Issue #147, opened 2020-08-06
Support Prometheus /-/reload endpoint


rfratto:
When running in non-scraping-service mode, currently the Agent has to be completely restarted to apply updates made to the config file. A `/-/reload` endpoint that functions like Prometheus' should be added.

Related: #140

amckinley:
+1 for this feature!

rfratto:
I'm starting to work on this this week. I suspect it'll be available in the 0.13.0 release.

PR #461, opened 2021-03-10
Add pkg/prom/cluster package with membership management

rfratto:
#### PR Description 
This is the first of a few PRs to migrate the `pkg/prom/ha` logic to a `pkg/prom/cluster` package.
This first PR re-implements membership within the cluster to a dedicated `node` type. The `node` type will be used for ensuring the Agent maintains appropriate membership in the proper cluster as the config changes. 

Reshards are triggered during the membership cycle:

- All Agents will be told to reshard when a node joins (including the joining Agent)
- All Agents will be told to reshard when a node leaves (excluding the leaving Agent)

Believe it or not, still working on #147.

#### Notes to the Reviewer
This isn't hooked into anything yet, and won't be hooked up until all pieces are migrated. Tests still needed, so I'm opening this in draft. 

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/prom/cluster/config.go
+++ b/pkg/prom/cluster/config.go
@@ -45,4 +45,5 @@ func (c *Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
 	f.DurationVar(&c.ReshardTimeout, prefix+"reshard-timeout", time.Second*30, "timeout for cluster-wide reshards and local reshards. Timeout of 0s disables timeout.")
 	c.KVStore.RegisterFlagsWithPrefix(prefix+"config-store.", "configurations/", f)
 	c.Lifecycler.RegisterFlagsWithPrefix(prefix, f)
+	c.Client.GRPCClientConfig.RegisterFlagsWithPrefix(prefix, f)
 }
--- a/pkg/prom/cluster/config.go
+++ b/pkg/prom/cluster/config.go
@@ -1,0 +1,48 @@
+package cluster
+
+import (
+	"flag"
+	"time"
+
+	"github.com/cortexproject/cortex/pkg/ring"
+	"github.com/cortexproject/cortex/pkg/ring/kv"
+	"github.com/grafana/agent/pkg/prom/ha/client"
+	flagutil "github.com/grafana/agent/pkg/util"
+)
+
+// DefaultConfig provides default values for the config
+var DefaultConfig = *flagutil.DefaultConfigFromFlags(&Config{}).(*Config)
+
+// Config describes how to instantiate a scraping service Server instance.
+type Config struct {
+	Enabled         bool                  `yaml:"enabled"`
+	ReshardInterval time.Duration         `yaml:"reshard_interval"`
+	ReshardTimeout  time.Duration         `yaml:"reshard_timeout"`
+	Client          client.Config         `yaml:"client"`
+	KVStore         kv.Config             `yaml:"kvstore"`
+	Lifecycler      ring.LifecyclerConfig `yaml:"lifecycler"`
+}
+
+// UnmarshalYAML implements yaml.Unmarshaler.
+func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	*c = DefaultConfig
+
+	type plain Config
+	return unmarshal((*plain)(c))
+}
+
+// RegisterFlags adds the flags required to config the Server to the given
+// FlagSet.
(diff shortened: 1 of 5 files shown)


rfratto:
The tests are pretty flaky, CI might fail here. I'll look into why it flakes tomorrow.

rfratto:
Ah, it's flaking because it takes a little bit of time for the lifecycler to join the ring in a separate goroutine. ApplyConfig should probably find a way to wait to verify that the node is in the ring.

pkg/prom/cluster/config.go
```diff
@@ -0,0 +1,49 @@
...
+
+// DefaultConfig provides default values for the config
+var DefaultConfig = *flagutil.DefaultConfigFromFlags(&Config{}).(*Config)
+
+// Config describes how to instantiate a scraping service Server instance.
+type Config struct {
+	Enabled         bool                  `yaml:"enabled"`
+	ReshardInterval time.Duration         `yaml:"reshard_interval"`
+	ReshardTimeout  time.Duration         `yaml:"reshard_timeout"`
+	Client          client.Config         `yaml:"client"`
```
rfratto, line 21: I moved the client config here, which will be a change that I need to handle when this is wired in.

rfratto:
Ready for review. I don't think I can make the implementation any simpler than it is now. I tried finding a better way to handle the tests, but I'm not sure there's a ton of room for improvement, even if they are a little ugly.

rfratto:
cc @56quarters since this will eventually be the new scraping service package.

pkg/prom/cluster/node.go
```diff
@@ -0,0 +1,327 @@
...
+		return fmt.Errorf("failed to start ring: %w", err)
+	}
+	n.ring = r
+
+	lc, err := ring.NewLifecycler(cfg.Lifecycler, n, "agent", agentKey, false, n.reg)
+	if err != nil {
+		return fmt.Errorf("failed to create lifecycler: %w", err)
+	}
+	if err := services.StartAndAwaitRunning(context.Background(), lc); err != nil {
+		r.StopAsync()
```
mattdurham, line 126: Should this block?

pkg/prom/cluster/node.go
```diff
@@ -0,0 +1,327 @@
...
+	}
+
+	level.Info(n.log).Log("msg", "node run loop exiting")
+}
+
+// performClusterReshard informs the cluster to immediately trigger a reshard
+// of their workloads. if includeSelf is true, the server provided to newNode will
+// also be informed. includeSelf should be true when joining the cluster, and false
+// when leaving.
+func (n *node) performClusterReshard(ctx context.Context, includeSelf bool) error {
```
mattdurham, line 202: Thoughts on `includeSelf` renaming to `isJoiningCluster` or similar. Feels more descriptive to the comment.

pkg/prom/cluster/node.go
```diff
@@ -0,0 +1,327 @@
...
+	return firstError
+}
+
+// Flush implements ring.FlushTransferer. It's a no-op.
+func (n *node) Flush() {}
+
+// TransferOut implements ring.FlushTransferer. It connects to all other healthy agents and
+// tells them to reshard. TransferOut should NOT be called manually unless the mutex is
+// held.
+func (n *node) TransferOut(ctx context.Context) error {
```
mattdurham, line 324: Is there a reason we don't hold the mutex here?

rfratto: Yep, if we hold the mutex here then it deadlocks with `ApplyConfig` when stopping the previous lifecycler 🙃. I'm not thrilled about this, but I can't think of a better way of handling it.

mattdurham: Brainstorming, since this looks like it is only called when leaving the ring and not called again under normal use can we use [Once](https://golang.org/pkg/sync/#Once) to help in anyway?

mattdurham (approved):

pkg/prom/cluster/node.go
rfratto: Nah, that would normally help but not in this case. Every time you call `ApplyConfig` you will leave the old ring from the old config before joining the new ring with the new config, so `sync.Once` would break here.

pkg/prom/cluster/node.go
rfratto: I don't think it would matter much here, other than it being a little bit slower, but I'll make it blocking to be safe.

PR #462, opened 2021-03-11
pkg/prom/cluster: add config watcher

rfratto:
#### PR Description 
`configWatcher` is an internal component that watches a ConfigStore and applies config as they get changed. It also has a polling mechanism to do a full sync against the state of the config store. This is the other half of the existing functionality of `pkg/prom/ha` and pairs with the work from #461. 

#### Notes to the Reviewer
Still doing #147 🙃 

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/prom/cluster/config_watcher.go
+++ b/pkg/prom/cluster/config_watcher.go
@@ -11,6 +11,15 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/grafana/agent/pkg/prom/instance"
 	"github.com/grafana/agent/pkg/prom/instance/configstore"
+	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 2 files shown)


pkg/prom/cluster/config_watcher.go
```diff
@@ -0,0 +1,249 @@
...
+func (w *configWatcher) Stop() error {
+	w.cfgMut.Lock()
+	defer w.cfgMut.Unlock()
+
+	if w.stopped {
+		return fmt.Errorf("already stopped")
+	}
+	w.stopped = true
+
+	w.stop()
```
mattdurham, line 236: Nit: should the setting of stopped and call be switched?

mattdurham (approved):

PR #470, opened 2021-03-16
Glue together the new cluster package

rfratto:
#### PR Description 
Glues together the new cluster package.

#### Notes to the Reviewer
Notably no integration tests here. I think enough of the logic is covered to fall back on the smoke tests, which I'll do in the next PR when I replace the old package with the new one. 

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/prom/cluster/node.go
+++ b/pkg/prom/cluster/node.go
@@ -3,6 +3,8 @@ package cluster
 import (
 	"context"
 	"fmt"
+	"hash/fnv"
+	"net/http"
 	"sync"
 	"time"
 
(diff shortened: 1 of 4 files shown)


pkg/prom/cluster/node.go
```diff
@@ -137,37 +140,6 @@ func (n *node) ApplyConfig(cfg Config) error {
 	return nil
 }
 
-// WaitJoined waits for the node the join the cluster and enter the
```
rfratto, line 140: I moved this function down towards the end of the file so everything here was in usage-order.

mattdurham (approved):

PR #435, opened 2021-02-26
pkg/prom: implement ApplyConfig

rfratto:
#### PR Description 
This PR implements ApplyConfig for the Prometheus instance manager. 

#### Which issue(s) this PR fixes 
Partial implementation of #147.  
Similar PRs: #416, #424, #427. 

#### Notes to the Reviewer
The behavior here is easily the most drastic: changing the scraping service config can only be handled by leaving the cluster, changing the config, and rejoining. 

#### PR Checklist

- [x] CHANGELOG updated (internal change, not needed) 
- [x] Documentation added (internal change, not needed)
- [x] Tests updated

PR #473, opened 2021-03-17
implement ApplyConfig for pkg/prom

rfratto:
#### PR Description 
Implements ApplyConfig for `pkg/prom.Agent`. 

Related to #147.

#### Notes to the Reviewer
Alternative implementation of #435.

I don't really like how this works, and needing to have a goroutine to apply changed configs in the background feels bad. I've given some thought on other ways to do this, but this seems like the less ridiculous so far. 

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/prom/agent.go
+++ b/pkg/prom/agent.go
@@ -269,14 +269,14 @@ func (a *Agent) syncInstances(oldConfig, newConfig Config) {
 	a.mut.RLock()
 	defer a.mut.RUnlock()
 
-	// Apply new configs from cfg
+	// Apply the new configs
 	for _, c := range newConfig.Configs {
 		if err := a.mm.ApplyConfig(c); err != nil {
 			level.Error(a.logger).Log("msg", "failed to apply config", "name", c.Name, "err", err)
 		}
 	}
 
-	// Remove configs from a.cfg that are no longer in cfg
+	// Remove any configs from oldConfig that aren't in newConfig.
 	for _, oc := range oldConfig.Configs {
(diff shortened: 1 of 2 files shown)


pkg/prom/agent.go
```diff
@@ -202,13 +195,113 @@ func (a *Agent) newInstance(c instance.Config) (instance.ManagedInstance, error)
...
+
+	if err := a.cluster.ApplyConfig(cfg.ServiceConfig); err != nil {
+		return fmt.Errorf("failed to apply cluster config: %w", err)
+	}
+
+	// Queue an actor in the background to sync the instances. This is required
+	// because creating both this function and newInstance grab the mutex.
+	oldConfig := a.cfg
+
+	a.actor <- func() {
```
mattdurham, line 257: As you mentioned this feels less than ideal. This is fine if this is the best way to go forward given the situation, but feels like something we may have to pay off later.

mattdurham (approved):

pkg/prom/agent.go
rfratto: I'm thinking about it some more and we'll eventually be able to get rid of it once we consolidate the operational modes here and make the Agent always act as if it was clustered. We'd have a `configstore.Local` implementation and the config watcher would continue to sync the state with the local store in the background. It's still not great that it'll always be asynchronous. Maybe once we make that change another opportunity to simplify will pop up 🤞

Issue #466, opened 2021-03-15
Tempo: Support Mirroring a traffic pipeline


joe-elliott:
Tempo currently allows multiple independent pipelines, but it would be nice if we could also add the ability to mirror a single pipeline to two different backends.  The underlying OTel Collector technology supports this so it would simply be a matter of setting up the correct configuration.

Here is an example mirrored pipeline configuration:
```
receivers:
  jaeger:
    protocols:
      grpc:
exporters:
  otlp:
    endpoint: example.com:12345
  otlp/2:
    endpoint: otherexample.com:12345
service:
  pipelines:
    traces:
      exporters: ["otlp","otlp/2"]
      processors: []
      receivers: ["jaeger"]
```

rfratto:
👍 I'd like to see this implemented as a `remote_write` block and deprecate `push_config`.

ponson-thankavel-finastra:
Is it possible to support write_relabel_configs as part of push_config? below config example illustrates the requirement

    tempo:
      configs:
        - name: tempo-config
           push_config: 
            - name: tenant1
              basic_auth:
                username: 12345
                password_file: /etc/agent/secrets/tenant1/password
              endpoint: tempo-us-central1.grafana.net
              write_relabel_configs:
                - action: keep
                  regex: tenant1-ns1|tenant1-ns2
                  source_labels:
                    - namespace
            - name: tenant2
              basic_auth:
                username: 45678
                password_file: /etc/agent/secrets/tenant1/password
              endpoint: tempo-us-central1.grafana.net
              write_relabel_configs:
                - action: keep
                  regex: tenant2-ns1|tenant2-ns2
                  source_labels:
                    - namespace
          receivers:
            otlp:
              protocols:
                grpc: 
                  endpoint: "localhost:55680"
          scrape_configs:
            - bearer_token_file: /var/run/secrets/kubernetes.io/serviceaccount/token
              job_name: kubernetes-pods
              kubernetes_sd_configs:
                - role: pod

rfratto:
Hi, we don't support that today but I'm open for discussion. Can you make a new issue for the feature request?

ponson-thankavel-finastra:
@rfratto created this issue - https://github.com/grafana/agent/issues/502

PR #474, opened 2021-03-18
Support mirror traffic pipelines

mapno:
#### PR Description 

Adds support for multiple backends in a tracing pipeline. `push_config` is deprecated in favor of `remote_write` and `batch`, which configure the multiple exporters.

`remote_write` still uses `PushConfig`, but the idea is to rename it and remove the `Batch` field once the `push_config` is removed from the config.

Closes https://github.com/grafana/agent/issues/466

#### PR Checklist

- [X] CHANGELOG updated 
- [x] Documentation added
- [X] Tests updated

### DIFF
--- a/pkg/tempo/config.go
+++ b/pkg/tempo/config.go
@@ -97,10 +97,9 @@ type PushConfig struct {
 	Insecure           bool                   `yaml:"insecure"`
 	InsecureSkipVerify bool                   `yaml:"insecure_skip_verify"`
 	BasicAuth          *prom_config.BasicAuth `yaml:"basic_auth,omitempty"`
+	Batch              map[string]interface{} `yaml:"batch,omitempty"`            // https://github.com/open-telemetry/opentelemetry-collector/blob/7d7ae2eb34b5d387627875c498d7f43619f37ee3/processor/batchprocessor/config.go#L24
 	SendingQueue       map[string]interface{} `yaml:"sending_queue,omitempty"`    // https://github.com/open-telemetry/opentelemetry-collector/blob/7d7ae2eb34b5d387627875c498d7f43619f37ee3/exporter/exporterhelper/queued_retry.go#L30
 	RetryOnFailure     map[string]interface{} `yaml:"retry_on_failure,omitempty"` // https://github.com/open-telemetry/opentelemetry-collector/blob/7d7ae2eb34b5d387627875c498d7f43619f37ee3/exporter/exporterhelper/queued_retry.go#L54
-	// Deprecated. Configure the batch processor with InstanceConfig.Batch instead.
-	Batch map[string]interface{} `yaml:"batch,omitempty"` // https://github.com/open-telemetry/opentelemetry-collector/blob/7d7ae2eb34b5d387627875c498d7f43619f37ee3/processor/batchprocessor/config.go#L24
 }
 
 // UnmarshalYAML implements yaml.Unmarshaler.
--- a/pkg/tempo/config.go
+++ b/pkg/tempo/config.go
@@ -61,10 +61,17 @@ func (c *Config) Validate() error {
 type InstanceConfig struct {
 	Name string `yaml:"name"`
 
+	// Deprecated in favor of RemoteWrite and Batch.
 	PushConfig PushConfig `yaml:"push_config"`
+
+	// RemoteWrite defines one or multiple backends that can receive the pipeline's traffic.
+	RemoteWrite []PushConfig `yaml:"remote_write"`
 
 	// Receivers: https://github.com/open-telemetry/opentelemetry-collector/blob/7d7ae2eb34b5d387627875c498d7f43619f37ee3/receiver/README.md
 	Receivers map[string]interface{} `yaml:"receivers"`
+
+	// Batch: https://github.com/open-telemetry/opentelemetry-collector/blob/7d7ae2eb34b5d387627875c498d7f43619f37ee3/processor/batchprocessor/config.go#L24
+	Batch map[string]interface{} `yaml:"batch,omitempty"`
 
 	// Attributes: https://github.com/open-telemetry/opentelemetry-collector/blob/7d7ae2eb34b5d387627875c498d7f43619f37ee3/processor/attributesprocessor/config.go#L30
 	Attributes map[string]interface{} `yaml:"attributes"`
@@ -90,9 +97,10 @@ type PushConfig struct {
 	Insecure           bool                   `yaml:"insecure"`
 	InsecureSkipVerify bool                   `yaml:"insecure_skip_verify"`
 	BasicAuth          *prom_config.BasicAuth `yaml:"basic_auth,omitempty"`
-	Batch              map[string]interface{} `yaml:"batch,omitempty"`            // https://github.com/open-telemetry/opentelemetry-collector/blob/7d7ae2eb34b5d387627875c498d7f43619f37ee3/processor/batchprocessor/config.go#L24
-	SendingQueue       map[string]interface{} `yaml:"sending_queue,omitempty"`    // https://github.com/open-telemetry/opentelemetry-collector/blob/7d7ae2eb34b5d387627875c498d7f43619f37ee3/exporter/exporterhelper/queued_retry.go#L30
-	RetryOnFailure     map[string]interface{} `yaml:"retry_on_failure,omitempty"` // https://github.com/open-telemetry/opentelemetry-collector/blob/7d7ae2eb34b5d387627875c498d7f43619f37ee3/exporter/exporterhelper/queued_retry.go#L54
+	// Deprecated. Configure the batch processor with InstanceConfig.Batch instead.
+	Batch          map[string]interface{} `yaml:"batch,omitempty"`            // https://github.com/open-telemetry/opentelemetry-collector/blob/7d7ae2eb34b5d387627875c498d7f43619f37ee3/processor/batchprocessor/config.go#L24
+	SendingQueue   map[string]interface{} `yaml:"sending_queue,omitempty"`    // https://github.com/open-telemetry/opentelemetry-collector/blob/7d7ae2eb34b5d387627875c498d7f43619f37ee3/exporter/exporterhelper/queued_retry.go#L30
+	RetryOnFailure map[string]interface{} `yaml:"retry_on_failure,omitempty"` // https://github.com/open-telemetry/opentelemetry-collector/blob/7d7ae2eb34b5d387627875c498d7f43619f37ee3/exporter/exporterhelper/queued_retry.go#L54
 }
 
 // UnmarshalYAML implements yaml.Unmarshaler.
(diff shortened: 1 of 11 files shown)


mapno:
Wasn't really sure how to proceed with all the changes needed in `/docs` and `/production`. Also didn't want to go too far before checking that the approach is valid. Let me know what you think.

pkg/tempo/config.go
```diff
@@ -90,9 +97,10 @@ type PushConfig struct {
...
 	InsecureSkipVerify bool                   `yaml:"insecure_skip_verify"`
 	BasicAuth          *prom_config.BasicAuth `yaml:"basic_auth,omitempty"`
-	Batch              map[string]interface{} `yaml:"batch,omitempty"`            // https://github.com/open-telemetry/opentelemetry-collector/blob/7d7ae2eb34b5d387627875c498d7f43619f37ee3/processor/batchprocessor/config.go#L24
-	SendingQueue       map[string]interface{} `yaml:"sending_queue,omitempty"`    // https://github.com/open-telemetry/opentelemetry-collector/blob/7d7ae2eb34b5d387627875c498d7f43619f37ee3/exporter/exporterhelper/queued_retry.go#L30
-	RetryOnFailure     map[string]interface{} `yaml:"retry_on_failure,omitempty"` // https://github.com/open-telemetry/opentelemetry-collector/blob/7d7ae2eb34b5d387627875c498d7f43619f37ee3/exporter/exporterhelper/queued_retry.go#L54
+	// Deprecated. Configure the batch processor with InstanceConfig.Batch instead.
+	Batch          map[string]interface{} `yaml:"batch,omitempty"`            // https://github.com/open-telemetry/opentelemetry-collector/blob/7d7ae2eb34b5d387627875c498d7f43619f37ee3/processor/batchprocessor/config.go#L24
```
rfratto, line 101: Can you move this below the other fields to make it clear that the deprecation only applies to Batch?

rfratto (commented):
I think this is pretty reasonable but I'm going to tag @joe-elliott for a review. 

Does batch need to be applied across all the exporters or is it possible to have (or want) per-endpoint batch settings?

pkg/tempo/config.go
```diff
@@ -167,10 +169,61 @@ func (c *InstanceConfig) otelConfig() (*configmodels.Config, error) {
...
+
+	if len(c.Receivers) == 0 {
+		return nil, errors.New("must have at least one configured receiver")
+	}
+
+	if len(c.RemoteWrite) == 0 && len(c.PushConfig.Endpoint) == 0 {
+		return nil, errors.New("must have a configured at least one remote_write or push_config.endpoint")
+	}
+
+	if len(c.RemoteWrite) != 0 && len(c.PushConfig.Endpoint) != 0 {
```
joe-elliott, line 208: instead of testing this we could just concatenate PushConfig with the new RemoteWrite slice?  we could remove this error checking.

maybe just have a 
```
if len(c.PushConfig.Endpoint) > 0 {
  // print friendly deprecation warning here.
}
```

pkg/tempo/config.go
```diff
@@ -167,10 +169,61 @@ func (c *InstanceConfig) otelConfig() (*configmodels.Config, error) {
...
+		return nil, errors.New("must not configure push_config and remote_write. push_config is deprecated in favor of remote_write")
 	}
 
+	if c.Batch != nil && c.PushConfig.Batch != nil {
```
joe-elliott, line 212: we need to move batch out of the PushConfig and onto the InstanceConfig. InstanceConfig represents a single otel processor pipeline and we only want one batch processor per pipeline.

with these changes PushConfig represents an otel exporter.

joe-elliott (changes requested):

pkg/tempo/config.go
```diff
@@ -110,49 +118,43 @@ func (c *PushConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
-func (c *InstanceConfig) otelConfig() (*configmodels.Config, error) {
-	otelMapStructure := map[string]interface{}{}
-
-	if len(c.Receivers) == 0 {
-		return nil, errors.New("must have at least one configured receiver")
-	}
-
-	if len(c.PushConfig.Endpoint) == 0 {
+// exporter builds an OTel exporter from PushConfig
+func (c *InstanceConfig) exporter(remoteWriteConfig PushConfig) (map[string]interface{}, error) {
```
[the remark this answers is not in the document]
mapno: You're right, removed!

pkg/tempo/config.go
mapno: Nice suggestion. Will do!

pkg/tempo/config.go
```diff
@@ -167,10 +169,61 @@ func (c *InstanceConfig) otelConfig() (*configmodels.Config, error) {
...
+}
+
+func (c *InstanceConfig) otelConfig() (*configmodels.Config, error) {
+	otelMapStructure := map[string]interface{}{}
+
+	if len(c.Receivers) == 0 {
+		return nil, errors.New("must have at least one configured receiver")
+	}
+
+	if len(c.RemoteWrite) == 0 && len(c.PushConfig.Endpoint) == 0 {
```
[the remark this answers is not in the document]
mapno: yea, it is a redundant check. nice catch, thanks!

pkg/tempo/config.go
mapno: That's a great idea. I hadn't thought of that.

However, I'm not entirely convinced.  I think it could result in unexpected or undesired behaviour when defining exporters with both `push_config` and `remote_write`. I think it's convenient to check that only one of the methods is being used.

pkg/tempo/config.go
mapno: Oh totally.  Right now it is staying there to support backwards compatibility with `push_config`.  It is meant to be removed in the future.  If we want to remove it now, I'm all for it.

This check is to make sure only one of the configs is being used.

joe-elliott: I mean that RemoteWrite is a slice of PushConfigs which represents a set of exporters on a single OTel pipeline.  Batch is only supported at the pipeline level and not at the exporter level. So we need to move Batch out of the PushConfig and onto the InstanceConfig.

mapno:
Applied the suggested changes and updated the config examples, as well as the config in `/production`.

pkg/tempo/config.go
```diff
@@ -90,9 +97,10 @@ type PushConfig struct {
 	Insecure           bool                   `yaml:"insecure"`
 	InsecureSkipVerify bool                   `yaml:"insecure_skip_verify"`
 	BasicAuth          *prom_config.BasicAuth `yaml:"basic_auth,omitempty"`
-	Batch              map[string]interface{} `yaml:"batch,omitempty"`            // https://github.com/open-telemetry/opentelemetry-collector/blob/7d7ae2eb34b5d387627875c498d7f43619f37ee3/processor/batchprocessor/config.go#L24
```
joe-elliott, line 93: Since PushConfig and []RemoteWriteConfigs are mutually exclusive, I think we should leave Batch in the PushConfig for backwards compat?

Either the user configures the old way which doesn't allow split pipelines and the BatchConfig can be directly in the push config or they configure the new way which does allow split pipelines and the BatchConfig has to be on the InstanceConfig.

joe-elliott (commented):
One final thought about backwards compat.

pkg/tempo/config.go
mapno: Yes, you're right. Addressed in 7dcc416f428e935f28f824e99973b46c952f4255

pkg/tempo/config.go
```diff
@@ -110,49 +117,74 @@ func (c *PushConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+func (c *RemoteWriteConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	*c = DefaultRemoteWriteConfig
+
+	type plain RemoteWriteConfig
+	if err := unmarshal((*plain)(c)); err != nil {
+		return err
+	}
+
+	if c.Compression != compressionGzip && c.Compression != compressionNone {
+		return fmt.Errorf("unsupported compression '%s', expected 'gzip' or 'none'", c.Compression)
```
rfratto, line 146: Something I learned about recently :) 
```suggestion
		return fmt.Errorf("unsupported compression %q, expected 'gzip' or 'none'", c.Compression)
```

pkg/tempo/config.go
```diff
@@ -167,10 +199,65 @@ func (c *InstanceConfig) otelConfig() (*configmodels.Config, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		exporterName := fmt.Sprintf("otlp/%d", i)
+		exporters[exporterName] = exporter
+	}
+	return exporters, nil
+}
+
+func (c *InstanceConfig) otelConfig() (*configmodels.Config, error) {
```
rfratto, line 235: Can we log a warning somewhere here if push_config is configured to note that it's deprecated in favor of remote_write?

rfratto (approved):
Thanks for changing the Tanka stuff too, great work :) 

I think my open question from last time got missed. What are the odds someone is going to watch different batch settings per remote_write? Prometheus does this with its queue_config settings. If the same thing doesn't make as much sense for traces, then I think this PR is fine as-is. Otherwise, I think we might want to consider setting up one pipeline per exporter, maybe in a follow-up PR. 

(I'm not sure if that actually works though, i.e., if OTel lets you re-use receivers across multiple pipelines.)

joe-elliott (approved):

pkg/tempo/config.go
rfratto: Actually, it would make the error look worse by mixing double quotes and single quotes. 🤷 Going to mark this as resolved.

mapno:
> Thanks for changing the Tanka stuff too, great work :)
> 
> I think my open question from last time got missed. What are the odds someone is going to watch different batch settings per remote_write? Prometheus does this with its queue_config settings. If the same thing doesn't make as much sense for traces, then I think this PR is fine as-is. Otherwise, I think we might want to consider setting up one pipeline per exporter, maybe in a follow-up PR.
> 
> (I'm not sure if that actually works though, i.e., if OTel lets you re-use receivers across multiple pipelines.)

Yes, sorry I totally missed it before. I'm not sure how often people are going to want per-exporter batch configs, but is something doable. As you mention, the OTel collector lets you set up pipelines re-using components (i.e. receivers, processors, exporters) as much as you'd like.

I think that different batch configs is most interesting when the exporters are different, which for the agent is not the case (we're always using the otlp exporter). But maybe @joe-elliott can give a better answer since he has more experience operating it.

pkg/tempo/config.go
mapno: Added in 53d7f969f43d38d6a36a49da13a6604505da6bf9

joe-elliott:
The PR as is covers the necessary cases. The OTEL collector does not allow a batch processor per exporter in a mirrored setup. There can be only one batch exporter for the pipeline that is then mirrored to the two exporters.

Independent pipelines can (and should) have independent batch processors but this is already covered at the `[]InstanceConfig` level.

rfratto:
Thanks for the input! Merging. Thanks again @mapno!

Issue #120, opened 2020-06-11
Generate configuration-reference.md


rfratto:
The [configuration reference](https://github.com/grafana/agent/blob/master/docs/configuration-reference.md) document is currently hand written, large, and subject to change when updating vendors. It would be nice to have a script that autogenerates the entire document. [Cortex has something like this](https://github.com/cortexproject/cortex/tree/master/tools/doc-generator) for its equivalent config reference that could be used for inspiration.

quinndiggity:
Just a note, there are a *ton* of references in the https://grafana.com docs that refer to items like https://github.com/grafana/agent/blob/master/docs/configuration-reference.md which are all 404'ing - for example, several of the first pieces of documentation users see, one of which being https://grafana.com/docs/grafana-cloud/how-do-i/agent/

marctc:
@rfratto I didn't find this document anymore, so I guess the problem is already fixed.

PR #475, opened 2021-03-18
Windows exporter integration

### LINKED ISSUE #285: Add Windows Exporter to Grafana Agent integrations
Since windows does still have a large footprint in the server world it would be great to have the Windows Exporter integration added to the Grafana agent as well. This would also allow for wider adoption across enterprise environments that usually have a mix of both Linux and Windows servers and would also allow for a smooth transition should they move from on-prem to Grafana Cloud or vice-versa.

mattdurham:
#### PR Description 

This add windows_exporter to the list of default exporters for the windows platform. This currently uses a fork of windows_exporter located at github.com/grafana/windows_exporter. It uses the same type of configuration has the normal windows_exporter.  

Closes #285 

#### PR Checklist

- [X] CHANGELOG updated 
- [X] Documentation added <- In separate [PR](https://github.com/mattdurham/agent/pull/3)
- [X] Tests updated


go.mod
```diff
@@ -92,3 +93,5 @@ replace github.com/gomodule/redigo => github.com/gomodule/redigo v1.8.2
 
 // Excluding fixes a conflict in test packages and allows "go mod tidy" to run.
 exclude google.golang.org/grpc/examples v0.0.0-20200728065043-dfc0c05b2da9
+
+replace github.com/prometheus-community/windows_exporter => github.com/grafana/windows_exporter v0.15.1-0.20210318175846-f7531024d7fc
```
rfratto, line 97: Can you move this into the TODO block with all the other exporter-related replaces?

pkg/integrations/install/install_windows.go
```diff
@@ -0,0 +1,5 @@
+// +build windows
+
+package install
+
+import _ "github.com/grafana/agent/pkg/integrations/windows_exporter" // register windows_exporter
```
rfratto, line 5: It's likely that people are going to try to use the same Agent config for all their projects. Can you include this in the normal install and create a no-op integration like I did for process_exporter?

pkg/integrations/windows_exporter/config.go
```diff
@@ -0,0 +1,40 @@
+// +build windows
```
rfratto, line 1: This doesn't need to be marked as windows-only since there's nothing Windows specific in here.

pkg/integrations/windows_exporter/config.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/grafana/agent/pkg/integrations/config"
+)
+
+// Config controls the node_exporter integration.
+type Config struct {
+	Common config.Common `yaml:",inline"`
+
+	WindowsConfig interface{} `yaml:"config"`
+
+	EnabledCollectors string `yaml:"enabledCollectors"`
```
rfratto, line 17: `enabledCollectors` should be snake_cased to be consistent with our other options.

pkg/integrations/windows_exporter/windows_exporter.go
```diff
@@ -0,0 +1,91 @@
...
+// New creates a new node_exporter integration.
+func New(log log.Logger, c *Config) (*Integration, error) {
+
+	bytes, _ := yaml.Marshal(c.WindowsConfig)
+	cb := string(bytes)
+	wc, _ := exporter.NewWindowsCollector(c.Name(), c.EnabledCollectors, cb)
+
+	level.Info(log).Log("msg", "Enabled windows_exporter collectors")
+
+	return &Integration{
```
rfratto, line 38: It looks like you can use NewCollectorIntegration here since there's nothing specific you need to do for the handler.

pkg/integrations/windows_exporter/windows_exporter.go
```diff
@@ -0,0 +1,91 @@
...
+
+	exporterMetricsRegistry *prometheus.Registry
+}
+
+// New creates a new node_exporter integration.
+func New(log log.Logger, c *Config) (*Integration, error) {
+
+	bytes, _ := yaml.Marshal(c.WindowsConfig)
+	cb := string(bytes)
+	wc, _ := exporter.NewWindowsCollector(c.Name(), c.EnabledCollectors, cb)
```
rfratto, line 34: I'm not super thrilled about unmarshaling to an `interface{}` and marshaling it back. It's going to be more code, but I'd rather we re-define some structs at this level and use it to configure whatever we need to at the collector. The main benefit of doing this is a config generator (#120) would work, plus errors would be more readable.

pkg/integrations/windows_exporter/windows_exporter.go
```diff
@@ -0,0 +1,91 @@
...
+	logger log.Logger
+	wc     *exporter.WindowsCollector
+
+	exporterMetricsRegistry *prometheus.Registry
+}
+
+// New creates a new node_exporter integration.
+func New(log log.Logger, c *Config) (*Integration, error) {
+
+	bytes, _ := yaml.Marshal(c.WindowsConfig)
```
rfratto, line 32: Let's not ignore any errors here, please 🙏

rfratto (commented):
Looking good! I wanted to give this a review before I take some PTO.

pkg/integrations/windows_exporter/config.go
```diff
@@ -0,0 +1,128 @@
...
+
+func (c *Config) ConvertToMap() map[string]string {
+	configMap := make(map[string]string)
+	mapToConfig(c, configMap)
+	return configMap
+}
+
+// This is to create a map from the config using the exporter struct tag as the key. This is to be ONLY used on pointer configs
+// and pointer strings. This is because we need to know if a value was set so that we don't overwrite defaults.
+func mapToConfig(config interface{}, cm map[string]string) {
```
mattdurham, line 55: This feels like technical debt and would love any feedback. Also going to check it out to see if I can simplify/harden it.

pkg/integrations/windows_exporter/config.go
```diff
@@ -0,0 +1,132 @@
...
+// The Windows Collector takes a map of configuration to set, so we need to convert from agent config to a key value
+// using the windows_exporter key name 'collector.iis.site-whitelist' for example.
+func (c *Config) ConvertToMap() map[string]string {
+	configMap := make(map[string]string)
+	mapToConfig(c, configMap)
+	return configMap
+}
+
+type ExchangeConfig struct {
+	EnabledList *string `yaml:"enabled_list" exporter:"collectors.exchange.enabled"`
```
rfratto, line 61: Do these need to be pointers?

pkg/integrations/windows_exporter/config.go
```diff
@@ -0,0 +1,132 @@
...
+		// Currently agent only support translating string pointer vars
+		if en != "" && valueField.Kind() == reflect.Ptr && valueField.Elem().Kind() == reflect.String {
+			cm[en] = valueField.Elem().String()
+			continue
+		}
+		if valueField.Kind() == reflect.Ptr {
+			mapToConfig(valueField.Interface(), cm)
+		}
+	}
+}
```
rfratto, line 132: In cases like these, I'd prefer to avoid reflection for a simpler, more verbose solution, especially recursive reflection. Can we find a way to retrieve the collector configs and manually map to their settings for each of our config structs? I can kind of imagine something like this:

```go
type LogicalDiskConfig struct {
  WhiteList string `yaml:"whitelist"` 
  BlackList string `yaml:"blacklist"` 
}

func (c LogicalDiskConfig) Sync(v interface{}) {
  if other, ok := v.(*exporter.LogicalDiskConfig); ok {
    other.WhiteList = c.WhiteList 
    other.BlackList = c.BlackList 
  }
}
```

This is more verbose and repetitive, but I believe it's simpler and also means we don't have to have a 1:1 mapping of fields.

rfratto (commented):
This is really coming along well, thank you for putting effort into this. I'm not super happy about reflection, and I'm worried its cognitive cost is too high compared to keeping the lines of code smaller.

pkg/integrations/windows_exporter/config.go
mattdurham: I think so. The issue arises that passing in "" would overwrite the defaults that windows_exporter uses. If there is another way to do it then would def like that route.

pkg/integrations/windows_exporter/config.go
```diff
@@ -0,0 +1,7 @@
+// +build !windows
+
+package windows_exporter //nolint:golint
+
+// Config placeholder config for dummy windows_exporter
+type Config struct {
+}
```
rfratto, line 7: This will cause problems if it's configured on non-Windows. Can you make the config struct cross-platform? You can make the ApplyConfig method Windows-only by putting it in its own file.

pkg/integrations/windows_exporter/config_windows.go
```diff
@@ -0,0 +1,226 @@
+package windows_exporter //nolint:golint
```
rfratto, line 1: Can you rebase? I think this is going to fail lints since it's missing godocs for some exported methods, and I don't want the lint to fail after the merge.

pkg/integrations/windows_exporter/windows_exporter.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/gorilla/mux"
+	"github.com/grafana/agent/pkg/integrations/config"
+)
+
+// Integration is the windows_exporter integration. On non-Windows platforms,
+// this integration does nothing and will print a warning if enabled.
+type Integration struct {
+	c *Config
```
rfratto, line 17: You don't need to store the config here, feel to just ignore both arguments to New.

rfratto (approved):
LGTM, though this needs to be rebased to pick up the latest changes to the integrations interface, otherwise it's going to fail to build after merging.

PR #486, opened 2021-03-24
ApplyConfig for integrations

rfratto:
#### PR Description 
Implements ApplyConfig for integrations.

This requires a change to the integration interface to no longer register routes directly to mux. This is because mux is unable to unregister routes. The new approach is to expose the HTTP handlers that expose metrics, and look up the handler at request time.

Related to #147. 

#### Notes to the Reviewer
Everything still seems to work after local testing. The new approach here is pretty similar, but stores the goroutines in a map (kind of like the basic instance manager) so they can be stopped independently. 

One interesting detail here: when "updating" an integration (stopping the old goroutine and starting a new one), the new goroutine does not wait for the previous one to stop. I don't believe this will cause any problems, but we might have to change this behavior later. 

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/integrations/manager.go
+++ b/pkg/integrations/manager.go
@@ -8,6 +8,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
 	config_util "github.com/prometheus/common/config"
 
 	"github.com/go-kit/kit/log"
@@ -124,37 +125,25 @@ func (c *ManagerConfig) DefaultRelabelConfigs() ([]*relabel.Config, error) {
 type Manager struct {
 	c                     ManagerConfig
 	logger                log.Logger
-	integrations          map[Config]Integration
 	hostname              string
 	defaultRelabelConfigs []*relabel.Config
 
+	ctx    context.Context
+	cancel context.CancelFunc
+	wg     sync.WaitGroup
+
 	im        instance.Manager
-	cancel    context.CancelFunc
 	done      chan bool
 	validator configstore.Validator
+
+	integrationsMut sync.RWMutex
+	integrations    map[string]*integrationProcess
 }
 
 // NewManager creates a new integrations manager. NewManager must be given an
 // InstanceManager which is responsible for accepting instance configs to
 // scrape and send metrics from running integrations.
 func NewManager(c ManagerConfig, logger log.Logger, im instance.Manager, validate configstore.Validator) (*Manager, error) {
-	integrations := make(map[Config]Integration)
-
-	for _, integrationCfg := range c.Integrations {
-		if integrationCfg.CommonConfig().Enabled {
-			l := log.With(logger, "integration", integrationCfg.Name())
-			i, err := integrationCfg.NewIntegration(l)
-			if err != nil {
-				return nil, err
-			}
-			integrations[integrationCfg] = i
-		}
-	}
-
-	return newManager(c, logger, im, integrations, validate)
-}
-
-func newManager(c ManagerConfig, logger log.Logger, im instance.Manager, integrations map[Config]Integration, validate configstore.Validator) (*Manager, error) {
 	defaultRelabels, err := c.DefaultRelabelConfigs()
 	if err != nil {
 		return nil, fmt.Errorf("cannot get default relabel configs: %w", err)
@@ -165,12 +154,13 @@ func newManager(c ManagerConfig, logger log.Logger, im instance.Manager, integra
 	m := &Manager{
 		c:                     c,
 		logger:                logger,
-		integrations:          integrations,
 		defaultRelabelConfigs: defaultRelabels,
 		im:                    im,
-		cancel:                cancel,
 		done:                  make(chan bool),
 		validator:             validate,
+
+		ctx:    ctx,
+		cancel: cancel,
 	}
 
 	if c.UseHostnameLabel {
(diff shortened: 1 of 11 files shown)


pkg/integrations/manager.go
```diff
@@ -116,155 +113,238 @@ func (c *ManagerConfig) DefaultRelabelConfigs() ([]*relabel.Config, error) {
...
+			}
+			p.stop()
+			delete(m.integrations, key)
+		}
+
+		l := log.With(m.logger, "integration", ic.Name())
+		i, err := ic.NewIntegration(l)
 		if err != nil {
-			return nil, err
+			return fmt.Errorf("error initializing integration %q: %w", ic.Name(), err)
```
56quarters, line 209: What happens when this returns an error? Is it a problem that new configuration has been applied to some integrations an not others? Or is the idea that the agent would exit if this happened?

pkg/integrations/manager.go
```diff
@@ -116,155 +113,238 @@ func (c *ManagerConfig) DefaultRelabelConfigs() ([]*relabel.Config, error) {
...
+	case <-m.ctx.Done():
+		return fmt.Errorf("Manager already stopped")
+	default:
+		// No-op
 	}
 
-	if c.UseHostnameLabel {
-		var err error
-		m.hostname, err = instance.Hostname()
+	// Iterate over our integrations. New or changed integrations will be
```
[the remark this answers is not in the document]
rfratto: I'm comfortable changing this into an implementation that builds up a new map and swaps them, though it'll still be very similar to how it works now; we don't want to stop all existing integrations to prevent any gaps in the metrics for integrations that didn't change.

pkg/integrations/manager.go
rfratto: I've been thinking about this as I've been implementing these, and I see what you're getting at. It's a problem that this returns an early error, since it leaves the Agent in a hard to predict state, where all valid changes were applied, but all (potentially valid) pending changes were aborted after coming across an invalid change. I can change this to log each error and return the first error after iterating through all of them. 

An error returned here would eventually turn into a 4xx/5xx HTTP response for the `/-/reload` caller. 

Given we're going to call /-/reload a beta feature, I think we should commit to saying it always updates the state of components that received a valid configuration, but any invalid configuration will leave that component in an undefined 
state (either running with the last known config or shut down and no longer running). 

It's hard to commit to anything more than this, especially since it's not always possible to bring up the component with the new settings before shutting down the old one. One example of this is the Tempo receivers, which open up TCP/UDP ports and need to be shut down so the new config can take its place. 

I hacked up some quick docs below, let me know what you think with the (beta feature) commitment for what we can guarantee /-/reload can/can't do. 

> `/-/reload` is an endpoint that updates the various components of the Agent to reflect the newest config on disk. It is currently in beta, so please open issues for any problems you see or ways to improve the functionality, especially around 
> handing runtime errors. 
> 
> When `/-/reload` is invoked, the config file provided to the Agent's `-config.file` flag will be re-read. Invalid YAML, or 
> YAML that doesn't pass basic validation will be immediately rejected. Otherwise, all components will be updated to 
> match the state of the new config file. 
> 
> When updating component state, runtime errors can happen that wouldn't be caught when reading in the YAML. 
> For example, setting the WAL to an invalid directory would be caught at runtime, not when the config file is read. 
> Runtime errors that prevent a component from being update will leave that component in an undefined state, where it is 
> either completely shut down or running with its last known config. Runtime errors must be fixed by addressing the 
> problem in the configuration file and invoking `/-/reload` again. 
> 
> Runtime errors encountered while updating the state do not stop the update from completing, and all valid changes will
> always be applied. The Agent has many "components" that receive updates:
>
> 1. The Server config 
> 2. Each Prometheus instance 
> 3. Each Loki instance 
> 4. Each Tempo instance 
> 5. Each Integration
> 6. The scraping service (which is the clustering mechanism + the key value store)
>
> If a call to /-/reload returns an error, always examine the logs to learn exactly what went wrong and which components 
> have been shut down as a result of the runtime error.

pkg/integrations/manager.go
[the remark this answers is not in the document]
56quarters: Great point

pkg/integrations/manager.go
56quarters: I think that's a good compromise. "the component with invalid configuration is in an undefined state but other things are fine" feels like the right tradeoff when a user provides a bad config. As long as users can fix the config and call `/-/reload` again this seems like very straightforward behavior.

mattdurham: Is it possible/wise to reject the update entirely and re-apply the old config? IE should we treat this has a transactional update. If server config is in an invalid state, can that propagate into all the other components being in an undefined state, even if their individual config is correct?

rfratto: It's possible, and I can see arguments on both sides here. I've been using Prometheus as the reference implementation which does the same thing as what we're doing now. Having transactional reloads will be an implementation detail of /-/reload itself though, where it just re-applies the last valid config before returning the error back to the user. 

I have some concerns about transactional reloads being noisy at the cluster level if that's being used, but this is something that can be discussed later. 

> If server config is in an invalid state, can that propagate into all the other components being in an undefined state, even if their individual config is correct?

Kind of. It's hard to do this in a strict way, but other components depend on the listen address / port being valid and open when the server starts listening. So if the server config is invalid, stuff will just subtly break at runtime.

mattdurham: Sounds reasonable, especially if we let it chill in Beta for a bit and get feedback, with the knowledge to users that the semantics could change.

mattdurham (approved):

Issue #484, opened 2021-03-24
No support for Kafka receiver as mentioned in the documentation


jithinchandranj:
As per the blog post mentioned here, [https://grafana.com/blog/2020/11/17/tracing-with-the-grafana-agent-and-grafana-tempo/](url), receivers allow the Grafana Agent to accept tracing data from a multitude of systems. We currently support receiving spans from Jaeger, Kafka, OpenCensus, OTLP, and Zipkin.

When integrated with the Kafka receiver config, error is shown on startup that the receiver is not supported. When checked the source code further, could see that the Kafka receiver is not added into the list of supported receivers.

joe-elliott:
@mapno take a look when you get a chance.  _Should_ be relatively straightforward to add the kafka receiver.

Perhaps we want this in Tempo proper as well?

PR #489, opened 2021-03-25
Add kafka traces receiver

mapno:
#### PR Description 

Adds the kafka receiver 

Closes https://github.com/grafana/agent/issues/484

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated


joe-elliott (approved):

rfratto (approved):
LGTM! Before I merge, can you update the changelog?

@joe-elliott Typically we add all the receivers to the [install manifest](https://github.com/grafana/agent/blob/main/production/kubernetes/build/templates/tempo/main.jsonnet) (rendered as YAML [here](https://github.com/grafana/agent/blob/main/production/kubernetes/agent-tempo.yaml)), but since the Kafka receiver works so differently, I don't think we can do the same thing here. WDYT?

joe-elliott:
Yeah, agree, it doesn't belong in the install manifest. Is there a place that makes sense to document this new option?

rfratto:
Our docs here are really weak. IMO for now let's just have a list [in the docs for tempo_instance_config](https://github.com/grafana/agent/blob/main/docs/configuration-reference.md#tempo_instance_config), something like:

```yaml
# Receiver configurations are mapped directly into the OpenTelmetry receivers block.
#   At least one receiver is required. Supported receivers: otlp, jaeger, kafka, [...]
#   Documentation for each receiver can be found at https://github.com/open-telemetry/opentelemetry-collector/blob/7d7ae2eb34b5d387627875c498d7f43619f37ee3/receiver/README.md
receivers:
```

mapno:
Added the documentation and manually tested the receiver by modifying the docker-compose example. I didn't think it was worth modifying the code. Let me know if you want me to test it some other way (e.g. integration tests).

rfratto (approved):
Thanks again! I'll merge once the conflicts are resolved.

mapno:
> Thanks again! I'll merge once the conflicts are resolved.

They should be solved now

PR #490, opened 2021-03-25
Implement Applyconfig for Entrypoint

rfratto:
#### PR Description 
Implements ApplyConfig for entrypoint.

This required making a few changes to drastically simplify the implementation: all subsystems are now always enabled. This adds very little resource overhead, but makes the logs a little noisier. Prometheus will now only fail to validate if the WAL isn't configured when at least one instance, integration, or the scraping service is configured. 

Related to #147.

#### Notes to the Reviewer
This doesn't change the logger to reflect a change in logging level. I'll follow up with a second PR for that.

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated


pkg/integrations/manager.go
```diff
@@ -116,6 +109,35 @@ func (c *ManagerConfig) DefaultRelabelConfigs(hostname string) []*relabel.Config
...
+		if !ic.CommonConfig().Enabled {
+			continue
+		}
+
+		scrapeIntegration := c.ScrapeIntegrations
+		if common := ic.CommonConfig(); common.ScrapeIntegration != nil {
+			scrapeIntegration = *common.ScrapeIntegration
+		}
+		if scrapeIntegration {
+			willCreateInstance = true
```
mattdurham, line 130: Could you return early here instead of using the willCreateInstance?

mattdurham (approved):
Minor comment but LGTM

pkg/integrations/manager.go
rfratto: ah, good point!

PR #491, opened 2021-03-25
Allow changing log levels at runtime

rfratto:
#### PR Description 
Ability to reload the logger with a new log level.

Part of #147. 

#### Notes to the Reviewer
🤷 

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/util/logger.go
+++ b/pkg/util/logger.go
@@ -13,10 +13,11 @@ import (
 // Logger implements Go Kit's log.Logger interface. It supports being
 // dynamically updated at runtime.
 type Logger struct {
+	// mut protects against race conditions accessing l, which can be modified
+	// and accessed concurrently if ApplyConfig and Log are called at the same
+	// time.
 	mut sync.RWMutex
-
-	// l is the constructed log.Logger.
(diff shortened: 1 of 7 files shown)


pkg/util/logger.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return l, nil
+}
+
+// Log logs a log line.
+func (l *Logger) Log(kvps ...interface{}) error {
+	l.mut.RLock()
```
mattdurham, line 72: Is this needed? From my semi-coherent reading of the path's it takes it looks like it should already be safe for concurrent actions.

mattdurham (approved):

pkg/util/logger.go
rfratto: yeah, we need it to protect against a concurrent call to Log and ApplyConfig, where there'd be a data race accessing `l.l`. I'll add a comment to that effect, since I agree it's not completely clear.

PR #493, opened 2021-03-26
Implement /-/reload

### LINKED ISSUE #147: Support Prometheus /-/reload endpoint

rfratto:
#### PR Description 
Adds `/-/reload`. Supporting this change required ensuring that Entrypoint.Run doesn't exit when re-creating the server. 

Closes #147 (finally!) 

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated


rfratto:
I've moved the handling of the server recreation logic to a new package, which helped a bit. I'll do a little more thinking about if it can be simplified tomorrow, and then test it one more time before opening this up for review.

cmd/agent/entrypoint.go
```diff
@@ -67,52 +97,32 @@ func NewEntrypoint(logger *util.Logger, cfg *config.Config) (*Entrypoint, error)
...
 func (srv *Entrypoint) ApplyConfig(cfg config.Config) error {
 	srv.mut.Lock()
 	defer srv.mut.Unlock()
 
-	// The server config uses some unexported fields which can't be compared by
-	// default. Since only exported fields are used by YAML, we'll only compare
-	// those here.
-	var ignoreUnexported = cmpopts.IgnoreUnexported(logging.Format{}, logging.Level{})
+	// Unregister any metrics that the server registered in the last apply.
+	srv.unreg.UnregisterAll()
```
rfratto, line 109: Ah, this is wrong. This should only be happening when recreating the server.

cmd/agent/entrypoint.go
```diff
@@ -67,52 +94,28 @@ func NewEntrypoint(logger *util.Logger, cfg *config.Config) (*Entrypoint, error)
...
 	defer srv.mut.Unlock()
 
-	// The server config uses some unexported fields which can't be compared by
-	// default. Since only exported fields are used by YAML, we'll only compare
-	// those here.
-	var ignoreUnexported = cmpopts.IgnoreUnexported(logging.Format{}, logging.Level{})
-
-	if cmp.Equal(srv.cfg, cfg, ignoreUnexported) {
-		return nil
+	if cfg.Server.Log == nil {
```
mattdurham, line 105: Do these two lines need to be reversed? IE set the src.cfg.Server.Log?

rfratto: I pushed this logic to the caller to make it a little less awkward.

mattdurham (approved):
Tested with tls and had some feedback on documentation. Ran through the code and tried calling the reload config endpoint with a variety of valid and invalid configs without issues. Did parallels calls to reload on both the primary and reload-port around 10k times with no observed memory leak or issues.

docs/api.md
```diff
@@ -203,6 +203,44 @@ Response on success:
...
+`/-/reload` will return with a status code of 200 once all subsystems have been
+updated. Malformed configuration files (invalid YAML, failed validation checks)
+will be immediately rejected with a status code of 400.
+
+If the configuration for the HTTP server is changed, it will be restarted.
+Because of this, it is not recommended to call `/-/reload` against the main HTTP
+server, as restarting it will prevent an HTTP client from reading the response
+of the reload. Instead, use the `--reload-addr` and `--reload-port` flags
+supported by the Agent. That will launch a secondary HTTP server that only
+responds to `/-/reload` and can be used to safely reload the system.
```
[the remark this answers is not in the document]
rfratto: 🤷‍♂️ I'm okay with leaving them both open and just recommend not using the primary server. I can see both arguments here though.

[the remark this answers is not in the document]
rfratto: Going to leave this as-is for now and let's come back later to see if we want to restrict it.

PR #499, opened 2021-03-30
Support generating metrics from spans

### LINKED ISSUE #357: Add support for metrics from traces
There is an incoming PR to the opentelemetry-collector-contrib repo that will allow the generation of metrics from traces:

https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/2012

Once this is merged let's vendor and expose this config to our users.

yeya24:
Signed-off-by: yeya24 <yb532204897@gmail.com>

#### PR Description 

This pr adds the support for `spanmetricsprocessor` to generate metrics from spans.
It will create two metrics, one is a counter called `calls` and another one is a histogram called `latency`.
For its configuration and usage, please check https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.23.0/processor/spanmetricsprocessor/README.md.

`spanmetricsprocessor` requires a metrics exporter to expose the generated metrics. Here I think it might be better to allow users to configure the exporter. Now I only add prometheus exporter support.

I haven't updated the doc because I am not sure about my implementation and the configuration. If the impl looks good, I can add the document later.

Fixes #357

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [x] Tests updated


yeya24:
I already tested this manually with the config below. After traces are pushed, I can get the generated metrics from `localhost:8889/metrics`.

``` yaml
server:
  http_listen_port: 8081
tempo:
  configs:
  - name: tempo
    push_config:
      ...

    spanmetrics:
      latency_histogram_buckets: [ 2ms, 6ms, 10ms, 100ms, 250ms ]
      dimensions:
        - name: http.method
          default: GET
        - name: http.status_code
      metrics_exporter:
        name: prometheus
        config:
          endpoint: "0.0.0.0:8889"
          namespace: promexample

    receivers:
      jaeger:
        protocols:
          grpc:
          thrift_binary:
          thrift_compact:
          thrift_http:
      zipkin:
      otlp:
        protocols:
          http:
          grpc:
      opencensus:
```

pkg/tempo/config.go
```diff
@@ -143,6 +152,22 @@ func (c *RemoteWriteConfig) UnmarshalYAML(unmarshal func(interface{}) error) err
 	return nil
 }
 
+// SpanMetricsConfig controls the configuration of spanmetricsprocessor and the related metrics exporter.
+type SpanMetricsConfig struct {
+	LatencyHistogramBuckets []time.Duration                  `yaml:"latency_histogram_buckets,omitempty"`
+	Dimensions              []spanmetricsprocessor.Dimension `yaml:"dimensions,omitempty"`
+	MetricsExporter         MetricsExporterConfig            `yaml:"metrics_exporter"`
```
mapno, line 159: Should we allow multiple exporters to be configured?

mapno (commented):
Thank you for your contribution! Great work!
The changes are in line with what I had in mind.  I think we can move forward with the approach.  I just left some minor comments.

I also tried the example you provided while adding the prometheus exporter to `scrape_configs` and they get reported, so very cool!

pkg/tempo/dummyreceiver/factory.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/config/configmodels"
+	"go.opentelemetry.io/collector/consumer"
+	"go.opentelemetry.io/collector/receiver/receiverhelper"
+)
+
+const (
+	// TypeStr for dummy receiver.
+	TypeStr = "dummy"
```
mapno, line 14: What do you think of naming it `noop` instead?

mapno (commented):
On second thought, would it make more sense to configure a prometheus exporter by default?  The main use case will be to have the agent self-scrape the exported prometheus metrics and have them sent to the configured `remote_write`.  I think it makes it a simpler feature for the users to enable (i.e. less configuration).

The future goal will be to hook this metrics directly to the agent's remote write, instead of having to scrape them locally.

pkg/tempo/dummyreceiver/factory.go
yeya24: Yeah, I am good with this.

pkg/tempo/config.go
```diff
@@ -276,24 +299,50 @@ func (c *InstanceConfig) otelConfig() (*configmodels.Config, error) {
 		processorNames = append(processorNames, "batch")
 	}
 
-	otelMapStructure["processors"] = processors
+	if c.SpanMetrics != nil {
+		// Configure the metrics exporter.
+		exporters[c.SpanMetrics.MetricsExporter.Name] = c.SpanMetrics.MetricsExporter.Config
```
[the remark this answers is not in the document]
yeya24: > On second thought, would it make more sense to configure a prometheus exporter by default? The main use case will be to have the agent self-scrape the exported prometheus metrics and have them sent to the configured `remote_write`. I think it makes it a simpler feature for the users to enable (i.e. less configuration).
> 
> The future goal will be to hook this metrics directly to the agent's remote write, instead of having to scrape them locally.

pkg/tempo/config.go
```diff
@@ -143,6 +152,22 @@ func (c *RemoteWriteConfig) UnmarshalYAML(unmarshal func(interface{}) error) err
...
+	Dimensions              []spanmetricsprocessor.Dimension `yaml:"dimensions,omitempty"`
+	MetricsExporter         MetricsExporterConfig            `yaml:"metrics_exporter"`
+}
+
+// MetricsExporterConfig controls the configuration of the exporter for generated metrics.
+type MetricsExporterConfig struct {
+	Name string `yaml:"name"`
+
+	// Configuration for metrics exporters.
+	// Now only otlp and prometheus exporter are supported.
```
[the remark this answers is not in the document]
yeya24: Good idea.

yeya24:
> would it make more sense to configure a prometheus exporter by default
If we want to have a prometheus exporter by default and we still need to support the remote write, then what would the configuration file look like? Can you please give me an example?

> The future goal will be to hook this metrics directly to the agent's remote write, instead of having to scrape them locally.
Do you mean we want to implement a new exporter for the existing Prometheus remote write part? Without using the otlp prometehus remote write exporter?

pkg/tempo/config.go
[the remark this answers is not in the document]
yeya24: We need to define the exporter configuration first. If we just want to support Prometheus exporter then having a default namespace is fine.

rfratto:
> The future goal will be to hook this metrics directly to the agent's remote write, instead of having to scrape them locally.

My $0.02: I understand this is the future goal, but I personally think we should *not* have this use the Prometheus Exporter, and would rather keep the current approach of having the Agent scrape itself. The Prometheus Exporter doesn't have a WAL, and I would feel weird having some of the Agent's metrics be more vulnerable to data loss. 

I think moving towards the future goal should be done in a follow up PR (especially because gut feeling tells me it's not going to be as simple as it sounds) and we should avoid temporarily using the Prometheus exporter. 

Just a suggestion though! Ultimately I want to leave the call up to you and/or @joe-elliott to decide what to do for this iteration of the feature.

docs/configuration-reference.md
```diff
@@ -2046,6 +2046,28 @@ receivers:
...
+  [ dimensions: <spanmetricsprocessor.dimensions> ]
+
+  # metrics_exporter config embeds the configuration for opentelemetry prometheus exporter.
+  # https://github.com/open-telemetry/opentelemetry-collector/blob/v0.23.0/exporter/prometheusexporter/README.md
+  metrics_exporter:
+    [ endpoint: <prometehusexporter.endpoint> ]
+    [ const_labels: <prometehusexporter.const_labels> ]
+    [ namespace: <prometehusexporter.namespace> ]
+    [ send_timestamps: <prometehusexporter.send_timestamps> ]
+    [ metric_expiration: <prometehusexporter.metric_expiration> ]
```
mapno, line 2070: This option does not exist in the version of the OTel collector we're using

```suggestion
```

mapno (approved):
LGTM. Left just 3 minor comments. Thanks again for the contribution!

yeya24:
@mapno Hello, I have updated my pr.

CHANGELOG.md
```diff
@@ -21,6 +21,9 @@ cross-compilation issue, but will return in v0.13.0.
   configuration file with defaults applied that the Agent has loaded from disk.
   (@rfratto)
 
+- [FEATURE] Support generating metrics and exposing them via a Prometheus exporter
```
rfratto, line 24:
```suggestion
- [FEATURE] (beta) Support generating metrics and exposing them via a Prometheus exporter
```

rfratto (approved):
This LGTM, but I'm going to leave it up to @mapno to merge since I've mostly been disengaged.

cc @joe-elliott in case you want to give a final review

docs/configuration-reference.md
```diff
@@ -2046,6 +2046,27 @@ receivers:
...
+spanmetrics:
+  # latency_histogram_buckets and dimensions are the same as the configs in spanmetricsprocessor. 
+  # https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.21.0/processor/spanmetricsprocessor/config.go#L38-L47
+  [ latency_histogram_buckets: <spanmetricsprocessor.latency_histogram_buckets> ]
+  [ dimensions: <spanmetricsprocessor.dimensions> ]
+
+  # metrics_exporter config embeds the configuration for opentelemetry prometheus exporter.
+  # https://github.com/open-telemetry/opentelemetry-collector/blob/v0.21.0/exporter/prometheusexporter/README.md
+  metrics_exporter:
+    [ endpoint: <prometehusexporter.endpoint> ]
```
joe-elliott, line 2066: spelling: `prometehus` => `prometheus`

docs/configuration-reference.md
```diff
@@ -2046,6 +2046,27 @@ receivers:
 # If a match is found then relabeling rules are applied.
 scrape_configs:
   - [<scrape_config>]
+  
+# spanmetrics supports aggregating Request, Error and Duration (R.E.D) metrics from span data.
+# https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.21.0/processor/spanmetricsprocessor/README.md.
+# spanmetrics generates two metrics from spans and exposes them to an HTTP endpoint (prometheus exporter).
```
joe-elliott, line 2052: it's not clear to me if this is a remote write configuration or a scrape endpoint configuration. i'm guessing this just opens an http server on a port and exposes the metrics at `/metrics` ?

either way can we get a little more verbiage to clarify?

joe-elliott (commented):
one nit and one request. fine with merging this otherwise.

thank you for the contribution!

docs/configuration-reference.md
yeya24: Exactly. Can you take a another look?

joe-elliott (approved):

PR #503, opened 2021-03-31
Add /-/config endpoint

rfratto:
#### PR Description 
Marks all fields as `omitempty` and adds support for a `/-/config` endpoint that marshals out the current loaded configuration over HTTP. 

#### Notes to the Reviewer
Some of this implementation accidentally snuck into #493, sorry!

There was a bug where an empty list of integrations was not marshaling correctly, but that has been fixed with this PR (and a new test to make sure we don't regress).

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/integrations/register.go
+++ b/pkg/integrations/register.go
@@ -98,6 +98,9 @@ func MarshalYAML(v interface{}) (interface{}, error) {
 	for i, n := 0, inType.NumField(); i < n; i++ {
 		if inType.Field(i).Type == configsType {
 			configs = inVal.Field(i).Interface().(Configs)
+			if configs == nil {
+				configs = Configs{}
+			}
 		}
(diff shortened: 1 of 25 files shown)


pkg/integrations/register.go
```diff
@@ -98,20 +98,23 @@ func MarshalYAML(v interface{}) (interface{}, error) {
 	for i, n := 0, inType.NumField(); i < n; i++ {
 		if inType.Field(i).Type == configsType {
 			configs = inVal.Field(i).Interface().(Configs)
+			if configs == nil {
+				configs = Configs{}
+			}
```
rfratto, line 103: The slice defaults to nil, so we have to do this here to make sure the function returns without erroring.

mattdurham (approved):

PR #509, opened 2021-04-01
Add stability information

rfratto:


CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ cross-compilation issue, but will return in v0.13.0.
 
 # Main (unreleased)
 
+As of this release, functionality will start being tagged as "experimental" or
```
56quarters, line 6: What's the difference between these two? Is "experimental" something that may not get added at all vs "beta" which will definitely be added but might change?

rfratto: no difference, but I think we already mix names in a few places. I'll re-phrase

mattdurham (approved):

PR #513, opened 2021-04-05
Add support for writing to Windows event log while running as a service

### LINKED ISSUE #287: Support running as a Windows Service
Add an optional `-windows-svc` flag that starts the Agent in Windows service mode. The flag is needed to maintain backwards compatbility with users using tools like NSSM to run the Agent as a service; it seems like NSSM may break if Windows service mode is blindly entered. 

Note that if this service is used, logs should be sent to the Windows Event Log. A shim to map go-kit logs to the Windows Event Log will need to be included.

The [`golang.org/x/sys/windows/svc`](https://pkg.go.dev/golang.org/x/sys/windows/svc) package can be used to run as a service. The Go team has an [example application that uses this package](https://github.com/golang/sys/blob/master/windows/svc/example/main.go).

mattdurham:
#### PR Description 

While running as a Windows Service write to the windows event log

Closes #287

#### Notes to the Reviewer

There is some oddness with regards to having three io.Writers to write to the event log, I want the comments to be clear so please look at that spot with a heavy eye.

#### PR Checklist

- [X] CHANGELOG updated 
- [X] Documentation added
- [X] Tests updated


pkg/util/logger_windows.go
```diff
@@ -0,0 +1,134 @@
...
+func (w *winLoggerFmt) Log(keyvals ...interface{}) error {
+	lvl, err := getLevel(keyvals...)
+	// If we don't know what happened move on
+	if err != nil {
+		w.infoLogger.Log(keyvals...)
+		return err
+	}
+	// If the messages level matches one of these we try to write to the logger
+	// the loggers are configured to reject the message if it isn't allowed.
+	if lvl == level.DebugValue() {
```
mattdurham, line 96: This is possibly the oddest bit of code, essentially if the infoLogger can't handle the level it will return the error we registered down below. We use that to filter out anything that shouldn't be at the correct level.

pkg/util/logger_windows.go
```diff
@@ -0,0 +1,134 @@
...
+	// Cleanup the handle when exits scope, this could be handled via explicit close but would need to change
+	// more upstream, and its not a huge issue if it hangs around slightly longer than it should
+	runtime.SetFinalizer(il, func(l *el.Log) {
+		l.Close()
+	})
+
+	//  These are setup to be writers for each Windows log level
+	//  Setup this way so we can utilize all the benefits of logformatter
+	infoWriter := &winLogWriter{writer: func(p []byte) {
+		il.Info(1, string(p))
```
rfratto, line 53: The error of these should be returned just in case something goes wrong.

pkg/util/logger_windows.go
```diff
@@ -0,0 +1,134 @@
...
+	notAllowedError := errors.New("not_allowed")
+
+	// Setup the log in windows events
+	err := el.InstallAsEventCreate(ServiceName, el.Error|el.Info|el.Warning)
+
+	// We expect an error of 'already exists' for subsequent runs,
+	if err != nil && !strings.Contains(err.Error(), "already exists") {
+		return nil, err
+	}
+	il, err := el.Open(ServiceName)
```
rfratto, line 39: Is it valid to have two of these open at the same time? Can both of them still write?

With using the finalizer, it means there will be a window where two event logs for the same service name are used.

pkg/util/logger_windows.go
```diff
@@ -0,0 +1,134 @@
...
+	runtime.SetFinalizer(il, func(l *el.Log) {
+		l.Close()
+	})
+
+	//  These are setup to be writers for each Windows log level
+	//  Setup this way so we can utilize all the benefits of logformatter
+	infoWriter := &winLogWriter{writer: func(p []byte) {
+		il.Info(1, string(p))
+	}}
+	infoLogger := log.NewLogfmtLogger(infoWriter)
```
rfratto, line 55: `*server.Config` can also request JSON here, not just logfmt

pkg/util/logger_windows.go
```diff
@@ -0,0 +1,134 @@
...
+		il.Warning(1, string(p))
+	}}
+	warningLogger := log.NewLogfmtLogger(warningWriter)
+	warningLogger = level.NewFilter(warningLogger, cfg.LogLevel.Gokit, level.ErrNotAllowed(notAllowedError))
+
+	errorWriter := &winLogWriter{writer: func(p []byte) {
+		il.Error(1, string(p))
+	}}
+	errorLogger := log.NewLogfmtLogger(errorWriter)
+	errorLogger = level.NewFilter(errorLogger, cfg.LogLevel.Gokit, level.ErrNotAllowed(notAllowedError))
```
rfratto, line 68: I don't think each of these needs their own filter. `winLoggerFmt` should be wrapped with `level.NewFilter`, so `Log` only gets called if the individual log message is supposed to be logged to one of the underlying writers.

pkg/util/logger_windows.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+// Looks through the key value pairs in the log for level and extract the value
+func getLevel(keyvals ...interface{}) (level.Value, error) {
+	if len(keyvals)%2 == 1 {
+		keyvals = append(keyvals, nil)
+	}
+	for i := 0; i < len(keyvals); i += 2 {
+		k, v := keyvals[i], keyvals[i+1]
+		if k == "level" {
```
rfratto, line 116: The `github.com/go-kit/log/level` package doesn't seem to care about the key, just the value implementing `level.Value`, we should do the same. (It also means the only error we'd have to return is for "no level found", which should default to info-level, as you have now.)

pkg/util/logger_windows.go
```diff
@@ -0,0 +1,134 @@
...
+			}
+			return nil, fmt.Errorf("unknown level")
+
+		}
+	}
+	return nil, fmt.Errorf("no level found")
+}
+
+type winLogWriter struct {
+	writer func(p []byte)
```
rfratto, line 128:
```suggestion
	writer func(p []byte) error
```

pkg/util/logger_windows.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+func (i *winLogWriter) Write(p []byte) (n int, err error) {
+	i.writer(p)
+	return len(p), nil
```
rfratto, line 133:
```suggestion
	return len(p), i.writer(p)
```

pkg/util/logger_windows.go
```diff
@@ -0,0 +1,134 @@
...
+	// the loggers are configured to reject the message if it isn't allowed.
+	if lvl == level.DebugValue() {
+		return w.infoLogger.Log(keyvals...)
+	} else if lvl == level.InfoValue() {
+		return w.infoLogger.Log(keyvals...)
+	} else if lvl == level.ErrorValue() {
+		return w.errorLogger.Log(keyvals...)
+	} else if lvl == level.WarnValue() {
+		return w.warningLogger.Log(keyvals...)
+	}
```
rfratto, line 104: This should probably use a switch statement instead to be a little cleaner, and then have the default case always log to info.Logger:

```go
switch lvl {
case level.DebugValue():
  return w.infoLogger.Log(keyvals...)
// ...
default:
  return w.infoLogger.Log(keyvals...)
} 
```

pkg/util/logger_windows.go
```diff
@@ -0,0 +1,134 @@
...
+	warningLogger log.Logger
+}
+
+func (w *winLoggerFmt) Log(keyvals ...interface{}) error {
+	lvl, err := getLevel(keyvals...)
+	// If we don't know what happened move on
+	if err != nil {
+		w.infoLogger.Log(keyvals...)
+		return err
+	}
```
rfratto, line 93: We can probably have `getLevel` return nil if it didn't find a level and rely on the default case of the switch statement to clean this function up.

pkg/util/logger_windows.go
```diff
@@ -0,0 +1,134 @@
...
+// Default name for the Grafana Agent under Windows
+const ServiceName = "Grafana Agent"
+
+//NewWindowsEventLogger creates a new logger that writes to the event log
+func NewWindowsEventLogger(cfg *server.Config) *Logger {
+	l := Logger{makeLogger: makeWindowsEventLogger}
+	if err := l.ApplyConfig(cfg); err != nil {
+		panic(err)
+	}
+	return &l
```
rfratto, line 25: Should this call `newLogger(cfg, makeWindowsEventLogger)` instead of re-implementing it?

rfratto (commented):
Nice! Mostly looks good, few nits about how to order the wrapping and style choices.

pkg/util/logger_windows.go
mattdurham: Yes, and yes, I had double-checked that you could write to two but also tested it in parallel with no issue.

docs/windows.md
```diff
@@ -32,4 +32,4 @@ If the Grafana Agent is installed using the installer, it can be uninstalled via
 
 ## Logs
 
-Logs from the Grafana Agent service are currently not stored anywhere. In the future version, logs will be written to Event Viewer.
+When Grafana Agent is running as a Windows Service the Grafana Agent will write logs to Windows Event Logs. When running as executable, Grafana Agent will write to standard out.
```
rfratto, line 35: Can you add the service name used for logging here?

pkg/util/logger_windows.go
```diff
@@ -0,0 +1,117 @@
...
+	return nil
+}
+
+func newWinLogWrapper(format logging.Format, write func(p []byte) error) log.Logger {
+	infoWriter := &winLogWriter{writer: write}
+	infoLogger := log.NewLogfmtLogger(infoWriter)
+	if format.String() == "json" {
+		infoLogger = log.NewJSONLogger(infoWriter)
+	}
+	return infoLogger
```
rfratto, line 81: Hmm, I don't think we need one of those for each underlying writer, we can do the same shared wrapper like we do with the filter.

pkg/util/logger_windows.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+type winLogger struct {
+	internalLog *el.Log
```
rfratto, line 85: This field appears to be unused, do we need it?

pkg/util/logger_windows_test.go
```diff
@@ -0,0 +1,28 @@
...
+	wl := NewWindowsEventLogger(&server.Config{
+		LogLevel: logging.Level{
+			Logrus: logrus.InfoLevel,
+			Gokit:  level.AllowAll(),
+		},
+	})
+	level.Debug(wl).Log("msg", "debug")
+	level.Info(wl).Log("msg", "info")
+	level.Warn(wl).Log("msg", "warn")
+	level.Error(wl).Log("msg", "error")
```
rfratto, line 26: This doesn't seem to be testing a lot here. We can probably do two different kinds of tests:

1. Make sure there's an error returned by `Log` when the level is too high 
2. Make sure there's no error while logging messages that should show up in the event log

pkg/util/logger_windows.go
```diff
@@ -0,0 +1,117 @@
...
+
+	"github.com/go-kit/kit/log"
+	"github.com/weaveworks/common/server"
+	el "golang.org/x/sys/windows/svc/eventlog"
+)
+
+// Default name for the Grafana Agent under Windows
+const ServiceName = "Grafana Agent"
+
+//NewWindowsEventLogger creates a new logger that writes to the event log
```
rfratto, line 19:
```suggestion
// NewWindowsEventLogger creates a new logger that writes to the event log
```

pkg/util/logger_windows.go
```diff
@@ -0,0 +1,117 @@
...
+func NewWindowsEventLogger(cfg *server.Config) *Logger {
+	return newLogger(cfg, makeWindowsEventLogger)
+}
+
+func makeWindowsEventLogger(cfg *server.Config) (log.Logger, error) {
+	// Setup the log in windows events
+	err := el.InstallAsEventCreate(ServiceName, el.Error|el.Info|el.Warning)
+
+	// We expect an error of 'already exists' for subsequent runs,
```
rfratto, line 28:
```suggestion
	// We expect an error of 'already exists' for subsequent runs.
```

pkg/util/logger_windows.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil && !strings.Contains(err.Error(), "already exists") {
+		return nil, err
+	}
+	il, err := el.Open(ServiceName)
+	if err != nil {
+		return nil, err
+	}
+
+	// Cleanup the handle when exits scope, this could be handled via explicit close but would need to change
+	// more upstream, and its not a huge issue if it hangs around slightly longer than it should
```
rfratto, line 38: nit: "not a huge issue" makes it sound like it might be a tiny issue :) Can you re-write this to clarify that it's valid to have more than one event logger open concurrently?

```suggestion
	// Ensure the logger gets closed when the GC runs. It's valid to have more than one win logger open concurrently. 
```

pkg/util/logger_windows.go
```diff
@@ -0,0 +1,117 @@
...
+	runtime.SetFinalizer(il, func(l *el.Log) {
+		l.Close()
+	})
+
+	//  These are setup to be writers for each Windows log level
+	//  Setup this way so we can utilize all the benefits of logformatter
```
rfratto, line 44: nit: double spaces after `//`:
```suggestion
	// These are setup to be writers for each Windows log level
	// Setup this way so we can utilize all the benefits of logformatter
```

rfratto (approved):
LGTM!

pkg/util/logger_windows_test.go
mattdurham: Agreed on the tests not having value.

pkg/util/logger_windows.go
rfratto: Ahhh, right. Never mind!

PR #514, opened 2021-04-05
Create the skeleton for an Agent Operator

rfratto:
#### PR Description 
Lays out the base for an Agent Operator. 

#### Notes to the Reviewer
There's probably going to be a lot of changes here, so I'm opening this up against an `operator` branch until we have a MVP. After an MVP is working, we can merge operator into main and continue development against main. 

Very little is working here, but if you follow the instructions in `DEVELOPMENT.md` you can see it react to applying the custom resources. One thing that needs to be done later is allow makefiles to generate the resources instead of doing it all by hand.


mattdurham (approved):
Seems reasonable for a skeleton looked over a chunk of it the other day.

PR #536, opened 2021-04-13
Add headers field to Tempo's remote_write  

### LINKED ISSUE #535: Support specifying additional headers in the remote_write in Tempo config
`remote_write`s  for the Prometheus configuration support specification of additional headers that are passed to the remote writer. Unfortunately this doesn't apply for the `remote_write`s in Tempo's config.  

At work we have a requirement where this would come in very handy. I looked into the code and it seems to be pretty straightforward to add and I could provide a PR.

alexbiehl:
#### PR Description 

This PR adds an optional `headers` field to `remote_write`s in Tempo. Headers specified through that field are passed on to remote writers. 

Closes #535 

#### PR Checklist

I will update documentation and CHANGELOG in upcoming commits.

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/configuration-reference.md
+++ b/docs/configuration-reference.md
@@ -2017,6 +2017,12 @@ remote_write:
   # host:port to send traces to
   - endpoint: <string>
 
+    # Custom HTTP headers to be sent along with each remote write request.
+    # Be aware that 'authorization' header will be overwritten in presence
+    # of basic_auth.
+    headers:
+      [ <string>: <string> ... ]
+
     # Controls whether compression is enabled.
     [ compression: <string> | default = "gzip" | supported = "none", "gzip"]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ cross-compilation issue, but will return in v0.13.0.
 
 # Main (unreleased)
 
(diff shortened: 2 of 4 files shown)


rfratto (approved):
This is great, thanks! Tagging @mapno for a code owner review. (Though we should hold off on merging until you have the changelogs/docs updated and we briefly review those as well)

alexbiehl:
Thanks for confirming this is the right approach @rfratto. I have added a CHANGELOG entry and adjusted the documentation.

docs/configuration-reference.md
```diff
@@ -2017,6 +2017,9 @@ remote_write:
   # host:port to send traces to
   - endpoint: <string>
 
+    # Custom HTTP headers to be sent along with each remote write request.
```
rfratto, line 2020: Can you note that "authorization" will always be overridden here?

alexbiehl: Hm, is it really always or just in presence of basic_auth?

rfratto: Ah, that's right, my bad. Let's note that configuring basic auth will override the value for authorization if set in custom headers, then.

mapno (approved):
LGTM, thanks for the contribution!

PR #539, opened 2021-04-14
Tempo: Use correct logger in Prom SD

joe-elliott:
#### PR Description 
Noticed this while working on other changes. I believe we were using the wrong logger in our SD processor.


rfratto (approved):
LGTM. Ideally we wouldn't use a logger based off of util.Logger (line 39), but it should be mostly equivalent and I know how hard it is to propagate custom objects through the OTel constructors. Merge at will!

PR #551, opened 2021-04-20
[Tempo] Automatic Logging

joe-elliott:
#### PR Description 
Adds an automatic logging processor, documentations, tests and updates the docker-compose example to use it.

Example lines:
```
span=/cart dur=410000000ns svc=frontend status=STATUS_CODE_UNSET tid=00000000000000002906906e1a662c34
span=/cart dur=461000000ns svc=frontend status=STATUS_CODE_UNSET tid=000000000000000041b1b78d4adc4bd7
span=/product dur=388000000ns svc=frontend status=STATUS_CODE_UNSET tid=0000000000000000b58662a921244284
span=/cart dur=421000000ns svc=frontend status=STATUS_CODE_UNSET tid=00000000000000004e7d51d0834bccf0
```

#### Notes to the Reviewer
Also updated Grafana, Loki and added Tempo to the default docker-compose example.

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated


docs/configuration-reference.md
```diff
@@ -2045,6 +2045,21 @@ remote_write:
     [ sending_queue: <otlpexporter.sending_queue> ]
     [ retry_on_failure: <otlpexporter.retry_on_failure> ]
 
+    # automatically log lines to Loki for discovery/metrics
+    automatic_logging:
+      [ spans: <boolean> | default = false ]               # log one line per span. Warning! possibly very high volume
+      [ roots: <boolean> | default = false ]               # log one line for every root span of a trace.
+      [ processes: <boolean> | default = false ]           # log one line for every process
+      [ span_attributes: <string array> | default = []]    # additional span attributes to log
+      [ process_attributes: <string array> | default = []] # additional process attributes to log
```
rfratto, line 2054: FYI: you can omit defaults if they are the zero value here

example/docker-compose/docker-compose.yaml
```diff
@@ -44,15 +44,19 @@ services:
       - agent
 
   # tracing backend
-  otel-collector:
-    image: otel/opentelemetry-collector:0.9.0
-    volumes:
-      - ./otel-collector:/etc/otel-collector
-    command:
-      - --config=/etc/otel-collector/config.yaml
+  tempo:
```
rfratto, line 47: Ooh, exciting to have Tempo in here :)

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ cross-compilation issue, but will return in v0.13.0.
 - [ENHANCEMENT] Add  `headers` field in `remote_write` config for Tempo. `headers`
   specifies HTTP headers to forward to the remote endpoint. (@alexbiehl)
 
+- [FEATURE] Added Automatic Logging feature for Tempo (@joe-elliott)
```
rfratto, line 25: I try to keep these ordered by FEATURE > ENHANCEMENT > BUGFIX > CHANGE, can you move this to the top?

pkg/tempo/automaticloggingprocessor/automaticloggingprocessor.go
```diff
@@ -0,0 +1,242 @@
...
+func (p *automaticLoggingProcessor) Start(ctx context.Context, _ component.Host) error {
+	loki := ctx.Value(contextkeys.Loki).(*loki.Loki)
+	if loki == nil {
+		return fmt.Errorf("key %s does not contain a Loki instance", contextkeys.Loki)
+	}
+	lokiInstance := loki.Instance(p.cfg.LokiName)
+	if lokiInstance == nil {
+		return fmt.Errorf("loki instance %s not found", p.cfg.LokiName)
+	}
+	p.lokiChan = lokiInstance.Promtail().Client().Chan()
```
rfratto, line 129: This isn't going to work if the promtail restarts 😞 We'll need this to use a channel on the instance (which should be immutable) and have the Loki instance forward from the active promtail over the instance's channel.

pkg/tempo/automaticloggingprocessor/automaticloggingprocessor.go
```diff
@@ -0,0 +1,242 @@
...
+	return component.ProcessorCapabilities{}
+}
+
+// Start is invoked during service startup.
+func (p *automaticLoggingProcessor) Start(ctx context.Context, _ component.Host) error {
+	loki := ctx.Value(contextkeys.Loki).(*loki.Loki)
+	if loki == nil {
+		return fmt.Errorf("key %s does not contain a Loki instance", contextkeys.Loki)
+	}
+	lokiInstance := loki.Instance(p.cfg.LokiName)
```
rfratto, line 125: What happens if the Agent config gets changed and the LokiName used here goes away? We'll need some kind of validation somewhere at the pkg/config level to make sure you can't break this. 

It *may* be easier in the short-term to launch our own promtail from within here instead of referencing one from the logging block, but it's up to you.

rfratto (commented):
Fancy! I have some general concerns about how this interacts with the reloading system.

pkg/tempo/automaticloggingprocessor/factory.go
```diff
@@ -0,0 +1,70 @@
...
+	"context"
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/config/configmodels"
+	"go.opentelemetry.io/collector/consumer"
+	"go.opentelemetry.io/collector/processor/processorhelper"
+)
+
+// TypeStr is the unique identifier for the Automatic Logging processor.
+const TypeStr = "automatic_logging_processor"
```
mapno, line 13: nit: `automatic_logging` would be more in line with the resto of the processors which don't include the processor suffix.

pkg/tempo/automaticloggingprocessor/automaticloggingprocessor.go
```diff
@@ -0,0 +1,242 @@
...
+	if !cfg.Roots && !cfg.Processes && !cfg.Spans {
+		return nil, errors.New("automaticLoggingProcessor requires one of roots, processes, or spans to be enabled")
+	}
+
+	cfg.Overrides.LokiTag = override(cfg.Overrides.LokiTag, defaultLokiTag)
+	cfg.Overrides.ServiceKey = override(cfg.Overrides.ServiceKey, defaultServiceKey)
+	cfg.Overrides.SpanNameKey = override(cfg.Overrides.SpanNameKey, defaultSpanNameKey)
+	cfg.Overrides.StatusKey = override(cfg.Overrides.StatusKey, defaultStatusKey)
+	cfg.Overrides.DurationKey = override(cfg.Overrides.DurationKey, defaultDurationKey)
+	cfg.Overrides.TraceIDKey = override(cfg.Overrides.TraceIDKey, defaultTraceIDKey)
```
mapno, line 65: I think the argument to `override` has to be the non-override values

pkg/tempo/automaticloggingprocessor/automaticloggingprocessor.go
joe-elliott: Yeah, I did some mild testing and it seemed to work, but my expectation is that in the general case it would not. I was considering:

- Shutdown Tempo subsystem
- Reload Loki 
- Reload Tempo

I think this would work.

pkg/tempo/automaticloggingprocessor/automaticloggingprocessor.go
joe-elliott: In that case ApplyConfig would fail: 
https://github.com/grafana/agent/blob/main/pkg/tempo/tempo.go#L48

which means that this would fail as well?
https://github.com/grafana/agent/blob/main/cmd/agent/entrypoint.go#L101

I think that's ok?

pkg/tempo/automaticloggingprocessor/automaticloggingprocessor.go
joe-elliott: I think this is fine? It needs tests regardless.

pkg/tempo/automaticloggingprocessor/automaticloggingprocessor.go
rfratto: So, the main problem here is if a user goes through these steps:

1. User starts the Agent using the automatic logger 
2. User modifies Agent config, removes the Loki entry, but the Tempo config is left unchanged 
3. User reloads the Agent. 
4. Automatic logging processor breaks, referencing a Promtail that has shut down

This is because we [cache the input to ApplyConfig](https://github.com/grafana/agent/blob/e874ac465849e33ddb77bd121ce5357bdf894b99/pkg/tempo/instance.go#L54) so we're not always restarting systems that didn't change.

pkg/tempo/automaticloggingprocessor/automaticloggingprocessor.go
rfratto: It'd work, but I'd prefer to try to get this to work with the existing ApplyConfig mechanism and not introducing shutdowns if we can. What were your concerns with forwarding the channel?

pkg/tempo/automaticloggingprocessor/automaticloggingprocessor.go
rfratto: (But if we have config validation in the [global ApplyDefaults](https://github.com/grafana/agent/blob/e874ac465849e33ddb77bd121ce5357bdf894b99/pkg/config/config.go#L51), we can return an error before any part of the system starts to reload, which would leave it in a working state)

pkg/tempo/automaticloggingprocessor/automaticloggingprocessor.go
mapno: Yes, sorry. I got confused

pkg/tempo/config.go
```diff
@@ -61,11 +62,11 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if err := unmarshal((*plain)(c)); err != nil {
 		return err
 	}
-	return c.Validate()
+	return nil
```
rfratto, line 65: Can we just have this changed to `return unmarshal((*plain)(c))` and remove the if above?

pkg/tempo/config.go
```diff
@@ -61,11 +62,11 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
 }
 
 // Validate ensures that the Config is valid.
-func (c *Config) Validate() error {
+func (c *Config) Validate(lokiConfig loki.Config) error {
```
rfratto, line 69: Nit: can lokConfig be a pointer?

pkg/tempo/automaticloggingprocessor/automaticloggingprocessor.go
joe-elliott: @rfratto have a first pass attempt at doing this with the `ApplyConfig`. Let me know what you think.

pkg/tempo/automaticloggingprocessor/automaticloggingprocessor.go
```diff
@@ -0,0 +1,248 @@
...
+	atts = append(atts, p.cfg.Overrides.DurationKey)
+	atts = append(atts, spanDuration(span))
+
+	atts = append(atts, p.cfg.Overrides.ServiceKey)
+	atts = append(atts, svc)
+
+	atts = append(atts, p.cfg.Overrides.StatusKey)
+	atts = append(atts, span.Status().Code())
+
+	span.Status().Code()
```
mapno, line 180:
```suggestion
```

pkg/tempo/automaticloggingprocessor/automaticloggingprocessor.go
```diff
@@ -0,0 +1,248 @@
...
+			atts = append(atts, name)
+			atts = append(atts, attributeValue(att))
+		}
+	}
+
+	return atts
+}
+
+func (p *automaticLoggingProcessor) spanKeyVals(span pdata.Span, svc string) []interface{} {
+	atts := make([]interface{}, 0, 8) // 8 for name, duration and service name
```
mapno, line 166:
```suggestion
	atts := make([]interface{}, 0, 8) // 8 for name, duration, service name and status
```

mapno (approved):
Nice! LGTM 👍

docs/configuration-reference.md
```diff
@@ -2045,6 +2045,22 @@ remote_write:
...
     [ retry_on_failure: <otlpexporter.retry_on_failure> ]
 
+    # automatically log lines to Loki for discovery/metrics
+    automatic_logging:
+      [ spans: <boolean> ]                      # log one line per span. Warning! possibly very high volume
+      [ roots: <boolean> ]                      # log one line for every root span of a trace.
+      [ processes: <boolean> ]                  # log one line for every process
+      [ span_attributes: <string array> ]       # additional span attributes to log
+      [ process_attributes: <string array> ]    # additional process attributes to log
+      [ timeout: <duration> | default = 100ms ] # timeout on sending logs to Loki
```
rfratto, line 2055: Nit: can documentation be placed on their own line above the field?

pkg/tempo/automaticloggingprocessor/automaticloggingprocessor.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+
+	keyvals = append(keyvals, []interface{}{p.cfg.Overrides.TraceIDKey, traceID}...)
+	line, err := logfmt.MarshalKeyvals(keyvals...)
+	if err != nil {
+		level.Warn(p.logger).Log("msg", "unable to marshal keyvals", "err", err)
+		return
+	}
+
+	sent := p.lokiInstance.SendEntry(api.Entry{
```
rfratto, line 205: In the future we'll have to be careful to make sure the Loki instances don't get replaced so this doesn't end up with a stale reference. I think we're fine for now though.

We could be defensive and perform the lookup each time, but I'm not sure it's worth it just yet.

docs/configuration-reference.md
```diff
@@ -2045,6 +2045,22 @@ remote_write:
     [ sending_queue: <otlpexporter.sending_queue> ]
     [ retry_on_failure: <otlpexporter.retry_on_failure> ]
 
+    # automatically log lines to Loki for discovery/metrics
```
rfratto, line 2048: Can you add a little context here about why someone would want to enable this?

docs/configuration-reference.md
```diff
@@ -2045,6 +2045,22 @@ remote_write:
...
+    automatic_logging:
```
rfratto, line 2049: Just curious: what enables this? Is it defining `automatic_logging` in the config or is it providing a value for `loki_name`? Can we document which?

docs/configuration-reference.md
```diff
@@ -2045,6 +2045,22 @@ remote_write:
...
+      [ spans: <boolean> ]                      # log one line per span. Warning! possibly very high volume
```
rfratto, line 2050: I can't tell if my comments are blocking everything but `loki_name` seems missing here :)

rfratto (approved):
Nice! Mostly nits about the docs.

docs/configuration-reference.md
joe-elliott: Adding `automatic_logging` will instantiate the processor, but it will fail with an error message if it can't find the provided `loki_name`.

docs/configuration-reference.md
```diff
@@ -2045,6 +2045,32 @@ remote_write:
...
+      # log one line for every root span of a trace.
+      [ roots: <boolean> ]
+      # log one line for every process
+      [ processes: <boolean> ]
+      # additional span attributes to log
+      [ span_attributes: <string array> ]
+      # additional process attributes to log
+      [ process_attributes: <string array> ]
+      # timeout on sending logs to Loki
+      [ timeout: <duration> | default = 100ms ]
```
rfratto, line 2065:
```suggestion
      [ timeout: <duration> | default = 1ms ]
```

rfratto: (noticed the default changed)

joe-elliott: good eyeballs

PR #558, opened 2021-04-23
Add security warning for scraping service API

rfratto:


mattdurham (approved):

56quarters (commented):
Amazing, I like this approach a lot more than just adding the warning in the docs. I'd love to see tests for this but this approach is :100:

56quarters (approved):
LGTM

pkg/prom/cluster/validation.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	for _, check := range checks {
+		if check.check() {
+			return fmt.Errorf("%s must be empty unless dangerous_allow_reading_files is set", check.name)
+		}
+	}
+	return nil
+}
+
+func validateDiscoveryNoFiles(disc discovery.Config) error {
```
simonswine, line 69: I would consider using reflection, to avoid breakage when new discovery mechanisms are added

mattdurham (approved):

PR #565, opened 2021-04-27
Cloud install script: Prompt for overwriting config

### LINKED ISSUE #553: Script installer overwrites /etc/grafana-agent.yaml
Hello,

I attempted to upgrade the already installed Grafana Agent by using the installer script. dpkg realizes that there is the new version provided, so it asks whether to overwrite the already existing config. But, the installer itself overwrites the config nevertheless:

```
--- Using package system deb. Downloading and installing package for amd64
(Reading database ... 227578 files and directories currently installed.)
Preparing to unpack /tmp/grafana-agent.deb ...
Unpacking grafana-agent (0.13.1-1) over (0.11.0-1) ...
Setting up grafana-agent (0.13.1-1) ...

Configuration file '/etc/grafana-agent.yaml'
 ==> Modified (by you or by a script) since installation.
 ==> Package distributor has shipped an updated version.
   What would you like to do about it ?  Your options are:
    Y or I  : install the package maintainer's version
    N or O  : keep your currently-installed version
      D     : show the differences between the versions
      Z     : start a shell to examine the situation
 The default action is to keep your current version.
*** grafana-agent.yaml (Y/I/N/O/D/Z) [default=N] ? 
--- Retrieving config and placing in /etc/grafana-agent.yaml
integrations:
  node_exporter:
    enabled: true
  prometheus_remote_write:
  - basic_auth:
      password: xxx
      username: yyy
    url: https://prometheus-us-central1.grafana.net/api/prom/push
prometheus:
  configs:
  - name: integrations
    remote_write:
    - basic_auth:
        password: xxx
        username: yyy
      url: https://prometheus-us-central1.grafana.net/api/prom/push
  global:
    scrape_interval: 15s
  wal_directory: /tmp/grafana-agent-wal
server:
  http_listen_port: 12345

--- Enabling and starting grafana-agent.service


Grafana Cloud Agent is now running! To check the status of your Agent, run:
   sudo systemctl status grafana-agent.service
```

These are the problematic lines in the installer:

```bash
  log '--- Retrieving config and placing in /etc/grafana-agent.yaml'
  retrieve_config | sudo tee /etc/grafana-agent.yaml
```

This should default to either keeping the existing file or prompting the user if he wants to overwrite it, like dpkg does.

rfratto:
#### PR Description 
Changes the installer to prompt for overwriting the existing config file, and making a backup of the existing one before doing an overwrite. 

Closes #553.

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/production/grafanacloud-install.sh
+++ b/production/grafanacloud-install.sh
@@ -73,7 +73,7 @@ main() {
 
   if [ -f /etc/grafana-agent.yaml ]; then
     while true; do
-      read -p "Do you wish to overwrite your configuration with the latest template? A backup will be made. y/N (default N) " yn
+      read -r -p "Do you wish to overwrite your configuration with the latest template? A backup will be made. y/N (default N) " yn
       case $yn in
         [Yy]*)
           log '--- Creating numbered backup of /etc/grafana-agent.yaml in same directory'
--- a/production/grafanacloud-install.sh
+++ b/production/grafanacloud-install.sh
@@ -71,8 +71,30 @@ main() {
       ;;
   esac
 
-  log '--- Retrieving config and placing in /etc/grafana-agent.yaml'
-  retrieve_config | sudo tee /etc/grafana-agent.yaml
+  if [ -f /etc/grafana-agent.yaml ]; then
+    while true; do
+      read -p "Do you wish to overwrite your configuration with the latest template? A backup will be made. y/N (default N) " yn
+      case $yn in
+        [Yy]*)
+          log '--- Creating numbered backup of /etc/grafana-agent.yaml in same directory'
+          sudo cp -p --force --backup=numbered /etc/grafana-agent.yaml /etc/grafana-agent.yaml
+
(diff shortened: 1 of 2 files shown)


production/grafanacloud-install.sh
```diff
@@ -71,8 +71,30 @@ main() {
       ;;
   esac
 
-  log '--- Retrieving config and placing in /etc/grafana-agent.yaml'
-  retrieve_config | sudo tee /etc/grafana-agent.yaml
+  if [ -f /etc/grafana-agent.yaml ]; then
```
56quarters, line 74: This looks like the first use of needing user input in this install script. Maybe another option would be to abort if the file already exists and require a `--overwrite-config` option (or something similar)? That would prevent automated uses of this script from hanging waiting for user input.

rfratto: This is a great point. This script is a bit of a special case: we don't expect it to be involved in automation and only plan for it being used for a manual walk through, but I still think there's value in supporting flags here.

PR #569, opened 2021-04-28
Remote write for span metrics

mapno:
#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/tempo/config.go
+++ b/pkg/tempo/config.go
@@ -22,6 +22,7 @@ import (
 	"go.opentelemetry.io/collector/config"
 	"go.opentelemetry.io/collector/config/configmodels"
 	"go.opentelemetry.io/collector/exporter/otlpexporter"
+	"go.opentelemetry.io/collector/exporter/prometheusexporter"
 	"go.opentelemetry.io/collector/processor/attributesprocessor"
 	"go.opentelemetry.io/collector/processor/batchprocessor"
 	"go.opentelemetry.io/collector/receiver/jaegerreceiver"
@@ -207,6 +208,8 @@ type SpanMetricsConfig struct {
 	ConstLabels map[string]interface{} `yaml:"const_labels,omitempty"`
 	// PromInstance is the Agent's prometheus instance that will be used to push metrics
 	PromInstance string `yaml:"prom_instance"`
+	// HandlerEndpoint is the address where a prometheus exporter will be exposed
+	HandlerEndpoint string `yaml:"handler_endpoint"`
 }
 
 // tailSamplingConfig is the configuration for tail-based sampling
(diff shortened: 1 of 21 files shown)


joe-elliott (commented):
There is an existing remote write exporter:

https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter/prometheusremotewriteexporter

Should we use this?

And instead of all of that, is it possible to reuse the remote write that is already present in the agent? @rfratto may have thoughts here.

If we did it would give us wal support as well.

mapno:
Changed the implementation to use a prometheus instance from the Agent to remote write. There are still things missing, but it's a closer look into the approach. The docker-compose example should correctly report span metrics.

pkg/tempo/config.go
```diff
@@ -202,21 +202,17 @@ func (c *RemoteWriteConfig) UnmarshalYAML(unmarshal func(interface{}) error) err
...
 	Dimensions              []spanmetricsprocessor.Dimension `yaml:"dimensions,omitempty"`
-
-	// MetricsExporter is a Prometheus metrics exporter
-	MetricsExporter metricsExporterConfig `yaml:"metrics_exporter,omitempty"`
+	// Namespace if set, exports metrics under the provided value.
+	Namespace string `yaml:"namespace,omitempty"`
+	// ConstLabels are values that are applied for every exported metric.
+	ConstLabels map[string]interface{} `yaml:"const_labels,omitempty"`
+	// Exporter is the configuration to export the metrics
+	Exporter metricsExporterConfig `yaml:"exporter"`
```
joe-elliott, line 210: unless we think we will add more configuration in `metricsExporterConfig` I don't think we should add this extra config level

joe-elliott (commented):
I didn't go through all the conversion code, but I'm sure it's flawless. This approach makes sense to me.

pkg/tempo/remotewriteexporter/exporter.go
```diff
@@ -0,0 +1,231 @@
...
+				case pdata.MetricDataTypeDoubleSum, pdata.MetricDataTypeIntSum, pdata.MetricDataTypeDoubleGauge, pdata.MetricDataTypeIntGauge:
+					if err := e.processScalarMetric(app, m); err != nil {
+						return fmt.Errorf("failed to process metric %s", err)
+					}
+				case pdata.MetricDataTypeDoubleHistogram, pdata.MetricDataTypeIntHistogram:
+					if err := e.processHistogramMetrics(app, m); err != nil {
+						return fmt.Errorf("failed to process metric %s", err)
+					}
+				case pdata.MetricDataTypeDoubleSummary:
+					return fmt.Errorf("%s processing unimplemented", m.DataType())
```
joe-elliott, line 89: i'm fine with leaving this off for now

pkg/tempo/remotewriteexporter/exporter.go
```diff
@@ -0,0 +1,231 @@
...
+					return fmt.Errorf("unsupported m data type %s", m.DataType())
+				}
+			}
+		}
+	}
+
+	return app.Commit()
+}
+
+func (e *remoteWriteExporter) processHistogramMetrics(app storage.Appender, m pdata.Metric) error {
```
joe-elliott, line 100: this is obviously quite complex. i say add some testing and ship it. we can confirm internally that these metrics are in sync with others.

pkg/tempo/config.go
mapno: Right, the exporter will be configured via the prometheus instance. Removed in 521f7c346f2370b3d71de07e078d7e6d8c84face

pkg/tempo/remotewriteexporter/exporter.go
mapno: Added a few unit tests, although I'm not extremely happy with them. I'd like to work something like https://github.com/prometheus/compliance/tree/main/remote_write out to the exporter. Maybe in a future pass?

mapno:
Found an issue while updating the documentation. Agents in scraping mode can't have prometheus instances, so the exporter can't get one to push metrics.

joe-elliott:
Does this PR remove the ability to publish locally? I was thinking we would want both options available. I was also thinking that we need to the ability to log locally to the automatic logging feature.

PR #572, opened 2021-04-29
add ability to get an instance by name

rfratto:
#### PR Description 
An internal-only change for now that adds the ability to retrieve a managed instance by name. In the case of instances being grouped, the running group gets returned instead. 

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/prom/instance/manager.go
+++ b/pkg/prom/instance/manager.go
@@ -34,6 +34,9 @@ var (
 // Manager represents a set of methods for manipulating running instances at
 // runtime.
 type Manager interface {
+	// GetInstance retrieved a ManagedInstance by name.
+	GetInstance(name string) (ManagedInstance, error)
+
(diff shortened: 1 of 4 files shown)


pkg/prom/instance/manager.go
```diff
@@ -34,6 +34,9 @@ var (
 // Manager represents a set of methods for manipulating running instances at
 // runtime.
 type Manager interface {
+	// GetInstance retrieved a ManagedInstance by name.
```
rfratto, line 37:
```suggestion
	// GetInstance retrieves a ManagedInstance by name.
```

rfratto: Great suggestion, thanks!

mapno (approved):

PR #575, opened 2021-04-30
metrics: read global config when reloading instances

rfratto:
This PR fixes an issue where the global settings were not being read by instances when reloading. This was because the global settings were passed to the constructor of instances, but never updated when applying the new instance settings.

This could be reproduced by adding a new external_label and reloading the config. The new label will not be present on newer sent metrics. 

This PR fixes that problem by adding the global config to be a part of the instance config struct. It cannot be marshaled into the instance, and must be set manually by the code before applying the changes.

There are two code paths to consider here:

1. The scraping service will re-read the global config on the next reshard tick.
2. Non-scraping service will apply the global config as soon as the new config is read in.

This also changes the group manager to ensure the global settings are kept when creating a group.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ service config.
   This fixes a problem where integrations didn't work if `prometheus:` wasn't
   configured. (@rfratto)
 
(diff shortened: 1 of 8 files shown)


mattdurham (approved):
LGTM

PR #580, opened 2021-05-04
update cortex for example/docker-compose and example/k3d to use blocks

rfratto:
#### PR Description 
Updates from a very old Cortex to the latest one to take advantage of blocks storage. Uses the [example single-process config](https://github.com/cortexproject/cortex/blob/master/docs/configuration/single-process-config-blocks.yaml) for configuring Cortex.

#### PR Checklist
I'd consider this an "internal change," so nothing to do here.

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/example/docker-compose/docker-compose.integrations.yaml
+++ b/example/docker-compose/docker-compose.integrations.yaml
@@ -5,7 +5,7 @@
 #
 # that enable local testing of all components of the Agent.
 #
-# The agent/config/agent-local.yaml file holds a config to 
+# The agent/config/agent-local.yaml file holds a config to
 # scrape all of these services.
 version: "2"
 services:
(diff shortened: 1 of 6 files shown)


example/docker-compose/docker-compose.integrations.yaml
```diff
@@ -15,7 +15,7 @@ services:
   #
 
   grafana:
-    image: grafana/grafana:6.6.1
+    image: grafana/grafana:7.5.4
```
rfratto, line 18: This syncs with the default docker-compose file which was using a newer version of Grafana.

mattdurham (approved):
LGTM

PR #591, opened 2021-05-14
Retry getting prom instance in Tempo start up

mapno:
#### PR Description 

Fixes a race condition where the agent will fail to boot if the prometheus
instance that it's set to use to remote write metrics in the tracing
pipeline is not ready yet.

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/tempo/instance.go
+++ b/pkg/tempo/instance.go
@@ -13,12 +13,13 @@ import (
 	"github.com/grafana/agent/pkg/util"
 	"github.com/prometheus/client_golang/prometheus"
 	"go.opencensus.io/stats/view"
-	"go.uber.org/zap"
-
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/config/configmodels"
 	"go.opentelemetry.io/collector/service/builder"
+	"go.uber.org/zap"
 )
+
+const defaultRetryInterval = time.Second * 5
 
 // Instance wraps the OpenTelemetry collector to enable tracing pipelines
 type Instance struct {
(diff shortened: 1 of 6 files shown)


pkg/tempo/instance.go
```diff
@@ -141,11 +142,19 @@ func (i *Instance) buildAndStartPipeline(ctx context.Context, cfg InstanceConfig
 	}
 
 	if cfg.SpanMetrics != nil && len(cfg.SpanMetrics.PromInstance) != 0 {
-		prom, err := promManager.GetInstance(cfg.SpanMetrics.PromInstance)
-		if err != nil {
-			return fmt.Errorf("failed to get prometheus instance %s", cfg.SpanMetrics.PromInstance)
+		for {
+			select {
+			case <-ctx.Done():
```
joe-elliott, line 147: i don't think this context has a timeout. i think this will spin forever. 

we need to either add a timeout or max retries or something.

pkg/tempo/instance.go
```diff
@@ -141,11 +142,19 @@ func (i *Instance) buildAndStartPipeline(ctx context.Context, cfg InstanceConfig
...
+				return fmt.Errorf("failed to get prometheus instance %s", cfg.SpanMetrics.PromInstance)
+			default:
+				prom, err := promManager.GetInstance(cfg.SpanMetrics.PromInstance)
+				if err == nil {
+					ctx = context.WithValue(ctx, contextkeys.Prometheus, prom)
+					break
+				}
+				<-time.After(defaultRetryInterval)
```
rfratto, line 155: Instead of looping when building the pipeline, what do you think about passing the whole manager to the span metrics processor and perform the look up of the instance per span? It should be a fairly speedy lookup.

pkg/tempo/instance.go
mapno: Nice catch, thanks. I ended up going for @rfratto's suggestion of passing the manager to the exporter.

pkg/tempo/instance.go
mapno: Good idea, made an implementation in 944efc7b160c2867d88e85ea806d3f15d9718e81.

I think losing a few metrics during start up is reasonable

rfratto: That commit looks good, but did you mean to push it to this PR? I'm not seeing it here, only via that link.

mapno: I pushed it to the main repository, my bad. It should be fixed now!

rfratto (approved):
LGTM, though you might want to let Joe take a second look. Are the docs up to date?

mapno:
> LGTM, though you might want to let Joe take a second look. Are the docs up to date?

They are from the previous PR. This one doesn't change anything for the user.

PR #597, opened 2021-05-19
[Tempo] Automatic Logging: Fixed double logging of "svc"

joe-elliott:
#### PR Description 
Automatic logging double logs the "svc" name.  This fixes it so "svc" is only logged along with process key vals.

- [x] CHANGELOG updated 
- [ ] Documentation added
- [x] Tests updated


mapno (approved):
LGTM

PR #602, opened 2021-05-23
Support building MIPS (LE)

marefr:
#### PR Description 
Wanted to run the agent on an [Ubiquiti EdgeRouter X](https://www.ui.com/edgemax/edgerouter-x/). This device architecture is MIPS, hence the need for a MIPS little endian (LE) build of the agent. 

Not sure you're interested of building and releasing MIPS version, but it would be nice if you could support it via the Makefile at least. Probably a few network devices/routers out there which is MIPS-based. For now I'll try to keep these changes in my fork. Let me know if this is interesting and/or whether I should add support for the other MIPS architectures in the Makefile or anything else that's missing.

Reference/guide/scripts/binaries for running the agent on Ubibuiti EdgeRouter: https://github.com/marefr/grafana-agent-edgeos

```
> lscpu  
Architecture:          mips
Byte Order:            Little Endian
CPU(s):                4
On-line CPU(s) list:   0-3
Thread(s) per core:    2
Core(s) per socket:    2
Socket(s):             1
BogoMIPS:              581.63
L1d cache:             32K
L1i cache:             32K
L2 cache:              256K
```

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -184,6 +184,10 @@ example-dashboards:
 # A custom grafana/agent/seego image is built on top of the base image with
 # specific overrides. grafana/agent/seego is not pushed to Docker Hub and
 # can be built with "make seego".
+#
+# Note that dist/agent(ctl)-linux-mipsle targets are not included in dist target
+# and are included as separate targets to make it easier for users to build them
+# manually.
 dist: dist-agent dist-agentctl dist-packages
 	for i in dist/agent*; do zip -j -m $$i.zip $$i; done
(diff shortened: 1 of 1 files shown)


rfratto:
Since we're not going to include it with releases, can you add a comment above the dist/* targets explaining that the mipsle target is just there to make it easier for users to build manually?

marefr:
Sure thing 👍

rfratto (approved):
LGTM!

PR #611, opened 2021-05-25
Exemplars support

mapno:
#### PR Description 

Upgrade vendor of Prometheus to v2.27 which add exemplars support

None

#### PR Checklist

- [x] CHANGELOG updated 
- [ ] Documentation added
- [x] Tests updated


pkg/prom/wal/wal.go
```diff
@@ -272,6 +285,17 @@ func (w *Storage) loadWAL(r *wal.Reader) (err error) {
 					}
 				}
 				decoded <- samples
+			case record.Exemplars:
```
mapno, line 288: Probably no need to decode exemplars if we're not updating the series' ts

pkg/prom/instance/instance.go
```diff
@@ -189,6 +189,9 @@ func (c *Config) ApplyDefaults(global GlobalConfig) error {
 			return fmt.Errorf("found duplicate remote write configs with name %q", cfg.Name)
 		}
 		rwNames[cfg.Name] = struct{}{}
+
+		// Send exemplars by default
+		cfg.SendExemplars = true
```
rfratto, line 194: (Note that this means you can't ever disable it)

mapno: Ideally it should enabled by default and be disable-able through config, right?

I couldn't find a clean way to do this. Specially because `RemoteWrite` is a slice, so it's harder to apply a default. If you have any thoughts on how to achieve this, let me know. If not, I think leaving it as false by default and apply the config manually will be best.

mapno: Removed the default. We will need to explicitly enable it. But that is how prometheus works too.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ guide](./docs/migration-guide.md) for details.
 
 - [ENHANCEMENT] Add the option to log to stdout instead of a Loki instance. (@joe-elliott)
 
+- [ENHANCEMENT] Update Prometheus dependency to v2.27.0 and add exemplars support (@mapno)
```
rfratto, line 9: Can you also note the dependency bumps for Cortex and Loki/Promtail?

rfratto, line 9: Also, can you update the Prometheus entry in .github/depcheck.yml?

pkg/prom/wal/wal.go
```diff
@@ -62,13 +63,19 @@ func newStorageMetrics(r prometheus.Registerer) *storageMetrics {
...
+	})
+
 	if r != nil {
 		r.MustRegister(
 			m.numActiveSeries,
 			m.numDeletedSeries,
 			m.totalCreatedSeries,
 			m.totalRemovedSeries,
 			m.totalAppendedSamples,
+			m.totalAppendedExemplars,
```
rfratto, line 78: Can you make sure this gets unregistered in the Unregister method?

pkg/prom/wal/wal.go
```diff
@@ -622,9 +631,23 @@ func (a *appender) AddFast(ref uint64, t int64, v float64) error {
 	return nil
 }
 
-func (a *appender) AppendExemplar(ref uint64, l labels.Labels, e exemplar.Exemplar) (uint64, error) {
-	// remote_write doesn't support exemplars yet, so do nothing here.
-	return 0, nil
+func (a *appender) AppendExemplar(ref uint64, _ labels.Labels, e exemplar.Exemplar) (uint64, error) {
+	s := a.w.series.getByID(ref)
+	if s == nil {
+		return 0, fmt.Errorf("unknown series ref. when trying to add exemplar: %d", ref)
```
rfratto, line 637: Can you double check that this will never be true for the first exemplar for any series? i.e., when the first sample for a new series comes in, ref is always 0. I'd like to make sure that's not true for AppendExemplar.

pkg/prom/wal/wal.go
```diff
@@ -622,9 +631,23 @@ func (a *appender) AddFast(ref uint64, t int64, v float64) error {
...
+	}
+
+	// Ensure no empty labels have gotten through.
+	e.Labels = e.Labels.WithoutEmpty()
```
rfratto, line 641: I think this should be doing the same check for an empty label set and returning an error as the normal Append here.

rfratto (commented):
Great work!

CHANGELOG.md
mapno: I wasn't sure on how to mention this, as it's using unreleased commits for both projects.

pkg/prom/wal/wal.go
mapno: Yes, the current tsdb [implementation](https://github.com/prometheus/prometheus/blob/ef584a9df6b4d5b360aa99ddee605b9b92f5ad45/tsdb/head.go#L1385-L1388) for AppendExemplar doesn't create the series entry.  The series will already exist because of the way exemplars are scraped today:   on the same line as a sample, and the sample is processed first ([ref](https://github.com/prometheus/prometheus/blob/ef584a9df6b4d5b360aa99ddee605b9b92f5ad45/scrape/scrape.go#L1422-L1450)).

CHANGELOG.md
mapno: Updated in 92c0e4c8cd9e2e517323f0e0a1e38a922feea77c

pkg/prom/wal/wal.go
mapno: Fixed in e317dca929e613b1083c16f8c255f49e1564acd6, thanks!

pkg/prom/wal/wal.go
mapno: Added compliance with OpenMetrics in 28d4de2a6bc29648dcd46b4ff5b2290bf5303c30

CHANGELOG.md
rfratto: This is how I've done it before: 

> [ENHANCEMENT] Update Loki to 59a34f9867ce. This is a non-release build, and was needed to support multiple Loki instances. (@rfratto)

rfratto (approved):
Nice! This looks good from my perspective. @joe-elliott do you want to review before it gets merged?

joe-elliott:
I'm good if you're good!

PR #619, opened 2021-05-26
Add Grok Exporter Integration

roobalimsab:
#### PR Description 
This PR adds [grok_exporter](https://github.com/fstab/grok_exporter) integration to grafana agent

There is no issue created for this. We wanted to have this integration so that we can remote write grok exporter metrics seamlessly.

#### Notes to the Reviewer
The Grok Exporter is not coded for the prometheus.Collector interface. So I have reused pieces of code from the grok_exporter itself and wired the integration.

#### PR Checklist

- [v] CHANGELOG updated 
- [v] Documentation added
- [v] Tests updated


roobalimsab:
@rfratto sure thanks. The PR is complete. Do let me know your feedback/comments.

rfratto:
Hi, sorry for the delay, things have been busy. I'm still planning on taking a look at this this week.

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Main (unreleased)
 
+- [FEATURE] New integration: grok_exporter (@roobalimsab)
```
rfratto, line 3: Can you move this below the breaking change notification?

pkg/integrations/grok_exporter/config.go
```diff
@@ -0,0 +1,132 @@
...
+	"gopkg.in/yaml.v3"
+)
+
+// Config controls the grok_exporter integration.
+type Config struct {
+	Common config.Common `yaml:",inline"`
+
+	GrokConfig v3.Config `yaml:",inline"`
+
+	IncludeExporterMetrics bool `yaml:"include_exporter_metrics"`
```
rfratto, line 20: Nit: can you remove the extra newlines here?

pkg/integrations/manager.go
```diff
@@ -476,6 +476,12 @@ func (m *Manager) WireAPI(r *mux.Router) {
 		handler := loadHandler(key)
 		handler.ServeHTTP(rw, r)
 	})
+
+	for _, process := range m.integrations {
+		for route, handler := range process.i.CustomHandlers() {
+			r.Handle(route, handler)
+		}
+	}
```
rfratto, line 484: This bit isn't going to work with /-/reload support. If the integration gets restarted due to a configuration change, the handler will still be pointing to the old instance.

pkg/integrations/grok_exporter/grok_exporter.go
```diff
@@ -0,0 +1,262 @@
...
+	registry := prometheus.NewRegistry()
+	for _, m := range metrics {
+		registry.MustRegister(m.Collector())
+	}
+
+	nLinesTotal, nMatchesByMetric, procTimeMicrosecondsByMetric, nErrorsByMetric := initSelfMonitoring(metrics, registry)
+
+	if config.IncludeExporterMetrics {
+		registry.MustRegister(prometheus.NewProcessCollector(prometheus.ProcessCollectorOpts{}))
+		registry.MustRegister(prometheus.NewGoCollector())
```
rfratto, line 80: These should already be exposed via the Agent, I'm not sure we need them at the integration-level, too.

rfratto (commented):
Hi, thanks again for your PR. I've given it a high-level overview and left some small pieces of feedback where things would break.

In general, I'm a little concerned this might be out of scope for an integration: grok_exporter doesn't appear to do anything that you can't already do with the existing log support within the Agent, and I'd rather not have two mechanisms to do the same thing (ie create metrics from logs). Can you explain your use case a little more to me and why Loki support (ie Promtail) wouldn't work?

In the future, I'd recommend opening an issue to propose adding a feature like this first. I appreciate the work you've done, but this is a significantly sized integration and we'll need to justify accepting it before we start taking responsibility for it.

roobalimsab:
Hi, Thank you for the feedback.

>Can you explain your use case a little more to me and why Loki support (ie Promtail) wouldn't work?
My team provides monitoring and alerting as a solution for all the other teams in my company. We use Cortex at a large scale for collecting all kinds of metrics. We already have barito and new relic for logging and there are no plans to move to Loki.

**Why we use grok_exporter?**
We have built a SLO platform where SLIs are based on public apis. Currently, we are extending that to internal APIs that are behind HAProxy. HAProxy prometheus exporter does not publish percentile metrics for response time. To achieve that we are using grok_exporter to parse HAProxy logs and publish response time histogram, so that users can configure latency SLIs.

And sure, will open issues for similar scenarios in future.

CHANGELOG.md
roobalimsab: Done

pkg/integrations/grok_exporter/config.go
roobalimsab: Done

pkg/integrations/grok_exporter/grok_exporter.go
roobalimsab: Checked the metrics endpoint of the agent and you are right, the metrics are present. So removing these.

pkg/integrations/manager.go
roobalimsab: Refactored the code to work with agent reload

roobalimsab:
Hi @rfratto  Any updates/feedback on this?

rfratto:
Hey @roobalimsab, thanks for the ping. I've been giving this a fair amount of thought over the last few days. Unfortunately, I'm not interested in accepting this as an integration for now. I don't want to imply that we'll never agree to support this, but I'm erring on the side of caution for now.

This isn't an insignificant amount of code for an integration, and accepting it means we'll have to become familiar with grok_exporter so we can maintain the integration and answer support questions we get from users. Since the Grafana Agent can already transform logs into metrics through its embedded Promtail, it's very hard to justify adding a second way of doing that. 

I think your specific use case of "we want to transform metrics into logs but don't want to send logs to Loki" is interesting, and something that the Loki/Promtail maintainers might be interested in entertaining. One way of doing this is allowing Promtail to run without sending logs anywhere, which would still let it generate metrics from the logs it collects. If you're not interested in proposing that, I would encourage you to continue running grok_exporter as an independent application.

PR #623, opened 2021-05-27
Improve performance of kubernetes_sd_configs when using host_filter

### LINKED ISSUE #48: Explore reducing Kubernetes SD load when using host_filter
It appears that the overhead of Kubernetes SD increases across the agents as:

a) the number of nodes increase 
b) the number of pods increase

We noticed that a DaemonSet of 20 agents was using more CPU combined compared to a single instance of Prometheus that included more CPU-intensive tasks, such as TSDB, Recording Rules, and Alerts.

When the agents are running as a DaemonSet with host_filter, it would be more less CPU intensive if each agent only scraped pods on the node it's running on rather than pods across the entire cluster. 

I'm not sure how/if this can be done.

rfratto:
#### PR Description 
This change introduces a patching mechanism for `scrape_configs` to make performance-related tweaks for host filtering. The first performance tweak is to add a selector for `role: pod` to only look for pods on the same node as the Agent. 

This code is heavily inspired by (copied from) [Promtail](https://github.com/grafana/loki/blob/966bed72d3247a15c7f94a14ca13b536edb30f74/clients/pkg/promtail/targets/file/filetargetmanager.go#L95-L105).

Fixes #48. 

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/prom/instance/host_filter.go
+++ b/pkg/prom/instance/host_filter.go
@@ -91,6 +91,7 @@ func (f *HostFilter) PatchSD(scrapes []*config.ScrapeConfig) {
 
 }
 
+// SetRelabels updates the relabeling rules used by the HostFilter.
 func (f *HostFilter) SetRelabels(relabels []*relabel.Config) {
 	f.relabelMut.Lock()
 	defer f.relabelMut.Unlock()
--- a/pkg/prom/instance/host_filter.go
+++ b/pkg/prom/instance/host_filter.go
@@ -2,9 +2,13 @@ package instance
 
 import (
 	"context"
+	"fmt"
 	"net"
+	"sync"
 
 	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/config"
+	"github.com/prometheus/prometheus/discovery/kubernetes"
 	"github.com/prometheus/prometheus/discovery/targetgroup"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/pkg/relabel"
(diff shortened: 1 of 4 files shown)


pkg/prom/instance/instance.go
```diff
@@ -261,10 +263,16 @@ func New(reg prometheus.Registerer, cfg Config, walDir string, logger log.Logger
...
+	}
+
 	i := &Instance{
-		cfg:    cfg,
-		logger: logger,
-		vc:     vc,
+		cfg:        cfg,
+		logger:     logger,
+		vc:         vc,
+		hostFilter: NewHostFilter(hostname, cfg.HostFilterRelabelConfigs),
```
rfratto, line 275: I moved the host filter to span the lifetime of the instance. This lets us update the host filter relabel configs, which wasn't working before (and wasn't treated as an invalid update, either).

pkg/prom/instance/host_filter.go
```diff
@@ -58,13 +63,41 @@ func NewHostFilter(host string, relabels []*relabel.Config) *HostFilter {
...
+// that HostFilter will output as possible.
+func (f *HostFilter) PatchSD(scrapes []*config.ScrapeConfig) {
+	for _, sc := range scrapes {
+		for _, d := range sc.ServiceDiscoveryConfigs {
+			switch d := d.(type) {
+			case *kubernetes.SDConfig:
+				if d.Role == kubernetes.RolePod {
+					d.Selectors = []kubernetes.SelectorConfig{{
+						Role:  kubernetes.RolePod,
+						Field: fmt.Sprintf("spec.nodeName=%s", f.host),
```
mattdurham, line 85: Do we need to do any escaping here or is f.host input considered safe?

mattdurham (approved):
One comment, everything else looks good

pkg/prom/instance/host_filter.go
rfratto: Hmm, good question. It should be safe here, but I'll do a little investigation before merging.

rfratto:
Yeah, I can't see anything that would indicate we'd need to sanitize this. I'm going to merge.

PR #630, opened 2021-06-03
Display specific error messages when cloud-config fails

### LINKED ISSUE #360: agentctl: Show HTTP response on error from agentctl cloud-config
`agentctl cloud-config` is the command used during the install script to retrieve the Grafana Cloud-generated config. If an invalid stack ID is passed to the install script, the error is just reported as a generic 404. However, the response body contains more information:

```{"status":"error","error":"stack '1' not found"}```

The error in the response body should be reported back to the user.

### LINKED ISSUE #629: 500 error from cloud-config with no further information
I have used the install script to install the Grafana Cloud Agent, but the config file /etc/grafana-agent.yaml is empty.

Looking at the script, I see that it fetches the config by calling `grafana-agentctl cloud-config` so I'm trying to do the same.

I'm running:

`grafana-agentctl cloud-config -u 'xxx' -p 'xxxx'`

It exits with only:

`could not retrieve agent cloud config: unexpected status code 500`

I don't think it's a problem with the credentials being supplied, as if I e.g. change the API key being passed as the `-p` argument, I then get a 401 error.

The lack of additional detail in the response is somewhat related to #360 - it's possible that there's something in the response which would tell me what was wrong, but the response I get from cloud-config gives me nothing to go on.

rfratto:
#### PR Description 
Changes the cloud-config command to attempt parsing the response body before returning an error that the response code was unexpected. 

Fixes #360 
Closes #629 (dupe of #360)

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 - [FEATURE] (beta) A Grafana Agent Operator is now available. (@rfratto)
 
+- [ENHANCEMENT] Error messages when installing the Grafana Agent for Grafana
+  Cloud will now be shown. (@rfratto)
+
 # v0.15.0 (2021-06-03)
 
 BREAKING CHANGE: Configuration of Tempo Autologging changed in this release.
--- a/pkg/client/grafanacloud/client.go
+++ b/pkg/client/grafanacloud/client.go
@@ -49,10 +49,6 @@ func (c *Client) AgentConfig(ctx context.Context, stackID string) (string, error
 	}
 	defer resp.Body.Close()
 
-	if resp.StatusCode != 200 {
-		return "", fmt.Errorf("unexpected status code %d", resp.StatusCode)
-	}
-
 	// Even though the API returns json, we'll parse it as YAML here so we can
 	// re-encode it with the same order it was decoded in.
 	payload := struct {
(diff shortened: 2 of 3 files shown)


cristiangreco (approved):

bigpresh:
Am I being blind - I don't see where you're actually outputting any extra info here?

The `return "", fmt.Errorf("unexpected status code %d", resp.StatusCode)` looks like the same as before; did you mean to add something from the parsed response into it?  Or would the addition of `dec.SetStrict(true)` cause it to omit useful info (presumably only if the problem is that the response didn't contain valid YAML, but not if it did?

I'm not a Go coder, and I'm not very familiar with the Grafana Agent, so I may well be completely missing something obvious here.

I've tested it, but I'm getting the same result I did before - just `could not retrieve agent cloud config: unexpected status code 500`

rfratto:
> I don't see where you're actually outputting any extra info here?

It's (unfortunately) hidden from the diff but the extra info gets printed out a [few lines below](https://github.com/grafana/agent/blob/62a36feb2197a43c31b467077f8916ec10e4d04a/pkg/client/grafanacloud/client.go#L70-L72). That logic was always there, but was being cut out by me prematurely reading the response code.

The new flow here is:

1. Try to parse the response as JSON.
2. If it fails, and the result wasn't 200, return the status code as a best-effort error message. Otherwise, return the failure about reading the response body. This shouldn't happen in practice. 
3. If it succeeds, and the response JSON represents an error message, return that underlying error message, which is the part that was being hidden before. (If the response JSON doesn't represent an error message, use the success payload as the config file)

Once we do a release (sometime next week), the new error message will be `could not retrieve agent cloud config: request was not successful: <issue from server>`. Until then, if you're willing to hack around with things, you can manually run agentctl from the main branch :)

bigpresh:
Ohhh, I see - yeah, that makes sense now, thanks for the explanation!

I'll have a crack at building from the main branch and seeing if it gets me any more info.

PR #641, opened 2021-06-10
Upgrade OTel collector to v0.27.0

mapno:
#### PR Description

Upgrades OTel collector from `v0.21.0` to `v0.27.0`. The builder pkg is made internal and now we need to use the `service.Application` interface from two layers above.

None.

#### Notes to the Reviewer

It's using a personal fork to propagate the context to the builders, which is needed for our custom processors and exporters.

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated


pkg/tempo/internal/tempoutils/server.go
```diff
@@ -73,121 +73,144 @@ func NewServerWithRandomPort(callback func(pdata.Traces)) (srv *Server, addr str
...
+	var (
+		startInfo component.ApplicationStartInfo
+	)
 
-	var cfg map[string]interface{}
-	if err := yaml.NewDecoder(strings.NewReader(conf)).Decode(&cfg); err != nil {
-		panic("could not decode config: " + err.Error())
+	params := service.Parameters{
+		Factories:            factories,
+		ApplicationStartInfo: startInfo,
```
kvrhdn, line 87: Nitpick: this can be simplified a tiny bit, avoiding a var block.

```suggestion
		ApplicationStartInfo: component.ApplicationStartInfo{},
```

pkg/tempo/internal/tempoutils/server.go
```diff
@@ -73,121 +73,144 @@ func NewServerWithRandomPort(callback func(pdata.Traces)) (srv *Server, addr str
...
+			return nil, fmt.Errorf("failed to merge in mapstructure config: %w", err)
+		}
 
-	receivers, err := builder.NewReceiversBuilder(logger, startInfo, otelCfg, pipelines, factories.Receivers).Build()
-	if err != nil {
-		return nil, fmt.Errorf("failed to build receivers: %w", err)
-	}
-	if err := receivers.StartAll(context.Background(), nil); err != nil {
-		return nil, fmt.Errorf("failed to start receivers: %w", err)
+		otelCfg, err := config.Load(v, factories)
```
kvrhdn, line 167: The docs ask to verify the config using `ValidateConfig`. I'm not sure if the config is validated later but might be helpful to fail fast here?
```go
// Load loads a Config from Viper.
// After loading the config, need to check if it is valid by calling `ValidateConfig`.
func Load(
  ...
```

kvrhdn (commented):
Some small comments, didn't get through all the changes yet.

pkg/tempo/internal/tempoutils/server.go
mapno: Applied, thanks

pkg/tempo/internal/tempoutils/server.go
mapno: Thanks for pointing it out. The interface has changed in v0.27, so it no longer applies.

pkg/tempo/instance.go
```diff
@@ -148,47 +116,63 @@ func (i *Instance) buildAndStartPipeline(ctx context.Context, cfg InstanceConfig
...
 	}
 
-	err = i.exporter.StartAll(ctx, i)
-	if err != nil {
-		return fmt.Errorf("failed to start exporters: %w", err)
+	params := service.Parameters{
+		Factories:      factories,
+		BuildInfo:      buildInfo,
+		ParserProvider: &cfg,
+		LoggingOptions: []zap.Option{zap.Development()},
```
mapno, line 129: Maybe we want to disable the app's own logging. I'm not sure with this one

go.mod
```diff
@@ -108,5 +108,7 @@ exclude google.golang.org/grpc/examples v0.0.0-20200728065043-dfc0c05b2da9
 // Used for /-/reload
 replace github.com/weaveworks/common => github.com/rfratto/weaveworks-common v0.0.0-20210326192855-c95210d58ba7
 
-// loadbalancingexporter uses non-fixed version of batchpertrace which fetches latest and causes problems
-replace github.com/open-telemetry/opentelemetry-collector-contrib/pkg/batchpertrace => github.com/open-telemetry/opentelemetry-collector-contrib/pkg/batchpertrace v0.21.0
+// loadbalancingexporter uses non-fixed version of batchpersignal which fetches latest and causes problems
+replace github.com/open-telemetry/opentelemetry-collector-contrib/pkg/batchpersignal => github.com/open-telemetry/opentelemetry-collector-contrib/pkg/batchpersignal v0.27.0
+
+replace go.opentelemetry.io/collector => github.com/mapno/opentelemetry-collector v0.24.1-0.20210610132247-eb06509beaa2
```
rfratto, line 114: Can you move this to the grafana org? IMO we should have as few personal forks as possible. (I'm the worst offender of this and need to remove the rfratto instances here)

rfratto (commented):
I'm going to leave the real review of this to people more familiar with the collector, but just a small nit about using personal orgs

pkg/tempo/config.go
```diff
@@ -400,7 +398,19 @@ func formatPolicies(cfg []map[string]interface{}) ([]map[string]interface{}, err
 	return policies, nil
 }
 
-func (c *InstanceConfig) otelConfig() (*configmodels.Config, error) {
+// Get satisfies the service.ParserProvider interface
+func (c *InstanceConfig) Get() (*config.Parser, error) {
```
kvrhdn, line 402: Possible improvement: we can verify this at compile-time by adding one of these checks: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/tailsamplingprocessor/sampling/always_sample.go#L26
With this the code will not compile if the struct doesn't satisfy the interface anymore.

It's a nice extra check but don't know if this is super common in Go 😅

pkg/tempo/config.go
```diff
@@ -400,7 +398,19 @@ func formatPolicies(cfg []map[string]interface{}) ([]map[string]interface{}, err
...
+	otelConfig, err := c.otelConfig()
+	if err != nil {
+		return nil, err
+	}
+
+	return config.NewParserFromStringMap(otelConfig), nil
+}
+
+// configFactory satisfies service.ConfigFactory
```
kvrhdn, line 411: What is `service.ConfigFactory`? I can't find this in my IDE.

pkg/tempo/instance.go
```diff
@@ -81,60 +79,30 @@ func (i *Instance) Stop() {
 }
 
 func (i *Instance) stop() {
```
kvrhdn, line 81: Having a function `Stop` and `stop` seems confusing. Could you clarify the difference? We can't rename `Stop` (part of the OTel interface) but maybe `stop` can be described as `stopApp` or `shutdownApp`.

Basically: why would you want to call `stop` instead of `Stop`?

kvrhdn (commented):
Left some remarks, but overall this seems good. Nice work Mario! 💪🏻 

Disclaimer: I'm not familiar with the collector and how it's supposed to work, so I might be missing stuff 😓

go.mod
joe-elliott: Any particular reason we're not vendoring otelcol directly?

mapno: > What does v0.24.1-... refer to? Is it this commit: mapno/opentelemetry-collector@eb06509 Those Go version always confuse me.

Correct. It's referencing that commit. I also find go modules confusing, but the reason for that _version_ is that it's referencing an unreleased commit, thus it's using a [pseudo-version](https://golang.org/ref/mod#pseudo-versions). If I'm not mistaken, it's using the 3rd form for pseudo-versions, but I don't know why it's v0.24.x and not v0.27.x.

> Are you planning to PR these changes? If so, we can link the PR as comment, makes it easier to track when we can remove the fork.

> Any particular reason we're not vendoring otelcol directly?

Yes, that's the plan. I've just opened the [PR](https://github.com/open-telemetry/opentelemetry-collector/pull/3418) to the collector. Initially I used a fork to get things moving fast, so we could integrate rapidly the new sampling policies we're working on. It'll probably won't be necessary to use my fork or Grafana's fork if it gets merged soon.

> Can you move this to the grafana org?

I'll wait a bit to get a response from the upstream. If it doesn't get merged we can create the org fork and make it _official_.

pkg/tempo/config.go
mapno: Nice suggestion, added in 383c8c1857e93f39ef7d379d063cbdb4bbdf2995

pkg/tempo/automaticloggingprocessor/factory.go
```diff
@@ -59,20 +59,18 @@ func NewFactory() component.ProcessorFactory {
...
 
-func createDefaultConfig() configmodels.Processor {
+func createDefaultConfig() config.Processor {
 	return &Config{
-		ProcessorSettings: configmodels.ProcessorSettings{
-			TypeVal: TypeStr,
-			NameVal: TypeStr,
-		},
+		ProcessorSettings: config.NewProcessorSettings(config.NewIDWithName(TypeStr, TypeStr)),
+		LoggingConfig:     &AutomaticLoggingConfig{},
```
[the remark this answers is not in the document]
mapno: Fixed in 383c8c1857e93f39ef7d379d063cbdb4bbdf2995, thanks

pkg/tempo/instance.go
mapno: Yea, you're right. Ideally you only want to call the exported function, the other one is for internal use only. But it's confusing, changed in 383c8c1857e93f39ef7d379d063cbdb4bbdf2995.

kvrhdn (approved):
We just have to decide what to do with the fork, but other than that LGTM :+1:

go.mod
mapno: :woohoo: merged! https://github.com/open-telemetry/opentelemetry-collector/pull/3418

mapno:
Bumping this one more time to `v0.28.0` with the merged cmd's ctx commit.

pkg/prom/instance/instance.go
```diff
@@ -757,7 +756,7 @@ func Hostname() (string, error) {
 }
 
 func getHash(data interface{}) (string, error) {
-	bytes, err := json.Marshal(data)
+	bytes, err := yaml.Marshal(data)
```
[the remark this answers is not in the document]
kvrhdn: Okay, that's interesting! 😀  Thanks for sharing.

mapno:
Closed in favor of https://github.com/grafana/agent/pull/700

Issue #646, opened 2021-06-11
Adding windows images to docker manifest


MouleeGee:
In order to deploy Grafana agent as a windows container running on windows node pools

rfratto:
I'm going to mark this as keepalive; I think this is important enough to track long term.

rfratto:
Windows images will be available in 0.31 thanks to the work done by @erikbaranowski

PR #198, opened 2020-09-25
Add windows container build and production setup.

cyriltovena:
This adds build step and production setup to run the agent on windows container with Kubernetes.

Tested it and it works like a charm.

We need to discuss how we should be doing release of this image, I don't have much time to invest so I would hope someone can take over.

PR #647, opened 2021-06-11
Adding windows/amd64 into the Docker manifest

jtyr:
#### PR Description 
This PR is adding `windows/amd64` into the Docker manifest so we can use images from Windows nodes.

It should fix issue #646.

#### Notes to the Reviewer
N/A

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Main (unreleased)
 
+- [FEATURE] windows/amd64 added into the Docker manifest. (@jtyr)
+
 - [FEATURE] (beta) A Grafana Agent Operator is now available. (@rfratto)
 
 - [ENHANCEMENT] Error messages when installing the Grafana Agent for Grafana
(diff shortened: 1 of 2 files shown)


rfratto:
Hi, thanks for your PR. Have you tested this locally with buildx? Windows support in the Docker manifest is more complicated than this. If you look at the Agent's [Dockerfile.buildx](https://github.com/grafana/agent/blob/main/cmd/agent/Dockerfile.buildx), you'll see that the final image is based off of [debian:buster-slim](https://hub.docker.com/layers/debian/library/debian/buster-slim/images/sha256-5f53426c1746b2f4e7ce7f11acb8e20755687ee13ea036e56ddc027aa218f61a?context=explore), which for obvious reasons will not work when building for a Windows image target. 

It's likely the build scripts will need to change to support a single image that has both Linux and Windows-compatible Docker images in a single manifest. #198 did some preliminary work with this, but it gets pushed as a separate image, so it's not completely where I'd like it to be yet.

One possible way of handling this is to split up the `docker buildx build` commands, changing them to `--load` (so they don't immediately get pushed) and having them build one platform at a time. Then, `docker buildx imagetools create` can combine the individual images into one manifest. There may be other ways to do this, but I'm not aware of any at the moment.

Issue #655, opened 2021-06-15
Operator: changing secret does not reconcile deployment


rfratto:
Steps to reproduce:

1. Create a PrometheusInstance that uses `additionalScrapeConfigs`, referencing a list of empty jobs.
2. Modify the secret, giving it a job.
3. Notice that a reconcile did not occur.

It looks like the event handlers are not currently watching referenced assets.

PR #656, opened 2021-06-15
Operator: tests and fix various small issues

rfratto:
#### PR Description 

This PR adds tests to validate the operator and fixes various small issues:

1. BearerTokenFile doesn't exist for PodMonitors
2. MatchLabels may be null
3. Fix `pod_monitor.libsonnet` import
4. Use `yaml.v3` for Jsonnet `unmarshalYAML`, which unmarshals to
   `map[string]interface{}` instead of `map[interface{}]interface{}`
5. Define `SHARDS` as an environment variable so custom scrape jobs can
   reference it

#### Notes to the Reviewer
The unit tests are quite a bit longer than I'd like them to be, but I can't think of a way to break them up any more, and they were useful for finding a few extra things I missed.

Feel free to just 🙈 through the tests, I suspect the Jsonnet code might not survive very long. Cue one day?

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/operator/config/config_test.go
+++ b/pkg/operator/config/config_test.go
@@ -6,10 +6,10 @@ import (
 
 	"github.com/grafana/agent/pkg/operator/assets"
 	"github.com/grafana/agent/pkg/util"
-	prom_v1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/api/core/v1"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8s_yaml "sigs.k8s.io/yaml"
 
 	grafana "github.com/grafana/agent/pkg/operator/apis/monitoring/v1alpha1"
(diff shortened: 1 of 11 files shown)


example/k3d/environment/operator.jsonnet
```diff
@@ -0,0 +1,279 @@
...
+};
+
+local configMap = k.core.v1.configMap;
+local container = k.core.v1.container;
+local daemonSet = k.apps.v1.daemonSet;
+local deployment = k.apps.v1.deployment;
+local policyRule = k.rbac.v1.policyRule;
+local serviceAccount = k.core.v1.serviceAccount;
+local namespace = k.core.v1.namespace;
+local secret = k.core.v1.secret;
```
rfratto, line 15: We don't need all of these.

pkg/operator/config/config_test.go
```diff
@@ -137,3 +139,294 @@ func TestBuildConfig(t *testing.T) {
 		})
 	}
 }
+
+// TestFullConfig tests generation of a config that may be used to collect
+// metrics from Kubernetes.
+func TestFullConfig(t *testing.T) {
```
rfratto, line 145: Split this up into unit tests. Test individual components directly. We can have an all-in-one test just to make sure different components get invoked.

pkg/operator/reconciler.go
```diff
@@ -66,16 +66,20 @@ func (r *reconciler) Reconcile(ctx context.Context, req controller.Request) (con
...
 
 	// Update our notifiers with the objects we discovered from building the
 	// deployment. This allows us to re-reconcile when any of the objects that
 	// composed our final deployment changes.
 	for _, secondary := range secondaryResources {
 		r.eventHandlers[secondary].Notify(req.NamespacedName, builder.ResourceSelectors[secondary])
 	}
+	for _, assetRef := range assetRefs {
+		_ = assetRef
+	}
```
rfratto, line 79: Didn't mean to commit this :)

rfratto (commented):
#655 is another open issue, but will probably need a larger fix and should be put in a separate PR.

rfratto:
The FAQ change and k3d change should be moved to a separate PR.

rfratto:
This is ready for review now. The unit tests are quite a bit longer than I'd like them to be, but I can't think of a way to break them up any more, and they were useful for finding a few extra things I missed.

pkg/operator/config/templates/component/probe.libsonnet
```diff
@@ -207,23 +209,8 @@ function(
...
     ])
   ),
-
-  metric_relabel_configs: if endpoint.MetricRelabelConfigs != null then optionals.array(
-    std.filterMap(
-      function(c) !(c.TargetLabel != '' && enforcedNamespaceLabel != '' && c.TargetLabel == enforcedNamespaceLabel),
-      function(c) new_relabel_config(c),
-      k8s.array(endpoint.MetricRelabelConfigs),
-    )
-  ),
```
mapno, line 221: Is this just removed?

pkg/operator/resources.go
```diff
@@ -191,6 +191,12 @@ func generateStatefulSetSpec(
 	d config.Deployment,
 	shard int32,
 ) (*apps_v1.StatefulSetSpec, error) {
+
+	shards := minShards
+	if reqShards := d.Agent.Spec.Prometheus.Shards; reqShards != nil && *reqShards > 1 {
```
mapno, line 196: `d.Agent` could be nil too, right?

mapno (commented):
LGTM, just left to minor comments

pkg/operator/config/templates/component/probe.libsonnet
rfratto: Yep, probes don't have this defined.

pkg/operator/resources.go
rfratto: It shouldn't be, this code only gets invoked we're operating on a non-nil resource hierarchy

mapno (approved):
LGTM

PR #660, opened 2021-06-16
don't return ErrNotFound on bad ref in Append

### LINKED ISSUE #659: Do not return ErrNotFound when appending an invalid ref ID
See the following code:
 
https://github.com/grafana/agent/blob/58467d45d85ec4af77a2ef41342b58943a5c1d95/pkg/prom/wal/wal.go#L615-L619

This isn't what the TSDB Head does anymore, and the scrape cache no longer will create a new entry if `storage.ErrNotFound` is returned.

rfratto:
#### PR Description 
Prometheus 2.26 changed the storage interface, removing the `AddFast` and `Add` methods, replacing with a single `Append` method. This also changed the expectation that `Append` should _not_ return `storage.ErrNotFound` on an invalid ref ID like `AddFast` used to. 

The 2.26 scraper will fail the scrape if `storage.ErrNotFound` is returned, and failed scrapes won't clear the cache, so the target will continue to fail to be scraped for the remainder of the process lifetime. This can be triggered primarily by a target going down for 2 GC cycles, when ref IDs of stale series are purged.

Fixes #659

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/prom/wal/wal.go
+++ b/pkg/prom/wal/wal.go
@@ -624,7 +624,7 @@ func (a *appender) getOrCreate(l labels.Labels) (series *memSeries, created bool
 func (a *appender) AppendExemplar(ref uint64, _ labels.Labels, e exemplar.Exemplar) (uint64, error) {
 	s := a.w.series.getByID(ref)
 	if s == nil {
-		return 0, errors.Wrapf(storage.ErrNotFound, "unknown series ref %d when trying to add exemplar", ref)
+		return 0, fmt.Errorf("unknown series ref. when trying to add exemplar: %d", ref)
 	}
 
 	// Ensure no empty labels have gotten through.
--- a/pkg/prom/wal/wal.go
+++ b/pkg/prom/wal/wal.go
@@ -578,18 +578,17 @@ func (a *appender) Append(ref uint64, l labels.Labels, t int64, v float64) (uint
 			return 0, errors.Wrap(tsdb.ErrInvalidSample, fmt.Sprintf(`label name "%s" is not unique`, lbl))
 		}
 
-		ref = a.w.ref.Inc()
-		series = &memSeries{ref: ref, lset: l}
+		var created bool
+		series, created = a.getOrCreate(l)
+		if created {
+			a.series = append(a.series, record.RefSeries{
+				Ref:    series.ref,
+				Labels: l,
+			})
 
-		a.series = append(a.series, record.RefSeries{
-			Ref:    ref,
-			Labels: l,
-		})
-
-		a.w.series.set(l.Hash(), series)
-
-		a.w.metrics.numActiveSeries.Inc()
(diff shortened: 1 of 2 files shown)


pkg/prom/wal/wal.go
```diff
@@ -566,78 +565,54 @@ type appender struct {
...
-	return ref, a.AddFast(ref, t, v)
-}
-
-func (a *appender) Add(l labels.Labels, t int64, v float64) (uint64, error) {
-	hash := l.Hash()
-	series := a.w.series.getByHash(hash, l)
-	if series != nil {
-		return series.ref, a.AddFast(series.ref, t, v)
-	}
-
```
rfratto, line 581: tsdb.Head in 2.26 doesn't do a getByHash first 🤷‍♂️ Leaving it out of the new impl.

pkg/prom/wal/wal.go
```diff
@@ -144,10 +144,7 @@ func NewStorage(logger log.Logger, registerer prometheus.Registerer, path string
 		deleted: map[uint64]int{},
 		series:  newStripeSeries(),
 		metrics: newStorageMetrics(registerer),
-
-		// The first ref ID must be non-zero, as the scraping code treats 0 as a
-		// non-existent ID and won't cache it.
-		nextRef: 1,
+		ref:     atomic.NewUint64(0),
```
rfratto, line 147: Changed to 0 by default because `Inc` will return the value post-increment.

pkg/prom/wal/wal.go
```diff
@@ -566,78 +565,66 @@ type appender struct {
...
+	series = &memSeries{ref: a.w.ref.Inc(), lset: l}
+	a.w.series.set(l.Hash(), series)
+	return series, true
 }
 
 func (a *appender) AppendExemplar(ref uint64, _ labels.Labels, e exemplar.Exemplar) (uint64, error) {
 	s := a.w.series.getByID(ref)
 	if s == nil {
-		return 0, fmt.Errorf("unknown series ref. when trying to add exemplar: %d", ref)
+		return 0, errors.Wrapf(storage.ErrNotFound, "unknown series ref %d when trying to add exemplar", ref)
```
rfratto, line 627: Oh, I see, this is taken right out of the TSDB Head. I'll revert this line.

```suggestion
		return 0, fmt.Errorf("unknown series ref. when trying to add exemplar: %d", ref)
```

mapno (approved):
LGTM. My only question is probably not related to this change: when can a series be found by hash but not ref ID?

rfratto:
> LGTM. My only question is probably not related to this change itself: when can a series be found by hash but not ref ID?
Yeah, I was wondering the same thing. It could happen if a target goes away from the scraper and comes back. The scraper wouldn't have the ref ID cached anymore, so it would pass a `0` instead, but we have the ability to look up the existing entry.

PR #676, opened 2021-06-21
Initialize biggest ref to existing ref when reading a segment

### LINKED ISSUE #675: Grafana agent 0.16.0 reporting incorrect metric values
Our setup:
We have a blue-green Grafana Agent cluster doing remote writes to Cortex.

We upgraded our Grafana Agent deployment to 0.16.0 from 0.13.1. We deployed 0.16.0 to our blue cluster which consisted of brand new machines; no WALs from the previous installation. During the blue cluster deployment, our cortex cluster were entirely accepting metrics from out green cluster; so the metrics are still correct at this point. Once all of our blue cluster machines were running we triggered the upgrade on our green cluster; at this point cortex now starts accepting writes from out 0.16.0 GAs which are reporting incorrect values (ex: we expect 64 for certain values but are getting 7 Billion). All metrics being reported were significantly incorrect. At this point we will immediately reverted to 0.13.1, which again reported the correct values.

Attached is one of our dashboard showing the number of agents we expect in one of our clusters. Please note I've converted the axis to Log10 as linear scale makes the original value looks like its 0

Screen Shot 2021-06-21 at 2 06 52 PM

rfratto:
#### PR Description 
`biggestRef` was incorrectly initialized to 0, which only works when there's no other segments, since that code is being called per-segment. If there was a segment with no series in it (which always happened with a new WAL is opened for writing), then the series ID will always be set to 0 and will always overlap with other metrics. 

This bug depends on an instance being restarted or the process crashing and restarting to trigger.

Fixes #675

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 - [FEATURE] Add TLS config options for tempo `remote_write`s. (@mapno)
 
+- [BUGFIX] Fix issue where replaying a WAL caused data corruption. (@rfratto)
+
 # v0.16.0 (2021-06-17)
 
 - [FEATURE] (beta) A Grafana Agent Operator is now available. (@rfratto)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 - [FEATURE] Add TLS config options for tempo `remote_write`s. (@mapno)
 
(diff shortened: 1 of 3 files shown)


mattdurham (approved):
LGTM

PR #689, opened 2021-06-25
Integrations always running regardless of enabled flag

### LINKED ISSUE #686: Can't seem to limit series with integrations
Hi there,

We are deploying a fleet of raspberry pi 4s (raspian) with grafana-agent installed to ship logs and some selected metrics -- except we can't seem to select the metrics.

Given a configuration file like this on agent v 0.15:

```yaml
integrations:
  agent:
    enabled: false
  node_exporter:
    enabled: true
prometheus:
   ....
server:
   .....
loki:
   ....
```

We still receive a boggling number of series, including all of the `agent_` series. I've experimented quite a bit with the options can't find anything that works. Originally I thought that I could use `set_collectors` as documented in the config reference, but that doesn't seem to get me there to reduce the number of collectors --even disabling it doesn't appear to work.

For a single pi:
Screen Shot 2021-06-24 at 2 21 54 PM

mattdurham:
#### PR Description 

The manager is ALWAYS collecting the integration even if enabled is set to false.  

Honor the enabled flag, when the manager was changed this behavior looks to have been lost.

Closes #686

#### PR Checklist

- [ X] CHANGELOG updated


pkg/integrations/manager.go
```diff
@@ -280,6 +280,11 @@ func (m *Manager) ApplyConfig(cfg ManagerConfig) error {
 			shouldCollect = *common.ScrapeIntegration
 		}
 
+		// If Enabled is false then we should override the should collect
+		if p.cfg.CommonConfig().Enabled == false {
```
rfratto: Hmm, I don't think this is being done in the right place. The loop in 210 would still spawn the integration, which it probably shouldn't be doing.

One way to handle this might be to add a function to return the list of only enabled integrations and iterate over that instead when calling `ApplyConfig`. That way disabling an integration and removing it from the file are treated as the same.

rfratto (commented):
Nice catch!

pkg/integrations/manager.go
mattdurham: My thought on that was looking at the documentation is that `enabled` flag enables the scraping but if you include the integration in the configuration with `enabled = false` it still will get the statistics but won't send them. I guess *should* there be a difference in disabling an integration vs removing it, if the answer is no then I will put in my original fix. Specifically with the agent integration having it available but not sending for debugging purposes seemed useful. Can def go either way though.

rfratto: Nah, the intent is that `enabled` drives whether the integration should be ran (so disabling it is the same as removing it), and `scrape_integration` determines if that running integration should be scraped by the Agent automatically.

cmd/agent/main.go
```diff
@@ -47,6 +47,17 @@ func main() {
...
+		// Iterate over the configurations and remove any that Enabled = false, this works by setting i to the last element
+		//   and then returning the array minus the last element. This loses order but is performant.
+		for i := 0; i < len(cfg.Integrations.Integrations);{
+			if !cfg.Integrations.Integrations[i].CommonConfig().Enabled {
+				cfg.Integrations.Integrations[i] = cfg.Integrations.Integrations[len(cfg.Integrations.Integrations)-1]
+				cfg.Integrations.Integrations = cfg.Integrations.Integrations[:len(cfg.Integrations.Integrations)-1]
+			} else {
+				i++
+			}
+		}
```
rfratto, line 59: Can you create a new function and call it from integrations.Manager.ApplyConfig instead? Filtering the actual slice here will omit the config from the results of `/-/config` which might be a little confusing to users.

mattdurham: This got more complicated. When you remove a configuration it did not stop the scraper, so when a configuration is disabled/removed and reload called Agent would list scrape_error = 404. Since the integration had stopped but the scraper was still running.

pkg/integrations/manager.go
```diff
@@ -220,9 +222,14 @@ func (m *Manager) ApplyConfig(cfg ManagerConfig) error {
 				continue
 			}
 			p.stop()
+			m.im.DeleteConfig(key)
```
mattdurham, line 223: Beforehand the scraper was not stopping, but the integration was.

pkg/integrations/manager.go
```diff
@@ -268,6 +279,7 @@ func (m *Manager) ApplyConfig(cfg ManagerConfig) error {
 
 		_ = m.im.DeleteConfig(key)
 		process.stop()
+		m.im.DeleteConfig(key)
```
rfratto, line 280: Why do we need to call this twice?

pkg/integrations/manager.go
```diff
@@ -220,9 +222,14 @@ func (m *Manager) ApplyConfig(cfg ManagerConfig) error {
 				continue
 			}
 			p.stop()
+			m.im.DeleteConfig(key)
 			delete(m.integrations, key)
 		}
 
+		if !ic.CommonConfig().Enabled {
+			continue
+		}
```
rfratto, line 229: I still think a function to return a slice of enabled integrations would be easier to follow than having to maintain this check across the two loops, IMO.

pkg/integrations/manager.go
rfratto: I'm not following how this is necessary, the second loop should be deleting the config?

rfratto:
Does this fix #686? That reports that set_collectors doesn't appear to be working with node_exporter.

pkg/integrations/manager.go
mattdurham: With the addition of the enabled check around line 227 the integration process is not started so the integration isn't added to the to m.integrations.

pkg/integrations/manager.go
mattdurham: We still need the integrations in the array to disable/remove them, ie going from enabled -> disabled on reload.

pkg/integrations/manager.go
mattdurham: Not at all and removed.

mattdurham:
It fixes the enable issue that cropped during #686 , I was unable to reproduce an actual issue with set collectors. I tested several variations and they behaved as expected.

rfratto (commented):
Thanks! Can you add a regression test for this? 

I'd also like to see a test that provides we need m.im.DeleteConfig in there twice, because it definitely _seems_ unnecessary which makes it a candidate for accidental removal. Having a test would help prevent this.

tbrandsma-rw:
Any updates on when this fix will be merged in? I am currently dealing with the fallout of this issue because the enabled flag isn't being honored.

mattdurham:
Updated with unit tests and explanation below. Also had to update the mock integration so marshaling would work in the comparison. No tests were broken.

Situation removal of config on line 225 in manager.go

1. Config is ENABLED
2. Integrations and Configs and Processes are all ENABLED
   1. integration process is started
   2. instance process is started
   3. configs added
3. ApplyConfig is called with integration DISABLED
   1. Enter loop at 210
   2. exist check succeeds
   3. yaml is different
   4. integration process is stopped
   5. integration is removed from manager integrations
   6. config is DISABLED so line 229 moves to next integration
   7. loop over manager integrations is called
   8. since the integration was removed in the above the DeleteConfig is never called
   9. the instance process continues running

If you run TestManager_IntegrationEnabledToDisabledReload with line 225 commented out, then the TestManager_IntegrationEnabledToDisabledReload will fail.

mattdurham:
As far as a release, currently unknown, it is going to be bundled with some other large changes.

pkg/integrations/manager.go
```diff
@@ -223,6 +226,10 @@ func (m *Manager) ApplyConfig(cfg ManagerConfig) error {
...
```
rfratto, line 231: Looks like this is in here twice :)

rfratto (approved):
LGTM after removing the duplicated if statement in the first loop

PR #690, opened 2021-06-25
fix race condition in initializing cluster

### LINKED ISSUE #688: 'invalid memory address or nil pointer dereference' occurred and the agent crashed when restart
Hi there,

I have deployed an agent cluster with 3 node, and using a 3 node consul cluster as the config-store and ring-store in kubernete cluster. When I scale down the replicas of agent to 0 and  scale up the replicas to 3 again, following errors occurred and the agent crashed.

image

The produce steps as below
---
1. With consul as the ring store
2. Scale down  the agent replica count to 0
3. Scale up the agent replica count to 3
4. The errors come out like that occurred

Agent version is v0.15 or v0.16

rfratto:
#### PR Description 
Fixes a race condition while initializing the cluster. There was even a comment warning about this exact thing happening but the lock wasn't ever held when calling `Reshard`. 

Fixes #688.

#### Notes to the Reviewer
The race checker should've caught this but it's difficult to reproduce in a test since it depends on the speed of CPUs, the work done when spinning up the cluster, and how many CPUs are available for scheduling goroutines.  

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@
   the matching release version of the Grafana Agent instead of v0.14.0.
   (@rfratto)
 
+- [BUGFIX] Fix race condition that may occur and result in a panic when
+  initializing scraping service cluster. (@rfratto)
+
 # v0.16.1 (2021-06-22)
 
 - [BUGFIX] Fix issue where replaying a WAL caused incorrect metrics to be sent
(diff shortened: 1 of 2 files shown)


mattdurham (approved):
LGTM

PR #693, opened 2021-06-28
Add promqtt integration

sh0rez:
#### PR Description 

Adds an integration for promqtt, a Prometheus <-> MQTT bridge I've
written over the weekend. For more details consider the readme at
https://github.com/sh0rez/promqtt

None

#### Notes to the Reviewer

This does not yet work as expected. The endpoint `agent:12345/integrations/promqtt/mqtt` returns 404, I cannot figure out why though :/

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -34,8 +34,8 @@ require (
 	github.com/prometheus-community/windows_exporter v0.0.0-00010101000000-000000000000
 	github.com/prometheus-operator/prometheus-operator v0.47.0
 	github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.47.0
-	github.com/prometheus/client_golang v1.10.0
-	github.com/prometheus/common v0.23.0
+	github.com/prometheus/client_golang v1.11.0
+	github.com/prometheus/common v0.26.0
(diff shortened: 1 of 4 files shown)


RichiH:
+1 on intention.

I will be able to test this by the middle of the week, hopefully.

My high level concern before being able to play & test is to ensure we have feature parity with
* https://github.com/inovex/mqtt_blackbox_exporter
* https://github.com/hikhvar/mqtt2prometheus
* https://github.com/kpetremann/mqtt-exporter

soonish and that intended future work is put into issues in upstream repo.

All in all things should ideally be ready to be used before EOW when hackathon ends.

rfratto:
> This does not yet work as expected. The endpoint agent:12345/integrations/promqtt/mqtt returns 404, I cannot figure out why though :/

Ah, we hard code each integration to be exposed over /integrations/<name>/metrics at the moment 🙃 You can find this logic in WireAPI in manager.go (I'm on my phone right now so it's hard to link directly)

Issue #661, opened 2021-06-16
Autologging: span attributes as log labels


mapno:
Automatic logging searches for a given set of span attributes and logs them as key-value pairs. This allows for automatically building a mechanism for trace discovery, in which traces can be searched with LogQL. However, most of the search requires parsing the log line (e.g. `logfmt`), which involves extra overhead compared to using log labels.

We should consider allowing to use labels for this attributes. Labels are passed to the log [entry](https://github.com/grafana/agent/blob/main/pkg/tempo/automaticloggingprocessor/automaticloggingprocessor.go#L226-L228).

eamonryan:
This would help a ton with Grafana jumping from Metrics -> Logs in split pane view, as it relies on common labels between the two to automatically build a Loki query when you open a split!

PR #703, opened 2021-07-02
Support autologging span attributes as loki labels

mapno:
#### PR Description 

Add support autologging span attributes as log labels. This skips the need to format the log line and allows to directly search via labels. But as a downside, it increases cardinality.

Fixes https://github.com/grafana/agent/issues/661

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated


mapno:
I'm not completely sold on this approach. It depends on the attribute/tag to be logged in the log line, duplicating data. It also does not support to add default values.

pkg/tempo/automaticloggingprocessor/automaticloggingprocessor.go
```diff
@@ -85,12 +87,18 @@ func newTraceProcessor(nextConsumer consumer.Traces, cfg *AutomaticLoggingConfig
 	cfg.Overrides.DurationKey = override(cfg.Overrides.DurationKey, defaultDurationKey)
 	cfg.Overrides.TraceIDKey = override(cfg.Overrides.TraceIDKey, defaultTraceIDKey)
 
+	labels := make(map[string]bool, len(cfg.Labels))
```
joe-elliott, line 90: `map[string]struct{}` is i think the "right" way to do a flag map like this

pkg/tempo/automaticloggingprocessor/automaticloggingprocessor.go
```diff
@@ -191,7 +217,7 @@ func (p *automaticLoggingProcessor) spanKeyVals(span pdata.Span) []interface{} {
 	atts = append(atts, spanDuration(span))
 
 	atts = append(atts, p.cfg.Overrides.StatusKey)
-	atts = append(atts, span.Status().Code())
+	atts = append(atts, span.Status().Code().String())
```
joe-elliott, line 220: to reduce spam i wonder if we should only log this if it does not equal STATUS_CODE_UNSET

docs/configuration-reference.md
```diff
@@ -2095,6 +2095,8 @@ remote_write:
       [ process_attributes: <string array> ]
       # timeout on sending logs to Loki
       [ timeout: <duration> | default = 1ms ]
+      # configures a set of key values that will be logged as loki labels
```
joe-elliott, line 2098: note that it does not impact the stdout backend.

pkg/tempo/automaticloggingprocessor/automaticloggingprocessor.go
```diff
@@ -134,6 +145,21 @@ func (p *automaticLoggingProcessor) ConsumeTraces(ctx context.Context, td pdata.
...
+func (p *automaticLoggingProcessor) spanLabels(keyValues []interface{}) model.LabelSet {
+	if len(keyValues) == 0 {
+		return nil
+	}
+	ls := make(map[model.LabelName]model.LabelValue, len(keyValues))
+	for i := 0; i < len(keyValues); i += 2 {
+		k := keyValues[i]
+		v := keyValues[i+1]
+		if p.labels[k.(string)] {
+			ls[model.LabelName(k.(string))] = model.LabelValue(v.(string))
```
joe-elliott, line 157: if `k.(string)` is guaranteed to not panic can we just take a `[]string` instead?

pkg/tempo/automaticloggingprocessor/automaticloggingprocessor.go
```diff
@@ -222,10 +248,11 @@ func (p *automaticLoggingProcessor) exportToLoki(kind string, traceID string, ke
 		return
 	}
 
+	// Add loki label
+	labels[model.LabelName(p.cfg.Overrides.LokiTag)] = model.LabelValue(kind)
```
joe-elliott, line 252: spanLabels can return nil, does this work?

pkg/tempo/automaticloggingprocessor/automaticloggingprocessor_test.go
```diff
@@ -233,3 +234,32 @@ func TestDefaults(t *testing.T) {
 	require.Equal(t, defaultDurationKey, p.(*automaticLoggingProcessor).cfg.Overrides.DurationKey)
 	require.Equal(t, defaultTraceIDKey, p.(*automaticLoggingProcessor).cfg.Overrides.TraceIDKey)
 }
+
+func TestLabels(t *testing.T) {
```
joe-elliott, line 238: make sure to test no labels

pkg/tempo/automaticloggingprocessor/automaticloggingprocessor.go
mapno: `map[string]struct{}` uses less memory since an empty struct is 0 bytes, but it's very similar in performance to `map[string]bool` (see https://gist.github.com/davecheney/3be245c92b61e5045f75). I think `map[string]bool` makes for a bit clearer code.

rfratto: FWIW I try to use `map[string]struct{}` whenever I can, because there's always a chance that a bug could creep in that sets an entry to `false` in the map which would be treated as if it didn't exist.

pkg/tempo/automaticloggingprocessor/automaticloggingprocessor.go
mapno: Yeah, good point. I don't have a strong opinion on this, I'm fine removing it.

docs/configuration-reference.md
mapno: Noted!

pkg/tempo/automaticloggingprocessor/automaticloggingprocessor.go
mapno: Changed the logic a bit to make it cleaner and safer.

pkg/tempo/automaticloggingprocessor/automaticloggingprocessor.go
mapno: Changed to return a zero value, nice catch!

pkg/tempo/automaticloggingprocessor/automaticloggingprocessor_test.go
mapno: Done, thanks!

pkg/tempo/automaticloggingprocessor/automaticloggingprocessor.go
mapno: That's a good point, changed in 7efae7e3517625a38c1d3c769b5763f53091f770

joe-elliott (approved):
one nit. nice work.

this will be nice in our infra. everytime i query this log stream i wish there were more filters.

docs/configuration/tempo-config.md
```diff
@@ -107,6 +107,11 @@ automatic_logging:
   [ process_attributes: <string array> ]
   # Timeout on writing logs to Loki when backend is "logs_instance."
   [ timeout: <duration> | default = 1ms ]
+  # Configures a set of key values that will be logged as labels
+  # They need to be span or process attributes logged in the log line
+  #
+  # This feature only affects logs_instace backends
```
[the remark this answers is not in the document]
rfratto: Also I misspelled instance :)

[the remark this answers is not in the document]
mapno: Changed, thanks!

pkg/tempo/automaticloggingprocessor/automaticloggingprocessor.go
```diff
@@ -190,8 +228,11 @@ func (p *automaticLoggingProcessor) spanKeyVals(span pdata.Span) []interface{} {
 	atts = append(atts, p.cfg.Overrides.DurationKey)
 	atts = append(atts, spanDuration(span))
 
-	atts = append(atts, p.cfg.Overrides.StatusKey)
-	atts = append(atts, span.Status().Code())
+	// Skip STATUS_CODE_UNSET to be less spammy
```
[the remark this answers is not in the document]
mapno: Nice suggestion, mentioned in the changelog

Issue #691, opened 2021-06-27
When using -config.expand-env flag, the agent stops sending metrics to Grafana Cloud


brpaz:
Hello.
I have deployed Grafana Agent into a Kubernetes cluster and I am pushing metrics to Grafana cloud. I was trying to remove the credentials stored into the Config file, stored as config map, and found out that I use could envrionment variables, that would be replaced at runtime by the agent with the correct values.

I follow the instructions to set the `-config.expand-env=true` flag, but after set it the agent just stop sending metrics to Grafana Cloud. I haven´t done anything more. I have remove the flag, the metrics start appearing again:

![graph](https://user-images.githubusercontent.com/184563/123551107-49d1f800-d768-11eb-8d1b-e62b974262b3.png)

I don´t see any error log on the agent pod.

Probably I am doing a very basic mistake, but can´t figure what.

Here is my deployment manifest:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: grafana-agent
  namespace: monitoring
spec:
  minReadySeconds: 10
  replicas: 1
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      name: grafana-agent
  template:
    metadata:
      labels:
        name: grafana-agent
    spec:
      containers:
        - args:
            - -config.file=/etc/agent/agent.yaml
            - -config.expand-env=true
          command:
            - /bin/agent
          image: grafana/agent:v0.16.1
          imagePullPolicy: IfNotPresent
          name: agent
          ports:
            - containerPort: 12345
              name: http-metrics
          volumeMounts:
            - mountPath: /etc/agent
              name: grafana-agent
          resources:
            limits:
              cpu: 200m
              memory: 256Mi
      serviceAccount: grafana-agent
      volumes:
        - configMap:
            name: grafana-agent
          name: grafana-agent
```

And the agent config (without any envrionment variables interpolation yet:

```yaml
server:
  http_listen_port: 12345
prometheus:
  wal_directory: /tmp/grafana-agent-wal
  global:
    scrape_interval: 15s
    external_labels:
      cluster: cloud
  configs:
    - name: integrations
      remote_write:
        - url: https://prometheus-us-central1.grafana.net/api/prom/push
          basic_auth:
            username: <redacted>
            password: <redacted>
      scrape_configs:
        - job_name: integrations/go
          static_configs:
            - targets: ["localhost:8080"]
        - job_name: integrations/kubernetes/cadvisor
[...]

PR #707, opened 2021-07-02
Update config

justinwalz:
#### PR Description 

This PR fixes an issue that users of config.expand-env face.

This doesn't solve the issue, but is a workaround for users of this jsonnet. https://github.com/grafana/agent/issues/691

#### Notes to the Reviewer

I can update the CHANGELOG and documentation if this change is something that would be accepted. 

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/production/kubernetes/agent-sigv4.yaml
+++ b/production/kubernetes/agent-sigv4.yaml
@@ -133,7 +133,7 @@ data:
                   - replacement: https
                     target_label: __scheme__
                   - regex: (.+)
-                    replacement: /api/v1/nodes/${1}/proxy/metrics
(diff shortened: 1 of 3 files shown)


rfratto (commented):
LGTM, thanks for fixing this! Can you run `make example-kubernetes` so the example YAML manifests get updated? If you're having trouble running it locally, I can try pushing to your PR.

justinwalz:
@rfratto Good call, I just updated with commit with those changes as well.

rfratto (approved):
Thanks for fixing this!

PR #709, opened 2021-07-03
Add post method for config reload api

### LINKED ISSUE #610: Feature parity - requesting config reload API on Grafana Agent to match with Prometheus reload API
It would be nice to have the config reload API on Grafana Agent match with Prometheus reload API, for ease of use / operability 
 
Grafana Agent:
`GET /-/reload` 

Prometheus:
`PUT  /-/reload`
`POST /-/reload`

tharun208:
#### PR Description 
Add `POST` method for `reload` config http endpoint.

Fixes #610

#### PR Checklist

- [ ] CHANGELOG updated 
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/cmd/agent/entrypoint.go
+++ b/cmd/agent/entrypoint.go
@@ -63,7 +63,7 @@ func NewEntrypoint(logger *util.Logger, cfg *config.Config, reloader Reloader) (
 		}
 
 		reloadMux := mux.NewRouter()
-		reloadMux.HandleFunc("/-/reload", ep.reloadHandler)
+		reloadMux.HandleFunc("/-/reload", ep.reloadHandler).Methods("GET", "POST")
 		ep.reloadServer = &http.Server{Handler: reloadMux}
 	}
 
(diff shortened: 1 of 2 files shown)


tharun208:
also, I found that the `wal_directory` config for `Prometheus is missing in this file, can I check the config value for it in this PR?

rfratto (approved):
Perfect!

> also, I found that the wal_directory config for `Prometheus is missing in this file, can I check the config value for it in this PR?

I'm not sure I understand what you mean here, could you explain?

tharun208:
> Perfect!
> 
> > also, I found that the wal_directory config for `Prometheus is missing in this file, can I check the config value for it in this PR?
> 
> I'm not sure I understand what you mean here, could you explain?

This https://github.com/grafana/agent/blob/main/cmd/agent/agent-local-config.yaml is missing the `wal_directory` value which failed when ran the locally built agent with the config file.
So, can I include the change with this same pr?

rfratto:
Ah, thanks for explaining. The intent for that file was that users pass `--prometheus.wal-directory` as a flag since everyone will likely want to put it in a different place. I'd be fine with a change that configures `wal_directory` to a relative directory like `./data`, though.

tharun208:
> Ah, thanks for explaining. The intent for that file was that users pass `--prometheus.wal-directory` as a flag since everyone will likely want to put it in a different place. I'd be fine with a change that configures `wal_directory` to a relative directory like `./data`, though.

ok, understood, then it is not required to include the change.

rfratto (approved):

PR #726, opened 2021-07-08
Docs: change operator custom scrape jobs example to use integrations job_name

rfratto:


rgeyer (approved):
LGTM. Thanks for the quick turnaround! :)

As an aside, are there instructions for adding the `cluster` external label found elsewhere?

rfratto:
> As an aside, are there instructions for adding the cluster external label found elsewhere?
The Operator should be adding a cluster label automatically and matching it to `<namespace>/<name>` of the GrafanaAgent resource 🤔 If it's not, that's a bug.

docs/operator/add-custom-scrape-jobs.md
```diff
@@ -121,3 +121,9 @@ spec:
 ```
 
 The Secret **MUST** be in the same namespace as the PrometheusInstance.
+
+There is a known [issue](https://github.com/grafana/agent/issues/655) that
+currently prevents the Grafana Agent Operator from updating Grafana Agent
+deployments when `additionalScrapeConfigs` or the underlying secret changes.
+Until the issue is resolved, you should restart the Operator to force it to pick
```
rgeyer, line 128: Does it make sense to add an appropriate command? I.E. `kubectl rollout restart deployment agent-operator`, or similar?

rfratto: Good point. It's tricky because we don't prescribe the name of the deployment, but we can assume they followed the getting started guide and use that. Updating

rfratto: PTAL :)

Issue #722, opened 2021-07-07
Change subsystem names to remove confusion


rfratto:
Naming the subsystems of the Agent `prometheus`, `loki`, and `tempo` was well-intentioned but mostly confusing. We should rename them as following:

* Change `prometheus` to `metrics`
* Change `loki` to `logs` 
* Change `tempo` to `traces`

For backwards compatibility, both names should be supported in the config file, but the old names should be marked as deprecated so we can remove them in a future release.

rfratto:
This change should also be made in the Grafana Agent Operator, with the same backwards compatibility (ie support both PrometheusInstance and MetricsInstance).

PR #727, opened 2021-07-08
Rename "loki" to "logs"

rfratto:
#### PR Description 
Renames the `loki` subsystem to `logs`, deprecating the old name. `loki`-named fields will continue to work until a further release that fully removes them.

Related to #722.

#### Notes to the Reviewer
@mapno does `logs_tag` make sense for the overrides in automatic_logging? Should it be `logs_instance_tag`? I'm not 100% sure what it does. 

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
 - [DEPRECATION] The `loki` key at the root of the config file has been
   deprecated in favor of `logs`. `loki`-named fields in `automatic_logging`
   have been renamed accordinly: `loki_name` is now `logs_instance_name`,
-  `loki_tag` is now `logs_tag`, and `backend: loki` is now
+  `loki_tag` is now `logs_instance_tag`, and `backend: loki` is now
   `backend: logs_instance`.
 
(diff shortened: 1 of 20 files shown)


mattdurham (approved):

mapno:
> @mapno does logs_tag make sense for the overrides in automatic_logging? Should it be logs_instance_tag? I'm not 100% sure what it does.

Yeah, `logs_instance_tag` would be more precise, since it only applies to ~Loki~ the logs instance. What it does is it overrides the label key it uses when logging to loki (i.e. `{"tempo": "root"}` <- it's the `tempo` key)

rfratto:
> Yeah, logs_instance_tag would be more precise, since it only applies to Loki the logs instance. What it does is it overrides the label key it uses when logging to loki (i.e. {"tempo": "root"} <- it's the tempo key)

Makes sense, thanks! PTAL :)

mapno (approved):
LGTM!

PR #728, opened 2021-07-08
Set DNSPolicy correctly when host network is used in daemonset

rgeyer:
#### PR Description 

When an integration is applied using the tanka library, it sets the daemonset or deployment to use host networking, but it does not change the DNS policy.

In my environment, this meant it could not write to my local cortex at `cortex.cortex.svc.cluster.local` because it couldn't resolve it.

N/A

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated


rfratto (approved):
LGTM, thanks!

PR #732, opened 2021-07-12
Implement Drone Build system, moving away from release via Github Actions

mattdurham:
#### PR Description 

# Drone overview

## Why replace github actions?

Github actions was failing due to constraints and not building reliably

## Why drone?

The drone build is meant to replace github actions for releases, github actions will continue to have a place in dependency checking and other purposes. Drone was chosen because grafana is already a user. That being said the drone build should run on any vanilla drone instance.

## Goals

Simplify the build process and make it easier to build.

## Changes made

Several shell scripts were removed and integrated into the makefile to reduce complexity and mental switches when dealing. Also much of the build output was suppressed and now is revealed for help while debugging various issues. 

## How to test?

The easiest way is to download drone and use the local exec. `drone exec --trusted --pipeline <name> ./.drone/drone.yaml`, you will need docker installed. The trusted is because it uses docker within docker and needs the privileged execution. 

#### PR Checklist

- [X] CHANGELOG updated 
- [X] Documentation added
- [NA] Tests updated


GOVERNANCE.md
```diff
@@ -1,156 +0,0 @@
----
```
rfratto, line 1: Did you mean to delete GOVERNANCE.md?

rfratto (commented):
Thanks for working on this! I'm a little concerned with how this changes the role of the Makefile. 

The Makefile took inspiration from Loki and Cortex where it's intended to be the entrypoint for all developers (at least on a Unix system) that are working on the Agent. All `make` commands are intended to work on any system with minimal extra dependencies needed. This meant that CI then only automates the commands a developer would run normally, with the exception of some specialized jobs like the dependency checker cronjob. 

Assuming that the Makefile is being invoked from Drone changes that behavior, and I think we either need to avoid changing the behavior or discuss a new approach for local development. For example, `make agent-image` will always try to push images  now, and `make dist` won't work out of the box at all on macOS anymore, since it looks for locally installed cross-compilers (and would probably still fail even then since it looks for the systemd headers when cross-compiling for Linux).  

I am open to discussing changing the developer workflow, but if this is what we choose to do, we'll need to put effort into making it as consistent as possible so people don't get confused. The goal should be that if I'm working on the Agent for platform X, I should always interact with the same tool to do my work (e.g., not `make Y` for some jobs and `drone` for others).

mattdurham:
The intent is to get everything back to supporting both drone and the previous functionality, I didn't think you would review the draft =) The go.mod changes will get rolled back once I merge the otel replace changes in. I had to update them to get the OTEL changes to build on arm v6/v6 (32 bit platforms). I should have everything behaving correctly tomorrow and will flip the PR to ready.

tools/test-packages
```diff
@@ -40,5 +40,5 @@ function test_with_systemd() {
     docker exec -i "${container}" /bin/bash -c "${remove_command}"
 }
 
-test_with_systemd "${IMAGE_PREFIX}"/centos-systemd "rpm -i /dist/grafana-agent-${VERSION}-${RELEASE}.amd64.rpm" "rpm -e grafana-agent"
-test_with_systemd "${IMAGE_PREFIX}"/debian-systemd "dpkg -i /dist/grafana-agent-${VERSION}-${RELEASE}.amd64.deb" "dpkg -r grafana-agent"
+test_with_systemd "${IMAGE_PREFIX}"/centos-systemd "rpm -i /dist/grafana-agent-${VERSION}-${RELEASE}.x86_64.rpm" "rpm -e grafana-agent"
+test_with_systemd "${IMAGE_PREFIX}"/debian-systemd "dpkg -i /dist/grafana-agent-${VERSION}-${RELEASE}.x86_64.deb" "dpkg -r grafana-agent"
```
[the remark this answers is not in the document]
mattdurham: Done!

GOVERNANCE.md
mattdurham: Not at all and its back

Makefile
```diff
@@ -119,27 +113,36 @@ all: protos agent agentctl
...
+	export CGO_ENABLED=1 GOOS=linux GOARCH=amd64 GO111MODULE=auto CC=gcc CCX=g++ ; go build $(CGO_FLAGS) -o $@ ./$(@D)
+endif
+ifeq ($(TARGETPLATFORM),linux/arm64)
+	export CGO_ENABLED=1 GOOS=linux GOARCH=arm64 CC=aarch64-linux-gnu-gcc CCX=aarch64-linux-gnu-g++; go build $(CGO_FLAGS) -o $@ ./$(@D)
+endif
+ifeq ($(TARGETPLATFORM),linux/arm/v7)
+	export CGO_ENABLED=1 GOOS=linux GOARCH=arm GOARM=7 CC=arm-linux-gnueabi-gcc CCX=arm-linux-gnueabi-g++; go build $(CGO_FLAGS) -o $@ ./$(@D)
+endif
+ifeq ($(TARGETPLATFORM),linux/arm/v6)
+	export CGO_ENABLED=1 GOOS=linux GOARCH=arm GOARM=6 CC=arm-linux-gnueabi-gcc CCX=arm-linux-gnueabi-g++; go build  $(CGO_FLAGS) -o $@ ./$(@D)
```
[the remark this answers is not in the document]
mattdurham: This is cleaner, but still not perfect.

go.mod
```diff
@@ -41,15 +42,18 @@ require (
...
 	github.com/wrouesnel/postgres_exporter v0.0.0-00010101000000-000000000000
 	go.opencensus.io v0.23.0
 	go.opentelemetry.io/collector v0.21.0
 	go.uber.org/atomic v1.7.0
 	go.uber.org/zap v1.16.0
-	golang.org/x/sys v0.0.0-20210503173754-0981d6026fa6
+	golang.org/x/oauth2 v0.0.0-20210628180205-a41e5a781914 // indirect
+	golang.org/x/sys v0.0.0-20210630005230-0f9fa26af87c
 	google.golang.org/grpc v1.37.0
+	google.golang.org/protobuf v1.27.1 // indirect
```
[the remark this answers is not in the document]
mattdurham: Merged from the otel changes.

Makefile
```diff
@@ -193,61 +197,50 @@ dist: dist-agent dist-agentctl dist-packages
...
+dist/agent-linux-armv7: 
+	export CGO_ENABLED=1 GOOS=linux GOARCH=arm GOARM=7  CC=arm-linux-gnueabi-gcc CCX=arm-linux-gnueabi-g++; go build $(CGO_FLAGS) -o $@ ./cmd/agent
+dist/agent-linux-mipsle: 
+	export CGO_ENABLED=1 GOOS=linux GOARCH=mipsle GOMIPS=softfloat CC=mipsel-linux-gnu-gcc CCX=mipsel-linux-gnu-f++;go build $(CGO_FLAGS) -o $@ ./cmd/agent
+dist/agent-darwin-amd64: 
+	export CGO_ENABLED=1 GOOS=darwin GOARCH=amd64 CC=x86_64-apple-darwin20.2-clang CCX=x86_64-apple-darwin20.2-clang++ LD_LIBRARY_PATH=$(OSXCROSS_PATH)/lib; go build $(CGO_FLAGS) -o $@ ./cmd/agent
+dist/agent-darwin-arm64: 
+	export CGO_ENABLED=1 GOOS=darwin GOARCH=arm64 CC=arm64-apple-darwin20.2-clang CCX=arm64-apple-darwin20.2-clang++ LD_LIBRARY_PATH=$(OSXCROSS_PATH)/lib; go build $(CGO_FLAGS) -o $@ ./cmd/agent
+dist/agent-windows-amd64.exe: 
+	export CGO_ENABLED=1 GOOS=windows GOARCH=amd64 CC=x86_64-w64-mingw32-gcc CCX=x86_64-w64-mingw32-g++; go build $(CGO_FLAGS) -o $@ ./cmd/agent
```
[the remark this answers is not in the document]
mattdurham: Verified this works =)

.drone/go_wrapper.sh
```diff
@@ -0,0 +1,161 @@
+#!/usr/bin/env bash
```
rfratto, line 1: Do we need to copy this from seego into this repository? I'm worried about keeping them in sync.

.github/workflows/test.yml
```diff
@@ -1,54 +0,0 @@
-name: Test
```
rfratto, line 1: Can you keep this file but only have it do `test` for `macos-latest` and `windows-latest`? (ie remove lint, test-packages)

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
   the matching release version of the Grafana Agent instead of v0.14.0.
   (@rfratto)
 
+- [ENHANCEMENT] Switch to drone for releases. (@mattdurham)
```
rfratto, line 11: nit: I try to keep the CHANGELOG in type -> date order, can you add it as the last enhancement instead of the first?

Makefile
```diff
@@ -75,17 +96,20 @@ PACKAGE_RELEASE := 1
...
-
 docker-build = docker buildx build --push --platform linux/amd64,linux/arm64,linux/arm/v6,linux/arm/v7 $(DOCKER_BUILD_FLAGS)
 endif
 
-ifeq ($(BUILD_IN_CONTAINER),false)
-seego = "/seego.sh"
+# If being called from the drone executable this will be set to true, in this case we want the go_wrapper to set the
+#	CC and CCX and other flags for the appropriate architecture
+ifeq ($(DRONE),true)
+	seego = bash ./.drone/go_wrapper.sh
```
rfratto, line 112: Shouldn't `/seego.sh` (or `/go_wrapper.sh`) still exist from Drone since it's running from the seego container?

Makefile
```diff
@@ -138,39 +162,57 @@ agent: cmd/agent/agent
...
+
+cmd/agent/agent: check-seego  cmd/agent/main.go
+ifeq ($(DRONE),true)
+	$(CGO_CC_CCX); go build $(CGO_FLAGS) -o $@ ./$(@D)
+else
 ifeq ($(CROSS_BUILD),false)
 	CGO_ENABLED=1 go build $(CGO_FLAGS) -o $@ ./$(@D)
 else
-	@CGO_ENABLED=1 GOOS=$(GOOS) GOARCH=$(GOARCH) GOARM=$(GOARM); $(seego) build $(CGO_FLAGS) -o $@ ./$(@D)
+	CGO_ENABLED=1 GOOS=$(GOOS) GOARCH=$(GOARCH) GOARM=$(GOARM); $(seego) build $(CGO_FLAGS) -o $@ ./$(@D)
```
rfratto, line 177: Can we leave this as-is? the extra output line is noisy

Makefile
```diff
@@ -138,39 +162,57 @@ agent: cmd/agent/agent
...
-cmd/agent/agent: check-seego cmd/agent/main.go
+
+
+
+# In general DRONE variable should overwrite any other options, if DRONE is not set then fallback to normal behavior
+
+cmd/agent/agent: check-seego  cmd/agent/main.go
+ifeq ($(DRONE),true)
+	$(CGO_CC_CCX); go build $(CGO_FLAGS) -o $@ ./$(@D)
+else
```
rfratto, line 173: Where does this get called from? IIUC nothing from Drone should ever make `cmd/agent/agent` directly, right?

Makefile
```diff
@@ -138,39 +162,57 @@ agent: cmd/agent/agent
...
-	@CGO_ENABLED=1 GOOS=$(GOOS) GOARCH=$(GOARCH) GOARM=$(GOARM); $(seego) build $(CGO_FLAGS) -o $@ ./$(@D)
+	CGO_ENABLED=1 GOOS=$(GOOS) GOARCH=$(GOARCH) GOARM=$(GOARM); $(seego) build $(CGO_FLAGS) -o $@ ./$(@D)
+endif
 endif
 	$(NETGO_CHECK)
 
+
 cmd/agentctl/agentctl: check-seego cmd/agentctl/main.go
+ifeq ($(DRONE),true)
+	$(CGO_CC_CCX); go build $(CGO_FLAGS) -o $@ ./$(@D)
```
rfratto, line 185: This can use the seego entrypoint script instead of having to set CGO_CC_CXX here.

Makefile
```diff
@@ -138,39 +162,57 @@ agent: cmd/agent/agent
...
+else
 ifeq ($(CROSS_BUILD),false)
 	CGO_ENABLED=1 go build $(CGO_FLAGS) -o $@ ./$(@D)
 else
-	@CGO_ENABLED=1 GOOS=$(GOOS) GOARCH=$(GOARCH) GOARM=$(GOARM); $(seego) build $(CGO_FLAGS) -o $@ ./$(@D)
+	CGO_ENABLED=1 GOOS=$(GOOS) GOARCH=$(GOARCH) GOARM=$(GOARM); $(seego) build $(CGO_FLAGS) -o $@ ./$(@D)
```
rfratto, line 190: Can we leave this as-is?

Makefile
```diff
@@ -138,39 +162,57 @@ agent: cmd/agent/agent
...
 endif
-	$(NETGO_CHECK)
```
rfratto, line 155: Is there a reason we needed to take this out?

Makefile
```diff
@@ -223,61 +265,140 @@ dist: dist-agent dist-agentctl dist-packages
...
 
+####################
+# BEGIN AGENT DIST #
+####################
+
 dist-agent: seego dist/agent-linux-amd64 dist/agent-linux-arm64 dist/agent-linux-armv6 dist/agent-linux-armv7 dist/agent-darwin-amd64 dist/agent-darwin-arm64 dist/agent-windows-amd64.exe dist/agent-freebsd-amd64 dist/agent-windows-installer.exe
+
 dist/agent-linux-amd64: seego
-	@CGO_ENABLED=1 GOOS=linux GOARCH=amd64; $(seego) build $(CGO_FLAGS) -o $@ ./cmd/agent
+	export CGO_ENABLED=1 GOOS=linux GOARCH=amd64;        $(seego) build $(CGO_FLAGS)  -o $@ ./cmd/agent
```
rfratto, line 275: `export` won't be necessary if the entrypoint was `/seego.sh` and not `bash <script>`

tools/seego/Dockerfile
```diff
@@ -9,3 +9,5 @@ RUN  rm -rf /usr/local/go                                           \
   && echo "$GOLANG_DOWNLOAD_SHA256  golang.tar.gz" | sha256sum -c - \
   && tar -C /usr/local -xzf golang.tar.gz                           \
   && rm golang.tar.gz
+RUN apt-get update
+RUN apt-get install -y nsis
```
rfratto, line 13: I don't think this Dockerfile is being used anywhere now, we can take it out

rfratto (commented):
Please don't rename the GOVERNANCE.md file :) 

Looks good for a first pass, but a lot of lines changed so hard to give it a thorough review. Once we slim it down a little I'll give it a second pass

.drone/go_wrapper.sh
mattdurham: Nope, your correct, I can reference the built in one.

Makefile
mattdurham: Sure, I would not like to change the drone or cross-builds though since they are more complex/harder to debug.

Makefile
mattdurham: Since we already have to define the GOOS, GOARCH, GOARM  including the CC CCX ect seems easier than calling out to an external script. Less context switching than going to the seego repo and looking at the go_wrapper.sh there. Another alternative would be making seego/go_wrapper aware of TARGETPLATFORM. Though I think that feels like even more indirection.

tools/seego/Dockerfile
mattdurham: Still used by cross-build local builds

Makefile
rfratto: > Another alternative would be making seego/go_wrapper aware of TARGETPLATFORM. 

Isn't that what https://github.com/grafana/agent/blob/main/tools/cross_build.bash does? 

> Since we already have to define the GOOS, GOARCH, GOARM including the CC CCX ect seems easier than calling out to an external script.

🤷‍♂️ seego is designed to be an abstraction over the go binary, and there's no guarantee that the same CC/CXX/etc values will be used between seego releases. I really think we should just rely on having seego figure it out.

mattdurham:
Simplified the makefile, a function at the top drives setting all the CC, CCX and other flags for builds that don't call the seego within the makefile. Drone builds are built inside seego. Tested on osx/linux, tests pass in github actions.

Makefile
mattdurham: I personally like seeing all the output on builds, makes it easier to debug. For someone who hasn't worked in make in a while or hasn't, it is hard to know what is wrong when you don't realize the output isn't being written out. (Like I was initially). It does add considerable noise though.

.github/workflows/test.yml
mattdurham: Done

.drone/drone.yml
```diff
@@ -0,0 +1,142 @@
...
+steps:
+- name: distribute
+  image: rfratto/seego
+  commands:
+  - apt-get update &&  apt-get install -y rubygems rpm nsis
+  - gem install --no-document fpm
+  - mkdir -p /usr/local/go/bin
+  - wget -q https://golang.org/dl/go1.16.5.linux-amd64.tar.gz
+  - tar -C /usr/local -xzf go1.16.5.linux-amd64.tar.gz
+  - rm go1.16.5.linux-amd64.tar.gz
```
rfratto, line 31: nit: since this isn't a Dockerfile we probably don't have to remove the tarball, though it doesn't really matter to me if we leave this in or not.

.drone/drone.yml
```diff
@@ -0,0 +1,142 @@
...
+- name: distribute
+  image: rfratto/seego
+  commands:
+  - apt-get update &&  apt-get install -y rubygems rpm nsis  apt-transport-https ca-certificates curl gnupg lsb-release
+  - gem install --no-document fpm
+  - mkdir -p /usr/local/go/bin
+  - wget -q https://golang.org/dl/go1.16.5.linux-amd64.tar.gz
+  - tar -C /usr/local -xzf go1.16.5.linux-amd64.tar.gz
+  - rm go1.16.5.linux-amd64.tar.gz
+  - export PATH=$PATH:/usr/local/go/bin
```
rfratto, line 52: Is there a way we can dedupe some of this stuff? Does Drone support Yaml anchors? If there's no easy/sane way to dedupe, I'm fine with leaving it as-is for now and then generating the drone.yml later on. (Generating with Jsonnet is what Loki did when moving to Drone).

.github/workflows/test.yml
```diff
@@ -1,54 +0,0 @@
...
-on: [pull_request]
-jobs:
```
rfratto, line 3: I think every other file in this directory ends in .yml, right? Can we keep it consistent?

.drone/drone.yml
```diff
@@ -0,0 +1,142 @@
...
+name: Containerize
+platform:
+  os: linux
+  arch: amd64
+trigger:
+  event:
+  - tag
+
+steps:
+- name: Build Containers
```
rfratto, line 111: Does this always run? Ideally we'd want to prevent Containerize from running if Lint/tests fail

Makefile
```diff
@@ -8,6 +8,34 @@ SHELL = /usr/bin/env bash
...
+$(if $(filter $(1),linux/arm/v6),export CGO_ENABLED=1 GOOS=linux GOARCH=arm GOARM=6 CC=arm-linux-gnueabi-gcc CCX=arm-linux-gnueabi-g++, \
+$(if $(filter $(1),darwin/amd64),export CGO_ENABLED=1 GOOS=darwin  GOARCH=amd64 CC=x86_64-apple-darwin20.2-clang CCX=x86_64-apple-darwin20.2-clang++ LD_LIBRARY_PATH=$(OSXCROSS_PATH)/lib, \
+$(if $(filter $(1),darwin/arm64),export CGO_ENABLED=1 GOOS=darwin GOARCH=arm64 CC=arm64-apple-darwin20.2-clang CCX=arm64-apple-darwin20.2-clang++ LD_LIBRARY_PATH=$(OSXCROSS_PATH)/lib, \
+$(if $(filter $(1),windows),export CGO_ENABLED=1 GOOS=windows GOARCH=amd64 CC=x86_64-w64-mingw32-gcc CCX=x86_64-w64-mingw32-g++, \
+$(if $(filter $(1),mipls),export CGO_ENABLED=1 GOOS=linux GOARCH=mipsle GOMIPS=softfloat CC=mipsel-linux-gnu-gcc CCX=mipsel-linux-gnu-f++, \
+$(if $(filter $(1),freebsd),export CGO_ENABLED=1 GOOS=freebsd GOARCH=amd64 CC=clang CCX=clang++ CGO_CFLAGS="-target x86_64-pc-freebsd11 --sysroot=/usr/freebsd/x86_64-pc-freebsd11" CGO_CXX_CFLAGS="-target x86_64-pc-freebsd11 --sysroot=/usr/freebsd/x86_64-pc-freebsd11" CGO_LDFLAGS="-target x86_64-pc-freebsd11 --sysroot=/usr/freebsd/x86_64-pc-freebsd11"  , $(error invalid flag $(1))) \
+)))))))))
+endef
```
rfratto, line 34: As discussed offline, this still makes me nervous, especially since we use `:latest` of seego (which needs to change IMO but that's my fault and a separate issue). I think for now we should keep the setting of CC/CXX/CGO_FLAGS/etc as the responsibility of the seego entrypoint.

Makefile
```diff
@@ -75,17 +102,19 @@ PACKAGE_RELEASE := 1
...
 
+# seego is used by default when running bare make commands such as `make dist` this uses an image that has all the necessary libraries to cross build
+#	when using drone the docker in docker is more problematic so instead drone uses seego has the base image then calls make running "raw" commands
+seego = docker run --rm -t -v "$(CURDIR):$(CURDIR)" -w "$(CURDIR)" -e "CGO_ENABLED=$$CGO_ENABLED" -e "GOOS=$$GOOS" -e "GOARCH=$$GOARCH" -e "GOARM=$$GOARM" -e "GOMIPS=$$GOMIPS"  grafana/agent/seego
+docker-build = docker build $(DOCKER_BUILD_FLAGS)
 ifeq ($(CROSS_BUILD),true)
 DOCKERFILE = Dockerfile.buildx
-
 docker-build = docker buildx build --push --platform linux/amd64,linux/arm64,linux/arm/v6,linux/arm/v7 $(DOCKER_BUILD_FLAGS)
+# docker-build = docker buildx build --push --platform linux/arm64 $(DOCKER_BUILD_FLAGS)
```
rfratto, line 113: you left this in by mistake :)

Makefile
```diff
@@ -138,39 +167,34 @@ agent: cmd/agent/agent
...
-else
-	@CGO_ENABLED=1 GOOS=$(GOOS) GOARCH=$(GOARCH) GOARM=$(GOARM); $(seego) build $(CGO_FLAGS) -o $@ ./$(@D)
-endif
+
+
+
+# In general DRONE variable should overwrite any other options, if DRONE is not set then fallback to normal behavior
+
+cmd/agent/agent: check-seego  cmd/agent/main.go
+	$(ALL_CGO_BUILD_FLAGS) ; $(seego) build $(CGO_FLAGS) -o $@ ./$(@D)
```
rfratto, line 176: Nice, having `$(seego)` == `go`  when not calling out to a container is a neat idea and really simplifies the Makefile.

Makefile
```diff
@@ -223,61 +247,100 @@ dist: dist-agent dist-agentctl dist-packages
...
-	@CGO_ENABLED=1 GOOS=windows GOARCH=amd64; $(seego) build $(CGO_FLAGS) -o $@ ./cmd/agent
+	$(call SetBuildVarsConditional,windows) ;      $(seego) build $(CGO_FLAGS) -o $@ ./cmd/agent
+
 dist/agent-windows-installer.exe: dist/agent-windows-amd64.exe
-	cp dist/agent-windows-amd64.exe ./packaging/windows
-	cp LICENSE ./packaging/windows
-	docker build ./packaging/windows -t windows_nsis
-	docker run -e VERSION=${RELEASE_TAG} --rm -t -v $(CURDIR)/dist:/app windows_nsis
+ifeq ($(DRONE),true)
+	cp dist/agent-windows-amd64.exe ./packaging/windows && cp LICENSE ./packaging/windows && makensis -V4 -DVERSION=${RELEASE_TAG} -DOUT="../../dist/grafana-agent-installer.exe" ./packaging/windows/install_script.nsis
```
rfratto, line 281: I noticed the DOUT path is different from the else on line 289, can you help me understand what's going on there?

rfratto (commented):
Nice, this is getting a lot closer. Thanks for working on this.

.drone/drone.yml
mattdurham: Its really annoying with drone =) If you run multiple drone instances it created a ton of .gz files that sit around.

rfratto: Ah, gotcha 👍

.drone/drone.yml
mattdurham: I tried using the anchors but could not get them to be happy. From light reading on drone issues, they have changed the behavior a few times.

rfratto: Alright, no worries, a config language can clean this up down the line.

Makefile
mattdurham: As previously discussed rolled back to using the bash file or default go build.

Makefile
mattdurham: Fixed this and tested on windows native and dist/containers.

mattdurham:
Simplified windows build, and cleaned up various comments.

.drone/drone.yml
```diff
@@ -0,0 +1,152 @@
...
+---
+kind: pipeline
+type: docker
+name: Containerize
+platform:
+  os: linux
+  arch: amd64
+trigger:
+  event:
+  - tag
```
rfratto, line 69: We still want containers per main-branch commit (Tempo team relies on them). I think it boils down to:

Main branch commit:
- Test 
- Containerize (tag being `main-sha`)

Tag:
- Test 
- Containerize (tag being `vX.Y.Z`)
- Dist 
- Publish

Makefile
```diff
@@ -223,61 +248,98 @@ dist: dist-agent dist-agentctl dist-packages
...
 
 seego: tools/seego/Dockerfile
+ifeq ($(DRONE),false)
+ifeq ($(BUILD_IN_CONTAINER),true)
 	docker build -t grafana/agent/seego tools/seego
+endif
+endif
 
 # Makes seego if CROSS_BUILD is true.
 check-seego:
```
rfratto, line 335: This isn't working if you just run `make agent` directly

rfratto (approved):
LGTM, great work. Thanks for putting time into this.

PR #721, opened 2021-07-07
Operator: create CRDs for logging

rfratto:
#### PR Description 
Adds logging-specific CRDs including a `PodLogs` CRD that is intended to discover pods to collect logs from. 

#### Notes to the Reviewer
Not hooked in anywhere yet, just doing this PR separately since it's a huge change. 

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

Issue #628, opened 2021-06-03
Proposal: logging support in the Grafana Agent Operator


rfratto:
I've created a Google doc describing how we can support logs within the Grafana Agent Operator. Feedback welcome!

Design doc: https://docs.google.com/document/d/1nBR4NQhx0lTBDmVsXAY4ehAd8O_BXgW6R89MgXc1fuo/edit#

PR #740, opened 2021-07-14
Grafana Agent Operator: generate logs config

rfratto:
#### PR Description 
Generates a Grafana Agent config based on the logging CRDs. 

Related to #721. 
Related to #628. 

#### Notes to the Reviewer
This is 90% tests and most of the logic in all in Jsonnet. I'd like a less verbose way to test this stuff but 🤷‍♂️ 

_Still_ not hooked in anywhere. Next PR will actually create a DaemonSet and call this code. 

Also, turns out `enforcedNamespaceLabel` was never a boolean and was always a string. Oh, the issues with manually-managed types 🙃. 

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated


pkg/operator/apis/monitoring/v1alpha1/types_logs.go
```diff
@@ -234,7 +243,7 @@ type PipelineStageSpec struct {
 	// the Grafana Agent pod. The Grafana Agent Operator should be configured
 	// with a MetricsInstance that discovers the logging DaemonSet to collect
 	// metrics created by this stage.
-	Metrics map[string]*MetricsStageSpec `json:"metrics,omitempty"`
+	Metrics map[string]MetricsStageSpec `json:"metrics,omitempty"`
```
rfratto, line 246: It's invalid for the value to be nil here so we're forcing it to at least be the zero value.

pkg/operator/apis/monitoring/v1alpha1/types_logs.go
```diff
@@ -412,9 +421,9 @@ type MetricsStageSpec struct {
 	// to a positive float.
 	Action string `json:"action,omitempty"`
 
-	// Buckets to create.
+	// Buckets to create. Bucket values must be convertible to floats.
 	// Only valid for type: histogram.
-	Buckets []int `json:"bucket,omitempty"`
+	Buckets []string `json:"bucket,omitempty"`
```
rfratto, line 426: Buckets are supposed to be floats, but Kubernetes doesn't support floats, so....strings it is.

pkg/operator/config/config.go
```diff
@@ -91,7 +113,7 @@ func createVM(secrets assets.SecretStore) (*jsonnet.VM, error) {
 		return nil, err
 	}
 
-	vm.Importer(NewFSImporter(templatesContents))
+	vm.Importer(NewFSImporter(templatesContents, []string{"./"}))
```
rfratto, line 116: This allows us to stop doing relative imports for everything.

pkg/operator/config/fs_importer.go
```diff
@@ -15,30 +16,45 @@ type FSImporter struct {
...
 }
 
 // Import implements jsonnet.Importer.
 func (i *FSImporter) Import(importedFrom, importedPath string) (contents jsonnet.Contents, foundAt string, err error) {
-	cleanedPath := path.Clean(
-		path.Join(path.Dir(importedFrom), importedPath),
-	)
-	cleanedPath = strings.TrimPrefix(cleanedPath, "./")
 
+	tryPaths := append([]string{importedFrom}, i.paths...)
```
rfratto, line 34: This will default to checking files relative to the current directory and then relative to each path in `i.paths`.

pkg/operator/apis/monitoring/v1alpha1/types_logs.go
mattdurham: Wut? That's intriguing.

mattdurham: Does this matter if its float 32/64?

mattdurham (approved):
Looks good, one small comment about if we need to specify 32/63 on the floats.

pkg/operator/apis/monitoring/v1alpha1/types_logs.go
rfratto: Nope, no floats period: https://github.com/kubernetes-sigs/controller-tools/blob/abcf7d8d54b0cca0f42c470b705fd3cfe768ede3/pkg/crd/schema.go#L387

rfratto: Oh, I'm not sure if we need to specify float32/float64 (if your buckets are that small you're probably going to have other problems), but I'm pretty sure these will get read as float64s.

rfratto: Looked into a it a little more and go-jsonnet stores everything as a float64. I'm not super confident in how accurate the string -> float64 conversion is going to be, so I'll add some kind of warning here about risk of precision loss at very large or very small numbers.

PR #743, opened 2021-07-15
Split trace agreggation

mapno:
#### PR Description

Split trace aggregation from tail sampling, as it can be useful for other purposes too, such as service graphs or other processors that benefit from having complete traces.

The pipeline will use `groupbytrace` processor for aggregation, except when using tail sampling, in which case grouping will be performed by the embedded aggregator in the sampling processor. It's not possible to have the tail sampling processor only sampling and leave aggregation to `groupbytrace`. Tail sampling has to buffer traces for at least 1s, so it's just better not to use `groupbytrace` in that case for simplicity and performance.

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/configuration/tempo-config.md
+++ b/docs/configuration/tempo-config.md
@@ -175,7 +175,7 @@ spanmetrics:
 #
 # In order to make a correct sampling decision it's important that the agent has
 # a complete trace. This is achieved by waiting a given time for all the spans
-# before evaluating the trace.
+# before evaluating the trace. In order to get complete traces, configure group_by_trace.
 #
 # Tail sampling also supports multiple agent deployments, allowing to group all
 # spans of a trace in the same agent by load balancing the spans by trace ID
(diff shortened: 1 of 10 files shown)


mapno:
There are a few things still missing from the PR, mostly documentation, so I marked it as a draft.

mapno:
It's ready for review now

docs/configuration/tempo-config.md
```diff
@@ -226,6 +221,24 @@ tail_sampling:
...
+# This is of particular interest for processor that benefit from having complete
+# traces, such as tail-based sampling
+#
+# A trace will be considered complete after the defined `wait` duration has
+# passed since its first span arrived to the processors.
+# If the trace is not complete by then, it'll be split into more than one trace.
+#
+# Longer waiting times will increase the number of traces that are correctly
+# grouped. However, it will also increase the memory overhead of the processor.
+group_by_trace:
```
joe-elliott, line 235: shouldn't the "load balancing" settings be moved out as well? maybe renamed something to suggest it's routing by trace id.

docs/configuration/tempo-config.md
```diff
@@ -226,6 +221,24 @@ tail_sampling:
...
+  
+  # Defines the time to wait for a complete trace before considering it complete
+  [ wait: <duration> | default="5s" ]
+    
+  # Configures the max amount of traces to keep in memory waiting for the duration
+  [ num_traces: <int> | default="1_000_000" ]
```
joe-elliott, line 241: max_traces?

PR #710, opened 2021-07-03
trigger reload configuration using SIGHUP

tharun208:
#### PR Description 
 This PR introduces reload of config using `SIGHUP`  signal. 

#### Which issue(s) this PR fixes 
Fixes #537.

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

PR #744, opened 2021-07-15
Use buffered channel for signal.Notify

rfratto:
The documentation for [`signal.Notify`](https://pkg.go.dev/os/signal#Notify) says:

> Package signal will not block sending to c: the caller must ensure that c has sufficient buffer space to keep up with the expected signal rate. For a channel used for notification of just one signal value, a buffer of size 1 is sufficient. 

This PR adds a buffer to the channel introduced in #710.


mapno (approved):
Nice catch

PR #752, opened 2021-07-19
Add install-bare.sh warning

rfratto:
#### PR Description 
Directs users away from the install-bare.sh script which is used only for Grafana Cloud and will eventually be removed in favor of Helm charts.

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/production/kubernetes/README.md
+++ b/production/kubernetes/README.md
@@ -26,6 +26,11 @@ applying out of the box and you will have to manually perform the following step
 
 3. Apply the modified manifest file: `kubectl -ndefault apply -f manifest.yaml`.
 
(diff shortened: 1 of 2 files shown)


hjet:
LGTM, the only thing I would add (which I should have done) is this:

> :warning: **These manifests do not include the Agent's configuration (ConfigMaps)**, which are necessary to run the Agent. For sample configurations and detailed installation instructions, ... 

In https://github.com/grafana/agent/blob/main/production/kubernetes/README.md

For more clarity... but not necessary

hjet (approved):

rfratto:
Thanks, that's a good idea, I'll add that in.

PR #754, opened 2021-07-20
Ignore regex capture groups when expanding environment vars

### LINKED ISSUE #692: Ignore patterns in the form ${number} when expanding env vars
`-config.expand-vars` will read anything in the form `${var}`, which includes pattern commonly used in Prometheus relabel_rules like `${1}`. This tends to be surprising as some scrape jobs will suddenly fail, and users must replace the relabel rule with `$${1}` or `$1`. 

`envsubst` supports a custom [`Eval`](https://pkg.go.dev/github.com/drone/envsubst#Eval) function that accepts a mapping function as an arugment. We should use `Eval` and give it a mapper that checks to see if the input string is a number and output `${number}` if it is. Otherwise, fall back to `os.Getenv` as normal.

Related: #691

rfratto:
#### PR Description 
Environment variable patterns from `-config.expand-env` (`${HOSTNAME}`) can look exactly like regex capture group references (`${1}`), causing the latter to get trimmed out from the file. This PR checks to see if the environment name is a number, and if it is, will keep the reference intact.

Closes #692.

#### Notes to the Reviewer
This is kind of a hacky fix. Windows technically supports numeric environment variables (IIRC), and also this will replace other functions like `${1=default}` with `${1}`. But it might be good enough for now, since the latter would only be valid on Windows and I suspect would be an extremely rare use case.

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@
   deprecated in favor of `logs`. `loki`-named fields in `automatic_logging`
   have been renamed accordinly: `loki_name` is now `logs_instance_name`,
   `loki_tag` is now `logs_instance_tag`, and `backend: loki` is now
-  `backend: logs_instance`. (@rfratto)[]
+  `backend: logs_instance`. (@rfratto)
 
 # v0.16.1 (2021-06-22)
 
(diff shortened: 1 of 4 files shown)


mapno (approved):
LGTM. One question regarding

> this will replace other functions like ${1=default} with ${1}

won't it be left as `${1=default}`?

mattdurham (approved):
Yes, the windows edge case seems incredibly unlikely.

PR #756, opened 2021-07-20
Service graphs processor

mapno:
#### PR Description

Opened to get early feedback if someone has any. It's a WIP and many things have been left out.

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated


pkg/tempo/servicegraphprocessor/processor.go
```diff
@@ -0,0 +1,227 @@
...
+func (p *processor) Capabilities() consumer.Capabilities {
+	return consumer.Capabilities{}
+}
+
+func (p *processor) ConsumeTraces(ctx context.Context, td pdata.Traces) error {
+	if p.closed.Load() {
+		return nil
+	}
+
+	if p.store.ItemCount() >= p.maxEdges {
```
annanay25, line 142: `maxEdges` feels like it relates to the total number of edges in the service graph, but here it is the number of items in the store (each unique span is an item)

pkg/tempo/servicegraphprocessor/factory.go
```diff
@@ -0,0 +1,53 @@
...
+	"go.opentelemetry.io/collector/consumer"
+	"go.opentelemetry.io/collector/processor/processorhelper"
+)
+
+const (
+	// TypeStr is the unique identifier for the Prometheus service graph exporter.
+	TypeStr = "service_graphs"
+
+	defaultWait     = time.Second * 10
+	defaultMaxEdges = 10_000
```
annanay25, line 18: I have some operational concerns of storing 10k edge objects in the store at any given time, even though our current scale of operation is much higher (something like 70k spans/sec through each agent pod?). We might end up either blowing up the agent memory _or_ having inaccurate graphs because we are unable to process all spans. Thoughts?

annanay25 (commented):
The pairing logic is nicely implemented, Service graphs will be a nice addition!

Left a few comments.

pkg/tempo/servicegraphprocessor/processor.go
```diff
@@ -0,0 +1,227 @@
...
+	}, []string{"client", "server"})
+	p.serviceGraphRequestFailedTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "tempo_service_graph_request_failed_total",
+		Help: "Total count of failed requests between two nodes",
+	}, []string{"client", "server"})
+	p.serviceGraphRequestHistogram = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name:    "tempo_service_graph_request_seconds",
+		Help:    "Time for a request between two nodes",
+		Buckets: prometheus.ExponentialBuckets(0.01, 2, 12),
+	}, []string{"client", "server"})
```
joe-elliott, line 98: an additional counter metric for unpaired edges would be nice. i.e. if i drop an edge due to the timeout or max edges before it gets paired.

we should also record the client/server on this metric as it might indicate an incoming or outgoing untraced edge.

pkg/tempo/servicegraphprocessor/processor.go
```diff
@@ -0,0 +1,227 @@
...
+	return consumer.Capabilities{}
+}
+
+func (p *processor) ConsumeTraces(ctx context.Context, td pdata.Traces) error {
+	if p.closed.Load() {
+		return nil
+	}
+
+	if p.store.ItemCount() >= p.maxEdges {
+		return ErrTooManyEdges
```
joe-elliott, line 143: unsure what returning err from this method does. logging/metricing this may make more sense. we don't want to interrupt the flow of spans if this processor is unable to generate a service graph.

mapno: Changed to log the error and return nil to not drop spans

pkg/tempo/servicegraphprocessor/processor.go
mapno: Added, thanks for the suggestion!

pkg/tempo/servicegraphprocessor/processor.go
mapno: Renamed to `maxItems`

pkg/tempo/servicegraphprocessor/processor.go
```diff
@@ -0,0 +1,227 @@
...
+type edge struct {
+	serverService, clientService string
+	serverLatency, clientLatency time.Duration
+
+	failed bool
+}
+
+func (e *edge) complete() bool {
+	return len(e.clientService) != 0 && len(e.serverService) != 0
+}
```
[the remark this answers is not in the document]
mapno: Correct! I've just merged and rebased https://github.com/grafana/agent/pull/765, which brings extracts load balancing from tail sampling.

pkg/tempo/servicegraphprocessor/processor.go
```diff
@@ -0,0 +1,227 @@
...
+
+func (p *processor) registerMetrics() error {
+	p.serviceGraphRequestTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "tempo_service_graph_request_total",
+		Help: "Total count of requests between two nodes",
+	}, []string{"client", "server"})
+	p.serviceGraphRequestFailedTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "tempo_service_graph_request_failed_total",
+		Help: "Total count of failed requests between two nodes",
+	}, []string{"client", "server"})
```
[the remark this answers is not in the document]
mapno: Yeah, I considered that too. I'm still not decided between two metrics or one metric that signals the status. WDYT?

pkg/tempo/servicegraphprocessor/factory.go
mapno: It's difficult to say. The objects stored are way smaller that the traces that are naturally in memory in the pipeline. If you consider that tail sampling might be buffering a similar amount of entire traces, it doesn't sound that bad.

Also, most items should be very short lived. If the edge is _complete_, it's immediately removed.

My initial approach was based on processing entire traces by extracting the grouping by trace functionality out of tail sampling (see https://github.com/grafana/agent/pull/743), so both features leveraged grouping only in one place. The main drawback is that you have to buffer traces for at least 1s in the sampling processor anyway.

If the deprecation of the tail sampling processor moves forward (see https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/1797), that approach is the winner for me, since we won't have to buffer in the processor anymore. But it's not the case at the moment.

pkg/tempo/servicegraphprocessor/processor.go
```diff
@@ -0,0 +1,251 @@
...
+	for _, trace := range batchpersignal.SplitTraces(td) {
+		if err := p.consume(trace); err != nil {
+			if errors.Is(err, ErrTooManyItems) {
+				level.Warn(p.logger).Log("msg", "skipped processing of spans", "maxItems", p.maxItems, "err", ErrTooManyItems)
+				break
+			}
+			errs = multierror.Append(errs, err)
+		}
+	}
+	if errs != nil {
```
joe-elliott, line 170: i don't think we should return err here either. an error in this processor should just be logged, but we should continue passing spans through the client?

maybe we should check other otel processors and see what they do.

pkg/tempo/servicegraphprocessor/processor.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+}
+
+func (p *processor) Capabilities() consumer.Capabilities {
+	return consumer.Capabilities{}
+}
+
+func (p *processor) ConsumeTraces(ctx context.Context, td pdata.Traces) error {
+	level.Info(p.logger).Log("msg", "consuming traces")
+	if p.closed.Load() {
```
joe-elliott, line 156: do we need this? the otel shutdown shuts down receivers first and then processors.

mapno: Addressed in 7122551971a9044c815835d8f8b4c8d65b0385c2

pkg/tempo/servicegraphprocessor/processor.go
mapno: Changed in 94703c34116daba96176137f8a11df8505670a4c. Other processor usually don't drop traces either.

pkg/tempo/servicegraphprocessor/processor.go
```diff
@@ -0,0 +1,255 @@
...
+				case pdata.SpanKindClient:
+					k := key(span.TraceID().HexString(), span.SpanID().HexString())
+
+					var e edge
+					if v, ok := p.store.Get(k); ok {
+						e = v.(edge)
+					}
+					e.clientService = svc.StringVal()
+					e.clientLatency = spanDuration(span)
+					p.store.SetDefault(k, e)
```
kvrhdn, line 227: I might be confused about how we are collecting our latency data... but if I understand this correctly:
- for every span, we update the store with a new latency for that service
- once per `ConsumeTraces` we observe this latency (in `collectMetrics`)

If we, for example, receive multiple spans with the same service name and durations `10s`, `5s` and `7s`. Am I right that we will only observe the latest duration (`7s`)? The previous durations will be replaced before we get to `collectMetrics`.

Second question about the latency: if a service emits multiple spans for a single request there might be one long span with a bunch of short child spans. Is the latency equal to the longest span or to the average of all spans?

pkg/tempo/servicegraphprocessor/processor.go
```diff
@@ -0,0 +1,255 @@
...
+
+func (p *processor) collectMetrics() {
+	for k, v := range p.store.Items() {
+		e := v.Object.(edge)
+		if e.complete() {
+			p.serviceGraphRequestTotal.WithLabelValues(e.clientService, e.serverService).Inc()
+			if e.failed {
+				p.serviceGraphRequestFailedTotal.WithLabelValues(e.clientService, e.serverService).Inc()
+			}
+			p.serviceGraphRequestHistogram.WithLabelValues(e.clientService, e.serverService).Observe(e.serverLatency.Seconds())
```
kvrhdn, line 190: Shouldn't we also observe `e.clientLatency` here?

pkg/tempo/servicegraphprocessor/processor.go
```diff
@@ -0,0 +1,256 @@
...
+
+func (p *processor) consume(trace pdata.Traces) error {
+	rSpansSlice := trace.ResourceSpans()
+	for i := 0; i < rSpansSlice.Len(); i++ {
+		rSpan := rSpansSlice.At(i)
+
+		svc, ok := rSpan.Resource().Attributes().Get(conventions.AttributeServiceName)
+		if !ok {
+			return ErrNoServiceName
+		}
```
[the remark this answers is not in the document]
kvrhdn: Oh lol, my comment is already outdated before even submitting it 🤣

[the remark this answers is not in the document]
mapno: Ha! Fixed, thanks!

pkg/tempo/servicegraphprocessor/processor.go
mapno: Aggregation is done by the parent (client) span, so unless two child (server) spans share the same parent span, the problem should not occur. I understand that should not happen if instrumentation is done correctly, but I will check what the specification says.

pkg/tempo/servicegraphprocessor/processor.go
mapno: Yeah, I wasn't sure how to approach this one. Maybe having two metrics for client and server latencies.

pkg/tempo/servicegraphprocessor/processor.go
```diff
@@ -0,0 +1,256 @@
...
+		ilsSlice := rSpan.InstrumentationLibrarySpans()
+		for j := 0; j < ilsSlice.Len(); j++ {
+			ils := ilsSlice.At(j)
+
+			for k := 0; k < ils.Spans().Len(); k++ {
+				span := ils.Spans().At(k)
+
+				if p.store.ItemCount() >= p.maxItems {
+					return ErrTooManyItems
+				}
```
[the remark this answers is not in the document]
mapno: That's a good point. I specially like the idea of trying to free the store if it reaches or it's close to reaching the limit.

> if we keep hitting maxItems, reduce the default expiration

That would affect the _correctness_ of the graphs, it could negatively affect the graphs more than not processing spans. It's something we can test with enough traffic.

pkg/tempo/servicegraphprocessor/processor.go
kvrhdn: Oh okay, I missed that we only gather latencies from client and server spans. Other spans will be internal and thus ignored 🤦🏻  The design makes way more sense now 😅 

A single parent span might still create multiple child spans, but I think in general it's a good practice to create one span per outgoing request.
We could log a warning if there is already a latency set, but I'm worried this could flood the logs.

PR #760, opened 2021-07-21
Operator: use reflection to discover references to secrets and configsmaps

rfratto:
#### PR Description 
Modifies the AssetReferences code to no longer manually traverse the resource hierarchy and instead use reflection to find all secret and configmap selectors.

This allows AssetReferences to always work regardless of changes we make to the CRDs and removes the risk of forgetting to update it. 

#### PR Checklist

- [x] CHANGELOG updated  (N/A)
- [x] Tests updated


mattdurham (approved):
Very light review but LGTM

PR #765, opened 2021-07-22
Split load balancing

mapno:
#### PR Description 

Split load balancing from tail sampling, as it can be useful for other purposes too, such as service graphs or other processors that benefit from having complete traces.

Need to be merged before https://github.com/grafana/agent/pull/756

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated


joe-elliott (commented):
Can we add a warning if the user configures tail based sampling (or service graphs in the future) but does not configure load balancing?

joe-elliott (approved):
sry, nit:

let's say "multi agent deployments" so it's clear that it applies if there's more than one grafana agent.

PR #773, opened 2021-07-22
Grafana Agent Crow

rfratto:
#### PR Description 
_Croooow!_

This PR creates a metrics validation tool called Grafana Agent Crow.

This is similar to Cortex's [text-exporter](https://github.com/cortexproject/cortex/tree/master/cmd/test-exporter), which wasn't working well for me. text-exporter assumes that the process is always being scraped from process boot time, which can cause false positive errors when rolling out new agents. 

Crow instead will queue samples for validation at scrape time, querying against Prometheus looking for those samples a configurable number of times before marking them as failed. Crow is _not_ a load tester. It isn't designed to generate a significant amount of series. It is only used to make sure that:

1. Scraped samples get sent.
2. Scraped samples get sent with the proper value.

This will be used alongside loki-canary and tempo-vulture in an automated smoke testing environment.

#### Notes to the Reviewer
I'm opening this in draft to get some early feedback on the approach. 

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated


pkg/crow/samples.go
```diff
@@ -0,0 +1,70 @@
...
+		samples[s] = &sample{
+			ScrapeTime: scrapeTime,
+			Labels:     prometheus.Labels{},
+			Value:      float64(sg.r.Int63n(1_000_000)),
+		}
+
+		ch <- prometheus.MustNewConstMetric(
+			desc,
+			prometheus.GaugeValue,
+			samples[s].Value, fmt.Sprintf("sample_%d", s),
```
rfratto, line 65: I don't think generating the same label each time is going to cut it here. Since series will stay alive for 4 minutes, it's pretty unlikely that we'd ever see a "missing" error after the first scrape. We'd just see a "mismatch" error since it'd be validating the sample from the previous scrape. 

Maybe generate a random label each time? There's not a ton of metrics so a churn of ~10 series per 5-60 seconds isn't a huge deal.

pkg/crow/crow.go
```diff
@@ -0,0 +1,309 @@
...
+// RegisterFlags registers flags for the config to the given FlagSet.
+func (c *Config) RegisterFlags(f *flag.FlagSet) {
+	f.StringVar(&c.PrometheusAddr, "prometheus-addr", DefaultConfig.PrometheusAddr, "Root URL of the Prometheus API to query against")
+	f.IntVar(&c.GenerateSamples, "generate-samples", DefaultConfig.GenerateSamples, "Number of samples to generate when being scraped")
+	f.StringVar(&c.UserID, "user-id", DefaultConfig.UserID, "UserID to attach to query. Useful for querying multi-tenated Cortex.")
+	f.StringVar(&c.ExtraSelectors, "extra-selectors", DefaultConfig.ExtraSelectors, "Extra selectors to include in queries, useful for identifying different instances of this job.")
+
+	f.DurationVar(&c.QueryDuration, "query-timeout", DefaultConfig.QueryTimeout, "timeout for querying")
+	f.DurationVar(&c.QueryDuration, "query-duration", DefaultConfig.QueryDuration, "time before and after sample to search")
+	f.DurationVar(&c.QueryDuration, "query-step", DefaultConfig.QueryStep, "step between samples when searching")
```
mapno, line 51: This and `query-timeout` have the wrong field

```suggestion
	f.DurationVar(&c.QueryStep, "query-step", DefaultConfig.QueryStep, "step between samples when searching")
```

pkg/crow/crow.go
```diff
@@ -0,0 +1,309 @@
...
+	cli, err := api.NewClient(api.Config{
+		Address: cfg.PrometheusAddr,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to create prometheus client: %w", err)
+	}
+
+	c := &Crow{
+		cfg:        cfg,
+		m:          newMetrics(),
```
mapno, line 130: I may have missed it, but are these metrics registered at any point?

mapno (commented):
I like the overall approach. Since generating testing metrics depends on crow being scraped, I think it would be good to alert on not being scraped.

pkg/crow/crow.go
rfratto: whoops :) thanks

pkg/crow/crow.go
rfratto: Nope, I'm trying a different pattern here. Instead of registering metrics directly, Crow exposes its two collectors (internal state & test sample generator) via a method. This means the caller has to register Crow's metrics, but it allows the caller to arbitrarily register and unregister the metrics at runtime.

Makefile
```diff
@@ -217,12 +219,14 @@ install:
...
-# for packages that have known race detection issues
-# Run tests with -online flag set to true, to include tests requiring network connectivity in CI
+# We have to run test twice: once for all packages with -race and then once
+# more without -race for packages that have known race detection issues.
+#
+# Run tests with -tags=has_network to include tests that require network
+# connectivity.
 test:
-	CGO_ENABLED=1 go test $(CGO_FLAGS) -race -cover -coverprofile=cover.out -p=4 ./... -- -online
-	CGO_ENABLED=1 go test $(CGO_FLAGS) -cover -coverprofile=cover-norace.out -p=4 ./pkg/integrations/node_exporter ./pkg/logs -- -online
```
rfratto, line 225: Turns out flags after `--` aren't read and you can't omit the `--` otherwise `-online` will be passed to every package, even packages that don't have them defined. Build tags are the best choice here.

example/docker-compose/agent/config/agent.yaml
```diff
@@ -88,8 +96,5 @@ tempo:
       spans: true
       processes: true
       roots: true
-      labels:
-        - svc
-        - status
```
rfratto, line 93: The `labels` field was removed at some point.

rfratto:
This is ready for review. I let it run for 10 minutes in docker-compose and there's a 100% success rate:

image

pkg/crow/samples.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// Ready checks if this sample is ready to be validated.
+func (s *sample) Ready(now time.Time) bool {
+	backoff := sampleBackoff(s.ValidationAttempt)
+	return now.After(s.ScrapeTime.Add(backoff))
+}
+
+func sampleBackoff(attempt int) time.Duration {
+	// Exponential backoff from 1s up to 1s + (500ms * 2^attempt).
```
rfratto, line 27:
```suggestion
	// Exponential backoff from 1s up to 1s + (250ms * 2^attempt).
```

pkg/operator/selector_eventhandler_test.go
```diff
@@ -1,12 +1,11 @@
 // These tests depend on test assets from controller-runtime which don't work on Windows.
 
-// +build !windows
+// +build !windows,has_network
```
flokli, line 3: This seems to execute despite me not setting `has_network` during a build…

flokli: This can be reproduced by running `nix-build -A grafana-agent` from a checkout of https://github.com/flokli/nixpkgs, `grafana-agent-remove-patch` branch.

rfratto: 🤔 That's strange, I don't see the same behavior:

```
$ go test -v ./pkg/operator
?       github.com/grafana/agent/pkg/operator   [no test files]
$ go test -tags=has_network -v ./pkg/operator
=== RUN   TestEnqueueRequestForSelector
# ... 
```

What version of Go are you using?

rfratto: (We had to change this because of [this comment](https://github.com/grafana/agent/pull/773#discussion_r687958362): turns out I was wrong about how the flags passing worked and the build tags are the only way to do this sanely, otherwise we'd need to add support for the network flag in every package)

flokli: That's `go 1.16.7`. If you grok at https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/go-modules/generic/default.nix#L168, that's how we invoke the build.

rfratto: Ah, thanks, I was struggling to find that. 

I see what the problem is now: v0.18.2 is a cherry-picked release containing neither your flag change or this PR: https://github.com/grafana/agent/blob/v0.18.2/pkg/operator/selector_eventhandler_test.go. 

We've been having to do a few cherry-pick releases since main contains unfinished work that we can't release yet. We thinks it's confusing (leads to situations like this) and also it's a pain to manage. We're currently working on unblocking main and coming up with a better defined branching and release strategy to avoid getting in this situation again. The current plan is to have a full release of main prior to the 10th of September.

flokli: Hm, I can reproduce the commands above do the same. It might be something specific to our test phase - we might not honor tags there properly. I'll check back on this, sorry for the noise.

flokli: Well, apparently this landed in main after the `v0.18.2` tag, so I can debug this all day :facepalm:

flokli: Okay, I can confirm this works just fine on current main. Sorry for the noise.

rfratto: No worries, sorry for the confusion with v0.18.2!

PR #790, opened 2021-08-03
CRDs: Fix typo: EnforcedNamepsaceLabel -> EnforcedNamespaceLabel

flokli:
#### PR Description 
This fixes a typo, `EnforcedNamepsaceLabel` -> `EnforcedNamespaceLabel`, and regenerates the CRDs.

#### Notes to the Reviewer
Best reviewed on a commit-by-commit base, as the previous commit introduces a lot of diff due to wrapping.

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated


rfratto:
Thanks for working on this!

> Best reviewed on a commit-by-commit base, as the previous commit introduces a lot of diff due to wrapping.

If you run `make crds` it will use a Docker image to regenerate the CRDs with a fixed version of controller-gen that won't add all the extra line wrapping. I'd like to do that before merging, just so the next person that runs `make crds` doesn't reintroduce the longer lines.

flokli:
I didn't use the container, but followed the instructions. I guess they then should not recommend using @latest, but the same version as the pinned container uses (plus cross-references in the comments on both sides)?

I'll re-roll this later today or tomorrow.

flokli:
I added a commit on top, explicitly pinning versions, and pinned `controller-gen` to its latest released version.

There's probably few reason to stick with the older version present in the container image right now - in the current state, it'd silently self-upgrade anyways.

build-image/Dockerfile
```diff
@@ -11,7 +11,8 @@ RUN apt-get update && apt-get install -y \
   && rm -rf /var/lib/apt/lists/*
 
 # Install dependencies used for the operator
-RUN go install sigs.k8s.io/controller-tools/cmd/controller-gen@latest
```
flokli, line 14: Someone/something probably needs to build/tag a new container here, and bump the version in the Makefile while merging this.

rfratto (approved):
LGTM, thanks again!

PR #796, opened 2021-08-04
Don't override existing span attributes

### LINKED ISSUE #795: Auto-tagging spans provides misleading results for pods belonging in a istio service mesh
I was trying to leverage grafana-agent's auto-tagging of spans and grafan's feature of jumping from Tempo traces to Loki logs.
My Pods are a part of a istio service mesh. They all have a istio-proxy sidecar and a istio-init initContainer. To little suprise I've learned that grafana agent was incorrectly tagging my spans with the istio-init container (didn't debug why but my guess is that because it's the first container). So I decided on adding the container attribute from within the application
```dotnet
var containerName = Environment.GetEnvironmentVariable("CONTAINER_NAME");
if (!string.IsNullOrWhiteSpace(containerName))
{
  attributes.Add(new KeyValuePair<string, object>("container", containerName));
}
```
but that didn't work because `UpsertString` is called on attributes.
https://github.com/grafana/agent/blob/997675b53a8e1685ee34f00a6dcc18e8ffa852f8/pkg/tempo/promsdprocessor/prom_sd_processor.go#L115

So I'd propose changing that to `InsertString` which would leave the original value. I've tested that on my environment and it works as expected.

PatTheSilent:
#### PR Description 

Don't overwrite existing span attributes with service discovery data

#795

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/tempo/promsdprocessor/prom_sd_processor.go
+++ b/pkg/tempo/promsdprocessor/prom_sd_processor.go
@@ -112,7 +112,7 @@ func (p *promServiceDiscoProcessor) processAttributes(attrs pdata.AttributeMap) 
 	}
 
 	for k, v := range labels {
-		attrs.UpsertString(string(k), string(v))
+		attrs.InsertString(string(k), string(v))
 	}
 }
 


rfratto:
Hey @PatTheSilent, thanks for your contribution. Given the discussion with @joe-elliott in #795, is this PR still valid?

PatTheSilent:
@rfratto no, it's not. Thanks for reminding me of it!

PR #798, opened 2021-08-05
Only for use in comparing previous release.

mattdurham:
#### PR Description 

Only used for comparing v0.18.0 to v0.18.1

#### PR Checklist

- [X] CHANGELOG updated

### DIFF
--- a/pkg/prom/instance/configstore/remote.go
+++ b/pkg/prom/instance/configstore/remote.go
@@ -24,11 +24,11 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
-/****************************************************88
+/***********************************************************************************************************************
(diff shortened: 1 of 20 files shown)


pkg/prom/instance/configstore/remote.go
```diff
@@ -162,6 +219,36 @@ func (r *Remote) List(ctx context.Context) ([]string, error) {
 	return r.kv.List(ctx, "")
 }
 
+// ListConsul returns Key Value Pairs instead of []string
```
jdbaldry, line 222: nit: Comment doesn't match function signature
```suggestion
// listConsul returns Key Value Pairs instead of []string
```

jdbaldry (commented):
Looks like a good workaround to facilitate the Consul short circuit logic which should be really valuable!

PR #808, opened 2021-08-10
Ensure drone tag always has a value, default to v0.0.0

mattdurham:
#### PR Description 

DRONE_TAG is often set to a blank string which causes issues when pushing containers.


rfratto (approved):
thanks!

PR #825, opened 2021-08-13
Automated smoke tests for metrics

rfratto:
This PR provides new `k3d` environment to be used for validating releases. Alert rules are used to flag issues, taking away the guesswork from "is this good to release or not?" The smoke test environment can be improved over time by adding more alerts to check for. 

This initial PR is limited to just validating the behavior of the metrics subsystem. Future work may be done to perform the same validations for logs, traces, and integrations. 

This PR also includes a v2 of the agent Tanka libraries which remove assumptions that previous versions made. Old libraries have been updated to point to the v2 lib where possible, removing the duplication around scrape jobs. 

The following pieces of software are deployed:

1. One Grafana Agent in non-clustered mode 
2. Three Grafana Agents in clustered mode 
3. Two Grafana Agent Crows, one per agent deployment (#773)
4. Cortex for sending metrics
5. Prometheus for monitoring everything and generating alerts 
6. Miscellaneous supporting software (Grafana, KV store for clustering, Avalanche pods, etc)

You can perform the smoke tests by running `bash ./example/k3d/scripts/smoke-test.bash`. By default, it will run for 3 hours. See the  script for flags to modify its behavior. 

The smoke test script uses two loops on a timer to simulate a mutating environment:

1. Every 5 minutes, the number of Avalanche pods will change to 0-10 replicas. 
2. Every 30 minutes, the non-clustered Grafana Agent pod and a random clustered pod will be force killed.  

After the smoke test, the script will check to see if any alerts were generated. If an alert got generated within the smoke testing period, the script will exit with a failure. 

The smoke testing k3d environment is left online for manual examination after the smoke test script completes. To clean up, rerun the smoke testing script with the `-d` flag.


rfratto:
This is working and kind of ready for review. The last bit is coming up with some more basic tests to alert on. We can always improve over time, but I want to at least add a test for agent memory and cpu usage.

production/tanka/grafana-agent/v1/internal/kubernetes_logs.libsonnet
```diff
@@ -1,185 +1,7 @@
-local gen_scrape_config(job_name, pod_uid) = {
```
mattdurham, line 1: Heads up this looks like it got changed from gen_scrape_config -> gen_logs_scrape_config which broke the bash script.

mattdurham (approved):
Ran it locally and tested all the changes listed below.

PR #748, opened 2021-07-16
Add logging support to operator

rfratto:
#### PR Description 
Completes logging support for the operator. 

#### Which issue(s) this PR fixes 
Closes #628.

#### Notes to the Reviewer
No docs changes, I'll do those in a follow up PR.

I found a few issues. Once this is merged I'll open issues and separate PRs for each.

- [x] GrafanaAgent finds LogsInstances even when not configured to find them. (fixed in #830)
- [x] Per-instance positions directory not found when starting up Agent. (#757)
- [x] Cluster label missing from logs. (#758)

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

PR #830, opened 2021-08-16
Reconcile when asset references change

### LINKED ISSUE #655: Operator: changing secret does not reconcile deployment

### LINKED ISSUE #793: operator: PrometheusInstance's {serviceMonitor,podMonitor,probe}NamespaceSelector are hard to use
Disclaimer: This is not a bug in the agent itself, but the way the CRD is built makes things more complicated.

I followed https://github.com/grafana/agent/blob/main/docs/operator/getting-started.md to deploy the operator in the cluster, and created a `PrometheusInstance`.

I'm not using bare manifests directly, but `terraform-provider-kubernetes` and `terraform-provider-kubernetes-alpha` to describe all resources there.

PrometheusInstance's `{serviceMonitor,podMonitor,probe}NamespaceSelector` fields match everything when we pass a dict, and only match the current namespace when the field is not passed.

For `terraform-provider-kubernetes-alpha`, this is the same (and I'm not sure if it's possible to tell this matters in the CRD). If I pass an empty dict there, it'll omit adding the field in the yaml entirely.

This will cause the operator to not discover any of the probes that are in another namespace.

The comment in `pkgs/operator/deployment_builder.go:104` says:

```
  // Set up our namespace label and object label selectors. By default, we'll                                                                                                                                                                                                                                                  
  // match everything (the inverse of the k8s default). If we specify anything,                                                                                                                                                                                                                                                
  // we'll narrow it down.
```

So the operator explicitly deviates from common k8s conventions.

Is there a way to tell in the CRD this makes a difference? Or de you really want to deviate from these conventions?

Right now, I have to work this around by adding a bogus `DoesNotExist` matching expression for all three fields:

```hcl
      *Selector = {
        # FIXME: this is to work around a bug where the empty {} is not
        #        serialized properly.
        #
        # Match all of the namespaces by checking that a key that is random
        # doesn't exist.
        matchExpressions = [{
          key = "xxx-AiPhua8omae5"
          operator = "DoesNotExist"
          values = []
        }]
      }
```

rfratto:
#### PR Description 
Configures the event handlers to trigger a reconcile when any referenced ConfigMap or Secret changes. 

A few changes are made to facilitate doing this nicely:

1. The event handlers will now operate on a set of interfaces that can implement custom logic for matching objects 
2. Reflection is used when discovering CRDs to reduce code deduplication 

On top of this change, a potential fix for #793 is included: we will default to Kubernetes' (and Prometheus Operator's) behavior w/r/t label selectors: select nothing when the selector is nil, and select everything when the selector is `{}`. The exception for this continues to be namespace selectors, which will match the namespace of the parent object in the hierarchy when the namespace selector is nil. 

Fixes #655 
Fixes #793 (maybe?)

#### Notes to the Reviewer
In draft for now. Opening for early discussion, but I need to test this beyond the unit level and consider doing a little more cleanup with the deployment builder. 

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated


docs/operator/getting-started.md
```diff
@@ -97,6 +97,7 @@ rules:
   resources:
   - secrets
   - services
+  - configmaps
```
rfratto, line 100: The operator's ClusterRole should've always been configured to watch for configmaps but wasn't. cc @hjet for the helm charts.

rfratto:
Testing this is blocked by #748 because the operator CRDs do not currently work against main.

rfratto:
I tested this in two scenarios:

1. Operator running outside of Kubernetes
2. Operator running inside of Kubernetes

This revealed a few final bugs in our guide, which have now been fixed. This is ready for review.

cmd/agent-operator/agent-example-config.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
   labels:
     app: grafana-agent-example
 spec:
-  image: grafana/agent:v0.18.2
+  image: grafana/agent:latest
```
rfratto, line 9: I needed to temporarily make this change since v0.18.2 doesn't have any of the logs changes needed for the operator to work & pull logs. Maybe we should leave it like this?

mattdurham (commented):
LGTM

mattdurham (approved):
Double LGTM

PR #832, opened 2021-08-18
Fix issue with v0.0.0 builds.

### LINKED ISSUE #823: Drone: merges to main not tagging images properly
Merging #773 should've created `grafana/agent-crow:main-cbde019`, but instead it pushed [`grafana/agent-crow:v0.0.0`](https://hub.docker.com/layers/grafana/agent-crow/v0.0.0/images/sha256-bde1472798173f35b24433c0fde2c842d474735e2bcdfb4f1a3d02cd623c2267?context=explore).

mattdurham:
#### PR Description 

This will fix the issue with v0.0.0 versions being pushed.

Closes #823

#### PR Checklist

- [X] CHANGELOG updated


Makefile
```diff
@@ -297,7 +298,7 @@ dist/agent-windows-installer.exe: dist/agent-windows-amd64.exe
 	cp LICENSE ./packaging/windows
 ifeq ($(BUILD_IN_CONTAINER),true)
 	docker build -t windows_installer ./packaging/windows
-	docker run --rm -t -v "${PWD}:/home" windows_installer
+	docker run --rm -t -v "${PWD}:/home" -e VERSION=${RELEASE_TAG} windows_installer
```
rfratto, line 301: Couldn't we just use `${IMAGE_TAG}` here?

rfratto, line 301: Hmm, can't we use ${IMAGE_TAG} here?

mattdurham: They are equivalent here, we could likely clean up to only use one of them.

mattdurham: Though there is a situation where you can manually override each one and they would be different, not even sure if that's a valid case?

rfratto: Yeah, I'm not sure that's something we'd want to explicitly support

rfratto (approved):
LGTM, up to you whether you want to try removing the reference to RELEASE_TAG when building the Windows installer

PR #841, opened 2021-08-20
Fix issue with global scrape_interval changes not reloading integrations

### LINKED ISSUE #827: Reload: Changing global scrape_interval does not modify integrations
For the following config file:

```yaml
prometheus:
  global:
    scrape_interval: 60s 
    remote_write: []
integrations:
  node_exporter:
    enabled: true 
```

Do the following to reproduce: 

1. Start the agent 
2. Modify the config file and change the scrape_interval to 15s 
3. Call the `/-/reload` endpoint 
4. Observe that node_exporter is being scraped every 60 seconds

If you configure remote_write, you can observe the scrape_interval being unchanged by querying for `count_over_time(node_uname_info[1m])`. It will be 1 when there's only 1 sample per minute (60s scrape_interval) and 4 when there are 4 samples per minute (15s scrape_interval)

kgeckhart:
#### PR Description 
We are doing a yaml compare of the IntegrationsManager config on reload to see if anything changed. Prometheus settings are not applied to the Integration configs the IntegrationManager has so it looks like nothing changed and we short circuit out. 

This change adds the global Prometheus Config to the IntegrationManager config to avoid short circuiting. There's also some small logging changes I kept that made it a little easier while troubleshooting. 

Fixes #827 

#### PR Checklist

- [x] CHANGELOG updated 
- [NA] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/integrations/manager.go
+++ b/pkg/integrations/manager.go
@@ -20,7 +20,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/prometheus/common/model"
-	"github.com/prometheus/prometheus/config"
+	promConfig "github.com/prometheus/prometheus/config"
 	"github.com/prometheus/prometheus/discovery"
 	"github.com/prometheus/prometheus/pkg/relabel"
 )
@@ -59,7 +59,7 @@ type ManagerConfig struct {
 	Labels model.LabelSet `yaml:"labels,omitempty"`
 
 	// Prometheus RW configs to use for all integrations.
-	PrometheusRemoteWrite []*config.RemoteWriteConfig `yaml:"prometheus_remote_write,omitempty"`
+	PrometheusRemoteWrite []*promConfig.RemoteWriteConfig `yaml:"prometheus_remote_write,omitempty"`
 
 	IntegrationRestartBackoff time.Duration `yaml:"integration_restart_backoff,omitempty"`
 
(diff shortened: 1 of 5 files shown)


pkg/integrations/manager.go
```diff
@@ -194,8 +199,13 @@ func (m *Manager) ApplyConfig(cfg ManagerConfig) error {
 	m.integrationsMut.Lock()
 	defer m.integrationsMut.Unlock()
 
-	if util.CompareYAML(m.cfg, cfg) {
+	// The global prometheus config settings don't get applied to integrations until later. This
+	// causes us to skip reload when those settings change.
+	if util.CompareYAML(m.cfg, cfg) && util.CompareYAML(m.cfg.PrometheusGlobalConfig, cfg.PrometheusGlobalConfig) {
+		level.Info(m.logger).Log("msg", "Integrations config is unchanged skipping apply")
```
rfratto, line 205: nit: can we make this log (and the log in the else case) debug-level?

pkg/integrations/manager.go
```diff
@@ -194,8 +199,13 @@ func (m *Manager) ApplyConfig(cfg ManagerConfig) error {
...
 		return nil
+	} else {
+		level.Info(m.logger).Log("msg", "Applying integrations config changes")
```
rfratto, line 208: nit: can we drop the dangling else here?

rfratto (approved):
Thanks! Only minor nits, but I'm happy with this.

rfratto:
Please do put this in the changelog though at the end of the list of bugfixes :)

PR #845, opened 2021-08-23
Operator: breaking change: rename Prometheus* to Metrics*

### LINKED ISSUE #840: Operator: remove hypens from CRD names
The hypens in the names we use for CRDs are unidiomatic. We should change the following prior to 1.0:

1. `grafana-agents` to `grafanaagents` 
2. `prometheus-instances` to `prometheusinstances` (or `metricsinstances` after renaming the CRD)
3. `logs-instances` to `logsinstances` 

This is technically a breaking change, though it only affects how users retrieve the objects from the command line, and should be minimally disruptive.

rfratto:
#### PR Description 
Renames `Prometheus*` CRDs and fields to `Metrics*`. 

Since the Operator is still young, this is done as a fully breaking change. Users **must** reconfigure their custom resources after updating the Operator and installing the new CRDs. 

Also closes #840 since it wasn't too much extra work to do that on top of this. 
Related to #722. The rename of the subsystem within the Agent will be a deprecation instead of a hard breaking change like this. 

#### Notes to the Reviewer
Opening up for early feedback. I don't feel great about the breaking change, but I failed to find a good way to handle this otherwise. I think we should make this change now while the operator's userbase is still small.

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated


mattdurham (approved):
LGTM

rfratto:
Given how disruptive of a change this is, I think this probably deserves some exploration of the alternative:

* Is there a sane way to do this with a new CRD version? (`v1alpha2`)
* When upgrading CRD versions, does every CRD from the same GroupKind go with it or just the one you're modifying?
* Is it worthwhile trying to do the upgrade with a MutatingWebHookConfig or should we do upgrades fully in-process without Kubernetes being aware of it?

rfratto:
We've discussed this in more detail and agreed to go with the breaking change for now. I'll explore making a script to do a zero-downtime upgrade, but odds are we'll just document how to upgrade without losing 1-2 minutes of metrics.

PR #860, opened 2021-08-30
Register missing metric

### LINKED ISSUE #855: Configstore: consul latency histogram unregistered
https://github.com/grafana/agent/blob/95f4504792f7e52e41e37f6e73fe2b10b43ab171/pkg/metrics/instance/configstore/remote.go#L31-L37

This metric is defined but never registered.

rfratto:
#### PR Description 
Registers `agent_configstore_consul_request_duration_seconds` to track the time it takes to do a consul list using the optimization shortcut. 

Note that the metric must have a different name due to it being a separate histogram, and the underlying metric used by Cortex's kv store is not exposed so we have to make our own.

Fixes #855 

#### PR Checklist

- [x] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,8 @@ for specific instructions.
 
 - [BUGFIX] Fix yaml marshalling tag for cert_file in kafka exporter agent config. (@rgeyer)
 
+- [BUGFIX] Register missing metric for configstore consul request duration.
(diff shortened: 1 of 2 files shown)


mattdurham (approved):
LGTM

Issue #857, opened 2021-08-30
Scraping service: reshard timeout misapplied to cluster-wide reshard


rfratto:
https://github.com/grafana/agent/blob/95f4504792f7e52e41e37f6e73fe2b10b43ab171/pkg/metrics/cluster/node.go#L176-L186

The timeout for reshard is being applied to the reshard of the entire cluster, which:

- Runs in parallel across the other nodes, with jitter of 500ms 
- Then runs locally 

This means that if any of the other nodes in the cluster time out, the local reshard will never get to run. I propose changing the gRPC reshard to schedule a background refresh and return immediately.

mattdurham:
Agreed on the background refresh. On the returning immediately and spawning a background routine.  I think has more limited benefit. The transfer out requires a return of error for the flushtransferer interface, which requires a wait.

PR #864, opened 2021-08-31
Fixes for honoring timeout and having lifecycle reshard use background context

### LINKED ISSUE #856: Scraping service: timeouts do not apply to periodic reshard
https://github.com/grafana/agent/blob/95f4504792f7e52e41e37f6e73fe2b10b43ab171/pkg/metrics/cluster/config_watcher.go#L100-L102

Periodic reshards use the context of the config watcher and are not subject to timeouts. This leads to potentially unbounded local reshards.

mattdurham:
#### PR Description 

The poling refresh interval should use the timeout specified, added additional logging to distinguish between a context deadline exceeded from waiting on mutex and waiting on kv store.

Lifecycle refresh also happens in the background on a background thread. 

Closes #856 and #857 

#### PR Checklist

- [X] CHANGELOG updated 
- [X] Documentation added
- [X] Tests updated

### DIFF
--- a/pkg/prom/cluster/node.go
+++ b/pkg/prom/cluster/node.go
@@ -159,7 +159,6 @@ func newRing(cfg ring.Config, name, key string, reg prometheus.Registerer) (*rin
 func (n *node) run() {
 	for range n.reload {
 		n.mut.RLock()
-
 		if err := n.performClusterReshard(context.Background(), true); err != nil {
 			level.Warn(n.log).Log("msg", "dynamic cluster reshard did not succeed", "err", err)
 		}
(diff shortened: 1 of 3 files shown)


pkg/prom/cluster/node.go
```diff
@@ -214,8 +206,8 @@ func (n *node) performClusterReshard(ctx context.Context, joining bool) error {
 			return nil, nil
 		}
 
-		ctx = user.InjectOrgID(ctx, "fake")
-		return nil, n.notifyReshard(ctx, id)
+		c = user.InjectOrgID(c, "fake")
```
mattdurham, line 209: Using the parent ctx seemed like a legit issue, since it is not the same as the context passed in the function.

pkg/prom/cluster/node.go
```diff
@@ -214,8 +206,8 @@ func (n *node) performClusterReshard(ctx context.Context, joining bool) error {
 			return nil, nil
 		}
 
-		ctx = user.InjectOrgID(ctx, "fake")
-		return nil, n.notifyReshard(ctx, id)
+		notifyCtx := user.InjectOrgID(c, "fake")
```
mattdurham, line 209: The original code seemed like a bug, since the ctx != c context passed into this function.

PR #876, opened 2021-09-01
Honor config refresh timeout, remove timeout for lifecycle and instead use background context

### LINKED ISSUE #856: Scraping service: timeouts do not apply to periodic reshard

### LINKED ISSUE #857: Scraping service: reshard timeout misapplied to cluster-wide reshard

### LINKED ISSUE #877: Scraping service can deadlock when reloading during join
The optimized consul path in the config store has the potential to deadlock. `node.performClusterReshard` and `node.Owns` are called in the same goroutine, both trying to hold a read lock. This means that if the config is reloaded (which grabs a write lock)  in the middle of self-assigning configs, the next call to `node.Owns` will deadlock. 

1. `node.run` grabs a read mutex, held for the duration of the join. 
2. In the same goroutine (eventually), `Reshard` is called against the local process. 
3. `Reshard` will start calling `node.Owns` (via `configWatcher.Refresh`) against the node to self-assign configs. 
4. The user invokes `/-/reload` with changes to the cluster. This grabs a write lock. 
5. The next call to `node.Owns` deadlocks, since there is a pending write lock, but `node.run` still has the read lock. 

This can be resolved by having performClusterReshard do background work or to hold the read lock for a shorter period of time in a way that can't ever recursively lock on read.

mattdurham:
#### PR Description 

Config refresh timeout should honor the timeout, but the lifecycle timeout should use a background context.

Closes #856 
Closes #857 
Closes #877 

#### PR Checklist

- [X] CHANGELOG updated 
- [X] Documentation added
- [NA] Tests updated

### DIFF
--- a/docs/configuration/prometheus-config.md
+++ b/docs/configuration/prometheus-config.md
@@ -73,7 +73,7 @@ agents distribute discovery and scrape load between nodes.
 [reshard_timeout: <duration> | default = "30s"]
 
 # The timeout for a cluster reshard events. A timeout of 0 indicates no timeout.
-[refresh_timeout: <duration> | default = "30s"]
+[cluster_reshard_event_timeout: <duration> | default = "30s"]
 
 # Configuration for the KV store to store configurations.
 kvstore: <kvstore_config>
--- a/docs/configuration/prometheus-config.md
+++ b/docs/configuration/prometheus-config.md
@@ -61,14 +61,19 @@ agents distribute discovery and scrape load between nodes.
 # cannot be used.
 [enabled: <boolean> | default = false]
 
-# How often should the agent manually reshard. Useful for if KV change
+# Note these next 3 configuration options are confusing. Due to backwards compatibility the naming
+# is less than ideal.
+
+# How often should the agent manually refresh the configuration. Useful for if KV change
 # events are not sent by an agent.
 [reshard_interval: <duration> | default = "1m"]
 
-# The timeout for a reshard. Applies to a cluster-wide reshard (done when
-# joining or leaving the cluster) and local reshards (done every
-# reshard_interval). A timeout of 0 indicates no timeout.
+# The timeout for configuration refreshes. This can occur on cluster events or 
+# on the reshard interval. A timeout of 0 indicates no timeout.
 [reshard_timeout: <duration> | default = "30s"]
+
+# The timeout for a cluster reshard events. A timeout of 0 indicates no timeout.
+[refresh_timeout: <duration> | default = "30s"]
 
 # Configuration for the KV store to store configurations.
 kvstore: <kvstore_config>
--- a/pkg/metrics/cluster/cluster.go
+++ b/pkg/metrics/cluster/cluster.go
@@ -107,10 +107,9 @@ func (c *Cluster) storeValidate(cfg *instance.Config) error {
 // Reshard implements agentproto.ScrapingServiceServer, and syncs the state of
 // configs with the configstore.
 func (c *Cluster) Reshard(ctx context.Context, _ *agentproto.ReshardRequest) (*empty.Empty, error) {
-	c.mut.RLock()
-	defer c.mut.RUnlock()
-
 	go func() {
+		c.mut.RLock()
+		defer c.mut.RUnlock()
 		err := c.watcher.Refresh(context.Background())
 		if err != nil {
 			level.Error(c.log).Log("msg", "failed to perform local reshard", "err", err)
--- a/pkg/metrics/cluster/cluster.go
+++ b/pkg/metrics/cluster/cluster.go
@@ -110,11 +110,13 @@ func (c *Cluster) Reshard(ctx context.Context, _ *agentproto.ReshardRequest) (*e
 	c.mut.RLock()
 	defer c.mut.RUnlock()
 
-	err := c.watcher.Refresh(ctx)
-	if err != nil {
-		level.Error(c.log).Log("msg", "failed to perform local reshard", "err", err)
-	}
-	return &empty.Empty{}, err
+	go func() {
+		err := c.watcher.Refresh(context.Background())
+		if err != nil {
+			level.Error(c.log).Log("msg", "failed to perform local reshard", "err", err)
+		}
+	}()
+	return &empty.Empty{}, nil
 }
 
 // ApplyConfig applies configuration changes to Cluster.
--- a/pkg/metrics/cluster/config.go
+++ b/pkg/metrics/cluster/config.go
@@ -15,12 +15,12 @@ var DefaultConfig = *flagutil.DefaultConfigFromFlags(&Config{}).(*Config)
 
 // Config describes how to instantiate a scraping service Server instance.
 type Config struct {
-	Enabled         bool                  `yaml:"enabled"`
-	ReshardInterval time.Duration         `yaml:"reshard_interval"`
-	ReshardTimeout  time.Duration         `yaml:"reshard_timeout"`
-	RefreshTimeout  time.Duration         `yaml:"refresh_timeout"`
-	KVStore         kv.Config             `yaml:"kvstore"`
-	Lifecycler      ring.LifecyclerConfig `yaml:"lifecycler"`
+	Enabled                    bool                  `yaml:"enabled"`
+	ReshardInterval            time.Duration         `yaml:"reshard_interval"`
+	ReshardTimeout             time.Duration         `yaml:"reshard_timeout"`
+	ClusterReshardEventTimeout time.Duration         `yaml:"cluster_reshard_event_timeout"`
+	KVStore                    kv.Config             `yaml:"kvstore"`
+	Lifecycler                 ring.LifecyclerConfig `yaml:"lifecycler"`
 
 	DangerousAllowReadingFiles bool `yaml:"dangerous_allow_reading_files"`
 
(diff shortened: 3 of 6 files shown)


pkg/metrics/cluster/config_watcher.go
```diff
@@ -131,6 +136,11 @@ func (w *configWatcher) Refresh(ctx context.Context) (err error) {
 		reshardDuration.WithLabelValues(success).Observe(time.Since(start).Seconds())
 	}()
 
+	// This is used to determine if the context was already exceeded before calling the kv provider
+	if err = ctx.Err(); err != nil {
+		return fmt.Errorf("context deadline exceeded before calling store.all")
+	}
```
rfratto, line 142: There's a chance that the context will cancel right after this check but still before `w.store.All` gets invoked. Can you supplant this with a debug-level log that happens after w.store.All?

pkg/metrics/cluster/node.go
```diff
@@ -176,13 +175,6 @@ func (n *node) run() {
 func (n *node) performClusterReshard(ctx context.Context, joining bool) error {
 	if n.ring == nil || n.lc == nil {
 		level.Info(n.log).Log("msg", "node disabled, not resharding")
-		return nil
-	}
-
-	if n.cfg.ReshardTimeout > 0 {
-		var cancel context.CancelFunc
-		ctx, cancel = context.WithTimeout(ctx, n.cfg.ReshardTimeout)
-		defer cancel()
```
rfratto, line 185: This isn't quite right. Now cluster wide reshards aren't subject to timeouts but refreshes are. This will cause the same problem we've had before but just with the other type of reshard.

The alternative here is this method initiated background (ie asynchronous, in a separate goroutine) reshards, so the cluster-wide reshard would exit immediately.

rfratto (changes requested):

pkg/metrics/cluster/node.go
rfratto: (Or the Refresh method itself could determine a timeout, but still this method would take quite a long time to execute. I'm not sure that necessarily matters, though).

mattdurham: Per discussion moved timeout to refresh function, and the node triggers the refresh in a background to minimize mutex locking.

docs/configuration/prometheus-config.md
```diff
@@ -61,15 +61,20 @@ agents distribute discovery and scrape load between nodes.
...
 
-# The timeout for a reshard. Applies to a cluster-wide reshard (done when
-# joining or leaving the cluster) and local reshards (done every
-# reshard_interval). A timeout of 0 indicates no timeout.
+# The timeout for configuration refreshes. This can occur on cluster events or 
+# on the reshard interval. A timeout of 0 indicates no timeout.
 [reshard_timeout: <duration> | default = "30s"]
 
+# The timeout for a cluster reshard events. A timeout of 0 indicates no timeout.
+[refresh_timeout: <duration> | default = "30s"]
```
rfratto, line 76: nit: What about `cluster_reshard_event_timeout`? Your comment seems to nail the description.

pkg/metrics/cluster/config.go
```diff
@@ -50,8 +51,10 @@ func (c *Config) RegisterFlags(f *flag.FlagSet) {
 // FlagSet with a specified prefix.
 func (c *Config) RegisterFlagsWithPrefix(prefix string, f *flag.FlagSet) {
 	f.BoolVar(&c.Enabled, prefix+"enabled", false, "enables the scraping service mode")
-	f.DurationVar(&c.ReshardInterval, prefix+"reshard-interval", time.Minute*1, "how often to manually reshard")
-	f.DurationVar(&c.ReshardTimeout, prefix+"reshard-timeout", time.Second*30, "timeout for cluster-wide reshards and local reshards. Timeout of 0s disables timeout.")
+	f.DurationVar(&c.ReshardInterval, prefix+"reshard-interval", time.Minute*1, "how often to manually refresh configuration")
+	f.DurationVar(&c.ReshardTimeout, prefix+"reshard-timeout", time.Second*30, "timeout for refreshing the configuration. Timeout of 0s disables timeout.")
+	f.DurationVar(&c.RefreshTimeout, prefix+"refresh-timeout", time.Second*30, "timeout for the cluster reshard. Timeout of 0s disables timeout.")
```
rfratto, line 56: I think we can set this a lot smaller since it kicks off an async job. 5s maybe?

pkg/metrics/cluster/config_watcher.go
```diff
@@ -131,6 +144,12 @@ func (w *configWatcher) Refresh(ctx context.Context) (err error) {
 		reshardDuration.WithLabelValues(success).Observe(time.Since(start).Seconds())
 	}()
 
+	// This is used to determine if the context was already exceeded before calling the kv provider
+	if err = ctx.Err(); err != nil {
+		return fmt.Errorf("context deadline exceeded before calling store.all")
```
rfratto, line 149: nit: I'm a little uncomfortable with this since it breaks common expectations about what a timeout error looks like in Go, plus the deadline can still expire in between this check and invoking w.store.All.

pkg/metrics/cluster/node.go
```diff
@@ -176,59 +175,66 @@ func (n *node) run() {
 func (n *node) performClusterReshard(ctx context.Context, joining bool) error {
 	if n.ring == nil || n.lc == nil {
 		level.Info(n.log).Log("msg", "node disabled, not resharding")
-		return nil
```
rfratto, line 179: I think you accidentally removed `return nil` here, it should be added back

pkg/metrics/cluster/node.go
```diff
@@ -176,59 +175,66 @@ func (n *node) run() {
...
-		firstError error
 	)
 
+	// This cancel has to exist outside since it is used in the goroutine
+	// originally it was put in a if !joining block which would logically
+	// ensure it would be called, but the linter does not process those paths
+	// so instead its sprinkled around before the returns.
+	var cancel context.CancelFunc
+	if n.cfg.RefreshTimeout > 0 {
+		ctx, cancel = context.WithTimeout(ctx, n.cfg.RefreshTimeout)
```
rfratto, line 191: There should be a `defer cancel()` here so it can't leak.

pkg/metrics/cluster/node.go
```diff
@@ -176,59 +175,66 @@ func (n *node) run() {
...
+	}
+
 	backoff := cortex_util.NewBackoff(ctx, backoffConfig)
 	for backoff.Ongoing() {
+		if ctx.Err() != nil {
+			cancel()
```
rfratto, line 197: This can be removed after adding the defer above.

pkg/metrics/cluster/node.go
```diff
@@ -176,59 +175,66 @@ func (n *node) run() {
...
+			level.Info(n.log).Log("msg", "running local reshard")
+			if _, reshardErr := n.srv.Reshard(ctx, &pb.ReshardRequest{}); reshardErr != nil {
+				level.Warn(n.log).Log("msg", "dynamic local reshard did not succeed", "err", err)
+			}
+		}()
 	}
-
-	return firstError
+	cancel()
+	return err
```
rfratto, line 237: This chunk here is still subject to timeout issues as we discussed offline. I strongly think we need to have the implementation of `Reshard` to kick off a Goroutine, and remove spawning a Goroutine here. 

As it is now, the calls to `.notifyReshard` will block until all other nodes finish refreshing, which may also include waiting for one of the tick refreshes to finish. It would lead to a similar failure issue with the new timeout we added.

pkg/metrics/cluster/node.go
mattdurham: Yes I can do this now since we moved the goroutine to the reshard, beforehand this was needed because it would cancel before the goroutine was ran and always be cancelled.

pkg/metrics/cluster/node.go
mattdurham: Per our conversation moved to the reshard function. Tested to see if any negative effects would occur and none were found.

pkg/metrics/cluster/config.go
mattdurham: I am a bit hesitant to drop that low. 30 seconds is likely overkill but is inline with our previous values.

pkg/metrics/cluster/config_watcher.go
mattdurham: Reformatted this

pkg/metrics/cluster/cluster.go
```diff
@@ -110,11 +110,13 @@ func (c *Cluster) Reshard(ctx context.Context, _ *agentproto.ReshardRequest) (*e
...
-	err := c.watcher.Refresh(ctx)
-	if err != nil {
-		level.Error(c.log).Log("msg", "failed to perform local reshard", "err", err)
-	}
-	return &empty.Empty{}, err
+	go func() {
+		err := c.watcher.Refresh(context.Background())
+		if err != nil {
+			level.Error(c.log).Log("msg", "failed to perform local reshard", "err", err)
+		}
```
rfratto, line 117: The read mutex needs to be held in the goroutine here, not in the function.

pkg/metrics/cluster/node.go
```diff
@@ -179,56 +179,56 @@ func (n *node) performClusterReshard(ctx context.Context, joining bool) error {
...
+
+	if err != nil {
+		level.Error(n.log).Log("msg", "notifying other nodes failed", "err", err)
 	}
 
+	// This is ran in the background to help prevent unnecessary mutex locking.
 	if joining {
 		level.Info(n.log).Log("msg", "running local reshard")
-		if _, err := n.srv.Reshard(ctx, &pb.ReshardRequest{}); err != nil {
+		if _, reshardErr := n.srv.Reshard(ctx, &pb.ReshardRequest{}); reshardErr != nil {
```
rfratto, line 227: You changed the name of the variable but the log needs to be updated still

rfratto (commented):
This also closes #877, can you update the comment to link the issue? (And mention it in the CHANGELOG, please)

pkg/metrics/cluster/cluster.go
mattdurham: Done

pkg/metrics/cluster/node.go
mattdurham: Reverted back to the old behavior entirely.

rfratto (approved):
LGTM, thanks!

PR #881, opened 2021-09-02
Move port to load_balancing config as receiver_port

mapno:
#### PR Description

When extracting load balancing from tail sampling, only the the load_balancing block was moved into its own block, but the receiver's port was left in tail_sampling.

This means that the receiver port for load_balancing could not be configured without tail_sampling.

Now it's moved to load_balancing as receiver_port.

It also includes a fix for when using load balancing where pipelines wouldn't be split if tail sampling wasn't enabled as well.

#### Notes to the Reviewer

Change is not in the CHANGELOG since the load balancing change hasn't been released yet.

#### PR Checklist

- [ ] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/upgrade-guide/_index.md
+++ b/docs/upgrade-guide/_index.md
@@ -30,6 +30,7 @@ Example old config:
 tail_sampling:
   policies:
     - always_sample:
+  port: 4318
   load_balancing:
     exporter:
       insecure: true
@@ -37,7 +38,6 @@ tail_sampling:
       dns:
         hostname: agent
         port: 4318
-  port: 4318
 ```
 
 Example new config:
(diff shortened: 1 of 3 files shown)


mapno:
Mirror PR for `dev.servicehraphs`: https://github.com/grafana/agent/pull/891

docs/upgrade-guide/_index.md
```diff
@@ -36,6 +37,7 @@ tail_sampling:
       dns:
         hostname: agent
         port: 4318
+  port: 4318
```
rfratto, line 40: nit: could we move the port here to be above load balancing to make it clear where it was nested? Just at a glance it's hard to tell whether it's a child of `tail_sampling` or `load_balancing` here

rfratto (approved):
LGTM, will 👍 other PR once this is merged and once it reflects any changes you make here

docs/upgrade-guide/_index.md
mapno: Good point, moved in 2b567a273e4f62c08598a783478adb0a17dd1e40. Thanks

PR #882, opened 2021-09-02
Rename prometheus to metrics

rfratto:
#### PR Description 
Aliases `prometheus` to `metrics` and deprecates the former. 

Related to #722. 

#### Notes to the Reviewer
Does not touch the tracing configs yet, that will be done in a follow up PR.

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,11 @@ for specific instructions.
   `loki_tag` is now `logs_instance_tag`, and `backend: loki` is now
   `backend: logs_instance`. (@rfratto)
 
+- [DEPRECATION] The `prometheus` key at the root of the config file has been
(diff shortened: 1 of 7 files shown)


rfratto:
I _think_ this will cause problems since not all defaults are being stored to the old name. Let me try to verify in a test, though the fix is simple.

rfratto:
> I think this will cause problems since not all defaults are being stored to the old name. Let me try to verify in a test, though the fix is simple.

There's nothing for metrics that depends on defaults being only loaded in from flags (which shouldn't be the case anyway). I'll still add something defensively until we can clean all of this up.

mattdurham (approved):
LGTM

Issue #875, opened 2021-09-01
Introduce an agent-wide clustering mechanism


rfratto:
Today, users can cluster their agents through the scraping service, a very specialized mode of the metrics subsystem that has its own code path (related: #873). 

Other subsystems are not able to take advantage of this clustering mechanism, like integrations (which would always run an integration locally) or tracing (which could use it for tail-based sampling). 

I propose that we move the clustering mechanism to be done agent-wide to allow other subsystems to take advantage of it. 

This is also an opportunity to introduce a _simpler_ clustering mechanism. The Grafana Agent does not need all the features found from the Cortex hash ring, and can be drastically simplified, making it easier to understand and maintain. 

I [prototyped](https://github.com/rfratto/agent/tree/per-target-sharding) an initial version of a simplified cluster for per-target sharding (#872). This used [rfratto/croissant](https://github.com/rfratto/croissant) for clustering, but it revealed significant issues with distribution of targets. This led me to re-think the approach, leading to the creation of [rfratto/ckit](https://github.com/rfratto/ckit), which I believe will be better suited for the clustering mechanism.

rfratto:
Right now, there's no plans to do this, so I'm going to close it as won't do. I'm still interested in supporting agent-wide clustering, but it seems more likely to be done within Grafana Agent Flow instead of the original idea suggested here.

PR #887, opened 2021-09-03
[dev.metrics-v2] Agent-wide cluster mechanism

rfratto:
#### PR Description 
Introduces an Agent-wide clustering mechanism based off of [rfratto/ckit](https://github.com/rfratto/ckit) and my previous experiments with Pastry to implement a prototype of [per-target sharding](https://github.com/rfratto/agent/tree/per-target-sharding). 

A lot is missing here. Nothing is actually using the cluster, there are no metrics, no HTTP page to investigate the state. I'll at least add support for metrics and an HTTP page into ckit and update this PR. 

Related: #875

#### Notes to the Reviewer
See unit tests for an idea of how to run this locally.

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/cmd/agent/entrypoint.go
+++ b/cmd/agent/entrypoint.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"net"
 	"net/http"
@@ -8,8 +9,13 @@ import (
 	"os/signal"
 	"sync"
 	"syscall"
+	"time"
 
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
 	"github.com/gorilla/mux"
+	"github.com/grafana/agent/pkg/cluster"
+	"github.com/grafana/agent/pkg/config"
 	"github.com/grafana/agent/pkg/integrations"
 	"github.com/grafana/agent/pkg/logs"
 	loki "github.com/grafana/agent/pkg/logs"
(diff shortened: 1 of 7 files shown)


cmd/agent/entrypoint.go
```diff
@@ -273,6 +286,23 @@ func (ep *Entrypoint) Start() error {
...
+	{
+		ctx, cancel := context.WithCancel(context.Background())
+		defer cancel()
+
+		g.Add(func() error {
+			// Wait just a litle bit of time to start the cluster so that our APIs
+			// and other components are running, allowing peers to immediately invoke
+			// them.
+			time.Sleep(100 * time.Millisecond)
+			ep.node.Start()
```
rfratto, line 298: This seems like a bit of a code smell, and is obscuring a bigger underlying problem. The original intent was to protect peers from invoking HTTP or gRPC endpoints before the listeners are open. (Note that that wouldn't actually happen, since the ports are open prior to this goroutine ever running, but that's beside the point). 

However, there generally may be transient HTTP or gRPC disconnects, and there is going to need to be a recommended way of handling that. 

If a function needs to deliver a message to its owner, but the owner is not responding, the function has two choices:

- Return a failure (either immediately or after a retry period)
- Failover to the next owner 

Failover requires specific scenarios to be acceptable; if you are transmitting stateful data, you would not want to do premature failover before an unhealthy node is completely evicted from the cluster. Generally, returning a failure is the safest approach, where the caller would either retry or drop the failed request. 

rfratto/ckit may want to extend itself to support node health checks to further filter down the memberlist to the set of working nodes.

rfratto:
I'm going to close this. This has been open for long enough that the dev branch just needs to be re-done instead of having this sit around :)

PR #901, opened 2021-09-08
Check span error

mapno:
#### PR Description

Checks span status and http status code to see if the span failed. If either the client or server span contains an error, the request will be considered as failed.

#### PR Checklist

- [ ] ~CHANGELOG updated~
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/tempo/servicegraphprocessor/processor.go
+++ b/pkg/tempo/servicegraphprocessor/processor.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"net/http"
 	"time"
 
 	util "github.com/cortexproject/cortex/pkg/util/log"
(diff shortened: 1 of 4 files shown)


pkg/tempo/servicegraphprocessor/processor.go
```diff
@@ -261,6 +265,11 @@ func (p *processor) consume(trace pdata.Traces) error {
 	return nil
 }
 
+func spanFailed(span pdata.Span) bool {
+	httpStatusCode, ok := span.Attributes().Get("http.status_code")
+	return span.Status().Code() == pdata.StatusCodeError || (ok && httpStatusCode.IntVal()/100 != 2)
```
joe-elliott, line 270: We should make this configurable. 300s are not errors and 400s are debatable on whether or not you'd want them included as a "failed" request. 

We should also add support for `rpc.grpc.status_code`
 https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/rpc.md#grpc

mapno: Made configurable, thanks for the suggestion

joe-elliott (approved):

PR #905, opened 2021-09-09
Rename prom_instance -> metrics_instance

rfratto:
I think this is the only thing mentioning `prom`/`prometheus` in the tracing code. Let me know if I missed something else.

Related to #722.


mapno (approved):
LGTM. We will need to update Tempo's docs too.

PR #906, opened 2021-09-10
tempo: if relabel config drops target, don't attempt to process (#904)

### LINKED ISSUE #904: Handle target being dropped in Tempo relabel config
The Agent uses Prometheus' SD in Tempo to relabel spans via `relabel_configs`. This is commonly used to add metadata about the source, such as kubernetes labels.

Relabel configs allow to drop targets. When dropping a label set, [`relabel.Process`](https://github.com/grafana/agent/blob/main/pkg/tempo/promsdprocessor/prom_sd_processor.go#L188) returns an empty label set. This is an issue because then the Agent will warn that it failed to find the `__address__` label ([source code](https://github.com/grafana/agent/blob/main/pkg/tempo/promsdprocessor/prom_sd_processor.go#L196-L200)).

The Agent should differentiate between the address being missing and dropping a label set.

james-callahan:
#### PR Description 

tempo agent: if relabel config drops target, don't attempt to process it.
This avoids e.g. logging a warning when a target is intentionally dropped

#904

#### Notes to the Reviewer

The current setup for testing doesn't have any easy path to test output of logging; so I wasn't able to add a test for this.

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/tempo/promsdprocessor/prom_sd_processor.go
+++ b/pkg/tempo/promsdprocessor/prom_sd_processor.go
@@ -187,6 +187,9 @@ func (p *promServiceDiscoProcessor) syncTargets(jobName string, group *targetgro
 		}
 		processedLabels := relabel.Process(labels.FromMap(labelMap), relabelConfig...)
 		level.Debug(p.logger).Log("processedLabels", processedLabels)
+		if processedLabels == nil { // dropped
+			continue
+		}
 
 		var labels = make(model.LabelSet)
 		for k, v := range processedLabels.Map() {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,8 @@ for specific instructions.
 
 - [BUGFIX] Fix yaml marshalling tag for cert_file in kafka exporter agent config. (@rgeyer)
 
+- [BUGFIX] Fix warn-level logging of dropped targets. (@james-callahan)
+
 - [CHANGE] Breaking change: reduced verbosity of tracing autologging
   by not logging `STATUS_CODE_UNSET` status codes. (@mapno)
 


pkg/tempo/promsdprocessor/prom_sd_processor.go
```diff
@@ -187,6 +187,9 @@ func (p *promServiceDiscoProcessor) syncTargets(jobName string, group *targetgro
 		}
 		processedLabels := relabel.Process(labels.FromMap(labelMap), relabelConfig...)
 		level.Debug(p.logger).Log("processedLabels", processedLabels)
+		if processedLabels == nil { // dropped
```
mapno, line 190: I think we can move this check before the debug log above it.

mapno (commented):
Thanks for the contribution! Please, add the bugfix to the CHANGELOG and @ yourself.

WDYT of adding a debug log about the dropped label set as you suggested in the issue?

james-callahan:
> WDYT of adding a debug log about the dropped label set as you suggested in the issue?

I felt like that was covered by the logging of `processedLabels {}`

mapno:
> > WDYT of adding a debug log about the dropped label set as you suggested in the issue?
> 
> I felt like that was covered by the logging of `processedLabels {}`

Sorry for the delay. Yeah, that works too. Please update the CHANGELOG with the bugfix and we can merge. Thanks!

james-callahan:
> Sorry for the delay. Yeah, that works too. Please update the CHANGELOG with the bugfix and we can merge. Thanks!

Done.

mapno (approved):

PR #907, opened 2021-09-10
Operator: Add support for managing a kubelet service

### LINKED ISSUE #885: Operator: Automatically generate service for cAdvisor/Kubelet
Prometheus Operator automatically manages a cAdvisor/Kubelet service for each node, allowing users to configure a ServiceMonitor to monitor nodes themselves. 

This was missed in the initial implementation and needs to be added.

rfratto:
#### PR Description 
Enables the operator to manage a Kubelet service, creating one endpoint per node in the cluster. This allows users to define a cAdvisor/Kubelet ServiceMonitor for collecting metrics. Previously, this had to be done by using additional manually-created scrape configs with custom `kubernetes_sd_configs` rules. 

Fixes #885.

#### Notes to the Reviewer
This copies the Prometheus Operator default of not providing this behavior OOTB and _must_ be configured by passing the flag. cc @hjet for the Helm charts.

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/cmd/agent-operator/agent-example-config.yaml
+++ b/cmd/agent-operator/agent-example-config.yaml
@@ -102,6 +102,99 @@ spec:
   pipelineStages:
     - cri: {}
 
+---
+
+# With -kubelet-service=default/kubelet provided as a flag to the Grafana Agent
+# Operator, it will maintain a Service called "kubelet" in the default namespace
+# with one endpoint per Node. This allows using the ServiceMonitor below to
+# monitor Kubernetes itself.
+
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  labels:
+    app.kubernetes.io/name: kubelet
+    instance: primary
+  name: kubelet
+  namespace: default
+spec:
+  endpoints:
+  - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+    honorLabels: true
+    interval: 30s
+    metricRelabelings:
+    - action: drop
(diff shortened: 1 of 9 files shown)


rfratto:
I tested this out with the local operator environment and it seems to work well, pulling in 25k metrics using the example ServiceAccount from [kube-prometheus](https://github.com/prometheus-operator/kube-prometheus/blob/0df510d1fa7a3cb88e95a83796a8ac241698b7fc/manifests/kubernetes-serviceMonitorKubelet.yaml).

cmd/agent-operator/agent-example-config.yaml
```diff
@@ -139,6 +232,7 @@ rules:
   resources:
   - nodes
   - nodes/proxy
+  - nodes/metrics
```
rfratto, line 235: This change, along with adding `/metrics/cadvisor`, should be reflected in the operator docs, but that is already covered by #923 so I'll let that get merged instead.

rfratto:
This is ready for review, though merging is blocked for the v0.19.0 release.

pkg/operator/kubelet.go
```diff
@@ -0,0 +1,151 @@
...
+		ObjectMeta: meta_v1.ObjectMeta{
+			Name:      r.kubeletName,
+			Namespace: r.kubeletNamespace,
+			Labels:    labels,
+		},
+		Subsets: []core_v1.EndpointSubset{{
+			Addresses: nodeAddrs,
+			Ports: []core_v1.EndpointPort{
+				// Taken from https://github.com/prometheus-operator/prometheus-operator/blob/2c81b0cf6a5673e08057499a08ddce396b19dda4/pkg/prometheus/operator.go#L593
+				{Name: "https-metrics", Port: 10250},
```
mattdurham, line 71: Are these ports guaranteed?

mattdurham (approved):
some of the configuration is over my head but looks good

pkg/operator/kubelet.go
rfratto: I don't think so. They're the defaults but you can technically change each Kubelet to have a different one. I'm not sure there's a good way to support that, and I also can't find a way to access it from within the cluster.

mattdurham: Seems good enough and we can always see what responses we get

PR #912, opened 2021-09-14
Ensure reshards are scheduled every reshard_interval

rfratto:


pkg/metrics/cluster/config_watcher.go
```diff
@@ -115,6 +116,24 @@ func (w *configWatcher) run(ctx context.Context) {
...
+	w.mut.Unlock()
+
+	now := time.Now()
+	if now.After(nextReshard) || now.Equal(nextReshard) {
+		ch := make(chan time.Time, 1)
+		ch <- now
+		return ch
+	}
+
+	remaining := time.Until(nextReshard)
```
mattdurham, line 133: Given some of our issues do you think having some additional logging here on remaining time would be good.

mattdurham (approved):
LGTM with one comment

PR #914, opened 2021-09-14
Merge v0.18.4

rfratto:


rfratto:
Release testing looks good. The config-sycner reproduces the case of events being handled every 5 minutes:

image

And despite this, reshards continue to happen nearly exactly every 60s, regardless of the extra events being handled without any gradual drift:

image

(The spike in reshards is from when I restarted the cluster)

mattdurham (approved):
LGTM

PR #920, opened 2021-09-14
Making configurable what method is used when adding k/v to spans

mapno:
#### PR Description 

Making configurable what method is used when adding k/v to spans. Options are `update`, `insert` and `upsert`

Fixes https://github.com/grafana/agent/issues/795

#### PR Checklist

- [ ] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,9 @@ for specific instructions.
 
 - [ENHANCEMENT] Update jsonnet-libs to 1.21 for Kubernetes 1.21+ compatability. (@MurzNN)
 
+- [ENHANCEMENT] Make method used to add k/v to spans in prom_sd processor
+  configurable. (@mapno)
+
 - [BUGFIX] Regex capture groups like `${1}` will now be kept intact when
   using `-config.expand-env`. (@rfratto)
 
--- a/cmd/agent/service.go
+++ b/cmd/agent/service.go
@@ -1,3 +1,4 @@
+//go:build !windows
 // +build !windows
 
 package main
--- a/cmd/agent/service.go
+++ b/cmd/agent/service.go
@@ -1,4 +1,3 @@
-//go:build !windows
 // +build !windows
 
 package main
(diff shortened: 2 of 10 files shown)


docs/configuration/tempo-config.md
```diff
@@ -126,11 +126,17 @@ automatic_logging:
...
-# configs have their __address__ matched against the ip on incoming spans. If a
-# match is found then relabeling rules are applied.
-scrape_configs:
-  - [<scrape_config>]
+prom_sd:
+  # A list of prometheus scrape configs.  Targets discovered through these scrape
+  # configs have their __address__ matched against the ip on incoming spans. If a
+  # match is found then relabeling rules are applied.
+  scrape_configs:
+    - [<scrape_config>]
```
rfratto, line 134: imo I'd prefer not nesting scrape_configs so it feels similar to the hierarchy used by logs / metrics. Can `sd_operation_type` be another top-level key maybe?

mapno: Yeah, that's the other option I considered. I chose this approach for consistency with other processors' configs, but I'm not a huge fan of another breaking change. I'll change it 👍

mapno: Changed in ea2ccc0b279d79aad4855954debe2c0d2c8d5821

pkg/tempo/config.go
```diff
@@ -444,9 +445,14 @@ func (c *InstanceConfig) otelConfig() (*config.Config, error) {
 	processors := map[string]interface{}{}
 	processorNames := []string{}
 	if c.ScrapeConfigs != nil {
+		opType := promsdprocessor.OperationTypeUpsert
```
joe-elliott, line 448: is this validated anywhere?

mapno: Added in ba939c7f

joe-elliott (approved):

PR #925, opened 2021-09-20
Convert in k8s RBAC api calls version from /v1beta1 to stable /v1

MurzNN:
#### PR Description 

This PR converts old RBAC resources endpoints from
```
apiVersion: rbac.authorization.k8s.io/v1beta1
```
to
```
apiVersion: rbac.authorization.k8s.io/v1
```

Installing Grafana Agent on Kubernetes versions >= 1.22 using instruction https://grafana.com/docs/grafana-cloud/quickstart/agent-k8s/k8s_agent_metrics/?src=blog&camp=timeshift_51#step-1-deploy-grafana-agent-resources fails with those errors:
```
serviceaccount/grafana-agent created
deployment.apps/grafana-agent created
unable to recognize "STDIN": no matches for kind "ClusterRole" in version "rbac.authorization.k8s.io/v1beta1"
unable to recognize "STDIN": no matches for kind "ClusterRoleBinding" in version "rbac.authorization.k8s.io/v1beta1"
```

From the documentation: RBAC resources

The rbac.authorization.k8s.io/v1beta1 API version of ClusterRole, ClusterRoleBinding, Role, and RoleBinding is no longer served as of v1.22.

Migrate manifests and API clients to use the rbac.authorization.k8s.io/v1 API version, available since v1.8.
All existing persisted objects are accessible via the new APIs

So the solution is to convert API calls from beta suffix to stable.

#### PR Checklist

- [x] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated


BitProcessor:
Link to relevant k8s docs: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#rbac-resources-v122

rfratto:
Thanks for making these changes! These files are generated from jsonnet and will be overwritten. We likely need to bump up the version of the k8s jsonnet libs we're using so it will generate this instead. If you're not sure how to do that, I can add a commit on top of this PR later today. 

In the meantime, can you modify the CHANGELOG and credit this as an enhancement? Be sure to give yourself credit for the change :)

MurzNN:
As I understand, we must bump the version in this https://github.com/grafana/agent/blob/main/production/kubernetes/build/lib/k.libsonnet file to something like `github.com/jsonnet-libs/k8s-alpha/1.21/main.libsonnet`, right?

And about changelog - I can simply add the text line to the original file, without any special procedure?

rfratto:
> As I understand, we must bump the version in this https://github.com/grafana/agent/blob/main/production/kubernetes/build/lib/k.libsonnet file to something like github.com/jsonnet-libs/k8s-alpha/1.21/main.libsonnet, right?

Right, the general procedure would be:

1. Update the dependency in the [jsonnetfile.json](https://github.com/grafana/agent/blob/main/production/kubernetes/build/jsonnetfile.json)
2. Run `jb install` (link to [jsonnet-bundler](https://github.com/jsonnet-bundler/jsonnet-bundler) which provides the command)
3. Update the ref in [the file you linked](https://github.com/grafana/agent/blob/main/production/kubernetes/build/lib/k.libsonnet) to use the new dependency version. 
4. Run `make example-kubernetes` (depends on [Tanka](https://github.com/grafana/tanka) being installed) to regenerate the manifests. 

Please let me know if you run into any trouble with this. I know jsonnet is complex and I'm happy to help or make the change if it proves to be too annoying for you.

> And about changelog - I can simply add the text line to the original file, without any special procedure?

Yes :) The changelog is just a plain text file at the moment.

MurzNN:
I've updated the version, also switched from archived alpha to stable version of jsonnet-libs repo.
Hope I done all right :)

MurzNN:
Also the folder [production/kubernetes/build/vendor](https://github.com/grafana/agent/tree/main/production/kubernetes/build/vendor), that presents in git repo, confuses me, shouldn't he be in gitignore?

rfratto:
> Also the non-empty folder production/kubernetes/build/vendor, that presents in git repo, confuses me, shouldn't it be in gitignore?

You can check in changes to that vendor folder for now. We used to check in all vendor folders, but recently stopped doing that for Go code. We still do it for Jsonnet for now.

When you push the changes from the vendor folder, I'll 👍 this and merge :)

MurzNN:
I've forcely added modified files from `vendor` folder, cuz it is already in .gitignore:
```
agent$ git add production/kubernetes/build/vendor
The following paths are ignored by one of your .gitignore files:
production/kubernetes/build/vendor
hint: Use -f if you really want to add them.
hint: Turn this message off by running
hint: "git config advice.addIgnoredFile false"
```
Now it's ok?

rfratto (approved):
LGTM, thank you!

I can't find a .gitignore here that's excluding it, do you have a global gitignore, maybe? 🤔

rfratto:
Ah, you're right. Thanks, definitely a silly mistake on my part.

rfratto:
I opened #932 to be more explicit about what paths we're ignoring; thanks for the tip!

PR #927, opened 2021-09-20
[dev.scraping-service-scalability] Fork the cortex ring/kv package for use with the configstore

rfratto:
#### PR Description 
The cortex kv package is focused heavily on a hash ring, making it increasingly inappropriate to use as a KV store for the configstore, which has different expectations about how it will be consumed. This initial PR does a basic copy
of the cortex package (from the specific version of Cortex we're already depending on) into configstore.

Initial changes from the Cortex package:
- Memberlist is removed
- cortex_ metrics have been reprefixed with agent_

This is a medium-term fix. I do not anticipate that the configstore package will exist forever, but we need a way to make it performant for now until we have something long-term in place.

This also unveiled a bug in updating the cluster config. Oops!

#### Notes to the Reviewer
This PR also immediately uses the new package, which doesn't require much except to point at the new location.

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,9 @@ for specific instructions.
 
 - [BUGFIX] Fix warn-level logging of dropped targets. (@james-callahan)
 
+- [BUGFIX] Reloading the scraping service kvstore config for loading instance
+  configs will no longer use the clustering config instead. (@rfratto)
+
 - [CHANGE] Breaking change: reduced verbosity of tracing autologging
   by not logging `STATUS_CODE_UNSET` status codes. (@mapno)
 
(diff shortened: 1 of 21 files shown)


56quarters:
You probably want to fork the one in dskit which has recently had some work done to strip down its dependencies: https://github.com/grafana/dskit/pull/43

rfratto:
> You probably want to fork the one in dskit which has recently had some work done to strip down its dependencies: grafana/dskit#43

Thanks for the suggestion! I had considered this, but even if nothing has really changed, I want to be extra certain (for now) that we remain as close as possible to the existing code. I'll definitely want to revisit this after 0.19.0 and see if we can pull over stuff from dskit instead.

rfratto:
I'm a little surprised the macOS tests are failing since I didn't really change anything. I'll look into it more tomorrow morning, but otherwise this should be a pretty small thing to review.

mattdurham (approved):
LGTM

PR #930, opened 2021-09-21
[dev.scraping-service-scalability] Batch the applying of configurations while running in scraping service mode

mattdurham:
#### PR Description 

This batches the configurations, so instead of handling them one at a time which at >1k configs can have a significant performance and scalability hit, this batches them and lessons that hit. 

#### PR Checklist

- [X] CHANGELOG updated 
- [NA] Documentation added
- [X] Tests updated

### DIFF
--- a/pkg/metrics/instance/batchapplyerror.go
+++ b/pkg/metrics/instance/batchapplyerror.go
@@ -1,0 +1,63 @@
+package instance
+
+import "fmt"
+
+// BatchApplyError contains the failed configs with error, successful configurations, and then if there is an error
+// that is not tied to a specific config then NonConfigError contains it
+type BatchApplyError struct {
+	Failed []BatchFailure
+	// NonConfigError is used when the error is not with a specific configuration, but likely the grouping
+	// or applying the group.
+	NonConfigError error
+}
+
+// BatchFailure for a given config has the associated error
+type BatchFailure struct {
+	Err    error
+	Config Config
+}
+
+// CreateBatchApplyErrorOrNil will create an error if failed has > 0 elements on nonConfigError is not nil, else will
+// return nil
+func CreateBatchApplyErrorOrNil(failed []BatchFailure, nonConfigError error) *BatchApplyError {
+	if len(failed) == 0 && nonConfigError == nil {
+		return nil
+	}
+	return &BatchApplyError{
+		Failed:         failed,
+		NonConfigError: nonConfigError,
+	}
+}
+
+// FindSuccessfulConfigs will take the list of all configs and compare them against the failed
+// return the successful based on name
+func FindSuccessfulConfigs(e *BatchApplyError, allConfigs []Config) []Config {
+	if e == nil || len(e.Failed) == 0 {
+		return allConfigs
+	}
+	succeeded := make([]Config, 0)
+	// Need to get the list of successful configs that were applied
+	for _, c := range allConfigs {
+		var didFail = false
+		for _, failed := range e.Failed {
+			if failed.Config.Name == c.Name {
+				didFail = true
+				break
+			}
+
+		}
+		if didFail {
+			continue
+		}
+		succeeded = append(succeeded, c)
+	}
+
+	return succeeded
+}
+
+func (e *BatchApplyError) Error() string {
+	if len(e.Failed) == 0 && e.NonConfigError == nil {
+		return ""
+	}
+	return fmt.Sprintf("%d configs failed to apply. Non config error %s ", len(e.Failed), e.NonConfigError)
+}
--- a/pkg/metrics/instance/batchapplyerror.go
+++ b/pkg/metrics/instance/batchapplyerror.go
@@ -13,8 +13,8 @@ type BatchApplyError struct {
 
 // BatchFailure for a given config has the associated error
 type BatchFailure struct {
-	Err    error
-	Config Config
+	Err        error
+	ConfigName string
 }
 
 // CreateBatchApplyErrorOrNil will create an error if failed has > 0 elements on nonConfigError is not nil, else will
@@ -40,7 +40,7 @@ func FindSuccessfulConfigs(e *BatchApplyError, allConfigs []Config) []Config {
 	for _, c := range allConfigs {
 		var didFail = false
 		for _, failed := range e.Failed {
-			if failed.Config.Name == c.Name {
+			if failed.ConfigName == c.Name {
 				didFail = true
 				break
 			}
--- a/pkg/metrics/instance/batchapplyerror.go
+++ b/pkg/metrics/instance/batchapplyerror.go
@@ -19,7 +19,7 @@ func FindSuccessfulConfigs(e *BatchApplyError, allConfigs []Config) []Config {
 	if e == nil || len(e.Failed) == 0 {
 		return allConfigs
 	}
-	succeeded := make([]Config, 0)
+	var succeeded []Config
 	// Need to get the list of successful configs that were applied
 	for _, c := range allConfigs {
 		var didFail = false
--- a/pkg/metrics/instance/batchapplyerror.go
+++ b/pkg/metrics/instance/batchapplyerror.go
@@ -19,7 +19,7 @@ type BatchFailure struct {
 
 // CreateBatchApplyErrorOrNil will create an error if failed has > 0 elements on nonConfigError is not nil, else will
 // return nil
-func CreateBatchApplyErrorOrNil(failed []BatchFailure, nonConfigError error) *BatchApplyError {
+func CreateBatchApplyErrorOrNil(failed []BatchFailure, nonConfigError error) error {
 	if len(failed) == 0 && nonConfigError == nil {
 		return nil
 	}
@@ -55,7 +55,7 @@ func FindSuccessfulConfigs(e *BatchApplyError, allConfigs []Config) []Config {
 	return succeeded
 }
 
-func (e *BatchApplyError) Error() string {
+func (e BatchApplyError) Error() string {
 	if len(e.Failed) == 0 && e.NonConfigError == nil {
 		return ""
 	}
--- a/pkg/metrics/instance/batchapplyerror.go
+++ b/pkg/metrics/instance/batchapplyerror.go
@@ -15,7 +15,7 @@ type BatchFailure struct {
 
 // FindSuccessfulConfigs will take the list of all configs and compare them against the failed
 // return the successful based on name
-func FindSuccessfulConfigs(e *BatchApplyError, allConfigs []Config) []Config {
+func (e *BatchApplyError) FindSuccessfulConfigs(allConfigs []Config) []Config {
 	if e == nil || len(e.Failed) == 0 {
 		return allConfigs
 	}
--- a/pkg/metrics/instance/batchapplyerror.go
+++ b/pkg/metrics/instance/batchapplyerror.go
@@ -2,31 +2,15 @@ package instance
 
 import "fmt"
 
-// BatchApplyError contains the failed configs with error, successful configurations, and then if there is an error
-// that is not tied to a specific config then NonConfigError contains it
+// BatchApplyError contains the failed configs with error, successful configurations
 type BatchApplyError struct {
 	Failed []BatchFailure
-	// NonConfigError is used when the error is not with a specific configuration, but likely the grouping
-	// or applying the group.
-	NonConfigError error
 }
 
 // BatchFailure for a given config has the associated error
 type BatchFailure struct {
 	Err        error
 	ConfigName string
-}
-
-// CreateBatchApplyErrorOrNil will create an error if failed has > 0 elements on nonConfigError is not nil, else will
-// return nil
-func CreateBatchApplyErrorOrNil(failed []BatchFailure, nonConfigError error) error {
-	if len(failed) == 0 && nonConfigError == nil {
-		return nil
-	}
-	return &BatchApplyError{
-		Failed:         failed,
-		NonConfigError: nonConfigError,
-	}
 }
 
 // FindSuccessfulConfigs will take the list of all configs and compare them against the failed
@@ -56,8 +40,8 @@ func FindSuccessfulConfigs(e *BatchApplyError, allConfigs []Config) []Config {
 }
 
 func (e BatchApplyError) Error() string {
-	if len(e.Failed) == 0 && e.NonConfigError == nil {
+	if len(e.Failed) == 0 {
 		return ""
 	}
-	return fmt.Sprintf("%d configs failed to apply. Non config error %s ", len(e.Failed), e.NonConfigError)
+	return fmt.Sprintf("%d configs failed to apply", len(e.Failed))
 }
--- a/pkg/metrics/instance/group_manager.go
+++ b/pkg/metrics/instance/group_manager.go
@@ -271,17 +271,7 @@ func (m *GroupManager) ListConfigs() map[string]Config {
 // exists, the group will have its settings merged with the Config and
 // will be updated.
 func (m *GroupManager) ApplyConfig(c Config) error {
-	cfgs := []Config{c}
-	err := m.ApplyConfigs(cfgs)
-	var bae BatchApplyError
-
-	if err != nil && errors.As(err, &bae) {
-		if len(bae.Failed) > 0 {
-			return bae.Failed[0].Err
-		}
-	}
-
-	return nil
+	return m.ApplyConfigs([]Config{c})
 }
 
 // DeleteConfig will remove a Config from its associated group. If there are
--- a/pkg/metrics/instance/group_manager.go
+++ b/pkg/metrics/instance/group_manager.go
@@ -101,9 +101,9 @@ func (m *GroupManager) applyConfigs(configs []Config, isRollback bool) error {
 	}
 	mergedHashes := make(map[string]int)
 	// The whole batch fails or succeeds here, so precompute if failures happen
-	var failedConfigs []BatchFailure
+	var groupFailedConfigs []BatchFailure
 	for _, c := range configs {
-		failedConfigs = append(failedConfigs, BatchFailure{
+		groupFailedConfigs = append(groupFailedConfigs, BatchFailure{
 			Err:        errors.New("failed during batch apply"),
 			ConfigName: c.Name,
 		})
@@ -115,7 +115,7 @@ func (m *GroupManager) applyConfigs(configs []Config, isRollback bool) error {
 			return err
 		} else if err != nil {
 			m.rollbackConfigs(oldConfigs)
-			return &BatchApplyError{Failed: failedConfigs}
+			return &BatchApplyError{Failed: groupFailedConfigs}
 		}
 		newHash, err := createMergedConfigHash(mergedConfig)
 
@@ -125,7 +125,7 @@ func (m *GroupManager) applyConfigs(configs []Config, isRollback bool) error {
 			return err
 		} else if err != nil {
 			m.rollbackConfigs(oldConfigs)
-			return &BatchApplyError{Failed: failedConfigs}
+			return &BatchApplyError{Failed: groupFailedConfigs}
 		}
 
 		// If we have the exact same config no need to reload it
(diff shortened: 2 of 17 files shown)


pkg/metrics/instance/manager.go
```diff
@@ -50,6 +50,9 @@ type Manager interface {
 	// one with Config.Name already exists.
 	ApplyConfig(Config) error
 
+	// ApplyConfigs batch applies the config
+	ApplyConfigs([]Config) (err error, successful []Config, failed []Config)
```
rfratto, line 54: This is a bit weird for a Go return signature. You could use custom error types here:

```go
type BatchApplyError struct {
   Failed map[Config]error 
   Successful []Config
}

func (e BatchApplyError) Error() string {
    return fmt.Sprintf("%d configs failed to apply", len(e.Failed)) 
} 
```

(Though I'm not sure if it really needs to return successful configs, since you could compute that if you really needed it)

Then you could use `errors.As` to unpack the error:

```go
var bae BatchApplyError 
if bae := BatchApplyError{}; errors.As(err, &bae) {
  // do something if needed 
}
```

pkg/traces/remotewriteexporter/exporter_test.go
```diff
@@ -85,6 +85,29 @@ type mockManager struct {
...
+			successful = append(successful, v)
+		}
+	}
+	return instance.BatchApplyResult{
+		Failed:         failed,
+		Successful:     successful,
+		NonConfigError: nil,
+	}
+
+}
```
rfratto, line 109: Is it possible for the remotewriteexporter to use the existing mock manager in the instance package?

pkg/metrics/instance/group_manager.go
```diff
@@ -53,11 +176,12 @@ func (g groupedConfigs) Copy() groupedConfigs {
 
 // NewGroupManager creates a new GroupManager for combining instances of the
 // same "group."
-func NewGroupManager(inner Manager) *GroupManager {
+func NewGroupManager(inner Manager, log log.Logger) *GroupManager {
```
rfratto, line 179: nit: can you put log first? We generally put registerers/logs as the first argument.

pkg/metrics/instance/manager.go
mattdurham: Cool! Created an object to hold the result, note the nonconfigerror isn't used now but will be once the configwatcher batch change goes in.

rfratto (commented):
quick first round of feedback, I'll follow up with a close review of GroupManager.ApplyConfigs shortly

pkg/traces/remotewriteexporter/exporter_test.go
mattdurham: It uses some non exported fields, which could probably be rewritten.  Would rather do in a separate PR.

pkg/metrics/instance/batchapplyerror.go
```diff
@@ -0,0 +1,63 @@
...
+	Failed []BatchFailure
+	// NonConfigError is used when the error is not with a specific configuration, but likely the grouping
+	// or applying the group.
+	NonConfigError error
+}
+
+// BatchFailure for a given config has the associated error
+type BatchFailure struct {
+	Err    error
+	Config Config
```
rfratto, line 17: nit: I'd put the Config name here instead of moving the entire Config around

pkg/metrics/instance/group_manager.go
```diff
@@ -30,11 +37,152 @@ type GroupManager struct {
...
+		}
+		groupsInConfigs[groupName][c.Name] = c
+		groupLookup[c.Name] = groupName
+		activeConfigs = append(activeConfigs, c)
+	}
+	// If we have group that no longer exist then we need to delete them
+	for groupName := range m.groups {
+		if _, exists := groupsInConfigs[groupName]; !exists {
+			if err := m.inner.DeleteConfig(groupName); err != nil {
+				level.Error(m.log).Log("err", fmt.Sprintf("failed to delete group named  %s with error %s", groupName, err))
```
rfratto, line 103: I think this should probably be a more fatal error. Recover and panic if recovery fails

pkg/metrics/instance/group_manager.go
```diff
@@ -30,11 +37,152 @@ type GroupManager struct {
...
+			}
+		}
+	}
+	groups := make(map[string]groupedConfigs)
```
rfratto, line 107: This ends up being a copy of `groupsInConfigs`. Can't we just use that instead?

pkg/metrics/instance/group_manager.go
```diff
@@ -30,11 +37,152 @@ type GroupManager struct {
...
+		newHash, err := createMergedConfigHash(mergedConfig)
+		if err != nil {
+			level.Error(m.log).Log("err", fmt.Sprintf("failed to create merged hash named  %s with error %s", groupName, err))
+			continue
+		}
+		groups[groupName] = grouped
+		mergedHashes[newHash] = 0
+		if _, exists := m.mergedConfigHashes[newHash]; exists {
+			continue
+		}
```
rfratto, line 125: This would cause unexpected issues if two groups happen to have the same hash. We should find a way to associate the hash with the group.

pkg/metrics/instance/group_manager.go
```diff
@@ -30,11 +37,152 @@ type GroupManager struct {
...
+		// always has a valid configuration. If we are already in a rollback then dont try again
+		if err != nil && !isRollback {
+			m.rollbackConfigs(oldConfigs)
+			return CreateBatchApplyErrorOrNil(failed, err)
+		}
+		err = m.inner.ApplyConfig(mergedConfig)
+		if err != nil && !isRollback {
+			m.rollbackConfigs(oldConfigs)
+			return CreateBatchApplyErrorOrNil(failed, err)
+		}
```
rfratto, line 137: I'm not sure this is right: if we're inside of a rollback and one of these calls fails, it _should_ be a panic, but it won't because the error is effectively ignored.

pkg/metrics/instance/batchapplyerror.go
```diff
@@ -0,0 +1,63 @@
...
+
+import "fmt"
+
+// BatchApplyError contains the failed configs with error, successful configurations, and then if there is an error
+// that is not tied to a specific config then NonConfigError contains it
+type BatchApplyError struct {
+	Failed []BatchFailure
+	// NonConfigError is used when the error is not with a specific configuration, but likely the grouping
+	// or applying the group.
+	NonConfigError error
```
rfratto, line 11: This seems like a different error. I think BatchApplyError should only be responsible for indicating that the grouping was successful, but individual configs had issues. 

If at any point the grouping fails, we recover anyway, so returning both doesn't have a ton of value here.

pkg/metrics/instance/batchapplyerror.go
```diff
@@ -0,0 +1,63 @@
...
+
+// BatchFailure for a given config has the associated error
+type BatchFailure struct {
+	Err    error
+	Config Config
+}
+
+// CreateBatchApplyErrorOrNil will create an error if failed has > 0 elements on nonConfigError is not nil, else will
+// return nil
+func CreateBatchApplyErrorOrNil(failed []BatchFailure, nonConfigError error) error {
```
rfratto, line 22: Given my comment w/r/t the struct, we should be able to remove this function an inline the call. At the very least, this shouldn't be an exposed function anyway.

pkg/metrics/instance/batchapplyerror.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	return &BatchApplyError{
+		Failed:         failed,
+		NonConfigError: nonConfigError,
+	}
+}
+
+// FindSuccessfulConfigs will take the list of all configs and compare them against the failed
+// return the successful based on name
+func FindSuccessfulConfigs(e *BatchApplyError, allConfigs []Config) []Config {
```
rfratto, line 34: IMO make this a method on the error struct.

pkg/metrics/instance/batchapplyerror.go
```diff
@@ -0,0 +1,63 @@
...
+		succeeded = append(succeeded, c)
+	}
+
+	return succeeded
+}
+
+func (e BatchApplyError) Error() string {
+	if len(e.Failed) == 0 && e.NonConfigError == nil {
+		return ""
+	}
```
rfratto, line 61: (This will never be true)

pkg/metrics/instance/batchapplyerror.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+// FindSuccessfulConfigs will take the list of all configs and compare them against the failed
+// return the successful based on name
+func FindSuccessfulConfigs(e *BatchApplyError, allConfigs []Config) []Config {
+	if e == nil || len(e.Failed) == 0 {
+		return allConfigs
+	}
+	succeeded := make([]Config, 0)
```
rfratto, line 38: nit: don't use make for slices unless you want to give it a non-zero length:

```var succeeded []Config```

File: pkg/metrics/instance/group_manager.go [hunk 14]
```diff
@@ -30,11 +37,152 @@ type GroupManager struct {
...
+
 	// groups is a map of group name to the grouped configs.
 	groups map[string]groupedConfigs
 
 	// groupLookup is a map of config name to group name.
 	groupLookup map[string]string
+
+	mergedConfigHashes map[string]int
+
+	log log.Logger
```
rfratto, line 51: nit: can you group these so there's no empty lines in between the members being protected by `mtx`? If you want a comment for a member, you can do it on the right:

```go
mtx sync.Mutex
activeConfigs []Config // List of individual configs currently applied
// etc
```

pkg/metrics/instance/group_manager.go
```diff
@@ -30,11 +37,152 @@ type GroupManager struct {
...
+	return m.applyConfigs(configs, false)
+}
+
+// applyConfigs is used to take the currently running set of configs, plus the new ones. (update/append configs).
+// isRollback determines whether we are rolling back any changes and if so avoid recursion in calling rollback again
+func (m *GroupManager) applyConfigs(configs []Config, isRollback bool) error {
+	if len(configs) == 0 {
+		return nil
+	}
+	failed := make([]BatchFailure, 0)
```
rfratto, line 67: nit: don't use `make` for slices with a non-zero size. You can also just have this be a BatchApplyError:

```go
var bae BatchApplyError 

// ... 
bae.Failed = append(bae.Failed, BatchFailure{Config: c.Name, Err: err})
```

pkg/metrics/instance/group_manager.go
```diff
@@ -30,11 +37,152 @@ type GroupManager struct {
...
+	// This will form our master set of configurations to apply by their group, this includes any that are currently
+	// running but not in the configs parameter.
+	groupsInConfigs := make(map[string]groupedConfigs)
+
+	// Combined group of current and new configs
+	combinedConfigs := m.getCombinedConfigs(configs)
+
+	oldConfigs := make([]Config, 0)
+	// In case of an error in applying the new configs, we need to grab the old config for rollback
+	copy(oldConfigs, m.activeConfigs)
```
rfratto, line 78: `m.configs` isn't modified at this point, so we can pass that to the rollback instead of making a copy here.

pkg/metrics/instance/group_manager.go
```diff
@@ -30,11 +37,152 @@ type GroupManager struct {
...
+	activeConfigs := make([]Config, 0)
```
rfratto, line 79: nit: make with empty slice

pkg/metrics/instance/group_manager.go
```diff
@@ -30,11 +37,152 @@ type GroupManager struct {
...
+	groupLookup := make(map[string]string)
+	// Iterate through the combined configurations
+	for _, c := range combinedConfigs {
+		groupName, err := hashConfig(c)
+		if err != nil {
+			failed = append(failed, BatchFailure{
+				Err:    err,
+				Config: c,
+			})
+			level.Error(m.log).Log("err", fmt.Sprintf("failed to get group name for config %s: %s", c.Name, err))
```
rfratto, line 89: nit: add config name as a field here to make writing fancy loqql metrics queries a little easier:

```
level.Error(m.log).Log("msg", "failed to get group name for config", "config", c.Name, "err", err)
```

pkg/metrics/instance/group_manager.go
rfratto: nit: use field names instead of just a giant error message

pkg/metrics/instance/group_manager.go
```diff
@@ -30,11 +37,152 @@ type GroupManager struct {
...
+		}
+	}
+	groups := make(map[string]groupedConfigs)
+	mergedHashes := make(map[string]int)
+	// Now that we have grouped all the new and existing configurations we can apply them
+	for groupName, grouped := range groupsInConfigs {
+		mergedConfig, err := groupConfigs(groupName, grouped)
+		if err != nil {
+			level.Error(m.log).Log("err", fmt.Sprintf("failed to group configs with groupname  %s with error %s", groupName, err))
+		}
```
rfratto, line 114: Ineffective assign to error here: it will be overwritten by the next call that assigns to err

pkg/metrics/instance/group_manager.go
```diff
@@ -30,11 +37,152 @@ type GroupManager struct {
...
+			m.rollbackConfigs(oldConfigs)
+			return CreateBatchApplyErrorOrNil(failed, err)
+		}
+	}
+	m.groups = groups
+	m.activeConfigs = activeConfigs
+	m.groupLookup = groupLookup
+	m.mergedConfigHashes = mergedHashes
+
+	return CreateBatchApplyErrorOrNil(failed, nil)
```
rfratto, line 144: nit: inline the creation of the error

pkg/metrics/instance/group_manager.go
```diff
@@ -30,11 +37,152 @@ type GroupManager struct {
...
+}
+
+func createMergedConfigHash(mergedConfig Config) (string, error) {
+	bb, err := yaml.Marshal(mergedConfig)
```
rfratto, line 148: nit: `yaml.Marshal` for configs has overhead because it creates a yaml.Map. Use `MarshalConfig` directly to reduce allocations.

pkg/metrics/instance/group_manager.go
```diff
@@ -30,11 +37,152 @@ type GroupManager struct {
...
+func (m *GroupManager) rollbackConfigs(oldConfigs []Config) {
+	// If restoring a config fails, we've left the Agent in a really bad
+	// state: the new config can't be applied and the old config can't be
+	// brought back. Just crash and let the Agent start fresh.
+	//
+	// Restoring the config _shouldn't_ fail here since applies only fail
+	// if the config is invalid. Since the config was running before, it
+	// should already be valid. If it does happen to fail, though, the
+	// internal state is left corrupted since we've completely lost a
+	// config.
```
rfratto, line 165: nit: move/reword this around a little so it's a comment on the method itself.

pkg/metrics/instance/group_manager.go
```diff
@@ -30,11 +37,152 @@ type GroupManager struct {
...
+	if err := m.applyConfigs(oldConfigs, true); err != nil {
+		level.Error(m.log).Log("err", fmt.Sprintf("failed to rollback configs with error %s", err))
+		panic(err)
+	}
+}
+
+func (m *GroupManager) getCombinedConfigs(newConfigs []Config) map[string]Config {
+	// There is also a known issue in that this DOES not handle deleted configs
+	combinedConfigsMap := make(map[string]Config)
```
rfratto, line 174: nit: do we want to allocate space for at _least_ `len(m.activeConfigs)` here?

pkg/metrics/instance/group_manager.go
```diff
@@ -104,89 +254,17 @@ func (m *GroupManager) ListConfigs() map[string]Config {
...
-		oldConfig, ok := m.groups[oldGroup][c.Name]
-		if !ok {
-			panic("failed to properly move config to new group. THIS IS A BUG!")
-		}
-
-		err = m.deleteConfig(c.Name)
-		if err != nil {
-			err = fmt.Errorf("cannot apply config %s because deleting it from the old group failed: %w", c.Name, err)
-			return
+	if err != nil && errors.As(err, &bae) {
```
rfratto, line 261: nit: `err != nil` can be dropped here, `errors.As` is no-op if err is nil

pkg/metrics/instance/group_manager_test.go
```diff
@@ -5,11 +5,13 @@ import (
...
 	"testing"
 
+	"github.com/go-kit/log"
+
 	"github.com/stretchr/testify/require"
 )
 
 func TestGroupManager_ListInstances_Configs(t *testing.T) {
-	gm := NewGroupManager(newFakeManager())
+	gm := NewGroupManager(log.NewNopLogger(), newFakeManager())
```
rfratto, line 14: nit: use `util.TestLogger(t)` here so we get logs when a test fails :)
(ditto for rest of tests)

pkg/metrics/instance/manager.go
```diff
@@ -89,6 +92,21 @@ type BasicManager struct {
...
+	for _, v := range configs {
+		err := m.ApplyConfig(v)
+		if err != nil {
+			failed = append(failed, BatchFailure{
+				Err:    err,
+				Config: v,
+			})
+		}
+	}
+	return CreateBatchApplyErrorOrNil(failed, nil)
```
rfratto, line 107: nit: inline call here

pkg/metrics/instance/manager.go
```diff
@@ -323,6 +341,20 @@ type MockManager struct {
...
+	for _, v := range configs {
+		if err := m.ApplyConfig(v); err != nil {
+			failed = append(failed, BatchFailure{
+				Err:    err,
+				Config: v,
+			})
+		}
+	}
+	return CreateBatchApplyErrorOrNil(failed, nil)
+}
```
rfratto, line 356: nit: this should use a `func()` instead of implementing the actual function

pkg/metrics/instance/modal_manager.go
```diff
@@ -68,6 +69,30 @@ type ModalManager struct {
...
+// ApplyConfigs is used to batch configurations for performance instead of the singular ApplyConfig
+func (m *ModalManager) ApplyConfigs(configs []Config) error {
+	m.mut.Lock()
+	defer m.mut.Unlock()
+
+	err := m.active.ApplyConfigs(configs)
+	var bae BatchApplyError
+	if err != nil {
+		errors.As(err, &bae)
+	}
```
rfratto, line 81: nit: `_ = errors.As(err, &bae)`  to make it clear we're just trying to blindly convert.

pkg/traces/remotewriteexporter/exporter_test.go
```diff
@@ -85,6 +85,22 @@ type mockManager struct {
...
+	for _, v := range configs {
+		err := m.ApplyConfig(v)
+		if err != nil {
+			failed = append(failed, instance.BatchFailure{
+				Err:    err,
+				Config: v,
+			})
+		}
+	}
+	return instance.CreateBatchApplyErrorOrNil(failed, nil)
```
rfratto, line 100: nit: this should call out to a mocked function instead of actually implementing something.

rfratto (commented):
thanks! mostly looks good, just a few nits and small comments around cleanup and allocations

pkg/metrics/instance/group_manager.go
mattdurham: Then we also have that concern with the hashConfig function also?

pkg/metrics/instance/batchapplyerror.go
mattdurham: I think it would be simpler if we assume that the batch either fails or succeeds as a block. Any failure rolls back the configs.  Then we can return a standard error.

rfratto: I don't think we should do that. If a user happens to push a bad config through, it would prevent reshards from ever succeeding until that one config is resolved; it'd essentially be a noisy neighbor problem.

pkg/metrics/instance/group_manager.go
rfratto: Yeah, I think you're right. In that case you do want hash collisions at least at the group level, but it's still possible that two incompatible configs have the same group hash.

rfratto:
This has fallen pretty far out of sync, let's close for now and revisit if we need to.

PR #878, opened 2021-09-01
Update metrics in mixin

kgeckhart:
#### PR Description 
I took a pass at the mixin to update referenced metrics which have changed overtime. I also moved the collection of samples queries in to the samples row.

#### PR Checklist

- [ na? ] CHANGELOG updated 
- [ na ] Documentation added
- [ na ] Tests updated

PR #931, opened 2021-09-21
Mixin Dashboard Updates

kgeckhart:
#### PR Description 
This PR has a handful of small changes to the mixin dashboards

- Adds a samples rate panel back to the remote write dashboard based on the newer prom one, https://github.com/prometheus/prometheus/blob/main/documentation/prometheus-mixin/dashboards.libsonnet#L143
- Flips `Created Series` back to `Appended Samples` (I changed this here, https://github.com/grafana/agent/pull/878/files#diff-0ffb78c3d3889793bebfb96a62bc7291cff9b69cc100278a96f16a6461de784c, because I didn't see the new metric)
- Partitions targets by pod on the main dashboard
- Adds the operational dashboard to the mixin
- All dashboards should be 1h range with a 30s refresh now

I tested out the changes using the k3d example

#### PR Checklist

- [ na ] CHANGELOG updated 
- [ na ] Documentation added
- [ na ] Tests updated

### DIFF
--- a/example/docker-compose/grafana/dashboards/agent-operational.json
+++ b/example/docker-compose/grafana/dashboards/agent-operational.json
@@ -7,7 +7,7 @@
    "graphTooltip": 0,
    "hideControls": false,
    "links": [ ],
-   "refresh": "10s",
+   "refresh": "1m",
    "rows": [
       {
          "collapse": false,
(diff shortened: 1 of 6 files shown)


rfratto (approved):
Thanks!

Issue #789, opened 2021-08-03
Potential inefficiency retrieving configurations from consul in scraping service mode


mattmendick:
Per an offline conversation with @mattdurham and @jdbaldry, there may be an inefficiency in how the agent's scraping service mode is retrieving scraping configs from consul. Currently it seems that the agent will perform one request to obtain the list of keys only, and then iterates over each of the keys to retrieve the value. As the agent has more and more configurations loaded, this will increase load on consul linearly to a breaking point of consul being unable to handle the requests.

Relevant code in consul's api code: https://github.com/hashicorp/consul/blob/main/api/kv.go#L85

This would be more efficient if we are able to do a single call to request the keys and values from consul so that all the configs are retrieved with a single call (or at least fewer than the n number of configurations that exist).

mattdurham:
Specifically, the Key-Value Pairs are transformed into an array of keys, then the values are retrieved later. This loss of information happens in the imported cortex consul client https://github.com/cortexproject/cortex/blob/master/pkg/ring/kv/consul/client.go#L337

PR #933, opened 2021-09-21
[dev.scraping-service-scalability] Move #802 fix to kv fork

rfratto:
#### PR Description 
#802 introduced a short-term fix for #789. Now that we're making a growing amount of changes to how we interact with the KV store, it's a good opportunity to move those changes to the fork of the KV package introduced in #927. 

This PR should be equivalent in performance to the Consul change, but now works for ETCD as well. 

#### Notes to the Reviewer
This is the first change to the kv interface to make `List` always return key value pairs instead of just keys.

#### PR Checklist
Need to update the CHANGELOG to reference that the metrics changed names again.

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/metrics/instance/configstore/remote.go
+++ b/pkg/metrics/instance/configstore/remote.go
@@ -2,17 +2,11 @@ package configstore
 
 import (
 	"context"
-	"errors"
 	"fmt"
-	"net/http"
 	"strings"
 	"sync"
 
 	"github.com/weaveworks/common/instrument"
-
-	"github.com/hashicorp/go-cleanhttp"
-
-	"github.com/hashicorp/consul/api"
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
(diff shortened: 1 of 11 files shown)


pkg/metrics/instance/configstore/remote.go
```diff
@@ -347,111 +268,32 @@ func (r *Remote) all(ctx context.Context, keep func(key string) bool) (<-chan in
...
-
-	var wg sync.WaitGroup
-	wg.Add(len(keys))
-	go func() {
-		wg.Wait()
-		close(ch)
-	}()
-
-	for _, key := range keys {
-		go func(key string) {
```
rfratto, line 428: @mattdurham I decided that your approach of iterating over all the keys in a single goroutine and then flushing to a channel will likely be more performant, especially if there's thousands of keys.

mattdurham (approved):
LGTM aside from that odd test error

PR #936, opened 2021-09-22
[dev.scraping-service-scalability] Catch delete events from kv.WatchPrefix

rfratto:
#### PR Description 
The Cortex KV package had ignore deleted events from etcd because other implementations didn't detect when a key had gone away. 

This PR detects deleted keys in Consul by keeping track of configs that go away in between List calls. 

#### Notes to the Reviewer
Note that there's no guarantee a delete event will be detected if there are multiple sequential calls to WatchPrefix. If a key is deleted in between calls, the subsequent call will be unaware that the key used to exist. I believe the same effect exists for the etcd watch though: the etcd watch only returns mutation events from the start of the watch, and keys may be deleted in between. 

Despite this, it still may be worthwhile to have this enhancement, since it's better than it was before and will detect most deletes. 

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/metrics/instance/configstore/kv/consul/client.go
+++ b/pkg/metrics/instance/configstore/kv/consul/client.go
@@ -265,10 +265,12 @@ func (c *Client) WatchKey(ctx context.Context, key string, f func(interface{}) b
 // Values in Consul are assumed to be JSON. This function blocks until the context is cancelled.
 func (c *Client) WatchPrefix(ctx context.Context, prefix string, f func(string, interface{}) bool) {
 	var (
-		backoff = util.NewBackoff(ctx, backoffConfig)
-		index   = uint64(0)
-		limiter = c.createRateLimiter()
+		backoff  = util.NewBackoff(ctx, backoffConfig)
+		index    = uint64(0)
+		limiter  = c.createRateLimiter()
+		lastKeys = map[string]struct{}{}
 	)
+
 	for backoff.Ongoing() {
 		err := limiter.Wait(ctx)
 		if err != nil {
@@ -288,6 +290,7 @@ func (c *Client) WatchPrefix(ctx context.Context, prefix string, f func(string, 
 		}
 
 		kvps, meta, err := c.kv.List(prefix, queryOptions.WithContext(ctx))
+
 		// kvps being nil here is not an error -- quite the opposite. Consul returns index,
 		// which makes next query blocking, so there is no need to detect this and act on it.
 		if err != nil {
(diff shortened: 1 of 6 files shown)


pkg/metrics/instance/configstore/kv/consul/mock.go
```diff
@@ -192,7 +192,12 @@ func (m *mockKV) List(prefix string, q *consul.QueryOptions) (consul.KVPairs, *c
 func (m *mockKV) Delete(key string, q *consul.WriteOptions) (*consul.WriteMeta, error) {
 	m.mtx.Lock()
 	defer m.mtx.Unlock()
+
+	// Deletes increment the ModifyIndex
 	delete(m.kvps, key)
+	m.current++
+	m.cond.Broadcast()
```
rfratto, line 199: I was very confused about why I couldn't get the delete to generate a new ModifyIndex since I could reproduce this perfectly locally with Consul. Turns out the mock consul here was accidentally missing this behavior.

pkg/metrics/instance/configstore/kv/consul/mock.go
```diff
@@ -181,8 +181,7 @@ func (m *mockKV) List(prefix string, q *consul.QueryOptions) (consul.KVPairs, *c
 
 	result := consul.KVPairs{}
 	for _, kvp := range m.kvps {
-		if strings.HasPrefix(kvp.Key, prefix) && kvp.ModifyIndex >= q.WaitIndex {
-			// unfortunately real consul doesn't do index check and returns everything with given prefix.
```
rfratto, line 185: This is another behavior of Consul that we're relying on also being true for the Mock client.

pkg/metrics/instance/configstore/kv/consul/client.go
```diff
@@ -302,7 +305,13 @@ func (c *Client) WatchPrefix(ctx context.Context, prefix string, f func(string,
 			continue
 		}
 
+		newKeys := make(map[string]struct{}, len(kvps))
+
 		for _, kvp := range kvps {
+			// Keep track of all returned keys. This *must* be done before the check
+			// below otherwise an unchanged key will be treated as deleted.
+			newKeys[kvp.Key] = struct{}{}
```
mattdurham, line 313: Do we also need to keep track of the keys in List? We might be good if the watch always executes before the list but relying on that feels not great.

rfratto: Hmm, not without also changing etcd to do the same. I think treating WatchPrefix as "you will only get CRUD events from the start of the watch" makes sense, though.

mattdurham (approved):
Looks good to me.

PR #942, opened 2021-09-28
Add note about using backslashes and backticks

MurzNN:
#### PR Description 
We need to note users about using backslashes in agent config file, and about not supported backticks too.
Also I've added error messages text for quicker finding the solution by error text.

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/configuration/logs-config.md
+++ b/docs/configuration/logs-config.md
@@ -67,9 +67,13 @@ scrape_configs:
 > * [`promtail.scrape_config`](https://github.com/grafana/loki/tree/master/docs/sources/clients/promtail#scrape_config)
 > * [`promtail.target_config`](https://github.com/grafana/loki/tree/master/docs/sources/clients/promtail#target_config)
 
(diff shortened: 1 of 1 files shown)


rfratto (commented):
Can we copy the note about escaping from the [regex stage](https://grafana.com/docs/loki/latest/clients/promtail/stages/regex/#regex-stage) before your note, and then make it clearer that your note only applies to regex inside the match stage?

rfratto:
I don't think it applies Agent-wide (you certainly don't have to do it for relabel rules). The double processing should only apply to `match`, and I believe it's from some combination of:

* how LogQL is being parsed
* `"` nested inside `'` that makes the escaping rules not very obvious

rfratto (approved):
LGTM though, will merge when it goes green. It looks like the CLA bot isn't working right now, but I know you've contributed before so I'll just force the merge through when everything else passes.

Issue #941, opened 2021-09-28
automatic_logging labels containing period cause loki to return 400 error


frankbohman:
Should it be possible to add for example service.name or span.kind as labels to index traces in loki ? 

Or perhaps have some way to rewrite them in the agent?

Loki return: Bad Request (400): error parsing labels: parse error at line 1, col 9: syntax error: unexpected ., expecting = or != or =~ or !~"

My configuration:

```yaml
automatic_logging:
          backend: logs_instance
          logs_instance_name: default
          overrides:
            service_key: app
          spans: true
          processes: true
          roots: true
          process_attributes:
            - service.name
          span_attributes:
            - span.kind
          labels:
            - service.name
            - span.kind
```

mapno:
Hi! Thanks for bringing this up.

Looked a bit around and this seems to be caused because Loki only accepts alphanumeric and "_" as valid characters in labels. Thus, `service.name` and `span.kind` are not _valid_ labels.

I've opened a [PR](https://github.com/grafana/agent/pull/951) with a proposal to fix this issue by replacing dots with underscores.

PR #951, opened 2021-09-29
Replace dots for underscores in autologged labels

mapno:
#### PR Description 

Replace dots for underscores in autologged labels.

Loki only accepts alphanumeric and "_" as valid characters for labels. Attributes containing dots are replaced for underscores

Fixes https://github.com/grafana/agent/issues/941

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,9 @@ for specific instructions.
 
 - [BUGFIX] Standardize scrape_interval to 1m in examples. (@mattdurham)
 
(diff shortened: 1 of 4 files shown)


rfratto:
Thanks Mario! We have a label sanitization function in https://github.com/grafana/agent/blob/ad0b4146ec1bfe9b8180b687991655a57f5e4a13/pkg/operator/config/utils.go#L130, maybe we should move that to util and re-use it here?

rfratto (approved):
LGTM :)

PR #953, opened 2021-09-30
Add note about ports in tempo remote_write enpoint

MurzNN:
Added comment with examples how to fill port of `remote_write` tempo endpoint for local and on-premises instances.
For new users it's unclear which port to choose, and in most cases they try to fill the default `3200` Tempo port, instead of gRPC receiver port.

#### PR Checklist

- [ ] CHANGELOG updated 
- [x] Documentation added
- [ ] Tests updated


mattdurham (approved):
Looks good, appreciated!

PR #957, opened 2021-10-01
docs: deploy main -> /next, release -> /latest and /<tag>

rfratto:
Deploys main commits to the /next folder to make it clearer that it's documenting unreleased software. Tags, when one is made, will be pushed to both that tag name and the /latest folder.

This will need a commit to manually make do the /next, /latest and /<tag> changes for v0.19.0

### DIFF
--- a/.github/workflows/publish_release.yml
+++ b/.github/workflows/publish_release.yml
@@ -1,0 +1,45 @@
+name: publish_release_docs
+
+on:
+  push:
+    tags:
+      - '*'
+
+jobs:
+  sync:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - run: git clone --single-branch --no-tags --depth 1 -b master https://grafanabot:${{ secrets.GH_BOT_ACCESS_TOKEN }}@github.com/grafana/website-sync ./.github/actions/website-sync
+
+    # Sync tag docs to versioned folder
+    - name: publish-tag-to-git
+      uses: ./.github/actions/website-sync
+      id: publish_tag
+      with:
+        repository: grafana/website
+        branch: master
+        host: github.com
+        github_pat: '${{ secrets.GH_BOT_ACCESS_TOKEN }}'
+        source_folder: docs
+        target_folder: 'content/docs/agent/${GITHUB_REF##*/}'
+    - shell: bash
+      run: |
+        test -n "${{ steps.publish_tag.outputs.commit_hash }}"
+        test -n "${{ steps.publish_tag.outputs.working_directory }}"
+
+    # Also send it to latest folder
+    - name: publish-latest-to-git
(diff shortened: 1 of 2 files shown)


.github/workflows/publish_release.yml
```diff
@@ -0,0 +1,45 @@
...
+    - name: publish-tag-to-git
+      uses: ./.github/actions/website-sync
+      id: publish_tag
+      with:
+        repository: grafana/website
+        branch: master
+        host: github.com
+        github_pat: '${{ secrets.GH_BOT_ACCESS_TOKEN }}'
+        source_folder: docs
+        target_folder: 'content/docs/agent/${GITHUB_REF##*/}'
```
rfratto, line 25: I _think_ this should work? 🤔 We used to use `${GITHUB_REF##*/}` for automating releases, but I never tried it with a action stage.

mattdurham (dismissed):
Looks good enough to see if it works ;)

GrafanaWriter:
I like the technical approach here, but we follow a different approach in Grafana OSS and Grafana Enterprise. For these docs, we maintain a /next directory for future content and a /latest directory for current content. We don't use tags for that purpose.

I'd prefer we follow a similar approach to avoid maintenance issues.

rfratto:
Thanks for the feedback! I'd also like to be consistent with how other projects handle this, and I believe Tempo and Loki are inline with what Grafana does as well.

What causes a deployment to next vs latest? Where are next/latest docs stored for Grafana?

GrafanaWriter:
For Grafana, we keep the docs in `latest` and `next`. `next` mirrors `main`, and `latest` mirrors the last published version. For example, `/docs/grafana/latest/` and `/docs/grafana/next/`

When we cut Beta for code, we don't update the docs in our website. When the code goes GA, we will switch.

Bug fixes are merged into `next`. To get it into `latest`, you backport. If the branch has been cut for a beta of the next major release, you must also backport to that branch.

Thus, `next` is really a bucket of future content.

GrafanaWriter:
I believe it is part of the release process. @achatterjee-grafana - can you respond, please?

achatterjee-grafana:
In the `website` repo, we have 2 directories for grafana, latest and next. You can follow a similar structure @rfratto for the agent docs. 
For context: https://docs.google.com/document/d/1jJ2pdNysgI2zt8Gf3nrEAMgTFcfAAQxAhm7G4ISmVsY/edit. Hope this helps!

rfratto:
It doesn't look like have access to that document, unfortunately. Along these lines, we proposed a shift in thinking around this a few months ago to make automation easier, and I'd love to get your opinions on it: https://docs.google.com/document/d/1vI9N5NEclR08uLTUVssxucpTGt6WjFyOLbBhxf2hz2o/edit 

(to external users reading: doc is internal-only at the moment since it's dependent on non-public code)

PR #965, opened 2021-10-06
Update mongodb_exporter to release-0.27.0-grafana

rfratto:
#### PR Description 
Updates mongodb_exporter to a new version in a Grafana fork. Note that upstream has since merged our changes, so the next release (0.29.0, presumable) can be done completely upstream without a replace directive.

#### Notes to the Reviewer
Also adds dependency update tracking for this exporter.

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated


rfratto:
Note that since this uses an upstream /metrics endpoint, it's losing the build info metric we force to be created for all other integrations, which might break some dashboards. We would need another approach if we wanted to add it back (like maybe pass registrars to the handler method to inject custom metrics)

Issue #883, opened 2021-09-02
Grafana Agent Operator: Add support for integrations


rfratto:
Grafana Agent Operator needs support for integrations for feature parity with Grafana Agent.

Design doc: https://docs.google.com/document/d/11BQeqpMkEwMaC69buESrVTAP6sM0UTmdUUrA8rWV4UQ/edit#heading=h.yuclk376lw30

rfratto:
There's a problem with this proposal: it's going to rely on #1059 but #1059 creates unpredictable metrics endpoints (i.e., the metrics endpoint is now determined by the instance label of an integration, which is calculated based both on integration's config). 

One way to work around this is to override the instance label for integrations to be set to the `AgentIntegration` CR namespace/name. I'm not sure if this is the nicest solution, though. 

I'm thinking this proposal might need to be re-written once #1059 is merged, and it might need more changes as prerequisites.

rfratto:
> There's a problem with this proposal: it's going to rely on #1059 but #1059 creates unpredictable metrics endpoints (i.e., the metrics endpoint is now determined by the instance label of an integration, which is calculated based both on integration's config).

#1133 would make this not a problem.

rfratto:
Superseded by #1224.

PR #968, opened 2021-10-06
[dev.operator-integrations-support] pkg/operator: create CRDs for integrations

rfratto:
This PR creates the basic outline for integration-specific CRDs. Opening in draft to get some feedback going.

One of my current concerns is settling on a way for CRs to be discovered for creating pods. If they are treated as root CRs in their own hierarchy, then we need to add a ton of options for them that control image/volumeMounts/tolerations/etc (but there would only be one deployment across the cluster). On the other hand, if we have them be a child of a GrafanaAgent hierarchy, then they have much less to configure (but they need to be deployed once per parent GrafanaAgent CR).


rfratto:
Related to #883.

PR #970, opened 2021-10-08
Merge of main into dev.servicegraphs

mapno:
#### PR Description

Merge of main into dev.servicegraphs


mapno:
I also rebased `dev.servicegraphs` into `main`. But that would require force pushing the branch, which I think is a riskier alternative, since it's not tracked. All the main commits here will disappear once `dev.servicegraphs` is merged into `main`.

PR #995, opened 2021-10-15
Attempting to get or delete a config which does not exist 404's

kgeckhart:
#### PR Description 
Right now if you want to ignore the error for a delete of a config which does not exist you have to do some error string checking. This adjusts the behavior in the config API to 404 instead of 400 when attempting to get or delete a non-existent config. 

It's possible to get a 404 from all endpoints if there's no configstore configured which makes sense. It's also possible to get a 404 if there's [no connection to the config store](https://github.com/grafana/agent/blob/main/pkg/metrics/instance/configstore/api.go#L122). This one feels like it might be a 500 instead of 404 since it seems to indicate a config/system issue vs no config store at all like the other 404. I can't tell if it's purely a safety check or possible to configure the agent in a certain way to hit it though.

N/A

#### Notes to the Reviewer
N/A

#### PR Checklist

- [x] CHANGELOG updated 
- [na] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,9 @@
 
 - [BUGFIX] Traces: remove extra line feed/spaces/tabs when reading password_file content (@nicoche)
 
+- [CHANGE] Configuration API now returns 404 instead of 400 when attempting to get or delete a config 
+  which does not exist. (@kgeckhart)
+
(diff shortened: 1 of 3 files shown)


rfratto:
@mattdurham can you review this?

PR #1011, opened 2021-10-20
Document changes due to OTel upgrade

mapno:
#### PR Description 

Document changes to receiver configs due to OTel upgrade from v0.30 to v0.36.

These changes should become less frequent or non-existent now that tracing is at v1 at OTel.

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/upgrade-guide/_index.md
+++ b/docs/upgrade-guide/_index.md
@@ -12,7 +12,7 @@ releases and how to migrate to newer versions.
 
 These changes will come in a future version.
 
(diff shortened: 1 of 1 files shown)


mapno:
I didn't find a way of supporting the old config that didn't involve a lot of code, since we're just passing the receiver's block directly to the embedded collector.

The exporters also use a TLS config block to define TLS params. And, although they're not identical, I think it brings both approaches closer. I don't see much value in supporting the old config right now.

PR #1020, opened 2021-10-22
promtail update

### LINKED ISSUE #1015: Update Promtail dependency to 0ae0d4da1
grafana/loki commit 0ae0d4da122d0cb043a2eb64b7365a48290683de contains a [fix](https://github.com/grafana/loki/pull/4461) for a behavior where if a static_target job is used that discovers no files, the target will immediately be removed and never run again. 

Promtail is stable enough that we can use non-release commits, and we should update to this commit for the next minor release.

rlankfo:
#### PR Description 

Upgrade to loki to version containing this fix: https://github.com/grafana/loki/pull/4461 

Fixes #1015

#### Notes to the Reviewer

Note this is blocked on a couple of things:
* ~~updating prometheus~~
* ~~removing dependencies on github.com/prometheus/common/log embedded in exporters that use it~~

#### PR Checklist

- [x] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated


go.mod
```diff
@@ -97,6 +93,8 @@ require (
 // Replace directives from Prometheus
 replace k8s.io/klog => github.com/simonpasquier/klog-gokit v0.1.0
 
+//replace github.com/gophercloud/gophercloud => github.com/gophercloud/gophercloud v0.17.0
```
rfratto, line 96: Remove?

go.mod
```diff
@@ -135,7 +134,7 @@ replace github.com/gomodule/redigo => github.com/gomodule/redigo v1.8.2
 exclude google.golang.org/grpc/examples v0.0.0-20200728065043-dfc0c05b2da9
 
 // Used for /-/reload
-replace github.com/weaveworks/common => github.com/rfratto/weaveworks-common v0.0.0-20210326192855-c95210d58ba7
+replace github.com/weaveworks/common => github.com/rlankfo/weaveworks-common v0.0.0-20211028201502-f0486066ff0a
```
rfratto, line 137: I wish we didn't have to bounce around which repo we're using all the time :) Can we see if the changes I made to weaveworks/common can be moved to dskit + support the changes you made too?

(This might have to be a PR all on its own, honestly)

go.mod
```diff
@@ -145,9 +144,6 @@ replace github.com/open-telemetry/opentelemetry-collector-contrib/pkg/batchpersi
 // This is a temporary solution while a new configuration design is discussed for the collector (ref: https://github.com/open-telemetry/opentelemetry-collector/issues/3482).
 replace go.opentelemetry.io/collector => github.com/grafana/opentelemetry-collector v0.4.1-0.20211008102704-a5d23001cb71
 
-// Pin prometheus dependencies
-replace github.com/prometheus/common => github.com/prometheus/common v0.23.0
```
rfratto, line 149: 🎉 Happy to see that this pin can be removed now. 
Can we add one entry each in the CHANGELOG for "primary" dependencies that we updated? Anything in depcheck.yaml should be included.

pkg/logs/logs_test.go
```diff
@@ -1,4 +1,5 @@
-//+build !race
+//go:build !race
+// +build !race
```
rfratto, line 2: IMO a small PR that does `go fmt ./...` using Go 1.17 would be nice to make this a little easier to review.

pkg/logs/logs_test.go
```diff
@@ -13,10 +14,11 @@ import (
 	"testing"
 	"time"
 
+	"github.com/grafana/loki/pkg/loghttp/push"
```
rfratto, line 17: (For the curious: loghttp is still Apache-2.0 licensed)

pkg/metrics/instance/host_filter_test.go
```diff
@@ -175,9 +175,11 @@ func TestHostFilter_PatchSD(t *testing.T) {
 	follow_redirects: true
 	kubernetes_sd_configs:
 		- role: service
+		  kubeconfig_file: ""
```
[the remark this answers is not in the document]
rfratto: I think it's safe to ignore the change as long as it unmarshals. We can put some trust into Prometheus' implementation of the SD.

pkg/metrics/instance/instance.go
```diff
@@ -843,7 +843,10 @@ func newScrapeManager(logger log.Logger, app storage.Appendable) *scrape.Manager
 	// data race of modifying that global, we lock a mutex here briefly.
 	managerMtx.Lock()
 	defer managerMtx.Unlock()
-	return scrape.NewManager(logger, app)
+	options := &scrape.Options{
+		ExtraMetrics: false,
```
rfratto, line 847: What are the extra metrics here? Does Prometheus default to true?

pkg/metrics/instance/marshal_test.go
```diff
@@ -60,6 +60,7 @@ remote_write:
     max_backoff: 100ms
   follow_redirects: true
   metadata_config:
+    max_samples_per_send: 500
```
rfratto, line 63: This change worries me a little bit. Does one of our defaults need to change to re-sync with Prometheus or something?

pkg/metrics/wal/util.go
```diff
@@ -126,6 +126,8 @@ func (c *walDataCollector) StoreSeries(series []record.RefSeries, _ int) {
 
 func (c *walDataCollector) SeriesReset(_ int) {}
 
+func (*walDataCollector) UpdateSeriesSegment([]record.RefSeries, int) {}
```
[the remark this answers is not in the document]
rfratto: walDataCollector is only used for tests I think. It's probably safe to ignore.

rfratto (commented):
Looks good!

pkg/metrics/instance/host_filter_test.go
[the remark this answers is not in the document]
rlankfo: sounds good 👍  we still need it here in the test case however though or our actual output doesn't match what we're expecting.

pkg/metrics/instance/marshal_test.go
rlankfo: This new config param was added in 2.29.0 according to the prometheus change log. Here's the PR: https://github.com/prometheus/prometheus/pull/8959

pkg/logs/logs_test.go
rlankfo: running this results in no changes, looks like we're already formatted?

pkg/metrics/instance/instance.go
rlankfo: this defaults to `false`, but looking at the code in prometheus, we should be able to just pass nil here:
https://github.com/prometheus/prometheus/blob/main/scrape/manager.go#L103-L105

rlankfo: https://prometheus.io/docs/prometheus/latest/feature_flags/#extra-scrape-metrics

pkg/logs/logs_test.go
rfratto: Ah, I guess this already covered all files that had a build directive, never mind :)

pkg/metrics/instance/marshal_test.go
rlankfo: @rfratto looking into this closer, I'm not sure this is something we need to be worried it about. It seems previously there was no max size on the amount of metrics that could be sent in one shot, resulting in some rather large message sizes. this appears to be adding an upper bound on the max and500 is the default value. we should be able to adjust this in our configurations if necessary.

go.mod
rlankfo: no additional changes were made here aside from rebasing the `custom-registerer` branch from your fork on latest upstream changes. I agree bouncing around isn't ideal. Do you have an open PR to get the changes merged upstream? I'd prefer not to block this PR on figuring this out if we can avoid it.

pkg/metrics/instance/marshal_test.go
rfratto: Hmm, I think this called out to me because I remember documenting it before the docs rewrite where we link out to Prometheus. I think what happened here is we accidentally documented a field that didn't exist (for us) yet. 

Anyways, I'm not concerned anymore, we can leave this alone now that the mystery is solved :)

go.mod
rfratto: Yeah, there's these two PRs:

* https://github.com/weaveworks/common/pull/228
* https://github.com/weaveworks/common/pull/230

but I think since all movement is going to dskit it might be more worthwhile replaying the changes there instead with a PR.

go.mod
```diff
@@ -61,41 +58,44 @@ require (
...
 	sigs.k8s.io/controller-runtime v0.9.0-beta.5
 	sigs.k8s.io/yaml v1.2.0
 )
 
 // Replace directives from Prometheus
-replace k8s.io/klog => github.com/simonpasquier/klog-gokit v0.1.0
+replace (
+	github.com/go-logr/logr => github.com/go-logr/logr v0.4.0
+	k8s.io/klog => github.com/simonpasquier/klog-gokit v0.3.0
+	k8s.io/klog/v2 => github.com/rlankfo/klog-gokit/v3 v3.0.1-0.20211103030435-2602604e10dd
```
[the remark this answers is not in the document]
rfratto: 😢 Can you add a comment explaining the replace and when we might be able to remove it?

go.mod
rlankfo: @rfratto added CHANGELOG entries for Prometheus, Cortex, and Loki.

pkg/metrics/instance/instance.go
rlankfo: @rfratto resolving this, doesn't seem like something we need/want to support in agent. Let me know if you disagree and we can unresolve.

go.mod
rlankfo: @rfratto I see that weaveworks/common#228 was merged. This PR doesn't attempt to pull in the weaveworks/common#230 but it seems like there's some movement there. For now looks like we can go back to weaveworks/common instead of this fork

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Main (unreleased)
 
+- [ENHANCEMENT] Update Cortex dependency to v1.10.0-92-g85c378182. (@rlankfo)
+
+- [ENHANCEMENT] Update Loki dependency to v2.1.0-656-g0ae0d4da1. (@rlankfo)
+
+- [ENHANCEMENT] Update Prometheus dependency to v2.31.0-1-g89bb32ee4. (@rlankfo)
```
rfratto, line 7: Can this change to just saying v2.31.0?

pkg/metrics/instance/instance.go
rfratto: Let's open an issue for this to make this configurable. It's an experimental feature in Prometheus, so we might not want to do it right away, but we should have an issue to start a discussion around it.

pkg/operator/apis/monitoring/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -1,3 +1,4 @@
+//go:build !ignore_autogenerated
```
rfratto, line 1: Can you re-run `make crds` to see if this gets removed?

rfratto (approved):
LGTM % small nits

PR #1022, opened 2021-10-25
Service graphs performance improvements

mapno:
#### PR Description 

The implementation with the cache library had a ver bad performance,
specially when calling `Items()`, since it would create a new map
and copy all items to it.

This commit changes the implementation to use a map and a list with a mutex.

The list is used to keep track of the oldest items.

Saw significant improvements in CPU usage. A new profile showed the processor to take ~2.55% of total CPU time.

#### PR Checklist

- [x] CHANGELOG updated 
- ~[ ] Documentation added~
- [x] Tests updated

### DIFF
--- a/pkg/traces/servicegraphprocessor/processor.go
+++ b/pkg/traces/servicegraphprocessor/processor.go
@@ -217,7 +217,7 @@ func (p *processor) ConsumeTraces(ctx context.Context, td pdata.Traces) error {
 
 	if err := p.consume(td); err != nil {
 		if errors.Is(err, errTooManyItems) {
-			level.Info(p.logger).Log("msg", "skipped processing of spans", "maxItems", p.maxItems, "err", errTooManyItems)
+			level.Warn(p.logger).Log("msg", "skipped processing of spans", "maxItems", p.maxItems, "err", errTooManyItems)
 		} else {
 			level.Error(p.logger).Log("msg", "failed consuming traces", "err", err)
 		}
--- a/pkg/traces/servicegraphprocessor/processor.go
+++ b/pkg/traces/servicegraphprocessor/processor.go
@@ -18,9 +18,13 @@ import (
 	"google.golang.org/grpc/codes"
 )
 
-var (
-	errTooManyItems = errors.New("too many items in memory")
-)
+type tooManySpansError struct {
+	droppedSpans int
+}
+
+func (t tooManySpansError) Error() string {
+	return fmt.Sprintf("dropped %d spans", t.droppedSpans)
+}
 
 // edge is an edge between two nodes in the graph
 type edge struct {
@@ -216,8 +220,8 @@ func (p *processor) ConsumeTraces(ctx context.Context, td pdata.Traces) error {
 	p.store.expire()
 
 	if err := p.consume(td); err != nil {
-		if errors.Is(err, errTooManyItems) {
-			level.Warn(p.logger).Log("msg", "skipped processing of spans", "maxItems", p.maxItems, "err", errTooManyItems)
+		if errors.As(err, &tooManySpansError{}) {
+			level.Warn(p.logger).Log("msg", "skipped processing of spans", "maxItems", p.maxItems, "err", err)
 		} else {
 			level.Error(p.logger).Log("msg", "failed consuming traces", "err", err)
 		}
@@ -243,6 +247,7 @@ func (p *processor) collectEdge(e *edge) {
 }
 
 func (p *processor) consume(trace pdata.Traces) error {
+	var totalDroppedSpans int
 	rSpansSlice := trace.ResourceSpans()
 	for i := 0; i < rSpansSlice.Len(); i++ {
 		rSpan := rSpansSlice.At(i)
@@ -261,10 +266,12 @@ func (p *processor) consume(trace pdata.Traces) error {
 				// Check if the store can process the remaining spans
 				if p.store.len() >= p.maxItems {
 					// todo: try to evict expired items before dropping more
-					remainingSpans := float64(ils.Spans().Len() - k)
-					p.serviceGraphDroppedSpansTotal.WithLabelValues(svc.StringVal()).Add(remainingSpans)
+					droppedSpans := ils.Spans().Len() - k
+					totalDroppedSpans += droppedSpans
 
-					return errTooManyItems
+					p.serviceGraphDroppedSpansTotal.WithLabelValues(svc.StringVal()).Add(float64(droppedSpans))
+
+					continue
 				}
 
 				span := ils.Spans().At(k)
@@ -299,6 +306,12 @@ func (p *processor) consume(trace pdata.Traces) error {
 				default:
 				}
 			}
+		}
+	}
+
+	if totalDroppedSpans > 0 {
+		return &tooManySpansError{
+			droppedSpans: totalDroppedSpans,
 		}
 	}
 	return nil
--- a/pkg/traces/servicegraphprocessor/processor.go
+++ b/pkg/traces/servicegraphprocessor/processor.go
@@ -7,6 +7,8 @@ import (
 	"sync"
 	"time"
 
+	"runtime"
+
 	util "github.com/cortexproject/cortex/pkg/util/log"
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
@@ -34,10 +36,12 @@ type edge struct {
 	// the edge will be considered as failed.
 	failed bool
 
+	// expiration is the time at which the edge expires, expressed as Unix time
 	expiration int64
 }
 
-func newEdge(w time.Duration) *edge {
+func newEdge(w time.Duration, expireFn func()) *edge {
+	go expireFn()
 	return &edge{
 		expiration: time.Now().Add(w).Unix(),
 	}
@@ -61,9 +65,11 @@ type processor struct {
 
 	// store is a local storage for edge between graphs nodes
 	store    map[string]*edge
-	storeMtx sync.Mutex
+	storeMtx sync.RWMutex
 	maxItems int
 	wait     time.Duration
+
+	firesCh chan string
 
 	closedCh chan struct{}
 
@@ -108,9 +114,11 @@ func newProcessor(nextConsumer consumer.Traces, cfg *Config) *processor {
 		logger:       logger,
 
 		store:    map[string]*edge{},
-		storeMtx: sync.Mutex{},
+		storeMtx: sync.RWMutex{},
 		maxItems: cfg.MaxItems,
 		wait:     cfg.Wait,
+
+		firesCh: make(chan string, runtime.NumCPU()),
 
 		closedCh: make(chan struct{}, 1),
 	}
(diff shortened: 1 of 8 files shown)


pkg/traces/servicegraphprocessor/processor.go
```diff
@@ -211,23 +264,80 @@ func (p *processor) ConsumeTraces(ctx context.Context, td pdata.Traces) error {
 		level.Error(p.logger).Log("msg", "failed consuming traces", "err", errs)
```
joe-elliott, line 264: instead of failing to consume traces maybe we should metric unprocessed spans but let the trace pipeline continue working.

pkg/traces/servicegraphprocessor/processor.go
```diff
@@ -158,21 +191,38 @@ func (p *processor) registerMetrics() error {
 		}
 	}
 
-	// Collect unpaired spans when evicting items from the store during
-	// periodic cleanup
-	p.store.OnEvicted(func(s string, i interface{}) {
-		e := i.(edgeRequest)
-		if !e.complete() {
-			p.serviceGraphUnpairedSpansTotal.WithLabelValues(e.clientService, e.serverService).Inc()
+	go func() {
```
joe-elliott, line 194: what throughput have you tested this at? at some point your channel is going to back up and you are going to start refusing traces.

should we have a configurable number of workers?

joe-elliott (commented):
Lot of thoughts, but overall I love the direction taken. Smart use of a linked list/map to get the desired performance characteristics.

pkg/traces/servicegraphprocessor/processor.go
```diff
@@ -197,6 +247,9 @@ func (p *processor) Capabilities() consumer.Capabilities {
 func (p *processor) ConsumeTraces(ctx context.Context, td pdata.Traces) error {
 	level.Debug(p.logger).Log("msg", "consuming traces")
 
+	// Evict expired edges
+	p.expire()
```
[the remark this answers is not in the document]
mapno: I thought of this too, but I'm not very convinced.

Running the function when consuming traces makes it scale with the throughput of the pipeline (i.e. more `ConsumeTraces` calls, more times it tries to expire). Finding the right amount of _expiry_ otherwise could be difficult.

This approach could be a problem for extremely low throughput scenarios, but I'm not sure that is a frequent occurrence.

pkg/traces/servicegraphprocessor/processor.go
```diff
@@ -22,34 +25,59 @@ import (
...
+
 var _ component.TracesProcessor = (*processor)(nil)
 
 type processor struct {
 	nextConsumer consumer.Traces
 	reg          prometheus.Registerer
 
-	// store is a local storage for request between graphs nodes
-	store    *cache.Cache
+	store    *list.List
```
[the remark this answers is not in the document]
mapno: Done in 48d234c1

pkg/traces/servicegraphprocessor/processor.go
```diff
@@ -197,6 +247,9 @@ func (p *processor) Capabilities() consumer.Capabilities {
 func (p *processor) ConsumeTraces(ctx context.Context, td pdata.Traces) error {
 	level.Debug(p.logger).Log("msg", "consuming traces")
 
+	// Evict expired edges
+	p.expire()
+
 	var errs error
 	for _, trace := range batchpersignal.SplitTraces(td) {
```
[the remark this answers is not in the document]
mapno: Actually no, nice catch. Removed in eeb762f67d744e92104c8642ee43ca89f517ae8c

pkg/traces/servicegraphprocessor/processor.go
```diff
@@ -211,23 +264,80 @@ func (p *processor) ConsumeTraces(ctx context.Context, td pdata.Traces) error {
...
-			p.serviceGraphRequestTotal.WithLabelValues(e.clientService, e.serverService).Inc()
-			if e.failed {
-				p.serviceGraphRequestFailedTotal.WithLabelValues(e.clientService, e.serverService).Inc()
-			}
-			p.serviceGraphRequestServerHistogram.WithLabelValues(e.clientService, e.serverService).Observe(e.serverLatency.Seconds())
-			p.serviceGraphRequestClientHistogram.WithLabelValues(e.clientService, e.serverService).Observe(e.clientLatency.Seconds())
-			p.store.Delete(k)
+// collectEdge records the metrics for the given edge.
+// Returns true if the edge is completed or expired and should be deleted.
+func (p *processor) collectEdge(e *edge) bool {
```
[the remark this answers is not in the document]
mapno: This is a remanent of the old implementation. You're right it's not needed. Removed in eeb762f67d744e92104c8642ee43ca89f517ae8c

pkg/traces/servicegraphprocessor/processor.go
mapno: Agree, changed in eeb762f67d744e92104c8642ee43ca89f517ae8c

pkg/traces/servicegraphprocessor/processor.go
mapno: I've only tested it at ~6500 spans/sec, not very high. I've the logic to spawn multiple workers.

pkg/traces/servicegraphprocessor/processor.go
```diff
@@ -158,21 +191,38 @@ func (p *processor) registerMetrics() error {
...
+				p.storeMtx.Lock()
+
+				ele := p.storeMap[k]
+				if ele == nil { // it may already have been processed
+					p.storeMtx.Unlock()
+					continue
+				}
+
+				edge := ele.Value.(*edge)
+				if shouldDelete := p.collectEdge(edge); shouldDelete {
```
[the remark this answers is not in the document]
mapno: You're right, this is no longer necessary. Removed in eeb762f67d744e92104c8642ee43ca89f517ae8c

pkg/traces/servicegraphprocessor/processor.go
```diff
@@ -87,14 +115,19 @@ func newProcessor(nextConsumer consumer.Traces, cfg *Config) *processor {
...
+		logger:       logger,
+
+		store:    list.New(),
+		storeMtx: &sync.RWMutex{},
+		storeMap: make(map[string]*list.Element),
 		maxItems: cfg.MaxItems,
-		logger:   logger,
+		wait:     cfg.Wait,
+
+		collectCh: make(chan string, runtime.NumCPU()),
```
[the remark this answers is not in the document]
mapno: Changed to run a configurable num of workers in 48d234c1

pkg/traces/servicegraphprocessor/processor.go
```diff
@@ -22,34 +19,56 @@ import (
...
 
 type processor struct {
 	nextConsumer consumer.Traces
 	reg          prometheus.Registerer
 
-	// store is a local storage for request between graphs nodes
-	store    *cache.Cache
+	store *store
+
+	wait     time.Duration
```
joe-elliott, line 66: nit: instead of having these 3 fields you could have a config object.

pkg/traces/servicegraphprocessor/processor.go
```diff
@@ -260,27 +273,28 @@ func (p *processor) consume(trace pdata.Traces) error {
 				case pdata.SpanKindClient:
 					k := key(span.TraceID().HexString(), span.SpanID().HexString())
 
-					var r edgeRequest
-					if v, ok := p.store.Get(k); ok {
-						r = v.(edgeRequest)
+					edge := p.store.upsertEdge(k, func(e *edge) {
+						e.clientService = svc.StringVal()
+						e.clientLatency = spanDuration(span)
+						e.failed = p.spanFailed(span)
```
joe-elliott, line 279: failed can be overwritten if one side shows success and one side shows failure. edge failure or success would then be determine by whichever span it randomly processed last.

if failed is true we should never change it back to false

joe-elliott (commented):
I find the flow of edges confusing in this iteration.

- processor.consume() calls store.upsertEdge()
  - callback in order to set specific fields
- push to a channel created by store if expired/completed
  - received by a store worker
  - call back into processor using evictCallback()

Can all of the channel-y flow stuff be owned by processor and the store just be a simple set of methods used to coordinate the list/mtx/map? Am I just overcomplicating this in my head?

pkg/traces/servicegraphprocessor/processor.go
```diff
@@ -247,11 +280,15 @@ func (p *processor) consume(trace pdata.Traces) error {
 
 			for k := 0; k < ils.Spans().Len(); k++ {
 
-				if p.store.ItemCount() >= p.maxItems {
-					remainingSpans := float64(ils.Spans().Len() - k)
-					p.serviceGraphDroppedSpansTotal.WithLabelValues(svc.StringVal()).Add(remainingSpans)
+				// Check if the store can process the remaining spans
+				if p.store.len() >= p.maxItems {
+					// todo: try to evict expired items before dropping more
+					droppedSpans := ils.Spans().Len() - k
```
joe-elliott, line 286: i think this will now double count spans b/c droppedSpans includes all future spans in `ils.Spans()` and you're looping all the way to the end.

also you don't really know for sure that you can't store the incoming edge b/c you haven't checked yet. at this point adding the span may simply match it up with another edge that already exists in the store or it might not even really be an edge span at all.

maybe `upsertEdge` should return errTooManyItems and then you can just: `totalDroppedSpans++`

joe-elliott (commented):
i find the flow of edges way more understandable in this one. we are close!

i still don't think we're counting dropped spans quite right. let's get that patched up and merge.

pkg/traces/servicegraphprocessor/processor.go
mapno: Good idea. The dropped spans logic has been broken all this time 🤦‍♂️ .

It makes more sense to move `maxItems` to the store, than having the processor handle it.

pkg/traces/servicegraphprocessor/processor.go
```diff
@@ -21,36 +18,61 @@ import (
...
+	return &edge{
+		key: key,
+
+		expiration: time.Now().Add(ttl).Unix(),
+	}
+}
+
+// completed returns true if the corresponding client and server
+// pair spans have been processed for the given edge
+func (e *edge) isCompleted() bool {
```
joe-elliott, line 54: `isCompleted()` has a weird edge case where if a span has a `service.name` attribute but it's an empty string then this function will never return completed even if both sides of an edge has been discovered.

When we get the service name perhaps we could do something like:

```
svc, ok := rSpan.Resource().Attributes().Get(semconv.AttributeServiceName)
if !ok {
	continue
}
// substitute in a service name ....
if len(svc) == 0 {
  svc = "<empty>"
}
```

pkg/traces/servicegraphprocessor/processor.go
```diff
@@ -247,46 +280,66 @@ func (p *processor) consume(trace pdata.Traces) error {
...
-					var r edgeRequest
-					if v, ok := p.store.Get(k); ok {
-						r = v.(edgeRequest)
+					edge, err := p.store.upsertEdge(k, func(e *edge) {
+						e.clientService = svc.StringVal()
+						e.clientLatency = spanDuration(span)
+						e.failed = e.failed || p.spanFailed(span) // keep request as failed if any span is failed
+					})
+
+					if err != nil {
```
joe-elliott, line 295: unnest for clarity:
```
if errors.Is(err, errTooManyItems) {
  ...
  continue
}
if err != nil {
  return err
}
```

joe-elliott (approved):
two small changes requested, but lgtm!

pkg/traces/servicegraphprocessor/processor.go
mapno: changed

pkg/traces/servicegraphprocessor/processor.go
mapno: You're right, nice catch. Although I'm not convinced that we should process spans with empty service name. I'd rather count them as unpaired.

mapno: Pushed a change to skip processing of spans with `""` service name. It can be changed later.

joe-elliott: I'm good on this, but make sure that the documentation lists `service.name` (along with `span.kind`) as a required attributes for this to work.

PR #1033, opened 2021-10-26
Change integrations to have a unique instance key

### LINKED ISSUE #1019: Change instance label used for integrations
By default, integrations use known values for the `job` and `instance` labels: `job` is set to `integrations/<integration name>` and `instance` is set to the hostname:port that the agent is running and listening on. 

#350 would introduce label collision, as there would be two instances of the same integration using the same labels. The obvious workaround is to add another label, but as most dashboards are only aware of job and instance, this workaround would not be sufficient for users, as disambiguating between targets would not be obvious.  

I propose that integrations be given some kind of primary key that identifies the value of the instance label. Integrations that connect to a system will have the primary key set to the hostname:port of that SUO (such as the MySQL server). Other integrations will continue to use the agent's hostname:port. The value used for the instance label will be customizable, and the primary key will only be used for the default value.

There is one more edge case this change causes. If two agents are running on two different machines and both configured to scrape from a localhost SUO, then you return to having a label collision on the combination of job/instance. As a workaround, I also propose undeprecating and requiring the `agent_hostname` label to exist. 

This means that for each integration, there will be at least three labels:

- `job`: `integrations/<integration name>`
- `instance`: `<suo address:port>`
- `agent_hostname`: `<agent hostname:port>`

This re-introduces the disambiguation problem mentioned earlier. However, I believe this form is more acceptable. Users can disambiguate through the following:

1. Identify that they have multiple targets with the same job & instance label
2. Examine the agent_hostname label of these duplicated target labels 
3. Modify the instance label for each integration to fix the collision.

rfratto:
#### PR Description 
Changes integrations to have a unique instance key. The instance key determines the `instance` label assigned to metrics.  

The URL used to scrape these integrations is left unchanged, but should probably be changed in a future PR. Changing the URL would be a breaking change (since some users manually scrape the endpoints from integrations), and we would need to be considerate about how to ease the transition.

Closes #1019

#### Notes to the Reviewer
This will need careful review to decide whether the defaults I gave to each integration make sense. This adds costs to future integrations since the author has to consider what an appropriate instance label default would be. 

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated


mapno (approved):

rlankfo (approved):
lgtm

PR #1047, opened 2021-10-29
Prototype adding a readiness check for metrics

### LINKED ISSUE #1009: `/-/ready` returns ready when WAL is still loading
We'd like to add a readiness check to our agent deployment, but we noticed that the `/-/ready` endpoint tells us the agent is ready even when the WAL is still reloading.

Would it be possible to _beef this up_ to check the readiness of the components more thoroughly?

rfratto:
#### PR Description 
Extends the `/-/ready` check to ensure that all current instances have been initialized. Waits for the first apply on startup to finish to ensure that at least the instances defined in the config file are available for checking.

Fixes #1009

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/cmd/agent/entrypoint.go
+++ b/cmd/agent/entrypoint.go
@@ -158,6 +158,13 @@ func (ep *Entrypoint) wire(mux *mux.Router, grpc *grpc.Server) {
 	ep.manager.WireAPI(mux)
 
 	mux.HandleFunc("/-/healthy", func(w http.ResponseWriter, r *http.Request) {
+		if !ep.promMetrics.Ready() {
+			w.WriteHeader(http.StatusServiceUnavailable)
+			fmt.Fprintf(w, "Metrics are still loading.\n")
+
+			return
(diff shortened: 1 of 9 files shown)


cmd/agent/entrypoint.go
```diff
@@ -158,6 +158,13 @@ func (ep *Entrypoint) wire(mux *mux.Router, grpc *grpc.Server) {
 	ep.manager.WireAPI(mux)
 
 	mux.HandleFunc("/-/healthy", func(w http.ResponseWriter, r *http.Request) {
+		if !ep.promMetrics.Ready() {
```
rfratto, line 161: Shouldn't this be on /-/ready and not /-/healthy?

rfratto:
@tpaschalis can you take a look at this when you get a chance?

rfratto:
Closing in favor of #1369

PR #1051, opened 2021-11-01
Show remote write latency on dashboard

bboreham:
#### PR Description 

If samples are getting backlogged, it may be due to latency between the agent and the backend it is sending to.

Replaced the differential panel which doesn't add much to the timestamp panel to the left, and essentially duplicates the differential samples rate panel below.

Before:
image

After:
image

#### PR Checklist

- [x] CHANGELOG updated 
- NA Documentation added
- NA Tests updated


rfratto (approved):
Nice, thanks!

PR #1053, opened 2021-11-01
Use correct user/group env variables

simonc6372:
The env variables in add_to_logging_group() did not match the names set at the start of the post install script.

#### PR Description 
Correct names of user & group environment variables so uniform name is used throughout. 

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated


rfratto (approved):
Thank you!

rfratto:
Before I merge, can you add this as a BUGFIX to the changelog?

simonc6372:
Done.

PR #1055, opened 2021-11-02
RFC: Design in the open

rfratto:
Proposes RFC0001 to change the design and proposal process for the Grafana Agent to be open by default.


docs/rfcs/0001-designing-in-the-open.md
```diff
@@ -0,0 +1,138 @@
...
+make the review process easier for readers.
+
+## Considered alternatives
+
+A few existing public proposal processes have been examined for inspiration:
+
+* [IETF's RFCs](https://www.ietf.org/standards/rfcs/)
+* [Rust's RFCs](https://github.com/rust-lang/rfcs)
+* [Joyent's Requests for Discussions](https://github.com/joyent/rfd)
+* [OpenTelemetry's OTEPs](https://github.com/open-telemetry/oteps)
```
rlankfo, line 135: additional inspiration:
```suggestion
* [OpenTelemetry's OTEPs](https://github.com/open-telemetry/oteps)
* [Kubernetes Enhancement Proposals (KEPs)](https://github.com/kubernetes/enhancements)
```

docs/rfcs/0001-designing-in-the-open.md
```diff
@@ -0,0 +1,138 @@
...
+# Designing in the Open
+
+* Date: 2021-11-02
+* Author: Robert Fratto (@rfratto)
+* PR: [grafana/agent#1055](https://github.com/grafana/agent/pull/1055)
+
+## Summary
+
+Traditionally, design proposals to Grafana Agent have been done in private by
+default, and then made only made public after internal consensus was reached.
```
rlankfo, line 15:
```suggestion
default, and then only made public after internal consensus was reached.
```

docs/rfcs/0001-designing-in-the-open.md
```diff
@@ -0,0 +1,138 @@
...
+2. A bulletpoint list of metadata that MUST include original authorship date
+   and the author's name. Author name SHOULD include their GitHub username as a
+   parenthethical.
+3. A summary section describing the context for what is being proposed.
+4. A list of goals and non-goals for the proposal.
+5. One or more sections dedicated to the proposal.
+
+RFCs MUST be placed into the `docs/rfcs` folder of the Grafana Agent repository.
+
+RFCs SHOULD have a `Terminology` section following the goals and non goals that
```
rlankfo, line 80: In the spirit of avoiding rigid process, I could see this SHOULD here being a MAY.

rfratto: Maybe! I think RFC2119 really helps clarify the language of documents. But it's also one of those things where I haven't found many believers, and might need a push to start making it a habit.

docs/rfcs/0001-designing-in-the-open.md
```diff
@@ -0,0 +1,139 @@
...
+Sections dedicated to the proposal SHOULD include alternatives considered.
+
+#### RFC IDs
+
+The filename of the RFC MUST be prefixed with the ID of the RFC. If an RFC does
+not yet have an ID, the ID MAY be omitted until one is assigned to it.
+
+#### RFC Mutability
+
+Once an RFC is accepted, all non-metadata information MUST remain immutable.
```
rfratto, line 92: Immutability makes sense for IETF, but since all of these will live in Git with its own version history, maybe we don't need to be as restrictive here?

Or, maybe, we're ok with errata fixes to an RFC, but the spirit/proposal of the RFC should not change?

rfratto: I think there's probably still some value in having immutability. It allows you to use an RFC ID alone for what is being implemented. Otherwise, you'd have to say "RFC-ID as of YYYY-MM" to make it clear which iteration of a document you're referring to.

docs/rfcs/0001-designing-in-the-open.md
```diff
@@ -0,0 +1,139 @@
...
+* Smaller changes are the inverse of the above.
+
+### RFCs
+
+RFCs MUST be valid markdown composed of at least the following:
+
+1. A title.
+2. A bulletpoint list of metadata that MUST include original authorship date
+   and the author's name. Author name SHOULD include their GitHub username as a
+   parenthethical.
```
rfratto, line 73: IMO there should be a section here iterating over the base set of metadata we'd expect to be included.

docs/rfcs/0001-designing-in-the-open.md
rlankfo: Imo we probably would want to avoid immutability where possible but if there are valid reasons for changing something, I don't think we should restrict it. It's likely not as big as an issue as it would be for the IETF as you mention, where they may be dealing with protocol specifications and multiple implementers.

rfratto:
Should "be as least restrictive as possible" be a goal?

docs/rfcs/0001-designing-in-the-open.md
```diff
@@ -0,0 +1,139 @@
...
+Issue proposals are the second form of proposal, useful for smaller proposals
+that don't warrant an entire RFC.
+
+Issue proposals are freeform, but SHOULD include:
+
+* Background on what is being proposed
+* The proposal itself
+* If relevant, code or configuration examples which compliment the proposal
+
+## Concerns
```
rfratto, line 118: I think by the time an RFC gets accepted, open issues / concerns should be removed from the RFC and resolved via discussion in the corresponding PR.

docs/rfcs/0001-designing-in-the-open.md
```diff
@@ -0,0 +1,139 @@
...
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL
+NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and
+"OPTIONAL" in this document are to be interpreted as described in
+[RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+## Proposal
+
+Public propsals come in two forms:
+
+* RFCs (i.e., this document) for larger changes
```
rfratto, line 52: Should RFC be renamed to something else to uniquely identify what project an RFC refers to? RFC seems to be a bloated term now and I think most people read RFC and think IETF.

(By most people I mean "me," I think IETF when I read RFC).

rlankfo: I also tend to associate RFC with the IETF as well. Joyent changed the terminology of their process for exactly this reason, which makes sense to me: 'We use the term "Requests for Discussion" in lieu of "Requests for Comments" to avoid conflation with the IETF construct -- and the more formal writing that it has come to represent.' 

Did you have another name in mind?

rfratto: Nope, but it's also worth noting that if this process leaks to other Grafana-hosted projects, we might have the same problem multiple times. 

There's always something like RFC-Agent-0001 🙈

docs/rfcs/0001-designing-in-the-open.md
```diff
@@ -0,0 +1,139 @@
...
+
+RFCs MUST be valid markdown composed of at least the following:
+
+1. A title.
+2. A bulletpoint list of metadata that MUST include original authorship date
+   and the author's name. Author name SHOULD include their GitHub username as a
+   parenthethical.
+3. A summary section describing the context for what is being proposed.
+4. A list of goals and non-goals for the proposal.
+5. One or more sections dedicated to the proposal.
```
thejosephstevens, line 76: A couple sections I've found useful in RFCs as technical proposals:
Drawbacks (what are the downsides to this proposal)
Rationale and Alternatives (why does the author think this is the best path forward, what other options were considered)
Prior Art (most designs aren't fully original, these can be helpful to orient discussion)
Unresolved Questions (where does the author see uncertainty/what questions should get answered before considering this a complete discussion)

I don't know how formal/strict these RFCs are likely to end up, but it could be useful to add these as suggested sections for authors who are interested in bringing a well-structured proposal forward.

docs/rfcs/0001-designing-in-the-open.md
RichiH: If you want to copy IETF, RFCs MUST NOT be mutable. They can have addendums, corrections, references, be superseeded by others. But they can not change.

RichiH: Imagine a variant of semver where you can sometimes maybe change what a version number points to.

docs/rfcs/0001-designing-in-the-open.md
```diff
@@ -0,0 +1,139 @@
...
+* Community concerns are less likely to have an impact as core maintainers have
+  already agreed on a solution.
+* Initial design is gated to Grafana Labs employees.
+* Historical design docs with context and discussions become hard to find both
+  internally and even harder publicly.
+
+In the spirit of building a broader community, this document proposes inverting
+the design process to be public by default.
+
+## Goals
```
rfratto, line 27: Should "be as least prescriptive as possible" be a goal?

The initial approach here seems probably a little too prescriptive, and the more things that are marked as required, the less people will want to do them. 

I think this is probably worth a second iteration with even less structural requirements.

docs/rfcs/0001-designing-in-the-open.md
```diff
@@ -0,0 +1,139 @@
...
+
+* Outline a process to support public-first design documents
+* Lower the barrier to entry to becoming a maintainer that is not employed at
+  Grafana Labs
+* Encourage a community to contribute input to design documents
+* Encourage a community to contribute their own design documents
+
+## Non-Goals
+
+* Make everything public. If something can't be public for confidentiality
```
wiardvanrij, line 37: This sentence has a lot of ambiguity. I would suggest removing "everything" and "something". Then define what should be confidential and what the scope is.  

I understand the goal here, but now (don't get me wrong, I know it's not the intention) it's a weird loophole.

docs/rfcs/0001-designing-in-the-open.md
```diff
@@ -0,0 +1,139 @@
...
+
+In the spirit of building a broader community, this document proposes inverting
+the design process to be public by default.
+
+## Goals
+
+* Outline a process to support public-first design documents
+* Lower the barrier to entry to becoming a maintainer that is not employed at
+  Grafana Labs
+* Encourage a community to contribute input to design documents
```
wiardvanrij, line 32: This document already has some form of guidelines on how to deal with RFCs. What I'm missing is how an RFC can get accepted. Is this the document for it to clarify that or should that become a follow-up?

docs/rfcs/0001-designing-in-the-open.md
rfratto: I'll rephrase that public by default should be best effort. 
I don't feel comfortable promising that everything will always be made public from the start (what if it's a security issue, etc.), and I'm not sure that any maintainer of this repository has authority to promise that either. 

The main problem I'm trying to address is how as a maintainer, I've personally been doing a lot of communication about the project internally first out of habit when I really didn't have to be doing that.

docs/rfcs/0001-designing-in-the-open.md
rfratto: Sure, we can clarify that proposals are up to the discretion/consensus of the governance team. While that's all Grafana Labs employees at the moment, I'm keeping a close eye out for external collaborators that want to be involved with the project at that level.

docs/rfcs/0001-designing-in-the-open.md
```diff
@@ -0,0 +1,139 @@
...
+strongly define categories, but these guidelines can be used for a gut check:
+
+* Larger changes require proposals that require thought and explore multiple
+  options, OR
+* Larger changes have a lasting effect on the project as a whole, OR
+* Larger changes touch multiple systems throughout the project or dramatic
+  changes to a single system.
+* Smaller changes are the inverse of the above.
+
+### RFCs
```
rfratto, line 66: I don't think designing in the open necessarily means RFCs. It could just mean issues that link to a google doc, which are already being used internally. 

RFCs help make things more discoverable and keep more context, but it may be an orthogonal problem to attempting to be more transparent when possible.

docs/rfcs/0001-designing-in-the-open.md
wiardvanrij: I understand the goals and I really love what you want to achieve. Though it's just some nuances I guess. If we take the example of security; as a user and/or in the future contributor, it would please me to know how it's handled. 
Having a complete open project on multiple levels, yet with some parts confidential can become a bit weird. What if I get hit by a security issue and Grafana Labs knew this issue internally for a few weeks? 
I'm not saying everything must be public, though describing a few things on the how/why/what gives openness in a different form. If that makes any sense (trying hard here to describe my thoughts :) )

rfratto: Absolutely, and I do think it's worth thinking more about this. Describing things about openness more broadly as a mission statement sounds like it could work. 

> What if I get hit by a security issue and Grafana Labs knew this issue internally for a few weeks?

Yeah, gut feeling is that legal issues would prevent premature disclosure here. I think we do well w/r/t when we release public security issue postmortems, but I also believe that disclosing issues before there's an available fix and a CVE would also just put people more at risk.

rfratto:
(I'll rebase on main to fix conflicts before merging, don't want to have to rebase+force push a bunch of times during review)

rfratto:
CC'ing governance team one last time for rough consensus: @mattdurham @joe-elliott @rfratto @RichiH @hjet @mapno

docs/README.md
```diff
@@ -1,21 +1,10 @@
...
-  (only focusing on `remote_write`-related functionality).
-- Grafana Agent allows for deploying multiple instances of the Agent in a
-  cluster and only scraping metrics from targets that are running on the same host.
-  This allows for distributing memory requirements across the cluster
-  rather than pressurizing a single node.
+* `user/`: Documentation for users. This directory is hosted on
+  [grafana.com/docs/agent](https://grafana.com/docs/agent/latest/) and we
+  recommend interacting with it there instead of viewing the markdown on
+  Github.
+* `rfcs/`: RFCs for proposals relating to Grafana Agent.
```
RichiH, line 10: My gut is to just have `rfcs` in the normal user docs. https://prometheus.io/docs/introduction/design-doc/ has not led to major confusion, either.

docs/rfcs/0001-designing-in-the-open.md
```diff
@@ -0,0 +1,148 @@
...
+from the varied backgrounds of its contributors: different people with
+different use cases combining together to make a widely useful piece of
+software.
+
+However, even with an intent to foster a community of developers, it's natural
+to accidentally build a habit from the closed-door design process. Even when
+once-private proposals are made public, potential external contributors can
+find themselves simply as spectators:
+
+* Initial design is gated to core maintainers, leaving less room for new people
```
RichiH, line 25:
```suggestion
* Initial design is gated to core maintainers, in particular if they all work for the same company. This leaves less room for new people
```

docs/rfcs/0001-designing-in-the-open.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Non-Goals
+
+* Enforce that every change originates from a fully public proposal or
+  discussion. While all maintainers and contributors will be encouraged to
+  design openly, there may be legal, security, privacy, or business reasons
+  that prevent some or all context from being made public.
+
+* Be overly prescriptive: too many rules can hinder adoption of a process. This
+  outlines more options than it is requirements.
```
RichiH, line 48:
```suggestion
  document outlines intention, not hard policy.
```

docs/rfcs/0001-designing-in-the-open.md
```diff
@@ -0,0 +1,148 @@
...
+`0000-template.md` contains a template to use for writing proposals that
+conforms to these rules.
+
+The remainder of the proposal may be formatted however the author wishes. Some
+example sections in the RFC may be:
+
+* Summary: What is the background that lead to this proposal?
+* Goals: What are the main goals of the proposal?
+* Non-Goals: What _aren't_ the main goals of the proposal?
+* Proposal: What is the proposal?
```
RichiH, line 90:
```suggestion
* Proposal 1: What is the proposal?
```

docs/rfcs/0001-designing-in-the-open.md
```diff
@@ -0,0 +1,148 @@
...
+* Pros: What are the upsides to this proposal?
+* Cons: What are the downsides to this proposal?
```
RichiH, line 92: Indent?

docs/rfcs/0001-designing-in-the-open.md
```diff
@@ -0,0 +1,148 @@
...
+* Considered Alternatives: Why is this proposal the best path forward? What were the alternatives?
```
RichiH, line 93:
```suggestion
* [Proposal 2]
* Considered Alternatives: Why is this proposal the best path forward? What were the alternatives?
```

docs/rfcs/0001-designing-in-the-open.md
```diff
@@ -0,0 +1,148 @@
...
+title with `RFC:` to easily identify it amongst other PRs.
+
+#### RFC Mutability
+
+Once an RFC is accepted, it must be treated as immutable. There are two exceptions:
+
+* The author's name and username may be changed, as long as authorship is still
+  credited to the same individuals.
+
+* Errata and typos may be fixed as long as it does not change the
```
RichiH, line 109: Define how they are referenced, how deprecations, and referals to new RFCs work?

You don't _need_ to, but if they're "immutable", you need to be more specific about what's mutable, and how.

docs/rfcs/0001-designing-in-the-open.md
```diff
@@ -0,0 +1,148 @@
...
+  interpretation of the proposal.
+
+### Google Docs Proposals
+
+It is not recommended to use Google Docs for writing proposals:
```
RichiH, line 114: In my experience, the first round or three are a _lot_ quicker in GDocs.

RichiH (changes requested):
See inline comments. Any luck reaching @rfratto, @rfratto?

docs/README.md
rfratto: I think it'd be nice to host them, but I'm concerned that they might be subject to extra scrutiny if they're hosted on grafana.com. Keeping them separate seems the most expedient for now.

docs/rfcs/0001-designing-in-the-open.md
rfratto: Similarly not sure why there are proposal 1/proposal 2 here 🤔

docs/rfcs/0001-designing-in-the-open.md
rfratto: 👍 I've taken out the immutability requirement for now. Too much to think about for what we're trying to do at this stage.

docs/rfcs/0001-designing-in-the-open.md
rfratto: Yeah, it is faster. But it still risks contextual loss, and I'm not sure if that can be fixed. Maybe it's a mistake to try so hard to keep all history intact?

rfratto:
> Any luck reaching @rfratto, @rfratto?

@rfratto hasn't replied yet. I'm starting to think he might not be very engaged in this project.

rfratto:
@RichiH have I addressed your concerns?

docs/rfcs/0001-designing-in-the-open.md
```diff
@@ -0,0 +1,142 @@
...
+it is not recommended for the permanent home of a proposal:
+
+* Change and comment history may not be available to all viewers.
+
+* The file owner may delete the proposal, leading to a gap in historical
+  context.
+
+Google Docs proposals will be permitted if linked to from an issue proposal.
+
+It is recommended that proposals from Google Docs are eventually converted by
```
mattdurham, line 116: I think this should be required. The repo should be the source of truth.

mattdurham (approved):
LGTM

docs/rfcs/0001-designing-in-the-open.md
rfratto: Good idea, let me know what you think of the rephrased wording.

Issue #1059, opened 2021-11-03
Proposal: Multiple instances of integrations


rfratto:
(This is a more design-y proposal for #350)

It is not currently possible to run multiple instances of integrations. This is a proposal to enable this to happen. 

## Configuration 

Currently, each integration may be defined in the YAML exactly once, via a `<integration name>` key inside of the `integrations` block. 

I propose adding a new `<integrations name>_configs` key for each integration, which holds an array of integration configs. Elements in this array will default to `enabled: true`, a change from the previous key. 

The old style of defining integrations will be _marked as deprecated and eventually removed_. Config files will fail to load if both the old style and new style are used simultaneously for a given integration. 

Old YAML:

```yaml
integrations:
  memcached_exporter:
    enabled: true
    memcached_address: localhost:11211 
```

New YAML:

```yaml
integrations:
  memcached_exporter_configs:
    - memcached_address: localhost:11211     
    - memcached_address: localhost:11212     
    - memcached_address: localhost:11213 
```

## HTTP Path 

The HTTP path used for integration metrics today must change. Today, each integration a unique HTTP endpoint at `/integrations/<integration name>/metrics`. Individual instances of the same integration would need to be given different endpoints so they can be scraped as separate targets. 

I propose using the instance key from #1033 as part of the path name for the HTTP endpoint. This key would need to be unique (it's currently not validated, IIRC) for each integration instance. The new path for an integration would be collectable via `/integration/<integration name>/<instance key>/metrics`.  

For example, this config:

```yaml
integrations:
  memcached_exporter_configs:
    - memcached_address: localhost:11211     
    - memcached_address: localhost:11212     
    - memcached_address: localhost:11213 
```

Will generate three endpoints (using URL-encoding for the instance key where required):

* `/integration/memcached_exporter/localhost%3A11211/metrics` 
* `/integration/memcached_exporter/localhost%3A11212/metrics` 
* `/integration/memcached_exporter/localhost%3A11213/metrics` 

This new HTTP path is only used for the `_configs`-style integrations, so old paths will continue to work. Once a user migrates to the _configs-style format, they must update any scrape jobs they have that use that old path. 

## Integration Instance Listing

Navigating to `/integrations/<integration name>` should show a listing of all metrics endpoints as HTML to make it easier for users to manually discover and configure scrape jobs for these integrations.

rfratto:
> Will generate three endpoints (using URL-encoding for the instance key where required):

[...]

Issue #868, opened 2021-09-01
conf.d-ish suport


rfratto:
Not all users of the Grafana Agent use config management tools to deploy the Grafana Agent. Such users may end up with a massive config file that is hard to understand at a glance. 

Allowing users to have a folder of config files would enable people to logically break up their config file and make it easier to digest. This could then be paired with better documentation for features like integrations, telling users to create a file within a directory instead of having them place it in the monolithic file.

I have created a design document proposing one way to implement this system: https://docs.google.com/document/d/1J2cFnWhf-FLgpKnbjq3eYO0_odHrOpkJ6DfDoQYvDkE/edit?usp=sharing

PR #1062, opened 2021-11-04
[dev.multiple-integrations] Enable present integrations by default, deprecate enabled field

rfratto:
#### PR Description 
Integrations will now default to being enabled when they are present in the `integrations` block. This is a breaking change from before, where `enabled: true` must be explicitly provided to enable an integration that was defined.

Note that here "defined" means that its key is inside `integrations`:

```yaml
integrations:
  # node_exporter is defined, agent is not. 
  node_exporter: {} 
```

Integrations that are not defined will not be run regardless of the value of `enabled`.
Integrations that are defined may still be explicitly disabled with `enabled: false`. 

The difference between "enabled" and "defined" is _really_ confusing, and I don't think 
provides any value. This PR also marks `enabled` as deprecated in favor of recommending 
that users un-define (through removal or commenting out) integrations instead of 
setting `enabled`.  

Related to #1059.

#### Notes to the Reviewer
I know #1059 proposed that `enabled` would have a different default based on the context, but it was difficult to get cleanly in the code. I think it makes sense to just make this change now instead of pushing it off. 

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,19 @@
 # Main (unreleased)
 
+BREAKING CHANGE: Integrations have changed in this release.
+Please review the [migration guide](./docs/migration-guide.md) for details.
+
 - [BUGFIX] Packaging: Use correct user/group env variables in RPM %post script (@simonc6372)
 
 - [BUGFIX] Validate logs config when using logs_instance with automatic logging processor (@mapno)
 
 - [CHANGE] Self-scraped integrations will now use an SUO-specific value for the `instance` label. (@rfratto)
+
+- [CHANGE] Integrations present in the `integrations:` map will now default to
+  being enabled. (@rfratto)
+
+- [DEPRECATION] Integrations: The `enabled` field is now deprecated and will be
+  removed in a future release. (@rfratto)
 
 # v0.20.0 (2021-10-28)
 
--- a/docs/configuration/integrations/_index.md
+++ b/docs/configuration/integrations/_index.md
@@ -14,7 +14,11 @@ manually write `scrape_configs`:
 agent:
   # Enables the Agent integration, allowing the Agent to automatically
   # collect and send metrics about itself.
-  [enabled: <boolean> | default = false]
+  #
+  # Enabled is DEPRECATED and will be removed in a future release. To disable
+  # an integration, comment it out or remove it from your config instead of
+  # setting `enabled: false`.
+  [enabled: <boolean> | default = true]
 
   # Sets an explicit value for the instance label when the integration is
   # self-scraped. Overrides inferred values.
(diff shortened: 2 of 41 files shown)


gaantunes (approved):
Makes total sense to me. If it`s defined it should be running, otherwise explicitly disable or comment it. I am only worried a lot of integrations might start unwillingly before the users realize it. How do we plan to communicate this change?

rfratto:
> I am only worried a lot of integrations might start unwillingly before the users realize it. How do we plan to communicate this change?

That's a good point. I documented the deprecation in the docs and in the upgrade guide, but I think there's probably a few more things to do here (in followup PRs, maybe):

1. Log a deprecation warning if `enabled: false` is set.
2. Remove `enabled: true` from all documentation and generated configs. I've already done the work to that within the agent repo, but Grafana Cloud and other vendors will need to do that extra step on their side once the agent is released with this change.

james-callahan:
> This PR also marks `enabled` as deprecated in favor of recommending
> that users un-define (through removal or commenting out) integrations instead of
> setting `enabled`.

I'm not sure this should be deprecated.
I think it makes sense to have a way (via e.g. `-config.expand-env`) to turn a configured integration on/off.

```yaml
integrations:
  # node_exporter is defined, agent is not. 
  node_exporter:
    enabled: ${ENABLE_NODE_EXPORTER:-false}
    disable_collectors:
      - meminfo_numa
```

rfratto:
> I'm not sure this should be deprecated.
I think it makes sense to have a way (via e.g. -config.expand-env) to turn a configured integration on/off.

My general concern is that the combination of present and the two `enabled` and `scrape_integration` configs form a confusing matrix about the behavior of an individual integration that is hard to document and explain. 

I'm definitely open to doing something different though, can you explain your use case?

james-callahan:
> I'm definitely open to doing something different though, can you explain your use case?

One config file that I use across multiple various machines, enabling/disabling integrations via environment variable.

gaantunes:
> > I am only worried a lot of integrations might start unwillingly before the users realize it. How do we plan to communicate this change?
> 
> That's a good point. I documented the deprecation in the docs and in the upgrade guide, but I think there's probably a few more things to do here (in followup PRs, maybe):
> 
> 1. Log a deprecation warning if `enabled: false` is set.
> 2. Remove `enabled: true` from all documentation and generated configs. I've already done the work to that within the agent repo, but Grafana Cloud and other vendors will need to do that extra step on their side once the agent is released with this change.

Sounds good, I can take a look if any integration has this flag configured on the agent file.

rfratto:
> (Sorry if this was hard to follow, it's the start of my morning and I haven't had a chance to make coffee yet)

Ah, I forgot to include the question part of my comment @james-callahan :) 

What do you think about the options? Would you be interested in a general solution for managing the config of multiple agents that might rely on extra tooling? Or do you want a hybrid approach where some things are configurable within the agent, but others might have to be deferred to something else?

rgeyer:
I agree that the `enable` property is confusing, and should be deprecated. It's clear to me that when an integration is defined, it should be enabled by default.

WRT the ability to have a dynamic configuration that can be used for several different agent deployments. I believe that the [conf.d](https://github.com/grafana/agent/issues/868) support that's forthcoming adequately answers that.

Namely, each integration definition can be in it's own conf.d file, and you enable/disable it by including or excluding that file from the conf.d directory.

Then the deprecation of `enable`, and only defining the integration makes sense, because it's the most terse, and declarative way to define what you want.

rfratto:
I know there's still a pending discussion, but since this is a PR against a dev branch and the discussion seems to have died down, I'm going to merge this for now.

PR #1076, opened 2021-11-09
Document new config options

mapno:
#### PR Description

- Document new config options in the processor `workers` and `success_codes`.
- Fix `success_codes` logic and actually use them 😅 

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/configuration/traces-config.md
+++ b/docs/configuration/traces-config.md
@@ -295,6 +295,24 @@ service_graphs:
   # a higher max number of items increases the max throughput of processed spans
   # with a higher memory consumption.
   [ max_items: <integer> | default = 10_000 ]
+  
+  # configures the number of workers that will process completed edges concurrently.
+  # as edges are completed, they get queued to be collected as metrics for the graph.
+  [ workers: <integer> | default = 10]
+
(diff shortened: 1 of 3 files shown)


annanay25 (approved):
LGTM, nice work @mapno!

docs/configuration/traces-config.md
```diff
@@ -295,6 +295,24 @@ service_graphs:
...
+  #   1. HTTP status is not 2XX
+  #   1. gRPC status code is not OK
+  #   1. span status is Error
+  success_codes:
+    # http status codes not to be considered as failure
+    http:
+      [ - <int> ... ]
+    # grpc status codes not to be considered as failure
+    grpc:
+      [ - <int> ... ]
```
KMiller-Grafana, line 315: This definition could be improved. The numbering is within a comment, so not part of rendering markdown. And, the wording confuses me.  I dislike double negatives ("codes not to be failure" ought to be "codes considered successful"). We should also word it to say something more like "Identifies which status codes are considered successful, independent of what those status codes are defined to mean. For example, an HTTP 404 error code  status can be configured to be considered successful."

PR #1081, opened 2021-11-10
cAdvisor Integration

rgeyer:
#### PR Description
This PR adds an integration for cAdvisor

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 
 - [BUGFIX] Fix usage of POSTGRES_EXPORTER_DATA_SOURCE_NAME when using postgres_exporter integration (@f11r)
 
+- [FEATURE] Added [cAdvisor](https://github.com/google/cadvisor) integration. (@rgeyer)
+
 # v0.21.2 (2021-12-08)
 
 - [SECURITY] This release contains a fix for
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,11 +3,11 @@
 - [FEATURE] (beta) Enable experimental config urls for fetching remote configs. Currently,
    only HTTP/S is supported. Use `-experiment.config-urls.enable` flag to turn this on. (@rlankfo)
 
+- [FEATURE] Added [cAdvisor](https://github.com/google/cadvisor) integration. (@rgeyer)
+
 - [ENHANCEMENT] Traces: Improved pod association in PromSD processor (@mapno)
 
 - [BUGFIX] Fix usage of POSTGRES_EXPORTER_DATA_SOURCE_NAME when using postgres_exporter integration (@f11r)
-
-- [FEATURE] Added [cAdvisor](https://github.com/google/cadvisor) integration. (@rgeyer)
 
 # v0.21.2 (2021-12-08)
 
--- a/docs/configuration/integrations/cadvisor-config.md
+++ b/docs/configuration/integrations/cadvisor-config.md
@@ -58,15 +58,15 @@ Full reference of options:
   [store_container_labels: <boolean> | default = true]
 
   # List of container labels to be converted to labels on prometheus metrics for each container. store_container_labels must be set to false for this to take effect.
-  whitelisted_container_labels:
+  allowlisted_container_labels:
     [ - <string> ]
 
   # List of environment variable keys matched with specified prefix that needs to be collected for containers, only support containerd and docker runtime for now.
-  env_metadata_whitelist:
+  env_metadata_allowlist:
     [ - <string> ]
 
   # List of cgroup path prefix that needs to be collected even when docker_only is specified.
-  raw_cgroup_prefix_whitelist:
+  raw_cgroup_prefix_allowlist:
     [ - <string> ]
 
   # Path to a JSON file containing configuration of perf events to measure. Empty value disabled perf events measuring.
--- a/docs/configuration/integrations/cadvisor-config.md
+++ b/docs/configuration/integrations/cadvisor-config.md
@@ -1,0 +1,115 @@
++++
+title = "cadvisor_config"
++++
+
+# cadvisor_config
+
+The `cadvisor_config` block configures the `cadvisor` integration,
+which is an embedded version of
+[`cadvisor`](https://github.com/google/cadvisor). This allows for the collection of container utilization metrics.
+
+The cAdvisor integration requires some broad privileged permissions to the host. Without these permissions the metrics will not be accessible. This means that the agent must *also* have those elevated permissions.
+
+A good example of the required file, and system permissions can be found in the docker run command published in the [cAdvisor docs](https://github.com/google/cadvisor#quick-start-running-cadvisor-in-a-docker-container).
+
+Full reference of options:
+
+```yaml
+  # Enables the cadvisor integration, allowing the Agent to automatically
+  # collect metrics for the specified github objects.
+  [enabled: <boolean> | default = false]
+
+  # Sets an explicit value for the instance label when the integration is
+  # self-scraped. Overrides inferred values.
+  #
+  # The default value for this integration is inferred from the hostname portion
+  # of api_url.
+  [instance: <string>]
+
+  # Automatically collect metrics from this integration. If disabled,
+  # the cadvisor integration will be run but not scraped and thus not
+  # remote-written. Metrics for the integration will be exposed at
+  # /integrations/cadvisor/metrics and can be scraped by an external
+  # process.
+  [scrape_integration: <boolean> | default = <integrations_config.scrape_integrations>]
+
+  # How often should the metrics be collected? Defaults to
+  # prometheus.global.scrape_interval.
+  [scrape_interval: <duration> | default = <global_config.scrape_interval>]
+
+  # The timeout before considering the scrape a failure. Defaults to
+  # prometheus.global.scrape_timeout.
+  [scrape_timeout: <duration> | default = <global_config.scrape_timeout>]
+
+  # Allows for relabeling labels on the target.
+  relabel_configs:
+    [- <relabel_config> ... ]
+
+  # Relabel metrics coming from the integration, allowing to drop series
+  # from the integration that you don't care about.
+  metric_relabel_configs:
+    [ - <relabel_config> ... ]
+
+  # How frequent to truncate the WAL for this integration.
+  [wal_truncate_frequency: <duration> | default = "60m"]
+
+  #
+  # cAdvisor-specific configuration options
+  #
+
+  # Convert container labels and environment variables into labels on prometheus metrics for each container. If false, then only metrics exported are container name, first alias, and image name.
+  [store_container_labels: <boolean> | default = true]
+
+  # List of container labels to be converted to labels on prometheus metrics for each container. store_container_labels must be set to false for this to take effect.
+  white_listed_container_labels:
+    [ - <string> ]
+
+  # List of environment variable keys matched with specified prefix that needs to be collected for containers, only support containerd and docker runtime for now.
+  env_metadata_whitelist:
+    [ - <string> ]
+
+  # List of cgroup path prefix that needs to be collected even when docker_only is specified.
+  raw_cgroup_prefix_whitelist:
+    [ - <string> ]
+
+  # Path to a JSON file containing configuration of perf events to measure. Empty value disabled perf events measuring.
+  [perf_events_config: <boolean>]
+
+  # resctrl mon groups updating interval. Zero value disables updating mon groups.
+  [resctrl_interval: <int> | default = 0]
+
+  # List of `metrics` to be disabled.
+  disabled_metrics:
+    [ - <string> ]
+
+  # List of `metrics` to be enabled. If set, overrides disabled_metrics
+  enabled_metrics:
+    [ - <string> ]
+  
+  # Length of time to keep data stored in memory
+  [storage_duration: <duration> | default = "2m"]
+
+  # Containerd endpoint
+  [containerd: <string> | default = "/run/containerd/containerd.sock"]
+
+  # Containerd namespace
+  [containerd_namespace: <string> | default = "k8s.io"]
+
+  # Docker endpoint
+  [docker: <string> | default = "unix:///var/run/docker.sock"]
+
+  # Use TLS to connect to docker
+  [docker_tls: <boolean> | default = false]
+
+  # Path to client certificate for TLS connection to docker
+  [docker_tls_cert: <string> | default = "cert.pem"]
+
+  # Path to private key for TLS connection to docker
+  [docker_tls_key: <string> | default = "key.pem"]
+
+  # Path to a trusted CA for TLS connection to docker
+  [docker_tls_ca: <string> | default = "ca.pem"]
+
+  # Only report docker containers in addition to root stats
+  [docker_only: <boolean> | default = false]
+```
--- a/docs/configuration/integrations/cadvisor-config.md
+++ b/docs/configuration/integrations/cadvisor-config.md
@@ -21,10 +21,7 @@ Full reference of options:
 
   # Sets an explicit value for the instance label when the integration is
   # self-scraped. Overrides inferred values.
-  #
-  # The default value for this integration is inferred from the hostname portion
-  # of api_url.
-  [instance: <string>]
+  [instance: <string> | default = <integrations_config.instance>]
 
   # Automatically collect metrics from this integration. If disabled,
   # the cadvisor integration will be run but not scraped and thus not
@@ -61,7 +58,7 @@ Full reference of options:
   [store_container_labels: <boolean> | default = true]
 
   # List of container labels to be converted to labels on prometheus metrics for each container. store_container_labels must be set to false for this to take effect.
-  white_listed_container_labels:
+  whitelisted_container_labels:
     [ - <string> ]
 
   # List of environment variable keys matched with specified prefix that needs to be collected for containers, only support containerd and docker runtime for now.
--- a/pkg/integrations/cadvisor/cadvisor.go
+++ b/pkg/integrations/cadvisor/cadvisor.go
@@ -16,7 +16,7 @@ var DefaultConfig Config = Config{
 	StoreContainerLabels: true,
 	ResctrlInterval:      0,
 	// Matching the default disabled set from cadvisor - https://github.com/google/cadvisor/blob/3c6e3093c5ca65c57368845ddaea2b4ca6bc0da8/cmd/cadvisor.go#L78-L93
-	DisabledMetricsSet: container.MetricSet{
+	disabledMetricsSet: container.MetricSet{
 		container.MemoryNumaMetrics:              struct{}{},
 		container.NetworkTcpUsageMetrics:         struct{}{},
 		container.NetworkUdpUsageMetrics:         struct{}{},
@@ -29,7 +29,7 @@ var DefaultConfig Config = Config{
 		container.ResctrlMetrics:                 struct{}{},
 		container.CPUSetMetrics:                  struct{}{},
 	},
-	EnabledMetricsSet: container.MetricSet{},
+	enabledMetricsSet: container.MetricSet{},
 
 	// Containerd config defaults
 	Containerd:          "/run/containerd/containerd.sock",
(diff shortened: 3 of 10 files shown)


rgeyer:
cAdvisor configuration, combined with the way that they register plugins for container types makes configuring it challenging. This is especially true in the case of supporting multiple integrations of the same type #1062 

cAdvisor configures everything through the `flags` module. When a given plugin is installed, the global code will run, and assign values to global package variables from CLI flags as seen [here](https://github.com/google/cadvisor/blob/master/container/containerd/factory.go#L34-L35).

In a world where there is only one instance of an integration per agent process (arguably the only logical mode for a cAdvisor integration?) I could simply set those globals I.E. `containerd.ArgContainerdEndpoint = c.Containerd`, then use the built in plugin factories and collector manager internal to cAdvisor by more-or-less copying [this code](https://github.com/google/cadvisor/blob/master/cmd/cadvisor.go#L136-L168).

However, if we need to support multiple cAdvisor instances per agent process, I'm going to need to peel back the onion of the cAdvisor logic model a bit more, and probably end up with a lot more boilerplate code instantiating collectors.

Trying to determine how far down the rabbit-hole to go to satisfy our most common cAdvisor use-cases. 🤔

rfratto:
Just curious, what's the affect on binary size when this is included?

pkg/integrations/cadvisor/cadvisor.go
```diff
@@ -0,0 +1,133 @@
...
+		container.HugetlbUsageMetrics:            struct{}{},
+		container.ReferencedMemoryMetrics:        struct{}{},
+		container.CPUTopologyMetrics:             struct{}{},
+		container.ResctrlMetrics:                 struct{}{},
+		container.CPUSetMetrics:                  struct{}{},
+	},
+	EnabledMetricsSet: container.MetricSet{},
+
+	// Containerd config defaults
+	Containerd:          "/run/containerd/containerd.sock",
```
rfratto, line 35: Will this break on non-Linux? Do we need to do something similar to windows_exporter to make this Linux-only?

pkg/integrations/cadvisor/cadvisor.go
```diff
@@ -0,0 +1,133 @@
...
+	PerfEventsConfig string `yaml:"perf_events_config,omitempty"`
+
+	// ResctrlInterval resctrl mon groups updating interval. Zero value disables updating mon groups.
+	ResctrlInterval int `yaml:"resctrl_interval,omitempty"`
+
+	// DisableMetrics list of `metrics` to be disabled.
+	DisabledMetrics []string `yaml:"disabled_metrics,omitempty"`
+
+	// DisabledMetricsSet list of `metrics` to be disabled, in the form required by the cadvisor collector(s)
+	DisabledMetricsSet container.MetricSet
```
rfratto, line 66: is this missing a yaml tag? I'd probably make this un-exported or process it in the integration

pkg/integrations/cadvisor/cadvisor.go
```diff
@@ -0,0 +1,133 @@
...
+
+	// EnableMetrics list of `metrics` to be enabled. If set, overrides 'disable_metrics'.
+	EnabledMetrics []string `yaml:"enabled_metrics,omitempty"`
+
+	// EnabledMetricsSet list of `metrics` to be enabled, in the form required by the cadvisor collector(s)
+	EnabledMetricsSet container.MetricSet
```
rfratto, line 72: Is this missing a yaml tag?

rfratto:
> cAdvisor configures everything through the flags module. When a given plugin is installed, the global code will run, and assign values to global package variables from CLI flags as seen here.

Ah, this is unfortunate. Yeah, we'll need to work with them to stop using globals for configuration or find a way to enforce a single instance for some integrations when #1062 is in place.

pkg/integrations/cadvisor/cadvisor.go
rgeyer: Not missing a tag, it was intended to be private/internal. Same with the enabled one.

rgeyer:
> Just curious, what's the affect on binary size when this is included?

No idea, I been doing `go run` to spot test a few things, and `make agent` is failing for me right now, so I'll have to troubleshoot.

rgeyer:
> > Just curious, what's the affect on binary size when this is included?
> 
> No idea, I been doing `go run` to spot test a few things, and `make agent` is failing for me right now, so I'll have to troubleshoot.

`-rwxr-xr-x 1 ryan ryan 97M Nov  9 17:09 cmd/agent/agent`

rfratto:
> -rwxr-xr-x 1 ryan ryan 97M Nov 9 17:09 cmd/agent/agent

Thanks! And what's the size if you don't include this? (Your size will be specific to your OS / Go version, so I don't want to assume by doing a main build on my end)

rgeyer:
> > -rwxr-xr-x 1 ryan ryan 97M Nov 9 17:09 cmd/agent/agent
> 
> Thanks! And what's the size if you don't include this? (Your size will be specific to your OS / Go version, so I don't want to assume by doing a main build on my end)

Circling back on this..

It turns out, the difference is only a few KB in size, so much so that when using `ls -alh` it truncates both to 97MB.

pkg/integrations/cadvisor/cadvisor.go
rgeyer: It does, indeed on anything non linux. After a couple sloppy attempts with build tags, I *think* got this straight.

rfratto: Your changes for that LGTM 👍

pkg/integrations/cadvisor/cadvisor.go
```diff
@@ -0,0 +1,255 @@
...
+
+// New creates a new cadvisor integration
+func New(logger log.Logger, c *Config) (integrations.Integration, error) {
+	klog.SetLogger(logger)
+
+	// Do gross global configs. This works, so long as there is only one instance of the cAdvisor integration
+	// per host.
+	// Containerd
+	containerd.ArgContainerdEndpoint = &c.Containerd
+	containerd.ArgContainerdNamespace = &c.ContainerdNamespace
```
rgeyer, line 206: So, these are why this is a singleton integration.

I could probably break it down and instantiate the collectors directly with a bit more effort, but this feels "okay" as an MVP.

pkg/integrations/cadvisor/cadvisor.go
```diff
@@ -0,0 +1,255 @@
...
+		return nil, fmt.Errorf("failed to start manager: %w", err)
+	}
+
+	containerLabelFunc := metrics.DefaultContainerLabels
+	if !c.StoreContainerLabels {
+		containerLabelFunc = metrics.BaseContainerLabels(c.WhitelistedContainerLabels)
+	}
+
+	goCol := collectors.NewGoCollector()                                         // This is already emitted by the agent, but not with the integration job name.
+	procCol := collectors.NewProcessCollector(collectors.ProcessCollectorOpts{}) // Same as above
```
rgeyer, line 240: This feels heavy handed. Ultimately the agent emits these metrics on it's own, and should probably own anything that's process/host wide.

That said, the dashboards in the onboarding integrations expect everything to have labels unique to the integration.

Thoughts?

rgeyer:
I have *not* updated the changelog yet. I'll be happy to rebase and add that *just* before merging.

I wanted to allow ample time for review and fixes before I add that tho.

pkg/integrations/cadvisor/cadvisor.go
rlankfo: @rgeyer where are these dashboards located? I just wanted to take a peek 👀

pkg/integrations/cadvisor/cadvisor.go
```diff
@@ -0,0 +1,255 @@
...
+		container.NetworkAdvancedTcpUsageMetrics: struct{}{},
+		container.ProcessSchedulerMetrics:        struct{}{},
+		container.ProcessMetrics:                 struct{}{},
+		container.HugetlbUsageMetrics:            struct{}{},
+		container.ReferencedMemoryMetrics:        struct{}{},
+		container.CPUTopologyMetrics:             struct{}{},
+		container.ResctrlMetrics:                 struct{}{},
+		container.CPUSetMetrics:                  struct{}{},
+	},
+	enabledMetricsSet: container.MetricSet{},
```
rfratto, line 55: nit: this is already the zero value, you don't have to define it here :)

pkg/integrations/cadvisor/cadvisor.go
rfratto: nit: All of the globals being set here should instead be set in the integration's `Run` method the integration instead of New. 

`Run` is guaranteed to never be invoked concurrently for the same integration. However, `New` _may_ be called while `Run` is being invoked, which would potentially lead to a race condition of these globals here. 

Setting them in `Run` will avoid that potential race condition. Changing where these are being initialized might unfortunately push down a lot of logic to Run, though.

pkg/integrations/cadvisor/cadvisor_stub.go
```diff
@@ -0,0 +1,86 @@
...
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+
+	"github.com/grafana/agent/pkg/integrations"
+	"github.com/grafana/agent/pkg/integrations/config"
+)
+
+// DefaultConfig holds the default settings for the cadvisor integration
+var DefaultConfig Config = Config{}
```
rfratto, line 18: nit: we might end up having a few of these stubs :) Can you add a helper somewhere to make having them less boilerplate?

rfratto (approved):
LGTM, though this needs a CHANGELOG entry and docs prior to merging

pkg/integrations/cadvisor/cadvisor.go
rlankfo: I was curious as to why/how they might be using them so this is an interesting find as we may not need this after all... ?

pkg/integrations/cadvisor/cadvisor.go
rgeyer: All good. I *did* have to offload all of the logic to `Run` but it's okay, because it forced me to go back and realize I hadn't stopped the resource manager, which I am now doing after the context ends. 😎

rlankfo (approved):

pkg/integrations/stub_integration.go
```diff
@@ -0,0 +1,73 @@
...
+	_      map[string]interface{} `yaml:"-"`
+}
+
+// UnmarshalYAML implements yaml.Unmarshaler for StubConfig
+func (c *StubConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	*c = StubConfig{}
+	type plain StubConfig
+	err := unmarshal((*plain)(c))
+	return err
+}
```
rgeyer, line 26: This works, in as much as it compiles, and is happy.

However, the desired behavior is that the config will still unmarshal, throwing away undefined values, and only log a warning when `StubConfig.NewIntegration` is called.

In practice, this isn't what happens, as we get yaml unmarshalling errors.

```
.\agent-windows-amd64.exe '-config.file' cadvisorcfg.yml
2021-12-14 16:37:01.286665 I | error loading config file cadvisorcfg.yml: yaml: unmarshal errors:
  line 4: field containerd_namespace not found in type integrations.plain
  line 5: field disabled_metrics not found in type integrations.plain
```

Ideally, I'd like to sort that out. 🤔

pkg/integrations/cadvisor/cadvisor_stub.go
```diff
@@ -0,0 +1,10 @@
+//go:build !linux
+// +build !linux
+
+package cadvisor //nolint:golint
+
+import "github.com/grafana/agent/pkg/integrations"
+
+func init() {
+	integrations.RegisterIntegration(integrations.NewStubConfig(name, "the cadvisor integration only works on linux; enabling it on other platforms will do nothing"))
+}
```
rgeyer, line 10: One liner stub @rfratto 😎

rgeyer:
> LGTM, though this needs a CHANGELOG entry and docs prior to merging

Geez, I knew I hadn't added the Changelog, but I managed to totally forget about the docs. Thanks for the reminder! I'll add both tomorrow.

pkg/integrations/cadvisor/cadvisor_stub.go
rfratto: Love it :) 

You're losing out on the field checking here though (as you've noticed in the other comment. Another way of doing it is the following:

1. Define the Config struct for all platforms, no build tags. 
2. Have two implementations of New: one for Linux, one for everything else. 
3. The "everything else" implementation would do something like the following:

```go
func New(logger log.Logger, c *Config) (integrations.Integration, error) {
   level.Warn(logger).Log("msg", "cadvisor ...") 

   // return something that just implements a fake Run method, running 
   // until the ctx is canceled. 
   // 
   // This might already exist? 
   return integrations.NoOp, nil
}
```

(sorry, this is probably more straightforward but it didn't occur to me until now)

rgeyer: Yeah, the problem with this one in particular, is that my config actually uses modules from cAdvisor, and it won't compile in anything but Linux.

I may have to circle back here on this integration, and try to prune out those bits from my config to clean this up. 🤔

rfratto: Ah, is that because you import `github.com/google/cadvisor/container`? Thankfully it looks like in this case you can replace `container.MetricSet` with a plain `map[string]interface{}` and validate later, but it's not very pretty. 

You might be able to change the stub to workaround the strict unmarshaling, but it wouldn't be pretty either, something like:

```go
// Unmarshal into a map so no field is rejected. 
var raw map[string]interface{}
if err := unmarshal(&raw); err != nil {
  return err 
}

// Convert the map back into YAML so we could unmarshal
// again (lol)
rawBytes, err := yaml.Marshal(raw) 
if err != nil {
  return err 
}

// Then do a final non-strict unmarshal into our stub struct.
return yaml.Unmarshal(rawBytes, (*plain)(c))
```

rgeyer: I decided to remove the `StubConfig` since it was a bad code smell. The *right* pattern here is to design a config that can be used on any platform.

For cAdvisor, I added an accessor func on the config struct that "does the magic" to convert the string slice to the desired native cAdvisor type, and it's only built on windows.

I think this is a pretty workable solution and pattern for integrations like this.

pkg/integrations/stub_integration.go
rgeyer: Resolved by not having a `StubConfig` see [this conversation](https://github.com/grafana/agent/pull/1081#discussion_r769162251)

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
 
 - [BUGFIX] Fix usage of POSTGRES_EXPORTER_DATA_SOURCE_NAME when using postgres_exporter integration (@f11r)
 
+- [FEATURE] Added [cAdvisor](https://github.com/google/cadvisor) integration. (@rgeyer)
```
rfratto, line 10: Can you add this to the top? We try to keep a grouping of FEATURES > ENHANCEMENTS > BUGFIXES > CHANGES > DEPRECATIONS, where each group has things listed in merge order, where newer PRs are at the end of the group.

docs/configuration/integrations/cadvisor-config.md
```diff
@@ -0,0 +1,115 @@
...
+```yaml
+  # Enables the cadvisor integration, allowing the Agent to automatically
+  # collect metrics for the specified github objects.
+  [enabled: <boolean> | default = false]
+
+  # Sets an explicit value for the instance label when the integration is
+  # self-scraped. Overrides inferred values.
+  #
+  # The default value for this integration is inferred from the hostname portion
+  # of api_url.
```
rfratto, line 26: This seems wrong, there is no api_url :)

docs/configuration/integrations/cadvisor-config.md
```diff
@@ -0,0 +1,115 @@
...
+  white_listed_container_labels:
+    [ - <string> ]
+
+  # List of environment variable keys matched with specified prefix that needs to be collected for containers, only support containerd and docker runtime for now.
+  env_metadata_whitelist:
+    [ - <string> ]
+
+  # List of cgroup path prefix that needs to be collected even when docker_only is specified.
+  raw_cgroup_prefix_whitelist:
+    [ - <string> ]
```
rfratto, line 73: nit: Can we have these be allowlist?

pkg/integrations/cadvisor/cadvisor.go
```diff
@@ -0,0 +1,166 @@
...
+	_ "github.com/google/cadvisor/container/containerd/install" // register containerd container plugin
+	_ "github.com/google/cadvisor/container/crio/install"       // register crio container plugin
+	"github.com/google/cadvisor/container/docker"
+	_ "github.com/google/cadvisor/container/docker/install" // register docker container plugin
+	"github.com/google/cadvisor/container/raw"
+	_ "github.com/google/cadvisor/container/systemd/install" // register systemd container plugin
+)
+
+// Matching the default disabled set from cadvisor - https://github.com/google/cadvisor/blob/3c6e3093c5ca65c57368845ddaea2b4ca6bc0da8/cmd/cadvisor.go#L78-L93
+var disabledMetricsSet container.MetricSet = container.MetricSet{
```
rfratto, line 37: Hmm, there's another way you can do this:

```go 
var disabledMetricsSet = *flag.Lookup("disable_metrics").Value.(*container.MetricsSet)
```

That would help you stay in sync with upstream.

pkg/integrations/cadvisor/cadvisor.go
```diff
@@ -0,0 +1,166 @@
...
+
+	// Raw
+	raw.DockerOnly = &i.c.DockerOnly
+
+	// Only using in-memory storage, with no backup storage for cadvisor stats
+	memoryStorage := memory.New(i.c.StorageDuration, []storage.StorageDriver{})
+
+	sysFs := sysfs.NewRealSysFs()
+
+	collectorHTTPClient := http.Client{}
```
rfratto, line 113: nit: use `var` for zero value declarations: 

```go
var collectorHTTPClient http.Client
```

(From [uber-go/style](https://github.com/uber-go/guide/blob/master/style.md))

pkg/integrations/stub_integration.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"context"
+	"net/http"
+
+	"github.com/grafana/agent/pkg/integrations/config"
+)
+
+// StubIntegration implements a no-op integration for use on platforms not supported by an integration
+type StubIntegration struct {
+}
```
rfratto, line 12:
```suggestion
type StubIntegration struct{}
```

rfratto (approved):
nits aside, this looks good to me. Two things I'd like you to check on before merging:

1. Running this doesn't add any Prometheus metrics to the global registry used by the agent 
2. Running this alongside node_exporter doesn't cause any flag collisions (as both are writing flags to the global flagset, IIUC)

docs/configuration/integrations/cadvisor-config.md
rgeyer: Yes.. But.. That would make the terminology different than the upstream project. I'm *certain* that the meaning won't be lost, but I dunno if this might be confusing to those who are used to using cAdvisor directly.🤔

rfratto: We can reference the original flag names in the comment, if it helps

CHANGELOG.md
rgeyer: > 1. Running this doesn't add any Prometheus metrics to the global registry used by the agent

How would I validate this one? Just run it without the integration enabled, and check `/metrics`? Then compare that to the metrics emitted when the integration is enabled?

rgeyer: Also, you have NO IDEA how long I agonized over where in the changelog this should go, and I still managed to get it wrong 😂 🤦‍♂️

rfratto: > How would I validate this one? Just run it without the integration enabled, and check /metrics? Then compare that to the metrics emitted when the integration is enabled?

Yeah, that seems close enough. Some kind of `wc -l` check. It might be off by a few, I'm just looking to make sure there's not like, hundreds more.

> Also, you have NO IDEA how long I agonized over where in the changelog this should go, and I still managed to get it wrong 😂 🤦‍♂️

No worries, it's something we don't document well yet :)

docs/configuration/integrations/cadvisor-config.md
rgeyer: Meh, I'm good changing this all to allowlist, and waiting to see if people complain ;)

pkg/integrations/cadvisor/cadvisor.go
rgeyer: Ohh.. I see why I didn't do this. Because I'd have to import the `github.com/google/cadvisor/cmd` package. IIRC that caused some conflicts and wasn't worth the drama :(

rfratto: Ahhh, I didn't realize it was from another package. That's fine, this is good as-is.

rgeyer:
> 1. Running this doesn't add any Prometheus metrics to the global registry used by the agent

```
root@18n1l:/home/ryan# wc -l without-cadvisor.om
328 without-cadvisor.om
root@18n1l:/home/ryan# wc -l with-cadvisor.om
462 with-cadvisor.om
```

> 2. Running this alongside node_exporter doesn't cause any flag collisions (as both are writing flags to the global flagset, IIUC)

No concise way to "show" my output on this, but I was able to enable both, and hit both `/integrations/node_exporter/metrics` and `/integrations/cadvisor/metrics` and get the expected outputs.

pkg/integrations/cadvisor/cadvisor.go
rgeyer: Put it back for now. Addressed everything else. When the build steps pass, I'm gonna go ahead and merge this.

Thanks for the repeated reviews @rfratto :)

PR #1085, opened 2021-11-10
handle retrieve_config failures

### LINKED ISSUE #599: agent does not detect a failure to download the configuration
When the Grafana cloud agent is installed via the https://raw.githubusercontent.com/grafana/agent/release/production/grafanacloud-install.sh script, there is no checking to make sure the configuration downloaded successfully in `retrieve_config` 

In some circumstances the download can fail, (such as running on non a shell where variable substation does not work in the same way), but this is not detected, and the user sees an ugly error. 

```--- Retrieving config and placing in /etc/grafana-agent.yaml
could not retrieve agent cloud config: failed to make request: Get "https://integrations-api.grafana.net/stacks/xxxxxx/agent_config": context deadline exceeded
Failed to retrieve config
--- Enabling and starting grafana-agent.service
```

rlankfo:
#### PR Description 

A bug in `production/grafanacloud-install.sh` results in the script not exiting properly when `retrieve_config` fails. It also prints an ugly error message which this hides.

Fixes #599 

#### Notes to the Reviewer

Typically I would use something like `set -o pipefail` in bash scripts to handle a case like this but I believe that's not POSIX compliant nor available in dash.

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/production/grafanacloud-install.sh
+++ b/production/grafanacloud-install.sh
@@ -13,6 +13,8 @@
 # PACKAGE_SYSTEM can be passed as an environment variable with either rpm or
 # deb.
 set -eu
+trap "exit 1" TERM
+MY_PID=$$
 
 log() {
   echo "$@" >&2
@@ -20,7 +22,7 @@ log() {
 
 fatal() {
   log "$@"
-  exit 1
+  kill -s TERM "$MY_PID"
 }
 
 #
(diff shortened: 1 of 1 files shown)


production/grafanacloud-install.sh
```diff
@@ -73,7 +73,21 @@ main() {
...
+  # This is a POSIX-compliant work around for dash not having pipefail shellopt.
+  local t
+  t=$(mktemp -d)
+  mkfifo "$t/config"
+  # SC2024 can be safely ignored as we want to redirect using the current user's
+  # permissions. See https://github.com/koalaman/shellcheck/issues/2070
+  # shellcheck disable=SC2024
+  sudo tee /etc/grafana-agent.yaml < "$t/config"  &
+  if ! retrieve_config 2>/dev/null 1>"$t/config"; then
+    fatal 'Failed to retrieve config'
```
rfratto, line 87: Would it make sense if `fatal` was changed to [signal an abort to the top-level script](https://stackoverflow.com/a/9894126)? That might help clean this up a little bit and make it easier to follow, if it fixes the problem at all.

rlankfo: @rfratto that would work as well. would you prefer I update this to use that approach?

rfratto: I think it might be easier to follow, but I'll leave it up to you. LMK when you want me to merge 👍

rlankfo: updated 👍

rlankfo:
@rfratto once all checks have completed, this should be ready to merge

rfratto (approved):
Thank you!

PR #1094, opened 2021-11-12
Update go-kit logger package to remove debug logs

mapno:
#### PR Description 

Some upstream code is now using `go-kit/log` instead of `go-kit/kit/log`, some debug logging from that code will pass through the level filter. `go-kit/kit` v0.12.0 addresses this by making `go-kit/kit/log` just a pass-through to `go-kit/log`, but upgrading to that version causes an issue while importing `github.com/prometheus/procfs/sysfs`.

#### PR Checklist

- NA CHANGELOG updated 
- [x] Documentation added
- NA Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,8 +18,6 @@
 
 - [BUGFIX] Operator: Fix relabel_config directive for PodLogs resource (@hjet)
 
-- [BUGFIX] Update go-kit logger package to remove debug logs (@mapno)
-
 - [CHANGE] Self-scraped integrations will now use an SUO-specific value for the `instance` label. (@rfratto)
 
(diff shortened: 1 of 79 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,8 @@
 
 - [BUGFIX] Operator: Fix relabel_config directive for PodLogs resource (@hjet)
 
+- [BUGFIX] Update go-kit logger package to remove debug logs (@mapno)
```
rfratto, line 21: This was introduced prior to a release so a CHANGELOG entry isn't necessary, can you remove?

rfratto (approved):
Thanks!

CHANGELOG.md
mapno: Oh that's right. Removed

PR #1102, opened 2021-11-15
upgrade to loki v2.4.1

### LINKED ISSUE #1127: Update to Loki v2.4.1
We're currently using a non-release commit of Loki (for Promtail) between v2.3.x and v2.4.1. We should update to the latest release (v2.4.1) to bring in new Promtail features like grafana/loki#4495. 

This change may be non-trivial due to the change Loki's migration over to grafana/dskit. 

cc @rlankfo

rlankfo:
#### PR Description 

* upgrade to loki v2.4.1. 
* replace cortex ring with dskit ring package

Fixes #1127 

#### Notes to the Reviewer

The `github.com/cortexproject/cortex/pkg/ring` has been migrated to a more generic version of the ring in `github.com/grafana/dskit/ring`. Loki has moved to this version. There were some minor API differences this PR addresses, and this ring version uses `promauto` so we no longer need the calls to register. I believe we could also remove explicit calls to Unregister but I have not done that here.

#### PR Checklist

- [x] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/metrics/cluster/node.go
+++ b/pkg/metrics/cluster/node.go
@@ -53,7 +53,7 @@ type node struct {
 // newNode creates a new node and registers it to the ring.
 func newNode(reg prometheus.Registerer, log log.Logger, cfg Config, s pb.ScrapingServiceServer) (*node, error) {
 	n := &node{
-		reg: util.WrapWithUnregisterer(prometheus.WrapRegistererWithPrefix("cortex_", reg)),
+		reg: util.WrapWithUnregisterer(reg),
 		srv: s,
 		log: log,
 
@@ -120,7 +120,7 @@ func (n *node) ApplyConfig(cfg Config) error {
 	}
 	n.ring = r
 
-	lc, err := ring.NewLifecycler(cfg.Lifecycler, n, "agent", agentKey, false, n.log, n.reg)
+	lc, err := ring.NewLifecycler(cfg.Lifecycler, n, "agent", agentKey, false, n.log, prometheus.WrapRegistererWithPrefix("cortex_", n.reg))
 	if err != nil {
 		return fmt.Errorf("failed to create lifecycler: %w", err)
 	}
@@ -152,7 +152,7 @@ func newRing(cfg ring.Config, name, key string, reg prometheus.Registerer, log l
 	if err != nil {
 		return nil, err
 	}
-	return ring.NewWithStoreClientAndStrategy(cfg, name, key, store, ring.NewIgnoreUnhealthyInstancesReplicationStrategy(), reg, log)
+	return ring.NewWithStoreClientAndStrategy(cfg, name, key, store, ring.NewIgnoreUnhealthyInstancesReplicationStrategy(), prometheus.WrapRegistererWithPrefix("cortex_", reg), log)
 }
 
 // run waits for connection to the ring and kickstarts the join process.
--- a/pkg/metrics/cluster/node.go
+++ b/pkg/metrics/cluster/node.go
@@ -8,7 +8,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/cortexproject/cortex/pkg/ring"
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	"github.com/gorilla/mux"
@@ -17,6 +16,7 @@ import (
 	"github.com/grafana/agent/pkg/util"
 	"github.com/grafana/dskit/backoff"
 	"github.com/grafana/dskit/kv"
+	"github.com/grafana/dskit/ring"
 	"github.com/grafana/dskit/services"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/weaveworks/common/user"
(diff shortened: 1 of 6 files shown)


pkg/metrics/cluster/node.go
```diff
@@ -53,12 +53,13 @@ type node struct {
 // newNode creates a new node and registers it to the ring.
 func newNode(reg prometheus.Registerer, log log.Logger, cfg Config, s pb.ScrapingServiceServer) (*node, error) {
 	n := &node{
-		reg: util.WrapWithUnregisterer(reg),
+		reg: util.WrapWithUnregisterer(prometheus.WrapRegistererWithPrefix("cortex_", reg)),
```
rfratto, line 56: Are these equivalent to what it did before?

rlankfo: Yes that's correct. You can see Loki made the same changes when they migrated as well: https://github.com/grafana/loki/pull/4641/files#diff-8e51742acf8f88fa2ae01855fb19125feb5c54c41023f1eaa8beded9f1b4bcc3R102

rlankfo: cc: @treid314

pkg/metrics/cluster/node.go
```diff
@@ -114,15 +115,12 @@ func (n *node) ApplyConfig(cfg Config) error {
...
-	if err := n.reg.Register(r); err != nil {
-		return fmt.Errorf("failed to register ring metrics: %w", err)
-	}
 	if err := services.StartAndAwaitRunning(context.Background(), r); err != nil {
 		return fmt.Errorf("failed to start ring: %w", err)
 	}
 	n.ring = r
 
-	lc, err := ring.NewLifecycler(cfg.Lifecycler, n, "agent", agentKey, false, n.reg)
+	lc, err := ring.NewLifecycler(cfg.Lifecycler, n, "agent", agentKey, false, n.log, n.reg)
```
treid314, line 123: You'll want to wrap the registerer here as well. Since we removed the `cortex_` prefix on the ring metrics in dskit  you'll want to wrap with the `cortex_` prefix here to match the previous cortex metrics.

pkg/metrics/cluster/node.go
```diff
@@ -150,10 +148,11 @@ func newRing(cfg ring.Config, name, key string, reg prometheus.Registerer, log l
 		kv.RegistererWithKVName(reg, name+"-ring"),
 		log,
 	)
+
 	if err != nil {
 		return nil, err
 	}
-	return ring.NewWithStoreClientAndStrategy(cfg, name, key, store, ring.NewIgnoreUnhealthyInstancesReplicationStrategy())
+	return ring.NewWithStoreClientAndStrategy(cfg, name, key, store, ring.NewIgnoreUnhealthyInstancesReplicationStrategy(), reg, log)
```
treid314, line 155: You'll want to wrap the registerer here as well. Since we removed the cortex_ prefix on the ring metrics in dskit you'll want to wrap with the cortex_ prefix here to match the previous cortex metrics.

pkg/metrics/cluster/node_test.go
```diff
@@ -169,7 +171,7 @@ func startNode(t *testing.T, srv agentproto.ScrapingServiceServer) {
 	lcConfig.Addr = l.Addr().(*net.TCPAddr).IP.String()
 	lcConfig.Port = l.Addr().(*net.TCPAddr).Port
 
-	lc, err := ring.NewLifecycler(lcConfig, ring.NewNoopFlushTransferer(), "agent", "agent", false, prometheus.NewRegistry())
+	lc, err := ring.NewLifecycler(lcConfig, ring.NewNoopFlushTransferer(), "agent", "agent", false, logger, prometheus.NewRegistry())
```
treid314, line 174: You can wrap here if it matters for the tests.

pkg/metrics/cluster/node.go
treid314: I'm not sure if you want to wrap here since you'll wrap that registerer for all metrics registered with this n.reg with the `cortex_` prefix

pkg/metrics/cluster/node_test.go
treid314: Or if we don't use the registry in tests you can just pass `nil` instead of `prometheus.NewRegistry()`

pkg/metrics/cluster/node_test.go
```diff
@@ -169,7 +171,7 @@ func startNode(t *testing.T, srv agentproto.ScrapingServiceServer) {
 	lcConfig.Addr = l.Addr().(*net.TCPAddr).IP.String()
 	lcConfig.Port = l.Addr().(*net.TCPAddr).Port
 
-	lc, err := ring.NewLifecycler(lcConfig, ring.NewNoopFlushTransferer(), "agent", "agent", false, prometheus.NewRegistry())
+	lc, err := ring.NewLifecycler(lcConfig, ring.NewNoopFlushTransferer(), "agent", "agent", false, logger, prometheus.WrapRegistererWithPrefix("cortex_", prometheus.NewRegistry()))
```
treid314, line 174: Actually in thinking, you can pass `nil` here in place of the registry since we don't even save it off to a variable, so I don't think we even look at it.

rfratto:
I'd like to return to this, but I think given the risk of upgrading to dskit means we should wait for #1326 first so we can be notified if something breaks here.

rlankfo:
Closing in favor of https://github.com/grafana/agent/pull/1422

PR #1103, opened 2021-11-16
Setup CD for agent

gouthamve:
This deploys agent to dev on a merge to main.

Depends on https://github.com/grafana/deployment_tools/pull/19936

Currently this automation only opens a PR, but I want to double check things and then I'll deploy directly to dev without a PR.

Signed-off-by: Goutham Veeramachaneni <gouthamve@gmail.com>

### DIFF
--- a/.drone/drone.yml
+++ b/.drone/drone.yml
@@ -129,6 +129,32 @@ volumes:
 ---
 kind: pipeline
 type: docker
+name: Deploy-To-Deployment-Tools 
+platform:
+  os: linux
+  arch: amd64
+trigger:
+  ref:
+    - refs/heads/main
+
+steps:
+  - name: Update Deployment Tools 
+    image: us.gcr.io/kubernetes-dev/drone/plugins/updater
+    settings:
+      config_json: "{\r\n  \"destination_branch\": \"master\",\r\n  \"pull_request_branch_prefix\": \"cd-agent-\",\r\n  \"pull_request_enabled\": true,\r\n  \"pull_request_existing_strategy\": \"fail\",\r\n  \"pull_request_message\": \"Created by Drone build https:\/\/drone.grafana.net\/api\/repos\/grafana\/agent\/builds\/${DRONE_BUILD_NUMBER}\\n\\nBuild commit: ${DRONE_COMMIT_LINK}\\n\",\r\n  \"pull_request_team_reviewers\": [\r\n    \"agent-squad\"\r\n  ],\r\n  \"repo_name\": \"agent\",\r\n  \"update_jsonnet_attribute_configs\": [\r\n    {\r\n      \"file_path\": \"ksonnet\/environments\/platform-observability\/images.libsonnet\",\r\n      \"jsonnet_key\": \"agent\",\r\n      \"jsonnet_value\": \"grafana\/agent:$DRONE_TAG\"\r\n    }\r\n  ]\r\n}"
+      github_token:
+        from_secret: gh_token
+ 
+depends_on:
+  - Containerize
+
+volumes:
+  - name: docker
+    host:
+      path: /var/run/docker.sock
+---
+kind: pipeline
+type: docker
 name: Release
 platform:
   os: linux
--- a/.drone/drone.yml
+++ b/.drone/drone.yml
@@ -138,6 +138,10 @@ trigger:
     - refs/heads/main
 
 steps:
+  - name: put image tag in a file
+    image: alpine
+    commands:
+      - echo grafana/agent:$(shell ./tools/image-tag) > .image-tag
   - name: Update Deployment Tools 
     image: us.gcr.io/kubernetes-dev/drone/plugins/updater
     settings:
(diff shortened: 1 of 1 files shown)


.drone/drone.yml
```diff
@@ -122,6 +122,32 @@ steps:
...
+  arch: amd64
+trigger:
+  ref:
+    - refs/heads/main
+
+steps:
+  - name: Update Deployment Tools 
+    image: us.gcr.io/kubernetes-dev/drone/plugins/updater
+    settings:
+      config_json: "{\r\n  \"destination_branch\": \"master\",\r\n  \"pull_request_branch_prefix\": \"cd-agent-\",\r\n  \"pull_request_enabled\": true,\r\n  \"pull_request_existing_strategy\": \"fail\",\r\n  \"pull_request_message\": \"Created by Drone build https:\/\/drone.grafana.net\/api\/repos\/grafana\/agent\/builds\/${DRONE_BUILD_NUMBER}\\n\\nBuild commit: ${DRONE_COMMIT_LINK}\\n\",\r\n  \"pull_request_team_reviewers\": [\r\n    \"agent-squad\"\r\n  ],\r\n  \"repo_name\": \"agent\",\r\n  \"update_jsonnet_attribute_configs\": [\r\n    {\r\n      \"file_path\": \"ksonnet\/environments\/platform-observability\/images.libsonnet\",\r\n      \"jsonnet_key\": \"agent\",\r\n      \"jsonnet_value\": \"grafana\/agent:$DRONE_TAG\"\r\n    }\r\n  ]\r\n}"
```
rfratto, line 144: Haha, this makes me wish we were using sonnet to generate this.

rfratto (commented):
LGTM, though you'll need to run `make drone` to sign it. If you don't have DRONE_TOKEN set up in your environment, you'll need to have that defined in advance.

.drone/drone.yml
```diff
@@ -122,6 +122,32 @@ steps:
...
+  os: linux
+  arch: amd64
+trigger:
+  ref:
+    - refs/heads/main
+
+steps:
+  - name: Update Deployment Tools 
+    image: us.gcr.io/kubernetes-dev/drone/plugins/updater
+    settings:
```
rfratto, line 143: For the curious, these are [converted](https://docs.drone.io/plugins/overview/#plugin-inputs) into `PLUGIN_<settings key>` environment variables for the container to use.

.drone/drone.yml
jvrplmlmn, line 144: You can do this as a multi-line config, something like:

```
  settings:
    config_json: |-
      {
        "key1": "value1",
        "key2": "value2"
      }
```

gouthamve:
@rfratto and @jvrplmlmn Thanks for the reviews, I've updated the PR! I think this is good to go, but I'll only know if the tags and images are sane once this is merged. Once I confirm that I'll remove the PR requirement and directly commit to `master`.

.drone/drone.yml
```diff
@@ -122,6 +122,52 @@ steps:
...
+            "rfratto",
+            "mattdurham",
+            "gouthamve"
+          ],
+          "repo_name": "deployment_tools",
+          "update_jsonnet_attribute_configs": [
+            {
+              "file_path": "ksonnet/environments/platform-observability/waves/agent.libsonnet",
+              "jsonnet_key": "dev_canary",
+              "jsonnet_value": "grafana/agent:$DRONE_TAG"
```
jvrplmlmn, line 161: If I understand https://docs.drone.io/pipeline/environment/reference/drone-tag/ correctly ,`DRONE_TAG` will be empty for this pipeline for some of the events, see the trigger section:

```
trigger:
  ref:
    - refs/heads/main
```

You are not filtering on a specific event (such as: `push`, `pull_request` or `tag`) so it would apply to all fo them.

.drone/drone.yml
```diff
@@ -122,6 +122,52 @@ steps:
...
+            }
+          ]
+        }
+      github_token:
+        from_secret: gh_token
+ 
+depends_on:
+  - Containerize
```
jvrplmlmn, line 169: Looking at the whole `drone.yml` plus the Makefile, I think that this won't work and you should depend instead on `Release`. If I understand correctly, the `Containerize` builds but does not push the Docker image to the registry, that is done in `Release` via `make ... publish`.

rfratto: Containerize does push the images through dockerx, release is for non-container release assets :)

jvrplmlmn: Then you are golden 👍

gouthamve: Yep, I was looking at the logs and noticed that `Containerize` does push things.

.drone/drone.yml
gouthamve: Hrm, this is true. I need to somehow get `IMAGE_TAG ?= $(shell ./tools/image-tag)` in there. Not sure how. Looking at how other repos are doing it.

rfratto: I feel responsible for this being a pain so I opened grafana/deployment_tools#20048 to add `jsonnet_value_file` support. 

Once merged and the updated image gets deployed, we can use an initial stage to write the output of `$(shell ./tools/image-tag)` to a file and use the file name in the config_json instead.

rfratto: That PR has been merged and the docker image is updated to include the new feature.

gouthamve:
So this works! And it directly commits to master and doesn't open a PR. Which is what we want for the dev rollouts.

.drone/drone.yml
```diff
@@ -122,6 +122,56 @@ steps:
...
+    image: us.gcr.io/kubernetes-dev/drone/plugins/updater
+    settings:
+      config_json: |- 
+        {
+          "destination_branch": "master",
+          "pull_request_branch_prefix": "cd-agent",
+          "pull_request_enabled": false,
+          "pull_request_team_reviewers": [
+            "agent-squad"
+          ],
```
rfratto, line 159: We can probably take this part out, but I think it's useful to have it in for when we want to copy this for non-dev flows

rfratto (approved):
Love it!

mattdurham (approved):
Looks reasonable

PR #1106, opened 2021-11-17
0.21.0 release prep

rlankfo:
Prepare for v0.21.0 release by following the [release procedure](https://github.com/grafana/agent/blob/main/cmd/agent/DEVELOPERS.md#performing-the-release)

### DIFF
--- a/pkg/operator/defaults.go
+++ b/pkg/operator/defaults.go
@@ -15,6 +15,7 @@ var (
 		"v0.18.4",
 		"v0.19.0",
 		"v0.20.0",
+		"v0.21.0",
 
 		// NOTE(rfratto): when performing an upgrade, add the newest version above instead of changing the existing reference.
 	}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Main (unreleased)
 
+# v0.21.0 (2021-11-17)
+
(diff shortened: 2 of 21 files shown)


rfratto:
Your branch is three commits behind main, can you fast forward?

pkg/operator/defaults.go
```diff
@@ -14,7 +14,7 @@ var (
 		"v0.18.3",
 		"v0.18.4",
 		"v0.19.0",
-		"v0.20.0",
+		"v0.21.0",
```
rfratto, line 17: This should add a new version instead of changing the old one

rfratto (commented):
LGTM aside from two comments

pkg/operator/defaults.go
rlankfo: woops, did some find / sed here and missed excluded this one.

rlankfo:
> Your branch is three commits behind main, can you fast forward?

fixed. did not fetch upstream before creating my branch locally.

rfratto (approved):
LGTM, though now that you pulled there's one more v0.20.0 to update :)

rlankfo:
> LGTM, though now that you pulled there's one more v0.20.0 to update :)

got it 😅

mattdurham (approved):
Lgtm assuming we don’t sneak anymore changes in :)

PR #1117, opened 2021-11-18
Default to using latest agent release version in operator

### LINKED ISSUE #1029: Operator: GrafanaAgent CRD should default to latest known version
Reconciling the GrafanaAgent CRD should deploy the latest known version of the agent. This is not currently the case: the version field is used but not assigned a default if it is empty. This leads to an invalid image reference if `version` is not set in the CRD.

(First noticed in #1026)

As an alternative to this, the version field can be made required, but I think defaulting to the latest known release makes more sense.

rfratto:
Fixes #1029

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 - [BUGFIX] Statsd Integration: Fix issue where mapped metrics weren't exposed to the integration. (@mattdurham)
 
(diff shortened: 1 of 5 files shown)


mattdurham (approved):
Lgtm

PR #1119, opened 2021-11-18
Don't create a WAL cleaner when there's no configured WAL directory.

### LINKED ISSUE #735: WAL watcher runs even when Prometheus not in use
The WAL watcher will run even when Prometheus is not in use. This will lead to errors if it defaults to reading a non-empty directory (ie `/`): 

```
ts=2021-07-12T15:59:58.668240754Z level=warn agent=prometheus component=cleaner msg="unable to find segment mtime of WAL" name=sys err="unable to open WAL: open sys/wal: no such file or directory"
```

Either the WAL watcher should be disabled if no Prometheus configs are present, or it should default to an empty directory, or that warning-level log should be a debug-level log.

rfratto:
#### PR Description 

This avoids noisy logs when someone uses the agent only for logs/traces.

Fixes #735

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 - [BUGFIX] Operator: fix bug where version was a required field (@rfratto)
 
+- [BUGFIX] Metrics: Only run WAL cleaner when metrics are being used and a WAL is configured. (@rfratto)
+
 # v0.21.0 (2021-11-17)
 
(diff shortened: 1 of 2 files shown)


mattdurham (approved):
Lgtm

mapno (approved):
Nice!

Issue #1129, opened 2021-11-23
[dev.multiple-integrations] Proposal: singleton-only integrations


rfratto:
While we are working on supporting multiple instances of integrations, it's becoming more clear that:

* Some integrations don't make sense to support more than one instance 
* Some integrations require many upstream changes to be runnable multiple times in process 

I propose that some integrations be marked as singletons. These would normally be your machine-wide integrations. As of today, the following would be singletons:

* agent
* node_exporter
* windows_exporter 
* process_exporter

Singleton integrations would _not_ support the new `*_configs` syntax proposed in #1059. 

This change also has implications for #883. The specifics of how the operator could handle singleton integrations is out of scope, but as an example, it could reject reconciling if it discovers multiple `type: per-node` integrations of the same name.

rfratto:
Closed by #1196.

PR #1130, opened 2021-11-23
[dev.multiple-integrations] pkg/integrations: support *_configs field for integrations

rfratto:
#### PR Description 
Creates the basic code to unmarshal integrations from a YAML field called <integration name>_configs, which is a slice of that integration.

Note that this is NOT wired up to the integrations manager yet, and trying to run the agent with more than one integration of the same type will likely cause problems.

#### Notes to the Reviewer
As a side effect to this PR, this config file:

```yaml
integrations:
  memcached_exporter:
    memcached_address: localhost:11211 
```

will always marshal into the array form (i.e., when returned from `/-/config`):

```yaml
integrations:
  memcached_exporter_configs:
    - memcached_address: localhost:11211 
```

This might be a little confusing to users. I'm open to suggestions for how it could be handled differently.

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated


rfratto:
> This might be a little confusing to users. I'm open to suggestions for how it could be handled differently.

Actually, I think #1129 may help and remove the problem. We could store instantiations for singleton integrations separately, which would be any integration from a non-`*_config` form. That would help with being consistent with the marshaling.

mattdurham:
I think having the config look a bit different from what is passed in is fine. I think there can be a difference in what is passed in versus how the agent sees the config has long as they are functionally the same.

mattdurham (approved):
Looks solid

PR #1132, opened 2021-11-24
Remove WAL on startup

gouthamve:
Agent / Remote Write code currently doesn't send the samples from
the WAL on startup at all. We only send the new samples so replaying
the WAL doesn't give us much.

It does let us load series references from the WAL but we save much more
by not reading the WAL at all.

See: https://github.com/prometheus/prometheus/issues/9848 and https://github.com/prometheus/prometheus/pull/8918

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/metrics/wal/wal.go
+++ b/pkg/metrics/wal/wal.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"math"
+	"os"
 	"sync"
 	"time"
 	"unicode/utf8"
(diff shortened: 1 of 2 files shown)


rfratto:
Related: I just opened https://github.com/prometheus/prometheus/pull/9862 to enable support for remote_write starting to queue samples from an older existing segment.

mattdurham (changes requested):
Need to check reload/scraping service behavior

Issue #888, opened 2021-09-04
metrics: Remove internal instance manager system


rfratto:
The [instance manager](https://github.com/grafana/agent/blob/2007c468203241f6bff596fbfa22fc54fc743e54/pkg/metrics/instance/manager.go#L37) system manages a set of [instance.Configs](https://github.com/grafana/agent/blob/2007c468203241f6bff596fbfa22fc54fc743e54/pkg/metrics/instance/instance.go#L60) at runtime with a CRUD-ish  interface. I believe that it's an overcomplicated component of the metrics subsystem: it's hard to understand, has performance implications at scale, and likely doesn't really need to exist. The performance limitations of the way the system was designed starts after being combined with a [GroupManager](https://github.com/grafana/agent/blob/2007c468203241f6bff596fbfa22fc54fc743e54/pkg/metrics/instance/group_manager.go#L28), where if you have 20,000 configs that group to a single instance, the total time to apply the full set of configs grows exponentially as each intermediate application must update the underlying Prometheus components. 

I propose replacing the instance manager system with a simpler mechanism that requires the full set of `instance.Configs` that should be ran all at once. This _may_ be done in `agent.ApplyConfig`, but until the scraping service is replaced (which requires the ability to dynamically update configs at runtime), we will need some kind of separate method just for accepting instance configs (like `agent.ApplyInstances`). 

Aside from resolving #849, this change is fully internal and will be invisible to users.

This change will push the responsibility of defining the full set of configs beyond individual subsystems and up to the agent's `main` method. For example, integrations, another subsystem which currently depends on the instance manager, will need to expose the configs for `main` to combine with those statically defined in the config file. This static set will then be combined with the runtime set used by the scraping service until it can be replaced (#875, #873, #872, #870, #869). 

Note that the metrics subsystem will still require a way to expose the underlying instances, as they are used by the tracing subsystem to write metrics directly to a WAL. That the lookups use names from the config file, so a mapping of original names to the running instance will need to be maintained if instance sharing is performed.

rfratto:
I'm going to close this as unplanned. This work has been superseded by the recent work on Flow.

Issue #873, opened 2021-09-01
Reduce operational modes to one


rfratto:
The number of ways you can operate the Grafana Agent is by far the most confusing aspect of it. Metrics supports "Prometheus-compatible mode," "host filtering mode," and "scraping service mode." This complexity is pushed onto users to understand which mode is best for them, and it also makes it harder for the developers to test and maintain three separate code paths. 

I propose removing host filtering mode, and have the metrics code path _always_ use the clustering code path. This depends on per-target sharding (#872) to be completed. The "prometheus-compatible mode" will always exist, but it is now equivalent to having an unclustered agent.  

Host filtering can be manually supplanted by not clustering agents and using environment variables in the configuration file with relabel_configs that ignore targets from different machines.

rfratto:
I'm going to close this; Flow supersedes this work and only has one "operational mode."

PR #1140, opened 2021-11-29
[RFC] New Metrics subsystem

### LINKED ISSUE #872: Per-target sharding
The scraping service shards on individual configuration files and _not_ discovered targets. This means efficient use of the scraping service requires users to have [single static configurations for each target](https://github.com/grafana/agent/blob/v0.16.0/docs/scraping-service.md#best-practices). 

This also caused the need for instance sharing, which has performance issues at scale (amplifying the effects of #849), and confuses users due to how it renames remote_write configuration blocks. 

I propose a second layer of sharding for clustered agents, one that shards individual targets and distributes per-target load amongst the cluster. Note that this treats each target as equal, so very large targets with a lot of series would still cause some imbalance. 

I have a design document for per-target sharding [here](https://docs.google.com/document/d/1JI804iaut6bKvZprOydes3Gb5Awo_J0sX-3ORlyc5l0/edit). It has also been partially prototyped [here](https://github.com/rfratto/agent/tree/per-target-sharding), though the prototype uses Pastry for clustering, which I would recommend swapping out with [rfratto/ckit](https://github.com/rfratto/ckit) for the "real" implementation.

rfratto:
Proposes a new metrics subsystem to resolve all of #872, #873, #875, and #888. 

Historically based on the existing [per-target sharding][per-target sharding] proposal, though has some changes to simplify the implementation.  

[per-target sharding]: https://docs.google.com/document/d/1JI804iaut6bKvZprOydes3Gb5Awo_J0sX-3ORlyc5l0

### DIFF
--- a/docs/rfcs/0003-new-metrics-subsystem.md
+++ b/docs/rfcs/0003-new-metrics-subsystem.md
@@ -1,12 +1,9 @@
----
-title: New metrics subsystem
-weight: xxxx
----
-
 # New metrics subsystem
 
 * Date: 2021-11-29
 * Author: Robert Fratto (@rfratto)
+* PR: [grafana/agent#1140](https://github.com/grafana/agent/pull/1140)
+* Status: Draft
 
 ## Background
 
@@ -25,12 +22,10 @@ genuinely beneficial for the maintenance of the project.
 
 This proposal suggests a redesign of the metrics subsystem which has native
 support for target sharding and lacks the technical debt from the current
-subsystem. Given the size of the change, I also propose a way to gradually
-implement the new subsystem to ensure that this proposal is practical.
+subsystem.
 
 ## Goals
 
-* Propose a strategy to implement this revamp incrementally
 * Enable dynamically target scraping with:
   * Automatic scaling
   * Automatic failover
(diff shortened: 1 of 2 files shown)


mattdurham:
Is there an inherent relationship between discovery sources and targets, so that if a discovery source changes those changes are propagated downstream? IE discovery source is no longer valid so all the targets need to be removed from the running systems?

rfratto:
> Thoughts on splitting up the config distribution and target distribution into separate items/executables? Then the agent can assume it is consuming the already "distributed" targets. This feels like a lot of responsibility in one executable.
 
It's not impossible, this is what OpenTelemetry Operator is going to do (or already does, I'm not up to date on the status). I have some concerns, but just so I understand, how would targets be given to the agent? Could you write this up a little more somewhat formally? 

> Is there an inherent relationship between discovery sources and targets, so that if a discovery source changes those changes are propagated downstream? IE discovery source is no longer valid so all the targets need to be removed from the running systems?

It would work like Prometheus: if a SD config becomes invalid, the last set of discovered targets will continue to exist until either the SD config becomes valid again or is updated. To put another way, the proposal here keeps the same semantics of the SD channel but adds sharding and pushes the targets over the network.

mattdurham:
Thoughts on having some scaling tests/environment/conceptual that need to pass before this becomes out of experimental? 

Something like 1k nodes with 100k configs with performance X, or some comparison.

rfratto:
> Thoughts on having some scaling tests/environment/conceptual that need to pass before this becomes out of experimental?

Definitely a good idea, though maybe out of scope for the RFC, and can be discussed after the experiment is initially released and people start playing around with it.

rfratto:
It feels like we've had consensus on this for a while now; taking out of draft.

rfratto:
I'm assigning this RFC-0003 and enabling auto-merge.

Issue #1121, opened 2021-11-18
Proposal: Remote Configuration


rlankfo:
# Remote Configuration

This is a more detailed proposal to the original idea proposed in #869 

Currently the only way to provide configuration to the agent is to pass a local file. Ideally this feature would provide functionality for fetching remote files via several different sources including the following:

* ~~AWS Secrets Manager~~
* AWS S3
* Azure Blob Storage
* Google Cloud Storage
* ~~HashiCorp Vault~~
* HTTP/HTTPS
* Remote Git repos

Access to these remote sources should be made available through URLs. Using URLs for retrieving remote configuration will simplify passing parameters to the agent on the command line, and make it possible to retrieve other remote files which may be configured for use in the agent.

### Implementation Details
This proposal suggests adding the ability to fetch configurations from remote locations by utilizing [go-fsimpl](https://github.com/hairyhenderson/go-fsimpl). 
go-fsimpl enables access to these sources through [URL schemes](https://github.com/hairyhenderson/go-fsimpl/blob/main/url_schemes.md).

### Goals

* allowing instances of the agent to fetch configuration files from a remote location
* support several common remote sources (such as all of the URL schemes made available by go-fsimpl)
* reload remote configuration on requests to `/-/reload` endpoint

### Non-Goals

* polling remote sources for configuration changes
* merging configurations from different sources

###  Pros

* multiple sources supported resulting in more flexibility for end users
* the heavy-lifting is delegated to go-fsimpl
* easily source control agent configs by using git repository source

### Cons

* added complexity with more surface area to support
* different scheme types require different types of credentials
* each scheme type would need to be tested and well documented
* introduces several new dependencies

## Proposed Usage

The feature would be enabled by passing a valid URL to the `-config.file` command line parameter. See the go-fsimpl documentation for supported URL schemes. Default to a local file if no scheme is passed.

Several sources will require providing credentials; however, this should typically be done as documented by the source itself.  For example, when retrieving a config from S3, credentials would be provided in `~/.aws/credentials`. 

### Examples
```bash
./agent -config.file=s3://<bucket>/<path>/agent-config.yaml
```

<TODO: will provide more examples of how authorization would/could work>

rfratto:
I know I initially opened #869 and lightly proposed this, but I do love this idea and I do have to give all the credit to @mattdurham for planting the seed here. 

[...]

PR #1143, opened 2021-11-30
remote config with http(s) provider

rlankfo:
#### PR Description 

This PR adds a simple remote configuration framework and implements an http(s) provider. More providers to come in following PRs.

https://github.com/grafana/agent/issues/1121  (partially)

#### Notes to the Reviewer

Currently the http remote provider supports basic auth, by passing the password, or a password file. To enable the feature, pass the `-experiment.config-urls.enable` flag when starting the agent.

The following command line flags have been added for auth for the http/https provider:
* `-config.url.basic-auth-user`
* `-config.url.basic-auth-password-file`

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -7,18 +7,18 @@ import (
 	"os"
 	"unicode"
 
+	"github.com/drone/envsubst/v2"
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
-	"github.com/weaveworks/common/server"
-
-	"github.com/drone/envsubst/v2"
 	"github.com/grafana/agent/pkg/integrations"
 	"github.com/grafana/agent/pkg/logs"
 	"github.com/grafana/agent/pkg/metrics"
 	"github.com/grafana/agent/pkg/traces"
 	"github.com/grafana/agent/pkg/util"
 	"github.com/pkg/errors"
+	"github.com/prometheus/common/config"
 	"github.com/prometheus/common/version"
+	"github.com/weaveworks/common/server"
 	"gopkg.in/yaml.v2"
 )
 
@@ -45,6 +45,10 @@ type Config struct {
 
 	// Deprecated fields user has used. Generated during UnmarshalYAML.
 	Deprecations []string `yaml:"-"`
+
+	// Remote config options
+	BasicAuthUser     string `yaml:"-"`
+	BasicAuthPassFile string `yaml:"-"`
 }
 
 // UnmarshalYAML implements yaml.Unmarshaler.
@@ -149,6 +153,8 @@ func (c *Config) RegisterFlags(f *flag.FlagSet) {
 
 	f.StringVar(&c.ReloadAddress, "reload-addr", "127.0.0.1", "address to expose a secondary server for /-/reload on.")
 	f.IntVar(&c.ReloadPort, "reload-port", 0, "port to expose a secondary server for /-/reload on. 0 disables secondary server.")
+	f.StringVar(&c.BasicAuthUser, "basic-auth-user", "", "basic auth username for fetching remote config.")
+	f.StringVar(&c.BasicAuthPassFile, "basic-auth-password-file", "", "path to file containing basic auth password for fetching remote config.")
 }
 
 // LoadFile reads a file and passes the contents to Load
@@ -162,7 +168,28 @@ func LoadFile(filename string, expandEnvVars bool, c *Config) error {
 
 // LoadRemote reads a config from url
 func LoadRemote(url string, expandEnvVars bool, c *Config) error {
-	rc, err := NewRemoteConfig(url, nil)
+	remoteOpts := &RemoteOpts{
+		ExpandEnv: expandEnvVars,
+	}
+	if c.BasicAuthUser != "" || c.BasicAuthPassFile != "" {
+		remoteOpts.HTTPClientConfig = &config.HTTPClientConfig{
+			BasicAuth: &config.BasicAuth{
+				Username:     c.BasicAuthUser,
+				PasswordFile: c.BasicAuthPassFile,
+			},
+		}
+
+	}
+
+	if remoteOpts.HTTPClientConfig != nil {
+		dir, err := os.Getwd()
+		if err != nil {
+			return err
+		}
+		remoteOpts.HTTPClientConfig.SetDirectory(dir)
+	}
+
+	rc, err := NewRemoteConfig(url, remoteOpts)
 	if err != nil {
 		return errors.Wrap(err, "error reading remote config")
 	}
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -47,8 +47,9 @@ type Config struct {
 	Deprecations []string `yaml:"-"`
 
 	// Remote config options
-	BasicAuthUser     string `yaml:"-"`
-	BasicAuthPassFile string `yaml:"-"`
+	ExperimentalConfigURLs bool   `yaml:"-"`
+	BasicAuthUser          string `yaml:"-"`
+	BasicAuthPassFile      string `yaml:"-"`
 }
 
 // UnmarshalYAML implements yaml.Unmarshaler.
@@ -155,6 +156,7 @@ func (c *Config) RegisterFlags(f *flag.FlagSet) {
 	f.IntVar(&c.ReloadPort, "reload-port", 0, "port to expose a secondary server for /-/reload on. 0 disables secondary server.")
 	f.StringVar(&c.BasicAuthUser, "config.url.basic-auth-user", "", "basic auth username for fetching remote config.")
 	f.StringVar(&c.BasicAuthPassFile, "config.url.basic-auth-password-file", "", "path to file containing basic auth password for fetching remote config.")
+	f.BoolVar(&c.ExperimentalConfigURLs, "experiment.config-urls.enable", false, "enable experimental remote config URLs feature")
 }
 
 // LoadFile reads a file and passes the contents to Load
@@ -240,7 +242,7 @@ func getenv(name string) string {
 // to the flagset before parsing them with the values specified by
 // args.
 func Load(fs *flag.FlagSet, args []string) (*Config, error) {
-	return load(fs, args, LoadRemote)
+	return load(fs, args, LoadFile)
 }
 
 // load allows for tests to inject a function for retrieving the config file that
@@ -266,6 +268,10 @@ func load(fs *flag.FlagSet, args []string, loader func(string, bool, *Config) er
 	if printVersion {
 		fmt.Println(version.Print("agent"))
 		os.Exit(0)
+	}
+
+	if cfg.ExperimentalConfigURLs {
+		loader = LoadRemote
 	}
 
 	if file == "" {
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -153,8 +153,8 @@ func (c *Config) RegisterFlags(f *flag.FlagSet) {
 
 	f.StringVar(&c.ReloadAddress, "reload-addr", "127.0.0.1", "address to expose a secondary server for /-/reload on.")
 	f.IntVar(&c.ReloadPort, "reload-port", 0, "port to expose a secondary server for /-/reload on. 0 disables secondary server.")
-	f.StringVar(&c.BasicAuthUser, "basic-auth-user", "", "basic auth username for fetching remote config.")
-	f.StringVar(&c.BasicAuthPassFile, "basic-auth-password-file", "", "path to file containing basic auth password for fetching remote config.")
+	f.StringVar(&c.BasicAuthUser, "config.url.basic-auth-user", "", "basic auth username for fetching remote config.")
+	f.StringVar(&c.BasicAuthPassFile, "config.url.basic-auth-password-file", "", "path to file containing basic auth password for fetching remote config.")
 }
 
 // LoadFile reads a file and passes the contents to Load
@@ -168,9 +168,7 @@ func LoadFile(filename string, expandEnvVars bool, c *Config) error {
 
 // LoadRemote reads a config from url
 func LoadRemote(url string, expandEnvVars bool, c *Config) error {
-	remoteOpts := &RemoteOpts{
-		ExpandEnv: expandEnvVars,
-	}
+	remoteOpts := &RemoteOpts{}
 	if c.BasicAuthUser != "" || c.BasicAuthPassFile != "" {
 		remoteOpts.HTTPClientConfig = &config.HTTPClientConfig{
 			BasicAuth: &config.BasicAuth{
@@ -178,7 +176,6 @@ func LoadRemote(url string, expandEnvVars bool, c *Config) error {
 				PasswordFile: c.BasicAuthPassFile,
 			},
 		}
-
 	}
 
 	if remoteOpts.HTTPClientConfig != nil {
@@ -191,18 +188,17 @@ func LoadRemote(url string, expandEnvVars bool, c *Config) error {
 
 	rc, err := NewRemoteConfig(url, remoteOpts)
 	if err != nil {
-		return errors.Wrap(err, "error reading remote config")
+		return errors.Errorf("error reading remote config: %s", err.Error())
 	}
 	// fall back to file if no scheme is passed
 	if rc == nil {
 		return LoadFile(url, expandEnvVars, c)
 	}
-	config, err := rc.Retrieve()
+	bb, err := rc.Retrieve()
 	if err != nil {
-		return errors.Wrap(err, "error retrieving remote config")
+		return errors.Errorf("error retrieving remote config: %s", err.Error())
 	}
-	*c = *config
-	return nil
+	return LoadBytes(bb, expandEnvVars, c)
 }
 
 // LoadBytes unmarshals a config from a buffer. Defaults are not
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -105,7 +105,7 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 // messages.
 func (c *Config) LogDeprecations(l log.Logger) {
 	for _, d := range c.Deprecations {
-		level.Warn(l).Log("msg", fmt.Sprintf("DEPRECATION NOTICE: %s", d))
+		_ = level.Warn(l).Log("msg", fmt.Sprintf("DEPRECATION NOTICE: %s", d))
 	}
 }
 
@@ -160,6 +160,24 @@ func LoadFile(filename string, expandEnvVars bool, c *Config) error {
 	return LoadBytes(buf, expandEnvVars, c)
 }
 
+// LoadRemote reads a config from url
+func LoadRemote(url string, expandEnvVars bool, c *Config) error {
+	rc, err := NewRemoteConfig(url, nil)
+	if err != nil {
+		return errors.Wrap(err, "error reading remote config")
+	}
+	// fall back to file if no scheme is passed
+	if rc == nil {
+		return LoadFile(url, expandEnvVars, c)
+	}
+	config, err := rc.Retrieve()
+	if err != nil {
+		return errors.Wrap(err, "error retrieving remote config")
+	}
+	*c = *config
+	return nil
+}
+
 // LoadBytes unmarshals a config from a buffer. Defaults are not
 // applied to the file and must be done manually if LoadBytes
 // is called directly.
@@ -199,7 +217,7 @@ func getenv(name string) string {
 // to the flagset before parsing them with the values specified by
 // args.
 func Load(fs *flag.FlagSet, args []string) (*Config, error) {
-	return load(fs, args, LoadFile)
+	return load(fs, args, LoadRemote)
 }
 
 // load allows for tests to inject a function for retrieving the config file that
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -105,7 +105,7 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 // messages.
 func (c *Config) LogDeprecations(l log.Logger) {
 	for _, d := range c.Deprecations {
-		_ = level.Warn(l).Log("msg", fmt.Sprintf("DEPRECATION NOTICE: %s", d))
+		level.Warn(l).Log("msg", fmt.Sprintf("DEPRECATION NOTICE: %s", d))
 	}
 }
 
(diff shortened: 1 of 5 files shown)


pkg/config/config.go
```diff
@@ -105,7 +105,7 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 // messages.
 func (c *Config) LogDeprecations(l log.Logger) {
 	for _, d := range c.Deprecations {
-		level.Warn(l).Log("msg", fmt.Sprintf("DEPRECATION NOTICE: %s", d))
+		_ = level.Warn(l).Log("msg", fmt.Sprintf("DEPRECATION NOTICE: %s", d))
```
rfratto, line 108: You shouldn't have to add the `_` here FWIW:
https://github.com/grafana/agent/blob/e6745e60df86b1848fa8341aedc17b61cc2f0950/.golangci.yml#L70-L74

rlankfo: My linter kept complaining about it in Goland so I just added this.

pkg/config/remote_config.go
```diff
@@ -0,0 +1,142 @@
...
+
+		request   *http.Request
+		response  *http.Response
+		basicAuth *config.BasicAuth
+
+		result = &Config{}
+		client = &http.Client{}
+	)
+
+	if p.httpClientConfig != nil {
```
rfratto, line 96: nit: Use `config.NewClientFromConfig` here and call `SetDirectory` from the caller instead of in the handler

rfratto (commented):
I think this is a good start, but I think it's worth some discussion about how quickly we're going to want to expand the usage here. 

If we suspect that we might also support remote_urls for #868, then it might be worthwhile having RemoteConfig implement fs.FS sooner rather than later. Otherwise, if we don't think we're going to expand the remote URL concept any further than we probably don't need a broader interface yet. 

(fs.FS would be useful for #868 in particular because conf.d would walk over directories)

pkg/config/config.go
rfratto: Can you disable the linting rule in Goland? I'd really rather not have to add `_ =` to every time we log :)

pkg/config/remote_config.go
rlankfo: @rfratto any reason we want to use the prometheus `http.Client` vs using Go's `http.DefaultClient`?  I didn't look into what all the differences were but assumed the prometheus client was likely optimized for pushing metrics

rfratto: I think if we're going to use Prometheus' HTTP client config, then we should reuse their existing code that generates a client from it :) It's not very Prometheus-specific, anything in prometheus/common is pretty broadly usable.

rlankfo: > If we suspect that we might also support remote_urls for #868, then it might be worthwhile having RemoteConfig implement fs.FS sooner rather than later. Otherwise, if we don't think we're going to expand the remote URL concept any further than we probably don't need a broader interface yet.

This makes sense to me. I intentionally kept this simple / extendable so it shouldn't be a problem to support conf.d style configs if we decide to in the future. Instead of having future requirements creep into the scope of this feature, I think we should implement local conf.d configuration support first, and then add it for remote as a separate piece of work.

pkg/config/remote_config.go
```diff
@@ -0,0 +1,140 @@
...
+	if basicAuth != nil {
+		if basicAuth.PasswordFile != "" {
+			bs, err := ioutil.ReadFile(basicAuth.PasswordFile)
+			if err != nil {
+				return nil, fmt.Errorf("unable to read basic auth password file %s: %s", basicAuth.PasswordFile, err)
+			}
+			basicAuth.Password = config.Secret(strings.TrimSpace(string(bs)))
+		}
+		request.SetBasicAuth(basicAuth.Username, string(basicAuth.Password))
+	}
```
rfratto, line 119: Now that you're using `config.NewClientFromConfig` you can take this out; it'll inject basic auth for you :)

rlankfo:
> I think this is a good start, but I think it's worth some discussion about how quickly we're going to want to expand the usage here.
> 
> If we suspect that we might also support remote_urls for #868, then it might be worthwhile having RemoteConfig implement fs.FS sooner rather than later. Otherwise, if we don't think we're going to expand the remote URL concept any further than we probably don't need a broader interface yet.
> 
> (fs.FS would be useful for #868 in particular because conf.d would walk over directories)

This makes sense to me. I intentionally kept this simple / extendable so it shouldn't be a problem to support conf.d style configs if we decide to in the future. Instead of having future requirements creep into the scope of this feature, I think we should implement local conf.d configuration support first, and then add it for remote as a separate piece of work.

rfratto:
I'm not opposed to putting it off, but I do want to at least make it clear that implementing fs.FS here would be a pretty significant rewrite.

rlankfo:
> I'm not opposed to putting it off, but I do want to say that I think implementing fs.FS at any point would be a pretty significant rewrite.

I would think it should be possible to retrieve `Config`s using this framework by implementing a new provider that uses go-fsimpl under the covers w/o having to do much rewrite. Does that make sense?

mattdurham (commented):
Code looks reasonable. Small and lightweight. Would love to see some doc, ie examples and the configuration options in our config block. With some comments on this being beta?

pkg/config/config.go
```diff
@@ -149,6 +153,8 @@ func (c *Config) RegisterFlags(f *flag.FlagSet) {
 
 	f.StringVar(&c.ReloadAddress, "reload-addr", "127.0.0.1", "address to expose a secondary server for /-/reload on.")
 	f.IntVar(&c.ReloadPort, "reload-port", 0, "port to expose a secondary server for /-/reload on. 0 disables secondary server.")
+	f.StringVar(&c.BasicAuthUser, "basic-auth-user", "", "basic auth username for fetching remote config.")
+	f.StringVar(&c.BasicAuthPassFile, "basic-auth-password-file", "", "path to file containing basic auth password for fetching remote config.")
```
rfratto, line 157: I think the flag names should be namespaced to make it clearer what they're used for (i.e., `config.url.basic-auth-user`)

pkg/config/remote_config.go
```diff
@@ -0,0 +1,127 @@
...
+	switch u.Scheme {
+	case "":
+		// if no scheme, assume local file path, return nil and let caller handle.
+		return nil, nil
+	case HTTP, HTTPS:
+		return &RemoteConfig{
+			RemoteProvider: newHTTPProvider(opts),
+		}, nil
+	}
+	return nil, fmt.Errorf("remote config scheme not supported: %s", u.Scheme)
```
rfratto, line 60: nit: move to default case

pkg/config/remote_config.go
```diff
@@ -0,0 +1,127 @@
...
+		result = &Config{}
+	)
+
+	if p.httpClientConfig != nil {
+		err = p.httpClientConfig.Validate()
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		p.httpClientConfig = &config.HTTPClientConfig{}
```
rfratto, line 99: nit: why use a pointer at all here if it's always going to be non-nil?

pkg/config/remote_config.go
```diff
@@ -0,0 +1,127 @@
...
+
+// Remote Config Providers
+// httpP - http/https provider
+type httpP struct {
+	myURL            *url.URL
+	expandEnv        bool
+	httpClientConfig *config.HTTPClientConfig
+}
+
+func newHTTPProvider(opts *RemoteOpts) *httpP {
```
rfratto, line 71: nit: validate http client config here instead of at retrieve time

pkg/config/remote_config.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	request, err = http.NewRequest(http.MethodGet, p.myURL.String(), nil)
+	if err != nil {
+		return nil, err
+	}
+	response, err = client.Do(request)
+	if err != nil {
+		return nil, err
+	}
```
rfratto, line 113: nit: `client.Get` instead of NewRequest + client.Do

pkg/config/remote_config.go
```diff
@@ -0,0 +1,127 @@
...
+
+	if response.StatusCode < http.StatusOK || response.StatusCode >= http.StatusMultipleChoices {
+		return nil, fmt.Errorf("error fetching config: status code: %d", response.StatusCode)
+	}
```
rfratto, line 117: nit: `response.StatusCode/100 != 2`

pkg/config/remote_config.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	request, err = http.NewRequest(http.MethodGet, p.myURL.String(), nil)
+	if err != nil {
+		return nil, err
+	}
+	response, err = client.Do(request)
+	if err != nil {
```
rfratto, line 111: You need to `defer response.Body.Close()` if there wasn't an error here

pkg/config/remote_config.go
```diff
@@ -0,0 +1,127 @@
...
+		return &RemoteConfig{
+			RemoteProvider: newHTTPProvider(opts),
+		}, nil
+	}
+	return nil, fmt.Errorf("remote config scheme not supported: %s", u.Scheme)
+}
+
+// Remote Config Providers
+// httpP - http/https provider
+type httpP struct {
```
rfratto, line 65: nit: `httpProivder`

pkg/config/remote_config.go
```diff
@@ -0,0 +1,127 @@
...
+	var (
+		bb  []byte
+		err error
+
+		request  *http.Request
+		response *http.Response
+		client   *http.Client
+
+		result = &Config{}
+	)
```
rfratto, line 91: very minor nit: I don't think you need to declare every variable you might use at the top like this, but I also don't really care that much

pkg/config/config.go
```diff
@@ -160,6 +166,45 @@ func LoadFile(filename string, expandEnvVars bool, c *Config) error {
 	return LoadBytes(buf, expandEnvVars, c)
 }
 
+// LoadRemote reads a config from url
+func LoadRemote(url string, expandEnvVars bool, c *Config) error {
```
rfratto, line 170: nit: I'd rather `rc.Retrieve` return `([]byte, error)` so we don't have to propagate expandEnvVars (and so it could exist in its own package)

pkg/config/config.go
```diff
@@ -160,6 +166,45 @@ func LoadFile(filename string, expandEnvVars bool, c *Config) error {
...
+		dir, err := os.Getwd()
+		if err != nil {
+			return err
+		}
+		remoteOpts.HTTPClientConfig.SetDirectory(dir)
+	}
+
+	rc, err := NewRemoteConfig(url, remoteOpts)
+	if err != nil {
+		return errors.Wrap(err, "error reading remote config")
```
rfratto, line 194: nit: favor `fmt.Errorf` over `errors.Wrap`

pkg/config/remote_config.go
```diff
@@ -0,0 +1,127 @@
...
+
+// RemoteProvider ...
+type RemoteProvider interface {
+	Retrieve() (*Config, error)
+}
+
+// RemoteConfig ...
+type RemoteConfig struct {
+	RemoteProvider
+}
```
rfratto, line 35: I'm not sure I understand why we have a struct that only embeds the interface? IMO, just return the interface from `NewRemoteConfig`

rfratto (commented):
looking good, just nits. 

would also like to see this documented and added to the changelog before merge though.

maybe also command line flag to enable "experimental remote url support?"

pkg/config/remote_config.go
rlankfo: Typically just got into a habit of doing this since it can be helpful for GC, especially when re-assigning to declared variables in a loop that gets executed lots of times but in this case probably isn't needed.

pkg/config/remote_config.go
rlankfo: I just did this for flexibility so if necessary additional values could be carried with this struct instead of just on the underlying `RemoteProvider` implementations.

rfratto: YAGNI :) I'd say keep it as an interface for now.

pkg/config/config.go
rlankfo: This seems reasonable; I almost went this route initially.

pkg/config/remote_config.go
rlankfo: the `RemoteOpts` struct will be shared between other provider implementations (git://, s3://, etc). Most won't need an http client config.

pkg/config/remote_config.go
rlankfo: ok -- switching to an interface is fine and it's easy to switch back if need be. it doesn't change the semantics for the caller. If we do end up with additional providers, conf.d style support, etc. I can see us wanting to carry other values here potentially.

rlankfo:
@rfratto @mattdurham addressed some of the minor nits if you guys want to take another pass at some point. I'm still leaving this as a draft for now as I wanted to implement a couple more providers before trying to get this merged in.

rfratto:
@rlankfo IMO http/s is a good starting point for a first PR, other providers can be done in future PRs

rlankfo:
> @rlankfo IMO http/s is a good starting point for a first PR, other providers can be done in future PRs

Ok sounds good, I can switch gears to finalizing this for merge. Do you think we *need* the experimental URL flag? Also, do think we should support token authorization as well as basic before merge?

rfratto:
I'm not sure if we _need_ an experimental flag. Experimental flags isn't something we've done historically, but I'm starting to think about when they might be useful to signal to the user that something is subject to change given #1140. 

> Also, do think we should support token authorization as well as basic before merge?

It's not strictly necessary, I don't think. We still don't know long-term if plain URLs is strictly the format we want to use over something like `provider=https url=http://example.com/agent.yaml basic_auth_username=foo basic_auth_passwordfile=pass.txt`

rfratto:
> It's not strictly necessary, I don't think. We still don't know long-term if plain URLs is strictly the format we want to use 

Haha, now that I said this, maybe an experimental flag is the way to go. Let's chat offline about this.

rfratto:
It looks like it's going to be difficult to find a nice, fast way to do experimental flags that error when configuring a flag for an experiment that is not enabled.

For now, I think we should have an `--experimental.<experiment name>.enable` flag that enables the code. Other flags can exist without the experimental prefix and can be parsed without any errors regardless of the other flag being set. We can figure out how to make it error later.

pkg/config/config.go
```diff
@@ -268,6 +270,10 @@ func load(fs *flag.FlagSet, args []string, loader func(string, bool, *Config) er
 		os.Exit(0)
 	}
 
+	if cfg.ExperimentalConfigURLs {
+		loader = LoadRemote
+	}
```
rfratto, line 275: I think this would be unexpected from a testing point of view, see suggestion for another way of handling it

pkg/config/config.go
```diff
@@ -240,7 +242,7 @@ func getenv(name string) string {
 // to the flagset before parsing them with the values specified by
 // args.
 func Load(fs *flag.FlagSet, args []string) (*Config, error) {
-	return load(fs, args, LoadRemote)
+	return load(fs, args, LoadFile)
```
rfratto: I don't see the suggestion button for some reason about what about doing this:

```go
return load(fs, args, func(url string, expand bool, c *Config) error {
  if c.cfg.ExperimentalConfigURLs {
    return LoadRemote(url, expand, c)
  }
  return LoadFile(url, expand, c)
}
```

rlankfo: good call, just made this update :+1:

pkg/config/config.go
```diff
@@ -149,6 +154,9 @@ func (c *Config) RegisterFlags(f *flag.FlagSet) {
 
 	f.StringVar(&c.ReloadAddress, "reload-addr", "127.0.0.1", "address to expose a secondary server for /-/reload on.")
 	f.IntVar(&c.ReloadPort, "reload-port", 0, "port to expose a secondary server for /-/reload on. 0 disables secondary server.")
+	f.StringVar(&c.BasicAuthUser, "config.url.basic-auth-user", "", "basic auth username for fetching remote config.")
+	f.StringVar(&c.BasicAuthPassFile, "config.url.basic-auth-password-file", "", "path to file containing basic auth password for fetching remote config.")
```
rfratto, line 158: can the help text say that they require the experiment to be enabled?

pkg/config/config.go
```diff
@@ -160,6 +168,41 @@ func LoadFile(filename string, expandEnvVars bool, c *Config) error {
...
+func LoadRemote(url string, expandEnvVars bool, c *Config) error {
+	remoteOpts := &RemoteOpts{}
+	if c.BasicAuthUser != "" || c.BasicAuthPassFile != "" {
+		remoteOpts.HTTPClientConfig = &config.HTTPClientConfig{
+			BasicAuth: &config.BasicAuth{
+				Username:     c.BasicAuthUser,
+				PasswordFile: c.BasicAuthPassFile,
+			},
+		}
+	}
```
rfratto, line 181: does basic auth work if you just have a password but no username? 🤔

pkg/config/config.go
```diff
@@ -160,6 +168,41 @@ func LoadFile(filename string, expandEnvVars bool, c *Config) error {
...
+		dir, err := os.Getwd()
+		if err != nil {
+			return err
+		}
+		remoteOpts.HTTPClientConfig.SetDirectory(dir)
+	}
+
+	rc, err := NewRemoteConfig(url, remoteOpts)
+	if err != nil {
+		return errors.Errorf("error reading remote config: %s", err.Error())
```
rfratto, line 193: please wrap the errors instead of using errors.Errorf so they can be unwrapped: `fmt.Errorf("error reading remote config: %w", err)`  (ditto for other places you do this)

pkg/config/remote_config.go
```diff
@@ -0,0 +1,100 @@
+package config
```
rfratto, line 1: IMO we should unexport everything in this file since we don't intend for it to be used by other packages (yet)

pkg/config/remote_config.go
```diff
@@ -0,0 +1,100 @@
...
+	// TODO: add s3, gcs, blob, and git providers backed by go-fsimpl
+)
+
+// RemoteOpts struct contains agent remote config options
+type RemoteOpts struct {
+	URL              *url.URL
+	HTTPClientConfig *config.HTTPClientConfig
+}
+
+// RemoteProvider ...
```
rfratto, line 24: nit: fix godoc comment

pkg/config/remote_config.go
```diff
@@ -0,0 +1,100 @@
...
+type RemoteProvider interface {
+	Retrieve() ([]byte, error)
+}
+
+// NewRemoteConfig ...
```
rfratto, line 29: nit: fix godoc comment

pkg/config/remote_config.go
```diff
@@ -0,0 +1,100 @@
...
+
+// RemoteOpts struct contains agent remote config options
+type RemoteOpts struct {
+	URL              *url.URL
+	HTTPClientConfig *config.HTTPClientConfig
+}
+
+// RemoteProvider ...
+type RemoteProvider interface {
+	Retrieve() ([]byte, error)
```
rfratto, line 26: in the future, we should allow this to have a timeout

pkg/config/remote_config.go
```diff
@@ -0,0 +1,100 @@
...
+// supported remote config provider schemes
+const (
+	HTTP  = "http"
+	HTTPS = "https"
+	// TODO: add s3, gcs, blob, and git providers backed by go-fsimpl
+)
+
+// RemoteOpts struct contains agent remote config options
+type RemoteOpts struct {
+	URL              *url.URL
```
rfratto, line 20: minor DX nit: it's a little unexpected that the user is not supposed to supply URL, but can supply the other field. I'd say make it unexported and keep HTTPClientConfig exported.

pkg/config/remote_config.go
```diff
@@ -0,0 +1,100 @@
...
+	default:
+		return nil, fmt.Errorf("remote config scheme not supported: %s", u.Scheme)
+	}
+}
+
+// Remote Config Providers
+// httpProvider - http/https provider
+type httpProvider struct {
+	myURL            url.URL
+	httpClientConfig config.HTTPClientConfig
```
rfratto, line 60: nit: I'd just store the opts here
(second nit: keep myURL as a pointer, since that seems to be common)

pkg/config/remote_config.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	return &httpProvider{
+		myURL:            *opts.URL,
+		httpClientConfig: httpClientConfig,
+	}, nil
+}
+
+// Retrieve implements RemoteProvider and fetches the config
+// TODO: token auth, oauth2, etc.
+func (p httpProvider) Retrieve() ([]byte, error) {
```
rfratto, line 80: can we add more wrapped errors here? It would be useful to know if this fails at which point it failed

pkg/config/remote_config.go
```diff
@@ -0,0 +1,100 @@
...
+	client, err := config.NewClientFromConfig(p.httpClientConfig, "remote-config")
+	if err != nil {
+		return nil, err
+	}
```
rfratto, line 84: nit: do this when the provider is created rather than at retrieve time, since the generated client will never change.

pkg/config/remote_config_test.go
```diff
@@ -0,0 +1,120 @@
+package config
```
rfratto, line 1: please use github.com/stretchr/testify/require here for assertions instead of directly invoking t.Error/t.FailNow :)

pkg/config/config.go
rlankfo: no I don't think so. this condition should be `&&` instead of `||`

pkg/config/remote_config.go
rlankfo: thinking about this, I probably only need to store the `http.Client` here instead of the opts.

pkg/config/remote_config.go
rlankfo: Agreed. For now, if we want to set a timeout for HTTP/S, we can pass the `WithIdleConnTimeout` `HTTPClientOption` to `NewClientFromConfig`.  We can revisit managing our own timeouts once we have more providers to deal with.

pkg/config/remote_config.go
rlankfo: yeah, that irked me a bit but seemed like the lesser evil compared to having to pass url separately down to the providers. I think un-exporting it is a good solution for now.

rfratto (approved):
great work, thank you!

PR #1154, opened 2021-12-08
docs: improve windows_exporter integration documentation

rfratto:
Make reference to how any collector can be enabled if it's non
experimental, but some collectors accept extra configuration.


rlankfo (approved):
LGTM

PR #1158, opened 2021-12-09
add DarwinConfig for macos_exporter integration

rlankfo:
#### PR Description 

This adds a new node_exporter integration config named macos_exporter. I'm adding this so we can have a separate config for the new MacOS integration we're creating as part of the hackathon. This is necessary in case a user enables the linux node integration AND the macos integration. These should be at least named differently to avoid unmarshalling errors in integrations-api: 
```level=error ts=2021-12-09T17:50:05.3051756Z revision=622c63b0-WIP caller=http.go:560 traceID=1daa1865c456ce16 stack_id=1 org_id=1 msg="could not unmarshal agent config" err="yaml: unmarshal errors:\n  line 33: mapping key \"node_exporter\" already defined at line 29" ```

#### Notes to the Reviewer

I thought about putting this in a file called `node_exporter_darwin.go` which would only enable it if the agent is built on darwin hosts, however, the trade off here is that if the end user has `macos_exporter` integration enabled in their config, it will blow up on linux, windows, etc.

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated


rlankfo:
@rfratto @mattdurham @gaantunes Do you guys mind reviewing this and letting me know your thoughts here? We want to add a config in our new integration that doesn't conflict with existing node_exporter integration.

rlankfo:
> I thought about putting this in a file called `node_exporter_darwin.go` which would only enable it if the agent is built on darwin hosts, however, the trade off here is that if the end user has `macos_exporter` integration enabled in their config, it will blow up on linux, windows, etc.

maybe this is ok and actually what we would want?

mattdurham:
We normally add a dummy file for the operating systems that don’t support it so the config doesn’t explode if ran on those

rlankfo:
> We normally add a dummy file for the operating systems that don’t support it so the config doesn’t explode if ran on those

Do you have an example of this?

mattdurham:
https://github.com/grafana/agent/blob/main/pkg/integrations/windows_exporter/windows_exporter.go

rlankfo:
thanks @mattdurham this makes sense to me, I could add register a dummy darwin config for this so it doesn't blow up but will do nothing if not run on darwin. 👍

rfratto:
I don't think we should do this, unless the config schemas are going to be different. Plus, if a user enables both node_exporter and this new exporter, this will actually not break: we set global config flags for node_exporter to work and two instances of node_exporter (i.e., the actual node_exporter, not the integration name) can't coexist today. 

If a macOS integration were going to exist, I think it reveals more necessity around agent-specific configs from Grafana Cloud instead of generating one giant config to give every agent.

rlankfo:
> I don't think we should do this, unless the config schemas are going to be different. Plus, if a user enables both node_exporter and this new exporter, this will actually not break: we set global config flags for node_exporter to work and two instances of node_exporter (i.e., the actual node_exporter, not the integration name) can't coexist today.

hey @rfratto sorry to clarify, what are your main concerns here? How could we address them in the short term in the agent w/o having to make significant changes to the way agent configs are generated in Grafana Cloud?

After running this for a few days, the configs will need to be different. Some of the collectors we enable for the regular node_exporter integration don't work on Darwin, and there are some Darwin specific ones we'll need to enable.]

mattdurham:
Looking at this again, if we ensured via build directives that both couldn't be enabled at the same time, then node_exporter would be a noop on osx and macos_exporter would be a noop on linux.

mattdurham:
The major issue is that it would be a logical breaking change, ie if you had node_exporter before then it would still run the agent without erroring but nothing would happen.

PR #1162, opened 2021-12-13
Fix POSTGRES_EXPORTER_DATA_SOURCE_NAME usage for postgres_exporter

f11r:
A recent change broke the usage of POSTGRES_EXPORTER_DATA_SOURCE_NAME for the postgres_exporter.
As the incorrect variable was checked in the if clause, it always raises an error.

#### Notes to the Reviewer

As I'm not familiar with the codebase (and more generally go) it would take me a lot of time to add a regression test for this. It would be great if someone else could do this.

#### PR Checklist

- [x] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Main (unreleased)
 
+- [BUGFIX] Fix usage of POSTGRES_EXPORTER_DATA_SOURCE_NAME when using postgres_exporter integration (@f11r)
+
 - [FEATURE] (beta) Enable experimental config urls for fetching remote configs. Currently,
    only HTTP/S is supported. Use `-experiment.config-urls.enable` flag to turn this on. (@rlankfo)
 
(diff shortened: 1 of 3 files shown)


f11r:
Really great to have had this merged so quickly :+1: 

Will you also create a new release soon? Currently everyone is forced to choose either upgrading for the CVE or using the environment variable for the postgres exporter.

rfratto:
Given the holidays, we're nervous about doing a release with many people on PTO (I'll be off myself after this week for the rest of the year). 

It's unfortunate, but for now I'd suggest a few workarounds: 

1. Avoid using the environment variable entirely 
2. If possible, you can use `--config.expand-env` with the YAML:

```yaml
integrations:
  postgres_exporter:
    enabled: true
    data_source_names:
      - ${MY_ENV_VAR}
```

3. For those who are comfortable, run from the main branch (If using Docker, images are pushed for every main commit, `docker pull grafana/agent:main-d4f9cb4`)

f11r:
Alright. Thanks for the quick reply.

Issue #1135, opened 2021-11-25
Explore supporting additional tracing exporters


mapno:
Currently, the Agent is only capable of exporting traces via OTLP (gRPC and HTTP). The main reasoning is that the Agent is built with the intention of sending telemetry data to the opinionated Grafana observability stack. While Tempo can ingest jaeger and other formats, with OTLP this goal is already achieved.

Exporting has a custom config block in the Agent (`remote_write`), which is translated to OTLP's exporter config. This translation is ad-hoc for this exporter. Adding new exporter would require changing or adding fields to the config, to suit the new different options. This would increase complexity in the configuration, in a way that is not beneficial to working with Tempo.

Outside OTLP, 4 of the 5 available exporters that Tempo could ingest from are built on top of gRPC or HTTP ([jaeger](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/jaegerexporter), [jaeger thrift](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/jaegerthrifthttpexporter), [opencensus](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/opencensusexporter) and [zipkin](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/zipkinexporter)). It's possible that `remote_write` is compatible with these exporters, but this requires further inspection to verify.

nicoche:
Hey!

This is something that we would really like to use. I've hacked around and got something running. The config looks like this:

```yaml
traces:
  configs:
    - name: default
      receivers:
        zipkin:
      remote_write:
        - endpoint: localhost:14250
          format: jaeger
          insecure: true
          retry_on_failure:
            enabled: false
```

With `docker run --rm --name jaeger -p 14250:14250 -p 9412:9411 -p 16686:16686 jaegertracing/all-in-one:latest`, spans are correctly forwarded:

```shell
curl --verbose -X POST 127.0.0.1:9411/v1/traces -H 'Content-Type: application/json' -d '[{ "id": "f5402e865c61400e", "traceId": "0123456789abcdef", "timestamp": 1639437423899795, "duration": 1000030, "name": "span from bash!", "tags": { "http.method": "GET", "http.path": "/api" }, "localEndpoint": { "serviceName": "shell script" } }]'
```

[...]

PR #1165, opened 2021-12-14
traces: add jaeger exporter

nicoche:
#### PR Description 

WIP

https://github.com/grafana/agent/issues/1135

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/configuration/traces-config.md
+++ b/docs/configuration/traces-config.md
@@ -59,6 +59,11 @@ remote_write:
     # Only "grpc" is supported in Grafana Cloud.
     [ protocol: <string> | default = "grpc" | supported = "grpc", "http" ]
 
+    # Controls what format to use when exporting traces, in combination with protocol.
+    # protocol/format supported combinations are grpc/otlp, http/otlp and grpc/jaeger
+    # Only grpc/otlp is supported in Grafana Cloud.
+    [ format: <string> | default = "otlp" | supported = "otlp", "jaeger" ]
+
     # Controls whether or not TLS is required.  See https://godoc.org/google.golang.org/grpc#WithInsecure
     [ insecure: <boolean> | default = false ]
 
--- a/pkg/traces/config.go
+++ b/pkg/traces/config.go
@@ -16,6 +16,7 @@ import (
 	"github.com/grafana/agent/pkg/traces/promsdprocessor"
 	"github.com/grafana/agent/pkg/traces/remotewriteexporter"
 	"github.com/grafana/agent/pkg/traces/servicegraphprocessor"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/exporter/jaegerexporter"
 	"github.com/open-telemetry/opentelemetry-collector-contrib/exporter/loadbalancingexporter"
 	"github.com/open-telemetry/opentelemetry-collector-contrib/exporter/prometheusexporter"
 	"github.com/open-telemetry/opentelemetry-collector-contrib/processor/attributesprocessor"
@@ -147,10 +148,16 @@ const (
 	protocolHTTP    = "http"
 )
 
-// DefaultRemoteWriteConfig holds the default setting	s for a PushConfig.
+const (
+	formatOtlp   = "otlp"
+	formatJaeger = "jaeger"
+)
+
+// DefaultRemoteWriteConfig holds the default settings for a PushConfig.
 var DefaultRemoteWriteConfig = RemoteWriteConfig{
 	Compression: compressionGzip,
 	Protocol:    protocolGRPC,
+	Format:      formatOtlp,
 }
 
 // RemoteWriteConfig controls the configuration of an exporter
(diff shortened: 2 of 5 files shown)


pkg/traces/config.go
```diff
@@ -235,7 +248,13 @@ func exporter(rwCfg RemoteWriteConfig) (map[string]interface{}, error) {
 		return nil, errors.New("must have a configured a backend endpoint")
 	}
 
-	headers := map[string]string{}
+	// TODO: investigate
+	// For some reason, jaeger exporter factory makes this field nil when otlp's makes it a 0len map
```
mapno, line 252: So, it turns out that the `otlpexporter` sets the headers to an empty map in the [default config](https://github.com/open-telemetry/opentelemetry-collector/blob/v0.41.0/exporter/otlpexporter/factory.go#L49). This is what's causing the diff between otlp and jaeger. Can we document this briefly?

Also, I think it'd be more correct to change the code to
```go
if rwCfg.Format == formatOtel {
```

pkg/traces/config.go
```diff
@@ -147,10 +148,16 @@ const (
 	protocolHTTP    = "http"
 )
 
-// DefaultRemoteWriteConfig holds the default setting	s for a PushConfig.
+const (
+	formatOtel   = "otlp"
```
mapno, line 152: Nit: can we change it to `formatOTLP`?

```suggestion
	formatOTLP   = "otlp"
```

mapno (commented):
Hey, thanks for the contribution!

I like the direction of the PR. I think `format` is a good choice to configure multiple exporters and addresses the main problem described in https://github.com/grafana/agent/issues/1135.

Left a couple of minor comments.

The next step for this PR would be to document this new parameter in the [config reference](https://github.com/grafana/agent/blob/main/docs/configuration/traces-config.md). Also, please add this change to the changelog.

nicoche:
Hey!
Thanks for taking the time to review this. I'll clean it up.

But before I'd like your opinion on something. I believe that it might be beneficial to deprecate the protocol field.
Exporter formats, as defined by otel, imply a protocol already:
- format [jaegerthrift](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/jaegerthrifthttpexporter) is http
- format [jaeger](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/jaegerexporter) is grpc
- format [otlphttp](https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter/otlphttpexporter) is http
- format [otlp](https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter/otlpexporter) is grpc

I guess it's mostly ok for jaeger and otlp as they both have an an http and a grpc format and we could understand that `protocol:http format:jaeger` refer to jaegerthrift, but regarding formats like [zipkin](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/zipkinexporter), where only http is valid, we could run into this kind of invalid conf:

```yaml
remote_write:
  - endpoint: xxx
    protocol: grpc
    format: zipkin
```

Deprecating `protocol` key in favor of `format` would be clearer imo. Targer confs would be:

```yaml
remote_write:
  - endpoint: xxx
    format: zipkin # or anything from that list actually https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter
```

Thoughts?

mapno:
Hi! Thanks for posting back. These are good points. 

> But before I'd like your opinion on something. I believe that it might be beneficial to deprecate the protocol field.
Exporter formats, as defined by otel, imply a protocol already

I'm not fully convinced in this part. It's my understanding that exporters for different protocols being in separate pkg's is more due to simpler development than being considered each its own format.

Take the OTLP HTTP exporter's definition for instance.

> Exports traces and/or metrics via HTTP using OTLP format.

I think we can run into invalid configs either way. It should be the config code's responsibility to validate them and return meaningful errors.

It's unclear yet how many exporters the Agent will support other than OTLP (gRPC+HTTP) and Jaeger (HTTP), so we will need to list those available in the docs. Similar to listing the different `format` and `protocol` combinations. I don't see a huge gain in that front.

Personally, I'd like to avoid breaking changes for this addition if possible. A new `format` field can be make fully backwards compatible, so I'm reticent to deprecate `protocol`.

nicoche:
Looks fair to me. I've pushed the changes, please let me know if it's missing anything. Once this gets merged, I believe that we could add jaegerthrift / zipkin support pretty easily in subsequent PRs

docs/configuration/traces-config.md
```diff
@@ -59,6 +59,11 @@ remote_write:
     # Only "grpc" is supported in Grafana Cloud.
     [ protocol: <string> | default = "grpc" | supported = "grpc", "http" ]
 
+    # Controls what format to use when exporting traces, in combination with protocol.
+    # protocol/format supported combinations are grpc/otlp, http/otlp and grpc/jaeger
+    # Only grpc/otlp is supported in Grafana Cloud.
```
nicoche, line 64: > Only grpc/otlp is supported in Grafana Cloud.

Not sure about that one tbh

mapno: This is correct :)

mapno (approved):
Sorry for the delay, things are moving slow during these dates.

Thanks for the contribution!

nicoche:
No worries, thanks for taking time to review it. Do you have an idea on when the next release would land?

rfratto:
> No worries, thanks for taking time to review it. Do you have an idea on when the next release would land?

We don't have a standard release schedule (yet) but we were avoiding doing a release during the holiday season. We'll probably have a release within the next few weeks, though right now I don't have a specific date for you.

PR #1166, opened 2021-12-14
Tracing pipeline dashboard and alerts

mapno:
#### PR Description 

Adds new `Agent Tracing Pipeline` jsonnet dashboard and prometheus alerts to production mixin.

Other changes:
- Updates jsonnet libraries
- Linted jsonnet

image

#### PR Checklist

- [x] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated


example/k3d/smoke/main.jsonnet
```diff
@@ -34,11 +34,11 @@ local smoke = {
...
-        metric_count: 1000,
-        series_interval: 300,
-        metric_interval: 600,
-      }),
+    avalanche: avalanche.new(replicas=3, namespace='smoke', config={
+      metric_count: 1000,
+      series_interval: 300,
+      metric_interval: 600,
+    }),
   }),
```
rfratto, line 42: Sorry, you're getting a nit since this got auto formatted :) 

This has actually been broken for a while and a PR hasn't fixed it yet. This is accidentally nesting a second call to avalanche.new when it shouldn't be, instead just giving the config settings. Can you fix that?

production/grafana-agent-mixin/alerts.libsonnet
```diff
@@ -0,0 +1,70 @@
+local config = import 'config.libsonnet';
+local _config = config._config;
+
+{
+  prometheusAlerts+:: {
+    groups+: [
+      {
+        name: 'agent-tracing',
```
rfratto, line 8: nit: grafana-agent-tracing? We do have a few agents here.

production/grafana-agent-mixin/dashboards.libsonnet
```diff
@@ -363,5 +363,230 @@ local template = grafana.template;
         row.new('Misc. Rates')
         .addPanel(enqueueRetries)
       ),
+
+    'agent-tracing-pipeline.json':
```
rfratto, line 367: nit: can you make sure this ends up in a folder? I'm not sure if it will in its current state.

rfratto (approved):
LGTM from jsonnet perspective, but I didn't validate the dashboards or alerts since you're the expert there

annanay25 (commented):
Great work on adding the agent tracing mixin!

Question: Wondering if some of the info on the panels is redundant. Do we need individual panels for Received, Accepted and Rejected spans? Similarly for exported spans.

example/k3d/smoke/main.jsonnet
mapno: Done! Fixed formatting in a couple other places too 35cdcaecad7e0aceaf3bc399af4da8fd4818ebaa

mapno:
> Question: Wondering if some of the info on the panels is redundant. Do we need individual panels for Received, Accepted and Rejected spans? Similarly for exported spans.

Sure, it's kind of the same information but displayed in different ways. The panels in the first row give per-pod details, while the other one give a more general view. Does that make sense?

annanay25:
Ok, might be good to add a legend `{{pod}}` on the first row panels.

mapno:
> Ok, might be good to add a legend `{{pod}}` on the first row panels.

It's added, but it's hidden because it could become too cluttered in big deployments. Similar to Tempo's operational dashboard which does the same (you can see where I took inspiration from :))

PR #1142, opened 2021-11-30
[dev.multiple-integrations] Prototype new integrations subsystem

rfratto:
#### PR Description 
Prototype a new integrations subsystem that would be able to natively handle multiple instances of integrations and an SD-compatible API for discovering integration targets (#1133).

This new subsystem is extension based, meaning there are extra interfaces that integrations may optionally implement to expose extra behaviors. The only extension interfaces provided here are for metrics.  

#### Notes to the Reviewer
The design of this subsystem could hypothetically be used to implement integrations with a dynamically changing set of targets (i.e., integrations using SD).

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

Issue #581, opened 2021-05-04
Consider making certain configuration options available via flag only


rfratto:
The upcoming /-/reload support has made a lot of sacrifices to be implemented. Many components do not support dynamic configuration and need to be fully shut down and replaced to emulate a configuration change. 

I've not been happy with the code I've had to do this, and I'm concerned about its stability. The user experience of /-/reload is also challenging. Given these components aren't meant to be dynamically updated, the system can be left in an undefined state when providing an invalid config file. 

I suggest a breaking change of **removing all configuration options that should be static** from the config file and exposing them only via flags. 

This would trim down the config file to only consist of the following:

1. Prometheus instance configs
2. Loki instance configs 
3. Tempo instance configs 
4. Integrations

Loki, Tempo, and Integrations would still need to be fully shut down to be updated, but this is OK. The sources of trouble implementing reload was the HTTP server and scraping service. 

I'm on the fence about this. I considered it when initially implementing config reload, but opted for the current approach instead to try it out. Now, I think it'd be better to simplify the config file and code required to make reload work.

56quarters:
FWIW, this is what Prometheus does prometheus/prometheus#6188

Anything that isn't hot-reloadable goes into a CLI flag.

rfratto:
Yep, following Prometheus' lead is a good reason to want to do this. 

I'm wary about breaking change fatigue (if such a thing exists). We've been making quite a few breaking changes lately, and although they've been small, it's been frequent. Maybe this isn't something we want to do immediately, but after a few months or so.

kfp:
Consider making this consumable via environment variables instead of/in addition to command args. Command args are more buried in k8s land where env variables are first class.

rfratto:
This is something that is going on in the background with the development of Flow. I'm going to close this as not planned.

PR #1170, opened 2021-12-15
config/features: create package to standardize experimental features

### LINKED ISSUE #1163: Proposal: experimental features
The maintainers of Grafana Agent try to keep the `main` branch in a state where it is as close as possible to being releasable. This makes it easy for us to arbitrarily cut a release to fix a critical bug. 

The primary tool used to keep the main branch in this state are the [development branches](https://github.com/grafana/agent/branches/all?query=dev.). Dev branches are useful when merging code that would otherwise block a release.  However, dev branches aren't perfect, and they separate themselves from the potential audience for a feature. 

This issue proposes a formal method for introducing _experimental features_, similar to Prometheus' [feature flags](https://prometheus.io/docs/prometheus/latest/feature_flags/). 

### Context 

We are starting to merge experimental features into main with the introduction of #1143. We should agree on a standard way of enabling and handling experimental features before the main branch is released.

### Requirements 

1. We must be able to add flags that exist in other packages (as we inherit flags from other modules) 
2. We must be able to interoperate with flags from the standard `flags` package (as the modules from requirement 1 use the standard package)

### Proposal 

Add a new a global flag, `--enable-features`, which accepts a comma-delimited list of features to enable. 

Then, define a standardized method `Validate(*flag.FlagSet)` on relevant config structs which does the following:

1. Applies default values where appropriate 
2. Performs validations of values where appropriate, including flags. 

The provided `*flag.FlagSet` can compare the value of the `--enable-features` flag against set flags from the config struct. See [this playground example](https://go.dev/play/p/NU_ql9QhTwr) for one way of doing this. 

If, for some reason, a Config struct needs to track that a specific feature is enabled, it may do so from its Validate method. 

#### Environment Variable

Support reading a new environment variable, `ENABLED_AGENT_EXPERIMENTS`, which contains a comma-separated list of experiments to enable. At runtime, Grafana Agent may define extra flags or change behavior based on whether a specific experiment is enabled. 

I would prefer not having to add environment variables into the bucket of ways Grafana Agents' behavior could be modified. If possible, I would prefer to keep static configuration options as flags and dynamic configuration options as YAML values (#581).

rfratto:
#### PR Description 
This commit introduces a new package, pkg/config/features, which allows defining a set of features and validating whether flags associated with those features are allowed to be set.

Closes #1163

#### Notes to the Reviewer
Renames the `config-urls` feature flag to `remote-configs`, since they might not necessarily be URLs forever.

Removes the old `--experiment.config-urls.enable` flag in favor of `--enable-features=remote-configs`.

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -13,6 +13,7 @@ import (
 	"github.com/drone/envsubst/v2"
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
+	"github.com/grafana/agent/pkg/config/features"
 	"github.com/grafana/agent/pkg/integrations"
 	"github.com/grafana/agent/pkg/logs"
 	"github.com/grafana/agent/pkg/metrics"
@@ -26,6 +27,14 @@ import (
 	"github.com/stretchr/testify/require"
 	"github.com/weaveworks/common/server"
 	"gopkg.in/yaml.v2"
+)
+
+var (
+	featRemoteConfigs = features.Feature("remote-configs")
+
+	allFeatures = []features.Feature{
+		featRemoteConfigs,
+	}
 )
 
 // DefaultConfig holds default settings for all the subsystems.
@@ -54,9 +63,8 @@ type Config struct {
 	Deprecations []string `yaml:"-"`
 
 	// Remote config options
-	ExperimentalConfigURLs bool   `yaml:"-"`
-	BasicAuthUser          string `yaml:"-"`
-	BasicAuthPassFile      string `yaml:"-"`
+	BasicAuthUser     string `yaml:"-"`
+	BasicAuthPassFile string `yaml:"-"`
 
 	// Toggle for config endpoint(s)
 	EnableConfigEndpoints bool `yaml:"-"`
@@ -157,8 +165,8 @@ func (c *Config) LogDeprecations(l log.Logger) {
 	}
 }
 
-// ApplyDefaults sets default values in the config
-func (c *Config) ApplyDefaults() error {
+// Validate validates the config, flags, and sets default values.
+func (c *Config) Validate(fs *flag.FlagSet) error {
 	if err := c.Metrics.ApplyDefaults(); err != nil {
 		return err
 	}
@@ -187,7 +195,15 @@ func (c *Config) ApplyDefaults() error {
 
 	c.Metrics.ServiceConfig.APIEnableGetConfiguration = c.EnableConfigEndpoints
 
-	return nil
+	// Don't validate flags if there's no FlagSet. Used for testing.
+	if fs == nil {
+		return nil
+	}
+	deps := []features.Dependency{
+		{Flag: "config.url.basic-auth-user", Feature: featRemoteConfigs},
+		{Flag: "config.url.basic-auth-password-file", Feature: featRemoteConfigs},
+	}
+	return features.Validate(fs, deps)
 }
 
 // RegisterFlags registers flags in underlying configs
@@ -201,10 +217,9 @@ func (c *Config) RegisterFlags(f *flag.FlagSet) {
 	f.IntVar(&c.ReloadPort, "reload-port", 0, "port to expose a secondary server for /-/reload on. 0 disables secondary server.")
 
 	f.StringVar(&c.BasicAuthUser, "config.url.basic-auth-user", "",
-		"basic auth username for fetching remote config. (requires config-urls experiment to be enabled")
+		"basic auth username for fetching remote config. (requires remote-configs experiment to be enabled")
 	f.StringVar(&c.BasicAuthPassFile, "config.url.basic-auth-password-file", "",
-		"path to file containing basic auth password for fetching remote config. (requires config-urls experiment to be enabled")
-	f.BoolVar(&c.ExperimentalConfigURLs, "experiment.config-urls.enable", false, "enable experimental remote config URLs feature")
+		"path to file containing basic auth password for fetching remote config. (requires remote-configs experiment to be enabled")
 
 	f.BoolVar(&c.EnableConfigEndpoints, "config.enable-read-api", false, "Enables the /-/config and /agent/api/v1/configs/{name} APIs. Be aware that secrets could be exposed by enabling these endpoints!")
 }
@@ -293,7 +308,7 @@ func getenv(name string) string {
 // args.
 func Load(fs *flag.FlagSet, args []string) (*Config, error) {
 	return load(fs, args, func(url string, expand bool, c *Config) error {
-		if c.ExperimentalConfigURLs {
+		if features.Enabled(fs, featRemoteConfigs) {
 			return LoadRemote(url, expand, c)
 		}
 		return LoadFile(url, expand, c)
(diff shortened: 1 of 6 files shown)


pkg/config/features/features.go
```diff
@@ -0,0 +1,155 @@
+// Package features enables a way to encode enabled enabled features in a
```
rfratto, line 1: disable disbable the enabled enabled typo 

```suggestion
// Package features enables a way to encode enabled features in a
```

rfratto: Typo in my comment to fix my typo not intentional :)

pkg/config/config.go
```diff
@@ -157,8 +165,8 @@ func (c *Config) LogDeprecations(l log.Logger) {
 	}
 }
 
-// ApplyDefaults sets default values in the config
-func (c *Config) ApplyDefaults() error {
+// Validate validates the config, flags, and sets default values.
+func (c *Config) Validate(fs *flag.FlagSet) error {
```
rfratto, line 169: I think this, alongside what #1142 (short term) and #1140 (longer term) may need, is perhaps calling for a better way of how we handle configs across the codebase. 

Today, we have many config structs spread across many packages. These config structs are (essentially) globals for a subsystem, where they contain their YAML schema, command line flags, and any other runtime value we needed to pass to them. Combining these like this complicates how we need to load config files, where the current process is as follows:

1. First, we define flags, which sets defaults against our struct. Flags take precedence over YAML, where a setting may be defined in both places (shoutout to my best friend #581).
2. Then, we parse flags just so we know what the config file flag is set to.
3. Next, we retrieve and unmarshal the config file as YAML. 
4. Finally, we parse the flags _again_, since the values set from the definition have almost definitely be overridden. Parsing again will give provided flags the proper precedence. 

Unmarshaling as YAML also tends to be a troublemaker because of how we attempt to provide defaults. We implement default settings through the following:

```go
var Default = Config{ ... } 
type Config struct { ... }

// UnmarshalYAML implements yaml.Umarshaler and sets c to Default 
// and unmarshals from YAML. Fields set in the YAML will override 
// the default values.
func (c *Config) UnmarshalYAML(f func(interface{}) error) error {
  *c = Default 
  type config Config 
  return f((*config)(c)) 
}
```

To make this even more complicated, #1142 will almost definitely require using a different schema for integrations if the integrations v2 feature is enabled. Maybe we can find a way to avoid this for now, but I generally don't think you can always avoid the need for runtime schema changes in a project where you do experimental features. #1140 is another good candidate for maybe having a different schema for the metrics subsystem if metrics v2 is enabled via flag. 

Even in the future with #581, we're still going to have this problem where loading configs is very convoluted and smelly. We may be able to separate out types in subsystems into dynamically configurable (YAML, `Config`) statically configurable (flags, `Options`). That might help with the complexity, but not much with the dynamic schema bit, and it all generally still smells a little stinky.

This is a lot of context with no answers, and I'll think on this more tomorrow. It feels like we have a few separate problems right now:

1. Flags and YAML are fighting with each other. We don't _want_ to support both. The YAML file will always need to be dynamically reloadable. Anything that is not dynamically reloadable should be a flag. 

2. Our configs struct are data monoliths that hold a cosmic horror of configs coming from a combination of YAML, flags, or generated at runtime after the config is loaded. 

3. There is the possibility of a near future where we will need to change the schema used for subsets of the config at runtime depending on the set of enabled feature flags. 

I'm hoping there's an obvious solution that I'm missing here that everyone else is already using :)

rfratto: > There is the possibility of a near future where we will need to change the schema used for subsets of the config at runtime depending on the set of enabled feature flags.

@sdboyer not to continue to complicate things, but...

rfratto: > We might be able to [...] decouple the config of a subsystem from how it is populated.

One final thought: if I were to rewrite everything from scratch, this is definitely the approach I would try taking first. I would [try] to treat our subsystems as libraries which are agnostic to how they are configured, and push the responsibility of how configs are represented to end users closer to the main package. 

Doing it that way would almost definitely give you a depressing amount of repetition when you inevitably define all your types twice to create a mapping of YAML -> agnostic config, but it would also make it a lot easier to be consistent and avoid the four problems we're struggling with now.

sdboyer: >> There is the possibility of a near future where we will need to change the schema used for subsets of the config at runtime depending on the set of enabled feature flags.

> @sdboyer not to continue to complicate what you and I were talking about

i ain't scurred 😝 

Reading over this main system/subsystem distinction, it sounds like what you're aiming for is dynamic schema composition. That's actually quite similar to what Grafana needs - eventually, when a plugin is installed on a given Grafana instance, we'll need to be able to load up the CUE representing its scuemata lineage and inject it into the overall schema for dashboards. Accommodating different injection patterns is why we have three 'variants' of the dashboard: [base](https://github.com/grafana/grafana/blob/main/packages/grafana-schema/src/scuemata/dashboard/dashboard.cue#L385) (purely the root/system properties), [dist](https://github.com/grafana/grafana/blob/main/packages/grafana-schema/src/scuemata/dashboard/dist/family.cue) (base + all core plugins), instance (instance + plugins installed on a particular instance).

That composition system, specifically the "instance" case, seems like precisely what you're looking for here. And while the dist approach is now encoded directly in CUE, my original implementation was assembled dynamically in pure Go, and this [compose](https://github.com/grafana/grafana/blob/main/packages/grafana-schema/src/scuemata/dashboard/dashboard.cue#L385) pattern is designed specifically to be usable in either approach.

rfratto: Ah, lovely, scuemata to the rescue yet again :)

pkg/config/features/features.go
```diff
@@ -0,0 +1,155 @@
...
+// Validate returns an error if any flags from deps were used without the
+// corresponding feature being enabled.
+//
+// If deps references a flag that is not in fs, Validate will panic.
+func Validate(fs *flag.FlagSet, deps []Dependency) error {
+	depLookup := make(map[string]Dependency, len(deps))
+
+	for _, dep := range deps {
+		if fs.Lookup(dep.Flag) == nil {
+			panic(fmt.Sprintf("flag %q does not eixst in fs", dep.Flag))
```
rlankfo, line 82:
```suggestion
			panic(fmt.Sprintf("flag %q does not exist in fs", dep.Flag))
```

rlankfo (approved):
LGTM overall :+1: nice work

pkg/config/config.go
rfratto: > To make this even more complicated, #1142 will almost definitely require using a different schema for integrations if the integrations v2 feature is enabled. This would mean that Validate would need to exist purely for the purpose of reading flags so it can influence how UnmarshalYAML works. Maybe we can avoid needing this for #1142, but I generally think we'll struggle to avoid experimental schema changes. (#1140 seems likely to do it when the metrics v2 feature is enabled).
> 
> [...]
>
> I'm hoping there's an obvious solution that I'm missing here that everyone else is already using :)

🎉: Deferred/partial unmarshaling: https://github.com/grafana/agent/pull/1185

PR #1177, opened 2021-12-16
Remove log-level flag from systemd unit file

### LINKED ISSUE #1172: systemd services shouldn't pass `--log.level`
https://github.com/grafana/agent/blob/d4f9cb4dac4c719f4e302074aa0dc31ddccf37c1/packaging/deb/grafana-agent.service#L12

Most of our example configs set `log_level` in the agent's YAML file. Users might not expect that this is also configurable via a flag, that the flag takes precedence, and is set by default when you install the agent as a DEB or RPM. I've seen a few people get tripped up by this, and I've forgotten a few times myself. 

We should stop passing the flag for both services and just fall back to the config file defaults.

jpkrohling:
#### PR Description 

This PR removes the log-level flag from the systemd unit file, so that only the value from the config file is taken into consideration.

Fixes #1172

#### Notes to the Reviewer

None.

#### PR Checklist

- [x] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

Signed-off-by: Juraci Paixão Kröhling <juraci@kroehling.de>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
 
 - [BUGFIX] Fix usage of POSTGRES_EXPORTER_DATA_SOURCE_NAME when using postgres_exporter integration (@f11r)
 
+- [CHANGE] Remove log-level flag from systemd unit file (@jpkrohling)
+
 # v0.21.2 (2021-12-08)
 
 - [SECURITY] This release contains a fix for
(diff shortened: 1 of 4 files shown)


rfratto:
Thanks! Can you remove it for the RPM service file too, and take the environment variable out of the var file? (All of those should be in the packaging folder)

jpkrohling:
Done!

rfratto (approved):
Great, thank you!

PR #1184, opened 2021-12-16
Merge main into dev.multiple-integrations

rfratto:

PR #1185, opened 2021-12-17
[dev.multiple-integrations] Hide integrations/v2 behind a feature flag

rfratto:
#### PR Description 
This PR hides integrations/v2 behind a feature flag, `integrations-next`. The default is for the stable integrations subsystem to run.

To do this, we use a new pattern: deferred unmarshaling. Until we know which integrations subsystem version to use, the Config struct initially only holds the raw YAML. After the version is provided, the raw YAML is unmarshaled again into the final representation. The deferred unmarshaling pattern is one of the more consistent ways to deal with the problem I [wrote about](https://github.com/grafana/agent/pull/1170#discussion_r769253988) in #1170.

#### Notes to the Reviewer
I went through several attempts at this before ending up with deferred unmarshaling. While there are other methods that used slightly less code, deferred unmarshaling is the least surprising and most usable throughout the various ways you might unmarshal a struct from YAML.

There was code related to defaults of the v1 subsystem that was removed in a PR. That code has been reintroduced and moved to `pkg/integrations.(*ManagerConfig).ApplyDefaults`. 

Note that the dev branch hasn't merged in main yet, so it doesn't have access to the official feature flag package. Merging main in is blocked by this PR, since the tests for integrations expect the v1 library.

Once this PR is merged, the merge main PR (#1184) will be updated to use a real feature flag.

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated


cmd/agent/entrypoint.go
```diff
@@ -41,9 +40,9 @@ type Entrypoint struct {
 
 	srv          *server.Server
 	promMetrics  *metrics.Agent
-	lokiLogs     *loki.Logs
+	lokiLogs     *logs.Logs
```
rfratto, line 43: My editor was complaining about this; we had the package imported twice with a different alias.

rfratto:
@rgeyer @gaantunes Sorry for all the emails :) This is ready for review now.

rfratto:
I'ma merge this, I think it's inoffensive enough and can get a deeper review later.

Issue #1183, opened 2021-12-16
[dev.multiple-integrations] Proposal: integrations v2 should send auto-collected metrics to a specific instance


rfratto:
Today, integrations work by dynamically creating a metrics instance for every single integration. Without instance sharing, which is scheduled to be deprecated (#398), this has a significant overhead when running many integrations, as each integration would receive its own dedicated WAL. 

Work has been ongoing in a dev.multiple-integrations branch which recently introduced a new iteration of the integrations subsystem. This new subsystem currently works by generating one single instance for all running integrations. While this is better, it still has three problems:

1. It still has _some_ overhead. This overhead can be avoided by disabling self-scraping and manually modifying an instance to use the new SD API to collect integrations. However, it's an annoying workaround that shouldn't have to be required to reduce resource usage. 

2. More features being added to the agent work by referencing an existing instance, such as auto-logging traces. This leaves the integration mechanism of creating a new instance feel weirdly out of place. 

3. It is unclear to users why `prometheus_remote_write` exists in integrations, which will be extra unclear once we start creating non-metrics integrations.

With integrations in the middle of a re-write which will be hidden behind a feature flag, I propose that we use the opportunity to align the rewrite with the rest of the codebase and have auto-collect metrics work by referencing a metrics instance defined in the config file.  

The schema would be as follows:

```yaml
integrations:
  # Config for metrics-based integrations.
  metrics:
    # Automatic scraping defaults. Individual instances of integrations 
    # may override these. 
    autoscrape: 
      [enable: <boolean> | default = true] 
      [metrics_instance: <string> | default = "default" ] 
      [scrape_interval: <duration> | default = <metrics.global.scrape_interval> ]  
      [scrape_timeout: <duration> | default = <metrics.global.scrape_timeout> ]  
    
  # ... integration configs follow ... 
```

The instance autoscraped metrics will be sent to defaults to "default." If a metrics instance with that name doesn't exist, the config will fail to load.

Additionally, the base set of settings for metrics-based integrations will change to the following schema:

```yaml
# Automatic scraping settings. Overrides defaults. Unlike the common autoscrape 
# settings, this includes relabeling rules.
autoscrape:
  [enable: <boolean> | default = <integrations.metrics.autoscrape.enable> ] 
  [metrics_instance: <string> | default = <integrations.metrics.autoscrape.metrics_instance> ] 
  [scrape_interval: <duration> | default = <integrations.metrics.autoscrape.scrape_interval> ]  
[...]

PR #1192, opened 2021-12-17
pkg/integrations: decouple common config from integration-specific configs

rfratto:
#### PR Description 
This PR removes the requirement that `integrations.Config` implement a method to expose common config settings.

Instead, common configs are now automatically unmarshaled alongside integrations. 

While this is somewhat useful for reducing the smallest amount of boilerplate, it is more immediately useful for #1183, which will use a _different_ set of common configs when running with the integrations-next feature flag. 

#### Notes to the Reviewer
This is some unfortunate complexity 😞 The requirement of strictly marshaling/unmarshaling into two inlined types where one may implement yaml.Unmarshaler is mostly why this is a pain.

#### PR Checklist
Internal change, nothing here needed.

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/util/yaml.go
+++ b/pkg/util/yaml.go
@@ -30,6 +30,7 @@ func (r RawYAML) MarshalYAML() (interface{}, error) {
 	return r.Map()
 }
 
+// Map converts the raw YAML into a yaml.MapSlice.
 func (r RawYAML) Map() (yaml.MapSlice, error) {
 	var ms yaml.MapSlice
 	if err := yaml.Unmarshal(r, &ms); err != nil {
--- a/pkg/util/yaml.go
+++ b/pkg/util/yaml.go
@@ -1,0 +1,119 @@
+package util
+
+import (
+	"errors"
+	"regexp"
+	"sort"
+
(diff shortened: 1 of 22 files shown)


rgeyer (approved):
LGTM 🚀

rfratto:
For the anxious: I tested these changes against our docker-compose integrations environment, and everything is working.

pkg/util/yaml.go
```diff
@@ -0,0 +1,120 @@
+package util
```
rlankfo, line 1: Should we add unit tests to cover the new functionality introduced by this yaml util?

rlankfo (approved):
LGTM, just quick q on unit tests :+1:

pkg/util/yaml.go
rfratto: Nah, not strictly necessary, the existing tests cover this. Though at some point we'd want to make smaller unit tests and test these directly.

rfratto:
I don't know what's up with the CLA bot but I've definitely signed it 🤷

PR #1202, opened 2021-12-21
Remote write dashboard: show the sample rates

bboreham:
Instead of showing the diff between in and out rates, show them both.
The total rate is key information alongside shards, latency, etc., and the differential can be inferred from looking at in and out rates, or the backlog on the top line.

Making these panels the same width as those nearby also aids readability.

Dropped samples already had a separate panel.

Before:
image

After:
image

#### PR Checklist

- [x] CHANGELOG updated 
- [ ] Documentation added - couldn't find any to update
- NA Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
 
 - [ENHANCEMENT] Updated OTel to v0.40.0 (@mapno)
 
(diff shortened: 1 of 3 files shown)


grafanalf (approved):

rfratto (approved):
Thanks! (I fixed the CHANGELOG conflict)

PR #1203, opened 2021-12-21
Remote write dashboard: add mean latency

bboreham:
While p99 latency is an indicator of problems, overall throughput is more dependent on mean latency so add it to the panel.

Before:
image

After:
image

Or, selecting just one mean:
image

#### PR Checklist

- [x] CHANGELOG updated 
- [ ] Documentation added - couldn't find any to update
- NA Tests updated


grafanalf (approved):

rfratto (approved):
Thanks Bryan!

PR #1204, opened 2021-12-21
Windows Service accept commands immediately

### LINKED ISSUE #1128: Windows Service start timeout if WAL takes to long to load
While replaying the WAL the service does not respond to Windows Service commands, which means if it takes over 30s then the service fails to start and is stopped.

The most likely change is to have the service start handling service commands immediately.

mattdurham:
#### PR Description 

This PR starts the command loop for handling windows service messages immediately instead of waiting for the agent to fully load. It still ensures that the config is loaded correctly. There are some possible race conditions but those are if the service is told to shutdown so impact is low. ie if you start the service then immediately send the stop signal, the entrypoint stop command may not be called.

Closes #1128 

#### PR Checklist

- [x] CHANGELOG updated 
- [na] Documentation added
- [na] Tests updated

### DIFF
--- a/cmd/agent/service_windows.go
+++ b/cmd/agent/service_windows.go
@@ -45,6 +45,11 @@ func (m *AgentService) Execute(args []string, serviceRequests <-chan svc.ChangeR
 		log.Fatalln(err)
 	}
 
+	// Pause is not accepted, we immediately set the service as running and trigger the entrypoint load in the background
+	// this is because the WAL is reloaded and the timeout for a windows service starting is 30 seconds. In this case
+	// the service is running but Agent may still be starting up reading the WAL and doing other operations.
+	changes <- svc.Status{State: svc.Running, Accepts: cmdsAccepted}
+
 	// After this point we can start using go-kit logging.
 	logger := util.NewWindowsEventLogger(&cfg.Server)
 	util_log.Logger = logger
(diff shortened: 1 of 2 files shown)


cmd/agent/service_windows.go
```diff
@@ -74,20 +80,23 @@ loop:
 			case svc.Interrogate:
 				changes <- c.CurrentStatus
 			case svc.Stop, svc.Shutdown:
-				ep.Stop()
```
rfratto, line 77: Hmm, doesn't this mean there's no way to stop the service? Finding a way to tie this in cleanly would be a PITA but IMO for now we can give `ep.Start` a context that we can cancel whenever we need it to shut down.

rfratto (approved):
Sorry, I know I'm supposed to be on vacation but just happened to notice this. LGTM % nit

cmd/agent/service_windows.go
mattdurham: When you say service do you mean the Windows service or entrypoint? The windows service will stop when this returns and the entrypoint will normally stop on line 98. Though there is a possible race condition.

rfratto: Oh, I see, right. Yeah, this is probably fine as-is. Never mind :)

rlankfo (approved):

PR #1206, opened 2021-12-21
Jsonnet: set HOSTNAME to ensure the log subsystem works

56quarters:
Promtail requires the HOSTNAME env var to be set otherwise it will
silently do nothing resulting in logs not being scraped with no indication
anything is wrong.

Signed-off-by: Nick Pillitteri <nick.pillitteri@grafana.com>

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/production/tanka/grafana-agent/v2/internal/base.libsonnet
+++ b/production/tanka/grafana-agent/v2/internal/base.libsonnet
@@ -4,6 +4,7 @@ function(name='grafana-agent', namespace='') {
   local container = k.core.v1.container,
   local configMap = k.core.v1.configMap,
   local containerPort = k.core.v1.containerPort,
+  local envVar = k.core.v1.envVar,
   local policyRule = k.rbac.v1.policyRule,
   local serviceAccount = k.core.v1.serviceAccount,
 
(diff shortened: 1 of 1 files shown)


rfratto (approved):
Perfect 👍

rfratto:
This might cause label collisions when using metrics integrations and not a DaemonSet. Can this be made optional or only modify DaemonSet controllers?

rfratto (changes requested):
Sorry, reviewed too fast :) 

This might cause label collisions when using metrics integrations and not a DaemonSet. Can this be made optional or only modify DaemonSet controllers?

56quarters:
> Sorry, reviewed too fast :)
> 
> This might cause label collisions when using metrics integrations and not a DaemonSet. Can this be made optional or only modify DaemonSet controllers?

Sure, which would you prefer (optional or only for DaemonSet or both)?

rfratto:
I think for now it makes sense to always exist for DaemonSet. We can make it optional later if there's a reason for it to be.

rfratto (approved):
Now it’s perfect :)

PR #1208, opened 2021-12-21
Fix cadvisor metrics configuration

pkoenig10:
#### PR Description 

The `cadvisor` metrics configuration has a couple bugs:
- When configuring `disabled_metrics` or `enabled_metrics`, only the last metric value will be applied.

    The [`MetricSet.Set`](https://github.com/google/cadvisor/blob/e0fab6233991e198fe53c25ee031e3d5a42c3f05/container/factory.go#L123-L136) method takes a comma-separated list of metrics and overwrites the metric set. It does not add the given metric to the set.

- Explicitly configuring `disabled_metrics` to be an empty list does not enable all metrics. Instead, the default disable metrics are still disabled.

    The current implementation does not differentiate between an absent configuration and an explicit empty configuration.

#### PR Checklist

- [ ] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/integrations/cadvisor/cadvisor.go
+++ b/pkg/integrations/cadvisor/cadvisor.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/go-kit/kit/log"
@@ -35,8 +36,8 @@ import (
 
 // Matching the default disabled set from cadvisor - https://github.com/google/cadvisor/blob/3c6e3093c5ca65c57368845ddaea2b4ca6bc0da8/cmd/cadvisor.go#L78-L93
 // Note: This *could* be kept in sync with upstream by using the following. However, that would require importing the github.com/google/cadvisor/cmd package, which introduces some dependency conflicts that weren't worth the hassle IMHO.
-// var disabledMetricsSet = *flag.Lookup("disable_metrics").Value.(*container.MetricSet)
-var disabledMetricsSet = container.MetricSet{
+// var disabledMetrics = *flag.Lookup("disable_metrics").Value.(*container.MetricSet)
+var disabledMetrics = container.MetricSet{
 	container.MemoryNumaMetrics:              struct{}{},
 	container.NetworkTcpUsageMetrics:         struct{}{},
 	container.NetworkUdpUsageMetrics:         struct{}{},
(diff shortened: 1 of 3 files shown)


pkoenig10:
I was surprised to discover that some metrics are disabled by default. While this matches the behavior of cadvisor, it seems like a fairly unintuitive default. I was initially confused why some metrics in the [cadvisor documentation](https://github.com/google/cadvisor/blob/e0fab6233991e198fe53c25ee031e3d5a42c3f05/docs/storage/prometheus.md) were not being produced.

Why does cadvisor exclude some metrics by default? And why was that behavior replicated here? Would you be willing to consider enabling all metrics by default?

mattdurham:
Pinging @rgeyer for what metrics we do for default, though it might be til after the holidays before we get an answer.

pkg/integrations/cadvisor/cadvisor.go
```diff
@@ -52,30 +53,27 @@ var disabledMetricsSet = container.MetricSet{
...
-	// Clear default disabled metrics if explicit disabled metrics are configured
-	if len(c.DisabledMetrics) > 0 {
-		disabledMetricsSet = container.MetricSet{}
-	}
-	for _, d := range c.DisabledMetrics {
-		if err := disabledMetricsSet.Set(d); err != nil {
-			return includeMetrics, fmt.Errorf("failed to set disabled metric: %w", err)
+	var enabledMetrics, includedMetrics container.MetricSet
+
+	if c.DisabledMetrics != nil {
```
rgeyer, line 58: Both `c.DisabledMetrics` and `c.EnabledMetrics` have zero values that come from the `Config` struct , so these conditionals would always pass, and set the enabled. and disabled sets to empty lists.

https://go.dev/play/p/Brw6Dc27WHn

rgeyer (changes requested):
@pkoenig10 First, thank you for catching those logic errors.

As for the standard set of disabled metrics.. I don't know the exact reasoning behind the standard list of disabled metrics, however, in looking at them, they all seem to be metrics which would be expensive to collect and report. For example fetching all TCP and UDP metrics can be very taxing on the host CPU.

I would propose that we do two things.

1. Allow a "special" metric name, something like `none`, which when specified in the list of `disabled_metrics` would override the defaults. I found [this response](https://github.com/google/cadvisor/issues/1298) to an issue on cAdvisor that recommends setting an empty `--disabled_metrics` which we'd be duplicating here.
2. Document the behavior better in `./docs/configuration/integrations/cadvisor-config.md`. I did a poor job of this, and should have included more of the detail from the [cAdvisor runtime options](https://github.com/google/cadvisor/blob/master/docs/runtime_options.md). Ours would need to explicitly document the "special" metric name proposed in #1 above.

I'm happy to author a PR with my proposed changes, and have you take a look to see if it resolves your concerns.

pkg/integrations/cadvisor/cadvisor.go
pkoenig10: The zero value of a slice is `nil`. It can be a bit confusing because printing a `nil` slice results in `[]`.

https://github.com/golang/go/issues/12780
https://go.dev/play/p/Q_PaZe6c3Go

pkoenig10:
> I don't know the exact reasoning behind the standard list of disabled metrics, however, in looking at them, they all seem to be metrics which would be expensive to collect and report.

Yeah this would be my guess. In that case, omitting these by default feels like a reasonable choice. It also feels reasonable to match the cAdvisor default behavior to make it easier for users to migrate.

> Allow a "special" metric name, something like none, which when specified in the list of disabled_metrics would override the defaults.

This feels like a bad design choice to me. If cAdvisor adds a metric named `none` (however unlikely that may be), we'd be in a tough spot. It's also unintuitive for users to have these sort of "magic" values.

I think a better approach would be to simply match the cAdvisor configuration and require users to explicitly configure an empty list of `disabled_metrics`. That was what I originally tried to do before discover the logic error here.

pkg/integrations/cadvisor/cadvisor.go
rgeyer: TIL. Thanks for setting me straight on this!

rgeyer:
> > Allow a "special" metric name, something like none, which when specified in the list of disabled_metrics would override the defaults.
> 
> This feels like a bad design choice to me. If cAdvisor adds a metric named `none` (however unlikely that may be), we'd be in a tough spot. It's also unintuitive for users to have these sort of "magic" values.
> 
> I think a better approach would be to simply match the cAdvisor configuration and require users to explicitly configure an empty list of `disabled_metrics`. That was what I originally tried to do - I only discovered the logic error here after being surprised that using an empty list didn't work.

Yep, totally agree with this. I was proposing this only because of my ignorance of the zero value of an empty slice. I assumed we had to distinguish between "not set" (`nil`) and "slice with zero elements".

Now that I understand that behavior better, I'm happy with this implemented as-is. In the yaml config we can unset the default disabled metrics with `disabled_metrics: []`.

Thanks!

rgeyer (approved):
LGTM based on above conversation. 😄

pkoenig10:
Thanks for the quick review!

rgeyer:
@mattdurham or @rfratto does this warrant a changelog entry? If so, I can add it, else I'll merge this. :)

mattdurham:
Yes please, given that this seems impactful having a changelog entry is nice. Something like

`- [BUGFIX] Fix cAdvisor so it collects all defined metrics instead of the last (@pkoenig10)`

rgeyer:
Merged. Thanks again @pkoenig10!

PR #1209, opened 2021-12-22
Jsonnet: add withService() helper for all v2 controllers

56quarters:
Add a helper method that allows people to create associated services
for deployment, statefulset, or daemonset Agents. This is particularly
useful for Agents that need to accept traces since they are pushed
to the agent and clients need a consistent URL.

Signed-off-by: Nick Pillitteri <nick.pillitteri@grafana.com>

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/production/tanka/grafana-agent/v2/main.libsonnet
+++ b/production/tanka/grafana-agent/v2/main.libsonnet
@@ -38,4 +38,6 @@ local container = k.core.v1.container;
     (import './internal/helpers/logs.libsonnet').volumeMounts(config),
   withLogPermissions(config)::
     (import './internal/helpers/logs.libsonnet').permissions(config),
+  withService(config)::
+    (import './internal/helpers/service.libsonnet').service(config),
 }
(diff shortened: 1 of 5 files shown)


56quarters:
NB: Adding a service for a daemonset controller was necessary in my work refactoring the TNS project. With this extension, a single daemonset controller could be used for metrics, logs, and traces.

production/tanka/grafana-agent/v2/main.libsonnet
```diff
@@ -38,4 +38,6 @@ local container = k.core.v1.container;
     (import './internal/helpers/logs.libsonnet').volumeMounts(config),
   withLogPermissions(config)::
     (import './internal/helpers/logs.libsonnet').permissions(config),
+  withService(config)::
+    (import './internal/helpers/service.libsonnet').service(config),
```
rfratto, line 42: All of these should probably default to `config={}`. The original intent was that they were optional, but I didn't realize the optional wouldn't propagate from the implementation to the exposed API.

Anyways, not a blocker, just a thought.

rfratto (approved):
LGTM!

PR #1217, opened 2021-12-31
Add dependabot to update go modules and github actions.

tomwilkie:
Signed-off-by: Tom Wilkie <tom.wilkie@gmail.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    open-pull-requests-limit: 2


rfratto (commented):
I'm a little mixed on this. I was looking into dependency tracking a while ago, and initially considered dependabot, but decided against it. 

There were two reasons for this:

1. So many of our dependencies are exposed directly to users and need a careful eye when updating. e.g., beyond increasing the dependency version, we might need to update docs or add new fields to configs. 
2. We rely on ephemeral forks to iterate quickly. Forks go away when upstream PRs are merged, but it means that we lose out on dependency checking while they're in use. 

Instead of using dependabot, I ended up writing [rfratto/depcheck](https://github.com/rfratto/depcheck) to signal to devs when we need to look into updating something. It's run [weekly](https://github.com/grafana/agent/blob/main/.github/workflows/depcheck.yml) and it manages [issues](https://github.com/grafana/agent/issues?q=is%3Aopen+is%3Aissue+label%3Aoutdated-dependency) for us to keep track of.

rfratto:
We've discussed this in our bug scrub and decided to merge. Thanks Tom!

rfratto (approved):

tomwilkie:
Oh, thanks Robert! I owed you a response but sorry didn't get back in time.

Do you think we might be able to use rfratto/depcheck to keep prometheus up to date in eg cortex, where dependabot stuggles?

rfratto:
> Do you think we might be able to use rfratto/depcheck to keep prometheus up to date in eg cortex, where dependabot stuggles?

Maybe! depcheck is the most useful when maintainers are struggling to keep track of when one of their dependencies using replace directives (esp. for forks) are out of date, but it can quickly become noisy and be something that you might end up [deprioritizing](https://github.com/grafana/agent/issues?q=is%3Aopen+is%3Aissue+label%3Aoutdated-dependency). 

I'd say we've had mixed success with the original intent, but it is at least useful as a signal that future work needs to be done.

PR #1223, opened 2022-01-03
[dev.operator-integrations-support] Update integration CRD names

rfratto:
Given #1198 and the intent that we will have more types of integrations that aren't just for metrics, it feels like a good move to rename IntegrationsInstance and IntegrationMonitor to MetricsIntegrationsInstance and MetricsIntegrationMonitor respectively. 

This change will allow us to configure fields that we know are commonly available on all metrics integrations (e.g.,, `instance` or `autoscrape`) but not necessarily available on integrations of different types. 

Additionally, the old "mode" field has been changed to "type", where a type is either `daemonset` (run on every node), `singleton` (run once per GrafanaAgent deployment) or `normal` (run any number of times per GrafanaAgent deployment). This aligns with the multiplexing functionality present in #1198.

"daemonset" and "singleton" both refer to integrations that can only be defined once per Grafana Agent. This is a concept unique to the operator; while node_exporter and statsd_exporter are both single-instance integrations, you would definitely want node_exporter to run on every node, but that's not necessarily also true of statsd_exporter.

This PR is some light prep work while #1198 is being reviewed. A new design doc supplanting the original one in #883 will follow, since #1998 will changes the implementation details and I've learned some lessons during my prototyping that should be written down.


rgeyer (approved):
Makes sense to me

rfratto:
This work has been supplanted by another dev branch.

PR #1224, opened 2022-01-04
[RFC] Integrations in Grafana Agent Operator

rfratto:
This RFC proposes a way to add support for integrations to Grafana Agent Operator by: 

* Defining two new CRDs:
  * `MetricsIntegration`: an integration to run
  * `IntegrationMonitor`: collection rules for integrations, similar to ServiceMonitor.
* Update `GrafanaAgent` to discover `MetricsIntegration`s and run integrations.
* Update `MetricsInstance` to discover `IntegrationMonitor`s and collect metrics from integrations.

Supersedes #883.


rfratto:
@flokli I know you've used the operator in the past and given valuable feedback, so just giving you a heads up for this in case you wanted to weigh in :)

docs/rfcs/xxxx-integrations-in-operator.md
```diff
@@ -0,0 +1,234 @@
...
+  than one `daemonset` integrations with the same `name` . Example integration:
+  `node_exporter`.
+
+* `singleton`: Declares that the `name` integration may only be run once per
+  deployment. It is invalid to have a GrafanaAgent CRD discover more than one
+  `singleton` integration with the same `name.` Example integration:
+  `statsd_exporter`.
+
+* `normal`: Declares that the integration may exist any number of times per
+  deployment. Example integration: `mysqld_exporter`. This is the default.
```
flokli, line 84:
```suggestion
  deployment. This is the default. Example integration: `mysqld_exporter`.
```

Flip the two phrases.

docs/rfcs/xxxx-integrations-in-operator.md
```diff
@@ -0,0 +1,234 @@
...
+* `type`: The type of integration
+* `config`: A YAML string declaring the configuration for the integration.
+
+`type` is critical for determining how the integration should be deployed.
+There are three supported values for `type`:
+
+* `daemonset`: Declares that the `name` integration should be run on every Node in the
+  Kubernetes cluster. It is invalid to have a GrafanaAgent CRD discover more
+  than one `daemonset` integrations with the same `name` . Example integration:
+  `node_exporter`.
```
flokli, line 76: While practical for sure, I'm not convinced we should ship all those exporters in the grafana-agent binary at all. It's probably perfectly okay to tell people to install `prometheus-node-exporter` from https://prometheus-community.github.io/helm-charts with the values necessary to get monitors installed and labels in the right shape dashboards expect them to be (and document this prominently)

flokli (commented):
Some comments. Thanks for the highlight!

flokli:
Orthogonal question: Tracing wouldn't be a daemonset integration, but its own thing, right?

Right now, I define manage that daemonset manually, outside of the operators control (yet)

docs/rfcs/xxxx-integrations-in-operator.md
```diff
@@ -0,0 +1,234 @@
...
+
+[#1198][] introduces a new service discovery (SD) API to retrieve a list of
+running integrations. The endpoint is available at
+`/agent/api/v1/metrics/integrations/sd` and returns JSON which conforms to
+Prometheus' [`http_sd_config`][http_sd_config] API.
+
+Note that integration targets returned by this API always have the following
+base labels:
+
+* `instance`: The inferred `instance` key based on integration settigs.
```
rlankfo, line 163:
```suggestion
* `instance`: The inferred `instance` key based on integration settings.
```

docs/rfcs/xxxx-integrations-in-operator.md
```diff
@@ -0,0 +1,234 @@
...
+* `__meta_agentoperator_grafanaagent_namespace`: Namespace of the owning GrafanaAgent.
+
+As a result, there is a centralized API to discover metrics endpoints for all
+running integrations, including information about the integration, the owning
+MetricsIntegration CR, and the root GrafanaAgent resource.
+
+### Scraping integrations
+
+The other new CRD, IntegrationMonitor, will be used to scrape running
+integrations. IntegrationsMonitors are discovered by a MetricsInstance, and
```
rlankfo, line 206:
```suggestion
integrations. IntegrationMonitor's are discovered by a MetricsInstance, and
```

docs/rfcs/xxxx-integrations-in-operator.md
```diff
@@ -0,0 +1,234 @@
...
+result in the generation of a integrations-specific scrape job. It is
+configured similarly to a `ServiceMonitor`, though without settings for
+endpoint port and path.
+
+IntegrationsMonitor use `http_sd_config` and the SD endpoint described in the
```
rlankfo, line 211:
```suggestion
IntegrationMonitor use `http_sd_config` and the SD endpoint described in the
```

docs/rfcs/xxxx-integrations-in-operator.md
```diff
@@ -0,0 +1,234 @@
...
+resource hierarchy, alongside MetricsInstances and LogsInstances. During
+reconcile, the following Kubernetes objects will be deployed:
+
+* One DaemonSet and Secret for all `daemonset` integrations, if
+  present in the resource hierarchy.
+
+* One Deployment and Secret for all other integrations, if non-`daemonset`
+  integrations are present in the resource hierarchy.
+
+* One Service for all integrations, if any integration was present in the
```
rlankfo, line 117: Is this one service for each integration? Or all MetricsIntegrations will share a service?

docs/rfcs/xxxx-integrations-in-operator.md
rfratto: The apostrophe isn't correct here :)

docs/rfcs/xxxx-integrations-in-operator.md
rfratto: There will be exactly one service created as part of reconciling a GrafanaAgent if its resource hierarchy contains at least one MetricsIntegration. I'll try to clear that up in the doc :)

docs/rfcs/xxxx-integrations-in-operator.md
```diff
@@ -0,0 +1,234 @@
...
+
+The Service is used for exposing the endpoint for the individual pod's
+integrations service discovery API introduced in [#1198][]. We will discuss
+this in more detail later.
+
+Some integrations may require changes to the deployed Pods to function
+properly. MetricsIntegrations will additionally support declaring `volumes`,
+`volumeMounts`, `secrets` and `configMaps`. These fields will be merged with
+the fields of the same name from the root GrafanaAgent resource when creating
+integration pods.
```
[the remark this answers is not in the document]
rfratto: Right, (unfortunately) not all of the exporters we embed support file-based secrets. If an integration doesn't support it, you'd have to declare the secret verbatim in the YAML block. But you could use it for things like [kafka_exporter](https://grafana.com/docs/agent/latest/configuration/integrations/kafka-exporter-config/):

```yaml
kind: MetricsIntegration
metadata:
  name: kafka
  namespace: default
spec:
  name: kafka_exporter
  type: normal
  config: |
    ca_file: /etc/grafana-agent/secrets/kafka-ca-file
    # ... 
  # Same "secrets" field present in GrafanaAgent.spec, where each secret 
  # is loaded from the same namespace and gets exposed at 
  # /etc/grafana-agent/secrets/<secret name>
  secrets: [kafka-ca-file]
```

docs/rfcs/xxxx-integrations-in-operator.md
```diff
@@ -0,0 +1,234 @@
...
+* Date: 2022-01-04
+* Author: Robert Fratto (@rfratto)
+
+## Background
+
+At the time of writing, Grafana Agent Operator can be used to collect metrics
+and logs from a Kubernetes cluster by deploying and managing Grafana Agent
+pods. However, logs and metrics are just a subset of the total functionality
+available in Grafana Agent. This document proposes adding support for
+integrations into the Grafana Agent Operator.
```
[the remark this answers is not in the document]
rfratto: Agreed that naming is hard :) In the context of Grafana Agent, an integration is _something_ that helps you generate telemetry data. Historically that's been embedded Prometheus exporters, but people are starting work on other types of integrations for things like logging. 

I've tried to clarify this in the doc.

docs/rfcs/xxxx-integrations-in-operator.md
rfratto: We'll probably always want to continue embedding _some_ exporters in the agent, since it makes things easier for non-Kubernetes users, which is still a userbase we're interested in helping. 

Though if an exporter only makes sense for Kubernetes (or isn't popular enough to justify embedding it), we would definitely prefer recommending installing that exporter with Helm and using a ServiceMonitor/PodMonitor/Probe for collecting metrics from it instead of embedding it directly into the agent.

rfratto:
> Orthogonal question: Tracing wouldn't be a daemonset integration, but its own thing, right?
>
> Right now, I define manage that daemonset manually, outside of the operators control (yet)

Right, Tracing support would be its own thing with its own set of CRDs at some point. The agent has four "subsytems:"

1. Metrics
2. Logs
3. Integrations 
4. Traces

We've added support for the first two subsystems in the operator, and this adds support for the third. Traces would be next up after this.

docs/rfcs/xxxx-integrations-in-operator.md
rlankfo: intended just to switch IntegrationsMonitor to IntegrationMonitor

rfratto: ah, didn't see that. fixed, thanks!

rgeyer (dismissed):
Slow to respond, but this LGTM.

rfratto:
I've been trying to find a way to support integrations in the operator since September, and the most recent attempt at implementing this proposal revealed yet more issues. 

My biggest takeaway is that integrations are really complicated, and we need to take baby steps towards a nicer solution. I've just pushed a third attempt at this proposal which supports all integrations in a generic way, but does so at the expense of some dramatic tradeoffs. It's not perfect, but I can't think of a simpler solution that we can build on top of.

PTAL @hjet @captncraig @rlankfo

docs/rfcs/xxxx-integrations-in-operator.md
```diff
@@ -0,0 +1,265 @@
...
+
+* `name`: The name of the integration (e.g., `node_exporter`, `mysqld_exporter`)
+* `type`: Information about the integration being deployed
+* `config`: YAML configuration block for the integration
+
+The `type` field is an object with the following fields:
+
+* `allNodes`: True when the `name` integration should run on all Kubernetes
+  Nodes.
+* `unique`: True when the `name` integration must be unique across a
```
hjet, line 69: To motivate this would maybe provide a sample use case here / example of an integration where you'd need this to be the case

docs/rfcs/xxxx-integrations-in-operator.md
```diff
@@ -0,0 +1,265 @@
...
+> ```yaml
+> apiVersion: monitoring.grafana.com/v1alpha1
+> kind: Integration
+> metadata:
+>   name: kafka
+>   namespace: default
+> spec:
+>   name: kafka_exporter
+>   type:
+>     hasMetrics: true
```
hjet, line 144: Is this another `type` field?

docs/rfcs/xxxx-integrations-in-operator.md
```diff
@@ -0,0 +1,265 @@
...
+>   name: mysqld_exporter
+>   config:
+>     autoscrape:
+>       enable: true
+>       # MetricsInstance <namespace>/<name> to send metrics to
+>       metrics_instance: default/primary
+>     data_source_name: root@(mysql.default:3306)/
+>     disable_collectors: [slave_status]
+> ```
+
```
hjet, line 190: For something like the `eventhandler` integration how would this work?

Would the config look something like this?

```yaml
kind: Integration
metadata:
   name: eventhandler
   namespace: default
 spec:
   name: eventhandler
   config:
       logs_instance: ?
       misc_eventhandler_config ...
```

I guess some sort of local "ghost' logs instance would be enabled (with what config?) and the integration could use that one?

docs/rfcs/xxxx-integrations-in-operator.md
rfratto: Relic of an older revision, this should be removed, sorry :)

docs/rfcs/xxxx-integrations-in-operator.md
rfratto: You'd set `logs_instance` to the `<namespace>/<name>` of the LogsInstance CR in your resource hierarchy. Does that make sense? 

```yaml 
kind: GrafanaAgent
metadata:
  name: grafana-agent-example
  namespace: default
  labels:
    app: grafana-agent-example
spec:
  logs:
    instanceSelector:
      matchLabels:
        agent: grafana-agent-example
  integrations:
    instanceSelector:
      matchLabels:
        agent: grafana-agent-example
---
kind: LogsInstance
metadata:
  name: primary
  namespace: default
  labels:
    agent: grafana-agent-example
spec:
  clients:
  - url: http://loki:8080/loki/api/v1/push
---
kind: Integration
metadata:
   name: eventhandler
   namespace: default
spec:
   name: eventhandler
   config:
     logs_instance: default/primary
     misc_eventhandler_config: ...
```

hjet: Sorry if this was already assumed, but just in case: eventhandler works by handing off log entries to a channel managed by a logs instance running in the _same_ Agent (in other words, the same Agent must run a logs instance and eventhandler integration).

The integration can be modified to not do this, and maybe write out to a file or something...

But given the above IIUC: 

- User defines a `LogsInstance`
- Operator reconciles and creates an Agent DaemonSet
- User creates a "Deployment" integration (so `allNodes: false`)
- Operator reconciles and creates an Agent Deployment running the integration
- Operator also detects LogsInstance and so will create config enabling a `logs` instance on that integrations Agent with the same config as the LogsInstance CR? 

What happens if several `LogsInstance`s are defined? Are these all appended to the integrations Agent's config? (sorry if this doesn't make sense...)

rfratto: I'll try to make this more obvious in the doc, but the idea is that the above config generates the following agent config file:

```yaml
logs:
  configs:
  - name: default/primary
    clients:
    - url: http://loki:8080/loki/api/v1/push

integrations:
  eventhandler:
    logs_instance: default/primary 
    misc_eventhandelr_config: ... 
```

Any LogsInstance or MetricsInstance in the same hierarchy as the integration will be injected into the generated config file so integrations can use them.

hjet: Gotcha understood

rlankfo (approved):
LGTM :+1:

I think we can iterate on this to make configuration easier for users in the future but with that we would likely have some trade-offs you've already considered (e.g. operator is aware of specific integrations and how to handle them).

hjet (approved):
LGTM as well! This is a great start and like @rlankfo said we can incrementally hide more config from the user / build more validation logic into operator. In most cases we'll be generating these for our users anyways....

docs/rfcs/xxxx-integrations-in-operator.md
captncraig: Also, what happens if you say `allNodes: true` and `unique: true`. Not all combinations make sense. Maybe a single field with three possibilities is the way to go (although I don't love singleton to describe it either)

captncraig (approved):
I like it. My biggest point of confusion is just the "unique" constraint and what it means. I'd like to make sure we have the correct terminology to make that distinction clearer.

docs/rfcs/xxxx-integrations-in-operator.md
rfratto: Here's a truth table for relevant integrations:

| allNodes | unique | example integration |
| -------- | ------ | ------------------- |
| false    | false  | mysqld_exporter     |
| false    | true   | statsd_exporter     |
| true     | false  | process_exporter    |
| true     | true   | node_exporter       |

Some of the table is arbitrary; there's no reason you _couldn't_ deploy statsd_exporter on all nodes. 

Also it's noteworthy to say that allNodes has no bearing on the generated agent config, while unique does.

rfratto:
> I like it. My biggest point of confusion is just the "unique" constraint and what it means. I'd like to make sure we have the correct terminology to make that distinction clearer.

Yeah, I've been struggling with the naming convention too. I'll merge the RFC as-is in Draft, and we can revisit brainstorming names to use consistently across the operator and metrics-next before this is included in a release.

rfratto:
I'm assigning this RFC-0002.

Issue #582, opened 2021-05-05
Rename binary to grafana-agent-* ?


RichiH:
In particular with https://github.com/prometheus/prometheus/pull/8785 it would make sense to be more specific in naming with the binary, I think.

rfratto:
I'm open to this. This isn't even something we're internally consistent with; the system packages install a grafana-agent while everything else is just called agent. 

I'd probably prefer not to rename the docker image though.

RichiH:
The image itself has grafana in the path name, that seems fine.

rfratto:
Hoping to make this change this week.

PR #1225, opened 2022-01-04
Agent smoke test update

rlankfo:
This PR converts the functionality from the existing smoke script to a Go program.


Makefile
```diff
@@ -307,6 +314,13 @@ endif
 dist/agent-freebsd-amd64: seego
 	$(call SetBuildVarsConditional,freebsd);  $(seego) build $(CGO_FLAGS) -o $@ ./cmd/agent
 
+#########################
+# BEGIN AGENTSMOKE DIST #
+#########################
+
+dist/agent-smoke-darwin-arm64: seego
+	$(call SetBuildVarsConditional,darwin/arm64) ;     $(seego) build $(CGO_FLAGS) -o $@ ./cmd/agent-smoke
+
```
mattdurham, line 323: Do we want to support darwin x64?

cmd/agent-smoke/main.go
```diff
@@ -0,0 +1,67 @@
...
+		kubeconfig        string
+		withTimeout       time.Duration
+		chaosFrequency    time.Duration
+		mutationFrequency time.Duration
+	)
+
+	flag.BoolVar(&debugLog, "debug", false, "enable debug logging")
+	flag.StringVar(&namespace, "namespace", "agent-smoke-test", "namespace smoke test should run in")
+	flag.StringVar(&kubeconfig, "kubeconfig", "", "absolute path to the kubeconfig file")
+	flag.DurationVar(&withTimeout, "with-timeout", time.Duration(0), "timeout after duration, example 3h")
```
mattdurham, line 27: Timeout verbiage feels off, maybe `smoke-test-duration`?

mattdurham (approved):
Looks solid, small nites.

Makefile
rlankfo: I can remove this, I had originally just added this target to test locally but we won't want to ship this.

cmd/agent-smoke/main.go
rlankfo: I think `timeout` is common terminology but I'm not opposed to changing this to just `duration` perhaps. `smoke-test-duration` seems kind of redundant.

rfratto:
Thanks! I'll look at this tomorrow

cmd/agent-smoke/main.go
```diff
@@ -0,0 +1,67 @@
...
+	flag.Parse()
+
+	logger := log.NewLogfmtLogger(log.NewSyncWriter(os.Stderr))
+	logger = log.With(logger, "ts", log.DefaultTimestampUTC, "caller", log.DefaultCaller)
+	level.AllowInfo()
+	level.Info(logger).Log("msg", "starting agent smoke framework")
+	if debugLog {
+		level.AllowDebug()
+		level.Debug(logger).Log("msg", "debug logging enabled")
+	}
```
rfratto, line 39: Can we use our existing log helper package here? I wouldn't be opposed to changing the arguments so it accepts a log level/format directly instead of a server config.  

https://github.com/grafana/agent/blob/5bf8cf452fa76c75387e30b6373630923679221c/pkg/util/logger.go#L28

(just calling `level.AllowInfo` doesn't actually do anything, it's an option you pass to `level.NewFilter`)

cmd/agent-smoke/main.go
```diff
@@ -0,0 +1,67 @@
...
+	"time"
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	smoke "github.com/grafana/agent/cmd/agent-smoke/internal"
+)
+
+func main() {
+	var (
+		debugLog          bool
```
rfratto, line 16: Nit: use `github.com/weaveworks/common/logging.Level` here instead, which implements flag.Value:

```yaml
var (
  logLevel logging.Level 
)

logLevel.RegisterFlags(flag.CommandLine)
```

cmd/agent-smoke/main.go
```diff
@@ -0,0 +1,67 @@
...
+		level.AllowDebug()
+		level.Debug(logger).Log("msg", "debug logging enabled")
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	if withTimeout > 0 {
+		level.Debug(logger).Log("msg", "starting with timeout", "duration", withTimeout.String())
+		ctx, cancel = context.WithTimeout(context.Background(), withTimeout)
+	}
+	defer cancel()
```
rfratto, line 46: This will leak the old cancel context when timeout is set, try this:

```go
ctx := context.Background() 
if withTimeout > 0 {
  var cancel context.CancelFunc 
  ctx, cancel = context.WithTimeout(ctx, withTimeout) 
  defer cancel()  
}
```

cmd/agent-smoke/main.go
```diff
@@ -0,0 +1,67 @@
...
+		smoke.WithMutationFrequency(mutationFrequency),
+	}
+	if kubeconfig != "" {
+		opts = append(opts, smoke.WithKubeConfig(kubeconfig))
+	}
+
+	smokeTest, err := smoke.NewSmokeTest(opts...)
+	if err != nil {
+		level.Error(logger).Log("msg", "error constructing smoke test", "err", err)
+		os.Exit(-1)
```
rfratto, line 61: very petty nit: `os.Exit(1)` here and below

cmd/agent-smoke/internal/smoke.go
```diff
@@ -0,0 +1,158 @@
...
+	kubeconfig string
+	namespace  string
+	tasks      []Task
+
+	chaosFrequency    time.Duration
+	mutationFrequency time.Duration
+}
+
+// NewSmokeTest is the constructor for a Smoke object.
+func NewSmokeTest(opts ...Option) (*Smoke, error) {
```
rfratto, line 27: stylistic suggestion, I don't care if you actually make this change: 

I would personally cut down on boilerplate until there's a reason it can't be avoided. I think the options pattern is neat and useful if your options need to do logic, but right now you don't need that and it's bubbling throughout the code. 

As an alternative, if you had an options strut:

```go
type Options struct {
  Kubeconfig string 
  Namespace string 
  MutationFrequency, ChaosFrequency time.Duration 
}
```

then you could avoid the boilerplate to define/call the options and even bind flags to field members directly in the main package:

```go 
package main 

func main() {
  var (
     ll logging.Level 
     o Options 
  )
  
  ll.RegisterFlags(flag.CommandLine) 
  flag.StringVar(&o.Namespace, "namespace", "agent-smoke-test", "namespace smoke test should run in")
  flag.StringVar(&o.Kubeconfig, "kubeconfig", "", "absolute path to the kubeconfig file")
  // etc 
  
  logger := util.NewLogger(ll) 
  st, err := smoke.NewSmokeTest(logger, o) 
  if err != nil {
     // ...
  } 
}
```

cmd/agent-smoke/internal/smoke.go
```diff
@@ -0,0 +1,158 @@
...
+
+// Run starts the smoke test and runs the tasks in an errgroup Group.
+func (s *Smoke) Run(ctx context.Context) error {
+	g, ctx := errgroup.WithContext(ctx)
+	taskFn := func(t Task) func() error {
+		fn, freq := t.Task()
+		return func() error {
+			for {
+				select {
+				case <-time.After(freq):
```
rfratto, line 80: nit: I'd just use a ticker since it will be a little more efficient vs creating many timers:

```go
tick := time.NewTicker(freq) 
defer tick.Stop() 

for {
  select {
  case <-tick.C: 
     // call fn 
  case <-ctx.Done():
     return nil 
  }
}
```

cmd/agent-smoke/internal/smoke.go
```diff
@@ -0,0 +1,158 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return s, nil
+}
+
+// Run starts the smoke test and runs the tasks in an errgroup Group.
+func (s *Smoke) Run(ctx context.Context) error {
+	g, ctx := errgroup.WithContext(ctx)
```
rfratto, line 74: nit: we already use `github.com/oklog/run` in other packages for task runners, we should re-use it here to be consistent.

cmd/agent-smoke/internal/smoke.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+	}
+	for _, task := range s.tasks {
+		g.Go(taskFn(task))
+	}
+	return g.Wait()
+}
+
+func (s *Smoke) logDebug(keyvals ...interface{}) {
+	if s.logger != nil {
```
rfratto, line 97: nit: instead of checking every time, make sure the logger is never nil when `NewSmokeTest` is called:

```go 
if s.logger == nil {
   s.logger = log.NewNopLogger()
}
```

cmd/agent-smoke/internal/tasks.go
```diff
@@ -0,0 +1,71 @@
...
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// The Task interface represents units of work performed by some function on
+// some frequency interval by the smoke framework.
+type Task interface {
+	Task() (func(context.Context, *Smoke) error, time.Duration)
+}
```
rfratto, line 16: The nested functions and return arguments read a little awkwardly here. WDYT about:

```go
type Task interface {
  Frequency() time.Duration 
  Run(context.Context) error 
}

type deletePodTask struct {
  clientset *kubernetes.Clientset

  namespace string 
  pod string 
  duration time.Duration 
}

func (t *deletePodTask) Frequency() time.Duration { return t.duration } 
func (t *deletePodTask) Run(ctx context.Context) error {
  var zero int64 
  err := t.clientset.CoreV1().Pods(t.namespace).Delete(ctx, t.pod, metav1.DeleteOptions{
    GracePeriod: &zero, 
  })
  if err != nil {
    // caller can log the error and decide whether to abort the task 
    // or continue scheduling it  
    return fmt.Errorf("failed to delete pod %q: %w", t.pod, err) 
  }
  return nil
}
```

rfratto: Alternatively, it might not make sense for the Task to store its frequency at all if it's only purpose is to be returned back to the parent which used to know the value in the first place. 

You could potentially track tasks with their frequency instead:

```go
type Smoke struct {
  tasks []scheduledTask 
}

type scheduledTask struct {
  Task 
  Frequency time.Duration 
}

func NewSmokeTest(/* args */) (*Smoke, error) {
  var s Smoke 
  // init, whatever
 
  s.tasks = append(
     s.tasks, 
     repeatingTask{
        Frequency: /* something */, 
        Task: deletePodTask{namespace: namespace, pod: pod},
     },
  )

  return &s, nil 
}
```

Makefile
```diff
@@ -198,6 +199,10 @@ cmd/grafana-agent-crow/grafana-agent-crow: cmd/grafana-agent-crow/main.go
 	$(ALL_CGO_BUILD_FLAGS) ; $(seego) build $(CGO_FLAGS) -o $@ ./$(@D)
 	$(NETGO_CHECK)
 
+cmd/agent-smoke/agent-smoke: seego cmd/agent-smoke/main.go
```
rfratto, line 202: nit: we should start prefixing binaries `grafana-agent-` instead of just `agent-` (see #582)

rfratto, line 202: In general, I also think it was a mistake for me to put crow in `cmd/`, and I think both crow and this should probably be in tools instead. These are things we wrote for development, and don't want to put them in the folder that signals "please use me"

Makefile
```diff
@@ -206,6 +211,8 @@ agent-operator-image:
 	$(docker-build)  -t $(IMAGE_PREFIX)/agent-operator:latest -t $(IMAGE_PREFIX)/agent-operator:$(IMAGE_TAG) -f cmd/agent-operator/$(DOCKERFILE) .
 grafana-agent-crow-image:
 	$(docker-build)  -t $(IMAGE_PREFIX)/agent-crow:latest -t $(IMAGE_PREFIX)/agent-crow:$(IMAGE_TAG) -f cmd/grafana-agent-crow/$(DOCKERFILE) .
+agent-smoke-image:
+	$(docker-build)  -t $(IMAGE_PREFIX)/agent-smoke:latest -t $(IMAGE_PREFIX)/agent-smoke:$(IMAGE_TAG) -f cmd/agent-smoke/$(DOCKERFILE) .
```
rfratto, line 215: The growing number of agent-related images on dockerhub are giving me some pause 😞 Can we consider pushing both this and crow to our internal `us.gcr.io/kubernetes-dev` registry instead? 

(They also don't need to support seego)

rfratto (approved):
nice!

cmd/agent-smoke/main.go
rlankfo: Thanks for pointing this out, this makes the logging set up easier and requires less boiler plate. I didn't want to change the signature of our existing `NewLogger` API (although I imagine it probably isn't called in too many places), but opted to add a new `NewLoggerFromLevel` API. This allows me to take advantage of this existing helper package w/o having to change any other callers of this package. Let me know what you think.

cmd/agent-smoke/main.go
rlankfo: thanks, did this for logging.Format as well :+1:

cmd/agent-smoke/main.go
rlankfo: fixed :+1:

cmd/agent-smoke/main.go
rlankfo: https://github.com/grafana/agent/pull/1225/commits/d05a28860042c0a38fca41b5c6e03778cf96f40b

cmd/agent-smoke/internal/smoke.go
rlankfo: I initially thought some of the options might require some logic when I started coding on this but that didn't end up being the case. I got rid of the functional options and the code is definitely less clunky.

cmd/agent-smoke/internal/smoke.go
rlankfo: updated :+1:

cmd/agent-smoke/internal/tasks.go
rlankfo: I think the repeatingTask struct makes the most sense here, made this update.

cmd/agent-smoke/internal/smoke.go
```diff
@@ -0,0 +1,105 @@
...
+				logger:    s.logger,
+				clientset: clientset,
+				namespace: opts.Namespace,
+				pod:       "grafana-agent-0",
+			},
+			frequency: opts.ChaosFrequency,
+		},
+		repeatingTask{
+			Task: &scaleDeploymentTask{
+				logger:      s.logger,
```
rfratto, line 65: very light suggestion: you can save yourself on typing here by making this 

```go
logger: log.With(s.logger, "task", "scale_deployment", "deployment", "avalanche") 
```

then you would be able to avoid all the formatting of the log messages and have something more generic: 

```go
// logs as ts=<time> level=error task=scale_deployment deployment=avalanche msg="failed to get autoscalingv1.Scale" err=<error> 
level.Error(t.logger).Log("msg", "failed to get autoscalingv1.Scale", "err", err)

```

rfratto (commented):
nice, much easier to follow, good work 👍

cmd/agent-smoke/internal/smoke.go
rlankfo: nice :+1: this is clean

tools/smoke/internal/smoke.go
```diff
@@ -0,0 +1,105 @@
...
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return nil, err
+	}
+
+	// add default tasks
+	// TODO: need to add a deletePodTask for cluster pods,
+	// TODO: currently script generates random number and appends it
+	// TODO: to the pod name. Should do this with a label selector,
+	// TODO: by adding a deletePodByLabelTask or such
```
rlankfo, line 52: I plant to knock out this todo after I get stuff up and running in dev cluster

production/tanka/grafana-agent/smoke/main.libsonnet
```diff
@@ -0,0 +1,15 @@
+local smoke = import './internal/smoke.libsonnet';
+
+{
+    _images:: {
+        agentsmoke: 'us.gcr.io/kubernetes-dev/grafana/agent-smoke:latest',
+    },
+
+    new(name='grafana-agent-smoke', namespace='grafana-agent-smoke', image=self._images.agentsmoke):: {
+        local this = self,
```
rlankfo, line 9: probably don't need this, honestly this whole file seems a bit unnecessary if I just moved smoke.libsonnet to main.libsonnet

production/tanka/grafana-agent/smoke/internal/smoke.libsonnet
```diff
@@ -0,0 +1,35 @@
+local k = import 'ksonnet-util/kausal.libsonnet';
+
+local policyRule = k.rbac.v1.policyRule;
+local serviceAccount = k.core.v1.serviceAccount;
+local container = k.core.v1.container;
+local deployment = k.apps.v1.deployment;
+
+{
+    newSmoke(name='grafana-agent-smoke', namespace='default', image):: {
```
rlankfo, line 9: need to pass additional arguments here like duration, frequency, etc.

example/k3d/smoke/main.jsonnet
```diff
@@ -23,6 +25,9 @@ local new_crow(name, selector) =
     },
   });
 
+local new_smoke(name) = smoke.new(name, namespace='smoke', image=images.agentsmoke);
```
rlankfo, line 28: need to pass `duration` here for local smoke test

example/k3d/scripts/smoke-test.bash
```diff
@@ -82,33 +83,30 @@ run() {
   if [[ ! -z "$IMPORT_IMAGES" ]]; then
     echo "--- Importing local images"
 
+    docker pull us.gcr.io/kubernetes-dev/grafana/agent-smoke:latest
     k3d image import -c $K3D_CLUSTER_NAME \
       grafana/agent:latest \
       grafana/agentctl:latest \
-      grafana/agent-crow:latest
+      grafana/agent-crow:latest \
```
rlankfo, line 90: I'll update this to pull crow from gcr now as well

example/k3d/scripts/smoke-test.bash
```diff
@@ -37,6 +37,7 @@ ROOT=$(git rev-parse --show-toplevel)
 K3D_CLUSTER_NAME="agent-smoke-test"
 MUTATION_FREQUENCY="5m"
 CHAOS_FREQUENCY="30m"
```
rlankfo, line 39: these end up being unused at this point, need to figure out how to pass these values during tk apply if that's possible

production/tanka/grafana-agent/smoke/internal/smoke.libsonnet
```diff
@@ -0,0 +1,35 @@
...
+            'log.level': 'debug',
+            'namespace': namespace,
+          })),
+
+        agentsmoke:
+          deployment.new(name, 1, [self.container]) +
+          deployment.mixin.metadata.withNamespace(namespace) +
+          deployment.mixin.spec.template.spec.withServiceAccount(name),
+    },
+}
```
rlankfo, line 35:
```suggestion
}
```

example/k3d/scripts/smoke-test.bash
rfratto: Yep: `tk apply <path> -V key1=value1 -V key2=value2`

Then in jsonnet: `std.extVar("key1")`

example/k3d/lib/crow/main.libsonnet
```diff
@@ -14,7 +14,7 @@ local volume = k.core.v1.volume;
     local this = self,
```
rlankfo, line 14: I'll need crow in deployment_tools so it might make sense to move this to our grafana-agent-mixin? I could also just duplicate this file in deployment_tools for now as well.

example/k3d/scripts/smoke-test.bash
```diff
@@ -82,20 +83,16 @@ run() {
   if [[ ! -z "$IMPORT_IMAGES" ]]; then
     echo "--- Importing local images"
 
+    docker pull us.gcr.io/kubernetes-dev/grafana/agent-smoke:latest
```
rfratto, line 86: IMO you should be building this locally or pulling it outside of the smoke script instead of having it pull from gcr every time.

example/k3d/smoke/main.jsonnet
```diff
@@ -23,6 +25,11 @@ local new_crow(name, selector) =
     },
   });
 
+local chaosFrequency = std.extVar("chaos_frequency");
+local mutationFrequency = std.extVar("mutation_frequency");
```
rfratto, line 29: Can we give these a default if they're not defined? I do local applies a lot when playing around with the smoke environment.

rfratto (approved):

example/k3d/smoke/main.jsonnet
rlankfo: @rfratto From what I can tell, `std.extVar` will return an error if this value is not passed and I'm not sure how to handle that error. Is there a pattern I can use here for assigning default values to external vars?

rlankfo: If we're ok modifying these values in the main.jsonnet, I could remove them from the smoke script and not pass them in as external vars. Let me know what you think.

example/k3d/scripts/smoke-test.bash
rlankfo: I agree -- I'll remove this so the end user can decide if they want to build it locally, pull from the registry, etc.

PR #1234, opened 2022-01-07
Create an e2e framework with support for running tests against k8s

rfratto:
#### PR Description 
This PR introduces an end-to-end testing framework that includes support for launching a Kubernetes cluster through k3d.

#### Notes to the Reviewer
I've defined `e2e` as its own module because it brings in a ton of new imports we otherwise don't really need. We'll likely want to eventually import agent/operator code in `e2e`, which unfortunately means we'll need to manually sync the replace directives. There might be a tool we can write for automating that later.

Also, while I've included a sample test that brings up a cluster, it's not likely we'll want to create a cluster per test (it's slow). Rather, we'll probably want to define `TestMain` in each package that needs a cluster to run tests.
 
#### PR Checklist

- [x] CHANGELOG updated (N/A) 
- [x] Tests updated


rfratto:
I'm guessing the CI is going to fail here. I think I'll create a new e2e-test Drone pipeline.

rfratto:
Oh 😞 This is exposing the k3d cluster on the host machine, not the docker container running the test. I'm not super sure what to do about that, honestly. I'll play around with something.

rfratto:
@mattdurham @rlankfo this is actually ready for review now 🙃

e2e/k8s/k8s.go
```diff
@@ -0,0 +1,330 @@
...
+		for _, nw := range lbInfo.NetworkSettings.Networks {
+			httpAddr = fmt.Sprintf("%s:80", nw.IPAddress)
+			break
+		}
+	}
+
+	serverInfo, err := cli.ContainerInspect(ctx, server.Name)
+	if err != nil {
+		return "", "", fmt.Errorf("failed to inspect loadbalancer: %w", err)
+	} else if len(serverInfo.NetworkSettings.Networks) == 0 {
```
rlankfo, line 238: nit: unnecessary else

e2e/k8s/k8s.go
```diff
@@ -0,0 +1,330 @@
...
+		return "", "", fmt.Errorf("no API server address from worker")
+	} else {
```
rlankfo, line 240: nit: unnecessary else

e2e/k8s/k8s.go
```diff
@@ -0,0 +1,330 @@
...
+		for _, nw := range serverInfo.NetworkSettings.Networks {
+			serverAddr = fmt.Sprintf("%s:6443", nw.IPAddress)
+			break
+		}
```
rlankfo, line 244: Will this cause an issue if there is more than one network? Would it be better to look up the network endpoint settings in this map by name? Since the iteration order of a map isn't guaranteed, it seems like this could possibly cause some intermittent issues that are hard to debug when several docker networks exist. Same for the lbInfo loop above.

rlankfo (approved):
LGTM overall but did have question on the docker network setup :+1:

PR #1235, opened 2022-01-07
integrations-next: fix bug where v2 integrations were not being strictly unmarshaled

rfratto:
(No changelog entry because integrations-next isn't released yet)
(This was only a problem for v2 integrations, the v1 shims were fine)


mattdurham (approved):
LGTM

Issue #1227, opened 2022-01-04
Grafana Operator does not allow args modification to Grafana Agent deployments


shamsalmon:
Grafana Operator does not expose the ability to add additional args, such as `config.enable-read-api`. It would be great if an additional option was added to the Grafana Agent CRD to expose the ability to modify EnableConfigEndpoints and set it to true or false.

shamsalmon:
From a quick glance, this seems to be the problem: https://github.com/grafana/agent/blob/a8aa6b411685db524ae4b07fc2aa4ff82f08838e/pkg/operator/resources_metrics.go#L228. These args are static.

rfratto:
Closed by #1248

PR #1248, opened 2022-01-10
Add option for Operator to pass arguments to GrafanaAgent #1227

shamsalmon:
#### PR Description 

Adds Args CRD to agent. Mostly for new config read endpoint options (disabled by default).

https://github.com/grafana/agent/issues/1227

#### Notes to the Reviewer

Can be enabled like so:
```
kind: GrafanaAgent
metadata:
  name: grafana-agent
  namespace: operations
  labels:
    app: grafana-agent
spec:
  image: grafana/agent:v0.21.2
  logLevel: info
  serviceAccountName: grafana-agent
  args:
    - --config.enable-read-api
  metrics:
    instanceSelector:
      matchLabels:
        agent: grafana-agent-metrics
    externalLabels:
      cluster: devqa
```

#### PR Checklist

- [x] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated


shamsalmon:
Also https://github.com/grafana/helm-charts/pull/951 for helm support for this new feature

pkg/operator/apis/monitoring/v1alpha1/types.go
```diff
@@ -147,6 +147,9 @@ type GrafanaAgentSpec struct {
 	// Logs controls the logging subsystem of the Agent and settings unique to
 	// logging-specific pods that are deployed.
 	Logs LogsSubsystemSpec `json:"logs,omitempty"`
+
+	// Args controls command line additional arguments sent to the Agent pod.
+	Args []string `json:"args,omitempty"`
```
rlankfo, line 152:
```suggestion
	// +kubebuilder:default={"-config.file=/var/lib/grafana-agent/config/agent.yml","-config.expand-env=true","-reload-port=8081"}
	Args []string `json:"args,omitempty"`
```

rlankfo (approved):
LGTM :+1: but I would suggest using the default marker to move the arguments still defined in code to the default value for args as demonstrated in review. This will hopefully prevent confusion if a user wants to use this args field to set any of the values that are still currently hardcoded. 

This should allow us to get rid of `agentArgs` slice and pass `d.Agent.Spec.Args` on line 403. Note you'll need to re-run `make crds`.

production/operator/crds/monitoring.grafana.com_grafanaagents.yaml
```diff
@@ -1054,6 +1054,16 @@ spec:
...
+                - -config.file=/var/lib/grafana-agent/config/agent.yml
+                - -config.expand-env=true
+                - -reload-port=8081
+                description: Args controls command line additional arguments sent
+                  to the Agent pod.
+                items:
+                  type: string
+                type: array
               configMaps:
                 description: ConfigMaps is a liset of config maps in the same namespace
```
rlankfo, line 1068:
```suggestion
                description: ConfigMaps is a list of config maps in the same namespace
```

This is unrelated to your change but I just noticed this typo

rfratto:
Thanks for the PR @shamsalmon!

I'm a little concerned about the idea of generally exposing the container args: 

* We might not want to allow full-override of the arguments, since the reload port *must* always be on 8081, otherwise things will silently break.
* On the other hand, we might not want to allow appending to the arguments, since providing an already-defined argument would also cause things to break (though it would break more noisily with container restart loops). 

I also _think_ it's already possible to use the `containers` field of the CRD to fully override the container arguments. That field was copied from the Prometheus Operator, and it's documented as explicitly an advanced option that has no backwards compatibility guarantees between releases. 

My gut feeling is:

1. This new option should append to arguments rather than be a full replace (to supplement the existing merge functionality), and it should be similarly documented as an advanced option with no backwards compatibility guarantees.
2. There should also be a new boolean option in the CRD specifically for enabling `enableConfigReadAPI` which will automatically provide the argument.

WDYT @rlankfo?

rfratto:
> I'm fine with adding a new option in the CRD for a specific flag, but that sets a precedent of needing to do this for every flag we may want to expose which would be more maintenance.

Right, I at least think enabling the read API is something should explicitly expose, but I agree we can't be expected to have a CRD option for literally every flag we might have. The operator is intended to be a Kubernetes-tailored experience, which isn't necessarily conducive to have a 1:1 mapping of flags.

rlankfo:
@rfratto I'll spend some time to verify if our args can be overridden using the `containers` field or if this just gets merged with the existing args we have hardcoded.

In any case, it sounds like the path forward here is probably just to enable the boolean for `enableConfigReadAPI`. I could see us wanting to use remote config here (one reason I wanted to be able to add args for experimental flags, and for modifying -config.file) but this could also be an explicit field.

Personally I think we should avoid appending user defined args with hardcoded args but I'm fine if that ends up being the direction we want to take.

Let me know what you think.

rfratto:
I agree that only adding an extra option for enabling the config read API is the least contentious path for now. We can revisit a general option for appending new options later if this comes up again, but we aren't adding many options that often.

@shamsalmon sorry for the back and forth, but would you mind changing your PR so that the only new field to the CRD is for explicitly enabling the config read API?

shamsalmon:
> I agree that only adding an extra option for enabling the config read API is the least contentious path for now. We can revisit a general option for appending new options later if this comes up again, but we aren't adding many options that often.
> 
> @shamsalmon sorry for the back and forth, but would you mind changing your PR so that the only new field to the CRD is for explicitly enabling the config read API?

will do!

shamsalmon:
> Hey @shamsalmon, I just wanted to check in on this PR. Is there anything we can do to help out on this? Thanks.

Sorry for the delay. I got busy with work and forgot about it! I will get the change pushed out today / tomorrow.

shamsalmon:
Just wrapped up my changes and tested them out. Let me know if you have any other suggestions. I tested with the param enabled, with it disabled, and with it missing and it worked as expected.

rfratto (approved):
Hey @shamsalmon, sorry for the delay! LGTM, thanks for contributing!

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ This release has deprecations. Please read [DEPRECATION] entries and consult
 the [upgrade guide](https://github.com/grafana/agent/blob/main/docs/upgrade-guide/_index.md)
 for detailed information.
 
+- [FEATURE] Added argument support to GrafanaAgent Custom Resource Definition.
```
rfratto, line 7: Can you rebase on top of current main and move this to the unreleased section?

rfratto (commented):
Before we merge, can you sign the CLA?

shamsalmon:
> Before we merge, can you sign the CLA?

I signed the CLA but for some reason its still showing as not signed. Any ideas?

rfratto:
> I signed the CLA but for some reason its still showing as not signed. Any ideas?

It looks like the CLA bot is thrown off by the committer being sshannon, which doesn't match your GitHub username (& there's no email address to fall back on)

rfratto:
If it's too much work, I can re-open this PR under my account and give you authorship credit :)

shamsalmon:
> If it's too much work, I can re-open this PR under my account and give you authorship credit :)

Looks like I have to rewrite history to change my username etc. Is that ok or would you rather me open a new PR? Also ok if you would like to do it.

CHANGELOG.md
shamsalmon: done

rfratto:
> Looks like I have to rewrite history to change my username etc. Is that ok or would you rather me open a new PR? Also ok if you would like to do it.
>
> Edit - Tried rewiritng history - looks like it worked. Sorry I somehow seem to have duplicated a bunch of writes.

rewriting history is fine 👍 I set this to automerge once the CI goes green. Thanks again for the contribution!

PR #1031, opened 2021-10-26
Integrations: frontend receiver

domasx2:
#### PR Description 

This PR adds a ["frontend receiver" integration](https://github.com/grafana/grafana-frontend-telemetry-receiver) which adds an endpoint that can receive payloads from the [grafana frontend agent](https://github.com/grafana/grafana-frontend-agent).

To achieve this I had to extend internal integration API to:
* provide logs and tempo instances to the integration constructor 
* support arbitrary http endpoints instead of just `/metrics`

TODOS, other than checklist below:
- [ ] add metrics handler
- [ ] forward traces to tempo

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

PR #1253, opened 2022-01-11
App o11y exporter

kpelelis:
NOTE: This integration is implemented using the v2 Integration interface

This commit is the first draft of integrating the [App Observability Telemetry Receiver][1] into the agent. For now the following components are supported:

### Logs

The exporter forwards the logs received onto the HTTP endpoint into the logs subsystem.

### Exceptions

Same as logs, but with fixed labels.

### Metrics

The integration is capable of reporting both some predetermined performance metrics (web vitals in this iteration) but also some user defined ones using a specific schema in the configuration file. Internally, a Prometheus Summary type is used to keep track of the incoming values and is also using the self scrape system to report the metrics.

The exporter itself is basically a modular HTTP endpoint that uses different exporters to forward received data into the appropriate subsystems of the agent.

[1]: https://github.com/grafana/grafana-frontend-telemetry-receiver

- [X] CHANGELOG updated 
- [X] Documentation added
- [X] Tests updated


docs/configuration/integrations/integrations-next/app-o11y-exporter.md
```diff
@@ -0,0 +1,58 @@
...
+The `app_o11y_config` block configures the `app-o11y-exporter`
+integration, which is a data exporter for logs/metrics/traces/exceptions
+received by the respectful JavaScript agent
+
+Full reference of options:
+
+```yaml
+  # Enables the app_o11y_config integration, allowing the Agent to automatically
+  # collect metrics,logs and exceptions from the grafana frontend agent
+  [enabled: <boolean> | default = false]
```
rfratto, line 16: We've moved away from integrations-next integrations having an enabled field. Now, we treat them as enabled if they're defined in the YAML.

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ This release has deprecations. Please read [DEPRECATION] entries and consult
 the [upgrade guide](https://github.com/grafana/agent/blob/main/docs/upgrade-guide/_index.md)
 for detailed information.
 
+- [FEATURE] (beta) Add app o11y integration. (@kpelelis)
```
rfratto, line 7: Can you add a note that this depends on integrations-next being enabled?

docs/configuration/integrations/integrations-next/app-o11y-exporter.md
```diff
@@ -0,0 +1,58 @@
...
+
+# app_o11y_config
+
+The `app_o11y_config` block configures the `app-o11y-exporter`
+integration, which is a data exporter for logs/metrics/traces/exceptions
+received by the respectful JavaScript agent
+
+Full reference of options:
+
+```yaml
```
rfratto, line 13: It seems like the metrics autoscrape/etc fields are missing here

pkg/integrations/v2/app_o11y_exporter/app_o11y_exporter.go
```diff
@@ -0,0 +1,206 @@
...
+		host = "0.0.0.0"
+		port = 3029
+	)
+	if ai.conf.Server.Host != "" {
+		host = ai.conf.Server.Host
+	}
+
+	if ai.conf.Server.Port != 0 {
+		port = ai.conf.Server.Port
+	}
```
rfratto, line 185: IMO defaults should be applied when the config is unmarshaled rather than at start time

pkg/integrations/v2/app_o11y_exporter/app_o11y_exporter.go
```diff
@@ -0,0 +1,206 @@
...
+
+	if err != nil {
+		return err
+	}
+	err = http.Serve(ln, r)
+	if err != nil {
+		return err
+	}
+
+	<-ctx.Done()
```
rfratto, line 200: If `ctx` gets canceled, the integration needs to shut down. I don't think that's currently happening here.

rfratto (commented):
Gave this a quick first pass, some small comments

CHANGELOG.md
kpelelis: 82ddd252

docs/configuration/integrations/integrations-next/app-o11y-exporter.md
kpelelis: a4a5f090

pkg/integrations/v2/app_o11y_exporter/app_o11y_exporter.go
kpelelis: 70998f9d

pkg/integrations/v2/app_o11y_exporter/app_o11y_exporter.go
kpelelis: a6dfcf0c

kpelelis:
I am working on the linting errors atm

domasx2 (changes requested):
Great job going forward with this :) 

I have some high level notes:
* I believe current sourcemap implementation is not exactly in the right direction. To make it usable, I think we should support two modes: 
   a) if opted-in: download the source file, exctract embedded sourceamp url, download the sourcemap, cache it
   b) allow specifying filesystem dirs where sourcemaps are to be found, per  **http origin** and **release**, assuming path is same as path of the source in the URL + `.map` 
I suggest we drop sourceamp support from this PR and make a separate one later. I'll create an issue with details

* When logging to loki, instead of logging just the stacktrace or just the log message we should log a logfmt line that also includes meta, see previous PR. Just a message with no context is not super useful :) https://github.com/grafana/agent/pull/1031/files#diff-35f3193cfb79b2ce91b8fcfc4427168437c8bd8fd4fc74b6bc2d7e84e46d5dbaR98.

* Should be possible to configure static labels or derive labels from meta when logging to loki, this was also in the previous PR 

* I have doubts about collecting browser metrics to prometheus. Either we provide no useful labels, or cardinality will be an issue. Would rather for first version log metrics to loki along with all metadata, use logql to derive metrics. I feel this will be a better fit and more flexible for now. Design doc on this https://docs.google.com/document/d/1rGo6fn8o2qk_m_shXPi6pwClEFfAmxCRp6aNmM7JKkU/edit#heading=h.5sybau7waq2q

pkg/integrations/v2/app_o11y_exporter/app_o11y_exporter.go
```diff
@@ -0,0 +1,212 @@
...
+		Config:   cfg,
+	}}
+}
+
+// RunIntegration implements Integratin
+func (i *appo11yIntegration) RunIntegration(ctx context.Context) error {
+	r := mux.NewRouter()
+	r.Handle("/collect", i.receiver.ReceiverHandler(&i.logger))
+
+	srv := &http.Server{
```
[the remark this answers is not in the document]
kpelelis: I believe the HTTP server of the agent is intended to be use as a prometheus endpoint.
I am concerned with [this](https://github.com/grafana/agent/blob/main/pkg/integrations/v2/integrations.go#L144-L145) comment which states that the URL is not guaranteed to be predictable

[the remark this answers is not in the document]
kpelelis: > I suggest we drop sourceamp support from this PR and make a separate one later. I'll create an issue with details

I am onboard with making the implementation as thin as possible.

docs/configuration/integrations/integrations-next/app-o11y-exporter.md
```diff
@@ -0,0 +1,80 @@
...
+
+  # User defined prometheus metrics to be scraped. The sending end of the agent can
+  # update this metrics using the specified payload. Since these are performance metrics
+  # a Summary type metric is used internally for each metric defined
+  custom_measurements:
+    [name: <string>]
+    [description: <string>]
+
+  # Extra loki lables to append in each Loki Entry (exceptions logs and measurements)
+  extra_loki_lablels:
```
domasx2, line 73:
```suggestion
  extra_logs_labels:
```
I believe the convention is to call it "logs" rather than "loki" in agent configs. Not everything loki-based is "loki", eg "logs enterprise"

pkg/integrations/v2/app_o11y_exporter/config/config.go
```diff
@@ -0,0 +1,80 @@
...
+// AppExporterConfig is the configuration struct of the
+// integration
+type AppExporterConfig struct {
+	CORSAllowedOrigins    []string           `yaml:"cors_allowed_origins,omitempty"`
+	RateLimiting          RateLimitingConfig `yaml:"rate_limiting,omitempty"`
+	MaxAllowedPayloadSize int64              `yaml:"max_allowed_payload_size,omitempty"`
+	Server                ServerConfig       `yaml:"server,omitempty"`
+	LogsInstance          string             `yaml:"logs_instance"`
+	Measurements          []Measurement      `yaml:"custom_measurements"`
+	ExtraLokiLabels       map[string]string  `yaml:"extra_loki_lablels"`
```
domasx2, line 62:
```suggestion
	ExtraLogsLabels       map[string]string  `yaml:"extra_loki_lablels"`
```

pkg/integrations/v2/app_o11y_exporter/exporters/loki.go
```diff
@@ -0,0 +1,113 @@
...
+			Timestamp: time.Now(),
+			Line:      string(line),
+		},
+	}, time.Duration(le.seTimeout))
+	if !sent {
+		level.Warn(le.logger).Log("msg", "failed to log frontend log event to loki pipeline")
+	}
+}
+
+func (le *LokiExporter) labelSet(kv *utils.KeyVal) prommodel.LabelSet {
```
domasx2, line 98: this is not correct: it shouldn't set all of the key value pairs as labels for the loki message. It would result in high cardinality and unnecessarily impact loki performance. User needs to explicitly configure what should be set as labels.

So, it should only set keys that are configured in `config.logsLabels`. If value is provided for a label, this value should be set directly. If value is empty string, it should be taken from entry in `kv` matching the label name. See the previous impl here: https://github.com/grafana/agent/pull/1031/files#diff-0805f060c9a440ed60d107f64f4adb402891b547a6300c3f58d6e80356c79f81

domasx2 (changes requested):

pkg/integrations/v2/app_o11y_exporter/exporters/exporter.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+// AppReceiverExporter is the base interface
+type AppReceiverExporter interface {
+	Init() error
+}
+
+// AppDataExporter  an interface for exporters
+// that are forwarding data to a different service
+type AppDataExporter interface {
```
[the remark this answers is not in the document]
domasx2: functionally they seem to be exactly the same though, unless I'm missing something. We can just have a single kind of exporter interface, it's up to the implementation of what it wants to do with the payload

domasx2:
closing, reopened from source branch https://github.com/grafana/agent/pull/1301 so both me and Kostas can work on it

PR #1254, opened 2022-01-11
Operator: use new k8s lib for tests

### LINKED ISSUE #1251: Invert condition used to exempt network tests
Today, we have tests which only execute when `-tags has_network` is provided: https://github.com/grafana/agent/blob/727f9089b3c67dea416353b9e24062a85c6f3c3f/pkg/operator/kubelet_test.go#L3

This is useful for running tests in a sandbox environment where things like the network aren't available. We initially started doing this after @flokli's #333, which needs to skip over tests for NixOS. 

However, adding the extra tag makes it more difficult to write the tests, as it is excluded from build by default, and therefore excluded from gopls by default. 

I'd suggest that we invert the condition instead:

```
//go:build !nonetwork
```

```
//go:build !nodocker 
```

(etc)

This makes it easier to add to the tests, and ensures that the default testing behavior is to run everything.

rfratto:
#### PR Description 
This PR is a follow-up to #1234. After playing around with some example e2e tests, it became clear that there was more testability power by doing things one level lower at the integration test level. 

This PR does some number of small things: 

* Removes `e2e/` and moves the k8s package to `pkg/util/k8s` so other packages can use it. 

* Changes the existing operator tests to use `pkg/util/k8s` instead of downloading the controller-runtime testing tools. As a benefit, you can now run these tests locally on Windows! (as long as you have Docker)

* Stops the macOS/Windows CIs from running the Operator tests. There's no platform specific behavior to test there, plus they don't have Docker installed. 

* Inverts exempting network tests as suggested in #1251. 

Closes #1251 

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated


rfratto:
Superseded by #1268.

PR #1259, opened 2022-01-12
Use new `metrics:` key in example config

### LINKED ISSUE #1255: Agent cluster fails to start up on k3d example due to invalid config
When running the k3d example, grafana-agent-cluster fails to start due to a configuration error as the provided config contains both a `prometheus:` and a `metrics:` key.

## Issue
```
$ kg pods
NAME                                    READY   STATUS             RESTARTS   AGE
...
grafana-agent-cluster-ffd99fc85-9h4z5   0/1     CrashLoopBackOff   8          17m

$ k logs grafana-agent-cluster-ffd99fc85-9h4z5
2022-01-12 11:26:22.338671 I | error loading config file /etc/agent/agent.yml: at most one of prometheus and metrics should be specified
```

Looking at the deployed `configMap` validates that this is indeed the issue.
```
$ k describe configmap grafana-agent-cluster
Name:         grafana-agent-cluster
...
agent.yml:
----
metrics:
...
prometheus:
...
```

## Cause
A possible cause could be the following:

In k3d's environment/main.jsonette, we indirectly use [grafana-agent/config.libsonnet](https://github.com/grafana/agent/blob/41cf32cc4051647854815daa079ce23af1615e38/production/tanka/grafana-agent/config.libsonnet#L56-L81) which already contains the `prometheus:` key.

Next, in `scraping-svc` we import this config, add a few attributes and load it up as a configMap.

```
local config = import '../config.libsonnet';

    _config+:: {
...
      agent_config+: {
        prometheus+: {
          // No configs are used in the scraping service mode.
          configs:: [],

          scraping_service: {
            enabled: true,
            kvstore: this._config.agent_config_kvstore,
            lifecycler: {
              ring: {
                kvstore: this._config.agent_ring_kvstore,
              },
            },
          },
        },
      },
    },

      agent_config+: {
        prometheus+: {
          // No configs are used in the scraping service mode.
          configs:: [],

          scraping_service: {
            enabled: true,
            kvstore: this._config.agent_config_kvstore,
            lifecycler: {
              ring: {
                kvstore: this._config.agent_ring_kvstore,
              },
            },
          },
        },
      },

...

    configMap:
      configMap.new(this._config.agent_configmap_name) +
      configMap.withData({
        'agent.yml': k.util.manifestYaml(this._config.agent_config),
      }),
```

While later on, the k3d example's main.jsonnet proceeds to _also_ add the `metrics:` key
```
      agent_config+: {
        metrics+: {
          global+: {
            external_labels+: {
              cluster: cluster_label,
            },
          },

          scraping_service+: {
            dangerous_allow_reading_files: true,
          },
        },
      },
```

## Solution
Right now, I don't know enough about the agent's config to say whether/how we can merge the two configuration keys, but removing the additional `metrics:` does indeed allow `grafana-agent-cluster` to come to life.

I'll look more into it and come back with a proper solution.

## Misc
As a side note, it was somewhat confusing context switching between `agent.yml` and `agent.yaml`

tpaschalis:

#### PR Description 
Since we can't have both `prometheus:` and `metrics:` keys in our config, use the first one until it's time for its proper deprecation.

This PR fixes #1255 

#### Notes to the Reviewer
We could also go the other way around, and change the `prometheus:` key in scraping-svc to `metrics:`.

However I'm not sure if there's anything else depending on scraping-svc; in any case the changes could be made to both places in a later deprecation/cleanup PR.

What do you think?

#### PR Checklist
- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/production/tanka/grafana-agent/config.libsonnet
+++ b/production/tanka/grafana-agent/config.libsonnet
@@ -58,7 +58,7 @@ local k8s_v2 = import './v2/internal/helpers/k8s.libsonnet';
         log_level: 'info',
       },
 
-      prometheus: {
+      metrics: {
         global: {
           scrape_interval: '1m',
         },
--- a/production/tanka/grafana-agent/scraping-svc/main.libsonnet
+++ b/production/tanka/grafana-agent/scraping-svc/main.libsonnet
@@ -33,7 +33,7 @@ local policyRule = k.rbac.v1.policyRule;
       agent_ring_kvstore: error 'must configure ring kvstore',
 
       agent_config+: {
-        prometheus+: {
+        metrics+: {
           // No configs are used in the scraping service mode.
           configs:: [],
 


example/k3d/environment/main.jsonnet
```diff
@@ -136,7 +136,7 @@ local images = {
 
       local cluster_label = 'k3d-agent/cluster',
       agent_config+: {
-        metrics+: {
+        prometheus+: {
```
rfratto, line 139: Shouldn't we switch everything to use `metrics` since prometheus is deprecated?

rfratto (commented):
Thanks :)

example/k3d/environment/main.jsonnet
tpaschalis: Yeah, I wanted your input on that. It certainly makes more sense, but I wasn't sure if anyone else may be depending on `scraping-svc`, and that change may break their work. WDYT?

rfratto: Nah, I think it's safe to update. It might break people, but the library should've been updated with the name change ages ago and it's arguably a bug that it hasn't been yet :)

tpaschalis: Okay, I've done that! I'll sanity-test the example once more just to be sure everything works as expected.

tpaschalis:
Okay, so the changes seem to be working. The configmap in the example contains only the `metrics:` key now, and so the pods get to run without any errors.

```
$ k describe configmap grafana-agent-cluster
...
Data
====
agent.yml:
----
metrics:
    global:
...
    scraping_service:
```

I still don't think that this doesn't warrant any changes to the documentation or the release notes, right?

rlankfo (approved):
LGTM :+1:

rfratto:
> I still don't think that this doesn't warrant any changes to the documentation or the release notes, right?

Nah, I think we can leave everything else as is. Thank you!

PR #1261, opened 2022-01-12
Introduce experimental cluster-aware metrics subsystem

rfratto:
This PR provides an initial implementation for #1140.

It does the following:

* Introduces an agent-wide clustering mechanism based on rfratto/ckit, allowing any subsystem to perform hash ring lookups for node ownership checking. 

* Introduces a revamped cluster-aware metrics subsystem outlined by #1140, hidden behind a feature flag.

* The revamped metrics subsystem has API endpoints to examine its state. These endpoints support aggregation, allowing you to request the API status of one agent and get the result for the whole cluster.

We may want to close this PR and split it up across multiple smaller PRs instead.


pkg/metrics/v2/sender.go
```diff
@@ -0,0 +1,263 @@
...
+type storageSet interface {
+	Appendable(name string) (storage, error)
+}
+
+type storage interface {
+	prom_storage.Appendable
+
+	// bindScraper binds the scraper to the storage. This is used for sending
+	// metadata.
+	bindScraper(sm *scrape.Manager)
```
rfratto, line 63: This was actually a pretty big issue I ran into. If we don't attach the scrape manager to remote write, then we don't get metadata for any of our scraped metrics (help/text info). 

Unfortunately, this also means that v2 integrations aren't generating any metadata because they're being scraped by an external scraper. I'm not sure how we should handle that yet, but I'll open up an issue for tracking.

rfratto (commented):
I did this for fun over some PTO so I wanted to give it a quick scan for some self feedback. Few high level thoughts so far:

1. I don't think we should have this in a dev branch forever, but I'm not sure if this is "ready" for main yet, even as a feature flag. 

2. I've talked a few times about trying to use gRPC as a transport instead of gossip. I'm not sure if that's a blocker, but it would simplify at least some of the setup code for the cluster. 

3. It feels a little awkward how much code was written for pkg/cluster. Should ckit be handling more?

4. I like the idea of calling metrics API endpoints and having them aggregate for every endpoint in the cluster, but I think maybe the wrong approach is being taken here. Should API endpoints default to being aggregated? Should aggregated API endpoints have their own dedicated path?

5. Maybe the k3d stuff should be a different PR since this is big enough as it is.

There's a lot of TODOs here, not all are blockers but I need to go through everything again a little bit later.

pkg/metrics/v2/scraper.go
```diff
@@ -0,0 +1,318 @@
...
+		metrics: newScraperMetrics(),
+
+		instanceGroups:   make(map[string]targetGroups),
+		scraperInstances: make(map[string]*scraper),
+	}
+}
+
+// ApplyConfig will update the managed set of scrapers. ApplyConfig should not
+// be called util the storageSet is reconfigured first.
+func (sm *scraperManager) ApplyConfig(cfg *Config) error {
```
rfratto, line 99: This should detect if a scrape job goes away, since the Discoverer won't send an empty target set for scrape jobs it doesn't know about.

pkg/metrics/v2/sender.go
rfratto: #1262

rfratto:
I'm going to close this and split it up across a couple of PRs.

PR #1264, opened 2022-01-13
Use RELEASE_TAG to choose between `:main` and `:latest` docker tags

### LINKED ISSUE #1256: Docker images: use `:latest` for latest stable release, `:main` for main branch builds
It's common in the Docker ecosystem for there to be a rolling-release stable image tag. Usually, this is `:latest`. 

Today, Grafana Agent doesn't have any stable image tag. Instead, it has:

* A set of explicit versioned tags (`:v0.21.0`) 
* A set of tags for every main-branch commit (`:main-<sha>`) and dev-branch commit (`:dev.<feature>-<sha>`)
* `:latest` pointing at the latest main branch commit. 

This can be surprising for users who want to automatically update their deployed Grafana Agents without having to rely on unstable builds. While I personally wouldn't recommend using `:latest` because of the chance of breaking changes, we still should support this use case given how common it is. 

I propose the following:

* Use `:latest` to point to the latest release 
* Use `:main` to point to the latest main branch commit 
* (Optionally) Use `:dev-<feature>` to point to the latest commit for a specific dev branch.

Note that this would be a breaking change for any user that currently relies on `:latest` being an unstable build. Those users, if they exist, would need to switch over to `:main` and make sure that it always gets pulled when a container is started.

tpaschalis:

#### PR Description 
This is an attempt to introduce a `:latest` rolling-release stable tag for release versions of the Agent, along with a `:main` tag for every new commit merged in main.

This approach depends on `$RELEASE_TAG` existing. This variable is set by [`DRONE_TAG`](https://github.com/grafana/agent/blob/0b8d63d795d85c4a1202ed795e74e8c35eb1414d/.drone/drone.yml#L159) in Drone and [`{{ github.event.inputs.tag }}`](https://github.com/grafana/agent/blob/4af620a03c39108edfefba34ea94d5efc7ffbcfe/.github/workflows/force_release.yml#L53) in Github Actions. I thought this could be one of the more harmless ways to go about it, without changing anything about our pipeline.

This PR closes #1256 
```bash
# If RELEASE_TAG and IMAGE_BRANCH_TAG are empty, then the tag will default to :main
$ echo $RELEASE_TAG $IMAGE_BRANCH_TAG

$ make agentctl-image
docker build --build-arg RELEASE_BUILD=false <...>  -t grafana/agentctl:"main" -t grafana/agentctl:fix-latest-docker-tag-2dbe3a3-WIP -f cmd/agentctl/Dockerfile .
                                                                         ^^^
# If Drone or Github Actions set ${IMAGE_BRANCH_TAG}, then it will be used instead of :main
$ export IMAGE_BRANCH_TAG="dev-fooo"
$ make agentctl-image
docker build --build-arg RELEASE_BUILD=false <...> -t grafana/agentctl:dev-fooo -t grafana/agentctl:fix-latest-docker-tag-ba946f9-WIP -f cmd/agentctl/Dockerfile .
                                                                         ^^^
# If Drone or Github Actions set ${RELEASE_TAG}, then it will take precedence
$ export RELEASE_TAG=v0.22.0
$ make agentctl-image
docker build --build-arg RELEASE_BUILD=false <...>  -t grafana/agentctl:"latest" -t grafana/agentctl:v0.22.0 -f cmd/agentctl/Dockerfile .
                                                                         ^^^
                                                                         

```

#### Notes to the Reviewer
- I'm not sure about possible drawbacks as I haven't gone through the entire release process yet.
- Should we default to something else other than `:main` for eg. local development? If yes, this could be handled via some env var indicating whether the process started from one of our pipeline (personally, not a big fan of it), or something like `git branch --show-current` to pick up the current branch name.
- Again, if we default to `:main` tags for local development, is that okay for our workflow or do we need to modify something else? (eg. we should probably run the smoke test before the release using :main instead of :latest)

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated


rfratto:
Thanks! I think this is pretty close, but maybe we should be able to override what the tag name ends up being.

We probably want to configure Drone to pass through `$DRONE_BRANCH` for non-release builds. Today, any commit to a dev branch also builds an image, and I don’t think we want that to override :main.

tpaschalis:
> Thanks! I think this is pretty close, but maybe we should be able to override what the tag name ends up being.
> 
> We probably want to configure Drone to pass through `$DRONE_BRANCH` for non-release builds. Today, any commit to a dev branch also builds an image, and I don’t think we want that to override :main.

Oh, I see what you mean! I just saw that we also trigger builds for `refs/heads/dev.*` branches and the tag will also be `:main`.

I'll amend with a commit. Are we still using the  _Force Release_ Github Action? (last use like 7 months ago), we can use sth like `${{ github.event.inputs.branch }}`for the same purpose.

mattdurham:
@rfratto thoughts on getting rid of force release? I don't think it will function correctly anymore since we switched to drone.

rfratto:
> I don't recall why we had to do a force release for v0.16.0, could be when we had issues with github actions working? I would assume its unlikely we would reuse it.
>
> @rfratto thoughts on getting rid of force release? I don't think it will function correctly anymore since we switched to drone.

Yeah, we were using it when we struggled with using actions for releases and while our Drone setup was still a WIP.

I think it's safe to remove now, we can always bring it back if we need to for some reason.

rfratto (approved):
This is good as-is but I think we can probably remove the force release job. If you're not going to do that LMK and I'll merge. Thank you!

rfratto (approved):
Thanks!

PR #1268, opened 2022-01-13
Add basic testing framework for operator

### LINKED ISSUE #1251: Invert condition used to exempt network tests

rfratto:
#### PR Description 
This PR includes the changes from #1251 and additionally introduces exactly one test for Grafana Agent Operator: it ensures that the end-to-end reconcile loop works, where the operator watches for resources and deploys at least what we expect it to.

While one new test doesn't add a lot of coverage, I'm hopeful that this will be the start of a series of tests and help kickstart stronger confidence in the correctness of the operator code.

Closes #1254 (this PR supersedes it)
Closes #1251 

#### PR Checklist

- [x] CHANGELOG updated (N/A)
- [x] Tests updated


.github/workflows/test.yml
```diff
@@ -16,6 +16,6 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v2
     - name: Build
-      run: make BUILD_IN_CONTAINER=false cmd/agent/agent cmd/agentctl/agentctl cmd/agent-operator/agent-operator
+      run: make GOFLAGS="-tags=netgo,nodocker" BUILD_IN_CONTAINER=false cmd/agent/agent cmd/agentctl/agentctl cmd/agent-operator/agent-operator
```
rlankfo, line 19: do we need netgo here since it's already defined in the Makefile where this GOFLAGS variable is expanded?

.github/workflows/test.yml
```diff
@@ -16,6 +16,6 @@ jobs:
...
     - name: Test
-      run: make BUILD_IN_CONTAINER=false test
\ No newline at end of file
+      run: make GOFLAGS="-tags=netgo,nodocker" BUILD_IN_CONTAINER=false test
```
rlankfo, line 21: same here

pkg/util/k8s/objects.go
```diff
@@ -0,0 +1,191 @@
...
+func ReadObjects(r io.Reader, cli client.Client) ([]client.Object, error) {
+	var (
+		objects []client.Object
+
+		scheme     = cli.Scheme()
+		rawDecoder = yaml.NewYAMLOrJSONDecoder(r, 4096)
+		decoder    = serializer.NewCodecFactory(scheme).UniversalDecoder(scheme.PrioritizedVersionsAllGroups()...)
+	)
+
+NextObject:
```
rlankfo, line 62: doesn't seem like there's any need for this label

pkg/util/k8s/objects.go
```diff
@@ -0,0 +1,191 @@
...
+
+// ReadUnstructuredObjects will read the set of objects from r as unstructured
+// objects.
+func ReadUnstructuredObjects(r io.Reader) ([]*unstructured.Unstructured, error) {
+	var (
+		objects    []*unstructured.Unstructured
+		rawDecoder = yaml.NewYAMLOrJSONDecoder(r, 4096)
+	)
+
+NextObject:
```
rlankfo, line 100: same here, unless I'm missing something

rlankfo (approved):
LGTM :+1:

.github/workflows/test.yml
rfratto: Yeah unfortunately, passing GOFLAGS here overrides the entire set of tags so you need to give netgo again

pkg/util/k8s/objects.go
rfratto: image

`break` in the switch will break out of the switch statement instead of the loop without the label 😞

rlankfo: ahh didn't notice that :+1:

PR #1274, opened 2022-01-14
Set custom User-Agent header when scraping metrics

### LINKED ISSUE #910: Metrics scraping: UserAgent reports "Prometheus/v0.x"
I noticed in logs `User-Agent: Prometheus/v0.17.0` (using Grafana Agent 0.17.0).
IMO it should say `GrafanaAgent` not `Prometheus`.

From what I see in the code, Grafana Agent sets its own name for Prometheus remote storage and for logs, but not for scraping.

tpaschalis:

#### PR Description 
After Robert's PR in (prometheus/prometheus#9333) we can now set `scrape.UserAgent` when the `instance` package is imported. I believe this may be the only required change, but let me know if I've missed anything.

This PR closes #910. I think at least, I'm still looking into how scraping mode works and  whether we have more work to do.

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/metrics/instance/instance.go
+++ b/pkg/metrics/instance/instance.go
@@ -35,6 +35,7 @@ import (
 
 func init() {
 	remote.UserAgent = fmt.Sprintf("GrafanaAgent/%s", build.Version)
+	scrape.UserAgent = fmt.Sprintf("GrafanaAgent/%s", build.Version)
 }
 
 var (


rfratto (approved):
Thanks!

PR #1283, opened 2022-01-18
Codegen to create static config

mattdurham:
#### PR Description 

Instead of using reflection to load the configuration which can be hard to understand use a code generator to generate golang structs for each integration. The goal is to make the code easier to understand and debug, while doing the dynamic configuration had to spend far too much time going over the reflection and this will also simplify dynamic configuration. 

#### Notes to the Reviewer

The number of files changed is scary! Thankfully most of them are package name changes, the primary ones that need to be looked at are the files in the:
* codegen directory
* v1/v2 config.go
* node_exporter PostProcessing

This is very open to if it should be done and cleanup. The package names changed and shared came about because the cyclic imports were like fighting a hydra. The generator creates roughly 700 lines of code (v1/config.go & v2/config.go). I removed the Agent V2 operator since I am unsure if it is needed, if we can always use v1 for as long as possible that seems simpler IMO.  

Another side effect is some special logic the Agent was doing in Unmarshal needs to be moved to a separate process. IE node_exporter's conversion. 

#### PR Checklist

- [X] CHANGELOG updated 
- [X] Documentation added
- [X] Tests updated


rfratto (commented):
I only gave this a quick skim, but I'm interested in seeing the final trimmed approach. 

In the PR's current state I think it's maybe generating _too_ much. I personally wouldn't try to generate methods to implement the interface, which feels like it's adding some newer magic that wasn't present in the original, especially since it runs the risk of generating incorrect implementations of methods. 

Also, I kind of like the registration system, honestly. I don't think it needs to use generics, but maybe there's a more dynamic way for the code generator to find the list of integrations it should be putting in the monolith config instead of hard coding a list of imports. (Or maybe there's an easier way to expose that list of imports that's not halfway down the file?)

mattdurham:
Fixed the issue with the imports, I can use a smidge of reflection to get the package path. I deduped some of the code generation by wrapping them in another struct. Not in love with this and the more verbose might be cleaner. I wonder if we need a `v2` version of the config at all, can we just keep making `v1` style and convert them as necessary. It does simplify some things. 

Also brought back the concept of a registry that lives in the agent instead of the codegen. I don't feel too strongly about where the registry lives, since it literally just contains a the metadata to use with codegen.

mattdurham:
On the plus side I am net negative lines of code =)

pkg/integrations/integrations.go
```diff
@@ -66,44 +70,53 @@ func (c VersionedIntegrations) MarshalYAML() (interface{}, error) {
...
-		return yaml.UnmarshalStrict(c.raw, c.configV1)
-	case integrationsVersion2:
+		c.ConfigV1 = &cfg
+		err := yaml.UnmarshalStrict(c.raw, c.ConfigV1)
+		// Node exporter has some post-processing that has to be done for migrations
+		if err != nil {
+			return err
+		}
+		if c.ConfigV1.Integrations.NodeExporter != nil {
+			return c.ConfigV1.Integrations.NodeExporter.Config.PostProcessing()
```
mattdurham, line 105: This is one thing that needed to change, instead of doing the node_exporter old -> new field name conversion during unmarshal move it to post processing.

CHANGELOG.md
```diff
@@ -34,6 +34,8 @@ for detailed information.
 
 - [ENHANCEMENT] Cherry-pick Prometheus PR #10102 into our Prometheus dependency (@rfratto).
 
+- [ENHANCEMENT] Use code generator to generator integration configuration instead of reflection (@mattdurham)
```
rfratto, line 37: This should be moved up to unreleased :)

rfratto:
I want to give this a good review, but is there a smaller set of files I should look at to get the gist of the improvements before I review the entire set of changes?

mattdurham:
Primarily how does the `/v1/config.go` and `/v2/config.go` feel, for the most part, it behaves like normal yaml. The codegen will need to evolve when we get more complex integrations, but it's pretty straightforward code.

mattdurham:
Also the actual codegenfile, the most important part is how it grabs the registry and mutates it.

cmd/codegen/codegenbuilder.go
```diff
@@ -0,0 +1,365 @@
...
+	{{ range $index, $element := . -}}
+	if v.{{ $element.Name }} != nil {
+        activeConfigs = append(activeConfigs, newConfigWrapper(&v.{{ $element.Name}}.Config, v.{{ $element.Name}}.Common))
+    }
+	{{ end -}}
+    for _, i := range v.TestConfigs {
+        activeConfigs = append(activeConfigs, i)
+    }
+    return activeConfigs
+}
```
rfratto, line 63: Serious question: do you feel like this is easier to understand than reflection?

rfratto: I don't mean to sound snarky, but genuinely want your opinion on if you think this is better than what we had before. I personally find complex text/templates pretty hard to follow and reason about, but I don't know how this compares

rfratto: Ultimately I think this will come down to what we feel is easier to understand, reason about, and maintain long term as integrations get more complex and we start supporting weirder niche use cases as the list grows.

pkg/integrations/v2/agent/agent.go
```diff
@@ -1,46 +0,0 @@
-// Package agent is an "example" integration that has very little functionality,
```
rfratto, line 1: Just curious: why was this taken out? Removing it means we have no v2 integrations now for testing, and my plan was to slowly migrate the existing v1 integrations to v2 over time.

cmd/codegen/main.go
```diff
@@ -0,0 +1,24 @@
...
+	"io/fs"
+	"io/ioutil"
+	"path/filepath"
+)
+
+func main() {
+
+	codegen := codeGen{}
+	v1Config := codegen.createV1Config()
+	err := ioutil.WriteFile(filepath.Join(".", "pkg", "integrations", "v1", "config.go"), []byte(v1Config), fs.ModePerm)
```
rfratto, line 13: Tiny nit: use 0664 here, `fs.ModePerm` is 0777 and gives the generated files the execute bit :)

rfratto (changes requested):
First of all, thank you for working on this. I think generating the integrations config struct monolith is easier to follow and generally less prone to being used incorrectly than the reflection. 

I also think this PR changes a little too much, and we might be better off splitting it into multiple smaller PRs now that you've done enough experimentation to have a good idea of what's necessary to do for this change. 

I have a few high level concerns with the current approach:

1. It feels a little unidiomatic to introduce `pkg/integrations/v1`. `/v2` packages are more common in Go for introducing breaking changes to a package, but there's usually not a `/v1`. I don't feel _super_ strongly about this one. It's possible that `/v2` was also a poor name since the intent is to remove the old code, and maybe `/next` would've been better. 

2. The introduction of `/shared` makes me a little uneasy, and I'm worried what it's signals to people writing new integrations. I view `/v2` as a complete rewrite of the integrations system, and _none_ of the legacy code should be imported with the exception of the temporary shim to use legacy integrations in the new subsystem. Both legacy and v2 code got put in the `/shared` package, which might signal "anything in here can be used for any version, and I don't think we should be encouraging that. 
 
3. There's a lot being exported that isn't used from external packages. I'm assuming that this is an artifact of how long you've been iterating on this, so splitting it up into newer PRs would be helpful here. 

4. I'm nervous about removing the ability to implement `UnmarshalYAML` from individual config types, and I think we should explore a general way to continue to permit it, maybe by using `util.UnmarshalYAMLMerged`. See my comment for one way it can go wrong. 

IMO, it feels like it makes sense to treat this PR as exploratory. Outside of my high-level concerns, I'm definitely in favor of the general approach here. But because of my high-level concerns, it might be less work to start over with a reduced amount of scope to get us to the same end result where code gen is being used. WDYT?

mattdurham:
That all sounds reasonable, I can start with a clean branch and start moving things over piecemeal.

rfratto:
Closing this for now after discussion in our bug scrub; we'd like to investigate this further but want to want for integrations-next to graduate out of a feature flag.

PR #1288, opened 2022-01-19
[dev.operator-integrations] operator: add CRDs for integrations

rfratto:
This PR introduces CRDs for integrations:

* MetricsIntegration is a child of GrafanaAgent and is used to specify a
  running metrics-based integration.

* IntegrationMonitor is a child of MetricsInstance and is used to
  specify which MetricsIntegrations should be scraped for metrics.

The CRDs have been added to the resource hierarchy and will
automatically be discovered when reconciling a GrafanaAgent resource.
While the resources are discovered, this commit does *not* run
integrations or collect metrics from them.

As a part of this change, the old config.Deployment type has been moved
to pkg/operator/apis/monitoring/v1alpha1. This allows it to take
advantage of the DeepCopy function generation, a method which was
previously manually maintained.


pkg/operator/reconciler.go
```diff
@@ -88,27 +87,26 @@ func (r *reconciler) Reconcile(ctx context.Context, req controller.Request) (con
...
 	secret := core_v1.Secret{
 		ObjectMeta: v1.ObjectMeta{
-			Namespace: d.Agent.Namespace,
-			Name:      fmt.Sprintf("%s-secrets", d.Agent.Name),
+			Namespace: h.Agent.Namespace,
+			Name:      fmt.Sprintf("%s-secrets", h.Agent.Name),
 			OwnerReferences: []v1.OwnerReference{{
-				APIVersion:         d.Agent.APIVersion,
+				APIVersion:         h.Agent.APIVersion,
 				BlockOwnerDeletion: &blockOwnerDeletion,
```
rlankfo, line 106: no `pointer.Bool(true)` here? :D

rlankfo (commented):
LGTM overall but I believe the renaming of `IgnoreNamespaceSelectors` was unintentional and it's updating the generated CRD.

PR #1289, opened 2022-01-20
Skip non-ready entries when listing instances

### LINKED ISSUE #1126: WAL cleanup panics if the WAL is not yet loaded
```
ts=2021-11-23T12:26:09.634937274Z caller=node.go:85 level=info agent=prometheus component=cluster msg="applying config"
ts=2021-11-23T12:26:09.63557965Z caller=remote.go:180 level=info agent=prometheus component=cluster msg="not watching the KV, none set"
ts=2021-11-23T12:26:09Z level=info caller=traces/traces.go:123 msg="Traces Logger Initialized" component=traces
ts=2021-11-23T12:26:09.648577389Z caller=server.go:77 level=info msg="server configuration changed, restarting server"
ts=2021-11-23T12:26:09.649497502Z caller=gokit.go:72 level=info http=[::]:3090 grpc=[::]:9095 msg="server listening on addresses"
ts=2021-11-23T12:26:09.786774813Z caller=wal.go:182 level=info agent=prometheus instance=14d8e3c77664fe89fcfcb980c9ba0874 msg="replaying WAL, this may take a while" dir=/var/lib/agent/data/14d8e3c77664fe89fcfcb980c9ba0874/wal
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x1a293f0] 
goroutine 253 [running]:
github.com/grafana/agent/pkg/metrics/instance.(*Instance).StorageDirectory(0xc000889e40, 0xc08d627830, 0xc00063fd78)
    /src/agent/pkg/metrics/instance/instance.go:559 +0x30
github.com/grafana/agent/pkg/metrics.(*WALCleaner).getManagedStorage(0xc00013e1e0, 0xc08d627830, 0x1)
    /src/agent/pkg/metrics/cleaner.go:154 +0x9d
github.com/grafana/agent/pkg/metrics.(*WALCleaner).cleanup(0xc00013e1e0)
    /src/agent/pkg/metrics/cleaner.go:248 +0xb7
github.com/grafana/agent/pkg/metrics.(*WALCleaner).run(0xc00013e1e0)
    /src/agent/pkg/metrics/cleaner.go:237 +0x98
created by github.com/grafana/agent/pkg/metrics.NewWALCleaner
    /src/agent/pkg/metrics/cleaner.go:145 +0x205
```

Here the WAL loading takes longer than 30mins, at which point the WALCleaner runs. But because WAL is `nil`, it panics. We should fix that.

tpaschalis:

#### PR Description
As mentioned in the linked issue, the ListInstances() method may return nil values in `map[string]ManagedInstance`, when an instance's process has not launched yet.

For instances that aren't ready yet this can result to panics in two places
- When WALCleaner runs, trying to get an instance's storage directory
- When someone calls `/agent/api/v1/targets` to list an instance's active targets

This PR tries to amend the issue by skipping over nil values in ListInstances(); so these instances will have their WAL cleaned and will be listed on the endpoint the next time that the method is called.

This PR closes #1126.

#### Notes to the Reviewer
Do y'all think we should add a test for this?

#### PR Checklist

- [x] CHANGELOG updated (N/A)
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 - [ENHANCEMENT] Go 1.17 is now used for all builds of the Agent. (@tpaschalis)
 
+- [BUGFIX] Fixed issue where Grafana Agent may panic if there is a very large
+  WAL loading while old WALs are being deleted or the `/agent/api/v1/targets`
+  endpoint is called. (@tpaschalis)
+
 # v0.22.0 (2022-01-13)
 
 This release has deprecations. Please read [DEPRECATION] entries and consult
--- a/pkg/metrics/instance/manager.go
+++ b/pkg/metrics/instance/manager.go
@@ -149,6 +149,9 @@ func (m *BasicManager) ListInstances() map[string]ManagedInstance {
 
 	res := make(map[string]ManagedInstance, len(m.processes))
 	for name, process := range m.processes {
+		if process == nil {
+			continue
+		}
 		res[name] = process.inst
 	}
 	return res


rfratto:
Thanks! Is there other code that calls `ListInstances` that might also panic here? Should we change the implementation of that method instead so it never returns nil instances instead?

tpaschalis:
Good call, there's a similar piece of code on `ListTargetsHandler` when calling the `/agent/api/v1/targets` endpoint that might panic.
https://github.com/grafana/agent/blob/b890ef98ef25ea129041c7eb79bd7e5656e390bc/pkg/metrics/http.go#L42-L46

@rfratto you're right, it probably makes sense to fix the root cause in `ListInstances()` and only return the non-nil instances there (we'll just possibly be over-allocating space in that map).

Not sure if there's any other side effects to that, so I'll check once again and come back to you^^

rfratto (approved):
LGTM, thank you!

PR #1292, opened 2022-01-20
Experiment with simplifying integrations reflection code

rfratto:
This PR uses the existing `util.UnmarshalYAMLMerged` and `util.MarshalYAMLMerged` fucntions to see if it helps with reducing the code complexity of unmarshaling integrations.

Notably, this has performance tradeoffs because of the need to generate a `util.RawYAML` for unmarshaling.


rfratto:
I'm going to close this. After playing around with it, there's too many footguns compared to #1283, and unlike Prometheus, we don't need the ability to opt in to which integrations are available. 

cc @mattdurham

PR #1293, opened 2022-01-21
cadvisor config Fixes #1281

### LINKED ISSUE #1281: [Bug]: cadvisor can't initial correctly
I'm trying to use latest v0.22.0 and cadvisor interintegration.

agent is install by deb packages with root user and this is my config:

```yaml
integrations:
  cadvisor:
    enabled: true
    scrape_interval: 15s
    store_container_labels: false
    docker_only: true
    storage_duration: 10m
    docker: "unix:///var/run/docker.sock"
    enabled_metrics:
      - disk
```

But I got this error log: ` ts=2022-01-18T12:48:44.074899092Z caller=manager.go:372 level=error msg="integration has panicked. THIS IS A BUG!" err="runtime error: index out of range [0] with length 0" integration=cadvisor`

try to compile the latest HEAD and add print to show panic trace:
```text
github.com/grafana/agent/pkg/integrations.(*integrationProcess).Run.func1()
        /home/trim21/proj/agent/pkg/integrations/manager.go:373 +0x7b
panic({0x3870bc0, 0xc00163e2a0})
        /home/trim21/.gvm/versions/go1.17.2.linux.amd64/src/runtime/panic.go:1038 +0x215
github.com/google/cadvisor/container/raw.(*rawFactory).CanHandleAndAccept(0xc0028d3720, {0xc0028d3720, 0x2c})
        /home/trim21/go/pkg/mod/github.com/google/cadvisor@v0.43.0/container/raw/factory.go:74 +0x127
github.com/google/cadvisor/container.NewContainerHandler({0xc0028d3720, 0x19}, 0x15, {0x0, 0x0, 0x0}, 0x25)
        /home/trim21/go/pkg/mod/github.com/google/cadvisor@v0.43.0/container/factory.go:249 +0x1c6
github.com/google/cadvisor/manager.(*manager).createContainerLocked(0xc000d4c2c0, {0xc0028d3720, 0x0}, 0x0)
        /home/trim21/go/pkg/mod/github.com/google/cadvisor@v0.43.0/manager/manager.go:919 +0xca
github.com/google/cadvisor/manager.(*manager).createContainer(0xc000d4c2c0, {0xc0028d3720, 0x19}, 0x9c8f50)
        /home/trim21/go/pkg/mod/github.com/google/cadvisor@v0.43.0/manager/manager.go:906 +0xa5
github.com/google/cadvisor/manager.(*manager).detectSubcontainers(0xc000d4c2c0, {0x3e74258, 0x23})
        /home/trim21/go/pkg/mod/github.com/google/cadvisor@v0.43.0/manager/manager.go:1121 +0x235
github.com/google/cadvisor/manager.(*manager).Start(0xc000d4c2c0)
        /home/trim21/go/pkg/mod/github.com/google/cadvisor@v0.43.0/manager/manager.go:305 +0x26c
github.com/grafana/agent/pkg/integrations/cadvisor.(*Integration).Run(0xc0007bd5d0, {0x445f188, 0xc00197d640})
        /home/trim21/proj/agent/pkg/integrations/cadvisor/cadvisor.go:125 +0x3be
github.com/grafana/agent/pkg/integrations.(*CollectorIntegration).Run(0x0, {0x445f188, 0xc00197d640})
        /home/trim21/proj/agent/pkg/integrations/collector_integration.go:104 +0x2d
github.com/grafana/agent/pkg/integrations.(*integrationProcess).Run(0xc000189930)
        /home/trim21/proj/agent/pkg/integrations/manager.go:384 +0x113
created by github.com/grafana/agent/pkg/integrations.(*Manager).ApplyConfig
        /home/trim21/proj/agent/pkg/integrations/manager.go:290 +0x146f
```

rgeyer:
#### PR Description
This adds the expected zero value for `raw_cgroup_prefix_allowlist` as the default for the cadvisor integration configuration.

Fixes #1281 

#### PR Checklist

- [ ] CHANGELOG updated 
- N/A Documentation added
- N/A Tests updated

### DIFF
--- a/pkg/integrations/cadvisor/cadvisor.go
+++ b/pkg/integrations/cadvisor/cadvisor.go
@@ -153,7 +153,7 @@ func (i *Integration) Run(ctx context.Context) error {
 	if err := rm.Stop(); err != nil {
 		return fmt.Errorf("failed to stop manager: %w", err)
 	}
-	return ctx.Err()
+	return nil
 }
 
 // New creates a new cadvisor integration
--- a/pkg/integrations/cadvisor/cadvisor.go
+++ b/pkg/integrations/cadvisor/cadvisor.go
@@ -91,6 +91,10 @@ type Integration struct {
 func (i *Integration) Run(ctx context.Context) error {
 	// Do gross global configs. This works, so long as there is only one instance of the cAdvisor integration
 	// per host.
+
+	// klog
+	klog.SetLogger(i.c.logger)
+
 	// Containerd
 	containerd.ArgContainerdEndpoint = &i.c.Containerd
 	containerd.ArgContainerdNamespace = &i.c.ContainerdNamespace
@@ -154,7 +158,7 @@ func (i *Integration) Run(ctx context.Context) error {
 
 // New creates a new cadvisor integration
 func New(logger log.Logger, c *Config) (integrations.Integration, error) {
-	klog.SetLogger(logger)
+	c.logger = logger
 
 	ci := integrations.NewCollectorIntegration(c.Name())
 	integration := Integration{
--- a/pkg/integrations/cadvisor/cadvisor_test.go
+++ b/pkg/integrations/cadvisor/cadvisor_test.go
@@ -1,3 +1,6 @@
+//go:build !nonetwork && !nodocker
+// +build !nonetwork,!nodocker
+
 package cadvisor
 
 import (
--- a/pkg/integrations/cadvisor/cadvisor_test.go
+++ b/pkg/integrations/cadvisor/cadvisor_test.go
@@ -1,0 +1,49 @@
+package cadvisor
+
+import (
+	"context"
+	"testing"
+
+	"github.com/go-kit/log"
+	"github.com/stretchr/testify/assert"
+	"gopkg.in/yaml.v3"
+)
+
+func run_integration(t *testing.T, cfgStr string) error {
+	var cfg Config
+
+	err := yaml.Unmarshal([]byte(cfgStr), &cfg)
+	assert.NoError(t, err)
+	ig, err := cfg.NewIntegration(log.NewNopLogger())
+	assert.NoError(t, err)
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
+	return ig.Run(ctx)
+}
+
+func TestConfig_DockerOnly(t *testing.T) {
+	t.Run("docker_only with default configuration is successful", func(t *testing.T) {
+		// Run it once with the default config, expecting success.
+		defaultCfg := `
+docker_only: true
+`
+		var err error
+
+		assert.NotPanics(t, func() { err = run_integration(t, defaultCfg) })
+		assert.ErrorIs(t, err, context.Canceled)
+	})
+
+	// 	t.Run("docker_only with empty raw_cgroup_prefix_allowlist panics", func(t *testing.T) {
+	// 		// then again when docker_only is true, and raw_cgroup_prefix_allowlist is an empty array,
+	// 		// expecting the cadvisor collectors to panic. If this suddenly starts hanging, or does not panic, the default
+	// 		// value for raw_cgroup_prefix_allowlist should be returned to a zero value string slice.
+	// 		//
+	// 		var err error
+	// 		panicCfgStr := `
+	// docker_only: true
+	// raw_cgroup_prefix_allowlist: []
+	// `
+	// 		assert.Panics(t, func() { err = run_integration(t, panicCfgStr) })
+	// 		assert.NoError(t, err)
+	// 	})
+}
--- a/pkg/integrations/cadvisor/cadvisor_test.go
+++ b/pkg/integrations/cadvisor/cadvisor_test.go
@@ -9,7 +9,7 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
-func run_integration(t *testing.T, cfgStr string) error {
+func runIntegration(t *testing.T, cfgStr string) error {
 	var cfg Config
 
 	err := yaml.Unmarshal([]byte(cfgStr), &cfg)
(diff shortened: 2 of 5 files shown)


pkg/integrations/cadvisor/common.go
```diff
@@ -12,6 +12,9 @@ const name = "cadvisor"
 var DefaultConfig Config = Config{
 	// Common cadvisor config defaults
 	StoreContainerLabels: true,
+	AllowlistedContainerLabels: []string{},
+	EnvMetadataAllowlist: []string{},
+	RawCgroupPrefixAllowlist: []string{"",},
```
rfratto, line 17:
```suggestion
	RawCgroupPrefixAllowlist: []string{""},
```

rfratto (approved):
Is there a way we can add a test for this? Or would that require docker or something to be running?

rgeyer:
> Is there a way we can add a test for this? Or would that require docker or something to be running?

It would fail... Differently, if there was no docker daemon to actually fetch cadvisor metrics from.

In fact, at first I couldn't even reproduce this error, until I used the config provided in the issue. It seems to only get used by cadvisor when the `docker_only` is set.

So.. We could build a test that expects failure, but the message has to do with connecting to docker (whatever that message ends up being), rather than the config initialization error that was initially caught. 🤔

rfratto:
> It would fail... Differently, if there was no docker daemon to actually fetch cadvisor metrics from.

Our CI [runs with access to a Docker daemon now](https://github.com/grafana/agent/blob/0759ad39817f9dfc7e82985b52d0dedbbc742508/.drone/drone.yml#L44-L46), and the only requirement on the testing side is that you add [build tags](https://github.com/grafana/agent/blob/0759ad39817f9dfc7e82985b52d0dedbbc742508/pkg/operator/hierarchy/hierarchy_test.go#L1-L2) so people can opt out of running tests that require it.

Would that help write a test for this?

rgeyer:
> Would that help write a test for this?

No problem! I've got to wrap up another couple priorities, but I'll be happy to implement this ASAP.

pkg/integrations/cadvisor/cadvisor.go
```diff
@@ -91,6 +91,10 @@ type Integration struct {
 func (i *Integration) Run(ctx context.Context) error {
 	// Do gross global configs. This works, so long as there is only one instance of the cAdvisor integration
 	// per host.
+
+	// klog
+	klog.SetLogger(i.c.logger)
```
rgeyer, line 96: When running in tests, with parallelization, and the `-race` flag, this triggers a race condition that I can't seem to easily workaround. I'm tempted to remove this completely.

pkg/integrations/cadvisor/cadvisor_test.go
```diff
@@ -0,0 +1,49 @@
...
+	// 		// value for raw_cgroup_prefix_allowlist should be returned to a zero value string slice.
+	// 		//
+	// 		var err error
+	// 		panicCfgStr := `
+	// docker_only: true
+	// raw_cgroup_prefix_allowlist: []
+	// `
+	// 		assert.Panics(t, func() { err = run_integration(t, panicCfgStr) })
+	// 		assert.NoError(t, err)
+	// 	})
```
rgeyer, line 48: Ideally, I'd be able to perform this test too, which validates that there *is* a panic when the config is wrong. If this were enabled, and began hanging or failing to panic, we'd know that upstream cadvisor changed in some significant way. Specifically, that a raw_cgroup_prefix_allowlist that has one empty string in a slice, is no longer an expected default.

rgeyer (commented):
I spent ENTIRELY too long figuring out how best to test this. In the end, I ran into a lot of issues with parallelization, specifically because of assigning the logger from the agent to `klog` which cadvisor uses internally.

pkg/integrations/cadvisor/cadvisor_test.go
```diff
@@ -0,0 +1,49 @@
...
+	// 		// expecting the cadvisor collectors to panic. If this suddenly starts hanging, or does not panic, the default
+	// 		// value for raw_cgroup_prefix_allowlist should be returned to a zero value string slice.
+	// 		//
+	// 		var err error
+	// 		panicCfgStr := `
+	// docker_only: true
+	// raw_cgroup_prefix_allowlist: []
+	// `
+	// 		assert.Panics(t, func() { err = runIntegration(t, panicCfgStr) })
+	// 		assert.NoError(t, err)
```
colega, line 47: This check wouldn't make sense since you asserted that `runIntegration` panics so `err =` assignment is never executed.

pkg/integrations/cadvisor/cadvisor.go
mattdurham: Possible use of `sync.Once`?

rgeyer: Yeah, that *might* have helped. Ultimately, it seems to be that cadvisor isn't properly shutting everything down, and waiting for background go routines to finish before returning. So there's still something which writes to the log after my test is "finished", and that races the next test executing and setting the logger.

That said, with some collab w/ @rfratto, there's only a single test, so we're dodging the bullet for now.

pkg/integrations/cadvisor/cadvisor_test.go
rgeyer: Yep, excellent point. I was struggling to get the tests working at all, and it never actually executed to that point.

Thanks for the heads up!

rfratto:
@rgeyer how much more work do we need here to get this merged?

rgeyer:
> @rgeyer how much more work do we need here to get this merged? Is it ready for another review round?

Right, sorry, this fell off my radar.

I believe this is "done". If someone wants to give it one more round of review that's fine, else I'm ready to merge.

rfratto (approved):

PR #1301, opened 2022-01-24
[dev.frontend-receiver] Add app o11y receiver integration

domasx2:
NOTE: This integration is implemented using the v2 Integration interface

This commit is the first draft of integrating the [App Observability Telemetry Receiver][1] into the agent. For now the following components are supported:

### Logs

The exporter forwards the logs received onto the HTTP endpoint into the logs subsystem.

### Exceptions

Same as logs, but with fixed labels.

### Metrics

The integration is capable of reporting both some predetermined performance metrics (web vitals in this iteration) but also some user defined ones using a specific schema in the configuration file. Internally, a Prometheus Summary type is used to keep track of the incoming values and is also using the self scrape system to report the metrics.

The exporter itself is basically a modular HTTP endpoint that uses different exporters to forward received data into the appropriate subsystems of the agent.

[1]: https://github.com/grafana/grafana-frontend-telemetry-receiver

- [X] CHANGELOG updated 
- [X] Documentation added
- [X] Tests updated
- [X] remove traces from payload, until frontend agent implements support


pkg/integrations/v2/app_o11y_exporter/receiver/receiver.go
```diff
@@ -76,25 +77,12 @@ func (ar *AppReceiver) ReceiverHandler(logger *log.Logger) http.Handler {
...
-
-			go func(exp exporters.AppReceiverExporter) {
-				defer wg.Done()
-				// Data exporters, export in sync with the user request
-				if de, ok := exp.(exporters.AppDataExporter); ok {
-					if err = de.Export(p); err != nil {
-						errChan <- err
-					}
+				if err = exp.Export(p); err != nil {
+					level.Error(logger).Log("msg", "exporter error", "exporter", exp.Name(), "error", err.Error())
```
kpelelis: I think this could be debug level instead of log

docs/configuration/integrations/integrations-next/app-o11y-exporter.md
```diff
@@ -55,15 +57,10 @@ Full reference of options:
     [host: <string> | default = "0.0.0.0"]
     [port: <number> | default = 8080]
 
-  # User defined prometheus metrics to be scraped. The sending end of the agent can
-  # update this metrics using the specified payload. Since these are performance metrics
-  # a Summary type metric is used internally for each metric defined
-  custom_measurements:
```
kpelelis: Why is this removed?

pkg/integrations/v2/app_o11y_receiver/models/trace.go
```diff
@@ -0,0 +1,83 @@
+package models
```
kpelelis, line 1: I wouldn't add tracing just yet as they will not be part of the first draft

pkg/integrations/v2/app_o11y_receiver/handler/handler.go
```diff
@@ -73,33 +73,20 @@ func (ar *AppO11yHandler) HTTPHandler(logger log.Logger) http.Handler {
...
-		select {
-		case <-wgDone:
-			w.WriteHeader(http.StatusAccepted)
-			_, _ = w.Write([]byte("ok"))
-		case err := <-errChan:
-			close(errChan)
-			http.Error(w, err.Error(), http.StatusInternalServerError)
-		}
+		wg.Wait()
+		w.WriteHeader(http.StatusAccepted)
```
kpelelis, line 88: What should happen in case of an error? Shouldn't we inform the consumer somehow?

kpelelis (changes requested):
Looks good from the first draft. I would only skip tracing for now to reduce complexity + some minor comments

pkg/integrations/v2/app_o11y_exporter/receiver/receiver.go
domasx2: It's `error` level. Imho this is correct, exporter failures should be a unexpected and actionable issue, not ok to swallow them or hide under debug level

pkg/integrations/v2/app_o11y_exporter/exporters/loki.go
```diff
@@ -38,16 +38,12 @@ func (le *LokiExporter) Init() error {
...
 
 	return &LokiExporter{
-		logger:      logger,
-		li:          conf.LokiInstance,
-		seTimeout:   time.Duration(conf.SendEntryTimeout),
-		extraLabels: el,
+		logger:    logger,
+		li:        conf.LokiInstance,
+		seTimeout: time.Duration(conf.SendEntryTimeout),
+		labels:    conf.Labels,
```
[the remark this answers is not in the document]
domasx2: I'd rather contain Loki-specific types to the internals of  Loki exporter, and not have outside callers have to deal with them. Also I remember this made code simpler somehow, but now forget the details :)

docs/configuration/integrations/integrations-next/app-o11y-exporter.md
domasx2: Any metrics should be defined by the frontend. AFAIK no other tool requires making backend agent config changes to add a metric.
 Also as discussed, removed prometheus exporter for now. Imho if we do a prometheus exporter later, measurements should be registered dynamically based on what front is sending

pkg/integrations/v2/app_o11y_receiver/models/trace.go
domasx2: My initial idea was to define accepted payload first, then implement various services using hte payload as a reference.
But as discussed, I'll comment out the traces member for now

pkg/integrations/v2/app_o11y_receiver/handler/handler.go
domasx2: Couple of things: 
* `StatusAccepted` does not imply that request was successfully proccessed, just that it was a valid request
* Previous implementation intended to report a single error, but would segfault if there was more than one expoter due to concurrency issues. Imho to do proper error reporting we'd need to do some design work first and figure out how best to report multiple errors, what do if some exporters succeeded and some not. I'd wait to see if there's a practical reason for it first

pkg/integrations/v2/app_o11y_receiver/models/trace.go
domasx2: removed tracing for now

pkg/integrations/v2/app_o11y_exporter/exporters/loki.go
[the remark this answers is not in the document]
kpelelis: If you want to keep the configuration separated from Loki, I am onboard. You could pre-evaluate the loki label set in the Initialisation phase of the exporter though (Init method)

docs/configuration/integrations/integrations-next/app-o11y-exporter.md
kpelelis: Then we should remove the [respective config](https://github.com/grafana/agent/pull/1301/files#diff-4deef3ff88ba2ad90cc619164577b484819249b25747d2821763e264055427dfR48-R51) as well

pkg/integrations/v2/app_o11y_receiver/handler/handler.go
kpelelis: If we don't really care about the result then, we can remove the wait group altogether. If the user waits for the full time of the request then we might as well let them know what happened. I do agree that 202 is the proper response here if we want to make this async.

My proposal(s) are:

* Remove blocking wait and return 202
or
* Execute everything synchronously and return the result

I lean towards the first option, as these systems are supposed to be fault tolerant

pkg/integrations/v2/app_o11y_exporter/exporters/loki.go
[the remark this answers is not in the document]
domasx2: some of the label values are taken from data, so can't pre-evaluate some of it. What is the concern though? I don't think there are performance implications, this is only compile-time type checking thing...

docs/configuration/integrations/integrations-next/app-o11y-exporter.md
domasx2: done!

pkg/integrations/v2/app_o11y_receiver/handler/handler.go
domasx2: I have thought about this, and would like to keep current approach because:
* Removing blocking makes writing tests harder :frog: how then to block the test until all exporters have finished? probably solvable, but don't wanna spend time on it
* Also by removing blocking frontend won't be able to tell as well if receiver is struggling and limit concurrent requests 
* Again, collecting & errors and figuring out how to communicate them is not easy and not immediately useful, would rather leave this for a separate issue/PR down the line

kpelelis: I agree we can skip this for now. As a heads up, I am not questioning one technique over the other, but rather the semantic correctness. The 202 status generally means that the the request has been accepted for processing, which at the point we don't do. We are completing the request normally. To comment on your points

> how then to block the test until all exporters have finished

It depends on the scope of the test. If it is testing the HTTP server, in terms of expected input/output, then testing that the server returns 202 is a form of a test. The async process that completes the request can be treated as a blackbox. I do agree that go makes extremely hard to test this

> Also by removing blocking client won't be able to tell as well if receiver is struggling and limit concurrent requests

Good point. We can solve this by releasing the rate limiting "lease" once the request is completed.

I think we can resolve this convo if that is okay by you

pkg/integrations/v2/app_o11y_exporter/exporters/loki.go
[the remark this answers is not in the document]
kpelelis: I was mostly talking about [this](https://github.com/grafana/agent/pull/1301/files#diff-3575e066eabbf90a39f6b09547a5ef9bf97946f39de6f529a3adc8f08be77643R104) line in which the key and the value of the `LokiExporter.labels` are getting casted into a label and value respectively. This is a nitpick in any case, so feel free to ignore

pkg/integrations/v2/app_o11y_receiver/handler/handler.go
domasx2: when testing http handler, I also wanna make sure that it **does the thing (tm)** , otherwise it's not super useful. Thanks, I'm sure we'll get back to these points soon and improve

kpelelis (approved):
LETS GOOOO

Issue #869, opened 2021-09-01
Load configuration files via URL


rfratto:
Today, the agent's base config file may only be retrieved from disk. An exception to this rule is the scraping service, where a KV store must be used to pull scrape configs. The KV store was built to target huge deployments of the Agent, but is still being forced on smaller users of the scraping serivice, as it is the only available option. An alternative approach to allow remote configuration of both clustered and non-clustered agents could allow the agent to be far more generally useful. 

I suggest we allow users to specify a config file via URL, perhaps through using [go-fsimpl](https://github.com/hairyhenderson/go-fsimpl). This would allow users to receive configs from a storage bucket, an HTTP endpoint, a file, or more. 

Once implemented, you can imagine that other file fields in the agent could also be extended to support URLs. One example may be #868, where the combination of this issue and conf.d support could fully supplant the config management API with something that has no external dependencies. That is out of scope for this issue, but is something I'd like to work towards.

Credit to @mattdurham for the idea of URL-based config files.

rlankfo:
This request was initially implemented for HTTP/HTTPS URLs in https://github.com/grafana/agent/pull/1143

PR #1308, opened 2022-01-24
[RFC] Dynamic Documentation

mattdurham:
This covers only the documentation which broadly defines what happens, and the feature flag itself.

This is to address issues #1121, #869, and #868. 

The goal is to 
* Allow the loading of all configuration files locally and remotely from a variety of sources. 
* Allow the templating of configuration files from a variety of sources, for instance pull secrets from Consul/Vault other sources
* Allow the configuration file itself to be broken down into logical sections that are recombined in a set order.  

The major mechanism it will be using to do this is by a custom fork of the [gomplate](https://github.com/hairyhenderson/gomplate) tool, that allows one to embed it has a library.

### DIFF
--- a/docs/configuration/dynamic-config.md
+++ b/docs/configuration/dynamic-config.md
@@ -7,7 +7,7 @@ Dynamic Configuration is the combination of two things:
 * Loading from multiple files
 * Using templates and datasources
 
-Both of these make heavy use of the excellent [gomplate](https://github.com/hairyhenderson/gomplate). The idea goal is
+Both of these make heavy use of the excellent [gomplate](https://github.com/hairyhenderson/gomplate). The goal is
 that as the configuration grows that it can be split it up into smaller segments to allow better readability and handling.
 The configurations cannot be patched in any order and instead are allowed at several levels.
 
--- a/docs/configuration/dynamic-config.md
+++ b/docs/configuration/dynamic-config.md
@@ -1,0 +1,118 @@
+# Dynamic Configuration - BETA
+
+**This is BETA and subject to change at anytime, feedback is much appreciated**
+
+Dynamic Configuration is the combination of two things:
+
+* Loading from multiple files
+* Using templates and datasources
+
+Both of these make heavy use of the excellent [gomplate](https://github.com/hairyhenderson/gomplate). The idea goal is
+that as the configuration grows that it can be split it up into smaller segments to allow better readability and handling.
+The configurations cannot be patched in any order and instead are allowed at several levels.
+
+The second goal is to allow the use of templating, functions for [gomplate doc](https://docs.gomplate.ca/) go into detail
+on what functions are available.
+
+## Configuration
+
+Location of the dynamic configuration is used via the feature flag `dynamic-config`, then it will use `-config.file` to
+load the configuration for dynamic configuration.
+
+```yaml
+# Sources to pull template values 
+[sources: <sources_config>]
+
+# Locations to use searching for templates, the system does NOT look into subdirectories. Follows gomplate schema
+# from [gomplate datasources](https://docs.gomplate.ca/datasources/). File and S3/GCP templates are currently supported
+templates: 
+[ - string ]
+
+``` 
+
+### sources_config
+```yaml
+# Name of the source to use when templating
+name: string
+
+# Path to datasource using schema from [gomplate datasources](https://docs.gomplate.ca/datasources/) 
+url: string
+
+```
+
+## Templates
+
+Note when adding a template you do NOT need to add the type as the top level yaml field. For instance if using traces:
+
+Incorrect
+
+```yaml
+traces:
+  configs:
+  - name: default
+    automatic_logging:
+      backend: loki
+      loki_name: default
+      spans: true
+```
+
+Correct
+
+```yaml
+configs:
+- name: default
+  automatic_logging:
+    backend: loki
+    loki_name: default
+    spans: true
+```
+
+Configurations are loaded in the order as they are listed below.
+
+### Agent
+
+
+Agent template is the standard agent configuration file in its entirety. It is defined by the pattern `agent-*.yml`. Only
+one file is supported. This is processed first then any subsequent configurations found REPLACE the values here, it is
+not additive.
+
+Reference {{< relref "./" >}})
+
+### Server
+
+Server is defined using the pattern `server-*.yml`, only ONE server file is supported.
+
+Reference {{< relref "./server-config.md" >}})
+
+
+### Metrics
+
+Metrics are defined using the pattern `metrics-*.yml`, only ONE metrics file is supported.
+
+Reference {{< relref "./metrics-config.md" >}})
+
+### Metric Instances
+
+Metric Instances are defined using the pattern `metrics_instances-*.yml`.
+
+Reference {{< relref "./metrics-config.md#metrics_instance_config" >}}) in the metrics instance
+
+
+### Integrations
+
+Integrations are defined using the pattern `integrations-*.yml`, these support more than one file, and multiple
+integrations can be defined in a file. Do not assume any order of loading for integrations.
+
+Reference {{< relref "./integrations/" >}})
+
+### Traces
+
+Traces are defined using the pattern `traces-*.yml`. This supports one file.
+
+Reference {{< relref "./traces-config.md" >}})
+
+### Logs
+
+Logs are defined using the pattern `logs-*.yml`. This supports on file.
+
+Reference {{< relref "./logs-config.md" >}})
--- a/docs/configuration/dynamic-config.md
+++ b/docs/configuration/dynamic-config.md
@@ -21,14 +21,23 @@ load the configuration for dynamic configuration.
 
 ```yaml
 # Sources to pull template values 
-sources: 
+datasources: 
   [- <sources_config>]
 
 # Locations to use searching for templates, the system does NOT look into subdirectories. Follows gomplate schema
 # from [gomplate datasources](https://docs.gomplate.ca/datasources/). File and S3/GCP templates are currently supported
 template_paths: 
   [ - string ]
+  
+# Filters allow you to override the default naming convention
 
+agent_filter:            string # defaults to agent-*.yml
+server_filter:           string # defaults to server-*.yml
+metrics_filter:          string # defaults to metrics-*.yml
+metrics_instance_filter: string # defaults to metrics_instances-*.yml
+integrations_filter:     string # defaults to integrations-*.yml
+logs_filter:             string # defaults to logs-*.yml
+traces_filter:           string # defaults to traces-*.yml
 ``` 
 
 ### sources_config
@@ -73,7 +82,7 @@ Configurations are loaded in the order as they are listed below.
 ### Agent
 
 
-Agent template is the standard agent configuration file in its entirety. It is defined by the pattern `agent-*.yml`. Only
+Agent template is the standard agent configuration file in its entirety. The default filter is `agent-*.yml`. Only
 one file is supported. This is processed first then any subsequent configurations found REPLACE the values here, it is
 not additive.
 
@@ -81,38 +90,38 @@ Reference {{< relref "./" >}})
 
 ### Server
 
-Server is defined using the pattern `server-*.yml`, only ONE server file is supported.
+The default filter is `server-*.yml`, only ONE server file is supported.
 
 Reference {{< relref "./server-config.md" >}})
 
 
 ### Metrics
 
-Metrics are defined using the pattern `metrics-*.yml`, only ONE metrics file is supported.
+The default filter is `metrics-*.yml`, only ONE metrics file is supported.
 
 Reference {{< relref "./metrics-config.md" >}})
 
 ### Metric Instances
 
-Metric Instances are defined using the pattern `metrics_instances-*.yml`. Any metric instances are appended to the instances defined in Metrics above. Any number of metric instance files are supporter.
+The default filter is `metrics_instances-*.yml`. Any metric instances are appended to the instances defined in Metrics above. Any number of metric instance files are supporter.
 
 Reference {{< relref "./metrics-config.md#metrics_instance_config" >}}) in the metrics instance
 
 
 ### Integrations
 
-Integrations are defined using the pattern `integrations-*.yml`, these support more than one file, and multiple integrations can be defined in a file. Do not assume any order of loading for integrations. For any integration that is a singleton, loading multiple of those will result in an error. 
+The default filter is `integrations-*.yml`, these support more than one file, and multiple integrations can be defined in a file. Do not assume any order of loading for integrations. For any integration that is a singleton, loading multiple of those will result in an error. 
 
 Reference {{< relref "./integrations/" >}})
 
 ### Traces
 
-Traces are defined using the pattern `traces-*.yml`. This supports ONE file.
+The default filter is `traces-*.yml`. This supports ONE file.
 
 Reference {{< relref "./traces-config.md" >}})
 
 ### Logs
 
-Logs are defined using the pattern `logs-*.yml`. This supports ONE file.
+The default filter is `logs-*.yml`. This supports ONE file.
 
 Reference {{< relref "./logs-config.md" >}})
--- a/docs/configuration/dynamic-config.md
+++ b/docs/configuration/dynamic-config.md
@@ -1,6 +1,6 @@
-# Dynamic Configuration - BETA
+# Dynamic Configuration - Experimental
 
-**This is BETA and subject to change at anytime, feedback is much appreciated**
+**This is experimental and subject to change at anytime, feedback is much appreciated. This is a feature that MAY NOT make it production.**
 
 Dynamic Configuration is the combination of two things:
 
@@ -16,17 +16,18 @@ on what functions are available.
 
 ## Configuration
 
-Location of the dynamic configuration is used via the feature flag `dynamic-config`, then it will use `-config.file` to
+Location of the dynamic configuration is used via the feature flag `dynamic-config`, then it will use `-config.dynamic-config-path` to
 load the configuration for dynamic configuration.
 
 ```yaml
 # Sources to pull template values 
-[sources: <sources_config>]
+sources: 
+  [- <sources_config>]
 
 # Locations to use searching for templates, the system does NOT look into subdirectories. Follows gomplate schema
 # from [gomplate datasources](https://docs.gomplate.ca/datasources/). File and S3/GCP templates are currently supported
-templates: 
-[ - string ]
+template_paths: 
+  [ - string ]
 
 ``` 
 
@@ -42,7 +43,7 @@ url: string
 
 ## Templates
 
-Note when adding a template you do NOT need to add the type as the top level yaml field. For instance if using traces:
+Note when adding a template you MUST NOT add the type as the top level yaml field. For instance if using traces:
 
 Incorrect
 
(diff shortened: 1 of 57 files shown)


mattdurham:
Requesting a review of the concept, more so than the actual code, since it is all just stubs in this PR.

rfratto:
Is there an example folder/set of configs that uses this? From the docs alone I'm having a hard time envisioning how this ties together.

mattdurham:
I will translate some unit tests into a structure that helps lay it out.

mattdurham:
Uploaded some documentation, all these are legit examples I did with my fork with dynamic configuration. Should give a better overview, there are a lot more I want to add but didn't want to spend forever on the examples. Its much easier to view if you download or look at the repo and not the diff. 

@rfratto @rlankfo @tpaschalis

docs/configuration/dynamic-config.md
```diff
@@ -0,0 +1,118 @@
...
+
+Traces are defined using the pattern `traces-*.yml`. This supports one file.
+
+Reference {{< relref "./traces-config.md" >}})
+
+### Logs
+
+Logs are defined using the pattern `logs-*.yml`. This supports on file.
+
+Reference {{< relref "./logs-config.md" >}})
```
rlankfo, line 118: This naming convention seems foreign to me -- I haven't seen other projects that work this way.  Usually projects with conf.d style configurations will have a main config file (say an agent.yaml in our case) and then a config directory (like conf.d/ or agent.d/)  where all *.conf or *.yaml files are in the directory are imported.  Would something like this be possible instead of relying on naming prefixes?

docs/configuration/dynamic-config.md
```diff
@@ -0,0 +1,118 @@
+# Dynamic Configuration - BETA
+
+**This is BETA and subject to change at anytime, feedback is much appreciated**
+
+Dynamic Configuration is the combination of two things:
+
+* Loading from multiple files
+* Using templates and datasources
+
+Both of these make heavy use of the excellent [gomplate](https://github.com/hairyhenderson/gomplate). The idea goal is
```
rlankfo, line 10:
```suggestion
Both of these make heavy use of the excellent [gomplate](https://github.com/hairyhenderson/gomplate). The goal is
```

docs/cookbook/dynamic-configuration/01_Basics/01_Structure.md
```diff
@@ -0,0 +1,65 @@
...
+
+`docker run -v ${PWD}/:/etc/grafana grafana/agentctl:latest template-parse /etc/grafana/02_config.yml`
+
+## Dynamic Configuration
+
+[config.yml](./01_config.yml)
+
+```yaml
+template_paths:
+  - "file:///etc/grafana/01_assets"
```
rlankfo, line 13: is the file:// scheme required here?

docs/cookbook/dynamic-configuration/01_Basics/01_Structure.md
```diff
@@ -0,0 +1,65 @@
...
+```
+
+Tells the Grafana Agent where to load files from. It is important to note that dynamic configuration does NOT traverse directories. It will look at the directory specified only, if you need more directories then add them to the `template_paths` array.
+
+## Agent
+
+Dynamic Configuration will find the first file matching pattern `agent-*.yml` and load that as the base. You can only have
```
rlankfo, line 20: Is `-config.file` no longer required with `-dynamic-config`?  I'm not a fan of the naming prefixes and to me it makes more sense to pass the main config file to `-config.file` and have a `-config.directory` flag to import additional *.yaml files. (or configure this directory in the main config).

docs/configuration/dynamic-config.md
```diff
@@ -0,0 +1,118 @@
+# Dynamic Configuration - BETA
+
+**This is BETA and subject to change at anytime, feedback is much appreciated**
+
+Dynamic Configuration is the combination of two things:
+
+* Loading from multiple files
+* Using templates and datasources
```
rlankfo, line 8: Are these features going to be tightly coupled? e.g. will users be able to make use of multiple config files w/o enabling the dynamic config feature that parses the gomplate templates? In my mind these seem like they should be separate.

rlankfo (commented):
Looks good to me for the most part, I like the cookbook style recipes.  The naming conventions proposed seem a bit wonky to me though vs just importing all files in a directory for example.

tpaschalis:
Nice work man! Thanks for providing nice the cookbook, it was great to see how we envision that it would work in action! 🚀 
And this looks something that's going to really help with maintaining complicated configs!

Some thoughts:

> Dynamic Configuration will find the first file matching pattern agent-*.yml and load that as the base. You can only have one agent template.
 

First off, it sounded a little weird that we'll only accept the first result of the wildcard (I suppose in lexicographic order) and silently discard the rest. ​I'm not sure which is the best solution since we have both files that are additive and ones that are not. Have we given any thought to erroring out when there are more than one matching files, or being able to add a matcher for what's in `*`? 


Secondly, do we explicitly fill in with every available field and add empty values as default? Is that how for example [01_Basics/02_assets/final.yml](https://github.com/grafana/agent/blob/doc_and_settings/docs/cookbook/dynamic-configuration/01_Basics/02_assets/final.yml) came to be larger than its inputs? Where do we get these default values from?

Finally, nice one with forking gomplate to query EC2; much simpler than I thought it would be! In order to be able to query URL-based datasources (eg. GCP Buckets or Azure Blobs) with authentication, will we need something similar?

mattdurham:
@tpaschalis in regards to the final.yaml, it is the result of our unmarshaling of the config which includes all the fields and defaults. It will error out if it finds multiple files that match: agent, server, metrics, traces, or logs. Will update the documentation to reflect that. The cloud provider queries will use the default auth which is generally environment variables or profiles (AWs/GCP, not used Azure much so unsure).

docs/configuration/dynamic-config.md
mattdurham: It's definitely possible, we would need to run all possible config types loaded on each file to see what it actually contains. They would be in memory at that point so the cost should be limited.

docs/cookbook/dynamic-configuration/01_Basics/01_Structure.md
mattdurham: I can change this here in the template_paths and datasources to default to `file://` if none specified BUT datasources embedded in templates would still need need `file://` and that seems inconsistent.

docs/cookbook/dynamic-configuration/01_Basics/01_Structure.md
mattdurham: If you are using dynamic configuration then `-config-file` points to the config file that handles the dynamic configuration, basically a list of folders to scan and any datasources you want to define. We could have `-config.directory` but you would either need to allow an array of them and also additional flags to specify datasources. 

You also do not need to specify an `agent-*.yml` at all, only import the parts you need.

docs/configuration/dynamic-config.md
mattdurham: The templating is optional, if not needed inside the config files then you don't include the template markup. I do think it is integral to the feature though, ie if they were separate PRs to keep the code review easier that's fine but releasing together is important.

docs/cookbook/dynamic-configuration/01_Basics/01_Structure.md
rlankfo: Oh I see.  IMO overriding the `-config.file` flag seems strange -- could a dynamic config block be including in the main agent config (what we're loading as `agent-*.yml` according to the proposed design)

mattdurham: The dynamic config block tells the system where to find the `agent-*.yml` if we didn't want to reuse `-config.file` then a new flag could be added. I resued it for simplicity but if it makes it clearer then that's worth it. The dynamic config tells the system where to load the `agent-*.yml` so it needs to be added first. 

Example `/etc/grafana/dynamic-config.yml`
```
template_paths:
  - "file:///etc/grafana/config"
```

The directory `/etc/grafana/config` has one file `agent-1.yml`. 

Running `./agent -config.file=/etc/grafana/dynamic-config.yml -enable-features=dynamic-config,integrations-next` will 

1. Load the configuration at `/etc/grafana/dynamic-config.yml`
2. Search the directory at `/etc/grafana/config`
3. Find the `agent-1.yml`
4. Unmarshal the config and start the agent

docs/cookbook/dynamic-configuration/01_Basics/03_Integrations.md
```diff
@@ -0,0 +1,43 @@
...
+
+```yaml
+redis_exporter_configs:
+  - redis_addr: localhost:6379
+    autoscrape:
+      metric_relabel_configs:
+        - source_labels: [__address__]
+          target_label: "banana"
+          replacement: "apple"
+  - redis_addr: localhost:6380
```
rfratto, line 35: I think this might be a pretty big problem for the "config snippet" concept. I'd argue it's a requirement that you're able to have one file per instantiation of a redis_exporter.

docs/cookbook/dynamic-configuration/01_Basics/03_assets/final.yml
```diff
@@ -0,0 +1,253 @@
+server:
```
rfratto, line 1: Can we manually trim out zero values from the final.yml files so it's not overwhelming to readers? (even if that means it's not synced with the actual output of /-/config anymore)

docs/cookbook/dynamic-configuration/02_Templates/01_Looping.md
```diff
@@ -0,0 +1,42 @@
...
+    scrape_interval: 60s
+    remote_write:
+      - url: https://prometheus-us-central1.grafana.net/api/prom/push
+        basic_auth:
+          username: 12345
+          password: secretpassword
+  configs:
+    - name: default
+  {{ range slice "apple" "banana" "pear" }}
+    - name: {{ . }}
```
rfratto, line 27: Nice! I think we should make it clear what happens if you have metrics instances defined in the base config file and also in templates. (IMO, it should be valid to have both and just append the arrays, if that's not already what happens)

docs/cookbook/dynamic-configuration/02_Templates/01_Looping.md
```diff
@@ -0,0 +1,42 @@
...
+  {{ end }}
```
rfratto, line 28: WDYT about the current limitation that looping can only be done at the base config template level? Is that going to restrict any common use cases?

docs/cookbook/dynamic-configuration/03_Advanced_Datasources/01_config.yml
```diff
@@ -0,0 +1,2 @@
+template_paths:
+  - "file:///Users/mdurham/source/agents/agent_clean/agent/docs/cookbook/dynamic-configuration/03_Advanced_Datasources/01_assets"
```
rfratto, line 2: Should this have mdurham in it? :)

docs/cookbook/dynamic-configuration/01_Basics/01_Structure.md
rlankfo: My thinking is that it makes more sense to do the following:

1. Load the `agent.yaml` config by passing it to `-config.file`.  This would contain the contents of `agent-*.yml` as well as the contents of `dynamic-config.yaml`
2. If the `dynamic-config` feature is enabled, parse the template paths.

This seems more "conventional" to me. It would mean we don't need a separate config when using dynamic config, and we don't have to discover the base agent config file.  I may be missing something here, but does this seem sane to you?

rfratto (approved):
Nice! I'm supportive of exploring the general idea of using templates to support dynamic configuration. 

I have some high-level concerns about relying on templates; I'm worried it might be really hard to debug problems and add a lot of overhead to the mental map you build of your configs. That being said, it's a good idea to start with this anyway. It's easier to implement than the alternatives I can think of, and that'll be really useful to collect feedback and use cases. I'd also love to be wrong about my concerns and see this graduate out of a feature flag :) 

My only "big" concern is making sure that you can have more than one file per specific instantiation of an multiplexed integration (i.e., one file for mysql_exporter against server-a, another for server-b). Splitting up integrations across multiple files aligns with the CRD concept from the operator, and also makes it easier to write instructions that users can copy and paste into new files. 

(I have some language / styling nits that I've left out of the review, I don't think they're important right now as an RFC)

docs/cookbook/dynamic-configuration/01_Basics/01_Structure.md
mattdurham: My thought is you will want to have a base agent config file that itself will be templated/re-used. The benefit of having them separate is allowing you to configure a fleet of agent instances.  

If you have a dozen agent instances running and want to have a base agent that has organization defaults (endpoints, username, password, agent metrics enabled), then you can have a folder with just your `agent-1.yml` and in the dynamic configuration, template_paths you have the directory where your base `agent-1.yml` resides then the individual instances would have different directories they load the specific details of what they want to scrape/observe. The dynamic-configuration template_paths directories could even be templated with environment vars. 

```
template_paths:
  - "s3://agent_configs/base" # contains agent-1.yml
  - "s3://agent_configs/{{agent_purpose}} # contains specific overrides for this agents purpose
  - "s3://agent_configs/{{hostname}} # contains specific overrides for this host
```

If you needed to globally update a value or setting you could do it in the base and when the reload timer (as of yet undiscussed) triggered it would propagate. 

This might be a use case that is not useful or I am overcomplicating it =)

docs/configuration/dynamic-config.md
```diff
@@ -0,0 +1,118 @@
+# Dynamic Configuration - BETA
```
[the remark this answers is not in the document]
mattdurham: This makes a lot of sense and will change verbiage

docs/configuration/dynamic-config.md
```diff
@@ -0,0 +1,118 @@
...
+that as the configuration grows that it can be split it up into smaller segments to allow better readability and handling.
+The configurations cannot be patched in any order and instead are allowed at several levels.
+
+The second goal is to allow the use of templating, functions for [gomplate doc](https://docs.gomplate.ca/) go into detail
+on what functions are available.
+
+## Configuration
+
+Location of the dynamic configuration is used via the feature flag `dynamic-config`, then it will use `-config.file` to
+load the configuration for dynamic configuration.
```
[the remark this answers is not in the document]
mattdurham: Yeah, that is a good point. Right now it needs the enabled-feature flag to change the meaning which ensures no overlap but based on this and @rlankfo feedback will switch to a dedicated flag.

docs/configuration/dynamic-config.md
```diff
@@ -0,0 +1,118 @@
...
+
+```yaml
+# Sources to pull template values 
+[sources: <sources_config>]
```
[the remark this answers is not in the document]
mattdurham: Yes, will correct doc

docs/configuration/dynamic-config.md
```diff
@@ -0,0 +1,118 @@
...
+name: string
+
+# Path to datasource using schema from [gomplate datasources](https://docs.gomplate.ca/datasources/) 
+url: string
+
+```
+
+## Templates
+
+Note when adding a template you do NOT need to add the type as the top level yaml field. For instance if using traces:
```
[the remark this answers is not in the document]
mattdurham: Good point, will correct.

docs/cookbook/dynamic-configuration/01_Basics/03_Integrations.md
mattdurham: You can but you need to include the `redis_exporter_configs` at the top of each file. The above could be split into two separate files with

```
redis_exporter_configs:
  - redis_addr: localhost:6379
    autoscrape:
      metric_relabel_configs:
        - source_labels: [__address__]
          target_label: "banana"
          replacement: "apple"
```

```
redis_exporter_configs: 
- redis_addr: localhost:6380
```

 The templates are additive except singletons, which throws an error. 

Does that address your concern?

rfratto: Yes, that'd be fine 👍

docs/cookbook/dynamic-configuration/01_Basics/03_assets/final.yml
mattdurham: I struggled with this, partly due to laziness and partly since I wanted to match exactly what agentctl dryrun outputs. I tried to call out what should be in the final. I may see if I can add an option to dryrun to compare a default config with the generated and only output the differences.

docs/cookbook/dynamic-configuration/01_Basics/01_Structure.md
rlankfo: Why can't the base still be templated or re-used if you combine the dynamic config?  I think you'd still be able to accomplish what you've described here (e.g. env vars in the template_paths section of the agent.yml) if you combine the dynamic config with the base config. It does sound a bit overcomplicated to me 😅

docs/cookbook/dynamic-configuration/02_Templates/01_Looping.md
mattdurham: It is appended, but I will add more verbiage on what is additive vs what is replaced.

docs/cookbook/dynamic-configuration/01_Basics/03_assets/final.yml
rfratto: 👍 might be a good opportunity for us to add missing `omitempty` all over the place (though it doesn't help with all our dependencies that don't have it)

docs/cookbook/dynamic-configuration/02_Templates/01_Looping.md
mattdurham: I am unsure what you mean by base config template level. Can you elaborate?

docs/cookbook/dynamic-configuration/03_Advanced_Datasources/01_config.yml
mattdurham: Yeah, that one I couldnt do in docker so I cheated and it slipped through. Thanks for seeing

docs/cookbook/dynamic-configuration/02_Templates/01_Looping.md
rfratto: Ah, sorry, I mean that AFAICT you can't iterate inside of an individual template file i.e., `metrics-instance-*.yml`

mattdurham:
I feel the only large outstanding concern is whether to use `file-names` or have a header on each file to determine the config type it should load. 

Filenames would be the existing `agent-*.yml`, 
Header would be adding the "root" element back into the config, so 
```
integrations:
  redis_exporter_configs:
  - redis_addr: localhost
```

then scan the top element to find which configuration to load. (Or at least something similar) The reason I went with filenames is that it enforces figuring out what the file contains when looking in a directory. Else people can name them config1.yml, config2.yml etc. Granted they still can name them in a way that makes it easy to understand in the header case. 

Thoughts? @rfratto @rlankfo

mattdurham:
@rfratto thoughts on removing the server loading and embedding it into the dynamic configuration block. That way it is NOT reloaded.

rfratto:
> I feel the only large outstanding concern is whether to use file-names or have a header on each file to determine the config type it should load.

I'm not opposed to the header so that all files have the same schema, but FWIW my suggestion was instead of 

```yaml
template_paths:
   - "/etc/grafana/01_assets" 
``` 

to have a user explicitly specify: 

```yaml
base_template_path: /etc/grafana/01_assets/agent-1.yml 
metrics_instance_template_paths: 
  - /etc/grafana/01_assets/metrics-instance-*.yml 
integration_template_paths: 
  - /etc/grafana/01_assets/integrations-*.yml 
traces_template_path: /etc/grafana/01_assets/traces.yml
logs_template_path: /etc/grafana/01_assets/logs.yml
```

That way they're in control of the naming and folder structure conventions, and we still know what the contents of the files are expected to be without having to add the header.

mattdurham:
It's a bit more config but solves the singleton problem and allows them to override the pattern. Let me prototype the changes and see how it feels.

mattdurham:
Prototype this and handling some of the nonfile providers is difficult as a single string, so would need to be the below, which feels verbose. 
```
base_template_path:
  url: s3:///mybucket/
  filter: integrations-*.yml
```

Alternative would be to have the template paths as is BUT allow the override of filter, so by default would use the ones listed above but allow override for each filter. 

```
template_paths:
   - "/etc/grafana/01_assets" 
integrations_filter:  myspecific_filter_integrations-*.yml
```

PR #1309, opened 2022-01-24
Remove reflect, remove tests that use test configs, comment out tests…

mattdurham:
This is part 1 of the codegen, instead of one giant PR smaller ones. This one involves the removal of the reflect sections of marshalling the config. In particular, removes any tests that arent likely to carry over, comments out a few that will need to be re-enabled once done, and removes any of the reflect code used for marshaling.


pkg/integrations/v2/metricsutils/versionshim.go
```diff
@@ -15,20 +15,13 @@ import (
 	"github.com/grafana/agent/pkg/util"
 )
 
-// CreateShim creates a shim between the v1.Config and v2.Config. The resulting
-// config is NOT registered.
-func CreateShim(before v1.Config, common common.MetricsConfig) (after v2.UpgradedConfig) {
-	return &configShim{orig: before, common: common}
-}
```
rfratto, line 22: Should this be removed at this stage? We might need it later

pkg/integrations/v2/register.go
```diff
@@ -1,85 +1,5 @@
...
-
-// UpgradedConfig is a v2 Config that was constructed through a legacy
-// v1.Config. It allows unwrapping to retrieve the original config for
-// the purposes of marshaling or unmarshaling.
-type UpgradedConfig interface {
-	Config
-
-	// LegacyConfig returns the old v1.Config.
-	LegacyConfig() (v1.Config, common.MetricsConfig)
-}
```
rfratto, line 81: Should these be kept until we know we definitely don't need them later on?

rfratto (approved):

pkg/integrations/v2/metricsutils/versionshim.go
mattdurham: If needed later, then I can re-add. If they are needed then they will likely take on a slightly different form.

pkg/integrations/v2/register.go
mattdurham: If needed later, then I can re-add. If they are needed then they will likely take on a slightly different form.

PR #1316, opened 2022-01-26
Stop appending duplicate exemplars

### LINKED ISSUE #1282: Metrics: stop writing duplicate exemplars
@bboreham noticed that 95% of exemplars being sent by Grafana Agent are duplicates, each costing some amount of disk I/O on the receiving side. 

Prometheus supports a [circular buffer](https://github.com/prometheus/prometheus/blob/4855a0c067e25df969debbe880448164116f625e/tsdb/exemplar.go#L118) for storing exemplars in memory. This can be utilized during validation of an exmplar to ignore existing exemplars, preventing duplicate writes. 

cc @mapno

tpaschalis:
#### PR Description 
This PR attempts to stop the Grafana Agent from sending duplicate exemplars, to avoid incurring extra I/O costs to the receiving side.

To achieve that, it adds a memExemplar struct to keep exemplars in memory. It's used to cache the latest exemplar (for series that have them) in a new field of stripeSeries, utilizing the same mechanism to distribute the lock contention.

This fixes #1282 

#### Notes to the Reviewer
- I sketched out this with a couple of other approaches, either moving this one level up as a separate entry in the `Storage` struct, or in a separate `exemplars.go` file next to `series.go`. I didn't really like either of those options, I feel that this might be an acceptable solution comparing the added complexity with the benefit of keeping the code in one place
- Resources-wise, this will allocate defaultStripeSize pointers on startup (~130KB). Each series that has at least one exemplar appended will incur some move overhead (95 bytes on avg, 328 bytes on worst case) as an extra heap allocation.
- Running a benchmark that includes appending a valid exemplar (with its timestamp set to b.N in the benchmark loop) to gauge the performance of `AppendExemplar` before and after this change gives us

```
name              old time/op    new time/op    delta
AppendExemplar-8     255ns ±21%     332ns ±16%  +30.32%  (p=0.000 n=10+10)

name              old alloc/op   new alloc/op   delta
AppendExemplar-8      805B ± 4%      872B ± 7%   +8.37%  (p=0.001 n=10+10)

name              old allocs/op  new allocs/op  delta
AppendExemplar-8      2.00 ± 0%      3.00 ± 0%  +50.00%  (p=0.000 n=10+10)
```

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
 - [BUGFIX] Operator: Add missing proxy_url field from generated remote_write
   configs. (@rfratto)
 
+- [ENHANCEMENT] The agent no longer appends duplicate exemplars. (@tpaschalis)
+
 # v0.22.0 (2022-01-13)
 
 This release has deprecations. Please read [DEPRECATION] entries and consult
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 - [ENHANCEMENT] integrations-next: Add `extra_labels` to add a custom set of
   labels to integration targets. (@rfratto)
 
+- [ENHANCEMENT] The agent no longer appends duplicate exemplars. (@tpaschalis)
+
 - [BUGFIX] Fixed issue where Grafana Agent may panic if there is a very large
   WAL loading while old WALs are being deleted or the `/agent/api/v1/targets`
   endpoint is called. (@tpaschalis)
@@ -15,8 +17,6 @@
   configs. (@rfratto)
 
 - [BUGFIX] Honor the specified log format in the traces subsystem (@mapno)
-
-- [ENHANCEMENT] The agent no longer appends duplicate exemplars. (@tpaschalis)
 
 # v0.22.0 (2022-01-13)
 
--- a/pkg/metrics/wal/series.go
+++ b/pkg/metrics/wal/series.go
@@ -84,6 +84,12 @@ func (m seriesHashmap) del(hash uint64, ref uint64) {
 	}
 }
 
+type memExemplar struct {
+	ts     int64
+	value  float64
+	labels labels.Labels
+}
+
 const (
 	// defaultStripeSize is the default number of entries to allocate in the
 	// stripeSeries hash map.
@@ -97,10 +103,11 @@ const (
 //
 // This code is copied from the Prometheus TSDB.
 type stripeSeries struct {
-	size   int
-	series []map[uint64]*memSeries
-	hashes []seriesHashmap
-	locks  []stripeLock
+	size      int
+	series    []map[uint64]*memSeries
+	hashes    []seriesHashmap
+	exemplars []map[uint64]*memExemplar
+	locks     []stripeLock
 }
 
 type stripeLock struct {
@@ -112,10 +119,11 @@ type stripeLock struct {
 func newStripeSeries() *stripeSeries {
 	stripeSize := defaultStripeSize
 	s := &stripeSeries{
-		size:   stripeSize,
-		series: make([]map[uint64]*memSeries, stripeSize),
-		hashes: make([]seriesHashmap, stripeSize),
-		locks:  make([]stripeLock, stripeSize),
+		size:      stripeSize,
+		series:    make([]map[uint64]*memSeries, stripeSize),
+		hashes:    make([]seriesHashmap, stripeSize),
+		exemplars: make([]map[uint64]*memExemplar, stripeSize),
+		locks:     make([]stripeLock, stripeSize),
 	}
 
 	for i := range s.series {
@@ -123,6 +131,9 @@ func newStripeSeries() *stripeSeries {
 	}
 	for i := range s.hashes {
 		s.hashes[i] = seriesHashmap{}
+	}
+	for i := range s.exemplars {
+		s.exemplars[i] = map[uint64]*memExemplar{}
 	}
 	return s
 }
@@ -171,6 +182,10 @@ func (s *stripeSeries) gc(mint int64) map[uint64]struct{} {
 			delete(s.series[i], series.ref)
 			s.hashes[j].del(seriesHash, series.ref)
 
+			// Since the series is gone, we'll also delete
+			// the latest stored exemplar.
+			delete(s.exemplars[i], series.ref)
+
 			if i != j {
 				s.locks[j].Unlock()
 			}
@@ -213,6 +228,24 @@ func (s *stripeSeries) set(hash uint64, series *memSeries) {
 	i = series.ref & uint64(s.size-1)
 	s.locks[i].Lock()
 	s.series[i][series.ref] = series
+	s.locks[i].Unlock()
+}
+
+func (s *stripeSeries) getLatestExemplar(id uint64) *memExemplar {
+	i := id & uint64(s.size-1)
+
+	s.locks[i].RLock()
+	exemplar := s.exemplars[i][id]
+	s.locks[i].RUnlock()
+
+	return exemplar
+}
+
+func (s *stripeSeries) setLatestExemplar(id uint64, exemplar *memExemplar) {
+	i := id & uint64(s.size-1)
+
+	s.locks[i].Lock()
+	s.exemplars[i][id] = exemplar
 	s.locks[i].Unlock()
 }
 
--- a/pkg/metrics/wal/series.go
+++ b/pkg/metrics/wal/series.go
@@ -239,8 +239,11 @@ func (s *stripeSeries) getLatestExemplar(id uint64) *exemplar.Exemplar {
 func (s *stripeSeries) setLatestExemplar(id uint64, exemplar *exemplar.Exemplar) {
 	i := id & uint64(s.size-1)
 
+	// Make sure that's a valid series id and record its latest exemplar
 	s.locks[i].Lock()
-	s.exemplars[i][id] = exemplar
+	if s.series[i][id] != nil {
+		s.exemplars[i][id] = exemplar
+	}
 	s.locks[i].Unlock()
 }
 
--- a/pkg/metrics/wal/series.go
+++ b/pkg/metrics/wal/series.go
@@ -3,6 +3,7 @@ package wal
 import (
 	"sync"
 
+	"github.com/prometheus/prometheus/pkg/exemplar"
 	"github.com/prometheus/prometheus/pkg/intern"
 	"github.com/prometheus/prometheus/pkg/labels"
 )
@@ -84,12 +85,6 @@ func (m seriesHashmap) del(hash uint64, ref uint64) {
 	}
 }
 
-type memExemplar struct {
-	ts     int64
-	value  float64
-	labels labels.Labels
-}
-
 const (
 	// defaultStripeSize is the default number of entries to allocate in the
 	// stripeSeries hash map.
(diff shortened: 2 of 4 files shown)


pkg/metrics/wal/series.go
```diff
@@ -112,10 +119,11 @@ type stripeLock struct {
...
 	stripeSize := defaultStripeSize
 	s := &stripeSeries{
-		size:   stripeSize,
-		series: make([]map[uint64]*memSeries, stripeSize),
-		hashes: make([]seriesHashmap, stripeSize),
-		locks:  make([]stripeLock, stripeSize),
+		size:      stripeSize,
+		series:    make([]map[uint64]*memSeries, stripeSize),
+		hashes:    make([]seriesHashmap, stripeSize),
+		exemplars: make([]map[uint64]*memExemplar, stripeSize),
```
tpaschalis, line 125: It's only ~130KB, but still we could preallocate a smaller slice here, maybe stripeSize/10?

pkg/metrics/wal/series.go
```diff
@@ -97,10 +103,11 @@ const (
 //
 // This code is copied from the Prometheus TSDB.
```
mattdurham, line 104: nit: Add a comment that this is inspired from the Prometheus TSDB

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@
 
 - [BUGFIX] Honor the specified log format in the traces subsystem (@mapno)
 
+- [ENHANCEMENT] The agent no longer appends duplicate exemplars. (@tpaschalis)
```
mattdurham, line 19: Bump this up to the bottom of the enhancements above, generally try to group categories of changes together.

pkg/metrics/wal/series.go
```diff
@@ -84,6 +84,12 @@ func (m seriesHashmap) del(hash uint64, ref uint64) {
 	}
 }
 
+type memExemplar struct {
+	ts     int64
+	value  float64
+	labels labels.Labels
+}
```
rfratto, line 91: I think we should probably just use `exemplar.Exemplar` here instead of redefining it. That'd help clean up the code a bit (ie you can call `.Equals` for equality checking) and it also means you'd get back the HasTs checking, which seems to be potentially important.

pkg/metrics/wal/wal.go
```diff
@@ -646,6 +646,14 @@ func (a *appender) AppendExemplar(ref uint64, _ labels.Labels, e exemplar.Exempl
 		}
 	}
 
+	// Check for duplicate vs last stored exemplar for this series, and discard those.
+	// Otherwise, record the current exemplar as the latest
+	prevExemplar := a.w.series.getLatestExemplar(ref)
+	if prevExemplar != nil && exemplarsEqual(*prevExemplar, e) {
+		return 0, nil
```
rfratto, line 653: I think this needs to return `ref` to satisfy interface expectations

rfratto (commented):
Nice! Seems reasonable to me. 

Tagging other interested parties for an optional review: @mapno @bboreham 

Also cc people who have done work on exemplars to check this since this will eventually make its way into Prometheus Agent: @cstyan @mdisibio.

CHANGELOG.md
tpaschalis: Thanks, it's done!

pkg/metrics/wal/wal.go
tpaschalis: The intention was to mirror what Prometheus is doing [here](https://github.com/prometheus/prometheus/blob/0df3489275ed4357b12d98e737290992e19915ce/tsdb/head_append.go#L341-L343).

The original interface [mentions](https://github.com/prometheus/prometheus/blob/fe06f16c116a109330cce0bcb61c2c6728ab7227/storage/interface.go#L217) something about cases where the (returned?) reference is 0 and not being used for caching, so I'll look around and see if there's a specific reason they're doing that or I'll change it to return ref.

rfratto: Ah, actually, that's fine then, but let's copy the comment so it's obvious why we're returning 0.

pkg/metrics/wal/series.go
rfratto: It's fine to import Prometheus packages 👍 We only duplicated memSeries because it was too complex to adapt it to support a mode that didn't create chunks on disk.

rfratto:
> Running a benchmark that includes appending a valid exemplar (with its timestamp set to b.N in the benchmark loop) to gauge the performance of AppendExemplar before and after this change 

Love the benchmark stats! We should consider including the benchmark you wrote in the PR so we can run it again for future changes.

tpaschalis:
> One extra thing: we should consider pre-populating `stripeSeries.exemplars` [like we do with `stripeSeries.series`](https://github.com/grafana/agent/blob/459b277c09abe17958ac020b4eb1ef7eff27322f/pkg/metrics/wal/wal.go#L303-L307) on initial startup of the WAL storage.
> 
> If we don't do this, then we might duplicate exemplars once per initialization of the WAL. It's likely not a huge deal, but would be consistent with how we handle other data in the WAL.

@rfratto Right now, we don't in decode exemplars in `loadWAL`. Should we also handle them during initialization in the same way there?

https://github.com/grafana/agent/blob/459b277c09abe17958ac020b4eb1ef7eff27322f/pkg/metrics/wal/wal.go#L283-L285

pkg/metrics/wal/wal_test.go
```diff
@@ -372,6 +372,24 @@ func TestStorage_TruncateAfterClose(t *testing.T) {
...
+
+	s, _ := NewStorage(log.NewNopLogger(), nil, walDir)
+	app := s.Appender(context.Background())
+	sRef, _ := app.Append(0, labels.Labels{{Name: "a", Value: "1"}}, 0, 0)
+
+	e := exemplar.Exemplar{Labels: labels.Labels{{Name: "a", Value: "1"}}, Value: 20, Ts: 10, HasTs: true}
+	for i := 0; i < b.N; i++ {
+		e.Ts = int64(i)
+		_, _ = app.AppendExemplar(sRef, nil, e)
+	}
```
rfratto, line 387: You should probably surround this in calls to `b.StartTimer` and `b.StopTimer` so it doesn't count the time to write to the WAL (or to remove the directory)

pkg/metrics/wal/wal_test.go
```diff
@@ -372,6 +372,24 @@ func TestStorage_TruncateAfterClose(t *testing.T) {
 	require.Error(t, ErrWALClosed, s.Truncate(0))
 }
 
+func BenchmarkAppendExemplar(b *testing.B) {
+	walDir, _ := ioutil.TempDir(os.TempDir(), "wal")
+	defer os.RemoveAll(walDir)
+
+	s, _ := NewStorage(log.NewNopLogger(), nil, walDir)
+	app := s.Appender(context.Background())
```
rfratto, line 380: You might want to consider closing the storage here on exit even if it's not strictly necessary

pkg/metrics/wal/wal_test.go
```diff
@@ -372,6 +372,24 @@ func TestStorage_TruncateAfterClose(t *testing.T) {
...
+	sRef, _ := app.Append(0, labels.Labels{{Name: "a", Value: "1"}}, 0, 0)
+
+	e := exemplar.Exemplar{Labels: labels.Labels{{Name: "a", Value: "1"}}, Value: 20, Ts: 10, HasTs: true}
```
rfratto, line 383: Hmm, I think this is going to inflate the results since every exemplar is always the same. (i.e., with the fix, you're always doing label comparisons, which is going to be slower). 

I'm not sure how dramatic the impact would be though, so it might not be worth changing to generate some new + existing exemplars every time.

rfratto:
> @rfratto Right now, we don't in decode exemplars in loadWAL. Should we also start handling them during initialization there?

Yeah, we'd need to start decoding them if we were going to do this :)

cstyan:
:+1: +1 for importing the prometheus exemplar code and reusing it's `Equals` function.

pkg/metrics/wal/wal_test.go
tpaschalis: After changing to use exemplars.Equal(), I checked the difference between changing the label, the timestamp and the value inside the benchmark loop. Indeed, there's quite some difference there, related to which field we compare first/last.

One thought is that the inflated result may be okay since (at least regarding exemplar comparison) we're getting an upper bound? We could also have three exemplars and modulo choose a different one each time.

```bash
# Comparing Label values in benchmark loop
# e.Label[0].Value = strconv.Itoa(i)
# Extra allocation due to Itoa conversion
name              time/op        alloc/op        allocs/op
AppendExemplar-8  97.6ns ± 4%     56.0B ± 0%       2.00 ± 0%

# Comparing exemplar timestamp
# e.Ts = int64(i)
name              time/op        alloc/op        allocs/op
AppendExemplar-8  144ns ± 2%     335B ± 2%       1.00 ± 0%

# Comparing exemplar values in benchmark loop
# e.Value = float64(i)
name              time/op        alloc/op        allocs/op
AppendExemplar-8  146ns ± 5%     340B ± 1%       1.00 ± 0%
```

pkg/metrics/wal/wal_test.go
tpaschalis: Done, thanks!

pkg/metrics/wal/series.go
tpaschalis: I've changed this in 249600d

pkg/metrics/wal/wal_test.go
rfratto: Hmm, I think this is probably fine as-is and we can return to it if appending exemplars turns out to be a cpu/memory bottleneck.

pkg/metrics/wal/wal.go
```diff
@@ -282,6 +282,8 @@ func (w *Storage) loadWAL(r *wal.Reader) (err error) {
 				decoded <- samples
 			case record.Tombstones, record.Exemplars:
 				// We don't care about decoding tombstones or exemplars
+				// If decide to decode exemplars, we should make sure to prepopulate
```
rfratto, line 285: nit: can you mark this with TODO so it's more discoverable through a search?

pkg/metrics/wal/series.go
```diff
@@ -216,6 +226,24 @@ func (s *stripeSeries) set(hash uint64, series *memSeries) {
...
+	s.locks[i].RUnlock()
+
+	return exemplar
+}
+
+func (s *stripeSeries) setLatestExemplar(id uint64, exemplar *exemplar.Exemplar) {
+	i := id & uint64(s.size-1)
+
+	s.locks[i].Lock()
+	s.exemplars[i][id] = exemplar
```
rfratto, line 243: nit: this has a very low chance of leaking if there's no series for `id`. Maybe this should check to see if `s.series[i][id]` is non-nil?

rfratto (approved):
LGTM after final nits :)

mattdurham (approved):
Looks awesome, love all the benchmarks and details.

pkg/metrics/wal/series.go
tpaschalis: Yeah, I hadn't considered this since supposedly we garbage collect keys from both series and exemplars in one pass, but better to seal this.

PR #1318, opened 2022-01-26
integrations-next: wait for integrations to exit after stopping them

### LINKED ISSUE #1317: integrations-next: Agent should wait for integrations to return before exiting
Discovered while working on https://github.com/grafana/agent/issues/1013

rfratto:
#### PR Description 
Modifies integrations-next to wait for integrations to fully exit before stopping them. This prevents two instances of the same integration (with different configs) from being run concurrently, and also ensures that integrations fully stop when shutting down the agent. 

Fixes #1317. 

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/integrations/v2/controller.go
+++ b/pkg/integrations/v2/controller.go
@@ -60,47 +60,67 @@ func (c *controller) run(ctx context.Context) {
 		exited chan struct{}
 	}
 	var (
+		runningWorkers sync.WaitGroup
+
 		workersMut sync.Mutex
 		workers    = make(map[*controlledIntegration]worker)
 	)
 
-	runWorker := func(ctx context.Context, w worker) {
-		defer func() {
-			// Have a worker remove itself from the map when it exits. Doing this
-			// allows exited integrations to re-start when the config is reloaded.
-			workersMut.Lock()
-			defer workersMut.Unlock()
-			delete(workers, w.ci)
-		}()
-
-		defer close(w.exited)
-
-		w.ci.running.Store(true)
-		defer w.ci.running.Store(false)
-
-		err := w.ci.i.RunIntegration(ctx)
-		if err != nil {
-			level.Warn(c.logger).Log("msg", "integration exited with error", "instance", w.ci.id, "err", err)
-		}
-	}
-
+	// Shut down all workers on shutdown.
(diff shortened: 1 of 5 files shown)


pkg/integrations/v2/controller.go
```diff
@@ -57,64 +57,108 @@ func newController(l log.Logger, cfg controllerConfig, globals Globals) (*contro
...
 // run starts the controller and blocks until ctx is canceled.
 func (c *controller) run(ctx context.Context) {
+	type worker struct {
+		ci     *controlledIntegration
+		stop   context.CancelFunc
+		exited chan struct{}
+	}
+	var (
+		workersMut sync.Mutex
+		workers    = make(map[*controlledIntegration]worker)
```
rfratto, line 67: I keep going back and forth between using a mutex and using a sync.Map here. Using a sync.Map would be easier to follow, but using a mutex makes sure that we can prevent workers from deleting themselves for a while.

pkg/integrations/v2/workers.go
```diff
@@ -0,0 +1,122 @@
...
+
+		// When the integration stops running, we want to free any of our
+		// resources that will notify watchers waiting for the worker to stop.
+		//
+		// Afterwards, we'll block until we remove ourselves from the map; having
+		// an worker remove itself on shutdown allows exited integrations to
+		// re-start when the config is reloaded.
+		defer func() {
+			ci.running.Store(false)
+			close(w.exited)
```
tpaschalis, line 109: We're relying on that close() will implicitly return the "zero value" when we're waiting in the Reload loop above, right? Was caught off-guard as to where the value was sent from, good to know!

tpaschalis (approved):
I'm not super familiar with the integrations code yet, but this LGTM!

mattdurham:
Like the workerpool concept!

mattdurham (approved):
LGTM

pkg/integrations/v2/workers.go
rfratto: Yes 👍 this is a pattern you'll see pretty common to signal a goroutine exited throughout the code + many of our dependencies

PR #1329, opened 2022-01-28
Allow remote_write URL credentials

### LINKED ISSUE #1211: Inline username and password no longer works for remote write after security fix
When using `http://user:password@example.com` for remote_write the password is redacted when generating the client URL for the HTTP client.

The instance.go:199 Clone function takes the unsanitized input and then redacts it in `github.com/prometheus/common/config/http_config.go:127`, the workaround is to separate the username and password into their own fields.

tpaschalis:
#### PR Description 
This is an attempt to re-allow inlined credentials in remote_write URLs.

The underlying issue will be dealt with by metrics V2, and while no issues have been raised by the community yet, it might be worth fixing.

This fixes #1211 

#### Notes to the Reviewer
I checked that when calling /-/config, the secret there is scrubbed.

Also, I kinda abused an existing RetainSecrets test (adding credentials both inline and in a basic_auth block), but I don't believe this small change warrants 40 extra lines of test code.

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@
 
 - [BUGFIX] Fix typo in node_exporter for runit_service_dir. (@mattdurham)
 
+- [BUGFIX] Allow inlining credentials in remote_write url. (@tpaschalis)
+
 # v0.22.0 (2022-01-13)
 
 This release has deprecations. Please read [DEPRECATION] entries and consult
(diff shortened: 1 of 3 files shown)


rfratto (approved):
Thanks!

Issue #1013, opened 2021-10-20
Add a Kubernetes Events integration


hjet:
Add an [eventrouter](https://github.com/heptiolabs/eventrouter)-style integration to Agent that watches for [Events](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#event-v1-events-k8s-io) and potentially other objects of interest (Conditions?), and ships them to a Loki sink.

rfratto:
This is currently being worked on, from what I understand.

hjet:
> This is currently being worked on, from what I understand.

Yes, confirmed :)

PR #1330, opened 2022-01-28
Add K8s eventhandler integration

hjet:
#### PR Description 
This adds an Agent integration that watches for changes to Event objects in a K8s cluster and creates Loki entries that are handed off to Agent's logs subsystem (embedded promtail). It uses integrations subsystem v2.

**Non-goals**:
To start, labels are hardcoded and users will have to use LogQL to do any transformations. In the future, I'll add pipelines etc. to the logs subsystem API so integrations can take advantage of these features as well as relabel_configs, etc.

https://github.com/grafana/agent/issues/1013

#### Notes to the Reviewer
Some of the design decisions here have already been discussed with @rfratto. One of the major ones was the need for a disk cache: upon restart, any informer/controller/etc. will perform a List() operation that will create duplicate shipped entries without some notion of "what's already been shipped." The technique used in this PR was chosen after evaluating other options.

#### PR Checklist
- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated


hjet:
Adding @bboreham as an optional reviewer

docs/configuration/integrations/integrations-next/eventhandler-config.md
```diff
@@ -0,0 +1,300 @@
...
+
+This integration uses Agent's embedded Loki-comptaible `logs` subsystem to ship entries, and a logs client and sink must be configured to use the integration. Please see the sample Agent config below for an example configuration. [Pipelines](https://grafana.com/docs/loki/latest/clients/promtail/pipelines/) and relabel configuration are not yet supported, but these features will be added soon. You should use the `job=eventhandler cluster=...` labels to query your events (you can then use LogQL on top of the result set).
```
rlankfo, line 13:
```suggestion
This integration uses Grafana Agent's embedded Loki-compatible `logs` subsystem to ship entries, and a logs client and sink must be configured to use the integration. Please see the sample Agent config below for an example configuration. [Pipelines](https://grafana.com/docs/loki/latest/clients/promtail/pipelines/) and relabel configuration are not yet supported, but these features will be added soon. You should use the `job=eventhandler cluster=...` labels to query your events (you can then use LogQL on top of the result set).
```

docs/configuration/integrations/integrations-next/eventhandler-config.md
```diff
@@ -0,0 +1,300 @@
...
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        name: grafana-agent-eventhandler
+    spec:
+      terminationGracePeriodSeconds: 10
+      containers:
+      - name: agent
+        image: grafana/agent:dev.eventlogger-781ee75
```
rlankfo, line 267:
```suggestion
        image: grafana/agent:main
```

This probably doesn't matter much but once it's merged in, users can pull the main image.

docs/configuration/integrations/integrations-next/eventhandler-config.md
```diff
@@ -0,0 +1,300 @@
...
+
+  ## K8s informer resync interval (seconds). You should use defaults here unless you are 
+  ## familiar with K8s informers.
+  [informer_resync: <int> | default = 120]
+
+  ## The integration will flush the last event shipped out to disk every flush_interval seconds.
+  [flush_interval: <int> | default = 10]
+
+  ## If you would like to limit events to a given namespace, use this parameter.
+  [namespace: <string>]
```
rlankfo, line 48: Just curious, are the options for namespace filtering currently limited to all or 1?

pkg/integrations/v2/eventhandler/eventhandler.go
```diff
@@ -0,0 +1,437 @@
...
+func (eh *EventHandler) RunIntegration(ctx context.Context) error {
+	var wg sync.WaitGroup
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	// Quick check to make sure logs instance exists
+	if i := eh.LogsClient.Instance(eh.LogsInstance); i == nil {
+		level.Error(eh.Log).Log("msg", "Logs instance not configured", "instance", eh.LogsInstance)
+	}
```
rlankfo, line 341: Is there much point in continuing here? It seems we'll have a panic in `handleEvent` when we try to send the entry later if this is nil.

rlankfo (approved):
LGTM :+1:

docs/configuration/integrations/integrations-next/eventhandler-config.md
hjet: Yes, due to https://github.com/kubernetes/kubernetes/issues/74415 (and others).. if there's sufficient demand for list input I can implement it

pkg/integrations/v2/eventhandler/eventhandler.go
hjet: Missed this, good catch!

PR #1326, opened 2022-01-27
smoke framework refactor

rlankfo:
#### PR Description 

This refactors the agent smoke test in a few ways in preparation for running in k8s cluster:

- Moves a portion of the bash script to a go program that is containerized and can run in-cluster to perform replica scaling for avalanche and agent deployments
- Adds a smoke/ directory in our production libsonnet that can be imported by deployment_tools
- Moves crow/ and smoke/ out of cmd and into tools
- Deploys crow and smoke images to internal registry instead of dockerhub

#### Notes to the Reviewer

* I encourage reviewers to pull these changes down and run the smoke test locally and let me know if you run into any issues that we need to resolve. 
* Most of this has been reviewed in separate PR's going into dev.smoke-framework already.

#### PR Checklist

- [ ] CHANGELOG updated  -- doesn't seem worthy of a changelog entry as this is just smoke test refactoring
- [ ] Documentation added
- [ ] Tests updated

PR #1338, opened 2022-02-01
add notes in smoke readme

rlankfo:
This is in response to some feedback @tpaschalis had in https://github.com/grafana/agent/pull/1326 when attempting to run the smoke test locally. Since we now push crow and smoke images to an internal registry,  users who do not have access will need to build these images locally before running the smoke test.


rfratto:
Thanks!

PR #1340, opened 2022-02-01
operator/hierarchy: stop using field selector when listing Secrets & ConfigMaps

### LINKED ISSUE #1339: Operator: main build fails to reconcile secrets
@stevo-f3 [reported][original-report] in #1271 that the operator is failing to reconcile on secrets:

> 
```
level=info ts=2022-02-01T18:57:31.693552638Z component=controller.grafanaagent msg="Starting workers" workercount=1
level=info ts=2022-02-01T18:57:31.693623302Z component=controller.grafanaagent name=foo namespace=bar msg="reconciling grafana-agent"
level=error ts=2022-02-01T18:57:31.694405371Z component=controller.grafanaagent name=foo namespace=bar msg="unable to build hierarchy" err="failed to discover secrets: failed to find \"/v1, Kind=SecretList\" resource: list failed: Index with name field:metadata.name does not exist"
```

This is likely caused by the ongoing refactoring to help test and maintain the code more easily. 

[original-report]: https://github.com/grafana/agent/issues/1271#issuecomment-1027214694_

rfratto:
#### PR Description 
The initial implementation of hierarchy.KeySelector injected a FieldSelector when listing Secrets and ConfigMaps to immediately return the single object being queried for.

This causes a problem with the client generated by the controller-runtime framework, where the client is wrapped in a cache and field indexer (where only the namespace is indexed by default).

This commit avoids using the field selector and the index lookup. The resulting behavior aligns more closely with discovering other resources in the hierarchy (i.e., ServiceMonitors), where the List call is also insufficient and needs post-processing via Matches to find the final list of resources.

Given the controller-runtime client uses an informer for reads, all relevant Secrets and ConfigMaps are already in-memory anyway, and using the index for a faster List is a bit of an over-optimization at the moment.

Closes #1339. 

#### Notes to the Reviewer
Relevant tests have been updated to use a client which is identical to the client generated by the operator (i.e., one that uses a cache informer & index).

#### PR Checklist

- [x] CHANGELOG updated (N/A: commit introducing bug not released) 
- [x] Documentation added 
- [x] Tests updated

### DIFF
--- a/pkg/operator/hierarchy/selector.go
+++ b/pkg/operator/hierarchy/selector.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -79,7 +78,6 @@ var _ Selector = (*KeySelector)(nil)
 // ApplyToList implements Selector.
 func (ks *KeySelector) ApplyToList(lo *client.ListOptions) {
 	lo.Namespace = ks.Namespace
-	lo.FieldSelector = fields.OneTermEqualSelector("metadata.name", ks.Name)
 }
 
 // Matches implements Selector.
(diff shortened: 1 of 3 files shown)


pkg/operator/hierarchy/selector.go
```diff
@@ -79,7 +78,6 @@ var _ Selector = (*KeySelector)(nil)
 // ApplyToList implements Selector.
 func (ks *KeySelector) ApplyToList(lo *client.ListOptions) {
 	lo.Namespace = ks.Namespace
-	lo.FieldSelector = fields.OneTermEqualSelector("metadata.name", ks.Name)
```
rfratto, line 82: (Note that `Matches` below already performs filtering of the objects so we don't need to do anything other than remove the selector here)

rlankfo (approved):

PR #1345, opened 2022-02-02
Add stale check Github Action

rfratto:
Stalebot doesn't seem to be working anymore, so this PR introduces similar functionality as an action. 

One downside with this: stale labels won't be removed automatically whenever someone comments, and instead will only be removed the next time the stale check runs. I've updated the stale message to try to make it clear that's the new behavior.


mattdurham (approved):
LGTM

.github/workflows/stale.yml
```diff
@@ -0,0 +1,38 @@
...
+          if there is new activity. The issue will be closed in 7 days if there
+          is no new activity.
+
+          Thank you for your contributions!
+        stale-pr-message: >
+          This PR has been automatically marked as stale because it has not had
+          any activity in the past 30 days.
+
+          The next time this stale check runs, the stale label will be removed
+          if there is new activity. The issue will be closed in 7 days if there
```
rfratto, line 31:
```suggestion
          if there is new activity. The PR will be closed in 7 days if there
```

rlankfo (approved):

Issue #1116, opened 2021-11-18
Request: default send_exemplars to true in remote_write config


bboreham:
We want to make [exemplars](https://grafana.com/docs/grafana/latest/basics/exemplars/) easy to use for customers.

There is a setting in `remote_write` which must be on to send them.

Turning it on by default will impact people who are scraping lots of exemplars but don't have a back-end to make use of them.
That doesn't seem like a widespread problem, and those people can always configure it to `false`.

bboreham:
Still needed.

PR #1352, opened 2022-02-02
default send_exemplars to true in remote_write config

rlankfo:
#### PR Description 

initially I was going to make this config update in our dependency on `grafana/prometheus` to a version that defaults send_exemplars to true in remote_write config. See https://github.com/grafana/prometheus/pull/26. Opting to do this in init() function in the agent instead since this isn't something that would end up going upstream.

Closes https://github.com/grafana/agent/issues/1116

#### PR Checklist

- [x] CHANGELOG updated 
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
 
 - [ENHANCEMENT] Added Kubernetes eventhandler integration (@hjet)
 
+- [ENHANCEMENT] Enables sending of exemplars over remote write by default. (@rlankfo)
+
 - [BUGFIX] Fixed issue where Grafana Agent may panic if there is a very large
   WAL loading while old WALs are being deleted or the `/agent/api/v1/targets`
   endpoint is called. (@tpaschalis)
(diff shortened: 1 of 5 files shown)


tpaschalis:
I think two different ways we could achieve this would be a) using the existing init function to set that exported variable to true, or b) setting it in our DefaultGlobalConfig and handling it in ApplyDefaults. 

I don't have a strong opinion and I like the simplicity here, but these options might be worth it if we'd like to depend less on our custom Prometheus fork going forward. WDYT?

mattdurham:
I would like to wait for the prometheus ticket to get merged. Given it is prometheus and our linking to prometheus is tricky already.

mattdurham:
And of course, I hit the wrong button and closed this PR instead of just submitting feedback. Reopened, apologies for email spam.

mattdurham (changes requested):
Would prefer to wait for merge in prometheus fork.

tpaschalis (approved):
LGTM! 👌

bboreham (approved):

mattdurham (approved):
👍

PR #1360, opened 2022-02-04
Code to support dynamic configuration.

mattdurham:
#### PR Description 

Code changes to support Dynamic Configuration

#### PR Checklist

- [X] CHANGELOG updated 
- [X] Documentation added
- [X] Tests updated


pkg/config/dynamicloader_test.go
```diff
@@ -0,0 +1,759 @@
+package config //nolint:golint
```
mattdurham, line 1: Going to look through this today and see what I can move into functions to deduplicate some code.

go.mod
```diff
@@ -421,3 +479,9 @@ replace github.com/jaegertracing/jaeger => github.com/jaegertracing/jaeger v1.27
 
 // Replacement necessary for windows_exporter so that we can use gokit logging and not the old prometheus logging
 replace github.com/leoluk/perflib_exporter v0.1.0 => github.com/grafana/perflib_exporter v0.1.1-0.20211013152516-e37e14fb8b0a
+
+// Required until fork merged upstream
```
rfratto, line 483: nit: Can you comment that this is used for dynamic configuration?

pkg/config/config.go
```diff
@@ -262,9 +262,21 @@ func LoadRemote(url string, expandEnvVars bool, c *Config) error {
 }
 
 // LoadDynamicConfiguration is used to load configuration from a variety of sources using
-// ConfigLoader, this is a templated approach
+// dynamic loader, this is a templated approach
 func LoadDynamicConfiguration(url string, _ bool, c *Config, fs *flag.FlagSet) error {
```
rfratto, line 266: nit: should expandEnv be used? (Is it possible for it to be used?)

pkg/config/config.go
```diff
@@ -311,10 +323,9 @@ func Load(fs *flag.FlagSet, args []string) (*Config, error) {
 		if features.Enabled(fs, featRemoteConfigs) {
 			return LoadRemote(url, expand, c)
 		}
-		if features.Enabled(fs, featDynamicConfig) {
-			if !features.Enabled(fs, featIntegrationsNext) {
-				panic("integrations-next must be enabled for dynamic configuration to work")
-			}
+		if features.Enabled(fs, featDynamicConfig) && !features.Enabled(fs, featIntegrationsNext) {
+			panic("integrations-next must be enabled for dynamic configuration to work")
```
rfratto, line 327: nit: I'd prefer only showing a panic to a user if it's from a logic issue that should never happen (ie not the user's fault). For a misconfiguration, I'd like for us to return errors instead.

pkg/config/config.go
```diff
@@ -327,12 +338,15 @@ func load(fs *flag.FlagSet, args []string, loader func(string, bool, *Config) er
...
-		file            string
-		configExpandEnv bool
+		printVersion             bool
+		file                     string
+		dynamicConfigurationPath string
+		configExpandEnv          bool
 	)
 
 	fs.StringVar(&file, "config.file", "", "configuration file to load")
+	fs.StringVar(&dynamicConfigurationPath, "config.dynamic-config-path", "", "dynamic configuration path; supports file:// and s3://")
```
rfratto, line 348: nit: Can we add to the help text to say that these two are mutually exclusive?

pkg/config/config.go
```diff
@@ -327,12 +338,15 @@ func load(fs *flag.FlagSet, args []string, loader func(string, bool, *Config) er
 	var (
 		cfg = DefaultConfig
 
-		printVersion    bool
-		file            string
-		configExpandEnv bool
+		printVersion             bool
+		file                     string
+		dynamicConfigurationPath string
+		configExpandEnv          bool
```
rfratto, line 344: nit: use `config` or `configuration` for both of these for consistency, but don't mix :)

pkg/config/config.go
```diff
@@ -347,7 +361,14 @@ func load(fs *flag.FlagSet, args []string, loader func(string, bool, *Config) er
 		os.Exit(0)
 	}
 
-	if file == "" {
+	if features.Enabled(fs, featDynamicConfig) {
+		if dynamicConfigurationPath == "" {
+			return nil, fmt.Errorf("-config.dynamic-config-path flag required when using dynamic configuration")
+		} else if err := loader(dynamicConfigurationPath, configExpandEnv, &cfg); err != nil {
+			return nil, fmt.Errorf("error loading dynamic configuration file %s: %w", dynamicConfigurationPath, err)
+		}
```
rfratto, line 369: nit: hmm, I don't know if it should be _required_ when the feature flag is used. When it graduates out of being a feature flag, we wouldn't want it to be required then either. 

Can we make it optional, and let the user decide between normal/dynamic config?

pkg/config/dynamicloader.go
```diff
@@ -0,0 +1,495 @@
...
+	cfg    *LoaderConfig
+}
+
+// NewDynamicLoader instantiates a new DynamicLoader
+func NewDynamicLoader() (*DynamicLoader, error) {
+
+	return &DynamicLoader{
+		loader: nil,
+		mux:    newFSProvider(),
+		cfg:    nil,
```
rfratto, line 45: nit: avoid explicitly setting fields to their zero value to keep things a little tidier

pkg/config/dynamicloader.go
```diff
@@ -0,0 +1,495 @@
...
+// DynamicLoader is used to load configs from a variety of sources and squash them together.
+// This is used by the dynamic configuration feature to load configurations from a set of templates and then run them through
+// gomplate producing an end result.
+type DynamicLoader struct {
+	loader *loader.ConfigLoader
+	mux    fsimpl.FSMux
+	cfg    *LoaderConfig
+}
+
+// NewDynamicLoader instantiates a new DynamicLoader
```
rfratto, line 39: nit: I'm probably guilty of this too, but IMO godocs should be complete sentences ending in a period

pkg/config/dynamicloader.go
```diff
@@ -0,0 +1,495 @@
...
+	} else if strings.HasPrefix(path, "s3://") {
+		blobURL, err := url.Parse(path)
+		if err != nil {
+			return err
+		}
+		buf, err = data.ReadBlob(*blobURL)
+		if err != nil {
+			return err
+		}
+	}
```
rfratto, line 87: nit: If the else case isn't relevant, can you add one anyway with a comment explaining why it's a no-op?


(it might be easier to read as a switch statement, but if/else if/else is fine too)

pkg/config/dynamicloader.go
```diff
@@ -0,0 +1,495 @@
...
+
+func (c *DynamicLoader) processServer() (*server.Config, error) {
+	filter := "server-*.yml"
+	if c.cfg.ServerFilter != "" {
+		filter = c.cfg.ServerFilter
+	}
+
+	for _, path := range c.cfg.TemplatePaths {
+		result, err := c.generateConfigsFromPath(path, filter, func() interface{} {
+			return &server.Config{}
```
rfratto, line 200: Do we need to apply defaults here?

pkg/config/dynamicloader.go
```diff
@@ -0,0 +1,495 @@
...
+	}
+	fBytes := make([]byte, stats.Size())
+	bytesRead, err := file.Read(fBytes)
+	if bytesRead == 0 {
+		return nil, errors.New("no bytes read")
+	}
+
+	if err != nil && err != io.EOF {
+		return nil, err
+	}
```
rfratto, line 401: All of this can be replaced with a call to fs.ReadFile :) 

```suggestion
	fBytes, err := fs.ReadFile(handler, f.Name())
	if err != nil {
		return nil, err	
	}  
```

(ditto for the other handlers)

pkg/integrations/v2/metricsutils/versionshim.go
```diff
@@ -18,49 +18,49 @@ import (
 // CreateShim creates a shim between the v1.Config and v2.Config. The resulting
```
rfratto, line 18: I'd like to avoid exposing the things here; see my tips in the test file for how we can do that.

pkg/integrations/v2/register.go
```diff
@@ -419,3 +430,14 @@ func replaceYAMLTypeError(err error, oldTyp, newTyp reflect.Type) error {
...
+// UnmarshalYamlToExporters attempts to convert the contents of yaml string into a set of exporters and then return
+// those configurations.
+func UnmarshalYamlToExporters(contents string) ([]Config, error) {
+	o := &SubsystemOptions{}
+	err := yaml.Unmarshal([]byte(contents), o)
+	if err != nil {
+		return nil, err
+	}
+	return o.Configs, nil
+}
```
rfratto, line 443: nit: I probably think this function can be defined unexported in the dynamic loader to keep things tidy.

rfratto (commented):
Nice work! I don't have a lot of high level concerns at this point, though I'd like if it was easier to follow how unloading works, and making sure the deferred unmarshaling for integrations-next only happens in exactly one place.

cmd/agent/entrypoint.go
```diff
@@ -73,6 +73,14 @@ func NewEntrypoint(logger *util.Logger, cfg *config.Config, reloader Reloader) (
 		level.Info(ep.log).Log("msg", "reload server started", "url", reloadURL)
 	}
 
+	// If there is no default instance and v2 integrations exist then create "default"
+	if len(cfg.Metrics.Configs) == 0 && !cfg.Integrations.IsZero() && cfg.Integrations.IsV2() {
+		if cfg.Metrics.Configs == nil {
+			cfg.Metrics.Configs = make([]instance.Config, 0)
+		}
+		cfg.Metrics.Configs = append(cfg.Metrics.Configs, instance.Config{Name: "default"})
+	}
```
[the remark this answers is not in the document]
mattdurham: What do you mean by "might confuse them when they don't see metrics appearing anywhere"? I would assume if they are setting the metrics instance target they are knowledgeable enough to know they need to add the instance. I think the default instance for most users is they run it, agent warns they need to add a default instance, they google it and eventually ad the lines.

If they are going the easy path then it just works with less config. I can def remove it and add a separate pr later to discuss it.

pkg/config/config.go
mattdurham: gomplate itself has support for environment var substitution so didn't want conflicts between the two.

pkg/integrations/v2/register.go
mattdurham: 👍

pkg/config/config.go
mattdurham: 👍

pkg/config/config.go
rfratto: Should we return an error if expand-env is set with dynamic configs then?

cmd/agent/entrypoint.go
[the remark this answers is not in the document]
mattdurham: Removed

pkg/integrations/v2/metricsutils/versionshim.go
mattdurham: I change the struct from being exposed but I still need to expose the fields for marshaling, seem reasonable?

pkg/config/config.go
```diff
@@ -262,9 +262,24 @@ func LoadRemote(url string, expandEnvVars bool, c *Config) error {
 }
 
 // LoadDynamicConfiguration is used to load configuration from a variety of sources using
-// ConfigLoader, this is a templated approach
-func LoadDynamicConfiguration(url string, _ bool, c *Config, fs *flag.FlagSet) error {
-	// will be filled in by future pr
+// dynamic loader, this is a templated approach
+func LoadDynamicConfiguration(url string, expandvar bool, c *Config, fs *flag.FlagSet) error {
+	if expandvar {
```
tpaschalis, line 267: I may be missing some context here about the `loader func` below, but do we need this argument if we're always going to exit? 

Could this failure be handled more explicitly at the caller's site, where we also list other failure modes?

pkg/config/dynamicconfig.go
```diff
@@ -0,0 +1,25 @@
...
+type LoaderConfig struct {
+	// Sources is used to define sources for variables using gomplate
+	Sources []Datasource `yaml:"datasources"`
+
+	// TemplatePaths is the "directory" to look for templates in, they will be found and matched to configs but various
+	// naming conventions. They can be S3/gcp, or file based resources. The directory structure is NOT walked.
+	TemplatePaths []string `yaml:"template_paths"`
+
+	AgentFilter           string `yaml:"agent_filter,omitempty"`
+	ServerFilter          string `yaml:"sever_filter,omitempty"`
```
tpaschalis, line 13: s/sever/server

mattdurham: Good catch!

pkg/config/dynamicloader_test.go
```diff
@@ -0,0 +1,612 @@
...
+		TemplatePaths: []string{fileFS},
+	}
+	cmf := generateLoader(t, loaderCfg)
+	configs, err := cmf.processIntegrations()
+	assert.NoError(t, err)
+	assert.Len(t, configs, 2)
+	for _, cfg := range configs {
+		switch v := cfg.(type) {
+		default:
+			fmt.Printf("unexpected type %T", v)
```
tpaschalis, line 194: "Useless comment of the week" candidate here, but did we mean for this to t.Errorf if a different type was encountered?

mattdurham: t.Errorf is far better! Thanks

pkg/integrations/v2/metricsutils/versionshim.go
rfratto: Hmm, I'm not sure I understand why it's needed. Weren't they being marshaled before to /-/config without the fields being exported?

pkg/config/dynamicloader_test.go
```diff
@@ -0,0 +1,612 @@
+//nolint:golint,goconst
```
rfratto, line 1: Ditto with my old comment about this file being quite long and maybe having an opportunity to make the set of assertions smaller by finding another way to make comparisons

pkg/config/dynamicconfig.go
```diff
@@ -0,0 +1,25 @@
+package config
```
rfratto, line 1: Hmm, are we creating enough new types here that it makes sense to create a new package for dynamic config? `pkg/config/dynconfig` or something?

rfratto (commented):
Gave a second pass, but there's some outstanding comments from my previous review that I still stand by

pkg/integrations/v2/metricsutils/versionshim.go
mattdurham: Good call here, found the underlying issue with a bug in my code where the v1config was not nil so it wasn't exported correctly.

cmd/agentctl/main.go
```diff
@@ -474,6 +477,41 @@ config that may be used with this agent.`,
...
 
+func templateDryRunCmd() *cobra.Command {
+
+	cmd := &cobra.Command{
+		Use:   "template-parse [directory]",
+		Short: "dry run dynamic configuration",
+		Long:  `This will load the dynamic configuration, load configs, run templates and then output the full config as yaml`,
+		Args:  cobra.ExactArgs(1),
+
+		RunE: func(_ *cobra.Command, args []string) error {
```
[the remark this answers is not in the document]
mattdurham: Sounds good, I am not entirely happy with the verbiage around the command but I think it's sufficient for now.

pkg/config/dynamicloader_test.go
mattdurham: Thoughts on splitting the tests between config loading and template? I also removed some redundant tests. Splitting/removing redundant tests means one file is 175 lines and the other is 350 lines

rfratto: Seems reasonable 👍

pkg/config/dynamicconfig.go
mattdurham: Though its only 3 new types: Loader, LoaderConfig, Datasource

pkg/config/config.go
mattdurham: Do you mean to use the usage of the `--config.dynamic-config-path` as the implicit opt in?

rfratto: Yeah, i.e., still let people use `-config.file` even if the feature flag is enabled, but using `-config.dynamic-config-path` requires the feature flag to be set.

mattdurham: Would the behavior in that scenario be that the normal file will be run through the template engine?

rfratto: IMO, I think there's two types of feature flags:

1. Enables optional new behavior (like `remote-configs`)
2. Switches out an implementation for a new one (like `integrations-next`)

I feel like this falls under the first bucket, where the intent is after graduating, it becomes an optional feature, so it might feel weird for users if during the feature flag phase, it's required when the feature is set. But it'd also be unlikely for someone to set the feature flag without already having the intent to use this functionality anyway. 

I don't really feel too strongly about this, and it's something we can fix if someone has a problem.

mattdurham:
Updated code based on feedback and pining @rfratto for another round of feedback.

pkg/config/config.go
mattdurham: Sounds reasonable, if it becomes an issue more than happy to look at the feedback and change accordingly.

pkg/config/dynamicloader.go
mattdurham: Switched to switch with error if not s3 or file

pkg/config/dynamicloader.go
```diff
@@ -0,0 +1,495 @@
...
+
+// ProcessConfigs loads the configurations in a predetermined order to handle functioning correctly.
+func (c *DynamicLoader) ProcessConfigs(cfg *Config, fs *flag.FlagSet) error {
+	if c.cfg == nil {
+		return errors.New("LoadConfig or LoadConfigByPath must be called")
+	}
+	var returnErr error
+	err := c.processAgent(cfg)
+	if err != nil {
+		returnErr = multierror.Append(returnErr, err)
```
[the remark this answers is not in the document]
mattdurham: Created a wrapper that checks for the err argument, since multierror.append creates a new error

pkg/config/dynamicloader.go
```diff
@@ -0,0 +1,495 @@
...
+		cfg.Traces = *traceConfigs
+	}
+
+	integrations, err := c.processIntegrations()
+	if err != nil {
+		returnErr = multierror.Append(returnErr, err)
+	}
+	// If integrations havent already been defined then we need to do
+	// some setup
+	if cfg.Integrations.configV2 == nil {
```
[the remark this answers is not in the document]
mattdurham: If you don't define an Agent file with integrations then this will be nil. Or conversely, if you don't define an Agent file at all.

pkg/config/dynamicloader.go
```diff
@@ -0,0 +1,495 @@
...
+
+	err = cfg.Validate(fs)
+	if err != nil {
+		returnErr = multierror.Append(returnErr, err)
+	}
+	return returnErr
+}
+
+func (c *DynamicLoader) processAgent(cfg *Config) error {
+	filter := "agent-*.yml"
```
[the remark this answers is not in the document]
mattdurham: Done and made the code cleaner!

pkg/config/dynamicloader.go
```diff
@@ -0,0 +1,495 @@
...
+
+	for _, path := range c.cfg.TemplatePaths {
+		result, err := c.generateConfigsFromPath(path, filter, func() interface{} {
+			return cfg
+		}, c.handleAgentMatch)
+		if err != nil {
+			return err
+		}
+		if len(result) > 1 {
+			return errors.New("multiple agent configurations found")
```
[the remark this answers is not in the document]
mattdurham: 👍

pkg/config/dynamicloader.go
```diff
@@ -0,0 +1,495 @@
...
+	}
+	for _, cfg := range configs {
+		if _, ok := singletonCheck[cfg.Name()]; ok {
+			singletonCheck[cfg.Name()]++
+			if singletonCheck[cfg.Name()] > 1 {
+				// TODO (mattdurham) would be nice if it said where they were found
+				returnError = multierror.Append(returnError, fmt.Errorf("found multiple instances of singleton integration %s", cfg.Name()))
+			}
+		}
+	}
```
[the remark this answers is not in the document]
mattdurham: Done and good call. Updated tests to reflect the move

pkg/config/dynamicloader.go
mattdurham: Jeez, yeah thats a ton cleaner.

pkg/config/dynamicloader.go
```diff
@@ -0,0 +1,495 @@
...
+	// Parse the template
+	processedConfigString, err := c.loader.GenerateTemplate("", fString)
+	if err != nil {
+		return nil, err
+	}
+	cfg := configMake()
+	err = yaml.Unmarshal([]byte(processedConfigString), cfg)
+	if err != nil && err != io.EOF {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
mattdurham: I dont see a way that inputs could trigger an EOF.  Removing eof check.

pkg/config/dynamicloader.go
```diff
@@ -0,0 +1,495 @@
...
+	// Parse the template
+	processedConfigString, err := c.loader.GenerateTemplate("", fString)
+	if err != nil {
+		return nil, err
+	}
+	cfg, err := v2.UnmarshalYamlToExporters(processedConfigString)
+	if cfg == nil || err != nil {
+		return nil, err
+	}
+	// TODO (mattdurham) there has to be a better way to handle this conversion
```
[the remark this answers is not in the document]
mattdurham: 😢

pkg/integrations/v2/controller.go
```diff
@@ -198,6 +200,23 @@ func (c *controller) UpdateController(cfg controllerConfig, globals Globals) err
 	c.mut.Lock()
 	defer c.mut.Unlock()
 
+	var returnError error
+	// Check to ensure no duplicate singleton exist
+	singletonCheck := make(map[string]interface{})
```
rfratto, line 205: Can you add this to the CHANGELOG as a bugfix for integrations?

rfratto, line 205: Use `struct{}` since it has no size (vs an `interface{}` which is always a fat pointer to something, even if it's nil)

```suggestion
	// Check to ensure no duplicate singleton exist
	singletonCheck := make(map[string]struct{})
```

rfratto (approved):

rfratto (approved):
LGTM! I've left a few comments but I think this is good to go whenever you're ready. Nice work with this, I'm excited for people to start using it.

PR #1361, opened 2022-02-04
add password file and basic auth round tripper in crow

rlankfo:
#### PR Description 

Adding password-file flag to crow for use with querying protected cortex endpoints. 

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/crow/crow.go
+++ b/pkg/crow/crow.go
@@ -4,11 +4,13 @@ package crow
 
 import (
 	"context"
+	"crypto/tls"
 	"errors"
 	"flag"
 	"fmt"
 	"math"
 	"math/rand"
+	"net"
 	"net/http"
 	"strings"
 	"sync"
@@ -21,6 +23,7 @@ import (
 	promapi "github.com/prometheus/client_golang/api/prometheus/v1"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	commonCfg "github.com/prometheus/common/config"
 	"github.com/prometheus/common/model"
 	"github.com/weaveworks/common/user"
 )
@@ -30,6 +33,7 @@ type Config struct {
 	PrometheusAddr string // Base URL of Prometheus server
 	NumSamples     int    // Number of samples to generate
 	UserID         string // User ID to use when querying.
+	PasswordFile   string // Password File for auth when querying.
 	ExtraSelectors string // Extra selectors for queries, i.e., cluster="prod"
 
 	// Querying Params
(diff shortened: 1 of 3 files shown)


pkg/crow/crow.go
```diff
@@ -145,7 +150,21 @@ func newCrow(cfg Config) (*Crow, error) {
...
+			Proxy: http.ProxyFromEnvironment,
+			DialContext: (&net.Dialer{
+				Timeout:   30 * time.Second,
+				KeepAlive: 30 * time.Second,
+			}).DialContext,
+			TLSHandshakeTimeout: 10 * time.Second,
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: true,
+			},
+		}
```
rlankfo, line 164: I prefer to use the `DefaultRoundTripper` here but I was getting x509 certificate errors when hitting cortex endpoints (with valid certificates) so I've added this temporarily.  Is it possible we need to update `seego` to make sure we have latest CA certificates in the container?

rlankfo: To simplify this for now, I could add another flag to toggle `InsecureSkipVerify` so this isn't the default 🤔

pkg/crow/crow.go
```diff
@@ -145,7 +148,9 @@ func newCrow(cfg Config) (*Crow, error) {
 	apiCfg := api.Config{
 		Address: cfg.PrometheusAddr,
 	}
-	if cfg.UserID != "" {
+	if cfg.UserID != "" && cfg.PasswordFile != "" {
```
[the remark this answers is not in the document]
rlankfo: yeah -- that looked a bit confusing to me when I was adding this so I didn't want to change any existing behavior. Looking closer, I don't think any of our stuff is currently using this field so I suggest we keep the userID flag the same, and add a new orgID flag to inject in this header. Does that sound reasonable?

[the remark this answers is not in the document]
rfratto: Sure 👍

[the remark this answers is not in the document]
rlankfo: ok updated -- let me know what you think.

pkg/crow/crow.go
```diff
@@ -60,7 +63,9 @@ func (c *Config) RegisterFlagsWithPrefix(f *flag.FlagSet, prefix string) {
 	f.StringVar(&c.PrometheusAddr, prefix+"prometheus-addr", DefaultConfig.PrometheusAddr, "Root URL of the Prometheus API to query against")
 	f.IntVar(&c.NumSamples, prefix+"generate-samples", DefaultConfig.NumSamples, "Number of samples to generate when being scraped")
 	f.StringVar(&c.UserID, prefix+"user-id", DefaultConfig.UserID, "UserID to attach to query. Useful for querying multi-tenated Cortex.")
+	f.StringVar(&c.PasswordFile, prefix+"password-file", DefaultConfig.PasswordFile, "Password file to use with auth. Useful for querying multi-tenated Cortex.")
```
rfratto, line 66: The help text should probably change here to say it's for basic auth (and remove the "Useful for querying multi-tenanted Cortex" part since that no longer applies)

pkg/crow/crow.go
```diff
@@ -60,7 +63,9 @@ func (c *Config) RegisterFlagsWithPrefix(f *flag.FlagSet, prefix string) {
...
 	f.StringVar(&c.ExtraSelectors, prefix+"extra-selectors", DefaultConfig.ExtraSelectors, "Extra selectors to include in queries, useful for identifying different instances of this job.")
+	f.StringVar(&c.OrgID, prefix+"org-id", DefaultConfig.OrgID, "Org ID to inject in X-Org-ScopeID header when querying.")
```
rfratto, line 68: Maybe add "Useful for querying multi-tenated Cortex directly." to the help-text here?

rfratto (approved):
LGTM, thanks!

PR #1367, opened 2022-02-08
SNMP prototype

mattdurham:
Did not add support for v1, but simplified config

This creates endpoints like `\integrations\snmp_exporter\test` that are equivalent to `\integrations\snmp_exporter\metrics?target=localhost&...` it also self scrapes and adds `snmp_target` as a label without user having to manually do relabel configs.

Below is the config used with v2 integrations
```
server:
  http_listen_port: 12345
  log_level: debug
metrics:
  wal_directory: /tmp/grafana-agent-normal
  global:
    scrape_interval: 15s
    remote_write:
    - url: https://prometheus-us-central1.grafana.net/api/prom/push 
      basic_auth: 
        username: 1234 
        password: pw 
  configs:
  - name: default
integrations:
  snmp_exporter: 
    # config_file: snmp.yml
    walk_params:
      private:
        retries: 1
        version: 2
        auth:
          community: asljfdknlj123
      public:
        retries: 1
        version: 2
        auth:
          community: public
    snmp_targets:
    - name: test
      address: localhost
      module: if_mib
      walk_params: public
```
Screen Shot 2022-02-08 at 3 38 08 PM


mattdurham:
No longer needed

PR #1374, opened 2022-02-09
Update changelog

rfratto:
Remove BUGFIX entries that fix a bug introduced by main (i.e., bugs which were never part of a release)


hjet (approved):
my bad, lgtm

Issue #1073, opened 2021-11-08
support prometheus extra-scrape-metrics


rlankfo:
Prometheus v2.30.0 introduced the extra-scrape-metrics feature flag, which when enabled adds `scrape_timeout_seconds` and `scrape_sample_limit metric`. These metrics about the scrape targets could be useful in the agent as well.

We should make this configurable in the agent. This setting currently defaults to false.

Note this is an experimental upstream feature.

rlankfo:
https://prometheus.io/docs/prometheus/latest/feature_flags/#extra-scrape-metrics

PR #1379, opened 2022-02-10
enable extra scrape metrics

rlankfo:
#### PR Description 

This adds a feature flag for enabling [extra scrape metrics](https://prometheus.io/docs/prometheus/2.31/feature_flags/#extra-scrape-metrics). 

Closes https://github.com/grafana/agent/issues/1073

#### Notes to the Reviewer

This is still experimental in prometheus so I've included the feature flag (instead of defaulting this to true). The latest version of prometheus includes additional scrape metrics which aren't included yet in the agent. See https://prometheus.io/docs/prometheus/latest/feature_flags/

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,8 @@ Main (unreleased)
   network. As a result of this change, the `client_config` field has been
   removed. (@rfratto)
 
+-  `extra-scrape-metrics` can now be enabled with the `--enable-features=extra-scrape-metrics` feature flag. See https://prometheus.io/docs/prometheus/2.31/feature_flags/#extra-scrape-metrics for details. (@rlankfo)
+
 ### Other changes
 
 - Update base image of official Docker containers from Debian buster to Debian
--- a/docs/user/configuration/flags.md
+++ b/docs/user/configuration/flags.md
@@ -30,6 +30,7 @@ Valid feature names are:
 * `remote-configs`: Enable [retrieving]({{< relref "./_index.md#remote-configuration-experimental" >}}) config files over HTTP/HTTPS
 * `integrations-next`: Enable [revamp]({{< relref "./integrations/integrations-next/" >}}) of the integrations subsystem
 * `dynamic-config`: Enable support for [dynamic configuration]({{< relref "./dynamic-config" >}})
+* `extra-scrape-metrics`: When enabled, additional time series  are exposed for each metrics instance scrape. See [Extra scrape metrics](https://prometheus.io/docs/prometheus/latest/feature_flags/#extra-scrape-metrics).
 
 ## Configuration file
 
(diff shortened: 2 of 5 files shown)


mattdurham:
Thoughts on adding a tiny unit test to make sure it propagates?

rfratto (commented):
$0.02: I think we can enable this by default, and make it optional if there ends up being problems. Prometheus can be a little more aggressive with feature flags, but I think this is small enough to just enable.

rlankfo:
> $0.02: I think we can enable this by default, and make it optional if there ends up being problems. Prometheus can be a little more aggressive with feature flags, but I think this is small enough to just enable.

@rfratto It does seem minor so I agree we could just enable this by default if we want access to these metrics. My only concern would be unexpected increase in cardinality caused by these additional metrics and their labels for users who have a substantial amount of scrape targets. I believe this is why prometheus initially wrapped this in a feature flag.

rfratto:
> My only concern would be the unexpected increase in cardinality caused by these additional metrics and their labels for users who have a substantial amount of scrape targets. I believe this is why prometheus initially wrapped this in a feature flag.

That's a good point. If we want to make it opt-out, maybe it should be a permanent flag or a YAML option? Feature flag implies that it will eventually be permanently on by default once it graduates out of experimental.

rlankfo:
> That's a good point. If we want to make it opt-out, maybe it should be a permanent flag or a YAML option? Feature flag implies that it will eventually be permanently on by default once it graduates out of experimental.

Personally I think the feature flag makes sense here (opt-in) vs deviating from what [Prometheus is doing](https://prometheus.io/docs/prometheus/latest/feature_flags/#extra-scrape-metrics) with these.  In their docs they state the behavior of these features could change in future releases, or they *may* end up permanently on by default, which seems reasonable for this feature.

rlankfo:
> the behavior of these features could change in future releases, or they _may_ end up permanently on by default

Expanding on this a bit in general, it probably isn't a good decision to enable experimental Prometheus features by default in the agent.

mattdurham:
I am for leaving it as a feature flag is legit if there is considerable uncertainty within Prometheus about the future of it.

mattdurham (changes requested):
LGTM, aside from a tiny test

mattdurham (approved):
Thanks!

mattdurham:
Apologies, one more thing. We should document this in the configuration guide somewhere. Note sure if [here](https://github.com/grafana/agent/blob/585e058c64fac7caec907471eeede33c93aab6d2/docs/user/configuration/_index.md) is the best place? Given our increasing usage of feature flags somewhere dedicated might make sense but that doesn't have to be decided in this PR has long as we have some doc somewhere.

rlankfo:
> Given our increasing usage of feature flags somewhere dedicated might make sense but that doesn't have to be decided in this PR has long as we have some doc somewhere.

I think this makes sense -- I'll go ahead and address https://github.com/grafana/agent/issues/1383 in this PR and create a dedicated section in the docs.

rfratto:
@rlankfo gentle reminder about this :)

rlankfo:
> @rlankfo gentle reminder about this :)

Rebased and updated docs. This should be good to go, sorry for delay on this

rfratto (approved):

PR #1386, opened 2022-02-14
1250 oauth2 tracing

### LINKED ISSUE #1250: Feature Request: Oauth2 extension OTEL / Grafana agent tracing
**Is your feature request related to a problem? Please describe.**

Currently the tracing part of the Grafana Agent is missing Oauth2 authentication capabilities. These are available for the metrics and logs part. As the tracing part is using the opentelemetry collector, there is an extension that enables this: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/oauth2clientauthextension

**Describe the solution you'd like**

Would be great if it would be possible to use Oauth2 authentication for all 3 data streams within the Agent. By implementing this extension or make it possible to configure extensions we can achieve this goal.

Canuteson:
#### PR Description 

This PR adds support for configuring opentelemetry exporters with oauth support, utilizing the oauth2clientauthextension:
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/oauth2clientauthextension/README.md

#1250 

#### Notes to the Reviewer

The added `oauth` section of the traces exporter configuration uses the opentelemetry auth config format, simplifying the grafana-agent configuration and dynamically creating the extension configuration when needed. This PR supports having different oauth configurations for different exporters if necessary.

Example trace exporter oauth config:
```
receivers:
  jaeger:
    protocols:
      grpc:
remote_write:
  - endpoint: example.com:12345
    protocol: http
    oauth2:
      client_id: someclientid
      client_secret: someclientsecret
      token_url: https://example.com/oauth2/default/v1/token
      scopes: ["api.metrics"]
      timeout: 2s
```

#### PR Checklist

- [x] CHANGELOG updated 
- [X] Documentation added
- [X] Tests updated


pkg/traces/config.go
```diff
@@ -354,11 +408,39 @@ func (c *InstanceConfig) exporters() (map[string]interface{}, error) {
 			return nil, err
 		}
 		exporterName = fmt.Sprintf("%s/%d", exporterName, i)
+		if remoteWriteConfig.Oauth2 != nil {
```
joe-elliott, line 411: this makes more sense to me in the `exporter()` function since we build the rest of the config there.

I believe it's here b/c the `exporterName` is here? Perhaps we could adjust `exporter()` to take its own name and use it?

Oof, it also seems odd that `getExporterName()` doesn't take `i`. Tacking it on afterwards seems error prone.

joe-elliott (commented):
minor nit about the way the the exporters map is built, but overall this looks good. nice addition of extensions to our otel collector config and pipeline.

will leave for @mattdurham or @mapno to give a review as well. despite my nit, i'm fine with merging as is.

docs/configuration/traces-config.md
```diff
@@ -92,6 +92,31 @@ remote_write:
       [ password: <secret> ]
       [ password_file: <string> ]
 
+    # Configures opentelemetry exporters to use the OpenTelemetry auth extension `oauth2clientauthextension`.
+    # Can not be used in combination with `basic_auth`.
+    # See https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/oauth2clientauthextension/README.md
+    oauth2:
+    # Configures the TLS settings specific to the oauth2 client
```
mapno, line 99: I think this should go before the `tls` block.

pkg/traces/instance.go
```diff
@@ -213,6 +245,11 @@ func (i *Instance) buildAndStartPipeline(ctx context.Context, cfg InstanceConfig
 		return fmt.Errorf("failed to start receivers: %w", err)
 	}
 
+	err = i.extensions.NotifyPipelineReady()
```
mapno, line 248: This error could be returned directly.

```go
return i.extensions.NotifyPipelineReady()
```

pkg/traces/instance.go
```diff
@@ -180,14 +198,28 @@ func (i *Instance) buildAndStartPipeline(ctx context.Context, cfg InstanceConfig
 		MeterProvider:  metric.NewNoopMeterProvider(),
 	}
 
+	// start extensions
+	i.extensions, err = extensions.Build(settings, appinfo, otelConfig, factories.Extensions)
+	if err != nil {
+		i.logger.Error(fmt.Sprintf("failed to build extensions:%s", err.Error()))
+		return fmt.Errorf("failed to create extensions builder: %w", err)
+	}
+	i.logger.Info(fmt.Sprintf("starting extensions:%+v", i.extensions))
```
mapno, line 207: I would drop this log and the one in line 219. The collector's logger already handles logging during start-up.

Also, they will print the pointers to the extensions/exporters, which isn't very helpful.

mapno (approved):
Just left a few minor comments. Overall, it looks good to me. Nice addition. Thanks!!

pkg/traces/config.go
```diff
@@ -160,6 +164,50 @@ var DefaultRemoteWriteConfig = RemoteWriteConfig{
...
+// TLSClientSetting configures the oauth2client extension TLS; compatible with configtls.TLSClientSetting
+type TLSClientSetting struct {
+	CAFile             string `yaml:"ca_file,omitempty"`
+	CertFile           string `yaml:"cert_file,omitempty"`
+	KeyFile            string `yaml:"key_file,omitempty"`
+	MinVersion         string `yaml:"min_version,omitempty"`
+	MaxVersion         string `yaml:"max_version,omitempty"`
+	Insecure           bool   `yaml:"insecure"`
+	InsecureSkipVerify bool   `yaml:"insecure_skip_verify"`
+	ServerName         string `yaml:"server_name_override,omitempty"`
```
mattdurham, line 176: Nit: `ServerNameOverride`

pkg/traces/config.go
Canuteson: Good point. I agree.
66315ced

pkg/traces/config.go
Canuteson: 17596c13

pkg/traces/instance.go
Canuteson: 17596c13

pkg/traces/instance.go
Canuteson: Good call.
17596c13

docs/configuration/traces-config.md
Canuteson: 17596c13

mattdurham:
Roughly tested this out and it worked well, will review it once more tomorrow now that the checks are running.

mapno (commented):
Changes look good to me. Thanks for addressing them!

One thing that I missed in the previous review, the config test is flaky now, since `extensions` is built from a map and order is not guaranteed. There is a function `sortPipelines` to avoid this very issue with exporters and receivers. We will need a similar function to sort extensions so tests are stable.

PR #1391, opened 2022-02-17
RFC-0001: Add rules for when RFC PRs should be merged

rfratto:
This PR adds rules for how RFC PRs should be merged.

This became a clear need after thinking about how the work for integrations in Grafana Agent Operator has been ongoing since [September](https://github.com/grafana/agent/issues/883) and [changed](https://github.com/grafana/agent/pulls/1224) multiple times since then during exploratory implementation.


mattdurham (approved):
LGTM

PR #1399, opened 2022-02-18
[dev.frontend-receiver] sourcemap support

### LINKED ISSUE #1295: [dev.frontend-receiver] add sourcemap support
Add opt-in sourcemap support where receiver is able to use sourcemaps to transform exception stacktrace frames from compiled locations to original source locations. 

Several modes should be supported to retrieve sourcemaps:
1. Download compiled source, read source map url comment, download source map, cache 
2. Download compiled source, read inline sourcemap, cache
3. Configure sourcemap folder location on file system. Per http origin + release

For example:

```yaml
sourcemaps:
  download: false # if true and sourcemap not found on fs, attempt to download
  filesystem:
     - origins: "https://fooapp.1.cdn.com,https://fooap.2.cdn.com" # or just "*" for any origin
       folder: "/etc/fooapp/${release}/"
```

Then if location is  "https://fooapp.1.cdn.com/public/js/main.js" and `app.release` provided in payload is `abcd`, will attempt to find sourcemap at `/etc/fooapp/abcd/public/js/main.js.map`

domasx2:
#### PR Description
Adds sourcemap support for the app o11y receiver. It can be configured to:
* Download minified source, parse sourcemap URL, download sourcemap
* Look up sourcemaps on file system. Config allows mapping minified source URL prefix to file system folder, using URL path after the prefix as path within the defined folder

Fixes #1295

#### PR Checklist

- [ ] ~~CHANGELOG updated~~ Let's update changelog when we merge dev branch
- [x] Documentation added
- [x] Tests updated

TODOS:
- [x] add test coverage for sanitizing url paths


pkg/integrations/v2/app_o11y_receiver/sourcemaps/sourcemaps.go
```diff
@@ -0,0 +1,331 @@
...
+
+	//attempt to download
+	if strings.HasPrefix(sourceURL, "http") && utils.URLMatchesOrigins(sourceURL, store.config.DownloadFromOrigins) {
+		return store.downloadSourceMapContent(sourceURL)
+	}
+	return nil, "", nil
+}
+
+func (store *RealSourceMapStore) getSourceMap(sourceURL string, release string) (*sourceMap, error) {
+	store.Lock()
```
kpelelis, line 234: I am not 100% with lazy loading the source maps and I will share my thoughts. First of all, this will make the first request from each instance to be delayed a bit. I would probably go for eager loading in this case on startup.

kpelelis: That being said the lock will delay all concurrent requests while the app is loading

pkg/integrations/v2/app_o11y_receiver/sourcemaps/sourcemaps.go
```diff
@@ -0,0 +1,331 @@
...
+	//attempt to find in fs
+	for _, fileconf := range store.config.FileSystem {
+		content, sourceMapURL, err = store.getSourceMapFromFileSystem(sourceURL, release, fileconf)
+		if content != nil || err != nil {
+			return content, sourceMapURL, err
+		}
+	}
+
+	//attempt to download
+	if strings.HasPrefix(sourceURL, "http") && utils.URLMatchesOrigins(sourceURL, store.config.DownloadFromOrigins) {
```
kpelelis, line 227: Just as note, we need to always prefix the links with `http{s}://` or else the links won't work right?

pkg/integrations/v2/app_o11y_receiver/sourcemaps/sourcemaps.go
```diff
@@ -0,0 +1,331 @@
...
+		return nil, "", err
+	}
+	return result, resolvedSourceMapURL, nil
+}
+
+func (store *RealSourceMapStore) getSourceMapFromFileSystem(sourceURL string, release string, fileconf config.SourceMapFileLocation) (content []byte, sourceMapURL string, err error) {
+	if len(sourceURL) == 0 || !strings.HasPrefix(sourceURL, fileconf.MinifiedPathPrefix) || strings.HasSuffix(sourceURL, "/") {
+		return nil, "", nil
+	}
+	pathParts := []string{strings.Replace(fileconf.Path, "{RELEASE}", cleanForFilePath(release), 1)}
```
kpelelis, line 193: nit/suggestion: Instead of using our custom templating language, we use go's [template](https://pkg.go.dev/text/template) package (i.e. "{{ .Release }}" or whatnot)

pkg/integrations/v2/app_o11y_receiver/sourcemaps/sourcemaps.go
```diff
@@ -0,0 +1,331 @@
...
+	return &models.Exception{
+		Type:       ex.Type,
+		Value:      ex.Value,
+		Stacktrace: &models.Stacktrace{Frames: frames},
+		Timestamp:  ex.Timestamp,
+	}
+}
+
+func cleanForFilePath(x string) string {
+	return strings.TrimLeft(strings.ReplaceAll(strings.ReplaceAll(x, "\\", ""), "/", ""), ".")
```
kpelelis, line 322: The x string needs to be [cleaned](https://pkg.go.dev/path#Clean) to prevent [path traversal](https://owasp.org/www-community/attacks/Path_Traversal). In this case it won't pose an actual problem as the path undergoes a `path.Join` function afterwards which will cleanup the path, but I would add it as an extra security measure.

pkg/integrations/v2/app_o11y_receiver/sourcemaps/sourcemaps.go
```diff
@@ -0,0 +1,331 @@
...
+	resp, err := store.httpClient.Get(url)
+	if err != nil {
+		store.metrics.downloads.WithLabelValues(getOrigin(url), "?").Inc()
+		return nil, err
+	}
+	store.metrics.downloads.WithLabelValues(getOrigin(url), fmt.Sprint(resp.StatusCode)).Inc()
+	if resp.StatusCode != 200 {
+		return nil, fmt.Errorf("unexpected status %v", resp.StatusCode)
+	}
+	defer resp.Body.Close()
```
kpelelis, line 127: Move that on line 122. If 124 fails, you will have a dangling reader

pkg/integrations/v2/app_o11y_receiver/sourcemaps/sourcemaps.go
```diff
@@ -0,0 +1,331 @@
...
+	body, err := io.ReadAll(resp.Body)
```
kpelelis, line 128: Maybe we can check content-type as well here to make sure we do not download non javascript files

pkg/integrations/v2/app_o11y_receiver/sourcemaps/sourcemaps.go
```diff
@@ -0,0 +1,331 @@
...
+	defer store.Unlock()
+
+	cacheKey := fmt.Sprintf("%s__%s", sourceURL, release)
+
+	if smap, ok := store.cache[cacheKey]; ok {
+		return smap, nil
+	}
+	content, sourceMapURL, err := store.getSourceMapContent(sourceURL, release)
+	if err != nil || content == nil {
+		store.cache[cacheKey] = nil
```
kpelelis, line 244: Why is this needed? Are we terminating the sequence after that? If not, this can cause a potential memory leak. I suggest using the [delete](https://pkg.go.dev/builtin#delete) function to remove the key completely

pkg/integrations/v2/app_o11y_receiver/sourcemaps/sourcemaps.go
```diff
@@ -0,0 +1,331 @@
...
+		return nil, err
+	}
+	if content != nil {
+		consumer, err := sourcemap.Parse(sourceMapURL, content)
+		if err != nil {
+			store.cache[cacheKey] = nil
```
kpelelis, line 250: Same as above, use delete instead

pkg/integrations/v2/app_o11y_receiver/sourcemaps/sourcemaps.go
```diff
@@ -0,0 +1,331 @@
...
+	}
+	frames := []models.Frame{}
+
+	for _, frame := range ex.Stacktrace.Frames {
+		mappedFrame, err := store.ResolveSourceLocation(&frame, release)
+		if err != nil {
+			level.Error(store.l).Log("msg", "Error resolving stack trace frame source location", "err", err)
+			frames = append(frames, frame)
+		} else {
+			if mappedFrame != nil {
```
kpelelis, line 305: This can be lowered a level of indentation with `else if mappedFrame != nil { } .... else { }`

kpelelis (changes requested):
Really great (and a lot of) work. I like how you added a lot of tests, which makes reviewing much much easier. Also kudos on including all cases of source maps.

My main argument with this implementation is the lazy loading of the files. This service is intended to be high volume and I delaying the first request in each restart of it might cause some issues.

There are also a couple of potential mem leaks which are trivial to fix though

pkg/integrations/v2/app_o11y_receiver/sourcemaps/sourcemaps.go
```diff
@@ -0,0 +1,331 @@
...
+		}
+	}
+
+	if fileService == nil {
+		fileService = &osFileService{}
+	}
+
+	metrics := &sourceMapMetrics{
+		cacheSize: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: "app_o11y_receiver_sourcemap_cache_size",
```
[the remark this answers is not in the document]
domasx2: I think it adds some labels, but not a prefix

- [ ] move prefix to constant

pkg/integrations/v2/app_o11y_receiver/sourcemaps/sourcemaps.go
domasx2: I have thought about these issues, here's my thinking.

Main argument:
* This is the simplest approach with least effort and complexity. I suspect that it might be _good enough_, so would rather give it a try and collect some data before optimizing

Supporting arguments:
* Slight delay does not really impact antyhing, frontend is not relying on quick response
* There's unlikely to be *many* sourcemaps to load in a typical project. I'd say Grafana is a fairly big frontend, and it builds to only dozens of source bundles
* In case of downloads, sourcemap locations are not known ahead of time so cannot be eagerly loaded
* Loading from file is fast, and we don't expect to have to load many files

pkg/integrations/v2/app_o11y_receiver/sourcemaps/sourcemaps.go
domasx2: This URL comes from error stacktrace, not a link. So it should always be `http{s}://`, unless you've opened a html page from local file system and it's `file://`

pkg/integrations/v2/app_o11y_receiver/sourcemaps/sourcemaps.go
domasx2: Good catch!
- [ ] move

pkg/integrations/v2/app_o11y_receiver/sourcemaps/sourcemaps.go
domasx2: I'm not sure that it's safe to limit only to `.js`. It could be at least `wasm` too?

pkg/integrations/v2/app_o11y_receiver/sourcemaps/sourcemaps.go
domasx2: This is overwriting an existing item in the map, this is adding a not yet existing key with `nil` value so we don't try to download again if first time didn't work. So should be ok mem leak wise.

That said, accumulation of stuff in the cache might be a problem down the road that perhaps could be solved with adding some sort of mechanism to delete items that haven't been used for a while

pkg/integrations/v2/app_o11y_receiver/sourcemaps/sourcemaps.go
domasx2: - [ ] refac

pkg/integrations/v2/app_o11y_receiver/sourcemaps/sourcemaps.go
domasx2: This is for sanitizing a single path item rather than the entire path, actual sanitization for path traversal is happening [here](https://github.com/grafana/agent/blob/563dad3763ccad9823b7747dbef8a71427901dcc/pkg/integrations/v2/app_o11y_receiver/sourcemaps/sourcemaps.go#L195).


`path.Clean` **does not prevent path traversal**, they really should put a sticker warning on it :/ See http://technosophos.com/2016/03/31/go-quickly-cleaning-filepaths.html.

domasx2: - [ ] improve function name to mention path part

domasx2: thanks for  looking out for this 👍  great review

pkg/integrations/v2/app_o11y_receiver/sourcemaps/sourcemaps.go
domasx2: good point!
- [ ] use go templating lang

pkg/integrations/v2/app_o11y_receiver/sourcemaps/sourcemaps.go
kpelelis: Lets go with that then and keep an eye on it

pkg/integrations/v2/app_o11y_receiver/sourcemaps/sourcemaps.go
kpelelis: I guess we have a bit too many types indeed. I am bit extra careful about sanitising/validating everything as we load resources from the filesystem/internet based on user data

kpelelis (changes requested):
Great! Thanks for looking into everything. I am rejecting the pull request, even though it is in a great state just so that we resolve my last questions and it is good to go

dev.sh
```diff
@@ -0,0 +1 @@
+go run ./cmd/agent --config.file=./cmd/agent/agent-dev.yaml -enable-features=integrations-next
```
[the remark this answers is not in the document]
domasx2: oops :see_no_evil:

kpelelis (approved):
🚀 🚗

PR #1411, opened 2022-02-22
[operator] - Use _file variants for basic auth credentials.

### LINKED ISSUE #1061: Operator: MetricsInstance should use password file
@rfratto to add more context

captncraig:
#### PR Description 

Operator-generated configs now use password_file and username_file instead of embedding directly. Should allow for faster recognition of changes, and marginally more secure configs.

Fixes #1061 

#### PR Checklist

- [ ] CHANGELOG updated 
- [ ] Documentation added
- [x ] Tests updated

### DIFF
--- a/pkg/operator/config/config_test.go
+++ b/pkg/operator/config/config_test.go
@@ -79,10 +79,10 @@ func TestBuildConfigMetrics(t *testing.T) {
 								basicAuth:
 									username:
 										name: example-secret
-										key: key
+										key: uname
 									password:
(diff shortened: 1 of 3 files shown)


rfratto (approved):
LGTM! I don't think it needs a CHANGELOG entry either (or docs). It's a fairly internal change, though it will handle changes faster than reconciling a new secret. I don't think users _need_ to necessarily know that, though. Up to you if you want to mark it as an ENHANCEMENT or not.

Issue #1414, opened 2022-02-22
Operator: Implement support for integrations 


rfratto:
This issue tracks completion for implementing [RFC-0002](https://github.com/grafana/agent/blob/main/docs/rfcs/0002-integrations-in-operator.md), first introduced by #1224.

PR #1417, opened 2022-02-23
[dev.operator-integrations] Add Integration into operator resource hierarchy

rfratto:
#### PR Description 
This PR is a reimplementation of #1288 given the recent changes to the operator integrations RFC (#1224).

It introduces one new CRD, Integration, which defines an integration to run. The new CRD is injected into the resource hierarchy and discovered during reconcile, which is verified by a test.

Additional changes:

* `pkg/operator/config.Deployment` has been moved to `pkg/operator/apis/monitoring/v1alpha1.Deployment`. This allows us to take advantage of the DeepCopy function generation, which was previously manually maintained.
* `.gitattributes` has been modified to mark CRDs as generated so they don't inflate the size of diffs during PR review, though I think this will only change future PRs and not this one.
* Imports to `pkg/operator/apis/monitoring/v1alpha1` are now always aliases to `gragent` for consistency. `gragent` was picked for its brevity and should be mostly grokkable by readers.

Related to #1414.

#### Notes to the Reviewer
While the new resources are discovered during reconcile, nothing is done with them at the moment. Incorporating into the reconcile and deploying the integrations will be done over a few follow up PRs.

#### PR Checklist

- [x] CHANGELOG updated (N/A) 
- [x] Tests updated


.gitattributes
```diff
@@ -0,0 +1 @@
+pkg/operator/crds/*.yaml linguist-generated=true
```
rfratto, line 1: I think this will impact future PRs, but not this one 🤔

pkg/operator/apis/monitoring/v1alpha1/types_integrations.go
```diff
@@ -0,0 +1,122 @@
...
+	// Label selector to find Integration resources to run. When nil, no
+	// integration resources will be defined.
+	Selector *metav1.LabelSelector `json:"selector,omitempty"`
+
+	// Label selector for namespaces to search when discovering integration
+	// resources. If nil, integration resources are only discovered in the
+	// namespace of the GrafanaAgent resource.
+	//
+	// Set to `{}` to search all namespaces.
+	NamespaceSelector *metav1.LabelSelector `json:"namespaceSelector,omitempty"`
```
rfratto, line 21: These are called `instanceSelector` and `instanceNamespaceSelector` for metrics and logs, so this is inconsistent with that. 

OTOH, those refer to selecting `MetricsInstances` and `LogsInstances`, so it might make sense to drop the prefix here.

pkg/operator/apis/monitoring/v1alpha1/types_integrations.go
```diff
@@ -0,0 +1,122 @@
...
+// configure a supported integration's autoscrape block with:
+//
+//     autoscrape:
+//       enable: true
+//       metrics_instance: default/production
+//
+// There is currently no way for telemetry created by an Operator-managed
+// integration to be collected from outside of the integration itself.
+type Integration struct {
+	metav1.TypeMeta   `json:",inline"`
```
tpaschalis, line 47: This works for bson, but does it work for json as well? 🤯

tpaschalis: After a discussion, it's probably for the best to be explicit about our need to inline TypeMeta here. Resolving this comment.

captncraig (approved):
LGTM - My biggest concern is usability with the config and volumes and things being pretty free form fields.

It will probably be hard for users to know how to use this without pre-built snippets for each particular integration in the docs.

But no major design objections from me.

pkg/operator/apis/monitoring/v1alpha1/types_integrations.go
```diff
@@ -0,0 +1,122 @@
...
+	//
+	// Secrets will be mounted at
+	// /etc/grafana-agent/integration-secrets/<secret_namespace>/<secret_name>.
+	Secrets []string `json:"secrets,omitempty"`
+
+	// An extra list of ConfigMaps in the same namespace as the Integration which
+	// will be mounted into the Grafana Agent pod running this integration.
+	//
+	// ConfigMaps will be mounted at
+	// /etc/grafana-agent/integration-secrets/<secret_namespace>/<secret_name>.
```
rlankfo, line 95: looks like this should be updated for config maps?

rlankfo (approved):
LGTM +1

PR #1421, opened 2022-02-24
Extend linting configuration file

rfratto:
* Add depguard linter to reject packages we tend to avoid
* Replace golint with revive, since golint is deprecated 
* Remove interfacer, which is deprecated with no replacement 
* Add makezero linter to detect misuse of make with append
* Add tenv to prefer t.Setenv over os.Setenv in tests
* Add whitespace to report unnecessary blank lines 
* Ignore test files for errcheck

In addition to the above, I refactored the following:

* Remove settings that just re-set default values, instead pointing to the website to retrieve defaults. 
* Simplify the errcheck rule to only include functions we actually need to ignore.


.golangci.yml
```diff
@@ -72,3 +73,12 @@ issues:
     # and in most cases it's ok. This is copied from the golangci-lint defaults
     # mut modified to include go-kit logging.
     - Error return value of .((os\.)?std(out|err)\..*|.*Close|.*Flush|os\.Remove(All)?|.*print(f|ln)?|os\.(Un)?Setenv|.*\.Log). is not checked
+
+linters-settings:
+  depguard:
+    list-type: denylist
+    include-go-root: true
+    packages-with-error-message:
```
rlankfo, line 81: Maybe we want to add one for golang.org/x/sync/errgroup to use github.com/oklog/run instead?

rlankfo (approved):

mattdurham (approved):
Lgtm

mattdurham (approved):
LGTM

Issue #1394, opened 2022-02-17
Update to Loki v2.4.2 (or newer) dependency


mattdurham:
This change may be non-trivial due to the change Loki's migration over to grafana/dskit. #1127 also has some details related to this

rfratto:
Do we want to make this more generic? i.e., update to Loki v2.4.2 and close #1127?

PR #1422, opened 2022-02-24
upgrade to loki v2.4.2

rlankfo:
#### PR Description 

* upgrade to loki v2.4.2
* replace cortex ring with dskit ring package

Closes https://github.com/grafana/agent/issues/1394

#### Notes to the Reviewer

The github.com/cortexproject/cortex/pkg/ring has been migrated to a more generic version of the ring in github.com/grafana/dskit/ring. Loki has moved to this version. There were some minor API differences this PR addresses, and this ring version uses promauto so we no longer need the calls to register. I believe we could also remove explicit calls to Unregister but I have not done that here.

#### PR Checklist

- [x] CHANGELOG updated 
- [ ] Documentation added
- [x] Tests updated


pkg/metrics/cluster/node.go
```diff
@@ -114,15 +114,13 @@ func (n *node) ApplyConfig(cfg Config) error {
...
-		return fmt.Errorf("failed to register ring metrics: %w", err)
-	}
+
 	if err := services.StartAndAwaitRunning(context.Background(), r); err != nil {
 		return fmt.Errorf("failed to start ring: %w", err)
 	}
 	n.ring = r
 
-	lc, err := ring.NewLifecycler(cfg.Lifecycler, n, "agent", agentKey, false, n.reg)
+	lc, err := ring.NewLifecycler(cfg.Lifecycler, n, "agent", agentKey, false, n.log, prometheus.WrapRegistererWithPrefix("cortex_", n.reg))
```
mattdurham, line 123: Random Q: I assume this is for backward compatibility but should we change this from `cortex_` to something more agent specific?

rlankfo: That's a good question -- I'm not sure. This is just added for backward compatibility as you noted.

mattdurham (approved):
LGTM

Issue #451, opened 2021-03-05
grafana-agent-logs failing to boot on k3d


clux:
With latest loki yaml from README (run today):

```sh
NAMESPACE="monitoring" /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/grafana/agent/release/production/kubernetes/install-loki.sh)" > loki.yml
```

Leave the `grafana-agent-logs` pods unable to start. From `k describe`:

```
      Message:      failed to create containerd task: OCI runtime create failed: container_linux.go:370: starting container process caused: process_linux.go:459: container init caused: rootfs_linux.go:59: mounting "/etc/machine-id" to rootfs at "/run/k3s/containerd/io.containerd.runtime.v2.task/k8s.io/8cfe261f373ea2017618ffcec8958a9a2040044974f550bb2703a7733549e953/rootfs/etc/machine-id" caused: not a directory: unknown
```

but my `/etc/machine-id` file is the expected 32 character hex.

The loki install worked on a k3d cluster last week on a cluster with `k3d cluster create grafana --servers 1 --agents 1` (when it was still called promtail). `k3d` 4.2.0, `kubernetes 1.20.1`.

rfratto:
Thanks for reporting! I ran into this when working on #406 and worked around it by bind mounting that file to the k3d container: https://github.com/grafana/agent/blob/main/example/k3d/scripts/create.bash

`/etc/machine-id` is typtically used for promtail's journald suppport. Since that's not active in the k3d example, it makes sense to take out the bind mount from that container.

rfratto:
I can confirm that I'm able to reproduce your problem with `k3d` 4.2.0 and that it works as expected when you create the cluster with:

```
k3d cluster create grafana --servers 1 --agents 1 -v /etc/machine-id:/etc/machine-id
```

(As long as /etc/machine-id exists on the host)

clux:
Thanks for the quick response. I am able to work around it with the extra `k3d` volume mount at cluster creation time :-)

Issue #1178, opened 2021-12-16
jsonnet v2: /etc/machine-id cannot be mounted as is


56quarters:
Hello agent team,

I'm attempting to use the [v2 Agent jsonnet](https://github.com/grafana/agent/tree/main/production/tanka/grafana-agent/v2) to scrape logs. I'm running into the following issues:

* `/etc/machine-id` can't be mounted since it isn't a directory. Instead `/etc` has to be mounted and `machine-id` specifified as a `subPath`
* However, `/etc` can't be mounted read-only because the agent configuration file needs to be mounted to `/etc/agent/agent.yml`

The Jsonnet I'm using can be found at https://github.com/grafana/tns/compare/grafana:5d8f820...grafana:c9baaaa

Given the following jsonnet:

```
  daemonset_agent:
    gragent.new(name='grafana-agent', namespace='default') +
    gragent.withDaemonSetController() +
    gragent.withLogVolumeMounts(config={}) +
    gragent.withAgentConfig({ /* config goes here */ })
```

Kubernetes (using k3d and tanka) fails to start the pod with the following error:

```
    Last State:     Terminated
      Reason:       StartError
      Message:      failed to create containerd task: OCI runtime create failed: container_linux.go:380: starting container process caused: process_linux.go:545: container init caused: rootfs_linux.go:76: mounting "/etc/machine-id" to rootfs at "/etc/machine-id" caused: mount through procfd: not a directory: unknown
      Exit Code:    128
```

Modifying the above jsonnet to the following fixes that particular error:

```
  daemonset_agent:
    gragent.new(name='grafana-agent', namespace='default') +
    gragent.withDaemonSetController() +
    gragent.withAgentConfig({ /* config goes here */ })
     {
        controller+:
          // For reading docker containers. /var/log is used for the positions file
          // and shouldn't be set to readonly.
          $.util.hostVolumeMount('varlog', '/var/log', '/var/log') +
          $.util.hostVolumeMount('varlibdockercontainers', '/var/lib/docker/containers', '/var/lib/docker/containers', readOnly=true) +

          // For reading journald
          $.util.hostVolumeMount('etc', '/etc', '/etc', readOnly=true, volumeMountMixin={subPath: 'machine-id'}),
      }
```

But results in the following problem due to `/etc` being read-only

```
    Last State:     Terminated
      Reason:       StartError
[...]

PR #1424, opened 2022-02-24
Remove journald mount by default

hjet:
#### PR Description
Since these sample manifests are mostly used in tandem with the Cloud agent/logs [quickstart](https://grafana.com/docs/grafana-cloud/kubernetes/agent-k8s/k8s_agent_logs/) ConfigMap and soon the K8s integration ConfigMap (which will be even more streamlined), and neither scrapes logs from the journal, I propose omitting the mount and volume from the default manifests entirely. 

An alternative solution is changing the `etcmachineid` volume type to `file` but this is cleaner IMO. I prefer reducing the scope of our defaults.

Fixes https://github.com/grafana/agent/issues/451
Also see https://github.com/grafana/agent/issues/1178

#### PR Checklist

- [x] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/production/tanka/grafana-agent/v1/lib/logs.libsonnet
+++ b/production/tanka/grafana-agent/v1/lib/logs.libsonnet
@@ -70,10 +70,10 @@ local container = k.core.v1.container;
       // For reading docker containers. /var/log is used for the positions file
       // and shouldn't be set to readonly.
       k.util.hostVolumeMount('varlog', '/var/log', '/var/log') +
-      k.util.hostVolumeMount('varlibdockercontainers', '/var/lib/docker/containers', '/var/lib/docker/containers', readOnly=true) +
+      k.util.hostVolumeMount('varlibdockercontainers', '/var/lib/docker/containers', '/var/lib/docker/containers', readOnly=true),
 
       // For reading journald
-      k.util.hostVolumeMount('etcmachineid', '/etc/machine-id', '/etc/machine-id', readOnly=true),
+      //k.util.hostVolumeMount('etcmachineid', '/etc/machine-id', '/etc/machine-id', readOnly=true),
   },
 
   // scrapeKubernetesLogs defines a Logs config that can collect logs from
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
 
 - [FEATURE] Added config read API support to GrafanaAgent Custom Resource Definition.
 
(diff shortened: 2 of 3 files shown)


production/tanka/grafana-agent/v1/lib/logs.libsonnet
```diff
@@ -70,10 +70,10 @@ local container = k.core.v1.container;
       // For reading docker containers. /var/log is used for the positions file
       // and shouldn't be set to readonly.
       k.util.hostVolumeMount('varlog', '/var/log', '/var/log') +
-      k.util.hostVolumeMount('varlibdockercontainers', '/var/lib/docker/containers', '/var/lib/docker/containers', readOnly=true) +
+      k.util.hostVolumeMount('varlibdockercontainers', '/var/lib/docker/containers', '/var/lib/docker/containers', readOnly=true),
 
       // For reading journald
-      k.util.hostVolumeMount('etcmachineid', '/etc/machine-id', '/etc/machine-id', readOnly=true),
+      //k.util.hostVolumeMount('etcmachineid', '/etc/machine-id', '/etc/machine-id', readOnly=true),
```
rfratto, line 76: I'm not sure about commenting this out, because people still might want it. Can we make it opt-in instead? i.e., logsJournaldMixin or something?

rgeyer: +1 for making this a parameter, and off by default.

rfratto: This also might be a good opportunity to pull the bandaid and migrate everything to the v2 jsonnet, removing the versions. This is already optionally configurable in v2.

hjet: yeah, ok. agreed on pulling the bandaid and migrating to `v2` jsonnet. 

@rgeyer this will delay delivering pod logs a bit (but not by much). i think it's important as we still want to maintain this non-operator deployment path going forward.

another advantage is we can finally provide a statefulset as the default (and i can quickly write a config to turn on eventrouter that we can provide in the onboarding plugin)

closing this PR

@rfratto i couldn't find how to configure the `machine-id` vol in the v2 jsonnet (https://github.com/grafana/agent/blob/main/production/tanka/grafana-agent/v2/internal/helpers/logs.libsonnet#L13) - could you point me in the right direction?

PR #1439, opened 2022-03-01
Consulagent sd

### LINKED ISSUE #1419: Evaluate consulagent_sd_config for Metrics and Traces
The newest version of Promtail introduced consulagent_sd_config which only looks for targets on the local machine. Once we update our Promtail dependency, we should evaluate whether this SD mechanism can be added to metrics and traces as well for consistency. It would only work for metrics and traces if it emits enough labels to find a target's address and port. 

Depends on #1394.

chuckyz:
#### PR Description 
This enables the [consulagent_sd_config](https://github.com/grafana/loki/blob/main/docs/sources/clients/promtail/configuration.md#consulagent_sd_config) provider which is presently part of Loki via Promtail.

This allows those with either large Consul clusters or those with very limited hardware backing their Consul clusters to scale metrics by running a grafana agent per node and just scraping the agent on the host it's running on for services, causing no requests to go back to the main cluster.

#1419 

#### Notes to the Reviewer

There doesn't seem to be any documentation for any of the service discovery mechanisms within Grafana Agent?  Also from the discussion in #1419 I don't think this PR needs tests.

#### PR Checklist

- [x] CHANGELOG updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
 
 - [FEATURE] Added config read API support to GrafanaAgent Custom Resource Definition.
 
+- [FEATURE] Added consulagent_sd to target discovery. (@chuckyz)
+
 # v0.23.0 (2022-01-13)
 
 - [ENHANCEMENT] Go 1.17 is now used for all builds of the Agent. (@tpaschalis)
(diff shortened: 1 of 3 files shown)


rfratto (approved):
LGTM, thank you!

PR #1440, opened 2022-03-01
Parse const_labels correctly in exporter

fredr:
#### PR Description 

Config field `const_labels` have different types in the remotewriterexporter than it has in the [main config type](https://github.com/grafana/agent/blob/main/pkg/traces/config.go#L255). This causes an error when starting the agent:
```
error reading exporters configuration for \"remote_write\": 1 error(s) decoding:\n\n* 'const_labels[0]' has invalid keys: key, some"
```
With this example config:
```yaml
server:
  log_level: info
  http_listen_port: 12345

traces:
  configs:
  - name: test
    receivers:
      zipkin:
        endpoint: 0.0.0.0:9999
    remote_write:
      - endpoint: traces.local:5555
        insecure: false
        tls_config:
            insecure_skip_verify: true
    spanmetrics:
      metrics_instance: test
      const_labels:
        some: "somevalue"
        key: "value"
```

#### Notes to the Reviewer

The previous implementation also had a bug in the mapstructure field, where both value and name where named `name`.

#### PR Checklist

- [x] CHANGELOG updated 
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/traces/remotewriteexporter/exporter.go
+++ b/pkg/traces/remotewriteexporter/exporter.go
@@ -55,8 +55,8 @@ func newRemoteWriteExporter(cfg *Config) (component.MetricsExporter, error) {
 	if cfg.ConstLabels != nil {
 		ls = make(labels.Labels, 0, len(*cfg.ConstLabels))
 
-		for name := range *cfg.ConstLabels {
-			ls = append(ls, labels.Label{Name: name, Value: (*cfg.ConstLabels)[name]})
+		for name, value := range *cfg.ConstLabels {
+			ls = append(ls, labels.Label{Name: name, Value: value})
 		}
 	}
 
--- a/pkg/traces/remotewriteexporter/exporter.go
+++ b/pkg/traces/remotewriteexporter/exporter.go
@@ -50,12 +50,14 @@ type remoteWriteExporter struct {
 func newRemoteWriteExporter(cfg *Config) (component.MetricsExporter, error) {
 	logger := log.With(util.Logger, "component", "traces remote write exporter")
 
-	ls := make(labels.Labels, 0, len(cfg.ConstLabels))
-	for _, constLabel := range cfg.ConstLabels {
-		ls = append(ls, labels.Label{
-			Name:  constLabel.Name,
-			Value: constLabel.Value,
-		})
+	var ls labels.Labels
+
+	if cfg.ConstLabels != nil {
+		ls = make(labels.Labels, 0, len(*cfg.ConstLabels))
+
+		for name := range *cfg.ConstLabels {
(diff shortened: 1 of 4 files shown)


fredr:
Test failed with 
```
failed to start manager: inotify_add_watch /sys/fs/cgroup/devices: no space left on device
```
I think that is unrelated, I can run all the tests locally

pkg/traces/remotewriteexporter/exporter.go
```diff
@@ -50,12 +50,14 @@ type remoteWriteExporter struct {
...
-			Value: constLabel.Value,
-		})
+	var ls labels.Labels
+
+	if cfg.ConstLabels != nil {
+		ls = make(labels.Labels, 0, len(*cfg.ConstLabels))
+
+		for name := range *cfg.ConstLabels {
+			ls = append(ls, labels.Label{Name: name, Value: (*cfg.ConstLabels)[name]})
+		}
```
mapno, line 60: This could be simplified to loop with the value as well

```suggestion
		for name, value := range *cfg.ConstLabels {
			ls = append(ls, labels.Label{Name: name, Value: value})
		}
```

pkg/traces/remotewriteexporter/factory.go
```diff
@@ -13,20 +14,15 @@ const (
...
 var _ config.Exporter = (*Config)(nil)
 
 // Config holds the configuration for the Prometheus remote write processor.
 type Config struct {
 	config.ExporterSettings `mapstructure:",squash"`
 
-	ConstLabels  []label `mapstructure:"const_labels"`
-	Namespace    string  `mapstructure:"namespace"`
-	PromInstance string  `mapstructure:"metrics_instance"`
+	ConstLabels  *prometheus.Labels `mapstructure:"const_labels"`
```
mapno, line 23: Why use a pointer to `prometheus.Labels`? It makes handling later a bit more complicated having to dereference it.

mapno (commented):
Hi! Thanks for addressing this issue, it's an awful bug.

It'd be great if we add `const_labels` to the test. Please, add an entry to the CHANGELOG :)

pkg/traces/remotewriteexporter/factory.go
fredr: I just made it the same signature as in https://github.com/grafana/agent/blob/main/pkg/traces/config.go#L255
But that is probably not necessary, I'll switch to a non-pointer!

mapno (approved):
> I added a new test that sets the spanmetrics section and creates a traces collector.

Nice! Looks good to me. Thanks!

> Would you rather that I add it to the TestTraces-test?

That's fine by me. I think ideally the test should be in `config_test.go`, but it's fine as long as it's tested.

Thanks for the PR and addressing the comments! :)

PR #1449, opened 2022-03-02
remove duplicate suffix from calls_total spanmetric

fredr:
#### PR Description 
Because of an [upstream fix](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/2837/files), the metric have duplicate `_total` suffixes and comes out as `traces_spanmetrics_calls_total_total`

#### Notes to the Reviewer
I wanted to write a test for this as well, but didn't find a way of doing it, since the name is being created in the processor upstream. If you have any suggestions on a test, please let me know.

#### PR Checklist

- [x] CHANGELOG updated 
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/traces/remotewriteexporter/exporter.go
+++ b/pkg/traces/remotewriteexporter/exporter.go
@@ -22,14 +22,13 @@ import (
 )
 
 const (
-	nameLabelKey  = "__name__"
-	sumSuffix     = "sum"
-	countSuffix   = "count"
-	bucketSuffix  = "bucket"
-	leStr         = "le"
-	infBucket     = "+Inf"
-	counterSuffix = "total"
-	noSuffix      = ""
+	nameLabelKey = "__name__"
+	sumSuffix    = "sum"
+	countSuffix  = "count"
+	bucketSuffix = "bucket"
+	leStr        = "le"
+	infBucket    = "+Inf"
+	noSuffix     = ""
 )
 
 type dataPoint interface {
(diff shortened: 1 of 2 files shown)


mapno (approved):
LGTM, thanks for the contribution again!

pkg/traces/remotewriteexporter/exporter.go
```diff
@@ -161,25 +160,17 @@ func (e *remoteWriteExporter) handleHistogramIntDataPoints(app storage.Appender,
...
-		dataPoints := m.Sum().DataPoints()
-		if err := e.handleScalarIntDataPoints(app, m.Name(), counterSuffix, dataPoints); err != nil {
-			return err
-		}
-	case pdata.MetricDataTypeGauge:
-		dataPoints := m.Gauge().DataPoints()
-		if err := e.handleScalarIntDataPoints(app, m.Name(), noSuffix, dataPoints); err != nil {
-			return err
-		}
+	dataPoints := m.Gauge().DataPoints()
```
mapno, line 163: The data type check is necessary to correctly get the datapoints. While maybe spanmetrics only generated gauges (unsure), it's safer to keep it.

mapno (changes requested):
Sorry, missed a change that I believe should be reverted.

pkg/traces/remotewriteexporter/exporter.go
fredr: Ah, my bad, sorry, missed that they where different, thought the only diff where the prefix!

fredr: Fixed! Thanks for catching that!

mapno (approved):
Thanks :)

PR #1452, opened 2022-03-02
Increase swap space on windows for running tests

rfratto:
This fixes our issues 🤷

### DIFF
--- a/.github/workflows/bump-formula-pr.yml
+++ b/.github/workflows/bump-formula-pr.yml
@@ -5,9 +5,7 @@ on:
   push:
     tags:
       - 'v*'
-    tags-ignore:
-      # Ignore release candidates
-      - 'v*-rc.*'
+      - '!v*-rc.*' # Ignore release candidates
 jobs:
   homebrew-core:
     name: homebrew-core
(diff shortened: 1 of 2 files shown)


.github/workflows/bump-formula-pr.yml
```diff
@@ -5,9 +5,7 @@ on:
   push:
     tags:
       - 'v*'
-    tags-ignore:
-      # Ignore release candidates
-      - 'v*-rc.*'
```
rfratto, line 10: We were [getting some errors](https://github.com/grafana/agent/actions/workflows/bump-formula-pr.yml) about setting both tags and tags-ignore so I'm sneaking in the fix here

.github/workflows/test.yml
```diff
@@ -5,9 +5,18 @@ jobs:
     name: Test
     strategy:
       matrix:
-        platform: [macos-latest, windows-2019]
+        platform: [macos-latest, windows-latest]
```
rfratto, line 8: If windows-latest works with the pagefile increase, we probably should just stick to it instead of forcing 2019 IMO

mattdurham (approved):
Woot!

PR #1456, opened 2022-03-03
Clean up the docker compose example environment

rfratto:
This PR replaces the various docker-compose.*.yaml files with a single docker-compose.yaml file.

By default, this file will spawn and expose ports for Loki, Grafana, Tempo, and Cortex so telemetry data can be stored and visualized. This allows developers to bring up dependencies needed for testing a local Grafana Agent process, useful for validating code changes.

Version 3 of Docker Compose is used for profile support, allowing docker-compose.yaml to optionally run extra services:

* The `agent` profile will also run a Grafana Agent that collects metrics from itself and sends them to Cortex.

* The `integrations` profile will run example services that integrations can target, such as MySQL and Consul. Each of these `integrations` services can also be enabled independently by using the profile name matching the relevant integration name (i.e., the `mysqld_exporter` profile enables the MySQL service). All integration services will expose their ports for a local agent to connect to.

You can pass a comma-delimited list of profiles to Docker Compose at the command line:
 `docker compose --profile=agent,mysqld_exporter,consul_exporter up -d`


rfratto:
At some point I suggested that we might want to replace Docker Compose completely with k3d, but I'm starting to question whether that's the right idea. 

Sometimes I want to quickly spin up _just_ Loki/Cortex/Tempo/Grafana for testing things, which would be pretty hard to do with k3d, but is pretty simple with Docker Compose 3 and its profiles support.

rlankfo (approved):
lgtm

PR #1461, opened 2022-03-03
add container restarts alert

### LINKED ISSUE #1459: Smoke test environment: add alert for unhealthy exit
We stumbled on #1450 by luck: it happened rarely enough that it didn't fire any flapping alerts, and just happened to trigger the "unexpected memory usage" alert during a particularly unlucky period of a few restarts in a row. 

We should have an alert that checks for an exit status code of 1. This won't be possible with any existing agent metrics, and we'd have to either make a log-based alert looking for fatal/panic/etc or check k8s metrics to detect a pod crashing.

rlankfo:
Adds a smoke test alert for container restarts

Closes #1459

### DIFF
--- a/production/grafana-agent-mixin/alerts.libsonnet
+++ b/production/grafana-agent-mixin/alerts.libsonnet
@@ -135,7 +135,6 @@ local _config = config._config;
             expr: |||
               sum by (pod) (rate(kube_pod_container_status_restarts_total{namespace="agent-smoke-test"}[10m])) > 0
             |||,
-            'for': '2m',
             annotations: {
               summary: '{{ $labels.pod }} has a high rate of container restarts',
             },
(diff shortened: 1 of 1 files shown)


production/grafana-agent-mixin/alerts.libsonnet
```diff
@@ -130,6 +130,16 @@ local _config = config._config;
               summary: '{{ $labels.job }} has used more than 10KB per series for more than 5 minutes',
             },
           },
+          {
+            alert: 'GrafanaAgentContainerRestarts',
+            expr: |||
+              sum by (pod) (rate(kube_pod_container_status_restarts_total{namespace="agent-smoke-test"}[10m])) > 0
+            |||,
+            'for': '2m',
```
rfratto, line 138: I think we can drop the `for` so it fires immediately whenever there's an unexpected restart. WDYT?

rfratto (approved):

production/grafana-agent-mixin/alerts.libsonnet
rlankfo: that sounds good :+1:

PR #1463, opened 2022-03-03
integrations-next: allow all integrations to be unmarshaled as an array or object

rfratto:
This PR relaxes the restrictions on how v2 integrations are unmarshaled so that any integration, regardless of type, can be unmarshaled as an JSON object or an array.

Introducing this change makes it *much* easier for the operator to generate configs for integrations, as we no longer have to switch between the _configs and non-_configs key based on the integration being generated.

With this change, configurations like these are now valid:

```yaml
integrations:
  node_exporter_configs:
  - {}
```

Such a configuration would not be recommended to users, but as there is only one defined instance of node_exporter, it still satisfies the singleton requirements. It is unlikely that we would document this is possible.

Instead, the following configuration would still be documented and recommended for singleton integrations:

```yaml
integrations:
  node_exporter: {}
```

## Re-marshaling 

Regardless of how the integrations config is unmarshaled, it will be re-marshaled as the non-array type. This means this config:

```yaml 
integrations:
  node_exporter_configs:
  - {}
```

Will display as the following at the `/-/config` endpoint:

```yaml
integrations:
  node_exporter: {}
```

### DIFF
--- a/pkg/integrations/v2/controller.go
+++ b/pkg/integrations/v2/controller.go
@@ -11,8 +11,6 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/hashicorp/go-multierror"
-
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	"github.com/gorilla/mux"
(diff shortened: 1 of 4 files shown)


pkg/integrations/v2/controller.go
```diff
@@ -101,21 +99,28 @@ func (c *controller) UpdateController(cfg controllerConfig, globals Globals) err
...
+		singletonSet[cfg.Name()] = struct{}{}
 	}
-	if returnError != nil {
-		return returnError
+	if len(duplicatedSingletons) == 1 {
+		return fmt.Errorf("integration %q may only be defined once", duplicatedSingletons[0])
+	} else if len(duplicatedSingletons) > 1 {
+		list := strings.Join(duplicatedSingletons, ", ")
+		return fmt.Errorf("the following integrations may only be defined once each: %s", list)
 	}
```
rfratto, line 124: Just doing a little bit of tidying here so there's less indentation

pkg/integrations/v2/register.go
```diff
@@ -86,9 +86,12 @@ type UpgradedConfig interface {
 type Type int
 
 const (
+	// TypeInvalid is an invalid type.
+	TypeInvalid Type = iota
```
rfratto, line 90: Introducing TypeInvalid as the zero value removes the need for examining `ok` from RegisteredType

mattdurham (approved):
Looks good and nice simplification!

PR #1468, opened 2022-03-07
refactor makefile hardcode

laiwei:
#### PR Description 
Refactor makefile to reduce hardcode, and fix a typo of `mipls `.

#### PR Checklist

- [x] CHANGELOG updated 
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -335,8 +335,8 @@ dist/$(BINARY_PREFIX)agent-windows-installer: dist/$(BINARY_PREFIX)agent-windows
 	cp ./dist/$(BINARY_PREFIX)agent-windows-amd64.exe ./packaging/windows
 	cp LICENSE ./packaging/windows
 ifeq ($(BUILD_IN_CONTAINER),true)
-	docker build -t windows_installer -e BINARY_PREFIX=${BINARY_PREFIX} ./packaging/windows
-	docker run --rm -t -v "${PWD}:/home" -e VERSION=${RELEASE_TAG} windows_installer
+	docker build -t windows_installer ./packaging/windows
+	docker run --rm -t -v "${PWD}:/home" -e VERSION=${RELEASE_TAG} -e BINARY_PREFIX=${BINARY_PREFIX} windows_installer
 else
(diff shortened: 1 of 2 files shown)


rfratto:
Hey @laiwei! Can you give us a brief description of the goals you're trying to achieve here? 
I've also been considering a major refactor of the Makefike to make it easier to follow, and if I go forward with that I'd like to make sure I'm supporting the same use cases you're trying to get here.

laiwei:
> Hey @laiwei! Can you give us a brief description of the goals you're trying to achieve here? I've also been considering a major refactor of the Makefike to make it easier to follow, and if I go forward with that I'd like to make sure I'm supporting the same use cases you're trying to get here.

hi @rfratto, thanks for the awesome project.

This PR trying to solve following issues:
- the binary name "agent-linux-amd64" is too "general" to distinguished with other agents running on the same node. So I'd like to add a BINARY-PREFIX to the name of agent binary to make more readable such as "ps aux | grep xxx-agent"
- when building customized agent docker image and to push it to a private registry, I'd like to use a customized image name, such as "myorg/agent:latest" instead of "grafana/agent".
- currently, supported platforms and archs are hardcode, so i want to make it as a list defined in a variable (still work in progress).

rfratto:
Thanks for the feedback @laiwei! 

> the binary name "agent-linux-amd64" is too "general" to distinguished with other agents running on the same node. So I'd like to add a BINARY-PREFIX to the name of agent binary to make more readable such as "ps aux | grep xxx-agent"

Definitely agree; we've had #582 open for a while to prefix all of the binaries with `grafana-*` (i.e., `grafana-agent` and `grafana-agentctl`). I'm not sure how I feel about custom prefixes in the Makefile, but would the `grafana-` prefix be enough differentiation for you here? 

> when building customized agent docker image and to push it to a private registry, I'd like to use a customized image name, such as "myorg/agent:latest" instead of "grafana/agent".

> currently, supported platforms and archs are hardcode, so i want to make it as a list defined in a variable (still work in progress).

Both of these points make sense to me, thanks for clarifying :)

PR #1477, opened 2022-03-10
CHANGELOG: fix entries

rfratto:
This restores the ordering we try to keep for our CHANGELOG, and adds a
note into the pull_request_template for contributors to retain that
order.

I also did some formatting and reclassified/rephrased some entries.


tpaschalis (approved):
👍

PR #1478, opened 2022-03-10
Add new `ebpf` integration

tpaschalis:
#### PR Description
This PR adds a v2 integration based on cloudflare's [ebpf_exporter](https://github.com/cloudflare/ebpf_exporter). 

One limitation of the specific exporter is that it doesn't support uprobes/uretprobes. While I was able to fork the exporter and patch in support for those, I thought we'd keep this simple for the time being and create a first version which worked as-is.

The user configures a number of ebpf "programs", which are then attached to the correct place in the kernel. The exporter reads and decodes values from an eBPF map, exposing them as Prometheus metrics.

```
integrations:
  ebpf:
    programs:
    # Count timers fired in the kernel
    - name: timers
      metrics:
        counters:
          - name: timer_start_total
            help: Timers fired in the kernel
            table: counts
            labels:
              - name: function
                size: 8
                decoders:
                  - name: ksym
      tracepoints:
        timer:timer_start: tracepoint__timer__timer_start
      code: |
        BPF_HASH(counts, u64);

        // Generates function tracepoint__timer__timer_start
        TRACEPOINT_PROBE(timer, timer_start) {
            counts.increment((u64) args->function);
            return 0;
        }
```

No issue, this came out of the Hackathon :)

#### Notes to the Reviewer
If you agree with the general premise and idea, I can start working on adding documentation, tests, etc.

Do you think there should be behind an additional `ebpf_enabled` build tag? The current PR ~fails~ failed because it couldn't build the gobpf dependency on the Linux image that we run golangci-lint on. I don't think we should add ebpf/bcc bindings into our linting step, just to stop it from complaining, so an extra build tag might be a solution here.

I thought we'd start with a v2 integration outright, as a way to encourage more people to start using them. How do you feel about this?

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [ ] Tests updated


pkg/integrations/v2/ebpf/integration_stub.go
```diff
@@ -0,0 +1,25 @@
...
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+)
+
+func init() {
+	integrations.Register(&config{}, integrations.TypeSingleton)
+}
+
+type config struct{}
```
mattdurham, line 17: I think we need the config defined here also *or better yet in a shared file*. Unmarshal strict will fail if the fields cant find a home on a nonlinux platform.

pkg/integrations/v2/ebpf/integration.go
```diff
@@ -0,0 +1,87 @@
...
+
+func (c *config) NewIntegration(l log.Logger, globals integrations.Globals) (integrations.Integration, error) {
+	c.globals = globals
+	ebpf := &ebpfHandler{cfg: c}
+	h, err := ebpf.createHandler()
+	if err != nil {
+		return nil, err
+	}
+
+	return metricsutils.NewMetricsHandlerIntegration(l, c, c.common, globals, h)
```
tpaschalis, line 66: Thanks to @mattdurham for the suggestion to use the helper here!

pkg/integrations/v2/ebpf/integration_stub.go
tpaschalis: You're right! I opted to re-define the only yaml field we need here, instead of having a common file, as the stub config is slimmer.

It just felt better in the moment, but I'm not opposed to changing this.

pkg/integrations/v2/ebpf/integration.go
```diff
@@ -0,0 +1,87 @@
+//go:build linux && ebpf_enabled
+// +build linux,ebpf_enabled
```
rfratto, line 2: IMO for now I'd ready have stuff like this be opt out (i.e., the default behavior is to build everything and you use tags to disable specific things). 

Can the tag be `!noebpf`? We use `nodocker` and `nonetwork` for disabling specific types of tests.

tpaschalis: The issue which prompted the opt-in tag is that the Linux agent builds would fail unless some specific system packages are installed (due to the Go->eBPF bindings dependency, which in turn requires [specific kernel versions](https://github.com/iovisor/bcc/blob/master/docs/kernel-versions.md)).

```
$ go build .          # on linux system without support for bcc
# github.com/iovisor/gobpf/bcc
../../vendor/github.com/iovisor/gobpf/bcc/module.go:32:10: fatal error: 'bcc/bcc_common.h' file not found
#include <bcc/bcc_common.h>
         ^~~~~~~~~~~~~~~~~~
1 error generated.
```

I'm not sure what's an elegant way to handle this. Bundling the required C headers, and somewhat complicating our build pipeline?

rfratto: You can update tools/seego/Dockerfile (it has apt, it's Debian-based) to install dependencies. We do this already in the main Dockerfile to install [systemd headers](https://github.com/rfratto/seego/blob/7257eaba401374ce0a6aa53a63317475aff3905c/Dockerfile#L85-L89), and extending the build image to include our specific changes would be the way to do this IMO. 

I think we can assume the kernel version is new enough, the platforms we'll be mainly building on should have kernels newer than ones from 2017.

tpaschalis: I see. We'll also need to add the dependencies when [linting](https://github.com/grafana/agent/blob/b7dd9dbd22f4eeaf3b00112651d5e52599f90280/.drone/drone.yml#L21). 

My main concern here is that people who build from source on Linux, might be confused when `go build` fails with that cryptic message. On one hand, they would already need the systemd headers as you mentioned, but I'm not sure that bcc is also something they'd expect to need if they're not using eBPF.

I'll try to find a way to build this and report back :)

rfratto: That's fair :) We can document it [in our contribution guide](https://github.com/grafana/agent/blob/main/docs/developer/contributing.md#compiling-the-agent) (we should probably be mentioning systemd there too but we're not at the moment). Do you think that would be helpful enough?

tpaschalis: Yeah, that would make sense ^^

pkg/integrations/v2/ebpf/integration.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/grafana/agent/pkg/integrations/v2/metricsutils"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+type config struct {
+	Programs []ebpf_config.Program `yaml:"programs,omitempty"`
+
+	common  common.MetricsConfig
+	globals integrations.Globals
```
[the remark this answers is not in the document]
tpaschalis: You're right; after using the NewMetricsHandlerIntegration helper, they're no longer needed and they've been removed.

tpaschalis:
Commenting here to un-stale. I want to check if any new ebpf packaging has been contributed upstream, especially after we changed to a new Debian base image.

pkg/integrations/v2/ebpf/integration.go
```diff
@@ -0,0 +1,87 @@
...
+func (e *ebpfHandler) createHandler() (http.HandlerFunc, error) {
+	exp, err := exporter.New(ebpf_config.Config{Programs: e.cfg.Programs})
+	if err != nil {
+		return nil, fmt.Errorf("failed to create ebpf exporter with input config: %s", err)
+	}
+
+	err = exp.Attach()
+	if err != nil {
+		return nil, fmt.Errorf("failed to attach ebpf exporter: %s", err)
+	}
```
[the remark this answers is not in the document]
tpaschalis: No, I don't think that there will be any issues.

I tested on a Linux VPS, changing various bits of configuration, enabling/disabling the integration and reloading each time; it seemed to work nicely.

pkg/integrations/v2/ebpf/integration.go
tpaschalis: I added a small note about these two dependencies on our contribution guide https://github.com/grafana/agent/pull/1478/commits/323fb0084cadcf6a62a44c712a4fa8df5c2ef5ed.

Do we still want to have a `noebpf` flag to make it easier for people to contribute?

docs/developer/contributing.md
```diff
@@ -82,6 +82,12 @@ The Makefile provides several targets:
 * `test`: run the tests
 * `lint`: run linting checks
 
+### Compiling on Linux
+Compiling the Agent on Linux requires a couple of extra dependencies.
+These are 
+* [systemd headers](https://github.com/grafana/agent/blob/main/cmd/agent/Dockerfile#L8-L9) for Promtail
+* [bcc tools](https://github.com/grafana/agent/blob/main/cmd/agent/Dockerfile#L12-L13) for the eBPF integration
```
marctc, line 89: Is this also required even if you don't use the eBPF integration?

tpaschalis: Sorry for the delay Marc! This was implemented in b7bf8ecdc734c2b7328ccc35d0bcc0e1554ad533.

The eBPF integration will be built requiring the `bcc/bcc_common.h` only on Linux and when the `noebpf` tag is missing. Conversely, if the platform is not linux or _is_ Linux but the noebpf tag is passed, then the stub will be used.

marctc: Is it possible to do in on the other way around? Like you need to specify an `ebpf` tag in case you want to compile agent with ebpf integration?

tpaschalis: We discussed this with Robert [here](https://github.com/grafana/agent/pull/1478#discussion_r826909113) and ultimately preferred the opt-out approach. We're using opt-out in other places throughout the project (eg. `nodocker`, `norace`, `nonetwork` build tags) so it might be worth to continue that convention.

brendamuir (approved):

mattdurham (approved):
Looks good to me

PR #1479, opened 2022-03-10
packaging: fix issue where .debs and .rpms always overwrote config files

rfratto:
(Also add unit tests for the .deb/.rpms)


packaging/deb/control/postinst
```diff
@@ -26,11 +26,8 @@ case "$1" in
             usermod -a -G systemd-journal "$GRAFANA_AGENT_USER"
         fi
 
-        if [ ! -d /var/lib/grafana-agent ]; then
-          mkdir -p /var/lib/grafana-agent
-          chown $GRAFANA_AGENT_USER:$GRAFANA_AGENT_GROUP /var/lib/grafana-agent
-          chmod 770 /var/lib/grafana-agent
-        fi
+        chown $GRAFANA_AGENT_USER:$GRAFANA_AGENT_GROUP /var/lib/grafana-agent
```
tpaschalis, line 29: I thought of suggesting this last time around, but thought that if a user had manually changed things, they'd have some reason for doing so.

Should we leave ownership/permissions if the folder already exists?

rfratto: Is there a way we can give it permissions by default? It's set to the home directory of the user so it's being created with unpredictable permissions unless we force it (ditto with the config file)

rfratto:
Haha, drone is making this difficult. I'll figure that out tomorrow to see how to run the tests both locally and from CI.

rfratto:
I tried a few more times to get it running on Drone but had too many issues. It ran on GH Actions on my first attempt, so maybe we just keep it as a GHA job 🤷

mattdurham (approved):
LGTM

packaging/deb/control/postinst
tpaschalis: I'm not sure. I don't want to block things, let's go ahead with this, should be fine.

tpaschalis (approved):

PR #1480, opened 2022-03-10
tools/crow: make sure that default settings are applied to crow

rfratto:
Follow up from #1476

### DIFF
--- a/tools/crow/main.go
+++ b/tools/crow/main.go
@@ -28,8 +28,8 @@ func main() {
 	var (
 		fs = flag.NewFlagSet(os.Args[0], flag.ExitOnError)
 
-		serverCfg   server.Config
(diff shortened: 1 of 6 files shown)


tools/crow/main.go
```diff
@@ -28,8 +28,8 @@ func main() {
 	var (
 		fs = flag.NewFlagSet(os.Args[0], flag.ExitOnError)
 
-		serverCfg   server.Config
-		crowCfg     crow.Config
+		serverCfg   = server.DefaultConfig
+		crowCfg     = crow.DefaultConfig
```
rfratto, line 32: This is the main fix, but I'm also updating the jsonnet to be defensive

mattdurham (approved):
LGTM

Issue #852, opened 2021-08-27
Default config exposes /debug/pprof publicly


terorie:
## Issue

When installing the Grafana Cloud Agent (version 0.17.0) according to documented instructions on a server, the agent is going to expose port `12345` to any address (listening at `0.0.0.0`) by default.

This server is also exposing the Go pprof debugging API under `http://<server>:12345/debug/pprof/`.

This is a minor security vulnerability, and common across the Go ecosystem: https://mmcloughlin.com/posts/your-pprof-is-showing
It is likely that any default installations of Grafana Cloud Agent suffer from this problem.

`kubelet` had the same issue two years ago (CVE-2019-11248). It was rated CVSS 6.4.
- https://github.com/kubernetes/kubernetes/issues/81023
- https://www.cvedetails.com/cve/CVE-2019-11248/

### Method 1: All in one script

Follow Grafana Cloud's walkthrough instructions to install Grafana Agent.

### Method 2: Manual config generator

Run the following command:

```sh
grafana-agentctl cloud-config -u <user> -p "<api_key>"
```

Substitute `<user>` for your Grafana Cloud username (numeric) and `<api_key>` for your API key (beginning with `ey...`).

It would generate a config like so:

```yaml
...
server:
  http_listen_port: 12345
```

When looking at the server config, this will indeed run the service under `0.0.0.0:12345`: https://grafana.com/docs/agent/latest/configuration/server-config/

```
# HTTP server listen host. Used for Agent metrics, integrations, and the Agent
# API.
[http_listen_address: <string> | default = "0.0.0.0"]

# HTTP server listen port
[http_listen_port: <int> | default = 80]
```

Finally, on your server:

```
$ sudo netstat -ltupn | grep 12345
tcp6       0      0 :::12345                :::*                    LISTEN      15538/grafana-agent 
```

## Fix

The fix is simple: Default `http_listen_address` to `127.0.0.1`.

PS, I could not find any channels for responsible disclosure on your website nor this repo, so I decided to post it here, since the severity is quite low.

PR #1481, opened 2022-03-10
server: Change default HTTP/gRPC listen address

rfratto:
This commit changes the default listen addresses to be 127.0.0.1:12345
(for HTTP) and 127.0.0.1:12346 (for gRPC). This makes listening on all
interfaces opt-in rather than opt-out, avoiding accidental overexposure
of access.

Closes grafana/agent#852.

Additionally, the `-reload-addr` and `-reload-port` flags have been
removed as a follow up to #1476. Now that the HTTP and gRPC server are
static for the lifetime of the application, it is impossible for the
user to change their configuration file in such a way to cause it to
shut down while performing a reload. This means that the `-reload-addr`
and `-reload-port` no longer have a use and can be removed safely.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,16 @@
   deprecated server block fields. Changing a deprecated field will be ignored
   and cause the reload to fail. (@rfratto)
 
+- [CHANGE] The default HTTP listen address is now `127.0.0.1:12345`. Use the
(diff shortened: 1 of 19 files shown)


rfratto:
@mattdurham this is ready for review now :)

mattdurham (approved):
LGTM

PR #1484, opened 2022-03-11
Remove server.http.address flag in main k8s manifest :)

ushtipak:
#### PR Description
Remove server.http.address flag in main to fix CrashLoopBackOff when following docs grafana-easystart-app/integrations-management/integrations/kubernetes

N / A

#### Notes to the Reviewer
"Happy path" doesn't work when following agent installation on K8s (grafana-easystart-app/integrations-management/integrations/kubernetes)

### DIFF
--- a/production/kubernetes/agent-bare.yaml
+++ b/production/kubernetes/agent-bare.yaml
@@ -59,7 +59,6 @@ spec:
       containers:
       - args:
         - -config.file=/etc/agent/agent.yaml
-        - -server.http.address=0.0.0.0:12345
         command:
         - /bin/agent
         env:


rfratto:
Hi, thanks for the PR! I might merge this short term just so we don't break the guide, but I think the guide really shouldn't be pointing to main, which seems like it's asking for trouble when we need to make changes to prepare for future releases. 

Can you sign the CLA so we can merge?

cc @hjet

ushtipak:
Hi @rfratto,

I completely agree that guide shouldn't be pointing to main, but if you go to that quick-start page I've mentioned - that is the case... The ConfigMap for K8s integration is correctly dynamically created (populating Prometheus with actual endpoint), so similar concept could be used here, but for now, as you've said, this might be a temporary solution :smile:

If this option is removed from the latest version and is not planned of "returning", removing it from the manifest completely could be still useful :smiley: 

CLA signed, sorry for not doing it earlier :smile:

rfratto (approved):
I'm working with the team that maintains that guide to change what branch it's using :) Thanks for the fix! I'll revert after we've changed the link to something more stable.

PR #1485, opened 2022-03-11
Try using actions/cache@v2 to speed up macOS/Windows tests

rfratto:


rfratto:
I looked at the last ~25 test runs: on average, running the tests for Windows/macOS takes 22m43s. 

If you have the module cache already built, that average goes down to 15m04s (across 5 runs). The first run took 25m48s to build the cache, and subsequent runs will take somewhere in between to update the cache when a subset of dependencies change. 

I'd estimate that caching the modules makes Windows/macOS tests complete roughly 30% faster on average, which seems like a net improvement.

PR #1494, opened 2022-03-14
Restructure CHANGELOG to make type groupings more explicit

rfratto:
Opening as a draft to get some feedback. Adding explicit groupings to the CHANGELOG might make it easier to contribute to and read compared to seeing a `[CHANGETYPE]` label for every single line.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -8,9 +8,8 @@
 
 <!-- When updating the CHANGELOG, please:
 
-1. Keep entries grouped by SECURITY/FEATURE/ENHANCEMENT/BUGFIX/CHANGE/DEPRECATION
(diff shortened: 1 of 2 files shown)


rfratto:
You can see it rendered [on my branch](https://github.com/rfratto/agent/blob/restructure-changelog/CHANGELOG.md), with the [unrendered representation](https://github.com/grafana/agent/blob/main/CHANGELOG.md?plain=1) also being interesting.

rfratto:
I initially tried [sub-headers](https://github.com/rfratto/agent/blob/10e85a85d01d56c2e015b7dbc635016d2f8d402f/CHANGELOG.md) instead of nested bullet points, but I was concerned that the headers would make it harder to scan through the file and easily identify different releases.

rfratto:
Closing for now, we can return to this some other time.

PR #1497, opened 2022-03-15
fix(mixin): grafanaDashboardFolder is a top-level attribute

Duologic:
#### PR Description
The mixin used to make the `grafanaDashboardFolder` key a part of the dashboard json, this
is not correct. This PR moves the attribute to the same level as the `grafanaDashboards`,
as such it will be consumed by the Grafana jsonnet library as a mixin.

NA

#### PR Checklist

- [x] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,9 @@
 - [BUGFIX] Fix issue where installing the DEB/RPM packages would overwrite the
   existing config files and environment files. (@rfratto)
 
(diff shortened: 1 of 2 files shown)


Duologic:
The CLA link gives me a broken form.

Duologic:
🥳

rfratto (approved):

PR #1498, opened 2022-03-15
WIP otel upgrade

mapno:
WIP Otel upgrade to v0.46.0

PR #1499, opened 2022-03-15
Update Prometheus, OpenTelemetry-collector, Loki

### LINKED ISSUE #1376: Cherry pick remote_write shard-up fixes for agent Prometheus fork
Internally at Grafana we are running a _dev build_ of the Agent, with the changes for https://github.com/prometheus/prometheus/pull/9274 merged in. We'd like to get back onto proper Agent releases.

## Checking if grafana/prometheus has this fix

Following the [replace directive](https://github.com/grafana/agent/blob/ef00c8e9cd690e1690ae1a78256ac7e16c5dd8ec/go.mod#L382), we can see that we're running https://github.com/grafana/prometheus/commit/aae84190631a.

And examining [the code of the fork at this commit](https://github.com/grafana/prometheus/blob/aae84190631a914a94cadfa83c437c85de3721aa/storage/remote/queue_manager.go#L815-L818), we can see that the fix isn't there.

## Moving forward

Having asked around internally, it seems as if updating to Prometheus 2.33.0, which has that fix merged in, is not such an easy task. So I'd like to propose we cherry-pick the fix to the Agent's fork of Prometheus in order to unblock ourselves. In https://github.com/iainlane/prometheus/tree/grafana/2.31.0-grafana I've tested this and it cherry-picks cleanly. (Further review would be required.)

The first step then would be to propose my cherry-picks as a PR. ~~But the commit we're using in Agent isn't on any branch:~~

```
laney@melton(main|✔)> git branch --contains aae84190631a
laney@melton(main|✔)> 
```

~~so there is no target for review.~~ (edit: this is done now)

## Steps

I'd suggest the following steps

  - [x] Create a branch in `grafana/prometheus` at `aae84190631a` (someone with commit access to that repo)
  - [x] Create a pull request to bring https://github.com/iainlane/prometheus/tree/grafana/2.31.0-grafana into that branch (@iainlane) - done as https://github.com/grafana/prometheus/pull/29
  - [ ] Review that, iterate until happy (someone who knows Agent/Prometheus)
  - [ ] Update Agent's `replace` directive to use this (@grafana/agent-squad)
  - [ ] When happy, make an Agent release containing this new version of the Prometheus fork

And then we can get back onto real releases.

Thoughts (or edits to this issue) welcome. 👍

### LINKED ISSUE #1415: Update github.com/prometheus/prometheus to v2.33.4
An update for `github.com/prometheus/prometheus` (version `v2.33.4`) is now available. Version `v2.27.0` is currently in use.

rfratto:
This PR updates:

* Prometheus to grafana/prometheus release-2.33.5-grafana
* Loki to k87
* OpenTelemetry-collector to v0.46.0
* cAdvisor to v0.44.0

They all have to be updated at once: changes to the Prometheus package layout causing the others to need updating.

Supersedes #1498.
Closes #1415.
Closes #1376.


pkg/config/config.go
```diff
@@ -129,9 +128,6 @@ func (c Config) MarshalYAML() (interface{}, error) {
 	enc.SetHook(func(in interface{}) (ok bool, out interface{}, err error) {
 		// Obscure the password fields for known types that do not obscure passwords.
 		switch v := in.(type) {
-		case etcd.Config:
-			v.Password = "<secret>"
-			return true, v, nil
```
rfratto, line 134: etcd.Config no longer needs an override (and has changed types; it's no longer a string) so we can remove this.

pkg/operator/logutil/log.go
```diff
@@ -13,13 +13,13 @@ import (
 
 // Wrap wraps a log.Logger into a logr.Logger.
 func Wrap(l log.Logger) logr.Logger {
-	return &goKitLogger{l: l}
+	return logr.New(&goKitLogger{l: l})
```
rfratto, line 16: logr.Logger changed from an interface to a struct so we had to rework this file to make everything compile again.

pkg/operator/logutil/log.go
```diff
@@ -30,23 +30,15 @@ type goKitLogger struct {
...
+var _ logr.LogSink = (*goKitLogger)(nil)
 
-func (l *goKitLogger) Log(keyvals ...interface{}) error {
-	return l.namedLogger().Log(keyvals...)
+func (l *goKitLogger) Init(info logr.RuntimeInfo) {
+	// no-op
 }
 
-func (l *goKitLogger) Enabled() bool { return true }
+func (l *goKitLogger) Enabled(level int) bool { return true }
```
mattdurham, line 39: Is the intent to return if the specified log level is enabled? Or conversely are we hard coding to true to keep it the same as before?

rfratto: I'm just keeping the same behavior (where the logger is always enabled) but the interface changed so we had to add the unused parameter.

pkg/traces/config.go
```diff
@@ -323,8 +323,8 @@ func exporter(rwCfg RemoteWriteConfig) (map[string]interface{}, error) {
 	}
 
 	compression := rwCfg.Compression
-	if compression == compressionNone {
-		compression = ""
+	if compression == "" {
```
mattdurham, line 326: Was this an existing bug?

rfratto: Everything in pkg/traces was changed by @mapno, but I think what happened here is we needed to change our logic after the OTel upgrade to get things to pass again

mattdurham (approved):
Looks good!

mapno (approved):
Nice, LGTM

PR #1314, opened 2022-01-25
[dev.operator-integrations] Generate integrations config

rfratto:
#### PR Description 
This PR extends the operator config generator with the ability to generate an agent config which runs integrations.

This is not wired up anywhere, and no additional Kubernetes resources will be generated as of this commit.

#### Which issue(s) this PR fixes 
Related to #1224.

#### Notes to the Reviewer
See the `TestBuildConfigIntegrations` test for a full example of a hierarchy and the resulting agent config file. 

#### PR Checklist

- [x] CHANGELOG updated (N/A) 
- [x] Documentation added (N/A)
- [x] Tests updated

PR #1503, opened 2022-03-15
[dev.operator-integrations] pkg/operator/config: Generate config for integrations

rfratto:
This PR adds the ability to generate an integrations-specific agent config for the operator. It is not wired in anywhere; nothing will currently generate the config.

Related to #1414 
(Supersedes the original attempt at #1314)

### DIFF
--- a/pkg/operator/config/templates/agent-integrations.libsonnet
+++ b/pkg/operator/config/templates/agent-integrations.libsonnet
@@ -1,5 +1,5 @@
-// agent-metrics.libsonnet is the entrypoint for rendering a Grafana Agent
-// config file for metrics based on the Operator custom resources.
+// agent-integrations.libsonnet is the entrypoint for rendering a Grafana Agent
+// config file for integrations based on the Operator custom resources.
 //
 // When writing an object, any field will null will be removed from the final
 // YAML. This is useful as we don't want to always translate unfilled values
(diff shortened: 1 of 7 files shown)


pkg/operator/config/templates/agent-integrations.libsonnet
```diff
@@ -0,0 +1,135 @@
...
+local new_logs_instance = import './logs.libsonnet';
+local new_metrics_instance = import './metrics.libsonnet';
+local new_external_labels = import 'component/metrics/external_labels.libsonnet';
+local new_remote_write = import 'component/metrics/remote_write.libsonnet';
+
+local calculateShards(requested) =
+  if requested == null then 1
+  else if requested > 1 then requested
+  else 1;
+
```
captncraig, line 27: This is a little confusing. Is this because the default is zero? I'd kinda prefer to explode here if somehow you pass a negative. But maybe this is a deep enough utility this behavior is safe enough?

rfratto: Ah, right, I copied this from the existing metrics code since `agent-integrations.libsonnet` is a combination of metrics + logs + integrations. 

We do the logic here because `0` is the default, and we want at least 1 shard. We could probably do a better job at validating input / applying the proper defaults before we pass things over to the jsonnet generator, though.

PR #1509, opened 2022-03-16
integrations-next: Support in-memory autoscraping

rfratto:
This PR extends `pkg/server` to include a way to form in-memory connections to the HTTP and gRPC server. It then modifies integrations-next to use this for the autoscrape SD and scraper.

Two worthy things to note:
* This uses github.com/rfratto/ckit for the in-memory net.Listeners, which doesn't have understood performance or resource utilization stats yet. (Plus, since it exists in my own user, contributing to it is bottlenecked by me, though I'm happy to give other people commit access)
* This uses an extra [change to Prometheus](https://github.com/grafana/prometheus/commit/bee3c63e5e7256ff82ab6d4ba1bcf2e2d5812a25) that I haven't opened a PR for upstream. I'm going to do that, but I'm not positive whether it'll get accepted.


tpaschalis:
This looks okay at a first glance, I'll take a proper view by tomorrow. The change to Prometheus was ultimately upstreamed in https://github.com/prometheus/prometheus/commit/44a5e705be503e5fb171cb67e9c47237873e6c49, right?

rfratto:
> This looks okay at a first glance, I'll take a proper view by tomorrow. The change to Prometheus was ultimately upstreamed in https://github.com/prometheus/prometheus/commit/44a5e705be503e5fb171cb67e9c47237873e6c49, right?

Yes 👍

docs/user/configuration/flags.md
```diff
@@ -78,6 +79,7 @@ These flags require the `dynamic-config` feature to be enabled:
 * `-server.grpc.max-concurrent-streams` Maximum number of concurrent gRPC streams (0 = unlimited)
 * `-server.grpc.max-recv-msg-size-bytes` Maximum size in bytes for received gRPC messages
 * `-server.grpc.max-send-msg-size-bytes` Maximum size in bytes for send gRPC messages
+* `-server.grpc.in-memory-addr`: Internal address used for the agent to make in-memory gRPC connections to itself. (default `agent.internal:12346`).
```
mattdurham, line 82: To be clear, this in no way actually opens a port for 12346? May want to explicitly state that.

mattdurham:
Are metrics also exposed in the normal http/https manner? (Looping over the listeners leads me to believe they are) If so we may want to explicitly call that out.

mattdurham (approved):
Looks good, aside from a bit of thoughts on verbiage/notes

PR #1514, opened 2022-03-17
Show curl progress bar agent install

mshahzeb:
Adds progress bar and removes silent flag when downloading agent deb package via curl. Sometimes the deb package can take a long time to download - Showing progress bar will provide feedback on when the agent install will complete.

Ref: https://curl.se/docs/manpage.html


rfratto (approved):
Thanks!

PR #1516, opened 2022-03-18
[dev.frontend-receiver] Add traces support to app o11 receiver

domasx2:
#### PR Description
* App o11y receiver integration can receive traces (otel format) in the payload and forward them to a tracing instance on the agent.
* Log, measurement, exception events can be annotated with trace id and span id. Tese are logged in a conventional way that Grafana understands

#### Notes to the Reviewer
Design doc: https://docs.google.com/document/d/1JkXQiMH_FL_jdY8B4zJZ_0416mm1TsB0d5N8w5JqNCs/edit
Based on Mario's suggested approach: https://github.com/grafana/agent/pull/1046/files

It's kinda hacky. Otel collector deliberately hides internals to not allow easy programatic access to a consumer. This PR  Adds a stub `push_receiver` to tracing pipelines that exposes a `consumer.Traces` instance via it's factory. Hey, it works though

#### PR Checklist

- [ ] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated


pkg/integrations/v2/app_o11y_receiver/exporters/traces.go
```diff
@@ -0,0 +1,35 @@
...
+	return "traces exporter"
+}
+
+// Export implements the AppDataExporter interface
+func (te *TracesExporter) Export(payload models.Payload) error {
+	if payload.Traces == nil {
+		return nil
+	}
+	if te.factory.Consumer != nil {
+		return te.factory.Consumer.ConsumeTraces(context.Background(), payload.Traces.Traces)
```
mapno, line 32: I would try to use the request's context

pkg/integrations/v2/app_o11y_receiver/app_o11y_receiver.go
```diff
@@ -88,27 +90,45 @@ func (c *Config) NewIntegration(l log.Logger, globals integrations.Globals) (int
...
+		exp = append(exp, lokiExporter)
+	}
+
+	if len(c.ExporterConfig.TracesInstance) > 0 {
+		tracesInstance := globals.Tracing.Instance(c.ExporterConfig.TracesInstance)
+		if tracesInstance == nil {
+			return nil, fmt.Errorf("traces instance \"%s\" not found", c.ExporterConfig.TracesInstance)
+		}
+
+		factory := tracesInstance.GetFactory(component.KindReceiver, "push_receiver")
```
mapno, line 124: Nit: use push_receiver's const

```suggestion
		factory := tracesInstance.GetFactory(component.KindReceiver, pushreceiver.TypeStr)
```

pkg/traces/config.go
```diff
@@ -434,6 +435,8 @@ func (c *InstanceConfig) otelConfig() (*config.Config, error) {
 	if len(c.Receivers) == 0 {
 		return nil, errors.New("must have at least one configured receiver")
 	}
+	c.Receivers[pushreceiver.TypeStr] = nil // add a hacky push receiver for when an integration
```
mapno, line 438: I'd try to avoid to add the push_receiver always if possible. Maybe we can add a non-yaml field to the traces config that is set based on the integrations config?

mapno (commented):
Left a few comments. Overall, lgtm.

pkg/traces/config.go
domasx2: I wanted to do that initially, but cound't figure out a way that didn't require significant refac... Didn't occur to me to use a non-yaml field though. Thanks for the hint, will investigate

domasx2: - [ ] only insert push receiver for traces instances used by the frontend receiver integration

pkg/integrations/v2/app_o11y_receiver/exporters/traces.go
domasx2: - [ ] use request's context

pkg/traces/config.go
domasx2: took some doing, but got it done!

mapno (approved):
Nice, thanks for addressing the comments. It looks good on the traces side :)

pkg/integrations/v2/app_o11y_receiver/exporters/exporter.go
```diff
@@ -1,11 +1,13 @@
...
+	"context"
+
 	"github.com/grafana/agent/pkg/integrations/v2/app_o11y_receiver/models"
 )
 
 // AppO11yReceiverExporter is the base interface
 type AppO11yReceiverExporter interface {
 	Name() string
-	Export(payload models.Payload) error
+	Export(ctx context.Context, payload models.Payload) error
```
kpelelis, line 12: I am not a huge fan of this change as it breaks the current API but to be honest, context is a runtime argument and it can be potentially useful to other consumers so 👍

pkg/integrations/v2/app_o11y_receiver/app_o11y_receiver.go
```diff
@@ -86,38 +91,56 @@ func (c *Config) NewIntegration(l log.Logger, globals integrations.Globals) (int
...
 	})
 
 	var exp = []exporters.AppO11yReceiverExporter{
-		// Loki
-		lokiExporter,
 		receiverMetricsExporter,
 	}
 
-	handler := handler.NewAppO11yHandler(c.ExporterConfig, exp, reg)
+	if len(c.ReceiverConfig.LogsInstance) > 0 {
```
kpelelis, line 104: Ι think we should make both the logs and the tracing instances hard dependencies (i.e. if the integration is enabled and the instances are missing for some reason, exit), or have default fallbacks

kpelelis (changes requested):
Alright there are a lot of different things going on here. I will explain the reason that I am rejecting this. First of all, thank for taking on with this. It is really difficult thing to do given the current architecture.

I would suggest breaking this pull request into quantifiable/reversable commits as there are distinct changes that are contextually different (e.g. changing the default configuration of CORS origins).
Second thing is making the logs and traces instances a soft dependency. Can we make sure that things wont break if we bootstrap the agent with just one of the two (or none) and we get a payload that needs to use the missing modules?
Last but not least, although not a reject reason, I will leave the core changes of the tracing receiver to be reviewed by someone in the agent team

Overall great work 🚀

pkg/integrations/v2/app_o11y_receiver/app_o11y_receiver.go
domasx2: Hm, I'm not certain that every user will have both logging and tracing backends and want to use them with frontend, would rather allow flexibility. It doesn't cost us anything

pkg/integrations/v2/app_o11y_receiver/app_o11y_receiver.go
```diff
@@ -73,7 +78,7 @@ var (
 	_ integrations.MetricsIntegration = (*appo11yIntegration)(nil)
 )
 
-// NewIntegration converts this config into an instance of an integratin
```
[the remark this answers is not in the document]
domasx2: - [ ] fix typo

pkg/config/config.go
```diff
@@ -186,6 +187,24 @@ func (c *Config) Validate(fs *flag.FlagSet) error {
 
 	c.Metrics.ServiceConfig.APIEnableGetConfiguration = c.EnableConfigEndpoints
 
+	// find every traces instance that is to be used by app o11y receiver
+	// and configure it to include a PushReceiver to be able to push traces internally
+	// this cannot happen in unmarshall stage since integrations are not yet unmarshalled
+	if c.Integrations.configV2 != nil {
+		for _, conf := range c.Integrations.configV2.Configs {
+			if conf.Name() == app_o11y_receiver.IntegrationName {
```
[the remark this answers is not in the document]
domasx2: - [ ] avoid nesting

domasx2:
> I would suggest breaking this pull request into quantifiable/reversable commits as there are distinct changes that are contextually different (e.g. changing the default configuration of CORS origins).

Sorry, I'm gonna need some convincing before spending time on this :) the whole thing is gonna be squashed into a single commit anyway, doesn't feel worthile to worry about it 

> Second thing is making the logs and traces instances a soft dependency. Can we make sure that things wont break if we bootstrap the agent with just one of the two (or none) and we get a payload that needs to use the missing modules?

That's exactly how it is. If you don't specify logs/traces instance, receiver wont log or send traces. If you specify an invalid instance, agent will fail to start with appropriate error

Last but not least, although not a reject reason, I will leave the core changes of the tracing receiver to be reviewed by someone in the agent team
 
> Changelog should be update as well probably

It's updated on `dev.frontend-receiver` branch where I recently merged main with refactored changelog

pkg/integrations/v2/app_o11y_receiver/app_o11y_receiver.go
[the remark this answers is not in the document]
domasx2: love culinary references :D

kpelelis (approved):

PR #1518, opened 2022-03-18
Reformat changelog to improve readability

rfratto:
Follow up on #1494. 

This PR reformats the entirety of the changelog to be (subjectively) more readable. There are now h3-level headers for change types rather than changes being tagged with `[TYPE]`. 

I've also changed the order in which we show changes to:

1. Security fixes 
2. Breaking changes 
3. Deprecations 
4. Features 
5. Enhancements
6. Bugfixes
7. Other changes 

Previously, the order was:

1. Features 
2. Enhancements 
3. Bugfixes 
4. Changes (breaking + non-breaking) 
5. Deprecations 

Security fixes has been given its own section rather than being merged in as an enhancement or bugfix. I have hopefully chosen an order which puts a focus on things we want people to read first before digging into more details.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,3 +1,13 @@
+<!--
+
+CONTRIBUTORS GUIDE: https://github.com/grafana/agent/blob/main/docs/developer/contributing.md#updating-the-changelog
+
+If this is your first PR or you have not contributed in a while, we recommend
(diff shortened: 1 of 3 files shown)


tpaschalis (approved):
I like the changes here. Approving, if anything, for the tremendous amount of work refactoring the previous changelog entries.

Hopefully, there wasn't too much vim magic there :)

Issue #1496, opened 2022-03-14
Implement new metrics subsystem (RFC-0003)


rfratto:
This is a high-level issue to track the implementation of [RFC-0003: new metrics subsystem](https://github.com/grafana/agent/blob/main/docs/rfcs/0003-new-metrics-subsystem.md).

PR #1527, opened 2022-03-22
pkg/cluster: introduce clustering package

rfratto:
#### PR Description
This commit introduces a [ckit-based](https://github.com/rfratto/ckit) pkg/cluster package as a top-level subsystem for clustering.

Related to #1496.

#### Notes to the Reviewer
Currently, the package is not wired in anywhere, but tests are included for validating correctness. This will be used for #1496 over time.

#### PR Checklist

- [x] CHANGELOG updated (N/A)
- [x] Tests updated

### DIFF
--- a/pkg/cluster/gossip.go
+++ b/pkg/cluster/gossip.go
@@ -24,6 +24,25 @@ import (
 // extraDiscoverProviders used in tests.
 var extraDiscoverProviders map[string]discover.Provider
 
+// tokensPerNode is used to decide how many tokens each node should be given in
+// the hash ring. All nodes must use the same value, otherwise they will have
+// different views of the ring and assign work differently.
+//
+// Using 256 tokens strikes a good balance between distribution accuracy and
+// memory consumption. A cluster of 1,000 nodes with 256 tokens per node
+// requires 6MB for the hash ring, while 12MB is used for 512 tokens per node.
+//
+// Distribution accuracy measures how close a node was to being responsible for
+// exactly 1/N keys during simulation. Simulation tests used a cluster of 10
+// nodes and hashing 100,000 random keys:
+//
+//    256 tokens per node: min 94.0%, median 96.3%, max 115.3%
+//    512 tokens per node: min 96.1%, median 99.9%, max 103.2%
+//
+// While 512 tokens per node is closer to perfect distribution, 256 tokens per
+// node is good enough, optimizing for lower memory usage.
+const tokensPerNode = 256
+
 // GossipConfig controls clustering of Agents through gRPC-based gossip.
 // GossipConfig cannot be changed at runtime.
 type GossipConfig struct {
@@ -157,7 +176,7 @@ func NewGossipNode(l log.Logger, srv *grpc.Server, c *GossipConfig) (*GossipNode
 		l = log.NewNopLogger()
 	}
 
-	sharder := shard.Ring(256)
+	sharder := shard.Ring(tokensPerNode)
 
 	ckitConfig := ckit.Config{
 		Name:          c.NodeName,
@@ -182,16 +201,10 @@ func NewGossipNode(l log.Logger, srv *grpc.Server, c *GossipConfig) (*GossipNode
 
 // ChangeState changes the state of n. ChangeState will block until the state
 // change has been receieved by another node; cancel the context to stop
-// waiting.
+// waiting. ChangeState will fail if the current state cannot move to the
+// target state.
 //
-// Nodes must be a StateParticipant to be returned during Lookup.
-//
-// The following state transitions are valid:
-//
-// 		StatePending -> StateParticipant|StateViewer|StateTerminating|StateGone
-// 		StateParticipant -> StateTerminating|StateGone
-// 		StateViewer -> StateTerminating|StateGone
-// 		StateTerminating -> StateGone
+// Nodes must be a StateParticipant to receive writes.
 func (n *GossipNode) ChangeState(ctx context.Context, to peer.State) error {
 	if !n.started.Load() {
 		return fmt.Errorf("node not started")
--- a/pkg/cluster/gossip.go
+++ b/pkg/cluster/gossip.go
@@ -1,0 +1,248 @@
+package cluster
+
+import (
+	"context"
+	"fmt"
+	"io"
+	stdlog "log"
+	"net"
+	"os"
+
+	"github.com/go-kit/log"
+	"github.com/grafana/dskit/flagext"
+	"github.com/hashicorp/go-discover"
+	"github.com/hashicorp/go-discover/provider/k8s"
+	"github.com/rfratto/ckit"
+	"github.com/rfratto/ckit/advertise"
+	"github.com/rfratto/ckit/chash"
+	"github.com/rfratto/ckit/clientpool"
+	"go.uber.org/atomic"
+	"google.golang.org/grpc"
+)
+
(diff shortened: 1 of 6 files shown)


pkg/cluster/gossip.go
```diff
@@ -0,0 +1,253 @@
...
+
+		addrs, err := d.Addrs(c.DiscoverPeers, stdlog.New(io.Discard, "", 0)) // TODO(rfratto): log to log.Logger?
+		if err != nil {
+			return fmt.Errorf("discovering peers: %w", err)
+		}
+		c.JoinPeers = addrs
+	}
+
+	for i := range c.JoinPeers {
+		// Default to using the same advertise port as the local node. This may
```
mattdurham, line 118: Thoughts on making this mandatory?

rfratto: We can't in all cases 😞 The port can be mandatory if you're providing a hard-coded list, but discovered peers from go-discover don't always have ports (they didn't when I used the k8s provider)

pkg/cluster/gossip.go
```diff
@@ -0,0 +1,253 @@
...
+// itself as a gRPC service to srv. GossipConfig is expected to be valid and
+// have already had ApplyDefaults called on it.
+//
+// GossipNode operations are unavailable until the node is started.
+func NewGossipNode(l log.Logger, srv *grpc.Server, c *GossipConfig) (*GossipNode, error) {
+	if l == nil {
+		l = log.NewNopLogger()
+	}
+
+	sharder := shard.Ring(256)
```
mattdurham, line 160: This looks like a magic number. Is this a carryover from ckit?

rfratto: Yeah, you can specify the number of tokens per node in the hash ring but I'm just hard coding it. Making it configurable would cause issues if people get their nodes out of sync. I can make it a package-level constant with a comment explaining this. WDYT?

mattdurham: A constant is fine. I was looking for more why 256 and not 512 (or some other value)? I don't have any knowledge of the trade-offs between the two.

pkg/cluster/gossip.go
```diff
@@ -0,0 +1,253 @@
...
+	return &GossipNode{
+		cfg:       c,
+		innerNode: inner,
+		log:       l,
+		sharder:   sharder,
+	}, nil
+}
+
+// ChangeState changes the state of n. ChangeState will block until the state
+// change has been receieved by another node; cancel the context to stop
```
mattdurham, line 184: nit: received

pkg/cluster/gossip.go
```diff
@@ -0,0 +1,253 @@
...
+			n.started.Store(true)
+		}
+	}()
+	return n.innerNode.Start(n.cfg.JoinPeers)
+}
+
+// Stop leaves the cluster and terminates n. n cannot be re-used after
+// stopping.
+//
+// It is advisble to ChangeState to StateTerminating and StateGone before
```
mattdurham, line 249: nite: advisable

pkg/cluster/gossip.go
```diff
@@ -0,0 +1,253 @@
...
+
+// ChangeState changes the state of n. ChangeState will block until the state
+// change has been receieved by another node; cancel the context to stop
+// waiting.
+//
+// Nodes must be a StateParticipant to be returned during Lookup.
+//
+// The following state transitions are valid:
+//
+// 		StatePending -> StateParticipant|StateViewer|StateTerminating|StateGone
```
mattdurham, line 191: Does StatePending exist? StateViewer seems to be the default state

mattdurham (changes requested):
StatePending is the only thing I have a real question on, everything else looks good aside from a few nits.

pkg/cluster/gossip.go
rfratto: Ah, TL;DR 256 tokens gives you decent enough distribution between nodes while using half of the memory. 

Long answer:

There's a [test](https://github.com/rfratto/ckit/blob/0091693232407b3f8ecc5be4df19e86e0700704c/internal/chash/chash_test.go#L64) in ckit which reports on the utilization accuracy of each algorithm when using 10 nodes:

```
ring 256 tokens distribution stats: min 94.0%, median 96.3%, max 115.3%
ring 512 tokens distribution stats: min 96.1%, median 99.9%, max 103.2%
```

Some explanation of the numbers here:
* 100% means a node saw perfect distribution (given N nodes, it was assigned exactly 1/N of the total load after performing 10,000 * N lookups). 
* >100% means a node received more work than other nodes, <100% means a node received less work than other nodes 
* min/median/max refers to the most underdistributed/median distribution/most overdistributed node during the test. 

IMO, a median distribution accuracy of 96% when using 256 tokens is good enough, and 94%/115% being the worst cases for over/under distributed are likely acceptable. 

If we look at the memory usage of how this scales to a wide number of nodes:

```
BenchmarkHashResource/1_nodes/ring_256_tokens-10         	      9948 ns/op	    6232 B/op
BenchmarkHashResource/1_nodes/ring_512_tokens-10         	     23798 ns/op	   12376 B/op
BenchmarkHashResource/10_nodes/ring_256_tokens-10        	    216016 ns/op	   65624 B/op
BenchmarkHashResource/10_nodes/ring_512_tokens-10        	    487229 ns/op	  122968 B/op
BenchmarkHashResource/50_nodes/ring_256_tokens-10        	   1330522 ns/op	  311384 B/op
BenchmarkHashResource/50_nodes/ring_512_tokens-10        	   2890132 ns/op	  614488 B/op
BenchmarkHashResource/100_nodes/ring_256_tokens-10       	   2981066 ns/op	  614490 B/op
BenchmarkHashResource/100_nodes/ring_512_tokens-10       	   6330703 ns/op	 1228888 B/op
BenchmarkHashResource/500_nodes/ring_256_tokens-10       	  16414820 ns/op	 3072090 B/op
BenchmarkHashResource/500_nodes/ring_512_tokens-10       	  34621860 ns/op	 6144088 B/op
BenchmarkHashResource/1000_nodes/ring_256_tokens-10      	  34484273 ns/op	 6144088 B/op
BenchmarkHashResource/1000_nodes/ring_512_tokens-10      	  75791608 ns/op	12288088 B/op
```

You can see that 256 tokens uses roughly half the memory and is twice as fast to construct the ring (the hash times are similar though; ring lookups is `O(log N)`.

My takeaway from all of the investigation is that you'll hit diminishing returns fairly quickly while the time to construct the ring and memory used for the ring continues to grow exponentially.

pkg/cluster/gossip.go
rfratto: Stale comment, I need to update this or stop trying to sync it with the upstream comment

rfratto: (but yes, StatePending used to exist and was since removed)

mattdurham (approved):

pkg/cluster/gossip.go
tpaschalis, line 249: It might be worth considering to make nodes perform these state transitions themselves when shutting down

tpaschalis (approved):
🚢

pkg/cluster/gossip.go
rfratto: 👍 we'll probably end up doing this, we'll need to see once everything is wired in if we can automate the shutdown transitions and stop exposing this code

PR #1529, opened 2022-03-24
Set timeout 2x as default HTTP Transport for agentctl

### LINKED ISSUE #1410: "context deadline exceeded" during agent install - Cause found
I was attempting to install the agent for communicating with my cloud account. I got the install to run and kept getting "Failed to retrieve config". I found the function in the shell script that called `grafana-agentctl cloud-config` and invoking that manually with my stack ID and API key, I got a little more verbose of an error message:

```
could not retrieve agent cloud config: failed to make request: Get "https://integrations-api-us-central.grafana.net/stacks/<my ID>/agent_config": context deadline exceeded
```

However, invoking this through curl worked:

```
curl -H "Authorization: Bearer eyJ<my API key>" https://integrations-api-us-central.grafana.net/stacks/<my ID>/agent_config
```

Response:

```
{"status":"success","data":{"integrations":{"node_exporter":{"enabled":true},"prometheus_remote_write":[{"basic_auth":{"password":"eyJ<removed>},"server":{"http_listen_port":12345}}}
```

On a hunch, I checked my `/etc/resolv.conf` and the first DNS server was a (currently) down DNS resolver. I commented that out and `grafana-agentctl` worked, despite having a second working resolver.

**Bug:**

Make `grafana-agentctl` more resilient in DNS lookups.

marctc:
#### PR Description

Setting a timeout for the execution of this command lower than the HTTP Transport timeout (30s) 
leads to show messages that are not very clear to users when this is exceeded as it explained in 
#1410. This PR increased the timeout for  `context.Background()` to 60s. 

Change this would  trigger first the HTTP client transport timeout first and show 
`dial tcp: i/o timeout`.

Fixes #1410

### DIFF
--- a/cmd/agentctl/main.go
+++ b/cmd/agentctl/main.go
@@ -4,6 +4,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"os"
 	"path/filepath"
 	"sort"
@@ -452,13 +453,13 @@ config that may be used with this agent.`,
 				return fmt.Errorf("--api-key must be provided")
 			}
 
-			cli := grafanacloud.NewClient(nil, apiKey, apiURL)
-
 			// setting timeout 2x as the default HTTP transport timeout (30s)
-			ctx, cancel := context.WithTimeout(context.Background(), time.Second*60)
-			defer cancel()
+			httpClient := &http.Client{
+				Timeout: time.Minute,
+			}
+			cli := grafanacloud.NewClient(httpClient, apiKey, apiURL)
 
-			cfg, err := cli.AgentConfig(ctx, stackID)
+			cfg, err := cli.AgentConfig(context.Background(), stackID)
 			if err != nil {
 				fmt.Fprintf(os.Stderr, "could not retrieve agent cloud config: %s\n", err)
 				os.Exit(1)
--- a/cmd/agentctl/main.go
+++ b/cmd/agentctl/main.go
@@ -454,7 +454,8 @@ config that may be used with this agent.`,
 
 			cli := grafanacloud.NewClient(nil, apiKey, apiURL)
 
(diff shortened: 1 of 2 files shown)


cmd/agentctl/main.go
```diff
@@ -454,7 +454,8 @@ config that may be used with this agent.`,
 
 			cli := grafanacloud.NewClient(nil, apiKey, apiURL)
 
-			ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
+			// setting timeout 2x as the default HTTP transport timeout (30s)
+			ctx, cancel := context.WithTimeout(context.Background(), time.Second*60)
```
rfratto, line 458: Sorry, I meant setting the read timeout on the HTTP client itself:

```go
cli := &http.Client{
  Timeout: time.Minute, 
}

cli := grafanacloud.NewClient(cli, apiKey, apiUrl) 

cfg, err := cli.AgentConfig(context.Background(), stackID)
// ...
```

marctc:
> LGTM, but IMO we should still set a HTTP client read timeout to something that's ~2-3x bigger than the transport timeout just in case the connection established but the server isn't responding.

Done, could you take a second look? Thanks!

cmd/agentctl/main.go
rfratto: I think the timeout on the http.Client should give more detailed error messages; let me double check.

rfratto: OK, yeah, it looks like http.Client.Timeout will add `(Client.Timeout exceeded while awaiting headers)` to the error messages, while the context itself won't.

marctc: Cool, thanks for checking!

rfratto (approved):
LGTM, thank you! Before I merge, can you add this to the changelog?

PR #1532, opened 2022-03-29
Introduce endpoint for accepting remote_write requests

### LINKED ISSUE #1447: Support remote_write interface to push metrics to Agent
We should support an endpoint to send Prometheus metrics directly to a metrics instance, similar to the Prometheus [/api/v1/write](https://prometheus.io/docs/prometheus/latest/storage/#overview) endpoint. 

We couldn't use the same URL; it'd have to be scoped to an instance, like `/agent/api/v1/metrics/instance/<instance name>/write`.

This shouldn't be terribly difficult to implement, see the [Prometheus implementation for reference](https://github.com/prometheus/prometheus/pull/8424). The important bit is that metrics from the endpoint are pushed directly into a WAL.

tpaschalis:
#### PR Description
This PR adds a new endpoint which allows users to push metrics directly into an instance's WAL. 
To do that, we reuse Prometheus' [writeHandler](https://github.com/prometheus/prometheus/blob/83a2e52bc2b9962aeb27071346abeab9fc860434/storage/remote/write_handler.go#L44-L66) which expects a snappy-compressed request that can be unmarshalled into a `prompb.WriteRequest`.

This PR fixes #1447, where the idea was detailed.

#### Notes to the Reviewer
I've added some rudimentary tests, mostly following the tests in the Prometheus implementation. If we think makes the http_test.go file too large, and since the handler is already tested upstream, I could add a different/more minimal kind of test.

To see this in action on a 'real' instance, you can do the following:

1) Checkout the PR, build and run the agent with a config like the following:
```
server:
  log_level: info

metrics:
  global:
    scrape_interval: 1m
  wal_directory: /tmp/wal
  configs:
    - name: tpaschalisInstance
      remote_write:
      - url: https://prometheus-prod-01-eu-west-0.grafana.net/api/prom/push
        basic_auth:
          username: 297344
          password: ey<my-api-key>
```

2) Go to `TestRemoteWriteHandler` and dump the contents of a 'valid' push request using something like `os.WriteFile("/tmp/request-bytes", buf, 0644)`

3) Use cURL to POST the binary data to the new endpoint like
```
curl -i -X POST --data-binary "@/tmp/request-bytes" http://localhost:12345/agent/api/v1/metrics/instance/tpaschalisInstance/write
```

4) Go to the Prometheus instance and see that the data has been recorded correctly.
image

#### PR Checklist

- [X] CHANGELOG updated
- [ ] Documentation added
- [X] Tests updated


pkg/metrics/http.go
```diff
@@ -125,3 +129,36 @@ type TargetInfo struct {
...
+	// Get the metrics instance and serve the request.
+	managedInstance, err := a.InstanceManager().GetInstance(instanceName)
+	if err != nil || managedInstance == nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	handler := remote.NewWriteHandler(a.logger, managedInstance)
+	handler.ServeHTTP(w, r)
+}
```
rfratto, line 152: IMO this is something that's so obviously correct at a glance that we might not need the test, especially since the test is pretty large. Up to you to keep/remove though :)

rfratto (approved):
LGTM, thanks! This needs a changelog entry as a feature, I think.

rfratto:
(This also needs to be documented in our API docs)

pkg/metrics/http.go
tpaschalis: I didn't know what the team would prefer, but I'd also be inclined to remove it, since it's mostly reiterating the upstream Prometheus test.

I'll remove it and see if anyone else raises a concern.

docs/user/api/_index.md
```diff
@@ -220,6 +220,19 @@ Response on success:
...
 ```
 
+### Accept remote_write requests
+
+```
+POST /agent/api/v1/metrics/instance/{instance}/write
+```
+
+This endpoint accepts Prometheus-compatible remote_write POST requests, and
+appends their contents into an instance's WAL.
```
rfratto, line 230: Can we add a sentence to replace `{instance}` with the instance name from the config? Something like:

``````
Replace `{instance}` with the name of the metrics instance from your config file. For example, this config file defines the instance names "dev" and "prod":

```yaml
metrics:
  configs:
  - name: dev  # /agent/api/v1/metrics/instance/dev/write
  - name: prod # /agent/api/v1/metrics/instance/prod/write
```
``````

tpaschalis: Yeah, you're right. I've added an extra paragraph, let me know how it looks!

rfratto (approved):
🎉

PR #1535, opened 2022-03-29
builds for linux/ppc64le

aklyachkin:
#### PR Description

This PR adds builds for linux/ppc64le into Makefile. 

#### Notes to the Reviewer

It builds correctly on my test system and creates all the packages. Obvious change to add a new architecture. ppc64le is in use on some of the biggest supercomputers (https://www.top500.org/lists/top500/2021/11/ - Nr. 2 and 3), on OpenPOWER (https://osuosl.org/, CERN, etc.) and on IBM Power Systems.

#### PR Checklist

- [X] CHANGELOG updated
- [ ] Documentation added - no platform-dependent documentation
- [ ] Tests updated - no platform-dependent tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,6 +84,8 @@ Main (unreleased)
 
 - Introduced endpoint that accepts remote_write requests and pushes metrics data directly into an instance's WAL. (@tpaschalis)
 
+- Added builds for linux/ppc64le. (@aklyachkin)
+
 ### Enhancements
 
 - Tracing: Exporters can now be configured to use OAuth. (@canuteson)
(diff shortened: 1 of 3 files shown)


rfratto:
Thanks! The PR is straightforward, but I think we should be considerate about whether we're able to easily officially support ppc64le. We've had OS/architecture related bugs in the past, so there's a non-zero maintenance cost for adding a new one. 

WDYT @mattdurham?

mattdurham:
Can we add EXPERIMENTAL/BETA verbiage to OS/arches? I am okay with making it a binary someone can download; less fine saying it's something with the same support level as x64 linux.

aklyachkin:
I am OK with that. Most of projects, which have ppc64le builds, says it is "Tier 2" or "experimental". It is difficult to get access to ppc64le hardware, if you never had it, and yes, it there could be some platform specific problems. I would like to know what problems did you have in the past, but if builds can be produced, it would be nice to have them, and if problems will appear in the future, I could help with them (if I can :-))

rfratto:
Good idea, I'd change the phrasing just slightly:

> ppc64le builds are currently considered secondary release targets and do not have the same level of support and testing as other platforms.

WDYT?

mattdurham:
@aklyachkin want to add that into the `release-notes.md`?

aklyachkin:
@mattdurham can you check if it suits?

aklyachkin:
now it looks better.

mattdurham (approved):
Great work, thank you!

PR #1539, opened 2022-03-30
[RFC] pkg/config: use -config.file.type flag to derive file type

rfratto:
Introduce a -config.file.type which acts as a hint for the contents of the config file being loaded. This then replaces the -config.dynamic-config-path flag with -config.file.type=dynamic.

Having a dedicated flag for file should be useful for avoiding having too many flags which load a config file and allows us to change the default type in the future (i.e., potentially default to `-config.file.type=flow` in the far future based on the success #1538).

Since the code change is small enough, I'm opening this as an RFC instead of proposing the flag as an issue. If nobody is opposed, we can merge as-is.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -39,6 +39,13 @@ var (
 	}
 )
 
+var (
+	fileTypeYAML    = "yaml"
+	fileTypeDynamic = "dynamic"
+
+	fileTypes = []string{fileTypeYAML, fileTypeDynamic}
+)
+
 // DefaultConfig holds default settings for all the subsystems.
 var DefaultConfig = Config{
 	// All subsystems with a DefaultConfig should be listed here.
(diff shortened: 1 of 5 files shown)


pkg/config/config.go
```diff
@@ -307,34 +314,47 @@ func getenv(name string) string {
...
+		case fileTypeDynamic:
+			if !features.Enabled(fs, featDynamicConfig) {
+				return fmt.Errorf("feature %q must be enabled to use file type %s", featDynamicConfig, fileTypeDynamic)
+			} else if !features.Enabled(fs, featIntegrationsNext) {
+				return fmt.Errorf("feature %q must be enabled to use file type %s", featIntegrationsNext, fileTypeDynamic)
+			} else if features.Enabled(fs, featRemoteConfigs) {
+				return fmt.Errorf("feature %q can not be enabled with file type %s", featRemoteConfigs, fileTypeDynamic)
+			} else if expandArgs {
+				return fmt.Errorf("-config.expand-env can not be used with file type %s", fileTypeDynamic)
+			}
```
rfratto, line 333: Note that this introduces some extra validations that didn't exist before, where unexpected combinations of flags would be silently ignored (cc @mattdurham)

tpaschalis (commented):
You two have had the most experience with config, but for my 2c, this looks like a good idea. ✅

mattdurham (approved):
I think it is reasonable

PR #1545, opened 2022-03-31
Expose `Params` field for integration scrape_config

tpaschalis:

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/integrations/config/config.go
+++ b/pkg/integrations/config/config.go
@@ -3,6 +3,7 @@
 package config
 
 import (
+	"net/url"
 	"time"
 
 	"github.com/prometheus/prometheus/model/relabel"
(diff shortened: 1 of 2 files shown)


v-zhuravlev (approved):

mattdurham (approved):
LGTM

rfratto:
Shouldn't this be added to v2 integrations too? Otherwise the v2 SNMP shim wouldn't work

tpaschalis:
@rfratto I think that v2 uses Prometheus' scrape_config [directly](https://github.com/grafana/agent/blob/vzhuravlev/snmp_exporter/pkg/integrations/v2/snmp_exporter/snmp.go#L69), so in contrast with v1 we already have access to the entire Prometheus' `ScrapeConfig` struct.
https://github.com/grafana/agent/blob/8f1f99d4b2b074756c00188e57378e541b73bd7d/pkg/integrations/v2/autoscrape/autoscrape.go#L63-L66

Also, for v2 we're using the `__param_target` label to set the target query param when building the target group.
https://github.com/grafana/agent/blob/8f1f99d4b2b074756c00188e57378e541b73bd7d/pkg/integrations/v2/snmp_exporter/snmp.go#L62

Do you think it's okay, or am I missing something here?

rfratto:
Ah, you're right, thank you 👍

PR #1546, opened 2022-03-31
Rfc agent flow

mattdurham:
Problem statement covered in https://github.com/grafana/agent/pull/1538

This is for discussion around expression/message based

### DIFF
--- a/docs/rfcs/0004-agent-flow.md
+++ b/docs/rfcs/0004-agent-flow.md
@@ -35,11 +35,13 @@ Agent Flow refers to both the execution, configuration and visual configurator o
 
 The primary goals of Agent Flow are
 
-1. Allow users to easily configure the system
-2. Allow users to understand the system
-3. Allow users to inspect the system
-4. Allow developers to easily add components to the system
-4. Ensure the Agent still maintains high performance on all platforms that the Agent currently supports
+1. Allow users to more easily understand the impact of their configuration
+2. Allow users to collect metrics from all running integrations across a  set of agents without installing extra dependencies
+3. Allow users to run integrations based on a dynamic environment
+4. Eliminate the need for duplication in the configuration file
+5. Allow developers to easily add components to the system
+6. Ensure the Agent still maintains high performance on all platforms that the Agent currently supports
+7. Allow the capability to support a GUI for configuration
 
 Goals of this design document
 
@@ -51,3 +53,109 @@ This document represents ideals and not technical implementation.
 Non Goals of this design document
 
 * Define a technical implementation or configuration implementation
+
+# 2. Broad Solution Path
+
+Conversation around should the components be assembled via message passing, via expressions, or a hybrid approach.
+
+**Note: Consider all examples pseudoconfig**
+
+## 2.1 Expression Based
+
+Expression based is writing expressions that allow referencing other components streams/outputs/values and using them directly. 
+
+**Pros**
+
+* Easier to Implement, evaluating expressions can map directly to existing config structs
+* Components are more reusable, you can pass basic types arounds (string, int, bool)
+
+**Cons**
+* Harder for users to wire things together
+* Harder to build a GUI for
+
+
+## 2.2 Message Based
+
+Message based is where components have no knowledge of other components and information is passed strictly via input and output streams. 
+
+**Pros**
+
+* Easier for users to understand the dependencies between components
+* Easier to build a GUI for
+* Ability to use more configuration formats easier
+
+**Cons**
+
+* More time consuming to implement, existing integrations/items would need to be componentized
+* Larger type system needed
+* More structured to keep the amount of types down
+
+Messages would require a more rigid and well defined type structure. For instance for getting credentials from various sources and passing those credentials around we would want to avoid the following.
+
+* MySQLCredentials
+* RedisCredentials
+* RemoteWriteCredentials
+* MySQLCredentialsSourceS3
+* MySQLCredentialsSourceVault
+* MySQLCredentialsSourceConsul
+* RedisCredentialsSourceS3
+* RedisCredentialsSourceVault
+* RedisCredentialsSourceConsul
+* RemoteWriteCredentialsSourceS3
+* RemoteWriteCredentialsSourceVault
+* RemoteWriteCredentialsSourceConsul
+
+And instead have one component that can read from many sources and outputs a single `Credential Type` and its up the destination component to intepret that correctly. 
+
+## 2.3 Hybrid
+
+## 2.4 Examples
+
+### 2.4.1 Simple Example Mysql from Target Discovery
+
+**Expression**
+
+```
+discovery "mysql_pods" {
+    # some sort of config here to find pods
+}
+
+
+integration "mysql" {
+  # Create one mysql integration for every element in the array here 
+  for_each = discovery.mysql_pods.targets
+
+  # Each spawned mysql integration has its data_source_name derived from 
+  # the address label of the input target.
+  data_source_name = "root@(${each.labels["__address__"]})"
+}
+```
+
+**Message**
+
+```
+discovery "mysqlpods" {
+    relabel_config {
+        [
+            {
+                source = "__address__"
+                match = "*mysql"
+                action = "replace"
+                replacement = "root@$1"
+            }
+        ]
+    }
+}
+
+# I think this would depend on convention, mysql would look at __address__ , and maybe optionally look for username/password
+integration "mysql" {}
+
+connections {
+    [
+        {
+            source = mysqlpods
+            destination = mysql
+        }
+    ]
+}
+```
--- a/docs/rfcs/0004-agent-flow.md
+++ b/docs/rfcs/0004-agent-flow.md
@@ -9,6 +9,7 @@
 
 The Agents configuration and onboarding is difficult to use. Viewing the effect of configuration changes on telemetry data is difficult. Making the configuration simplier, composable and intuitive allieviates these concerns.
 
+
 ## Description
 
 Agent Flow is intended to solve real world needs that the Grafana Agent team have idenfified in conversations with users and developers. 
@@ -24,6 +25,8 @@ These broadly include
     - Remote Write, Different Input Formats
     - Different output formats
     - Filtering ala Relabel Configs is complex and hard to figure out when they occur
+- Lack of understanding how telemetry data moves through agent
+    - Other systems use pipeline/extensions to allow users to understand how data moves through the system
 
 # 1. Introduction and Goals 
 
--- a/docs/rfcs/0004-agent-flow.md
+++ b/docs/rfcs/0004-agent-flow.md
@@ -2,7 +2,7 @@
 
 * Date: 2022-03-30
 * Author: Matt Durham (@mattdurham)
-* PR: [grafana/agent#XXXX](https://github.com/grafana/agent/pull/XXXX)
+* PR: [grafana/agent#1538](https://github.com/grafana/agent/pull/1538)
 * Status: Draft
 
 ## Overarching Problem Statement
--- a/docs/rfcs/0004-agent-flow.md
+++ b/docs/rfcs/0004-agent-flow.md
@@ -1,0 +1,51 @@
+# Agent Flow - Agent Utilizing Components 
+
+* Date: 2022-03-30
+* Author: Matt Durham (@mattdurham)
+* PR: [grafana/agent#XXXX](https://github.com/grafana/agent/pull/XXXX)
+* Status: Draft
+
+## Overarching Problem Statement
+
+The Agents configuration and onboarding is difficult to use. Viewing the effect of configuration changes on telemetry data is difficult. Making the configuration simplier, composable and intuitive allieviates these concerns.
+
+## Description
+
+Agent Flow is intended to solve real world needs that the Grafana Agent team have idenfified in conversations with users and developers. 
+
+These broadly include 
+
+- Current configuration is complex and opaque
+- Lack of introspection within the agent
+    - Questions about what telemetry data are being sent
+    - Are rules applying correctly?
+    - How does filtering work?
+- Users have been requesting additional capabilites and adding new features is hard due to coupling between systems, some examples include
+    - Remote Write, Different Input Formats
+    - Different output formats
+    - Filtering ala Relabel Configs is complex and hard to figure out when they occur
+
+# 1. Introduction and Goals 
+
+This design document describes Agent Flow, which is a system for describing the execution and configuration of telemetry data, along with any secondary configuration required to configure the Grafana Agent. 
+
+Agent Flow refers to both the execution, configuration and visual configurator of data flow.
+
+The primary goals of Agent Flow are
+
+1. Allow users to easily configure the system
+2. Allow users to understand the system
+3. Allow users to inspect the system
+4. Allow developers to easily add components to the system
+
+Goals of this design document
+
+* Define high level concepts and goals
+* Determine whether Agent Flow is a feature that is valuable and useful
+* Define high level concepts of the execution path
+
+This document represents ideals and not technical implementation. 
+
+Non Goals of this design document
+
+* Define a technical implementation or configuration implementation
--- a/docs/rfcs/0004-agent-flow.md
+++ b/docs/rfcs/0004-agent-flow.md
@@ -37,11 +37,10 @@ The primary goals of Agent Flow are
 
 1. Allow users to more easily understand the impact of their configuration
 2. Allow users to collect metrics from all running integrations across a  set of agents without installing extra dependencies
-3. Allow users to run integrations based on a dynamic environment
-4. Eliminate the need for duplication in the configuration file
-5. Allow developers to easily add components to the system
-6. Ensure the Agent still maintains high performance on all platforms that the Agent currently supports
-7. Allow the capability to support a GUI for configuration
+3. Allow users to run components based on a dynamic environment
+4. Allow developers to easily add components to the system
+5. Ensure the Agent still maintains high performance on all platforms that the Agent currently supports
+6. Flow configs must be easily machine-readable and machine-writable to support tooling, such as formatting and GUI
 
 Goals of this design document
 
--- a/docs/rfcs/0004-agent-flow.md
+++ b/docs/rfcs/0004-agent-flow.md
@@ -39,6 +39,7 @@ The primary goals of Agent Flow are
 2. Allow users to understand the system
 3. Allow users to inspect the system
 4. Allow developers to easily add components to the system
+4. Ensure the Agent still maintains high performance on all platforms that the Agent currently supports
 
 Goals of this design document
 
--- a/docs/rfcs/0004-agent-flow.md
+++ b/docs/rfcs/0004-agent-flow.md
@@ -16,7 +16,6 @@ Agent Flow is intended to solve real world needs that the Grafana Agent team hav
 
 These broadly include 
 
-- Current configuration is complex and opaque
 - Lack of introspection within the agent
     - Questions about what telemetry data are being sent
     - Are rules applying correctly?
@@ -44,7 +43,6 @@ The primary goals of Agent Flow are
 Goals of this design document
 
 * Define high level concepts and goals
-* Determine whether Agent Flow is a feature that is valuable and useful
 * Define high level concepts of the execution path
 
 This document represents ideals and not technical implementation. 


docs/rfcs/0004-agent-flow.md
```diff
@@ -0,0 +1,162 @@
...
+
+# 2. Broad Solution Path
+
+Conversation around should the components be assembled via message passing, via expressions, or a hybrid approach.
+
+**Note: Consider all examples pseudoconfig**
+
+## 2.1 Expression Based
+
+Expression based is writing expressions that allow referencing other components streams/outputs/values and using them directly.
```
rfratto, line 66: Should we clarify that expressions also implies functions and arithmetic? something like `length(some_service_discovery.targets)`

docs/rfcs/0004-agent-flow.md
```diff
@@ -0,0 +1,162 @@
...
+**Note: Consider all examples pseudoconfig**
+
+## 2.1 Expression Based
+
+Expression based is writing expressions that allow referencing other components streams/outputs/values and using them directly. 
+
+**Pros**
+
+* Easier to Implement, evaluating expressions can map directly to existing config structs
+* Components are more reusable, you can pass basic types arounds (string, int, bool)
```
rfratto, line 71:
```suggestion
* Components are more reusable, you can pass basic types around (string, int, bool) in addition to custom types
```

docs/rfcs/0004-agent-flow.md
```diff
@@ -0,0 +1,162 @@
...
+
+**Cons**
+* Harder for users to wire things together
+* Harder to build a GUI for
```
rfratto, line 75: nit: this needs the "why"

light suggestion:

> * References to components are more complicated, which may be harder for users to understand how to wire things together
> * Every field of a component is potentially dynamic, making it harder to represent visually

docs/rfcs/0004-agent-flow.md
```diff
@@ -0,0 +1,162 @@
...
+
+## 2.2 Message Based
+
+Message based is where components have no knowledge of other components and information is passed strictly via input and output streams. 
+
+**Pros**
+
+* Easier for users to understand the dependencies between components
+* Easier to build a GUI for
+* Ability to use more configuration formats easier
```
rfratto, line 86: nit: 

```suggestion
* References between components are no more than strings, making the text-based representation language agnostic (e.g., it could be YAML, JSON, or any language)  
```

docs/rfcs/0004-agent-flow.md
```diff
@@ -0,0 +1,162 @@
...
+
+
+## 2.2 Message Based
+
+Message based is where components have no knowledge of other components and information is passed strictly via input and output streams. 
+
+**Pros**
+
+* Easier for users to understand the dependencies between components
+* Easier to build a GUI for
```
rfratto, line 85: nit: need the "why" it's easier :)

docs/rfcs/0004-agent-flow.md
```diff
@@ -0,0 +1,162 @@
...
+* Ability to use more configuration formats easier
+
+**Cons**
+
+* More time consuming to implement, existing integrations/items would need to be componentized
+* Larger type system needed
```
rfratto, line 91:
```suggestion
* Larger type system needed to express custom message streams
```

docs/rfcs/0004-agent-flow.md
```diff
@@ -0,0 +1,162 @@
...
+* MySQLCredentialsSourceVault
+* MySQLCredentialsSourceConsul
+* RedisCredentialsSourceS3
+* RedisCredentialsSourceVault
+* RedisCredentialsSourceConsul
+* RemoteWriteCredentialsSourceS3
+* RemoteWriteCredentialsSourceVault
+* RemoteWriteCredentialsSourceConsul
+
+And instead have one component that can read from many sources and outputs a single `Credential Type` and its up the destination component to intepret that correctly.
```
rfratto, line 109: I'm not sure this is possible, though. What would a single Credential type look like There's many different types of credentials; how would you distinguish between Basic Auth and Bearer Auth?

docs/rfcs/0004-agent-flow.md
```diff
@@ -0,0 +1,162 @@
...
+
+Message based is where components have no knowledge of other components and information is passed strictly via input and output streams. 
+
+**Pros**
+
+* Easier for users to understand the dependencies between components
+* Easier to build a GUI for
+* Ability to use more configuration formats easier
+
+**Cons**
```
rfratto, line 88: I'm not sure if this is a con, or if it still counts as an "expression," but we would need to define custom types for things that can't be represented as a basic (number, bool, string) or complex type (array, dictionary). 

For example, we would require a type which is `log stream` for components which filter logs. 

I'm not sure if that makes it a message-based system (I don't think it does, things are still wired via expressions), but it does mean that it's impossible for us to represent 100% of the agent today in terms of existing basic / complex data types.

docs/rfcs/0004-agent-flow.md
```diff
@@ -0,0 +1,162 @@
...
+
+```
+discovery "mysqlpods" {
+    relabel_config {
+        [
+            {
+                source = "__address__"
+                match = "*mysql"
+                action = "replace"
+                replacement = "root@$1"
```
rfratto, line 145: nit: doesn't a valid DSN require the parenthesis? `root@($1)`?

docs/rfcs/0004-agent-flow.md
```diff
@@ -0,0 +1,162 @@
...
+            }
+        ]
+    }
+}
+
+# I think this would depend on convention, mysql would look at __address__ , and maybe optionally look for username/password
+integration "mysql" {}
```
rfratto, line 152: If there's no input target, would it be possible to still set `data_source_name` manually?

docs/rfcs/0004-agent-flow.md
```diff
@@ -0,0 +1,162 @@
...
+
+connections {
+    [
+        {
+            source = mysqlpods
+            destination = mysql
+        }
+    ]
+}
```
rfratto, line 161: Is it an artifact of the design that connections between components are defined separate to the components?

docs/rfcs/0004-agent-flow.md
rfratto: Oops, this should've been on expressions and not messages.

docs/rfcs/0004-agent-flow.md
```diff
@@ -0,0 +1,162 @@
...
+
+* Easier to Implement, evaluating expressions can map directly to existing config structs
+* Components are more reusable, you can pass basic types arounds (string, int, bool)
+
+**Cons**
+* Harder for users to wire things together
+* Harder to build a GUI for
+
+
+## 2.2 Message Based
```
rfratto, line 78: One thing that just came to mind: it's less obvious to users what fields a connection between two message-based components is overriding; you'd have to pull up the docs to figure out what the connection sets. Comparatively, in expressions, the settings being made dynamic is explicit and can be inferred just by reading the config file.

docs/rfcs/0004-agent-flow.md
mattdurham: Likely an enum or some other flag. May force the user into storing their credentials in a specific format.  

```
{
  "bearer-token" : "1234"
}
```

or
```
{
"basic-auth" 
  {
     "username":"1234"
      "password": "1234"
  }
}
```

docs/rfcs/0004-agent-flow.md
mattdurham: Agreed, I think custom types are likely needed in both cases. Messages need considerably more than expressions.

docs/rfcs/0004-agent-flow.md
mattdurham: Yes, the configuration can still be set internally. So the below would be valid.

```
integration "mysql" {
  datasource = root@(localhost)
  timeout = 10s
}
```

docs/rfcs/0004-agent-flow.md
mattdurham: Ideally, for most components, the user shouldn't need to worry about it. I do think there will be cases where they do and that will not be ideal.

rfratto: The most common case I see is a user configuring a component: they need to make sure they don't configure both the static field and the connection, so they'll need to pull up the documentation to see what values are still required.

I think this will be the case when configuring a new component or editing an existing component to have a connection or to remove a connection and make it static.

rfratto:
It seems to be that if we opt for expressions, we're implicitly saying that configuration the agent will have the same complexity as configuring Terraform, since that's where the expression idea came from. 

It's not quite a 1:1 comparison, though. Many of the blocks in Terraform are standalone, whereas very few (if any) blocks in Agent Flow do anything without being connected to something else. Despite that, I think the comparison might be a good gauge to measure the perceived complexity that using expressions would bring. If the common perception of Terraform is that expressions and using the output of other resource/data blocks are difficult to understand, then we might not want to opt for expressions.

docs/rfcs/0004-agent-flow.md
```diff
@@ -0,0 +1,166 @@
...
+
+This document represents ideals and not technical implementation. 
+
+Non Goals of this design document
+
+* Define a technical implementation or configuration implementation
+
+# 2. Broad Solution Path
+
+Conversation around should the components be assembled via message passing, via expressions, or a hybrid approach.
```
rfratto, line 60:
```suggestion
At a high level, Agent Flow:

* Breaks apart the existing hierarchical configuration file into reusable components 
* Allows components to be connected, resulting in a programmable pipeline of telemetry data

This document considers three potential approaches to allow users to connect components together: 

1. Message passing (i.e., an actor model) 
2. Expressions (i.e., directly referencing the output of another component)
3. A hybrid of both messages and expressions  
```

docs/rfcs/0004-agent-flow.md
```diff
@@ -0,0 +1,166 @@
...
+## Overarching Problem Statement
+
+The Agents configuration and onboarding is difficult to use. Viewing the effect of configuration changes on telemetry data is difficult. Making the configuration simplier, composable and intuitive allieviates these concerns.
+
+
+## Description
+
+Agent Flow is intended to solve real world needs that the Grafana Agent team have idenfified in conversations with users and developers. 
+
+These broadly include
```
rfratto, line 19:
```suggestion
These broadly include:
```

docs/rfcs/0004-agent-flow.md
```diff
@@ -0,0 +1,166 @@
...
+- Users have been requesting additional capabilites and adding new features is hard due to coupling between systems, some examples include
+    - Remote Write, Different Input Formats
+    - Different output formats
+    - Filtering ala Relabel Configs is complex and hard to figure out when they occur
+- Lack of understanding how telemetry data moves through agent
+    - Other systems use pipeline/extensions to allow users to understand how data moves through the system
+
+# 1. Introduction and Goals 
+
+This design document describes Agent Flow, which is a system for describing the execution and configuration of telemetry data, along with any secondary configuration required to configure the Grafana Agent.
```
rfratto, line 34:
```suggestion
This design document outlines Agent Flow, a system for describing a programmable pipeline for telemetry data. 
```

docs/rfcs/0004-agent-flow.md
```diff
@@ -0,0 +1,166 @@
...
+Goals of this design document
+
+* Define high level concepts and goals
+* Define high level concepts of the execution path
+
+This document represents ideals and not technical implementation. 
+
+Non Goals of this design document
+
+* Define a technical implementation or configuration implementation
```
rfratto, line 56:
```suggestion
### Goals 

* Allow users to more easily understand the impact of their configuration
* Allow users to collect integration metrics across a set of agents
* Allow users to run components based on a dynamic environment
* Allow developers to easily add components to the system
* Maintain high performance on all currently-supported platforms
* Produce machine-readable and machine-writable configs for tooling such as formatters or a GUI.

### Non-goals

* Discuss technical details: we instead focus on how a user would interact with a hypothetical implementation of Agent Flow.
```

docs/rfcs/0004-agent-flow.md
```diff
@@ -0,0 +1,166 @@
...
+
+## 2.2 Message Based
+
+Message based is where components have no knowledge of other components and information is passed strictly via input and output streams. 
+
+**Pros**
+
+* Easier for users to understand the dependencies between components
+* Easier to build a GUI for
+    * Inputs and Outputs are well typed
```
rfratto, line 88: Expressions are well-typed too, there's just more basic types like string vs only having custom types.

docs/rfcs/0004-agent-flow.md
```diff
@@ -0,0 +1,166 @@
...
+    * Messages mean users are only dragging outputs to inputs. The components themselves determine how to interpret the input needing less configuration from the user.
```
rfratto, line 89:
```suggestion
    * Connections are made by connecting two components directly, compared to expressions which connect subsets of a component's output
```

docs/rfcs/0004-agent-flow.md
```diff
@@ -0,0 +1,166 @@
...
+* MySQLCredentialsSourceVault
+* MySQLCredentialsSourceConsul
+* RedisCredentialsSourceS3
+* RedisCredentialsSourceVault
+* RedisCredentialsSourceConsul
+* RemoteWriteCredentialsSourceS3
+* RemoteWriteCredentialsSourceVault
+* RemoteWriteCredentialsSourceConsul
+
+And instead have one component that can read from many sources and outputs a single `Credential Type` and its up the destination component to intepret that correctly.
```
rfratto, line 113:
```suggestion
Messages require a more rigid type structure to minimize the number of total components. 

For example, it would be preferable to have a single `Credential` type that can be emitted by an s3, Vault, or Consul component. These components would then need to set a field that marks their output as a specific kind of Credential (such as Basic Auth or Bearer Auth).

If, instead, you had multiple Credential types, like `MySQLCredentials` and `RedisCredentials`, you would have the following components:

* Vault component for MySQL credentials 
* Vault component for Redis credentials 
* S3 component for MySQL credentials 
* S3 component for Redis credentials 
* (and so on)
```

rfratto (commented):
Some editing nits to maybe hopefully make it more clear to new readers what we're trying to build:

docs/rfcs/0004-agent-flow.md
mattdurham: Will play with the verbiage there

rfratto:
Another point we may want to consider regarding messages/expressions: 
with the message pattern, existing components need to be updated to support new types of components (for example, extending Credentials to support OAuth would require updating components that accept Credentials to be aware of OAuth support). 

A similar situation may still be possible with expressions but it's far less likely as component outputs would more heavily have basic data types like strings that can be referenced.

rfratto:
Given my recent thoughts, I'm currently in favor of using expressions as the _basis_ for the raw representation of the flow config file. While the messages/actor-based pattern is inarguably simpler to reason about the flow of data, I'm primarily concerned about: 

* The implementation effort needed to convert the Agent codebase into supporting actor-based messages across the board
* The design overhead needed to ensure components are reusable enough to avoid creating too many unique types 
* The maintenance cost of extending an existing type, requiring updating a subset of components that consume it
* The inability to understand at a glance what fields get set by forming a connection

However: Anything defined by the messages pattern can be represented in terms of expressions, but not the other way around; i.e., messages can be considered a subset of expressions. This means that tooling can be layered on top of expressions to provide a simpler model (such as a GUI), while still leaving the raw config file open for more advanced expression-based use cases.

So far, I think using expressions as the basis is the best of both worlds: the config file has the most flexibility, it's easier to code, and still allows for the simpler message/actor pattern to exist on top.

mattdurham:
One thought is that streams/expressions will likely need to be immutable. How/does that impact solutions and overall proposal? IE if an array of logs goes to two separate nodes, those two arrays (and contents) will need to be entirely distinct from one another.

rfratto:
> One thought is that streams/expressions will likely need to be immutable. How/does that impact solutions and overall proposal? IE if an array of logs goes to two separate nodes, those two arrays (and contents) will need to be entirely distinct from one another. 

Right, expressions build a declarative system: the input values are immutable each time a component gets evaluated (eg it can get reevaluated later with different a different input), but the values can still encapsulate something that changes internally, like Go channels. This means whatever Go values we'd consider for binding components together using messages can still be used with expressions.

rfratto:
> One thought is that streams/expressions will likely need to be immutable. How/does that impact solutions and overall proposal? IE if an array of logs goes to two separate nodes, those two arrays (and contents) will need to be entirely distinct from one another. 

We'll have to be careful when implementing components so that they don't mutate their inputs (which would also be a race condition if components run in parallel), but I don't think it's a significant problem. We'll need to be clear in the RFC and in documentation that it's a requirement if components to do copy-on-write.

rfratto:
I was considering the initial list of components we would have and how it maps to the codebase today. We may have some complications around metric storage and remote_write. 

Today, we clean up the WAL using the lowest sent timestamp across all remote_writes which are reading from it. I see three ways this impacts the design of components:

* If remote_write and WAL are separate components, they need some form of bidirectional communication so remote_writes can report back what timestamp can be used for deletion 
* Alternatively, remote_write + WAL can be one component
  * However, this would break the messages pattern as you would no longer have a way to send different credentials to different remote_writes that need them 
* As a final option, we could attempt to find a different way of determining how to clean up data from the WAL that wouldn't break remote_writes  
  * This would allow remote_write and WAL to be separate components without needing any kind of bidirectional communication.

mattdurham:
While I was playing with this over the weekend, I think this is an opportunity to bring the configuration in house. I think when it comes to certain fields like credentials/?others ?we should support loading from an external datastore inherent to the component. And not use expression/messages for things that aren't incredibly dynamic.

IE in the the remote_write configblock. 
```
remote_write:
    url: https://example.com
    remote_write_credential_file: "s3://mybucket/credentials/rw_credentials.yaml"
```

and also support

```
remote_write: 
    config_file: "s3://mybucket/configs/rw.yaml"
```

rfratto:
I think we're moving away from the core concept of reusable components and are shifting back towards predetermined configuration patterns.

mattdurham:
I am not saying we should limit credentials being passed around, but I think most people would use a path based parser located within the node config to get credentials instead of using a credentials provider component for either expressions/messages. My assumption is in most cases a user is not mutating credentials.

IE if I have 10 nodes that need credentials I am unlikely to want to create 10 credential components to stream/grab via expressions the needed credentials. With expressions I could have reference the path to file to a particular key from a single provider, but that seems like the full path with extra steps.  

```
credentials "cred" {
  base_path = s3://bucket/credentials
}

remote_write "rw" {
   url: https://example.com
     username: 12345
     password: cred.file("/rw.txt")
}

```

vs

```
remote_write "rw" {
   url: https://example.com
     username: 12345
     password: s3://bucket/credentials/rw.txt
}
```

rfratto:
I was imagining a expressions-based system would have external-system specific components which output strings for things they were loading:

```tf
s3 "grafanacloud-api-key" {
  key     = "bucket/credentials/rw.txt"
  refresh = "10s" # Frequency to refresh value

  # Output: 
  # data = "secretpassword" 
}

remote_write "example" {
  url      = "http://example.com" 
  username = "12345" 
  password = s3.grafanacloud-api-key.data 
}

# not shown: 
# re-use s3.grafanacloud-api-key.data for log & trace writing
```

This would have the following benefits:

1. **Consistency**: Using components for everything gives a nice consistency to the design: no special strings to learn, just discovering a component that might do what you want. 
2. **Flexibility**: Having a component that loads something from S3 and outputs a string allows it to be used in other scenarios. It could be passed to an `include` attribute that loads in extra config files, allowing Agent Flow to be dynamically reconfigured from a remote source. 
3. **Secret re-use**: The output of a component such as this may be re-used multiple times. This is more of a performance optimization: you wouldn't need to download the same S3 file N times, and we wouldn't have to build a cache to reduce overhead of the string-based form. 
4. **More obvious distinct authentication support**: While I was initially in favor of the URL idea when we initially explored #1121, the complexities around authentication made me move away from it. How do you format the URL if you require multiple distinct sets of credentials to authenticating against HTTP and S3 providers? I originally considered using query parameters, but you have to consider what to do if the HTTP server you're authenticating against expects query parameters as part of the request. It got messy fast.
    1. Alternatively, the strings could be a special format, like `key=value` pairs, but that feels like a lot of workarounds compared to what components could handle natively. 
5. **Direct introspection**: As being a distinct component, their health becomes independently introspectable: you can see if just the `s3.grafanacloud-api-key` component is unhealthy and look for errors, compared to finding out that the URL is unhealthy after investigating why the `remote_write.example` component isn't working.   
6. **Implementation effort**: I agree we should stop depending on downstream configuration and bring things in-house, but I still believe mapping to a string would take less work. When I'd like to support things like refresh, I'm concerned that it pushes just a small amount of extra complexities on components that, in an expression-based world, they wouldn't have to be aware of. 

> My assumption is in most cases a user is not mutating credentials.

I think this is true most of the time, but rolling API keys does happen (some people even roll them automatically on a fixed period).

rfratto:
To devil's advocate my own argument, if the biggest selling point for separate components for credentials is component flexibility, it might not be all that valuable if we can't identify that many use cases. So far I can only imagine it being used for loading secrets and loading extra flow config files to include, but other uses cases aren't immediately obvious to me right now.

mattdurham:
In the above example expressions are easier for credentials vs messages for sure. In your above example are we assuming `key=value` for the txt file? (Or working off the file extension?) 

I like the idea of a credentials object, but that could very easily be my love for strict types systems. I can def understand the below being to rigid but I like the cleanness of it. 

```
basic_auth:
- user: 12345
   password: example!
mysql:
- user: root
  password: password
github:
- token: 12345qwers
```

rfratto:
> In the above example expressions are easier for credentials vs messages for sure. In your above example are we assuming key=value for the txt file? (Or working off the file extension?)

Oh, sorry, I just copied the same object name you were using. The component just returns a string; it's not interpreted by the component itself. However, components referencing the string can choose to parse the strings: 

```hcl 
s3 "arbitrary-data" {
  key     = "bucket/some/data"
}

# Extra Flow files to include.
include = [
  # Function to return file contents as string. Parsed 
  # as a Flow file. 
  file("/some/file/agent-extra-settings.flow"),

  # Function to return environment variable contents as 
  # string. Parse as a Flow file. (Storing file contents 
  # in an environment variable would be a bad idea in 
  # practice) 
  env("MY_CONFIG_FILE"),

  # Resolves to string contents of bucket/some/data. Parsed 
  # as a Flow file.  
  s3.arbitrary-data.data,   

  # Literal string. Parsed as a Flow file.  
  <<-EOF 
    remote_write "remote-write2" {
      url = "http://example2.com"
    }
  EOF
]

remote_write "example" {
  url      = "http://example.com" 
  username = "12345" 

  # Resolves to string contents of bucket/some/data. Used 
  # literally as password. 
  password = s3.arbitrary-data.data
}
```

> I like the idea of a credentials object, but that could very easily be my love for strict types systems. 

Can you explain what you mean by strict type systems? It's been mentioned a few times but we might have different definitions; I'm saying a component literally returns a `string` and can only be passed to things expects `string`s, which is strict typing.

rfratto:
It might be more clear to demonstrate what I'm talking about in terms of the literal Go representation:

```go 
package main

import (
	"fmt"
	"os"
)

type RemoteWriteConfig struct {
	URL string

	// Optional basic auth username + password
	Username, Password string
}

type FlowConfig struct {
	// Contents of files to include as their string contents
        // (i.e., file *contents*, not literally file names)
	Include []string

	RemoteWrite []RemoteWriteConfig
}

type S3Config struct {
	// Key to load
	Key string
}

type S3Result struct {
	// Data from Key.
	Data string
}

func resolveS3(cfg S3Config) S3Result {
	return S3Result{
		Data: "contents-from-s3",
	}
}

func main() {
	arbitraryData := resolveS3(S3Config{
		Key: "bucket/some/data",
	})

	flowFile, _ := os.ReadFile("/some/file/agent-extra-settings.flow")

	cfg := FlowConfig{
		Include: []string{
			string(flowFile),
			os.Getenv("MY_CONFIG_FILE"),
			arbitraryData.Data,
			`
			  remote_write "remote-write2" {
			    url = "http://example2.com"
			  }
			`,
		},

		RemoteWrite: []RemoteWriteConfig{{
			URL:      "http://example.com",
			Username: "12345",
			Password: arbitraryData.Data,
		}},
	}

	fmt.Printf("%#v\n", cfg)
}

```

tpaschalis:
Leaving some random thoughts here :) I'm still trying to wrap my head around on how this would look in practice, and how it would impact the Agent's future.

First off, I feel we're talking about two things at once here; how we (internally) handle the representation using expression-based components or message-based actors, versus what are the ways we allow users to define and configure those components (externally). Is a decision here mainly based on what we expose as configuration (with the tradeoffs of implementing them), or are these options tightly coupled and a decision would have a greater impact on how it internally works?

Secondly, I'm having a harder time grasping the message-based approach. The example config, defines some 'messages', like the arrows in the diagram? That's where the Actor paradigm comes in, each being able to 'react' to different kind of messages, right? Messages will _both_ be configuration-specific things like credentials, scrape targets, transformers but also actual _data_ like Prometheus metrics, log lines and traces?

Having spent only a couple of hours thinking about it, I haven't formed any strong opinions yet, but the expression-based case _does_ sound simpler to support and extend, and if it can also be compatible with message-sending, that would also be a point in its favor.

Finally, (hoping this isn't way off topic), but have we looked at [OTLP's pipelines](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/design.md#pipelines)? They're more restrictive, start from the receiver and only allow for a more linear model of combining, but it would be interesting to see how their pieces communicate with each other, and there are any learned lessons that we can reuse.

mattdurham:
So far we haven't defined a configuration format, only what the format needs to support. I would consider all our examples "pseudo-config". I think there are two things to discuss at this point: 

1. Are using components worthwhile? 
2. How do those components communicate?

You can see an example of message-based agent flow at https://github.com/mattdurham/agentflow, hoping to do something similar with expressions. 

Pipelines are something we looked at, and provided some validation for the component concept.

mattdurham:
Regarding strict typing very poor verbiage on my part. I am thinking complex types to represent credentials if we went to a message based system.  IE in the s3 file `s3://bucket/credentials.yaml` would be this text and it would parse it as a credentials object.

```
basic_auth:
- user: 12345
   password: example!
mysql:
- user: root
  password: password
github:
- token: 12345qwers
```

rfratto:
It's not fully wired together, but here is the expressions based example that I [tweeted](https://twitter.com/robertfratto/status/1501220651083960324) about a month ago: https://github.com/rfratto/graph-agent

hedss:
Apologies for 'shoving my oar in' here, as well as being a bit late. The FE team has been looking at this with interest, especially from the PoV of prospects/customers who have interesting use cases (as well as our own).
My own preference here is expression based. Whilst not initially the 'simplest' way to create these connections, the 'modular' approach to pull in blocks you need and to easily reference these (including intrinsic types), gives a huge amount of flexibility to creating sources/sinks and then reference these. 
And this includes the ability, potentially, for users to create their own blocks (I know, I know, way down the road), to carry out operations that allow them to customise their experience (as well as carry out operations such as canarying, etc.).
There's also the fact that a lot of teams that would be configuring the agent out in the wild most likely have some HCL experience, and for them, this would not be a slightly different paradigm to pick up to configure the agent.

rfratto:
@mattdurham I think we're past the point of a new RFC for Flow, but since this started the discussions, I think it should be merged as-is to provide historical context for people.

PR #1550, opened 2022-03-31
Makefile: ensure that Docker containers can be canceled by the user

rfratto:
It was previously impossible to abort a `make` target which ran a docker container. Ensuring that the `--init -it` flags are passed to docker run guarantee that the user can always Ctrl+C to cancel the target and the container.

This commit also adds the `--rm` flag to commands that were missing it.


mattdurham (approved):
Yeeeeeesssssss!

PR #1555, opened 2022-04-04
Improve documentation on how to run Smoke Tests

marctc:
#### PR Description

Following the steps of how to run smoke tests led to have pods crash looping because
images where not found. This PR clarifies the extra step where the images need to
be imported in order to run the smoke test.

### DIFF
--- a/example/k3d/README.md
+++ b/example/k3d/README.md
@@ -11,7 +11,7 @@ that implements a full Grafana Agent environment for testing.
 - Kubectl
 - Docker
 - [Tanka >= v0.9.2](https://github.com/grafana/tanka)
-- [k3d >= v4.0.0](https://github.com/rancher/k3d)
+- [k3d >= v4.0.0,<= v5.2.2](https://github.com/rancher/k3d)
 - [jsonnet-bundler >= v0.4.0](https://github.com/jsonnet-bundler/jsonnet-bundler)
 
 ### Getting Started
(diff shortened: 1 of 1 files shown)


example/k3d/README.md
```diff
@@ -11,7 +11,7 @@ that implements a full Grafana Agent environment for testing.
 - Kubectl
 - Docker
 - [Tanka >= v0.9.2](https://github.com/grafana/tanka)
-- [k3d >= v4.0.0](https://github.com/rancher/k3d)
+- [k3d >= v4.0.0,<= v5.2.2](https://github.com/rancher/k3d)
```
marctc, line 14: Installing the last version (v5.4.1) causes that running the agent in a k3d cluster doesn't work. I tried with the version that was working for @tpaschalis (v5.2.2) and it worked. Not sure if we have to fix those issues instead of pining this version here.

tpaschalis: For more context, the issue _seems_ to be with the version of Kubernetes client used by k3d.

The error message when trying to interact with the k3d cluster using `tk apply` is the following. Instead of providing the interactive diff, prompting the user to type 'yes' and apply the changes, we get:
```
$ tk apply ./environment
error: rbac.authorization.k8s.io/v1beta1, Kind=ClusterRole doesn't support dry-run
Error diffing: exit status 2
```

This error message is emitted by Kubernetes' DryRunVerifier, but I couldn't figure out which commit prompted the change and when it was bundled as a dependency in k3d.

PR #1565, opened 2022-04-05
RFC: deprecate agent-operator

captncraig:
This is a proposal to deprecate the agent-operator. Details as to reasons are in the included doc. 

This is a request for comments, NOT necessarily a final decision. We have noticed a high maintenance burden to the operator, and wonder if it is worth continuing, or if we should pursue other avenues to make kubernetes deployment easier.

### DIFF
--- a/docs/rfcs/0004-deprecate-operator.md
+++ b/docs/rfcs/0004-deprecate-operator.md
@@ -33,7 +33,7 @@ It allows users with existing Prometheus Operator infrastructure to reuse their 
 
 ### Alternatives
 
-The above lead me to wonder if the operator is worth the effort of maintaining as a seperate app. If our goal is to make deployment of the agent to kubernetes as easy as possible, perhaps we would be better served by providing a happy path to deploying customized manifests directly. 
+The above led me to wonder if the operator is worth the effort of maintaining as a separate app. If our goal is to make deployment of the agent to Kubernetes as easy as possible, perhaps we would be better served by providing a happy path to deploying customized manifests directly. 
 
 The first large step we should take is publishing an official Helm chart. It should be usable to deploy the agent with a working config extremely easily, while still allowing fully custom agent config.
 


docs/rfcs/0004-deprecate-operator.md
```diff
@@ -0,0 +1,40 @@
...
+It does so by introducing custom resources to represent grafana agent instances, metrics and logging backends, as well as
+scrape configs. One problem that comes up consistently is that these CRDs are yet another way to represent the grafana agent config. The operator
+requires extra maintenance to make new agent features availible to users, and it can get out of sync easily. Without being well understood,
+the operator model can lead to confusion for new users, and make debugging the agent more difficult.
+
+I propose we deprecate the agent-operator, and instead focus on making agent deployments as seamless as possible for users to configure and deploy it directly.
+
+### Purpose of the agent
+
+The agent-operator is designed to be the easiest way to deploy Grafana Agent to kubernetes. It watches two main sets of CRDs:
```
rfratto, line 20:
```suggestion
The operator is designed to be the easiest way to deploy Grafana Agent to Kubernetes. It watches two main sets of CRDs:
```

docs/rfcs/0004-deprecate-operator.md
```diff
@@ -0,0 +1,40 @@
...
+
+- Prometheus operator crds for podmonitors, servicemoniters, etc.
```
rfratto, line 22:
```suggestion
- Prometheus Operator CRDs for PodMonitors, ServiceMonitors, and Probes.
```

docs/rfcs/0004-deprecate-operator.md
```diff
@@ -0,0 +1,40 @@
...
+- Agent specific operators for declaring agent instances, metrics endpoints and so forth.
```
rfratto, line 23:
```suggestion
- Grafana Agent-specific CRDs for declaring agent instances, metrics endpoints and so forth.
```

docs/rfcs/0004-deprecate-operator.md
```diff
@@ -0,0 +1,40 @@
...
+
+It allows users with existing prometheus-operator infrastructure to reuse those monitor resources without needing to redefine scrape configs for the agent.
+
+### Problems with the agent
+
+- The agent specific CRDs are similar to, but distinct from the internal agent config types.
+- Operator config is not always up to date with agent config. Major agent features (tracing for example), take additional work to port into the operator crds.
+- Kubernetes operators are less understood by the larger agent squad. Maintenance on the operator is slower than on the agent itself.
+- Documentation must take into account that there are two fully supported ways to deploy and configure the agent. Operator crds are a completely different format than the agent config yaml.
```
rfratto, line 32:
```suggestion
- Documentation must take into account that there are two fully supported ways to deploy and configure the agent. Operator CRDs are a completely different format than the agent config yaml.
```

docs/rfcs/0004-deprecate-operator.md
```diff
@@ -0,0 +1,40 @@
...
+
+### Alternatives
+
+The above lead me to wonder if the operator is worth the effort of maintaining as a seperate app. If our goal is to make deployment of the agent to kubernetes as easy as possible, perhaps we would be better served by providing a happy path to deploying customized manifests directly.
```
rfratto, line 36:
```suggestion
The above led me to wonder if the operator is worth the effort of maintaining as a separate app. If our goal is to make deployment of the agent to Kubernetes as easy as possible, perhaps we would be better served by providing a happy path to deploying customized manifests directly. 
```

docs/rfcs/0004-deprecate-operator.md
```diff
@@ -0,0 +1,40 @@
...
+- The agent specific CRDs are similar to, but distinct from the internal agent config types.
+- Operator config is not always up to date with agent config. Major agent features (tracing for example), take additional work to port into the operator crds.
+- Kubernetes operators are less understood by the larger agent squad. Maintenance on the operator is slower than on the agent itself.
+- Documentation must take into account that there are two fully supported ways to deploy and configure the agent. Operator crds are a completely different format than the agent config yaml.
+
+### Alternatives
+
+The above lead me to wonder if the operator is worth the effort of maintaining as a seperate app. If our goal is to make deployment of the agent to kubernetes as easy as possible, perhaps we would be better served by providing a happy path to deploying customized manifests directly. 
+
+The first large step we should take is publishing an official helm chart. It should be usable to deploy the agent with a working config extremely easily, while still allowing fully custom agent config.
```
rfratto, line 38:
```suggestion
The first large step we should take is publishing an official Helm chart. It should be usable to deploy the agent with a working config extremely easily, while still allowing fully custom agent config.
```

docs/rfcs/0004-deprecate-operator.md
```diff
@@ -0,0 +1,40 @@
...
+
+The biggest missing features in deprecating the operator would probably be native support for the prometheus operator Monitor crds. We should explore the possibility of having the agent itself watch those types and merge into its own config, or some other method for supporting those (if they are important to our user base).
```
rfratto, line 40:
```suggestion
The biggest missing features in deprecating the operator would probably be native support for the Prometheus Operator's CRDs. We should explore the possibility of having the agent itself watch those types and merge into its own config, or some other method for supporting those (if they are important to our user base).
```

rfratto (commented):
Gave a quick first pass with nits for spelling/capitalization

IMO we can refer to "the agent" or "the operator" without capitalization, but "Grafana Agent" and "Grafana Agent Operator" should be capitalized.

rfratto:
All said, I'm in favor of deprecating the operator, provided any functionality users rely on the operator for are moved to the agent itself. We'd only need to port PodLogs, ServiceMonitors, Probes, and PodMonitors though; other CRDs don't have a lot of value and would be confusing at the agent level (i.e., MetricsInstance, Integration).

I'm nervous about whether we're ready for a good Helm chart yet. We need to think about the experience we want the Helm chart to provide and whether there's any agent-level features we need to add first. In particular, I'm worried about the deployment model: having to manage multiple sets of agents is usually a pain point [mostly] abstracted away by the operator, and pushing it back to the user with a Helm chart would be un-fun. Or, if the Helm chart includes that abstraction today, what's the migration path when we make the experience better such that the abstraction is no longer needed?

rfratto:
Can this be renamed to RFC-0005 so it doesn't collide with https://github.com/grafana/agent/pull/1546?

flokli:
From the RFC itself:

>  The biggest missing features in deprecating the operator would probably be native support for the prometheus operator Monitor crds. We should explore the possibility of having the agent itself watch those types and merge into its own config, or some other method for supporting those (if they are important to our user base). 

From @rfratto:
> All said, I'm in favor of deprecating the operator, provided any functionality users rely on the operator for are moved to the agent itself. We'd only need to port PodLogs, ServiceMonitors, Probes, and PodMonitors though; other CRDs don't have a lot of value and would be confusing at the agent level (i.e., MetricsInstance, Integration).

I very much like the "rendering of the *Monitors/`Probes` resources into operator config. It became the de-facto standard to mark metrics for scraping, and composes nicely - services don't need to know if we deployed grafana-agent-operator, prometheus-operator or something else. I definitely wouldn't want to go back to manually writing scraping config by myself, so if the operator gets deprecated, the agent itself *really* would *need* to learn how to interpret monitoring CRDs. This should be an integral part of the RFC, not just a "we should explore the possibility" promise.

If it gets support to interpret the CRDs, we probably should also talk about how to show the resulting scraping config for debugging purporses. Being able to peek at the generated scraping config in the current setup for debugging purposes was very helpful. That could probably just be some http api.

I'm not so sure about the log collection bits. The complicated work IMHO was getting the mounts properly described, and I really liked the operator taking this away from me. If it's now on the user to write `DaemonSet` resources for this part of grafana-agent directly, at least there'd need to be a lot of "handholding", aka a Helm chart (with a bunch of overrides).

Comment: I didn't need to use the sharding feature so far, so I'm not sure if there would be any problems by suddenly not having an operator in the cluster knowing about the "bigger picture". Has this been evaluated?

Whyeasy:
> I'm nervous about whether we're ready for a good Helm chart yet. We need to think about the experience we want the Helm chart to provide and whether there's any agent-level features we need to add first. In particular, I'm worried about the deployment model

Don't know if I can(or allowed to) chip in. You could take the same route as the Loki helm charts. Where you have 2 different helm charts. One with the all-in-one solution which is a daemon set. Or the "distributed" version which deploys 3 agents with each of their own responsibility. 

And it might be even possible to have this in one chart and a variable in values in the like of `all-in-one.enabled: true`

tesharp:
Just chipping in as well since I can :)

Using both the grafana agent and vector.dev now for different purposes and what I wished was possible was to convert any of the ServiceMonitor etc CRD's to simple configmaps / secrets using some custom, or predefined, transformation. I dont really need / want the MetricsInstance, GrafanaAgent CRD's as I can deploy those myself, and have to for tracing atm. If there was an controller that could read the monitoring CRD's and transform them into prometheus config, or vector config, and write to a configmap I could just use that config map and handle the reloading myself.

That way the agent itself also doesn't need to know about those CRD's which in my opinion would be better as the agent is not only for Kubernetes. Keep it simple and dont add too much to the agent it doesn't need to know about.

NissesSenap:
In general i enjoy operators, I think they abstract away config that me as a user shouldn't have to think about.

For me that have multiple loki instances that i talk to for different reasons (one for platform and one for application) I think it's very useful to not have to manage the helm chart and all the different config that this would force me to do.

I also agree with @tesharp point that, if you would move this logic from the operator in to the agent the agent would have to become more advanced to support both kubernetes and none kubernetes workloads.
The value of separating the deployment logic i think might be worth maintaining the operator just because of that.

Also grafana have recently started supporting the loki operator so gaining more knowledge about managing operators for the grafana company is probably a good thing. But i don't work at grafana but that's my point of view from the outside. 

In one of the operator issues there have been talk about having a single instance that supports tracing, metrics and log gathering. My guess that solving this though a helm chart would make the helm chart very complex. 

Another point to make for operators is the testability of them vs a helm chart.

These are my 2 cents, I'm not a maintainer of this operator but I'm a happy user. And i hope i will be able to be that in the future as well. 

No matter what happens with the agent operator, thanks for creating great open-source projects :).

rfratto:
@captncraig is this ready for another round of reviews? Are you still backing the idea that we deprecate the operator in favor of Flow components?

andreabrduque:
Hi all, Id love to chip in. Our use case in my company is to monitor machine learning workloads running in GPU node pools. Im a K8s enthusiast, but very new to the technology. I started using the Grafana Agent Operator because it was the easiest way to install a unified system to get metrics + logs from Kubernetes. I got a bit confused at first because of the difference between the configuration format from the operator vs the agent, but in general was an OK effort compared to the  benefits that it brought.  At the end, at least for us, it all went down to the fact that we could have one system to have metrics + logs in one go, and that it would automatically detect all the endpoints with the metrics into a namespace.

rfratto:
From a maintainers point of view, I definitely understand all of the concerns about deprecating the operator too early.

I want to make sure we communicate this very well: we won't mark the operator as deprecated until its replacement is widely available and known to cover the use cases the operator enables. We'll provide the replacement openly for feedback before the operator gets deprecated, and we'll revisit the decision to keep the operator if the replacement isn't received well. 
 
The intention is to try adding support for Prometheus Operator CRDs (and PodLogs) directly into the agent itself. From what I understand, nearly all use cases for the operator revolve around just wanting to read these CRDs, rather than wanting something that deploys agent resources for you.  When it comes to deploying agent resources, I think a Helm chart can carry most of the load here.

Personally, I think operators can be really difficult to do well: operators tend to end up exposing a subset over some underlying software (in this case Grafana Agent), and are constantly at risk of the chosen subset being inappropriate for some users. This risk can turn into a slow crawl towards feature bloat as more and more things are injected into CRDs to make sure everything users need is exposed. 

I think Helm charts suffer from the same issue (I see many Helm charts with a silly number of options in values.yaml), which make them hard to maintain. However, Helm charts that aren't exposing options still have an escape hatch: users can always forgo the Helm chart and deploy resources manually, setting options they need. The ability to escape-hatch away from Helm charts isn't something you can do with an operator, where the only option is to wait and deploy a new version of the operator with the features you were waiting for. 

I hope people find this context useful; this is how I've been thinking of the reasoning behind exploring other routes. But to reiterate the earlier point: the operator will only be deprecated once a replacement which covers the operator's use cases is available and the replacement is deemed to be sufficient.

Until this happens, the operator will continue to receive bug fixes, but may not see any new features until we know what its future looks like.

captncraig:
It is clear from interacting with people using the operator that it has a valuable use case, and should stick around for now. If / when more of the features are native in the agent itself, we may bring this up again. But for now I am going to close this to avoid confusion. The operator is an official part of the grafana agent repo for the enforceable future.

PR #1566, opened 2022-04-05
Makefile: support release candidate versions

rfratto:
$IMAGE_BRANCH_TAG must not be an empty string for building images. This defaults $IMAGE_BRANCH_TAG for release candidates to be the same as $RELEASE_TAG.

This will redundantly cause the image to be tagged twice as $RELEASE_TAG, but will avoid build errors.


rlankfo (approved):
LGTM

mattdurham (approved):
LGTM

PR #1575, opened 2022-04-07
Initial windows support for certificate store

mattdurham:
#### PR Description

This introduces support for configuring TLS to use the Windows certificate store. This overrides the existing tls configuration with a new subset that allows you to specify the Client CA root and the server certificate to be used. 

There are unit tests that cover the filtering, unfortunately, due to the windows-only nature and having to install a certificate true testing is a manual process. The certificates necessary are included in `/testdata/windows`, you can manually install the certificates and test. 

This is BETA and subject to change. 

- [X] CHANGELOG updated
- [X] Documentation added
- [X] Tests updated


pkg/server/tls_windows.go
```diff
@@ -0,0 +1,342 @@
...
+			defer cn.winMut.Unlock()
+			cert := &tls.Certificate{
+				Certificate: [][]byte{cn.serverCert.Raw},
+				PrivateKey:  cn.serverSigner,
+				// These seem to the be safest to use, tested on Win10, Server 2016, 2019, 2022
+				SupportedSignatureAlgorithms: []tls.SignatureScheme{
+					tls.PKCS1WithSHA256,
+					tls.PKCS1WithSHA384,
+					tls.PKCS1WithSHA512,
+				},
```
rfratto, line 65: I'm not sure I understand why we need to provide this list yet. It restricts which algorithms are used, so if it's not provided, everything is valid, right?

pkg/server/tls_windows.go
```diff
@@ -0,0 +1,342 @@
...
+		return err
+	}
+	certPool := x509.NewCertPool()
+	certPool.AddCert(cn.clientRootCA)
+	config := &tls.Config{
+		VerifyPeerCertificate: cn.VerifyPeer,
+		GetCertificate: func(info *tls.ClientHelloInfo) (*tls.Certificate, error) {
+			cn.winMut.Lock()
+			defer cn.winMut.Unlock()
+			cert := &tls.Certificate{
```
rfratto, line 57: Do we want to set `Leaf` here?

```go
// Leaf is the parsed form of the leaf certificate, which may be initialized
// using x509.ParseCertificate to reduce per-handshake processing. If nil,
// the leaf certificate will be parsed as needed.
Leaf *x509.Certificate
```

It might make handshakes a bit faster.

pkg/server/tls_windows.go
```diff
@@ -0,0 +1,342 @@
...
+		config.ClientAuth = tls.RequireAnyClientCert
+	case "VerifyClientCertIfGiven":
+		config.ClientAuth = tls.VerifyClientCertIfGiven
+	case "RequireAndVerifyClientCert":
+		config.ClientAuth = tls.RequireAndVerifyClientCert
+	case "", "NoClientCert":
+		config.ClientAuth = tls.NoClientCert
+	default:
+		return fmt.Errorf("invalid ClientAuth %q", c.ClientAuth)
+	}
```
rfratto, line 86: Can we move this to a shared location so we don't have to duplicate it for both TLS mechanisms?

pkg/server/tls_windows.go
```diff
@@ -0,0 +1,342 @@
...
+		subjectRegEx, err = regexp.Compile(c.WindowsCertificateFilter.ClientSubjectRegEx)
+		if err != nil {
+			return fmt.Errorf("error compiling subject common name regular expression %w", err)
+		}
+	}
+
+	// If there is an existing windows certhandler notify it to stop refreshing
+	if l.windowsCertHandler != nil {
+		l.windowsCertHandler.stopUpdateTimer <- struct{}{}
+	}
```
rfratto, line 39: This will block forever if the handler goroutine exits or panics. I'd recommend passing a context to the cert handler and storing the cancel function, calling it if it's not nil:

```
if l.cancelCertHandler != nil {
   l.cancelCertHandler() 
   l.cancelCertHandler = nil 
}
```

pkg/server/tls_windows.go
```diff
@@ -0,0 +1,342 @@
...
+	serverIdentity certstore.Identity
+	clientRootCA   *x509.Certificate
+
+	stopUpdateTimer chan struct{}
+}
+
+func (c *winCertStoreHandler) startUpdateTimer() {
+	if c.cfg.ServerRefreshInterval == 0 {
+		c.cfg.ServerRefreshInterval = 5 * time.Minute
+	}
```
rfratto, line 115: `c.cfg` is protected by a mutex, we need to grab a write lock here or move defaulting logic elsewhere

pkg/server/tls_windows.go
```diff
@@ -0,0 +1,342 @@
...
+		c.serverCert = nil
+		c.serverSigner = nil
+		return
+	case <-time.After(c.cfg.ServerRefreshInterval):
+		err := c.refreshCerts()
+		if err != nil {
+			return
+		}
+
+	}
```
rfratto, line 131: This will only refresh once. Did you mean to wrap this in a for loop?

pkg/server/tls_windows.go
```diff
@@ -0,0 +1,342 @@
...
+
+	select {
+	case <-c.stopUpdateTimer:
+		if c.serverIdentity != nil {
+			c.serverIdentity.Close()
+		}
+		c.serverCert = nil
+		c.serverSigner = nil
+		return
+	case <-time.After(c.cfg.ServerRefreshInterval):
```
rfratto, line 125: You should store the refresh interval in a variable which is initialized with a read lock on c.cfg

pkg/server/tls_windows.go
```diff
@@ -0,0 +1,342 @@
...
+		err := c.refreshCerts()
+		if err != nil {
+			return
+		}
+
+	}
+}
+
+// refreshCerts is the main work item in certificate store, responsible for finding the right certificate
+func (c *winCertStoreHandler) refreshCerts() error {
```
rfratto, line 135: Use a named return here:

```
func (c *winCertStoreHandler) refreshCerts() (err error) {
   // ... 
}
``` 

named returns are guaranteed to have their values set by a `return`, ensuring that you can check for the value of the error on return without worrying about if a local variable was set.

pkg/server/tls_windows.go
```diff
@@ -0,0 +1,342 @@
...
+	c.serverIdentity = serverIdentity
+	return nil
+}
+
+func (c *winCertStoreHandler) findServerCertificate() (certstore.Identity, error) {
+	return c.findCertificate(c.cfg.ServerSystemStore, c.cfg.ServerStore, c.cfg.ServerIssuerCommonNames, c.cfg.ServerTemplateID, nil, c.getStore)
+}
+func (c *winCertStoreHandler) findClientCertificate() (certstore.Identity, error) {
+	return c.findCertificate(c.cfg.ClientSystemStore, c.cfg.ClientStore, c.cfg.ClientIssuerCommonNames, c.cfg.ClientTemplateID, c.subjectRegEx, c.getStore)
+}
```
rfratto, line 189: Is it required to set both server and client certificates? IMO client should be optional (though server must be set if client is set)

pkg/server/tls_windows.go
```diff
@@ -0,0 +1,342 @@
...
+		return err
+	}
+	clientIdentity, err = c.findClientCertificate()
+	if err != nil {
+		return err
+	}
+	cc, err := clientIdentity.Certificate()
+	if err != nil {
+		return err
+	}
```
rfratto, line 176: IMO these should be wrapped errors so we know what specifically is failing

pkg/server/tls_windows.go
```diff
@@ -0,0 +1,342 @@
...
+			c.serverIdentity.Close()
+		}
+		c.serverCert = nil
+		c.serverSigner = nil
+		return
+	case <-time.After(c.cfg.ServerRefreshInterval):
+		err := c.refreshCerts()
+		if err != nil {
+			return
+		}
```
rfratto, line 129: Two things:

1. This should be logging the error 
2. Is returning expected? Or should we try again later?

pkg/server/tls_windows.go
```diff
@@ -0,0 +1,342 @@
...
+	var clientIdentity certstore.Identity
+	var err error
+	// This handles closing all our various handles
+	defer func() {
+		// we have to keep the server identity open if we want to use it, BUT only if an error occurred, else we need it
+		// open to sign
+		if serverIdentity != nil && err != nil {
+			serverIdentity.Close()
+		}
+		// Client identity does need to be open since we dont need to sign
```
rfratto, line 152: s/does/doesn't?

pkg/server/tls_windows.go
```diff
@@ -0,0 +1,342 @@
...
+		return nil, err
+	}
+	if len(filtered) == 0 {
+		return nil, fmt.Errorf("no certificates found")
+	}
+	// order oldest to newest
+	sort.Slice(filtered, func(i, j int) bool {
+		// Already accessed this so the error will not happen
+		a, _ := filtered[i].Certificate()
+		b, _ := filtered[j].Certificate()
```
rfratto, line 245: nit: `certI`/`certJ` to make it clearer that they align with the i/j variables

pkg/server/tls_windows.go
```diff
@@ -0,0 +1,342 @@
...
+		}
+	}
+	if validIdentity == nil {
+		return nil, fmt.Errorf("no certificates found")
+	}
+
+	return validIdentity, nil
+}
+
+func (c *winCertStoreHandler) filterByIssuerCommonNames(input []certstore.Identity, commonNames []string) ([]certstore.Identity, error) {
```
rfratto, line 265: The filter functions are kind of repetitive, can we make a generic filter function that these can use?

```go
func (c *winCertStoreHandler) filterCerts(in []certstore.Identity, f func(*x509.Certificate) bool} ([]certstore.Identity, error) {
  res := make([]certstore.Identity, 0, len(in)) 
  for _, ident := range in {
    cert, err := ident.Certificate()
    if err != nil {
      return err 
    }
    if f(cert) {
      res = append(res, ident)
    }
  }
  return res
}
```

pkg/server/tls_windows.go
```diff
@@ -0,0 +1,342 @@
...
+	_, err := clientCert.Verify(opts)
+	return err
+
+}
+
+type templateInformation struct {
+	Template     asn1.ObjectIdentifier
+	MajorVersion int
+	MinorVersion int
+}
```
rfratto, line 342: nit: can this be declared directly after filterByTemplateID? I was thrown off by what was using this on first glance.

pkg/server/tls_windows_test.go
```diff
@@ -0,0 +1,295 @@
...
+	serverSt.identities = append(serverSt.identities, newFakeIdentity(sc))
+	findCert := func(systemStore, _ string) (certstore.Store, error) {
+		return serverSt, nil
+	}
+	serverIdentity, err := c.findCertificate(c.cfg.ServerSystemStore, c.cfg.ServerStore, c.cfg.ServerIssuerCommonNames, c.cfg.ServerTemplateID, nil, findCert)
+	require.NoError(t, err)
+	foundCert, err := serverIdentity.Certificate()
+	require.NoError(t, err)
+	require.NotNil(t, serverIdentity)
+	require.True(t, foundCert == sc)
```
rfratto, line 34: nit: use `require.Equal` here so the error message shows the two non-equal objects

(ditto for other similar checks in this file)

pkg/server/tls_windows_test.go
```diff
@@ -0,0 +1,295 @@
...
+	sc := makeCert(time.Now().Add(time.Duration(-1)*time.Minute), time.Now().Add(5*time.Minute), []int{1, 2, 3}, "", "")
+	serverSt.identities = append(serverSt.identities, newFakeIdentity(sc))
+	findCert := func(systemStore, _ string) (certstore.Store, error) {
+		return serverSt, nil
+	}
+	serverIdentity, err := c.findCertificate(c.cfg.ServerSystemStore, c.cfg.ServerStore, c.cfg.ServerIssuerCommonNames, c.cfg.ServerTemplateID, nil, findCert)
+	require.NoError(t, err)
+	foundCert, err := serverIdentity.Certificate()
+	require.NoError(t, err)
+	require.NotNil(t, serverIdentity)
```
rfratto, line 33: Should this be moved directly below line 30? Ditto for other instances of this check in the file

pkg/server/tls_windows_test.go
```diff
@@ -0,0 +1,295 @@
...
+func (f fakeStore) Identities() ([]certstore.Identity, error) {
+	ids := make([]certstore.Identity, len(f.identities))
+	for i, id := range f.identities {
+		ids[i] = id
+	}
+	return ids, nil
+}
+
+func (f fakeStore) Import(data []byte, password string) error {
+	panic("implement me")
```
rfratto, line 219: Should this be changed to "should not be called" if we're not intending to use this? Ditto for the other panics in this file

rfratto (commented):
🎉 nice, this is looking good. Most of my concerns so far are around the worker goroutine semantics

docs/user/configuration/server-config.md
```diff
@@ -60,8 +60,7 @@ The `server_tls_config` configures TLS.
 
 ## windows_certificate_filter
 
-The `windows_certificate_filter` configures usage of the Windows Certificate store
-NOTE: cert_file, key_file, and client_ca_file are not used when this is enabled
+The `windows_certificate_filter` configures usage of the Windows Certificate store. It is invalid to set cert_file, key_file, and client_ca_file when using windows_certificate_filter.
```
brendamuir, line 63:
```suggestion
The `windows_certificate_filter` configures the use of the Windows Certificate store. Set cert_file, key_file, and client_ca_file are invalid settings when using the windows_certificate_filter.
```

brendamuir (changes requested):

pkg/server/tls_windows.go
mattdurham: I originally played around with something similar, but I think it reduced clarity. For instance, you need to pass in two functions (or add the check in the caller) to see if you need to do any matching. I could move them out of inline funcs but the gain is minimal.

Will think on if there is a cleaner path. 

```
c.filterCertificate(identities, func() bool{
	return len(commonNames) != 0
}, func(cert *x509.Certificate) bool {
	for _, cfgName := range commonNames {
		if cert.Issuer.CommonName == cfgName {
			return true
		}
	}
	return false
})


func (c *winCertStoreHandler) filterCertificate(input []certstore.Identity, shouldMatch func() bool, match func(certificate *x509.Certificate) bool) ([]certstore.Identity, error) {
	if !shouldMatch() {
		return input, nil
	}
	returnIdentities := make([]certstore.Identity, 0)
	for _, identity := range input {
		cert, err := identity.Certificate()
		if err != nil {
			return nil, err
		}
		if match(cert) {
			returnIdentities = append(returnIdentities, identity)
		}
	}
	return returnIdentities, nil
}
```

pkg/server/tls_windows.go
mattdurham: Yes, good catch!

pkg/server/tls_windows.go
mattdurham: The config is protected by the parent tls mutex. Winmutex protects the found certificates.

pkg/server/tls_windows.go
mattdurham: `does NOT` is what I meant to write, good catch =)

pkg/server/tls_windows.go
mattdurham: Agreed, and with fixing the for loop it will try again at the interval. In the case of an error in refreshCerts the previous certificates will be used.

pkg/server/tls_windows.go
mattdurham: Adding mutex around the checking of the refresh interval

docs/user/configuration/server-config.md
```diff
@@ -53,4 +53,46 @@ The `server_tls_config` configures TLS.
...
+
+## windows_certificate_filter
+
+The `windows_certificate_filter` configures usage of the Windows Certificate store
+NOTE: cert_file, key_file, and client_ca_file are not used when this is enabled
+
+```yaml
+
+# Name of the system store to look for the Client Certificate ex LocalMachine, CurrentUser 
+[client_system_store: <string>]
```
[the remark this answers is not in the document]
mattdurham: I don't think we can define a default value, it's likely LocalMachine, but it could also be the Services location too.

docs/user/configuration/server-config.md
```diff
@@ -53,4 +53,46 @@ The `server_tls_config` configures TLS.
...
+[client_store: <string>]
+
+# Array of issuer common names to check against
+[client_issuer_common_names: [<string>]]
+
+# Regular expression to match Subject name
+[client_subject_regex: <string>]
+
+# Client Template ID to match in ASN1 format ex "1.2.3"
+[client_template_id: <string>]
```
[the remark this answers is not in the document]
mattdurham: Took a look and it is cleaner to go separate blocks so done!

pkg/server/tls_windows.go
mattdurham: Refactored so that it's more intelligent about when to throw if client certs are not set.

docs/user/configuration/server-config.md
```diff
@@ -53,4 +53,61 @@ The `server_tls_config` configures TLS.
...
+
+
+### windows_client_config
+
+```yaml
+# Name of the system store to look for the Client Certificate ex LocalMachine, CurrentUser 
+[system_store: <string>]
+
+# Name of the store to look for the Client Certificate ex My, CA
+[store: <string>]
```
rfratto, line 81: Are we sure these aren't required? What will happen if there's no system store / store set?

pkg/server/tls_certstore_windows.go
```diff
@@ -0,0 +1,374 @@
+package server
```
rfratto, line 1: `*tlsListener.Close` should stop the winCertStoreHandler goroutine if it's set

rfratto (approved):
🎉

PR #1581, opened 2022-04-08
App o11y receiver intergration

domasx2:
#### PR Description

This PR introduces the app o11y receiver intergation (v2), built on `dev.frontend-receiver` branch. It opens a HTTP server with an endpoint that can accept custom telemetry payloads from [the Grafana Javascript Agent](https://github.com/grafana/grafana-javascript-agent). These payloads include logs, exceptions, measurements and traces. Integration will format logs, exceptions and measurements into log lines and forward them to a logs instance. Traces are forwarded to a traces instance. Stack traces can be transformed using sourcemaps, which can be downloaded or read from file system. HTTP endpoint supports CORS, basic rate limiting and basic API key auth. 

#### Notes to the Reviewer

[Grafana Javascript Agent](https://github.com/grafana/grafana-javascript-agent) still doesn't support sending traces, actively working on that. Otel JS libraries are giving us some headaches :) 

I haven't been able to easily come up with tests for the integration setup (`app_o11y_receiver.go`), due to all the dependencies that can't easily be mocked. Willing to take another stab though if required

Trace ingestion from within agent is hacked together by introducing a `PushReceiver` that is added to traces pipelines that are referenced by this integration. It exposes a traces consumer through it's factory. 

*some musings about present and future:* So currently this is using a custom payload and http handler impl, simple and obvious solution. Traces use otel format, but are embedded within our custom payload. However as we watch otel js mature a bit, and talk to our own otel folks, we're warming up to the idea of using otel traces channel for everything. Embedding logs, exceptions etc into span events and such, then using custom processors to extract and transform. Trying to POC this, might eventually become a v2 down the road

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated


CHANGELOG.md
```diff
@@ -10,6 +10,12 @@ internal API changes are not present.
 Main (unreleased)
 -----------------
 
+### Features
+
+- (beta) Add app o11y integration. This depends on integrations-next being enabled
```
rfratto, line 15: I think given we're already thinking about a v2 that goes all in on OTLP payloads, maybe this should be experimental instead of beta?

@mattdurham do you think this should have a feature flag?

docs/configuration/integrations/integrations-next/app-o11y-receiver-config.md
```diff
@@ -0,0 +1,118 @@
++++
+title = "app_o11y_config"
++++
+
+# app_o11y_receiver_config
```
rfratto, line 5: nit: Not a lot of people know what `o11y` is, I think we should pick a different name. Maybe something like frontend_agent_receiver?

docs/configuration/integrations/integrations-next/app-o11y-receiver-config.md
```diff
@@ -0,0 +1,118 @@
...
+
+The `app_o11y_receiver_config` block configures the `app-o11y-receiver`
+integration. This integration exposes a http endpoint that can receive telemetry
+from the [grafana javascript agent](https://github.com/grafana/grafana-javascript-agent)
```
rfratto, line 9: nit: I don't want to bog this down with a full grammar review, but I think any mention of Grafana Javascript Agent should be capitalized like it is in the project.

docs/configuration/integrations/integrations-next/app-o11y-receiver-config.md
```diff
@@ -0,0 +1,118 @@
...
+  # Configure rate limiting. The HTTP server of the App observability implements
+  # a token bucket rate limitng algorithm in which we can configure the maximum RPS
+  # as well as the burstiness (peaks of RPS)
+  rate_limiting:
+    [enabled: <boolean> | default = false]
+    [rps: <number> | default = 100]
+    [burstiness: <number> | default = 50]
+
+  # If configured, incoming requests will be required to specify this key in "x-api-key" header
+  [api_key: <string>]
```
rfratto, line 48: naming nit: maybe this should be `validate_api_key` so it's clearer to users that we're talking about API keys to check for and not an API key that the receiver uses? 

Also, we should support a `_file`-based form of this so secrets aren't stored in the config file directly.

docs/configuration/integrations/integrations-next/app-o11y-receiver-config.md
```diff
@@ -0,0 +1,118 @@
...
+
+  # Configure rate limiting. The HTTP server of the App observability implements
+  # a token bucket rate limitng algorithm in which we can configure the maximum RPS
+  # as well as the burstiness (peaks of RPS)
+  rate_limiting:
+    [enabled: <boolean> | default = false]
+    [rps: <number> | default = 100]
+    [burstiness: <number> | default = 50]
+
+  # If configured, incoming requests will be required to specify this key in "x-api-key" header
```
rfratto, line 47: Should this be using Bearer auth instead of a unique header?

docs/configuration/integrations/integrations-next/app-o11y-receiver-config.md
```diff
@@ -0,0 +1,118 @@
...
+  # Max allowed payload size in bytes for the JSON payload. Interanlly the
+  # Content-Length header is used to make this check
+  [max_allowed_payload_size: <number> | default = 0]
+
+  # Traces instance to send traces to. This assumes that you have a traces config with such instance defined
+  [traces_instance: <string> | default = ""]
+
+  # Logs instance to send logs and exceptions to. This assumes that you have a logs
+  # config with the instance defined
+  [logs_instance: <string> | default = ""]
```
rfratto, line 59: We generally try to default these to `"default"` instead of empty strings since that's the normal naming convention.

docs/configuration/integrations/integrations-next/app-o11y-receiver-config.md
```diff
@@ -0,0 +1,118 @@
...
+
+  # Server config refres to the HTTP endpoint that the integration will be exposing
+  # to receive data from.
+  server:
+    [host: <string> | default = "0.0.0.0"]
```
rfratto, line 64: We're trying to lock things down by default more often now:

```suggestion
    [host: <string> | default = "127.0.0.1"]
```

docs/configuration/integrations/integrations-next/app-o11y-receiver-config.md
```diff
@@ -0,0 +1,118 @@
...
+  # Traces instance to send traces to. This assumes that you have a traces config with such instance defined
+  [traces_instance: <string> | default = ""]
+
+  # Logs instance to send logs and exceptions to. This assumes that you have a logs
+  # config with the instance defined
+  [logs_instance: <string> | default = ""]
+
+  # Server config refres to the HTTP endpoint that the integration will be exposing
+  # to receive data from.
+  server:
```
rfratto, line 63: IMO the CORS and rate limiter settings should be moved to this block since they're server-specific.

docs/configuration/integrations/integrations-next/app-o11y-receiver-config.md
```diff
@@ -0,0 +1,118 @@
...
+    [host: <string> | default = "0.0.0.0"]
+    [port: <number> | default = 8080]
```
rfratto, line 65: Should we maybe default this to 12347 since the official agent ports start at 12345 (http) and 12346 (grpc)?

pkg/config/config.go
```diff
@@ -185,6 +186,26 @@ func (c *Config) Validate(fs *flag.FlagSet) error {
...
+			if o11yConf.ReceiverConfig.TracesInstance == "" {
+				continue
+			}
+			for idx, instanceConf := range c.Traces.Configs {
+				if instanceConf.Name == o11yConf.ReceiverConfig.TracesInstance {
+					c.Traces.Configs[idx].PushReceiver = true
+				}
+			}
+		}
+	}
```
rfratto, line 207: Can we have every trace instance have an internal push receiver by default? That we don't need a special loop here.

pkg/integrations/install/install.go
```diff
@@ -27,6 +27,6 @@ import (
 	// v2 integrations
 	//
 
-	_ "github.com/grafana/agent/pkg/integrations/v2/agent" // register agent
-	_ "github.com/grafana/agent/pkg/integrations/v2/eventhandler"
+	_ "github.com/grafana/agent/pkg/integrations/v2/agent"             // register agent
+	_ "github.com/grafana/agent/pkg/integrations/v2/app_o11y_receiver" // register app_o11y_receiver
```
rfratto, line 31: You accidentally removed the other integration here :)

domasx2:
> @domasx2 heads up, the design docs you linked to aren't viewable by the public; should they be?

oops, removed references for now

pkg/integrations/v2/app_o11y_receiver/app_o11y_receiver.go
```diff
@@ -0,0 +1,241 @@
...
+
+	if len(c.ReceiverConfig.LogsInstance) > 0 {
+		logsInstance := globals.Logs.Instance(c.ReceiverConfig.LogsInstance)
+		if logsInstance == nil {
+			return nil, fmt.Errorf("logs instance \"%s\" not found", c.ReceiverConfig.LogsInstance)
+		}
+		lokiExporter := exporters.NewLogsExporter(
+			l,
+			exporters.LogsExporterConfig{
+				LogsInstance:     logsInstance,
```
rfratto, line 112: You can't hold a pointer to a logs instance throughout the lifetime of an integration; the implementation of the logs subsystem will create new pointers if the logs instance changes. 

That means you have to do a lookup by name any time the integration needs to access the instance. The same should be done for metrics and tracing (when added).

pkg/integrations/v2/app_o11y_receiver/app_o11y_receiver.go
```diff
@@ -0,0 +1,241 @@
...
+				Labels:           c.ReceiverConfig.LogsLabels,
+				SendEntryTimeout: c.ReceiverConfig.LogsSendTimeout,
+			},
+			sourcemapStore,
+		)
+		exp = append(exp, lokiExporter)
+	}
+
+	if len(c.ReceiverConfig.TracesInstance) > 0 {
+		tracesInstance := globals.Tracing.Instance(c.ReceiverConfig.TracesInstance)
```
rfratto, line 122: ditto here: you can't hold a pointer to the instance

pkg/integrations/v2/app_o11y_receiver/app_o11y_receiver.go
```diff
@@ -0,0 +1,241 @@
...
+		},
+		Source: fmt.Sprintf("%s/%s", i.integrationName, i.instanceID),
+	}
+
+	group.Targets = append(group.Targets, model.LabelSet{
+		model.AddressLabel:     model.LabelValue(ep.Host),
+		model.MetricsPathLabel: model.LabelValue(path.Join(ep.Prefix, "/metrics")),
+	})
+
+	return []*targetgroup.Group{group}
```
rfratto, line 179: Can we refactor this code so that it's not being duplicated across multiple integrations?

pkg/integrations/v2/app_o11y_receiver/app_o11y_receiver.go
```diff
@@ -0,0 +1,241 @@
...
+	cfg.ServiceDiscoveryConfigs = sd
+	cfg.ScrapeInterval = i.common.Autoscrape.ScrapeInterval
+	cfg.ScrapeTimeout = i.common.Autoscrape.ScrapeTimeout
+	cfg.RelabelConfigs = i.common.Autoscrape.RelabelConfigs
+	cfg.MetricRelabelConfigs = i.common.Autoscrape.MetricRelabelConfigs
+
+	return []*autoscrape.ScrapeConfig{{
+		Instance: i.common.Autoscrape.MetricsInstance,
+		Config:   cfg,
+	}}
```
rfratto, line 201: Can we refactor this code so that it's not being duplicated across multiple integrations?

pkg/integrations/v2/app_o11y_receiver/app_o11y_receiver.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+// RunIntegration implements Integration
+func (i *appo11yIntegration) RunIntegration(ctx context.Context) error {
+	mdlw := middleware.New(middleware.Config{
+		Recorder: metrics.NewRecorder(metrics.Config{Registry: i.reg, Prefix: utils.MetricsNamespace}),
+	})
```
rfratto, line 208: Instead of using a new middleware package, can we use the existing one that pkg/server uses from github.com/weaveworks/common?

pkg/integrations/v2/app_o11y_receiver/app_o11y_receiver.go
```diff
@@ -0,0 +1,241 @@
...
+
+	r := mux.NewRouter()
+	r.Handle("/collect", i.handler.HTTPHandler(i.logger))
```
rfratto, line 211: Should this give an allowlist of HTTP methods? (i.e., PUSH only, GET only, etc)

pkg/integrations/v2/app_o11y_receiver/app_o11y_receiver.go
```diff
@@ -0,0 +1,241 @@
...
+	srv := &http.Server{
+		Addr:    fmt.Sprintf("%s:%d", i.conf.Server.Host, i.conf.Server.Port),
+		Handler: std.Handler("", mdlw, r),
+	}
+	errChan := make(chan error)
+
+	go func() {
+		level.Info(i.logger).Log("msg", "starting app o11y receiver", "host", i.conf.Server.Host, "port", i.conf.Server.Port)
+		if err := srv.ListenAndServe(); err != http.ErrServerClosed {
+			errChan <- err
```
rfratto, line 222: This may block forever if the integration exits before an error gets written; use a buffered channel for errChan to avoid that.

pkg/integrations/v2/app_o11y_receiver/config/config.go
```diff
@@ -0,0 +1,98 @@
...
+	DefaultRateLimitingRPS = 100
+	// DefaultRateLimitingBurstiness is the default burstiness factor of the
+	// token bucket algorigthm
+	DefaultRateLimitingBurstiness = 50
+	// DefaultMaxPayloadSize is the max paylad size in bytes
+	DefaultMaxPayloadSize = 5e6
+)
+
+// DefaultConfig holds the default configuration of the receiver
+var DefaultConfig = AppO11yReceiverConfig{
```
rfratto, line 17: nit: we don't need to explicitly set zero values in the default config.

pkg/integrations/v2/app_o11y_receiver/config/config.go
```diff
@@ -0,0 +1,98 @@
...
+	LogsInstance          string             `yaml:"logs_instance,omitempty"`
+	LogsLabels            map[string]string  `yaml:"logs_labels,omitempty"`
+	LogsSendTimeout       int                `yaml:"logs_send_timeout,omitempty"`
+	SourceMaps            SourceMapConfig    `yaml:"sourcemaps,omitempty"`
+}
+
+// UnmarshalYAML implements the Unmarshaller interface
+func (c *AppO11yReceiverConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	*c = DefaultConfig
+	type cA AppO11yReceiverConfig
```
rfratto, line 87: nit: can you give this a more readable name? If there's an unmarshal error, the type name here appears to the user, and `cA` wouldn't be underable to them.

pkg/integrations/v2/app_o11y_receiver/exporters/logs.go
```diff
@@ -0,0 +1,128 @@
...
+)
+
+type logsInstance interface {
+	SendEntry(entry api.Entry, dur time.Duration) bool
+}
+
+// LogsExporterConfig holds the configuration of the logs exporter
+type LogsExporterConfig struct {
+	SendEntryTimeout int
+	LogsInstance     logsInstance
```
rfratto, line 27: nit: the type of an exported field on an exported struct should also be exported (i.e., export `logsInstance`)

pkg/integrations/v2/app_o11y_receiver/exporters/logs.go
```diff
@@ -0,0 +1,128 @@
...
+	return err
+}
+
+func (le *LogsExporter) sendKeyValsToLogsPipeline(kv *utils.KeyVal) error {
+	line, err := logfmt.MarshalKeyvals(utils.KeyValToInterfaceSlice(kv)...)
+	if err != nil {
+		level.Error(le.logger).Log("msg", "failed to logfmt a frontend log event", "err", err)
+		return err
+	}
+	sent := le.li.SendEntry(api.Entry{
```
rfratto, line 94: (This is where we should be looking up the instance instead of holding a pointer)

pkg/integrations/v2/app_o11y_receiver/exporters/logs.go
```diff
@@ -0,0 +1,128 @@
...
+		Labels: le.labelSet(kv),
+		Entry: logproto.Entry{
+			Timestamp: time.Now(),
+			Line:      string(line),
+		},
+	}, le.seTimeout)
+	if !sent {
+		level.Warn(le.logger).Log("msg", "failed to log frontend log event to logs pipeline")
+		return fmt.Errorf("Failed to send app o11y event to logs pipeline")
```
rfratto, line 103: nit: error messages should not start with an uppercase letter or end with punctuation

```suggestion
		return fmt.Errorf("failed to send app o11y event to logs pipeline")
```

pkg/integrations/v2/app_o11y_receiver/exporters/logs.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	return nil
+}
+
+func (le *LogsExporter) labelSet(kv *utils.KeyVal) prommodel.LabelSet {
+	set := make(prommodel.LabelSet)
```
rfratto, line 109:
```suggestion
	set := make(prommodel.LabelSet, len(le.labels))
```

pkg/integrations/v2/app_o11y_receiver/exporters/logs_test.go
```diff
@@ -0,0 +1,120 @@
...
+	"strings"
+	"testing"
+	"time"
+
+	kitlog "github.com/go-kit/log"
+	"github.com/grafana/agent/pkg/integrations/v2/app_o11y_receiver/models"
+	"github.com/grafana/loki/clients/pkg/promtail/api"
+	prommodel "github.com/prometheus/common/model"
+
+	"github.com/stretchr/testify/assert"
```
rfratto, line 16: nit: we tend to use /require instead of assert, since most assertions that are being made are blockers for the rest of the test to pass.

pkg/integrations/v2/app_o11y_receiver/exporters/receiver_metrics.go
```diff
@@ -0,0 +1,60 @@
...
+		totalMeasurements: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: utils.MetricsNamespace,
+			Name:      "total_measurements",
+			Help:      "Total number of ingested measurements",
+		}),
+		totalExceptions: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: utils.MetricsNamespace,
+			Name:      "total_exceptions",
+			Help:      "Total number of ingested exceptions",
+		}),
```
rfratto, line 41: nits:

1. Don't use Namespace from the opts, because that makes it harder to search through the codebase to find where a metric is defined. Instead, pass the full literal name in to the Name field.
2. The naming conventions here don't match Prometheus; the counters should be `logs_total`, not `total_logs`.

pkg/integrations/v2/app_o11y_receiver/exporters/receiver_metrics_test.go
```diff
@@ -0,0 +1,149 @@
...
+		{},
+		{},
+		{},
+	}
+	payload.Exceptions = []models.Exception{
+		{},
+		{},
+		{},
+		{},
+	}
```
rfratto, line 65: This can be simplified pretty heavily:

```suggestion
        payload.Logs = make([]models.Log, 2)
        payload.Measurements = make([]models.Measurement, 3)
        payload.Exceptions = make([]models.Exception, 4)
```

ditto this suggestion for the other instances in this file

pkg/integrations/v2/app_o11y_receiver/config/config.go
```diff
@@ -0,0 +1,98 @@
...
+	DownloadFromOrigins []string                `yaml:"download_origins,omitempty"`
+	DownloadTimeout     time.Duration           `yaml:"download_timeout,omitempty"`
+	FileSystem          []SourceMapFileLocation `yaml:"filesystem,omitempty"`
+}
+
+// AppO11yReceiverConfig is the configuration struct of the
+// integration
+type AppO11yReceiverConfig struct {
+	CORSAllowedOrigins    []string           `yaml:"cors_allowed_origins,omitempty"`
+	RateLimiting          RateLimitingConfig `yaml:"rate_limiting,omitempty"`
```
rfratto, line 73: nit: make this a pointer with its own UnmarshalYAML implementation, so:

1. You don't need the enabled field (if it's not defined, it's not enabled)
2. You can more easily default the rate limit when it is provided

pkg/integrations/v2/app_o11y_receiver/handler/handler.go
```diff
@@ -0,0 +1,124 @@
...
+		if conf.RateLimiting.Burstiness > 0 {
+			b = conf.RateLimiting.Burstiness
+		}
+		rateLimiter = ratelimiting.NewRateLimiter(rps, b)
+	}
+
+	exporterErrorsCollector := prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: utils.MetricsNamespace,
+		Subsystem: "exporter",
+		Name:      "errors",
```
rfratto, line 48: ditto nit about only using the Name field for defining new metrics

pkg/integrations/v2/app_o11y_receiver/handler/handler.go
```diff
@@ -0,0 +1,124 @@
...
+
+		var wg sync.WaitGroup
+
+		for _, exporter := range ar.exporters {
+			wg.Add(1)
+			go func(exp exporters.AppO11yReceiverExporter) {
+				defer wg.Done()
+				if err := exp.Export(r.Context(), p); err != nil {
+					level.Error(logger).Log("msg", "exporter error", "exporter", exp.Name(), "error", err)
+					ar.exporterErrorsCollector.WithLabelValues(exp.Name()).Inc()
```
rfratto, line 105: Should this cause a non-200 to be returned?

pkg/integrations/v2/app_o11y_receiver/models/measurement_test.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	assert.Equal(t, len(tcs.expected.Values), len(m.Values))
+	for k, v := range m.Values {
+		assert.NotEmpty(t, tcs.expected.Values[k])
+		assert.Equal(t, v, tcs.expected.Values[k])
+	}
+}
+
+func TestMeasureMents(t *testing.T) {
```
rfratto, line 52:
```suggestion
func TestMeasurements(t *testing.T) {
```

pkg/integrations/v2/app_o11y_receiver/models/meta.go
```diff
@@ -0,0 +1,150 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/grafana/agent/pkg/integrations/v2/app_o11y_receiver/utils"
+)
+
+// Sdk holds metadata about the app agent that produced the event
+type Sdk struct {
```
rfratto, line 11: nit: this should be renamed to SDK to match with go naming conventions for exported acronyms.

pkg/integrations/v2/app_o11y_receiver/tools/ratelimiting/ratelimiter.go
```diff
@@ -0,0 +1,73 @@
+package ratelimiting
```
rfratto, line 1: Can we use golang.org/x/time/rate instead of creating our own rate limiter?

rfratto (commented):
👋 thanks for working on this! This is my first pass review. Given the size of the PR, I suspect we'll go through a few review cycles before this is ready to merge. 

Aside from my comments, I have a few high-level thoughts:

I'm concerned the files and packages are being overpartitioned here, making it harder to read files as a story: i.e., why does `exporters/exporter.go` define nothing but a single interface? You can't look at that file alone and understand how it fits into the broader picture.  

IMO, the need for something like a `models/` package generally points to overpartitioning: types should be moved into the packages containing the business logic that consumes them, and packages should be combined into one if that's not possible. We generally treat a package that only defines types as an anti-pattern in this specific codebase. 

Similarly, it seems like a lot of types added for testing are being exported and polluting the consumable API design: `NewSourceMapStore` accepts an HTTPClient interface and a FileService interface, which are never set except for tests: this tells me they should be kept as internal to clean up the API and reduce the scope of the mental model for the overall integration.

docs/configuration/integrations/integrations-next/app-o11y-receiver-config.md
domasx2: - [ ] captialize

pkg/integrations/v2/app_o11y_receiver/handler/handler.go
domasx2: Given that a) Partial success is possible, eg ingesting logs succeeds but traces fail b) Client shouldn't do anything if this fails, we decided to just return `202 Accepted `always. The thinking is that if at some point we need client to know the final outcome, we'd implement some sort of multiple result response

pkg/integrations/v2/app_o11y_receiver/models/meta.go
domasx2: dang, I knew this! 
- [ ] capitalize

docs/configuration/integrations/integrations-next/app-o11y-receiver-config.md
rfratto: I'll leave it up to you, I just would prefer the name to not have `o11y` in it :)

docs/configuration/integrations/integrations-next/app-o11y-receiver-config.md
domasx2: - [ ] default to 12347

pkg/config/config.go
domasx2: That was original impl, but @mapno  suggested to add only when necessary. Shall I revert to including by default?

pkg/integrations/install/install.go
domasx2: - [ ] oops, bad merge. Restore

pkg/integrations/v2/app_o11y_receiver/app_o11y_receiver.go
domasx2: - [ ] attempt to commonize

pkg/integrations/v2/app_o11y_receiver/app_o11y_receiver.go
domasx2: Hmm, there was some issue with using it I think, but now am not sure... will check again
- [ ] try using weaveworks http metrics middleware

pkg/integrations/v2/app_o11y_receiver/app_o11y_receiver.go
domasx2: - [ ] limit to PUSH and OPTIONS

pkg/integrations/v2/app_o11y_receiver/app_o11y_receiver.go
domasx2: - [ ] use buffered channel

pkg/integrations/v2/app_o11y_receiver/config/config.go
domasx2: - [ ] dont explicitly set zero values

pkg/integrations/v2/app_o11y_receiver/config/config.go
domasx2: - [ ] better naming

pkg/integrations/v2/app_o11y_receiver/exporters/logs_test.go
domasx2: - [ ] use `/require` instead

pkg/integrations/v2/app_o11y_receiver/config/config.go
domasx2: It should be enabled by default, it's a bad idea to go without, even accidentally :thinking:  Maybe we should just remove the `enabled` flag

pkg/integrations/v2/app_o11y_receiver/exporters/receiver_metrics.go
domasx2: - [ ] update

rfratto:
Could you resolve the merge conflicts? 🙏 

I didn't have as much time as I wanted to today to give a second review on this, but I'll give it a closer look tomorrow.

domasx2:
> Could you resolve the merge conflicts? pray
Done!

docs/configuration/integrations/integrations-next/app-agent-receiver-config.md
```diff
@@ -0,0 +1,118 @@
++++
+title = "app_agent_config"
++++
+
+# app_agent_receiver_config
+
+The `app_agent_receiver_config` block configures the `app-agent-receiver`
```
rfratto, line 7: Should this use underscores since that's the name of the integration you'd specify in the config file?

```suggestion
The `app_agent_receiver_config` block configures the `app_agent_receiver`
```

pkg/config/config.go
```diff
@@ -188,6 +189,26 @@ func (c *Config) Validate(fs *flag.FlagSet) error {
...
+			if appAgentReceiverConf.TracesInstance == "" {
+				continue
+			}
+			for idx, instanceConf := range c.Traces.Configs {
+				if instanceConf.Name == appAgentReceiverConf.TracesInstance {
+					c.Traces.Configs[idx].PushReceiver = true
+				}
+			}
+		}
+	}
```
rfratto, line 210: @mapno WDYT about having every tracing instance always have a PushReceiver? Domas mentioned you suggested only enabling it when needed, but I'm concerned about the effect it has on pkg/config here. 

Can you talk a little bit about the concerns you raised?

pkg/integrations/v2/app_agent_receiver/app_agent_receiver.go
```diff
@@ -0,0 +1,193 @@
...
+	_ integrations.MetricsIntegration = (*appAgentReceiverIntegration)(nil)
+)
+
+// NewIntegration converts this config into an instance of an integration
+func (c *Config) NewIntegration(l log.Logger, globals integrations.Globals) (integrations.Integration, error) {
+	reg := prometheus.NewRegistry()
+	sourcemapLogger := log.With(l, "subcomponent", "sourcemaps")
+	sourcemapStore := NewSourceMapStore(sourcemapLogger, c.SourceMaps, reg, nil, nil)
+
+	receiverMetricsExporter := NewReceiverMetricsExporter(ReceiverMetricsExporterConfig{
```
rfratto, line 48: nit: IMO, I don't think anything in this package aside from the config structs need to be exported 🤷

pkg/integrations/v2/app_agent_receiver/handler.go
```diff
@@ -0,0 +1,124 @@
...
+// AppAgentReceiverHandler struct controls the data ingestion http handler of the receiver
+type AppAgentReceiverHandler struct {
+	exporters               []appAgentReceiverExporter
+	config                  *Config
+	rateLimiter             *rate.Limiter
+	exporterErrorsCollector *prometheus.CounterVec
+}
+
+// NewAppAgentReceiverHandler creates a new AppReceiver instance based on the given configuration
+func NewAppAgentReceiverHandler(conf *Config, exporters []appAgentReceiverExporter, reg *prometheus.Registry) AppAgentReceiverHandler {
```
rfratto, line 32: nit: pass around `prometheus.Registrar` instead of `*prometheus.Registry` to allow for different implementations (e.g., prometheus.WrapRegistrarWith, etc)

pkg/integrations/v2/app_agent_receiver/handler.go
```diff
@@ -0,0 +1,124 @@
...
+		// Check rate limiting state
+		if ar.config.Server.RateLimiting.Enabled {
+			if rsv := ar.rateLimiter.Reserve(); !rsv.OK() {
+				http.Error(w, http.StatusText(http.StatusTooManyRequests), http.StatusTooManyRequests)
+				return
+			}
+		}
+
+		// check API key if one is provided
+		if len(ar.config.Server.APIKey) > 0 && subtle.ConstantTimeCompare([]byte(r.Header.Get("x-api-key")), []byte(ar.config.Server.APIKey)) == 0 {
```
rfratto, line 79: nit: `"x-api-key"` feels a little magic-stringy, can you make it a package-level constant?

rfratto (approved):
Nice work! 

There's a lot going on here. I didn't give this an extremely in depth review, but the integration is isolated enough that I think it's safe to merge. 

Before merging, I'd request:

* Giving a chance for @mattdurham to review, if he wants to take a look 
* Considering my nits (you don't have to change anything, just consider them and resolve them if you're choosing not to make any changes). I don't need to review any code changes from the result of a nit comment :) 
* Waiting for @mapno to respond to my question about the push receiver and deciding if we'll change it or leave it as-is.  

@brendamuir is on PTO at the moment, but she'll be back next week. She might have suggestions about the docs changes, and you might have to resolve those in a follow up PR if this gets merged before then. 

I'm excited to have this!

docs/configuration/integrations/integrations-next/app-agent-receiver-config.md
```diff
@@ -0,0 +1,118 @@
...
+  [instance: <string>]
+
+  # Traces instance to send traces to. This assumes that you have a traces config with such instance defined
+  [traces_instance: <string> | default = ""]
+
+  # Logs instance to send logs and exceptions to. This assumes that you have a logs
+  # config with the instance defined
+  [logs_instance: <string> | default = ""]
+
+  # Server config refres to the HTTP endpoint that the integration will be exposing
```
mattdurham, line 42: "refers"?

docs/configuration/integrations/integrations-next/app-agent-receiver-config.md
```diff
@@ -0,0 +1,118 @@
...
+[download: <boolean> | default = false]
+
+# List of HTTP origins to download sourcemaps for
+[download_origins: []<string> | default = ["*"]]
+
+# Timeout for downloading compiled sources and sourcemaps
+[download_timeout: <duration> | default = "1s"]
+
+# Sourcemap locations on filesystem. Takes precedence over downloading if both methods are enabled
+[filesystem:
```
mattdurham, line 93: Brackets look odd here

pkg/integrations/v2/app_agent_receiver/config.go
```diff
@@ -0,0 +1,114 @@
...
+
+// Config is the configuration struct of the
+// integration
+type Config struct {
+	Common          common.MetricsConfig `yaml:",inline"`
+	Server          ServerConfig         `yaml:"server,omitempty"`
+	TracesInstance  string               `yaml:"traces_instance,omitempty"`
+	LogsInstance    string               `yaml:"logs_instance,omitempty"`
+	LogsLabels      map[string]string    `yaml:"logs_labels,omitempty"`
+	LogsSendTimeout int                  `yaml:"logs_send_timeout,omitempty"`
```
mattdurham, line 82: Better to be a duration?

docs/configuration/integrations/integrations-next/app-agent-receiver-config.md
```diff
@@ -0,0 +1,118 @@
...
+  # Traces instance to send traces to. This assumes that you have a traces config with such instance defined
+  [traces_instance: <string> | default = ""]
+
+  # Logs instance to send logs and exceptions to. This assumes that you have a logs
+  # config with the instance defined
+  [logs_instance: <string> | default = ""]
+
+  # Server config refres to the HTTP endpoint that the integration will be exposing
+  # to receive data from.
+  server:
```
mattdurham, line 44: Does this inherit the TLS config (or new tls), or is this only HTTP?

CHANGELOG.md
```diff
@@ -28,6 +28,10 @@ Main (unreleased)
 - Grafana Agent Operator: add support for integrations through an `Integration`
   CRD which is discovered by `GrafanaAgent`. (@rfratto)
 
+- (beta) Add app agent receiver integration. This depends on integrations-next being enabled
```
mattdurham, line 31: Thoughts on labeling experimental? The meaning behind beta is a bit broad but in general, BETA indicates we plan to move to production where experimental is much more fluid. This feels experimental. @rfratto thoughts?

mattdurham (commented):
Very large and gave it a relatively quick run-through. Only large thought is on tls and experimental vs beta. Overall looks like an awesome addition.

CHANGELOG.md
rfratto: I think that's going to be up to @domasx2. It sounds like @domasx2 and team are confident this is going to be something that gets delivered as stable, so calling it beta makes sense IMO.

pkg/integrations/v2/app_agent_receiver/app_agent_receiver.go
domasx2: I think there could be cases where we want to import these in another project. For example implementing a reicever endpoint in Grafana's backend, so that ootb it could receive and log frontend events w/o running an agent additionally.

pkg/integrations/v2/app_agent_receiver/handler.go
domasx2: - [ ] use`prometheus.Registrar`

pkg/integrations/v2/app_agent_receiver/handler.go
domasx2: - [ ] make `x-api-key` a package-level constant

docs/configuration/integrations/integrations-next/app-agent-receiver-config.md
domasx2: - [ ] "refers"!

docs/configuration/integrations/integrations-next/app-agent-receiver-config.md
domasx2: - [ ] refac

pkg/integrations/v2/app_agent_receiver/config.go
domasx2: - [ ] make it a `time.Duration`

CHANGELOG.md
domasx2: hm, I think let's make it experimental at least while we dogfood and build confidence
- [ ] experimental

docs/configuration/integrations/integrations-next/app-agent-receiver-config.md
domasx2: This is only HTTP for now.
If we have to support TLS, might need some help with implementing it

pkg/config/config.go
domasx2: Will do :)

docs/user/configuration/integrations/integrations-next/app-agent-receiver-config.md
```diff
@@ -0,0 +1,118 @@
...
+  [logs_send_timeout: <duration> | default = 2s]
+
+  # Sourcemap configuration for enabling stack trace transformation to original source locations
+  [sourcemaps: <sourcemap_config>]
+```
+
+## sourcemap_config
+
+```yaml
+# Wether agent should attempt to download compiled sources and source maps
```
brendamuir, line 83:
```suggestion
# Whether agent should attempt to download compiled sources and source maps
```

docs/user/configuration/integrations/integrations-next/app-agent-receiver-config.md
```diff
@@ -0,0 +1,118 @@
...
+# Sourcemap locations on filesystem. Takes precedence over downloading if both methods are enabled
+filesystem:
+  [- <sourcemap_file_location>]
+```
+
+## sourcemap_file_location
+
+```yaml
+# Source URL prefix. If a minified source URL matches this prefix,
+# A filepath is constructed by removing the prefix, prepending path below and appending ".map".
```
brendamuir, line 101:
```suggestion
# a filepath is constructed by removing the prefix, prepending path below and appending ".map".
```

docs/user/configuration/integrations/integrations-next/app-agent-receiver-config.md
```diff
@@ -0,0 +1,118 @@
...
++++
+
+# app_agent_receiver_config
+
+The `app_agent_receiver_config` block configures the `app_agent_receiver`
+integration. This integration exposes a http endpoint that can receive telemetry
+from the [Grafana Javascript Agent](https://github.com/grafana/grafana-javascript-agent)
+and forward it to logs, traces or metrics backends.
+
+Full reference of options:
```
brendamuir, line 12:
```suggestion
These are the options you have for configuring the app_agent_receiver integration.
```

brendamuir (changes requested):

mattdurham:
Looks good with those changes Brenda requested

mattdurham (approved):
Looks good once the changes brenda requested are done

brendamuir (approved):

PR #1611, opened 2022-04-13
Set up the default logger parameters

### LINKED ISSUE #1586: Default log level changed to debug in v0.24.0
I'm running v0.24.0 and noticed that the default log level implicitly changed to debug. 

Steps to reproduce:

1. Run the agent with at least command line flags `-server.http.address=127.0.0.1:9090 -server.grpc.address=127.0.0.1:9091 -config.enable-read-api`. Use a config file that does not set `log_level` in the server block.
2. curl `http://localhost:9090/-/config`.
3. Observe that `log_level: ""` is returned.

Expected default value is `info`.

tpaschalis:

#### PR Description
This PR re-introduces the correct default log level and log format, after being removed in https://github.com/grafana/agent/commit/b0c1e45c4ab884fd51c85687f51142ef85d2ca5d.

They used to have their default values implicitly set when registering the `-log.level` and `-log.format` flags, but it fell through when changing from weaveworks/common/server with agent/pkg/server.

There are a couple of places and ways to go about it, reintroducing them in main.go, in the server.Config.RegisterFlags method, or the server.Config.UnmarshalYAML method.

I prefer the option presented here, allowing the correct default values to be set, whether the `server:` top-level block exists in the YAML or not.

Fixes #1586 

#### Notes to the Reviewer
Is there any test we should add?

#### PR Checklist

- [x] CHANGELOG updated
- [x] Tests updated

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -344,6 +344,9 @@ type loaderFunc func(path string, fileType string, expandArgs bool, target *Conf
 // load allows for tests to inject a function for retrieving the config file that
 // doesn't require having a literal file on disk.
 func load(fs *flag.FlagSet, args []string, loader loaderFunc) (*Config, error) {
+	// Set up the default logger parameters.
+	DefaultConfig.Server.LogLevel.Set("info")
+	DefaultConfig.Server.LogFormat.Set("json")
 	var (
 		cfg = DefaultConfig
 
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -49,6 +49,7 @@ var (
 // DefaultConfig holds default settings for all the subsystems.
 var DefaultConfig = Config{
 	// All subsystems with a DefaultConfig should be listed here.
+	Server:                server.DefaultConfig,
 	Metrics:               metrics.DefaultConfig,
 	Integrations:          DefaultVersionedIntegrations,
 	EnableConfigEndpoints: false,
@@ -344,9 +345,6 @@ type loaderFunc func(path string, fileType string, expandArgs bool, target *Conf
 // load allows for tests to inject a function for retrieving the config file that
 // doesn't require having a literal file on disk.
 func load(fs *flag.FlagSet, args []string, loader loaderFunc) (*Config, error) {
-	// Set up the default logger parameters.
-	DefaultConfig.Server.LogLevel.Set("info")
-	DefaultConfig.Server.LogFormat.Set("json")
 	var (
 		cfg = DefaultConfig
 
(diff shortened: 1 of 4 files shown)


mattdurham (approved):
Looks good and straight forward

pkg/config/config.go
```diff
@@ -344,6 +344,9 @@ type loaderFunc func(path string, fileType string, expandArgs bool, target *Conf
 // load allows for tests to inject a function for retrieving the config file that
 // doesn't require having a literal file on disk.
 func load(fs *flag.FlagSet, args []string, loader loaderFunc) (*Config, error) {
+	// Set up the default logger parameters.
+	DefaultConfig.Server.LogLevel.Set("info")
+	DefaultConfig.Server.LogFormat.Set("json")
```
rfratto, line 349: Hmm, this might be a little unreliable and easy to accidentally remove during refactoring. (Also, the default log format should be logfmt, not json :)

It might be better for us to do a bit of a hack in `pkg/server.DefaultConfig` to make sure the defaults are set properly, something like:

```go
var DefaultConfig = Config{
  LogLevel: func() logging.Level {
     var l logging.Level
     // Pass a fake flag set to set the defaults
     l.RegisterFlags(flag.NewFlagSet("", flag.PanicOnError))
     return l
  }(),
  LogFormat: func() logging.Format {
     var l logging.Format
     l.RegisterFlags(flag.NewFlagSet("", flag.PanicOnError))
     return l
  }(),

  // ...
}
```

also, can you add a unit test for this and a changelog entry?

rfratto (changes requested):
My comments are mostly nits but at the very least this shouldn't be merged until the log format default is fixed to be logfmt and not json

pkg/config/config.go
tpaschalis: Yeah, unfortunately this change trips up the linter for not checking the error. I was just trying to find a way to set it up directly in the DefaultConfig, but it would still be kinda ugly.

rfratto: We could probably make our own structs that we can default more easily, though that would be slightly more work for you

pkg/server/config.go
```diff
@@ -55,6 +57,18 @@ var (
 	DefaultGRPCConfig = GRPCConfig{
 		// No non-zero defaults yet
 	}
+
+	emptyFlagSet    = flag.NewFlagSet("", flag.ExitOnError)
+	DefaultLogLevel = func() logging.Level {
```
tpaschalis, line 62: Really ugly, but unfortunately I think it's indeed the quickest way of getting valid logging.Level and logging.Format values 😕

rfratto: I've written worse :)

PR #1612, opened 2022-04-13
pkg/operator/config: don't set non-existent username_file field

### LINKED ISSUE #1601: Agent deployed with Operator and default configuration does not work out of the box
After installing the operator from helm chart and following https://grafana.com/docs/agent/latest/operator/custom-resource-quickstart/#step-1-deploy-grafanaagent-resource, the `grafana-agent-0` pod fails to initialise with the following message `flag provided but not defined: -reload-port`. I see that this option has been removed for v0.24, and the operator does not support versions below v0.24, so I'm not really sure why the operator is configuring the StatefulSet with that flag.

rfratto:
#### PR Description

PR #1441 introduced a change to use password_file for basic auth, but
accidentally also generated a username_file field which doesn't exist.

This is a partial revert of #1441, keeping only the password_file change.

Fixes #1601

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,9 @@ Main (unreleased)
 - Set the `Content-Type` HTTP header to `application/json` for API endpoints
   returning json objects. (@marctc)
 
+- Operator: fix issue where a `username_file` field was incorrectly set.
+  (@rfratto)
+
 ### Other changes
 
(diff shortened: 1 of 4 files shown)


rlankfo (approved):
LGTM

PR #1623, opened 2022-04-18
Update README.md to remove Cortex and swap in Mimir

09jvilla:
Swapped out Cortex for Mimir.

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ with:
 
 * [Grafana Cloud](https://grafana.com/products/cloud/)
 * [Grafana Enterprise Stack](https://grafana.com/products/enterprise/)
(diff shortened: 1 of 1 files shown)


rfratto (approved):
Thanks! It'll automerge after you sign the CLA and tests pass :)

09jvilla:
@rfratto note that there's still one hanging reference to Cortex which points to something in the 'examples'. I didn't want to make that change because the example still uses cortex. At some point we should probably switch the example (assuming you believe the example is still useful/gets traffic).

rfratto:
> I didn't want to make that change because the example still uses cortex. At some point we should probably switch the example (assuming you believe the example is still useful/gets traffic).

It should probably be renamed, the example became more of something for us to run a deployment for testing locally. But yes, changing it to use Mimir is a good idea and something we can do as a follow-up.

PR #1634, opened 2022-04-21
Make gossip test work on Ubuntu

domasx2:
#### PR Description

`gossip_test.go` fails on Ubuntu 20.04. It seems it checks for hardcoded network interfaces `eth0` or `en0`, but Ubuntu and most likely other distros use [predictable network interface names](https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/), so ethernet can look like `enp4s0f1`. 

This PR fixes it  by getting interface list via `net.Interfaces()`. Not 100% this is appropriate, but it does make the test pass :) I opened an [issue on the upstream ckit repo as well
](https://github.com/rfratto/ckit/issues)
#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/cluster/gossip_test.go
+++ b/pkg/cluster/gossip_test.go
@@ -3,8 +3,8 @@ package cluster
 import (
 	"fmt"
 	stdlog "log"
+	"net"
 	"os"
-	"runtime"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


rfratto (approved):
Thanks, I think this is OK in terms of the test :+1:

PR #1652, opened 2022-04-28
Add amd64-specific build tag for eBPF integration

tpaschalis:

#### PR Description
This PR constraints the eBPF integration to be only built on AMD64 systems, since cross-compilation fails on other architectures

No filed issues.

#### Notes to the Reviewer
If we think the current solution (limiting access to users running on a certain architecture) is acceptable, we can go ahead with it. It's not a defacto limitation of eBPF as a technology, but a limitation of our current dependencies.

If we would like to amend this, we can go ahead with the other PR which reverts the entire integration and revisit it once upstream packaging for the bcc toolchain is improved.

#### PR Checklist

- [x] CHANGELOG updated
- [x] Tests updated (N/A)

### DIFF
--- a/pkg/integrations/v2/ebpf/integration.go
+++ b/pkg/integrations/v2/ebpf/integration.go
@@ -1,5 +1,5 @@
-//go:build linux && !noebpf
-// +build linux,!noebpf
+//go:build linux && amd64 && !noebpf
+// +build linux,amd64,!noebpf
 
 package ebpf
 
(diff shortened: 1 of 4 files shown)


pkg/integrations/v2/ebpf/integration.go
```diff
@@ -1,5 +1,5 @@
-//go:build linux && !noebpf
-// +build linux,!noebpf
+//go:build linux && amd64 && !noebpf
+// +build linux,amd64,!noebpf
```
mattdurham, line 2: Have we documented that eBPF only works for linux amd64?

mattdurham (approved):
LGTM

brendamuir (approved):

PR #1665, opened 2022-04-29
Revert everything related to ebpf integration

tpaschalis:
#### PR Description
This PR reverts the four commits related to the addition of the eBPF integration.

6b5313a159b9fdfba76f06dec29837b2052f792b
9f03dcf4f001ae2c7cd178b5ddc997428db92a14
bb0b4478a29449ff5ee4923e6c93b09d5a6c3740
2cf47b80a07951bb7dba3aad54223126b0426c09

#### PR Checklist

- [X] CHANGELOG updated
- [X] Documentation added
- [X] Tests updated


captncraig (approved):

iainlane:
I just pushed f2ef6735092d142e2e8bd6e2addece7f025cc1bb to a branch - if/when you bring this back, I think you'll find it helpful.

Basically, the problem you had was that libbpfcc is not available for all architectures the agent is built for, so when buildx uses emulation rather than cross-compilation, as it does for the final stage, then it will fail to find the library if it's not built there. What broke in 2cf47b80a07951bb7dba3aad54223126b0426c09 is that we installed it for _all_ architectures irrespective of whether it was available, and it is not available for `linux/arm/v6` which Debian calls `armel`.

In f2ef6735092d142e2e8bd6e2addece7f025cc1bb I made a change to only install when the build target is linux/amd64 which is the only target eBPF is enabled for anyway.

(aside: There are [upstream docs](https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/) on the cross compilation feature, and [here](https://github.com/grafana/agent/blob/bc42ae89b8d632e2df1c15abf1d39235434aaf55/cmd/agent/Dockerfile.buildx#L1) is where it's used in the agent - note the `--platform=$BUILDPLATFORM` which is what enables cross-compilation.)

PR #1705, opened 2022-05-16
Rewrite all front matter as YAML with current alias defined

jdbaldry:
#### PR Description

The addition of the current alias has no effect on the website as Hugo
will not create an HTML redirect if there is already content in the
web server directory.

The reason for the addition of the alias is to reduce the likelihood
of 404s occuring when files are moved.

Signed-off-by: Jack Baldry <jack.baldry@grafana.com>

#### PR Checklist

- [na] CHANGELOG updated
- [na] Documentation added
- [na] Tests updated

### DIFF
--- a/docs/user/README.md
+++ b/docs/user/README.md
@@ -1,6 +1,8 @@
-+++
-draft = "True"
-+++
+---
+aliases:
+- /docs/agent/latest/README/
+draft: "True"
+---
 
 <p align="center"><img src="assets/logo_and_name.png" alt="Grafana Agent logo"></p>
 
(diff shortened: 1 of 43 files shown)


docs/user/README.md
```diff
@@ -1,6 +1,8 @@
-+++
-draft = "True"
-+++
+---
+aliases:
+- /docs/agent/latest/README/
+draft: "True"
```
jdbaldry, line 4: As far as I can tell, this is working on the website. https://grafana.com/docs/agent/latest/README/ is a 404. I was expecting this to be a boolean value though.

brendamuir (approved):

PR #1708, opened 2022-05-17
Use serviceAccountName instead of deprecated serviceAccount in V2 templates

glindstedt:
Switch to using the non-deprecated `serviceAccountName` instead of `serviceAccount` in V2 templates. The deprecated field produces errors when we lint our kubernetes manifests.


rfratto (approved):
LGTM, thank you!

glindstedt:
@rfratto I assume the test failure was a flake, could you retrigger?

PR #1715, opened 2022-05-18
Remove production/README.md

rfratto:
production/README.md is redocumenting existing things found in the main documentation. Let's remove it so we don't have to keep the two in sync over time.


mattdurham (approved):
LGTM

PR #1723, opened 2022-05-23
cmd/agentflow: introduce temporary agentflow main package for prototyping

rfratto:
This commit introduces cmd/agentflow as a separate binary which allows for developers to test components end-to-end and merge them into the main branch without having to add flow support into cmd/agent itself.

After Flow is ready for wider use, cmd/agentflow will be replaced with native cmd/agent support.


mattdurham (approved):
LGTM

PR #1724, opened 2022-05-23
agentflow: add /debug/graph endpoint to render current dependency DAG

rfratto:
Adds a /debug/graph endpoint to render the currently loaded dependency graph (with transitive reduction applied).


mattdurham (approved):
Kind of hate having to have graphviz installed but good for now!

rfratto:
> Kind of hate having to have graphviz installed but good for now!

Yeah, that's fair. Most Go developers probably have it installed already since you need it for `go tool pprof` to work, but I'd be happy to replace this with a frontend renderer when we have time to build one :)

PR #1727, opened 2022-05-24
Flow: add /-/config endpoint

rfratto:
The `/-/config` endpoint renders the current state of all components, with expressions evaluated, back as HCL to the user. 

`/-/config?debug=1` may be invoked to add health information per-component along with component-specific debug information if a component implements the `component.DebugComponent` interface.


mattdurham (approved):
Lgtm

PR #1729, opened 2022-05-24
flow/hcltypes: add a Secret and OptionalSecret type

### LINKED ISSUE #1688: Flow controller: hide secrets from config endpoint
A `/-/config` HTTP endpoint will be made available to examine the current state of Flow. This endpoint should hide any sensitive values. 

The details of how this should work is up to the discretion of the implementer, though I suggest the following: 

The Flow prototype introduced a [Secret type](https://github.com/grafana/agent/blob/agent-flow/expressions/pkg/flow/hcltypes/secret.go#L16) which can be used by a component's config or exported fields. This type allows any string to be converted to a Secret, but not in the other direction, avoiding accidental secret leakage. 

However, this is only part of the solution: some components may export values which are sometimes-sensitive, sometimes-not. For example, the `remote.s3` component (#1689) may retrieve sensitive info (like an API key) or non-sensitive info (like a config file). 

Such components will need some way to configure if they have secret information so that they export the Secret type instead of a string.

rfratto:
Closes #1688.

Secret
------

hcltypes.Secret allows components to declare a value as sensitive.
Secret types are never shown to the user at the /-/config endpoint,
appearing as `(secret)`.

Both strings and OptionalSecrets may be converted into a Secret.

Secrets may be converted back to OptionalSecrets, but may never be
converted back into a string.

OptionalSecret
--------------

hcltypes.OptionalSecret allows components to configure whether a string
value is sensitive. This is useful for components like local.file where
some files contain sensitive information and others don't.

Both strings and Secrets may be convertd into an OptionalSecret.

OptionalSecrets may be converted into Secrets. OptionalSecrets may only
be converted back into a string if their underlying value is not marked
as sensitive.

### DIFF
--- a/pkg/flow/hcltypes/secret.go
+++ b/pkg/flow/hcltypes/secret.go
@@ -1,0 +1,75 @@
+package hcltypes
+
+import (
+	"reflect"
+
+	"github.com/hashicorp/hcl/v2/hclsyntax"
+	"github.com/hashicorp/hcl/v2/hclwrite"
+	"github.com/rfratto/gohcl"
+	"github.com/zclconf/go-cty/cty"
+)
+
+// Secret holds a sensitive value. Secrets are never displayed to the user when
+// rendering HCL.
+//
+// String values may be implicitly converted to a Secret, but the inverse isn't
+// true: this ensures that it's not possible for a user to accidentally leak a
+// sensitive value.
+type Secret string
+
+var secretTy cty.Type
+
+func init() {
+	secretTy = cty.CapsuleWithOps("secret", reflect.TypeOf(Secret("")), &cty.CapsuleOps{
+		// We allow strings to be converted into secrets, but don't allow secrets
+		// to be converted back to strings to prevent them from being leaked.
+
+		ConversionFrom: func(src cty.Type) func(interface{}, cty.Path) (cty.Value, error) {
+			switch {
+			case src.Equals(optionalSecretTy): // Secret -> OptionalSecret
+				return func(v interface{}, _ cty.Path) (cty.Value, error) {
+					return cty.CapsuleVal(optionalSecretTy, &OptionalSecret{
+						Sensitive: true,
+						Value:     string(*v.(*Secret)),
+					}), nil
+				}
+			default:
+				return nil
+			}
+		},
+
+		ConversionTo: func(dst cty.Type) func(cty.Value, cty.Path) (interface{}, error) {
+			switch {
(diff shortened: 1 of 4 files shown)


pkg/flow/hcltypes/secret.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/rfratto/gohcl"
+	"github.com/zclconf/go-cty/cty"
+)
+
+// Secret holds a sensitive value. Secrets are never displayed to the user when
+// rendering HCL.
+//
+// String values may be implicitly converted to a Secret, but the inverse isn't
+// true: this ensures that it's not possible for a user to accidentally leak a
+// sensitive value.
```
tpaschalis, line 17: I'm not sure what we mean by this. We certainly can do the following

```go
s := Secret("hello world!")
fmt.Println(string(s)) // hello world!
```

This is tied to the general question about how the users/developers will interact with these types. People who write components will make use of Secret/OptionalSecret and the unexported secretTy/optionalSecretTy are to be used when encoding/decoding HCL, right?

tpaschalis (approved):
LGTM! Just trying to see how this will be used when developing new components.

pkg/flow/hcltypes/secret.go
tpaschalis: On second reading, I guess it means that it can't be converted back when referred to as arguments and exports of a Flow component which are go-cty values in the background.

rfratto: > People who write components will make use of Secret/OptionalSecret and the unexported secretTy/optionalSecretTy are to be used when encoding/decoding HCL, right?

Right, component authors will use Secret/OptionalSecret directly and that will be converted to registered cty capsule types under the hood. 

> But will they also handle conversions between these types manually? Or is this also something that's going to happen on the HCL layer as Flow components pass values between each other?

Conversions happen automatically when decoding. A component author could expect a password to be a Secret and they don't have to know whether the final value came from a Secret or a string. 

> On second reading, I guess it means that it can't be converted back when referred to as arguments and exports of a Flow component which are go-cty values in the background.

Yeah, I'm talking about the go-cty conversion being disallowed so users can't convert (but developers can). I'll update the comment to try to make that more clear.

tpaschalis: Thanks, made more sense after a second reading! I played around with the tests a little, and I don't have anything to add ^^

PR #1733, opened 2022-05-25
flow: add support for log level and format in config file

rfratto:


cmd/agentflow/example-config.flow
```diff
@@ -1 +1,5 @@
+log_level  = "debug"
+log_format = "logfmt"
```
rfratto, line 2: Should we have a logging block instead of separate fields?

```hcl 
logging {
  level  = "debug"
  format = "logfmt"
}
```

pkg/flow/flow.go
```diff
@@ -158,6 +169,9 @@ func (c *Flow) run(ctx context.Context) {
 	}
 }
 
+// Logger returns the logger used by the Flow controller.
+func (c *Flow) Logger() log.Logger { return c.log }
```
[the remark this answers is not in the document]
rfratto: We could do something similar to the existing agent, where the logger is maintained higher up, but then it's a little weird for the logging parameters to be defined in the root HCL block at all. 🤔

cmd/agentflow/example-config.flow
mattdurham: The block feels better to me. Also eventually the agent should expose its own logger and logging have an output makes logical sense.

rfratto: Any opinions on the block name? `logging`/`logger`/something else?

mattdurham (approved):
LGTM

PR #1735, opened 2022-05-25
Add Hugo link checking to PR check_docs workflow

jdbaldry:
#### PR Description

Builds the technical documentation in the grafana.com website as part of the PR `check_docs` workflow. Hugo performs link checking of relrefs and refs and the passed environment variable causes this build to fail if any of those links are broken.

#### PR Checklist

- [na] CHANGELOG updated
- [ ] Documentation added
- [na] Tests updated

PR #1745, opened 2022-05-26
Remove old docs check in favor of Hugo check from #1735

rfratto:
Remove the old flaky check in favor of the new check introduced in #1735.


tpaschalis (approved):
I'm not sure if Brenda is on today, but I'm giving it a ✅

PR #1746, opened 2022-05-26
Flow: introduce local.file component

### LINKED ISSUE #1725: Flow: support for reading file contents
Grafana Agent Flow should allow users to read files from disk. This can be used for any string, but most commonly will be used to read passwords or API keys. 

There are two ways this can be implemented: as a component, or as a function. 

### Component approach 

The `local.file` component will read & watch a file from disk, updating its exports whenever the file changes. 

It is crucial that the `local.file` component can be marked as sensitive so that its value does not appear at the `/-/config` endpoint.  

```hcl 
local "file" "some-api-key" {
  path      = "./some-api-key.txt"
  sensitive = true 
}
```

### Function approach 

The `file` function will return the latest contents of a file. The file will be watched for changes, and cause components using it to have their arguments re-evaluated. 

```hcl 
file("./some-api-key.txt") 
```

### Pros / Cons 

Pros for component: 

* Easily implementable as a component 
* Fits in well with the `remote.*` components

Pros for a function: 

* Way less tedious for users to load many secrets from many files, which is likely a common use case. 
* Because the function isn't a separate component, you wouldn't need to mark it as sensitive, and its value will be hidden by default if being assigned to a secret field. 

## Proposed path 

Whatever we pick can be changed down the line. 

I think functions make more sense than components if the user is going to define many of them for one-off use cases. Components should be re-usable; functions should be more limited. I think we're bound to eventually have component-like functions: they encapsulate some logic and can change their resulting value over time.

However, since we are going to have `remote.*` components, it probably makes the most sense to have `local.file` exist as a component too. 

Function-like components can remain as a potential solution for some other problem, like a function for checking ownership of a key within a cluster which is re-evaluated when the cluster changes.

rfratto:
#### PR Description
Introduces a new `local.file` component which can expose the contents of files on disk to other components. 

`local.file` will watch the file for updates and will export the latest content when it changes. 

There are two update detection mechanisms:

* `fsnotify` (the default): Uses `fsnotify` to detect when the file changes 
* `poll`: Polls to check for file modification 

The `poll` mechanism will be required for users who are watching files on filesystem where changed events are not emitted (e.g., NFS mounts).

Closes #1725

#### PR Checklist

- [x] CHANGELOG updated (N/A)
- [x] Tests updated

### DIFF
--- a/component/local/file/detector.go
+++ b/component/local/file/detector.go
@@ -65,10 +65,14 @@ func (ut *Detector) UnmarshalText(text []byte) error {
 }
 
 type fsNotify struct {
-	opts       fsNotifyOptions
-	watcherMut sync.Mutex // Needed to prevent race conditions on Windows
+	opts   fsNotifyOptions
+	cancel context.CancelFunc
+
+	// watcherMut is needed to prevent race conditions on Windows. This can be
+	// removed once fsnotify/fsnotify#454 is merged and included in a patch
+	// release.
+	watcherMut sync.Mutex
 	watcher    *fsnotify.Watcher
-	cancel     context.CancelFunc
 }
 
 type fsNotifyOptions struct {


rfratto:
I found #1747 while testing this, but since it's a bug unrelated to this specific PR I'll leave it for a separate fix.

rfratto:
Looks like the tests are flaky, I'm investigating.

component/local/file/detector.go
```diff
@@ -0,0 +1,204 @@
...
+			// Continually re-adding it to the watcher acts as a fallback polling
+			// mechanism, where we'll eventually rewatch the file once it gets added
+			// again.
+			err := fsn.watcher.Add(fsn.opts.Filename)
+			if err != nil {
+				level.Warn(fsn.opts.Logger).Log("msg", "failed re-watch file", "err", err)
+			} else {
+				fsn.forwardNotification()
+			}
+		case err := <-fsn.watcher.Errors:
```
tpaschalis, line 126: Under what circumstances will we receive errors here? Why do we want to forwardNotification() and read the file here, so that it fails and the output gets cleared?

tpaschalis (approved):
LGTM! Finally, bringing the first 'real' component to life!

component/local/file/detector.go
rfratto: Looks like the errors you could get depends on the platform, but the intent was to force the file to reload so any file-level issues would be reported as a health problem on the component. 

Do you think this should take a different approach or just needs more comments to explain what's going on?

tpaschalis: No, I just wanted to clarify the logic here. Thanks!

rfratto: Gotcha, I'll add comments then :) If it wasn't obvious from your first read, it won't be obvious to others either and they shouldn't have to find the PR to look for an explanation 👍

PR #1751, opened 2022-05-30
Fix rendering of cookbook/dynamic-configuration docs

tpaschalis:

#### PR Description
Currently, the documentation page has some 'phantom' links with no title. These correspond to the [cookbook/dynamic-configuration](https://github.com/grafana/agent/tree/main/docs/user/cookbook/dynamic-configuration) pages. Hugo encounters those Markdown files under the docs/ directory and tries to render them, but doesn't find the necessary headers.

This is a general problem with linking and presenting documentation which would be viewable via _both_ Hugo and Github.

image

This PR includes two new `_index.md` files so that Hugo generates the folder structure under "cookbook/dynamic-configuration", plus headers for each of the relevant pages.

The second _index.md contains the same content as README.md, only difference is that one is built for viewing and working links via Hugo and the other via Github. It also adds a `draft: true` front-matter/header to README.md so it's ignored by Hugo. I'm not super happy about them, but they're one solution. 

image

#### Notes to the Reviewer

* One idea would be to move the cookbook _outside_ of the docs/ directory and have it be only available from the Github repo. But that would defeat its purpose, right? 

* Unfortunately, to make the links to YAML snippets (eg. 04_config.yml) to work correctly both in Github and Hugo, I had to make them hardcoded to the main branch, instead of relative directories 😞  Let me know if you think this is a big deal.

#### PR Checklist

- [ ] CHANGELOG updated (N/A)
- [X] Documentation added
- [ ] Tests updated (N/A)


tpaschalis:
We talked about this in the past, but I totally forgot it. Here's a possible solution.

It's a little unwieldy when we want to have documentation that works both in rendered Hugo _and_ through the Github interfaces so this has some compromises, but let me know how it looks!

tpaschalis:
@mattdurham I don't want to step on your toes, so I'd like to hear your opinion.

How does this look to you? Are you okay with how it tries to work both in Hugo and when viewed from the Github interface, or should we focus on one of those?  
Also, after talking it through with @brendamuir, we could think about grouping this under the "Configure Grafana Agent" section and dropping the "Cookbook" section? WDYT?

mattdurham:
Its less ideal but if it cleans up the grafana.com page then I am fine with it.

mattdurham (approved):
LGTM

tpaschalis:
@mattdurham one last thing. I know you were into the Cookbook idea. Do you think we'll have good use-cases for it in the future and keep it, or do you prefer that we include the "Dynamic Configuration" section under the general "Configure Grafana Agent" list?

brendamuir:
Can we also adjust the titles so it doesn't appear as Basics/Structure etc. ?

docs/user/cookbook/_index.md
```diff
@@ -0,0 +1,15 @@
+---
+aliases:
+- /docs/agent/latest/cookbook
+title: Cookbook
+weight: 900
+---
+
+## Cookbook 
+
+The cookbook section aims to provide small guides for more advanced use cases, experimental features, or certain concepts around the Grafana Agent.
```
brendamuir, line 10:
```suggestion
The cookbook section aims to provide small guides for more advanced use cases, experimental features, or certain concepts around Grafana Agent.
```

brendamuir: Apparantly we should just say "Grafana Agent" - no articles needed

brendamuir (commented):
Looks better

mattdurham (approved):
LGTM

PR #1752, opened 2022-05-30
Add --platforms flag to agentctl cloud-config command

tpaschalis:

#### PR Description
This PR adds a `--platforms` flag to the `agentctl cloud-config` command. 

Last quarter, we enhanced our cloud API to receive filters based on platforms and include/exclude options. This enabled us to have both Linux and macOS integrations based on `node_exporter` and avoiding errors about duplicate keys; the Grafana Cloud API would be smart enough to request platform-specific configuration bring only one of the two keys.

Currently, when new users go through onboarding, they're using the `grafanacloud-install.sh` script to retrieve their Cloud configuration locally. If they have both the Linux and macOS integrations, calling the cloud-config command will return a 500 error with a message along the lines of 
```msg="could not unmarshal agent config" err="yaml: unmarshal errors:\n  line XX: mapping key \"node_exporter\" already defined at line XX"```

This new flag is instructs the Linux-specific script mentioned above to only request configuration for Linux-based integrations and ignore everything else.

No issue filed.

#### Notes to the Reviewer
Should we also add flags for the `include` and `exclude` filters in the Grafana Cloud API? Also, if you think there's a cleaner way of achieving the same result, please let me know.

#### PR Checklist

- [ ] CHANGELOG updated
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ Main (unreleased)
 
 - Configure the agent to report the use of feature flags to grafana.com. (@marctc)
 
+- Add `--platforms` flag to agentctl cloud-config command. (@tpaschalis)
 
 ### Enhancements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,8 +36,6 @@ Main (unreleased)
 
 - Configure the agent to report the use of feature flags to grafana.com. (@marctc)
 
-- Add `--platforms` flag to agentctl cloud-config command. (@tpaschalis)
-
 ### Enhancements
 
 - integrations-next: Integrations using autoscrape will now autoscrape metrics
--- a/cmd/agentctl/main.go
+++ b/cmd/agentctl/main.go
@@ -431,9 +431,10 @@ func filterAgentOwners(refs []meta_v1.OwnerReference) (filtered []meta_v1.OwnerR
 
 func cloudConfigCmd() *cobra.Command {
 	var (
-		stackID string
-		apiURL  string
-		apiKey  string
+		stackID   string
+		apiURL    string
+		apiKey    string
+		platforms string
 	)
 
 	cmd := &cobra.Command{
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Review by v-zhuravlev — APPROVED** [review]
LGTM, thank you!

**Comment by rfratto** [comment]
Determining it dynamically sounds nice, but for local testing I like the flag. Maybe runtime.GOOS can determine the default value for `--platforms`, but still allow for overriding?

File: cmd/agentctl/main.go [hunk 1]
```diff
@@ -474,6 +475,7 @@ config that may be used with this agent.`,
 	cmd.Flags().StringVarP(&stackID, "stack", "u", "", "stack ID to get a config for")
 	cmd.Flags().StringVarP(&apiKey, "api-key", "p", "", "API key to authenticate against Grafana Cloud's API with")
 	cmd.Flags().StringVarP(&apiURL, "api-url", "e", "", "Grafana Cloud's API url")
+	cmd.Flags().StringVar(&platforms, "platforms", "", "Platforms/OSes to retrieve configuration for")
```
rfratto, line 478: Pluralization feels a little weird here. Is it valid to provide multiple platforms? If so, how does the user provide more than one? Comma-delimited?

CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ Main (unreleased)
 
 - Configure the agent to report the use of feature flags to grafana.com. (@marctc)
 
+- Add `--platforms` flag to agentctl cloud-config command. (@tpaschalis)
```
rfratto, line 39: Heads up: `agentctl cloud-config` is left undocumented so we tend to not mention it in the changelog. Up to you if you want to mention it, but I've historically left it as an internal detail of our cloud installation process so that the changelog and other things can be OSS focused.

rfratto (commented):
I'm OK with this, but since Matt raised a concern I'll leave this for him to approve.

CHANGELOG.md
tpaschalis: You're right, makes more sense as an internal detail. Removing the changelog entry.

cmd/agentctl/main.go
rfratto: We can add it later if necessary. For now, I think changing the help text of the flag to note that the platforms are comma-delimited would be good 👍

tpaschalis:
@mattdurham I've gone with the intermediate solution. The default value of the flag gets set by runtime.GOOS, so you don't _have_ to pass it, but it still allows for overriding.

No strong opinion on either way of going ahead with it, let me know what you think :)

mattdurham (approved):
LGTM

Issue #1721, opened 2022-05-23
Traces: Enable including client metadata be default


mapno:
Collecting metadata from incoming requests in OTLP is hidden behind an experimental config option `include_metadata` ([http](https://github.com/open-telemetry/opentelemetry-collector/blob/v0.46.0/config/confighttp/confighttp.go#L220-L222) & [grpc](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configgrpc/configgrpc.go#L157-L159)).

Adding this metadata to requests is extremely useful for the PromSD processor, which relies on getting the IP of the batch sender in order to attach attributes to traces. One of the methods it uses for getting the IP is inspecting the context received, with the hope that it contains metadata from the connection from where the batch came from. This is rarely the case since it's disabled by default.

PR #1771, opened 2022-06-03
Set include_metadata to true by default for OTLP traces receivers

mapno:
#### PR Description

Set include_metadata to true by default for OTLP traces receivers.

Adding this metadata to requests is extremely useful for the PromSD processor, which relies on getting the IP of the batch sender in order to attach attributes to traces. One of the methods it uses for getting the IP is inspecting the context received, with the hope that it contains metadata from the connection from where the batch came from. This is rarely the case since it's disabled by default.

Fixes https://github.com/grafana/agent/issues/1721

#### PR Checklist

- [x] CHANGELOG updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/traces/config.go
+++ b/pkg/traces/config.go
@@ -156,13 +156,18 @@ func (r *ReceiverMap) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	for k := range *r {
 		if strings.HasPrefix(k, otlpReceiverName) {
 			// for http and grpc receivers, include_metadata is set to true by default
+			protocolsCfg, ok := (*r)[k].(map[interface{}]interface{})["protocols"].(map[interface{}]interface{})
+			if !ok {
+				return fmt.Errorf("failed to parse OTLP receiver config: %s", k)
+			}
+
 			for _, p := range protocols {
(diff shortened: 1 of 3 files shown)


pkg/traces/config.go
```diff
@@ -141,6 +144,35 @@ type InstanceConfig struct {
...
+		return err
+	}
+
+	protocols := []string{protocolHTTP, protocolGRPC}
+	// enable include_metadata by default if receiver is OTLP
+	for k := range *r {
+		if strings.HasPrefix(k, otlpReceiverName) {
+			// for http and grpc receivers, include_metadata is set to true by default
+			for _, p := range protocols {
+				if cfg, ok := (*r)[k].(map[interface{}]interface{})["protocols"].(map[interface{}]interface{})[p]; ok {
```
joe-elliott, line 160: this is quite the line. let's break it up for clarity. this code panics if `(*r)[k]` is not a `(map[interface{}]interface{})`?

let's do these casts once and store them in variables to be used later. will help with code clarity.

joe-elliott (commented):
I fully support changing this default if it helps people more easily be successful with the promdsdprocessor.

pkg/traces/config.go
mapno: Simplified casting. Done!

PR #1780, opened 2022-06-06
flow: fail if component returns error while building

### LINKED ISSUE #1747: Flow controller: component errors do not prevent initial load
While testing #1746, I noticed that purposefully providing an invalid config file did not prevent the controller from running:

```hcl
local "file" "bad-file" {
  filename = "this/file/does/not/exist"
}
```

Logs:

```
level=error component=local.file.this-file msg="failed to read file" path=data-agent/local.file.bad-file err="open this/file/does/not/exist: no such file or directory"
level=error msg="failed to evaluate component" component=local.file.bad-file err="building component: no diagnostics"
level=info msg="scheduling loaded components"
level=info msg="now listening for http traffic" addr=127.0.0.1:12345
```

I expect the first load of the file to fail and exit the process if any component fails to evaluate, but it looks like:

* The failed evaluation isn't propagating the error back up to the caller 
* The error from the component isn't being returned by ComponentNode

captncraig:
Fixes #1747 

Error could potentially be passed up another layer or 2, but I thought this was a sensible place to exit without modifying too many more signatures.

### DIFF
--- a/pkg/flow/internal/controller/component.go
+++ b/pkg/flow/internal/controller/component.go
@@ -233,7 +233,7 @@ func (cn *ComponentNode) evaluate(ectx *hcl.EvalContext) error {
 		// We haven't built the managed component successfully yet.
 		managed, err := cn.reg.Build(cn.managedOpts, argsCopy)
 		if err != nil {
-			return fmt.Errorf("building component: %w", diags)
+			return fmt.Errorf("building component: %s", err)
 		}
 		cn.managed = managed
 		cn.args = argsCopy
--- a/pkg/flow/internal/controller/component.go
+++ b/pkg/flow/internal/controller/component.go
@@ -233,7 +233,7 @@ func (cn *ComponentNode) evaluate(ectx *hcl.EvalContext) error {
 		// We haven't built the managed component successfully yet.
 		managed, err := cn.reg.Build(cn.managedOpts, argsCopy)
 		if err != nil {
-			return fmt.Errorf("building component: %s", err)
+			return fmt.Errorf("building component: %w", err)
 		}
 		cn.managed = managed
 		cn.args = argsCopy
--- a/pkg/flow/internal/controller/loader.go
+++ b/pkg/flow/internal/controller/loader.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"sync"
 
+	stdlog "log"
+
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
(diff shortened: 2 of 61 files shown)


pkg/flow/internal/controller/loader.go
```diff
@@ -81,7 +83,9 @@ func (l *Loader) Apply(parentContext *hcl.EvalContext, blocks hcl.Blocks) hcl.Di
 		// We cache both arguments and exports during an initial load in case the
 		// component is new; we want to make sure that all fields are available
 		// before the component updates its exports for the first time.
-		l.evaluate(parentContext, n.(*ComponentNode), true, true)
+		if err := l.evaluate(parentContext, n.(*ComponentNode), true, true); err != nil {
+			stdlog.Fatal(err)
```
rfratto, line 87: We don't want this to always exit the process; this will get called every time the file gets loaded, even after the first load. 

Can this instead add an error-level diagnostic to report errors back to the caller? 

(Also, we already have a logger attached to the Loader which isn't stdlog)

pkg/flow/internal/controller/component.go
```diff
@@ -233,7 +233,7 @@ func (cn *ComponentNode) evaluate(ectx *hcl.EvalContext) error {
 		// We haven't built the managed component successfully yet.
 		managed, err := cn.reg.Build(cn.managedOpts, argsCopy)
 		if err != nil {
-			return fmt.Errorf("building component: %w", diags)
+			return fmt.Errorf("building component: %s", err)
```
rfratto, line 236: This should still wrap the error with `%w`, otherwise the inner error type information gets lost.

pkg/flow/internal/controller/loader.go
```diff
@@ -202,23 +209,23 @@ func (l *Loader) EvaluateDependencies(parentContext *hcl.EvalContext, c *Compone
 			// arguments will need re-evaluation.
 			return nil
 		}
-		l.evaluate(parentContext, n.(*ComponentNode), true, false)
-		return nil
+		return l.evaluate(parentContext, n.(*ComponentNode), true, false)
```
rfratto, line 212: We need to put some thought into whether this is correct or not. Anything returning an error during Walk will cause the Walk to abort. This means that if an evaluation fails, any dependency won't be re-evaluated, even if some of its expressions would've properly changed values. 

We could just ignore the evaluation here too; at this point we're probably more interested in whether errors cause a component to go unhealthy than acting on a failed eval.

PR #1783, opened 2022-06-08
Add component to transform Service Discovery targets

### LINKED ISSUE #1683: Flow component for mutating targets
This is the counterpart to #1680; there should be a component in Flow which can mutate discovered targets, dropping specific targets or changing their labels. 

Like #1680, for the initial implementation, it is recommended that the configuration for this component matches Prometheus relabeling rules, where you can provide multiple relabel configs in the component's config block.

The name of this component also needs some thought, as an exercise to the implementor.

tpaschalis:

#### PR Description

This is an early draft of the PR that will introduce a component that can mutate/modify/edit/transform scrape targets coming from a variety of service discovery mechanisms.

An early test, with the following configuration
```
discovery "transformer" "one" {
  targets = [
    { "__address__" = "localhost:9091", "name" = "one"},
    { "__address__" = "localhost:9092", "name" = "two"},
  ]

  scrape_labels = {
    job_name = "my_scrape_job"
    interval = "1m"
    timeout = "59s"
    metrics_path = "/metrics"
    scheme = "https"
  }

  query_params = {"user" = ["tpaschalis"] }

  relabel_configs  {
	source_labels = ["__address__", "name"]
	separator = "/"
	target_label = "destination"
        action = "replace"
        regex = "(.*)"
  }

  relabel_configs  {
	source_labels = ["name"]
	target_label = "instance"
  }
}
```

results in the original target set to be transformed into the following label set
```
  // Exported fields:
  output_targets = [{
    __address__         = "localhost:9091"
    __metrics_path__    = "/metrics"
    __param_user        = "tpaschalis"
    __scheme__          = "https"
    __scrape_interval__ = "1m"
    __scrape_timeout__  = "59s"
    destination         = "localhost:9091/one"
    instance            = "one"
    job                 = "my_scrape_job"
    name                = "one"
  }, {
    __address__         = "localhost:9092"
    __metrics_path__    = "/metrics"
    __param_user        = "tpaschalis"
    __scheme__          = "https"
    __scrape_interval__ = "1m"
    __scrape_timeout__  = "59s"
    destination         = "localhost:9092/two"
    instance            = "two"
    job                 = "my_scrape_job"
    name                = "two"
  }]
```

Fixes #1683 

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/component/discovery/transformer/transformer.go
+++ b/component/discovery/transformer/transformer.go
@@ -73,7 +73,7 @@ type ScrapeLabels struct {
 var DefaultRelabelConfig = RelabelConfig{
 	Action:      Replace,
 	Separator:   ";",
-	Regex:       MustNewRegexp("(.*)"),
+	Regex:       mustNewRegexp("(.*)"),
 	Replacement: "$1",
 }
 
@@ -241,7 +241,7 @@ func hclToPromRelabelConfigs(rcs []*RelabelConfig) []*relabel.Config {
 			TargetLabel:  rc.TargetLabel,
 			Replacement:  rc.Replacement,
 			Action:       relabel.Action(rc.Action),
-			Regex:        relabel.Regexp{Regexp: rc.Regex.re}, // TODO (@tpaschalis) not super happy with how this turned out, let's check it again.
+			Regex:        relabel.Regexp{Regexp: rc.Regex.Regexp}, // TODO (@tpaschalis) not super happy with how this turned out, let's check it again.
 		}
 	}
 
--- a/component/discovery/transformer/transformer.go
+++ b/component/discovery/transformer/transformer.go
@@ -32,7 +32,7 @@ type Arguments struct {
 	Targets []Target `hcl:"targets,optional"`
 
 	// The relabelling steps to apply to the each target's label set.
-	RelabelConfigs []*RelabelConfig `hcl:"relabel_configs,block"`
+	RelabelConfigs []*RelabelConfig `hcl:"relabel_config,block"`
 }
 
 // Target refers to a singular HTTP or HTTPS endpoint that will be used for scraping.
--- a/component/discovery/transformer/transformer.go
+++ b/component/discovery/transformer/transformer.go
@@ -2,6 +2,7 @@ package transformer
 
 import (
 	"context"
+	"fmt"
 	"net/url"
 	"sync"
 	"time"
@@ -9,7 +10,6 @@ import (
 	"github.com/grafana/agent/component"
 	"github.com/grafana/regexp"
 	"github.com/hashicorp/hcl/v2"
-	"github.com/pkg/errors"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/config"
 	"github.com/prometheus/prometheus/model/labels"
@@ -91,25 +91,25 @@ func (rc *RelabelConfig) DecodeHCL(body hcl.Body, ctx *hcl.EvalContext) error {
 	}
 
 	if rc.Action == "" {
-		return errors.Errorf("relabel action cannot be empty")
+		return fmt.Errorf("relabel action cannot be empty")
 	}
 	if rc.Modulus == 0 && rc.Action == HashMod {
-		return errors.Errorf("relabel configuration for hashmod requires non-zero modulus")
+		return fmt.Errorf("relabel configuration for hashmod requires non-zero modulus")
 	}
 	if (rc.Action == Replace || rc.Action == HashMod || rc.Action == Lowercase || rc.Action == Uppercase) && rc.TargetLabel == "" {
-		return errors.Errorf("relabel configuration for %s action requires 'target_label' value", rc.Action)
+		return fmt.Errorf("relabel configuration for %s action requires 'target_label' value", rc.Action)
 	}
 	if (rc.Action == Replace || rc.Action == Lowercase || rc.Action == Uppercase) && !relabelTarget.MatchString(rc.TargetLabel) {
-		return errors.Errorf("%q is invalid 'target_label' for %s action", rc.TargetLabel, rc.Action)
+		return fmt.Errorf("%q is invalid 'target_label' for %s action", rc.TargetLabel, rc.Action)
 	}
 	if (rc.Action == Lowercase || rc.Action == Uppercase) && rc.Replacement != DefaultRelabelConfig.Replacement {
-		return errors.Errorf("'replacement' can not be set for %s action", rc.Action)
+		return fmt.Errorf("'replacement' can not be set for %s action", rc.Action)
 	}
 	if rc.Action == LabelMap && !relabelTarget.MatchString(rc.Replacement) {
-		return errors.Errorf("%q is invalid 'replacement' for %s action", rc.Replacement, rc.Action)
+		return fmt.Errorf("%q is invalid 'replacement' for %s action", rc.Replacement, rc.Action)
 	}
 	if rc.Action == HashMod && !model.LabelName(rc.TargetLabel).IsValid() {
-		return errors.Errorf("%q is invalid 'target_label' for %s action", rc.TargetLabel, rc.Action)
+		return fmt.Errorf("%q is invalid 'target_label' for %s action", rc.TargetLabel, rc.Action)
 	}
 
 	if rc.Action == LabelDrop || rc.Action == LabelKeep {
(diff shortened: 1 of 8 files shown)


component/discovery/transformer/transformer.go
```diff
@@ -0,0 +1,265 @@
...
+
+// Arguments holds values which are used to configure the discovery.transformer component.
+type Arguments struct {
+	// Targets contains the input 'targets' passed by a Service Discovery component.
+	Targets []Target `hcl:"targets,optional"`
+
+	// The following values can be used to modify the original target label set.
+	RelabelConfigs []*RelabelConfig `hcl:"relabel_configs,block"`
+	ScrapeLabels   ScrapeLabels     `hcl:"scrape_labels,optional"`
+	QueryParams    url.Values       `hcl:"query_params,optional"`
```
rfratto, line 41: I'm a little confused about what purpose these serve for transforming discovered targets. They both seem relevant to specifically metrics scraping and not just general discovery; should they be removed from here?

component/discovery/transformer/transformer.go
```diff
@@ -0,0 +1,265 @@
...
+	c.healthMut.RLock()
+	defer c.healthMut.RUnlock()
+	return c.health
+}
+
+func (c *Component) setHealth(h component.Health) {
+	c.healthMut.Lock()
+	defer c.healthMut.Unlock()
+	c.health = h
+}
```
rfratto, line 225: I think you can remove this; the controller will update the health of the component when evaluations finish, and component-specific health can be reserved for runtime specific problems (which this component wouldn't have).

component/discovery/transformer/transformer_test.go
```diff
@@ -0,0 +1 @@
+package transformer
```
rfratto, line 1: Are you planning on adding tests? If not, can we remove this?

component/discovery/transformer/types.go
```diff
@@ -0,0 +1,107 @@
...
+type Action string
+
+// Regexp encapsulates the Regexp type from Grafana's
+// fork of the Go stdlib regexp package.
+// TODO (@tpaschalis) This encapsulation already exists in Prometheus' relabel.Regexp
+// so not sure whether to also move it here for now.
+type Regexp struct {
+	re       *regexp.Regexp
+	original string
+}
```
rfratto, line 19: nit: move the declarative for Regexp after all the Action-related things so the file reads more naturally from top down.

component/discovery/transformer/types.go
```diff
@@ -0,0 +1,107 @@
...
+	return []byte(a.String()), nil
+}
+
+// UnmarshalText implements encoding.TextUnmarshaler for Action.
+func (a *Action) UnmarshalText(text []byte) error {
+	if _, exists := actions[Action(text)]; exists {
+		*a = Action(text)
+		return nil
+	}
+	return fmt.Errorf("unrecognized action type %s", string(text))
```
rfratto, line 65: nit: use `%q` for strings so accidental whitespace is more obvious:

```
unrecognized action type  replace
```

vs 

```
unrecognized action type " replace"
```

component/discovery/transformer/types.go
rfratto, line 19: I think this can be `type Regexp *regexp.Regexp`. It's also not necessary to keep the original string; you can just call [`regexp/Regexp.String`](https://pkg.go.dev/regexp#Regexp.String) if you need the original input.

component/discovery/transformer/types.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// MustNewRegexp works like NewRegexp, but panics if the regular expression does not compile.
+func MustNewRegexp(s string) Regexp {
+	re, err := NewRegexp(s)
+	if err != nil {
+		panic(err)
+	}
+	return re
+}
```
rfratto, line 85: Can we unexport these?

component/discovery/transformer/transformer.go
```diff
@@ -0,0 +1,265 @@
...
+	QueryParams    url.Values       `hcl:"query_params,optional"`
+}
+
+// Target refers to a singular HTTP or HTTPS endpoint that will be used for scraping.
+// It mimics github.com/prometheus/prometheus/scrape.Target.
+// Here, we're using a map[string]string instead of labels.Labels; if the label ordering
+// is important, we can change to follow the upstream logic instead.
+// TODO (@tpaschalis) Maybe the target definitions should be part of the
+// Service Discovery components package. Let's reconsider once it's ready.
+type Target map[string]string
```
rfratto, line 50: I think there's two types of Targets we will be dealing with: 


1. Service Discovery targets contain a label set which identify some identified service. They may have common labels like `__address__`. 
2. Scraping targets encapsulate a service discovery target and add extra information like query parameters. 

I'm not sure we need to be dealing with the latter, since the scrape targets are an internal concept to the scrape manager (and are only exposed for the scrape targets API). 

We can likely remove all code related to scraping and just focus on relabeling target sets. WDYT?

rfratto (commented):
First pass looks good :)

component/discovery/transformer/types.go
tpaschalis: Nice detail, gonna use it more often!

component/discovery/transformer/transformer.go
tpaschalis: Oh, I see! In any case, we should leave `CurrentHealth()` intact so that the component implements the HealthComponent interface, right?

component/discovery/transformer/types.go
tpaschalis: I tried using `type Regexp *regexp.Regexp` but turns out we can't define methods on pointer types.

We can either use `type Regexp regexp.Regexp` (without the pointer), which involves some back-and-forth casting or just embedding it in a struct a-la
```
type Regexp struct {
	*regexp.Regexp
}
```

Does that look okay to you?

component/discovery/transformer/transformer.go
tpaschalis: Yeah, I agree. To provide some context, initially the extra scrape labels and query params were to be able to call the Prometheus `scrape.PopulateLabels` method. 

Since this is a higher-level component, it makes sense to only deal with the discovered label set and keep it generic enough; these scrape-specific details can be handled by a downstream component. I'll make this change and get back to you.

tpaschalis: To provide more information: a list of things that [scrape.PopulateLabels](https://github.com/prometheus/prometheus/blob/867d3bd78f514ffffcfd5da751cb2bd84c6aba72/scrape/target.go#L354) does when dealing with targets:

* Override scrape-specific parameters if they weren't set
* Encode query parameters as labels
* Drop the target if no __address__ is set
* Perform all relabelling steps
* Check if we should add a default port to the address
* Check that the address doesn't contain any url paths (no '/')
* Set scrape intervals and scrape timeout, ensure the timeout is no longer than the interval
* If the instace label isn't set, default it to the value of __address__
* Remove `__meta*` labels
* Checks if all label values are valid UTF8 strings

I've simplified the logic keeping only two parts. Applying the relabelling rules, and dropping the target if there's no __address__ or if a label is not a valid UTF8 string. Is that more in line with what you were thinking?

component/discovery/transformer/types.go
rfratto: Either one is fine IMO, but embedding a pointer might be potentially more prone to developer error.

component/discovery/transformer/transformer.go
rfratto: I think you only need to call relabel.Process, and leave any processing of the labels alone :) 

All the rules you mentioned seem specific to the implementation of the scraper, and so we'd probably want to leave that to the scraper to handle.

component/discovery/transformer/types.go
tpaschalis: Also, this can be improved upstream, one less string to keep around :) I've opened https://github.com/prometheus/prometheus/pull/10846, I'll check if we can generate a benchmark for it.

component/discovery/transformer/transformer.go
tpaschalis: Gotcha! Changed the implementation to only run relabelling rules and just drop the target if the result labelset is empty.

component/discovery/transformer/transformer.go
```diff
@@ -0,0 +1,212 @@
...
+	return nil
+}
+
+// Update implements component.Component.
+func (c *Component) Update(args component.Arguments) error {
+	newArgs := args.(Arguments)
+
+	c.mut.Lock()
+	defer c.mut.Unlock()
+	c.args = newArgs
```
rfratto, line 154: c.args is never read, so it looks like you can remove both the mutex and that field :)

rfratto (approved):
LGTM, merge at will! (I know you were planning on adding unit tests, so you can merge after you've added those and once you're happy with them)

component/discovery/transformer/transformer.go
tpaschalis: You mean that we don't ever need it asynchronously so we can just use the component.Arguments passed in the Update() function, right?

tpaschalis: Nice catch, thanks!

PR #1789, opened 2022-06-10
flow: add graph cycles validation

### LINKED ISSUE #1687: Flow controller: reject cyclic dependencies
The prototype Flow controller [lacked](https://github.com/grafana/agent/blob/513c16a2aa8ddd07c4841fca8dd71953062b1a94/pkg/flow/flow.go#L63) any kind of validation logic to ensure that components do not directly or indirectly depend on themselves. 

The DAG used by the Flow controller must be updated to detect [strongly connected components](https://en.wikipedia.org/wiki/Strongly_connected_component) and reject configs where cyclic dependencies exist. 

Rejected configs should inform the user about where the cycle is so they can identify and fix the problem.

marctc:
#### PR Description

This PR adds a validation method which detect cycles for the defined
Flow graph using [strongly connected components algorithm](https://en.wikipedia.org/wiki/Strongly_connected_component).

The detect cycles are returned as `diags` in order to provide debug
info for users which declared their graphs with cycles.

#1687 

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/flow/internal/dag/tarjan.go
+++ b/pkg/flow/internal/dag/tarjan.go
@@ -59,7 +59,6 @@ func (t *tarjan) tarjan(g *Graph, n Node) {
 		// Add current strongly connected component to result
 		t.result = append(t.result, scc)
 	}
-
 }
 
 // visit marks node as visited and pushes to the stack


rlankfo (approved):
LGTM 👍  nice work @marctc

PR #1790, opened 2022-06-10
docs/flow: Document component transforming sd targets

### LINKED ISSUE #1683: Flow component for mutating targets

tpaschalis:

#### PR Description
This PR adds documentation for the recently added `discovery.transformer` component. It also renames it to **targets.mutate**.

~If we have reservations about the name, this might be a good place to kickstart that discussion.~

Fixes #1683

#### Notes to the Reviewer

Do you think we're okay to close #1683 after this?

#### PR Checklist

- [ ] CHANGELOG updated (N/A for now, until we tell people what Flow is about)
- [X] Documentation added
- [ ] Tests updated (N/A)

### DIFF
--- a/docs/flow/components/discovery.transformer.md
+++ b/docs/flow/components/discovery.transformer.md
@@ -1,0 +1,116 @@
+# discovery.transformer
+
+The `discovery.transformer` component rewrites the label set of the input
+targets by applying one or more [`relabel_config`][] steps. If no relabeling
+steps are defined, then the input targets will be exported as-is.
+
+The most common use of `discovery.transformer` is to filter Prometheus targets
+or standardize the label set that will be passed to a downstream component.
+The `relabel_config` blocks will be applied to the label set of each target in
+order of their appearance in the configuration file.
+
+Multiple `discovery.transformer` components can be specified by giving them
+different name labels like "keep-backend-only" in the following example.
+
+## Example
+
+```hcl
+discovery "transformer" "keep-backend-only" {
+  targets = [ 
+    { "__meta_foo" = "foo", "__address__" = "localhost", "instance" = "one",   "app" = "backend"  },
+    { "__meta_bar" = "bar", "__address__" = "localhost", "instance" = "two",   "app" = "database" },
+    { "__meta_baz" = "baz", "__address__" = "localhost", "instance" = "three", "app" = "frontend" }
+  ]
+  
+  relabel_config {
+    source_labels = ["__address__", "instance"]
+    separator     = "/"
+    target_label  = "destination"
+    action        = "replace"
+  } 
+  
+  relabel_config {
+    source_labels = ["app"]
+    action = "keep"
+    regex  = "backend"
+  }
+}
+```
+
+## Arguments
+
+The following arguments are supported and can be referenced by other
+components:
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`targets` | `[]map[string]string` | The input targets | | no
+`relabel_config` | `RelabelConfig` | The relabeling steps to apply | `DefaultRelabelConfig` | no
+
+
+### Targets
+The `targets` field contains the array of input targets. Each target consists
+of one or more string key-value pairs, referred to as its label set.
+
+### RelabelConfig
+The `relabel_config` block contains the definition of any relabeling rules
+that can be applied to an input target. If you want an overview of how
+relabeling works you can refer to the [Prometheus docs][] or this [blogpost][].
+
+```go
+type RelabelConfig struct {
+	SourceLabels []string `hcl:"source_labels,optional"`
+	Separator    string   `hcl:"separator,optional"`
+	Regex        Regexp   `hcl:"regex,optional"`
+	Modulus      uint64   `hcl:"modulus,optional"`
+	TargetLabel  string   `hcl:"target_label,optional"`
+	Replacement  string   `hcl:"replacement,optional"`
+	Action       Action   `hcl:"action,optional"`
+}
+```
+
+### DefaultRelabelConfig
+The `DefaultRelabelConfig` object contains the default values for omitted fields
+when decoding an HCL block to a RelabelConfig struct.
+
+```go
+var DefaultRelabelConfig = RelabelConfig{
+	Action:      Replace,
+	Separator:   ";",
+	Regex:       "(.*)",
+	Replacement: "$1",
+}
+```
+
+## Exported fields
+
+The following fields are exported and can be referenced by other components:
+
+Name | Type | Description
+---- | ---- | -----------
+`output_targets` | `[]map[string]string` | The resulting targets' label set.
+
+The number of targets exposed in `output_targets` will be less or equal than
+the number of input targets, as some of them may be dropped by the
+relabeling process.
+
+## Component health
+
+Any `discovery.transformer` component will be reported as healthy whenever
+it is able to correctly apply all relabeling steps to the input targets.
+If the either the input `targets`, or `relabel_config` blocks could not
+be parsed or the relabeling steps could not be applied, the component will
+be reported as unhealthy. In those cases, exported fields will be kept at
+the last healthy values.
+
+## Debug information
+
+`discovery.transformer` does not expose any component-specific debug information.
+
+### Debug metrics
+
+`discovery.transformer` does not expose any component-specific debug metrics.
+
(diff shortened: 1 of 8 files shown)


tpaschalis:
I'm not sure how we want to document custom object/structs. 

I'll take a look at Hashicorp's documentation to see how they handle Terraform modules and act accordingly.

marctc:
> 
> This PR adds documentation for the recently added `discovery.transformer` component.
> 
> If we have reservations about the name, this might be a good place to kickstart that discussion.

I'm not fully convinced about the name either. It feels that the `discovery` module would enable
developing other components to discover other resources than services. The name `sd` is not
very comprehensive either unless you now the context of Prometheus very well. I'd go for
`service-discovery`, it's long but it's descriptive.

> I'm not sure how we want to document custom object/structs.
> 
> I'll take a look at Hashicorp's documentation to see how they handle Terraform modules and act accordingly.

I don't recall seeing in any terraform's documentation page any reference of Go. I wouldn't add
those to documentation regardless. It feels more like an implementation detail that Flow users shouldn't
care.

docs/flow/components/discovery.transformer.md
```diff
@@ -0,0 +1,104 @@
...
+```
+
+## Arguments
+
+The following arguments are supported and can be referenced by other
+components:
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`targets` | `[]map[string]string` | The input targets | | no
```
rfratto, line 47: Hmm, shouldn't targets be required though? Why have this component if you're not providing targets?

rfratto, line 47: Let's use the type definitions Terraform uses here: `list(map(string))`. We'll need to revisit later to see if this is the type syntax we want to use long term, but something other than the Go type makes sense to me.

docs/flow/components/discovery.transformer.md
```diff
@@ -0,0 +1,104 @@
...
+### Targets
+The `targets` field contains the array of input targets. Each target consists
+of one or more string key-value pairs, referred to as its label set.
+
+### RelabelConfig
+The `relabel_config` block contains the definition of any relabeling rules
+that can be applied to an input target. If you want an overview of how
+relabeling works you can refer to the [Prometheus docs][] or this [blogpost][].
+
+#### Argument Reference
```
rfratto, line 60: Drop the header here so there's only one "argument" section in the docs.

docs/flow/components/discovery.transformer.md
```diff
@@ -0,0 +1,104 @@
...
+* `replacement` - The value against which a regex replace is performed, if the regex matched the extracted value. Supports previously captured groups. Defaults to `$1`.
+* `action` - The relabeling action to perform. Defaults to "replace".
+
+## Exported fields
+
+The following fields are exported and can be referenced by other components:
+
+Name | Type | Description
+---- | ---- | -----------
+`output_targets` | `[]map[string]string` | The resulting targets' label set.
```
rfratto, line 79: nit: like the arguments, I think this should also be `list(map(string))`.

rfratto, line 79: nit: For the description, I'd like it to be a little more descriptive: 

> The set of targets after applying transformations.

docs/flow/components/discovery.transformer.md
```diff
@@ -0,0 +1,104 @@
...
+
+The number of targets exposed in `output_targets` will be less or equal than
+the number of input targets, as some of them may be dropped by the
+relabeling process.
```
rfratto, line 83: It could be more too, if you use label mapping. I think we can just drop this paragraph.

docs/flow/components/discovery.transformer.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Component health
+
+Any `discovery.transformer` component will be reported as healthy whenever
+it is able to correctly apply all relabeling steps to the input targets.
+If the either the input `targets`, or `relabel_config` blocks could not
+be parsed or the relabeling steps could not be applied, the component will
+be reported as unhealthy. In those cases, exported fields will be kept at
+the last healthy values.
```
rfratto, line 92: I think we can shorten this quite a bit:

> The `discovery.transformer` component will only be reported as unhealthy when given an invalid configuration.

docs/flow/components/discovery.transformer.md
```diff
@@ -0,0 +1,104 @@
...
+    target_label  = "destination"
+    action        = "replace"
+  } 
+  
+  relabel_config {
+    source_labels = ["app"]
+    action = "keep"
+    regex  = "backend"
+  }
+}
```
rfratto, line 37: This should to be reformatted to conform to the HCL formatting rules:

```suggestion
  relabel_config {
    source_labels = ["app"]
    action        = "keep"
    regex         = "backend"
  }
}
```

docs/flow/components/discovery.transformer.md
tpaschalis: I'm okay with dropping the paragraph, but can they actually be more? I thought that labelmap will just rename labels from a specific target.

docs/flow/components/discovery.transformer.md
rfratto: Oh, sorry, required means that the field has to be set by the user, even if it's the zero value. i.e., for required targets, this is still valid:

```hcl 
targets = []
```

but omitting the targets field isn't.

docs/flow/components/discovery.transformer.md
rfratto: It technically creates more. Given this label set: 

```
{ 
  "__label_foo" = "1", 
  "__label_bar" = "2",
}
```

And a relabel config of

```hcl
action      = "labelmap"
regex       = "__label_(.*)" 
replacement = "${1}" 
```

the resulting label set will be:

```
{ 
  "__label_foo" = "1", 
  "__label_bar" = "2", 
  "foo" = "1",
  "bar" = "2",
}
```

rfratto:
Something that just occurred to me: I think the syntax used for replacing regex capture groups doesn't work in HCL: `"${1}"` would be interpreted as an HCL interpolated expression and the user would need to escape with `$${1}`. Can we document that?

docs/flow/components/discovery.transformer.md
rfratto: Ahh, you're right, my bad :)

tpaschalis:
I've added a small note that users should refer to capture groups using either the `$1` or `$${1}` notations, and renamed the package to `targets.mutate`.

Let me know how it looks!

docs/flow/components/targets.mutate.md
```diff
@@ -0,0 +1,102 @@
...
+```
+
+## Arguments
+
+The following arguments are supported and can be referenced by other
+components:
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+targets | list(map(string)) | The targets to mutate. | | no
```
rfratto, line 47:
```suggestion
targets | list(map(string)) | The targets to mutate. | | **yes**
```

docs/flow/components/targets.mutate.md
```diff
@@ -0,0 +1,102 @@
...
+relabel_config | RelabelConfig | The relabeling steps to apply. | | no
```
rfratto, line 48: Hmm, for now, can we drop the block as being mentioned as an argument? I think blocks will have different documentation semantics than attributes.

```suggestion
```

docs/flow/components/targets.mutate.md
```diff
@@ -0,0 +1,102 @@
...
+
+
+### RelabelConfig
```
rfratto, line 51: And then change this to be about a block:

```suggestion
### `relabel_config` block
```

rfratto (commented):
Few nits about how we want to document inner blocks, but LGTM, thanks!

rfratto (approved):

docs/flow/components/targets.mutate.md
tpaschalis: Whoops, thanks for the catch.

docs/flow/components/targets.mutate.md
tpaschalis: Yeah, I wasn't sure how to handle them. Makes sense.

PR #1796, opened 2022-06-15
Fix packaging reference link

dohnto:
Not sure what the correct link is, but the current one is broken. PTAL


tpaschalis (approved):
Thanks Tomas!

PR #1801, opened 2022-06-16
v0.25.1 image tags

rlankfo:
Update image tags for v0.25.1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ internal API changes are not present.
 Main (unreleased)
 -----------------
 
-v0.25.1 (TBD)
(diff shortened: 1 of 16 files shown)


rlankfo:
> Could we also bump the tag referenced in [this](https://github.com/grafana/agent/blob/main/docs/user/set-up/install-agent-docker.md) file? Otherwise, LGTM!

good catch, I'll update this one as well.

rfratto (commented):
LGTM, but this is probably missing a few references. Not all of our references start with v, so you may want to do a search for 0.24 and 0.25 to see if anything got left out

rfratto (approved):

PR #1810, opened 2022-06-21
feat(docs): use job=cadvisor label consistently

### LINKED ISSUE #1606: docs/user/operator/custom-resource-quickstart.md: wrong `up{job=…}` metrics
A bunch of Grafana Kubernetes Dashboards (like the default ones produced by `kubernetes-monitoring/mixins` - rendered at https://github.com/monitoring-mixins/website/tree/master/assets/kubernetes/dashboards, use the following query to construct the `$cluster` dashboard variable:

`label_values(up{job="cadvisor"}, cluster)`

Likewise, there's other queries using `job="kubelet"`.

This means, it'd be convenient if the quickstart in `docs/user/operator/custom-resource-quickstart.md` would provide the same `job` labels that are expected there, to provide a nice out of the box experience.

However, things are a bit convoluted.

The `up` metric seems to have the `job=kubelet` label (this seems to come from the job name in the rendered scraping config), as the `metricRelabelings` line only seems to apply to individual metrics, not the `up` job describing the scrape job.

```
    - action: replace
      targetLabel: job
      replacement: integrations/kubernetes/cadvisor
```

Also, as can be seen there, this should be set to `cadvisor`, not `integrations/kubernetes/cadvisor`.

flokli:
This ensures cadvisor metrics have the job=cadvisor label set, which is
used in many Grafana Kubernetes Dashboards, Alerts and Recording rules,
and required to make them work.

Compatibility with Grafana Cloud's Kubernetes integration is dropped,
which was already behind a big warning sign.

In some cases, the labels were also set using metricRelabelings, which
only affected metrics themselves, and not the `up` metric.
Consistently use `relabelings` everywhere.

Fixes #1606.

Fixes #1606.

#### Notes to the Reviewer
`production/tanka/grafana-agent/v2/internal/helpers/k8s.libsonnet` might need an update too.

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/user/operator/custom-resource-quickstart.md
+++ b/docs/user/operator/custom-resource-quickstart.md
@@ -229,14 +229,14 @@ spec:
       regex: kubelet_cgroup_manager_duration_seconds_count|go_goroutines|kubelet_pod_start_duration_seconds_count|kubelet_runtime_operations_total|kubelet_pleg_relist_duration_seconds_bucket|volume_manager_total_volumes|kubelet_volume_stats_capacity_bytes|container_cpu_usage_seconds_total|container_network_transmit_bytes_total|kubelet_runtime_operations_errors_total|container_network_receive_bytes_total|container_memory_swap|container_network_receive_packets_total|container_cpu_cfs_periods_total|container_cpu_cfs_throttled_periods_total|kubelet_running_pod_count|node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate|container_memory_working_set_bytes|storage_operation_errors_total|kubelet_pleg_relist_duration_seconds_count|kubelet_running_pods|rest_client_request_duration_seconds_bucket|process_resident_memory_bytes|storage_operation_duration_seconds_count|kubelet_running_containers|kubelet_runtime_operations_duration_seconds_bucket|kubelet_node_config_error|kubelet_cgroup_manager_duration_seconds_bucket|kubelet_running_container_count|kubelet_volume_stats_available_bytes|kubelet_volume_stats_inodes|container_memory_rss|kubelet_pod_worker_duration_seconds_count|kubelet_node_name|kubelet_pleg_relist_interval_seconds_bucket|container_network_receive_packets_dropped_total|kubelet_pod_worker_duration_seconds_bucket|container_start_time_seconds|container_network_transmit_packets_dropped_total|process_cpu_seconds_total|storage_operation_duration_seconds_bucket|container_memory_cache|container_network_transmit_packets_total|kubelet_volume_stats_inodes_used|up|rest_client_requests_total
       sourceLabels:
       - __name__
-    - action: replace
-      targetLabel: job
-      replacement: integrations/kubernetes/kubelet
     port: https-metrics
     relabelings:
     - sourceLabels:
       - __metrics_path__
       targetLabel: metrics_path
+    - action: replace
+      targetLabel: job
+      replacement: kubelet
     scheme: https
     tlsConfig:
       insecureSkipVerify: true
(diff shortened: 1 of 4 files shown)


docs/user/operator/custom-resource-quickstart.md
```diff
@@ -289,7 +288,7 @@ spec:
       app.kubernetes.io/name: kubelet
 ```
 
-These two ServiceMonitors configure Agent to scrape all the Kubelet and cAdvisor endpoints in your Kubernetes cluster (one of each per Node). In addition, it defines a `job` label which you may change (it is preset here for compatibility with Grafana Cloud's Kubernetes integration), and allowlists a core set of Kubernetes metrics to reduce remote metrics usage. If you don't need this allowlist, you may omit it, however note that your metrics usage will increase significantly.
+These two ServiceMonitors configure Agent to scrape all the Kubelet and cAdvisor endpoints in your Kubernetes cluster (one of each per Node). In addition, they define a `job` label, which is `kubelet` and `cadvisor` respectively, to match the label values used in various Grafana Kubernetes Dashboards, and allowlists a core set of Kubernetes metrics to reduce remote metrics usage. If you don't need this allowlist, you may omit it, however note that your metrics usage will increase significantly.
```
hjet, line 291:
```suggestion
These two ServiceMonitors configure Agent to scrape all the Kubelet and cAdvisor endpoints in your Kubernetes cluster (one of each per Node). In addition, they define a `job` label, which is `integrations/kubernetes/kubelet` and `integrations/kubernetes/cadvisor` respectively, to match the label values used in Grafana Cloud's Kubernetes integration. They also allowlist a core set of Kubernetes metrics to reduce remote metrics usage. If you don't need this allowlist, you may omit it, however note that your metrics usage will increase significantly.

> **Note:** You can omit the `integrations/kubernetes/` prefix (so set the `job` label to just `cadvisor` or `kubelet`) to align with the OSS kubernetes-mixin default `job` labels. You can also modify these `job` selectors [here](https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/config.libsonnet#L21). 
```

flokli:
> all the changes lgtm, but for custom-resource-quickstart.md i'd like to preserve compatibility with the cloud k8s integration (so keep the labels as integrations/kubernetes/*) as we point our users and engineers there at the moment.

The main idea for this PR was to fix compatibility with the dashboards and alerts out there. Reading the docs, I assumed the Grafana Cloud Kubernetes integration was somewhat expected to break eventually, until it aligns with the OSS label names.

The suggested changes don't make things less confusing. If you want to avoid having a time the quickstart here doesn't work with the Grafana Cloud Integration, I'd rather prefer to keep this open and merge it the same time the Grafana Cloud Integration switched to the new labels.

flokli:
@hjet 

> […] this will soon change (in the next month or so), at which point we can make the `job` labels totally compatible with the OSS mixin […]

Is there an update on that? Can we merge this in soonish?

hjet:
@flokli these docs were all hand-written and will soon be updated with K8s yaml generated from jsonnet (PR [here](https://github.com/grafana/agent/pull/1921))

the job labels are [top level parameters](https://github.com/grafana/agent/pull/1921/files#diff-627c7e1e815f14848d3dbbcd8bb4153acf15c5e2222fe4590e10c6278ad68162R41) so we will set them to the OSS defaults in the OSS docs

flokli:
@hjet with https://github.com/grafana/agent/pull/1921 merged, what's the state of this?

I saw that PR didn't touch the markdown files - has there been a followup PR? Should this be rebased?

hjet:
> @hjet with #1921 merged, what's the state of this?
> 
> I saw that PR didn't touch the markdown files - has there been a followup PR? Should this be rebased?

hey there @flokli, 
- https://github.com/grafana/agent/pull/2070 has been merged which fixes the original issue (https://github.com/grafana/agent/issues/1606) - the issue being broken/duplicate `up` metrics
- at this time, we would like to keep `integrations/kubernetes/` in the default `job` labels in this custom resource quickstart guide, until we deliver these operator manifests to users directly in grafana cloud (which will happen very soon)
- there is a more recent set of example manifests that can be found [here](https://github.com/grafana/agent/blob/main/production/operator/templates/agent-operator.yaml), which includes kube-state-metrics, the k8s events integration, and some other updates
  - we need to fix the same relabelings/metricsrelabelings issue in this set of example manifests
  - we can set the default job labels to `cadvisor` and `kubelet` [here](https://github.com/grafana/agent/blob/main/production/kubernetes/build/templates/operator/main.jsonnet#L40) and remove `integrations/kubernetes`
  - ideally this should be split into multiple manifests, one for each component, and then the custom resource quickstart can reference these manifests directly, instead of having them hard coded in the doc. the custom resource quickstart also is in desperate need of an update...
  - PRs welcome for the above 3 bullets, as i likely won't have time to tackle this in the immediate future

hjet:
update: i set the `job` labels to the OSS defaults in the new sample manifests, and fixed the relabel configs (see https://github.com/grafana/agent/pull/2073)

rfratto:
Fixed by #2073, closing

PR #1812, opened 2022-06-21
Global refid usage in wal

mattdurham:
#### PR Description

This allows overriding the reference id with a global reference id provider. By default `NewStorage` works are before, but if calling `NewStorageWithRefidSource` you can pass in a custom provider. A global provider is available. This will allow us to move forward with flow
#### PR Checklist

- [NA] CHANGELOG updated
- [NA] Documentation added
- [X] Tests updated

### DIFF
--- a/pkg/metrics/wal/wal.go
+++ b/pkg/metrics/wal/wal.go
@@ -102,6 +102,36 @@ func (m *storageMetrics) Unregister() {
 	}
 }
 
+// RefIdSource contains a refid that can be reused across multiple instances
+type RefIdSource struct {
+	refid *atomic.Uint64
+}
+
+// Inc increments the refid and returns the new result
+func (r *RefIdSource) Inc() uint64 {
+	return r.refid.Inc()
+}
+
+func (r *RefIdSource) Load() uint64 {
+	return r.refid.Load()
+}
+
+func (r *RefIdSource) Store(v uint64) {
+	r.refid.Store(v)
+}
+
+var globalCache *RefIdSource
+
+// GetGlobalCache returns a refid source that ensures the refids are globally unique
+func GetGlobalCache() *RefIdSource {
+	if globalCache == nil {
+		globalCache = &RefIdSource{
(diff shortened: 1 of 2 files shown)


pkg/metrics/wal/wal.go
```diff
@@ -122,30 +151,30 @@ type Storage struct {
...
 	deleted    map[chunks.HeadSeriesRef]int // Deleted series, and what WAL segment they must be kept until.
 
 	metrics *storageMetrics
+
+	refIdSource *RefIdSource
 }
 
-// NewStorage makes a new Storage.
-func NewStorage(logger log.Logger, registerer prometheus.Registerer, path string) (*Storage, error) {
+func NewStorageWithRefidSource(logger log.Logger, registerer prometheus.Registerer, path string, refid *RefIdSource) (*Storage, error) {
```
rlankfo, line 164: Do we want to add this new constructor API?  It seems like it gives consumers a way to re-create the initial problem by creating storages with different `RefIdSources`.  Couldn't we use the `GlobalRefId` in `NewStorage`?

mattdurham: At some point, once we are more comfortable with the changes we could switch to using the GlobalRefID.

rlankfo: that makes sense -- in that case instead of passing in a `RefIdSource` parameter maybe just have this function use the `GlobalRefID`?

rlankfo: Either way here, we're changing the code path for the existing Agent, I'm just trying to avoid adding an API that makes it easy to shoot ourselves in the foot later while being hard to debug.

rlankfo (approved):
LGTM +1

pkg/metrics/wal/wal.go
mattdurham: Ended up simplifying it, for this initial pass we dont need the wrapper around atomic, so now just passing around atomics for even less change. Still have the two constructors but hopefully can eventually narrow those down.

PR #1814, opened 2022-06-22
Metric Working Group

mattdurham:
#### PR Description

PR to add Metrics/Appender to Flow. Consider the remote_write and scraping components proof of concept to show the workings of the communication between components. Primarily reviewing the receiver/appendable/globalref mapping. The scraper and remote_write are just there to show how the  receiver/appendable/globalref work and will likely be redone later.


cmd/agentflow/basic_scraper.flow
```diff
@@ -0,0 +1,39 @@
+metrics scraper "default"  {
+    targets = [
+        {
+            targets = [{ "__address__" = "localhost:9001"}]
```
mattdurham, line 4: This assumes you have avalanche going `docker run -p 9001:9001 quay.io/freshtracks.io/avalanche --metric-count=1`

cmd/agentflow/basic_scraper.flow
```diff
@@ -0,0 +1,39 @@
...
+            labels = {}
+        }
+    ]
+    receivers = [metrics.remote_write.default.receiver, metrics.remote_write.rw2.receiver]
```
mattdurham, line 8: receivers accepts an array of receivers. The naming of receivers feels overloaded and not specific so VERY open to opinion here.

component/metrics/receiver.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/storage"
+)
+
+// Receiver is used to pass an array of metrics to another receiver
+type Receiver struct {
+	// metrics should be considered immutable
+	Receive func(timestamp int64, metrics []FlowMetric) `hcl:"receiver"`
```
mattdurham, line 11: This is the func that needs to be set by the child component, and the parent calls Receive. The timestamp can apply to all the metrics. We may need to return a mapping of reference ids to flowmetric(s) in the short term.

component/metrics/receiver.go
```diff
@@ -0,0 +1,19 @@
...
+}
+
+// FlowMetric is a wrapper around a single metric without the timestamp
+type FlowMetric struct {
+	Ref    storage.SeriesRef
```
mattdurham, line 16: Maybe add the reference id here so on the return you can get the results. The results then need to be synchronous which might be fine?

component/metrics/remotewrite/remote_write.go
```diff
@@ -0,0 +1,288 @@
...
+	return res, nil
+}
+
+func startTime() (int64, error) { return 0, nil }
+
+var _ component.Component = (*Component)(nil)
+
+// Run implements Component.
+func (c *Component) Run(ctx context.Context) error {
+	c.opts.OnStateChange(Export{Receiver: c.receiver})
```
mattdurham, line 138: This causes the remote writes receiver to bind to the scrapers receivers.

component/metrics/remotewrite/remote_write.go
```diff
@@ -0,0 +1,288 @@
...
+	res := &Component{
+		log:  o.Logger,
+		opts: o,
+		reg:  reg,
+
+		walStore:    walStorage,
+		remoteStore: remoteStore,
+		storage:     storage.NewFanout(o.Logger, walStorage, remoteStore),
+	}
+	res.receiver = &metrics.Receiver{Receive: res.Receive}
```
mattdurham, line 125: Set the receiver, I have to do it in a separate statement since it references itself.

component/metrics/scraper/appendable.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func newScrapeAppendable(receiver []*metrics.Receiver) *scrapeAppendable {
+	return &scrapeAppendable{
+		buffer:    make(map[int64][]metrics.FlowMetric),
+		receivers: receiver,
+	}
+}
+
+func (s *scrapeAppendable) Append(ref storage.SeriesRef, l labels.Labels, t int64, v float64) (storage.SeriesRef, error) {
```
mattdurham, line 27: This converts appendable into the array based format.

component/metrics/scraper/appendable.go
```diff
@@ -0,0 +1,82 @@
...
+	// BIG TODO is that we probably want to move refid creation and caching into a shared cache between wal and scraper at some point.
+	// 	in the below the refid is never cached by the scrape pool which makes it sad. Its likely this / refid caching should
+	//  should create the refid cache
+	s.mut.Lock()
+	defer s.mut.Unlock()
```
mattdurham, line 32: Maybe we dont need the mutexes? I dont think appendables allow multiple writes from different sources to appendable.

component/metrics/scraper/scraper.go
```diff
@@ -0,0 +1,237 @@
...
+	Scheme                string              `hcl:"scheme,optional"`
+	BodySizeLimit         units.Base2Bytes    `hcl:"body_size_limit,optional"`
+	SampleLimit           uint                `hcl:"sample_limit,optional"`
+	TargetLimit           uint                `hcl:"target_limit,optional"`
+	LabelLimit            uint                `hcl:"label_limit,optional"`
+	LabelNameLengthLimit  uint                `hcl:"label_name_length_limit,optional"`
+	LabelValueLengthLimit uint                `hcl:"label_value_length_limit,optional"`
+
+	// TODO(rfratto): http client config
+	Receivers []*metrics.Receiver `hcl:"receivers"`
```
mattdurham, line 58: There is some stuttering that goes on here with flow, in testing it got called with 0 receivers, then 1 and then 2 receivers. Not a big deal but does create some churn.

component/metrics/scraper/appendable.go
```diff
@@ -0,0 +1,86 @@
...
+	_, found := s.buffer[t]
+	if !found {
+		set := make([]*metrics.FlowMetric, 0)
+		s.buffer[t] = set
+	}
+	// If ref is 0 then lets grab a global id
+	if ref == 0 {
+		ref = storage.SeriesRef(metrics.GlobalRefMapping.CreateGlobalRefID(l))
+	}
+	s.buffer[t] = append(s.buffer[t], &metrics.FlowMetric{
```
tpaschalis, line 45: What do we win/lose by batching metrics together by timestamp? 

Reading through Prometheus' scrape loop code, Append() will be called by the same timestamp for every scrape iteration (rounded to avoid jitter) so I _think_ that this buffer map would end up having as many entries as scrapers.

mattdurham: Number of function calls is probably the primary win. If you have a scraper -> remote write there isn't any value since they both get converted to single calls. But adding filters in between would increase the overhead. IE without batching if you had 10 metrics then scraper -> remote write would be 20 total calls , but if you had two filters in between it would be 40, whereas with batching it would 23. 

The loss is some possible memory usage but I think that is a wash. I should prove that though.

tpaschalis: Sorry, I didn't get the point about the filter. 😕  

I'm trying to understand what would change if the buffer would be a `[]metrics.FlowMetric` and it was all batched into a single call to Receive().

component/metrics/globalrefmap.go
```diff
@@ -0,0 +1,96 @@
+package metrics
```
mattdurham, line 1: This is a work in progress ignore please =)

component/metrics/remotewrite/remote_write.go
tpaschalis: I suppose we can't have the function itself as the component's Export, right? It doesn't look so bad though.

mattdurham: I actually didnt try that to see if cty can accept funcs, good call and I will check that.

component/metrics/remotewrite/remote_write.go
```diff
@@ -0,0 +1,290 @@
...
+	if err != nil {
+		return err
+	}
+
+	c.cfg = cfg
+	return nil
+}
+
+func (c *Component) Receive(ts int64, metricArr []*metrics.FlowMetric) {
+	app := c.walStore.Appender(context.Background())
```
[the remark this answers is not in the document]
mattdurham: Good question, will take a look. I imagine we actually want the fanout.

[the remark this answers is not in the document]
tpaschalis: Did the data end up in the remote_write when you tested? If so, I can't figure out the _how_.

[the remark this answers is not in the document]
mattdurham: It did, since the data still goes to the walstorage.

component/metrics/remotewrite/remote_write.go
```diff
@@ -0,0 +1,289 @@
...
+	for _, m := range metricArr {
+		// TODO this should all be simplified into one call
+		if m.GlobalRefID == 0 {
+			globalID := metrics.GlobalRefMapping.CreateGlobalRefID(m.Labels)
+			m.GlobalRefID = globalID
+		}
+		localID := metrics.GlobalRefMapping.GetLocalRefID(c.opts.ID, m.GlobalRefID)
+		newLocal, err := app.Append(storage.SeriesRef(localID), m.Labels, ts, m.Value)
+		if localID == 0 {
+			metrics.GlobalRefMapping.GetOrAddLink(c.opts.ID, uint64(newLocal), m.Labels)
```
tpaschalis, line 253: The Appender implementation here passed from other components will most likely be Prometheus' [headAppender](https://github.com/prometheus/prometheus/blob/f389711943daca716675fdd3b37901f4d70100f7/tsdb/head_append.go#L244-L301).

Do we also want to check that the returned refID is non-zero (i.e. the Append succeeded) before adding the link?

component/metrics/scraper/appendable.go
```diff
@@ -0,0 +1,93 @@
...
+	if len(s.receivers) == 0 {
+		return 0, nil
+	}
+	_, found := s.buffer[t]
+	if !found {
+		set := make([]*metrics.FlowMetric, 0)
+		s.buffer[t] = set
+	}
+	// If ref is 0 then lets grab a global id
+	if ref == 0 {
```
tpaschalis, line 43: @mattdurham Looking at the [Appendable interface](https://github.com/prometheus/prometheus/blob/f389711943daca716675fdd3b37901f4d70100f7/storage/interface.go#L204-L213) in Prometheus, the provided RefID is _optional_ and is used to accelerate calls. If zero is passed, then the matching happens through the labels, where either a new series is created, or the old refID is returned by matching the labels (code is [here](https://github.com/prometheus/prometheus/blob/f389711943daca716675fdd3b37901f4d70100f7/tsdb/head_append.go#L244-L301)).

The call to Append in the scrape loop along with the refID matching happens [here](https://github.com/prometheus/prometheus/blob/f389711943daca716675fdd3b37901f4d70100f7/scrape/scrape.go#L1552). As far as I can tell, Append will be called with a zero refID only if that series is not cached in the scrape loop, (and proceed with the label matching).

Here, we do things a little differently if I get it correctly; we depend on the seriesRef being zero to recognize that it's a new series, at which point we generate a new global refID.

I'm not certain how a series could be dropped from the scrape loop's cache, (a reload of an upstream component maybe?) but what's the worst thing that could happen in our usecase? Is it just that we'd be creating a new refID for a previously-seen series? Would that previous entry ultimately get GCed?

component/metrics/remotewrite/remote_write.go
mattdurham: Thats a good call

component/metrics/scraper/appendable.go
mattdurham: No worries, it was a dumb name. Code that is hard to read is just as bad as bad code ;)

tpaschalis (approved):
LGTM. I'm basing the scraping component over at #1850 on this, let's get more of the puzzle pieces together ^^

PR #1815, opened 2022-06-22
[docs] fix config example errors

marctc:
#### PR Description

Config example for traces is using old configuration entries and leads to
an error if using as it is. Removing the section of `queues` as is not necessary
for a minimal example. 

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated


rfratto:
@mapno Do you remember if it was important for the example config to have retries enabled?

marctc:
@rfratto should we merge it?

rfratto (approved):
Sorry, didn't catch the reply from Mario. Looks good, thanks!

PR #1816, opened 2022-06-22
remove deprecated server YAML fields in favor of static flags

rfratto:
Many `server` YAML fields were deprecated in v0.24.0 and scheduled for removal in v0.26.0. This commit removes those fields in favor of just using the flags.

This also removes the logic for determining listen address by examining both the configured listen address and listen host/port combination, as the latter was only available in the flags. This simplifies the fix introduced by #1762, which can reliably look up the port used for the
gRPC server.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,15 @@ internal API changes are not present.
 Main (unreleased)
 -----------------
 
+> **BREAKING CHANGES**: This release has breaking changes. Please read entries
+> carefully and consult the [upgrade guide][] for specific instructions.
+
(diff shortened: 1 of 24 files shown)


mattdurham:
I dont think all the documentation usages have been changed, are you planning on doing them in another PR? IE [http_listen_port](https://github.com/grafana/agent/blob/main/docs/user/configuration/integrations/integrations-next/snmp-config.md#prometheus-service-discovery-use-case)

rfratto:
Nice catch @mattdurham, I updated the remaining files and I don't see any more references to the removed flags.

PR #1839, opened 2022-06-29
[RFC] River: A Flow-optimized config language

rfratto:
This RFC expands on the expressions design proposed by #1546 with a suggestion that a custom config language for Flow expressions is required to make Flow be as useful and powerful as possible. The current implementation of Flow uses HCL, but with the number of changes I had to make to HCL to get things to work, it doesn't appear to be the best fit. 

I would prefer not creating a new language, but I see it as our only option for Flow. Most of this RFC focuses on the rationale for how I reached this conclusion, why existing languages aren't the best fit, and why it's important to do this for Flow earlier instead of later.

Additionally, in light of proposing a new language, I focus on reducing the learning curve as much as possible to reduce friction from users migrating from the existing YAML configs to Flow.

The current prototype of River is in the main branch, but is subject to removal if we abandon the proposal.

### DIFF
--- a/docs/rfcs/xxxx-river.md
+++ b/docs/rfcs/xxxx-river.md
@@ -21,10 +21,11 @@ interfaces or channels, enabling component developers to easily allow users to
 construct data pipelines using Go APIs without requiring knowledge of the
 underlying implementation.
 
-The initial expressions prototype used [HCL][], which Flow's needs at the time.
-However, our growing dependency on passing around arbitrary Go values started
-to conflict with the limitations of HCL, making HCL increasingly insufficient
-for our specific use case.
+The initial expressions prototype used [HCL][], which initially fit Flow's
+needs during early prototyping. However, the growing dependency on passing
+around arbitrary Go values to build pipelines started to conflict with the
+limitations of HCL, making HCL increasingly insufficient for Flow's specific
+use case.
 
 We examined alternatives to HCL such as YAML, CUE, Jsonnet, Lua, and Go itself.
 Eventually, we determined that the way we use arbitrary Go values in
--- a/docs/rfcs/xxxx-river.md
+++ b/docs/rfcs/xxxx-river.md
@@ -2,7 +2,7 @@
 
 * Date: 2022-06-27
 * Author: Robert Fratto (@rfratto), Matt Durham (@mattdurham)
-* PR: TODO
+* PR: [grafana/agent#1839](https://github.com/grafana/agent/pull/1839)
 * Status: Draft
 
 ## Summary
--- a/docs/rfcs/xxxx-river.md
+++ b/docs/rfcs/xxxx-river.md
@@ -137,7 +137,7 @@ a library to interoperate with tagged Go structs, it was insufficient for
 passing around arbitrary Go values, requiring me to [fork][gohcl] both
 github.com/hashicorp/hcl/v2/gohcl and github.com/zclconf/go-cty/cty/gocty to
 reduce boilerplate. This fork contains a non-trivial amount of changes that
-would need to be maintained or contributed upstream to be tenable.
+would need to be contributed upstream to be tenable long-term.
 
 Additionally, there is desired functionality that is not supported today in
 HCL/go-cty:
--- a/docs/rfcs/xxxx-river.md
+++ b/docs/rfcs/xxxx-river.md
@@ -70,7 +70,8 @@ enable users to manipulate values to meet their own use cases in ways that
 otherwise would require dedicated feature work, such as:
 
 * Allowing users to merge metadata together from distinct sources when adding
-  labels to metrics
+  labels to metrics, such as merging labels from discovered Kubernetes
+  namespaces with discovered Kubernetes pods.
 
 * Allowing users to chain Prometheus service discoveries (e.g., feed the output
   of Kubernetes Service Discovery into HTTP Service Discovery)
@@ -95,11 +96,11 @@ concept.
 
 ### Why an embedded language?
 
-Embedded languages are typically known for the ability for maintainers of the
-project to expose APIs to users of the embedded language, such as the Lua API
-used by Neovim. Embedded languages typically imply tight integration with the
-application embedding them, as opposed to something like YAML which is a
-language consumed once at load time.
+We are using the term "embedded languages" to refer to languages typically
+known for the ability for maintainers of the project to expose APIs to users of
+the embedded language, such as the Lua API used by Neovim. Embedded languages
+typically imply tight integration with the application embedding them, as
+opposed to something like YAML which is a language consumed once at load time.
 
 An embedded language is a good fit for Flow:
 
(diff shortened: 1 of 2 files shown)


docs/rfcs/xxxx-river.md
```diff
@@ -0,0 +1,399 @@
...
+
+This document proposes River, an HCL-inspired declarative expressions-based
+language for continuous runtime evaluation. The decision to propose a new
+language is not taken lightly, and is seen as the last resort. As such, much of
+this proposal will focus on the rationale leading to this choice.
+
+## Goals
+
+* Minimize learning curve as much as possible to reduce friction
+* Make it easy for developers to consume for Flow components
```
mattdurham, line 43: This sentence doesn't flow. Is it 'Make it easy for developers to write flow components'? Or for users to easily use flow components?

docs/rfcs/xxxx-river.md
```diff
@@ -0,0 +1,399 @@
...
+## Rationale
+
+### Why an expression language? Why not YAML?
+
+The entire rationale for creating a new language depends on the rationale that
+expressions provide a useful amount of capabilities to users. Expressions
+enable users to manipulate values to meet their own use cases in ways that
+otherwise would require dedicated feature work, such as:
+
+* Allowing users to merge metadata together from distinct sources when adding
```
mattdurham, line 71: Metadata feels loose here. Is there a more concrete example?

docs/rfcs/xxxx-river.md
```diff
@@ -0,0 +1,399 @@
...
+expressions, you can expose functions for users to call, and its syntax has a
+small learning curve.
+
+However, I found the [schema-driven processing][] API exposed by HCL to be
+difficult to work with for Flow, requiring a lot of boilerplate. While there is
+a library to interoperate with tagged Go structs, it was insufficient for
+passing around arbitrary Go values, requiring me to [fork][gohcl] both
+github.com/hashicorp/hcl/v2/gohcl and github.com/zclconf/go-cty/cty/gocty to
+reduce boilerplate. This fork contains a non-trivial amount of changes that
+would need to be contributed upstream to be tenable long-term.
```
mattdurham, line 140: Do we know the rough code footprint of the changes between cty/hcl to river lines of code?

docs/rfcs/xxxx-river.md
rfratto: oops, not sure what happened here. fixed

docs/rfcs/xxxx-river.md
```diff
@@ -0,0 +1,400 @@
...
+   block like as seen in Terraform.
+
+The combination of desired changes across gohcl and go-cty, the fork that was
+already necessary to make it easier to adopt HCL for Flow, and the desire to
+have a stronger interaction with arbitrary Go values led to the decision that a
+new Flow-specific language was warranted.
+
+### Why now?
+
+Grafana Agent Flow is already a dramatic change to the Agent. To avoid users
```
marctc, line 162: Is this new language necessary for an MVP? I was thinking if HCL is enough for the scope of the MVP, should we invest time on design a new language if later users wouldn't be happy to adopt Flow?

docs/rfcs/xxxx-river.md
```diff
@@ -0,0 +1,400 @@
...
+> For some background, it's important to note that HCL can be considered two
+> separate projects: hashicorp/hcl (the language and expression evaluator) and
+> zclconf/go-cty (the value and type system used by HCL).
+
+HCL was the obvious first choice for the Flow prototype: it supports
+expressions, you can expose functions for users to call, and its syntax has a
+small learning curve.
+
+However, I found the [schema-driven processing][] API exposed by HCL to be
+difficult to work with for Flow, requiring a lot of boilerplate. While there is
```
marctc, line 136: Could you elaborate this a bit more? Where would be this boilerplate in the codebase and why is worst that writing a new languange?

docs/rfcs/xxxx-river.md
```diff
@@ -0,0 +1,404 @@
...
+describing it to allow implementation details to change over time.
+
+## Rationale
+
+### Why an expression language? Why not YAML?
+
+The entire rationale for creating a new language depends on the rationale that
+expressions provide a useful amount of capabilities to users. Expressions
+enable users to manipulate values to meet their own use cases in ways that
+otherwise would require dedicated feature work, such as:
```
tpaschalis, line 70: This would be my #1 point to highlight, in a future-looking mindset ☝️

docs/rfcs/xxxx-river.md
```diff
@@ -0,0 +1,404 @@
...
+  answered.
+
+* Ensuring that error messages explain the problem and the resolution is
+  obvious.
+
+## Proposal
+
+River's syntax is inspired by HCL. However, some of the syntax will be changed
+from HCL to make River more easily identifiable as a different language and
+avoid a situations where users confuse the two.
```
tpaschalis, line 189: Not related to the RFC, but we might want to balance new changes with being able to use existing HCL tooling (eg. syntax highlighting)

docs/rfcs/xxxx-river.md
tpaschalis: An example could be the [tenant](https://grafana.com/docs/loki/latest/clients/promtail/stages/tenant/) pipeline stage from Loki; a metrics component that would allow multitenancy based on specific rules?

rfratto: Added the example I was thinking of here:

> * Allowing users to merge metadata together from distinct sources when adding
  labels to metrics, such as merging labels from discovered Kubernetes
  namespaces with discovered Kubernetes pods.

docs/rfcs/xxxx-river.md
rfratto: Added detail:

> While the fork lets us avoid the boilerplate of hand-crafting
schema definitions for components, it contains a non-trivial amount of changes
that would need to be contributed upstream to be tenable long-term.

docs/rfcs/xxxx-river.md
```diff
@@ -0,0 +1,400 @@
...
+
+YAML by itself does not support expressions. While expressions could be added
+to YAML through the use of templates (e.g., `field_a: {{ some_variable + 5
+}}`), it is beyond the scope of what YAML was intended for and would be more
+cumbersome to use compared to a language where expressions are a first-class
+concept.
+
+### Why an embedded language?
+
+Embedded languages are typically known for the ability for maintainers of the
```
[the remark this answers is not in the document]
rfratto: I don't think there's a formal definition. Lua [describes itself](https://www.lua.org/about.html) as embeddable, but it's just their way of describing how easy it is to integrate Lua. 

My definition of embedded languages here makes some assumptions about what tooling you want for a language being embedded. I've modified the RFC to make it clearer it's a term I'm making up.

docs/rfcs/xxxx-river.md
```diff
@@ -0,0 +1,404 @@
...
+language is not taken lightly, and is seen as the last resort. As such, much of
+this proposal will focus on the rationale leading to this choice.
+
+## Goals
+
+* Minimize learning curve as much as possible to reduce friction
+* Make it easy for developers to create Flow components which operate with
+  arbitrary Go values (interfaces, channels, etc.)
+* Expose error messages in an easily understandable and actionable way
+* Natively support using Go values of any type in config expressions
```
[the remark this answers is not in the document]
rfratto: I feel like I'm already covering this when I first brought up the term:

> Values can refer to arbitrary Go values like
interfaces or channels, enabling component developers to easily allow users to
construct data pipelines using Go APIs without requiring knowledge of the
underlying implementation.

Is there a way I could make this clearer?

docs/rfcs/xxxx-river.md
tpaschalis: Yeah, feel free to disregard this comment. When I saw how quickly you were able to produce VSCode/Vim plugins, it makes less sense to worry about tooling, and as you said gives us absolute freedom to choose our path.

marctc:
@rfratto we probably should merge this before releasing any flow version, wdyt?

rfratto:
> @rfratto we probably should merge this before releasing any flow version, wdyt?

Yeah, agreed, I'll open it up for final review now.

mattdurham (approved):
Ship it

PR #1850, opened 2022-07-05
flow: Add scraping component

### LINKED ISSUE #1679: Flow component for scraping metrics

tpaschalis:
#### PR Description

This PR introduces the `metrics.scrape` component, which kickstarts a Prometheus `scrape.Manager` to scrape metrics, and passes in a custom Appendable which will allow us to fanout these metrics towards multiple receivers defined by some other component.

The following config file will set up two scrape jobs (with separate query parameters), to scrape two instances of the blackbox exporter on ports :9115 and :9116.

```hcl
logging {
  level  = "debug"
  format = "logfmt"
}

metrics "scrape" "blackbox-scraper" {
  targets = [ 
    {"__address__" = "127.0.0.1:9115", "name" = "one"},
    {"__address__" = "127.0.0.1:9116", "name" = "two"},
  ]
  receivers = []

  scrape_config {
    job_name = "grafana"
    scrape_interval = "10s"
    params = { "target" = ["grafana.com"], "module" = ["http_2xx"]}
    metrics_path = "/probe"
  }

  scrape_config {
    job_name = "google"
    scrape_interval = "13s"
    params = { "target" = ["google.com"], "module" = ["http_2xx"]}
    metrics_path = "/probe"
  }
}
```

After the first scrape iteration, our logs show that Append() has been called from both scrape jobs, towards both blackbox_exporter targets
```
Called to Append with 0 {__name__="probe_duration_seconds", instance="127.0.0.1:9115", job="metrics.scrape.blackbox-scraper/grafana", name="one"} 1657105599553 0.02185725
Called to Append with 0 {__name__="probe_duration_seconds", instance="127.0.0.1:9116", job="metrics.scrape.blackbox-scraper/grafana", name="two"} 1657105602575 0.001875458

Called to Append with 0 {__name__="probe_duration_seconds", instance="127.0.0.1:9115", job="metrics.scrape.blackbox-scraper/google", name="one"} 1657105607235 0.0018175
Called to Append with 0 {__name__="probe_duration_seconds", instance="127.0.0.1:9116", job="metrics.scrape.blackbox-scraper/google", name="two"} 1657105600944 0.019284375
```

Fixes #1679 

#### TODOs
- [x] Properly wire up the new Appendable
- [ ] Implement HealthComponent interface?
- [x] Test with proper remote_write implementation 

#### Notes to the Reviewer
Some design decisions to be aware of as you're reviewing :) 

1) When bridging to Prometheus config, there are some fields that were purposefully left out. They are, the SD mechanisms, relabel_configs and metrics_relabel_configs.

As we're breaking the Agent into components, it's most likely that the SD mechanism will be split out to a separate component, whose output will be used as the Arguments `target` field, and thus the SD process will not be run by the scrape manager itself. Similarly, there is already a separate component for mutating targets, plus there will be another one for filtering metrics. I'm not sure if this separation of concerns will have any performance impact, but it definitely maps better to the mental model of how the different moving parts are tied together.

2) As we're experimenting with moving away from the notion of target groups, and having all targets just be a map[string]string, there was some weirdness in the way that we needed to feed the targetSet channel to the scrape manager.

The relevant Prometheus code is [here](https://github.com/prometheus/prometheus/blob/27559e7b4fb3dc9853927757fe1640b219e1832b/scrape/manager.go#L200-L228) for anyone wanting to follow along. To be brief, when reloading the target set, the scrape manager will loop through the given `map[string][]*targetgroup.Group` and will try to match the string key to the manager's cached scrapeConfigs.

Since I suppose that we'd like all scrape_configs to be applied to all targets, I'm using the convention of naming the target sets and scrape configs' job names as "{componentID}/{jobName}".

#### PR Checklist

- [ ] CHANGELOG updated (N/A for now, until we tell people what Flow is about)
- [x] Documentation added
- [ ] Tests updated


component/metrics/scrape/types.go
```diff
@@ -0,0 +1,284 @@
...
+// Config holds all of the attributes that can be used to configure a scrape
+// component.
+type Config struct {
+	// TODO (@tpaschalis) I think we need to override this to be the same value
+	// as the key in the targetGroups map that is being passed through the
+	// channel, and not allow it to be freely set.
+	// The job name to which the job label is set by default.
+	JobName string `hcl:"job_name,attr"`
+
+	// Indicator whether the scraped metrics should remain unmodified.
```
mattdurham, line 32: I think for the MVP doing a clone of the configuration is fine, but I think once we move to more maturity we should take a hard look at what settings we can remove/mutate.

mattdurham: IE  moving all the limit commands to a separate component.

component/metrics/scrape/appendable.go
```diff
@@ -0,0 +1,89 @@
...
+func newFlowAppendable(receivers ...*metrics.Receiver) *flowAppendable {
+	return &flowAppendable{
+		buffer:    make(map[int64][]*metrics.FlowMetric),
+		receivers: receivers,
+	}
+}
+
+func (app *flowAppendable) Appender(_ context.Context) storage.Appender {
+	return app
+}
```
rfratto, line 35: This will cause problems. 


The Prometheus scrape loop will request one appender per scrape job. If that scrape fails, it'll call .Rollback on the appender it received. 

If you used a shared appender for all scrape jobs, one failed scrape will cause the samples for all other metrics to be rolled back.

component/metrics/scrape/scrape.go
```diff
@@ -0,0 +1,209 @@
...
+type Arguments struct {
+	Targets   []Target            `hcl:"targets"`
+	Receivers []*metrics.Receiver `hcl:"receivers"`
+
+	// Scrape Options
+	ExtraMetrics bool `hcl:"extra_metrics,optional"`
+	// TODO(@tpaschalis) enable HTTPClientOptions []config_util.HTTPClientOption
+
+	// Scrape Config
+	ScrapeConfigs []Config `hcl:"scrape_config,block"`
```
rfratto, line 44: I don't think it makes a lot of sense for a scrape component to have more than one scrape job, especially if the set of targets are shared across all jobs. Do you have a use case in mind here?

component/metrics/scrape/types.go
```diff
@@ -0,0 +1,286 @@
...
+
+var emptyAuthorization = common_config.Authorization{}
+var emptyBasicAuth = common_config.BasicAuth{}
+var emptyOAuth2 = &common_config.OAuth2{}
+
+const bearer string = "Bearer"
+
+// Config holds all of the attributes that can be used to configure a scrape
+// component.
+type Config struct {
```
rfratto, line 27: Oh my, this object is big. I'd recommend trimming it down for the MVP or splitting into multiple children blocks so we don't have to prefix fields like `tls_config_insecure_skip_verify`.

component/metrics/scrape/types.go
```diff
@@ -0,0 +1,286 @@
...
+}
+
+// Helper function to bridge the in-house configuration with the Prometheus
+// scrape_config.
+// As explained in the Config struct, the following fields are purposefully
+// missing out, as they're being implemented by another components.
+// - RelabelConfigs
+// - MetricsRelabelConfigs
+// - ServiceDiscoveryConfigs
+func (c *Component) getPromScrapeConfigs(scs []Config) []*config.ScrapeConfig {
```
rfratto, line 135: nit: Implement a method on Config instead to convert into a ScrapeConfig so you don't need an entire Component to convert them.

component/metrics/scrape/scrape.go
tpaschalis: Hmm, mainly for multi-target exporters (and being able to present their grouped status), and only having a single instance of the scrape cache around. Come to think of it, I'm not sure the second point has any gains after all.

I see what you're saying though, and if it also enables us to avoid the issue with the shared Appenders then maybe it's the way to go.

rfratto: It doesn't avoid that problem :) There'd still be one scrape loop per target. I'm just not sure it makes sense to have many scrape jobs that share the same set of targets.

component/metrics/scrape/scrape.go
```diff
@@ -0,0 +1,209 @@
...
+		},
+	})
+	component.RegisterGoStruct("MetricsReceiver", metrics.Receiver{})
+}
+
+// Arguments holds values which are used to configure the metrics.scrape
+// component.
+type Arguments struct {
+	Targets   []Target            `hcl:"targets"`
+	Receivers []*metrics.Receiver `hcl:"receivers"`
```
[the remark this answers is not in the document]
tpaschalis: Agreed. Do you have anything specific in mind? Something like `{write|send|append|forward|transmit_to}` that we used in previous example? Something more abstract like `direction` or  `appender`?

I personally like `forward_to` but open to opinions.

component/metrics/scrape/appendable.go
tpaschalis: Nice detail, I hadn't seen that! Hm, I'm trying to think of a solution here, by looking what the upstream code does.

- headAppender uses an atomically incremented appendID
- agent is using an AppenderPool 

I'll check if we're able to conditionally single out and rollback only the failed appender by using the shared timestamps, otherwise we can reuse one of the other solutions.

component/metrics/scrape/types.go
tpaschalis: I trimmed it down to multiple pointer-type optional blocks on https://github.com/grafana/agent/pull/1850/commits/6709f3436b05f19d5667dfd3e962fd4edf80a2d6. It looks cleaner, simplifies checking for invalid configs and doesn't need prefixes, so I'd consider it an improvement.

component/metrics/scrape/appendable.go
tpaschalis: Your suggestion was around this, right?https://github.com/grafana/agent/pull/1850/commits/f7ed6647aaf5319221bc3e3e2ed3281f3629112c

rfratto: Yep, looks good :)

tpaschalis:
@mattdurham let me know what you think! Having to separate the appender to fix the issue described [here](https://github.com/grafana/agent/pull/1850#discussion_r915875585), we also don't require the mutexes anymore.

mattdurham:
That seems a win-win

PR #1853, opened 2022-07-05
add verbiage around dynamic provisioning

mattdurham:
#### PR Description

Add verbiage around dynamic provisioning


mattdurham:
@captncraig I am still pretty rough on the operator, does this make sense? There was an issue with the user using local storage and not dynamically provisioning the storage. They had to create a PV themselves before it would work.

tpaschalis (approved):

captncraig:
I don't think this is quite accurate. If you don't provide any [storage config](https://github.com/grafana/agent/blob/main/pkg/operator/apis/monitoring/v1alpha1/types.go#L97), the operator will generate empty dir volumes by default. These will use local disk.

If you provide a pvc template, and they fail to dynamically provision, then you will need to manually create them. The names could be more complicated than just `agent-wal` though if you have statefulsets and sharding and all this stuff.

brendamuir (approved):

Issue #1861, opened 2022-07-07
Flow: consider removing singleton component concept


rfratto:
The initial implementation of the Flow controller allows components to register themselves as "singletons," meaning only one of that component may be instantiated:

https://github.com/grafana/agent/blob/80ae7af17ad8ca772cef0f36b3e51c0e3919ad54/component/registry.go#L85-L106

This was introduced because some integrations only support existing once per process:

* agent 
* cAdvisor 
* eventhandler
* node_exporter
* process_exporter
* snmp_exporter
* statsd_exporter
* windows_exporter

I think the singleton component will cause long-term problems, especially around if we ever wanted to support modules, which seems inevitable: two modules that both launch node_exporter with different settings will conflict, and one will be rejected based on load order. 

I think we should slowly invest effort in:

1. Determining which of these integrations breaks if more than one exists per process (I know node_exporter and cAdvisor do due to using global flags to function)
2. Ensuring that the Flow components of these integrations do not function as singletons 
3. Removing the singleton flag from component registration

This is probably a blocker for Flow leaving experimental. Alternative suggestions are welcome for how we can potentially avoid conflict for modules (which although unplanned at the moment, feel inevitable for Flow).

Integrations which need updates to support multiple instances:

- [ ] cAdvisor  
- [ ] node_exporter 

List of integrations which could support multiple instances today: 

* agent 
* eventhandler

List of integrations which need investigation:

* process_exporter
* snmp_exporter
* statsd_exporter
* windows_exporter

tpaschalis:
cc @rfratto @captncraig while looking at #2095, I realised that when we moved from HCL to using River throughout the controller in #1954, we lost the handling around `Singleton` field.

Right now, it's not being utilised anywhere and for example, you can have two instances of `prometheus.integrations.node_exporter` at the same time. Does this make it any more pressing to figure out a solution?

rfratto:
Singleton components aren't allowed in the context of modules, which would prevent remove configuration of components like `prometheus.exporter.unix`. That's all the more reason, IMO, to replace all singleton components with non-singletons and remove the flag.

marctc:
After merging cadvisor and unix flow components removing Singleton, I think we can close this issue. I didn't see any references of Singleton in main.

PR #1872, opened 2022-07-07
(flow) node_exporter integration

### LINKED ISSUE #1684: Flow component for node_exporter
To test integrations with Flow, we need a `node_exporter` component. 

It is not yet obvious how these components will expose metrics. The prototype had an [`integration.github`](https://github.com/grafana/agent/blob/agent-flow/expressions/component/integration/github/github.go) component which exported an integration-specific metrics endpoint as its state. 

The implementation of this component may do something similar, or it may push scraped metrics directly to a receiver. It is up to the discretion of the implementor to weigh the trade offs and pick a path forward, though it should be noted that the chose made will likely influence how other exporter components are implemented.

It is recommended to re-use the existing node_exporter integration code as much as possible for building this component.

captncraig:
Trying to reuse the maximum amount of code from the existing integration. The flow version simply adds hcl tags to the existing config type and invokes the integration directly.

Http handlers are a new frontier here. My first attempt here just adds a subtree at /component/{id}/ that gets delegated to the component itself if it implements HTTPComponent. This is functioning with this config:
```
prometheus.integration.node_exporter "n_ex_test"{
  
}
```
you can read metrics at `http://localhost:12345/component/prometheus.integration.node_exporter.n_ex_test/metrics`.

This is a bit ugly, but it does work. If we wanted to go to in-memory pipes or whatever, the components shouldn't need to change anything, since they just expose a handler.

Fixes #1684


component/all/all.go
```diff
@@ -2,7 +2,8 @@
 package all
 
 import (
-	_ "github.com/grafana/agent/component/local/file"          // Import local.file
-	_ "github.com/grafana/agent/component/metrics/remotewrite" // Import metrics.remotewrite
-	_ "github.com/grafana/agent/component/targets/mutate"      // Import targets.mutate
+	_ "github.com/grafana/agent/component/integrations/node_exporter" // Import targets.mutate
```
tpaschalis, line 5: s/targets.mutate/integration.node_exporter/

pkg/flow/flow_http.go
```diff
@@ -43,6 +44,35 @@ func (f *Flow) ConfigHandler() http.HandlerFunc {
 	}
 }
 
+// ComponentHandler returns an http.HandlerFunc which will delegate all requests to
+// a component named by the first path segment
+func (f *Flow) ComponentHandler() http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		parts := strings.Split(r.URL.Path, "/")
+		id := parts[1]
+		newPath := "/" + strings.Join(parts[2:], "/")
```
rfratto, line 53: I find this a little hard to follow and reason about. Can we take more advantage of mux routing featuers here to make it easier to understand? Something like using `/component/{componentID}` in the route name and potentially a middleware handler to trim the prefix?

captncraig: yeah, I agree its ugly. I'll hit their docs and see the right way to do it. Trickiest part is grabbing the entire trimmed subpath and passing it along as the new root.

component/component.go
```diff
@@ -100,3 +103,11 @@ type DebugComponent interface {
...
 	DebugInfo() interface{}
 }
+
+// HTTPComponent is an extension interface for components which contain their own HTTP Handlers
+type HTTPComponent interface {
+	Component
+
+	// Handler should return a valid http handler for the component
+	Handler() http.Handler
+}
```
rfratto, line 113: Can you document the behavior of how the URL gets trimmed?

component/integrations/node_exporter/node_exporter.go
```diff
@@ -0,0 +1,92 @@
...
+
+		Build: func(opts component.Options, args component.Arguments) (component.Component, error) {
+			return NewComponent(opts, args.(node_integration.Config))
+		},
+	})
+}
+
+// Exports are simply a list of targets for a scraper to consume
+type Exports struct {
+	Output []scrape.Target `river:"output,attr"`
```
rfratto, line 29: `output` feels a little weird here, even though it's consistent with targets.mutate. Should this be `targets` or `metrics`?

component/integrations/node_exporter/node_exporter.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/go-kit/log"
+	"github.com/grafana/agent/component"
+	"github.com/grafana/agent/component/metrics/scrape"
+	node_integration "github.com/grafana/agent/pkg/integrations/node_exporter"
+	"github.com/prometheus/common/model"
+)
+
+func init() {
+	component.Register(component.Registration{
+		Name:    "integration.node_exporter",
```
rfratto, line 17: I think the name is fine for now, but we probably want to rethink it at some point.

What is an integration? 🤷 What are the types of things which go into the integration component group?

`exporter.node` or something along those lines might make more sense down the line as we add more components.

component/integrations/node_exporter/node_exporter.go
```diff
@@ -0,0 +1,92 @@
...
+	if err := c.Update(args); err != nil {
+		return nil, err
+	}
+
+	return c, nil
+}
+
+// Run implements component.Component.
+func (c *Component) Run(ctx context.Context) error {
+	c.log.Log("Msg", "Running")
```
rfratto, line 60: remove please

component/integrations/node_exporter/node_exporter.go
```diff
@@ -0,0 +1,92 @@
...
+	<-ctx.Done()
+	return nil
+}
+
+// Update implements component.Component.
+func (c *Component) Update(args component.Arguments) error {
+	c.log.Log("Msg", "Update")
```
rfratto, line 67: remove please

component/integrations/node_exporter/node_exporter.go
```diff
@@ -0,0 +1,92 @@
...
+	var err error
+	c.integration, err = node_integration.New(c.log, c.cfg)
+	targets := []scrape.Target{{
+		model.AddressLabel:     "127.0.0.1:12345",
```
rfratto, line 71: We need to expose the port number used to components, since it's not always 12345 and the user can override it with a flag.

component/integrations/node_exporter/node_exporter.go
```diff
@@ -0,0 +1,92 @@
...
+	}}
+	c.opts.OnStateChange(Exports{
+		Output: targets,
+	})
+	return err
+}
+
+// Handler serves node_exporter metrics endpoint
+func (c *Component) Handler() http.Handler {
+	if c.integration != nil {
```
rfratto, line 84: Is `c.integration` ever null?

component/integrations/node_exporter/node_exporter.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// Component for node exporter
+type Component struct {
+	log  log.Logger
+	opts component.Options
+
+	cfg *node_integration.Config
+
+	integration *node_integration.Integration
```
rfratto, line 39: You need to protect this with a mutex.

pkg/flow/flow_http.go
```diff
@@ -49,6 +51,36 @@ func (f *Flow) ConfigHandler() http.HandlerFunc {
...
+			w.WriteHeader(http.StatusNotFound)
+			return
+		}
+		handler := node.HTTPHandler()
+		if handler == nil {
+			w.WriteHeader(http.StatusNotFound)
+			return
+		}
+		// remove /component/{id} from front of path, so each component can handle paths from their own root path
+		r.URL.Path = strings.TrimPrefix(r.URL.Path, "/component/"+id)
```
rfratto, line 79: Can we put a helper method in the `component` package to get a path prefix by component ID? It's being spread across two packages (here and the component) and it feels a little magic-stringy.

pkg/integrations/node_exporter/config.go
```diff
@@ -80,61 +80,61 @@ func init() {
 
 // Config controls the node_exporter integration.
 type Config struct {
```
rfratto, line 82: This should really be broken up into multiple blocks (one subblock per configurable collector)  since it's such a massive list of attributes, but that can be a follow up change.

marctc:
Should we add metrics like the added in this [PR](https://github.com/grafana/agent/pull/1982)?

component/integrations/node_exporter/node_exporter.go
rfratto: We have an opportunity to fix the naming with Flow. I felt like the integration terminology got a bit overused. Things like the app_agent_receiver probably deserves to be in its own component namespace separate from all the components which are Prometheus exporters. 

Generally, I'm also willing to delay naming decisions right now until we have more components, since more components will make it easier to see the emerging pattern for how we want to name things. I did want to bring this up so that we start thinking about it, though.

captncraig: I completely agree integrations are super overloaded. I'm gonna go with `exporter.node`.

pkg/flow/flow_http.go
captncraig: Yes, I added HTTPPath to the options passed in at component creation time.

component/integrations/node_exporter/node_exporter.go
rfratto: #2059 suggests specifically `prometheus.integration.node_exporter` and explains why the "exporter" terminology isn't recommended (since it collides with OpenTelemetry's usage of the term)

We can discuss what the final name should be, but I think it should be scoped in the `prometheus` namespace at least.

component/component.go
```diff
@@ -111,3 +114,14 @@ type DebugComponent interface {
 	// DebugInfo must be safe for calling concurrently.
 	DebugInfo() interface{}
 }
+
+// HTTPComponent is an extension interface for components which contain their own HTTP Handlers
+type HTTPComponent interface {
+	Component
+
+	// Handler should return a valid http handler for the component.
```
rfratto, line 122:
```suggestion
	// Handler should return a valid HTTP handler for the component.
```

component/component.go
```diff
@@ -111,3 +114,14 @@ type DebugComponent interface {
 	// DebugInfo must be safe for calling concurrently.
 	DebugInfo() interface{}
 }
+
+// HTTPComponent is an extension interface for components which contain their own HTTP Handlers
```
rfratto, line 118:
```suggestion
// HTTPComponent is an extension interface for components which contain their own HTTP handlers.
```

component/exporters/node/node_exporter.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// Component for node exporter
+type Component struct {
+	log  log.Logger
+	opts component.Options
+
+	integration *node_integration.Integration
+
+	sync.Mutex
```
rfratto, line 41: nit: I'd prefer we avoid embedding sync.Mutex since it shows up at the public API level. 

It also makes it unclear _what_ the mutex is protecting, I prefer to group struct fields by the mutex that is protecting them to make it more clear to maintainers when the mutex is supposed to be locked:

```go
mut         sync.Mutex 
integration *node_integration.Integration
```

component/exporters/node/node_exporter.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// Handler serves node_exporter metrics endpoint
+func (c *Component) Handler() http.Handler {
+	c.Lock()
+	defer c.Unlock()
+	h, err := c.integration.MetricsHandler()
+	if err != nil {
+		c.log.Log(fmt.Errorf("Creating metrics handler: %e", err))
+	}
```
rfratto, line 95: Should this return `http.NotFoundHandler()` or a custom handler which returns a 5xx?

component/exporters/node/node_exporter.go
```diff
@@ -0,0 +1,97 @@
...
+	return err
+}
+
+// Handler serves node_exporter metrics endpoint
+func (c *Component) Handler() http.Handler {
+	c.Lock()
+	defer c.Unlock()
+	h, err := c.integration.MetricsHandler()
+	if err != nil {
+		c.log.Log(fmt.Errorf("Creating metrics handler: %e", err))
```
rfratto, line 94: This will render weird to users, our go-kit/log logger uses key-value pairs and only providing one argument makes it the only key, which will have all spaces stripped. 

```suggestion
		level.Error(c.log).Log("msg", "failed to creating metrics handler", "err", err)
```

component/registry.go
```diff
@@ -53,6 +53,13 @@ type Options struct {
 
 	// Registerer allows components to add their own metrics. The register will come pre-wrapped with the component ID. It is not necessary for components to unregister metrics on shutdown.
 	Registerer prometheus.Registerer
+
+	// HTTPListenAddr is the address the server is listening on. For information only.
```
rfratto, line 57: "for information only" is a little unclear. Can we clarify what that means?

rfratto (commented):
Nice, this is looking better.

component/exporters/node/node_exporter.go
captncraig: 500 for sure

component/all/all.go
```diff
@@ -2,11 +2,12 @@
...
-	_ "github.com/grafana/agent/component/discovery/relabel"      // Import discovery.relabel
-	_ "github.com/grafana/agent/component/local/file"             // Import local.file
-	_ "github.com/grafana/agent/component/prometheus/relabel"     // Import prometheus.relabel
-	_ "github.com/grafana/agent/component/prometheus/remotewrite" // Import prometheus.remote_write
-	_ "github.com/grafana/agent/component/prometheus/scrape"      // Import prometheus.scrape
-	_ "github.com/grafana/agent/component/remote/s3"              // Import s3.file
+	_ "github.com/grafana/agent/component/discovery/kubernetes"                  // Import discovery.k8s
+	_ "github.com/grafana/agent/component/discovery/relabel"                     // Import discovery.relabel
+	_ "github.com/grafana/agent/component/local/file"                            // Import local.file
+	_ "github.com/grafana/agent/component/prometheus/integrations/node_exporter" // Import exporter.node
```
rfratto, line 8:
```suggestion
	_ "github.com/grafana/agent/component/prometheus/integrations/node_exporter" // Import prometheus.integration.node_exporter
```

component/component.go
```diff
@@ -111,3 +114,14 @@ type DebugComponent interface {
...
 }
+
+// HTTPComponent is an extension interface for components which contain their own HTTP handlers.
+type HTTPComponent interface {
+	Component
+
+	// Handler should return a valid HTTP handler for the component.
+	// All requests to the component will have the path trimmed such that the component is at the root.
+	// For example, f a request is made to `/component/{id}/metrics`, the component
+	// will receive a request to just `/metrics`
```
rfratto, line 125:
```suggestion
	// will receive a request to just `/metrics`.
```

component/prometheus/integrations/node_exporter/node_exporter.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/grafana/agent/component/discovery"
+	node_integration "github.com/grafana/agent/pkg/integrations/node_exporter"
+	"github.com/prometheus/common/model"
+)
+
+func init() {
+	component.Register(component.Registration{
+		Name:    "prometheus.integration.node_exporter",
+		Args:    node_integration.Config{},
+		Exports: Exports{},
```
rfratto, line 21: This is our first component which must be a singleton until #1861 is resolved. We need to set `Singleton: true` here to mark it as one.

component/registry.go
```diff
@@ -164,12 +167,14 @@ func parseComponentName(name string) (parsedName, error) {
...
-		if len(other) != len(check) {
-			return fmt.Errorf("%q cannot be used because it is incompatible with %q", check, other)
+		for i := 0; ; i++ {
+			if i >= len(other) || i >= len(check) {
+				// one name is a complete prefix of another
+				return fmt.Errorf("%q cannot be used because it is incompatible with %q", check, other)
+			}
+			if other[i] != check[i] {
+				break
+			}
```
rfratto, line 177: I'm finding this a little hard to follow. If you can't find a way to make it easier to follow, I think a comment explaining what's going on here will be the most useful, particularly about how the loop is always guaranteed to exit and why `i >= len(other) || i >= len(check)` means that the names are incompatible.

pkg/flow/flow_http.go
```diff
@@ -49,6 +51,36 @@ func (f *Flow) ConfigHandler() http.HandlerFunc {
...
+		}
+		if node == nil {
+			w.WriteHeader(http.StatusNotFound)
+			return
+		}
+		handler := node.HTTPHandler()
+		if handler == nil {
+			w.WriteHeader(http.StatusNotFound)
+			return
+		}
```
rfratto, line 77: This might eventually be expensive in the future since we're asking to create a new HTTP handler for every request. Could we add some kind of TODO here noting that we might want to cache the handlers in the future?

rfratto (commented):
Looks good. The only other major thing will be breaking up the massive config struct, and potentially documentation (though maybe documentation can go in another PR).

PR #1874, opened 2022-07-07
river: add internal/value package

rfratto:
This commit introduces the first pass at a value package for River (grafana/agent#1839). It is not wired in anywhere yet, and further PRs will improve the package and eventually wire it in.

The value package handles the internals of the River evaluator, and can transcode to and from Go values.

Note that this commit does not indicate we will ship River, just that we are continuing the prototyping process in the main branch. pkg/river may still be removed in the future in favor of an alternative solution for Flow.

This commit purposefully leaves three larger tasks incomplete to be done in follow up commits:

1. Ability to invoke function values
2. Proper encoding/decoding to Go structs with rvr block tags (currently,
   labels are ignored)
3. Decoding to Go structs with missing required attributes should fail


pkg/river/internal/rivertags/rivertags.go
```diff
@@ -0,0 +1,195 @@
...
+		if !field.IsExported() {
+			panic(fmt.Sprintf("river: rvr tag found on unexported field at %s", printPathToField(ty, field.Index)))
+		}
+
+		options := strings.SplitN(tag, ",", 2)
+		if len(options) == 0 {
+			panic(fmt.Sprintf("river: unsupported empty tag at %s", printPathToField(ty, field.Index)))
+		}
+		if len(options) != 2 {
+			panic(fmt.Sprintf("river: field %s tag is missing options", printPathToField(ty, field.Index)))
```
tpaschalis, line 121: So we'll not be allowing a tag like this with no options, right?
```
Field struct{} `rvr:"my_block"`
```

pkg/river/internal/value/decode.go
```diff
@@ -0,0 +1,382 @@
...
+				return ElementError{Value: val, Index: i, Inner: err}
+			}
+		}
+		rt.Set(res)
+
+	case reflect.Array:
+		res := reflect.New(rt.Type()).Elem()
+		for i := 0; i < val.rv.Len(); i++ {
+			// Stop processing elements if the target array is too short.
+			// TODO(rfratto): should we force array length to be identical?
```
tpaschalis, line 166: Better than silently dropping elements IMHO. I think if someone's going through the trouble of using a Go array, they'd have a reason for wanting exactly N elements.

pkg/river/internal/value/number_value.go
```diff
@@ -0,0 +1,120 @@
...
+	default:
+		panic("river/value: unrecognized Go number type " + v.Kind().String())
+	}
+
+	return numberValue{val, uint8(bits), nk}
+}
+
+func (nv numberValue) Int() int64 {
+	if nv.k == numberKindFloat {
+		return int64(math.Float64frombits(nv.value))
```
tpaschalis, line 90: I suppose this truncates the floating-point part. Maybe we want to raise an error here instead?

pkg/river/internal/value/type.go
```diff
@@ -0,0 +1,121 @@
...
+	case reflect.Func:
+		switch t.NumOut() {
+		case 1:
+			if t.Out(0) == goError {
+				return TypeCapsule
+			}
+			return TypeFunction
+		case 2:
+			if t.Out(1) != goError {
+				return TypeCapsule
```
tpaschalis, line 108: Edge case, but do we want to disallow a river function return (error, error)?

rfratto: Ah, good point. We do, thanks :) 

In the future we might want to consider making it an error to use an invalid function instead of having it fall back to capsules, depending on how confusing developers find the behavior to be.

pkg/river/internal/value/number_value.go
rfratto: It would implicitly truncate, yeah. If we make this return an error, then we'd have to give the user a function to manually convert from a float-type-number to an int-type-number, which may be tedious (or confusing since there's currently only one River number type which encompasses all the Go number types). 

I'd like to leave it as-is for now and revisit later if it becomes a source of confusion.

tpaschalis: Yeah, makes sense. Thanks!

pkg/river/internal/value/decode.go
rfratto: Yeah, I agree. Silently dropping is the behavior encoding/json uses, but I find it weird. I'll change it to enforce the length.

pkg/river/internal/rivertags/rivertags.go
rfratto: Yeah, right now you have to specify if it's an attribute or a block.

pkg/river/internal/value/decode.go
```diff
@@ -0,0 +1,385 @@
...
+		}
+		into = into.Elem()
+	}
+
+	// Fastest cases: we can directly assign values.
+	switch {
+	case val.Type() == TypeNull:
+		// TODO(rfratto): Does it make sense for a null to always decode into the
+		// zero value? Maybe only objects and arrays should support null?
+		into.Set(reflect.Zero(into.Type()))
```
mattdurham, line 68: This feels like something we may pivot on has we get experience but seems reasonable at the moment. I wonder if we could get away with a null-less language

mattdurham (approved):
Looks good, some ideas around null but nothing to get hung up on.

pkg/river/internal/value/decode.go
rfratto: Might be possible, the challenges would be figuring out: 

1. What does a Go nil value encode to? Does it fail to encode?
2. If we have a decode_json function that includes a null, what does that convert to?

We can keep it in the back of our minds as we build this out, we might get some ideas of what we can do

PR #1881, opened 2022-07-10
river: add river/printer package

rfratto:
#### PR Description
river/printer is a pretty-printer of River AST nodes, which will be used for error message formatting when a specific AST node should be displayed to the user.

Much of the implementation river/printer is inspired by go/printer, and so the way it formats files is very similar. For example, this input:

```
secure = true // Enable secure mode
access_log = "/tmp/access_log.txt" // Enable access log
```

is transformed to be aligned with [elastic tabstops](https://nickgravgaard.com/elastic-tabstops/):

```
secure     = true                  // Enable secure mode
access_log = "/tmp/access_log.txt" // Enable access log
```

The formatting rules applied by river/printer are purposefully not formally specified to allow them to change over time as formatting recommendations are established.

river/printer will only print out files with LF; CRLF output is not supported.

Additionally, since river/printer pretty-prints nodes, it also can serve as the basis for a formatter. While a user-facing tool like `riverfmt` is possible, one isn't introduced in this commit; more tests around river/printer are needed for validating that the printer can never produce a different AST than the one it formatted.

See pkg/river/printer/testdata for before-and-after formatting examples, where *.in denote files prior to formating and *.expect denote what files should look like after formatting.

Related to #1839.

#### Notes to the Reviewer

Please let me know if the code is hard to follow and needs a comment explaining why something is being done. It took me some time to understand go/printer, and even though I added more comments here that should make things easier to understand for first-time readers, it's possible there are more opportunities for making reasoning clearer.  

#### PR Checklist

- [x] CHANGELOG updated (N/A)
- [x] Tests updated (N/A)

### DIFF
--- a/pkg/river/ast/ast.go
+++ b/pkg/river/ast/ast.go
@@ -280,7 +280,7 @@ func EndPos(n Node) token.Pos {
 	case *Comment:
 		return n.Start.Add(len(n.Text) - 1)
 	case *AttributeStmt:
-		return EndPos(n.Name)
+		return EndPos(n.Value)
 	case *BlockStmt:
 		return n.RCurly
 	case *IdentifierExpr:
--- a/pkg/river/ast/ast.go
+++ b/pkg/river/ast/ast.go
@@ -48,7 +48,7 @@ type CommentGroup []*Comment
 // removed, EndPos will not be accurate for any comment which contained
 // carriage returns.
 type Comment struct {
-	Start token.Pos // Starting position of comment
+	StartPos token.Pos // Starting position of comment
 	// Text of the comment. Text will not contain '\n' for line comments.
 	Text string
 }
@@ -66,7 +66,7 @@ type BlockStmt struct {
 	Label   string
 	Body    Body
 
-	LCurly, RCurly token.Pos
+	LCurlyPos, RCurlyPos token.Pos
 }
 
 // IdentifierExpr refers to a named value.
@@ -88,14 +88,14 @@ type LiteralExpr struct {
 
 // ArrayExpr is an array of values.
 type ArrayExpr struct {
-	Elements       []Expr
-	LBrack, RBrack token.Pos
+	Elements             []Expr
+	LBrackPos, RBrackPos token.Pos
 }
 
 // ObjectExpr declares an object of key-value pairs.
 type ObjectExpr struct {
-	Fields         []*ObjectField
-	LCurly, RCurly token.Pos
+	Fields               []*ObjectField
+	LCurlyPos, RCurlyPos token.Pos
 }
 
 // ObjectField defines an individual key-value pair within an object.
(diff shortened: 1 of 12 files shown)


pkg/river/ast/ast.go
```diff
@@ -280,7 +280,7 @@ func EndPos(n Node) token.Pos {
 	case *Comment:
 		return n.Start.Add(len(n.Text) - 1)
 	case *AttributeStmt:
-		return EndPos(n.Name)
+		return EndPos(n.Value)
```
rfratto, line 283: This was wrong, an attribute's ending position is the end of the value, not the end of the name (i.e., `attr = true` ends at `true`, not at `attr`)

pkg/river/printer/walker.go
```diff
@@ -0,0 +1,281 @@
...
+	// TODO(rfratto): Should blocks have a oneline format if they're short or
+	// empty? e.g.: `empty_block { attr = 5 }`, `empty_block {}`
+
+	w.p.Write(
+		s.NamePos,
+		&ast.IdentifierExpr{Name: joined, NamePos: s.NamePos},
+	)
+
+	if s.Label != "" {
+		label := fmt.Sprintf("%q", s.Label)
```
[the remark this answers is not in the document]
rfratto: You'd have to do `foo["bar baz"].output`, but we can also update the parser to force labels to be valid identifiers too

pkg/river/printer/walker.go
```diff
@@ -0,0 +1,281 @@
...
+
+		w.p.Write(
+			wsBlank,
+			&ast.LiteralExpr{Kind: token.STRING, Value: label},
+		)
+	}
+
+	w.p.Write(
+		wsBlank,
+		s.LCurly, token.LCURLY, wsIndent,
```
mattdurham, line 122: This took a bit to understand, but I assume its patterned off Gos printer?

pkg/river/printer/walker.go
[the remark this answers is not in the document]
tpaschalis: Oh, makes sense, sorry I missed that.

mattdurham (approved):
LGTM

pkg/river/printer/walker.go
rfratto: yeah, the Write stuff is based off of go/printer. Which part of it took a while to understand? Maybe I can add more comments explaining why something is happening

mattdurham: `s.LCurly, token.LCURLY` the initial look was why are we adding two curlys.

pkg/river/printer/walker.go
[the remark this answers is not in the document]
rfratto: Either of you want to open a PR to require labels to be valid identifiers? I'm comfortable doing it but it should be a small change and a good experience diving into the parser.

pkg/river/printer/walker.go
rfratto: Done, every `token.Pos` field in the AST is now suffixed with `Pos` (i.e., `LCurlyPos` instead of `LCurly`).

pkg/river/printer/walker.go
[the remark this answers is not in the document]
rfratto: Opened as a new issue: #1893

PR #1882, opened 2022-07-11
fix release step

mattdurham:
Testing the release process.

### DIFF
--- a/.drone/drone.yml
+++ b/.drone/drone.yml
@@ -244,7 +244,7 @@ steps:
       - export PATH="$(go env GOPATH)/bin:$PATH"
       - make -j4 DOCKER_OPTS="" BUILD_IN_CONTAINER=false RELEASE_BUILD=true RELEASE_TAG=${DRONE_TAG} publish
 depends_on:
-  - Test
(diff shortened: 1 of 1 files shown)


.drone/drone.yml
```diff
@@ -228,7 +228,7 @@ steps:
       GITHUB_TOKEN:
         from_secret: GITHUB_KEY
     commands:
-      - apt-get install -y apt-transport-https ca-certificates curl gnupg lsb-release
+      - apt-get update &&  apt-get install -y rubygems rpm nsis apt-transport-https ca-certificates curl gnupg lsb-release libbpfcc-dev
```
rfratto, line 231: Why didn't this break before? 🤔

mattdurham: I have no idea. Which is a bit scary.

rfratto: Oh, I bet it's just `libbpfcc-dev`, which wasn't needed prior to v0.26.0 since that's when the eBPF integration was introduced. The previous install step probably was a no-op, but I think the seego based image was already installing those anyway.

PR #1894, opened 2022-07-13
CI: automate creation of build images

rfratto:
Automate the creation of build images any time a tag matching `build-image/vX.Y.Z` is pushed. This tag matcher aligns with the pattern used by projects like golang/tools for creating releases for subprojects within the monorepo.

Assuming a tag `build-image/v0.14.0` is pushed, two Docker images will be pushed:

1. `grafana/agent-build-image:0.14.0`
2. `grafana/agent-build-image:0.14.0-windows`

The former build image is for Linux, while the latter is for Windows.

This change will allow us to more easily manage build images, as well as extremely easily create custom build images for Windows.

Given this change, the existing pipelines are made to consistently match on `refs/tags/v*` so that they don't trigger when `refs/tags/build-image/v*` is pushed.

Further commits to use the custom build images consistently throughout drone.yml are needed.


rfratto:
One downside of this is that it'll pollute our tag namespace with extra tags, especially tags which are used for development and not user-facing projects. We could use another trigger that's not a tag, but a tag felt like the easiest thing to do.

rlankfo (approved):
LGTM 👍

PR #1896, opened 2022-07-13
pkg/usagestats: fix rare race condition in usagestats test

rfratto:
A data race was reported here: https://github.com/grafana/agent/runs/7321137500?check_suite_focus=true

This can happen if the usagestats happen to tick exactly after the context was canceled.


marctc (approved):
Thanks!

PR #1899, opened 2022-07-13
Drone: use grafana/agent-build-image:0.14.0-windows

rfratto:
Additionally, this removes Lint as a dependency to running the Test jobs so they can run in parallel without having to wait for the Lint to finish.

If Lint fails, the pipeline will still be marked as a failure.


rfratto:
5 minutes faster 🎉

mattdurham (approved):
Awesome!

PR #1901, opened 2022-07-13
river/internal/value: support calling function values

rfratto:
Adds a `Value.Call` method which can invoke function values. 

This supports variadic functions; passing all Values as arguments to the function. See the tests for examples of calling functions in various configurations.

Additionally, if the function call has a `(T, error)` return type, and the error is non-nil, the error will be returned as the error for the Call method.


tpaschalis (approved):
LGTM 💯

PR #1908, opened 2022-07-15
flow: Add component for mutating metrics

### LINKED ISSUE #1680: Flow component for mutating samples
There should be a component in Flow which can mutate scraped samples. It should act as a processor: it will receive scraped samples from another component, and forward the mutated result to another component. 

For the initial implementation, it is recommended that the configuration for this component matches Prometheus relabeling rules, where you can provide multiple relabel configs in the component's config block. 

This component needs a thoughtful name.

tpaschalis:

#### PR Description
This is a small PR that allows us to use Prometheus relabelling rules for mutating samples between two receivers.

The PR does a couple of things:
* Extracts the `flowAppendable` functionality into a new package under components/common. This allows the struct to be reused between metrics.mutate and metrics.scrape, as well as a way for us to easily improve on it 
* Extracts the `RelabelConfig` structs, decoding logic, and helpers bridging with the Prometheus struct in a new package, so it can be reused between targets.mutate and metrics.mutate
* Introduces a new `metrics.mutate` component. Looking past the boilerplate, its functionality is pretty simple and is contained in its `Receive` method. All it does is loop through all samples passed in its receiver, apply the relabelling rules, and then append the batch to be forwarded to the next receiver.

The following configuration is a standalone example pipeline where all the available components can be tested

<details>
  <summary>example-config.flow</summary>

```
logging {
  level  = "debug"
  format = "logfmt"
}

targets mutate "blackbox-exporters" {
  targets = [ 
    { "__address__" = "localhost:9115", "instance" = "one"},
    { "__address__" = "localhost:9116", "instance" = "two"},
    { "__address__" = "rblackbox:9115", "instance" = "three"},
    { "__address__" = "rblackbox:9116", "instance" = "four"}
  ]
  
  relabel_config {
    source_labels = ["instance", "__address__"]
    separator     = "@"
    target_label  = "destination"
    action        = "replace"
  } 
  
  relabel_config {
    action        = "keep"
    regex         = "localhost(.*)"
  }
}

metrics scrape "blackbox" {
  targets = targets.mutate.blackbox-exporters.output
  forward_to = [metrics.mutate.cow.receiver]
  scrape_config {
    job_name        = "tpaschalis"
    scrape_interval = "10s"
    params          = { "target" = ["grafana.com"], "module" = ["http_2xx"]}
    metrics_path    = "/probe"
  }
}

metrics mutate "dog" {
  forward_to = [metrics.remote_write.grafanacloud.receiver]
  metric_relabel_config {
    action       = "replace"
    target_label = "mylabel"
    replacement  = "passed-through-metrics-mutate-component"
  }
  metric_relabel_config {
    action       = "replace"
    target_label = "sound"
    replacement  = "woof"
  }
}

metrics remote_write "grafanacloud" {
  external_labels = {
    cluster = "flow-demo"
  }

  remote_write {
    url = "https://prometheus-prod-01-eu-west-0.grafana.net/api/prom/push"
    basic_auth {
      username = "297344"
      password = "eyJr"
    }
  }
}

```

</details>

Fixes #1680 

#### Notes to the Reviewer
We've already discussed that going this route (instead of plugging the samples relabeling functionality in the ScrapeManger as Prometheus does) is going to increase memory usage somewhat. On the other hand, it ties much better to the user's mental model and the whole idea of Flow, and of course has room for improvement on future iterations.

#### PR Checklist

- [ ] CHANGELOG updated (N/A)
- [ ] Documentation added
- [ ] Tests updated


component/metrics/mutate/mutate.go
```diff
@@ -0,0 +1,104 @@
...
+	// Where the relabelled metrics should be forwarded to.
+	ForwardTo []*metrics.Receiver `hcl:"forward_to"`
+
+	// The relabelling steps to apply to each metric before it's forwarded.
+	MetricRelabelConfigs []*flow_relabel.Config `hcl:"metric_relabel_config,block"`
+}
+
+// Exports holds values which are exported by the metrics.mutate component.
+type Exports struct {
+	Receiver *metrics.Receiver `hcl:"receiver"`
```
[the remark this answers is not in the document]
tpaschalis: Agree, it was mind-bending for me as well.

I didn't want to not change too many things on _this_ PR, but I'd also like for us to think about improving the name.

marctc (approved):

component/common/appendable/appendable.go
```diff
@@ -16,13 +16,15 @@ type FlowMetric struct {
 	Value  float64
 }
 
-type flowAppendable struct {
-	receivers []*metrics.Receiver
+// FlowAppendable is a flow-specific implementation of an Appender.
+type FlowAppendable struct {
+	Receivers []*metrics.Receiver
 }
```
rfratto, line 22: nit: why not `type FlowAppendable []*metrics.Receiver`?

component/metrics/mutate/mutate.go
```diff
@@ -0,0 +1,104 @@
...
+
+// Receive implements the receiver.Receive func that allows an array of metrics
+// to be passed around.
+func (c *Component) Receive(ts int64, metricArr []*metrics.FlowMetric) {
+	app := c.appendable.Appender(context.Background())
+	for _, m := range metricArr {
+		m.Labels = relabel.Process(m.Labels, c.mrc...)
+		if m.Labels == nil {
+			continue
+		}
```
rfratto, line 94: This will have some serious CPU overhead if it's done every time new metrics are received. 

This should do something similar to what the Prometheus scraper does: cache the result of relabeling per series ID, completely discarding the cache if the config for the component changes. Entries can be removed from the cache if a staleness marker is received. 

You don't have to do this in this commit, but you can leave it as a TODO and open an issue for it. I'd say making this performant would be a blocker for the release, but not for the MVP.

tpaschalis: That's true, I've added a TODO for now, but I'd like to see if there's any better solution available (chaining two `metrics.mutate` components shouldn't need double the cache memory).

component/common/appendable/appendable.go
tpaschalis: Hmm, this used to contain more fields when we hadn't separated the `FlowAppendable` and the per-scrape job  derived `appender`.

I'll change it to this one-liner type alias, and we can return to the struct definition if we ever need more fields again.

PR #1910, opened 2022-07-15
river/token/builder: add builder package to build files

rfratto:
#### PR Description
The river/token/builder package exposes an API to construct a River file, inspired by github.com/hashicorp/v2/hclwrite. It builds a list of tokens and then formats them back to the user.

It can also encode Go values directly into tokens.

Related to #1839. 

#### Notes to the Reviewer
This also fixes some new small formatting bugs found when printing out lists that contain objects.

#### PR Checklist

- [x] CHANGELOG updated (N/A)
- [x] Tests updated

### DIFF
--- a/pkg/river/token/builder/builder.go
+++ b/pkg/river/token/builder/builder.go
@@ -1,0 +1,186 @@
+// Package builder exposes an API to create a River configuration file by
+// constructing a set of tokens.
+package builder
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+
+	"github.com/grafana/agent/pkg/river/token"
+)
+
+// A File reprents a River configuration file.
+type File struct {
+	body *Body
+}
+
+// NewFile creates a new File.
+func NewFile() *File { return &File{body: newBody()} }
+
+// Tokens returns the File as a set of Tokens.
+func (f *File) Tokens() []Token { return f.Body().Tokens() }
+
+// Body returns the Body contents of the file.
+func (f *File) Body() *Body { return f.body }
+
+// WriteTo renders and formats the File, writing the contents to w.
+func (f *File) WriteTo(w io.Writer) (int64, error) {
+	n, err := printTokens(w, f.Tokens())
+	return int64(n), err
+}
+
+// Bytes renders the File to a formatted byte slice.
+func (f *File) Bytes() []byte {
+	var buf bytes.Buffer
+	_, _ = f.WriteTo(&buf)
+	return buf.Bytes()
+}
+
+// Body is a list of block and attribute statements. A Body cannot be manually
+// created, but is retrieved from a File or Block.
+type Body struct {
+	nodes []tokenNode
+}
+
+// A tokenNode is a structural element which can be converted into a set of
+// Tokens.
+type tokenNode interface {
+	// Tokens builds the set of Tokens from the node.
+	Tokens() []Token
+}
+
+func newBody() *Body {
+	return &Body{}
+}
+
+// Tokens returns the File as a set of Tokens.
+func (b *Body) Tokens() []Token {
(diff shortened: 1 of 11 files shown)


pkg/river/token/builder/builder_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+	expect := format(t, `
+		null_value = null
+	
+		num     = 15 
+		string  = "Hello, world!"
+		bool    = true
+		list    = [0, 1, 2]
+		func    = function
+		capsule = capsule("chan int")
```
rfratto, line 81: This is kind of a weird way to represent function and capsule values, and needs some thought in the future for how we should display them to users. 

Note that however they get displayed _must_ be valid River syntax, since we need to be able to parse the tokens to format them.

pkg/river/token/builder/builder.go
```diff
@@ -0,0 +1,186 @@
...
+// Tokens returns the File as a set of Tokens.
+func (b *Body) Tokens() []Token {
+	var rawToks []Token
+	for i, node := range b.nodes {
+		rawToks = append(rawToks, node.Tokens()...)
+
+		if i+1 < len(b.nodes) {
+			// Append a terminator between each statement in the Body.
+			rawToks = append(rawToks, Token{
+				Type: token.ILLEGAL,
```
tpaschalis, line 66: Was this meant to be token.ILLEGAL, or just a terminator?

pkg/river/token/builder/builder.go
```diff
@@ -0,0 +1,186 @@
...
+	var toks []Token
+
+	for i, frag := range b.Name {
+		toks = append(toks, Token{Type: token.IDENT, Lit: frag})
+		if i+1 < len(b.Name) {
+			toks = append(toks, Token{Type: token.DOT})
+		}
+	}
+
+	toks = append(toks, Token{Type: token.ILLEGAL, Lit: " "})
```
tpaschalis, line 168: Same here and below, not sure I get the ILLEGAL part. Is it because it's just for presentation so it doesn't matter?

rfratto: See the Token docstring:

```
// A Token is a wrapper around token.Token which contains the token type
// alongside its literal. Use token.ILLEGAL to write literal characters such as
// whitespace.
```

We just unfortunately don't have a good way to say "this should be written as-is" other than re-using token.ILLEGAL.

tpaschalis (approved):
LGTM 🎉

mattdurham (approved):
LGTM , question on literaltoken but thats it

pkg/river/token/builder/token.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/grafana/agent/pkg/river/token"
+)
+
+// LiteralTok is used as a token.Token value for writing literal sequences of
+// text into a built file.
+//
+// river/token doesn't have the capability of representing literal strings, so
+// LiteralTok uses token.ILLEGAL as a way to uniquely identify when literal
+// strings should be injected.
+const LiteralTok = token.ILLEGAL
```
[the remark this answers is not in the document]
rfratto: It could be any literal text. We could add token.LITERAL, but we'd have to make it clear that the scanner never emits it and it's just for the token builder package.

WDYT?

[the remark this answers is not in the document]
mattdurham: That feels better to me than the verbiage of illegal, but if it makes more sense to you to use illegal then keep it.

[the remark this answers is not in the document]
rfratto: I think it's probably fine to add a token.LITERAL since there's multiple consumers of the token package anyway

PR #1921, opened 2022-07-18
(agent-operator) Add jsonnet libs for remaining custom resources

hjet:
#### PR Description
This adds a jsonnet library to generate a default set of CRs for agent operator.

Closes https://github.com/grafana/grafana-k8s-plugin/issues/178.

#### PR Checklist

- [x] CHANGELOG updated
- [ ] Documentation added


production/tanka/grafana-agent-operator/internal/logsinstance.libsonnet
```diff
@@ -0,0 +1,40 @@
...
+    local clients = li.spec.clients,
+
+    _config:: {
+        li_labels: {agent: "grafana-agent-logs"},
+        logs_rw_url: 'YOUR_HL_URL',
+        logs_rw_user: 'YOUR_HL_USER',
+        logs_rw_pass: 'YOUR_HL_PASS',
+        logs_secret_name: 'primary-credentials-logs',
+        external_labels: {cluster: 'cloud'}
+    } + config,
```
rgeyer, line 18: In hindsight, these are all "optional" as well. Even though a config without remoteWrite seems useless, we don't want to impose that on the user.

They should be able to configure it without labels, without external labels, and without any remote write. Each of these can be `with{...}` functions. It would be good to allow multiple remoteWrite/client definitions as well.

Something [like this](https://github.com/grafana/agent/blob/main/production/tanka/grafana-agent/v1/lib/metrics.libsonnet#L87-L102).

This would all apply to the metrics instance as well.

production/tanka/grafana-agent-operator/internal/operator.libsonnet
```diff
@@ -0,0 +1,57 @@
...
+
+    local container = k.core.v1.container,
+    local policyRule = k.rbac.v1.policyRule,
+    local serviceAccount = k.core.v1.serviceAccount,
+    local deployment = k.apps.v1.deployment,
+
+    local this = self,
+
+    _images:: {
+      agent_operator: 'grafana/agent-operator:v0.26.0-rc.0',
```
rgeyer, line 12: This could just be a parameter to the function. Then callers know how to change the image, and can store it in a top level config that get's passed down when instantiating the operator.

production/tanka/grafana-agent-operator/internal/operator.libsonnet
```diff
@@ -0,0 +1,57 @@
...
+    },
+
+    _config:: {
+      name: name,
+      namespace: namespace,
```
rgeyer, line 17: This is never actually used AFAIKT.

rgeyer (commented):
Few high level design suggestions and comments.

hjet:
Made the API a bit simpler per your suggestions, PTAL @rgeyer !

rgeyer (commented):
Thanks for cleaning up the api @hjet. Great stuff!

I made several more comments, however.. These are all coming from the perspective of using these tanka files as a library which would be consumed downstream, as I'm assuming that's the intent.

If that *is* the intent, we would want to avoid wrapping the upstream objects in our own custom objects, and re-implementing functionality the upstream object already exposes. I.E. Anywhere that there is a combination of `resource:` and something else returned by a `new..` or `with...` function, it makes it difficult to extend that upstream object.

These libraries expose everything that is needed for a very opinionated use-case, and will "just work" with some config flexibility, but if a user wanted to change anything, such as adding pod monitoring rather than service monitoring, they'd have to re-implement a lot of what's here using the upstream objects.

Happy to chat about this in real-time to add clarity.

Also sorry if it seems that I'm changing my opinion slightly with each review round. As it evolves, I'm forced to think about it slightly differently so new things are shaking out.

production/kubernetes/build/templates/operator/main.jsonnet
```diff
@@ -1,7 +1,157 @@
...
+  },
+
+  _config:: {
+    namespace: '${NAMESPACE}',
+    metrics_url: '${METRICS_URL}',
+    metrics_user: '${METRICS_USER}',
+    metrics_key: '${METRICS_KEY}',
+    logs_url: '${LOGS_URL}',
+    logs_user: '${LOGS_USER}',
+    logs_key: '${LOGS_KEY}',
```
rgeyer, line 38: I haven't seen this pattern before. Are we doing SED replace, or accepting environment variables?

rgeyer (approved):
This looks good to me. Thank you for all of the changes and collab on this.

There are a couple helper functions that I'd like to add, but that can be done in a follow-up PR, I don't want to block this any further.

I did have one question/comment, which I suspect is fine, but if you could enlighten me I'd appreciate it.

production/kubernetes/build/templates/operator/main.jsonnet
hjet: > Are we doing SED replace, or accepting environment variables?

the latter. generated manifests will have `${VAR_NAME}` style placeholders throughout that can be filled in using env vars and `envsubst` - example [here](https://github.com/grafana/agent/blob/main/production/kubernetes/agent-loki.yaml).

since this will be a more complex set of manifests, when i return i'll try to document this a bit better. the end user should not have to generate anything or use jsonnet, though.

captncraig (approved):
Looks pretty good to me. Hopefully if we get to dogfooding, we will find the rough edges and see what additional helpers will be useful.

PR #1926, opened 2022-07-19
river/internal/value: fix block representation in objects and lists

rfratto:
This PR fixes how blocks are represented from objects and lists with the following rules:

1. A block name `a.b.c` is represented as a deeply nested object, where `a` is the top level key, `b` is a key inside `a`, and `c` is a key inside `b`. The block itself is then inside of `c`. 
2. A labeled block is represented as a key following the block name, where the key name is the label value. The block itself is then inside the label key. 
3. An array or slice of labeled blocks is rolled up into a single object, with each label being a key. 
  1. This implicitly requires that labels are unique per block name. 
  2. Like normal, an array or slice of _unlabeled_ blocks is still an array. 

For example, this file:

```river
discovery.kubernetes "pods" {} 
discovery.kubernetes "namespaces" {} 
discovery.static "healthchecks" {} 

metrics.scrape "default" {} 

node_exporter {}
```

is represented as the following object:

```river 
{
  discovery = {
    kubernetes = {
      pods       = { /* ... */ },
      namespaces = { /* ... */ },
    },
    static = {
      healthchecks = { /* ... */ },
    },
  },
  metrics = {
    scrape = {
      default = { /* ... */ }
    },
  },
  node_exporter = { /* ... */ }
}
```

This is accomplished by introducing `structWrapper`, a special type of struct which allows us to partially traverse through a real struct as we navigate keys. In support of this, the cached tags for a struct are now built into a tree.

This is a difficult chunk of code to follow. Please let me know if there's anything I can document or do to make it clearer.

### DIFF
--- a/pkg/river/internal/rivertags/rivertags.go
+++ b/pkg/river/internal/rivertags/rivertags.go
@@ -45,9 +45,9 @@ func (f Flags) GoString() string { return f.String() }
 
 // Field is a tagged field within a struct.
 type Field struct {
-	Name  string // Name of tagged field
-	Index []int  // Index into field (reflect.Value.FieldByIndex)
-	Flags Flags  // Flags assigned to field
+	Name  []string // Name of tagged field
+	Index []int    // Index into field (reflect.Value.FieldByIndex)
+	Flags Flags    // Flags assigned to field
 }
 
 // Get returns the list of tagged fields for some struct type ty. Get panics if
@@ -121,15 +121,17 @@ func Get(ty reflect.Type) []Field {
 			panic(fmt.Sprintf("river: field %s tag is missing options", printPathToField(ty, field.Index)))
 		}
 
+		fullName := options[0]
+
 		tf := Field{
-			Name:  options[0],
+			Name:  strings.Split(fullName, "."),
 			Index: field.Index,
 		}
 
-		if first, used := usedNames[tf.Name]; used && tf.Name != "" {
-			panic(fmt.Sprintf("river: field name %s already used by %s", tf.Name, printPathToField(ty, first)))
+		if first, used := usedNames[fullName]; used && fullName != "" {
+			panic(fmt.Sprintf("river: field name %s already used by %s", fullName, printPathToField(ty, first)))
 		}
-		usedNames[tf.Name] = tf.Index
+		usedNames[fullName] = tf.Index
 
 		switch options[1] {
 		case "attr":
@@ -148,8 +150,12 @@ func Get(ty reflect.Type) []Field {
 
 		// Validate field
 
+		if len(tf.Name) > 1 && tf.Flags&FlagBlock == 0 {
+			panic(fmt.Sprintf("river: field names with `.` may only be used by blocks (found at %s)", printPathToField(ty, tf.Index)))
+		}
+
 		if tf.Flags&FlagLabel != 0 {
-			if tf.Name != "" {
+			if fullName != "" {
 				panic(fmt.Sprintf("river: label field at %s must not have a name", printPathToField(ty, tf.Index)))
 			}
 			if field.Type.Kind() != reflect.String {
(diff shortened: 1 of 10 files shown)


pkg/river/internal/value/decode.go
```diff
@@ -147,6 +146,62 @@ func decode(val Value, into reflect.Value) error {
 	}
 }
 
+func decodeAny(val Value, into reflect.Value) error {
```
rfratto, line 149: This allows us to guarantee the types when decoding into an `interface{}`, whereas before it could've been anything, which is fairly difficult to reason about. 

For example, if you use `interface{}` and decode an object, it always is a `map[string]interface{}` now, where before it could've been a map or a struct from another package.

pkg/river/internal/rivertags/rivertags.go
```diff
@@ -45,9 +45,9 @@ func (f Flags) GoString() string { return f.String() }
 
 // Field is a tagged field within a struct.
 type Field struct {
-	Name  string // Name of tagged field
-	Index []int  // Index into field (reflect.Value.FieldByIndex)
-	Flags Flags  // Flags assigned to field
+	Name  []string // Name of tagged field
```
rfratto, line 48: This needed to be changed into a slice since block names may be separated into fragments (e.g., `a.b.c` should be `[]string{"a", "b", "c"}` so that you only need to do the split once). This also helps with building the tree.

pkg/river/internal/value/decode.go
mattdurham, line 149: There needs to be more doc/clarification around the differences between `decodeAny` and `decode`. That they can recursively call into each other is confusing. They have very similar names and same args? I *think* decodeAny is the backdoor for when we dont concretely know the type and its an interface{}?

rfratto: Right, `decodeAny` is when we're decoding into a `interface{}` and we want to provide a set of fixed types based on the River value. We recursively call `decode` with the non-`interface{}` type to decode into. 

I'll try to update the docs here to make it more clear that's what's going on.

pkg/river/internal/value/tag_cache.go
```diff
@@ -9,9 +9,9 @@ import (
 // tagsCache caches the river tags for a struct type. This is never cleared,
 // but since most structs will be statically created throughout the lifetime
 // of the process, this will consume a negligible amount of memory.
-var tagsCache = make(map[reflect.Type]objectFields)
+var tagsCache = make(map[reflect.Type]*objectFields)
 
-func getCachedTags(t reflect.Type) objectFields {
+func getCachedTags(t reflect.Type) *objectFields {
```
mattdurham, line 14: Can this create duplicate tag trees? (likely reasonable but want for my understanding)

metrics.reciever "default" {
    targets = [ {
        url = "localhost"
     }]
}

exporter.redis "default" {
    targets = [ {
       url = "localhost"
    }]
}

type Target struct {
    URL string `river:"url"`
}

Running getCachedTags on metrics.receiver and exporter.redis will return two trees that have independent child nodes of Target? As opposed to two trees that point to a common Target `tree`?

rfratto: It doesn't dig down that deep. The tree it creates only covers the fields for a struct and how it maps to keys in an object representing that struct. 

if you have 

```go
type MetricsReceiverArguments struct {
   Targets []Target `river:"targets,attr"` 
}

type ExporterRedisArguments struct {
   Targets []Target `river:"targets,attr"` 
}
```

then `getCachedTags` returns two separate trees:

```go
// For MetricsReceiverArguments
&objectFields{
  fields: map[string]rivertags.Field{
    "targets": rivertags.Field{
       Name: []string{"targets"},
       Index: []int{0}, 
       Flags: rivertags.FlagAttr,
    },
  },
}

// For ExporterRedisArguments
&objectFields{
  fields: map[string]rivertags.Field{
    "targets": rivertags.Field{
       Name: []string{"targets"},
       Index: []int{0}, 
       Flags: rivertags.FlagAttr,
    },
  },
}
```

Notice that this isn't recursing into inner struct types. The "tree" only represents the hierarchy of keys within a given struct, so if you have a block name `a.b.c`, we hold leaves for `b` and `c` but that's it.

mattdurham (approved):
LGTM

tpaschalis (approved):

PR #1933, opened 2022-07-20
River: capsule markers, capsule conversion, and custom tokenization 

rfratto:
This PR introduces three small additions required for a River equivalent of our [hcltypes][] package:

1. Define a Capsule interface which explicitly marks a Go type as being a capsule. This is needed for Go types which should be capsules but whose underlying types are non-capsules by default. For example: 

```go
type Secret string

// RiverCapsule ensures that Secret values are treated as a capsule.
func (s Secret) RiverCapsule() {}
``` 

2. Define a ConvertibleFromCapsule and ConvertibleIntoCapsule extension interfaces to allow converting capsule values from and to other Go values respectively. For a Secret type, this will be used to allow converting a string into a Secret, but not a Secret into a string. Two interface types are defined to allow ConvertibleFromCapsule to be implemented on a non-pointer type.

3. Define a Tokenizer interface which is used to print custom tokens for a Go value. This will allow custom printing of types like an OptionalSecret, which should be shown as a literal string when it is marked as cleartext, and hidden when it is marked as sensitive.

[hcltypes]: https://github.com/grafana/agent/tree/main/pkg/flow/hcltypes

### DIFF
--- a/pkg/river/internal/value/capsule.go
+++ b/pkg/river/internal/value/capsule.go
@@ -1,0 +1,51 @@
+package value
+
+import (
+	"fmt"
+)
+
+// Capsule is a marker interface for Go values which forces a type to be
+// represented as a River capsule. This is useful for types whose underlying
+// value is not a capsule, such as:
+//
+//   // Secret is a secret value. It would normally be a River string since the
+//   // underlying Go type is string, but it's a capsule since it implements
+//   // the Capsule interface.
+//   type Secret string
+//
+//   func (s Secret) RiverCapsule() {}
+//
+// Extension interfaces are used to describe additional behaviors for Capsules.
+// ConvertibleCapsule allows defining custom conversion rules to convert
+// between other Go values.
+type Capsule interface {
+	RiverCapsule()
+}
+
+// ErrNoConversion is returned by implementations of ConvertibleCapsule to
+// denote that a custom conversion from or to a specific type is unavailable.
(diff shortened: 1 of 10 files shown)


pkg/river/internal/value/capsule.go
```diff
@@ -0,0 +1,53 @@
...
+// into any Go type which is not the same as the capsule type.
+type ConvertibleIntoCapsule interface {
+	Capsule
+
+	// ConvertInto should convert its value and store it into dst. dst will be a
+	// pointer to a value which ConvertInto is expected to update.
+	//
+	// ConvertInto should return ErrNoConversion if no conversion into dst is
+	// available.
+	ConvertInto(dst interface{}) error
```
tpaschalis, line 52: So the implementations of ConvertFrom/ConvertInto should contain all the available conversions from/to compatible Capsule types, right?

tpaschalis (approved):
LGTM ⛵

pkg/river/internal/value/capsule.go
rfratto: Not just from/to other capsule types, but from/to _any_ Go type (i.e., you could allow a string to be converted into a capsule, or a capsule into a string). 

You also don't have to implement both, allowing you to do the asymmetric conversion that we need for secrets.

mattdurham (approved):
LGTM

PR #1936, opened 2022-07-21
river/vm: introduce vm package for evaluating expressions

rfratto:
This introduces river/vm, the last primary package needed for River to work end-to-end. river/vm exposes an evaluator which can evaluate AST nodes and convert them into River values, which are then decoded into a Go value.

For example, the River expression `3 + 5` would be evaluated by evaluating both sides of the expression and adding them together to get 8.

The evaluator should work mostly as expected:

1. `==` and `!=` can be performed against any two values of any type (`3 == true` returns false and isn't a type error).

2. `+` can be performed against two numbers or two strings.

3. `||` and `&&` can be performed against two bools.

4. Other math binary operators, `-`, `*`, `/`, and `%`, can be performed against two numbers.

5. Other inequality operators, `<`, `<=`, `>`, `>=`, can be performed against two numbers or two strings.

6. A field may be accessed within an object through the `a.b` syntax or through `a[idx]`, where the index is a string.

7. An element in an array may be accessed through `a[idx]` syntax, where the index is a number.

8. Variables are looked up using a scope provided to the evaluator. If a variable is not found in the scope, it is searched in the stdlib (which only contains functions).

9. The evaluator ensures that all required attributes and blocks are set when evaluating a block or file.


pkg/river/internal/value/decode.go
```diff
@@ -64,15 +64,12 @@ func decode(val Value, into reflect.Value) error {
...
-	// Fastest cases: we can directly assign values.
+	// Fastest cases: we can directly assign values without converting.
 	switch {
 	case val.Type() == TypeNull:
 		// TODO(rfratto): Does it make sense for a null to always decode into the
 		// zero value? Maybe only objects and arrays should support null?
 		into.Set(reflect.Zero(into.Type()))
-	case val.rv.Type() == into.Type():
-		into.Set(cloneGoValue(val.rv))
-		return nil
```
rfratto, line 75: I had to remove this because it led to too much unpredictable decoding, and was causing tests for `concat` function to fail, which were expecting `int`s from decodeAny, but that path wasn't getting hit because of this check. 

We might be able to add it back in the future, but removing it is easier for now and gives me more confidence in knowing what the result of decoding is.

pkg/river/internal/value/value.go
```diff
@@ -67,8 +67,10 @@ func Object(m map[string]Value) Value {
 // Array creates an array from the given values. A copy of the vv slice is made
 // for producing the Value.
 func Array(vv ...Value) Value {
-	ty := reflect.ArrayOf(len(vv), goAny)
-	raw := reflect.New(ty).Elem()
+	// Arrays should be slices otherwise any reference to them gets copied by
+	// value into a new pointer.
+	arrayType := reflect.SliceOf(goAny)
+	raw := reflect.MakeSlice(arrayType, len(vv), len(vv))
```
rfratto, line 73: This was causing me a headache with error decorating. Each reference to the array was getting a different pointer, so we weren't able to match the failing value to a node in the AST.

Keeping the array as a slice prevents creating new copies of the array every time it's referenced, though it does lead to one more allocation.

pkg/river/parser/internal.go
```diff
@@ -276,10 +276,12 @@ func (p *parser) parseBlockName() *blockName {
 
 	// [ string ]
 	if p.tok != token.ASSIGN && p.tok != token.LCURLY {
-		if p.tok == token.STRING && len(p.lit) > 2 {
-			bn.Label = p.lit[1 : len(p.lit)-1] // Strip quotes from label
-			if !isValidIdentifier(bn.Label) {
-				p.addErrorf("expected block label to be a valid identifier")
```
rfratto, line 282: This was a small unexpected parse error: `a.b.c ""` was returning an error at the parse level, while I was expecting to see an issue at the evaluator level.

pkg/river/token/builder/builder.go
```diff
@@ -10,6 +10,38 @@ import (
...
+	n, err := printExprTokens(w, e.Tokens())
+	return int64(n), err
+}
+
+// Bytes renders the File to a formatted byte slice.
+func (e *Expr) Bytes() []byte {
+	var buf bytes.Buffer
+	_, _ = e.WriteTo(&buf)
+	return buf.Bytes()
+}
```
rfratto, line 43: This new chunk here allows us to print values as an expression without having to create an entire file. This is used for pretty error messages from the evaluator.

pkg/river/vm/error.go
```diff
@@ -0,0 +1,125 @@
...
+	if ve.Node != nil {
+		return fmt.Sprintf("%s: %s", ast.StartPos(ve.Node).Position(), ve.Message)
+	}
+	return ve.Message
+}
+
+// convertValueError tries to convert err into a ValueError. err must be an
+// error from the river/internal/value package, otherwise err will be returned
+// unmodified.
+func convertValueError(err error, assoc map[value.Value]ast.Node) error {
```
rfratto, line 45: The code here is fairly confusing, and probably needs some cleanup work. We'll need to decide if cleaning it up is a blocker for merging or if it can wait a little.

pkg/river/vm/op_binary.go
```diff
@@ -0,0 +1,324 @@
...
+	case token.SUB: // number - number
+		lhsNum, rhsNum := lhs.Number(), rhs.Number()
+		switch fitNumberKinds(lhsNum.Kind(), rhsNum.Kind()) {
+		case value.NumberKindUint:
+			return value.Uint(lhsNum.Uint() - rhsNum.Uint()), nil
+		case value.NumberKindInt:
+			return value.Int(lhsNum.Int() - rhsNum.Int()), nil
+		case value.NumberKindFloat:
+			return value.Float(lhsNum.Float() - rhsNum.Float()), nil
+		}
```
rfratto, line 71: I would love to make these less repetitive but I haven't found a way to do that 😞

pkg/river/vm/vm_stdlib_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+	tt := []struct {
+		name   string
+		input  string
+		expect interface{}
+	}{
+		{"env", `env("TEST_VAR")`, string("Hello!")},
+		{"concat", `concat([true, "foo"], [], [false, 1])`, []interface{}{true, "foo", false, 1}},
+		{"unmarshal_json object", `unmarshal_json("{\"foo\": \"bar\"}")`, map[string]interface{}{"foo": "bar"}},
+		{"unmarshal_json array", `unmarshal_json("[0, 1, 2]")`, []interface{}{float64(0), float64(1), float64(2)}},
```
tpaschalis, line 22: I tried toying around with these functions a little, I got some issues when decoding JSON `null`s.

When decoding them as struct fields or array elements, I got
```
		{"unmarshal_json nil field", `unmarshal_json("{\"foo\": null}")`, map[string]interface{}{"foo": nil}},
		{"unmarshal_json nil array element", `unmarshal_json("[0, null]")`, []interface{}{float64(0), nil}},

pkg/river/internal/value/decode.go|89| panic: reflect: call of reflect.Value.Type on zero Value 
```

and when trying to use concat with an array that contained a null value, I got
``` 
		{"unmarshal_json array", `concat(unmarshal_json("[0, 1, 2]"), unmarshal_json(true, null, "3"))`, []interface{}{float64(0), float64(1), float64(2), true, nil, float64(3)}},

panic: NULL is not a valid token 
```

I haven't dived in to figure why, just noting it to look into it.

rfratto: Thanks, this is likely another instance where we're not handling nils gracefully. 

The null issue might be a parser error, though that call also looks incorrect since there should only be one string argument

rfratto: I take the "NULL is not a valid token" theory back, I have no idea where that's coming from :)

rfratto: Both of these are fixed now :)

pkg/river/vm/op_unary.go
```diff
@@ -0,0 +1,33 @@
...
+
+	case token.SUB:
+		if val.Type() != value.TypeNumber {
+			return value.Null, value.TypeError{Value: val, Expected: value.TypeNumber}
+		}
+
+		valNum := val.Number()
+		switch valNum.Kind() {
+		case value.NumberKindUint:
+			return value.Uint(-valNum.Uint()), nil
```
tpaschalis, line 24: Do we want to implicitly return an Int here?

pkg/river/vm/op_binary.go
```diff
@@ -0,0 +1,324 @@
...
+	return b
+}
+
+var numberKindPrec = map[value.NumberKind]int{
+	value.NumberKindUint:  0,
+	value.NumberKindInt:   1,
+	value.NumberKindFloat: 2,
+}
+
+func intPow[Number int64 | uint64](n, m Number) Number {
```
tpaschalis, line 315: Nice sneak in here 😅

rfratto: This seems like one of the cases where generics are useful :)

pkg/river/vm/op_binary.go
```diff
@@ -0,0 +1,324 @@
...
+		return value.Bool(lhs.Bool() && rhs.Bool()), nil
+
+	case token.ADD: // number + number, string + string
+		if lhs.Type() == value.TypeString {
+			return value.String(lhs.Text() + rhs.Text()), nil
+		}
+
+		lhsNum, rhsNum := lhs.Number(), rhs.Number()
+		switch fitNumberKinds(lhsNum.Kind(), rhsNum.Kind()) {
+		case value.NumberKindUint:
```
tpaschalis, line 54: Is the `vm` package the right place to check for over/underflows? If we want to have those kind of checks, we can leave a TODO for later on.

tpaschalis (approved):
LGTM, I don't have anything to add, but will play some more with `rivereval` ⛵

mattdurham (approved):
Awesome work, I know the repetitive sections are unruly but seems the most straightforward simple approach

PR #1941, opened 2022-07-22
flow: kubernetes SD

### LINKED ISSUE #1682: Flow component for k8s SD

captncraig:
Implementation of k8s sd, using prometheus components. Includes common helper function that can be used for hopefully any other prom discovery mechanisms.

Closes #1682


component/discovery/kubernetes/configtypes.go
```diff
@@ -0,0 +1,151 @@
+package kubernetes
+
+import (
+	"net/url"
+
+	"github.com/prometheus/common/config"
+)
+
+// TODO (cpeterson) move HTTPClientConfig and subtypes to dedicated package. make sure all custom serializers work
+type HTTPClientConfig struct {
```
tpaschalis, line 10: I also had to implement the HTTPClientConfig from Prometheus [here](https://github.com/grafana/agent/blob/e24c6fab53130b51b8387562322cef1f7d3f97f1/component/metrics/scrape/types.go#L65-L69); we should definitely be reusing them in a shared package.

component/discovery/kubernetes/k8s.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// Run implements component.Component.
+func (c *Component) Run(ctx context.Context) error {
+	f := func(t []scrape.Target) {
+		c.opts.OnStateChange(Exports{Targets: t})
+	}
+	discovery.RunDiscovery(ctx, c.ch, f)
+	// if we get here, we've canceled
+	if c.cancel != nil {
```
tpaschalis, line 115: I might be missing here, but why do we need to use the context.CancelFunc here? Is it not enough to depend on `<-ctx.Done` like we do for other components?

component/discovery/kubernetes/k8s.go
```diff
@@ -0,0 +1,136 @@
...
+		c.cancel()
+	}
+	return nil
+}
+
+// Update implements component.Compnoent.
+func (c *Component) Update(args component.Arguments) error {
+	newArgs := args.(SDConfig)
+	fmt.Println(newArgs)
```
tpaschalis, line 124: Let's remove this

component/discovery/kubernetes/configtypes.go
tpaschalis: Also, might be a good idea to call the HTTPClientConfig's [Validate method](https://github.com/prometheus/common/blob/49b36038aead362ce78f09337dc5872000bd048a/config/http_config.go#L278-L280) to check that the user didn't define conflicting auth mechanisms.

component/discovery/kubernetes/k8s.go
```diff
@@ -0,0 +1,136 @@
...
+		Args:    SDConfig{},
+		Exports: Exports{},
+
+		Build: func(opts component.Options, args component.Arguments) (component.Component, error) {
+			return New(opts, args.(SDConfig))
+		},
+	})
+}
+
+type SDConfig struct {
```
tpaschalis, line 29: Do we want to apply some defaults and validation checks like [Prometheus does](https://github.com/prometheus/prometheus/blob/d9c650b92e0a52a39151792ef3697825e20589c9/discovery/kubernetes/kubernetes.go#L173-L174) in a DecodeHCL method? Or will the errors gracefully float up to our layer?

component/discovery/discovery.go
```diff
@@ -0,0 +1,67 @@
...
+	"time"
+
+	"github.com/grafana/agent/component/metrics/scrape"
+	"github.com/prometheus/prometheus/discovery/targetgroup"
+)
+
+// RunDiscovery is a utility for consuming and forwarding target groups from a discoverer.
+// It will handle collating targets (and clearing), as well as time based throttling of updates.
+// The channel provided should be the same channel passed to the Discovery's Run method.
+func RunDiscovery(ctx context.Context, ch <-chan []*targetgroup.Group, f func([]scrape.Target)) {
```
tpaschalis, line 14: I kinda like the idea presented here; I _think_ it will easily allow us to plug in new SD mechanisms. Do we have anything else in mind that would enable portability?

Couple of small things; 
a) do we want to make the frequency configurable? I think that was your original intent here, right? 
b) I think we could improve readability by simplifying some `if dirty` and `nextValidTime` bits with a simple time.Ticker. Is it just so that we avoid an initial empty change log, or is there another reason?
c) Have we encountered cases where throttling is going to be required, or is it an early optimization?

component/discovery/discovery.go
```diff
@@ -0,0 +1,67 @@
...
+		}
+		select {
+		case <-ctx.Done():
+			return
+		case <-timeChan:
+			continue
+		case groups := <-ch:
+			for _, group := range groups {
+				if len(group.Targets) == 0 {
+					delete(cache, group.Source)
```
tpaschalis, line 59: Let's try and make sure this clears the cache correctly.  The cache would only be cleared if the `group.Source` key was present with zero targets, but would remain forever there if the key was simply never fetched.

What is the `source` field of each group here? How is it commonly updated through the different Prometheus SD mechanisms?

captncraig: For k8s sd, the source is the pod id (namespace/pod name or something like that). It is a unique key, then the targets have lable sets for the pod. By convention, if a pod is deleted, the discovery will send an update with 0 targets as a signal that the pod is gone and to remove its targets.

component/discovery/discovery.go
captncraig: With the way I've done it now, infrequent updates would get sent right away, while frequent ones will be sent as soon as the time has elapsed. With a ticker, all updates would be stuck waiting for the next tick. Maybe that is a premature optimization though. It certainly makes the code quite a bit more cumbersome. 

I think the throttling is modeled from prometheus, and their "contract" that says discovery should have a maximum update frequency. @rfratto may know more. If that is the case, we may or may not want frequency to be too configurable depending on the rigidity of that requirement.

component/discovery/kubernetes/k8s.go
captncraig: We definitely do.

component/discovery/kubernetes/k8s.go
captncraig: The underlying discovery gets recreated every time the component updates. The component creates a context for it to signal it to exit anytime we get an update. The cancel func here is how we create that cancellation signal. `<-ctx.Done` is how you respond to the signal.

component/discovery/kubernetes/configtypes.go
captncraig: Yes, now is definitely the right time to make a shared package. Any suggestions for the right import path?

component/discovery/discovery.go
rfratto, line 14: It feels like this should be `RunDiscovery(ctx context.Context, d discovery.Discoverer, f func([]scrape.Target))`.

rfratto: We can consider making it configurable in the future, but not required for this PR IMO. It's also not currently configurable in the Prometheus SD manager either; no more than once every 5 sec is [hard coded](https://github.com/prometheus/prometheus/blob/00ec720c29bce637c1580a2ce43b32122d133f83/discovery/manager.go#L119).

component/discovery/discovery.go
```diff
@@ -0,0 +1,67 @@
...
+// The channel provided should be the same channel passed to the Discovery's Run method.
+func RunDiscovery(ctx context.Context, ch <-chan []*targetgroup.Group, f func([]scrape.Target)) {
+	cache := map[string]*targetgroup.Group{}
+
+	dirty := false
+
+	const maxChangeFreq = 5 * time.Second
+	// this should give us 2 seconds at startup to collect some changes before sending
+	var lastChange time.Time = time.Now().Add(-3 * time.Second)
+	for {
```
rfratto, line 22: This whole thing is pretty hard to follow for me. I'd like to simplify this and sacrifice the completeness of the throttling functionality for now just so it's easier for us to reason about.

Something like what Prometheus does:

```go
var (
 ticker      = time.NewTicker(5 * time.Second)
 queueUpdate = make(chan struct{}, 1)
)


for {
  select {
  case <-ctx.Done():
    return 
    
  case <-ticker.C:
    select {
    case <-queueUpdate: 
      // ... calculate new targets and send to f ...  
    default:
      // No targets queued; wait to next tick.
    }
    
  case groups := <-ch:
    // ... update cache ... 

    // Schedule an update for the next tick if the channel is non-empty.
    select {
    case queueUpdate <- struct{}{}:
    default:
      // Update already queued; no-op. 
    }  
  }
}
```

component/discovery/discovery.go
tpaschalis: I would personally prefer a snippet like the one presented below; I don't think it would be too much of a performance hit.

component/discovery/discovery.go
tpaschalis: Okay, then it would work correctly. Can we leave a comment for this behavior, eg. in the docstring where we mention that the cache is being cleared?

component/discovery/discovery.go
captncraig: The problem with that is lifecycle of the Discoverer. If the component updates, we will make a new Discoverer. 

But we don't need to make a new channel. This loop can continue uninterrupted.

But maybe restarting this is viable too. Lemme see if that works.

captncraig: On second thought, I think we **need** to restart this loop on update. Otherwise you could get orphaned cached targets.

component/discovery/kubernetes/configtypes.go
tpaschalis: Could be under `component/common` but I'm not sure about the package name. `promhttpclientconfig` sounds unwieldy; `promhttp` maybe?

component/discovery/kubernetes/k8s.go
```diff
@@ -0,0 +1,139 @@
...
+	return &promk8s.SelectorConfig{
+		Role:  promk8s.Role(sc.Role),
+		Label: sc.Label,
+		Field: sc.Field,
+	}
+}
+
+// Exports holds values which are exported by the discovery.k8s component.
+type Exports struct {
+	Targets []scrape.Target `hcl:"targets,optional"`
```
tpaschalis, line 89: Let's remember to turn HCL tags over to `river`.

component/discovery/discovery.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/grafana/agent/component/metrics/scrape"
+	"github.com/prometheus/prometheus/discovery"
+	"github.com/prometheus/prometheus/discovery/targetgroup"
+)
+
+const MaxUpdateFrequency = 5 * time.Second
+
+// RunDiscovery is a utility for consuming and forwarding target groups from a discoverer.
+// It will handle collating targets (and clearing), as well as time based throttling of updates.
+func RunDiscovery(ctx context.Context, d discovery.Discoverer, f func([]scrape.Target)) {
```
tpaschalis, line 16: Maybe leave a comment on what we expect `f()` to be, so that people won't have to go back to where it's being used?

component/common/config/types.go
```diff
@@ -0,0 +1,153 @@
...
+// DefaultHTTPClientConfig for initializing objects
+var DefaultHTTPClientConfig = HTTPClientConfig{
+	FollowRedirects: true,
+	EnableHTTP2:     true,
+}
+
+// BasicAuth configures Basic HTTP authentication credentials.
+type BasicAuth struct {
+	Username     string `river:"username,attr,optional"`
+	Password     string `river:"password,attr,optional"`
```
rfratto, line 47: This needs to use `rivertypes.Secret` instead of `string` to avoid leaking the value at /debug/config.

component/common/config/types.go
```diff
@@ -0,0 +1,153 @@
...
+	return config.URL{URL: urlp}
+}
+
+// Secret mirrors config.Secret
+type Secret string
+
+// Authorization sets up HTTP authorization credentials.
+type Authorization struct {
+	Type            string `river:"type,attr,optional"`
+	Credentials     string `river:"credentials,attr,optional"`
```
rfratto, line 82: This needs to use `rivertypes.Secret` instead of `string` to avoid leaking the value at /debug/config.

component/common/config/types.go
```diff
@@ -0,0 +1,153 @@
...
+		ServerName:         t.ServerName,
+		InsecureSkipVerify: t.InsecureSkipVerify,
+		MinVersion:         config.TLSVersion(t.MinVersion),
+	}
+}
+
+// OAuth2Config sets up the OAuth2 client.
+type OAuth2Config struct {
+	ClientID         string            `river:"client_id,attr,optional"`
+	ClientSecret     string            `river:"client_secret,attr,optional"`
```
rfratto, line 129: This needs to use `rivertypes.Secret` instead of `string` to avoid leaking the value at /debug/config.

component/common/config/types.go
```diff
@@ -0,0 +1,153 @@
...
+
+	"github.com/prometheus/common/config"
+)
+
+// HTTPClientConfig mirrors config.HTTPClientConfig
+type HTTPClientConfig struct {
+	BasicAuth       *BasicAuth     `river:"basic_auth,block,optional"`
+	Authorization   *Authorization `river:"authorization,block,optional"`
+	OAuth2          *OAuth2Config  `river:"oauth2,block,optional"`
+	BearerToken     Secret         `river:"bearer_token,attr,optional"`
```
rfratto, line 15: This needs to use `rivertypes.Secret` instead of `string` to avoid leaking the value at /debug/config.

component/common/config/types.go
```diff
@@ -0,0 +1,153 @@
...
+	if u == "" {
+		return config.URL{}
+	}
+	urlp, _ := url.Parse(string(u))
+
+	return config.URL{URL: urlp}
+}
+
+// Secret mirrors config.Secret
+type Secret string
```
rfratto, line 77: Remove this and use rivertypes.Secret instead.

component/common/config/types.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	return &config.Authorization{
+		Type:            a.Type,
+		Credentials:     config.Secret(a.Credentials),
+		CredentialsFile: a.CredentialsFile,
+	}
+}
+
+// TLSVersion mirrors config.TLSVersion
+type TLSVersion uint16
```
rfratto, line 99: This is missing the implementation for encoding.TextMarshaler/encoding.TextMarshaler. You can implement the missing methods now or comment out any field which references it and add a TODO.

component/common/config/types.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	return &config.BasicAuth{
+		Username:     b.Username,
+		Password:     config.Secret(b.Password),
+		PasswordFile: b.PasswordFile,
+	}
+}
+
+// URL mirrors config.URL
+type URL string
```
rfratto, line 64: This should be 

```go
type URL struct {
  *url.URL 
}
```

with an implmeentation of encoding.TextMarshaler/encoding.TextUnmarshaler to validate on load, just like how prometheus/common/config does it.

component/discovery/discovery.go
```diff
@@ -0,0 +1,71 @@
...
+
+	"github.com/grafana/agent/component/metrics/scrape"
+	"github.com/prometheus/prometheus/discovery"
+	"github.com/prometheus/prometheus/discovery/targetgroup"
+)
+
+// MaxUpdateFrequency is the minimum time to wait between updating targets.
+// Currently not settable, since prometheus uses a static threshold, but
+// we could reconsider later.
+const MaxUpdateFrequency = 5 * time.Second
```
rfratto, line 15: nit: this doesn't need to be exported

component/discovery/discovery.go
```diff
@@ -0,0 +1,71 @@
...
+	ch := make(chan []*targetgroup.Group)
+	go d.Run(ctx, ch)
+
+	// function to convert and send targets in format scraper expects
+	send := func() {
+		allTargets := []scrape.Target{}
+		for _, group := range cache {
+			for _, target := range group.Targets {
+				labels := map[string]string{}
+				for k, v := range group.Labels {
```
rfratto, line 33: nit: can you add a comment here explaining that it's important for the group to go first since the labels from target take precedence over the group labels?

component/discovery/kubernetes/k8s.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// SDConfig is a conversion of discover/kubernetes/SDConfig to be compatible with flow
+type SDConfig struct {
+	APIServer          config.URL              `river:"api_server,attr,optional"`
+	Role               string                  `river:"role,attr"`
+	KubeConfig         string                  `river:"kubeconfig_file,attr,optional"`
+	HTTPClientConfig   config.HTTPClientConfig `river:"http_client_config,attr,optional"`
+	NamespaceDiscovery NamespaceDiscovery      `river:"namespaces,attr,optional"`
+	Selectors          []SelectorConfig        `river:"selectors,attr,optional"`
```
rfratto, line 32: All three of these should be blocks and not attributes; blocks are for grouping related settings together while attributes are for values which can be passed around as a whole.

component/common/config/types.go
```diff
@@ -0,0 +1,153 @@
+// Package config contains types from github.com/prometheus/common/config, but changes them to be compatible with river
```
rfratto, line 1: general nit: please use proper capitalization for nouns like River/Prometheus, and use proper grammar in comments so comments start with capital letters and end in full stops. You should be writing comments like documentation that external users are going to read, because they will.

component/discovery/kubernetes/k8s.go
```diff
@@ -0,0 +1,140 @@
...
+func (nd *NamespaceDiscovery) convert() *promk8s.NamespaceDiscovery {
+	return &promk8s.NamespaceDiscovery{
+		IncludeOwnNamespace: nd.IncludeOwnNamespace,
+		Names:               nd.Names,
+	}
+}
+
+// SelectorConfig mirroring prometheus type
+type SelectorConfig struct {
+	Role  string `river:"role,attr,optional"`
```
rfratto, line 78: Is role really optional? What does it mean to define a selector with nothing inside of it?

component/discovery/kubernetes/k8s.go
```diff
@@ -0,0 +1,140 @@
...
+func (c *Component) Update(args component.Arguments) error {
+	newArgs := args.(SDConfig)
+	disc, err := promk8s.New(c.opts.Logger, newArgs.Convert())
+	if err != nil {
+		return err
+	}
+	// cancel any previously running discovery
+	if c.cancel != nil {
+		c.cancel()
+	}
```
rfratto, line 129: Reads and writes to c.cancel need to be protected by a mutex since it gets read in Component.Run.

component/discovery/kubernetes/k8s.go
```diff
@@ -0,0 +1,140 @@
...
+	// function to send updates on change
+	f := func(t []scrape.Target) {
+		c.opts.OnStateChange(Exports{Targets: t})
+	}
+	// create new context so we can cancel it if we get any future updates
+	newCtx, cancel := context.WithCancel(context.Background())
+	c.cancel = cancel
```
rfratto, line 136: Reads and writes to c.cancel need to be protected by a mutex since it gets read in Component.Run.

component/common/config/types.go
```diff
@@ -0,0 +1,153 @@
...
+
+// URL mirrors config.URL
+type URL string
+
+// Convert converts our type to the native prometheus type
+func (u URL) Convert() config.URL {
+	if u == "" {
+		return config.URL{}
+	}
+	urlp, _ := url.Parse(string(u))
```
rfratto, line 71: Don't ignore errors please. I realize this was in the code you copied, but we should fix this now before merging.

component/discovery/kubernetes/k8s.go
```diff
@@ -0,0 +1,140 @@
...
+	// finally run discovery
+	go discovery.RunDiscovery(newCtx, disc, f)
```
rfratto, line 138: Can you add a TODO here to eventually move this call to Component.Run? The update of a component should ideally not start any new goroutines and instead notify the Run goroutine to rework things. 

This doesn't have to be fixed now, but we shouldn't lose track of it so the TODO comment will be good enough.

component/common/config/types.go
captncraig: And what to do here? Passing up chain would require changing a lot of signatures. I could return an empty url in that case perhaps?

rfratto: No, the user should be told that the URL is invalid and the component should fail to evaluate. 

https://github.com/grafana/agent/pull/1941#discussion_r934896820 

If you follow this comment then you won't need to parse it though in this call tree though, since it'd be parsed at decoding time.

component/common/config/types.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+}
+
+// URL mirrors config.URL
+type URL struct {
+	*url.URL
+}
+
+// MarshalText implements encoding.TextMarshaler
+func (u *URL) MarshalText() (text []byte, err error) {
```
rfratto, line 72: Can you change this to be a non-pointer receiver? There's issues in Flow where MarshalText doesn't get invoked for pointer receivers because the arguments is rendered as a non-pointer type.

Ditto with TLSVersion.

component/discovery/kubernetes/k8s.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// Update implements component.Compnoent.
+func (c *Component) Update(args component.Arguments) error {
+	newArgs := args.(SDConfig)
+	disc, err := promk8s.New(c.opts.Logger, newArgs.Convert())
+	if err != nil {
+		return err
+	}
+	c.newDiscoverer <- disc
```
rfratto, line 147: Hmm. 

I think this is probably fine, but it might cause a deadlock if someone Update gets called twice when a component it still exiting. 

Just to be safe, can we change this to

```go
select {
case c.newDiscoverer <- disc: 
default: 
}
```

(Though honestly that's not ideal either; storing the latest discoverer in a field and using a notification channel might be more 'correct' but I think we need to change something here just to be safe, especially while the component controller is still young)

component/discovery/kubernetes/k8s.go
```diff
@@ -0,0 +1,165 @@
...
+		opts: o,
+		// buffered to avoid deadlock from the first immediate update
+		newDiscoverer: make(chan struct{}, 1),
+	}
+	return c, c.Update(args)
+}
+
+// Run implements component.Component.
+func (c *Component) Run(ctx context.Context) error {
+	var cancel context.CancelFunc
```
rfratto, line 118: This will always leak the last context. Can you add

```go
defer func() {
  if cancel != nil {
     cancel() 
  }
}
```

?

rfratto: Ah, I think I'm misreading the docs: https://pkg.go.dev/context#WithCancel

It seems like it's fine as long as either the parent or the child context gets canceled, which will definitely happen here.

rfratto (approved):
LGTM, nice work!

PR #1942, opened 2022-07-22
river: fix custom capsule conversions & support time.Duration

rfratto:
This PR makes two small changes:

1. Fixes a bug where it wasn't possible to perform custom conversions between two capsule types.

2. Implements custom logic for encoding and decoding time.Duration as strings.


tpaschalis (approved):
LGTM 💯

PR #1948, opened 2022-07-25
Fix metrics.scrape panic with nil receiver

### LINKED ISSUE #1945: Flow: metrics.scrape panics after first scrape
Use the following config file with the current Flow implementation in main:

```hcl
logging {
  level  = "debug"
  format = "logfmt"
}

metrics "scrape" "default" {
  targets = [{
    "__address__" = "demo.robustperception.io:9090",
  }]
  forward_to = [metrics.remote_write.default.receiver]

  scrape_config {
    job_name = "default"
  }
}

metrics "remote_write" "default" {
  remote_write {
  }
}
```

After the first scrape, the component will panic:

```
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x10462a528]

goroutine 115 [running]:
github.com/grafana/agent/component/common/appendable.(*flowAppender).Commit(0x14000bce6e0)
	/Users/robert/dev/grafana/agent/component/common/appendable/appendable.go:74 +0xf8
github.com/prometheus/prometheus/scrape.(*scrapeLoop).scrapeAndReport.func1()
	/Users/robert/dev/go/pkg/mod/github.com/grafana/prometheus@v1.8.2-0.20220413182558-6b32d0b957c5/scrape/scrape.go:1263 +0x48
github.com/prometheus/prometheus/scrape.(*scrapeLoop).scrapeAndReport(0x14000b84680, {0xeda6f5fa8?, 0x1053a51e0?, 0x0?}, {0xeda6f5fa8?, 0x1053a51e0?, 0x1053a51e0?}, 0x0)
	/Users/robert/dev/go/pkg/mod/github.com/grafana/prometheus@v1.8.2-0.20220413182558-6b32d0b957c5/scrape/scrape.go:1334 +0xb5c
github.com/prometheus/prometheus/scrape.(*scrapeLoop).run(0x14000b84680, 0x14000290240?)
	/Users/robert/dev/go/pkg/mod/github.com/grafana/prometheus@v1.8.2-0.20220413182558-6b32d0b957c5/scrape/scrape.go:1216 +0x284
created by github.com/prometheus/prometheus/scrape.(*scrapePool).sync
	/Users/robert/dev/go/pkg/mod/github.com/grafana/prometheus@v1.8.2-0.20220413182558-6b32d0b957c5/scrape/scrape.go:587 +0x880
exit status 2
```

The issue seems to be that the appendable is created [once](https://github.com/grafana/agent/blob/1fda90adc917c8d107e2e2a0207611adc43222b4/component/metrics/scrape/scrape.go#L71-L74), and the appendable the scraper uses never updates after construction.

`metrics.remote_write` initially exports a nil receiver, which is where `metrics.scrape` is getting it from. This also shouldn't be happening, but is a separate issue (#1946).

`metrics.scrape` needs to:

1. Protect against trying to invoke nil receivers 
2. Allow the receivers list to change as the component gets updated

tpaschalis:

#### PR Description
This behavior appeared when we changed the `type struct { receivers []*metrics.Receiver }` definition to the `type FlowAppendable []*metrics.Receiver` type alias in 7be5f6c04c3d8b814c1c74de40d46bb7ce7c1f37.

This PR fixes the panic which occurs when the `metrics.scrape` component tries to forward to a nil receiver by reverting to the previous behavior

Fixes #1945

#### PR Checklist

- [ ] CHANGELOG updated (N/A) until Flow is out of the PoC
- [ ] Tests updated

### DIFF
--- a/component/common/appendable/appendable.go
+++ b/component/common/appendable/appendable.go
@@ -17,11 +17,15 @@ type FlowMetric struct {
 }
 
 // FlowAppendable is a flow-specific implementation of an Appender.
-type FlowAppendable []*metrics.Receiver
+type FlowAppendable struct {
+	receivers []*metrics.Receiver
+}
 
 // NewFlowAppendable initializes the appendable.
-func NewFlowAppendable(receivers ...*metrics.Receiver) FlowAppendable {
-	return receivers
+func NewFlowAppendable(receivers ...*metrics.Receiver) *FlowAppendable {
+	return &FlowAppendable{
+		receivers: receivers,
+	}
 }
 
 type flowAppender struct {
@@ -33,8 +37,13 @@ type flowAppender struct {
 func (app FlowAppendable) Appender(_ context.Context) storage.Appender {
 	return &flowAppender{
 		buffer:    make(map[int64][]*metrics.FlowMetric),
-		receivers: app,
+		receivers: app.receivers,
 	}
+}
+
+// SetReceivers defines the list of receivers for this appendable.
+func (app *FlowAppendable) SetReceivers(receivers []*metrics.Receiver) {
+	app.receivers = receivers
 }
 
 func (app *flowAppender) Append(ref storage.SeriesRef, l labels.Labels, t int64, v float64) (storage.SeriesRef, error) {
--- a/component/common/appendable/appendable.go
+++ b/component/common/appendable/appendable.go
@@ -53,6 +53,13 @@ func (app *FlowAppendable) SetReceivers(receivers []*metrics.Receiver) {
 	app.mut.Unlock()
 }
 
+// ListReceivers is a test method for exposing the Appender's receivers.
+func (app *FlowAppendable) ListReceivers() []*metrics.Receiver {
+	app.mut.RLock()
+	defer app.mut.RUnlock()
+	return app.receivers
+}
+
 func (app *flowAppender) Append(ref storage.SeriesRef, l labels.Labels, t int64, v float64) (storage.SeriesRef, error) {
 	if len(app.receivers) == 0 {
 		return 0, nil
--- a/component/common/appendable/appendable.go
+++ b/component/common/appendable/appendable.go
@@ -87,7 +87,7 @@ func (app *flowAppender) AppendExemplar(ref storage.SeriesRef, l labels.Labels, 
 func (app *flowAppender) Commit() error {
 	for _, r := range app.receivers {
 		for ts, metrics := range app.buffer {
-			if r.Receive == nil {
+			if r == nil || r.Receive == nil {
 				continue
 			}
 			r.Receive(ts, metrics)
--- a/component/common/appendable/appendable.go
+++ b/component/common/appendable/appendable.go
@@ -2,6 +2,7 @@ package appendable
 
 import (
 	"context"
+	"sync"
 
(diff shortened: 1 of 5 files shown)


component/common/appendable/appendable.go
```diff
@@ -37,6 +37,13 @@ func (app FlowAppendable) Appender(_ context.Context) storage.Appender {
 	}
 }
 
+// SetReceivers sets the list of receivers the appendable should forward to.
+func (app FlowAppendable) SetReceivers(receivers ...*metrics.Receiver) {
+	if receivers != nil {
+		app = FlowAppendable(receivers)
+	}
+}
```
rfratto, line 45: I don't think this will work, since it's a non-pointer receiver. Can you write a test? 

It might be better to switch back to the struct to make it easier to reason about ensuring the correct slice of receivers is being used for appending.

tpaschalis: The non-pointer receiver is a type alias for a pointer-type though, so that's why I thought it'd work 🤷 

In any case, I'd also be in favor of returning to a the struct, I've got a PR ready.

component/metrics/scrape/scrape.go
```diff
@@ -128,7 +128,7 @@ func (c *Component) Update(args component.Arguments) error {
 	defer c.mut.Unlock()
 	c.args = newArgs
 
-	c.appendable = fa.FlowAppendable(newArgs.ForwardTo)
```
rfratto, line 131: Just want to point this out: this was incorrect behavior :) 

This assigns a new value to c.appendable (i.e., a new pointer to memory containing a slice), while the scraper continues using the previous value of c.appendable (i.e., a different pointer to memory containing a different slice).

rfratto: i.e., for the update to work correctly the value of c.appendable _must_ stay the same, and instead the underlying memory c.appendable points to changes.

component/common/appendable/appendable.go
rfratto: > The non-pointer receiver is a type alias for a pointer-type though,

It still makes a copy of the slice header. So you have an extra value pointing to a location in memory containing a slice, and by doing `app = FlowAppendable(receivers)`, you are just changing where that new copy is pointing, not changing the slice.

tpaschalis: I see, both this and the next comment make sense now, sorry for not catching this in the first place.

rfratto: no worries :)

component/common/appendable/appendable.go
```diff
@@ -33,10 +37,15 @@ type flowAppender struct {
 func (app FlowAppendable) Appender(_ context.Context) storage.Appender {
 	return &flowAppender{
 		buffer:    make(map[int64][]*metrics.FlowMetric),
-		receivers: app,
+		receivers: app.receivers,
```
rfratto, line 40: I'm pretty sure you need to protect this with a mutex now to prevent race conditions between updating the component and scraping happening at the same time

rfratto: Probably a RWMutex since many scrape loops might want a rear lock at roughly the same time

rfratto (approved):
LGTM!

PR #1951, opened 2022-07-25
River: various small fixes

rfratto:
This PR fixes 7 small issues I found while testing River for Flow, and adds one new feature (river/diag). 

The small changes / bug fixes are as follows: 

* river/internal/value: Add support for an Unmarshaler interface so types can hook custom logic into decoding (e.g., like for applying default values) 

* river/ast: Create an ast.Walk function to allow for walking the AST. This also changes separates IdentExpr into an Ident so AST visitors can guarantee when an IdentExpr is part of an expression. This will be used for determining what references a Flow component is making to other Flow components. 

* river/token/builder: Support writing out comments. 

* river/token/builder: Allow encoding bodies from Go values. This was a missing piece of functionality needed to implement the /-/config endpoint. 

* river/internal/value: Minimize usage of Value.rv when decoding. Value.rv should be considered internal-only and not used throughout the decoder. This removes most usages, but complete elimination will require future work. 

* river/internal/value: Move logic for TextMarshaler/time.Duration to the value package so values of these types are actually treated as strings, instead of only displaying as strings. 

* river/internal/value: Properly detect River type for values which only implement an interface for a pointer receiver.   

The new `river/diag` package implements a shared error type and a pretty-printer to show these errors to users. The pretty-printer produces Rust-inspired error messages where the range of text from the offending file is displayed to the user:

```
1 |   test.block "label" {
2 |   	attr       = 1
  |   	^^^^
3 |   	other_attr = 2
```

These changes should be merged _prior_ to #1936, which will need to be updated to also handle some of the changes here. 

I've separated each of these changes into their own commits for easier reviewing. If reviewers would prefer separate PRs, please let me know. I decided to try for a single commit given the number of small changes needed.

### DIFF
--- a/pkg/river/diag/diag.go
+++ b/pkg/river/diag/diag.go
@@ -1,0 +1,84 @@
+// Package diag exposes error types used throughout River and a method to
+// pretty-print them to the screen.
+package diag
+
+import (
+	"fmt"
+
+	"github.com/grafana/agent/pkg/river/token"
+)
+
+// Severity denotes the severity level of a diagnostic. The zero value of
+// severity is invalid.
+type Severity int
+
+// Supported severity levels.
+const (
+	SeverityLevelWarn Severity = iota + 1
+	SeverityLevelError
+)
+
+// Diagnostic is an individual diagnostic message. Diagnostic messages can have
+// different levels of severities.
+type Diagnostic struct {
+	// Severity holds the severity level of this Diagnostic.
+	Severity Severity
+
+	// StartPos refers to a position in a file where this Diagnostic starts.
+	StartPos token.Position
+
+	// EndPos refers to an optional position in a file where this Diagnostic
+	// ends. If EndPos is the zero value, the Diagnostic should be treated as
+	// only covering a single character (i.e., StartPos == EndPos).
+	//
+	// When defined, EndPos must have the same Filename value as the StartPos.
+	EndPos token.Position
+
+	Message string
+	Value   string
+}
+
+// Error implements error.
+func (d Diagnostic) Error() string {
+	return fmt.Sprintf("%s: %s", d.StartPos, d.Message)
+}
+
+// Diagnostics is a collection of diagnostic messages.
+type Diagnostics []Diagnostic
+
+// Add adds an individual Diagnostic to the diagnostics list.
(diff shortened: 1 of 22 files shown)


mattdurham (approved):
LGTM

pkg/river/token/builder/builder_test.go
```diff
@@ -93,6 +95,73 @@ func TestBuilder_GoEncode(t *testing.T) {
...
+		ReqFieldA string `river:"req_field_a,attr"`
+		ReqFieldB string `river:"req_field_b,attr"`
+	}
+
+	f := builder.NewFile()
+	f.Body().AppendFrom(Structure{
+		OptFieldA: "some_value",
+		OptFieldB: "", // Zero value
+		ReqFieldA: "some_value",
+		ReqFieldB: "", // Zero value
```
tpaschalis, line 153: Nice to see this working 'correctly'! ^^

pkg/river/diag/diag.go
```diff
@@ -0,0 +1,84 @@
...
+
+// Error implements error.
+func (ds Diagnostics) Error() string {
+	switch len(ds) {
+	case 0:
+		return "no errors"
+	case 1:
+		return ds[0].Error()
+	default:
+		return fmt.Sprintf("%s (and %d more diagnostics)", ds[0], len(ds)-1)
```
tpaschalis, line 62: I'm not sure if this would ever reach the user, but I'd rather we provide a more complete image, as these errors might be related to each other. Should we print out more of them?

tpaschalis (approved):
LGTM! 💯

pkg/river/diag/diag.go
rfratto: Do you have a suggestion of what it should return?

rfratto: Going to merge for now, but let's revisit this later and iterate on what the raw error string does

tpaschalis: Yeah, didn't mean for this to be a blocker of course.

I'm not sure if/where this is going to be presented to the user I was just asking if we should for example, show the 5 top errors (skipping warnings maybe?) before offering the "and N more diagnostics" message.

PR #1954, opened 2022-07-26
Use River throughout Flow

rfratto:
This commit replaces usage of HCL for the Flow POC with River. All components have been updated to use river tags in their struct definition.

Struct tags were replaced with the following pattern:

* `hcl:"KEY"` becomes `river:"KEY,attr"`
* `hcl:"KEY,optional"` becomes `river:"KEY,attr,optional"`
* `hcl:"KEY,block"` becomes `river:"KEY,block"`

Additionally, if `hcl:"KEY,block"` is tagged against a slice or a pointer field, it is made optional: `river:"KEY,block,optional"`.

It is no longer required to register Go structs to use as capsules with River, so the documentation for that has been removed.

As discussed with @mattdurham, this removes the ability for expressions to reference the arguments of other components; expressions may now only reference the exports of other components. The ability to reference both may be restored in the future.

Low-level documentation about how to use River is spread across a few packages. High-level documentation is out of scope of this PR.


cmd/agentflow/example-config.river
```diff
@@ -0,0 +1,29 @@
+logging {
```
rfratto, line 1: Note that running this example file will eventually panic because of #1945.

pkg/flow/config.go
```diff
@@ -1,75 +1,86 @@
...
+	"github.com/grafana/agent/pkg/river/diag"
+	"github.com/grafana/agent/pkg/river/parser"
+	"github.com/grafana/agent/pkg/river/vm"
 )
 
 // File holds the contents of a parsed Flow file.
 type File struct {
 	Name string    // File name given to ReadFile.
-	HCL  *hcl.File // Raw HCL file.
+	Node *ast.File // Raw File now.
```
rfratto, line 17: What happened here?

```suggestion
	Node *ast.File // Raw File node.
```

pkg/flow/config_test.go
```diff
@@ -1,72 +1,49 @@
...
-	content := `
-		doesnotexist "hello-world" {
-		}
-	`
-
-	f, diags := flow.ReadFile(t.Name(), []byte(content))
-	require.Nil(t, f)
-	require.True(t, diags.HasErrors())
-	require.Equal(t, `Blocks of type "doesnotexist" are not expected here.`, diags[0].Detail)
-}
```
rfratto, line 53: This error was returned because of HCL's schema checks, which doesn't exist in River. Now, the controller will return an error for invalid components.

pkg/flow/internal/controller/component.go
```diff
@@ -137,21 +141,6 @@ func NewComponentNode(globals ComponentGlobals, b *hcl.Block) *ComponentNode {
...
-	// twice: once with all name fragments in the ID and once without the last
-	// one (e.g., the one that represents the user ID).
-	reg, ok := component.Get(id.String())
-	if ok {
-		return reg, ok
-	}
-
-	reg, ok = component.Get(id[:len(id)-1].String())
-	return reg, ok
-}
```
rfratto, line 153: This is no longer necessary because now we can cleanly tell what part of a block is a user label and what part is the name, unlike before where the label slice contained part of a user label and part of the component name.

pkg/flow/internal/controller/value_cache.go
```diff
@@ -165,36 +135,5 @@ func (vc *valueCache) buildValue(from []ComponentID, offset int) cty.Value {
...
-		mergedMap[key] = value
-	}
-	for key, value := range stateMap {
-		if _, exist := mergedMap[key]; exist {
-			panic(fmt.Sprintf("component exports overrides arguments key %s", key))
-		}
-		mergedMap[key] = value
-	}
-
-	return cty.ObjectVal(mergedMap)
```
rfratto, line 199: This was removed after a discussion with @mattdurham; it's too confusing to explain to users how arguments get converted into values, but much easier to explain exports. For now, since we can't think of a use case where a user would need to reference an argument, only the exports can be referenced.

pkg/flow/root_eval_context.go
```diff
@@ -1,76 +0,0 @@
...
-	"github.com/zclconf/go-cty/cty/function"
-	"github.com/zclconf/go-cty/cty/function/stdlib"
-)
-
-// rootEvalContext is a set of global variables and functions which will be
-// available to all expressions in HCL.
-//
-// controller.Loader creates a child of this context which includes values for
-// running components.
-var rootEvalContext = &hcl.EvalContext{
```
rfratto, line 15: This has been removed in favor of the native River stdlib. River's stdlib is much smaller, but we can add to it over time as we identify functions we want to expose.

pkg/river/types.go
```diff
@@ -0,0 +1,75 @@
+package river
```
rfratto, line 1: Some of the River interfaces are hidden in internal packages, so I'm using pkg/river to document and re-expose them.

component/local/file/file.go
```diff
@@ -55,20 +54,20 @@ var DefaultArguments = Arguments{
...
 	type arguments Arguments
-	return gohcl.DecodeBody(body, ctx, (*arguments)(a))
+	return f((*arguments)(a))
 }
 
 // Exports holds values which are exported by the local.file component.
 type Exports struct {
 	// Content of the file.
-	Content *hcltypes.OptionalSecret `hcl:"content,attr"`
+	Content rivertypes.OptionalSecret `river:"content,attr"`
```
mattdurham, line 70: Was it intentional to change from a pointer?

rfratto: not intentional, but I think a non-pointer is correct actually (go-cty might've required a pointer, I can't remember)

mattdurham (approved):
Large change, but a good chunk of it is renaming! LGTM

cmd/agentflow/main.go
```diff
@@ -77,6 +80,23 @@ func run() error {
 	}
 
 	if err := reload(); err != nil {
+		var diags diag.Diagnostics
+		if errors.As(err, &diags) {
+			bb, _ := os.ReadFile(configFile)
+
+			p := diag.NewPrinter(diag.PrinterConfig{
+				Color:              !color.NoColor,
```
tpaschalis, line 88: This gets set by the library automatically based on the environment it's running at. Do we want to negate it, or just set it to always true?

tpaschalis (approved):
LGTM!

cmd/agentflow/main.go
rfratto: Right, negating it was the intent so you don't get the color codes for things like logs sent to Loki. I don't think it should be always true.

PR #1956, opened 2022-07-27
flow: Expose metrics on flow controller

### LINKED ISSUE #1939: Flow: expose metrics on Flow controller
The Flow controller should expose metrics about itself. There's a few metrics I'm mainly interested in to start with:

* Number of running components by health (i.e., how many components are unhealthy)  
* Histogram of component evaluations (i.e., how many times do we re-evaluate components and how long do they take?)

There's probably more metrics we could expose, but I think this is a good starting set; the implementation which adds metrics for the controller may choose to add more.

Given the proposal described in #1914, metric names should probably _not_ contain `flow`, and instead be prefixed with either `agent_controller` or `agent_component_controller`, or something along these lines.

marctc:
#### PR Description

This PR adds metrics to flow controller package in order to
instrument the process of evaluating and running Flow components.

#1939

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [x] Tests updated


pkg/flow/internal/controller/metrics.go
```diff
@@ -0,0 +1,38 @@
+package controller
+
+import "github.com/prometheus/client_golang/prometheus"
+
+var (
+	controllerEvaluation = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "agent_component_controller_evaluating",
```
tpaschalis, line 7: Let's keep in mind to have a consistent approach with for prefixes in future metrics.

tpaschalis (approved):
LGTM

rfratto (changes requested):

rfratto:
Sorry, I wrote my comments yesterday and forgot to submit them. I'm not sure if they're still relevant.

pkg/flow/internal/controller/metrics.go
```diff
@@ -0,0 +1,44 @@
...
+		Help: "Tracks if the controller is currently in the middle of a graph evaluation",
+	})
+	componentEvaluationTime = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Name: "agent_component_evaluation_seconds",
+			Help: "Time spent performing component evaluation",
+		},
+	)
+
+	componentLabels := []string{"id", "health_type"}
```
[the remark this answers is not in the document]
mattdurham: My worry on this would be for clustering, so would also be pro per health status instead of id.

mattdurham (approved):
LGTM

pkg/flow/internal/controller/component.go
```diff
@@ -418,6 +421,7 @@ func (cn *ComponentNode) setRunHealth(t component.HealthType, msg string) {
 	cn.healthMut.Lock()
 	defer cn.healthMut.Unlock()
 
+	cn.cm.runningComponentTotal.WithLabelValues(cn.id.String(), t.String()).Inc()
```
rfratto, line 424: This never gets decremented so it doesn't accurately reflect the number of running components. 

Also the labels don't match up with the metric you constructed.

pkg/flow/internal/controller/component.go
```diff
@@ -99,7 +101,7 @@ var (
 
 // NewComponentNode creates a new ComponentNode from an initial ast.BlockStmt.
 // The underlying managed component isn't created until Evaluate is called.
-func NewComponentNode(globals ComponentGlobals, b *ast.BlockStmt) *ComponentNode {
+func NewComponentNode(globals ComponentGlobals, b *ast.BlockStmt, cm *componentMetrics) *ComponentNode {
```
rfratto, line 104: nit: don't pass unexported structs to an exported method. I think you can export componentMetrics for now.

pkg/flow/internal/controller/component.go
marctc: Does it make more sense now?

pkg/flow/internal/controller/component.go
```diff
@@ -91,6 +91,8 @@ type ComponentNode struct {
 
 	exportsMut sync.RWMutex
 	exports    component.Exports // Evaluated exports for the managed component
+
+	cm *ControllerMetrics
```
rfratto, line 95: You can remove this since it's no longer used

rfratto: (You can also unexport ControllerMetrics now I think)

rfratto (approved):
LGTM, nice job!

PR #1959, opened 2022-07-27
Add using environment var for flow

mattdurham:
#### PR Description

Add environment var check for flow in cmd/agent and remove cmd/agentflow. Specifically not adding to changelog/doc since it's incredibly expermental.

#### PR Checklist

- [NA] CHANGELOG updated
- [NA] Documentation added
- [NA] Tests updated


cmd/agent/main.go
```diff
@@ -29,7 +29,8 @@ func init() {
 }
 
 func main() {
-	// If Windows is trying to run us as a service, go through that
+
+	// If Windows is trying to runFlow us as a service, go through that
```
rfratto, line 33: runFlow us? :)

rfratto: Should we leave a TODO to allow Flow to work as a Windows service?

mattdurham: Silly goland rename and me not double checking. Added a todo but likely not supported til flow graduates to full support.

rfratto (approved):
Looks fine, though I think it's worth spending time figuring out what the entrypoint for Flow should actually be:

* What are the HTTP endpoints it exposes? 
* Does it use a command lib like cobra? 

cmd/agentflow was just a temporary solution, and while it's fine to bring that directly into cmd/agent it still needs some love and attention before we release.

PR #1962, opened 2022-07-27
river/token/builder: print keys in sorted order for unordered objects

rfratto:
This change consistently prints object keys which come from unordered objects (i.e., a map[string]V). All other types of River objects have ordered keys, and can be printed in normal order.

### DIFF
--- a/pkg/river/internal/value/value.go
+++ b/pkg/river/internal/value/value.go
@@ -289,6 +289,25 @@ func makeValue(v reflect.Value) Value {
 	return Value{rv: v, ty: riverType}
 }
 
+// OrderedKeys reports if v represents an object with consistently ordered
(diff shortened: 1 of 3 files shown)


pkg/river/internal/value/value.go
```diff
@@ -289,6 +289,25 @@ func makeValue(v reflect.Value) Value {
...
+	switch {
+	case v.rv.Kind() == reflect.Map:
+		// Maps aren't ordered since you can't iterate over their keys in a
+		// consistent order.
+		return false
+
+	default:
+		// Everything else is ordered.
+		return true
+	}
```
rfratto, line 308: I think I'm tired :) Why isn't this a simple return?

tpaschalis (approved):

PR #1965, opened 2022-07-28
flow: Use time.Duration for metrics.scrape config block

tpaschalis:

#### PR Description
The upstream model.Duration type is an alias for time.Duration. 

By using that alias from the decoding level, the `/debug/config` endpoint will report these values as numbers.

```
		scrape_interval  = 60000000000
		scrape_timeout   = 10000000000
```

To make use of the fix on https://github.com/grafana/agent/commit/1fda90adc917c8d107e2e2a0207611adc43222b4 where we added support for time.Duration marshalling as a readable string in River, we have to defer this conversion to a later point in time.

With this change, the durations are reported as expected.
```
		scrape_interval  = "1m0s"
		scrape_timeout   = "10s"
```

No issue filed

#### PR Checklist

- [ ] CHANGELOG updated (N/A)
- [ ] Tests updated (N/A)

### DIFF
--- a/component/metrics/scrape/types.go
+++ b/component/metrics/scrape/types.go
@@ -28,9 +28,9 @@ type Config struct {
 	// A set of query parameters with which the target is scraped.
 	Params url.Values `river:"params,attr,optional"`
 	// How frequently to scrape the targets of this scrape config.
-	ScrapeInterval model.Duration `river:"scrape_interval,attr,optional"`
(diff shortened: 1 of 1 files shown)


rfratto (approved):
LGTM! 

The main problem here is that model.Duration only implements encoding.TextMarshaler for the pointer receiver, and the Arguments/Exports are always rendered as values and not pointers, so the interface doesn't get called.

Using pointers would be a long term fix, but this is more than good enough for now.

PR #1967, opened 2022-07-28
Flow: Add /debug/scope endpoint

rfratto:
/debug/scope prints the variables varaible to be referenced by other
components.

For cmd/agent/example-config.river, the following is printed:

```river
{
  local = {
    file = {
      this_file = {
        content = "Change me!",
      },
    },
  },
  metrics = {
    remote_write = {
      default = {
        receiver = capsule("metrics.Receiver"),
      },
    },
    scrape = {
      default = {},
    },
  },
}
```

This is useful for debugging errors where the scope isn't updated with expected values, something @captncraig and I ran into while testing #1941.

### DIFF
--- a/cmd/agent/flow.go
+++ b/cmd/agent/flow.go
@@ -112,6 +112,7 @@ func runFlow() error {
 		r.Handle("/metrics", promhttp.Handler())
 		r.Handle("/debug/config", f.ConfigHandler())
 		r.Handle("/debug/graph", f.GraphHandler())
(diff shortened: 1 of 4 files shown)


mattdurham:
Seems reasonable for now, though at some point pondering if file/s3 should be binary with a string accessor, but we can cross that bridge when we need to.

mattdurham (approved):
LGTM

PR #1972, opened 2022-07-29
Merge s3 flow component into main

mattdurham:
#### PR Description

This adds S3 flow component support to the flow system. Has a variety of settings, as time goes on these will likely need to be expanded but allows us to run unit tests and hit was S3. Other providers are currently not supported but s3 compatible ones could be added


component/s3/file/s3.go
```diff
@@ -0,0 +1,223 @@
...
+		Exports: Exports{},
+		Build: func(opts component.Options, args component.Arguments) (component.Component, error) {
+			return New(opts, args.(Arguments))
+		},
+	})
+}
+
+// S3 handles reading content from a file located in S3
+type S3 struct {
+	mut        sync.Mutex
```
mattdurham, line 33: Not incredibly happy with the amount of mutex lock/unlocks I have, but not sure if there is a better way

component/s3/file/s3.go
```diff
@@ -0,0 +1,223 @@
...
+		s.cancel()
+	}
+
+	return nil
+}
+
+// Update is called whenever the arguments have changed
+func (s *S3) Update(args component.Arguments) error {
+	newArgs := args.(Arguments)
+	if newArgs.PollFrequency <= time.Second*30 {
```
mattdurham, line 93: Randomly chosen minimum.

component/s3/file/s3.go
```diff
@@ -0,0 +1,223 @@
...
+		s.health.Message = err.Error()
+	}
+	s.health.UpdateTime = time.Now()
+}
+
+func getPathBucketAndFile(path string) (bucket, file string) {
+	parts := strings.Split(path, "/")
+	file = parts[len(parts)-1]
+	bucket = strings.Join(parts[:len(parts)-1], "/")
+	bucket = strings.ReplaceAll(bucket, "s3://", "")
```
mattdurham, line 220: We are only supporting the s3 style accessor; later on probably accept different styles and this would need to change.

component/s3/file/s3_test.go
```diff
@@ -0,0 +1,76 @@
...
+		Options: AWSOptions{
+			Endpoint:     srv.URL,
+			DisableSSL:   true,
+			UsePathStyle: true,
+		},
+	})
+	require.NoError(t, err)
+	ctx, cancel := context.WithCancel(context.Background())
+	go s3File.Run(ctx)
+	time.Sleep(200 * time.Millisecond)
```
mattdurham, line 50: I low-key hate this sleep. Really low poll and sleep values caused it to be inconsistent in passing though.

component/s3/file/s3.go
```diff
@@ -0,0 +1,223 @@
...
+	updateChan chan []byte
+	errorChan  chan error
+	content    string
+	cancel     context.CancelFunc
+}
+
+var (
+	_ component.Component       = (*S3)(nil)
+	_ component.HealthComponent = (*S3)(nil)
+	_ river.Unmarshaler         = (*Arguments)(nil)
```
rfratto, line 47: nit: move this check to where Arguments is being declared

component/s3/file/s3.go
rfratto: IMO validations should be done at the Arguments level during unmarshal so it fails fast. 

(Also I'd make the minimum a global constant since it's a magic number)

component/s3/file/s3.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+// Update is called whenever the arguments have changed
+func (s *S3) Update(args component.Arguments) error {
+	newArgs := args.(Arguments)
+	if newArgs.PollFrequency <= time.Second*30 {
+		return fmt.Errorf("poll_frequency must be greater than 30s")
+	}
+
+	// TODO detect if args are different
```
rfratto, line 97: The controller won't call Update unless the args changes, IIRC. If it doesn't do that, I think it should.

component/s3/file/s3.go
```diff
@@ -0,0 +1,223 @@
...
+			return aws.Credentials{
+				AccessKeyID:     args.Options.AccessKey,
+				SecretAccessKey: string(args.Options.Secret),
+			}, nil
+		})
+		credProvider := aws_config.WithCredentialsProvider(credFunc)
+		configOptions = append(configOptions, credProvider)
+	}
+
+	cfg, err := aws_config.LoadDefaultConfig(context.TODO(), configOptions...)
```
rfratto, line 163: Will this support the environment variables the AWS SDK expects? Or do users have to configure the options block to authenticate?

component/s3/file/s3.go
```diff
@@ -0,0 +1,223 @@
...
+func (s *S3) handleContentPolling(err error) {
+	if err == nil {
+		s.opts.OnStateChange(Exports{
+			Content: rivertypes.OptionalSecret{
+				IsSecret: s.args.IsSecret,
+				Value:    s.content,
+			},
+		})
+		s.health.Health = component.HealthTypeHealthy
+		s.health.Message = "s3 file updated"
```
rfratto, line 208: You're going to need to protect s.health with a mutex since CurrentHealth can be called at any time.

component/s3/file/watcher.go
```diff
@@ -0,0 +1,96 @@
...
+			err = w.download()
+			if err != nil {
+				w.outError <- err
+			}
+			w.mut.Lock()
+			if currentFrequency != w.frequency {
+				currentFrequency = w.frequency
+				dlTick.Reset(currentFrequency)
+			}
+			w.mut.Unlock()
```
rfratto, line 71: nit: make the ticker a field on watcher instead so you can reset the frequency as soon as updateValues is called.

component/s3/file/inout.go
```diff
@@ -0,0 +1,46 @@
+package s3
```
rfratto, line 1: Naming the file inout.go is fine for now but I think we'll want to make things consistent eventually so it's easier to find code.

component/s3/file/inout.go
```diff
@@ -0,0 +1,46 @@
...
+// Arguments implements the input for the s3 component
+type Arguments struct {
+	Path string `river:"path,attr"`
+	// PollFrequency determines the frequency to check for changes
+	// defaults to 10m
+	PollFrequency time.Duration `river:"poll_frequency,attr,optional"`
+	// IsSecret determines if the content should be displayed to the user
+	IsSecret bool `river:"is_secret,attr,optional"`
+	// Options allows you to override default settings
+	Options AWSOptions `river:"options,block,optional"`
```
rfratto, line 18: Having these settings in an options block feels a little strange. Aren't arguments already the options for a component? Should this just be a flat struct instead?

rfratto (commented):
This is coming along nicely! 

(Though the tests failed for me locally when I ran them)

component/s3/file/s3.go
mattdurham: It supports profiles and env var, the AWS SDK handles that under the scenes.

component/s3/file/s3.go
mattdurham: Should be handled by the caller but moving the mutexes around to make it clearer.

component/s3/file/inout.go
mattdurham: How would you feel about naming them `aws_options`? They are overriding default behavior, so I don't want them flattened out.

component/s3/file/inout.go
mattdurham: changed to `types.go` to match scrape, so at least there is a hint of commonality ;)

component/s3/file/inout.go
rfratto: `client` or `client_options`, maybe? I'm not sure about the `aws_` prefix because there's a lot of non-AWS S3 implementations nowadays.

mattdurham: I was thinking any non-aws would have it's own block ie `gcp_options` for custom overrides. TBF I dont know how many of those there actually are or even if we want them in the same component.

rfratto: I'd take a look at Loki's docs to see if its s3 storage backend needs any options that are specific to a vendor (I'd be surprised, because vendor-specific settings would probably not work well in the s3 community)

mattdurham: They use s3, gcs, and azure object storage as separate blocks entirely with their specific settings. Granted they are likely using the specific providers instead of s3 compatible mode. https://grafana.com/docs/loki/latest/configuration/#storage

component/s3/file/s3.go
```diff
@@ -0,0 +1,213 @@
...
+		},
+	})
+}
+
+// S3 handles reading content from a file located in S3
+type S3 struct {
+	mut        sync.Mutex
+	opts       component.Options
+	args       Arguments
+	health     *component.Health
```
rfratto, line 36: nit: this doesn't need to be a pointer

component/s3/file/s3.go
```diff
@@ -0,0 +1,213 @@
...
+	watcher    *watcher
+	updateChan chan []byte
+	errorChan  chan error
+	content    string
+	cancel     context.CancelFunc
```
rfratto, line 41: nit: can you group these with what actually needs to be protected by the mutex and what doesn't need to?

component/s3/file/s3.go
```diff
@@ -0,0 +1,213 @@
...
+	})
+
+	bucket, file := getPathBucketAndFile(args.Path)
+	s := &S3{
+		opts:       o,
+		args:       args,
+		health:     &component.Health{},
+		updateChan: make(chan []byte),
+		errorChan:  make(chan error),
+	}
```
rfratto, line 67: Can this try to do an initial load at construction time so it immediately fails if the file can't be retrieved? (local.file does this)

component/s3/file/watcher.go
```diff
@@ -0,0 +1,88 @@
...
+
+	return &watcher{
+		bucket:     bucket,
+		file:       file,
+		output:     out,
+		outError:   outError,
+		dlTicker:   time.NewTicker(frequency),
+		downloader: downloader,
+	}, nil
+}
```
rfratto, line 40: nit: this never fails, remove the error return

component/s3/file/watcher.go
```diff
@@ -0,0 +1,88 @@
...
+		Key:    aws.String(w.file),
+	})
+	if err != nil {
+		return err
+	}
+	buf := make([]byte, output.ContentLength)
+	_, err = output.Body.Read(buf)
+	if err != nil && err != io.EOF {
+		return err
+	}
```
rfratto, line 85: nit: use `errors.Is(err, io.EOF)`

component/s3/file/watcher.go
```diff
@@ -0,0 +1,88 @@
...
+	w.output <- buf
```
rfratto, line 86: This can deadlock if the component has exited. You need to pass the context to `download` and do 

```
select {
  case <-ctx.Done():
     return nil
  case w.output <- buf:
}
```

so it doesn't hang forever if the component shuts down.

component/s3/file/watcher.go
```diff
@@ -0,0 +1,88 @@
...
+		w.outError <- err
+	}
+	defer w.dlTicker.Stop()
+	for {
+		select {
+		case <-w.dlTicker.C:
+			err = w.download()
+			if err != nil {
+				w.outError <- err
+			}
```
rfratto, line 63: Check for context to be closed here too.

component/s3/file/watcher.go
```diff
@@ -0,0 +1,88 @@
...
+	w.file = file
+	w.dlTicker.Reset(frequency)
+	w.downloader = downloader
+}
+
+func (w *watcher) run(ctx context.Context) {
+	err := w.download()
+	if err != nil {
+		w.outError <- err
+	}
```
rfratto, line 55: Check for context to be closed here too.

component/s3/file/s3.go
```diff
@@ -0,0 +1,213 @@
...
+			},
+		)
+		configOptions = append(configOptions, httpOverride)
+	}
+
+	// check credentials
+	if args.Options.AccessKey != "" {
+		if args.Options.Secret == "" {
+			return nil, fmt.Errorf("if accesskey or secret are specified then the other must also be specified")
+		}
```
rfratto, line 147: Note to self: it'd be great if we could return an error for specifically this field but we can't do that right now, so the error will be shown at the component level.

rfratto (approved):
Nice! Few things that need to be fixed, some nits, but this will be good to go.

component/s3/file/watcher.go
rfratto: Also, small tip for anyone reading who doesn't know this: every time you need to check for the context being closed it _must_ be in the form of

```go
select {
case <-ctx.Done():
  return nil 
case ch <- data:
}
```

and not 

```go
if ctx.Done() != nil {
  return nil 
}
ch <- data 
```

since the latter can still deadlock.

component/s3/file/s3.go
mattdurham: I didn't do it that way since we don't have a parent context, especially given it's going over the network and could take some time.

rfratto: I think that's probably fine to start with, but it does mean other components will start with referencing an empty string until the file gets retrieved.

component/s3/file/watcher.go
mattdurham: Thank you much!

component/s3/file/s3.go
mattdurham: Changed it up so this downloads synchronously on New.

PR #1973, opened 2022-07-29
docs/flow: create skeleton of Flow documentation

rfratto:
This changes docs/flow to have the skeleton including the table of
contents for an initial revision of the Flow documentation.
Documentation already written for existing components has been ported.

Run `make docs` within `docs/` to be able to locally view documentation.

For now, this is not being hosted in docs/sources, so it won't appear in grafana.com/docs/agent. Once we're happy with the documentation, we can move it.


mattdurham (approved):
LGTM

PR #1974, opened 2022-07-29
jsonnet/river: add jsonnet library to build River configs

rfratto:
Add a `jsonnet/river` library which can be used to construct River configs. 

I decided to put this at the top-level rather than in `production/tanka`. Let me know if you think it should be moved.


rfratto:
We should probably move this to an `operations/river-jsonnet` folder instead to align with [Mimir's operations folder](https://github.com/grafana/mimir/tree/main/operations). 

I'm not a fan of production since it implies production readiness, while operations wouldn't.

mattdurham (approved):
LGTM

PR #1975, opened 2022-07-29
Flow: use cobra for the CLI and add an `agent run` command

rfratto:
This changes

```
ENABLE_FLOW_EXPERIMENTAL=true agent --config.file=/path/to/file.river
```

to 

```
ENABLE_FLOW_EXPERIMENTAL=true agent run /path/to/file.river
```

Help text is added to help users understand basic information about Flow. 

Additionally, `ENABLE_FLOW_EXPERIMENTAL=1` is also accepted to enable Flow.

### DIFF
--- a/cmd/agent/flow_run.go
+++ b/cmd/agent/flow_run.go
@@ -43,8 +43,8 @@ River file wasn't specified, can't be loaded, or contains errors, run will exit
 immediately.
 
 run starts an HTTP server which can be used to debug Grafana Agent Flow or
-force it to reload (by POSTing to /-/reload). The listen address can be changed
-through the --server.http.listen-addr flag.
+force it to reload (by sending a GET or POST request to /-/reload). The listen
+address can be changed through the --server.http.listen-addr flag.
 
 By default, the HTTP server exposes the following debug endpoints:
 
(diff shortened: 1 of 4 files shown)


cmd/agent/flow_run.go
```diff
@@ -0,0 +1,211 @@
...
+		Short: "Run Grafana Agent Flow",
+		Long: `The run subcommand runs Grafana Agent Flow in the foreground until an interrupt
+is received.
+
+run must be provided an argument pointing at the River file to use. If the
+River file wasn't specified, can't be loaded, or contains errors, run will exit
+immediately.
+
+run starts an HTTP server which can be used to debug Grafana Agent Flow or
+force it to reload (by POSTing to /-/reload). The listen address can be changed
```
tpaschalis, line 45: Nitpick, but we don't currently enforce the request to be a POST, which I think is okay.

tpaschalis (approved):
LGTM, looks pretty straightforward I don't have much to add 🎉

docs/flow/reference/cli/run.md
```diff
@@ -6,3 +6,29 @@ weight: 100
 ---
 
 # `agent run` command
+
+The `agent run` command runs Grafana Agent Flow in the foreground until an
```
tpaschalis, line 10: Do we want to mention the `EXPERIMENTAL_ENABLE_FLOW` flag here, or leave it just in the code for now?

rfratto: All of the Flow docs only apply when `EXPERIMENTAL_ENABLE_FLOW` is set so I think we'll mention it somewhere higher-level than this.

mattdurham (approved):
LGTM

PR #1978, opened 2022-08-01
Minor changes to remote.s3 documentation

rfratto:
1. Fix file path (should be in reference/components)
2. File file name (should be remote.s3.md)
3. Fix sentences missing a period
4. Standardize sections from other reference docs
5. Minor rewording

### DIFF
--- a/docs/flow/reference/components/remote.s3.md
+++ b/docs/flow/reference/components/remote.s3.md
@@ -1,14 +1,25 @@
+---
+aliases:
+- /docs/agent/latest/flow/reference/components/remote.s3
+title: remote.s3
+---
+
 # remote.s3
 
-The `remote.s3` component exposes the contents of a file located in an S3 compatible system to other components. The file will be polled for changes so that the most recent content is always available.
+`remote.s3` exposes the contents of a file located in an S3 compatible system
+to other components. The file will be polled for changes so that the most
+recent content is always available.
 
-Multiple `remote.s3` components can be specified by giving them different name labels. By default the component will use AWS environment vars or profiles that are setup. If those are not setup or overrides are required then `key` and `secret` can be used.
+Multiple `remote.s3` components can be specified by giving them different name
+labels. By default, AWS environment variables are used to authenticate against
+S3. The `key` and `secret` arguments can be used to provide custom
+authentication.
 
 ## Example
 
 ```river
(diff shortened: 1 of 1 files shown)


docs/flow/reference/components/remote.s3.md
```diff
@@ -18,12 +29,14 @@ The following arguments are supported:
 
 Name | Type | Description                                                             | Default | Required
 ---- | ---- |-------------------------------------------------------------------------| ------- | --------
-`path` | `string` | Path in the format of `"s3://bucket/file"`                              | | **yes**
+`path` | `string` | Path in the format of `"s3://bucket/file"` | | **yes**
 `poll_frequency` | `duration` | How often to poll the file for changes, must be greater than 30 seconds | `"10m"` | no
-`is_secret` | `bool` | Marks the file as containing a [secret][]                               | `false` | no
-`client_options` | `client_options` | Additional [options](#client_options) for configuring the client      | | no
```
rfratto, line 24: Historically we haven't been documenting blocks as arguments. This might change, but I'd like to remove it to make it consistent with the other docs for now.

mattdurham: How would we represent nested blocks? IE if client_options also had an `http_config`?

rfratto: Subheaders, maybe? That's a good point, though. Maybe we do something like this below the table: 

```
The following inner blocks can be specified:

* [`client_options`](#client_options): Additional options for configuring the client 
```

WDYT?

rfratto: Maybe a second table?

The issue IMO is that they _are_ different things: attributes are typed values, blocks are structural elements.

mattdurham: I am down for a second table. I am unsure if they are different from a user's perspective. They are config. We can go with two tables and see if we get any feedback.

rfratto: We'll have to try to teach the users to see them as different, because one supports expressions and the other doesn't :)

rfratto: PTAL @mattdurham

mattdurham (approved):
LGTM

PR #1982, opened 2022-08-02
Add component metrics to /metrics endpoint

### LINKED ISSUE #1938: Flow: allow components to expose metrics
Components should be allowed to expose metrics to be collected from the centralized /metrics endpoint. This is most important for the metrics.remote_write component, since we'll want to collect remote_write stats. Metrics from components should always have a label injected which corresponds to the component ID of that component. 

The Flow prototype did this by having an extension interface where components could be Prometheus Collectors, but it was hacky and there wasn't a lot of thought put into it. If there's another solution that's cleaner and makes more sense, go from it.

mattdurham:
#### PR Description

Add support for passing prometheus register to components and wrapping them with the component id. I used `component_id` since I thought `id` might be used elsewhere since it is so generic. Adds ready endpoint that is triggered from the initial load/reload.

Closes #1938

#### PR Checklist

- [X] Tests updated


cmd/agent/flow.go
```diff
@@ -119,6 +129,7 @@ func runFlow() error {
 
 		r.HandleFunc("/-/reload", func(w http.ResponseWriter, _ *http.Request) {
 			err := reload()
+			ready = err == nil
```
rfratto, line 132: This would need a mutex to protect ready or to use an atomic bool. Can we just have it print "ready" for now instead of trying to slowly add in health checks?

cmd/agent/flow.go
```diff
@@ -43,6 +43,7 @@ func runFlow() error {
 		httpListenAddr = "127.0.0.1:12345"
 		configFile     string
 		storagePath    = "data-agent/"
+		ready          = true
```
rfratto, line 46: Wouldn't this be false by default?

component/local/file/file.go
```diff
@@ -136,6 +147,7 @@ func (c *Component) Run(ctx context.Context) error {
 	for {
 		select {
 		case <-ctx.Done():
+			c.opts.Registerer.Unregister(c.lastAccessed)
```
rfratto, line 150: This isn't guaranteed to be called if the component never starts up (ie because construction failed)

pkg/flow/internal/controller/component.go
```diff
@@ -148,6 +150,9 @@ func getManagedOptions(globals ComponentGlobals, cn *ComponentNode) component.Op
 		Logger:        log.With(globals.Logger, "component", cn.nodeID),
 		DataPath:      filepath.Join(globals.DataPath, cn.nodeID),
 		OnStateChange: cn.setExports,
+		Registerer: prometheus.WrapRegistererWith(prometheus.Labels{
+			"component_id": cn.nodeID,
+		}, globals.Registerer),
```
rfratto, line 155: Generally I feel like passing the registerer along to components is going to be a common source of bugs, where metrics start leaking because the component implementation didn't unregister them consistently. 

I would much rather us pass a custom implementation of `Registerer` which implements Collector and is invoked by the controller when collecting metrics instead of actually registering metrics globally, so that way components don't need to unregister anything.

rfratto (changes requested):

cmd/agent/flow.go
mattdurham: The http listener is only available once the config has been loaded successfully. We could start as false but would have to add an additional line of code to flip it to true after reload().

cmd/agent/flow.go
mattdurham: Will use atomic bool.

component/local/file/file.go
```diff
@@ -98,11 +101,19 @@ func New(o component.Options, args Arguments) (*Component, error) {
 		opts: o,
 
 		reloadCh: make(chan struct{}, 1),
+		lastAccessed: prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "local_file_timestamp_last_accessed_unix_seconds",
```
marctc, line 105: Don't we want this with `agent` or `agent_component` namespace?

mattdurham: `agent_` sounds reasonable

component/metrics/mutate/mutate.go
```diff
@@ -57,9 +58,17 @@ func New(o component.Options, args Arguments) (*Component, error) {
 	c := &Component{opts: o}
 	c.appendable = fa.NewFlowAppendable(args.ForwardTo...)
 	c.receiver = &metrics.Receiver{Receive: c.Receive}
+	c.metricsProcessed = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "metrics_mutate_metrics_processed",
```
marctc, line 62: Same here :smile:

pkg/flow/internal/controller/component.go
mattdurham: Added a custom registerer that handles unregistering the component and any children objects.

component/metrics/remotewrite/remote_write.go
```diff
@@ -85,7 +85,7 @@ type BasicAuthConfig struct {
 type Component struct {
 	log  log.Logger
 	opts component.Options
-	reg  *metrics.CollectorRegistry
+	reg  prometheus.Registerer
```
rfratto, line 88: This field is never read now, we can remove it.

component/registry.go
```diff
@@ -49,6 +50,9 @@ type Options struct {
 	// by the component; a component must use the same Exports type for its
 	// lifetime.
 	OnStateChange func(e Exports)
+
+	//Registerer allows components to add their own metrics. The register will come pre-wrapped with the component id
+	Registerer Registerer
```
rfratto, line 55: Components shouldn't need to call `RegisterComponent` or `UnregisterComponent`, this should just be the normal prometheus.Registerer interface.

component/metrics/scrape/scrape_test.go
```diff
@@ -16,7 +18,10 @@ import (
 func TestForwardingToAppendable(t *testing.T) {
 	l, err := logging.New(os.Stderr, logging.DefaultOptions)
 	require.NoError(t, err)
-	opts := component.Options{Logger: l}
+	opts := component.Options{
+		Logger:     l,
+		Registerer: &fakeRegisterer{},
```
rfratto, line 23: Once you're using the normal prometheus.Registerer interface, this can be `Registerer: prometheus.NewRegistry()` to avoid having to make a fake impl.

pkg/flow/internal/controller/component.go
```diff
@@ -272,6 +277,7 @@ func (cn *ComponentNode) Run(ctx context.Context) error {
 	}
 
 	cn.setRunHealth(component.HealthTypeHealthy, "started component")
+	defer cn.managedOpts.Registerer.UnregisterComponent()
```
rfratto, line 280: This has the same leaking problem as before (Run won't always get called even after metrics get registered), but now it's at the controller level.

I still think we need to implement a collector which iterates over existing ComponentNodes and collects from their registries similar to the [collector Marc wrote](https://github.com/grafana/agent/blob/1bc903abb68d4889a5742e69ae937f5e6a0a56b2/pkg/flow/internal/controller/metrics.go#L38-L70).

component/registry.go
mattdurham: Tradeoff is the controller will need to do some casting and it's a bit more boilerplate in the components themselves. I could use MustRegister to register multiple metrics but I don't think we should panic in this case.

rfratto: I think it's probably fine to panic IMO, you'd really want to know fast if you registered invalid metrics, otherwise it'll just be hidden and you won't know why it's not showing up

marctc:
I was wondering if instead to have an specific metric per component we could have generic metrics (accessed, errors, etc) with a label as component name. That would give us the ability to easily add metrics to new components without having to create new metrics per component, like in this ongoing [PR](https://github.com/grafana/agent/pull/1872/files)

mattdurham:
There is probably room for some generic cases: `errors` springs to mind. Could be once we have a stable of components see which ones have high commonality.

component/registry.go
```diff
@@ -49,6 +50,9 @@ type Options struct {
 	// by the component; a component must use the same Exports type for its
 	// lifetime.
 	OnStateChange func(e Exports)
+
+	//Registerer allows components to add their own metrics. The register will come pre-wrapped with the component id
```
rfratto, line 54:
```suggestion
	// Registerer allows components to add their own metrics. The register will come pre-wrapped with the component ID. It is not necessary for components to unregister metrics on shutdown.
```

pkg/flow/internal/controller/metrics.go
```diff
@@ -67,4 +68,7 @@ func (cc *controllerCollector) Collect(ch chan<- prometheus.Metric) {
 
 func (cc *controllerCollector) Describe(ch chan<- *prometheus.Desc) {
 	ch <- cc.runningComponentsTotal
+	for _, component := range cc.l.components {
+		component.register.Describe(ch)
+	}
```
rfratto, line 73: We shouldn't do this: Describe is only called once and will permanently register checked metrics. 

We should leave metrics from components as unchecked and remove this.

(Also, you needed to do cc.l.Components() to prevent a race condition)

pkg/flow/internal/controller/wrap_registerer.go
```diff
@@ -0,0 +1,65 @@
...
+
+// newWrappedRegisterer creates a wrapped register
+func newWrappedRegisterer() *wrappedRegisterer {
+	return &wrappedRegisterer{internalCollectors: make(map[prometheus.Collector]struct{})}
+}
+
+// Describe implements the interface
+func (w *wrappedRegisterer) Describe(descs chan<- *prometheus.Desc) {
+	w.mut.Lock()
+	defer w.mut.Unlock()
```
rfratto, line 22: nit: Can you change this to an RLock and RUnlock? Ditto in Collect.

pkg/flow/internal/controller/component.go
mattdurham: Added a custom wrapper that feeds to the collector and feeds to the prometheus registry.

rfratto (approved):
🎉 Nice! 

I tested it out with a few reloads after removing / adding components and everything looks clean.

PR #1989, opened 2022-08-04
operations/river-jsonnet: add jsonnetfile.json

rfratto:
This allows river-jsonnet to be versioned separately from other dependencies.

Created using `jb init`.


mattdurham (approved):
LGTM

PR #1991, opened 2022-08-04
Add client_config block to remote_write

mattdurham:
#### PR Description
Add http_client_config to remote write flow

### DIFF
--- a/component/common/config/types.go
+++ b/component/common/config/types.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"net/url"
 
+	"github.com/grafana/agent/pkg/river"
+
 	"github.com/grafana/agent/pkg/flow/rivertypes"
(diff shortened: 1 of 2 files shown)


rfratto:
Can you also make sure the client config in metrics.scrape is used? Right now it's exposed but not hooked in: https://github.com/grafana/agent/blob/87fb3779723f319b1f1dc4b0541a4479043874c3/component/metrics/scrape/scrape.go#L43

rfratto:
> Can you also make sure the client config in metrics.scrape is used? Right now it's exposed in the arguments but not hooked in:

My bad, it is being used, the comment is just stale :)

rfratto (approved):

mattdurham:
I think the comment refers to the dialerfunc so should be good.

PR #1994, opened 2022-08-04
metrics.remote_write: use correct storage.Appendable when writing metrics

rfratto:
The fanout appender should be used for appending metrics so the prometheus_remote_storage_highest_timestamp_in_seconds metric gets updated.

Also fixes the `agent_component_evaluation_seconds` histogram and the `agent_component_controller_evaluating` metrics to track whether a runtime reevaluation is ongoing, not just graph loads.


mattdurham (approved):
Lgtm

PR #1996, opened 2022-08-08
flow: evaluate everything in queue when queue changes

rfratto:
This fixes a bug where only a single changed component was reevaluated, even if other components were in the queue. This potentially caused some components to never get evaluated, depending on the luck of the draw from what gets dequeued.


gouthamve (approved):
Not the most familiar with this codebase, but I've gone through the `controller.Queue` implementation and this makes sense.

PR #2008, opened 2022-08-08
vSphere integration

rlankfo:
#### PR Description
This PR introduces a v2 integration for the vmware_exporter (vsphere)

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/configuration/integrations/integrations-next/vsphere-config.md
+++ b/docs/sources/configuration/integrations/integrations-next/vsphere-config.md
@@ -30,8 +30,9 @@ Configuration reference:
     [scrape_interval: <duration> | default = <metrics.global.scrape_interval>]
     [scrape_timeout: <duration> | default = <metrics.global.scrape_timeout>]
 
-  # Integration instance name
-  [instance: <string>]
+  # Integration instance name. This will default to the host:port of the configured
+  # vsphere_url.
+  [instance: <string> | default = <vsphere_url>]
 
   # Number of managed objects to include in each request to vsphere when
   # fetching performance counters.
@@ -48,13 +49,13 @@ Configuration reference:
   [enable_exporter_metrics: <boolean> | default = true]
 
   # The url of the vCenter SDK endpoint
-  [vsphere_url: <string>]
+  vsphere_url: <string>
 
   # vCenter username
-  [vsphere_user: <string>]
+  vsphere_user: <string>
 
   # vCenter password
-  [vsphere_password: <string>]
+  vsphere_password: <string>
 
 ```
 
--- a/docs/sources/configuration/integrations/integrations-next/vsphere-config.md
+++ b/docs/sources/configuration/integrations/integrations-next/vsphere-config.md
@@ -10,6 +10,54 @@ The `vsphere_config` block configures the `vmware_exporter` integration, an embe
 version of [`vmware_exporter`](https://github.com/grafana/vmware_exporter), configured
 to collect vSphere metrics. This integration is considered beta.
 
+Configuration reference:
+
+```yaml
+  autoscrape:
+    # Enables autoscrape of integrations.
+    [enable: <boolean> | default = true]
+
+    # Specifies the metrics instance name to send metrics to. Instance
+    # names are located at metrics.configs[].name from the top-level config.
+    # The instance must exist.
+    #
+    # As it is common to use the name "default" for your primary instance,
+    # we assume the same here.
+    [metrics_instance: <string> | default = "default"]
+
+    # Autoscrape interval and timeout. Defaults are inherited from the global
+    # section of the top-level metrics config.
+    [scrape_interval: <duration> | default = <metrics.global.scrape_interval>]
+    [scrape_timeout: <duration> | default = <metrics.global.scrape_timeout>]
+
+  # Integration instance name
+  [instance: <string>]
+
+  # Number of managed objects to include in each request to vsphere when
+  # fetching performance counters.
+  [request_chunk_size: <int> | default = 256]
+
+  # Number of concurrent requests to vsphere when fetching performance counters.
+  [collect_concurrency: <int> | default = 8]
+
+  # Interval on which to run vsphere managed object discovery. Setting this to a
+  # non-zero value will result in object discovery running in the background. Each
+  # scrape will use object data gathered during the last discovery.
+  # When this value is 0, object discovery occurs per scrape.
+  [discovery_interval: <duration> | default = 0]
+  [enable_exporter_metrics: <boolean> | default = true]
+
+  # The url of the vCenter SDK endpoint
+  [vsphere_url: <string>]
+
+  # vCenter username
+  [vsphere_user: <string>]
+
+  # vCenter password
+  [vsphere_password: <string>]
+
+```
+
 ## Quick configuration example
 
 ```yaml
@@ -18,7 +66,7 @@ integrations:
     - vsphere_url: https://127.0.0.1:8989/sdk
       vsphere_user: user
       vsphere_password: pass
-      chunk_size: 256
+      request_chunk_size: 256
       collect_concurrency: 8
       instance: vsphere
       autoscrape:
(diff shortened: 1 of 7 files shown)


docs/sources/configuration/integrations/integrations-next/vsphere-config.md
```diff
@@ -0,0 +1,32 @@
+---
+aliases:
+  - /docs/agent/latest/configuration/integrations/integrations-next/vsphere-config/
+title: vsphere_config
+---
+
+# vsphere config (beta)
```
rfratto, line 7: We should definitely document the schema for the block before merging IMO

pkg/integrations/v2/vmware_exporter/vmware_exporter.go
```diff
@@ -0,0 +1,92 @@
...
+
+var DefaultConfig = Config{
+	ChunkSize:               256,
+	CollectConcurrency:      8,
+	ObjectDiscoveryInterval: 0,
+	EnableExporterMetrics:   true,
+}
+
+type Config struct {
+	ChunkSize               int                  `yaml:"chunk_size,omitempty"`
```
mattdurham, line 28: Add unit to name/yaml name?

docs/sources/configuration/integrations/integrations-next/vsphere-config.md
```diff
@@ -0,0 +1,80 @@
...
+    # As it is common to use the name "default" for your primary instance,
+    # we assume the same here.
+    [metrics_instance: <string> | default = "default"]
+
+    # Autoscrape interval and timeout. Defaults are inherited from the global
+    # section of the top-level metrics config.
+    [scrape_interval: <duration> | default = <metrics.global.scrape_interval>]
+    [scrape_timeout: <duration> | default = <metrics.global.scrape_timeout>]
+
+  # Integration instance name
```
rfratto, line 33: Can you document that this defaults to the vSphere URL if unset?

docs/sources/configuration/integrations/integrations-next/vsphere-config.md
```diff
@@ -0,0 +1,80 @@
...
+  [enable_exporter_metrics: <boolean> | default = true]
+
+  # The url of the vCenter SDK endpoint
+  [vsphere_url: <string>]
+
+  # vCenter username
+  [vsphere_user: <string>]
+
+  # vCenter password
+  [vsphere_password: <string>]
```
rfratto, line 57: Are any of these required? If so, drop the surrounding brackets for required options

docs/sources/configuration/integrations/integrations-next/vsphere-config.md
```diff
@@ -0,0 +1,80 @@
...
+
+```yaml
+integrations:
+  vsphere_configs:
+    - vsphere_url: https://127.0.0.1:8989/sdk
+      vsphere_user: user
+      vsphere_password: pass
+      request_chunk_size: 256
+      collect_concurrency: 8
+      instance: vsphere
```
rfratto, line 71: The example should probably have a `vsphere` metrics instance

pkg/integrations/v2/vmware_exporter/vmware_exporter.go
```diff
@@ -0,0 +1,99 @@
...
+
+// ApplyDefaults applies the integration's default configuration.
+func (c *Config) ApplyDefaults(g integrations.Globals) error {
+	c.Common.ApplyDefaults(g.SubsystemOpts.Metrics.Autoscrape)
+	return nil
+}
+
+// Identifier returns a string that identifies the instance of the integration.
+func (c *Config) Identifier(g integrations.Globals) (string, error) {
+	return *c.Common.InstanceKey, nil
```
rfratto, line 63: This isn't always set, right? It seems like this could panic

rfratto: Ah. I see you copied this from other v2 integrations, but they're also doing the wrong thing. 

This should be:

```go
if c.Common.InstanceKey != nil {
  return *c.Common.InstanceKey, nil 
} 

u, err := url.Parse(c.VSphereURL)
if err != nil {
  return "", err
}
return fmt.Sprintf("%s:%s", u.Hostname(), u.Port()), nil
```

I'm not sure why other v2 integrations are implementing the InstanceKey method; that's a v1 integration thing and isn't used in v2.

pkg/integrations/v2/vmware_exporter/vmware_exporter.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// InstanceKey returns the vsphere instance configured.
+func (c *Config) InstanceKey(agentKey string) (string, error) {
```
rfratto, line 67: see above comment: this is never called in v2 integrations; move logic to Identifier instead.

pkg/integrations/v2/vmware_exporter/vmware_exporter.go
```diff
@@ -0,0 +1,99 @@
...
+// NewIntegration constructs a new instance of this integration.
+func (c *Config) NewIntegration(log log.Logger, g integrations.Globals) (integrations.Integration, error) {
+	vsphereURL, err := url.Parse(c.VSphereURL)
+	if err != nil {
+		return nil, err
+	}
+	vsphereURL.User = url.UserPassword(c.VSphereUser, string(c.VSpherePass))
+
+	exporterConfig := vsphere.Config{
+		TelemetryPath:           "/integrations/vsphere/metrics",
```
rfratto, line 84: What does this do? The full URL of an integration isn't consistent; it's based on the integration instance identifier a user supplies or the default autogenerated identifier.

docs/sources/configuration/integrations/integrations-next/vsphere-config.md
rfratto: Oops, sorry, I was mistaking this for autoscrape.metrics_instance. Never mind :)

pkg/integrations/v2/vmware_exporter/vmware_exporter.go
rfratto: It did in v1, but in v2 there might be many instances of the same integration so the unique instance key is used instead. 

It might be worth finding a way bypass the mux to avoid any potential bugs with this IMO

pkg/integrations/v2/vmware_exporter/vmware_exporter.go
rlankfo: Thanks 🙏  I was actually wondering about this when I initially did it. I wonder if some other v2 integrations do this because they also register as legacy integration?

mattdurham:
How do the EnableExporterMetrics get shared?

rlankfo:
> How do the EnableExporterMetrics get shared?

They are just additional metrics exposed on the /metrics endpoint of the metrics handler.

pkg/integrations/v2/vmware_exporter/vmware_exporter.go
```diff
@@ -0,0 +1,97 @@
+package vmware_exporter
```
marctc, line 1: nit: consistency between package/file name and integration name?

marctc (approved):
LGTM. Did you test this with real environment? That would be exciting to see!

pkg/integrations/v2/vmware_exporter/vmware_exporter.go
rlankfo: @marctc I named the package after the project but we could change this if necessary. My thinking was that eventually the exporter could container collectors for things besides vSphere, such as NSX.

rlankfo:
> LGTM. Did you test this with real environment? That would be exciting to see!

Yes, I tested against a live VMC environment as well as the vcsim simulator.

pkg/integrations/v2/vmware_exporter/vmware_exporter.go
marctc: Gotcha, thanks

mattdurham (approved):
Awesome, excited to see it in action

PR #2009, opened 2022-08-08
river: reuse memory when decoding a map

rfratto:
Benchmark results:

```
ObjectDecode-10    5.83µs ± 2%    5.68µs ± 2%   -2.54%  (p=0.000 n=10+10)

ObjectDecode-10    3.97kB ± 0%    3.67kB ± 0%   -7.66%  (p=0.000 n=9+10)

ObjectDecode-10       125 ± 0%       106 ± 0%  -15.20%  (p=0.000 n=10+10)
```

This gives a tiny performance boost for decoding map values, but it creates far less garbage.

Other low-hanging fruit optimizations are possible, but I wanted to start with this one.


mattdurham (approved):
LGTM

PR #2011, opened 2022-08-08
Remove any port number from the incoming connection address for PromSD processing

### LINKED ISSUE #1984: Matching of sending IP address to find pod info should strip the port number first
I'm using the Grafana Agent, version 0.26.1, configured using the tk/jsonnet configuration alternative.  My configuration file includes this to enable translation of the sending IP address into pod information:

```yaml
scrape_configs:
  - bearer_token_file: /var/run/secrets/kubernetes.io/serviceaccount/token
    job_name: kubernetes-pods
    kubernetes_sd_configs:
      - role: pod
    relabel_configs:
      - action: replace
        source_labels:
          - __meta_kubernetes_namespace
        target_label: namespace
      - action: replace
        source_labels:
          - __meta_kubernetes_pod_name
        target_label: pod
      - action: replace
        source_labels:
        target_label: container
    tls_config:
      ca_file: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
      insecure_skip_verify: false
```

However, I see no pod information in the captured traces.  After enabling debug-level logging in the Grafana Agent, I see this error message when the trace is received:

```
ts=2022-08-03T13:07:49.133188793Z caller=prom_sd_processor.go:166 level=debug component="traces service disco" msg="unable to find matching hostLabels" ip=172.30.114.52:54443
```

I believe the Grafana Agent knows about the IP address because I see messages like this earlier in the log:
```
ts=2022-08-03T13:07:26.142710099Z caller=prom_sd_processor.go:284 level=debug component="traces service disco" msg="adding host to hostLabels" host=172.30.114.52
```

I believe the presence of the port number in the string being looked up is causing the failure to find the match.  The port number should be stripped before doing the lookup.

jphx:
#### PR Description

When the PromSD processor is using the `connection`  method to deduce the IP address of the client sending the span, remove any port number from the address string before matching against known IP addresses.

Fixes #1984

#### Notes to the Reviewer

I tested this change in the scenario where my trace spans were not getting any pod information attached.  After this change, there are `pod` and `container` labels in the `Process` section of the spans.

#### PR Checklist

- [X] CHANGELOG updated
- [ ] Documentation added
- [X] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,9 @@ Main (unreleased)
 
 ### Bugfixes
 
+- Tracing: Fixed issue with the PromSD processor using the `connection` method to discover the IP
+  address.  It was failing to match because the port number was included in the address string. (@jphx)
+
 ### Other changes
 
  - Update several go dependencies to resolve warnings from certain security scanning tools. None of the resolved vulnerabilities were known to be exploitable through the agent. (@captncraig)
--- a/pkg/traces/promsdprocessor/prom_sd_processor.go
+++ b/pkg/traces/promsdprocessor/prom_sd_processor.go
@@ -118,13 +118,25 @@ func stringAttributeFromMap(attrs pcommon.Map, key string) string {
 	return ""
 }
 
-func getConnectionIP(ctx context.Context) string {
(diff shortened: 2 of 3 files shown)


rfratto (approved):
LGTM, thank you for fixing this!

PR #2012, opened 2022-08-08
Fix short-circuiting of scope and add name

### LINKED ISSUE #1999: Flow: Failure for scope to be populated if a component fails to load

mattdurham:
#### PR Description

This ensures the scope is filled in a better fashion. There are still edge cases such as component naming issues.

Closes #1999

### DIFF
--- a/pkg/flow/internal/controller/loader.go
+++ b/pkg/flow/internal/controller/loader.go
@@ -191,7 +191,7 @@ func (l *Loader) wireGraphEdges(g *dag.Graph) diag.Diagnostics {
 // Variables returns the Variables the Loader exposes for other Flow components
 // to reference.
 func (l *Loader) Variables() map[string]interface{} {
-	return l.cache.BuildContext(nil).Variables
+	return l.cache.BuildContext(nil, "").Variables
 }
 
 // Components returns the current set of loaded components.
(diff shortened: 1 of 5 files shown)


pkg/river/vm/vm.go
```diff
@@ -506,6 +506,9 @@ func (vm *Evaluator) evaluateExpr(scope *Scope, assoc map[value.Value]ast.Node,
 
 // A Scope exposes a set of variables available to use during evaluation.
 type Scope struct {
+	// Name is the name of this particular scope/block
+	Name string
```
rfratto, line 510: Sorry, why do we want to name the scopes?

mattdurham: It's mostly debugging, to figure out which one we are on.

rfratto: I'm not sure about this IMO, every component effectively has the same scope 🤔

mattdurham: Isn't each component a separate scope?

rfratto: No, the scope is the object-ized form of the entire graph, each component gets a copy of it at evaluation time.

mattdurham: That's true, it was helpful when I was trying to dig through the code to see what part of the evaluation I was at but I can walk the stack to see that.

pkg/flow/internal/controller/loader.go
```diff
@@ -268,14 +268,15 @@ func (l *Loader) EvaluateDependencies(parentScope *vm.Scope, c *ComponentNode) {
...
 func (l *Loader) evaluate(parent *vm.Scope, c *ComponentNode) error {
 	ectx := l.cache.BuildContext(parent)
-	if err := c.Evaluate(ectx); err != nil {
-		level.Error(l.log).Log("msg", "failed to evaluate component", "component", c.NodeID(), "err", err)
-		return err
-	}
+	err := c.Evaluate(ectx)
 	// Always update the cache both the arguments and exports, since both might
-	// change when a component gets re-evaluated.
+	// change when a component gets re-evaluated. We also want to cache the arguments and exports in case of an error
```
rfratto, line 273: tiniest nit: end comments in periods please :) I'd prefer to treat comments as if they were public-facing documentation.

rfratto (approved):
🎉 Nice, LGTM. 

Should we have a test for this?

PR #2014, opened 2022-08-08
Introduce agent-flow-mixin

rfratto:
This introduces an agent-flow-mixin for users to visualize and alert on the behavior of Grafana Agent Flow.

This first commit introduces a prototype of a controller dashboard that can be used for monitoring Flow controller information. A grizzly.jsonnet file has been included for locally prototyping the mixin.

The code isn't very pretty, but hopefully maintainable enough.

Future PRs will be used to add alerts and other dashboards.

image


mattdurham (approved):
LGTM

PR #2015, opened 2022-08-09
Make FlowMetric mostly immutable

mattdurham:
#### PR Description

Make FlowMetrics semi-immutable, before a flowmetric could be easily changed. There could be more memory alloc/gced due to LabelsCopy() but that is likely unavoidable.

### DIFF
--- a/component/common/appendable/appendable.go
+++ b/component/common/appendable/appendable.go
@@ -63,17 +63,15 @@ func (app *flowAppender) Append(ref storage.SeriesRef, l labels.Labels, t int64,
 		set := make([]*metrics.FlowMetric, 0)
 		app.buffer[t] = set
 	}
-	// If ref is 0 then lets grab a global id
-	if ref == 0 {
-		ref = storage.SeriesRef(metrics.GlobalRefMapping.GetGlobalRefIDByLabels(l))
-	}
-	// If it is stale then we can remove it
+	// The incoming ref is a global refid.
+    fm := metrics.NewFlowMetric(metrics.RefID(ref), l, v)
+	// If it is stale then we can remove it.
 	if value.IsStaleNaN(v) {
-		metrics.GlobalRefMapping.AddStaleMarker(uint64(ref), l)
+		metrics.GlobalRefMapping.AddStaleMarker(fm.GlobalRefID(), l)
 	} else {
-		metrics.GlobalRefMapping.RemoveStaleMarker(uint64(ref))
+		metrics.GlobalRefMapping.RemoveStaleMarker(fm.GlobalRefID())
 	}
-	app.buffer[t] = append(app.buffer[t], metrics.NewFlowMetric(uint64(ref), l, v))
+	app.buffer[t] = append(app.buffer[t],fm)
 	return ref, nil
 }
 
--- a/component/common/appendable/appendable.go
+++ b/component/common/appendable/appendable.go
@@ -11,12 +11,6 @@ import (
 	"github.com/prometheus/prometheus/storage"
 )
 
-// FlowMetric is a wrapper around a single sample without the timestamp.
-type FlowMetric struct {
-	Labels labels.Labels
-	Value  float64
-}
-
 // FlowAppendable is a flow-specific implementation of an Appender.
 type FlowAppendable struct {
 	mut       sync.RWMutex
--- a/component/common/appendable/appendable.go
+++ b/component/common/appendable/appendable.go
@@ -63,7 +63,8 @@ func (app *flowAppender) Append(ref storage.SeriesRef, l labels.Labels, t int64,
 		set := make([]*metrics.FlowMetric, 0)
 		app.buffer[t] = set
 	}
-	// The incoming ref is a global refid.
+	// The incoming ref is a global refid. The refid might be 0, and in that case assigning it to a flow metric
+	// will ensure that it is not 0.
     fm := metrics.NewFlowMetric(metrics.RefID(ref), l, v)
 	// If it is stale then we can remove it.
 	if value.IsStaleNaN(v) {
--- a/component/common/appendable/appendable.go
+++ b/component/common/appendable/appendable.go
@@ -88,10 +88,11 @@ func (app *flowAppender) AppendExemplar(
 func (app *flowAppender) Commit() error {
 	for _, r := range app.receivers {
 		for ts, metrics := range app.buffer {
-			if r == nil || r.Receive == nil {
+			if r == nil {
 				continue
 			}
-			r.Receive(ts, metrics)
+
+			r.Send(ts, metrics)
 		}
 	}
 	app.buffer = make(map[int64][]*metrics.FlowMetric)
--- a/component/common/appendable/appendable.go
+++ b/component/common/appendable/appendable.go
@@ -81,7 +81,7 @@ func (app *flowAppender) AppendExemplar(
 	l labels.Labels,
 	e exemplar.Exemplar,
 ) (storage.SeriesRef, error) {
-
+	// TODO actually implement this before flow production.
 	return 0, nil
 }
 
--- a/component/common/appendable/appendable.go
+++ b/component/common/appendable/appendable.go
@@ -65,18 +65,23 @@ func (app *flowAppender) Append(ref storage.SeriesRef, l labels.Labels, t int64,
 	}
 	// The incoming ref is a global refid. The refid might be 0, and in that case assigning it to a flow metric
 	// will ensure that it is not 0.
-    fm := metrics.NewFlowMetric(metrics.RefID(ref), l, v)
+	fm := metrics.NewFlowMetric(metrics.RefID(ref), l, v)
 	// If it is stale then we can remove it.
 	if value.IsStaleNaN(v) {
 		metrics.GlobalRefMapping.AddStaleMarker(fm.GlobalRefID(), l)
 	} else {
 		metrics.GlobalRefMapping.RemoveStaleMarker(fm.GlobalRefID())
 	}
-	app.buffer[t] = append(app.buffer[t],fm)
+	app.buffer[t] = append(app.buffer[t], fm)
 	return ref, nil
 }
 
-func (app *flowAppender) AppendExemplar(ref storage.SeriesRef, l labels.Labels, e exemplar.Exemplar) (storage.SeriesRef, error) {
+func (app *flowAppender) AppendExemplar(
+	ref storage.SeriesRef,
+	l labels.Labels,
+	e exemplar.Exemplar,
+) (storage.SeriesRef, error) {
+
 	return 0, nil
 }
 
--- a/component/common/appendable/appendable.go
+++ b/component/common/appendable/appendable.go
@@ -79,11 +79,7 @@ func (app *flowAppender) Append(ref storage.SeriesRef, l labels.Labels, t int64,
 	} else {
 		metrics.GlobalRefMapping.RemoveStaleMarker(uint64(ref))
 	}
-	app.buffer[t] = append(app.buffer[t], &metrics.FlowMetric{
-		GlobalRefID: uint64(ref),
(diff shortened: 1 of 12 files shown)


component/metrics/mutate/mutate.go
```diff
@@ -101,19 +102,21 @@ func (c *Component) Update(args component.Arguments) error {
 // series. This is a blocker for releasing a production-grade  of the metrics.mutate
 // component.
 func (c *Component) Receive(ts int64, metricArr []*metrics.FlowMetric) {
-	app := c.appendable.Appender(context.Background())
+	c.mut.Lock()
+	defer c.mut.Unlock()
```
rfratto, line 106: This should probably be a RLock/RUnlock since many things might be sending metrics at the same time

component/metrics/mutate/mutate.go
```diff
@@ -101,19 +102,21 @@ func (c *Component) Update(args component.Arguments) error {
...
-		if m.Labels == nil {
+		newLabels := relabel.Process(m.LabelsCopy(), c.mrc...)
+		if newLabels == nil {
 			continue
 		}
-		_, err := app.Append(storage.SeriesRef(m.GlobalRefID), m.Labels, ts, m.Value)
-		if err != nil {
-			level.Error(c.opts.Logger).Log("msg", "failed to forward sample from metrics.mutate component", "err", err, "componentID", c.opts.ID)
-		}
+		relabelledMetrics = append(relabelledMetrics, metrics.NewFlowMetric(metrics.GlobalRefMapping.GetOrAddGlobalRefID(newLabels), newLabels, m.Value()))
```
rfratto, line 114: nit: kind of a long line, can we assign the new metric to a variable rather than constructing it in the append?

component/metrics/receiver.go
```diff
@@ -13,9 +14,35 @@ type Receiver struct {
...
 func (r Receiver) RiverCapsule() {}
 
-// FlowMetric is a wrapper around a single metric without the timestamp
+// FlowMetric is a wrapper around a single metric without the timestamp, FlowMetric is *mostly* immutable
+// the globalrefid can only be set if it is 0
 type FlowMetric struct {
-	GlobalRefID uint64
-	Labels      labels.Labels
-	Value       float64
+	globalRefID *atomic.Uint64
```
rfratto, line 20: This represents a single in-flight sample, right? I'm not sure I understand when you would change this in-flight.

component/metrics/receiver.go
```diff
@@ -13,9 +14,35 @@ type Receiver struct {
...
+		value:       value,
+	}
+}
+
+// GlobalRefID Retrieves the GlobalRefID
+func (fw *FlowMetric) GlobalRefID() uint64 { return fw.globalRefID.Load() }
+
+// SetGlobalRefID will only set the GlobalRefID if it is 0
+func (fw *FlowMetric) SetGlobalRefID(globalRefID uint64) {
+	fw.globalRefID.CAS(0, globalRefID)
```
rfratto, line 39: The CAS only works if the previous ref ID was 0, which it almost never will be.

component/metrics/remotewrite/remote_write.go
```diff
@@ -174,15 +174,15 @@ func (c *Component) Receive(ts int64, metricArr []*metrics.FlowMetric) {
...
-			globalID := metrics.GlobalRefMapping.GetOrAddGlobalRefID(m.Labels)
-			m.GlobalRefID = globalID
+		if m.GlobalRefID() == 0 {
+			globalID := metrics.GlobalRefMapping.GetOrAddGlobalRefID(m.LabelsCopy())
+			m.SetGlobalRefID(globalID)
 		}
-		localID := metrics.GlobalRefMapping.GetLocalRefID(c.opts.ID, m.GlobalRefID)
-		newLocal, err := app.Append(storage.SeriesRef(localID), m.Labels, ts, m.Value)
+		localID := metrics.GlobalRefMapping.GetLocalRefID(c.opts.ID, m.GlobalRefID())
+		newLocal, err := app.Append(storage.SeriesRef(localID), m.LabelsCopy(), ts, m.Value())
```
rfratto, line 182: This seems a bit like a code smell. If the intent is that senders should generate their own ref IDs, then we should rely on that here and never mutate the in-flight samples. 

Plus, if the metrics are in-flight and they were generated ad-hoc from the sender, it's not going to get the updated ref ID as it is. 

I think this should just be 

```go
localID := metrics.GlobalRefMapping.GetOrAddGlobalRefID(m.Labels()) 
```

component/metrics/receiver.go
```diff
@@ -13,9 +14,35 @@ type Receiver struct {
...
+}
+
+// Value returns the value
+func (fw *FlowMetric) Value() float64 { return fw.value }
+
+// LabelsCopy returns a copy of the labels structure
+func (fw *FlowMetric) LabelsCopy() labels.Labels {
+	return fw.labels.Copy()
 }
```
rfratto, line 48: There should be a read-only method for labels. Making a copy every time we want to read will be expensive.

rfratto (commented):
This is shaping up ok, but I'm a bit concerned about the changes as they are.

component/metrics/receiver.go
mattdurham: It's an escape hatch to ensure items have a global id set for the remote_write. It might not be needed, but ensures any metrics have an id assigned.

component/metrics/receiver.go
mattdurham: That is by design, if we don't want the double checking then we can make it immutable.

rfratto: Sorry, I missed the doc comment. The API here feels a little hard to follow, I think we should spend some time investing in making it easier to reason about.

component/metrics/receiver.go
mattdurham: When that use case arises, my plan was to add features on the FlowMetric itself. 
* ContainsLabel
* GetValue
etc

rfratto: If you're firm on copying the labels, can we write benchmarks for metrics.remote_write's receiver? 

We need to be cautious of changes that adds more overhead.

mattdurham: When benching `GlobalRefMapping.GetGlobalRefIDByLabels(fms[j].labels)` vs `GlobalRefMapping.GetGlobalRefIDByLabels(fms[j].LabelsCopy())` the second is roughly 57% slower.

mattdurham: Fiddling around to where the GlobalRefMapping grabs the underlying labels since it is in the same package and is safe to use in this specific instance.

mattdurham:
Simplified the flowmetric and the surface area of the GlobalRefMapping, GlobalRefMapping now uses the fact its in the same space as the FlowMetric to access the internal labels. It knows that it is safe to do this. Was able to remove several function calls or unexport them for smaller surface area. Finally made the globalrefid its own type so it wasn't easy to confuse a local id for a global.

component/metrics/remotewrite/remote_write.go
mattdurham: Simplified this code path

component/common/appendable/appendable.go
```diff
@@ -69,25 +63,25 @@ func (app *flowAppender) Append(ref storage.SeriesRef, l labels.Labels, t int64,
...
 
-func (app *flowAppender) AppendExemplar(ref storage.SeriesRef, l labels.Labels, e exemplar.Exemplar) (storage.SeriesRef, error) {
+func (app *flowAppender) AppendExemplar(
+	ref storage.SeriesRef,
+	l labels.Labels,
+	e exemplar.Exemplar,
+) (storage.SeriesRef, error) {
+
 	return 0, nil
 }
```
rfratto, line 86: Can you add a TODO here? This is something we'll want to implement before release.

component/metrics/globalrefmap.go
```diff
@@ -89,8 +89,8 @@ func (g *GlobalRefMap) GetOrAddGlobalRefID(l labels.Labels) uint64 {
 	return g.globalRefID
 }
 
-// GetGlobalRefID returns the global refid for a component local combo, or 0 if not found
-func (g *GlobalRefMap) GetGlobalRefID(componentID string, localRefID uint64) uint64 {
+// GetGlobalRefIDForComponent returns the global refid for a component local combo, or 0 if not found
+func (g *GlobalRefMap) GetGlobalRefIDForComponent(componentID string, localRefID uint64) RefID {
```
rfratto, line 93: I just noticed that `g.mappings` never gets cleaned up, even after a component goes away, so mappings will leak. Can you add a TODO somewhere to fix the leak?

component/metrics/receiver.go
```diff
@@ -13,9 +13,37 @@ type Receiver struct {
 // RiverCapsule marks receivers as a capsule.
 func (r Receiver) RiverCapsule() {}
 
-// FlowMetric is a wrapper around a single metric without the timestamp
+// FlowMetric is a wrapper around a single metric without the timestamp, FlowMetric is *mostly* immutable
+// the globalrefid can only be set if it is 0
```
rfratto, line 17: Is the "mostly" comment stale now?

component/metrics/remotewrite/remote_write.go
```diff
@@ -173,16 +173,11 @@ func (c *Component) Update(newConfig component.Arguments) error {
...
 	for _, m := range metricArr {
-		// TODO this should all be simplified into one call
-		if m.GlobalRefID == 0 {
-			globalID := metrics.GlobalRefMapping.GetOrAddGlobalRefID(m.Labels)
-			m.GlobalRefID = globalID
-		}
-		localID := metrics.GlobalRefMapping.GetLocalRefID(c.opts.ID, m.GlobalRefID)
-		newLocal, err := app.Append(storage.SeriesRef(localID), m.Labels, ts, m.Value)
+		localID := metrics.GlobalRefMapping.GetLocalRefID(c.opts.ID, m.GlobalRefID())
+		newLocal, err := app.Append(storage.SeriesRef(localID), m.LabelsCopy(), ts, m.Value())
```
rfratto, line 177: I'm still concerned about this making a copy of the labels for every single sample.

mattdurham: If we ever expose relabel_configs for remote_write then they could mutate the labels.

component/metrics/globalrefmap.go
mattdurham: Fixed the leak.

component/common/appendable/appendable.go
```diff
@@ -69,34 +63,35 @@ func (app *flowAppender) Append(ref storage.SeriesRef, l labels.Labels, t int64,
...
+	e exemplar.Exemplar,
+) (storage.SeriesRef, error) {
+	// TODO actually implement this before flow production.
 	return 0, nil
 }
 
 func (app *flowAppender) Commit() error {
 	for _, r := range app.receivers {
 		for ts, metrics := range app.buffer {
-			if r == nil || r.Receive == nil {
```
rfratto, line 97: Why are we removing this check? IIRC this was added at some point to prevent a panic.

rfratto (changes requested):
I'd prefer for us to not merge this in its current state. 

Since we've discovered how much time we're spending in the GC, I'm hard against us copying labels for every received sample, which will exacerbate the problem. We should be matching Prometheus' behavior for labels: do copy-on-write prior to when label modifications are going to be done, not copy-on-read. 

I'd suggest building a Docker image for this and running it in our dev clusters for a while to see what impact it has on CPU. If I'm wrong, I'll retract my concerns, but I'm fairly certain this will have a significant performance hit.

component/common/appendable/appendable.go
mattdurham: I was going to ensure that never happens at the receiver level but will add that back in until I make that change in a separate PR.

PR #2016, opened 2022-08-09
River: preserve pointers and references when decoding

### LINKED ISSUE #1947: River: capsules of Go structs always get copied by value
The following test will fail: 

```go
package value_test 

import (
  "github.com/grafana/agent/pkg/river/internal/value"
)

type capsule struct{} 
func (capsule) RiverCapsule() {}

func Test(t *testing.T) {
  var rawValue capsule 
  expect := &rawValue 

  var actual *capsule 
  err := value.Decode(value.Encode(expect), &actual) 
  require.NoError(t, err) 
  require.Equal(t, unsafe.Pointer(expect), unsafe.Pointer(actual)) 
}
```

Currently, decoding into a pointer _always_ allocates new memory, even if the capsule value being decoded from is also a pointer. 

This will break any code which is trying to share memory through capsules (e.g., sending a shared cache to another component). It seems like if both sides of the decode are capsules of struct pointers, the memory should be shared. If either side is not a pointer, then it is fine to do a copy-by-value. 

The temporary workaround is to use capsules backed by interface values instead; those never get deferenced and so the memory gets shared properly. Another workaround is to use a capsule struct where every field is a pointer, map, func, or interface; inner fields within a capsule get copied as-is.

rfratto:
This change modifies River so that decoded values only make a copy when the Go values do not match directly.

This means that pointer addresses, map references, and slice references are preserved.

When direct assignment is performed, River bypasses invoking the TextUnmarshaler and river.Unmarshaler interfaces.

As an exception, copies are made if the target Go type has an interface{} anywhere within its type definition, since interface{} values are expected to decode into known types (where River objects always decode into map[string]interface{}, etc).

This change make sa dramatic improvement to object decoding:

```
ObjectDecode-10    5.84µs ± 2%    0.04µs ± 1%  -99.27%  (p=0.000 n=19+18)

ObjectDecode-10    3.67kB ± 0%    0.01kB ± 0%  -99.75%  (p=0.000 n=20+20)

ObjectDecode-10       106 ± 0%         2 ± 0%  -98.11%  (p=0.000 n=20+20)
```

However, this change puts larger pressure on components to treat Arguments and Exports as immutable. Documentation has been updated to reflect this new requirement.

Closes #1947.

### DIFF
--- a/component/component.go
+++ b/component/component.go
@@ -33,6 +33,17 @@
 // Default values for Arguments may be provided by implementing
 // river.Unmarshaler.
 //
+// Arguments and Exports immutability
+//
+// Arguments passed to a component should be treated as immutable, as memory
+// can be shared between components as an optimization. Components should make
+// copies for fields they need to modify. An exception to this is for fields
+// which are expected to be mutable (e.g., interfaces which expose a
+// goroutine-safe API).
+//
+// Similarly, Exports and the fields within Exports must be considered
+// immutable after they are written for the same reason.
+//
 // Mapping River strings to custom types
 //
 // Custom encoding and decoding of fields is available by implementing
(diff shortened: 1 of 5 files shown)


mattdurham (approved):
Great work, excited to see the improvement

PR #2019, opened 2022-08-09
Flow: use common Target type amongst related components

rfratto:
To be able to take advantage of the performance improvement from #2016, the various Flow components need to use a common Target type, since values can only be directly assigned if the Go types are exactly the same.

### DIFF
--- a/component/discovery/discovery.go
+++ b/component/discovery/discovery.go
@@ -4,10 +4,13 @@ import (
 	"context"
 	"time"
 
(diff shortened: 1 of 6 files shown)


rfratto:
I think even with this change, we'll still miss cases where calling `concat` causes things to bypass the #2016 logic, so after this PR finding a way to optimize concat would be next.

marctc (approved):

PR #2021, opened 2022-08-10
Register metrics

mattdurham:
#### PR Description

Register the prometheus discovery metrics. Its not ideal since it uses the default register but during normal operation that is the register passed into the New agent. 

- [X] CHANGELOG updated


rfratto (approved):
Ah nice, I was wondering why those metrics never showed up :) Can we open an upstream PR to Prometheus to allow the registerer to be specified?

PR #2022, opened 2022-08-10
river: have `concat` bypass decoder

rfratto:
This introduces the concept of a "raw function" which can be used to implement River functions that don't decode to Go values.

Then, `concat` is changed to be implemented as a raw function. This has a dramatic impact on its performance:

```
name       old time/op    new time/op    delta
Concat-10    23.3µs ± 1%     2.3µs ± 2%  -90.27%  (p=0.000 n=19+18)

name       old alloc/op   new alloc/op   delta
Concat-10    21.3kB ± 0%     1.5kB ± 0%  -93.14%  (p=0.000 n=20+20)

name       old allocs/op  new allocs/op  delta
Concat-10       451 ± 0%        20 ± 0%  -95.57%  (p=0.000 n=20+20)
```

`concat` is in the hot path for Flow when there are many service discoveries updating at the maximum rate (once per 5 seconds).

### DIFF
--- a/pkg/river/internal/stdlib/stdlib.go
+++ b/pkg/river/internal/stdlib/stdlib.go
@@ -4,7 +4,12 @@ package stdlib
 import (
 	"encoding/json"
 	"os"
+	"reflect"
+
+	"github.com/grafana/agent/pkg/river/internal/value"
 )
+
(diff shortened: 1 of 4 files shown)


mattdurham (approved):
Woot! Ship it

PR #2025, opened 2022-08-10
simulate errors on remote write endpoint

rlankfo:
#### PR Description

This adds an option to return 500 on the fake remote write endpoint used in the smoke test 1% of the time if enabled.

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/tools/smoke/internal/smoke.go
+++ b/tools/smoke/internal/smoke.go
@@ -19,6 +19,8 @@ type Smoke struct {
 	cfg    *Config
 	logger log.Logger
 	tasks  []repeatingTask
+
+	fakeRemoteWriteHandler http.HandlerFunc
 }
 
 // Config struct to pass configuration to the Smoke constructor.
@@ -27,6 +29,7 @@ type Config struct {
 	Namespace         string
 	PodPrefix         string
 	FakeRemoteWrite   bool
+	SimulateErrors    bool
 	ChaosFrequency    time.Duration
 	MutationFrequency time.Duration
 }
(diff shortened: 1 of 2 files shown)


tools/smoke/internal/smoke.go
```diff
@@ -37,6 +40,7 @@ func (c *Config) RegisterFlags(f *flag.FlagSet) {
 	f.StringVar(&c.Kubeconfig, "kubeconfig", DefaultConfig.Kubeconfig, "absolute path to the kubeconfig file")
 	f.StringVar(&c.PodPrefix, "pod-prefix", DefaultConfig.PodPrefix, "prefix for grafana agent pods")
 	f.BoolVar(&c.FakeRemoteWrite, "fake-remote-write", DefaultConfig.FakeRemoteWrite, "remote write endpoint for series which are discarded, useful for testing and not storing metrics")
+	f.BoolVar(&c.SimulateErrors, "simulate-errors", DefaultConfig.SimulateErrors, "remote write endpoint will return a 500 error response 1% of the time.")
```
rfratto, line 43: Nice, should the percentage be configurable though? Maybe something like `--simulate-error-percentage=0.01` (for 1%, expecting a value between 0 and 1)?

rlankfo: Yeah, I could add this. The way I've implemented the 1% is basically to just return an error every 100th request. I think it might make sense to use the error percentage as a weighted probability so if it's set to 0.05, then there's a 5% chance each request will return an error. What do you think?

rfratto (approved):

PR #2029, opened 2022-08-11
metrics.remote_write: expose truncate & min/max keepalive time

rfratto:
This adds a `wal` block to metrics.remote_write which can be used to configure how the WAL gets cleaned up over time.

Three attributes are exposed within this new block:

1. `truncate_frequency`: How often the WAL gets cleaned up, with new checkpoints being added and stale series being deleted from memory.

2. `min_keepalive_time`: How old samples should be before they're considered from being deleted during the next truncate.

3. `max_keepalive_time`: How old samples can be before they're forced to be deleted during the next truncate.

The remote flush duration is left unexposed for now because there's no way to dynamically change it without shutting down the remote_write.

Closes #2023.

### DIFF
--- a/component/metrics/remotewrite/remote_write.go
+++ b/component/metrics/remotewrite/remote_write.go
@@ -24,14 +24,8 @@ import (
 
 // Options.
 //
-// TODO(rfratto): These should be flags. How do we want to handle static
-// options for components?
-var (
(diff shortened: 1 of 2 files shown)


mattdurham (approved):
Lgtm

PR #2031, opened 2022-08-11
operations/river-jsonnet: allow river.expr() to be used with std.prune()

rfratto:
`std.prune` removes all hidden fields. This commit changes River expressions to have a non-hidden field which indicates that it's an expression.

The new object is constructed with a name which is very unlikely to appear as a key in an object literal to avoid invalid manifests from being generated.


mattdurham (approved):
Ahhh jsonnet

PR #2032, opened 2022-08-11
docs/flow: Add documentation for Flow concepts

rfratto:
The "concepts" section should explain everything there is to know about Flow at a high-level: 

1. What components are 
2. What the component controller is, how components get evaluated, component health
3. What the configuration language is 

If this is missing any other high-level documentation for Flow, please let me know and I'll add it. 

I think we'll want to eventually have high-level documentation for concepts like Prometheus metrics and service discovery, but I'm leaving that out for now to just focus on Flow-specific things.

### DIFF
--- a/docs/flow/concepts/component_controller.md
+++ b/docs/flow/concepts/component_controller.md
@@ -6,3 +6,115 @@ weight: 200
 ---
 
 # Component controller
+
+The _component controller_ is the core part of Grafana Agent Flow which manages
+components at runtime.
+
+It is responsible for:
+
+* Reading and validating the config file
+* Managing the lifecycle of defined components
+* Evaluating the arguments used to configure components
+* Reporting the health of defined components
+
+## Component graph
+
+As discussed in [Components][], a relationship between components is created
+when an expression is used to set the argument of one component to an export of
+another component.
+
+The set of all components and the relationships between them define a [directed
+acyclic graph][DAG] (DAG), which informs the component controller which
+references are valid and in what order components must be evaluated.
+
+For a config file to be valid, components must not reference themselves or
+contain a cyclic reference:
+
+```river
+// INVALID: local.file.some_file may not reference itself:
+local.file "self_reference" {
+  filename = local.file.self_reference.content
+}
+```
+
+```river
+// INVALID: cyclic reference between local.file.a and local.file.b:
+local.file "a" {
+  filename = local.file.b.content
+}
+local.file "b" {
+  filename = local.file.a.content
+}
+```
+
+## Component evaluation
+
+A component is _evaluated_ when its expressions are computed into concrete
+values and used to configure the component. The component controller is
+finished loading once all components have been evaluated and are running.
+
+The component controller is guaranteed to only evaluate a given component after
+evaluating all of the components the given component references. Components
+which do not reference other components may be evaluated at any time during the
+evaluation process.
+
+## Component reevaluation
+
+As mentioned in [Components][], a comonent is dynamic: a component may update
+its exports any number of times throughout its lifetime.
+
+When a component updates its exports, a _controller reevaluation_ is triggered:
+the component controller will reevaluate any component which references the
+changed component, any components which reference those components, and so on,
+until all affected components have been reevaluated.
+
+## Component health
+
+At any given time, a component may have one of the following health states:
+
+1. Unknown: default state, the component isn't running yet
+2. Healthy: the component is working as expected
+3. Unhealthy: the component is not working as expected
+4. Exited: the component has stopped and is no longer running
+
+By default, the component controller determines the health of a component. The
+component controller will mark a compoent has healthy as long as that component
+is running and its most recent evaluation was successful.
+
+Some components may report their own component-specific health information. For
+example, the `local.file` component will report itself unhealthy if the file it
+was watching gets deleted.
+
+The overall health of a component is reported by combining the
+controller-reported health of the component with the component-specific health
+information.
+
+An individual component's health is independent from the health of any other
+components it references: a component may be marked as healthy even if it
+references the exports of an unhealthy component.
+
+## Handling evaluation failures
+
+When a component fails to evaluate, it is marked as unhealthy with the reason
+for why the evaluation failed.
+
+When an evaluation fails, the component continues operating as normal: it
+continues using its previous set of evaluated arguments, and it may continue
+exporting new values.
+
+This prevents failure propagation: if your `local.file` component which watches
+API keys suddenly stops working, other components will continue using the last
+valid API key until the component returns to a healthy state.
+
+## Updating the config file
+
+Both the `/-/reload` HTTP endpoint and the `SIGHUP` singal can be used to
+inform the component controller to reload the config file. When this happens,
+the component controller will synchronize the set of running components with
+the ones in the config file, removing components which are no longer defined in
+the config file and creating new components which were added to the config
+file. All components managed by the controller will be reevaluated after
+reloading.
+
+[Components]: {{< relref "./components.md" >})
+[DAG]: https://en.wikipedia.org/wiki/Directed_acyclic_graph
--- a/docs/flow/concepts/component_controller.md
+++ b/docs/flow/concepts/component_controller.md
@@ -60,7 +60,7 @@ evaluation process.
 
 ## Component reevaluation
 
-As mentioned in [Components][], a comonent is dynamic: a component may update
+As mentioned in [Components][], a component is dynamic: a component may update
 its exports any number of times throughout its lifetime.
 
 When a component updates its exports, a _controller reevaluation_ is triggered:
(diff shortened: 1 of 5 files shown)


rfratto:
@mattdurham @tpaschalis can both of you review this when you get a chance?

docs/flow/concepts/component_controller.md
```diff
@@ -6,3 +6,115 @@ weight: 200
...
+finished loading once all components have been evaluated and are running.
+
+The component controller is guaranteed to only evaluate a given component after
+evaluating all of the components the given component references. Components
+which do not reference other components may be evaluated at any time during the
+evaluation process.
+
+## Component reevaluation
+
+As mentioned in [Components][], a comonent is dynamic: a component may update
```
tpaschalis, line 63: s/comonent/component

docs/flow/concepts/component_controller.md
```diff
@@ -6,3 +6,115 @@ weight: 200
...
+
+At any given time, a component may have one of the following health states:
+
+1. Unknown: default state, the component isn't running yet
+2. Healthy: the component is working as expected
+3. Unhealthy: the component is not working as expected
+4. Exited: the component has stopped and is no longer running
+
+By default, the component controller determines the health of a component. The
+component controller will mark a compoent has healthy as long as that component
```
tpaschalis, line 81: s/compoent/component

docs/flow/concepts/components.md
```diff
@@ -6,3 +6,111 @@ weight: 100
...
+depends on another component's output (exports). The input of the component
+will now be re-evaluated any time the exports of the components it references
+get updated.
+
+The flow of data through the set of references between components forms a
+_pipeline_.
+
+An example pipeline may look like this:
+
+1. A `local.file` component watches a file on disk containing an API key.
```
tpaschalis, line 57: Doesn't really matter as Markdown will render the numbered list correctly, but the numbering is off.

docs/flow/concepts/components.md
```diff
@@ -6,3 +6,111 @@ weight: 100
...
+* Arguments: settings which configure a component
+* Exports: named values which a component exposes to other components
+
+Each component has a name which describes what that component is responsible
+for. For example, the `local.file` component is responsible for retrieving the
+contents of files on disk.
+
+Components are specified by users by first providing the component's name with
+a user-specified label, and then by providing arguments to configure the
+component.
```
tpaschalis, line 25: How about a sentence to explain that this is the way you can have multiple components of the same type?

docs/flow/concepts/configuration_language.md
```diff
@@ -6,3 +6,86 @@ weight: 300
...
+}
+
+metrics.remote_write "default" {
+  remote_write {
+    url = "http://localhost:9009/api/prom/push"
+  }
+}
+```
+
+River was designed with two requirements in mind:
```
tpaschalis, line 30: s/two/three

docs/flow/concepts/configuration_language.md
```diff
@@ -6,3 +6,86 @@ weight: 300
...
+* `remote_write`: An unlabeled block inside the component which configures an
+  endpoint to send metrics to. This block sets the `url` attribute to specify
+  what the endpoint is.
+
+## More information
+
+River is documented in detail in [Configuration language][config-docs] section
+of the Grafana Agent Flow docs.
+
+[config-docs]: {{< relref "../config-language/" >}}
```
tpaschalis, line 91: By this time, I think the answer to "Why not YAML?" should be more obvious. Do we want to add a section for a 2-sentence explanation?

docs/flow/concepts/components.md
```diff
@@ -6,3 +6,111 @@ weight: 100
...
+3. A `metrics.scrape` component references the exports of the previous
+   component, and sends collected metrics to the `metrics.remote_write`
+   component.
+
+A user would use this config file to represent the above pipeline:
+
+```river
+// Get our API key from disk.
+//
+// This component has an exported value called "contents", holding the contents
```
tpaschalis, line 72: We reference the export both as `content` and `contents` in different places.

tpaschalis (approved):
LGTM, small nits only. I think it's a good base for someone learning about Flow.

docs/flow/concepts/components.md
```diff
@@ -6,3 +6,111 @@ weight: 100
...
+   for authentication.
+2. A `discovery.kubernetes` component discovers and exports Kubernetes Pods
+   where metrics can be collected.
+3. A `metrics.scrape` component references the exports of the previous
+   component, and sends collected metrics to the `metrics.remote_write`
+   component.
+
+A user would use this config file to represent the above pipeline:
+
+```river
```
marctc, line 69: Is this gonna be rendered properly? Should we use `hcl` in the meantime (if it does anything, ofc)?

rfratto: It doesn't add any syntax highlighting because it doesn't know what `river` is but it shows "River" when you hover over the code block. Adding HCL might be misleading since it would show on the top-right (and might highlight improperly in some cases where the languages differ)

docs/flow/concepts/configuration_language.md
```diff
@@ -6,3 +6,86 @@ weight: 300
 ---
 
 # Configuration language
+
+The Grafana Agent Flow _configuration language_ refers to the language used in
+configuration files which define and configure components to run.
+
+The configuration language is called River, a Terraform/HCL-inspired language:
```
marctc, line 13: Link to the RFC?

docs/flow/concepts/configuration_language.md
```diff
@@ -6,3 +6,86 @@ weight: 300
...
+metrics.remote_write "default" {
+  remote_write {
+    url = "http://localhost:9009/api/prom/push"
+  }
+}
+```
+
+River was designed with two requirements in mind:
+
+* _Fast_: The configuration language must be fast so the component controller
```
marctc, line 32: This could trigger discussions if River is faster than HCL if we don't detail why is fast. Probably same for the rest of requirements.

docs/flow/concepts/component_controller.md
```diff
@@ -6,3 +6,115 @@ weight: 200
...
+continues using its previous set of evaluated arguments, and it may continue
+exporting new values.
+
+This prevents failure propagation: if your `local.file` component which watches
+API keys suddenly stops working, other components will continue using the last
+valid API key until the component returns to a healthy state.
+
+## Updating the config file
+
+Both the `/-/reload` HTTP endpoint and the `SIGHUP` singal can be used to
```
mattdurham, line 111: signal

docs/flow/concepts/components.md
```diff
@@ -6,3 +6,111 @@ weight: 100
...
+The flow of data through the set of references between components forms a
+_pipeline_.
+
+An example pipeline may look like this:
+
+1. A `local.file` component watches a file on disk containing an API key.
+1. A `metrics.remote_write` component is configured to receive metrics and
+   forward them to an external database using the API key from the `local.file`
+   for authentication.
+2. A `discovery.kubernetes` component discovers and exports Kubernetes Pods
```
mattdurham, line 61: I think this is `discovery.k8s`

docs/flow/concepts/components.md
```diff
@@ -6,3 +6,111 @@ weight: 100
...
+      password = local.file.api_key.contents
+    }
+  }
+}
+
+// Find Kubernetes pods where we can collect metrics.
+//
+// This component exports a "targets" value which contains the list of
+// discovered pods.
+discovery.kubernetes "pods" {
```
mattdurham, line 107: `discovery.k8s`

mattdurham (approved):
Minor nits but overall looks good

docs/flow/concepts/configuration_language.md
rfratto: Yeah, I want to keep this page light and just explain high-level concepts. I think the detailed config language documentation can cover more of the rationale behind the language. 

(There's also an argument to be made about whether the docs should be defensive and rationalize decisions, but I think this is something people are going to ask about so it makes sense to explain somewhere)

docs/flow/concepts/configuration_language.md
rfratto: Given that the RFC mainly just talks about justification and not about any language details, I wonder if linking to it would just cause more confusion 🤔

docs/flow/concepts/configuration_language.md
rfratto: Agreed, but I think the detailed docs for River are a better place to talk more about the speed rather than the high-level overview page this acts as.

docs/flow/concepts/configuration_language.md
rfratto: I'll leave this as-is for now, but we can consider whether the detailed section for River can link to the RFC somewhere.

PR #2033, opened 2022-08-11
simplified metrics fix

### LINKED ISSUE #2030: Flow: metrics.mutate applies the mutation on the same input flow metric
metrics.mutate applies the relabel change to the input label which violates the immutability of flowmetric. Since multiple mutates can be working on the same label set this can cause undefined behavior if they are all mutating the same labelset onto itself.

mattdurham:
#### PR Description

Fixes the issue in the mutate function where it is assigning what should be an immutable label set. This is a tighter PR with a low amount of changes to fix the mutate issue and help guard against bad uses of labels.

Closes #2030

- [X] Tests updated

### DIFF
--- a/component/common/appendable/appendable.go
+++ b/component/common/appendable/appendable.go
@@ -79,11 +79,7 @@ func (app *flowAppender) Append(ref storage.SeriesRef, l labels.Labels, t int64,
 	} else {
 		metrics.GlobalRefMapping.RemoveStaleMarker(uint64(ref))
 	}
-	app.buffer[t] = append(app.buffer[t], &metrics.FlowMetric{
-		GlobalRefID: uint64(ref),
-		Labels:      l,
-		Value:       v,
-	})
+	app.buffer[t] = append(app.buffer[t], metrics.NewFlowMetric(uint64(ref), l, v))
 	return ref, nil
 }
 
--- a/component/metrics/globalrefmap.go
+++ b/component/metrics/globalrefmap.go
@@ -44,7 +44,7 @@ func newGlobalRefMap() *GlobalRefMap {
 }
 
 // GetOrAddLink is called by a remote_write endpoint component to add mapping and get back the global id.
-func (g *GlobalRefMap) GetOrAddLink(componentID string, localRefID uint64, l labels.Labels) uint64 {
+func (g *GlobalRefMap) GetOrAddLink(componentID string, localRefID uint64, fm *FlowMetric) uint64 {
 	g.mut.Lock()
 	defer g.mut.Unlock()
 
@@ -59,7 +59,7 @@ func (g *GlobalRefMap) GetOrAddLink(componentID string, localRefID uint64, l lab
 		g.mappings[componentID] = m
 	}
 
-	labelHash := l.Hash()
+	labelHash := fm.labels.Hash()
 	globalID, found := g.labelsHashToGlobal[labelHash]
 	if found {
 		m.localToGlobal[localRefID] = globalID
--- a/component/metrics/mutate/mutate.go
+++ b/component/metrics/mutate/mutate.go
@@ -2,15 +2,13 @@ package mutate
 
 import (
 	"context"
+	"sync"
 
-	"github.com/go-kit/log/level"
 	"github.com/grafana/agent/component"
-	fa "github.com/grafana/agent/component/common/appendable"
 	flow_relabel "github.com/grafana/agent/component/common/relabel"
 	"github.com/grafana/agent/component/metrics"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/prometheus/model/relabel"
-	"github.com/prometheus/prometheus/storage"
 )
 
 func init() {
@@ -41,10 +39,10 @@ type Exports struct {
 
 // Component implements the metrics.mutate component.
 type Component struct {
-	opts component.Options
-	mrc  []*relabel.Config
-
-	appendable       *fa.FlowAppendable
+	mut              sync.RWMutex
+	opts             component.Options
+	mrc              []*relabel.Config
+	forwardto        []*metrics.Receiver
 	receiver         *metrics.Receiver
 	metricsProcessed prometheus.Counter
 }
@@ -56,7 +54,6 @@ var (
 // New creates a new metrics.mutate component.
 func New(o component.Options, args Arguments) (*Component, error) {
 	c := &Component{opts: o}
-	c.appendable = fa.NewFlowAppendable(args.ForwardTo...)
 	c.receiver = &metrics.Receiver{Receive: c.Receive}
 	c.metricsProcessed = prometheus.NewCounter(prometheus.CounterOpts{
 		Name: "agent_metrics_mutate_metrics_processed",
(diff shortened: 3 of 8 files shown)


component/metrics/receiver.go
```diff
@@ -13,9 +14,51 @@ type Receiver struct {
...
+
+// Relabel applies normal prometheus relabel rules and returns a flow metric. NOTE this may return itself.
+func (fw *FlowMetric) Relabel(cfgs ...*promrelabel.Config) *FlowMetric {
+	retLbls := promrelabel.Process(fw.labels)
+	if retLbls == nil {
+		return nil
+	}
+	if retLbls.Hash() == fw.labels.Hash() {
+		return fw
+	}
```
rfratto, line 62: There's a really minuscule chance of a hash collision here, I'd also make sure that if the hashes match the labels are also identical.

rfratto (commented):
I'll give this a closer look tomorrow, but it seems solid

component/metrics/receiver.go
mattdurham: Also entirely forgot to use cfgs...

mattdurham: Fixed and added tests.

component/common/appendable/appendable.go
```diff
@@ -79,11 +79,7 @@ func (app *flowAppender) Append(ref storage.SeriesRef, l labels.Labels, t int64,
 	} else {
 		metrics.GlobalRefMapping.RemoveStaleMarker(uint64(ref))
 	}
-	app.buffer[t] = append(app.buffer[t], &metrics.FlowMetric{
-		GlobalRefID: uint64(ref),
-		Labels:      l,
-		Value:       v,
-	})
+	app.buffer[t] = append(app.buffer[t], metrics.NewFlowMetric(uint64(ref), l, v))
```
rfratto, line 82: Thinking aloud: 

One of the nice things about the storage.Appendable API is that it runs on the stack: you don't have to allocate a new pointer for each sample when you send it to the WAL. 

Since the buffer is a slice and will exist in the heap anyway, there might be some nice improvements in allocations by making it store a non-pointer value so there's only one big thing to clean up instead of many small pointers.

Our profiles aren't showing that this is a problem right now, so it's not something I'd spend time on immediately, and it's also probably not worth working on while the API is still being shaped out.

component/metrics/globalrefmap.go
```diff
@@ -44,7 +44,7 @@ func newGlobalRefMap() *GlobalRefMap {
 }
 
 // GetOrAddLink is called by a remote_write endpoint component to add mapping and get back the global id.
-func (g *GlobalRefMap) GetOrAddLink(componentID string, localRefID uint64, l labels.Labels) uint64 {
```
rfratto, line 47: Note for later: 

GlobalRefMap is going to be the cause of a lot of bottlenecking because of how much locking it does. The biggest problem is how this is using a Lock and not a RWLock, but even if that weren't true, there's still many concurrent write operations that will be bottlenecked by the other one. 

I would suggest to start thinking about how to make the cache and global IDs as lock free as possible.

component/metrics/mutate/mutate.go
```diff
@@ -101,19 +101,22 @@ func (c *Component) Update(args component.Arguments) error {
 // series. This is a blocker for releasing a production-grade  of the metrics.mutate
 // component.
 func (c *Component) Receive(ts int64, metricArr []*metrics.FlowMetric) {
-	app := c.appendable.Appender(context.Background())
+	c.mut.RLock()
+	defer c.mut.RUnlock()
+
+	relabelledMetrics := make([]*metrics.FlowMetric, 0)
```
rfratto, line 107: Performance nit: this should have an initial capacity of `len(metricArr)` so that Go doesn't overallocate when appending metrics for small scrape jobs.

component/metrics/mutate/mutate.go
```diff
@@ -101,19 +101,22 @@ func (c *Component) Update(args component.Arguments) error {
 // series. This is a blocker for releasing a production-grade  of the metrics.mutate
 // component.
 func (c *Component) Receive(ts int64, metricArr []*metrics.FlowMetric) {
```
rfratto, line 103: Note for later:

It feels suspicious that this is an operation that never fails.

component/metrics/receiver.go
```diff
@@ -13,9 +14,51 @@ type Receiver struct {
...
+// GlobalRefID Retrieves the GlobalRefID
+func (fw *FlowMetric) GlobalRefID() uint64 { return fw.globalRefID }
+
+// Value returns the value
+func (fw *FlowMetric) Value() float64 { return fw.value }
+
+// LabelsCopy returns a copy of the labels structure
+func (fw *FlowMetric) LabelsCopy() labels.Labels {
+	return fw.labels.Copy()
+}
```
rfratto, line 46: API nit: There's no reason to have a method for this, the caller can call `fw.RawLabels().Copy()` if they want to make a copy.

component/metrics/remotewrite/remote_write.go
```diff
@@ -173,16 +173,13 @@ func (c *Component) Update(newConfig component.Arguments) error {
...
+		// Currently it doesn't look like the storage interfaces mutate the labels, but thats not a strong
+		// promise. So this should be treated with care.
+		newLocal, err := app.Append(storage.SeriesRef(localID), m.RawLabels(), ts, m.Value())
 		// Add link if there wasn't one before, and we received a valid local id
 		if localID == 0 && newLocal != 0 {
-			metrics.GlobalRefMapping.GetOrAddLink(c.opts.ID, uint64(newLocal), m.Labels)
+			metrics.GlobalRefMapping.GetOrAddLink(c.opts.ID, uint64(newLocal), m)
 		}
 		if err != nil {
 			_ = app.Rollback()
```
rfratto, line 185: Nit on the log line below (which GitHub won't let me comment on):

Receiving metrics didn't fail, but appending metrics to the WAL did.

rfratto (approved):

component/metrics/globalrefmap.go
mattdurham: I was going to hold off on making significant changes to the global ref mapping until it either a. becomes a problem or b. we do submodules. Whatever we decide to do with submodules the global ref mapping will probably have to change a fair amount (not to mention other code).

rfratto: It doesn't have to be changed right now, but I just wanted to point it out as a concern because it introduces much more lock contention across multiple scrape jobs than Prometheus Agent or the static Grafana Agent does.

component/metrics/mutate/mutate.go
mattdurham: It was hard to handle the failure, ie if it branches out N times and X succeed and Y fail what do you do? The decision was built around making it fire and forget, then have the individual components handle their own failures and not push those failures up the tree.

rfratto: Prometheus scrapers have logic that mark the scrape as failed if it can't write to the WAL. We're losing that capability here.

component/metrics/receiver.go
mattdurham: I really wanted to bury RawLabels usage as much as possible since it's very easy to forget to add the copy.

component/metrics/mutate/mutate.go
mattdurham: Created issue #2035 for further discussion.

PR #2041, opened 2022-08-16
Append APPDATA to bookmark files in agent config for Windows event logs

mshahzeb:
For the windows integration - the default windows event log configuration contains relative bookmark file paths.

```
job_name: integrations/windows-exporter-application
        windows_events:
          use_incoming_timestamp: true
          bookmark_path: "./bookmark-application.xml"
          eventlog_name: "Application"
          xpath_query: '*'
          labels:
            job: integrations/windows_exporter
        relabel_configs:
          - source_labels: ['computer']
            target_label: 'agent_hostname'
        pipeline_stages:
          - json:
              expressions:
                source: source
          - labels:
              source:
```

When running the agent with relative paths in bookmark_path - the agent service fails to start as it is not able to create these files in the relative paths. This PR appends %APPDATA% into all bookmark_path (s) present in the agent config when installing the agent for windows though the powershell script.

The agent itself does not support expanding the %APPDATA% variable in Windows. Therefore the powershell script is updated to do the replacement when installing the agent for windows through the script.

The resulting bookmark path becomes:
```
bookmark_path: C:\Users\shb\AppData\Roaming\bookmark-application.xml
```

### DIFF
--- a/production/grafanacloud-install.ps1
+++ b/production/grafanacloud-install.ps1
@@ -81,7 +81,7 @@ if ($jsonObj.status -eq "success") {
 	Start-Service "Grafana Agent"
 
 	# Wait for service to startup after restart
-	Write-Host "Wait for Grafana service to initialize"
+	Write-Host "Wait for Grafana service to initialize after restart"
 	Start-Sleep -s 10
 
 	# Show Grafana agent service status
--- a/production/grafanacloud-install.ps1
+++ b/production/grafanacloud-install.ps1
@@ -53,13 +53,26 @@ $response = Invoke-WebRequest $CONFIG_URI -Method 'GET' -Headers $headers
 
 $jsonObj = $response | ConvertFrom-Json
 if ($jsonObj.status -eq "success") {
+	$config_file = ".\agent-config.yaml"
 	Write-Host "Saving and updating agent configuration file"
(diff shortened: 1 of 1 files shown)


production/grafanacloud-install.ps1
```diff
@@ -53,13 +53,26 @@ $response = Invoke-WebRequest $CONFIG_URI -Method 'GET' -Headers $headers
...
+            $bookmark_filename = $split_line[1] -replace "./",""
+            $content = $content -replace $_,"$($prefix): $($env:APPDATA)\$($bookmark_filename)"
+        }
+        $content | Set-Content $config_file
+    }
+    Move-Item $config_file "C:\Program Files\Grafana Agent\agent-config.yaml" -force
 
 	# Wait for service to initialize after first install
+	Write-Host "Wait for Grafana service to initialize"
 	Start-Sleep -s 5
```
mattdurham, line 76: Unrelated, but can we bump this to 10s, I have seen it take a bit longer on enterprise machines.

mshahzeb: Actually this is the wait step before attempting to start/stop the service after copying the config file.
If we dont wait at this step the Stop service command tends to fail immediately after start.

However I have added a 10 second wait after the restart step as well before displaying the final status of the service.

mattdurham (approved):
Lgtm!

PR #2044, opened 2022-08-17
agent-operator: base image off of distroless base image.

captncraig:
This should lower the security footprint of the operator significantly, since it relies on many fewer things than the agent itself.

[distroless](https://github.com/GoogleContainerTools/distroless) is a much smaller, debian based image. It still has a few CVE hits from libc and openssl, but fewer than the full debian image. Tested working in my local cluster.


marctc (approved):
Nice one!

PR #2048, opened 2022-08-18
Flow UI api

mattdurham:
Main issue is moving some river constructs out of internal.


pkg/flow/json_walker.go
```diff
@@ -0,0 +1,238 @@
...
+	Key          string      `json:"key,omitempty"`
+	Name         string      `json:"name,omitempty"`
+	Label        string      `json:"label,omitempty"`
+	Type         string      `json:"type,omitempty"`
+	References   []string    `json:"references,omitempty"`
+	ReferencedBy []string    `json:"reference_by,omitempty"`
+	Health       *Health     `json:"health,omitempty"`
+	Original     string      `json:"original,omitempty"`
+	Value        interface{} `json:"value,omitempty"`
+}
```
rfratto, line 27: This is definitely doing way too much and should be split into multiple types

pkg/flow/json_walker.go
```diff
@@ -0,0 +1,238 @@
...
+	})
+}
+
+func convertExports(exports component.Exports) *Field {
+	return convertField(exports, &rivertags.Field{
+		Name:  []string{"exports"},
+		Index: nil,
+		Flags: 0,
+	})
+}
```
rfratto, line 75: For context, these should be returning River JSON bodies

pkg/flow/json_walker.go
```diff
@@ -0,0 +1,238 @@
...
+		nf.Value = &Field{
+			Type:  "string",
+			Value: "(secret)",
+		}
+		maybeSecret := in.(rivertypes.OptionalSecret)
+		if !maybeSecret.IsSecret {
+			nf.Value.(*Field).Value = maybeSecret.Value
+		}
+		return nf
+	}
```
rfratto, line 132: This should be more generic; check to see if it calls RiverTokenize instead and tokenize it into a literal string to report as a capsule.

pkg/flow/json_walker.go
```diff
@@ -0,0 +1,238 @@
+package flow
```
rfratto, line 1: The bulk of this logic is River-specific and can be moved to pkg/river so the other packages can remain internal.

pkg/flow/json_walker.go
```diff
@@ -0,0 +1,238 @@
...
+// convertToField converts to a generic field for json
+func convertToField(in interface{}, name string) *Field {
+	return convertField(in, &rivertags.Field{
+		Name:  []string{name},
+		Index: nil,
+		Flags: 0,
+	})
+}
+
+func convertField(in interface{}, f *rivertags.Field) *Field {
```
rfratto, line 86: There's two types of things to convert:

1. River bodies (slices of River statements; either an attribute or a body)
2. River values 

Separating out the logic for conversion of those two things will help simplify the checks here (like checking to see if something should be a block or an object literal)

Issue #2052, opened 2022-08-19
Proposal: clean up Makefile & CI


rfratto:
## Background

Originally, the Makefile was intended to be the entrypoint for everything, where the Makefile had the same set of outputs that any of our CI jobs would.

This hasn't been true for a while; Drone configs and GitHub Actions no longer act as wrapps around the Makefile. Despite this, the Makefile has a lot of historic complexity from when it did try to be the sole entrypoint.

While the original goal for the Makefile was nice, I don't think it's tenable anymore: we are slowly moving to a world where not all parts of the Agent project can be done by a single machine, and the Makefile should be as simple as possible so developers can understand it.

## Proposal

I propose a set of changes to how we build things to introduce consistency and generally clean things up.

At a high level:

* The Makefile should only have targets useful to both local development and CI
* We should consistently use the same build image for CI 
* Dockerfiles should no longer build the artifacts themselves 
* `//go:generate` should be used for asset generation.
* Persistent build caches should be used for improving the local development
  feedback loop.

### Changes to the Makefile

The Makefile should be focused on build primitives which are useful for both local development and CI. If a target would only be used in CI, it should not go in the Makefile.

The Makefile would include code generation (`make generate`), testing (`make test`, `make lint`), and build artifacts (binaries and Docker containers).

### Changes to CI

All Drone jobs should use grafana/agent-build-image for their work. This includes both Linux and Windows runners, as grafana/agent-build-image has a Linux and Windows container variants.

The build image should be based on an existing image which exposes cross compilers so we can build Cgo code (`rfratto/seego` or similar images).

GitHub Actions may be used if Drone is not sufficient.

### Changes to Dockerfiles

There should only be one Dockerfile per produced image instead of two.

Dockerfiles should copy built assets from the host machine rather than building it directly. This helps improve iteration time of testing changes in containers.

### Changes to code generation

`//go:generate` directives should be used as the primary interface for generating assets in the repository. Makefile targets may only be used for generating assets if the generated assets aren't consumed by Go code.  

### Changes to local builds

For local development, the Makefile should support proxying builds to the build image as it does today. 

The Makefile should be modified to support a persistent build container or persistent cache volume so that caches can be shared between builds to dramatically speed up build and iteration time.

The build container path is useful for local development when building Linux images on a non-Linux host, but is optional: developers may continue to use local tooling if they have it.

PR #2054, opened 2022-08-22
Introduce new consolidated build-image

rfratto:
This will introduce 0.15.0 of the build-image, which contains all the tooling needed for building the agent locally and through CI.

The base image is rfratto/viceroy, which is a rewrite of rfratto/seego which is a little easier to work with. Note that while rfratto/viceroy has an arm64 image, grafana/agent-build-image doesn't (yet) and only amd64 images will be produced for the 0.15.0 release. 

If there are any missing tools here needed for any Github Action or Drone job, please let me know: this should be the only image we need and have all of the dependencies already installed. 

Related to #2052.

### DIFF
--- a/build-image/Dockerfile
+++ b/build-image/Dockerfile
@@ -13,8 +13,8 @@ FROM golang:1.17 as ghr
 RUN GO111MODULE=on go get github.com/tcnksm/ghr
 
 # Dependency: docker (for building images)
-FROM alpine:edge as docker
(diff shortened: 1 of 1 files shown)


rfratto:
It looks like the docker CLI doesn't work, I'm working on fixing that (and making sure buildx is installed properly)

mattdurham (approved):
LGTM

tpaschalis (approved):

PR #2056, opened 2022-08-23
Introduce settings to disable Keep-Alive and setting IdleConnectionTimeouts for scraping connections

### LINKED ISSUE #1953: Allow settings for keep alives and idle connection timeouts for scrape configs
Allow the usage of custom keep alive duration for scrape configs (both normal and scraping service). This hopefully can be done via the custom DialerFunc

tpaschalis:

#### PR Description
This PR introduces new configuration options for allowing users to disable Keep-Alives in outgoing connections or introducing a timeout for Idle Keep-Alive connections when scraping.

This PR fixes #1953

#### Notes to the Reviewer
There are a couple of different ways to achieve this, so I guess the main thing of this PR is choosing one of them. One thing that need is to avoid breaking compatibility with configuration blocks that we share with Prometheus; wherever we say that we can copy/paste blocks between the Agent and Prometheus today, we should be able to do so in the future.

1) Propose and introduce such settings in Prometheus' configuration. While we can try kickstarting this conversation, this would take a little bit longer.

2) Add those options on the metrics instance level (as in the first commit). This is the most straightforward way of wiring this new functionality in and provides some flexibility in to which metrics instances should perform their scrapes that way.

3) Add those options under the top-level metrics key. This requires some more thought on wiring things around and removes some of the flexibility, but would be easier to enforce for more complicated configurations.

~Initially, I've went with Option 2, but let me know if you have a different opinion.~

Went with Option 3 and it seems to work; checking that there's no other path the configuration can take where the setting will not be honored.

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [X] Tests updated

### DIFF
--- a/docs/sources/configuration/metrics-config.md
+++ b/docs/sources/configuration/metrics-config.md
@@ -45,6 +45,7 @@ define one instance.
 
 # Allows to configure the maximum amount of time an idle Keep-Alive connection
 # can remain idle before closing itself. Zero means no limit.
+# The setting is ignored when `http_disable_keepalives` is enabled.
 [http_idle_conn_timeout: <duration> | default = "5m"]
 
 # The list of Prometheus instances to launch with the agent.
--- a/pkg/metrics/agent.go
+++ b/pkg/metrics/agent.go
@@ -81,10 +81,6 @@ func (c *Config) ApplyDefaults() error {
 		return errors.New("cannot use configs when scraping_service mode is enabled")
 	}
 
-	if c.DisableKeepAlives && c.IdleConnTimeout > 0 {
-		return errors.New("cannot disable keep-alives and set the idle connection timeout at the same time")
-	}
-
 	c.Global.DisableKeepAlives = c.DisableKeepAlives
 	c.Global.IdleConnTimeout = c.IdleConnTimeout
 	usedNames := map[string]struct{}{}


pkg/metrics/instance/instance_test.go
```diff
@@ -245,6 +247,82 @@ func TestInstance(t *testing.T) {
...
+
+	// Wait until mockWalStorage has had a series added to it.
+	test.Poll(t, 30*time.Second, true, func() interface{} {
+		mockStorage.mut.Lock()
+		defer mockStorage.mut.Unlock()
+		return len(mockStorage.series) > 0
+	})
+}
+
+func TestIdleConnectionTimeouts(t *testing.T) {
```
rfratto, line 282: IMO we don't need to test Prometheus code and we don't really need tests for this since we can tell at a glance the options get passed properly.

pkg/metrics/agent.go
```diff
@@ -47,6 +47,8 @@ type Config struct {
 	Configs                []instance.Config     `yaml:"configs,omitempty,omitempty"`
 	InstanceRestartBackoff time.Duration         `yaml:"instance_restart_backoff,omitempty"`
 	InstanceMode           instance.Mode         `yaml:"instance_mode,omitempty"`
+	DisableKeepAlives      bool                  `yaml:"http_disable_keepalives,omitempty"`
+	IdleConnTimeout        time.Duration         `yaml:"http_idle_conn_timeout,omitempty"`
```
rfratto, line 51: IMO these settings make more sense to be exposed in the `metrics.global` block, WDYT?

pkg/metrics/instance/instance.go
```diff
@@ -427,7 +428,15 @@ func (i *Instance) initialize(ctx context.Context, reg prometheus.Registerer, cf
 	i.storage = storage.NewFanout(i.logger, i.wal, i.remoteStore)
 
 	opts := &scrape.Options{
-		ExtraMetrics: cfg.global.ExtraMetrics,
+		ExtraMetrics:      cfg.global.ExtraMetrics,
+		HTTPClientOptions: []config_util.HTTPClientOption{},
```
rfratto, line 432: tip: You don't have to set this to an empty slice since it's valid to append to a nil slice, though doing this makes it more clear that it should be empty to start with.

rfratto (approved):

pkg/metrics/agent.go
tpaschalis: I thought about it, but currently our metrics.global block is something that can be copy/pasted between Prometheus and the Agent and I didn't want to break this symmetry. Do you feel more comfortable putting it there?

rfratto: I'm personally fine with breaking the symmetry, but it's your call

tpaschalis: My call would be to leave them as-is, except if @mattdurham also has a strong opinion.

Finally, these two options are non-compatible meaning the IdleConnTimeout will only matter for non-disabled Keep-Alive connections. Are we okay with putting a check on the `ApplyDefaults` method?

pkg/metrics/agent.go
```diff
@@ -81,6 +81,10 @@ func (c *Config) ApplyDefaults() error {
 		return errors.New("cannot use configs when scraping_service mode is enabled")
 	}
 
+	if c.DisableKeepAlives && c.IdleConnTimeout > 0 {
+		return errors.New("cannot disable keep-alives and set the idle connection timeout at the same time")
+	}
```
rfratto, line 86: I think this might cause more confusion than is necessary since if you only set `http_disable_keepalives: true` it'll fail because of the default idle conn timeout. 

We can just document that `http_idle_conn_timeout` is ignored when `http_disable_keepalives` is enabled IMO.

docs/sources/configuration/metrics-config.md
```diff
@@ -39,6 +39,15 @@ define one instance.
 # A value of 0 disables periodic cleanup of abandoned WALs
 [wal_cleanup_period: <duration> | default = "30m"]
 
+# Allows to disable HTTP Keep-Alives when scraping; the Agent will only use
+# outgoing each connection for a single request.
```
mattdurham, line 43: Something about this verbiage seems off. should `outgoing` and `each` be switched?

pkg/metrics/agent.go
mattdurham: Global means use as default. These settings don't have specific instance settings to apply so I think NOT being in global is reasonable.

docs/sources/configuration/metrics-config.md
tpaschalis: Yeah sorry, fixed!

mattdurham (approved):
LGTM

PR #2066, opened 2022-08-24
docs/flow: remove reference/flags.md

rfratto:
There is no general set of command-line flags that should be documented; instead, each cli command will document its command line flags. Command line flags for `run` are already documented in reference/cli/run.md.


mattdurham (approved):
-8 lines of code nothing added...sweet

PR #2067, opened 2022-08-24
Update version numbers for 0.27.0-rc.0

captncraig:
Update docs and deployments for 0.27.0-rc.0

### DIFF
--- a/production/kubernetes/install-bare.sh
+++ b/production/kubernetes/install-bare.sh
@@ -25,7 +25,7 @@ check_installed() {
 check_installed curl
 check_installed envsubst
 
-MANIFEST_BRANCH=v0.26.1
+MANIFEST_BRANCH=v0.27.0
 MANIFEST_URL=${MANIFEST_URL:-https://raw.githubusercontent.com/grafana/agent/${MANIFEST_BRANCH}/production/kubernetes/agent-bare.yaml}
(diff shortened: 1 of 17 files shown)


production/kubernetes/install-bare.sh
```diff
@@ -25,7 +25,7 @@ check_installed() {
 check_installed curl
 check_installed envsubst
 
-MANIFEST_BRANCH=v0.26.1
+MANIFEST_BRANCH=v0.27.0
```
rfratto, line 28: I wonder if this should actually be release-0.27 given our new branch convention?

rfratto (commented):
production/grafanacloud-install.sh also needs to be updated

production/kubernetes/install-bare.sh
captncraig: I interpreted it that way. It looks like it will pull from the release branch, not the rc tag.

rfratto: This will resolve the URL below to https://raw.githubusercontent.com/grafana/agent/v0.27.0/production/kubernetes/agent-bare.yaml 

I think `MANIFEST_BRANCH` should be changed to either be the literal branch name `release-v0.27` or the tag `v0.27.0-rc.0`

rfratto (approved):
LGTM after fixing MANIFEST_BRANCH

PR #2071, opened 2022-08-24
SNMP Exporter Scrape Configs

rgeyer:
#### PR Description
The SNMP config was not correctly setting additional scrape params (`module`, and `walk_params` specifically) when they defined in the configuration file. This would result in incomplete service discovery, and scraping a target without those values.

Partially #2017

#### Notes to the Reviewer
I am also experiencing a yaml unmarshal error which I'm trying to track down when using the agent operator and the SNMP integration. Time will tell if I can sort it out.

#### PR Checklist

- [x] CHANGELOG updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,10 @@ Main (unreleased)
 
 - Fix grafanacloud-install.ps1 web request internal server error when fetching config. (@rlankfo)
 
+- Fix snmp integration not passing module or walk_params parameters when scraping. (@rgeyer)
+
+- Fix unmarshal errors (key "<walk_param name>" already set in map) for snmp integration config when walk_params is defined, and the config is reloaded. (@rgeyer)
+
 ### Other changes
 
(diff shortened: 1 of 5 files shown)


marctc (approved):
LGTM. Could you update changelog as well?

rgeyer:
> LGTM. Could you update changelog as well?

Done. Also, I found my unmarshalling error. Because the snmp config defines `walk_params` as a map with a string key, when the [reload endpoint](https://github.com/grafana/agent/blob/main/cmd/agent/entrypoint.go#L233-L243) is called, the config struct for the snmp integration already contains values in `walk_params` and a subsequent unmarshal results in an error that the key(s) are already set in the map.

I added a commit which (heavy handedly?) nukes the `walk_params`, so that when it is unmarshaled again, there are no existing values to conflict.

Would you mind taking another quick peek?

Thanks!

gloriasc (approved):
Good catch! LGTM

marctc (approved):

PR #2072, opened 2022-08-25
Use go install to install binaries in drone 

captncraig:
go get for installing binaries is deprecated in go 1.18.

Trying to see if this fixes the release.

https://go.dev/doc/go-get-install-deprecation


mattdurham (approved):
Lgtm

PR #2078, opened 2022-08-26
Use ubuntu base image

captncraig:
`ubuntu:jammy` is slightly smaller than `debian:bullseye-slim`, and has 0 high or critical vulnerabilities in trivy.

If it works, this could solve a lot of problems.

It runs for me locally, but still needs verification of the cgo bits:

- [x] test ebpf (@tpaschalis plz). I'm not even sure if that works in the current main image though.
- [x] test systemd log scraping (@captncraig )

### DIFF
--- a/cmd/agent-operator/Dockerfile
+++ b/cmd/agent-operator/Dockerfile
@@ -6,7 +6,7 @@ ARG IMAGE_TAG
 
 RUN make clean && make IMAGE_TAG=${IMAGE_TAG} RELEASE_BUILD=${RELEASE_BUILD} BUILD_IN_CONTAINER=false agent-operator
 
-FROM debian:bullseye-slim
+FROM ubuntu:focal
 
 RUN apt-get update && \
   apt-get install -qy tzdata ca-certificates && \
--- a/cmd/agent-operator/Dockerfile
+++ b/cmd/agent-operator/Dockerfile
@@ -6,7 +6,7 @@ ARG IMAGE_TAG
 
 RUN make clean && make IMAGE_TAG=${IMAGE_TAG} RELEASE_BUILD=${RELEASE_BUILD} BUILD_IN_CONTAINER=false agent-operator
 
-FROM ubuntu:focal
+FROM ubuntu:jammy
 
 RUN apt-get update && \
   apt-get install -qy tzdata ca-certificates && \
--- a/cmd/agent-operator/Dockerfile.buildx
+++ b/cmd/agent-operator/Dockerfile.buildx
@@ -19,7 +19,7 @@ ARG IMAGE_TAG
 
 RUN make clean && make IMAGE_TAG=${IMAGE_TAG} RELEASE_BUILD=${RELEASE_BUILD} BUILD_IN_CONTAINER=false DRONE=true agent-operator
 
-FROM debian:bullseye-slim
+FROM ubuntu:jammy
 
 RUN apt-get update && \
   apt-get install -qy tzdata ca-certificates && \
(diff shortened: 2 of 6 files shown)


tpaschalis:
I just created a DigitalOcean VPS with
```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04 LTS
Release:	22.04
Codename:	jammy
```

I installed Docker, Go, `build-essential` and `libsystemd-dev` and ran `make agent-image` on this branch. When running the image, i get some warnings but no errors.
```
$ docker image ls
REPOSITORY      TAG                      IMAGE ID       CREATED         SIZE
grafana/agent   cmp_ubuntubase-2b7d8fc   b4942d7a7318   5 minutes ago   395MB

$ docker run --privileged -v /tmp/agent:/etc/agent/data -v /root/agent-config.yaml:/etc/agent/agent.yaml -v /lib/modules:/lib/modules -v /sys:/sys -v /usr/src:/usr/src grafana/agent:cmp_ubuntubase-2b7d8fc -config.file /etc/agent/agent.yaml
...
...
include/linux/compiler-clang.h:43:9: warning: '__HAVE_BUILTIN_BSWAP16__' macro redefined [-Wmacro-redefined]
#define __HAVE_BUILTIN_BSWAP16__
        ^
<command line>:3:9: note: previous definition is here
#define __HAVE_BUILTIN_BSWAP16__ 1
        ^
3 warnings generated.
```

If I exec into the pod, I can see the eBPF integration running successfully, listed as a target and reporting metrics, which are correctly forwarded to Grafana Cloud.
```
$ docker ps
CONTAINER ID   IMAGE                                  COMMAND                  CREATED              STATUS              PORTS     NAMES
d34c50f0b7c8   grafana/agent:cmp_ubuntubase-2b7d8fc   "/bin/agent -config.…"   About a minute ago   Up About a minute             strange_greider

$ docker exec -it d34c50f0b7c8 bash

$ curl localhost:12345/integrations/ebpf/metrics
# HELP ebpf_build_info A metric with a constant '1' value labeled by version, revision, branch, and goversion from which ebpf was built.
# TYPE ebpf_build_info gauge
ebpf_build_info{branch="cmp_ubuntubase",goversion="go1.18",revision="2b7d8fc9",version="cmp_ubuntubase-2b7d8fc"} 1
...

# curl localhost:12345/agent/api/v1/metrics/targets  | jq
{
  "status": "success",
  "data": [
    {
      "instance": "2ba88ba02a3b5fc6a1e8efb4565826b0",
      "target_group": "integrations/ebpf",
      "endpoint": "http://127.0.0.1:12345/integrations/ebpf/metrics",
      "state": "up",
      "labels": {
        "agent_hostname": "d34c50f0b7c8",
        "instance": "ebpf",
        "job": "integrations/ebpf"
      },
...
```

So all in all, it worked without any hassle.

One hunch is that the issue that we saw together might have to do with building and running the image on different distro versions; in that case it might be indeed a case with having to install the necessary linux headers for jammy.

marctc:
> `ubuntu:jammy` is slightly smaller than `debian:bullseye-slim`

How come `slim` version of Debian is bigger than Ubuntu? :open_mouth: Are those ubuntu images prepared to be slim?

tpaschalis:
@captncraig I tried running the same sequence as above on a new Focal VPS to check out if I'll get the error when building and running the image from a non-jammy machine.
```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.4 LTS
Release:	20.04
Codename:	focal
```

and everything worked out normally again (eBPF integration runs even without warnings, data goes through), so I couldn't reproduce the failure we saw.

rfratto (approved):

PR #2083, opened 2022-08-29
prepare for v0.27.0-rc.1

captncraig:
Since we've already pushed rc.0, it makes sense to roll the new commits into an rc.1. Then we can bake it in dev and push 0.27 later this week.

### DIFF
--- a/cmd/agent/Dockerfile
+++ b/cmd/agent/Dockerfile
@@ -13,9 +13,13 @@ RUN apt-get update && apt-get install -qy libbpfcc-dev
 
 RUN make clean && make IMAGE_TAG=${IMAGE_TAG} RELEASE_BUILD=${RELEASE_BUILD} BUILD_IN_CONTAINER=false agent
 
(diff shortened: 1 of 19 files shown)


cmd/agent/Dockerfile
```diff
@@ -13,13 +13,9 @@ RUN apt-get update && apt-get install -qy libbpfcc-dev
 
 RUN make clean && make IMAGE_TAG=${IMAGE_TAG} RELEASE_BUILD=${RELEASE_BUILD} BUILD_IN_CONTAINER=false agent
 
-FROM debian:bullseye-slim
+FROM debian:bookworm-slim
```
rfratto, line 16: Is this a safe change?

rfratto (approved):

cmd/agent/Dockerfile
captncraig: backed out, accidentally included a stale experiment.

PR #2084, opened 2022-08-29
Introduce Makefile-v2.mk

rfratto:
This commit temporarily introduces a new Makefile, Makefile-v2.mk, introduced initially as a separate file for easier comparison to the existing Makefile.

Makefile-v2.mk is intended to eventually replace the original Makefile and follows the designed outlined in grafana/agent#2052.

It includes targets only needed for local development and testing. Targets only intended for CI are purposefully omitted to keep the Makefile as simple as possible.

It is controlled primarily by environment variables, which are documented at the comment at the top of the Makefile. That comment may be printed to stdout by calling `make -f Makefile-v2.mk help`.

The new build image, grafana/agent-build-image:v0.15.0, is used for proxying Makefile commands when the `USE_CONTAINER=1` environment variable is set.

A Docker volume cache can be initialized to speed up builds in the build container. Run `make -f Makefile-v2.mk build-container-cache` to create the cache and `make -f Makefile-v2.mk clean` to remove it. The cache stores the Go module cache and Go build cache. The Makefile will detect the cache and automatically use it when proxying commands to the build container. When the cache is present, subsequent builds are sped up dramatically, from multiple minutes to 32 seconds. Credit to @captncraig for the idea of using Docker volumes for caching builds.

This first commit is opened early for feedback on the approach. Missing functionality will be added in future commits if the approach seems OK:

1. There is currently no equivalent target to test-packages
2. The Dockerfiles still build the agent instead of copying over a built binary


rfratto:
> When the cache is present, subsequent builds are sped up dramatically, from multiple minutes to 32 seconds. 

This currently only applies to non-arm64 hosts since there's no arm64 variant of grafana/agent-build-image yet. All of the base images have arm64 variants, so arm64 users can create the build image locally to make sure builds are as fast as possible until there is an official arm64 image.

rfratto:
I'm looking for feedback on the general approach right now to see if it makes more sense over the existing Makefile. Not all of the build targets work right now (in particular, the `dist*` one don't work), so YMMV if you try running these locally.

marctc:
> It includes targets only needed for local development and testing. Targets only intended for CI are purposefully omitted to keep the Makefile as simple as possible.

I like this approach, it keep the Makefile for the simpler purpose to make development faster and reduce burden to memorize commands.

Alternatively we could reduce the size of the Makefile by creating Makefiles per package and importing them in the main one. I believe in `docs` we have something like that, but we could apply also to other packages (for example to generate CRDs). But it's only an idea, I already like it as it is.

rfratto:
Marking this as ready for review now. Everything works with the exception of the packaging targets, which I'll finish tomorrow.

rfratto:
Just pushed a change to add all the packaging targets. All the targets should work locally now.

Makefile-v2.mk
```diff
@@ -0,0 +1,296 @@
...
+##   CGO_ENABLED      Set to 0 to disable Cgo for builds
+##   RELEASE_BUILD    Set to 1 to build release binaries
+##   VERSION          Version to inject into built binaries.
+##   GO_TAGS          Extra tags to use when building.
+
+# TODO(rfratto): test-packages target
+#
+# This depends on some reworking of how the Go tests in ./packaging works to
+# assume that packages have already been built so we don't have to hook in
+# packaging code here.
```
rfratto, line 89: I'm leaving this purposefully unfinished for this PR, I'll add the target once this replaces the existing Makefile.

Makefile-v2.mk
```diff
@@ -35,6 +35,14 @@
 ##   crow-image      Builds crow Docker image.
 ##   smoke-image     Builds smoke test Docker image.
 ##
+## Targets for packaging:
+##
+##   dist                   Produce release assets for everything.
+##   dist-agent-binaries    Produce release-ready agent binaries.
+##   dist-agentctl-binaries Produce release-ready agentctl binaries.
+##   dist-packages          Produce release-ready DEB and RPM packages.
+##   dist-agent-intaller    Produce a Windows installer for Grafana Agent.
```
rlankfo, line 44:
```suggestion
##   dist-agent-installer    Produce a Windows installer for Grafana Agent.
```

Makefile-v2.mk
```diff
@@ -0,0 +1,296 @@
...
+##                          subsequent proxied builds
+##   clean                  Clean caches and built binaries
+##   help                   Displays this message
+##   info                   Print Makefile-specific environment variables
+##
+## Environment variables:
+##
+##   USE_CONTAINER    Set to 1 to enable proxying commands to build container
+##   AGENT_IMAGE      Image name:tag built by `make agent-image`
+##   AGENTCTL_IMAGE   Image name:tag built by `make agentctl-image`
```
rlankfo, line 67: probably out of scope for this PR, but to simplify things a bit it seems we could combine agentctl, smoke, and crow images into a single agent tools image.

Makefile-v2.mk
```diff
@@ -0,0 +1,296 @@
...
+SMOKE_BINARY    ?= build/agent-smoke
+GOOS            ?= $(shell go env GOOS)
+GOARCH          ?= $(shell go env GOARCH)
+GOARM           ?= $(shell go env GOARM)
+CGO_ENABLED     ?= 1
+RELEASE_BUILD   ?= 0
+
+# This should contain the list of all environment variables should should
+# propagate to the build container. USE_CONTAINER should _not_ be included to
+# avoid infinite recursion.
```
rlankfo, line 111:
```suggestion
# This should contain the list of all environment variables which will
# propagate to the build container. USE_CONTAINER should _not_ be included to
# avoid infinite recursion.
```

Makefile-v2.mk
```diff
@@ -0,0 +1,296 @@
...
+PROPAGATE_VARS := \
+	AGENT_IMAGE AGENTCTL_IMAGE OPERATOR_IMAGE CROW_IMAGE SMOKE_IMAGE \
+	BUILD_IMAGE GOOS GOARCH GOARM CGO_ENABLED RELEASE_BUILD \
+	AGENT_BINARY AGENTCTL_BINARY OPERATOR_BINARY CROW_BINARY SMOKE_BINARY \
+	VERSION GO_TAGS
+
+#
+# Contants for targets
```
rlankfo, line 119:
```suggestion
# Constants for targets
```

Makefile-v2.mk
```diff
@@ -0,0 +1,296 @@
...
+.PHONY: lint
+lint:
+	golangci-lint run -v --timeout=10m
+
+.PHONY: test
+# We have to run test twice: once for all packages with -race and then once
+# more without -race for packages that have known race detection issues.
+test:
+	$(GO_ENV) go test $(CGO_FLAGS) -race ./...
+	$(GO_ENV) go test $(CGO_FLAGS) ./pkg/integrations/node_exporter ./pkg/logs ./pkg/operator ./pkg/util/k8s
```
rlankfo, line 160: I assume this step will succeed as long as the second go test is successful. How will we easily detect when a new package introduces race conditions?

Makefile-v2.mk
```diff
@@ -0,0 +1,296 @@
...
+	@printf "GOARM           = $(GOARM)\n"
+	@printf "CGO_ENABLED     = $(CGO_ENABLED)\n"
+	@printf "RELEASE_BUILD   = $(RELEASE_BUILD)\n"
+	@printf "VERSION         = $(VERSION)\n"
+	@printf "GO_TAGS         = $(GO_TAGS)\n"
+
+# awk magic to print out the comment block at the top of this file.
+.PHONY: help
+help:
+	@awk 'BEGIN {FS="## "} /^##\s*(.*)/ { print $$2 }' $(MAKEFILE_LIST)
```
rlankfo, line 296: Definitely like this help target 👍  I believe grafana/grafana does something similar, taken from this blog post I believe: https://www.thapaliya.com/en/writings/well-documented-makefiles/.  It seems we could take this a step further, take a look and see what you think.

cmd/agent-operator/Dockerfile.new
```diff
@@ -0,0 +1,35 @@
+# syntax=docker/dockerfile:1.4
+
+# NOTE: This Dockerfile can only be built using BuildKit. BuiltKit is used by
```
rlankfo, line 3:
```suggestion
# NOTE: This Dockerfile can only be built using BuildKit. BuildKit is used by
```

rlankfo: Looks like this typo is in other Dockerfiles as well

Makefile-v2.mk
```diff
@@ -0,0 +1,296 @@
+## Build, test, and generate code for various parts of Grafana Agent.
+##
+## At least Go 1.18, git, and a moderately recent version of Docker is required
```
rlankfo, line 3: might want to mention protoc is required as well to run generate-protos

tools/make/build-container.mk
```diff
@@ -0,0 +1,96 @@
...
+#
+
+GO_CACHE_VOLUME    := grafana-agent-build-container-gocache
+GO_MODCACHE_VOLUME := grafana-agent-build-container-gomodcache
+
+define volume_exists
+$(shell docker volume inspect $(1) >/dev/null 2>&1 && echo 1 || echo "")
+endef
+
+# Figure out if our build container
```
rlankfo, line 55: ?

tools/make/packaging.mk
```diff
@@ -0,0 +1,230 @@
...
+	$(call generate_fpm,deb,amd64,amd64,$(PACKAGE_PREFIX).amd64.deb)
+	$(call generate_fpm,rpm,x86_64,amd64,$(PACKAGE_PREFIX).amd64.rpm)
+endif
+
+.PHONY: dist-packages-arm64
+dist-packages-arm64: dist/agent-linux-arm64 dist/agentctl-linux-arm64
+ifeq ($(USE_CONTAINER),1)
+	$(RERUN_IN_CONTAINER)
+else
+	$(call generate_fpm,deb,arm64,arm64,$(PACKAGE_PREFIX).arm64.deb)
```
rlankfo, line 185: arm64 twice?

tools/make/packaging.mk
```diff
@@ -0,0 +1,230 @@
...
+	$(call generate_fpm,rpm,armhfp,armv7,$(PACKAGE_PREFIX).armv7.rpm)
+endif
+
+.PHONY: dist-packages-ppc64le
+dist-packages-ppc64le: dist/agent-linux-ppc64le dist/agentctl-linux-ppc64le
+ifeq ($(USE_CONTAINER),1)
+	$(RERUN_IN_CONTAINER)
+else
+	$(call generate_fpm,deb,ppc64el,ppc64le,$(PACKAGE_PREFIX).ppc64el.deb)
+	$(call generate_fpm,rpm,ppc64le,ppc64le,$(PACKAGE_PREFIX).ppc64le.rpm)
```
rlankfo, line 213: it's annoying if fedora and debian use different nomenclature to reference the same architecture, e.g. ppc64le vs ppc64el.  I can't tell if these lines are correct at. glance but it seems it should be double checked.

rlankfo (approved):
Left a few comments, but overall it looks good to me 👍

Makefile-v2.mk
rfratto: Makes sense for the future (though I'm also tempted to combine agentctl + agent into a single binary by utilizing Cobra in Flow)

Makefile-v2.mk
rfratto: I originally started with that, but eventually moved to this because I felt it was cleaner and more flexible to have a giant doc comment at the top of the Makefile, since that means it's also more accessible to people reading the Makefile directly.

Makefile-v2.mk
rfratto: Yeah, this wasn't supposed to be an exhaustive list since it'll probably get out of sync pretty fast. I'll elaborate the comment, thanks!

tools/make/build-container.mk
rfratto: oops :)

tools/make/packaging.mk
rfratto: Yep, see our existing Makefile: https://github.com/grafana/agent/blob/cbbfabb8b7844d702409dc62dcc32fd330973ec5/Makefile#L468-L469

tools/make/packaging.mk
rfratto: They're correct, see our existing Makefile: 

https://github.com/grafana/agent/blob/cbbfabb8b7844d702409dc62dcc32fd330973ec5/Makefile#L474-L475

https://github.com/grafana/agent/blob/cbbfabb8b7844d702409dc62dcc32fd330973ec5/Makefile#L483-L484

(I triple-checked what I wrote here against the current Makefile, since I definitely agree it's confusing)

PR #2087, opened 2022-08-29
docs/flow: explain how to reference a component's exports

rfratto:
Give a detailed explanation for how to reference the exports of a component.

### DIFF
--- a/docs/flow/concepts/configuration_language.md
+++ b/docs/flow/concepts/configuration_language.md
@@ -68,6 +68,10 @@ The most common expression is to reference the exports of a component like
 formed by merging the component's name (e.g., `local.file`), label (e.g.,
 `password_file`), and export name (e.g., `content`), delimited by period.
 
+For components that don't use labels, like
+`prometheus.integration.node_exporter`, only combine the component name with
+export name: `prometheus.integration.node_exporter.targets`.
+
 ## Blocks
 
 _Blocks_ are used to configure components and groups of attributes. Each block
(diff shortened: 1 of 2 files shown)


docs/flow/concepts/configuration_language.md
```diff
@@ -61,6 +61,13 @@ supports more complex expressions, such as:
 
 Expressions may be used for any attribute inside a component definition.
 
+### Referencing component exports
+
+The most common expression is to reference the exports of a component like
+`local.file.password_file.content`. A reference to a component's exports is
+formed by merging the component's name (e.g., `local.file`), label (e.g.,
```
mattdurham, line 68: should we add `label if exists`?

rfratto: Good point. It feels a little weird to squeeze into the sentence, but I'll add a second sentence for components that don't use labels

docs/flow/concepts/configuration_language.md
```diff
@@ -61,6 +61,17 @@ supports more complex expressions, such as:
...
+`password_file`), and export name (e.g., `content`), delimited by period.
+
+For components that don't use labels, like
+`prometheus.integration.node_exporter`, only combine the component name with
+export name: `prometheus.integration.node_exporter.targets`.
```
rfratto, line 73: We don't actually have a merged component which doesn't use a label right now, so we might need to remove this by v0.28 if #1872 doesn't get merged.

mattdurham (approved):
LGTM

PR #2091, opened 2022-08-29
remove armv6 container image

captncraig:
After switching to [ubuntu base images](#2078), we noticed that there isn't support for armv6 in the base image anymore.

This leaves us with a few options:

1. Switch back to debian for armv6 support. This would continue supporting rpis under 3. But debian has more security issues than ubuntu.
2. Discontinue armv6 docker images. We still publish binaries and packages, and we can close some of the cves for other platforms.

This PR is for the second one. If we want to go that route.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -135,7 +135,7 @@ seego = docker run --init --rm $(DOCKER_OPTS) $(MOD_MOUNT) -v "$(CURDIR):$(CURDI
 docker-build = docker build $(DOCKER_BUILD_FLAGS)
 ifeq ($(CROSS_BUILD),true)
 DOCKERFILE = Dockerfile.buildx
-docker-build = docker buildx build --push --platform linux/amd64,linux/arm64,linux/arm/v6,linux/arm/v7,linux/ppc64le $(DOCKER_BUILD_FLAGS)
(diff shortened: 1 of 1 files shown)


rfratto:
This would be a breaking change that we'd need to call out very explicitly in the changelog and upgrade guide. 

We don't have pull stats per image/platform, but there is at least one machine which is pulling the arm/v6 images, though it might be a bot. That being said, I believe the userbase of people running Docker on ARMv6 is probably pretty low, so this is probably a mostly safe change. 

If anyone _is_ using ARMv6 (which hasn't been used since Raspberry Pi 2), then we'll still be producing armv6 binaries for them to run on the machine itself, and they could produce their own Docker images if running the agent on Docker is absolutely needed.

rfratto:
I'd personally prefer this approach over mixing in Debian images for armv6. We can find a solution if it becomes a problem, but I suspect it won't become an issue.

rfratto:
After discussing offline, we agreed to move forward with this for now, but there needs to be a big notice in the changelog that armv6 Docker images will no longer be produced due to the switch, but armv6 binaries and system packages will continue to be distributed.

rfratto (approved):

PR #2094, opened 2022-08-30
Add flow API support

### LINKED ISSUE #2007: Flow: Compress /debug/graph and /debug/config endpoints

mattdurham:
#### PR Description

This adds the initial support for the flow API, necessary for the UI portion.

Fixes #2007.


pkg/flow/flow.go
```diff
@@ -209,9 +214,66 @@ func (c *Flow) LoadFile(f *File) error {
 	return diags.ErrorOrNil()
 }
 
+// ComponentInfos returns the component infos.
+func (c *Flow) ComponentInfos() []*river.ComponentField {
+	cns := c.loader.Components()
+	infos := make([]*river.ComponentField, len(cns))
+	refByBacktrack := make(map[string]*river.ComponentField, 0)
+	for i, com := range cns {
+		refs, _ := controller.ComponentReferences(com, c.loader.Graph())
```
rfratto, line 223: Can you add a TODO here to change this to use a graph that doesn't have transitive reductions applied? That'll produce some unexpected results for users.

pkg/flow/flow.go
```diff
@@ -209,9 +214,66 @@ func (c *Flow) LoadFile(f *File) error {
 	return diags.ErrorOrNil()
 }
 
+// ComponentInfos returns the component infos.
+func (c *Flow) ComponentInfos() []*river.ComponentField {
+	cns := c.loader.Components()
+	infos := make([]*river.ComponentField, len(cns))
+	refByBacktrack := make(map[string]*river.ComponentField, 0)
+	for i, com := range cns {
```
rfratto, line 222: You should be able to look at the edges of the graph directly instead of having to call `ComponentReferences` (which isn't cheap) or having to construct the backtracking map

pkg/flow/flow.go
```diff
@@ -209,9 +214,66 @@ func (c *Flow) LoadFile(f *File) error {
...
 func (c *Flow) Close() error {
 	c.cancel()
 	<-c.exited
 	return c.sched.Close()
 }
+
+func newFromNode(cn *controller.ComponentNode, refs []controller.Reference) *river.ComponentField {
+	refsStr := make([]string, len(refs))
+	for i, r := range refs {
+		refsStr[i] = strings.Join(r.Target.ID(), ".")
```
rfratto, line 247: nit: use `r.Target.ID().String()` here

pkg/flow/flow.go
```diff
@@ -209,9 +214,66 @@ func (c *Flow) LoadFile(f *File) error {
...
+	}
+	ci := &river.ComponentField{
+		Field: river.Field{
+			ID:   cn.NodeID(),
+			Name: strings.Join(cn.ID()[0:2], "."),
+			Type: "block",
+			Key:  strings.Join(cn.ID(), "."),
```
rfratto, line 254: This worries me a little bit, the first two strings in the ID won't always be the namespace (i.e. `prometheus.integration.node_exporter`.)

I think we should update ComponentNode to store the component name & component label separately instead of us having to extract them from the ID.

pkg/flow/flow_http.go
```diff
@@ -91,3 +94,38 @@ func (f *Flow) configBytes(w io.Writer, debugInfo bool) (n int64, err error) {
...
+			UpdatedTime: ci.Health.UpdatedTime,
+		},
+		"")
+	if err != nil {
+		return 0, err
+	}
+	bb, err := json.MarshalIndent(field, "", "    ")
+	if err != nil {
+		return 0, err
+	}
```
rfratto, line 129: nit: don't emit pretty-printed JSON since it'll really blow up the output. Let tooling format the JSON if we need to look at it as humans.

pkg/flow/flow_http.go
```diff
@@ -91,3 +94,38 @@ func (f *Flow) configBytes(w io.Writer, debugInfo bool) (n int64, err error) {
 
 	return file.WriteTo(w)
 }
+
+// ComponentJSON returns the json representation of the flow component.
+func (c *Flow) ComponentJSON(w io.Writer, ci *river.ComponentField) (int, error) {
```
rfratto, line 99: nit: is the int return field here useful?

pkg/river/json_walker.go
```diff
@@ -0,0 +1,409 @@
...
+	Value interface{} `json:"value,omitempty"`
+	Body  []*Field    `json:"body,omitempty"`
+}
+
+// Health represents the health of a component.
+type Health struct {
+	State       string    `json:"state"`
+	Message     string    `json:"message"`
+	UpdatedTime time.Time `json:"updatedTime"`
+}
```
rfratto, line 43: I would prefer avoiding adding Flow-specific pieces into the River code and we can decouple the two here. The code here can force on creating a JSON value for a River block, body, or value, and then the rest can live in pkg/flow.

pkg/river/json_walker.go
```diff
@@ -0,0 +1,409 @@
+package river
```
rfratto, line 1: I don't think this logic belongs in the top-level river package. Maybe `pkg/river/encoding/riverjson`?

pkg/river/json_walker.go
```diff
@@ -0,0 +1,409 @@
...
+	}
+	// Handle items that explicitly use tokenizer, these are always considered capsule values.
+	if tkn, ok := in.(builder.Tokenizer); ok {
+		tokens := tkn.RiverTokenize()
+		return &Field{
+			Type:  "capsule",
+			Value: tokens[0].Lit,
+		}, nil
+	}
+	newV := value.MakeValue(vIn)
```
rfratto, line 273: This should be a call to `value.Encode`

pkg/river/json_walker.go
```diff
@@ -0,0 +1,409 @@
...
+	rt := value.RiverType(reflect.TypeOf(in))
```
rfratto, line 274: This should be a call to `newV.Type()`

pkg/river/json_walker.go
```diff
@@ -0,0 +1,409 @@
...
+}
+
+func isValue(t reflect.Type) bool {
+	rt := value.RiverType(t)
+	switch rt {
+	case value.TypeNull, value.TypeNumber, value.TypeString, value.TypeBool, value.TypeFunction, value.TypeCapsule:
+		return true
+	}
+	return false
+}
```
rfratto, line 256: Unless I'm missing something, this always returns true

rfratto (commented):
I gave this a first pass, but I need more time to carefully review json_walker.go. It feels like there's more opportunities for it to reuse logic from river/internal/value and I want to take the time to find them so we don't have to spend too much time maintaining the json conversion.

pkg/river/json_walker.go
rfratto: Oh I see, you're not considering arrays or objects to be values, but they are. I think this might need a more clear name

pkg/river/json_walker.go
```diff
@@ -0,0 +1,409 @@
...
+			if found != nil {
+				fields = append(fields, found)
+			}
+		}
+	}
+	return fields, nil
+}
+
+// convertThing is the switchboard for conversions
+func convertThing(in interface{}, f *rivertags.Field) (*Field, error) {
```
rfratto, line 143: Can we give this a better name? The naming is making this a bit harder to review. 

If it helps, the hierarchy is always:

- Root is a body 
- A body contains a list of attributes and blocks 
- An attribute is a name, value pair 
- A block contains a name, optional label, and inner body
- Values are either nulls, bools, numbers, strings, functions, capsules, arrays, or objects.
- Arrays contain a list of values.
- Objects contain a list of name, value pairs.

pkg/river/json_walker.go
rfratto: To add a little extra flavor on _why_ I think we should decouple things: IMO this introduces a "soft" cyclic dependency, where `pkg/flow` imports `pkg/river` but `pkg/river` has Flow-specific concepts like component components.

IMO it makes the separation of concerns a little blurry and might make it harder to reason about code as we continue adding new functionality.

cmd/agent/flow_run.go
```diff
@@ -145,6 +151,13 @@ func (fr *flowRun) Run(configFile string) error {
 		}
 
 		r := mux.NewRouter()
+		// Add gzip if the client requests it to all requests.
+		r.Use(func(h http.Handler) http.Handler {
+			return gziphandler.GzipHandler(h)
```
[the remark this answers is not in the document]
mattdurham: Sounds good

pkg/flow/flow_http.go
mattdurham: With GZIP enabled it added 10% size to pretty print. From a debugging standpoint, it's nice to not have to pretty print but if we think that's important I can remove it.

pkg/flow/flow.go
mattdurham: Good idea.

pkg/flow/flow.go
mattdurham: Cached the edges pre-reduce for later usage.

pkg/river/json_walker.go
mattdurham: Moved things around, so the actual river json only knows about the most basic building blocks: blocks and attrs.

mattdurham:
Rewrote / changed pretty much the whole thing, its more code but I think much clearer on what is happening.

pkg/river/encoding/array.go
```diff
@@ -0,0 +1,87 @@
...
+	type temp ArrayField
+	return json.Marshal((*temp)(af))
+}
+
+func (af *ArrayField) isValid() bool {
+	if af == nil {
+		return false
+	}
+	return len(af.valueFields)+len(af.structFields)+len(af.mapFields)+len(af.arrayFields) > 0
+}
```
rfratto, line 49: The "isValid" naming here is a little confusing. Why is an empty array "invalid?" Should the method name here be different? Ditto for the other isValid implementations.

pkg/river/internal/rivertags/rivertags.go
```diff
@@ -184,6 +184,28 @@ func Get(ty reflect.Type) []Field {
...
+
+		_, tagged := field.Tag.Lookup("river")
+		if !tagged {
+			continue
+		}
+		return true
+	}
+
+	return false
+}
```
rfratto, line 207: It seems like this isn't called anywhere, should it be removed?

pkg/river/encoding/json.go
```diff
@@ -0,0 +1,149 @@
+package encoding
```
rfratto, line 1: nit: I think this file should be called encoding.go so it's clearer that it's the entrypoint of the package.

rfratto (commented):
Nice! This is looking much better.

pkg/flow/flow.go
```diff
@@ -209,9 +213,81 @@ func (c *Flow) LoadFile(f *File) error {
...
+}
+
+// ComponentField represents a component in river.
+type ComponentField struct {
+	encoding.Field `json:",omitempty"`
+	ID             string           `json:"id,omitempty"`
+	Label          string           `json:"label,omitempty"`
+	References     []string         `json:"referencesTo"`
+	ReferencedBy   []string         `json:"referencedBy"`
+	Health         *ComponentHealth `json:"health"`
```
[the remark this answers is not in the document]
mattdurham: I was torn a bit, I didn't know if there would be a situation where a component wouldn't have health. Not really sure how we treat `global` state. Though that might be a special case the API handles differently.

pkg/river/encoding/json.go
```diff
@@ -0,0 +1,149 @@
...
+
+	"github.com/grafana/agent/pkg/river/internal/rivertags"
+	"github.com/grafana/agent/pkg/river/internal/value"
+)
+
+const attr = "attr"
+const object = "object"
+
+// ConvertComponentChild is used to convertBase arguments, exports, health and debuginfo.
+func ConvertComponentChild(input interface{}) ([]interface{}, error) {
```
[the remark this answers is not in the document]
mattdurham: Do we consider debuginfo and health blocks?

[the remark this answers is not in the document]
mattdurham: I went with child since I couldnt think of a better word, exports was taken, info is already over used but def open to changing the naming

pkg/river/encoding/array.go
```diff
@@ -0,0 +1,87 @@
...
+
+func (af *ArrayField) convertArray(val value.Value) error {
+	if !isArray(val) {
+		return fmt.Errorf("convertArray requires a field that is an slice/array got %T", val.Interface())
+	}
+	af.Type = "array"
+	structs := make([]*StructField, 0)
+	values := make([]*ValueField, 0)
+	arrays := make([]*ArrayField, 0)
+	maps := make([]*MapField, 0)
```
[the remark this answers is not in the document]
mattdurham: IMO it makes it easier to understand and allows for better type checking. I tried to get away from using `interface{}` where I could.

pkg/river/encoding/json.go
```diff
@@ -0,0 +1,149 @@
...
+	return fields, nil
+}
+
+func isValue(val value.Value) bool {
+	switch val.Type() {
+	case value.TypeNull, value.TypeNumber, value.TypeString, value.TypeBool, value.TypeFunction, value.TypeCapsule:
+		return true
+	}
+	return false
+}
```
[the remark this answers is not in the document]
mattdurham: Will use `isFieldValue` since that is what we are really checking.

pkg/river/encoding/json.go
```diff
@@ -0,0 +1,149 @@
...
+	return val.Reflect().Kind() == reflect.Array || val.Reflect().Kind() == reflect.Slice
+}
+
+func isMap(val value.Value) bool {
+	return val.Reflect().Kind() == reflect.Map
+}
+
+func isStruct(val value.Value) bool {
+	return val.Reflect().Kind() == reflect.Struct
+}
```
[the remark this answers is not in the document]
mattdurham: Refactored out isStruct which allowed simplification.

pkg/river/encoding/array.go
rfratto: hasValue seems better, thanks!

pkg/flow/flow_http.go
```diff
@@ -91,3 +95,52 @@ func (f *Flow) configBytes(w io.Writer, debugInfo bool) (n int64, err error) {
...
+	}
+
+	var err error
+	args, err := encoding.ConvertComponentChild(foundComponent.Arguments())
+	if err != nil {
+		return err
+	}
+	if args != nil && !reflect.ValueOf(args).IsNil() {
+		ci.Arguments = args
+	}
```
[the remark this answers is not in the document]
mattdurham: Cleaned this up by using a length check.

pkg/flow/flow.go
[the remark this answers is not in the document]
rfratto: Not a blocker, we can leave it alone as-is for now.

pkg/flow/internal/controller/component.go
```diff
@@ -169,6 +170,12 @@ func getExportsType(reg component.Registration) reflect.Type {
 // ID returns the component ID of the managed component from its River block.
 func (cn *ComponentNode) ID() ComponentID { return cn.id }
 
+// Label returns the label for the block or "" if none was specified.
+func (cn *ComponentNode) Label() string { return cn.label }
+
+// NodeType returns the component's type, i.e. `local.file.test` returns `local.file`.
+func (cn *ComponentNode) NodeType() string { return cn.nodeType }
```
rfratto, line 177: Since we're aligning on terminology, this should probably be renamed to `ComponentName()` rather than `NodeType()`

pkg/flow/internal/dag/dag.go
```diff
@@ -157,6 +158,9 @@ func (g *Graph) Edges() []Edge {
 	return edges
 }
 
+// NonTransitiveEdges is the original edge list before the reduce function.
+func (g *Graph) NonTransitiveEdges() []Edge { return g.originalEdges }
```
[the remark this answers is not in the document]
rfratto: Bump on this; I'd still like to see this changed.

pkg/river/encoding/encoding.go
```diff
@@ -0,0 +1,139 @@
...
+
+func isArray(val value.Value) bool {
+	return val.Type() == value.TypeArray
+}
+
+func isMapOrStruct(val value.Value) bool {
+	return val.Type() == value.TypeObject
+}
+
+func convertRiverValue(val value.Value) (vf *ValueField, af *ArrayField, mf *MapField, err error) {
```
rfratto, line 126: The function signature here is a little intense and will suffer from scaling issues if we add new field types. I'd prefer it if we switch this to returning some interface, error tuple. 

It doesn't have to strictly be `interface{}` and can be a marker interface like River's [`ast.Expr`](https://github.com/grafana/agent/blob/116237054a2c2d09b43ff03cc0e20f56ee1be8d8/pkg/river/ast/ast.go#L24-L28), which is referenced anywhere that accepts an expression like [`ast.AttributeStmt`](https://github.com/grafana/agent/blob/116237054a2c2d09b43ff03cc0e20f56ee1be8d8/pkg/river/ast/ast.go#L56-L60).

For example, you could do:

```go 
type encodingField interface {
  json.Marshaler 
  isEncodingField() 
}
```

(json.Marshaler is embedded since you expect them to implement MarshalJSON).

This should let you remove the isValid/hasValue methods which should help clean up the code quite a bit.

pkg/river/encoding/encoding.go
```diff
@@ -0,0 +1,139 @@
...
+	default:
+		return nil, fmt.Errorf("unable to convert %T", val.Interface())
+	}
+}
+
+func isArray(val value.Value) bool {
+	return val.Type() == value.TypeArray
+}
+
+func isMapOrStruct(val value.Value) bool {
```
rfratto, line 122: Let's name these two isRiverArray and isRiverObject.

pkg/river/encoding/array_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/grafana/agent/pkg/river/internal/value"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSimpleArray(t *testing.T) {
+	intArr := make([]int, 10)
+	for i := 0; i < 10; i++ {
+		intArr[i] = i
+	}
+	reqString := `{"type":"array","value":[{"type":"number","value":0},{"type":"number","value":1},{"type":"number","value":2},{"type":"number","value":3},{"type":"number","value":4},{"type":"number","value":5},{"type":"number","value":6},{"type":"number","value":7},{"type":"number","value":8},{"type":"number","value":9}]}`
```
rfratto, line 16: Can we split this across multiple lines to make it easier to read? JSONEq ignores formatting of the JSON strings. Ditto with the other tests in this file and other files.

pkg/river/encoding/array.go
```diff
@@ -0,0 +1,79 @@
...
+		return fmt.Errorf("convertArray requires a field that is an slice/array got %T", val.Interface())
+	}
+	af.Type = "array"
+	values := make([]*ValueField, 0)
+	arrays := make([]*ArrayField, 0)
+	maps := make([]*MapField, 0)
+
+	for i := 0; i < val.Len(); i++ {
+		arrEle := val.Index(i).Interface()
+		arrVal := value.Encode(arrEle)
```
rfratto, line 58: This is expensive and does a copy, just use val.Index(i) directly:

```suggestion
		arrVal := val.Index(i)
```

pkg/river/encoding/attribute.go
```diff
@@ -0,0 +1,80 @@
...
+	return json.Marshal((*temp)(af))
+}
+
+func (af *AttributeField) convertAttribute(val value.Value, f rivertags.Field) error {
+	if !f.IsAttr() {
+		return fmt.Errorf("convertAttribute requires a field that is an attribute got %T", val.Interface())
+	}
+	if val.Reflect().IsZero() {
+		return nil
+	}
```
rfratto, line 61: The value API already has an IsZero method which behaves slightly differently; you should be calling that instead here.

```suggestion
	if val.IsZero() {
		return nil
	}
```

pkg/river/encoding/block.go
```diff
@@ -0,0 +1,86 @@
...
+	return len(bf.blockFields)+len(bf.attributeFields) > 0
+}
+
+func (bf *BlockField) convertBlock(val value.Value, f rivertags.Field) error {
+	if val.Reflect().Kind() != reflect.Struct {
+		return fmt.Errorf("convertBlock cannot work on interface or slices")
+	}
+	if val.Interface() == nil {
+		return nil
+	}
```
rfratto, line 54: Will this ever be true? Do you mean to check for `val.Type() == value.TypeNull` here?

pkg/river/encoding/block.go
```diff
@@ -0,0 +1,86 @@
...
+	if val.Reflect().IsZero() {
```
rfratto, line 55: Can you bring out `val.Reflect()` into a variable? It's called a few times in this method and I'm worried about the cost of copying the value each time.

pkg/river/encoding/encoding.go
```diff
@@ -0,0 +1,139 @@
+package encoding
```
rfratto, line 1: API nit: we can greatly reduce the number of things that need to get exported here if the only exported method is 

```go 
func MarshalBodyJSON(input interface{}) ([]byte, error) 
``` 

Then in types from other packages you would use [`json.RawMessage`](https://pkg.go.dev/encoding/json#RawMessage) to expose the rendered JSON:

```go
// ComponentInfo represents a component in river.
type ComponentInfo struct {
	encoding.Field `json:",omitempty"`
	ID             string           `json:"id,omitempty"`
	Label          string           `json:"label,omitempty"`
	References     []string         `json:"referencesTo"`
	ReferencedBy   []string         `json:"referencedBy"`
	Health         *ComponentHealth `json:"health"`
	Original       string           `json:"original"`
	Arguments      json.RawMessage  `json:"arguments,omitempty"`
	Exports        json.RawMessage  `json:"exports,omitempty"`
	DebugInfo      json.RawMessage  `json:"debugInfo,omitempty"`
}
```

That would be a huge improvement over the exposed API and would keep things very tidy.

pkg/river/encoding/block.go
```diff
@@ -0,0 +1,86 @@
...
+	Label           string        `json:"label,omitempty"`
+	Body            []interface{} `json:"body,omitempty"`
+	attributeFields []*AttributeField
+	blockFields     []*BlockField
+}
+
+func newBlock(val value.Value, f rivertags.Field) (*BlockField, error) {
+	bf := &BlockField{}
+	return bf, bf.convertBlock(val, f)
+}
```
rfratto, line 26: Blocks are structural and are _not_ values (even though they can be encoded as them); this is the one place where we should be using the reflect API down the call chain.

(i.e., use reflect.Value here in `newBlock` and in `convertBlock`)

pkg/river/encoding/encoding.go
```diff
@@ -0,0 +1,139 @@
...
+
+	"github.com/grafana/agent/pkg/river/internal/rivertags"
+	"github.com/grafana/agent/pkg/river/internal/value"
+)
+
+const attr = "attr"
+const object = "object"
+
+// ConvertRiverBlock is used to convertBase arguments, exports, health and debuginfo.
+func ConvertRiverBlock(input interface{}) ([]interface{}, error) {
```
rfratto, line 17: nit: `input` will always be a River body, never a River block. 

River blocks are named bodies, and the name information is always lost from a Go value. This is why in `river/token/builder` you have to [create a block separately](https://github.com/grafana/agent/blob/9e279c9bb43291f4d1b8242334219519779562eb/pkg/river/token/builder/builder.go#L271-L280) before you can [encode the body into it](https://github.com/grafana/agent/blob/9e279c9bb43291f4d1b8242334219519779562eb/pkg/river/token/builder/builder.go#L119-L132).

rfratto (commented):
Getting much closer! I think this is my final round of comments.

pkg/river/encoding/encoding.go
mattdurham: Simplified it quite a bit in the latest version.

pkg/flow/internal/controller/loader.go
```diff
@@ -87,7 +88,8 @@ func (l *Loader) Apply(parentScope *vm.Scope, blocks []*ast.BlockStmt) diag.Diag
 		diags = append(diags, multierrToDiags(err)...)
 		return diags
 	}
-
+	// Copy the original graph
```
rfratto, line 91: Can you add detail in the comment to explain why we're doing this? (so we can expose the non-reduced graph to users to examine original connections)

pkg/river/encoding/array.go
rfratto: bump

pkg/river/encoding/attribute.go
rfratto: bump

pkg/river/encoding/encoding.go
rfratto: bump

This is a nit for now, but I'd like for us to start thinking about the interfaces our packages expose.

pkg/river/encoding/encoding.go
rfratto: bump

pkg/river/encoding/encoding.go
rfratto: bump

pkg/river/encoding/encoding.go
```diff
@@ -0,0 +1,140 @@
...
+
+func convertRiverValue(val value.Value) (RiverField, error) {
+	if isFieldValue(val) {
+		return convertValue(val)
+	} else if isArray(val) {
+		return newArray(val)
+	} else if isMapOrStruct(val) {
+		return newMap(val)
+	}
+	return nil, fmt.Errorf("unknown value %T", val.Interface())
```
rfratto, line 139: style nit: this should probably be a `switch` case at this point

rfratto:
(The CI is failing because of changes to the Makefile; you'll need to rebase on top of latest main)

pkg/river/encoding/encoding.go
mattdurham: Removed them and implemented the check inline since it was used in one place.

pkg/river/encoding/array_test.go
mattdurham: It's a lot more verbose to format it out adds 150 lines but definitely can.

pkg/river/encoding/block.go
mattdurham: I wanted to handle the case: the two should be the equivalent but if value.TypeNull is easier to understand I will change to that.

rfratto: My reasoning here is that `val.Interface()` and `val.Reflect()` effectively convert the River value back to a Go value, which should be used extremely sparingly and only when it makes sense. 

If we're dealing with River values, we should be using the River API as much as we can.

pkg/river/encoding/encoding.go
mattdurham: Reduced the scope significantly.

pkg/river/encoding/attribute.go
```diff
@@ -0,0 +1,61 @@
...
+	af := &attributeField{}
+	return af, af.convertAttribute(val, f)
+}
+
+func (af *attributeField) isValid() bool {
+	if af == nil || af.valField == nil {
+		return false
+	}
+	return af.valField.hasValue()
+}
```
rfratto, line 28: Did you mean to rename this to hasValue rather than isValid? There's a few other methods still named isValid too.

pkg/river/encoding/encoding.go
```diff
@@ -0,0 +1,148 @@
...
+
+const attr = "attr"
+const object = "object"
+
+// riverField is an interface that wraps the various concrete options for a river value.
+type riverField interface {
+	hasValue() bool
+}
+
+// ConvertRiverBodyToJSON is used to convertBase arguments, exports, health and debuginfo.
```
rfratto, line 22: Comment is stale and still mentions Flow-specific comments

```suggestion
// ConvertRiverBodyToJSON converts a River body value to a JSON representation.
```

web/api/api.go
```diff
@@ -0,0 +1,89 @@
...
+//
+//   Go runtime, build information (like the Prometheus page)
+//
+// /api/v0/web/status/flags
+//
+//   Command-line flags used to launch application
+//
+// /api/v0/web/status/config-file
+//
+//   Parsed config file (*not* evaluated config file)
```
rfratto, line 89:
```suggestion
```

Let's take all this out for now.

pkg/river/encoding/block.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"github.com/grafana/agent/pkg/river/internal/rivertags"
+	"github.com/grafana/agent/pkg/river/internal/value"
+)
+
+// blockField represents the json representation of a river block.
+type blockField struct {
+	field           `json:",omitempty"`
+	ID              string        `json:"id,omitempty"`
+	Label           string        `json:"label,omitempty"`
```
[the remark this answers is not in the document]
rfratto: See pkg/river/token/builder for an example of how I pull out the label: https://github.com/grafana/agent/blob/e6240381773ddd61cc150e17ddbe42625a33f430/pkg/river/token/builder/builder.go#L188-L191

pkg/flow/flow.go
```diff
@@ -215,9 +218,73 @@ func (c *Flow) LoadFile(f *File) error {
...
+		Health: &ComponentHealth{
+			State:       h.Health.String(),
+			Message:     h.Message,
+			UpdatedTime: h.UpdateTime,
+		},
+	}
+	return ci
+}
+
+// ComponentInfo represents a component in river.
```
rfratto, line 270:
```suggestion
// ComponentInfo represents a Flow component.
```

rfratto, line 270: The comment here feels a little weird. It's not representing a component in river, necessarily. Should this say "represents a component in Flow?"

pkg/river/encoding/attribute.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/grafana/agent/pkg/river/internal/rivertags"
+	"github.com/grafana/agent/pkg/river/internal/value"
+)
+
+// attributeField represents a json representation of a river attribute.
```
rfratto, line 12:
```suggestion
// attributeField represents a JSON representation of a River attribute.
```

pkg/river/token/builder/builder.go
```diff
@@ -131,6 +131,18 @@ func (b *Body) AppendFrom(goValue interface{}) {
...
+func GetBlockLabel(rv reflect.Value) string {
+	tags := rivertags.Get(rv.Type())
+	for _, tag := range tags {
+		if tag.Flags&rivertags.FlagLabel != 0 {
+			return rv.FieldByIndex(tag.Index).String()
+		}
+	}
+
+	return ""
+}
```
rfratto, line 144: I'd prefer not to make this part of the public API for token/builder since it feels out of scope for the purpose of the package. Can we just make a copy of the unexported function or move this into river/internal/value or something?

pkg/river/encoding/block.go
```diff
@@ -0,0 +1,90 @@
...
+
+func (bf *blockField) convertBlock(reflectValue reflect.Value, f rivertags.Field) error {
+	for reflectValue.Kind() == reflect.Pointer {
+		if reflectValue.IsNil() {
+			return nil
+		}
+		reflectValue = reflectValue.Elem()
+	}
+	if reflectValue.Kind() != reflect.Struct {
+		return fmt.Errorf("convertBlock cannot work on interface or slices")
```
rfratto, line 58:
```suggestion
		return fmt.Errorf("convertBlock can only be called on struct kinds, got %s", reflectValue.Kind())
```

pkg/river/encoding/block.go
```diff
@@ -0,0 +1,90 @@
...
+		// Blocks can only have sub blocks and attributes
+		if rf.IsBlock() {
+			newBF, err := newBlock(reflect.ValueOf(fieldVal), rf)
+			if err != nil {
+				return nil
+			}
+			if newBF.hasValue() {
+				bf.blockFields = append(bf.blockFields, newBF)
+			}
+		} else if rf.IsAttr() {
```
rfratto, line 79: Array blocks can happen here too, not just for the top-level block. I'd recommend creating a helper function that both `convertBlock` and `ConvertRiverBodyToJSON` use to convert the body of a block into a list of output fields.

pkg/river/encoding/encoding.go
```diff
@@ -0,0 +1,148 @@
...
+	"reflect"
+
+	"github.com/grafana/agent/pkg/river/token/builder"
+
+	"github.com/grafana/agent/pkg/river/internal/rivertags"
+	"github.com/grafana/agent/pkg/river/internal/value"
+)
+
+const attr = "attr"
+const object = "object"
```
rfratto, line 15: nit: decorate the names here to make it more clear how they're intended to be used, like `attrType` and `objectType`.

pkg/river/encoding/encoding.go
```diff
@@ -0,0 +1,148 @@
...
+
+// riverField is an interface that wraps the various concrete options for a river value.
+type riverField interface {
+	hasValue() bool
+}
+
+// ConvertRiverBodyToJSON is used to convert a River body value to a JSON representation.
+func ConvertRiverBodyToJSON(input interface{}) (json.RawMessage, error) {
```
rfratto, line 23: nit: return `[]byte` here which is more idiomatic for encoding functions.

pkg/river/internal/value/number_value.go
```diff
@@ -23,6 +23,7 @@ const (
 	NumberKindFloat
 )
 
+// makeNumberKind converts a go kind to a river kind.
```
rfratto, line 26:
```suggestion
// makeNumberKind converts a Go kind to a River kind.
```

web/api/api.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/gorilla/mux"
+	"github.com/grafana/agent/pkg/flow"
+)
+
+// FlowAPI wraps several calls for the component health api.
+type FlowAPI struct {
+	flow *flow.Flow
+}
+
+// NewFlowAPI instantiates a new flow api.
```
rfratto, line 24:
```suggestion
// NewFlowAPI instantiates a new Flow API.
```

web/api/api.go
```diff
@@ -0,0 +1,77 @@
...
+	"net/http"
+	"path"
+
+	"github.com/prometheus/prometheus/util/httputil"
+
+	"github.com/gorilla/mux"
+	"github.com/grafana/agent/pkg/flow"
+)
+
+// FlowAPI wraps several calls for the component health api.
```
rfratto, line 19:
```suggestion
// FlowAPI wraps several calls for the component health API.
```

rfratto: (also, this comment seems inaccurate/incomplete. "component health API?")

web/api/api.go
```diff
@@ -0,0 +1,77 @@
...
+type FlowAPI struct {
+	flow *flow.Flow
+}
+
+// NewFlowAPI instantiates a new flow api.
+func NewFlowAPI(flow *flow.Flow, r *mux.Router) *FlowAPI {
+	return &FlowAPI{flow: flow}
+}
+
+// RegisterRoutes registers all the routes.
```
rfratto, line 29:
```suggestion
// RegisterRoutes registers all the API's routes.
```

web/api/api.go
```diff
@@ -0,0 +1,77 @@
...
+				_, _ = w.Write(bb.Bytes())
+				return
+			}
+		}
+
+		http.NotFound(w, r)
+	}
+}
+
+// JSON returns the json representation of ComponentInfoDetailed.
```
rfratto, line 69:
```suggestion
// JSON returns the JSON representation of c.
```

web/api/api.go
```diff
@@ -0,0 +1,77 @@
...
+func (f *FlowAPI) JSON(c *flow.ComponentInfo) (bytes.Buffer, error) {
```
rfratto, line 70: Returning `bytes.Buffer` seems not very idiomatic, I'd suggest `[]byte` here instead.

rfratto, line 70: I think we can unexport this method, it's only used internally within the package.

rfratto (approved):
LGTM after final comments. Nice work!

pkg/river/token/builder/builder.go
mattdurham: Will make a copy.

PR #2093, opened 2022-08-30
docs/flow: add config-language/_index.md

rfratto:

PR #2096, opened 2022-08-31
river: add documentation for language concepts

tpaschalis:

#### PR Description
This is an initial stab at adding slightly more detailed documentation for River's files, the language syntax and handling of components.

#### Notes to the Reviewer
I think that there's significant overlap with what we thought would go under `config-language/_index.md` in #2093 which leads me to think that we should have something more or different here? 

I'm not entirely sure _what_ that would be though right now, looking forward to your feedback

#### PR Checklist

- [ ] CHANGELOG updated (N/A)
- [X] Documentation added
- [ ] Tests updated (N/A)

### DIFF
--- a/docs/flow/config-language/components.md
+++ b/docs/flow/config-language/components.md
@@ -16,6 +16,8 @@ controller_]({{< relref "../concepts/component_controller.md" >}}), who is
 responsible for scheduling them, reporting their health and debug status, and
 continuously re-evaluating their inputs and outputs.
 
+
+## Configuring components
 Each top-level River _block_ will instantiate a new component. All components
 are identified by their name, describing what the component is responsible for,
 while some allow or require to provide an extra user-specified _label_.
@@ -62,6 +64,7 @@ local.file "targets" {
 }
 ```
 
+## Referencing components
 To wire components together, one can use the Exports of one as the Arguments
 to another.
 
@@ -96,10 +99,4 @@ you can wire components together.
 In the previous example, the contents of the `local.file.target.content`
 expression must first be evaluated in a concrete value then type-checked and
 substituted into `prometheus.scrape.default` for it to be configured in turn.
-
-River attributes that appear as Arguments and Exports can carry around more
-than the basic types you'd expect such as strings, numbers or booleans; they
-are able to natively wrap any Go value (eg. channels or interfaces), as well as
-pass around and invoke Go functions which makes for powerful tooling when
-building more complex pipelines.
 
--- a/docs/flow/config-language/components.md
+++ b/docs/flow/config-language/components.md
@@ -16,42 +16,34 @@ controller_]({{< relref "../concepts/component_controller.md" >}}), who is
 responsible for scheduling them, reporting their health and debug status, and
 continuously re-evaluating their inputs and outputs.
 
-Each top-level River _block_ will instantiate a new component. A component
-consists of its name (which describes what the component is responsible for) as
-well as an optional user-specified label.
+Each top-level River _block_ will instantiate a new component. All components
+are identified by their name, describing what the component is responsible for,
+while some allow or require to provide an extra user-specified _label_.
 
 You can see a list of all available components [here]({{< relref "../components/_index.md" >}}).
-Each component features a complete reference page, so getting a component to
-work for you should be as easy as reading its documentation and copy/pasting
-from an example.
+Each one features a complete reference page, so getting a component to work for
+you should be as easy as reading its documentation and copy/pasting from an
+example.
 
 ## Arguments and Exports
 Most user interactions with components will center around two basic concepts;
 _Arguments_ and _Exports_.
 
-* _Arguments_ work much like function arguments in your favorite programming
- language. They can be any number of attributes or nested unlabeled blocks and
-are used to configure the component's behavior. Most components will feature
-both required and optional arguments. Any arguments that are not provided will
-take on their default values.
+* _Arguments_ are settings which modify the behavior of a component. They can
+ be any number of attributes or nested unlabeled blocks, some of them being
+required for the component to work and some being optional. Any optional
+arguments that are not overriden, will take on their default values.
 
 * _Exports_ are zero or more named return values that can be referred to by
- other components. There is no limit to what an export might be; it can take on
-any River value or Go type (eg. channels, interfaces).
-
-Each component's documentation page contains the list of all available
-Arguments and resulting Exports, along their underlying type. Users can
-configure Arguments either using a constant value or an
-[_expression_]({{< relref "./expressions/_index.md" >}}) to continuously
-re-evaluate values who are dynamic in nature.
+ other components and can be any River value.
 
 Here's a quick example; the following block defines a `local.file` component
 labelled "targets". The `local.file.targets` component will then expose the
 file `content` as a string in its Exports.
 
 The `filename` attribute is a _required_ argument; the user can also define a
-number of _optional_ ones (in this case  `detector`, `poll_frequency` and
-`is_sensitive`), which configure how and how often the file should be polled
+number of _optional_ ones, in this case `detector`, `poll_frequency` and
+`is_sensitive`, which configure how and how often the file should be polled
 as well as whether its contents are safe to be presented as plaintext back to
 the user.
 
@@ -60,19 +52,18 @@ local.file "targets" {
 	// Required Argument
 	filename = "/etc/agent/targets" 
 
-	/* Optional Arguments
-	is_sensitive   = <boolean>
-	detector       = < "fsnotify" | "poll" >
-	poll_frequency = <duration> 
-	*/
+	// Optional Arguments
+	//   is_sensitive   = <boolean>
+	//   detector       = < "fsnotify" | "poll" >
+	//   poll_frequency = <duration> 
 
 	// Exports: a single field named `content`
 	// It can be referred to as `local.file.token.content`
 }
 ```
 
-To wire components together, you just use the Exports of one as the Arguments
-of another.
+To wire components together, one can use the Exports of one as the Arguments
+to another.
 
 For example, here's a component that scrapes Prometheus metrics. The `targets`
 field is populated with two scrape targets; a constant one `localhost:9001` and
@@ -102,61 +93,13 @@ River will type-check expressions before assigning a value to an attribute; the
 documentation of each component will have more information about the ways that
 you can wire components together.
 
+In the previous example, the contents of the `local.file.target.content`
+expression must first be evaluated in a concrete value then type-checked and
+substituted into `prometheus.scrape.default` for it to be configured in turn.
+
 River attributes that appear as Arguments and Exports can carry around more
 than the basic types you'd expect such as strings, numbers or booleans; they
 are able to natively wrap any Go value (eg. channels or interfaces), as well as
 pass around and invoke Go functions which makes for powerful tooling when
 building more complex pipelines.
 
-## DAG
-The relationships between components form a [Directed Acyclic Graph](https://en.wikipedia.org/wiki/Directed_acyclic_graph).
-
-This graph is used by River to resolve dependencies between components, set the
-correct order for component evaluation, and ensure the validity of references.
-River does not allow components that reference themselves, any other
-relationship that might end up in cyclical references, and will reject outright
-invalid references in expressions.
-
-In the previous example, the contents of the `local.file.target.content`
-expression must first be evaluated in a concrete value then type-checked and
-substituted into `prometheus.scrape.default` for it to be configured in turn.
-
-## Component Lifecycle
-Each time a River configuration file is reloaded or re-evaluated, Flow will:
-
-- Parse the provided configuration file. In case any errors are found, the
-  Agent will continue using the last valid River configuration.
-- Parse and apply settings for predefined global blocks (like `logging`).
-- Create a graph of component relationships and validate it.
-- Create new components that weren't present in the last loaded configuration
-  file.
-- Drop components that no longer exist in the new configuration file.
-- Update any pre-existing components whose arguments have changed.
-
-## Error reporting
-Like in the case of syntax errors, River will also enrich component-related
-errors with useful information to help with debugging.
-
-```
-Error: ./cmd/agent/example-config.river:16:21: component "local.file.this_files.content" does not exist
-
-15 |         "__address__"   = "localhost:12345",
-16 |         "dynamic_label" = local.file.this_files.content,
-   |                           ^^^^^^^^^^^^^^^^^^^^^^^
-17 |     }]
-
-Error: ./cmd/agent/example-config.river:25:1: Failed to build component: building component: duplicate remote write configs are not allowed, found duplicate for URL: http://localhost:9009/api/prom/push
-
-24 |
-25 | | prometheus.remote_write "default" {
-   |  _^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-26 | |     remote_write {
-27 | |         url = "http://localhost:9009/api/prom/push"
-28 | |     }
-29 | |     remote_write {
-30 | |         url = "http://localhost:9009/api/prom/push"
-31 | |     }
-32 | | }
-   | |_^
-33 |
-```
--- a/docs/flow/config-language/components.md
+++ b/docs/flow/config-language/components.md
@@ -45,9 +45,8 @@ file `content` as a string in its Exports.
 
 The `filename` attribute is a _required_ argument; the user can also define a
 number of _optional_ ones, in this case `detector`, `poll_frequency` and
-`is_sensitive`, which configure how and how often the file should be polled
-as well as whether its contents are safe to be presented as plaintext back to
-the user.
+`is_secret`, which configure how and how often the file should be polled
+as well as whether its contents are sensitive or not.
 
 ```river
 local.file "targets" {
@@ -55,7 +54,7 @@ local.file "targets" {
 	filename = "/etc/agent/targets" 
 
 	// Optional Arguments
-	//   is_sensitive   = <boolean>
+	//   is_secret      = <boolean>
 	//   detector       = < "fsnotify" | "poll" >
 	//   poll_frequency = <duration> 
 
--- a/docs/flow/config-language/components.md
+++ b/docs/flow/config-language/components.md
@@ -8,7 +8,7 @@ weight: 300
 # Components
 Components are the defining feature of Grafana Agent Flow. They are autonomous
 pieces of business logic that perform a single task (like retrieving secrets or
-collecting Prometheus metrics) and are wired together to form programmable
+collecting Prometheus metrics) and can be wired together to form programmable
 pipelines of telemetry data.
 
 Under the hood, components are orchestrated via the [_component
@@ -33,11 +33,11 @@ _Arguments_ and _Exports_.
 
 * _Arguments_ are settings which modify the behavior of a component. They can
  be any number of attributes or nested unlabeled blocks, some of them being
-required for the component to work and some being optional. Any optional
-arguments that are not overriden, will take on their default values.
+required and some being optional. Any optional arguments that are not
+overriden, will take on their default values.
 
-* _Exports_ are zero or more named return values that can be referred to by
- other components and can be any River value.
+* _Exports_ are zero or more output values that can be referred to by other
+  components, and can be any River type.
 
 Here's a quick example; the following block defines a `local.file` component
 labelled "targets". The `local.file.targets` component will then expose the
@@ -70,7 +70,7 @@ to another.
 
 For example, here's a component that scrapes Prometheus metrics. The `targets`
 field is populated with two scrape targets; a constant one `localhost:9001` and
-an expression that ties the target to the contents of
+an expression that ties the target to the value of
 `local.file.target.content`.
 
 ```river
--- a/docs/flow/config-language/components.md
+++ b/docs/flow/config-language/components.md
@@ -39,6 +39,12 @@ take on their default values.
  other components. There is no limit to what an export might be; it can take on
 any River value or Go type (eg. channels, interfaces).
 
+Each component's documentation page contains the list of all available
+Arguments and resulting Exports, along their underlying type. Users can
+configure Arguments either using a constant value or an
+[_expression_]({{< relref "./expressions/_index.md" >}}) to continuously
+re-evaluate values who are dynamic in nature.
+
 Here's a quick example; the following block defines a `local.file` component
 labelled "targets". The `local.file.targets` component will then expose the
 file `content` as a string in its Exports.
@@ -87,6 +93,10 @@ prometheus.scrape "default" {
 }
 ```
 
+Every time the file contents change and the `local.file` component re-evaluates
+its Exports, the value will propagated to `prometheus.scrape` so it can start
+scraping the new target.
+
 All Arguments and Exports have an underlying [type]({{< relref "./expressions/types_and_values.md" >}}).
 River will type-check expressions before assigning a value to an attribute; the
 documentation of each component will have more information about the ways that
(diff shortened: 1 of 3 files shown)


docs/flow/config-language/syntax.md
```diff
@@ -6,3 +6,138 @@ weight: 200
...
+file is not important; the language will consider all direct and indirect
+dependencies between elements to determine the order which must be used for the
+evaluation.
+
+## Attributes and Blocks
+
+### Attributes
+_Attributes_ are used to configure individual settings. They always take the
+form of `<ATTRIBUTE_NAME> = <ATTRIBUTE_VALUE>`. They can appear either as
+top-level elements or nested within blocks.
```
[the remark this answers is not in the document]
tpaschalis: I mean they currently cannot, but not sure whether that's due to the language design or because we currently don't have any use for them to.

docs/flow/config-language/components.md
```diff
@@ -6,3 +6,157 @@ weight: 300
...
+Each top-level River _block_ will instantiate a new component. A component
+consists of its name (which describes what the component is responsible for) as
+well as an optional user-specified label.
+
+You can see a list of all available components [here]({{< relref "../components/_index.md" >}}).
+Each component features a complete reference page, so getting a component to
+work for you should be as easy as reading its documentation and copy/pasting
+from an example.
+
+## Arguments and Outputs
```
rfratto, line 28:
```suggestion
## Arguments and Exports
```

docs/flow/config-language/syntax.md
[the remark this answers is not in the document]
rfratto: They can. The root of the AST is a body which is a list of attributes and blocks, we just don't have any top-level attributes defined.

rfratto (commented):
I gave this a very quick first pass, but I'll need to spend more time later today slowly reviewing it. 

My initial concern is how much Flow-specific concepts are being taught here, like the DAG and component lifecycle, which have nothing to do with the configuration language. I think we should be careful to minimize the amount of Flow-specific information here so readers aren't forced to read about River when they want to know how the Flow DAG and graph behaves (for example). 

IMO, the only Flow-specific things users need to know in the River documentation is:

1. A component is configured via arguments, exposes exports 
2. How to define a component 
3. How to reference the exports of a component

docs/flow/config-language/components.md
```diff
@@ -6,3 +6,157 @@ weight: 300
...
+
+## Arguments and Exports
+Most user interactions with components will center around two basic concepts;
+_Arguments_ and _Exports_.
+
+* _Arguments_ work much like function arguments in your favorite programming
+ language. They can be any number of attributes or nested unlabeled blocks and
+are used to configure the component's behavior. Most components will feature
+both required and optional arguments. Any arguments that are not provided will
+take on their default values.
```
rfratto, line 36: Not all of our users will know how to write code; can we change analogies here? I think we can think of them as settings which modify the behavior of the component.

docs/flow/config-language/components.md
```diff
@@ -6,3 +6,157 @@ weight: 300
...
+
+* _Exports_ are zero or more named return values that can be referred to by
```
rfratto, line 38: Let's not mention Go things here which might overwhelm people. Our audience might not be programmers. We can say something like an export might be any River value, and then later explain in types and values that a capsule is a value which abstracts something internally used by Flow.

docs/flow/config-language/components.md
```diff
@@ -6,3 +6,157 @@ weight: 300
...
+pipelines of telemetry data.
+
+Under the hood, components are orchestrated via the [_component
+controller_]({{< relref "../concepts/component_controller.md" >}}), who is
+responsible for scheduling them, reporting their health and debug status, and
+continuously re-evaluating their inputs and outputs.
+
+Each top-level River _block_ will instantiate a new component. A component
+consists of its name (which describes what the component is responsible for) as
+well as an optional user-specified label.
```
rfratto, line 21: I think we need to be more specific here: labels _aren't_ optional. Some components require them, some components require that you _don't_ have them.

docs/flow/config-language/components.md
```diff
@@ -6,3 +6,157 @@ weight: 300
...
+documentation of each component will have more information about the ways that
+you can wire components together.
+
+River attributes that appear as Arguments and Exports can carry around more
+than the basic types you'd expect such as strings, numbers or booleans; they
+are able to natively wrap any Go value (eg. channels or interfaces), as well as
+pass around and invoke Go functions which makes for powerful tooling when
+building more complex pipelines.
+
+## DAG
```
rfratto, line 111: I think we should drop this section and explain it elsewhere.

docs/flow/config-language/components.md
```diff
@@ -6,3 +6,157 @@ weight: 300
...
+correct order for component evaluation, and ensure the validity of references.
+River does not allow components that reference themselves, any other
+relationship that might end up in cyclical references, and will reject outright
+invalid references in expressions.
+
+In the previous example, the contents of the `local.file.target.content`
+expression must first be evaluated in a concrete value then type-checked and
+substituted into `prometheus.scrape.default` for it to be configured in turn.
+
+## Component Lifecycle
```
rfratto, line 124: I think we should drop this section and explain it elsewhere.

docs/flow/config-language/components.md
```diff
@@ -6,3 +6,157 @@ weight: 300
 ---
 
 # Components
```
rfratto, line 8: Overall, I would prefer that this document talk solely about components from the perspective of River: 

1. How does a user configure components? 
2. What are arguments? What are exports? 
3. How does a user reference a component? 

Going into this page, readers should already be familiar with the concepts of what components are and how the component controller works; we can link them to the other pages as a reminder without having to reexplain those ideas.

This is the one page of the River docs where I think we _can_ mention some Flow-specific things, since this is the only part of River that Flow directly impacts. Everything else should probably avoid saying things which are only true for Flow.

docs/flow/config-language/components.md
```diff
@@ -6,3 +6,157 @@ weight: 300
...
+
+* _Exports_ are zero or more named return values that can be referred to by
+ other components. There is no limit to what an export might be; it can take on
+any River value or Go type (eg. channels, interfaces).
+
+Each component's documentation page contains the list of all available
+Arguments and resulting Exports, along their underlying type. Users can
+configure Arguments either using a constant value or an
+[_expression_]({{< relref "./expressions/_index.md" >}}) to continuously
+re-evaluate values who are dynamic in nature.
```
rfratto, line 46: I think the concepts page already links to the reference list of components and explains how reevaluation works, doesn't it? We should probably remove this paragraph and leave it to the concept-level explanation.

docs/flow/config-language/components.md
```diff
@@ -6,3 +6,157 @@ weight: 300
...
+```river
+local.file "targets" {
+	// Required Argument
+	filename = "/etc/agent/targets" 
+
+	/* Optional Arguments
+	is_sensitive   = <boolean>
+	detector       = < "fsnotify" | "poll" >
+	poll_frequency = <duration> 
+	*/
```
rfratto, line 67: nit: because we won't have syntax highlighting on the rendered website, block comments might look confusing to the user. I'd prefer always using `//`-style comments until then.

Can we find another way of representing this?

docs/flow/config-language/components.md
```diff
@@ -6,3 +6,157 @@ weight: 300
...
+- Parse the provided configuration file. In case any errors are found, the
+  Agent will continue using the last valid River configuration.
+- Parse and apply settings for predefined global blocks (like `logging`).
+- Create a graph of component relationships and validate it.
+- Create new components that weren't present in the last loaded configuration
+  file.
+- Drop components that no longer exist in the new configuration file.
+- Update any pre-existing components whose arguments have changed.
+
+## Error reporting
```
rfratto, line 136: I think we can move the enriched error messages to the index file as a generic "this is what River can do" rather than giving specific examples for errors you might get for components, syntax, or expressions.

docs/flow/config-language/files.md
```diff
@@ -6,3 +6,18 @@ weight: 100
...
+file may be referred to as a 'configuration file', or a 'River configuration'.
+
+River files are required to be UTF-8 encoded, and are permitted to contain
+Unicode characters. River files can use both Unix-style line endings (LF) and
+Windows-style line endings (CRLF), but formatters may replace all line endings
+with Unix-style ones.
+
+There is experimental support for River in
+[vim](https://github.com/rfratto/vim-river) and in
+[VSCode](https://github.com/rfratto/vscode-river).
```
rfratto, line 19: Can we put tooling like this in its own header and call out that it's community tooling?

docs/flow/config-language/files.md
```diff
@@ -6,3 +6,18 @@ weight: 100
...
+
+To learn more about the basic elements that form a River configuration file,
+check out the [Syntax]({{< relref "../../controller-config/logging.md" >}})
+guide next.
```
rfratto, line 23: nit: we shouldn't have to guide people through paragraphs, they can use the sidebar to navigate and expect that they should be reading top to bottom.

docs/flow/config-language/syntax.md
```diff
@@ -6,3 +6,135 @@ weight: 200
...
+```river
+local.file "token" {
+	filename  = "/etc/agent/service-account-token"
+	is_secret = true
+}
+```
+
+## Comments
+River configuration files support single-line `//` as well as `/* */` multiline
+comments.
```
rfratto, line 86: nit: call `/*`-style comments "block" comments instead of multiline.

docs/flow/config-language/syntax.md
```diff
@@ -6,3 +6,135 @@ weight: 200
...
+
+## Error reporting
+One of the main improvements River brings to the table is its error reporting.
+Whenever a syntax error is emitted, River will decorate it with information
+about the position of the error, some surrounding context as well as a possible
+solution.
```
rfratto, line 92: Ditto from before: I think we can have one error reporting section in the index file for the config language instead of defining it multiple times.

docs/flow/config-language/syntax.md
```diff
@@ -6,3 +6,135 @@ weight: 200
...
+		// Nested block body
+	}
+}
+```
+
+The `<BLOCK_NAME>` must correspond to a registered Flow [component]({{< relref "./components.md" >}}).
+
+If the optional `<BLOCK_LABEL>` is set, it has to be a valid River identifier:
+consisting of one or more UTF-8 letters, digits or underscores, but it cannot
+begin with a digit.
```
rfratto, line 67: Rather than giving the rules for a valid identifier multiple times, can we have a dedicated header for defining what an "identifier" is?

rfratto (commented):
nice work, this is coming along nicely :)

docs/flow/config-language/components.md
tpaschalis: I've changed it to use single-line comments. Not sure about another way to represent this I'll try to come up with something, but ideally, I'd like to show that a component can work without defining all of its Arguments.

docs/flow/config-language/files.md
tpaschalis: I personally find it useful to have another way of saying "hey, click here to continue reading" instead of expecting that they're following the sidebar, especially in mobile devices or resized browser windows where the sidebar might be hidden altogether.

I can still remove it if you still think it looks out of place, just let me know ^^

docs/flow/config-language/syntax.md
tpaschalis: Yes, it was there in the first place, putting it back now.

docs/flow/config-language/syntax.md
```diff
@@ -6,3 +6,135 @@ weight: 200
...
+```
+// Pattern for creating a block with an optional user-label:
+<BLOCK NAME> ["<BLOCK LABEL>"] {
+	// Block body can contain attributes and nested unlabeled blocks
+	<IDENTIFIER> = <EXPRESSION> // Attribute
+
+	<NESTED_BLOCK_NAME> {
+		// Nested block body
+	}
+}
```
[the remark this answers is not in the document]
tpaschalis: Eh, yes. I didn't want to re-introduce the same templates you provided in _index.md, but it looks cleaner that way.

docs/flow/config-language/syntax.md
```diff
@@ -6,3 +6,135 @@ weight: 200
...
+The River syntax is designed to be easy to read and write. Essentially, there
+are just two high-level elements to it; _Attributes_ and  _Blocks_. 
+
+The main purpose of River is to define components for the Grafana Agent Flow
+mode, so that they can be wired together to form programmable telemetry
+pipelines.
+As such, the ordering of blocks and attributes within the River configuration
+file is not important; the language will consider all direct and indirect
+dependencies between elements to determine the order which must be used for the
+evaluation.
```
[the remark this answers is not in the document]
tpaschalis: You're right. Still, it's the language itself which resolves the direct and indirect dependencies between block and expressions, right? Or is it Flow again?

docs/flow/config-language/components.md
tpaschalis: I've added the relevant headers but I'm not sure if we should add more content to each.

docs/flow/config-language/components.md
tpaschalis: I've reworded this, let me know if it reads okay or is still awkward.

docs/flow/config-language/components.md
tpaschalis: Should we talk about "scope" in one of these documents? Also, should we talk about the config endpoint and how we're re-creating the config file to be shown back to the user?

rfratto: > Should we talk about "scope" in one of these documents? 

I think the "scope" is probably still evolving so we might not want to be too detailed, but it does makes sense to say in the components page that only components can reference other components.

> Also, should we talk about the config endpoint and how we're re-creating the config file to be shown back to the user?

Maybe not right now, that config endpoint might go away in favor of the UI. (Though how the endpoint works feels like an implementation detail that the user shuoldn't need to know about)

docs/flow/config-language/syntax.md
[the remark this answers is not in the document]
rfratto: > You're right. Still, it's the language itself which determines the direct and indirect dependencies between block and expressions, right? Or is it Flow again?

Flow is responsible for converting the component blocks into a graph and using that to determine evaluation order, River has no concept of blocks having any semantic meaning other than a grouping of related attributes.

docs/flow/config-language/components.md
```diff
@@ -6,3 +6,96 @@ weight: 300
 ---
 
 # Components
+Components are the defining feature of Grafana Agent Flow. They are autonomous
+pieces of business logic that perform a single task (like retrieving secrets or
+collecting Prometheus metrics) and can be wired together to form programmable
+pipelines of telemetry data.
+
+Under the hood, components are orchestrated via the [_component
+controller_]({{< relref "../concepts/component_controller.md" >}}), who is
```
rfratto, line 15:
```suggestion
controller_]({{< relref "../concepts/component_controller.md" >}}), which is
```

docs/flow/config-language/components.md
```diff
@@ -6,3 +6,96 @@ weight: 300
...
+responsible for scheduling them, reporting their health and debug status, and
+continuously re-evaluating their inputs and outputs.
```
rfratto, line 17: I'd like to avoid the word "continuous" here, since it sounds like components are always being reevaluated no matter what, implying that it might be heavy on CPU usage.

Can we change this to say that the component controller determines when components need to be reevaluated?

rfratto: (also, nit: exports don't get evaluated, only arguments do)

docs/flow/config-language/components.md
```diff
@@ -6,3 +6,96 @@ weight: 300
...
+
+
+## Configuring components
+Each top-level River _block_ will instantiate a new component. All components
+are identified by their name, describing what the component is responsible for,
+while some allow or require to provide an extra user-specified _label_.
```
rfratto, line 23: There's non-component blocks like `logging` that are permissible at the top level.

docs/flow/config-language/components.md
```diff
@@ -6,3 +6,96 @@ weight: 300
...
+
+You can see a list of all available components [here]({{< relref "../components/_index.md" >}}).
+Each one features a complete reference page, so getting a component to work for
+you should be as easy as reading its documentation and copy/pasting from an
+example.
```
rfratto, line 28: I vaguely remember seeing a guideline that recommended avoiding using "see here" type links, though I can't remember from where.

Can we rephrase to something like 

```
The [components reference](LINK) shows a list of all defined components. Each component has a compete reference page, ...  
```

docs/flow/config-language/components.md
```diff
@@ -6,3 +6,96 @@ weight: 300
...
+
+## Arguments and Exports
+Most user interactions with components will center around two basic concepts;
+_Arguments_ and _Exports_.
+
+* _Arguments_ are settings which modify the behavior of a component. They can
+ be any number of attributes or nested unlabeled blocks, some of them being
+required and some being optional. Any optional arguments that are not set, will
+take on their default values.
```
rfratto, line 37: comma splice, I think

```suggestion
required and some being optional. Any optional arguments that are not set will
take on their default values.
```

docs/flow/config-language/components.md
```diff
@@ -6,3 +6,96 @@ weight: 300
...
+	// Required Argument
+	filename = "/etc/agent/targets" 
+
+	// Optional Arguments
+	//   is_secret      = <boolean>
+	//   detector       = < "fsnotify" | "poll" >
+	//   poll_frequency = <duration> 
+
+	// Exports: a single field named `content`
+	// It can be referred to as `local.file.token.content`
```
rfratto, line 62:
```suggestion
	// It can be referred to as `local.file.targets.content`
```

docs/flow/config-language/components.md
```diff
@@ -6,3 +6,96 @@ weight: 300
...
+Each top-level River _block_ will instantiate a new component. All components
+are identified by their name, describing what the component is responsible for,
+while some allow or require to provide an extra user-specified _label_.
+
+You can see a list of all available components [here]({{< relref "../components/_index.md" >}}).
+Each one features a complete reference page, so getting a component to work for
+you should be as easy as reading its documentation and copy/pasting from an
+example.
+
+## Arguments and Exports
```
rfratto, line 30: nit: I'm genuinely not sure if we should be using proper nouns for arguments/exports. We can leave it as is for now and talk to a docs writer about it later.

docs/flow/config-language/components.md
```diff
@@ -6,3 +6,96 @@ weight: 300
...
+```
+
+## Referencing components
+To wire components together, one can use the Exports of one as the Arguments
+to another.
+
+For example, here's a component that scrapes Prometheus metrics. The `targets`
+field is populated with two scrape targets; a constant one `localhost:9001` and
+an expression that ties the target to the value of
+`local.file.target.content`.
```
rfratto, line 73:
```suggestion
`local.file.targets.content`.
```

docs/flow/config-language/components.md
```diff
@@ -6,3 +6,96 @@ weight: 300
...
+
+```river
+prometheus.scrape "default" {
+	targets = [
+		{ "__address__" = local.file.target.content }, // tada!
```
rfratto, line 78:
```suggestion
		{ "__address__" = local.file.targets.content }, // tada!
```

docs/flow/config-language/components.md
```diff
@@ -6,3 +6,96 @@ weight: 300
...
+	forward_to = [prometheus.remote_write.default.receiver]
+	scrape_config {
+		job_name = "default"
+	}
+}
+```
+
+Every time the file contents change and the `local.file` component re-evaluates
+its Exports, the value will propagated to `prometheus.scrape` so it can start
+scraping the new target.
```
rfratto, line 91: nit: New exports get exported/exposed/provided, but not reevaluted: 

> Every time the file changes on disk, the `local.file` component will re-read the file and expose the latest contents through its exports. This event will be propagated to the `prometheus.scrape` referencing the file so it can start scraping the new target.

docs/flow/config-language/components.md
```diff
@@ -6,3 +6,96 @@ weight: 300
...
+
+All Arguments and Exports have an underlying [type]({{< relref "./expressions/types_and_values.md" >}}).
+River will type-check expressions before assigning a value to an attribute; the
+documentation of each component will have more information about the ways that
+you can wire components together.
+
+In the previous example, the contents of the `local.file.target.content`
```
rfratto, line 98:
```suggestion
In the previous example, the contents of the `local.file.targets.content`
```

docs/flow/config-language/syntax.md
```diff
@@ -6,3 +6,88 @@ weight: 200
 ---
 
 # Syntax
+The River syntax is designed to be easy to read and write. Essentially, there
+are just two high-level elements to it: _Attributes_ and  _Blocks_.
```
rfratto, line 10: there's two spaces before blocks :)

```suggestion
are just two high-level elements to it: _Attributes_ and _Blocks_. 
```

docs/flow/config-language/syntax.md
```diff
@@ -6,3 +6,88 @@ weight: 200
...
+dependencies between elements to determine their relationships.
+
+## Attributes and Blocks
+
+### Attributes
+_Attributes_ are used to configure individual settings. They always take the
+form of `<ATTRIBUTE_NAME> = <ATTRIBUTE_VALUE>`. They can appear either as
+top-level elements or nested within blocks.
+
+The following example  sets the `log_level` attribute to `"debug"`.
```
rfratto, line 24: two spaces before sets:

```suggestion
The following example sets the `log_level` attribute to `"debug"`.
```

docs/flow/config-language/syntax.md
```diff
@@ -6,3 +6,88 @@ weight: 200
...
+
+```river
+log_level = "debug"
+```
+
+All `<ATTRIBUTE_NAME>`s must be valid River identifiers.
```
rfratto, line 30: nit: can we link to the identifiers header here?

docs/flow/config-language/syntax.md
```diff
@@ -6,3 +6,88 @@ weight: 200
...
+	// Block body can contain attributes and nested unlabeled blocks
+	<IDENTIFIER> = <EXPRESSION> // Attribute
+
+	<NESTED_BLOCK_NAME> {
+		// Nested block body
+	}
+}
+```
+
+The `<BLOCK_NAME>` must correspond to a registered Flow [component]({{< relref "./components.md" >}}).
```
rfratto, line 67: like above, this is not necessarily true: there are permissible top-level blocks like logging which are not components. 

This can probably say something like "the block names have to be recognized by Flow."

docs/flow/config-language/syntax.md
```diff
@@ -6,3 +6,88 @@ weight: 200
...
+
+If the optional `<BLOCK_LABEL>` is set, it has to be a valid River identifier
```
rfratto, line 69: nit: link to the identifier header here

rfratto (commented):
Looking much better!

docs/flow/config-language/syntax.md
```diff
@@ -6,3 +6,88 @@ weight: 200
 ---
 
 # Syntax
```
rfratto, line 8: Somewhere in here I think we need to explain "terminators," since terminators are needed after every attribute and block. Terminators will show up in error messages for the lexer/parser, so I think we need to define it (I thought about changing the error message but I'm not confident) 

Terminators are currently defined as newlines. Terminators are found if there is a newline following an expression, `]`, `)`, or `}`. Other newlines are effectively ignored and a user may enter as many newlines as they want.

docs/flow/config-language/components.md
tpaschalis: I've changed all instances to lowercase in both documentation PRs. We can change this later if needed.

docs/flow/config-language/components.md
tpaschalis: > but it does makes sense to say in the components page that only components can reference other components.

👍 I've added this under the expressions page that talks about referencing exports.

docs/flow/config-language/syntax.md
```diff
@@ -91,3 +91,10 @@ comments.
 River considers an identifier as valid if it consists of one or more UTF-8
 letters, digits or underscores, but doesn't start with a digit.
 
+## Terminators
+All block and attribute definitions are followed by a newline, which River
+calls a _terminator_, as it terminates the current statement.
+A terminator is found when there's a newline following any expression, `]`,
+`)` or `}`. Other newlines are effectively ignored and a user can enter as many
+newlines as they want.
```
rfratto: I'd like to tweak this slightly:

```suggestion
A newline is treated as terminator when it follows any expression, `]`,
`)` or `}`. Other newlines are ignored by River and and a user can enter as many
newlines as they want.
```

rfratto (approved):
Lovely, ship it!

PR #2106, opened 2022-09-02
river: add documentation for expressions

tpaschalis:

#### PR Checklist

- [ ] CHANGELOG updated (N/A)
- [X] Documentation added
- [ ] Tests updated (N/A)

### DIFF
--- a/docs/flow/config-language/expressions/_index.md
+++ b/docs/flow/config-language/expressions/_index.md
@@ -6,3 +6,25 @@ weight: 400
 ---
 
 # Expressions
+
+Expressions represent or compute values that can be assigned to attributes
+within a configuration.
+
+Basic expressions are literal values, like `"Hello, world!"` or `true`.
+Expressions may also do things like [refer to values][] exported by components,
+perform arithmetic, or [call functions][] from the River standard library as
+well as custom ones, dynamically created by components on the spot.
+
+Expressions can be used when configuring any component. As all component
+Arguments have an underlying [type][], River will type-check expressions before
+assigning the resolved value to an attribute. 
+
+The only exception to this rule are some special reserved blocks like
+[`logging`][] which are used to configure the global behavior of Grafana Agent
+Flow and do not support expressions but can only use literal values.
+
+[refer to values]: {{< relref "./referencing_exports.md" >}}
+[call functions]: {{< relref "./function_calls.md" >}}
+[type]: {{< relref "./expressions/types_and_values.md" >}}
+[`logging`]: {{< relref "../../controller-config/logging.md" >}}
+
--- a/docs/flow/config-language/expressions/_index.md
+++ b/docs/flow/config-language/expressions/_index.md
@@ -15,12 +15,8 @@ Expressions may also do things like [refer to values][] exported by components,
 perform arithmetic, or [call functions][].
 
 Expressions can be used when configuring any component. As all component
-Arguments have an underlying [type][], River will type-check expressions before
+arguments have an underlying [type][], River will type-check expressions before
 assigning the result to an attribute. 
-
-The only exception to this rule are some special reserved blocks like
-[`logging`][] which are used to configure the global behavior of Grafana Agent
-Flow and do not support expressions but can only use literal values.
 
 [refer to values]: {{< relref "./referencing_exports.md" >}}
 [call functions]: {{< relref "./function_calls.md" >}}
--- a/docs/flow/config-language/expressions/_index.md
+++ b/docs/flow/config-language/expressions/_index.md
@@ -16,7 +16,7 @@ perform arithmetic, or [call functions][].
 
 Expressions can be used when configuring any component. As all component
 Arguments have an underlying [type][], River will type-check expressions before
-assigning the resolved value to an attribute. 
+assigning the result to an attribute. 
 
 The only exception to this rule are some special reserved blocks like
 [`logging`][] which are used to configure the global behavior of Grafana Agent
--- a/docs/flow/config-language/expressions/_index.md
+++ b/docs/flow/config-language/expressions/_index.md
@@ -12,8 +12,7 @@ within a configuration.
 
 Basic expressions are literal values, like `"Hello, world!"` or `true`.
 Expressions may also do things like [refer to values][] exported by components,
-perform arithmetic, or [call functions][] from the River standard library as
-well as custom ones, dynamically created by components on the spot.
+perform arithmetic, or [call functions][].
 
 Expressions can be used when configuring any component. As all component
 Arguments have an underlying [type][], River will type-check expressions before
--- a/docs/flow/config-language/expressions/_index.md
+++ b/docs/flow/config-language/expressions/_index.md
@@ -20,6 +20,6 @@ assigning the result to an attribute.
 
 [refer to values]: {{< relref "./referencing_exports.md" >}}
 [call functions]: {{< relref "./function_calls.md" >}}
-[type]: {{< relref "./expressions/types_and_values.md" >}}
+[type]: {{< relref "./types_and_values.md" >}}
 [`logging`]: {{< relref "../../controller-config/logging.md" >}}
 
--- a/docs/flow/config-language/expressions/function_calls.md
+++ b/docs/flow/config-language/expressions/function_calls.md
@@ -6,3 +6,25 @@ weight: 400
 ---
 
 # Function Calls
+Function calls is one more River feature that allows users to build richer
+expresisons.
+
+Functions take zero or more arguments as their input and always return a single
+value as their output. Functions cannot be constructed by users, but can be
+either called from River's standard library, or exported by a compoonent.
+
+## Standard library functions
+River contains a [standard library][] of useful functions. Some enable
+interaction with the host system (eg. reading from an environment variable), or
+allow for more complex expressions (eg. concatenating arrays or decoding JSON
+strings into objects).
+```river
+env("HOME")
+json_decode(local.file.cfg.contents)["namespace"]
+```
+
+## Exported by components
+Some components also expose functions in their named Exports, documented on the
+component's reference page. 
+
+[standard library]: {{< relref "../../reference/stdlib/_index.md" >}}
--- a/docs/flow/config-language/expressions/function_calls.md
+++ b/docs/flow/config-language/expressions/function_calls.md
@@ -13,6 +13,9 @@ Functions take zero or more arguments as their input and always return a single
 value as their output. Functions cannot be constructed by users, but can be
 either called from River's standard library, or when exported by a component.
 
+In case a function fails, the expression will not be evaluated and an error
+will be reported.
+
 ## Standard library functions
 River contains a [standard library][] of useful functions. Some enable
 interaction with the host system (eg. reading from an environment variable), or
--- a/docs/flow/config-language/expressions/function_calls.md
+++ b/docs/flow/config-language/expressions/function_calls.md
@@ -6,7 +6,7 @@ weight: 400
 ---
 
 # Function Calls
-Function calls is one more River feature that allows users to build richer
+Function calls is one more River feature that lets users build richer
 expresisons.
 
 Functions take zero or more arguments as their input and always return a single
@@ -24,7 +24,7 @@ json_decode(local.file.cfg.contents)["namespace"]
 ```
 
 ## Exported by components
-Some components also expose functions in their named Exports, documented on the
-component's reference page. 
+Components can also expose functions in their named Exports, as documented on
+their reference page. 
 
 [standard library]: {{< relref "../../reference/stdlib/_index.md" >}}
--- a/docs/flow/config-language/expressions/function_calls.md
+++ b/docs/flow/config-language/expressions/function_calls.md
@@ -11,7 +11,7 @@ expresisons.
 
 Functions take zero or more arguments as their input and always return a single
 value as their output. Functions cannot be constructed by users, but can be
-either called from River's standard library, or exported by a compoonent.
+either called from River's standard library, or exported by a component.
 
 ## Standard library functions
 River contains a [standard library][] of useful functions. Some enable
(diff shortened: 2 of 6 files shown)


docs/flow/config-language/expressions/function_calls.md
```diff
@@ -6,3 +6,25 @@ weight: 400
...
+
+## Standard library functions
+River contains a [standard library][] of useful functions. Some enable
+interaction with the host system (eg. reading from an environment variable), or
+allow for more complex expressions (eg. concatenating arrays or decoding JSON
+strings into objects).
+```river
+env("HOME")
+json_decode(local.file.cfg.contents)["namespace"]
+```
```
tpaschalis, line 24: The standard library is pretty barren for now. Should we mention that we're looking forward to expand as we identify new use cases, or what would be empower users?

docs/flow/config-language/expressions/operators.md
```diff
@@ -6,3 +6,96 @@ weight: 300
...
+## Assignment operator
+River uses `=` as its assignment operator.
+
+An assignment statement may only assign a single value to a single attribute.
+In assignments, each value must be _assignable_ to the attribute to which it is
+assigned.
+
+* The `null` value can be assigned to any attribute.
+* Numerical, string, boolean, array and object types are assignable to
+  attributes of the corresponding type.
```
tpaschalis, line 68: I think that _functions_ are also assignable, right?

docs/flow/config-language/expressions/referencing_outputs.md
```diff
@@ -1,8 +1,69 @@
...
-- /docs/agent/latest/flow/configuration-language/expressions/referencing-outputs
-title: Referencing component exports
+- /docs/agent/latest/flow/configuration-language/expressions/referencing-exports
+title: Referencing component Exports
 weight: 200
 ---
 
-# Types and Values
+# Referencing component Exports
+Referencing component Exports is what enables River to dynamically configure
```
marctc, line 9: nit: does exports need to be capitalized?

docs/flow/config-language/expressions/referencing_outputs.md
```diff
@@ -1,8 +1,69 @@
...
+prometheus.remote_write "onprem" {
+	remote_write {
+		url = "http://prometheus:9009/api/prom/push"
+	}
+}
+```
+
+In the previous example, we managed to wire together a very simple pipeline by
+writing a few River expressions.
+```
```
marctc, line 45: Do we want to keep this kind of diagrams across the docs? Do they render properly in our docs? else we could use mermaid or something like that

docs/flow/config-language/expressions/function_calls.md
rfratto: We say exactly this in docs/flow/reference/stdlib/_index.md: 

https://github.com/grafana/agent/blob/cbbfabb8b7844d702409dc62dcc32fd330973ec5/docs/flow/reference/stdlib/_index.md?plain=1#L10-L13

docs/flow/config-language/expressions/operators.md
rfratto: Yeah, functions are first-class values here. Capsules are also assignable. 

There are also special exceptions where some types can be [converted](https://github.com/grafana/agent/blob/cbbfabb8b7844d702409dc62dcc32fd330973ec5/pkg/river/internal/value/value.go#L463):

- Numbers can always be converted to strings 
- Strings can be converted to numbers if the string represents a number (otherwise evaluation will fail)

docs/flow/config-language/expressions/types_and_values.md
```diff
@@ -6,3 +6,145 @@ weight: 100
...
+}
+```
+
+## Numbers
+River handles integers, unsigned integers and floating-point values as a single
+'number' type which simplifies writing and reading River configuration files.
+
+When a value has to be decoded into a Go type, River will perform the necessary
+conversion. The language also supports defining numbers using the scientific e
+notation.
```
rfratto, line 73: nit: we shouldn't be talking about implementation details since those for change. I think it's fair to drop this sentence entirely for now.

docs/flow/config-language/expressions/types_and_values.md
```diff
@@ -6,3 +6,145 @@ weight: 100
...
+
+#### Capsules
+
+River has a special type called a `capsule`, which represents an internal type
+used by Grafana Agent for building telemetry pipelines. Capsule values can
+never be constructed by the user, but can be used in expressions and assigned
+to attributes. Capsules are represented to the user as
+`capsule("SOME_INTENRAL_NAME")`. An attribute expecting a capsule can only be
+given a capsule of the same internal type. The specific type of capsule
+expected is explicitly documented for any component which uses or exports them.
```
rfratto, line 46: I know this is something I wrote earlier, but I think I want to change how we communicate what capsules are to users to hide the internal implementation details a little bit. 

I've been thinking that we say that capsules are a _category_ of types which are internal to Grafana Agent, and that each capsule type has a unique name like `capsule("metrics.Receiver")`. 

Then we can say that capsule types are not inter-compatible, and that an attribute expecting a `capsule("metrics.Receiver")` can only be given a value of the `capsule("metrics.Receiver")` type.

docs/flow/config-language/expressions/types_and_values.md
```diff
@@ -6,3 +6,145 @@ weight: 100
...
+used in the `forward_to` attribute of `metrics.scrape`, which expects a list of
+`capsule("metrics.Receiver")`:
+
+```river
+metrics.remote_write "default" {
+  remote_write {
+    url = "http://localhost:9090/api/v1/write"
+  }
+}
+metrics.scrape "default" {
```
rfratto, line 60: nit: format this like riverfmt would:

```suggestion
}

metrics.scrape "default" {
```

docs/flow/config-language/expressions/_index.md
```diff
@@ -6,3 +6,24 @@ weight: 400
...
+Expressions may also do things like [refer to values][] exported by components,
+perform arithmetic, or [call functions][].
+
+Expressions can be used when configuring any component. As all component
+Arguments have an underlying [type][], River will type-check expressions before
+assigning the result to an attribute. 
+
+The only exception to this rule are some special reserved blocks like
+[`logging`][] which are used to configure the global behavior of Grafana Agent
+Flow and do not support expressions but can only use literal values.
```
rfratto, line 23: this isn't strictly true. `logging` can use expressions, it just can't refer to other components in its expressions.

Sorry, this reads like something I wrote; I was wrong :)

docs/flow/config-language/expressions/function_calls.md
```diff
@@ -6,3 +6,25 @@ weight: 400
...
+
+## Exported by components
+Components can also expose functions in their named Exports, as documented on
+their reference page.
```
rfratto, line 28: Since we don't have any components that do this yet, should we remove this section?

docs/flow/config-language/expressions/operators.md
```diff
@@ -6,3 +6,97 @@ weight: 300
...
+```
+{ }    used for defining blocks and objects
+( )    used to group and prioritize expressions
+[ ]    used for defining arrays
+```
+
+In the following example we can see the use of curly braces and square brackets
+to define an object and an array.
+```river
+obj = { app = "agent", namespace = "dev"}
```
rfratto, line 81: nit: formatting like riverfmt would print

```suggestion
obj = { app = "agent", namespace = "dev" }
```

docs/flow/config-language/expressions/operators.md
```diff
@@ -6,3 +6,97 @@ weight: 300
...
+<=    less or equal
+>     greater
+>=    greater or equal
+```
+
+The equality operators `==` and `!=` can be applied to any operands.
+
+On the other hand, for the ordering operators the two operands must both be
+_orderable_ and of the same type. These terms and the result of the comparisons
+are defined as follows:
```
rfratto, line 39: You use the term "ordering operators" but don't define it here. Can we explain which operators are ordering operators?

docs/flow/config-language/expressions/operators.md
```diff
@@ -6,3 +6,97 @@ weight: 300
...
+```
+
+Logical operators apply to boolean values and yield a boolean result.
+
+## Assignment operator
+River uses `=` as its assignment operator.
+
+An assignment statement may only assign a single value to a single attribute.
+In assignments, each value must be _assignable_ to the attribute to which it is
+assigned.
```
rfratto, line 64: nit: I'm not really sure if this is an "expression / operator" thing, it's more of a block thing. 

(also `=` is used in object literals to assign a value to a key, it's not just for attributes)

rfratto (approved):
:shipit: Great start. 

I think we're experiencing some challenges with splitting up the docs work, since it's not clear what content belongs where or if something was already written. 

I think our best move is to merge everything and then clean up duplicated content later.

docs/flow/config-language/expressions/operators.md
tpaschalis: Okay, I'll add some bulletpoints to the assignment rules!

docs/flow/config-language/expressions/referencing_outputs.md
tpaschalis: I'm not sure, but we when decide on something it should be enforced throughout our documentation.

docs/flow/config-language/expressions/referencing_outputs.md
tpaschalis: Not sure if the diagrams are a good idea, looking for feedback on that :) 

I just want to make sure that the concept of referencing outputs is well-explained.

docs/flow/config-language/expressions/function_calls.md
tpaschalis: Yeah, I removed the dedicated section but left the mention of the feature in the overview paragraph above.

docs/flow/config-language/expressions/operators.md
tpaschalis: The original intent of the first sentence here was to say that you cannot have multiple assignments a la `foo, bar = 1, 2`. 

Sorry, I'm not sure what you meant by 'it's more of a block thing' here, could you expand?

rfratto: Ah, sorry, I meant it might be more appropriate to explain the rules of `=` when we talk about blocks and attributes.

docs/flow/config-language/expressions/_index.md
tpaschalis: You're right, I should have verified that. Removing this sentence.

docs/flow/config-language/expressions/operators.md
tpaschalis: Do you recommend dropping the whole 'Assignment operator' section and move it to `syntax.md`? I feel that it's somewhat well tied here as we're discussing the different types under the expressions directory.

rfratto: I don't feel too strongly about it, it's your call. I can see good arguments for both sides here

docs/flow/config-language/expressions/referencing_outputs.md
tpaschalis: I just checked with a mermaid fenced codeblock, but it only seems to be rendered in Github, it doesn't show when browsing after running `make docs`. So let's either keep the ASCII graph or get rid of it altogether.

docs/flow/config-language/expressions/function_calls.md
```diff
@@ -6,3 +6,21 @@ weight: 400
 ---
 
 # Function Calls
+Function calls is one more River feature that lets users build richer
```
rfratto, line 9: Should we also mention that some function calls can fail, and if they fail, evaluation of the expression fails? (json_decode is currently the only function which can fail)

docs/flow/config-language/expressions/referencing_exports.md
```diff
@@ -0,0 +1,69 @@
+---
+aliases:
+- /docs/agent/latest/flow/configuration-language/expressions/referencing-exports
+title: Referencing component exports
+weight: 200
+---
+
+# Referencing component exports
```
rfratto, line 8: I think we should find a place here to note that component exports may only be referenced inside the block or subblocks of another component, and non-component blocks like `logging` can't do it.

docs/flow/config-language/expressions/referencing_exports.md
```diff
@@ -0,0 +1,69 @@
...
+While components can work in isolation, they can be much more versatile when
+one component's behavior and data flow is bound to the exports of another,
+building a dependency relationship between the two.
+
+## Using references
+These references are built by combining the component's name, label and named
+export with dots.
+
+For example, the contents of a file exported by the `local.file` component
+might be referenced as `local.file.target.content`, while a
```
rfratto, line 21: `target` seems like it appears out of nowhere here. Can we be more explicit?

> For example, the contents of a file exported by a `local.file` component labeled `target` is referenced as `local.file.target.content`

rfratto: Ditto for the example right below it.

docs/flow/config-language/expressions/types_and_values.md
```diff
@@ -6,3 +6,143 @@ weight: 100
...
+
+## Numbers
+River handles integers, unsigned integers and floating-point values as a single
+'number' type which simplifies writing and reading River configuration files.
+
+```
+3   == 3.00     // true
+5.0 == (10 / 2) // true
+1e+2 == 100    // true
+2e-3 == 0.002  // true
```
rfratto, line 35: Are these supposed to be aligned?

```suggestion
3    == 3.00     // true
5.0  == (10 / 2) // true
1e+2 == 100      // true
2e-3 == 0.002    // true
```

rfratto (approved):

docs/flow/config-language/expressions/referencing_exports.md
tpaschalis: Yeah, it's better to be more explicit. I'm expanding on this.

PR #2108, opened 2022-09-02
river: fix error message about allowed operands

### LINKED ISSUE #2107: river: broken reporting of allowed operands for binary operations
If someone tries to use a binary operator with non-matching operand types (eg. `is_secret = 3 < true`), the error message reported by River is missing the hint about the supported operand types.

```
ts=2022-09-02T14:01:29.107614Z level=error msg="failed to evaluate component" component=local.file.this_file err="decoding River: ./cmd/agent/example-config.river:11:18: true should be one of \u0001\u0002 for binop <, got bool"

Error: ./cmd/agent/example-config.river:11:18: true should be one of  for binop <, got bool

10 |     detector = "fsnotify"
11 |     is_secret = 3 < true
   |                     ^^^^
12 | }
```

~The reason for that, weirdly enough, is that the `%s` verb will not call the String() method for each element in an array, even though it will do so for a map. The corresponding log shows that Go will try to print the uint8 values instead.~

This issue is showcased [here](https://go.dev/play/p/ivXHtYeHOov); the simplest solution will be to see where we're exposing arrays of `value.Type` and using the `%v` verb instead to construct the error message.

## Edit
The underlying reason is explained [here](https://pkg.go.dev/fmt). While `%s` should be recursively applied to each slice element, a small exception is present for uint8 (byte) slices and string-like verbs 😅 .

The issue disappears if eg. we had defined the types as `type Type uint16`. TIL!

> For compound operands such as slices and structs, the format applies to the elements of each operand, recursively, not to the operand as a whole. Thus %q will quote each element of a slice of strings, and %6.2f will control formatting for each element of a floating-point array.

> However, when printing a byte slice with a string-like verb (%s %q %x %X), it is treated identically to a string, as a single item.

tpaschalis:

#### PR Description
This PR allows us to side-step an edge case in Go's string formatting described in #2107 where `%s` in uint8 (byte) slices doesn't recurse each element to call its String() but treats the whole slice as a string.

Fixes #2107

#### Notes to the Reviewer
The allowed operands are now reported correctly.
```
ts=2022-09-02T14:14:25.417997Z level=error msg="failed to evaluate component" component=local.file.this_file err="decoding River: ./cmd/agent/example-config.river:11:18: true should be one of [number string] for binop <, got bool"
ts=2022-09-02T14:14:25.418241Z component=prometheus.remote_write.default subcomponent=rw level=info remote_name=618d80 url=http://localhost:9009/api/prom/push msg="Replaying WAL" queue=618d80
Error: ./cmd/agent/example-config.river:11:18: true should be one of [number string] for binop <, got bool

10 |     detector = "fsnotify"
11 |     is_secret = 3 < true
   |                     ^^^^
12 | }
```
#### PR Checklist

- [ ] CHANGELOG updated (N/A)
- [ ] Tests updated (N/A)

### DIFF
--- a/pkg/river/vm/op_binary.go
+++ b/pkg/river/vm/op_binary.go
@@ -25,12 +25,12 @@ func evalBinop(lhs value.Value, op token.Token, rhs value.Value) (value.Value, e
 	if !acceptableBinopType(lhs, op) {
 		return value.Null, value.Error{
 			Value: lhs,
-			Inner: fmt.Errorf("should be one of %s for binop %s, got %s", binopAllowedTypes[op], op, lhs.Type()),
+			Inner: fmt.Errorf("should be one of %v for binop %s, got %s", binopAllowedTypes[op], op, lhs.Type()),
 		}
 	} else if !acceptableBinopType(rhs, op) {
 		return value.Null, value.Error{
 			Value: rhs,
-			Inner: fmt.Errorf("should be one of %s for binop %s, got %s", binopAllowedTypes[op], op, rhs.Type()),
+			Inner: fmt.Errorf("should be one of %v for binop %s, got %s", binopAllowedTypes[op], op, rhs.Type()),
 		}
 	}
 


rfratto (approved):
Nice catch! 

I'm not sure what I was originally thinking the user should see here. Long term, I don't know if `[number string]` is very readable to non-programming users, but this is a good improvement for now. :shipit:

PR #2116, opened 2022-09-05
Add field to Operator to disable report usage

### LINKED ISSUE #1856: Allow disabling report usage when using the Operator
Currently, we can't use the Agent's [command-line flags](https://grafana.com/docs/agent/latest/configuration/flags/) when deploying it via the Operator. This means that there's no easy way to disable the usage report to stats.grafana.org (normally done via the `-disable-reporting` flag).

We could provide a way for users to do that; I think a new flag on the agent-operator binary could be a decent option but I'm not sure if we have any other alternatives in mind.

marctc:
#### PR Description

This PR adds the ability to disable the reporting of usage of feature flags
to grafana when using the Agent Operator

Closes #1856 

#### Notes to the Reviewers

After running `make crd` it changed quite a lot the current CRDs.
Please, look if anything critical changed.

#### PR Checklist

- [ ] CHANGELOG updated
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/operator/custom-resource-quickstart.md
+++ b/docs/sources/operator/custom-resource-quickstart.md
@@ -135,7 +135,7 @@ Deploying a GrafanaAgent resource on its own will not spin up any Agent Pods. Ag
 
 ### Disable feature flags reporting
 
-If you would like to disable the [reporting]({{< relref "../configuration/flags.md/#report-use-of-feature-flags" >}}) reporting usage of feature flags to Grafana, set `disableReporting` field to `true`.
+If you would like to disable the [reporting]({{< relref "../configuration/flags.md/#report-use-of-feature-flags" >}}) usage of feature flags to Grafana, set `disableReporting` field to `true`.
 
 ## Step 2: Deploy a MetricsInstance resource
 
--- a/docs/sources/operator/custom-resource-quickstart.md
+++ b/docs/sources/operator/custom-resource-quickstart.md
@@ -133,6 +133,10 @@ The full hierarchy of custom resources is as follows:
 
 Deploying a GrafanaAgent resource on its own will not spin up any Agent Pods. Agent Operator will create Agent Pods once MetricsInstance and LogsIntance resources have been created. In the next step, you'll roll out a `MetricsInstance` resource to scrape cAdvisor and Kubelet metrics and ship these to your Prometheus-compatible metrics endpoint.
 
+### Disable feature flags reporting
+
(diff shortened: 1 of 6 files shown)


docs/sources/operator/custom-resource-quickstart.md
```diff
@@ -133,6 +133,10 @@ The full hierarchy of custom resources is as follows:
 
 Deploying a GrafanaAgent resource on its own will not spin up any Agent Pods. Agent Operator will create Agent Pods once MetricsInstance and LogsIntance resources have been created. In the next step, you'll roll out a `MetricsInstance` resource to scrape cAdvisor and Kubelet metrics and ship these to your Prometheus-compatible metrics endpoint.
 
+### Disable feature flags reporting
+
+If you would like to disable the [reporting]({{< relref "../configuration/flags.md/#report-use-of-feature-flags" >}}) reporting usage of feature flags to Grafana, set `disableReporting` field to `true`.
```
tpaschalis, line 138: Double 'reporting' here.

Otherwise LGTM, but would like someone with more insight to verify the many changes in the CRDs.

pkg/operator/apis/monitoring/v1alpha1/types.go
```diff
@@ -169,6 +169,9 @@ type GrafanaAgentSpec struct {
 	// config port 8080 on the agent.
 	// +kubebuilder:default=false
 	EnableConfigReadAPI bool `json:"enableConfigReadAPI,omitempty"`
+
+	// disableReporting disable reporting of enabled feature flags to Grafana.
+	DisableReporting bool `json:"disableReporting,omitempty"`
```
tpaschalis, line 174: Just curious, but what is the `+kubebuilder:default` directive that exists on some fields (eg. on EnableConfigReadAPI) above? Is it something autogenerated?

marctc: Good that you pointed out! This set values when running `make crd`. I will also add the directive here to default to false

rfratto (approved):
LGTM! `make crds` probably hasn't been run in a while since it's only common to rerun it when actually modifying our CRDs downstream, but any change to doc comments from any type we reference (from Kubernetes or Prometheus Operator) will cause changes like these. 

Since the only other changes to the CRDs were for descriptions, this looks perfectly safe to merge.

rfratto (commented):
Can you make sure to update the CHANGELOG before merging, please?

PR #2117, opened 2022-09-05
flow: polish prometheus.scrape component

tpaschalis:

#### PR Description
This PR aims to provide a round of polish to the `prometheus.scrape` component.

More specifically: 
* Removes the requirement to pass a `job_name`, which was a leftover from when the component could configure multiple targets at the same time. The reported job_name will be set to the component's ID (prometheus.scrape.label).
* Removes the `scrape_config` block to allow for a flatter configuration. The only nested blocks now is `http_client_config`. Note: I tried to simply embed the previously-defined Config field, but unfortunately it broke River's unmarshalling, and ended up not recognizing even the previously-present fields, so I had to widen the scope of changes.
* Made a final comparison between the fields defined in Prometheus' scrape config and the component's Arguments. The only field we're missing is the [custom HTTPClientOptions](https://github.com/prometheus/common/blob/dfbc25bd00225c70aca0d94c3c4bb7744f28ace0/config/http_config.go#L379) which are defined as functional options. We _could_ add some custom logic to hook them up if the user requires to tinker with these settings, but they are not currently configurable not even in Prometheus' YAML configuration, so I'm not sure if it makes sense for now.
* Changed the example configuration to match the new Argument structure.

#### Notes to the Reviewer
If you can, take 5' to verify; have I missed anything else? :)

We're getting close, it'd be nice to have the component as polished as possible! 🙏 
 
#### PR Checklist

- [ ] CHANGELOG updated (N/A)
- [X] Documentation added
- [X] Tests updated

### DIFF
--- a/docs/flow/reference/components/prometheus.scrape.md
+++ b/docs/flow/reference/components/prometheus.scrape.md
@@ -29,7 +29,7 @@ prometheus.scrape "blackbox_scraper" {
 	forward_to = [prometheus.remote_write.grafanacloud.receiver, prometheus.remote_write.onprem.receiver]
 	
 	scrape_interval = "10s"
-	params          = { "target" = ["grafana.com"], "module" = ["http_2xx"]}
+	params          = { "target" = ["grafana.com"], "module" = ["http_2xx"] }
 	metrics_path    = "/probe"
 }
 ```
--- a/docs/flow/reference/components/prometheus.scrape.md
+++ b/docs/flow/reference/components/prometheus.scrape.md
@@ -18,26 +18,32 @@ different labels.
 The following example will set up the job with certain attributes (scrape
 intervals, query parameters) and let it scrape two instances of the blackbox
 exporter. The received metrics will be sent over to the provided list of
-remote_writes, as defined by other components.
+remote writes, as defined by other components.
 
 ```river
 prometheus.scrape "blackbox_scraper" {
-  targets = [
-    {"__address__" = "blackbox-exporter:9115", "instance" = "one"},
-    {"__address__" = "blackbox-exporter:9116", "instance" = "two"},
-  ]
-  forward_to = [prometheus.remote_write.grafanacloud.receiver, prometheus.remote_write.onprem.receiver]
-
-  scrape_config {
-    job_name        = "grafana"
-    scrape_interval = "10s"
-    params          = { "target" = ["grafana.com"], "module" = ["http_2xx"]}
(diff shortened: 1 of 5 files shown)


docs/flow/reference/components/prometheus.scrape.md
```diff
@@ -18,56 +18,40 @@ different labels.
 The following example will set up the job with certain attributes (scrape
 intervals, query parameters) and let it scrape two instances of the blackbox
 exporter. The received metrics will be sent over to the provided list of
-remote_writes, as defined by other components.
+remote writes, as defined by other components.
```
rfratto, line 21: nit: receivers instead of "remote writes?" Elements in forward_to aren't always prometheus.remote_write, sometimes it's prometheus.relabel, and the list of things it could be will probably grow over time.

rfratto (approved):
LGTM, thanks!

docs/flow/reference/components/prometheus.scrape.md
```diff
@@ -18,56 +18,40 @@ different labels.
...
 ## Arguments
 
+The component will configure and start a new scrape job to scrape all of the
+input targets. The list of arguments that can be used to configure the block is
+presented below.
+
+Any omitted fields will take on their default values. In case that conflicting
+attributes are being passed (eg. defining both a BearerToken and
+BearerTokenFile or configuring both Basic Authorization and OAuth2 at the same
+time), the component will report an error.
```
mattdurham, line 46: Do we define the behavior here? I assume error means that it will not work, but what if its an update? Does it go with last good configuration (like local.file) or does it fault entirely?

mattdurham: Not sure; probably legit but maybe it should be referenced as part of the lifecycle pipeline. LGTM

mattdurham:
Holup, thoughts on having job_name as an optional setting.

tpaschalis:
> Holup, thoughts on having job_name as an optional setting.

@mattdurham yeah, we could go about it. My initial thought was that standardizing around the component names+labels could be helpful in cross-referencing other component metrics, but we should let the users override if they prefer to do so.

I'll ping you for a re-review once I've pushed the commit

tpaschalis:
I've brought back the job_name as an optional argument in 5da4d5b9b2452c1a57e0c6ecd815d33a0d54a21f. Let me know how it looks!

mattdurham (approved):
LGTM

PR #2121, opened 2022-09-07
Support arm64 variant of grafana/agent-build-image

rfratto:
Updates Drone to build arm64 variants of grafana/agent-build-image, primarily for Apple Silicon users. 

Also ensures that `libbpfcc-dev` is always installed for amd64, which is the only architecture that eBPF works on right now. 

This will tag 0.15.1 of the build image.

### DIFF
--- a/build-image/Dockerfile
+++ b/build-image/Dockerfile
@@ -38,12 +38,12 @@ FROM rfratto/viceroy:v0.2.1
 
 # NOTE(rfratto): musl is installed so the Docker binaries from alpine work
 # properly.
-RUN apt-get update                               \
- && apt-get install -qy                          \
-      build-essential file zip unzip gettext git \
-      musl libsystemd-dev nsis libbpfcc-dev nsis \
-      rpm ruby ruby-dev rubygems                 \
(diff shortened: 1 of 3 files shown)


build-image/Dockerfile
```diff
@@ -38,12 +38,12 @@ FROM rfratto/viceroy:v0.2.1
 
 # NOTE(rfratto): musl is installed so the Docker binaries from alpine work
 # properly.
-RUN apt-get update                               \
- && apt-get install -qy                          \
-      build-essential file zip unzip gettext git \
-      musl libsystemd-dev nsis libbpfcc-dev nsis \
```
rfratto, line 44: (nsis was accidentally included twice)

captncraig (approved):
You will need to re-sign the drone yaml. Other than that, lgtm.

rfratto:
> What does the second part of libbpfcc-dev:amd64 do? It's the same time I'm seeing this, does it only install it for that architecture?

It installs the amd64 package regardless of the host architecture, making it accessible from cross compilers

PR #2122, opened 2022-09-07
Drone: Use new Makefile and build image

rfratto:
Updates the Drone config to use the new Makefile and build images.

Jobs moved:

- [x] Test 
- [x] Lint 
- [x] Containerize
- [x] Release

### DIFF
--- a/.drone/drone.yml
+++ b/.drone/drone.yml
@@ -190,7 +190,9 @@ steps:
         # stable release (e.g., not a release candiate, we'll update :latest,
         # otherwise we update the branch name).
         #
-        # If there's no branch, we'll default to the version.
+        # If there's no branch, we'll default to the version. This means the
+        # same Docker tag will be created twice, effectively turning the branch
+        # tag into a no-op.
         if [ ! -z "$DRONE_TAG" ] && [[ "$DRONE_TAG" != *"-rc."* ]]; then
           export BRANCH_TAG=$DRONE_TAG
         elif [ ! -z "$DRONE_BRANCH" ]; then
@@ -426,6 +428,6 @@ get:
 
 ---
 kind: signature
-hmac: e1796875c78e260d9a6ae096f6f1ac47f9b09ff687e5350442fdd11fa2c2f020
+hmac: 802581e2c69988f181ee3cedcdf0e936fc153f4c4587abbb9192292604cda123
 
 ...
--- a/.drone/drone.yml
+++ b/.drone/drone.yml
@@ -110,6 +110,11 @@ steps:
       - name: docker
         path: /var/run/docker.sock
     commands:
+      # Ensure the binaries can build and also run Go tests.
+      #
+      # The operator tests require K8S_USE_DOCKER_NETWORK=1 to be set when
+      # tests are being run inside of a Docker container so it can access the
+      # created k3d cluster properly.
       - make -f Makefile-v2.mk binaries
       - K8S_USE_DOCKER_NETWORK=1 make -f Makefile-v2.mk test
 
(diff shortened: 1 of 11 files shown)


.drone/drone.yml
```diff
@@ -177,20 +170,85 @@ steps:
...
+
+      - >-
+        docker buildx build --push
+        --platform linux/amd64,linux/arm64,linux/arm/v7,linux/ppc64le
+        --build-arg RELEASE_BUILD=1
+        --build-arg VERISON=$VERSION
+        -t grafana/agent:$VERSION
+        -t grafana/agent:$BRANCH_TAG
+        -f cmd/agent/Dockerfile.new
+        .
```
rfratto, line 210: This is a bit wordy, but it doesn't feel appropriate to go in the Makefile since:

1. You _have_ to push the images when doing a multiplatform build, which is inappropriate for local development
2. There's a lot of Drone-specific logic around determining what the tags should be.

This is probably something we can clean up a bit by using Jsonnet instead of YAML for Drone, but that's a different PR for a different day.

captncraig: I'd be ok with a local script for this if is is kinda ugly in the drone (which it is). Might be easier to test locally in that case, even if you have to contrive some vars.

.drone/drone.yml
```diff
@@ -106,26 +105,19 @@ trigger:
...
-      - rm -r /usr/local/go
-      - mkdir -p /usr/local/go/bin
-      - wget -q https://golang.org/dl/go1.18.linux-amd64.tar.gz
-      - tar -C /usr/local -xzf go1.18.linux-amd64.tar.gz
-      - rm go1.18.linux-amd64.tar.gz
-      - export PATH=$PATH:/usr/local/go/bin
-      - make DOCKER_OPTS="" cmd/agent/agent cmd/agentctl/agentctl cmd/agent-operator/agent-operator tools/crow/grafana-agent-crow tools/smoke/grafana-agent-smoke
-      - make DOCKER_OPTS="" K8S_USE_DOCKER_NETWORK=1 DRONE=true BUILD_IN_CONTAINER=false test
+      - make -f Makefile-v2.mk binaries
+      - K8S_USE_DOCKER_NETWORK=1 make -f Makefile-v2.mk test
```
captncraig, line 114: can we have a comment why this env is needed? It seems kinda out of thin air now that the command is simpler.

.drone/drone.yml
```diff
@@ -292,22 +354,12 @@ steps:
...
-      - wget -q https://golang.org/dl/go1.18.linux-amd64.tar.gz
-      - tar -C /usr/local -xzf go1.18.linux-amd64.tar.gz
-      - rm go1.18.linux-amd64.tar.gz
-      - export PATH=$PATH:/usr/local/go/bin
-      - go install github.com/mitchellh/gox@latest
-      - go install github.com/tcnksm/ghr@latest
-      - export PATH="$(go env GOPATH)/bin:$PATH"
-      - make -j4 DOCKER_OPTS="" BUILD_IN_CONTAINER=false RELEASE_BUILD=true RELEASE_TAG=${DRONE_TAG} publish
+      - make -f Makefile-v2.mk -j4 RELEASE_BUILD=1 VERISON=${DRONE_TAG} dist
+      - for i in dist/agent*; do zip -j -m $$i.zip $$i; done
```
captncraig, line 359: what is the decision for what parts of this go in the makefile, and which go in drone? 

Some of this seems like it could be hidden in makefile or a script. But I would believe theres reasons for it this way too.

captncraig (approved):

.drone/drone.yml
rfratto: 👍 I'll create a new script for building cross-platform images.

rlankfo (approved):

captncraig (approved):
Lgtm

PR #2123, opened 2022-09-08
Separate Github Actions to avoid publishing rc documentation as stable

### LINKED ISSUE #1782: Release candidates may cause new versioned docs to be created prematurely
From what I can tell of our [publish_docs action](https://github.com/grafana/agent/blob/main/.github/workflows/publish.yml) and the [docs-target action implementation](https://github.com/grafana/grafana-github-actions/tree/main/docs-target), any commit to one of our `release-*` branches will cause versioned docs to be created. 

If my understanding is correct, this means if we create a new release candidate in a `release-*` branch, we'll prematurely create versioned docs for it which will be set to the latest version. 

It seems like the preferred behavior is that we only create new versioned docs after the initial tag is made, though maybe the behavior I described is fine as-is. 

cc @jdbaldry @tpaschalis

tpaschalis:

#### PR Description
This PR separates the `publish.yml` GitHub action into two
a) A new, simpler one which just pushes docs whenever there's a new commit on `main`.
b) An updated version of the previous one, which uses the `has-matching-release-tag` custom action to perform a stricter check and only push documentation if there's already a published tag with the same name.

The same approach is used by Mimir's [workflows](https://github.com/grafana/mimir/tree/main/.github/workflows) and while I think we could maybe fit everything into one Action, I believe it's worth it to follow a common approach, with simpler building blocks that other projects can reuse more easily.

Fixes #1782

#### PR Checklist

- [ ] CHANGELOG updated (N/A)
- [ ] Tests updated (N/A)

### DIFF
--- a/.github/workflows/publish-documentation-next.yml
+++ b/.github/workflows/publish-documentation-next.yml
@@ -1,61 +1,45 @@
-name: publish_docs
+name: "publish-technical-documentation-next"
 
 on:
   push:
     branches:
       - main
-      - 'release-*'
-    tags:
-      - 'v[0-9]+.[0-9]+.[0-9]+'
     paths:
       - 'docs/sources/**'
+  workflow_dispatch:
 
 jobs:
   test:
     runs-on: ubuntu-latest
     steps:
-      - name: Check out code
+      - name: "Check out code"
         uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


rfratto (approved):
👏 LGTM!

PR #2128, opened 2022-09-08
prometheus.scrape: ensure that target groups match configured job name

rfratto:
The target groups must match the job name of the scrape_config, otherwise the scrape pool will fail to recognize the targets.

I didn't add tests to check for this so we could fix main quickly, but we should probably have tests in the future to assert this condition.


rlankfo (approved):

rfratto:
It looks like this wasn't a problem because because the configured `job_name` was accidentally getting ignored, so that it was always using the component ID.

PR #2130, opened 2022-09-09
Remove old makefile, dockerfiles

### LINKED ISSUE #2052: Proposal: clean up Makefile & CI

rfratto:
Removes the old Makefile and Dockerfiles in favor of the new Makefile and Dockerfiles introduced in #2084. 

Nothing is changed except for renaming the files and renaming references to the old names.

Closes #2052.


rfratto:
I need to add test-packages back into the new Makefile since I forgot to do that. Working on that now.

rlankfo (approved):

PR #2133, opened 2022-09-12
flow: add reference page for prometheus.relabel component

tpaschalis:

#### PR Description
Add a reference page for the prometheus.relabel component, much like discovery.relabel.
No issue filed.

#### PR Checklist

- [ ] CHANGELOG updated (N/A)
- [x] Documentation added
- [ ] Tests updated (N/A)

### DIFF
--- a/docs/flow/reference/components/prometheus.relabel.md
+++ b/docs/flow/reference/components/prometheus.relabel.md
@@ -64,7 +64,7 @@ Name | Type | Description | Default | Required
 `replacement`   | `string`       | The value against which a regex replace is performed, if the regex matches the extracted value. Supports previously captured groups. | $1 | no
 `action`        | `string`       | The relabeling action to perform. | replace | no
 
-Here's a list of the available actions along with a brief description of their usage.
+Here's a list of the available actions, along with a brief description of their usage.
 
 * `replace`   - Matches `regex` to the concatenated labels. If there's a match, it replaces the content of the `target_label` using the contents of the `replacement` field.
 * `keep`      - Keeps metrics where `regex` matches the string extracted using the `source_labels` and `separator`.
--- a/docs/flow/reference/components/prometheus.relabel.md
+++ b/docs/flow/reference/components/prometheus.relabel.md
@@ -91,23 +91,17 @@ metric_a{host = "cluster_a/production",  __address__ = "cluster_a", instance = "
 ```
 
 The third and final relabeling step which uses the `labeldrop` action would
-remove the `instance` label from the set of labels
+remove the `instance` label from the set of labels.
+
+So, after filtering down the samples passed through the exported receiver,
+the two resulting metrics would then be propagated to each receiver passed in
+the `forward_to` argument.
 
 ```
 metric_a{host = "localhost/development", __address__ = "localhost", app = "backend"}	2
 metric_a{host = "cluster_a/production",  __address__ = "cluster_a", app = "backend"}	9
 ```
 
-Finally, the `__address__` label, as it starts with a double underscore and is
-reserved as an 'internal' label and would be implicitly removed.
-
-After filtering down the samples passed through the exported receiver, these
-two final metrics would then be propagated to each receiver passed in the
-`forward_to` argument.
-```
-metric_a{host = "localhost/development", app = "backend"}	2
-metric_a{host = "cluster_a/production",  app = "backend"}	9
-```
 
 ## Arguments
 
@@ -115,7 +109,7 @@ The following arguments are supported:
 
 Name | Type | Description | Default | Required
 ---- | ---- | ----------- | ------- | --------
-forward_to | list(receiver) | Where the metrics should be forwarded to after relabeling takes place | | **yes**
+`forward_to` | `list(receiver)` | Where the metrics should be forwarded to after relabeling takes place | | **yes**
 
 The following subblocks are supported:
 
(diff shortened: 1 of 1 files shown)


docs/flow/reference/components/prometheus.relabel.md
```diff
@@ -0,0 +1,181 @@
...
+Multiple `prometheus.relabel` components can be specified by giving them
+different labels.
+
+## Example
+
+```river
+prometheus.relabel "keep_mysql_only" {
+  forward_to [prometheus.remote_write.onprem.receiver]
+
+  metric_relabel_config {
```
rfratto, line 41: Seeing this in action, `metric_relabel_config` feels a little awkward. I'm not sure what a better name for this would be. Maybe `rule`? 

Maybe it's not something we need to change for v0.28, but might be worth keeping in the back of our heads.

docs/flow/reference/components/prometheus.relabel.md
```diff
@@ -0,0 +1,181 @@
...
+
+The third and final relabeling step which uses the `labeldrop` action would
+remove the `instance` label from the set of labels
+
+```
+metric_a{host = "localhost/development", __address__ = "localhost", app = "backend"}	2
+metric_a{host = "cluster_a/production",  __address__ = "cluster_a", app = "backend"}	9
+```
+
+Finally, the `__address__` label, as it starts with a double underscore and is
```
rfratto, line 101: This isn't quite right.

Reserved labels are removed from _targets_ prior to scraping, but never from metrics.

docs/flow/reference/components/prometheus.relabel.md
```diff
@@ -0,0 +1,181 @@
...
+metric_a{host = "cluster_a/production",  app = "backend"}	9
+```
+
+## Arguments
+
+The following arguments are supported:
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+forward_to | list(receiver) | Where the metrics should be forwarded to after relabeling takes place | | **yes**
```
rfratto, line 118: Nit: I'd like for us to wrap names and types in backticks so forward_to appears as `forward_to`. 

Ditto for elsewhere in this file and when we describe the list of supported relabel actions.

rfratto (commented):
This is a lovely amount of detail, nice job!

rfratto:
> This is a lovely amount of detail, nice job!

To add a little extra flavor: I think this is what we should be doing with all of our component documentation: we should be explaining as much as possible so users understand exactly what components do. Implementation details don't need to be documented, but anything a user would need to be aware of definitely should. 

Nice work 🎉

docs/flow/reference/components/prometheus.relabel.md
tpaschalis: We're talking about relabeling _rules_ and relabeling _steps_. I think that `rule` sounds reasonable, and maybe we should coordinate this with the discovery.relabel component, too.

We still have enough time, I believe it makes sense to change this before launch but I'd prefer to make bunch these changes in a different PR.

rfratto: > We still have enough time, I believe it makes sense to change this before launch but I'd prefer to make bunch these changes in a different PR.

👍

mattdurham: I prefer `rule` over `metric_relabel_config` also but fine with shipping this without it.

docs/flow/reference/components/prometheus.relabel.md
```diff
@@ -0,0 +1,172 @@
...
+aliases:
+- /docs/agent/latest/flow/reference/components/prometheus.relabel
+title: prometheus.relabel
+---
+
+# prometheus.relabel
+
+Prometheus metrics follow the [OpenMetrics](https://openmetrics.io/) format.
+Each time series is uniquely identified by its metric name, plus optional
+key-value pairs called labels. Each sample, i.e. a datapoint in the time
```
rfratto, line 11: Apparently it's recommended to avoid e.g. and i.e. because people tend to mix them up:

https://developers.google.com/style/abbreviations?hl=en#dont-use

mattdurham (approved):
Looks great!

docs/flow/reference/components/prometheus.relabel.md
```diff
@@ -0,0 +1,172 @@
...
+transformations will be applied in top-down order.
+
+The following arguments can be used to configure a `metric_relabel_config`
+block.
+All arguments are optional and any omitted fields will take on their default
+values.
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`source_labels` | `list(string)` | The list of labels whose values should be selected. Their content is concatenated using the `separator` and matched against `regex`. | | no
```
karengermond, line 131:
```suggestion
`source_labels` | `list(string)` | The list of labels whose values are to be selected. Their content is concatenated using the `separator` and matched against `regex`. | | no
```

docs/flow/reference/components/prometheus.relabel.md
```diff
@@ -0,0 +1,172 @@
...
+`separator`     | `string`       |  The separator used to concatenate the values present in `source_labels`. | ; | no
+`regex`         | `string`       | A valid RE2 expression with support for parenthesized capture groups. Used to match the extracted value from the combination of the `source_label` and `separator` fields or filter labels during the `labelkeep/labeldrop/labelmap` actions. | `(.*)` | no
+`modulus`       | `uint`         | A positive integer used to calculate the modulus of the hashed source label values. | | no
+`target_label`  | `string`       | Label to which the resulting value will be written to. | | no
+`replacement`   | `string`       | The value against which a regex replace is performed, if the regex matched the extracted value. Supports previously captured groups. | $1 | no
```
karengermond, line 136:
```suggestion
`replacement`   | `string`       | The value against which a regex replace is performed, if the regex matches the extracted value. Supports previously captured groups. | $1 | no
```

docs/flow/reference/components/prometheus.relabel.md
```diff
@@ -0,0 +1,172 @@
...
+`action`        | `string`       | The relabeling action to perform. | replace | no
+
+Here's a list of the available actions along with a brief description of their usage.
```
karengermond, line 139:
```suggestion
Here's a list of the available actions, along with a brief description of their usage.
```

docs/flow/reference/components/prometheus.relabel.md
```diff
@@ -0,0 +1,172 @@
...
+
+* `replace`   - This action matches `regex` to the concatenated labels. If there's a match, it replaces the content of the `target_label` using the contents of the `replacement` field.
+* `keep`      - This action only keeps the metrics where `regex` matches the string extracted using the `source_labels` and `separator`.
```
karengermond, line 142:
```suggestion
* `keep`      - Keeps only the metrics where `regex` matches the string extracted using the `source_labels` and `separator`.
```

docs/flow/reference/components/prometheus.relabel.md
```diff
@@ -0,0 +1,172 @@
...
+`separator`     | `string`       |  The separator used to concatenate the values present in `source_labels`. | ; | no
+`regex`         | `string`       | A valid RE2 expression with support for parenthesized capture groups. Used to match the extracted value from the combination of the `source_label` and `separator` fields or filter labels during the `labelkeep/labeldrop/labelmap` actions. | `(.*)` | no
+`modulus`       | `uint`         | A positive integer used to calculate the modulus of the hashed source label values. | | no
+`target_label`  | `string`       | Label to which the resulting value will be written to. | | no
+`replacement`   | `string`       | The value against which a regex replace is performed, if the regex matched the extracted value. Supports previously captured groups. | $1 | no
+`action`        | `string`       | The relabeling action to perform. | replace | no
+
+Here's a list of the available actions along with a brief description of their usage.
+
+* `replace`   - This action matches `regex` to the concatenated labels. If there's a match, it replaces the content of the `target_label` using the contents of the `replacement` field.
```
karengermond, line 141:
```suggestion
* `replace`   - Matches `regex` to the concatenated labels. If there's a match, it replaces the content of the `target_label` using the contents of the `replacement` field.
```

docs/flow/reference/components/prometheus.relabel.md
```diff
@@ -0,0 +1,172 @@
...
+* `keep`      - This action only keeps the metrics where `regex` matches the string extracted using the `source_labels` and `separator`.
+* `drop`      - This action drops the metrics where `regex` matches the string extracted using the `source_labels` and `separator`.
```
karengermond, line 143:
```suggestion
* `drop`      - Drops the metrics where `regex` matches the string extracted using the `source_labels` and `separator`.
```

docs/flow/reference/components/prometheus.relabel.md
```diff
@@ -0,0 +1,172 @@
...
+* `hashmod`   - This action hashes the concatenated labels, calculates its modulo `modulus` and writes the result to the `target_label`.
```
karengermond, line 144:
```suggestion
* `hashmod`   - Hashes the concatenated labels, calculates its modulo `modulus` and writes the result to the `target_label`.
```

docs/flow/reference/components/prometheus.relabel.md
```diff
@@ -0,0 +1,172 @@
...
+* `labelmap`  - This action matches `regex` against all label names. Any labels that match will be renamed according to the contents of the `replacement` field.
```
karengermond, line 145:
```suggestion
* `labelmap`  - Matches `regex` against all label names. Any labels that match are renamed according to the contents of the `replacement` field.
```

docs/flow/reference/components/prometheus.relabel.md
```diff
@@ -0,0 +1,172 @@
...
+* `labeldrop` - This action matches `regex` against all label names. Any labels that match will be removed from the metric's label set.
```
karengermond, line 146:
```suggestion
* `labeldrop` - Matches `regex` against all label names. Any labels that match are removed from the metric's label set.
```

docs/flow/reference/components/prometheus.relabel.md
```diff
@@ -0,0 +1,172 @@
...
+* `labelkeep` - This action matches `regex` against all label names. Any labels that don't match will be removed from the metric's label set.
```
karengermond, line 147:
```suggestion
* `labelkeep` - Matches `regex` against all label names. Any labels that don't match are removed from the metric's label set.
```

docs/flow/reference/components/prometheus.relabel.md
```diff
@@ -0,0 +1,172 @@
...
+Finally, note that the regex capture groups can be referred to using either the
+`$1` or `$${1}` notation.
+
+## Exported fields
+
+The following fields are exported and can be referenced by other components:
+
+Name | Type | Description
+---- | ---- | -----------
+`receiver` | `receiver` | The input receiver where samples should be sent to, for them to be relabeled.
```
karengermond, line 158:
```suggestion
`receiver` | `receiver` | The input receiver where samples are sent to be relabeled.
```

docs/flow/reference/components/prometheus.relabel.md
```diff
@@ -0,0 +1,172 @@
...
+
+## Component health
+
+`prometheus.relabel` will only be reported as unhealthy when given an invalid
```
karengermond, line 162:
```suggestion
`prometheus.relabel` is only reported as unhealthy if given an invalid
```

docs/flow/reference/components/prometheus.relabel.md
```diff
@@ -0,0 +1,172 @@
...
+configuration. In those cases, exported fields will be kept at their last
```
karengermond, line 163:
```suggestion
configuration. In those cases, exported fields are kept at their last
```

karengermond (approved):
LGTM!

rfratto:
Copying [comment](https://github.com/grafana/agent/pull/2145#discussion_r969978045) from #2145 over here: 

> It might be better to move this section to be after the example, potentially even after arguments and exports. Ditto for the intro section in https://github.com/grafana/agent/pull/2133.
>
> Reference: https://developers.google.com/style/api-reference-comments?hl=en
>
>> On each unique page (for a class, interface, etc.), include a code sample (~5-20 lines) at the top.
>
> It feels like we'll want to keep information users will access the most frequently closest to the top (brief description, example, arguments, exports) and less frequently accessed information closer to the bottom of the page (behavior).

docs/flow/reference/components/prometheus.relabel.md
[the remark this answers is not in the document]
tpaschalis: Removed i.e. altogether in an effort to be more descriptive :)

rfratto (approved):
LGTM!

PR #2153, opened 2022-09-14
Flow: finish local.file documentation

rfratto:


rfratto:
I re-read through local.file and didn't see anything else that needed to change; I think all the information a user would need to know is already being covered. 

The only thing that was missing was the new metric added recently.

karengermond (approved):
LGTM!

mattdurham (approved):
LGTM

PR #2154, opened 2022-09-14
prometheus.remote_write: finalize prometheus.remote_write component

rfratto:
The following changes are made to prometheus.remote_write:

* Naming patterns are changed to align with other components. For example, using an Arguments struct to describe the arguments of the component.

* The nested `remote_write` block has been renamed to `endpoint`.

* Missing options are added for the `endpoint` block:
  * `remote_timeout`
  * `headers`
  * `metadata_config`

* `client_config` is renamed to `http_client_config` to align with other components.

* The `basic_auth` block is removed, as that block is already defined in `http_client_config`.

The remainder of the changes are minor code cleanup. A test has been added to assert that prometheus.remote_write works as expected by sending synthetic metrics to a test server.

The following has been purposefully scoped out of this PRs and will be done in later PRs:

* Supporting the sigv4 authentication method in the `endpoint` block.
* Updating documentation to reflect the naming changes.


rfratto:
> The following has been purposefully scoped out of this PRs and will be done in later PRs:
>
> * Supporting the sigv4 authentication method in the endpoint block.

I left this one out because it's kind of tricky. Since we don't currently supporting inlined fields, it feels a bit strange to have `sigv4` sit alongside `http_client_config` as a sibling. I'd like to leave it out for now because usage of sigv4 should be uncommon while Flow is still experimental, and it'll give us time to figure out how to make it fit in more naturally.

component/prometheus/remotewrite/types.go
```diff
@@ -108,84 +182,33 @@ func (o *WALOptions) UnmarshalRiver(f func(interface{}) error) error {
...
-		rwConfigs = append(rwConfigs, rwc)
+		rwConfigs = append(rwConfigs, &config.RemoteWriteConfig{
+			URL:                 &common.URL{URL: parsedURL},
+			RemoteTimeout:       model.Duration(rw.RemoteTimeout),
+			Headers:             rw.Headers,
+			WriteRelabelConfigs: nil, // WriteRelabelConfigs are currently not supported
+			Name:                rw.Name,
+			SendExemplars:       rw.SendExemplars,
+
+			HTTPClientConfig: *rw.HTTPClientConfig.Convert(),
```
tpaschalis, line 207: WDYT about reusing [this](https://github.com/grafana/agent/blob/917b8f3eaedcef34fa8316a416ae3fc4c07286d2/component/prometheus/scrape/types.go#L52) validation method before applying the HTTP client config here? 

We use it in prometheus.scrape where we fail the component creation if conflicting configuration is passed eg. defining both basic_auth _and_ oauth2. I picked it up from [prometheus/common](https://github.com/prometheus/common/blob/dfbc25bd00225c70aca0d94c3c4bb7744f28ace0/config/http_config.go#L278-L338) where it's called in their UnmarshalYAML/UnmarshalJSON methods.


The validation could even go in the struct's Convert method, (although it's a little awkward) or the UnmarshalRiver itself.

component/prometheus/remotewrite/remote_write_test.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/prompb"
+	"github.com/prometheus/prometheus/storage/remote"
+	"github.com/stretchr/testify/require"
+)
+
+// Test is an integration-level test which ensures that metrics can get sent to
+// a prometheus.remote_write component and forwarded to a
+// remote_write-compatible server.
+func Test(t *testing.T) {
```
tpaschalis, line 25: Nice one! 🙌

tpaschalis (approved):
LGTM, I don't have anything to add other than that suggestion!

component/prometheus/remotewrite/types.go
```diff
@@ -41,45 +51,109 @@ var (
...
+type QueueOptions struct {
 	Capacity          int           `river:"capacity,attr,optional"`
 	MaxShards         int           `river:"max_shards,attr,optional"`
 	MinShards         int           `river:"min_shards,attr,optional"`
 	MaxSamplesPerSend int           `river:"max_samples_per_send,attr,optional"`
 	BatchSendDeadline time.Duration `river:"batch_send_deadline,attr,optional"`
 	MinBackoff        time.Duration `river:"min_backoff,attr,optional"`
 	MaxBackoff        time.Duration `river:"max_backoff,attr,optional"`
-	RetryOn429        bool          `river:"retry_on_http_429,attr,optional"`
+	RetryOnRateLimit  bool          `river:"retry_on_http_429,attr,optional"`
```
mattdurham, line 103: Should we keep the 429 variable name so it aligns more closely? Or my preference would be to align on `retry_on_rate_limit`.

mattdurham (approved):
LGTM, minor naming suggestion

component/prometheus/remotewrite/types.go
rfratto: I'll keep the variable name 👍

rfratto: Changed to RetryOnHTTP429 so the name is more clear.

component/prometheus/remotewrite/types.go
rfratto: Good idea. I made Validate a method on the struct and called it during unmarshal.

rfratto: @tpaschalis can you take another look to see if what I did looks ok to you?

tpaschalis: Yeah, works fine for both components during initial load and reloaded configuration. Thanks!

tpaschalis (approved):
⛵

PR #2156, opened 2022-09-15
river: improve diagnostic for invalid object keys

tpaschalis:

#### PR Description
After a recent discussion with @rfratto about River only allowing string map keys, we noticed that the error was not terminating correctly.

This PR contains an approach for fixing the reported error by advancing directly to the assignment operator instead of attempting to simply consume any strings or identifiers in sequence.

No issue filed.

#### Notes to the Reviewer
A side-effect of this change is that `parseField()` no longer returns a nil `ast.ObjectField`, but rather an empty one in case of errors. I didn't see any issues with that in any caller, but I'll look once more.

#### PR Checklist

- [ ] CHANGELOG updated (N/A)
- [X] Tests updated

### DIFF
--- a/pkg/river/parser/internal.go
+++ b/pkg/river/parser/internal.go
@@ -683,29 +683,28 @@ func (p *parser) parseFieldList(until token.Token) []*ast.ObjectField {
 //
 //    Field = ( string | identifier ) "=" Expression
 func (p *parser) parseField() *ast.ObjectField {
-	if p.tok != token.STRING && p.tok != token.IDENT {
-		p.addErrorf("expected field name (string or identifier), got %s", p.tok)
-		p.advanceAny(fieldStarter)
-		return nil
-	}
+	var field ast.ObjectField
 
-	field := &ast.ObjectField{
-		Name: &ast.Ident{
+	if p.tok == token.STRING || p.tok == token.IDENT {
+		field.Name = &ast.Ident{
 			Name:    p.lit,
 			NamePos: p.pos,
-		},
+		}
+		if p.tok == token.STRING && len(p.lit) > 2 {
+			// The field name a string literal; unwrap the quotes.
+			field.Name.Name = p.lit[1 : len(p.lit)-1]
+			field.Quoted = true
+		}
+		p.next() // Consume field name
+	} else {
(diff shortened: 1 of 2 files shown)


pkg/river/parser/internal.go
```diff
@@ -683,34 +683,28 @@ func (p *parser) parseFieldList(until token.Token) []*ast.ObjectField {
...
 	}
-	p.next() // Consume field name
 
 	p.expect(token.ASSIGN)
 
 	field.Value = p.ParseExpression()
-	return field
-}
-
-var fieldStarter = map[token.Token]struct{}{
```
tpaschalis, line 711: We no longer have a use for this map, so let's remove it. 

We _could_ use it to check that p.tok is in the `fieldStarter` set instead of checking directly for token.STRING and token.IDENT, but I'm not sure it makes any difference.

pkg/river/parser/testdata/invalid_object_key.river
```diff
@@ -0,0 +1,8 @@
+obj {
+  map = {
+    "string_field"    = "foo", 
+    identifier_string = "bar", 
+    1337 /* ERROR "expected field name \(string or identifier\), got NUMBER" */ = "baz",
```
rfratto, line 5: Can you add another valid field after this to make sure it doesn’t report any more errors?

pkg/river/parser/internal.go
```diff
@@ -683,34 +683,28 @@ func (p *parser) parseFieldList(until token.Token) []*ast.ObjectField {
...
 			NamePos: p.pos,
-		},
-	}
-	if p.tok == token.STRING && len(p.lit) > 2 {
-		// The field name a string literal; unwrap the quotes.
-		field.Name.Name = p.lit[1 : len(p.lit)-1]
-		field.Quoted = true
+		}
+		if p.tok == token.STRING && len(p.lit) > 2 {
+			// The field name a string literal; unwrap the quotes.
```
rfratto, line 694: Found a typo in the comment that got copied over :)
```suggestion
			// The field name is a string literal; unwrap the quotes.
```

rfratto (approved):
LGTM!

PR #2157, opened 2022-09-15
Upgrade Loki dependency to v2.6.1

rfratto:


pkg/config/config.go
```diff
@@ -430,12 +419,19 @@ func load(fs *flag.FlagSet, args []string, loader loaderFunc) (*Config, error) {
...
-	err := LoadBytes([]byte(rawCfg), false, cfg)
+	var cfg Config
+	err := LoadBytes([]byte(rawCfg), false, &cfg)
 	require.NoError(t, err)
-	bb, err := yaml.Marshal(cfg)
+
+	// Set integrations version to make sure our marshal function goes through
+	// the custom marshaling code.
+	err = cfg.Integrations.setVersion(integrationsVersion1)
 	require.NoError(t, err)
```
rfratto, line 429: I'm genuinely not sure why this worked before. The version needs to be set for remarshaling to work properly.

pkg/config/config.go
```diff
@@ -430,12 +419,19 @@ func load(fs *flag.FlagSet, args []string, loader loaderFunc) (*Config, error) {
...
-	scrubbedCfg := string(bb)
-	require.True(t, strings.Contains(scrubbedCfg, "<secret>"))
-	require.False(t, strings.Contains(scrubbedCfg, originalValue))
+
+	bb, err := yaml.Marshal(&cfg)
+	require.NoError(t, err)
+
+	fmt.Println(string(bb))
```
rfratto, line 434:
```suggestion
```

pkg/config/integrations.go
```diff
@@ -62,7 +62,7 @@ func (c VersionedIntegrations) MarshalYAML() (interface{}, error) {
 	case c.configV2 != nil:
 		return c.configV2, nil
 	default:
-		return c.raw, nil
+		return &c.raw, nil
```
rfratto, line 65:
```suggestion
		// A pointer is needed for the yaml.Marshaler implementation to work.
		return &c.raw, nil
```

mattdurham (approved):
LGTM

PR #2159, opened 2022-09-15
Add Karen and Eve as maintainers of documentation

rfratto:
Also add Karen as a CODEOWNER for documentation.


karengermond (approved):
Thanks!

PR #2161, opened 2022-09-15
Add metrics for config reloads and config hash

### LINKED ISSUE #2040: Add metrics for config reloads
We need to add the following metrics to help debug issues with config reloads:

`agent_config_last_reload_success_timestamp_seconds`
`agent_config_last_reload_successful`
`agent_config_hash`

The first two exist in Prometheus and the third exists in Alertmanager and the Loki, Tempo, Mimir ecosystem.

jcreixell:
#### PR Description

- Add metrics for config reloads
  - Adds gauges to help debug issues with config reloads
- Add metric reporting config hash
  - Adds a gauge reporting a sha256 hash of the config (before
    expansion) via label. This should be useful to debug config reload
    issues.
  - The gauge is implemented for agent, agentctl and agent_flow and
    supports static, dynamic and remote config.
 
Fixes #2040 

#### Notes to the Reviewer

- I think I instrumented the code in the right places but might have missed something, please double check
- ~I had to namespace the metric name for flow as I would get a double registration otherwise. Amy ideas on how to do this better without too much complexity?~
- Is there a risk of cardinality explosion for the hash metric? (for example with dynamic configs)
- I assume we don't test internal metrics
- I assume we don't document these metrics, please correct if I am wrong
- Anything else I missed?

#### PR Checklist

- [x]  CHANGELOG updated


cmd/agent/entrypoint.go
```diff
@@ -61,6 +65,14 @@ func NewEntrypoint(logger *server.Logger, cfg *config.Config, reloader Reloader)
 		ep = &Entrypoint{
 			log:      logger,
 			reloader: reloader,
+			configReloadSuccess: promauto.NewGauge(prometheus.GaugeOpts{
+				Name: "agent_config_last_reload_successful",
+				Help: "Config loaded successfully.",
+			}),
+			configReloadSeconds: promauto.NewGauge(prometheus.GaugeOpts{
```
mattdurham, line 72: Check with Goutham but I wonder if its better to have success/failure be a label so we can track the two separately, it does break some standing with prometheus but might be acceptable.  So would become `agent_config_last_reload_timestamp_seconds` with label like `result=success` or `result=failure`.

cmd/agent/entrypoint.go
```diff
@@ -250,17 +262,21 @@ func (ep *Entrypoint) TriggerReload() bool {
 
 	cfg, err := ep.reloader()
 	if err != nil {
+		ep.configReloadSuccess.Set(0)
```
mattdurham, line 265: Also in thinking about this I wonder if we need to change it to just `load`, or set values to good on startup. If I am alerting based on these they will not be valid til a reload happens.

pkg/config/config.go
```diff
@@ -434,3 +452,11 @@ func CheckSecret(t *testing.T, rawCfg string, originalValue string) {
 	require.True(t, strings.Contains(string(bb), "<secret>"))
 	require.False(t, strings.Contains(string(bb), originalValue))
 }
+
+// Create a sha256 hash of the config before expansion and expose it via
+// the agent_config_hash metric
```
mattdurham, line 457: nit: Robert likes comments to end with a .

cmd/agent/entrypoint.go
```diff
@@ -61,6 +65,14 @@ func NewEntrypoint(logger *server.Logger, cfg *config.Config, reloader Reloader)
 		ep = &Entrypoint{
 			log:      logger,
 			reloader: reloader,
+			configReloadSuccess: promauto.NewGauge(prometheus.GaugeOpts{
```
mattdurham, line 68: Do these show up under the flow system? We should probably have them there to.

mattdurham (changes requested):
Looks solid with some feedback requested.

cmd/agent/entrypoint.go
jcreixell: I thought about, it's probably a good idea, let's do it

cmd/agent/entrypoint.go
gouthamve: Shouldn't the default be success? This is because the "first" load of the config is a success and if its bad config, the startup itself will fail. I would default to success tbh.

jcreixell: uhm, let's move this to config.go so that we keep track of loads and not just reloads

cmd/agent/flow_run.go
```diff
@@ -25,6 +27,24 @@ import (
...
+var (
+	configHash = promauto.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "agent_flow_config_hash",
+			Help: "Hash of the currently active config file.",
+		},
+		[]string{"sha256"},
+	)
+	configReloadSuccess = promauto.NewGauge(prometheus.GaugeOpts{
+		Name: "agent_flow_config_last_reload_successful",
```
jcreixell, line 39: still not super happy about this one, as both flow and non-flow metrics show up in the metrics endpoint, instead of being registered lazily. Ideas welcome

jcreixell:
@mattdurham I have addressed your comments, please have another look when you have time. The only remaining issue I can think of is that both flow and non-flow metrics are exposed, however I am not sure that fixing this is worth the hassle. Thank you!

cmd/agent/flow_run.go
tpaschalis: Do you mean that Flow metrics would show up even in the 'static' Agent mode?

If so, it might be a little annoying but not too much. We _could_ stuff everything into the Run method, but I'm not sure the readability is a good tradeoff.

tpaschalis (approved):
LGTM! Nothing to add personally, simple and straightforward.

cmd/agent/flow_run.go
mattdurham: For Flow side: In `flow_run.go` instead of using promauto you could register it with the DefaultRegister. Same with static agent , instead of promauto register with the reg in `entrypoint.go`

jcreixell:
@mattdurham See my last commit for a refactoring moving the duplicated code into a new package. 

I tried the simple solution of manually registering the metrics during execution, but it was quite intrusive, especially in config.go (since Load can be called multiple times, I would need to register the metrics outside that package and pass them as a parameter). In the end, I opted for using a singleton instance for the metrics, reused both in flow and regular agent.

Location of this new package and naming are up for discussion.

pkg/config/instrumentation/config_metrics.go
```diff
@@ -0,0 +1,59 @@
...
+// Create a sha256 hash of the config before expansion and expose it via
+// the agent_config_hash metric.
+func (c *configMetrics) InstrumentConfig(buf []byte) {
+	hash := sha256.Sum256(buf)
+	c.configHash.Reset()
+	c.configHash.WithLabelValues(fmt.Sprintf("%x", hash)).Set(1)
+}
+
+// Expose metrics for reload success / failures.
+func (c *configMetrics) InstrumentLoad(isError bool) {
```
mattdurham, line 51: The `isError` through me off a bit; I was expecting more success bool.

cmd/agent/flow_run.go
```diff
@@ -108,11 +109,15 @@ func (fr *flowRun) Run(configFile string) error {
 	reload := func() error {
 		flowCfg, err := loadFlowFile(configFile)
 		if err != nil {
+			instrumentation.ConfigMetrics.InstrumentLoad(true)
```
mattdurham, line 112: Could this be simplified to one `defer instrumentation.ConfigMetrics.InstrumentLoad(err != nil)`?

pkg/config/instrumentation/config_metrics.go
jcreixell: sounds good, let's change it

pkg/config/instrumentation/config_metrics.go
```diff
@@ -0,0 +1,59 @@
...
+
+	m.configHash = promauto.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "agent_config_hash",
+			Help: "Hash of the currently active config file.",
+		},
+		[]string{"sha256"},
+	)
+	m.configReloadSuccess = promauto.NewGauge(prometheus.GaugeOpts{
+		Name: "agent_config_last_reload_successful",
```
jcreixell, line 31: question: shall we be using the term `load` instead of `reload` here? Reload could be confusing given that we instrument every load, including the initial one

mattdurham: Sounds reasonable.

mattdurham (approved):
Looks good to me, clean and simple

pkg/config/instrumentation/config_metrics.go
```diff
@@ -0,0 +1,59 @@
...
+		[]string{"sha256"},
+	)
+	m.configLoadSuccess = promauto.NewGauge(prometheus.GaugeOpts{
+		Name: "agent_config_last_load_successful",
+		Help: "Config loaded successfully.",
+	})
+	m.configLoadSeconds = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "agent_config_last_load_timestamp_seconds",
+		Help: "Timestamp of the last configuration load by result.",
+	}, []string{"result"})
```
rfratto, line 37: Hm, one of my concerns with the `result` label is that every instance of the label will continue to be reported for the process lifetime, which might be confusing and hard to query for exactly what you're looking for. 

I think there's two choices here:

1. Remove the result label 
2. Delete all the labels on the vector every time we instrument a new load, so the /metrics endpoint only reports the metric and label corresponding to the most recent load.  

Personally, I'd go for the first; a user can still know when unsuccessful loads happened by querying for `agent_config_last_load_timestamp_seconds and agent_config_last_load_successful == 0`.

pkg/config/instrumentation/config_metrics.go
```diff
@@ -0,0 +1,59 @@
...
+// Create a sha256 hash of the config before expansion and expose it via
+// the agent_config_hash metric.
+func (c *configMetrics) InstrumentConfig(buf []byte) {
+	hash := sha256.Sum256(buf)
+	c.configHash.Reset()
+	c.configHash.WithLabelValues(fmt.Sprintf("%x", hash)).Set(1)
+}
+
+// Expose metrics for load success / failures.
+func (c *configMetrics) InstrumentLoad(success bool) {
```
rfratto, line 51: nit: Passing success to a function like this feels a little weird to me. You could implement this using some middleware-y function:

```go
cm.InstrumentLoad(bb, func(bb []byte) error {
  // do the load 
  // return if there was an error 
})
```

which would set all the appropriate metrics for you. 

Honestly, I'm not sure if that's really worth implementing right now. This is mostly just a nit, it feels a little strange but I wouldn't say it's a blocker.

rfratto (approved):
LGTM. I have some concerns about the result label for `agent_config_last_load_timestamp_seconds`, but how you solve it is ultimately up to you and you don't need a second approval from me. 

Let @grafana/grafana-agent-maintainers know when you're ready for it to be merged and someone can put it in main :)

pkg/config/instrumentation/config_metrics.go
rfratto: cc @mattdurham since I'm suggesting that the `result` label might cause issues

mattdurham: Seems reasonable.

rfratto: SGTM 👍

pkg/config/instrumentation/config_metrics.go
jcreixell: I agree with your point here, however looking at the config code, it is hard to agree on a single API that would work for all current implementations (including dynamic config). I am also purposely trying to separate config hash instrumentation from config load (under the assumption that it is better to build the hash before applying dynamic transformations like variable substitution, etc to prevent noise, while it is better to instrument loads as late as possible, to cover all possible failure modes). I am also open to have these assumptions challenged :)

rfratto: I don't feel too strongly about it, it's something we can always change later and it's not intended to be a public facing API. It's more than fine for now 👍

jcreixell:
@grafana/grafana-agent-maintainers I think I am done here, feel free to merge if you are ok with my last commits and comment

Issue #2095, opened 2022-08-31
flow: panic when failing to build unlabelled component


tpaschalis:
The following configuration panics about 50% of the time. It likely has to do with the unlabelled local.file component, as adding a label to it makes it so that it never panics. 

```river
logging {
	level  = "debug"
	format = "logfmt"
}

local.file "this_file" {
	filename = "./cmd/agent/test-local-file.txt"
	detector = "fsnotify"
}

local.file {
	foo = "bar"
}

prometheus.scrape "default" {
	targets = [{
		"__address__"   = "localhost:12345",
		"dynamic_label" = local.file.this_file.content,
	}]
	forward_to = [prometheus.remote_write.default.receiver]

	scrape_config {
		job_name = "default"
	}
}

prometheus.remote_write "default" {
	remote_write {
		url = "http://localhost:9009/api/prom/push"
	}
}
```

The produced panic
```
$ EXPERIMENTAL_ENABLE_FLOW=true ./cmd/agent/agent run ./cmd/agent/example-config.river
ts=2022-08-31T08:39:07.098555Z level=error msg="failed to evaluate component" component=local.file err="decoding River: missing required attribute \"filename\""
panic: runtime error: index out of range [2] with length 2

goroutine 1 [running]:
github.com/grafana/agent/pkg/flow/internal/controller.(*valueCache).buildValue(0x14000d9a390, {0x14000d9b260, 0x2, 0x4?}, 0x2)
	/Users/tpaschalis/GitRepos/agent/pkg/flow/internal/controller/value_cache.go:133 +0x310
github.com/grafana/agent/pkg/flow/internal/controller.(*valueCache).buildValue(0x14000d9a390, {0x14000d9b200, 0x2, 0x5?}, 0x1)
	/Users/tpaschalis/GitRepos/agent/pkg/flow/internal/controller/value_cache.go:139 +0x17c
github.com/grafana/agent/pkg/flow/internal/controller.(*valueCache).BuildContext(0x14000d9a390, 0x0)
	/Users/tpaschalis/GitRepos/agent/pkg/flow/internal/controller/value_cache.go:105 +0x2cc
github.com/grafana/agent/pkg/flow/internal/controller.(*Loader).evaluate(0x14000dd8630, 0x0?, 0x14000da0540)
	/Users/tpaschalis/GitRepos/agent/pkg/flow/internal/controller/loader.go:270 +0x30
[...]

PR #2164, opened 2022-09-16
flow: handle singleton components

tpaschalis:

#### PR Description
This PR enforces that singleton components _must_ have a label and that non-singleton components _must not_ have one. 

The error reporting looks like this

```
Error: ./example-config.river:13:1: Component "local.file" must have a label

12 |
13 | local.file {
   | ^^^^^^^^^^
14 | }

Error: ./example-config.river:16:1: Singleton component "prometheus.integration.node_exporter" must not have a label

15 |
16 | prometheus.integration.node_exporter "foo" {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
17 | }
```

No issue filed specifically for the handling of singleton components, but should deal with #2095.

#### Notes to the Reviewer
* We could add this check in an earlier point eg. in `flow.ReadFile()`, but that would stop at the first error and not allow us to report all offending components.
* The error reporting here reads a little repetitive, but I would prefer we don't get too smart about it yet with a helper
* The `singleton` concept is not explained in the docs, and might even be on its way out as per #1861. In any case, we might want to update our docs to be more explicit on which components _must_ have a label and which components must not.

#### PR Checklist

- [ ] CHANGELOG updated (N/A)
- [ ] Documentation added
- [X] Tests updated

### DIFF
--- a/pkg/flow/internal/controller/loader.go
+++ b/pkg/flow/internal/controller/loader.go
@@ -168,9 +168,9 @@ func (l *Loader) populateGraph(g *dag.Graph, blocks []*ast.BlockStmt) diag.Diagn
 			if registration.Singleton && block.Label != "" {
 				diags.Add(diag.Diagnostic{
 					Severity: diag.SeverityLevelError,
-					Message:  fmt.Sprintf("Singleton component %q must not have a label", componentName),
-					StartPos: block.NamePos.Position(),
-					EndPos:   block.NamePos.Add(len(componentName) - 1).Position(),
+					Message:  fmt.Sprintf("Component %q does not support labels", componentName),
+					StartPos: block.NamePos.Add(len(componentName) + 1).Position(),
+					EndPos:   block.NamePos.Add(len(componentName) + 1).Add(len(block.Label) + 1).Position(),
 				})
 				continue
 			}
--- a/pkg/flow/internal/controller/loader.go
+++ b/pkg/flow/internal/controller/loader.go
@@ -154,10 +154,31 @@ func (l *Loader) populateGraph(g *dag.Graph, blocks []*ast.BlockStmt) diag.Diagn
 			c.UpdateBlock(block)
 		} else {
 			componentName := strings.Join(block.Name, ".")
-			if _, exists := component.Get(componentName); !exists {
+			registration, exists := component.Get(componentName)
+			if !exists {
 				diags.Add(diag.Diagnostic{
 					Severity: diag.SeverityLevelError,
 					Message:  fmt.Sprintf("Unrecognized component name %q", componentName),
+					StartPos: block.NamePos.Position(),
+					EndPos:   block.NamePos.Add(len(componentName) - 1).Position(),
+				})
+				continue
+			}
+
+			if registration.Singleton && block.Label != "" {
+				diags.Add(diag.Diagnostic{
+					Severity: diag.SeverityLevelError,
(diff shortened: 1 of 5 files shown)


rfratto (approved):
nice! Easier to fix than I thought :)

pkg/flow/internal/controller/loader.go
```diff
@@ -164,6 +165,26 @@ func (l *Loader) populateGraph(g *dag.Graph, blocks []*ast.BlockStmt) diag.Diagn
 				continue
 			}
 
+			if registration.Singleton && block.Label != "" {
+				diags.Add(diag.Diagnostic{
+					Severity: diag.SeverityLevelError,
+					Message:  fmt.Sprintf("Singleton component %q must not have a label", componentName),
+					StartPos: block.NamePos.Position(),
+					EndPos:   block.NamePos.Add(len(componentName) - 1).Position(),
```
rfratto, line 173: Do we have the position of the label available? It might be useful to point to the label instead of the component name

pkg/flow/internal/controller/loader.go
```diff
@@ -164,6 +165,26 @@ func (l *Loader) populateGraph(g *dag.Graph, blocks []*ast.BlockStmt) diag.Diagn
 				continue
 			}
 
+			if registration.Singleton && block.Label != "" {
+				diags.Add(diag.Diagnostic{
+					Severity: diag.SeverityLevelError,
+					Message:  fmt.Sprintf("Singleton component %q must not have a label", componentName),
```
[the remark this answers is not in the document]
mattdurham: Agreed, I think we should veer away from singleton. Once we do submodules the term singleton may get a bit complicated to explain whereas no labels is easier.

pkg/flow/internal/testcomponents/singleton.go
```diff
@@ -0,0 +1,59 @@
+package testcomponents
+
+import (
+	"context"
+
+	"github.com/go-kit/log"
+	"github.com/grafana/agent/component"
+)
+
+func init() {
```
mattdurham, line 10: Discussion point related to my above comment: IMO Singleton is fine for now, but I feel submodules will force us to define this more concretely or use different verbiage.

pkg/flow/internal/controller/loader.go
tpaschalis: Hm, not directly, but we can offset for the component name, and component name + label (including the quotation marks), and it seems to be working nicely
```
StartPos: block.NamePos.Add(len(componentName) + 1).Position(),
EndPos:   block.NamePos.Add(len(componentName) + 1).Add(len(block.Label) + 1).Position(),

Error: ./example-config.river:16:38: Component "prometheus.integration.node_exporter" does not support labels

15 |
16 | prometheus.integration.node_exporter "foo" {
   |                                      ^^^^^
17 | }
```

rfratto: That won't always work depending on spacing, I don't think. Do you want to try updating the AST + parser to include the label position?

It's probably something we generally need, I remember a few times where I wish I had access to the label position but didn't.

tpaschalis: Yes! I see that you're already setting the LabelPos when you're parseBlockName, so it's a one-line change.

I'll see that it get properly handled in other places as well and update this PR!

rfratto (approved):

PR #2167, opened 2022-09-16
river/encoding: sort object keys

rfratto:
Sort object keys so River objects backed by Go maps appear in a consistent order.

### DIFF
--- a/pkg/river/encoding/map.go
+++ b/pkg/river/encoding/map.go
@@ -31,7 +31,7 @@ func (mf *mapField) convertMap(val value.Value) error {
 
 	keys := val.Keys()
 
-	// If v isn't an ordered object (i.e., a go map), sort the keys so they have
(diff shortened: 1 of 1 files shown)


pkg/river/encoding/map.go
```diff
@@ -27,7 +28,16 @@ func (mf *mapField) hasValue() bool {
 func (mf *mapField) convertMap(val value.Value) error {
 	mf.Type = objectType
 	fields := make([]*keyField, 0)
-	for _, key := range val.Keys() {
+
+	keys := val.Keys()
+
+	// If v isn't an ordered object (i.e., a go map), sort the keys so they have
```
rfratto, line 34:
```suggestion
	// If v isn't an ordered object (i.e., a Go map), sort the keys so they have
```

mattdurham (approved):
LGTM

PR #2166, opened 2022-09-16
web/ui: initial commit of Flow UI

rfratto:
This is the initial commit of the Flow UI. It has two views:

* A component table view 
* A component detail view, to debug individual components 

This initial commit does not yet wire the UI into the Grafana Agent binary. This will happen in a future PR.

To run locally, do the following:

* Run Grafana Agent Flow 
* In another tab, run `yarn` in the `web/ui` folder.
* Then run `yarn start` in the `web/ui` folder.
* Navigate to http://localhost:3000 to access the UI. 

You must have the following installed to run the UI locally:

* Yarn >= v1.22
* Node.js >= v16

### Component table view

image

### Component detail view 

image

PR #2171, opened 2022-09-16
web/ui: Add graph view

rfratto:
This PR follows up on #2166 and adds a Graph view which can be used to visualize the relationships between running components:

image

The logic here is the most complicated and probably needs the most attention.

### DIFF
--- a/web/ui/src/features/component/ComponentGraph.tsx
+++ b/web/ui/src/features/component/ComponentGraph.tsx
@@ -217,10 +217,11 @@ export const ComponentGraph: FC<ComponentGraphProps> = (props) => {
       .append('path')
       .attr(
         'd',
+        // Draw an arrow shape
         d3.line()([
-          [0, 0],
-          [0, 20],
-          [20, 10],
+          [0, 0], // Bottom left of arrow
+          [0, 20], // Top left of arrow
+          [20, 10], // Middle point of arrow
         ])
       )
       .attr('fill', '#c8c9ca');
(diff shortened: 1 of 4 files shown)


web/ui/src/features/component/ComponentGraph.tsx
```diff
@@ -0,0 +1,394 @@
...
+      });
+
+    svgSelection.call(zoom).call(zoom.transform, d3Zoom.zoomIdentity);
+
+    // Add a marker element so we can draw an arrow pointing between nodes.
+    svgWrapper
+      .append('defs')
+      .append('marker')
+      .attr('id', 'arrow')
+      .attr('viewBox', [0, 0, 20, 20])
```
mattdurham, line 211: What are the 0,0,20,20 referenced a few times? Are those the dimensions for the box with the component name?

rlankfo (approved):
¯\_(ツ)_/¯

web/ui/src/features/component/ComponentGraph.tsx
rfratto: Yeah, this assigns a [viewBox attribute](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/viewBox) with x=0, y=0, w=20, h=20. It's basically the viewport size where elements can be drawn.

Do you see 0,0,20,20 anywhere else? I only see the one instance here

mattdurham: Ten or so lines down that I assume is drawing the border

rfratto: Oh, that's slightly different. That's the actual arrow shape. The 0,0,20,20 is the actual box the arrow shape can be drawn in. 


I added some comments to try to clear it up, is it better now?

mattdurham (approved):
LGTM

PR #2172, opened 2022-09-16
cmd/agent: expose the Flow UI from the HTTP server

rfratto:
This commit exposes the Flow UI from Grafana Agent Flow's HTTP server. By default, the UI is available via the the `/` path. The path where the UI (and the UI's API) is hosted can be changed via the `--server.http.ui-path-prefix` flag.

By default, Grafana Agent Flow will look for the UI at the `./web/ui/build` path relative to the working directory from where Grafana Agent Flow is running. A built UI at `./web/ui/build` can be baked into the binary by setting a `builtinassets` build flag when building.

To build the UI, run `go generate -tags builtinassets ./web/ui` or run `yarn && yarn build` from the `./web/ui` directory. 

This change does not yet modify the Makefile or Dockerfile to include the builtin UI from release assets. This will be done in a future change.

### DIFF
--- a/docs/flow/reference/cli/run.md
+++ b/docs/flow/reference/cli/run.md
@@ -32,3 +32,4 @@ The following flags are supported:
 * `--server.http.listen-addr`: Address to listen for HTTP traffic on (default `127.0.0.1:12345`).
 * `--storage.path`: Base directory where components can store data (default `data-agent/`).
 * `--debug.endpoints.enabled`: Whether to enable HTTP debug endpoints (default `true`).
+* `--server.http.ui-path-prefix`: Base path where the UI will be exposed (default `/`).
--- a/docs/flow/reference/cli/run.md
+++ b/docs/flow/reference/cli/run.md
@@ -29,7 +29,7 @@ various debugging needs.
 
 The following flags are supported:
 
-* `--server.http.listen-addr`: Address to listen for HTTP traffic on (default `127.0.0.1:12345`).
-* `--storage.path`: Base directory where components can store data (default `data-agent/`).
 * `--debug.endpoints.enabled`: Whether to enable HTTP debug endpoints (default `true`).
+* `--server.http.listen-addr`: Address to listen for HTTP traffic on (default `127.0.0.1:12345`).
 * `--server.http.ui-path-prefix`: Base path where the UI will be exposed (default `/`).
+* `--storage.path`: Base directory where components can store data (default `data-agent/`).
--- a/web/ui/assets_nobuiltin.go
+++ b/web/ui/assets_nobuiltin.go
@@ -1,0 +1,11 @@
+//go:build !builtinassets
+// +build !builtinassets
+
+package ui
+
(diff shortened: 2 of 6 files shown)


docs/flow/reference/cli/run.md
```diff
@@ -32,3 +32,4 @@ The following flags are supported:
 * `--server.http.listen-addr`: Address to listen for HTTP traffic on (default `127.0.0.1:12345`).
 * `--storage.path`: Base directory where components can store data (default `data-agent/`).
 * `--debug.endpoints.enabled`: Whether to enable HTTP debug endpoints (default `true`).
+* `--server.http.ui-path-prefix`: Base path where the UI will be exposed (default `/`).
```
rfratto, line 35: Maybe these should be sorted alphabetically

web/ui/assets_nobuiltin.go
```diff
@@ -0,0 +1,11 @@
+//go:build !builtinassets
+// +build !builtinassets
+
+package ui
+
+import "net/http"
+
+// Assets contains the UI's assets.
+func Assets() http.FileSystem {
+	return http.Dir("./web/ui/build")
```
mattdurham, line 10: Does this work on windows?

web/ui/ui.go
```diff
@@ -0,0 +1,204 @@
...
+	"github.com/prometheus/common/server"
+)
+
+// RegisterRoutes registers routes to the provided mux.Router for serving the
+// Grafana Agent Flow UI. The UI will be served relative to pathPrefix. If no
+// pathPrefix is specified, the UI will be served at root.
+//
+// By default, the UI is retreived from the ./web/ui/build directory relative
+// to working directory, assuming that the Agent is run from the repo root.
+// However, if the builtinassets Go tag is present, the built UI will be
```
mattdurham, line 26: Is there a reason we offer both ways? For ease of debugging?

rfratto: Yeah, debugging is one benefit, but it also simplifies the build process for people who aren't working with the UI, since otherwise you need NodeJS and Yarn installed just to build the agent at all.

web/ui/assets_nobuiltin.go
rfratto: According to the docs this should be split by filepath.Separator, thanks.

web/ui/ui.go
```diff
@@ -0,0 +1,204 @@
...
+		_ = f.Close()
+		return nil, err
+	}
+
+	// Return the underlying file if we got a directory. Otherwise, we're going
+	// to create our own synthethic file.
+	if fi.IsDir() {
+		return f, nil
+	}
+	defer f.Close()
```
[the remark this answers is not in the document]
rfratto: No, the caller is supposed to close it. If it's not a directory, we return a new file, which is why we close the original one on defer in that path. I've expanded the comment to make this more clear

mattdurham (approved):
LGTM

PR #2189, opened 2022-09-20
docs/flow: revamp documentation page for discovery.relabel

tpaschalis:

#### PR Description
This PR adds a small note to the reference page for discovery.relabel about internal labels.

No issue filed.

#### Notes to the Reviewer
Let me know if I'm missing anything important we should let users know!

#### PR Checklist

- [ ] CHANGELOG updated (N/A)
- [X] Documentation added
- [ ] Tests updated (N/A)

### DIFF
--- a/docs/flow/reference/components/discovery.relabel.md
+++ b/docs/flow/reference/components/discovery.relabel.md
@@ -6,14 +6,27 @@ title: discovery.relabel
 
 # discovery.relabel
 
-`discovery.relabel` rewrites the label set of the input targets by applying one
-or more relabeling `rule`s. If no rules are defined, then the input targets
-will be exported as-is.
+In Flow, targets are defined as sets of key-value pairs called _labels_.
 
-The most common use of `discovery.relabel` is to filter Prometheus targets or
-standardize the label set that will be passed to a downstream component. The
-`rule` blocks will be applied to the label set of each target in
-order of their appearance in the configuration file.
+`discovery.relabel` rewrites the label set of the input targets by applying one
+or more relabeling rules. If no rules are defined, then the input targets are
+exported as-is.
+
+The most common use of `discovery.relabel` is to filter targets or standardize
+the target label set that is passed to a downstream component. The `rule`
+blocks are applied to the label set of each target in order of their appearance
+in the configuration file.
+
+Target labels which start with a double underscore `__` are considered
+internal, and may be removed by other Flow components prior to telemetry
+collection. To retain any of these labels, use a `labelmap` action to remove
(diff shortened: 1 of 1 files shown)


docs/flow/reference/components/discovery.relabel.md
```diff
@@ -6,15 +6,27 @@ title: discovery.relabel
...
 # discovery.relabel
 
+In Flow, targets are defined as sets of key-value pairs called _labels_.
+
 `discovery.relabel` rewrites the label set of the input targets by applying one
 or more `relabel_config` steps. If no relabeling steps are defined, then the
-input targets will be exported as-is.
+input targets are exported as-is.
 
 The most common use of `discovery.relabel` is to filter Prometheus targets or
```
rfratto, line 15: Do we want to drop the mention of Prometheus here? These targets will be used in a lot of places, even if the `prometheus.*` components are the only real use case for them at the moment.

We could just say "filter targets" instead.

docs/flow/reference/components/discovery.relabel.md
```diff
@@ -6,15 +6,27 @@ title: discovery.relabel
...
 
+Target labels which start with a double underscore `__` are considered as
+internal by Prometheus and are removed prior to scraping. To retain any of
+these labels post scraping, use a `labelmap` action. Service discovery
+mechanisms usually group their labels under `__meta_*`. For example, the
+discovery.kubernetes component populates a set of `__meta_kubernetes_*` labels
+to provide information about the discovered Kubernetes resources. If a
+relabeling step needs to store a label value temporarily, for example as the
+input to a subsequent relabeling step, use the `__tmp` label name prefix, as it
+is guaranteed to never be used by Prometheus itself.
```
rfratto, line 28: Can we remove the reference to Prometheus and scraping here to allow the discovery component to be logically decoupled from Prometheus? 

I think we can say something like:

> Target labels which start with a double underscore `__` are considered internal, and may be removed by other Flow components prior to telemetry collection. To retain any ...

rfratto (approved):
LGTM!

docs/flow/reference/components/discovery.relabel.md
tpaschalis: You're right. I've removed references to Prometheus and scraping, so that the description reads more naturally.

Do you think we should explain anything else related to relabeling here that a newcomer might find useful?

rfratto: Not that I can think of. This is a great first version to ship, and if anything comes up  we can add it later 👍

PR #2192, opened 2022-09-20
flow: Report usage of enabled Components

marctc:
#### PR Description

This PR enables the reporting of enabled components to grafana. This
would help us to identify which components are more used than others.

The information sent it's completely anonymous. This is the [full list](https://grafana.com/docs/agent/next/configuration/flags/#report-information-usage) of all
information sent by the reporter. The reporting can always be disabled using 
`--disable-reporting`.

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/cmd/agent/entrypoint.go
+++ b/cmd/agent/entrypoint.go
@@ -95,7 +95,7 @@ func NewEntrypoint(logger *server.Logger, cfg *config.Config, reloader Reloader)
 		return nil, err
 	}
 
-	ep.reporter, err = usagestats.NewReporter(logger, cfg)
+	ep.reporter, err = usagestats.NewReporter(logger)
 	if err != nil {
 		return nil, err
 	}
@@ -242,6 +242,11 @@ func (ep *Entrypoint) reloadHandler(rw http.ResponseWriter, r *http.Request) {
 	}
 }
 
+// getReporterMetrics creates the metrics map to send to usage reporter
+func (ep *Entrypoint) getReporterMetrics() map[string]interface{} {
+	return map[string]interface{}{"enabled-features": ep.cfg.EnabledFeatures}
+}
+
 // TriggerReload will cause the Entrypoint to re-request the config file and
 // apply the latest config. TriggerReload returns true if the reload was
 // successful.
@@ -327,7 +332,7 @@ func (ep *Entrypoint) Start() error {
 	ep.mut.Unlock()
 	if cfg.EnableUsageReport {
 		g.Add(func() error {
-			return ep.reporter.Start(srvContext)
+			return ep.reporter.Start(srvContext, ep.getReporterMetrics())
 		}, func(e error) {
 			srvCancel()
 		})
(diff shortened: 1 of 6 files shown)


pkg/usagestats/reporter.go
```diff
@@ -103,7 +100,7 @@ func agentSeedFileName() string {
 }
 
 // Start inits the reporter seed and start sending report for every interval
-func (rep *Reporter) Start(ctx context.Context) error {
+func (rep *Reporter) Start(ctx context.Context, metrics map[string]interface{}) error {
```
rfratto, line 103: `metrics` is going to be static throughout the lifetime of the agent, which wouldn't necessarily represent the actual state of the agent at the time a report is generated. Can we change `metrics` into `func() map[string]interface{}` and call it every time we generate a report instead?

cmd/agent/entrypoint.go
```diff
@@ -242,6 +242,11 @@ func (ep *Entrypoint) reloadHandler(rw http.ResponseWriter, r *http.Request) {
 	}
 }
 
+// getReporterMetrics creates the metrics map to send to usage reporter
+func (ep *Entrypoint) getReporterMetrics() map[string]interface{} {
+	return map[string]interface{}{"enabled-features": ep.cfg.EnabledFeatures}
```
rfratto, line 247: You need to hold `ep.mut` here before you can read `ep.cfg.EnabledFeatures`.

cmd/agent/flow_run.go
```diff
@@ -214,6 +219,24 @@ func (fr *flowRun) Run(configFile string) error {
 		defer func() { _ = srv.Shutdown(ctx) }()
 	}
 
+	// Report usage of enabled components
+	if !fr.disableReporting {
+		infos := f.ComponentInfos()
+		components := []string{}
+		for _, info := range infos {
+			components = append(components, info.Name)
+		}
```
rfratto, line 228: There may be many components with the same name (such as more than one `discovery.kubernetes` component). We'll probably want to only put each component name once.

cmd/agent/flow_run.go
```diff
@@ -83,6 +85,8 @@ depending on the nature of the reload error.
 	cmd.Flags().StringVar(&r.storagePath, "storage.path", r.storagePath, "Base directory where components can store data")
 	cmd.Flags().
 		BoolVar(&r.enableDebugEndpoints, "debug.endpoints.enabled", r.enableDebugEndpoints, "Enables /debug/ HTTP endpoints")
+	cmd.Flags().
+		BoolVar(&r.disableReporting, "disable-reporting", false, "Disable reporting of enabled components to Grafana.")
```
rfratto, line 89: nit: Get the default from `r` for consistency with other flags.

```suggestion
		BoolVar(&r.disableReporting, "disable-reporting", r.disableReporting, "Disable reporting of enabled components to Grafana.")
```

rfratto (approved):
LGTM, thanks!

PR #2193, opened 2022-09-20
Flow: Remove /debug/* endpoints in favor of Flow UI

### LINKED ISSUE #2000: Flow: /debug/graph does not work in containers if graphviz is not installed

### LINKED ISSUE #2006: Flow: Breakup /debug/scope and /debug/config into multiple endpoints
"Give me everything" can still be an option, but production workloads are going to want to take things granularly so you can load it into a browser.

rfratto:
This also removes the `--debug.endpoints.enabled` flag, which had little little purpose after the removal of the endpoints.

It may be desired in the future to add a flag to disable the debugging UI, but that is out of scope for the work of this PR.

This resolves a few issues specific to /debug/* endpoints:

* Closes #2000
* Closes #2006


pkg/flow/flow_http.go
```diff
@@ -85,49 +44,6 @@ func (f *Flow) ComponentHandler() http.HandlerFunc {
...
-			http.Error(w, err.Error(), http.StatusInternalServerError)
-		} else {
-			_, _ = io.Copy(w, &buf)
-		}
-	}
-}
-
-// configBytes dumps the current state of the flow config as River.
-func (f *Flow) configBytes(w io.Writer, debugInfo bool) (n int64, err error) {
-	file := builder.NewFile()
```
rfratto, line 107: It looks like we could remove the pkg/river/token/builder package since it's not being used, but I'd like to keep that in for now since it serves as a good reference for how to work with River values.

tpaschalis (approved):
LGTM for reducing the scope.

mattdurham (approved):
LGTM

PR #2196, opened 2022-09-20
Resolve issues #2186 and #2187

### LINKED ISSUE #2186: Flow UI: HTTP 500 with some discovery.kubernetes components
I got an HTTP 500 response when looking at a `discovery.kubernetes.stackdriver_exporter` component:

```
unable to find key __meta_kubernetes_pod_annotation_cni_projectcalico_org_podIP for value type 5
```

The component is defined as follows:

```river
discovery.kubernetes "stackdriver_exporter" {
  role = "pod"
}
```

### LINKED ISSUE #2187: Flow UI: HTTP 500 with some discovery.relabel components
I got an HTTP 500 response when looking at a discovery.relabel component:

```
unable to find value for []discovery.Target in convertAttribute
```

<details>
<summary>Component definition</summary>

The component is defined as follows:

```river
discovery.relabel "example" {
	targets = concat(discovery.kubernetes.example.targets)

	relabel_config {
		action = "keep"
		regex = ".*-metrics"
		source_labels = ["__meta_kubernetes_pod_container_port_name"]
	}
	relabel_config {
		action = "drop"
		regex = ""
		source_labels = ["__meta_kubernetes_pod_label_name"]
	}
	relabel_config {
		action = "drop"
		regex = "Succeeded|Failed"
		source_labels = ["__meta_kubernetes_pod_phase"]
	}
	relabel_config {
		action = "replace"
		regex = "(https?)"
		replacement = "$1"
		source_labels = ["__meta_kubernetes_pod_annotation_prometheus_io_scheme"]
		target_label = "__scheme__"
	}
	relabel_config {
		action = "replace"
		regex = "(.+)"
		replacement = "$1"
		source_labels = ["__meta_kubernetes_pod_annotation_prometheus_io_path"]
		target_label = "__metrics_path__"
	}
	relabel_config {
		action = "replace"
		regex = "(.+?)(\\:\\d+)?;(\\d+)"
		replacement = "$1:$3"
		source_labels = ["__address__", "__meta_kubernetes_pod_annotation_prometheus_io_port"]
		target_label = "__address__"
	}
	relabel_config {
		action = "labelmap"
		regex = "__meta_kubernetes_pod_annotation_prometheus_io_param_(.+)"
		replacement = "__param_$1"
	}
	relabel_config {
		action = "labelmap"
		regex = "__meta_kubernetes_pod_label_prometheus_io_label_(.+)"
	}
	relabel_config {
		action = "labelmap"
		regex = "__meta_kubernetes_pod_annotation_prometheus_io_label_(.+)"
	}
	relabel_config {
		action = "replace"
		replacement = "$1"
		separator = "/"
		source_labels = ["__meta_kubernetes_namespace", "__meta_kubernetes_pod_label_name"]
		target_label = "job"
	}
	relabel_config {
		action = "replace"
		source_labels = ["__meta_kubernetes_namespace"]
		target_label = "namespace"
	}
	relabel_config {
		action = "replace"
		source_labels = ["__meta_kubernetes_pod_name"]
		target_label = "pod"
	}
	relabel_config {
		action = "replace"
		source_labels = ["__meta_kubernetes_pod_container_name"]
		target_label = "container"
	}
	relabel_config {
		action = "replace"
		separator = ":"
		source_labels = ["__meta_kubernetes_pod_name", "__meta_kubernetes_pod_container_name", "__meta_kubernetes_pod_container_port_name"]
		target_label = "instance"
	}
	relabel_config {
		action = "keep"
		regex = "example-pod"
		source_labels = ["__meta_kubernetes_pod_label_name"]
	}
	relabel_config {
		replacement = "dev-us-central-0"
		target_label = "cluster"
	}
	relabel_config {
		action = "hashmod"
		modulus = 4
		target_label = "__tmp_hash"
	}
	relabel_config {
		action = "keep"
		regex = "^3$"
		source_labels = ["__tmp_hash"]
	}
	relabel_config {
		action = "labeldrop"
		regex = "__tmp_hash"
	}
}
```
</details>

mattdurham:
### PR Description

Resolve errors with API where it was not handling empty values correctly.

Closes #2186 
Closes #2187

### DIFF
--- a/pkg/river/encoding/map.go
+++ b/pkg/river/encoding/map.go
@@ -42,8 +42,9 @@ func (mf *mapField) convertMap(val value.Value) error {
 
 		kf.Key = key
 		mapVal, found := val.Key(key)
+		// The above val.Key(key) will return false if the value is nil or zero.
 		if !found {
-			return fmt.Errorf("unable to find key %s for value type %d", key, val.Type())
+			continue
 		}
 		rv, err := convertRiverValue(mapVal)
 		if err != nil {
--- a/pkg/river/encoding/map.go
+++ b/pkg/river/encoding/map.go
@@ -42,7 +42,8 @@ func (mf *mapField) convertMap(val value.Value) error {
 
 		kf.Key = key
 		mapVal, found := val.Key(key)
-		// The above val.Key(key) will return false if the value is nil or zero.
+		// The above val.Key(key) will return false if the value is nil or zero. This is an issue that will be evaluated
+		// later.
 		if !found {
 			continue
 		}
--- a/pkg/river/encoding/attribute.go
+++ b/pkg/river/encoding/attribute.go
@@ -54,7 +54,7 @@ func (af *attributeField) convertAttribute(val value.Value, f rivertags.Field) e
 		return err
 	}
 	if !rv.hasValue() {
-		return fmt.Errorf("unable to find value for %T in convertAttribute", val.Interface())
+		return nil
 	}
 	af.valField = rv
 	return nil
--- a/pkg/river/encoding/attribute_test.go
+++ b/pkg/river/encoding/attribute_test.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"testing"
 
+	"github.com/grafana/agent/component/discovery"
 	"github.com/grafana/agent/pkg/river/internal/rivertags"
 	"github.com/grafana/agent/pkg/river/internal/value"
 	"github.com/stretchr/testify/require"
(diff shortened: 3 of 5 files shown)


pkg/river/encoding/map.go
```diff
@@ -42,8 +42,9 @@ func (mf *mapField) convertMap(val value.Value) error {
 
 		kf.Key = key
 		mapVal, found := val.Key(key)
```
rfratto, line 44: pkg/river/token/builder ignores the found attribute and always converts the value: https://github.com/grafana/agent/blob/f77453c35f9de489faf04126305fab2cda691aa2/pkg/river/token/builder/value_tokens.go#L77

This is useful when the map has an empty value, but you'd still expect the key to show up. I think something similar should happen probably happen here.

mattdurham: That could get noisy, but will add it for now and re-evaluate later.

mattdurham: The only wrinkle I see is that `""` is returned as TypeNull instead of a blank string.

rfratto: I think doing a `continue` for now is ok, we need to think more about how everything behaves here. That being said, I think this is a bug: we're iterating over the list of keys from `val.Keys()`, so it's unexpected that `found` could ever be false here. 

Can you add a comment noting this as a TODO to fix the issue? We can revisit after the initial release since I don't think this edge case is a release blocker.

rfratto (approved):
LGTM after adding a TODO comment for fixing the issue where value.Value.Key returns false for a key that we definitely know exists.

pkg/river/encoding/map.go
mattdurham: Discussed and this looks like a bug/adverse behavior in `Key` func, so added a todo and will address it later.

PR #2197, opened 2022-09-21
web/ui: respond better to reverse proxies using Nginx subfiltering

rfratto:
This changes the UI to use a <base> HTML element to determine the root path of the application. This is useful in scenarios where an NGINX reverse proxy is used with the subfiltering module; this allows NGINX to modify the <base> HTML element so generated URLs are still correct.

In this scenario, where a reverse proxy is using subfiltering to modify the HTTP response, the `--server.http.ui-path-prefix` flag should **not** be set. The reverse proxy should instead trim out the common prefix on requests, and inject the prefix back into responses through subfiltering.


mattdurham (approved):
LGTM

PR #2198, opened 2022-09-21
web/ui: reduce total graph size

rfratto:
* Assign synthetic nodes a size of [0, 0] to prevent extra unnecessary padding in the graph.

* Determine the width of the node based on the component name and label indepednently, rather than the combined ID. This is a safe change because the full ID is never rendered, while the name and label are rendered on separate lines.


mattdurham (approved):
LGTM

PR #2202, opened 2022-09-21
script for generating runt.sh and runt.sh itself

mattdurham:
#### PR Description

These are scripts to generate the tutorial script and the tutorial script itself. Split the PR into two parts, this one concerns the scripts and the second will implement those changes in the tutorials itself, along with general cleanup of tutorials and images from the UI. 

This is also linux/osx specific, at some point might do a powershell version for windows.

### DIFF
--- a/docs/sources/flow/tutorials/assets/generate.sh
+++ b/docs/sources/flow/tutorials/assets/generate.sh
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 
-# This script generates runt.sh script that pulls down the needed files and runs them.
+# This script generates runt.sh script that pulls down the needed files and runs them for the tutorials themselves.
+# This needs to be reran anytime new tutorials are added with new docker composes.
+# The runt.sh file is meant to download all the needed files for the example and for them to be used. 
 
 echo "#!/usr/bin/env bash" > runt.sh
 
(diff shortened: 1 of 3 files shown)


karengermond (approved):
LGTM

docs/sources/flow/tutorials/assets/generate.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
jcreixell, line 1: not super important, but you may want to use `#!/usr/bin/env bash` for better portability and consistency with other scripts in this repo

docs/sources/flow/tutorials/assets/generate.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+
+# This script generates runt.sh script that pulls down the needed files and runs them.
```
rfratto, line 3: I think this comment needs to explain what's going on here a little more. None of us are bash experts, so I find it a little hard to follow. 

I think explaining what the "needed files" are would go a long way here.

I think we should also comment when this script would need to be re-run.

rfratto (approved):
The approval here is a bit rubber-stampy. If the scripts lint, that's good enough for me, but I find the generate.sh hard to understand and would like to see more comments to explain what's going on so we know what correct output looks like.

PR #2203, opened 2022-09-21
web/ui: fix issue where clicking on navigation didn't change position

rfratto:
While #2200 fixed the links, the <base> tag and <Link> component still changed the behavior of how the page position changes when clicking on a link in the sidebar.

Adding target="_top" ensures clicking the link will change the proper browsing context to scroll the proper place in the page.


mattdurham (approved):
👍

PR #2211, opened 2022-09-23
river: delay reporting of component traversal errors

### LINKED ISSUE #2110: river: duplicate/conflicting error reporting for non-existing identifier
When someone tries to configure an attribute with a non-existing idenitifier such as in the following example, he's presented with two different error messages, about it not being a valid identifier _and_ not being a valid component.

One of the messages comes from Flow's component references resolving, while the other one from River's VMs when looking up an identifier from the scope.

```
Error: ./cmd/agent/example-config.river:11:14: component "foo" does not exist

10 |     detector = "fsnotify"
11 |     is_secret = foo
   |                 ^
12 | }

Error: ./cmd/agent/example-config.river:11:14: identifier "foo" does not exist

10 |     detector = "fsnotify"
11 |     is_secret = foo
   |                 ^^^
12 | }
```

I'm not sure what's the best approach here, so I'll try to come back with an actual proposal; maybe having both errors is a valid approach?

tpaschalis:

#### PR Description
This PR removes the error reporting from graph traversals in the controller. Instead, these errors are reported later on by River's VM as not found identifiers.

Since there are no more errors being reported from this branch of the codebase, the diagnostics are unwired as well.

To be honest, I'm not 100% sure whether this is a good solution and if it makes sense, so maybe this will kickstart a conversation instead.

Fixes #2110

#### PR Checklist

- [ ] CHANGELOG updated (N/A)
- [ ] Tests updated

### DIFF
--- a/pkg/flow/internal/controller/component_references.go
+++ b/pkg/flow/internal/controller/component_references.go
@@ -5,7 +5,6 @@ import (
 
 	"github.com/grafana/agent/pkg/flow/internal/dag"
 	"github.com/grafana/agent/pkg/river/ast"
-	"github.com/grafana/agent/pkg/river/diag"
 	"github.com/grafana/agent/pkg/river/vm"
 )
 
@@ -26,12 +25,8 @@ type Reference struct {
 
 // ComponentReferences returns the list of references a component is making to
 // other components.
-func ComponentReferences(cn *ComponentNode, g *dag.Graph) ([]Reference, diag.Diagnostics) {
-	var (
-		traversals = componentTraversals(cn)
-
-		diags diag.Diagnostics
-	)
+func ComponentReferences(cn *ComponentNode, g *dag.Graph) []Reference {
+	traversals := componentTraversals(cn)
 
 	refs := make([]Reference, 0, len(traversals))
 	for _, t := range traversals {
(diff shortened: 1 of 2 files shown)


rfratto:
Something I'd like to understand about #2110 is why both errors showed up. 

```
Error: ./cmd/agent/example-config.river:11:14: component "foo" does not exist

10 |     detector = "fsnotify"
11 |     is_secret = foo
   |                 ^
12 | }

Error: ./cmd/agent/example-config.river:11:14: identifier "foo" does not exist

10 |     detector = "fsnotify"
11 |     is_secret = foo
   |                 ^^^
12 | }
```

The second error is implying that the VM was trying to evaluate the component. IMO the VM shouldn't be trying to evaluate it if the graph failed to process the component. 

Generally, I'm concerned with the scalability of having the graph loader report invalid references. If/when we add support for variables, the graph loader would have to check to see if an identifier is referencing a variable to prevent it from reporting errors, and that seems well beyond the problem domain for the Flow controller. Maybe this isn't too much of a problem; we might be able to eventually have `river/vm` report if something is a valid identifier so the Flow graph loader wouldn't have to do it.

tpaschalis:
I'm closing this PR as the linked issue has already been fixed along the way.

PR #2212, opened 2022-09-23
pkg/river: introduce high-level marshal/unmarshal API

rfratto:
This commit introduces a high-level river.Marshal* and river.Unmarshal* API. The API serves mostly as documentation for how to properly use the low level APIs.

Flow component tests which previously used the low-level APIs have been replaced with the high-level APIs.

Sorry for the size of this PR :) Almost of the entirety of the new size is documentation.

### DIFF
--- a/pkg/river/river.go
+++ b/pkg/river/river.go
@@ -116,7 +116,8 @@ func Marshal(v interface{}) ([]byte, error) {
 // String values encode to River strings.
 //
 // Array and slice values encode to River arrays, except that []byte is
-// converted into a string. Nil slices encode as an empty array.
+// converted into a River string. Nil slices encode as an empty array and nil
+// []byte slices encode as an empty string.
 //
 // Structs encode to River objects, using Go struct field tags to determine the
 // resulting structure of the River object. Each exported struct field with a
--- a/pkg/river/river.go
+++ b/pkg/river/river.go
@@ -178,7 +178,7 @@ func (enc *Encoder) Encode(v interface{}) error {
 // EncodeValue converts the value pointed to by v into a River value and writes
 // the result to the Decoder's output stream.
 //
-// See the documentation for MarshalValue for details about the converstion of
+// See the documentation for MarshalValue for details about the conversion of
 // Go values into River values.
 func (enc *Encoder) EncodeValue(v interface{}) error {
 	expr := builder.NewExpr()
--- a/pkg/river/river.go
+++ b/pkg/river/river.go
@@ -1,0 +1,322 @@
+// Package river implements a high-level API for decoding and encoding River
+// configuration files. The mapping between River and Go values is described in
+// the documentation for the Unmarshal and Marshal functions.
+//
+// Lower-level APIs which give more control over configuration evaluation are
+// available in the inner packages. The implementation of this package is
+// minimal and serves as a reference for how to consume the lower-level
+// packages.
+package river
+
+import (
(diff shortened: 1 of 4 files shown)


pkg/river/river.go
```diff
@@ -0,0 +1,322 @@
...
+//
+// The river tag specifies a name, possibly followed by a comma-separated list
+// of options. The name must be empty if the provided options do not support a
+// name being defined. The following provides examples for all supported struct
+// field tags with their meanings:
+//
+//	// Field appears as a block named "example". It will always appear in the
+//	// resulting encoding. When decoding, "example" is treated as a required
+//	// block and must be present in the source text.
+//	Field struct{...} `river:"example,block"`
```
tpaschalis, line 51: Is there a difference to explain between having `struct {...}` or `*struct{...}` here?

pkg/river/river.go
```diff
@@ -0,0 +1,322 @@
...
+	expr := builder.NewExpr()
+	expr.SetValue(v)
+
+	_, err := expr.WriteTo(enc.w)
+	return err
+}
+
+// Unmarshal converts the River configuration file specified by in and stores
+// it in the struct value pointed to by v. If v is nil or not a pointer,
+// Unmarshal panics. The configuration specified by in may use expressions to
```
tpaschalis, line 193: I'm not sure what was meant to be here "The configuration specified by in ..."

pkg/river/river.go
```diff
@@ -0,0 +1,322 @@
...
+// allocating maps, slices, and pointers as necessary, with the following
+// additional rules:
+//
+// After converting a River value into its Go value counterpart, the Go value
+// may be converted into a capsule if the capsule type implements
+// ConvertibleIntoCapsule.
+//
+// To unmarshal a River object into a struct, UnmarshalValue matches incoming
+// object fields to the river struct tags specified by v. Incoming object
+// fields which do not match to a river struct tag cause a decoding error.
```
tpaschalis, line 237: That means that superfluous fields will cause a decoding error, and for the unmarshaling to work correctly the incoming fields must match 1:1 the exported fields that have a `river` tag, right?

tpaschalis (approved):
LGTM, not much to add at the moment!

pkg/river/river.go
```diff
@@ -0,0 +1,322 @@
...
+// Otherwise, MarshalValue uses the following type-dependent default encodings:
+//
+// Boolean values encode to River bools.
+//
+// Floating point, integer, and Number values encode to River numbers.
+//
+// String values encode to River strings.
+//
+// Array and slice values encode to River arrays, except that []byte is
+// converted into a string. Nil slices encode as an empty array.
```
mattdurham, line 119: Does a nil []byte convert to "" or []?

pkg/river/river.go
rfratto: Yeah, if you want to support arbitrary field names you need to unmarshal into a map instead. 

Should I make that more clear in the comments?

tpaschalis: No, just wanted to verify. Our unmarshaller will be a little stricter than say json.Unmarshal which IIRC allows duplicate or missing fields, but I don't think that's a bad thing.

pkg/river/river.go
rfratto: Nope, either is fine.

mattdurham (approved):
LGTM

Issue #2213, opened 2022-09-24
Flow: add OpenTelemetry Collector components


rfratto:
This issue proposes a set of `otelcol` components to add first-class OpenTelemetry Collector support in Grafana Agent Flow. 

Adding first-class OpenTelemetry Collector support enables Grafana Agent to embrace Grafana's big tent philosophy, allowing the agent to support multiple vendors outside of the LGTM stack. 

## Proposal 

We will create a set of OpenTelemetry Collector components across three namespaces:

* `otelcol.receiver`: A list of OpenTelemetry Collector receivers 
* `otelcol.processor`: A list of OpenTelemetry Collector processors 
* `otelcol.exporter`: A list of OpenTelemetry Collector exporters 

A description of how to handle extensions is out of scope for this proposal. However, extensions for at least authentication will be included. 

> **NOTE**: Throughout this proposal, I use the term "consumer" in the same sense as OpenTelemetry Collector; a consumer is a component which can consume OpenTelemetry data sent from other components.  

### Pipeline construction

Grafana Agent Flow creates a pipeline by having each component reference other components where data will be sent. This contrasts OpenTelemetry Collector, where components are defined separately from the pipeline. 

For consistency with other Flow components, the set of `otelcol` components will form a pipeline the same way: by having each component reference other components where data will be sent. 
 
### Arguments 

All `otelcol` Flow components which send data to a consumer will support the following arguments: 

* `output.traces`: A list of consumers where traces should be forwarded. Only available if the configured component supports traces.  
* `output.metrics`: A list of consumers where metrics should be forwarded. Only available if the configured component supports metrics.
* `output.logs`: A list of consumers where logs should be forwarded. Only available if the configured component supports logs.

### Exported fields 

`otelcol` Flow components which act as consumers will support the following exported fields:

* `input`: A consumer for telemetry data. Any telemetry type supported by the OTel component can be sent to this consumer. 

Note that there is only one exported field for input for all telemetry types. This simplifies the config, and there doesn't seem to be much value in separating the fields. 

### Example config 

```river
otelcol.receiver.jaeger "default" {
  // Define the server settings where applications 
  // may push Jaeger-formatted traces to. 
  server {
    // Create a gRPC server with the default settings. 
    grpc {} 
  }

  output {  
    traces = [otelcol.processor.batch.default.input] 
  } 
}

otelcol.processor.batch "default" {
  output {
    traces = [otelcol.exporter.logging.default.input]
  }
}

[...]

PR #2224, opened 2022-09-27
component/otelcol: set up groundwork for OpenTelemetry Collector components

rfratto:
This commit introduces the component/otelcol package, which will eventually hold a collection of `otelcol.*` components.

Initial prototyping work for Flow OpenTelemetry Collector components was done in #1843, which demonstrated that the full set of code needed to implement OpenTelemetry Components is quite large. I will be splitting up the needed code across a few changes; this is the first.

This initial commit starts setting up the framework for running OpenTelemetry Collector components inside of Flow with a `Scheduler` struct.

Related to #2213.


component/otelcol/host.go
```diff
@@ -0,0 +1,41 @@
...
+	otelcomponent "go.opentelemetry.io/collector/component"
+	otelconfig "go.opentelemetry.io/collector/config"
+)
+
+// host implements otelcomponent.Host for Grafana Agent Flow.
+type host struct {
+	log log.Logger
+
+	extensions map[otelconfig.ComponentID]otelcomponent.Extension
+	exporters  map[otelconfig.DataType]map[otelconfig.ComponentID]otelcomponent.Exporter
```
mattdurham, line 16: Thoughts on adding more structure to this? Maps in maps always feel off to me.

rfratto: 🤷 It's the type OpenTelemetry requires for the Host interface. I can name the type but it still needs to be a map of maps.

mattdurham: Fair enough.

mattdurham (approved):
LGTM

rfratto:
@tpaschalis I'm going to merge this so I can open another PR, but if you want to review and have feedback, feel free and I'll address tomorrow :)

component/otelcol/internal/scheduler/scheduler.go
```diff
@@ -0,0 +1,150 @@
...
+
+// Scheduler implements manages a set of OpenTelemetry Collector components.
+// Scheduler is intended to be used from Flow components which need to schedule
+// OpenTelemetry Collector components; it does not implement the full
+// component.Component interface.
+//
+// Each OpenTelemetry Collector component has one instance per supported
+// telemetry signal, hence supporting multiple OpenTelemetry Collector
+// components inside the scheduler. Scheduler should only be used to manage
+// multiple instances of the same OpenTelemetry Collector component.
```
tpaschalis, line 26: @rfratto I'll read on and maybe things get clearer, but I'm trying to understand what we mean here.

An OpenTelemetry Collector components can support all three of metrics/logs/traces at the same time, right? What do we mean by the following phrase? Should we use one scheduler for multiple OTel Collector component for logs, and then for example, _another_ scheduler for the rest of OTel Collector component for traces+metrics?

> Scheduler should only be used to manage multiple instances of the _same_ OpenTelemetry Collector component.

component/otelcol/internal/scheduler/scheduler_test.go
```diff
@@ -0,0 +1,127 @@
...
+		// should cause our running component to terminate.
+		require.NoError(t, started.Wait(5*time.Second), "component did not start")
+		cancel()
+		require.NoError(t, stopped.Wait(5*time.Second), "component did not shutdown")
+	})
+}
+
+func newTriggerComponent() (component otelcomponent.Component, started, stopped *util.WaitTrigger) {
+	started = util.NewWaitTrigger()
+	stopped = util.NewWaitTrigger()
```
tpaschalis, line 92: Just wanted to say this looked nice :)

component/otelcol/internal/scheduler/scheduler.go
rfratto: Ah, sorry, I'll try to make this clearer over in #2227. 

It probably should say something like: 

> Schedule should only be use to manage the various telemetry signals of the same OpenTelemetry Collector Component.

Would that make more sense? (for example, use the schedule to manage the logs instance, metrics instance, and traces instance for `otelcol.receiver.otlp`)

tpaschalis: Is "otlp" here the label, right? So if we have an `otelcol.receiver.prod` and `otelcol.receiver.dev`, we'd have to have one scheduler for each?

Yes then, it makes sense in that context.

rfratto: `otelcol.receiver.otlp` is the component name :) But yes, that still applies. I'll make some tweaks in #2227 and ping you there.

PR #2225, opened 2022-09-27
Update tutorial

mattdurham:
Added images, commands, actual todos for the reader and general cleanup.


rfratto (commented):
The Graph screenshot looks like it's not rendering correctly 😞 What browser did you use for that? @eamonryan noticed that Firefox doesn't seem to display the SVG properly. The SVG works well for me in Safari and Chrome though.

rfratto (approved):
Nice work. I'll let Karen do editing reviews, but I'd like to see the graph screenshot replaced with one from Chrome where it renders properly.

mattdurham:
I was using Firefox, grabbed one from Chrome that is aligned properly.

hedss:
So whilst I understand the change to show multiple inputs, could this not be a separate tutorial (ideally, you need two separate sources though, as opposed to the same one). How about enabling metrics from the Grafana instance and using that as the second source.

Then you could keep the `cool`/`not_cool` service to show single source, mulitple sinks which is also a great tutorial demo, too!

(@eamonryan and I are currently going through the tutorials :) )

docs/sources/flow/tutorials/collecting-prometheus-metrics.md
```diff
@@ -10,16 +10,33 @@ Grafana Agent is a telemetry collector with the primary goal of moving telemetry
 # Prerequisites
 
 * [Docker](https://www.docker.com/products/docker-desktop)
-* Clone the [Agent Repository](https://github.com/grafana/agent) `git clone git@github.com:grafana/agent.git`
 
 # Scraping the Agent
 
-To quickly spin up an example environment, run the following: `CONFIG_FILE=agent.flow docker-compose -f ./assets/docker-compose.yaml up` with the [flow file](../assets/flow_configs/agent.flow). Allow the service to run for a few minutes, then navigate to the [metrics browser](http://localhost:3000/explore?orgId=1&left=%5B%22now-1h%22,%22now%22,%22Mimir%22,%7B%22refId%22:%22A%22,%22instant%22:true,%22range%22:true,%22exemplar%22:true,%22expr%22:%22agent_build_info%7B%7D%22%7D%5D).
+Run the following `curl https://raw.githubusercontent.com/grafana/agent/main/docs/sources/flow/tutorials/assets/runt.sh -O && bash ./runt.sh agent.flow`.
```
karengermond, line 16: Are users going to be comfortable running a script like this?  Just want to make sure in the long run that we're doing what people expect. I think what you had before was more typical, like in this tutorial: https://grafana.com/tutorials/play-with-grafana-mimir/.

docs/sources/flow/tutorials/collecting-prometheus-metrics.md
```diff
@@ -10,16 +10,33 @@ Grafana Agent is a telemetry collector with the primary goal of moving telemetry
...
 
 ![](../assets/agent_build_info.png)
 
 This example scrapes the Grafana Agent's `http://localhost:12345/metrics` endpoint and pushes those metrics to the Mimir instance.
 
+Navigate to `http://localhost:12345/graph` to view the Grafana Agent Flow UI. 
+
+![](../assets/graph.png)
+
+Clicking on the nodes will navigate to the associated component page. This allows viewing the state, health information and if applicable the debug information.
```
karengermond, line 34:
```suggestion
Click the nodes to navigate to the associated component page. There, you can view the state, health information, and, if applicable, the debug information.
```

docs/sources/flow/tutorials/collecting-prometheus-metrics.md
```diff
@@ -10,16 +10,33 @@ Grafana Agent is a telemetry collector with the primary goal of moving telemetry
...
+Allow the Grafana Agent to run for two minutes, then navigate to [Grafana](http://localhost:3000/explore?orgId=1&left=%5B%22now-1h%22,%22now%22,%22Mimir%22,%7B%22refId%22:%22A%22,%22instant%22:true,%22range%22:true,%22exemplar%22:true,%22expr%22:%22agent_build_info%7B%7D%22%7D%5D). 
 
 ![](../assets/agent_build_info.png)
 
 This example scrapes the Grafana Agent's `http://localhost:12345/metrics` endpoint and pushes those metrics to the Mimir instance.
 
+Navigate to `http://localhost:12345/graph` to view the Grafana Agent Flow UI. 
+
+![](../assets/graph.png)
+
```
karengermond, line 33:
```suggestion
The Agent displays the component pipeline in a dependency graph.  See the [Scraping component](url) and [Remote Write ](url) sections for details about the components. 
```

karengermond: Sorry, not sure yet about the hugo syntax...  I just want them to know that the details exist in the topic.

docs/sources/flow/tutorials/collecting-prometheus-metrics.md
tpaschalis: Also, the current approach creates a bunch of files and folders in the current directory.

Personal preference, but how about adding a `mkdir flow_tutorials && cd flow_tutorials` first?

docs/sources/flow/tutorials/collecting-prometheus-metrics.md
karengermond: We should probably think about how best to arrange the topics in these tutorials -- a longer term goal. I wonder if we might want to create a pattern where we use expanding text near the top of a tutorial that would be titled "Understand the configuration" -- and we'd include the explanations of the components there.  We'd probably need to work with Jack, though?

karengermond:
Some general feedback on the tutorials

- Could we add an H1 to each, with the name of the tutorial?
- The headers for the examples are inconsistent. "Collecting Prometheus metrics" has "Scraping the Agent."  "Filtering" has "Prometheus Relabel component," and "Chaining" has "Example."  I think we need to figure out how best to arrange these tutorials but in the short term, could we just use "Run the example"?  (Long term, we might want to string these into a longer form tutorial where the learner could add on capabilities as they work through it (like start with collecting, then add filtering, then chaining?)

karengermond (approved):
Just a few suggestions for the tutorial headers that are important for this release. I'm hoping to experiment with the team on new patterns for the tutorials and Agent docs in general but great stuff here for the release to get people playing!

docs/sources/flow/tutorials/collecting-prometheus-metrics.md
mattdurham: I will add the `mkdir` and `cd.` Cloning and moving to the correct directory feels like a lot of friction to me. Will give it some thought to see if there is a cleaner method.

docs/sources/flow/tutorials/collecting-prometheus-metrics.md
mattdurham: If we have expanding tags would love to use them, or spoiler style. I was going to put the answers to the tutorials in spoiler style sections but couldn't do it without custom CSS.

PR #2226, opened 2022-09-27
Introduce agentctl command to test logs

### LINKED ISSUE #1180: Add promtail command line flag -dry-run
`promtail` has a flag `-dry-run` which lets you run and see how the logs are being parsed, etc. It's very useful for debugging.

With the grafana agent, though, that's not possible.

It would be really handy to have that in the agent too.

jcreixell:
#### PR Description

Allows running promtail using `--dry-run`, which effectively redirects logs to standard output instead of sending them to Loki.

Usage:

```
./agentctl test-logs [config file]
```

Fixes #1180

#### Notes to the Reviewer

  - It exposes an extra `dryRun` param in a few places to make this possible while reusing the as much code as possible from the regular agent. IMHO it is not too intrusive.

#### PR Checklist

- [x] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated


rfratto:
Thanks Jorge! I'll give this a review tomorrow morning :)

cmd/agentctl/main.go
```diff
@@ -517,6 +522,43 @@ func templateDryRunCmd() *cobra.Command {
...
+			file := args[0]
+
+			cfg := config.Config{}
+			err := config.LoadFile(file, false, &cfg)
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "failed to validate config: %s\n", err)
+				os.Exit(1)
+			}
+
+			logger := log.NewLogfmtLogger(log.NewSyncWriter(os.Stdout))
```
tpaschalis, line 543: Do we want to reuse the logging parameters from the `config.Server` block here with the server.NewLogger helper?

If we prefer to have our own, I'd say maybe it's worth sending it over to stderr instead. Both for duplicating what the Agent is currently doing, as well as allowing a user to distinguish between the two streams and possibly redirect the dry-run logs to a file for further manipulation.

jcreixell: Thanks for the review @tpaschalis. I thought about using the server logger, but find it a bit strange to re-use the config from the server, since this is a CLI tool and there is no concept of server. I think it separating these two streams makes sense. I can switch to stderr for now.

Does anybody else have opinion on this?

jcreixell: uhm, although we are writing both to stdout and stderr in the command, so we are still mixing logs :thinking:

tpaschalis: Eh, more of a nitpicky comment, if someone is testing with a pre-existing config file, they'd probably also have the server block, the choice is yours. :)

Also, what I meant originally is that by splitting the streams we'd allow the user to redirect the stdout only to a file to verify that everything's working properly. Not sure if it's terribly useful though.

jcreixell: discussed this, since promtail uses its own logger and it is [hardcoded](https://github.com/grafana/loki/blob/main/clients/pkg/promtail/client/logger.go#L56) to use stdout, the server logger config would not affect the output of promtail. We don't seem to log much more in the agent code itself for this code path, so probably using the server logger isn't needed.

We could use the ad-hoc logger to log all output for the command itself, but that would be inconsistent with the other commands.

I will change the only place where we write to stdout in the command to stderr, and set stderr for the agent logs as well (which as I understand is consistent with how we log when running the agent in normal mode).

rfratto (approved):
Thanks Jorge! Very straightforward :shipit:

PR #2227, opened 2022-09-27
component/otelcol/exporter: create exporter component abstraction

rfratto:
This commit introduces a new package, `component/otelcol/exporter`, which exposes a generic Flow component implementation which can run OpenTelemetry Collector exporters.

There is some stuff left to do for this implementation to be complete:

* A Zap logging adapter needs to be created to correctly process logs from OpenTelemetry Collector components.

All of the above will be done in separate PRs.

As of this commit, there are no registered `otelcol.exporter.*` components. Implementations for OpenTelemetry Collector Flow components will be done in future PRs.

Related to #2213, and based off of the original prototyping done in #1843.

### DIFF
--- a/component/otelcol/consumer.go
+++ b/component/otelcol/consumer.go
@@ -1,0 +1,19 @@
+package otelcol
+
+import (
+	otelconsumer "go.opentelemetry.io/collector/consumer"
+)
+
+// Consumer is a compbined OpenTelemetry Collector consumer which can consume
+// any telemetry signal.
+type Consumer interface {
+	otelconsumer.Traces
+	otelconsumer.Metrics
+	otelconsumer.Logs
+}
+
+// ConsumerExports is a common Exports type for Flow components which are
+// otelcol processors or otelcol exporters.
+type ConsumerExports struct {
+	Input Consumer `river:"input,attr"`
+}
--- a/component/otelcol/consumer.go
+++ b/component/otelcol/consumer.go
@@ -4,7 +4,7 @@ import (
 	otelconsumer "go.opentelemetry.io/collector/consumer"
 )
 
-// Consumer is a compbined OpenTelemetry Collector consumer which can consume
+// Consumer is a combined OpenTelemetry Collector consumer which can consume
 // any telemetry signal.
 type Consumer interface {
 	otelconsumer.Traces
--- a/component/otelcol/exporter/exporter.go
+++ b/component/otelcol/exporter/exporter.go
@@ -1,0 +1,165 @@
+// Package exporter exposes utilities to create a Flow component from
+// OpenTelemetry Collector exporters.
+package exporter
+
+import (
+	"context"
+	"errors"
+	"os"
+
+	"github.com/grafana/agent/component"
+	"github.com/grafana/agent/component/otelcol"
+	"github.com/grafana/agent/component/otelcol/internal/lazyconsumer"
+	"github.com/grafana/agent/component/otelcol/internal/scheduler"
+	"github.com/grafana/agent/pkg/build"
+	otelcomponent "go.opentelemetry.io/collector/component"
+	otelconfig "go.opentelemetry.io/collector/config"
+	"go.opentelemetry.io/otel/metric/nonrecording"
+	"go.opentelemetry.io/otel/trace"
+	"go.uber.org/zap"
+)
+
+// Arguments is an extension of component.Arguments which contains necessary
+// settings for OpenTelemetry Collector exporters.
+type Arguments interface {
+	component.Arguments
+
+	// Convert converts the Arguments into an OpenTelemetry Collector exporter
+	// configuration.
+	Convert() otelconfig.Exporter
+
+	// Extensions returns the set of extensions that the configured component is
+	// allowed to use.
+	Extensions() map[otelconfig.ComponentID]otelcomponent.Extension
+
+	// Exporters returns the set of exporters that are exposed to the configured
+	// component.
+	Exporters() map[otelconfig.DataType]map[otelconfig.ComponentID]otelcomponent.Exporter
+}
+
+// Exporter is a Flow component shim which manages an OpenTelemetry Collector
+// exporter component.
+type Exporter struct {
+	ctx    context.Context
+	cancel context.CancelFunc
+
+	opts     component.Options
+	factory  otelcomponent.ExporterFactory
+	consumer *lazyconsumer.Consumer
+
+	sched *scheduler.Scheduler
+}
+
+var (
+	_ component.Component       = (*Exporter)(nil)
+	_ component.HealthComponent = (*Exporter)(nil)
+)
+
+// New creates a new Flow component which encapsulates an OpenTelemetry
+// Collector exporter. args must hold a value of the argument type registered
+// with the Flow component.
+//
+// The registered component must be registered to export the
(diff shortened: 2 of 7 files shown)


component/otelcol/exporter/exporter_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+func createTestTraces() ptrace.Traces {
+	data := ptrace.NewTraces()
+	rss := data.ResourceSpans().AppendEmpty()
+	ss := rss.ScopeSpans().AppendEmpty()
+	s := ss.Spans().AppendEmpty()
+	s.SetName("TestSpan")
+
+	return data
+}
```
rfratto, line 170: This feels a bit silly, I'll look around for a JSON decoder to maybe make creating test traces a little easier.

rfratto: There's `ptrace.NewJSONUnmarshaler()`, which might be what I'm looking for here (as long as I can figure out what the JSON format looks like)

component/otelcol/exporter/exporter.go
```diff
@@ -0,0 +1,165 @@
...
+	settings := otelcomponent.ExporterCreateSettings{
+		TelemetrySettings: otelcomponent.TelemetrySettings{
+			// TODO(rfratto): create an adapter from zap -> go-kit/log
+			Logger: zap.NewNop(),
+
+			// TODO(rfratto): expose tracing and logging statistics.
+			//
+			// We may want to put off tracing until we have native tracing
+			// instrumentation from Flow, but metrics should come sooner since we're
+			// already set up for supporting component-specific metrics.
```
rfratto, line 117: Here's an example of what we can do here once we get around to implementing this:
https://github.com/open-telemetry/opentelemetry-go/blob/example/prometheus/v0.32.1/example/prometheus/main.go#L35

It's going to require updating the OpenTeletry Collector dependency, but we're overdue for that anyway.

component/otelcol/consumer.go
```diff
@@ -0,0 +1,19 @@
+package otelcol
+
+import (
+	otelconsumer "go.opentelemetry.io/collector/consumer"
+)
+
+// Consumer is a compbined OpenTelemetry Collector consumer which can consume
```
tpaschalis, line 7:
```suggestion
// Consumer is a combined OpenTelemetry Collector consumer which can consume
```

rfratto: 😞 sorry, one day I'll figure out how to get spell checking in Vim to be reliable. Right now I usually have it off because there's just way too many false positives.

component/otelcol/internal/scheduler/scheduler.go
```diff
@@ -20,10 +20,13 @@ import (
...
-// telemetry signal, hence supporting multiple OpenTelemetry Collector
-// components inside the scheduler. Scheduler should only be used to manage
-// multiple instances of the same OpenTelemetry Collector component.
+// Each OpenTelemetry Collector component has one instance per support
+// telemetry signal, which is why Scheduler supports multiple components. For
+// example, when creating the otlpreceiver component, you would have three
+// total instances: one for logs, one for metrics, and one for traces.
+// Scheduler should only be used to manage the different signals of the same
+// OpenTelemetry Collector component; this means that otlpreceiver and
+// jaegerreceiver should not share the same Scheduler.
```
rfratto, line 29: @tpaschalis how's this?

component/otelcol/internal/scheduler/host.go
```diff
@@ -26,6 +23,18 @@ func NewHost(l log.Logger) *Host {
 
 var _ otelcomponent.Host = (*Host)(nil)
 
+// SetExtensions sets the extensions available from the Host. It is not valid
+// to call this after the Host is in use.
+func (h *Host) SetExtensions(extensions map[otelconfig.ComponentID]otelcomponent.Extension) {
```
mattdurham, line 28: Do we need any rails around this? Is it important that the `h.Extensions` not change or the contents of the map?

rfratto: Honestly it should just be a part of the constructor instead of a method. It's mainly that I don't know if changes would necessarily get picked up by implementations of otelcol components, so it probably shouldn't be changed after you pass the host to components.

WDYT? Should I just move setting the extensions and exporters to the host?

rfratto: I moved these to the constructor using the options pattern (since the type names are _very_ long). LMK what you think :)

mattdurham: I like the new way! 👍

mattdurham (approved):
LGTM

component/otelcol/internal/scheduler/scheduler.go
tpaschalis: Yeah, reads much cleaner, thanks for the explanation!

component/otelcol/consumer.go
```diff
@@ -0,0 +1,19 @@
...
+	otelconsumer "go.opentelemetry.io/collector/consumer"
+)
+
+// Consumer is a combined OpenTelemetry Collector consumer which can consume
+// any telemetry signal.
+type Consumer interface {
+	otelconsumer.Traces
+	otelconsumer.Metrics
+	otelconsumer.Logs
+}
```
tpaschalis, line 13: So the way that the OTel collector works here is that an implementation can choose to handle a signal with a consumer, or return `ErrDataTypeIsNotSupported`, right?

component/otelcol/consumer.go
```diff
@@ -0,0 +1,19 @@
...
+
+// ConsumerExports is a common Exports type for Flow components which are
+// otelcol processors or otelcol exporters.
+type ConsumerExports struct {
+	Input Consumer `river:"input,attr"`
```
tpaschalis, line 18: I don't _love_ the name Input here, but we can leave it for now. If I get this correctly, it's like what we used as a `prometheus.Receiver` right?

rfratto: yeah, we could also call it `receiver` if you think that would help (or if that might confuse people since they're not compatible, we could call it `consumer`)

tpaschalis: Let's leave it as-is for now, just wanted a quick verification.

tpaschalis (approved):
LGTM!

tpaschalis:
FYI we have a flaky test from the previous merged PR, got it on another branch as well.
```
--- FAIL: TestExporter (1.00s)
    exporter_test.go:52: 
        	Error Trace:	/drone/src/component/otelcol/exporter/exporter_test.go:52
        	            				/drone/src/component/otelcol/exporter/asm_amd64.s:1571
        	Error:      	Received unexpected error:
        	            	telemetry type is not supported
        	Test:       	TestExporter
    exporter_test.go:57: 
        	Error Trace:	/drone/src/component/otelcol/exporter/exporter_test.go:57
        	Error:      	testcomponent did not receive traces
        	Test:       	TestExporter

```

component/otelcol/consumer.go
rfratto: Right, though ErrDataTypeIsNotSupport isn't really intended to be used by consumers; we're just reusing it here for convenience. OpenTelemetry would give you a warning at load time if you're trying to send traces to a component that doesn't support it, but since we're combining everything into one consumer we have emit the errors at runtime instead.

PR #2234, opened 2022-09-28
user node_exporter for chaining

mattdurham:
Updated to use the node_exporter for a more "real" world usage when using multiple inputs. Kept using the agent metrics to keep the flow files as small as possible for the other examples.


tpaschalis (approved):
LGTM ⛵

PR #2235, opened 2022-09-28
docs/flow: replace ascii art in referencing_exports.md

tpaschalis:

#### PR Description
Replace the ASCII art with a proper diagram in config-language/expressions/referencing_exports.md 

No issue filed

#### Notes to the Reviewer
I'll make sure the inlined image works fine, not 100% sure if we should have other kind of links.

#### PR Checklist

- [ ] CHANGELOG updated (N/A)
- [X] Documentation added
- [ ] Tests updated (N/A)


rfratto (approved):
Later we should replace with an SVG but a PNG is good for now 👍

rfratto (commented):
@tpaschalis Wait sorry, shouldn't the local.file be pointing in the other direction so it represents the data flow?

tpaschalis:
@rfratto 
1) Yes, better make this an SVG, no reason to inflate the repo history forever
2) I represented how the DAG would look, that prometheus.scrape depends on the other two components, no the other way around. Do you want to keep it simpler and have them in sequential order?

rfratto:
> I represented how the DAG would look, that prometheus.scrape depends on the other two components, no the other way around. Do you want to keep it simpler and have them in sequential order?

Your call, but IMO if we're talking about the pipeline, it should be the flow of data. 
cc @mattdurham what do you think here?

karengermond:
> Later we should replace with an SVG but a PNG is good for now 👍


> 

Argh, the issue of DAG versus data flow, huh?

rfratto:
> Yes, better make this an SVG, no reason to inflate the repo history forever

Let me know if you'd like help creating the SVG, I found it strangely difficult to do

karengermond:
I wonder if we can come up with a convention for when we're showing data flow versus the DAG?  I just asked for access to Lucidchart -- maybe we can experiment later.

PR #2238, opened 2022-09-28
docs/flow: fix instances where default strings overflow page width

rfratto:
This isn't ideal, but allows the page to be navigated until we can make style changes for it to display better inside the table.


karengermond (approved):
LGTM

PR #2240, opened 2022-09-28
CI: use cache option for actions/setup-go

rfratto:
I've noticed that there's some cache misses for the CI even though they should exist. I want to see if using the builtin cache option for actions/setup-go fixes our issues.


rfratto:
I can’t tell if this is fixing anything, but subsequent runs do seem faster 🤷‍♂️ Either way it’s simpler.

mattdurham (approved):
This does indeed seem far simplier.

PR #2252, opened 2022-09-30
Rename scripts from flow to river

mattdurham:
Update the scripts from flow to river extension.


docs/sources/flow/tutorials/assets/flow_configs/README.md
```diff
@@ -0,0 +1,3 @@
+There are two copies of each file, one ending in .river and one ending in .flow. The appropriate extension is .river but
+when the tutorials were released they had .flow. After the next release the .flow files will be removed but in the meantime
+are here for compatibility.
```
tpaschalis, line 3: Are we going to backport this change into v0.28.0? Trying to understand if it makes sense to only keep one copy of each.

mattdurham: I don't think it's worth backporting. I don't want to break anyone trying it out until our next release.

karengermond (approved):
LGTM

docs/sources/flow/tutorials/assets/flow_configs/README.md
rfratto: These are just docs changes, we don't have to wait for a release to do them and we can just cherry pick this directly into the release branch.

Though we'd have to also edit the blog post to change the extension, but that's not a big deal.

mattdurham: My main concern is removing the `.flow` is breaking someone playing with it at the moment, after 6 weeks of using the new extension I have less of a concern.

PR #2254, opened 2022-09-30
component/otelcol/receiver: create receiver component abstraction

rfratto:
This commit introduces a new package, component/otelcol/receiver, which exposes a generic Flow component implementation which can run OpenTelemetry Collector receiver.

Like #2227, it leaves some work unfinished for future PRs:

* A Zap logging adapter needs to be created to correctly process logs from OpenTelemetry Collector components.

As of this commit, there are no registered `otelcol.receiver.*` components. Implementations for OpenTelemetry Collector Flow components will be done in future PRs.

Related to #2213.

### DIFF
--- a/component/otelcol/consumer.go
+++ b/component/otelcol/consumer.go
@@ -12,6 +12,17 @@ type Consumer interface {
 	otelconsumer.Logs
 }
 
+// ConsumerArguments is a common Arguments type for Flow components which can
+// send data to otelcol consumers.
+//
+// It is expected to use ConsumerArguments as a block within the top-level
+// arguments block for a component.
+type ConsumerArguments struct {
+	Metrics []Consumer `river:"metrics,attr,optional"`
+	Logs    []Consumer `river:"logs,attr,optional"`
+	Traces  []Consumer `river:"traces,attr,optional"`
+}
+
 // ConsumerExports is a common Exports type for Flow components which are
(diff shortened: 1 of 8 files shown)


tpaschalis (approved):
Looks clean!

component/otelcol/consumer.go
```diff
@@ -12,6 +12,17 @@ type Consumer interface {
...
 
+// ConsumerArguments is a common Arguments type for Flow components which can
+// send data to otelcol consumers.
+//
+// It is expected to use ConsumerArguments as a block within the top-level
+// arguments block for a component.
+type ConsumerArguments struct {
+	Metrics []Consumer `river:"metrics,attr,optional"`
+	Logs    []Consumer `river:"logs,attr,optional"`
+	Traces  []Consumer `river:"traces,attr,optional"`
```
tpaschalis, line 23: We'll always use the `fanoutconsumer` package implementations for these internally, right?

rfratto: No, fanoutconsumer is only used when passing the list of consumers to the OpenTelemetry code. These will instead almost always be conusmers from the `lazyconsumer` package (unless one day there's different components which are otelcol-compatible that have their own approach for exposing consumers)

mattdurham (approved):
LGTM

PR #2257, opened 2022-10-03
CI: remove stale bot

rfratto:
We are exploring a reworking of our project management process. As part of this process, the stale bot is being removed in favor of directly interacting with issues that are not planned to be worked on. 

We may revert this decision in the future if the new process doesn't work out.


mattdurham (approved):
LGTM

PR #2323, opened 2022-10-03
component/otelcol: use zap adapter to accept logs from wrapped components

rfratto:
This introduces component/otelcol/internal/zapadapter, which creates a *zap.Logger instance from a github.com/go-kit/log.Logger instance. This is then used when creating OpenTelemetry Collector components, allowing us to continue to use github.com/go-kit/log consistently throughout Flow.

Related to #2213.

### DIFF
--- a/component/otelcol/internal/zapadapter/zapadapter.go
+++ b/component/otelcol/internal/zapadapter/zapadapter.go
@@ -1,0 +1,240 @@
+// Package zapadapter allows github.com/go-kit/log to be used as a Zap core.
+package zapadapter
+
+import (
+	"fmt"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+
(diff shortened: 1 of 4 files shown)


component/otelcol/internal/zapadapter/zapadapter.go
```diff
@@ -0,0 +1,240 @@
...
+	fe.fields = append(fe.fields, fe.keyName(key), "<array>")
+	return nil
+}
+
+func (fe *fieldEncoder) AddObject(key string, marshaler zapcore.ObjectMarshaler) error {
+	// TODO(rfratto): allow this to write the value of the object instead of
+	// placeholder text.
+	fe.fields = append(fe.fields, fe.keyName(key), "<object>")
+	return nil
+}
```
rfratto, line 126: I didn't implement these here because:

1. I think they're going to be a massive pain.
2. I suspect logging arrays and objects is going to be fairly uncommon.

mattdurham (approved):
Some odd underlying things with Zap but overall LHTM.

mattdurham (approved):
LGTM

Issue #2286, opened 2022-10-03
Flow: otelcol.receiver.otlp component


rfratto:
Expose the OpenTelemtry Collector `otlp` receiver as a Flow component called `otelcol.receiver.otlp`. This component is from the `otelcol` distribution.

PR #2326, opened 2022-10-04
Flow: add otelcol.receiver.otlp component

rfratto:
Introduce an otelcol.receiver.otlp component which can receive OTLP data over gRPC or HTTP. There is more code here than would normally be necessary since I needed to introduce common types otelcol receiver components will share. 

This is a nearly 1:1 match to the upstream component with one exception: Custom authentication methods are not yet supported.

Future work will be needed to figure out how we want to expose authentication to components. The way it currently works in OpenTelemetry is by exposing an Authentication extension to components. We will likely need to do something different. Work on this will be done in a future PR.

Closes grafana/agent#2286.

### DIFF
--- a/component/all/all.go
+++ b/component/all/all.go
@@ -5,6 +5,7 @@ import (
 	_ "github.com/grafana/agent/component/discovery/kubernetes"                 // Import discovery.kubernetes
 	_ "github.com/grafana/agent/component/discovery/relabel"                    // Import discovery.relabel
 	_ "github.com/grafana/agent/component/local/file"                           // Import local.file
+	_ "github.com/grafana/agent/component/otelcol/receiver/otlp"                // Import otelcol.receiver.otlp
 	_ "github.com/grafana/agent/component/prometheus/integration/node_exporter" // Import prometheus.integration.node_exporter
 	_ "github.com/grafana/agent/component/prometheus/relabel"                   // Import prometheus.relabel
 	_ "github.com/grafana/agent/component/prometheus/remotewrite"               // Import prometheus.remote_write


mattdurham:
I want to give this a closer look than I can at the moment but will give a deeper review tomorrow morning.

docs/sources/flow/reference/components/otelcol.receiver.otlp.md
```diff
@@ -0,0 +1,183 @@
+---
+aliases:
+- /docs/agent/latest/flow/reference/components/otelcol.receiver.otlp
+title: otelcol.receiver.otlp
+---
+
+# otelcol.receiver.otlp
```
rfratto, line 7: I want to be able to give an example pipeline at the bottom of the page, but I can't at the moment, not until there's at least one exporter component.

tpaschalis (approved):
LGTM!

component/otelcol/config_grpc.go
```diff
@@ -0,0 +1,122 @@
...
+type GRPCServerArguments struct {
+	Endpoint  string `river:"endpoint,attr,optional"`
+	Transport string `river:"transport,attr,optional"`
+
+	TLS *TLSServerArguments `river:"tls,block,optional"`
+
+	MaxRecvMsgSizeMiB    uint64 `river:"max_recv_msg_size_mib,attr,optional"`
+	MaxConcurrentStreams uint32 `river:"max_concurrent_streams,attr,optional"`
+	ReadBufferSize       int    `river:"read_buffer_size,attr,optional"`
+	WriteBufferSize      int    `river:"write_buffer_size,attr,optional"`
```
rfratto: After a discussion offline, it'd be nice to have these be River strings which describe the unit (`"500B"`).

Ditto for other config types.

karengermond (approved):
LGTM!

PR #2338, opened 2022-10-10
Add support bundles functionality

### LINKED ISSUE #2381: Enable exporting of 'support bundles'
Many times, when trying to debug the Grafana Agent, we take specific steps.

* Try to see what version, Platform/OS and mode it has been deployed with
* Check the current configuration file
* Look at debug-level logs
* Look at the agent's internal metrics endpoint
* Use the API to check on the active metrics/logs subsystem instances and discovered targets
* Check if there's any experimental features in use (eg. dynamic-config or integrations-next)
* Export and examine pprof profiles 

We could add a new endpoint (or agentctl command?) that gathers all this information and packages into a single zip file, so that it can provide a baseline of information when trying to debug an issue or be easily shared between teams.

The 'bundle' should contain all information in plain text so that it can be inspected before sharing, and we should also make sure that it it not possible to leak any sensitive information.

tpaschalis:

#### PR Description
This PR is the first part of adding "support bundles" functionality to the Agent

Closes #2381.

#### Notes to the Reviewer
I'm mostly looking for feedback on the API exposed by the `supportbundle` package,

There are still a few things to be done; you can let me know if you prefer them to be in this PR or a follow-up:
1) Add unit test for single-flight request approach
2) Wire this API to the Flow code path
3) Add support for integrations (intercepting logs, reporting usage)
~4) Check whether it's we can temporarily hijack all log levels to debug~
~5) Add CLI flag for disabling this functionality eg. `--disable-debug-endpoints`~

This provides another way of fetching the configuration, circumventing the `/-/config` endpoint, for which we had a security advisory and need a dedicated flag to enable. Did anything change on that front, are do we know if we're covered? I'll check, but meant to ask if you knew off the top of your head.

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [ ] Tests updated


cmd/agent/entrypoint.go
```diff
@@ -249,6 +252,74 @@ func (ep *Entrypoint) getReporterMetrics() map[string]interface{} {
...
+	}()
+
+	// Hijack subsystems loggers.
+	ep.log.SetLogger(logfmtLogger)
+	ep.promMetrics.SetLogger(logfmtLogger)
+	ep.lokiLogs.SetLogger(logfmtLogger)
+	ep.tempoTraces.SetLogger(zapLogger)
+
+	var g singleflight.Group
+	_, err, _ := g.Do("/-/support", func() (interface{}, error) {
```
rfratto, line 308: I don't think this works; it looks like you're creating a new `singleflight.Group` each time the handler is called.

Personally, I'd just have a mutex and lock it for the duration of the request.

cmd/agent/entrypoint.go
```diff
@@ -249,6 +252,74 @@ func (ep *Entrypoint) getReporterMetrics() map[string]interface{} {
...
+		ep.promMetrics.SetLogger(metricsLogger)
+		ep.lokiLogs.SetLogger(logsLogger)
+		ep.tempoTraces.SetLogger(tracesLogger)
+	}()
+
+	// Hijack subsystems loggers.
+	ep.log.SetLogger(logfmtLogger)
+	ep.promMetrics.SetLogger(logfmtLogger)
+	ep.lokiLogs.SetLogger(logfmtLogger)
+	ep.tempoTraces.SetLogger(zapLogger)
```
rfratto, line 305: nit: I think it's less to keep track of if we have a persistent middleware-y logger so we don't have to maintain a list of loggers which need to be wrapped and don't have to change out pointers at runtime. 

Example: This isn't updating the logger for the integrations subsystem, which you'd get for free with the middleware approach.

rfratto (commented):
Looking great!

cmd/agent/entrypoint.go
tpaschalis: I played around with this and leads to a much cleaner solution. Implementing this and will come back for comments.

pkg/supportbundle/supportbundle.go
```diff
@@ -0,0 +1,310 @@
...
+		Architecture:    runtime.GOARCH,
+		Uptime:          ut.Seconds(),
+		EnabledFeatures: enabledFeatures,
+	}
+	meta, err := yaml.Marshal(m)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal support bundle metadata: %s", err)
+	}
+	// Gather current configuration.
+	config, err := yaml.Marshal(cfg)
```
marctc, line 71: Is this already removing the secrets from the config?

tpaschalis: Supposedly, yes, but I'll have to double check before enabling this endpoint.

pkg/supportbundle/supportbundle.go
```diff
@@ -0,0 +1,310 @@
...
+	// Gather Agent's own metrics.
+	resp, err := http.DefaultClient.Get("http://" + srvAddress + "/metrics")
+	if err != nil {
+		return nil, fmt.Errorf("failed to get internal Agent metrics: %s", err)
+	}
+	agentMetrics, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read internal Agent metrics: %s", err)
+	}
+
```
marctc, line 85: What about metrics from integrations? Or too much data?

pkg/supportbundle/supportbundle.go
```diff
@@ -0,0 +1,310 @@
...
+type Metadata struct {
+	BuildVersion    string   `yaml:"build_version"`
+	OS              string   `yaml:"os"`
+	Architecture    string   `yaml:"architecture"`
+	Uptime          float64  `yaml:"uptime"`
+	EnabledFeatures []string `yaml:"enabled_features"`
+}
+
+// Export gathers the information required for the support bundle.
+func Export(enabledFeatures []string, cfg config.Config, srvAddress string, duration float64) (*Bundle, error) {
```
marctc, line 53: Would it make sense to split this function a bit?

pkg/supportbundle/supportbundle.go
tpaschalis: I initially left out the integrations as they required different handling.

With the current approach of using the middleware-y logger, we get (at least some) of the integrations' logs for free, so I'll check out how to get their metrics, too.

pkg/traces/traces.go
```diff
@@ -134,11 +136,17 @@ func newLogger(zapLevel zapcore.LevelEnabler, fmt logging.Format) *zap.Logger {
...
+
+	logger := zap.New(
+		zapcore.NewTee(
+			zapcore.NewCore(
+				encoder,
+				os.Stdout,
+				zapLevel,
+			),
+			zapAdapter.Core()),
+		zap.AddCaller())
```
tpaschalis, line 149: This keep the same definition of the previous logger, but tees it off to the zapadapter where we passed HookedLogger in.

pkg/supportbundle/supportbundle.go
```diff
@@ -0,0 +1,310 @@
...
+			return nil, err
+		}
+	}
+	runtime.SetBlockProfileRate(1)
+	if p := pprof.Lookup("block"); p != nil {
+		if err := p.WriteTo(&blockBuf, 0); err != nil {
+			return nil, err
+		}
+	}
+	runtime.SetBlockProfileRate(0)
```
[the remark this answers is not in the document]
tpaschalis: The block profiler's rate is 0 by default (disabled), so it wouldn't make any difference. But you're right, we have to read and restore the mutex profiler's setting to its previous value.

pkg/supportbundle/supportbundle.go
```diff
@@ -0,0 +1,310 @@
...
+
+	// TODO(@tpaschalis) Figure out how to better correlate CPU profile
+	// duration to server timeout settings. Also, ideally a CPU profile should
+	// include at least one scrape or some log collection, but we can't
+	// guarantee that.
+	err = pprof.StartCPUProfile(&cpuBuf)
+	if err != nil {
+		return nil, err
+	}
+	time.Sleep(time.Duration(duration-1) * time.Second)
```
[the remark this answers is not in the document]
tpaschalis: Like starting a context.WithTimeout and then using it to wait until `ctx.Deadline()`?

pkg/supportbundle/supportbundle.go
tpaschalis: Sorry for not responding for so long, I refactored some things about the PR, but now getting to the supportbundle package itself. What do you have in mind?

pkg/supportbundle/supportbundle.go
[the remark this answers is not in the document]
rfratto: Yes 👍

docs/sources/api/_index.md
```diff
@@ -363,6 +363,35 @@ configuration file on disk.
...
+```
+
+This endpoint returns a 'support bundle', a zip file that contains information
+about a running agent, and can be used as a baseline of information when trying
+to debug an issue.
+
+The duration parameter is optional, must be less or equal than the configured
+HTTP server write timeout, and if no provided, defaults to it.
+The endpoint is only exposed to the agent's HTTP server listen address, which
+defaults to `localhost:12345`.
```
karengermond, line 378:
```suggestion
The duration parameter is optional, must be less than or equal to the configured
HTTP server write timeout, and if not provided, defaults to it.
The endpoint is only exposed to the agent's HTTP server listen address, which
defaults to `localhost:12345`.
```

docs/sources/api/_index.md
```diff
@@ -363,6 +363,35 @@ configuration file on disk.
...
+
+The support bundle contains all information in plain text, so that it can be
+inspected before sharing, to verify that no sensitive information has leaked.
+
+In addition, you can inspect the [supportbundle package](https://github.com/grafana/agent/tree/main/pkg/supportbundle)
+to verify the code that is being used to generate these bundles.
+
+A support bundle contains the following data
```
karengermond, line 386:
```suggestion
A support bundle contains the following data:
```

docs/sources/api/_index.md
```diff
@@ -363,6 +363,35 @@ configuration file on disk.
...
+* `agent-config.yaml` contains the current agent configuration.
+* `agent-logs.txt` contains the agent logs during the bundle generation.
+* `agent-metadata.yaml` contains the agent's build version, operating system, architecture, uptime, plus a string payload defining which extra agent features have been enabled via command-line flags.
+* `agent-metrics-instances.json` and `agent-metrics-targets.json` contain the active metric subsystem instances, and the discovered scraped targets for each one.
+* `agent-logs-instances.json` contains the active logs subsystem instances.
+* `agent-metrics.txt` contains a snapshot of the agent's internal metrics.
+* the `pprof/` directory contains Go runtime profiling data (CPU, heap, goroutine, mutex, block profiles) as exported by the pprof package.
```
karengermond, line 393:
```suggestion
* The `pprof/` directory contains Go runtime profiling data (CPU, heap, goroutine, mutex, block profiles) as exported by the pprof package.
```

docs/sources/configuration/flags.md
```diff
@@ -57,6 +57,25 @@ This list may change over time. All newly reported data will also be documented
 If you would like to disable the reporting, Grafana Agent provides the flag `-disable-reporting`
 to stop the reporting.
 
+## Support Bundles
+Grafana Agent allows exporting 'Support Bundles' on the `/-/support` endpoint.
+These are zip files, that contain commonly-used information to provide a
+baseline when starting to debug an issue with the Agent.
```
karengermond, line 63:
```suggestion
Grafana Agent allows exporting of 'support bundles' on the `/-/support` endpoint.
Support bundles are zip files containing commonly-used information that provide a
baseline for debugging an issue with the Agent.
```

docs/sources/configuration/flags.md
```diff
@@ -57,6 +57,25 @@ This list may change over time. All newly reported data will also be documented
 If you would like to disable the reporting, Grafana Agent provides the flag `-disable-reporting`
 to stop the reporting.
 
+## Support Bundles
```
karengermond, line 60:
```suggestion
## Support bundles
```

docs/sources/configuration/flags.md
```diff
@@ -57,6 +57,25 @@ This list may change over time. All newly reported data will also be documented
...
+Grafana Agent allows exporting 'Support Bundles' on the `/-/support` endpoint.
+These are zip files, that contain commonly-used information to provide a
+baseline when starting to debug an issue with the Agent.
+
+The support bundle contains all information in plain text, so that it can be
```
karengermond, line 65:
```suggestion
Support bundles contain all information in plain text, so that they can be
```

docs/sources/configuration/flags.md
```diff
@@ -57,6 +57,25 @@ This list may change over time. All newly reported data will also be documented
...
+inspected before sharing, to verify that no sensitive information has leaked.
```
karengermond, line 66:
```suggestion
inspected before sharing to verify that no sensitive information has leaked.
```

docs/sources/configuration/flags.md
```diff
@@ -57,6 +57,25 @@ This list may change over time. All newly reported data will also be documented
...
+The bundle's structure is:
```
karengermond, line 67:
```suggestion
Support bundles contain the following data:
```

docs/sources/configuration/flags.md
```diff
@@ -57,6 +57,25 @@ This list may change over time. All newly reported data will also be documented
...
+* `agent-config.yaml` contains the current agent configuration.
+* `agent-logs.txt` contains the agent logs during the bundle generation.
+* `agent-metadata.yaml` contains the agent's build version, operating system, architecture, uptime, plus a string payload defining which extra agent features have been enabled via command-line flags.
+* `agent-metrics-instances.json` and `agent-metrics-targets.json` contain the active metric subsystem instances, and the discovered scraped targets for each one.
+* `agent-logs-instances.json` contains the active logs subsystem instances.
+* `agent-metrics.txt` contains a snapshot of the agent's internal metrics.
+* the `pprof/` directory contains Go runtime profiling data (CPU, heap, goroutine, mutex, block profiles) as exported by the pprof package.
```
karengermond, line 74:
```suggestion
* The `pprof/` directory contains Go runtime profiling data (CPU, heap, goroutine, mutex, block profiles) as exported by the pprof package.
```

karengermond (approved):
LGTM -- minor suggestions

cmd/agent/entrypoint.go
```diff
@@ -249,6 +257,68 @@ func (ep *Entrypoint) getReporterMetrics() map[string]interface{} {
...
 }
 
+func (ep *Entrypoint) supportHandler(rw http.ResponseWriter, r *http.Request) {
+	ep.mut.Lock()
+	cfg := ep.cfg
+	ep.mut.Unlock()
+
+	if cfg.DisableSupportBundle {
+		rw.WriteHeader(http.StatusNotFound)
+		_, _ = rw.Write([]byte("404 - support bundle generation is disabled"))
```
mattdurham, line 267: Since we are adding context, would be nice to return the command to enable it in the message.

pkg/supportbundle/supportbundle.go
```diff
@@ -0,0 +1,236 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get internal Agent metrics: %s", err)
+	}
+	agentMetrics, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read internal Agent metrics: %s", err)
+	}
+
+	// Collect the Agent metrics instances and target statuses.
+	resp, err = httpClient.Get("http://" + srvAddress + "/agent/api/v1/metrics/instances")
```
mattdurham, line 94: Do we need to check the tls flag and switch to HTTP is TLS is enabled?

pkg/supportbundle/supportbundle.go
```diff
@@ -0,0 +1,236 @@
...
+		goroutineBuf bytes.Buffer
+		blockBuf     bytes.Buffer
+		mutexBuf     bytes.Buffer
+	)
+	err = pprof.StartCPUProfile(&cpuBuf)
+	if err != nil {
+		return nil, err
+	}
+	deadline, _ := ctx.Deadline()
+	time.Sleep(time.Until(deadline) - 100*time.Millisecond)
```
mattdurham, line 144: This looks like a magic value. Do we want to document why 100ms?

pkg/traces/traces.go
rfratto, line 149: Is this necessary? It feels like we can use the following here to simplify:

```go
logger := zap.New(zapadapter.New(l))
```

(This is what we do for the otelcol Flow components)

pkg/supportbundle/supportbundle.go
```diff
@@ -0,0 +1,236 @@
...
+type Metadata struct {
+	BuildVersion string                 `yaml:"build_version"`
+	OS           string                 `yaml:"os"`
+	Architecture string                 `yaml:"architecture"`
+	Uptime       float64                `yaml:"uptime"`
+	Payload      map[string]interface{} `yaml:"payload"`
+}
+
+// Export gathers the information required for the support bundle.
+func Export(ctx context.Context, enabledFeatures []string, cfg config.Config, srvAddress string, dialContext server.DialContextFunc) (*Bundle, error) {
```
rfratto, line 48: nit: just to play it safe, we should probably move the mutex here as as a global to enforce there's no way to call `Export` more than once concurrently, rather than leaving that up to the caller.

pkg/supportbundle/supportbundle.go
rfratto: (I'm pretty sure this is using the in-memory HTTP client so TLS isn't required here)

cmd/agent/entrypoint.go
rfratto, line 267: Is 404 appropriate for this? The endpoint was valid, just denied. Another 4xx code (like `405 Method Not Allowed`) might be appropriate here.

cmd/agent/entrypoint.go
```diff
@@ -249,6 +257,68 @@ func (ep *Entrypoint) getReporterMetrics() map[string]interface{} {
...
+		}
+		if durationInt < 1 {
+			http.Error(rw, "duration value (in seconds) should be larger than 1", http.StatusBadRequest)
+			return
+		}
+		duration = float64(durationInt)
+		if duration > ep.srv.HTTPServer.WriteTimeout.Seconds() {
+			http.Error(rw, "duration value exceeds the server's write timeout", http.StatusBadRequest)
+			return
+		}
```
rfratto, line 288: pprof has an interesting way of doing this:

https://cs.opensource.google/go/go/+/refs/tags/go1.19.2:src/net/http/pprof/pprof.go;l=104-107

Maybe this is fine as-is, but getting the server from the context _might_ be more safe just in case `ep.srv.HTTPServer.WriteTimeout` gets mutated at read time.

cmd/agent/entrypoint.go
```diff
@@ -249,6 +257,68 @@ func (ep *Entrypoint) getReporterMetrics() map[string]interface{} {
...
+	// TODO(@tpaschalis) Can we also dynamically up the log_level while the
+	// support bundle is being generated?
+	var logsBuffer bytes.Buffer
+	logger := log.NewSyncLogger(log.NewLogfmtLogger(&logsBuffer))
+	defer func() {
+		ep.log.HookLogger.Enabled = false
+		ep.log.HookLogger.Logger = nil
+	}()
+	ep.log.HookLogger.Enabled = true
+	ep.log.HookLogger.Logger = logger
```
rfratto, line 309: This isn't goroutine safe; Enabled/Logger might be read by other goroutines (from different subsytems) at the same time this gets modified. 

You should add a method to the HookLogger to set these which needs a mutex (see next comment).

pkg/server/logger.go
```diff
@@ -81,9 +90,21 @@ func makeDefaultLogger(lvl logging.Level, fmt logging.Format) (log.Logger, error
...
+	}
 	return l.l.Log(kvps...)
 }
 
+// Log implements log.Logger.
+func (hl *HookedLogger) Log(kvps ...interface{}) error {
+	if hl.Enabled {
+		return hl.Logger.Log(kvps...)
+	}
+	return nil
```
rfratto, line 105: This needs a mutex to prevent concurrent read/writes in the middle of a support bundle generation while subsystems are actively logging.

cmd/agent/entrypoint.go
```diff
@@ -249,6 +257,68 @@ func (ep *Entrypoint) getReporterMetrics() map[string]interface{} {
...
+
+	ep.mut.Lock()
+	var (
+		enabledFeatures = ep.cfg.EnabledFeatures
+		httpSrvAddress  = ep.cfg.ServerFlags.HTTP.InMemoryAddr
+	)
+	ep.mut.Unlock()
+
+	// TODO(@tpaschalis) Can we also dynamically up the log_level while the
+	// support bundle is being generated?
```
rfratto, line 301: As long as the hook logger executes _before_ the level filterer, you will always get debug level logs here.

CHANGELOG.md
```diff
@@ -40,6 +40,10 @@ collected logs to standard output. This can be useful for debugging. (@jcreixell
   - `otelcol.receiver.jeager` receives Jaeger-formatted traces. Data can then
     be forwarded to other `otelcol` components. (@rfratto)
 
+- Introduce `/-/support` endpoint for generating 'support bundles'; zip files
+  of commonly-requested information that can be used to debug a running agent.
+  (@tpaschalis)
```
rfratto, line 45: Should we make this clear this is for the non-Flow ("static") agent?

rfratto (approved):
Nice work!

cmd/agent/entrypoint.go
tpaschalis: We do the same for when we the /-/config endpoint is disabled, but you're right. I feel like 403 Forbidden might make sense here, WDYT?

tpaschalis: The support bundles are enabled by default, and disabled via the `-disable-support-bundles` flag. Do you think we should remind the user why it's disabled, or is it okay?

cmd/agent/entrypoint.go
tpaschalis: Just saw this, was positively surprised!

CHANGELOG.md
tpaschalis: Yeah, if you're also ok with wiring this to Flow in a different PR. I'll update the changelog to reflect that.

pkg/traces/traces.go
tpaschalis: I believe we want to keep the previous definition of the traces logger as-is (utilising the customer encoder and level, plus the caller annotation), and just tee off to our own. WDYT?

cmd/agent/entrypoint.go
mattdurham: I would still probably add the flag name so they if they find they want to turn it off they dont have to dig example `and support bundle generation can be enabled by removing the --disable-support-bundles` flag.

pkg/traces/traces.go
rfratto: I personally would prefer removing the old encoder and force every log line to go through the same log.Logger instance that all our subsystems use. 

One of the big issues with _not_ using the log.Logger instance here is that changing the logging level at runtime doesn't change the verbosity of traces, since it only reads the desired logging level once. It also goes straight to os.Stdout, which means the Windows Event logs never include anything for tracing. 

Using zadapater exclusively will help us fix these issues IMO.

pkg/supportbundle/supportbundle.go
rfratto: (It might make sense to have two mutexes; one in main and the other here. Whatever works; but I think this should still be protected with a mutex somehow)

cmd/agent/entrypoint.go
rfratto: Yeah, 403 seems better than 405 👍

CHANGELOG.md
rfratto: > Yeah, if you're also ok with wiring this to Flow in a different PR. I'll update the changelog to reflect that.

Yeah, it's ok to put Flow in another PR 👍 (though if we close out the original issue we'll want to create a new issue for Flow support)

cmd/agent/entrypoint.go
tpaschalis: Sounds good, done!

pkg/supportbundle/supportbundle.go
tpaschalis: Sorry for not responding to this before.

The idea was to spend _most_ of the given `duration` to let the profiles gather some data, and then give it a small amount of time to write those profiles and the result of the bundle.

Do you have any better way in mind to do this? Otherwise, I'll just add a comment as you suggested.

pkg/traces/traces.go
tpaschalis: This is done in https://github.com/grafana/agent/pull/2338/commits/3ebecda9fdddf48d6af8874148b4ec78dcdec261.

We can now use the server block to reload the logging level and format for traces, too.

pkg/config/config.go
```diff
@@ -54,6 +54,7 @@ var DefaultConfig = Config{
 	ServerFlags:           server.DefaultFlags,
 	Metrics:               metrics.DefaultConfig,
 	Integrations:          DefaultVersionedIntegrations,
+	DisableSupportBundle:  false,
```
mattdurham, line 57: One thing I dislike and its probably a broader conversation but that we use both enable* and disable* flag names. Makes it harder to grasp which means which.

pkg/config/config.go
```diff
@@ -79,6 +80,9 @@ type Config struct {
 	// Toggle for config endpoint(s)
 	EnableConfigEndpoints bool `yaml:"-"`
 
+	// Toggle for support bundle generation.
+	DisableSupportBundle bool `yaml:"-"`
```
marctc, line 84: Would it make sense to add it to operator config as well?

pkg/supportbundle/supportbundle.go
tpaschalis: We went ahead with the approach of not exposing the configuration, unless the -config.enable-read-api flag is passed, so as to not provide a way to sidestep it.

We should revisit them both soon-ish, as it will make support bundles more useful.

PR #2341, opened 2022-10-11
Docs: reuse flow config blocks

rfratto:
This PR headless pages to reuse Flow config blocks.

To view locally, run `make docs` from docs/ and look at an example page using the new shared snippet: http://localhost:3002/docs/agent/latest/flow/reference/components/otelcol.processor.batch/#output-block

### DIFF
--- a/docs/sources/flow/reference/components/otelcol.processor.batch.md
+++ b/docs/sources/flow/reference/components/otelcol.processor.batch.md
@@ -72,20 +72,7 @@ output | [output][] | Configures where to send received telemetry data. | yes
 
 ### output block
 
-The `output` block configures a set of components to send batched telemetry
-data to.
-
-The following arguments are supported:
-
-Name | Type | Description | Default | Required
----- | ---- | ----------- | ------- | --------
-`metrics` | `list(otelcol.Consumer)` | List of consumers to send metrics to. | `[]` | no
-`logs` | `list(otelcol.Consumer)` | List of consumers to send logs to. | `[]` | no
-`traces` | `list(otelcol.Consumer)` | List of consumers to send traces to. | `[]` | no
-
-The `output` block must be specified, but all of its arguments are optional. By
-default, telemetry data is dropped. To send telemetry data to other components,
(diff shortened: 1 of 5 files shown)


rfratto:
This isn't working at the moment; the shortcode is replaced with blank content.

rfratto:
> I believe the problem is that the docs/sources/shared/ directory is missing an index.md file that should probably itself be headless: true. The shortcode relies on using the "resources" of the shared page bundle to find children pages and include the content. With the absence of the index.md file, the lookup is failing silently and injecting empty content.

@jdbaldry Thanks! I added docs/sources/shared/index.md, but it still doesn't seem to be working 🤔

nate-walters-grafana:
@jdbaldry @rfratto Does it work if the "index.md" file name is prepended with an underscore: _index.md?

rfratto:
> @jdbaldry @rfratto Does it work if the "index.md" file name is prepended with an underscore: _index.md? Index files in Hugo usually need the "_".

@nate-walters-grafana No, I just tried `_index.md` and it doesn't work either (also I see other projects which use `docs/sources/index.md` without the underscore)

jdbaldry:
Quoting myself from the open PR in a private repository:
> I believe it must be `index.md` as we rely on the child files being recognized as page resources of the shared page bundle. I'm still getting back into the swing of things and Hugo regularly confuses me so forgive me if I'm getting this muddled up.

nate-walters-grafana:
@jdbaldry is right. It _should_ be index.md (without the underscore). Apologies. I wasn't fully understanding how this was meant to work.

jdbaldry:
An updated `grafana/docs-base:latest` image has been published that should allow you to start reusing content. If you encounter any problems, the previous image is available as `grafana/docs-base:4b6e372e035b7561e110e1025fbafc05d58fc10c`.

rfratto:
Thanks @jdbaldry, it works now! 🎉 

@karengermond are you comfortable with us merging this? It doesn't deduplicate _everything,_ but should provide a base for others to build on.

karengermond (approved):
LGTM!

PR #2351, opened 2022-10-12
Duplicate linux install script, remove fetching cloud config.

cristiangreco:
#### PR Description
The cloud functionality to generate the agent config file is soon to be removed, hence here we duplicate the linux install script so that both versions can be used while the cloud change is being worked on.

The new script is identical to the old one, except for `grafana-agentctl cloud-config` being removed (together with env vars for auth).

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/production/grafanacloud-install-linux.sh
+++ b/production/grafanacloud-install-linux.sh
@@ -1,0 +1,115 @@
+#!/usr/bin/env sh
+# shellcheck shell=dash
+# This script should run in all POSIX environments and Dash is POSIX compliant.
+
+# grafanacloud-install-linux.sh installs the Grafana Agent on supported
+# Linux systems for Grafana Cloud users. Those who aren't users of Grafana Cloud
+# or need to install the Agent on a different architecture or platform should
+# try another installation method.
+#
+# grafanacloud-install.sh has a hard dependency on being run on a supported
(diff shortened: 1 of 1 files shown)


rfratto:
@cristiangreco Rather than hosting all the install scripts in the public repository, can they be moved to a GCS bucket? I'm not sure Grafana Cloud should depend on a git repo for installation scripts.

cristiangreco:
@rfratto do you think the win and linux scripts should be removed from this repo? I'm ok with moving maintenance to the connections-exp team.

@matthewnolf ?

rfratto:
> @rfratto do you think the win and linux scripts should be removed from this repo? I'm ok with moving maintenance to the connections-exp team.

Yeah, I generally think removing the scripts is a good idea. It'll give you more control and not be beholden to our release process.

matthewnolf:
> > @rfratto do you think the win and linux scripts should be removed from this repo? I'm ok with moving maintenance to the connections-exp team.
> 
> Yeah, I generally think removing the scripts is a good idea. It'll give you more control and not be beholden to our release process.

Yep agreed, makes sense to move the ownership

PR #2284, opened 2022-10-03
component/otelcol/processor: create processor component abstraction

rfratto:
This commit introduces a new package, component/otelcol/processor, which exposes a generic Flow component implementation which can run OpenTelemetry Collector processors.

This package acts like a combination of #2227 and #2254, both exposing consumers for other consumers to send telemetry data to be processed, and accepting a list of consumers to determine where to forward processed data. 

Like #2227 and #2254, it leaves some work unfinished for future PRs:

* Component-specific metrics are currently ignored.
* Component-specific traces are currently ignored.

As of this commit, there are no registered `otelcol.processor.*` components. Implementations for OpenTelemetry Collector Flow components will be done in future PRs.

Related to #2213.

PR #2352, opened 2022-10-12
component/otelcol/auth: create auth extension component abstraction

rfratto:
OpenTelemetry Collector supports many extensions. Extensions are used as a generic way to put "everything else" into OpenTelemetry. To quote their [documentation](https://pkg.go.dev/go.opentelemetry.io/collector@v0.61.0/component#Extension):

> Extension is the interface for objects hosted by the OpenTelemetry Collector that don't participate directly on data pipelines but provide some functionality to the service, examples: health check endpoint, z-pages, etc.

There are two types of extensions relevant to us:

* [Authentication extensions][auth-ext]: used for both client and server authentication.
* [Storage extensions][storage-ext]: used for external storage of state.

Other extensions, such as [awsproxy][] are useful but better suited as generic Flow components rather than being shoved in the otelcol namespace, since they are unrelated to telemetry pipelines and aren't referenced by other otelcol components in the upstream configuration.

This commit introduces a new package, component/otelcol/auth, which exposes a generic Flow component implementation which can run OpenTelemetry Collector extensions meant for authentication.

While storage extensions may end up being Flow components eventually, it's currently marked as experimental upstream. We will reevaluate storage extension components once things have stabilized a little more.

Like #2227, #2254, and #2284, it leaves some work unfinished for future PRs:

As of this commit, there are no registered `otelcol.auth.*` components. Implementations for OpenTelemetry Collector Flow components will be done in future PRs.

Related to #2213.

[auth-ext]: https://pkg.go.dev/go.opentelemetry.io/collector@v0.61.0/config/configauth
[storage-ext]: https://pkg.go.dev/go.opentelemetry.io/collector@v0.61.0/extension/experimental/storage
[awsproxy]: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.61.0/extension/awsproxy

### DIFF
--- a/component/otelcol/auth/auth.go
+++ b/component/otelcol/auth/auth.go
@@ -1,0 +1,150 @@
+// Package auth provides utilities to create a Flow component from
+// OpenTelemetry Collector authentication extensions.
+//
+// Other OpenTelemetry Collector extensions are better served as generic Flow
+// components rather than being placed in the otelcol namespace.
+package auth
+
+import (
+	"context"
+	"os"
+
+	"github.com/grafana/agent/component"
+	"github.com/grafana/agent/component/otelcol/internal/scheduler"
+	"github.com/grafana/agent/component/otelcol/internal/zapadapter"
+	"github.com/grafana/agent/pkg/build"
+	otelcomponent "go.opentelemetry.io/collector/component"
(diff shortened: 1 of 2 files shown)


mattdurham (approved):
LGTM

PR #2365, opened 2022-10-14
Fix identifier in v2 snmp integration

### LINKED ISSUE #2017: Many v2 integrations do not use identifiers properly

marctc:
#### PR Description

This PR fixes the usage of the identifier (instance key) for SNMP v2 integration

Closes #2017

#### PR Checklist

- [x] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,8 @@ collected logs to standard output. This can be useful for debugging. (@jcreixell
 
 - Remove empty port from the `apache_http` integration's instance label. (@katepangLiu)
 
+- Fix identifier on target creation for SNMP v2 integration. (@marctc)
+
 v0.28.0 (2022-09-29)
 --------------------
 


rfratto (approved):
Thanks!

PR #2369, opened 2022-10-14
river/internal/value: change representation of objects and maps

### LINKED ISSUE #2362: river: interface-based capsules lose type when put in an array
To reproduce, run the following config file with Grafana Agent Flow:

```river 
otelcol.receiver.jaeger "default" {
	protocols {
		thrift_http {}
	}

	output {
		traces = [otelcol.processor.batch.default.input] 
	}
}

otelcol.processor.batch "default" {
	output {
		traces = [otelcol.exporter.otlp.tempo.input]
	}
}

otelcol.exporter.otlp "tempo" {
	client {
		endpoint = "tempo:4317"

		tls {
			insecure             = true
			insecure_skip_verify = true
		}
	}
}
```

You will observe the following error:

```
Error: /etc/agent-config/example.river:13:12: [otelcol.exporter.otlp.tempo.input][0] should be capsule, got object

12 |     output {
13 |         traces = [otelcol.exporter.otlp.tempo.input]
   |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
14 |     }
```

(As an aside, the error message is formatted strangely; it should be an error on the element and not the whole array).

I dug into this a little bit, and from what I can tell, it's caused from when an array is constructed:

https://github.com/grafana/agent/blob/5beae3036616b66c82154b8651fe4016949def70/pkg/river/internal/value/value.go#L73-L89

This will coerce everything into an `interface{}`. Most values will get boxed, but other interface types will get cast and lose their type information that marks them as a capsule. We haven't seen this until now because the `otelcol` components are the first components which return interfaces as capsule values.

Storing the raw `reflect.Value` or `value.Value` instead of `interface{}` fixes the issue, though we need to see what impact that has on the performance benchmarks.

rfratto:
This commit changes the representation of objects and maps from `[]interface{}` and `map[string]interface{}` to use `[]value.Value` and `map[string]value.Value` respectively.

This fixes #2362, an issue where interface-based capsule values where losing type information and being coerced to a type based on the actual value backing the interface.

This also greatly improves the performance of creating arrays and objects, which no longer need to be inidividually copied in a loop:

```
name                   old time/op    new time/op    delta
Object/Non-capsule-10    1.84µs ± 8%    0.00µs ± 0%   -99.89%  (p=0.000 n=19+19)
Object/Capsule-10        1.82µs ± 4%    0.00µs ± 0%   -99.89%  (p=0.000 n=20+19)
Array/Non-capsule-10      389ns ± 2%      19ns ± 2%   -95.20%  (p=0.000 n=19+18)
Array/Capsule-10          389ns ± 1%      19ns ± 0%   -95.19%  (p=0.000 n=20+20)

name                   old alloc/op   new alloc/op   delta
Object/Non-capsule-10    1.88kB ± 0%    0.00kB       -100.00%  (p=0.000 n=20+20)
Object/Capsule-10        1.88kB ± 0%    0.00kB       -100.00%  (p=0.000 n=20+20)
Array/Non-capsule-10       344B ± 0%       24B ± 0%   -93.02%  (p=0.000 n=20+20)
Array/Capsule-10           344B ± 0%       24B ± 0%   -93.02%  (p=0.000 n=20+20)

name                   old allocs/op  new allocs/op  delta
Object/Non-capsule-10      42.0 ± 0%       0.0       -100.00%  (p=0.000 n=20+20)
Object/Capsule-10          42.0 ± 0%       0.0       -100.00%  (p=0.000 n=20+20)
Array/Non-capsule-10       2.00 ± 0%      1.00 ± 0%   -50.00%  (p=0.000 n=20+20)
Array/Capsule-10           2.00 ± 0%      1.00 ± 0%   -50.00%  (p=0.000 n=20+20)
```

As part of this change, the logic used for the concat function has been greatly simplified. This has seemingly no impact on performance, as the benchmark results below show:

```
name       old time/op    new time/op    delta
Concat-10    1.10µs ± 4%    1.09µs ± 1%   ~     (p=0.735 n=18+19)

name       old alloc/op   new alloc/op   delta
Concat-10    1.30kB ± 0%    1.30kB ± 0%   ~     (all equal)

name       old allocs/op  new allocs/op  delta
Concat-10      20.0 ± 0%      20.0 ± 0%   ~     (all equal)
```

I didn't update the CHANGELOG because #2362 only manifested for the `otelcol` components, which haven't been released yet.

### DIFF
--- a/pkg/river/internal/stdlib/stdlib.go
+++ b/pkg/river/internal/stdlib/stdlib.go
@@ -4,12 +4,9 @@ package stdlib
 import (
 	"encoding/json"
 	"os"
-	"reflect"
 
 	"github.com/grafana/agent/pkg/river/internal/value"
 )
-
-var goAny = reflect.TypeOf((*interface{})(nil)).Elem()
 
 // Functions returns the list of stdlib functions by name. The interface{}
 // value is always a River-compatible function value, where functions have at
@@ -53,39 +50,14 @@ var Functions = map[string]interface{}{
 			return args[0], nil
 		}
 
-		// If the imcoming Go slices have the same type, we can have our resulting
-		// slice use the same type. This will allow decoding to use the direct
-		// assignment optimization.
-		//
-		// However, if the types don't match, then we're forced to fall back to
-		// returning []interface{}.
-		//
-		// TODO(rfratto): This could fall back to checking the elements if the
-		// array/slice types don't match. It would be slower, but the direct
-		// assignment optimization probably justifies it.
-		useType := args[0].Reflect().Type()
-		for i := 1; i < len(args); i++ {
-			if args[i].Reflect().Type() != useType {
-				useType = reflect.SliceOf(goAny)
-				break
-			}
-		}
-
-		// Build out the final array.
-		raw := reflect.MakeSlice(useType, finalSize, finalSize)
-
-		var argNum int
+		raw := make([]value.Value, 0, finalSize)
 		for _, arg := range args {
 			for i := 0; i < arg.Len(); i++ {
-				elem := arg.Index(i)
-				if elem.Type() != value.TypeNull {
-					raw.Index(argNum).Set(elem.Reflect())
-				}
-				argNum++
+				raw = append(raw, arg.Index(i))
 			}
 		}
 
-		return value.Encode(raw.Interface()), nil
+		return value.Array(raw...), nil
 	}),
 
 	"json_decode": func(in string) (interface{}, error) {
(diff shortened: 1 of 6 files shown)


mattdurham (approved):
This is way simpler and cleaner. A little surprised there are no gotchas but looks great!

rfratto:
> This is way simpler and cleaner. A little surprised there are no gotchas but looks great!

Yeah, I'm a little nervous that I'm missing something and that something is going to break after this is merged, but we'll see what happens 🤷‍♂️

PR #2382, opened 2022-10-18
Add initial support for prometheus relabel caching

### LINKED ISSUE #2324: Flow: Add caching mechanism to `prometheus.relabel`
We shoud maybe reuse the upstream Prometheus approach of having a per-refID cache for the results of the relabeling process, and clearing entries for dropped and stale series. Currently, the relabeling process runs each time for all metrics, resulting in increased CPU usage.

mattdurham:
#### PR Description

Adds caching to prometheus relabel, it's probably not the end goal because there are cases where the cache can leak. IE if you have two relabel components in a series, the second one starts caching results from the first, and the configuration on the first changes so it never sends the value that the second was caching. 

The more complex solution would be to also add a time-based eviction notice, but that seemed to complex for an edge case. Adding it to the global cache would likely allow one component to have a timer and notify callers that the cache can be evicted. Since all metrics call back to get a global id it has the most knowledge. I think submodules complicate the global cache and am wary of adding that to global cache at the moment.

Closes #2324

#### PR Checklist

- [X] CHANGELOG updated
- [ ] Documentation added
- [X] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@ This document contains a historical list of changes between releases. Only
 changes that impact end-user behavior are listed; changes to documentation or
 internal API changes are not present.
 
-## Main (unreleased)
+Main (unreleased)
+-----------------
 
 > **BREAKING CHANGES**: This release has breaking changes. Please read entries
 > carefully and consult the [upgrade guide][] for specific instructions.
@@ -49,7 +50,7 @@ internal API changes are not present.
 
 - Update Prometheus dependency to v2.38.0. (@rfratto)
 
-- Add caching to Prometheus relabel flow component. (@mattdurham)
+- Add caching to Prometheus relabel component. (@mattdurham)
 
 ### Bugfixes
 
@@ -57,7 +58,8 @@ internal API changes are not present.
 
 - Fix identifier on target creation for SNMP v2 integration. (@marctc)
 
-## v0.28.0 (2022-09-29)
+v0.28.0 (2022-09-29)
+--------------------
 
 ### Features
 
@@ -91,7 +93,8 @@ internal API changes are not present.
 
 - Add metrics for config reloads and config hash (@jcreixell)
 
-## v0.27.1 (2022-09-09)
+v0.27.1 (2022-09-09)
+--------------------
 
 > **NOTE**: ARMv6 Docker images are no longer being published.
 >
@@ -105,7 +108,8 @@ internal API changes are not present.
 
 - Switch docker image base from debian to ubuntu. (@captncraig)
 
-## v0.27.0 (2022-09-01)
+v0.27.0 (2022-09-01)
+--------------------
 
 ### Features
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1713,12 +1715,12 @@ v0.2.0 (2020-04-09)
   (@gotjosh)
 
   These metrics are available to monitor metadata being sent:
-    - `prometheus_remote_storage_succeeded_metadata_total`
```
mattdurham, line 1716: Looks like the formatter changed some values, though it looks like the new ones are more correct?

rfratto: I'm not sure what happened here, but this looks fine to me

CHANGELOG.md
```diff
@@ -43,13 +43,15 @@ collected logs to standard output. This can be useful for debugging. (@jcreixell
...
     exceeded. (@tpaschalis)
-  
+
 ### Enhancements
 
 - Update OpenTelemetry Collector dependency to v0.61.0. (@rfratto)
 
 - Update Prometheus dependency to v2.38.0. (@rfratto)
 
+- Add caching to Prometheus relabel component. (@mattdurham)
```
rfratto, line 53: * Can you prefix this indicating that it's a Flow-specific change 
* Can you change this to explain the effect to end users rather than what was done? (for example: reduce CPU usage of relabeling)

component/prometheus/relabel/relabel.go
```diff
@@ -95,19 +98,27 @@ func (c *Component) Update(args component.Arguments) error {
...
+		var relabelledFm *prometheus.FlowMetric
+		if fm, found := c.cache[m.GlobalRefID()]; found {
+			relabelledFm = fm
+		} else {
+			relabelledFm = m.Relabel(c.mrc...)
+			c.cache[m.GlobalRefID()] = relabelledFm
+		}
+
+		// If stale remove from the cache, the reason we don't exit early is so the stale value can propagate.
+		// TODO: (@mattdurham) This caching can leak and likely needs a timed eviction at some point, but this is simple.
```
rfratto, line 117: Yeah, leaking will _definitely_ happen if a component is removed, since stale markers don't get sent on component shutdown (and Prometheus-team doesn't recommend it working that way). 

This is probably fine for now. As you mentioned, improvements to the global cache to assist in garbage collecting will help with this. 

(Please do create a new issue for us after this is merged to track the improvements so we don't forget that we need to do this though)

component/prometheus/relabel/relabel_test.go
```diff
@@ -0,0 +1,81 @@
+package relabel
```
rfratto, line 1: Can we add a benchmark here so we can start tracking the performance of this?

rfratto (approved):
LGTM!

CHANGELOG.md
rfratto: (also, mentioning the specific component name `prometheus.relabel` here would be nice)

component/prometheus/relabel/relabel_test.go
mattdurham: Adding an incredibly simple best case benchmark. 146ns with cache, 740ns without cache.

Screen Shot 2022-10-18 at 3 59 45 PM

component/prometheus/relabel/relabel.go
mattdurham: Created issue #2383 to handle the longer term thoughts.

PR #2385, opened 2022-10-19
docs/flow: Fix capture group reference documentation

tpaschalis:

#### PR Description
This PR fixes the documentation of capture groups for the prometheus.relabel and discovery.relabel components. The old notation was valid when using HCL in previous Flow iterations, but is different now as we switched to River.

No issue filed

#### PR Checklist

- [ ] CHANGELOG updated (N/A)
- [ ] Tests updated (N/A)

### DIFF
--- a/docs/sources/shared/flow/reference/components/rule-block.md
+++ b/docs/sources/shared/flow/reference/components/rule-block.md
@@ -33,4 +33,4 @@ Here's a list of the available actions, along with a brief description of their 
 * `labelkeep` - Matches `regex` against all label names. Any labels that don't match are removed from the metric's label set.
 
 Finally, note that the regex capture groups can be referred to using either the
-`$1` or `${1}` notation.
+`$CAPTURE_GROUP_NUMBER` or `${CAPTURE_GROUP_NUMBER}` notation.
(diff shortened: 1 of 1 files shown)


marctc (approved):

docs/sources/shared/flow/reference/components/rule-block.md
```diff
@@ -33,4 +33,4 @@ Here's a list of the available actions, along with a brief description of their
 * `labelkeep` - Matches `regex` against all label names. Any labels that don't match are removed from the metric's label set.
 
 Finally, note that the regex capture groups can be referred to using either the
-`$1` or `$${1}` notation.
+`$1` or `${1}` notation.
```
karengermond, line 36: @rfratto and I were discussing whether to use Google's placeholder suggestion (https://developers.google.com/style/placeholders#placeholders-in-inline-text) in [this thread](https://raintank-corp.slack.com/archives/C02GSU8SHBN/p1666123999924289?thread_ts=1666123989.274949&cid=C02GSU8SHBN). In that case, we would want to use something like: 

```suggestion
`$CAPTURE_GROUP_NUMBER` or `${CAPTURE_GROUP_NUMBER}` notation.
```
I leave the decision up to you all...

karengermond (approved):
LGTM

docs/sources/shared/flow/reference/components/rule-block.md
rfratto: It's a GitHub action which adds issues and PRs to our project board, but it doesn't work for PRs from forks. I need to remove the PR automation; you can ignore it failing for now.

PR #2388, opened 2022-10-19
cmd/agent: include basic process and machine metrics

rfratto:
This commit introduces new `agent_resources_*` metrics which explain basic information about the agent and machine the agent is running on:

* CPU usage of agent process
* Memory (virtual and resident) usage of agent process
* Unix start time of agent process
* Machine-wide network usage across all network interfaces
  * Process-specific network usage isn't provided since there doesn't appear to currently be support for collecting this across all platforms at the moment.

Unlike the `agent_process_*` metrics from github.com/prometheus/client_golang, these metrics are fully cross platform and will work on all platforms Grafana Agent is compatible with.

These metrics are intended to provide very basic information about consumption of the host Grafana Agent is running on, and aren't intended to replace fully-featured exporters or Flow components. The intended use case is to visualize these metrics on an Agent-specific resource consumption dashboard.

This metrics are only made available when using Grafana Agent Flow.


cmd/agent/resources_collector.go
```diff
@@ -0,0 +1,159 @@
...
+var _ prometheus.Collector = (*resourcesCollector)(nil)
+
+// newResourcesCollector creates a new resourcesCollector.
+func newResourcesCollector(l log.Logger) *resourcesCollector {
+	rc := &resourcesCollector{
+		log: l,
+
+		processStartTime: prometheus.NewDesc(
+			"agent_resources_process_start_time_seconds",
+			"Start time of the process since Unix epoch in seconds",
```
mattdurham, line 37: nit: Period at end of seconds.

mattdurham: tx/rx bytes also

mattdurham (approved):
Very small nits but looks good. I double-checked the supported architectures and they seem in alignment.

PR #2398, opened 2022-10-20
Do not store the value, only store the labels and global reference id…

mattdurham:
#### PR Description

Instead of storing the whole flowmetric, which includes the value. We only cache the global cache id and the labels. Storing the global ref id saves a lookup and in the simple benchmark saves about 15% . This resolves the issue where the first value is locked in and always reused.

### DIFF
--- a/component/prometheus/relabel/relabel.go
+++ b/component/prometheus/relabel/relabel.go
@@ -8,6 +8,8 @@ import (
 	flow_relabel "github.com/grafana/agent/component/common/relabel"
 	"github.com/grafana/agent/component/prometheus"
 	prometheus_client "github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/prometheus/model/labels"
+
 	"github.com/prometheus/prometheus/model/relabel"
 	"github.com/prometheus/prometheus/model/value"
 )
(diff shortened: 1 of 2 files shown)


component/prometheus/relabel/relabel.go
```diff
@@ -154,12 +159,26 @@ func (c *Component) clearCache() {
...
 }
 
 func (c *Component) addToCache(originalID uint64, fm *prometheus.FlowMetric) {
 	c.cacheMut.Lock()
 	defer c.cacheMut.Unlock()
 
-	c.cache[originalID] = fm
+	if fm == nil {
+		c.cache[originalID] = nil
+		return
```
rfratto, line 171: Should this delete the entry instead of setting it to nil?

rfratto (commented):
Can we add a test for the behavior we experienced? That's the type of thing I'd like to catch in a unit test to avoid regressions.

mattdurham:
Will add some more tests.

component/prometheus/relabel/relabel.go
mattdurham: This handles the case of dropping a metric entirely.

rfratto: Ah, got it 👍

rfratto (approved):

PR #2402, opened 2022-10-21
flow: Allow config blocks to reference component exports

### LINKED ISSUE #2214: Flow: allow config blocks to reference components
Currently, [config blocks][] like [`logging`][] don't support being able to reference other components. I would like to make this possible.

I don't think this is strictly necessary today, but it is a requirement to implement #2215 as  it's listed.
  
## Impleentation

One way to do this would be to represent the set of config blocks as a single non-component node in the controller's DAG. Then it would be allowed to depend on component nodes, being reevaluated whenever a component it depends on changes. 

In the future, we may wish to add each config blocks as separate nodes in the DAG, but I think starting with one node for all config blocks would be good enough for an initial step.

This isn't a small amount of work; a lot of controller code right now assumes that everything in the graph is a component, and that will need to be reworked. 

[config blocks]: https://grafana.com/docs/agent/latest/flow/reference/config-blocks/
[`logging`]: https://grafana.com/docs/agent/latest/flow/reference/config-blocks/logging/

tpaschalis:

#### PR Description
This is a first approach for allowing Flow's special [config blocks](https://grafana.com/docs/agent/latest/flow/reference/config-blocks/) to reference other components, that can be extended to allow for more config blocks in the future.

One way to achieve this is to pre-populate the DAG with a new type of ConfigNode.

This approach gathers all config blocks (just logging for now) when `Apply`ing a configuration file, and uses them to build a NewConfigNode before adding the rest of the components to the graph. The ConfigNode is then wired along all other ComponentNodes, to detect references to components.

The ConfigNode is prepopulated with the default argument options for the various config blocks, and uses the passed blocks to Evaluate the river expressions that were given.

Fixes #2214

#### Notes to the Reviewer
The ConfigNode works a little different than the other ComponentNodes. It does not produce exports, does not report health data, so might be more difficult to debug issues with it.

I'm divided on how we want to handle configuration issues with config blocks after reporting the error; do we act like other components do and keep on with the previous working config, or do we fall back to the defaults?

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,8 @@ Main (unreleased)
     drops data or forces a garbage collection if the defined limits are
     exceeded. (@tpaschalis)
 
+- Flow: Allow config blocks to reference component exports. (@tpaschalis)
+
 ### Enhancements
 
 - Update OpenTelemetry Collector dependency to v0.61.0. (@rfratto)
--- a/pkg/flow/config.go
+++ b/pkg/flow/config.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/grafana/agent/pkg/flow/logging"
 	"github.com/grafana/agent/pkg/river/ast"
 	"github.com/grafana/agent/pkg/river/diag"
 	"github.com/grafana/agent/pkg/river/parser"
@@ -14,8 +13,6 @@ import (
 type File struct {
 	Name string    // File name given to ReadFile.
 	Node *ast.File // Raw File node.
-
-	Logging logging.Options
 
 	// Components holds the list of raw River AST blocks describing components.
 	// The Flow controller can interpret them.
@@ -70,12 +67,9 @@ func ReadFile(name string, bb []byte) (*File, error) {
 		}
 	}
 
-	loggingOpts := logging.DefaultOptions
-
 	return &File{
 		Name:         name,
 		Node:         node,
-		Logging:      loggingOpts,
 		Components:   components,
 		ConfigBlocks: configs,
 	}, nil
--- a/pkg/flow/config.go
+++ b/pkg/flow/config.go
@@ -8,7 +8,6 @@ import (
 	"github.com/grafana/agent/pkg/river/ast"
 	"github.com/grafana/agent/pkg/river/diag"
 	"github.com/grafana/agent/pkg/river/parser"
-	"github.com/grafana/agent/pkg/river/vm"
 )
 
 // File holds the contents of a parsed Flow file.
@@ -20,7 +19,8 @@ type File struct {
 
 	// Components holds the list of raw River AST blocks describing components.
 	// The Flow controller can interpret them.
-	Components []*ast.BlockStmt
+	Components   []*ast.BlockStmt
+	ConfigBlocks []*ast.BlockStmt
 }
 
 // ReadFile parses the River file specified by bb into a File. name should be
@@ -37,8 +37,8 @@ func ReadFile(name string, bb []byte) (*File, error) {
 	// TODO(rfratto): should this code be brought into a helper somewhere? Maybe
 	// in ast?
 	var (
-		loggerBlock *ast.BlockStmt
-		components  []*ast.BlockStmt
+		components []*ast.BlockStmt
+		configs    []*ast.BlockStmt
 	)
 
 	for _, stmt := range node.Body {
@@ -55,7 +55,7 @@ func ReadFile(name string, bb []byte) (*File, error) {
 			fullName := strings.Join(stmt.Name, ".")
 			switch fullName {
 			case "logging":
-				loggerBlock = stmt
+				configs = append(configs, stmt)
 			default:
 				components = append(components, stmt)
 			}
@@ -71,16 +71,12 @@ func ReadFile(name string, bb []byte) (*File, error) {
 	}
 
 	loggingOpts := logging.DefaultOptions
-	if loggerBlock != nil {
-		if err := vm.New(loggerBlock.Body).Evaluate(nil, &loggingOpts); err != nil {
-			return nil, err
-		}
-	}
 
 	return &File{
-		Name:       name,
-		Node:       node,
-		Logging:    loggingOpts,
-		Components: components,
+		Name:         name,
+		Node:         node,
+		Logging:      loggingOpts,
+		Components:   components,
+		ConfigBlocks: configs,
 	}, nil
 }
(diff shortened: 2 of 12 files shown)


pkg/flow/internal/controller/config.go
```diff
@@ -0,0 +1,110 @@
...
+
+// ConfigNode is a controller node which manages agent configuration.
+type ConfigNode struct {
+	mut          sync.RWMutex
+	blocks       []*ast.BlockStmt // Current River blocks to derive config from
+	logger       log.Logger
+	loggingArgs  logging.Options // Evaluated logging arguments for the config
+	loggingBlock *ast.BlockStmt
+	loggingEval  *vm.Evaluator
+}
```
rfratto, line 29: Is the intent that all of our config blocks are represented by one ConfigNode instance? Just want to check :)

rfratto (commented):
Thanks Paschalis! I gave this a quick glance but I'm going to give this a much deeper review later today.

pkg/flow/internal/controller/config.go
tpaschalis: Yeah. Built this with the intention that there'd _always_  exactly one single ConfigNode instance on the graph.

Multiple config blocks would be stored in the slice, and each one would be used for specific subset of the struct fields. Not 100% sure about this approach yet, still playing around to see how multiple/none config nodes could work.

rfratto:
> I'm divided on how we want to handle configuration issues with config blocks after reporting the error; do we act like other components do and keep on with the previous working config, or do we fall back to the defaults?

I'll think about it more but IMO we should always do our best to retain last valid state rather than reverting to defaults.

rfratto:
> The ConfigNode works a little different than the other ComponentNodes. It does not produce exports, does not report health data, so might be more difficult to debug issues with it.

Yeah, we'll have to think about how we want to expose config blocks in the UI. It might make sense to have one DAG node per config block just for making it a more natural fit in the UI, but I think we can kick this can down the hill a little bit for now.

CHANGELOG.md
```diff
@@ -50,6 +50,8 @@ Main (unreleased)
     drops data or forces a garbage collection if the defined limits are
     exceeded. (@tpaschalis)
 
+- Flow: Allow config blocks to reference component exports. (@tpaschalis)
```
rfratto, line 53: Can you update the docs to remove references to the limitation that config blocks can't reference component exports?

pkg/flow/config.go
```diff
@@ -71,16 +71,12 @@ func ReadFile(name string, bb []byte) (*File, error) {
 	}
 
 	loggingOpts := logging.DefaultOptions
```
rfratto, line 73: Does this have any negative effect when logging gets evaluated? For example, if you enable debug logging, is there any guarantee you'd get debug logs during the initial evaluation for all components which the logging block _doesn't_ depend on?

I'm not sure this is something we have to address right now, but I'm generally concerned about finding a way to make the logging behavior easy to understand when it depends on another component in the graph.

pkg/flow/internal/controller/config.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/grafana/agent/pkg/river/vm"
+)
+
+const (
+	configNodeID = "configNode"
+
+	loggingBlockID = "logging"
+)
+
+// ConfigNode is a controller node which manages agent configuration.
```
rfratto, line 21: Can you expand this comment that makes it more clear the intent is that this is some kind of super-node which contains the state for all config blocks?

pkg/flow/internal/controller/config.go
```diff
@@ -0,0 +1,110 @@
...
+		return cn.loggingBlock, fmt.Errorf("decoding River: %w", err)
+	}
+	cn.loggingArgs = args
+
+	l, ok := cn.logger.(*logging.Logger)
+	if ok {
+		err := l.Update(cn.loggingArgs)
+		if err != nil {
+			return cn.loggingBlock, fmt.Errorf("could not update logger: %v", err)
+		}
```
rfratto, line 100: It looks like we may want this to be the _only_ call to Update:

https://github.com/grafana/agent/blob/4373af81da5e4328d5bab565ee7378f163647396/pkg/flow/flow.go#L200-L203

As it is now, every time the file is reloaded it'll reset to the defaults here.

pkg/flow/internal/controller/loader.go
```diff
@@ -75,6 +75,15 @@ func (l *Loader) Apply(parentScope *vm.Scope, blocks []*ast.BlockStmt) diag.Diag
 		newGraph dag.Graph
 	)
 
+	// Pre-populate graph with a ConfigNode.
+	// TODO (@tpaschalis) Are we okay with recreating the config node on each
+	// Apply? Should we check if it already exists and just update its blocks
+	// instead? if exist := l.graph.GetByID(configNodeID); exist == nil { ...
```
rfratto, line 81: Yeah, this is probably fine for now. The main reason we don't want to recreate other nodes is because they manage an instance of a component. Recreating vs updating here doesn't seem to have a similar impact that recreating a component node does.

rfratto (approved):
Nice! This looks good. 

I think in the future, as more types of nodes get added, we'll want to design a common interface that both component nodes and config block nodes implement, but this looks perfect for now.

CHANGELOG.md
tpaschalis: Yep, done!

pkg/flow/internal/controller/loader.go
tpaschalis: Ok, makes sense.

pkg/flow/config.go
tpaschalis: I've removed these leftovers referring to the loggingOpts and storing them in the File struct.

Right now, the logger will get initialized once from cmd/agent/flow_run.go

tpaschalis: I've removed these leftovers referring to the loggingOpts and storing them in the File struct, as they weren't being anymore.

Right now, the logger will get initialized once from cmd/agent/flow_run.go and then Updated only from the ConfigNode.

> is there any guarantee you'd get debug logs during the initial evaluation for all components which the logging block doesn't depend on?

No, there's no such guarantee. One solution would be to evaluate the ConfigNode by itself separately instead of when walking the graph, but I'm not too fond of it.

pkg/flow/internal/controller/config.go
tpaschalis: True. I've removed this, and now we retain last valid state rather than reverting to defaults.

karengermond (commented):
LGTM

karengermond (commented):
LGTM -- this was already merged when I was on PTO but wanted to review for issues that we could fix in another PR.  No issue -- perfect.

Issue #2347, opened 2022-10-12
panic with invalid traces configuration


tpaschalis:
The following configuration snippet panics during YAML decoding without providing any user-friendly indication of what went wrong

```yaml
server:
  log_level: info

traces:
  configs:
    - name: default
      receivers:
        otlp:
```

<details>
panic: interface conversion: interface {} is nil, not map[interface {}]interface {} [recovered]
	panic: interface conversion: interface {} is nil, not map[interface {}]interface {} [recovered]
	panic: interface conversion: interface {} is nil, not map[interface {}]interface {} [recovered]
	panic: interface conversion: interface {} is nil, not map[interface {}]interface {}

goroutine 1 [running]:
gopkg.in/yaml%2ev2.handleErr(0x1400103fae8)
	/Users/tpaschalis/go/pkg/mod/github.com/rfratto/go-yaml@v0.0.0-20211119180816-77389c3526dc/yaml.go:256 +0x84
panic({0x107bdbca0, 0x140010776e0})
	/usr/local/go/src/runtime/panic.go:838 +0x204
gopkg.in/yaml%2ev2.handleErr(0x1400103f7c0)
	/Users/tpaschalis/go/pkg/mod/github.com/rfratto/go-yaml@v0.0.0-20211119180816-77389c3526dc/yaml.go:256 +0x84
panic({0x107bdbca0, 0x140010776e0})
	/usr/local/go/src/runtime/panic.go:838 +0x204
gopkg.in/yaml%2ev2.handleErr(0x1400103f1f0)
	/Users/tpaschalis/go/pkg/mod/github.com/rfratto/go-yaml@v0.0.0-20211119180816-77389c3526dc/yaml.go:256 +0x84
panic({0x107bdbca0, 0x140010776e0})
	/usr/local/go/src/runtime/panic.go:838 +0x204
github.com/grafana/agent/pkg/traces.(*ReceiverMap).UnmarshalYAML(0x14000cb7578, 0x1030b1cc8?)
	/Users/tpaschalis/GitRepos/agent/pkg/traces/config.go:160 +0x488
gopkg.in/yaml%2ev2.(*decoder).callUnmarshaler(0x140001adc80, 0x140002d1110, {0x133a449c8, 0x14000cb7578})
	/Users/tpaschalis/go/pkg/mod/github.com/rfratto/go-yaml@v0.0.0-20211119180816-77389c3526dc/decode.go:270 +0xa4
[...]

PR #2422, opened 2022-10-24
Avoid a panic with an invalid traces configuration

ptodev:
#### PR Description
There used to be a panic due to a configuration such as this:
```
server:
  log_level: info

traces:
  configs:
    - name: default
      receivers:
        otlp:
```
The panic message was not clear and user friendly. The new message makes it much more clear to the user why the agent failed to start.

Fixes https://github.com/grafana/agent/issues/2347

#### Notes to the Reviewer
I wrote a new "Test" function, instead of adding the test to an already existing function such as ```TestOTelConfig()```. This is because I would have had to add a 2nd ```expectedError``` field, which would only be used by this test. However, if you think there is a value in adding this test to the ```TestOTelConfig()``` function, please let me know and I will move it there.

Also, I think updating documentation is not needed., so I'll ignore that checkbox below.

#### PR Checklist

- [x] CHANGELOG updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,8 @@ Main (unreleased)
 
 - Fix bug when specifying Blackbox's modules when using Blackbox integration. (@marctc)
 
+- Tracing: fix a panic when the required `protocols` field was not set in the `otlp` receiver. (@ptodev)
+
 v0.28.0 (2022-09-29)
 --------------------
 
--- a/pkg/traces/config.go
+++ b/pkg/traces/config.go
@@ -157,9 +157,14 @@ func (r *ReceiverMap) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	for k := range *r {
 		if strings.HasPrefix(k, otlpReceiverName) {
 			// for http and grpc receivers, include_metadata is set to true by default
-			protocolsCfg, ok := (*r)[k].(map[interface{}]interface{})["protocols"].(map[interface{}]interface{})
+			receiverCfg, ok := (*r)[k].(map[interface{}]interface{})
 			if !ok {
 				return fmt.Errorf("failed to parse OTLP receiver config: %s", k)
+			}
+
+			protocolsCfg, ok := receiverCfg["protocols"].(map[interface{}]interface{})
+			if !ok {
+				return fmt.Errorf("otlp receiver requires a \"protocols\" field which must be a YAML map: %s", k)
 			}
 
 			for _, p := range protocols {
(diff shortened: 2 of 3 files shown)


pkg/traces/config.go
```diff
@@ -157,11 +157,16 @@ func (r *ReceiverMap) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
 			// for http and grpc receivers, include_metadata is set to true by default
-			protocolsCfg, ok := (*r)[k].(map[interface{}]interface{})["protocols"].(map[interface{}]interface{})
+			receiverCfg, ok := (*r)[k].(map[interface{}]interface{})
 			if !ok {
 				return fmt.Errorf("failed to parse OTLP receiver config: %s", k)
 			}
 
+			protocolsCfg, ok := receiverCfg["protocols"].(map[interface{}]interface{})
+			if !ok {
+				return fmt.Errorf("failed to parse OTLP receiver protocol config: %s", k)
```
rfratto, line 167: Can this error message explain why the config is invalid so users know what to fix?

rfratto:
@ptodev When you get a chance, could you sign the CLA please?

CHANGELOG.md
```diff
@@ -69,6 +69,8 @@ Main (unreleased)
 
 - Fix identifier on target creation for SNMP v2 integration. (@marctc)
 
+- Avoid a panic with an invalid traces configuration. (@ptodev)
```
rfratto, line 72: Can this be changed to better explain what the panic was caused by?

Something like this:

> Tracing: avoid a panic when the required `protocols` field was not set in the `otlp` receiver.

ptodev: Thank you, this is done now. I agree that the new description sounds much better.

pkg/traces/config.go
ptodev: Good idea, but I wasn't sure what would be an accurate error message. We could say something like "no protocol config found", but I'm not sure if this would be accurate all the time. For example, if the type assertion failed because the yaml has a string instead of a map - then the protocol info isn't empty, but it's not what we expected (a map). Maybe an error like "no OTLP receiver protocol information in a valid key:value format was found"? Or maybe somehow we can report the line in the yaml config which has a problem?

rfratto: A generic message mentioning the requirements is probably sufficient: 

> otlp receiver requires a "protocol" field which must be a YAML map

ptodev: Thank you, I changed it to the error message you said and I do like it a lot more now. Just one thing - I wrote it as "protocols" instead of "protocol", I think it's plural in the config.

tpaschalis (approved):

CHANGELOG.md
ptodev: I added a test for this error message as well.

PR #2423, opened 2022-10-24
`prometheus.*` components: immediately export receiver on component construction

### LINKED ISSUE #1946: Flow: prometheus.remote_write initially exports nil receiver
Use the following config file with the current Flow implementation in main:

```river
logging {
  level  = "debug"
  format = "logfmt"
}

  endpoint {
  }
}
```

`prometheus.remote_write` initially exports a nil value for its receiver. This propagates through the graph and opens opportunities for components to crash (#1945). 

`prometheus.remote_write` should export a non-nil value on component creation, even if it's a no-op appendable. This will prevent other components from receiving a nil value.

rfratto:
This commit changes Flow Prometheus components which export receivers to immediately export their receivers on component construction. Receivers for these components remain the same through the process lifetime, so there's no need for them to emit exports at run or update time.

This is also a minor (but likely immeasurable) performance enhancement, where components depending on `prometheus.remote_write` or `prometheus.relabel` won't be forced to update twice in a row near the process start time. 

The CHANGELOG has been updated to reflect the performance improvement angle, since it is the only way for the changes to be observed by end users.

Closes #1946.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,10 @@ Main (unreleased)
   Grafana Agent, but are not meant to act as a replacement for fully featured
   components like `prometheus.integration.node_exporter`. (@rfratto)
 
+- Grafana Agent Flow: `prometheus.remote_write` and `prometheus.relabel` will
+  now export receivers immediately, removing the need for dependant components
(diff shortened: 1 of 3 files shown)


tpaschalis (approved):
LGTM!

PR #2426, opened 2022-10-25
Document how operator hierarchy is merged.

jeschkies:
#### PR Description
I took me a while to figure out how the hierarchy of the Grafana Agent Operator resources are picked up and merged. So I've created a visualization and added a description that would have helped me.

#### PR Checklist

- [ ] CHANGELOG updated
- [x] Documentation added
- [ ] Tests updated


docs/sources/operator/architecture.md
```diff
@@ -43,6 +43,24 @@ A single resource can belong to multiple hierarchies. For example, if two
...
+<img src="../../assets/hierarchy.svg" alt="Hierarchy" />
+
+### Debug the Hierarchy
+
+The generated configurations are save in secrets. They can be downloaded and
+validated manually with the following commands.
+
+```
+$ kubectl get secrets <???>-logs-config -o json | jq -r '.data."agent.yml"' | base64 --decode
+$ kubectl get secrets <???>-config -o json | jq -r '.data."agent.yml"' | base64 --decode
```
jeschkies, line 61: How are the config names defined? What is the pattern?

docs/sources/operator/architecture.md
```diff
@@ -43,6 +43,24 @@ A single resource can belong to multiple hierarchies. For example, if two
 GrafanaAgents use the same Probe, modifying that Probe will cause both
 GrafanaAgents to be reconciled.
 
+### Build the Hierarchy
+
+The hierarchy is built via label matching on the custom resources. Below is a
+graph illustrating the matching. The `GrafanaAgent` picks up the `MetricsInstance`
+and `LogsInstance` that match the labels `app.kubernetes.io/name: loki` and `app.kubernetes.io/instance: release-name`. The instances pick up the resources the same way.
```
jeschkies, line 50: We probably should use different labels for an example.

karengermond, line 50:
```suggestion
Grafana Agent Operator builds the hierarchy using label matching on the custom resources. The following figure illustrates the matching. The `GrafanaAgent` picks up the `MetricsInstance`
and `LogsInstance` that match the labels `app.kubernetes.io/name: loki` and `app.kubernetes.io/instance: release-name`, respectively. The instances pick up the resources the same way.
```

docs/sources/operator/architecture.md
```diff
@@ -43,6 +43,24 @@ A single resource can belong to multiple hierarchies. For example, if two
...
+
+<img src="../../assets/hierarchy.svg" alt="Hierarchy" />
+
+### Debug the Hierarchy
+
+The generated configurations are save in secrets. They can be downloaded and
+validated manually with the following commands.
```
karengermond, line 57:
```suggestion
The generated configurations are saved in secrets. To download and
validate them manually, use the following commands:
```

docs/sources/operator/architecture.md
```diff
@@ -43,6 +43,24 @@ A single resource can belong to multiple hierarchies. For example, if two
 GrafanaAgents use the same Probe, modifying that Probe will cause both
 GrafanaAgents to be reconciled.
 
+### Build the Hierarchy
+
+The hierarchy is built via label matching on the custom resources. Below is a
+graph illustrating the matching. The `GrafanaAgent` picks up the `MetricsInstance`
+and `LogsInstance` that match the labels `app.kubernetes.io/name: loki` and `app.kubernetes.io/instance: release-name`. The instances pick up the resources the same way.
+
+<img src="../../assets/hierarchy.svg" alt="Hierarchy" />
```
karengermond, line 52: The text on the edges in this graph is always the same and it makes the image busy.  Should we leave it out since you say this in the text above the image?  Another approach might be to add a callout on the left that points to some of the edges that says:

"The operator matches each instance against:

```matchLabels
   app.kubernetes.io/name: loki
   app.kubernetes.io/instance: release-name
```

docs/sources/operator/architecture.md
karengermond: What would be good labels for the example?  Might want to do it now because it might slip through the cracks...

karengermond (commented):
Thanks for adding this!  I'm working on updates to the Agent Operator topics and this filled in a gap.

karengermond (approved):

docs/sources/operator/architecture.md
jeschkies: I believe the pattern is very common. Unfortunately I don't know the space well enough.

docs/sources/operator/architecture.md
karengermond: I can work on it later because I'm working on these topics anyways.  It was created by a former writer, Karen M (not to be confused with me, Karen G. 😄 ) .

docs/sources/operator/architecture.md
karengermond: @captncraig Do you know what labels we could use instead?

captncraig: In the quickstart we use:

```
  metrics:
    instanceSelector:
      matchLabels:
        agent: grafana-agent-metrics
  logs:
    instanceSelector:
      matchLabels:
        agent: grafana-agent-logs
```

and 

```
podMonitorSelector:
    matchLabels:
      instance: primary
```

I'd usually recommend using the `name` of the parent object as the label value. And it looks like we have been using `instance` as the key. Maybe we should standardize on a more namespaced label key, but thats a different discussion entirely.

docs/sources/operator/architecture.md
captncraig: I am also confused about the match labels. The diagram just has `name:` not the full `app.kubernetes.io/name:`. I'd probably update the diagram with the full names?

captncraig (approved):

PR #2428, opened 2022-10-25
Report usage of enabled integrations

### LINKED ISSUE #1830: Add integration usage to telemetry
Recently, in https://github.com/grafana/agent/issues/1488, some telemetry data was introduced.

I propose to leverage this feature to get a better grip on the usage of integrations in the agent. Which integrations are configured? How many metrics do they collect, how many logs? This could for example be an average over the currently used reporting-interval.

marctc:
#### PR Description

This PR enables the reporting of enabled integrations to grafana. This
would help us to identify which integrations are more used than others.

The information sent it's completely anonymous. This is the [full list](https://grafana.com/docs/agent/next/configuration/flags/#report-information-usage) of all
information sent by the reporter. The reporting can always be disabled using 
`--disable-reporting`.

Closes #1830 

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/config/integrations.go
+++ b/pkg/config/integrations.go
@@ -114,6 +114,22 @@ func (c *VersionedIntegrations) setVersion(v integrationsVersion) error {
 	}
 }
 
+// GetIntegrationsList returns a slice of enabled integrations
+func (c *VersionedIntegrations) GetIntegrationsList() []string {
+	integrations := []string{}
+	if c.configV1 != nil {
+		for _, integration := range c.configV1.Integrations {
+			integrations = append(integrations, integration.Name())
+		}
+	}
+	if c.configV2 != nil {
+		for _, integration := range c.configV2.Configs {
+			integrations = append(integrations, integration.Name())
+		}
+	}
+	return integrations
+}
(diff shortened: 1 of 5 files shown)


karengermond (approved):
LGTM

pkg/config/integrations.go
```diff
@@ -114,6 +114,22 @@ func (c *VersionedIntegrations) setVersion(v integrationsVersion) error {
...
+func (c *VersionedIntegrations) EnabledIntegrations() []string {
+	integrations := []string{}
+	if c.configV1 != nil {
+		for _, integration := range c.configV1.Integrations {
+			integrations = append(integrations, integration.Name())
+		}
+	}
+	if c.configV2 != nil {
+		for _, integration := range c.configV2.Configs {
+			integrations = append(integrations, integration.Name())
```
rfratto, line 127: Won't this double count multiple instances of the same integration?

rfratto (approved):
LGTM but you might need to check for name uniqueness

PR #2431, opened 2022-10-26
Convert to using prometheus appendable

### LINKED ISSUE #2037: Flow: use Prometheus directly instead of abstraction
Our Prometheus components (`metrics.scrape`, `metrics.mutate`, `metrics.remote_write`) should behave exactly as Prometheus does. 

Currently, Flow uses an abstraction for its metrics pipeline:

* A [`metrics.Receiver`](https://github.com/grafana/agent/blob/8335534a0efcd4b3115a35b225e82f25dc4100ff/component/metrics/receiver.go#L11) receives samples at a given timestamp
* A [`metrics.FlowMetric`](https://github.com/grafana/agent/blob/8335534a0efcd4b3115a35b225e82f25dc4100ff/component/metrics/receiver.go#L17-L22) represents an individual sample

This abstraction is insufficient in its current state:

* It does yet not support writing exemplars  
* It will not immediately support the upcoming work to write metadata to the WAL 
* It will not immediately support the upcoming sparse histogram work  

It also has behavioral differences:

* #2035 notes that failing to append metrics don't fail the scrape, which may cause `up != 1` alerts to no longer fire. 

The list of behavioral differences may or may not be complete: abstractions introduce uncertainty, where it is harder to know if there is any other documented or undocumented behaviors of the underlying code which the abstraction does not satisfy. 

For all of the issues above to be resolved, the abstraction would have to be changed to become Prometheus-like. Instead of maintaining an abstraction over time to match the behavior of a moving target, it would be easier to remove the abstraction and use Prometheus code directly. 

Good abstractions are hard to make, and should only be attempted after accumulating a ton of experience with all the things it is abstracting over. I would recommend we write components to use their native APIs instead of trying to introduce abstractions. This means that Prometheus components should use Prometheus code, and OpenTelemetry Collector components should use OpenTelemetry Collector code. 

If components use their native APIs, then a shim is needed to move between APIs. This is similar to an abstraction, but it reduces the surface area of where bugs can be introduced, allowing the native code path to always work fully as intended. How these shims would be exposed to users is out of scope of this proposal.

mattdurham:
#### PR Description

This switches over to using the storage.appendable/appender interfaces. Exemplars and metadata aren't supported but the hooks are all there. I didn't want to overload this PR.

Closes #2037

#### PR Checklist

- [ ] CHANGELOG updated
- [X] Documentation added
- [X] Tests updated


component/prometheus/fanout.go
```diff
@@ -0,0 +1,132 @@
...
+func (a *appender) AppendExemplar(ref storage.SeriesRef, l labels.Labels, e exemplar.Exemplar) (storage.SeriesRef, error) {
+	// TODO implement me
+	panic("implement me")
+}
+
+// UpdateMetadata satisifies the Appender interface.
+func (a *appender) UpdateMetadata(ref storage.SeriesRef, l labels.Labels, m metadata.Metadata) (storage.SeriesRef, error) {
+	// TODO implement me
+	panic("implement me")
+}
```
rfratto, line 132: Let's remove the panic before merging please, since exemplars definitely get pushed in our deployment env. Returning an error should be safe though.

component/prometheus/relabel/relabel_test.go
```diff
@@ -120,29 +79,30 @@ func TestNil(t *testing.T) {
...
-		Receive: func(timestamp int64, metrics []*prometheus.FlowMetric) {
-		},
-	}
 	l := log.NewSyncLogger(log.NewLogfmtLogger(os.Stderr))
 
-	relabeller, _ := New(component.Options{
+	fanout := prometheus.NewFanout(func(ref storage.SeriesRef, l labels.Labels, tt int64, v float64) (storage.SeriesRef, labels.Labels, int64, float64, error) {
+		if !l.Has("new_label") {
+			panic("must have new label")
+		}
```
rfratto, line 92: Remove the panic please (also I'm not sure what new_label means)

rfratto: Oh, this is in test code, nvm :) 

This would be more idiomatic for our existing tests: 

```go
requires.True(t, l.Has("new_label"), "new_label must be provided") 
```

component/prometheus/fanout.go
```diff
@@ -0,0 +1,131 @@
...
+
+type intercept func(ref storage.SeriesRef, l labels.Labels, t int64, v float64) (storage.SeriesRef, labels.Labels, int64, float64, error)
+
+// Fanout supports the default Flow style of appendables since it can go to multiple outputs. It also allows the intercepting of appends.
+type Fanout struct {
+	mut sync.RWMutex
+	// intercept allows one to intercept the series before it fans out to make any changes. If labels.Labels returns nil the series is not propagated.
+	// Intercept shouuld be thread safe and can be called across appenders.
+	intercept func(ref storage.SeriesRef, l labels.Labels, t int64, v float64) (storage.SeriesRef, labels.Labels, int64, float64, error)
+
```
rfratto, line 25: Have you considered making a separate storage.Appendable interface just for intercept? It would behave similarly to the middleware pattern at that point. As it is now, it feels like Fanout is responsible for a little too much.

component/prometheus/fanout.go
```diff
@@ -0,0 +1,131 @@
...
+		children:    make([]storage.Appender, len(f.children)),
+		intercept:   f.intercept,
+		componentID: f.componentID,
+	}
+	for i, x := range f.children {
+		if x == nil {
+			continue
+		}
+		app.children[i] = x.Appender(ctx)
+	}
```
rfratto, line 64: What about setting the capacity of `children` instead of the length? That way you can remove all the nil checks in the appender itself and guarantee that each child is non-nil.

component/prometheus/fanout.go
```diff
@@ -0,0 +1,131 @@
...
+
+// Append satisfies the Appender interface.
+func (a *appender) Append(ref storage.SeriesRef, l labels.Labels, t int64, v float64) (storage.SeriesRef, error) {
+	if ref == 0 {
+		ref = storage.SeriesRef(GlobalRefMapping.GetOrAddGlobalRefID(l))
+	}
+	nRef := ref
+	nL := l
+	nT := t
+	nV := v
```
rfratto, line 84: Can we use longer names here please? I personally find that `nL` is harder to read over `l`.

component/prometheus/globalrefmap.go
```diff
@@ -79,6 +79,10 @@ func (g *GlobalRefMap) GetOrAddGlobalRefID(l labels.Labels) uint64 {
 	g.mut.Lock()
 	defer g.mut.Unlock()
 
+	if l == nil {
+		return 0
+	}
```
rfratto, line 84: Can we add a comment here explaining why we return 0 here? Presumably it's because a nil label set indicates a dropped series, right?

component/prometheus/relabel/relabel_test.go
```diff
@@ -15,100 +17,57 @@ import (
...
+	lbls := labels.FromStrings("__address__", "localhost")
+	relabeller.relabel(0, lbls)
 	require.Len(t, relabeller.cache, 1)
 	_ = relabeller.Update(Arguments{
 		MetricRelabelConfigs: []*flow_relabel.Config{},
 	})
 	require.Len(t, relabeller.cache, 0)
 }
 
-func TestThatValueIsNotReused(t *testing.T) {
```
rfratto, line 57: Is this test no longer necessary? It felt like a good regression test to have

component/prometheus/remotewrite/remote_write.go
```diff
@@ -71,7 +70,10 @@ func NewComponent(o component.Options, c Arguments) (*Component, error) {
 		remoteStore: remoteStore,
 		storage:     storage.NewFanout(o.Logger, walStorage, remoteStore),
 	}
-	res.receiver = &prometheus.Receiver{Receive: res.Receive}
+	res.receiver = &appendable{
+		inner:       res.storage,
+		componentID: o.ID,
+	}
```
rfratto, line 76: If we make the interceptor its own storage.Appendable, then you should be able to replace the custom appendable struct here with the interceptor to reduce some code, I think.

component/prometheus/relabel/relabel_test.go
mattdurham: Can benchmarks access the testing.T?

rfratto: oh, it'd be `b` here then. `require` can be given any `testing.T` or `testing.B` instance

component/prometheus/fanout.go
mattdurham: I originally did have more specific appendables, but it's really handy to reuse the fanout for tests. There was a fair amount of duplicated code so I refactored to use one that allows some middleware in between.

component/prometheus/fanout.go
mattdurham: Removed the nil checks from the appender and simplified a bit of the logic. Nils are allowed in the appendable but not the appender.

component/prometheus/globalrefmap.go
mattdurham: Moreso a guard against bad input and not pollute the cache. The bad input could be a dropped series.

component/prometheus/relabel/relabel_test.go
mattdurham: With the removal of flowmetric the cache can never be passed the value so it's not necessary. The test was mostly obsolete earlier but now its more explicit that values cannot enter the cache.

component/prometheus/remotewrite/remote_write.go
mattdurham: I was torn on this one and did it both ways, and felt that remote_write having its own custom one was warranted because anything interfacing with remote_write will have to do the translation ie appendexemplar and updatemetadata. Where my gut is that other components wont have any customer behavior on those.

component/prometheus/relabel/relabel_test.go
mattdurham: Doh!

component/prometheus/fanout.go
rfratto: I'd like a second opinion from @tpaschalis here. My gut feeling is that we shouldn't put both intercept + fanout logic in the same thing, but I don't know if I'm being too idealistic with this.

tpaschalis: I think I'd be in favor of a separate appendable here.

If it's only the testing code that gets more complicated by splitting the intercept+fanout logic, then maybe we should give it a try, since we won't be altering the test code that much (it's already tested upstream and we only need to validate our component's behavior).

mattdurham: This is resolved now.

component/prometheus/fanout.go
```diff
@@ -0,0 +1,101 @@
...
+	return nil
+}
+
+// Rollback satisifies the Appender interface.
+func (a *appender) Rollback() error {
+	for _, x := range a.children {
+		_, _ = x, a.Rollback()
+	}
+	return nil
+}
```
rfratto, line 91: Have you considered using a multierror here and returning the error back to the caller so it can make the decision for what to do with error handling?

Ultimately I don't think the fanout should be hiding errors and it should be the callers responsibility.

rfratto (approved):
LGTM! My only big concern is to change the fanout to _not_ hide errors, since there's a few important things that depend on errors being returned. For example, the target should be marked as failed if it couldn't be written to the WAL (such as having invalid or duplicate labels).

PR #2447, opened 2022-10-28
separate fanout and intercept patterns

mattdurham:
Separate the fanout and interceptor.

### DIFF
--- a/component/prometheus/interceptor.go
+++ b/component/prometheus/interceptor.go
@@ -1,0 +1,125 @@
+package prometheus
+
+import (
+	"context"
+	"fmt"
+	"sync"
+
+	"github.com/prometheus/prometheus/model/exemplar"
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/model/metadata"
+	"github.com/prometheus/prometheus/storage"
+)
+
+type intercept func(ref storage.SeriesRef, l labels.Labels, t int64, v float64) (storage.SeriesRef, labels.Labels, int64, float64, error)
+
+// Interceptor supports the concept of an appendable/appender that you can add a func to be called before
+// the values are sent to the child appendable/append.
+type Interceptor struct {
+	mut sync.RWMutex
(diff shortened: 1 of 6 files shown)


component/prometheus/interceptor.go
```diff
@@ -0,0 +1,125 @@
...
+	"fmt"
+	"sync"
+
+	"github.com/prometheus/prometheus/model/exemplar"
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/model/metadata"
+	"github.com/prometheus/prometheus/storage"
+)
+
+type intercept func(ref storage.SeriesRef, l labels.Labels, t int64, v float64) (storage.SeriesRef, labels.Labels, int64, float64, error)
```
rfratto, line 14: What do you think about making this more similar to the middleware pattern, meaning:

```go
type intercept func(
  ref storage.SeriesRef, 
  l labels.Labels, 
  t int64, 
  v float64,
  next storage.Appender,
) (storage.SeriesRef, error)
```

This gives the interceptor more control, including skipping the next appender. It also means the `Append` method can be made more minimal: 

```go
// Append satisfies the Appender interface.
 func (a *interceptappender) Append(ref storage.SeriesRef, l labels.Labels, t int64, v float64) (storage.SeriesRef, error) {
 	if ref == 0 {
 		ref = storage.SeriesRef(GlobalRefMapping.GetOrAddGlobalRefID(l))
 	}

        return a.intercept(ref, l, t, v, a.child) 
}
```

component/prometheus/interceptor.go
```diff
@@ -0,0 +1,125 @@
...
+
+	// child is where to fan out.
+	child storage.Appendable
+
+	// ComponentID is what component this belongs to.
+	componentID string
+}
+
+// NewInterceptor creates a interceptor appendable.
+func NewInterceptor(inter intercept, child storage.Appendable, componentID string) *Interceptor {
```
rfratto, line 32: IMO at least the interceptor should be a required argument here.

mattdurham:
Updated to middleware pattern and other PR feedback.

rfratto (approved):
LGTM, I like it 👍

PR #2455, opened 2022-10-31
Use Go 1.19 for builds

rfratto:
This updates the GitHub Actions and build image to use Go 1.19.2 for builds. After this PR, a new build image tag will need to be cut and a follow up PR will be needed to update everything. 

We should probably wait until the [1.19.3 release](https://groups.google.com/g/golang-nuts/c/0D1HwgYge94/m/SO3Tr1zBBwAJ) is available and use that instead, but I'm still opening this in draft just to queue up the work. 

This also removes `-gcflags "all=-N -l"` from our debug flags to allow tests to pass again. Those flags, which disable optimizations and inlining, should now be the responsibility of tools like delve and GoLang to inject. Passing the flags in Go 1.19 causes issues for Macs (see issue 54291 in the Go repo).

### DIFF
--- a/build-image/Dockerfile
+++ b/build-image/Dockerfile
@@ -19,7 +19,7 @@ FROM alpine:3.16 as docker
 RUN apk add --no-cache docker-cli docker-cli-buildx
 
 # Dependency: Go and Go dependencies
-FROM golang:1.18-bullseye as golang
+FROM golang:1.19.3-bullseye as golang
 
 # Keep in sync with cmd/agent-operator/DEVELOPERS.md
 ENV CONTROLLER_GEN_VERSION v0.8.0
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,10 +14,10 @@ jobs:
     steps:
     - name: Checkout code
       uses: actions/checkout@v3
-    - name: Set up Go 1.18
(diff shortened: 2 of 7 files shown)


build-image/Dockerfile
```diff
@@ -19,7 +19,7 @@ FROM alpine:3.16 as docker
 RUN apk add --no-cache docker-cli docker-cli-buildx
 
 # Dependency: Go and Go dependencies
-FROM golang:1.18-bullseye as golang
+FROM golang:1.19.2-bullseye as golang
```
rfratto, line 22: I thought about changing this to just `1.19` but since we don't automatically update the build image, it's better for us to be explicit about which minor version we're using so we can upgrade it when necessary.

tpaschalis (approved):
LGTM. I haven't followed the release notes, are we temporarily removing the `-gcflags "all=-N -l"` flags until that issue is resolved, or are they no longer required?

rfratto:
> LGTM. I haven't followed the release notes, are we temporarily removing the -gcflags "all=-N -l" flags until that issue is resolved, or are they no longer required?

IMO they've never been required; they're only required when using a debugger but you'd probably never run the binary produced through make through a debugger anyway. We can definitely add them back after the upstream issue is resolved, though.

PR #2462, opened 2022-11-02
otelcol.receiver.prometheus: new component

### LINKED ISSUE #2301: Flow: otelcol.receiver.prometheus component
Create an OpenTelemetry Collector receiver component called `prometheus` and expose it as a Flow component called `otelcol.receiver.prometheus`. 

This component would act as a bridge between the `otelcol.*` component and the `prometheus.*` components, where it exports a Prometheus `storage.Appendable` interface for Prometheus components to send metrics to. The Prometheus metrics would then get converted to OpenTelemetry metrics on the fly before being send to the `otelcol.*` consumers configured with the component. 

```river 
prometheus.scrape "default" {
  targets = [...]
  send_to = [otelcol.receiver.prometheus.default.receiver] 
}

otelcol.receiver.prometheus "default" {
  output {
    metrics = [otelcol.exporter.otlp.default.input] 
  } 
}

otelcol.exporter.otlp "default" {
  ...
}
```

tpaschalis:
#### PR Description
This PR adds a new `otelcol.receiver.prometheus` Flow component. The new component exports a Prometheus storage.Appendable, where all metrics are converted to an OTel-compatible format and then forwarded to one or more OTel metric consumers.

Fixes #2301 

#### Notes to the Reviewer

* The biggest chunk of this PR is bringing in the `internal` package from the opentelemetry-collector-contrib repo's `prometheusreceiver. I copied over the package as it existed on v0.61.0, but removed the test files.
https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.61.0/receiver/prometheusreceiver/internal
* The prometheusreceiver code expects the context passed to the Appender to have a scrape Target and a MetadataStore implementation encoded. I've amended the fanout appendable to add an empty Target and a NoopMetadataStore to the context. Unfortunately, this will be true for _every_ call of the appender, not only those tied to the otelcol bridge component. I don't _think_ this will be too much of a performance penalty though.
* There's a GC interval according to the upstream documentation, should be 'at least as long as the longest scrape interval plus a delta to avoid jobs getting GC'ed between scrapes'. We don't have that information from upstream components, so I've hardcoded to 5 minutes. We _could_ add it as an argument, not sure if it's worth it.

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [X] Tests updated

### DIFF
--- a/component/otelcol/receiver/prometheus/prometheus.go
+++ b/component/otelcol/receiver/prometheus/prometheus.go
@@ -1,0 +1,173 @@
+// Package prometheus provides an otelcol.receiver.prometheus component.
+package prometheus
+
+import (
+	"context"
+	"os"
+	"regexp"
+	"sort"
+	"sync"
+	"time"
+
+	"github.com/go-kit/log"
+	"github.com/grafana/agent/component"
+	"github.com/grafana/agent/component/otelcol"
+	"github.com/grafana/agent/component/otelcol/internal/fanoutconsumer"
+	"github.com/grafana/agent/component/otelcol/receiver/prometheus/internal"
+	"github.com/grafana/agent/pkg/build"
+	"github.com/grafana/agent/pkg/util/zapadapter"
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/storage"
+	otelcomponent "go.opentelemetry.io/collector/component"
+	otelconfig "go.opentelemetry.io/collector/config"
+	"go.opentelemetry.io/otel/metric"
+	"go.opentelemetry.io/otel/trace"
+)
+
+func init() {
+	component.Register(component.Registration{
+		Name:    "otelcol.receiver.prometheus",
+		Args:    Arguments{},
+		Exports: Exports{},
+
+		Build: func(o component.Options, a component.Arguments) (component.Component, error) {
+			return NewComponent(o, a.(Arguments))
+		},
+	})
+}
+
+// Arguments configures the otelcol.receiver.prometheus component.
+type Arguments struct {
+	ExternalLabels map[string]string `river:"external_labels,attr,optional"`
+
+	// Output configures where to send received data. Required.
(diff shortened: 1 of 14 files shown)


component/prometheus/fanout.go
```diff
@@ -45,6 +46,9 @@ func (f *Fanout) Appender(ctx context.Context) storage.Appender {
 	f.mut.RLock()
 	defer f.mut.RUnlock()
 
+	ctx = scrape.ContextWithMetricMetadataStore(ctx, NoopMetadataStore{})
+	ctx = scrape.ContextWithTarget(ctx, &scrape.Target{})
```
rfratto, line 50: Can we add a note explaining why we need this for now and a TODO to remove it when possible?

component/otelcol/receiver/prometheus/prometheus.go
```diff
@@ -0,0 +1,149 @@
...
+
+		Build: func(o component.Options, a component.Arguments) (component.Component, error) {
+			return NewComponent(o, a.(Arguments))
+		},
+	})
+}
+
+// Arguments configures the otelcol.receiver.prometheus component.
+type Arguments struct {
+	ExternalLabels map[string]string `river:"external_labels,attr,optional"`
```
rfratto, line 41: I feel a little bit weird about supporting injecting custom labels; it seems like you should be using prometheus.relabel for this. 

Can we take this out of the component for now?

rfratto (commented):
👏 Implementation looks good to me, I'll give the approval once the docs are in

component/otelcol/receiver/prometheus/prometheus.go
tpaschalis: Yeah, I'm on it.

rfratto (approved):

karengermond (approved):
LGTM!  One minor update.

PR #2482, opened 2022-11-04
Flow: add tracing instrumentation

### LINKED ISSUE #2215: Add tracing instrumentation to Flow controller
I would like to add tracing instrumentation to the Flow controller, mainly as a way to dogfood #2213. 

To start with, it would be nice to create a new trace every time the controller performs a graph reevaluation. 

This should be coupled with a new [config block][] for tracing. The config block would customize arguments like sampling and specify a set of components to send spans to, something like:

```river
tracing {
  // Sample 10% of traces 
  sampling_fraction = 0.1
  write_to          = [otelcol.exporter.otlp] 
}
```

Fully adding instrumentation has two dependencies: 

* Referencing a Flow component in a config block depends on #2214 being implemented 
* OpenTelemetry components depends on at least #2288 being implemented 

We may choose to completely wait until the dependencies are implemented before adding instrumentation, or it may be valid to take incremental steps and rely on environment variables for sampling/delivering traces until we can provide the full solution.

[config block]: https://grafana.com/docs/agent/latest/flow/reference/config-blocks/

rfratto:
This PR adds a new block called `tracing` which configures Flow-wide tracing support. The `tracing` block supports sending spans to any `otelcol.*` component which exports a consumer:

```river
tracing {
  sampling_percentage = 0.1

  write_to = [otelcol.exporter.otlp.tempo.input]
}

otelcol.exporter.otlp "tempo" {
  client {
    endpoint = "localhost:4317"

    tls {
      insecure = true
    }
  }
}
```

If the tracing block is not provided, or the `write_to` attribute is empty, traces will be discarded.

The Flow controller will give an instance of the tracer to all spanwed components. Components which use this tracer will automatically have an attribute injected to spans denoting the component ID where the span came from.

The tracer is also used globally for any code which currently doesn't support accepting a custom tracer (like Prometheus Remote Write) and for instrumenting the gorilla mux router with traces.

You can try this out yourself by running the Docker Compose environment in example and running Grafana Agent Flow with the example configuration in `./cmd/agent`. 

Closes #2215.


docs/sources/flow/reference/config-blocks/tracing.md
```diff
@@ -0,0 +1,56 @@
+---
+aliases:
+- /docs/agent/latest/flow/reference/config-blocks/tracing
+title: tracing
+weight: 100
+---
+
+# `tracing` block
```
karengermond, line 8:
```suggestion
# tracing block
```

karengermond: The rendering of the code tag in the title looks strange. If you make this change, could you do so for the logging topic, too?

karengermond (approved):
Looks great, one minor edit.

rfratto:
Can you explain what you mean?

mattdurham:
Imagine the below

```
agent {
    log_level = "debug"
    tracing {
        sampling_percentage = 0.1
        write_to = [otelcol.exporter.otlp.tempo.input]
    }
    logging {
        forward_to = logging.sink.file //Implicitly include stdout
    }
}
```

rfratto:
I'm confused, what's the reasoning here? Another option is to add `forward_to` into the existing logging block without putting it under `agent`, right?

mattdurham:
IMO to collapse all the agent-specific items as its own component with subblocks instead of spreading it around to multiple high-level components.

mattdurham (approved):
LGTM, we can discuss the higher level agent component, if needed, later

Issue #2435, opened 2022-10-26
`redis_exporter` fails to connect to Redis when password is provided via `redis_password_file` or environment variable


jesseops:
I can successfully authenticate to redis when putting the password directly in the configuration file, but if I use an environment variable or password file, I get only `Couldn't connect to redis instance`.

I enabled verbose redis logging to check that the connection is initiated:

```
1:M 26 Oct 2022 15:50:56.320 - Accepted 172.18.0.1:59848
1:M 26 Oct 2022 15:50:56.321 - Client closed connection
```
this repeats every 60s and coincides with the message logged by grafana-agent:
```
Oct 26 15:50:56 grafanapoc grafana-agent[3448008]: time="2022-10-26T15:50:56Z" level=error msg="Couldn't connect to redis instance"
```

I have not been able to find any way to increase the redis_exporter log level to get more details. The password file is at least _potentially_ being read, as I can change the configuration to reference a non-existent file and observe that the integration fails to load.

Examples:

## Does work

```file=/etc/grafana-agent.yaml
...
redis_exporter:
    enabled: true
    relabel_configs:
      - replacement: <hostname>
        target_label: instance
    redis_addr: "redis://localhost:6379"
    redis_password: "<password>"
...
```

## Does not work

Example config using environment variable, added `-config.expand-env` to `/etc/default/grafana-agent`:`CUSTOM_ARGS` and appended `REDIS_PASSWORD=<password>` to `/etc/default/grafana-agent`:

```file=/etc/grafana-agent.yaml
...
  redis_exporter:
    enabled: true
    relabel_configs:
      - replacement: <hostname>
        target_label: instance
    redis_addr: "redis://localhost:6379"
    redis_password: $REDIS_PASSWORD
...
```

Example config using a password file:

```file=/etc/grafana-agent.yaml
...
  redis_exporter:
    enabled: true
    relabel_configs:
      - replacement: <hostname>
        target_label: instance
    redis_addr: "redis://localhost:6379"
    redis_password_file: /etc/grafana-agent/redis.json
...
```

We have tried both a simple file containing just the redis password, and the upstream `redis_exporter` formats for the password file:

```file=/etc/grafana-agent/redis.json
<password>
```
and
```file=/etc/grafana-agent/redis.json
{"redis://localhost:6379": "<password>"}
```

spartan0x117:
Hello! I'm taking a look into this, could you try ensuring there is no newline at the end of the simple password file format?

[...]

PR #2484, opened 2022-11-05
Bugfix: Fix issue connecting to redis when specifying password file

spartan0x117:
#### PR Description
Trim whitespace from password file, similar to how [prometheus does it](https://github.com/prometheus/common/blob/8c9cb3fa6d01832ea16937b20ea561eed81abd2f/config/http_config.go#L625-L629).

[#2435](https://github.com/grafana/agent/issues/2435), at least partially. This will fix whitespace being sent as part of password when specifying a password file but does not fix the issue of the environment variable not being used.

#### PR Checklist

- [X] CHANGELOG updated
- [ ] Documentation added
- [X] Tests updated


CHANGELOG.md
```diff
@@ -111,6 +111,8 @@ Main (unreleased)
 
 - Support Bearer tokens for metric remote writes in the Grafana Operator (@jcreixell, @marctc)
 
+- Fix issue when using a password file for `redis_exporter`. (@spartan0x117)
```
rfratto, line 114: Can you update the changelog entry [to describe what the issue was?](https://github.com/grafana/agent/blob/main/docs/developer/contributing.md#changelog-entry-style-tips)

rfratto:
This feels like a breaking change to me, since it means a file containing just the password won't work anymore. Do we understand why the existing code isn't working?

spartan0x117:
Yeah, as-is I realized this would be a breaking change so added a TODO to fallback to the old way (with the slight modification of trimming any newlines at the end, if they exist).

The original issue looks to have been caused when a newline was placed at the end of the file. The newline would be read as a part of the `os.ReadFile` and then would be passed to redis for authentication and most likely fail.

We could solve this either by doing the `re.LoadPwdFile`, falling back to `os.ReadFile` + trimming newlines, _or_ just `os.ReadFile` + trimming newlines. I'm happy to do either way

rfratto:
I'd prefer `os.ReadFile` + trimming newlines, since Prometheus will [also trim newlines when reading password files](https://github.com/prometheus/common/blob/8c9cb3fa6d01832ea16937b20ea561eed81abd2f/config/http_config.go#L629). 

If we want to support the password map from redis_exporter, personally I'd suggest making it something you have to opt into by setting a flag or through a completely distinct `password_map_file` setting. I don't think this is necessary for fixing #2435, but you're more than free to add it if you think supporting the password map file is important :)

CHANGELOG.md
spartan0x117: Done!

rfratto (approved):
LGTM, thanks! 

I'd like to hold off on merging until Monday after the v0.29.0-rc.0 CHANGELOG entries are brought over to main to avoid any confusion with which changes go where.

PR #2496, opened 2022-11-08
Flow: start HTTP server before doing initial load

### LINKED ISSUE #2001: Flow: HTTP server should start immedietly instead of waiting for evaluation

rfratto:
This commit ensures that the HTTP server used by Flow is loaded prior to reading the config file. This allows metrics and pprof data to be collected while the Flow controller is still loading.

One negative to this approach is that it complicates eventual TLS support: if a user intends to protect the server with TLS using a (hypothetical) `server` configuration block, then we should _not_ launch the HTTP server before Flow loads. I'm not sure if it's worth worrying about that right now, though.

Closes #2001


mattdurham (approved):
LGTM

PR #2521, opened 2022-11-08
Flow UI: skip over config nodes in the DAG

rfratto:
If a config node appeared in the references list for a given node, the graph page would fail to load because a reference to a non-existant node would be defined.

This temporarily works around the issue by skipping over DAG edges to a config node.

A longer term fix would be to add support for config blocks in the UI so users can see how they evaluated.

This is behavior which can be observed from v0.29.0, but I'm not sure if it warrants a patch release.


mattdurham (approved):
Good catch

mattdurham (approved):

PR #2532, opened 2022-11-10
flow/internal/controller: Trace graph evaluations

rfratto:
This PR adds traces to graph evaluations:

image

A log line is emitted whenever a partial or full graph evaluation starts including the trace id to assist in finding traces from logs.


mattdurham (approved):
LGTM

PR #2538, opened 2022-11-15
Update agent-loki.yaml to use environment variables in the configuration file

go4real:
Add "-config.expand-env" arguments to use environment variables in the configuration

#### PR Description
Add a command-line flag '-config.expand-env' to use environment variables

Without this flag, the HOSTNAME environment variable is not available even if ${HOSTNAME} is set in the agent configuration file.
As a result, we can't set labels with actual hostnames.

#### PR Checklist

- [x] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ internal API changes are not present.
 Main (unreleased)
 -----------------
 
+
+v0.30.0-rc.0 (2022-12-15)
+--------------------
+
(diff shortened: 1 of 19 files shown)


tpaschalis:
Hey Sungjin! 👋  Apologies, this fell through the cracks.

Could you add a new CHANGELOG entry for this? It will also re-trigger the CI pipeline which seems like it was stuck.

PR #2546, opened 2022-11-22
Flow: add loki.source.file component

### LINKED ISSUE #2503: `loki.source.file`: Retrieves logs from files on disk.
Create a component that receives a list of 'targets' (key-label pairs, or labels), uses a label to infer the file location and starts tailing the files on disk.

The component will read tailed lines and will forward them to other `loki.*` components.

tpaschalis:

#### PR Description
This PR adds a component that expects a list of `targets`, uses the `__path__` label to find the files on disk, and tails these files. The tailed contents are fanned out to the receivers passed in `forward_to`.

Most of the PR size is due to
-  test fixtures
- code copied from Promtail's file target, specifically the Reader interface which is implemented either through a `tailer` or a `decompressor`.
- copying over Promtail's `positions` and `api` package so that we control the way that loki.* components will communicate with each other.

This first iteration goes uses a simplistic approach that whenever new targets are passed, it stops all previous Readers and re-starts them one by one.

There is a subtle concurrency issue when simply deleting Readers, that has their stranded wrapped handlers passing around duplicate messages for a while until they're GC'ed. This approach of explicitly Stop()ing everything is a quick way out, but a more mature solution should test around these behaviors.

#### Open Questions
* Should we periodically re-sync for the target set? Or should this be the responsibility of the upstream discovery component? Right now, if we pass some static targets and the file gets created _after_ that point, it will never get picked up until an Update is called. I'm thinking it could be a configurable parameter.

Fixes #2503 

#### Notes to the Reviewer
The PR adds a `loki.sink.stdout` component that can be used for testing, eg. with a configuration such as the following.

I'm leaving it in for now, but will be removed before the PR is merged. We could revisit the idea in a future PR as an alternative to Promtail's `--dry-run` mode, but it needs more thought put into it.
```
loki.source.file "default" {
  targets = [
    {__path__ = "/tmp/foo.txt", "color" = "pink"},
    {__path__ = "/tmp/bar.txt", "color" = "blue"},
    {__path__ = "/tmp/baz.txt", "color" = "green"},
  ]
  forward_to = [loki.sink.stdout.a.receiver, loki.sink.stdout.c.receiver]
}

loki.sink.stdout "a" {
}

loki.sink.stdout "b" {
}

loki.sink.stdout "c" {
}
```
#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated


component/loki/source/file/decompresser.go
```diff
@@ -0,0 +1,311 @@
+package file
```
rfratto, line 1: This file is from Promtail, right? I think it's ok to merge as-is but I'd be interested in doing some refactoring later to clean it up a bit.

component/loki/source/file/file.go
```diff
@@ -0,0 +1,277 @@
...
+// DefaultArguments defines the default settings for loki.source.file.
+var DefaultArguments = Arguments{}
+
+// UnmarshalRiver implements river.Unmarshaler.
+func (arg *Arguments) UnmarshalRiver(f func(interface{}) error) error {
+	*arg = DefaultArguments
+
+	type args Arguments
+	return f((*args)(arg))
+}
```
rfratto, line 52: nit: you can take this out since there's no defaults to apply. If we want to add defaults later, then we can add it back.

component/loki/source/file/file.go
```diff
@@ -0,0 +1,277 @@
...
+const (
+	pathLabel     = "__path__"
+	filenameLabel = "filename"
+)
+
+// Arguments holds values which are used to configure the loki.source.file
+// component.
+type Arguments struct {
+	Targets   []discovery.Target `river:"targets,attr,optional"`
+	ForwardTo []chan api.Entry   `river:"forward_to,attr,optional"`
```
rfratto, line 40: Are either of these truly optional? They feel required to me.

component/loki/source/file/file.go
```diff
@@ -0,0 +1,277 @@
...
+	"path/filepath"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/go-kit/log/level"
+	"github.com/grafana/agent/component"
+	"github.com/grafana/agent/component/discovery"
+	"github.com/grafana/loki/clients/pkg/promtail/api"
+	"github.com/grafana/loki/clients/pkg/promtail/positions"
```
rfratto, line 16: Do we plan to replace these with agent code (probably a fork of Promtail) at some point in a future PR?

component/loki/source/file/file.go
```diff
@@ -0,0 +1,277 @@
...
+	c.receivers = newArgs.ForwardTo
+
+	oldPaths := make(map[string]struct{})
+
+	// Stop all readers and recreate them below. This avoids the issue we saw
+	// with stranded wrapped handlers staying behind until they were GC'ed and
+	// sending duplicate message to the global handler. It also makes sure that
+	// we update everything with the new labels. Simply zeroing out the
+	// c.readers map did not work correctly to shut down the wrapped handlers
+	// in time.
```
rfratto, line 145: I'm assuming this comment is from Promtail, right? We might want to add a notice indicating as much. 

Recreating readers sounds expensive, so I'm interested in the future optimization, but the "wrapped handlers staying behind until they were GC'd and sending duplicate messages to the global handler" sounds a little scary. We'll probably want to learn more about this before doing the optimization.

component/loki/source/file/file.go
```diff
@@ -0,0 +1,277 @@
...
+		return nil
+	}
+
+	for _, target := range newArgs.Targets {
+		path := target[pathLabel]
+		c.reportSize(path)
+
+		var labels = make(model.LabelSet)
+		for k, v := range target {
+			if strings.HasPrefix(k, "__") {
```
rfratto, line 171: nit: use `model.ReservedLabelPrefix` (from github.com/prometheus/common/model) here instead of `"__"`.

component/loki/source/file/file.go
```diff
@@ -0,0 +1,277 @@
...
+			}
+			labels[model.LabelName(k)] = model.LabelValue(v)
+		}
+
+		handler := api.AddLabelsMiddleware(labels).Wrap(api.NewEntryHandler(c.handler, func() {}))
+
+		reader, err := c.startTailing(path, handler)
+		if err != nil {
+			continue // TODO (@tpaschalis) return err maybe?
+		}
```
rfratto, line 182: Is this the part where you're worried we might want to re-sync targets on an interval? I think it's ok to retry tailing failed targets after a certain period. 

I definitely don't think we should return the error here; we don't want one failed target to cause the sync to fail and disrupt log collection for other files. 

Also: we should make sure the error is _at least_ being logged here, but we'll also want to expose these targets as debug information just like we do for prometheus.scrape so people can see tailing errors.

component/loki/source/file/file_test.go
```diff
@@ -0,0 +1,61 @@
...
+	defer os.RemoveAll(dataPath) // clean up
+
+	opts := component.Options{Logger: l, DataPath: dataPath}
+
+	f, err := os.CreateTemp(dataPath, "example")
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer f.Close()
+	defer os.Remove(f.Name()) // clean up
```
rfratto, line 33: Are these in the right order? Shouldn't files normally be closed before they're removed?

component/loki/source/file/metrics.go
```diff
@@ -0,0 +1,61 @@
...
+	filesActive      prometheus.Gauge
+}
+
+// newMetrics creates a new set of file metrics. If reg is non-nil, the metrics
+// will be registered.
+func newMetrics(reg prometheus.Registerer) *metrics {
+	var m metrics
+	m.reg = reg
+
+	m.readBytes = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
rfratto, line 24: nit: Don't use the Namespace field when creating metrics because it makes it harder to find where in the code metrics are coming from. I'd prefer us only using the name field, like `Name: "loki_source_file_read_bytes_total"`.

component/loki/source/file/reader.go
```diff
@@ -0,0 +1,9 @@
+package file
+
+// Reader contains the set of methods the loki.source.file component uses.
+type Reader interface {
+	Stop()
+	IsRunning() bool
+	Path() string
+	MarkPositionAndSize() error
+}
```
rfratto, line 9: nit: this doesn't need to be exposed, I don't think.

component/loki/source/file/decompresser_test.go
```diff
@@ -0,0 +1,178 @@
...
+		file string
+	}{
+		{
+			name: "2000 lines of log .tar.gz compressed",
+			file: "test_fixtures/short-access.tar.gz",
+		},
+		{
+			name: "100000 lines of log .gz compressed",
+			file: "test_fixtures/long-access.gz",
+		},
```
rfratto, line 57: nit: Can we put these files in `testdata` instead of `test_fixtures`? `testdata` seems more common in Go projects.

rfratto (approved):

rfratto:
Also, while we're still maturing these components, I think we should add a comment indicating when something was copied from Promtail directly. We can remove the comments over time, but I feel it'll be useful early on so we know where context originates.

component/loki/source/file/file.go
tpaschalis: No, this is my own comment when I tried to explain the approach taken here.

After coming back to this today, I think it might be just as easy to compare the previous and new targets, and stop any readers that no longer exist or that had their labels changed, and re-start them accordingly, so I might include it in this PR.

component/loki/source/file/reader.go
tpaschalis: You're right, it doesn't. Fixing this.

component/loki/source/file/decompresser.go
tpaschalis: Yes, I refrained from making any changes for now. As you suggested I will add a comment to all copied files and start seeing how we can simplify things we import.

component/loki/source/file/file.go
tpaschalis: Yes, the plan is to start porting over everything, maybe as well do it right away, right?

The `api` package is a single file which multiple components will reuse and while the `positions` package is a little larger, it's not super complicated by any means.

tpaschalis: I've brought both packages over under `component/common` and am using our own copies for the new component.

component/loki/source/file/file.go
tpaschalis: Good points both about logging and exposing debug information about these targets, I'll take care of it.

Yes, that's the point. For example here, if we receive a target for a __path__ that doesn't yet exist, then we will never know about it until a new Update comes in.

cc @mattdurham  Some of Promtail's SD mechanisms use a `refresh_interval` that re-read the target set. I can see an argument that this should be the responsibility of a discovery component to provide a 'correct' list of targets and update accordingly, but I'd personally lean to adding a relatively longer `refresh_interval` ticker in the Run's select that calls Update every now and then. WDYT?

component/loki/source/file/file.go
```diff
@@ -0,0 +1,277 @@
...
+
+		handler := api.AddLabelsMiddleware(labels).Wrap(api.NewEntryHandler(c.handler, func() {}))
+
+		reader, err := c.startTailing(path, handler)
+		if err != nil {
+			continue
+		}
+
+		c.readers[path] = reader
```
tpaschalis, line 184: cc @mattdurham @rfratto This implementation doesn't allow two targets with the same __path__ but different labels, as the last one would win.

I'll try to find a way to fix this, just wondering if you had any thoughts on allowing/disallowing this.

rfratto: I can't imagine a use case off the top of my head, but it's probably still surprising behavior. I'd suggest we key the map by target labels + path:

```go
c.readers[labels.String() + path] = reader
```

component/loki/source/file/file.go
mattdurham: This sounds similar to local.file and remote.s3 components that refresh and then update. Maybe steal from those?

component/loki/source/file/file.go
mattdurham: IMO if the below works reasonably well, I would go for checking this in and then revisiting in another PR. That way its more focused and gets this PR which imports Promtail out of the way.

tpaschalis:
Changes since the last review

- Added `DebugInfo` method. Now the Flow UI reports for each tailed path, whether the reader is running and how far we've read into it.

```
"targets_info": [{
	path        = "/Users/tpaschalis/GitRepos/agent/test/bar.txt",
	is_running  = true,
	read_offset = 72,
}, {
	path        = "/Users/tpaschalis/GitRepos/agent/test/foo.txt",
	is_running  = true,
	read_offset = 54,
}]
```

- Moved the `api` package that we ported over from Promtail one level up, so we can now refer to `loki.Entry` instead.


As suggested here and in offline conversation, I'll leave a couple of things for the next PR;
a) Keying targets by both paths _and_ labels
b) Fixing the brute-force approach of always stopping and recreating _all_ targets

docs/sources/flow/reference/components/loki.source.file.md
```diff
@@ -0,0 +1,96 @@
...
+## Arguments
+The component starts a new reader for each of the given `targets` and fans out
+log entries to the list of receivers passed in `forward_to`.
+The list of arguments that can be used to configure the block is presented
+below.
+
+Name         | Type                   | Description          | Default | Required
+------------ | ---------------------- | -------------------- | ------- | --------
+`targets`    | `list(map(string))`    | List of files to read from. | | yes
+`forward_to` | `list(chan loki.Entry)` | List of receivers to send log entries to. | | yes
```
tpaschalis, line 33: Should we alias this to a more generic type name like `LogsReceiver`? Or leave it as-is for now?

docs/sources/flow/reference/components/loki.source.file.md
```diff
@@ -0,0 +1,96 @@
...
+loki.source.file "tmpfiles" {
+  targets    = [
+    {__path__ = "/tmp/foo.txt", "color" = "pink"},
+    {__path__ = "/tmp/bar.txt", "color" = "blue"},
+    {__path__ = "/tmp/baz.txt", "color" = "grey"},
+  ] 
+  forward_to = [loki.write.local.receiver]
+}
+
+loki.write "local" {
```
tpaschalis, line 91: The component does not exist yet (but will exist by the next release). Should I remove the example and add it back when it's ready?

docs/sources/flow/reference/components/loki.source.file.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Debug information
+
+`loki.source.file` does not expose any component-specific debug information. ????
+
+## Component behavior
+Each element in the list of `targets` as a set of key-value pairs called
+_labels_.
+The set of targets can either be _static_, or dynamically provided periodically
+by a service discovery component such as `discovery.fileglob`. The special
```
tpaschalis, line 57: The fileglob component does not exist yet (but will exist by the next release). Should I remove the example and add it back when it's ready?

component/loki/source/file/file.go
```diff
@@ -0,0 +1,302 @@
...
+func (c *Component) DebugInfo() interface{} {
+	var res readerDebugInfo
+	for _, reader := range c.readers {
+		path := reader.Path()
+		offset, _ := c.posFile.Get(path)
+		res.TargetsInfo = append(res.TargetsInfo, targetInfo{
+			Path:       path,
+			IsRunning:  reader.IsRunning(),
+			ReadOffset: offset,
+		})
```
rfratto, line 206: Can this include the full set of labels from the target?

rfratto: Also, do we have access to the last time a log line was read to expose it here?

tpaschalis: > The component does not exist yet (but will exist by the next release). Should I remove the example and add it back when it's ready?

It'd be much easier if/when we key readers both by path and labels; for now, we'd have to loop through the targets and match the path, which has a caveat.  
For targets with the same path, when creating the reader the last one wins. Here the first one wins, so it might be misleading.

> Also, do we have access to the last time a log line was read to expose it here?
No we don't, but I can add it.

rfratto: 👍 Returning to decorate this with more information later sounds good to me.

docs/sources/flow/reference/components/loki.source.file.md
rfratto: IMO yes, I'd prefer avoiding referencing unimplemented components in our documentation even if we'll add them before the release (some people may still be reading the `next` version documentation)

component/loki/source/file/file.go
tpaschalis: Ok, will add a TODO for now.

docs/sources/flow/reference/components/loki.source.file.md
rfratto: I'd say use something generic to reduce the chances that someone asks what something like `chan loki.Entry` means.

docs/sources/flow/reference/components/loki.source.file.md
tpaschalis: I've removed it and added a TODO in an HTML comment to add it back once it's merged.

docs/sources/flow/reference/components/loki.source.file.md
```diff
@@ -0,0 +1,109 @@
...
+  targets    = TARGET_LIST
+  forward_to = RECEIVER_LIST
+}
+```
+
+## Arguments
+The component starts a new reader for each of the given `targets` and fans out
+log entries to the list of receivers passed in `forward_to`.
+The list of arguments that can be used to configure the block is presented
+below.
```
karengermond, line 28:
```suggestion

`loki.source.file` supports the following arguments:
```

karengermond: Suggestion for simplifying and making this consistent with other component Arguments sections.

docs/sources/flow/reference/components/loki.source.file.md
```diff
@@ -0,0 +1,109 @@
...
+passed to other `loki.*` components.
+
+The component uses its data path (a directory named after the domain's
+fully qualified name) to store its _positions file_. The positions file is used
+to store read offsets, so that in case of a component or Agent restart,
+`loki.source.file` can pick up tailing from the same spot. 
+
+In case a file is removed from the `targets` list, its positions file entry
+is also removed; that means that when it's added back on, `loki.source.file`
+starts reading it from the beginning.
```
karengermond, line 83:
```suggestion
If a file is removed from the `targets` list, its positions file entry
is also removed. When it's added back on, `loki.source.file`
starts reading it from the beginning.
```

karengermond (approved):
LGTM!

PR #2552, opened 2022-11-24
Adds `phlare.scrape` and `phlare.write` component to support Profiling.

cyriltovena:
#### PR Description

This adds two new components to support profiling.

## `phlare.scrape`

Which scrapes pprof profiles and attach labels based on service discovery.

## `phlare.write`

Which sends pprof profiles and labels attached to Grafana Phlare.

PR is still in the building.

Example of config:

```

logging {
	level  = "debug"
	format = "logfmt"
}

phlare.write "local" {
	endpoint {
		 url = "http://localhost:4100"
	}
	external_labels = {
		"foo" = "bar",
	}
}


phlare.scrape "default" {
	targets    = [
		 {"__address__" = "localhost:4100","app"="phlare"},
		 {"__address__" = "localhost:12345","app"="agent"},
	]
	forward_to = [phlare.write.local.receiver]
	profiling_config {
		pprof_config = {
			fgprof = {
				path = "/debug/fgprof",
				delta = true,
				enabled = true,
			},
			block = { enabled = false },
		}
	}
}
```

#### Notes to the Reviewer

I won't be available in December so please feel free to update and contribute to the PR.

Once this is merged the Phlare teams plan on depending on this component in Phlare to build the all-in-one (including agent) binary mode (ala Prometheus).

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated


rfratto:
That was fast :)

component/pprof/scrape/scrape.go
```diff
@@ -0,0 +1,310 @@
...
+}
+
+type PprofConfig map[string]*PprofProfilingConfig
+
+type PprofProfilingConfig struct {
+	Enabled *bool  `river:"enabled,attr,optional"`
+	Path    string `river:"path,attr,optional"`
+	Delta   bool   `river:"delta,attr,optional"`
+}
+
```
rfratto, line 91: I don't know about making this a map. Labeled blocks might be more appropriate so you have

```river 
profile.memory {
  path = "/debug/pprof/allocs" 
}

profile.block {
  path = "/debug/pprof/block" 
}

...

profile.custom "label" {
  path = ... 
}
```

Using different block names for the recognized profile names allows us to retain the defaults, but we can still support custom ones with the `profile.custom` block. 

Also: I'm not sure why you'd want to pass `enabled` here to the individual profile configs.

component/pprof/scrape/scrape.go
```diff
@@ -0,0 +1,310 @@
...
+		}
+	}
+
+	return scrape.ScraperStatus{TargetStatus: res}
+}
+
+func trueValue() *bool {
+	a := true
+	return &a
+}
```
rfratto, line 310: Use `pointer.Bool` instead of creating a new function

component/phlare/push.v1.proto
```diff
@@ -0,0 +1,38 @@
+syntax = "proto3";
+
+package push.v1;
```
rfratto, line 3: We need to try to wire this up into `go generate` and our build image so we can regenerate the files as needed.

docs/sources/flow/reference/components/pprof.scrape.md
```diff
@@ -0,0 +1,246 @@
...
+configuration.
+
+## Debug information
+
+`pprof.scrape` reports the status of the last scrape for each configured
+scrape job on the component's debug endpoint.
+
+## Debug metrics
+
+* `agent_pprof_fanout_latency` (histogram): Write latency for sending to direct and indirect components.
```
tpaschalis, line 168: I'm not sure if we have a convention around this, but maybe we should start one. Do we prefix all debug metrics coming from Flow components with `agent_`, or do we just use the component name with underscores?

docs/sources/flow/reference/components/phlare.write.md
```diff
@@ -0,0 +1,139 @@
+---
+aliases:
+- /docs/agent/latest/flow/reference/components/phlare.write
+title: phlare.write
+---
+
+# phlare.write
+
+`phlare.write` forward over the network profiles sent from other components to a series of
+user-supplied endpoints. Profiles are sent over the network using [Phlare' Push API](https://grafana.com/oss/phlare/).
```
tpaschalis, line 10:
```suggestion
`phlare.write` receives network profiles from other components and forwards them to a series of user-supplied endpoints using [Phlare' Push API](https://grafana.com/oss/phlare/).
```

component/phlare/write/write.go
```diff
@@ -0,0 +1,263 @@
...
+	return f((*arguments)(r))
+}
+
+// Component is the prometheus.remote_write component.
+type Component struct {
+	opts component.Options
+	cfg  Arguments
+}
+
+// Exports are the set of fields exposed by the phalre.write component.
```
tpaschalis, line 88:
```suggestion
// Exports are the set of fields exposed by the phlare.write component.
```

component/phlare/write/write.go
```diff
@@ -0,0 +1,263 @@
...
+
+// UnmarshalRiver implements river.Unmarshaler.
+func (r *EndpointOptions) UnmarshalRiver(f func(v interface{}) error) error {
+	*r = DefaultEndpointOptions()
+
+	type arguments EndpointOptions
+	return f((*arguments)(r))
+}
+
+// Component is the prometheus.remote_write component.
```
tpaschalis, line 82:
```suggestion
// Component is the phlare.write component.
```

component/phlare/write/write.go
```diff
@@ -0,0 +1,263 @@
...
+type Component struct {
+	opts component.Options
+	cfg  Arguments
```
tpaschalis, line 85: I think in most our components we blindly add a mutex here, but just making sure we don't need it here, right? Every time new arguments are passed we only use them in one goroutine to recreate the receiver.

component/pprof/scrape/scrape.go
```diff
@@ -0,0 +1,310 @@
...
+
+func init() {
+	component.Register(component.Registration{
+		Name: "pprof.scrape",
+		Args: Arguments{},
+
+		Build: func(opts component.Options, args component.Arguments) (component.Component, error) {
+			return New(opts, args.(Arguments))
+		},
+	})
```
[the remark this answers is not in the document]
tpaschalis: +1 to this approach. Then the parts which are needed by both could go under components/common/phlare.

component/pprof/scrape/scrape.go
```diff
@@ -0,0 +1,310 @@
...
+				arg.ProfilingConfig.PprofConfig[pt].Enabled = trueValue()
+			}
+			if arg.ProfilingConfig.PprofConfig[pt].Path == "" {
+				arg.ProfilingConfig.PprofConfig[pt].Path = pc.Path
+			}
+		}
+	}
+
+	if arg.ScrapeTimeout > arg.ScrapeInterval {
+		return fmt.Errorf("scrape timeout must be larger or equal to inverval")
```
tpaschalis, line 155:
```suggestion
		return fmt.Errorf("scrape interval must be larger or equal to inverval")
```

component/pprof/scrape/scrape.go
```diff
@@ -0,0 +1,310 @@
...
+			if c.args.JobName != "" {
+				jobName = c.args.JobName
+			}
+			c.mut.RUnlock()
+			promTargets := c.componentTargetsToProm(jobName, tgs)
+
+			select {
+			case targetSetsChan <- promTargets:
+				level.Debug(c.opts.Logger).Log("msg", "passed new targets to scrape manager")
+			case <-ctx.Done():
```
tpaschalis, line 233: Do we need to return here if the context is done?

component/pprof/scrape/manager.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+	m.mtxScrape.Unlock()
+	wg.Wait()
+}
+
+// ApplyConfig resets the manager's target providers and job configurations as defined by the new cfg.
+func (m *Manager) ApplyConfig(cfg Arguments) error {
+	m.mtxScrape.Lock()
+	defer m.mtxScrape.Unlock()
+	if reflect.DeepEqual(m.config, cfg) {
```
tpaschalis, line 114: Same as Robert's comment above, that Update will only be called with the updated Arguments, even if it's in a different package.

component/pprof/scrape/scrape.go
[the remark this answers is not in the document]
rfratto: We discussed offline and for now we're going to name the component `phlare.scrape`.

component/pprof/scrape/scrape.go
rfratto: We discussed this offline and we're going to try to do these as blocks, keeping in mind the need for defaults for known profile endpoints and the ability to add custom profile endpoints. 

We will be conscious about the ability to convert it into a map so it fits well with the existing code.

docs/sources/flow/reference/components/pprof.scrape.md
rfratto: We discussed this offline, and for now we're going to continue the convention of naming metrics based on the component name, and without the `agent_` prefix, so that this would be called `phlare_scrape_fanout_latency` (assuming pprof.scrape is renamed to phlare.scrape). 

If there's a name collision in dashboards, we can use `and` to filter down the set of metrics to the applications we want:

```
go_goroutines{cluster="$cluster",namespace="$namespace",instance=~"$instance"}
and on(instance)
agent_build_info{cluster="$cluster",namespace="$namespace",instance=~"$instance"}
```

component/pprof/scrape/scrape.go
rfratto: It doesn't necessarily need to because it'll return on the next loop, but I think for clarity it would be good to return here anyway.

component/phlare/push.v1.proto
```diff
@@ -0,0 +1,38 @@
+syntax = "proto3";
```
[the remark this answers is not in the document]
rfratto: We discussed this offline. Since the intent is to be able to share the scrape and push code between Grafana Agent and Phlare, _and_ I am concerned about forcing our module dependency hell on Phlare, we should be careful about how we do this. 

It's likely we're going to eventually have the scrape and push code in its own module for sharing. That module cannot import any Grafana Agent package (otherwise it introduces the problem mentioned above), so we will do the following:

1. Create an `api` package which contains the scrape manager and client code to push profiles to Phlare. This `api` package will eventually be its own module, but will remain in the base module for now. 
2. We will suggest that protobufs move to a shared location / yet another module that clients can import from  to avoid a circular dependency between Grafana Agent and Phlare, but this isn't a blocker for this PR.

docs/sources/flow/reference/components/pprof.scrape.md
```diff
@@ -0,0 +1,246 @@
...
+
+## Scraping behavior
+
+The `pprof.scrape` component borrows the scraping behavior of Prometheus.
+Prometheus, and by extent this component, uses a pull model for scraping
+profiles from a given set of _targets_.
+Each scrape target is defined as a set of key-value pairs called _labels_.
```
karengermond, line 175:
```suggestion

Each scrape target is defined as a set of key-value pairs called _labels_.
```

docs/sources/flow/reference/components/pprof.scrape.md
```diff
@@ -0,0 +1,246 @@
...
+        enabled = true,
+      },
+      block = { enabled = false },
+      mutex = { enabled = false },
+    }
+  }
+}
+```
+
+Here's the the endpoints that are being scraped every 15 seconds:
```
karengermond, line 235:
```suggestion
Here are the the endpoints that are being scraped every 15 seconds:
```

karengermond (approved):
LGTM -- a few suggestions.

component/phlare/push.v1.proto
cyriltovena: Yeah I think we can create an issue for this. Phlare use buff but I don't want to force anything, this was just to get a first version.

component/pprof/scrape/scrape.go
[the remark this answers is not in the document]
cyriltovena: pprof scraping is actually supported by other backends than Phlare. I think it's more open, but I'm also not super opinionated.

component/pprof/scrape/scrape.go
cyriltovena: This was a bug fixing it.

docs/sources/flow/reference/components/pprof.scrape.md
cyriltovena: This is resolved.

component/pprof/scrape/scrape.go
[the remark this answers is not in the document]
cyriltovena: I've renamed it as requested.

component/phlare/push.v1.proto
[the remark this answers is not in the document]
cyriltovena: That sounds like more my problem. 

Grafana also has to generate the[ query side ](https://github.com/grafana/grafana/tree/0845ac2f53ca17cf698341b382af5262335e295c/pkg/tsdb/phlare/gen/querier/v1)and I've been copying code so far, but I think it's time to have a Phlare API module as you suggest where protobuf generation happens.

I'll investigate it.

component/pprof/scrape/scrape.go
cyriltovena: Sounds good, let me know if I can help.

There are a couple of interesting issues related to this already in Phlare, but the most noteworthy https://github.com/grafana/phlare/issues/456. This configuration varies from one application to another.

We can also consider fgprof not custom but disable by default. Having the ability to add custom profile types is nice to have.

I suspect in the future we will add more per profile configuration, for example, to exclude some sample types (just send sample value, not sample count).

component/phlare/push.v1.proto
cyriltovena: After thinking a bit more I'm planing to remove this from here.

component/phlare/push.v1.proto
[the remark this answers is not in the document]
cyriltovena: Trying this here https://github.com/grafana/phlare/pull/466

component/phlare/push.v1.proto
cyriltovena: This is removed.

component/phlare/write/write.go
cyriltovena: Checking .....

cyriltovena: Yeah doesn't seems necessary to me.

rfratto:
@cyriltovena Is there anything left you were planning on doing? If not, I'll address changing the profiling configs to blocks as suggested [here](https://github.com/grafana/agent/pull/2552#discussion_r1037468309) and add doc comments on exported things wherever they're missing.

cyriltovena:
Yeah I need to send a commit to fix the breaking change introduce by the API module change

cyriltovena:
> Yeah I need to send a commit to fix the breaking change introduce by the API module change

LGTM now !

cyriltovena:
Not sure why the CI is flaky.

component/pprof/scrape/scrape.go
rfratto: Done:

```river 
phlare.scrape "local" {
  targets    = [
    {"__address__" = "localhost:4100", "app"="phlare"},
    {"__address__" = "localhost:12345", "app"="agent"},
  ]
  forward_to = [phlare.write.local.receiver]
  profiling_config {
    profile.fgprof {
      enabled = true
    }
    profile.block {
      enabled = false
    }
    profile.mutex {
      enabled = false
    }
  }
}
```

rfratto (approved):
LGTM! This needs doc comments in a few places, but I'll add those later in #2676.

PR #2574, opened 2022-11-30
Support exemplars in Prometheus component pipelines

### LINKED ISSUE #2420: Grafana Agent Flow doesn't support Prometheus exemplars
Grafana Agent Flow currently doesn't support Prometheus exemplars. 

It looks like `prometheus.remote_write` will properly send exemplars if they were written to the WAL, but the current FlowMetric abstraction means that no exemplars get written to the WAL. 

See #2037 for one proposal which can fix this issue.

rfratto:
This commit allows exemplars to be passed around in Prometheus component pipelines. To support this change, the prometheus.Interceptor struct was extended to support intercepting any call to the Appendable interface.

Support for metadata is also added, though that code path is currently not invoked in our usage of the Prometheus scraper.

Fixes #2420.

### DIFF
--- a/component/prometheus/interceptor.go
+++ b/component/prometheus/interceptor.go
@@ -13,23 +13,60 @@ import (
 // getting data. Interceptor should not be modified once created. All callback
 // fields are optional.
 type Interceptor struct {
-	OnAppend         func(ref storage.SeriesRef, l labels.Labels, t int64, v float64, next storage.Appender) (storage.SeriesRef, error)
-	OnAppendExemplar func(ref storage.SeriesRef, l labels.Labels, e exemplar.Exemplar, next storage.Appender) (storage.SeriesRef, error)
-	OnUpdateMetadata func(ref storage.SeriesRef, l labels.Labels, m metadata.Metadata, next storage.Appender) (storage.SeriesRef, error)
+	onAppend         func(ref storage.SeriesRef, l labels.Labels, t int64, v float64, next storage.Appender) (storage.SeriesRef, error)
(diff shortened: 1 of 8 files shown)


component/prometheus/interceptor.go
```diff
@@ -2,81 +2,55 @@ package prometheus
...
-	// intercept allows one to intercept the series before it fans out to make any changes. If labels.Labels returns nil the series is not propagated.
-	// Intercept shouuld be thread safe and can be called across appenders.
-	intercept Intercept
-	// next is where to send the next command.
-	next storage.Appendable
-
-	// ComponentID is what component this belongs to.
-	componentID string
-}
+	OnAppend         func(ref storage.SeriesRef, l labels.Labels, t int64, v float64, next storage.Appender) (storage.SeriesRef, error)
```
mattdurham, line 16: Would it be safer to hide this in a constructor, and use wrapper functions to call the underlying funcs?

rfratto: Maybe, I was hesitant to introduce a constructor that took 4 parameters, especially since that list will keep growing. 

If you think it needs to be a constructor, I wouldn't be opposed to the functional option pattern like `NewInterceptor(next, WithAppendHook(fn), WithMetadataHook(fn))` though.

mattdurham: Realistically it's unlikely to get misused but if the functional pattern is easy and quick to do then that avoids the issue entirely.

rfratto: I implemented the functional option pattern, WDYT?

mattdurham (approved):
LGTM

PR #2575, opened 2022-11-30
Operator: Allow disabling http2

captncraig:
- Update prometheus apis to get latest CRDs and clients
- Manually update crd yamls in `./production`
- Add new field to jsonnet for PodMonitor and ServiceMonitor


marctc (approved):
LGTM

PR #2577, opened 2022-12-01
 Flow: add loki.relabel component

### LINKED ISSUE #2514: `loki.transform.relabel`: Runs relabeling rules on log entries labels.
Create a component that can run relabeling rules on the labels of a log entry, after it has been discovered and tailed.

tpaschalis:

#### PR Description
The PR adds a `loki.relabel` component that receives log entries and rewrites their label set using relabeling rules. This functionality was not available in Promtail, where relabeling would only happen for targets, and any label manipulation after the log had been scraped had to be done with stages pipelines, so it needs some back-and-forth.

We didn't have to bring any new Promtail code for this one, for now.

Fixes #2514 

#### Notes to the Reviewer
* Since relabeling can be resource-intensive, I've introduced a cache field in the Component struct. Since the relabeling happens only on the same goroutine as the component is running, there was no need to protect it with a mutex.
* As the log entries do not provide any indication of 'staleness', like Prometheus metric do, we'd have to choose another way to evict items from the cache, so it doesn't keep growing forever. I've opted to use [hashicorp/golang-lru](https://github.com/hashicorp/golang-lru) which is already on our dependencies list, and comes with an [MPL-2.0 license](https://github.com/hashicorp/golang-lru/blob/master/LICENSE).

#### Open Questions
* Do we want the allow the user to configure the cache size? If we take Loki's default label limits (30 labels, up to 1024 chars for name and 2048 chars for value), the current `maxCacheSize = 10_000` value,  caps to ~1GB of memory in use.
* Is it okay to `import "reflect"` in order to detect when new relabel configs are passed? Should we use the fingerprinting idea on Flow's relabel structs to make the comparison that way?
* Should we document anything about the 'footgun' factor here? Like changing labels may result in logs being present in different 'streams'?
 
#### PR Checklist

- [X] CHANGELOG updated
- [X] Documentation added
- [X] Tests updated

### DIFF
--- a/component/loki/relabel/relabel.go
+++ b/component/loki/relabel/relabel.go
@@ -1,0 +1,214 @@
+package relabel
+
+import (
+	"context"
+	"reflect"
+	"sync"
+
+	"github.com/go-kit/log/level"
+	"github.com/grafana/agent/component"
+	"github.com/grafana/agent/component/common/loki"
+	flow_relabel "github.com/grafana/agent/component/common/relabel"
+	lru "github.com/hashicorp/golang-lru"
+	prometheus_client "github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/model/relabel"
+)
+
+func init() {
+	component.Register(component.Registration{
+		Name:    "loki.relabel",
+		Args:    Arguments{},
+		Exports: Exports{},
+		Build: func(opts component.Options, args component.Arguments) (component.Component, error) {
+			return New(opts, args.(Arguments))
+		},
+	})
+}
+
+// Arguments holds values which are used to configure the loki.relabel
+// component.
+type Arguments struct {
+	// Where the relabelled metrics should be forwarded to.
+	ForwardTo []loki.LogsReceiver `river:"forward_to,attr"`
+
+	// The relabelling rules to apply to each log entry before it's forwarded.
+	RelabelConfigs []*flow_relabel.Config `river:"rule,block,optional"`
+}
+
+// Exports holds values which are exported by the loki.relabel component.
+type Exports struct {
+	Receiver loki.LogsReceiver `river:"receiver,attr"`
+}
+
+// Component implements the loki.relabel component.
+type Component struct {
+	mut      sync.RWMutex
+	opts     component.Options
+	rcs      []*relabel.Config
+	receiver loki.LogsReceiver
+	fanout   []loki.LogsReceiver
+
+	entriesProcessed prometheus_client.Counter
+	entriesOutgoing  prometheus_client.Counter
+	cacheHits        prometheus_client.Counter
+	cacheMisses      prometheus_client.Counter
+	cacheSize        prometheus_client.Gauge
+
+	cache *lru.Cache
+}
+
+var (
+	_ component.Component = (*Component)(nil)
+
+	maxCacheSize = 10_000 // TODO(@tpaschalis) Do we make this configurable?
+)
+
+// New creates a new loki.relabel component.
+func New(o component.Options, args Arguments) (*Component, error) {
+	cache, err := lru.New(maxCacheSize)
+	if err != nil {
+		return nil, err
+	}
+
+	c := &Component{
+		opts:  o,
+		cache: cache,
+	}
+	c.entriesProcessed = prometheus_client.NewCounter(prometheus_client.CounterOpts{
+		Name: "loki_relabel_entries_processed",
+		Help: "Total number of log entries processed",
+	})
+	c.entriesOutgoing = prometheus_client.NewCounter(prometheus_client.CounterOpts{
+		Name: "loki_relabel_entries_written",
+		Help: "Total number of log entries forwarded",
+	})
+	c.cacheMisses = prometheus_client.NewCounter(prometheus_client.CounterOpts{
+		Name: "loki_relabel_cache_misses",
+		Help: "Total number of cache misses",
+	})
+	c.cacheHits = prometheus_client.NewCounter(prometheus_client.CounterOpts{
+		Name: "loki_relabel_cache_hits",
+		Help: "Total number of cache hits",
+	})
+	c.cacheSize = prometheus_client.NewGauge(prometheus_client.GaugeOpts{
+		Name: "loki_relabel_cache_size",
+		Help: "Total size of relabel cache",
+	})
+
+	for _, metric := range []prometheus_client.Collector{c.entriesProcessed, c.entriesOutgoing, c.cacheMisses, c.cacheHits, c.cacheSize} {
+		err := o.Registerer.Register(metric)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// Create and immediately export the receiver which remains the same for
+	// the component's lifetime.
+	c.receiver = make(loki.LogsReceiver)
+	o.OnStateChange(Exports{Receiver: c.receiver})
+
+	// Call to Update() to set the relabelling rules once at the start.
+	if err := c.Update(args); err != nil {
+		return nil, err
+	}
+
+	return c, nil
+}
+
+// Run implements component.Component.
+func (c *Component) Run(ctx context.Context) error {
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case entry := <-c.receiver:
+			lbls := c.relabel(entry)
+			if len(lbls) == 0 {
+				level.Debug(c.opts.Logger).Log("msg", "dropping entry after relabeling", "labels", entry.Labels.String())
+				continue
+			}
+
+			c.entriesOutgoing.Inc()
+			entry.Labels = lbls
+			for _, f := range c.fanout {
+				select {
+				case <-ctx.Done():
+					return nil
+				case f <- entry:
+				}
+			}
+		}
+	}
+}
+
+func (c *Component) relabel(e loki.Entry) model.LabelSet {
+	c.entriesProcessed.Inc()
+	hash := e.Labels.Fingerprint().String()
+	found, ok := c.cache.Get(hash)
+	if ok {
+		c.cacheHits.Inc()
+		return found.(model.LabelSet)
+	}
+
+	c.cacheMisses.Inc()
+
+	// TODO(@tpaschalis) It's unfortunate how we have to cast back and forth
+	// between model.LabelSet (map) and labels.Labels (slice). Promtail does
+	// not have this issue as relabel config rules are only applied to targets.
+	// Do we want to use labels.Labels in loki.Entry instead?
+	var lbls labels.Labels
+	for k, v := range e.Labels {
+		lbls = append(lbls, labels.Label{
+			Name:  string(k),
+			Value: string(v),
+		})
+	}
+	lbls = relabel.Process(lbls, c.rcs...)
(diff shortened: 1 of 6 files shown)


component/loki/relabel/relabel.go
```diff
@@ -0,0 +1,214 @@
...
+		Help: "Total number of cache misses",
+	})
+	c.cacheHits = prometheus_client.NewCounter(prometheus_client.CounterOpts{
+		Name: "loki_relabel_cache_hits",
+		Help: "Total number of cache hits",
+	})
+	c.cacheSize = prometheus_client.NewGauge(prometheus_client.GaugeOpts{
+		Name: "loki_relabel_cache_size",
+		Help: "Total size of relabel cache",
+	})
```
rfratto, line 98: nit: to clean this up a little bit, what do you think about declaring a metrics struct and putting it & these in metrics.go? That's a pattern we sometimes use in the codebase.

component/loki/relabel/relabel.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+func (c *Component) relabel(e loki.Entry) model.LabelSet {
+	c.entriesProcessed.Inc()
+	hash := e.Labels.Fingerprint().String()
+	found, ok := c.cache.Get(hash)
+	if ok {
+		c.cacheHits.Inc()
+		return found.(model.LabelSet)
+	}
```
rfratto, line 153: What if there's a hash collision?

component/loki/relabel/relabel.go
```diff
@@ -0,0 +1,214 @@
...
+
+	c.cacheMisses.Inc()
+
+	// TODO(@tpaschalis) It's unfortunate how we have to cast back and forth
+	// between model.LabelSet (map) and labels.Labels (slice). Promtail does
+	// not have this issue as relabel config rules are only applied to targets.
+	// Do we want to use labels.Labels in loki.Entry instead?
```
rfratto, line 160: IMO we should switch everything to use labels.Labels as soon as we can. We should probably find out if there's a reason Promtail uses model.LabelSet, though.

component/loki/relabel/relabel.go
```diff
@@ -0,0 +1,214 @@
...
+	newArgs := args.(Arguments)
+	newRCS := flow_relabel.ComponentToPromRelabelConfigs(newArgs.RelabelConfigs)
+	if relabelingChanged(c.rcs, newRCS) {
+		level.Debug(c.opts.Logger).Log("msg", "received new relabel configs, purging cache")
+		c.cache.Purge()
+		c.cacheSize.Set(0)
+	}
+	c.rcs = newRCS
+	c.cache.Purge()
+	c.fanout = newArgs.ForwardTo
```
rfratto, line 195: We're always purging the cache here, was that a mistake?

component/loki/relabel/relabel.go
```diff
@@ -0,0 +1,214 @@
...
+
+	return nil
+}
+
+// TODO(@tpaschalis) This is an attempt to not purge the cache if the
+// relabeling rules are the same. One way to go about it would be to use the
+// fingerprinting idea on Flow's relabel structs, but this seemed more
+// straightforward. Are we okay with importing reflect here?
```
rfratto, line 203: > Are we ok with importing reflect here? 

Yeah, this is fine IMO.

component/loki/relabel/relabel_test.go
```diff
@@ -0,0 +1,100 @@
+package relabel
```
rfratto, line 1: Can you write a benchmark for this component, please?

docs/sources/flow/reference/components/loki.relabel.md
```diff
@@ -0,0 +1,182 @@
...
+}
+```
+
+## Arguments
+
+The following arguments are supported:
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`forward_to` | `list(receiver)` | Where the log entries should be forwarded to, after relabeling takes place. | | yes
```
rfratto, line 49: You might be able to make this a bit shorter, WDYT? 

```suggestion
`forward_to` | `list(receiver)` | Where to forward log entries after relabeling. | | yes
```

docs/sources/flow/reference/components/loki.relabel.md
```diff
@@ -0,0 +1,182 @@
...
+## Debug metrics
+
+
+* `loki_relabel_entries_processed` (counter): Total number of log entries processed.
+* `loki_relabel_entries_written` (counter): Total number of log entries forwarded.
+* `loki_relabel_cache_misses` (counter): Total number of cache misses.
+* `loki_relabel_cache_hits` (counter): Total number of cache hits.
+* `loki_relabel_cache_size` (gauge): Total size of relabel cache.
+
+## Example
```
rfratto, line 91: This reads like more of a tutorial than an example to me. What do you think about making this briefer, and saving the detailed prose for a more fleshed out logging tutorial?

rfratto:
> Do we want the allow the user to configure the cache size? If we take Loki's default label limits (30 labels, up to 1024 chars for name and 2048 chars for value), the current maxCacheSize = 10_000 value, caps to ~1GB of memory in use.

Yeah, I think making it configurable makes sense to allow users to choose between memory and CPU.

rfratto:
> Should we document anything about the 'footgun' factor here? Like changing labels may result in logs being present in different 'streams'?

I'm not sure I understand what the issue is, but if you think improper relabeling can cause unexpected issues, we should definitely document it :)

docs/sources/flow/reference/components/loki.relabel.md
```diff
@@ -0,0 +1,182 @@
...
+its receiver by applying one or more relabeling `rule`s and forwards the
+results to the list of receivers in the component's arguments.
+
+<!-- TODO(@tpaschalis) Add note about loki.transform
+  To manipulate the log entry itself, you can look at the loki.transform
+  component which allows you to run one or more 'stages' on the log entries.
+ -->
+
+If no labels remain after the relabeling rules are applied, then the log entry
+are dropped.
```
karengermond, line 19:
```suggestion
If no labels remain after the relabeling rules are applied, then the log entries
are dropped.
```

docs/sources/flow/reference/components/loki.relabel.md
karengermond: Nice edit, Robert!   I'm not needed anymore.  :-(

docs/sources/flow/reference/components/loki.relabel.md
karengermond: That sounds like a good idea to me.

docs/sources/flow/reference/components/loki.relabel.md
```diff
@@ -0,0 +1,182 @@
...
+* `loki_relabel_cache_hits` (counter): Total number of cache hits.
+* `loki_relabel_cache_size` (gauge): Total size of relabel cache.
+
+## Example
+
+One of Loki's core features is that, unlike other logging systems, it does not
+index the log message itself, but works off indexing metadata about the
+received log streams. These metadata are stored as key-value pairs called
+_labels_, much like Prometheus metric labels. Having a small and efficient
+index, along being able to store the log data itself in highly compressed
```
karengermond, line 97:
```suggestion
index, along with being able to store the log data itself in highly compressed
```

karengermond (approved):
LGTM!  I didn't edit the last example thoroughly since it will likely turn into a tutorial.

component/loki/relabel/relabel.go
tpaschalis: Unfortunately, then we'd just get wrong relabeling results here.

From what I see, it's the same Fingerprint that Loki uses internally eg. in their ChunkRefs, or the Distributor hashes, taken from Prometheus. It uses the 64-bit variant of FNV-1A. I couldn't find a theoretical collision rate for the algorithm, but [this](https://www.strchr.com/hash_functions#:~:text=Georgi%20%27Sanmayce%27\)%3A-,Core%20i5%20processor,-Wikipedia) source talks about 2M collisions in the set of the words in english wikipedia (~12M words), which is not really an impressive performance...

Our cache sizes shouldn't reach _those_ sizes (we'd probably have more pressing problems, much earlier on), but do we have any real-world feedback of hashing with FNV-1A?

docs/sources/flow/reference/components/loki.relabel.md
tpaschalis: 🤦 thanks!

component/loki/relabel/relabel.go
rfratto: The [hashmap code in Prometheus](https://github.com/prometheus/prometheus/blob/0372e259baf014bbade3134fd79bcdfd8cbdef2c/tsdb/agent/series.go#L48-L61) handles collisions by allowing each entry to be a list of label sets. Whenever there's a hashmap lookup, every label set will be compared against the input to check if it matches.

IMO, collisions are unlikely until they happen, and then they're a huge pain to track down. My gut feeling is even if it's a little extra work to implement (and a bit slower on CPU), protecting against collisions with the approach Prometheus uses will probably save us future pain.

docs/sources/flow/reference/components/loki.relabel.md
tpaschalis: Thanks for the feedback! I think that an end-to-end logging tutorial will be worth putting some time into.

For now, I've shortened the example to be like the rest of the components!

component/loki/relabel/relabel.go
tpaschalis: My idea was to get an end-to-end logging pipeline ready and then comparing if there's any big difference between the models.LabelSet and labels.Labels performance-wise, but I wouldn't think so for the number of labels usually in-use.

I'll reach out and see if there was a reason for this choice.

component/loki/relabel/relabel.go
tpaschalis: You're right, I just bluntly copy/pasted from another component. Let me tidy this up :)

component/loki/relabel/relabel.go
tpaschalis: I've gone ahead with this, and also wrote a test with and without collisions to see that the implementation works as intended.

component/loki/relabel/relabel_test.go
rfratto: That's fine, I just want us to start having benchmarks to make it easier to track performance over time 👍 We don't need to worry about if 5ms per series is too high or not for now.

component/loki/relabel/relabel.go
```diff
@@ -0,0 +1,235 @@
...
+	if !found {
+		// We've never seen this hash; initialize it in the cache with a new
+		// cacheItem and return the relabeled result.
+		c.metrics.cacheMisses.Inc()
+		relabeled := c.process(e)
+
+		c.cache.Add(hash, []cacheItem{{e.Labels, relabeled}})
+		c.metrics.cacheSize.Set(float64(c.cache.Len()))
+
+		return relabeled
```
rfratto, line 195: Is there a way we can dedupe this with the code below? They're very similar except for the append.

rfratto (approved):
Nice work, LGTM!

component/loki/relabel/relabel.go
tpaschalis: Yes, I've gone ahead with a small refactor, hopefully it's easier to follow now.

PR #2583, opened 2022-12-02
[RFC] Multi-tenancy support in Flow

### LINKED ISSUE #521: Flow: Add component for multi-tenant remote_write support
The Prometheus remote_write protocol doesn't have the notion of multi-tenancy itself. For this reason different backends offer use various methods to enable multi-tenancy, most often using an `X-Scope-OrgID` header, or labels with a special meaning.

When using a `prometheus.remote_write` component, the Prometheus queue_manager reads WAL segments sequentially and enqueues metrics opportunistically to be batched off as remote_write requests. This behaviour offers no fine-grained control at the request level. Our current suggestion to users is that they add an extra header to their [endpoint block](https://grafana.com/docs/agent/latest/flow/reference/components/prometheus.remote_write/#endpoint-block), but these per-endpoint headers are static, and we don't support write_relabel_config filtering in Flow yet.

The new component would act as a remote_write middleware, receiving metrics from upstream components, extracting a given label (say `tenant`), and batching timeseries grouped by this label value.

It would then send discrete remote_write requests for these batches while also adding the correct `X-Scope-OrgID` header for each one.

### Notes
* This was inspired by the conversation in grafana/agent#1821 and existing solutions like [cortex-tenant](https://github.com/blind-oracle/cortex-tenant).
* I don't believe it would be easy to try and shoehorn this behaviour into the current `prometheus.remote_write` component.
* Using this component would entail some performance penalty which should be measured and explicitly documented.

ptodev:
Proposes several approaches to support (or not) multi-tenancy within Flow. Solves grafana/alloy#521.

### DIFF
--- a/docs/rfcs/0007-multi-tenant-remote-write-in-flow.md
+++ b/docs/rfcs/0007-multi-tenant-remote-write-in-flow.md
@@ -1,0 +1,212 @@
+# Multi-tenancy support in Flow
+
+* Date: 2022-12-01
+* Author: Paulin Todev (@todevpau)
+* PR: [grafana/agent#2583](https://github.com/grafana/agent/pull/2583)
+* Status: Draft
+
+## Background
+
+Grafana databases (e.g. Mimir/Loki/Tempo) support the notion of "tenancy". In GEx, a "tenant" is both a piece of metadata and an authentication/trust boundary. It can be used to manage rate limits, distribution of data across the cluster, limiting the impact of expensive queries, etc. 
+
+In order for the Grafana databases to know the tenant of a series, they expect to see the tenant ID in the "X-Scope-OrgID" HTTP header of the remote write request. In GEx this is done via a [Gateway](https://grafana.com/docs/enterprise-traces/latest/gateway/) component which sits between the Agent and the database. The gateway assigns tenancy information based on the security credentials which the Agent used to connect to it.
+
+The above works fine for GEx. However, there are also Grafana customers who use on-prem hosted databases. They are not able to do multi-tenancy for two reasons:
+* The aforementioned "gateway" component is not open source
+* The agent and the database could be communicating via a non-secure connection. Hence, even if a Gateway is used, deriving tenancy information from authentication details would not be an option.
+
+## Problem statement
+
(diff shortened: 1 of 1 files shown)


docs/rfcs/0007-multi-tenant-remote-write-in-flow.md
```diff
@@ -0,0 +1,212 @@
...
+* Good security. Allowing the Agent to set the tenant could mean that an Agent is configured to send data as the wrong tenant (either due to malicious activity or by mistake). Setting tenant information within the Gateway, a server-side component, would be more restricted and secure.
+
+#### Cons
+
+* It is extra work for customers to introduce and operate an extra component.
+* Development work on the Gateway is required in order to open source it and make it more configurable.
+
+### Solution 2 - A generic "sharding" configuration for the Agent remote write
+
+The [```prometheus.remote_write```](https://grafana.com/docs/agent/latest/flow/reference/components/prometheus.remote_write/) Flow component could be expanded to be able to shard the writes into different remote writes under the hood. Each of these shards would have its own write-ahead log (WAL). Also, each shard would be able to remote write to a different endpoint address and with different HTTP header information.
```
mattdurham, line 47: Note: Caching of series refs is per WAL, so if tenants have a lot of overlap with series, this can lead to ballooning memory usage.

mattdurham:
There is also https://github.com/blind-oracle/cortex-tenant which seems to fit the needs?

docs/rfcs/0007-multi-tenant-remote-write-in-flow.md
ptodev: Good point, thank you, I added this to the "cons" list. Although not sure if it's a "con" it's more like just something to keep in mind.

ptodev:
> There is also https://github.com/blind-oracle/cortex-tenant which seems to fit the needs?

Of course! How could I forget 😆 Thanks for reminding me, I added that too.

ptodev:
I'll close this PR because it has been stale for a really long time... I'm not sure what a good multi tenancy solution would look like and I really lack the time to look into it.

PR #2588, opened 2022-12-05
Force directconnect for mongodb_exporter

v-zhuravlev:
#### PR Description
By default, `mongodb_exporter` enforce [DirectConnect](https://www.mongodb.com/docs/drivers/go/current/fundamentals/connection/#direct-connection) to mongodb replicaset node:
https://github.com/percona/mongodb_exporter/blob/main/main.go#L42

But AFAIK, `mongodb_exporter` integration set DirectConnect to `false` by default. In this case, even if you point mongodb mongodb_uri to secondary node, it will actually use it only to discover primary node's URI and connect to it instead. 

Such approach (DirectConnect: false) results in situation, where secondary nodes return primary node metrics (including replset state) instead of their own, messing up Replication state picture on dashboards.

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/integrations/mongodb_exporter/mongodb_exporter.go
+++ b/pkg/integrations/mongodb_exporter/mongodb_exporter.go
@@ -63,6 +63,7 @@ func New(logger log.Logger, c *Config) (integrations.Integration, error) {
 		// configurable in the future.
 		CompatibleMode: true,
 		CollectAll:     true,
+		DirectConnect:  true,
 	})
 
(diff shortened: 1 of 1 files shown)


gaantunes (approved):
Ooh thanks for looking into it, I did not know that ws the situation. Nice catch!

rfratto:
Can you update the CHANGELOG before this gets merged?

v-zhuravlev:
> Can you update the CHANGELOG before this gets merged?

Added.

PR #2589, opened 2022-12-05
Flow: add loki.process component

### LINKED ISSUE #2515: `loki.process`: Use Promtail stages to transform log labels and entries
Create a component that encapsulates the functionality of Promtail's [stages](https://grafana.com/docs/loki/latest/clients/promtail/stages/) to parse/transform/filter incoming log entries. 

The component reads runs the given stages and forwards the log entries to other `loki.*` components.

The proposed configuration approach that is sketched below is a little repetitive, but allows running the 'stages' blocks in top-down order, without the need for dedicated River features.

```river
loki.transform.stages "example" {
  stage {
    match { ... } 
  }

  stage {
    json { ... } 
  }

  forward_to = [ /* */ ]
}
```

tpaschalis:

#### PR Description

The PR adds a `loki.process` component that receives log entries and runs one or more [processing stages](https://grafana.com/docs/loki/latest/clients/promtail/stages/) on them.

The stages are brought over from Loki's `logentry` package. The packge's total size is nearly ~10k lines of code split between 21 stages, which are not as straightforward as copy-and-pasting to port over.

For this reason, we've only brought in two stages for now, `json` and `labels`, and the plan is to port the rest of them gradually.

The changes while between the files in _this_ PR and the stages as they exist on the current version of Loki we're importing can be seen on the following link, as the PR size may make it harder to view at a glance.
https://github.com/tpaschalis/loki/compare/stages-baseline-0...tpaschalis:loki:stages-baseline-1

I've opted to only bring in entire files and _not_ parts of them, and comment out tests that cannot be run until the rest of the components are ported over, to minimize conflicts later on. Let me know if you think differently.

Fixes #2515 

#### Notes to the Reviewer
* Do we think we should document the stages in a different page? Or sprinkle in small explanations as-is in the component's doc page?
* I'm still looking how much the effort would be to improve the current configuration format by adding new River code. 

The current approach is a little tedious to write; it would be nice if we could do `stage "json" { }` and infer the contents from the block's label, but I haven't put any serious time looking into it.
```
loki.transform "default" {
  forward_to = [loki.stdout.a.receiver, loki.stdout.b.receiver]

  stage { 
    json { 
      expressions = {"genre" = "" }
      drop_malformed = true
    }
  }
  stage { 
    labels = { "genre" = "" }
  }
}
```

#### PR Checklist

- [X] CHANGELOG updated
- [X] Documentation added
- [X] Tests updated


component/loki/process/stages/inspector.go
```diff
@@ -0,0 +1,133 @@
+package stages
```
rfratto, line 1: Do we want to expose this API right now? Should we move stages to an internal package?

component/loki/process/stages/json.go
```diff
@@ -0,0 +1,180 @@
...
+
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	"github.com/jmespath/go-jmespath"
+	json "github.com/json-iterator/go"
+	"github.com/pkg/errors"
```
rfratto, line 15: Can you change any usage of `github.com/pkg/errors` to the stdlib errors package?

component/loki/process/stages/stage.go
```diff
@@ -0,0 +1,221 @@
...
+	// 	if err != nil {
+	// 		return nil, err
+	// 	}
+	// case StageTypeStaticLabels:
+	// 	s, err = newStaticLabelsStage(logger, cfg)
+	// 	if err != nil {
+	// 		return nil, err
+	// 	}
+	default:
+		return nil, errors.Errorf("Unknown stage type")
```
rfratto, line 218: nit: include the stage type provided in the error message

component/loki/process/stages/pipeline.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/grafana/agent/component/common/loki"
+	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
+	"golang.org/x/time/rate"
+)
+
+// StageConfig defines a single stage in a processing pipeline.
+type StageConfig struct {
+	JSONConfig   *JSONConfig         `river:"json,block,optional"`
+	LabelsConfig *map[string]*string `river:"labels,attr,optional"`
```
rfratto, line 23: It's a little weird to me that `json` is a block but `labels` is an attribute. I think this will make the documentation a little confusing to follow, since users will have to be aware that not all stages are subblocks. 

I'd suggest we do something like 

```
loki.process "example" {
  stage {
    labels {
      values = { ... } 
    }
  }
} 
```

just for consistency with the other stages, even though it's kind of ugly and verbose.

docs/sources/flow/reference/components/loki.process.md
```diff
@@ -0,0 +1,191 @@
...
+aliases:
+- /docs/agent/latest/flow/reference/components/loki.process
+title: loki.process
+---
+
+# loki.process
+
+`loki.process` receives log entries from other loki components, applies one or
+more processing _stages_, and forwards the results to the list of receivers
+in the component's arguments.
```
rfratto, line 11: Italicizing stages here seems a bit weird to me, since you don't really follow up on the specific word "stages." Do we want to explicitly define stage in the next paragraph, saying "A stage is a ..."?

docs/sources/flow/reference/components/loki.process.md
```diff
@@ -0,0 +1,191 @@
...
+
+The `loki.process` component is a multi-purpose tool that can parse, transform,
+and filter log entries before they're passed to a downstream component. The
```
rfratto, line 14: This feels like a second intro paragraph to the component, can this be reworded?

docs/sources/flow/reference/components/loki.process.md
```diff
@@ -0,0 +1,191 @@
...
+
+[stage]: #stage-block
+[json]: #json-block
+
+### stage block
+
+The `stage` block describes a single processing step to run log entries
+through. As such, each block must have exactly _one_ inner block or argument,
+to match the type of stage to configure. Multiple processing stages must be
+defined in different blocks.
```
rfratto, line 65: Can you you explain that stages are applied top-down in the order defined by the user?

docs/sources/flow/reference/components/loki.process.md
```diff
@@ -0,0 +1,191 @@
...
+```river
+{"log":"log message\n","extra":"{\"user\":\"agent\"}"}
+
+stage { json {
+  expressions = { output = log, extra = "" }
+}
+stage { json {
+  source      = "extra"
+  expressions = { username = "user" }
+}
```
rfratto, line 162: Can you give this example as a component rather than the body of the component? Also, please format it as riverfmt would :)

docs/sources/flow/reference/components/loki.process.md
```diff
@@ -0,0 +1,191 @@
...
+parse as JSON. By default, this is the log line itself, but it can also be a
+previously extracted value.
+
+The `expressions` field is the set of key-value pairs of MESPath expressions to
+run. The map key defines the name with which the data is extracted as, while
+the map value is the expression used to populate the value.
+
+<!-- TODO(@tpaschalis) Should we move this example here to a different tutorial?
+I don't feel the fields are enough for someone to understand what the processing
+stage is doing -->
```
rfratto, line 149: Yeah, I think examples should be more copy-and-paste focused rather than explaining details. I think explanations belong either closer to where the thing is originally defined (i.e., in the "json block" header) or in a tutorial where we can dive deeper.

rfratto (approved):
🎉 I'm happy with this for now, but would love to invest time into figuring out how we can make the component less tedious to write after we merge. 

> Do we think we should document the stages in a different page? Or sprinkle in small explanations as-is in the component's doc page?

For now I'd suggest we keep the same pattern we use for other components and keep everything in the same page. It'll be one of our larger component references, but we can see later if it warrants breaking up into multiple pages.

> The current approach is a little tedious to write; it would be nice if we could do stage "json" { } and infer the contents from the block's label, but I haven't put any serious time looking into it.

I'd suggest we try to reserve labels as something that doesn't impact parsing for consistency; block names are the standard way of determining what the schema of something is, we just have to figure out a nice API to make `stage.json` and `stage.labels` retain their order after unmarshaling.

rfratto:
> I'd suggest we try to reserve labels as something that doesn't impact parsing for consistency; block names are the standard way of determining what the schema of something is, we just have to figure out a nice API to make stage.json and stage.labels retain their order after unmarshaling.

I'd like to see a proposal for all the ways we could solve this, but I'd like to contribute one idea: 

```go
-- process.go -- 
package process  

type Arguments struct {
  ...

  // Mark Stages as a "oneof," which will unmarshal exactly one 
  // field as found in the source. This allows unmarshaling multiple 
  // blocks of different names but retain the order they were 
  // unmarshaled in.     
  Stages []stages.StageConfig `river:",oneof,optional"` 
}

-- stages/stages.go -- 
package stages 

type StageConfig struct {
  JSON *JSONConfig `river:"stage.json,block,optional"` 
  Labels *LabelsConfig `river:"stage.labels,block,optional"`  
}
```

component/loki/process/stages/pipeline.go
tpaschalis: I also had the same dillemma here, not particularly fond of either solution, but as you said, let's keep the consistency.

docs/sources/flow/reference/components/loki.process.md
tpaschalis: Okay, I'll move them back to each subblock definition, hope it doesn't make them too verbose.

component/loki/process/stages/stage.go
tpaschalis: I think that right now, we'd never reach this case. We don't switch on stage names anymore, but rather on which of the fields is non-nil.

Either we were able to successfully decode into exactly one of the defined config types, or we weren't which would fail during the Unmarshaling. This may change if we find a good way to handle `stage.json`, so I'm leaving the default case as-is for now.

tpaschalis:
> There's likely going to be a bunch of ways we can do this though, so enumerating over them in a proposal would probably be ideal

Ok, since most of the legwork around this component is here, I'll take some time to look into this then.

component/loki/process/process.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// Run implements component.Component.
+func (c *Component) Run(ctx context.Context) error {
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case entry := <-c.receiver:
+			for _, f := range c.fanout {
```
mattdurham, line 76: Is there a race condition between this and update on c.fanout?

docs/sources/flow/reference/components/loki.process.md
```diff
@@ -0,0 +1,192 @@
...
+more processing _stages_, and forwards the results to the list of receivers
+in the component's arguments.
+
+A stage is a multi-purpose tool that can parse, transform, and filter log
+entries before they're passed to a downstream component. These stages are
+applied to each log entry in order of their appearance in the configuration
+file. All stages within a `loki.process` block have access to the log entry's
+label set, the log line, the log timestamp, as well as a shared map of
+'extracted' values; this allows results of one stage to be used in a subsequent
+one.
```
karengermond, line 19:
```suggestion
file. All stages within a `loki.process` block have access to the log entry's
label set, the log line, the log timestamp, as well as a shared map of
'extracted' values, so the results of one stage to be used in a subsequent
one.
```

docs/sources/flow/reference/components/loki.process.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Blocks
+
+The following blocks are supported inside the definition of `loki.process`:
+
+Hierarchy      | Block      | Description | Required
+-------------- | ---------- | ----------- | --------
+stage          | [stage][]  | Processing stage to run. | no
+stage > json   | [json][]   | Configures a JSON processing stage.  | no
+stage > labels | [labels][] | Configures a Labels processing stage. | no
```
karengermond, line 53: I think "Labels" should be lowercase here.

docs/sources/flow/reference/components/loki.process.md
```diff
@@ -0,0 +1,192 @@
...
+
+### stage block
+
+The `stage` block describes a single processing step to run log entries
+through. As such, each block must have exactly _one_ inner block to match the
+type of stage to configure. Multiple processing stages must be defined in
+different blocks and are applied on the incoming log entries in top-down order.
+
+The block does not support any arguments and is configured only via inner
+blocks.
```
karengermond, line 70:
```suggestion
The `stage` block does not support any arguments and is configured only via inner
blocks.
```

docs/sources/flow/reference/components/loki.process.md
```diff
@@ -0,0 +1,192 @@
...
+`source`         | `string`      | Source of the data to parse as JSON. | `""` | no
+`drop_malformed` | `bool`        | Drop lines whose input cannot be parsed as valid JSON.| `false` | no
+
+When configuring a JSON stage, the `source` field defines the source of data to
+parse as JSON. By default, this is the log line itself, but it can also be a
+previously extracted value.
+
+The `expressions` field is the set of key-value pairs of MESPath expressions to
+run. The map key defines the name with which the data is extracted as, while
+the map value is the expression used to populate the value.
```
karengermond, line 91:
```suggestion
The `expressions` field is the set of key-value pairs of MESPath expressions to
run. The map key defines the name with which the data is extracted, while
the map value is the expression used to populate the value.
```

docs/sources/flow/reference/components/loki.process.md
```diff
@@ -0,0 +1,192 @@
...
+The second stage uses the value in `extra` as the input and appends the
+following key-value pair to the set of extracted data.
+```
+username: agent
+```
+
+### labels block
+
+The `labels` inner block configures a Labels processing stage that can read
+data from the extracted values map and set new labels on incoming log entries.
```
karengermond, line 131:
```suggestion
The `labels` inner block configures a labels processing stage that can read
data from the extracted values map and set new labels on incoming log entries.
```

docs/sources/flow/reference/components/loki.process.md
```diff
@@ -0,0 +1,192 @@
...
+
+Name                  | Type          | Description                               | Default        | Required
+--------------------- | --------------| ----------------------------------------- | -------------- | --------
+`labels`              | `map(string)` | Configures a `labels` processing stage.   | `{}`           | no
+
+In a Labels stage, the map's keys define the label to set and the values are
+how to look them up.  If the value is empty, it is inferred to be the same as
+the key.
```
karengermond, line 139:
```suggestion
In a labels stage, the map's keys define the label to set and the values are
how to look them up.  If the value is empty, it is inferred to be the same as
the key.
```

karengermond (approved):
LGTM!

tpaschalis:
> Does it make sense to have json and labels as separate components? This feels like components inside components.

@mattdurham apologies for the super belated response here. I thought about this, but have some gripes:
* There's 20+ of these stages, and one component for each feels like too much, especially since it would be mainly boilerplate. I agree that `loki.process` feels more like a module for now though.
* These stages have access to some shared data; the log entry being passed (log line, timestamp, labels), as well as a map of 'extracted' values that they can read/write from. This shared map is now per-component, so if we have two loki.process components, each would have their own instance of this shared cache. I'm not sure how much complexity this would add if we had each stage as a separate component.
* It would be a little harder to write for the user; keeping track of N forward_to attributes in each component.

What do you think? Would better configuration help here?

docs/sources/flow/reference/components/loki.process.md
tpaschalis: Ok, keeping this in mind as a convention for the stages that we'll gradually add as well.

component/loki/process/process.go
tpaschalis: Yes, I think you're right and that we should grab an RLock here before fanning out. This is not the case why the test finds a race condition though.

tpaschalis:
@rfratto @mattdurham The current implementation contains a race condition. The way that Promtail's pipeline works, is that it acts as a middleware; it wraps around an input channel, and returns a new, send-only channel with an EntryHandler which is responsible for receiving, transforming and feeding back the entries to the original channel in a new goroutine on the background.

The current code wraps around _each_ of the channels in `forward_to`, to create N pipelines running in parallel, that expose their send-only channels stored in c.fanout
```go
	for _, f := range newArgs.ForwardTo {
		entryHandler := loki.NewEntryHandler(f, func() {})
		c.fanout = append(c.fanout, pipeline.Wrap(entryHandler).Chan())
	}
```

The issue is that we have N (identical processing) jobs acting on incoming entries, which touch the labels slice concurrently. One way to circumvent this would be to Clone() the entry's labels for each of the fanout jobs.

A better solution would be to have the processing logic in a single pipeline. The component would keep track of a pair of `processIn` and `processOut` channels, that would be what links entries coming in to the receiver and entries being fanned out to forward_to.

```go
	for {
		select {
		case entry := <-c.receiver:
			c.processIn <- entry
		case entry := <-c.processOut:
			for _, f := range c.forward_to {
				f <- entry:
				}
			}
		}
```

If you don't have any objections, I'll go ahead with this solution; I hope it doesn't make it all that more difficult to follow.

component/loki/process/process.go
```diff
@@ -0,0 +1,135 @@
...
+			c.mut.RLock()
+			for _, f := range c.fanout {
+				select {
+				case <-ctx.Done():
+					return nil
+				case f <- entry:
+					// no-op
+				}
+			}
+			c.mut.RUnlock()
```
rfratto, line 97: How many components are going to need to do this? It's a little hard to follow, so if we expect needing to do this more than once, we might want to consider trying to abstract this logic somehow and documenting it in detail to help readers understand what's going on.

rfratto (approved):
LGTM, I only re-reviewed the fix for the race condition

component/loki/process/process.go
tpaschalis: I don't really expect more components to do so. This is complicated for two reasons: the `processIn` channel has to be send-only (except if we refactor the Loki code) so we couldn't wrap around c.receiver directly, plus each created pipeline running on its own goroutine.

If this pops up again in more logging components (eg. when we bring over the OTel<->Loki adapters) I'd like to see this under a common package eg. similarly to how we do for our own version of Prometheus' Appendable.

PR #2592, opened 2022-12-07
Add sanitization of BRANCH variable used when building

spartan0x117:
Adds sanitization of the git branch name used by the operator to prevent issues when the branch name is an invalid Kubernetes label.

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -87,6 +87,9 @@ Main (unreleased)
 - Fix issue where on checking whether to restart integrations the Integration Manager was comparing
   configs with secret values scrubbed, preventing reloads if only secrets were updated. (@spartan0x117)
 
+- Fix issue where branch names that were invalid Kubernetes labels would cause operator tests
+  to fail. (@spartan0x117)
+
 ### Other changes
 
 - Grafana Agent Flow has graduated from experimental to beta.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -87,9 +87,6 @@ Main (unreleased)
 - Fix issue where on checking whether to restart integrations the Integration Manager was comparing
   configs with secret values scrubbed, preventing reloads if only secrets were updated. (@spartan0x117)
 
-- Fix issue where branch names that were invalid Kubernetes labels would cause operator tests
-  to fail. (@spartan0x117)
-
 ### Other changes
 
 - Grafana Agent Flow has graduated from experimental to beta.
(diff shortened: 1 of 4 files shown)


tools/image-tag
```diff
@@ -10,7 +10,11 @@ fi
 set -o nounset
 
 WIP=$(git diff --quiet || echo '-WIP')
-BRANCH=$(git rev-parse --abbrev-ref HEAD | sed 's#/#-#g')
+
+# In order to be a valid Kubernetes label, this replaces '/' with '-', then replaces
+# any other invalid characters with '.', and removes all non-alphanumeric characters
+# from the beginning of the string. 
+BRANCH=$(git rev-parse --abbrev-ref HEAD | sed 's#/#-#g' | sed 's/[^A-Za-z0-9\._-]/./g' | sed 's/^[^A-Za-z0-9]*//' | head -c51)
```
rfratto, line 17: IMO this might be surprising to users. WDYT about having the sanitization in the operator code instead, since that's the only place that needs it?

CHANGELOG.md
```diff
@@ -87,6 +87,9 @@ Main (unreleased)
 - Fix issue where on checking whether to restart integrations the Integration Manager was comparing
   configs with secret values scrubbed, preventing reloads if only secrets were updated. (@spartan0x117)
 
+- Fix issue where branch names that were invalid Kubernetes labels would cause operator tests
+  to fail. (@spartan0x117)
```
rfratto, line 91: This isn't a user facing change, so the CHANGELOG wouldn't be updated here :)

pkg/operator/resources_pod_template.go
```diff
@@ -269,6 +270,26 @@ func generatePodTemplate(
...
+	labelBytes = invalidBeginningChars.ReplaceAll(labelBytes, []byte(""))
+	return trimLabel(string(labelBytes))
+}
+
+func trimLabel(label string) string {
+	if len(label) > 63 {
+		return label[:63]
+	}
+	return label
+}
```
rfratto, line 291: Would it be possible to call [clientutil.SanitizeVolumeName][src] here instead of creating new functions?

The function name might need renaming to be more clear about what it does, but it also performs validation to ensure a string is treated as valid DNS-1123 as expected in volume names and labels. 

[src]: https://github.com/grafana/agent/blob/46473875e30510e159cb889e3de5666174b73dd1/pkg/operator/clientutil/clientutil.go#L24

WDYT, are these equivalent?

spartan0x117: Good catch, that works!

rfratto (approved):
LGTM, thanks!

PR #2594, opened 2022-12-07
tools/agentlint: introduce custom linting

rfratto:
This PR introduces a custom linter as `tools/agentlint`. Initially, it only has one pass: ensure that Flow components are imported.

It will investigate `./component/all`, doing the following:

1. Retrieve the list of direct imports of `./component/all`.
2. Examine packages in `./component/...` to determine which declare a Flow component.
   1. A package declares a Flow component if it has an `init` function which calls `component.Register`.
3. Ensure that `./component/all` imports all of the discovered component packages.

Run `make agentlint` to build the linter at `./build/agentlint`. 

To run it locally, run the following, and observe that no errors are reported:

```
agentlint ./...
```

You can remove an entry from `./component/all` and rerun to observe errors being reported. 

The linter is hooked up to the `make lint` target, and we will be able to add more custom linter passes in the future to agentlint as needed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,21 +88,22 @@
 
 include tools/make/*.mk
 
-AGENT_IMAGE     ?= grafana/agent:latest
-AGENTCTL_IMAGE  ?= grafana/agentctl:latest
-OPERATOR_IMAGE  ?= grafana/agent-operator:latest
-CROW_IMAGE      ?= us.gcr.io/kubernetes-dev/grafana/agent-crow:latest
-SMOKE_IMAGE     ?= us.gcr.io/kubernetes-dev/grafana/agent-smoke:latest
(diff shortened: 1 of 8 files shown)


rfratto:
This already [found](https://drone.grafana.net/grafana/agent/5061/1/2) one issue, so I'd say it has value :)

tpaschalis (approved):

rfratto:
> Other idea for a linter; enforce river tags conventions. Right now, a tag like the following (that doesn't define whether it's an attribute or a block) will only panic at runtime. river:"listen_protocol,optional"

Ah, yeah, that’s a good idea too :) Maybe it could also report using camelCase instead of snake_case for the names, too.

mattdurham (approved):
LGTM

PR #2595, opened 2022-12-07
Flow: Add discovery.file component

### LINKED ISSUE #2525: Flow: Support file globbing for logging components
Like Promtail does, we should support glob patterns expansion for locating files on disk.

Since we opted to move that functionality out of `loki.source.file`, this could be an opportunity to go ahead and create a new `discovery` component that will perform the globbing and expose the matched files as a list of targets, similar to other service discovery mechanisms.

mattdurham:
#### PR Description

Add the ability to discover arbitrary files on the filesystem.

Closes #2525

#### Notes to the Reviewer

I wonder if we even need FSNotify, the file target in promtail uses it but it also does a sync glob check every 10 seconds. this basically does the same, and having both seems cumbersome. 

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/component/discovery/file/file.go
+++ b/component/discovery/file/file.go
@@ -26,7 +26,7 @@ func init() {
 	})
 }
 
-// Arguments holds values which are used to configure the discovery.loki.path
+// Arguments holds values which are used to configure the discovery.file
 // component.
 type Arguments struct {
 	Paths         []string      `river:"paths,attr"`
@@ -34,13 +34,14 @@ type Arguments struct {
 	UpdatePeriod  time.Duration `river:"update_period,attr,optional"`
 }
 
+// Exports exposes targets.
 type Exports struct {
 	Targets []discovery.Target `river:"targets,attr"`
 }
 
 var _ component.Component = (*Component)(nil)
 
-// Component implements the discovery.loki.path component.
+// Component implements the discovery.file component.
 type Component struct {
 	opts component.Options
 
@@ -51,7 +52,7 @@ type Component struct {
 	watchedFiles   map[string]struct{}
 }
 
-// New creates a new loki.source.file component.
+// New creates a new discovery.file component.
 func New(o component.Options, args Arguments) (*Component, error) {
 	c := &Component{
 		opts:         o,
@@ -69,7 +70,6 @@ func New(o component.Options, args Arguments) (*Component, error) {
 	if err := c.Update(args); err != nil {
 		return nil, err
 	}
-
 	return c, nil
 }
 
@@ -84,21 +84,26 @@ func (a *Arguments) UnmarshalRiver(f func(interface{}) error) error {
 	return f((*arguments)(a))
 }
 
+// Update satisfies the component interface.
 func (c *Component) Update(args component.Arguments) error {
 	c.mut.Lock()
 	defer c.mut.Unlock()
+
 	c.args = args.(Arguments)
 	c.reconcileWatchesWithWatcher()
 	return nil
 }
 
+// Run satisfies the component interface.
 func (c *Component) Run(ctx context.Context) error {
 	watchDog := time.NewTicker(c.args.UpdatePeriod)
 	timerDuration := c.args.UpdatePeriod
 	update := func() {
 		c.mut.Lock()
 		defer c.mut.Unlock()
+		// See if there is anything new we need to check.
 		c.reconcileWatchesWithWatcher()
+		// Update the exports with the targets. Should only be called if changes occured.
 		c.checkOnStateChanged()
 		// Check to see if our ticker timer needs to be reset.
 		if timerDuration != c.args.UpdatePeriod {
@@ -106,6 +111,7 @@ func (c *Component) Run(ctx context.Context) error {
 			timerDuration = c.args.UpdatePeriod
 		}
 	}
+	// Trigger initial check
 	update()
 	defer watchDog.Stop()
 	for true {
@@ -179,7 +185,7 @@ func (c *Component) reconcileWatchesWithWatcher() {
 		}
 	}
 	// Find all the removed paths.
-	pathsToRemove := make([]string, 0)
+	filesToRemove := make([]string, 0)
 	for p := range c.watchedFiles {
 		found := false
 		for _, np := range expandedPaths {
@@ -188,16 +194,31 @@ func (c *Component) reconcileWatchesWithWatcher() {
 				break
 			}
 		}
+
 		if !found {
-			pathsToRemove = append(pathsToRemove, p)
+			filesToRemove = append(filesToRemove, p)
+		}
+		// Scan to see if we need to exclude any new files.
+		for _, exclude := range c.args.ExcludedPaths {
+			matched, _ := doublestar.PathMatch(exclude, p)
+			if matched {
+				filesToRemove = append(filesToRemove, p)
+			}
 		}
 	}
-	if len(pathsToRemove) > 0 {
+
+	if len(filesToRemove) > 0 {
 		c.watchesUpdated = true
 	}
-	for _, p := range pathsToRemove {
+	for _, p := range filesToRemove {
 		cleaned := filepath.Dir(p)
-		_ = c.watcher.Remove(cleaned)
+		// Check to see if there are paths we no longer need to watch.
+		for _, exclude := range c.args.ExcludedPaths {
+			excludeDir, _ := doublestar.PathMatch(exclude, cleaned)
+			if excludeDir {
+				_ = c.watcher.Remove(cleaned)
+			}
+		}
 		delete(c.watchedFiles, p)
 	}
 }
--- a/component/discovery/file/file.go
+++ b/component/discovery/file/file.go
@@ -115,7 +115,7 @@ func (c *Component) Run(ctx context.Context) error {
 	// Trigger initial check
 	update()
 	defer watchDog.Stop()
-	for true {
+	for {
 		select {
 		case fe := <-c.watcher.Events:
 			c.fsnotifyTrigger(fe)
@@ -134,7 +134,6 @@ func (c *Component) Run(ctx context.Context) error {
 			return nil
 		}
 	}
-	return nil
 }
 
 // reconcileWatchesWithWatcher checks for any new directories that have been added along with verifying
--- a/component/discovery/file/file.go
+++ b/component/discovery/file/file.go
@@ -94,7 +94,6 @@ func New(o component.Options, args Arguments) (*Component, error) {
 		return nil, err
 	}
 	c.watcher = watcher
-	// Call to Update() to start readers and set receivers once at the start.
 	if err := c.Update(args); err != nil {
 		return nil, err
 	}
@@ -325,7 +324,17 @@ func (c *Component) getWatchedFiles() []discovery.Target {
 
 	foundFiles := make([]discovery.Target, 0)
 	for k := range c.watchedFiles {
-		foundFiles = append(foundFiles, discovery.Target{"__path__": k})
+		// This means that if a single file matches multiple outputs it will create targets with same path but different labels.
+		for _, inc := range c.args.Paths {
+			if match, _ := doublestar.PathMatch(inc["__path__"], k); match {
+				dt := discovery.Target{}
+				for dk, v := range inc {
+					dt[dk] = v
+				}
+				dt["__path__"] = k
+				foundFiles = append(foundFiles, dt)
+			}
+		}
 	}
 	return foundFiles
 }
--- a/component/discovery/file/file.go
+++ b/component/discovery/file/file.go
@@ -73,6 +73,17 @@ func New(o component.Options, args Arguments) (*Component, error) {
 	return c, nil
 }
 
+func getDefault() Arguments {
+	return Arguments{UpdatePeriod: 10 * time.Second}
+}
+
+// UnmarshalRiver implements river.Unmarshaler.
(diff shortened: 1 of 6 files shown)


docs/sources/flow/reference/components/discovery.file.md
```diff
@@ -0,0 +1,76 @@
...
+
+```river
+discovery.file "files" {
+    paths = ["/tmp/logs/**/*.log"]
+}
+loki.source.file "files" {
+    targets = discovery.file.files.targets
+    forward_to = []
+}
+```
```
tpaschalis, line 76: (Still haven't seen any of the PR code yet, just went straight to the docs)

I think one of the most sought-after usecase would be to tie this with discovery.kubernetes/discovery.relabel to read logs of all pods.

Should we include this in an example so people can easily copy-paste reuse it? Promtail has a similar issue, where it needs a lot of digging to find out the exact magic required for it to work.

mattdurham: I thought about that (not specifically k8s but more examples), and maybe it’s fine to put it in reference for now. I know @karengermond has an idea for a section that is task based and that would be perfect for this

tpaschalis: Ok, at least let's make sure between ourselves that it can fit our purpose (the specific pipeline is a little fuzzy on my mind).

mattdurham:
I wonder if we even need FSNotify, the filetarget in promtail uses it but it also does a sync glob check every 10 seconds.

docs/sources/flow/reference/components/discovery.file.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Arguments
+
+The following arguments are supported:
+
+Name | Type           | Description                                                              | Default | Required
+---- |----------------|--------------------------------------------------------------------------|-----| --------
+`paths` | `list(string)` | Doublestar compatible paths to search.                                   |     | yes
+`exclude_paths` | `list(string)` | Doublestar compatible paths to exclude, exclude_paths supercedes paths . |     | no
```
karengermond, line 28:
```suggestion
`paths` | `list(string)` | Doublestar-compatible paths to search.                                   |     | yes
`exclude_paths` | `list(string)` | Doublestar-compatible paths to exclude, exclude_paths supersedes paths . |     | no
```

docs/sources/flow/reference/components/discovery.file.md
```diff
@@ -0,0 +1,76 @@
...
+`update_period` | `duration`     | How often to sync filesystem and targets.                                | `"10s"` | no
+
+`paths` and `exclude_paths` use [doublestar][] style paths.
+* `/tmp/**/*.log` will match all subfolders of `tmp` and include any files that end in `*.log`.
+* `/tmp/apache/*.log` will match only files in `/tmp/apache/` that end in `*.log`.
+* `/tmp/**` will match all subfolders of `tmp` and `tmp` itself and all files.
```
karengermond, line 34:
```suggestion
* `/tmp/**` will match all subfolders of `tmp`, `tmp` itself, and all files.
```

docs/sources/flow/reference/components/discovery.file.md
```diff
@@ -0,0 +1,76 @@
...
+`discovery.file` does not expose any component-specific debug information.
+
+### Debug metrics
+
+`discovery.file` does not expose any component-specific debug metrics.
+
+## Examples
+
+This example discovers all files and folders under `/tmp/logs` and the absolute paths are 
+used by `loki.source.file.files` targets.
```
karengermond, line 66:
```suggestion
This example discovers all files and folders under `/tmp/logs`. The absolute paths are 
used by `loki.source.file.files` targets.
```

karengermond (approved):
LGTM!

component/discovery/file/file.go
```diff
@@ -0,0 +1,316 @@
...
+	})
+}
+
+// Arguments holds values which are used to configure the discovery.file
+// component.
+type Arguments struct {
+	Paths         []string      `river:"paths,attr"`
+	ExcludedPaths []string      `river:"excluded_paths,attr,optional"`
+	UpdatePeriod  time.Duration `river:"update_period,attr,optional"`
+}
```
rfratto, line 35: As is, this doesn't enable prevent Kubernetes-focused use cases, which IMO are a requirement for the initial release of these components. 

For collecting logs from Kubernetes pods which are mounted to the host, the expectation is that you would chain `discovery.kubernetes` to `discovery.relabel` to `discovery.file` to `loki.source.file` to be able to discover log files from containers and decorate it with metadata about the pods, something like:

```river 
discovery.kubernetes "example" {
  role = "pod"
}

discovery.relabel "example" {
  targets = [discovery.kubernetes.example.targets] 
 
  rule {
    source_labels = ["__meta_kubernetes_namespace", "__meta_kubernetes_pod_label_name"]
    target_label  = "job"
    separator     = "/"
  }

  rule {
    source_labels = ["__meta_kubernetes_pod_uid", "__meta_kubernetes_pod_container_name"]
    target_label  = "__path__"
    separator     = "/" 
    replacement   = "/var/log/pods/*$1/*.log"
  } 
}

discovery.file "example" {
  targets = [discovery.relabel.example.output] 
}

loki.source.file "example" {
  targets    = [discovery.file.example.targets] 
  forward_to = [...]
}
```

This is the equivalent of the following Promtail config file:

```yaml 
scrape_config:
- job_name: kubernetes-pods 
  kubernetes_sd_configs:
  - role: pod 
  relabel_configs:
  - source_labels:
    - __meta_kubernetes_namespace 
    - __meta_kubernetes_pod_label_name 
    target_label: job 
    separator: /
  - source_labels: 
    - __meta_kubernetes_pod_uid
    target_label: __path__ 
    separator: / 
    replacement: /var/log/pods/*$1/*.log
```

I know we're planning on exploring retrieving logs directly from the Kubernetes API server with #2504, but it won't ship with v0.30 and it's not guaranteed it'll even work at scale. At least in the interrim, we need to allow `discovery.file` to be used in a pipeline where targets are discovered from Kubernetes.

docs/sources/flow/reference/components/discovery.file.md
rfratto: As it is now, this isn't enabling use cases with discovery.kubernetes, see my comment here: https://github.com/grafana/agent/pull/2595#discussion_r1043780810

component/discovery/file/file.go
mattdurham: Good point and resolved to take in targets.

docs/sources/flow/reference/components/discovery.file.md
```diff
@@ -0,0 +1,76 @@
...
+}
+```
+
+## Arguments
+
+The following arguments are supported:
+
+Name | Type           | Description                                                                                                      | Default | Required
+---- |----------------|------------------------------------------------------------------------------------------------------------------|-----| --------
+`paths` | `list(map(string))` | Doublestar-compatible paths to search, looks for keys with `__path__`.                                           |     | yes
```
tpaschalis, line 27: Should we expand on this a little more, maybe in the intro paragraph? We can talk about how the component receives a set of targets which _must_ define the `__path__` key, similar to how Prometheus uses `__address__`, but can have other identifying labels accompanying it etc.

docs/sources/flow/reference/components/discovery.file.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Component health
+
+`discovery.file` is only reported as unhealthy when given an invalid
+configuration. In those cases, exported fields retain their last healthy
+values.
+
+## Debug information
+
+`discovery.file` does not expose any component-specific debug information.
```
tpaschalis, line 57: I think it would be nice if we could expose the expanded set of targets, so users can see with a glance how their glob patterns work. WDYT?

component/discovery/file/file.go
```diff
@@ -0,0 +1,346 @@
...
+	if !c.watchesUpdated {
+		return
+	}
+	c.watchesUpdated = false
+	output := make([]discovery.Target, len(c.watchedFiles))
+	i := 0
+	for k := range c.watchedFiles {
+		output[i] = discovery.Target{"__path__": k}
+		i++
+	}
```
tpaschalis, line 267: I think this isn't quite right. It looks like that here we'd strip the targets of all their other labels, which isn't what we'd like (expand from N->M targets based on their `__path__` and leaving the other data intact).

component/discovery/file/file.go
```diff
@@ -0,0 +1,346 @@
...
+			exclude := false
+			for _, excluded := range excludedPaths {
+				if match, _ := doublestar.Match(excluded, n); match {
+					exclude = true
+					break
+				}
+			}
+			if exclude {
+				continue
+			}
```
tpaschalis, line 208: I think maybe this snippet might belong inside of the `getPaths` helper. This way we'd get what we want for `expandedPaths` from the start.

As a plus, I think these would never appear in the list of watchedFiles so we might simplify the similar check that appears below.

component/discovery/file/file.go
```diff
@@ -0,0 +1,346 @@
...
+
+// Component implements the discovery.file component.
+type Component struct {
+	opts component.Options
+
+	mut            sync.RWMutex
+	args           Arguments
+	watcher        *fsnotify.Watcher
+	watchesUpdated bool
+	watchedFiles   map[string]struct{}
```
tpaschalis, line 80: This has the same issue as the current implementation of `loki.source.file` where it can only keep track of each file _once_ and in this case, it's 'first one wins' (eg. it cannot have the same file or pattern with different labels twice).

If it's not practical to fix now, I think we should at least put out a TODO and/or document this behavior, but let's explore how we could improve this.

component/discovery/file/file.go
```diff
@@ -0,0 +1,346 @@
...
+		args:         args,
+		watchedFiles: make(map[string]struct{}),
+	}
+
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return nil, err
+	}
+	c.watcher = watcher
+	// Call to Update() to start readers and set receivers once at the start.
```
tpaschalis, line 97: Nitpick: Let's fix the comment as we have no readers/receivers.

component/discovery/file/file.go
mattdurham: Fair point, if a file matches multiple paths, do we export all labels, merge, more? Goes back to earlier and lower discussion on unique by path or label.

docs/sources/flow/reference/components/discovery.file.md
mattdurham: Any concerns if it is thousands of files?

tpaschalis: Yes, might be an issue indeed.

The component would have the same (if not more) targets as listed as our discovery.kubernetes ones, which makes the UI unusable after a certain number of pods listed, right? I'll leave this up to you.

docs/sources/flow/reference/components/discovery.file.md
```diff
@@ -0,0 +1,124 @@
...
+
+`discovery.file` discovers files on the local filesystem using the [doublestar][] library.
+
+[doublestar]: https://github.com/bmatcuk/doublestar
+
+## Usage
+
+```river
+discovery.file "LABEL" {
+  paths = ["DOUBLESTAR_PATH"]
```
tpaschalis, line 17: Let's fix this, since now it doesn't expect just the glob, but a list of targets

Couple of more points
a) Do you think we should we rename the Argument to `targets`?
b) Do you think we should expliticly call out that the component looks out for the `__path__` key in the input targets?
c) Seems like we're not using the `exclude_paths` Argument anymore, but getting them from the labels, right?

mattdurham: Good call on b and c. As far as A I could go either way, paths feel more intuitive to me, but targets are more type correct.

tpaschalis: I'd recommend we go with 'targets' so it can more easily be read (eg. using the output of `kubernetes.discovery.default.targets`)

mattdurham: IMO after thinking on it for 5, I think it should not be targets just in case we allow referencing arguments at some point in river.

component/discovery/file/file.go
tpaschalis: I'm not sure I understood what you meant here.

I feel the logic here doesn't exactly match what Promtail did and what we want to do here.

In the current code, we gather all the paths, glob expand them, and then drop them if _any_ of the targets happened to match the excludion.

I believe what we'd like to do is treat each target _separately_, so each `__path__` and possibly `__path_exclude__` that we encounter is globbed and exported with its own set of labels.

component/discovery/file/file.go
mattdurham: This is resolved now

docs/sources/flow/reference/components/discovery.file.md
mattdurham: After conversation went with path_targets

component/discovery/file/file.go
tpaschalis: One thing I noticed is that we now export directories as targets, too. I'm not sure if that was the case with the fsnotify approach before.

This will make the downstream components just reject those targets, but do you think we should skip them over here as well?

mattdurham: We should skip thats a bug I introduced

tpaschalis (approved):

PR #2599, opened 2022-12-08
Flow: Add support for labels in loki components positions file

tpaschalis:

#### PR Description
In #2503 we introduced a new `loki.source.file` component, which tails files from disk and sends scraped log entries to other loki.* components.

There are a few omissions with the component current component implementation.

* On Update() it stops and re-creates all readers, even if it's not required
* It does not support tailing multiple files with the same path but different label sets (last one currently wins)
* It cannot expose target labels in DebugInfo

Some of this behavior stems from Promtail's positions file implementation which indexes _only by path_.

This PR updates the `positions` package to index files by both path _and_ the label set.
This will allow for smarter handling when updating the status of targets, better insight on the current status of active readers, and finer-grained handling of targets in future PRs.

Since this is a breaking change for the Positions file format (rendering previously-used positions files useless) this would be a breaking change if it were to be included in a future release, so I'd like to try and include this with v0.30.

The new positions file format looks like this
```
--- positions.yml ---
$ cat data-agent/loki.source.file.default/positions.yml
positions:
  ? path: /Users/tpaschalis/GitRepos/agent/test/foo.txt
    labels: '{color="pink"}'
  : "1472"
  ? path: /Users/tpaschalis/GitRepos/agent/test/bar.txt
    labels: '{color="blue"}'
  : "22"
  ? path: /Users/tpaschalis/GitRepos/agent/test/baz.txt
    labels: '{}'
  : "29"
```

The question mark character makes use of YAML's 'complex key mapping' feature ([1](https://stackoverflow.com/questions/33987316/what-is-a-complex-mapping-key-in-yaml) [2](https://yaml.org/spec/1.2.2/#:~:text=Example%202.11%20Mapping%20between%20Sequences)) which allows to index a key using a _structure_ and not a single string.

An alternative would be to use a separator like `filename~~~labels` or `filename|labels`, but since we want the format to be portable between *nix-like and Windows machines _and_  be able to parse back and forth from it without any ambiguous meaning; I couldn't find a good way to do that (eg. a character that cannot be a part of a path in any Windows _or_ Linux filesystems).

Follow-up to #2503.

#### Notes to the Reviewer
This is still a WIP. At a first glance, it works locally, but I'd like to verify its behavior and increase our confidence with a test suite.

Also, will have to verify the positions file structure.

#### PR Checklist

- [ ] CHANGELOG updated (N/A, not a user-facing change)
- [ ] Documentation added
- [X] Tests updated

### DIFF
--- a/component/common/loki/positions/positions.go
+++ b/component/common/loki/positions/positions.go
@@ -51,14 +51,21 @@ type positions struct {
 	logger    log.Logger
 	cfg       Config
 	mtx       sync.Mutex
-	positions map[string]string
+	positions map[Entry]string
 	quit      chan struct{}
 	done      chan struct{}
 }
 
+// Entry desribes a positions file entry consisting of an absolute path and
+// the accompanying labels.
+type Entry struct {
+	Path   string `yaml:"path"`
+	Labels string `yaml:"labels"`
+}
+
 // File format for the positions data.
 type File struct {
-	Positions map[string]string `yaml:"positions"`
+	Positions map[Entry]string `yaml:"positions"`
 }
 
 type Positions interface {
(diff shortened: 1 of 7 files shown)


mattdurham (approved):
Lgtm

PR #2600, opened 2022-12-08
component/otelcol: expose metrics from upstream otelcol components

### LINKED ISSUE #2372: Flow: expose telemetry from otelcol components
Currently, none of our otelcol components expose metrics to help users debug or introspect their status. 

Metrics from OpenTelemetry Collector components work primarily by using [obsreport](https://pkg.go.dev/go.opentelemetry.io/collector@v0.62.1/obsreport). `obsreport` provides helper methods for components to use to instrument operations for collecting traces, metrics, and logs. 

Internally, these helper methods [use OpenCensus](https://github.com/open-telemetry/opentelemetry-collector/blob/b45d0ed7f3c7f897fa59071a36371592c5f6d6a9/obsreport/obsreport_exporter.go#L110-L115) to record individual measurements for each component. The metrics exposed are defined as globals in an [internal package](https://github.com/open-telemetry/opentelemetry-collector/blob/b45d0ed7f3c7f897fa59071a36371592c5f6d6a9/internal/obsreportconfig/obsmetrics/obs_exporter.go#L50-L78). 

Our component wrappers should find a way to expose metrics from running OpenTelemetry Collector endpoints as Prometheus metrics. 

The biggest challenge is that Flow components are ephemeral, but OpenTelemetry Collector components are (currently) scoped to the process lifetime. This means that we should remove metrics for components which are no longer running in Grafana Agent Flow. 

There are two approaches we could take, in descending preference order:

1. Find a way to have our wrappers expose OpenCensus metrics as Prometheus metrics for _only_ the components being wrapped. This would automatically remove them once the component is removed. 
2. Globally register OpenCensus metrics from OpenTelemetry Collector, and figure out a way to remove metrics for components that have been terminated. 

cc @tpaschalis @jpkrohling

rfratto:
This commit configures the MeterProvider given to otelcol components to forward metrics to a Prometheus Registry. A new Prometheus Registry is generated every time otelcol components are created and passed to a lazy collector to hot-swap out the Registry being used.

For this to work, the telemetry.useOtelForInternalMetrics feature gate from OpenTelemetry Collector is forcibly enabled, which tells the go.opentelemetry.io/collector/obsreport package (which is used by some components for generating metrics) to use the provided MeterProvider instead of OpenCensus for metrics.

Note that the OpenTelemetry Collector dependency will need to be updated to v0.67 before we can make full use of this change, as v0.67 includes changes to obsreport so that the MeterProvider is used for all component types.

Closes #2372.


component/otelcol/receiver/receiver.go
```diff
@@ -99,14 +111,20 @@ func (r *Receiver) Update(args component.Arguments) error {
 		scheduler.WithHostExporters(rargs.Exporters()),
 	)
 
+	reg := prometheus.NewRegistry()
+	r.collector.Set(reg)
+
+	promExporter, err := sdkprometheus.New(sdkprometheus.WithRegisterer(reg), sdkprometheus.WithoutTargetInfo())
```
tpaschalis, line 117: Nitpick, but why is this WithoutTargetInfo? Is it to not pass unnecessary metrics to other prometheus.* components later on?

tpaschalis (approved):

tpaschalis:
LGTM, pretty straightforward!

Issue #1986, opened 2022-08-03
Grafana Agent Operator - Support for the same features as Prometheus Operator 


aengusrooneygrafana:
Following discussion with a number of users, it has been found that the Prometheus Operator and the Grafana Agent Operator are slightly out of feature alignment.  The Prometheus Operator supports Alerts and Alerting, which the Grafana Agent Operator does not.  This leads to a proliferation of tools required for maintenance of AM and Alerts outside of the operator.  

This proposal is to request the same level of support for Alerting in the Grafana Agent Operator, as is supported in the Prometheus Operator.  Ref:  https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/user-guides/alerting.md

rfratto:
This is similar to #523 where we discussed adding the ability to Grafana Agent to sync rules with the Cortex/Mimir ruler API.

It sounds reasonable to add that functionality into both the agent and the operator, though it's not going to be simple; we need to figure out how to identify rules that came from an agent so we know which should be added/removed/changed when reconciling the list.

flokli:
Prometheus Operator seems to store this info in the [`Prometheus` CRs](https://docs.openshift.com/container-platform/4.7/rest_api/monitoring_apis/prometheus-monitoring-coreos-com-v1.html), at `spec.{rules,ruleSelector}`.

While using different CRDs, VictoriaMetrics operator seems to have a alerts-specific [`VMAlert.spec.{ruleSelector,ruleNamespaceSelector}](https://docs.victoriametrics.com/operator/api.html#vmalertspec)` that describes where to select `*Rule` CRs from.

I assume to keep this as frictionless as possible, no new CRDs should be introduced, but Grafana Agent Operator could listen to `Prometheus` CRs?

Or, looking at https://github.com/grafana/agent/pull/1839 and similar proposals, Grafana Agent would watch `PrometheusRule` CRs, and selectors/namespaces would be a config of Grafana Agent itself?

azhurbilo:
remove stale

jhohertz:
This would be a boon if this support existed in the agent.

rfratto:
Hey all, support for recording and alerting rules is something still being considered. We don't have any updates right now.

With the recent [Grafana Agent Flow](https://grafana.com/blog/2022/09/29/introducing-programmable-pipelines-with-grafana-agent-flow/?mdm=social) announcement, it might make sense to support recording/alerting rules as Flow components, too.

[...]

Issue #523, opened 2021-04-08
Support defining Cortex Rules in Agent config


rfratto:
Based on [this discussion](https://github.com/grafana/agent/discussions/456), I'd like it if the Agent could support recording and alerting rule definitions that are synced against the Cortex Ruler API.

jpkrohling:
I'm removing myself from this task, as it's been for a while in my queue and I can't seem to even get started with it. I'll try to get a smaller issue first to get myself more comfortable with this codebase and might work on this again in the future. Sorry about that :-/

sambacha:
Any update on this issue? This is a blocker to adopting grafana

rfratto:
No updates for now, this isn't being actively worked on at the moment.

rfratto:
This was implemented as a Grafana Agent Flow component in #2604. I'm going to close this as completed. 

At this time, there's no plan to backport this functionality into the static mode (i.e., the default mode the agent runs in).

PR #2604, opened 2022-12-09
Add support for configuring Grafana Mimir via PrometheusRule CRDs

### LINKED ISSUE #1544: Listen for PrometheusRule resources and configure in Grafana Mimir/Cortex
It'd be really awesome if GrafanaAgent could listen for `PrometheusRule` resources, and create remote Mimir/Cortex configuration.

While it seems it's possible to configure these manually via Grafana, a lot of applications ship `PrometheusRule` resources to describe aggregations or alerts.

Using this would mean there's no need to deploy alertmanager in a downstream cluster, and do aggregation there, but it could all live in the monitoring system.

Logiraptor:
#### PR Description

This PR adds a new flow component: `mimir.rules.kubernetes`. This was my hackathon project for the December Grafana Labs hackathon. (I'm on the Mimir squad). Here's a list of items I managed to get done this week:

* Live config reload works
* Efficient use of the k8s api (informer pattern)
* Rate limited (both overall and per-item for backoff) interaction with Mimir
* Debug info available in agent UI
* Able to use many agents without clobbering each other’s rules
* Able to still use Grafana Cloud integration rules
* Metrics available for work queue stats, k8s response time, mimir response time
* Supports all methods of authentication with Mimir / GEM / Grafana cloud including anonymous, api key, basic auth, and oidc
* Supports all TLS settings on the mimir client side

Fixes #1544
Fixes https://github.com/grafana/mimir/issues/2133
Related to #523 and #1986

#### Notes to the Reviewer

~~I added a few tests including some integration tests for the core event loop, but please let me know if there are more tests I should add. Notably, some of the scaffolding code is currently untested.~~

~~Also missing as of now is documentation. Happy to contribute there, if you have any pointers?~~

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated
- [x] @Logiraptor to resolve the licensing issue for the Mimir Client code copied from github.com/grafana/mimir


pkg/mimir/client/client.go
```diff
@@ -0,0 +1,211 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+// Provenance-includes-location: https://github.com/grafana/cortex-tools/blob/main/pkg/client/client.go
+// Provenance-includes-license: Apache-2.0
+// Provenance-includes-copyright: The Cortex Authors.
```
rfratto, line 4: Is this AGPL code? We're not going to be allowed to merge this if so; we're Apache 2.0-only.

Logiraptor: Gotcha, it's copied from Mimir's repo, so I'm sure we can work it out considering we own the license for both sides. I'll take this as an action item.

rfratto: It looks like it's copied from grafana/cortex-tools, which seems to be Apache 2 (but I might be missing something here)

Logiraptor: Yeah originally cortex Apache 2, copied into mimir and became AGPL, then copied here.

component/mimir/rules/debug.go
```diff
@@ -0,0 +1,64 @@
+package rules
+
+import "fmt"
+
+type DebugInfo struct {
+	Error               string                   `river:"error,attr,optional"`
+	PrometheusRules     []DebugK8sPrometheusRule `river:"prometheusRules,attr,optional"`
+	MimirRuleNamespaces []DebugMimirNamespace    `river:"mimirRuleNamespaces,attr,optional"`
```
rfratto, line 8: These should probably be blocks (not attrs) so they get dedicated sections in the UI rather than all being shoved into the same table. Also, when changing them to blocks, make the names singular rather than plural.

component/mimir/rules/debug.go
```diff
@@ -0,0 +1,64 @@
...
+	Namespace     string `river:"namespace,attr"`
+	Name          string `river:"name,attr"`
+	UID           string `river:"uid,attr"`
+	NumRuleGroups int    `river:"numRuleGroups,attr"`
+}
+
+type DebugMimirNamespace struct {
+	Name          string `river:"name,attr"`
+	NumRuleGroups int    `river:"numRuleGroups,attr"`
+}
```
rfratto, line 21: Canonically we use snake_case for block and attribute names.

component/mimir/rules/rules.go
```diff
@@ -0,0 +1,309 @@
...
+
+	promExternalVersions "github.com/prometheus-operator/prometheus-operator/pkg/client/informers/externalversions"
+	promVersioned "github.com/prometheus-operator/prometheus-operator/pkg/client/versioned"
+)
+
+func init() {
+	component.Register(component.Registration{
+		Name:    "mimir.rules",
+		Args:    Arguments{},
+		Exports: Exports{},
```
rfratto, line 34: If a component doesn't export anything, you should leave this nil and remove the Exports type.

component/mimir/rules/types.go
```diff
@@ -0,0 +1,60 @@
...
+
+func setDefaultArguments(args *Arguments) {
+	if args.SyncInterval == 0 {
+		args.SyncInterval = 30 * time.Second
+	}
+
+	if args.MimirNameSpacePrefix == "" {
+		args.MimirNameSpacePrefix = "agent"
+	}
+}
```
rfratto, line 26: The normal way of setting default arguments is to implement river.Unmarshaler:

```go 
var DefaultArguments = Arguments{
  SyncInterval: 30 * time.Second,
  MimirNameSpacePrefix: "agent", 
}

func (args *Arguments) UnmarshalRiver(f func(interface{}) error) error {
  *args = DefaultArguments 

  type arguments Arguments 
  if err := f((*arguments)(args)); err != nil {
    return err 
  }

  if args.SyncInterval <= 0 {
    return fmt.Errorf("sync_interval must be greater than 0")
  }
  if args.MimirNameSpacePrefix == "" {
    return fmt.Errorf("mimir_namespace_prefix must not be empty")
  }

  return nil
}

```

component/mimir/rules/types.go
```diff
@@ -0,0 +1,60 @@
...
+
+type LabelSelector struct {
+	MatchLabels      map[string]string `river:"match_labels,attr,optional"`
+	MatchExpressions []MatchExpression `river:"match_expressions,attr,optional"`
```
rfratto, line 30: Is there value in having this be an attribute instead of a block? Blocks are typically the default way of grouping settings together. Object attributes are desirable only if you anticipate someone loading an entire `match_expression` value from another component/expression (such as `json_decode`).

component/mimir/rules/types.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+type TLSArguments struct {
+	CertPath           string `river:"tls_cert_path,attr,optional"`
+	KeyPath            string `river:"tls_key_path,attr,optional"`
+	CAPath             string `river:"tls_ca_path,attr,optional"`
+	ServerName         string `river:"tls_server_name,attr,optional"`
+	InsecureSkipVerify bool   `river:"tls_insecure_skip_verify,attr,optional"`
+	CipherSuites       string `river:"tls_cipher_suites,attr,optional"`
+	MinVersion         string `river:"tls_min_version,attr,optional"`
```
rfratto, line 56: We've defined this type already in `component/common/config`, is there a reason we need to redefine it?

component/mimir/rules/rules.go
```diff
@@ -0,0 +1,309 @@
...
+	_ "k8s.io/component-base/metrics/prometheus/workqueue"
+	controller "sigs.k8s.io/controller-runtime"
+
+	promExternalVersions "github.com/prometheus-operator/prometheus-operator/pkg/client/informers/externalversions"
+	promVersioned "github.com/prometheus-operator/prometheus-operator/pkg/client/versioned"
+)
+
+func init() {
+	component.Register(component.Registration{
+		Name:    "mimir.rules",
```
rfratto, line 32: There's probably some discussion to be had around the component name. If this is the only component we ever have for loading rules into Mimir, `mimir.rules` is OK, but I wouldn't be surprised if we have non-Kubernetes components to do something similar in the future. 

Maybe this should be `mimir.rules.kubernetes` or something?

component/mimir/rules/rules.go
```diff
@@ -0,0 +1,309 @@
...
+		ticker:        time.NewTicker(c.SyncInterval),
+		metrics:       metrics,
+	}, nil
+}
+
+func (c *Component) Run(ctx context.Context) error {
+	err := c.startup(ctx)
+	if err != nil {
+		return err
+	}
```
rfratto, line 148: I'm not sure about this returning an error; it'll terminate the component and won't spin back up. Should it log instead and set the health of the component to unhealthy so people can fix? 

Ditto with returning an error in the for loop.

component/mimir/rules/rules.go
```diff
@@ -0,0 +1,309 @@
...
+var _ component.Component = (*Component)(nil)
+var _ component.DebugComponent = (*Component)(nil)
+
+func NewComponent(o component.Options, c Arguments) (*Component, error) {
+	setDefaultArguments(&c)
+
+	metrics := newMetrics()
+	metrics.Register(o.Registerer)
+
+	return &Component{
```
rfratto, line 134: Ideally we call `init` here as a way to validate the config file and get any early errors returned to the user before the component is running. However, if `init` will be slow or launches any goroutines, it's best left out and delegated to the Run function.

component/mimir/rules/diff.go
```diff
@@ -0,0 +1,117 @@
...
+	RuleGroupDiffKindAdd    RuleGroupDiffKind = "add"
+	RuleGroupDiffKindRemove RuleGroupDiffKind = "remove"
+	RuleGroupDiffKindUpdate RuleGroupDiffKind = "update"
+)
+
+type RuleGroupDiff struct {
+	Kind    RuleGroupDiffKind
+	Actual  mimirClient.RuleGroup
+	Desired mimirClient.RuleGroup
+}
```
rfratto, line 23: I've been trying to encourage people to keep the publicly exposed API of new components small to have greater flexibility over how our code gets used. It doesn't look like this needs to be exported; I'd recommend unexporting it. D

itto with everything in `events.go`.

rfratto (commented):
Nice! This is off to a great start, and I'm really excited to have this component. 

We'll want a CHANGELOG entry and [reference documentation][refs] before we merge. Writing documentation is usually the most time consuming part of new components since we try to explain as much as users need to know in the reference docs. 

[refs]: https://github.com/grafana/agent/tree/main/docs/sources/flow/reference/components

> I added a few tests including some integration tests for the core event loop, but please let me know if there are more tests I should add. Notably, some of the scaffolding code is currently untested.

This will be a judgment call by you; if you think what you have tested is enough to have confidence in the correctness of the code (and any untested code is obviously correct), then you don't need to add any other tests :)

component/mimir/rules/types.go
Logiraptor: Nope, just didn't know it existed. :)

component/mimir/rules/rules.go
Logiraptor: I think I incorrectly assumed returning an error would terminate the pod (meaning normal CrashLoop monitoring would alert someone). I'll make the change

pkg/mimir/client/client.go
Logiraptor: As discussed in slack, I've solved this by rebasing onto the Apache 2 cortex client: https://github.com/grafana/agent/pull/2604/commits/e6feaad6a3399a0eb502155ec002b818643f5362, and adding a simple rename commit on top: https://github.com/grafana/agent/pull/2604/commits/74311ef9fc1ea30478fff94f5dbb81149b2e1241. All other commits were rebased from that point and are new code written for this hackathon. There is now a clear provenance back to cortex and all changes from there were done fresh for this PR. The end result is barely different, but it should be a valid use of the license now.

component/mimir/rules/diff.go
```diff
@@ -0,0 +1,111 @@
+package rules
+
+import (
+	"bytes"
+
+	mimirClient "github.com/grafana/agent/pkg/mimir/client"
+
+	"gopkg.in/yaml.v3"
```
rfratto, line 8: We typically try to use yaml.v2 (though at this point I don't remember why). Is v3 necessary here?

rfratto (commented):
🎉 I'm happy with this. We still need a few things before it can be merged:

* Changelog entry 
* Documentation 

Other than that, looks great, thank you for working on this!

component/mimir/rules/diff.go
Logiraptor: v3 is necessary because its used to unmarshal into structs that directly reference v3 types. For example the `RuleNode` type in prometheus: https://github.com/grafana/prometheus/blob/master/pkg/rulefmt/rulefmt.go#L124. Prometheus uses the yaml Node types directly in order to provide line numbers for syntax and validation errors.

I've tested it with v2, which fails to unmarshal.

component/mimir/rules/events.go
```diff
@@ -0,0 +1,261 @@
+package rules
+
+import (
+	"context"
+	"fmt"
+	"regexp"
+	"time"
+
+	"github.com/ghodss/yaml"
```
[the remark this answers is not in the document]
Logiraptor: `github.com/ghodss/yaml` is necessary here because it's used to marshal from structs that only have json annotations. Rather than coming up with a solution I just copied what the prometheus-operator does to maintain 100% compatibility: https://github.com/prometheus-operator/prometheus-operator/blob/3ee8ce54baea4e99c1e2dfe6bd6d0b6b8b265e76/pkg/operator/rules.go#L21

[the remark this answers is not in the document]
Logiraptor: I'll add a small comment to that line to clarify

component/mimir/rules/diff.go
Logiraptor: I'll add a small comment to that line to clarify

component/mimir/rules/events.go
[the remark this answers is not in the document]
rfratto: 👍 Thanks!

component/mimir/rules/events.go
```diff
@@ -0,0 +1,261 @@
...
+	"github.com/go-kit/log/level"
+	mimirClient "github.com/grafana/agent/pkg/mimir/client"
+	"github.com/grafana/dskit/multierror"
```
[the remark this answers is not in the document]
Logiraptor: Done in 713ed26

component/mimir/rules/events.go
```diff
@@ -0,0 +1,261 @@
...
+	groups, errs := rulefmt.Parse(buf)
+	if len(errs) > 0 {
+		return nil, multierror.New(errs...).Err()
+	}
+
+	mimirGroups := make([]mimirClient.RuleGroup, len(groups.Groups))
+	for i, g := range groups.Groups {
+		mimirGroups[i] = mimirClient.RuleGroup{
+			RuleGroup: g,
+			// TODO: allow setting remote write configs?
```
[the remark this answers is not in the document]
Logiraptor: This is a GEM-only extension to Mimir's Ruler API that allows you to forward the rule output to an external database. It's not used much as far as I know and isn't part of the core use-case for the agent. I think we should omit it unless a customer specifically asks for it. What do you think?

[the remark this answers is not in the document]
rfratto: Yeah, I think leaving it out makes sense for now, thanks for explaining 👍

component/mimir/rules/events.go
```diff
@@ -0,0 +1,261 @@
...
+
+// OnUpdate implements the cache.ResourceEventHandler interface.
+func (c *Component) OnUpdate(oldObj, newObj interface{}) {
+	c.publishEvent(newObj)
+}
+
+// OnDelete implements the cache.ResourceEventHandler interface.
+func (c *Component) OnDelete(obj interface{}) {
+	c.publishEvent(obj)
+}
```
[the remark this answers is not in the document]
Logiraptor: Done in 5b45501

component/mimir/rules/events.go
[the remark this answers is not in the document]
Logiraptor: Cool, I was able to remove some glue code in that case: 48223c1

component/mimir/rules/diff.go
```diff
@@ -0,0 +1,111 @@
...
+	ruleGroupDiffKindUpdate ruleGroupDiffKind = "update"
+)
+
+type ruleGroupDiff struct {
+	Kind    ruleGroupDiffKind
+	Actual  mimirClient.RuleGroup
+	Desired mimirClient.RuleGroup
+}
+
+func diffRuleState(desired map[string][]mimirClient.RuleGroup, actual map[string][]mimirClient.RuleGroup) map[string][]ruleGroupDiff {
```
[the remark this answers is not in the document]
Logiraptor: I tried doing that, but it ends up forcing me to allocate copies of the original maps just to satisfy the types. I wonder if it could be better to just define a type for the whole map, which is easier to just type convert.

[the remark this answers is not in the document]
rfratto: Maybe just a doc comment on the function then?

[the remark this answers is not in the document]
rfratto: Oh, a type for the whole map could work too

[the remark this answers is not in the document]
rfratto: Looks good, thanks!

Logiraptor:
Ok, @rfratto I've added a first pass at reference documentation based on some of the other docs I found. I think at this point it should be good. Let me know if anything else needs fixing!

CHANGELOG.md
```diff
@@ -107,8 +107,12 @@ v0.30.0 (2022-12-20)
   - `discovery.file` discovers files on the filesystem following glob
     patterns. (@mattdurham)
 
+  - `mimir.rules.kubernetes` discovers `PrometheusRule` kubernetes resources and
```
rfratto, line 110:
```suggestion
  - `mimir.rules.kubernetes` discovers `PrometheusRule` Kubernetes resources and
```

docs/sources/flow/reference/components/mimir.rules.kubernetes.md
```diff
@@ -0,0 +1,202 @@
+---
+aliases:
+- /docs/agent/latest/flow/reference/components/mimir.rules.kubernetes
+title: mimir.rules.kubernetes
+---
+
+# mimir.rules.kubernetes
+
+`mimir.rules.kubernetes` discovers `PrometheusRule` kubernetes resources and
```
rfratto, line 9:
```suggestion
`mimir.rules.kubernetes` discovers `PrometheusRule` Kubernetes resources and
```

docs/sources/flow/reference/components/mimir.rules.kubernetes.md
```diff
@@ -0,0 +1,202 @@
...
+loads them into a Mimir instance.
+
+* Multiple `mimir.rules.kubernetes` components can be specified by giving them
+  different labels.
+* [Kubernetes label selectors][] can be used to limit the `Namespace` and
+  `PrometheusRule` resources considered during reconciliation.
+* Compatible with the Ruler APIs of Grafana Mimir, Grafana Cloud, and Grafana Enterprise Metrics
+* Compatible with the `PrometheusRule` CRD from the [prometheus-operator][]
```
rfratto, line 17:
```suggestion
* Compatible with the Ruler APIs of Grafana Mimir, Grafana Cloud, and Grafana Enterprise Metrics.
* Compatible with the `PrometheusRule` CRD from the [prometheus-operator][].
```

docs/sources/flow/reference/components/mimir.rules.kubernetes.md
```diff
@@ -0,0 +1,202 @@
...
+
+### label_selector block
+
+The `label_selector` block describes a kubernetes label selector for rule or namespace discovery.
+
+The following arguments are supported:
+
+Name           | Type          | Description                                       | Default                     | Required
+---------------|---------------|---------------------------------------------------|-----------------------------|---------
+`match_labels` | `map(string)` | Label keys and values used to discover resources. | empty (match all resources) | yes
```
rfratto, line 93:
```suggestion
`match_labels` | `map(string)` | Label keys and values used to discover resources. | `{}` | yes
```

docs/sources/flow/reference/components/mimir.rules.kubernetes.md
```diff
@@ -0,0 +1,202 @@
...
+
```
rfratto, line 94:
```suggestion

When the `match_labels` argument is empty, all resources will be matched. 

```

docs/sources/flow/reference/components/mimir.rules.kubernetes.md
```diff
@@ -0,0 +1,202 @@
...
+The `match_expression` block describes a kubernetes label match expression for rule or namespace discovery.
+
+The following arguments are supported:
+
+Name       | Type       | Description                                        | Default | Required
+-----------|------------|----------------------------------------------------|---------|---------
+`key`      | `string`   | The label name to match against.                   |         | yes
+`operator` | `string`   | The operator used when matching. (in,notin,exists) |         | yes
+`values`   | `[]string` | The values used when matching.                     |         | no
+
```
rfratto, line 106:
```suggestion

The `operator` argument should be one of the following strings:

* `"in"` 
* `"notin"` 
* `"exists"` 

```

docs/sources/flow/reference/components/mimir.rules.kubernetes.md
```diff
@@ -0,0 +1,202 @@
...
+### match_expression block
+
+The `match_expression` block describes a kubernetes label match expression for rule or namespace discovery.
+
+The following arguments are supported:
+
+Name       | Type       | Description                                        | Default | Required
+-----------|------------|----------------------------------------------------|---------|---------
+`key`      | `string`   | The label name to match against.                   |         | yes
+`operator` | `string`   | The operator used when matching. (in,notin,exists) |         | yes
```
rfratto, line 104:
```suggestion
`operator` | `string`   | The operator to use when matching. |         | yes
```

docs/sources/flow/reference/components/mimir.rules.kubernetes.md
```diff
@@ -0,0 +1,202 @@
...
+## Component health
+
+`mimir.rules.kubernetes` is reported as unhealthy if given an invalid configuration or an error occurs during reconciliation.
+
+## Debug information
+
+`mimir.rules.kubernetes` exposes resource-level debug information.
+
+The following are exposed per discovered `PrometheusRule` resource:
+* The kubernetes namespace.
```
rfratto, line 142:
```suggestion
* The Kubernetes namespace.
```

rfratto (approved):
👏 fantastic. I'll leave the final doc review to @karengermond, but once that's done, we'll get this merged!

docs/sources/flow/reference/components/mimir.rules.kubernetes.md
```diff
@@ -0,0 +1,202 @@
...
+            agent = "yes",
+        }
+    }
+}
+```
+
+This example creates a `mimir.rules.kubernetes` component that loads discovered
+rules to Grafana Cloud.
+
+```
```
rfratto, line 190:
```suggestion
```river
```

rfratto (commented):
sorry, one extra thing :)

docs/sources/flow/reference/components/mimir.rules.kubernetes.md
```diff
@@ -0,0 +1,202 @@
...
+
+`mimir.rules.kubernetes` supports the following arguments:
+
+Name                     | Type       | Description                                             | Default | Required
+-------------------------|------------|---------------------------------------------------------|---------|---------
+`address`                | `string`   | URL of the Mimir ruler.                                 |         | yes
+`tenant_id`              | `string`   | Mimir tenant ID                                         |         | no
+`use_legacy_routes`      | `bool`     | Whether to use deprecated ruler API endpoints           | false   | no
+`sync_interval`          | `duration` | Amount of time between reconciliations with Mimir       | "30s"   | no
+`mimir_namespace_prefix` | `string`   | Prefix used to differentiate multiple agent deployments | "agent" | no
```
karengermond, line 40:
```suggestion
`tenant_id`              | `string`   | Mimir tenant ID.                                         |         | no
`use_legacy_routes`      | `bool`     | Whether to use deprecated ruler API endpoints.           | false   | no
`sync_interval`          | `duration` | Amount of time between reconciliations with Mimir.       | "30s"   | no
`mimir_namespace_prefix` | `string`   | Prefix used to differentiate multiple agent deployments. | "agent" | no
```

docs/sources/flow/reference/components/mimir.rules.kubernetes.md
```diff
@@ -0,0 +1,202 @@
...
+
+If no `tenant_id` is provided, the component assumes that the Mimir instance at
+`address` is running in single-tenant mode and no X-Scope-OrgID header is sent.
```
karengermond, line 43:
```suggestion
If no `tenant_id` is provided, the component assumes that the Mimir instance at
`address` is running in single-tenant mode and no `X-Scope-OrgID` header is sent.
```

docs/sources/flow/reference/components/mimir.rules.kubernetes.md
```diff
@@ -0,0 +1,202 @@
...
+
+The `sync_interval` argument determines how often Mimir's ruler API is accessed
+to reload the current state of rules. Interaction with the kubernetes API works
+differently. Updates are processed as events from the kubernetes API server
+according to the informer pattern.
```
karengermond, line 48:
```suggestion
The `sync_interval` argument determines how often Mimir's ruler API is accessed
to reload the current state of rules. Interaction with the Kubernetes API works
differently. Updates are processed as events from the Kubernetes API server
according to the informer pattern.
```

docs/sources/flow/reference/components/mimir.rules.kubernetes.md
```diff
@@ -0,0 +1,202 @@
...
+
+The following blocks are supported inside the definition of
+`mimir.rules.kubernetes`:
+
+Hierarchy                                  | Block                  | Description                                              | Required
+-------------------------------------------|------------------------|----------------------------------------------------------|---------
+rule_namespace_selector                    | [label_selector][]     | Label selector for `Namespace` resources                 | no
+rule_namespace_selector > match_expression | [match_expression][]   | Label match expression for `Namespace` resources         | no
+rule_selector                              | [label_selector][]     | Label selector for `PrometheusRule` resources            | no
+rule_selector > match_expression           | [match_expression][]   | Label match expression for `PrometheusRule` resources    | no
```
karengermond, line 64:
```suggestion
rule_namespace_selector                    | [label_selector][]     | Label selector for `Namespace` resources.                 | no
rule_namespace_selector > match_expression | [match_expression][]   | Label match expression for `Namespace` resources.         | no
rule_selector                              | [label_selector][]     | Label selector for `PrometheusRule` resources.            | no
rule_selector > match_expression           | [match_expression][]   | Label match expression for `PrometheusRule` resources.    | no
```

docs/sources/flow/reference/components/mimir.rules.kubernetes.md
```diff
@@ -0,0 +1,202 @@
...
+[basic_auth]: #basic_auth-block
+[authorization]: #authorization-block
+[oauth2]: #oauth2-block
+[tls_config]: #tls_config-block
+[label_selector]: #label_selector-block
+[match_expression]: #match_expression-block
+
+### label_selector block
+
+The `label_selector` block describes a kubernetes label selector for rule or namespace discovery.
```
karengermond, line 87:
```suggestion
The `label_selector` block describes a Kubernetes label selector for rule or namespace discovery.
```

docs/sources/flow/reference/components/mimir.rules.kubernetes.md
```diff
@@ -0,0 +1,202 @@
...
+
+The following arguments are supported:
+
+Name           | Type          | Description                                       | Default                     | Required
+---------------|---------------|---------------------------------------------------|-----------------------------|---------
+`match_labels` | `map(string)` | Label keys and values used to discover resources. | empty (match all resources) | yes
+
+### match_expression block
+
+The `match_expression` block describes a kubernetes label match expression for rule or namespace discovery.
```
karengermond, line 97:
```suggestion
The `match_expression` block describes a Kubernetes label match expression for rule or namespace discovery.
```

docs/sources/flow/reference/components/mimir.rules.kubernetes.md
```diff
@@ -0,0 +1,202 @@
...
+
+The following are exposed per discovered Mimir rule namespace resource:
+* The namespace name.
+* The number of rule groups.
+
+Only resources managed by the component are exposed - regardless of how many
+actually exist.
+
+## Debug metrics
+
```
karengermond, line 155: You might want to use a table here.

karengermond (approved):
LGTM, aside from minor edits.

docs/sources/flow/reference/components/mimir.rules.kubernetes.md
Logiraptor: Done in a0b5f4a

Logiraptor:
@rfratto All comments have been addressed 👍

nicoche:
Hi there!

Congrats on the work, I can't wait to run this in production 🙂. I've tried it on my lab, it worked great. Two things:

1 : I ran into an issue where the component would not boot correctly. cf. screenshot, it's neither healthy nor unhealthy, it did not finish booting. The issue was that the service account I used for Grafana Agent missed some capabilities. I used the one mentioned in the documentation (https://grafana.com/docs/grafana-cloud/kubernetes-monitoring/other-methods/k8s-agent-metrics/#deploy-grafana-agent-resources). Here is the diff for it to work:

```diff
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: grafana-agent
 rules:
+ - apiGroups: # This group is needed to list PrometheusRules
+  - "monitoring.coreos.com"
+  resources:
+  - prometheusrules
+  verbs:
+   - '*'
 - apiGroups:
   - ""
   resources:
   - nodes
   - nodes/proxy
   - nodes/metrics
   - services
   - endpoints
   - pods
   - events
+   - namespaces # This was missing to list namespaces
 (...)
```

image

This call blocks when the service account has not enough privileges: https://github.com/grafana/agent/blob/main/component/mimir/rules/kubernetes/rules.go#L304.

I believe that we should:
- At least edit the documentation to mention the permissions needed
- Potentially change the code such that we enforce a timeout on the blocking call _or_ verify the service account permissions when booting the component.

wdyt @rfratto  @Logiraptor  ?

2: I'm down to work on the AlertManager CRD integration. I believe that the approach could be largely similar to what has been done for PrometheusRules. wdyt? Shall I open a dedicated issue for that? On mimir and/or agent?

rfratto:
> I believe that we should:
>
> * At least edit the documentation to mention the permissions needed
> * Potentially change the code such that we enforce a timeout on the blocking call or verify the service account permissions when booting the component.

Definitely both great ideas 👍 

It might also be worth mentioning that the CRDs need to be installed in the cluster, though maybe that's obvious.

> 2: I'm down to work on the AlertManager CRD integration. I believe that the approach could be largely similar to what has been done for PrometheusRules. wdyt? Shall I open a dedicated issue for that? On mimir and/or agent?

Contributions here would be fantastic :) Creating an issue in the agent repo to track sounds good to me.

nicoche:
Gotcha, I popped an issue in the agent: https://github.com/grafana/agent/issues/2753. I'll try to open a PR in the coming days

rafilkmp3:
Can you provide an example creating this integration? I moved from Prometheus to Prometheus-agent and I need a way to make Mimir use my cluster prometheus rules and use mimir alert-manager to send alerts

Logiraptor:
@rafilkmp3 Have you checked the docs? https://grafana.com/docs/agent/latest/flow/reference/components/mimir.rules.kubernetes/#example

That should help point you in the right direction

rafilkmp3:
@Logiraptor  Yes but I can't find a way to make this work using grafana-agent-operator 

this is my configuration already with additional rbac

```
---
kind: GrafanaAgent
metadata:
  name: grafana-agent
  namespace: {{ .Release.Namespace }}
  labels:
    app: grafana-agent
spec:
  logLevel: info
  serviceAccountName: grafana-agent
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
          - matchExpressions:
              - key: eks.amazonaws.com/compute-type
                operator: NotIn
                values:
                  - fargate

  logs:
    instanceSelector:
      matchLabels:
        agent: grafana-agent

  integrations:
    selector:
      matchLabels:
        agent: grafana-agent

---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: grafana-agent
  namespace: {{ .Release.Namespace }}

---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: grafana-agent
  namespace: {{ .Release.Namespace }}
rules:
- apiGroups:
  - ""
  resources:
  - nodes
  - nodes/proxy
  - nodes/metrics
  - services
  - endpoints
  - pods
  - events
  - namespaces
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - networking.k8s.io
  resources:
  - ingresses
  verbs:
  - get
  - list
  - watch
- nonResourceURLs:
  - /metrics
  - /metrics/cadvisor
  verbs:
  - get
- apiGroups: # This group is needed to list PrometheusRules
  - "monitoring.coreos.com"
  resources:
  - prometheusrules
  verbs:
  - '*'
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: grafana-agent
  namespace: {{ .Release.Namespace }}
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: grafana-agent
subjects:
  - kind: ServiceAccount
    name: grafana-agent
    namespace: {{ .Release.Namespace }}

---
kind: Integration
metadata:
  labels:
    agent: grafana-agent
  name: agent-eventhandler
  namespace: {{ .Release.Namespace }}
spec:
  config:
    cache_path: /etc/eventhandler/eventhandler.cache
    logs_instance: {{ .Release.Namespace }}/grafana-agent-logs
  name: eventhandler
  type:
    unique: true
  volumeMounts:
    - mountPath: /etc/eventhandler
      name: agent-eventhandler
  volumes:
    - name: agent-eventhandler
      persistentVolumeClaim:
        claimName: agent-eventhandler

---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: agent-eventhandler
  namespace: {{ .Release.Namespace }}
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 1G
```

Can you help me know where to add this section ?

```
mimir.rules.kubernetes "default" {
    address = "GRAFANA_CLOUD_METRICS_URL"
    basic_auth {
        username = "GRAFANA_CLOUD_USER"
        password = "GRAFANA_CLOUD_API_KEY"
        // Alternatively, load the password from a file:
        // password_file = "GRAFANA_CLOUD_API_KEY_PATH"
    }
}
```

Logiraptor:
The agent operator doesn't support grafana agent flow as far as I know. It's really two separate modes of using the agent. I'd have to defer to someone on the agent team to help from here, as this was a cross-team hackathon project for me and I'm not actually working on the agent full time.  

Pinging @rfratto to help!

rfratto:
Unfortunately, Grafana Agent Operator is focused around static mode, which doesn't support PrometheusRule CRDs. 

Grafana Agent Flow is currently the only Grafana Agent variant which supports reading PrometheusRule CRDs, and we don't have any plans right now to add support for it to static mode or the static mode operator.

PR #2616, opened 2022-12-14
add multi label and remove fsnotify

mattdurham:


component/discovery/file/file.go
```diff
@@ -116,239 +77,56 @@ func (c *Component) Update(args component.Arguments) error {
...
+		// The component node checks to see if exports have actually changed.
+		c.opts.OnStateChange(Exports{Targets: paths})
+
 		// Check to see if our ticker timer needs to be reset.
-		if timerDuration != c.args.UpdatePeriod {
-			watchDog.Reset(c.args.UpdatePeriod)
-			timerDuration = c.args.UpdatePeriod
+		if timerDuration != c.args.SyncPeriod {
+			watchDog.Reset(c.args.SyncPeriod)
+			timerDuration = c.args.SyncPeriod
```
tpaschalis, line 104: Do we need to reset the ticker timer in the Update method instead?

I think that here, if we had an initial SyncPeriod of 1h, and reloaded config to change it to 20s, the change would only take place after the 1h has elapsed first, right?

mattdurham: That makes sense

tpaschalis (approved):
LGTM. I think we can merge this on to the proper PR branch and see how it all works together.

karengermond (commented):
LGTM!

PR #2629, opened 2022-12-18
Bump node_exporter version to v1.5.0

Thor77:
#### PR Description
Update node_exporter integration to v1.5.0
Specific reason for the update: in v1.3.1 the zfs_zpool_state metric is not available with my ZFS version, [prometheus/node_exporter#2451](https://github.com/prometheus/node_exporter/pull/2451) solves that issue and is included in v1.4.0.

#### PR Checklist

- [x] CHANGELOG updated


tpaschalis:
Hey there, thanks for the PR! 👋 

Just a small note, the version listed in the go.mod file `github.com/prometheus/node_exporter v1.3.1-0.20220127103407-4f27a4fd8efa` is a little misleading; the first part `v1.3.1` is a [pseudo-version](https://go.dev/ref/mod#pseudo-versions) (not really sure why go.mod uses that). The actual commit that we depend on is `4f27a4fd8efa`.

Could you also take a look at the failing test and mention the new collectors that were introduced in node_exporter's [1.4.0](https://github.com/prometheus/node_exporter/releases/tag/v1.4.0) in our documentation?

Thor77:
@tpaschalis Thanks for taking a look!
I see, thanks for the explanation, that pseudo version still doesn't "contain" the PR fixing the ZFS issue, though.

The test is failing because the flag is available for Darwin now (collector was refactored in https://github.com/prometheus/node_exporter/pull/2417).
I didn't realize it would need this many changes but I added documentation for the new collectors, a migration for the deprecated diskstats ignored devices flag and configuration options for the new sysctl collector.
Please let me know if I missed something or should implement it differently.
I'm particularly unsure about the changes in component/prometheus/integration/node_exporter as there doesn't seem to be a migration path for those? Should I just leave the config names as is then?

karengermond (approved):
LGTM

tpaschalis:
Hey Jonas! This looks reasonable at a first glance, but would like to sanity check it. Just a heads up, reviews might be a little slower due to most of the reviewers being away on PTO, but it shouldn't be too long!

pkg/integrations/node_exporter/config.go
```diff
@@ -102,7 +102,8 @@ type Config struct {
 	CPUEnableCPUGuest                bool                `yaml:"enable_cpu_guest_seconds_metric,omitempty"`
 	CPUEnableCPUInfo                 bool                `yaml:"enable_cpu_info_metric,omitempty"`
 	CPUFlagsInclude                  string              `yaml:"cpu_flags_include,omitempty"`
-	DiskStatsIgnoredDevices          string              `yaml:"diskstats_ignored_devices,omitempty"`
+	DiskStatsDeviceExclude           string              `yaml:"diskstats_device_exclude,omitempty"`
+	DiskStatsDeviceInclude           string              `yaml:"diskstats_device_include,omitempty"`
```
rfratto, line 106: This would be considered a breaking change, so we'd need to document it as such in the upgrade guide.

rfratto (commented):
Thanks for the PR! This needs a look from the CODEOWNERS for final review, but I left a quick comment :)

pkg/integrations/node_exporter/config.go
Thor77: This shouldn't be *breaking* because there's a migration from the old to the new value. But I added a note to the changelog and upgrade-guide and also rebased on main to resolve the merge conflict.

pkg/integrations/node_exporter/config.go
```diff
@@ -127,6 +128,8 @@ type Config struct {
 	PowersupplyIgnoredSupplies       string              `yaml:"powersupply_ignored_supplies,omitempty"`
 	RunitServiceDir                  string              `yaml:"runit_service_dir,omitempty"`
 	SupervisordURL                   string              `yaml:"supervisord_url,omitempty"`
+	SysctlInclude                    flagext.StringSlice `yaml:"sysctl_include,omitempty"`
```
marctc, line 131: Probably we want to expose this values here too: [component/prometheus/integration/node_exporter/config.go](https://github.com/grafana/agent/blob/main/component/prometheus/integration/node_exporter/config.go)

Thor77: added and also rebased again because of the merge conflicts in go.mod

CHANGELOG.md
```diff
@@ -25,6 +26,10 @@ v0.31.0-rc.0 (2023-01-26)
   - `agentctl` is now `grafana-agentctl`.
   - `agent-operator` is now `grafana-agent-operator`.
 
+- Node Exporter configuration options changed to align with new upstream version (@Thor77):
```
marctc, line 29: This should be in Main (unreleased) section

Thor77: Oh, I messed this up while rebasing, fixed.

Issue #2644, opened 2022-12-22
Release v0.30.0 no longer emits traces_receiver_* metrics


modulitos:
### Repro:
Start the `grafana/agent:v0.30.0` image with `/bin/agent -config.file=/etc/agent/agent.yaml -server.http.address=0.0.0.0:80 -config.expand-env`.

Here's my `/etc/agent/agent.yaml`, but I'm not sure if the specifics of this file matters for this bug:
```
server:
    log_level: info
    log_format: json
traces:
    configs
      - name: zipkin_traces
        batch:
            send_batch_size: 1000
            timeout: 5s
        receivers:
            zipkin:
                endpoint: 0.0.0.0:9411
        remove_write:
            //etc
        tail_sampling:
            //etc
```
Then run `curl localhost:80/metrics | grep receiver` to see the receiver metrics.

### Expected
We should see the `traces_receiver_accepted_spans` and `traces_receiver_accepted_spans` metrics in the output. For example:

```
# HELP traces_receiver_accepted_spans Number of spans successfully pushed into the pipeline.
# TYPE traces_receiver_accepted_spans counter
traces_receiver_accepted_spans{receiver="zipkin",traces_config="zipkin_traces",transport="http_v2_json"} 827851
# HELP traces_receiver_refused_spans Number of spans that could not be pushed into the pipeline.
# TYPE traces_receiver_refused_spans counter
traces_receiver_refused_spans{receiver="zipkin",traces_config="zipkin_traces",transport="http_v2_json"} 0
```
This is the case when I do all of the above, but with image `grafana/agent:v0.29.0` instead.

### Actual
There is no output for any metrics matching `receiver` on v0.30.0.

---

It seems like these `traces_receiver_*` metrics are expected - at least according to the Troubleshooting docs here: https://grafana.com/docs/tempo/latest/troubleshooting/agent/

Could this be a bug from the `v0.30.0` release?

tpaschalis:
Hey there, thanks for the report! 👋 

We have identified the issue about being about the way that our OpenTelemetry Collector dependency exposes these metrics in their v0.63. We'll share news about a fix in due time.

tpaschalis:
Closing this, as it has been fixed with the v0.30.1 patch release.

Again, thanks for bringing this to our attention!

modulitos:
@tpaschalis I can confirm this bug is fixed. Thank you for the quick fix!

PR #2647, opened 2022-12-23
Enable useOtelForInternalMetrics only when Flow mode is on

marctc:
#### PR Description

This PR disables the usage `useOtelForInternalMetrics` when Flow is not running.
This code path is being executed regardless Flow is one causing that some metrics
are not collected for tracing (see https://github.com/grafana/agent/issues/2644)

Fixes https://github.com/grafana/agent/issues/2644

#### Notes to the Reviewer

This will be included in the next release patch.

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/component/otelcol/internal/featuregate/install.go
+++ b/component/otelcol/internal/featuregate/install.go
@@ -4,6 +4,8 @@
 package featuregate
 
 import (
+	"os"
+
 	"go.opentelemetry.io/collector/featuregate"
 
 	_ "go.opentelemetry.io/collector/obsreport" // telemetry.useOtelForInternalMetrics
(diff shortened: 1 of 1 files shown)


component/otelcol/internal/featuregate/install.go
```diff
@@ -19,7 +21,26 @@ func init() {
...
 }
+
+func isFlowRunning() bool {
+	key, found := os.LookupEnv("AGENT_MODE")
+	if !found {
+		key, found := os.LookupEnv("EXPERIMENTAL_ENABLE_FLOW")
+		if !found {
+			return false
+		}
+		return key == "true" || key == "1"
```
ptodev, line 38: Should any of these values be case sensitive? Also, is it not better to get whether Flow is running from some kind of internal variable which is set on startup? The reason is that there would only be one place where we check the "EXPERIMENTAL_ENABLE_FLOW" and "AGENT_MODE" variables.

ptodev (approved):
I can't say I understand the change very well, but I'll trust your judgement that this is what we need 👍  I see that Paschalis mentioned more info will be shared soon.

component/otelcol/internal/featuregate/install.go
marctc: I copied the code from here: https://github.com/grafana/agent/blob/main/cmd/agent/mode.go#L17 I thought reusing that code but is not possible as it's in `cmd/main`. A refactor would be needed but since it's a quickfix, I rather copy this bit for now.

marctc:
> I can't say I understand the change very well,

Me neither, but due our capacity right now, is the best what I can come up for now.

PR #2661, opened 2022-12-30
Agent Management Initial

spartan0x117:
#### PR Description
Add configuration to poll a remote API for new configs in the agent. If the API call fails, falls back to a locally cached file of the most recent good config.

#### PR Checklist

- [X] CHANGELOG updated
- [ ] Documentation added
- [X] Tests updated


pkg/config/agentmanagement.go
```diff
@@ -0,0 +1,188 @@
...
+	"github.com/prometheus/common/config"
+)
+
+type BasicAuth struct {
+	Username     string `yaml:"username"`
+	PasswordFile string `yaml:"password_file"`
+}
+
+type RemoteConfiguration struct {
+	Labels       []string `yaml:"labels"`
```
marctc, line 25: I was trying to configure this and it took me a while to figure out how to do it. Why not do it more yaml-esque and do it something like this?

```
    labels:
      - team: team-a
      - hostname machine-1
```

spartan0x117: Ah yeah, this is a holdover from them being CLI flags. I'll fix this

spartan0x117: Should be resolved now, it now parses to a `map[string]string`, so specifying labels would look like:
```yaml
labels:
  team: "team-a"
  hostname: "machine-1"
```

cmd/agent/entrypoint.go
```diff
@@ -352,6 +352,22 @@ func (ep *Entrypoint) TriggerReload() bool {
 	return true
 }
 
+// Triggers a reload of the config on each tick of the ticker until the context
+// completes.
+func (ep *Entrypoint) PollConfig(ctx context.Context, t *time.Ticker) error {
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
```
rfratto, line 361: Generally we only return `ctx.Err()` if the canceled context means something failed. In this case, it's expected for the function to poll forever until the context is canceled, so this should be `return nil`.

cmd/agent/entrypoint.go
```diff
@@ -352,6 +352,22 @@ func (ep *Entrypoint) TriggerReload() bool {
 	return true
 }
 
+// Triggers a reload of the config on each tick of the ticker until the context
+// completes.
+func (ep *Entrypoint) PollConfig(ctx context.Context, t *time.Ticker) error {
```
rfratto, line 357: This doesn't need to be an exposed function :) I'd like for us to try to keep the public API as small as we can.

pkg/config/agentmanagement.go
```diff
@@ -0,0 +1,165 @@
...
+	Protocol        string    `yaml:"protocol"`
+	PollingInterval string    `yaml:"polling_interval"`
+
+	RemoteConfiguration RemoteConfiguration `yaml:"remote_configuration"`
+}
+
+// Gets the remote config specified in the initial config, falling back to a local, cached copy
+// of the remote config if the request to the remote fails. If both fail, an empty config and an
+// error will be returned.
+func GetRemoteConfig(dir string, expandEnvVars bool, initialConfig *Config, log *server.Logger) (*Config, error) {
```
rfratto, line 43: Ditto here: This also doesn't need to be exposed

pkg/config/config.go
```diff
@@ -232,6 +244,29 @@ func LoadFile(filename string, expandEnvVars bool, c *Config) error {
 	return LoadBytes(buf, expandEnvVars, c)
 }
 
+// Loads and merges a config from an Agent Management API.
+//  1. Read local initial config.
+//  2. Get the remote config.
+//     a) Fetch from remote. If this fails or is invalid:
+//     b) Read the remote config from cache. If this fails, return an error.
+//  4. Merge the initial and remote config into c.
+func LoadFromAgentManagementAPI(path string, expandEnvVars bool, c *Config, log *server.Logger) error {
```
rfratto, line 253: Ditto (doesn't need to be exposed)

pkg/config/config.go
```diff
@@ -232,6 +244,29 @@ func LoadFile(filename string, expandEnvVars bool, c *Config) error {
...
+	}
+
+	remoteConfig, err := GetRemoteConfig(filepath.Dir(path), expandEnvVars, c, log)
+	if err != nil {
+		return err
+	}
+	if err = mergo.Merge(c, remoteConfig); err != nil {
+		return fmt.Errorf("error trying to merge initial and remote configs: %w", err)
+	}
+	return nil
```
rfratto, line 267: This worries me a little bit, because reasoning about merge behaviors seems challenging. Is this what we're expecting to happen? I was under the impression that the remote config replaces the local config.

spartan0x117: Yeah, had a discussion and this will be replaced with a shallow, top-level merge, e.g.:
```golang
for k, v := range remoteConfig {
     c[k] = v
}
```

cmd/grafana-agent/entrypoint.go
```diff
@@ -352,6 +352,22 @@ func (ep *Entrypoint) TriggerReload() bool {
 	return true
 }
 
+// Triggers a reload of the config on each tick of the ticker until the context
```
marctc, line 355:
```suggestion
// pollConfig triggers a reload of the config on each tick of the ticker until the context
```
Comments for function in Go always start with the name of the function. This is needed for generating the godoc, in case of exported functions, but for consistency it's a good practice to use same format for unexported functions.

spartan0x117: Ah good to know, I'll go through all of the added functions and update their comments!

pkg/config/instrumentation/remote_config_metrics.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+var RemoteConfigMetrics = newRemoteConfigMetrics()
+
+func newRemoteConfigMetrics() *remoteConfigMetrics {
+	var remoteConfigMetrics remoteConfigMetrics
+
+	remoteConfigMetrics.fetchStatusCodes = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "remote_config_fetches",
```
marctc, line 22: It's usally good practice to add `_total` as suffix

pkg/config/agentmanagement.go
```diff
@@ -0,0 +1,165 @@
...
+
+type RemoteConfiguration struct {
+	Labels       labelMap `yaml:"labels"`
+	Namespace    string   `yaml:"namespace"`
+	BaseConfigId string   `yaml:"base_config_id"`
+}
+
+type BasicAuth struct {
+	Username     string `yaml:"username"`
+	PasswordFile string `yaml:"password_file"`
```
marctc, line 27: `Password` should be a valid option

jcreixell (approved):

pkg/config/config.go
```diff
@@ -224,6 +224,7 @@ func (c *Config) RegisterFlags(f *flag.FlagSet) {
 		"basic auth username for fetching remote config. (requires remote-configs experiment to be enabled")
 	f.StringVar(&c.BasicAuthPassFile, "config.url.basic-auth-password-file", "",
 		"path to file containing basic auth password for fetching remote config. (requires remote-configs experiment to be enabled")
+	f.StringVar(&c.AgentManagement.CacheLocation, "agentmanagement.cache-location", "", "Path to writeable directory where remote configs will be cached to disk")
```
jcreixell, line 227: let's specify a default here? for example, `-metrics.wal-directory` defaults to `data-agent/`, so it is probably a good default

cmd/grafana-agent/entrypoint.go
```diff
@@ -400,6 +416,20 @@ func (ep *Entrypoint) Start() error {
...
 
+	if ep.cfg.AgentManagement.Enabled {
+		managementContext, managementCancel := context.WithCancel(context.Background())
+		defer managementCancel()
+
+		// By this point ep.cfg.Validate() has already been called, so the configured time must be valid
+		sleepTime, _ := ep.cfg.AgentManagement.SleepTime()
+		t := time.NewTicker(sleepTime)
+		g.Add(func() error {
+			return ep.pollConfig(managementContext, t)
```
tpaschalis, line 427: Are we only setting the ticker interval once at the start, or is there a way to reset it? I couldn't find where this would happen at a glance.

pkg/config/config.go
jcreixell: @spartan0x117 I tried to run this via config flags but it is not working for me right now, getting:

```
2023/01/17 15:58:54 error loading config file remote-config.yaml: invalid initial config: path to cache must be specified in 'remote_config_cache_location'
```

cmd/grafana-agent/entrypoint.go
spartan0x117: The initial intention was not to allow this, but perhaps if so this should be a required CLI flag instead of a config element? cc: @jcreixell

spartan0x117: We also may leave this for a separate followup PR

pkg/config/agentmanagement.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+}
+
+// fetchConfig fetches the raw bytes of the config from the API specified in c.
+func fetchConfig(c *Config) ([]byte, error) {
+	httpClientConfig := &config.HTTPClientConfig{
+		BasicAuth: &c.AgentManagement.BasicAuth,
+	}
+
+	dir, err := os.Getwd()
```
mattdurham, line 93: How does this behave on windows installed as a service?

spartan0x117: That's a good question (that I don't know the answer to, yet). This would only affect the case where the password file is being specified as a relative path ([relevant code where that dir would be used](https://github.com/prometheus/common/blob/main/config/config.go#L83)). An alternative would be to require that the password file be specified as an absolute path, at least for now?

spartan0x117: Well, I guess the answer would be: if the password file is specified with a relative path, it will look for the password file relative to the working directory of the windows service. (But, I don't know what that working directory would typically be)

PR #2664, opened 2023-01-04
Ensure no empty fields are marshaled in YAML

jcreixell:
#### PR Description

  - Remote config merging requires marshaling valid but minimal configurations to avoid returning more fields than necessary, which may lead to unwanted overrides in the local and default configurations. 
  - This change makes sure no empty fields are marshaled.

#### Notes to the Reviewer

  - Although we omit most of the empty fields already, some types (especially imported ones) require implementing the IsZeroer interface for `omitempty` to properly work.
  - Since we cannot extend an external type with this interface, we need to define wrapper types.
  - Although this is relatively intrusive, it is probably safer than replacing the existing fields with pointers.
  - See https://pkg.go.dev/gopkg.in/yaml.v2#IsZeroer
  - See https://sentry.io/answers/alias-type-definitions/#the-solution

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -452,3 +452,10 @@ func TestLoadDynamicConfigurationExpandError(t *testing.T) {
 	assert.Error(t, err)
 	assert.True(t, strings.Contains(err.Error(), "expand var is not supported when using dynamic configuration, use gomplate env instead"))
 }
+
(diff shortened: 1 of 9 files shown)


rfratto (approved):
Thanks!

PR #2675, opened 2023-01-05
Switch from nanoseconds to seconds.

mattdurham:
#### PR Description

This fixes an issue where Windows runs would fail with the nanosecond sync period

- [X] Tests updated


rfratto (approved):
It seems like this definitely fixed the issue, but if nothing else it's an improvement. LGTM!

PR #2676, opened 2023-01-05
tools/agentlint: Lint for documentation on exported identifiers

rfratto:
golint was previously used for linting, which would enforce that all exported identifiers (i.e., types, variables, functions, and methods which are part of the public API of a package) have documentation. 

Since golint is no longer maintained, we have lost this linting rule, which I personally think is important to write good APIs. While one could argue methods like `String() string` "don't need documentation," the Go standard library shows that [you can still document the method with useful information](https://pkg.go.dev/time#Duration.String). 

golint had one exception: you did not need to document the methods used for implementing `sort.Interface`. That exception has been replicated here.

### DIFF
--- a/tools/agentlint/internal/exportedcomments/exportedcomments.go
+++ b/tools/agentlint/internal/exportedcomments/exportedcomments.go
@@ -1,0 +1,146 @@
+// Package exportedcomments exposes an Analyzer which will validate that all
+// exported identifiers have comments.
+package exportedcomments
+
+import (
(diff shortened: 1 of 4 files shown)


rfratto:
(The lint check is failing because it's calling out undocumented identifiers in existing code. I'll fix those if we want to merge this)

rfratto:
I'm going to close this for now. If we start enforcing this rule, we may want to consider reopening.

PR #2677, opened 2023-01-05
Prefix published binaries with `grafana-*`

### LINKED ISSUE #582: Rename binary to grafana-agent-* ?

rfratto:
* `agent` -> `grafana-agent`
* `agentctl` -> `grafana-agentctl`
* `agent-operator` -> `grafana-agent-operator`

Other binaries which are not published (agentlint, smoke, crow) have been left unrenamed.

Docker containers have been given a symbolic link so the old binary names (`/bin/agent*`) continue to work. These symbolic links will be removed in v0.33.

Fixes #582.


cmd/grafana-agent-operator/Dockerfile
```diff
@@ -31,5 +31,6 @@ RUN <<EOF
   rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 EOF
 
-COPY --from=build /src/agent/build/agent-operator /bin/agent-operator
-ENTRYPOINT ["/bin/agent-operator"]
+COPY --from=build /src/agent/build/agent-operator /bin/grafana-agent-operator
+RUN ln -sv /bin/grafana-agent-operator /bin/agent-operator
```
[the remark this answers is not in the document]
mattdurham: NVM, just needed to read further.

packaging/windows/install_script.nsis
```diff
@@ -194,7 +194,7 @@ Section "uninstall"
     Pop $0
 
     # Remove files
-    delete $INSTDIR\agent-windows-amd64.exe
+    delete $INSTDIR\grafana-agent-windows-amd64.exe
```
mattdurham, line 197: My gut says we should delete the old name here too.

mattdurham (approved):
One minor nit about deleting the old name though I don't think it's entirely necessary it will help for upgrades and then deletes. (which are probably rare)

packaging/windows/install_script.nsis
mattdurham: It does not https://nsis.sourceforge.io/Reference/Delete `The error flag is not set from trying to delete a file that does not exist. `

rfratto: 👍 thanks!

PR #2683, opened 2023-01-06
Flow UI: Long string out of bounds fix

### LINKED ISSUE #2207: Flow UI: long strings overflow box
Very long strings (such as very long labels on targets) will overflow the box they're in. We should either allow the box to scroll horizontally or allow word breaks to occur anywhere.

xiyu95:
#### PR Description

Fix the issue with long string going out of bounds. Screenshots of comparison between before and after the fix are shown below:

BEFORE 
Screen Shot 2023-01-05 at 4 08 45 PM

AFTER
Screen Shot 2023-01-05 at 4 08 11 PM

Fixes #2207 

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated


rfratto:
Was the monospace font removed from the entries in the name column on purpose? 

image

image

I personally preferred them being monospace since they're always code identifiers with underscores.

xiyu95:
Nice catch! It's probably being removed accidentally when I was changing the css. I will take a look at it.

xiyu95:
I just added the monospace font style back to the name column of the table. It was caused by the removal of the table style from the css file. Screenshot below:
Screen Shot 2023-01-06 at 10 26 07 AM

PR #2688, opened 2023-01-06
prometheus.operator.podmonitors component

### LINKED ISSUE #3382: Flow: add `prometheus.operator.podmonitors` component

captncraig:

Fixes #3382 

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated
- [x] Flag component as unhealthy on initialization failure


component/discovery/kubernetes_crds/kubernetes_crds.go
```diff
@@ -0,0 +1,272 @@
...
+	"k8s.io/client-go/tools/clientcmd"
+
+	v1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	promop "github.com/prometheus-operator/prometheus-operator/pkg/client/informers/externalversions"
+	versioned "github.com/prometheus-operator/prometheus-operator/pkg/client/versioned"
+)
+
+func init() {
+	component.Register(component.Registration{
+		Name:    "discovery.kubernetes_crds",
```
rfratto, line 31: I don't know if this is the right name for the component, since it's less about just dicsovering things and more focused on telemetry collection. Personally I'd go for `prometheus.kubernetes.pod_monitors`.

component/discovery/kubernetes_crds/kubernetes_crds.go
```diff
@@ -0,0 +1,272 @@
...
+		Args:    Config{},
+		Exports: struct{}{},
+
+		Build: func(opts component.Options, args component.Arguments) (component.Component, error) {
+			return New(opts, args)
+		},
+	})
+}
+
+type Config struct {
```
rfratto, line 41: I'd like to see selectors here similar to what one does with MetricsInstances: a label selector for which PodMonitors to find and a label selector which which namespaces to search for PodMonitors in.

component/discovery/kubernetes_crds/kubernetes_crds.go
```diff
@@ -0,0 +1,272 @@
...
+	BasicAuth *commonConfig.BasicAuth `json:"basicAuth,omitempty"`
+	// Bearer token for accessing apiserver.
+	BearerToken string `json:"bearerToken,omitempty"`
+	// File to read bearer token for accessing apiserver.
+	BearerTokenFile string `json:"bearerTokenFile,omitempty"`
+	// TLS Config to use for accessing apiserver.
+	TLSConfig commonConfig.TLSConfig `json:"tlsConfig,omitempty"`
+	// Authorization section for accessing apiserver
+	Authorization commonConfig.Authorization `json:"authorization,omitempty"`
+}
```
rfratto, line 72: Should the JSON tags here be river?

component/discovery/kubernetes_crds/kubernetes_crds.go
```diff
@@ -0,0 +1,272 @@
...
+	pm := oldObj.(*v1.PodMonitor)
+	level.Info(c.opts.Logger).Log("msg", "found pod monitor", "name", pm.Name)
+	c.clearConfigs("podMonitor", pm.Namespace, pm.Name)
+	c.addConfig(newObj.(*v1.PodMonitor))
+}
+func (c *Component) OnDeletePodMonitor(obj interface{}) {
+	pm := obj.(*v1.PodMonitor)
+	c.clearConfigs("podMonitor", pm.Namespace, pm.Name)
+	c.apply()
+}
```
rfratto, line 167: These three methods can be unexported; they don't need to be a part of the public API.

component/discovery/kubernetes_crds/promcfg.go
```diff
@@ -0,0 +1,388 @@
...
+	//TODO:
+	//cfg = addSafeTLStoYaml(cfg, m.Namespace, ep.TLSConfig.SafeTLSConfig)
+	//}
+
+	//TODO: Secret store needs to be figured out
+	// if ep.BearerTokenSecret.Name != "" {
+	// 	if s, ok := store.TokenAssets[fmt.Sprintf("podMonitor/%s/%s/%d", m.Namespace, m.Name, i)]; ok {
+	// 		cfg = append(cfg, yaml.MapItem{Key: "bearer_token", Value: s})
+	// 	}
+	// }
```
rfratto, line 133: Can't the component request the content of secrets from the K8s API client? It would search in the same namespace as the PodMonitor.

rfratto (commented):
nice, I'm excited to have this in. Gave a quick first review, I'll give something more in depth once you're ready for a deeper look

component/discovery/kubernetes_crds/promcfg.go
```diff
@@ -0,0 +1,469 @@
...
+			cfg.Regex = relabel.DefaultRelabelConfig.Regex
+		}
+		if cfg.Replacement == "" {
+			cfg.Replacement = relabel.DefaultRelabelConfig.Replacement
+		}
+		r.configs = append(r.configs, cfg)
+	}
+}
+
+// addFromMonitoring converts from an externally generated monitoringv1 RelabelConfig
```
jcreixell, line 375:
```suggestion
// addFromV1 converts from an externally generated monitoringv1 RelabelConfig
```

component/discovery/kubernetes_crds/promcfg.go
```diff
@@ -0,0 +1,469 @@
...
+
+func parseRegexp(s string, l log.Logger) relabel.Regexp {
+	r, err := relabel.NewRegexp(s)
+	if err != nil {
+		level.Error(l).Log("msg", "failed to parse regex", "err", err)
+	}
+	return r
+}
+
+func (cg *configGenerator) generatePodMonitorConfig(m *v1.PodMonitor, ep v1.PodMetricsEndpoint, i int) *config.ScrapeConfig {
```
jcreixell, line 106: this probably needs to be heavily unit tested

component/discovery/kubernetes_crds/config.go
```diff
@@ -0,0 +1,48 @@
...
+	// and will discover API servers automatically and use the pod's CA certificate
+	// and bearer token file at /var/run/secrets/kubernetes.io/serviceaccount/.
+	ApiServerConfig *APIServerConfig `river:"api_server,block,optional"`
+
+	ForwardTo []storage.Appendable `river:"forward_to,attr"`
+}
+
+// APIServerConfig defines a host and auth methods to access apiserver.
+// More info: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#kubernetes_sd_config
+// TODO: river!
```
jcreixell, line 25: could you elaborate?

jcreixell (commented):
Gave it a superficial pass and overall LGTM. I think we need a lot of tests to gain confidence that the config generation is doing what we expect. Other than that, I think it is relatively straightforward

component/discovery/kubernetes_crds/config.go
```diff
@@ -0,0 +1,74 @@
...
+
+	ForwardTo []storage.Appendable `river:"forward_to,attr"`
+
+	Namespaces    []string `river:"namespaces,attr,optional"`
+	LabelSelector string   `river:"label_selector,optional"`
+	FieldSelector string   `river:"field_selector,optional"`
+}
+
+func (args *Config) UnmarshalRiver(f func(interface{}) error) error {
+	*args = Config{}
```
rfratto, line 31: Do you intend to give this a default config? If not, we can take this out.

component/discovery/kubernetes_crds/config.go
```diff
@@ -0,0 +1,74 @@
...
+
+	commonConfig "github.com/grafana/agent/component/common/config"
+	promConfig "github.com/prometheus/common/config"
+	"github.com/prometheus/prometheus/storage"
+	apiv1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+type Config struct {
```
rfratto, line 14: For consistency with other components, this type should be called `Arguments`.

component/discovery/kubernetes_crds/config.go
```diff
@@ -0,0 +1,74 @@
+package kubernetes_crds
```
rfratto, line 1: For consistency with other components, this file should be called types.go.

component/discovery/kubernetes_crds/kubernetes_crds.go
```diff
@@ -0,0 +1,336 @@
...
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+
+	v1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	"github.com/prometheus-operator/prometheus-operator/pkg/client/versioned"
+)
+
+func init() {
+	component.Register(component.Registration{
+		Name:    "prometheus.crds",
```
rfratto, line 31: Is the intent that this one component handles all the Prometheus CRD types? I need to think about that one a little, I'm not sure if that's going to complicate the debug information or the configuration at all, and it turns this component into a bit of a monolith.

component/discovery/kubernetes_crds/kubernetes_crds.go
```diff
@@ -0,0 +1,336 @@
+package kubernetes_crds
```
rfratto, line 1: The package name and path needs to be updated to sync with the component name.

component/discovery/kubernetes_crds/kubernetes_crds.go
```diff
@@ -0,0 +1,336 @@
...
+
+	v1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	"github.com/prometheus-operator/prometheus-operator/pkg/client/versioned"
+)
+
+func init() {
+	component.Register(component.Registration{
+		Name:    "prometheus.crds",
+		Args:    Config{},
+		Exports: struct{}{},
```
rfratto, line 33: If you're not exporting anything, leave this out: 

```suggestion
```

component/discovery/kubernetes_crds/kubernetes_crds.go
```diff
@@ -0,0 +1,336 @@
...
+	config *Config
+
+	discovery        *discovery.Manager
+	scraper          *scrape.Manager
+	cg               configGenerator
+	discoveryConfigs map[string]discovery.Configs
+	scrapeConfigs    map[string]*config.ScrapeConfig
+
+	mut sync.Mutex
+}
```
rfratto, line 63: I'd prefer to see the crdManager stuff in a separate file to make things a little more separated.

component/discovery/kubernetes_crds/kubernetes_crds.go
```diff
@@ -0,0 +1,336 @@
...
+func (c *crdManager) onUpdateServiceMonitor(oldObj, newObj interface{}) {
+	sm := oldObj.(*v1.PodMonitor)
+	c.clearConfigs("serviceMonitor", sm.Namespace, sm.Name)
+	c.addServiceMonitor(newObj.(*v1.ServiceMonitor))
+}
+func (c *crdManager) onDeleteServiceMonitor(obj interface{}) {
+	sm := obj.(*v1.ServiceMonitor)
+	c.clearConfigs("serviceMonitor", sm.Namespace, sm.Name)
+	c.apply()
+}
```
rfratto, line 174: It doesn't feel necessary to have three different add/update/delete functions. You could have one of each with a type switch:

```
func (c *crdManager) onAdd(obj interface{}) {
  switch obj := obj.(type) {
  case *v1.ServiceMonitor:
    // ...

  case *v1.PodMonitor:
    // ... 

  case *v1.Probe:
    // ... 

  default:
    // log an error, presumably 
  }
}
```

component/discovery/kubernetes_crds/kubernetes_crds.go
```diff
@@ -0,0 +1,336 @@
...
+
+// Run implements component.Component.
+func (c *Component) Run(ctx context.Context) error {
+	// innerCtx gets passed to things we create, so we can restart everything anytime we get an update.
+	// Ideally, this component has very little dynamic config, and won't have frequent updates.
+	var innerCtx context.Context
+	// cancel is the func we use to trigger a stop to all downstream processors we create
+	var cancel func()
```
rfratto, line 198: I'd recommend calling cancel in a defer to be a little safer around ensuring it always gets canceled at the end: 

```go 
var cancel func() 
defer func() {
  if cancel != nil {
    cancel() 
  }
}()
```

component/discovery/kubernetes_crds/kubernetes_crds.go
```diff
@@ -0,0 +1,336 @@
...
+func (c *crdManager) run(ctx context.Context) {
+
+	config, err := c.config.restConfig()
+	if err != nil {
+		level.Error(c.logger).Log("msg", "failed to create rest config", "err", err)
+		return
+	}
+	clientset, err := kubernetes.NewForConfig(config)
+
+	promClientset := versioned.New(clientset.RESTClient())
```
rfratto, line 229: This seems risky; if NewForConfig fails, won't clientset be nil? We should move the error check to happen earlier.

component/discovery/kubernetes_crds/kubernetes_crds.go
```diff
@@ -0,0 +1,336 @@
...
+	c.scraper = scrape.NewManager(opts, c.logger, flowAppendable)
+	defer c.scraper.Stop()
+	targetSetsChan := make(chan map[string][]*targetgroup.Group)
+	go func() {
+		err := c.scraper.Run(targetSetsChan)
+		level.Info(c.logger).Log("msg", "scrape manager stopped")
+		if err != nil {
+			level.Error(c.logger).Log("msg", "scrape manager failed", "err", err)
+		}
+	}()
```
rfratto, line 313: I'm not sure all of these need to be associated with the `crdManager`. As it is now, if you change the config for the component, all scraping stops, which will interrupt the flow of metrics.

At least the scraper and appendable should be handled a layer up the goroutine tree, potentially also the discovery manager.

component/discovery/kubernetes_crds/kubernetes_crds.go
```diff
@@ -0,0 +1,336 @@
...
+			c.mut.Lock()
+			componentCfg := c.config
+			c.mut.Unlock()
+			crdMan := newManager(c.opts, c.opts.Logger, componentCfg)
+			go crdMan.run(innerCtx)
+		}
+	}
+}
+
+func (c *crdManager) run(ctx context.Context) {
```
rfratto, line 220: Because all the component updates are handled completely in the background, you should implement `HealthComponent` and set the component to unhealthy if an error occurs in `run`.

component/discovery/kubernetes_crds/probe.go
```diff
@@ -0,0 +1,232 @@
...
+		if u, err := url.Parse(m.Spec.ProberSpec.ProxyURL); err != nil {
+			level.Error(cg.logger).Log("msg", "failed to parse ProxyURL from probe", "err", err)
+		} else {
+			cfg.HTTPClientConfig.ProxyURL = commonConfig.URL{URL: u}
+		}
+	}
+	return cfg
+}
+
+// 	if m.Spec.Module != "" {
```
rfratto, line 41: What's going on here? If we're going to leave commented code in like this, I'd like another comment explaining what it is and why it's being left in the file.

component/discovery/kubernetes_crds/configGenerator.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+	return cfg
+}
+
+func parseRegexp(s string, l log.Logger) relabel.Regexp {
+	r, err := relabel.NewRegexp(s)
+	if err != nil {
+		level.Error(l).Log("msg", "failed to parse regex", "err", err)
+	}
+	return r
```
rfratto, line 78: We shouldn't just be dropping the errors here. This should bubble up to the caller so we can set information in the DebugInfo to indicate there's a problem with the resource we discovered.

rfratto (changes requested):
Thanks! I did a second pass of the code here. Things are looking good, but honestly this PR is doing way too much, and it's very hard to review fully in its current state. I'd like it if we could break this up into multiple PRs, where the first PR delivers some minimal MVP which works end-to-end with debug information, and we add on additional features/capabilities that were left out of the MVP (like ability to read secrets) in subsequent PRs.

component/discovery/kubernetes_crds/kubernetes_crds.go
captncraig: I think we're covered if the parent context cancels, but I have no problem being explicit for safety.

component/prometheus/kubernetes/podmonitors/operator.go
```diff
@@ -0,0 +1,89 @@
...
+	// Ideally, this component has very little dynamic config, and won't have frequent updates.
+	var innerCtx context.Context
+	// cancel is the func we use to trigger a stop to all downstream processors we create
+	var cancel func()
+	defer func() {
+		if cancel != nil {
+			cancel()
+		}
+	}()
+	errChan := make(chan error)
```
rfratto, line 50: Unused, remove?

component/prometheus/kubernetes/podmonitors/operator.go
```diff
@@ -0,0 +1,89 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			if cancel != nil {
+				cancel()
+			}
```
rfratto, line 56: nit: You can remove this since it's guaranteed to be called in the return.

component/prometheus/kubernetes/podmonitors/podMonitor.go
```diff
@@ -0,0 +1,269 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
+	regexFilterRunning = relabel.MustNewRegexp("(Failed|Succeeded)")
+	regexTrue          = relabel.MustNewRegexp("true")
+	regexAnything      = relabel.MustNewRegexp("(.+)")
+)
+
+func (cg *configGenerator) generatePodMonitorConfig(m *v1.PodMonitor, ep v1.PodMetricsEndpoint, i int) (cfg *config.ScrapeConfig, err error) {
```
rfratto, line 29: nit: I always find it surprising and hard to navigate the code base when methods for a struct are split across multiple files, unless the file names indicate that it's a continuation of something. 

I would generally expect this file to be called `config_generator_podmonitor.go`. I also wouldn't use camelCase for any of file names, since no other file in this code base is named that way, and I like being consistent across the board.

component/prometheus/kubernetes/podmonitors/configGenerator.go
```diff
@@ -0,0 +1,176 @@
...
+	"github.com/prometheus/prometheus/model/relabel"
+)
+
+type configGenerator struct {
+	config *Arguments
+}
+
+// the k8s sd config is mostly dependent on our local config for accessing the kubernetes cluster.
+// if undefined it will default to an in-cluster config
+func (cg *configGenerator) generateK8SSDConfig(namespaceSelector v1.NamespaceSelector, namespace string, role promk8s.Role, attachMetadata *v1.AttachMetadata) *promk8s.SDConfig {
```
rfratto, line 22: I would really like to see unit tests for each of these methods instead of just having the one massive test for generatePodMonitorConfig.

component/prometheus/kubernetes/podmonitors/configGenerator.go
```diff
@@ -0,0 +1,176 @@
...
+	"regexp"
+
+	v1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	commonConfig "github.com/prometheus/common/config"
+	"github.com/prometheus/common/model"
+	promk8s "github.com/prometheus/prometheus/discovery/kubernetes"
+	"github.com/prometheus/prometheus/model/relabel"
+)
+
+type configGenerator struct {
```
rfratto, line 16: nit: I'm not sure why some methods are exported and some aren't. For unexported types like `configGenerator`, I prefer to use exported methods to indicate "the caller is expected to use these" and unexported methods to indicate "the caller is not expected to use these."

component/prometheus/kubernetes/podmonitors/operator.go
```diff
@@ -0,0 +1,89 @@
...
+		Name: "prometheus.kubernetes.podmonitors",
+		Args: Arguments{},
+
+		Build: func(opts component.Options, args component.Arguments) (component.Component, error) {
+			return New(opts, args)
+		},
+	})
+}
+
+type Component struct {
```
rfratto, line 22: IMO, this is a component that desperately needs DebugInfo like loki.source.podlogs. If you're not planning on adding DebugInfo in this PR, we need TODO comments and issues created after this is merged to track what's remaining.

component/prometheus/kubernetes/podmonitors/crdManager.go
```diff
@@ -0,0 +1,189 @@
...
+	if err != nil {
+		return errors.Wrap(err, "creating prometheus clientset")
+	}
+
+	c.cg = configGenerator{
+		config: c.config,
+	}
+
+	for _, namespace := range c.config.Namespaces {
+		factory := informers.NewSharedInformerFactoryWithOptions(promClientset,
```
rfratto, line 130: Ideally we use the same library that loki.source.podlogs and loki.source.kubernetes_events use for informers (sigs.k8s.io/controller-runtime). If you don't want to switch to that library in this PR, can you add a TODO?

component/prometheus/kubernetes/podmonitors/crdManager.go
```diff
@@ -0,0 +1,189 @@
...
+func (c *crdManager) onUpdatePodMonitor(oldObj, newObj interface{}) {
+	pm := oldObj.(*v1.PodMonitor)
+	c.clearConfigs("podMonitor", pm.Namespace, pm.Name)
+	c.addPodMonitor(newObj.(*v1.PodMonitor))
+}
+func (c *crdManager) onDeletePodMonitor(obj interface{}) {
+	pm := obj.(*v1.PodMonitor)
+	c.clearConfigs("podMonitor", pm.Namespace, pm.Name)
+	c.apply()
+}
```
rfratto, line 113: nit: These are declared before they're used. I like to suggest that methods in a file be declared in use order so the file is easier to read top-down, but as this is now I'm forced to jump around in my reading. 

I would generally expect that the `on*PodMonitor` methods are defined _after_ `run`.

component/prometheus/kubernetes/podmonitors/crdManager.go
```diff
@@ -0,0 +1,189 @@
...
+	c.scraper = scrape.NewManager(opts, c.logger, flowAppendable)
+	defer c.scraper.Stop()
+	targetSetsChan := make(chan map[string][]*targetgroup.Group)
+	go func() {
+		err := c.scraper.Run(targetSetsChan)
+		level.Info(c.logger).Log("msg", "scrape manager stopped")
+		if err != nil {
+			level.Error(c.logger).Log("msg", "scrape manager failed", "err", err)
+		}
+	}()
```
rfratto, line 179: There's some risk here that changing the component is going to inject a lot of staleness markers. Ideally there's one scraper shared throughout the lifetime of the component.

Otherwise, from my current understanding the Prometheus scraper, when you shut the scraper down, staleness markers get propagated through the system for every target, even if that target isn't going away, which will cause gaps in Grafana.

component/all/all.go
```diff
@@ -2,62 +2,63 @@
...
+	_ "github.com/grafana/agent/component/phlare/write"                      // Import phlare.write
+	_ "github.com/grafana/agent/component/prometheus/exporter/apache"        // Import prometheus.exporter.apache
+	_ "github.com/grafana/agent/component/prometheus/exporter/blackbox"      // Import prometheus.exporter.blackbox
+	_ "github.com/grafana/agent/component/prometheus/exporter/consul"        // Import prometheus.exporter.consul
+	_ "github.com/grafana/agent/component/prometheus/exporter/github"        // Import prometheus.exporter.github
+	_ "github.com/grafana/agent/component/prometheus/exporter/mysql"         // Import prometheus.exporter.mysql
+	_ "github.com/grafana/agent/component/prometheus/exporter/process"       // Import prometheus.exporter.process
+	_ "github.com/grafana/agent/component/prometheus/exporter/redis"         // Import prometheus.exporter.redis
+	_ "github.com/grafana/agent/component/prometheus/exporter/unix"          // Import prometheus.exporter.unix
+	_ "github.com/grafana/agent/component/prometheus/kubernetes/podmonitors" // Import prometheus.exporter.apache
```
rfratto, line 58: I just want to call out that it's OK for us to merge this PR to make incremental progress, but the component is _not_ currently ready for release, not until the documentation is completed and configGenerator is finished. 

I don't know how we're going to track this (maybe we just remember), but we'll need to remind ourselves to pull the component from the v0.33 release if it's not fully ready by then.

rfratto (commented):
Thanks @captncraig, here's another round of review feedback. 

Are you planning on updating the CHANGELOG?

component/prometheus/operator/podmonitors/podmonitor_test.go
```diff
@@ -0,0 +1,231 @@
+package podmonitors
```
jcreixell, line 1: shall we rename this file to `config_gen_pod_monitor_test.go`? (And config_gen_podMonitor.go should probably be named `config_gen_pod_monitor.go`?)

component/prometheus/operator/podmonitors/config_gen.go
```diff
@@ -0,0 +1,176 @@
+package podmonitors
+
+// SEE https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/prometheus/promcfg.go
```
jcreixell, line 3: in general, I would like to see more tests for this. For example, some of the code is adapted from prometheus (as referenced in the link), so there are tests we could already reuse https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/prometheus/promcfg_test.go

jcreixell (commented):
my general feedback is that there is still a good chunk of non-trivial config mappings that are not covered by tests, including edge cases and errors. My concern is that if we don't add them now, we will never do it, which will make refactoring scarier down the line.

We should also think about failure modes and whether we want to restart the component when it fails, or mark it as failed, etc.

There are still some unaddressed code comments and the doc is not complete.

Other than that I think we should be aiming to merge this sooner than later and continue iterating on it.

rgeyer (commented):
As a consumer. Docs look good, and make sense to me. Looking forward to using this 👍

component/common/config/kubernetes.go
```diff
@@ -0,0 +1,87 @@
+package config
```
rfratto, line 1: Should these types go into a new package at `component/common/config/kubernetes`? I don't think `ClientArguments` is very descriptive with the package it's been moved to.

component/prometheus/operator/podmonitors/config_gen/config_gen.go
```diff
@@ -0,0 +1,164 @@
+package config_gen
+
+// SEE https://github.com/prometheus-operator/prometheus-operator/blob/main/pkg/prometheus/promcfg.go
```
rfratto, line 3: Do we want to pin this comment to the specific SHA we copied it from so it's easier to tell if we get out of sync?

rfratto: Ditto in the other file.

component/prometheus/operator/podmonitors/config_gen/config_gen_podmonitor.go
```diff
@@ -0,0 +1,262 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
+	regexFilterRunning = relabel.MustNewRegexp("(Failed|Succeeded)")
+	regexTrue          = relabel.MustNewRegexp("true")
+	regexAnything      = relabel.MustNewRegexp("(.+)")
+)
+
+func (cg *ConfigGenerator) GeneratePodMonitorConfig(m *v1.PodMonitor, ep v1.PodMetricsEndpoint, i int) (cfg *config.ScrapeConfig, err error) {
```
rfratto, line 27: I think the implementation of this function can probably be refactored a little bit to be smaller, but I'm not going to block the PR on this since the PR is big enough.  

Maybe the next PR for another component can take the opportunity to refactor?

component/prometheus/operator/podmonitors/crdmanager.go
```diff
@@ -0,0 +1,280 @@
...
+	DiscoveryConfigs map[string]discovery.Configs
+	ScrapeConfigs    map[string]*config.ScrapeConfig
+	DebugInfo        map[string]*DiscoveredPodMonitor
+	DiscoveryManager *discovery.Manager
+	ScrapeManager    *scrape.Manager
+
+	Opts      component.Options
+	Logger    log.Logger
+	Args      *Arguments
+	ConfigGen config_gen.ConfigGenerator
```
rfratto, line 45: Why are all of these exported? That feels like a code smell to me, especially when some of them are suppose to be protected by a mutex.

component/prometheus/operator/podmonitors/crdmanager.go
```diff
@@ -0,0 +1,280 @@
...
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// Generous timeout period for configuring all informers
+const informerSyncTimeout = 10 * time.Second
+
+// CRDManager is all of the fields required to run the component.
+// on update, this entire thing will be recreated and restarted
+type CRDManager struct {
```
rfratto, line 34: Does this need to be exported? I'd like for us to be cognizant of what types/methods/functions we export for non-internal packages as we inch towards v1.

component/prometheus/operator/podmonitors/config_gen/config_gen.go
```diff
@@ -0,0 +1,164 @@
+package config_gen
```
rfratto, line 1: It's typically unidiomatic to use underscores in package names, unless it's representing  a named entity that already uses underscores (like an integration referencing an exporter like node_exporter or a component like kubernetes_events).

component/prometheus/operator/podmonitors/operator.go
```diff
@@ -0,0 +1,133 @@
...
+					level.Error(c.opts.Logger).Log("msg", "error running crd manager", "err", err)
+					errChan <- err
+				}
+			}()
+		}
+	}
+}
+
+// Update implements component.Component.
+func (c *Component) Update(args component.Arguments) error {
```
rfratto, line 94: In the future, it'd be nice to have Update initialize as much as it can so it can return errors early back to the caller. 

Can we add a TODO here for that?

component/prometheus/operator/podmonitors/crdmanager.go
```diff
@@ -0,0 +1,280 @@
...
+		cache, err := cache.New(config, opts)
+		if err != nil {
+			return err
+		}
+
+		informers := cache
+
+		go func() {
+			err := informers.Start(ctx)
+			if err != nil && ctx.Err() != nil {
```
rfratto, line 141: Is this correct? Did you mean `err != && ctx.Err() == nil`?

component/prometheus/operator/podmonitors/crdmanager.go
```diff
@@ -0,0 +1,280 @@
...
+	"github.com/grafana/agent/component/prometheus/operator/podmonitors/config_gen"
+	"github.com/prometheus/prometheus/config"
+	"github.com/prometheus/prometheus/discovery"
+	"github.com/prometheus/prometheus/discovery/targetgroup"
+	"github.com/prometheus/prometheus/scrape"
+	toolscache "k8s.io/client-go/tools/cache"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	v1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
```
rfratto, line 24: Can we give this a package alias that makes it clearer it's for Prometheus Operator? That was `v1.AddToScheme` later in this file is more clear about which scheme is being imported without having to scroll way back up.

component/prometheus/operator/podmonitors/crdmanager.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+
+	return nil
+}
+
+// configureInformers configures the informers for the CRDManager to watch for PodMonitors changes.
+func (c *CRDManager) configureInformers(ctx context.Context, informers cache.Informers) error {
+	objects := []client.Object{
+		&v1.PodMonitor{},
+	}
```
rfratto, line 160: Do we intend to add more objects to this slice or was this left by mistake after a refactor? 

If we don't intend to ever add more objects, it might be a good opportunity to clean this up and remove the loop.

(But IIUC, we might add more things to here eventually like Secrets or ConfigMaps).

component/prometheus/operator/podmonitors/crdmanager.go
```diff
@@ -0,0 +1,280 @@
...
+	defer c.Mut.Unlock()
+	prefix := fmt.Sprintf("podMonitor/%s/%s", ns, name)
+	for k := range c.DiscoveryConfigs {
+		if strings.HasPrefix(k, prefix) {
+			delete(c.DiscoveryConfigs, k)
+			delete(c.ScrapeConfigs, k)
+		}
+	}
+	delete(c.DebugInfo, prefix)
+}
```
rfratto, line 222: It's be nice to move this function to be defined _after_ the onXPodMonitor methods, since I wasn't clear what it did at first.

Typically, I recommend sorting functions in terms of usage order (i.e., if function A calls function B, then define function A first). This makes code much easier to read top-down, as you don't have to hop around the file.

component/prometheus/operator/podmonitors/crdmanager.go
```diff
@@ -0,0 +1,280 @@
...
+			level.Error(c.Logger).Log("name", pm.Name, "err", err, "msg", "error generating scrapeconfig from podmonitor")
+			break
+		}
+		c.Mut.Lock()
+		c.DiscoveryConfigs[pmc.JobName] = pmc.ServiceDiscoveryConfigs
+		c.ScrapeConfigs[pmc.JobName] = pmc
+		c.Mut.Unlock()
+	}
+	if err != nil {
+		c.addDebugInfo(pm.Namespace, pm.Name, err)
```
rfratto, line 255: This will only add the most recent error to the debug info. Should this be a multierror instead?

component/prometheus/operator/podmonitors/crdmanager.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+	prefix := fmt.Sprintf("podMonitor/%s/%s", ns, name)
+	c.DebugInfo[prefix] = debug
+}
+
+func (c *CRDManager) addPodMonitor(pm *v1.PodMonitor) {
+	var err error
+	for i, ep := range pm.Spec.PodMetricsEndpoints {
+		var pmc *config.ScrapeConfig
+		pmc, err = c.ConfigGen.GeneratePodMonitorConfig(pm, ep, i)
```
rfratto, line 244: It would be great in the future if this wrote events to the Kubernetes API so you could see the errors from the resource when doing `kubectl get <podmonitor>`. Can we add a TODO for that?

component/prometheus/operator/podmonitors/operator.go
```diff
@@ -0,0 +1,133 @@
...
+			innerCtx, cancel = context.WithCancel(ctx)
+			c.mut.Lock()
+			componentCfg := c.config
+			manager := NewCRDManager(c.opts, c.opts.Logger, componentCfg)
+			c.manager = manager
+			c.mut.Unlock()
+			go func() {
+				if err := manager.Run(innerCtx); err != nil {
+					level.Error(c.opts.Logger).Log("msg", "error running crd manager", "err", err)
+					errChan <- err
```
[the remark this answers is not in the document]
rfratto: If you're going to keep the channel, it should be buffered, otherwise this goroutine will leak if the manager fails at the same time the component shuts down.

component/prometheus/operator/podmonitors/operator.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func (c *Component) reportHealthy() {
+	c.healthMut.Lock()
+	defer c.healthMut.Unlock()
+	c.health = component.Health{
+		Health:     component.HealthTypeHealthy,
+		UpdateTime: time.Now(),
+	}
+}
```
rfratto, line 133: Can we combine these two into a single `reportHealth(err error)` function which is healthy when `err == nil`?

component/prometheus/operator/podmonitors/operator.go
```diff
@@ -0,0 +1,133 @@
...
+
+// DebugInfo returns debug information for this component.
+func (c *Component) DebugInfo() interface{} {
+	var info DebugInfo
+	for _, pm := range c.manager.DebugInfo {
+		info.DiscoveredPodMonitors = append(info.DiscoveredPodMonitors, pm)
+	}
+	info.Targets = scrape.BuildTargetStatuses(c.manager.ScrapeManager.TargetsActive())
+	return info
```
rfratto, line 113: There's a race condition here; `c.manager.DebugInfo` is protected by a mutex.

component/prometheus/operator/podmonitors/operator.go
rfratto, line 133: Can we combine these two into a single `reportHealth(err error)` function which is healthy when `err == nil`?

component/prometheus/operator/podmonitors/operator.go
rfratto, line 113: There's a race condition here; `c.manager.DebugInfo` is protected by a mutex.

rfratto, line 113: There's a race condition here; `c.manager.DebugInfo` is protected by a mutex.

component/prometheus/operator/podmonitors/operator.go
rfratto, line 133: Can we combine these two into a single `reportHealth(err error)` function which is healthy when `err == nil`?

docs/sources/flow/reference/components/prometheus.operator.podmonitors.md
```diff
@@ -0,0 +1,193 @@
...
+  stage: beta
+---
+
+# prometheus.operator.podmonitors
+
+{{< docs/shared lookup="flow/stability/beta.md" source="agent" >}}
+
+`prometheus.operator.podmonitors` discovers [PodMonitor](https://prometheus-operator.dev/docs/operator/api/#monitoring.coreos.com/v1.PodMonitor) resources in your kubernetes cluster and scrape the targets they reference. This component performs three main functions:
+
+1. Discover `PodMonitor` resources from your kubernetes cluster.
```
rfratto, line 13:
```suggestion
1. Discover `PodMonitor` resources from your Kubernetes cluster.
```

docs/sources/flow/reference/components/prometheus.operator.podmonitors.md
```diff
@@ -0,0 +1,193 @@
...
+2. Discover `Pods` in your cluster that match those `PodMonitors`.
+3. Scrape metrics from those `Pods`, and forward them to a receiver.
+
+The default configuration assumes the agent is running inside a kubernetes cluster, and uses the in-cluster config to access the kubernetes api. It can be run from outside the cluster by supplying connection info in the `client` block, but network level access to pods is required to scrape metrics from them.
```
rfratto, line 17:
```suggestion
The default configuration assumes the agent is running inside a Kubernetes cluster, and uses the in-cluster config to access the Kubernetes API. It can be run from outside the cluster by supplying connection info in the `client` block, but network level access to pods is required to scrape metrics from them.
```

docs/sources/flow/reference/components/prometheus.operator.podmonitors.md
```diff
@@ -0,0 +1,193 @@
...
+`forward_to` | `list(MetricsReceiver)` | List of receivers to send scraped metrics to. | | yes
+`namespaces` | `list(string)` | List of namespaces to search for PodMonitor resources. If not specified, all namespaces will be searched. || no
+
+## Blocks
+
+The following blocks are supported inside the definition of `prometheus.operator.podmonitors`:
+
+Hierarchy | Block | Description | Required
+--------- | ----- | ----------- | --------
+client | [client][] | Configures Kubernetes client used to find podmonitors. | no
```
rfratto, line 42:
```suggestion
client | [client][] | Configures Kubernetes client used to find PodMonitors. | no
```

docs/sources/flow/reference/components/prometheus.operator.podmonitors.md
```diff
@@ -0,0 +1,193 @@
...
+client > basic_auth | [basic_auth][] | Configure basic_auth for authenticating to the endpoint. | no
```
rfratto, line 43:
```suggestion
client > basic_auth | [basic_auth][] | Configure basic authentication to the Kubernetes API. | no
```

docs/sources/flow/reference/components/prometheus.operator.podmonitors.md
```diff
@@ -0,0 +1,193 @@
...
+client > authorization | [authorization][] | Configure generic authorization to the endpoint. | no
```
rfratto, line 44:
```suggestion
client > authorization | [authorization][] | Configure generic authorization to the Kubernetes API. | no
```

docs/sources/flow/reference/components/prometheus.operator.podmonitors.md
```diff
@@ -0,0 +1,193 @@
...
+client > oauth2 | [oauth2][] | Configure OAuth2 for authenticating to the endpoint. | no
```
rfratto, line 45:
```suggestion
client > oauth2 | [oauth2][] | Configure OAuth2 for authenticating to the Kubernetes API. | no
```

docs/sources/flow/reference/components/prometheus.operator.podmonitors.md
```diff
@@ -0,0 +1,193 @@
...
+client > oauth2 > tls_config | [tls_config][] | Configure TLS settings for connecting to the endpoint. | no
```
rfratto, line 46:
```suggestion
client > oauth2 > tls_config | [tls_config][] | Configure TLS settings for connecting to the Kubernetes API. | no
```

docs/sources/flow/reference/components/prometheus.operator.podmonitors.md
```diff
@@ -0,0 +1,193 @@
...
+client > tls_config | [tls_config][] | Configure TLS settings for connecting to the endpoint. | no
```
rfratto, line 47:
```suggestion
client > tls_config | [tls_config][] | Configure TLS settings for connecting to the Kubernetes API. | no
```

docs/sources/flow/reference/components/prometheus.operator.podmonitors.md
```diff
@@ -0,0 +1,193 @@
...
+
+The `selector` block describes a Kubernetes label selector for `PodMonitors`.
+
+The following arguments are supported:
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`match_labels` | `map(string)` | Label keys and values used to discover resources. | `{}` | no
+
+When the `match_labels` argument is empty, all resources will be matched.
```
rfratto, line 113:
```suggestion
When the `match_labels` argument is empty, all PodMonitor resources will be matched.
```

docs/sources/flow/reference/components/prometheus.operator.podmonitors.md
```diff
@@ -0,0 +1,193 @@
+---
```
rfratto, line 1: I don't think we have to put PodMonitor/PodMonitors or Pod/Pods in code backticks; that might make this render a little more nicely in the UI.

docs/sources/flow/reference/components/prometheus.operator.podmonitors.md
```diff
@@ -0,0 +1,193 @@
...
+[basic_auth]: #basic_auth-block
+[authorization]: #authorization-block
+[oauth2]: #oauth2-block
+[tls_config]: #tls_config-block
+[selector]: #selector-block
+[match_expression]: #match_expression-block
+
+### client block
+
+The `client` block configures the Kubernetes client used discover PodMonitors. If the `client` block isn't provided, the default in-cluster
```
rfratto, line 65:
```suggestion
The `client` block configures the Kubernetes client used to discover PodMonitors. If the `client` block isn't provided, the default in-cluster
```

docs/sources/flow/reference/components/prometheus.operator.podmonitors.md
```diff
@@ -0,0 +1,193 @@
...
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`match_labels` | `map(string)` | Label keys and values used to discover resources. | `{}` | no
+
+When the `match_labels` argument is empty, all resources will be matched.
+
+### match_expression block
+
+The `match_expression` block describes a Kubernetes label match expression for
```
rfratto, line 117:
```suggestion
The `match_expression` block describes a Kubernetes label matcher expression for
```

docs/sources/flow/reference/components/prometheus.operator.podmonitors.md
```diff
@@ -0,0 +1,193 @@
...
+
+The `operator` argument must be one of the following strings:
+
+* `"In"`
+* `"NotIn"`
+* `"Exists"`
+* `"DoesNotExist"`
+
+`selector` can make use of multiple
+`match_expression` inner blocks which are treated as AND clauses.
```
rfratto, line 136: I found this a little hard to understand, what about this:

```suggestion
If there are multiple `match_expressions` blocks inside of a `selector` block, they are combined together with AND clauses. 
```

docs/sources/flow/reference/components/prometheus.operator.podmonitors.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Exported fields
+
+`prometheus.operator.podmonitors` does not export any fields. It will forward all metrics it scrapes to the receiver configures with the `forward_to` argument.
```
rfratto, line 140: Use present tense:

```suggestion
`prometheus.operator.podmonitors` does not export any fields. It forwards all metrics it scrapes to the receiver configures with the `forward_to` argument.
```

docs/sources/flow/reference/components/prometheus.operator.podmonitors.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Component health
+
+`prometheus.operator.podmonitors` is reported as unhealthy if any of its critical components fails during initialization (discovery manager, informers).
```
rfratto, line 144: I'd like for this to not reference as much internal names, since those can change easily:

```suggestion
`prometheus.operator.podmonitors` is reported as unhealthy when given an invalid configuration, Prometheus components fail to initialize, or the connection to the Kubernetes API could not be established properly.
```

rfratto (approved):
Let's get this merged! I reviewed everything _except_: 

1. I did not validate that the config generator matches Prometheus Operator's behavior. 
2. I did not look at any of the test code. 

Most of my feedback here is around style suggestions, but I found a couple of bugs, too. 

GitHub may hide some of my comments since there's so many; you might have to expand the review to see all 37.

component/prometheus/operator/podmonitors/crdmanager.go
jcreixell: we had a look at this one; looks like it tries to parse all of the endpoints defined in the PodMonitor, and if any of them fails (either due to parse errors or unsupported features) we reject the entire PodMonitor and report the error via DebugInfo. This means we don't need multierror here, as we never have more than one error.

component/prometheus/operator/podmonitors/crdmanager.go
jcreixell: I am not sure about the intent here, I will let @captncraig chime in on this one

component/prometheus/operator/podmonitors/crdmanager.go
jcreixell: not sure about this one, this is also used in other components like pod logs and k8s events similarly. It does make sense to me, could you please elaborate?

component/prometheus/operator/podmonitors/operator.go
jcreixell: do you mean doing the CRDManager initialization in this function? do we even need the channel then? (you mention returning errors back early, here we always return `nil`, and I don't see any errors being returned outside of newly spawned goroutine, so it might not be necessary?)

rfratto: Yes, CRDManager.Run currently both initializes _and_ runs things, but initializing things could fail:

1. Initializing the rest client, scheme, and label selector could fail because of invalid config
2. Creating a cache for the informers might fail
3. Creating informers before running them could fail 

It would be nice to decouple the initialization logic from the running logic so we can report those errors early in the Update call before the component is running. 

(I don't think this should be done in this PR, definitely something for later)

component/prometheus/operator/podmonitors/crdmanager.go
rfratto: Thanks, I missed the break statement above 👍

component/prometheus/operator/podmonitors/config_gen/config_gen_podmonitor.go
jcreixell: sounds good, we will refactor this once we add service monitor support

component/prometheus/operator/podmonitors/operator.go
```diff
@@ -90,6 +90,8 @@ func (c *Component) Run(ctx context.Context) error {
 }
 
 // Update implements component.Component.
+// TODO(jcreixell): Initialize manager here so we can return errors back early to the caller.
+// See https://github.com/grafana/agent/pull/2688#discussion_r1152384425
```
rfratto, line 94: nit: please don't attach TODO comments as documentation comments on functions to prevent it showing up as part of the documentation, rather put TODO comments inside the body of the function (or type if adding a TODO to a type) instead

component/prometheus/operator/podmonitors/crdmanager.go
jcreixell: looking at the [implementation](https://github.com/kubernetes-sigs/controller-runtime/blob/7982358244ab6ce4e3a919c1999dd79a5a4476c4/pkg/cache/internal/informers.go#L195) (I believe this is the one), seems like `Start` blocks until the the context is cancelled. Once the context is cancelled, `ctx.Err()` will always return an error (related to the cancellation).

Although this condition will never be triggered given this implementation, we want to log an error when `Start` returns an error and the error is not related to the cancellation, so your reasoning makes sense to me.

I will change this now (only in this component for now)

captncraig: I think that code is carried over from PodLogs. Your interpretation is as good as mine.

PR #2697, opened 2023-01-09
flow: Allow relabeling components to expose configured rules

tpaschalis:

#### PR Description
This PR expands on the three relabeling components (`discovery.relabel`, `prometheus.relabel`, `loki.relabel`) to add a new Export field named Rules. The field value is a function that when it's called, it returns the currently-configured relabeling rules in place.

This will be useful for components that generate telemetry themselves instead of scraping it from a third-party source (such as the upcoming `loki.source.*` components), so that they can call into relabeling rules and process any 'internal' labels before discarding them.

To see this in action, here's how it would look as River configuration and component implementation for the upcoming `loki.source.syslog` component.

Wiring in the new field in another component: https://github.com/tpaschalis/agent/commit/a4d105bc68e237a49f6fa61eb8a2ccf633800cc1

```
loki.relabel "keep_internal" {
	rule {
		action      = "labelmap"
		regex       = "__syslog_(.*)"
		replacement = "syslog_${1}"
	}
	forward_to = []
}

loki.source.syslog "local" {
	listener {
		address  = "127.0.0.1:51893"
	}

	relabel_configs = loki.relabel.keep_internal.rules
	forward_to      = [loki.write.grafanacloud.receiver]
}

loki.write "grafanacloud" {
  endpoint {
...
  }
}
```

No issue filed.

#### Notes to the Reviewer
- I'm not sure if such a small change warrants the tests, let me know if you'd like to remove them.
- This approach allows the relabeling components to be used in two separate ways (both with their receiver/forward_to attributes and by refering to the rules themselves). Are we okay with that?
- The current implementation means that a `loki.source.syslog` component might be able to use the exports of a `prometheus.relabel` block. Should we define separate types for the three namespaces to limit accidental reuse?
- After marking the Rules alias with the capsule interface, here's how it shows up on the Flow UI page
image

#### TODO
- Expand the documentation on this usecase.

#### PR Checklist

- [X] CHANGELOG updated
- [X] Documentation added
- [X] Tests updated

### DIFF
--- a/component/common/relabel/relabel.go
+++ b/component/common/relabel/relabel.go
@@ -184,3 +184,6 @@ func ComponentToPromRelabelConfigs(rcs []*Config) []*relabel.Config {
 
 	return res
 }
+
+// Rules returns the relabel configs in use for a relabeling component.
+type Rules func() []*relabel.Config
--- a/component/common/relabel/relabel.go
+++ b/component/common/relabel/relabel.go
@@ -187,3 +187,7 @@ func ComponentToPromRelabelConfigs(rcs []*Config) []*relabel.Config {
 
 // Rules returns the relabel configs in use for a relabeling component.
 type Rules func() []*relabel.Config
+
+// RiverCapsule marks the alias defined above as a "capsule type" so that it
+// cannot be invoked by River code.
+func (r Rules) RiverCapsule() {}
--- a/component/common/relabel/relabel.go
+++ b/component/common/relabel/relabel.go
@@ -186,7 +186,7 @@ func ComponentToPromRelabelConfigs(rcs []*Config) []*relabel.Config {
 }
 
 // Rules returns the relabel configs in use for a relabeling component.
-type Rules func() []*relabel.Config
+type Rules func() []*Config
 
 // RiverCapsule marks the alias defined above as a "capsule type" so that it
 // cannot be invoked by River code.
--- a/component/discovery/relabel/relabel.go
+++ b/component/discovery/relabel/relabel.go
@@ -34,11 +34,13 @@ type Arguments struct {
 // Exports holds values which are exported by the discovery.relabel component.
 type Exports struct {
 	Output []discovery.Target `river:"output,attr"`
+	Rules  flow_relabel.Rules `river:"rules,attr"`
 }
 
 // Component implements the discovery.relabel component.
 type Component struct {
 	opts component.Options
+	rcs  []*relabel.Config
 }
 
 var _ component.Component = (*Component)(nil)
@@ -67,6 +69,7 @@ func (c *Component) Update(args component.Arguments) error {
 
 	targets := make([]discovery.Target, 0, len(newArgs.Targets))
 	relabelConfigs := flow_relabel.ComponentToPromRelabelConfigs(newArgs.RelabelConfigs)
+	c.rcs = relabelConfigs
 
 	for _, t := range newArgs.Targets {
 		lset := componentMapToPromLabels(t)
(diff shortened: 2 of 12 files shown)


component/loki/relabel/relabel.go
```diff
@@ -229,3 +230,10 @@ func (c *Component) process(e loki.Entry) model.LabelSet {
 	}
 	return relabeled
 }
+
+func (c *Component) getRules() []*relabel.Config {
+	c.mut.RLock()
+	defer c.mut.RUnlock()
+
+	return c.rcs
+}
```
rfratto, line 239: Do we want to make this a capsule so:

1. It shows up as being less obscure when viewing the UI?
2. It prevents users from calling the function from River? (Which they can do for any exported function)

tpaschalis: Yes, agreed. Do you think that in the future we might have the need for an interface that can not only get all rules but do other stuff too? (eg. fetch first N rules, fetch only `keep` rules etc?)

tpaschalis:
cc @mattdurham to kickstart the conversation.

I've also added an example of wiring this to another component as well as River config that utilizes it.

component/loki/relabel/relabel.go
tpaschalis: Marked it as a capsule type.

I'm thinking about whether we're okay to have the same type for all three namespaces.

rfratto: Oops, sorry, I typed that before I saw your reply :) Sharing the type probably makes sense.

mattdurham:
How does using the function work for the propagation of updates?

```
local.file "action" {
	path = "/action.txt"
}


loki.relabel "keep_internal" {
	rule {
		action      = local.file.action
		regex       = "__syslog_(.*)"
		replacement = "syslog_${1}"
	}
	forward_to = []
}

loki.source.syslog "local" {
	listener {
		address  = "127.0.0.1:51893"
	}
	relabel_configs = loki.relabel.keep_internal.rules
	forward_to      = [loki.write.grafanacloud.receiver]
}
```

tpaschalis:
image

The dependency graph for this configuration looks like this.

So any update to the file will update the loki.relabel component and the will also trigger an Update of the loki.source.syslog component. Then, it's up to that component to call the function and use the new values.

tpaschalis:
One more thing to decide is whether the function should expose the Flow version or the Prometheus version of the relabeling blocks; this just changes which side will call the `ComponentToPromRelabelConfigs` helper.

component/common/relabel/relabel.go
```diff
@@ -184,3 +184,10 @@ func ComponentToPromRelabelConfigs(rcs []*Config) []*relabel.Config {
 
 	return res
 }
+
+// Rules returns the relabel configs in use for a relabeling component.
+type Rules func() []*relabel.Config
```
rfratto, line 189: w/r/t returning the Flow vs Prometheus type:

The Flow type probably makes more sense, but if it's going to be too much work, this is fine for now.

component/discovery/relabel/relabel.go
```diff
@@ -67,6 +69,7 @@ func (c *Component) Update(args component.Arguments) error {
 
 	targets := make([]discovery.Target, 0, len(newArgs.Targets))
 	relabelConfigs := flow_relabel.ComponentToPromRelabelConfigs(newArgs.RelabelConfigs)
+	c.rcs = relabelConfigs
```
rfratto, line 72: It looks like this needs to be protected by a mutex to prevent race conditions on reading/writing `c.rcs` because `getRules` and `Update` can be called in parallel. It's probably the same case in the other components, too.

tpaschalis: We do this for other components, but hadn't used a mutex in `discovery.relabel` yet. Fixing this.

component/common/relabel/relabel.go
tpaschalis: I agree with this take.
The components weren't keeping track of the Flow variant so I've added this, but I believe it makes more sense for inter-component communication.

karengermond (approved):
LGTM!

Issue #2640, opened 2022-12-21
[bug] Flow stops scraping kubelet metrics correctly after node update


payparain:
When the Agent is started in Flow mode on a kubernetes cluster, initially all is fine and metrics are scraped from the kubelet and cAdvisor endpoints as expected using the below configuration:

```
discovery.kubernetes "k8s_node" {
	role = "node"
}

discovery.relabel "configure_kubelet" {
	targets = discovery.kubernetes.k8s_node.targets

	rule {
		replacement  = "kubernetes.default.svc:443"
		target_label = "__address__"
	}

	rule {
		regex         = "(.+)"
		replacement   = "/api/v1/nodes/$1/proxy/metrics"
		source_labels = ["__meta_kubernetes_node_name"]
		target_label  = "__metrics_path__"
	}
}

prometheus.scrape "kubelet_scrape" {
	targets      = discovery.relabel.configure_kubelet.output
	job_name     = "integrations/kubernetes/kubelet"
	scheme       = "https"
	honor_labels = true
	forward_to   = [prometheus.relabel.kubelet_filters.receiver]

	http_client_config {
		bearer_token_file = "/var/run/secrets/kubernetes.io/serviceaccount/token"

		tls_config {
			ca_file              = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
			insecure_skip_verify = false
			server_name          = "kubernetes"
		}
	}
}
```

This allows us to view metrics such as `kubelet_volume_stats_capacity_bytes` and friends on a Prometheus instance that the Agent writes to (either Grafana Cloud or otherwise).  However, during operation, when the Agent encounters multiple consecutive updates to the kubernetes discovery component for the nodes, metrics no longer are accurately scraped (either missing labels, or not scraped at all). This is a graph of what this looks like on Prometheus:
<img width="1495" alt="Screen Shot 2022-12-21 at 1 06 08 pm" src="https://user-images.githubusercontent.com/106503014/208811825-2226ab26-9c5d-4a92-ada8-230aff41bf95.png">

And this is what the Agent logs right at that cutoff point when it stops sending accurate metrics:

```
ts=2022-12-21T03:02:50.641065661Z level=debug msg="handling component with updated state" node_id=discovery.kubernetes.k8s_node
ts=2022-12-21T03:02:50.64109186Z level=info trace_id=9261bea3c087ef8e119b1e503b908949 msg="starting partial graph evaluation"
[...]

PR #2701, opened 2023-01-09
prometheus.relabel: clone labels before relabeling

rfratto:
This PR clones the label set before applying relabels. Not cloning lead to two problems:

1. It forces the computed ID of the incoming series to change (as its labels changed), which causes duplicate relabeling caching.
2. It can cause obscure bugs with relabel rules being applied, such as a `keep` action which doesn't work after modifying the original slice.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,10 @@ Main (unreleased)
 
 - Flow UI: Fix the issue with long string going out of bound in the component detail page. (@xiyu95)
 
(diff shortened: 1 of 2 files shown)


mattdurham (approved):
Looks good, I remember we had a conversation about this about whether to not copy and we felt it was safe. But must have been some edge cases.

rfratto:
In the past, I think I believed that relabel.Process _didn't_ modify the original slice, but it definitely does. I talked about this some in #2640: https://github.com/grafana/agent/issues/2640#issuecomment-1360832103

PR #2705, opened 2023-01-09
flow: Add labelallow/labeldrop/static_labels stages to `loki.process`

tpaschalis:

#### PR Description
This PR updates `loki.process` to port over some more Promtail processing stages along with their tests.

More specifically, the PR brings over some stages that are responsible for manipulating the label set of incoming log entries, namely `labelallow`, `labeldrop` and `static_labels`.

Updates #2515 

#### Notes to the Reviewer
- The `labelallow` terminology is a little weird given the opposite `labeldrop`. Should we rename it to `labelkeep`, or keep the Promtail name?
- Any better names for `static_labels`?

#### PR Checklist

- [ ] CHANGELOG updated (add a changelog entry when all stages are ported over?)
- [X] Documentation added
- [X] Tests updated


component/loki/process/internal/stages/static_labels.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+// Process implements Stage
+func (l *StaticLabelStage) Process(labels model.LabelSet, extracted map[string]interface{}, t *time.Time, entry *string) {
+
+	for lName, lSrc := range l.Config.Values {
+		if lSrc == nil || *lSrc == "" {
+			continue
+		}
+		s, err := getString(*lSrc)
```
tpaschalis, line 64: Promtail uses this helper to get string values from an interface{} type.

Here, our input is `map[string]*string` so we don't need that. Most likely, we don't need the pointer in the map value as well, as it was used to enable YAML keys with empty values.

docs/sources/flow/reference/components/loki.process.md
```diff
@@ -153,6 +159,68 @@ stage {
...
+### labelallow block
+
+The `labelallow` inner block configures a labelallow processing stage that
+filters the label set of an incoming log entry to only keep a subset of them.
+
+The following arguments are supported:
+
+Name        | Type          | Description                                   | Default        | Required
+----------- | --------------| --------------------------------------------- | -------------- | --------
+`values`    | `map(string)` | Configures a `labelallow` processing stage.   | `{}`           | no
```
mattdurham, line 171: Is this a map or list?

component/loki/process/internal/stages/pipeline.go
```diff
@@ -20,8 +20,11 @@ import (
 // We define these as pointers types so we can use reflection to check that
 // exactly one is set.
 type StageConfig struct {
-	JSONConfig   *JSONConfig   `river:"json,block,optional"`
-	LabelsConfig *LabelsConfig `river:"labels,block,optional"`
+	JSONConfig         *JSONConfig         `river:"json,block,optional"`
+	LabelsConfig       *LabelsConfig       `river:"labels,block,optional"`
+	LabelAllowConfig   *LabelAllowConfig   `river:"labelallow,block,optional"`
```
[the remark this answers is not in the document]
tpaschalis: Yes, I'll change them both.

mattdurham (approved):
Requested some small doc changes but LGTM

docs/sources/flow/reference/components/loki.process.md
tpaschalis: Done!

component/loki/process/internal/stages/label_drop.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"errors"
+	"time"
+
+	"github.com/prometheus/common/model"
+)
+
+// ErrEmptyLabelDropStageConfig error returned if config is empty.
+const ErrEmptyLabelDropStageConfig = "labeldrop stage config cannot be empty"
```
rfratto, line 15: I know we just ported this over, but it's weird to have an error variable be a string. This should really be an error type instead of wrapping it in `errors.New()` every time it's returned.

tpaschalis: True, should have picked this up. Fixing it.

docs/sources/flow/reference/components/loki.process.md
```diff
@@ -46,18 +46,24 @@ Name              | Type                 | Description
...
-stage > json   | [json][]   | Configures a JSON processing stage.  | no
-stage > labels | [labels][] | Configures a labels processing stage. | no
+Hierarchy        | Block      | Description | Required
+---------------- | ---------- | ----------- | --------
+`stage`          | [stage][]  | Processing stage to run. | no
+`stage > json`   | [json][]   | Configures a JSON processing stage.  | no
+`stage > labels` | [labels][] | Configures a labels processing stage. | no
+`stage > label_keep`   | [label_keep][]    | Configures a label_keep processing stage. | no
+`stage > label_drop`   | [label_drop][]    | Configures a label_drop processing stage. | no
+`stage > static_labels`| [static_labels][] | Configures a static_labels processing stage. | no
```
rfratto, line 56: We don't put the hierarchy in backticks:

```suggestion
stage          | [stage][]  | Processing stage to run. | no
stage > json   | [json][]   | Configures a JSON processing stage.  | no
stage > labels | [labels][] | Configures a labels processing stage. | no
stage > label_keep   | [label_keep][]    | Configures a `label_keep` processing stage. | no
stage > label_drop   | [label_drop][]    | Configures a `label_drop` processing stage. | no
stage > static_labels | [static_labels][] | Configures a `static_labels` processing stage. | no
```

component/loki/process/internal/stages/pipeline.go
```diff
@@ -20,8 +20,11 @@ import (
 // We define these as pointers types so we can use reflection to check that
 // exactly one is set.
 type StageConfig struct {
```
rfratto, line 22: Hm, I just noticed our component is using this directly in its arguments, which means that a publicly exposed type has a public field of an internal type, preventing external callers from being able to construct arguments by hand. 

I'm not sure if it's worth fixing right now, but maybe we should have a comment here to resolve in the future?

rfratto (approved):

component/loki/process/internal/stages/pipeline.go
tpaschalis: Yeah that's true. I'm adding a TODO comment, would the proper solution be to make the package non-internal once we've gained some confidence?

component/loki/process/internal/stages/pipeline.go
```diff
@@ -19,9 +19,17 @@ import (
 // StageConfig defines a single stage in a processing pipeline.
 // We define these as pointers types so we can use reflection to check that
 // exactly one is set.
+// TODO (@tpaschalis) The fact that this type belongs to an internal package
+// _and_ is part of the loki.process component Arguments, means that an
+// external caller will not be able to construct the component Arguments by
+// hand. This will be fixed once we've gained confidence in the ported
+// processing stages and the package is made non-internal.
```
rfratto, line 26: nit: please put TODO comments inside the struct definition as opposed to attached to the doc comments of StageConfig. Otherwise this will show up in pkg.go.dev which is always kind of weird to see from other packages IMO

PR #2714, opened 2023-01-10
Add `loki.source.windowsevent` component

### LINKED ISSUE #2507: `loki.source.windowsevents`: Listens to the Windows event logs API

mattdurham:
#### PR Description

This adds `loki.source.windowsevent` component to flow. 

Close #2507 

#### Notes to the Reviewer

There is an odd relation between xpath_query and eventlog_name, one of them is required depending on their values. So for required, I added `see below` with a note.

#### PR Checklist

- [X] CHANGELOG updated
- [X] Documentation added
- [X] Tests updated


component/loki/source/windowsevent/component.go
```diff
@@ -0,0 +1,37 @@
...
+var (
+	_ component.Component = (*FakeComponent)(nil)
+)
+
+// FakeComponent implements the loki.source.windowsevent component for non-windows environments.
+type FakeComponent struct {
+}
+
+func (f *FakeComponent) Run(ctx context.Context) error {
+	<-ctx.Done()
```
rfratto, line 31: Should this log something to tell the user that the component won't do anything?

component/loki/source/windowsevent/component.go
```diff
@@ -0,0 +1,37 @@
+//go:build !windows
```
rfratto, line 1: nit: Can we rename this file component_stub? I would generally see `component.go` and think it's where the main implementation is, so calling it out as a stub would be slightly less confusing.

component/loki/source/windowsevent/component_windows.go
```diff
@@ -0,0 +1,148 @@
...
+	receivers []loki.LogsReceiver
+}
+
+func (c *Component) Chan() chan<- api.Entry {
+	return c.handler
+}
+
+func (c *Component) Stop() {
+	// This is a noop.
+}
```
rfratto, line 48: I'm assuming these implement some kind of interface for `windows.New`? Personally, I'd like us to create a new (unexported) struct for the business logic so implementing the interface for windows.New doesn't become a part of the public API for the component.

component/loki/source/windowsevent/arguments.go
```diff
@@ -0,0 +1,46 @@
...
+// UnmarshalRiver implements river.Unmarshaler.
+func (r *Arguments) UnmarshalRiver(f func(v interface{}) error) error {
+	*r = defaultArgs()
+
+	type arguments Arguments
+	if err := f((*arguments)(r)); err != nil {
+		return err
+	}
+
+	return nil
```
rfratto, line 45: Should we be doing any validations here? (Such as making sure `eventlog_name` is provided if `xpath_query` uses the short form)

rfratto (commented):
🎉 A nice and simple one :)

component/loki/source/windowsevent/component_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+func TestEventLogger(t *testing.T) {
+	var loggerName = "agent_test"
+	//Setup Windows Event log with the log source name and logging levels
+	_ = eventlog.InstallAsEventCreate(loggerName, eventlog.Info|eventlog.Warning|eventlog.Error)
+	wlog, err := eventlog.Open(loggerName)
+	require.NoError(t, err)
+	l, err := logging.New(os.Stdout, logging.DefaultOptions)
+	require.NoError(t, err)
+	dataPath, err := os.MkdirTemp("", "loki.source.windowsevent")
```
tpaschalis, line 27: Small nitpick, the component doesn't need a DataPath, right?

component/loki/source/windowsevent/component_windows.go
```diff
@@ -0,0 +1,148 @@
...
+		case <-ctx.Done():
+			return nil
+		case entry := <-c.handler:
+			lokiEntry := loki.Entry{
+				Labels: entry.Labels,
+				Entry:  entry.Entry,
+			}
+			for _, receiver := range c.receivers {
+				receiver <- lokiEntry
+			}
```
tpaschalis, line 87: Do we need to take an RLock/RUnlock for c.receivers? Also, can we send the entry directly?

component/loki/source/windowsevent/component_windows.go
```diff
@@ -0,0 +1,148 @@
...
+// Update implements component.Component.
+func (c *Component) Update(args component.Arguments) error {
+	newArgs := args.(Arguments)
+
+	c.mut.Lock()
+	defer c.mut.Unlock()
+
+	// If no bookmark specified create one in the datapath.
+	if newArgs.BookmarkPath == "" {
+		newArgs.BookmarkPath = path.Join(c.opts.DataPath, "bookmark.xml")
```
tpaschalis, line 102: The component's DataPath is not guaranteed to exist. I think we might have to create in New and cleanup after Run?

https://github.com/grafana/agent/blob/67627a3c30bf76dec414a1e78b014c0ee49cce27/component/registry.go#L42-L44

component/loki/source/windowsevent/component_windows.go
```diff
@@ -0,0 +1,148 @@
...
+		if err != nil {
+			return err
+		}
+		_, err = os.Create(newArgs.BookmarkPath)
+		if err != nil {
+			return err
+		}
+	}
+
+	winTarget, err := windows.New(c.opts.Logger, c, nil, convertConfig(newArgs))
```
tpaschalis, line 118: Does the Promtail code create any internal labels, similar to our discussion around `loki.source.syslog`?

Since we've merged #2697, we can have an optional Argument for this, and if it's not nil then pass it through `ComponentToPromRelabelConfigs` and add it as the argument here ^^

tpaschalis (approved):
Couple of small points, but nice one!

docs/sources/flow/reference/components/loki.source.windowsevent.md
```diff
@@ -0,0 +1,70 @@
...
+The component starts a new reader and fans out
+log entries to the list of receivers passed in `forward_to`.
+
+`loki.source.windowsevent` supports the following arguments:
+
+Name         | Type                 | Description                                                                    | Default                    | Required
+------------ |----------------------|--------------------------------------------------------------------------------|----------------------------| --------
+`locale`    | `number`             | Locale ID for event rendering. 0 default is Windows Locale.                    | `0` | no
+`eventlog_name`    | `string`             | Event log to read from.                                                        |                            | see below
+`xpath_query`    | `string`             | Event log to read from.                                                        | `"*"`                          | see below
```
karengermond, line 34:
```suggestion
`eventlog_name`    | `string`             | Event log to read from.                                                        |                            | See below.
`xpath_query`    | `string`             | Event log to read from.                                                        | `"*"`                          | See below.
```

component/loki/source/windowsevent/component_windows.go
mattdurham: Good call on lock, and we cant because it's a slightly different struct with almost the exact same fields.

docs/sources/flow/reference/components/loki.source.windowsevent.md
```diff
@@ -0,0 +1,70 @@
...
+`exclude_event_data`    | `bool`               | Exclude event data.                                                            | `false`                      | no
+`exclude_user_data`    | `bool`               | Exclude user data.                                                             | `false`                      | no
+`user_incoming_timestamp`    | `bool`               | When false will assign the current timestamp to the log when it was processed. | `false`                      | no
+`forward_to` | `list(LogsReceiver)` | List of receivers to send log entries to.                                      |                            | yes
+
+
+> **NOTE**: eventlog_name is required if xpath_query does not specify the event log.
+> The xpath_query can be defined in [short or xml form](https://docs.microsoft.com/en-us/windows/win32/wes/consuming-events).
+> When using the XML form you can specify the event_log in the xpath_query. 
+> If using short form, eventlog_name must be defined.
```
karengermond, line 46: I guess you should use code style for `eventlog_name` and `xpath_query`.   I think code within notes doesn't look as bad in the new website design.  Here's a note with code: https://grafana.com/docs/grafana-cloud/kubernetes-monitoring/navigate-k8s-monitoring/#pods.

component/loki/source/windowsevent/component_windows.go
mattdurham: This is handled componet_windows.go:110
```
_, err := os.Stat(newArgs.BookmarkPath) if os.IsNotExist(err) { 	err := os.MkdirAll(path.Dir(newArgs.BookmarkPath), 644) 	if err != nil { 		return err 	} 	_, err = os.Create(newArgs.BookmarkPath) 	if err != nil { 		return err 	} }
```

mattdurham: We shouldnt clean up so we dont send duplicate events.

karengermond (approved):
LGTM -- a few comments

component/loki/source/windowsevent/component_windows.go
tpaschalis: Whoops, sorry I missed this 🤦

component/loki/source/windowsevent/component_test.go
```diff
@@ -0,0 +1,73 @@
...
+	})
+	require.NoError(t, err)
+	ctx := context.Background()
+	ctx, cancelFunc := context.WithTimeout(ctx, 10*time.Second)
+	found := false
+	go c.Run(ctx)
+	tm := time.Now().Format(time.RFC3339Nano)
+	err = wlog.Info(2, tm)
+	require.NoError(t, err)
+	select {
```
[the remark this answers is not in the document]
mattdurham: The timeout will in the test file will trigger if the read fails.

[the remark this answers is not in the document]
mattdurham: I tested it by changing the poll interval to greater than the timeout.

component/loki/source/windowsevent/component_windows.go
mattdurham: Just checked and it does not create and labels prefixed with `__`, it adds channel (event log source) and the computer name as regular labels.

component/loki/source/windowsevent/component_test.go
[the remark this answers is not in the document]
tpaschalis: Oh, so you mean the context timeout? Makes sense then ^^

component/loki/source/windowsevent/arguments.go
mattdurham: I tried to do this a bit. 

The current implementation doesn't check and we could try to parse the field as XML but even with XML you may need the eventlog_name, so it felt more effort than it was worth. IMO we can ship it without and  add it later.

PR #2721, opened 2023-01-11
Flow: drop unqueued logs after 5 seconds when stopping tailer

### LINKED ISSUE #2716: [bug] Flow fails to display `loki.source.file` when port-forwarding on kubernetes
Using v0.30, when trying to click on a `loki.source.file` block via the UI on the graph, the page receives an empty response and no error/debug logs are emitted by the Agent. The agent _is_ reading log data; I can see it attempting to write it to our remote Loki endpoint, but the visual block for this component isn't viewable. 

Screen Shot 2023-01-11 at 10 20 29 am

rfratto:
Fix an issue where being unable to send logs to `loki.write` due to the client being permanently backlogged would deadlock the Flow controller.

The `loki.write` client may be permanently backlogged when:

* Limits are reached when sending logs to Loki, leading to endless request retries.
* Loki has an extended outage.

When an EntryHandler is stopped, it will wait for 5 seconds before forcibly stopping the goroutine which queues log entries. If this timeout is reached, any unqueued log entries are permanently lost, as the positions file will likely be updated past the point where the entry was read.

While losing logs is not ideal, it's unacceptable for any Flow component to be able to block the controller. This is a short-term solution to allow the Flow controller to continue working properly. A long term solution would be to use a Write-Ahead Log (WAL) for log entries. See grafana/loki#7993.

Fixes #2716.
Related to grafana/loki#2361.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,11 @@ Main (unreleased)
   metrics, which could lead to relabel rules applying incorrectly on subsequent
   relabels. (@rfratto)
 
+- Flow: `loki.source.file` will no longer deadlock other components if log
+  lines are unable to be sent to Loki. `loki.source.file` will wait for 5
+  seconds per file to finish flushing read logs to the client, after which it
+  will drop them, resulting in losing logs. (@rfratto)
+
 ### Other changes
 
 - Use Go 1.19.4 for builds. (@erikbaranowski)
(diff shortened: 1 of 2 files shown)


mattdurham (approved):
Fantastic and much cleaner than I thought it would be, LGTM

tpaschalis (approved):

erikbaranowski (approved):

PR #2732, opened 2023-01-12
(helm) Support exposing additional ports for the Grafana Agent

payparain:
#### PR Description
This PR adds support for specifying extra port maps to the Grafana Agent, to support exposing different ports for specific components (such as [`otelcol.receiver.jaeger`](https://grafana.com/docs/agent/latest/flow/reference/components/otelcol.receiver.jaeger/#thrift_http-block), which defaults to listening on `14268` for http-thrift traces). By supplying the values:

```
agent:
    extraPorts:
        - name: jaeger-thrift
           port: 14268
           targetPort: 14268
           protocol: TCP
```

A new port will be exposed in both the headless service and the container itself with the provided arguments, allowing one to use the Kube DNS to send otelcol data to the Agent.

N/A

#### Notes to the Reviewer
This approach is based on the way [`kube-prom-stack`](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/values.yaml#L359) adds additional ports for it's various components, but modified slightly due to the differences in how ports are specified on a container vs a service.
#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/operations/helm/charts/grafana-agent/CHANGELOG.md
+++ b/operations/helm/charts/grafana-agent/CHANGELOG.md
@@ -9,6 +9,10 @@ internal API changes are not present.
 
 Unreleased
 ----------
+
+0.2.0 (2023-1-12)
+----------
+
 ### Features
 
 - Introduce supporting extra ports on the Grafana Agent created by Helm Chart.
--- a/operations/helm/charts/grafana-agent/CHANGELOG.md
+++ b/operations/helm/charts/grafana-agent/CHANGELOG.md
@@ -9,6 +9,9 @@ internal API changes are not present.
 
 Unreleased
 ----------
+### Features
+
+- Introduce supporting extra ports on the Grafana Agent created by Helm Chart.
 
 0.1.0 (2023-01-11)
 ------------------
--- a/operations/helm/charts/grafana-agent/Chart.yaml
+++ b/operations/helm/charts/grafana-agent/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v2
 name: grafana-agent
 description: "Grafana Agent"
 type: application
-version: 0.1.0
+version: 0.1.1
 appVersion: "v0.30.2"
--- a/operations/helm/charts/grafana-agent/Chart.yaml
+++ b/operations/helm/charts/grafana-agent/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v2
 name: grafana-agent
 description: "Grafana Agent"
 type: application
-version: 0.1.1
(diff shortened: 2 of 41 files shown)


payparain:
I hope I understood the tests right; I just put the rendered output into the relevant folder and added values like the others. Apologies if I got this wrong!

operations/helm/charts/grafana-agent/templates/containers/_agent.yaml
```diff
@@ -25,6 +25,11 @@
   ports:
     - containerPort: {{ .Values.agent.listenPort }}
       name: http-metrics
+    {{- range $portMap := .Values.agent.extraPorts }}
+    - containerPort: {{ $portMap.port }}
```
rfratto, line 29: Should this be targetPort instead of port? (since port refers to the service port and targetPort refers to the port number from the container)

operations/helm/charts/grafana-agent/Chart.yaml
```diff
@@ -2,5 +2,5 @@ apiVersion: v2
 name: grafana-agent
 description: "Grafana Agent"
 type: application
-version: 0.1.0
+version: 0.1.1
```
rfratto, line 5: Can you bump this up to 0.2.0 instead since it's technically not fixing a bug?

operations/helm/charts/grafana-agent/CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ internal API changes are not present.
 
 Unreleased
```
rfratto, line 10: Looks like we'll have to release this immediately, can you add a section for 0.2.0?

We'll probably disable requiring version bumps for every change after this gets merged.

rfratto (approved):
Thank you! I'll merge once the feedback is addressed :)

operations/helm/charts/grafana-agent/templates/containers/_agent.yaml
payparain: Totally right! I'll change this now.

operations/helm/charts/grafana-agent/Chart.yaml
payparain: Yep, will do; wasn't sure what version to make it when bumping.

operations/helm/charts/grafana-agent/CHANGELOG.md
payparain: Will do

PR #2733, opened 2023-01-12
Refactor template for Operator API docs to use markdown

marctc:
#### PR Description

The previous implementation of the templates that generated automatic API documentation for the Agent Operator had a lot of issues to render tables properly and other formatting issues. This PR replaces the usage of HTML by markdown, making it more suitable for Hugo.

#### Notes to the Reviewer

This how is generated now:

I woud like to remove the links on the right side as the anchor is not properly linked and I'm not sure how to fix it. Any ideas how to remove them?

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated


jdbaldry:
I guess the right hand side bar isn't working because it doesn't understand the `<a>` tags and how it affects anchor resolution.

I'm not immediately sure of the difficulty of fixing this in the website. What is the purpose of the `<a>` tags? Could we consider removing them and falling back to the default anchor id behavior?

marctc:
> What is the purpose of the `<a>` tags? Could we consider removing them and falling back to the default anchor id behavior?

If I don't add those `<a>`, the links inside the page won't work (they use something like `#monitoring.grafana.com/v1alpha1.LogsDeployment` for the anchors) but the links on the side bar will work. I tried to replace those links by the name of the resource (`#LogsDeployment`) but links in the side bar are lowercase. I also tried to lowercase the links but with no success. So at this point I'm kinda stuck :shrug:

jdbaldry:
That makes sense, thanks for explaining :)

I think it would be nice for the website to support the links as they are now but I'm not sure of the complexity and wouldn't want to stall your PR.

If you wanted to instead use the GitHub generated ids, the algorithm for sanitizing headings is https://github.com/gohugoio/hugo/blob/3d5dbdcb1a11b059fc2f93ed6fadb9009bf72673/markup/goldmark/autoid.go#L43-L84

captncraig (approved):
Resolve conflict, and lgtm.

jdbaldry:
I updated the `doc-validator` tool to use the Hugo heading->id logic in https://github.com/grafana/technical-documentation/commit/ea5280812290669fcf189d1f3a5092ddbd7b270b

PR #2738, opened 2023-01-12
Add support for GELF source.

### LINKED ISSUE #2509: `loki.source.gelf`: Reads GELF messages over UDP

mattdurham:
#### PR Description

This adds support for reading Graylog logs (GELF).

Closes #2509 

#### Notes to the Reviewer

Some controversial thoughts, I added job to the output since by default it comes with no labels and loki rejects. Job is the ID of the component.

There are also a few `__` labels. @tpaschalis I know syslog uses separate rules set but would like to revisit that now since I have used it.

#### PR Checklist

- [X] CHANGELOG updated
- [X] Documentation added
- [X] Tests updated


docs/sources/flow/reference/components/loki.source.gelf.md
```diff
@@ -0,0 +1,67 @@
...
+`loki.source.gelf` supports the following arguments:
+
+Name         | Type                 | Description                                                                    | Default                    | Required
+------------ |----------------------|--------------------------------------------------------------------------------|----------------------------| --------
+`listen_address`    | `string`             | UDP address and port to listen for Graylog messages.                    | `0.0.0.0:12201` | no
+`use_incoming_timestamp`    | `bool`             | When false, assigns the current timestamp to the log when it was processed | `false`                            | no
+
+
+> **NOTE**: GELF logs can be sent uncompressed or compressed with GZIP or ZLIB. 
+> A `job` label will be added with the full name of the component `loki.source.gelf.LABEL`.
```
karengermond, line 36:
```suggestion
> **NOTE**: GELF logs can be sent uncompressed or compressed with GZIP or ZLIB. 
> A `job` label is added with the full name of the component `loki.source.gelf.LABEL`. 
```

docs/sources/flow/reference/components/loki.source.gelf.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Component health
+
+`loki.source.gelf` is only reported as unhealthy if given an invalid
+configuration.
+
+## Debug Metrics
+
+* `gelf_target_entries_total` (counter): Total number of successful entries sent to the gelf target.
+* `gelf_target_parsing_errors_total` (counter): Total number of parsing errors while receiving gelf messages.
```
karengermond, line 53:
```suggestion
* `gelf_target_entries_total` (counter): Total number of successful entries sent to the GELF target.
* `gelf_target_parsing_errors_total` (counter): Total number of parsing errors while receiving GELF messages.
```

karengermond (approved):
LGTM; minor edits

component/loki/source/gelf/component.go
```diff
@@ -0,0 +1,141 @@
...
+	"context"
+	"sync"
+
+	"github.com/prometheus/common/model"
+
+	"github.com/grafana/agent/component"
+	"github.com/grafana/agent/component/common/loki"
+	"github.com/grafana/loki/clients/pkg/promtail/api"
+	"github.com/grafana/loki/clients/pkg/promtail/scrapeconfig"
+	"github.com/grafana/loki/clients/pkg/promtail/targets/gelf"
```
tpaschalis, line 13: In `loki.source.syslog` (and some other Loki components) we went with a slightly different approach; 
a) copying the targets package as an internal package for the component, and
b) making them both use our own copy of the `github.com/grafana/loki/clients/pkg/promtail/api` package in `github.com/grafana/agent/component/common/loki`.

It shouldn't make the component itself any harder to work with, just will slightly inflate the PR size but allow us to have a straightforward way to deprecate Promtail

component/loki/source/gelf/component.go
```diff
@@ -0,0 +1,141 @@
...
+	c.target = t
+	return nil
+}
+
+// Arguments are the arguments for the component.
+type Arguments struct {
+	// ListenAddress only supports UDP.
+	ListenAddress        string              `river:"listen_address,attr,optional"`
+	UseIncomingTimestamp bool                `river:"use_incoming_timestamp,attr,optional"`
+	Receivers            []loki.LogsReceiver `river:"forward_to,attr"`
```
tpaschalis, line 87: This is a question for other `loki.source.*` components as well, so looking for a temperature check.

Do we allow each component to only define a single target? Or should we do what scrape_configs do and wrap these into a block so the user can define as many as they'd want under a single component?

After doing the block thing with the `syslog` target, I think that on one hand, I like the simplicity that this brings, plus having them as separate graph nodes, but not sure what's best from a UX perspective.

cc @mattdurham @rfratto

mattdurham: I like 1 per component; there *might* be exceptions, but this makes it cleaner and the config much cleaner.

rfratto: What does "target" mean in this case? Server?

tpaschalis: Yes, that's the idea. After catching up with Matt and arrived at the same conclusion.

We'll have sources like this or the Heroku only allow launching a single listener, and others like GCPLog that might need to read from multiple subscriptions or Syslog that allows both TCP and UDP allow defining multiple listeners.

> for the above example: at most one TCP listener AND at most one UDP listener, up to two total listeners

This is _not_ implemented currently in the Syslog receiver, but we can think about this restriction. I'm not sure if in the Syslog world it's valid to have multiple application log on different ports, but I'll check it out.

mattdurham:
Fixed a bug in calling deepequal that caused all comparisons to trigger.

component/loki/source/gelf/component.go
```diff
@@ -0,0 +1,147 @@
...
+	c.target = t
+	return nil
+}
+
+// Arguments are the arguments for the component.
+type Arguments struct {
+	// ListenAddress only supports UDP.
+	ListenAddress        string                 `river:"listen_address,attr,optional"`
+	UseIncomingTimestamp bool                   `river:"use_incoming_timestamp,attr,optional"`
+	RelabelConfigs       []*flow_relabel.Config `river:"rule,block,optional"`
```
tpaschalis, line 92: Are we going with the inner blocks after all instead of using the struct exported by relabeling components like loki.source.syslog does?

component/loki/source/gelf/component.go
```diff
@@ -0,0 +1,147 @@
...
+	Receivers            []loki.LogsReceiver    `river:"forward_to,attr"`
+}
+
+func defaultArgs() Arguments {
+	return Arguments{
+		ListenAddress:        "0.0.0.0:12201",
```
tpaschalis, line 98: 👍 I was debating whether to default to 0.0.0.0 and a non-random port when building loki.source.gcplog. Might follow suit here.

component/loki/source/gelf/component.go
```diff
@@ -0,0 +1,147 @@
...
+		return nil, err
+	}
+	return c, nil
+}
+
+type handler struct {
+	c chan api.Entry
+}
+
+func (h *handler) Chan() chan<- api.Entry {
```
tpaschalis, line 142: I think we can use our own`loki.Entry` and `loki.NewEntryHandler(c.handler, func() {})` directly here instead of the Promtail package. WDYT?

component/loki/source/gelf/component.go
```diff
@@ -0,0 +1,147 @@
+package gelf
```
tpaschalis, line 1: Super silly nitpick, but should we rename these files to `gelf.go` and `gelf_test.go` so it's easier to navigate to like other components do?

pkg/flow/internal/controller/component.go
```diff
@@ -232,40 +232,40 @@ func (cn *ComponentNode) evaluate(scope *vm.Scope) error {
...
 		}
 		cn.managed = managed
-		cn.args = argsCopy
+		cn.args = argsCopyValue
 
 		return nil
 	}
 
-	if reflect.DeepEqual(cn.args, args) {
+	if reflect.DeepEqual(cn.args, argsCopyValue) {
```
tpaschalis, line 256: Is this bug you mentioned? What do you mean about comparisons getting triggered?

component/loki/source/gelf/internal/target/metrics.go
```diff
@@ -0,0 +1,42 @@
...
+
+// NewMetrics creates a new set of gelf metrics. If reg is non-nil, the
+// metrics will be registered.
+func NewMetrics(reg prometheus.Registerer) *Metrics {
+	var m Metrics
+	m.reg = reg
+
+	m.gelfEntries = prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace: "promtail",
+		Name:      "gelf_target_entries_total",
```
tpaschalis, line 25: Let's remove the `promtail` namespace and have the metric names be `loki_source_gelf_*`. WDYT?

pkg/flow/internal/controller/component.go
mattdurham: Yes, previously it was comparing a pointer to a value type.

ptodev (approved):

component/loki/source/gelf/internal/target/gelftarget.go
```diff
@@ -0,0 +1,202 @@
...
+
+	"github.com/grafana/loki/clients/pkg/promtail/api"
+	"github.com/grafana/loki/clients/pkg/promtail/scrapeconfig"
+	"github.com/grafana/loki/clients/pkg/promtail/targets/target"
+
+	"github.com/grafana/loki/pkg/logproto"
+)
+
+// SeverityLevels maps severity levels to severity string levels.
+var SeverityLevels = map[int32]string{
```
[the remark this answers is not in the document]
mattdurham: It isn't possible to make a map const. It's in an internal package so it's mostly hidden.

pkg/flow/internal/controller/component.go
tpaschalis: Sorry for the trouble, but can we check again? Looks like that both types were the same i.e. the component.Arguments interface.

tpaschalis (approved):
LGTM. I am worried about how the bug fix impacts how we do relabeling rules in other components and keeping a convention around it, but it shouldn't block the loki source. Let me know if you think it should be on a different commit though.

PR #2747, opened 2023-01-13
production/tanka: stop explicitly setting command

rfratto:
Explicitly setting the command is not neccesary, but also currently won't work in the Jsonnet because the prefix name change (#2677) for the binaries hasn't been released yet, and the Jsonnet uses a stable release.

Removing an explicit setting of the command allows containers to use the Dockerfile default.


tpaschalis (approved):
LGTM

PR #2755, opened 2023-01-16
operator: fix enable_http2 flag that defaults to true.

captncraig:
The enable_http2 flag defaults to true in prometheus, so needs to be explicitly set to `false`. `optionals.bool`, was detecting false, and emitting null, but it should not. ADded a param to optionals.bool to supply the default value for cases where the default is true.


mattdurham (approved):
Lgtm

PR #2757, opened 2023-01-16
(helm) Update configmap.yaml to use correct value for custom configuration

### LINKED ISSUE #2756: (helm) Helm chart configmap has mismatched variables
In the latest version of the helm chart, to pass in new config the following checks are performed:

```
 {{- if .Values.agent.configMap.content }}
  config.river: {{ toYaml .Values.config | indent 2 }}
  {{- else }}
  ```
  
  At a glance this seems correct, but on closer inspection this is referring to two different things; a field called `config` in the root of the values file, which is what it accepts as the content to load in for the agent, and the field `agent.configMap.content` which is whats checked to determine if the contents of `config` should be provided to the configmap.
  
  This is confusing. In order to pass a custom `.river` file to the chart, one needs to both supply `.Values.config` _and_ put something that isn't applied in `agent.configMap.content`. I'm not sure the best approach to reconcile this, but I personally think we should just use the one `agent.configMap.content` variable and remove `config` from the values, ie
  
  ```
  {{- if .Values.agent.configMap.content }}
  config.river: {{ toYaml .Values.agent.configMap.content | indent 2 }}
  {{- else }}
  ```

### PR DESCRIPTION
**Pull request by payparain**
#### PR Description
Currently, the Agent helm chart checks a configMap variable to decide if it should populate a configmap with the contents of a separate variable, `config`. This is confusing, and makes it impossible to add custom config without also populating `agent.configMap.content` with something that never gets used, but is not `""`. This PR aligns the check with the content being populated.

#2756 

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [ ] Tests updated

### CODE REVIEW

**Comment by tpaschalis** [comment]
LGTM from a playing around locally with `helm template` locally, but let's wait to hear from @rfratto.

My other thought is: #2735 made it so we don't always have to version bump, but can batch many changes into one release. Do you think it makes sense to keep the version as-is so we can bundle it together with a fix to #2762?

File: operations/helm/charts/grafana-agent/CHANGELOG.md [hunk 1]
```diff
@@ -10,6 +10,12 @@ internal API changes are not present.
 Unreleased
 ----------
 
+0.2.2 (2023-01-16)
+------------------
```
rfratto, line 14:
```suggestion
0.2.2 (2023-01-17)
------------------
```

rfratto (approved):
Thank you for fixing this! It would be nice to bundle changes, but I unfortunately have limited time for reviews today, so I'll bump the other PR to 0.2.3 and release both.

PR #2758, opened 2023-01-16
flow: Add support for filtering stages to loki.process component

tpaschalis:

#### PR Description
This PR updates loki.process to port the two 'filtering' stages available in Promtail, namely `match` and `drop`.

The first one can conditionally either drop, or  run a nested set of processing stages on entries that match a LogQL selector.

The second one can drop entries based on some predefined criteria, which are treated as AND clauses.
 
Updates #2515 

#### Notes to the Reviewer
We're still trying to port these over in smaller batches, so we can review the code and documentation being brought in, as well as make any obvious improvements work.

For this PR, the main change from the upstream Promtail code is not using pointers to denote optional values but letting River handle those.

Finally, the `DropConfig` struct contains both an Expression string _and_ the Regex it's used to construct. We can keep only one if we do something similar to what we're doing on `flow_relabel.Config`, but didn't know if it was worth the hassle.

#### PR Checklist

- [ ] CHANGELOG updated (will add a changelog entry when all stages are ported over)
- [X] Documentation added
- [X] Tests updated

### DIFF
--- a/docs/sources/flow/reference/components/loki.process.md
+++ b/docs/sources/flow/reference/components/loki.process.md
@@ -773,12 +773,12 @@ Name            | Type      | Description                                       
 --------------- | --------- | ------------------------------------------------------------------- | ------- | --------
 `selector`      | `string`  | The LogQL stream selector and filter expressions to use.            |         | yes
 `pipeline_name` | `string`  | A custom name to use for the nested pipeline.                       | `""`    | no
-`action`        | `string`  | The action to take when the selector matches the log line. Supported values are "keep" and "drop" | `"keep"` | no
+`action`        | `string`  | The action to take when the selector matches the log line. Supported values are `"keep"` and `"drop"` | `"keep"` | no
 `drop_counter_reason` | `string` | A custom reason to report for dropped lines.                   | `"match_stage"` | no 
 
-The `match` block supports a number of `stage` inner blocks, same as
-the top-level one. These are used to construct the  nested set of stages to run
-if the selector matches the labels and content of the log entries.
+The `match` block supports a number of `stage` inner blocks, like the top-level
+block. These are used to construct the nested set of stages to run if the
+selector matches the labels and content of the log entries.
 
 The following blocks are supported inside the definition of `stage > match`:
 
@@ -789,7 +789,7 @@ stage          | [stage][]  | Processing stage to run. | no
 
 If the specified action is `"drop"`, the metric
 `loki_process_dropped_lines_total` is incremented with every line dropped.
-By default, the reason label is `match_stage`, but a custom reason can be
+By default, the reason label is `"match_stage"`, but a custom reason can be
 provided by using the `drop_counter_reason` argument.
 
 Let's see this in action, with the following log lines and stages
@@ -855,17 +855,17 @@ lines where the `applbl="foo"`. So, for the first line, the nested JSON stage
 adds `msg="app1 log line"` into the extracted map.
 
 The fourth stage uses the LogQL selector to only execute on lines where
-`applbl="qux"`; that means it won't match on any of our input, and the nested
-json stage does not run.
+`applbl="qux"`; that means it won't match any of the input, and the nested
+JSON stage does not run.
 
-The fifth stage drops any entries from on lines where `applbl` is set to
-'bar' and the line contents that matches the regex `.*noisy error.*`. It also
-increments the `loki_process_dropped_lines_total` metric with a label
-reason="discard_noisy_errors".
+The fifth stage drops entries from lines where `applbl` is set to 'bar' and the
+line contents matches the regex `.*noisy error.*`. It also increments the
+`loki_process_dropped_lines_total` metric with a label
+`drop_counter_reason="discard_noisy_errors"`.
 
 The final output stage changes the contents of the log line to be the value of
 `msg` from the extracted map. In this case, the first log entry's content is
-is changed to `app1 log line`.
+changed to `app1 log line`.
 
 
 ### drop block
@@ -891,13 +891,13 @@ not provided, the regex attempts to match the log line itself. If source is
 provided, the regex attempts to match the corresponding value from the
 extracted map.
 
-On the other hand, the `value` field can only work with values from the
-extracted map, and must be specified together with `source`. Entries are
-dropped when there is an exact match between the two.
+The `value` field can only work with values from the extracted map, and must be
+specified together with `source`. Entries are dropped when there is an exact
+match between the two.
 
 Whenever an entry is dropped, the metric `loki_process_dropped_lines_total`
-is incremented. By default, the reason label is `drop_stage`, but a custom
-one can be provided by using the `drop_counter_reason` argument.
+is incremented. By default, the reason label is `"drop_stage"`, but you can
+provide a custom label using the `drop_counter_reason` argument.
 
 The following stage drops log entries that contain the word `debug` _and_ are
 longer than 1KB.
@@ -911,8 +911,9 @@ stage {
 }
 ```
 
-The following stages drop entries that are 24h or older, are longer than
-8KB, _or_ the extracted value of 'app' is equal to foo.
+On the following example, we define multiple `drop` blocks so `loki.process`
+will drop entries that are either 24h or older, are longer than 8KB, _or_ the
+extracted value of 'app' is equal to foo.
 
 ```
 stage {
@@ -925,7 +926,7 @@ stage {
 stage {
 	drop {
 		older_than  = "8KB"
-		longer_than = "too long"
+		drop_reason = "too long"
 	}
 }
 
--- a/docs/sources/flow/reference/components/loki.process.md
+++ b/docs/sources/flow/reference/components/loki.process.md
@@ -366,25 +366,15 @@ Name          | Type      | Description                                         
 `source`      | `string`  | Name from extracted data to parse. If empty, uses the log message.  | `""`    | no
 
 
-The `expression` field needs to be a RE2 regex string. Every capture group (re) is set into the extracted map, so it must be named like: `(?P<name>re)`. The name of the capture group is then used as the key in the extracted map.
+The `expression` field needs to be a RE2 regex string. Every capture group (re)
+is set into the extracted map, so it must be named like: `(?P<name>re)`. The
+name of the capture group is then used as the key in the extracted map.
 
-<!--
-We don't care about YAML, what does River do instead???
+Because of how River treats backslashes in double-quoted strings, all
+backslashes in an `expression` must be escaped like `"\\S+"`.
 
-Because of how YAML treats backslashes in double-quoted strings, note that all backslashes in a regex expression must be escaped when using double quotes. For example, all of these are valid:
-
-expression: \w*
-expression: '\w*'
-expression: "\\w*"
-But these are not:
-
-expression: \\w* (only escape backslashes when using double quotes)
-expression: '\\w*' (only escape backslashes when using double quotes)
-expression: "\w*" (backslash must be escaped)
--->
-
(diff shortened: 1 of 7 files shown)


docs/sources/flow/reference/components/loki.process.md
```diff
@@ -530,6 +524,182 @@ Then, the second stage adds `user="John Doe"` to the label set of the log
...
+	drop {
+		older_than  = "24h"
+		drop_reason = "too old"
+	}
+}
+
+stage {
+	drop {
+		older_than  = "8KB"
+		longer_than = "too long"
```
mattdurham, line 691: Should this be `drop_reason`?

docs/sources/flow/reference/components/loki.process.md
```diff
@@ -530,6 +524,182 @@ Then, the second stage adds `user="John Doe"` to the label set of the log
...
+entry matches a configurable LogQL stream selector and filter expressions.
+
+The following arguments are supported:
+
+Name            | Type      | Description                                                         | Default | Required
+--------------- | --------- | ------------------------------------------------------------------- | ------- | --------
+`selector`      | `string`  | The LogQL stream selector and filter expressions to use.            |         | yes
+`pipeline_name` | `string`  | A custom name to use for the nested pipeline.                       | `""`    | no
+`action`        | `string`  | The action to take when the selector matches the log line. Supported values are "keep" and "drop" | `"keep"` | no
+`drop_counter_reason` | `string` | A custom reason to report for dropped lines.                   | `"match_stage"` | no
```
mattdurham, line 540: I assume there is a limit on length here? Should we define it.

docs/sources/flow/reference/components/loki.process.md
```diff
@@ -308,25 +312,15 @@ Name          | Type      | Description
...
 
-The `expression` field needs to be a Go RE2 regex string. Every capture group (re) is set into the extracted map, so it must be named like: `(?P<name>re)`. The name of the capture group is then used as the key in the extracted map.
+The `expression` field needs to be a RE2 regex string. Every capture group (re)
+is set into the extracted map, so it must be named like: `(?P<name>re)`. The
+name of the capture group is then used as the key in the extracted map.
 
-<!--
-We don't care about YAML, what does River do instead???
+Because of how River treats backslashes in double-quoted strings, all
+backslashes in an `expression` must be escaped like `"\\S+"`.
```
mattdurham, line 320: Would it be clearer to have `backslashes in an `expression` must be escaped with double backslash `\\` for example `"\\S+"``

Escaped like feels like you are showing that have to put more than \\ instead of calling it out specifically as an example.

docs/sources/flow/reference/components/loki.process.md
```diff
@@ -530,6 +524,182 @@ Then, the second stage adds `user="John Doe"` to the label set of the log
...
+
+Hierarchy      | Block      | Description | Required
+-------------- | ---------- | ----------- | --------
+stage          | [stage][]  | Processing stage to run. | no
+
+
+If the specified action is `"drop"`, the metric
+`loki_process_dropped_lines_total` is incremented with every line dropped.
+By default, the reason label is `match_stage`, but a custom reason can be
+provided by using the `drop_counter_reason` argument.
```
karengermond, line 556:
```suggestion
`loki_process_dropped_lines_total` is incremented with every line dropped.
By default, the reason label is `"match_stage"`, but a custom reason can be
provided by using the `drop_counter_reason` argument.
```

docs/sources/flow/reference/components/loki.process.md
```diff
@@ -530,6 +524,182 @@ Then, the second stage adds `user="John Doe"` to the label set of the log
...
+either apply a nested set of processing stages or drop an entry when a log
+entry matches a configurable LogQL stream selector and filter expressions.
+
+The following arguments are supported:
+
+Name            | Type      | Description                                                         | Default | Required
+--------------- | --------- | ------------------------------------------------------------------- | ------- | --------
+`selector`      | `string`  | The LogQL stream selector and filter expressions to use.            |         | yes
+`pipeline_name` | `string`  | A custom name to use for the nested pipeline.                       | `""`    | no
+`action`        | `string`  | The action to take when the selector matches the log line. Supported values are "keep" and "drop" | `"keep"` | no
```
karengermond, line 539:
```suggestion
`action`        | `string`  | The action to take when the selector matches the log line. Supported values are `"keep"` and `"drop"` | `"keep"` | no
```

karengermond: Made these consistent with other occurrences.

docs/sources/flow/reference/components/loki.process.md
karengermond, line 556:
```suggestion
If the specified action is `"drop"`, the metric
`loki_process_dropped_lines_total` is incremented with every line dropped.
By default, the reason label is `"match_stage"`, but a custom reason can be
provided by using the `drop_counter_reason` argument.
```

mattdurham (approved):
LGTM , a few documentation nits but nothing else. Also I assume we need to change the changelog.

docs/sources/flow/reference/components/loki.process.md
```diff
@@ -530,6 +524,182 @@ Then, the second stage adds `user="John Doe"` to the label set of the log
...
+The first two stages parse the log lines as JSON, decode the `app` value into
+the shared extracted map as `appname`, and use its value as the `applbl` label.
+
+The third stage uses the LogQL selector to only execute the nested stages on
+lines where the `applbl="foo"`. So, for the first line, the nested JSON stage
+adds `msg="app1 log line"` into the extracted map.
+
+The fourth stage uses the LogQL selector to only execute on lines where
+`applbl="qux"`; that means it won't match on any of our input, and the nested
+json stage does not run.
```
karengermond, line 622:
```suggestion
The fourth stage uses the LogQL selector to only execute on lines where
`applbl="qux"`; that means it won't match any of the input, and the nested
JSON stage does not run.
```

docs/sources/flow/reference/components/loki.process.md
```diff
@@ -530,6 +524,182 @@ Then, the second stage adds `user="John Doe"` to the label set of the log
...
+
+The fifth stage drops any entries from on lines where `applbl` is set to
+'bar' and the line contents that matches the regex `.*noisy error.*`. It also
+increments the `loki_process_dropped_lines_total` metric with a label
+reason="discard_noisy_errors".
```
karengermond, line 627:
```suggestion
The fifth stage drops entries from lines where `applbl` is set to
'bar' and the line contents match the regex `.*noisy error.*`. It also
increments the `loki_process_dropped_lines_total` metric with a label
`drop_counter_reason="discard_noisy_errors"`.
```

karengermond: This paragraph was confusing. Not sure my suggestion captures it correctly.

docs/sources/flow/reference/components/loki.process.md
```diff
@@ -530,6 +524,182 @@ Then, the second stage adds `user="John Doe"` to the label set of the log
...
+
+The final output stage changes the contents of the log line to be the value of
+`msg` from the extracted map. In this case, the first log entry's content is
+is changed to `app1 log line`.
```
karengermond, line 631:
```suggestion
The final output stage changes the contents of the log line to be the value of
`msg` from the extracted map. In this case, the first log entry's content is
changed to `app1 log line`.
```

docs/sources/flow/reference/components/loki.process.md
```diff
@@ -530,6 +524,182 @@ Then, the second stage adds `user="John Doe"` to the label set of the log
...
+`drop_counter_reason` | `string`   | A custom reason to report for dropped lines. | `"drop_stage"` | no
+
+The `expression` field needs to be a RE2 regex string. If `source` is empty or
+not provided, the regex attempts to match the log line itself. If source is
+provided, the regex attempts to match the corresponding value from the
+extracted map.
+
+On the other hand, the `value` field can only work with values from the
+extracted map, and must be specified together with `source`. Entries are
+dropped when there is an exact match between the two.
```
karengermond, line 659:
```suggestion
The `value` field can only work with values from the
extracted map, and must be specified together with `source`. Entries are
dropped when there is an exact match between the two.
```

docs/sources/flow/reference/components/loki.process.md
```diff
@@ -530,6 +524,182 @@ Then, the second stage adds `user="John Doe"` to the label set of the log
...
+
+Whenever an entry is dropped, the metric `loki_process_dropped_lines_total`
+is incremented. By default, the reason label is `drop_stage`, but a custom
+one can be provided by using the `drop_counter_reason` argument.
```
karengermond, line 663:
```suggestion
Whenever an entry is dropped, the metric `loki_process_dropped_lines_total`
is incremented. By default, the reason label is set to `drop_stage`, but you can
provide a custom label using the `drop_counter_reason` argument.
```

docs/sources/flow/reference/components/loki.process.md
```diff
@@ -530,6 +524,182 @@ Then, the second stage adds `user="John Doe"` to the label set of the log
...
+stage {
+	drop {
+		expression  = ".*debug.*"
+		longer_than = "1KB"
+	}
+}
+```
+
+The following stages drop entries that are 24h or older, are longer than
+8KB, _or_ the extracted value of 'app' is equal to foo.
```
karengermond, line 678: I think you should say why there's an OR here -- because there are multiple `drop` blocks.

ptodev:
Thanks for adding me as a reviewer, I browsed through it and learned a few things :)

docs/sources/flow/reference/components/loki.process.md
tpaschalis: You mean a limit on the length of the custom reason to report? Not sure about that, should be the maximum label length allowed by Loki, I'l look into it.

docs/sources/flow/reference/components/loki.process.md
```diff
@@ -308,25 +312,15 @@ Name          | Type      | Description
 `source`      | `string`  | Name from extracted data to parse. If empty, uses the log message.  | `""`    | no
 
 
-The `expression` field needs to be a Go RE2 regex string. Every capture group (re) is set into the extracted map, so it must be named like: `(?P<name>re)`. The name of the capture group is then used as the key in the extracted map.
+The `expression` field needs to be a RE2 regex string. Every capture group (re)
+is set into the extracted map, so it must be named like: `(?P<name>re)`. The
+name of the capture group is then used as the key in the extracted map.
```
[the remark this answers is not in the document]
tpaschalis: I've removed the (re) thing, as it was mainly confusing. This is how this reads now, is it any clearer?

```
The `expression` field must be a RE2 regex string. Every matched capture
group is added to the extracted map, so it must be named like: `(?P<name>re)`.
The name of the capture group is then used as the key in the extracted map for the 
matched value.
```

docs/sources/flow/reference/components/loki.process.md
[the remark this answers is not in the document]
tpaschalis: I've removed the reference to 'double-quoted' strings, as River only has that kind. How does that read now?

```
Because of how River strings work, any backslashes in `expression` must be
formatted as a double backslash; for example `"\\w"` or `"\\S+"`.
```

tpaschalis: I've removed the reference to 'double-quoted' strings, as River only has that kind. How does that read now?

```
Because of how River strings work, any backslashes in `expression` must be
formatted as a double backslash; for example `"\\w"` or `"\\S+"`.
```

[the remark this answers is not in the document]
mattdurham: Looks good

PR #2759, opened 2023-01-16
Do not marshal default or empty scraping service fields

jcreixell:
#### PR Description

  - Scraping service related fields produce high amounts of noise when marshaling YAML. This affects both the `/-/config` in the agent itself, as well remote agent management solutions using the Agent as a dependency.
  - This PR makes sure that configurations that do not explicitly enable scraping service or change its default configuration do not include such fields.
  - These fields include the keys `scraping_service`, `scraping_service_client` under metrics, as well as internal subkeys like `lifecycler` and `kvstore`.
  - Also excluded the `global` field under metrics, which is the only remaining top level field with nesting

Example YAML with an empty config:

```yaml
server:
  log_level: debug
  log_format: logfmt
metrics:
  wal_directory: data-agent/
  wal_cleanup_age: 12h0m0s
  wal_cleanup_period: 30m0s
  instance_restart_backoff: 5s
  instance_mode: shared
integrations:
  scrape_integrations: true
  integration_restart_backoff: 5s
  replace_instance_label: true
  use_hostname_label: true

```

#### Notes to the Reviewer

  - Unfortunately, when using `/-/config` the `scraping_service` field and most of its sub-fields are still marshaled. This is due to the way the agent applies default values *after* the Default configuration has been applied. This happens in multiple places, for example in `config.Config.Validate` (where the listen port for the lifecycler is set).
  - Some other top level fields we could clean up are less trivial (for example, `integrations` since we implement a custom marshaler using reflection. Probably not worth the hassle. 

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/metrics/agent_test.go
+++ b/pkg/metrics/agent_test.go
@@ -333,23 +333,11 @@ func makeInstanceConfig(name string) instance.Config {
 	return cfg
 }
 
-func TestAgent_MarshalYAMLOmitDefaultServiceClientConfig(t *testing.T) {
+func TestAgent_MarshalYAMLOmitDefaultConfigFields(t *testing.T) {
 	cfg := DefaultConfig
 	yml, err := yaml.Marshal(&cfg)
 	require.NoError(t, err)
 	require.NotContains(t, string(yml), "scraping_service_client")
-}
-
-func TestAgent_MarshalYAMLOmitDefaultServiceConfig(t *testing.T) {
-	cfg := DefaultConfig
-	yml, err := yaml.Marshal(&cfg)
-	require.NoError(t, err)
 	require.NotContains(t, string(yml), "scraping_service")
-}
-
-func TestAgent_MarshalYAMLOmitGlobalConfig(t *testing.T) {
-	cfg := DefaultConfig
-	yml, err := yaml.Marshal(&cfg)
-	require.NoError(t, err)
 	require.NotContains(t, string(yml), "global")
 }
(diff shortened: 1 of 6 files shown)


pkg/metrics/agent_test.go
```diff
@@ -332,3 +332,24 @@ func makeInstanceConfig(name string) instance.Config {
...
+	yml, err := yaml.Marshal(&cfg)
+	require.NoError(t, err)
+	require.NotContains(t, string(yml), "scraping_service")
+}
+
+func TestAgent_MarshalYAMLOmitGlobalConfig(t *testing.T) {
+	cfg := DefaultConfig
+	yml, err := yaml.Marshal(&cfg)
+	require.NoError(t, err)
+	require.NotContains(t, string(yml), "global")
```
tpaschalis, line 354: Silly nitpick; these along the others on the bottom could be a single test.

tpaschalis (approved):
LGTM, the empty configuration looks much more manageable now!

+1 to handling integrations on another path.

Side question, did we ever figure out the difference in Marshaling behavior of `gopkg.in/yaml.v2` and gopkg.in/yaml.v3`?

jcreixell:
> Side question, did we ever figure out the difference in Marshaling behavior of gopkg.in/yaml.v2 and gopkg.in/yaml.v3`?
I haven't looked into it. We seem to be using a combination of v2 and v3 in the codebase, but v2 is more prevalent. Definitely something to look into, maybe there is some room for consolidation

PR #2764, opened 2023-01-17
#2762 - fix(helm): Uses correct Arguments for starting in static mode

### LINKED ISSUE #2762: (helm) Grafana-Agent "-config.file" flag required error (fails when agent-mode is not flow)
I've been trying to debug an issue as part of installing the grafana-agent helm chart here: 
https://github.com/grafana/agent/blob/main/operations/helm/charts/grafana-agent/templates/containers/_agent.yaml#L5

There's some issues around the use of `AGENT_MODE` along with the need for the `run` arg.

The helm chart builds the following args (by default):

```
/bin/agent run /etc/agent/config.yaml
```

When agent-mode is not set to flow, startup fails:

```
docker run --entrypoint=bash -it docker.io/grafana/agent:v0.30.2
```
```
# create a dummy config file
root@8fc92b82b020:/# touch /tmp/config.yaml

# run it 
root@8fc92b82b020:/# /bin/agent run /tmp/config.yaml 
2023/01/16 18:06:42 -config.file flag required
```

If I remove the `run` arg and pass in `-config.file` it seems to work, but not sure this is valid...

```
root@813fe18a1fe1:/# /bin/agent -config.file=/tmp/config.yaml 
```

```
ts=2023-01-16T18:09:22.607032666Z caller=server.go:191 level=info msg="server listening on addresses" http=127.0.0.1:12345 grpc=127.0.0.1:12346 http_tls_enabled=false grpc_tls_enabled=false
...
```

With `AGENT_MODE=flow` it works (because it skips the config validation and returns early: https://github.com/grafana/agent/blob/main/cmd/grafana-agent/main.go#L49 )
```
AGENT_MODE=flow /bin/agent run /tmp/config.yaml 
```
```
ts=2023-01-16T18:15:35.495846284Z level=info msg="running usage stats reporter"
ts=2023-01-16T18:15:35.495817054Z level=info msg="now listening for http traffic" addr=127.0.0.1:12345
...
```

szczepad:
#### PR Description
The required arguments for starting the agent in flow vs. in static mode differ. Thus with the previous iteration of the chart it was only possible to start the agent in flow-mode. With this PR the default arguments are set under consideration of the selected mode.

Fixes #2762 

#### PR Checklist

- [x] CHANGELOG updated
- [ ] Documentation added
- [x] Tests updated


Ronan-WeScale:
Hi @szczepad !

I test your branch and is not working in static mode (help is displayed).  I think is needed to change the watch argument cmd too.
I try to fix it in my side.

operations/helm/charts/grafana-agent/templates/containers/_agent.yaml
```diff
@@ -2,13 +2,19 @@
...
     - run
     - /etc/agent/{{ include "grafana-agent.config-map.key" . }}
     - --storage.path={{ .Values.agent.storagePath }}
     - --server.http.listen-addr={{ .Values.agent.listenAddr }}:{{ .Values.agent.listenPort }}
     {{- if not .Values.agent.enableReporting }}
     - --disable-reporting
     {{- end}}
+    {{- end}}
+    {{- if eq .Values.agent.mode "static"}}
+    - -config-file=/etc/agent/{{ include "grafana-agent.config-map.key" . }}
```
nabadger, line 15: Think this should be `config.file` instead of `config-file` :)

szczepad: Naturally. 🙈 thank you! Adapted it.

Ronan-WeScale: it's work, what about the reload config cmd ? 
https://github.com/grafana/agent/blob/8c953cfd8ffb8aba4cbee2638cc077d3ba0ff6f8/operations/helm/charts/grafana-agent/templates/containers/_watch.yaml#L11

Ronan-WeScale:
I also made this change in my fork : 
https://github.com/Ronan-WeScale/grafana-agent-fix/commit/beaba62fc2f17864a06ff54722ecce31a8d675fb

to use agent.configMap.content instead of config; I think it's duplicate, "agent.configMap.content" is used to enable custom config and "config" is not referenced in values.yaml ..

tell me if I need to create PR for this

szczepad:
Hey @Ronan-WeScale thanks for your support! :) 
I've merged your changes **on the watcher** (and rendered the test-charts, as this seems to be necessary.)

Although I haven't merged the changes to the configMap, as there seems to be an Issue and PR already present for that topic. See #2756 . :) Naturally we might also fix that one. But that might be for a maintainer to decide.

tpaschalis:
Hey, thanks for the PR! Couple of small points:

a) I think that #2735 made it so we don't always have to version bump, but can batch many changes into one release. Do you think it makes sense to keep the version as-is and list the change under Unreleased for a future release?

b) Like you said, let's fix one issue at a time and leave the configMap changes to #2757.

c) Our CLA Assistant required every committer to have signed the CLA. Can @Ronan-WeScale sign the CLA to get listed as a co-author?

szczepad:
Hey, @tpaschalis 
a) Theoretically fine for me. :) On the other hand this seems to be a breaking bug for people running in static mode. As as far as I understood it's not possible to run the chart in this case without manually fiddling in the manifest of the agent. So a release/bump miiight also be justified, depending on the maturity of the chart. But yeah. :D doesn't really make a difference for me, so whatever means the least amount of effort.

Ronan-WeScale:
Hey @tpaschalis,
Thank for your answer ! 
a) We can version pump with this PR and #2757 ;) today, custom config not work at all.
b) ok thank, sorry I don't find it :/ 
c) done

rfratto (approved):
Thank you! I'll wait for #2757 to merge first and then release this as 0.2.3.

PR #2774, opened 2023-01-19
Lazy Initialize Remote Config Metrics

spartan0x117:
#### PR Description
Lazily registers `remote_config` metrics to avoid adding metrics unnecessarily when `remote-configs` and `agent-management` features are not enabled.

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,8 @@ Main (unreleased)
 
 - Flow: `prometheus.relabel` and `prometheus.remote_write` will now error if they have exited. (@ptodev)
 
+- Agent Management: Fix `remote_config_fetch_errors_total` metric being registered even when `remote-configs` and `agent-management` features were not enabled. (@spartan0x117)
+
 ### Other changes
 
 - Use Go 1.19.4 for builds. (@erikbaranowski)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,8 +66,6 @@ Main (unreleased)
 
 - Flow: `prometheus.relabel` and `prometheus.remote_write` will now error if they have exited. (@ptodev)
 
-- Agent Management: Fix `remote_config_fetch_errors_total` metric being registered even when `remote-configs` and `agent-management` features were not enabled. (@spartan0x117)
-
 ### Other changes
 
 - Use Go 1.19.4 for builds. (@erikbaranowski)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -66,6 +66,8 @@ Main (unreleased)
 
 - Flow: `prometheus.relabel` and `prometheus.remote_write` will now error if they have exited. (@ptodev)
 
+- Agent Management: Fix `remote_config_fetch_errors_total` metric being registered even when `remote-configs` and `agent-management` features were not enabled. (@spartan0x117)
```
rfratto, line 69: Since this PR is changing code that hasn't been released yet, it wouldn't go into the changelog (e.g., the changelog should only include things which have changed since the last release)

pkg/config/instrumentation/remote_config_metrics.go
```diff
@@ -34,10 +40,12 @@ func newRemoteConfigMetrics() *remoteConfigMetrics {
 	return &remoteConfigMetrics
 }
 
-func (r *remoteConfigMetrics) InstrumentRemoteConfigFetch(code int) {
-	r.fetchStatusCodes.WithLabelValues(fmt.Sprintf("%d", code)).Inc()
+func InstrumentRemoteConfigFetch(code int) {
```
marctc, line 43: nit: could we rename this param to `statusCode`? Same for label

pkg/config/instrumentation/remote_config_metrics.go
```diff
@@ -34,10 +40,12 @@ func newRemoteConfigMetrics() *remoteConfigMetrics {
...
+	metricsInitializer.Do(initializeRemoteConfigMetrics)
```
jcreixell, line 44: would it be easier to add `remoteConfigMetrics` to the `httpProvider` type and initialize it in `newHTTPProvider` (instead of globally)? or do we intent to use these metrics in other places?

spartan0x117: My thinking was that we may want to add some agent management-specific metrics here in the future, which would be used in `agentmanagement.go`

pkg/config/instrumentation/remote_config_metrics.go
spartan0x117: Makes sense, and done!

marctc (approved):

PR #2784, opened 2023-01-19
Add Default Cache Directory for Agent Management

spartan0x117:
#### PR Description
Adds `data-agent/` as the default location for caching remote configs, along with tests.

Also removes the unnecessary export of the `RemoteConfiguration` struct.

- [ ] CHANGELOG updated
- [ ] Documentation added
- [X] Tests updated

### DIFF
--- a/pkg/config/agentmanagement.go
+++ b/pkg/config/agentmanagement.go
@@ -21,7 +21,7 @@ var defaultConfig = AgentManagement{
 
 type labelMap map[string]string
 
-type remoteConfiguration struct {
+type RemoteConfiguration struct {
 	Labels    labelMap `yaml:"labels"`
 	Namespace string   `yaml:"namespace"`
 }
@@ -34,7 +34,7 @@ type AgentManagement struct {
 	PollingInterval string           `yaml:"polling_interval"`
 	CacheLocation   string           `yaml:"remote_config_cache_location"`
 
-	RemoteConfiguration remoteConfiguration `yaml:"remote_configuration"`
+	RemoteConfiguration RemoteConfiguration `yaml:"remote_configuration"`
 }
 
 // UnmarshalYAML implements yaml.Unmarshaler.
--- a/pkg/config/agentmanagement.go
+++ b/pkg/config/agentmanagement.go
@@ -15,9 +15,13 @@ import (
 
 const cacheFilename = "remote-config-cache.yaml"
 
+var defaultConfig = AgentManagement{
+	CacheLocation: "data-agent/",
+}
+
 type labelMap map[string]string
 
-type RemoteConfiguration struct {
+type remoteConfiguration struct {
 	Labels    labelMap `yaml:"labels"`
 	Namespace string   `yaml:"namespace"`
 }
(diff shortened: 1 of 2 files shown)


pkg/config/agentmanagement_test.go
```diff
@@ -138,3 +139,46 @@ func TestFullUrl(t *testing.T) {
...
+
+func TestUnmarshal(t *testing.T) {
+	cfg := `
+api_url: http://localhost:8080
+basic_auth:
+  username: test_user
+  password_file: /tmp/test/passfile
+protocol: http
+polling_interval: 1m
+remote_configuration:
```
jcreixell, line 161: shall we include here the cache _location to make sure the default is properly overriden?

pkg/config/agentmanagement.go
```diff
@@ -15,9 +15,13 @@ import (
 
 const cacheFilename = "remote-config-cache.yaml"
 
+var defaultConfig = AgentManagement{
+	CacheLocation: "data-agent/",
+}
+
 type labelMap map[string]string
 
-type RemoteConfiguration struct {
```
jcreixell, line 20: While this works for now, I would still export it for consistency with other parts of the code where we define configuration types

jcreixell (changes requested):
Looks good overall, added some comments, please have a look

pkg/config/agentmanagement.go
```diff
@@ -15,9 +15,13 @@ import (
 
 const cacheFilename = "remote-config-cache.yaml"
 
+var defaultConfig = AgentManagement{
+	CacheLocation: "data-agent/",
```
marctc, line 19: This path by default doesn't work  if I try to run the agent locally

marctc: Furthermore, the agent returns an error but doesn't panic. Polling works but it's not able to save the config file locally

mattdurham: `data-agent` can be defined differently based on the command line flags for the wal, so you may want to utilize those command line flags  as the default,`metrics.wal-directory`.

mattdurham: Are all these features linux only? IE tmp likely does not exist on windows.

spartan0x117: Ahh that's a good point, I often forget since I develop on Unix/Linux. Perhaps the best course of action would be to make it a required field, then, if agent management is enabled?

_Or_ the default could be the directory containing the initial config, which would always exist (though may not be writable). In the case that no directory is specified in the `agent_management` block and the directory containing the initial config is not writable, then it would panic.

jcreixell: I think there is plenty of good insight on how to do cross-platform persistence in this file https://github.com/grafana/agent/blob/main/pkg/usagestats/reporter.go#L93-L99, where we persist the uid for stats reporting. 

I see 2 options:

* Doing something similar to what we do to persist the uid. I do not like this option because defaulting to something like `/tmp` is dangerous: If the file is gone (tmp cleanup) and the API is down, agents will panic.
* Move the default WAL dir outside of the metrics package into a package with defaults under `config`. That way we can keep the data dir in sync and it is probably a good idea to have a default data folder for the entire agent anyway (and not just for the wal). We could eventually provide defaults for things like `positions_directory` there too.

However, the more I think about it, the more I think we shouldn't provide a default for this field at this point. Not at least until we agree on a default agent data dir we can use globally (changing the default dir later could potentially break existing installations).

What do the others think?

tpaschalis: Not sure if the chosen `data-agent` path here is chosen for compatibility with Flow's defaults or a coincidence.

For my 2c, gut feeling is the same as Jorge's; I personally would _not_ opt for a default agent dir to be used globally, but have it as a required flag when using agent-management.

rfratto: Flow supports an agent-wide data directory (named `data-agent` by default), so we have one going forward. I'm not sure if it's worth backporting one to the static mode just for agent management; it'd be a fair amount of work.

mattdurham: Agree with @jcreixell on making the setting required if Agent Management is turned on.

spartan0x117:
Closing this PR as the consensus looks to be not to set a default directory, at the very least for now.

I will follow up with a separate PR to a) change `remote_config_cache_location` -> `remote_config_cache_directory` to be consistent with some other similar fields and b) make it a required field.

PR #2787, opened 2023-01-20
WIP: Convert standard errors to Diagnostic errors

ptodev:
I am looking for comments on how to use the Diagnostic type in a good way. Especially when reporting position information.

### DIFF
--- a/pkg/river/vm/vm.go
+++ b/pkg/river/vm/vm.go
@@ -203,6 +203,8 @@ func prepareDecodeValue(v reflect.Value) reflect.Value {
 	return v
 }
 
+// TODO: Should diag.Diagnostic be returned as a pointer? Or should we make a Diagnostic interface?
+// TODO: For nil diagnostics, should we make a diagn.Null type, like the "var Null = Value{}" in value.go?
 func (vm *Evaluator) evaluateExpr(scope *Scope, assoc map[value.Value]ast.Node, expr ast.Expr) (v value.Value, err *diag.Diagnostic) {
 	defer func() {
 		if v != value.Null {
@@ -387,7 +389,7 @@ func (vm *Evaluator) evaluateExpr(scope *Scope, assoc map[value.Value]ast.Node, 
 				Severity: diag.SeverityLevelError,
 				StartPos: ast.StartPos(expr).Position(),
 				EndPos:   ast.EndPos(expr).Position(),
-				Message:  err.Error(),
+				Message:  plainErr.Error(),
 			}
 		}
 
@@ -420,7 +422,7 @@ func (vm *Evaluator) evaluateExpr(scope *Scope, assoc map[value.Value]ast.Node, 
 				Severity: diag.SeverityLevelError,
 				StartPos: ast.StartPos(expr).Position(),
 				EndPos:   ast.EndPos(expr).Position(),
-				Message:  err.Error(),
+				Message:  plainErr.Error(),
 			}
 		}
 
--- a/pkg/river/vm/vm.go
+++ b/pkg/river/vm/vm.go
@@ -203,7 +203,7 @@ func prepareDecodeValue(v reflect.Value) reflect.Value {
 	return v
 }
 
-func (vm *Evaluator) evaluateExpr(scope *Scope, assoc map[value.Value]ast.Node, expr ast.Expr) (v value.Value, err error) {
+func (vm *Evaluator) evaluateExpr(scope *Scope, assoc map[value.Value]ast.Node, expr ast.Expr) (v value.Value, err *diag.Diagnostic) {
 	defer func() {
 		if v != value.Null {
 			assoc[v] = expr
@@ -212,7 +212,17 @@ func (vm *Evaluator) evaluateExpr(scope *Scope, assoc map[value.Value]ast.Node, 
 
 	switch expr := expr.(type) {
 	case *ast.LiteralExpr:
-		return valueFromLiteral(expr.Value, expr.Kind)
+		val, err := valueFromLiteral(expr.Value, expr.Kind)
+		if err == nil {
+			return val, nil
+		} else {
+			return value.Null, &diag.Diagnostic{
+				Severity: diag.SeverityLevelError,
+				StartPos: ast.StartPos(expr).Position(),
+				EndPos:   ast.EndPos(expr).Position(),
+				Message:  err.Error(),
+			}
+		}
 
 	case *ast.BinaryExpr:
 		lhs, err := vm.evaluateExpr(scope, assoc, expr.Left)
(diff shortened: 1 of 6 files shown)


pkg/river/vm/vm.go
```diff
@@ -173,26 +173,59 @@ func (vm *Evaluator) evaluateBlockOrBody(scope *Scope, assoc map[value.Value]ast
 			continue
 
 		case tf.IsAttr() && len(blocks) > 0:
-			return fmt.Errorf("%q must be an attribute, but is used as a block", fullName)
+			return diag.Diagnostic{
+				Severity: diag.SeverityLevelError,
+				StartPos: ast.StartPos(blocks[0]).Position(),
+				EndPos:   ast.EndPos(blocks[len(blocks)-1]).Position(),
```
rfratto, line 179: The problem here is that the highlight range might include blocks which are fine. Imagine this scenario:

```river 
invalid_block_name { ... } 
valid_block_name { ... } 
invalid_block_name { ... } 
```

The error will report the entire range from the first and last block as invalid, even though the block in the middle is fine.

Rather than reporting one diagnostic for the entire set of blocks, you can return `diag.Diagnostics` (a slice of `diag.Diagnostic`) and add a diagnostic for each invalid block in `blocks`.

pkg/river/vm/vm.go
```diff
@@ -173,26 +173,59 @@ func (vm *Evaluator) evaluateBlockOrBody(scope *Scope, assoc map[value.Value]ast
...
+				Message:  fmt.Sprintf("%q must be an attribute, but is used as a block", fullName),
+			}
 		case tf.IsAttr() && len(attrs) == 0 && !tf.IsOptional():
-			return fmt.Errorf("missing required attribute %q", fullName)
+			return diag.Diagnostic{
+				Severity: diag.SeverityLevelError,
+				//TODO: We need to report the start and end position of the block where this attribute should be?
+				StartPos: ast.StartPos().Position(),
+				EndPos:   ast.EndPos().Position(),
```
rfratto, line 187: Use the starting and ending position of the `node` argument, which will be the entire surrounding block. 

For example, this would mean that 

```river 
prometheus.scrape {
  targets = [ ... ] 
}
```

will highlight the whole component as invalid because it is missing the "forward_to" required attribute.

pkg/river/vm/vm.go
```diff
@@ -173,26 +173,59 @@ func (vm *Evaluator) evaluateBlockOrBody(scope *Scope, assoc map[value.Value]ast
...
+				Message:  fmt.Sprintf("attribute %q may only be set once", fullName),
+			}
 
 		case tf.IsBlock() && len(attrs) > 0:
-			return fmt.Errorf("%q must be a block, but is used as an attribute", fullName)
+			return diag.Diagnostic{
+				Severity: diag.SeverityLevelError,
+				StartPos: ast.StartPos(attrs[0]).Position(),
+				EndPos:   ast.EndPos(attrs[len(attrs)-1]).Position(),
+				Message:  fmt.Sprintf("%q must be a block, but is used as an attribute", fullName),
```
rfratto, line 206: I have the same concern here as I do for the blocks: we should return one diagnostic per invalid attribute rather than one diagnostic overall for the condition.

pkg/river/vm/vm.go
```diff
@@ -173,26 +173,59 @@ func (vm *Evaluator) evaluateBlockOrBody(scope *Scope, assoc map[value.Value]ast
...
+			}
 		case tf.IsBlock() && len(blocks) == 0 && !tf.IsOptional():
 			// TODO(rfratto): does it ever make sense for children blocks to be required?
-			return fmt.Errorf("missing required block %q", fullName)
+			return diag.Diagnostic{
+				Severity: diag.SeverityLevelError,
+				//TODO: We need to report the start and end position of the block where this block should be?
+				StartPos: ast.StartPos().Position(),
+				EndPos:   ast.EndPos().Position(),
```
rfratto, line 214: same comment for required attribute: use the start and end position of the `node` argument to mark the surrounding block as invalid.

pkg/river/vm/vm.go
```diff
@@ -173,26 +173,59 @@ func (vm *Evaluator) evaluateBlockOrBody(scope *Scope, assoc map[value.Value]ast
...
 		case len(attrs) > 0 && len(blocks) > 0:
 			// NOTE(rfratto): it's not possible to reach this condition given the
 			// statements above, but this is left in defensively in case there is a
 			// bug with the validity checks.
-			return fmt.Errorf("%q may only be used as a block or an attribute, but found both", fullName)
+			return diag.Diagnostic{
+				Severity: diag.SeverityLevelError,
+				//TODO: We need to report the start and end position of the parent block?
+				StartPos: ast.StartPos().Position(),
+				EndPos:   ast.EndPos().Position(),
```
rfratto, line 226: It would be nice to report one of these for each declaration of the attribute/block fter the first declaration, but given it's not possible to reach this condition, I don't think it's worth figuring out if `attrs[0]` or `blocks[0]` comes first in the file.

For now, you can use the position of the `node` argument to report the error.

ptodev:
Hi @rfratto - do you think we should also change this function from [diag.go](https://github.com/grafana/agent/blob/main/pkg/river/diag/diag.go#L53-L55) to print the end position:
```
func (d Diagnostic) Error() string {
	return fmt.Sprintf("%s: %s", d.StartPos, d.Message)
}
```
Otherwise we won't see the end position in unit test error messages, and then it'd be harder to make sure it is reported correctly? Also, if we don't report the end position, I'm not sure if there is any point in filling it in when reporting the error?

rfratto:
> Otherwise we won't see the end position in unit test error messages, and then it'd be harder to make sure it is reported correctly? Also, if we don't report the end position, I'm not sure if there is any point in filling it in when reporting the error?

It's used by the pretty diagnostic printer: https://github.com/grafana/agent/blob/main/pkg/river/diag/printer.go

I'm a little worried about making the string form too hard to parse. If you need to make sure the end position is correct in a unit test, you can assert against the Diagnostic value as a whole instead of testing against the result of calling `String()`.

ptodev:
Hi @rfratto, would you mind if I make all functions in vm.go return `diag.Diagnostic` instead of an `error` in the function prototype? I think it might help make the code cleaner, because functions like evaluateExpr() themselves recursively and that way it wouldn't have to check whether it got a diagnostic or a regular error.

pkg/river/vm/vm.go
```diff
@@ -131,6 +132,9 @@ func (vm *Evaluator) evaluateBlockOrBody(scope *Scope, assoc map[value.Value]ast
 		Assoc:   assoc,
 		TagInfo: ti,
 	}
+	//TODO(ptodev): if Decode() returns an error which is not a diagnostic, we should assume that
+	// it was because the error applies to the whole block and we should repackage this error here
+	// as a diagnostic with the start and end position of the block.
```
ptodev, line 137: @rfratto We had a discussion on Slack where I mentioned that `sd.Decode` doesn't have the brace position of the block because we only pass the body and not the block itself. You suggested to pass the block to `sd.Decode` insteadn of the body. But in the case of evaluating the main body of the river file, we would not have a root level block and root level braces. Hence we cannot **only** pass a block to the `sd.Decode` function. 

I am not sure what is the best way to get `sd.Decode` to print diagnostic info with curly brace start and end position of the block for the `missing required attribute` and `missing required block` errors.

The easiest and most hacky solution feels like accepting an `ast.Node` as a parameter for `sd.Decode`, but I wonder if there is a more type safe way to do it which doesn't involve a type assertion.

Also, what if we have a `missing required attribute` or `missing required block` error on the root level, where there is no block? I guess we should print the start and end position of the whole river config?

Personally, I think the easiest solution is to report both the start and end position in the error message, and keep using the body instead of the block at all times for `sd.Decode`. Then, even though we don't report the position of the curly braces it'd be clear to the user that we are erroring on the whole range of the block. However, I think you had reservations for making the non-pretty error printing include more information?

rfratto:
River has moved to a different repo (grafana/river) so I'm going to close this PR. It's also probably not super pressing since we haven't seen a large amount of people complain about these specific errors.

Issue #2763, opened 2023-01-16
Flow: create `loki.source.podlogs` component to collect logs based on PodLogs CRDs 


rfratto:
Related to #2504, but instead of being given a list of targets, it will be given rules to discover PodLogs resources within the cluster and use that to define the set of targets to collect logs from. 

Like #2504, it can use the Kubernetes API to collect logs (instead of files mounted from the host). This would allow it to share a lot of code in common with `loki.source.kubernetes`.

PR #2788, opened 2023-01-20
pkg/runner: introduce common package for goroutine scheduling

rfratto:
Many of our Flow components share a behavior of managing goroutines, where they must synchronize the set of running goroutines with a set of tasks to perform:

* `loki.source.file` must manage tailer goroutines for each logs target.
* `phalre.scrape` must manager scraper goroutines for each profiling target.
* `prometheus.scrape` must manage scraper goroutines for each metrics target.

More components with this type of goroutine management will be added in the future (#2504, #2763). Even the Flow controller implements exactly the same goroutine management to manage the set of running components.

Throughout all of these uses, the goroutine management works the same way:

* Start new goroutines for tasks (e.g., targets) that have appeared.
* Shut down old goroutines for tasks that have gone away.
* Maintain existing goroutines as long as the task continues to exist.

This commit introduces a new utility package, `pkg/runner`, which creates an abstraction for this behavior. `pkg/runner` defines three concepts:

* Task: a unit of work that a goroutine should be performing, such as a metrics target.
* Worker: a goroutine which will act against a Task, such as scraping a metrics target on an interval.
* Runner: a mechanism that controls the set of running Workers based on Tasks given to it.

The Runner is a declarative interface; you call `ApplyTasks` with the entire set of Tasks to run, and it will synchronize the set of Workers with the given Tasks, shutting down Workers for Tasks that have gone away, creating new Workers for new Tasks, and maintaining existing Workers for Tasks which still exist.

The Task interface is defined as:

```go
type Task interface {
  Hash() uint64
  Equals(other Task) bool
}
```

Tasks are tracked internally in a hash map. The `Hash` method is invoked to figure out what bucket to put a Task in. If two Tasks happen to have the same hash, the `Equals` method is called to check if they are equal. Completely identical tasks (with the same hash and where `Equals` returns true) will be deduplicated so only one task exists.

Meanwhile, the Worker interface is defined as:

```go
type Worker interface {
  Run(ctx context.Context) error
}
```

The `Run` method is invoked when a Worker should start running. Workers should continue running until there is an error or the provided context is canceled, which is done by the Runner when a Task assigned to a Worker is removed.

There is exactly one Worker for each unique Task. Each Worker is responsible for exactly one task for its entire lifetime. A Runner is constructed with a function to map Tasks to Workers:

```go
func New[TaskType Task](newWorker func(t TaskType) Worker) *Runner[TaskType]
```

`newWorker` is only invoked when the Runner is given a Task for which no Runner is assigned to. To provide Tasks to a runner, call `ApplyTasks`.

See included test code for example of how to use the `runner` package.

Benchmarks are included to test the hashMap, where the bulk of the performance will be seen. Benchmark results show modest performance for the initial commit:

    Benchmark_hashMap/Add_element-10                                 7875709           133.3  ns/op
    Benchmark_hashMap/Remove_element-10                             30147565            48.13 ns/op
    Benchmark_hashMap/Lookup_element-10                             85129270            14.20 ns/op
    Benchmark_hashMap/Add_existing_element-10                       61546616            18.85 ns/op
    Benchmark_hashMap_collisions/1_collisions/Add-10                 5402604           215.6  ns/op
    Benchmark_hashMap_collisions/1_collisions/Remove-10             32215772            37.19 ns/op
    Benchmark_hashMap_collisions/1_collisions/Lookup-10             40548360            29.54 ns/op
    Benchmark_hashMap_collisions/2_collisions/Add-10                 4416198           270.9  ns/op
    Benchmark_hashMap_collisions/2_collisions/Remove-10             21474396            55.78 ns/op
    Benchmark_hashMap_collisions/2_collisions/Lookup-10             26042961            45.52 ns/op
    Benchmark_hashMap_collisions/5_collisions/Add-10                 3013473           396.2  ns/op
    Benchmark_hashMap_collisions/5_collisions/Remove-10             10754502           111.6  ns/op
    Benchmark_hashMap_collisions/5_collisions/Lookup-10             11606706           102.8  ns/op
    Benchmark_hashMap_collisions/10_collisions/Add-10                1863262           644.7  ns/op
    Benchmark_hashMap_collisions/10_collisions/Remove-10             5846124           204.8  ns/op
    Benchmark_hashMap_collisions/10_collisions/Lookup-10             5014551           241.2  ns/op
    Benchmark_hashMap_collisions/100_collisions/Add-10                 81848         14595    ns/op
    Benchmark_hashMap_collisions/100_collisions/Remove-10             635085          1877    ns/op
    Benchmark_hashMap_collisions/100_collisions/Lookup-10              90588         13244    ns/op

As expected, the performance steadily decreases when more hash collisions are introduced. Proper usage of pkg/runner will be to make sure that there is good hash coverage through different tasks.

### DIFF
--- a/pkg/runner/hash_map.go
+++ b/pkg/runner/hash_map.go
@@ -1,0 +1,96 @@
+package runner
+
+import "sync"
+
+type hashMap struct {
+	mut    sync.RWMutex
+	lookup map[uint64][]Task
+}
+
+// newHashMap creates a new hashMap allocated to handle at least size unique
+// hashes.
+func newHashMap(size int) *hashMap {
+	return &hashMap{
+		lookup: make(map[uint64][]Task, size),
+	}
+}
+
+// Add adds the provided Task into the hashMap. It returns false if t
+// already exists in the hashMap.
+func (hm *hashMap) Add(t Task) bool {
+	hm.mut.Lock()
+	defer hm.mut.Unlock()
+
+	hash := t.Hash()
+	for _, compare := range hm.lookup[hash] {
+		if compare.Equals(t) {
+			return false
+		}
+	}
+
+	hm.lookup[hash] = append(hm.lookup[hash], t)
+	return true
+}
+
+// Has returns true if t exists in the hashMap.
+func (hm *hashMap) Has(t Task) bool {
+	hm.mut.RLock()
+	defer hm.mut.RUnlock()
+
+	for _, compare := range hm.lookup[t.Hash()] {
+		if compare.Equals(t) {
+			return true
+		}
+	}
+
+	return false
+}
+
+// Delete removes the provided Task from the hashMap. It returns true if the
+// Task was found and deleted.
+func (hm *hashMap) Delete(t Task) (deleted bool) {
+	hm.mut.Lock()
+	defer hm.mut.Unlock()
+
+	hash := t.Hash()
+
+	var rem []Task
+	for _, s := range hm.lookup[hash] {
+		if s.Equals(t) {
+			deleted = true
+			continue
+		}
+		rem = append(rem, s)
+	}
+	if len(rem) == 0 {
+		delete(hm.lookup, hash)
+	} else {
+		hm.lookup[hash] = rem
+	}
+
+	return deleted
+}
+
+// Iterate returns a channel which iterates through all elements in the
+// hashMap. The channel *must* be fully consumed, otherwise the hashMap will
+// deadlock.
+//
+// The iteration order is not guaranteed.
+func (hm *hashMap) Iterate() <-chan Task {
+	taskCh := make(chan Task)
+
+	go func() {
+		hm.mut.Lock()
+		defer hm.mut.Unlock()
+
+		for _, set := range hm.lookup {
+			for _, task := range set {
+				taskCh <- task
+			}
+		}
+
+		close(taskCh)
+	}()
+
+	return taskCh
+}
--- a/pkg/runner/hash_map.go
+++ b/pkg/runner/hash_map.go
@@ -54,18 +54,18 @@ func (hm *hashMap) Delete(t Task) (deleted bool) {
 
 	hash := t.Hash()
 
-	var rem []Task
+	var remaining []Task
 	for _, s := range hm.lookup[hash] {
 		if s.Equals(t) {
 			deleted = true
 			continue
 		}
-		rem = append(rem, s)
+		remaining = append(remaining, s)
 	}
-	if len(rem) == 0 {
+	if len(remaining) == 0 {
 		delete(hm.lookup, hash)
 	} else {
-		hm.lookup[hash] = rem
+		hm.lookup[hash] = remaining
 	}
 
 	return deleted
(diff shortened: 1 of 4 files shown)


tpaschalis:
Looks good as a first pass, I'll take a second look later on. Two things pop to mind for now.

About having both a Hash() and Equals() methods. Is the intention here to allow each task type have its own implementation (eg. a very short and rapid firing task having a cheap hash, while long-running ones can work with an expensive one?). Would we ever just have a more expensive hash and only keep one of the two methods?

Should callers be able to access to the error returned by Run, or should we get rid of it?

rfratto:
> About having both a Hash() and Equals() methods. Is the intention here to allow each task type have its own implementation (eg. a very short and rapid firing task having a cheap hash, while long-running ones can work with an expensive one?). Would we ever just have a more expensive hash and only keep one of the two methods?

The intention is that `Hash()` should be cheaper to call than `Equals()`. Each task type should implement both, but given `Hash()` always returns the same value for the same tasks (tasks should be immutable), it can be cached by the task for future calls to be faster. 

`Hash()` is called to figure out what bucket to put the task in, and then `Equals()` is calls for tasks within that bucket (i.e., tasks which have the same hash) to do deduplication.

rfratto:
> Should callers be able to access to the error returned by Run, or should we get rid of it?

Ah, yeah. There's two things I could do here:

1. Have the runner log the error returned by Run
2. Remove the error from the implementation, and assume that Workers will be logging errors if there are any.

WDYT?

rfratto:
> Remove the error from the implementation, and assume that Workers will be logging errors if there are any.

@tpaschalis and I discussed this offline and he suggested this approach.

pkg/runner/hash_map.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+}
+
+// Add adds the provided Task into the hashMap. It returns false if t
+// already exists in the hashMap.
+func (hm *hashMap) Add(t Task) bool {
+	hm.mut.Lock()
+	defer hm.mut.Unlock()
+
+	hash := t.Hash()
```
mattdurham, line 24: Can this be simplified to use `Has` and move the locks to after the `Has` call?

rfratto: Technically no, if two `Add` calls run in parallel with the same task, there's a chance that the task gets added twice depending on if they both call `Has` at the same time before obtaining the lock.

rfratto: (But even if the lock was held for the entirety of Add, re-using Has here means the hash needs to get generated twice, so this performs better)

pkg/runner/hash_map.go
```diff
@@ -0,0 +1,96 @@
...
+func (hm *hashMap) Delete(t Task) (deleted bool) {
+	hm.mut.Lock()
+	defer hm.mut.Unlock()
+
+	hash := t.Hash()
+
+	var rem []Task
+	for _, s := range hm.lookup[hash] {
+		if s.Equals(t) {
+			deleted = true
```
mattdurham, line 60: This feels off in the naming. I read `rem` as `remove`, but I think it means `remaining`?

rfratto: Yep, I'll rename 👍

mattdurham (approved):
LGTM

PR #2798, opened 2023-01-23
Add loki source heroku component

### LINKED ISSUE #2511: `loki.source.heroku`: Receives logs from a Heroku application by using Heroku HTTPS Drain

erikbaranowski:
#### PR Description

The PR adds a loki.source.heroku component that can read logs from Heroku cloud resources by listening on a drain endpoint.

The log entries are then forwarded to other loki.* components.

Most of the PR size is due to bringing over Promtail's heroku package, so that we can use our own loki.Entry definition. This was also a chance to clean things up to not use the Promtail Target interface and distinguish the configuration struct for each strategy.

Fixes #2511 

#### PR Checklist

- [ X ] CHANGELOG updated
- [ X ] Documentation added
- [ X ] Tests updated


erikbaranowski:
much of the go.mod and go.sum changes were a result of a `go mod tidy`. If these should not be part of this PR let me know and I can back them out.

docs/sources/flow/reference/components/loki.source.heroku.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Usage
+
+```river
+loki.source.heroku "LABEL" {
+  listener {
+    address = "LISTEN_ADDRESS"
+    port = PORT
+  }
+  ...
```
tpaschalis, line 32: We can't have more than one `listener` block, so can we remove the ellipses?

docs/sources/flow/reference/components/loki.source.heroku.md
```diff
@@ -0,0 +1,136 @@
...
+
+```river
+loki.source.heroku "local" {
+  listener {
+    address  = "127.0.0.1"
+    port = 8080
+  }
+  use_incoming_timestamp = true
+  labels   = { component = "loki.source.heroku"} 
+  forward_to = [loki.write.local.receiver]
```
tpaschalis, line 127: Let's format these like `riverfmt` would do.

The easiest way would be to copy these over to a file and use the `agent fmt /path/to/file` command to get the formatted on stdout.

tpaschalis: Also, let's have the example listen on `0.0.0.0`, since that's most likely what users would have to do to be able to receive entries from Heroku.

docs/sources/flow/reference/components/loki.source.heroku.md
tpaschalis: Also, similar to other components, do we want this as a block, or have we considered flattening the structure?

docs/sources/flow/reference/components/loki.source.heroku.md
```diff
@@ -0,0 +1,136 @@
...
+`port`                   | `int`         | The `<port>` to listen to for heroku messages. | | yes
+
+Example Block
+
+```river
+listener {
+  address  = "127.0.0.1"
+  port = 8080
+}
+```
```
tpaschalis, line 80: Nitpick, but we don't usually give out examples for inner blocks, and this is a pretty simple one, so not sure if we should keep it.

docs/sources/flow/reference/components/loki.source.heroku.md
```diff
@@ -0,0 +1,136 @@
...
+
+The `labels` map is applied to every message that the component reads.
+
+The following internal labels are available for relabeling:
+- `__heroku_drain_host`
+- `__heroku_drain_app`
+- `__heroku_drain_proc`
+- `__heroku_drain_log_id`
+
+All url query params will be translated to `__heroku_drain_param_<name>`
```
tpaschalis, line 92: Can we add a note saying that these labels that start with a double underscore are treated as internal and will be discarded unless a relabeling rule is applied to keep them? Similar to how [loki.source.gelf](https://grafana.com/docs/agent/next/flow/reference/components/loki.source.gelf/#:~:text=These%20labels%20are%20stripped%20unless) notes it.

component/loki/source/heroku/heroku.go
```diff
@@ -0,0 +1,198 @@
...
+
+	newArgs := args.(Arguments)
+	c.fanout = newArgs.ForwardTo
+
+	var rcs []*relabel.Config
+	if newArgs.RelabelRules != nil && len(newArgs.RelabelRules.GetAll()) > 0 {
+		rcs = flow_relabel.ComponentToPromRelabelConfigs(newArgs.RelabelRules.GetAll())
+	}
+
+	if configsChanged(c.lc, newArgs.HerokuListener) {
```
tpaschalis, line 140: I think this comparison will always trigger as we're never storing a copy of the new config in `c.lc`.

Moreover, we might also have to check whether the relabeling rules have changed; otherwise they will never be propagated to NewHerokuTarget (same goes for the syslog source component, too).

component/loki/source/heroku/heroku_test.go
```diff
@@ -0,0 +1,126 @@
...
+	flow_relabel "github.com/grafana/agent/component/common/relabel"
+	"github.com/grafana/agent/component/loki/source/heroku/internal/herokutarget"
+	"github.com/grafana/agent/pkg/flow/logging"
+	"github.com/grafana/regexp"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/model"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPush(t *testing.T) {
```
tpaschalis, line 23: Nice and concise test, I like it.

component/loki/source/heroku/internal/herokutarget/herokutarget.go
```diff
@@ -0,0 +1,199 @@
...
+func (h *HerokuTarget) ListenPort() int {
+	return h.config.Server.HTTPListenPort
+}
+
+func (h *HerokuTarget) Endpoint() string {
+	return "/heroku/api/v1/drain"
+}
+
+func (h *HerokuTarget) Ready() bool {
+	return true
```
tpaschalis, line 187: I know we just copied this over from Promtail so don't sweat it, but we could add a TODO to use an atomic boolean and actually report when the server is _really_ ready.

tpaschalis (commented):
Solid work, just a few nits here and there! 🙌

component/loki/source/heroku/heroku_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+const address = "localhost"
+const port = 42421
+const testPayload = `270 <158>1 2022-06-13T14:52:23.622778+00:00 host heroku router - at=info method=GET path="/" host=cryptic-cliffs-27764.herokuapp.com request_id=59da6323-2bc4-4143-8677-cc66ccfb115f fwd="181.167.87.140" dyno=web.1 connect=0ms service=3ms status=200 bytes=6979 protocol=https
+`
+
+var rulesExport = &flow_relabel.Rules{
+	GetAll: func() []*flow_relabel.Config {
+		return []*flow_relabel.Config{
+			{
```
tpaschalis, line 84: After rebasing with main, this will probably have to be just
```
var rulesExport = flow_relabel.Rules{
	{ // a rule
         },
         { // a second rule
         },
... 
}

tpaschalis (approved):
LGTM, thanks! 🎉

PR #2799, opened 2023-01-23
drone: only rebuild containers when Dockerfiles change

rfratto:
Currently, it takes upwards of 40 minutes to run the "Check Windows containers" drone job for every push to a PR.

This change only runs jobs to check containers if the Dockerfile itself has changed, which should reduce the turnaround time on PRs.

Note that these rules come from a [Drone plugin which is being used on our runners](https://github.com/meltwater/drone-convert-pathschanged), so the specific syntax can't be seen in Drone's docs.

### DIFF
--- a/.drone/drone.yml
+++ b/.drone/drone.yml
@@ -189,6 +189,8 @@ steps:
 trigger:
   event:
   - pull_request
+  paths:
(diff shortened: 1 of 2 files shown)


erikbaranowski (approved):
LGTM

PR #2800, opened 2023-01-23
Flow API: Display the values of all River attributes unless they are nil

### LINKED ISSUE #2205: Flow API: empty exports not returned from the API
If a component has a zero value for everything in its exports, the entire exports section is omitted from the API result. 

This might mislead users, since it makes it seem like those components don't have any exported fields.

### LINKED ISSUE #2206: Flow API: empty values hidden from attributes
If a component has a zero value for any attribute, that attribute is omitted from the API result.

This might mislead or confuse users, since it makes it seem like that attribute isn't set (or potentially doesn't exist).

I think all required attributes should always show up in the UI, even if they're the zero value.

ptodev:
#### PR Description
The UI will now display the values of all River attributes unless they are nil.

Example UI screenshots:

<details>
<summary>remote_write exports</summary>

</details>

<details>
<summary>local_file exports</summary>

</details>

<details>
<summary>node_exporter exports</summary>

</details>

<details>
<summary>node_exporter arguments</summary>

</details>

Fixes #2205
Fixes #2206

#### PR Checklist

- [x] CHANGELOG updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/river/encoding/attribute.go
+++ b/pkg/river/encoding/attribute.go
@@ -29,6 +29,10 @@ func (af *attributeField) hasValue() bool {
 
 // MarshalJSON implements json marshaller.
 func (af *attributeField) MarshalJSON() ([]byte, error) {
+	if af.valField == nil {
+		return nil, fmt.Errorf("the value of the attribute field is nil")
+	}
+
 	if af.valField.hasValue() {
 		af.field.Value = af.valField
 	} else {
@@ -43,7 +47,7 @@ func (af *attributeField) convertAttribute(val value.Value, f rivertags.Field) e
 	if !f.IsAttr() {
 		return fmt.Errorf("convertAttribute requires a field that is an attribute got %T", val.Interface())
 	}
-	if !val.Reflect().IsValid() || val.Reflect().IsZero() {
+	if !val.Reflect().IsValid() {
 		return nil
 	}
 	af.Type = attrType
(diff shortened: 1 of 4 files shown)


pkg/river/encoding/attribute.go
```diff
@@ -29,6 +29,10 @@ func (af *attributeField) hasValue() bool {
 
 // MarshalJSON implements json marshaller.
 func (af *attributeField) MarshalJSON() ([]byte, error) {
+	if af.valField == nil {
+		return nil, fmt.Errorf("the value of the attribute field is nil")
+	}
```
rfratto, line 34: I'm not sure I understand intuitively when this condition can hit. Have you run into this?

rfratto: what was the river config which triggered the panic?

ptodev: I think it was the `TestSimpleZeroVal` test from attributes_test.go. I ran it prior to my change, so I can work in a more TDD way. I just tried to run that test again on the main branch, and indeed I get an error:
```
```

rfratto: It feels like this check shouldn't be hitting and it's indicative of a problem elsewhere in the encoder, but we'll take it for now to fix the bug and we can revisit refactoring the encoder when we have more time.

rfratto (approved):

PR #2801, opened 2023-01-23
Flow UI: Fix the issue for object layout

### LINKED ISSUE #2689: UI: Objects no longer indented and aligned
As of either #2683 or #2642, aligning of multiline values stopped working:

Before:

image

After:

image

Is this because extra whitespace is getting removed?

cc @xiyu95

xiyu95:
Fixes #2689 

#### Notes to the Reviewer
I refactored the code a little bit to fix the layout for the table value that contains objects, but I think the `RiverValue.tsx` will need to be recreated or at least a big refactor ahead since at the moment it uses empty spaces to calculate the indet level. This PR has a refactor for the object value to take in css properties instead of using white spaces, but I kept the old way of implementation wrapped in a condition render (This is because there will probably be a different data structure). I will add this note to the one pager UI enhancement doc in the agent drive. Below is the screenshots for one of the component detail pages after the fix:

Screen Shot 2023-01-23 at 10 37 21 AM
Screen Shot 2023-01-23 at 10 37 34 AM

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,8 @@ Main (unreleased)
 
 - Flow: `prometheus.relabel` and `prometheus.remote_write` will now error if they have exited. (@ptodev)
 
+- Flow UI: Fix the layout for the object value in component detail page. (@xiyu95)
+
 ### Other changes
 
 - Use Go 1.19.4 for builds. (@erikbaranowski)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,8 +66,6 @@ Main (unreleased)
 
 - Flow: `prometheus.relabel` and `prometheus.remote_write` will now error if they have exited. (@ptodev)
 
-- Flow UI: Fix the layout for the object value in component detail page. (@xiyu95)
-
 ### Other changes
 
 - Use Go 1.19.4 for builds. (@erikbaranowski)
--- a/web/ui/src/features/river-js/RiverValue.tsx
+++ b/web/ui/src/features/river-js/RiverValue.tsx
@@ -7,47 +7,31 @@ import styles from './RiverValue.module.css';
 export interface RiverValueProps {
   value: Value;
   name?: string;
-  nthChild?: number;
 }
-
-// TO track the length of the object key value pair
-const DIVISION_LENGTH = 20;
 
 /**
  * RiverValue emits a paragraph which represents a River value.
  */
-export const RiverValue: FC<RiverValueProps> = ({ value, name, nthChild }) => {
+export const RiverValue: FC<RiverValueProps> = ({ value, name }) => {
   return (
     <div className={styles.value}>
-      <ValueRenderer value={value} nthChild={nthChild} name={name} />
+      <ValueRenderer value={value} name={name} />
     </div>
   );
 };
 
-const ValueRenderer: FC<RiverValueProps> = ({ value, name, nthChild }) => {
-  const backgroundColor = nthChild && nthChild % 2 === 1 ? '#f4f5f5' : 'white';
-
+const ValueRenderer: FC<RiverValueProps> = ({ value, name }) => {
   /**
    * Renderer for the river format for target and label object
    */
   const renderGrid = (partition: ObjectField[]) => {
-    // const gridTemplateColumns = '10% 1% 69%';
-
-    let maxLength = 0;
-
-    partition.forEach(({ key }) => {
-      return key.length > maxLength ? (maxLength = key.length) : maxLength;
-    });
-
-    const gridTemplateColumns = maxLength < DIVISION_LENGTH ? '10% 1% 89%' : undefined;
-
     return partition.map(({ key, value }) => {
       return (
-        <div key={key} className={styles['grid-layout']} style={{ backgroundColor: backgroundColor, gridTemplateColumns }}>
+        <>
           <div className={`${styles['grid-item']} ${styles['grid-key']}`}>{key}</div>
           <div className={styles['grid-item']}>=</div>
           <ValueRenderer value={value} name={name} />
-        </div>
+        </>
       );
     });
   };
@@ -100,9 +84,11 @@ const ValueRenderer: FC<RiverValueProps> = ({ value, name, nthChild }) => {
         <>
           <span>&#123;</span>
           <br />
-          {partitions.map((partition) => {
-            return renderGrid(partition);
-          })}
+          <div className={styles['grid-layout']}>
+            {partitions.map((partition) => {
+              return renderGrid(partition);
+            })}
+          </div>
           <span>&#125;</span>
         </>
       );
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -66,6 +66,8 @@ Main (unreleased)
 
 - Flow: `prometheus.relabel` and `prometheus.remote_write` will now error if they have exited. (@ptodev)
 
+- Flow UI: Fix the layout for the object value in component detail page. (@xiyu95)
```
rfratto, line 69: Since the issue this is fixing was never part of a release, we wouldn't include it in bug fixes here

web/ui/src/features/river-js/RiverValue.tsx
```diff
@@ -80,19 +115,25 @@ const ValueRenderer: FC<valueRendererProps> = (props) => {
...
+            const gridTemplateColumns = '10% 1% 69%';
+
+            switch (name) {
+              case 'targets':
+                return renderGrid(simplifiedPartition);
+              case 'labels':
+                return renderGrid(simplifiedPartition, gridTemplateColumns);
+              default:
+                return renderDefault(partition, indentLevel, keyLength);
+            }
```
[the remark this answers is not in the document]
xiyu95: > Unfortunately this won't work in the general case. We can't guarantee that the only time a object or array value will be in attributes called `targets` and `labels`.

Thanks for the context. In this case, we can apply just the `renderGrid()` without the condition, and combine the `renderGrid` with `renderDefault`, which by adding the `<ValueRenderer>` in the `renderGrid` to fix the out of bound at the same time not breaking the layout. This way it will still call the recursive and detect the `ValueType` for the value. I will do a small adjustment on this.

After the combination it will still detect the color since we will not be using the `grid-value` color anymore. The only thing is to add a dynamic calculation on the width of the css.

web/ui/src/features/river-js/RiverValue.tsx
```diff
@@ -1,30 +1,56 @@
...
+    });
+
+    const gridTemplateColumns = maxLength < DIVISION_LENGTH ? '10% 1% 89%' : undefined;
+
+    return partition.map(({ key, value }) => {
+      return (
+        <div key={key} className={styles['grid-layout']} style={{ backgroundColor: backgroundColor, gridTemplateColumns }}>
+          <div className={`${styles['grid-item']} ${styles['grid-key']}`}>{key}</div>
+          <div className={styles['grid-item']}>=</div>
+          <ValueRenderer value={value} name={name} />
```
xiyu95, line 49: Quick explanation on the change here vs the last commit. There won't be a condition renderer anymore. In general, all the objects will follow this layout which will call the `ValueRenderer` recursively in order to deal with any other nested `ValueType` (including colors). `partition` passing in without simplified to detect the type of the value. Set a width on grid column depending on the max length of the key within the group (This way the `=` symbol will be lined up otherwise there will be a big gap in the shorter key group such as `labels`).

web/ui/src/features/river-js/RiverValue.tsx
```diff
@@ -1,30 +1,56 @@
...
+
+    let maxLength = 0;
 
-const ValueRenderer: FC<valueRendererProps> = (props) => {
-  const value = props.value;
+    partition.forEach(({ key }) => {
+      return key.length > maxLength ? (maxLength = key.length) : maxLength;
+    });
+
+    const gridTemplateColumns = maxLength < DIVISION_LENGTH ? '10% 1% 89%' : undefined;
```
xiyu95, line 42: Set a width on grid column depending on division length. (This way the `=` symbol will be lined up otherwise there will be a big gap in the shorter key group such as `labels`). Comparison ss below with dynamic/static grid width:
 
Screen Shot 2023-01-23 at 11 59 36 AM
 
 
Screen Shot 2023-01-23 at 11 59 58 AM

Due to the reuse of `<table>`, and the diversity of the key:value pair. We need to adjust the width in order to fit both the longer keys and shorter keys.

xiyu95:
@rfratto I have adjust the changes above and a quick explanation on the differences. Let me know if there is any other concern.

web/ui/src/features/river-js/RiverValue.tsx
```diff
@@ -1,30 +1,56 @@
...
+
+    return partition.map(({ key, value }) => {
+      return (
+        <div key={key} className={styles['grid-layout']} style={{ backgroundColor: backgroundColor, gridTemplateColumns }}>
```
rfratto, line 46: Is there a reason that each of the attributes is put in the DOM as a separate grid?

If you had each attribute as a separate row in the grid, then you could set `grid-template-columns` to `max-content auto auto` and it would fluidly scale as expected, but because each attribute is a different grid, we're forced to do the `10% 1% 89%` check, where even the keys could overflow if they're long enough: 

image

(Notice the overflowing on `__meta_kubernetes_pod_annotationpresent_scheduler_alpha_kubernetes_io_critical_pod`)

rfratto:
Another small thing I noticed: 

If you mouse over an entry in one of the tables now, the background-color changes inconsistently: 

image

web/ui/src/features/river-js/RiverValue.tsx
```diff
@@ -1,30 +1,56 @@
...
+      <ValueRenderer value={value} nthChild={nthChild} name={name} />
+    </div>
   );
 };
 
-type valueRendererProps = RiverValueProps & {
-  indentLevel: number;
-};
+const ValueRenderer: FC<RiverValueProps> = ({ value, name, nthChild }) => {
+  const backgroundColor = nthChild && nthChild % 2 === 1 ? '#f4f5f5' : 'white';
```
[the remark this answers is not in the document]
xiyu95: The reason we are doing this logic here is that the grid layout is within the table element. The reason for the logic to be here is that we want to sync up the color with the table's index, not the `nth-child` from the grid.

web/ui/src/features/river-js/RiverValue.tsx
xiyu95: If we set a `max-content auto auto` then the width will depend on the content length, which will cause 2 things to happen: 
1. the line up will be messed up as the content length will be different
2. the content will not wrap which will cause an overflow

xiyu95: > Is there a reason that each of the attributes is put in the DOM as a separate grid?

Oh I see what you mean now. I didn't realize I was setting the grid layout inside the loop. My bad. Updating it now. Nice catch thanks!

xiyu95:
@rfratto Thanks for the catch! It's a lot cleaner now. 😄 Was able to remove the css selector in the functional component and fix the hover effect. Please double check and let me know if the page is behaving as expected.

xiyu95:
I will close this one as will work on the river string function PR for more in the layout of river format

PR #2808, opened 2023-01-24
[Agent Management] [WIP] Refactor Agent Management for Testing

spartan0x117:
#### PR Description
Refactors Agent Management code to support passing a `remoteConfigProvider` interface that handles fetching and caching, in addition to adding tests for the following cases:

- An invalid _initial_ config results in an error and does not cache.
- An un-marshallable _remote_ config (i.e. invalid yaml) falls back to the cache and does not cache.
- If the remote fetch fails it falls back to the cached config and does not cache again.
- If the remote fetch succeeds and responds with a valid (note: at present `valid` means that `LoadBytes` succeeded so it is valid yaml, not necessarily that the config is a valid agent management config) there is no error and the remote config is cached.

- [ ] CHANGELOG updated
- [ ] Documentation added
- [X] Tests updated

### DIFF
--- a/pkg/config/agentmanagement.go
+++ b/pkg/config/agentmanagement.go
@@ -127,11 +127,6 @@ func getRemoteConfig(expandEnvVars bool, configProvider remoteConfigProvider, lo
 		return configProvider.GetCachedRemoteConfig(expandEnvVars)
 	}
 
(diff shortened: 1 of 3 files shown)


spartan0x117:
Closing in favor of an upcoming PR.

PR #2814, opened 2023-01-24
Decrease the CPU profiling period to avoid cancellations

cyriltovena:
#### PR Description

This PR decrease the profiling CPU period to interval-1second, we've seen some timeout in the agent because of that and we made the same change in Phlare. see https://github.com/grafana/phlare/commit/c5e677293d21424a444a147b217ddb4f7c150523

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated


rfratto (approved):

ptodev:
I guess that this will work correctly if `cfg.ScrapeInterval <= time.Second`?

cyriltovena:
> I guess that this will work correctly if `cfg.ScrapeInterval <= time.Second`?

Can't be below 2*seconds

PR #2817, opened 2023-01-24
Flow: add loki.source.kubernetes component

### LINKED ISSUE #2504: `loki.source.kubernetes`: Retrieves logs directly from the Kubernetes API server.

rfratto:
The `loki.source.kubernetes` component, when given a set of targets from `discovery.kubernetes`, tails logs from containers using the Kubernetes API.

One tailer is launched for each unique container. Tailers will reconnect to Kubernetes with exponential backoff if their connection to the Kubernetes API is closed. Tailers only exit if the target is removed from the input or if the tailer determines that the terminated container will not restart.

Closes #2504.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,24 +31,29 @@ Main (unreleased)
 
 - New Grafana Agent Flow components:
 
-  - `otelcol.receiver.kafka` receives telemetry data from Kafka. (@rfratto)
-  - `phlare.scrape` collects application performance profiles. (@cyriltovena)
-  - `phlare.write` sends application performance profiles to Grafana Phlare. (@cyriltovena)
-  - `otelcol.receiver.zipkin` receives Zipkin-formatted traces. (@rfratto)
-  - `otelcol.receiver.opencensus` receives OpenConsensus-formatted traces or metrics. (@ptodev)
-  - `loki.source.heroku` listens for Heroku messages over TCP a connection and forwards them to other `loki` components. (@erikbaranowski)
-  - `loki.source.windowsevent` reads logs from Windows Event Log. (@mattdurham)
-  - `loki.source.syslog` listens for Syslog messages over TCP and UDP
-    connections and forwards them to other `loki` components. (@tpaschalis)
   - `loki.source.cloudflare` reads logs from Cloudflare's Logpull API and
     forwards them to other `loki` components. (@tpaschalis)
   - `loki.source.gcplog` reads logs from GCP cloud resources using Pub/Sub
     subscriptions and forwards them to other `loki` components. (@tpaschalis)
+  - `loki.source.gelf` listens for Graylog logs. (@mattdurham)
+  - `loki.source.heroku` listens for Heroku messages over TCP a connection and
+    forwards them to other `loki` components. (@erikbaranowski)
+  - `loki.source.kubernetes` collects logs from Kubernetes pods using the
+    Kubernetes API. (@rfratto)
+  - `loki.source.syslog` listens for Syslog messages over TCP and UDP
+    connections and forwards them to other `loki` components. (@tpaschalis)
+  - `loki.source.windowsevent` reads logs from Windows Event Log. (@mattdurham)
   - `otelcol.exporter.loki` forwards OTLP-formatted data to compatible `loki`
     receivers. (@tpaschalis)
-  - `loki.source.gelf` listens for Graylog logs. (@mattdurham)
+  - `otelcol.receiver.kafka` receives telemetry data from Kafka. (@rfratto)
   - `otelcol.receiver.loki` receives Loki logs, converts them to the OTLP log
     format and forwards them to other `otelcol` components. (@tpaschalis)
+  - `otelcol.receiver.opencensus` receives OpenConsensus-formatted traces or
+    metrics. (@ptodev)
+  - `otelcol.receiver.zipkin` receives Zipkin-formatted traces. (@rfratto)
+  - `phlare.scrape` collects application performance profiles. (@cyriltovena)
+  - `phlare.write` sends application performance profiles to Grafana Phlare.
+    (@cyriltovena)
 
 - Flow components which work with relabeling rules (`discovery.relabel`,
   `prometheus.relabel` and `loki.relabel`) now export a new value named Rules.
--- a/component/loki/source/kubernetes/kubernetes.go
+++ b/component/loki/source/kubernetes/kubernetes.go
@@ -1,0 +1,311 @@
+// Package kubernetes implements the loki.source.kubernetes component.
+package kubernetes
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"reflect"
+	"sync"
+	"time"
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	"github.com/grafana/agent/component"
+	"github.com/grafana/agent/component/common/config"
+	"github.com/grafana/agent/component/common/loki"
+	"github.com/grafana/agent/component/common/loki/positions"
+	"github.com/grafana/agent/component/discovery"
+	"github.com/grafana/agent/component/loki/source/kubernetes/kubetail"
+	"github.com/grafana/agent/pkg/build"
+	"github.com/grafana/agent/pkg/river"
+	promconfig "github.com/prometheus/common/config"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+func init() {
+	component.Register(component.Registration{
+		Name: "loki.source.kubernetes",
+		Args: Arguments{},
+
+		Build: func(opts component.Options, args component.Arguments) (component.Component, error) {
+			return New(opts, args.(Arguments))
+		},
+	})
+}
+
+// Arguments holds values which are used to configure the loki.source.kubernetes
+// component.
+type Arguments struct {
+	Targets   []discovery.Target  `river:"targets,attr"`
+	ForwardTo []loki.LogsReceiver `river:"forward_to,attr"`
+
+	// Client settings to connect to Kubernetes.
+	Client ClientArguments `river:"client,block,optional"`
+}
+
+var _ river.Unmarshaler = (*Arguments)(nil)
+
+// DefaultArguments holds default settings for loki.source.kubernetes.
+var DefaultArguments = Arguments{
+	Client: ClientArguments{
+		HTTPClientConfig: config.DefaultHTTPClientConfig,
+	},
+}
+
+// UnmarshalRiver implements river.Unmarshaler and applies defaults.
+func (args *Arguments) UnmarshalRiver(f func(interface{}) error) error {
+	*args = DefaultArguments
+
+	type arguments Arguments
+	return f((*arguments)(args))
+}
+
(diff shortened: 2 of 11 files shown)


component/loki/source/kubernetes/kubetail/target.go
```diff
@@ -0,0 +1,238 @@
...
+	"sync"
+	"time"
+
+	"github.com/cespare/xxhash/v2"
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/model/labels"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// TODO(rfratto): health, log time, and offset in target info.
```
rfratto, line 15: Stale comment; all of this is done.

component/loki/source/kubernetes/kubetail/tailer.go
```diff
@@ -0,0 +1,323 @@
...
+	for bo.Ongoing() {
+		err := t.tail(ctx, handler)
+		if err == nil {
+			terminated, err := t.containerTerminated(ctx)
+			if terminated {
+				// The container shut down and won't come back; we can stop tailing it.
+				return
+			} else if err != nil {
+				level.Warn(t.log).Log("msg", "could not determine if container terminated; will retry tailing", "err", err)
+			}
```
rfratto, line 111: I originally didn't have this logic, but had to implement it once I noticed that it kept retrying forever for init containers that stopped (and would never come back).

I briefly considered whether retrying forever was fine if the same log lines don't get sent more than once, but I'm concerned about how much extra pressure it would cause on the API server for clusters that make heavy use of init containers.

component/loki/source/kubernetes/kubernetes.go
```diff
@@ -0,0 +1,311 @@
...
+
+	// Convert input targets into targets to give to tailer.
+	targets := make([]*kubetail.Target, 0, len(newArgs.Targets))
+
+	for _, inTarget := range newArgs.Targets {
+		lset := inTarget.Labels()
+		processed, err := kubetail.PrepareLabels(lset, c.opts.ID)
+		if err != nil {
+			// TODO(rfratto): should this set the health of the component?
+			level.Error(c.log).Log("msg", "failed to process input target", "target", lset.String(), "err", err)
```
tpaschalis, line 243: Should we `continue` and skip over the failed target?

component/loki/source/kubernetes/kubetail/kubetail.go
```diff
@@ -0,0 +1,144 @@
...
+
+	m.tasks = tasks
+	return nil
+}
+
+func entryForTarget(t *Target) positions.Entry {
+	// Prefix the path with "cursor-" so it doesn't get automatically deleted by
+	// positions.Positions, which otherwise will try to read the path as a file.
+	return positions.Entry{
+		Path:   "cursor-" + t.String(),
```
tpaschalis, line 96: nitpick: we could also use positions.CursorKey() here for the same effect.

component/loki/source/kubernetes/kubetail/target.go
```diff
@@ -0,0 +1,236 @@
...
+	"github.com/prometheus/prometheus/model/labels"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// Internal labels which indicate what container to tail logs from.
+const (
+	LabelPodNamespace     = "__pod_namespace__"
+	LabelPodName          = "__pod_name__"
+	LabelPodContainerName = "__pod_container_name__"
+	LabelPodUID           = "__pod_uid__"
```
tpaschalis, line 20: I might be missing something, but are these labels coming from somewhere, or a new fallback convention we'd like to enforce?

component/loki/source/kubernetes/kubetail/kubetail.go
```diff
@@ -0,0 +1,144 @@
...
+func (m *Manager) UpdateOptions(ctx context.Context, newOptions *Options) error {
+	m.mut.Lock()
+	defer m.mut.Unlock()
+
+	// Iterate through the previous set of tasks and create a new task with the
+	// new set of options.
+	tasks := make([]*tailerTask, 0, len(m.tasks))
+	for _, oldTask := range m.tasks {
+		tasks = append(tasks, &tailerTask{
+			Options: newOptions,
```
tpaschalis, line 113: Should we be updating the same task with new Options (basically a new Client)?

It sounds like a rare edge-case, but I'm thinking that would mean that a pod with the same `namespace/name:container` in another cluster could potentially reuse an older positions entry from another cluster with a different `SinceTime` and this gap might lead to losing logs.

component/loki/source/kubernetes/kubetail/tailer.go
```diff
@@ -0,0 +1,323 @@
...
+					Line:      entryLine,
+				},
+			}
+
+			select {
+			case <-ctx.Done():
+				return nil
+			case ch <- entry:
+				// Save position after it's been sent over the channel.
+				t.opts.Positions.Put(positionsEnt.Path, positionsEnt.Labels, entryTimestamp.UnixMicro())
```
tpaschalis, line 197: UnixMicro() by default returns the local time, should we store it in UTC?

I couldn't find a definitive answer as to whether Kubernetes clusters can set their timezones. This might be a footgun for someone who chooses to run their pods on different timezones, but I think we should standardize around the common usecase.

FWIW [RFC3339](https://www.rfc-editor.org/rfc/rfc3339#section-4.1) does state that "This specification does not cater to local time zone rules" encouraging UTC, but it _does_ allow offsets.

component/loki/source/kubernetes/kubetail/tailer.go
```diff
@@ -0,0 +1,323 @@
...
+		// * It terminated with any exit code and restartPolicy is Always
+		// * It terminated with non-zero exit code and restartPolicy is not Never
+		//
+		// https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#restart-policy
+		switch {
+		case containerInfo.State.Waiting != nil || containerInfo.State.Running != nil:
+			return false, nil // Container will restart
+		case restartPolicy == corev1.RestartPolicyAlways:
+			return false, nil // Container will restart
+		case containerInfo.State.Terminated != nil && restartPolicy == corev1.RestartPolicyOnFailure && containerInfo.State.Terminated.ExitCode != 0:
```
tpaschalis, line 259: Do we want to check for `restartPolicy == OnFailure` or `restartPolicy != Never` like the comment says?

component/loki/source/kubernetes/kubetail/tailer.go
```diff
@@ -0,0 +1,323 @@
...
+			return false, nil // Container will restart
+		default:
+			return true, nil // Container will *not* restart
+		}
+
+	case containerTypeEphemeral:
+		// Ephemeral containers never restart.
+		//
+		// https://kubernetes.io/docs/concepts/workloads/pods/ephemeral-containers/
+		return true, nil
```
tpaschalis, line 287: Does this mean we'll never tail ephemeral containers though? Should we at least check if they're running?

rfratto: Yeah, good point (though this line is only hit after we got an EOF, so it would at least be tailed for a little while)

CHANGELOG.md
```diff
@@ -31,21 +31,25 @@ Main (unreleased)
...
   - `otelcol.exporter.loki` forwards OTLP-formatted data to compatible `loki`
     receivers. (@tpaschalis)
-  - `loki.source.gelf` listens for Graylog logs. (@mattdurham)
+  - `otelcol.receiver.kafka` receives telemetry data from Kafka. (@rfratto)
+  - `otelcol.receiver.opencensus` receives OpenConsensus-formatted traces or
+    metrics. (@ptodev)
+  - `otelcol.receiver.zipkin` receives Zipkin-formatted traces. (@rfratto)
+  - `phlare.scrape` collects application performance profiles. (@cyriltovena)
+  - `phlare.write` sends application performance profiles to Grafana Phlare.
+    (@cyriltovena)
```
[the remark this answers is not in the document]
tpaschalis: I've merged another component so this might need another go.

component/loki/source/kubernetes/kubetail/target.go
rfratto: Yeah, I'd like it to be the preferred convention since the `__meta_kubernetes` labels could be considered optional. We'll fall back to the meta labels if these don't exist though.

component/loki/source/kubernetes/kubetail/kubetail.go
rfratto: Hm, I'm not sure, having the tasks be immutable is pretty nice to reason about. 

I'm not sure I'm following the edge case; are you talking about if someone changes the component to watch an entirely different cluster?

tpaschalis: Yes, that's the case I was thinking. It does sound pretty rare, but that's what would happen, right?

component/loki/source/kubernetes/kubetail/tailer.go
rfratto: Are you sure it returns the local time?

```
// UnixMicro returns t as a Unix time, the number of microseconds elapsed since
// January 1, 1970 UTC. The result is undefined if the Unix time in
// microseconds cannot be represented by an int64 (a date before year -290307 or
// after year 294246). The result does not depend on the location associated
// with t.
```

I can change this to `entryTimestamp.UTC().UnixMicro()` just to be safe though.

rfratto: Actually, I think we should store it in whatever format we get back from Kubernetes. Just it case it *does* change, we don't want to store a different time in the positions file and then load that different time back in for the next tail (since at that point, we wouldn't know how to convert it back to the local time Kubernetes recognizes)

tpaschalis: Whoops, I was looking at the docs for the UnixMicro function, not the method https://pkg.go.dev/time#UnixMicro 🤷

tpaschalis (approved):
Looks nice!

rfratto:
@karengermond I'd like to merge this so we can keep moving forward for the release tomorrow, but I'll fix any fixes you suggest in a follow up PR :)

docs/sources/flow/reference/components/loki.source.kubernetes.md
```diff
@@ -0,0 +1,146 @@
...
+
+[client]: #client-block
+[http_client_config]: #http_client_config-block
+
+### client block
+
+The `client` block configures the Kubernetes client used to tail logs from
+containers. If the `client` block isn't provided, default in-cluster
+configuration with the service account of the running Grafana Agent pod is
+used.
```
rfratto, line 86:
```suggestion
The `client` block configures the Kubernetes client used to tail logs from
containers. If the `client` block isn't provided, the default in-cluster
configuration with the service account of the running Grafana Agent pod is
used.
```

docs/sources/flow/reference/components/loki.source.kubernetes.md
```diff
@@ -0,0 +1,146 @@
...
+
+The following arguments are supported:
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`api_server` | `string` | URL of Kubernetes API server. | | no
```
rfratto, line 92:
```suggestion
`api_server` | `string` | URL of the Kubernetes API server. | | no
```

PR #2819, opened 2023-01-25
[Agent Management] Add Remote Config Validation

spartan0x117:
#### PR Description
Refactors `agentmanagement.go` so `getRemoteConfig` accepts `remoteConfigProvider`s, which allows for more testing. Adds tests for some cases.

Also splits out the code that parses the flagset (the second time) and sets integrations version in `pkg/config/config.go`, specifically `load`, into a separate function which `getRemoteConfig` also calls. `getRemoteConfig` is also modified to take the parameters needed by this function (the `flagSet`, `args`, and the config `path`).

This is done in order to allow for semantic validation of the remote config inside the agent management code. Otherwise, if the API responds with a semantically invalid config (i.e. it's valid yaml, but it's not a valid agent management config) the _invalid_ remote config will be cached and any agents attempting to use it upon startup will panic.

Finally, this adds a new metric `agent_remote_config_invalid_total`, which tracks the number of invalid configs that were returned by the API, with possible labels `invalid_yaml` and `invalid_agent_config`.

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [X] Tests updated

### DIFF
--- a/pkg/config/agentmanagement.go
+++ b/pkg/config/agentmanagement.go
@@ -18,11 +18,8 @@ import (
 const cacheFilename = "remote-config-cache.yaml"
 
 type remoteConfigProvider interface {
-	ValidateInitialConfig() error
-
 	GetCachedRemoteConfig(expandEnvVars bool) (*Config, error)
 	CacheRemoteConfig(remoteConfigBytes []byte) error
-
 	FetchRemoteConfig() ([]byte, error)
 }
 
@@ -30,14 +27,14 @@ type remoteConfigHTTPProvider struct {
 	InitialConfig *AgentManagementConfig
 }
 
-func newRemoteConfigHTTPProvider(c *Config) remoteConfigHTTPProvider {
-	return remoteConfigHTTPProvider{
-		InitialConfig: &c.AgentManagement,
+func newRemoteConfigHTTPProvider(c *Config) (*remoteConfigHTTPProvider, error) {
+	err := c.AgentManagement.Validate()
+	if err != nil {
+		return nil, err
 	}
-}
-
-func (r remoteConfigHTTPProvider) ValidateInitialConfig() error {
-	return r.InitialConfig.Validate()
+	return &remoteConfigHTTPProvider{
+		InitialConfig: &c.AgentManagement,
+	}, nil
 }
 
 // GetCachedRemoteConfig retrieves the cached remote config from the location specified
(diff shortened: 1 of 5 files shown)


pkg/config/agentmanagement.go
```diff
@@ -116,6 +91,78 @@ func fetchConfig(c *Config) ([]byte, error) {
...
+	err = LoadBytes(remoteConfigBytes, expandEnvVars, &remoteConfig)
+	if err != nil {
+		level.Error(log).Log("msg", "could not load the response from the API, falling back to cache", "err", err)
+		instrumentation.InstrumentInvalidRemoteConfig("invalid_yaml")
+		return configProvider.GetCachedRemoteConfig(expandEnvVars)
+	}
+
+	// this is done in order to validate the config semantically
+	if err = applyIntegrationValuesFromFlagset(fs, args, configPath, &remoteConfig); err != nil {
+		level.Error(log).Log("msg", "could not apply integration version to config", "err", err)
```
jcreixell, line 137: probably instrument here too with reason "invalid_flags" (arguably this should never happen but who knows)

pkg/config/agentmanagement.go
```diff
@@ -116,6 +91,78 @@ func fetchConfig(c *Config) ([]byte, error) {
...
+
+	err = LoadBytes(remoteConfigBytes, expandEnvVars, &remoteConfig)
+	if err != nil {
+		level.Error(log).Log("msg", "could not load the response from the API, falling back to cache", "err", err)
+		instrumentation.InstrumentInvalidRemoteConfig("invalid_yaml")
+		return configProvider.GetCachedRemoteConfig(expandEnvVars)
+	}
+
+	// this is done in order to validate the config semantically
+	if err = applyIntegrationValuesFromFlagset(fs, args, configPath, &remoteConfig); err != nil {
```
jcreixell, line 136: for a follow up refactoring PR: some of this logic is so entangled with config.go that it should probably moved there (these two files know and depend on each other, which isn't great from the modularity POV)

pkg/config/agentmanagement.go
```diff
@@ -116,6 +91,78 @@ func fetchConfig(c *Config) ([]byte, error) {
...
 }
 
+type labelMap map[string]string
+
+type RemoteConfiguration struct {
+	Labels    labelMap `yaml:"labels"`
+	Namespace string   `yaml:"namespace"`
+}
+
+type AgentManagement struct {
```
jcreixell, line 101: consider renaming this to AgentManagementConfig``

jcreixell (commented):
LGTM, I am gonna do a few tests locally and if everything looks good I will give it a :+1:

jcreixell (approved):
Good job, I have tested this thoroughly and seems to work as expected.

I found an unrelated issue that we will need to fix separately (adding, removing and adding again log scrape configs seems to cause an issue with registering prometheus metrics in promtail, I will create a separate issue for that)

pkg/config/agentmanagement_test.go
```diff
@@ -87,27 +124,14 @@ func TestMissingCacheLocation(t *testing.T) {
 	assert.Error(t, invalidConfig.Validate())
 }
 
-func TestGetCachedRemoteConfig(t *testing.T) {
-	cwd := filepath.Clean("./testdata/")
-	_, err := getCachedRemoteConfig(cwd, false)
-	assert.NoError(t, err)
-}
+// func TestGetCachedRemoteConfig(t *testing.T) {
```
marctc, line 127: We could probably delete this code

jcreixell:
DNM, adding some changes

jcreixell:
I am done with my changes, @marctc feel free to merge if you are happy with this

marctc (approved):

PR #2822, opened 2023-01-25
Gracefully restart scrape jobs for `phlare.scrape`

cyriltovena:
#### PR Description

This PR ensures we don't pull the plug when we're scraping and gracefully restart scrape goroutine when a configuration changes.

I also test if there's a need to restart scrape goroutines to avoid doing it unnecessarily

#### Notes to the Reviewer

Deployed and test

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated


rfratto (approved):
LGTM! 

By the way, we recently merged #2788 which introduces a new package to abstract some of this stuff away. It helps make it easier to make sure workers (like scrapers) don't get updated when nothing changes. We might want to consider updating `phlare.scrape` at some point to use that package.

PR #2826, opened 2023-01-25
loki.source.podlogs: new Flow component

### LINKED ISSUE #2763: Flow: create `loki.source.podlogs` component to collect logs based on PodLogs CRDs

rfratto:
This commit introduces a new Flow component called `loki.source.podlogs`. This component will look for PodLogs resources on Kubernetes, which in turn specifies a set of Pods to collect logs from.

`loki.source.podlogs` will then use the Kubernetes API to collect logs from the discovered Pods.

The PodLogs CRD used by the component is a new version (v1alpha2) of the one use used by Grafana Agent Operator (v1alpha1). There are some minor differences in this version:

* `jobLabel` and `podTargetLabels` are removed in favor of using `relabelings` to transfer labels from the Kubernetes Pod onto the target.

* `pipelineStages` has been removed in favor of passing the result of `loki.source.podlogs` to a `loki.process` component. `pipelineStages` can be added back in the future once all stages are made available.

* The `namespaceSelector` uses label selectors instead of the namespace selector object from Prometheus Operator. Prometheus Operator uses label selectors for namespaces wherever it can, but can't use them in PodMonitors due to restrictions in Prometheus SD (where labels from namespaces aren't exposed for relabeling rules). We don't have the same restrictions here, so we use label selectors instead. 

`loki.source.podlogs` does *not* support the v1alpha1 resouces and will ignore them if they are on the Kubernetes cluster.

A major tradeoff to this component is that it currently has no notion of sharding. There is currently no way for two agents to distribute the load of `loki.source.podlogs` by using something like hashmod. If multiple agents are running the `loki.source.podlogs` component, they must do manual sharding based on labels of PodLogs resources. 

To reduce the size of this PR, the new CRD is not yet made available in the agent Helm chart. This can be done in a follow up, along with creating the permissions for the agent to discover PodLogs.

Closes #2763.

## Running locally

If you wish to try this locally, do the following:

1. Run `go generate ./component/loki/source/podlogs/...` to build the CRDs. 
2. Apply the CRDs in `operations/helm/charts/grafana-agent/crds` to your Kubernetes cluster (ideally one created with `k3d` or `kind`) 
3. Run the following Flow config, replacing `RECEIVER` with a component to send the logs to: 
   ```river 
   logging {
     level = "debug"
   }

   loki.source.podlogs "example" {
     forward_to = [RECEIVER]

     client {
       kubeconfig_file = env("HOME") + "/.kube/config"
     }
   }
   ```
4. Finally, apply this object to Kubernetes to bring up a container which generates logs and a PodLogs to collect from it:

  <details>
<summary>Kubernetes manifest</summary>

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: logger
  namespace: default
  labels:
    app: logger
spec:
  replicas: 1
  selector:
    matchLabels:
      app: logger
  template:
    metadata:
      labels:
        app: logger
    spec:
      containers:
      - name: logger
        image: chentex/random-logger:latest

---

apiVersion: monitoring.grafana.com/v1alpha2
kind: PodLogs
metadata:
  name: logger
  namespace: default
  labels:
    app: logger
spec:
  selector:
    matchLabels:
      app: logger
  namespaceSelector: {}
  relabelings:
  - action: replace
    sourceLabels:
    - __meta_kubernetes_namespace
    targetLabel: namespace
  - action: replace
    sourceLabels:
    targetLabel: container
  - action: replace
    separator: /
    sourceLabels:
    - __meta_kubernetes_namespace
    - __meta_kubernetes_pod_name
    targetLabel: instance
  - action: drop
    regex: Succeeded|Failed|Pending
    sourceLabels:
    - __meta_kubernetes_pod_phase
```

  </details>


component/all/all.go
```diff
@@ -15,8 +15,10 @@ import (
 	_ "github.com/grafana/agent/component/loki/source/gelf"                     // Import loki.source.gelf
 	_ "github.com/grafana/agent/component/loki/source/heroku"                   // Import loki.source.heroku
 	_ "github.com/grafana/agent/component/loki/source/kubernetes"               // Import loki.source.kubernetes
+	_ "github.com/grafana/agent/component/loki/source/podlogs"                  // Import loki.source.podlogs
 	_ "github.com/grafana/agent/component/loki/source/syslog"                   // Import loki.source.syslog
 	_ "github.com/grafana/agent/component/loki/source/windowsevent"             // Import loki.source.windowsevent
+	_ "github.com/grafana/agent/component/loki/stdout"                          // Import loki.stdout
```
rfratto, line 21:
```suggestion
```

component/loki/source/podlogs/selector.go
```diff
@@ -0,0 +1,56 @@
...
+		Operator: metav1.LabelSelectorOperator(me.Operator),
+		Values:   me.Values,
+	}
+}
+
+// NamespaceSelector defines a selector to find namespaces.
+type NamespaceSelector struct {
+	All        bool     `river:"all,attr,optional"`
+	MatchNames []string `river:"match_names,attr,optional"`
+}
```
tpaschalis, line 56: Doesn't seem like we use this anywhere, we make use of LabelSelector for both the `selector` and `namespace_selector`.

I'm removing this.

docs/sources/flow/reference/components/loki.source.podlogs.md
```diff
@@ -0,0 +1,238 @@
...
+
+`loki.source.podlogs` supports the following arguments:
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`forward_to` | `list(LogsReceiver)` | List of receivers to send log entries to. | | yes
+
+`loki.source.podlogs` searches for `PodLogs` resources on Kubernetes. Each
+`PodLogs` resource describes a set of pods to collect logs from.
+`loki.source.podlogs` discovers pods based on the `PodLogs` resources, and
```
tpaschalis, line 49: I'm removing a duplicate sentence here.

docs/sources/flow/reference/components/loki.source.podlogs.md
```diff
@@ -0,0 +1,238 @@
...
+client > http_client_config | [http_client_config][] | HTTP client configuration for Kubernetes requests. | no
+selector | [selector][] | Label selector for which `PodLogs` to discover. | no
+selector > match_expression | [match_expression][] | Label selector expression for which `PodLogs` to discover. | no
+namespace_selector | [selector][] | Label selector for which namespaces to discover `PodLogs` in. | no
+namespace_selector > match_expression | [match_expression][] | Label selector expression for which namespaces to discover `PodLogs` in. | no
+
+The `>` symbol indicates deeper levels of nesting. For example, `client >
+http_client_config` refers to an `http_client_config` block defined
+inside a `client` block.
+
```
tpaschalis, line 140: I'm mentioning that users can supply multiple `match_expression` blocks, acting as AND clauses.

component/loki/source/podlogs/podlogs.go
```diff
@@ -0,0 +1,315 @@
...
+}
+
+// updateController updates the state of the controller. This must only be
+// called after updateReconciler. mut must be held when calling.
+func (c *Component) updateController(args Arguments) error {
+	// We only need to update the controller if we already have a rest config
+	// generated and our client args haven't changed since the last call.
+	if reflect.DeepEqual(c.args.Client, args.Client) && c.restConfig != nil {
+		return nil
+	}
```
tpaschalis, line 277: Do we mean if it _has_ changed here?

tpaschalis, line 277: Do we mean if it _has_ changed here? cc @rfratto

component/loki/source/podlogs/internal/apis/monitoring/v1alpha2/types.go
```diff
@@ -0,0 +1,41 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec PodLogsSpec `json:"spec,omitempty"`
+}
+
+// PodLogsSpec defines how to collect logs for a Pod.
+type PodLogsSpec struct {
+	// Selector to select Pod objects. Required.
+	Selector metav1.LabelSelector `json:"selector"`
```
tpaschalis, line 25: The selector docstring says the Selector is required in a PodLogs spec, but we have this block as optional. Or is it a different thing? cc @rfratto

tpaschalis (approved):
LGTM. Still looking through the reconciler/controller code to understand them better, but I think it looks good to go.

component/loki/source/podlogs/selector.go
rfratto: Yep, thanks, I added this in to test something and forgot to take it out.

docs/sources/flow/reference/components/loki.source.podlogs.md
rfratto: Can you add that in the section for match_expression?

component/loki/source/podlogs/internal/apis/monitoring/v1alpha2/types.go
rfratto: Yeah, nice catch, we should add a kubebuilder annotation to say this is required for anyone writing a PodLogs resource (we do this in some of the operator types if you need a reference on how to do that). I don't think the selectors on the Flow configs need to be required though.

component/loki/source/podlogs/controller.go
```diff
@@ -0,0 +1,196 @@
...
+		case <-ctrl.reconcileCh:
+			if err := ctrl.reconciler.Reconcile(ctx, client); err != nil {
+				level.Error(ctrl.log).Log("msg", "reconcile failed", "err", err)
+			}
+		}
+	}
+}
+
+// configureInformers starts the informers used by this controller to perform reconciles.
+func (ctrl *controller) configureInformers(ctx context.Context, informers cache.Informers) error {
```
mattdurham, line 158: Thoughts on passing in the RequestReconsole func and making this an unattached fun?

component/loki/source/podlogs/controller.go
```diff
@@ -0,0 +1,196 @@
...
+	// we can reevaluate selectors defined in PodLogs.
+	types := []client.Object{
+		&corev1.Namespace{},
+		&corev1.Pod{},
+		&monitoringv1alpha2.PodLogs{},
+	}
+	for _, ty := range types {
+		informer, err := informers.GetInformer(ctx, ty)
+		if err != nil {
+			return err
```
mattdurham, line 170: If we error early do we need to do anything with all the `onChangeEventHandler` remember the danging events from .net days.

component/loki/source/podlogs/podlogs.go
```diff
@@ -0,0 +1,315 @@
...
+	_ component.DebugComponent = (*Component)(nil)
+)
+
+// New creates a new loki.source.podlogs component.
+func New(o component.Options, args Arguments) (*Component, error) {
+	err := os.MkdirAll(o.DataPath, 0750)
+	if err != nil && !os.IsExist(err) {
+		return nil, err
+	}
+	positionsFile, err := positions.New(o.Logger, positions.Config{
```
mattdurham, line 98: @tpaschalis for future enhancement thoughts on building the path into the positions.New? So we can avoid the calls to mkdirall sprinkled around.

tpaschalis: Certainly yes!

PR #2829, opened 2023-01-26
Prefer deamonless containers

jdbaldry:
Use podman (https://podman.io) for running containers without a root Docker daemon if it is installed.

Signed-off-by: Jack Baldry <jack.baldry@grafana.com>


rfratto (approved):
feels a little niche, but sure :)

PR #2830, opened 2023-01-26
Add doc-validator for linting documentation

jdbaldry:
#### PR Description

`doc-validator` codifies best practices for technical documentation at Grafana. Linting errors are defined in https://github.com/grafana/technical-documentation/blob/main/tools/cmd/doc-validator/errata.hcl. 

The errors enumerated in the errata file are under active development and are intended to be helpful in explaining reasoning and fixes for linting issues. Not all errors meet this goal yet, so please report any errors that are confusing, seem to lack motivation, or are hard to resolve.

#### Notes to the Reviewer

I also limited the workflow to run on PRs that change documentation files and added manual dispatch. Let me know if this is undesirable and I will drop the commit.


clayton-cornell (commented):
Checked out this branch and tested `make doc-validator`. It works, and it generates a lot of output (errors plus explanation).  I think it does the job as intended.

jdbaldry:
Nice! There's a bunch of improvements since this PR was originally opened. I can configure the workflow to provide GitHub comments and suggestions.

@clayton-cornell I would love to help you get this into the repository. Given that it will cause CI to fail, do we want to enable this one piece at a time, or perhaps we can have a spike to fix up the existing errors in one go.

I can fix a large amount of the errors programmatically but things like `description` front matter might need your help.

jdbaldry:
I've pushed the updated version to a new branch `jdb/2023-07-add-doc-validator` but I won't open a PR because the automated review will post so many comments.

If you are keen to get this in, I'll fix up a bunch of those errors to make the review manageable and then I can open the PR for the final cleanups.

jdbaldry:
Sounds like a good plan to me.

erikbaranowski:
> Thank you for doing this, it'd be great! Over time maybe we could also add Agent-specific linting for docs. For example, there is a "river fmt" utility which could be used for formatting River code in the example section of the docs.

This would be very useful. I think we should make this available for docs linting in all repos if/when it is added since different docs may reference river markdown blocks.

clayton-cornell:
@jdbaldry Where are we with this PR?

jdbaldry:
I've put this PR on the back burner given your comment about bite size chunks.

In order to enable this check in CI we would need to fix all the errors or introduce a plan for how we limit which files are linted and which rules are enabled. All the current rules are documented in https://grafana.com/docs/writers-toolkit/review/doc-validator/errata/.

I'm happy to work around your availability to fix things. Is there a particular directory you want to start with?

Once we have a plan of how we want to enable it, I can bump the version in workflow and then we need to fix all the identified issues.

clayton-cornell:
I've taken a more in-depth look at the output of `make docs-validator` in this repo and I think it'd be helpful to resolve one major contributor to the error output... it'd dramatically reduce the output if this validation was fixed first... I am not sure where this needs to be corrected though.

The following error pattern shows up for every file in the repo:
```
ERROR: sources/latest/_index.md:1:1 The canonical 'https://grafana.com/docs/agent/latest/' in the front matter does not match the source file pretty URL path '/docs/agent/agent/latest/'. [canonical-does-not-match-pretty-URL]
Update the canonical URL to match the pretty URL for the source file path.
```
It looks like something is being parsed incorrectly in the pretty URL path and `agent` is added twice?

clayton-cornell:
Two PRs in Agent to fix up the documentation validation errors: 

* https://github.com/grafana/agent/pull/5296
* https://github.com/grafana/agent/pull/5319

These two PRs fix up all the error output from `make docs-validator` EXCEPT:

1. The `canonical-does-not-match-pretty-URL` errors - this looks like a possible bug in the validation scripts. This will have to be looked at separately.
2. The `anchor does not exist` errors in `operator/api.md`.  This file is generated from source code doc comments. The source looks OK, and the URLs all resolve in the live documentation. So either the validator needs to be made aware of `href` anchor labels, or we have to change the doc comments to create different link syntax.

So, I'm guessing that once these two issues are sorted out, we should be in a position to enable the validator in the Agent repo builds?

PR #2836, opened 2023-01-26
Changing the current version to v0.31.0-rc.0

ptodev:


rfratto:
We don't have a release strategy for updating the Helm charts right now, can we leave the entire `operations/helm` directory untouched?

captncraig:
I'd also prefer not to touch helm here, mostly because of the size of the changeset. But I think you could eval this by pointing at the branch to pull the chart, so it is probably valuable to have an updated chart. But unless there are changes to the chart itself, you could evaluate the rc by changing the image tag in values.

captncraig:
All that to say, I don't think the value outweighs the annoyance for updating the chart in a release branch.

tpaschalis (approved):

PR #2854, opened 2023-01-29
docs: update landing page for Flow

rfratto:
* Remove suggestion that Flow isn't suitable for production. Our testing has shown that it's stable in production, and the only thing people would need to be concerned about is breaking changes. 
  * This suggestion probably should've been removed when Flow graduated to beta, I left it in by mistake. 
* Update list of limitations, which has fallen out of sync since the v0.28 release.

Once this is merged to main, I'll create a second PR to cherry pick it to release-v0.31.


karengermond (approved):
LGTM!

mattdurham (approved):

PR #2855, opened 2023-01-30
[Agent Management] Add agent management-related alerts

spartan0x117:
#### PR Description
Adds some `warning` and `critical` alerts for agent management-related errors.

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/production/grafana-agent-mixin/alerts.libsonnet
+++ b/production/grafana-agent-mixin/alerts.libsonnet
@@ -242,12 +242,12 @@ local _config = config._config;
           {
             alert: 'AgentManagementBadAPIRequests',
             expr: |||
-              100 * sum(rate(agent_remote_config_fetches_total{status_code!~"2.."}[5m])) by (%(group_by_cluster)s)
+              100 * sum(rate(agent_remote_config_fetches_total{status_code="(4|5).."}[10m])) by (%(group_by_cluster)s)
                 /
-              sum(rate(agent_remote_config_fetches_total[5m])) by (%(group_by_cluster)s)
+              sum(rate(agent_remote_config_fetches_total[10m])) by (%(group_by_cluster)s)
                 > 5
             ||| % _config,
-            'for': '10m',
+            'for': '20m',
             labels: {
               severity: 'warning',
             },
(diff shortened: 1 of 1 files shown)


marctc (approved):

production/grafana-agent-mixin/alerts.libsonnet
```diff
@@ -235,6 +235,151 @@ local _config = config._config;
...
           },
         ],
+      },
+      {
+        name: 'GrafanaAgentManagement',
+        rules: [
+          {
+            alert: 'AgentManagementBadAPIRequests',
+            expr: |||
+              100 * sum(rate(agent_remote_config_fetches_total{status_code!~"2.."}[5m])) by (%(group_by_cluster)s)
```
jcreixell, line 245: shall this alert fire if we get `3..` responses?

jcreixell, line 245: I think it is probably safer to use 10m, I could see agents having multi-minute poll intervals to reduce request rates

jcreixell: probably safer to use `"(4|5).."`

jcreixell (approved):
please see my comments, other than that LGTM

production/grafana-agent-mixin/alerts.libsonnet
```diff
@@ -242,12 +242,12 @@ local _config = config._config;
...
             expr: |||
-              100 * sum(rate(agent_remote_config_fetches_total{status_code!~"2.."}[5m])) by (%(group_by_cluster)s)
+              100 * sum(rate(agent_remote_config_fetches_total{status_code="(4|5).."}[10m])) by (%(group_by_cluster)s)
                 /
-              sum(rate(agent_remote_config_fetches_total[5m])) by (%(group_by_cluster)s)
+              sum(rate(agent_remote_config_fetches_total[10m])) by (%(group_by_cluster)s)
                 > 5
             ||| % _config,
-            'for': '10m',
+            'for': '20m',
```
jcreixell, line 250: AFAIK we don't need to change this when increasing the rate interval (it just means it looks back a longer time frame for each data point, but the alert duration, meaning how much we wait before alerting, is not affected)

jcreixell (approved):
Please see my comment, but overall LGTM

PR #2868, opened 2023-01-31
Prefer relative aliases

jdbaldry:
Absolute aliases containing versions can redirect "latest" content to
old versions when a page has been removed in newer versions.
The redirected pages are indexed by search engines but our robots.txt forbids them crawling the non-latest page.

Relative aliases behave consistently across versions and do not suffer the same problems.

I've checked a bunch of aliases manually by:
1. Running `make docs`
1. Picking an absolute alias path from a random file
1. Checking that I end up at the page represented by the file containing the alias.

For example, to test the relative alias in the file `docs/sources/configuration/create-config-file.md`:
1. `$ make docs`
1. Browse to http://localhost:3002/docs/agent/latest/set-up/create-config-file/
1. Confirm that I am redirected to http://localhost:3002/docs/agent/latest/configuration/create-config-file/

### DIFF
--- a/docs/sources/README.md
+++ b/docs/sources/README.md
@@ -1,6 +1,4 @@
 ---
-aliases:
-- /docs/agent/latest/README/
 draft: "True"
 ---
 
--- a/docs/sources/_index.md
+++ b/docs/sources/_index.md
@@ -1,6 +1,4 @@
 ---
-aliases:
-- /docs/agent/latest/
 title: Grafana Agent
 weight: 1
 ---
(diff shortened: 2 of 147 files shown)


jdbaldry:
This will need to be backported to all versions from `v0.25` to `v0.30`

karengermond (approved):
These look good to me.

karengermond:
@rfratto, Jack updated the agent aliases to prevent broken links to different versions -- just wanted to make sure you're aware of this PR and that we're supposed to use relative paths for aliases.

PR #2878, opened 2023-01-31
helm: support Grafana Agent v0.31

rfratto:
Adds support for Grafana Agent v0.31 to the Helm chart:

* Installs the PodLogs CRD used by `loki.source.podlogs`. 
* Adds RBAC rules to allow `loki.source.podlogs` and `mimir.rules.kubernetes` to function. 

I chose to only include CRDs which we own, so the PrometheusRule CRD is not included and will have to be installed separately for mimir.rules.kubernetes to be able to discover resources. 

I tested this with a k3d cluster locally; everything works.

### DIFF
--- a/operations/helm/charts/grafana-agent/templates/rbac.yaml
+++ b/operations/helm/charts/grafana-agent/templates/rbac.yaml
@@ -23,6 +23,33 @@ rules:
       - get
       - list
       - watch
+  # Rules which allow loki.source.kubernetes and loki.source.podlogs to work.
+  - apiGroups:
+      - ""
+    resources:
+      - pods
+      - pods/log
+      - namespaces
+    verbs:
+      - get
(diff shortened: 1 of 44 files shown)


operations/helm/charts/grafana-agent/templates/rbac.yaml
```diff
@@ -23,6 +23,33 @@ rules:
       - get
       - list
       - watch
+  # Rules which allow loki.source.kubernetes and loki.source.podlogs to work.
+  - apiGroups:
+      - ""
+    resources:
+      - pods
+      - pods/log
```
rfratto, line 31: I triple checked this, but it is in fact `pods/log` and not `pod/logs` or `pods/logs`.

captncraig (approved):

PR #2879, opened 2023-01-31
Flow: reload when SIGHUP is sent

rfratto:
Three small changes:

* Flow will now reload when a SIGHUP is sent, matching the behavior of static mode.
* Update the deb and rpms to respond to `systemctl reload grafana-agent.service` by sending a SIHGUP to the agent process.
* Fix a broken River stanza introduced in #2629.


erikbaranowski (approved):
LGTM, using SIGHUP for reload seemed a little strange on first glance but considering that matches the agent functionality, that makes sense.

PR #2881, opened 2023-01-31
drone: test cross compiling all targets

rfratto:


tpaschalis:
Adding `-mlong-calls` in CGO_CFLAGS makes these two ARM-specific steps [pass](https://drone.grafana.net/grafana/agent/6017/6/2) (and other fails, ofc because the flag doesn't exist)

marctc (approved):

tpaschalis:
Are we leaving the cross-compiling as part of the normal CI runs?

erikbaranowski (approved):
LGTM

rfratto:
> Are we leaving the cross-compiling as part of the normal CI runs?

I'd like to. They're quick and it'll be good to catch this in a PR if there's platform/architecture-specific problems.

erikbaranowski:
Agreed, this is a great addition to the test drone pipelines to help catch more stuff before going to main

PR #2891, opened 2023-02-01
tools/agentlint: add linter for river tags

rfratto:
Extend tools/agentlint with a linter for `river` tags, which will help catch improper usages of tags where unit tests for unmarshaling aren't present (first seen in #2629).

Linting rules can be skipped when necessary by adding a `//nolint:rivertags` comment.

### DIFF
--- a/tools/agentlint/internal/rivertags/rivertags.go
+++ b/tools/agentlint/internal/rivertags/rivertags.go
@@ -17,7 +17,7 @@ var Analyzer = &analysis.Analyzer{
 	Run:  run,
 }
 
-var noLintRegex = regexp.MustCompile(`//\s*nolint:rivertags`)
+var noLintRegex = regexp.MustCompile(`//\s*nolint:(\S+)`)
 var riverTagRegex = regexp.MustCompile(`river:"([^"]*)"`)
 
 // Rules for river tag linting:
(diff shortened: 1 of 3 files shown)


tools/agentlint/internal/rivertags/rivertags.go
```diff
@@ -0,0 +1,248 @@
...
+	"golang.org/x/tools/go/analysis"
+)
+
+var Analyzer = &analysis.Analyzer{
+	Name: "rivertags",
+	Doc:  "perform validation checks on River tags",
+	Run:  run,
+}
+
+var noLintRegex = regexp.MustCompile(`//\s*nolint:rivertags`)
```
tpaschalis, line 20: In some places in our codebase we do something like this: `//nolint:golint,rivertags`. Can we amend the regex to catch this as well?

tpaschalis (approved):
LGTM, easy to follow!

tools/agentlint/internal/rivertags/rivertags.go
```diff
@@ -0,0 +1,248 @@
...
+	)
+
+	switch options {
+	case "attr", "attr,optional":
+		if len(nameParts) == 0 {
+			diagnostics = append(diagnostics, "attr field must have a name")
+		} else if len(nameParts) > 1 {
+			diagnostics = append(diagnostics, "attr field names must not contain `.`")
+		}
+		for _, name := range nameParts {
```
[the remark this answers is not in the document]
rfratto: Yeah, I'm iterating over it anyway just in case they incorrectly put more than one, it's still useful (imo) to show the various errors across all the name fragments..

tools/agentlint/internal/rivertags/rivertags.go
rfratto: Sure! Doing it at the regex level felt ugly, but I changed my approach to support this 👍

PR #2893, opened 2023-02-01
web/ui: add riverStringify function

rfratto:
`riverStringify` converts a Value into the canonical River representation with pretty-printing applied.


xiyu95:
I will directly use this pr as a playground and test in there if you don't mind.

rfratto:
> I will directly use this pr as a playground and test in there if you don't mind.

@xiyu95 Go for it!

rfratto:
Superseded by #3101.

PR #2895, opened 2023-02-01
Gonzalez/static as default

gonzalez:
#### PR Description
Updates the values file to deploy in static mode. 
Updates the readme to provide new defaults

- [x] CHANGELOG updated
- [x] Documentation added


rfratto:
Ah, I'm sorry, I misunderstood your offline question. I'm not sure about changing the default of the Helm chart. We're trying to make Flow the primary way to use the agent as fast as we can, and the Helm chart is being used as a way to make the choice to use static mode more explicit. 

I might change my mind later, but for now I think it's useful to have Flow as the default mode. 

cc @captncraig

PR #2896, opened 2023-02-01
Update release instructions

ptodev:
Me and @tpaschalis did a release recently, and noticed that some things we normally do are not in the instructions. I hope this update will make them more complete and more clear.

To see the rendered markdown, please go to [docs/developer/releasing.md](https://github.com/grafana/agent/blob/update-rel-instructions/docs/developer/releasing.md)

### DIFF
--- a/docs/developer/releasing.md
+++ b/docs/developer/releasing.md
@@ -66,7 +66,7 @@ The lifetime of a release branch is shepherded by a single person, picked from a
 rota of project maintainers and contributors. The release shepherd is
 responsible for managing all releases that compose the release branch.
 
-For a **new release branch**, the release shepherd will:
+For a **new release candidate**, the release shepherd will:
 
 1. Gather consensus on which commit should be used as a base for the release
    branch.
@@ -85,7 +85,11 @@ For a **new release branch**, the release shepherd will:
 
 5. Tag the release candidate using the tag naming convention `vX.Y.0-rc.N`.
 
-6. Run the release candidate in a testing environment for at least 48 hours. If
+6. [Publish the release candidate on Github](#publishing-the-release)
+
+7. Announce the release candidate on the community Slack channel.
+
+8. Run the release candidate in a testing environment for at least 48 hours. If
    you do not have a testing environment, one can be spawned locally using the
    sample environments in `example/k3d`.
 
@@ -94,26 +98,34 @@ For a **new release branch**, the release shepherd will:
    fixes are available, cherry-picking the fixes into the release branch and
    starting a new release candidate.
 
-7. Create a new PR to [update code](#updating-code) using the _stable_
-   release version. 
+   The `rc` version of the next release candidate would have to be incremented - 
+   e.g. from `v0.31.0-rc.0` to `v0.31.0-rc.1`.
 
-8. Create the stable release tag.
+If no new issues were found during the environment testing, the shepherd can continue with 
+creating a **stable release**:
 
-9. Make sure [Homebrew is updated](#updating-homebrew).
+1. Create a new PR to [update code](#updating-code) using the _stable_
+   release version (i.e. without `-rc.N`).
 
-10. Force-push the `release` branch to point at the stable release tag. This
+2. Create the stable release tag.
+
+3. Force-push the `release` branch to point at the stable release tag. This
    branch is used to externally reference files in the repository for a stable
    release, e.g. the "[latest](https://grafana.com/docs/agent/latest/)" documentation.
 
-11. Make sure the documentation is updated:
+4. Make sure the documentation is updated:
     1. A new version of the documentation should be visible on the 
     [version switcher](https://grafana.com/docs/versions/?project=/docs/agent/).
     2. The "[latest](https://grafana.com/docs/agent/latest/)" one should be identical to the latest version.
+    3. Check the "Upgrade guide" - the latest changes listed in it should be assigned to the latest version.
 
-12. [Publish the release](#publishing-the-release)
+5. [Publish the release on Github](#publishing-the-release)
 
-13. Post a comment on the community Slack channel to let the rest of the community 
-know about the release. You can post about release candidates too.
+6. [Update Agent Operator Helm chart version](#update-agent-operator-helm-chart-version)
+
+7. Make sure [Homebrew is updated](#updating-homebrew).
+
+8. Announce the release on community Slack channel.
 
 For **patch releases**, the release shepherd will:
 
@@ -126,16 +138,21 @@ For **patch releases**, the release shepherd will:
 
 4. [Publish the release](#publishing-the-release)
 
+5. [Update Agent Operator Helm chart version](#update-agent-operator-helm-chart-version)
+
 ### Updating code
 
 The codebase must be updated whenever a new release is being made to reference the upcoming release tag.
 
+NOTE: Any change done for a release should be done in `main` first and then moved to the release branch.
+
+NOTE: Branches used for PRs should have a name that doesn't start with `release-`, 
+otherwise branch protection rules apply to it. Alternatively, branches used for PRs 
+can come from forks.
+
 #### Update the changelog
 
-NOTE: Any time CHANGELOG.md is updated for a release, it should first be done
-via PR to the release branch, and then by a second PR to main.
-
-When creating a release **candidate** (an `-rc.N` version):
+When creating a release **candidate** (i.e. an `-rc.N` version):
 
 * Add a new section in the changelog for the new release candidate and include 
 today's date. E.g. "v0.31.0-rc.0 (2023-01-26)".
@@ -151,10 +168,17 @@ of the `-rc.N` version.
 * Make sure to also update the **date** in the title.
 * See [here](https://github.com/grafana/agent/pull/2873/files) for an example pull request.
 
+#### Update the "Upgrade guide"
+The "Upgrade guide" is located in `docs/sources/upgrade-guide/_index.md` and lists breaking changes 
+and deprecated features relevant to each release. Make sure that it is updated similarly to the changelog.
+
 #### Search and replace the old version with the new one
 
 Go through the entire repository and find references to the previous release
   version, updating them to reference the new version **where necessary**.
+
+NOTE: Please do not update the `operations/helm` directory. It is updated independently 
+from Agent releases for now.
 
 NOTE: There are files such as 
 "[pkg/operator/defaults.go](https://github.com/grafana/agent/blob/main/pkg/operator/defaults.go)", 
@@ -178,7 +202,12 @@ release process.
 ### Publishing the Release
 
 GitHub Actions will create release assets and open a release draft for every
-pushed tag. To publish the release:
+pushed tag.
+
+> **WARNING**: We should never force push a tag after the publish button is pressed 
+for a release.
+
+To publish the release:
 
 1. Go to the [GitHub releases page](https://github.com/grafana/agent/releases)
    and find the drafted release.
@@ -215,7 +244,12 @@ needs to be manually updated after the release is published. These are the requi
 
 1. Copy the content of the last CRDs into helm-charts: from agent's repo `production/operator/crds/` to
 helm-charts' repo `charts/agent-operator/crds`.
-2. Update references of agent-operator app version in helm-charts pointing to release version and bump helm chart version.
+2. Update references of agent-operator app version in helm-charts pointing to release version and bump 
+helm chart version.
+3. There is no need to update the README.md manually - running the 
+[helm-docs](https://github.com/norwoodj/helm-docs) utility in the `charts/agent-operator` directory 
+will update it automatically.
+
 You can use this [pull-request](https://github.com/grafana/helm-charts/pull/1831) as reference.
 
 ### Updating Homebrew
@@ -229,8 +263,8 @@ Brew installations use by default
 * [grafana/homebrew-grafana](https://github.com/grafana/homebrew-grafana) is a "tap" - 
 a Third-Party Repository from Grafana.
 
-When a release tag is published to the Agent repository, GitHub actions are triggered automatically 
-to update both of the above repositories. PRs are created automatically:
+When a new [release](https://github.com/grafana/agent/releases) is published, GitHub actions 
+are automatically triggered to update both of the above repositories. PRs are created automatically:
 * If there are no issues, they will be merged automatically.
 * The merges are done from these two grafanabot forks:
     * [grafanabot/homebrew-core](https://github.com/grafanabot/homebrew-core)
--- a/docs/developer/releasing.md
+++ b/docs/developer/releasing.md
@@ -140,13 +140,13 @@ For **patch releases**, the release shepherd will:
    
 2. Cherry-pick the fixes from `main` into the release branch.
 
-3. Follow the same instructions as the ones for a stable release.
+3. Follow the same instructions as the ones for a stable release, starting from step 1.
 
 Some notes on patch releases:
 
 - When creating patch releases, there is no need for a release candidate.
 
-- Changes made to patch releases are not listed in the changelog for the next stable version.
+- Changes made to patch releases are not listed in the CHANGELOG for the next stable version.
 
 - The tag of the patch release does yet not exist at the time when we update code for the 
 main branch to reference it.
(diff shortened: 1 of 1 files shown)


docs/developer/releasing.md
```diff
@@ -161,8 +218,31 @@ helm-charts' repo `charts/agent-operator/crds`.
...
 
+### Updating Homebrew
+[Homebrew](https://brew.sh/) is a package manager for MacOS. It installs binaries 
+(aka "[bottles](https://docs.brew.sh/Bottles)") from either the main Homebrew repository 
+or a third-party one (aka a "[tap](https://docs.brew.sh/Taps)").
+
+For every Agent release we need to update two Brew repositories:
+* [homebrew/homebrew-core](https://github.com/Homebrew/homebrew-core) is the main one which 
+Brew installations use by default
+* [grafana/homebrew-grafana](https://github.com/grafana/homebrew-grafana) is a "tap" -
```
rfratto, line 229: Rather than documenting our custom tap, should we follow in Mimir's footsteps and deprecate our own Formula in favor of the official Homebrew core one? See grafana/homebrew-grafana#41.

docs/developer/releasing.md
```diff
@@ -146,10 +187,26 @@ pushed tag. To publish the release:
...
+
+4. Optionally, have other team members review the release draft if you wish
    to feel more comfortable with it.
 
-4. Publish the release!
+5. Publish the release!
+
+NOTE: Release candidates should be retained on the 
+[GitHub releases page](https://github.com/grafana/agent/releases).
+Please do not remove them.
```
rfratto, line 209:
```suggestion
> **NOTE**: Release candidates should be retained on the 
[GitHub releases page](https://github.com/grafana/agent/releases).
Please do not remove them.
```

docs/developer/releasing.md
```diff
@@ -113,25 +124,55 @@ For patch releases, the release shepard will:
...
+Run `make generate-manifests` and `make generate-dashboards` to update
+manifests in case they are stale.
+
 ### Merge freezes
 
-Release shepards may request a merge freeze to main for any reason during the
+Release shepherds may request a merge freeze to main for any reason during the
 release process.
 
 ### Publishing the Release
```
rfratto, line 178: Can we add a warning somewhere to note that you should never force push a tag after the publish button is pressed for a release?

docs/developer/releasing.md
```diff
@@ -113,25 +124,55 @@ For patch releases, the release shepard will:
...
+At the time of this writing, `defaults.go` is the only such file.
+
+#### Update defaults.go
+
+Add the new version to the "[pkg/operator/defaults.go](https://github.com/grafana/agent/blob/main/pkg/operator/defaults.go)" file. If there is a release candidate (`-rc.N`) version, remove it.
+
+#### Update manifests and dashboards
+
+Run `make generate-manifests` and `make generate-dashboards` to update
+manifests in case they are stale.
```
rfratto, line 171: Ideally we remove this from the release instructions and instead run `make generate-manifests generate-dashboards` in CI and fail a PR which causes them to be stale.

docs/developer/releasing.md
```diff
@@ -113,25 +124,55 @@ For patch releases, the release shepard will:
...
+
+When creating a **stable** release:
+
+* Replace the version from the release candidate section to be the stable release instead 
+of the `-rc.N` version.
+* We do not leave release candidates in the changelog once there is a stable version for them.
+* Make sure to also update the **date** in the title.
+* See [here](https://github.com/grafana/agent/pull/2873/files) for an example pull request.
+
+#### Search and replace the old version with the new one
```
rfratto, line 154: Can we add a notice here to not touch the `operations/helm` directory? That should be updated independently from releases for now.

docs/developer/releasing.md
```diff
@@ -113,25 +124,55 @@ For patch releases, the release shepard will:
...
 NOTE: Any time CHANGELOG.md is updated for a release, it should first be done
 via PR to the release branch, and then by a second PR to main.
 
+When creating a release **candidate** (an `-rc.N` version):
+
+* Add a new section in the changelog for the new release candidate and include 
+today's date. E.g. "v0.31.0-rc.0 (2023-01-26)".
+* All items form the "unreleased" section will move to a new section for the upcoming release version.
+* See [here](https://github.com/grafana/agent/pull/2838/files) for an example pull request.
+* Sanity check that the changelog doesn't contain features which don't belong to it. Sometimes features get added to the wrong version in the changelog due to bad merges.
```
rfratto, line 144:
```suggestion
* Ensure that the changelog doesn't contain features which don't belong to it. Sometimes features get added to the wrong version in the changelog due to bad merges.
```

docs/developer/releasing.md
```diff
@@ -113,25 +124,55 @@ For patch releases, the release shepard will:
...
+The codebase must be updated whenever a new release is being made to reference the upcoming release tag.
 
-* Modify `CHANGELOG.md` with a new version number and its release date.
-* Go through the entire repository and find references to the previous release
-  version, updating them to reference the new version.
-* Run `make generate-manifests` and `make generate-dashboards` to update
-  manifests in case they are stale.
+#### Update the changelog
 
 NOTE: Any time CHANGELOG.md is updated for a release, it should first be done
```
rfratto, line 135: We should move this warning and re-word it a little. Any change done for a release should be done in main first and then moved to the release branch.

docs/developer/releasing.md
```diff
@@ -62,26 +62,26 @@ export GPG_TTY=$(tty)
...
 3. Create a PR to cherry-pick additional commits into the release branch as
    needed.
 
-4. Create a PR to [update code](#updating-code) for the upcoming release
-   candidate. A new section in the changelog should be added for the release
-   candidate, documenting all changes introduced by the release candidate.
-
-4. Update the changelog with a new section for the upcoming release candidate,
-   documenting changes that will be introduced in that release candidate.
+4. Create a PR to [update code](#updating-code) using the `-rc.N` release
```
rfratto, line 83: We should note that branches used for PRs should have a name that doesn't start with `release-`, otherwise branch protection rules apply to it. Alternatively, branches used for PRs can come from forks.

docs/developer/releasing.md
```diff
@@ -94,17 +94,28 @@ For a new release branch, the release shepard will:
    fixes are available, cherry-picking the fixes into the release branch and
    starting a new release candidate.
 
-7. Create a PR to update code for the upcoming stable release. The changelog
-   sections for the release candidates should be replaced with a single section
-   for the stable release.
+7. Create a PR to [update code](#updating-code) again - this time using the stable
```
rfratto, line 97:
```suggestion
7. Create a new PR to [update code](#updating-code) using the _stable_
```

docs/developer/releasing.md
```diff
@@ -94,17 +94,28 @@ For a new release branch, the release shepard will:
...
+   release version. 
 
 8. Create the stable release tag.
 
-9. Force-push the `release` branch to point at the stable release tag. This
+9. Make sure [Homebrew is updated](#updating-homebrew).
```
rfratto, line 102: The order here seems wrong. Updating Homebrew can only happen after the release is published.

docs/developer/releasing.md
```diff
@@ -62,26 +62,26 @@ export GPG_TTY=$(tty)
...
-4. Create a PR to [update code](#updating-code) for the upcoming release
-   candidate. A new section in the changelog should be added for the release
-   candidate, documenting all changes introduced by the release candidate.
-
-4. Update the changelog with a new section for the upcoming release candidate,
-   documenting changes that will be introduced in that release candidate.
+4. Create a PR to [update code](#updating-code) using the `-rc.N` release
+   candidate version.
 
 5. Tag the release candidate using the tag naming convention `vX.Y.0-rc.N`.
```
rfratto, line 86: I'm worried that the instructions here might be a little confusing. We should publish the release candidate too, but the instructions don't mention it. 

Maybe there should be a dedicated section for post-publishing steps, where:

For a release candidate:

* Publish the release candidate 
* Announce the release candidate in the public Slack channel 

For a stable release:

* Publish the release candidate 
* Ensure Homebrew gets updated 
* Ensure documentation got published 
* Announce the stable release in the public Slack channel

docs/developer/releasing.md
ptodev: Will do, thanks!

docs/developer/releasing.md
ptodev: Good point! I'll move it to the start of the "Updating code" section, like this:

```
### Updating code

The codebase must be updated whenever a new release is being made to reference the upcoming release tag.

NOTE: Any change done for a release should be done in `main` first and then moved to the release branch.
```

docs/developer/releasing.md
ptodev: Thank you, I'll add it like this:

```
4. Create a PR to [update code](#updating-code) using the `-rc.N` release
   candidate version.

   NOTE: PRs should have a name that doesn't start with `release-`, otherwise branch 
   protection rules apply to it. Alternatively, branches used for PRs can come from forks.
```

docs/developer/releasing.md
ptodev: True, it does get confusing. I was trying to think how to implement the  post-publishing steps suggestion, but couldn't think of a super clean way.

For now I split the entire "candidate" and the "stable" steps into two different lists. I feel like this makes sense, since they are usually done on two different days? That way we start from point 1 on the second day.

docs/developer/releasing.md
```diff
@@ -94,17 +98,36 @@ For a new release branch, the release shepard will:
...
+5. [Publish the release on Github](#publishing-the-release)
+
+6. [Update Agent Operator Helm chart version](#update-agent-operator-helm-chart-version)
+
+7. Make sure [Homebrew is updated](#updating-homebrew).
 
-For patch releases, the release shepard will:
+8. Announce the release on community Slack channel.
+
+For **patch releases**, the release shepherd will:
```
ptodev, line 130: Hi @rfratto, do you think that the steps below are good enough to describe patch releases? If you agree, I could update the md file.

#### For **patch releases**, the release shepherd will:
1. Merge the fixes we need for the patch to `main`.
2. Raise a PR to `main` to do two things:
    1. Add a new section to the changelog for the patch release. **Move** the fixes listed in the changelog via step 1 from the "unreleased" section to the section for the new patch release.
    2. Update code so that `main` is referencing the new patch version in all files.
4. After the PR in step 2 is merged, cherry pick the commit from step 2 into the release branch.
5. Create a new tag for the patch release.

The above instructions have a few implications:
- When creating patch releases, there is no need for a release candidate.
- Changes made to patch releases are not listed in the changelog for the next stable version.
- The tag of the patch release does yet not exist at the time when we update code for the main branch to reference it.

NOTE: When patching an old release branch, we would do some things differently:
- `main` would not be updated to reference the patch version
- The change applied to the old release branch would have to be listed in two sections in the changelog:
        1. The patch release's section
        2. The "unreleased" section

rfratto: These look good 👍

docs/developer/releasing.md
ptodev: The custom tap is still on 31.0, so it's been effectively deprecated...I'll remove it from the instructions for now and in the unlikely event that we are unable to decommission it we could add it back.

docs/developer/releasing.md
ptodev: If I understand you correctly, what these two `make` commands do is the same as what we would do by search and replace of the version through the codebase? Hence we can do search and replace for now as our release process, and use the `make` commands just to double check that we replaced everything correctly?

I am not too sure how to make sure those `make` commands completed successfully. Maybe the best way would be to have the makefile return an error code which causes Drone to fail the CI? Here I am guessing that Drone would fail when one of the `commands` in the docker pipeline returns a non-zero code.

But then, how would the makefile know that the command failed? I ran a `tk show` command just to see what happens when `tk` exits with an error, and the return code was 0. A 0 normally indicates success. Maybe there is a way other than error codes to check for failures?

If you want to, I could move this into a separate issue and could mention that we should remove the steps from the documentation when this is added to the CI?

rfratto: I'm suggesting that we have a CI job to make sure that the manifests are never stale. We do this for Helm chart tests: https://github.com/grafana/agent/blob/9c068ac3993983b1e3160af6e9af117fdda35f64/.github/workflows/helm-test.yml#L34-L41

That way, there would be no reason to include this step in the release step instructions because manifests would not be stale at release time.

ptodev: Ok, thank you, I'll delete the instructions from here and will make sure to update the CI on the other [PR](https://github.com/grafana/agent/pull/2966).

docs/developer/releasing.md
```diff
@@ -62,30 +62,39 @@ export GPG_TTY=$(tty)
...
 
-2. Create and push the release branch from the selected base commit.
+2. Create and push the release branch from the selected base commit:
+
+   * The name of the release branch should be the name of the stable version we intend 
+   to release, e.g. `release-v0.31`.
+   * Release branches do not contain the `-rc.N` release candidate suffix. I.e. there 
+   is no branch called `release-v0.31-rc.0`.
+   * Release branches do not contain the patch version. I.e. there is no branch called 
+   `release-v0.31.0` or `release-v0.31.4`.
```
rfratto, line 81: The Google documentation style guide recommends against using e.g. and i.e. since people tend to mix them up.

```suggestion
   * The name of the release branch should be the name of the stable version we intend 
   to release, such as `release-v0.31`.
   * Release branches do not contain the `-rc.N` release candidate suffix. This means
     there is no branch called `release-v0.31-rc.0`.
   * Release branches do not contain the patch version. This means there is no 
     branch called `release-v0.31.0` or `release-v0.31.4`.
```

docs/developer/releasing.md
```diff
@@ -62,30 +62,39 @@ export GPG_TTY=$(tty)
...
 ## Performing Releases
 
-The lifetime of a release branch is sheparded by a single person, picked from a
-rota of project maintainers and contributors. The release shepard is
+The lifetime of a release branch is shepherded by a single person, picked from a
+rota of project maintainers and contributors. The release shepherd is
 responsible for managing all releases that compose the release branch.
 
-For a new release branch, the release shepard will:
+For a **new release candidate**, the release shepherd will:
```
rfratto, line 69: Personally, I preferred the older instructions, because it talked about a release branch as a unit: a stream of releases for the same major/minor version tuple. 

I'm concerned saying "for a new release candidate" may be confusing, because the same minor release branch may contain any number of release candidates, depending on how many bugs are discovered. 

Maybe we need a new term here to refer to these branches? "Release streams?"

docs/developer/releasing.md
```diff
@@ -62,30 +62,39 @@ export GPG_TTY=$(tty)
...
 
-4. Create a PR to [update code](#updating-code) for the upcoming release
-   candidate. A new section in the changelog should be added for the release
-   candidate, documenting all changes introduced by the release candidate.
+4. Create a PR to [update code](#updating-code) on `main` using the `-rc.N` release
+   candidate version.
+
+5. Create a PR to cherry pick the update from step 4 to the release branch.
+
+5. [Create a tag](#create-a-tag).
```
rfratto, line 91: Your numbering here drifts (there's two 5s).

docs/developer/releasing.md
```diff
@@ -94,50 +103,139 @@ For a new release branch, the release shepard will:
...
+NOTE: Branches used for PRs should have a name that doesn't start with `release-`, 
+otherwise branch protection rules apply to it. Alternatively, branches used for PRs 
+can come from forks.
+
+#### Update the CHANGELOG
+
+When creating a release **candidate** (i.e. an `-rc.N` version):
+
+* Add a new section in the CHANGELOG for the new release candidate and include 
+today's date. E.g. "v0.31.0-rc.0 (2023-01-26)".
```
rfratto, line 170:
```suggestion
the date of the release, for example, "v0.31.0-rc.0 (2023-01-26)".
```

rfratto (commented):
This is a large PR which adds a lot of text to the documentation. I need to carefully review it again. I'm worried the instructions are becoming too lengthy, and if there's a way we can cut it down while delivering the same information.

docs/developer/releasing.md
ptodev: My reasoning for splitting this into sections of "new release candidate", "stable release" and "patch releases" is that people will likely do them at three separate times. When someone finishes with the "new release candidate" section they are supposed to pause for a few days and then resume from the next section on "stable release". It makes it a little easier to find where to resume, instead of looking for the correct point in a list.

If we could make the headings something like "Step 1: New release candidate", "Step 2: Stable release" and "Step 3: Patch releases", maybe that'd make it more clear that those are different steps of the same continuous process?

In my case, I suppose the "unit" is the Github release and the outcome of each of the three sections is a "release" as in a "Github release". I agree there is value in thinking of the release branch itself as a unit though. Maybe instead of "Performing Releases", this whole section should be called "Release branch lifecycle"?

marctc (approved):

rfratto:
Let's merge for now, we can always improve iteratively.

Issue #2783, opened 2023-01-19
Proposal: add function closures and map/reduce/filter to River


rfratto:
Early on, one of the benefits we planned to introduce into Grafana Agent Flow was more expressiveness. One of the ways I imagined this happening was by allowing River expressions to mutate values passed as input to a component. 

In particular, I'm interested in using River expressions to mutate targets from `discovery.` components and inject new labels as an alternative to relabeling rules: 

```river 
discovery.kubernetes "pods" {
  role = "pod"
}

prometheus.scrape "default" {
  targets = map(
    // (function closure syntax here just for example) 
    // object_merge is a hypothetical function which can combine two objects.
    func(target) => object_merge(target, { 
      extra_label = "test",
    }),
    discovery.kubernetes.pods.targets,
  )

  forward_to = [...]
}
```

### Function closures 

I propose that function closures be added to River. Function closures create a function value, which captures the scope of the block surrounding it (typically a component). 

I suggest the following syntax: 

```
func(ARGUMENT_LIST) => EXPRESSION 
``` 

This syntax was chosen to make it as easy as possible to parse. While `(ARGUMENT_LIST) => EXPRESSION` would be less effort to write, it will make the parser more difficult to maintain. 

In my prototype, I originally tried `func(ARGUMENT_LIST) EXPRESSION` for the function closure syntax. I found that this syntax was a little confusing as it wasn't clear from scanning the file how  `func(ARGUMENT_LIST)` and `EXPRESSION` necessarily related. Adding the arrow `=>` in makes it slightly clearer to new readers what is going on. 

Function closures can be called within River using the normal function call syntax. An example of defining and calling the identity function is as follows:

```river 
(func(id) => id)(15) // returns 15 
```

For the scope of this proposal, it will not be possible to convert a function closure to a component argument. This ability can be added in the future. 

### Map/reduce/filter 

I propose a basic set of functions that accept closures as input: 

* `map(map_func, array)` iterates over elements in the array and calls `map_func(element)` for each element. A new array composed of the result of calling `map_func` for each element is returned.

* `reduce(reduce_func, array, init)` reduces an array to a single value, starting with `init` for the accumulator. `reduce_func(element, accumulator)` is called for each element with the current accumulator value. `reduce_func` should return a new accumulator value. `reduce` returns the final return value (i.e., the final accumulator) provided by calling `reduce_func`.   

* `filter(filter_func, array)` returns a new array composed of elements from `array` where `filter_func(element)` returned true. 

[...]

Issue #2560, opened 2022-11-29
[Request] Supporting multiple `.river` files for Grafana Agent Flow


payparain:
Currently Grafana Agent Flow can only accept a single file for configuration of the Agent when first started (i.e. `agent run <file>`). It would be great if either River could support importing other `.river` files into a single entrypoint, or if Agent could accept multiple `.river` files on start up. 

The use case for this is when using Agent Flow in a multi-user organisation, different teams or individuals work on disparate pieces of a single deployment, and it becomes unwieldy to develop against a single file for defining how to scrape all metrics for a deployment. By supporting the ability to use multiple `.river` files, these teams can each maintain their own `.river` configuration that can individually be provided to the Agent on a deployment, where each team might point their metrics to a common exporter block. This lowers the impact of a single team potentially having a bad config, as it only takes out their stream of data and not the whole graph.

rfratto:
We've been thinking of a way to support including other River files to support code reuse (e.g., something similar to Terraform modules). It feels like this could fit into those plans a little bit (cc @mattdurham). 

One of the things we have to decide is what the behavior of loading multiple River files is: 

* Do all the files get loaded into the same graph? (If so, label names have to be globally unique across all loaded files)
* Do all the files get loaded into separate graphs? (If so, components across different files can't interact with one another)  

It sounds like you're suggesting the first approach, right?

mattdurham:
My solely personal opinion is that in general other files are loaded as subgraphs with dedicated arguments and exports to interact with the parent graph. This is something the team has talked about loosely and I plan on doing a more formal RFC soon(tm). @payparain if you have some time would love to get more details about your specific use case and to see how modules (name subject to change) can help with those. 

There is also a side use case of dynamic components that are a bit further down the line, but I think it is a different use case than the one your are suggesting.

payparain:
> It sounds like you're suggesting the first approach, right?

Hey @rfratto, correct, when writing this I was imagining extra files forming subgraphs that eventually all link into one large graph, which would mandate unique labels globally. The alternative isn't as suitable for what I was hoping as the ability to support multiple files writing to the same `prometheus.remote_write` was where this idea stemmed from, but I haven't thought too deeply about it.

> if you have some time would love to get more details about your specific use case and to see how modules (name subject to change) can help with those.

[...]

PR #2898, opened 2023-02-01
RFC: Design doc for flow `modules`

mattdurham:
#### PR Description

RFC for discussion of modules in flow.

### DIFF
--- a/docs/rfcs/0007-flow-modules.md
+++ b/docs/rfcs/0007-flow-modules.md
@@ -53,6 +53,8 @@ Modules cannot directly access children or parent modules except through predefi
 
 *Note* should security be a non-goal? For example can a submodule spawn a `local.file`, should there be the ability to pass a deny list of modules, and/or allow list? 
 
+*Note* should metadata be a non-goal? Metadata would allow better filtering the ability to tag modules, also will allow programmatic access to things that might otherwise be in comments. 
+
 ### Add additional capabilities to load strings
 
 Modules will not care about the source of the string that represents the river syntax, nor will modules have any inherent reload semantics. The component supplying the string will be responsible for the source and will notify the module when the input changes so that it can utilize the new river configuration.
--- a/docs/rfcs/0007-flow-modules.md
+++ b/docs/rfcs/0007-flow-modules.md
@@ -22,8 +22,7 @@ During this time the Agent team saw a lot of potential in modules outside of scr
 
 ## Goals
 
-* Support single module loading via `module.single`
-* Support multiple module loading via `module.multiple`
+* Support single module loading via `module.string`
 * Enable re-use of common patterns
 * Allow loading a module from a string
 * Allow modules to load other modules
@@ -35,7 +34,7 @@ Common functionality can be wrapped in a set of common components that form a mo
 
 ### Allow loading a module from a string
 
-Modules will not care about the source of a string. In the case of a `module.single` the module will take in one string of valid river configuration. In the case of `module.multiple` the module will take in a `map(string)` where the key uniquely denotes the module.
+Modules will not care about the source of a string. In the case of a `module.string` the module will take in one string of valid river configuration.
 
 ### Allow modules to load other modules
 
@@ -47,13 +46,15 @@ Modules cannot directly access children or parent modules except through predefi
 
 ## Non Goals
 
+Non goals represent capabilities that are not going to be done in the initial release of modules but may come in later versions.  
+
 * Add additional capabilities to load strings
 * Any type of versioning
 * Any user interface work beyond ensuring it works as the UI currently does
+* Any sort of security for modules and what modules are allowed
+* Any sort of metadata
 
-*Note* should security be a non-goal? For example can a submodule spawn a `local.file`, should there be the ability to pass a deny list of modules, and/or allow list? 
 
-*Note* should metadata be a non-goal? Metadata would allow better filtering the ability to tag modules, also will allow programmatic access to things that might otherwise be in comments. 
 
 ### Add additional capabilities to load strings
 
@@ -100,7 +101,7 @@ local.file "mysql" {
     filename = "/test/mysql.river"
 }
 
-module.single "mysql" {
+module.string "mysql" {
     content = local.file.mysql.content
     arguments = {
         {
@@ -111,24 +112,24 @@ module.single "mysql" {
 }
 
 prometheus.scrape "scraper" {
-    targets = module.single.mysql.exports.targets
+    targets = module.string.mysql.exports.targets
 }
 
 ```
 
 ## Limitations
 
-* Duplicate modules cannot be nested, this may or may not be enforced by the system
+* A module cannot directly or indirectly load itself, this will not be enforced by the system
 * Singleton components are not supported at this time. Example [node_exporter](https://grafana.com/docs/agent/latest/flow/reference/components/prometheus.integration.node_exporter/).
 * Modules will not prevent competing resources, such as starting a server on the same port
 * [Configuration blocks](https://grafana.com/docs/agent/latest/flow/reference/config-blocks/#configuration-blocks) will not be supported. 
-* Arguments and exports within a module must be unique
+* Names of arguments and exports within a module must be unique across that module.
 
 ## Proposal
 
 Add the ability to load `modules` as subgraphs to the primary `graph`. Modules may call other modules within a reasonable stack size depth. Modules are represented as a river string that is interpreted with a defined set of arguments and exports.
 
-The initial component will be `module.single` that will load a single module. Internally these modules will be namespaced so they cannot affect children or parent graphs except via arguments and exports.
+The initial component will be `module.string` that will load a single module. Internally these modules will be namespaced so they cannot affect children or parent graphs except via arguments and exports.
 
 Modules will have access to any standard function and any other component exempting singletons. Internally each component in the module will have an `id` that is prepended with the parent's `id` for identification purposes outside of the module. Within the module a component can reference another sibling component normally. There are no known limits on the datatype that a module can use as an argument or export.
 
@@ -137,7 +138,7 @@ Modules will have access to any standard function and any other component exempt
 ### Component Options
 
 
-Given the above example, the `id` of `integrations.mysql "server1"` would be `module.single.mysql.integrations.mysql.server1`. The `data-agent` field would also be prefixed. There are some inherent issues, deeply nested metrics are likely to run into Prometheus label value limits. On Windows platforms there could be issues with the `data-agent` length. These are issues that currently exist in Agent Flow but are more easily hit using deeply nested modules.
+Given the above example, the `id` of `integrations.mysql "server1"` would be `module.string.mysql.integrations.mysql.server1`. The `data-agent` field would also be prefixed. There are some inherent issues, deeply nested metrics are likely to run into Prometheus label value limits. On Windows platforms there could be issues with the `data-agent` length. These are issues that currently exist in Agent Flow but are more easily hit using deeply nested modules.
 
 
 ### Failure Modes
@@ -170,72 +171,7 @@ For example, `Module A` has two sub-modules `Module B` and `Module C`. During re
 * Can create undefined behavior
 * Complex to unload and reload
 
-## `modules.multiple`: Allowing multiple modules to be loaded at once
-
-Note: This feels the most experimental of the topics listed.
-
-Exports are accessed via `module.multiple.LABEL.exports.NAME`, `exports` is a `map(array)`. The `exports` label is used to prevent any collision if other fields are added to `modules.multiple`
-
-## Option 1: No filter 
-
-Depend on filtering the `map(string)` input before loading the modules.
-
-
-#### Filter prior
-
-Assume all files in `/configs/*river` are appropriate for the `module.multiple.load`. This has the advantage of being simplifying the usage of modules and development. This also mirrors `module.single` in the being simple and pushing verisoning/usage to other components. Downside is users are unable to put all configs in one folder and magically work.
-
-```river
-
-// Note this doesnt exist and should only be used for representative purposes.
-local.files "loadfolder" {
-    folder_path = "/configs" # Assume this outputs a map(string)
-    filter = "*.river"
-}
-
-module.multiple "load" {
-    source = local.files.loadfolder.contents
-}
-
-prometheus.scrape "module" {
-    // The module.multiple coalesces multiple exports into an array. 
-    targets = module.multiple.load.exports.targets
-}
-
-```
-
-## Option 2: Filter
-
-Allow `filter_arguments` and `filter_exports` to only include modules that define arguments and exports in the filters. Modules may have additional arguments that can optionally be set.
-
-
-
-#### Filter by arguments and exports
-
-Pass all files to both `load` and `load2` filtering if they have the appropriate input. This can lead to duplicate loading of modules if a module defines both `input1` and `input2`.
-
-```river
-
-// Note this doesnt exist and should only be used for representative purposes.
-local.files "loadfolder" {
-    folder_path = "/configs" # Assume this outputs a map(string)
-    filter = "*.river"
-}
-
-module.multiple "load" {
-    source = local.files.loadfolder.contents
-    filter = ["input1"]
-}
-
-module.multiple "load2" {
-    source = local.files.loadfolder.contents
-    filter = ["input2"]
-}
-
-```
-
-
-# Example Documentation for `argument`
+# Possible Example Documentation for `argument`
 
 ## Arguments
 
@@ -275,15 +211,16 @@ Name | Type | Description
 `value` | `any` | The represented value of the export.
 
 
-# Example Documentation for `module.single`
+# Example Documentation for `module.string`
 
 ## Arguments
 
 The following arguments are supported:
 
-Name            | Type                | Description                                                                                | Default | Required
---------------- | ------------------- | ------------------------------------------------------------------------------------------ |---------| --------
+Name            | Type            | Description                                                                                                                   | Default | Required
+--------------- | --------------- |-------------------------------------------------------------------------------------------------------------------------------|---------| --------
 `arguments`  | `map(string)` | Map of items to pass to module. It is possible to include arguments that are not needed. Any required arguments are required. |    "'{}'"     | no
+`content`  | `string` | River configuration to be loaded.                                                                                             |    "''"     | yes
 
 ## Exported fields
 
--- a/docs/rfcs/0007-flow-modules.md
+++ b/docs/rfcs/0007-flow-modules.md
@@ -5,6 +5,8 @@
 * PR: [grafana/agent#2898](https://github.com/grafana/agent/pull/2898)
 * Status: Draft
 
+[Formatted Link for ease of user](https://github.com/grafana/agent/blob/rfc_modules/docs/rfcs/0007-flow-modules.md)
+
 ## RFC Goals
 
 * Explain the use cases of modules
--- a/docs/rfcs/0007-flow-modules.md
+++ b/docs/rfcs/0007-flow-modules.md
@@ -1,0 +1,168 @@
+# Flow Modules
+
+* Date: 2023-01-27
+* Author: Matt Durham @mattdurham
+* PR: [grafana/agent#XXXX](https://github.com/grafana/agent/pull/XXXX)
+* Status: Draft
+
+## RFC Goals
+
+* Explain the use cases of modules
+* Explain what modules are
+* Go over possible syntax for modules
+* Go over pros and cons of modules
+
+## Summary
+
+One of the primary goals for the production usage of Agent Flow is parity with the static subsystem. One of the features of the static subsystem is [scraping service](). Scraping service allows a user to run a fleet of agents and have thousands of scrape configurations distributed among running Agents. Through discussions within the Agent Team, we did not want to limit dynamically loading content to only scrape configs but allow most components to be loaded and used.
+
+During this time the Agent team saw a lot of potential in modules outside of scraping service. Packaging up sets of components for specific workflows, publishing common use cases and allowing better usage for internal customers in the Agent as a Service model.
+
+## Goals
+
+* Enable re-use of common patterns
+* Allow loading a module from a string
+* Allow modules to load other modules
+* Sandbox modules except via arguments and exports
+* Allow multiple modules with the same interface to be loaded at once
+
+## Non Goals
+
+* Add additional capabilities to load strings
+* Any type of versioning
+* Any User Interface work
+
+## Example
+
+```river
+# module
+argument "password" {
+    optional = false
+    comment = "password for mysql"
+}
+
+argument "username" {
+    optional = false
+    comment = "username for mysql"
+}
+
+export "target" {
+    comment = "target for the integration"
+    value = integrations.mysql.targets
+}
+
+integrations.mysql "mysql" {
+    username = argument.username.value
+    password = argument.password.value
+}
+```
+
+```river
+# parent
+
+local.file "mysql" {
+    filename = "/test/mysql.river"
+}
+
+module.single "mysql" {
+    content = local.file.mysql.content
+    arguments = {
+        {
+            "password" : PASSWORD,
+            "username" : USERNAME,
+        }
+    }
+}
+
+prometheus.scrape "scraper" {
+    targets = module.single.mysql.target
+}
+
+```
+
+## Limitations
+
+* Duplicate modules cannot be nested, this may or may not be enforced by the system
+* Singleton components are not supported at this time
+* Modules will not prevent competing resources, such as starting a server on the same port
+* Component-like objects will not be supported, ie logging level
+* Arguments and exports within a module must be unique
+
+## Proposal
+
+Add the ability to load `modules` as subgraphs to the primary `graph`. Modules may call other modules with a reasonable stack size. Modules are represented as a river string that is interpreted with a defined set of arguments and exports.
+
+The initial component will be `module.single` that will load a single module. Internally these modules will be namespaced so they cannot affect children or parent graphs except via arguments and exports.
+
+Modules will have access to any standard function and any other component exempting singletons. Internally each component in the module will have an `id` that is prepended with the parent's `id` for identification purposes outside of the module. Within the module a component can reference another sibling component normally. There are no known limits on the datatype that a module can use as an argument or export.
+
+### Component Options
+
+
+Given the above example, the `id` of `integrations.mysql "mysql"` would be `module.single.mysql.integrations.mysql.mysql`. The `data-agent` field would also be prefixed. There are some inherent issues, deeply nested metrics are likely to run into prometheus label value limits. On windows platforms there could be issues with the `data-agent` length. These are issues that currently exist in Agent Flow but are more easily hit using deeply nested modules.
+
+
+### Failure Modes
+
+#### When a Module Fails Itself and Children Stops
+
+If an error occurs while re-evaluating a module then the module marks itself as unhealthy and unloads the original module.
+
+*Pros*
+
+* Simple to implement
+* Easy to understand
+
+*Cons*
+
+* One failure mode can cascade
+
+#### Modules Keep Last Good Value
+
+If an error occurs while re-evaluating a module then the module marks itself as unhealthy and attempts to keep the original module. This may have an issue with cascading failures, if a module depends on a module then the system may enter an inconsistent state while applying and then rolling back the change.
+
+For example, `Module A` has two sub-modules `Module B` and `Module C`. During reevaluation `Module B` reloads appropriately but `Module C` fails. `Module A` unloads both modules and then reloads the last good string. In the case that the last good string also fails then `Module A` is unhealthy and non-functional and `Module A's` submodules do not exist.
+
+*Pros*
+
+* Allows more resilient usage
+
+*Cons*
+
+* Can create undefined behavior
+* Complex to unload and reload
+
+## Allowing multiple modules to be loaded at once
+
+Note: This feels the most experimental of the topics listed.
+
+### module.multiple
+
+`module.multiple` will load a module if it matches the required_arguments and required_exports. The input for `source` is a `map(string)`. Note this has a problem of lack of type checking.
+
+Exports are a `map` with the key being the export name and the value of a concatenated array of individual module exports. Exports are accessed via `module.multiple.LABEL.exports.NAME`, `exports` is a `map(array)`.
+
+#### Example
+
+When using the modules for a scraping service there should be the capability to load all modules at once if they have the same arguments and exports.
+
+```river
+
+# Note this doesnt exist and should only be used for representative purposes.
+local.files "loadfolder" {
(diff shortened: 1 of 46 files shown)


rfratto:
Exciting! I'm going to give this a careful review tomorrow.

Related to #2560.

docs/rfcs/0007-flow-modules.md
```diff
@@ -0,0 +1,293 @@
...
+One of the primary goals for the production usage of Agent Flow is parity with the static subsystem. One of the features of the static subsystem is [scraping service](https://github.com/grafana/agent/blob/main/docs/sources/configuration/scraping-service.md). Scraping service allows a user to run a fleet of agents and have thousands of scrape configurations distributed among running Agents. Through discussions within the Agent Team, we did not want to limit dynamically loading content to only scrape configs but allow most components to be loaded and used.
+
+During this time the Agent team saw a lot of potential in modules outside of scraping service. Packaging up sets of components for specific workflows, publishing common use cases and allowing better usage for internal customers in the Agent as a Service model.
+
+## Goals
+
+* Support single module loading via `module.single`
+* Support multiple module loading via `module.multiple`
+* Enable re-use of common patterns
+* Allow loading a module from a string
```
mattdurham, line 28: Wanting to already shift this, I think it should be a label set with key ‘contents’ with the content and ‘id’ for an optional unique id. This will allow metadata to be passed since I also want to add a metadata component into the module.

docs/rfcs/0007-flow-modules.md
```diff
@@ -0,0 +1,293 @@
...
+}
+
+```
+
+## Limitations
+
+* Duplicate modules cannot be nested, this may or may not be enforced by the system
+* Singleton components are not supported at this time. Example node_exporter.
+* Modules will not prevent competing resources, such as starting a server on the same port
+* Component-like objects will not be supported. Example direct access to logging level
```
rfratto, line 122: I did a quick first pass, and this is the only part where I'm not sure what you're taking about here. Could you explain?

mattdurham: I should have used the term `Configuration blocks` but was trying to be vague and failed ;)

docs/rfcs/0007-flow-modules.md
mattdurham: Torn over this for an ease of use mechanism. Maybe `module.single` loads from a string and `module.multiple` loads from a labelset.

docs/rfcs/0007-flow-modules.md
```diff
@@ -0,0 +1,295 @@
...
+
+## Non Goals
+
+* Add additional capabilities to load strings
+* Any type of versioning
+* Any user interface work beyond ensuring it works as the UI currently does
+
+*Note* should security be a non-goal? For example can a submodule spawn a `local.file`, should there be the ability to pass a deny list of modules, and/or allow list? 
+
+*Note* should metadata be a non-goal? Metadata would allow better filtering the ability to tag modules, also will allow programmatic access to things that might otherwise be in comments.
```
mattdurham, line 56: I would like to see a component that wraps `map(string)` to allow for arbitrary key-value pairs that are then exposed somehow, but I don't think we have to determine that in this RFC. The use case would be being able to extract information from modules, and the ability to filter if we export modules as label sets. The metadata becomes a label `metadata_NAME = VALUE`.

docs/rfcs/0007-flow-modules.md
```diff
@@ -0,0 +1,295 @@
...
+
+Modules cannot directly access children or parent modules except through predefined arguments and exports. 
+
+## Non Goals
+
+* Add additional capabilities to load strings
+* Any type of versioning
+* Any user interface work beyond ensuring it works as the UI currently does
+
+*Note* should security be a non-goal? For example can a submodule spawn a `local.file`, should there be the ability to pass a deny list of modules, and/or allow list?
```
mattdurham, line 54: I think this is needed long term but not for this RFC and maybe not for Modules v1. I think there is nuance in that we cannot JUST filter by component name but by component functionality. IE don't allow network, don't allow file system.

docs/rfcs/0007-flow-modules.md
rfratto: Haha, ah, I see. Yeah, for the sake of reaching a wider audience with the RFC, using the terminology we use in the docs is probably the best bet to get everyone on the same page (so they can look up with a configuration block is if they don't already know)

docs/rfcs/0007-flow-modules.md
```diff
@@ -0,0 +1,295 @@
...
+Exports are accessed via `module.multiple.LABEL.exports.NAME`, `exports` is a `map(array)`. The `exports` label is used to prevent any collision if other fields are added to `modules.multiple`
+
+## Option 1: No filter 
+
+Depend on filtering the `map(string)` input before loading the modules.
+
+
+#### Filter prior
+
+Assume all files in `/configs/*river` are appropriate for the `module.multiple.load`. This has the advantage of being simplifying the usage of modules and development. This also mirrors `module.single` in the being simple and pushing verisoning/usage to other components. Downside is users are unable to put all configs in one folder and magically work.
```
erikbaranowski, line 186: versioning*

docs/rfcs/0007-flow-modules.md
```diff
@@ -0,0 +1,295 @@
...
+
+# Example Documentation for `argument`
+
+## Arguments
+
+The following arguments are supported:
+
+Name            | Type                | Description                                                                                | Default | Required
+--------------- | ------------------- | ------------------------------------------------------------------------------------------ |---------| --------
+`optional`  | `bool` | If an argument has to be specified. |    "false"     | no
```
erikbaranowski, line 246: I think we talked about this 1 and using the false as default which makes sense. If we decide it makes sense for the default of `optional` to be true we could always rename this to `required` default false.

erikbaranowski (approved):
Looks great! We talked previous and you already made updates from our discussion, so a bunch was covered there.

docs/rfcs/0007-flow-modules.md
```diff
@@ -0,0 +1,293 @@
...
+* Allows more resilient usage
+
+*Cons*
+
+* Can create undefined behavior
+* Complex to unload and reload
+
+## `modules.multiple`: Allowing multiple modules to be loaded at once
+
+Note: This feels the most experimental of the topics listed.
```
tpaschalis, line 175: If in the future we'd implement #2783 to have a language-level filter functionality, would you envision being able to selectively load modules using that?

Also, would we need modules.multiple then, or could they be loaded in sequence using a map function?

docs/rfcs/0007-flow-modules.md
```diff
@@ -0,0 +1,295 @@
...
+* Explain the use cases of modules
+* Explain what modules are
+* Go over possible syntax for modules
+* Go over pros and cons of modules
+
+## Summary
+
+One of the primary goals for the production usage of Agent Flow is parity with the static subsystem. One of the features of the static subsystem is [scraping service](https://github.com/grafana/agent/blob/main/docs/sources/configuration/scraping-service.md). Scraping service allows a user to run a fleet of agents and have thousands of scrape configurations distributed among running Agents. Through discussions within the Agent Team, we did not want to limit dynamically loading content to only scrape configs but allow most components to be loaded and used.
+
+During this time the Agent team saw a lot of potential in modules outside of scraping service. Packaging up sets of components for specific workflows, publishing common use cases and allowing better usage for internal customers in the Agent as a Service model.
```
rfratto, line 21: I don't want to do too much feedback on formatting for my first review pass, but I'd like to see this be more clear that you're creating a new term and defining it on the next sentence:

```suggestion
During this time the Agent team saw a lot of potential in the idea of "modules." Modules package up sets of components for specific workflows, publishing common use cases and allowing better usage for internal customers in the Agent as a Service model.
```

docs/rfcs/0007-flow-modules.md
```diff
@@ -0,0 +1,295 @@
...
+
+### Allow modules to load other modules
+
+Modules will be able to load other modules, with reasonable safe guards. There will be a stack limit for the depth of sub-modules and modules will be unable to load themselves.
+
+### Modules should be sandboxed except via arguments and exports
+
+Modules cannot directly access children or parent modules except through predefined arguments and exports. 
+
+## Non Goals
```
rfratto, line 48: Can we add a sentence explaining if the non-goals are simply out of scope and may be handled later, or if they're explicitly incompatible with the way we're designing modules?

docs/rfcs/0007-flow-modules.md
```diff
@@ -0,0 +1,295 @@
...
+
+* Add additional capabilities to load strings
```
rfratto, line 50: "load strings" here feels ambiguous. I'm assuming you're talking about module sources? 

I'd like to suggest we introduce three new terms for this RFC to make things clearer:

* Module source: a Flow config file which will be used as a module 
* Main module: the Flow config file passed into `agent run MAIN_MODULE`.  
* Module loader: a component which takes a module source and runs the components defined inside.

docs/rfcs/0007-flow-modules.md
```diff
@@ -0,0 +1,295 @@
...
+```river
+// module
+argument "password" {
+    optional = false
+    comment = "password for mysql"
+}
+
+argument "username" {
+    optional = false
+    comment = "username for mysql"
```
rfratto, line 81: We parse River comments into the AST, so having a comment attribute isn't necessary. We could do this:

```river 
// Username for MySQL.
argument "username" {
  optional = false
}
```

docs/rfcs/0007-flow-modules.md
```diff
@@ -0,0 +1,295 @@
...
+
+prometheus.scrape "scraper" {
+    targets = module.single.mysql.exports.targets
+}
+
+```
+
+## Limitations
+
+* Duplicate modules cannot be nested, this may or may not be enforced by the system
```
rfratto, line 121: Can we make a decision for this RFC whether module loaders prevent this?

Also, I'm not 100% sure what you mean by "duplicate modules cannot be nested." Do you mean that a "module cannot directly or indirectly load itself?" That may be clearer.

docs/rfcs/0007-flow-modules.md
```diff
@@ -0,0 +1,295 @@
...
+* Singleton components are not supported at this time. Example [node_exporter](https://grafana.com/docs/agent/latest/flow/reference/components/prometheus.integration.node_exporter/).
+* Modules will not prevent competing resources, such as starting a server on the same port
+* [Configuration blocks](https://grafana.com/docs/agent/latest/flow/reference/config-blocks/#configuration-blocks) will not be supported. 
+* Arguments and exports within a module must be unique
```
rfratto, line 125:
```suggestion
* Names of arguments and exports within a module must be unique across that module.
```

docs/rfcs/0007-flow-modules.md
```diff
@@ -0,0 +1,295 @@
...
+
+## Proposal
+
+Add the ability to load `modules` as subgraphs to the primary `graph`. Modules may call other modules within a reasonable stack size depth. Modules are represented as a river string that is interpreted with a defined set of arguments and exports.
+
+The initial component will be `module.single` that will load a single module. Internally these modules will be namespaced so they cannot affect children or parent graphs except via arguments and exports.
```
rfratto, line 131: I'm not sure about the name, but I don't think we need to settle on a name for the RFC.

docs/rfcs/0007-flow-modules.md
```diff
@@ -0,0 +1,295 @@
...
+* Any type of versioning
+* Any user interface work beyond ensuring it works as the UI currently does
+
+*Note* should security be a non-goal? For example can a submodule spawn a `local.file`, should there be the ability to pass a deny list of modules, and/or allow list? 
+
+*Note* should metadata be a non-goal? Metadata would allow better filtering the ability to tag modules, also will allow programmatic access to things that might otherwise be in comments. 
+
+### Add additional capabilities to load strings
+
+Modules will not care about the source of the string that represents the river syntax, nor will modules have any inherent reload semantics. The component supplying the string will be responsible for the source and will notify the module when the input changes so that it can utilize the new river configuration.
```
rfratto, line 60: Are we saying the initial module loader will only support strings and that we may add more module loaders in the future (i.e., `module.git`), or if we will explicitly only ever permit getting a module source from another component?

docs/rfcs/0007-flow-modules.md
```diff
@@ -0,0 +1,295 @@
...
+
+Modules will have access to any standard function and any other component exempting singletons. Internally each component in the module will have an `id` that is prepended with the parent's `id` for identification purposes outside of the module. Within the module a component can reference another sibling component normally. There are no known limits on the datatype that a module can use as an argument or export.
+
+`modules.multiple` will use the key of the `map(string)` to uniquely identify a specific module.
+
+### Component Options
+
+
+Given the above example, the `id` of `integrations.mysql "server1"` would be `module.single.mysql.integrations.mysql.server1`. The `data-agent` field would also be prefixed. There are some inherent issues, deeply nested metrics are likely to run into Prometheus label value limits. On Windows platforms there could be issues with the `data-agent` length. These are issues that currently exist in Agent Flow but are more easily hit using deeply nested modules.
```
rfratto, line 140: I'm not sure about using `.` to concatenate the ID of the module loader with the ID of the component, since it may be confused with referencing an export from the module loader. 

Can we use another character like `/` or `:`? (`/` would be better for filesystems though)

docs/rfcs/0007-flow-modules.md
```diff
@@ -0,0 +1,295 @@
...
+*Pros*
+
+* Simple to implement
+* Easy to understand
+
+*Cons*
+
+* One failure mode can cascade
+
+#### Option 2: Modules keep last good value
```
rfratto, line 158: I'm in favor of this. This is how the Flow main file works today. Failing the module with its children sounds potentially disastrous.

docs/rfcs/0007-flow-modules.md
```diff
@@ -0,0 +1,295 @@
...
+*Pros*
+
+* Allows more resilient usage
+
+*Cons*
+
+* Can create undefined behavior
+* Complex to unload and reload
+
+## `modules.multiple`: Allowing multiple modules to be loaded at once
```
rfratto, line 173: It feels like there can be a ton of discussion around this component alone. What do you think about leaving it out of the scope of the modules RFC and just starting with one way to load modules?

docs/rfcs/0007-flow-modules.md
```diff
@@ -0,0 +1,295 @@
...
+
+Note: This feels the most experimental of the topics listed.
+
+Exports are accessed via `module.multiple.LABEL.exports.NAME`, `exports` is a `map(array)`. The `exports` label is used to prevent any collision if other fields are added to `modules.multiple`
```
rfratto, line 177: It feels like we're missing information here for what exactly `module.multiple` would do.

docs/rfcs/0007-flow-modules.md
```diff
@@ -0,0 +1,295 @@
...
+}
+
+prometheus.scrape "module" {
+    // The module.multiple coalesces multiple exports into an array. 
+    targets = module.multiple.load.exports.targets
+}
+
+```
+
+## Option 2: Filter
```
rfratto, line 207: A third option would be to say that the set of module sources gets combined into a single module, such that arguments, exports, and component names must be globally unique. 

I'm not sure I understand what the value in having multiple distinct module sources being loaded as multiple modules from one component is yet, so it feels to me like combining the sources into a single module is equivalent in terms of convenience.

docs/rfcs/0007-flow-modules.md
```diff
@@ -0,0 +1,295 @@
...
+
+module.multiple "load2" {
+    source = local.files.loadfolder.contents
+    filter = ["input2"]
+}
+
+```
+
+
+# Example Documentation for `argument`
```
rfratto, line 238: Is this really example documentation? Or are you proposing what the `argument` block supports? It wasn't clear to me that this was relating back to the argument block schema until I saw that the comment attribute was defined here.

rfratto (commented):
For a first pass, looks good. It feels like we can maybe pare down the scope a bit, since `module.multiple` feels like its own separate can of worms. 

I'll do a review pass for editing/readability once you're ready to share this more broadly.

docs/rfcs/0007-flow-modules.md
polyrain: +1, this also makes the Goals section underneath much clearer as to what it's trying to communicate

docs/rfcs/0007-flow-modules.md
polyrain, line 186: `... has the advantage of *being simplifying*` can probably drop being here :)

docs/rfcs/0007-flow-modules.md
polyrain: Aiming for consistency is something I think modules should value, since the mental model of how to use them shouldn't differ too wildly between loading a single module or many at a time. If that means an initial hurdle on getting familiar with the system (since modules are already somewhat of an intermediate feature in terms of difficulty) for the single use case, personally that feels like an acceptable tradeoff.

docs/rfcs/0007-flow-modules.md
polyrain: +1 on your comment, having capability based permissions that can be applied on a module long term would be a nice addition long term, being able to define interactions with FS or remote_writes etc, but that feels like a much larger thing to discuss about the overall component eco-system in Flow :)

docs/rfcs/0007-flow-modules.md
polyrain, line 56: For this RFC I don't think so, but in order for modules to serve the use case of Agent-as-a-Service as laid out in the summary, metadata is pretty vital.

docs/rfcs/0007-flow-modules.md
```diff
@@ -0,0 +1,295 @@
...
+* [Configuration blocks](https://grafana.com/docs/agent/latest/flow/reference/config-blocks/#configuration-blocks) will not be supported. 
+* Arguments and exports within a module must be unique
+
+## Proposal
+
+Add the ability to load `modules` as subgraphs to the primary `graph`. Modules may call other modules within a reasonable stack size depth. Modules are represented as a river string that is interpreted with a defined set of arguments and exports.
+
+The initial component will be `module.single` that will load a single module. Internally these modules will be namespaced so they cannot affect children or parent graphs except via arguments and exports.
+
+Modules will have access to any standard function and any other component exempting singletons. Internally each component in the module will have an `id` that is prepended with the parent's `id` for identification purposes outside of the module. Within the module a component can reference another sibling component normally. There are no known limits on the datatype that a module can use as an argument or export.
```
polyrain, line 133: It might help to put an example of the `id`  you mention here. Correct me if I'm wrong, but how this reads to me is if I had the following graph of module dependencies:
`A -> B -> C -> D`
Then a component in module `D`, let's call it a `discovery.relabel "relabeler"` would have an `id` field on it that looks like `ABCD`. 

The word 'internally' is throwing me of here. Is this `id` intended to be used by the end user to reference/identify a component nestled deep in a module dependency path, or is this just for debugging/logging?

docs/rfcs/0007-flow-modules.md
```diff
@@ -0,0 +1,295 @@
...
+
+### Component Options
+
+
+Given the above example, the `id` of `integrations.mysql "server1"` would be `module.single.mysql.integrations.mysql.server1`. The `data-agent` field would also be prefixed. There are some inherent issues, deeply nested metrics are likely to run into Prometheus label value limits. On Windows platforms there could be issues with the `data-agent` length. These are issues that currently exist in Agent Flow but are more easily hit using deeply nested modules.
+
+
+### Failure Modes
+
+#### Option 1: When a module fails then fail itself and children.
```
polyrain, line 145: I do like this approach since it matches up with the idea that modules essentially append sub-graphs onto the overall graph that makes up the Agent configuration, so if the parent breaks you'd expect the failure to cascade down the children. I do see the issue of siblings to the module graph potentially breaking from this, but I guess the call on this is how Flow wants to shape how people should think about the tree of components. 

Option 2 is more relevant to a production use case, though, so in that context Option 2 seems the most correct option for where Flow wants to exist :)

docs/rfcs/0007-flow-modules.md
polyrain: I agree with this, mainly since the multi-use case introduces a lot of additional complexity that doesn't necessarily need to be discussed up front with the idea of modules. Flow describes itself as[ being made up of components](https://grafana.com/docs/agent/latest/flow/concepts/components/)that handle a single task (which of course _can_ go very deep, but thus far most components scope is not huge by any means).

Modules take this one step further and provide a wrapper over multiple components to give you a black box of sorts that does something more interesting/common use case/etc, which is great! Multi-module support feels like sugar on top of that for when you want to bundle _even more_, instead of just listing all the individual pieces together and keeping track of a large list. That's also great, but feels like jumping the gun while we're still figuring out what it means to load a _single module_; sugar before the tea, cart before the horse, and all that :^) 

That's just my 2c, but I 100% agree the multi-module aspect should be moved into its own RFC once what a module is has been more solidified.

docs/rfcs/0007-flow-modules.md
```diff
@@ -0,0 +1,295 @@
...
+
+## Option 2: Filter
+
+Allow `filter_arguments` and `filter_exports` to only include modules that define arguments and exports in the filters. Modules may have additional arguments that can optionally be set.
+
+
+
+#### Filter by arguments and exports
+
+Pass all files to both `load` and `load2` filtering if they have the appropriate input. This can lead to duplicate loading of modules if a module defines both `input1` and `input2`.
```
polyrain, line 215: I like this approach a lot, since it's essentially just filtering based on if a module implements some interface which is easy to understand and makes using third party modules simpler.

polyrain, line 215: Are you meaning arguments when you mention `input1` etc here? If so the terminology should be consistent

polyrain (commented):
Very excited to see modules being formalised, and the approach mentioned here looks solid to me. Left some comments :)

rfratto:
On the topic of module loaders which can load more than one module source: I would _love_ to see some kind of `module.kubernetes` component which can discover FlowModule CRDs and load them into memory. If we create a dedicated RFC for multi-source module loaders, could we explore how that would work?

docs/rfcs/0007-flow-modules.md
rfratto: Given my [comment](https://github.com/grafana/agent/pull/2898#issuecomment-1414827146) about FlowModule CRDs, I think module loaders should be conceptually flexible enough that some module loaders are given a module source, while others may discover module sources on their own.

docs/rfcs/0007-flow-modules.md
```diff
@@ -0,0 +1,295 @@
...
+During this time the Agent team saw a lot of potential in modules outside of scraping service. Packaging up sets of components for specific workflows, publishing common use cases and allowing better usage for internal customers in the Agent as a Service model.
+
+## Goals
+
+* Support single module loading via `module.single`
+* Support multiple module loading via `module.multiple`
+* Enable re-use of common patterns
+* Allow loading a module from a string
+* Allow modules to load other modules
+* Modules should be sandboxed except via arguments and exports
```
tpaschalis, line 30: Components are sandboxed in the same way too, right? Is there any other limitation you'd like to impose, or do you simply mean eg. not being able to modify the loaded module code?

docs/rfcs/0007-flow-modules.md
mattdurham: Absolutely.

docs/rfcs/0007-flow-modules.md
mattdurham: IMO out of scope but may change later. I will add the verbiage.

docs/rfcs/0007-flow-modules.md
mattdurham: Yes, I want to state that even if you know the full namespace id of a component you cannot access it.

docs/rfcs/0007-flow-modules.md
mattdurham: In this RFC, I want to only focus on the 3rd use case. The other two feel like solid things to do once we have solidified the core functionality.

docs/rfcs/0007-flow-modules.md
mattdurham: Agreed, go to change this from a `should we`, to a this is a non-goal for v1 but a long-term goal.

docs/rfcs/0007-flow-modules.md
mattdurham: For modules v1, a module input is explicitly a string. Will also mark that this is only a v1 consideration and other implied sources may happen. I am going to remove `module.multiple` from the RFC and mark it as a future goal.

docs/rfcs/0007-flow-modules.md
mattdurham: My fear here is that since comments would not be a field it will require people knowing it exists by convention. We can document the behavior but being a field makes it more inline with how components normally work.

docs/rfcs/0007-flow-modules.md
mattdurham: Let's go with its not enforced by the system. Will change the verbiage.

docs/rfcs/0007-flow-modules.md
mattdurham: Agreed with the above and removing `modules.multiple`.

docs/rfcs/0007-flow-modules.md
mattdurham: This is the ideal situation, and the one we can start with but I think there is some flexibility here. In the current reload we only have to check to see if the config is valid once, in modules that can happen many times and we will need to keep the many old and new sub-graphs running. I have concerns this may lead to duplication.

docs/rfcs/0007-flow-modules.md
mattdurham: Renamed `module.single` to `module.string` to give more room for different types of module loaders in the future.

mattdurham (commented):
RFC feedback

docs/rfcs/0007-flow-modules.md
rfratto: Sorry, I'm confused. What use cases? I'm just suggesting we break down what we're already describing in this RFC into those three terms.

rfratto: Yes, I was suggesting that the way Flow functions now is by loading the "main module." You could say that the main module has different constraints (i.e., configuration blocks are only valid in the main module, but no submodule).

docs/rfcs/0007-flow-modules.md
mattdurham: The configuration you write does not care but the only reason I included it is because of metric naming and passing the data path for components to use. So for a metric you may have `failed_requests{id="module.single.loki.loki.receiver.syslog.mysyslogreceiver"} 0`. In this case the `module.single.loki "loki"` module loads component `loki.receiver.syslog "mysyslogreceiver"` The data path for any file system usage would be equally odd.

docs/rfcs/0007-flow-modules.md
```diff
@@ -0,0 +1,232 @@
...
+
+prometheus.scrape "scraper" {
+    targets = module.string.mysql.exports.targets
+}
+
+```
+
+## Limitations
+
+* A module cannot directly or indirectly load itself, this will not be enforced by the system
```
tpaschalis, line 122: s/will not/will ?

docs/rfcs/0007-flow-modules.md
ptodev: Is the "comment" argument going to be read by something, or is it just meant to be a regular comment for the users who read and edit the river file? If it's used by other components/modules/UIs then I could see why it's an attribute, but if it's just a comment then I'm not sure I see the advantage vs. a regular river comment.

docs/rfcs/0007-flow-modules.md
ptodev: I think if we go with this approach, we would have to indicate to the users which modules/components failed to load  the latest value.

What are the main reasons for not being able to load the latest value? Is it only due to invalid config (e.g. River syntax issues)?

The safest approach imo would be that if a config reload fails a subsection of the graph, the whole graph needs to run on the last valid state, and not just a subsection of it.

docs/rfcs/0007-flow-modules.md
```diff
@@ -0,0 +1,232 @@
...
+* Singleton components are not supported at this time. Example [node_exporter](https://grafana.com/docs/agent/latest/flow/reference/components/prometheus.integration.node_exporter/).
+* Modules will not prevent competing resources, such as starting a server on the same port
+* [Configuration blocks](https://grafana.com/docs/agent/latest/flow/reference/config-blocks/#configuration-blocks) will not be supported. 
+* Names of arguments and exports within a module must be unique across that module.
+
+## Proposal
+
+Add the ability to load `modules` as subgraphs to the primary `graph`. Modules may call other modules within a reasonable stack size depth. Modules are represented as a river string that is interpreted with a defined set of arguments and exports.
+
+The initial component will be `module.string` that will load a single module. Internally these modules will be namespaced so they cannot affect children or parent graphs except via arguments and exports.
```
ptodev, line 132: If a module reports certain metrics, then what would happen if we import the module more than once? I guess that there will be a label to indicate the fully qualified module name which contains these metrics?

rfratto:
cc @slim-bean, since I know you were invested in modules as a way to enable centralized management.

docs/rfcs/0007-flow-modules.md
mattdurham: It's metadata. Having all other metadata be attributes and comments not feels weird.

docs/rfcs/0007-flow-modules.md
mattdurham: The namespace will include the parent labels, so that will prevent duplication. 

```river
module.string "modulo" {
    content = "some string that's valid"
}

module.string "samestringmodule" {
  content  = "some strings that's valid"
}
```

In the above, the components for `module.string.modulo` will be prefixed with `module.string.modulo` which trickles down to `data-agent` and metric nanes. Components under `module.string.samestringmodule` would be prefixed with `module.string.samestringmodule`.

docs/rfcs/0007-flow-modules.md
```diff
@@ -0,0 +1,232 @@
...
+## RFC Goals
+
+* Explain the use cases of modules
+* Explain what modules are
+* Go over possible syntax for modules
+* Go over pros and cons of modules
+
+## Summary
+
+One of the primary goals for the production usage of Agent Flow is parity with the static subsystem. One of the features of the static subsystem is [scraping service](https://github.com/grafana/agent/blob/main/docs/sources/configuration/scraping-service.md). Scraping service allows a user to run a fleet of agents and have thousands of scrape configurations distributed among running Agents. Through discussions within the Agent Team, we did not want to limit dynamically loading content to only scrape configs but allow most components to be loaded and used.
```
marctc, line 19: Should we add link to Agent Flow rfc?

rfratto (approved):
@mattdurham Let's get this merged!

PR #2904, opened 2023-02-02
Lazily register config metrics

spartan0x117:
#### PR Description
Changes config metrics to be lazily registered. In the case where a different module imports the agent it prevents these metrics from being exposed.

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated


jcreixell (approved):
I have verified this an it works as expected :100:

tpaschalis (approved):
Thanks!

PR #2905, opened 2023-02-02
Fix #integrations-api anchor in api docs

v-zhuravlev:
Fix #integrations-api anchor in api docs


marctc (approved):
Thanks!

PR #2915, opened 2023-02-02
replace go4.org/unsafe/assume-no-moving-gc until it's updated for Go 1.20

rfratto:


go.mod
```diff
@@ -83,112 +131,60 @@ require (
...
 	k8s.io/apimachinery v0.25.4
 	k8s.io/client-go v0.25.4
+	k8s.io/component-base v0.25.4
 	k8s.io/klog/v2 v2.80.1
 	k8s.io/utils v0.0.0-20221128185143-99ec85e7a448
 	sigs.k8s.io/controller-runtime v0.13.1
 	sigs.k8s.io/yaml v1.3.0
 )
 
-require (
```
rfratto, line 117: I combined the requires and re-ran `go mod tidy`, I'm not sure why we ended up with two of these at some point.

spartan0x117 (approved):
LGTM

Issue #2424, opened 2022-10-24
Flow: What labels should `prometheus.exporter.unix` export? 


rfratto:
Currently, the `prometheus.exporter.unix` component exports a target with a minimal set of labels:

https://github.com/grafana/agent/blob/18eeffa4bda2371ecd658f3ce1cbe427947434f2/component/prometheus/integration/node_exporter/node_exporter.go#L53-L58

The v1 integrations subsystem appends three labels:

* `instance`: Set to an integration-specific ID 
* `job`: Set to `integrations/INTEGRATION_NAME` 
* `agent_hostname`: Set to `$HOSTNAME` (if set) or to `os.Hostname()` otherwise. 

In addition to the labels listed above, the v2 integrations subsystem includes meta labels which can be useful for relabeling: 

* `__meta_agent_integration_name`: alias to `job`  
* `__meta_agent_integration_instance`: set to the name of the integration
* `__meta_agent_integration_autoscrape`: `"true"` when autoscrape is enabled for this integration

The only non-required label the Flow component sets is `name`, which is similar to `__meta_agent_integration_name` from the v2 integration subsystem. 

Not all of v1/v2 integration labels may be relevant to the Flow component equivalent, but I think we should determine of the current minimal list of labels was deliberate, and if not, which labels should be added to it. 

cc @captncraig

captncraig:
Great points. I think the most relevant ones are definitely the job and name labels if you want to manipulate them later.

hostname, I kinda feel like should come from relabels or external labels later in the pipeline, but I could be convinced otherwise if you have strong feelings. Autoscrape does not have an equivalent in flow right now.

So my new proposed set for #2916 is:

```
jobName := fmt.Sprintf("integrations/%s", name)

model.AddressLabel:                  opts.HTTPListenAddr,
model.SchemeLabel:                   "http",
model.MetricsPathLabel:              path.Join(opts.HTTPPath, "metrics"),
"instance":                          opts.ID,
"job":                               jobName,
"__meta_agent_integration_name":     jobName,
"__meta_agent_integration_instance": opts.ID,
```

rfratto:
> hostname, I kinda feel like should come from relabels or external labels later in the pipeline, but I could be convinced otherwise if you have strong feelings.

Removing hostname is fine with me, provided none of our integration dashboards rely on it being set. 

> So my new proposed set for https://github.com/grafana/agent/pull/2916 is:

LGTM 👍

captncraig:
[...]

PR #2916, opened 2023-02-02
flow: consul and apache integrations

captncraig:
Also moved a bunch of the code from node_exporter into a common package, much like discovery. Now all a new integration has to do is create a config type with a convert function, and wire it into the shared component.


component/prometheus/integration/integrationComponent.go
```diff
@@ -0,0 +1,119 @@
...
+// Update implements component.Component.
+func (c *Component) Update(args component.Arguments) error {
+	integration, err := c.creator(c.opts, args)
+	if err != nil {
+		return err
+	}
+	c.mut.Lock()
+	c.integration = integration
+	c.mut.Unlock()
+	c.reload <- struct{}{}
```
rfratto, line 98: Looks good, only have one drive-by comment:

```suggestion
	select {
	case c.reload <- struct{}{}:
	default:
	}
```

component/prometheus/integration/integrationComponent.go
```diff
@@ -0,0 +1,119 @@
...
+		// Call to Update() to set the output once at the start.
+		if err := c.Update(args); err != nil {
+			return nil, err
+		}
+		targets := []discovery.Target{{
+			model.AddressLabel:     opts.HTTPListenAddr,
+			model.SchemeLabel:      "http",
+			model.MetricsPathLabel: path.Join(opts.HTTPPath, "metrics"),
+			"name":                 name,
+		}}
```
rfratto, line 53: Reminder about #2424, it's not a blocker for this PR but it's still an open question for these integrations.

captncraig: Thanks for reminder. I will think it over and make sure we finalize before merging.

component/all/all.go
```diff
@@ -40,6 +40,7 @@ import (
 	_ "github.com/grafana/agent/component/otelcol/receiver/zipkin"              // Import otelcol.receiver.zipkin
 	_ "github.com/grafana/agent/component/phlare/scrape"                        // Import phlare.scrape
 	_ "github.com/grafana/agent/component/phlare/write"                         // Import phlare.write
+	_ "github.com/grafana/agent/component/prometheus/integration/consul"        // Import prometheus.integration.consul
 	_ "github.com/grafana/agent/component/prometheus/integration/node_exporter" // Import prometheus.integration.node_exporter
```
marctc, line 44: apache needs to be included too, right?

component/prometheus/integration/integrationComponent.go
```diff
@@ -0,0 +1,119 @@
+package integration
```
marctc, line 1: Could this code also be used for non prometheus integrations (eventhandler)? also nit: use `_` or `-` for a filename instead of camel case?

marctc (commented):
Looking good. We probably should add tests. For integrations should be enough to test conversion is properly done I guess

jcreixell (approved):
nice, 2 in 1, LGTM. The only question I have is if we want to have a consistent way of testing the configuration mapping (Convert function, in this case it is simple enough, in other cases it may be more complex).

If we don't want to test it and it isn't part of an interface, we may as well make it private

component/prometheus/integration/integrationComponent.go
captncraig: Yes, it possibly could. We may need to add additional extension points for that, but it could be done.

docs/sources/flow/reference/components/prometheus.integration.apache.md
```diff
@@ -0,0 +1,77 @@
...
+}
+```
+
+## Arguments
+The following arguments can be used to configure the exporter's behavior.
+All arguments are optional. Omitted fields take their default values.
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`scrape_uri`    | `string` | URI to apache stub status page. | `http://localhost/server-status?auto` | no
```
karengermond, line 29:
```suggestion
`scrape_uri`    | `string` | URI to Apache stub status page. | `http://localhost/server-status?auto` | no
```

docs/sources/flow/reference/components/prometheus.integration.apache.md
```diff
@@ -0,0 +1,77 @@
...
+## Exported fields
+The following fields are exported and can be referenced by other components.
+
+Name      | Type                | Description
+--------- | ------------------- | -----------
+`targets` | `list(map(string))` | The targets that can be used to collect `consul` metrics.
+
+For example, the `targets` could either be passed to a `prometheus.relabel`
+component to rewrite the metrics' label set, or to a `prometheus.scrape`
+component that collects the exposed metrics.
```
karengermond, line 42:
```suggestion
For example, the `targets` can either be passed to a `prometheus.relabel`
component to rewrite the metric's label set, or to a `prometheus.scrape`
component that collects the exposed metrics.
```

karengermond: I'm unsure of this one.  It's the label set of a particular metric, right? So it would be "metric's"...

karengermond (approved):
LGTM!

PR #2917, opened 2023-02-02
flow: add `constants` to stdlib

### LINKED ISSUE #2703: Flow: add `constants` identifier
I think it would be immensely useful to have some constants available for use throughout expressions:

* `constants.hostname`: hostname of the machine the Agent is running on 
* `constants.os`: Operating System the agent is running on (e.g., GOOS)
* `constants.arch`: Architecture of the system the agent is running on (e.g., GOARCH)

Maybe more would be useful over time, but these are the initial set that I'd personally like to see. 

These would be especially useful alongside #2638 to use them inside conditional expressions.

rfratto:
Add `constants` identifier to stdlib which returns an object of various constants exposing information about the system that Grafana Agent is running on. The initial set of constants is:

* `constants.hostname`: The hostname of the system that Grafana Agent is running on.
* `constants.os`: The operating system that Grafana Agent is running on. 
* `constants.arch`: The architecture of the system that Grafana Agent is running on.

Closes #2703.

### DIFF
--- a/docs/sources/flow/reference/stdlib/constants.md
+++ b/docs/sources/flow/reference/stdlib/constants.md
@@ -1,0 +1,28 @@
+---
+aliases:
+- ../../configuration-language/standard-library/constants/
+title: constants
+---
+
+# constants
+
+The `constants` object exposes a list of constant values about the system
+Grafana Agent is running on:
+
+* `constants.hostname`: The hostname of the machine Grafana Agent is running
+  on.
+* `constants.os`: The operating system Grafana Agent is running on.
+* `constants.arch`: The architecture of the system Grafana Agent is running on.
+
+## Examples
+
+```
+> constants.hostname
+"my-hostaname"
+
+> constants.os
+"linux"
+
+> constants.arch
+"amd64"
+```
--- a/docs/sources/flow/reference/stdlib/constants.md
+++ b/docs/sources/flow/reference/stdlib/constants.md
@@ -18,7 +18,7 @@ Grafana Agent is running on:
 
 ```
 > constants.hostname
-"my-hostaname"
+"my-hostname"
 
 > constants.os
 "linux"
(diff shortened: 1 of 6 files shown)


docs/sources/flow/reference/stdlib/constants.md
```diff
@@ -0,0 +1,28 @@
...
+* `constants.hostname`: The hostname of the machine Grafana Agent is running
+  on.
+* `constants.os`: The operating system Grafana Agent is running on.
+* `constants.arch`: The architecture of the system Grafana Agent is running on.
+
+## Examples
+
+```
+> constants.hostname
+"my-hostaname"
```
rfratto, line 21:
```suggestion
"my-hostname"
```

pkg/flow/internal/controller/component_references.go
```diff
@@ -48,7 +48,7 @@ func ComponentReferences(cn dag.Node, g *dag.Graph) ([]Reference, diag.Diagnosti
 		//
 		// Any call to an stdlib function is ignored.
 		var emptyScope vm.Scope
-		if _, ok := emptyScope.Lookup(t[0].Name); ok && len(t) == 1 {
```
rfratto, line 51: I'm not sure what was with this check before, I think I was worried that there may eventually be something in the standard library whose name matches a namespace of a Flow component, but I don't think this is practically going to happen. 

I removed this check so that `constants.*` worked properly, since otherwise it thought `constants.hostname` was a reference to a component.

pkg/river/internal/stdlib/stdlib.go
```diff
@@ -15,6 +15,9 @@ import (
 // least one non-error return value, with an optionally supported error return
 // value as the second return value.
 var Functions = map[string]interface{}{
```
rfratto, line 17: Maybe it's not a great idea to put non-functions in something called functions. I can rename this or put the constants somewhere else :)

tpaschalis: Maybe s/Functions/Identifiers? I don't think it's a big deal either way.

tpaschalis (approved):

karengermond (approved):
LGTM

PR #2922, opened 2023-02-03
feat: adding structured config and ingress to support faro

ts-mini:
#### PR Description

Adding some helm chart features to enable faro through ingress
- ingress support (stolen from grafana chart)
- structuredConfig (stolen from loki chart)

Because integrations-next is still behind feature flags this doesn't work 'out of the box' for most users, without proper config - but I feel like these changes fall in line with the rest of the grafana ecosystem in terms of configuration.

Here is the config on how i was able to deploy (my values.yaml)
```
agent:
  mode: static
  extraArgs:
    - -enable-features=integrations-next
  structuredConfig:
    server:
      log_level: info
      log_format: json
    traces:
      configs:
        - name: default
          receivers:
            otlp:
              protocols:
                grpc: null
            jaeger:
              protocols:
                grpc: null
          remote_write:
            - endpoint: https://tempo-gateway.internal.infra/otlp/v1/traces
              insecure: false
              headers:
                x-scope-orgid: prod-cluster
          batch:
            timeout: 5s
            send_batch_size: 100
    metrics:
      wal_directory: '/tmp/wal-client'
      global:
        external_labels:
          __cluster__: prod-cluster
        scrape_interval: '60s'
        remote_write:
          - url: https://mimir-gateway.internal.infra/api/v1/push
            headers:
              'X-Scope-OrgID': 'prod-cluster'
            remote_timeout: 30s
            queue_config:
              capacity: 1500
              max_shards: 1000
              min_shards: 1
              max_samples_per_send: 500
              batch_send_deadline: 5s
              min_backoff: 30ms
              max_backoff: 100ms
      configs:
        - name: default

    logs:
      configs:
        - name: default
          positions:
            filename: '/tmp/positions-client.yaml'
          clients:
            - url: https://loki-gateway.internal.infra/loki/api/v1/push
              tenant_id: prod-cluster
    integrations:
      app_agent_receiver:
        autoscrape:
          enable: true
          metrics_instance: default
        instance: default
        logs_instance: default
        logs_send_timeout: '5s'
        server:
          cors_allowed_origins:
            - '*'
          host: '0.0.0.0'
          max_allowed_payload_size: 5e+07
          port: 8080
          rate_limiting:
            burstiness: 100
            enabled: true
            rps: 100
        sourcemaps:
          download: true
        traces_instance: default

ingress:
  enabled: true
  ingressClassName: nginx
  hosts:
    - some-host.tld
  tls:
    - hosts:
        - some-host.tld
      secretName: some-tls
```

None

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [ ] Tests updated


CHANGELOG.md
```diff
@@ -7,8 +7,8 @@ This document contains a historical list of changes between releases. Only
 changes that impact end-user behavior are listed; changes to documentation or
 internal API changes are not present.
 
-Main (unreleased)
------------------
+## Main (unreleased)
```
rfratto, line 10: Can we leave the formatting of the h1's untouched? For reading the raw text, I find the `-------` to be easier to distinguish when scanning the file from the other `###` headers.

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ Main (unreleased)
 
 ### Features
 
+- Helm Chart: Add ingress and support for agent-receiver. (@ts-mini)
+
+- Helm chart: Add support for structuredConfig in values.yaml (@ts-mini)
```
rfratto, line 22: These should go into the changelog of the helm chart, not the root changelog: https://github.com/grafana/agent/blob/main/operations/helm/charts/grafana-agent/CHANGELOG.md

operations/helm/charts/grafana-agent/values.yaml
```diff
@@ -9,12 +9,12 @@ fullnameOverride: null
 agent:
   # -- Mode to run Grafana Agent in. Can be "flow" or "static".
   mode: 'flow'
-
+  structuredConfig: {}
```
rfratto, line 12: Can you explain why we need both `structuredConfig` and `configMap.content`? I'm worried this adds too much extra complexity for understanding which values you need to set.

ts-mini: I've found the structured config super useful when deploying things like loki - because it allows me to reference values that i've set in the yaml-structured way.

so lets say I set an override, and want to use it over and over again, i can do it this way

CHANGELOG.md
ts-mini: oops

CHANGELOG.md
ts-mini: fixed

operations/helm/charts/grafana-agent/values.yaml
rfratto: I'm not sure I'm following. Even though the default value of `configMap.content` is a string, because we pass it through `toYaml`, can't you also give it a structured object?

rfratto: Ok, so I tried 

```yaml
agent:
  mode: static
  configMap:
    content:
      server:
        log_level: debug
``` 

and it doesn't work. But I would rather find a way to support `agent.configMap.content` to be either a string or an object rather than introducing a new field.

ts-mini: So its more about 'the future' I think.  Currently this chart ships with no usable config for the agent - but I assume that will change in the future.  

Under that assumption this allows users to edit/override ANY key easily in a structured way - allowing a downstream user to inherit the 'sane defaults' of the chart, configuration, parameters, but easily provide a non-string value to change a key thing here-or-there.  

I'm not tied to this change personally - it is just what's been in place with [tempo](https://github.com/grafana/helm-charts/blob/main/charts/tempo-distributed/values.yaml#L74) and [loki](https://github.com/grafana/loki/blob/main/production/helm/loki/values.yaml#L281) and I've found that using it makes maintaining upgrades and configuration more pleasant for the end user relative to the alternative

ts-mini: there are also just silly nice-to-have-things, like because its not an inline string, it looks prettier in the IDE or that you can use jq to edit/parse/validate easier but thats not really super important to most

ts-mini: >But I would rather find a way to support agent.configMap.content to be either a string or an object rather than introducing a new field.

ok! let me back this out of THIS pr and ill throw up a separate one with that change

rfratto:
Sorry for the delays on reviewing, I'm hoping to review this either today or tomorrow.

operations/helm/charts/grafana-agent/CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ internal API changes are not present.
 Unreleased
 ----------
```
rfratto, line 11:
```suggestion
Unreleased 
----------

0.5.0 (2023-02-08)
------------------
```

rfratto (approved):

rfratto:
Thank you!

PR #2927, opened 2023-02-06
[Agent Management] Add random jitter to http API requests

spartan0x117:
#### PR Description
Adds random jitter (e.g. a short `sleep`) to requests in the `remoteConfigHTTPProvider`. It's done as late as possible, immediately before the actual retrieval, to ensure that any invalid configuration is surfaced beforehand.

- [ ] CHANGELOG updated
- [ ] Documentation added
- [X] Tests updated


pkg/config/agentmanagement.go
```diff
@@ -81,6 +88,9 @@ func (r remoteConfigHTTPProvider) FetchRemoteConfig() ([]byte, error) {
 		return nil, fmt.Errorf("error reading remote config: %w", err)
 	}
 
+	// sleep for random time to add jitter to API requests
+	time.Sleep(jitterTime(httpMinJitterTime, httpMaxJitterTime))
```
marctc, line 92: Wouldn't make more sense to add the jitter time to the polling interval?

pkg/config/agentmanagement.go
```diff
@@ -181,6 +191,21 @@ func (am *AgentManagementConfig) SleepTime() (time.Duration, error) {
 	return time.ParseDuration(am.PollingInterval)
 }
 
+// jitterTime returns a random duration in the range [minJitterDuration, maxJitterDuration).
+func jitterTime(minJitterDuration time.Duration, maxJitterDuration time.Duration) time.Duration {
+	// If either duration is negative, or the maximum duration is 0 return no jitter.
+	if minJitterDuration < 0 || maxJitterDuration <= 0 {
```
marctc, line 197: If min and max are not going to be configurable I'm not sure what's the use case of having these checks here

pkg/config/agentmanagement.go
```diff
@@ -15,7 +16,11 @@ import (
 	"github.com/prometheus/common/config"
 )
 
-const cacheFilename = "remote-config-cache.yaml"
+const (
+	cacheFilename     = "remote-config-cache.yaml"
+	httpMinJitterTime = time.Duration(0)
+	httpMaxJitterTime = 10 * time.Second
```
marctc, line 22: What are those values coming from? I'm not sure how jitter implementation should be but I thought that it would be related with the polling interval. How we ensure that requests are somehow equally distributed when we have big polling intervals (such 60s)?

pkg/config/agentmanagement_test.go
```diff
@@ -132,6 +132,36 @@ func TestSleepTime(t *testing.T) {
 	assert.Equal(t, time.Second*15, st)
 }
 
+func TestJitterTime(t *testing.T) {
```
marctc, line 135: Maybe is a good time to try [fuzz tests](https://go.dev/security/fuzz/)? We probably should also tests that given a min and a max, the function return values that are equally distributed among the possible return options

pkg/config/agentmanagement.go
jcreixell: yeah, I think the jitter should be a function of the polling interval, with min and max percentage of the polling interval as parameters (for example, between 0 and 20% of the polling interval).

pkg/config/agentmanagement.go
```diff
@@ -181,6 +184,12 @@ func (am *AgentManagementConfig) SleepTime() (time.Duration, error) {
 	return time.ParseDuration(am.PollingInterval)
 }
 
+// jitterTime returns a random duration in the range [0, am.PollingInterval).
+func (am *AgentManagementConfig) JitterTime() time.Duration {
+	maxJitterDuration, _ := am.SleepTime()
+	return time.Duration(rand.Int63n(int64(maxJitterDuration)))
```
jcreixell, line 190: LGTM overall but let's allow some level of configuration here? (for example a percentage of the sleep time with some acceptable ranges)

jcreixell (commented):
See my comment regarding making this somewhat configurable, other than that LGTM

marctc (approved):

PR #2929, opened 2023-02-06
flow: Add logs fanout abstraction

tpaschalis:
#### PR Description
This PR introduces the simplest abstraction to more easily manage fanning out to multiple logging receivers.

Components who wish to use it will create a `NewFanout`, use `UpdateReceivers` when their Update method. Components will then be able to work as:

```go
	for {
		select {
		case <-ctx.Done():
			return nil
		case entry := <-c.handler:
                          c.fanout.Send(ctx, entry)
		}
	}
```

If the context is cancelled while the fanout is blocked on a receiver, then Send will return prematurely and allow the Run method to exit as well. For performance-sensitive components, it will also allow users to configure a timeout and make use of the SendWithTimeout method. 

No issue filed.

#### Notes to the Reviewer
cc-ing people involved in loki components for feedback, so apologies for the noise.

Do you think this adds enough value to be justified? Would you like to also see anything more that would help in common cases with components?

TODO: Add metrics?

#### PR Checklist

- [ ] CHANGELOG updated (N/A)
- [ ] Tests updated (N/A)

### DIFF
--- a/component/common/loki/types.go
+++ b/component/common/loki/types.go
@@ -147,3 +147,55 @@ func AddLabelsMiddleware(additionalLabels model.LabelSet) EntryMiddleware {
 		})
 	})
 }
+
+// LogsFanout is an abstraction for fanning out log entries to multiple
+// receivers.
+type LogsFanout struct {
+	mut       sync.RWMutex
+	receivers []LogsReceiver
+}
+
+// NewFanout creates new LogsFanout.
+func NewFanout(r []LogsReceiver) *LogsFanout {
+	return &LogsFanout{
+		receivers: r,
+	}
+}
+
+// UpdateReceivers updates where LogsFanout should send entries to.
+func (f *LogsFanout) UpdateReceivers(r []LogsReceiver) {
+	f.mut.Lock()
+	f.receivers = r
+	f.mut.Unlock()
+}
+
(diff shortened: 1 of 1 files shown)


component/common/loki/types.go
```diff
@@ -147,3 +147,55 @@ func AddLabelsMiddleware(additionalLabels model.LabelSet) EntryMiddleware {
...
+
+// LogsFanout is an abstraction for fanning out log entries to multiple
+// receivers.
+type LogsFanout struct {
+	mut       sync.RWMutex
+	receivers []LogsReceiver
+}
+
+// NewFanout creates new LogsFanout.
+func NewFanout(r []LogsReceiver) *LogsFanout {
```
tpaschalis, line 159: Should we go one layer up, and include the handler which receives the entries within the Fanout itself, instead of having to receive it and call the Send method every time from the component's Run method?

LogsFanout could be started in its own rungroup and pass along entries in the background.

thampiotr: I would be in favour of including the logic that reads from a channel, as it will further reduce duplication. Could be part of `LogFanout`, but could also be a new struct that takes `LogFanout` in constructor.

thampiotr (commented):
👍 this looks like a good idea to me!

PR #2930, opened 2023-02-06
Prepare for release 0.31.1

rfratto:
I'll update the release branch with all the commits needed for 0.31.1 after this is merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ Main (unreleased)
 
 ### Features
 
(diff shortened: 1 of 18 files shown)


ptodev:
There is no release candidate for the patch release? Is that the case with most patch releases?

rfratto:
> There is no release candidate for the patch release? Is that the case with most patch releases?

Right, there's only a release candidate for the `.0` release.

ptodev (approved):

PR #2936, opened 2023-02-07
[Agent Management] Check remote cache for validity using url hash

spartan0x117:
#### PR Description
Adds a new `remoteConfigCache` struct. It contains a hash of the url as well as the raw config. When loading the config from the cache, the hash of the url currently being used is compared against that of the cache. If they don't match, that means a different url was used when generating the cache (perhaps a different namespace or set of labels) and it is considered an invalid cache.

This change also incidentally removes a call to `LoadFile`, which would call `InstrumentConfig` and set the `agent_config_hash` gauge briefly before it is overridden [here](https://github.com/grafana/agent/blob/70e03f7fb63d5b3ccc70707bf1ef3dabca79da4d/pkg/config/config.go#L276).

Finally, removes an unused test file.

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated


pkg/config/agentmanagement.go
```diff
@@ -37,22 +40,67 @@ func newRemoteConfigHTTPProvider(c *Config) (*remoteConfigHTTPProvider, error) {
 	}, nil
 }
 
+type remoteConfigCache struct {
+	UrlHash string `json:"url_hash"`
+	Config  string `json:"config"`
+}
+
+func hashUrl(u string) string {
```
jcreixell, line 48: would we want to invalidate the cache when other fields change? (for example the namespace or labelset)

spartan0x117: For `namespace`, at least, I'd imagine so, as it could have an entirely different base config. For `labelset`, I'd personally think also yes. 

For example, if one were to stop a service running on a node (say previously a database was running but now is not) and remove the associated label, if the agent were restarted with the new `labelset` and unable to fetch the `remote_config` it would use the old cached config that had the snippets related to the now-stopped database process.

pkg/config/agentmanagement.go
```diff
@@ -96,12 +144,13 @@ type RemoteConfiguration struct {
...
-	Protocol        string           `yaml:"protocol"`
-	PollingInterval string           `yaml:"polling_interval"`
-	CacheLocation   string           `yaml:"remote_config_cache_location"`
+	Enabled                    bool             `yaml:"-"` // Derived from enable-features=agent-management
+	Url                        string           `yaml:"api_url"`
+	BasicAuth                  config.BasicAuth `yaml:"basic_auth"`
+	Protocol                   string           `yaml:"protocol"`
+	PollingInterval            string           `yaml:"polling_interval"`
+	CacheLocation              string           `yaml:"remote_config_cache_location"`
+	InvalidateCacheOnUrlChange bool             `yaml:"invalidate_cache_on_url_change"` // Whether to invalidate the cached config if the url hash is different
```
jcreixell, line 153: I am not sure about this, probably we don't really need to make this configurable? (it makes sense in general not to fall back to something that is clearly a left over from a previous initial config)

marctc (approved):
LGTM. Is it worth/possible to add some tests?

Issue #2602, opened 2022-12-09
River: support squashing structs into slices


rfratto:
This proposal extends #2373 to support squashing structs into slices. 

Shoutout to @tpaschalis for the help creating this proposal :) 

## Background 

Component authors may wish to support multiple variant blocks, but retain the order they are decoded in. This need was first identified in #2589, where users can provide one or more stage blocks to transform log lines, where stages are applied in the order they were defined in the config file. Unlike prometheus.relabel, there are multiple types of stages that users can use. 

Due to how the River API currently works, users must write stages like so:

```river 
loki.transform "example" {
  forward_to = [...]

  stage { 
    json { 
      ...
    }
  }
  stage {
    labels {
      ...
    } 
  }
}
```

This is verbosity being pushed to users because of the current limitations of the River API. It would be much more preferable for users to write the following:

```river 
loki.transform "example" {
  forward_to = [...]

  stage.json {
    ...
  }
  stage.labels {
    ...
  }
}
```

This proposal details changes to make to the River API to support decoding this while retaining the order of the blocks for proper processing of the stages. 

## Proposal 

I propose extending #2373 to permit squashing struct fields into slices. Component authors (like loki.process) could then define their arguments as:

```go 
type Arguments struct {
  ForwardTo []loki.LogsReceiver  `river:"forward_to,attr"`
  Stages    []StageConfig        `river:",squash"`
}

type StageConfig struct {
 JSON   *JSONStage   `river:"stage.json,block,optional"` 
 Labels *LabelsStage `river:"stage.labels,block,optional"`  
}
```

## Implementation 

When the River VM encounters a block which is defined by a squashed slice, the block is evaluated and decoded into a new instance of the struct, which is then appended to the slice.  This means that each squashed slice element will only ever have one field decoded. 

### No attributes 

To prevent users from being able to define attributes more than once, squashing slices is only supported when the struct type being sliced contains no attributes. For example, this is invalid:

```go
type Person struct {
  Attrs []Attributes `river:",squash"` 
}

type Attributes struct {
  // INVALID: Attribute is squashed as a slice into Person,
  // which would otherwise allow this strange config:
  //
  //   name = "John Doe"
  //   name = "Jane Doe" 
  Name string `river:"name,attr,optional"` 
}
``` 

The River VM must panic when given a squashed slice which breaks these rules. 

### Encoding 

For working with the River Value encoder, blocks within a slice of squashed structs must start with the same name fragment. 

[...]

Issue #1957, opened 2022-07-27
River: diagnostics error type not used consistently


rfratto:
River currently doesn't consistently use the `diag.Diagnostics` consistently through all the exposed APIs. For example, using a block name as an attribute returns [a fmt.Errorf](https://github.com/grafana/agent/blob/0004e01dea47f6bfcf57c5ad41ab48ea99b2b19a/pkg/river/vm/vm.go#L186) instead of diag.Diagnostics.

We should make sure that diag.Diagnostics are used consistently so that multiple errors can be aggregated and reported against with correct position information.

ptodev:
Hi, I pushed some changes to my branch. For the ones that are implemented, does it makes sense? E.g.:
```
case tf.IsAttr() && len(blocks) > 0:
	return fmt.Errorf("%q must be an attribute, but is used as a block", fullName)
	return diag.Diagnostic{
		Severity: diag.SeverityLevelError,
		StartPos: ast.StartPos(blocks[0]).Position(),
		EndPos:   ast.EndPos(blocks[len(blocks)-1]).Position(),
		Message:  fmt.Sprintf("%q must be an attribute, but is used as a block", fullName),
	}
```

For some other ones, I wasn't quite sure what's the best way to get the parent or whether we should get the parent:
```
case tf.IsAttr() && len(attrs) == 0 && !tf.IsOptional():
	return fmt.Errorf("missing required attribute %q", fullName)
	return diag.Diagnostic{
		Severity: diag.SeverityLevelError,
		//TODO: We need to report the start and end position of the block where this attribute should be?
		StartPos: ast.StartPos().Position(),
		EndPos:   ast.EndPos().Position(),
		Message:  fmt.Sprintf("missing required attribute %q", fullName),
	}
```

rfratto:
Can we do code discussions via PR instead of comments, please?

PR #2962, opened 2023-02-09
river/vm: refactor decode order

rfratto:
Previously, decoding into a struct would iterate through the tags, and decode each tag in order. Blocks would be grouped by name, and then decoded in into the slice in bulk.

To faciliate #2602, where a slice may have multiple elements from blocks of different names, the iteration order must be reversed. Now, we iterate through statements in the AST, and decode them in order specified in the configuration file. This has the added benefit of returning evaluation errors in order of the config file rather than in order of the tags specified by the Go struct.

A subset of errors has been updated to return `diag.Diagnostic`, as part of #1957 (but does not complete #1957).

Benchmark results show a very minor slowdown in performance, but it is likely within acceptable bounds:

```
name                    old time/op    new time/op    delta
Exprs/object_to_map-10    1.10µs ± 1%    1.11µs ± 1%  +0.44%  (p=0.042 n=19+20)

name                    old alloc/op   new alloc/op   delta
Exprs/object_to_map-10    1.45kB ± 0%    1.45kB ± 0%    ~     (all equal)

name                    old allocs/op  new allocs/op  delta
Exprs/object_to_map-10      19.0 ± 0%      19.0 ± 0%    ~     (all equal)
```


pkg/flow/internal/controller/loader_test.go
```diff
@@ -205,7 +205,7 @@ func TestScopeWithFailingComponent(t *testing.T) {
 	diags := applyFromContent(t, l, []byte(testFile))
 	require.Error(t, diags.ErrorOrNil())
 	require.Len(t, diags, 1)
-	require.True(t, strings.Contains(diags.Error(), "Failed to build component: decoding River: missing required attribute \"frequency\""))
+	require.True(t, strings.Contains(diags.Error(), `unrecognized attribute name "frequenc"`))
```
rfratto, line 208: This needed to be changed because the new evaluation order will report unrecognized attributes first over missing a required attribute.

pkg/river/vm/vm_block_test.go
```diff
@@ -78,7 +78,7 @@ func TestVM_Block_Attributes(t *testing.T) {
 		eval := vm.New(parseBlock(t, input))
 
 		err := eval.Evaluate(nil, &block{})
-		require.EqualError(t, err, `"number" must be an attribute, but is used as a block`)
+		require.EqualError(t, err, `2:4: "number" must be an attribute, but is used as a block`)
```
rfratto, line 81: This error (and the one below it) is a diag.Diagnostic now, so I had to add the position information.

tpaschalis (approved):
LGTM after a first reading, let's go ahead with it!

PR #2966, opened 2023-02-10
Add CI tests to check generation of manifests and dashboards

ptodev:
#### PR Description
As advised by @rfratto on [this](https://github.com/grafana/agent/pull/2896#discussion_r1093587827) comment, it is good to have the CI check if any kubernetes environment definitions and dashboard definitions are out of date.

#### Notes to the Reviewer
I am not sure how to test this change. I tried changing the Agent version in these files, then running `make generate-dashboards` and `make generate-manifests`:
```
production/grafanacloud-install.sh
production/kubernetes/build/lib/version.libsonnet
```
However, I did not see any changes in any other files. 
What would be a good way to test the change?

#### PR Checklist

- [x] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated


.github/workflows/dashboard-test.yml
```diff
@@ -0,0 +1,21 @@
+name: Test dashboards
```
rfratto, line 1: If we add these to our Drone jobs instead, we can have it use the agent build image where `jb` is installed.

.github/workflows/manifests-test.yml
```diff
@@ -0,0 +1,21 @@
...
+
+      - name: Regenerate environment manifests
+        run: |
+          make generate-manifests
+          ERR_MSG="The environment manifests are out of date. Please run 'make generate-manifests' and commit changes!"
+          if ! git diff --exit-code; then
+            echo $ERR_MSG >&2
+            exit 1
+          fi
+          if [ ! -z "$(git status --porcelain)" ]; then
```
ptodev, line 18: Not sure if there is any  point in having this second `git status` check?

rfratto: It's needed; the second check verifies whether the make commands don't generate any _new_ files. The first command only verifies that existing files don't get updated.

ptodev: I see, thank you! Here `git status` would catch both tracked and untracked files. I suppose the reason we do the `git diff` is to give a more precise error message. In my PR I have the same message for both, so I could either delete the `git diff` or try to make a more precise error message for it. I would personally prefer to just do `git status` and report an error, without making too many assumptions about what the error is. Would you be ok with that? :)

.github/workflows/dashboard-test.yml
ptodev: Thanks! I got this to work on github actions as well with the agent build image, but it took longer. So I'll stick to drone for now.

CHANGELOG.md
```diff
@@ -59,6 +61,9 @@ Main (unreleased)
 - Use Go 1.20 for builds. Official release binaries are still produced using Go
   1.19. (@rfratto)
 
+- Check deployment manifests and dashboard definitions on every 
+  pull request (@ptodev)
```
rfratto, line 65: This isn't a user facing change, so we wouldn't document it.

.github/workflows/manifests-test.yml
rfratto: Sure, that's fine. I'll make the change since I'm quickly doing finishing touches so we can merge this now.

rfratto (approved):
Thanks!

PR #2970, opened 2023-02-12
Add timeout to loki.source.podlogs controller setup

### LINKED ISSUE #2866: Flow: add timeout to loki.source.podlogs controller setup
`loki.source.podlogs` can block forever (or until the component is reloaded) waiting for informers to sync. This is most common when the agent is run with inappropriate permissions (such as using a Role instead of a ClusterRole and watching all namespaces). 

Configuring informers should have a timeout, and if the timeout expires, log a helpful message about checking that permissions are correct or that the connection to Kubernetes API is stable.

polyrain:
#### PR Description
This PR adds a timeout when configuring Informers in `loki.source.podlogs`, which had the potential to block forever if 
the Agent had insufficient RBAC permissions or flaky connection to the control plane. After 10 seconds the context is cancelled and if the error branch is hit due to this, a more informative error message is returned explaining possible causes.

Fixes #2866

#### Notes to the Reviewer
Golang is an area I'm not the most fluent in, apologies if some aspects aren't idiomatic; feedback appreciated :) 

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,8 @@ Main (unreleased)
 - Flow: add `constants` into the standard library to expose the hostname, OS,
   and architecture of the system Grafana Agent is running on. (@rfratto)
 
+- Flow: add timeout to loki.source.podlogs controller setup. (@polyrain)
+
 ### Other changes
 
 - Use Go 1.20 for builds. Official release binaries are still produced using Go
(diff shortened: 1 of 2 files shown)


rfratto (approved):
LGTM, thank you!

PR #2983, opened 2023-02-14
docs: remove beta status from Flow

rfratto:
Flow is now implicitly considered production ready. A subset of components are singled out as still being beta or experimental: 

* `loki.echo` is explicitly marked as beta.
* `loki.source.kubernetes` is explicitly marked as experimental.
* `loki.source.podlogs` is explicitly marked as experimental.
* `mimir.rules.kubernetes` is explicitly marked as beta.
* `otelcol.exporter.loki` is explicitly marked as beta.
* `otelcol.exporter.prometheus` is explicitly marked as beta.
* `otelcol.receiver.loki` is explicitly marked as beta.
* `otelcol.receiver.prometheus` is explicitly marked as beta.
* `phlare.scrape` is explicitly marked as beta.
* `phlare.write` is explicitly marked as beta.

Other Flow components not listed here are considered production ready and generally available.


docs/sources/flow/_index.md
```diff
@@ -78,18 +72,3 @@ prometheus.remote_write "default" {
...
 [reference documentation]: {{< relref "./reference" >}}
 
-## Current limitations
-
-The goal of Grafana Agent Flow is to eventually support the same use cases that
-Grafana Agent does today. Some functionality may be missing while Grafana Agent
-Flow is still in development:
-
-* An equivalent list of integrations
-* An equivalent to the scraping service
```
rfratto, line 88: These limitations still currently exist, but since we're removing the beta status, it's probably appropriate to move to timeless documentation.

docs/sources/flow/_index.md
```diff
@@ -78,18 +72,3 @@ prometheus.remote_write "default" {
...
-
-## Provide feedback
-
```
tpaschalis, line 91: We could maybe keep this, no reason to throw it out right?

tpaschalis (approved):
👍

rfratto:
I'm going to bring this back in draft for a bit so it doesn't get accidentally merged. I think we need to consider whether we should single out any Flow components as experimental/beta before merging this.

docs/sources/flow/_index.md
rfratto: I wanted to remove the feedback section because I'm worried it implies that Flow is still young or not production ready. 

I think people can rely on the "prove feedback" links elsewhere, like at the top level or at the project README. 

I'm going to merge this as-is to keep things moving, but if you're still unsure, let's discuss it and we can bring this back in a follow up PR :)

PR #3007, opened 2023-02-15
Drone: build each Docker container in different stages

rfratto:
This should help speed up CI/CD time as the images are built and pushed in parallel.


erikbaranowski:
I think we need an update for Check Windows containers to split it into 2 pipelines or call the bash script twice. Passing in the specific executable it plans to build from these updates

erikbaranowski (approved):

PR #3008, opened 2023-02-15
squash http_client_config for prometheus.scrape

### LINKED ISSUE #2994: Flow: squash `http_client_config` in `prometheus.scrape`
The `http_client_config` block in `prometheus.scrape` should be squashed to match the squashing used in the upstream config.

erikbaranowski:
#### PR Description

Squash http_client_config for the first time and for prometheus.scrape.

Closes #2994 

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated


component/prometheus/scrape/scrape_test.go
```diff
@@ -9,12 +9,71 @@ import (
...
+	scrape_interval = "10s"
+	job_name        = "local-flow"
+
+	bearer_token = "token"
+	bearer_token_file = "/path/to/file.token"
+	proxy_url = "http://0.0.0.0:11111"
+	follow_redirects = true
+	enable_http2 = true
+
+	basic_auth {
```
erikbaranowski, line 32: I feel like it is good practice to test the blown up river config at least happy path in the tests for all these. With that, do we want to put a happy path http_client_config river format somewhere rather than copy paste in every test file. Same thought for tls_config and maybe the others.

component/prometheus/scrape/scrape_test.go
```diff
@@ -9,12 +9,71 @@ import (
...
+
+	oauth2 {
+		client_id = "client_id"
+		client_secret = "client_secret"
+		client_secret_file = "/path/to/file.oath2"
+		scopes = ["scope1", "scope2"]
+		token_url = "token_url"
+		endpoint_params = {"param1" = "value1", "param2" = "value2"}
+		proxy_url = "http://0.0.0.0:11111"
+		tls_config = {
```
erikbaranowski, line 52: this is an attribute here and block below. That doesn't feel right. should I update to block for both?

docs/sources/flow/reference/components/prometheus.scrape.md
```diff
@@ -54,38 +54,40 @@ Name | Type | Description | Default | Required
 `label_limit`              | `uint`     | More than this many labels post metric-relabeling causes the scrape to fail. | | no
 `label_name_length_limit`  | `uint`     | More than this label name length post metric-relabeling causes the scrape to fail. | | no
 `label_value_length_limit` | `uint`     | More than this label value length post metric-relabeling causes the scrape to fail. | | no
+`bearer_token`             | `secret`   | Bearer token to authenticate with. | | no
```
erikbaranowski, line 57: copy pasting this part of the table for http_client_config on every doc that is is squashed seems like a maintenance headache. Any thoughts on an alternative?

component/prometheus/scrape/scrape_test.go
rfratto: It seems like it was a mistake for this to be a block, and we probably document it as if it's supposed to be a block. Changing it makes sense.

docs/sources/shared/flow/reference/components/authorization-block.md
```diff
@@ -7,7 +7,7 @@ headless: true
 Name | Type | Description | Default | Required
 ---- | ---- | ----------- | ------- | --------
 `type` | `string` | Authorization type, for example, "Bearer". | | no
-`credential` | `secret` | Secret value. | | no
+`credentials` | `secret` | Secret value. | | no
```
erikbaranowski, line 10: this was documented incorrectly compared to the river type. Alternatively, I can update the river type to match the docs if we prefer `credential`

docs/sources/shared/flow/reference/components/http-client-config-block.md
```diff
@@ -10,7 +10,7 @@ Name | Type | Description | Default | Required
 `bearer_token_file` | `string` | File containing a bearer token to authenticate with. | | no
 `proxy_url` | `string` | HTTP proxy to proxy requests through. | | no
 `follow_redirects` | `bool` | Whether redirects returned by the server should be followed. | `true` | no
-`enable_http_2` | `bool` | Whether HTTP2 is supported for requests. | `true` | no
+`enable_http2` | `bool` | Whether HTTP2 is supported for requests. | `true` | no
```
erikbaranowski, line 13: this was documented incorrectly compared to the river type. Alternatively, I can update the river type to match the docs if we prefer `enable_http_2`

docs/sources/flow/reference/components/prometheus.scrape.md
```diff
@@ -54,38 +54,40 @@ Name | Type | Description | Default | Required
...
+`proxy_url`                | `string`   | HTTP proxy to proxy requests through. | | no
+`follow_redirects`         | `bool`     | Whether redirects returned by the server should be followed. | `true` | no
+`enable_http2`             | `bool`     | Whether HTTP2 is supported for requests. | `true` | no
+
+Only 1 of the following arguments or blocks can be provided
+- `bearer_token`
+- `bearer_token_file`
+- `basic_auth`
+- `authorization`
+- `oauth2`
```
rfratto, line 68: This feels like maybe the wrong place to put this. Should this be documented in the blocks section instead?

docs/sources/flow/reference/components/prometheus.scrape.md
```diff
@@ -54,38 +54,40 @@ Name | Type | Description | Default | Required
 `label_limit`              | `uint`     | More than this many labels post metric-relabeling causes the scrape to fail. | | no
 `label_name_length_limit`  | `uint`     | More than this label name length post metric-relabeling causes the scrape to fail. | | no
 `label_value_length_limit` | `uint`     | More than this label value length post metric-relabeling causes the scrape to fail. | | no
+`bearer_token`             | `secret`   | Bearer token to authenticate with. | | no
+`bearer_token_file`        | `string`   | File containing a bearer token to authenticate with. | | no
+`proxy_url`                | `string`   | HTTP proxy to proxy requests through. | | no
+`follow_redirects`         | `bool`     | Whether redirects returned by the server should be followed. | `true` | no
+`enable_http2`             | `bool`     | Whether HTTP2 is supported for requests. | `true` | no
+
+Only 1 of the following arguments or blocks can be provided
```
rfratto, line 63:
```suggestion
At most one of the following can be provided:
```

docs/sources/flow/reference/components/prometheus.scrape.md
rfratto, line 68:
```suggestion
- [`bearer_token` argument](#Arguments).
- [`bearer_token_file` argument](#Arguments). 
- [`basic_auth` block][basic_auth].
- [`authorization` block][authorization].
- [`oauth2` block][oauth2].
```

component/prometheus/scrape/scrape_test.go
rfratto: Yeah, I'm not sure honestly. Testing shared config in a common place might not be necessary. 

I don't disagree that we should have some kind of unmarshaling test, but I'm not sure of the value in testing every possible field, especially for components like these where specific blocks and arguments are mutually exclusive.

component/prometheus/scrape/scrape_test.go
rfratto: Changing this would technically be a breaking change, but if our documentation says that it's supposed to be block, then we can list it as a bug fix IMO.

rfratto (commented):
Nice, LGTM

CHANGELOG.md
```diff
@@ -16,6 +16,9 @@ Main (unreleased)
   - `diskstats_ignored_devices` is now `diskstats_device_exclude` in agent configuration.
   - `ignored_devices` is now `device_exclude` in flow configuration.
 
+- A number of components were updated now that type squashing is implemented in river. Please review the component docs.
```
erikbaranowski, line 19: this wording is a placeholder for better words on how we want to tell users about the squashing changes. I'm open to suggestion.

component/prometheus/scrape/scrape_test.go
erikbaranowski: see my blind stab at the changelog. I am wondering if we want a blanket statement on these acknowledging breaking changes to config in general or something more specific for each component.

CHANGELOG.md
```diff
@@ -16,6 +16,9 @@ Main (unreleased)
...
+  - `prometheus.scrape` (@erikbaranowski)
```
rfratto, line 20: IMO, this should be changed to talk more about why we're doing it from the user's perspective: 

```suggestion
- Some blocks in Flow components have been merged with their parent block to make the block hierarchy smaller:
  - `prometheus.scrape > client > http_client_config` is merged into the `client` block. (@erikbaranowski)
```

We also need to write a section for this in the upgrade guide.

docs/sources/shared/flow/reference/components/http-client-config-block.md
rfratto: No, I think `enable_http2` is probably what was intended.

docs/sources/shared/flow/reference/components/authorization-block.md
rfratto: No, I think `credentials` is probably what was intended.

docs/sources/flow/reference/components/prometheus.scrape.md
rfratto: What about a new shared doc which holds the common rows in the table when the HTTP client config is squashed?

component/prometheus/scrape/scrape_test.go
erikbaranowski: updated and documented as a bug fix for all impacted components

erikbaranowski:
I think I may have butchered the shared markdown but I'm looking for an assist on the best way to do that part @karengermond . Robert also commented on the location of 1 related section we could use an opinion on how best to do it.

There will be a number more like this with the same http_client_config so once we get the pattern here we can repeat.

rfratto (approved):
LGTM!

component/prometheus/scrape/scrape.go
```diff
@@ -99,7 +99,12 @@ func (arg *Arguments) UnmarshalRiver(f func(interface{}) error) error {
 	*arg = DefaultArguments
 
 	type args Arguments
-	return f((*args)(arg))
+	err := f((*args)(arg))
+	if err != nil {
+		return err
+	}
+
+	return arg.HTTPClientConfig.Validate()
```
rfratto, line 107: Can you call out in a comment here that squashed blocks don't have UnmarshalRiver called, so we need to manually call the validate method?

erikbaranowski: all set

docs/sources/flow/reference/components/prometheus.scrape.md
```diff
@@ -54,38 +54,29 @@ Name | Type | Description | Default | Required
 `label_limit`              | `uint`     | More than this many labels post metric-relabeling causes the scrape to fail. | | no
 `label_name_length_limit`  | `uint`     | More than this label name length post metric-relabeling causes the scrape to fail. | | no
 `label_value_length_limit` | `uint`     | More than this label value length post metric-relabeling causes the scrape to fail. | | no
+{{< docs/shared lookup="flow/reference/components/http-client-config-block.md" source="agent" >}}
```
karengermond, line 57: This included  block is messing up how the preceding table is rendering.  Just the names are showing and not columns 2 - 4. I think you need a blank line and also the "authorization block" header as in: https://grafana.com/docs/agent/latest/flow/reference/components/prometheus.remote_write/#authorization-block

karengermond (approved):
LGTM with the exception of a table rendering issue...

docs/sources/flow/reference/components/prometheus.scrape.md
erikbaranowski: I tried a new line, but I can't personally see if the table will render with the additional rows as intended. In this case the http_client_config is no longer a block section/table on its own but is now additional rows of the Arguments table.

karengermond: You run `make docs` to see how it will render.  It looks like the included file has a header row so that would prevent it from rendering the way you want.

erikbaranowski: Excellent, thank you for the pointer on running locally. This should be all set for now. We are looking into a future way to put the table data in a shared file.

PR #3010, opened 2023-02-16
river/diag: fix issue with multiline diagnostic printing

rfratto:
An extra `|` character was being accidentally injected when printing multiline diagnostics if the column on the starting range was 1.

Before the fix (notice there is an extra `|` on line 28):

    27 |
    28 | | prometheus.scrape "default" {
       |  _^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    29 | |     targets    = prometheus.integration.node_exporter.targets
    30 | |     forward_to = [prometheus.remote_write.default.receiver]
    31 | | }
       | |_^
    32 |

After the fix (notice that the extra `|` has been removed):

    27 |
    28 |   prometheus.scrape "default" {
       |  _^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    29 | |     targets    = prometheus.integration.node_exporter.targets
    30 | |     forward_to = [prometheus.remote_write.default.receiver]
    31 | | }
       | |_^
    32 |


mattdurham (approved):
LGTM

ptodev (approved):

PR #3013, opened 2023-02-16
loki.echo: new Flow component

### LINKED ISSUE #2736: flow: Better trouleshooting for `loki` pipelines
To troubleshoot a Loki pipeline, it might be useful to mimic Promtail's dry-run mode. 

One way to go about it would be to have a `loki.stdout` component that other components could forward entries to and allow the users to inspect the logs flowing in real time.

This has one crucial difference to Promtail's dry-run mode, where the positions file is _not_ updated, so when Promtail runs in its normal mode again, the entries will be re-read and passed through the pipeline. If we wanted to also add support for this, it would need to be at the tailers level, eg. with a new Argument on the `loki.source.file` component.

rfratto:
Introduce a new Flow component, `loki.echo`, which accepts logs from other `loki` components and writes the entries to stdout.

Closes #2736. 

Credit to @tpaschalis for the original component, I just added the changelog entry and the documentation. 

## Previous renaming discussion 

This component was originally called `loki.stdout`, and included the following discussion:

While I think this component is immediately useful, especially for debugging, I'm a little worried about its longevity. I can imagine that it might be useful to explicitly specify a file to write to (like `/dev/stdout` or `/dev/stderr`), or even have the component append to a real file on disk. 

If we want to be able to do the above, then I'm not sure `loki.stdout` is the best name; maybe `loki.debug`, `loki.echo`, or `loki.tee` (if it was possible to write to more than one file)? 

As noted by @tpaschalis in #2736, the positions file would get updated when using `loki.stdout`. One alternative way to handle this in the future would be to change the semantics of when the positions file gets updated, where the positions file only gets updated for sent entries, and sending an entry returns a signal for whether the positions file should be updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@ Main (unreleased)
 
   - `loki.source.docker` reads logs from Docker containers and forwards them to
     other `loki` components. (@tpaschalis)
-  - `loki.stdout` writes received logs to stdout. (@tpaschalis, @rfratto)
+  - `loki.echo` writes received logs to stdout. (@tpaschalis, @rfratto)
   - `otelcol.processor.tail_sampling` samples traces based on a set of defined
     policies from `otelcol` components before forwarding them to other
     `otelcol` components. (@erikbaranowski)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,12 +27,17 @@ Main (unreleased)
   - `gcp` (@kgeckhart, @ferruvich)
 
 - New Grafana Agent Flow components:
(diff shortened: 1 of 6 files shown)


karengermond (approved):
LGTM!

tpaschalis:
Damn, I spent more time than I'd like to admit considering the name and I'm still not 100% on it. 😅 

Bottom line is, I would prefer if the name did not lock us into a specific feature set; no matter the name, the component can start by just logging entries by default but might be able to pipe out to one or more different destinations, or perform filtering in the future.

So my 2c is let's go with `loki.echo for now. Then we could add a new Argument and an abstraction that allows to change how the entries are handled, defaulting to just logging them out but allows appending them to a file. WDYT?

PR #3025, opened 2023-02-17
feat: adding template support into the configmap

ts-mini:
#### PR Checklist

- [x] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

## Context
[Due to previous conversations in this PR](https://github.com/grafana/agent/pull/2922) I propose the following change.  This allows us/users to provide references `{{ .Values.someAbitraryKey }}` inside their configMap string-block that will interpolate based on values provided rather than mergeOverwrite with a structured block.  

One specific use case this enables is the use of `helm secrets` whose secret values (integrations/basic, basic auth, etc) can be stored in a separate (encrypted) file but used inside the string configuration block

### DIFF
--- a/operations/helm/charts/grafana-agent/CHANGELOG.md
+++ b/operations/helm/charts/grafana-agent/CHANGELOG.md
@@ -9,6 +9,9 @@ internal API changes are not present.
 
 Unreleased
 ----------
+### Enhancements
+- Helm chart: Add support for templates inside of configMap.content (@ts-mini)
+
 
 0.6.0 (2023-02-13)
 ------------------
--- a/operations/helm/charts/grafana-agent/README.md
+++ b/operations/helm/charts/grafana-agent/README.md
@@ -28,7 +28,7 @@ use the older mode (called "static mode"), set the `agent.mode` value to
 
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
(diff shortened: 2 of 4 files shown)


rfratto:
Ah, interesting, this seems like a good idea. 

I'm not super familiar with the Helm ecosystem; do you have a sense for how common doing this kind of templating is?

ts-mini:
> Ah, interesting, this seems like a good idea.
> 
> I'm not super familiar with the Helm ecosystem; do you have a sense for how common doing this kind of templating is?

So it is actually super common - even internal to grafana charts.  Here are some links below
https://github.com/grafana/loki/blob/main/production/helm/loki/values.yaml#L80
https://github.com/grafana/loki/blob/main/production/helm/loki/values.yaml#L101
https://github.com/grafana/mimir/blob/main/operations/helm/charts/mimir-distributed/templates/_config-render.tpl#L1

[and `bitnami` actually has in in their `common` chart which is used ALL OVER](https://github.com/bitnami/charts/blob/main/bitnami/common/templates/_tplvalues.tpl)

rfratto:
Great, good to know it's common so it won't be surprising to people. Can you update the doc comment for `.Values.agent.configMap.content` to note that it's a templated string?

rfratto (approved):
Sorry for the delay on the review, we've been busy preparing for an upcoming release.

LGTM, thanks!

PR #3028, opened 2023-02-18
river/printer: fix formatting issues with comments

### LINKED ISSUE #3027: Flow: `agent fmt` uses wrong indentation for comments in arrays
Given the following file:

```river
attr = [
	1,
	// Hello!
	2,
]
```

`agent fmt` will move the comment to be aligned with the indentation level of the attribute instead of the elements:

```river 
attr = [
	1,
// Hello!
	2,
]
```

Expected behavior: the comment is indented as the same level as an element in the array.

rfratto:
* Fix issue where adding an inline comment on the same line as a `[` or `{` character would cause indentation issues on subsequent lines.

  This fix involves making sure that wsFormfeed (which starts a new row run with different alignment) is used correctly over wsNewline.

  Also, if writing a comment drops a wsFormfeed, the formfeed character will now be injected back into the resulting config to make sure that desired alignment doesn't get dropped.

* Fix issue where adding a sole line comment interspersed between array elements would be given the wrong indentation.

  This was fixed by making sure indentation is set in the array prior to adding in characters, so when the comment is injected the proper indentation is set.

I'm really not happy with the state of river/printer; it's a bit hard to reason about. I'm hoping in the future we'll have an opportunity to fix the cruft to make it easier to maintain.

Closes #3027.


slim-bean (approved):
sick

slim-bean (approved):
no, i really like this, merge it

CHANGELOG.md
```diff
@@ -62,6 +62,13 @@ Main (unreleased)
 
 - Flow: add timeout to loki.source.podlogs controller setup. (@polyrain)
 
+- Flow: fix issues in `river fmt` where adding an inline comment on the same
+	line as a `[` or `{` would cause indentation issues on subsequent lines.
+	(@rfratto)
+
+- Flow: fix issues in `river fmt` where line comments in arrays would be given
+	the wrong identation level. (@rfratto)
```
rfratto, line 70:
```suggestion
- Flow: fix issues in `river fmt` where adding an inline comment on the same
  line as a `[` or `{` would cause indentation issues on subsequent lines.
  (@rfratto)

- Flow: fix issues in `river fmt` where line comments in arrays would be given
  the wrong identation level. (@rfratto)
```

cmd/grafana-agent/repr.river
```diff
@@ -1,39 +0,0 @@
-// This config file is designed to send traces and metrics to the docker
```
rfratto, line 1: I definitely committed this file by mistake in the past.

tpaschalis (approved):

PR #3030, opened 2023-02-20
Ensure server LogLevel and LogFormat have defaults, pointer approach (#2607)

### LINKED ISSUE #2607: Empty YAML server stanza causes debug-level logging
If `agent` is configured with an empty `server` stanza in YAML:

```yml
server:
metrics:
...

```

Then `log_level` and `log_format` will lose their defaults and be assigned empty strings, as shown by the `-/config` endpoint:

```yml
server:
  log_level: ""
  log_format: ""
  grpc_tls_config: {
...
```

The [default values are supposed to be](https://grafana.com/docs/agent/latest/configuration/server-config/) `info` and `logfmt`.

For the `log_level`, unfortunately an unknown level is mapped to "all" by the logging library; this generates *significant amounts* of unwanted log output. 
https://github.com/cortexproject/cortex/blob/v1.11.0/pkg/util/log/log.go#L124

Ironically, I triggered this issue as part of upgrading my config from `v0.24`:
```diff
 server:
-  http_listen_address: '{{ ansible_host }}'
-  http_listen_port: 9090
metrics:
``` 

Adding any other entry to the stanza alleviates the issue: 

```yml
server:
  http_tls_config:
```

I stepped through and thought about it for a bit, and I think that in the case of just `server:`, the `config.Config.Server` is getting Golang-default-initialized, which puts empty strings in those fields, whereas in the `server: http_tls_config:` case, it is specifically `config.Config.Server.HTTP.TLSConfig` that is getting default-initialized.  And of course if no `server:` is specified, there is nothing overwritten.

I think the best course of action is to specifically check for default-initialization of those fields, since Golang want `""` and Agent wants `info / logfmt`.   It is unlikely an operator would intentionally make blank and it's invalid anyway.

neomantra:
#### PR Description

This is an alternative to PR #2608, which takes the approach suggested by @tpaschalis of changing the member `config.Config.Server` to a pointer type.

#2607 

#### Notes to the Reviewer

This passes `TestConfig_EmptyServerConfig`.   Most of the changes are how values/pointers of `server.Config` are either referenced or dereferenced.  Most importantly, I did not reason about whether this affects how values are copied / mutated / preserved.

#### PR Checklist

- [X] CHANGELOG updated
- [X] Documentation added
- [X] Tests updated


neomantra:
Updated per comments in #2608.  The test is now `TestConfig_EmptyServerConfigFails` and `config.Config.Validate` returns an error if `config.Server` is `nil`.

The test shows it, but an empty configuration doesn't continue improperly, which is of course the whole point :) :
```
> ./build/grafana-agent -config.file=./pkg/config/testdata/server_empty.yml
2023/02/20 16:32:34 error in config file: an empty server config is invalid
```

tpaschalis:
Hey, thanks for the effort! Could you also check the small thing around the Windows test that CI is complaining about?

neomantra:
OK, I see that now and will fix it.  In general, is there a cross-compile `make` target I should use to check before pushing?  I've tried a few obvious targets and they don't work on my host.

tpaschalis:
The CI will check the cross-compilation, you shouldn't have to do anything ^^

tpaschalis (approved):
Thanks!

PR #3036, opened 2023-02-21
Avoid goroutine leaks in loki.process, loki.source.file components

tpaschalis:
#### PR Description
This PR fixes the issue where goroutines were leaking in the loki.process and loki.source.file components due to a) not properly shutting down the channels used to Wrap around a pipeline and b) not shutting down the positions file.

I originally believed that the input/output channels for the pipeline would be simply garbage collected, but I should have read it more closely.

No issue filed.

#### Notes to the Reviewer
This is a bit tedious and error prone. Should we make it easier for component authors by changing the types we ported over from Promtail in the `common/loki` package?

#### PR Checklist

- [ ] CHANGELOG updated (N/A)
- [X] Tests updated

### DIFF
--- a/component/loki/process/process.go
+++ b/component/loki/process/process.go
@@ -42,12 +42,13 @@ var (
 type Component struct {
 	opts component.Options
 
-	mut        sync.RWMutex
-	receiver   loki.LogsReceiver
-	fanout     []loki.LogsReceiver
-	processIn  chan<- loki.Entry
-	processOut chan loki.Entry
-	stages     []stages.StageConfig
+	mut          sync.RWMutex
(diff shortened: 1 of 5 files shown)


component/loki/process/process.go
```diff
@@ -71,6 +72,12 @@ func New(o component.Options, args Arguments) (*Component, error) {
 
 // Run implements component.Component.
 func (c *Component) Run(ctx context.Context) error {
+	defer func() {
+		if c.entryHandler != nil {
```
rfratto, line 76: You need to grab the mutex here

tpaschalis: 🤦 thanks.

PR #3048, opened 2023-02-21
loki.source.kubernetes_events: new component

### LINKED ISSUE #2321: Flow: add eventhandler integration component
Add a component for the `integrations-next` [eventhandler integration](https://grafana.com/docs/agent/latest/configuration/integrations/integrations-next/eventhandler-config/).

This requires logging support to be added to Flow first.

The `eventhandler` name doesn't make a ton of sense, and doesn't imply what it does. This may be a good opportunity to rename it to something a little more self-explanatory like `loki.source.kubernetes_events`.

rfratto:
`loki.source.kubernetes_events` is the equivalent to static mode's `eventhandler` integration (integrations-next) only. It watches for events in Kubernetes and converts them into log lines.

`loki.source.kubernetes_events` emits log lines formatted exactly as the `eventhandler` integration does, but its configuration has changed to align more closely with `loki.source.kubernetes` and `loki.source.podlogs`.

Closes #2321.

You can easily test this component locally with the following pipeline:

```river 
loki.source.kubernetes_events "example" {
	namespaces = [] // Empty array means all namespaces
	forward_to = [loki.echo.example.receiver]

	// Remove for in-cluster config.
	client {
		kubeconfig_file = env("HOME") + "/.kube/config"
	}
}

loki.echo "example" { }
```

This will use your current Kubernetes config context for tailing events.


CHANGELOG.md
```diff
@@ -40,9 +40,12 @@ Main (unreleased)
 
 - New Grafana Agent Flow components:
 
+  - `loki.echo` writes received logs to stdout. (@tpaschalis, @rfratto)
   - `loki.source.docker` reads logs from Docker containers and forwards them to
     other `loki` components. (@tpaschalis)
-  - `loki.echo` writes received logs to stdout. (@tpaschalis, @rfratto)
+  - `loki.source.kubernetes_events` watches for Kubernetes Events and converts
+    them into log lines to forward to other `loki` components. It is the
+    equivalent of the `eventhandler` integration. (@rfratto)
```
rfratto, line 48: Doing some re-ordering of the names here :)

ptodev (approved):

component/loki/source/kubernetes_events/kubernetes_events.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+	return c, nil
+}
+
+// Run implements component.Component.
+func (c *Component) Run(ctx context.Context) error {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	defer c.runner.Stop()
```
tpaschalis, line 141: Let's also defer a call to c.positions.Stop() to aboid leaking a goroutine.

component/loki/source/kubernetes_events/kubernetes_events.go
```diff
@@ -0,0 +1,258 @@
...
+				if err := c.runner.ApplyTasks(ctx, tasks); err != nil {
+					level.Error(c.log).Log("msg", "failed to apply event watchers", "err", err)
+				}
+			}
+		}
+	}, func(_ error) {
+		cancel()
+	})
+
+	// Runner to forard received logs.
```
tpaschalis, line 165:
```suggestion
	// Runner to forward received logs.
```

component/loki/source/kubernetes_events/event_controller.go
```diff
@@ -0,0 +1,306 @@
...
+		Labels: lset,
+	}
+
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case ctrl.handler.Chan() <- entry:
+		// Update position offset only after it's been sent to the next set of
+		// components.
+		ctrl.task.Positions.Put(ctrl.positionsKey, "", eventTs.UnixMicro())
```
tpaschalis, line 224: Should we also update initTimestamp here? Currently it's being read from the positions file only on task startup.

tpaschalis (approved):
LGTM, nice use of pkg/runner indeed!

component/loki/source/kubernetes_events/event_controller.go
rfratto: That's on purpose :) From a comment at the start of handleEvent:

> We use a best-effort attempt to not re-deliver any events we've already logged by checking the timestamp from when the worker started. 

Events don't have a guaranteed order, so updating initPosition as we receive events could cause us to drop many, many more events.

tpaschalis: Oh, I see!

PR #3051, opened 2023-02-21
fix indentation errors in metrics section

robertvandervoort:
I haven't tested the entire config but the indentation was incorrect and my agent would not start. We should prob double check these as well as the other sections mentioned in this page for proper indentation. YAML is a fickle beast.

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/configuration/create-config-file.md
+++ b/docs/sources/configuration/create-config-file.md
@@ -79,9 +79,9 @@ metrics:
   configs:
     - name: agent
       scrape_configs:
-        # PASTE scrape_configs SECTION HERE
-      remote_write:
-        # PASTE remote_write SECTION HERE
+      # PASTE scrape_configs SECTION HERE
+    remote_write:
+    # PASTE remote_write SECTION HERE
 ```
 
 For example, this configuration file configures the Grafana Agent to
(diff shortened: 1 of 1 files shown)


docs/sources/configuration/create-config-file.md
```diff
@@ -79,9 +79,9 @@ metrics:
   configs:
     - name: agent
       scrape_configs:
-        # PASTE scrape_configs SECTION HERE
-      remote_write:
-        # PASTE remote_write SECTION HERE
+      # PASTE scrape_configs SECTION HERE
+    remote_write:
+    # PASTE remote_write SECTION HERE
```
rfratto, line 84: Hm, this doesn't look quite right, `remote_write` is supposed to be a field inside of the array element; a neighbor to scrape_configs.

robertvandervoort: you can have a remote write in each scrape job, also globally. What is breaking this example was the indent. I beat on it for an hour and had Heds and Eamon look at it and they said check the indentation. I then messed with that until it worked. Attached is my WORKING agent config file so you can see it in context.

PS: I had to add a .txt to the end in order to upload it to Github.

[grafana-agent.yaml.txt](https://github.com/grafana/agent/files/10865880/grafana-agent.yaml.txt)

robertvandervoort: Attached is the agent config yaml so you can see it in full context.
[grafana-agent.yaml.txt](https://github.com/grafana/agent/files/10865908/grafana-agent.yaml.txt)

rfratto: This suggestion doesn't match what you're doing in your configuration. `remote_write` should not have the same indentation as the `-`, it should stay where it is: indented deeper than the `-` to signal that it's a field in the element of that array.

PR #3052, opened 2023-02-22
Add retries for Phlare write component

cyriltovena:
#### PR Description

This PR adds retries but also metrics for the Phlare write component.
Retries used the same backoff logic as other components like Loki. Not all failure are retried e.g 400 are not.
This is required since Phlare now has rate limiting.

#### Notes to the Reviewer

I've also tweaked the RemoteTimeout from 30s to 10s. Which in my opinion was too high.

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated


CHANGELOG.md
```diff
@@ -77,6 +77,8 @@ Main (unreleased)
 
 ### Enhancements
 
+- Flow: Add retries with backoff logic to Phlare write component. (@ctovena)
```
rfratto, line 80: Can you merge/rebase on top of main and move this to the unreleased enhancements section?

component/phlare/write/write.go
```diff
@@ -60,17 +61,23 @@ func (rc *Arguments) UnmarshalRiver(f func(interface{}) error) error {
...
-	Headers          map[string]string        `river:"headers,attr,optional"`
-	HTTPClientConfig *config.HTTPClientConfig `river:",squash"`
+	Name              string                   `river:"name,attr,optional"`
+	URL               string                   `river:"url,attr"`
+	RemoteTimeout     time.Duration            `river:"remote_timeout,attr,optional"`
+	Headers           map[string]string        `river:"headers,attr,optional"`
+	HTTPClientConfig  *config.HTTPClientConfig `river:",squash"`
+	MinBackoff        time.Duration            `river:"min_backoff_period,attr,optional"`  // start backoff at this level
+	MaxBackoff        time.Duration            `river:"max_backoff_period,attr,optional"`  // increase exponentially to this level
+	MaxBackoffRetries int                      `river:"max_backoff_retries,attr,optional"` // give up after this many; zero means infinite retries
```
rfratto, line 71: Should this be in a block instead of three flat attributes?  

```river
backoff {
  min_period  = ... 
  max_period  = ... 
  max_retries = ... 
}
```

(IMO the answer is yes if grouping them into a block will help guide documentation for the user).

component/phlare/write/write.go
```diff
@@ -202,6 +244,30 @@ func (f *fanOutClient) Push(ctx context.Context, req *connect.Request[pushv1.Pus
 	return connect.NewResponse(&pushv1.PushResponse{}), nil
 }
 
+func shouldRetry(err error) bool {
+	if err == context.DeadlineExceeded {
+		return true
+	}
```
rfratto, line 250: Should this be `errors.Is(err, context.DeadlineExceeded)` to handle wrapped errors? If not, can you add a comment to explain why it shouldn't so we don't accidentally swap it out?

rfratto (approved):

karengermond (approved):
LGTM!

component/phlare/write/write.go
cyriltovena: It's not for Loki so I went for consistency

cyriltovena:
Flakky tests on kafka/loki/

component/phlare/write/write.go
rfratto: Ok, SGTM

cyriltovena: but may be we should consider it for both ?

PR #3055, opened 2023-02-22
Loki process contention and channel issues

### LINKED ISSUE #3039: Flow: Fix `loki.process` deadlock and `loki.file` channel issue
There is a deadlock in loki.process and a channel was getting overwritten when using targets.

mattdurham:
#### PR Description

This fixes a channel contention in loki.process and also fixes a channel issue with loki.file that has multiple targets.

Closes #3039

#### Notes to the Reviewer

Going to work on tests to see if I can find a way to reproduce easily.

#### PR Checklist

- [X] CHANGELOG updated
- [NA] Documentation added
- [ ] Tests updated

### DIFF
--- a/component/loki/process/process.go
+++ b/component/loki/process/process.go
@@ -60,6 +60,7 @@ func New(o component.Options, args Arguments) (*Component, error) {
 	// Create and immediately export the receiver which remains the same for
 	// the component's lifetime.
 	c.receiver = make(loki.LogsReceiver)
+	c.processOut = make(loki.LogsReceiver)
 	o.OnStateChange(Exports{Receiver: c.receiver})
 
 	// Call to Update() to start readers and set receivers once at the start.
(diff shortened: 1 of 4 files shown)


component/loki/source/file/file.go
```diff
@@ -182,7 +182,6 @@ func (c *Component) Update(args component.Arguments) error {
 		}
 
 		c.reportSize(path, labels.String())
-		c.handler = make(loki.LogsReceiver)
```
tpaschalis, line 185: Each time we enter Update, the entryHandler will be stopped and c.handler will be closed. Don't we want to re-create it here?

component/loki/process/process.go
```diff
@@ -124,7 +105,6 @@ func (c *Component) Update(args component.Arguments) error {
 		if err != nil {
 			return err
 		}
-		c.processOut = make(chan loki.Entry)
```
tpaschalis, line 127: Same here, we might close c.processOut when calling Update. Don't we want to re-create it here?

component/loki/source/file/file.go
mattdurham: We probably shouldn't close the handler until the component leaves, let me take a look.

tpaschalis: If not, then we can remove it from the stop func. The test will show if we have a goroutine leak.

mattdurham: Moved the close to defer on the run method. Which should work since its only created once.

component/loki/process/process.go
mattdurham: Moved the close to defer on the run method. Which should work since its only created once.

mattdurham: I don't think processOut should never change during the lifecycle of the process component. Outs the output channel for the ll the ApiEventHandler channels/processes.

mattdurham:
Ran through some additional tests and things look good.

tpaschalis (approved):

PR #3071, opened 2023-02-23
ci: have crosscompile jobs build with journald support

rfratto:
When working on the v0.32.0-rc.0 release, we noticed the 32-bit ARMv6 build is broken and preventing us from releasing.

It seems that removing the embedded Flow UI for ARMv6 restores the builds. Since it doesn't seem like anyone on 32-bit ARM is using Flow, we are temporarily removing support while we wait for Go 1.20.2 which fixes the linker issue we're running into.

### DIFF
--- a/.drone/drone.yml
+++ b/.drone/drone.yml
@@ -343,8 +343,7 @@ platform:
 steps:
 - commands:
   - make generate-ui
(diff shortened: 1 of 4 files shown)


mattdurham (approved):
LGTM

erikbaranowski (approved):
LGTM

PR #3072, opened 2023-02-23
Temporarily remove support for 32-bit ARM builds

rfratto:
When working on the v0.32.0-rc.0 release, we noticed the 32-bit ARM builds are broken and preventing us from releasing. 

A fix for 32-bit ARM builds is pending the release of Go 1.20.2, so we have to temporarily pull support for these builds so we can continue releasing.

### DIFF
--- a/.drone/drone.yml
+++ b/.drone/drone.yml
@@ -366,36 +366,6 @@ trigger:
 type: docker
 ---
 kind: pipeline
-name: Build agent (Linux armv6)
-platform:
-  arch: amd64
-  os: linux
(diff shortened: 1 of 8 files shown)


rfratto:
After discussing offline, the builds will still work if we take out the Flow UI from 32-bit ARM. Since it seems like there are no Flow users currently on 32-bit ARM, removing just the UI instead of the set of builds seems like a good temporary workaround while we're waiting for the Go patch to release.

Closing in favor of another PR.

PR #3087, opened 2023-02-24
kubetail: add a max lifetime for tailing logs from a container

### LINKED ISSUE #3093: Flow: no log lines when using loki.source.kubernetes or loki.source.podlogs
In my test Kubernetes environment, I've noticed that log lines disappear after a few hours when using `loki.source.kubernetes`. 

I've tracked this down to a network connectivity issue with the connection to the Kubernetes API, where the component is waiting forever for a new log line to appear. 

Since `loki.source.kubernetes` and `loki.source.podlogs` share code for retrieving log lines, this bug affects both.

rfratto:
This adds a maximum lifetime (1 hour) for each tailing connection to a container. This fixes an issue where the connection between the agent, the Kubernetes API server, and the Kubelet can get disrupted without closing the TCP connection, leading to no logs being read.

This bug fix impacts `loki.source.kubernetes` and `loki.source.podlogs`, which both use the kubetail package. 

Closes #3093.

### DIFF
--- a/component/loki/source/kubernetes/kubetail/tailer.go
+++ b/component/loki/source/kubernetes/kubetail/tailer.go
@@ -79,7 +79,7 @@ func newLabelSet(l labels.Labels) model.LabelSet {
 }
 
 var retailBackoff = backoff.Config{
-	MinBackoff: time.Second,
+	MinBackoff: 0,
 	MaxBackoff: time.Minute,
 }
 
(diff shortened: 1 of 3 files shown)


component/loki/source/kubernetes/kubetail/tailer.go
```diff
@@ -79,7 +79,7 @@ func newLabelSet(l labels.Labels) model.LabelSet {
 }
 
 var retailBackoff = backoff.Config{
-	MinBackoff: time.Second,
+	MinBackoff: 0,
```
rfratto, line 82: I set this back to zero so the very first retry is instantaneous, which is useful to make sure that the maximum lifetime of the tailer doesn't have to wait a second before reestablishing the connection.

rfratto:
I deployed this locally and found that there are still some issues; mainly it seems like containers were falsely considered terminated after the lifetime expired. I'm still looking into this.

erikbaranowski (approved):
LGTM, nice catch on the UID!

rfratto:
I'm going to merge this and continue to run the fix in my Kubernetes cluster. If it still looks good after a couple of days, it would probably make sense to put this fix in a patch release (given the severity of the bug).

PR #3088, opened 2023-02-25
test: use `t.TempDir` to create temporary test directory

Juneezee:
#### PR Description

This pull request replaces `os.MkdirTemp` with `t.TempDir`. We can use the `t.TempDir` function from the `testing` package to create temporary directory. The directory created by `t.TempDir` is automatically removed when the test and all its subtests complete. 

This saves us at least 2 lines (error check, and cleanup) on every instance, or adds cleanup that we forgot.

Reference: https://pkg.go.dev/testing#T.TempDir

```go
func TestFoo(t *testing.T) {
	// before
	tmpDir, err := os.MkdirTemp("", "")
	require.NoError(t, err)
	defer os.RemoveAll(tmpDir)

	// now
	tmpDir := t.TempDir()
}
```

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/component/loki/source/file/file_test.go
+++ b/component/loki/source/file/file_test.go
@@ -22,9 +22,7 @@ func Test(t *testing.T) {
 	// Create opts for component
 	l, err := logging.New(os.Stderr, logging.DefaultOptions)
 	require.NoError(t, err)
-	dataPath, err := os.MkdirTemp("", "loki.source.file")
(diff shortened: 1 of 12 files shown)


rfratto (approved):
LGTM, thanks! This is much cleaner :)

PR #3089, opened 2023-02-25
Update _index.md

jzck:
idempotent means `function(function(x)) = function(x)`. I think what you mean here is [pure functions](https://en.wikipedia.org/wiki/Pure_function), similarly to terraform functions: https://developer.hashicorp.com/terraform/language/expressions/function-calls

#### PR Checklist

- [ ] CHANGELOG updated
- [x] Documentation added
- [ ] Tests updated


rfratto (approved):
Thanks!

PR #3094, opened 2023-02-27
helm: respect .image.pullPolicy from values

rfratto:
The `.image.pullPolicy` value was not being respected. 

I've also changed the documentation to no longer recommended `loki.source.kubernetes`, given the issues discovered in #3093.


mattdurham (approved):
Lgtm

PR #3101, opened 2023-02-27
add code block, and the refactored table from the previous pr

### LINKED ISSUE #2689: UI: Objects no longer indented and aligned

xiyu95:
#### PR Description
Code block view for river values using react-code-blocks

Fixes #2689 

#### Video demo
https://user-images.githubusercontent.com/112968312/221679846-32c8660c-a619-4993-9ffd-8e119c20300f.mov

#### Notes to the Reviewer
The `Table.tsx` was removed from main and back to the pre-refactored version of `<table>`. I brought it back as it is used in the `ComponentList` page (shown at 0:20 in the video above). The main change for this PR is to add the react-code-block in the component detail page to avoid the overflow of text. (shown Screen Shot 2023-02-27 at 12 40 00 PM)

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated


rfratto:
It looks like there's extra padding being introduced in each row, can that be fixed?

From demo video:
image

From current UI:
image

xiyu95:
> It looks like there's extra padding being introduced in each row, can that be fixed?

I tried reduce the padding of the table, and found that the library comes with it's own `line-height` which is making the gaps.

xiyu95:
@rfratto quick follow up. Let me know if you are okay with the spacing (description ☝️ )

rfratto:
Ideally we don't add extra padding just because of the libraries we choose to use. Can we get one of your peers to investigate?

xiyu95:
> Ideally we don't add extra padding just because of the libraries we choose to use. Can we get one of your peers to investigate?

Yea I tried removing the padding to see if it changes the spacing as well, and due to the `line-height` from the library the extra space is not able to be removed. If we don't want the extra spacing, I can try copying the component from the library instead of using the actual library and pass in full custom styling. I can try getting some ideas from my teammates during our 1:1 meetings.

L2D2Grafana:
This library is cool [react-code-blocks](https://github.com/rajinwonderland/react-code-blocks) in the fact that it mentions other libraries to use. react-code-block doesn't let you override the styles, and would need a PR to update that.

I would try this one instead [react-syntax-highlighter](https://github.com/react-syntax-highlighter/react-syntax-highlighter#props) it looks like you have control to overwrite the styles.

I tried the [highlight.js](https://github.com/bvaughn/react-highlight.js) one and the styling is easy but lost theme colors
Screen Shot 2023-03-07 at 3 07 57 PM

web/ui/src/features/widgets/Table.tsx
```diff
@@ -0,0 +1,31 @@
...
+}
+
+/**
+ * Simple table component that accept a custom header, and custom render
+ * function for the table data
+ */
+const Table = ({ tableHeaders, style = {}, renderTableData }: Props) => {
+  return (
+    <table className={styles.table}>
+      <col span={1} style={style} />
```
L2D2Grafana, line 18: I believe this is throwing an error
Screen Shot 2023-03-07 at 2 27 13 PM

xiyu95:
> I tried the [highlight.js](https://github.com/bvaughn/react-highlight.js) hmm, it doesn't let you customize the colors?

L2D2Grafana:
> > I tried the [highlight.js](https://github.com/bvaughn/react-highlight.js)
> 
> hmm, it doesn't let you customize the colors? `react-code-block` isn't very flexible in the color either, but it does let us pass in an object with pre-set field to modify some of the colors.

Right don't try `highlight.js` try [react-syntax-highlighter](https://github.com/react-syntax-highlighter/react-syntax-highlighter#props)

web/ui/src/features/widgets/Table.tsx
xiyu95: Fixed using `<colgroup>`

xiyu95:
> Right don't try `highlight.js` try [react-syntax-highlighter](https://github.com/react-syntax-highlighter/react-syntax-highlighter#props)

Thnx, will give it a try

xiyu95:
Okay so I have tried `react-syntax-highlighter`, below is the result:

https://user-images.githubusercontent.com/112968312/226470281-5f8e31e8-f676-4a88-800f-ee1d27a91e81.mov

@rfratto what do you think of this one? I've removed the spacing between each row by passing in custom style

rfratto:
@xiyu95 Much better, thanks!

xiyu95:
> @xiyu95 Much better, thanks!

Thanks! I am gonna clean up the code a bit and update with the main branch. After the code review I will merge this one.

rfratto:
One small thing I just noticed in the video:
 
image

Wasn't the header row bigger before?

xiyu95:
> One small thing I just noticed in the video:

> Wasn't the header row bigger before?

There! I adjusted around the style and forgot to put the padding for the header back. Would you like me to keep the header bold?
Screenshot 2023-03-20 at 2 38 10 PM

rfratto:
@xiyu95 Bolding it seems like a good idea, thanks!

web/ui/src/features/component/style.ts
```diff
@@ -0,0 +1,140 @@
+import React from 'react';
```
L2D2Grafana, line 1: Is this all for `react-syntax-highlighter` maybe a comment to mention it all works together or export as syntaxHighligher

web/ui/src/features/widgets/Table.module.css
```diff
@@ -0,0 +1,34 @@
+.table {
```
L2D2Grafana, line 1: I hesitate to name any folder `widget` and `component` as developers usually use these name to dump all generic reusable components into, but in this repo the folders seem to only pertain to one ui element, perhaps name it /table

L2D2Grafana (approved):
LGTM nice updates!

web/ui/src/features/component/style.ts
xiyu95: Good point. Yes, this is an entire object to pass in the `react-syntax-highlighter`, and it is pulled from the library's style object with some custom changes

web/ui/src/features/component/ComponentBody.tsx
```diff
@@ -0,0 +1,76 @@
...
+        {
+          // If the partition only has 1 key, then make it an h2.
+          // Otherwise, make it an h3.
+          partition.displayName.length === 1 ? (
+            <h2>{partition.displayName}</h2>
+          ) : (
+            <h3>
+              {partition.displayName.map((val, idx) => {
+                return (
+                  <Fragment key={idx.toString()}>
```
abannachGrafana, line 50: using an index as the key is not recommended practice. https://react.dev/learn/rendering-lists#why-does-react-need-keys See the "Pitfall" section.

web/ui/src/features/river-js/stringify.ts
```diff
@@ -0,0 +1,207 @@
...
+      case '"':
+        return '\\"';
+      case '\\':
+        return '\\\\';
+    }
+    return '';
+  });
+}
+
+function getLinePrefix(indentLevel: number): string {
```
abannachGrafana, line 114: suggest renaming to `indentLine`

web/ui/src/features/widgets/Table.tsx
```diff
@@ -0,0 +1,31 @@
...
+  tableHeaders: string[];
+  style?: React.CSSProperties;
+  renderTableData: () => JSX.Element[];
+}
+
+/**
+ * Simple table component that accept a custom header, and custom render
+ * function for the table data
+ */
+const Table = ({ tableHeaders, style = {}, renderTableData }: Props) => {
```
abannachGrafana, line 15: Do we not have a `Table` in `@grafana/ui`?

xiyu95: We do, but I find the Table from grafana ui more than necessary for a custom table, and harder to use in data conversion. I might using it in the future if there is more complex table needed in the agent app, but for now this simple Table widget will do.

abannachGrafana (approved):
LGTM

rfratto (commented):
💪 Thanks! I fixed the merge conflict so we can get this into main :)

xiyu95:
> 💪 Thanks! I fixed the merge conflict so we can get this into main :)

Thank you!

PR #3117, opened 2023-02-28
modules: Initial support for loading subgraphs.

### LINKED ISSUE #3113: Flow Modules: Add support in flow subsystem for loading modules
Add support in the Flow subsystem for loading "modules," which are configuration files encapsulated as components. 

Modules should support "arugments" which allow input to be fed to modules, and "exports" which allow values to be returned by modules.

mattdurham:
#### PR Description

This PR inherently makes no user-facing changes until modules are added in the next PR. The primary goal is to add subgraph support, I waffled on naming this submodules but there might be a world where this is used beyond modules. The primary point is to create a parent child tree of subgraphs that are loaded and unloaded on demand.

The general workflow is that when a component is created the SubgraphHandler is passed as an option, the component can use this to create a new subgraph. That component will then become the subgraph owner and a new child subgraph will be created. The subgraph.md file gives a rundown of how it will work with modules.

Most of the flow functionality moved into subgraph, it is possible we could have reused the flow.new and created full flow subsystems but I think its worth having an anchor for the main module (flow.go that spawns the main subgraph) and then child modules (subgraphs). There are likely some functions in the main module (flow.go) we would want to instantiate/control at a root level instead of pivoting in subgraphs.

This also introduces the concept of NamespaceID which is the full unique name, essentially the Parent NamespaceID + ComponentID. I could easily talked into having this just be the parent ID and manually merging the two where needed.

Fixes #3113

#### Notes to the Reviewer

This is part one of several PRs to introduce.

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/component/component.go
+++ b/component/component.go
@@ -128,6 +128,8 @@ type HTTPComponent interface {
 
 // SubgraphOwner is the component calling the SubgraphHandler.
 type SubgraphOwner interface {
+	// ID is the concatenated IDs of the owner, "module.string.mod1"
 	ID() string
+	// IDs is the separated IDs of the owner, []string{"module","string","mod1"}
 	IDs() []string
 }
--- a/component/component.go
+++ b/component/component.go
@@ -125,3 +125,9 @@ type HTTPComponent interface {
 	// will receive a request to just `/metrics`.
 	Handler() http.Handler
 }
+
+// SubgraphOwner is the component calling the SubgraphHandler.
+type SubgraphOwner interface {
+	ID() string
+	IDs() []string
+}
--- a/cmd/grafana-agent/flow_run.go
+++ b/cmd/grafana-agent/flow_run.go
@@ -151,7 +151,7 @@ func (fr *flowRun) Run(configFile string) error {
 		if err != nil {
 			return fmt.Errorf("reading config file %q: %w", configFile, err)
 		}
-		if err := f.LoadFile(flowCfg); err != nil {
+		if err := f.LoadFile(flowCfg, configFile); err != nil {
 			return fmt.Errorf("error during the initial gragent load: %w", err)
 		}
 
--- a/cmd/grafana-agent/flow_run.go
+++ b/cmd/grafana-agent/flow_run.go
@@ -287,7 +287,7 @@ func getEnabledComponentsFunc(f *flow.Flow) func() map[string]interface{} {
 	}
 }
 
-func loadFlowFile(filename string) (*flow.File, error) {
+func loadFlowFile(filename string) ([]byte, error) {
 	bb, err := os.ReadFile(filename)
 	if err != nil {
 		return nil, err
@@ -295,7 +295,7 @@ func loadFlowFile(filename string) (*flow.File, error) {
 
 	instrumentation.InstrumentConfig(bb)
 
-	return flow.ReadFile(filename, bb)
+	return bb, nil
 }
 
 func interruptContext() (context.Context, context.CancelFunc) {
--- a/component/loki/source/file/file_test.go
+++ b/component/loki/source/file/file_test.go
@@ -84,8 +84,11 @@ func TestTwoTargets(t *testing.T) {
 	if err != nil {
 		log.Fatal(err)
 	}
+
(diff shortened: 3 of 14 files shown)


component/component.go
```diff
@@ -125,3 +125,9 @@ type HTTPComponent interface {
 	// will receive a request to just `/metrics`.
 	Handler() http.Handler
 }
+
+// SubgraphOwner is the component calling the SubgraphHandler.
+type SubgraphOwner interface {
+	ID() string
+	IDs() []string
+}
```
tpaschalis, line 133: Could we add a docstring about what the methods mean? Is it the ID of the parent and the IDs of its children?

Also, is this the `DelegateComponent` interface we reference in a comment below?

rfratto (changes requested):
I'm worried about the added complexity that adding namespaces to the Flow controller (`pkg/flow`) introduces. 

In general, I'm not sure if the Flow controller is the right place to implement the bulk of this work. In my opinion, it would require less effort to have module loader components instantiate and run new Flow controllers by calling `flow.New` inside of the component. That way modules are completely isolated from the parent graph. 

My suggestion would still require some changes to the Flow controller API for a few things:

1. Ability to pass arguments to the Flow controller.
2. Ability to provide a base ID for components in the Flow controller. 
3. Find a way to have metrics from the inner Flow controller work, since the outer controller will try to override the `component_id` label.

mattdurham:
For comparison I did a quick and dirty [PR](https://github.com/grafana/agent/pull/3133/files#diff-c6dd75c886270ade02e70ce7f8bf69627616781e3bcc8495171164973c0fbc81) to use flow.new. Eyeballing the lines changes flow.new is a bit less, flow.new includes the module components but doesn't have the doc and unit test changes. Also has a few bugs that I would need to fix but it is good enough to run modules.

I think the subgraph adds some upfront complexity for an easier system to track and work with in the future. 

I believe subgraphs offer several advantages compared to flow.new:
* Reduced surface area exposed to the components
* Reduced complexity in building module components
* Easier to add features to modules subsystem since the core functionality is one place that can reach deep into the subgraph subsystem as a whole. Stack limits and the UI for example.

Many of these could be mitigated by using flow.new with additional wrapping or interfacing but then flow.new loses/ties the complexity issue.

If the particular concern is namespaces it's possible to remove those from some of the structs and only use them when creating the managed options like in the flow.new path. Inherently the namespace will exist just not as a concrete field.

rfratto:
These are my specific concerns: 

* I don't see it's how currently possible to unload subgraphs constructed by components. Right now you only unload the root subgraph.  
* We should not be using any kind of global-like value where components have the opportunity to trample over one another.  

> Reduced surface area exposed to the components

I'm not sure module loaders having to use more API functionality to work is necessary a bad thing, especially if it's safer and harder to abuse. 

> Reduced complexity in building module components

Abstractions can handle reducing the complexity. 

> Easier to add features to modules subsystem since the core functionality is one place that can reach deep into the subgraph subsystem as a whole. Stack limits and the UI for example.

This is true, though there are still ways to do both:

* Stack limits can be limited by checking the ID of the module loader and refusing to load if the path is too long. (But I'm not convinced in general that a stack limit is something we should even do)
* The UI can be handled by implementing HTTPComponent and exposing an API. 

Both of these can be handled per-component or by an abstraction to share the logic between multiple module loaders.

rfratto:
In general, my conceptual model of `pkg/flow` is that it is a controller for exactly one Flow module. I'm not convinced yet that it makes sense to expand that scope to handle the entire hierarchy of modules.

rfratto:
If we can't reach consensus on this, or disagree and commit, I think we might have to take this to a governance vote. I'm still too concerned about the drawbacks of having an effectively global declarative API for modules.

mattdurham:
I pushed some changes on unloading subgraphs and the cleanup. I also added a [PR](https://github.com/grafana/agent/pull/3141) to show modules and how they are used, I didn't include them here because was trying to piecemeal the changes in.

PR #3124, opened 2023-02-28
fix(operator/api): change /etc/grafana-agent to /var/lib/grafana-agent

GeorgeMac:
#### PR Description

Corrects the GrafanaAgent API documentation to reflect that extra `secrets` and `configmaps` added via the agent CRD will be manifested inside the pod under `/var/lib` instead of `/etc`.

At-least this appears to be what I am observing in my setup.

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated


marctc:
@GeorgeMac sorry for the late review! This document is auto generated after running `make generate-crds`, so you will need to find the comment in the code and the run this command.

Pinging here @captncraig to validate if assumption of paths is true (I don't have context for this).

captncraig:
That looks correct to me.

marctc:
@GeorgeMac do you still want to get this PR merged?

GeorgeMac:
@marctc sorry for the slow reply.

I just rebased my PR and did as you advised. There generated output actually touches a lot more than just the small comment I changed (I used the container version of `USE_CONTAINER=1 make generate-crds` to try and get something with the right dependency versions).
Do you want me to try and just pluck out the bits that are relevant, or is committing all of these changes ok?

marctc (approved):

marctc:
>  Do you want me to try and just pluck out the bits that are relevant, or is committing all of these changes ok?

Is good like this, thanks for the contribution!

Issue #3068, opened 2023-02-23
Add support for linux s390x


mattdurham:
Tested on a VM and should mainly require updating Viceroy with the appropriate build tags and creating the correct drone stages.

PR #3127, opened 2023-02-28
Update build image to viceroy v0.3.0

rfratto:
Viceroy v0.3.0 supports s390x.

Related to #3068.

I've also added Helm, which was missing to be able to run `make generate-helm-tests generate-helm-docs` via the build container without installing Helm or helm-docs.


mattdurham (approved):
LGTM

PR #3128, opened 2023-02-28
agentlint: alert when a Flow component uses the wrong field tag

rfratto:
This adds a linting rule where if a struct has at least one field with a `river` tag, an error is emitted if there is a field with `json`/`yaml` tags but no `river` tags. 

I fixed the one instance which fails this check.

### DIFF
--- a/tools/agentlint/internal/rivertags/rivertags.go
+++ b/tools/agentlint/internal/rivertags/rivertags.go
@@ -18,7 +18,12 @@ var Analyzer = &analysis.Analyzer{
 }
 
 var noLintRegex = regexp.MustCompile(`//\s*nolint:(\S+)`)
-var riverTagRegex = regexp.MustCompile(`river:"([^"]*)"`)
+
+var (
+	riverTagRegex = regexp.MustCompile(`river:"([^"]*)"`)
+	jsonTagRegex  = regexp.MustCompile(`json:"([^"]*)"`)
+	yamlTagRegex  = regexp.MustCompile(`yaml:"([^"]*)"`)
+)
 
 // Rules for river tag linting:
 //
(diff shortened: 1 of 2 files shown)


tools/agentlint/internal/rivertags/rivertags.go
```diff
@@ -63,7 +78,22 @@ func run(p *analysis.Pass) (interface{}, error) {
...
 
 			matches := riverTagRegex.FindAllStringSubmatch(s.Tag(i), -1)
-			if len(matches) == 0 {
+			if len(matches) == 0 && hasRiverTags {
+				// If this struct has River tags, but this field only has json/yaml
+				// tags, emit an error.
+				jsonMatches := jsonTagRegex.FindAllStringSubmatch(s.Tag(i), -1)
+				yamlMatches := yamlTagRegex.FindAllStringSubmatch(s.Tag(i), -1)
+
+				if len(jsonMatches) != 0 || len(yamlMatches) != 0 {
```
captncraig, line 87: a bit confusing logic. Would it be better to say `if len(jsonMatches) + len(yamlMatches) > 0` ? Maybe even just changing the `!=`s to `>` would make it read easier.

captncraig (approved):

tools/agentlint/internal/rivertags/rivertags.go
rfratto: I think this may be personal taste. I prefer the logical or statement because I read it literally as "if there are any JSON matches, or if there are any YAML matches." Changing it to a sum is a bit weirder to me. 

I can definitely change to `len(jsonMatches) > 0 || len(yamlMatches) > 0` though.

PR #3130, opened 2023-03-01
Add language around where to find default config file locations

shih-chris:
#### PR Description
I was having trouble finding where the default location for the grafana agent config file was located, particularly after installing it via an integration. I eventually found the info I needed in the Install guides, but it took some determination.

I figured that it might make sense to include a reference to it in the first place that I looked (and where Google brought me)

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/configuration/_index.md
+++ b/docs/sources/configuration/_index.md
@@ -31,6 +31,8 @@ well-documented.
 The configuration file can be reloaded at runtime using the `/-/reload` API
 endpoint or sending a SIGHUP signal to the process.
 
+For default config file locations, refer to the Install Grafana Agent guides under "Set up Grafana Agent"
+
 ## Variable substitution
 
 You can use environment variables in the configuration file to set values that


clayton-cornell:
@shih-chris Thanks for pointing this out - you're very right, the default config file locations is a bit hidden. I'm currently working on refreshing the Install and Config documentation for Flow in https://github.com/grafana/agent/pull/4232  This update includes explicitly stating where the default config files for Agent Flow are located. I plan to do a similar update for Static mode install and config docs, so I'd like to hold off on merging this PR for now so that we can implement a change that makes it even more clear where things are.

shih-chris:
> @shih-chris Thanks for pointing this out - you're very right, the default config file locations is a bit hidden. I'm currently working on refreshing the Install and Config documentation for Flow in #4232 This update includes explicitly stating where the default config files for Agent Flow are located. I plan to do a similar update for Static mode install and config docs, so I'd like to hold off on merging this PR for now so that we can implement a change that makes it even more clear where things are.

Sounds great! Really appreciate you working on it!

PR #3142, opened 2023-03-01
Improve Phlare target reloading

### LINKED ISSUE #2871: Flow: Reduce phlare.* update times
Evaluating the config for `phlare.*` components seems to take a lot longer than for the other components. This causes the "agent_component_evaluation_seconds" histogram to log much higher readings than usual. We should look into whether this update time can be reduced.

cyriltovena:
#### PR Description

This PRs improves the target reloading in `phlare.scrape` component.

We basically only wait for the scraping loop to end when shutting down the component, but not anymore when reloading targets.

This means the target will gracefully finish and terminate asynchronously.

Finally @rfratto help me figure a second bug that was caused by a lock taken while retrying push to the backend but it was required only to update the status.

Fixes #2871 

#### Notes to the Reviewer

Going to test this in dev.

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/component/phlare/scrape/manager.go
+++ b/component/phlare/scrape/manager.go
@@ -83,11 +83,6 @@ func (m *Manager) reloader() {
 }
 
 func (m *Manager) reload() {
-	now := time.Now()
-	level.Warn(m.logger).Log("msg", "phlare reloading scrape manager")
-	defer func() {
-		level.Warn(m.logger).Log("msg", "phlare reloaded scrape manager", "duration", time.Since(now))
-	}()
 	m.mtxScrape.Lock()
 	defer m.mtxScrape.Unlock()
 
--- a/component/phlare/scrape/manager.go
+++ b/component/phlare/scrape/manager.go
@@ -84,6 +84,8 @@ func (m *Manager) reloader() {
 
 func (m *Manager) reload() {
 	m.mtxScrape.Lock()
+	defer m.mtxScrape.Unlock()
+
 	var wg sync.WaitGroup
 	for setName, groups := range m.targetSets {
 		if _, ok := m.targetsGroups[setName]; !ok {
@@ -102,7 +104,6 @@ func (m *Manager) reload() {
 			wg.Done()
 		}(m.targetsGroups[setName], groups)
 	}
-	m.mtxScrape.Unlock()
 	wg.Wait()
 }
 
--- a/component/phlare/scrape/manager.go
+++ b/component/phlare/scrape/manager.go
@@ -83,6 +83,11 @@ func (m *Manager) reloader() {
 }
 
 func (m *Manager) reload() {
+	now := time.Now()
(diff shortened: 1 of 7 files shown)


component/phlare/scrape/manager.go
```diff
@@ -84,6 +84,8 @@ func (m *Manager) reloader() {
 
 func (m *Manager) reload() {
 	m.mtxScrape.Lock()
+	defer m.mtxScrape.Unlock()
```
tpaschalis, line 87: Why do we keep the lock until all target sets have been synced here? Is it so if a new reload comes in we avoid running new and old sync jobs in parallel? In any case, the scrapePools have their own locks so it might not be needed.

cyriltovena: yeah you probably right but I went for something safer.

component/phlare/scrape/scrape_loop.go
```diff
@@ -126,7 +130,7 @@ func (tg *scrapePool) stop() {
 	defer tg.mtx.Unlock()
 
 	for _, t := range tg.activeTargets {
-		t.stop()
+		t.stop(true)
```
tpaschalis, line 133: I kinda like the `wait` idea on the stop method, I might use it in the future. 

My question is, this is only called when shutting down the component (eg. removing it from the config), right? Is there a reason why we want to wait here eg. to avoid leaking any goroutines or something?

cyriltovena: My idea is that we might attempt to shutdown the component, because we restart the agent and we don;'t want to interrupt goroutines until they have properly finished their last scrape.

All of this is assuming flow wait for the component.Run to end when shutting down. When just removing the component, it's less important as the goroutine will not be forcefully exited by os.Exit.

tpaschalis: Yes, the Agent will wait for all the component's Run methods to exit when shutting down eg. when an interrupt signal is received.

Could we see the same long waiting times when exiting as we did for updates? If so, I'd be a little worried about scheduling in Kubernetes being disrupted.

cyriltovena: Yes we probably need to stop in parallel to avoid long time instead of sequential. Should this be part of the current PR ?

rfratto: That's a good suggestion @tpaschalis, but can we push that off to another PR? I'd like for this to be in so we can stop the bleeding

rfratto (approved):
LGTM, @tpaschalis are you OK with this being merged or do you have anything you'd like to see addressed?

component/phlare/scrape/scrape_loop.go
tpaschalis: Yeah, let's go ahead and merge this!

tpaschalis (approved):

cyriltovena:
Let me create an issue for that then.

PR #3145, opened 2023-03-01
Add documentation for best practices when writing Flow component docs

rfratto:
This PR adds a best practices guide for Flow component reference documentation.


rfratto:
Some of the markdown renders a little weird here; I'll look into it in the morning.

docs/developer/writing-flow-component-documentation.md
```diff
@@ -0,0 +1,512 @@
...
+backticks in the header.
+
+Block sections are similar to Arguments section, where it is composed of:
+
+1. A brief description of the block.
+2. A table of arguments supported by the block.
+3. Detailed description for how that block impacts component behavior, and how
+   the block's arguments can be used to modify that behavior.
+
+See [Arugments](#arguments) for a description of how to write the arguments
```
spartan0x117, line 331:
```suggestion
See [Arguments](#arguments) for a description of how to write the arguments
```

docs/developer/writing-flow-component-documentation.md
```diff
@@ -0,0 +1,512 @@
...
+
+If the component does not expose any debug information, the content of the
+section should be the following paragraph:
+
+```markdown
+`COMPONENT_NAME` does not expose any component-specific debug information.
+```
+
+Otherwise, write a high-level description for what debug information the
+component provides. Do not document the attributes or blocks which is exposed
```
spartan0x117, line 431:
```suggestion
component provides. Do not document the attributes or blocks which are exposed
```

docs/developer/writing-flow-component-documentation.md
```diff
@@ -0,0 +1,512 @@
...
+Values in the Type column should be a Prometheus metric type, one of `counter`,
+`gauge`, `histogram`, `native histogram`, or `summary`.
+
+A paragraph with the content "The following Prometheus metrics are exposed:"
+should always prefix the metrics table.
+
+### Examples
+
+The Examples section provides copy-and-paste River pipelines which use the
+River component. The section starts with an `h2` header called Examples. If
```
spartan0x117, line 466: nit: 
```suggestion
Flow component. The section starts with an `h2` header called Examples. If
```

docs/developer/writing-flow-component-documentation.md
```diff
@@ -0,0 +1,512 @@
...
+there is only one example, call the section Example instead.
+
+The Examples section should be composed of a brief description of each example,
+followed by the example in a code block. For example:
+
```
spartan0x117, line 471: It might be nice to add if we support adding clarifying comments inside the examples or not.

For example, do we want something like:
```
prometheus.exporter.example {
    // The exclude list will always be prioritized over the include list, so "db2" will not appear in
    // exported metrics.
    include = ["db1", "db2"]
    exclude = ["db2"]
}

Comment (rfratto) [reply docs/developer/writing-flow-component-documentation.md:471]: hmm, to me it feels like that kind of information should be put in the Arguments section (or the relevant Block section) instead of having people search for an example

Comment (spartan0x117) [reply docs/developer/writing-flow-component-documentation.md:471]: That's fair, maybe it would be good in certain circumstances to have it in both? That would mean that the information is available in Arguments/Block section, but with a concrete example to reinforce understanding as well?

File: docs/developer/writing-flow-component-documentation.md [hunk 5]
```diff
@@ -0,0 +1,512 @@
...
+should always prefix the metrics table.
+
+### Examples
+
+The Examples section provides copy-and-paste River pipelines which use the
+River component. The section starts with an `h2` header called Examples. If
+there is only one example, call the section Example instead.
+
+The Examples section should be composed of a brief description of each example,
+followed by the example in a code block. For example:
```
rfratto, line 470: I think this should be extended to say if multiple examples are being used, each example should have its own header so it can be navigated to from the sidebar.

PR #3151, opened 2023-03-02
[RFC] Agent clustering in Flow

tpaschalis:
RFC for discussing clustering in flow.


docs/rfcs/0006-clustering.md
```diff
@@ -0,0 +1,331 @@
...
+all nodes are working with the same configuration and have knowledge of each
+one of their peers in the cluster.The Sharder will be used to determine
+ownership of resources by hashing a desired value and checking the peers
+responsible for the corresponding ring token.
+
+When all nodes in the cluster have an up-to-date image of their peers, they
+will be able to independently agree to the ownership of a resource without
+having to communicate with each other, as local hashing will provide the same
+results for all nodes. For example, an agent node will be able to hash the
+label set of a target and check which of the peers is responsible for scraping
```
ptodev, line 84: Is the "label set" a collection of all the labels in a metrics endpoint? By "labels" here I mean the keys in the key value pairs?

docs/rfcs/0006-clustering.md
```diff
@@ -0,0 +1,331 @@
...
+that target. Similarly, an agent node will be able to hash the fully qualified
+component name and decide whether a component needs to be scheduled on this
+node or if another peer takes responsibility for it.
+
+On a more practical note, this clustering will most likely work with a
+[Kubernetes HPA](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/)
+so we can dynamically scale a fleet of agents. The target resource the
+autoscaler will look out for is most likely memory usage. In the future we may
```
ptodev, line 92: > most likely memory usage

I wonder what would happen if we have, say, 1 target with a huge number of data and 10 targets with very little data in their `/metrics` endpoint. In that case there'd be very little point in sharding because we cannot split this 1 big target in any way. I suppose we are making an assumption that the size of the `/metrics` payload is roughly the same for all targets, or at least that there are no big outliers? Not sure if it's worth documenting this.

Another use case could be having a much higher sampling frequency for some targets as opposed to others. The agents could be sharded in such a way that the more frequently sampled endpoints are given more resources.

rfratto:
@tpaschalis Clustering has been implemented, is it time to take this out of draft and merge it?

tpaschalis:
> @tpaschalis Clustering has been implemented, is it time to take this out of draft and merge it?

I'll do another round to clarify comments left by @ptodev and ping the team once more so we can merge this.

docs/rfcs/0006-clustering.md
tpaschalis: The label set consists of _both_ keys and their values (as every unique label set is a new resource, in this case a target.

docs/rfcs/0006-clustering.md
```diff
@@ -0,0 +1,331 @@
...
+idea of how this can work is to have something similar to how OnExportsChange
+works. This abstraction will allow components to communicate back to the
+controller whether a clustering-related Argument has changed. The controller
+will keep a record of these components and when a cluster change is detected
+these will have their Update method called (or just queued for re-evaluation).
+
+While the cluster is not in a consistent state, this might lead to temporarily
+missing or duplicated scrapes as not all nodes will have the same image of the
+cluster and agree on the distribution of tokens. As long as the cluster can
+stabilize within a time period comparable to the scraping interval, this should
```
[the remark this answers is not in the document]
tpaschalis: We'll introduce some alerts to the agent-flow-mixin. Do you have any other ideas about what we could do?

docs/rfcs/0006-clustering.md
tpaschalis: > I suppose we are making an assumption that the size of the /metrics payload is roughly the same for all targets, or at least that there are no big outliers?

Yes, your hunch is true; we're assuming that even if outliers exist, they are enough in numbers to be ~evenly distributed. The distribution will always be a little worse than hashmod sharding (as we're only distributing a certain number of tokens), but my back-of-the-envelope calculation in a realistic environment showed the stdev/mean over 4 nodes to be around 18% and not that worse than hashmod.

I've considered metrics-level sharding, but I think we shouldn't go about it right now. The target-level sharding is what we're already using with hashmod, is less resource intensive, easier to debug and seems works well enough for regular use cases.

> Another use case could be having a much higher sampling frequency for some targets as opposed to others. The agents could be sharded in such a way that the more frequently sampled endpoints are given more resources.

The scrape interval is dictated per `prometheus.scrape` component, so in this case, _all_ targets would be sampled at the same rate. Were you thinking about multiple scrape components running in parallel?

docs/rfcs/0006-clustering.md
```diff
@@ -0,0 +1,331 @@
...
+among the nodes in the cluster. The eventually consistent cluster state is when
+all nodes are working with the same configuration and have knowledge of each
+one of their peers in the cluster.The Sharder will be used to determine
+ownership of resources by hashing a desired value and checking the peers
+responsible for the corresponding ring token.
+
+When all nodes in the cluster have an up-to-date image of their peers, they
+will be able to independently agree to the ownership of a resource without
+having to communicate with each other, as local hashing will provide the same
+results for all nodes. For example, an agent node will be able to hash the
```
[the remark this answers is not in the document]
tpaschalis: > I'm not sure how this can be true in the general case

Could you expand on that? In our eventual consistent state, when all peers have knowledge of one another, their hashing function will provide the same output for the same input target set.

My current view is that we _don't_ need two set of agent clusters. Redundancy can happen by asking the same hashing function to return N owners for a certain token. This would be configured on the component itself as the clustering provides the overarching mechanism only, and similarly to N=1, no explicit communication or beforehand sharding is required.

docs/rfcs/0006-clustering.md
ptodev: > Were you thinking about multiple scrape components running in parallel?

Yes. If I understand correctly, in such case all scrape components will be sharded the same number of times. If we want to shard them differently, we would have to run two clusters. That sounds reasonable to me.

docs/rfcs/0006-clustering.md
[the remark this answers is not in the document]
ptodev: Sorry, I should have clarified this a bit more. What I mean is that "local hashing will provide the same results", but only if the hashing function is the same and if the "N owners" are the same. Which means that if we want to change either of them, we would need to restart the cluster to make sure it's consistent. I'm not sure if this is something worth mentioning explicitly or we can just assume it's obvious.

ptodev:
Thank you for the comments @tpaschalis, I'm personally happy if the PR is merged. And thanks very much for the great work!

ptodev (approved):

PR #3154, opened 2023-03-02
Phlare: Improve Component Shutdown

### LINKED ISSUE #3152: Phlare improve component shutdown
Currently as stated in https://github.com/grafana/agent/pull/3142 when we shutdown the `phlare.scrape `component, we will sequentially wait for all scrape pool to end. This can add up and be very slow.

But ultimately we should stop all of them in parallel and wait. This means we would only wait maximum 15s if the backend is available and a bit more if we have to retry.

cyriltovena:
#### PR Description

Ensure we close all scrape loop in parallel to reduce shutdown time.

#3152 

#### Notes to the Reviewer

Do we need a changelog for this ?

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/component/phlare/scrape/manager.go
+++ b/component/phlare/scrape/manager.go
@@ -189,8 +189,14 @@ func (m *Manager) Stop() {
 	m.mtxScrape.Lock()
 	defer m.mtxScrape.Unlock()
 
+	wg := sync.WaitGroup{}
 	for _, sp := range m.targetsGroups {
-		sp.stop()
+		wg.Add(1)
+		go func(sp *scrapePool) {
+			defer wg.Done()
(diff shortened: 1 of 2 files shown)


cyriltovena:
@tpaschalis That should do it.

@rfratto up to you know if you want to include this or not. I suggest not to and get it in the next version.

rfratto:
> Do we need a changelog for this ?

Nah, not if nobody's observed it being a problem. I think we're just being cautious here, which is good.

rfratto (approved):

tpaschalis (approved):

PR #3158, opened 2023-03-02
return an error if prometheus.exporter.unix or a config block are found inside a module

### LINKED ISSUE #3111: Flow Modules: Do not allow modules to load singletons
When iterating over the components the `module.*` components should fail if a config block or singleton is loaded.

Good place to do this is https://github.com/grafana/agent/blob/rfc_exploratory/component/module/string/string.go#L102

erikbaranowski:
#### PR Description

return an error if prometheus.exporter.unix or a config block are found inside a module.

Closes #3111 

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/component/module/string/string.go
+++ b/component/module/string/string.go
@@ -2,12 +2,14 @@ package string
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"sync"
 
 	"github.com/grafana/agent/component"
 	"github.com/grafana/agent/component/module/argument"
 	"github.com/grafana/agent/component/module/export"
+	"github.com/grafana/agent/component/prometheus/exporter/unix"
 )
 
 func init() {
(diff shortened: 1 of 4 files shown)


component/module/string/string.go
```diff
@@ -122,6 +124,8 @@ func (c *Component) Update(args component.Arguments) error {
 				// Go ahead and fill in the value
 				c.values[name] = val
 				c.exportComponents[cc.Name] = cc
+			case *unix.Component:
```
erikbaranowski, line 127: If there's a clever way to do this without copying `component/prometheus/exporter/exporter.go` into `component/prometheus/exporter/unix/exporter.go` let me know. Considering how soon it sounds like we will end up changing this to not be a singleton it may not be worth a ton of time.

PR #3176, opened 2023-03-03
[Flow] Add Postgres Component

### LINKED ISSUE #2310: Flow: Add postgres_exporter
Create `prometheus.integration.postgres_exporter` component

spartan0x117:
#### PR Description
Adds `prometheus.exporter.postgres` component for `postgres_exporter`.

Fixes #2310  

#### PR Checklist

- [X] CHANGELOG updated
- [X] Documentation added
- [X] Tests updated

### DIFF
--- a/component/prometheus/exporter/postgres/postgres.go
+++ b/component/prometheus/exporter/postgres/postgres.go
@@ -30,7 +30,8 @@ var DefaultArguments = Arguments{}
 // Arguments configures the prometheus.exporter.postgres component
 type Arguments struct {
 	// DataSourceNames to use to connect to Postgres. This is marked optional because it
-	// may also be supplied by the POSTGRES_EXPORTER_DATA_SOURCE_NAME env var
+	// may also be supplied by the POSTGRES_EXPORTER_DATA_SOURCE_NAME env var,
+	// though it is not recommended to do so.
 	DataSourceNames []rivertypes.Secret `river:"data_source_names,attr,optional"`
 
 	DisableSettingsMetrics bool     `river:"disable_settings_metrics,attr,optional"`
--- a/component/prometheus/exporter/postgres/postgres.go
+++ b/component/prometheus/exporter/postgres/postgres.go
@@ -38,7 +38,7 @@ type Arguments struct {
 	ExcludeDatabases       []string `river:"exclude_databases,attr,optional"`
 	IncludeDatabases       []string `river:"include_databases,attr,optional"`
 	DisableDefaultMetrics  bool     `river:"disable_default_metrics,attr,optional"`
-	QueryPath              string   `river:"query_path,attr,optional"`
+	CustomQueriesPath      string   `river:"custom_queries_path,attr,optional"`
 }
 
 // UnmarshalRiver implements River unmarshalling for Arguments.
@@ -57,7 +57,7 @@ func (a *Arguments) Convert() *postgres_exporter.Config {
 		ExcludeDatabases:       a.ExcludeDatabases,
 		IncludeDatabases:       a.IncludeDatabases,
 		DisableDefaultMetrics:  a.DisableDefaultMetrics,
-		QueryPath:              a.QueryPath,
+		QueryPath:              a.CustomQueriesPath,
 	}
 }
 
--- a/component/prometheus/exporter/postgres/postgres.go
+++ b/component/prometheus/exporter/postgres/postgres.go
@@ -1,0 +1,70 @@
+package postgres
+
+import (
+	"github.com/grafana/agent/component"
+	"github.com/grafana/agent/component/prometheus/exporter"
+	"github.com/grafana/agent/pkg/flow/rivertypes"
+	"github.com/grafana/agent/pkg/integrations"
+	"github.com/grafana/agent/pkg/integrations/postgres_exporter"
+	config_util "github.com/prometheus/common/config"
+)
+
+func init() {
+	component.Register(component.Registration{
+		Name:    "prometheus.exporter.postgres",
+		Args:    Arguments{},
+		Exports: exporter.Exports{},
+		Build:   exporter.New(createIntegration, "postgres"),
+	})
+}
+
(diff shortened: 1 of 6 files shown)


docs/sources/flow/reference/components/prometheus.exporter.postgres.md
```diff
@@ -0,0 +1,125 @@
...
+The following arguments are supported:
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`data_source_names`                  | `list(secret)`      | Specifies the Postgres server(s) to connect to.  |         | yes
+`disable_settings_metrics`           | `bool`              | Disables collection of metrics from pg_settings. | `false` | no
+`disable_default_metrics`            | `bool`              | When `true`, only exposes metrics supplied from `custom_queries_path`. | `false` | no
+`autodiscover_databases`             | `bool`              | Whether to automatically discover databases to scrape metrics from. | `false` | no
+`exclude_databases`                  | `list(string)`      | A list of databases to ignore when `autodiscover_databases` is `true` | | no
+`include_databases`                  | `list(string)`      | Includes only specific databases (excluding all others) when autodiscover_databases is `true` | | no
```
captncraig, line 39: This interaction between fields is confusing to me. Even after reading it a few times. Is it ever valid to have auto: true and an include list?

docs/sources/flow/reference/components/prometheus.exporter.postgres.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Debug metrics
+
+`prometheus.exporter.postgres` does not expose any component-specific
+debug metrics.
+
+## Examples
+
+This minimal example uses a `prometheus.exporter.postgres` component to collect metrics from a Postgres
+server running locally with all default settings:
```
captncraig, line 80: Are the defaults useful in this case? It seems like if autodiscover is off, it isn't going to collect very much. I've never used postgres exporter though, so I'm not sure.

component/prometheus/exporter/postgres/postgres.go
```diff
@@ -0,0 +1,71 @@
...
+// DefaultArguments holds the default arguments for the prometheus.exporter.postgres
+// component.
+var DefaultArguments = Arguments{}
+
+// Arguments configures the prometheus.exporter.postgres component
+type Arguments struct {
+	// DataSourceNames to use to connect to Postgres. This is marked optional because it
+	// may also be supplied by the POSTGRES_EXPORTER_DATA_SOURCE_NAME env var,
+	// though it is not recommended to do so.
+	DataSourceNames []rivertypes.Secret `river:"data_source_names,attr,optional"`
```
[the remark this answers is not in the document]
spartan0x117: There is some validation being done [inside the integration](https://github.com/grafana/agent/blob/main/pkg/integrations/postgres_exporter/postgres_exporter.go#L117), however it turns out that if neither was set [this line](https://github.com/grafana/agent/blob/main/pkg/integrations/postgres_exporter/postgres_exporter.go#L111) would result in a slice like: `[]string{""}` (since `os.Getenv` returns `""` if the env var is not set, and `strings.Split("", ",")` returns `[]string{""}`), so it was not detecting this since that is a slice of length 1.

I have updated the PR with a fix for this, but would be happy to make that a separate PR, if needed.

docs/sources/flow/reference/components/prometheus.exporter.postgres.md
```diff
@@ -0,0 +1,125 @@
...
+`exclude_databases`                  | `list(string)`      | A list of databases to ignore when `autodiscover_databases` is `true` | | no
+`include_databases`                  | `list(string)`      | Includes only specific databases (excluding all others) when autodiscover_databases is `true` | | no
+`custom_queries_path`                | `string`            | Path to YAML file containing custom queries to expose as metrics. | | no
+
+The format for connection strings in `data_source_names` can be found in the [official postgresql documentation](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING).
+
+See examples for the `custom_queries_path` file in the [postgres_exporter repository](https://github.com/prometheus-community/postgres_exporter/blob/master/queries.yaml).
+
+**NOTE**: There are a number of environment variables that are not recommended for use, as they will affect _all_ `prometheus.exporter.postgres` components. A full list can be found in the [postgres_exporter repository](https://github.com/prometheus-community/postgres_exporter#environment-variables).
```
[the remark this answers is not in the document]
spartan0x117: Personally, I would be hesitant to unset any environment variables as that could be pretty unexpected (picturing some shenanigans of having env vars in the config and then triggering a reload, at which point it would error as they had been unset the first go around).

As for outright forbidding it, I lean towards allowing the user to use them if they want while also noting we discourage it and why. It's not a strongly held opinion of mine, though, so would be open to discussion.

component/prometheus/exporter/postgres/postgres.go
```diff
@@ -0,0 +1,76 @@
...
+func init() {
+	component.Register(component.Registration{
+		Name:    "prometheus.exporter.postgres",
+		Args:    Arguments{},
+		Exports: exporter.Exports{},
+		Build:   exporter.New(createIntegration, "postgres"),
+	})
+}
+
+func createIntegration(opts component.Options, args component.Arguments) (integrations.Integration, error) {
```
marctc, line 21: nit: We called these `createExporter` in the other exporters (process exporters would need to be changed too)

docs/sources/flow/reference/components/prometheus.exporter.postgres.md
spartan0x117: I've changed the structure a bit to hopefully make it more clear that those config fields are linked together, as well as updated the documentation with more details and examples.

component/prometheus/exporter/postgres/postgres.go
spartan0x117: Done, thanks!

marctc (approved):

Issue #3115, opened 2023-02-28
Flow Modules: Fix UI so that the components screen can show all components and you can select them, but NOT show in the graph view


mattdurham:
This means passing the namespace id, and using that in most places.

Issue #3110, opened 2023-02-28
Flow Modules: Add a stack limit


mattdurham:
Add a limit on the depth that the modules can grow. Set a default of ?10?, this should be a command line parameter due to the nature of changing it.

erikbaranowski:
If we decide not to enforce a stack limit, test a reasonable depth and document it. Also document the recursive case as unenforced and not safe to do.

PR #3179, opened 2023-03-04
Flow: support modules through isolated controllers

### LINKED ISSUE #3113: Flow Modules: Add support in flow subsystem for loading modules

rfratto:
**NOTE TO REVIEWER:** I recommend reviewing this change commit-by-commit rather than looking across all changes to get a better understanding of what each change is doing.

This PR demonstrates how modules can be implemented by using fully isolated controllers, as originally suggested in https://github.com/grafana/agent/pull/3117#pullrequestreview-1317528752. The PR acts as an alternative to #3117 and #3170. 

The following are implemented in this PR: 

* `argument` and `export` blocks are supported. (#3113) 
* Components loaded by modules can be navigated to from the UI through navigating to a module loader component. (#3115) 

This PR also contains an example commit for a `module.string` component which loads a module from a string (#3114) but it isn't part of the final PR so it can be handled more thoughtfully later. 

To focus on demonstrating the approach, the following are **not** implemented in this PR: 

* Telemetry data (metrics, logs, traces) for components loaded from a module do not properly get propagated to the parent controller. 
* Singletons and logging/tracing blocks are not rejected (#3111). 
* There is no module stack limit. (#3110)
* The arguments and exports blocks are hacked in, but likely should be rethought so they exist as independent nodes in the DAG. 

Closes #3113.


rfratto:
If this is merged, the following would need to be addressed in follow up PRs:

* Telemetry data (metrics, logs, traces) for components loaded from a module do not properly get propagated to the parent controller.
* Singletons and logging/tracing blocks are not rejected (https://github.com/grafana/agent/issues/3111).
* There is no module stack limit. (https://github.com/grafana/agent/issues/3110)
* The arguments and exports blocks are hacked in, but likely should be rethought so they exist as independent nodes in the DAG.

erikbaranowski (approved):
LGTM

pkg/flow/internal/controller/component.go
```diff
@@ -151,6 +153,12 @@ func NewComponentNode(globals ComponentGlobals, b *ast.BlockStmt) *ComponentNode
 }
 
 func getManagedOptions(globals ComponentGlobals, cn *ComponentNode) component.Options {
+	// Make sure the prefix is always absolute.
+	prefix := globals.HTTPPathPrefix
+	if !strings.HasPrefix(prefix, "/") {
```
tpaschalis, line 159: This is only useful in case someone using a controller manually overwrites this, right?

rfratto: Either that, or if it's set to be the empty string.

pkg/flow/internal/controller/component.go
```diff
@@ -159,19 +160,30 @@ func getManagedOptions(globals ComponentGlobals, cn *ComponentNode) component.Op
...
+		ID:     globalID,
+		Logger: log.With(globals.Logger, "component", globalID),
 		Registerer: prometheus.WrapRegistererWith(prometheus.Labels{
-			"component_id": cn.nodeID,
+			"component_id": globalID,
 		}, wrapped),
-		Tracer:         wrapTracer(globals.TraceProvider, cn.nodeID),
+		Tracer: wrapTracer(globals.TraceProvider, globalID),
+
+		DataPath:       filepath.Join(globals.DataPath, cn.nodeID),
```
tpaschalis, line 182: Should we also use the globalID here so that different modules have a different DataPath?

rfratto: No, see the comments in flow.Options for DataPath: 

```
// Directory where components can write data. Constructed components will be
// given a subdirectory of DataPath using the local ID of the component.
//
// If running multiple Flow controllers, each controller must have a
// different value for DataPath to prevent components from colliding.
```

rfratto: Basically module loaders don't know what the shared DataPath is, because they _already_ get a unique path for their DataPath, and they would propagate that forward to the controllers they instantiate. 

It's a little confusing, but I couldn't find a better way of handling it right now other than documentation.

pkg/flow/flow.go
```diff
@@ -238,11 +240,35 @@ func (c *Flow) run(ctx context.Context) {
...
 // without any configuration errors.
-func (c *Flow) LoadFile(file *File) error {
+func (c *Flow) LoadFile(file *File, args map[string]any) error {
 	c.loadMut.Lock()
 	defer c.loadMut.Unlock()
 
-	diags := c.loader.Apply(nil, file.Components, file.ConfigBlocks)
+	// Fill out the values for the scope so that argument.NAME.value can be used
+	// to reference expressions.
+	evalatedArgs := make(map[string]any, len(file.Arguments))
```
tpaschalis, line 255: s/evalatedArgs/evaluatedArgs

pkg/flow/flow.go
```diff
@@ -106,6 +106,12 @@ type Options struct {
 	// The controller does not itself listen here, but some components
 	// need to know this to set the correct targets.
 	HTTPListenAddr string
+
+	// OnExportsChange is called when the exports of the controller change.
+	// Exports are controlled by "export" configuration blocks. If
+	// OnExportsChange is nil, export configuration blocks are not allowed in the
+	// loaded config file.
+	OnExportsChange func(exports map[string]any)
```
tpaschalis, line 114: What do we envision this function to be used for?

rfratto: See the commit which adds an example module.string component. It's used to propagate exports back out to the parent component/module.

tpaschalis: Ok, what I expected. Haven't gotten to that component yet.

component/module/string/string.go
```diff
@@ -0,0 +1,193 @@
...
+	// TODO(rfratto): sync exports with current set rather than updating it in
+	// full every time.
+	var emptyExports = map[string]any{}
+
+	for _, b := range f.ConfigBlocks {
+		if b.Name[0] == "export" {
+			emptyExports[b.Label] = nil
+		}
+	}
+	c.opts.OnStateChange(Exports{Values: emptyExports})
```
tpaschalis: I might be missing something, but why do we propagate an emptyExports here? Doesn't this mean we'll never inform the controller that it should propagate a change? Or is that something for upcoming PRs?

rfratto: (Note that the module loader component is taken out of the PR but the original commit is left in for referencing example usage of the controller)

tpaschalis (approved):
LGTM!

PR #3185, opened 2023-03-06
Cherry-pick "Improve Phlare target reloading (#3142)" into release v0.32

simonswine:
This cherry-picks #3142 into the v0.32 release branch

PR #3186, opened 2023-03-06
Cherry-pick "Improve Phlare target reloading (#3142)" into release v0.31

simonswine:
This cherry-picks #3142 into the v0.31 release branch


rfratto:
Closing in favor of #3185, since we typically don't put out patch releases for non-current minor versions.

PR #3200, opened 2023-03-07
Add an otelcol.auth.sigv4 component

### LINKED ISSUE #2358: Flow: otelcol.auth.sigv4 component

ptodev:
#### PR Description
Adding an otelcol.auth.sigv4 Flow component, which ports [sigv4auth](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/sigv4authextension).

Fixes #2358

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated


CHANGELOG.md
```diff
@@ -25,6 +25,8 @@ Main (unreleased)
     (@marctc).
   - `prometheus.exporter.mysql` collects metrics from a MySQL database.
     (@spartan0x117)
+  - `otelcol.auth.sigv4` performs `sigv4` authentication for making requests to AWS services
```
rfratto, line 28: Can we have this align with the documentation reference and initially spell this out as "AWS Signature Version 4 (SigV4)" for the first time it's referenced, and then use "SigV4" for other references?

component/otelcol/auth/sigv4/sigv4.go
```diff
@@ -0,0 +1,88 @@
...
+func (args Arguments) Convert() otelconfig.Extension {
+	res := sigv4authextension.Config{
+		ExtensionSettings: otelconfig.NewExtensionSettings(otelconfig.NewComponentID("sigv4")),
+		Region:            args.Region,
+		Service:           args.Service,
+		AssumeRole:        *args.AssumeRole.Convert(),
+	}
+	// sigv4authextension.Config has a private member called "credsProvider" which gets initialized when we call Validate().
+	// If we don't call validate, the unit tests for this component will fail.
+	res.Validate()
```
rfratto, line 44: This can fail, which makes me a little nervous. Should we implement `river.Unmarshaler` and have it do the validation and return an error if the validation fails? 

We would still have to do the validation again when calling Convert, but it would at least catch errors that we'd otherwise be potentially missing right now.

component/otelcol/auth/sigv4/sigv4.go
```diff
@@ -0,0 +1,88 @@
...
+// UnmarshalRiver implements river.Unmarshaler.
+func (h *AssumeRole) UnmarshalRiver(f func(interface{}) error) error {
+	type assumeRole AssumeRole
+
+	if err := f((*assumeRole)(h)); err != nil {
+		return err
+	}
+
+	return nil
+}
```
rfratto, line 75: I'm not sure if this is necessary when you're not applying defaults or changing the type we're unmarshaling to. Was there a reason you put this in?

component/otelcol/auth/sigv4/sigv4_test.go
```diff
@@ -0,0 +1,194 @@
...
+	"time"
+
+	"github.com/grafana/agent/component/otelcol/auth"
+	"github.com/grafana/agent/component/otelcol/auth/sigv4"
+	"github.com/grafana/agent/pkg/flow/componenttest"
+	"github.com/grafana/agent/pkg/river"
+	"github.com/grafana/agent/pkg/util"
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/collector/config/configauth"
+	"gotest.tools/assert"
```
rfratto, line 19: Please use `github.com/stretchr/testify/assert` for consistency with the assertion libraries we use

docs/sources/flow/reference/components/otelcol.auth.sigv4.md
```diff
@@ -0,0 +1,171 @@
...
+  service = "SERVICE"
+}
+```
+
+In some cases the exporter URL does contain enough information about the region and service as 
+described [here](#arguments). In that case there is no need to specify any arguments to this component:
+```river
+otelcol.auth.sigv4 "LABEL" {
+}
+```
```
rfratto, line 38: To follow the recommendations documented in #3145, this section should be composed of nothing but a River code block with required arguments. 

We can document when region/service may be required in the Arguments block without providing any information here.

docs/sources/flow/reference/components/otelcol.auth.sigv4.md
```diff
@@ -0,0 +1,171 @@
...
+
+## Arguments
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`region` | `string` | The AWS region for the service you are exporting to for AWS SigV4. | "" | no
+`service` | `string` | The AWS service for AWS SigV4. | "" | no
```
rfratto, line 45: Are these shorter descriptions still accurate?

```suggestion
`region` | `string` | The AWS region to sign with. | "" | no
`service` | `string` | The AWS service to sign with. | "" | no
```

docs/sources/flow/reference/components/otelcol.auth.sigv4.md
```diff
@@ -0,0 +1,171 @@
...
+
+Both `region` and `service` are required for the component to work, but if they are left empty their values 
+may be inferred from the exporter URL based on
```
rfratto, line 48: They're not required, then :) I think this needs some rephrasing to talk about how they must be provided if the URL in the exporters don't follow a set of rules.

docs/sources/flow/reference/components/otelcol.auth.sigv4.md
```diff
@@ -0,0 +1,171 @@
...
+## Arguments
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`region` | `string` | The AWS region for the service you are exporting to for AWS SigV4. | "" | no
+`service` | `string` | The AWS service for AWS SigV4. | "" | no
+
+Both `region` and `service` are required for the component to work, but if they are left empty their values 
+may be inferred from the exporter URL based on 
+[these](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.63.0/extension/sigv4authextension/signingroundtripper.go#L121) rules:
```
rfratto, line 49: IMO we should drop links to Go code since our readers might not be Go developers:

```suggestion
the following rules:
```

docs/sources/flow/reference/components/otelcol.auth.sigv4.md
```diff
@@ -0,0 +1,171 @@
...
+
+* If the exporter URL starts with `aps-workspaces` and `service` is empty, `service` will be set to `aps`.
+* If the exporter URL starts with `search-` and `service` is empty, `service` will be set to `es`.
+* If the exporter URL starts with either `aps-workspaces` or `search-` and `region` is empty, `region` .
+will be set to the value between the first and second `.` character in the exporter URL.
+
+For concrete examples, see the [examples](#examples) section.
```
rfratto, line 56: This shouldn't be necessary, readers will see on the sidebar that there is an examples section.

docs/sources/flow/reference/components/otelcol.auth.sigv4.md
```diff
@@ -0,0 +1,171 @@
...
+
+The AWS regions are listed [here](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.RegionsAndAvailabilityZones.html).
```
rfratto, line 58: Avoid calling links "here", instead try something like:

```suggestion
A list of valid AWS regions can be found on Amazon's documentation for [Regions, Availability Zones, and Local Zones](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.RegionsAndAvailabilityZones.html).
```

This will then align with the Google developer documentation recommendations for [Link text](https://developers.google.com/style/link-text).

docs/sources/flow/reference/components/otelcol.auth.sigv4.md
```diff
@@ -0,0 +1,171 @@
...
+
+## Blocks
+
+The following blocks are supported inside the definition of
+`otelcol.auth.sigv4`:
+
+Hierarchy | Block | Description | Required
+--------- | ----- | ----------- | --------
+assume_role | [assume_role][] | Specifies the configuration needed to assume a role. | no
```
rfratto, line 67:
```suggestion
assume_role | [assume_role][] | Configuration for assuming a role. | no
```

docs/sources/flow/reference/components/otelcol.auth.sigv4.md
```diff
@@ -0,0 +1,171 @@
...
+`otelcol.auth.sigv4` is only reported as unhealthy if given an invalid
+configuration.
+
+## Debug information
+
+`otelcol.auth.sigv4` does not expose any component-specific debug information.
+
+## Examples
+
+In the example below the exporter endpoint starts with `aps-workspaces`. Hence `service` is inferred to be `aps`
```
rfratto, line 105: Can you give each of these examples headers? That's not something mentioned (yet) in #3145 but I think it would help with the readability here.

docs/sources/flow/reference/components/otelcol.auth.sigv4.md
```diff
@@ -0,0 +1,171 @@
...
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`arn` | `string` | The Amazon Resource Name (ARN) of a role to assume. | "" | no
+`session_name` | `string` | The name of a role session. | "" | no
+`sts_region` | `string` | The AWS region where STS is used to assume the configured role. | "" | no
+
+Note that if a role is intended to be assumed and `sts_region` is not provided, then `sts_region`
+will default to the value for `region` if `region` is provided.
+
+`region` can be differentiated from `sts_region` to handle cross region authentication.
```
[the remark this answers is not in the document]
ptodev: Do you mean "can be set to different values"?

docs/sources/flow/reference/components/otelcol.auth.sigv4.md
```diff
@@ -0,0 +1,171 @@
...
+The `assume_role` block specifies the configuration needed to assume a role.
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`arn` | `string` | The Amazon Resource Name (ARN) of a role to assume. | "" | no
+`session_name` | `string` | The name of a role session. | "" | no
+`sts_region` | `string` | The AWS region where STS is used to assume the configured role. | "" | no
+
+Note that if a role is intended to be assumed and `sts_region` is not provided, then `sts_region`
+will default to the value for `region` if `region` is provided.
```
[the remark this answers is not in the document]
ptodev: Do you think this sounds better:

> If the `assume_role` block is specified in the config and `sts_region` is not set, then `sts_region` will default to the value for `region`

component/otelcol/auth/sigv4/sigv4.go
ptodev: Yes, I was not sure of the value this function adds by using `type assumeRole AssumeRole` so I decided to add this just in case. Now when I read your comment, I remembered why we have it - we avoid the infinite loop by calling assumeRole's UnmarshalRiver instead of AssumeRole's UnmarshalRiver. I'll remove the function, thank you!

component/otelcol/auth/sigv4/sigv4.go
ptodev: I'm not sure about having `river.Unmarshaler` do the validation for a couple of reasons:

- The validation is an implementation detail of Otel.
- The unmarshalling is simply all about creating a sigv4.Arguments structure out of the river string. It doesn't do anything Otel related, so I'm not sure how it'd have access to the Otel Validate() method?

Is the problem with Validate() erroring inside Convert() that it will panic? It is only used in the Update() function in auth.go, so we could just add error handling there in case Convert() fails?

component/otelcol/auth/sigv4/sigv4.go
```diff
@@ -0,0 +1,74 @@
...
+func (args Arguments) Extensions() map[otelconfig.ComponentID]otelcomponent.Extension {
+	return nil
+}
+
+// Exporters implements auth.Arguments.
+func (args Arguments) Exporters() map[otelconfig.DataType]map[otelconfig.ComponentID]otelcomponent.Exporter {
+	return nil
+}
+
+type AssumeRole struct {
```
tpaschalis, line 57: nit: let's put a docstring here so that the linter won't complain. (or maybe we don't even need to export this type?)

ptodev: Where did you see the linter issue? The only linter problem I can see is this one:

> component/otelcol/auth/sigv4/sigv4.go:43:14: Error return value of `res.Validate` is not checked (errcheck)

component/otelcol/auth/sigv4/sigv4.go
ptodev: Sorry, I don't understand this bit:

> have the UnmarshalRiver on top of what Validate() does.

How can UnmarshalRiver call Validate() if there is no otel object at that time? UnmarshalRiver is just to fill out sigv4.Arguments and there is no otel object during the unmarshalling.

rfratto:
```go
func (args *Arguments) UnmarshalRiver(f func(interface{}) error) error {
  type arguments Arguments 
  if err := f((*arguments)(args)); err != nil {
    return err 
  }

  return args.Convert().(sigv4authextension.Config).Validate()
}
```

ptodev: Thanks, I see. It didn't occur to me to do the `sigv4authextension.Config` conversion like that.

component/otelcol/auth/sigv4/sigv4.go
rfratto: > nit: let's put a docstring here so that the linter won't complain. (or maybe we don't even need to export this type?)


That check (which I love and miss) got removed when we upgraded golangci-lint. I want to bring it back in #2676 with a custom lint rule, but for now it's not required (but I still recommend documentation for all exported identifiers).

docs/sources/flow/reference/components/otelcol.auth.sigv4.md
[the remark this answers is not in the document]
rfratto: Yes, sorry

docs/sources/flow/reference/components/otelcol.auth.sigv4.md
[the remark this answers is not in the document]
rfratto: Much better, thanks!

docs/sources/flow/reference/components/otelcol.auth.sigv4.md
ptodev: Ok, I will delete this line.

docs/sources/flow/reference/components/otelcol.auth.sigv4.md
ptodev: Makes sense. What do you think about this instead:

> `region` and `service` only need to be specified if their values cannot be inferred from the exporter URL.
If they are left empty, their values may be inferred from the exporter URL based on the following rules:

rfratto: Maybe flip it around?

> If `region` and `service` are left empty, their values are inferred from the URL of the exporter using this component using the following rules:
>
> (RULES)
>
> If none of the above rules apply, then `region` and `service` must be specified.

docs/sources/flow/reference/components/otelcol.auth.sigv4.md
ptodev: I think they are accurate, and I do think they sound better, thank you.

component/otelcol/auth/sigv4/sigv4.go
```diff
@@ -0,0 +1,91 @@
...
+		ExtensionSettings: otelconfig.NewExtensionSettings(otelconfig.NewComponentID("sigv4")),
+		Region:            args.Region,
+		Service:           args.Service,
+		AssumeRole:        *args.AssumeRole.Convert(),
+	}
+	// sigv4authextension.Config has a private member called "credsProvider" which gets initialized when we call Validate().
+	// If we don't call validate, the unit tests for this component will fail.
+	err := res.Validate()
+	if err != nil {
+		panic(err)
```
ptodev, line 49: `Convert()` is unable to return errors, so the only ways to handle the error is to ignore it or panic. We can't ignore it, so I put in a panic statement. However, I think a better solution would be to have `Convert()` return an error, and then the `Update()` function in `component/otelcol/auth/auth.go` could catch the error and return an error as well? Otherwise we could panic during an update operation, which would not be great.

- [ ] Return an error instead of panicking?

docs/sources/flow/reference/components/otelcol.auth.sigv4.md
```diff
@@ -0,0 +1,168 @@
+---
+title: otelcol.auth.sigv4
+---
+
+# otelcol.auth.sigv4
+
+`otelcol.auth.sigv4` exposes a `handler` that can be used by other `otelcol`
+components to authenticate requests to AWS services using the AWS Signature Version 4 (SigV4) protocol. 
+See [here](https://docs.aws.amazon.com/general/latest/gr/signing-aws-api-requests.html) for more
```
rfratto, line 9: Can you make this a named link so users know what they're clicking on?

The Google documentation style guide recommends against links like "here" in favor of being more descriptive: https://developers.google.com/style/link-text

rfratto (approved):
LGTM! Just one final comment on the docs.

component/otelcol/auth/sigv4/sigv4.go
ptodev: Ok, so you are suggesting to remove the Validate() from Convert(), since we are already calling it in UnmarshalRiver? I agree that it is a bit redundant now that UnmarshalRiver calls it explicitly.

However, do you think it makes the code cleaner to have Convert() call Validate(), and UnmarshalRiver to just call Convert()? Then we wouldn't have to do the type assertion. I think if Convert() would need to have some sort of check that it produced a valid otel object inside of it, and return an error if it didn't manage to create it?

rfratto: No, I'm saying you can leave this as-is for now as a safety measure, and then we can clean it up in a follow-up PR.

ptodev: > Convert() would need to have some sort of check that it produced a valid otel object
Otherwise every function which calls Convert would also have to do a type assertion and a Validate().

rfratto: No, I think this should be left as-is for now. UnmarshalRiver should definitely not call a function which is possible to panic, since panics should be reserved for scenarios which the developers do not anticipate happening. 

The type assertion can be left in for now, but in the follow up PR it'll no longer be necessary to do the validation in UnmarshalRiver anymore since Convert will be able to report errors instead. That will remove the need for the type assertion.

ptodev: But if I leave it as it is, Convert will panic, then UnmarshalRiver will panic too and it will not even get to the type assertion?

rfratto: Thanks, I didn't catch that. You can remove the call to Validate in Convert for now and we'll add it back in later

rfratto: How about we open up a separate PR to change the interface first to allow Convert to return errors, and then after it's merged, rebase this PR on top of main so we can take advantage of it immediately?

ptodev: Or by "leave it as it is" you mean to ignore the error in Convert?

ptodev: I pushed a change to Convert() to ignore the error in Validate(). If you are happy with this, please feel free to merge it.
Thanks very much for the review!

PR #3201, opened 2023-03-07
Remove snowflake v2 integration code

marctc:
#### PR Description

There's been some confusion whether integrations should have a piece of code to
enable usage of `integrations-next` (v2). In most of the case, this can be simply
achieved using the following line in `init` function of the integration. This makes 
integration automatically registered and available for v2 config mode.

`integrations_v2.RegisterLegacy(&Config{}, integrations_v2.TypeMultiplex, metricsutils.NewNamedShim("snowflake"))`

This PR removes unnecessary code that enabled v2 integration for snowflake. Also
password field should be of the type `Secret`.

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated


Dasomeone (approved):
LGTM, especially with the context of why Snowflake doesn't need any custom v2 code given that it's a database integration

Thanks @marctc for explaining :)

PR #3202, opened 2023-03-07
Use missing Secret field in integrations

marctc:
#### PR Description

Some integrations were using plain strings instead of `Secret` type for
secret fields. This PR fixes it.

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated


Dasomeone (approved):
Nice catch, thanks! 

Will definitely keep a note of this for future exporter reviews :)

PR #3203, opened 2023-03-07
[issue-3173] Fix some values references and add new properties

### LINKED ISSUE #3173: [helm] grafana-agent chart has some configurability missing for my use case
- `hostNetwork` for `Deamonset`
- `envFrom` to mount all keys on a secret as environment variables. Faro API keys for each app
- multiple `Ingress`, one per `app_agent_receiver_configs` server `port`
- `structuredConfig` similar to Loki, Mimir... charts for `mode: static`

I run `grafana-agent` as a `Deamonset` with `hostNetwork` to allow Kubernetes PODs to find the traces host by using the node's IP where they are running. This is a consequence of the migration from Opentelemetry collector which was able to enrich traces with the EKS and EC2 node metadata.

By provisioning multiple RUM Faro applications I need to manage various Ingresses out of the Helm chart but this could be easily done within the chart

carlosjgp:
#### PR Description

Found some properties referencing the wrong key on the `values.yaml` file and I've added a few new properties to the chart to enable the deployment on my clusters, hopefully also useful to other users of the chart

Fixes #3173 

#### Notes to the Reviewer

N/A

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [x] Tests updated


rfratto:
Thanks for the PR! I'll try to review this within the next few days.

rfratto:
👋 This looks good, but I'm concerned about why a lot of the test files got deleted.

Also, there's a few other changes we might want to batch into 0.9.0, so can you remove the version bump for now? We'll release the changes ASAP after this is merged.

carlosjgp:
> wave This looks good, but I'm concerned about why a lot of the test files got deleted.
> 
> Also, there's a few other changes we might want to batch into 0.9.0, so can you remove the version bump for now? We'll release the changes ASAP after this is merged.

I've reverted the version

Regardless of the deleted test files I think the script `make generate-helm-tests` did delete them... or maybe I generated the test files there and I noticed my mistake and then deleted someone else's generated tests...?

Notice that the deleted test file paths are `/operations/helm/operations/helm/tests/create-statefulset/...` noticed the double `/operations/helm`

carlosjgp:
I've updated the script to be able to run 
```
agent > $ make generate-helm-tests
```
and 
```
agent/operations/helm > $ make rebuild-tests
```

into the same folder

rfratto:
Hi, I'm having a hard time reviewing the changes in this PR. I'm having a hard time understanding where our existing chart tests went.

This feels like it should be broken up into a few PRs. I would suggest that we focus this PR on just the changes you want to see added to the chart, and for us to discuss other changes (like changes to the rebuild-tests.sh) script in a follow up PR. 

If we can reduce the size of this PR down, I'll be able to review it much more quickly and we can get it merged.

carlosjgp:
split into these two PRs
- https://github.com/grafana/agent/pull/3340
- https://github.com/grafana/agent/pull/3342

PR #3204, opened 2023-03-07
river: allow referencing zero values in Go maps

rfratto:
This fixes an issue where any reference to a zero value in a Go map would be treated as if the key didn't exist. This was due to a misunderstanding of the Go documentation that said "the zero Value" will be returned if the key does not exist in the map. This documentation referred to an invalid `reflect.Value` instead of the zero value of the map's value type.

This allows to simplify the logic when performing a map lookup to determine if the key was found or not, and permits expressions which reference a map key set to the zero value of the value type.

Opening in draft because it still needs some unit tests.

### DIFF
--- a/pkg/river/internal/value/value_test.go
+++ b/pkg/river/internal/value/value_test.go
@@ -13,55 +13,95 @@ import (
 // implement an interface for a pointer receiver are retained correctly
 // throughout values.
 func TestInterfacePointerReceiver(t *testing.T) {
-	t.Run("Encode", func(t *testing.T) {
-		pm := &pointerMarshaler{}
+	tt := []struct {
+		name         string
+		encodeTarget any
+		key          string
 
-		val := value.Encode(pm)
-		require.Equal(t, value.TypeString, val.Type())
-		require.Equal(t, "Hello, world!", val.Text())
-	})
+		expectBodyType value.Type
+		expectBodyText string
 
-	t.Run("Struct Encode Key", func(t *testing.T) {
-		type Body struct {
-			Data pointerMarshaler `river:"data,attr"`
-		}
+		expectKeyExists bool
+		expectKeyValue  value.Value
+		expectKeyType   value.Type
+	}{
+		{
+			name:           "Encode",
+			encodeTarget:   &pointerMarshaler{},
+			expectBodyType: value.TypeString,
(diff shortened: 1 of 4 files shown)


pkg/river/internal/value/value_test.go
```diff
@@ -64,6 +100,10 @@ func (*pointerMarshaler) MarshalText() ([]byte, error) {
 	return []byte("Hello, world!"), nil
 }
 
+type Body struct {
+	Data pointerMarshaler `river:"data,attr"`
+}
```
rfratto, line 105: @erikbaranowski Since this is being declared for the scope of the one test, and it doesn't have any methods, can we move it to the top of the inside of the function?

```go
func TestInterfacePointerReceiver() {
  type Body { ... } 

  ...
}

File: pkg/river/internal/value/value_test.go [hunk 2]
```diff
@@ -13,29 +13,85 @@ import (
...
+		expectKeyExists bool
+		expectKeyValue  value.Value
+		expectKeyType   value.Type
+	}{
+		{
+			name:           "Encode",
+			encodeTarget:   &pointerMarshaler{},
+			expectBodyType: value.TypeString,
+			expectBodyText: "Hello, world!",
+		},
```
rfratto, line 33: Can we split this out into its own test function? It feels a bit weird that the rest of these tests are more about testing key lookups but this one isn't

rfratto, line 33: @erikbaranowski Can we split this out into its own test function? It feels a bit weird that the rest of these tests are more about testing key lookups but this one isn't

pkg/river/internal/value/value_test.go
```diff
@@ -9,33 +9,107 @@ import (
...
+		name         string
+		encodeTarget any
+		key          string
+
+		expectBodyType  value.Type
+		expectKeyExists bool
+		expectKeyValue  value.Value
+		expectKeyType   value.Type
+	}{
+		{
```
rfratto, line 29: @erikbaranowski Can you add a test case here for getting a zero value from a River object?

For example, for `map[string]string{"data": ""}` the lookup should succeed and the resulting value should be the empty string. 

After that, I think this is good to merge.

PR #3205, opened 2023-03-07
river: fix capsule type loss when converting capsules to any

### LINKED ISSUE #3171: Flow Modules: Exports returning nil that needs to be stuffed into capsule types fails
Using the export function returns an interface that wraps the capsule type and into value is a capsule fails. Likely these check needs to be made in value.decode.

rfratto:
This makes two changes to fix the loss of the capsule type when a capsule value is converted to interface{}:

1. Structs with no River tags are no longer considered objects, and are now considered capsules.

2. Capsules can be converted if the Go types can be converted (such as assigning to an interface).

Fixes #3171.
Related to #2547.

Opening in draft because it still needs some unit tests.


rfratto:
cc @tpaschalis do you think this is a safe move?

pkg/river/internal/value/decode.go
```diff
@@ -339,6 +339,17 @@ func tryCapsuleConvert(from Value, into reflect.Value, intoType Type) (ok bool,
 		}
 	}
 
+	// Last attempt: see if the Go types can be converted.
+	if from.Reflect().CanAddr() && from.Reflect().Addr().CanConvert(into.Type()) {
```
rfratto, line 343: The order here is probably important, since it's more ideal to convert the pointer to avoid a copy.

rfratto, line 343: @erikbaranowski This is primarily where we need the tests, I already covered the expanded definition of capsule types.

(I think the package-level documentation in pkg/river might need to be updated for this too)

tpaschalis (approved):
I can't think of any drawbacks right now.

PR #3216, opened 2023-03-08
Update Redis exporter  to v1.48.0

### LINKED ISSUE #3178: Update github.com/oliver006/redis_exporter to v1.48.0
An update for `github.com/oliver006/redis_exporter` (version `v1.48.0`) is now available. Version `v1.27.1` is currently in use.

spartan0x117:
#### PR Description
Bumps the redis_exporter dependency to the latest release. This includes a few new options, as well as several bugfixes and metrics additions.

New options:
- `incl_config_metrics`: Adds new gauges for Redis config elements.
- `redact_config_metrics`: Related to `incl_config_metrics`, controls redacting secrets in the Redis config.
- `IsCluster`: Whether the connection is to a cluster, for the purposes of monitoring individual keys with the `check_keys` and `check_single_keys` options.

Updated options:
- `script_path`: For static mode, now supports passing in a comma-separated list of paths to Lua scripts instead of only a single path. For flow mode, this argument has been changed from `string` to `list(string)` and has been renamed to `script_paths`.

Fixes #3178 

#### PR Checklist

- [X] CHANGELOG updated
- [X] Documentation added
- [X] Tests updated

### DIFF
--- a/component/prometheus/exporter/redis/redis.go
+++ b/component/prometheus/exporter/redis/redis.go
@@ -36,6 +36,7 @@ var DefaultConfig = Config{
 	SetClientName:           true,
 	CheckKeyGroupsBatchSize: 10000,
 	MaxDistinctKeyGroups:    100,
+	RedactConfigMetrics:     true,
 }
 
 type Config struct {
--- a/component/prometheus/exporter/redis/redis.go
+++ b/component/prometheus/exporter/redis/redis.go
@@ -60,18 +60,21 @@ type Config struct {
 	CheckStreams            []string          `river:"check_streams,attr,optional"`
 	CheckSingleStreams      []string          `river:"check_single_streams,attr,optional"`
 	CountKeys               []string          `river:"count_keys,attr,optional"`
-	ScriptPath              string            `river:"script_path,attr,optional"`
+	ScriptPaths             []string          `river:"script_paths,attr,optional"`
 	ConnectionTimeout       time.Duration     `river:"connection_timeout,attr,optional"`
 	TLSClientKeyFile        string            `river:"tls_client_key_file,attr,optional"`
 	TLSClientCertFile       string            `river:"tls_client_cert_file,attr,optional"`
 	TLSCaCertFile           string            `river:"tls_ca_cert_file,attr,optional"`
 	SetClientName           bool              `river:"set_client_name,attr,optional"`
 	IsTile38                bool              `river:"is_tile38,attr,optional"`
+	IsCluster               bool              `river:"is_cluster,attr,optional"`
 	ExportClientList        bool              `river:"export_client_list,attr,optional"`
 	ExportClientPort        bool              `river:"export_client_port,attr,optional"`
 	RedisMetricsOnly        bool              `river:"redis_metrics_only,attr,optional"`
 	PingOnConnect           bool              `river:"ping_on_connect,attr,optional"`
 	InclSystemMetrics       bool              `river:"incl_system_metrics,attr,optional"`
+	InclConfigMetrics       bool              `river:"incl_config_metrics,attr,optional"`
+	RedactConfigMetrics     bool              `river:"redact_config_metrics,attr,optional"`
 	SkipTLSVerification     bool              `river:"skip_tls_verification,attr,optional"`
 }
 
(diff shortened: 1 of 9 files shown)


component/prometheus/exporter/redis/redis.go
```diff
@@ -60,18 +61,21 @@ type Config struct {
 	CheckStreams            []string          `river:"check_streams,attr,optional"`
 	CheckSingleStreams      []string          `river:"check_single_streams,attr,optional"`
 	CountKeys               []string          `river:"count_keys,attr,optional"`
-	ScriptPath              string            `river:"script_path,attr,optional"`
+	ScriptPaths             []string          `river:"script_paths,attr,optional"`
```
rfratto, line 64: This is technically a breaking change, so we should be a little considerate here.

Do you want it to be a breaking change? If so, that's probably fine pre-v1. If not, you might have to consider a different route (like changing the documentation for `script_path` to mention delimiting multiple scripts by comma, or supporting both `script_path` and `script_paths` but have them be mutually exclusive)

spartan0x117: It should probably be left as a non-breaking change. I like the idea of having both options, are there any existing flow components you know of that have a pattern like that?

rfratto: > I like the idea of having both options, are there any existing flow components you know of that have a pattern like that?

There's a few fields in the squashed HTTPClientConfig blocks which are mutually exclusive, you could check there

pkg/integrations/redis_exporter/redis_exporter.go
```diff
@@ -135,11 +142,15 @@ func New(log log.Logger, c *Config) (integrations.Integration, error) {
 	}
 
 	if c.ScriptPath != "" {
-		ls, err := os.ReadFile(c.ScriptPath)
-		if err != nil {
-			return nil, fmt.Errorf("Error loading script file %s: %w", c.ScriptPath, err)
+		scripts := map[string][]byte{}
+		for _, path := range strings.Split(c.ScriptPath, ",") {
+			ls, err := os.ReadFile(path)
```
captncraig, line 147: Are we really going to load every file in a directory? Seems like a potential security issue. Especially if file names and contents get into the logs on parse errors.

spartan0x117: It shouldn't be loading every file in the directory, just loading every file in a comma-separated list specified in the config. So as an example if provided with the config:
```yaml
  [...]
  script_path: "/tmp/redis-metrics.lua,/tmp/extra-redis-metrics.lua"
```
This would be splitting `"/tmp/redis-metrics.lua,/tmp/extra-redis-metrics.lua"` into `[]string{"/tmp/redis-metrics.lua", "/tmp/extra-redis-metrics.lua"}` and reading each of those.

component/prometheus/exporter/redis/redis.go
```diff
@@ -80,10 +86,27 @@ func (c *Config) UnmarshalRiver(f func(interface{}) error) error {
...
 
 	type cfg Config
-	return f((*cfg)(c))
+	if err := f((*cfg)(c)); err != nil {
+		return err
+	}
+	return c.Validate()
+}
+
+func (c *Config) Validate() error {
```
marctc, line 95: Should we add test for this?

marctc (approved):

component/prometheus/exporter/redis/redis.go
```diff
@@ -80,10 +86,27 @@ func (c *Config) UnmarshalRiver(f func(interface{}) error) error {
...
+	if c.ScriptPath != "" && len(c.ScriptPaths) > 0 {
+		return fmt.Errorf("only one of script_path and script_paths should be specified")
```
rfratto, line 97: IMO, should implies it's optional like a warning, but it's being asserted here as a requirement:

```suggestion
		return fmt.Errorf("only one of script_path and script_paths can be specified")
```

On second read of this though, I'm not sure if "can" sounds any different than "should" so feel free to ignore this 🤷

docs/sources/flow/reference/components/prometheus.exporter.redis.md
```diff
@@ -42,27 +42,36 @@ Name | Type | Description | Default | Required
...
 
-The `check_key_groups_batch_size` option name reflects key groups for backwards compatibility, but applies to both key and key groups.
+The `check_key_groups_batch_size` argument name reflects key groups for backwards compatibility, but applies to both key and key groups.
+
+The `script_path` argument may also be specified as a comma-separated string of paths, though it is encouraged to use `script_paths` when using
+multiple Lua scripts.
 
 Any leftover key groups beyond `max_distinct_key_groups` are aggregated in the 'overflow' bucket.
 
+The `is_cluster` argument should be used when connecting to a Redis cluster and using the `check_keys` and/or `check_single_keys` arguments.
```
rfratto, line 73: Should? Or must?

```suggestion
The `is_cluster` argument must be set to `true` when connecting to a Redis cluster and using either of the `check_keys` and `check_single_keys` arguments.
```

rfratto (approved):

PR #3221, opened 2023-03-09
add autoscaling for deployment to helm chart

therealmanny:
#### PR Description

This change adds autoscaling value for deployment to the helm chart. Provides the ability to scale app agent receiver deployment.

None

#### Notes to the Reviewer

In this pull request, I am assuming it is a good idea to run more than one app agent receiver integration configured grafana agent replicas forwarding telemetry to the same location. If this is wrong, you can reject this pull request. Thanks!

#### PR Checklist

- [X] CHANGELOG updated
- [X] Documentation added
- [X] Tests updated


operations/helm/charts/grafana-agent/templates/hpa.yaml
```diff
@@ -0,0 +1,34 @@
...
+spec:
+  scaleTargetRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: {{ include "grafana-agent.fullname" . }}
+  {{- with .Values.controller.autoscaling }}
+  minReplicas: {{ .minReplicas }}
+  maxReplicas: {{ .maxReplicas }}
+  metrics:
+    {{- with .targetCPUUtilizationPercentage }}
```
captncraig, line 17: Does it always make sense to scale based on both cpu and memory? If we scale by memory, but are not so concerned about cpu, can we do that?

operations/helm/charts/grafana-agent/templates/hpa.yaml
```diff
@@ -0,0 +1,34 @@
+{{- if and (eq .Values.controller.type "deployment") .Values.controller.autoscaling.enabled }}
```
rfratto, line 1: Would it be possible to use an HPA with a StatefulSet too?

rfratto (commented):
Thanks for the PR! I'm a little nervous about this one, actually. I think we definitely want to support HPAs at some point, but I don't think we have a good grip on how to use them effectively when configuring the agent as a whole. 

If we're going to merge this, I think it there should probably be a section in the chart's readme which explains how to use the `controller.autoscaling` section and specifically what use cases it works with. Since you mentioned the app agent receiver, we can probably start there and expand it out as we gain more knowledge. 

@captncraig do you think this is something we can merge and support today or should we hold off for a while?

operations/helm/charts/grafana-agent/templates/hpa.yaml
therealmanny: Yes, we can scale by only memory or CPU. The documentation is updated to reflect that `targetCPUUtilizationPercentage` or `targetMemoryUtilizationPercentage` can be diable by setting them to 0. Thanks!

operations/helm/charts/grafana-agent/templates/hpa.yaml
therealmanny: Yes, it is possible to use an HPA with a StatefulSet. But to reduce the nervousness about this change, it is only targeted toward deployment. If we are comfortable with adding HPA for StatefulSet, I can make the change to add it. Thanks!

captncraig:
I agree with your concerns, but am still ok merging it, as long as the language is clear. This will not reshard or anything like that, and it may create a lot of redundant work if your config is not suited. This will be extremely useful once we have native clustering though.

rfratto:
@therealmanny would you like to take a shot at adding a header to document this to the chart's readme, or would you prefer for us to do it?

therealmanny:
> @therealmanny would you like to take a shot at adding a header to document this to the chart's readme, or would you prefer for us to do it?

@rfratto I can take a shot at documenting this to the chart's readme.

operations/helm/charts/grafana-agent/README.md
```diff
@@ -159,6 +159,11 @@ components like [discovery.kubernetes][] to work properly.
 
 [discovery.kubernetes]: https://grafana.com/docs/agent/latest/flow/reference/components/discovery.kubernetes/
 
+### controller.autoscaling
+
+`controller.autoscaling.enabled` enables the creation of a HorizontalPodAutoscaler for `controller.type` deployment. The use case for `controller.autoscaling` is an [app agent receiver](https://grafana.com/docs/agent/latest/configuration/integrations/integrations-next/app-agent-receiver-config/) configured Grafana Agent.
```
rfratto, line 164:
```suggestion
`controller.autoscaling.enabled` enables the creation of a HorizontalPodAutoscaler. It is only used when `controller.type` is set to `deployment`.

`controller.autoscaling` is intended to be used with an [app agent receiver](https://grafana.com/docs/agent/latest/configuration/integrations/integrations-next/app-agent-receiver-config/) configured Grafana Agent.
```

rfratto: (You'll also have to make this change in the readme template since the README is also autogenerated)

rfratto (approved):
Looks good, a little feedback on the README and we can get this merged!

rfratto:
Thanks for the contribution!

PR #3225, opened 2023-03-09
prometheus.exporter.unix should still use job:integrations/node_exporter

### LINKED ISSUE #3180: Flow: prometheus.exporter.unix not compatible with Grafana Cloud integration
Out of the box, you can't use `prometheus.exporter.unix` with the Grafana Cloud integration, because the Grafana Cloud integration expects the `job` label to be set to `integrations/node_exporter`.

Additionally, the `instance` label is different. The `instance` label used to be set to the agent's hostname and listen port, but it's been changed to the component name, which is less useful when there's a ton of running agents. 

To get it to work, I had to stick a `discovery.relabel` component in between the exporter and scraping:

```river 
prometheus.exporter.unix {
	disable_collectors = ["mdadm"]
}

// We need to change the output of prometheus.exporter.unix to be 
// compatible with the node_exporter integration:
//
// * Change the job label to integrations/node_exporter.
// * Change the instance label to be HOSTNAME:LISTEN_PORT.
discovery.relabel "node_exporter" {
	targets = prometheus.exporter.unix.targets

	rule {
		action       = "replace"
		target_label = "job"
		replacement  = "integrations/node_exporter"
	}

	rule {
		action       = "replace"
		target_label = "__hostname__"
		replacement  = "{{ ansible_facts['nodename'] }}"
	}

	rule {
		action        = "replace"
		target_label  = "__port__"
		regex         = ".*:(.*)"
		replacement   = "$1"
	}

	rule {
		action        = "replace"
		source_labels = ["__hostname__", "__port__"]
		separator     = ":"
		target_label  = "instance"
	}
}

prometheus.scrape "default" {
	targets    = discovery.relabel.node_exporter.output
	forward_to = [prometheus.remote_write.grafanacloud.receiver]

	scrape_interval = "60s"
}
```

Ideally the `discovery.relabel` component isn't necessary for remapping the data to what it used to be.

captncraig:
Related to #3180

### DIFF
--- a/component/prometheus/exporter/unix/unix.go
+++ b/component/prometheus/exporter/unix/unix.go
@@ -13,7 +13,9 @@ func init() {
 		Args:      Config{},
 		Exports:   exporter.Exports{},
 		Singleton: true,
-		Build:     exporter.New(createExporter, "unix"),
+		// set name to node_exporter instead of unix. This is for backward compatibility
+		// with cloud integrations, dashboards, and other systems that look for `job:integrations/node_exporter`
+		Build: exporter.New(createExporter, "node_exporter"),
 	})
 }
 


spartan0x117 (approved):
LGTM, are there any other components that would need similar treatment that we know of?

jcreixell (changes requested):
I think it would be better to change this in the cloud mixin to avoid inconsistent naming here. Let's discuss it before merging this

captncraig:
Superceded by #3873

PR #3227, opened 2023-03-09
module.string: new component

### LINKED ISSUE #3114: Flow Modules: Add `module.string` component

rfratto:
This commit adds a `module.string` component, which loads a flow module from a string and runs it in a nested controller.

The `module.string` component accept arguments to pass to the controller, and exposes the controller's exports back out to peer components.

TODO:

- [x] Resolve any TODOs in the code 
- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated(?) 

Closes #3114

### DIFF
--- a/component/module/string/string.go
+++ b/component/module/string/string.go
@@ -43,8 +43,8 @@ type Arguments struct {
 
 // Exports holds values which are exported from the run module.
 type Exports struct {
-	// Values exported from the running module.
-	Values map[string]any `river:"values,attr"`
+	// Exports exported from the running module.
+	Exports map[string]any `river:"exports,attr"`
 }
 
 // Component implements the module.string component.
@@ -89,10 +89,11 @@ func New(o component.Options, args Arguments) (*Component, error) {
 			HTTPListenAddr: o.HTTPListenAddr,
 
 			OnExportsChange: func(exports map[string]any) {
-				o.OnStateChange(Exports{Values: exports})
+				o.OnStateChange(Exports{Exports: exports})
 			},
 		}),
 	}
+
 	if err := c.Update(args); err != nil {
 		return nil, err
 	}
@@ -128,7 +129,6 @@ func (c *Component) updateHealth(err error) {
 
 // Update implements component.Component.
 func (c *Component) Update(args component.Arguments) error {
-	// TODO: Figure out a way for sibling components to have access to exports
 	newArgs := args.(Arguments)
 
 	f, err := flow.ReadFile(c.opts.ID, []byte(newArgs.Content))
--- a/component/module/string/string.go
+++ b/component/module/string/string.go
@@ -8,8 +8,6 @@ import (
 	"sync"
 	"time"
 
-	"go.uber.org/atomic"
-
 	"github.com/go-kit/log"
 	"github.com/gorilla/mux"
 	"github.com/grafana/agent/component"
@@ -55,10 +53,6 @@ type Component struct {
 	log  log.Logger
 	ctrl *flow.Flow
 
-	moduleArgs        atomic.Pointer[map[string]any]
-	file              atomic.Pointer[flow.File]
-	updateContollerCh chan struct{}
-
 	healthMut sync.RWMutex
 	health    component.Health
 }
@@ -98,8 +92,6 @@ func New(o component.Options, args Arguments) (*Component, error) {
 				o.OnStateChange(Exports{Values: exports})
 			},
 		}),
-
-		updateContollerCh: make(chan struct{}, 1),
 	}
 	if err := c.Update(args); err != nil {
 		return nil, err
@@ -109,17 +101,9 @@ func New(o component.Options, args Arguments) (*Component, error) {
 
 // Run implements component.Component.
 func (c *Component) Run(ctx context.Context) error {
-Loop:
-	for {
-		select {
-		case <-ctx.Done():
-			break Loop
-		case <-c.updateContollerCh:
-			err := c.ctrl.LoadFile(c.file.Load(), *c.moduleArgs.Load())
-			c.updateHealth(err)
-		}
-	}
+	c.ctrl.Run()
 
+	<-ctx.Done()
 	return c.ctrl.Close()
 }
 
@@ -144,6 +128,7 @@ func (c *Component) updateHealth(err error) {
 
 // Update implements component.Component.
 func (c *Component) Update(args component.Arguments) error {
+	// TODO: Figure out a way for sibling components to have access to exports
 	newArgs := args.(Arguments)
 
 	f, err := flow.ReadFile(c.opts.ID, []byte(newArgs.Content))
@@ -151,36 +136,8 @@ func (c *Component) Update(args component.Arguments) error {
 		return err
 	}
 
-	// TODO(rfratto): sync exports with current set rather than updating it in
-	// full every time.
-	var emptyExports = map[string]any{}
-
-	for _, b := range f.ConfigBlocks {
-		if b.Name[0] == "export" {
-			emptyExports[b.Label] = nil
-		}
-	}
-
-	// Create an initial exported value which contains all the keys that
-	// correspond with the exports from our module so that sibling components can
-	// properly reference them by name, albeit with the zero value.
-	//
-	// Currently, this runs every time Update is called, so there will always be
-	// two calls to OnStateChange happening every time: once which sets everything
-	// to the zero value, and once where the evaluated concrete values are
-	// exposed (in the callback to OnExportsChange).
-	//
-	// TODO(rfratto): find a way to not override the last evaluated value with
-	// null.
-	c.opts.OnStateChange(Exports{Values: emptyExports})
-
-	c.file.Store(f)
-	c.moduleArgs.Store(&newArgs.Arguments)
-
-	select {
-	case c.updateContollerCh <- struct{}{}:
-	default: // no-op
-	}
+	err = c.ctrl.LoadFile(f, newArgs.Arguments)
+	c.updateHealth(err)
 
 	return nil
 }
--- a/component/module/string/string.go
+++ b/component/module/string/string.go
@@ -36,8 +36,8 @@ func init() {
 // Arguments holds values which are used to configure the module.string
 // component.
 type Arguments struct {
-	// Source code to load for the module.
-	Source rivertypes.Secret `river:"source,attr"`
+	// Content to load for the module.
+	Content rivertypes.Secret `river:"content,attr"`
 
 	// Arguments to pass into the module.
 	Arguments map[string]any `river:"arguments,attr,optional"`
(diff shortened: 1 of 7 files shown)


pkg/flow/flow.go
```diff
@@ -192,18 +184,23 @@ func newFlow(o Options) (*Flow, context.Context) {
...
 		exited:       make(chan struct{}, 1),
 		loadFinished: make(chan struct{}, 1),
-	}, ctx
+	}
+}
+
+// Starts flow controller as a go routine
+func (c *Flow) Run() {
+	go c.run()
 }
```
rfratto, line 195: IMO this is a bit of an anti-pattern, and I would rather:

1. Have `Flow.Run` be a blocking call (so the caller has to decide whether it should be launched in a goroutine rather than us making that decision for them)
2. Pass the context to use to `Flow.Run` 
3. Remove `Flow.exited` and `Flow.Stop`, which will no longer be required.

erikbaranowski: That makes sense, I didn't know how tied we were to the existing Flow public API on this 1 so I didn't refactor it.

rfratto: Done, I just made these changes

component/module/string/string.go
```diff
@@ -0,0 +1,160 @@
...
+			return New(opts, args.(Arguments))
+		},
+	})
+}
+
+// Arguments holds values which are used to configure the module.string
+// component.
+type Arguments struct {
+	// Content to load for the module.
+	Content rivertypes.Secret `river:"content,attr"`
```
rfratto, line 38: @erikbaranowski WDYT about making this a rivertypes.OptionalSecret so the user can specify whether its sensitive or not?

component/module/string/string.go
```diff
@@ -0,0 +1,160 @@
...
+
+	// Arguments to pass into the module.
+	Arguments map[string]any `river:"arguments,attr,optional"`
+}
+
+// Exports holds values which are exported from the run module.
+type Exports struct {
+	// Exports exported from the running module.
+	Exports map[string]any `river:"exports,attr"`
+}
```
rfratto, line 48: @erikbaranowski I would like to make both arguments and exports blocks before we release, but given #3244 isn't fixed yet, should we merge these as attributes and revisit in a follow up PR once #3244 is fixed?

rfratto (commented):
Nice. The module loader implementation looks pretty clean.

component/module/string/string.go
erikbaranowski: I'd merge as attributes for now. Since we know there is more modules work to do we risk scope creeping this PR

component/module/string/string.go
erikbaranowski: I think that makes sense to experiment with.

erikbaranowski: tested and I like it, so it will be a secret or not based on whether or not the content passed in is a secret. We will want to do something about how the UI renders an entire file in the `is_secret = false` case in a future issue.

erikbaranowski: see #3251 for future UI improvement

docs/sources/flow/reference/components/module.string.md
```diff
@@ -0,0 +1,136 @@
...
+}
+```
+
+## Arguments
+
+The following arguments are supported:
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`content`   | `secret`    | The contents of the module to load as a secret string. | | yes
```
rfratto, line 31:
```suggestion
`content`   | `secret or string` | The contents of the module to load as a secret string. | | yes
```

docs/sources/flow/reference/components/module.string.md
```diff
@@ -0,0 +1,136 @@
+---
+title: module.string
+---
+
+# module.string
+
+`module.string` is a *Module loader* component. A *Module loader* is a Grafana Agent Flow
```
rfratto, line 7: I don't think it's a proper noun. 

Also, after the first usage you can drop the italics.

```suggestion
`module.string` is a *module loader* component. A module loader is a Grafana Agent Flow 
```

docs/sources/flow/reference/components/module.string.md
```diff
@@ -0,0 +1,136 @@
...
+## Arguments
+
+The following arguments are supported:
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`content`   | `secret`    | The contents of the module to load as a secret string. | | yes
+`arguments` | `map(any)`  | The values for the supported arguments in the module contents. | | no
+
+`content` is a string that contains all the arguments, exports and components for the module.
```
rfratto, line 34: Can we say that `content` is a string which contains the configuration source of the module to load?

docs/sources/flow/reference/components/module.string.md
```diff
@@ -0,0 +1,136 @@
...
+---- | ---- | ----------- | ------- | --------
+`content`   | `secret`    | The contents of the module to load as a secret string. | | yes
+`arguments` | `map(any)`  | The values for the supported arguments in the module contents. | | no
+
+`content` is a string that contains all the arguments, exports and components for the module. 
+`content` is typically loaded via the exports of another component. For example,
+
+- local.file.[label].content
+- remote.http.[label].content
+- remote.s3.[label].content
```
rfratto, line 39: Follow Google's style guide for how to [Format placeholders](https://developers.google.com/style/placeholders) here: 

```suggestion
- local.file.LABEL.content
- remote.http.LABEL.content
- remote.s3.LABEL.content
```

docs/sources/flow/reference/components/module.string.md
```diff
@@ -0,0 +1,136 @@
...
+
+`arguments` can contain, but are not limited to strings, components and component exports.
```
rfratto, line 41: I think this needs a few more pieces of detail, primarily: 

* Any argument marked non-optional in the module being loaded must be present in the `arguments` object 
* It is invalid to provide an argument that is not defined as an `argument` block in the loaded module. 

As it is now, it's not clear how you would know what to put in `arugments`. 

Also: I'm not sure we have to say what type of values are supported, we can just say it can be set to any River expression.

docs/sources/flow/reference/components/module.string.md
```diff
@@ -0,0 +1,136 @@
...
+Name | Type | Description
+---- | ---- | -----------
+`exports` | `map(any)` | The exports of the Module loader.
+
+`exports` can contain, but is not limited to strings, components and component exports.
+
+## Component health
+
+`module.string` will be reported as healthy whenever all of its components have
+been loaded successfully.
```
rfratto, line 56: Present tense:

```suggestion
`module.string` is reported as healthy if the most recent load of the module was 
successful. 
```

erikbaranowski (approved):
LGTM

rfratto (commented):
@erikbaranowski feedback on the docs, overall nice work but have some style and content feedback

docs/sources/flow/reference/components/module.string.md
rfratto: Changed to be "`secret`" or `string`" rather than `secret or string`.

docs/sources/flow/reference/components/module.string.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Debug information
+
+`module.string` does not expose any component-specific debug information.
+
+### Debug metrics
+
+`module.string` does not expose any component-specific debug metrics.
+
+## Example
```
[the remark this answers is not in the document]
erikbaranowski: this is now trimmed way down

docs/sources/flow/reference/components/module.string.md
```diff
@@ -0,0 +1,135 @@
...
+
+The following arguments are supported:
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`content`   | `secret` or `string` | The contents of the module to load as a secret or string. | | yes
+`arguments` | `map(any)`  | The values for the supported arguments in the module contents. | | no
+
+`content` is a string that contains the configuration of the module to load.
+`content` is typically loaded via the exports of another component. For example,
```
rfratto, line 35:
```suggestion
`content` is typically loaded by using the exports of another component. For example,
```

docs/sources/flow/reference/components/module.string.md
```diff
@@ -0,0 +1,135 @@
...
+
+- local.file.LABEL.content
+- remote.http.LABEL.content
+- remote.s3.LABEL.content
```
rfratto, line 39:
```suggestion
- `local.file.LABEL.content`
- `remote.http.LABEL.content`
- `remote.s3.LABEL.content`
```

docs/sources/flow/reference/components/module.string.md
```diff
@@ -0,0 +1,135 @@
...
+## Exported fields
+
+The following fields are exported and can be referenced by other components:
+
+Name | Type | Description
+---- | ---- | -----------
+`exports` | `map(any)` | The exports of the Module loader.
+
+`exports` exposes the `export` config block inside a module. It can be accessed from
+the parent river config via `module.string.LABEL.exports.EXPORT_LABEL
```
rfratto, line 57:
```suggestion
the parent config via `module.string.LABEL.exports.EXPORT_LABEL
```

docs/sources/flow/reference/components/module.string.md
```diff
@@ -0,0 +1,135 @@
...
+### Debug metrics
+
+`module.string` does not expose any component-specific debug metrics.
+
+## Example
+
+In this example, we pass credentials from a parent river config to a module for
+a `prometheus.remote_write` flow component. We then export that component for use
+in the parent river config. The export can then be used by the parent river config
+as a component.
```
rfratto, line 82: I found this a little hard to read, WDYT about:

```suggestion
In this example, we pass credentials from a parent config to a module which loads
a `prometheus.remote_write` component. The exports of the 
`prometheus.remote_write` component are exposed to parent config, allowing 
the parent config to pass metrics to it.
```

docs/sources/flow/reference/components/module.string.md
```diff
@@ -0,0 +1,135 @@
...
+Module:
+
+```river
+argument "username" { }
+
+argument "password" { }
+
+export "prometheus_remote_write" {
+	value = prometheus.remote_write.grafana_cloud
+}
```
rfratto, line 117: Huh, I didn't realize this would work honestly, but I understand why it does.

docs/sources/flow/reference/components/module.string.md
```diff
@@ -0,0 +1,135 @@
...
+		}
+	}
+}
+```
+
+> **NOTE**: For the purpose of the above example we have exported the entire
+> `prometheus.remote_write.grafana_cloud` component. Alternatively, the module above
+> could have exported `prometheus.remote_write.grafana_cloud.receiver` instead of
+> the entire component since that is the only part of the component we need in the
+> parent river config.
```
rfratto, line 135: I think we can drop this note, or have it only export the `receiver` if that will be the most common thing to do.

rfratto (commented):
@erikbaranowski final round of feedback!

docs/sources/flow/reference/components/module.string.md
erikbaranowski: I can drop the note. I think most common will be people exporting full components but we will see!

PR #3229, opened 2023-03-09
Introduce documentation to explain modules

### LINKED ISSUE #3105: Flow Modules: Documentation for concept of modules
Document modules more concretely along with some example use cases.

rfratto:
This PR adds concept-level documentation for Flow modules, and configuration block reference pages for `argument` and `export`.

I'm a little worried that this isn't being explained very well; let me know if there's anything I can do here to help make things clearer.

Closes #3105.

### DIFF
--- a/docs/sources/flow/concepts/modules.md
+++ b/docs/sources/flow/concepts/modules.md
@@ -47,7 +47,7 @@ refer to the documentation for the module loader you are using for more
 information.
 
 [Component controller]: {{< relref "./component_controller.md" >}}
-[Component]: {{< relref "../reference/components/" >}}
+[Components]: {{< relref "../reference/components/" >}}
 
 ## Module sources
 
@@ -84,7 +84,7 @@ This example module manages a pipeline which filters out debug- and info-level
 log lines which are given to it:
 
 ```river
-// argument.write_to is an optional argument which specifies where filtered
+// argument.write_to is a required argument which specifies where filtered
 // log lines should be sent.
 //
 // The value of the argument can be retrieved in this file with
--- a/docs/sources/flow/concepts/modules.md
+++ b/docs/sources/flow/concepts/modules.md
@@ -1,0 +1,145 @@
+---
+aliases:
+- ../../concepts/modules/
+title: Modules
+weight: 300
+---
+
+# Modules
+
+_Modules_ are a way to create Grafana Agent Flow configuration files which can
+laoded as a component. Modules are a great way to parameterize configuration
+files to create reusable pipelines.
+
+Modules are Grafana Agent Flow configuration files which have:
+
+* Arguments: settings which configure a module.
+* Exports: named values which a module exposes to the consumer of the module.
+* Components: Grafana Agent Flow Components to run when the module is running.
+
+Modules are loaded into Grafana Agent Flow by using a [Module
+loader](#module-loaders).
+
+Refer to the documentation for the [argument block][] and [export block][] to
+learn how to define arguments and exports for a module.
+
+[argument block]: {{< relref "../reference/config-blocks/argument.md" >}}
+[export block]: {{< relref "../reference/config-blocks/export.md" >}}
+
+## Module loaders
+
+A _Module loader_ is a Grafana Agent Flow component which retreives a module
+and runs the components defined inside of it.
+
+Module loader components are responsible for:
+
+* Retreiving the module source to run.
+* Creating a [Component controller][] for the module to run in.
+* Passing arguments to the loaded module.
+* Exposing exports from the loaded module.
+
+Module loaders typically are called `module.LOADER_NAME`. The list of module
+loader components can be found in the list of Grafana Agent Flow
+[Components][].
+
+Some module loaders may not support running modules with arguments or exports;
+refer to the documentation for the module loader you are using for more
+information.
+
+[Component controller]: {{< relref "./component_controller.md" >}}
+[Component]: {{< relref "../reference/components/" >}}
+
+## Module sources
+
+Modules are designed to be flexible, and can have their configuration retrieved
+from anywhere, such as:
+
+* The local filesystem
+* An S3 bucket
+* An HTTP endpoint
+
+Each module loader component will support different ways of retrieving module
+sources. The most generic module loader component, `module.string`, can load
+modules from the export of another Flow component:
+
+```river
+local.file "my_module" {
+  filename = "PATH_TO_MODULE"
+}
+
+module.string "my_module" {
+  source = local.file.my_module.content
+
+  arguments = {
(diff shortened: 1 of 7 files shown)


docs/sources/flow/concepts/modules.md
```diff
@@ -0,0 +1,145 @@
...
+module.string "log_filter" {
+  source = local.file.log_filter.content
+
+  arguments = {
+    // Configure the filter to forward filtered logs to loki.echo below.
+    logs_output = [loki.write.default.receiver],
+  }
+}
+
+loki.write "default" {
```
captncraig, line 140: maybe reorder the components here in order of references? I'd expect the loki.write up top, then the file, the module, and the source.

docs/sources/flow/concepts/modules.md
```diff
@@ -0,0 +1,145 @@
+---
+aliases:
+- ../../concepts/modules/
+title: Modules
+weight: 300
+---
+
+# Modules
+
+_Modules_ are a way to create Grafana Agent Flow configuration files which can
```
captncraig, line 10:
```suggestion
_Modules_ are a way to create Grafana Agent Flow configuration files which can be
```

docs/sources/flow/concepts/modules.md
```diff
@@ -0,0 +1,145 @@
...
+laoded as a component. Modules are a great way to parameterize configuration
+files to create reusable pipelines.
+
+Modules are Grafana Agent Flow configuration files which have:
```
captncraig, line 14: We should perhaps avoid calling them files if they are not necessarily files. Is there a better word here?

captncraig (approved):

docs/sources/flow/concepts/modules.md
rfratto: Maybe just "configurations?"

docs/sources/flow/concepts/modules.md
```diff
@@ -0,0 +1,145 @@
...
+[Components][].
+
+Some module loaders may not support running modules with arguments or exports;
+refer to the documentation for the module loader you are using for more
+information.
+
+[Component controller]: {{< relref "./component_controller.md" >}}
+[Component]: {{< relref "../reference/components/" >}}
+
+## Module sources
```
erikbaranowski, line 52: are we going with `source` over `content`? I don't have an opinion but think we should update the rfc if going with `source`.

rfratto: I'll change this to sync up with whatever the merged component uses, I just went with what's in the PR for now

docs/sources/flow/reference/config-blocks/argument.md
```diff
@@ -0,0 +1,70 @@
...
+The module loader is responsible for determining the values for arguments.
+Components in a module may use `argument.ARGUMENT_NAME.value` to retrieve the
+value provided by the module loader.
+
+## Example
+
+This example creates a module where agent metrics are collected. Collected
+metrics are then forwarded to the argument specified by the loader:
+
+```river
```
erikbaranowski, line 57: it may start to get slightly redundant, but I feel like we should have a 2nd config showing how to pass the argument into the module as part of this example for someone looking for an isolated example of using an argument end to end

docs/sources/flow/reference/config-blocks/export.md
```diff
@@ -0,0 +1,60 @@
...
+## Exported fields
+
+The `export` block does not export any fields.
+
+## Example
+
+This example creates a module where the output of discovering Kubernetes pods
+and nodes are exposed to the module loader:
+
+```river
```
erikbaranowski, line 45: it may start to get slightly redundant, but I feel like we should have a 2nd config showing how to reference the export in the parent config for an isolated example of using an export end to end

rfratto: > but I feel like we should have a 2nd config showing how to reference the export in the parent config for an isolated example of using an export end to end

I agree, but part of my concern here is that we'll have different module loaders which do things differently. I'm not sure if we're able to commit to a generic example right now, and we'll have to push some of the responsibility for usage to the component documentation per loader.

rfratto: For example, I'd like to see a `module.kubernetes` loader which can discover FlowModule CRDs on Kubernetes and run them, but those (probably) won't support export blocks at all since the module loader will find a dynamic number of modules.

docs/sources/flow/concepts/modules.md
rfratto: I'm going to leave this as-is mainly because it's a stylistic choice and we don't recommend a specific way yet; I normally write components in terms of the data flow top-down, with some exceptions like putting local.files before their usage.

docs/sources/flow/reference/config-blocks/argument.md
rfratto: I'd like to leave this out of the config block reference because, at the moment, how an argument block may be used isn't guaranteed to be the same across module loaders.

docs/sources/flow/reference/config-blocks/argument.md
```diff
@@ -0,0 +1,70 @@
+---
+title: argument
+---
+
+# argument block
+
+`argument` is an optional configuration block used to specify parameterized
+input to a [Module][Modules]. `argument` blocks must be given a label which
+determine the name of the argument.
```
erikbaranowski, line 9:
```suggestion
determines the name of the argument.
```

docs/sources/flow/concepts/modules.md
```diff
@@ -0,0 +1,145 @@
...
+Module loaders typically are called `module.LOADER_NAME`. The list of module
+loader components can be found in the list of Grafana Agent Flow
+[Components][].
+
+Some module loaders may not support running modules with arguments or exports;
+refer to the documentation for the module loader you are using for more
+information.
+
+[Component controller]: {{< relref "./component_controller.md" >}}
+[Component]: {{< relref "../reference/components/" >}}
```
erikbaranowski, line 50:
```suggestion
[Components]: {{< relref "../reference/components/" >}}
```

docs/sources/flow/concepts/modules.md
```diff
@@ -0,0 +1,145 @@
...
+}
+```
+
+## Example module
+
+This example module manages a pipeline which filters out debug- and info-level
+log lines which are given to it:
+
+```river
+// argument.write_to is an optional argument which specifies where filtered
```
erikbaranowski, line 87: Does this comment make sense? The argument.write_to is set as `optional = false` right below

erikbaranowski (approved):
LGTM, thanks for sorting those TODOs in module.string.md to save us another cycle on docs.

PR #3230, opened 2023-03-09
Fix issue with evaluating expressions

### LINKED ISSUE #3222: River: Parts of expression are ignored
When parsing the `{ a = 15 }.7` expression, it gets formatted as `map[string]interface {}` with one element - the string key of "a" and an int value of 15. The `.7` part is seemingly ignored. Instead, I'd have expected an error to be thrown that there is no element `7`. 

Similarly, the result of the expression `7.8.5` is a float64 value of 7.8 and the `.5` part is ignored. I'd have expected to see an error for the `.5` part here too.

ptodev:
#### PR Description
Fixes the issue mentioned in #3222.  The ParseExpression() function on the main branch does not currently raise an error if trailing parts of the expressions were not evaluated.

Fixes #3222

#### Notes to the reviewer
Apparently in River it's possible to define a float value without specifying the 0 before the dot. For example, `.8` is equivalent to `0.8`.

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,7 +50,7 @@ Main (unreleased)
 
 - Flow: fix issue where Flow would return an error when trying to access a key
   of a map whose value was the zero value (`null`, `0`, `false`, `[]`, `{}`).
-  Whether an error was returned dependend on the internal type of the value.
+  Whether an error was returned depended on the internal type of the value.
   (@rfratto)
 
 - Flow: fix issue where using the `jaeger_remote` sampler for the `tracing`
--- a/pkg/river/parser/parser.go
+++ b/pkg/river/parser/parser.go
@@ -32,6 +32,8 @@ func ParseExpression(expr string) (ast.Expr, error) {
 
 	e := p.ParseExpression()
 
+	// If the current token is not a TERMINATOR then the parsing did not complete
+	// in full and there are still parts of the string left unparsed.
 	p.expect(token.TERMINATOR)
 
 	if len(p.diags) > 0 {
--- a/pkg/river/parser/parser.go
+++ b/pkg/river/parser/parser.go
@@ -32,9 +32,7 @@ func ParseExpression(expr string) (ast.Expr, error) {
 
 	e := p.ParseExpression()
 
-	if p.tok != token.TERMINATOR && p.tok != token.EOF {
-		p.addErrorf("did not parse the expression in full - unexpected token of type %s", p.tok.String())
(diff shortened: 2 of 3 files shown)


pkg/river/parser/parser.go
```diff
@@ -30,6 +31,11 @@ func ParseExpression(expr string) (ast.Expr, error) {
 	p := newParser("", []byte(expr))
 
 	e := p.ParseExpression()
+
+	if p.tok != token.TERMINATOR && p.tok != token.EOF {
+		p.addErrorf("did not parse the expression in full - unexpected token of type %s", p.tok.String())
+	}
```
rfratto, line 37: I'm not totally sure on this one. I'd like to use `p.expect` because that's the standard way we assert a token. 

Expressions should _always_ end in token.TERMINATOR, and a token.TERMINATOR is expected to be injected prior to token.EOF (if it doesn't, that could be another bug that needs investigating).

I think it's safe to replace this entire section with `p.expect(token.TERMINATOR)`.

pkg/river/parser/parser.go
```diff
@@ -30,6 +31,9 @@ func ParseExpression(expr string) (ast.Expr, error) {
 	p := newParser("", []byte(expr))
 
 	e := p.ParseExpression()
+
+	p.expect(token.TERMINATOR)
+
```
rfratto, line 36: Should we add a comment here explaining why this is necessary?

CHANGELOG.md
```diff
@@ -34,9 +34,13 @@ Main (unreleased)
 
 - Flow: fix issue where Flow would return an error when trying to access a key
   of a map whose value was the zero value (`null`, `0`, `false`, `[]`, `{}`).
-  Whether an error was returned dependend on the internal type of the value.
+  Whether an error was returned depended on the internal type of the value.
   (@rfratto)
 
+- River: fix issue with evaluating expressions where if a trailing part of the 
+  expression is invalid, no error was raised. This issue does not affect Flow 
+  because it does not call ParseExpression() directly. (@ptodev)
```
rfratto, line 42: Since it's not exposed to Flow, it's not a user-facing change; can we take this out?

rfratto (approved):

rfratto (approved):
LGTM, thanks!

rfratto:
@ptodev Can you merge in main to fix the build issues?

PR #3231, opened 2023-03-09
helm: use consistent versioning when rendering chart tests

rfratto:
This adds an undocumented value, `$chart_tests`, which changes how the charts get rendered when creating the manifests. When `$chart_tests` is true, the chart identifier has its version removed and the app.kubernetes.io/version label is forced to vX.Y.Z.

This will be useful for reducing the size of PRs which update the Helm chart to focus on only the important changes.


captncraig (approved):
YES!

PR #3240, opened 2023-03-11
flow/tracing: fix issue where remote sampler could not be parsed

rfratto:
This fixes an issue where the remote sampler could not be parsed if it used strings for the strategy type instead of numbers. The resolution is to switch to the protobuf JSON encoding, which properly parses this type as a string.

Not being able to parse the remote sampler rules meant that sampling rate fell back to the default (10%).

This change is equivalent to
open-telemetry/opentelemetry-go-contrib#3183. We are still unable to move to the upstream type while
open-telemetry/opentelemetry-go-contrib#2981 remains unresolved.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,11 @@ Main (unreleased)
   Whether an error was returned dependend on the internal type of the value.
   (@rfratto)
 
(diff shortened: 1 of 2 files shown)


spartan0x117 (approved):
LGTM since it's pretty much identical to the otel change linked

PR #3248, opened 2023-03-13
Add redis address to usage docs

ptodev:
`redis_addr` should be listed in the "usage" section because it is a mandatory attribute.

### DIFF
--- a/component/prometheus/exporter/redis/redis.go
+++ b/component/prometheus/exporter/redis/redis.go
@@ -45,7 +45,7 @@ type Config struct {
 	//
 	// The exporter binary config differs to this, but these
 	// are the only fields that are relevant to the exporter struct.
-	RedisAddr               string            `river:"redis_addr,attr,optional"`
+	RedisAddr               string            `river:"redis_addr,attr"`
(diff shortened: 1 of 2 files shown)


ptodev:
Interesting, I just realised this attribute is optional in the component's river tag:
https://github.com/grafana/agent/blob/main/component/prometheus/exporter/redis/redis.go#L48

tpaschalis:
> Interesting, I just realised this attribute is optional in the component's river tag:

Then we should either make it mandatory in the River struct as well, or leave it out from the example configuration as we only include the required fields in the minimal examples.

ptodev:
> > Interesting, I just realised this attribute is optional in the component's river tag:
> 
> Then we should either make it mandatory in the River struct as well, or leave it out from the example configuration as we only include the required fields in the minimal examples.

The component would definitely [error](https://github.com/grafana/agent/blob/v0.32.1/pkg/integrations/redis_exporter/redis_exporter.go#L134) if this is not present, so I'll remove the "optional" label from the river tags.

spartan0x117 (approved):

Issue #3107, opened 2023-02-28
Flow Modules: Add metrics if needed


mattdurham:
Investigate if there are other places to add metrics specific to modules.

PR #3253, opened 2023-03-13
flow: experiment with explicit component telemetry types

rfratto:
> **NOTE TO REVIEWER**: this PR is easier to follow if reviewed commit-by-commit instead of looking at the entire diff. 

This PR prototypes a solution for #3107 by creating a logging type, `github.com/grafana/agent/pkg/flow/logger`, with two main loggers:

* `logger.Controller` 
* `logger.Component` 

Each of these is created from a Sink, which can be one of: 

* An io.Writer 
* A Controller logger 
* A Component logger 

Sinks propagate the component ID correctly, allowing Controller loggers to be created from Component loggers without losing the ID of the component which created the controller. 

An alternative approach could be to give generic telemetry types to components, with no automatic label injection of the component ID. This would make it much easier to create controllers, but would also make it much easier for components to do the wrong thing and trample over metrics generated by another component, causing scrapes to fail. 

This approach is a more complicated API, but I appreciate how it makes it hard to do the wrong thing when creating controllers.

### DIFF
--- a/pkg/flow/logger/component.go
+++ b/pkg/flow/logger/component.go
@@ -9,6 +9,8 @@ import (
 // Component is a logger passed to Grafana Agent Flow components. It implements
 // the [log.Logger] interface.
 type Component struct {
+	componentID string
+
 	orig log.Logger // Original logger before the component name was added.
 	log  log.Logger // Logger with component name injected.
 }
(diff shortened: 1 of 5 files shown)


erikbaranowski:
would it make sense to name them ComponentLogger (Logger) and FlowLogger/FlowControllerLogger (Subsytem) to differentiate the 2. Then we would have for example

```
NewFromComponentLogger(l *ComponentLogger) *FlowLogger { ... }
```

instead of

```
func NewFromLogger(l *Logger) *Subsystem { ... }
```

rfratto:
Hm, maybe. I'd like to avoid making the names too long because then it's a bit tedious to read/write, but I'll think about if there's a pair of names other than Subsystem/Logger which makes sense here.

rfratto:
@erikbaranowski What about renaming the package to `logger` and then having `logger.Component` and `logger.Controller` with methods like `logger.NewForComponent` and `logger.NewForController`?

rfratto:
@erikbaranowski The last commit added the new API for us to see if we like it, PTAL and let me know what you think

PR #3255, opened 2023-03-13
modules will fail to load if they contain a singleton

### LINKED ISSUE #3111: Flow Modules: Do not allow modules to load singletons

erikbaranowski:
#### PR Description

Introduces the idea of whether a flow controller is a module or not and uses it to prevent singletons in modules.

Closes #3111

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated


pkg/flow/flow.go
```diff
@@ -112,6 +112,9 @@ type Options struct {
 	// OnExportsChange is nil, export configuration blocks are not allowed in the
 	// loaded config file.
 	OnExportsChange func(exports map[string]any)
+
+	// IsModule is set to true if the Flow Controller is a module.
+	IsModule bool
```
rfratto, line 117: Do we want to make this another option, or do we want to assume that ControllerID being non empty means the controller is for a module?

pkg/flow/flow.go
```diff
@@ -261,7 +264,7 @@ func (c *Flow) LoadFile(file *File, args map[string]any) error {
 		},
 	}
 
-	diags := c.loader.Apply(argumentScope, file.Components, file.ConfigBlocks, c.opts.OnExportsChange)
+	diags := c.loader.Apply(argumentScope, file.Components, file.ConfigBlocks, c.opts.OnExportsChange, c.opts.IsModule)
```
rfratto, line 267: This should probably be passed when the loader is constructed (since it never changes) so we don't just keep piling on more and more arguments here. 

Honestly, `c.opts.OnExportsChange` should also probably be moved too.

rfratto (approved):
Clean 👍 LGTM!

PR #3256, opened 2023-03-13
operator: allow setting runtimeClassName for operator generated pods.

captncraig:


jcreixell:
could you provide some context on why this is needed and how it should be used? (we allow setting the value, but where do we read it and what for?)

captncraig:
[runtime class](https://kubernetes.io/docs/concepts/containers/runtime-class/) Is a way to specify the container runtime a pod should use, I think useful in hybrid clusters, or where additional security features may be available to different pods. 

It is a simple setting on the pod that operator users can't set by themselves, so we simply pass it through.

jcreixell (approved):
got it, thanks for the context, LGTM

PR #3259, opened 2023-03-13
flow/logging: rewrite logging subsystem

rfratto:
> **NOTE TO REVIEWER**: This PR is much easier to review commit-by-commit rather than looking at the change as a whole. 

This commit rewrites the Flow logging subsystem so it can be used as part of the solution for #3107. 

The new logging subsystem is composed of:

* A logging `Sink`: a place to send emitted logs.
* A `Logger`: a `go-kit/log.Logger` implementation which sends logs to a configured sink.

Compared to the previous logging subsystem, this breaks apart logging.Logger into the Sink and Logger pairs. Now, the Sink is what supports being updated with new Options.

Additionally, Loggers may now be constructed with options, including assigning a Component ID.

There are two different types of sinks:

* A `WriterSink` formats and writes logs to an `io.Writer`.
* A `LoggerSink` forwards logs to another `*Logger` instance.

When `LoggerSink` is used, the component ID of the Logger (if any) propagates properply to loggers generated with that sink, ensuring a global ID is generated.

The Flow controller and components have been updated to use this API explicitly to handle part of grafana/agent#3107.


rfratto:
(I won't keep this in a /v2 folder past this PR, it's just there right now to make this easier to review)

pkg/flow/logging/v2/sink.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+)
+
+// Sink is where a Controller logger will send log lines to.
+type Sink struct {
+	w         io.Writer // Raw writer to use
+	updatable bool      // Whether the sink supports being updated.
+
+	parentComponentID string // Whether the sink has a parent component ID associated with it.
```
erikbaranowski, line 16: nit: this comment seems a little odd for a non-bool

erikbaranowski (approved):
LGTM

PR #3262, opened 2023-03-14
Add backport workflow

jdbaldry:
The workflow automatically backports PRs to branches labeled with `backport <BRANCH>` on closure or labeling of PRs.

Requires GH_BOT_ACCESS_TOKEN to work and in the future this will need to be replaced by a GitHub App based workflow. Timeline TBC.

PR #3263, opened 2023-03-14
Avoid ambiguous relref lookups by forcing relative resolution

jdbaldry:
Follow up from #3233 where I left some behind.

Again needs to be backported all the way back to v0.28 but we can maybe use the workflow in #3262 for that.


rfratto (approved):
Thanks!

PR #3265, opened 2023-03-14
web/api: fix Flow UI for components with no arguments

rfratto:
Components with no arguments returned a `null` for the arguments block, but the UI contract expects that arguments is always an array.

cc @rgeyer


erikbaranowski (approved):
LGTM

PR #3289, opened 2023-03-15
Add error if the logging or tracing config block exists in a module

### LINKED ISSUE #3254: Flow Modules: Do not allow modules to load config blocks for tracing or logging

erikbaranowski:
Return err if a module component fails to start in module.string.go

Closes #3254 

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/flow/reference/components/module.string.md
+++ b/docs/sources/flow/reference/components/module.string.md
@@ -70,8 +70,11 @@ source.
 
 ## Component health
 
-`module.string` is reported as unhealthy if given an invalid configuration or if
-it fails to start any of the components defined inside of it.
+`module.string` is reported as healthy if the most recent load of the module was
+successful.
+
+If the module is not loaded successfully, the current health displays as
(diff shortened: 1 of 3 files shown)


docs/sources/flow/reference/components/module.string.md
```diff
@@ -70,11 +70,8 @@ source.
 
 ## Component health
 
-`module.string` is reported as healthy if the most recent load of the module was
-successful.
-
-If the module is not loaded successfully, the current health displays as
-unhealthy and the health includes the error from loading the module.
+`module.string` is reported as unhealthy if given an invalid configuration or if
```
rfratto, line 73: I feel like the new documentation here is hiding information. Nothing has changed, right? `module.string` is unhealthy if the most recent load of the module succeeded. 

Whether components can _start_ isn't part of the health definition, just if they can be created.

Issue #3029, opened 2023-02-19
Proposal: add aliases to River


tpaschalis:
River may benefit from being able to define references to values so that they can be reused and shared across different parts of a configuration file.

These aliases can be helpful to reducing deduplication when repeating the same values or expressions when writing complex configuration and give a degree of certainty that all references are updated in one go. 

At the same time, if overused, they can make a configuration file hard to read by hiding the actual values under one or more layers of indirection. For this reason, aliases should only be used in moderation, in situations where a single value or result is used in many places, or that should only be changed from a central place to help maintain a readable configuration file.

## Proposal
I propose that we change River to allow a new top-level statement that _aliases_, or assigns a _name_ to an expression, so that it can be reused multiple times throughout a configuration file.
An alias should be nothing more than a shorthand to the matching expression, and would use pass-by-value semantics.

### Defining aliases
Defining aliases can make use of the `let` keyword plus the alias name on the LHS, while the RHS must be any valid River expression.
```
let ALIAS = EXPRESSION
```
Such definitions can only appear as top-level River statements.
Each line can only define a single alias.
Aliases must be valid River identifiers.
Alias definitions cannot be recursive.
Aliases cannot conflict with any component or top-level block names.
The expression concept in River should now be changed to include references to aliases, similar to component exports.

### Using aliases
Aliases can only be used in the context of a River expression.

My initial proposal is that we wouldn’t introduce any new symbols or a special way for referring to aliased values (for example, avoid having `$name`, `locals.name` or `vals.name`) but that they should be used as an IdentifierExpr, much like components exports do.
## Usage
Here’s how a configuration file that makes use of aliases could look like.
The first alias is a static value that can be reused throughout the configuration file. The second and third aliases are the results of expressions that make use of River’s standard library and component exports.

```
let lvl = “debug”
let rw  = [prometheus.remote_write.staging.receiver]
let tgt = json_decode(remote.http.targets.content)

logging {
  log_level = lvl
}

remote.http "targets" {
  url = env("MY_TARGETS_URL")
}

prometheus.remote_write "staging" {
  endpoint {
    url = env(“RW_URL”)
  }
}

prometheus.scrape "app1" {
  targets    = tgt
  forward_to = rw
}
```

## Implementation
Aliases can be stored along component arguments and exports in the controller’s valueCache under a global NodeID.
[...]

PR #3297, opened 2023-03-16
Independent Config Blocks v1

### LINKED ISSUE #3223: Make config blocks independent nodes in the DAG
Right now, all config blocks are treated as [one giant node](https://github.com/grafana/agent/blob/a8e1d2bc53fa01c34b1501466a49b931ee18e99a/pkg/flow/internal/controller/config.go#L55) in the DAG. 

Ideally, we have one node per config block (one for logging, one for tracing, and one per export) so evaluating one of those blocks doesn't force the rest to evaluate, and so that they don't share unnecessary DAG dependencies.

erikbaranowski:
This is probably best viewed commit by commit.

This implements an approach to independent config blocks. It is WIP and has more prototyping coming until this comment has been modified to say it doesn't.

This is fully implemented and probably needs some additions
- unit testing written for config block scenarios
- a pass on comments in code
- any caching things that need to be fixed for reevaluation/reloads

Fixes #3223

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/flow/internal/controller/component.go
+++ b/pkg/flow/internal/controller/component.go
@@ -227,7 +227,7 @@ func (cn *ComponentNode) UpdateBlock(b *ast.BlockStmt) {
 	cn.eval = vm.New(b.Body)
 }
 
-// Evaluate implements dag.Node and updates the arguments for the managed component
+// Evaluate implements BlockNode and updates the arguments for the managed component
 // by re-evaluating its River block with the provided scope. The managed component
 // will be built the first time Evaluate is called.
 //
@@ -334,7 +334,7 @@ func (cn *ComponentNode) Arguments() component.Arguments {
 	return cn.args
 }
 
-// Block implements dag.Node and returns the current block of the managed component.
+// Block implements BlockNode and returns the current block of the managed component.
 func (cn *ComponentNode) Block() *ast.BlockStmt {
 	cn.mut.RLock()
 	defer cn.mut.RUnlock()
--- a/pkg/flow/internal/controller/component.go
+++ b/pkg/flow/internal/controller/component.go
@@ -228,9 +228,9 @@ func (cn *ComponentNode) UpdateBlock(b *ast.BlockStmt) {
 	cn.eval = vm.New(b.Body)
 }
 
-// Evaluate updates the arguments for the managed component by re-evaluating
-// its River block with the provided scope. The managed component will be built
-// the first time Evaluate is called.
+// Evaluate implements dag.Node and updates the arguments for the managed component
+// by re-evaluating its River block with the provided scope. The managed component
+// will be built the first time Evaluate is called.
 //
 // Evaluate will return an error if the River block cannot be evaluated or if
 // decoding to arguments fails.
@@ -335,7 +335,7 @@ func (cn *ComponentNode) Arguments() component.Arguments {
 	return cn.args
 }
 
-// Block returns the current block of the managed component.
+// Block implements dag.Node and returns the current block of the managed component.
 func (cn *ComponentNode) Block() *ast.BlockStmt {
 	cn.mut.RLock()
 	defer cn.mut.RUnlock()
--- a/pkg/flow/internal/controller/component.go
+++ b/pkg/flow/internal/controller/component.go
@@ -14,7 +14,6 @@ import (
 
 	"github.com/go-kit/log/level"
 	"github.com/grafana/agent/component"
-	"github.com/grafana/agent/pkg/flow/internal/dag"
 	"github.com/grafana/agent/pkg/flow/logging"
 	"github.com/grafana/agent/pkg/river/ast"
 	"github.com/grafana/agent/pkg/river/vm"
@@ -107,7 +106,7 @@ type ComponentNode struct {
 	exports    component.Exports // Evaluated exports for the managed component
 }
 
-var _ dag.Node = (*ComponentNode)(nil)
+var _ BlockNode = (*ComponentNode)(nil)
 
 // NewComponentNode creates a new ComponentNode from an initial ast.BlockStmt.
 // The underlying managed component isn't created until Evaluate is called.
--- a/pkg/flow/internal/controller/component.go
+++ b/pkg/flow/internal/controller/component.go
@@ -335,6 +335,13 @@ func (cn *ComponentNode) Arguments() component.Arguments {
 	return cn.args
 }
 
+// Block returns the current block of the managed component.
+func (cn *ComponentNode) Block() *ast.BlockStmt {
+	cn.mut.RLock()
+	defer cn.mut.RUnlock()
+	return cn.block
+}
+
 // Exports returns the current set of exports from the managed component.
 // Exports returns nil if the managed component does not have exports.
 func (cn *ComponentNode) Exports() component.Exports {
--- a/pkg/flow/internal/controller/config.go
+++ b/pkg/flow/internal/controller/config.go
@@ -23,22 +23,14 @@ const (
 // The graph will always have _exactly one_ instance of ConfigNode, which will
 // be used to contain the state of all config blocks.
 type ConfigNode struct {
-	globals ComponentGlobals
-	id      ComponentID
+	label         string
+	nodeID        string
+	componentName string
+	globals       ComponentGlobals
 
-	mut          sync.RWMutex
-	block        *ast.BlockStmt      // Current River blocks to derive config from
-	loggingArgs  logging.SinkOptions // Evaluated logging arguments for the config
-	loggingBlock *ast.BlockStmt
-	loggingEval  *vm.Evaluator
-	tracingArgs  tracing.Options
-	tracingBlock *ast.BlockStmt
-	tracingEval  *vm.Evaluator
-	exportBlocks map[string]*ast.BlockStmt
-	exportEvals  map[string]*vm.Evaluator
-
-	// Set to true for configuration blocks that were provided.
-	foundLogging, foundTracing bool
+	mut   sync.RWMutex
+	block *ast.BlockStmt // Current River blocks to derive config from
+	eval  *vm.Evaluator
 }
 
 // ConfigBlockID returns the string name for a config block.
@@ -54,24 +46,12 @@ func NewConfigNode(block *ast.BlockStmt, globals ComponentGlobals, isInModule bo
 	var (
 		diags diag.Diagnostics
 
-		loggingBlock ast.BlockStmt
-		tracingBlock ast.BlockStmt
-
-		exportBlocks = make(map[string]*ast.BlockStmt)
-		exportEvals  = make(map[string]*vm.Evaluator)
-
-		foundLogging = false
-		foundTracing = false
-	)
-
-	var (
-		name = strings.Join(block.Name, ".")
+		name   = strings.Join(block.Name, ".")
+		nodeID = BlockComponentID(block).String()
 	)
 
 	switch name {
 	case loggingBlockID:
-		loggingBlock = *block
-		foundLogging = true
 		if isInModule {
 			diags.Add(diag.Diagnostic{
 				Severity: diag.SeverityLevelError,
@@ -81,8 +61,6 @@ func NewConfigNode(block *ast.BlockStmt, globals ComponentGlobals, isInModule bo
 			})
 		}
 	case tracingBlockID:
-		tracingBlock = *block
-		foundTracing = true
 		if isInModule {
 			diags.Add(diag.Diagnostic{
 				Severity: diag.SeverityLevelError,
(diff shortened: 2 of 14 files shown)


pkg/flow/internal/controller/loader.go
```diff
@@ -182,18 +158,53 @@ func (l *Loader) Apply(parentScope *vm.Scope, blocks []*ast.BlockStmt, configBlo
...
+	// debug information.
+	l.originalGraph = g.Clone()
+	// Perform a transitive reduction of the graph to clean it up.
+	dag.Reduce(&g)
+
+	return g, diags
+}
+
+func (l *Loader) populateConfigBlockNodes(g *dag.Graph, configBlocks []*ast.BlockStmt, globals ComponentGlobals, isInModule bool) diag.Diagnostics {
+	c, diags := NewConfigNode(configBlocks, l.globals, l.isModule())
```
erikbaranowski, line 196: This is where we would loop on the configBlocks instead of inside NewConfigNode. NewConfigNode would only take in a single configBlock instead of the whole slice.

erikbaranowski:
Proposal: ConfigNode would be refactored to have a single *ast.BlockStmt instead of []*ast.BlockStmt. This will result in some significant changes to it as a whole and many of these properties would be refactored and the logic for them would have to exist elsewhere (some likely in populateConfigBlockNodes proposed above)

current definition of ConfigNode for reference

```golang
type ConfigNode struct {
	globals ComponentGlobals

	mut          sync.RWMutex
	blocks       []*ast.BlockStmt    // Current River blocks to derive config from
	loggingArgs  logging.SinkOptions // Evaluated logging arguments for the config
	loggingBlock *ast.BlockStmt
	loggingEval  *vm.Evaluator
	tracingArgs  tracing.Options
	tracingBlock *ast.BlockStmt
	tracingEval  *vm.Evaluator
	exportBlocks map[string]*ast.BlockStmt
	exportEvals  map[string]*vm.Evaluator

	// Set to true for configuration blocks that were provided.
	foundLogging, foundTracing bool
}
```

pkg/flow/internal/controller/loader_test.go
```diff
@@ -124,15 +124,8 @@ func TestLoader(t *testing.T) {
 		require.Error(t, diags.ErrorOrNil())
 
 		requireGraph(t, l.Graph(), graphDefinition{
```
erikbaranowski, line 126: see [here](https://github.com/grafana/agent/pull/3297/files#r1139349663) for context

pkg/flow/internal/dag/dag.go
```diff
@@ -1,12 +1,28 @@
...
-import "fmt"
+import (
+	"fmt"
+
+	"github.com/grafana/agent/pkg/river/ast"
+	"github.com/grafana/agent/pkg/river/vm"
+)
 
 // Node is an individual Vertex in the DAG.
 type Node interface {
```
rfratto, line 12: I'd like to keep the Node interface hyper-generic to leave the door open in the future for Nodes in the future which don't have blocks (i.e., #3029).

Instead, can you make a new interface in `pkg/flow/internal/controller` for this?

```go 
package controller 

import (
  "github.com/grafana/agent/pkg/flow/internal/dag"
)

// BlockNode is a node in the DAG which manages a River block 
// and can be evaluated.
type BlockNode interface {
	dag.Node 

	// Evaluate updates the arguments for the managed component
	// by re-evaluating its River block with the provided scope. The managed component
	// will be built the first time Evaluate is called.
	//
	// Evaluate will return an error if the River block cannot be evaluated or if
	// decoding to arguments fails.
	Evaluate(scope *vm.Scope) error

	// Block returns the current block of the managed config node.
	Block() *ast.BlockStmt
}
```

This means you may have to keep some of the type checks in, but instead you can type check based on the interface:

```go
switch cn := cn.(type) {
case BlockNode:
	traversals = expressionsFromBody(cn.Block().Body)
}
```

pkg/flow/internal/controller/config.go
```diff
@@ -6,249 +6,52 @@ import (
...
 	loggingBlockID = "logging"
 	tracingBlockID = "tracing"
-	exportBlockID  = "export"
 )
 
-// ConfigNode is a controller node which manages agent configuration.
-// The graph will always have _exactly one_ instance of ConfigNode, which will
-// be used to contain the state of all config blocks.
+// Shared config properties for all config block nodes
 type ConfigNode struct {
```
rfratto, line 21: At this point this is a helper type so I'd unexport it as `configNode`.

pkg/flow/internal/controller/config_export.go
```diff
@@ -0,0 +1,81 @@
...
+// Evaluate implements dag.Node and updates the arguments for the managed config block
+// by re-evaluating its River block with the provided scope. The managed config block
+// will be built the first time Evaluate is called.
+//
+// Evaluate will return an error if the River block cannot be evaluated or if
+// decoding to arguments fails.
+func (cn *ExportConfigNode) Evaluate(scope *vm.Scope) error {
+	cn.configNode.mut.Lock()
+	defer cn.configNode.mut.Unlock()
+	exports := make(map[string]any, 1)
```
rfratto, line 58: Is there a reason we always allocate it for at least one element? On first glance, `make(map[string]any)` should be sufficient here I think.

pkg/flow/internal/controller/config_export.go
```diff
@@ -0,0 +1,81 @@
...
+
+// Evaluate implements dag.Node and updates the arguments for the managed config block
+// by re-evaluating its River block with the provided scope. The managed config block
+// will be built the first time Evaluate is called.
+//
+// Evaluate will return an error if the River block cannot be evaluated or if
+// decoding to arguments fails.
+func (cn *ExportConfigNode) Evaluate(scope *vm.Scope) error {
+	cn.configNode.mut.Lock()
+	defer cn.configNode.mut.Unlock()
```
rfratto, line 57: IMO it's usually a code smell for a type to use the mutex of another type, since it tends to make it harder to know if protected fields are being accessed properly.

pkg/flow/internal/controller/loader.go
```diff
@@ -71,45 +71,21 @@ func NewLoader(globals ComponentGlobals) *Loader {
...
-
-	wireDiags := l.wireGraphEdges(parentScope, &newGraph)
-	diags = append(diags, wireDiags...)
-
-	// Validate graph to detect cycles
-	err := dag.Validate(&newGraph)
-	if err != nil {
-		diags = append(diags, multierrToDiags(err)...)
+	newGraph, diags := l.loadNewGraph(parentScope, componentNodeBlocks, configBlocks)
+	if diags.HasErrors() {
```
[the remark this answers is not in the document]
rfratto: Hmm, yeah, I'm struggling to tell if this matters. Ideally we collect as many diagnostics as possible so people don't have to keep reloading a bunch of times to see another set of errors, but it also might be a good idea to do an initial return if building the DAG or validating it had any issues, since it might compound unrelated errors when the graph is being walked.

Maybe this is fine as-is for now, and we can revisit if we get negative feedback?

pkg/flow/internal/controller/loader.go
```diff
@@ -71,45 +71,21 @@ func NewLoader(globals ComponentGlobals) *Loader {
 // The provided parentContext can be used to provide global variables and
 // functions to components. A child context will be constructed from the parent
 // to expose values of other components.
-func (l *Loader) Apply(parentScope *vm.Scope, blocks []*ast.BlockStmt, configBlocks []*ast.BlockStmt) diag.Diagnostics {
+func (l *Loader) Apply(parentScope *vm.Scope, componentNodeBlocks []*ast.BlockStmt, configBlocks []*ast.BlockStmt) diag.Diagnostics {
```
rfratto, line 74: Naming nit to make this shorter: `componentBlocks`?

pkg/flow/internal/controller/loader.go
```diff
@@ -144,27 +120,26 @@ func (l *Loader) Apply(parentScope *vm.Scope, blocks []*ast.BlockStmt, configBlo
...
 				if errors.As(err, &evalDiags) {
 					diags = append(diags, evalDiags...)
 				} else {
 					diags.Add(diag.Diagnostic{
 						Severity: diag.SeverityLevelError,
 						Message:  fmt.Sprintf("Failed to build component: %s", err),
-						StartPos: ast.StartPos(n.(*ComponentNode).block).Position(),
-						EndPos:   ast.EndPos(n.(*ComponentNode).block).Position(),
+						StartPos: ast.StartPos(n.(*ComponentNode).Block()).Position(),
+						EndPos:   ast.EndPos(n.(*ComponentNode).Block()).Position(),
```
rfratto, line 132: Looks like you can simplify this even further, since `c` is in scope as a `*ComponentNode`, and is an alias to `n`:

```suggestion
						StartPos: ast.StartPos(c.Block()).Position(),
						EndPos:   ast.EndPos(c.Block()).Position(),
```

rfratto (commented):
Did a first pass, this is coming along well!

pkg/flow/internal/controller/config_export.go
```diff
@@ -0,0 +1,81 @@
...
+	"fmt"
+
+	"github.com/grafana/agent/pkg/flow/internal/dag"
+	"github.com/grafana/agent/pkg/river/ast"
+	"github.com/grafana/agent/pkg/river/diag"
+	"github.com/grafana/agent/pkg/river/vm"
+)
+
+type ExportConfigNode struct {
+	configNode ConfigNode
```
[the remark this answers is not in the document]
erikbaranowski: It's useful from the sense of not copy pasting all the properties that are common. I also figured if we then have any specific properties for a specific config block type that we want to track it will be more obvious this way.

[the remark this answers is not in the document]
rfratto: Hm. I'd usually be in favor of being more explicit for clarity, even if it means copying a little extra code. I'll leave the final call up to you though, I don't think I'm that opinionated about this right now.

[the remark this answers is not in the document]
erikbaranowski: I will do a first pass refactoring for not accessing the mutex from the specific config node types then let's see what we think. I think that will help clarify.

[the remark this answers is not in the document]
erikbaranowski: I made some significant changes once I considered your comments and started cleaning up the code. The commit should document them and I'm thinking `ComponentNode` should use `sharedBlockNode` in place of the 6 properties it has in common. I didn't take this step yet without giving you a chance to consider but I can do that in addition if we think.

pkg/flow/internal/controller/config_export.go
```diff
@@ -0,0 +1,82 @@
...
+			StartPos: ast.StartPos(block).Position(),
+			EndPos:   ast.EndPos(block).Position(),
+		})
+	}
+
+	return &ExportConfigNode{
+		configNode: sharedBlockNode{
+			label:         block.Label,
+			nodeID:        BlockComponentID(block).String(),
+			componentName: block.GetBlockName(),
```
rfratto, line 36: `componentName` isn't right here though, since config blocks aren't components. 

I'm starting to lean a little more firmly against the idea of the sharedBlockNode struct and prefer copying the fields, but I'd like a second opinion from @tpaschalis to see if I'm overthinking it.

pkg/flow/internal/controller/component.go
```diff
@@ -228,9 +227,9 @@ func (cn *ComponentNode) UpdateBlock(b *ast.BlockStmt) {
 	cn.eval = vm.New(b.Body)
 }
 
-// Evaluate updates the arguments for the managed component by re-evaluating
-// its River block with the provided scope. The managed component will be built
-// the first time Evaluate is called.
+// Evaluate implements dag.Node and updates the arguments for the managed component
```
rfratto, line 230:
```suggestion
// Evaluate implements BlockNode and updates the arguments for the managed component
```

pkg/flow/internal/controller/component.go
```diff
@@ -335,6 +334,13 @@ func (cn *ComponentNode) Arguments() component.Arguments {
 	return cn.args
 }
 
+// Block implements dag.Node and returns the current block of the managed component.
```
rfratto, line 337:
```suggestion
// Block implements BlockNode and returns the current block of the managed component.
```

pkg/flow/internal/controller/config_export.go
```diff
@@ -0,0 +1,82 @@
...
+func NewExportConfigNode(block *ast.BlockStmt, globals ComponentGlobals, isInModule bool) (*ExportConfigNode, diag.Diagnostics) {
+	var diags diag.Diagnostics
+
+	if !isInModule {
+		diags.Add(diag.Diagnostic{
+			Severity: diag.SeverityLevelError,
+			Message:  "export blocks only allowed inside a module",
+			StartPos: ast.StartPos(block).Position(),
+			EndPos:   ast.EndPos(block).Position(),
+		})
```
rfratto, line 29: Should this be returning so the node doesn't get created anyway?

pkg/flow/internal/controller/config_logging.go
```diff
@@ -0,0 +1,73 @@
...
+func NewLoggingConfigNode(block *ast.BlockStmt, globals ComponentGlobals, isInModule bool) (*LoggingConfigNode, diag.Diagnostics) {
+	var diags diag.Diagnostics
+
+	if isInModule {
+		diags.Add(diag.Diagnostic{
+			Severity: diag.SeverityLevelError,
+			Message:  "logging block not allowed inside a module",
+			StartPos: ast.StartPos(block).Position(),
+			EndPos:   ast.EndPos(block).Position(),
+		})
```
rfratto, line 28: Should this be returning so the node doesn't get created anyway?

pkg/flow/internal/controller/config_tracing.go
```diff
@@ -0,0 +1,77 @@
...
+func NewTracingConfigNode(block *ast.BlockStmt, globals ComponentGlobals, isInModule bool) (*TracingConfigNode, diag.Diagnostics) {
+	var diags diag.Diagnostics
+
+	if isInModule {
+		diags.Add(diag.Diagnostic{
+			Severity: diag.SeverityLevelError,
+			Message:  "tracing block not allowed inside a module",
+			StartPos: ast.StartPos(block).Position(),
+			EndPos:   ast.EndPos(block).Position(),
+		})
```
rfratto, line 29: Should this be returning so the node doesn't get created anyway?

pkg/flow/internal/controller/loader.go
```diff
@@ -144,27 +119,26 @@ func (l *Loader) Apply(parentScope *vm.Scope, blocks []*ast.BlockStmt, configBlo
...
+		case BlockNode:
+			if err = l.evaluateConfigBlock(logger, parentScope, c); err != nil {
 				diags.Add(diag.Diagnostic{
 					Severity: diag.SeverityLevelError,
-					Message:  fmt.Sprintf("Failed to evaluate node for config blocks: %s", err),
-					StartPos: ast.StartPos(errBlock).Position(),
-					EndPos:   ast.EndPos(errBlock).Position(),
+					Message:  fmt.Sprintf("Failed to evaluate node for config block: %s", err),
+					StartPos: ast.StartPos(n.(BlockNode).Block()).Position(),
+					EndPos:   ast.EndPos(n.(BlockNode).Block()).Position(),
```
rfratto, line 141: Similar change here to the above:

```suggestion
					StartPos: ast.StartPos(c.Block()).Position(),
					EndPos:   ast.EndPos(c.Block()).Position(),
```

pkg/flow/internal/controller/loader.go
```diff
@@ -391,14 +425,14 @@ func (l *Loader) evaluate(logger log.Logger, parent *vm.Scope, c *ComponentNode)
...
-	errBlock, err := c.Evaluate(ectx)
+	err := c.Evaluate(ectx)
 	if err != nil {
 		level.Error(logger).Log("msg", "failed to evaluate config", "node", c.NodeID(), "err", err)
-		return errBlock, err
+		return err
 	}
-	return nil, nil
+	return nil
 }
```
rfratto, line 436: IMO it's a bit awkward that the function is called `evaluateConfigBlock` but it takes any `BlockNode`. Moving the calls to CacheArguments/CacheExports in evaluate out to a new function and refactoring a bit so we don't need two separate `evaluate*` methods?

pkg/flow/internal/dag/tarjan_test.go
```diff
@@ -88,3 +91,7 @@ func sortSlice(nodeSets [][]Node) [][]Node {
 type stringNode string
 
 func (s stringNode) NodeID() string { return string(s) }
+
+func (s stringNode) Block() *ast.BlockStmt { return nil }
+
+func (s stringNode) Evaluate(scope *vm.Scope) error { return nil }
```
rfratto, line 97: Can you revert this since it's no longer required?

pkg/flow/internal/controller/config_export.go
erikbaranowski: I guess it would just become `name` in that case. That being said, if it's a little too object oriented of an approach for normal golang patterns and not a pattern we typically use I'm happy to nuke it once @tpaschalis takes a quick look.

pkg/flow/internal/controller/config_export.go
erikbaranowski: good catch on these, will do

pkg/flow/internal/controller/config_export.go
erikbaranowski: This should be all set.

pkg/flow/internal/controller/loader.go
erikbaranowski: I agree in principle and will need slightly longer to digest and come up with an implementation for this 1 (after the community call probably)

pkg/flow/internal/controller/config_export.go
tpaschalis: Hey team, I'll look at it tomorrow if that's okay.

pkg/flow/internal/controller/loader.go
erikbaranowski: this should be improved

pkg/flow/internal/controller/loader.go
```diff
@@ -373,32 +405,25 @@ func (l *Loader) EvaluateDependencies(parentScope *vm.Scope, c *ComponentNode) {
...
-	// change when a component gets re-evaluated. We also want to cache the arguments and exports in case of an error
-	l.cache.CacheArguments(c.ID(), c.Arguments())
-	l.cache.CacheExports(c.ID(), c.Exports())
-	if err != nil {
-		level.Error(logger).Log("msg", "failed to evaluate component", "component", c.NodeID(), "err", err)
-		return err
+	err := bn.Evaluate(ectx)
+
+	switch c := bn.(type) {
+	case *ComponentNode:
```
rfratto, line 415: Nice, good way of handling this 👍

rfratto (approved):

pkg/flow/internal/controller/loader.go
```diff
@@ -373,32 +405,25 @@ func (l *Loader) EvaluateDependencies(parentScope *vm.Scope, c *ComponentNode) {
 	})
 }
 
-// evaluate constructs the final context for c and evaluates it. mut must be
-// held when calling evaluate.
-func (l *Loader) evaluate(logger log.Logger, parent *vm.Scope, c *ComponentNode) error {
+// evaluate constructs the final context for the special config Node and
+// evaluates it. mut must be held when calling evaluate.
```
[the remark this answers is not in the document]
rfratto: Oops, I merged this before this was addressed. Moving the feedback to #3335.

PR #3299, opened 2023-03-16
loki.process: fix deadlock on components with no stages

rfratto:
This fixes a deadlock when loki.process has no stages defined, since previously it needed at least one stage to create the input pipelines.


slim-bean (approved):
smooth

tpaschalis:
Whoops, thanks for the catch!

PR #3302, opened 2023-03-18
module.git: git-based module loader

### LINKED ISSUE #3454: Flow Modules: Add `module.git` component
a prototype exists here: https://github.com/grafana/agent/pull/3302

Ideally, we will use the common module.go code now that it exists before merging.

rfratto:
This PR introduces a `module.git` module loader. `module.git` clones and manages a Git repository and loads a modules from a file within the managed repository.

The revision of the repository to retrieve can be any recognized revision; including branch names, tag names, or individual commit SHAs.

The repository reference can be anything recognized by `git clone`, including the local filesystem.

The repository will be fetched with the given pull_frequency argument. If pull_frequency is 0, then the repository will only be re-fetched when the component is updated. Setting this argument is primarily useful when pointing `module.git` at a branch.

I have published a repository at https://github.com/rfratto/agent-modules to be used with this PR for examples.

This configuration file can load one of the example modules from that repository:

    logging {
            level = "debug"
    }

    module.git "add" {
            repository = "https://github.com/rfratto/agent-modules.git"
            revision   = "main"
            path       = "add/module.river"

            arguments {
                    a = 15
                    b = 45
            }
    }

Closes #3454.


component/module/git/git.go
```diff
@@ -0,0 +1,302 @@
...
+				} else {
+					if ticker != nil {
+						ticker.Stop()
+					}
+					ticker = nil
+					tickerC = nil
+				}
+
+			}
+			c.mut.Unlock()
```
rfratto, line 165: This code is pretty gross. I'd prefer cleaning this up if we want to move forward with the module loader.

ptodev:
I personally like the idea of using Git as a source, for two reasons:

- A lot of module code may be stored in Git. Being able to retreive it from there  easily is handy.
- It could be our recommended solution for use cases where module versioning is required. Then we could achieve this without building versioning into the modules functionality itself. 

There are just a few edge cases that I am wondering about:

- If the local clone of the repo gets modified (e.g. someone opens the folder and changes some files) then would the Agent recover gracefully? Should we maybe do a `git reset --hard` form time to time? Should we always use everything from `origin` to avoid local branches with the same names? Or in case of failure, maybe it's easiest to delete the directory?
- Do we want this to work with submodules?
- Will this work for private repositories?
- We'd need to document any dependencies on the global git configuration.

rfratto:
> Should we maybe do a git reset --hard form time to time?

The PR currently does a `git reset --hard` when updating the repository to the new SHA (after performing the fetch). 

> Should we always use everything from origin to avoid local branches with the same names?

The fetch is done from remote branches. But also, in general, if a user is going to mess with the data directory managed by the agent, any side effects/bugs it causes is totally on them, and I don't think we should go out of our way to try to support graceful recovery in that scenario. 

> Or in case of failure, maybe it's easiest to delete the directory?
 
That would have the unwanted side effect of preventing the module from working after a reboot; keeping the last valid state should be a priority IMO, even if it's been tampered with by a user and we're unable to update the repository. 

> Do we want this to work with submodules?

It already should, we initialize the submodules when the repo is cloned. 

> Will this work for private repositories?

I haven't tested that, but I believe if you're able to `git clone` on your local machine, this will work for you. I'd say explicit support for private repositories are probably outside of the initial scope.

> We'd need to document any dependencies on the global git configuration.

Good point, but AFAIK there are no dependencies on the global git configuration, at least for accessing public repositories. We can document that it uses the global Git config of the user running Grafana Agent, if that config exists.

ptodev:
Adding to Erik's example above, it'd get even more interesting if any of the modules in the Git repo use modules form other Git repos. And if there are circular dependancies, things could get interesting.

rfratto:
cc @mattdurham heads up that the prototype is here, and is nearly ready minus documentation

erikbaranowski:
@rfratto  @mattdurham what do we think about having a `remote.git` component based on this then embedding it in a `module.git` similar to how `module.file` works? That gives us a new component plus seperates the module loader logic from the git logic. If we think `remote.git` doesn't have any uses or is unsafe then let's not do that.

rfratto:
> @rfratto @mattdurham what do we think about having a remote.git component based on this then embedding it in a module.git similar to how module.file works? That gives us a new component plus seperates the module loader logic from the git logic. If we think remote.git doesn't have any uses or is unsafe then let's not do that.

I didn't want to have a `remote.git` component yet because it's not clear if we'll eventually allow `module.git` to behave differently if given a folder, which wouldn't be possible with a `remote.git` component. 

If we want a `remote.git` component in the future and we don't plan to have special behavior for folders, I think we can refactor this down the road.

component/module/git/git.go
```diff
@@ -0,0 +1,268 @@
...
+
+	c.mut.Lock()
+	defer c.mut.Unlock()
+
+	newArgs := args.(Arguments)
+
+	// TODO(rfratto): store in a repo-specific directory so changing repositories
+	// doesn't risk break the module loader if there's a SHA collision between
+	// the two different repositories.
+	repoPath := filepath.Join(c.opts.DataPath, "repo")
```
rfratto, line 186: I could try to fix this before merging, but honestly I'm not sure if it's necessary. I'm tempted to leave this as-is and see if it ever becomes a problem for anyone.

erikbaranowski:
ya outside even the context of git I want to consider in the future what it would look like to accept a list of module content for a module loader. There may be some way to do it sanely at that level then if a component can pass back that list things 'just work'.

rfratto:
> Instead of persisting to disk thoughts on doing it all in memory or only grabbing a specific file? Ala https://github.com/go-git/go-git#in-memory-example

My thought was persistent to disk allows the module loader to be resilient to network outages across reboots, similar to the problem that the agent management team has been tackling.

mattdurham:
That's reasonable, I am good with this approach as long as we are not guaranteeing the internal behavior of grabbing the whole repo. IE I think we can refine it to only get the file you need and persist to disk and ignore the others.

docs/sources/flow/reference/components/module.git.md
```diff
@@ -0,0 +1,129 @@
...
+`repository` | `string` | The Git repository address to retrieve the module from. | | yes
+`revision` | `string` | The Git revision to retrieve the module from. | | yes
+`path` | `string` | The path in the repository where the module is stored. | | yes
+`pull_frequency` | `duration` | The frequency to pull the repository for updates. | | `"60s"`
+
+The `repository` attribute must be set to a repository address that would be
+recognized by Git with a `git clone REPOSITORY_ADDRESS` command, such as
+`htts://github.com/grafana/agent.git`.
+
+The `revision` attribute must be set to a valid branch, tag, or commit SHA
```
mattdurham, line 49: IMO revision should be optional, assume HEAD if none given.

rfratto:
> That's reasonable, I am good with this approach as long as we are not guaranteeing the internal behavior of grabbing the whole repo. IE I think we can refine it to only get the file you need and persist to disk and ignore the others.

I couldn't find a way to do that, especially with subsequent updates. But in general I think the storage used by components is never guaranteed to be consistent and isn't a part of their public interface.

component/module/git/git.go
```diff
@@ -0,0 +1,269 @@
...
+
+	mut      sync.RWMutex
+	repo     *vcs.GitRepo
+	repoOpts vcs.GitRepoOptions
+	args     Arguments
+
+	argsChanged chan struct{}
+
+	healthMut sync.RWMutex
+	health    component.Health
```
erikbaranowski, line 74: Do we need to track a second health here or can we use the health on module.ModuleComponent? If that implementation is unclear maybe we update it but the intention is for module.ModuleComponent.health to manage the module health.

related, updateHealth looks like it could be common enough we should add it to module.go instead of in here.

rfratto: We need a second health for the health of the git cloning code, yes. Actually, let me update this to use the new `component.LeastHealthy` functionality

erikbaranowski: sounds good, this will make more sense to track separate in that context I think.

erikbaranowski (approved):
LGTM! I tested it against a version of my own config from a feature branch with an argument + export and everything is looking great.

You could update the health stuff as discussed or leave for another day from my perspective.

jkroepke:
I would like to know, if private git repostories are supported, too? For SSH access, the `ssh-agent` may take care. But what about Auth Basic (e.g. PAT, Tokens?)

If `https://user:password@github.com/` is the preffered way for passing credentials, can the URL marked as secret?

mattdurham:
We will add support for private git repos, but we are likely to check this in to get feedback and then circle back around to private repos later.

PR #3312, opened 2023-03-21
Make the Flow controller cluster-aware and use it to distribute targets between prometheus.scrape components

tpaschalis:
This is a PR to showcase a first use of clustering in the Agent.

For easier review, I suggest going commit-by-commit.

I'd like to throw out the idea that we may even split this up into multiple PRs if we at least agree in the way that we make the controller cluster-aware if we don't agree on the way we implement component behavior.

More specifically, this PR:

* Introduces a couple of new command-line arguments that enable clustered mode and specifying which peers to connect to.
* Introduces a `Clusterer` struct which is a thin layer over rfratto/ckit
* Propagates the Clusterer to be used by components through their Arguments
* Allows components to register and unregister themselves to have their Update method called 
* Uses the above to have the `prometheus.scrape` component wrap its input targets in a new struct and distribute between multiple components.

To test this, one can work with a config like the following
```river
prometheus.scrape "default" {
	clustering {
                enabled = true
        }

	targets = [
		{"__address__" = "http://demo.robustperception.io:9090/metrics", "color" = "pink"},
		{"__address__" = "http://demo.robustperception.io:9090/metrics", "color" = "yellow"},
		{"__address__" = "http://demo.robustperception.io:9090/metrics", "color" = "red"},
		{"__address__" = "http://demo.robustperception.io:9090/metrics", "color" = "blue"},
		{"__address__" = "http://demo.robustperception.io:9090/metrics", "color" = "white"},
	]
	forward_to = []
}
```

Then, start a bootstrap node on port 12345
```
$ AGENT_MODE=flow ./build/grafana-agent run ~/clustering_targets.river --server.http.listen-addr 127.0.0.1:12345 --cluster.enabled --cluster.join-address 127.0.0.1
```

and as many additional nodes as you'd like
```
$ AGENT_MODE=flow ./build/grafana-agent run ~/clustering_targets.river --server.http.listen-addr 127.0.0.1:12346 --cluster.enabled --cluster.join-address 127.0.0.1:12345
$ AGENT_MODE=flow ./build/grafana-agent run ~/clustering_targets.river --server.http.listen-addr 127.0.0.1:12347 --cluster.enabled --cluster.join-address 127.0.0.1:12345
$ AGENT_MODE=flow ./build/grafana-agent run ~/clustering_targets.river --server.http.listen-addr 127.0.0.1:12348 --cluster.enabled --cluster.join-address 127.0.0.1:12345
```

When you navigate to the UI of each of your nodes, you'll see the DebugInfo only listing the targets the Agent is responsible for.
```
http://localhost:12345/component/prometheus.scrape.default#Debug%20info-target_3
http://localhost:12346/component/prometheus.scrape.default#Debug%20info-target_3
```

### DIFF
--- a/cmd/internal/flowmode/cmd_run.go
+++ b/cmd/internal/flowmode/cmd_run.go
@@ -16,10 +16,13 @@ import (
 	"github.com/grafana/agent/web/ui"
 	"go.opentelemetry.io/otel"
 	"golang.org/x/exp/maps"
+	"golang.org/x/net/http2"
+	"golang.org/x/net/http2/h2c"
 
 	"github.com/fatih/color"
 	"github.com/go-kit/log/level"
 	"github.com/gorilla/mux"
+	"github.com/grafana/agent/pkg/cluster"
 	"github.com/grafana/agent/pkg/config/instrumentation"
 	"github.com/grafana/agent/pkg/flow"
 	"github.com/grafana/agent/pkg/flow/logging"
@@ -78,9 +81,13 @@ depending on the nature of the reload error.
 	}
 
 	cmd.Flags().
-		StringVar(&r.httpListenAddr, "server.http.listen-addr", r.httpListenAddr, "address to listen for HTTP traffic on")
+		StringVar(&r.httpListenAddr, "server.http.listen-addr", r.httpListenAddr, "Address to listen for HTTP traffic on")
 	cmd.Flags().StringVar(&r.storagePath, "storage.path", r.storagePath, "Base directory where components can store data")
 	cmd.Flags().StringVar(&r.uiPrefix, "server.http.ui-path-prefix", r.uiPrefix, "Prefix to serve the HTTP UI at")
+	cmd.Flags().
+		BoolVar(&r.clusterEnabled, "cluster.enabled", r.clusterEnabled, "Start in clustered mode")
+	cmd.Flags().
+		StringVar(&r.clusterJoinAddr, "cluster.join-address", r.clusterJoinAddr, "Address to join the cluster at")
 	cmd.Flags().
 		BoolVar(&r.disableReporting, "disable-reporting", r.disableReporting, "Disable reporting of enabled components to Grafana.")
 	return cmd
@@ -91,6 +98,8 @@ type flowRun struct {
 	storagePath      string
 	uiPrefix         string
 	disableReporting bool
+	clusterEnabled   bool
+	clusterJoinAddr  string
 }
 
 func (fr *flowRun) Run(configFile string) error {
@@ -138,9 +147,15 @@ func (fr *flowRun) Run(configFile string) error {
 	reg := prometheus.DefaultRegisterer
 	reg.MustRegister(newResourcesCollector(l))
 
+	clusterer, err := cluster.New(l, fr.clusterEnabled, fr.httpListenAddr, fr.clusterJoinAddr)
+	if err != nil {
+		return fmt.Errorf("building clusterer: %w", err)
+	}
+
 	f := flow.New(flow.Options{
 		LogSink:        logSink,
 		Tracer:         t,
+		Clusterer:      clusterer,
 		DataPath:       fr.storagePath,
 		Reg:            reg,
 		HTTPPathPrefix: "/api/v0/component/",
@@ -186,6 +201,7 @@ func (fr *flowRun) Run(configFile string) error {
 		r.Handle("/metrics", promhttp.Handler())
 		r.PathPrefix("/debug/pprof").Handler(http.DefaultServeMux)
 		r.PathPrefix("/api/v0/component/{id}/").Handler(f.ComponentHandler())
+		r.PathPrefix("/api/v1/ckit/transport").Handler(clusterer.Mux)
 
 		r.HandleFunc("/-/ready", func(w http.ResponseWriter, _ *http.Request) {
 			if f.Ready() {
@@ -217,7 +233,7 @@ func (fr *flowRun) Run(configFile string) error {
 		// will take precedence over anything else mapped in uiPrefix.
 		ui.RegisterRoutes(fr.uiPrefix, r)
 
-		srv := &http.Server{Handler: r}
+		srv := &http.Server{Handler: h2c.NewHandler(r, &http2.Server{})}
 
 		wg.Add(1)
 		go func() {
--- a/cmd/internal/flowmode/cmd_run.go
+++ b/cmd/internal/flowmode/cmd_run.go
@@ -87,6 +87,8 @@ depending on the nature of the reload error.
 	cmd.Flags().
 		BoolVar(&r.clusterEnabled, "cluster.enabled", r.clusterEnabled, "Start in clustered mode")
 	cmd.Flags().
+		StringVar(&r.clusterJoinAddr, "cluster.advertise-address", r.clusterAdvAddr, "Address to advertise to the cluster")
+	cmd.Flags().
 		StringVar(&r.clusterJoinAddr, "cluster.join-address", r.clusterJoinAddr, "Address to join the cluster at")
 	cmd.Flags().
 		BoolVar(&r.disableReporting, "disable-reporting", r.disableReporting, "Disable reporting of enabled components to Grafana.")
@@ -99,6 +101,7 @@ type flowRun struct {
 	uiPrefix         string
 	disableReporting bool
 	clusterEnabled   bool
+	clusterAdvAddr   string
 	clusterJoinAddr  string
 }
 
@@ -147,7 +150,7 @@ func (fr *flowRun) Run(configFile string) error {
 	reg := prometheus.DefaultRegisterer
 	reg.MustRegister(newResourcesCollector(l))
 
-	clusterer, err := cluster.New(l, fr.clusterEnabled, fr.httpListenAddr, fr.clusterJoinAddr)
+	clusterer, err := cluster.New(l, fr.clusterEnabled, fr.clusterAdvAddr, fr.clusterJoinAddr)
 	if err != nil {
 		return fmt.Errorf("building clusterer: %w", err)
 	}
--- a/cmd/internal/flowmode/cmd_run.go
+++ b/cmd/internal/flowmode/cmd_run.go
@@ -201,7 +201,10 @@ func (fr *flowRun) Run(configFile string) error {
 		r.Handle("/metrics", promhttp.Handler())
 		r.PathPrefix("/debug/pprof").Handler(http.DefaultServeMux)
 		r.PathPrefix("/api/v0/component/{id}/").Handler(f.ComponentHandler())
-		r.PathPrefix("/api/v1/ckit/transport").Handler(clusterer.Mux)
+
+		// Register routes for the clusterer.
+		cr, ch := clusterer.Node.Handler()
+		r.PathPrefix(cr).Handler(ch)
 
 		r.HandleFunc("/-/ready", func(w http.ResponseWriter, _ *http.Request) {
 			if f.Ready() {
--- a/cmd/internal/flowmode/cmd_run.go
+++ b/cmd/internal/flowmode/cmd_run.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/grafana/agent/web/api"
 	"github.com/grafana/agent/web/ui"
+	"github.com/rfratto/ckit/peer"
 	"go.opentelemetry.io/otel"
 	"golang.org/x/exp/maps"
 	"golang.org/x/net/http2"
@@ -250,6 +251,21 @@ func (fr *flowRun) Run(configFile string) error {
 		}()
 
 		defer func() { _ = srv.Shutdown(ctx) }()
+	}
+
+	// Start the clusterer after having started the HTTP server.
+	err = clusterer.Node.Start()
+	if err != nil {
+		level.Error(l).Log("msg", "failed to connect to any peers; including ourselves to bootstrap a new cluster", "err", err)
+		return err
+	}
+
+	// Nodes initially join the cluster in the Viewer state. We can move to the
+	// Participant state to signal that we wish to participate in reading or
+	// writing data.
+	err = clusterer.Node.ChangeState(context.Background(), peer.StateParticipant)
+	if err != nil {
+		return err
 	}
 
 	// Report usage of enabled components
--- a/cmd/internal/flowmode/cmd_run.go
+++ b/cmd/internal/flowmode/cmd_run.go
@@ -14,7 +14,6 @@ import (
 
 	"github.com/grafana/agent/web/api"
 	"github.com/grafana/agent/web/ui"
-	"github.com/rfratto/ckit/peer"
 	"go.opentelemetry.io/otel"
 	"golang.org/x/exp/maps"
 	"golang.org/x/net/http2"
@@ -251,21 +250,6 @@ func (fr *flowRun) Run(configFile string) error {
 		}()
 
 		defer func() { _ = srv.Shutdown(ctx) }()
-	}
-
-	// Start the clusterer after having started the HTTP server.
-	err = clusterer.Node.Start()
-	if err != nil {
-		level.Error(l).Log("msg", "failed to connect to any peers; including ourselves to bootstrap a new cluster", "err", err)
-		return err
-	}
-
-	// Nodes initially join the cluster in the Viewer state. We can move to the
-	// Participant state to signal that we wish to participate in reading or
-	// writing data.
-	err = clusterer.Node.ChangeState(context.Background(), peer.StateParticipant)
-	if err != nil {
-		return err
 	}
 
 	// Report usage of enabled components
--- a/component/component.go
+++ b/component/component.go
@@ -126,7 +126,7 @@ type HTTPComponent interface {
 	Handler() http.Handler
 }
 
-// ClusteredComponent is an extension interface for componens which implement
+// ClusteredComponent is an extension interface for components which implement
 // clustering-specific behavior.
 type ClusteredComponent interface {
 	Component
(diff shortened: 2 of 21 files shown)


cmd/grafana-agent/flow_run.go
```diff
@@ -138,9 +145,15 @@ func (fr *flowRun) Run(configFile string) error {
 	reg := prometheus.DefaultRegisterer
 	reg.MustRegister(newResourcesCollector(l))
 
+	clusterer, err := cluster.New(l, fr.clusterEnabled, fr.httpListenAddr, fr.clusterJoinAddr)
+	if err != nil {
+		return err
```
rfratto, line 150: Can you wrap the error here?

component/component.go
```diff
@@ -125,3 +125,10 @@ type HTTPComponent interface {
 	// will receive a request to just `/metrics`.
 	Handler() http.Handler
 }
+
+// ClusterComponent is an extension interface for components which implement
+// new behavior whenever the Flow controller is running in clustered mode.
+type ClusterComponent interface {
+	Component
+	ClusterWatch() bool
+}
```
rfratto, line 134: I'm not sure about this long term, but I'm OK with it for the scope of the initial PR.

I'm still interested some kind of signaling mechanism so components can register/unregister themselves for stuff like cluster events when their Update method is called. 

This could be handled by passing a context.Context with some special values in it, or maybe extra arguments to Update to allow us to give that kind of data to components.

component/discovery/discovery.go
```diff
@@ -2,20 +2,64 @@ package discovery
...
+	"github.com/grafana/agent/pkg/cluster"
 	"github.com/prometheus/prometheus/discovery"
 	"github.com/prometheus/prometheus/discovery/targetgroup"
 	"github.com/prometheus/prometheus/model/labels"
+	"github.com/rfratto/ckit/shard"
 )
 
+// Clustering determines whether to use clustering.
+type Clustering struct {
+	NodeUpdates bool `river:"node_updates,attr"`
```
rfratto, line 20: I think the River attribute name here is a little confusing. `enable_clustering`?

component/discovery/discovery.go
```diff
@@ -2,20 +2,64 @@ package discovery
...
+func (t *DistributedTargets) Get() []Target {
+	if t.node == nil {
+		return t.targets
+	}
+
+	res := make([]Target, 0, len(t.targets)/len(t.node.Peers()))
+	for _, tgt := range t.targets {
+		peers, err := t.node.Lookup(shard.StringKey(tgt.Labels().String()), 1, shard.OpReadWrite)
+		if err != nil {
+			continue
```
rfratto, line 50: If the cluster is a failing state, then no targets will get picked up by any agent. 
Maybe the default behavior should be to assign to self if an error happens here.

component/discovery/discovery.go
```diff
@@ -2,20 +2,64 @@ package discovery
...
+		}
+		if peers[0].Self {
+			res = append(res, tgt)
+		}
+	}
+	fmt.Println("Node is responsible for:")
+	for _, tgt := range res {
+		fmt.Println("  ", tgt)
+	}
```
rfratto, line 59: stray debug code?

component/registry.go
```diff
@@ -61,6 +62,14 @@ type Options struct {
 	// attribute denoting the component ID.
 	Tracer trace.TracerProvider
 
+	// Clusterer allows components to work in a clustered fashion. The
+	// clusterer is shared between all components initialized by a Flow
+	// controller.
+	//
+	// TODO(@tpaschalis) As noted in the ComponentGlobals, we most likely only
+	// need the Sharder implementation here.
+	Clusterer *cluster.Clusterer
```
rfratto, line 71: I think this potentially fits in with finding a better way to send signals to components through the Update method so we don't have to keep adding new fields to the options. 

But: not a blocker for this PR.

pkg/cluster/cluster.go
```diff
@@ -67,3 +74,124 @@ func (ln *localNode) Observe(ckit.Observer) {
 func (ln *localNode) Peers() []peer.Peer {
 	return []peer.Peer{ln.self}
 }
+
+// Clusterer implements the behavior required for operating Flow controllers
```
rfratto, line 78: Can you remove the previous implementation so we don't have both? The old implementation isn't being used anywhere.

pkg/cluster/cluster.go
```diff
@@ -67,3 +74,124 @@ func (ln *localNode) Observe(ckit.Observer) {
...
+			ComponentsToUpdate: make(map[string]struct{}),
+			ClusterChanges:     make(chan struct{}),
+		}, nil
+	}
+
+	// Gossip node.
+	var lis net.Listener
+	var err error
+
+	lis, err = net.Listen("tcp", "127.0.0.1:64583")
```
rfratto, line 106: Can you add a TODO so we don't accidentally commit this?

component/component.go
tpaschalis: I've removed this for now; components call directly the Register method through their Options. I'm not sure if this the way we'll go for, but it's much better than this ugly solution.

component/discovery/discovery.go
tpaschalis: This will error out only if we request more peers than they're available, which will never happen. I've gone ahead and done that as it's sensible in the general case though.

cmd/internal/flowmode/cmd_run.go
```diff
@@ -217,7 +233,7 @@ func (fr *flowRun) Run(configFile string) error {
 		// will take precedence over anything else mapped in uiPrefix.
 		ui.RegisterRoutes(fr.uiPrefix, r)
 
-		srv := &http.Server{Handler: r}
+		srv := &http.Server{Handler: h2c.NewHandler(r, &http2.Server{})}
```
tpaschalis, line 236: Are we okay with enabling h2c for all routes here?

pkg/cluster/cluster.go
```diff
@@ -67,3 +71,111 @@ func (ln *localNode) Observe(ckit.Observer) {
...
+	}
+	port, err := strconv.Atoi(portStr)
+	if err != nil {
+		return nil, err
+	}
+
+	mux := http.NewServeMux()
+
+	gossipConfig := DefaultGossipConfig
+	gossipConfig.NodeName = uuid.NewString()
```
tpaschalis, line 113: Should this be configurable? Should it default to the hostname? Any other ideas?

pkg/cluster/cluster.go
```diff
@@ -67,3 +71,111 @@ func (ln *localNode) Observe(ckit.Observer) {
...
+	// cluster by ourselves.
+	err = gossipNode.Start()
+	if err != nil {
+		level.Debug(log).Log("msg", "failed to connect to peers; bootstrapping a new cluster")
+		gossipConfig.JoinPeers = nil
+		err = gossipNode.Start()
+		if err != nil {
+			return nil, err
+		}
+	}
```
tpaschalis, line 137: This smells a bit hacky. We could delay the Start() call to after we start the HTTP server so we can always connect to ourselves without the need for special handling

pkg/cluster/cluster.go
```diff
@@ -67,3 +71,111 @@ func (ln *localNode) Observe(ckit.Observer) {
...
+
+	res := &Clusterer{
+		Node: gossipNode,
+		Mux:  mux,
+
+		RegisteredComponents: make(map[string]struct{}),
+	}
+
+	gossipNode.Observe(ckit.FuncObserver(func(peers []peer.Peer) (reregister bool) {
```
tpaschalis, line 154: This is useful only during development/testing. Remember to remove this.

pkg/flow/internal/controller/loader.go
```diff
@@ -56,6 +58,19 @@ func NewLoader(globals ComponentGlobals) *Loader {
...
+		for _, cmp := range l.Components() {
+			if globals.Clusterer.IsRegistered(cmp.ID().String()) {
+				err := cmp.Revisit()
+				if err != nil {
+					level.Error(globals.Logger).Log("msg", "failed to revisit component", "componentID", cmp.NodeID(), "err", err)
+				}
+			}
+		}
+		return true
+	}))
```
tpaschalis, line 72: Is this the best place for registering the observer? What other alternatives do we have?

tpaschalis (commented):
First self-review pass, will take a fresh look in a bit.

pkg/flow/internal/controller/loader.go
tpaschalis: Also, this does not easily allow for _other_ types of registrations; here components will only have their Update method called again. Keeping an eye out for other ideas.

cmd/internal/flowmode/cmd_run.go
rfratto: I'd say so 👍

component/prometheus/scrape/scrape.go
```diff
@@ -183,7 +184,11 @@ func (c *Component) Run(ctx context.Context) error {
 				jobName = c.args.JobName
 			}
 			c.mut.RUnlock()
-			promTargets := c.componentTargetsToProm(jobName, tgs)
+
+			// NOTE(@tpaschalis) First approach, manually building the
+			// 'clustered' targets implementation every time.
+			ct := discovery.NewDistributedTargets(c.opts.Clusterer.Node, tgs)
+			promTargets := c.componentTargetsToProm(jobName, ct.Get())
```
rfratto, line 191: Is this correct? From what I can tell, this runs even when `clustering_enabled` is false for the component, right? That would cause targets to be sharded and never update if the agent is participating in a cluster but the component isn't subscribed to updates.

pkg/cluster/cluster.go
rfratto: Yes: configurable but default to the host name. 

The node name should be consistent between process restarts, otherwise you'll pick up an entirely different set of targets when restarting.

component/registry.go
```diff
@@ -61,6 +62,11 @@ type Options struct {
 	// attribute denoting the component ID.
 	Tracer trace.TracerProvider
 
+	// Clusterer allows components to work in a clustered fashion. The
+	// clusterer is shared between all components initialized by a Flow
+	// controller.
+	Clusterer *cluster.Clusterer
```
rfratto, line 68: Can you update the modules code to propagate this into the flow.Options of controllers they create?

pkg/flow/internal/controller/component.go
```diff
@@ -247,6 +250,37 @@ func (cn *ComponentNode) Evaluate(scope *vm.Scope) error {
 	return err
 }
 
+// Revisit calls Update on the managed component with its last used arguments.
+// Revisit does not build the component if it is not already built and does not
+// re-evaluate the River block.
+// Its only use case is for components opting-in to clustering where calling
+// Update with the same Arguments may result in different functionality.
+func (cn *ComponentNode) Revisit() error {
```
rfratto, line 258: nit: name sugestion: `Reevaluate`

pkg/cluster/cluster.go
```diff
@@ -67,3 +71,111 @@ func (ln *localNode) Observe(ckit.Observer) {
...
+		delete(c.RegisteredComponents, nodeID)
+	}
+}
+
+// IsRegistered reports whether a nodeID is scheduled to have its Update method
+// called whenever the cluster state changes.
+func (c *Clusterer) IsRegistered(nodeID string) bool {
+	_, ok := c.RegisteredComponents[nodeID]
+	return ok
+}
```
rfratto, line 181: I'd like for this to not reference components/nodeIDs here. As it is now, this builds an implicit circular dependency with pkg/flow, as pkg/flow directly imports pkg/cluster, and pkg/cluster assumes knowledge about pkg/flow.

pkg/flow/internal/controller/loader.go
```diff
@@ -56,6 +58,19 @@ func NewLoader(globals ComponentGlobals) *Loader {
 	if globals.Registerer != nil {
 		globals.Registerer.MustRegister(cc)
 	}
+
+	globals.Clusterer.Node.Observe(ckit.FuncObserver(func(peers []peer.Peer) (reregister bool) {
+		for _, cmp := range l.Components() {
```
rfratto, line 63: You have a memory leak in your design: if a component registered for cluster updates is removed from the config file, its registration will persist forever. 

I think the marker interface you had before is probably preferable over the explicit registration design because it makes it harder for memory to leak.

component/prometheus/scrape/scrape.go
tpaschalis: You're right, this was erroneous. I've fixed this by removing the Registration logic and depending on the Extension interface for now and it works as expected.

I'm exploring the approach with the Capabilities/Registration approaches to see how they compare.

pkg/cluster/cluster.go
rfratto: We don't have to get the API perfect on the first PR, this is much better from the perspective of it not leaking 👍

rfratto (approved):
LGTM, though I'd prefer not to merge this until:

* rfratto/ckit#6 is merged in main 
* gRPC support is removed from rfratto/ckit 
* We update this PR to use latest rfratto/ckit 

Then, after this PR, we can migrate the ckit code over into the agent.

component/component.go
```diff
@@ -125,3 +125,11 @@ type HTTPComponent interface {
 	// will receive a request to just `/metrics`.
 	Handler() http.Handler
 }
+
+// ClusteredComponent is an extension interface for componens which implement
```
mattdurham, line 129: nit: typo for componen(t)s

component/discovery/discovery.go
```diff
@@ -7,15 +7,57 @@ import (
...
+	"github.com/rfratto/ckit/shard"
 )
 
 // Target refers to a singular discovered endpoint found by a discovery
 // component.
 type Target map[string]string
 
+// DistributedTargets uses the node's Lookup method to distribute discovery
+// targets when a Flow component runs in a cluster.
+type DistributedTargets struct {
```
mattdurham, line 23: Idea only: I wonder if we can make this more generic....with generics? ;) Something like this pseudo code. So that we dont have to write specific ones for each type of object.

```go
// Declaration
Owns[T](key func(v T),values []T) []T

// Implementation
Owns[Target](func(t Target){
    return tgt.Labels().String()
},
targets)
```

pkg/cluster/cluster.go
```diff
@@ -67,3 +75,99 @@ func (ln *localNode) Observe(ckit.Observer) {
...
+
+	cli := &http.Client{
+		Transport: &http2.Transport{
+			AllowHTTP: true,
+			DialTLS: func(network, addr string, _ *tls.Config) (net.Conn, error) {
+				return net.Dial(network, addr)
+			},
+			TLSClientConfig: &tls.Config{
+				InsecureSkipVerify: true,
+			},
```
mattdurham, line 131: This `InsecureSkipVerify` feels a bit suspect. TODO?

component/discovery/discovery.go
```diff
@@ -7,15 +7,57 @@ import (
...
+	return DistributedTargets{e, n, t}
+}
+
+// Get distributes discovery targets a clustered environment.
+//
+// If a cluster size is 1, then all targets will be returned.
+func (t *DistributedTargets) Get() []Target {
+	if !t.useClustering || t.node == nil {
+		return t.targets
+	}
```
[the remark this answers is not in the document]
tpaschalis: Ok, I'm adding a TODO here for now.

I think the localNode implementation that we already have in place will suit the 'cluster of one' concept.

pkg/cluster/cluster.go
tpaschalis: Hmm, commenting it out doesn't make any difference on the ckit tests, they're still passing with h2c. I'm taking a look as to what this _actually_ does and if we need it.

pkg/flow/internal/controller/loader.go
```diff
@@ -61,10 +61,12 @@ func NewLoader(globals ComponentGlobals) *Loader {
 
 	globals.Clusterer.Node.Observe(ckit.FuncObserver(func(peers []peer.Peer) (reregister bool) {
```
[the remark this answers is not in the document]
tpaschalis: Something like e6fc78dd8ef1707c602e572b413cc5644411a533?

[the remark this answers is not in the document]
rfratto: Yes, but then a child span for each reevaluated component :)

pkg/cluster/cluster.go
tpaschalis: And when _using_ TLS, the certificates should be able to match correctly, right? So I think we might be able to remove this

rfratto: > And when using TLS, the certificates should be able to match correctly, right? So I think we might be able to remove this

Yeah, we should never set this by default when using TLS unless the user specifies it.

component/prometheus/scrape/scrape.go
```diff
@@ -81,6 +80,8 @@ type Arguments struct {
 
 	// Scrape Options
 	ExtraMetrics bool `river:"extra_metrics,attr,optional"`
+
+	ClusteringEnabled bool `river:"clustering_enabled,attr,optional"`
```
rfratto, line 84: I would like to see this as a block rather than an attribute, even if it's just a block with a single attribute in it:

```river 
clustering {
  enabled = true 
}
```

The main reason I want to do this is because we'll probably document a lot of information relevant to clustering, and it would be nice to document that in its own dedicated section for the `clustering` block, whereas right now it would have to be all jammed into the arguments block documentation.

pkg/cluster/cluster.go
tpaschalis: I'm leaving this in for the time being. It's useful as an immediate signal to peers joining-leaving and isn't too noisy, but will eventually get removed.

Issue #3304, opened 2023-03-19
Non-obvious Flow map(string) syntax when special characters are used in a label.


jacobstr:
Looking to configure a prometheus rule that matches rules with a label that may contain `/` or `.` - both are legal characters in a typical [kubernetes label](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set).

E.g. The agent itself using chart defaults has this k8s label:

```
labels:
    app.kubernetes.io/instance: grafana-agent
```

The docs indicated `map(string)` as the type when using `match_labels` in the mimir rule forwarding component. I've tried the following:

YOLO:

```
mimir.rules.kubernetes "default" {
        match_labels {
          grafana.example.com/forward = "true"
        }
}
```

Double quote the key? This is my principle-of-least-surprise favorite given experience in jsonnet and yaml.
```
mimir.rules.kubernetes "default" {
        match_labels {
          "grafana.example.com/forward" = "true"
        }
}
```

Single-quotes?

```
mimir.rules.kubernetes "default" {
        match_labels {
          'grafana.example.com/forward' = "true"
        }
}
```

Escape a special character?

```
mimir.rules.kubernetes "default" {
        match_labels {
          grafana\/forward = "true"
        }
}
```

---

Random thoughts that I hope convey my confusion:

* Vaguely, I'd expect something in the [documentation here](https://grafana.com/docs/agent/latest/flow/config-language/expressions/types_and_values/#objects) specific to map types. It's really left up to guesswork like I did above to determine what the syntax is - if any.
* If the language can't support this - well that's a bit of a conundrum because kubernetes labels _do_ allow characters that might not typically comprise a valid `identifier` or `block label`. 
* Examples in [loki](https://grafana.com/docs/agent/latest/flow/reference/components/loki.source.podlogs/#usage) and [mimir](https://grafana.com/docs/agent/latest/flow/reference/components/mimir.rules.kubernetes/) would benefit from using non-trivial examples for these label selectors.
* The "BNF"-looking spec for the language suggests https://grafana.com/docs/agent/latest/flow/config-language/syntax/#identifiers  that `grafana.example.com/forward` is an identifier and claims that any UTF-8 ~character~ letter is valid.
  * Which would be somewhat difficult since the identifier obviously can't have spaces or an equal sign if it wants to get lexed.
[...]

PR #3316, opened 2023-03-21
docs: clarify objects, fmt based on feedback in #3304

rfratto:
Clarify documentation around object construction and the `fmt` command based on feedback in provided in #3304.

### DIFF
--- a/docs/sources/flow/config-language/expressions/types_and_values.md
+++ b/docs/sources/flow/config-language/expressions/types_and_values.md
@@ -123,7 +123,7 @@ a string:
 > An _object_ is a value assigned to an [Attribute][Attributes], where
 > commas **must** be provided between key-value pairs on separate lines.
 >
-> A [Block][Blocks] is a structural element composed of multiple attributes,
+> A [Block][Blocks] is a named structural element composed of multiple attributes,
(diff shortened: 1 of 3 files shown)


docs/sources/flow/config-language/expressions/types_and_values.md
```diff
@@ -82,25 +89,53 @@ is on a different line as the final value:
...
+  namespace                    = "default",
+}
+```
+
+> **NOTE**: Be careful not to confuse objects with blocks.
+>
+> An _object_ is a value assigned to an [Attribute][Attributes], where
+> commas **must** be provided between key-value pairs on separate lines.
+>
+> A [Block][Blocks] is a structural element composed of multiple attributes,
```
tpaschalis, line 126:
```suggestion
> A [Block][Blocks] is a named structural element composed of multiple attributes,
```

nit: does this make the difference any clearer? Not sure if it's worth it.

tpaschalis (approved):

docs/sources/flow/config-language/expressions/types_and_values.md
rfratto: sure, I like that

PR #3314, opened 2023-03-21
docs: create a Getting started folder for Flow

rfratto:
This commit creates a _Getting started_ folder in the Flow documentation. _Getting started_ will be used for various task-based documentation.

The existing _Getting started_ guide has been moved to a "Run Grafana Agent Flow" topic within that folder, and updated to follow the recommendations for [Task topics](https://grafana.com/docs/writers-toolkit/writing-guide/topic-types/task/).

PR #3320, opened 2023-03-21
docs: add Collect Prometheus metrics topic for Flow

rfratto:
Add a _Collect Prometheus metrics_ topic in the Grafana Agent Flow documentation. This topic is intended to be used for all common tasks around Prometheus metric collection.

For this initial commit, the only task is for configuring a prometheus.remote_write component. More tasks will be added in follow-up PRs to keep the review size as small as possible.

I've started with the task for configuring metrics delivery _first_ to allow each task section to be functional in isolation. Without being instructed to configure metrics delivery, there would be no point in collecting metrics. I plan to open a second PR for a metrics collection task ASAP after this is merged to reduce confusion for readers that stumble on it while there's only one task. 

Like #3314, this follows the recommendation for [Task topics](https://grafana.com/docs/writers-toolkit/writing-guide/topic-types/task/).

### DIFF
--- a/docs/sources/flow/getting-started/collect-prometheus-metrics.md
+++ b/docs/sources/flow/getting-started/collect-prometheus-metrics.md
@@ -24,12 +24,13 @@ This topic describes how to:
 
 ## Before you begin
 
-* Ensure that you basic familiarity with instrumenting applications with
+* Ensure that you have basic familiarity with instrumenting applications with
   Prometheus.
 * Have a set of Prometheus exports or applications exposing Prometheus metrics
   that you want to collect metrics from.
-* Identify where you will write collected metrics, such as Prometheus, Grafana
-  Mimir, Grafana Cloud, or Grafana Enterprise Metrics.
+* Identify where you will write collected metrics. Metrics may be written to
+  Prometheus or Prometheus-compatible endpoints such as Grafana Mimir, Grafana
+  Cloud, or Grafana Enterprise Metrics.
 * Be familiar with the concept of [Components][] in Grafana Agent Flow.
 
 [Components]: {{< relref "../concepts/components.md" >}}
--- a/docs/sources/flow/getting-started/collect-prometheus-metrics.md
+++ b/docs/sources/flow/getting-started/collect-prometheus-metrics.md
@@ -36,6 +36,9 @@ This topic describes how to:
 
 ## Configure metrics delivery
 
+Before components can collect Prometheus metrics, you must have a component
+responsible for writing those metrics somewhere.
+
 The [prometheus.remote_write][] component is responsible for delivering
 Prometheus metrics to one or Prometheus-compatible endpoints. Once a
 `prometheus.remote_write` component is defined, other Grafana Agent Flow
--- a/docs/sources/flow/getting-started/collect-prometheus-metrics.md
+++ b/docs/sources/flow/getting-started/collect-prometheus-metrics.md
@@ -76,7 +76,7 @@ complete the following steps:
     2. Replace `USERNAME` with the basic authentication username to use.
 
     3. Replace `PASSWORD_FILE` with a path to a file containing the basic
-       authentication password.
+       authentication password or API key.
 
 5. If you have more than one endpoint to write metrics to, repeat the
    `endpoint` block for additional endpoints.
--- a/docs/sources/flow/getting-started/collect-prometheus-metrics.md
+++ b/docs/sources/flow/getting-started/collect-prometheus-metrics.md
@@ -1,0 +1,114 @@
+---
+title: Collect Prometheus metrics
+weight: 200
+---
+
+# Collect Prometheus metrics
+
+Grafana Agent Flow can be configured to collect [Prometheus][] metrics and
+forward them to any Prometheus-compatible database.
+
+This topic describes how to:
+
+* Configure metrics delivery
+
+[Prometheus]: https://prometheus.io
+
+## Components used in this topic
+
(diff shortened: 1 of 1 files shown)


docs/sources/flow/getting-started/collect-prometheus-metrics.md
```diff
@@ -0,0 +1,117 @@
...
+    basic_auth {
+      username      = env("REMOTE_WRITE_USERNAME")
+      password_file = "/etc/secrets/api-key"
+    }
+  }
+}
+
+prometheus.scrape "example" {
+  targets = [{
+    __address__ = "my-application:80",
```
tpaschalis, line 109: Maybe also include a `localhost:12345` target here so that someone trying the example directly can have some metrics flow through?

docs/sources/flow/getting-started/collect-prometheus-metrics.md
```diff
@@ -0,0 +1,117 @@
...
+  }]
+
+  forward_to = [prometheus.remote_write.default.receiver]
+}
+```
+
```
tpaschalis, line 115: I don't want to make this any more difficult to follow, it's pretty succinct as-is. But do we want a sentence guiding our user to the Flow UI so they can see how data Flows, any debug info, etc?

rfratto: I don't think so, since we'd probably end up adding it to every single topic (users aren't expected to follow topics in any particular order). But if they follow the first topic, that does point them in the direction of the UI.

tpaschalis (approved):

docs/sources/flow/getting-started/collect-prometheus-metrics.md
ptodev: Just an idea: there could be a "further reading" or "related pages" at the end which list the next things users could be interested in.

rfratto: We can consider doing that after we've fleshed out the topics more, but I'm also worried that we'll forget to maintain them if we try to keep a list of related topics at the bottom.

docs/sources/flow/getting-started/collect-prometheus-metrics.md
```diff
@@ -0,0 +1,119 @@
...
+* [prometheus.remote_write][]
+* [prometheus.scrape][]
+
+[prometheus.remote_write]: {{< relref "../reference/components/prometheus.remote_write.md" >}}
+[prometheus.scrape]: {{< relref "../reference/components/prometheus.scrape.md" >}}
+
+## Before you begin
+
+* Ensure that you basic familiarity with instrumenting applications with
```
ptodev, line 27:
```suggestion
* Ensure that you have basic familiarity with instrumenting applications with
```

docs/sources/flow/getting-started/collect-prometheus-metrics.md
```diff
@@ -0,0 +1,119 @@
...
+  Prometheus.
+* Have a set of Prometheus exports or applications exposing Prometheus metrics
+  that you want to collect metrics from.
+* Identify where you will write collected metrics, such as Prometheus, Grafana
```
ptodev, line 31:
```suggestion
* Identify a destination where you will write collected metrics. For example: Prometheus, Grafana
```

ptodev: The current working almost makes it sound like Prometheus is the metric itself.

docs/sources/flow/getting-started/collect-prometheus-metrics.md
```diff
@@ -0,0 +1,119 @@
...
+
+   ```river
+   prometheus.remote_write "LABEL" {
+     endpoint {
+       url = "PROMETHEUS_URL"
+     }
+   }
+   ```
+
+2. Replace `LABEL` with a label to use for the component, such as `default`.
```
ptodev, line 60: It might be more clear if each main point is for adding a component, and there are subpoints for modifying the provided component template. So for example, replacing "LABEL" would be 1.1.
This would make it more consistent with the basic_auth component.

docs/sources/flow/getting-started/collect-prometheus-metrics.md
```diff
@@ -0,0 +1,119 @@
...
+   The label chosen must be unique across all `prometheus.remote_write`
+   components in the same configuration file.
+
+3. Replace `PROMETHEUS_URL` with the full URL of the Prometheus-compatible
+   endpoint where metrics will be sent, such as
+   `https://prometheus-us-central1.grafana.net/api/prom/push`.
+
+4. If your endpoint requires basic authentication, complete the following:
+
+    1. Paste the following inside of the `endpoint` block:
```
ptodev, line 70: Should we call this 4.1, so that we can refer to it more easily if needed? And similar for other sub-points.

ptodev (approved):

docs/sources/flow/getting-started/collect-prometheus-metrics.md
rfratto: No, that won't work in our Markdown.

docs/sources/flow/getting-started/collect-prometheus-metrics.md
rfratto: I'm not sure that follows the guidelines for steps, but I'm genuinely not sure. I'll merge as-is and we'll get feedback from the docs team afterwards and clean up if we have to.

PR #2710, opened 2023-01-10
Do not action if the component is not running

ptodev:
#### PR Description
Fixes #2216. The `prometheus.relabel` and  `prometheus.remote_write` components will now error if data is being sent to them while they are stopped. The reason they are in memory and stopped is that during a config refresh they got deleted, but a component which references them keeps a reference to them since it has to be preserved to be in the last good state (the state prior to the refresh). An issue has been raised to question this behaviour (#2709) - it may be better to preserve the last good state of the graph as a whole instead of the last good state of the individual components. If that were the case, we also would be less likely to see bugs such as this one.

#### Test evidence
1. An agent was ran locally with this river file:
```
prometheus.scrape "default" {
	targets    = [{"__address__" = "localhost:12345"}]
	forward_to = [prometheus.relabel.filter.receiver]
}

prometheus.relabel "filter" {
	rule {
		source_labels = ["__name__"]
		regex         = "(.+)"
		replacement   = "api_server"
		target_label  = "service"
	}
	forward_to = [prometheus.remote_write.prom.receiver]
}

prometheus.remote_write "prom" {
	endpoint {
		url = "http://localhost:9009/api/v1/push"
	}
}
```

2. Then the river file was changed to this:
```
prometheus.scrape "default" {
	targets    = [{"__address__" = "localhost:12345"}]
	forward_to = [prometheus.relabel.filter.receiver]
}

prometheus.remote_write "prom" {
	endpoint {
		url = "http://localhost:9009/api/v1/push"
	}
}
```

3. The config was refreshed with this command: `curl localhost:12345/-/reload`

4. The new `component is not running` warnings can be seen in the agent output:
```
% AGENT_MODE=flow ./grafana-agent run /Users/paulintodev/flow_tutorial/tutorials/flow_configs/relabel.river

ts=2023-01-10T10:31:55.289245Z level=info msg="now listening for http traffic" addr=127.0.0.1:12345
ts=2023-01-10T10:31:55.289302Z level=info msg="running usage stats reporter"
ts=2023-01-10T10:31:55.289594Z level=info trace_id=d53f49e021568fdfa2ad9a30a381566a msg="starting complete graph evaluation"
ts=2023-01-10T10:31:55.291231Z level=info component=prometheus.remote_write.prom subcomponent=wal msg="replaying WAL, this may take a while" dir=data-agent/prometheus.remote_write.prom/wal/prometheus.remote_write.prom/wal
[...]

PR #3325, opened 2023-03-22
WIP: Add RFC for export direction

### LINKED ISSUE #3319: Flow: Reevaluate direction of component exports
It may be confusing that component exports can be either the component's input or its output. At the moment Flow doesn't enforce either model. People are free to fix and match them using their best judgement. We need an RFC which evaluates the pros and cons of each approach, and also states if it'd be worth it to change our current approach.

ptodev:
#### PR Description
This is a draft RFC which reevaluates whether we should enforce a specific direction for the exports of Flow components.

Please feel free to drop comments with any suggestions, for example:

* Additional pros and cons
* Additional examples
* Your own conclusion

Fixes grafana/alloy#470

#### Notes to the Reviewer
Most of the lines are very long. It's easiest to edit the markdown this way and have them wrap around in the editor. For the final draft I will make the lines shorter, as is the convention for MD files.

### DIFF
--- a/docs/rfcs/0007-export-direction.md
+++ b/docs/rfcs/0007-export-direction.md
@@ -1,0 +1,221 @@
+# Direction of Flow component exports
+
+* Date: 2023-03-22
+* Author: Paulin Tpdev (@ptodev)
+* PR:
+* Status: Draft
+
+## Summary
+It may be confusing that component exports can be either the component's input or its output. At the moment Flow doesn't enforce either model. People are free to fix and match them using their best judgement. This RFC evaluates the pros and cons of each approach, and also states if it'd be worth it to change our current approach.
+
+## Outcome
+The users would like to see the following:
+
+- When reading a component's documentation, know what is an input and what is an output at a glance.
+- Is should be easy to tell which component's output can be piped to which component's input (composition).
+- Too much reliance on examples is a sign that the system is too hard to use. Browsing the documentation for a component's properties should be enough most of the time. 
+- Learning new components should be easy and feel familiar. It should not feel like a new experience.
+
+## Examples
+
+### Example 1: Prometheus scrape and remote write
+#### Example 1.1: How Flow works at the moment
+```
+prometheus.remote_write "staging" {
+  endpoint {
+    url = "http://mimir:9009/api/v1/push"
+  }
+}
+
+prometheus.scrape "demo1" {
+  targets = [{"__address__" = "127.0.0.1:12345"}]
+  forward_to = [prometheus.remote_write.staging.receiver]
+}
+
+prometheus.scrape "demo2" {
+  targets = [{"__address__" = "127.0.0.2:12345"}]
+  forward_to = [prometheus.remote_write.staging.receiver]
+}
+```
+
+#### Example 1.2: How Flow could work if we do not use exports as inputs
+```
+prometheus.remote_write "staging" {
+  receiver = [prometheus.scrape.demo1, prometheus.scrape.demo2] 
+  endpoint {
+    url = "http://mimir:9009/api/v1/push"
+  }
+}
+
+prometheus.scrape "demo1" {
+  targets = [{"__address__" = "127.0.0.1:12345"}]
+}
+
+prometheus.scrape "demo2" {
+  targets = [{"__address__" = "127.0.0.2:12345"}]
+}
+```
+
+### Example 2: Prometheus scrape converted to Otel format and sent to an Otel endpoint
+#### Example 2.1: How Flow works at the moment
+```
+prometheus.scrape "default" {
+  targets = [{"__address__"   = "127.0.0.1:12345"}]
+  forward_to = [otelcol.receiver.prometheus.default.receiver]
+}
+
+otelcol.receiver.prometheus "default" {
+  output {
+    metrics = [otelcol.exporter.otlp.default.input]
+  }
+}
+
+otelcol.exporter.otlp "default" {
+  client {
+    endpoint = env("OTLP_ENDPOINT")
+  }
+}
+```
+
+#### Example 2.2: How Flow could work if we do not use exports as inputs
+```
+prometheus.scrape "default" {
+    targets = [{"__address__"   = "127.0.0.1:12345"}]
+}
+
+otelcol.receiver.prometheus "default" {
+  receiver = prometheus.scrape.output
+}
+
+otelcol.exporter.otlp "default" {
+  // otelcol.receiver.prometheus.default.output will be of type 
+  // "otelcol.Consumer" instead of an "output" block
+  input = otelcol.receiver.prometheus.default.output
+  client {
+    endpoint = env("OTLP_ENDPOINT")
+  }
+}
+```
+
+## Potential approaches
+
+### Approach 1 - Use exports as both inputs and outputs (current status quo)
+With this approach the component's exports could be either its inputs or its outputs. The writer of each component uses their best judgement on whether having the exports as inputs or outputs would be most appropriate.
+
+#### Pros
+- Maximum flexibility.
+- Prometheus uses a pull model. Otel uses a push model. Supporting both models in Flow could make it easier to express Prometheus and Otel logic. 
+#### Cons
+- Lack of consistency between components can be confusing.
+- Having components in the same config file using different paradigms looks confusing.
+
+### Approach 2 - Use exports as inputs only
+With this approach the component's exports are its inputs. Each component will have to specify to which other components its output data is going to.
+
+#### Pros
+
+#### Cons
+- It may be awkward to modify components which are already configured, so that they send data to more downstream components.
+
+### Approach 3 - Use exports as outputs only
+With this approach the component's exports are its outputs. Other components will feed those outputs to their inputs. Each component will have to specify from which other components its input data is coming from.
+
+#### Pros
+- More intuitive. Newcomers to Flow will browse the documentation and look for inputs and outputs of each component. They will likely assume that the non-exported arguments are the component's inputs, whereas the exported ones are the component's outputs.
+
+#### Cons
+
+## Other considerations
+
+### Which approach would be less verbose to write?
+
+Depending on the situation, either approach could be more verbose. For example:
+
+- A single Prometheus remote write being referenced by multiple scrapers is less verbose when the remote write specifies its inputs in the remote write itself (as an input). This is example 1.2 above.
+- A single scraper referencing multiple remote writes would be less verbose if the remote writes are specified in the scraper itself (as an output).
+
+Hence, nether option can be considered less verbose in the general case.
+
+### Ease of reading
+
+Ideally, one should be able to read a River config file from top to bottom easily. In addition:
+* It should be easy to see where a component's inputs are coming from.
+* It should be easy to see where a component's outputs are going to.
+
+It is not possible to achieve both in practice, just by reading a config file. That's because we only specify either where data came from, or where it's going to. It would be redundant to specify this in both the source component and the destination component. The only way to achieve both of those things is to use a UI tool for visualizing the graph.
+
+Approach 1 makes tracing dependencies easy for some components like this one...
+```
+prometheus.scrape "default" {
+  targets    = discovery.kubernetes.pods.targets
+  forward_to = [prometheus.remote_write.default.receiver]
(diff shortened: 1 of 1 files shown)


docs/rfcs/0007-export-direction.md
```diff
@@ -0,0 +1,221 @@
...
+  forward_to = [prometheus.remote_write.staging.receiver]
+}
+
+prometheus.scrape "demo2" {
+  targets = [{"__address__" = "127.0.0.2:12345"}]
+  forward_to = [prometheus.remote_write.staging.receiver]
+}
+```
+
+#### Example 1.2: How Flow could work if we do not use exports as inputs
```
tpaschalis, line 41: Could we list potential solutions separately? I think this makes the document difficult to follow.

docs/rfcs/0007-export-direction.md
```diff
@@ -0,0 +1,221 @@
...
+  client {
+    endpoint = env("OTLP_ENDPOINT")
+  }
+}
+```
+
+## Potential approaches
+
+### Approach 1 - Use exports as both inputs and outputs (current status quo)
+With this approach the component's exports could be either its inputs or its outputs. The writer of each component uses their best judgement on whether having the exports as inputs or outputs would be most appropriate.
```
tpaschalis, line 103: Could we please define what 'inputs' and 'outputs' are in this case?

docs/rfcs/0007-export-direction.md
```diff
@@ -0,0 +1,221 @@
...
+#### Pros
+
+#### Cons
+- It may be awkward to modify components which are already configured, so that they send data to more downstream components.
+
+### Approach 3 - Use exports as outputs only
+With this approach the component's exports are its outputs. Other components will feed those outputs to their inputs. Each component will have to specify from which other components its input data is coming from.
+
+#### Pros
+- More intuitive. Newcomers to Flow will browse the documentation and look for inputs and outputs of each component. They will likely assume that the non-exported arguments are the component's inputs, whereas the exported ones are the component's outputs.
```
tpaschalis, line 124: I couldn't follow what this sentence meant. Let's try and use language consistent with the current Flow terminology and documentation.

docs/rfcs/0007-export-direction.md
```diff
@@ -0,0 +1,221 @@
...
+
+Hence, nether option can be considered less verbose in the general case.
+
+### Ease of reading
+
+Ideally, one should be able to read a River config file from top to bottom easily. In addition:
+* It should be easy to see where a component's inputs are coming from.
+* It should be easy to see where a component's outputs are going to.
+
+It is not possible to achieve both in practice, just by reading a config file. That's because we only specify either where data came from, or where it's going to. It would be redundant to specify this in both the source component and the destination component. The only way to achieve both of those things is to use a UI tool for visualizing the graph.
```
tpaschalis, line 145: The UI currently does _not_ display data flows, but graph dependencies. Would the proposed change help with that?

docs/rfcs/0007-export-direction.md
```diff
@@ -0,0 +1,221 @@
...
+```
+
+... but it makes it extra hard for other components such as this:
+```
+discovery.kubernetes "pods" {
+  role = "pod"
+}
+```
+
+It may have been easier to read the file if `discovery.kubernetes` listed it outputs like so:
```
tpaschalis, line 162: ditto not sure what we mean by `outputs` here.

docs/rfcs/0007-export-direction.md
```diff
@@ -0,0 +1,221 @@
...
+### Approach 2 - Use exports as inputs only
+With this approach the component's exports are its inputs. Each component will have to specify to which other components its output data is going to.
+
+#### Pros
+
+#### Cons
+- It may be awkward to modify components which are already configured, so that they send data to more downstream components.
+
+### Approach 3 - Use exports as outputs only
+With this approach the component's exports are its outputs. Other components will feed those outputs to their inputs. Each component will have to specify from which other components its input data is coming from.
```
tpaschalis, line 121: > Other components will feed those outputs to their inputs

Does this mean 'arguments'? I'd advise including a small example for each approach as now it's not really clear what is intended.

docs/rfcs/0007-export-direction.md
```diff
@@ -0,0 +1,221 @@
+# Direction of Flow component exports
+
+* Date: 2023-03-22
+* Author: Paulin Tpdev (@ptodev)
+* PR:
+* Status: Draft
+
+## Summary
+It may be confusing that component exports can be either the component's input or its output. At the moment Flow doesn't enforce either model. People are free to fix and match them using their best judgement. This RFC evaluates the pros and cons of each approach, and also states if it'd be worth it to change our current approach.
```
tpaschalis, line 9: Could we please define what 'inputs' and 'outputs' are early on and keep the terminology consistent throughout the doc? Do we mean _data_ flow? How does this impact components that do not flow data through (eg. an authentication component).

docs/rfcs/0007-export-direction.md
```diff
@@ -0,0 +1,221 @@
...
+
+### Learning curve
+
+The more consistent our approach, the easiest it is to explain in documentation. In order to link two components, at the moment sometimes users specify the next component in the "outputs" section, and sometimes they specify the previous component in the "inputs" section. This can be confusing. 
+
+### Using components from a River file which cannot or should not be modified (River libraries?)
+
+In this case, it is best for the upstream component not to specify all the downstream components which use its output. That way components which need data that the upstream exports could simply use it in any of the downstream's attributes.
+
+### Preventing dangling references to shut down components
```
ptodev, line 188: @rfratto mentioned in a global comment:
> I’m concerned we’re not touching enough on how it will be possible to reverse the direction of some components in a way which is easy to maintain and that isn’t prone to bugs.

I'll continue the discussion in this thread because it might get lengthy :) Before we discuss a solution, I'd like to firstly discuss the problem to make sure we understand it.

### What bugs could happen if we change the exports direction?
I suppose that the bugs @rfratto is referring to are mostly dangling references? Please let me know if there could be other ones too? Below I'll include my thoughts on dangling references.

### Can we decouple the representation of the graph in config from how the graph is built in the code?
I am surprised that our ability to treat dangling references is related to the way the graph is described in config. I would have expected this to be decoupled more. Do we want to try decoupling it so that we can do future changes with less friction?

Regardless of how the graph is represented in config, IMO all users infer from it is that there is an edge between two nodes on a graph. We could handle the internals as we see fit. The config representation is mostly for readability and aesthetics.

### When could we end up with dangling references? 
I can think of two situations:

1. A misconfiguration during a config reload. This is what this issue was patching:
#2710
Maybe such reload problems could be fixed for all components via:
#2709
2. A component stops working spontaneously. What should we do then?
    * Should we restart the component?
    * Restart the agent process? We almost certainly can't do that, because if we shut down we don't know if the Agent was set up with some sort of restart policy.
    * Is there any other way to attempt repairing it?
    * What do the components which depend on the failing component do? Do they restart? It could be that they are the real cause of the problem and maybe they need to be restarted?
    * What about buildup of data to the failing component?

### Have a solution at least as good as the one today. What issues does the current implementation prevent?
The current implementation prevents some dangling references. E.g. if `prometheus.remote_write` shuts down, it will remove references to other components. However, other components such as `prometheus.scrape` would still have a dangling reference to it. And as shown by ##2710, the current implementation could lead to bugs as well. Maybe we can't find a bug-free solution, but could make something at least as good as the current one.

### What should the user experience look like?
When we agree on the desired user experience during component failure, I think we could start discussing potential solutions.

ptodev:
An additional interesting use case is the one described in #3939 and #3656

docs/rfcs/0007-export-direction.md
rfratto: > I suppose that the bugs @rfratto is referring to are mostly dangling references? 

Let me give an example of what I'm talking about here:

Today, components which expose current information are pull-based: 

```river 
discovery.kubernetes "pods" {
  role = "pod" 
}

prometheus.scrape "pods" {
  // Pull from discovery.kubernetes.pods.
  targets = discovery.kubernetes.pods.targets 
}
```

If `discovery.kubernetes.pods` is removed from the configuration file, it will cause a evaluation error with prometheus.scrape.pods, and prometheus.scrape.pods will be left in its last valid state until the evaluation error is resolved. Once the reference is removed from prometheus.scrape.pods, everything becomes healthy again. 

Let's assume now that the direction is reversed:

```river 
discovery.kubernetes "pods" {
  role = "pod" 

  // Push targets to prometheus.scrape.pods. 
  send_to = [prometheus.scrape.pods.targets] 
}

prometheus.scrape "pods" { }
```

The implementation of this requires: 

1. Some kind of stateful operation for discovery.kubernetes.pods to notify prometheus.scrape.pods of new targets. 
2. prometheus.scrape.pods must be able to correctly combine targets received from multiple senders, implying it must cache targets. 

The implementation means that push-based discovery component authors must find a way to unregister the latest targets from whoever they sent it to. If they don't, and discovery.kubernetes.pods is removed from the config file, the targets leak, and prometheus.scrape.pods holds on to targets forever with no signal to the user; the user doesn't know things are unhealthy. 

This isn't an issue with pushing events, but it's a complexity introduced when you're talking about synchronizing state via pushes. 

Does that make sense?

rfratto: The same problem exists with swapping event-based components to be pull-based instead:

```river 
prometheus.scrape "example" {
  ... 
}

prometheus.remote_write "example" {
  read_from = [prometheus.scrape.example.output]
}
```

This implies that:

1. prometheus.remote_write.example must subscribe to prometheus.scrape.example. 
2. prometheus.scrape.example must keep track of all components which is subscribed. 
3. prometheus.remote_write.example must unsubscribe on shutdown, otherwise prometheus.scrape.example will keep a dangling reference forever.

rfratto: So, to summarize, the issue is that swapping the direction of these components _is_ decoupled from the controller; the current implementation being coupled with the controller means that components can't be implemented in a way which causes memory leaks in _other components._

rfratto: So when I say 

> I’m concerned we’re not touching enough on how it will be possible to reverse the direction of some components in a way which is easy to maintain and that isn’t prone to bugs.

I mean we need to find a way to implement an event system (for everything-is-push) or a subscription system (for everything-is-pull) which means that it's either impossible or extremely difficult for component authors to cause a memory leak in an unrelated component.

thampiotr: I'm slightly confused - on one hand we must do pull and cannot do push for targets, but on the other hand we must to push and we cannot do pull for events. What is the essential difference between targets and events that makes it so? 

I get it that particularly for one-to-many connections, the push model is easier to manage. But I don't get it why push model cannot be used with targets - can't we treat a discovery of targets like an event that is sent to many components on `forward_to` list? If the implementation can cause a leak for targets, why are there no leaks for 'events'?

rfratto: @thampiotr The primary difference is that events are stateless, while targets are stateful, as they declare the current state of things. 

> But I don't get it why push model cannot be used with targets - can't we treat a discovery of targets like an event that is sent to many components on forward_to list? 

I've tried to describe the challenges with it here: 

https://github.com/grafana/agent/pull/3325#discussion_r1200459648

Was there any part of this that wasn't clear that I can help clarify? 

> If the implementation can cause a leak for targets, why are there no leaks for 'events'?

Because events aren't stateful; you don't have to persist all of the events you've received over time to build an aggregated combined list like you have to with targets.

thampiotr: That's a fair point, but if you were able to hide this behind the scenes for pull-based approach, it should be possible to hide it also for a push-based approach. Right now `concat` will combine most recently exported values and call `Update` for components that consume it. We could have a receiver that does the opposite of fan-out - return a new channel for every producer that wants to write to it and combine the most recently sent values for the consumer (while timing out and defaulting to 0 for dead channels).

What is kinda interesting is that we already have a de-facto push-based model for targets, because River pushes the updates with a call to `Update` function ;)

ptodev:
Closing this since it's not a priority at the moment. It'll also likely need a whole new discussion if we're to consider it again.

PR #3326, opened 2023-03-22
Convert() method for otel auth extensions to return an error

ptodev:
#### PR Description
If the Convert() method returns an error, this allows us to handle errors in UnmarshalRiver without doing a type assertion. It also makes sense to return an error if a conversion could not be done. This change was discussed in a [comment](https://github.com/grafana/agent/pull/3200#discussion_r1143298235) in a prior PR.

#### Notes to the reviewer

It is such a minor change, that I think there is no need for a changelog entry?

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/component/otelcol/auth/auth.go
+++ b/component/otelcol/auth/auth.go
@@ -31,7 +31,7 @@ type Arguments interface {
 
 	// Convert converts the Arguments into an OpenTelemetry Collector
 	// authentication extension configuration.
-	Convert() otelconfig.Extension
+	Convert() (otelconfig.Extension, error)
 
 	// Extensions returns the set of extensions that the configured component is
 	// allowed to use.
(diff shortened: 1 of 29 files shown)


rfratto:
@ptodev I know it will bloat the PR size, but do you mind making the same change to receiver, processor, and exporter components? That way we keep things consistent.

> It is such a minor change, that I think there is no need for a changelog entry?

No, there'd be no changelog entry here because there's no user-visible behavior change.

ptodev:
> @ptodev I know it will bloat the PR size, but do you mind making the same change to receiver, processor, and exporter components? That way we keep things consistent.

Good point! I updated them now - please have a look.

rfratto (approved):
LGTM, thanks!

PR #3336, opened 2023-03-22
docs: add Collect metrics from Kubernetes Pods task

rfratto:
Add a _Collect metrics from Kubernetes Pods_ task to the _Collect Prometheus Metrics_ topic.

This is a follow-up PR on #3332.

Note that I always refer to Pods with capitalization here, since that's how the Kubernetes docs seem to exclusively refer to them.

### DIFF
--- a/docs/sources/flow/getting-started/collect-prometheus-metrics.md
+++ b/docs/sources/flow/getting-started/collect-prometheus-metrics.md
@@ -137,7 +137,7 @@ To collect metrics from Kubernetes Pods, complete the following steps:
 2. Discover Kubernetes Pods:
 
     1. Add the following `discovery.kubernetes` component to your configuration
-       file:
+       file to discover every Pod in the cluster across all Namespaces:
 
        ```river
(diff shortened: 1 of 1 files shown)


erikbaranowski (approved):
Content reads well and `make docs` test looked good.

tpaschalis (approved):

docs/sources/flow/getting-started/collect-prometheus-metrics.md
```diff
@@ -117,3 +120,148 @@ prometheus.scrape "example" {
...
+       file:
+
+       ```river
+       discovery.kubernetes "DISCOVERY_LABEL" {
+         role = "pod"
+       }
+       ```
+
+       This will generate one Prometheus target for every exposed port on every
+       discovered Pod.
```
tpaschalis, line 149: Nit: it's kinda implied that this gets _all_ pods it can reach, right? Do we also want to spell it out?

PR #3338, opened 2023-03-22
module.file new module loader component

### LINKED ISSUE #3311: Flow Modules: Add `module.file` component

erikbaranowski:

Closes #3311 

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated


component/module/file/file.go
```diff
@@ -0,0 +1,247 @@
...
+		return err
+	}
+
+	return c.LoadFlowContent()
+}
+
+// NewManagedLocalFileComponent creates the new local.file managed component.
+func (c *Component) NewManagedLocalFileComponent(o component.Options, args Arguments) (*file.Component, error) {
+	localFileOpts := o
+	localFileOpts.OnStateChange = func(e component.Exports) {
```
erikbaranowski, line 172: Is this the only component.Options variable that should be overridden when running a managed component? I wasn't entirely sure on this 1.

docs/sources/flow/reference/components/module.file.md
```diff
@@ -0,0 +1,165 @@
...
+
+`arguments` allows us to pass parameterized input into a module. The values
+passed in `arguments` correspond to [argument blocks][] defined in the module
+source.
+
+An `argument` marked non-optional in the module being loaded is required in the
+`arguments`. It is also not valid to provide an `argument` not defined in the
+module being loaded.
+
+### File change detectors
```
erikbaranowski, line 55: A bunch of this was copied from local.file.md but I'm not sure if we want to create a shared .md for situations or if that overcomplicates it.

component/module/file/file_test.go
```diff
@@ -93,12 +95,20 @@ func TestMissingFile(t *testing.T) {
...
 	require.ErrorContains(t, err, "failed to read file:")
 }
+
+func riverEscape(filePath string) string {
+	if runtime.GOOS == "windows" {
+		return strings.Replace(filePath, `\`, `\\`, -1)
+	}
+
+	return filePath
+}
```
rfratto, line 114: I think you can use filepath.FromSlash here instead of defining a new function.

erikbaranowski: I'm open to suggestion on a better way but I'm specifically escaping the filepath for putting it into the river config. The suggested function I don't think is quite what I need to do the transform from `C:\Users\erikr\AppData\Local\Temp\TestMissingFile1834357519\002\module.river` to `C:\\Users\\erikr\\AppData\\Local\\Temp\\TestMissingFile1834357519\\002\\module.river`

(I did try it just to sanity check)

```golang
// FromSlash returns the result of replacing each slash ('/') character
// in path with a separator character. Multiple slashes are replaced
// by multiple separators.
func FromSlash(path string) string {
	if Separator == '/' {
		return path
	}
	return strings.ReplaceAll(path, "/", string(Separator))
}
```

erikbaranowski: I thought about running the replace on all operating systems then chose to just mess with windows. I could also do the replace up above alternatively across all operating systems and not need a helper function but it may not be as clear as to why I am doing it.

rfratto: Doh, sorry, I didn't think it through enough. You could use `filepath.ToSlash` here but this is fine.

component/module/file/file.go
rfratto: You shouldn't override this ever, OnStateChange is a function you invoke to pass exports back to the controller.

rfratto: Argh, sorry. The creator of a component is free to modify component.Options for whatever it needs to do IMO

component/module/file/file.go
```diff
@@ -0,0 +1,247 @@
...
+
+		return err
+	}
+
+	return nil
+}
+
+// Handler implements component.HTTPComponent.
+func (c *Component) Handler() http.Handler {
+	r := mux.NewRouter()
```
mattdurham, line 220: Same as above, though part of me cringes at the mixing of api level info with a component.

docs/sources/flow/reference/components/module.file.md
mattdurham: How difficult is using a shared.md? Since we only have it copied twice it doesnt feel bad to manually sync unless making the shared.md is easy.

erikbaranowski: Not that hard, I will give it a go and we will see what it looks like.

component/module/file/file.go
erikbaranowski: > Is this the only component.Options variable that should be overridden when running a managed component? I wasn't entirely sure on this 1.

leaving as `unresolved` relative to the above question

mattdurham (approved):
LGTM

PR #3344, opened 2023-03-23
Add logging around the /-/reload endpoint

rfratto:
This is a workaround to have some logging around when reloads are requested until a longer-term fix for logging on the HTTP handler is introduced.


cyriltovena (approved):
LGTM

PR #3349, opened 2023-03-23
Add an otelcol.processor.attributes component

### LINKED ISSUE #2294: Flow: otelcol.processor.attributes component
Expose the OpenTelemtry Collector `attributes` processor as a Flow component called `otelcol.processor.attributes`. This component is from the `otelcol-contrib` distribution.

ptodev:
Adding an otelcol.processor.attributes component.

Fixes #2294

I left some comments in the PR about things I was not sure about.

Tests, docs, and changelog have been updated.


docs/sources/flow/reference/components/otelcol.processor.attributes.md
```diff
@@ -0,0 +1,582 @@
...
+    * `key` is required. It specifies the attribute to act upon.
+    * `action = "convert"` is required.
+    * `converted_type` is required and must be one of int, double or string.
+
+* `delete`: Deletes an attribute from the input data.
+
+    * `key` and/or `pattern` is required. It specifies the attribute to act upon.
+    * `action = "delete"` is required.
+
+### include/exclude blocks
```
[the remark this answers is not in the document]
ptodev: Also, title "include/exclude blocks" could be misinterpreted as "how to include or exclude blocks".

docs/sources/flow/reference/components/otelcol.processor.attributes.md
```diff
@@ -0,0 +1,582 @@
...
+
+This block specifies items to match the implementation library against.
+A match occurs if the span's implementation library matches at least one `library` block.
+
+The following arguments are supported:
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`name` | `string` | The attribute key. | | yes
+`version` | | The value to match against. | | yes
```
ptodev, line 222: `version` is a pointer. Should we include a note in the docs that leaving this unset and setting it to an empty string are not the same thing, and could lead to different outcomes?

component/otelcol/config_filter_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+import (
+	"testing"
+
+	"github.com/grafana/agent/component/otelcol"
+	"github.com/stretchr/testify/require"
+)
+
+func TestConvertMatchProperties(t *testing.T) {
+	version := "3.8.0"
```
ptodev, line 11: Idk if there is a better way to do this inline at `Version: &version,`?

docs/sources/flow/reference/components/otelcol.processor.attributes.md
```diff
@@ -0,0 +1,582 @@
...
+  grpc {}
+
+  output {
+    metrics = [otelcol.processor.attributes.default.input]
+    logs    = [otelcol.processor.attributes.default.input]
+    traces  = [otelcol.processor.attributes.default.input]
+  }
+}
+
+otelcol.processor.attributes "default" {
```
ptodev, line 264: Add an example for `from_context`?

docs/sources/flow/reference/components/otelcol.processor.attributes.md
```diff
@@ -0,0 +1,582 @@
...
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`key` | `string` | The attribute key. | | yes
+`value` | `any` | The value to match against. | | no
+
+If `value` is not set, any value will match.
+The type of `value` could be a number, a string or a boolean.
+
+### resource block
```
ptodev, line 197: This is the same as the attribute block. Should I repeat it, or try have it under the same heading ("attributes/resource block")?

component/otelcol/config_attraction.go
```diff
@@ -0,0 +1,103 @@
+package otelcol
+
+type AttrActionKeyValueSlice []AttrActionKeyValue
```
ptodev, line 3: This pattern is something I made up myself, since it makes sense to use actions as a slice and to have the convert function in config_atraction.go. If you think this pattern doesn't make senes though, let me know

docs/sources/flow/reference/components/otelcol.processor.attributes.md
```diff
@@ -0,0 +1,582 @@
...
+otelcol.processor.attributes "default" {
+    include {
+        // "match_type" of "regexp" defines that the "value" attributes 
+        // in the "attribute" blocks are regexp-es.
+        match_type = "regexp"
+
+        // This attribute ('db.statement') must exist in the span and match 
+        // the regex ('SELECT \* FROM USERS.*') for a match.
+        attribute {
+            key = "env"
```
ptodev, line 506: Shouldn't key = "env" be key = "db.statement"? This was taken from the [examples](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.63.0/processor/attributesprocessor/testdata/config.yaml#L303) in the otel repo.

docs/sources/flow/reference/components/otelcol.processor.attributes.md
```diff
@@ -0,0 +1,582 @@
...
+`services` | `list(string)` | A list of items to match the service name against. | `[]` | no
+`span_names` | `list(string)` | A list of items to match the span name against. | `[]` | no
+`log_bodies` | `list(string)` | A list of strings that the LogRecord's body field must match against. | `[]` | no
+`log_severity_texts` | `list(string)` | A list of strings that the LogRecord's severity text field must match against. | `[]` | no
+`metric_names` | `list(string)` | A list of strings to match the metric name against. | `[]` | no
+`span_kinds` | `list(string)` | A list of items to match the span kind against. | `[]` | no
+
+`match_type` is required and can be set to either `"regexp"` or `"strict"`.
+
+One of the following is also required:
```
ptodev, line 156: The Otel docs also mentions [expressions](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.63.0/processor/attributesprocessor#includeexclude-filtering). However, I don't see any "expressions" in the code, so I removed the reference to it here.

Also, the Otel docs didn't mention "span_kinds", which I included here.

I could raise a PR to the Otel repo to also fix it on their end?

component/otelcol/config_attraction.go
```diff
@@ -0,0 +1,103 @@
...
+	if args.FromAttribute != "" {
+		res["from_attribute"] = args.FromAttribute
+	}
+	if args.FromContext != "" {
+		res["from_context"] = args.FromContext
+	}
+	if args.ConvertedType != "" {
+		res["converted_type"] = args.ConvertedType
+	}
+	return res
```
rfratto, line 102: Is there a reason you can't do 

```go
return map[string]interface{}{ 
  "key": args.Key,
  "action": args.Action,
  "value": args.Value,
  "pattern": args.RegexPattern,
  "from_attribute": args.FromAttribute,
  "from_context": args.FromContext,
  "converted_type": args.ConvertedType,
}
```

here?

component/otelcol/config_filter.go
```diff
@@ -0,0 +1,215 @@
...
+	// A match occurs if the span's implementation library matches at least one item in this list.
+	Libraries []InstrumentationLibrary `river:"library,block,optional"`
+
+	// SpanKinds specify the list of items to match the span kind against.
+	// A match occurs if the span's span kind matches at least one item in this list.
+	SpanKinds []string `river:"span_kinds,attr,optional"`
+}
+
+// Convert converts args into the upstream type.
+func (args *MatchProperties) Convert() map[string]interface{} {
```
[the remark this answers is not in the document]
rfratto: No, this is a pretty big struct to use a non-pointer receiver for. Using a pointer here is the right move IMO.

rfratto, line 71: ditto with unexporting the Convert methods here and elsewhere wherever they're defined

component/otelcol/config_filter.go
```diff
@@ -0,0 +1,215 @@
...
+	if args == nil {
+		return nil
```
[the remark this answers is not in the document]
rfratto: nil so nothing gets allocated

component/otelcol/config_filter.go
```diff
@@ -0,0 +1,215 @@
...
+	//  <blank>  1       no
+	//  1        <blank> no
+	//  1        1       yes
+	Version *string `river:"version,attr"`
+}
+
+func (args InstrumentationLibrary) Convert() map[string]interface{} {
+	return map[string]interface{}{
+		"name":    args.Name,
+		"version": strings.Clone(*args.Version),
```
rfratto, line 195: I think this will panic if `version = null` is configured in the River config, can we add a if statement here for safety?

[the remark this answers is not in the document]
rfratto: It's better to make a copy, I think.

component/otelcol/config_filter_test.go
rfratto: You can use [pointer.String](https://pkg.go.dev/k8s.io/utils/pointer#String) from k8s.io/utils/pointer.

docs/sources/flow/reference/components/otelcol.processor.attributes.md
```diff
@@ -0,0 +1,582 @@
...
+include > attribute | [attribute] | A list of attributes to match against. | no
+include > resource | [resource] | A list of items to match the resources against. | no
+include > library | [library] | A list of items to match the implementation library against. | no
+include > log_severity_number | [library] | How to match against a log record's SeverityNumber, if defined. | no
+exclude | [include/exclude] | Filter for data to be excluded from this processor's actions | no
+exclude > regexp | [regexp] | Regex cache settings. | no
+exclude > attribute | [attribute] | A list of attributes to match against. | no
+exclude > resource | [resource] | A list of items to match the resources against. | no
+exclude > library | [library] | A list of items to match the implementation library against. | no
+exclude > log_severity_number | [library] | How to match against a log record's SeverityNumber, if defined. | no
```
rfratto, line 55:
```suggestion
include | [include/exclude][] | Filter for data to be included to this processor's actions. | no
include > regexp | [regexp][] | Regex cache settings. | no
include > attribute | [attribute][] | A list of attributes to match against. | no
include > resource | [resource][] | A list of items to match the resources against. | no
include > library | [library][] | A list of items to match the implementation library against. | no
include > log_severity_number | [library][] | How to match against a log record's SeverityNumber, if defined. | no
exclude | [include/exclude][] | Filter for data to be excluded from this processor's actions | no
exclude > regexp | [regexp][] | Regex cache settings. | no
exclude > attribute | [attribute][] | A list of attributes to match against. | no
exclude > resource | [resource][] | A list of items to match the resources against. | no
exclude > library | [library][] | A list of items to match the implementation library against. | no
exclude > log_severity_number | [log_severity_number][] | How to match against a log record's SeverityNumber, if defined. | no
```

docs/sources/flow/reference/components/otelcol.processor.attributes.md
```diff
@@ -0,0 +1,582 @@
...
+
```
rfratto, line 56: You're missing the sentence which explains the meaning of the `>` symbol: https://github.com/grafana/agent/blob/main/docs/developer/writing-flow-component-documentation.md#blocks

docs/sources/flow/reference/components/otelcol.processor.attributes.md
```diff
@@ -0,0 +1,582 @@
...
+---- | ---- | ----------- | ------- | --------
+`key` | `string` | The attribute that the action relates to. |  | yes
+`action` | `string` | The type of action to be performed. |  | yes
+`value` | `any` | The value to populate for the key. |  | no
+`pattern` | `string` | A regex pattern. | `""` | no
+`from_attribute` | `string` | The attribute from the input data to use to populate the attribute value. | `""` | no
+`from_context` | `string` | The context value to use to populate the attribute value.  | `""` | no
+`converted_type` | `string` | The type to convert the attribute value to. | `""` | no
+
+The type of `value` could be a number, a string or a boolean.
```
rfratto, line 82:
```suggestion
The type of `value` must be either a number, string, or boolean.
```

docs/sources/flow/reference/components/otelcol.processor.attributes.md
```diff
@@ -0,0 +1,582 @@
...
+
+The supported values for `action` are:
+
+* `insert`: Inserts a new attribute in input data where the key does not already exist.
+
+    * `key` is required. It specifies the attribute to act upon.
+    * One of `value`, `from_attribute` or `from_context` is required.
+    * `action = "insert"` is required.
```
rfratto, line 90: `action = "insert"` being required when `action` is `insert` is a little confusing to me. Can we drop this? 

Ditto with the other instances of this below (`action = "upsert"`, etc)

docs/sources/flow/reference/components/otelcol.processor.attributes.md
```diff
@@ -0,0 +1,582 @@
...
+
+* `update`: Updates an attribute in input data where the key does exist.
+
+    * `key` is required. It specifies the attribute to act upon.
+    * One of `value`, `from_attribute` or `from_context` is required.
+    * `action = "update"` is required.
+
+* `upsert`: Performs insert or update. Inserts a new attribute in input data where the key does not already exist and updates an attribute in input data where the key does exist.
```
rfratto, line 98:
```suggestion
* `upsert`: Either inserts a new attribute in input data where the key does not already exist, or updates an attribute in input data where the key does exist.
```

docs/sources/flow/reference/components/otelcol.processor.attributes.md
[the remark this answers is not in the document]
rfratto: > I don't want to have a ### include block and a ### exclude block separately, and copy paste the same contents.

I would prefer having both an `include` and `exclude` block; we should be optimizing for the reading experience over the writing experience. 

Instead of copying and pasting, you can put the content in docs/shared and use the docs/shared shortcode to deduplicate.

docs/sources/flow/reference/components/otelcol.processor.attributes.md
rfratto: > I could raise a PR to the Otel repo to also fix it on their end?

Raising it upstream would be good OSS citizen behavior 👍

docs/sources/flow/reference/components/otelcol.processor.attributes.md
```diff
@@ -0,0 +1,582 @@
...
+---- | ---- | ----------- | ------- | --------
+`match_type` | `string` | Controls how items in "services" and "span_names" arrays are interpreted. | | yes
+`services` | `list(string)` | A list of items to match the service name against. | `[]` | no
+`span_names` | `list(string)` | A list of items to match the span name against. | `[]` | no
+`log_bodies` | `list(string)` | A list of strings that the LogRecord's body field must match against. | `[]` | no
+`log_severity_texts` | `list(string)` | A list of strings that the LogRecord's severity text field must match against. | `[]` | no
+`metric_names` | `list(string)` | A list of strings to match the metric name against. | `[]` | no
+`span_kinds` | `list(string)` | A list of items to match the span kind against. | `[]` | no
+
+`match_type` is required and can be set to either `"regexp"` or `"strict"`.
```
rfratto, line 154:
```suggestion
`match_type` is required and must be set to either `"regexp"` or `"strict"`.
```

docs/sources/flow/reference/components/otelcol.processor.attributes.md
```diff
@@ -0,0 +1,582 @@
...
+
+One of the following is also required:
+* For spans, one of `services`, `span_names`, `span_kinds`, [attribute], [resource], or [library] must be specified with a non-empty value for a valid configuration. The `log_bodies`, `log_severity_texts`, `resource_attributes` and `metric_names` fields are invalid.
+* For logs, one of `log_bodies`, `log_severity_texts`, [attribute], [resource], or [library] must be specified with a non-empty value for a valid configuration. The `span_names`, `span_kinds`, `metric_names`, `resource_attributes`, and `services` fields are invalid.
+* For metrics, one of `metric_names`, [resource] must be specified with a valid non-empty value for a valid configuration. The `span_names`, `span_kinds`, `log_bodies`, `log_severity_texts` and `services` fields are invalid.
```
rfratto, line 159:
```suggestion
* For spans, one of `services`, `span_names`, `span_kinds`, [attribute][], [resource][], or [library][] must be specified with a non-empty value for a valid configuration. The `log_bodies`, `log_severity_texts`, `resource_attributes` and `metric_names` attributes are invalid.
* For logs, one of `log_bodies`, `log_severity_texts`, [attribute][], [resource][], or [library][] must be specified with a non-empty value for a valid configuration. The `span_names`, `span_kinds`, `metric_names`, `resource_attributes`, and `services` attributes are invalid.
* For metrics, one of `metric_names`, [resource][] must be specified with a valid non-empty value for a valid configuration. The `span_names`, `span_kinds`, `log_bodies`, `log_severity_texts` and `services` attributes are invalid.
```

docs/sources/flow/reference/components/otelcol.processor.attributes.md
rfratto: No, don't merge blocks please, even if they have the same arguments.

docs/sources/flow/reference/components/otelcol.processor.attributes.md
rfratto: I don't think it has to be a pointer on our side, right? We can turn it into a pointer when converting to the upstream type.

docs/sources/flow/reference/components/otelcol.processor.attributes.md
ptodev: Ok great, I raised a [PR](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/20855) :)

docs/sources/flow/reference/components/otelcol.processor.attributes.md
ptodev: The matching rules for null pointers and empty strings work differently:
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/internal/filter/filterconfig/config.go#L214-L226

If `version` is a null pointer, it will match any version. If `version` is an empty string, it will only match a version which is also an empty string.

I will try to add some unit tests to make sure this works. If it doesn't work, then I could probably add an extra bool flag in the river config so that our code sets the version to a null pointer during the conversion.

docs/sources/flow/reference/components/otelcol.processor.attributes.md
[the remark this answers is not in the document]
ptodev: It is not trivial to reuse code by putting it in `docs/shared`, because the "include block" and "exclude block" sections contain references to blocks such as `[attribute][]` and `[resource][]`. 

If we were to put the shared code in `docs/shared`, we could do one of those things:
1. Unlink the links in the shared code and leave them as text. This means changing `[attribute][]` to `attribute`. I think this wouldn't be a good solution, because the `attribute` block is described in the "Blocks" section. So the config in `docs/shared` for the "include block" would not be self sufficient.
2. Move the text with links to the "Blocks" section. This also wouldn't make sense, since the attributes in the "include" block are not mentioned yet. 
3. Move only the table to a shared md file, to avoid some repetition. But copy paste the bullet points about "One of the following is also required:"

Basically the problem is that in the "include block" description, we have to mention both its attributes and its blocks. But because the blocks definition is not in the "include block, it is hard to reuse md code.

For now I'll go with option 3, but please let me know if there is a better way.

The fact that blocks and attributes are defined separately in config suggests that they should be described separately, but in this case I cannot think of a reason why describing them in one place is wrong, and I can't think of a way to describe them separately.

docs/sources/flow/reference/components/otelcol.processor.attributes.md
```diff
@@ -0,0 +1,597 @@
...
+
+### log_severity_number block
+
+This block defines how to match based on a log record's SeverityNumber field.
+
+The following arguments are supported:
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`min` | `int` | The lowest severity that may be matched. | | yes
```
ptodev, line 231: Not sure what is the best way to document the allowed [plog values](https://github.com/open-telemetry/opentelemetry-collector/blob/v0.64.0/pdata/plog/logs.go#L76). I think I'll just copy all the values here, so that we don't link to the otel code.

rfratto: Yes, you should copy the values. But also, I think we should define our own string-based mapping here using encoding.TextMarshaler and encoding.TextUnmarshaler because `min = 7` is going to confuse people.

docs/sources/flow/reference/components/otelcol.processor.attributes.md
```diff
@@ -0,0 +1,637 @@
...
+
+### regexp block
+
+This block is optional configuration for the `match_type` of `"regexp"`.
+
+The following arguments are supported:
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`cacheenabled` | `bool` | Determines whether match results are LRU cached. | `false` | no
```
mattdurham, line 173: `cache_enabled` seems more inline with our naming. Same for `cache_max_num_entries`

mattdurham: Link to what LRU is  or spell it out?

component/otelcol/config_filter.go
```diff
@@ -0,0 +1,278 @@
...
+	"FATAL4": 24,
+}
+
+// UnmarshalText implements encoding.TextUnmarshaler for SeverityLevel.
+func (sl *SeverityLevel) UnmarshalText(text []byte) error {
+	if numVal, exists := severityLevels[SeverityLevel(text)]; exists {
+		// Check if this is a valid plog severity number
+		plogInt := plog.SeverityNumber(numVal)
+		plogStr := plogInt.String()
+		if plogStr == "SEVERITY_NUMBER_"+string(text) {
```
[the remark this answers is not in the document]
mattdurham: That feels entirely reasonable

docs/sources/flow/reference/components/otelcol.processor.attributes.md
```diff
@@ -0,0 +1,637 @@
...
+---- | ---- | ----------- | ------- | --------
+`min` | `string` | The lowest severity that may be matched. | | yes
+`match_undefined` | `bool` | Whether logs with "undefined" severity match. | | yes
+
+If `match_undefined` is true, entries with undefined severity will match.
+
+The severities supported by Otel are listed in the table below. 
+The value for `min` should be one of the values in the "Log Severity" column.
+
+Log Severity | Severity number
```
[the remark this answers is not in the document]
mattdurham: This feels a bit strange but assuming its a straight mapping from otel right?

mattdurham (approved):
LGTM

PR #3351, opened 2023-03-23
module.string: fix issue with nested controller API

### LINKED ISSUE #3294: Flow Modules: UI issues with nested modules
There are a few issues I've observed playing with nested modules. This issue may get broken up after review.

# Example river files

There's a redacted url in `parent.river` and you'll need to update some file paths in all example configs to what you have in order to use them.

`parent.river`
```river
logging {
    level  = "error"
	format = "logfmt"
}

local.file "logs" {
    filename  = "/mnt/c/workspace/config/grafana-agent/modules/debug/logs.river"
}

module.string "logs" {
    content = local.file.logs.content
    arguments = {
        url = "REDACTED",
    }
}
```

`logs.river`
```river
argument "url" { }

loki.source.file "local" {
  targets    = [
    {__path__ = "/home/erik/logs/flow.log", "color" = "pink"},
  ]
  forward_to = [loki.write.loki.receiver]
}

loki.write "loki" {
    endpoint {
        url = argument.url.value
    }
}

local.file "logs_2" {
  filename  = "/mnt/c/workspace/config/grafana-agent/modules/debug/logs2.river"
}

module.string "logs_2" {
  content = local.file.logs_2.content
}
```

`logs2.river`
```river
local.file "log_level_2" {
  filename  = "/mnt/c/workspace/config/grafana-agent/modules/debug_level.txt"
}

local.file "logs_3" {
  filename  = "/mnt/c/workspace/config/grafana-agent/modules/debug/logs3.river"
}

module.string "logs_3" {
  content = local.file.logs_3.content
}
```

`logs3.river`
```river
local.file "log_level_3" {
  filename  = "/mnt/c/workspace/config/grafana-agent/modules/debug_level.txt"
}
```

# flow_http.go routing

## Steps

From `http://localhost:12345/component/module.string.logs`

Click view button for module.string.logs_2
image

## Result

`http://localhost:12345/component/module.string.logs/module.string.logs_2` doesn't load
image

## Debugging

I attempted a number of fixes, narrowing in on ./pkg/flow/flow_http.go ComponentHandler. With some print statements, I observed that clicking that view button also called ComponentHandler an unexpected 5 times.
image

I then studied the routes and added a bit of code to get the child module page `http://localhost:12345/component/module.string.logs/module.string.logs_2` to load (still seeing more than expected calls to ComponentHandler)
image

But wait, there's more! Some pages are not loading but I believe there is an href routing issue on the UI side to address first.

# View button href

The module components inside module.string.logs_2 still won't load after the above debugging 'fix'. The fix also didn't help clicking the view button for module.string.logs_3 to go down another level.

Not working!
image

I took a look at the href on each button and spotted an issue that they don't include the full parent path.

href
- `local.file.logs_3`: `http://localhost:12345/component/module.string.logs_2/local.file.logs_3`
- `module.string.logs_3`:`http://localhost:12345/component/module.string.logs_2/module.string.logs_3`

I attempted the following quick fix for the hrefs but the property I was relying on is undefined where I tried to use it:

ComponentView.tsx
Image

updated href: `http://localhost:12345/component/undefined/module.string.logs/module.string.logs_2`

# api.go NewFlowAPI unused argument

This is a nonfunctional item. I noticed that api.go NewFlowAPI asks for an argument it doesn't use. I believe we can remove that.

./web/api/api.go
image

rfratto:
This commit fixes a bug where API endpoints for nested controllers did not propagate correctly. The controller handler assumed that the full HTTP path was being used rather than the truncated path given to component handlers.

This also fixes an issue in the UI where the URL for deeply nested modules was calculated incorrectly.

Fixes #3294

### DIFF
--- a/component/module/string/string.go
+++ b/component/module/string/string.go
@@ -4,6 +4,7 @@ package string
 import (
 	"context"
 	"net/http"
+	"path"
 
 	"github.com/go-kit/log"
 	"github.com/gorilla/mux"
@@ -114,6 +115,13 @@ func (c *Component) Handler() http.Handler {
 	fa := api.NewFlowAPI(c.ctrl, r)
 	fa.RegisterRoutes("/", r)
 
-	r.PathPrefix("/{id}/").Handler(c.ctrl.ComponentHandler())
+	r.PathPrefix("/{id}/").HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Re-add the full path to ensure that nested controllers propagate
+		// requests properly.
+		r.URL.Path = path.Join(c.opts.HTTPPath, r.URL.Path)
+
+		c.ctrl.ComponentHandler().ServeHTTP(w, r)
+	})
+
 	return r
 }
--- a/web/ui/src/features/component/ComponentView.tsx
+++ b/web/ui/src/features/component/ComponentView.tsx
@@ -1,7 +1,7 @@
-import { faCubes, faLink } from '@fortawesome/free-solid-svg-icons';
-import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { FC, Fragment, ReactElement } from 'react';
 import { Link } from 'react-router-dom';
+import { faCubes, faLink } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
 import { RiverValue } from '../../features/river-js/RiverValue';
 import { AttrStmt, Body, StmtType } from '../../features/river-js/types';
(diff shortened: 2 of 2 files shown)


component/module/string/string.go
```diff
@@ -114,6 +115,13 @@ func (c *Component) Handler() http.Handler {
...
 	fa.RegisterRoutes("/", r)
 
-	r.PathPrefix("/{id}/").Handler(c.ctrl.ComponentHandler())
+	r.PathPrefix("/{id}/").HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Re-add the full path to ensure that nested controllers propagate
+		// requests properly.
+		r.URL.Path = path.Join(c.opts.HTTPPath, r.URL.Path)
+
+		c.ctrl.ComponentHandler().ServeHTTP(w, r)
+	})
```
rfratto, line 124: @erikbaranowski Until there's shared code for abstracting away some of the repetitive work behind using a controller in a component, you'll need to mirror this change to #3338

erikbaranowski (approved):
LGTM, thank you for fixing this 1🚀

PR #3369, opened 2023-03-27
Introduce Grafana Agent Flow system packages

### LINKED ISSUE #3303: Proposal for additional package that includes configurations for running Grafana Agent in Flow mode for debian-based systems.
Hello world!
I would like to propose a package for **debian-based** systems (that can be tweaked for other OSes such as rpm) that I made earlier today that could be useful towards the adoption of Grafana agent's flow operational mode. This is an additional/alternative package that includes various configurations and systemd service unit for running Grafana agent in flow mode either side-by-side to standard Grafana agent or to allow switching solely to Grafana agent flow. 

This package that I made (which resides in this [repo I made](https://github.com/robigan/grafana-agent-flow)) I think is the best way to make it as easy as possible to adopt Grafana agent flow without much additional maintenance as it includes the systemd service unit for running Grafana agent flow which just makes it easier to switch over to flow without having to mess with systemd service unit files and the associated `EnvironmentFile` required by the systemd service unit. Also, given that the approach I am suggesting simply creates an additional service unit and configuration directory, it makes adopting flow an incremental process without having to ship two separate packages that both include the full grafana-agent binary. This is less significant, but I've seen various developers ship separate packages with essentially the same binary when phasing one codebase out and another in.

Could this be useful? I made this because I needed an easy way to run grafana-agent in flow mode which I could ship in a package to my various machines which had Grafana agent installed which I wished to switch over to flow mode.

**Quick links**
Repo: [https://github.com/robigan/grafana-agent-flow](https://github.com/robigan/grafana-agent-flow)

rfratto:
> **NOTE TO REVIEWER**: This PR is easier to review commit-by-commit.

This PR adds system packages for Grafana Agent Flow. It is installed with the same `grafana-agent` group so permissions can be shared between the two installations.

The system packages are primarily copies of the existing system packages, but with `-flow` suffixes added where appropriate. 

Closes #3303.


rfratto:
This is ready for review. I tested the new package on my Raspberry Pi (with it being installed alongside the static mode package), and everything works as expected.

packaging/grafana-agent-flow/grafana-agent-flow.river
```diff
@@ -0,0 +1,27 @@
...
+		[{
+			// Self-collect metrics
+			job         = "agent",
+			__address__ = "127.0.0.1:12345",
+		}],
+	)
+
+	forward_to = [
+	// TODO: components to forward metrics to (like prometheus.remote_write or
+	// prometheus.relabel).
```
rfratto, line 25: This is what the formatter is doing, even though I don't think this is where the comments should go :disappointed:

packaging/grafana-agent-flow/linux_packages_test.go
```diff
@@ -0,0 +1,267 @@
...
+		Size:     fi.Size(),
+		ModTime:  fi.ModTime(),
+		Mode:     0600,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	_, err = io.Copy(w, f)
+	return &buf, err
```
ptodev, line 248: if there was an error, `buf` may be in a corrupted state. Is it ok to still return it, along with the error? Is this an ok practice in general? Or is it better to return `nil, err`?

packaging/grafana-agent-flow/linux_packages_test.go
```diff
@@ -0,0 +1,267 @@
...
+// DEBEnvironment creates an Environment to install the agent RPM against.
+func DEBEnvironment(t *testing.T, pool *dockertest.Pool) Environment {
+	t.Helper()
+
+	container := environmentContainer(
+		t,
+		pool,
+		"../testdata/debian-systemd.Dockerfile",
+		"agent-test-debian-systemd",
+		fmt.Sprintf("../../dist/grafana-agent-flow-0.0.0-1.%s.deb", runtime.GOARCH),
```
ptodev, line 166: Same here - we are assuming that the current working directory is the one of the _test.go file?

tools/make/packaging.mk
```diff
@@ -150,106 +150,181 @@ dist.temp/grafana-agent-flow-linux-amd64: GO_TAGS += builtinassets promtail_jour
...
+endef
+
+AGENT_PACKAGE_VERSION := $(patsubst v%,%,$(VERSION))
+AGENT_PACKAGE_RELEASE := 1
+AGENT_PACKAGE_PREFIX  := dist/grafana-agent-$(AGENT_PACKAGE_VERSION)-$(AGENT_PACKAGE_RELEASE)
+
+.PHONY: dist-agent-packages
+dist-agent-packages: dist-agent-packages-amd64   \
+                     dist-agent-packages-arm64   \
+                     dist-agent-packages-arm64   \
```
ptodev, line 215: `dist-agent-packages-arm64` is listed twice?

tools/make/packaging.mk
```diff
@@ -150,106 +150,181 @@ dist.temp/grafana-agent-flow-linux-amd64: GO_TAGS += builtinassets promtail_jour
...
-.PHONY: dist-packages
-dist-packages: dist-packages-amd64   \
-               dist-packages-arm64   \
-               dist-packages-arm64   \
-               dist-packages-ppc64le \
-               dist-packages-s390x
+.PHONY: dist-agent-flow-packages
+dist-agent-flow-packages: dist-agent-flow-packages-amd64   \
+                          dist-agent-flow-packages-arm64   \
+                          dist-agent-flow-packages-arm64   \
```
ptodev, line 290: Here `dist-agent-flow-packages-arm64` is also repeated twice

.drone/drone.yml
```diff
@@ -1246,6 +1246,7 @@ steps:
 - commands:
   - DOCKER_OPTS="" make dist/grafana-agent-linux-amd64
   - DOCKER_OPTS="" make dist/grafana-agentctl-linux-amd64
+  - DOCKER_OPTS="" make dist.temp/grafana-agent-flow-linux-amd64
```
ptodev, line 1249: I don't know if `dist.flow` would make more sense than `dist.temp`?

ptodev (approved):

PR #3371, opened 2023-03-27
flow: prototype retrieving flags from config file

rfratto:
This commit adds a new `--flags-config-file` flag to the `run` command. When this flag is passed and given a file, the file is parsed as a River config representing the flags to pass to run.

For example, the default flags `--server.http.listen-addr=127.0.0.1:12345 --server.http.ui-prefix=/ --storage-path=data-agent/ --disable-reporting=false` is represented as the following file:

```
server {
  http {
    listen_addr = "127.0.0.1:12345"
    ui_prefix   = "/"
  }
}

storage {
  path = "data-agent/" 
}

disable_reporting = false
```

This can be useful for systems where it is not easy to change options passed to the service (like macOS; see grafana/homebrew-grafana#47).

It's not clear to me if this is a good idea, but it would genuinely help make things easier for macOS installations. 

It's also not clear to me if the flags config file should _also_ be River. Maybe it should be another language, like YAML or .ini?

### DIFF
--- a/cmd/internal/flowmode/cmd_run.go
+++ b/cmd/internal/flowmode/cmd_run.go
@@ -24,6 +24,7 @@ import (
 	"github.com/grafana/agent/pkg/flow"
 	"github.com/grafana/agent/pkg/flow/logging"
 	"github.com/grafana/agent/pkg/flow/tracing"
+	"github.com/grafana/agent/pkg/river"
 	"github.com/grafana/agent/pkg/river/diag"
 	"github.com/grafana/agent/pkg/usagestats"
 	"github.com/prometheus/client_golang/prometheus"
@@ -37,10 +38,16 @@ import (
 
 func runCommand() *cobra.Command {
 	r := &flowRun{
-		httpListenAddr:   "127.0.0.1:12345",
-		storagePath:      "data-agent/",
-		uiPrefix:         "/",
-		disableReporting: false,
+		Server: flowRunServer{
+			HTTP: flowRunServerHTTP{
+				ListenAddr: "127.0.0.1:12345",
+				UIPrefix:   "/",
+			},
+		},
+		Storage: flowRunStorage{
+			Path: "data-agent/",
+		},
+		DisableReporting: false,
 	}
 
 	cmd := &cobra.Command{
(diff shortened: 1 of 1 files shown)


rfratto:
My main concern with re-using River for this file is that people might confuse the two River files, since the flags config file has a different schema than the Flow config file, and users might confuse the two.

rfratto:
cc @rlankfo since you had reviewed grafana/homebrew-grafana#47.

ptodev:
I personally think that using River only makes sense if these settings are part of the normal River config file. E.g. there could be a singleton component-looking section at the top which sets these things. However, it's not clear how this could work if the configuration should also be used by static mode.

River makes a lot of sense for piping components, but I am not sure what is the benefit in using it here? I didn't think it was meant to be a general purpose configuration language. And I also agree it'd be confusing to use it for other purposes.

I think we already have too many ways to configure the Agent:

- environment variables
- cmd args
- river/yaml config files

This could be confusing - it is sometimes not clear which of the three needs to be changed when a user want to reconfigure something. And if we add another mechanism which overlaps with the current ones, then we also need to warn users about some configs taking precedence over others.

In terms of fixing the immediate Homebrew issue - is it not possible for Homebrew service to start an .sh file instead of the Agent directly, and then users can just add their env var settings to the .sh file?

ptodev:
> The problem is Homebrew will override things from packages it installs.

Could we update the official Homebrew recipe to start a .sh file instead of the agent directly? Or maybe to start a bash command line which runs an sh file and (`&&`) then the agent binary, thereby allowing the .sh file to customize the Agent environment?

> It wasn't a design goal, but we always assumed that we would eventually open source River and the Flow system in a new repository eventually so other people could use it.

Yes, but my question was more about the intended use case. I always thought of River as a domain-specific language, created to solve a specific problem. I didn't think of it as a replacement for general purpose config formats such as yaml, json and xml.

rfratto:
I'm going to close this for now. Homebrew's [recommendation](https://github.com/Homebrew/brew/blob/master/docs/FAQ.md#how-can-i-specify-different-configure-arguments-for-a-formula) is to do `brew edit FORMULA` to edit formulas prior to installation, so the steps for a user would be:

1. `brew edit grafana-agent-flow` 
2. Add custom flags as required 
3. `brew reinstall grafana-agent-flow` 
4. `brew services restart grafana-agent-flow`

IMO this isn't ideal, but it's much less abrasive than the alternative of forking the launchd service.

PR #3377, opened 2023-03-28
add ability to have a module with multiple exports

### LINKED ISSUE #3335: Flow: module exports override each other
There's a regression where each ExportConfigNode will be calling the same onExportsChange function and fighting with each other. This means that only one export at a time would be exposed to the module loader. 

_Originally posted by @rfratto in https://github.com/grafana/agent/pull/3297#discussion_r1145250601_

mattdurham:

Closes #3335

#### Notes to the Reviewer

Wondering why the exports on changed returns a map???

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/component/module/string/string.go
+++ b/component/module/string/string.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"net/http"
 	"path"
+	"sync"
 
 	"github.com/go-kit/log"
 	"github.com/gorilla/mux"
@@ -50,6 +51,9 @@ type Component struct {
 	opts component.Options
 	log  log.Logger
 	ctrl *flow.Flow
+
+	exportsMut sync.Mutex
+	exports    map[string]any
 }
 
 var (
@@ -65,24 +69,36 @@ func New(o component.Options, args Arguments) (*Component, error) {
 	flowRegistry := prometheus.NewRegistry()
 
 	c := &Component{
-		opts: o,
-		log:  o.Logger,
-
-		ctrl: flow.New(flow.Options{
-			ControllerID: o.ID,
-			LogSink:      logging.LoggerSink(o.Logger),
-			Tracer:       flowTracer,
-			Reg:          flowRegistry,
-
-			DataPath:       o.DataPath,
-			HTTPPathPrefix: o.HTTPPath,
-			HTTPListenAddr: o.HTTPListenAddr,
-
-			OnExportsChange: func(exports map[string]any) {
-				o.OnStateChange(Exports{Exports: exports})
-			},
-		}),
+		opts:    o,
+		log:     o.Logger,
+		exports: make(map[string]any),
 	}
+	c.ctrl = flow.New(flow.Options{
+		ControllerID: o.ID,
+		LogSink:      logging.LoggerSink(o.Logger),
+		Tracer:       flowTracer,
+		Reg:          flowRegistry,
+
+		DataPath:       o.DataPath,
+		HTTPPathPrefix: o.HTTPPath,
+		HTTPListenAddr: o.HTTPListenAddr,
+
+		OnExportsChange: func(exports map[string]any) {
+			c.exportsMut.Lock()
+			defer c.exportsMut.Unlock()
+
+			// Update our primary export map with all the values.
+			for k, v := range exports {
+				c.exports[k] = v
+			}
+			// This is to prevent any access to our primary export map. So we create a copy for other usages.
+			exportable := make(map[string]any)
+			for k, v := range c.exports {
+				exportable[k] = v
+			}
+			o.OnStateChange(Exports{Exports: exportable})
+		},
+	})
 
 	if err := c.Update(args); err != nil {
 		return nil, err
@@ -104,6 +120,10 @@ func (c *Component) Update(args component.Arguments) error {
 	if err != nil {
 		return err
 	}
+	// Unsure of our exports so we should refresh those.
+	c.exportsMut.Lock()
+	c.exports = make(map[string]any)
+	c.exportsMut.Unlock()
 
 	return c.ctrl.LoadFile(f, newArgs.Arguments)
 }


component/module/string/string.go
```diff
@@ -65,24 +69,36 @@ func New(o component.Options, args Arguments) (*Component, error) {
...
+		ControllerID: o.ID,
+		LogSink:      logging.LoggerSink(o.Logger),
+		Tracer:       flowTracer,
+		Reg:          flowRegistry,
+
+		DataPath:       o.DataPath,
+		HTTPPathPrefix: o.HTTPPath,
+		HTTPListenAddr: o.HTTPListenAddr,
+
+		OnExportsChange: func(exports map[string]any) {
```
mattdurham, line 86: Wondering if we can change this signature to string, any instead of map.

rfratto, line 86: This won't work if an export gets removed, because the component will keep it in cache forever. 

The reason OnExportsChange provides a map because it's intended to be the declarative state of all the currently defined exports in the module, with the OnExportsChange callback being used as a thin layer to expose it back to the caller; I think fixing this at the component level is too high-level of a fix.

pkg/flow/internal/controller/loader.go
```diff
@@ -392,6 +402,10 @@ func (l *Loader) EvaluateDependencies(parentScope *vm.Scope, c *ComponentNode) {
 		switch n := n.(type) {
 		case BlockNode:
 			err = l.evaluate(logger, parentScope, n)
+			if exp, ok := n.(*ExportConfigNode); ok {
+				name, val := exp.NameAndValue()
+				exports[name] = val
+			}
```
rfratto, line 408: Doesn't this suffer the same problem as before, where partial evaluates of a block node will cause other exports to get dropped? 

IMO we should probably be using a mechanism similar to `valueCache` (but for exports) here, where the cache gets cleared before a full evaluation, but updated in-place and re-exported any time a partial evaluation happens.

component/module/string/string.go
```diff
@@ -67,22 +67,21 @@ func New(o component.Options, args Arguments) (*Component, error) {
...
-			DataPath:       o.DataPath,
-			HTTPPathPrefix: o.HTTPPath,
-			HTTPListenAddr: o.HTTPListenAddr,
-
-			OnExportsChange: func(exports map[string]any) {
-				o.OnStateChange(Exports{Exports: exports})
-			},
-		}),
 	}
+	c.ctrl = flow.New(flow.Options{
```
erikbaranowski, line 71: does this change functionally do anything or just a styling change on initializing the flow controller?

erikbaranowski:
I think we mentioned cleaning up this comment on this PR so we don't forget https://github.com/grafana/agent/pull/3297#discussion_r1145239603 (copy/paste mistake we missed until right after merge)

pkg/flow/internal/controller/config_export.go
```diff
@@ -62,19 +62,22 @@ type exportBlock struct {
...
 	cn.mut.RLock()
 	defer cn.mut.RUnlock()
-	exports := make(map[string]any)
 
 	var export exportBlock
 	if err := cn.eval.Evaluate(scope, &export); err != nil {
 		return fmt.Errorf("decoding River: %w", err)
 	}
+	cn.value = export.Value
+	cn.name = cn.label
```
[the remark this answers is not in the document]
mattdurham: Doh, good catch.

pkg/flow/internal/controller/loader.go
erikbaranowski: I'm thinking how to recreate what you are worried on here to see how it works. I created a 2 export module loaded as `module.string` with content from a `local.file`. I then updated the value of 1 of the 2 exports and waited for `local.file` polling to reload the module. It seems to pick up the change and I can repeat the process numerous times.

Are we in a situation where since reload isn't technically what reloads the module contents in any way I can't recreate this yet? Or does this show it working as intended? Or maybe I'm missing the test case?

pkg/flow/internal/controller/loader.go
```diff
@@ -392,6 +402,10 @@ func (l *Loader) EvaluateDependencies(parentScope *vm.Scope, c *ComponentNode) {
 		switch n := n.(type) {
 		case BlockNode:
 			err = l.evaluate(logger, parentScope, n)
+			if exp, ok := n.(*ExportConfigNode); ok {
+				name, val := exp.NameAndValue()
+				exports[name] = val
```
erikbaranowski, line 407: I thought for a moment that we should validate if the key is already here but we have an upstream check on duplicate blocks that will catch this. Mentioning in case anyone thinks similar.

Example error:
image

pkg/flow/internal/controller/loader.go
mattdurham: Are you saying a situation where there is a failure in re-evaluating a module and the walk exits early? 

I think even using `valueCache` style behavior there will be issues if the reload fails. 

IMO until we define the failure mode of modules, failures during module parsing should be considered undefined behavior. For what its worth, I would prefer to cancel the whole load and reload from the last known good state for the WHOLE config in the case of any error.

rfratto: This should be able to reproduce the scenario:

```river 
discovery.kubernetes "nodes" {
  role = "node" 
}

discovery.kubernetes "pods" {
  role = "pod" 
}

export "nodes" {
  value = discovery.kubernetes.nodes.targets 
}

export "pods" {
  value = discovery.kubernetes.pods.targets 
}
```

On the initial load, both `export.nodes` and `export.pods` will be exposed to the parent module, but if either `discovery.kubernetes.nodes` or `discovery.kubernetes.pods` updates, it will trigger a partial reevaluation of the graph, and cause the other export to disappear.

mattdurham: Apologies, misunderstood the situation you were describing, looking into this now.

mattdurham: Re-used the `valueCache` since it generally mirrors the usage. Going to work on a unit test for this.

pkg/flow/internal/controller/loader.go
```diff
@@ -403,9 +415,13 @@ func (l *Loader) EvaluateDependencies(parentScope *vm.Scope, c *ComponentNode) {
 		}
 		return nil
 	})
+
+	if l.globals.OnExportsChange != nil {
+		l.globals.OnExportsChange(l.cache.CreateModuleExports())
+	}
```
rfratto, line 421: Can we do anything to detect whether a call to this is necessary to avoid stress on the parent component? 

If we can't do anything easily, a TODO comment to handle it later would be appreciated.

rfratto (approved):
LGTM, thanks!

erikbaranowski (approved):

pkg/flow/internal/controller/loader.go
mattdurham: I have a solution but not sure if I like it so will commit this PR and open another PR with the change.

PR #3378, opened 2023-03-28
[Agent Management] GetRemoteConfigCache now falls back to DefaultConfig

spartan0x117:
#### PR Description
Now adds a second fallback to `DefaultConfig()` in case both the call to the API fails and there is no local cache. This is done to ensure that the agent will continue to run, even if the very first attempt is not successful.

In order to make sure that this can be diagnosed more easily, also adds logging to the cache retrieval section as well as a new counter metric `agent_management_fallbacks_total{destination="default|cache"}`.

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [X] Tests updated


jcreixell (commented):
This LGTM overall, the only thing I am wondering about is whether we want to introduce a new metric so that we can alert when this happens (it is not the same to fallback to cache from time to time due to connectivity issues / outages, which would show in the fetch-related metrics, vs. falling back to default config, which indicates a bigger problem)

pkg/config/instrumentation/agent_management_metrics.go
```diff
@@ -0,0 +1,38 @@
...
+func initializeagentManagementMetrics() {
+	amMetrics = newAgentManagementMetrics()
+}
+
+func newAgentManagementMetrics() *agentManagementMetrics {
+	var agentManagementMetrics agentManagementMetrics
+
+	agentManagementMetrics.fallbacks = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "agent_management_fallbacks_total",
```
jcreixell, line 26: nit: rename to agent_management_config_fallbacks_total

jcreixell (approved):
pushed a small renaming change, other than that LGTM! :ok_hand:

pkg/config/instrumentation/agent_management_metrics.go
```diff
@@ -0,0 +1,38 @@
...
+
+func newAgentManagementMetrics() *agentManagementMetrics {
+	var agentManagementMetrics agentManagementMetrics
+
+	agentManagementMetrics.configFallbacks = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "agent_management_config_fallbacks_total",
+			Help: "Number of config fallbacks by fallback destination.",
+		},
+		[]string{"destination"},
```
spartan0x117, line 29: Thinking on it further, does this sound better?
```suggestion
			Help: "Number of config fallbacks by fallback source.",
		},
		[]string{"source"},
```

jcreixell: yeah I think it does

pkg/config/agentmanagement.go
```diff
@@ -191,11 +191,16 @@ func getRemoteConfig(expandEnvVars bool, configProvider remoteConfigProvider, lo
...
 
-func getCachedRemoteConfig(expandEnvVars bool, configProvider remoteConfigProvider, fs *flag.FlagSet) (*Config, error) {
+// getCachedRemoteConfig gets the cached remote config, falling back to the default config if the cache is invalid or not found.
+func getCachedRemoteConfig(expandEnvVars bool, configProvider remoteConfigProvider, fs *flag.FlagSet, log *server.Logger) (*Config, error) {
 	rc, err := configProvider.GetCachedRemoteConfig()
 	if err != nil {
-		return nil, fmt.Errorf("could not load cached config: %w", err)
+		level.Error(log).Log("msg", "could not get cached remote config, falling back to DefaultConfig", "err", err)
+		d := DefaultConfig()
+		instrumentation.InstrumentAgentManagementConfigFallback("default")
```
jcreixell, line 200: would it make sense to call this "empty_config"? not sure "default" is understood as "the agent does not do any work except trying to reconnect"

pkg/config/instrumentation/agent_management_metrics.go
```diff
@@ -0,0 +1,38 @@
...
+
+func newAgentManagementMetrics() *agentManagementMetrics {
+	var agentManagementMetrics agentManagementMetrics
+
+	agentManagementMetrics.configFallbacks = promauto.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "agent_management_config_fallbacks_total",
+			Help: "Number of config fallbacks by fallback source.",
+		},
+		[]string{"source"},
```
jcreixell, line 29: actually, what about just calling this "to"?

agent_management_config_fallbacks_total{to: "cache"}
agent_management_config_fallbacks_total{to: "empty_config"}

jcreixell (commented):
added some more naming ideas

PR #3388, opened 2023-03-29
component/loki: fix journaltarget after #3375

rfratto:
PR #3375 introduced an issue in journaltarget that was missed by CI.


spartan0x117 (approved):
LGTM! Just curious is the `promtail_journal_enabled` only available on certain OSs & architectures (seems like `amd64`, `arm64` non-darwin linux, but not macOS or Windows)

PR #3390, opened 2023-03-30
Use Defined types in client arguments

dosmanak:
use types as defined as https://grafana.com/docs/agent/latest/flow/config-language/expressions/types_and_values/

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/flow/reference/components/otelcol.exporter.otlphttp.md
+++ b/docs/sources/flow/reference/components/otelcol.exporter.otlphttp.md
@@ -69,13 +69,13 @@ Name | Type | Description | Default | Required
 `endpoint`           | `string`      | `host:port` to send telemetry data to. | | yes
 `read_buffer_size`   | `string`      | Size of the read buffer the HTTP client uses for reading server responses. | `0` | no
 `write_buffer_size`  | `string`      | Size of the write buffer the HTTP client uses for writing requests. | `"512KiB"` | no
-`timeout`            | `duration`    | Time to wait before marking a request as failed. | `"30s"` | no
-`headers`            | `map(string)` | Additional headers to send with the request. | `{}` | no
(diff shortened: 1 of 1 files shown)


tpaschalis (changes requested):

dosmanak:
Thanks for ultra fast reaction.
I was not sure how the type `map(string)` shall be formatted.
I decided to go with
```
  client {                                                                                              
    endpoint = env("TEMPO_URL")                                                                         
    auth = otelcol.auth.basic.tempocreds.handler                                                        
    headers = {                                                                                           
      "X-Scope-OrgID" = env("TEMPO_TENANT_ID")                                                          
    }                                                                                                   
  }        
```

tpaschalis:
> I was not sure how the type map(string) shall be formatted.

The idea here was that map keys are always strings, so the `map(T)` describes what `T` type the values should take, so in this case it's a `string -> string` map.

If you don't mind, I'm closing this PR. We'll keep an eye out for things we can improve in the documentation; feel free to propose such documentation changes via another PR or an issue.

Thanks again!

rfratto:
@tpaschalis Should we document our type naming convention somewhere so people know what "duration" means?

tpaschalis:
> @tpaschalis Should we document our type naming convention somewhere so people know what "duration" means?
Yes. I think the "Types and Values" page may be a good candidate, under "Special Types". WDYT?

rfratto:
> Yes. I think the "Types and Values" page may be a good candidate, under "Special Types". WDYT?

SGTM, but given the above comment I think we should also have a table or something somewhere which mentions every type used in the docs:

`boolean`
`number`
`string`
`map(T)` 
`array(T)` 
`duration` 
`capsule(T)`

This also might be a good opportunity to reconsider if `map(T)` should be `object(T)`.

PR #3402, opened 2023-03-31
Fix deadlock in loki.source.docker with a separate mutex

### LINKED ISSUE #3391: Flow: Deadlock when updating loki.source.docker
`loki.source.docker` can hit a race condition when updating and removing old targets; while old targets are shutting down, they will try to flush their logs, which will block forever because the mutex is held by the call to Update which is shutting down those targets. 

This is a similar bug to what was fixed in #3298. We should check that no other loki.source components are susceptible to this deadlock.

tpaschalis:
#### PR Description
This PR fixes the loki.source.docker deadlock by adding a separate mutex for updating the LogsReceiver fanout and allowing containers to flush their log lines before being shut down.

Fixes #3391

#### PR Checklist

- [X] CHANGELOG updated
- [ ] Tests updated (N/A? will look if there's a way of emulating docker containers and makes sense)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,6 +91,8 @@ Main (unreleased)
 - Fix issue where scraping native Prometheus histograms would leak memory.
   (@rfratto)
 
+- Fix issue where loki.source.docker component could deadlock. (@tpaschalis)
+
 ### Other changes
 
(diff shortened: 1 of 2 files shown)


rfratto (approved):
LGTM!

ptodev (approved):

PR #3406, opened 2023-03-31
Add default logging and tracing nodes back into the dag when not provided in the river config

### LINKED ISSUE #3334: Flow: log level can persist after removing logging block
#3259 introduced a rewrite of the logging subsystem. It introduced a bug where the last configured logging level persists after removing the logging block, because the logger settings only get configured when the block is provided. 

This bug is only reproducible against the main branch.

To reproduce, start with the following config file:

```river 
logging { level = "debug" }
```

And then reload to this (empty) config file:

```river 
```

Notice that debug logs will still be emitted.

erikbaranowski:

Fixes #3334

#### Notes to the Reviewer

This can be done without the opts caching that was added to sink.go but it seems nicer to not mess with the logger if nothing changed. I also considered if this should be done here or have the caller to Update track the state but I think this is much simpler and improves all consumers of the sink functionality.

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/flow/internal/controller/loader.go
+++ b/pkg/flow/internal/controller/loader.go
@@ -219,7 +219,13 @@ func (l *Loader) populateConfigBlockNodes(g *dag.Graph, configBlocks []*ast.Bloc
 
 	// If a logging config block is not provided, use the defaults.
 	if _, ok := blockMap[loggingBlockID]; !ok && !l.isModule() {
-		l.globals.LogSink.Update(logging.DefaultSinkOptions)
+		err := l.globals.LogSink.Update(logging.DefaultSinkOptions)
+		if err != nil {
+			diags.Add(diag.Diagnostic{
+				Severity: diag.SeverityLevelError,
+				Message:  err.Error(),
+			})
+		}
 	}
 
 	return diags
--- a/pkg/flow/internal/controller/loader.go
+++ b/pkg/flow/internal/controller/loader.go
@@ -217,6 +217,11 @@ func (l *Loader) populateConfigBlockNodes(g *dag.Graph, configBlocks []*ast.Bloc
 		g.Add(c)
 	}
 
+	// If a logging config block is not provided, use the defaults.
+	if _, ok := blockMap[loggingBlockID]; !ok && !l.isModule() {
+		l.globals.LogSink.Update(logging.DefaultSinkOptions)
+	}
+
 	return diags
 }
 
--- a/pkg/flow/internal/controller/loader.go
+++ b/pkg/flow/internal/controller/loader.go
@@ -225,15 +225,16 @@ func (l *Loader) populateConfigBlockNodes(g *dag.Graph, configBlocks []*ast.Bloc
 		g.Add(c)
 	}
 
-	// If a logging config block is not provided, use the defaults.
+	// If a logging config block is not provided, we create an empty node which uses defaults.
 	if _, ok := blockMap[loggingBlockID]; !ok && !l.isModule() {
-		err := l.globals.LogSink.Update(logging.DefaultSinkOptions)
-		if err != nil {
-			diags.Add(diag.Diagnostic{
-				Severity: diag.SeverityLevelError,
-				Message:  err.Error(),
-			})
-		}
+		c := NewDefaultLoggingConfigNode(l.globals)
+		g.Add(c)
+	}
+
+	// If a tracing config block is not provided, we create an empty node which uses defaults.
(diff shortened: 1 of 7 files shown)


pkg/flow/internal/controller/loader.go
```diff
@@ -217,6 +217,17 @@ func (l *Loader) populateConfigBlockNodes(g *dag.Graph, configBlocks []*ast.Bloc
 		g.Add(c)
 	}
 
+	// If a logging config block is not provided, use the defaults.
```
rfratto, line 220: This needs to be done for he tracing block too. 

Also, I think this is OK for now but I'm worried about how maintainable it is. Can we add a TODO to try to identify a solution which we're less likely to accidentally break?

rfratto: (Longer term, I think the logging/tracing config nodes always exist in the DAG, even when the block isn't provided in the config file. That way we establish a pattern of having config nodes which always exist and always have some kind of evaluation logic)

rfratto: > I'm open to adding it to the DAG if it's missing instead if we prefer that approach. 

IMO it'll be a little more code but I think it's preferable just from the perspective of keeping logging specific code isolated, and not having the loader be aware of specific subsystems. 

> What do you think about the opts caching in the logger?

I don't really have an opinion about that one, I think it's fine 👍

erikbaranowski: So I started implementing this and I may be going down the wrong path or an unexpected rabbit hole.

I think a couple things are true
- We need to do something in `loader.go` `populateConfigBlockNodes` to add the node after it has finished adding other nodes so we aren't totally free of custom logic for this type of situation in the loader.
- Creating a block without real values on it for NamePos, LabelPos, etc feels kind of wrong.
- I'm concerned what will happen if we walk the graph on a fake node how it will behave but maybe it's ok


Prototype I was working on...

loader.go populateConfigBlockNodes replace change in this PR with...
```golang
        If a logging config block is not provided add it to the graph.
	if _, ok := blockMap[loggingBlockID]; !ok && !l.isModule() {
		c, newConfigNodeDiags := NewDefaultLoggingConfigNode(l.globals, l.isModule())
		diags = append(diags, newConfigNodeDiags...)
		g.Add(c)
	}
```

config_logging.go
```golang
func NewDefaultLoggingConfigNode(globals ComponentGlobals, isInModule bool) (*LoggingConfigNode, diag.Diagnostics) {
        // Fill all this in with logging defaults
        block := &ast.BlockStmt{
		Name:      []string{},
		NamePos:   token.Pos{},
		Label:     "",
		LabelPos:  token.Pos{},
		Body:      []ast.Stmt{},
		LCurlyPos: token.Pos{},
		RCurlyPos: token.Pos{},
	}

	return NewLoggingConfigNode(block, globals, isInModule)
}
```

Were you thinking of something different or more drastic than this approach? I can keep thinking on it too.

erikbaranowski: I could make the case that populateConfigBlockNodes belongs in config.go and populateComponentNodes in component.go too (pass in the 2 things we need from `l *Loader`). Thoughts?

mattdurham (approved):
LGTM

pkg/flow/internal/controller/loader.go
erikbaranowski: we talked offline - I'm going to implement with a nil block on the logging node and see what it looks like

erikbaranowski: I added both logging and tracing back into the dag while I was at it. This should make it easier to similarly propagate the tracing for modules @mattdurham

pkg/river/ast/ast.go
```diff
@@ -221,7 +222,7 @@ func (n *ParenExpr) astExpr()      {}
 
 // StartPos returns the position of the first character belonging to a Node.
 func StartPos(n Node) token.Pos {
-	if n == nil {
+	if n == nil || reflect.ValueOf(n).IsZero() {
```
erikbaranowski, line 225: this was tested by forcing config_logging.go or config_tracing.go `Evaluate` to return an error instead of nil at the end. Without this change we get a go panic instead of a node evaluation error.

without the change
image

with the change!
image

pkg/flow/internal/controller/loader.go
erikbaranowski: This should be all set pending another review of the specific implementation details

rfratto (approved):

PR #3407, opened 2023-03-31
Disable git user warning on build container

rfratto:
Disables a warning by git to add /src or /drone/src as a safe directory. This was added as a security measure on systems, but it doesn't make sense in build containers which always use the root user.


erikbaranowski (approved):
LGTM

PR #3410, opened 2023-04-02
Fix typos, wording and grammar

darxriggs:


component/loki/process/internal/metric/counters.go
```diff
@@ -63,7 +63,7 @@ func (c *CounterConfig) UnmarshalRiver(f func(v interface{}) error) error {
 	return nil
 }
 
-// Counters is a vec tor of counters for a each log stream.
+// Counters is a vector of counters for a log stream.
```
darxriggs, line 66: @tpaschalis I am not 100% certain. Did you mean `a log stream` or `each log stream`?

component/discovery/relabel/relabel.go
```diff
@@ -28,7 +28,7 @@ type Arguments struct {
 	// Targets contains the input 'targets' passed by a service discovery component.
 	Targets []discovery.Target `river:"targets,attr"`
 
-	// The relabelling rules to apply to the each target's label set.
+	// The relabelling rules to apply to each target's label set.
```
darxriggs, line 31: @tpaschalis I am not 100% certain. Did you mean `the target's label set` or `each target's label set`?

marctc (approved):
Thanks for this :heart_eyes:! Did you use any tool to automatize this? I wonder if we should have some sort linter enabled to catch those.

GOVERNANCE.md
```diff
@@ -145,7 +145,7 @@ The ex-member is
 - removed from the list of [team members](#team-members). Ideally by sending a PR of their own, at least approving said PR. In case of forced removal, no approval is needed.
 - removed from the projects. Optionally, they can retain maintainership of one or more repositories if the [team](#team-members) agrees.
 - removed from the team mailing list and demoted to a normal member of the other mailing lists.
-- not allowed to call themselves an active team member any more, nor allowed to imply this to be the case.
+- not allowed to call themselves an active team member anymore, nor allowed to imply this to be the case.
```
marctc, line 148: This file is shared between different repos, so the change would need to be done in other repos. Could revert this change for now (governance team will care of it eventually)? :pray:

component/loki/process/internal/metric/counters.go
tpaschalis: The proposed change sounds okay (eg. for a given log stream)

pkg/integrations/v2/app_agent_receiver/sourcemaps.go
```diff
@@ -95,15 +95,15 @@ func NewSourceMapStore(l log.Logger, config SourceMapConfig, reg prometheus.Regi
 	metrics := &sourceMapMetrics{
 		cacheSize: prometheus.NewCounterVec(prometheus.CounterOpts{
 			Name: "app_agent_receiver_sourcemap_cache_size",
-			Help: "number of items in sourcemap cache, per origin",
```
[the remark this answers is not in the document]
darxriggs: Both variants (`source map` and `sourcemap`) have been used in this file. I made the spelling at least here consistent.

The decision to use `source map` is based on:
- the fact that in the code mostly variations of `SourceMap` are used to separate both words
- the "official" wording by Mozilla (e.g. see [here](https://developer.chrome.com/blog/sourcemaps/) and [here](https://developer.mozilla.org/en-US/docs/web/http/Headers/Sourcemap))

darxriggs:
How shall I resolve the conflict and incorporate review comments?
Merging `main` into this branch or rebasing onto `main`?

pkg/integrations/v2/app_agent_receiver/sourcemaps.go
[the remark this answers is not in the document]
darxriggs: If `sourcemap` is preferred nonetheless, let me know and I will change it.

darxriggs:
> Did you use any tool to automatize this? I wonder if we should have some sort linter enabled to catch those.

The locations have been located by a tool but every finding then has been checked manually and adapted or excluded.

I am not sure how well this can be automated. I think the tool would have to be really good. This is about natural language combined with code (including a lot of abbreviations, names, etc.). Maintenance could be more expensive than for a code linter (which can understand the syntax of a language).

marctc:
> How shall I resolve the conflict and incorporate review comments? Merging `main` into this branch or rebasing onto `main`?

Whatever your prefer. The changes will be squashed anyway.

darxriggs:
* rebased on `main`
* reverted the change in `GOVERNANCE.md` as suggested
* resolved the conflict in `pkg/flow/internal/controller/value_cache.go`

PR #3427, opened 2023-04-03
Reduce churn by tracking module exports changes

mattdurham:
#### PR Description

This attempts to reduce churn on export changes. The name of the function is horrendous `HasModuleExportsChangedSinceLastCall` but I didnt want to keep around a copy of the old map one but wanted to the name to represent what is going on.

#### PR Checklist

- [X] Tests updated


pkg/flow/internal/controller/loader.go
```diff
@@ -416,7 +416,7 @@ func (l *Loader) EvaluateDependencies(parentScope *vm.Scope, c *ComponentNode) {
 		return nil
 	})
 
-	if l.globals.OnExportsChange != nil {
+	if l.globals.OnExportsChange != nil && l.cache.HasModuleExportsChangedSinceLastCall() {
```
rfratto, line 419: This seems a little error prone since it updates the state of the cache. What do you think about having some kind of `ModuleExportsChangedTime()` method and comparing the changed time before/after the evaluate? 

Alternatively instead of returning the changed time, you can have it keep track of some kind of changed index and compare those instead.

mattdurham: Instead of using a time, switched to using an incrementing index.

pkg/flow/internal/controller/value_cache.go
```diff
@@ -97,18 +97,24 @@ func (vc *valueCache) ClearModuleExports() {
...
+	vc.moduleChangedIndex++
 	vc.moduleExports = make(map[string]any)
 }
 
-// HasModuleExportsChangedSinceLastCall returns true if the module exports changed since the last call to HasModuleExportsChangedSinceLastCall.
-func (vc *valueCache) HasModuleExportsChangedSinceLastCall() bool {
-	vc.mut.Lock()
-	defer vc.mut.Unlock()
+// HasModulesChanged returns true if the module exports index is different from changeIndex.
+func (vc *valueCache) HasModulesChanged(changeIndex int) bool {
```
rfratto, line 105: I'm not sure this function is necessary anymore, the caller can directly 
check `cache.ExportChangeIndex() != lastIndex`

mattdurham: Your correct, will remove

rfratto (approved):
LGTM

PR #3428, opened 2023-04-03
remote.vault: introduce component to retrieve secrets from Vault

rfratto:
The `remote.vault` component retrieves secrets from Vault. The initial implementation is scoped to only retrieving secrets from the KVv2 store, but it should be flexible enough to support logical database stores in the future. 

`remote.vault` supports a few different authentication methods, all from the Vault API module: 

* Raw Token 
* Approle
* AWS
* Azure 
* GCP
* Kubernetes
* LDAP
* Username/password 
* Custom 

Authentication tokens are automatically refreshed in the background if they have a lease or are renewable. 

Currently, the entire set of data retrieved from Vault is treated as a secret, and there is no way to mark a key from the data as non-sensitive. In the future, this may be handled by some kind of function to convert a secret into a string, or a configuration block in the component to list a subset of data field names as non-sensitive. 

When reviewing, note that the prototype for this one started in the extremely early days of Flow (e.g., pre-v0.28). There may be best practices I'm not following, and didn't catch in my self-review. If you notice anything out of place, please let me know.

Sorry about the size of this one 😞


docs/sources/flow/reference/components/remote.vault.md
```diff
@@ -0,0 +1,265 @@
...
+
+## HTTP endpoints
+
+> HTTP endpoints for components are always prefixed with
+> `/component/<component_id>/`.
+
+`remote.vault` exposes the following HTTP endpoints:
+
+* `POST /force-reread`: force a reread of the secret. This endpoint always
+  returns an HTTP 200 response with no response body.
```
rfratto, line 242: Documented but not implemented. Should this be removed or implemented?

docs/sources/flow/reference/components/remote.vault.md
```diff
@@ -0,0 +1,265 @@
...
+    url = "https://onprem-mimir:9009/api/v1/push"
+    basic_auth {
+      username = remote.vault.remote_write.data.username
+      password = remote.vault.remote_write.data.password
+    }
+  }
+}
+```
+
+## Arguments
```
rfratto, line 45: This is missing the Blocks section after Arguments.

docs/sources/flow/reference/components/remote.vault.md
```diff
@@ -0,0 +1,265 @@
...
+
+`remote.vault` connect to Vault to retrieve secrets. It can retrieve secrets
+using the [KV v2][] secrets engine.
+
+Multiple `remote.vault` components can be specified by giving them different
+labels.
+
+[KV v2]: https://www.vaultproject.io/docs/secrets/kv/kv-v2
+
+## Example
```
rfratto, line 17: Examples go at the bottom. Usage should go here.

rfratto (commented):
I prototyped this a long time ago and finally got around to pushing it now. Leaving comments for patterns that have changed in our documentation structure.

component/remote/vault/vault.go
```diff
@@ -0,0 +1,237 @@
...
+func (c *Component) CurrentHealth() component.Health {
+	c.healthMut.RLock()
+	defer c.healthMut.RUnlock()
+	return c.health
+}
+
+// DebugInfo returns debug information about the remote.vault component. It
+// includes non-sensitive metadata about the current secret.
+func (c *Component) DebugInfo() interface{} {
+	panic("NYI")
```
rfratto, line 236: Don't forget to remove this :)

component/remote/vault/auth.go
```diff
@@ -0,0 +1,339 @@
...
+type AuthArguments struct {
+	AuthToken      *AuthToken      `river:"token,block,optional"`
+	AuthAppRole    *AuthAppRole    `river:"approle,block,optional"`
+	AuthAWS        *AuthAWS        `river:"aws,block,optional"`
+	AuthAzure      *AuthAzure      `river:"azure,block,optional"`
+	AuthGCP        *AuthGCP        `river:"gcp,block,optional"`
+	AuthKubernetes *AuthKubernetes `river:"kubernetes,block,optional"`
+	AuthLDAP       *AuthLDAP       `river:"ldap,block,optional"`
+	AuthUserPass   *AuthUserPass   `river:"userpass,block,optional"`
+	AuthCustom     *AuthCustom     `river:"auth.custom,block,optional"`
```
rfratto, line 40:
```suggestion
	AuthCustom     *AuthCustom     `river:"custom,block,optional"`
```

component/remote/vault/vault.go
```diff
@@ -0,0 +1,237 @@
...
+	if err := c.Update(args); err != nil {
+		return nil, err
+	}
+	return c, nil
+}
+
+// Run runs the remote.vault component, managing the lifetime of the retrieved
+// secret and renewing/rereading it as necessary.
+func (c *Component) Run(ctx context.Context) error {
+	// TODO(rfratto): manage lifecycle of auth tokens
```
rfratto, line 176: Two things to do here:

* Re-read the secret:
  * On the configured refresh interval 
* If we need more than token auth:
  * Renew the auth token using [LifetimeWatcher](https://pkg.go.dev/github.com/hashicorp/vault/api#LifetimeWatcher), until it's no longer renewable; after which, get a new auth token.

component/remote/vault/vault.go
```diff
@@ -0,0 +1,237 @@
...
+	secretStore := newArgs.secretStore(newClient)
+	secret, err := secretStore.Read(context.Background(), &newArgs)
+	if err != nil {
+		return fmt.Errorf("reading secret: %w", err)
+	}
+
+	c.mut.Lock()
+	c.authToken = authToken
+	c.client = newClient
+	c.secret = secret
```
rfratto, line 212: This needs to be exported to components using OnStateChange.

pkg/util/k8s/k8s.go
```diff
@@ -11,16 +11,16 @@ import (
...
 	gragent "github.com/grafana/agent/pkg/operator/apis/monitoring/v1alpha1"
+	k3d_client "github.com/k3d-io/k3d/v5/pkg/client"
+	config "github.com/k3d-io/k3d/v5/pkg/config"
+	k3d_cfgtypes "github.com/k3d-io/k3d/v5/pkg/config/types"
+	k3d_config "github.com/k3d-io/k3d/v5/pkg/config/v1alpha4"
+	k3d_log "github.com/k3d-io/k3d/v5/pkg/logger"
+	k3d_runtime "github.com/k3d-io/k3d/v5/pkg/runtimes"
+	k3d_docker "github.com/k3d-io/k3d/v5/pkg/runtimes/docker"
+	k3d_types "github.com/k3d-io/k3d/v5/pkg/types"
+	k3d_version "github.com/k3d-io/k3d/v5/version"
```
rfratto, line 22: These had to be updated to sort out dependency hell 😬

docs/sources/flow/reference/components/remote.vault.md
```diff
@@ -0,0 +1,319 @@
+---
+aliases:
+- /docs/agent/latest/flow/reference/components/remote.vault
+title: remote.vault
+---
+
+# `remote.vault`
+
+`remote.vault` connect to Vault to retrieve secrets. It can retrieve secrets
+using the [KV v2][] secrets engine.
```
tpaschalis, line 10: Should we make a distinction and mention that it fetches a single secret that can hold multiple values?

docs/sources/flow/reference/components/remote.vault.md
```diff
@@ -0,0 +1,319 @@
...
+
+The `auth.token` block auhenticates to Vault using the [AppRole auth
+method][AppRole].
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`role_id` | `string` | Role ID to authenticate as. | | yes
+`secret` | `secret` | Secret to authenticate with. | | yes
+`wrapping_token` | `bool` | Whether to [unwrap][] the token. | `false` | no
+`mount_path` | `string` | Mount path for the login. | `"approle"` | no
```
tpaschalis, line 122: I think we're not setting the `mount_path` default in any of the authentication methods here.

docs/sources/flow/reference/components/remote.vault.md
```diff
@@ -0,0 +1,319 @@
...
+
+Credentials used to connect to AWS are specified by the environment variables
+`AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and `AWS_SESSION`. The
+environment variable `AWS_SHARED_CREDENTIALS_FILE` may be specified to use a
+credentials file instead.
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`type` | `string` | Mechanism to authenticate against AWS with. | | yes
+`region` | `string` | AWS region to connect to. | | no
```
tpaschalis, line 139: AFAICT, if the `region` is left empty, it will try to be inferred from an API call to the EC2 metadata service. If metadata is not available, it will default to `us-east-1`. Should we document this?

component/remote/vault/refresher.go
```diff
@@ -0,0 +1,278 @@
...
+
+func getSecretInfo(secret *vault.Secret, updateTime time.Time) secretInfo {
+	return secretInfo{
+		LatestRequestID:  secret.RequestID,
+		LastUpdateTime:   updateTime,
+		SecretExpireTime: secretExpireTime(secret),
+		Renewable:        secret.Renewable,
+		Warnings:         secret.Warnings,
+	}
+}
```
tpaschalis, line 269: This can panic in cases where a nil secret is returned (eg. from an authManager using the AuthToken method to authenticate).

component/remote/vault/refresher.go
```diff
@@ -0,0 +1,278 @@
...
+
+// needsLifecycleWatcher determines if a secret needs a lifecycle watcher.
+// Secrets only need a lifecycle watcher if they are renewable or have a lease
+// duration.
+func needsLifecycleWatcher(secret *vault.Secret) bool {
+	if secret.Auth != nil {
+		return secret.Auth.Renewable || secret.Auth.LeaseDuration > 0
+	}
+	return secret.Renewable || secret.LeaseDuration > 0
+}
```
tpaschalis, line 216: This can also panic in case this is hit on the codepath of authManager using certain AuthMethods

tpaschalis (commented):
Haven't put it to work versus an actual Vault instance to be honest; but looks good except for some minor points.

component/remote/vault/vault.go
```diff
@@ -0,0 +1,309 @@
...
+	"github.com/grafana/agent/component"
+	"github.com/grafana/agent/pkg/flow/rivertypes"
+	"github.com/oklog/run"
+
+	vault "github.com/hashicorp/vault/api"
+)
+
+func init() {
+	component.Register(component.Registration{
+		Name:    "remote.vault",
```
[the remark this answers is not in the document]
rfratto: > The authManager will need to re-authenticate to vault every once in a while so we can continue getting the newest value stored in the secret.

Yes. Authentication tokens (also a `*vault.Secret`) can have leases which are used to renew them. My understanding is that authenticating against Vault can have side effects depending on the authentication method you're using, so it's better to renew your authentication lease whenever possible instead of always re-authenticating whenever it's time to reread the secret.

component/remote/vault/auth.go
```diff
@@ -0,0 +1,339 @@
...
+		return a.AuthKubernetes
+	case a.AuthLDAP != nil:
+		return a.AuthLDAP
+	case a.AuthUserPass != nil:
+		return a.AuthUserPass
+	case a.AuthCustom != nil:
+		return a.AuthCustom
+	}
+
+	panic("remote.vault: unreachable")
```
mattdurham, line 65: Can we return an error here instead of panic? This feels like it could panic during an update.

mattdurham (changes requested):
Found some issues with attr naming and questions about naming.

component/remote/vault/auth.go
```diff
@@ -0,0 +1,339 @@
...
+		opts = append(opts, approle.WithMountPath(a.MountPath))
+	}
+
+	auth, err := approle.NewAppRoleAuth(a.RoleID, secret, opts...)
+	if err != nil {
+		return nil, fmt.Errorf("auth.approle: %w", err)
+	}
+	s, err := cli.Auth().Login(ctx, auth)
+	if err != nil {
+		return nil, fmt.Errorf("auth.approle: %w", err)
```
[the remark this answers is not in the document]
mattdurham: Probably not an issue since its a pattern your using in the other `auth` areas.

component/remote/vault/refresher.go
```diff
@@ -0,0 +1,278 @@
...
+	tm.cli = cli
+	tm.mut.Unlock()
+
+	ctx, cancel := context.WithTimeout(context.Background(), tokenManagerInitializeTimeout)
+	defer cancel()
+
+	tm.updateToken(ctx)
+}
+
+// SetRefrehInterval sets a forced refresh interval, separate from automatic
```
mattdurham, line 231: nit: `SetRefreshInterval`

component/remote/vault/vault.go
```diff
@@ -0,0 +1,309 @@
...
+
+	for key, value := range secret.Data {
+		switch value := value.(type) {
+		case string:
+			newExports.Data[key] = rivertypes.Secret(value)
+		case []byte:
+			newExports.Data[key] = rivertypes.Secret(value)
+
+		default:
+			// Non-string secrets are ignored.
```
mattdurham, line 281: Is there any benefit to logging anything here?

component/remote/vault/auth.go
```diff
@@ -0,0 +1,339 @@
...
+	}
+	return s, nil
+}
+
+// AuthAWS authenticates against Vault with AWS.
+type AuthAWS struct {
```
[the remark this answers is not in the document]
rfratto: I left it out on purpose, I don't think it makes sense to set a static nonce for the lifetime of the component, since the idea of a nonce is that it's used just once.

mattdurham (approved):
Awesome LGTM

tpaschalis (approved):
:shipit:

PR #3429, opened 2023-04-03
Add Grafana Agent Flow service manager for Windows

### LINKED ISSUE #3243: Move Windows service functionality to a dedicated program
I propose creating a dedicated program responsible for running Grafana Agent as a Windows service. This would remove the native Windows service capabilities from the agent. 

I see a few main reasons to do this:

* It will allow us to collect _all_ logs from stdout/stderr of the agent to forward to the Windows Event Logger, including panics and any logs which don't go through our logger. 
* It would remove the need for #2061, as the startup time of Grafana Agent would no longer impact the lifecycle of the service. 
* It may make it easier to add Windows service-specific functionality, such as specifying the command line arguments of the Agent binary via an `.ini` file or Windows registry keys.

This is similar to how Grafana does their Windows service; the Grafana Windows installer [installs NSSM](https://github.com/grafana/grafana/blob/d5a9a0cea0d0a187db08b7e60b44b23c165dc3f0/scripts/build/ci-msi-build/msigenerator/generator/utils.py#L14) which is responsible for managing the lifecycle of the Grafana server. 

However, instead of using NSSM, to keep things lightweight and avoid any surprises about installed software, I suggest the following:

* We create a new command, `./cmd/grafana-agent-service` which manages Grafana Agent as a child process on Windows, handling its lifecycle (automatic restarts?) and forwarding logs to the Windows Event logger. 
* `./cmd/grafana-agent-service` is only built and shipped as part of the Windows installer. 

I find the need for being able to collect and log potential panics the biggest reason to switch over to a dedicated program to run the agent, but being able to remove the need for #2061 is a nice bonus. 

cc @mattdurham for thoughts.

rfratto:
This PR introduces a new Grafana Agent Flow service manager for Windows installations. It is a replacement of the native Windows service support with the goals of addressing the proposal in #3243. 

While hypothetically the same program could be used for static mode with minimal changes, it only targets Flow mode right now. 

Closes #3243.


cmd/grafana-agent-service/config_windows.go
```diff
@@ -0,0 +1,46 @@
...
+	Args []string
+
+	// WorkingDirectory points to the working directory to run the Grafana Agent
+	// binary from.
+	WorkingDirectory string
+}
+
+// loadConfig loads the config from the Windows registry.
+func loadConfig() (*config, error) {
+	agentKey, err := registry.OpenKey(registry.LOCAL_MACHINE, `Software\Grafana\Grafana Agent`, registry.READ)
```
rfratto, line 26: Should this be `Software\Grafana\Grafana Agent Flow`? 

cc @erikbaranowski @mattdurham

packaging/grafana-agent-flow/windows/config.river
```diff
@@ -0,0 +1,3 @@
+logging {
+	level = "info"
+}
```
rfratto, line 3: Should we include any other examples here, even if they're commented out?

cmd/grafana-agent-service/config_windows.go
erikbaranowski: I'd say yes. It will be more obvious that this custom installer created stuff when debugging plus it doesn't have the flow/static installer bumping into each other.

If you do, reminder to update `install_script.nsis`

cmd/grafana-agent-service/logger_windows.go
```diff
@@ -0,0 +1,94 @@
...
+var (
+	_ log.Logger = (*logger)(nil)
+	_ io.Writer  = (*logger)(nil)
+)
+
+// newLogger creates a new logger which writes logs to the Windows Event
+// Logger.
+func newLogger() (*logger, error) {
+	eventTypes := uint32(eventlog.Info | eventlog.Warning | eventlog.Error)
+	_ = eventTypes
```
mattdurham, line 28: What does this line do?

rfratto: Oops, I added it during development as a way to remove the "unused variable" error, but forgot to take it out. Removing.

cmd/grafana-agent-service/config_windows.go
rfratto: I'll change to `Grafana Agent Flow`, we can always change it again later and have some kind of migration or support both paths.

cmd/grafana-agent-service/logger_windows.go
```diff
@@ -0,0 +1,83 @@
...
+// If the data contains the phrase "warn," then the text is logged as a
+// warn-level event. If the data contains the phrase "error," then the text is
+// logged as an error-level event.
+func (l *logger) Write(data []byte) (n int, err error) {
+	var (
+		leveledLogger = l.el.Info
+		msg           = string(data)
+	)
+
+	switch {
```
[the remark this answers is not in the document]
rfratto: More tolerant how? Like finding a different way to detect whether it's a warning/error log?

cmd/grafana-agent-service/service.go
```diff
@@ -0,0 +1,123 @@
...
+		select {
+		case <-ctx.Done():
+			return
+		case <-time.After(restartBackoff):
+		}
+
+		// Increase the backoff period for the next iteration of the loop.
+		restartBackoff = restartBackoff * 2
+		if restartBackoff > svc.maxRestartBackoff {
+			restartBackoff = svc.maxRestartBackoff
```
mattdurham, line 112: Does this mean it will try to restart forever? Maybe we have a maximum timeframe on it?

rfratto: Yeah, it will try to restart forever unless you stop the service.

Do you want me to add a maximum number of restarts?

mattdurham (approved):
Very minor nits but overall LGTM

PR #3438, opened 2023-04-03
Module shared library

erikbaranowski:
#### PR Description

Put shared code into a parent package for module components

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/component/module/file/file.go
+++ b/component/module/file/file.go
@@ -1,32 +1,23 @@
-// Package string defines the module.file component.
 package file
 
 import (
 	"context"
 	"fmt"
 	"net/http"
-	"path"
 	"sync"
 	"time"
 
-	"github.com/go-kit/log"
-	"github.com/gorilla/mux"
 	"github.com/grafana/agent/component"
 	"github.com/grafana/agent/component/local/file"
-	"github.com/grafana/agent/pkg/flow"
-	"github.com/grafana/agent/pkg/flow/logging"
-	"github.com/grafana/agent/pkg/flow/rivertypes"
-	"github.com/grafana/agent/pkg/flow/tracing"
+	"github.com/grafana/agent/component/module"
 	"github.com/grafana/agent/pkg/river"
-	"github.com/grafana/agent/web/api"
-	"github.com/prometheus/client_golang/prometheus"
 )
 
 func init() {
 	component.Register(component.Registration{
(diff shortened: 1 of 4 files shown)


component/module/file/file.go
```diff
@@ -163,85 +123,35 @@ func (c *Component) Update(args component.Arguments) error {
...
-			c.setHealth(component.Health{
-				Health:     component.HealthTypeHealthy,
-				Message:    "module content loaded",
-				UpdateTime: time.Now(),
-			})
-		}
+		defer c.mut.RUnlock()
 
-		return err
+		_ = c.mod.LoadFlowContent(c.args.Arguments, e.(file.Exports).Content.Value)
```
mattdurham, line 136: Is there a reason we are eating this error? I noticed that it does it in the previous commit too.

component/module/module.go
```diff
@@ -0,0 +1,131 @@
...
+			OnExportsChange: func(exports map[string]any) {
+				o.OnStateChange(Exports{Exports: exports})
+			},
+		}),
+	}
+}
+
+// LoadFlowContent loads the flow controller with the current component content.
+func (c *ModuleComponent) LoadFlowContent(arguments map[string]any, contentValue string) error {
+	c.mut.RLock()
```
mattdurham, line 63: I wonder if its clearer to wrap all this in a Rlock with defer instead of doing it in sections.

erikbaranowski: Good question, that's something I've been wondering about in go. Should I try to hold locks for the absolute minimal amount of time or value the clarity of the defer pattern when it's 'close enough'?

erikbaranowski: locks removed. I used to lock for the content/arguments but those are now function params after the refactor

component/module/file/file.go
erikbaranowski: good question, this got me playing with it and I want to make sure the way things work between the modules is consistent. I have to update a couple things.

erikbaranowski: take a look at the last commit. I put comments where I hope it answers your question. I also had to add back in the content tracking and call to `LoadFlowContent` from `Update`. I had cut a little too much in my refactoring.

mattdurham (approved):
Awesome LGTM

PR #3452, opened 2023-04-04
Operator: Check agent selector on each reconcile.

captncraig:
You should be able to run multiple operators, each using mutually exclusive --agent-selector flags. Currently, there are issues where they try and overwrite things that the other operator is managing. 

My guess is that controller-runtime is finding resources owned by other operators' grafana-agents (like statefulsets), and transitively reconciling the owners, even though the top-level agent is not found by the controller. 

The fix is to always check the selector before reconciling, and skip if it is not ours.

This also contains a fix for logging. Logs are full of `GrafanaAgent="unsupported value type"`, and this will replace it with `GrafanaAgent=namespace/name`.


jcreixell (commented):
nice catch, thanks for fixing this!

Looks like there are some tests in https://github.com/grafana/agent/blob/main/pkg/operator/operator_test.go#L28-L44, I think it should be relatively simple to add one reproducing this issue (we just define the input and expected output resources), WDYT?

captncraig:
@jcreixell I'm having quite a few problems with those tests, even getting them to run properly at all. As this is a relatively simple change, how do you feel about merging for 0.33 and spending some time looking at those integration tests together at a future time?

jcreixell (approved):

PR #3461, opened 2023-04-05
prometheus.remote_write: fix issue where unnecessary directory was used

rfratto:
prometheus.remote_write was originally creating a folder called `wal/COMPONENT_ID` relative to the data path given to the component.

Because the data path given to the component is already unique, AND because the Prometheus storage code creates a WAL directory, this means that the WAL was being stored at
`DATA_PATH/COMPONENT_ID/wal/COMPONENT_ID/wal`. This issue compounds itself when using modules, as each module adds an additional unnecessary directory to the path.

This commit changes the WAL to be stored at
`DATA_PATH/COMPONENT_ID/wal`.

### DIFF
--- a/component/prometheus/remotewrite/remote_write.go
+++ b/component/prometheus/remotewrite/remote_write.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"math"
+	"os"
 	"path/filepath"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


component/prometheus/remotewrite/remote_write.go
```diff
@@ -62,15 +63,23 @@ type Component struct {
...
+	//
+	// We best-effort attempt to delete the old path if it already exists to not
+	// leak storage space.
+	oldDataPath := filepath.Join(o.DataPath, "wal", o.ID)
+	_ = os.RemoveAll(oldDataPath)
+
 	walLogger := log.With(o.Logger, "subcomponent", "wal")
-	dataPath := filepath.Join(o.DataPath, "wal", o.ID)
-	walStorage, err := wal.NewStorage(walLogger, o.Registerer, dataPath)
+	walStorage, err := wal.NewStorage(walLogger, o.Registerer, o.DataPath)
```
rfratto, line 76: (wal.NewStorage creates a `wal` subdirectory from the path given to it)

tpaschalis (approved):
:shipit:

PR #3465, opened 2023-04-05
introduce limit to module depth

### LINKED ISSUE #3110: Flow Modules: Add a stack limit

erikbaranowski:
#### PR Description

This PR adds a new command line argument to control how deep modules can be nested. This takes care of validating the case where modules recursively refer to each other.

Closes #3110

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/component/module/file/file.go
+++ b/component/module/file/file.go
@@ -51,7 +51,7 @@ func (a *Arguments) UnmarshalRiver(f func(interface{}) error) error {
 
 // Component implements the module.file component.
 type Component struct {
-	mod module.ModuleComponent
+	mod *module.ModuleComponent
 
 	mut     sync.RWMutex
 	args    Arguments
@@ -68,12 +68,15 @@ var (
 
 // New creates a new module.file component.
 func New(o component.Options, args Arguments) (*Component, error) {
+	moduleComponent, err := module.NewModuleComponent(o)
+	if err != nil {
+		return nil, err
+	}
 	c := &Component{
-		mod:  module.NewModuleComponent(o),
+		mod:  moduleComponent,
 		args: args,
 	}
 
-	var err error
 	c.managedLocalFile, err = c.NewManagedLocalComponent(o)
 	if err != nil {
 		return nil, err
--- a/component/module/file/file.go
+++ b/component/module/file/file.go
@@ -71,15 +71,12 @@ var (
 
 // New creates a new module.file component.
 func New(o component.Options, args Arguments) (*Component, error) {
-	moduleComponent, err := module.NewModuleComponent(o)
-	if err != nil {
-		return nil, err
-	}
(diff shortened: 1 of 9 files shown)


component/registry.go
```diff
@@ -67,6 +67,10 @@ type Options struct {
 	// HTTPPath is the base path that requests need in order to route to this component.
 	// Requests received by a component handler will have this already trimmed off.
 	HTTPPath string
+
+	// ModuleDepth is the depth of the current module level for the component.
+	// The top level flow controller will have depth 0.
+	ModuleDepth uint8
```
rfratto, line 73: IMO I'd rather not continue adding extra fields to this or flow.Options, since it's bloating the API. The module depth can be inferred from the component and controller ID by counting the `/` characters.

component/module/file/file.go
erikbaranowski, line 1: These changes are largely due to weirdness uncovered when implementing module depth. They will be necessary to prevent `LoadFlowContent` from being called too many times.

component/module/module.go
```diff
@@ -33,13 +33,13 @@ type Exports struct {
 }
 
 // NewModuleComponent initializes a new ModuleComponent.
-func NewModuleComponent(o component.Options) ModuleComponent {
+func NewModuleComponent(o component.Options) *ModuleComponent {
```
erikbaranowski, line 36: Not necessary for this PR but I think a pointer may be more correct for a new function return

pkg/flow/internal/controller/loader.go
```diff
@@ -471,6 +471,5 @@ func multierrToDiags(errors error) diag.Diagnostics {
 
 // If the definition of a module ever changes, update this.
 func (l *Loader) isModule() bool {
-	// Either 1 of these checks is technically sufficient but let's be extra careful.
-	return l.globals.OnExportsChange != nil && l.globals.ControllerID != ""
+	return l.globals.ModuleDepth > 0
```
erikbaranowski, line 474: not necessary but clearer if we have ModuleDepth available to us imo

component/module/file/file.go
erikbaranowski, line 1: These changes are largely due to weirdness uncovered when implementing module depth. They will be necessary to prevent `LoadFlowContent` from being called too many times independent of module depth implementation

erikbaranowski:
Closing this PR.

- The `module.file` refactor was pulled into a different PR https://github.com/grafana/agent/pull/3488
- We have decided on a new API design for how to pass data to a module flow controller instead of pushing everything via options which will be worked on as a prereq for this issue. The final code will not be implemented in this way.

PR #3466, opened 2023-04-05
Update install-agent-macos.md

ricardolupo:
added full path to config.yml using example from mkdir command

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/set-up/install-agent-macos.md
+++ b/docs/sources/set-up/install-agent-macos.md
@@ -33,8 +33,13 @@ Use Homebrew to install the most recent released version of Grafana using the Ho
     ```
 
 1. Modify `config.yml` with your configuration requirements. 
+    
+    ```
+    vi $(brew --prefix)/etc/grafana-agent/config.yml
+    ```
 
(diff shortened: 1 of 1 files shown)


docs/sources/set-up/install-agent-macos.md
```diff
@@ -33,8 +33,13 @@ Use Homebrew to install the most recent released version of Grafana using the Ho
     ```
 
 1. Modify `config.yml` with your configuration requirements. 
+    
+    ```
+    vi $(brew --prefix)/etc/grafana-agent/config.yml
```
tpaschalis, line 38: @ricardolupo I'm a little cautious that not everyone might be familiar with Vi here and might get stuck. 

WDYT about telling users where to find the config file in the previous line instead?

tpaschalis:
I'm closing this PR as users will not be necessarily familiar with vi, and we're looking to add new sections on guiding users to building their first configurations.

PR #3488, opened 2023-04-06
refactor module.file with better embedded component handling

erikbaranowski:
#### PR Description

This fixes up some issues with module.file running an embedded local.file component properly

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated


component/module/file/file.go
```diff
@@ -69,24 +72,43 @@ var (
...
 	}
-
 	if err := c.Update(args); err != nil {
 		return nil, err
 	}
 	return c, nil
 }
 
+// NewManagedLocalComponent creates the new local.file managed component.
+func (c *Component) NewManagedLocalComponent(o component.Options) (*file.Component, error) {
```
rfratto, line 93: Does this need to be exported as part of the public API of the component? Seems like we can make it internal, right?

erikbaranowski: good catch, will fix tomorrow

erikbaranowski:
It looks like an armv6 build issue. I'm not sure what the latest status on that is.

erikbaranowski:
merged main which removes the armv6 builds and resolved the build fail

rfratto (approved):
LGTM

PR #3513, opened 2023-04-11
Update create-config-file.md

navbahl:
Correcting misleading "required" to "optional".

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/configuration/create-config-file.md
+++ b/docs/sources/configuration/create-config-file.md
@@ -41,7 +41,7 @@ integrations:
 ```
 
 In this example, we first must configure the `wal_directory` which is used to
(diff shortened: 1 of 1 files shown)


navbahl:
Minor but may be provides more clarity - Then the wording should be corrected. Instead of "This is required, but ensures that samples will be redelivered in case of failure", we should remove the "but" and may be replace it with an "and".

tpaschalis:
I'm closing this PR if that's okay. As mentioned, the WAL is required for metrics; we'll continue working on our documentation to improve the users' understanding of how the WAL works.

PR #3524, opened 2023-04-12
misc: add support for inline TLS settings for Prometheus-based components

rfratto:
This updates the prometheus/common dependency to use a fork until prometheus/common#472 is merged.

Components which use the prometheus/common dependency for TLS configs will automatically support inline TLS settings. Some components, particularly loki.source.kafka and loki.source.syslog had to be updated as they build their own TLS configs.


erikbaranowski (approved):
looks like a lint and test to sort but overall LGTM

PR #3528, opened 2023-04-12
drone: remove usage of Docker from test path

rfratto:
Remove the usage of Docker from the test path so CI tests run faster.


mattdurham (approved):
LGTM

PR #3532, opened 2023-04-13
river/encoding: support map[string]T blocks

rfratto:
Add support for map[string]T blocks to river/encoding, first added in #3232.

### DIFF
--- a/pkg/river/encoding/block.go
+++ b/pkg/river/encoding/block.go
@@ -37,20 +37,48 @@ func (bf *blockField) convertBlock(namePrefix []string, reflectValue reflect.Val
 		}
 		reflectValue = reflectValue.Elem()
 	}
-	if reflectValue.Kind() != reflect.Struct {
-		return fmt.Errorf("convertBlock can only be called on struct kinds, got %s", reflectValue.Kind())
-	}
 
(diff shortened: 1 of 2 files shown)


pkg/river/encoding/block.go
```diff
@@ -37,20 +37,48 @@ func (bf *blockField) convertBlock(namePrefix []string, reflectValue reflect.Val
...
-	bf.Name = strings.Join(mergeStringSlices(namePrefix, f.Name), ".")
-	bf.Type = "block"
-	bf.Label = getBlockLabel(reflectValue)
+	switch reflectValue.Kind() {
+	case reflect.Struct:
+		bf.Name = strings.Join(mergeStringSlices(namePrefix, f.Name), ".")
+		bf.Type = "block"
+		bf.Label = getBlockLabel(reflectValue)
+
+		fields, err := getFieldsForBlock(namePrefix, reflectValue.Interface())
```
erikbaranowski, line 47: should this be renamed to `getFieldsForBlockStruct` and a new function be written for the below case called something like `getFieldsForBlockMap`? I also noticed `getFieldsForBlockStruct` gets used 1 other place so sanity checking, do we need to do anothing additional there for maps?

mattdurham (approved):
LGTM

PR #3533, opened 2023-04-13
docs: use river tag consistently for River samples

rfratto:
Some markdown blocks were missing the `river` tag.


erikbaranowski (approved):
LGTM

PR #3536, opened 2023-04-13
component: change module loaders to use an `arguments` block

### LINKED ISSUE #3534: Flow Modules: Update arguments attribute to be a block instead

rfratto:
This commit changes module modules to pass arguments to the module via an arguments block, rather than an arguments attribute.

Closes #3534.

image

### DIFF
--- a/docs/sources/flow/reference/components/module.file.md
+++ b/docs/sources/flow/reference/components/module.file.md
@@ -60,7 +60,7 @@ arguments | [arguments][] | Arguments to pass to the module. | no
 
 ### arguments block
 
-The `arguments` block specifies the list of arguments to pass to the loaded
+The `arguments` block specifies the list of values to pass to the loaded
 module.
 
 The attributes provided in the `arguments` block are validated based on the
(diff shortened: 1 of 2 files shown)


erikbaranowski (approved):
LGTM!

docs/sources/flow/reference/components/module.file.md
```diff
@@ -42,20 +43,35 @@ Name | Type | Description | Default | Required
...
+
+Hierarchy        | Block      | Description | Required
+---------------- | ---------- | ----------- | --------
+arguments | [arguments][] | Arguments to pass to the module. | no
+
+[arguments]: #arguments-block
+
+### arguments block
+
+The `arguments` block specifies the list of arguments to pass to the loaded
```
mattdurham, line 63: Thoughts on going with:

"The `arguments` block specifies the list of values to pass to the loaded"

I hate using the same word in the definition of the word ;)

mattdurham (approved):
Small nit but looks good

docs/sources/flow/reference/components/module.file.md
rfratto: Works for me 👍

PR #3539, opened 2023-04-13
remote.http: Add fail_on_error property

jkroepke:
#### PR Description

Currently, `http.remote` always fails silently without any error on log. The error is maybe available on the UI.

In my use-case, if  `http.remote` fails, that should be result into a critical and visible error. If `remote.http` is used together with `module.string`, it could really hard to debug, what the error is.

#### PR Checklist

- [ ] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/component/remote/http/http.go
+++ b/component/remote/http/http.go
@@ -37,6 +37,7 @@ type Arguments struct {
 	URL           string        `river:"url,attr"`
 	PollFrequency time.Duration `river:"poll_frequency,attr,optional"`
 	PollTimeout   time.Duration `river:"poll_timeout,attr,optional"`
+	FailOnError   bool          `river:"fail_on_error,attr,optional"`
 	IsSecret      bool          `river:"is_secret,attr,optional"`
 
 	Client common_config.HTTPClientConfig `river:"client,block,optional"`
@@ -47,6 +48,7 @@ var DefaultArguments = Arguments{
 	PollFrequency: 1 * time.Minute,
 	PollTimeout:   10 * time.Second,
 	Client:        common_config.DefaultHTTPClientConfig,
+	FailOnError:   false,
 }
 
 var _ river.Unmarshaler = (*Arguments)(nil)
@@ -129,7 +131,7 @@ func (c *Component) Run(ctx context.Context) error {
 		case <-ctx.Done():
 			return nil
 		case <-time.After(c.nextPoll()):
-			c.poll()
+			return c.poll()
 		case <-c.updated:
 			// no-op; force the next wait to be reread.
 		}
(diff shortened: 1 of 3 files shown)


component/remote/http/http.go
```diff
@@ -129,7 +131,7 @@ func (c *Component) Run(ctx context.Context) error {
 		case <-ctx.Done():
 			return nil
 		case <-time.After(c.nextPoll()):
-			c.poll()
+			return c.poll()
```
rfratto, line 134: Components are expected to always run so they can continue to be updated with new parameters. 

I don't think allowing the component to exit is a good idea, and we should discuss alternatives to what you're trying to do. 

I would be comfortable adding new log messages if a poll fails, or finding other ways to signal to the user that there's a problem without expecting them to use the UI or metrics for the count of unhealthy components.

jkroepke: >  There's no mechanism to restart a component at runtime if it terminated on its own.

Off-topic: Does this also mean, if [`module.string`](https://grafana.com/docs/agent/next/flow/reference/components/module.string/) gets invalid config once (for example through `remote.http`), it will stop working?

rfratto: > Off-topic: Does this also mean, if [module.string](https://grafana.com/docs/agent/next/flow/reference/components/module.string/) gets invalid config once (for example through remote.http), it will stop working?

No, it will be marked as unhealthy but it will continue using the last valid config it received.

jkroepke: > Components are expected to always run so they can continue to be updated with new parameters. There's no mechanism to restart a component at runtime if it terminated on its own.

I was not aware of this before creating this PR. After local testing, I can confirm that everything works continue using the last valid config, but once I'm getting a `component exited with error` in the logs, at least any `remote.http` component will not longer pull periodically.

For example:

> ts=2023-04-13T17:47:25.673108Z level=error trace_id=53835d22852dffcd40e2b95b94386e38 msg="failed to evaluate config" node=module.string.default err="updating component: module.string.default:1:1: Component \"prometheus.exporter.unix\" is a singleton and unsupported inside a module"

Stops pulling on any `remote.http` (including on main module and inside the module).

rfratto: Wait, sorry, are you saying you're seeing `component exited with error` without this change?

jkroepke: @rfratto Since I'm running a self-complied version of the grafana agent based on the main branch, I'm trying to reconfirm this by rebuild the agent and running `go cache -clear` in front.

I'm not longer see such error messages like `component exited with error` as before reported.

Issue #3389, opened 2023-03-30
Proposal: add optional second argument to env() stdlib function


dosmanak:
[env()](https://grafana.com/docs/agent/latest/flow/reference/stdlib/env/) function returns empty string if environment variable is not defined.

I've found the implementation is pure go [os.Getenv()](https://github.com/grafana/agent/blob/main/pkg/river/internal/stdlib/stdlib.go#L21).

I propose to implement custom `env()` function with optional second argument as `env("POSSIBLE_VAR","reasonable default")` so the configuration can be finetuned using environment variables.

rfratto:
To make this not a breaking change, we'd probably want do this by one of:

* Add a new function that is some form of `env_or_default`
* Adding support for optional arguments (I'm not sure about this being a good idea)
* Rely on #2638, even if it's verbose, it would enable this use case 
* Have some kind of `or` function similar to Go templating where it will use the first non-zero value, so that `or(env("VAR_DOES_NOT_EXIST"), "default_value")` could be used.

rfratto:
Another option would be supporting an `envsubst` function to allow Bash-like syntax for defaulting and other patterns:

```river 
envsubst("${POSSIBLE_VAR:-reasonable default}")
```

PR #3540, opened 2023-04-13
stdlib/env: Add default value

jkroepke:
#### PR Description

This PR enrich the `env` functions from stdlib to provide a default value as fallback.

```
> env("HOME")
"/home/grafana-agent"
> env("DOES_NOT_EXIST")
""
> env("DOES_NOT_EXIST", "default")
"default"
```

#### PR Checklist

- [ ] CHANGELOG updated
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/flow/reference/stdlib/env.md
+++ b/docs/sources/flow/reference/stdlib/env.md
@@ -8,7 +8,8 @@ title: env
 
 The `env` function gets the value of an environment variable from the system
 Grafana Agent is running on. If the environment variable does not exist, `env`
-returns an empty string.
+returns an empty string. An optional default value can be provided, if the environment
+variable is not present.
 
 ## Examples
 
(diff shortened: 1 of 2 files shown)


rfratto:
Thanks for the contribution! I'm not sure if this is how we want to handle it though, and I think we need to give it more thought and get consensus across the board. There's some discussion about this in #3389, where there were a few options that were laid out.

rfratto:
@mattdurham @tpaschalis WDYT about adding optional values to the `env` function? Or should we find another way to support fallbacks when an environment variable doesn't exist?

mattdurham:
I would rather add a coalesce function, I may poke at that this week should be simple to do.

`coalesce(args...)`, and it goes with the first non zero/nil value. So `coalesce(env("HOME"),"/home/bob")`

PR #3542, opened 2023-04-13
Get pipeline secrets from Vault rather than Drone

julienduchesne:


.drone/pipelines/secrets.jsonnet
```diff
@@ -0,0 +1,23 @@
+local newSecret(name) = {
```
rfratto, line 1: The `pipelines` folder is for jobs which run in Drone, can this be moved to util instead?

julienduchesne: Done!

rfratto (approved):
Thanks!

PR #3545, opened 2023-04-13
Introduce `component.LeastHealthy` utility function and fix bug in module.file

rfratto:
This introduces a new function called `component.LeastHealth` which returns the least healthy Health from a list. 

This is useful when there are multiple Health values for a given component, such as with `module.file` containing the health of the managed `local.file` component and its module controller. 

Using this function in `module.file` fixes an issue where the health of `local.file` was never reported, which can be observed by using `module.file` and deleting the file on disk after the agent first loads successfully.

### DIFF
--- a/component/component_health.go
+++ b/component/component_health.go
@@ -101,7 +101,7 @@ func (ht *HealthType) UnmarshalText(text []byte) error {
 // considered to be the least healthy.
 //
 // Health types are first prioritized by [HealthTypeExited], followed by
-// [HealthTypeUnhealthy], [HealthTypeHealthy], and [HealthTypeUnknown].
+// [HealthTypeUnhealthy], [HealthTypeUnknown], and [HealthTypeHealthy].
 //
 // If multiple arguments have the same Health type, the Health with the most
 // recent timestamp is returned.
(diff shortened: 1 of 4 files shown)


component/component_health.go
```diff
@@ -96,3 +96,37 @@ func (ht *HealthType) UnmarshalText(text []byte) error {
 	}
 	return nil
 }
+
+// LeastHealthy returns the Health from the provided arguments which is
+// considered to be the least healthy.
+//
+// Health types are first prioritized by [HealthTypeExited], followed by
+// [HealthTypeUnhealthy], [HealthTypeHealthy], and [HealthTypeUnknown].
```
rfratto, line 104: I'm not sure if this order makes sense; it's a bit weird for `HealthTypeHealthy` to take precedence over `HealthTypeUnknown`. Should I flip the order of these two?

erikbaranowski: Ya I would flip. If one of the statuses is unknown I think we could argue the overall status is unknown

rfratto: Ok, flipped 👍

erikbaranowski (approved):
LGTM, nice touch considering update time as well.

PR #3550, opened 2023-04-14
[flow] prometheus.exporter.blackbox: Support embedded config

jkroepke:
#### PR Description

Add the possibility to pass the blackbox exporter config as string.

#### Notes to the Reviewer

Ref: https://grafana.slack.com/archives/C01050C3D8F/p1681416858928169

#### PR Checklist

- [ ] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/component/prometheus/exporter/blackbox/blackbox.go
+++ b/component/prometheus/exporter/blackbox/blackbox.go
@@ -1,6 +1,8 @@
 package blackbox
 
 import (
+	blackbox_config "github.com/prometheus/blackbox_exporter/config"
+	"gopkg.in/yaml.v2"
 	"time"
 
 	"github.com/grafana/agent/component"
@@ -76,9 +78,11 @@ func (t TargetBlock) Convert() []blackbox_exporter.BlackboxTarget {
 }
 
 type Arguments struct {
-	ConfigFile         string        `river:"config_file,attr"`
+	ConfigFile         string        `river:"config_file,attr,optional"`
+	Config             string        `river:"config,attr,optional"`
 	Targets            TargetBlock   `river:"target,block"`
 	ProbeTimeoutOffset time.Duration `river:"probe_timeout_offset,attr,optional"`
+	ConfigStruct       blackbox_config.Config
 }
 
 // UnmarshalRiver implements River unmarshalling for Arguments.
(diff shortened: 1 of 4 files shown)


component/prometheus/exporter/blackbox/blackbox_test.go
```diff
@@ -38,6 +38,36 @@ func TestUnmarshalRiver(t *testing.T) {
 	require.Contains(t, "http_2xx", args.Targets[1].Module)
 }
 
+func TestUnmarshalRiverWithInlineConfig(t *testing.T) {
```
marctc, line 41: Should we also add a test for the case that the content of `config` is invalid? I'm also curious how is this error message displayed when using this inline yaml format, and also how is displayed in combination of flow/river errors.

jkroepke: Good catch! I added some test case and the make the config related error more describable. (by adding an prefix)

docs/sources/flow/reference/components/prometheus.exporter.blackbox.md
```diff
@@ -33,7 +33,8 @@ Omitted fields take their default values.
 
 Name | Type | Description | Default | Required
 ---- | ---- | ----------- | ------- | --------
-`config_file`                 | `string`       | Blackbox configuration file with custom modules. | | yes
+`config_file`                 | `string`       | Blackbox configuration file with custom modules. | | no
+`config`                      | `string`       | Blackbox configuration with custom modules as YAML. | |no
 `probe_timeout_offset`        | `duration`     | Offset in seconds to subtract from timeout when probing targets.  | `"0.5s"` | no
 
 The `config_file` argument points to a YAML file defining which blackbox_exporter modules to use. See [blackbox_exporter]( https://github.com/prometheus/blackbox_exporter/blob/master/example.yml) for details on how to generate a config file.
```
marctc, line 40: Could we clarify that `config` is using inline yaml (or whatever is called, preferably with a link)?

component/prometheus/exporter/blackbox/blackbox.go
```diff
@@ -76,23 +81,39 @@ func (t TargetBlock) Convert() []blackbox_exporter.BlackboxTarget {
...
+		return err
+	}
+
+	if a.ConfigFile != "" && a.Config != "" {
+		return errors.New("config and config_file are mutually exclusive")
+	}
+
+	err := yaml.UnmarshalStrict([]byte(a.Config), &a.ConfigStruct)
+	if err != nil {
+		return fmt.Errorf("invalid config: %s", err)
```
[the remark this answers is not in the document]
jkroepke: looking at https://github.com/prometheus/blackbox_exporter/blob/master/CONFIGURATION.md, I would prefer blackbox config or blackbox_exporter config. Right now, only module can be configured but in theory more can be added.

docs/sources/flow/reference/components/prometheus.exporter.blackbox.md
jkroepke: I hope its more clear now!

marctc (approved):
LGTM, thanks!

PR #3557, opened 2023-04-17
`go fix` clean up & minor docs fixes

thampiotr:
#### PR Description
1. Ran `go fix ./...` that performed the following fixes:
  - it cleaned up old style of build tags. [As per Go 1.18 release notes:](https://tip.golang.org/doc/go1.18#go-command)
> Go 1.17 introduced //go:build lines as a more readable way to write build constraints, instead of // +build lines. As of Go 1.17, gofmt adds //go:build lines to match existing +build lines and keeps them in sync, while go vet diagnoses when they are out of sync.
>
> 
> Since the release of Go 1.18 marks the end of support for Go 1.16, all supported versions of Go now understand //go:build lines. In Go 1.18, go fix now removes the now-obsolete // +build lines in modules declaring go 1.18 or later in their go.mod files.
> 
> For more information, see https://go.dev/design/draft-gobuild.
  - it replaced "golang.org/x/net/context" with "context" import, as it was moved to standard library - see https://go.dev/doc/go1.7#context.
  -   We are using Go 1.19, so both of this changes should be safe.

2. A few minor fixes to documentation as I found them while reading the docs.
3. While running CI tests, I had a test fail on Windows run only. I suspect the timeout for it was too tight for Windows and increased it from 200ms to 1s. It didn't fail since.
4. A linter highlighted issue with not calling a cancel function. I've fixed this too.

A trivial change does not require an issue. This PR is mostly automated changes from `go fix`.


component/loki/source/internal/kafkatarget/target_syncer_test.go
```diff
@@ -66,7 +66,7 @@ func Test_TopicDiscovery(t *testing.T) {
 		}
 		group.mut.Unlock()
 		return reflect.DeepEqual([]string{"topic1"}, group.GetTopics())
-	}, 200*time.Millisecond, time.Millisecond, "expected topics: %v, got: %v", []string{"topic1"}, group.GetTopics())
+	}, time.Second, time.Millisecond, "expected topics: %v, got: %v", []string{"topic1"}, group.GetTopics())
```
thampiotr, line 69: This test failed on Windows only in CI. I've guessed that maybe timeout is too tight for Windows CI agent and increased it. It passed. If someone has more context, I'd be happy to hear about it.

component/loki/source/internal/kafkatarget/target_syncer_test.go
```diff
@@ -77,7 +77,7 @@ func Test_TopicDiscovery(t *testing.T) {
 		}
 		group.mut.Unlock()
 		return reflect.DeepEqual([]string{"topic1", "topic2"}, group.GetTopics())
-	}, 200*time.Millisecond, time.Millisecond, "expected topics: %v, got: %v", []string{"topic1", "topic2"}, group.GetTopics())
+	}, time.Second, time.Millisecond, "expected topics: %v, got: %v", []string{"topic1", "topic2"}, group.GetTopics())
```
thampiotr, line 80: See comment above.

component/discovery/file/file_test.go
```diff
@@ -30,7 +30,8 @@ func TestFile(t *testing.T) {
 	})
 	c := createComponent(t, dir, []string{path.Join(dir, "*.txt")}, nil)
 	ct := context.Background()
-	ct, _ = context.WithTimeout(ct, 5*time.Second)
+	ct, ccl := context.WithTimeout(ct, 5*time.Second)
+	defer ccl()
```
thampiotr, line 34: Here and below, the linter was not happy to discard the cancel function, so I made sure we run it. I suspect that the linting highlighted it now because it was a new rule added since this file was last time modified?

mattdurham: Should rerun the linter from scratch, are you running `make lint` or local linting? Also our make lint sometimes lags.

thampiotr: Yeah, I used `make lint` - it is now happy locally and in CI. But before this change I was getting a lot of:
```
component/discovery/file/file_test.go:33:6: lostcancel: the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak (govet)
	ct, _ = context.WithTimeout(ct, 5*time.Second)
	    ^
```

thampiotr:
Weird - the tests were all green before I've rebased. Will take a look into this.

ptodev (approved):
LGTM

PR #3560, opened 2023-04-17
Include file watching wal times from k146. 

### LINKED ISSUE #3551: Update Promtail depenendecy to include grafana/loki#8874
We need to port over grafana/loki#8874 so it can be included on the v0.33 release, since the PR was opened with the intent of bringing it downstream this release cycle. 

This should ideally wait for a new `k` release in Loki, but if there isn't one available which contains the PR prior to v0.33.0-rc.0 being cut, then an alternative solution can be used (e.g., temporarily using a fork which cherry-picks the functionality on top of the most recent `k` release, or just using the SHA of the merged PR specifically). 

The functionality introduced in grafana/loki#8874 should also be configurable in Flow, but this doesn't have to be done for the scope of this issue.

mattdurham:
#### PR Description

This primarily pulls in this change https://github.com/grafana/loki/pull/8874/files

Closes #3551

#### PR Checklist

- [X ] Documentation added
- [X] Tests updated

### DIFF
--- a/component/module/file/file_test.go
+++ b/component/module/file/file_test.go
@@ -1,4 +1,4 @@
-//go:build !windows
+//go:build linux
 
 package file
 
--- a/component/module/file/file_test.go
+++ b/component/module/file/file_test.go
@@ -1,5 +1,3 @@
-//go:build linux
-
 package file
 
 import (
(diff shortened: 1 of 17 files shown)


component/module/file/file_test.go
```diff
@@ -65,7 +67,7 @@ func TestModule(t *testing.T) {
 			require.NoError(t, err)
 
 			go c.Run(context.Background())
-			time.Sleep(200 * time.Millisecond)
+			time.Sleep(2 * time.Second)
```
mattdurham, line 70: This is dumb osx thing thats been popping up

pkg/util/eventually.go
```diff
@@ -11,7 +11,7 @@ import (
 var backoffRetry = backoff.Config{
 	MinBackoff: 10 * time.Millisecond,
 	MaxBackoff: 1 * time.Second,
-	MaxRetries: 5,
+	MaxRetries: 10,
```
mattdurham, line 14: Other dumb osx thing

tpaschalis (approved):
LGTM!

Issue #2216, opened 2022-09-24
Flow: `prometheus` components still partially function after removal


rfratto:
`prometheus` components which export receivers don't reject metrics after the component is removed. This can have some unintended side effects. 

To demonstrate the issue, start with this config file:

```river 
prometheus.scrape "default" {
  targets    = [{ __address__ = "localhost:12345" }]
  forward_to = [prometheus.relabel.default.receiver]
}

prometheus.relabel "default" {
  forward_to = [prometheus.remote_write.default.receiver] 

  // (no rules defined) 
}

prometheus.remote_write "default" {
  // (no endpoints defined) 
}
```

Initially, all components work as intended. Now, modify the config file to remove  `prometheus.relabel.default`, and reload the config by POSTing to `/-/reload`:

```river
prometheus.scrape "default" {
  targets    = [{ __address__ = "localhost:12345" }]
  forward_to = [prometheus.relabel.default.receiver]
}

prometheus.remote_write "default" {
  // (no endpoints defined) 
}
```

This causes `prometheus.scrape.default` to go unhealthy because the evaluation will fail. However, the component remains in its last valid state, and still holds a pointer to the now-removed `prometheus.relabel.default.receiver`, which was last configured to send to `prometheus.remote_write.default`. 

Because the prometheus components don't check to see if they were terminated, metrics will still be sent to `prometheus.remote_write` unexpectedly, even though there is no running component that would enable that to happen.

To avoid this issue, `prometheus` components which export a receiver should reject or ignore samples if they have been shut down.

ptodev:
I've been thinking more about this, and I cannot think of a good reason why we would want to allow an update with an invalid graph? And I cannot think of any reason why a user would want to knowingly send us an invalid one. The only reason I can think of is if it got sent by accident. Then, if we accept it, this could potentially lead to a loss of data. Would it not be better to just abort the whole update, like [this](https://github.com/grafana/agent/commit/fb4c8533049795aa7ac7218dac15cacbba7ca004)?

rfratto:
> I've been thinking more about this, and I cannot think of a good reason why we would want to allow an update with an invalid graph? 

One of the original goals of the Flow controller was that we'll try to apply as much as we can to reduce the blast radius of an invalid config. We can discuss changing that decision, but I feel it's separate from this issue still: it is invalid for the `prometheus` components to still accept metrics after they've been shut down.

PR #3566, opened 2023-04-19
flow: Add 'write_to' argument to logging block

### LINKED ISSUE #2700: flow: Allow the `logging` block to send the Agent's own logs to`loki.*` components
Since we've implemented #2214, and introduced an initial set of Loki components, we can now consider allowing the logging block to forward the Agent's own logging to one or more `loki.*` components.

This would require adding a new `write_to` optional field, that treats logged entries as `loki.Entry` values and propagates them to the given list of `LogsReceiver` capsule values.

```
logging {
  level    = "debug"
  format   = "logfmt"
  write_to = [loki.write.local.receiver, loki.process.redact.receiver]
}
```

### Comments/Open Questions
- Should logs continue to be written in `stderr`? My gut feeling is that yes, the `write_to` field should not otherwise interfere with the Agent's logging.
- As with other places in the codebase where we fan out Loki entries, this happens over Go channels in sequence; this means that a blocked pipeline might delay other log lines from being sent. Should this be best-effort and use a timeout?
- Users will be able to use the OTel <-> Loki converters in #2501 to send them to an OTel logging pipeline. I don't think that it's particularly useful to have write_to support _two_ kinds of receivers at the same time.

tpaschalis:
#### PR Description

This PR adds a `write_to` argument to the special `logging` configuration block so that it can tee off the Agent's own logs to one or more loki.* components' exported receivers.

Fixes #2700 

#### Notes to the Reviewer
Do you think there should be more docs/examples?

#### PR Checklist

- [X] CHANGELOG updated
- [X] Documentation added
- [X] Tests updated

### DIFF
--- a/pkg/flow/logging/sink.go
+++ b/pkg/flow/logging/sink.go
@@ -4,9 +4,13 @@ import (
 	"fmt"
 	"io"
 	"reflect"
+	"time"
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
+	"github.com/grafana/agent/component/common/loki"
+	"github.com/grafana/loki/pkg/logproto"
+	"github.com/prometheus/common/model"
 )
 
 // Sink is where a Controller logger will send log lines to.
@@ -81,6 +85,10 @@ func (s *Sink) Update(o SinkOptions) error {
 func writerSinkLogger(w io.Writer, o SinkOptions) (log.Logger, error) {
 	var l log.Logger
 
+	if len(o.Fanout) > 0 {
+		w = &fanoutWriter{w: w, f: o.Fanout}
+	}
+
 	switch o.Format {
 	case FormatLogfmt:
 		l = log.NewLogfmtLogger(log.NewSyncWriter(w))
@@ -95,5 +103,33 @@ func writerSinkLogger(w io.Writer, o SinkOptions) (log.Logger, error) {
 	if o.IncludeTimestamps {
 		l = log.With(l, "ts", log.DefaultTimestampUTC)
 	}
+
 	return l, nil
+}
+
+type fanoutWriter struct {
+	w io.Writer
+	f []loki.LogsReceiver
+}
+
+func (fw *fanoutWriter) Write(p []byte) (int, error) {
+	n, err := fw.w.Write(p)
+	if err != nil {
+		return n, err
+	}
+
+	for _, receiver := range fw.f {
+		select {
+		case receiver <- loki.Entry{
+			Labels: model.LabelSet{"component": "agent"},
+			Entry: logproto.Entry{
+				Timestamp: time.Now(),
+				Line:      string(p),
+			},
+		}:
+		default:
+		}
+	}
+
+	return n, nil
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -131,6 +131,9 @@ Main (unreleased)
 
 - Update Loki Dependency to k146 which includes configurable file watchers (@mattdurham)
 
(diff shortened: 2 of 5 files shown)


pkg/flow/logging/sink.go
```diff
@@ -95,5 +103,33 @@ func writerSinkLogger(w io.Writer, o SinkOptions) (log.Logger, error) {
...
+	for _, receiver := range fw.f {
+		select {
+		case receiver <- loki.Entry{
+			Labels: model.LabelSet{"component": "agent"},
+			Entry: logproto.Entry{
+				Timestamp: time.Now(),
+				Line:      string(p),
+			},
+		}:
+		default:
```
tpaschalis, line 130: This is to guard against issues like #2216, i.e. trying to send to the receiver of a component that is shutting down and getting GC'ed and its export is nil.

I'm not 100% if the logging block works similarly to components, will check one more time to make sure whether we need this.

pkg/flow/logging/sink.go
```diff
@@ -95,5 +103,33 @@ func writerSinkLogger(w io.Writer, o SinkOptions) (log.Logger, error) {
 	if o.IncludeTimestamps {
 		l = log.With(l, "ts", log.DefaultTimestampUTC)
 	}
+
 	return l, nil
 }
+
+type fanoutWriter struct {
+	w io.Writer
+	f []loki.LogsReceiver
```
tpaschalis, line 112: As far as I could see in the code, this will get recreated in-place whenever the `logging` block changes and is re-evaluated; I _think_ we don't need a mutex here.

pkg/flow/logging/sink.go
```diff
@@ -81,6 +85,10 @@ func (s *Sink) Update(o SinkOptions) error {
 func writerSinkLogger(w io.Writer, o SinkOptions) (log.Logger, error) {
 	var l log.Logger
 
+	if len(o.Fanout) > 0 {
+		w = &fanoutWriter{w: w, f: o.Fanout}
```
mattdurham, line 89: Mutating the passed in `w io.Writer` feels off. Is there a cleaner way?

erikbaranowski:
Is this the right way to set this up? I'm seeing my logs on the command line but not forwarded to loki

```river
logging {
	level  = "info"
	format = "logfmt"
	write_to = [loki.write.default.receiver]
}

loki.write "default" {
	endpoint {
		url = env("LOG_URL")
	}
}
```

erikbaranowski (approved):
LGTM and thank you from someone this will specifically help out right away!

Keep in mind we will want to reimplement this change if we decide to back out the logging refactor from #3107. Backing the logging refactor out was discussed as part of a module API refactor.

pkg/flow/logging/sink.go
tpaschalis: I'll take another look if you'd like; I tried wrapping around the logger instead but this meant we didn't have access to the log level and log format, while this makes sure the logs in the pipeline are same ones you see in the console.

mattdurham: Sounds reasonable, maybe add a comment to the funcs?

tpaschalis:
> Keep in mind we will want to reimplement this change if we decide to back out the logging refactor from https://github.com/grafana/agent/issues/3107. Backing the logging refactor out was discussed as part of a module API refactor.

I'll take a look at those conversations and see if I'd proceed in a different way.

In any case, I will hold on merging this for today, so we can keep refactor-related changes to a single release.

tpaschalis:
This approach is no longer feasible after a recent refactoring of the pkg/flow/logging package.

I'm closing this PR and come back with a new one.

PR #3575, opened 2023-04-20
[Agent Management] Move `agent_management.remote_config_cache_location`

jcreixell:
#### PR Description

  - Moves remote_config_cache_location config field under remote_configuration section.
  - Since we have a remote_config section, all fields specifically related to it should go there.

#### Notes to the Reviewer

  - Although this is a breaking change and last minute before the next rc, I think it is worth doing since we can bundle it together with other breaking changes to the config.

#### PR Checklist

- [x] CHANGELOG updated
- [x] Tests updated


spartan0x117 (approved):
LGTM!

PR #3581, opened 2023-04-20
Create shared server for `loki.source.+` network targets 

thepalbi:
#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated


component/common/loki/http/config.go
```diff
@@ -0,0 +1,10 @@
+package http
+
+import (
+	"github.com/weaveworks/common/server"
+)
+
+// Config is a wrapper around server.Config.
+type Config struct {
```
thepalbi, line 8: **incomplete**: the idea is for this `Config` object to expose a river block that is a subset of [this config](https://github.com/weaveworks/common/blob/master/server/server.go#L62). That way we can easily map it, extend it, and re-use the same kind of config block for all network loki sources. Let me know if that sounds good @tpaschalis ! 

Also, if yes, how easily can we break the river API from one version to another for things such as [loki.source.gcp](https://github.com/grafana/agent/blob/main/component/loki/source/gcplog/internal/gcplogtarget/types.go#L23)? Asking because I don't know which kind of api stability we have right now in flow

component/common/loki/http/server_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+	// assert all metrics have the prefix applied
+	metrics, err := reg.Gather()
+	require.NoError(t, err)
+	for _, m := range metrics {
+		require.True(t, strings.HasPrefix(m.GetName(), "test_namespace"))
+	}
+}
+
+func getServerConfigWithAvailablePort() (cfg server.Config, port int, err error) {
```
thepalbi, line 61: this bit is re-used from heroku and gcp tests, can be moved somewhere in common

component/common/loki/http/config.go
tpaschalis: I'd recommend that we go a slightly different way to be aligned with other Flow components work.

This struct can contain just the fields that Promtail's [server](https://grafana.com/docs/loki/next/clients/promtail/configuration/#server) block can already configure, but with River tags attached to them.

Then we can have an `UnmarshalRiver` method so that we apply defaults, performs validations etc implicitly during unmarshaling and finally a `Convert()` method so that it gets turned into a `github.com/weaveworks/common/server.Config` that the caller packages can use instead of building the config by hand. If this sounds complicated, [here's an example](https://github.com/grafana/agent/blob/6fc177a56a2f71c8980fac266bb23d78b6572f75/component/common/config/types.go#L19-L140) of doing that with our shared `HTTPClientConfig` struct. 

Then, packages like gcplog could embed this config in their Arguments, call Convert() to get the config and then call to NewTargetServer.

We could even go a step further and have the `NewTargetServer` receive the River struct itself and call Convert() inside of it. That means that caller packages don't have to construct the config themselves like we do [here](https://github.com/grafana/agent/pull/3581/files#diff-fb88003d94f0a346752ac1519c4b80e685c763f8e716a83a935934a9117143aeR43-R49) which has leaves some room for error.

tpaschalis: As for the component Arguments, I don't think there's too much hassle around changing the config, especially if it's for a good reason such as here. Flow itself as a whole is stable, but I believe that provided we document and communicate the change we should be okay.

thepalbi: Completely agree with the suggested approach. I went ahead and first renamed `Config` to `ServerConfig`, which exposes in River a some configuration blocks for tuning the underlying server. I haven't added all settings since it's a huge amount, and rather just used the ones we found the most usable for a scalable server (for http listen and port, and for grpc some more that control connections lifecycle).

Also, routed this new settings through heroku and gcp components. Let me know if we should keep backwards with this both, and we can discuss how, I'm assuming since we can break it between one release and the other.

thepalbi:
@tpaschalis still have some docs to add, and some cleanup, but let me know if this goes in the right direction

component/common/loki/http/config.go
```diff
@@ -0,0 +1,86 @@
...
+type GRPCConfig struct {
+	ListenAddress         string        `river:"listen_address,attr,optional"`
+	ListenPort            int           `river:"listen_port,attr,optional"`
+	ConnLimit             int           `river:"conn_limit,attr,optional"`
+	MaxConnectionAge      time.Duration `river:"max_connection_age,attr,optional"`
+	MaxConnectionAgeGrace time.Duration `river:"max_connection_age_grace,attr,optional"`
+	MaxConnectionIdle     time.Duration `river:"max_connection_idle,attr,optional"`
+}
+
+func (g *GRPCConfig) Config(c *server.Config) {
```
tpaschalis, line 49: nit: Maybe name this `Into`?

component/loki/source/gcplog/internal/gcplogtarget/types.go
```diff
@@ -21,22 +22,14 @@ type PullConfig struct {
...
-	PushTimeout          time.Duration     `river:"push_timeout,attr,optional"`
-	Labels               map[string]string `river:"labels,attr,optional"`
-	UseIncomingTimestamp bool              `river:"use_incoming_timestamp,attr,optional"`
-}
-
-// DefaultPushConfig sets the default listen address and port.
-var DefaultPushConfig = PushConfig{
-	HTTPListenAddress: "0.0.0.0",
-	HTTPListenPort:    8080,
+	Server               *lhttp.ServerConfig `river:"server,block,optional"`
```
tpaschalis, line 26: Right now, users have to go one level deeper just to access the `server` block
```
loki.source.gcplog "foo" {
  server { 
    http { }
    grpc { }
  }
}
```



We have the option of _squashing` the block ([here's](https://github.com/grafana/agent/blob/bec518aa9ce7913debc20e88b9d458309e9f31b0/component/prometheus/remotewrite/types.go#L77) an example) so that people can configure the two inner blocks directly, like this:
```
loki.source.gcplog "foo" {
  http { }
  grpc { }
}
```


Code-wise, you still keep using the `PushConfig.Server` block as-is, this is just syntactic sugar for people writing configuration.

component/common/loki/http/server.go
```diff
@@ -0,0 +1,94 @@
...
+		return nil, fmt.Errorf("metrics namespace is not prometheus compatiible: %s", metricsNamespace)
+	}
+
+	t := &TargetServer{
+		logger:           logger,
+		metricsNamespace: metricsNamespace,
+	}
+
+	// convert from River into the weaveworks config
+	serverCfg := config.Convert()
```
tpaschalis, line 37: I'm on the fence about this, even though there's not a super big difference right now. Do we want to call Convert() here, or ask component authors to call the Convert() method and pass you the weaveworks config directly?

Let's keep is as-is for now but keep it in mind in case we want to change it in the future.

tpaschalis (commented):
I think we're heading to the right direction. Looking forward to seeing how this progresses!

component/common/loki/http/config.go
```diff
@@ -0,0 +1,86 @@
...
+
+import (
+	"flag"
+	"github.com/weaveworks/common/server"
+	"time"
+)
+
+// ServerConfig is a wrapper around server.ServerConfig.
+type ServerConfig struct {
+	// HTTP configures the HTTP server. Note that despite the blog being present or not,
```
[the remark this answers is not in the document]
thepalbi: Solved in https://github.com/grafana/agent/pull/3581/commits/3fe2c07c4583b84d672aeb8fb7372392d0576007

component/common/loki/http/config.go
thepalbi: Like it

component/common/loki/http/server.go
thepalbi: I like the idea of TargetServer taking directly the River-based configuration, since it's a customized version of the WeaveWorks server. Also, by taking directly the River config we could narrow down the configs we expose, weaveworks model is rather complex.

About the component authors doing the convert, you mean so that they are able to tune some other param our river model doesn't expose?

component/loki/source/gcplog/internal/gcplogtarget/types.go
thepalbi: Uh like it! Using squash since a three level deep configurations looks annoying.

component/common/loki/http/config.go
```diff
@@ -0,0 +1,86 @@
...
+package http
+
+import (
+	"flag"
+	"github.com/weaveworks/common/server"
+	"time"
+)
+
+// ServerConfig is a wrapper around server.ServerConfig.
+type ServerConfig struct {
```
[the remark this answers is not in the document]
thepalbi: I think it's a good idea to expose fields defined in the Promtail server config, which just a subset applies here since there's things like `disabled`, `enable_runtime_reload` and other that don't apply to this case.
Exposing the whole of [weaveworks config](https://github.com/weaveworks/common/blob/master/server/server.go#L62) is too much in my opinion, and will be a PITA for users to configure through river.

I think a middle ground could be:
1. We expose the configs that are in the Promtail doc and apply to our case, which are:
```
http {
    [http_listen_address: <string>]
    [http_listen_port: <int> | default = 80]
    [http_server_read_timeout: <duration> | default = 30s]
    [http_server_write_timeout: <duration> | default = 30s]
    [http_server_idle_timeout: <duration> | default = 120s]
    [http_path_prefix: <string>] // could be ommited
}
grpc {
    [grpc_listen_address: <string>]
    [grpc_listen_port: <int> | default = 9095]
    [grpc_server_max_recv_msg_size: <int> | default = 4194304]
    [grpc_server_max_send_msg_size: <int> | default = 4194304]
    [grpc_server_max_concurrent_streams: <int> | default = 100]
}
[register_instrumentation: <boolean> | default = true] // can be ommited, since metrics are registered in the registered passed to target server
[graceful_shutdown_timeout: <duration> | default = 30s]
[log_level: <string> | default = "info"]
```
classified in their respective sub-block.

2. We do what you mentioned [here](https://github.com/grafana/agent/pull/3581#discussion_r1175542719) or requiring as argument for `NewTargetServer` the weaveworks config, so that component authors are able to tweak some specific setting available in weaveworks, but no in promtail. 
3. 
4. wdyt?

component/loki/source/heroku/heroku.go
```diff
@@ -30,38 +30,13 @@ func init() {
...
-}
-
-// DefaultListenerConfig provides the default arguments for a heroku listener.
-var DefaultListenerConfig = ListenerConfig{
-	ListenAddress: "0.0.0.0",
-}
-
-// UnmarshalRiver implements river.Unmarshaler.
-func (lc *ListenerConfig) UnmarshalRiver(f func(interface{}) error) error {
-	*lc = DefaultListenerConfig
```
thampiotr, line 54: Removing this default should have no effect, because we would be passing an empty string as an address to weaveworks server, which has the same meaning as `0.0.0.0` - is my understanding correct here?

component/common/loki/http/server.go
```diff
@@ -0,0 +1,94 @@
+package http
```
thampiotr, line 1: The `TargetServer` runs both GRPC and HTTP, so this package may need a rename?
AFAIU we will be using and documenting the GRPC endpoints too in the future.

component/loki/source/gcplog/internal/gcplogtarget/push_target.go
thampiotr, line 1: Note that these files diverge now a bit more from their upstream source. I'm not sure how this could be avoided though, maybe the mindset we should take is that actual request handlers should remain verbatim copies, but creation and lifecycle management of the server will need to be adapted to agent's needs.

thampiotr (approved):
LGMT, though I'm pretty new to this codebase.

I'm looking forward to this merging so I can incorporate the changes to my PR https://github.com/grafana/agent/pull/3648

component/loki/source/heroku/heroku.go
```diff
@@ -30,38 +30,13 @@ func init() {
 // Arguments holds values which are used to configure the loki.source.heroku
 // component.
 type Arguments struct {
-	HerokuListener       ListenerConfig      `river:"listener,block"`
```
thampiotr, line 33: Some of these changes will require documentation updates too.

thampiotr (commented):
Got a few more questions as I was working on docs for a new component.

component/common/loki/http/config.go
```diff
@@ -0,0 +1,87 @@
+// Package http contains a River serializable definition of the weaveworks server config in
+// https://github.com/weaveworks/common/blob/master/server/server.go#L62.
+package http
+
+import (
+	"flag"
+	"time"
+
+	"github.com/weaveworks/common/server"
```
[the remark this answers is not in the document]
thepalbi: Uhh I really like this, applying

component/common/loki/http/server.go
thepalbi: Yes, it's actually the intention of this PR to expose the grpc configs

component/common/loki/http/config.go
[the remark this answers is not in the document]
tpaschalis: Okay, let's start with these fields.

Some comments:
* As you said, let's remove the register_instrumentation and have it always be false. This will not be the 'main' flow HTTP server, and if we're passing the Registerer from the parent component, the metrics will be available on the root /metrics endpoint either way
* Let's use another default port for HTTP (9090 maybe?). Port 80 means that the server will need to run with superuser privileges which doesn't have to be the default

component/loki/source/heroku/heroku.go
```diff
@@ -30,38 +30,13 @@ func init() {
...
 	Labels               map[string]string   `river:"labels,attr,optional"`
 	UseIncomingTimestamp bool                `river:"use_incoming_timestamp,attr,optional"`
 	ForwardTo            []loki.LogsReceiver `river:"forward_to,attr"`
 	RelabelRules         flow_relabel.Rules  `river:"relabel_rules,attr,optional"`
 }
 
-// ListenerConfig defines a heroku listener.
-type ListenerConfig struct {
-	ListenAddress string `river:"address,attr,optional"`
-	ListenPort    int    `river:"port,attr"`
```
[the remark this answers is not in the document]
thepalbi: Yhea, I guess if we remove the `optional` in the river config tags it should make the field required. What do you think of making the http port required, and the grpc optional (since I think the usual case is to use HTTP)?

component/loki/source/gcplog/internal/gcplogtarget/push_target.go
thepalbi: With this PR the only moving piece in each of `push_target` and heroku is the endpoint handler, therefore to update we can just copy that func. For example for push_target it'd be [this one](https://github.com/grafana/loki/blob/main/clients/pkg/promtail/targets/gcplog/push_target.go#L103) + the other package files.

component/loki/source/heroku/heroku.go
thepalbi: Yup, weavewowrks applies no default, hence the resulting address in the [`net.Listen`](https://pkg.go.dev/net#Listen) call to get the listen socket is `:PORT`. AFAIK an empty listen address is the same as `0.0.0.0`, which is listening for every IP address.

component/loki/source/heroku/heroku.go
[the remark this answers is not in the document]
thepalbi: As discussed [here](https://github.com/grafana/agent/pull/3581#discussion_r1179284020), using a sane default for both http and grpc ports, so it's easy for a user to not configure the server at all and get it working.

thepalbi:
@tpaschalis @thampiotr added all settings to the config model as discussed above, and added some tests covering the river unmarshalling. Will add the docs tomorrow

docs/sources/flow/reference/components/loki.server.md
thepalbi, line 1: no sure of how to call this file. I'd like for us to keep the docs page on the common server component in a diff section, instead of copying in each component. Maybe the docs system allows us to keep it as a snippet and embed it in each necessary page? I think that'd be a better solution

docs/sources/flow/reference/components/loki.source.gcplog.md
```diff
@@ -80,19 +80,22 @@ a service account key.
...
 fields take their default values.
 
-Name                     | Type          | Description | Default | Required
------------------------- | ------------- | ----------- | ------- | --------
-`http_listen_address`    | `string`      | The address the server listens to.  | `"0.0.0.0"` | no
-`http_listen_port`       | `int`         | The port the server listens to.     |  `8080`     | no
+Name                     | Type          | Description                                                     | Default | Required
+------------------------ |---------------|-----------------------------------------------------------------| ------- | --------
+`http`    | [HTTP][]      | Configures the HTTP server that receives requests.              |  | no
+`grpc`    | [GRPC][]      | Configures the gRPC server that receives requests.              |  | no
```
thepalbi, line 92: Only drawback is we'll had to copy this in each config using it

thepalbi:
@tpaschalis @thampiotr added the docs, leaving some Qs on how to best organize them

docs/sources/flow/reference/components/loki.server.md
thampiotr: I think you can import a file using something like this: https://github.com/grafana/agent/blob/dedba9bb3731ca4a138be9a99b0d06c6291a5c23/docs/sources/flow/reference/components/loki.relabel.md#L63

tpaschalis: As Piotr said, let's put this on `docs/sources/shared/flow/reference/components/loki-server.md`

Afterwards, we'll be able to refer to it using a shorthand like the one above.

docs/sources/flow/reference/components/loki.server.md
```diff
@@ -0,0 +1,86 @@
...
+}
+```
+
+## Arguments
+
+`loki.server` supports the following arguments:
+
+ Name                        | Type           | Description                               | Default | Required 
+-----------------------------|----------------|-------------------------------------------|---------|----------
+ `graceful_shutdown_timeout` | `RelabelRules` | Relabeling rules to apply on log entries. | "{}"    | no
```
tpaschalis, line 34: This should be removed, right? There's no top-level arguments defined on the block.

docs/sources/flow/reference/components/loki.server.md
```diff
@@ -0,0 +1,86 @@
...
+The following blocks are supported inside the definition of
+`loki.source.gcplog`:
+
+ Hierarchy | Name     | Description                 | Required 
+-----------|----------|-----------------------------|----------
+ http      | [http][] | Configures the HTTP server. | no       
+ push      | [grpc][] | Configures the gRPC server. | no       
+
+The `pull` and `push` inner blocks are mutually exclusive; a component must
+contain exactly one of the two in its definition.
```
tpaschalis, line 47:
```suggestion
 grpc      | [grpc][] | Configures the gRPC server. | no       
```

docs/sources/flow/reference/components/loki.server.md
```diff
@@ -0,0 +1,86 @@
...
+
+The `http` configures the HTTP server.
+
+The following arguments can be used to configure the `http` block. Any omitted
+fields take their default values.
+
+ Name                   | Type       | Description                                                                                                          | Default  | Required 
+------------------------|------------|----------------------------------------------------------------------------------------------------------------------|----------|----------
+ `listen_address`       | `string`   | Network address on which the server will listen for new connections. Defaults to accepting all incoming connections. | `""`     | no       
+ `listen_port`          | `int`      | Port number on which the server will listen for new connections.                                                     | `9090`   | no
```
tpaschalis, line 63: By default, Prometheus listens to :9090.

I'd say we change this and the gRPC default port to 8080/8081 which should be a tad safer.

component/common/loki/net/config.go
```diff
@@ -0,0 +1,111 @@
...
+import (
+	"flag"
+	"time"
+
+	weaveworks "github.com/weaveworks/common/server"
+)
+
+const (
+	DefaultHTTPPort = 9090
+	DefaultGRPCPort = 9091
```
tpaschalis, line 14: Prometheus listens to 9090 by default. While rare, it _may_ trip up some users?

How do you feel about using 8080/8081 here?

docs/sources/flow/reference/components/loki.server.md
tpaschalis, line 34: Let's fix the type, default value and description here.

component/common/loki/net/config.go
thepalbi: Sure thing, updating

docs/sources/flow/reference/components/loki.server.md
thepalbi: Fixed

docs/sources/flow/reference/components/loki.server.md
thepalbi: Applied

component/loki/source/heroku/heroku.go
thepalbi: Docs added in the follow up commits

thepalbi:
@thampiotr @tpaschalis should be good now

component/common/loki/net/config.go
```diff
@@ -0,0 +1,115 @@
...
+		c.HTTP.Into(&cfg)
+	}
+	if c.GRPC != nil {
+		c.GRPC.Into(&cfg)
+	}
+	// If set, override. Don't allow a zero-value since it configure a context.WithTimeout, so the user should at least
+	// give a >0 value to it
+	if c.GracefulShutdownTimeout != 0 {
+		cfg.ServerGracefulShutdownTimeout = c.GracefulShutdownTimeout
+	}
```
tpaschalis, line 100: Just to make sure, what happens when the user does _not_ set this? Is the `"30s"` default applied? I'm not sure where exactly this happened.

component/common/loki/net/config.go
```diff
@@ -0,0 +1,115 @@
...
+	return cfg
+}
+
+// newDefaultConfig creates a new weaveworks.Config object with some overridden defaults.
+func newDefaultConfig() weaveworks.Config {
+	c := weaveworks.Config{}
+	c.RegisterFlags(flag.NewFlagSet("empty", flag.ContinueOnError))
+	// Opting by default 0, which used in net.Listen assigns a random port
```
tpaschalis, line 108: This comment is no longer up to date, right?

docs/sources/shared/flow/reference/components/loki-server-http.md
```diff
@@ -0,0 +1,19 @@
...
+---
+
+The `http` block configures the HTTP server.
+
+The following arguments can be used to configure the `http` block. Any omitted
+fields take their default values.
+
+ Name                   | Type       | Description                                                                                                          | Default  | Required
+------------------------|------------|----------------------------------------------------------------------------------------------------------------------|----------|----------
+ `listen_address`       | `string`   | Network address on which the server will listen for new connections. Defaults to accepting all incoming connections. | `""`     | no
```
tpaschalis, line 14: Is the default here `0.0.0.0`?

docs/sources/shared/flow/reference/components/loki-server-grpc.md
```diff
@@ -0,0 +1,22 @@
...
+---
+
+The `grpc` block configures the gRPC server.
+
+The following arguments can be used to configure the `grpc` block. Any omitted
+fields take their default values.
+
+ Name                            | Type       | Description                                                                                                          | Default      | Required
+---------------------------------|------------|----------------------------------------------------------------------------------------------------------------------|--------------|----------
+ `listen_address`                | `string`   | Network address on which the server will listen for new connections. Defaults to accepting all incoming connections. | `""`         | no
```
tpaschalis, line 14: Same here, does the description mean that the default is `0.0.0.0`? Or is it implied to be 0.0.0.0 by the weaveworks.Server code?

docs/sources/flow/reference/components/loki.source.gcplog.md
```diff
@@ -136,7 +147,7 @@ metrics:
 
 ## Example
 
-This example listens for GCP Pub/Sub PushRequests on `0.0.0.0:8080` and
+This example listens for GCP Pub/Sub PushRequests on `0.0.0.0:9090` and
```
tpaschalis, line 150: With the new default applied
```suggestion
This example listens for GCP Pub/Sub PushRequests on `0.0.0.0:8080` and
```

docs/sources/flow/reference/components/loki.source.gcplog.md
```diff
@@ -80,19 +80,23 @@ a service account key.
...
 fields take their default values.
 
-Name                     | Type          | Description | Default | Required
------------------------- | ------------- | ----------- | ------- | --------
-`http_listen_address`    | `string`      | The address the server listens to.  | `"0.0.0.0"` | no
-`http_listen_port`       | `int`         | The port the server listens to.     |  `8080`     | no
+Name                     | Type          | Description                                                     | Default | Required
+------------------------ |---------------|-----------------------------------------------------------------| ------- | --------
+`http`    | [HTTP][]      | Configures the HTTP server that receives requests.              |  | no
+`grpc`    | [gRPC][]      | Configures the gRPC server that receives requests.              |  | no
```
tpaschalis, line 92: Let's remove these from here as they are not arguments, and document them under the `## Blocks` header.

Their hierarchy would then be `push > http` and `push > grpc`. Also, let's make the `Type` column lowercase, like the block itself is to avoid confusion.

docs/sources/flow/reference/components/loki.source.heroku.md
```diff
@@ -52,19 +53,19 @@ The following blocks are supported inside the definition of `loki.source.heroku`
 
 Hierarchy | Name | Description | Required
 --------- | ---- | ----------- | --------
-listener | [listener] | Configures a listener for Heroku messages. | yes
+`http`    | [HTTP][]      | Configures the HTTP server that receives requests.              |  | no
+`grpc`    | [gRPC][]      | Configures the gRPC server that receives requests.              |  | no
```
tpaschalis, line 57: nit: the block names are lowercase; let's keep it 
```suggestion
`http`    | [http][]      | Configures the HTTP server that receives requests.              |  | no
`grpc`    | [grpc][]      | Configures the gRPC server that receives requests.              |  | no
```

component/common/loki/net/server.go
```diff
@@ -0,0 +1,92 @@
...
+	// Set the config to the new combined config.
+	// Avoid logging entire received request on failures
+	serverCfg.ExcludeRequestInLog = true
+	// Configure dedicated metrics registerer
+	serverCfg.Registerer = reg
+	// Persist crafter config in server
+	ts.config = &serverCfg
+	// To prevent metric collisions because all metrics are going to be registered in the global Prometheus registry.
+	ts.config.MetricsNamespace = ts.metricsNamespace
+	// We don't want the /debug and /metrics endpoints running, since this is not the main promtail HTTP server.
```
tpaschalis, line 47:
```suggestion
	// We don't want the /debug and /metrics endpoints running, since this is not the main Flow HTTP server.
```

tpaschalis (approved):
LGTM, just a few documentation-related nits and we're good to go!

component/common/loki/net/server.go
thampiotr, line 1: We will use this server also for metrics (e.g. https://github.com/grafana/agent/pull/3753) and perhaps other telemetry, so I think it would be better in a `component/common` package.

thampiotr (approved):
LGTM! Just one rename suggestion, happy for this to merge :)

component/common/loki/net/config.go
thepalbi: Yup, deleting!

docs/sources/shared/flow/reference/components/loki-server-http.md
thepalbi: Nop, is empty string which has the same effect of listening to tcp conns from every incoming ip. [Explained here](https://pkg.go.dev/net#Listen)

docs/sources/flow/reference/components/loki.source.gcplog.md
thepalbi: Done

component/common/loki/net/server.go
thepalbi: Yhea, saw your PR yesterday I think. Moving to `component/common` and deleting loki-specific comments.

thepalbi:
@thampiotr @tpaschalis should be good to go now 😃

tpaschalis:
Great work @thepalbi 🙌  thanks for making writing Loki components a lot easier!

PR #3582, opened 2023-04-21
Fix entry in CHANGELOG

marctc:
We mistyped `prometheus` instead of `otelcol` for `exporter.logging` in the CHANGELOG.


tpaschalis (approved):

tpaschalis:
Do we want to reorder them so they're alphabetical, too?

tpaschalis (approved):

PR #3584, opened 2023-04-21
Fix README doc links

tpaschalis:
_After_ the v0.33 release is out and the docs revamp gets published as `latest` we should change this link to point to the Flow installation docs instead (as our README basically is all about Flow).

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,9 +55,9 @@ Check out our [documentation][] to see:
 * The list of Grafana Agent Flow [Components][]
 
 [documentation]: https://grafana.com/docs/agent/latest/
-[Installation instructions]: https://grafana.com/docs/agent/latest/set-up/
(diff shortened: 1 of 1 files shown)


ptodev (approved):

ptodev:
I don't think we have linter checks for markdown files? If we do, it'd have been nice to run them after the docs changes to make sure there are no dead links.

tpaschalis:
> I don't think we have linter checks for markdown files? If we do, it'd have been nice to run them after the docs changes to make sure there are no dead links.

Yes we do, and it's going to trigger _after_ the release is cut and the rc is promoted to a stable tag, when the `next` docs are published as `latest. Just making the PR beforehand.

tpaschalis:
Already fixed in https://github.com/grafana/agent/pull/3639

PR #3588, opened 2023-04-21
operator: fix filter_running logic

captncraig:
Apparently null and false are different things in jsonnet. Adding this test broke the logic for anything that doesn't include the field. That means 0.33-rc0 has a broken version of operator.


spartan0x117 (approved):
:+1:

jcreixell (approved):

PR #3589, opened 2023-04-21
Revert "Add an otelcol.processor.attributes component (#3349)"

ptodev:
Removing `otelcol.processor.attributes` for now, because it does not seem to work with all signal types.

The [Test_ExcludeMulti](https://github.com/grafana/agent/blob/more_unit_tests_for_otelcol_processor_attributes/component/otelcol/processor/attributes/attributes_test.go#L489) test fails with this error:
```
services, span_names, and span_kinds are not valid for log records
```
The issue is that we [pass](https://github.com/grafana/agent/blob/main/component/otelcol/processor/processor.go#L178) the logs processor incompatible match properties. [ValidateForLogs](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.63.0/internal/coreinternal/processor/filterconfig/config.go#L188) complains because there are "services" specified, which otel doesn't expect for logs.


docs/sources/flow/reference/components/discovery.kubernetes.md
```diff
@@ -269,7 +269,7 @@ omitted, all namespaces are searched.
 Name | Type | Description | Default | Required
 ---- | ---- | ----------- | ------- | --------
 `own_namespace` | `bool`   | Include the namespace the agent is running in. | | no
-`names` | `list(string)` | List of namespaces to search. | | no
+`names` | `[]string` | List of namespaces to search. | | no
```
spartan0x117, line 272: Did we want to revert this change as well?

Suggestion in case you want to apply from GH:
```suggestion
`names` | `list(string)` | List of namespaces to search. | | no
```

docs/sources/flow/reference/components/mimir.rules.kubernetes.md
```diff
@@ -120,11 +120,11 @@ The `match_expression` block describes a Kubernetes label match expression for r
...
-Name       | Type           | Description                                        | Default | Required
------------|----------------|----------------------------------------------------|---------|---------
-`key`      | `string`       | The label name to match against.                   |         | yes
-`operator` | `string`       | The operator to use when matching.                 |         | yes
-`values`   | `list(string)` | The values used when matching.                     |         | no
+Name       | Type       | Description                                        | Default | Required
+-----------|------------|----------------------------------------------------|---------|---------
+`key`      | `string`   | The label name to match against.                   |         | yes
+`operator` | `string`   | The operator to use when matching. |         | yes
+`values`   | `[]string` | The values used when matching.                     |         | no
```
spartan0x117, line 127: Did we want to revert this change as well?

Suggestion in case you want to apply from GH:
```suggestion
Name       | Type           | Description                                        | Default | Required
-----------|----------------|----------------------------------------------------|---------|---------
`key`      | `string`       | The label name to match against.                   |         | yes
`operator` | `string`       | The operator to use when matching.                 |         | yes
`values`   | `list(string)` | The values used when matching.                     |         | no
```

spartan0x117 (approved):
:+1:

docs/sources/flow/reference/components/otelcol.processor.batch.md
```diff
@@ -112,6 +112,7 @@ otelcol.exporter.otlp "production" {
     endpoint = env("OTLP_SERVER_ENDPOINT")
   }
 }
+
```
ptodev, line 115:
```suggestion
```

PR #3600, opened 2023-04-21
Do not use hard-coded `kill` command path

mohammed90:
#### PR Description

The current systemd unit file specifies `/usr/bin/kill` for the `kill` command. The command isn't always placed in that location. It's best to call the command using `/usr/bin/env kill` to always receive the path to the actual location of the command.

N/A

#### PR Checklist

- [ ] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated


rfratto (approved):
Sorry it took so long to review this. LGTM, thanks for the contribution!

PR #3604, opened 2023-04-23
Added coalesce function

### LINKED ISSUE #3389: Proposal: add optional second argument to env() stdlib function

jkroepke:
#### PR Description

Adding an coalesce function as described in https://github.com/grafana/agent/pull/3540#issuecomment-1511352098

Fixes #3389
Replaces #3540 

#### Notes to the Reviewer

I currently, it's not tested, if updates from components like `local.file` or `remote.http` are passed. 

For example 

```
local.file "blackbox_exporter_config" {
  filename  = "/etc/grafana-agent/blackbox_exporter.yaml"
}

prometheus.exporter.blackbox "this" {
  config = coalesce(local.file.blackbox_exporter_config.content, "{ modules: { http_2xx: { prober: http, timeout: 5s } } }")
  
  target "TARGET_NAME" {
    address = "TARGET_ADDRESS" 
  }
}

```

If `local.file.blackbox_exporter_config.content` switches from empty string to content, the `coalesce` function needs to aware of this. Not sure, if river functions supporting this.

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/flow/reference/stdlib/coalesce.md
+++ b/docs/sources/flow/reference/stdlib/coalesce.md
@@ -6,8 +6,7 @@ title: coalesce
 
 # coalesce
 
-`coalesce` takes any number of arguments and returns the first one that isn't null, an empty string, empty list or empty object. It can
-be combined with other functions to provide a default.
+`coalesce` takes any number of arguments and returns the first one that isn't null, an empty string, empty list, or an empty object. It is useful for obtaining a default value, such as if an environment variable isn't defined.
 
 ## Examples
 
--- a/docs/sources/flow/reference/stdlib/coalesce.md
+++ b/docs/sources/flow/reference/stdlib/coalesce.md
@@ -1,0 +1,21 @@
+---
+aliases:
+- ../../configuration-language/standard-library/coalesce/
+title: coalesce
+---
+
+# coalesce
+
+`coalesce` takes any number of arguments and returns the first one that isn't null, an empty string, empty list or empty object. It can
+be combined with other functions to provide a default.
+
+## Examples
+
+```
+> coalesce("a", "b")
+a
+> coalesce("", "b")
+b
+> coalesce(env("DOES_NOT_EXIST"), "c")
+c
+```
--- a/docs/sources/flow/reference/stdlib/coalesce.md
+++ b/docs/sources/flow/reference/stdlib/coalesce.md
@@ -6,7 +6,9 @@ title: coalesce
 
 # coalesce
 
-`coalesce` takes any number of arguments and returns the first one that isn't null, an empty string, empty list, or an empty object. It is useful for obtaining a default value, such as if an environment variable isn't defined.
+`coalesce` takes any number of arguments and returns the first one that isn't null, an empty string, empty list, or an empty object.
+It is useful for obtaining a default value, such as if an environment variable isn't defined.
+The last argument is always returned, if there is no other non-empty or non-zero argument.
 
 ## Examples
 
--- a/docs/sources/flow/reference/stdlib/coalesce.md
+++ b/docs/sources/flow/reference/stdlib/coalesce.md
@@ -8,7 +8,7 @@ title: coalesce
 
 `coalesce` takes any number of arguments and returns the first one that isn't null, an empty string, empty list, or an empty object.
 It is useful for obtaining a default value, such as if an environment variable isn't defined.
-The last argument is always returned, if there is no other non-empty or non-zero argument.
+If no argument is non-empty or non-zero, the last argument is returned. 
 
 ## Examples
 
--- a/pkg/river/internal/stdlib/stdlib.go
+++ b/pkg/river/internal/stdlib/stdlib.go
@@ -73,4 +73,18 @@ var Identifiers = map[string]interface{}{
 		}
 		return res, nil
 	},
+
+	"coalesce": value.RawFunction(func(funcValue value.Value, args ...value.Value) (value.Value, error) {
+		for _, arg := range args {
+			if !arg.Reflect().IsZero() {
+				if argType := value.RiverType(arg.Reflect().Type()); (argType == value.TypeArray || argType == value.TypeObject) && arg.Len() == 0 {
+					continue
+				}
+
+				return arg, nil
+			}
+		}
+
+		return value.Null, nil
+	}),
 }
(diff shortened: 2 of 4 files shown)


rfratto:
I'm in favor of the function, but I'm on the fence about calling it coalesce. 

I know that's what's commonly used in Excel and SQL, but I'm not sure how widely understood it is (at least I had never heard of it before @mattdurham mentioned it)
 
Are there other ideas for what we may be able to call this? Or is coalesce the most descriptive we can get?

jkroepke:
I do not have any opinion about the name coalesce. What I know is that coalesce is used in many programs as name:

* https://developer.hashicorp.com/terraform/language/functions/coalesce
* https://www.postgresqltutorial.com/postgresql-tutorial/postgresql-coalesce/
* https://www.rdocumentation.org/packages/hutils/versions/1.8.1/topics/coalesce

or used as language operator (null coalescing operator)

* https://en.wikipedia.org/wiki/Null_coalescing_operator

An other idea would be something like `default`, `isempty` or `iszero`

pkg/river/internal/stdlib/stdlib.go
```diff
@@ -73,4 +73,18 @@ var Identifiers = map[string]interface{}{
 		}
 		return res, nil
 	},
+
+	"coalesce": value.RawFunction(func(funcValue value.Value, args ...value.Value) (value.Value, error) {
+		for _, arg := range args {
+			if !arg.Reflect().IsZero() {
+				if argType := value.RiverType(arg.Reflect().Type()); (argType == value.TypeArray || argType == value.TypeObject) && arg.Len() == 0 {
```
mattdurham, line 80: I think we need to add some conditions for `null`

To handle `coalesce([],null)`, I doubt anyone will slam null in there but its possible exports from other components could be null. 

```
if arg.Type() == value.TypeNull {
				continue
			}
```

pkg/river/internal/stdlib/stdlib.go
```diff
@@ -73,4 +73,18 @@ var Identifiers = map[string]interface{}{
...
+					continue
+				}
+
+				return arg, nil
+			}
+		}
+
+		return value.Null, nil
```
mattdurham, line 88: Propose always returning the last value even if null/Zero Value. Functionally its very similar except in the case of arrays and maps.

rfratto:
> I do not have any opinion about the name coalesce. What I know is that coalesce is used in many programs as name:
> 
> * https://developer.hashicorp.com/terraform/language/functions/coalesce
> * https://www.postgresqltutorial.com/postgresql-tutorial/postgresql-coalesce/
> * https://www.rdocumentation.org/packages/hutils/versions/1.8.1/topics/coalesce
> 
> or used as language operator (null coalescing operator)
> 
> * https://en.wikipedia.org/wiki/Null_coalescing_operator
> 
> An other idea would be something like `default`, `isempty` or `iszero`

🤷 Alright, maybe this is just a gap in my knowledge :) If Terraform calls it coalesce, I think it's fine here too (though personally I think `default` is easier to grok).

mattdurham:
I wrote a whole comment and then closed the tab =( 

I like `coalesce`  but come from that background. I don't mind `default` verbiage wise, but I want to reserve that keyword for a later date.

tpaschalis:
Agree with Matt, and for my 2c, since we're borrowing from HCL/Terraform, having them also use coalesce is a plus for that term.

rfratto:
> For example
>
> ```
> local.file "blackbox_exporter_config" {
>  filename  = "/etc/grafana-agent/blackbox_exporter.yaml"
> }
> 
> prometheus.exporter.blackbox "this" {
>   config = coalesce(local.file.blackbox_exporter_config.content, "{ modules: { http_2xx: { prober: http, timeout: 5s } } }")
>
>   target "TARGET_NAME" {
>     address = "TARGET_ADDRESS" 
>   }
> }
> ```
>
> If local.file.blackbox_exporter_config.content switches from empty string to content, the coalesce function needs to aware of this. Not sure, if river functions supporting this.


Sorry for not responding to this earlier. Flow detects if a dependency changes its output, and will reevaluate the entire block; this means that it will re-call the `coalesce` function if `local.file.blackbox_exporter_config.content` and use the new value. River doesn't need to do any special handling here.

docs/sources/flow/reference/stdlib/coalesce.md
```diff
@@ -0,0 +1,21 @@
+---
+aliases:
+- ../../configuration-language/standard-library/coalesce/
+title: coalesce
+---
+
+# coalesce
+
+`coalesce` takes any number of arguments and returns the first one that isn't null, an empty string, empty list or empty object. It can
+be combined with other functions to provide a default.
```
rfratto, line 10:
```suggestion
`coalesce` takes any number of arguments and returns the first one that isn't null, an empty string, empty list, or an empty object. It is useful for obtaining a default value, such as if an environment variable isn't defined.
```

rfratto (approved):
LGTM, but I'll wait for @mattdurham to review the implementation as well.

Can you add this to the changelog as an enhancement?

docs/sources/flow/reference/stdlib/coalesce.md
mattdurham, line 10: Since we added that it will always return the last element even if null, then we should document that here.

docs/sources/flow/reference/stdlib/coalesce.md
```diff
@@ -0,0 +1,22 @@
...
+aliases:
+- ../../configuration-language/standard-library/coalesce/
+title: coalesce
+---
+
+# coalesce
+
+`coalesce` takes any number of arguments and returns the first one that isn't null, an empty string, empty list, or an empty object.
+It is useful for obtaining a default value, such as if an environment variable isn't defined.
+The last argument is always returned, if there is no other non-empty or non-zero argument.
```
rfratto, line 11: I can't remember the link to the rule for this, but I've been told at some point to always make statements in the form of "If CONDITION, then STATEMENT" rather than "STATEMENT if CONDITION":

```suggestion
If no argument is non-empty or non-zero, the last argument is returned. 
```

rfratto: Found it: https://developers.google.com/style/sentence-structure

jkroepke: Thanks, good to know.

mattdurham:
Awesome and thanks a ton! Merging now

PR #3605, opened 2023-04-24
flow: Parse default clustering port from --server.http.listen-addr

tpaschalis:
#### PR Description
The previous implementation had an oversight where the default port was only set from the `--cluster.advertise-address` flag, _if_ it was provided.

That meant that when using a DNS name as `--cluster.join-addresses`, it was expected that it the other nodes would be reacheable on port 80.

This PR adds a new CLI flag that appends a default port to the advertise and join addresses, if not one is explicitly defined.

Also, as we expect nodes to have their port numbers _mostly_ aligned to prevent networking issues, we continue using the advertise address to set the default port, if it is provided on it.

No issue filed

#### Notes to the Reviewer
Nothing to note. Tested this using our helm charts, but I should invest in getting some testing in the near future.

#### PR Checklist

- [ ] CHANGELOG updated (N/A)
- [X] Documentation added
- [ ] Tests updated (N/A)

### DIFF
--- a/cmd/internal/flowmode/cmd_run.go
+++ b/cmd/internal/flowmode/cmd_run.go
@@ -150,7 +150,7 @@ func (fr *flowRun) Run(configFile string) error {
 	reg := prometheus.DefaultRegisterer
 	reg.MustRegister(newResourcesCollector(l))
 
-	clusterer, err := cluster.New(l, fr.clusterEnabled, fr.clusterAdvAddr, fr.clusterJoinAddr)
+	clusterer, err := cluster.New(l, fr.clusterEnabled, fr.httpListenAddr, fr.clusterAdvAddr, fr.clusterJoinAddr)
 	if err != nil {
 		return fmt.Errorf("building clusterer: %w", err)
 	}
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -115,28 +115,25 @@ func getJoinAddr(addrs []string, in string) []string {
 }
 
 // New creates a Clusterer.
-func New(log log.Logger, clusterEnabled bool, addr, joinAddr string) (*Clusterer, error) {
+func New(log log.Logger, clusterEnabled bool, listenAddr, advertiseAddr, joinAddr string) (*Clusterer, error) {
 	// Standalone node.
 	if !clusterEnabled {
-		return &Clusterer{Node: NewLocalNode(addr)}, nil
+		return &Clusterer{Node: NewLocalNode(listenAddr)}, nil
 	}
 
 	gossipConfig := DefaultGossipConfig
 
 	defaultPort := 80
-	if addr != "" {
-		host, portStr, err := net.SplitHostPort(addr)
-		if err != nil {
-			return nil, err
-		}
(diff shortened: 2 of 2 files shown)


docs/sources/flow/reference/cli/run.md
```diff
@@ -71,3 +72,9 @@ other agent nodes; if this is empty it will attempt to find a suitable address
 to advertise from a list of default network interfaces. The agent must be
 reachable over HTTP on this address as communication happens over the agent's
 HTTP server.
+
+The `--cluster.port` value will be appended to the
```
ptodev, line 76: I might be missing something obvious, but I don't understand why this separate cmd argument is needed. If the default port 80 is not good enough for some use cases, then users can just specify a different port in `--cluster.advertise-address` and `--cluster.join-addresses`? Whether a DNS is used or not doesn't really matter?

In fact, I even feel like having an extra argument makes things more complicated because then it's not obvious which port was used - the one from `--cluster.port`, or the one included in the address.

If this change really has to go in, then I'd rename `--cluster.port` to `--cluster.default-port` to make it more clear that it is the default setting and will be overridden by an address-specific setting.

ptodev: Ok thank you, I get it now. `--cluster.default-port` would make more sense to me, since it makes it clear that this is a default which could be overridden by a more specific address. 

However, I'm am not sure if we should make this change. The problem is, even if the agent tries some default ports and end up connecting to a cluster, how does it know it is the correct cluster? It'd be most safe if each address in `--cluster.join-addresses` also contains a port to avoid ambiguity. If there is no port, IMO a default port should be non-configurable to avoid issues with differently-configured agents with different default ports. If the port is very important and the default is not suitable, the port should be explicitly mentioned in `--cluster.join-addresses`. 

So personally I'd either make the port mandatory in `--cluster.join-addresses`, or make it non-configurable as it is now.

I could be wrong, but the way I see clustering working in production is via:
1. A known set of `--cluster.join-addresses`. Those are agents started via explicitly setting their `--cluster.join-addresses`
2. Additional Agents started which use a default `--cluster.advertise-address` to  connect to this known set of Agents. 

So I don't see a strong need to also try guess the ports on which to join a cluster.

tpaschalis: > So I don't see a strong need to also try guess the ports on which to join a cluster.

There's a third option which this tries to cover, which maybe I didn't make it apparent here.

> It'd be most safe if each address in `--cluster.join-addresses` also contains a port to avoid ambiguity

Right now, we're going to try and make clustering work via a Kubernetes headless service; basically a name you perform a SRV DNS lookup on and use the discovered IP addresses as your peers. Problem is that this name _does not_ allow for a port to be defined, so it's a way to define it externally, which cannot happen if the advertise address does _also_ not define a port.

You _could_ hack around it to, by parsing `dnsname:port` and then using them separately but I'm not sure if it's a much better solution

mattdurham: IMO for testing of clustering I could see `two agents with same ip and different ports`

ptodev (approved):

PR #3615, opened 2023-04-24
Draft: add bespoke map handling

### LINKED ISSUE #3565: Flow Modules: pointer export incorrectly converted to a value when used as an argument
Background: This issue occurs in a very specific set of circumstances. I will show a couple working examples followed by 1 that demonstrates the issue. After that I have some additional technical details on the error and a possible hint on where things are having issues in the code.

# Working Example 1

In this example we get the receiver as an argument to the module loader. Working as expected!

parent.river
```river
	endpoint {
		url = env("METRICS_URL")

		basic_auth {
			username = env("METRICS_USERNAME")
			password = env("METRICS_PASSWORD")
		}
	}
}

module.file "agent_metrics" {
	filename = env("AGENT_CONFIG_FOLDER") + "/metrics/prometheus_scrape/module.river"
	arguments {
		address  = "http://0.0.0.0:12345/metrics"
		receiver = prometheus.remote_write.default.receiver
	}
}
```

module.river
```river
argument "address" {
    optional = false
}

argument "receiver" {
    optional = false
}

argument "scrape_interval" {
    optional = true
    default  = "10s"
}

prometheus.scrape "default" {
	targets         = [{"__address__" = argument.address.value}]
	forward_to      = [argument.receiver.value]
	scrape_interval = argument.scrape_interval.value
}
```

# Working Example 2

In this example we get the receiver as the export of another module loader. Working as expected!

parent.river
```river
module.file "agent_metrics" {
	filename = env("AGENT_CONFIG_FOLDER") + "/metrics/prometheus_scrape/module.river"
	arguments {
		address  = "http://0.0.0.0:12345/metrics"
	}
}
```

prometheus_scrape/module.river
```river
argument "address" {
    optional = false
}

argument "scrape_interval" {
    optional = true
    default  = "10s"
}

prometheus.scrape "default" {
	targets         = [{"__address__" = argument.address.value}]
        forward_to      = [module.file.prometheus_receiver.exports.receiver]
	scrape_interval = argument.scrape_interval.value
}

module.file "prometheus_receiver" {
	filename = env("AGENT_CONFIG_FOLDER") + "/metrics/prometheus_receiver/module.river"
	arguments {
		username = env("METRICS_USERNAME")
		password = env("METRICS_PASSWORD")
		url      = env("METRICS_URL")
	}
}
```

prometheus_receiver/module.river
```river
argument "username" {
	optional = false
}

argument "password" {
	optional = false
}

argument "url" {
	optional = false
}

export "receiver" {
	value = prometheus.remote_write.default.receiver
}

	endpoint {
		url = argument.url.value

		basic_auth {
			username = argument.username.value
			password = argument.password.value
		}
	}
}
```

# Broken Example

In this example we do a combination of the above examples. From the parent config we will get the receiver as an export from a module loader then pass it as an argument to a different module loader.

parent.river
```river
module.file "agent_metrics" {
	filename = env("AGENT_CONFIG_FOLDER") + "/metrics/prometheus_scrape/module.river"
	arguments {
		address  = "http://0.0.0.0:12345/metrics"
		receiver = module.file.prometheus_receiver.exports.receiver
	}
}

module.file "prometheus_receiver" {
	filename = env("AGENT_CONFIG_FOLDER") + "/metrics/prometheus_receiver/module.river"
	arguments {
		username = env("METRICS_USERNAME")
		password = env("METRICS_PASSWORD")
		url      = env("METRICS_URL")
	}
}
```

prometheus_scrape/module.river
```river
argument "address" {
    optional = false
}

argument "receiver" {
    optional = false
}

argument "scrape_interval" {
    optional = true
    default  = "10s"
}

prometheus.scrape "default" {
	targets         = [{"__address__" = argument.address.value}]
	forward_to      = [argument.receiver.value]
	scrape_interval = argument.scrape_interval.value
}
```

prometheus_receiver/module.river
```river
argument "username" {
	optional = false
}

argument "password" {
	optional = false
}

argument "url" {
	optional = false
}

export "receiver" {
	value = prometheus.remote_write.default.receiver
}

	endpoint {
		url = argument.url.value

		basic_auth {
			username = argument.username.value
			password = argument.password.value
		}
	}
}
```

Error starting the agent:

```Error: module.file.agent_metrics:26:21: argument.receiver.value expected capsule("storage.Appendable"), got capsule("prometheus.Interceptor")```

# Debugging

Diving in a little deeper into the code, `prometheus.remote_write` exports `receiver` as `*prometheus.Interceptor` not `prometheus.Interceptor`. Somewhere along the path it is getting translated from a pointer into its value and causing the error.

I stepped through the code in `pkg/flowinternal/controller/component.go` and confirmed that when evaluate is being called it does in fact incorrectly set the argument to the value instead of the pointer in this case.

image

### PR DESCRIPTION
**Pull request by mattdurham**
#### PR Description

This adds custom handling the case of `map[string]any`, this may not be the best solution but wanted to gather some feedback on it.

Fixes #3565

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [X] Tests updated

### DIFF
--- a/pkg/river/internal/value/decode.go
+++ b/pkg/river/internal/value/decode.go
@@ -77,7 +77,6 @@ func (d *decoder) decode(val Value, into reflect.Value) error {
 	if rawValue.CanAddr() {
 		rawValue = rawValue.Addr()
 	}
-
 	// Fully deference into and allocate pointers as necessary.
 	for into.Kind() == reflect.Pointer {
 		// Check for direct assignments before allocating pointers and dereferencing.
@@ -99,7 +98,6 @@ func (d *decoder) decode(val Value, into reflect.Value) error {
 		}
 		into = into.Elem()
 	}
-
 	// Ww need to preform the same switch statement as above after the loop to
 	// check for direct assignment one more time on the fully deferenced types.
 	//
@@ -108,6 +106,11 @@ func (d *decoder) decode(val Value, into reflect.Value) error {
 	switch {
 	case into.CanSet() && val.Type() == TypeNull:
 		into.Set(reflect.Zero(into.Type()))
+		return nil
+	// This handles the case of map[string]any, which fails on canDirectlyAssign.
+	// This is not great because it is a bespoke solution.
+	case into.CanSet() && into.Kind() == reflect.Map && reflect.TypeOf(into.Interface()).Elem() == goAny && d.canDirectlyAssignWithMap(val.rv.Type(), into.Type()):
+		into.Set(val.rv)
 		return nil
 	case into.CanSet() && d.canDirectlyAssign(val.rv.Type(), into.Type()):
 		into.Set(val.rv)
@@ -229,6 +232,20 @@ func (d *decoder) canDirectlyAssign(from reflect.Type, into reflect.Type) bool {
 		return false
 	}
 	return !containsAny(into)
+}
+
+// canDirectlyAssign returns true if the `from` type can be directly asssigned
+// to the `into` type. This always returns false if the decoder is set to make
+// copies or into contains an interface{} type anywhere in its type definition
+// to allow for decoding interfaces{} into a set of known types.
+func (d *decoder) canDirectlyAssignWithMap(from reflect.Type, into reflect.Type) bool {
+	if d.makeCopy {
+		return false
+	}
+	if from != into {
+		return false
+	}
+	return true
 }
 
 // containsAny recursively traverses through into, returning true if it
--- a/pkg/river/internal/value/decode_test.go
+++ b/pkg/river/internal/value/decode_test.go
@@ -641,3 +641,39 @@ func TestDecode_SquashedSlice_Pointer(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, expect, out)
 }
+
+func TestPointerToValue(t *testing.T) {
+	type fss struct{}
+	f := &fss{}
+	mm := make(map[string]any)
+	mm["t"] = f
+	out := make(map[string]any)
+	err := value.Decode(value.Encode(mm), &out)
+	require.NoError(t, err)
+	require.True(t, mm["t"] == out["t"])
+}
+
+func TestPointerToValueSpecific(t *testing.T) {
+	type fss struct{}
+	f := &fss{}
+	mm := make(map[string]*fss)
+	mm["t"] = f
+	out := make(map[string]*fss)
+	err := value.Decode(value.Encode(mm), &out)
+	require.NoError(t, err)
+	require.True(t, mm["t"] == out["t"])
+}
+
+func TestMapIntoNumber(t *testing.T) {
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Review by rfratto — COMMENTED** [review]
This PR breaks an [API guarantee](https://github.com/grafana/agent/blob/f954c56a9c90581a07672ee3c9c8c679a39312b1/pkg/river/river.go#L262-L272) where `any` has known values:

> ```go 
> // To unmarshal River into an interface value, Unmarshal stores one > of the
> // following:
> //
> //   - bool, for River bools
> //   - float64, for River numbers
> //   - string, for River strings
> //   - []interface{}, for River arrays
> //   - map[string]interface{}, for River objects
> //
> // Capsule and function types will retain their original type when > decoding
> // into an interface value.
> ```

You can validate that this breaks this guarantee by using this test:

```go 
func TestKnownValuesAny(t *testing.T) {
	in := map[string]any{
		"a": int8(1),
		"b": int16(2),
		"c": float32(4),
	}

	var actual map[string]any
	err := value.Decode(value.Encode(in), &actual)
	require.NoError(t, err)

	expect := map[string]any{
		"a": int(1),
		"b": int(2),
		"c": float64(4),
	}
	require.Equal(t, expect, actual)
}
```

This passes as expected when running against main. Unfortunately, the test you added checks that the _values_ match, but what we care about at the API level is the types.

mattdurham:
I am likely missing something but reading that shouldn't `a` and `b` be typed to `float64` on the expect?

mattdurham:
Converting floating point numbers offer a bit of ambiguity.

IE this test
```
func TestKnownValuesAny(t *testing.T) {
	in := map[string]any{
		"a": int8(1),
		"b": int16(2),
		"c": float32(4.2),
	}

	var actual map[string]any
	err := value.Decode(value.Encode(in), &actual)
	require.NoError(t, err)

	expect := map[string]any{
		"a": int(1),
		"b": int(2),
		"c": float64(4.2),
	}
	require.Equal(t, expect, actual)
}

```

rfratto:
> I am likely missing something but reading that shouldn't a and b be typed to float64 on the expect?

Ha, you're right. That's a bug in the current implementation of `decodeAny`:

https://github.com/grafana/agent/blob/f954c56a9c90581a07672ee3c9c8c679a39312b1/pkg/river/internal/value/decode.go#L388-L401

It should always be setting it to float64.

rfratto:
This is the right test, which currently fails everywhere:

```go
func TestKnownValuesAny(t *testing.T) {
	in := map[string]any{
		"a": int8(1),
		"b": int16(2),
		"c": float32(4),
	}

	var actual map[string]any
	err := value.Decode(value.Encode(in), &actual)
	require.NoError(t, err)

	expect := map[string]any{
		"a": float64(1),
		"b": float64(2),
		"c": float64(4),
	}
	require.Equal(t, expect, actual)
}
```

PR #3618, opened 2023-04-24
flow: add missing metrics to clustering mode

rfratto:
This PR ensures that clustering metrics are registered when clustering is enabled.


tpaschalis (approved):
Good catch. I tested the change on a my k8s cluster and metrics show up properly. I've just moved the changelog entry to the correct spot.

PR #3626, opened 2023-04-25
flow: fix typo in local.file argument

tpaschalis:
#### PR Description
Fixes a small typo `s/freqency/frequency` in the local.file component arguments. 

No issue filed.

#### Notes to the Reviewer
Arguably, this didn't come up as the default `fsnotify` detector is more handy and appears much more _frequently_ in configurations.

I'm not sure whether we should add a CHANGELOG entry or not; it's not a bug in itself so I'm adding it under "other changes" for now.

#### PR Checklist

- [X] CHANGELOG updated
- [X] Documentation added 
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@ Main (unreleased)
 
 - Add metrics when clustering mode is enabled. (@rfratto)
 
+- Fix spelling of the `frequency` argument on the `local.file` component.
+  (@tpaschalis)
+
 v0.33.0-rc.2 (2023-04-24)
 -------------------------
 
--- a/component/local/file/detector.go
+++ b/component/local/file/detector.go
@@ -76,10 +76,10 @@ type fsNotify struct {
 }
 
 type fsNotifyOptions struct {
-	Logger       log.Logger
-	Filename     string
-	ReloadFile   func()        // Callback to request file reload.
-	PollFreqency time.Duration // How often to do fallback polling
+	Logger        log.Logger
+	Filename      string
+	ReloadFile    func()        // Callback to request file reload.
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -23,6 +23,9 @@ Main (unreleased)
 
 - Add metrics when clustering mode is enabled. (@rfratto)
 
+- Fix spelling of the `frequency` argument on the `local.file` component.
+  (@tpaschalis)
+
```
thampiotr, line 28: Technically this is a breaking change?

tpaschalis: Well, maybe yes? It could also be a bug since it _now_ matches what the configuration used to call the field.

We may have to document this on our Flow-specific [upgrade guide page] (https://grafana.com/docs/agent/next/flow/upgrade-guide/). It might be a good conversation with the team on how we handle such cases.

thampiotr: > It might be a good conversation with the team
Agreed - also how does it work with components being in beta. 

In practice, if I have a config that passes the old spelling, I will get an error on upgrade, so I think it's worth highlighting this in docs in one way or the other.

tpaschalis: (not disagreeing here, just playing devil's advocate) it wouldn't even be possible to get a config that passes just by reading the docs, you'd have to delve into the code.

thampiotr: Right, the docs were correct all the time. Makes sense to keep this as a bugfix then :) LGTM

thampiotr (approved):
LGTM 👍

rfratto (approved):

PR #3630, opened 2023-04-25
river: match documentation when decoding numbers into any

### LINKED ISSUE #3629: River: Number conversion/respresentation
First documented in #3615 during discussion of maps there was an issue found with what the River API guarantees and what is occurring.

```
// To unmarshal River into an interface value, Unmarshal stores one of the
// following:
//
//   - bool, for River bools
//   - float64, for River numbers
//   - string, for River strings
//   - []interface{}, for River arrays
//   - map[string]interface{}, for River objects
//
// Capsule and function types will retain their original type when > decoding
// into an interface value.
```

```go
func TestKnownValuesAny(t *testing.T) {
	in := map[string]any{
		"a": int8(1),
		"b": int16(2),
		"c": float32(4),
	}

	var actual map[string]any
	err := value.Decode(value.Encode(in), &actual)
	require.NoError(t, err)

	expect := map[string]any{
		"a": int(1),
		"b": int(2),
		"c": float64(4),
	}
	require.Equal(t, expect, actual)
}
``` 

According to the guarantee those `int8/int16` should transform into `float64`. This is an issue in the existing river implementation.

rfratto:
Previously, all non-floating point numbers decoded into an int, and all floating-point numbers decoded into a float64. This did not match the documentation, which stated that all River numbers should decode into a float64.

No changelog entry is added because this is only a developer-facing change.

Closes #3629.

### DIFF
--- a/pkg/river/internal/value/decode.go
+++ b/pkg/river/internal/value/decode.go
@@ -386,19 +386,8 @@ func (d *decoder) decodeAny(val Value, into reflect.Value) error {
 		return nil
 
 	case TypeNumber:
-		switch val.Number().Kind() {
-		case NumberKindFloat:
-			var v float64
-			ptr = reflect.ValueOf(&v)
-		case NumberKindInt:
-			var v int
-			ptr = reflect.ValueOf(&v)
-		case NumberKindUint:
-			var v uint
-			ptr = reflect.ValueOf(&v)
-		default:
-			panic("river/value: unreachable")
-		}
+		var v = val.Number().Float()
+		ptr = reflect.ValueOf(&v)
 
 	case TypeArray:
 		var v []interface{}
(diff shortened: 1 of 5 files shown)


erikbaranowski:
This looks good since it matches the docs but seems weird to convert everything to float64 if an int or uint could work.

Is this causing issues as it works today? Should we consider updating the docs instead of changing the code to match the current docs?

rfratto:
It makes working with `any` really annoying, since you have to do three type-checks any time there's a number to see what kind of number it is.

The behavior was copied from encoding/json, where both JSON and River don't specify a difference in numbers (they're all essentially floating point)

erikbaranowski:
if there's precedent and we don't anticipate too much performance converting back to int when we need downstream then works for me

mattdurham:
Can we add a test with decimals?

rfratto:
@mattdurham Thanks for catching that, done!

Issue #3155, opened 2023-03-02
Proposal: remove singleton flag from `prometheus.exporter.unix`


rfratto:
It may make sense to prioritize removing the singleton flag from `prometheus.exporter.unix` as soon as possible so we break changes when Flow adoption is still low. 

I see two potential paths this could take:

1. Remove the singleton flag and validate at runtime that only one `prometheus.exporter.unix` component exists, likely through an atomic variable. 
2. Update our dependency on node_exporter to stop relying on globals and remove the singleton flag. 

Path one would be a short term fix, but path two would be the longer term fix that would take more time to do. 

Related to #1861.

rfratto:
I'm going to add flow/feature-parity to this. Technically it's not for feature parity, but it would be required for stable.

rfratto:
@marctc Since you've been working on this, I'm going to assign it to you.

PR #3637, opened 2023-04-25
Disable node_exporter on windows

jkroepke:
#### PR Description

This disables node_exporter on Windows. The node_exporter does not have Windows support. https://github.com/prometheus/node_exporter/pull/549

This change enables a common configuration across Windows and Linux nodes. While the windows_exporter can be conditionally enabled though an sub component, the node exporter always needs to stay on the root module. (see: https://github.com/grafana/agent/issues/3155)

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated


pkg/integrations/node_exporter/node_exporter_windows.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+// Integration is the node_exporter integration. On Windows platforms,
+// this integration does nothing and will print a warning if enabled.
+type Integration struct {
+}
+
+// New creates a fake node_exporter integration.
+func New(logger log.Logger, _ *Config) (*Integration, error) {
+	level.Warn(logger).Log("msg", "the node_exporter does not works on Windows; enabling it otherwise will do nothing")
```
mattdurham, line 19: minor nit: `works` -> `work`

mattdurham:
Super minor nit but looks great.

mattdurham (approved):
LGTM with a minor nit

pkg/integrations/node_exporter/node_exporter_windows.go
jkroepke:
```suggestion
	level.Warn(logger).Log("msg", "the node_exporter does not work on Windows; enabling it otherwise will do nothing")
```

jkroepke:
Thanks, done

PR #3648, opened 2023-04-26
Add `loki.source.api` component

### LINKED ISSUE #2767: `loki.source.http`: Receive log entries from HTTP calls
We should introduce a component that can spin up a [Loki Push API server](https://grafana.com/docs/loki/next/api/#push-log-entries-to-loki) and receive log entries via HTTP requests. This would act similar to Promtail's [`loki_push_api`](https://grafana.com/docs/loki/next/clients/promtail/configuration/#loki_push_api) 'scrape' config.

thampiotr:
#### PR Description
Adds a new component which embeds promtail's loki push API. 

Fixes #2767

#### Notes to the Reviewer

Looking for early feedback on:
- [x] confirm this is the right name for the component given it can also expose GRPC
- [x] confirm the river config format is following best practices
- [x] confirm what river config default values would be desired

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated
- [ ] Incorporate changes from https://github.com/grafana/agent/pull/3581


component/loki/source/http/http.go
```diff
@@ -0,0 +1,170 @@
...
+	"github.com/weaveworks/common/logging"
+	"github.com/weaveworks/common/server"
+)
+
+// TODO: this component also supports GRPC, so we may want to call it `loki.source.push_api` or something else.
+const componentName = "loki.source.http"
+
+type Arguments struct {
+	HTTPAddress          string              `river:"http_address,attr"`
+	HTTPPort             int                 `river:"http_port,attr"`
```
tpaschalis, line 27: We can default these to `0.0.0.0` and `3500` to match upstream defaults. I'm on the fence about making these required or optional.

These fields will be superseded by #3581 either way so there's no point nitpicking them too much at this stage, let's leave it as-is.

Check how other components make use of a `DefaultArguments` variable and an `UnmarshalRiver` method to enforce defaults and perform any validations.

component/loki/source/http/http.go
```diff
@@ -0,0 +1,170 @@
...
+
+type Component struct {
+	opts         component.Options
+	entriesChan  chan loki.Entry
+	unregisterer *util.Unregisterer
+	rwLock       sync.RWMutex
+
+	// The following fields must be guarded by the rwLock
+	args       Arguments
+	pushTarget *lokipush.PushTarget
```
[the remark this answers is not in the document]
tpaschalis: Also, I'd recommend we use a _separate_ mutex and field for the `forward_to` argument, similar to what we do in `loki.source.docker`. 

https://github.com/grafana/agent/blob/9156a10d29a6dae64e3535f1f9ce7ff64de44bf1/component/loki/source/docker/docker.go#L71-L72

https://github.com/grafana/agent/blob/9156a10d29a6dae64e3535f1f9ce7ff64de44bf1/component/loki/source/docker/docker.go#L129-L131

https://github.com/grafana/agent/blob/9156a10d29a6dae64e3535f1f9ce7ff64de44bf1/component/loki/source/docker/docker.go#L143-L146

This means we can avoid contentions between this mutex (when sending data) and when calling the Update method to re-create the server and avoid issues like #3402 or #3298.

component/loki/source/http/http.go
```diff
@@ -0,0 +1,170 @@
...
+	"github.com/grafana/agent/component/common/loki"
+	"github.com/grafana/agent/component/common/relabel"
+	"github.com/grafana/agent/component/loki/source/http/internal/lokipush"
+	"github.com/prometheus/common/model"
+	"github.com/weaveworks/common/logging"
+	"github.com/weaveworks/common/server"
+)
+
+// TODO: this component also supports GRPC, so we may want to call it `loki.source.push_api` or something else.
+const componentName = "loki.source.http"
```
tpaschalis, line 23: Good point, let's find a better name for the component.

My 2c, in order of preference:
* loki.source.api
* loki.source.network
* loki.source.push_api
* loki.source.receiver`

tpaschalis (commented):
Nice one, we're getting somewhere!

component/loki/source/api/api_test.go
```diff
@@ -0,0 +1,358 @@
+package api
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	client2 "github.com/grafana/agent/component/common/loki/client"
```
thampiotr, line 9: TODO: get rid of this alias

component/loki/write/write.go
```diff
@@ -5,9 +5,10 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/grafana/agent/component/common/loki/client"
```
tpaschalis, line 8: I'm not sure why all these were transferred into their own lines. Is it something from an IDE?

thampiotr: My bad! fixed this now

thampiotr (commented):
Addressed comments.

component/loki/source/api/api.go
```diff
@@ -0,0 +1,176 @@
...
+	newPushTargetConfig := c.pushTargetConfigForArgs(newArgs)
+
+	c.rwMut.Lock()
+	defer c.rwMut.Unlock()
+
+	pushTargetNeedsUpdate := c.pushTarget == nil || !reflect.DeepEqual(c.pushTarget.CurrentConfig(), *newPushTargetConfig)
+	if !pushTargetNeedsUpdate {
+		c.args = newArgs
+		return nil
+	}
```
tpaschalis, line 125: I think the address here will always be different; do we want to compare the struct itself?

And if no changes are needed, then we don't need to update c.args either, right?

docs/sources/flow/reference/components/loki.source.api.md
```diff
@@ -0,0 +1,82 @@
...
+```
+
+## Arguments
+
+`loki.source.api` supports the following arguments:
+
+ Name                     | Type                 | Description                                                | Default | Required 
+--------------------------|----------------------|------------------------------------------------------------|---------|----------
+ `http_port`              | `int`                | The host port for the HTTP server to listen on.            | `0`     | yes      
+ `http_address`           | `string`             | The host address for the HTTP server to listen on.         | `""`    | yes
```
tpaschalis, line 34: Since these fields are required, let's leave the Default empty.

docs/sources/flow/reference/components/loki.source.api.md
```diff
@@ -0,0 +1,82 @@
...
+
+This example starts an HTTP server on `localhost` address and port `9999`. The server receives log entries and forwards
+them to a `loki.echo` component while
+adding a `forwarded="true"` label.
+
+```river
+loki.echo "print" {}
+
+loki.source.api "loki_push_api" {
+    http_address = "localhost"
```
tpaschalis, line 74: Since people will be copying this example, should we make the listen address here `0.0.0.0` so the component can also receive requests from other machines over the network? (since I guess this will be its primary use case)

component/loki/source/api/internal/lokipush/pushtarget.go
```diff
@@ -0,0 +1,219 @@
...
+	t.server = srv
+	t.server.HTTP.Path("/api/v1/push").Methods("POST").Handler(http.HandlerFunc(t.handleLoki))
+	t.server.HTTP.Path("/api/v1/raw").Methods("POST").Handler(http.HandlerFunc(t.handlePlaintext))
+	t.server.HTTP.Path("/ready").Methods("GET").Handler(http.HandlerFunc(t.ready))
+
+	go func() {
+		err := srv.Run()
+		if err != nil {
+			level.Error(t.logger).Log("msg", "loki.source.api server shutdown with error", "err", err)
+		}
```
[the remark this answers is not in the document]
tpaschalis: Were you thinking of something in particular? In the simplest case, we could store a `false` for the server's readiness check and display in the component's debug info?

I don't think this should block us either way.

tpaschalis (approved):
LGTM % some small nits, nice work! 🙌 

I'm ok with putting this in now, and iterating on it when #3581 is merged.

component/loki/source/api/api.go
```diff
@@ -0,0 +1,176 @@
...
+
+func (c *Component) pushTargetConfigForArgs(newArgs Arguments) *lokipush.PushTargetConfig {
+	return &lokipush.PushTargetConfig{
+		Server: server.Config{
+			HTTPListenPort:          newArgs.HTTPPort,
+			HTTPListenAddress:       newArgs.HTTPAddress,
+			Registerer:              c.unregisterer,
+			MetricsNamespace:        "loki_source_api",
+			RegisterInstrumentation: false,
+			Log:                     logging.GoKit(c.opts.Logger),
```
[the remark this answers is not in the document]
tpaschalis: On second thought, the logger should already have the component-specific ID, the extra value would be to only disambiguate logs coming from the component itself vs the component's internal server which shouldn't be as useful.

docs/sources/flow/reference/components/loki.source.api.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Exported fields
+
+`loki.source.api` does not export any fields.
+
+## Component health
+
+`loki.source.api` is only reported as unhealthy if given an invalid configuration.
+
+## Metrics
```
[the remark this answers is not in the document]
thampiotr: There are more metrics that come with the weaveworks' server. I didn't think they are important enough to be on this list. Some other components don't mention them at all.

The metrics I've skipped:
```
# HELP loki_source_api_inflight_requests Current number of inflight requests.
# HELP loki_source_api_tcp_connections_limit The max number of TCP connections that can be accepted (0 means no limit).
```

Happy to try unify these, but again, this could be something for the Server unification.

docs/sources/flow/reference/components/loki.source.api.md
thampiotr: Good idea!

thampiotr (commented):
Thanks for the feedback!

component/loki/source/api/api.go
tpaschalis: My point was that reflect.DeepEqual would compare the addresses and the second part of the clause would _always_ turn to true, right? Shouldn't we have reflect.DeepEqual called on the dereferenced values?

thampiotr: I did wonder the same thing, but `reflect.DeepEqual` will follow the pointers and compare the actual values they are pointing to. I have a test for this here: https://github.com/grafana/agent/blob/21fcc26d7e86118c14c091b9921e7f4d72cf9552/component/loki/source/api/api_test.go#L230

I also have a simple example: https://go.dev/play/p/yBrKf2pyms9.

I think it goes via this code path: https://github.com/golang/go/blob/master/src/reflect/deepequal.go#L127

tpaschalis: Ok, thanks for the heads up!

component/loki/source/api/internal/lokipush/push_api_server.go
```diff
@@ -0,0 +1,239 @@
...
+	for i, r := range s.relabelRules {
+		var rCopy = *r
+		newRules[i] = &rCopy
+	}
+	return newRules
+}
+
+// NOTE: This code is copied from Promtail (3478e180211c17bfe2f3f3305f668d5520f40481) with changes kept to the minimum.
+// Only the HTTP handler functions are copied to allow for flow-specific server configuration and lifecycle management.
+func (s *PushAPIServer) handleLoki(w http.ResponseWriter, r *http.Request) {
```
thampiotr, line 124: I am in favour of adopting this approach instead - we can simplify a lot of surrounding code and only copy the handler functions.

thampiotr (commented):
Updated to use the new server - PTAL

docs/sources/flow/reference/components/loki.source.api.md
```diff
@@ -0,0 +1,93 @@
+---
+title: loki.source.api
+---
+
+# loki.source.api
+
+`loki.source.api` listens for HTTP requests containing Loki log entries and forwards them to other components capable of receiving log entries.
```
tpaschalis, line 7: How does that sound?

```suggestion
`loki.source.api` receives log entries over HTTP and forwards them to other `loki.*` components.
```

docs/sources/flow/reference/components/loki.source.api.md
```diff
@@ -0,0 +1,93 @@
...
+
+The HTTP API exposed is compatible with [Loki push API][loki-push-api], using Loki's `logproto` format. This means that other [`loki.write`][loki.write] components can be used as a client and send requests to `loki.source.api`. This enables Grafana Agents to send log entries over the network to other Grafana Agents.
```
tpaschalis, line 9:
```suggestion
The HTTP API exposed is compatible with [Loki push API][loki-push-api] and the `logproto` format. This means that other [`loki.write`][loki.write] components can be used as a client and send requests to `loki.source.api` and enables using the Agent as a proxy for logs.
```

docs/sources/flow/reference/components/loki.source.api.md
```diff
@@ -0,0 +1,93 @@
...
+* `agent_loki_source_api_request_message_bytes` (histogram): Size (in bytes) of messages received in the request.
+* `agent_loki_source_api_response_message_bytes` (histogram): Size (in bytes) of messages sent in response.
+* `agent_loki_source_api_tcp_connections` (gauge): Current number of accepted TCP connections.
+
+## Example
+
+This example starts an HTTP server on `0.0.0.0` address and port `9999`. The server receives log entries and forwards them to a `loki.echo` component while adding a `forwarded="true"` label.
+
+```river
+loki.echo "print" {}
```
tpaschalis, line 78: nit: could we forward to a loki.write instead and amend the previous sentence? 

I think it will be more useful for people who are looking to copy/paste from our docs

```
loki.write "local" {
  endpoint {
    url = "loki:3100/api/v1/push"
    basic_auth {
      username = env("LOKI_USERNAME")
      password = env("LOKI_PASSWORD")
    }
  }
}
```

docs/sources/flow/reference/components/loki.source.api.md
```diff
@@ -0,0 +1,93 @@
...
+`loki.source.api` is only reported as unhealthy if given an invalid configuration.
+
+## Debug metrics
+
+The following are some of the metrics that are exposed when this component is used. Note that the metrics include labels such as `status_code` where relevant, which can be used to measure request success rates.
+
+* `agent_loki_source_api_request_duration_seconds` (histogram): Time (in seconds) spent serving HTTP requests.
+* `agent_loki_source_api_request_message_bytes` (histogram): Size (in bytes) of messages received in the request.
+* `agent_loki_source_api_response_message_bytes` (histogram): Size (in bytes) of messages sent in response.
+* `agent_loki_source_api_tcp_connections` (gauge): Current number of accepted TCP connections.
```
tpaschalis, line 71: Let's remove the `agent_` prefix from our metrics (and fix that on our code as well) to conform with what other components are doing as well.

```suggestion
* `loki_source_api_request_duration_seconds` (histogram): Time (in seconds) spent serving HTTP requests.
* `loki_source_api_request_message_bytes` (histogram): Size (in bytes) of messages received in the request.
* `loki_source_api_response_message_bytes` (histogram): Size (in bytes) of messages sent in response.
* `loki_source_api_tcp_connections` (gauge): Current number of accepted TCP connections.
```

docs/sources/flow/reference/components/loki.source.api.md
```diff
@@ -0,0 +1,93 @@
...
+
+The HTTP API exposed is compatible with [Loki push API][loki-push-api], using Loki's `logproto` format. This means that other [`loki.write`][loki.write] components can be used as a client and send requests to `loki.source.api`. This enables Grafana Agents to send log entries over the network to other Grafana Agents.
+
+[loki.write]: {{< relref "./loki.write.md" >}}
+[loki-push-api]: https://grafana.com/docs/loki/latest/api/#push-log-entries-to-loki
+
+## Usage
+
+```river
+loki.source.api LABEL {
```
tpaschalis, line 17:
```suggestion
loki.source.api "LABEL" {
```

component/loki/source/api/internal/lokipush/push_api_server_test.go
```diff
@@ -0,0 +1,267 @@
...
+	"time"
+
+	"github.com/go-kit/log"
+	"github.com/grafana/agent/component/common/loki/client/fake"
+	fnet "github.com/grafana/agent/component/common/net"
+	frelabel "github.com/grafana/agent/component/common/relabel"
+	"github.com/grafana/agent/pkg/river"
+	"github.com/grafana/dskit/flagext"
+	"github.com/grafana/loki/clients/pkg/promtail/api"
+	"github.com/grafana/loki/clients/pkg/promtail/client"
```
tpaschalis, line 25: Let's avoid importing promtail directly and use the types in component/common/loki instead, like we did for the component code.

docs/sources/flow/reference/components/loki.source.api.md
```diff
@@ -0,0 +1,93 @@
+---
+title: loki.source.api
+---
+
+# loki.source.api
```
tpaschalis, line 5: Let's find some place to mention what the two exposed endpoints are and what different purpose they serve as currently people wouldn't know where to send logs to.

component/loki/source/api/internal/lokipush/push_api_server.go
```diff
@@ -0,0 +1,240 @@
...
+func (s *PushAPIServer) getRelabelRules() []*relabel.Config {
+	s.rwMutex.RLock()
+	defer s.rwMutex.RUnlock()
+	newRules := make([]*relabel.Config, len(s.relabelRules))
+	for i, r := range s.relabelRules {
+		var rCopy = *r
+		newRules[i] = &rCopy
+	}
+	return newRules
+}
```
tpaschalis, line 121: (nit, not a blocker) I'm not super convinced about the added complexity yet; in truth I don't imagine it will be often that people change the relabel rules or labels by hand that we'd win anything over not having to restart the server every time, and needing the setter/getter/having to copy the fields approach etc.

component/loki/source/api/internal/lokipush/push_api_server.go
```diff
@@ -0,0 +1,240 @@
...
+	registerer prometheus.Registerer,
+) (*PushAPIServer, error) {
+
+	s := &PushAPIServer{
+		logger:       logger,
+		serverConfig: serverConfig,
+		handler:      handler,
+	}
+
+	srv, err := fnet.NewTargetServer(logger, "agent_loki_source_api", registerer, serverConfig)
```
tpaschalis, line 53: As I mentioned, let's drop the metrics prefix.
```suggestion
	srv, err := fnet.NewTargetServer(logger, "loki_source_api", registerer, serverConfig)
```

tpaschalis (approved):
LGTM % docs fixes and few nits! ^^

docs/sources/flow/reference/components/loki.source.api.md
tpaschalis: There's a very concise definition [here](https://grafana.com/docs/loki/next/clients/promtail/configuration/#loki_push_api)

```
Promtail also exposes a second endpoint on /promtail/api/v1/raw which expects newline-delimited log lines. This can be used to send NDJSON or plaintext logs.
```

Also, from the codebase I think we could say that the raw endpoint does not honor `use_incoming_timestamp`.

component/loki/source/api/internal/lokipush/push_api_server_test.go
thampiotr: Yes! sorry, I missed this when I re-copied this test.

docs/sources/flow/reference/components/loki.source.api.md
thampiotr: That will do. Thanks!

PR #3651, opened 2023-04-26
clustering: Use only the address on DNS lookups

tpaschalis:
#### PR Description
This PR makes use of the address only when finding join addresses via a DNS lookup.

No issue filed.

#### Notes to the Reviewer
I found two clusters with different behavior; in one the lookup returned a port, in the other it didn't and made the cluster unable to bootstrap.

I think this ties better with the recommendation that all nodes use the same port; we should double down on that message instead of allowing wiggle room.

#### PR Checklist

- [ ] CHANGELOG updated (N/A)
- [ ] Tests updated (N/A)


rfratto (approved):
LGTM!

PR #3666, opened 2023-04-27
River: add support for string interpolation

### LINKED ISSUE #2972: River: support string interpolation
I'd like to see support for string interpolation added to River. 

String interpolation is a quality-of-life feature which will make it easier to write expressions which return a string. 

For example, this expression:

```river 
env("POD_NAMESPACE") + "/" + env("POD_NAME")
```

could be replaced with the following interpolated string:

```river 
"${env("POD_NAMESPACE")}/${env("POD_NAME")}" 
```

## Syntax 

String interpolation is initiated by placing `${EXPR}` into a string. String interpolation can be escaped by escaping the dollar sign: `\${EXPR}` returns the literal string `${EXPR}`. 

## Evaluation 

Each EXPR instead of `${EXPR}` will be evaluated, converted to a string, and injected into the resulting output. 

For example:

```river 
"3 + 5 is ${3 + 5}" 
```

evaluates to `"3 + 5 is 8"`. 

Expressions should be converted into a string form which best represents that value. The representation for basic types should match the form of that value written in River, with the exception of functions and capsules, which should be converted to some description of the type of the function or capsule.

## Considerations 

This would be a breaking change for users who are using some form of `${EXPR}` in their configuration files already. I'm not sure if any of our existing Flow components would lead to someone using `${EXPR}` inside of a string. 

If this is going to be an issue, we can change the syntax used to denote a string that supports interpolation, such as any of the following:

* C#-like: `$"3 + 5 is {3 + 5}"`
* Javascript-like: `` `3 + 5 is ${3 + 5}` ``

rfratto:
This PR adds support for string interpolation as specified in grafana/river#15. 

Given the size of the commit, I recommend reviewing it commit-by-commit.

I'm opening this in draft since grafana/river#15 is not formally approved, and it's not clear if we want this functionality.

String interpolation required more code changes than is usually necessary; parsing string interpolation required taking existing strings and parsing them further into fragments; `"1 + 2 is ${1 + 2}"` gets parsed into two fragments: `1 + 2 is ` (including the trailing space) and `${1 + 2}`. 

The pretty-printer formats expressions inside interpolated strings, but leaves the other whitespace in the string alone. 

Evaluation works by building a string from all the fragments. Fragments representing expressions (`${EXPR}`) are evaluated and converted into strings. Conversion into strings are only supported for a few types:

* Nulls
* Strings 
* Numbers 
* Bools 

This means string interpolation fails for any other type, including arrays, objects, capsules, and functions. 

Closes grafana/river#15. 

## Example 

<details>
<summary>Example</summary>

```river 
prometheus.exporter.unix { /* use defaults */ }

prometheus.scrape "default" {
	targets    = prometheus.exporter.unix.targets
	forward_to = [prometheus.remote_write.default.receiver]
}

	external_labels = {
		env = "Addr: ${prometheus.exporter.unix.targets[0].__address__}",
	}

	endpoint {
	}
}
```

image

</details>

### DIFF
--- a/docs/sources/flow/config-language/expressions/types_and_values.md
+++ b/docs/sources/flow/config-language/expressions/types_and_values.md
@@ -58,10 +58,41 @@ The supported escape sequences are as follows:
 | `\v` | The vertical tab character `U+000B` |
 | `\'` | The `'` character `U+0027` |
 | `\"` | The `"` character `U+0022`, which prevents terminating the string |
+| `\$` | The `$` character, which escapes [string interpolation][] |
 | `\NNN` | A literal byte (NNN is three octal digits) |
 | `\xNN` | A literal byte (NN is two hexadecimal digits) |
 | `\uNNNN` | A Unicode character from the basic multilingual plane (NNNN is four hexadecimal digits) |
 | `\UNNNNNNNN` | A Unicode character from supplementary planes (NNNNNNNN is eight hexadecimal digits) |
+
+[string interpolation]: #string-interpolation
+
+### String interpolation
+
+Sequences of `${EXPR}` within strings are evaluated as expressions, and their
+results are placed back into the surrounding string.
+
+The string:
+
+```river
+"1 + 2 is: ${1 + 2}"
+```
+
+Evaluates to the following:
+
+```river
+"1 + 2 is: 3"
+```
+
+String interpolation fails if the expression can not be converted into a
+string. String interpolation is compatible with the following types:
+
+* Nulls (`"${null}`)
+* Strings (`"${env("HOME")}`)
+* Numbers (`"${1 + 2}`)
+* Booleans (`${true || false}`)
+
+String interpolation can be escaped by escaping the dollar sign: `"\${EXPR}"`
+will produce the literal string `"${EXPR}"`.
 
 ## Bools
 
--- a/pkg/river/parser/internal.go
+++ b/pkg/river/parser/internal.go
@@ -2,6 +2,7 @@ package parser
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/grafana/agent/pkg/river/ast"
@@ -23,10 +24,11 @@ import (
 // list of errors to be returned to the user. The resulting AST should be
 // discarded if errors were encountered during parsing.
 type parser struct {
-	file     *token.File
-	diags    diag.Diagnostics
-	scanner  *scanner.Scanner
-	comments []ast.CommentGroup
+	file      *token.File
+	diags     diag.Diagnostics
+	scanner   *scanner.Scanner
+	comments  []ast.CommentGroup
+	posOffset int // Offset to add to any position information.
 
 	pos token.Pos   // Current token position
 	tok token.Token // Current token
@@ -57,6 +59,24 @@ func newParser(filename string, src []byte) *parser {
 	return p
 }
 
+func newParserWithOffset(f *token.File, src []byte, off int) *parser {
+	p := &parser{
+		file:      f,
+		posOffset: off,
+	}
+
+	p.scanner = scanner.New(f, src, func(pos token.Pos, msg string) {
+		p.diags.Add(diag.Diagnostic{
+			Severity: diag.SeverityLevelError,
+			StartPos: f.PositionFor(pos.Add(off)),
+			Message:  msg,
+		})
+	}, scanner.IncludeComments)
+
+	p.next()
+	return p
+}
+
 // next advances the parser to the next non-comment token.
 func (p *parser) next() {
 	p.next0()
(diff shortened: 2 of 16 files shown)


pkg/river/parser/interp_string.go
```diff
@@ -0,0 +1,282 @@
...
+	s.readOffset += width
+	s.ch = r
+}
+
+func (s *interpStringScanner) onError(offset int, msg string) {
+	if s.err != nil {
+		s.err(s.file.Pos(offset), msg)
+	}
+	s.numErrors++
+}
```
rfratto, line 121: These utility methods are all copied verbatim from river/scanner, but I don't think it makes sense to try to refactor them out into a shared location.

pkg/river/parser/interp_string.go
```diff
@@ -0,0 +1,282 @@
...
+	switch {
+	case ch >= '0' && ch <= '9':
+		return int(ch - '0')
+	case lower(ch) >= 'a' && lower(ch) <= 'f':
+		return int(lower(ch) - 'a' + 10)
+	}
+	return 16 // Larger than any legal digit val
+}
+
+func lower(ch rune) rune { return ('a' - 'A') | ch }
```
rfratto, line 282: These utility methods are copied verbatim from river/scanner, but I don't think it makes sense to try to refactor them into a shared location.

rfratto:
I've already noticed a bug in the pretty-printer, where `"${env(\"HOME\")}"` gets formatted as `"${env("HOME")}"`, which is incorrect.

rfratto:
> I've already noticed a bug in the pretty-printer, where "${env(\"HOME\")}" gets formatted as "${env("HOME")}", which is incorrect.

Fixed, though not very elegantly.

pkg/river/parser/internal.go
```diff
@@ -638,6 +665,91 @@ var statementEnd = map[token.Token]struct{}{
 	token.COMMA:      {},
 }
 
+// parseInterpString parses an interpolated string sequence.
+func (p *parser) parseInterpString() ast.Expr {
```
thampiotr, line 669: Were there any libraries that we considered using instead?

something like https://github.com/Knetic/govaluate? but there are probably better candidates out there too

rfratto: If there's a library out there that returns a parse tree for specifically the language we want to expose, and it's performant, then I wouldn't be against using a library.

thampiotr: Well, obviously there wouldn't be one for River, but evaluating expressions inside strings, like `"1 + 2"` -> `3` could probably be found.

rfratto: Right, but the problem is the expression inside of string should be a River expression, including referencing component exports like `"${prometheus.exporter.unix.targets[0].__address__}" like in the example attached to the PR. 

That needs to be parsed and not directly evaluated so we can make sure dependencies are hooked up properly in the DAG.

docs/sources/flow/config-language/expressions/types_and_values.md
```diff
@@ -58,11 +58,42 @@ The supported escape sequences are as follows:
...
+
+```river
+"1 + 2 is: 3"
+```
+
+String interpolation fails if the expression can not be converted into a
+string. String interpolation is compatible with the following types:
+
+* Nulls (`"${null}`)
+* Strings (`"${env("HOME")}`)
```
rfratto, line 90:
```suggestion
* Strings (`"${env(\"HOME\")}`)
```

This makes me want to support `'`-formatted strings 😬

rfratto:
>> I've already noticed a bug in the pretty-printer, where "${env(\"HOME\")}" gets formatted as "${env("HOME")}", which is incorrect.
> I temporarily fixed it but there's still a few places where it's buggy, so I'm thinking the pretty printer needs to be able to handle nested escaped quotes more gracefully if we were going to move forward with this.

One very easy way to work around this would be to change the syntax for interpolated strings to Javascript-style backticks, proposed by @jkroepke in https://github.com/grafana/agent/pull/4225#discussion_r1237681552.

pkg/river/parser/interp_string.go
```diff
@@ -0,0 +1,282 @@
+package parser
```
rfratto, line 1: On re-review, it seems pretty unnecessary for the interpStringScanner to be a near-copy of the regular scanner, especially with the escape handling. 

This can likely be replaced with either

* Regex, if regex can properly handle extra `{}` characters inside the interpolated string
* A simpler scanner which is only looking for a non-escaped `${` start sequence, and scans everything inside of that sequence until a matching `}` (while keeping track of inner `{}` pairs). 

If regex is used, performance should be benchmarked to see if it's reasonable, especially for longer strings.

rfratto:
Closing since the related proposal is closed.

PR #3676, opened 2023-04-27
Update create-config-file.md

MattiasSegerdahl:
#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/static/configuration/create-config-file.md
+++ b/docs/sources/static/configuration/create-config-file.md
@@ -182,3 +182,9 @@ integrations:
   node_exporter:
     enabled: true
 ```
+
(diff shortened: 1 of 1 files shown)


rfratto:
Thank you for the contributions this morning. 

I'm not sure this is the best way to expose information about the ports, and as a reviewer, it's a lot of extra burden to have to review 14 separate PRs. 

My gut feeling is that we should inform the user about the different port only where it's relevant (i.e., when installing static mode on Linux).

We'll accept contributions for instructing users that the default port has changed, but it should be in a single PR. I'm going to close this and the other PRs for now.

PR #3695, opened 2023-04-27
docs: Add kustomize considerations on flow mode install for kubernetes

Chewie:
#### PR Description

Following the discussion on slack, the project recommends against using doing new rollouts on config change, which would be the default behavior when using kustomize to generate the ConfigMap if one is not careful.

This PR adds a section to the installation docs mentioning this recommendation, along with a snippet of a kustomization with the appropriate config to disable the behaviour.


rfratto (approved):
Really great first contribution to the docs, thank you! 🎉 Once you sign the CLA we can get this merged :)

PR #3558, opened 2023-04-17
Default back to healthy when a component does not implement component.HealthComponent

erikbaranowski:
#### PR Description

This fixes a bug introduced here https://github.com/grafana/agent/pull/3545

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

PR #3740, opened 2023-05-01
flow: only use component health when component implements component.HealthComponent

rfratto:
This commit offers an alternative approach to #3558, where the overall health of a component only considers the component implementation of health if that component implements the component.HealthComponent interface.

It also has the benefit of reducing the overall public API surface. 

This reverts commit b2f8992ce0c8a7e9de53c871e64774c2ae8f8d8e. The previous commit introduced a bug where the "default health" message was always returned for healthy components, overriding the message informing the user when the last time the component evaluated was.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,11 @@ Main (unreleased)
 
 - Disable node_exporter on Windows systems (@jkroepke)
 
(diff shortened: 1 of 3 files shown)


erikbaranowski (approved):
LGTM, much better!

PR #3741, opened 2023-05-01
Rework module API to reveal less to components and propagate signals

### LINKED ISSUE #3107: Flow Modules: Add metrics if needed

mattdurham:
#### PR Description

This adds two new concepts, a module system and module delegate. The original conversation used the words controller but that got real confusing given the abundance of the word controller. I also think long term converting this to a `pull` pattern of `options.GetModuleDelegate` would be cleaner but that can be a task for another day. A bulk of the changes are removing the log sink code and returning to a simpler time. 

Closes #3107 

#### PR Checklist

- [X] CHANGELOG updated
- [ ] Documentation added
- [X] Tests updated


component/registry.go
```diff
@@ -28,16 +29,47 @@ var (
...
 
+// ModuleSystem is a mechanism responsible for allowing components to create other components, or
+// flow systems.
+type ModuleSystem interface {
+	// NewModuleDelegate creates a new, un-started ModuleDelegate.
+	NewModuleDelegate(id string) ModuleDelegate
+}
+
+// ModuleDelegate is a controller for running components within a Module.
+type ModuleDelegate interface {
```
mattdurham, line 40: Same as the above with making the name more specific to what it is doing, also open to using a different word than `Delegate`

cmd/internal/flowmode/cmd_run.go
```diff
@@ -180,7 +179,13 @@ func (fr *flowRun) Run(configFile string) error {
...
 			}
 		},
-	})
+		Modules: module.NewModule(&module.Options{
+			Logger:    l,
+			Tracer:    t,
+			Clusterer: clusterer,
+			Reg:       reg,
+		}),
+	}, "root")
```
mattdurham, line 188: Went with `root` , I didnt want a blank string but entirely open to ideas.

component/module/file/file.go
```diff
@@ -166,6 +161,11 @@ func (c *Component) CurrentHealth() component.Health {
 	return leastHealthy
 }
 
+// Handler returns the http handler.
```
rfratto, line 164: I feel like this is a less informative comment then what we had before. What happened? Ditto in the other module loader components.

pkg/module/module_test.go
```diff
@@ -0,0 +1,30 @@
+package module
```
rfratto, line 1: I do not want pkg/module to be a part of the public API. I would prefer it if the Flow controller hid the implementation of the interfaces.

pkg/flow/flow.go
```diff
@@ -142,9 +148,9 @@ type Flow struct {
 
 // New creates and starts a new Flow controller. Call Close to stop
 // the controller.
-func New(o Options) *Flow {
+func New(o Options, id string) *Flow {
```
rfratto, line 151: Why are we adding another id argument instead of using ControllerID as the id?

cmd/internal/flowmode/cmd_run.go
```diff
@@ -180,7 +179,13 @@ func (fr *flowRun) Run(configFile string) error {
 				return (&net.Dialer{}).DialContext(ctx, network, address)
 			}
 		},
-	})
+		Modules: module.NewModule(&module.Options{
+			Logger:    l,
+			Tracer:    t,
+			Clusterer: clusterer,
+			Reg:       reg,
+		}),
```
rfratto, line 187: I'm not a fan of this design, since it adds more complexity to the API of consuming a Flow controller. Callers shouldn't need to construct a module interface; the Flow controller should do it for them with a hidden implementation, especially since it shares the same values for the Logger/Tracer/Cluster/Registry.

rfratto (commented):
Gave this a first pass. The PR is huge, so I'm going to need to give a few rounds of feedback here.

cmd/internal/flowmode/cmd_run.go
mattdurham: I did this to avoid circular dependencies. IMO I think I might be able to get around this/mitigate by switching to a pull based system but that seemed outside the scope of an already large PR.

rfratto: This could just be unexported code in the flow controller package, right? It doesn't need to be isolated in its own package.

mattdurham: If I move the http handles back into component/module that removes the circular dependency. The other alternative is to abstract flow http but I would rather not do that at the moment.

mattdurham: I bit the bullet and abstracting the flow api

pkg/module/module_test.go
mattdurham: Moved this into flow, which required an abstraction for flow api, which was something long overdue since it kept tripping things up. Only module interfaces are exposed publicly now

component/jsontypes.go
```diff
@@ -0,0 +1,28 @@
+package component
+
+import (
+	"encoding/json"
+	"time"
+)
+
+// ComponentInfo represents a component in flow.
+type ComponentInfo struct {
```
mattdurham, line 9: To abstract flow api out I had to move this to a neutral spot and this seemed the most reasonable.

pkg/flow/options.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"go.opentelemetry.io/otel/trace"
+)
+
+// ModuleOptions holds static options for module.
+type ModuleOptions struct {
+
+	// Logger to use for controller logs and components. A no-op logger will be
+	// created if this is nil.
+	Logger *logging.Logger
```
mattdurham, line 15: I am a little torn on having this be separate from flow since flow has duplicate values but this feels slightly better referencing flow directly and creating that dependency.

web/api/api.go
```diff
@@ -7,22 +7,27 @@ package api
 import (
```
rfratto, line 7: At this point, it probably makes sense for this code to be moved into pkg/flow too since there's a logical circular dependency between them.

mattdurham: I would rather we work on removing the logical dependency, IMO FlowAPI should be a peer to Flow and not a soft-dependency.

component/module/file/file_test.go
```diff
@@ -56,6 +60,12 @@ func TestModule(t *testing.T) {
 				Registerer:    prometheus.NewRegistry(),
 				OnStateChange: func(e component.Exports) {},
 				DataPath:      t.TempDir(),
+				ModuleSystem: flow.NewModuleSystem(&flow.ModuleOptions{
+					Logger:    l,
+					Tracer:    trace.NewNoopTracerProvider(),
+					Clusterer: cl,
+					Reg:       prometheus.DefaultRegisterer,
+				}),
 			}
```
rfratto, line 69: Like with pkg/module before, I don't want `flow.NewModuleSystem` to be a part of our public API at all:

1. Components should be using `pkg/flow/componenttest` to test their implementations, where `pkg/flow/componenttest` can return some minimally useful interface specific for testing. 
2. Tests against the actual module code should be moved to pkg/flow rather than be done in individual module loaders.

mattdurham: Rewrote the tests to pass through flow instead of directly instantiating the component, for file_test it needs the real module loader. Checks most of the same things but now has to go through the ComponentInfos call to get information on the component.

rfratto: > file_test it needs the real module loader

Can you explain why it needs the real module loader?

pkg/flow/delegate.go
```diff
@@ -0,0 +1,83 @@
...
+// newDelegate creates a module delegate for a specific component.
+func newDelegate(id string, o *ModuleOptions) *delegate {
+	return &delegate{
+		o:  o,
+		id: id,
+	}
+}
+
+// LoadConfig parses River config and loads it.
+func (c *delegate) LoadConfig(config []byte, o component.Options, args map[string]any, onExport component.Export) error {
```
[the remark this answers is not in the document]
mattdurham: Refactored this to remove the dependency on component.Options

cmd/internal/flowmode/cmd_run.go
mattdurham: Didnt like this removed it.

component/module/file/file_test.go
mattdurham: It has a few tests that extract handle bad module loading. Its possible those arent the right place for the tests but I didnt want to make that determination in this PR.

erikbaranowski:
Can we merge main? I can help if needed then plan to deep dive into this next week at a point. That will save me from context swapping when I'm reading between old and current code.

mattdurham:
Going to bounce this into two PRs, one focusing on the API and one focusing on the implementation.

PR #3742, opened 2023-05-01
Use upstream Prometheus

rfratto:
This removes our dependency on the Prometheus fork. Unfortunately, we still need the replace directive because our dependency on Cortex makes Go think that Cortex's Prometheus dependency is newer than the API published version. 

We used to need this dependency because we had downstream changes we were using that we needed. As of 2023, this was just the custom shared string interner. I completed a test today which shows no difference in CPU and memory usage without the custom shared string interner, so there is no reason to use the fork anymore.

We still have some diverged code from upstream, primarily around two implementations of the WAL. I will open a second PR to move away from that, which will hopefully allow us to help test the upstream code.


mattdurham (approved):
LGTM

tpaschalis (approved):
:shipit:

PR #3744, opened 2023-05-02
agent-flow-mixin: add clustering-related dashboards and alerts

tpaschalis:
#### PR Description
This PR ports some internally-used dashboards (all credit goes to @rfratto for the nice-looking graphs) and a couple of rudimentary alerts to the Flow mixin.

Testing was done via [grafana/grizzly](https://github.com/grafana/grizzly/) and was surprisingly easy.

No issue filed.

#### Notes to the Reviewer
I'm having the PR in draft mode until I structure the alerts in jsonnet. I'm having some trouble syncing the alerts to my AlertManager instance right now, looking to see if it's me that's doing something wrong or some Grizzly issue.

Here's the dashboards in action.
image

image

image

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated (N/A)


operations/agent-flow-mixin/grizzly.jsonnet
```diff
@@ -34,4 +34,18 @@ local mixin = import './mixin.libsonnet';
...
+      kind: 'PrometheusRuleGroup',
+      metadata: {
+        folder: $.folder.metadata.name,
+        name: std.split(file, '.')[0],
+        namespace: std.split(file, '.')[0],
+      },
+      spec: mixin.prometheusAlerts[file],
+    }
+    for file in std.objectFields(mixin.prometheusAlerts)
+  },
```
tpaschalis, line 50: This gets rendered as following, which I'm not sure is correct yet.
```
$ jsonnet mixin.libsonnet 
...
...
   },
   "prometheusAlerts": {
      "clustering.json": {
         "rules": [
            {
               "alert": "ClusterNotConverged",
               "annotations": {
                  "message": "Cluster is not converging"
               },
               "expr": "stddev(cluster_node_peers) != 0",
               "for": "2m"
            },
            {
               "alert": "ClusterSplitBrain",
               "annotations": {
                  "message": "Cluster nodes have entered a split brain state"
               },
               "expr": "(sum without (state) (cluster_node_peers)) != count(cluster_node_info)",
               "for": "2m"
            }
         ]
      }
   }
}
```

operations/agent-flow-mixin/dashboards/cluster-overview.libsonnet
```diff
@@ -0,0 +1,231 @@
...
+          },
+          showHeader: true,
+          sortBy: [
+            {
+              desc: true,
+              displayName: 'instance',
+            },
+          ],
+        }) +
+        panel.withFieldConfigs({
```
rfratto, line 87: Some of these field configs may not be necessary. Anything which is the default supplied by Grafana can be omitted, which will be helpful to make the libraries smaller. 

The best way of checking to see if something matters is to run `grr watch` and see if anything changes when modifying the file.

For example, in the case of the table panels, there's no reason to supply the thresholds for values, since there's no values to color green/red.

operations/agent-flow-mixin/dashboards.libsonnet
```diff
@@ -2,5 +2,7 @@
   grafanaDashboards+:
     (import './dashboards/controller.libsonnet') +
     (import './dashboards/resources.libsonnet') +
-    (import './dashboards/prometheus.remote_write.libsonnet'),
+    (import './dashboards/prometheus.remote_write.libsonnet') +
+    (import './dashboards/cluster-node.libsonnet') +
+    (import './dashboards/cluster-overview.libsonnet'),
```
rfratto, line 7: Should these two dashboards be given the deployment annotations like the other dashboards have? The deployment annotation will be especially useful on the clustering dashboards, I feel.

operations/agent-flow-mixin/grizzly.jsonnet
rfratto: Have you checked to see what it renders as in the [mimir-mixin](https://github.com/grafana/mimir/tree/main/operations/mimir-mixin)?

rfratto: FWIW, it might be correct from the mixin's perspective, but then might need to be transformed to be compatible with Grizzly, like with what I have to do for the dashboards: 

https://github.com/grafana/agent/blob/fff8fb246bc87caf80d59982d5e4b5273cc63a16/operations/agent-flow-mixin/grizzly.jsonnet#L25-L35

tpaschalis: When calling `$ jsonnet grizzly.jsonnet`, the output contains the grizzly-specific wrapper, will check against Mimir's mixin output.

operations/agent-flow-mixin/dashboards/cluster-node.libsonnet
```diff
@@ -0,0 +1,299 @@
...
+        panel.withQueries([
+          panel.newQuery(
+            expr='sum(cluster_node_peers{instance="$instance"})',
+          ),
+        ]) +
+        panel.withFieldConfigs({
+          defaults: {
+            unit: 'suffix:peers',
+          },
+        })
```
tpaschalis, line 128: Let's keep this as-is, or extract it in a `withUnit` method? Similarly, we could use a `withOverrides` below instead of passing the entire struct.

operations/agent-flow-mixin/dashboards.libsonnet
tpaschalis: Yes, right! They've been added now.

operations/agent-flow-mixin/dashboards/cluster-node.libsonnet
tpaschalis: Oh, you'd already added `withUnit`. I added `withOverrides` in a similar fashion and removed the generic `withFieldConfig` altogether.

operations/agent-flow-mixin/dashboards/cluster-node.libsonnet
```diff
@@ -0,0 +1,257 @@
...
+      (
+        panel.new('Node Info', 'row') +
+        panel.withPosition({ h: 1, w: 24, x: 0, y: 0 })
+      ),
+      // Node Info
+      (
+        panel.new('Node Info', 'table') +
+        panel.withDescription(|||
+          Information about a specific cluster node.
+        |||) +
```
rfratto, line 40: Can you add documentation to the description? i.e., describe what each of the rows mean, and when the values might be concerning. I can help come up with the description if it's not clear.

operations/agent-flow-mixin/dashboards/cluster-node.libsonnet
```diff
@@ -0,0 +1,257 @@
...
+          y: 18,
+        }) +
+        panel.withQueries([
+          panel.newQuery(
+            expr='1 - ( \r\n    rate(cluster_transport_tx_packets_failed_total{instance="$instance"}[$__rate_interval]) / \r\n    rate(cluster_transport_tx_packets_total{instance="$instance"}[$__rate_interval])\r\n)',
+            legendFormat='Tx success %',
+          ),
+          panel.newQuery(
+            expr='1 - ( \r\n    rate(cluster_transport_rx_packets_failed_total{instance="$instance"}[$__rate_interval]) / \r\n    rate(cluster_transport_rx_packets_total{instance="$instance"}[$__rate_interval])\r\n)',
+            legendFormat='Rx success %',
```
rfratto, line 172: Can we make these multiline Jsonnet strings just so it's easier for us to read? Ditto with any other query with \r\n in it.

operations/agent-flow-mixin/dashboards/cluster-node.libsonnet
```diff
@@ -0,0 +1,257 @@
...
+          ),
+        ]) +
+        panel.withUnit('percentunit')
+      ),
+      // Pending packet queue
+      (
+        panel.new('Pending packet queue', 'timeseries') +
```
rfratto, line 179: Can we add a description here so the user knows what the pending packet queue is, and when they should be concerned about it?

operations/agent-flow-mixin/dashboards/cluster-node.libsonnet
```diff
@@ -0,0 +1,257 @@
...
+          panel.newQuery(
+            expr='1 - ( \r\n    rate(cluster_transport_stream_rx_packets_failed_total{instance="$instance"}[$__rate_interval]) / \r\n    rate(cluster_transport_stream_rx_packets_total{instance="$instance"}[$__rate_interval])\r\n)',
+            legendFormat='Rx success %'
+          ),
+        ]) +
+        panel.withUnit('percentunit')
+      ),
+      // Open transport streams
+      (
+        panel.new('Open transport streams', 'timeseries') +
```
rfratto, line 242: Ditto: can we add a description + detail around when it should be concerning?

operations/agent-flow-mixin/dashboards/cluster-overview.libsonnet
```diff
@@ -0,0 +1,173 @@
...
+        panel.withDescription(|||
+          "Whether the cluster state has converged.
+
+          It is normal for the cluster state to be diverged briefly as gossip events propagate. It is not normal for the cluster state to be diverged for a long period of time.
+
+          This will show one of the following:
+
+          * Converged: Nodes are aware of all other nodes, with the correct states.
+          * Not converged: A subset of nodes aren't aware of their peers, or don't have an updated view of peer states."
+        |||) +
```
rfratto, line 123: Ooh, after we merge this, I would love to work on a state history for a follow up PR so we can show how long the converged/not-converted state has been going on for.

operations/agent-flow-mixin/dashboards/cluster-node.libsonnet
tpaschalis: How does this sound?

```
          The number of packets enqueued currently to be decoded or encoded and sent during communication with other nodes.

          The incoming and outgoing packet queue should be as empty as possible; a growing queue means that the Agent cannot keep up with the number of messages required to have all nodes informed of cluster changes, and the nodes may not converge in a timely fashion.
```

operations/agent-flow-mixin/dashboards/cluster-node.libsonnet
tpaschalis: Yeah, will do.

I've added the following, let me know how it sounds

```
          Information about a specific cluster node.

          Lamport clock time: The observed Lamport time on the specific node's clock used to provide partial ordering around gossip messages. Nodes should ideally be observing roughly the same time, meaning they are up-to-date on the cluster state. If a node is falling behind, it means that it has not recently processed the same number of messages and may have an outdated view of its peers.

          Internal cluster state observers: The number of Observer functions that are registered to run whenever the node detects a cluster change.

          Gossip health score: A health score assigned to this node by the memberlist implementation. The lower, the better.

          Gossip protocol version: The protocol version used by nodes to communicate with one another. It should match across all nodes.```

File: operations/agent-flow-mixin/dashboards/cluster-node.libsonnet (hunk 8 above)
Comment (rfratto) [reply operations/agent-flow-mixin/dashboards/cluster-node.libsonnet:242]: Looks good. I would add information around how the transport streams should eventually close, and the graph should tend towards zero. Anything else is problematic.

> continuously gossip messages

continuously gossiping messages*

File: operations/agent-flow-mixin/dashboards/cluster-node.libsonnet (hunk 7 above)
Comment (rfratto) [reply operations/agent-flow-mixin/dashboards/cluster-node.libsonnet:179]: Perfect 👍

File: operations/agent-flow-mixin/dashboards/cluster-node.libsonnet (hunk 5 above)
Comment (rfratto) [reply operations/agent-flow-mixin/dashboards/cluster-node.libsonnet:40]: Perfect! Make it a bullet point list though, since the description can be Markdown:

```
          Information about a specific cluster node.

          * Lamport clock time: The observed Lamport time on the specific node's clock used to provide partial ordering around gossip messages. Nodes should ideally be observing roughly the same time, meaning they are up-to-date on the cluster state. If a node is falling behind, it means that it has not recently processed the same number of messages and may have an outdated view of its peers.

          * Internal cluster state observers: The number of Observer functions that are registered to run whenever the node detects a cluster change.

          * Gossip health score: A health score assigned to this node by the memberlist implementation. The lower, the better.

          * Gossip protocol version: The protocol version used by nodes to communicate with one another. It should match across all nodes.
```

operations/agent-flow-mixin/dashboards/cluster-node.libsonnet
tpaschalis: > how the transport streams should eventually close

Is that true? My 6-node cluster has been stable for a couple of days and each node has 3-5 open streams at any time. Do you mean during cluster changes?

rfratto: I think it's supposed to be true 😬 I thought transport streams should only open for the streaming API calls, and should close after the net.Conn is closed by memberlist?

(And transport streams are only opened for a push/pull, which happens once a minute with a random node)

tpaschalis: Whoops, will have to check. Not sure if it could be due to the timeout hapenning on a similar timeframe? :/

PR #3745, opened 2023-05-02
WIP: Add process component to phlare, derived from Loki's

simonswine:
This will add a component `phlare.process`, which allows to skip/rewrite/modify profiles after scraping and before sending it of to Phlare API.

The first stage to implement in this PR would be calculate `delta` of a cumulative profile since the last profile received. This is currently done by Pyroscope & Phlare on the server side.

### DIFF
--- a/component/common/phlare/types.go
+++ b/component/common/phlare/types.go
@@ -1,0 +1,138 @@
+package phlare
+
+import (
+	"context"
+	"sync"
+	"time"
+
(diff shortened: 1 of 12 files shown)


tpaschalis:
Hey there @simonswine.

I'm going through a round of cleaning up our PRs backlog. Is there any possibility that we continue with a new `pyroscope.process` PR and want to leave this up as a reference? Or do you think this can be closed?

PR #3748, opened 2023-05-02
docs: document naming convention for types

### LINKED ISSUE #3657: Flow: component docs mention types not present in language reference
At various points, the component docs mention types such as `map(string)` and ` list(string)` (see [loki.write](https://grafana.com/docs/agent/v0.33/flow/reference/components/loki.write/#arguments)). However, the [language reference](https://grafana.com/docs/agent/v0.33/flow/config-language/expressions/types_and_values/) does not mention these types.

In particular, the map type was a head scratcher to write a literal for. I was able to find an example in the [phlare.write](https://grafana.com/docs/agent/v0.33/flow/reference/components/phlare.write/#example) after grepping the repo, but it would be good to have a central example showing how to fill such attributes.

rfratto:
Closes #3657.


tpaschalis (approved):
LGTM, two minor points if you think they make sense to add.

a) Should we also mention `ms` and `ns`?
b) Should we mention that map keys are _always_ strings (and that's why they're omitted)

docs/sources/flow/config-language/expressions/types_and_values.md
```diff
@@ -23,6 +23,25 @@ River uses the following types for its values:
   or more arguments as input and always return a single value as output.
 * `null`: A type that has no value.
 
+### Naming convention
+
+In addition to the types above, [component reference][] documentation will use
+the following conventions for refering to types:
```
erikbaranowski, line 29:
```suggestion
the following conventions for referring to types:
```

docs/sources/flow/config-language/expressions/types_and_values.md
```diff
@@ -23,6 +23,25 @@ River uses the following types for its values:
...
+
+In addition to the types above, [component reference][] documentation will use
+the following conventions for referring to types:
+
+* `any`: A value of any type.
+* `map(T)`: an `object` where the value type is `T`. For example, `map(string)`
+  is an object where all the values are strings. The key type of an object is
+  always a string, or an identifier which is converted into a string.
+* `list(T)`: an `array` where the value type is `T`. For example, `list(string`
+  is an array where all the values are strings).
```
erikbaranowski, line 36:
```suggestion
* `list(T)`: an `array` where the value type is `T`. For example, `list(string)`
  is an array where all the values are strings.
```

Issue #2059, opened 2022-08-23
Flow: naming convention for components


rfratto:
## Background

If you include #1872 and #1843, this is our current list of Flow components:

* `discovery.k8s`
* `integrations.node_exporter`
* `local.file`
* `metrics.mutate`
* `metrics.remote_write`
* `metrics.scrape`
* `otel.exporter_jaeger`
* `otel.exporter_otlp`
* `otel.exporter_zipkin`
* `otel.receiver_jaeger`
* `otel.receiver_otlp`
* `otel.receiver_zipkin`
* `remote.s3`
* `targets.mutate`

To speed up development time, these components were initially named without much consideration to how they interact with other components.

We will have many more components as time goes on, and so we have a growing need for a naming convention to align related components.

### Namespaces

Every component is in a "namespace" where related components are placed. For example, the `remote.s3` component above is in the `remote` namespace.

Namespaces may be nested when appropriate. For example, an `a.b.c` component is in the `a.b` namespace (e.g., the `b` namespace inside the `a` namespace). Nesting a namespace is appropriate when components in a namespace can be divided into multiple responsibilities.

## Proposal

Namespaces should:

1. Group components by the data they act on in a pipeline
2. Be named with _singular_ nouns

Components should:

1. Be named with a verb if they transform data
2. Otherwise, either verbs or nouns are acceptable

I am proposing our list of components becomes:

* `discovery.k8s`
* `discovery.relabel`
* `local.file`
* `otelcol.exporter.jaeger`
* `otelcol.exporter.otlp`
* `otelcol.exporter.zipkin`
* `otelcol.receiver.jaeger`
* `otelcol.receiver.otlp`
* `otelcol.receiver.zipkin`
* `prometheus.integration.node_exporter`
* `prometheus.relabel`
* `prometheus.scrape`
* `prometheus.remote_write`
* `remote.s3`

### `discovery` namespace

The `discovery` namespace holds components which expose targets. Its name is unchanged from the existing namespace.

The existing `targets.mutate` component is renamed to `discovery.relabel`, since it also has the purpose of exposing targets to other components in the pipeline.

### `local` namespace

The `local` namespace holds components which expose data from the local machine. Its name is unchanged from the existing namespace.

### `otelcol` namespace

The `otelcol` namespace holds components which expose OpenTelemetry Collector components as Flow components. It was renamed from the `otel` namespace to make the intent clearer; "otel" is a generic term for all of OpenTelemetry, and should be avoided.

[...]

PR #3753, opened 2023-05-02
Add `prometheus.receive_http` component to allow receiving metrics over HTTP

### LINKED ISSUE #2780: Flow: allow receiving remote_write metrics over HTTP
The static mode [supports sending metrics over HTTP](https://grafana.com/docs/agent/latest/api/#accept-remote_write-requests) to a metrics instance, but there is no Flow equivalent for this yet. 

I see two paths:

1. Add an API endpoint for `prometheus.remote_write` that allows external processes to send metrics to it over the network (i.e., `/components/prometheus.remote_write.default/remote_write`) 
2. Create a new component dedicated to receiving metrics over the network, which would allow it to be chained into other components for processing. 

Option 2 is probably immediately preferable for its increased flexibility.

thampiotr:
#### PR Description

Adds a new `prometheus.receive_http` component to allow receiving metrics over HTTP.

Fixes #2780 

#### PR Checklist

- [x] Finish the draft
- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated


component/prometheus/source/api/api.go
```diff
@@ -0,0 +1,179 @@
...
+	defer func() {
+		c.updateMut.Lock()
+		defer c.updateMut.Unlock()
+		c.shutdownServer()
+	}()
+
+	for {
+		select {
+		case err := <-c.serverErr:
+			return fmt.Errorf("server terminated with error: %v", err)
```
rfratto, line 82: We don't recommend having any kind of signal which would exit the Run function, that's effectively fatal for the component; nothing will cause it to rerun. 

Instead, if there's an error like this, expose the component as unhealthy by implementing [component.HealthComponent](https://github.com/grafana/agent/blob/93cb1a2718f6fc90fd06ef33b6bcff519dbed662/component/component_health.go#L14), and allow the component to return to a healthy state once the user reconfigures the component or triggers a reload.

thampiotr: Thanks for spotting this and the suggestion. I've implemented this now :)

go.mod
```diff
@@ -171,7 +171,7 @@ require (
 	golang.org/x/time v0.3.0
 	google.golang.org/api v0.109.0
 	google.golang.org/grpc v1.52.3
-	google.golang.org/protobuf v1.28.1
+	google.golang.org/protobuf v1.30.0
```
thampiotr, line 174: The update was needed so that I can use the `protoadapt` package in api_test.go - see comments in that file for more details.

component/prometheus/source/api/api_test.go
```diff
@@ -0,0 +1,392 @@
...
+
+	client, err := remote.NewWriteClient("remote-write-client", &remote.ClientConfig{
+		URL:     &config.URL{URL: remoteWriteURL},
+		Timeout: model.Duration(30 * time.Second),
+	})
+	if err != nil {
+		return err
+	}
+
+	buf, err := proto.Marshal(protoadapt.MessageV2Of(req))
```
thampiotr, line 371: We need to use `protoadapt.MessageV2Of` since the Agent repo uses the new V2 [google.golang.org/protobuf](https://pkg.go.dev/mod/google.golang.org/protobuf) library, while prometheus uses the V1 [github.com/golang/protobuf](https://pkg.go.dev/mod/github.com/golang/protobuf) library. Importing the V1 is a linter error.

In order to use the `protoadapt`, I had to upgrade the `google.golang.org/protobuf` to 1.30.0 - see `go.mod`.

I don't think it's a problem since it's a minor version and we would eventually want to update this dependency anyway.

component/prometheus/source/api/api.go
```diff
@@ -0,0 +1,141 @@
...
+	defer func() {
+		c.updateMut.Lock()
+		defer c.updateMut.Unlock()
+		c.shutdownServer()
+	}()
+
+	for range ctx.Done() {
+		level.Info(c.opts.Logger).Log("msg", "terminating due to context done")
+		return nil
+	}
```
thampiotr, line 77: Linter told me to use this syntax. Personally I don't mind a more verbose, but more intuitive version of
```
for {
  select {
    ...
  }
}
```
But seems that when there is only one `case` in the `select` statement, the `range` syntax is recommended.

component/prometheus/source/api/api.go
```diff
@@ -0,0 +1,141 @@
...
+	agentprom "github.com/grafana/agent/component/prometheus"
+	"github.com/grafana/agent/pkg/util"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/prometheus/storage"
+	"github.com/prometheus/prometheus/storage/remote"
+)
+
+func init() {
+	component.Register(component.Registration{
+		Name: "prometheus.source.api",
```
tpaschalis, line 23: We usually use multiple dots to namespace components (some [conventions](https://github.com/grafana/agent/issues/2059) from back when we were first naming components).

I don't think we'll introduce more `prometheus.source` components, so wdyt about reconsidering the name? Throwing some ideas out there, but I'd like to hear your suggestions

* prometheus.http
* prometheus.httpreceiver
* prometheus.remotewritereceiver

component/prometheus/source/api/api.go
tpaschalis: Why do we need the `for {}` here? 

Since there's nothing else on the processing loop (like we have for logging components), we could just block on `<-ctx.Done()`, right?

component/prometheus/source/api/api.go
```diff
@@ -0,0 +1,141 @@
...
+		Args: Arguments{},
+		Build: func(opts component.Options, args component.Arguments) (component.Component, error) {
+			return New(opts, args.(Arguments))
+		},
+	})
+}
+
+type Arguments struct {
+	Server    *fnet.ServerConfig   `river:",squash"`
```
tpaschalis, line 32: Up to you if you prefer to use the shared server implementation (probably for the best), just taking the opportunity to let you know about the HTTPComponent extension interface which can register a component's handler on the Agent's root HTTP server.

The handler is registered on a path unique to the component; so a call to `/component/{component_id}//api/v1/metrics/write` would reach the endpoint of the `prometheus.source.api "component_id"` instance.

https://github.com/grafana/agent/blob/d5b1d7dc1ee8b742f3a196e9c9c53181d85c27eb/component/component.go#L118-L127

component/prometheus/source/api/api_test.go
```diff
@@ -0,0 +1,392 @@
+package api
```
tpaschalis, line 1: Nice and readable testing! :)

docs/sources/flow/reference/components/prometheus.source.api.md
```diff
@@ -0,0 +1,111 @@
...
+prometheus.source.api "LABEL" {
+  http {
+    listen_address = "LISTEN_ADDRESS"
+    listen_port = PORT 
+  }
+  forward_to = RECEIVER_LIST
+}
+```
+
+The component will start HTTP server supporting the following requests:
```
tpaschalis, line 26:
```suggestion
The component will start an HTTP server supporting the following endpoint:
```

docs/sources/flow/reference/components/prometheus.source.api.md
```diff
@@ -0,0 +1,111 @@
...
+  }
+  forward_to = [prometheus.remote_write.local.receiver]
+}
+
+// Writes metrics to a specified address, e.g. cloud-hosted Prometheus instance
+prometheus.remote_write "local" {
+  endpoint {
+    url = "http://my-cloud-prometheus-instance.com/api/prom/push"
+  }
+}
```
tpaschalis, line 90: super nitpicky (apologies), but Prometheus itself couldn't receive request to an endpoint like this; let's alter the example a bit.

```suggestion
// Send metrics to a locally running Mimir.
prometheus.remote_write "local" {
  endpoint {
    url = "http://mimir:9009/api/v1/push"
    
    basic_auth {
      username = "example-user"
      password = "example-password"
    }
  }
}
```

docs/sources/flow/reference/components/prometheus.source.api.md
```diff
@@ -0,0 +1,111 @@
...
+
+// Writes metrics to localhost:9999/api/v1/metrics/write - e.g. served by 
+// the prometheus.source.api component from the example above.
+prometheus.remote_write "local" {
+  endpoint {
+    url = "http://localhost:9999/api/v1/metrics/write"
+  }  
+}
+```
```
tpaschalis, line 111: Could we put a new header (`## Proxying metrics` maybe?) on this example? Just so that people scanning the page don't end up copying both.

docs/sources/flow/reference/components/prometheus.source.api.md
```diff
@@ -0,0 +1,111 @@
...
+* `prometheus_source_api_request_duration_seconds` (histogram): Time (in seconds) spent serving HTTP requests.
+* `prometheus_source_api_request_message_bytes` (histogram): Size (in bytes) of messages received in the request.
+* `prometheus_source_api_response_message_bytes` (histogram): Size (in bytes) of messages sent in response.
+* `prometheus_source_api_tcp_connections` (gauge): Current number of accepted TCP connections.
+* `agent_prometheus_fanout_latency` (histogram): Write latency for sending metrics to other components.
+* `agent_prometheus_forwarded_samples_total` (counter): Total number of samples sent to downstream components.
+
+## Example
+
+This example creates a `prometheus.source.api` component which starts an HTTP server on `0.0.0.0` address and port `9999`. The server receives metrics and forwards them to a `prometheus.remote_write` component which writes these metrics to a specified HTTP endpoint.
```
tpaschalis, line 73:
```suggestion
This example creates a `prometheus.source.api` component which starts an HTTP server listening on `0.0.0.0` and port `9999`. The server receives metrics and forwards them to a `prometheus.remote_write` component which writes these metrics to the specified HTTP endpoint.
```

component/prometheus/source/api/api.go
```diff
@@ -0,0 +1,141 @@
...
+	c.shutdownServer()
+
+	err, s := c.createNewServer(newArgs)
+	if err != nil {
+		return err
+	}
+	c.server = s
+
+	err = c.server.MountAndRun(func(router *mux.Router) {
+		router.Path("/api/v1/metrics/write").Methods("POST").HandlerFunc(c.handler.ServeHTTP)
```
tpaschalis, line 103: Is it the same as `Methods("POST").Handler(c.handler)`?

tpaschalis (commented):
Some good progress; comments are mainly around documentation and naming. Looking forward to seeing this land!

docs/sources/flow/reference/components/prometheus.source.api.md
thampiotr: That's a good suggestion, thanks!

PR #3764, opened 2023-05-03
cluster: honor deadline when opening TCP connection

rfratto:
Although synchronous writes to peers have a timeout, this timeout only affected the HTTP/2 request and not the net.Dial call. This commit attempts to honor the deadline when establishing a connection to a peer.

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -155,14 +155,17 @@ func New(log log.Logger, reg prometheus.Registerer, clusterEnabled bool, listenA
 		Transport: &http2.Transport{
 			AllowHTTP: true,
 			DialTLSContext: func(ctx context.Context, network, addr string, _ *tls.Config) (net.Conn, error) {
-				if deadline, ok := ctx.Deadline(); ok {
-					return net.DialTimeout(network, addr, time.Until(deadline))
+				// Set a maximum timeout for establishing the connection. If our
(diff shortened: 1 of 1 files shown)


tpaschalis (approved):
LGTM. According to the docs on x/net/http2, `DialTLS` is deprecated and they point to using `DialTLSContext` instead so this is a positive change in any case.

pkg/cluster/cluster.go
```diff
@@ -153,7 +154,14 @@ func New(log log.Logger, reg prometheus.Registerer, clusterEnabled bool, listenA
...
 			AllowHTTP: true,
-			DialTLS: func(network, addr string, _ *tls.Config) (net.Conn, error) {
+			DialTLSContext: func(ctx context.Context, network, addr string, _ *tls.Config) (net.Conn, error) {
+				if deadline, ok := ctx.Deadline(); ok {
+					return net.DialTimeout(network, addr, time.Until(deadline))
+				}
+
+				// TODO(rfratto): should we use a default timeout if one isn't set from
+				// the context? [http.DefaultTransport] configures a net.Dialer with a
+				// timeout of 30 seconds by default.
```
tpaschalis, line 164: I'm trying to understand whether this will have any effect right now with the transport _we_ are using for clustetring. If not, I think setting a default would be a good start

PR #3765, opened 2023-05-03
phlare.scrape: support clustering

rfratto:
This commit adds clustering support to `phlare.scrape`, allowing profile collection to be distributed amongst agents participating in a cluster.

Let me know if there's anything I missed to allow clustering to work with phlare.scrape.


tpaschalis (approved):
cc @cyriltovena @simonswine 

LGTM, should be as straightforward as this. I've fixed a test instance that didn't have access to a Node implementation, should be alright to merge.

PR #3776, opened 2023-05-04
Fix issue where not providing config could result in panic

thampiotr:
#### PR Description
When not providing either `http` or `grpc` configuration blocks for `loki.source.(heroku|gcplog)`, the user may get a panic when starting the server.
This PR fixes this by replacing `nil` with default.

N/A

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/flow/reference/components/loki.source.heroku.md
+++ b/docs/sources/flow/reference/components/loki.source.heroku.md
@@ -51,10 +51,10 @@ before they're forwarded to the list of receivers in `forward_to`.
 
 The following blocks are supported inside the definition of `loki.source.heroku`:
 
-Hierarchy | Name | Description | Required
---------- | ---- | ----------- | --------
-`http`    | [http][]      | Configures the HTTP server that receives requests.              |  | no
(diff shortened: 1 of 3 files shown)


docs/sources/flow/reference/components/loki.source.heroku.md
```diff
@@ -51,10 +51,10 @@ before they're forwarded to the list of receivers in `forward_to`.
 
 The following blocks are supported inside the definition of `loki.source.heroku`:
 
-Hierarchy | Name | Description | Required
---------- | ---- | ----------- | --------
-`http`    | [http][]      | Configures the HTTP server that receives requests.              |  | no
```
thampiotr, line 56: Count of columns didn't match the header.

thepalbi (approved):
nice catch

tpaschalis (approved):
LGTM. This could be introduced a layer above, when calling the Convert() method and applying some defaults.

Let's merge this for now and reconsider when we start utilizing this in components.

PR #3711, opened 2023-04-28
loki.source.heroku: use an unchecked collector for server metrics

rfratto:
Every time an HTTP server for `loki.source.heroku` is created, it creates a new weaveworks/common server, and registers the same set of metrics to the provided collector.

By default, these are checked metrics, which causes the agent-wide `/metrics` endpoint to stop working with an error:

    An error has occurred while serving metrics:

    4 error(s) occurred:
    * collected metric "loki_source_heroku_drain_target_tcp_connections_limit" { label:<name:"component_id" value:"loki.source.heroku.repro" > label:<name:"protocol" value:"http" > gauge:<value:0 > } was collected before with the same name and label values
    * collected metric "loki_source_heroku_drain_target_tcp_connections_limit" { label:<name:"component_id" value:"loki.source.heroku.repro" > label:<name:"protocol" value:"grpc" > gauge:<value:0 > } was collected before with the same name and label values
    * collected metric "loki_source_heroku_drain_target_tcp_connections" { label:<name:"component_id" value:"loki.source.heroku.repro" > label:<name:"protocol" value:"http" > gauge:<value:0 > } was collected before with the same name and label values
    * collected metric "loki_source_heroku_drain_target_tcp_connections" { label:<name:"component_id" value:"loki.source.heroku.repro" > label:<name:"protocol" value:"grpc" > gauge:<value:0 > } was collected before with the same name and label values

This commit adds a new utility method to create an unchecked prometheus.Collector implementation, and then creates a new prometheus.Registry to store server metrics any time a new server is created. The prometheus.Registry instance for the server is then passed to the unchecked prometheus.Collector for handling.

Closes #3646.

PR #3780, opened 2023-05-04
Prototype using our own embedded HTTP server

thampiotr:
#### Problem 
Thinking about our recent components that embed an HTTP server, there are a few issues with weaveworks sever:
- Starts GRPC server always which can lead to port conflicts and will use more resources. We only require HTTP embedded servers currently. 
- Has issues with re-registering existing metrics when we want to restart it - which requires workarounds: https://github.com/grafana/agent/pull/3711
- Requires adapter code for converting configs from River to weaveworks, which is a bit boilerplate (even though the shared river configs are much nicer than what we had before).

#### Proposed solution
The [weaveworks server](https://github.com/weaveworks/common/blob/master/server/server.go) is not very complex though (500LOC) and I think we can maintain a similar implementation of our own that would be simpler and suit our needs better.

**This PR is a quick hack to show how this could work in practice and verify how much code would this require.**

#### Alternatives
Alternatives are:
- Keep working around weavework server's issues, which I feel results in more boilerplate
- Contribute upstream so we can e.g. disable GRPC and re-register metrics as needed - I suspect this wouldn't be desirable for the upstream, since if someone wants just HTTP, why not just use standard built-in which would make things simpler?

### PR Checklist

- [ ] Fix invalid comments
- [ ] Clean up the prototype 
- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/component/common/net/config.go
+++ b/component/common/net/config.go
@@ -3,30 +3,62 @@
 package net
 
 import (
-	"flag"
+	"github.com/gorilla/mux"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/weaveworks/common/logging"
+	"github.com/weaveworks/common/middleware"
 	"time"
-
-	weaveworks "github.com/weaveworks/common/server"
 )
 
 const (
 	DefaultHTTPPort = 8080
-	DefaultGRPCPort = 8081
 )
 
-// ServerConfig is a River configuration that allows one to configure a weaveworks.Server. It
-// exposes a subset of the available configurations.
+// ServerConfig for a Server
 type ServerConfig struct {
+
+	// ==== configuration exposed in River configs ===
+
 	// HTTP configures the HTTP weaveworks. Note that despite the block being present or not,
 	// the weaveworks is always started.
 	HTTP *HTTPConfig `river:"http,block,optional"`
 
-	// GRPC configures the gRPC weaveworks. Note that despite the block being present or not,
-	// the weaveworks is always started.
-	GRPC *GRPCConfig `river:"grpc,block,optional"`
-
 	// GracefulShutdownTimeout configures a timeout to gracefully shut down the server.
 	GracefulShutdownTimeout time.Duration `river:"graceful_shutdown_timeout,attr,optional"`
+
+	// ==== configuration NOT exposed in River configs ===
+
+	MetricsNamespace  string
+	HTTPListenNetwork string
+
+	CipherSuites  string
+	MinVersion    string
+	HTTPTLSConfig TLSConfig
+
+	RegisterInstrumentation  bool
+	ExcludeRequestInLog      bool
+	DisableRequestSuccessLog bool
+
+	HTTPMiddleware                []middleware.Interface
+	Router                        *mux.Router
+	DoNotAddDefaultHTTPMiddleware bool
+
+	LogFormat             logging.Format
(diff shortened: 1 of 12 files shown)


component/common/net/config.go
```diff
@@ -3,30 +3,62 @@
...
 
-	// GRPC configures the gRPC weaveworks. Note that despite the block being present or not,
-	// the weaveworks is always started.
-	GRPC *GRPCConfig `river:"grpc,block,optional"`
-
 	// GracefulShutdownTimeout configures a timeout to gracefully shut down the server.
 	GracefulShutdownTimeout time.Duration `river:"graceful_shutdown_timeout,attr,optional"`
+
+	// ==== configuration NOT exposed in River configs ===
+
```
thampiotr, line 30: Exposing these in River would be trivial now and all components could benefit from new configuration options, e.g. TLS.

component/common/net/config.go
```diff
@@ -3,30 +3,62 @@
...
 
 const (
 	DefaultHTTPPort = 8080
-	DefaultGRPCPort = 8081
 )
 
-// ServerConfig is a River configuration that allows one to configure a weaveworks.Server. It
-// exposes a subset of the available configurations.
+// ServerConfig for a Server
 type ServerConfig struct {
```
thampiotr, line 18: This struct is a merge of our ServerConfig and the weaveworks' server Config.
But we would be in control of how this works and what is available in River.

component/common/net/tls_config.go
```diff
@@ -0,0 +1,55 @@
+package net
```
thampiotr, line 1: This file is verbatim copied from weaveworks common.

component/common/net/server.go
```diff
@@ -1,97 +1,353 @@
 package net
```
thampiotr, line 1: This is a modified copy of weaveworks' common server. I've removed the GRPC support and made it work with our merged configuration struct.

component/common/net/server.go
```diff
@@ -1,97 +1,353 @@
...
+	if gatherer == nil {
+		gatherer = prometheus.DefaultGatherer
+	}
+
+	tcpConnections := prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: cfg.MetricsNamespace,
+		Name:      "tcp_connections",
+		Help:      "Current number of accepted TCP connections.",
+	}, []string{"protocol"})
+	reg.MustRegister(tcpConnections)
```
thampiotr, line 113: We will now be able to use `Register` and reuse the existing metric if it's already registered. Avoiding the need for workaround here https://github.com/grafana/agent/pull/3711

thampiotr (commented):
Left some comments to help read the code.

thampiotr:
@thepalbi @tpaschalis When you have some time, I would appreciate some feedback on this idea.
(note, this is not ready for merge, just to demonstrate how this could work)

thampiotr:
> @thepalbi @tpaschalis When you have some time, I would appreciate some feedback on this idea. (note, this is not ready for merge, just to demonstrate how this could work)

Had a chat with @tpaschalis about this and we'll explore a couple of alternatives before that, but we keep it as a an open alternative for the future. 

We should note that we already have a fork of weaveworks server in `pkg/server/server.go` and maybe these could be unified?

PR #3782, opened 2023-05-04
Sync WAL code with upstream

rfratto:
This commit synchronizes the WAL code with the upstream tsdb/agent code in Prometheus. I synced the code  against the latest main commit at the time of starting the commit; prometheus/prometheus@7c2de14b0bd74303c2ca6f932b71d4585a29ca75.  

For now, I opted to synchronize the code instead of importing upstream literally for two reasons:

1. Importing upstream literally will require more effort and be a breaking change, as the metrics and metric names used by upstream are different than the names we had selected. 
2. For now, we may want to continue to experiment with changes to WAL storage downstream, such as for clustering and cleaning up memory faster. 

This change should have little-to-no noticeable impact on end users, with one exception being that each series should use slightly less memory (2 bytes less, due to removing two bools) and will GC stale series after one truncate rather than two.

### DIFF
--- a/pkg/metrics/wal/series.go
+++ b/pkg/metrics/wal/series.go
@@ -11,40 +11,29 @@ import (
 type memSeries struct {
 	sync.Mutex
 
-	ref    chunks.HeadSeriesRef
-	lset   labels.Labels
+	ref  chunks.HeadSeriesRef
+	lset labels.Labels
+
+	// Last recorded timestamp. Used by gc to determine if a series is stale.
 	lastTs int64
-
-	// TODO(rfratto): this solution below isn't perfect, and there's still
-	// the possibility for a series to be deleted before it's
-	// completely gone from the WAL. Rather, we should have gc return
-	// a "should delete" map and be given a "deleted" map.
-	// If a series that is going to be marked for deletion is in the
-	// "deleted" map, then it should be deleted instead.
-	//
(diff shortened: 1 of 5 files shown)


pkg/metrics/wal/series.go
```diff
@@ -8,46 +8,36 @@ import (
 	"github.com/prometheus/prometheus/tsdb/chunks"
 )
```
mattdurham, line 9: Should we add a disclaimer similar to what we did when porting promtail over? Basically its ported over with only changes to metric names

pkg/metrics/wal/wal.go
```diff
@@ -145,9 +151,9 @@ func NewStorage(logger log.Logger, registerer prometheus.Registerer, path string
 		wal:     w,
 		logger:  logger,
 		deleted: map[chunks.HeadSeriesRef]int{},
-		series:  newStripeSeries(),
+		series:  newStripeSeries(tsdb.DefaultStripeSize),
 		metrics: newStorageMetrics(registerer),
-		ref:     atomic.NewUint64(0),
+		nextRef: atomic.NewUint64(0),
```
mattdurham, line 156: This name feels odd, but its what is their now.

mattdurham (approved):
LGTM

tpaschalis (approved):

PR #3788, opened 2023-05-05
agent-flow-mixin: tweaks to clustering alerts

rfratto:
1. Use an alert period of 5m, since an alert period of 2m may be too aggressive for a cluster to converge considering startup times.

2. Allow the ClusterSplitBrain alert to be tolerant of multiple ckit clusters across different namespaces in the same Kubernetes cluster.


tpaschalis (approved):
LGTM, missed amending the second alert yesterday.

While this can be applied using Grizzly, `mixtool` reports a slightly different structure than the one reported by `mimir-mixin` so I'll have a follow-up PR to fix it.

PR #3792, opened 2023-05-05
remove replace for github.com/iamseth/oracledb_exporter

### LINKED ISSUE #3415: Get changes of `oracledb_exporter` merged upstream and remove `replace` in go.mod

schmikei:
#### PR Description

With the merges of 

https://github.com/iamseth/oracledb_exporter/pull/308
and
https://github.com/iamseth/oracledb_exporter/pull/312

This exporter can be used without the replace statement 🚀 

During the refactors there is no need to specify the `metrics_scrape_interval` as the desired behavior should be encompassed in `scrape_interval` of the integration and that parameter should be configured in . I am okay following a deprecation for the parameter, just didn't know how that process should go!

Fixes #3415

#### Notes to the Reviewer

Have validated integration notes already and tested the integration by sending to Grafana Cloud 🎉 

image

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ Main (unreleased)
 ### Breaking changes
 
 - The experimental dynamic configuration feature has been removed in favor of Flow mode. (@mattdurham)
+- The `oracledb` integration configuration has removed a redundant field `metrics_scrape_interval`, `scrape_interval` usage of the integration is recommended. (@schmikei)
```
rgeyer, line 16: I'm not sure I understand what's being recommended here.

This states that the integration configuration has changed, and that using the integration is recommended.

I understand the removal of the redundant field, but..

Has the use of the integration not always been the best-practice recommendation? What has changed here?

rgeyer (approved):
Aside from one comment about the changelog (I may just be confused), this LGTM.

CHANGELOG.md
rgeyer: Ahh, okay, the change of the parameters makes much more sense now. I don't think you need to elaborate on the change in the changelog.

I think my confusion comes from the grammar. Maybe a change like this would be more clear, assuming I'm understanding it correctly.

```suggestion
- The `oracledb` integration configuration has removed a redundant field `metrics_scrape_interval`. Use the `scrape_interval` parameter of the integration if a custom scrape interval is required. (@schmikei)
```

jcreixell (approved):

jcreixell:
LGTM, thanks!

PR #3794, opened 2023-05-05
flow/operator: Move label matching into service discovery configs

captncraig:
This is a significant change from how the current operator generates scrape configs, but I think a useful one.

Currently, for each PodMonitor, we generate a kubernetesSD entry that catches ALL pods in the cluster. We also add relabel rules for each specified label match the PodMonitor specifies. This means we will repeatedly be fetching huge lists of pods, and passing them from the discoveryManager to the scrapeManager, only for the scrapeManager to immediately discard almost all of them in the relabeling phase.

This changes a little bit of that. Any labels specified in a Monitor's `Spec.Selector.MatchLabels` will be converted to a selector on the service discovery config. There may be tradeoffs here, but I think we still come out ahead:

### Pros:
- Fewer targets returned by discovery means we use less memory and cpu to relabel and discard uneeded targets.
- We will consume less bandwidth to the k8s api server as less data is transfered.

### Cons:
- We may be shifting the load of filtering into the k8s api server. This change should not add or remove connection (pooling and sharing discoveries across all monitors would be a much larger optimization, but also much more difficult). My gut says the api server should deal with this just fine, but it is the main thing we'd need to watch out for.

There is a note in the prometheus config docs that may be relevant:

> Note: When making decision about using field/label selector make sure that this
 is the best approach - it will prevent Prometheus from reusing single list/watch
 for all scrape configs. This might result in a bigger load on the Kubernetes API,
 because per each selector combination there will be additional LIST/WATCH. On the other hand,
 if you just want to monitor small subset of pods in large cluster it's recommended to use selectors.
 Decision, if selectors should be used or not depends on the particular situation.

I am not sure this is relevant to this situation though, because each SD is running independently. UNLESS there is code deep in the bowels that is combining these queries, I do not think we are saving on list/watch requests.

### DIFF
--- a/component/prometheus/operator/common/crdmanager.go
+++ b/component/prometheus/operator/common/crdmanager.go
@@ -109,13 +109,8 @@ func (c *crdManager) Run(ctx context.Context) error {
 		select {
 		case <-ctx.Done():
 			return nil
-		case pools := <-c.discoveryManager.SyncCh():
-			for pName, groups := range pools {
-				for _, group := range groups {
-					fmt.Println(pName, group.Source, len(group.Targets), "!!!!!")
-				}
-			}
-			targetSetsChan <- pools
+		case m := <-c.discoveryManager.SyncCh():
+			targetSetsChan <- m
 		}
 	}
 }
--- a/component/prometheus/operator/common/crdmanager.go
+++ b/component/prometheus/operator/common/crdmanager.go
@@ -109,8 +109,13 @@ func (c *crdManager) Run(ctx context.Context) error {
 		select {
 		case <-ctx.Done():
 			return nil
-		case m := <-c.discoveryManager.SyncCh():
-			targetSetsChan <- m
+		case pools := <-c.discoveryManager.SyncCh():
+			for pName, groups := range pools {
+				for _, group := range groups {
+					fmt.Println(pName, group.Source, len(group.Targets), "!!!!!")
+				}
+			}
+			targetSetsChan <- pools
 		}
 	}
 }
--- a/component/prometheus/operator/configgen/config_gen.go
+++ b/component/prometheus/operator/configgen/config_gen.go
@@ -32,6 +32,7 @@ func (cg *ConfigGenerator) generateK8SSDConfig(
 	attachMetadata *promopv1.AttachMetadata,
 	matchLabels map[string]string,
 ) *promk8s.SDConfig {
+
 	cfg := &promk8s.SDConfig{
 		Role: role,
 	}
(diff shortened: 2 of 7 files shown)


component/prometheus/operator/common/crdmanager.go
```diff
@@ -109,8 +109,13 @@ func (c *crdManager) Run(ctx context.Context) error {
 		select {
 		case <-ctx.Done():
 			return nil
-		case m := <-c.discoveryManager.SyncCh():
-			targetSetsChan <- m
+		case pools := <-c.discoveryManager.SyncCh():
+			for pName, groups := range pools {
+				for _, group := range groups {
+					fmt.Println(pName, group.Source, len(group.Targets), "!!!!!")
```
rfratto, line 115: ?

rfratto (commented):
I don't have full context for why Prometheus Operator uses relabeling to filter targets instead of doing filtering via API calls.

I suspect it's related to this comment from the [Prometheus configuration documentation](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#kubernetes_sd_config):

> ```yaml
> # Note: When making decision about using field/label selector make sure that this
> # is the best approach - it will prevent Prometheus from reusing single list/watch
> # for all scrape configs. This might result in a bigger load on the Kubernetes API,
> # because per each selector combination there will be additional LIST/WATCH. On the other hand,
> # if you just want to monitor small subset of pods in large cluster it's recommended to use selectors.
> # Decision, if selectors should be used or not depends on the particular situation.
> ```

This comes from how the Prometheus Service Discovery manager will [deduplicate SD mechanisms](https://github.com/prometheus/prometheus/blob/f5fcaa3872ce03808567fabc56afc9cf61c732cb/discovery/manager.go#L423-L427) which are exactly the same.

So this change puts higher load on the API by adding more watch calls per config.

component/prometheus/operator/common/crdmanager.go
captncraig: debug code, removed now.

captncraig:
That deduplication does make a difference. So it would only run one list/watch for ALL pod monitors (or rather unique namespace combinations I think). That would probably make this not worth adding.

Although passing around huge lists of pods is not free either. We could be wasting a lot of memory and cpu relabeling things repeatedly. I think we may need a good way to measure all this before proceeding.

captncraig:
After thinking about it more, I realize that discovery relabeling is not on the hot path for metrics. Discovery should be happening way less often than metrics are coming in, so optimizing a few copy operations on that case may not be worth it, if we are saving concurrent connections to the api server.

PR #3800, opened 2023-05-06
agent-flow: fix S3 object fetch when path has parent directories

### LINKED ISSUE #3784: Flow: S3 file fetch failing when file is under a directory in a bucket
When S3 Module is utilized with path having one or more parent paths then the path splitting logic fails.

Currently if we pass path : `s3://BUCKET_NAME/PARENT_1/FILENAME` then it will assign
- `bucket`: `BUCKET_NAME/PARENT_1`
- `file` : `FILENAME`

But it should assign
- `bucket`: `BUCKET_NAME`
- `file` : `PARENT_1/FILENAME`

When printed error in `getObject` function inside `watcher.go` it was presenting following error:
```
Error: operation error S3: GetObject, https response error StatusCode: 403, RequestID: REPLACED_FOR_SECURITY_REASONS, HostID: REPLACED_FOR_SECURITY_REASONS, api error SignatureDoesNotMatch: The request signature we calculated does not match the signature you provided. Check your key and signing method.
```

## Current Code
https://github.com/grafana/agent/blob/5309e385b64ed0a2e853db0e03ec4ca7b619eab7/component/remote/s3/s3.go#L223-L229

## Fix
```go
func getPathBucketAndFile(path string) (bucket, file string) {
	parts := strings.Split(path, "/")
	file = strings.Join(parts[3:], "/")
	bucket = strings.Join(parts[:3], "/")
	bucket = strings.ReplaceAll(bucket, "s3://", "")
	return
}
```

Version: `v0.33.1`

## References:
[1]: [php - Amazon S3 - How to fix 'The request signature we calculated does not match the signature' error? - Stack Overflow](https://stackoverflow.com/questions/30518899/amazon-s3-how-to-fix-the-request-signature-we-calculated-does-not-match-the-s)

jastisriradheshyam:
## Agent Flow S3 object fetch fix

This PR fix S3 object fetch when path contains one or more parent directories before the filename. 

Fixes #3784

### DIFF
--- a/component/remote/s3/s3.go
+++ b/component/remote/s3/s3.go
@@ -222,8 +222,8 @@ func (s *S3) handleContentPolling(newContent string, err error) {
 // getPathBucketAndFile takes the path and splits it into a bucket and file.
 func getPathBucketAndFile(path string) (bucket, file string) {
 	parts := strings.Split(path, "/")
-	file = parts[len(parts)-1]
-	bucket = strings.Join(parts[:len(parts)-1], "/")
+	file = strings.Join(parts[3:], "/")
+	bucket = strings.Join(parts[:3], "/")
 	bucket = strings.ReplaceAll(bucket, "s3://", "")
 	return
 }


mattdurham:
Looks great and thank you for the PR, can we add a unit test so it doesnt happen in the future?

mattdurham:
Also feel free to add your name and short description to the changelog under bugs.

jastisriradheshyam:
ok, will add a unit test and update the changelog as suggested.

mattdurham:
Thanks! Restarting the windows test then will merge.

mattdurham (approved):
Thank you and LGTM

PR #3803, opened 2023-05-07
Add `kustomization.yaml` in CRDs folder for easy installation of the definitions using Kustomize

ssakuh:

Having the kustomization.yaml allows for referencing this repo in kustomize such that you can update the CRDs based on a reference.

Example:

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
  - github.com/ssakuh/agent/production/operator/crds?ref=patch-1
```

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated


ssakuh:
build does not seem to fail because of my change

captncraig:
Is it possible to put kustomization.yaml in another folder? I am totally ok with enabling kustomize install, but as this is, it would potentially break somebody doing a `kubectl apply -f *.yaml` or something like that.

captncraig:
I don't think this is something we can fully support in this form. We have instructions for people to just install all objects in this folder. If they don't have kustomize, they will get errors doing that with this. If there is a workaround for that, feel free to reopen.

PR #3808, opened 2023-05-08
Add API signatures for module system

mattdurham:
#### PR Description

This is more of a focus on the API signatures and moving some items around. Specifically some code in the ModuleSystem and ModuleDelegates is set to nil/commented out and will be correctly set when doing the functional changes.

Furthermore will merge into module_staging so that when several of the PRs are done they can be merged atomicly.


component/jsontypes.go
```diff
@@ -0,0 +1,28 @@
+package component
```
mattdurham, line 1: Moved this to help separate flow.api and flow so they are no longer tightly coupled.

rfratto: I'd like for this change to be temporary, since it feels weird to me that type definitions used only for the API are moved to a high-level package like this. 

Do you have any long term ideas for how we can keep all API code isolated so that type definitions don't leak into shared packages? 

If we're looking for a temporary solution to work out circular dependencies, what do you think about putting these in a web/api/apitypes package?

pkg/flow/delegate.go
```diff
@@ -0,0 +1,125 @@
...
+		r.URL.Path = path.Join(c.o.HTTPPath, r.URL.Path)
+
+		c.f.ComponentHandler().ServeHTTP(w, r)
+	})
+
+	return r
+}
+
+// ModuleOptions holds static options for module.
+type ModuleOptions struct {
```
rfratto, line 94: The intent is that the Flow controller creates module options, right? 

If that's the case, can we reuse the existing options type instead of defining a new one? 

If not, can we unexport this until we think it must be exported?

mattdurham: I can unexport it, my fear in reusing flow.Options was coupling between the two.

component/jsontypes.go
mattdurham: That sounds reasonable.

component/registry.go
```diff
@@ -28,9 +29,40 @@ var (
...
 
+// ModuleSystem is a mechanism responsible for allowing components to create other components, or
+// flow systems.
+type ModuleSystem interface {
+	// NewModuleDelegate creates a new, un-started ModuleDelegate.
+	NewModuleDelegate(id string) ModuleDelegate
+}
+
+// ModuleDelegate is a controller for running components within a Module.
+type ModuleDelegate interface {
```
rfratto, line 40: Not in love with the names here. What about `ModuleController` and `Module`? My concern with introducing "system" as a term is that people might get confused about if there's a semantic difference between an "X system" and an "X controller," even though the terms are used interchangeably. 

I don't think ModuleDelegate needs to be called a "delegate;" from the perspective of a user it's no different from a module, where they load the module source, run the module, and retrieve an HTTP handler representing the module's API. 

I'm more opinionated that "ModuleDelegate" should be "Module," but I can be convinced that ModuleSystem is good enough for now if there's no opportunity to get confused by both system and controller being used in different places to refer to the same thing.

A third option is to make `NewModule` a function in Options rather than passing an interface, though that would prevent us from ever adding more methods.

component/registry.go
```diff
@@ -28,9 +29,40 @@ var (
...
+	// Run blocks until the provided context is canceled.
+	Run(context.Context)
+
+	// ComponentHandler returns an HTTP handler which exposes endpoints of
+	// components managed by the ModuleDelegate.
+	ComponentHandler() http.Handler
+}
+
+// Export is used for onExport of the ModuleDelegate
+type Export func(exports map[string]any)
```
rfratto, line 58: `Export` as a type name might be a little confusing, since it sounds like it holds data rather than being a function. Can this be called `ExportFunc`, `ExportHandler`, or something along those lines that make it clearer from the name it's a callback?

pkg/flow/delegate.go
```diff
@@ -0,0 +1,125 @@
...
+		o: &delegateOptions{
+			ID:            id,
+			moduleOptions: o,
+		},
+	}
+}
+
+// LoadConfig parses River config and loads it.
+func (c *delegate) LoadConfig(config []byte, args map[string]any, onExport component.Export) error {
+	if c.f == nil {
```
rfratto, line 43: This smells like it's going to need a mutex for protecting c.f. Should we be cautious and add one now?

pkg/flow/delegate.go
```diff
@@ -0,0 +1,125 @@
...
+	return &delegate{
+		o: &delegateOptions{
+			ID:            id,
+			moduleOptions: o,
+		},
+	}
+}
+
+// LoadConfig parses River config and loads it.
+func (c *delegate) LoadConfig(config []byte, args map[string]any, onExport component.Export) error {
```
rfratto, line 42: There's a surprising behavior here: callers are expected to provide a `onExport` value every time `LaodConfig` is called, but only the first `onExport` value from the first call is used. If users try to be clever and supply different values on each call, they will be surprised to learn the value is ignored on subsequent calls. 

IMO `onExport` should be passed when the module is created, not every time when the config is loaded.

web/api/types/types.go
```diff
@@ -0,0 +1,28 @@
+package types
```
rfratto, line 1: nit: `apitypes` instead of `types`; `types` is an overused name and it's not the only package called `types` we import in our code. 

Even though it's redundant in the package path, `apitypes` makes it clearer in code that imports it what kind of types are being used.

rfratto (commented):
Mostly LGTM, some questions about naming and a concern around the design of how to pass the exports changed handler around to modules.

pkg/flow/delegate.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/gorilla/mux"
+	"github.com/grafana/agent/component"
+	"github.com/grafana/agent/pkg/cluster"
+	"github.com/grafana/agent/pkg/flow/logging"
+	"github.com/grafana/agent/web/api"
+	"github.com/prometheus/client_golang/prometheus"
+	"go.opentelemetry.io/otel/trace"
+)
+
+type delegate struct {
```
erikbaranowski, line 17: this type being called only delegate in the flow package is a bit confusing to me since it is actually a module delegate. The file being delegate.go also implies to me it will contain any delegates for flow. I saw @rfratto 's comment about removing the term delegate in general and this would be impacted as well.

component/registry.go
mattdurham: I am not a fan of any of the names TBH, I can live with `controller` and `module`. I just hate how we have something like 11 files with controller in the name/type.

pkg/flow/delegate.go
mattdurham: Will change.

pkg/flow/delegate.go
mattdurham: I concur, I think are some point we should revisit `export` and how it is passed to `flow` since it has always felt a little odd. IMO at some point I think we may need to pass in the calling component as an interface and allow `module.module` to access id, export and other fields as needed but outside the scope of this change.

mattdurham:
@erikbaranowski @rfratto I think I addressed all the feedback here for the naming/layout

pkg/flow/module.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// LoadConfig parses River config and loads it.
+func (c *module) LoadConfig(config []byte, args map[string]any) error {
+	c.mut.Lock()
+	defer c.mut.Unlock()
+	if c.f == nil {
+		f := New(Options{
+			ControllerID:   c.o.ID,
+			Tracer:         nil,
```
rfratto, line 48: Should there be TODOs here for filling out logger and tracer?

pkg/flow/module.go
```diff
@@ -0,0 +1,127 @@
...
+		o: o,
+	}
+}
+
+// LoadConfig parses River config and loads it.
+func (c *module) LoadConfig(config []byte, args map[string]any) error {
+	c.mut.Lock()
+	defer c.mut.Unlock()
+	if c.f == nil {
+		f := New(Options{
```
rfratto, line 46: It's not obvious to me, but where is the unwrapping of logging/tracing/metrics happening? Is that not done yet?

rfratto (approved):
LGTM with two questions

pkg/flow/module.go
mattdurham: I didnt add a todo because the next PR fills it all in.

PR #3810, opened 2023-05-08
Update grafana agent docs to indicate core purpose

moxious:
#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/_index.md
+++ b/docs/sources/_index.md
@@ -5,20 +5,7 @@ weight: 1
 
 # Grafana Agent
 
-Grafana Agent is a vendor-neutral, batteries-included telemetry collector with
-configuration inspired by [Terraform][]. It is designed to be flexible,
-performant, and compatible with multiple ecosystems such as Prometheus and
-OpenTelemetry.
-
-Grafana Agent is based around **components**. Components are wired together to
-form programmable observability **pipelines** for telemetry collection,
-processing, and delivery.
-
-> **NOTE**: This page focuses mainly on "[Flow mode][Grafana Agent Flow]," the
-> Terraform-inspired variant of Grafana Agent.
->
-> For information on other variants of Grafana Agent, refer to [About Grafana
-> Agent][].
+Grafana Agent is software that can run on any of your hosts, and which collects events and metrics from hosts and sends them to Grafana.  Grafana Agent is capable of gathering data for [all three pillars of observability](https://grafana.com/docs/grafana-cloud/fundamentals/intro-to-observability/); logs, metrics, and traces.
 
 Grafana Agent can collect, transform, and send data to:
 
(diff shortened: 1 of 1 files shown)


docs/sources/_index.md
```diff
@@ -5,20 +5,7 @@ weight: 1
...
-Grafana Agent is based around **components**. Components are wired together to
-form programmable observability **pipelines** for telemetry collection,
-processing, and delivery.
-
-> **NOTE**: This page focuses mainly on "[Flow mode][Grafana Agent Flow]," the
-> Terraform-inspired variant of Grafana Agent.
->
-> For information on other variants of Grafana Agent, refer to [About Grafana
-> Agent][].
+Grafana Agent is software that can run on any of your hosts, and which collects events and metrics from hosts and sends them to Grafana.  Grafana Agent is capable of gathering data for [all three pillars of observability](https://grafana.com/docs/grafana-cloud/fundamentals/intro-to-observability/); logs, metrics, and traces.
```
rfratto, line 8: Thank you for working on this. I have a few concerns with this rewrite: 

* The "inspired by Terraform" mention is useful for getting people excited who have some familiarity with Terraform, and there's no other place where it's mentioned. 
* The list of pillars of observability omits profiles. Profiles are a newer pillar, and omitting mention of it may mislead people to think we don't support it. 
* The "sends them to Grafana" is downplaying the vendor neutrality of Grafana Agent, which is an important message we want to continue conveying. 
* I'm overall concerned we have a limited opportunity to capture the attention of readers. If this is the first page a reader visits, I would argue we should be trying to get them interested as soon as possible. 

I would personally prefer explaining what a telemetry collector is to a second paragraph:

> Grafana Agent is a vendor-neutral, batteries-included telemetry collector with
configuration inspired by [Terraform][]. It is designed to be flexible,
performant, and compatible with multiple ecosystems such as Prometheus and
OpenTelemetry.
> 
> A telemetry collector is software which collects telemetry signals from machines and sends them to a database. Grafana Agent is capable of gathering data for the primary pillars of observability: logs, metrics, traces, and continuous profiles. 

cc @clayton-cornell

moxious: @rfratto PR was intentionally put in to start a discussion to include Clayton rather than to settle the matter.

As a going in position, I think it's useful to position the overview page towards a fairly intro-level understanding of the space.  It shouldn't be presumed that people know what a telemetry collector is.   In that spirit, this language:

> vendor-neutral, batteries-included telemetry collector with configuration inspired by [Terraform][]

is accurate, but fairly opaque to newbies.

Rather than differing though on what the page should say or how best phrased, it might be most fruitful to think about who the audience is and what we want someone to learn when they hit this page.   For me as someone who focuses on educating developers or early customers it's:

1. What is this thing?  (It's a telemetry collector, what's that?)
2. What can I do with it?  (By itself, collectors may not be useful, they forward data to something else that does something cool)
3. How can I get it installed quickly and do something minimally useful with it? 

It's going to be tricky how to balance all of the different perspectives and signaling.  For example I understand the point about vendor neutrality and not just "it sends them to grafana" as positioning element about the agent itself, but again the question becomes "who is the audience of this page"....would it not be primarily people using the agent for that purpose?

PR #3816, opened 2023-05-09
docs: update prometheus links to point to the latest version

theSuess:
#### PR Description

Similar to #1859 this updates the documentation links as version `2.42` has been removed from the Prometheus website. By changing it to `latest`, it should be available longer at the cost of maybe referencing features which are not yet available in the agent.

#### Notes to the Reviewer

If it makes more sense to pin a specific version, just let me know and I'll figure out which one is currently used 

#### PR Checklist

- [x] ~~CHANGELOG updated~~
- [x] Documentation added
- [x] ~~Tests updated~~

### DIFF
--- a/docs/sources/static/configuration/metrics-config.md
+++ b/docs/sources/static/configuration/metrics-config.md
@@ -259,7 +259,7 @@ remote_write:
   - [<remote_write>]
 ```
 
-> **Note:** For more information on remote_write, refer to the [Prometheus documentation](https://prometheus.io/docs/prometheus/2.42/configuration/configuration/#remote_write)
+> **Note:** For more information on remote_write, refer to the [Prometheus documentation](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write)
 
 ## metrics_instance_config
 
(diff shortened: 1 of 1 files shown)


rfratto:
Hi, sorry it took so long to review this. 

> If it makes more sense to pin a specific version, just let me know and I'll figure out which one is currently used

I do think it makes more sense to pin to a specific version, since it's fairly difficult to find out what version we're actually using, and it might not be immediately clear that the latest docs aren't what's being used in the agent. 

At the time of writing, we _are_ using 2.42, but I'm not sure why those links don't work anymore, since it still shows up for me: https://prometheus.io/docs/prometheus/2.42/configuration/configuration/

clayton-cornell:
> linking to a pinned version of dependencies vs latest

This is a tough one. The Writer's Toolkit doesn't provide a lot of guidance here. 

Generally, when you're deep linking to external sites you run the risk of links going stale and eventually breaking. This adds maintenance steps in the future (validating links, repairing 404s, etc.). Where possible we should try to reduce external linking. When we need to deep link into another website, it's better to use `latest` where possible. 

In this case, since we are referring to a specific version of Prometheus, we should probably link to that version unless having 2.44 (and higher) topic content isn't going to confuse things). If we leave at a deep link to 2.42, then we need to make sure that when we internally bump to 2.43 of higher that we update all deep links.

theSuess:
> At the time of writing, we are using 2.42, but I'm not sure why those links don't work anymore, since it still shows up for me: https://prometheus.io/docs/prometheus/2.42/configuration/configuration/

Hm, those links seem to be working again. However, the live docs still point to https://prometheus.io/docs/prometheus/2.34/configuration/configuration/#relabel_config :thinking:. I'll do some further digging around the publishing pipeline and report back

theSuess:
Alright, the issue is pretty simple: https://github.com/grafana/agent/commit/d7e3c37365c98d8d808dcba3d5d3371aed79aefd is not yet on the `release-0.33` branch. Cherypicking this commit to the release branch should get the docs working as well so I'll close this PR

PR #3817, opened 2023-05-09
river/stdlib: add nonsensitive function

rfratto:
> **NOTE TO REVIEWER**: I had to do some refactoring to make this PR possible without import cycles. I recommend reviewing commit-by-commit. 

The `nonsensitive` function allows users to explicitly convert a secret value back to a string.
 
`nonsensitive` is primarily useful when combined with the remote.vault component, since remote.vault always returns secrets, even if some data stored in Vault is non-sensitive (such as a TLS public key).

`nonsensitive` is the equivalent to the Terraform function [of the same name](https://developer.hashicorp.com/terraform/language/functions/nonsensitive).


ptodev:
I personally like the approach with `nonsensitive` as opposed to trying to add extra non-secret config to `remote.vault` for two reasons:
* It's quite flexible and could work with other components which only export secrets in the future.
* It also gives the user full responsibility regarding what gets converted from a secret. The less under-the-hood-magic is happening with secrets, the better.

ptodev (approved):

docs/sources/flow/reference/stdlib/nonsensitive.md
```diff
@@ -0,0 +1,28 @@
...
+
+# nonsensitive
+
+`nonsensitive` converts a [secret][] value back into a string.
+
+> **WARNING**: Only use `nonsensitive` when you are positive that the value
+> being converted back to a string is not a sensitive value.
+>
+> Strings resulting from calls to `nonsensitive` will be shown in plaintext in
+> the UI.
```
mattdurham, line 15: `and API`? Or maybe a broader term?

mattdurham (approved):
Small nit but looks good.

docs/sources/flow/reference/stdlib/nonsensitive.md
rfratto: I put "and HTTP calls to the UI," since the API the UI uses is undocumented and essentially internal with no stability guarantees.

Issue #2861, opened 2023-01-30
flow: clustering auto-distribution


tpaschalis:
We should find a way to allow Flow components to automatically distribute load across a cluster for feature parity with scraping service. Let's sketch out how this might look like!

mattdurham:
Closing since the only outstanding item is making it the default and that can be talked about later

PR #3831, opened 2023-05-10
helm: add headless service

tpaschalis:
#### PR Description

This PR adds a headless service to our helm chart. Its main use case is for use with clustering, so that new pods can use it for an SRV lookup to fins the rest of the cluster peers to connect to.

It is a copy of the existing service definition but with a `-headless` suffix to the name and ClusterIP set explicitly to None.

Comparing the previous manifests (exported with `helm template .`) and the new ones gives:

```diff
$ diff manifests-old.yaml manifests-new.yaml
156a157,178
> # Source: grafana-agent/templates/service-headless.yaml
> apiVersion: v1
> kind: Service
> metadata:
>   name: RELEASE-NAME-grafana-agent-headless
>   labels:
>     helm.sh/chart: grafana-agent-0.13.0
>     app.kubernetes.io/name: grafana-agent
>     app.kubernetes.io/instance: RELEASE-NAME
>     app.kubernetes.io/version: "v0.33.1"
>     app.kubernetes.io/managed-by: Helm
> spec:
>   clusterIP: 'None'
>   selector:
>     app.kubernetes.io/name: grafana-agent
>     app.kubernetes.io/instance: RELEASE-NAME
>   ports:
>     - name: http-headless-service
>       port: 80
>       targetPort: 80
>       protocol: "TCP"
> ---
```

Related to #2861 and ease-of-use.

#### Notes to the Reviewer
Nothing to note.

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [X] Tests updated


rfratto (commented):
Thinking about this more: what if we made the service toggleable and then just relied on turning the primary service into a headless service instead of adding a new service?

operations/helm/charts/grafana-agent/manifests-new.yaml
```diff
@@ -0,0 +1,271 @@
+---
```
rfratto: Accidental check in? (Ditto with manifests-old)

rfratto (approved):

PR #3832, opened 2023-05-10
helm: allow using HPA autoscaling with StatefulSets

tpaschalis:
#### PR Description
This PR amends the HPA manifest to be used not only for deployments but also for StatefulSets, to be used in clustering mode.

Related to https://github.com/grafana/agent/issues/2861 and ease-of-use.

#### Notes to the Reviewer

As we've discussed, this will leak volumes when the HPA scales up and then down.

One option would be to switch using ephemeral volumes, but would lead to losing the WAL between upgrades. Should we switch to that when autoscaling is defined? If we don't find a solution to clearing up the remaining volumes we should at least document the behavior so people know what to look for and how to clean them up.
 
#### PR Checklist

- [X] CHANGELOG updated
- [ ] Documentation added
- [X] Tests updated


operations/helm/charts/grafana-agent/templates/hpa.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if and (eq .Values.controller.type "deployment") .Values.controller.autoscaling.enabled }}
+{{- if and (or (eq .Values.controller.type "deployment") (eq .Values.controller.type "statefulset" ))  .Values.controller.autoscaling.enabled }}
```
tpaschalis, line 1: We could simplify this to `(not eq .Values.controller.type "daemonset")` if you think it's preferred.

tpaschalis:
In my low-power DigitalOcean cluster, you can't get pod-level metrics

```
$ k top pod
error: Metrics API not available
```

As such, when checking out the HPA, the used memory is reported as `<unknown>`, and the HPA complains about not being able to get the target's scale.


```bash
$ kg hpa
NAME                  REFERENCE                        TARGETS         MINPODS   MAXPODS   REPLICAS   AGE
agent-grafana-agent   Deployment/agent-grafana-agent   <unknown>/80%   2         5         0          4m3s

$ k describe hpa agent-grafana-agent
Name:                                                     agent-grafana-agent
Namespace:                                                agent
Labels:                                                   app.kubernetes.io/instance=agent
                                                          app.kubernetes.io/managed-by=Helm
                                                          app.kubernetes.io/name=grafana-agent
                                                          app.kubernetes.io/version=v0.33.1
                                                          helm.sh/chart=grafana-agent-0.13.0
Annotations:                                              meta.helm.sh/release-name: agent
                                                          meta.helm.sh/release-namespace: agent
CreationTimestamp:                                        Wed, 10 May 2023 14:48:01 +0300
Reference:                                                Deployment/agent-grafana-agent
Metrics:                                                  ( current / target )
  resource memory on pods  (as a percentage of request):  <unknown> / 80%
Min replicas:                                             2
Max replicas:                                             5
Deployment pods:                                          0 current / 0 desired
Conditions:
  Type         Status  Reason          Message
  ----         ------  ------          -------
  AbleToScale  False   FailedGetScale  the HPA controller was unable to get the target's current scale: deployments/scale.apps "agent-grafana-agent" not found
Events:
  Type     Reason          Age                  From                       Message
  ----     ------          ----                 ----                       -------
  Warning  FailedGetScale  9s (x16 over 3m54s)  horizontal-pod-autoscaler  deployments/scale.apps "agent-grafana-agent" not found
```

Is it something you've seen before and should we be warning users about it?

rfratto (approved):
LGTM, ship it!

rfratto:
> One option would be to switch using ephemeral volumes, but would lead to losing the WAL between upgrades. Should we switch to that when autoscaling is defined?

I have some suggestions for this that you might want to implement before merging this PR, but I'll need to get to a computer first.

tpaschalis:
No rush to merge, take your time and we can discuss it first.

rfratto:
> I have some suggestions for this that you might want to implement before merging this PR, but I'll need to get to a computer first.

I recommend we do two things: 

1. Make usage of `persistentVolumeClaimRetentionPolicy` like [Loki does](https://github.com/grafana/loki/blob/d7dd2e9f74f901215cec0e22a6b92291a8d64108/production/helm/loki/templates/read/statefulset-read.yaml#L19-L27). Note that because it's behind an alpha feature gate, we can't use it personally, so it will need to be a toggle you could optionally turn on in the values.yaml. 

2. Definitely document it. Leaking is only possible when:
 
   * You are using the StatefulSet controller, AND 
   * You are using autoscaling, AND 
   * You configured volumeClaimTemplates to be created with the StatefulSet, AND 
   * You cannot enable persistentVolumeClaimRetentionPolicy on your cluster. 
  
   Furthermore, it only leaks up to the maximum replicas of PVCs controlled by the HPA, rather than creating new PVCs infinitely. 

   Because of all of the above, I don't think it's extra scary, but it should be something users are aware of when it applies to them and what the alternatives are (don't use PVCs, etc).

operations/helm/tests/create-statefulset-autoscaling/grafana-agent/templates/hpa.yaml
```diff
@@ -0,0 +1,26 @@
...
+  labels:
+    helm.sh/chart: grafana-agent
+    app.kubernetes.io/name: grafana-agent
+    app.kubernetes.io/instance: grafana-agent
+    app.kubernetes.io/version: "vX.Y.Z"
+    app.kubernetes.io/managed-by: Helm
+spec:
+  scaleTargetRef:
+    apiVersion: apps/v1
+    kind: Deployment
```
laurovenancio, line 16: The scaleTargetRef also has to be changed.

PR #3834, opened 2023-05-10
Update and rename config.yaml to config.yml

rfratto:
GitHub requires the config file for issue templates to be called `config.yml` and _not_ `config.yaml`.


mattdurham (approved):
LGTM

PR #3840, opened 2023-05-10
Fix module example

ptodev:
Minor change to fix the example in the module docs.


erikbaranowski (approved):
Looks good, thanks for catching this.

PR #3849, opened 2023-05-11
agent-flow-mixin: add more alerts around clustering

### LINKED ISSUE #3758: Add clustering-specific dashboards and alerts to agent-flow-mixin
We should provide a set of dashboards and alerts that users can easily set up using our `agent-flow-mixin` and start having some visibility into how their cluster works.

tpaschalis:
#### PR Description
This PR adds some more alerts around clustering.

Fixes #3758 

#### Notes to the Reviewer
My PromQL is bad, so let me know if you spot anything amiss. The queries _do_ work with our internal data.

#### PR Checklist

- [ ] CHANGELOG updated (N/A)
- [ ] Tests updated (N/A)

### DIFF
--- a/operations/agent-flow-mixin/alerts/clustering.libsonnet
+++ b/operations/agent-flow-mixin/alerts/clustering.libsonnet
@@ -6,7 +6,7 @@ alert.newGroup(
     // Cluster not converging.
     alert.newRule(
       'ClusterNotConverging',
-      'stddev by (cluster, namespace) ((sum without (state) (cluster_node_peers))) != 0',
+      'stddev by (cluster, namespace) (sum without (state) (cluster_node_peers)) != 0',
       'Cluster is not converging.',
       '5m',
     ),
@@ -38,7 +38,7 @@ alert.newGroup(
     alert.newRule(
       'ClusterNodeUnhealthy',
       |||
-        sum by (cluster, namespace, instance) (cluster_node_gossip_health_score) > 0
+        cluster_node_gossip_health_score > 0
       |||,
       'Cluster node is reporting a health score > 0.',
       '5m',
@@ -63,7 +63,7 @@ alert.newGroup(
     // Node stuck in Terminating state.
     alert.newRule(
       'ClusterNodeStuckTerminating',
-      'sum by (cluster, namespace, instance) (cluster_node_peers{state="terminating"}) != 0',
+      'sum by (cluster, namespace, instance) (cluster_node_peers{state="terminating"}) > 0',
       'Cluster node stuck in Terminating state.',
       '5m',
     ),
--- a/operations/agent-flow-mixin/alerts/clustering.libsonnet
+++ b/operations/agent-flow-mixin/alerts/clustering.libsonnet
@@ -68,33 +68,7 @@ alert.newGroup(
       '5m',
     ),
 
-    // Failed packet writes.
-    alert.newRule(
-      'ClusterNodePacketWritesFailing',
-      |||
-        1 - (
-        sum by (cluster, namespace) (rate(cluster_transport_tx_packets_failed_total[2m])) /
-        sum by (cluster, namespace) (rate(cluster_transport_tx_packets_total[2m]))
-        ) < 1
-
-      |||,
-      'Cluster node is having its packet writes failing.',
-      '5m',
-    ),
-
-    // Failed stream writes.
-    alert.newRule(
-      'ClusterNodeStreamWritesFailin',
-      |||
-        1 - (
-        sum by (cluster, namespace) (rate(cluster_transport_stream_tx_packets_failed_total[2m])) /
-        sum by (cluster, namespace) (rate(cluster_transport_stream_tx_packets_total[2m]))
-        ) < 1
-      |||,
-      'Cluster node is having its packet writes failing.',
-      '5m',
-    ),
-
-    // TODO(@tpaschalis) Alert on open transport streams once we investigate their behavior.
+    // TODO(@tpaschalis) Alert on open transport streams once we investigate
+    // their behavior.
   ]
 )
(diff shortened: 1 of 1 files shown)


operations/agent-flow-mixin/alerts/clustering.libsonnet
```diff
@@ -25,5 +25,50 @@ alert.newGroup(
...
       '5m',
     ),
+
+    // Standard Deviation of Lamport clock time between nodes is too high
+    alert.newRule(
+      'ClusterLamportClockDrift',
+      'stddev by (cluster, namespace) (cluster_node_lamport_time) != 0',
+      "Cluster nodes' lamport clocks are not converging.",
+      '5m'
+    ),
```
rfratto, line 35: I think this one needs a little tweaking. Lamport clocks won't ever converge on large clusters, but you just don't want it to drift _too_ much; we don't want clocks to deviate too far from the mean clock value. 

We should start with `> 2` here, since conventionally a deviation of +/-2 is considered to be statically significant (according to [Wikipedia](https://en.wikipedia.org/wiki/Standard_deviation)). 

We should also change the description to talk more about the deviation:

```
Cluster nodes' lamport clocks time has deviated too far from the mean.
```

operations/agent-flow-mixin/alerts/clustering.libsonnet
```diff
@@ -25,5 +25,50 @@ alert.newGroup(
...
+
+    // Nodes health score is not zero.
+    alert.newRule(
+      'ClusterNodeUnhealthy',
+      |||
+        sum by (cluster, namespace, instance) (cluster_node_gossip_health_score) > 0
```
rfratto, line 41: There's nothing to aggregate here, since there's just one of these metrics per node, and we want to track it at the node level.

```suggestion
        cluster_node_gossip_health_score > 0
```

operations/agent-flow-mixin/alerts/clustering.libsonnet
```diff
@@ -25,5 +25,50 @@ alert.newGroup(
...
+      |||,
+      'Cluster node is reporting a health score > 0.',
+      '5m',
+    ),
+
+    // Lamport clock of a node is not progressing at all.
+    alert.newRule(
+      'ClusterLamportClockStuck',
+      'sum by (cluster, namespace, instance) (rate(cluster_node_lamport_time[2m])) == 0',
```
rfratto, line 50: nit: use `increase` instead of `rate` here, we don't care what the per-second rate is, we just want to know if it changed in the last two minutes.

operations/agent-flow-mixin/alerts/clustering.libsonnet
```diff
@@ -25,5 +25,50 @@ alert.newGroup(
...
+      'ClusterNodeNameConflict',
+      'sum by (cluster, namespace) (rate(cluster_node_gossip_received_events_total{event="node_conflict"}[2m])) > 0',
+      'A node tried to join the cluster with a name conflicting with an existing peer.',
+      '5m',
+    ),
+
+    // Node stuck in Terminating state.
+    alert.newRule(
+      'ClusterNodeStuckTerminating',
+      'sum by (cluster, namespace, instance) (cluster_node_peers{state="terminating"}) != 0',
```
rfratto, line 66:
```suggestion
      'sum by (cluster, namespace, instance) (cluster_node_peers{state="terminating"}) > 0',
```

rfratto (approved):

rfratto:
Should we add an alert for config drift using `agent_config_hash`? We'd only want to alert on that for agents in the same namespace which have clustering enabled, though.

operations/agent-flow-mixin/alerts/clustering.libsonnet
rfratto: I'm not sure if that expression will catch every scenario that we'd want to alert on, since the drift you're calculating is used in the calculation of stddev. 

I think if there's only one outlier, it wouldn't be alerted on since its own deviation is not double of itself. 

Consider a lamport clock set of (2, 2, 2, 2, 20). We want to alert on the 20 here. Your alert would resolve to

```
abs(20 - 5.6) > 2 * 7.2  => 
abs(14.4) > 14.4 => 
false 
```

And then if you add a second problematic node (2, 20, 2, 2, 20), nothing would be alerted on, because no node deviates more from the mean than twice the stddev value ~8.82, even though there's clearly a problem here. 

Let's take a step back:

We're trying to validate that the lamport clock is functioning properly, right? Lamport clocks must be roughly synchronized between nodes, where clocks are synchronized through sending gossiped messages. Any time a node N receives a newer clock time T, it sets its own clock time to T+1. 

I think we can pick an arbitrary value here to say that the standard deviation should be less than some target of clock ticks: clocks should always roughly be within X ticks from the mean (let's say 4 for now). If the stddev goes above 4, it means that there's a subset of nodes that are not gossiping correctly. 

What do you think?

tpaschalis:
> Should we add an alert for config drift using agent_config_hash? We'd only want to alert on that for agents in the same namespace which have clustering enabled, though.

@rfratto yeah, that sounds like a good idea. Here's what my limited PromQL-fu got me, would you mind sanity-checking?

```
sum by (cluster, namespace) (cluster_node_info)
!=
on (cluster, namespace) group_left avg by (cluster, namespace) ((count by (cluster, namespace, sha256) (agent_config_hash)))
```

rfratto:
> ```sum by (cluster, namespace) (cluster_node_info)
> !=
> on (cluster, namespace) group_left avg by (cluster, namespace) ((count by (cluster, namespace, sha256) (agent_config_hash)))
>```


I think there's a slightly more clear way of expressing what you're looking for here:

```
count without (sha256) ( 
    max by (cluster, namespace, sha256) (agent_config_hash and on(cluster, namespace) cluster_node_info)
) > 1 
```

The inner `max` creates one time series for each SHA256 found for agents which are running clustering (`agent_config_hash and on(cluster, namespace) cluster_node_info`). The outer `count` then returns the number of time series there were from the inner expression. 

If there is more than one series, it means you have two sets of clustered agents in the same namespace that both have different configuration files.

operations/agent-flow-mixin/alerts/clustering.libsonnet
tpaschalis: Yeah, let's go with this approach and see how it works in practice.

PR #3854, opened 2023-05-11
operator: add logging if we fail to update a controller

captncraig:
There are some cases where kubernetes will not allow you to update a StatefulSet if certain fields have changed. In that case, we will delete and recreate it as a fallback.

This adds a little logging to help diagnose that.


spartan0x117 (approved):
LGTM!

PR #3855, opened 2023-05-11
flow: allow loading a configuration "source" from a folder

rfratto:
> **NOTE TO REVIEWER**: This PR contains some refactoring; review the PR commit-by-commit to make things easier. 

This PR allows `grafana-agent run` to be given a directory. When given a directory, it will search for all *.river files in that directory (ignoring nested directories) and load them in as a single configuration source. 

This is different from modules; files in the same directory must have unique component names, and can not repeat configuration blocks. 

This allows users to use directories to split up configurations by domain for readability but still share components, similar to Terraform modules or Go packages. 

Module loaders are not modified, and still only support a loading a single configuration source, but could be updated in the future to allow similar behavior, such as pointing module.git at a directory.

### DIFF
--- a/cmd/internal/flowmode/cmd_run.go
+++ b/cmd/internal/flowmode/cmd_run.go
@@ -184,6 +184,7 @@ func (fr *flowRun) Run(configFile string) error {
 
 	reload := func() (*flow.Source, error) {
 		flowSource, err := loadFlowSource(configFile)
+		instrumentation.InstrumentSHA256(flowSource.SHA256())
 		defer instrumentation.InstrumentLoad(err == nil)
 
 		if err != nil {
@@ -349,9 +350,6 @@ func loadFlowSource(filename string) (*flow.Source, error) {
 	if err != nil {
 		return nil, err
 	}
-
-	instrumentation.InstrumentConfig(bb)
-
 	return flow.ParseSource(filename, bb)
 }
 
--- a/cmd/internal/flowmode/cmd_run.go
+++ b/cmd/internal/flowmode/cmd_run.go
@@ -4,11 +4,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io/fs"
 	"net"
 	"net/http"
 	"os"
 	"os/signal"
 	"path"
+	"path/filepath"
+	"strings"
 	"sync"
 	"syscall"
 
@@ -49,14 +52,17 @@ func runCommand() *cobra.Command {
 	}
 
 	cmd := &cobra.Command{
-		Use:   "run [flags] file",
+		Use:   "run [flags] path",
 		Short: "Run Grafana Agent Flow",
 		Long: `The run subcommand runs Grafana Agent Flow in the foreground until an interrupt
 is received.
 
-run must be provided an argument pointing at the River file to use. If the
-River file wasn't specified, can't be loaded, or contains errors, run will exit
+run must be provided an argument pointing at the River path to use. If the
+River path wasn't specified, can't be loaded, or contains errors, run will exit
 immediately.
+
+If path is a directory, all *.river files in that directory will be combined
+into a single unit. Subdirectories are not searched.
 
 run starts an HTTP server which can be used to debug Grafana Agent Flow or
 force it to reload (by sending a GET or POST request to /-/reload). The listen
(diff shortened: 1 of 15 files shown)


cmd/internal/flowmode/cmd_run.go
```diff
@@ -346,15 +351,73 @@ func getEnabledComponentsFunc(f *flow.Flow) func() map[string]interface{} {
...
+
+		if strings.HasSuffix(path, ".river") || path == pathName {
+			bb, err := os.ReadFile(path)
+			if err != nil {
+				return err
+			}
+			fileContents[path] = bb
+
+			// TODO(rfratto): this breaks when there's multiple files
+			instrumentation.InstrumentConfig(bb)
```
rfratto, line 379: Note that multiple files breaks InstrumentConfig, since it's only meant to track a single config at a time, and we'd need another solution for folders.

cmd/internal/flowmode/cmd_run.go
```diff
@@ -346,15 +351,50 @@ func getEnabledComponentsFunc(f *flow.Flow) func() map[string]interface{} {
 	}
 }
 
-func loadFlowFile(filename string) (*flow.File, error) {
-	bb, err := os.ReadFile(filename)
+func loadFlowSource(path string) (*flow.Source, error) {
```
erikbaranowski, line 354: The contents of this feels like it may belong in the pkg/flow space with a light wrapper here like it prior. This would also make us 1 step closer to other things (ie. modules) reusing this logic.

mattdurham:
Can we add some tests for load sources that include multiple files? Also would be helpful to add file notation to config blocks  around `populateConfigBlockNodes`.

Using the test 
```river

func TestSources(t *testing.T) {
	content := `
        logging {
		    log_format = "json"
		}

		testcomponents.tick "ticker_a" {
			frequency = "1s"
		}
	`

	content2 := `
        logging {
		    log_format = "json"
		}

		testcomponents.tick "ticker_a" {
			frequency = "1s"
		}
	`

	sources := make(map[string][]byte)
	sources["t1"] = []byte(content)
	sources["t2"] = []byte(content2)

	s, err := ParseSources(sources)
	require.NoError(t, err)
	ctrl := New(testOptions(t))
	err = ctrl.LoadSource(s, nil)
	require.Error(t, err)
}
```

I get `Component testcomponents.tick.ticker_a already declared at t1:6:3` which is awesome, but get `"logging" block already declared` for the logging block which is less awesome.

pkg/flow/source.go
```diff
@@ -0,0 +1,152 @@
...
+func ParseSource(name string, bb []byte) (*Source, error) {
+	node, err := parser.ParseFile(name, bb)
+	if err != nil {
+		return nil, err
+	}
+
+	// Look for predefined non-components blocks (i.e., logging), and store
+	// everything else into a list of components.
+	//
+	// TODO(rfratto): should this code be brought into a helper somewhere? Maybe
```
erikbaranowski, line 38: Not for this PR since I noticed after the fact this is really moved code, but I agree the logic for what blocks are config blocks should be moved. More specifically, there is overlap in pkg/flow/internal/controller/config.go now as well with the constants for what blocks are config blocks.

cmd/internal/flowmode/cmd_run.go
```diff
@@ -49,15 +52,18 @@ func runCommand() *cobra.Command {
...
 	cmd := &cobra.Command{
-		Use:   "run [flags] file",
+		Use:   "run [flags] path",
 		Short: "Run Grafana Agent Flow",
 		Long: `The run subcommand runs Grafana Agent Flow in the foreground until an interrupt
 is received.
 
-run must be provided an argument pointing at the River file to use. If the
-River file wasn't specified, can't be loaded, or contains errors, run will exit
+run must be provided an argument pointing at the River path to use. If the
```
erikbaranowski, line 60: somewhere in here I think we should explicitly say we also accept a filepath. This through me off for a minute until I read the code.

erikbaranowski (commented):
Looks solid overall. I added a few thoughts from me for your consideration.

rfratto:
Heads up, this has fallen down on my priority list but I hope to address comments in the next few days.

rfratto:
I'm going to close this for now, since I haven't been able to find the time to finish this up. If I have time in the future I'll re-open, otherwise someone else is free to continue this work if they're interested.

PR #3862, opened 2023-05-11
river-jsonnet: support block indexes

rfratto:
Support adding an index argument to river.block to ensure that blocks get marshaled in a specific order.


erikbaranowski (approved):
LGTM

PR #3864, opened 2023-05-11
docs: remove patch versions from the upgrade guide unless required

rfratto:
This removes patch versions from the upgrade guide, since upgrade guide applies to minor versions as a whole and not individual patch releases.

The only instance where the patch release mattered was for a security release in the past; this has been explicitly listed with the patch version.

### DIFF
--- a/docs/sources/static/upgrade-guide.md
+++ b/docs/sources/static/upgrade-guide.md
@@ -21,10 +21,10 @@ static mode.
 
 ### Removal of Dynamic Configuration
 
-The experimental feature Dynamic Configuration has been removed. The use case of dynamic configuration will be replaced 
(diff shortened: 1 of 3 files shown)


captncraig (approved):

docs/sources/static/upgrade-guide.md
```diff
@@ -49,7 +49,7 @@ options for the `diskstats` collector have changed names:
 - `ignored_devices` is now `device_exclude` in the Flow component
   configuration.
 
-## v0.31.1
+## v0.31
```
erikbaranowski, line 52: you've got 2 v0.31's here and below

erikbaranowski (approved):
1 spot to check how you want to handle with v0.31

PR #3866, opened 2023-05-11
[release-v0.33] Backport #3864 to release v0.33

rfratto:
This removes patch versions from the upgrade guide, since upgrade guide applies to minor versions as a whole and not individual patch releases.

The only instance where the patch release mattered was for a security release in the past; this has been explicitly listed with the patch version.

(cherry picked from commit db7f9f28ff625e4ea82c464f97af4dec17cc9c4f)


spartan0x117 (approved):
LGTM!

PR #3872, opened 2023-05-12
feat(gcplog): import feature from loki repository

eraac:
#### PR Description

Import changes from Promtail https://github.com/grafana/loki/pull/9044
Discussion: https://github.com/grafana/agent/discussions/3842

#### Notes to the Reviewer

It's not feel right the way I've done the "lift and shifting", because of the light edits from the Loki. If something is wrong tell me

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [X] Tests updated

### DIFF
--- a/component/loki/source/gcplog/internal/gcplogtarget/pull_target.go
+++ b/component/loki/source/gcplog/internal/gcplogtarget/pull_target.go
@@ -4,6 +4,11 @@ package gcplogtarget
 // configure and run the targets that can read log entries from cloud resource
 // logs like bucket logs, load balancer logs, and Kubernetes cluster logs
 // from GCP.
+// /!\ But
+// - newPullTarget is transform to NewPullTarget
+// - scrapeconfig import is deleted
+// - import 'github.com/grafana/agent/component/common/loki' is added, and api.* is replace by loki.*
+// - change return 'pullTarget.Details', from `interface{}' to 'map[string]string'
 
 import (
 	"context"
@@ -19,11 +24,25 @@ import (
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/model/relabel"
 	"google.golang.org/api/option"
+
+	"github.com/grafana/loki/clients/pkg/promtail/targets/target"
 )
 
-// PullTarget represents a target that scrapes logs from a GCP project id and
-// subscription and converts them to Loki log entries.
-type PullTarget struct {
+var defaultBackoff = backoff.Config{
+	MinBackoff: 1 * time.Second,
+	MaxBackoff: 10 * time.Second,
+	MaxRetries: 0, // Retry forever
+}
+
+// pubsubSubscription allows us to mock pubsub for testing
+type pubsubSubscription interface {
+	Receive(ctx context.Context, f func(context.Context, *pubsub.Message)) error
+}
+
+// pullTarget represents the target specific to GCP project, with a pull subscription type.
+// It collects logs from GCP and push it to Loki.
+// nolint:revive
+type pullTarget struct {
 	metrics       *Metrics
 	logger        log.Logger
 	handler       loki.EntryHandler
@@ -43,20 +62,21 @@ type PullTarget struct {
 	msgs chan *pubsub.Message
 }
 
-// TODO(@tpaschalis) Expose this as River configuration in the future.
-var defaultBackoff = backoff.Config{
-	MinBackoff: 1 * time.Second,
-	MaxBackoff: 10 * time.Second,
-	MaxRetries: 0, // Retry forever
-}
-
-// pubsubSubscription allows us to mock pubsub for testing
-type pubsubSubscription interface {
-	Receive(ctx context.Context, f func(context.Context, *pubsub.Message)) error
-}
-
-// NewPullTarget returns the new instance of PullTarget.
-func NewPullTarget(metrics *Metrics, logger log.Logger, handler loki.EntryHandler, jobName string, config *PullConfig, relabel []*relabel.Config, clientOptions ...option.ClientOption) (*PullTarget, error) {
+// NewPullTarget returns the new instance of pullTarget for
+// the given `project-id`. It scraps logs from the GCP project
+// and push it Loki via given `loki.EntryHandler.`
+// It starts the `run` loop to consume log entries that can be
+// stopped via `target.Stop()`
+// nolint:revive,govet
+func NewPullTarget(
+	metrics *Metrics,
+	logger log.Logger,
+	handler loki.EntryHandler,
+	relabel []*relabel.Config,
+	jobName string,
+	config *PullConfig,
+	clientOptions ...option.ClientOption,
+) (*pullTarget, error) {
 	ctx, cancel := context.WithCancel(context.Background())
 	ps, err := pubsub.NewClient(ctx, config.ProjectID, clientOptions...)
 	if err != nil {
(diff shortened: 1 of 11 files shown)


component/loki/source/gcplog/internal/gcplogtarget/types.go
```diff
@@ -14,19 +15,23 @@ type Target interface {
...
 }
 
 // PushConfig configures a GCPLog target with the 'push' strategy.
+// struct is a derived of 'GcplogTargetConfig', but specific for Push configuration from Promtail code (https://github.com/grafana/loki/blob/main/clients/pkg/promtail/scrapeconfig/scrapeconfig.go)
 type PushConfig struct {
-	Server               *fnet.ServerConfig `river:",squash"`
-	PushTimeout          time.Duration      `river:"push_timeout,attr,optional"`
-	Labels               map[string]string  `river:"labels,attr,optional"`
-	UseIncomingTimestamp bool               `river:"use_incoming_timestamp,attr,optional"`
+	Server               server.Config  `river:",squash"`
```
tpaschalis, line 30: The river implementation needs to use the previous definition server. Let's revert this import change to _not_ use the weaveworks/common/server package instead.

component/loki/source/gcplog/internal/gcplogtarget/push_translation.go
```diff
@@ -4,22 +4,24 @@ package gcplogtarget
...
 	"github.com/grafana/agent/component/common/loki"
-	"github.com/grafana/loki/pkg/util"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/model/labels"
 	"github.com/prometheus/prometheus/model/relabel"
-)
 
-// ReservedLabelTenantID reserved to override the tenant ID while processing
-// pipeline stages
-const ReservedLabelTenantID = "__tenant_id__"
```
tpaschalis, line 22: Let's add this back in instead of relying on importing the `lokiClient` package just for a string definition.

component/loki/source/gcplog/internal/gcplogtarget/push_target_test.go
```diff
@@ -1,25 +1,34 @@
 package gcplogtarget
 
+// This code is copied from Promtail. The gcplogtarget package is used to
```
tpaschalis, line 3: I think we should revert changes to this file. We're not adding any new test cases here, and simply reverting to use Promtail structs, which isn't useful.

component/loki/source/gcplog/internal/gcplogtarget/pull_target.go
```diff
@@ -137,25 +151,31 @@ func (t *PullTarget) consumeSubscription() {
...
+func (t *pullTarget) DiscoveredLabels() model.LabelSet {
+	return nil
+}
+
+func (t *pullTarget) Labels() model.LabelSet {
+	return t.config.Labels
+}
+
+func (t *pullTarget) Details() map[string]string {
+	return nil
```
tpaschalis, line 175: Same here, we've removed these Promtail methods since we don't need to implement the same interface and have no use for them at the moment.

component/loki/source/gcplog/internal/gcplogtarget/push_target.go
```diff
@@ -4,6 +4,12 @@ package gcplogtarget
 // configure and run the targets that can read log entries from cloud resource
 // logs like bucket logs, load balancer logs, and Kubernetes cluster logs
 // from GCP.
+// /!\ But
```
tpaschalis, line 7: We also need to take another shot at this; we're not running the server in the way that Promtail does, and we must not revert to it

tpaschalis (changes requested):
Hey @eraac, thank you for the PR and your efforts to improve Grafana Agent! (and also apologies for the belated review).

I'd like to recommend we go through this once more, with the following in mind.

a) We're trying to minimize the code we're directly importing from Promtail. As such, we shouldn't be simply copying the new Promtail files over and reuse the older structs, but port just the _code changes_ that we want instead.
b) We should keep the same way that the current component implementation tests itself, kickstarts its server, exports its structs etc. I've added some comments to better explain what I mean.

So, my suggestion would be to start over, and take it step by step; first adding the new `UseFullLine` field to be exposed in the component's River config with its default value. Then we can work from there to pass it down to `parseGCPLogsEntry`, change the way the formatter works, add the test alongside the existing ones, and finally document the new field and labels on the component's doc pages!

Is that any clearer? Let me know in what way I can help and I will make time to do so!

eraac:
@tpaschalis Thanks for the review. Indeed, in this way the implementation is a lot simpler. 

I misunderstood but from the discussion https://github.com/grafana/agent/discussions/3842 and the comments in the files, I didn't expected to re-implement the feature in the Agent as well (but more like a copy/paste). Maybe a documentation may be good for the next contributors

docs/sources/flow/reference/components/loki.source.gcplog.md
```diff
@@ -91,12 +92,13 @@ push requests from GCP's Pub/Sub servers.
...
+| Name                        | Type          | Description                                                                        | Default | Required |
+|-----------------------------|---------------|------------------------------------------------------------------------------------|---------|----------|
+| `graceful_shutdown_timeout` | `duration`    | Timeout for servers graceful shutdown. If configured, should be greater than zero. | "30s"   | no       |
+| `push_timeout`              | `duration`    | Sets a maximum processing time for each incoming GCP log entry.                    | `"0s"`  | no       |
+| `labels`                    | `map(string)` | Additional labels to associate with incoming entries.                              | `"{}"`  | no       |
+| `use_incoming_timestamp`    | `bool`        | Whether to use the incoming entry timestamp.                                       | `false` | no       |
+| `use_full_line`             | `bool`        | Send the full line from Cloud Logging even if `textPayload` is available.          | `false` | no       |
```
tpaschalis, line 102: Could we also add a sentence about

```
By default, if `textPayload` is present in the line, then it's used as log line
```

component/loki/source/gcplog/internal/gcplogtarget/push_target.go
```diff
@@ -79,39 +79,39 @@ func (p *PushTarget) push(w http.ResponseWriter, r *http.Request) {
 	bs, err := io.ReadAll(r.Body)
 	if err != nil {
 		p.metrics.gcpPushErrors.WithLabelValues("read_error").Inc()
-		level.Warn(p.logger).Log("msg", "failed to read incoming gcp push request", "err", err.Error())
+		_ = level.Warn(p.logger).Log("msg", "failed to read incoming gcp push request", "err", err.Error())
```
tpaschalis, line 82: Nitpicky comment, but could we revert all these cosmetic changes from commit https://github.com/grafana/agent/pull/3872/commits/50d877ff7c2056435ddfbf547088ae991dee2d3c?

These are more IDE-specific options, and we don't enforce them in our codebase right now; to me it just makes the code a little harder to read.

tpaschalis (commented):
LGTM, just a nitpick remaining about not introducing extra stylistic changes!

Once we've fixed this, it's ready to ship!

component/loki/source/gcplog/internal/gcplogtarget/push_target.go
eraac: No problem, it's exactly why I've push this in separate commit. I've revert this change

tpaschalis (approved):
LGTM, thanks for the contribution Kévin!

PR #3873, opened 2023-05-12
Flow: provide hostname as default instance label (prometheus exporters)

jcreixell:
#### PR Description

  - The instance label in static mode defaults to `<host>:<port>`, and can be overriden by specific exporters. It is commonly used in Grafana Cloud to differentiate metrics coming from different agent instances.
  - In Flow exporters, it was statically set to the component ID. This is less useful, especially for integrations that do not include other labels during discovery (like node_exporter).
  - This change allows defining custom instance label values for each integration, and provides a default one for the hostname of the node. The port is not included as it is not available in flow components and would require extensive refactoring.
  - Migrates existing integrations for better compatibility with static-mode (`instance` label should reflect the previous value).

  - Partially fixes https://github.com/grafana/agent/issues/3180

#### Notes to the Reviewer

- Failure to customize the `instance` label will fallback to the default value (host:port) instead of returning an error or panicking (as static mode does). Currently, there is no logging when that happens, might be worth adding.
- Extends existing function to expand a single target into multiple ones, which provides full flexibility for customizing targets (including label overrides or extension).

#### PR Checklist

- [x] CHANGELOG updated
- [ ] Documentation added
- [x] Tests updated


component/prometheus/exporter/exporter.go
```diff
@@ -134,6 +107,46 @@ func (c *Component) Update(args component.Arguments) error {
...
 }
 
+// Handler serves metrics endpoint from the integration implementation.
+func (c *Component) Handler() http.Handler {
+	c.mut.Lock()
+	defer c.mut.Unlock()
+	return c.metricsHandler
+}
+
+func newExporter(creator Creator, name string, instance string, multiTargetFunc func(discovery.Target, component.Arguments) []discovery.Target) func(component.Options, component.Arguments) (component.Component, error) {
```
jcreixell, line 117: this injection of instance is not going to work, as we have not evaluated the arguments yet :thinking:

jcreixell: I will try tweakaing the `multiTargetFunc` to turn it into a more generic custom target generator that also allows overriding the instance label, as discussed with @captncraig and @spartan0x117

component/prometheus/exporter/exporter.go
```diff
@@ -134,6 +108,44 @@ func (c *Component) Update(args component.Arguments) error {
...
+
+		c.baseTarget = discovery.Target{
+			model.AddressLabel:                  opts.HTTPListenAddr,
+			model.SchemeLabel:                   "http",
+			model.MetricsPathLabel:              path.Join(opts.HTTPPath, "metrics"),
+			"instance":                          instance,
+			"job":                               jobName,
+			"__meta_agent_integration_name":     jobName,
+			"__meta_agent_integration_instance": instance,
+			"__meta_agent_integration_id":       opts.ID,
```
jcreixell, line 137: note: this is a new label

captncraig: would `__meta_agent_component_id` be better?

component/prometheus/exporter/mysql/mysql.go
```diff
@@ -23,6 +27,26 @@ func createExporter(opts component.Options, args component.Arguments) (integrati
 	return a.Convert().NewIntegration(opts.Logger)
 }
 
+func customizeTarget(baseTarget discovery.Target, args component.Arguments) []discovery.Target {
+	a := args.(Arguments)
+	target := baseTarget
+
+	m, err := mysql.ParseDSN(string(a.DataSourceName))
```
captncraig, line 34: It feels a bit odd to do this level of validation here, that the component will also do on update. I don't have a super good alternative though, if the arguments are all we are given. Will the unmarshalRiver validation handle this as well?

captncraig (approved):

component/prometheus/exporter/mysql/mysql.go
jcreixell: yeah it is weird and kind of ugly, currently `ParseDSN` is only called in `go-sql-driver`, I think it is a good idea to also call it early as a validation

component/prometheus/exporter/exporter.go
jcreixell: I will add `__meta_agent_hostname` (we used to have `agent_hostname` in static mode)

PR #3883, opened 2023-05-14
flow: Add toggle to enable /debug/pprof endpoint

jkroepke:
#### PR Description

This PR introduce a CLI flag to conditionally enable the go profiling endpoints. Leaving the default enable to preserve backwards compatibility.

#### PR Checklist

- [x] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,7 +92,7 @@ Main (unreleased)
 - Add `agent_wal_out_of_order_samples_total` metric to track samples received
   out of order. (@rfratto)
 
-- Add CLI flag `--enable-go-profiling` to grafana-agent-flow to conditionally enable `/debug/pprof` endpoints (@jkroepke)
+- Add CLI flag `--server.http.enable-pprof` to grafana-agent-flow to conditionally enable `/debug/pprof` endpoints (@jkroepke)
 
 - Use Go 1.20.4 for builds. (@tpaschalis)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -92,6 +92,8 @@ Main (unreleased)
 - Add `agent_wal_out_of_order_samples_total` metric to track samples received
   out of order. (@rfratto)
 
+- Add CLI flag `--enable-go-profiling` to grafana-agent-flow to conditionally enable `/debug/pprof` endpoints (@jkroepke)
+
 - Use Go 1.20.4 for builds. (@tpaschalis)
 
 v0.33.2 (2023-05-11)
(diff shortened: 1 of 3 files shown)


rfratto:
I'm not against this, but I want to make sure we expose the capability in the right way.  I'm trying to understand if a flag is the best way to expose it, or if it should it be configurable at runtime? 

What's your use case for this. Is it security-related?

I also want to note that disabling the pprof endpoints will prevent phlare.scrape from being able to collect profiles on the running agent, and will also prevent the support bundle endpoint (once it's added to Flow) from working too.

jkroepke:
> What's your use case for this? Is it security-related?

Yes, I aware we are on 127.0.0.1 by default, but the profiling endpoint may not enable by default

> I also want to note that disabling the pprof endpoints will prevent phlare.scrape from being able to collect profiles on the running agent, and will also prevent the support bundle endpoint (once it's added to Flow) from working too.

Is it not possible that some endpoints are only available on the in-memory listener?

cmd/internal/flowmode/cmd_run.go
```diff
@@ -95,18 +96,21 @@ depending on the nature of the reload error.
 		StringVar(&r.clusterJoinAddr, "cluster.join-addresses", r.clusterJoinAddr, "Comma-separated list of addresses to join the cluster at")
 	cmd.Flags().
 		BoolVar(&r.disableReporting, "disable-reporting", r.disableReporting, "Disable reporting of enabled components to Grafana.")
+	cmd.Flags().
+		BoolVar(&r.enableGoProfiling, "enable-go-profiling", r.enableGoProfiling, "Enable /debug/pprof profiling endpoints.")
```
rfratto, line 100: I'd like to group this under `server.http.`; can we call this `server.http.enable-pprof`?

CHANGELOG.md
```diff
@@ -68,6 +68,8 @@ Main (unreleased)
 - Add `agent_wal_out_of_order_samples_total` metric to track samples received
   out of order. (@rfratto)
 
+- Add CLI flag `--enable-go-profiling` to grafana-agent-flow to conditionally enable `/debug/pprof` endpoints (@jkroepke)
```
rfratto, line 71: Can you document the new flag at https://github.com/grafana/agent/blob/main/docs/sources/flow/reference/cli/run.md?

rfratto (commented):
LGTM, I think let's move forward with this approach for now and we can refine it over time. 

My only comment is about the flag name; once that's resolved this will be good to merge.

CHANGELOG.md
jkroepke: Done

cmd/internal/flowmode/cmd_run.go
jkroepke: Done

rfratto (approved):
Thank you!

Issue #3891, opened 2023-05-16
The grafana-agent binary supplied in the grafana-agent-0.33.2-1.s390x.deb seems to have been compiled for a PPC64 processor architecture instead of IBM S390x.  


fionnw29:
### What's wrong?

The grafana-agent binary supplied in the grafana-agent-0.33.2-1.s390x.deb seems to have been compiled for a PPC64 processor architecture instead of IBM S/390x processor architecture.

After installing the grafana-agent-0.33.2-1.s390x.deb package on my S/390x test system, the grafana-agent failed to start.

Running the "file" command on the binary, installed in "/usr/bin/grafana-agent" shows the following:

`file /usr/bin/grafana-agent
/usr/bin/grafana-agent: ELF 64-bit LSB executable, 64-bit PowerPC or cisco 7500, version 1 (SYSV), dynamically linked, interpreter /lib64/ld64.so.2, BuildID[sha1]=df14874e9df93797e4971564e68699dc775229ed, for GNU/Linux 3.10.0, stripped`

Running the "file" command on the binary "/usr/bin/grafana-agentctl"  (installed in the same package as the grafana-agent) shows the following:

`file /usr/bin/grafana-agentctl 
/usr/bin/grafana-agentctl: ELF 64-bit MSB executable, IBM S/390, version 1 (SYSV), dynamically linked, interpreter /lib/ld64.so.1, BuildID[sha1]=8cff33c9c8607cdccbff5d1638dd6911062b55f9, for GNU/Linux 3.2.0, stripped`

Output of the "dpkg -L grafana-agent" command:

/.
/etc
/etc/default
/etc/default/grafana-agent
/etc/grafana-agent.yaml
/usr
/usr/lib
/usr/lib/systemd
/usr/lib/systemd/system
/usr/lib/systemd/system/grafana-agent.service
/usr/bin
/usr/bin/grafana-agent
/usr/bin/grafana-agentctl
/usr/share
/usr/share/doc
/usr/share/doc/grafana-agent
/usr/share/doc/grafana-agent/changelog.gz

### Steps to reproduce

Installed the downloaded "grafana-agent-0.33.2-1.s390x.deb" package with this command: 

`dpkg -i grafana-agent-0.33.2-s390x.deb`

### System information

Ubuntu Server 20.04 LTS  S/390x

### Software version

Grafana Agent Version 0.33.2

### Configuration

```text
# Sample config for Grafana Agent
[...]

PR #3892, opened 2023-05-16
tools: fix cases where s390x artifacts were built using the wrong GOARCH

tpaschalis:
#### PR Description
This PR fixes some occurrences where the s390x artifacts were built using the PowerPC GOARCH value, rendering them unusable.

Related to #3891

#### Notes to the Reviewer
I'm not sure whether this should be backported to the previous release. I don't think we can rebuild the already-published artifacts, but we can advise people on how to build their own.

WDYT?

#### PR Checklist

- [x] CHANGELOG updated
- [ ] Tests updated (N/A)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,9 @@ Main (unreleased)
 - Fix an issue where not specifying either `http` nor `grpc` blocks could result
   in a panic for `loki.source.heroku` and `loki.source.gcplog` components. (@thampiotr)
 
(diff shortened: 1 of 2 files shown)


mattdurham (approved):
LGTM

tpaschalis:
Apologies for the noise, just wanted some more input from you two that have been more involved with our release tooling.

Okay, we will _not_ be backporting these, but I will advise the issue author on how to manually build the package.

Issue #3871, opened 2023-05-12
Grafana Agent Flow install fails: start error due to missing config file


cyrille-leclerc:
### What's wrong?

Following the installation doc [Install Grafana Agent Flow on macOS](https://grafana.com/docs/agent/v0.33/flow/install/macos/) fails with

* Success message on the start command
```
$ brew services start grafana-agent-flow
==> Successfully started `grafana-agent-flow` (label: homebrew.mxcl.grafana-agent-flow)
```

* but then verification shows that startup failed
```
$ brew services list
Name               Status     User           File
grafana            none
grafana-agent      none
grafana-agent-flow error  256 cyrilleleclerc ~/Library/LaunchAgents/homebrew.mxcl.grafana-agent-flow.plist
...

$ brew services info grafana-agent-flow
grafana-agent-flow (homebrew.mxcl.grafana-agent-flow)
Running: ✘
Loaded: ✔
Schedulable: ✘
```

* Troubleshooting shows that the `config.river` was not created which is probably the cause of the start failure

```
cat /opt/homebrew/etc/grafana-agent-flow/config.river
cat: /opt/homebrew/etc/grafana-agent-flow/config.river: No such file or directory
```

The documentation [Install Grafana Agent Flow on macOS](https://grafana.com/docs/agent/v0.33/flow/install/macos/) don't say the config file has to be created and filled.
The shell output of `brew install grafana-agent-flow` is not very clear as it refers to customization rather than creation of the config file. I thought the agent would start in a NOOP mode
```
==> Caveats
The agent uses a configuration file that you must customize before running:
  /opt/homebrew/etc/grafana-agent-flow/config.river
```

### Steps to reproduce

Follow https://grafana.com/docs/agent/v0.33/flow/install/macos/

### System information

MacBook Air M2 - MacOS 13.3.1 (a) (22E772610a)

### Software version

Grafana Agent Flow agent, version v0.33.2 (branch: HEAD, revision: unknown)   build user:       grafana   build date:       2023-05-11T20:26:42Z   go version:       go1.20.4   platform:         darwin/arm64   tags:             builtinassets,noebpf

### Logs

```text
$ brew tap grafana/grafana

==> Tapping grafana/grafana
Cloning into '/opt/homebrew/Library/Taps/grafana/homebrew-grafana'...
remote: Enumerating objects: 269, done.
remote: Counting objects: 100% (49/49), done.
remote: Compressing objects: 100% (26/26), done.
[...]

PR #3895, opened 2023-05-16
docs: fix macOS log locations for Flow

rfratto:
Previously, an instance of the grafana/grafana-agent-flow Homebrew formula installed on macOS would generate logs in the same location as the static mode grafana-agent formula.

The documentation here was incorrect, saying that the stderr file would write to a Flow-specific file.

grafana/homebrew-grafana#52 fixes the formula to log files in the proper, Flow-specific, place, and this PR updates the documentation to use the correct file names (assuming grafana/homebrew-grafana#52 is merged).

Related to #3871.

### DIFF
--- a/docs/sources/flow/install/macos.md
+++ b/docs/sources/flow/install/macos.md
@@ -130,7 +130,7 @@ To expose the UI to other machines, complete the following steps:
 
 ### Viewing Grafana Agent Flow logs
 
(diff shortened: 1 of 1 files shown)


rfratto:
cc @cyrille-leclerc

chri2547 (approved):

cyrille-leclerc:
This changes matches with what I have on my test environment.

PR #3898, opened 2023-05-16
Config API v1

### LINKED ISSUE #3736: Create Convert API and the internal pattern for the API
Introduce the Convert API for converting config from other formats into River config.

In addition, `prometheus scrape config -> river` or comparable should be implemented inside the Convert API to demonstrate how it will work and be tested.

erikbaranowski:
#### PR Description

Introduce Config Convert API with an implementation for the prometheus.scrape and prometheus.remote_write Flow components.

Known features for config conversion that this PR does not include
- A way for components to include capsules from other components #3902
- Print secrets as strings #3901
- Automatic handling of default parameters for more concise output #3900

Closes #3736 

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated


erikbaranowski:
I'm open to feedback regarding naming of the 'reconvert' functions. This name was chosen to not intersect with the 'convert' functions that translate from outside config to Flow. In the case we decide to move this functionality to those packages it is intended to differentiate the direction of the conversion.

rfratto (approved):
Overall LGTM, but I'd like for us to add a second Test function which applies converted files from `testdata/` against a Flow controller to make sure that things load as we expect. We don't have to actually run the controller; just call LoadConfig so the graph evaluates.

Also: I'm not sure about creating all the packages for the different bits of Prometheus we're trying to convert; I'm worried it will be more tedious to discover the various functions, but I'm not too opinionated about it since an internal API. Having all the Prometheus-related reconvert functions in a single flat package might be simpler and avoids any chance of circular dependencies between convert functions.

converter/internal/prometheusconvert/scrape/scrape.go
```diff
@@ -0,0 +1,59 @@
...
+		HTTPClientConfig:      *common.ReconvertHttpClientConfig(&scrapeConfig.HTTPClientConfig),
+		ExtraMetrics:          false,
+		Clustering:            scrape.Clustering{Enabled: false},
+	}
+}
+
+func getTargets(scrapeConfig *promconfig.ScrapeConfig) []discovery.Target {
+	targets := []discovery.Target{}
+
+	for _, serviceDiscoveryConfig := range scrapeConfig.ServiceDiscoveryConfigs {
```
rfratto, line 45: Can you go through this entire PR and add TODO comments for anything from Prometheus which isn't being converted yet? 

Since there's a lot of surface area, I'm worried we'll forget something.

converter/internal/prometheusconvert/testdata/prometheus_agent.out.river
```diff
@@ -0,0 +1,85 @@
...
+	forward_to      = []
+	job_name        = "prometheus1"
+	scrape_interval = "10s"
+	scrape_timeout  = "5s"
+	metrics_path    = "/metrics"
+	scheme          = "http"
+
+	basic_auth {
+		username = "user"
+		password = (secret)
```
rfratto, line 66: This is broken; this will remove any inline credentials from a user. It's also invalid River and will fail to evaluate because `secret` isn't a valid identifier.

rfratto (commented):
Sorry, I noticed a bigger issue in the testdata.

mattdurham:
Bikeshedding, thoughts on `ToRiver` instead of `Reconvert`?

mattdurham:
Wondering about the namespacing with `prometheus`, since we will likely add actual Prometheus  config conversations at some point but that could be a refactor for another day

mattdurham:
Overall LGTM with the issue with secrets and receivers we talked about earlier. IMO I like having the smaller scoped like this PR functions instead of larger ones.

mattdurham (approved):
LGTM aside from, receivers and secrets

converter/internal/prometheusconvert/prometheusconvert_test.go
```diff
@@ -0,0 +1,114 @@
...
+					evalBlock(t, node)
+				}
+			})
+		}
+
+		return nil
+	})
+}
+
+func evalBlock(t *testing.T, block *ast.BlockStmt) {
```
erikbaranowski, line 98: @rfratto  @mattdurham  This testing was updated to eval river blocks into their specific argument types. It's a little more maintenance which is why I wanted to avoid it, but evaluating the whole graph calls a whole ton of code in component New/Update functions. It was specifically crashing the test in windows but in general it's a lot of code I don't necessarily want to test here.

If there is some alternate way to do this without specifying the argument types based on the block name I am all ears.

erikbaranowski: removed the test for now per a side conversation

PR #3912, opened 2023-05-17
change scrape intervals in docs to 60s

### LINKED ISSUE #3411: Docs: Align scrape_interval in docs to 60s
When setting the scrape_interval to e.g. 15s and you are using Grafana Cloud please make sure the scrape interval is also reflected in your corresponding datasource settings. Otherwise the agent will scrape at an 15s interval but Grafana will still show just one data point per minute.

mattdurham:
#### PR Description

Change the places in docs where we still specify 15s instead of 60s.

Closes #3411

### DIFF
--- a/docs/sources/static/configuration/agent-management.md
+++ b/docs/sources/static/configuration/agent-management.md
@@ -120,7 +120,7 @@ snippets:
     config: |
       metrics_scrape_configs:
       - job_name: 'prometheus'
-        scrape_interval: 15s
+        scrape_interval: 60s
         static_configs:
         - targets: ['localhost:9090']
       logs_scrape_configs:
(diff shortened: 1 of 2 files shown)


spartan0x117 (approved):

clayton-cornell:
@rfratto Did you want this topic backported to v0.33? The topic is currently only in `Next`, and doesn't exist in v0.33.

rfratto:
@clayton-cornell The node_exporter page exists in v0.33, doesn't it? https://grafana.com/docs/agent/latest/static/configuration/integrations/node-exporter-config/

clayton-cornell:
Ah sorry I didn't note which page... the `node-exporter-config` page is in v0.33,  but the other page that's included in this PR, `static/configuration/agent-management.md` is only in next

Issue #3686, opened 2023-04-27
otelcol.auth.headers only works if TLS is enabled


ptodev:
A customer [reported](https://grafana.slack.com/archives/C01050C3D8F/p1682590848693829) reported this error:

`
ts=2023-04-27T07:00:53.129758843Z level=error component=otelcol.exporter.otlp.agentTempo msg="failed to start scheduled component" err="grpc: the credentials require transport level security (use grpc.WithTransportCredentials() to set)"
`

This is the config used:
<details>
<summary>River config</summary>

```
otelcol.receiver.otlp "otelInput" {
  http {}
  grpc {}
​
  output {
    traces  = [otelcol.processor.batch.otelBatch.input]
  }
}
​
otelcol.processor.batch "otelBatch" {
  output {
    traces  = [otelcol.exporter.otlp.agentTempo.input]
  }
}
​
otelcol.exporter.otlp "agentTempo" {
  client {
    endpoint = "tempo-distributor.d10devops-grafana-tempo:4317"
    auth     = otelcol.auth.headers.creds.handler
    tls {
      insecure_skip_verify = true
      insecure = true
    }
  }
}
​
otelcol.auth.headers "creds" {
  header {
    key          = "x-scope-orgid"
    value        = "test"
  }
}
```
</details>

The official [headerssetterextension](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/headerssetterextension/README.md) does not say that TLS is a requirement, and neither does the [otelcol.auth.headers](https://grafana.com/docs/agent/latest/flow/reference/components/otelcol.auth.headers/) documentation. If the need for TLS is intentional, it should be mentioned in the docs. If it is not intentional, the code should be fixed to not need TLS.

ptodev:
On a sidenote, the example in [otelcol.auth.headers](https://grafana.com/docs/agent/latest/flow/reference/components/otelcol.auth.headers/#example) uses it to set headers for otelcol.exporter.otlp. However, users could do this by setting the `headers` in the grpc [client](https://grafana.com/docs/agent/latest/flow/reference/components/otelcol.exporter.otlp/#client-block) block instead. Maybe we should write another example with a component where such a `headers` config attribute doesn't exist. It'd probably be more useful.

Also, the docs are not clear what would happen if we try to set the exact same headers with different values from both `otelcol.auth.headers` and the grpc client block. We should probably say which would take precedence.

ptodev:
[...]

PR #3916, opened 2023-05-17
Document credentials over non-TLS connections

jcreixell:
#### PR Description

  - Several users have fallen into this while attempting to configure otlp in internal networking environments, so we add a clarifying note to the documentation.

  - Fixes https://github.com/grafana/agent/issues/3686


docs/sources/flow/reference/components/otelcol.exporter.otlp.md
```diff
@@ -107,6 +107,11 @@ server.
 
 {{< docs/shared lookup="flow/reference/components/otelcol-tls-config-block.md" source="agent" >}}
 
+> **NOTE**: `otelcol.exporter.otlp` uses gRPC, which disallows sending sensitive credentials (like `auth`) over insecure channels for security reasons.
+> If you need to send sensitive credentials over non-TLS connections (via the `insecure` argument), you may want to use [otelcol.exporter.otlphttp][] instead.
```
clayton-cornell, line 111:
```suggestion
> If you need to send sensitive credentials over non-TLS connections (via the `insecure` argument), you can use [otelcol.exporter.otlphttp][] instead.
```

docs/sources/flow/reference/components/otelcol.exporter.otlp.md
```diff
@@ -107,6 +107,11 @@ server.
 
 {{< docs/shared lookup="flow/reference/components/otelcol-tls-config-block.md" source="agent" >}}
 
+> **NOTE**: `otelcol.exporter.otlp` uses gRPC, which disallows sending sensitive credentials (like `auth`) over insecure channels for security reasons.
```
clayton-cornell, line 110:
```suggestion
> **NOTE**: `otelcol.exporter.otlp` uses gRPC, which does not allow you to send sensitive credentials (like `auth`) over insecure channels.
```

docs/sources/flow/reference/components/otelcol.exporter.otlp.md
ptodev:
```suggestion
> Sending sensitive credentials over insecure non-TLS connections is supported by non-gRPC exporters such as [otelcol.exporter.otlphttp][.
```
What about this wording?

clayton-cornell: From a language standpoint, that's good. It's just a clear statement without giving specific direction (which seems to be more appropriate here).

jcreixell: I am fine with this, @rfratto WDYT?

rfratto: SGTM 👍

rfratto (approved):

rfratto:
Should this get the backport-v0.33 label to be backported to live docs?

PR #3918, opened 2023-05-17
cluster: Fix targets deduplication

laurovenancio:
#### PR Description
Prometheus deduplicates targets after removing the `__meta_` labels.
If these labels are used to define the targets distribution, targets
that must be deduplicated could end in different nodes, resulting in
duplicated metrics.

#### PR Checklist

- [X] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -87,6 +87,8 @@ Main (unreleased)
 - Fix an issue where the cluster advertise address was overwriting the join
   addresses. (@laurovenancio)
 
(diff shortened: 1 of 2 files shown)


rfratto:
Nice catch! This makes sense to me but I want @tpaschalis to give it a look to make sure I'm not missing something.

rfratto (approved):
I chatted with Paschalis offline, looks good, thanks!

PR #3924, opened 2023-05-18
clusterer: use a 512-token ring

rfratto:
Use a 512-token ring to achieve better distribution. Because a 512-token ring only needs 12MB for 1,000 nodes (compared to 6MB for 1,000 nodes in a 256-token ring), its overhead is marginal enough to be worth it for the improved distribution.

However, this change means that any existing cluster being upgraded to the new version will suffer from incorrect or missing hashing assignments until all nodes are using 512 tokens.

This is the type of change we shouldn't make often; I think it's only feasible now because clustering is still experimental. Future changes would likely need to be opt-in via a flag.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,7 +62,7 @@ Main (unreleased)
 - Embed Google Fonts on Flow UI (@jkroepke)
 
 - Enable Content-Security-Policies on Flow UI (@jkroepke)
-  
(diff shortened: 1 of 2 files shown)


mattdurham (approved):
512 > 256 =)

tpaschalis (approved):
⛵

PR #3926, opened 2023-05-18
helm: make faro port optional

### LINKED ISSUE #2957: (helm) grafana-agent `faroPort` should be optional
Hi

https://github.com/grafana/agent/pull/2922 introduces ingress support for faro.

This exposes the agent `faroPort` on the generated service and pod resources which it probably shouldn't do if faro is not enabled (via config in `app_agent_receiver_configs`).

---

A sidenote here is that multiple faro-ports could actually be exposed since `app_agent_receiver_configs` is a list. That could lead to multiple ingresses or a single ingress with multiple backends.

It's not really clear what the recommend setup should be here, i.e: multiple grafana-agents with a single faro-endpoint defined, or 1 grafana-agent with multiple endpoints defined.

tpaschalis:
#### PR Description
This PR makes the faro port on the Agent pods optional; people who still wish to use `:12347` for Faro can opt-in to it by defining it on `extraPorts`, like in the provided example.

Notice that the `faro-ingress` test is virtually the same.

Fixes #2957

#### Notes to the Reviewer

I just saw the pre-existing issue #2957 which outlines some possible solutions. For now, I've moved `faroPort` inside of the ingress which sounds like it's custom-made for it.

Would we prefer to allow a list of ports that we loop over to create a `rule` for each on the same ingress so we don't have to explicitly refer to Faro (and leave it as an example so people can enable it with ease)? This is not an ideal long-term solution but at least a step towards it.

#### PR Checklist

- [x] CHANGELOG updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/operations/helm/charts/grafana-agent/CHANGELOG.md
+++ b/operations/helm/charts/grafana-agent/CHANGELOG.md
@@ -14,6 +14,8 @@ Unreleased
 
 - Add HPA support for Deployments and StatefulSets. (@tpaschalis)
 
+- Make the Faro port optional. (@tpaschalis)
+
 0.14.0 (2023-05-11)
 -------------------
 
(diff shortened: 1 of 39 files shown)


rfratto (approved):
LGTM, thanks!

PR #3927, opened 2023-05-18
Add AWS Firehose loki.source component

thepalbi:
#### PR Description

This PR adds a new agent flow component, making the agent capable to receive [AWS Firehose batches through HTTP](https://docs.aws.amazon.com/firehose/latest/dev/httpdeliveryrequestresponse.html). 

The new component works much like gcplogs push or heroku, exposing a network server that exposes a single endpoints under: `POST /awsfirehose/api/v1/push`

This endpoints implements the firehose HTTP api quoted above. Apart from that, it's able to handle the special case of [CloudWatch Logs](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/SubscriptionFilters.html#FirehoseExample) events being shipped to firehose, further decoding them and:
- Splitting a single firehose record into the multiple cloudwatch log messages contained
- Adding all meta-information about cloudwatch as optional labels, available for relabel

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Implement pending TODO on using the incoming timestamp
- [x] Tests updated


component/loki/source/aws_firehose/internal/handler.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	// Using the same header check as the gzip library, but inlining the check to avoid unnecessary boilerplate
+	// code from creating the reader.
+	//
+	// https://github.com/golang/go/blob/master/src/compress/gzip/gunzip.go#L185
+	if !(decodedRec[0] == gzipID1 && decodedRec[1] == gzipID2 && // the first two represent the 1f8b magic bytes
+		decodedRec[2] == gzipDeflate) { // the third byte represents the gzip compression method DEFLATE
+		// no gzip, return decoded data
+		return decodedRec, OriginDirectPut, nil
```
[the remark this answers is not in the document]
kgeckhart: Thanks CloudWatch, WDYT about linking out to the CloudWatch docs to help describe why we need to do this?

component/loki/source/aws_firehose/internal/handler.go
```diff
@@ -0,0 +1,199 @@
...
+	if err != nil {
+		return nil, OriginCloudwatchLogs, fmt.Errorf("error creating gzip reader: %w", err)
+	}
+	defer gzipReader.Close()
+
+	b := bytes.Buffer{}
+	if _, err := io.Copy(bufio.NewWriter(&b), gzipReader); err != nil {
+		return nil, OriginCloudwatchLogs, fmt.Errorf("error reading gzipped bytes: %w", err)
+	}
+	return b.Bytes(), OriginCloudwatchLogs, nil
```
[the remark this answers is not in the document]
kgeckhart: 👍 for merge it and see how it scales

component/loki/source/aws_firehose/internal/handler.go
```diff
@@ -0,0 +1,199 @@
...
+		if err != nil {
+			h.metrics.errors.WithLabelValues("decode").Inc()
+			level.Error(h.logger).Log("msg", "failed to decode request record", "err", err.Error())
+			h.respondError(w, firehoseReq.RequestID, "failed to decode record", http.StatusBadRequest)
+
+			// todo(pablo): is ok this below?
+			// since all individual data record are packed in a bigger record, responding an error
+			// here will mean we'll get the same individual record on the retry. Continue processing
+			// the rest.
+			return
```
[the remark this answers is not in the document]
thepalbi: Yup, did the following:
- Make the handler continue when one single record fails
- Splitted error metrics into request decoding, and record decoding
- Added batch size metrics

component/loki/source/aws_firehose/component.go
```diff
@@ -0,0 +1,171 @@
...
+
+	serverNeedsUpdate := !reflect.DeepEqual(c.args.Server, newArgs.Server)
+	if !serverNeedsUpdate {
+		c.args = newArgs
+		return nil
+	}
+
+	c.shutdownServer()
+
+	jobName := strings.Replace(c.opts.ID, ".", "_", -1)
```
tpaschalis, line 135: Why do we need this, does the jobName end up at a label? (I also saw we use in the gcplog component, I'm just not sure we do need it.)

component/loki/source/aws_firehose/component.go
```diff
@@ -0,0 +1,171 @@
...
+
+	registry := prometheus.NewRegistry()
+	c.serverMetrics.SetCollector(registry)
+
+	wlog := log.With(c.logger, "component", "aws_firehose_logs")
+	c.server, err = fnet.NewTargetServer(wlog, jobName, registry, newArgs.Server)
```
tpaschalis, line 141: I think the logger is already decorated with the component ID. Did we add this to override it with a different value instead?

component/loki/source/aws_firehose/internal/handler.go
```diff
@@ -0,0 +1,255 @@
...
+}
+
+// sendAPIResponse responds to AWS Firehose API in the expected response format. To simplify error handling,
+// it uses a string template instead of marshalling a struct.
+func sendAPIResponse(w http.ResponseWriter, firehoseID, errMsg string, status int) {
+	timestamp := time.Now().Unix()
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(status)
+	if errMsg != "" {
+		_, _ = w.Write([]byte(fmt.Sprintf(errorResponseTemplate, firehoseID, timestamp, errMsg)))
```
tpaschalis, line 179: Does AWS firehose expect that we respond with http.StatusOK even if an error occurred?

component/loki/source/aws_firehose/component.go
```diff
@@ -0,0 +1,171 @@
...
+	if err != nil {
+		return err
+	}
+
+	if err = c.server.MountAndRun(func(router *mux.Router) {
+		// re-create handler when server is re-computed
+		// todo(pablo): should use unchecked collector here?
+		handler := internal.NewHandler(c, c.logger, c.opts.Registerer, c.rbs)
```
tpaschalis, line 149: I _think_ we're okay with the component registerer here? We can check what happens when we reload with a changed configuration as well as when we remove the component, reload, then add it back again and reload once more.

tpaschalis (commented):
I think this is progressing well! I can't say much about the decoding part as I haven't tested it (although I thought that Go's HTTP handler would handle gzip decoding implicitly), but looks alright at a first glance!

component/loki/source/aws_firehose/component.go
thepalbi: In both gcp logs and here the component name with that replace, for example `loki_souce_awsfirehose` is used as prefix for all weaveworks server metrics

component/loki/source/aws_firehose/component.go
```diff
@@ -0,0 +1,173 @@
...
+
+	c.server, err = fnet.NewTargetServer(c.logger, jobName, registry, newArgs.Server)
+	if err != nil {
+		return err
+	}
+
+	if err = c.server.MountAndRun(func(router *mux.Router) {
+		// re-create handler when server is re-computed
+		handler := internal.NewHandler(c, c.logger, registry, c.rbs)
+		router.Path("/awsfirehose/api/v1/push").Methods("POST").Handler(handler)
```
mattdurham, line 152: IMO we should make this pather configurable, with this has the default. I dont see where we are prepending the component id. If we dont make it configurable then it needs to be a singleton and that we are trying to avoid singletons.

component/loki/source/aws_firehose/component.go
```diff
@@ -0,0 +1,173 @@
...
+	c.shutdownServer()
+
+	// update relabel rules in component if needed
+	if relabelRulesChanged {
+		c.rbs = newRelabels
+	}
+
+	jobName := strings.Replace(c.opts.ID, ".", "_", -1)
+
+	registry := prometheus.NewRegistry()
```
mattdurham, line 141: `NewRegistry` feels off, is there a reason we arent using the passed in registry?

component/loki/source/aws_firehose/internal/handler.go
```diff
@@ -0,0 +1,255 @@
...
+
+	h.metrics.batchSize.WithLabelValues().Observe(float64(len(firehoseReq.Records)))
+
+	for _, rec := range firehoseReq.Records {
+		// cleanup err since it might have failed in the previous iteration
+		err = nil
+
+		decodedRecord, recordType, err := h.decodeRecord(rec.Data)
+		if err != nil {
+			h.metrics.errorsRecord.WithLabelValues(getReason(err)).Inc()
```
mattdurham, line 117: Do we know the cardinality of reason? It seems it could be high, but in actual usage maybe its minimal?

mattdurham (changes requested):
Found some things I am unsure of and want to avoid making this a singleton

component/loki/source/aws_firehose/component.go
thepalbi: Regarding the path, just following what we've been using in promtail (hence moving into flow as well). If you check the rest of the `loki.source` components that expose a network server:
- [api](https://grafana.com/docs/agent/next/flow/reference/components/loki.source.api/): `/api/v1/push`
- [heroku](https://grafana.com/docs/agent/latest/flow/reference/components/loki.source.heroku/): `/heroku/api/v1/drain`
- [gcplogs](https://grafana.com/docs/agent/next/flow/reference/components/loki.source.gcplog/#push-block): `/gcp/api/v1/push`

Following that pattern I chose `/awsfirehose/api/v1/push`

What do you mean that it needs to be a singleton?

component/loki/source/aws_firehose/component.go
thepalbi: Since both the component and the inner server register metrics, I'm using the unchecked collector registered when the component is created. If not, when either the server or the handler or the server re-created, the registry will try to re-register those metrics and panic. 
For the server there's no other way, all components that use that server use an unchecked collector.
For the handler metrics I could create the metrics object outside, and keep it in the component state. That way we avoid the re-registering panic

component/loki/source/aws_firehose/internal/handler.go
thepalbi: Yes, it's something I've added for easily propagating some error codes out of some inner functions. It has <10 values. You can check all the `errWithReason` usages, and if the error is not that type `getReason` fallsback to `unknown`

component/loki/source/aws_firehose/component.go
rfratto: @mattdurham AFAICT It's not a singleton because each instance of the component has its own HTTP server

component/loki/source/aws_firehose/component.go
mattdurham: The latter sounds reasonable and ensures the metrics get captured, if there isnt value to the metrics and your just using the registry to satisfy the function interface then thats a different store. If so then a comment explaining the behavior would suffice.

component/loki/source/aws_firehose/component.go
mattdurham: Its a bit painful to setup multiple ones (mainly due to weaveworks always creating a grpc server) but it works. So resolving this.

component/loki/source/aws_firehose/component.go
thampiotr: Related: https://github.com/weaveworks/common/issues/285 - I will work on making sure we can get rid of the unchecked collector eventually.

mattdurham (changes requested):
Only major change requested is having docs.

component/loki/source/aws_firehose/component.go
thepalbi: Implemented the latter to avoid passing the registry.

thepalbi:
@mattdurham just pushed the commits implementing the missing `UseIncomingTimestamp` feature, and all docs.

docs/sources/flow/reference/components/loki.source.awsfirehose.md
```diff
@@ -0,0 +1,189 @@
...
+with [Firehose HTTP Delivery API](https://docs.aws.amazon.com/firehose/latest/dev/httpdeliveryrequestresponse.html).
+Since the API model that AWS Firehose uses to deliver data over HTTP is generic enough, the same component can be used
+to receive data from multiple origins:
+
+- [AWS CloudWatch logs](https://docs.aws.amazon.com/firehose/latest/dev/writing-with-cloudwatch-logs.html)
+- [AWS CloudWatch events](https://docs.aws.amazon.com/firehose/latest/dev/writing-with-cloudwatch-events.html)
+- Custom data through [DirectPUT requests](https://docs.aws.amazon.com/firehose/latest/dev/writing-with-sdk.html)
+
+The component uses a heuristic to try to decode as much information as possible from each log record, falling back to writing
+to Loki the raw line. The decoding process goes as follows:
```
clayton-cornell, line 21:
```suggestion
The component uses a heuristic to try to decode as much information as possible from each log record, falling back to writing
the raw line to Loki . The decoding process goes as follows:
```

Reverse the order of the words

docs/sources/flow/reference/components/loki.source.awsfirehose.md
```diff
@@ -0,0 +1,189 @@
...
+
+- AWS Firehose sends batched requests
+- Each individual record is treated individually
+- For `record` received in each request:
+  - If the `record` comes from a [CloudWatch logs subscription filter](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/SubscriptionFilters.html#DestinationKinesisExample), it will be further decoded and each logging event will be written to Loki individually
+  - If not, the whole record is written raw to Loki
+
+The component exposes some internal labels, available for re-labeling. The following tables describes internal labels available
+in records coming from any source.
```
clayton-cornell, line 30:
```suggestion
The component exposes some internal labels available for re-labeling. The following tables describe the internal labels available
in records coming from any source.
```

docs/sources/flow/reference/components/loki.source.awsfirehose.md
```diff
@@ -0,0 +1,189 @@
...
+| `__aws_cw_log_stream`       | The log stream name of the originating log data.                                                                                                                                                    | `111111111111_CloudTrail/logs_us-east-1`                                 |
+| `__aws_cw_matched_filters`  | The list of subscription filter names that matched with the originating log data. The list is encoded as a comma-separated list.                                                                    | `Destination,Destination2`                                               |
+| `__aws_cw_msg_type`         | Data messages will use the `DATA_MESSAGE` type. Sometimes CloudWatch Logs may emit Kinesis Data Streams records with a `CONTROL_MESSAGE` type, mainly for checking if the destination is reachable. | `DATA_MESSAGE`                                                           |
+
+See [Examples](#example) for a full example configuration on how to enrich each log entry with these labels.
```
clayton-cornell, line 48:
```suggestion
See [Examples](#example) for a full example configuration showing how to enrich each log entry with these labels.
```

docs/sources/flow/reference/components/loki.source.awsfirehose.md
```diff
@@ -0,0 +1,189 @@
...
+loki.source.awsfirehose "LABEL" {
+    http {
+        listen_address = "LISTEN_ADDRESS"
+        listen_port = PORT 
+    }
+    forward_to = RECEIVER_LIST
+}
+```
+
+The component will start HTTP server on the configured port and address with the following endpoints:
```
clayton-cornell, line 62:
```suggestion
The component will start an HTTP server on the configured port and address with the following endpoints:
```

docs/sources/flow/reference/components/loki.source.awsfirehose.md
```diff
@@ -0,0 +1,189 @@
...
+  with [AWS Firehose HTTP Specifications](https://docs.aws.amazon.com/firehose/latest/dev/httpdeliveryrequestresponse.html).
+
+## Arguments
+
+`loki.source.awsfirehose` supports the following arguments:
+
+| Name                     | Type                 | Description                                                | Default | Required |
+ |--------------------------|----------------------|------------------------------------------------------------|---------|----------|
+| `forward_to`             | `list(LogsReceiver)` | List of receivers to send log entries to.                  |         | yes      |
+| `use_incoming_timestamp` | `bool`               | Whether or not to use the timestamp received from request. | `false` | no       |
```
clayton-cornell, line 74:
```suggestion
| `use_incoming_timestamp` | `bool`               | Whether or not to use the timestamp received from the request. | `false` | no       |
```

docs/sources/flow/reference/components/loki.source.awsfirehose.md
```diff
@@ -0,0 +1,189 @@
...
+---
+
+# loki.source.awsfirehose
+
+`loki.source.awsfirehose` receives log entries over HTTP
+from [AWS Firehose](https://docs.aws.amazon.com/firehose/latest/dev/what-is-this-service.html)
+and forwards them to other `loki.*` components.
+
+The HTTP API exposed is compatible
+with [Firehose HTTP Delivery API](https://docs.aws.amazon.com/firehose/latest/dev/httpdeliveryrequestresponse.html).
```
clayton-cornell, line 12:
```suggestion
with the [Firehose HTTP Delivery API](https://docs.aws.amazon.com/firehose/latest/dev/httpdeliveryrequestresponse.html).
```

docs/sources/flow/reference/components/loki.source.awsfirehose.md
```diff
@@ -0,0 +1,189 @@
...
+
+- [AWS CloudWatch logs](https://docs.aws.amazon.com/firehose/latest/dev/writing-with-cloudwatch-logs.html)
+- [AWS CloudWatch events](https://docs.aws.amazon.com/firehose/latest/dev/writing-with-cloudwatch-events.html)
+- Custom data through [DirectPUT requests](https://docs.aws.amazon.com/firehose/latest/dev/writing-with-sdk.html)
+
+The component uses a heuristic to try to decode as much information as possible from each log record, falling back to writing
+to Loki the raw line. The decoding process goes as follows:
+
+- AWS Firehose sends batched requests
+- Each individual record is treated individually
```
clayton-cornell, line 24:
```suggestion
- Each record is treated individually
```

docs/sources/flow/reference/components/loki.source.awsfirehose.md
```diff
@@ -0,0 +1,189 @@
...
+- For `record` received in each request:
+  - If the `record` comes from a [CloudWatch logs subscription filter](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/SubscriptionFilters.html#DestinationKinesisExample), it will be further decoded and each logging event will be written to Loki individually
+  - If not, the whole record is written raw to Loki
```
clayton-cornell, line 27:
```suggestion
- For each `record` received in each request:
  - If the `record` comes from a [CloudWatch logs subscription filter](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/SubscriptionFilters.html#DestinationKinesisExample), it is decoded and each logging event is written to Loki
  - All other records are written raw to Loki
```

docs/sources/flow/reference/components/loki.source.awsfirehose.md
```diff
@@ -0,0 +1,189 @@
...
+
+The component exposes some internal labels, available for re-labeling. The following tables describes internal labels available
```
clayton-cornell, line 29:
```suggestion
The component exposes some internal labels, available for relabeling. The following tables describes internal labels available
```

docs/sources/flow/reference/components/loki.source.awsfirehose.md
```diff
@@ -0,0 +1,189 @@
...
+| `__aws_cw_log_group`        | The log group name of the originating log data.                                                                                                                                                     | `CloudTrail/logs`                                                        |
+| `__aws_cw_log_stream`       | The log stream name of the originating log data.                                                                                                                                                    | `111111111111_CloudTrail/logs_us-east-1`                                 |
+| `__aws_cw_matched_filters`  | The list of subscription filter names that matched with the originating log data. The list is encoded as a comma-separated list.                                                                    | `Destination,Destination2`                                               |
```
clayton-cornell, line 45:
```suggestion
| `__aws_cw_matched_filters`  | The list of subscription filter names that match the originating log data. The list is encoded as a comma-separated list.                                                                    | `Destination,Destination2`                                               |
```

docs/sources/flow/reference/components/loki.source.awsfirehose.md
```diff
@@ -0,0 +1,189 @@
...
+
+`loki.source.awsfirehose` does not export any fields.
+
+## Component health
+
+`loki.source.awsfirehose` is only reported as unhealthy if given an invalid configuration.
+
+## Debug metrics
+
+The following are some of the metrics that are exposed when this component is used. Note that the metrics include labels
```
clayton-cornell, line 114:
```suggestion
The following are some  metrics that are exposed when you use this component. Note that the metrics include labels
```

docs/sources/flow/reference/components/loki.source.awsfirehose.md
```diff
@@ -0,0 +1,189 @@
...
+such as `status_code` where relevant, which can be used to measure request success rates.
```
clayton-cornell, line 115:
```suggestion
The following are some of the metrics that are exposed when this component is used. 
{{% admonition type="note" %}}
The metrics include labels  such as `status_code` where relevant, which you can use to measure request success rates.
{{%/admonition %}}
```

docs/sources/flow/reference/components/loki.source.awsfirehose.md
```diff
@@ -0,0 +1,189 @@
...
+    }
+    forward_to = [
+        loki.write.local.receiver,
+    ]
+}
+```
+
+As another example, if we are receiving records that are originated from a CloudWatch logs subscription, we can enrich each
+received entry by re-labeling internal labels. The following configuration builds upon the one above, but keeps the origin
+log stream and group as `log_stream` and `log_group` respectively.
```
clayton-cornell, line 152:
```suggestion
As another example, if you are receiving records that originated from a CloudWatch logs subscription, you can enrich each
received entry by relabeling internal labels. The following configuration builds upon the one above but keeps the origin
log stream and group as `log_stream` and `log_group`, respectively.
```

clayton-cornell (commented):
Added some suggestions for the docs. Mostly just minor cleanup adding in a few missing words and changing the word order in  some places

thepalbi:
@clayton-cornell addressed all comments. Had to apply them in diff batches due to GH not liking overlapping comments or sth 🤦

PR #3942, opened 2023-05-22
Update loki version to fix promtail goroutine leak

thampiotr:
#### PR Description

Ran `go get github.com/grafana/loki@ac4a442291cceab54c4abd5962f4f57b66121b84` in order to get this fix: https://github.com/grafana/loki/commit/ac4a442291cceab54c4abd5962f4f57b66121b84

Fixes the static mode code path of https://github.com/grafana/agent/issues/3885

Flow mode fix still pending.

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/Shopify/sarama v1.38.1
 	github.com/alecthomas/kingpin/v2 v2.3.2
 	github.com/alecthomas/units v0.0.0-20211218093645-b94a6e3cc137
-	github.com/aws/aws-sdk-go v1.44.187
+	github.com/aws/aws-sdk-go v1.44.217
 	github.com/aws/aws-sdk-go-v2 v1.17.2
 	github.com/aws/aws-sdk-go-v2/config v1.17.10
 	github.com/aws/aws-sdk-go-v2/service/s3 v1.27.11
@@ -31,7 +31,7 @@ require (
 	github.com/docker/docker v23.0.3+incompatible
 	github.com/docker/go-connections v0.4.0
 	github.com/drone/envsubst/v2 v2.0.0-20210730161058-179042472c46
-	github.com/fatih/color v1.14.1
+	github.com/fatih/color v1.15.0
 	github.com/fatih/structs v1.1.0
 	github.com/fsnotify/fsnotify v1.6.0
 	github.com/github/smimesign v0.2.0
@@ -53,16 +53,16 @@ require (
 	github.com/gorilla/mux v1.8.0
 	github.com/grafana/ckit v0.0.0-20230518140533-fbd338b33964
 	github.com/grafana/cloudflare-go v0.0.0-20230110200409-c627cf6792f2
-	github.com/grafana/dskit v0.0.0-20230201083518-528d8a7d52f2
+	github.com/grafana/dskit v0.0.0-20230417151531-1725bcc1e9a1
 	github.com/grafana/go-gelf/v2 v2.0.1
-	github.com/grafana/loki v1.6.2-0.20230414223651-220cbdd4f172 // k146 branch
+	github.com/grafana/loki v1.6.2-0.20230522153400-ac4a442291cc // k146 branch
 	github.com/grafana/phlare/api v0.1.2
 	github.com/grafana/regexp v0.0.0-20221122212121-6b5c0a4cb7fd
 	github.com/grafana/snowflake-prometheus-exporter v0.0.0-20221213150626-862cad8e9538
 	github.com/grafana/tail v0.0.0-20230328181249-aa6682d7843a
 	github.com/grafana/vmware_exporter v0.0.4-beta
 	github.com/grpc-ecosystem/go-grpc-middleware v1.3.0
-	github.com/hashicorp/consul/api v1.18.0
+	github.com/hashicorp/consul/api v1.20.0
 	github.com/hashicorp/go-cleanhttp v0.5.2
 	github.com/hashicorp/go-discover v0.0.0-20220105235006-b95dfa40aaed
 	github.com/hashicorp/go-multierror v1.1.1
@@ -85,7 +85,7 @@ require (
 	github.com/k3d-io/k3d/v5 v5.4.8
 	github.com/lib/pq v1.10.7
 	github.com/mackerelio/go-osstat v0.2.3
-	github.com/miekg/dns v1.1.50
+	github.com/miekg/dns v1.1.53
 	github.com/minio/pkg v1.5.8
 	github.com/mitchellh/mapstructure v1.5.0
 	github.com/mitchellh/reflectwalk v1.0.2
@@ -150,7 +150,7 @@ require (
 	github.com/testcontainers/testcontainers-go v0.19.0
 	github.com/uber/jaeger-client-go v2.30.0+incompatible
 	github.com/vincent-petithory/dataurl v1.0.0
-	github.com/weaveworks/common v0.0.0-20221201103051-7c2720a9024d
+	github.com/weaveworks/common v0.0.0-20230119144549-0aaa5abd1e63
 	github.com/webdevops/azure-metrics-exporter v0.0.0-20230502203721-b2bfd97b5313
 	github.com/webdevops/go-common v0.0.0-20230502000651-d37d46be8ee7
 	github.com/wk8/go-ordered-map v0.2.0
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -101,14 +100,14 @@ require (
 	github.com/open-telemetry/opentelemetry-collector-contrib/exporter/jaegerexporter v0.79.0
 	github.com/open-telemetry/opentelemetry-collector-contrib/exporter/kafkaexporter v0.79.0
 	github.com/open-telemetry/opentelemetry-collector-contrib/exporter/loadbalancingexporter v0.79.0
-	github.com/open-telemetry/opentelemetry-collector-contrib/exporter/prometheusexporter v0.79.0
+	github.com/open-telemetry/opentelemetry-collector-contrib/exporter/prometheusexporter v0.77.0
```
ptodev, line 103: Not sure why this is getting downgraded?

darxriggs:
Could the update even use the newer `k157` branch? It contains another fix for promtail (https://github.com/grafana/loki/commit/954df433e98f659d006ced52b23151cb5eb2fdfa).

go.mod
```diff
@@ -48,15 +47,15 @@ require (
...
 	github.com/google/renameio/v2 v2.0.0
 	github.com/google/uuid v1.3.0
 	github.com/gorilla/mux v1.8.0
 	github.com/grafana/ckit v0.0.0-20230628095927-30fb40ef3315
 	github.com/grafana/cloudflare-go v0.0.0-20230110200409-c627cf6792f2
-	github.com/grafana/dskit v0.0.0-20230201083518-528d8a7d52f2
+	github.com/grafana/dskit v0.0.0-20230518162305-3c92c534827e
 	github.com/grafana/go-gelf/v2 v2.0.1
-	github.com/grafana/loki v1.6.2-0.20230414223651-220cbdd4f172 // k146 branch
+	github.com/grafana/loki v1.6.2-0.20230605151533-4d184de6debd // k146 branch
```
darxriggs, line 58: The comment should be updated too.

ptodev:
Thanks for updating Loki! I had a look at the build errors, and there were two of them. I took the liberty to push a fix for them in a [commit](https://github.com/grafana/agent/pull/3942/commits/09883a640d0577ecb3a375b8e15ffb67e6c23a1a) to your branch. Hopefully you don't mind! :) 

### Problem 1: Old gRPC version

The Agent build failed due to errors like these:

```
# github.com/sercand/kuberesolver/v4
../../../go/pkg/mod/github.com/sercand/kuberesolver/v4@v4.0.0/builder.go:96:56: invalid operation: cannot call non-function target.Endpoint (variable of type string)
```

I checked the kuberesolver [code](https://github.com/sercand/kuberesolver/blob/v4.0.0/builder.go), and it seems the variable that's erroring has a type from the gRPC module. The version of gRPC which they use in their [go.mod](https://github.com/sercand/kuberesolver/blob/v4.0.0/go.mod) is v1.53.0. The Agent currently uses an old version of gRPC for a reason which is out of date now. In the Agent's go.mod is a `replace` statement to substitute the version of gRPC, so when I removed that the error went away. 

### Problem 2: Old version of Prometheus

The other errors were like this:
```
../../../go/pkg/mod/github.com/grafana/loki@v1.6.2-0.20230706113538-2eb6232f9818/pkg/logql/vector/heap.go:16:21: s[i].F undefined (type promql.Sample has no field or method F)
```

Apparently the [new](https://github.com/prometheus/prometheus/blame/main/promql/value.go#L174-L180) Prometheus code differs from the [old](https://github.com/grafana/prometheus/blob/22c61d1811b2dc464fbe48c0c883df375eb43c7f/promql/value.go#L175-L179) one. There is also a `replace` statement for Prometheus, which is there for three reasons:

```
// TODO(rfratto): remove replace directive once:
//
// * We remove our dependency on Cortex, which forces Prometheus to an older
//   version since Go thinks v1 is newer than v0.
// * There is a release of Prometheus which contains prometheus/prometheus#12349.
// * We upgrade our OpenTelemetry dependency which will allow us to update Prometheus.
replace github.com/prometheus/prometheus => github.com/grafana/prometheus v1.8.2-0.20230511165250-22c61d1811b2 // release-2.42.0-grafana
```

At. this point in time, I think only the Cortex depenedency is holding us back from upgrading. I replaced the grafana fork with the latest Prometheus version in the `replace` statement, and the build mostly worked. I think we still need the `replace` statement, so that Cortex uses that new version. I'm not 100% sure if it's ok to merge the Prometheus update - it would be best to double check with @rfratto .

There was a recent [change](https://github.com/prometheus/prometheus/commit/b987afa7ef948461100b914fb6aea5e7195fd078) to the Prometheus label builders so I had to change some code as well.

rfratto:
> At this point in time, I think only the Cortex depenedency is holding us back from upgrading. 

To clarify my comment, the Cortex dependency (AFAIK) doesn't block us upgrading Prometheus, it just requires us to use a replace directive to specify the SHA of the specific commit we want to use. Without the replace directive, Go tooling will try to "upgrade" to the version of Prometheus that Cortex specifies (v1), which is actually an older version, since Prometheus recently started tagging their API as v0, since Go tooling also didn't know how to handle the v2 tags. 

TL;DR: We can update Prometheus, we just need a replace directive to explicitly set the version we're upgrading to.

converter/internal/prometheusconvert/testdata/azure.river
```diff
@@ -53,14 +53,8 @@ prometheus.remote_write "default" {
...
-		queue_config {
-			capacity             = 2500
-			max_shards           = 200
-			max_samples_per_send = 500
-		}
+		queue_config { }
 
-		metadata_config {
-			max_samples_per_send = 500
-		}
```
thampiotr, line 64: @erikbaranowski after the prometheus version upgrade, I believe the defaults set by prometheus library match those that we set in Flow Mode, thus they are no longer here.

I ran a debugger to be sure and took a look at what are the values right after we run this line:
```
promConfig, err := prom_config.Load(string(in), false, log.NewNopLogger())
```
And I can see that they match our [Flow Mode defaults](https://grafana.com/docs/agent/latest/flow/reference/components/prometheus.remote_write/#queue_config-block):
image

So I think this is an expected change.

It bring an interesting "issue": it matters what version of prometheus config are you migrating from... We should probably raise in documentation clearly that we support only the prometheus config version that matches our `go.mod`.

thampiotr (commented):
Thanks so much @ptodev! This was super helpful explanation and your changes look good - I'm always happy for you to push to my branches 😄 

@erikbaranowski - I've tagged you here because of a change to defaults in prometheus convert. 

@darxriggs - thanks for the suggestion, I've updated to `k157` branch.

thampiotr:
I just fixed one test, but we still have another issue, this time with agent operator:

```
➜  agent git:(thampiotr/promtail-goroutine-leak-update-loki) go test github.com/grafana/agent/pkg/operator
# github.com/thanos-io/thanos/pkg/store/storepb/prompb
../../go/pkg/mod/github.com/thanos-io/thanos@v0.30.0/pkg/store/storepb/prompb/samples.go:27:47: undefined: promql.Point
FAIL	github.com/grafana/agent/pkg/operator [build failed]
FAIL
```

```
➜  agent git:(thampiotr/promtail-goroutine-leak-update-loki) go mod why github.com/thanos-io/thanos/pkg/store/storepb
# github.com/thanos-io/thanos/pkg/store/storepb
github.com/grafana/agent/pkg/operator
github.com/prometheus-operator/prometheus-operator/pkg/operator
github.com/thanos-io/thanos/pkg/store/storepb
```

I tried upgrading `prometheus-operator` dependency to `0.66.0`, but I ran into more missing packages:
```
github.com/grafana/agent/component/prometheus/operator/configgen imports
	github.com/prometheus-operator/prometheus-operator/pkg/namespace-labeler: module github.com/prometheus-operator/prometheus-operator@latest found (v0.66.0), but does not contain package github.com/prometheus-operator/prometheus-operator/pkg/namespace-labeler
```

Will be looking into this a bit more, but any suggestions welcome, I don't have much context on the operator.

component/prometheus/operator/configgen/config_gen_podmonitor.go
```diff
@@ -9,7 +9,7 @@ import (
 	"strings"
 
 	promopv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
-	namespacelabeler "github.com/prometheus-operator/prometheus-operator/pkg/namespace-labeler"
+	"github.com/prometheus-operator/prometheus-operator/pkg/namespacelabeler"
```
thampiotr, line 12: This package was renamed in the latest `prometheus-operator` module.

converter/internal/staticconvert/testdata/prom_remote_write.river
```diff
@@ -1,65 +1,41 @@
...
-		queue_config {
-			capacity             = 2500
-			max_shards           = 200
-			max_samples_per_send = 500
-		}
+		queue_config { }
 
-		metadata_config {
-			max_samples_per_send = 500
-		}
```
thampiotr, line 14: I believe this is the same story as for the prometheus configs, with another interesting issue: when we update prometheus version and it changes defaults, the static mode's defaults will change too.

thampiotr:
Let's _not_ merge this before we cut 0.35, because I would like to get more testing of the prometheus upgrade.

converter/internal/prometheusconvert/testdata/azure.river
erikbaranowski: it's probably worth a look why queue_config shows up at all now. We should also update the doc page which does specify a version currently.

https://grafana.com/docs/agent/next/flow/reference/cli/convert/

converter/internal/staticconvert/testdata/prom_remote_write.river
erikbaranowski: good catch 😄 we should make sure the static config docs reflect that or override the static defaults to what they were. 

https://grafana.com/docs/agent/next/static/configuration/

converter/internal/prometheusconvert/testdata/azure.river
thampiotr: 1) I've fixed the docs now to point to right version of prom
2) The reason the empty blocks are printed out, I believe, is that in the default value of `endpoint`, the `queue_config` is `nil`, while when constructing the `endpoint` in conversion code, we instead set it to a struct with default values. See this debug screenshot:

image

I'm thinking that to solve it, we may need to call `reflect.New` and `SetToDefault()` on every field that is `nil` in the loop over fields above, but there may be better ways. The issue is that we detect it's a default only on a recursive call to `encodeFields`, after a block for `queue_config` is already added. We should detect it before we call into the recursive call, I think.

thampiotr: I did some more testing and found a few potential issues and generally it's a bit hard to determine how things should work in such cases where we have nested blocks with pointers and defaults. See: https://github.com/grafana/agent/pull/4430

converter/internal/staticconvert/testdata/prom_remote_write.river
thampiotr: Good point. We seem to defer to prometheus documentation and I've updated the link to point to the right version here: https://grafana.com/docs/agent/next/static/configuration/metrics-config/#metrics_instance_config 

We don't talk about defaults in staticmode, so I think we're good with the link update.

clayton-cornell (commented):
No changes for docs

PR #3953, opened 2023-05-23
feat: pyroscope.ebpf

korniltsev:
#### PR Description

This is a ebpf profiler from [pyroscope](https://github.com/grafana/pyroscope/tree/main/pkg/agent/ebpfspy),  with some changes:
 - libbpf replaced with cilium
 - libbcc functionality rewriten in go  (symtab package)

#### Notes to the Reviewer
There are binary files commited, not sure it is a good idea
 - symtab/testdata - elf files for symbol resolution tests
 -  compiled bpf programs which are embeded as //go:embed)  
#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated

#### Future improvements 
-  https://github.com/grafana/phlare/issues/787
- https://github.com/grafana/phlare/issues/789
- https://github.com/grafana/phlare/issues/790
- https://github.com/grafana/phlare/issues/788
- https://github.com/grafana/phlare/issues/791
- https://github.com/grafana/phlare/issues/792


component/pyroscope/ebpf/ebpfspy/session.go
```diff
@@ -0,0 +1,275 @@
...
+		return
+	}
+	var stackFrames []string
+	for i := 0; i < 127; i++ {
+		it := stack[i*8 : i*8+8]
+		ip := binary.LittleEndian.Uint64(it)
+		if ip == 0 {
+			break
+		}
+		sym := s.symCache.resolve(pid, ip, s.roundNumber)
```
luisgerhorst, line 232: Does this allow a user to distinguish between user space and kernel symbols? The kernel contains a lot of functions. Some of these could easily overlap with symbols from user applications or libraries. If one then uses Pyroscope's Sandwich-View feature, wouldn't these be merged?

luisgerhorst: Also, is there an option to exclude user and/or kernel symbols from the trace?

korniltsev: As of right now there is no way to distinguish between user/kernel.
There's no option to exclude user/kernel traces right now, but i think it's useful and i will add them.

component/pyroscope/ebpf/ebpfspy/session.go
```diff
@@ -0,0 +1,313 @@
...
+	_ "embed"
+	"encoding/binary"
+	"fmt"
+	"reflect"
+	"unsafe"
+
+	"github.com/cilium/ebpf"
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	"github.com/pyroscope-io/pyroscope/pkg/agent/ebpfspy/cpuonline"
```
luisgerhorst, line 18: I noticed this dependency because Pyroscope is AGPLv3 (https://github.com/grafana/pyroscope/blob/v0.37.2/LICENSE). Isn't this incompatible to Apache 2? I guess cpuonline is not that large and can be moved into the agent repo?

docs/sources/flow/reference/components/phlare.ebpf.md
```diff
@@ -0,0 +1,187 @@
...
+| `forward_to`       | `list(ProfilesReceiver)` | List of receivers to send collected profiles to.                       |                | yes      |  
+| `targets`          | `list(map(string))`      | List of targets to group profiles by container id                      |                | no       |   
+| `default_target`   | `map(string)`            | Default target to use when a PID is not associated with a container id |                | no       |    
+| `targets_only`     | `bool`                   | A flag to ignore profiles not associated with a container id           | false          | no       |     
+| `job_name`         | `string`                 | The job name to override the job label with.                           | component name | no       |      
+| `collect_interval` | `duration`               | How frequently to collect profiles                                     | `10s`          | no       |       
+| `sample_rate`      | `int`                    | How many times per second to collect samples                           | 100            | no       |
```
luisgerhorst: This says the default sampling rate is 100Hz. This is a bit odd to me because it is usually recommended to use odd/prime sampling rates (e.g., 19Hz for parca, or 97Hz)  in order to reduce the likelihood that the sampling interrupt always/frequentlty overlaps with some other event occurring at a certain frequency. However, I was not able to quickly find any hard data on this. Was this considered / was 100Hz chosen for a particular reason?

CHANGELOG.md
```diff
@@ -47,10 +47,12 @@ v0.34.0-rc.0 (2023-06-01)
   - `prometheus.exporter.oracledb` collects metrics from oracledb (@jonathanwamsley)
   - `prometheus.exporter.dnsmasq` collects metrics from a dnsmasq server. (@spartan0x117)
   - `loki.source.awsfirehose` - receive Loki log entries from AWS Firehose via HTTP (@thepalbi)
+  - `phlare.ebpf` collects system-wide performance profiles from the current host (@korniltsev)
```
cyriltovena, line 50: Can we rename this to pyroscope instead ?

cmd/grafana-agent/Dockerfile
```diff
@@ -44,5 +44,5 @@ COPY --from=build /src/agent/build/grafana-agent /bin/grafana-agent
 COPY cmd/grafana-agent/agent-local-config.yaml /etc/agent/agent.yaml
 
 
-ENTRYPOINT ["/bin/grafana-agent"]
-CMD ["--config.file=/etc/agent/agent.yaml", "--metrics.wal-directory=/etc/agent/data"]
+#ENTRYPOINT ["/bin/grafana-agent"]
```
cyriltovena, line 47: I think this needs to reverted.

component/pyroscope/ebpf/ebpfspy/bpf/profile.bpf.c
```diff
@@ -0,0 +1,67 @@
...
+
+	val = bpf_map_lookup_elem(&counts, &key);
+	if (val)
+		(*val)++;
+	else
+		bpf_map_update_elem(&counts, &key, &one, BPF_NOEXIST);
+	return 0;
+}
+
+char _license[] SEC("license") = "GPL"; //todo
```
cyriltovena, line 67: I think everything is Apache 2 here.

component/pyroscope/ebpf/ebpfspy/session.go
cyriltovena: Yes I agree either change back the license of `cpuonline` package to Apache 2.0 or move the code here.

docs/sources/flow/reference/components/pyroscope.ebpf.md
```diff
@@ -0,0 +1,178 @@
+---
+title: phlare.ebpf
+labels:
+  stage: beta
+---
+
+# phlare.ebpf
```
cyriltovena, line 7:
```suggestion
# pyroscope.ebpf
```

docs/sources/flow/reference/components/pyroscope.ebpf.md
```diff
@@ -0,0 +1,178 @@
+---
+title: phlare.ebpf
```
cyriltovena, line 2:
```suggestion
title: pyroscope.ebpf
```

docs/sources/flow/reference/components/pyroscope.ebpf.md
```diff
@@ -0,0 +1,178 @@
...
+The component configures and starts a new ebpf profiling job to collect performance profiles from the current host.
+
+The following arguments can be used to configure a `pyroscope.ebpf`. Only the
+`forward_to` field is required and any omitted fields take their default
+values.
+
+| Name                      | Type                     | Description                                                            | Default | Required |
+|---------------------------|--------------------------|------------------------------------------------------------------------|---------|----------|
+| `forward_to`              | `list(ProfilesReceiver)` | List of receivers to send collected profiles to.                       |         | yes      |  
+| `targets`                 | `list(map(string))`      | List of targets to group profiles by container id                      |         | no       |
```
cyriltovena, line 37: If this only works with containers we should specify it in the documentation summary.

docs/sources/flow/reference/components/pyroscope.ebpf.md
```diff
@@ -0,0 +1,178 @@
...
+`forward_to` field is required and any omitted fields take their default
+values.
+
+| Name                      | Type                     | Description                                                            | Default | Required |
+|---------------------------|--------------------------|------------------------------------------------------------------------|---------|----------|
+| `forward_to`              | `list(ProfilesReceiver)` | List of receivers to send collected profiles to.                       |         | yes      |  
+| `targets`                 | `list(map(string))`      | List of targets to group profiles by container id                      |         | no       |   
+| `default_target`          | `map(string)`            | Default target to use when a PID is not associated with a container id |         | no       |    
+| `targets_only`            | `bool`                   | A flag to ignore profiles not associated with a container id           | true    | no       |      
+| `collect_interval`        | `duration`               | How frequently to collect profiles                                     | `10s`   | no       |
```
cyriltovena, line 40: Can we default to 15s for consistency

docs/sources/flow/reference/components/pyroscope.ebpf.md
```diff
@@ -0,0 +1,178 @@
...
+components.
+
+## Component health
+
+`pyroscope.ebpf` is only reported as unhealthy if given an invalid
+configuration.
+
+## Debug information
+
+`pyroscope.ebpf` does not expose any component-specific debug information.
```
cyriltovena, line 60: Not for this PR but I think in the future we probably want to show active targets. May be we should create an issue for this.

docs/sources/flow/reference/components/pyroscope.ebpf.md
```diff
@@ -0,0 +1,178 @@
...
+The `pyroscope.ebpf` component is designed to collect stacktraces associated with a process running on the current host.
+Stack traces are collected according to the defined `sample_rate`, meaning traces are gathered this many times
+per second.
+
+The following labels are automatically injected to the collected profiles if not found and can help pin down a
+profiling target.
+
+| Label              | Description                                                                                                                                                                                            |
+|--------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `service_name`     | Pyroscope service name. It's automatically selected from discovery meta labels if possible. Otherwise defaults to the fully formed component name or a value from `pyroscope.ebpf.service_name` field. |
```
cyriltovena, line 85: is `pyroscope.ebpf.service_name` a label name ? Why not keep the `service_name` if it's provided via the target labels ?

docs/sources/flow/reference/components/pyroscope.ebpf.md
```diff
@@ -0,0 +1,178 @@
...
+### Container ID
+
+Each collected stack trace is then associated with a specified target from the targets list, determined by a
+container ID. This association process involves checking the `__container_id__`, `__meta_docker_container_id`,
+and `__meta_kubernetes_pod_container_id` labels of a target against the `/proc/{pid}/cgroup` of a process.
+
+If a corresponding container ID is found, the stack traces are aggregated per target based on the container ID.
+However, if a container ID is not identified, the stack trace is then associated with a `default_target`.
+
+If the `targets_only` is set to true, any stacktrace not associated with a listed target
```
cyriltovena, line 98:
```suggestion
If the `targets_only` is set to true, any stacktraces not associated with a listed target
```

docs/sources/flow/reference/components/pyroscope.ebpf.md
cyriltovena: If this is the case I think we should also create an issue to support non-container workload.

We could probably create a service discovery finds process, allow to filter them and send target with the associated `___process_pid__`.

docs/sources/flow/reference/components/pyroscope.ebpf.md
```diff
@@ -0,0 +1,178 @@
...
+
+The following arguments can be used to configure a `pyroscope.ebpf`. Only the
+`forward_to` field is required and any omitted fields take their default
+values.
+
+| Name                      | Type                     | Description                                                            | Default | Required |
+|---------------------------|--------------------------|------------------------------------------------------------------------|---------|----------|
+| `forward_to`              | `list(ProfilesReceiver)` | List of receivers to send collected profiles to.                       |         | yes      |  
+| `targets`                 | `list(map(string))`      | List of targets to group profiles by container id                      |         | no       |   
+| `default_target`          | `map(string)`            | Default target to use when a PID is not associated with a container id |         | no       |
```
cyriltovena, line 38: I wonder if we should get rid of this and just drop those. I'd rather defer this to another service discovery component ? WDYT ?

cyriltovena: This would also allow to get rid of `targets_only`, less config option the better.

docs/sources/flow/reference/stdlib/regex_quote.md
```diff
@@ -0,0 +1,17 @@
...
+
+# regex_quote
+
+The `regex_quote` function returns a string that escapes all regular expression metacharacters inside the argument 
+text. The returned string is a regular expression matching the literal text.
+
+## Examples
+
+```
+> regex_escape("ip-10-0-69-8.ec2.internal")
```
cyriltovena, line 15: regex_escape or regex_quote ?

do we actually need/use this ? @rfratto can you share your feedback on this one ?

component/pyroscope/ebpf/ebpf_linux.go
```diff
@@ -0,0 +1,229 @@
...
+					return nil
+				}
+				c.appendable.UpdateChildren(newArgs.ForwardTo)
+				if c.args.CollectInterval != collectInterval {
+					t.Reset(c.args.CollectInterval)
+					collectInterval = c.args.CollectInterval
+				}
+			case <-t.C:
+				var err error
+				pprof.Do(
```
cyriltovena, line 147: Can you comment what this does and why it's required ?

component/pyroscope/ebpf/ebpf_linux.go
```diff
@@ -0,0 +1,229 @@
...
+	c.targetFinder.ResizeContainerIDCache(c.args.ContainerIDCacheSize)
+}
+
+func (c *Component) Update(args component.Arguments) error {
+	newArgs := args.(Arguments)
+	c.argsUpdate <- newArgs
+	return nil
+}
+
+func (c *Component) reset() error {
```
cyriltovena, line 196: reset sounds like a component reset but in fact it sends data, may be we should find a better name.

docs/sources/flow/reference/components/pyroscope.ebpf.md
luisgerhorst: I think it also works without containers for the whole system if you set `targets_only=false` and `default_target={service_name="System"}` but this took me some time to click.

component/pyroscope/ebpf/ebpfspy/sd/target.go
```diff
@@ -0,0 +1,194 @@
...
+	l          log.Logger
+	cid2target map[containerID]*Target
+
+	// todo make it never evict during a reset
+	containerIDCache *lru.Cache[uint32, containerID]
+	defaultTarget    *Target
+	metrics          *metrics.Metrics
+}
+
+func NewTargetFinder(l log.Logger, containerIdCacheSize int, metrics *metrics.Metrics) (*TargetFinder, error) {
```
cyriltovena, line 95: Would be great to add some tests for the target finder.

component/pyroscope/ebpf/ebpfspy/sd/target.go
```diff
@@ -0,0 +1,194 @@
...
+			t, err := NewTarget(cid, target)
+			if err != nil {
+				_ = level.Error(s.l).Log(
+					"msg", "target skipped",
+					"target", target.Labels().String(),
+					"err", err,
+				)
+				continue
+			}
+			//_ = level.Debug(s.l).Log("created target", t.labels.String())
```
cyriltovena, line 129:
```suggestion
```

component/pyroscope/ebpf/ebpfspy/bpf/profile.bpf.c
cyriltovena: ho interesting ! Thank you.

cyriltovena: If that's the case we need a license header or in the folder.

docs/sources/flow/reference/components/pyroscope.ebpf.md
cyriltovena: yeah I want to make it more explicit via a new service discovery component.

docs/sources/flow/reference/stdlib/regex_quote.md
korniltsev: no, it is not needed anymore.  Initially I was filtering nodes by regex as discovery relabel, but it is a wrong approach and discovery.kubernetes.selectors is now used.
I will remove it.

component/pyroscope/ebpf/ebpfspy/bpf/profile.bpf.c
korniltsev: Added a license header

cmd/grafana-agent/Dockerfile
korniltsev: reverted

component/pyroscope/ebpf/ebpfspy/session.go
korniltsev: moved it here

docs/sources/flow/reference/components/phlare.ebpf.md
korniltsev: made it 97, thanks

docs/sources/flow/reference/components/pyroscope.ebpf.md
korniltsev: made it 15s

docs/sources/flow/reference/components/pyroscope.ebpf.md
korniltsev: I agree it is confusing and I needed it mostly for local development.
I removed both default_targets and targets_only

docs/sources/flow/reference/stdlib/regex_quote.md
korniltsev: removed

component/pyroscope/ebpf/ebpfspy/sd/target.go
```diff
@@ -0,0 +1,194 @@
...
+
+func inferServiceName(target discovery.Target) string {
+	k8sServiceName := target[labelServiceNameK8s]
+	if k8sServiceName != "" {
+		return k8sServiceName
+	}
+	k8sNamespace := target["__meta_kubernetes_namespace"]
+	k8sContainer := target["__meta_kubernetes_pod_container_name"]
+	if k8sNamespace != "" && k8sContainer != "" {
+		return fmt.Sprintf("ebpf/%s/%s", k8sNamespace, k8sContainer)
```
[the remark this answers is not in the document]
korniltsev: I like the idea to add agent=ebpf label
I don't like the idea to remove the prefix because if we remove, the profiles will be merged for both ebfp and scraped profiles because of the same service_name and it may be confusing as to why I see 4 hours of data when I expect just 2 hours of data, the stacktraces will mostlikely differ nad not merge perfectly, so there will be 2 almost identical profiles. 
Coming from og pyroscope as a user I realy just want to do a one mouse click and get the correct and expected app. And now I have to always  keep in mind to query for 2 labels for both ebpf and scraped profiles

docs/sources/flow/reference/components/pyroscope.ebpf.md
luisgerhorst: Hmm, actually I found the feature to be very useful. What about removing the documentation and keeping the options until something like `discovery.whole_system` is available?

component/pyroscope/ebpf/ebpf_linux.go
korniltsev: we don't need it. removed it.

it was just handy to separate ebpf.reset profile from the rest (mostly service discovery) during development

docs/sources/flow/reference/components/pyroscope.ebpf.md
korniltsev: no, it was a leftover. 

Now we have the following section

```
### Service name

The special label `service_name` is required and must always be present. If it's not specified, it is
attempted to be inferred from multiple sources:

- `__meta_kubernetes_pod_annotation_pyroscope_io_service_name` which is a `pyroscope.io/service_name` pod annotation.
- `__meta_kubernetes_namespace` and `__meta_kubernetes_pod_container_name`
- `__meta_docker_container_name`

If `service_name` is not specified and could not be inferred it is set to `unspecified`.
```

And the following row in the table

```
| `service_name`     | Pyroscope service name. It's automatically selected from discovery meta labels if possible. Otherwise defaults to `unspecified`. |
```

Does it look good now?

component/pyroscope/ebpf/ebpfspy/sd/target.go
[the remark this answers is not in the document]
cyriltovena: >the profiles will be merged for both ebfp and scraped profiles because of the same service_name

Yes but this our problem, we should never have configured the same service name to begging with. We should use relabeling on our ebpf config to add the prefix, not make it default. WDYT ?

component/pyroscope/ebpf/ebpfspy/metrics/metrics.go
```diff
@@ -0,0 +1,16 @@
+package metrics
+
+import "github.com/prometheus/client_golang/prometheus"
+
+type Metrics struct {
```
cyriltovena, line 5: May be let's start with 2 metrics.

How many targets are active (Gauge)
Total profiling sessions succeed/failed.

docs/sources/flow/reference/components/pyroscope.ebpf.md
cyriltovena: I think you added this ? We should update the doc then.

component/pyroscope/ebpf/ebpfspy/session.go
```diff
@@ -0,0 +1,321 @@
...
+	"github.com/grafana/agent/component/pyroscope/ebpf/ebpfspy/cpuonline"
+
+	"github.com/cilium/ebpf"
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	"github.com/grafana/agent/component/pyroscope/ebpf/ebpfspy/symtab"
+	"golang.org/x/sys/unix"
+
+	"github.com/grafana/agent/component/pyroscope/ebpf/ebpfspy/metrics"
+
```
cyriltovena, line 24: I think you need to reorder those import for the linter.

component/pyroscope/ebpf/ebpfspy/session.go
```diff
@@ -0,0 +1,321 @@
...
+	CollectUser   bool
+	CollectKernel bool
+}
+
+type Session struct {
+	logger     log.Logger
+	pid        int
+	sampleRate int
+
+	targetFinder *sd.TargetFinder
```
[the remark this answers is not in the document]
cyriltovena: I see you also use `sd.Target` so not sure if my comment is applicable.

component/pyroscope/ebpf/ebpfspy/session.go
```diff
@@ -0,0 +1,321 @@
...
+
+func (s *Session) UpdateProfileOptions(options ProfileOptions) {
+	s.ProfileOptions = options
+}
+
+func (s *Session) PidCacheDebugInfo() symtab.GCacheDebugInfo[symtab.ProcTableDebugInfo] {
+	return s.symCache.PidCacheDebugInfo()
+}
+
+func reverse(s []string) {
```
cyriltovena, line 290: You can use `lo.Reverse(`

component/pyroscope/ebpf/metrics.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+func NewMetrics(reg prometheus.Registerer) *Metrics {
+	m := &Metrics{
+		TargetsActive: prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "pyroscope_ebpf_active_targets",
+			Help: "Current number of active targets being tracked by the ebpf component",
+		}),
+		ProfilingSessionsTotal: prometheus.NewCounter(prometheus.CounterOpts{
+			Name: "pyroscope_ebpf_profiling_sessions_total",
```
cyriltovena, line 18: might be interesting to have success vs fail there. This way we can alert if it's failing.

component/pyroscope/ebpf/ebpfspy/pprof.go
```diff
@@ -0,0 +1,99 @@
...
+		return res
+	}
+	builder := &ProfileBuilder{
+		locations: make(map[string]*profile.Location),
+		functions: make(map[string]*profile.Function),
+		Labels:    labels,
+		Profile: &profile.Profile{
+			Mapping: []*profile.Mapping{
+				{
+					ID: 1,
```
cyriltovena, line 33: Any way we can set the mapping filename based on the name of the binary ? On the database side we're using this to shard stacktraces.

Nothing urgent since we fallback to `service_name`

cyriltovena:
Should we remove profile_bpfel.o and profile_bpfeb.o ?

component/pyroscope/ebpf/ebpfspy/pprof.go
korniltsev: Sure, we can add it. 
It may be a bit confusing, because there may be multiple executable and multiple mappings with same addresses, but different files because the pprof contains stacktraces from multiple processes.

korniltsev:
>  Should we remove profile_bpfel.o and profile_bpfeb.o ?
I have same question regarding `elf/testdata`

It looks like according to pkg/integrations/snmp_exporter/common/common.go, the way to do it here in agent repo is to have a generation step, but commit it anyway, because the generation steps are currently not invoked during a build.

component/pyroscope/ebpf/ebpfspy/session.go
```diff
@@ -0,0 +1,321 @@
...
+		symCache:       symCache,
+		sampleRate:     sampleRate,
+		targetFinder:   serviceDiscovery,
+		ProfileOptions: profileOptions,
+	}, nil
+}
+
+func (s *Session) Start() error {
+	var err error
+	if err = unix.Setrlimit(unix.RLIMIT_MEMLOCK, &unix.Rlimit{
```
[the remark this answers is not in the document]
korniltsev: It is mostly for old kernels to increase probability to load bpf programs successfully

https://lore.kernel.org/bpf/20201201215900.3569844-1-guro@fb.com/t/#u
https://github.com/cilium/ebpf/blob/1140a754d780a64d78cd2e7d02d96f0aab694200/prog.go#L360
https://github.com/cilium/ebpf/blob/1140a754d780a64d78cd2e7d02d96f0aab694200/rlimit/rlimit.go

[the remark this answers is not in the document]
korniltsev: oh I should switch to cilium/rlimit because it is smarter and not lifting the limit if not necessery

component/pyroscope/ebpf/ebpfspy/sd/procfs.go
```diff
@@ -0,0 +1,37 @@
...
+func (tf *TargetFinder) getContainerIDFromPID(pid uint32) containerID {
+	f, err := tf.fs.Open(fmt.Sprintf("proc/%d/cgroup", pid))
+	if err != nil {
+		return ""
+	}
+	defer f.Close()
+
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		line := scanner.Text()
```
cyriltovena, line 22: I think Scanner has a method `Bytes()` and you can run the regex on it. This way you can avoid an allocation for each line.

component/pyroscope/ebpf/ebpfspy/pprof.go
cyriltovena: Ok let's start like this then we can experiment later.

cyriltovena: Don't change anything :)

component/pyroscope/ebpf/ebpf_linux.go
```diff
@@ -0,0 +1,253 @@
...
+
+	if err != nil {
+		return fmt.Errorf("ebpf session CollectProfiles %w", err)
+	}
+	level.Debug(c.options.Logger).Log("msg", "ebpf  CollectProfiles done", "profiles", len(builders.Builders))
+	bytesSent := 0
+	for _, builder := range builders.Builders {
+		c.metrics.PprofsTotal.Inc()
+		var buf bytes.Buffer
+		err := builder.Profile.Write(&buf)
```
cyriltovena, line 214: Looking at `Write` implementation https://github.com/google/pprof/blob/91b7bce49751/profile/profile.go#L341

I think it might be better to use `WriteUncompressed` so that you can pool gzip writers (by resetting) and use this one instead https://pkg.go.dev/github.com/klauspost/compress/gzip#NewWriter

cyriltovena: Not sure if this will help but worth trying since it ~45% of alloc space 
image

component/pyroscope/ebpf/ebpfspy/sd/procfs.go
```diff
@@ -0,0 +1,37 @@
+package sd
+
+import (
+	"bufio"
+	"fmt"
+	"regexp"
+)
+
+var (
+	cgroupContainerIDRe = regexp.MustCompile(`^.*/(?:.*-)?([0-9a-f]+)(?:\.|\s*$)`)
```
cyriltovena, line 10: Can you add a comment for reference how it looks like

```
sudo cat /proc/3288/cgroup
0::/system.slice/docker-7afe91f93cb1d9a392e9e3229da9c6c4e6d98eba756e443a35055a0c7c9290e8.scope
```

component/pyroscope/ebpf/ebpfspy/session.go
```diff
@@ -0,0 +1,325 @@
...
+	return res
+}
+
+func (s *session) walkStack(sb *stackBuilder, stack []byte, pid uint32) {
+	if len(stack) == 0 {
+		return
+	}
+	var stackFrames []string
+	for i := 0; i < 127; i++ {
+		it := stack[i*8 : i*8+8]
```
cyriltovena, line 256: Would be great to explain/comment the layout of `stack []byte` it and ip doesn't ring a bell to me.

component/pyroscope/ebpf/ebpfspy/symtab/gcache.go
```diff
@@ -0,0 +1,166 @@
...
+	round int
+}
+type entry[V Resource] struct {
+	v     V
+	round int
+}
+
+type GCacheOptions struct {
+	Size       int
+	KeepRounds int
```
cyriltovena, line 30: What does `KeepRounds` do ? May be add a comment also about size that is it in amount of element not size of them.

component/pyroscope/ebpf/ebpfspy/sd/procfs.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+func (tf *TargetFinder) getContainerIDFromPID(pid uint32) containerID {
+	f, err := tf.fs.Open(fmt.Sprintf("proc/%d/cgroup", pid))
```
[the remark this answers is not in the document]
cyriltovena: Ho! You do this already in the session.

component/pyroscope/ebpf/ebpfspy/sd/target.go
```diff
@@ -0,0 +1,220 @@
...
+	res := tf.findTarget(pid)
+	if res != nil {
+		return res
+	}
+	return tf.defaultTarget
+}
+
+func (tf *targetFinder) findTarget(pid uint32) *Target {
+	cid, ok := tf.containerIDCache.Get(pid)
+	if ok && cid != "" {
```
cyriltovena, line 176: What are the chances that a PID that has no associated containerID gets one at some point ? I wonder if we should cache miss at least for a few hours ?

component/pyroscope/ebpf/ebpfspy/bpf/profile.bpf.c
```diff
@@ -0,0 +1,69 @@
+// SPDX-License-Identifier: GPL-2.0-only
```
rfratto, line 1: We can't import GPL code into Grafana Agent.

rfratto:
This is quite a bit of code to bring into the project. Does it make sense for this code to exist elsewhere and for us to import it?

component/pyroscope/ebpf/ebpfspy/bpf/profile.bpf.c
luisgerhorst: For reference, here's an earlier discussion on that: https://github.com/grafana/agent/pull/3953#discussion_r1231996675

rfratto: Ah, thanks! That got buried a bit in the history so I missed it

korniltsev:
I like the idea to keep it elsewhere and use it as a library

I guess one downside could be that it would receive less attention during development and codereview

What could be other downsides?

cyriltovena:
>This is quite a bit of code to bring into the project. Does it make sense for this code to exist elsewhere and for us to import it?

We can have in in pyroscope as a go module or have it here. We can always changed our mind at anytime.

I would personally prefer it closer to the agent so we can work and test locally smoothly but not opposed to use a go module.

component/pyroscope/ebpf/ebpfspy/pprof/pprof.go
```diff
@@ -123,10 +123,10 @@ func (p *ProfileBuilder) addFunction(function string) *profile.Function {
 }
 
 func (p *ProfileBuilder) Build() ([]byte, error) {
```
cyriltovena, line 125: I would do `Write(io.Writer) error` instead. This way the user can deal with lifecycle of the io.Writer.

In this method you can get from the gzip writer from the pool, reset to the io.Writer received, and finally put back the writer.

component/pyroscope/ebpf/ebpf_linux.go
korniltsev: pooling gzip writers increased inuse mem by ~40M and decreased cpu ~10%. lets keep it pooled and if we ever need to save more memory we could switch back or introduce a configuration option.

docs/sources/flow/reference/components/pyroscope.ebpf.md
```diff
@@ -0,0 +1,250 @@
...
+
+If you only see module names (e.g., `/lib/x86_64-linux-gnu/libc.so.6`) without corresponding function names, this
+indicates that the symbols couldn't be mapped to their respective function names.
+
+This can occur for a number of reasons:
+
+- The binary has been stripped, leaving no .symtab, .dynsym, or .gopclntab sections in the ELF file.
+- The debug file is missing or could not be located.
+
+To fix this for your binaries, ensure that they either not stripped or that you have separate
```
clayton-cornell, line 148:
```suggestion
To fix this for your binaries, ensure that they are either not stripped or that you have separate
```

docs/sources/flow/reference/components/pyroscope.ebpf.md
```diff
@@ -0,0 +1,250 @@
...
+executing:
+
+```bash
+apt install libc6-dbg
+```
+
+### Understanding Flat Stack Traces
+
+If your profiles show a lof of shallow stack traces, typically 1-2 frames deep, it might mean that
+your binary was compiled without frame pointers.
```
clayton-cornell, line 167:
```suggestion
If your profiles show many shallow stack traces, typically 1-2 frames deep, your binary might have been compiled without frame pointers.
```

docs/sources/flow/reference/components/pyroscope.ebpf.md
```diff
@@ -0,0 +1,250 @@
...
+
+To compile your code with frame pointers, include the `-fno-omit-frame-pointer` flag in your compiler options.
+
+### Profiling Interpreted Languages
+
+Currently, profiling interpreted languages like Python, Ruby, JavaScript, etc., is not ideal using this implementation.
```
clayton-cornell, line 173:
```suggestion
Profiling interpreted languages like Python, Ruby, JavaScript, etc., is not ideal using this implementation.
```

docs/sources/flow/reference/components/pyroscope.ebpf.md
```diff
@@ -0,0 +1,250 @@
...
+The JIT-compiled methods in these languages are typically not in ELF file format, demanding additional steps for
+profiling. For instance, using perf-map-agent and enabling frame pointers for Java.
+
+Interpreted methods will display the interpreter function’s name, rather than the actual function. Future releases of
+the profiler are expected to support profiling of Ruby and Python code more effectively.
```
clayton-cornell, line 178:
```suggestion
Interpreted methods will display the interpreter function’s name rather than the actual function. Future releases of
the profiler are expected to support profiling of Ruby and Python code more effectively.
```

Is this (Future releases of the profiler...) a forward looking statement?  Do we want to make this claim? Generally we do NOT want to have any forward looking statements making functionality claims that "might" be available.

docs/sources/flow/reference/components/pyroscope.ebpf.md
```diff
@@ -0,0 +1,250 @@
...
+
+## Example
+
+### Kubernetes discovery
+
+In the following example, performance profiles are collected from pods on the same node, discovered using
+`discovery.kubernetes`. Pod selection relies on the `HOSTNAME` environment variable, which is a pod name if the agent is
+used as a Grafana agent helm chart. Service name is set to `{namespace}/{container_name}` from kubernetes meta labels.
```
clayton-cornell, line 186:
```suggestion
used as a Grafana agent helm chart. The service name is set to `{namespace}/{container_name}` from Kubernetes meta labels.
```

Is Service name a variable? If it is a variable, we need to format it correctly here. Looking at the Docker example it's called "the `service_name` label"... is it the same thing here.

Is it line 204 that is being referred to here? If so there seems to be a disconnect between 

> Service name is set to `{namespace}/{container_name}`

and this line from the example:

> source_labels = ["__meta_kubernetes_namespace", "__meta_kubernetes_pod_container_name"]

docs/sources/flow/reference/components/pyroscope.ebpf.md
```diff
@@ -0,0 +1,250 @@
...
+  endpoint {
+    url = "http://pyroscope:4100"
+  }
+}
+```
+
+### Docker discovery
+
+The following example collects performance profiles from containers discovered by `discovery.docker` and ignores all
+other profiles collected from outside any docker container. `service_name` label is set to
```
clayton-cornell, line 223:
```suggestion
other profiles collected from outside any docker container. The `service_name` label is set to the
```

clayton-cornell (commented):
Some inline suggestions and several questions that need clarification

docs/sources/flow/reference/components/pyroscope.ebpf.md
korniltsev: I removed the forward looking statement. Removed the coma manually.

docs/sources/flow/reference/components/pyroscope.ebpf.md
korniltsev: > Is Service name a variable?

Service name is a label -  `service_name`.

> Is it line 204 that is being referred to here?

Yes

> If so there seems to be a disconnect between

Do you mean it should be `{__meta_kubernetes_namespace}/{__meta_kubernetes_pod_container_name}`

docs/sources/flow/reference/components/pyroscope.ebpf.md
```diff
@@ -0,0 +1,250 @@
...
+- `/usr/lib/debug/.build-id/01/0123456789abcdef.debug`
+- `/lib/x86_64-linux-gnu/libc.so.6.debug`
+- `/lib/x86_64-linux-gnu/.debug/libc.so.6.debug`
+- `/usr/lib/debug/lib/x86_64-linux-gnu/libc.so.6.debug`
+
+### Dealing with Unknown Symbols
+
+If you encounter unknown symbols in the profiles you’ve collected, this indicates that we couldn't access an ELF file
+associated to a given address in trace. This might be due to the ELF file being inaccessible, or the absence of an entry
+in `/proc/pid/maps` for a particular address (this can be the case with JIT-compiled functions).
```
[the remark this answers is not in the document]
korniltsev: > Who is we here?

The ebpf profiler component of grafana agent. Specifically the part that is responsible for stack trace symbolization. Let's replace "we" with "ebpf profiler" or "pyroscope.ebpf component" or "ebpf profiler component"

docs/sources/flow/reference/components/pyroscope.ebpf.md
```diff
@@ -0,0 +1,250 @@
...
+## Troubleshooting Unknown Symbols
+
+Symbols are extracted from various sources including:
+
+- The `.symtab` and `.dynsym` sections in the ELF file.
+- The `.symtab` and `.dynsym` sections in the debug ELF file.
+- The `.gopclntab` section in Go language ELF files.
+
+The search for debug files follows [gdb algorithm](https://sourceware.org/gdb/onlinedocs/gdb/Separate-Debug-Files.html).
+For example if it wants to find the debug file
```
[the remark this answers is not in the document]
korniltsev: same as in https://github.com/grafana/agent/pull/3953/files#r1252552212

docs/sources/flow/reference/components/pyroscope.ebpf.md
```diff
@@ -0,0 +1,250 @@
...
+* `pyroscope_ebpf_profiling_sessions_failing_total` (counter): Number of profiling sessions failed.
+* `pyroscope_ebpf_pprofs_total` (counter): Number of pprof profiles collected by ebpf component.
+
+## Profile collecting behavior
+
+The `pyroscope.ebpf` component is designed to collect stacktraces associated with a process running on the current host.
+Stack traces are collected according to the defined `sample_rate`, meaning traces are gathered this many times
+per second.
+
+The following labels are automatically injected to the collected profiles if not found and can help pin down a
```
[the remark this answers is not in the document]
korniltsev: The labels.

So if `service_name` is provided by the user configuration - it is used as is, if not - it is injected by the profiler. 
Same for other labels `__name__` and `__container_id__`

docs/sources/flow/reference/components/pyroscope.ebpf.md
```diff
@@ -0,0 +1,250 @@
...
+* `pyroscope_ebpf_active_targets` (gauge): Number of active targets tracked by the component.
+* `pyroscope_ebpf_profiling_sessions_total` (counter): Number of profiling sessions completed.
+* `pyroscope_ebpf_profiling_sessions_failing_total` (counter): Number of profiling sessions failed.
+* `pyroscope_ebpf_pprofs_total` (counter): Number of pprof profiles collected by ebpf component.
+
+## Profile collecting behavior
+
+The `pyroscope.ebpf` component is designed to collect stacktraces associated with a process running on the current host.
+Stack traces are collected according to the defined `sample_rate`, meaning traces are gathered this many times
+per second.
```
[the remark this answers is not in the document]
korniltsev: `sample_rate` is the profiler argument which has default value so it may be defined by the user.

> The sample rate is measured in traces collected per second.
The sample rate is not measured.
If user specified `sample_rate` argument as 239, then stack traces are collected 239 times per second
If user does not specify `sample_rate`, then default value of 97 is used and stack traces are collected 97 times per second.

docs/sources/flow/reference/components/pyroscope.ebpf.md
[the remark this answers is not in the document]
korniltsev: What I meant with "if not found" is that these labels will not be overwritten if specified by the user, they are only injected if not specified by the user

docs/sources/flow/reference/components/pyroscope.ebpf.md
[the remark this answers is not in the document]
korniltsev: I've replaced it with "the profiler"

docs/sources/flow/reference/components/pyroscope.ebpf.md
korniltsev: I've replaced the sentence with 

The `service_name` label is set to `{__meta_kubernetes_namespace}/{__meta_kubernetes_pod_container_name}` from kubernetes meta labels.

korniltsev:
@clayton-cornell Thank you for your feedback.  I've applied fixes and replied to the questions. I'd appreciate another round  of docs review

docs/sources/flow/reference/components/pyroscope.ebpf.md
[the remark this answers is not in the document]
clayton-cornell: OK, can we word it this way:

> The following labels are automatically injected into the collected profiles if you have not defined them. These labels can help you pin down a profiling target.

clayton-cornell (commented):
@korniltsev I've replied and tried to clarify the suggestions that were still open.

clayton-cornell:
Looks good for docs now :-)

korniltsev:
@rfratto @cyriltovena The PR is ready for review and potential merge. Apologies for a long and dirty branch.

cyriltovena (approved):
LGTM @rfratto can we get this merge please ?

docs/sources/flow/reference/components/pyroscope.ebpf.md
```diff
@@ -0,0 +1,245 @@
...
+the list of receivers passed in
+`forward_to`.
+
+You can specify multiple `pyroscope.ebpf` components by giving them different labels, however it is not recommended.
+
+## Usage
+
+```river
+pyroscope.ebpf "LABEL" {
+  forward_to = RECEIVER_LIST
```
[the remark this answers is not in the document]
korniltsev: good catch.

It `targets` used to be not required at the beginning - may be omitted if used with `targets_only=false && default_target={...}` .
Then we made it required and removed `targets_only` and `default_target` arguments
And now it is required (but only in documentation) and we also returned back `targets_only` and `default_target` arguemnts, but we do not document them and do not commit to support them. This may be useful if someone want's to profile the whole system, not only pods. This is "temporary"  :)  until I implement other service discovery options to profile specific processes/systemd units or something else.

I've added targets to the example as you suggested. Thanks.

component/pyroscope/ebpf/ebpf_linux.go
```diff
@@ -0,0 +1,266 @@
...
+				opts.Logger,
+				targetFinder,
+				convertSessionOptions(arguments, ms),
+			)
+			if err != nil {
+				return nil, fmt.Errorf("ebpf session create: %w", err)
+			}
+
+			return New(opts, arguments, session, targetFinder, ms)
+		},
```
rfratto, line 48: Normally, the Build function just calls New, only passing the options and arugments to it. Does this extra logic need to be here or can it be moved to New?

component/pyroscope/ebpf/ebpf_placeholder.go
```diff
@@ -0,0 +1,3 @@
+//go:build !linux
+
+package ebpf
```
rfratto, line 3: Can you register a dummy implementation of the component here which logs a notice that pyroscope.ebpf is no-op on non-Linux platforms similarly to what we do in static mode for the [node_exporter integration](https://github.com/grafana/agent/blob/474fe21362810889f7c55206483613a61ccdb874/pkg/integrations/node_exporter/node_exporter_windows.go#L18-L21)?

docs/sources/flow/reference/components/pyroscope.ebpf.md
```diff
@@ -0,0 +1,246 @@
...
+---
+
+# pyroscope.ebpf
+
+{{< docs/shared lookup="flow/stability/beta.md" source="agent" >}}
+
+`pyroscope.ebpf` configures an ebpf profiling job for the current host. The collected performance profiles are forwarded
+to
+the list of receivers passed in
+`forward_to`.
```
rfratto, line 14: nit:

```suggestion
`pyroscope.ebpf` configures an ebpf profiling job for the current host. The collected performance profiles are forwarded
to the list of receivers passed in `forward_to`.
```

docs/sources/flow/reference/components/pyroscope.ebpf.md
```diff
@@ -0,0 +1,246 @@
...
+
+You can specify multiple `pyroscope.ebpf` components by giving them different labels, however it is not recommended.
```
rfratto, line 16: Can you expand on why it's not recommended in the documentation? Users might accidentally end up in a situation where they have more than one pyroscope.ebpf component if they are using two different external modules which both happen to use this component.

docs/sources/flow/reference/components/pyroscope.ebpf.md
```diff
@@ -0,0 +1,246 @@
...
+The special label `service_name` is required and must always be present. If it's not specified, it is
+attempted to be inferred from multiple sources:
+
+- `__meta_kubernetes_pod_annotation_pyroscope_io_service_name` which is a `pyroscope.io/service_name` pod annotation.
+- `__meta_kubernetes_namespace` and `__meta_kubernetes_pod_container_name`
+- `__meta_docker_container_name`
+
+If `service_name` is not specified and could not be inferred, it is set to `unspecified`.
+
+## Troubleshooting Unknown Symbols
```
rfratto, line 108: Docs recommendation are to use sentence case for headers:

```suggestion
## Troubleshooting unknown symbols
```

docs/sources/flow/reference/components/pyroscope.ebpf.md
```diff
@@ -0,0 +1,246 @@
...
+For example, if the profiler wants to find the debug file
+for `/lib/x86_64-linux-gnu/libc.so.6`
+with a `.gnu_debuglink` set to `libc.so.6.debug` and a build ID `0123456789abcdef`. The following paths are examined:
+
+- `/usr/lib/debug/.build-id/01/0123456789abcdef.debug`
+- `/lib/x86_64-linux-gnu/libc.so.6.debug`
+- `/lib/x86_64-linux-gnu/.debug/libc.so.6.debug`
+- `/usr/lib/debug/lib/x86_64-linux-gnu/libc.so.6.debug`
+
+### Dealing with Unknown Symbols
```
rfratto, line 126:
```suggestion
### Dealing with unknown symbols
```

docs/sources/flow/reference/components/pyroscope.ebpf.md
```diff
@@ -0,0 +1,246 @@
...
+
+Unknown symbols in the profiles you’ve collected indicate that the profiler couldn't access an ELF file ￼associated with a given address in the trace.
+
+This can occur for several reasons:
+
+- The process has terminated, making the ELF file inaccessible.
+- The ELF file is either corrupted or not recognized as an ELF file.
+- There is no corresponding ELF file entry in `/proc/pid/maps` for the address in the stack trace.
+
+### Addressing Unresolved Symbols
```
rfratto, line 136:
```suggestion
### Addressing unresolved symbols
```

docs/sources/flow/reference/components/pyroscope.ebpf.md
```diff
@@ -0,0 +1,246 @@
...
+```
+
+For system libraries, ensure that debug symbols are installed. On Ubuntu, for example, you can install them by
+executing:
+
+```bash
+apt install libc6-dbg
+```
+
+### Understanding Flat Stack Traces
```
rfratto, line 162:
```suggestion
### Understanding flat stack traces
```

docs/sources/flow/reference/components/pyroscope.ebpf.md
```diff
@@ -0,0 +1,246 @@
...
+
+If your profiles show many shallow stack traces, typically 1-2 frames deep, your binary might have been compiled without frame pointers.
+
+To compile your code with frame pointers, include the `-fno-omit-frame-pointer` flag in your compiler options.
+
+### Profiling Interpreted Languages
```
rfratto, line 168:
```suggestion
### Profiling interpreted languages
```

rfratto (commented):
LGTM after these comments are resolved :)

docs/sources/flow/reference/components/pyroscope.ebpf.md
clayton-cornell: Ooops I missed that in my review. :-P

component/pyroscope/ebpf/ebpf_linux.go
korniltsev: `session` and it's dependencies were created here to allow passing the `session`  as an argument to `New` function to be able to swap it in unit test. 

 I've moved and hardcoded the dependencies to the `New` function, now it is not  testable, so I've created another one in test `newTestComponent`

docs/sources/flow/reference/components/pyroscope.ebpf.md
korniltsev:
```
You can specify multiple `pyroscope.ebpf` components by giving them different labels, however it is not recommended as
it can lead to additional memory and CPU usage.
```
Do you think it is enough?

component/pyroscope/ebpf/ebpf_placeholder.go
korniltsev: done

PR #3956, opened 2023-05-23
Add new override hook for token encoding and use it for config secret conversion

### LINKED ISSUE #3901: Allow outputing of secrets as string when doing config conversion to Flow
Secrets in a config conversion to Flow current print out as `(secret)`. Implement a way to print them out as strings and apply it to existing conversions.

For example,

```river
basic_auth {
	username = "user"
	password = (secret)
}
```

becomes

```river
basic_auth {
	username = "user"
	password = "pass"
}
```

erikbaranowski:
#### PR Description

- Implements a new override hook for token encoding
- Leverage new hook for override secret token encoding value

Closes #3901 

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/converter/internal/common/common.go
+++ b/converter/internal/common/common.go
@@ -6,9 +6,9 @@ import (
 	"github.com/grafana/agent/pkg/river/token/builder"
 )
 
-// AppendBlock appends Flow component arguments using the convert value
-// override hook to the file we are building.
-func AppendBlock(f *builder.File, name []string, label string, args component.Arguments) {
+// AppendBlockWithOverride appends Flow component arguments using the convert
+// value override hook to the file we are building.
+func AppendBlockWithOverride(f *builder.File, name []string, label string, args component.Arguments) {
 	block := builder.NewBlock(name, label)
 	block.Body().SetValueOverrideHook(getValueOverrideHook())
 	block.Body().AppendFrom(args)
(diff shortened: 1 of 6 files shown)


converter/internal/common/common.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"github.com/grafana/agent/component"
+	"github.com/grafana/agent/pkg/river/rivertypes"
+	"github.com/grafana/agent/pkg/river/token/builder"
+)
+
+// AppendBlock appends Flow component arguments using the convert value
+// override hook to the file we are building.
+func AppendBlock(f *builder.File, name []string, label string, args component.Arguments) {
```
erikbaranowski, line 11: This is to save repeated code. It may evolve as more gets added to the convert API or removed if the pattern becomes inconsistent but I'd like it for now.

mattdurham: Should `AppendBlock` be called something more descriptive since we already have an append block? Maybe `AppendBlockWithOverride`? Does this cascade?

erikbaranowski: Updated name, good call. What did you mean by cascade in this context?

mattdurham (approved):
Lgtm

PR #3960, opened 2023-05-23
Convert API unsupported config return warnings

### LINKED ISSUE #3917: Implement a way to warn users if unsupported config is provided
For example, prometheus scrape config supports `nomad_sd_config` but Flow does not support `discovery.nomad`. A warning should be output if `nomad_sd_config` is provided to the converter.

erikbaranowski:
#### PR Description

This prototype implements a way to return warnings for known unsupported config conversion.

Closes #3917 

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [x] Tests updated


converter/converter.go
```diff
@@ -28,10 +29,16 @@ const (
 // because of mismatched functionality, an error is returned with no resulting
 // config. If the conversion completed successfully but generated warnings, an
 // error is returned alongside the resulting config.
-func Convert(in []byte, kind Input) ([]byte, error) {
+func Convert(in []byte, kind Input) ([]byte, diag.Diagnostics) {
```
rfratto, line 32: Personally, I'd prefer creating a new error type for this instead of trying to reuse river/diag; it doesn't feel like a natural fine since there's never any file name or line information.

converter/internal/prometheusconvert/prometheusconvert.go
```diff
@@ -51,7 +61,32 @@ func Convert(in []byte) ([]byte, error) {
...
+	for _, scrapeConfig := range promConfig.ScrapeConfigs {
+		for _, sdConfig := range scrapeConfig.ServiceDiscoveryConfigs {
+			switch sdConfig.(type) {
+			case *nomad.SDConfig:
+				diags.Add(diag.Diagnostic{
+					Severity: diag.SeverityLevelWarn,
+					Message:  "unsupported nomad_sd_config was provided",
+				})
+			default:
+			}
```
rfratto, line 87: I know this is just a prototype, but protip:


```suggestion
			switch sdConfig.(type) {
			default:
				diags.Add(diag.Diagnostic{
					Severity: diag.SeverityLevelWarn,
					Message:  fmt.Sprintf("unsupported %s was provided", sdConfig.Name()),
				})
			}
```

converter/internal/prometheusconvert/prometheusconvert_test.go
```diff
@@ -2,55 +2,73 @@ package prometheusconvert_test
...
 func TestConvert(t *testing.T) {
-	filepath.WalkDir("testdata", func(path string, d fs.DirEntry, _ error) error {
-		if d.IsDir() {
-			return nil
-		}
-
-		if strings.HasSuffix(path, promSuffix) {
-			inputFile := path
-			expectFile := strings.TrimSuffix(path, promSuffix) + flowSuffix
+	tt := []struct {
```
rfratto, line 14: Aw man, I kind of liked that test cases were automatically defined before just by adding new files. 

What do you think about having a separate Test for errors, and then the existing gold master tests by using the before/after files?

You could also do something similar to the parser tests, where it looks for a regex in the source file holding an error message to check against, which would help with creating a bunch of these easily without having to have a giant list of defined test cases.

erikbaranowski: To do this via directory/file alone l propose having a third optional TEST.errors file in the testdata directory that has error messages (or not).

converter/internal/prometheusconvert/prometheusconvert.go
erikbaranowski: I think I may need more granular decision making here since we do implement a number of discovery components and I think this will warn on all of them as suggested. I could do it this way as long as I can use the sdConfig.Name() to differentiate ones we have from ones we don't or leave as is.

rfratto: Sorry, I was noting that you could use `default` for warning against any discovery which isn't supported, and add specific cases for ones that are :)

erikbaranowski: ah yes reverse the logic and auto catch misses. I like that

converter/internal/prometheusconvert/prometheusconvert_test.go
erikbaranowski: I have to switch out the river diags for a new error type so it might look a little different in the end, but I have added support for .errors files and put the file driven testing back.

converter/internal/common/diagnostics.go
erikbaranowski, line 1: I made this implementation as minimal as possible to start. I suspect we will want another helper function or 2 when something starts using the Convert API.

converter/converter.go
```diff
@@ -28,10 +29,13 @@ const (
 // because of mismatched functionality, an error is returned with no resulting
 // config. If the conversion completed successfully but generated warnings, an
 // error is returned alongside the resulting config.
-func Convert(in []byte, kind Input) ([]byte, error) {
+func Convert(in []byte, kind Input) ([]byte, common.Diagnostics) {
```
rfratto, line 32: This doesn't work, because it's a public API exposing an internal type, meaning callers don't have access to it.

The typical way of dealing with this is to either create aliases (see [river.ErrNoConversion](https://github.com/grafana/agent/blob/4b8807a517071556c7fd520d7ddd792fbcd311f7/pkg/river/types.go#L51)) or to create a lower-level type which is used internally, and to convert that to a higher-level public API error type.

erikbaranowski: good call, should be fixed now

converter/internal/prometheusconvert/prometheusconvert.go
```diff
@@ -31,12 +32,17 @@ import (
...
 //	discovery.lightsail
 //	discovery.relabel
-func Convert(in []byte) ([]byte, error) {
+func Convert(in []byte) ([]byte, common.Diagnostics) {
+	var diags common.Diagnostics
+
 	promConfig, err := promconfig.Load(string(in), false, log.NewNopLogger())
 	if err != nil {
-		return nil, err
+		diags.Add(common.SeverityLevelError, err.Error())
```
rfratto, line 40: Can you wrap the error here to make it clear that the error is from reading the file (e.g., before the conversion even starts)

Something like:

```suggestion
		diags.Add(common.SeverityLevelError, fmt.Sprintf("failed to parse Prometheus config: %s", err))
```

converter/internal/common/diagnostics.go
```diff
@@ -0,0 +1,38 @@
+// Package common exposes error types used throughout converter and a method to
+// pretty-print them to the screen.
+package common
```
rfratto, line 3: I think this just needs to be in its own package since you effectively have to alias everything (and you unfortunately can't put it in converter right now because it would be an import cycle).

rfratto (approved):
LGTM % comments

converter/internal/common/diagnostics.go
rfratto: Though IMO it's a code smell that we have to put it in its own package to avoid circular dependencies, I don't think internal package layout should impact the API

PR #3967, opened 2023-05-24
helm: rename pod template's serviceAccount to serviceAccountName

tpaschalis:
#### PR Description
As per the [Kubernetes docs](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#podspec-v1-core), in PodSpec the `serviceAccount` field is a deprecated alias for `serviceAccountName`, which should be used in its place.

This change should be a no-op to people using the Helm chart.

```
serviceAccount (string) | Deprecated - ServiceAccount is a depreciated alias for ServiceAccountName. Deprecated: Use serviceAccountName instead.
```

No issue filed

#### Notes to the Reviewer
Do we want to notify users that this is a no-op change so that they're not worried if they see a diff in their rendered manifests?

#### PR Checklist

- [X] CHANGELOG updated
- [X] Tests updated

### DIFF
--- a/operations/helm/charts/grafana-agent/templates/controllers/_pod.yaml
+++ b/operations/helm/charts/grafana-agent/templates/controllers/_pod.yaml
@@ -10,7 +10,7 @@ metadata:
     {{- toYaml . | nindent 4 }}
     {{- end }}
 spec:
-  serviceAccount: {{ include "grafana-agent.serviceAccountName" . }}
+  serviceAccountName: {{ include "grafana-agent.serviceAccountName" . }}
   {{- with .Values.image.pullSecrets }}
   imagePullSecrets:
     {{- toYaml . | nindent 4 }}
--- a/operations/helm/charts/grafana-agent/CHANGELOG.md
+++ b/operations/helm/charts/grafana-agent/CHANGELOG.md
@@ -16,8 +16,8 @@ Unreleased
 
 - Make the Faro port optional. (@tpaschalis)
 
-- Rename the deprecated `serviceAccount` alias to to `serviceAccountName` in
-  pod template. This is a no-op for existing deployments. (@tpaschalis)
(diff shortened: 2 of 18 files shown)


operations/helm/charts/grafana-agent/templates/controllers/_pod.yaml
```diff
@@ -10,7 +10,7 @@ metadata:
     {{- toYaml . | nindent 4 }}
     {{- end }}
 spec:
-  serviceAccount: {{ include "grafana-agent.serviceAccountName" . }}
+  serviceAccountName: {{ include "grafana-agent.serviceAccountName" . }}
```
rfratto, line 13: Should we check for Kubernetes API version here to choose between serviceAccount/serviceAccountName? 

The Loki Helm chart uses version checks in a few places: https://github.com/grafana/loki/blob/2c57d2e2b91a5fe8efe852350b9ca5971b73ebf3/production/helm/loki/templates/read/statefulset-read.yaml#L20

rfratto: (It depends on when serviceAccountName was first introduced. If it was introduced years ago, we probably don't need to bother to support the old field name)

rfratto (approved):
LGTM % comment

operations/helm/charts/grafana-agent/templates/controllers/_pod.yaml
tpaschalis: I see a reference to the deprecation back in v1.20, which was released on Dec 8, 2020 so I'd say we go ahead?

https://devdocs.io/kubernetes~1.20/workload-resources/pod-v1/index#PodSpec

rfratto: yeah, that's old enough for me, I think it's fine to just use the new field

PR #3975, opened 2023-05-25
update github.com/sirupsen/logrus and github.com/docker/distribution

captncraig:
These were referenced in various (dubious) CVEs. There is no way the vulns specified were exploitable through agent, but updates are fun anyway.


tpaschalis (approved):
👍

tpaschalis (approved):
👍

tpaschalis (approved):
👍

PR #3981, opened 2023-05-25
RFC: Backwards compatibility guarantees

rfratto:
As we approach a 1.0 release, we must discuss what parts of Grafana Agent will be considered semantically versioned and receive backwards compatibility.

This RFC proposes the scope of backwards compatibility once we cut our first major release.  

Note that this does not discuss a timeline for a 1.0 release, only the implications of backwards compatibility for when a 1.0 release is eventually made available. 

Please read in full before commenting; some gut reactions to things earlier in the RFC may be addressed later in the RFC.


docs/rfcs/xxxx-backwards-compatibility.md
```diff
@@ -0,0 +1,176 @@
...
+The following will be protected by backwards compatibility between minor
+releases: 
+
+- **User configuration**, including configuration files and the command-line
+  interface. 
+
+- **Versioned network APIs**, if any versioned APIs are introduced prior to the
+  1.0 release.
+
+- **Telemetry data used in official dashboards**. This means that users will
```
mattdurham, line 46: Can we define official? Is it any dashboard contained inside the Agent Repo or any dashboard used in a grafana blessed mixin?

docs/rfcs/xxxx-backwards-compatibility.md
```diff
@@ -0,0 +1,176 @@
...
+not used as a way to hype a release. If hyping releases is required, there
+should be a version split between the API version of Grafana Agent and a
+project version (such as API v1.5, project version 2023.0).   
+
+### Supporting previous major releases
+
+When a new major release is published, the previous major release should
+continue to receive security and bug fixes for a set amount of time.
+Announcement of a new major release should be coupled with a minimum Long-Term
+Support (LTS) period for the previous major release. For example, we may choose
```
marctc, line 159: Do we want to expand the definition of LTS in terms on number of months or this is out of the scope of this document?

marctc (approved):
LGTM. Nice write up!

docs/rfcs/xxxx-backwards-compatibility.md
```diff
@@ -0,0 +1,176 @@
...
+
+  - It allows us to get in the habit of being deliberate around when something
+    should be consumable downstream, and will help us strengthen our skills
+    with designing good APIs in general. Ideally, this helps us get better at
+    writing Go code _and_ writing Flow components.
+
+  - It will help train us to more easily recognize when any kind of change is
+    breaking. 
+
+  The easiest way to commit to this without any effort is to put everything in
```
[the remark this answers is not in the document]
jcreixell: agreed, if we want to publish certain parts of the agent for usage as a library, we should create deliberate public interfaces that provide stability and isolate internal changes. These interfaces should probably be documented and tested for compatibility. Creating such interfaces should be a deliberate and informed decision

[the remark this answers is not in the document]
ptodev: I also agree. IMO we should make as many things as possible internal, and then advise users to contact us if they have a use case for making something external. At the moment I'm not even sure who imports our code and for what reasons, so there is no point trying to guess - we could just internalise it and let the various use cases surface via GitHub issues and PRs.

docs/rfcs/xxxx-backwards-compatibility.md
```diff
@@ -0,0 +1,176 @@
...
+
+- **Externally importable Go packages**. If a user is importing our code as a
+  dependency, they should be able to upgrade to a new minor release without
+  having to make changes to their code.
+
+  This means that names, signatures, and meanings may not change; a function
+  called Calculate may not be changed from returning the sum of two numbers to
+  returning the difference of two numbers.
+
+  Protecting Go packaged with backwards compatibility is expected to be a
```
ptodev, line 58: Do you mean "Protecting Go packages"?

docs/rfcs/xxxx-backwards-compatibility.md
```diff
@@ -0,0 +1,176 @@
...
+### Exceptions to backwards compatibility 
+
+It's impossible to guarantee that full backwards compatibility is achieved.
+There are some exceptions which may cause a breaking change without a new major
+version:
+
+- Security: a breaking change may be made if a security fix requires making a
+  breaking change. 
+
+- Non-versioned network APIs: Internal network APIs, such as the internal API
```
ptodev, line 97:
```suggestion
- Non-versioned network APIs: internal network APIs, such as the internal API
```

docs/rfcs/xxxx-backwards-compatibility.md
```diff
@@ -0,0 +1,176 @@
...
+
+- Undocumented behavior: relying on undocumented behavior may break between
+  minor releases. 
+
+- Upstream dependencies: part of the public API of Grafana Agent may directly
+  expose the public API of an upstream dependency. In these cases, if an
+  upstream dependency introduces a breaking change, we may be required to make
+  a breaking change to our public API as well.   
+
+- Other telemetry data: Metrics, logs, and traces may change between releases.
```
ptodev, line 109:
```suggestion
- Other telemetry data: metrics, logs, and traces may change between releases.
```

ptodev: Not sure if you intend to have a capital letter after the colon or the other way around.

docs/rfcs/xxxx-backwards-compatibility.md
```diff
@@ -0,0 +1,176 @@
...
+- Set expectations for what is covered by backwards compatibility. 
+- Set expectations for when upgrades to new major versions will be forced.
+
+## Proposal 
+
+Backwards compatibility means that a user can upgrade their version of Grafana
+Agent without needing to make any changes with the way they interact with
+Grafana Agent. 
+
+### Scope of backwards compatibility
```
ptodev, line 35: The main thing which bothers me is whether backwards compatibility guarantees are more effort that it's worth? I wish there was more information about this in the RFC. For example:
* Is there anecdotal evidence from users that they are struggling to upgrade? Or that there could be changes which would prevent them from upgrading?
* Are any user adoption metrics showing that adoption of new versions is lagging behind, and this is something this RFC could help with? E.g. did we notice that in the past when we introduced a big breaking change, adoption of the new version was not as quick as usual?
* Even if there are pain points which compatibility could help with, if we simply list them in the exception category, then our compatibility guarantees still wouldn't help the users. Knowing more about the problem we're solving could help us determine if our exceptions list is too broad.

docs/rfcs/xxxx-backwards-compatibility.md
```diff
@@ -0,0 +1,176 @@
...
+compatibility. 
+
+## Goals 
+
+- Set expectations for what is covered by backwards compatibility. 
+- Set expectations for when upgrades to new major versions will be forced.
+
+## Proposal 
+
+Backwards compatibility means that a user can upgrade their version of Grafana
```
ptodev, line 31: If the Agent is extensible via modules and plugins, then backwards compatibility for the base Agent image would be much easier to achieve. That's more of a v2 or v3 thing, but maybe it is something you'd want to mention?

docs/rfcs/xxxx-backwards-compatibility.md
rfratto: I try to keep it out of scope in the last paragraph:

> The support timeframe for an LTS version is not fixed, and may change between
major releases. For example, version 0.X may receive at least 12 months of LTS,
while version 1.X may receive at least 4 months of LTS. Project maintainers
will need to decide how long to support previous major versions based on the
difficulty for upgrading to the latest major version.

docs/rfcs/xxxx-backwards-compatibility.md
rfratto: That doesn't change the definition of backwards compatibility though: a new major version is required whenever there is a breaking change caused by updating Grafana Agent.

docs/rfcs/xxxx-backwards-compatibility.md
rfratto: I'd like to clarify that this proposal is less about guaranteeing backwards compatibility, and more to scope what backwards compatibility means given that we follow semantic versioning. 

There are a few things that are leading to this:

* We've been asked in the past to release 1.0, since 1.0 gives the impression of stability and maturity. 
* We have always followed semantic versioning, where versions >0.X must be backwards compatible with the same major release version. 
* So, to release 1.0, we need to define the scope of what is in our public API to know when a release 2.0 would be required. 

To your questions: 

We do see in our usage stats that more than half of reporting installations are several versions behind; only about 7% of users are on the latest v0.33 release. 

In my experience, most users typically don't upgrade unless they need something new or need a specific fix. This makes sense, since when the agent is acting perfectly you leave it in the background and don't think about it. This does mean that when users _do_ need to upgrade, they're likely several versions behind. 

Ignoring backwards compatibility guarantees can lead to accumulating a growing backlog of work to do when a user _does_ need to finally upgrade. My concern is that this is a bad user experience: we're either asking them to stay on top of updating software they don't need to update just for the sake of staying as compatible as possible, or risk being blocked on functionality they need by going through a growing checklist of breaking change items to address.

I think we've been told once or twice to not make as many breaking changes as we have, but disregarding that, I'm mainly proposing we build software that I personally would want to use: clear expectations about breaking changes so that most upgrades are as pain-free as possible.

docs/rfcs/xxxx-backwards-compatibility.md
ptodev: True, although it does make it easier for both developers and customers:

- Developers won't have to create major versions as often
- Customers would be able to upgrade the core Agent binary without upgrading the version of plugins used in Agent configs. Moreover, they could mix and match plugins at different versions. Currently they have to use a set bundle of components, all at that version.

I know this might be a bit outside the scope of the RFC, but it could imply that component versions are one day treated separately and thought this might be worth mentioning in a "future" section for example. Up to you, I personally don't insist 😄

docs/rfcs/xxxx-backwards-compatibility.md
```diff
@@ -0,0 +1,179 @@
...
+### Avoiding major release burnout 
+
+As a new major release implies a user must put extra effort into upgrading, it
+is possible to burn out users by releasing breaking changes too frequently. 
+
+We will attempt to limit new major versions no more than once every 12 calendar
+months. This means that if Grafana Agent 1.0 was hypothetically released on
+August 4th, Grafana Agent 2.0 should not be released until at least August 4th
+of the following year. This is best-effort; if a new major release is required
+earlier, then one should be published.
```
thampiotr, line 150: I reckon that we'd have a batch of breaking changes for each major release and they would need to be parked on some branch and would be available to the public anyway. Maybe we should define what 'released' and 'published' mean?

docs/rfcs/xxxx-backwards-compatibility.md
```diff
@@ -0,0 +1,179 @@
...
+  backwards compatibility between major versions. 
+
+If a breaking change is introduced in a minor change accidentally, and that
+breaking change is not covered by one of the [exceptions][] defined below, it
+is a bug. In these cases, a patch release should be introduced to undo the
+breaking change. 
+
+[exceptions]: #exceptions-to-backwards-compatibility
+
+### Exceptions to backwards compatibility
```
thampiotr, line 91: I think we should have an exception for "Experimental" or "Beta" components or features - as long as they are clearly documented as such. 

In the spirit of iterative development, we should have an option to publish an experimental component, gather feedback and later decide to change e.g. the configuration format in order to make it is more user friendly. 

Accidentally introducing a poor config file for a component that is not yet mature shouldn't lock us in for up to 12 months until we can make a breaking change for the next major release.

thampiotr (commented):
Looks good, it's useful to have a doc like this. Left a few comments.

docs/rfcs/xxxx-backwards-compatibility.md
rfratto: I agree with this, though we might want to find a different way to categorize changes to experimental/beta components in the changelog. Calling them "breaking changes" might mislead people to think we should've done a major release. 

Maybe, after 1.0, breaking changes to non-stable components are just considered "other changes?"

docs/rfcs/xxxx-backwards-compatibility.md
rfratto: I've tried to clarify this 👍

docs/rfcs/xxxx-backwards-compatibility.md
rfratto: (I don't think we need to document how we communicate changes to non-stable features here, but it's worth thinking about)

thampiotr: Yeah, a section for just `Experimental` or something similar would work. Sure we can find a good name with some help from tech writers.

docs/rfcs/xxxx-backwards-compatibility.md
```diff
@@ -0,0 +1,111 @@
...
+
+  Protecting Go packages with backwards compatibility is expected to be a contentious proposal. However, it has positive benefits: 
+
+  - It makes us good open-source citizens, allowing people to use Grafana Agent as a library with the same upgrade confidence as people who use Grafana Agent as a binary.  
+
+  - It allows us to get in the habit of being deliberate around when something should be consumable downstream, and will help us strengthen our skills with designing good APIs in general. Ideally, this helps us get better at writing Go code _and_ writing Flow components.
+
+  - It will help train us to more easily recognize when any kind of change is breaking. 
+
+  The easiest way to commit to this without any effort is to put everything in an top-level internal folder, effectively saying that Grafana Agent has no code API that can be consumed downstream.
```
mattdurham, line 48: Not a fan of this easy step. Everything to internal feels less open then breaking having api changes. If someone is using grafana agent as a library and things they need are behind internal then that requires a fork instead of changing their api signatures.

docs/rfcs/xxxx-backwards-compatibility.md
```diff
@@ -0,0 +1,111 @@
...
+
+- **User configuration**, including configuration file syntax and the command-line interface. 
+
+- **Versioned network APIs**, if any versioned APIs are introduced prior to the 1.0 release.
+
+- **Telemetry data used in official dashboards**. This means that users will continue to be able to use the same set of dashboards we provide when upgrading minor releases.    
+
+  - Official dashboards are dashboards in the repository's `operations/` [directory](../../operations/).
+
+- **Externally importable Go packages**. If a user is importing our code as a dependency, they should be able to upgrade to a new minor release without having to make changes to their code.
```
mattdurham, line 36: Instead of using Agent as a library how about instead moving the Agent systems that would make good libraries into their own repositories. Then our usage would be the exact same as people wanting to use the subsystems of the Agent as a library. Then we could semantically version the subsystems independently of each other while maintaining that the Agent itself does not expose a stable api. This would also promote more usable, composable systems instead of tight coupling.

docs/rfcs/xxxx-backwards-compatibility.md
rfratto: > Not a fan of this easy step. Everything to internal feels less open then breaking having api changes. If someone is using grafana agent as a library and things they need are behind internal then that requires a fork instead of changing their api signatures.

That's exactly the intent; having something be exposed should be done with intent that we're ready for it to be consumed publicly and it's been designed in a way that's broadly useful. 

Forking instead of importing isn't always bad, as [A little copying is better than a little dependency](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=9m28s).

docs/rfcs/xxxx-backwards-compatibility.md
rfratto: > My concern is that we will have API changes that are unavoidable which will then trigger major version changes at a rate we do not like.

If we expect we need to have API changes that are unavoidable, we should not be exposing those APIs.

docs/rfcs/xxxx-backwards-compatibility.md
```diff
@@ -0,0 +1,111 @@
...
+
+  The easiest way to commit to this without any effort is to put everything in an top-level internal folder, effectively saying that Grafana Agent has no code API that can be consumed downstream.
+
+- **The scope of backwards compatibility**. Backwards compatibility is only defined for major version 1; we reserve the right to change the definition of backwards compatibility between major versions. 
+
+If a breaking change is introduced in a minor change accidentally, and that breaking change is not covered by one of the [exceptions][] defined below, it is a bug. In these cases, a patch release should be introduced to undo the breaking change. 
+
+[exceptions]: #exceptions-to-backwards-compatibility
+
+### Exceptions to backwards compatibility
```
rfratto, line 56: I'd like to add two new exceptions: 

1. Default values are not subject to backwards compatibility and may change between minor releases. 
2. Deprecating and making something a no-op is not a breaking change because the old configuration file is still accepted.

ptodev: > Default values are not subject to backwards compatibility and may change between minor releases.

I suppose the reasoning is that the previous default value would still be configurable, even though it is not a default.

It would be good if backwards compatibility gives an answer to the question "if I upgrade the Agent, will my old config continue to work in the same way?". If we change default values, the answer to that would be "no", so IMO it'd be better to not consider this backwards compatible.

mattdurham: There is some nuance here on what backward compatibility means, IMO the document seems to indicate though maybe not spelled out that; backward compatibility means "the Agent will run but the behavior may change without changes". Or more simply "the inputs (config/api) won't change but the output of the Agent is not guaranteed".

thampiotr: > Default values are not subject to backwards compatibility and may change between minor releases.

I think default values should be part of the backwards compatibility guarantee. It can completely break you if we e.g. change the default port of some component. When I depend on software and I see it's using semver and its a minor release, I expect it to not break anything and I am happy to quickly apply the update. If the software starts breaking with minor releases, I will lose trust and will not want to update often any more. 

IMO we should make default values part of the guarantee and be very careful about what we set up as default, because it will need to stay this way until the next breaking changes opportunity.

thampiotr: > Deprecating and making something a no-op is not a breaking change ...

IMO making something a no-op is a breaking change, because e.g. my telemetry pipeline may stop working after a minor version update. I think that should not be the case - I'll quickly loose trust in minor updates if this is the case. 

Deprecation is not the same thing as making something a no-op though, we _can_ mark things as deprecated in minor releases - in docs (and maybe print a warning) and remove the feature only in the next major release. [See here](https://semver.org/#how-should-i-handle-deprecating-functionality). 

> not a breaking change because the old configuration file is still accepted

I don't think we should go with this definition of what is in scope of backwards compatibility guarantee. Imagine if k8s made a minor release where `Deployment` resource in your config file would become a no-op. It would be a breaking change. 

For me the TL;DR; should be that if I am using non-experimental, documented features, I should always be able to make a minor or a bugfix update without any extra work - just a version bump and nothing should break. And I don't need to read through changelog etc.

docs/rfcs/xxxx-backwards-compatibility.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Proposal 
+
+Backwards compatibility means that a user can upgrade their version of Grafana Agent without needing to make any changes with the way they interact with Grafana Agent, provided that interaction is within scope of being covered by backwards compatibility.
+
+### Scope of backwards compatibility   
+
+The following will be protected by backwards compatibility between minor releases: 
+
+- **User configuration**, including configuration file syntax and the command-line interface.
```
thampiotr, line 28: Based on discussions in other threads, I think we should also be clear that functional meaning / semantics of the configuration should also be covered - for CLI, API and config files. 

Imagine a change where `loki.write` starts to just print to stdout, like `loki.echo` - that would break users' telemetry pipelines, even though it may not break the config files.

docs/rfcs/xxxx-backwards-compatibility.md
rfratto: > I think default values should be part of the backwards compatibility guarantee. It can completely break you if we e.g. change the default port of some component. 

The rationale behind the default values is to allow us to change things like the default value for the frequency at which the WAL gets truncated, but not things like port numbers. 

Maybe the language should be tweaked here a bit.

thampiotr: > like the default value for the frequency at which the WAL gets truncated

I see - so this would be some knob for which we provide reasonable defaults, but allow customers to change as well if needed. I think for these we could consider one of the following options:
* Describe the default as `auto` and not really document what value it will be. If this changes, it'd be an undocumented change.
* Call out in docs that this configuration property may be tuned with minor releases. That should also be a property that only affects performance, but not the functionality. 
* Have a namespace/prefix for config properties that may change outside of the backwards compatibility guarantees. 

A general statement about defaults does sound like we give ourselves permission to break things like ports etc, so I think we should tweak this.

jcreixell, line 56: shall we include licensing issues with upstream dependencies as an exception?

rfratto: Yeah, I want to change the security section to "Security and legal requirements"

rfratto: I'm going to say that denoting that the "functional semantics" as being covered by backwards compatibility implies that non-functional semantics (i.e., performance tunable settings) are not covered. 

Marking as resolved.

docs/rfcs/xxxx-backwards-compatibility.md
rfratto: That's interesting. It seems like such a case isn't covered by [Go's](https://go.dev/doc/go1compat) backwards compatibility guarantees either: 

> Methods. As with struct fields, it may be necessary to add methods to types. Under some circumstances, such as when the type is embedded in a struct along with another type, the addition of the new method may break the struct by creating a conflict with an existing method of the other embedded type. We cannot protect against this rare case and do not guarantee compatibility should it arise. 

They don't specifically call out the case you lined out, but it's the same. 

I think we should link out to Go's compatibility guarantees for our API compatibility guarantees to say we're a superset of this.

rfratto: I've changed the language to explicitly say that our expectations around backwards compatible API changes match Go's.

docs/rfcs/xxxx-backwards-compatibility.md
```diff
@@ -0,0 +1,104 @@
...
+Backwards compatibility guarantees introduce new challenges while writing code,
+as it means finding creative ways to iterate without making breaking changes
+immediately. Here are some example strategies one could use to avoid making a
+breaking change:
+
+- Keep things internal: the easiest route is to just not expose things publicly. For example, put Go code in `internal/` packages, or use unexported methods or types.
+
+- Rely on deprecations: if a publicly exposed thing must be changed, consider deprecating it.
+
+  - Instead of removing something, consider making it a no-operation and adding a deprecation warning.
```
thampiotr, line 78: I'm not sure about this approach, would prefer if we don't include it here and not use it. 

Maybe I'm misunderstanding something, but it seems to me like we could have e.g. `func SendMetrics(...) error` that is being change to `return nil`. In most cases this would break people's setup and hence would violate the:

> can upgrade their version of Grafana Agent without needing to make any changes with the way they interact with Grafana Agent

thampiotr (approved):
Looks good! I just have a couple of comments, but I wouldn't say these are blocking.

docs/rfcs/xxxx-backwards-compatibility.md
rfratto: I agree, I already removed some recommendations from the RFC, and I think I should remove the recommendation here too.

docs/rfcs/xxxx-backwards-compatibility.md
```diff
@@ -0,0 +1,104 @@
...
+
+- **User configuration**, including the syntax and functional semantics of the configuration file and command-line interface.
+
+- **Versioned network APIs**, if any versioned APIs are introduced prior to the 1.0 release.
+
+- **Telemetry data used in official dashboards**. This means that users will continue to be able to use the same set of dashboards we provide when upgrading minor releases.    
+
+  - Official dashboards are dashboards in the repository's `operations/` [directory](../../operations/).
+
+- **Externally importable Go packages**. If a user is importing our code as a dependency, they should be able to upgrade to a new minor release without having to make changes to their code.
```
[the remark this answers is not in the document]
rfratto: Yeah, I think what we expose publicly with a 1.0 tag is potentially out of scope and needs more discussion.

docs/rfcs/xxxx-backwards-compatibility.md
mattdurham: Sounds reasonable

erikbaranowski (approved):
LGTM, with this context for a future discussion https://github.com/grafana/agent/pull/3981#discussion_r1294794073

PR #3822, opened 2023-05-09
Restore attributes processor

ptodev:
#### PR Description
`otelcol.processor.attributes` was originally added via #3349 but had to be removed via #3589 due to the fact that some functionality wasn't working. This PR fixes the issues which were discovered and adds extra testing so that we run the actual Otel attributes processor and inspect its output.

#### Notes to the Reviewer
There are lots of different combinations of attributes filters and actions, and it's very hard to test all of them. However, I believe the testing done here is sufficient for us to have confidence in the component.

As reviewers I have added the same colleagues who reviewed the original PR - #3349.

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated

PR #3983, opened 2023-05-25
River: Unmarshal integers from interface{} into int64, uint64 or float64

ptodev:
#### PR Description

Currently, if River unmarshals a number into an `interface{}` it will always turn it into a float64.  This is similar to json unmarshaling which always unmarshals numbers to floats. 

However, it is a problem in some cases, when we need to be yaml-compatible. For example, in #3822 we have an `interface{}` value in our River struct which could be an int, float or a string. Depending on which type is used, OpenTelemetry will create an attribute (label) in the signal with this same type. Since River always unmarshals numbers into a float when the variable is of type `interface{}`, all Otel attributes of a numerical type will be floats and none will be integers. 

This change makes River more in line with yaml, whose behaviour could be summarised as:

1. Try parsing the string into an integer
2. If that fails, try parsing into an unsigned
3. If that fails, try parsing into a float

[Here](https://github.com/go-yaml/yaml/blob/v2.4.0/resolve.go#L151) is the relevant yaml code. 

In the River code, we only use "64 bit" types for simplicity.

This PR reverts #3630

#### Notes to the reviewer

Do you think I should go through all the components we currently have and identify the ones which use `interface{}`? We may want to list this as a breaking change for these components, since depending on how the users use them there could be user-facing consequences like the ones described with the attributes processor.

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/river/internal/value/decode.go
+++ b/pkg/river/internal/value/decode.go
@@ -397,15 +397,17 @@ func tryCapsuleConvert(from Value, into reflect.Value, intoType Type) (ok bool, 
 // decodeAny is invoked by decode when into is an interface{}. We assign the
 // interface{} a known type based on the River value being decoded:
 //
-//	Null values:   nil
-//	Number values: float64, int, or uint depending on the underlying Go type
-//	               of the River value
-//	Arrays:        []interface{}
-//	Objects:       map[string]interface{}
-//	Bool:          bool
-//	String:        string
-//	Function:      Passthrough of the underlying function value
-//	Capsule:       Passthrough of the underlying capsule value
+//		Null values:           nil
+//		Integer values:        int64, uint64 or float64, in this order of preference;
+//	                           If the number overflows an int64 then we try uint64,
+//	                           and if that also overflows we try a float64.
+//		Floating point values: float64
+//		Arrays:                []interface{}
+//		Objects:               map[string]interface{}
+//		Bool:                  bool
+//		String:                string
+//		Function:              Passthrough of the underlying function value
+//		Capsule:               Passthrough of the underlying capsule value
 //
 // In the cases where we do not pass through the underlying value, we create a
 // value of that type, recursively call decode to populate that new value, and
@@ -419,8 +421,20 @@ func (d *decoder) decodeAny(val Value, into reflect.Value) error {
 		return nil
 
 	case TypeNumber:
-		var v = val.Number().Float()
-		ptr = reflect.ValueOf(&v)
+
+		switch val.Number().Kind() {
+		case NumberKindFloat:
+			var v float64
+			ptr = reflect.ValueOf(&v)
+		case NumberKindUint:
+			var v uint64
+			ptr = reflect.ValueOf(&v)
+		case NumberKindInt:
+			var v int64
+			ptr = reflect.ValueOf(&v)
+		default:
+			panic("river/value: unreachable")
+		}
 
 	case TypeArray:
 		var v []interface{}
--- a/pkg/river/internal/value/decode.go
+++ b/pkg/river/internal/value/decode.go
@@ -398,16 +398,16 @@ func tryCapsuleConvert(from Value, into reflect.Value, intoType Type) (ok bool, 
 // decodeAny is invoked by decode when into is an interface{}. We assign the
 // interface{} a known type based on the River value being decoded:
 //
-//			Null values:           nil
-//			Number values:         float64, int64, or uint64 depending on the underlying Go type
-//		                           of the River value. If the underlying type is not a float and the
-//	                               number fits within an "int", then an "int" will be used.
-//			Arrays:                []interface{}
-//			Objects:               map[string]interface{}
-//			Bool:                  bool
-//			String:                string
-//			Function:              Passthrough of the underlying function value
-//			Capsule:               Passthrough of the underlying capsule value
+//	Null values:           nil
+//	Number values:         float64, int64, or uint64 depending on the underlying Go type
+//	                       of the River value. If the underlying type is not a float and the
+//	                       number fits within an "int", then an "int" will be used.
+//	Arrays:                []interface{}
+//	Objects:               map[string]interface{}
+//	Bool:                  bool
+//	String:                string
+//	Function:              Passthrough of the underlying function value
+//	Capsule:               Passthrough of the underlying capsule value
 //
 // In the cases where we do not pass through the underlying value, we create a
 // value of that type, recursively call decode to populate that new value, and
--- a/pkg/river/internal/value/decode.go
+++ b/pkg/river/internal/value/decode.go
@@ -399,9 +399,9 @@ func tryCapsuleConvert(from Value, into reflect.Value, intoType Type) (ok bool, 
 // interface{} a known type based on the River value being decoded:
 //
 //	Null values:           nil
-//	Number values:         float64, int64, or uint64 depending on the underlying Go type
-//	                       of the River value. If the underlying type is not a float and the
-//	                       number fits within an "int", then an "int" will be used.
+//	Number values:         float64, int, int64, or uint64.
+//	                       If the underlying type is a float, always decode to a float64.
+//	                       For non-floats the order of preference is int -> int64 -> uint64.
 //	Arrays:                []interface{}
 //	Objects:               map[string]interface{}
 //	Bool:                  bool
@@ -430,6 +430,9 @@ func (d *decoder) decodeAny(val Value, into reflect.Value) error {
 			uint64Val := val.Uint()
 			if uint64Val <= math.MaxInt {
 				var v int
+				ptr = reflect.ValueOf(&v)
+			} else if uint64Val <= math.MaxInt64 {
+				var v int64
 				ptr = reflect.ValueOf(&v)
 			} else {
 				var v uint64
--- a/pkg/river/internal/value/decode.go
+++ b/pkg/river/internal/value/decode.go
@@ -4,6 +4,7 @@ import (
 	"encoding"
 	"errors"
 	"fmt"
+	"math"
 	"reflect"
 	"time"
 
@@ -397,17 +398,16 @@ func tryCapsuleConvert(from Value, into reflect.Value, intoType Type) (ok bool, 
 // decodeAny is invoked by decode when into is an interface{}. We assign the
 // interface{} a known type based on the River value being decoded:
 //
-//		Null values:           nil
-//		Integer values:        int64, uint64 or float64, in this order of preference;
-//	                           If the number overflows an int64 then we try uint64,
-//	                           and if that also overflows we try a float64.
-//		Floating point values: float64
-//		Arrays:                []interface{}
-//		Objects:               map[string]interface{}
-//		Bool:                  bool
-//		String:                string
-//		Function:              Passthrough of the underlying function value
-//		Capsule:               Passthrough of the underlying capsule value
+//			Null values:           nil
+//			Number values:         float64, int64, or uint64 depending on the underlying Go type
+//		                           of the River value. If the underlying type is not a float and the
+//	                               number fits within an "int", then an "int" will be used.
+//			Arrays:                []interface{}
+//			Objects:               map[string]interface{}
+//			Bool:                  bool
+//			String:                string
+//			Function:              Passthrough of the underlying function value
+//			Capsule:               Passthrough of the underlying capsule value
 //
 // In the cases where we do not pass through the underlying value, we create a
 // value of that type, recursively call decode to populate that new value, and
@@ -427,11 +427,24 @@ func (d *decoder) decodeAny(val Value, into reflect.Value) error {
 			var v float64
 			ptr = reflect.ValueOf(&v)
 		case NumberKindUint:
-			var v uint64
-			ptr = reflect.ValueOf(&v)
+			uint64Val := val.Uint()
+			if uint64Val <= math.MaxInt {
+				var v int
+				ptr = reflect.ValueOf(&v)
+			} else {
+				var v uint64
+				ptr = reflect.ValueOf(&v)
+			}
 		case NumberKindInt:
-			var v int64
-			ptr = reflect.ValueOf(&v)
+			int64Val := val.Int()
+			if math.MinInt <= int64Val && int64Val <= math.MaxInt {
+				var v int
+				ptr = reflect.ValueOf(&v)
+			} else {
+				var v int64
+				ptr = reflect.ValueOf(&v)
+			}
+
 		default:
 			panic("river/value: unreachable")
 		}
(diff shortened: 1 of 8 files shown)


pkg/river/vm/constant.go
```diff
@@ -18,11 +18,22 @@ func valueFromLiteral(lit string, tok token.Token) (value.Value, error) {
...
+			return value.Uint(uintVal), nil
 		}
-		return value.Int(v), nil
+
+		floatVal, err3 := strconv.ParseFloat(lit, 64)
+		if err3 == nil {
+			return value.Float(floatVal), nil
+		}
+
+		return value.Null, err3
```
rfratto, line 36: Can you write a benchmark for this and provide before/after results? This may make decoding floating point numbers significantly slower.

pkg/river/internal/value/decode.go
```diff
@@ -364,15 +364,17 @@ func tryCapsuleConvert(from Value, into reflect.Value, intoType Type) (ok bool,
...
-//	Arrays:        []interface{}
-//	Objects:       map[string]interface{}
-//	Bool:          bool
-//	String:        string
-//	Function:      Passthrough of the underlying function value
-//	Capsule:       Passthrough of the underlying capsule value
+//		Null values:           nil
+//		Integer values:        int64, uint64 or float64, in this order of preference;
+//	                           If the number overflows an int64 then we try uint64,
+//	                           and if that also overflows we try a float64.
```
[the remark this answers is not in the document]
rfratto: I think your implementation is correct, and just the comment is wrong; I do believe that the VM should be handling what type to use, and the decoder should just assign directly based on that.

mattdurham:
Lurking here but I +1 this =)

rfratto:
> In the River code, we only use "64 bit" types for simplicity.

I know I suggested that we do this on Slack, but I think I'd like to backpedal from that because an untyped number constant in Go will default to `int`, so as it is now we always have to typecast numbers to int64. 

Maybe the order should be int -> int64 -> uint64 -> float64?

pkg/river/vm/constant.go
thampiotr: What triggers calling this codepath? I wonder how much of a hot path this is.

thampiotr (commented):
Happy with this change, no major comments.

pkg/river/vm/constant.go
rfratto: It's any time a number is in the config file and a component is evaluated, so it can happen multiple times a second

ptodev:
> > In the River code, we only use "64 bit" types for simplicity.
> 
> I know I suggested that we do this on Slack, but I think I'd like to backpedal from that because an untyped number constant in Go will default to `int`, so as it is now we always have to typecast numbers to int64 in tests, which isn't very ergonomic.
> 
> Maybe the order should be int -> int64 -> uint64 -> float64?

Sounds fine. I don't have a strong opinion so I'm happy to go with what you suggest. These are all my thoughts:

- Returning int64 might be a bit faster than int for the parser since [strconv.ParseInt](https://pkg.go.dev/strconv#ParseInt) returns an int64. To return an int we'd have to do an extra check and cast to an int.
- Returning an int would be more in line with [yaml](https://github.com/go-yaml/yaml/blob/v2.4.0/resolve.go#L153). I think this is also why we never get `int64` on the Go [playground](https://go.dev/play/p/CCAfq1miPAt). I guess the playground runs on 64 bit and it'd only return an explicit int64 if it was running on 32 a bit system.
- Returning an int would be more efficient for the very, very few users who are still on 32 bit systems...

If returning an int doesn't cost too much in the benchmarks compared to returning an int64, I'd be happy to go with it. I'll try to compare those but I suspect there won't be much difference.

rfratto (commented):
@ptodev Can we pair on this tomorrow so we're on the same page?

pkg/river/internal/value/number_value.go
```diff
@@ -15,8 +15,10 @@ var (
 type NumberKind uint8
 
 const (
-	// NumberKindInt represents an int-like type (e.g., int, int8, etc.).
+	// NumberKindInt represents an int-like type (e.g., int, int8, etc.) except an int64.
 	NumberKindInt NumberKind = iota
+	// NumberKindInt64 represents an int64.
+	NumberKindInt64
```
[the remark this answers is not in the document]
ptodev: I was thinking the same, but considering that we don't want the decoder to do anything fancy ([this](https://github.com/grafana/agent/pull/3983#discussion_r1205841216) comment) I thought that's the only way to pass the int64 information to the decoder.

pkg/river/internal/value/decode_test.go
```diff
@@ -650,31 +651,32 @@ func TestDecode_KnownTypes_Any(t *testing.T) {
...
+		{int(15), int(15)},
+		{int8(15), int(15)},
+		{int16(15), int(15)},
+		{int32(15), int(15)},
+		{int64(15), int64(15)},
+		{uint(15), uint64(15)},
+		{uint8(15), uint64(15)},
+		{uint16(15), uint64(15)},
+		{uint32(15), uint64(15)},
+		{uint64(15), uint64(15)},
```
rfratto, line 663: All of these should fit within an `int`, so I'm not sure this is matching the behavior we were looking for.

pkg/river/internal/value/number_value.go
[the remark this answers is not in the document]
rfratto: Given our discussion, I think we should revert this change and split the concerns for typing across both the evaluator (river/vm) and the decoder (river/internal/value).

pkg/river/internal/value/value.go
```diff
@@ -49,8 +49,11 @@ var Null = Value{}
 // Uint returns a Value from a uint64.
 func Uint(u uint64) Value { return Value{rv: reflect.ValueOf(u), ty: TypeNumber} }
 
-// Int returns a Value from an int64.
-func Int(i int64) Value { return Value{rv: reflect.ValueOf(i), ty: TypeNumber} }
+// Int returns a Value from an int.
+func Int(i int) Value { return Value{rv: reflect.ValueOf(i), ty: TypeNumber} }
+
+// Int64 returns a Value from an int64.
+func Int64(i int64) Value { return Value{rv: reflect.ValueOf(i), ty: TypeNumber} }
```
rfratto, line 56: Ditto for removing this; the River representation of numbers will always be either an int64/uint64/float64, even though the decoding of those numbers may downcast when possible.

pkg/river/vm/constant.go
```diff
@@ -18,11 +18,27 @@ func valueFromLiteral(lit string, tok token.Token) (value.Value, error) {
...
-		if err != nil {
-			return value.Null, err
+		int64Val, err1 := strconv.ParseInt(lit, 0, 64)
+		if err1 == nil {
+			intVal := int(int64Val)
+			if int64Val == int64(intVal) {
+				return value.Int(intVal), nil
+			} else {
+				return value.Int64(int64Val), nil
+			}
```
rfratto, line 28: Since we're always going to change the River representation to be either an int64/uint64/float64, you can remove this conditional here to speed things up a little.

pkg/river/vm/constant.go
rfratto: Paulin and I wrote benchmarks together, but it's not possible to compare to the existing code since the existing code forced number literals to fit within an int64. 

The new code is definitely slower in the worst case (float64), where it has to parse the same string three times, but the worst case is still roughly 415ns/op on my machine, which is roughly twice as slow as the best case in main (239ns/op). 

I don't think users are usually going to have number literals which are so big that they fall into the worse case, so I think this change is roughly safe to make without much performance impact; if we do find it to be a performance bottleneck, we can revisit how the numbers are parsed to optimize it.

pkg/river/internal/value/decode.go
```diff
@@ -386,8 +388,23 @@ func (d *decoder) decodeAny(val Value, into reflect.Value) error {
...
+		switch val.Number().Kind() {
+		case NumberKindFloat:
+			var v float64
+			ptr = reflect.ValueOf(&v)
+		case NumberKindUint:
+			var v uint64
+			ptr = reflect.ValueOf(&v)
+		case NumberKindInt:
+			var v int
+			ptr = reflect.ValueOf(&v)
```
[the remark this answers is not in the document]
rfratto: Just an aside: I don't think float64s should ever be downcast, just int64/uint64 to int whenever possible.

ptodev:
> Maybe the order should be int -> int64 -> uint64 -> float64?

Hi @rfratto , thanks very much for the feedback! I updated the code now. There's just one thing - in theory this order won't be honoured on a 32 bit system for an integer between MaxInt and MaxUint64. With my current code we'll end up. with a uint64 instead of an int64. I'd have to add an extra clause to the decoder to fix that. But I feel like it's an overkill at this point?

mattdurham:
At the moment I dont think we support any 32bit architectures.

pkg/river/internal/value/decode.go
```diff
@@ -397,15 +398,16 @@ func tryCapsuleConvert(from Value, into reflect.Value, intoType Type) (ok bool,
...
-//	Arrays:        []interface{}
-//	Objects:       map[string]interface{}
-//	Bool:          bool
-//	String:        string
-//	Function:      Passthrough of the underlying function value
-//	Capsule:       Passthrough of the underlying capsule value
+//			Null values:           nil
+//			Number values:         float64, int64, or uint64 depending on the underlying Go type
+//		                           of the River value. If the underlying type is not a float and the
+//	                               number fits within an "int", then an "int" will be used.
```
rfratto, line 404: The alignment got a little messed up here:

```suggestion
//			Number values:         float64, int64, or uint64 depending on the underlying Go type
//			                       of the River value. If the underlying type is not a float and 
//			                       the number fits within an "int", then an "int" will be
//			                       used.
```

pkg/river/internal/value/decode_test.go
```diff
@@ -663,31 +664,42 @@ func TestDecode_KnownTypes_Any(t *testing.T) {
...
+		{float64(math.MinInt64) - 10, float64(math.MinInt64) - 10},
+		{float64(math.MaxInt64) + 10, float64(math.MaxInt64) + 10},
 
 		{bool(true), bool(true)},
 		{string("Hello"), string("Hello")},
 
 		{
 			input:  []int{1, 2, 3},
-			expect: []any{float64(1), float64(2), float64(3)},
+			expect: []any{int(1), int(2), int(3)},
```
rfratto, line 697: I don't think the wrapping is needed here anymore, since I'm pretty sure untyped whole number constants default to `int`:

```suggestion
			expect: []any{1, 2, 3},
```

This change can be replicated in all other files where float64(N) is changed to int(N); it should be possible to remove the explicit type cast, which is half the reason to support `int` anyway.

pkg/river/vm/vm_benchmarks_test.go
```diff
@@ -92,7 +99,7 @@ func BenchmarkExprs(b *testing.B) {
 
 			for i := 0; i < b.N; i++ {
 				vPtr := reflect.New(expectType).Interface()
-				_ = eval.Evaluate(scope, vPtr)
+				require.NoError(b, eval.Evaluate(scope, vPtr))
```
rfratto, line 102: Note that adding in require.NoError into the benchmark makes the test slower and skews the results, which is why errors were being ignored before.

rfratto (commented):
Nearly there! This should be the last round of feedback, nice work :)

rfratto:
> Hi @rfratto , thanks very much for the feedback! I updated the code now. There's just one thing - in theory this order won't be honoured on a 32 bit system for an unsigned integer between MaxInt and MaxInt64. With my current code we'll end up. with a uint64 instead of an int64. I'd have to add an extra clause to the decoder to fix that. But I feel like it's an overkill at this point?

Even though we don't have any 32-bit architecture support right now, that sounds like it would be a really difficult bug to track down if we added it in the future. IMO given that it's just a small extra conditional, it'd be a good idea to add it for future portability, especially in the case where river/flow get brought out into their own repo.

pkg/river/vm/vm_benchmarks_test.go
ptodev: This is also what I thought, but I added the check because I doubted that it would slow down the code much. I ran a few benchmarks just now and wow, it does slow it down noticeably. 

That said, I think it wouldn't slow it down too much if we add a more lightweight check like an `err != nil` + a `b.Fatalf` and would be worth the tradeoff given that otherwise we could easily miss an error.

These are the variants which I tested:

[no_error_check.txt](https://github.com/grafana/agent/files/11706930/no_error_check.txt) (baseline):
```
_ = eval.Evaluate(scope, vPtr)
```

[require_no_error.txt](https://github.com/grafana/agent/files/11706931/require_no_error.txt):
```
require.NoError(b, eval.Evaluate(scope, vPtr))
```

[simple_error_check.txt](https://github.com/grafana/agent/files/11706932/simple_error_check.txt):
```
err = eval.Evaluate(scope, vPtr)
if err != nil {
	b.Fatalf("evaluation failed with error: %q", err)
}
```

[no_checks_even_for_parsing.txt](https://github.com/grafana/agent/files/11706929/no_checks_even_for_parsing.txt):
```
expr, _ := parser.ParseExpression(tc.input)
```
...and...
```
_ = eval.Evaluate(scope, vPtr)
```


The benchmark comparisons are:

* [no_error_check_VS_require_no_error.txt](https://github.com/grafana/agent/files/11706976/no_error_check_VS_require_no_error.txt): +45.70% geomean in sec/op

* [no_error_check_VS_simple_error_check.txt](https://github.com/grafana/agent/files/11706977/no_error_check_VS_simple_error_check.txt): +0.48% geomean in sec/op

* [no_error_check_VS_no_checks_even_for_parsing.txt](https://github.com/grafana/agent/files/11706975/no_error_check_VS_no_checks_even_for_parsing.txt): -2.48% geomean in sec/op

Based on this, IMO we could add such a simple check , and also replace the check for `parser.ParseExpression` with a simple check too. If we do both, probably it'd be a net performance increase even :) I haven't tested it though

rfratto: I'm just not sure we need our benchmarks to also be testing the code TBH, just executing it.

ptodev: For now I'm happy to just the error check I added, but if you want me to replace both with the simple checks just let me know.

ptodev: Ohh I just realised there is a `b.StopTimer()` and `b.StartTimer()`. Maybe we can just wrap the error check in them 😅

rfratto: I think that's probably overthinking it still (also stopping and starting the timer is slow too IME). Like I said, benchmarks really don't need to test that code is correct, they're just trying to measure the performance of code.

ptodev:
I think it's all ready for merging now. Thanks very much for the comments! Please feel free to take a final look.

PR #3990, opened 2023-05-25
Update Docker Agent Static install

clayton-cornell:
#### PR Description

Update the Grafana Agent install on Docker topic

- Clean up the makrdown used
- Align better with Agent Flow content/layout
- Fix notes
- Fix TOC entry

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/static/set-up/install-agent-docker.md
+++ b/docs/sources/static/set-up/install-agent-docker.md
@@ -1,61 +1,60 @@
 ---
-title: Run static mode on Docker
+title: Docker
 weight: 110
 aliases:
 - ../../set-up/install-agent-docker/
 ---
 
-## Run static mode on Docker
+## Install Grafana Agent on Docker
 
-Install Grafana Agent and get it up and running on Docker.
+Grafana Agent Flow is available as Docker images on the following platforms:
+
+* [Linux containers][] for AMD64 and ARM64 machines.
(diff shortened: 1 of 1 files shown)


docs/sources/static/set-up/install-agent-docker.md
```diff
@@ -1,61 +1,62 @@
...
-```
+ - Ensure that [Docker][] is installed and running on your machine.
+ - Ensure that you have created a Grafana Agent [configuration file]({{< relref "../configuration/create-config-file/" >}}).
 
-2. Replace `/tmp/agent` with the folder you want to store WAL data in.
+[Docker]: https://docker.io
+## Run a Linux Docker container
 
-    WAL data is where metrics are stored before they are sent to Prometheus. Old WAL data is cleaned up every hour and is used for recovery if the process happens to crash.
+1. To install a Grafana Agent Docker container on Linux, run the following command in a terminal:
```
rfratto, line 26: nit about the terminology here: I think "installing on Docker" is misleading, because it might sound like we're going to be adding Grafana Agent to an existing Docker container.

I think "running on Docker" is more correct for this specific platform.

rfratto (approved):
LGTM % the nit about "install" vs "run" in the specific case of Docker containers.

PR #3991, opened 2023-05-26
Flow: auto-reload on config file changes

### LINKED ISSUE #2691: Flow: add automatically watching & reloading config file
It's typical to deploy Grafana Agent on Kubernetes with a sidecar container to watch for changes to the config file and automatically invoke the `/-/reload` endpoint when it changes.

It'd be great to incorporate this directly into the agent, which would be useful on platforms that don't have the sidecar container (e.g., Raspberry Pi). 

I imagine auto-reloading would be enabled by default, but would support being disabled by a flag.

joshuapare:
#### PR Description
Adds the feature to automatically reload the config when it changes. This feature is enabled by default, and is controlled with the flag `--config.auto-reload`

Fixes #2691 

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/cmd/internal/flowmode/cmd_run.go
+++ b/cmd/internal/flowmode/cmd_run.go
@@ -11,6 +11,7 @@ import (
 	"path"
 	"sync"
 	"syscall"
+	"time"
 
 	"github.com/grafana/agent/component"
 	"github.com/grafana/agent/converter"
@@ -33,6 +34,7 @@ import (
 	"github.com/grafana/agent/pkg/flow/tracing"
 	"github.com/grafana/agent/pkg/river/diag"
 	"github.com/grafana/agent/pkg/usagestats"
+	"github.com/grafana/agent/pkg/util"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/spf13/cobra"
@@ -44,13 +46,15 @@ import (
 
 func runCommand() *cobra.Command {
 	r := &flowRun{
-		inMemoryAddr:     "agent.internal:12345",
-		httpListenAddr:   "127.0.0.1:12345",
-		storagePath:      "data-agent/",
-		uiPrefix:         "/",
-		disableReporting: false,
-		enablePprof:      true,
-		configFormat:     "flow",
+		inMemoryAddr:       "agent.internal:12345",
+		httpListenAddr:     "127.0.0.1:12345",
+		storagePath:        "data-agent/",
+		uiPrefix:           "/",
+		disableReporting:   false,
+		enablePprof:        true,
+		configFormat:       "flow",
+		configAutoReload:   true,
+		configPollInterval: 3 * time.Second,
 	}
 
 	cmd := &cobra.Command{
@@ -88,6 +92,10 @@ depending on the nature of the reload error.
 	}
 
 	cmd.Flags().
+		BoolVar(&r.configAutoReload, "config.auto-reload", r.configAutoReload, "Automatically reload the config file when it changes on disk")
+	cmd.Flags().
+		DurationVar(&r.configPollInterval, "config.auto-reload.poll-interval", r.configPollInterval, "Interval at which to poll the config file for changes when polling is used")
+	cmd.Flags().
 		StringVar(&r.httpListenAddr, "server.http.listen-addr", r.httpListenAddr, "Address to listen for HTTP traffic on")
 	cmd.Flags().StringVar(&r.inMemoryAddr, "server.http.memory-addr", r.inMemoryAddr, "Address to listen for in-memory HTTP traffic on. Change if it collides with a real address")
 	cmd.Flags().StringVar(&r.storagePath, "storage.path", r.storagePath, "Base directory where components can store data")
@@ -122,6 +130,8 @@ type flowRun struct {
 	clusterJoinAddr                string
 	configFormat                   string
 	configBypassConversionWarnings bool
+	configAutoReload               bool
+	configPollInterval             time.Duration
 }
 
 func (fr *flowRun) Run(configFile string) error {
@@ -344,6 +354,17 @@ func (fr *flowRun) Run(configFile string) error {
 		return err
 	}
 
+	// Setup config file watching
+	if fr.configAutoReload {
+		fw := util.NewFileWatcher(configFile, reload, fr.configPollInterval)
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			fw.Watch(l, ctx)
+		}()
+	}
+
+	// Reload on SIGHUP
 	reloadSignal := make(chan os.Signal, 1)
 	signal.Notify(reloadSignal, syscall.SIGHUP)
 	defer signal.Stop(reloadSignal)
--- a/pkg/util/filewatch.go
+++ b/pkg/util/filewatch.go
@@ -1,0 +1,77 @@
+package util
+
+import (
+	"context"
+	"os"
+	"time"
+
+	"github.com/fsnotify/fsnotify"
+	"github.com/go-kit/log/level"
+	"github.com/grafana/agent/pkg/flow/logging"
+)
+
+type ReloadFunc func() error
+
+type FileWatcher struct {
+	filePath     string
+	reloadFunc   ReloadFunc
+	pollInterval time.Duration
+}
+
+// NewFileWatcher creates a new file watcher that will call reloadFunc when the file at filePath is modified.
+func NewFileWatcher(filePath string, reloadFunc ReloadFunc, pollInterval time.Duration) *FileWatcher {
+	return &FileWatcher{
+		filePath:     filePath,
+		reloadFunc:   reloadFunc,
+		pollInterval: pollInterval,
+	}
+}
+
+// Watch starts watching the file for changes and calls reloadFunc when the file is modified.
+func (fw *FileWatcher) Watch(l *logging.Logger, ctx context.Context) {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		level.Error(l).Log("msg", "failed to create watcher", "err", err)
+		return
+	}
+	defer watcher.Close()
+
+	err = watcher.Add(fw.filePath)
+	if err != nil {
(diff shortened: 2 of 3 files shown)


clayton-cornell (approved):
Minor rephrase suggestion.
Approved from a docs standpoint.

sbhrule15:
just realized after looking back at this this morning that this would continue to attempt to reload the config even if it recieved a sigint and sigkill, which could be undesireable. Going to add in those signal handlers to stop the file watch.

cmd/internal/flowmode/cmd_run.go
```diff
@@ -83,6 +85,8 @@ depending on the nature of the reload error.
 		},
 	}
 
+	cmd.Flags().
+		BoolVar(&r.configAutoReload, "config.auto-reload", r.configAutoReload, "Automatically reload the config file when it changes on disk")
```
rfratto, line 89: Some config files might not be hosted on a volume where file change events would be emitted (e.g., NFS). 

WDYT about also having a `--config.auto-reload.poll-interval` flag for polling for file changes in addition to the fsnotify watcher?

cmd/internal/flowmode/cmd_run.go
```diff
@@ -334,6 +339,48 @@ func (fr *flowRun) Run(configFile string) error {
...
+			level.Error(l).Log("msg", "failed to reload config", "err", err)
+		}
+		defer watcher.Close()
+
+		// Stop reloading on SIGINT and SIGTERM
+		sigCh := make(chan os.Signal, 1)
+		signal.Notify(sigCh, syscall.SIGINT, syscall.SIGTERM)
+		defer signal.Stop(sigCh)
+
+		go func() {
```
rfratto, line 355: Can you synchronize this goroutine with the waitgroup so we ensure it closes before Flow exits?

```suggestion
		wg.Add(1)
		go func() {
			defer wg.Done()
```

rfratto (commented):
Looks great! Just one comment and a question about a polling fallback.

cmd/internal/flowmode/cmd_run.go
sbhrule15: Oh that's a great point - we have some NFS and FUSE systems that definitely wouldn't work here. I have a few projects I can test this with :)

Are you thinking auto-fallback to polling with a default interval, or disabling auto-reload if the filesystem doesn't support inotify events?

rfratto: > Are you thinking auto-fallback to polling with a default interval, or disabling auto-reload if the filesystem doesn't support inotify events?

Personally, I'm thinking that it's always both polling and watching for events when auto-reload is enabled, but fsnotify would potentially catch changes to files faster if the filesystem supports events.

Something like:

```go 
for {
  select {
  case <-poll.C: // re-read file based on poll 
  case event := <-watcher.Events: 
    // ... 

  // (other cases as before)
  }
}
```

mattdurham: IMO for config reloading I would be fine with just polling since its simple, should almost always work and given we are polling for a single file (or folder for future) it shouldnt be a ton of overhead.

sbhrule15: I'll be honest, having the instant reload there is quite nice just from developing this and testing with some test configs. It looks like the default sync period on the `discovery.file` component is 10 seconds, which from personal experience, I would probably end up just end up reloading the service, especially if a misconfig brought down the agent by mistake.

@mattdurham 100% understand where you're coming from, but I'd like to do both methods - I suspect other DevOps engineers and devs running local compose agents debugging configs would appreciate the instant reload (and honestly want to work it in just for the fun of it, I'm still relatively new to Go and have only done work in Kubebuilder and Fiber and this is a joy to work on)

rfratto: I personally understand both sides of the debate here: it's nice to have near-instantaneous feedback for a great local editing loop, but using fsnotify adds some extra complexity, and may make PRs like #3855 harder to implement. 

Given file notifications aren't available on all filesystems, and fsnotify events may be dropped (which is why Promtail doesn't use them), polling is definitely required to cast the largest net, but I'm not too opinionated about whether fsnotify should runs on top of that as a convenience. 

@mattdurham Since the fsnotify code is already here, should we keep in t for now? We can always remove it in the future without a breaking change, since `--config.auto-reload` doesn't necessarily imply how auto-reload is implemented.

cmd/internal/flowmode/cmd_run.go
```diff
@@ -334,6 +339,50 @@ func (fr *flowRun) Run(configFile string) error {
 		return err
 	}
 
+	// Reload on file change
+	if fr.configAutoReload {
```
mattdurham, line 343: I wonder if this code block is getting to long and deeply nested and if we should move it to its own func.

cmd/internal/flowmode/cmd_run.go
mattdurham: I would like to see the polling method additionally added since this will default to true but fine with having that be the backup to allow the ideal case. I want to avoid the case where someone on a system that doesn't support inotify wonders why the auto-reload isn't working.

rfratto:
@sbhrule15 Just checking in, do you need help with anything for adding in the polling code?

cmd/internal/flowmode/cmd_run.go
sbhrule15: I agree, especially if polling is going to be added. That'll come in with the change

sbhrule15:
@rfratto all good for now, thanks! currently working through it

sbhrule15:
@rfratto spoke too soon on that, but actually around testing, hoping you could give me some guidance here on how to go about the test suite. I'm having some trouble figuring out how I'm going to simulate the filesystem in test to make sure the fallback works when events aren't supported. Any ideas?

Also was hoping to just get your general opinions on the refactor!

mattdurham:
You might be able to re-use the pattern in `component/local/file/file_test.go`.

clayton-cornell (commented):
Docs look OK.

clayton-cornell (commented):
Docs look OK.

cmd/internal/flowmode/cmd_run.go
```diff
@@ -83,6 +88,12 @@ depending on the nature of the reload error.
 		},
 	}
 
+	cmd.Flags().
+		BoolVar(&r.configAutoReload, "config.auto-reload", r.configAutoReload, "Automatically reload the config file when it changes on disk")
+	cmd.Flags().
+		BoolVar(&r.configForcePolling, "config.force-polling", r.configForcePolling, "Default to using only polling instead of filesystem events for auto-reloading")
+	cmd.Flags().
+		DurationVar(&r.configPollInterval, "config.poll-interval", r.configPollInterval, "Interval at which to poll the config file for changes when polling is used")
```
rfratto, line 96: I'm not sure about the design of these flags at the moment. 

I don't understand the value of `--config.force-polling`; from my perspective it seems like users would always want fsnotify + polling enabled at the same time, or neither. Do you have a use case in mind? If not, I'd suggest taking this flag out to keep the API surface smaller until we know we need it. 

I'd also suggest having the auto-reload and polling interval to be in the same flag namespace so they appear together in the documentation for easy discovery:

* `--config.auto-reload` to toggle whether auto-reloading is enabled (fsnotify + polling)
* `--config.auto-reload.poll-interval` to customize the frequency at which auto-reloading manually polls the file for changes. 

Does this make sense? What do you think?

cmd/internal/flowmode/cmd_run.go
```diff
@@ -334,6 +348,13 @@ func (fr *flowRun) Run(configFile string) error {
 		return err
 	}
 
+	// Setup config file watching
+	if fr.configAutoReload {
+		fw := util.NewFileWatcher(configFile, reload, fr.configPollInterval, fr.configForcePolling)
+		fw.Watch(l, ctx, &wg)
```
rfratto, line 354: Passing the WaitGroup to watch is leaking the implementation details of the caller to the shared function, which is a bit of a smell IMO. 

I'd suggest to change the details here a bit:

* Make `fw.Watch` a blocking call. 
* Call `fw.Watch` from a goroutine launched by the caller:

  ```go 
  wg.Add(1) 
  go func() {
    defer wg.Done()
    fw.Watch(l, ctx) 
  }()
  ```

This will allow util.NewFileWatcher to be used in scenarios where a WaitGroup isn't necessary.

pkg/util/filewatch.go
```diff
@@ -0,0 +1,98 @@
...
+
+	err = watcher.Add(fw.filePath)
+	if err != nil {
+		level.Error(l).Log("msg", "failed to add file to watcher", "err", err)
+		return
+	}
+
+	sigCh := make(chan os.Signal, 1)
+	signal.Notify(sigCh, syscall.SIGINT, syscall.SIGTERM)
+	defer signal.Stop(sigCh)
```
rfratto, line 59: These signals are already handled by the caller so you don't need to watch for them here; instead you can just wait until the provided context gets canceled to exit.

rfratto (commented):
Looking good, just a few final comments!

rfratto:
> @rfratto spoke too soon on that, but actually around testing, hoping you could give me some guidance here on how to go about the test suite. I'm having some trouble figuring out how I'm going to simulate the filesystem in test to make sure the fallback works when events aren't supported. Any ideas?

Honestly, I'm not too worried about the testing; the code is small enough that it can be verified by eye. We can add unit tests later if it turns out there's problems here.

cmd/internal/flowmode/cmd_run.go
sbhrule15: Great point - I'll correct

sbhrule15:
@rfratto sorry for the long wait! In the process of rebasing from main and squashing the commits I accidentally synced my fork from main. Could the PR be reopened to keep the context of the conversations above?

rfratto:
@sbhrule15 Done :)

korniltsev:
@joshuapare are you going to continue working on this PR/issue?

rfratto:
Hi @joshuapare, I'm going to close the PR since it appears to be abandoned. If you want us to reopen it let us know, but for now I'd prefer to close it just in case others want to pick up this work.

Issue #3646, opened 2023-04-26
`loki.source.heroku` update can break the metrics endpoint


thampiotr:
When `loki.source.heroku` is `Update`'d and the update requires restarting the embedded server, the Agent's `/metrics` endpoint can break with `<metric> was collected before with the same name and label values` errors.

Steps to reproduce:

1. Start agent with this config:
```
loki.relabel "test_relabel" {
    forward_to = []
    rule {
        action = "labeldrop"
        regex  = "rubbish"
    }
}

loki.source.heroku "repro" {
    listener {
        address = "localhost"
        port = "8888"
    }
    relabel_rules = loki.relabel.test_relabel.rules
    forward_to = []
}
```

2. Go to `http://localhost:12345/metrics` and observe metrics are as expected.
3. Change the config's relabel rules, e.g. change the regex attribute in `loki.relabel.test_relabel` rule.
4. `curl http://localhost:12345/-/reload`
5. Go to `http://localhost:12345/metrics` and observe an error similar to:
```
An error has occurred while serving metrics:

4 error(s) occurred:
* collected metric "loki_source_heroku_drain_target_tcp_connections_limit" { label:<name:"component_id" value:"loki.source.heroku.repro" > label:<name:"protocol" value:"http" > gauge:<value:0 > } was collected before with the same name and label values
* collected metric "loki_source_heroku_drain_target_tcp_connections_limit" { label:<name:"component_id" value:"loki.source.heroku.repro" > label:<name:"protocol" value:"grpc" > gauge:<value:0 > } was collected before with the same name and label values
* collected metric "loki_source_heroku_drain_target_tcp_connections" { label:<name:"component_id" value:"loki.source.heroku.repro" > label:<name:"protocol" value:"http" > gauge:<value:0 > } was collected before with the same name and label values
* collected metric "loki_source_heroku_drain_target_tcp_connections" { label:<name:"component_id" value:"loki.source.heroku.repro" > label:<name:"protocol" value:"grpc" > gauge:<value:0 > } was collected before with the same name and label values
```

Could be related to two different servers registering metrics with the same names?

thampiotr:
We seem to have two main options here:

- Don't use server's metrics and expose our own, component-specific metrics instead. 
- Try to change the upstream server behaviour to not call `MustRegister`, but instead call `Register` and use the existing collector if it's returned:
```
	// If the provided Collector is equal to a Collector already registered
	// (which includes the case of re-registering the same Collector), the
[...]

PR #3995, opened 2023-05-26
Remove wrapped registerer

thampiotr:
#### PR Description
While working on a proper fix for https://github.com/grafana/agent/issues/3646 that is implemented in https://github.com/weaveworks/common/pull/289, I found that it's impossible in the flow components to get the previously registered metrics as per prometheus documentation:

```
type Registerer interface {
      ...
	// If the provided Collector is equal to a Collector already registered
	// (which includes the case of re-registering the same Collector), the
	// returned error is an instance of AlreadyRegisteredError, which
	// contains the previously registered Collector.
     ...
	Register(Collector) error
```

this is because we are using `wrappedRegisterer` which does not implement the semantics described above. 

While `wrappedRegisterer` works around some re-registration issues, we found that it didn't fully solve the problem and we had to use another workaround: https://github.com/grafana/agent/pull/3711

`wrappedRegisterer` removes some of the checks that prometheus does (e.g. duplicate metrics) and can easily lead to breaking agent's global `/metrics` endpoint. 

It seems to be a better option to use Prometheus' implementation instead, which should work as expected with 3rd party libraries and provide validation.

The downside is that we may uncover now some dormant issues that can result in a panic. These should be possible to address by reusing existing metrics or unregistering them first.

Fixes/relates to #3646

#### Notes to the Reviewer

Part of this PR is reviewing our existing components to verify they don't register new metrics on `Update` calls. If this is found to be the case, we may be able to either change the code or provide workarounds only for that specific component. 

An alternative to this change (if we're worried about getting a bunch of panics in existing components) could be creating a Registry inside each component that wants a more strict prometheus validation and registering it onto the wrappedRegisterer we get from options... but that still feels like a workaround to me.

#### PR Checklist

- [x] check all components for any metrics registration in `Update`
- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated


rfratto (approved):
LGTM 

I can't remember why we had this originally. 

Your PR here looks correct, since metrics exposed from components are currently collected by an unchecked collector. 

If you've done some local testing with component reloads (of components which expose metrics) and everything is still fine, this should be ok 🤷‍♂️

mattdurham:
@rfratto, I did some archeological digging and I think it links back to `mustRegister` in the wal and remote write and a few other underlying structs. At some point in flow we handled them differently and they could get recreated more often or handled incorrectly. I dont think it is an issue anymore though. Its possible this will expose bad behavior in `Update` for some components but I think that becomes a code smell and something we should fix instead of the original work around.

mattdurham (approved):
LGTM, should simplify things and make them more in line.

component/loki/source/internal/kafkatarget/target_syncer.go
```diff
@@ -33,7 +32,6 @@ type TopicManager interface {
 type TargetSyncer struct {
 	logger log.Logger
 	cfg    Config
-	reg    prometheus.Registerer
```
thampiotr, line 36: The `TargetSyncer` never used this field.

component/prometheus/remotewrite/remote_write_test.go
```diff
@@ -108,9 +82,146 @@ func Test(t *testing.T) {
...
 	select {
-	case <-time.After(time.Minute):
+	case <-time.After(5 * time.Second):
+		require.FailNow(t, "timed out waiting for metrics")
+	case res := <-writeResult:
+		require.Equal(t, expect, res.Timeseries)
+	}
+}
+
+func TestUpdate(t *testing.T) {
```
thampiotr, line 92: Added a test for update of `remote.write`. This was part of debugging to make sure some code paths do not panic when we Update components now - seems like we're safe. Worth keeping the test.

thampiotr (commented):
Added a few small changes - PTAL
e.g. found and fixed one instance of potential panic, but everything else looks safe 🙏

component/loki/process/internal/stages/metric.go
```diff
@@ -108,6 +117,17 @@ func newMetricStage(logger log.Logger, config MetricsConfig, registry prometheus
...
 
+func registerOrReuse(registerer prometheus.Registerer, c prometheus.Collector) (prometheus.Collector, error) {
+	err := registerer.Register(c)
+	if err == nil {
+		return c, nil
+	}
+	if previous, ok := err.(prometheus.AlreadyRegisteredError); ok {
+		return previous.ExistingCollector, nil
+	}
+	return nil, err
```
rfratto, line 128: Should this use [errors.As](https://pkg.go.dev/errors#As) to future proof against wrapped error values?

thampiotr: I made this change, then thought that I should test this and when testing I've found that it's safe to MustRegister, because we are using unchecked metrics: https://github.com/grafana/agent/blob/c6b4c3749863553907ee4fb24bc7263fe3efe7c4/component/loki/process/internal/metric/metricvec.go#L36 

It confused me, so I've added comments instead of my change. Not sure I like that checked/unchecked metrics have the same type, but fundamentally different behaviour, but it's a different problem.

thampiotr:
> Thanks Piotr! Do you think this should be merged for v0.34 or would you prefer to wait until after the release branch is cut?

I don't think it will cause issues, but since it also won't hurt or cause merge conflicts if we wait a bit, I'm happy to delay merging. Thanks!

rfratto:
The release branch was cut, so merging this now 👍

PR #3998, opened 2023-05-26
Prototype Defaulter and Validator for River.

### LINKED ISSUE #3988: Introduce new interface for managing river defaults and validation
Today, Flow components must implement UnmarshalRiver in order to implement applying their own defaults and validations.

Introduce new interface[s] so that river API access to defaults and validation exists. These should be leveraged by at least 1 component to demonstrate the practical usage.

For example, the interface[s] will look something like this.

```golang
type Defaulter interface {
	SetToDefault()
}
```

```golang
type Validator interface {
	Validate() error
}
```

erikbaranowski:
#### PR Description

Prototype to get the ball rolling on a way to implement default and validate methods for river types. Includes an update to the evaluate functionality so that Unmarshaler doesn't have a stack overflow risk built in.

Closes #3988

#### Notes to the Reviewer

- This does not remove the need to implement river.Unmarshaler from a component. Let's discuss what we want to do about that in this context.

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [x] Tests updated


pkg/river/internal/value/value.go
```diff
@@ -25,9 +25,11 @@ var (
 	goCapsule         = reflect.TypeOf((*Capsule)(nil)).Elem()
 	goDuration        = reflect.TypeOf((time.Duration)(0))
 	goDurationPtr     = reflect.TypeOf((*time.Duration)(nil))
-	goRiverDecoder    = reflect.TypeOf((*Unmarshaler)(nil)).Elem()
-	goRawRiverFunc    = reflect.TypeOf((RawFunction)(nil))
-	goRiverValue      = reflect.TypeOf(Null)
+	// goRiverDefaulter  = reflect.TypeOf((*Defaulter)(nil)).Elem()
```
erikbaranowski, line 28: I suspect updates to decode.go will be needed and these commented out lines will become useful. I wanted to start with this prototype before going too far into it.

pkg/river/vm/vm.go
erikbaranowski, line 1: `evaluateBlockOrBody` was refactored to be a series of cascading functions instead of recursive. This implements the new features plus makes it unnecessary to assign to a type not implementing river.Unmarshaler.

rfratto (commented):
On first glance, mostly looks good, but needs some consideration for whether unmarshal or default setting should happen first. 

Can you add unit tests before I give this a more in-depth review?

erikbaranowski:
> On first glance, mostly looks good, but needs some consideration for whether unmarshal or default setting should happen first. 
> 
> Can you add unit tests before I give this a more in-depth review? 

For sure, I will work on it Tuesday.

pkg/river/internal/value/value.go
rfratto: Yeah, Defaulter/Validator should be called roughly in the same place as where Unmarshaler is used.

pkg/river/vm/vm.go
```diff
@@ -85,16 +85,47 @@ func (vm *Evaluator) evaluateBlockOrBody(scope *Scope, assoc map[value.Value]ast
...
+			return err
+		}
+	}
+
+	if decodeElem {
+		return vm.evaluateBlockOrBodyDecode(scope, assoc, node, rv.Elem())
+	}
+
+	return vm.evaluateBlockOrBodyDecode(scope, assoc, node, rv)
+}
```
rfratto, line 126: This doesn't feel right to me. We're validating before decoding?

pkg/river/vm/vm.go
```diff
@@ -85,16 +85,47 @@ func (vm *Evaluator) evaluateBlockOrBody(scope *Scope, assoc map[value.Value]ast
 		rv = rv.Addr()
 	}
 
+	return vm.evaluateBlockOrBodyDefault(scope, assoc, node, rv)
+}
+
+func (vm *Evaluator) evaluateBlockOrBodyDefault(scope *Scope, assoc map[value.Value]ast.Node, node ast.Node, rv reflect.Value) error {
```
rfratto, line 91: I'm on the fence about whether this makes sense as one longer method vs splitting it up like this; introducing four new intermediate methods might make the mental model of the call tree harder to all keep track of in (brain) memory at once.

pkg/river/vm/vm.go
```diff
@@ -85,16 +85,47 @@ func (vm *Evaluator) evaluateBlockOrBody(scope *Scope, assoc map[value.Value]ast
...
+	if ru, ok := rv.Interface().(value.Defaulter); ok {
+		ru.SetToDefault()
+	}
+
+	return vm.evaluateBlockOrBodyUnmarshal(scope, assoc, node, rv)
+}
+
+func (vm *Evaluator) evaluateBlockOrBodyUnmarshal(scope *Scope, assoc map[value.Value]ast.Node, node ast.Node, rv reflect.Value) error {
```
rfratto, line 99: I'm not sure about the order here. It feels like Unmarshal should take precedence over Defaults, so that it becomes:

1. Unmarshaler 
2. Defaulter
3. (decode)
4. Validator 

This is because Unmarshaler primarily allows you to change the type used for decoding, which we do constantly in all of our implementations. It feels like if you're changing the type, that new type should be checked for Defaulter/Validator instead.

WDYT?

rfratto (commented):
Looking good, some general comments though.

pkg/river/vm/vm.go
erikbaranowski: That seems reasonable and I'll experiment with this order.

pkg/river/vm/vm.go
erikbaranowski: I might have an idea I can try here to get us the best of both worlds. It's a little more refactoring but probably worth.

component/prometheus/scrape/scrape.go
```diff
@@ -91,27 +91,26 @@ type Clustering struct {
...
 func (arg *Arguments) UnmarshalRiver(f func(interface{}) error) error {
-	*arg = DefaultArguments
-
-	type args Arguments
-	err := f((*args)(arg))
-	if err != nil {
-		return err
-	}
+	return f(arg)
+}
```
thampiotr, line 110: I realise it's a draft - just a reminder to remove the entire UnmarshalRiver here - AFAICT we won't need it.

pkg/river/types.go
```diff
@@ -26,6 +28,22 @@ type Unmarshaler interface {
 	UnmarshalRiver(f func(v interface{}) error) error
 }
 
+// The Defaulter interface allows a type to implement default functionality
+// in River evaluation.
```
thampiotr, line 32: Q1: If my arguments have a block, then a `SetToDefault()` will be called for it and all its nested blocks?

Q2: If my arguments have a `squash`-ed block will `SetToDefault()` be called on that squashed struct and its nested blocks?

Would be good to document answers to these.

thampiotr (commented):
I like this concept and the approach, thanks for taking a look!

component/prometheus/scrape/scrape.go
erikbaranowski: removed

pkg/river/vm/vm.go
erikbaranowski: I flattened it out a bit but still wrote it in a way the func doesn't grow and grow with all the implementation. Let me know what you think.

erikbaranowski:
I've got a bit more to do here but wanted to push some of the improvements before a day of PTO when I will forget where I left off.

pkg/river/vm/vm.go
```diff
@@ -85,16 +85,42 @@ func (vm *Evaluator) evaluateBlockOrBody(scope *Scope, assoc map[value.Value]ast
 		rv = rv.Addr()
 	}
 
+	if tryUnmarshal {
```
rfratto, line 88: I'm not sure if it's necessary to add this new `tryUnmarshal` conditional in, since we can look at the boolean return value from evaluateBlockOrBodyUnmarshalRiver, right?

```go 
err, unmarshaled := vm.evaluateBlockOrBodyUnmarshalRiver(scope, assoc, node, rv) 
if err != nil {
  return err 
} else if unmarshaled { 
  // vm.evaluateBlockOrBodyUnmarshalRiver recursively unmarshaled, so we don't need 
  // to do anything else now.
  return nil 
}
```

pkg/river/vm/vm.go
rfratto: Definitely an improvement, but left a comment here: https://github.com/grafana/agent/pull/3998/files#r1210837896

pkg/river/vm/vm.go
erikbaranowski: I believe that condition will stop the stack overflow so we don't have to define a new type in each UnmarshalRiver method. If I take it out we will get the existing behavior.

erikbaranowski: This stack overflow prevention has been removed in favor of not introducing more than we need to on this already complex PR. The thought is that with the new default/validate functionality components won't need to implement UnmarshalRiver and therefore not need to do the type casting.

pkg/river/types.go
erikbaranowski: added documentation

pkg/river/internal/value/decode.go
```diff
@@ -126,6 +142,14 @@ func (d *decoder) decode(val Value, into reflect.Value) error {
 		return err
 	}
 
+	if into.CanAddr() {
+		if into.Addr().Type().Implements(goRiverDefaulter) {
+			into.Addr().Interface().(Defaulter).SetToDefault()
+		}
+	} else if into.Type().Implements(goRiverDefaulter) {
+		into.Interface().(Defaulter).SetToDefault()
+	}
```
rfratto, line 151: nit to clean up the depth:

```suggestion
	if into.CanAddr() && into.Addr().Type().Implements(goRiverDefaulter) {
		into.Addr().Interface().(Defaulter).SetToDefault()
	} else if into.Type().Implements(goRiverDefaulter) {
		into.Interface().(Defaulter).SetToDefault()
	}
```

pkg/river/internal/value/decode.go
```diff
@@ -217,6 +247,18 @@ func (d *decoder) decode(val Value, into reflect.Value) error {
 	}
 }
 
+func validate(into reflect.Value) error {
```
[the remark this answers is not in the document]
rfratto: Yeah, I'm a little worried about it being easy to accidentally forgetting how to validate. You could do a named return and a defer statement to validate on close.

rfratto (approved):
LGTM % nit and making it harder to forget to validate on decode

PR #4006, opened 2023-05-29
Update Agent static install on macOS topic

clayton-cornell:
#### PR Description

Updated the Agent install on macOS topic

* Reorganized content and cleaned up existing documentation
* Aligned a bit better with Agent Flow on macOS content

TODO:

- [X]  Fill in missing information for binary install steps - removed ref to binary install
- [X]  Expose UI section from flow is not included in static, is it applicable to static?  No change

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated


clayton-cornell:
@rfratto The Agent static install steps do not cover the binary install. Do we need this? Also the Flow docs detail exposing the UI to other machines. Is this applicable to Static or a Flow-only thing?

rfratto:
> The Agent static install steps do not cover the binary install. Do we need this?

What binary install are you referring to here? 

> Also the Flow docs detail exposing the UI to other machines. Is this applicable to Static or a Flow-only thing?

That's Flow-only, there's no web UI in static mode.

clayton-cornell:
@rfratto the original text has a line `You can also install Grafana Agent on macOS using the macOS binary.`  GitHub has grafana-agent-darwin-amd64.zip which is what I'm guessing this line is referring to?

clayton-cornell:
@rfratto I agree. I've removed the note about the binary version. It's no just the Homebrew solution - one other point in the Homebrew steps.. there's a note about v2 and v3 Homebrew. I question if we need this. Homebrew is on v4 now  and v3 was released in Feb 2021. Is it safe to assume everyone would most likely be on at least v3.x by now? and we could simplify by removing the references to v2?

rfratto:
> and we could simplify by removing the references to v2?

Yes, this makes sense to me 👍

rfratto (approved):

PR #4008, opened 2023-05-30
Agent deployed with ArgoCD remains in endless sync loop

tiithansen:
By reordering metrics argocd will no longer be stuck in endless loop. Related issue in ArgoCD project https://github.com/argoproj/argo-cd/issues/1079

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -118,6 +118,9 @@ Main (unreleased)
   work, leading to better distribution. However, rolling out this change will
   cause some incorrerct or missing assignments until all nodes are updated. (@rfratto)
 
+- Change the Docker base image for Linux containers to `ubuntu:lunar`.
+  (@rfratto)
+
 v0.33.2 (2023-05-11)
 --------------------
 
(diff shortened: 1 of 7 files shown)


rfratto (commented):
Weird, but ok! Thanks for the contribution!

Can you do a few things before we can merge this?

* Can you add an entry to the unreleased section in the chart's [CHANGELOG](https://github.com/grafana/agent/blob/main/operations/helm/charts/grafana-agent/CHANGELOG.md) noting what problem is being fixed? (i.e., the ArgoCD endless sync loop).
* Can you add a comment in the template noting that the order needs to memory before CPU so nobody else accidentally reorders it?

tiithansen:
> Weird, but ok! Thanks for the contribution!
> 
> Can you do a few things before we can merge this?
> 
>     * Can you add an entry to the unreleased section in the chart's [CHANGELOG](https://github.com/grafana/agent/blob/main/operations/helm/charts/grafana-agent/CHANGELOG.md) noting what problem is being fixed? (i.e., the ArgoCD endless sync loop).
> 
>     * Can you add a comment in the template noting that the order needs to memory before CPU so nobody else accidentally reorders it?

Added

rfratto (approved):
Thank you!

PR #4027, opened 2023-05-31
Modules code for implementing tracing and metrics while reverting logging to previous state.

mattdurham:
Provides the bulk of the code for modules


pkg/flow/internal/controller/component.go
```diff
@@ -219,6 +232,9 @@ func (cn *ComponentNode) ComponentName() string { return cn.componentName }
 // block.
 func (cn *ComponentNode) NodeID() string { return cn.nodeID }
 
+// GlobalNodeID returns the parent + nodeid of the component. It is guaranteed to be unique across all dags.
```
mattdurham, line 235: I noticed in a lot of places we logged with `NodeID` which doesnt use the full path.

component/module/file/file_test.go
```diff
@@ -2,6 +2,7 @@ package file
 
 import (
```
mattdurham, line 3: This test had to change a lot since it no longer had direct access to `flow`, I didnt want to remove it entirely so now it uses the http api to test for health.

pkg/flow/internal/controller/metrics.go
```diff
@@ -11,18 +13,20 @@ type controllerMetrics struct {
 }
 
 // newControllerMetrics inits the metrics for the components controller
-func newControllerMetrics(r prometheus.Registerer) *controllerMetrics {
+func newControllerMetrics(r prometheus.Registerer, id string) *controllerMetrics {
```
rfratto, line 16: Our mixin will need to be updated to have some awareness of modules, can you start thinking about how these should be monitored/visualized?

pkg/flow/internal/controller/metrics.go
```diff
@@ -11,18 +13,20 @@ type controllerMetrics struct {
...
 	})
 
 	cm.componentEvaluationTime = prometheus.NewHistogram(
 		prometheus.HistogramOpts{
-			Name: "agent_component_evaluation_seconds",
-			Help: "Time spent performing component evaluation",
+			Name:        "agent_component_evaluation_seconds",
+			Help:        "Time spent performing component evaluation",
+			ConstLabels: map[string]string{"controller_id": id},
 		},
```
rfratto, line 30: These metrics will leak now if a module terminates, since they're registered with MustRegister and never removed.

pkg/flow/module_controller.go
```diff
@@ -13,11 +13,11 @@ var (
 )
 
 // newModuleController is the entrypoint into creating module instances.
-/*func newModuleController(o *moduleControllerOptions) component.ModuleController {
-	return &module{
+func newModuleController(o *moduleControllerOptions) component.ModuleController {
```
rfratto, line 16: This file is so tiny, I feel like it just be rolled up into pkg/flow/module.go.

pkg/util/test_logger.go
```diff
@@ -29,13 +29,12 @@ func TestFlowLogger(t require.TestingT) *logging.Logger {
 		t.Helper()
 	}
 
-	sink, err := logging.WriterSink(os.Stderr, logging.SinkOptions{
+	l, err := logging.New(os.Stderr, logging.Options{
```
rfratto, line 32: TestFlowLogger isn't needed anymore since the interface is back to log.Logger; this can be removed and all instances of using it can be reverted back to `util.TestLogger(t)`.

pkg/flow/internal/dag/dag.go
```diff
@@ -7,6 +7,9 @@ import "fmt"
 type Node interface {
 	// NodeID returns the display name of the Node.
 	NodeID() string
+	// GlobalNodeID returns a globally unique node, parent + node id.
+	// Primarily used for logging or when you absolutely need the global id.
+	GlobalNodeID() string
```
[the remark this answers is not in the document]
mattdurham: GlobalNodeID was only used for tracing and logging so using a wrapper in the loader and passing the controller_id works and allows the removal of GlobalNodeID.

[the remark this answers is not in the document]
erikbaranowski: I think GlobalNodeID() should be added to the BlockNode interface instead of Node.

erikbaranowski (commented):
Looking good, a few comments/questions

component/common/loki/client/client.go
```diff
@@ -363,6 +363,8 @@ func (c *client) sendBatch(tenantID string, batch *batch) {
 				}
 			}
 			return
+		} else {
+			level.Error(c.logger).Log("err", err)
```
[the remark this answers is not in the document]
mattdurham: It does not, I was trying to catch the case where the status check causes it to break before logging, but it will be caught by the final error check so removing.

component/module/file/file_test.go
```diff
@@ -102,11 +102,6 @@ func TestBadFile(t *testing.T) {
 			moduleContents:        `local.fake "fake" {}`,
 			expectedErrorContains: `Unrecognized component name "local.fake"`,
 		},
-		{
```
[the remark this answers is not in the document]
mattdurham: Yes, an empty river config is acceptable. Also allows for noops if using modules with coalesce.

mattdurham:
Oh my, doing a main merge triggered all the reviewers. Though I wonder if we should add a carve out when not using main. Going to give this a twice over and add relevant comments.

pkg/flow/internal/controller/component.go
```diff
@@ -180,21 +191,23 @@ func getManagedOptions(globals ComponentGlobals, cn *ComponentNode) component.Op
...
 		Clusterer: globals.Clusterer,
 
 		DataPath:       filepath.Join(globals.DataPath, cn.nodeID),
 		HTTPListenAddr: globals.HTTPListenAddr,
 		DialFunc:       globals.DialFunc,
 		HTTPPath:       path.Join(prefix, cn.nodeID) + "/",
 
-		OnStateChange: cn.setExports,
+		OnStateChange:    cn.setExports,
+		ModuleController: globals.ModuleController,
```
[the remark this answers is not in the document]
mattdurham: I flipped this around and now throw an error if id is not unique. This could lead to a situation where modules don't know what ID to pass. The current implementation forces modules to use `options.ID` which is unique when used appropriately. I think at some point when introduce multi loading then we can backtrack a bit. Though in general I think module loaders should always have the calling component ID in them so you can backtrack if given a log entry.

go.mod
```diff
@@ -201,6 +201,13 @@ require (
 	sigs.k8s.io/yaml v1.3.0
 )
 
+require (
+	github.com/google/pprof v0.0.0-20230111200839-76d1ae5aea2b
+	github.com/richardartoul/molecule v1.0.1-0.20221107223329-32cfee06a052
+	github.com/spaolacci/murmur3 v1.1.0
+	github.com/zeebo/xxh3 v1.0.2
+)
```
rfratto, line 209: Can you bring these in to the main `require` block and re-run `go mod tidy` so we don't have more than two require blocks?

pkg/flow/module.go
```diff
@@ -2,19 +2,51 @@ package flow
...
+		ids: map[string]struct{}{},
+	}
+}
+
+// NewModule creates a new, unstarted Module.
+func (m *moduleController) NewModule(id string, export component.ExportFunc) (component.Module, error) {
+	if _, found := m.ids[id]; found {
+		return nil, fmt.Errorf("id %s already exists", id)
+	}
+	m.ids[id] = struct{}{}
```
rfratto, line 42: Unless I'm missing something, this leaks now. If you remove a module loader component from a config and then bring it back, it will fail once the config is reloaded.

rfratto (commented):
Gave a quick pass

component/module/file/file_test.go
erikbaranowski, line 1: Was this deleted on purpose? Or did I miss where the tests were moved to? I'm for putting module tests in a more centralized place but would like to leave these in until we do.

mattdurham: I removed it since most of the test overlapped with `string_test.go`. The Good Module test was pretty flakey with going through the the json api and the heart of it is contained within the local file test. I will synthesize these into generic module tests.

mattdurham: @erikbaranowski I added module_test.go that encapsulates all the tests in string/file.

pkg/flow/module.go
```diff
@@ -2,19 +2,63 @@ package flow
...
+
+// NewModule creates a new, unstarted Module.
+func (m *moduleController) NewModule(id string, export component.ExportFunc) (component.Module, error) {
+	m.mut.Lock()
+	defer m.mut.Unlock()
+
+	if _, found := m.ids[id]; found {
+		return nil, fmt.Errorf("id %s already exists", id)
+	}
+	m.ids[id] = struct{}{}
```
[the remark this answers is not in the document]
mattdurham: Its possible to move the ID into the ModuleController instead of the NewModule, which would ensure uniqueness at the cost of creating a ModuleController for each component. But is entirely doable.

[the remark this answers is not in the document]
mattdurham: This does create a confusion if a component wants to create multiple modules since the ID must be unique. In our current cases we don't need that behavior and its best if the value is blank.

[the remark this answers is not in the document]
rfratto: > This does create a confusion if a component wants to create multiple modules since the ID must be unique. In our current cases we don't need that behavior and its best if the value is blank.

What do you mean? They already have to pass an ID to create multiple modules, and the ID must be unique to some scope; I just think it should only be locally unique to that component instance.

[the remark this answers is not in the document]
mattdurham: PR for what it would look like https://github.com/grafana/agent/pull/4117/files.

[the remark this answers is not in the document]
rfratto: > PR for what it would look like https://github.com/grafana/agent/pull/4117/files.

That's fine with me. I'd have some feedback on the documentation for the interface to help make it more clear for users what the expectations are, but otherwise that looks good to help isolate components.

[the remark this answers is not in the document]
mattdurham: Pulled it into this repo, would love any feedback on the documentation on the interface. Adjusted the docs a bit to indicate possible consequences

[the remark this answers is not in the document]
mattdurham: Pulled it into this repo, would love any feedback on the documentation on the interface. Adjusted the docs a bit to indicate possible consequences

mattdurham:
@rfratto other changes you would like to see? I know there is an outstanding removal of `TestFlowLogger` but will handle that in a separate PR to not pollute this one further.

component/registry.go
```diff
@@ -31,8 +31,10 @@ var (
 
 // ModuleController is a mechanism responsible for allowing components to create other components via modules.
 type ModuleController interface {
-	// NewModule creates a new, un-started Module.
-	NewModule(id string, export ExportFunc) Module
+	// NewModule creates a new, un-started Module. The id must be unique to the ModuleController.
+	// ID is appended to the global id of the calling component, and can be left blank.
+	// Though any id used needs to be unique for that component.
```
rfratto, line 36: WDYT about this re-write? I tried to write in in a way so the uniqueness of id only needs to be mentioned once:

```suggestion
	// NewModule creates a new, un-started Module with a given ID. Multiple calls to 
	// NewModule must provide unique values for id. The empty string is a valid unique 
	// value for id. 
```

pkg/flow/internal/controller/component.go
```diff
@@ -63,18 +65,18 @@ type DialFunc func(ctx context.Context, network, address string) (net.Conn, erro
...
+	Clusterer            *cluster.Clusterer                         // Clusterer shared between all managed components.
+	DataPath             string                                     // Shared directory where component data may be stored
+	OnComponentUpdate    func(cn *ComponentNode)                    // Informs controller that we need to reevaluate
+	OnExportsChange      func(exports map[string]any)               // Invoked when the managed component updated its exports
+	Registerer           prometheus.Registerer                      // Registerer for serving agent and component metrics
+	HTTPPathPrefix       string                                     // HTTP prefix for components.
+	HTTPListenAddr       string                                     // Base address for server
+	DialFunc             DialFunc                                   // Function to connect to HTTPListenAddr.
+	ControllerID         string                                     // ID of controller.
+	ModuleControllerFunc func(id string) component.ModuleController // Func to generate a module controller.
```
rfratto, line 79: nit: I was confused what this function did at first. `NewModuleController` might help make it more clear.

pkg/flow/module.go
```diff
@@ -70,6 +120,7 @@ func (c *module) LoadConfig(config []byte, args map[string]any) error {
 //
 // Run blocks until the provided context is canceled.
 func (c *module) Run(ctx context.Context) {
+	defer c.o.parent.removeID(c.o.ID)
```
rfratto, line 123: Can you add documentation to `component.Module.Run` that says the ID of a module won't be freed for re-use until the module stops running?

I want to use that documentation to make sure callers know that modules must _always_ be run, otherwise their IDs will be registered forever.

pkg/flow/module_test.go
```diff
@@ -0,0 +1,250 @@
+//go:build linux
```
rfratto, line 1: Can we remove the linux build tag from this?

rfratto (approved):
Final bits of feedback. Looks really good now, thank you for working on this. 

Most of my feedback at this point is around distributing some of the documentation so users know how to use the API correctly, primarily with how to call NewModule correctly and around making sure that they run the module so the ID can be freed in the future.

PR #4034, opened 2023-06-01
Call out string operators

petewall:
Call out the string operators (just the "+" concatenation operator) by putting them into their own table.

#### PR Checklist

- [ ] CHANGELOG updated
- [X] Documentation added
- [ ] Tests updated


petewall:
Does a change like this go into CHANGELOG?

clayton-cornell:
> Does a change like this go into CHANGELOG?

I'd guess probably not since it's not new content or a new feature, it's a correction to content we already had.

clayton-cornell (approved):
Looks good

clayton-cornell:
Since we discussed this already in Slack I'll merge this one.

PR #4035, opened 2023-06-01
chore(pyroscope): Ensure presence of service_name label

korniltsev:
#### PR Description

The special label `service_name` is required and must always be present. If it's not specified, it is
attempted to be inferred from multiple sources: 
- `__meta_kubernetes_pod_annotation_pyroscope_io_service_name` which is a `pyroscope.io/service_name` pod annotation.
- `__meta_kubernetes_namespace` and `__meta_kubernetes_pod_container_name`
- `__meta_docker_container_name`

If `service_name` is not specified and could not be inferred it is set to `unspecified`.

https://github.com/grafana/phlare/issues/711

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated


component/pyroscope/scrape/target.go
```diff
@@ -400,3 +400,27 @@ func targetsFromGroup(group *targetgroup.Group, cfg Arguments) ([]*Target, []*Ta
...
 }
+
+func inferServiceName(lset labels.Labels) string {
+	if k8sNamespace := lset.Get("__meta_kubernetes_namespace"); k8sNamespace != "" {
+		k8sNameLabels := []string{
+			"__meta_kubernetes_pod_label_pyroscope_name",
+			"__meta_kubernetes_pod_label_app_kubernetes_io_name",
+			"__meta_kubernetes_pod_container_name",
+			"__meta_kubernetes_pod_name",
+		}
```
kolesnikovae, line 411: Should we make agent to understand `pyroscope.io/application-name` annotation (see [example](https://github.com/grafana/pyroscope/tree/main/examples/golang-pull/kubernetes)) for backward compatibility with the existing deployments profiled with pyroscope OSS?

A nit: we could allocate the slice globally

korniltsev: I think if we try to support pyroscope.io annotations, we should support all of them, not just one and supporting all of the may complicate logic

Rperry2174: lets definitely not support all of them. I think if we support any we should support only the applicaiton name, but I'm also fine not supporting this and when we release pyroscope 1.0 we can have people make the change

eh-am: Plus not sure about using the `pyroscope.io` prefix.

docs/sources/flow/reference/components/pyroscope.scrape.md
```diff
@@ -305,6 +305,10 @@ by a service discovery component such as `discovery.kubernetes`. The special
 label `__address__` _must always_ be present and corresponds to the
 `<host>:<port>` that is used for the scrape request.
 
+The special label `service_name` is required and must always be present. If `service_name` is not explicitly provided, it
+will be inferred from the discovery meta labels. If the service name cannot be determined from the discovery meta
+labels, it will default to `unknown_service_name`.
```
kolesnikovae, line 310: We also should keep in mind that as long as we don't support removal from the db, it's might be quite frustrating experience for users to have "unknown" service in the dropdown (same with high cardinality values, such as pod name) – we should be ready to answer this question. On the other hand, provided that we fetch labels only from the head block, the "unknown" app will eventually disappear from the list once the user configuration is fixed.

component/pyroscope/scrape/target.go
```diff
@@ -400,3 +400,28 @@ func targetsFromGroup(group *targetgroup.Group, cfg Arguments) ([]*Target, []*Ta
 
 	return targets, droppedTargets, nil
 }
+
+var k8sNameLabels = []string{
+	"__meta_kubernetes_pod_label_pyroscope_io_name",
```
kolesnikovae, line 405: The way I see it: a diligent user is supposed to create a relabeling config to populate `service_name` from an annotation/label or put it somehow else. If we provide a way to avoid this extra config by putting `pyroscope.io/name` label (for consistency, may be we should call it `pyroscope.io/service-name`), it would be nice to have it mentioned in the documentation.

kolesnikovae (approved):

docs/sources/flow/reference/components/pyroscope.scrape.md
Rperry2174: just trying to understand what this will look like in the UI:
image

cc @eh-am

component/pyroscope/scrape/target.go
kolesnikovae: My idea is to make the configuration process less prone to human errors. I expect that a good share of users that deal with this configuration already have experience of configuring Pyroscope OSS, where we have the `application-name` annotation. I anticipate that due to the human factor, these two might be mixed mistakenly.

docs/sources/flow/reference/components/pyroscope.scrape.md
eh-am: If it defaults to `unknown_service_name`, that's how it will look like in the UI, yes.

Differently from Anton, I feel like it's fine to have `unknown` or even fail to scrape altogether. IMO it's better to fail than trying to be too clever. When one is first setting up it's expected to have a few hiccups. As long as we document (and have proper error messages) it should be fine.

>On the other hand, provided that we fetch labels only from the head block, the "unknown" app will eventually disappear from the list once the user configuration is fixed.
But my understanding is that this is going to change sometime in the future, no?

component/pyroscope/scrape/target.go
Rperry2174: lets go with the concatenation (and risk the repetitions)

kolesnikovae: Yeah, that works. But I wouldn't defer to `pod_container_name` and `pod_name` labels even as a last resort. Moreover, I'd question necessity of falling back to "unknown" service. Maybe we should give up early.

---

On a side note: k8s labels vs annotations – I think it may be slightly better to use annotations only for our purposes. I understand that not everyone follows the rules, but still.

[Annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/): 
> You can use Kubernetes annotations to attach arbitrary non-identifying metadata to [objects](https://kubernetes.io/docs/concepts/overview/working-with-objects/#kubernetes-objects). Clients such as tools and libraries can retrieve this metadata.

[Labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/):
> Labels are intended to be used to specify identifying attributes of objects that are meaningful and relevant to users, but do not directly imply semantics to the core system.

Rperry2174: if "give up early" what happens from a user perspective?

docs/sources/flow/reference/components/pyroscope.scrape.md
kolesnikovae: > Differently from Anton, I feel like it's fine to have unknown or even fail to scrape altogether. IMO it's better to fail than trying to be too clever. When one is first setting up it's expected to have a few hiccups. As long as we document (and have proper error messages) it should be fine.

Just to clarify: I think that the label must be mandatory and explicit in both push and pull modes in the same way it is implemented in Pyroscope, and a missing service name should be threat as a misconfiguration. Profiles without the label should be discarded, and a corresponding message should be logged.

OTOH, I don't know how that works for ebpf.

Rperry2174: If it's not set it still seems like it would be easier to handle in the UI if there was some value that _means_ that it was not set (as opposed to not setting the label at all) wdyt @eh-am? would you rather have no `service_name` label or `service_name: undefined`

component/pyroscope/scrape/target.go
kolesnikovae: 1. No profiling data appears in the UI. So the user don't need to clean up after the unsuccessful attempt
2. A message indicating the misconfiguration is written to the agent log every time the target is discovered

korniltsev: I like the idea to give up early and not being clever for scrape.

For ebpf I still think we should go with concatenation or something else

component/pyroscope/scrape/target.go
```diff
@@ -271,10 +272,6 @@ func populateLabels(lset labels.Labels, cfg Arguments) (res, orig labels.Labels,
 		return nil, nil, errors.New("no address")
 	}
 
-	if v := lset.Get(model.AddressLabel); v == "" {
-		return nil, nil, fmt.Errorf("no address")
-	}
-
```
rfratto, line 277: Is there a reason this is being removed? Isn't an `__address__` label still required so pyroscope.scrape knows where to collect profiles from?

component/pyroscope/scrape/target.go
eh-am: RE: eBPF concatenation, I don't think we need `namespace`, since it already be there (if you can infer via PID -> service discovery), right?

And based on the same principle, you can also look up the same tags you are looking for normal scrape (service_name, name etc).

Then as a last resort, the binary name could be used (somebody suggested it before), no?

component/pyroscope/scrape/target.go
rfratto: Ah, sorry, I missed that. Thanks!

component/pyroscope/scrape/target.go
```diff
@@ -230,9 +230,10 @@ func (ts Targets) Less(i, j int) bool { return ts[i].URL().String() < ts[j].URL(
 func (ts Targets) Swap(i, j int)      { ts[i], ts[j] = ts[j], ts[i] }
 
 const (
-	ProfilePath      = "__profile_path__"
-	ProfileName      = "__name__"
-	ProfileTraceType = "trace"
+	ProfilePath        = "__profile_path__"
+	ProfileName        = "__name__"
+	serviceNameLabel   = "service_name"
```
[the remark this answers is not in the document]
Rperry2174: yeah confirming we decided on `service_name` and _not_ `__service_name__` so that this label is queryable

korniltsev:
Now that we made `service_name` a required label it became another breaking change.
I suppose there will be changes required to be done in deployment_tools. Should I go ahead and prepare PR for deployment_tools or this code is owned by grafana agent team and I should not touch it?
@rfratto @cyriltovena

rfratto:
> I suppose there will be changes required to be done in deployment_tools. Should I go ahead and prepare PR for deployment_tools or this code is owned by grafana agent team and I should not touch it?

You're more than free to touch that code, it's in a weird shared ownership state right now :)

docs/sources/flow/upgrade-guide.md
```diff
@@ -56,6 +56,31 @@ pyroscope.scrape "default" {
 }
 ```
 
+### Breaking change: `service_name` label in is required in `pyroscope.scrape`
```
clayton-cornell, line 59: Is there a word missing in this heading?

korniltsev: thanks! fixed, made it 
```
`service_name` label in is required in `pyroscope.scrape`
```

component/pyroscope/scrape/target.go
```diff
@@ -321,6 +318,13 @@ func populateLabels(lset labels.Labels, cfg Arguments) (res, orig labels.Labels,
 		lb.Set(model.InstanceLabel, addr)
 	}
 
+	if serviceName := lset.Get(serviceNameLabel); serviceName == "" {
+		if serviceName = lset.Get(serviceNameK8SLabel); serviceName == "" {
+			return nil, nil, errors.New("no address")
```
kolesnikovae, line 323:
```suggestion
			return nil, nil, errors.New("service_name label is not specified")
```

clayton-cornell (commented):
Looks OK for the docs.

docs/sources/flow/reference/components/pyroscope.scrape.md
Rperry2174: after discussing further in slack we decided that `service_name:undefined` may be better for the frontend, but will make querying significantly slower (compared to having an explicit service name that querying can rely on).

component/pyroscope/scrape/target.go
```diff
@@ -321,6 +318,13 @@ func populateLabels(lset labels.Labels, cfg Arguments) (res, orig labels.Labels,
 		lb.Set(model.InstanceLabel, addr)
 	}
 
+	if serviceName := lset.Get(serviceNameLabel); serviceName == "" {
+		if serviceName = lset.Get(serviceNameK8SLabel); serviceName == "" {
+			return nil, nil, errors.New("service_name label is not specified")
```
petethepig, line 323: same comment as [here](https://github.com/grafana/phlare/pull/745/files#diff-6ed840df23b67314462615ca210b760563632f290c2bd683c6dc716b1c622dc4)

korniltsev:
I guess we need to decide on https://github.com/grafana/phlare/issues/751 before merging

korniltsev:
@rfratto can we merge this?

PR #4044, opened 2023-06-02
Fix issue where the in-memory dialer was not being passed to modules.

mattdurham:
#### PR Description

The dialerfunc is not being passed to the child modules, this caused exporters to fail since they were exporting agent.internal but serving on normal http.

#### PR Checklist

- [X ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ internal API changes are not present.
 Main (unreleased)
 -----------------
 
+### Bugs
+
+- Fix issue where using exporters in modules failed due to not passing the in-memory address dialer. (@mattdurham)
+
 v0.34.0-rc.0 (2023-06-01)
 --------------------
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -114,6 +114,8 @@ v0.34.0-rc.0 (2023-06-01)
 
 - Fix issue where updating some modules' config (e.g. `loki.process`) could lead to a panic. (@thampiotr) 
 
+- Fix issue where using exporters in modules failed due to not passing the in-memory address dialer. (@mattdurham)
```
rfratto, line 117: This is in the -rc.0 section of the changelog, should it be put to unreleased/main?

mattdurham: Yes, I was jumping the gun a bit.

PR #4045, opened 2023-06-02
Integrations: Add Squid Integration 

armstrmi:
#### PR Description
Adding an integration for the squid_exporter

#### Notes to the Reviewer
Metrics exposed on http://hostname:3182

Notes:
The end goal of this integration is to use the exporter [built by boynux](https://github.com/boynux/squid-exporter)

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ Main (unreleased)
 
 - The Pyroscope scrape component computes and sends delta profiles automatically when required to reduce bandwidth usage. (@cyriltovena)
 
+- Integrations: Introduce the `squid` integration. (@armstrmi)
+
 ### Enhancements
 
 - `prometheus.operator.podmonitors` and `prometheus.operator.servicemonitors` can now access cluster secrets for authentication to targets. (@captncraig)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,7 +71,6 @@ v0.34.0-rc.0 (2023-06-01)
   - `coalesce` returns the first non-zero value from a list of arguments. (@jkroepke)
   - `nonsensitive` converts a River secret back into a string. (@rfratto)
 
-
 ### Enhancements
 
 - Add error value hashing to Faro collector. (@eskirk)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,7 @@ changes that impact end-user behavior are listed; changes to documentation or
 internal API changes are not present.
 
 Main (unreleased)
------------------ 
-
-### Features
-
-- Integrations: Introduce the `squid` integration. (@armstrmi)
+-----------------
 
 ### Features
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,7 @@ v0.34.0-rc.0 (2023-06-01)
   - `coalesce` returns the first non-zero value from a list of arguments. (@jkroepke)
   - `nonsensitive` converts a River secret back into a string. (@rfratto)
 
+
 ### Enhancements
 
 - Add error value hashing to Faro collector. (@eskirk)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,11 @@ changes that impact end-user behavior are listed; changes to documentation or
 internal API changes are not present.
 
 Main (unreleased)
------------------
+----------------- 
+
+### Features
+
+- Integrations: Introduce the `squid` integration. (@armstrmi)
 
 ### Features
 
--- a/docs/sources/static/configuration/integrations/squid-config.md
+++ b/docs/sources/static/configuration/integrations/squid-config.md
@@ -7,7 +7,7 @@ aliases:
 # squid_config
 
 The `squid_config` block configures the `squid` integration,
-which is an embedded version of a forked version of the [`Squid_exporter`](https://github.com/boynux/squid-exporter). This allows the collection of third party [Squid](http://www.squid-cache.org/) metrics.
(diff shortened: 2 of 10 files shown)


pkg/integrations/squid_exporter/squid_exporter.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// InstanceKey returns the addr of the squid instance.
+func (c *Config) InstanceKey(agentKey string) (string, error) {
+	return c.Address, nil
+}
+
+// NewIntegration returns the Squid Exporter Integration
+func (c *Config) NewIntegration(_ log.Logger) (integrations.Integration, error) {
+	return New(c)
```
schmikei, line 85: I notice that we still aren't passing in a logger here, I just want to note that any errors that the exporter here would not get viewed in the agent log. This is definitely undesireable but I'm not sure if it warrants blocking this PR. I think the agent team can determine if this is a blocker. 

Do not resolve this before getting their review :)

pkg/integrations/squid_exporter/squid_exporter.go
```diff
@@ -0,0 +1,111 @@
...
+
+	port, _ := strconv.Atoi(scheme)
+	c.Port = port
+
+	return nil
+}
+
+// UnmarshalYAML implements yaml.Unmarshaler for Config
+func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	*c = DefaultConfig
```
schmikei, line 67: Is the port for squid generally just localhost:3128? I wonder if its worth trying to determine if only a host is provided if we can assume the default port if that's the case.

schmikei (changes requested):
I'm still not up to date on the documentation standards so that may need some touch up as well. Looks promising though!

pkg/integrations/squid_exporter/squid_exporter.go
```diff
@@ -0,0 +1,111 @@
...
+	if host == "" {
+		return errNoHostname
+	}
+	c.Host = host
+
+	if scheme == "" {
+		return errNoPort
+	}
+
+	port, _ := strconv.Atoi(scheme)
```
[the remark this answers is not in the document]
armstrmi: Hey Keith, do you mean like ignoring the error portion?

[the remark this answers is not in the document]
schmikei: Yeah the `Atoi` call returns an error which we are specifically ignoring on this line in the validate. Just feels like if it fails we should return an error rather than ignoring it and getting a potential panic somewhere down the road.

pkg/integrations/squid_exporter/squid_exporter.go
armstrmi: I believe that a default port for squid is 3128, but apparently 8080 is also common, but you're proposing that if a hostname is provided that then we could assume that the port is 3128 and fill it in?

schmikei: I'm not super familiar with Squid but yeah if there is a reasonable default AND the library doesn't assume it and we are validating existence of the port I think it might be applicable to assume a user meant whatever is the default port.

pkg/integrations/squid_exporter/squid_exporter.go
schmikei: I guess I just wanted to note that this component as currently written appears to swallow errors and the Grafana team should note that before accepting this component.

pkg/integrations/squid_exporter/squid_exporter_test.go
```diff
@@ -0,0 +1,141 @@
...
+	cases := []struct {
+		name        string
+		getConfig   func() Config
+		expectedErr error
+	}{
+		{
+			name: "valid",
+			getConfig: func() Config {
+				c := DefaultConfig
+				c.Address = "squid-service.sample-apps.svc.cluster.local:3128"
```
schmikei, line 23: I dont think you wanted to commit this up.

docs/sources/static/configuration/integrations/squid-config.md
```diff
@@ -0,0 +1,81 @@
...
+  # The address used to connect to the Squid instance in the format
+  # of <HOST>:<PORT>.
+  # i.e. "localhost:3128"
+  [address: <string>]
+
+  # The username for squid instance.
+  [username: <string>]
+
+  # The password for username above.
+  [password: <string> | default = "password"]
```
StefanKurek, line 66: do passwords usually have a default? seems a little odd?

StefanKurek (commented):
Looks generally good once you address @schmikei 's concerns

CHANGELOG.md
```diff
@@ -8,12 +8,18 @@ changes that impact end-user behavior are listed; changes to documentation or
 internal API changes are not present.
 
 Main (unreleased)
------------------
+----------------- 
+
+### Features
+
+- Integrations: Introduce the `squid` integration. (@armstrmi)
```
StefanKurek, line 15: @armstrmi mistake?

clayton-cornell (commented):
Docs look OK. I've added one comment/suggestion.

StefanKurek (approved):

marctc (commented):
Hey, I took a quick look and overall looks good. Just a reminder that this PR should also include the Flow component for the integration. More info how to add it [here](https://github.com/boynux/squid-exporter).

marctc:
> @marctc Hello. I had asked Mischa Thomson about if we needed to have a single PR or if we could have two separate ones, and he mentioned that two should be fine. If this is incorrect, let me know.

Oh sure, no problem!

marctc (approved):

armstrmi:
@marctc Thanks for the approval! Would you mind merging? I would myself but can't lol

PR #4048, opened 2023-06-02
Tanka config: retain cAdvisor metrics for system processes

bboreham:
#### PR Description

Currently, CPU metrics for Kubelet, Containerd, etc. are dropped by a relabel rule since they don't have a container image.

On most systems they run in their own cgroup called `system.slice`, so we can add another rule which uses that fact to give the `image` label a value and then they won't be dropped.

#### Notes to the Reviewer

I haven't tried this out yet.

#### PR Checklist

- [x] CHANGELOG updated
- NA Documentation added
- NA Tests updated


tpaschalis (approved):
LGTM

PR #4052, opened 2023-06-05
Support producing delta profiles in Pyroscope scrape component

cyriltovena:
#### PR Description

This PR introduces the computation of delta profiles in `pyroscope.scrape` component before shipping them to the target. This is only done for some cumulative profile types. We're doing this to reduce the network bandwidth between the Agent and the server.

Right now the pyroscope server does this but we think the agent is better place to reduce bandwidth and resource usage.

The first profile will return an error and be discarded.

Fixes https://github.com/grafana/phlare/issues/694

#### Notes to the Reviewer

I had to copy datadog [`fastdelta`](https://github.com/DataDog/dd-trace-go/tree/v1.50.0/profiler/internal/fastdelta) package as it is internal.

#### PR Checklist

- [x] CHANGELOG updated
~~- [ ] Documentation added~~
- [x] Tests updated


cyriltovena:
I haven't made any documentation change because I don't think this is worth knowing from a user standpoint.

korniltsev (approved):
LGTM

mattdurham (approved):
LGTM.

PR #4053, opened 2023-06-05
[Flow] Fix attribute docs for the Prometheus Windows Exporter

ptodev:
I think there was a mixup in the documentation for the windows exporter. `enabled_list` was documented in the way that `enabled_collectors` should have been documented, and vice versa. 

### Note to reviewers

I raised this PR just to fix the docs, but I think we should make another PR to also convert the `enabled_list` to a `list(string)`


clayton-cornell (approved):
Looks good to me

clayton-cornell:
@ptodev Since changing the datatype for `enabled_list` from `string` to `list(string)` requires a code change, I'll push this docs-only correction through.

PR #4057, opened 2023-06-05
matching defaults removed when encoding optional river attributes and blocks

### LINKED ISSUE #3900: Don't include defaults in the output converted config if the value matches the Flow default
When doing a conversion to Flow config, implement a way to not include config that matches the defaults in Flow. Update existing conversions to use this mechanism.

For example,

```river
	endpoint {
		name             = "remote1"
		url              = "http://remote-write-url1"
		remote_timeout   = "30s"
		follow_redirects = true
		enable_http2     = true

		queue_config {
			capacity             = 2500
			max_shards           = 200
			min_shards           = 1
			max_samples_per_send = 500
			batch_send_deadline  = "5s"
			min_backoff          = "30ms"
			max_backoff          = "5s"
		}

		metadata_config {
			send                 = true
			send_interval        = "1m0s"
			max_samples_per_send = 500
		}
	}
```

becomes

```river
	endpoint {
		name             = "remote1"
		url              = "http://remote-write-url1"
	}
```

erikbaranowski:

Closes #3900

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [x] Tests updated


component/common/relabel/relabel.go
```diff
@@ -166,7 +170,7 @@ func (rc *Config) UnmarshalRiver(f func(interface{}) error) error {
 	}
 
 	if rc.Action == KeepEqual || rc.Action == DropEqual {
-		if rc.Regex != DefaultRelabelConfig.Regex ||
+		if !reflect.DeepEqual(*rc.Regex.Regexp, *DefaultRelabelConfig.Regex.Regexp) ||
```
erikbaranowski, line 173: This was updated because the comparison was checking pointers previously rather than the values which won't work anymore.

go.mod
erikbaranowski, line 1: `go mod tidy`

component/common/relabel/relabel_test.go
```diff
@@ -75,6 +75,8 @@ func TestParseConfig(t *testing.T) {
 			t.Parallel()
 			var cfg Config
 			err := river.Unmarshal([]byte(tt.cfg), &cfg)
+			cfg.SetToDefault()
+			cfg.Validate()
```
rfratto, line 79: I'm not sure what's going on here; why is SetToDefault being called after Unmarshal, and why is the error of cfg.Validate() not checked?

converter/internal/prometheusconvert/testdata/relabel.river
```diff
@@ -4,58 +4,34 @@ prometheus.remote_write "default" {
 	}
 
 	endpoint {
-		name             = "remote1"
-		url              = "http://remote-write-url1"
-		remote_timeout   = "30s"
-		follow_redirects = true
-		enable_http2     = true
+		name = "remote1"
```
rfratto, line 7: Very clean 👍

go.mod
```diff
@@ -200,6 +200,8 @@ require (
 	sigs.k8s.io/yaml v1.3.0
 )
 
+require github.com/klauspost/compress v1.15.15
```
rfratto, line 203: Can this be put in the other require section above and re-run `go mod tidy` so we don't have too many require sections?

pkg/river/token/builder/builder.go
```diff
@@ -131,6 +132,10 @@ func (b *Body) AppendBlock(block *Block) {
 // Body. If any value reachable from goValue implements Tokenizer, the printed
 // tokens will instead be retrieved by calling the RiverTokenize method.
 //
+// Optional attributes and blocks may be trimmed. If the goValue implements
+// Defaulter, any values matching the defaults will be trimmed. If not, any
+// values matching the golang defaults will be trimmed.
```
rfratto, line 137: WDYT about this rewrite to make the language stronger about trimming?

```suggestion
// Optional attributes and blocks set to default values are trimmed. 
// If goValue implements Defaulter, default values are retrieved by 
// calling SetToDefault against a copy. Otherwise, default values are 
// the zero value of the respective Go types.
```

pkg/river/token/builder/builder.go
```diff
@@ -166,9 +171,21 @@ func (b *Body) encodeFields(rv reflect.Value) {
 	}
 
 	fields := rivertags.Get(rv.Type())
+	clone := reflect.New(rv.Type()).Elem()
```
rfratto, line 174: Rather than calling it `clone`, what do you think about a name more descriptive of what we're using it for, like `defaults`?

pkg/river/token/builder/builder.go
```diff
@@ -166,9 +171,21 @@ func (b *Body) encodeFields(rv reflect.Value) {
...
+
+	var goRiverDefaulter = reflect.TypeOf((*value.Defaulter)(nil)).Elem()
```
rfratto, line 176: Can this be a package-level variable so we're not always calculating it at runtime for each call?

pkg/river/token/builder/builder.go
```diff
@@ -166,9 +171,21 @@ func (b *Body) encodeFields(rv reflect.Value) {
...
+	if rv.CanAddr() && rv.Addr().Type().Implements(goRiverDefaulter) {
+		clone.Addr().Interface().(value.Defaulter).SetToDefault()
+	}
```
rfratto, line 179: Because we're calling SetToDefault on `clone`, we should be checking whether `clone.CanAddr()` is true rather than `rv.CanAddr()`; the two aren't linked, so it's possible that clone could be addressible when rv isn't. 

In fact, clone will _always_ be addressible because we create it as a pointer above with `reflect.New`.

Given that we need to check the addressability of clone instead, it may make more sense to make the whole condition about clone instead of rv:

```suggestion
	if clone.CanAddr() && clone.Addr().Type().Implements(goRiverDefaulter) {
		clone.Addr().Interface().(value.Defaulter).SetToDefault()
	}
```

rfratto (approved):
👍 Very nice, love how small the implementation in river/token/builder is. 

Just a few comments, but this is ready to merge after they're addressed!

component/common/relabel/relabel_test.go
erikbaranowski: I think the technical term is a 'brain fart'

PR #4059, opened 2023-06-05
Apply river.Defaulter and river.Validator [almost] everywhere

erikbaranowski:
#### PR Description

This PR is intended to have 0 behavioral impact on any Flow component.

This PR will bring [almost] all components up to speed with the new river.Defaulter and river.Validator interfaces rather than using the river.Unmarshaler interface. There are 2 exceptions which require slightly more refactoring and should be part of a separate PR so they can be looked at closer.

image

This PR positions us to have a single standardized way for managing defaults and validation for components rather than a split brain with components doing defaults/validation different ways. It will also do the necessary wiring so that the new Config Converter API will not need to update specific flow components as it gets implemented.

#### Notes to the Reviewer

Apologies for the size of the PR in advance. This is applying a similar pattern to every flow component so the width of the PR is large.

Common patterns I saw
- If UnmarshalRiver was implemented for defaulting behavior, it was replaced by SetToDefault
- If UnmarshalRiver was implemented for validation behavior, it was replaced by Validate
- If Validate was already implemented then called by UnmarshalRiver, UnmarshalRiver could be safely removed
- If anything different was happening in UnmarshalRiver, UnmarshalRiver was left alone for now

#### PR Checklist

- [ ] CHANGELOG updated
- [x] Documentation added
- [ ] Tests updated


thampiotr (approved):
LGTM! 

The only small nit I could think of is that the `var _ river.Defaulter = (*Arguments)(nil)` type conformance checks are quite nice and we could have them. But I would say it's more valuable to merge this soon than to be spending extra time adding these :)

component/otelcol/auth/sigv4/sigv4.go
```diff
@@ -50,6 +50,7 @@ func (args Arguments) Convert() (otelconfig.Extension, error) {
 	return &res, nil
 }
 
+// UnmarshalRiver implements river.Unmarshaler
```
rfratto, line 53:
```suggestion
// UnmarshalRiver implements river.Unmarshaler.
```

rfratto (approved):
🎉

PR #4078, opened 2023-06-07
fix(azure_event_hubs): default scope

akselleirv:
#### PR Description

Fixed an issue I had when using workload identity in AKS to authenticate to Azure Event Hubs. It was fixed by adding the path `/.default` to the scope. 

#### PR Checklist

- [x] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,8 @@ Main (unreleased)
   could lead to propagating issues downstream to other components which expect
   the export to be non-empty. (@rfratto)
 
+- Fix oauth default scope in `loki.source.azure_event_hubs`. (@akselleirv)
+
 ### Other changes
 
 - Add logging to failed requests in `remote.http`. (@rfratto)
--- a/component/loki/source/azure_event_hubs/azure_event_hubs.go
+++ b/component/loki/source/azure_event_hubs/azure_event_hubs.go
@@ -192,7 +192,7 @@ func (a *Arguments) Convert() (kt.Config, error) {
 			if err != nil {
 				return kt.Config{}, fmt.Errorf("unable to extract host from fully qualified namespace: %w", err)
 			}
-			a.Authentication.Scopes = []string{fmt.Sprintf("https://%s", host)}
+			a.Authentication.Scopes = []string{fmt.Sprintf("https://%s/.default", host)}
 		}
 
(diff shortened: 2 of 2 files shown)


rfratto:
Thanks for the fix! 

I'm trying to figure out why Promtail isn't doing this, and whether always providing the `/.default` might break a subset of use-cases.

rfratto:
Ah, it seems you contributed the component in the first place :) Do you know if `/.default` should work for everyone?

akselleirv:
I might make sense moving this to the documentation. When I was testing it then it worked fine using az CLI to authenticate to the event hub. However, yesterday when I used Workload Identity I got an error which indicated that the scope was wrong. Then after enabling more logging for the azidentity library I was able to get a hint about that scope.

I was unfortunately not able to find any documentation regarding this when I tried yesterday.

So maybe move this to the documentation instead?

rfratto:
I'll admit I don't understand enough about Azure Workload Identity so I'm going to have to rely on your knowledge here. If you think moving this to documentation makes sense, that's fine with me. 

That being said, what would we document? Based on the code, the default OAuth scopes for Workload Identity isn't correct, right? Should the documentation say to always manually add OAuth scopes?

akselleirv:
Sorry for being so back and forth, but I'm also not 100% sure what the correct thing to do is, as I have yet to find any clear documentation for the correct scope.

I found, however,  some indications that the adding the `/.default` to the scope is the correct thing to do. 
- For example at the [azeventhubs library](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/messaging/azeventhubs/internal/sbauth/token_provider.go#L94). 
- Also at [this stack overflow thread](https://stackoverflow.com/questions/63525478/how-do-you-get-an-access-token-from-azure-active-directory-v2-to-allow-access).
- This is also what works for me when I'm using this the component.

So it seems that the correct thing is to have the `/.default` path to the scope.

rfratto:
No worries, we're in this together 😛 

Let's merge this for now and see what happens.

Can you add this to the CHANGELOG as a bug fix before we merge?

akselleirv:
Good :D 

I've updated the CHANGELOG.

rfratto (approved):
LGTM, thanks!

PR #4079, opened 2023-06-07
Modify the address pointed to by the Agent integration link

oliver-zhang:
#### PR Description
Agent integration link address does not exist，modify to the correct place


rfratto (approved):
Thanks for the fix!

PR #4094, opened 2023-06-08
Add servicemonitor support

QuentinBisson:
#### PR Description

This PR adds ServiceMonitor support to the grafana helm chart to ensure the grafana agent can be monitored

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated


rfratto:
Thank you! Looks reasonable to me, but I'll let @captncraig give the final review since he self-requested one.

You'll have to fix some merge conflicts since we just put out a new Helm chart release alongside v0.34, sorry about that!

operations/helm/charts/grafana-agent/Chart.yaml
```diff
@@ -3,4 +3,4 @@ name: grafana-agent
 description: 'Grafana Agent'
 type: application
 version: 0.15.0
-appVersion: 'v0.34.0'
+appVersion: 'v0.34.0'
```
QuentinBisson, line 6:
```suggestion
appVersion: 'v0.34.0'

```

QuentinBisson:
@rfratto Thanks I fixed the conflicts, I will let @captncraig review it :)

TheoBrigitte (approved):

captncraig:
I'm not necessarily opposed to this, but I am slightly concerned that flag is sending a wrong message. 

Activating this alone is not enough to monitor the agent. You would still need to scrape the metrics endpoint with a `prometheus.operator.servicemonitors` component, or through a direct scraping of the target. My main concern would be that somebody will activate `serviceMonitor.enabled` and then be confused why their agent metrics are not showing up. 

Maybe this just needs a good comment at the top of the `serviceMonitor` block in values.yaml?

I personally prefer to use a `prometheus.scrape` component with a static target to `localhost`. It has fewer "moving parts" to do it that way. That doesn't necessarily mean this approach is wrong, and yes, we need to better document that kind of recommendation.

QuentinBisson:
@captncraig I can definitely add some documentation but for my use case, I'm not using the agent to monitor anything, I am only using it to move prometheus rules CR to mimir :)

operations/helm/charts/grafana-agent/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,26 @@
+{{- if and .Values.service.enabled .Values.serviceMonitor.enabled -}}
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: {{ include "grafana-agent.fullname" . }}
```
rfratto, line 5: Based on the lint test failing, it looks like this needs a namespace field for the chart to install successfully.

QuentinBisson:
@rfratto I think the issue is that the kind cluster is not aware about servicemonitor CRDs. Do you know where I could add it so the CRD is created by the CI in kind only?

rfratto:
> @rfratto I think the issue is that the kind cluster is not aware about servicemonitor CRDs. Do you know where I could add it so the CRD is created by the CI in kind only?

I'm not completely sure. The Kind cluster is created [here](https://github.com/grafana/agent/blob/1f28204254c155ccf87b430d9d1aea2506de5542/.github/workflows/helm-test.yml#L74) and supposedly it can be given a [Kind config](https://github.com/helm/kind-action#inputs), but it doesn't look like Kind configs support being given a list of CRDs. 

It seems like the most straightforward path would be to add a new step after the cluster is created to install the CRDs

nlamirault:
any news on this feature ? 

you could add a step to install Prometheus Operator CRDs:

```yaml
- name: Add dependency chart repos
  run: |
    helm repo add prometheus-community https://prometheus-community.github.io/helm-charts

- name: Install Prometheus Operator CRDs
  run: |
    helm install my-prometheus-operator-crds prometheus-community/prometheus-operator-crds --version 6.0.0 
```

QuentinBisson:
thank you @nlamirault sorry i was working on other topics :(

clayton-cornell:
Does there need to be any addition docs added other than the READMEs? I guess no, but want to be sure.

QuentinBisson:
@clayton-cornell i'm not sure, I would think the values file would be enough but I can check on monday

rfratto:
Hi, I'm on call this week so I have very limited time for code reviews. Since this has been opened for a while, I'd like to see you get this merged without waiting for me to have more time, so I'm going to hand this off to someone else for final review so we can get it in sooner :)

captncraig (approved):

Issue #2715, opened 2023-01-10
Flow: support TLS config similar to static agent


rfratto:
We need to bring over the TLS server options into the config file, similarly to the config exposed in the static agent.

See also #2984 to ensure integrations still working as expected when TLS is enabled. #2984 and #4253 is a prerequisite to being able to implement this functionality.

PR #4097, opened 2023-06-08
Flow: add new methods to get and list components

rfratto:
> **NOTE TO REVIEWER**: I've split this up across a few commits to hopefully make this easier to review. 

This PR refactors the Flow controller API to allow it to return detail on running components via the new `(*Flow).ListComponents` and `(*Flow).GetComponents` methods. These replace the old `(*Flow).ComponentInfos` API used by the UI as it exposes more raw information to the caller.  

As probing detail about components can be expensive, these new methods support options to control how much detail is returned about components, such as omitting debug information altogether. 

This is slightly more code than what we previously had, but gives callers more power over what to do with the list of running components. 

## Background 

#2715 is one of the remaining items required for feature parity: the Flow HTTP server should support being configured with TLS in the configuration file. However, due to the nature of Flow, this means that TLS settings should _ideally_ support being retrieved via components. The most obvious use case would be to retrieve TLS keys from Vault and use that to configure the agent. 

However, this introduces a problem: the HTTP server is started prior to Flow. For the HTTP server to depend on a component for settings means that the HTTP server must exist in the DAG so it can be resolved when needed. This will eventually lead to a "services" proposal. 

The intent for the "services" proposal will be to introduce services which are decoupled from the Flow controller. Because of this, this PR is necessary as a prerequisite so that code running outside of the controller can get extremely detailed information about the state of components. For example, an HTTP service will need to poll the list of components and see if any of them implement the `component.HTTPComponent` interface.

> **NOTE**: I have done some forward thinking here, where I am fairly confident we will want these APIs to also be able to return component detail from components running in modules. I have prepared the API surface to be able to do this, but it's not implemented yet and is documented as a known bug. 
>
> The ability to have the controller dig into modules is blocked until the refactoring work in #4027 is completed.


pkg/flow/flow_components.go
```diff
@@ -0,0 +1,240 @@
...
+	)
+
+	if references == nil {
+		references = []string{}
+	}
+	if referencedBy == nil {
+		referencedBy = []string{}
+	}
+
+	arguments, err = riverjson.MarshalBody(cd.Arguments)
```
erikbaranowski, line 88: Sanity check me on these. Looking at the riverjson.MarshalBody tests are we getting the wanted behavior when we don't have arguments, exports, etc. I see the omitempty tags above so maybe that covers it?

image

erikbaranowski (commented):
Looking solid overall. I added some questions and sanity checks along the way.

cmd/internal/flowmode/cmd_run.go
```diff
@@ -355,12 +355,13 @@ func (fr *flowRun) Run(configFile string) error {
 // getEnabledComponentsFunc returns a function that gets the current enabled components
 func getEnabledComponentsFunc(f *flow.Flow) func() map[string]interface{} {
 	return func() map[string]interface{} {
-		infos := f.ComponentInfos()
-		components := map[string]struct{}{}
-		for _, info := range infos {
-			components[info.Name] = struct{}{}
+		components := f.ListComponents(flow.ComponentDetailOptions{})
+
+		componentNames := map[string]struct{}{}
```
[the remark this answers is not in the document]
rfratto: `map[string]struct{}` is a common pattern in Go when you want a O(1) lookup table for keys but don't need any values; `struct{}` is a zero-sized value so it takes up no memory compared to doing something like a `map[string]bool` or a `map[string]interface{}`.

pkg/flow/flow_components.go
```diff
@@ -0,0 +1,240 @@
...
+// [ComponentDetailOptions] for more information.
+//
+// GetComponent returns an error if a component is not found.
+//
+// BUG(rfratto): The ModuleID in the component ID is ignored.
+func (f *Flow) GetComponent(id ComponentID, opts ComponentDetailOptions) (*ComponentDetail, error) {
+	f.loadMut.RLock()
+	defer f.loadMut.RUnlock()
+
+	// TODO(rfratto): navigate running modules and return a component within a
```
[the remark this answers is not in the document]
rfratto: No data loss; the existing code already couldn't navigate through modules, but I'm thinking the new API should be able to in the near future.

pkg/flow/flow_components.go
```diff
@@ -0,0 +1,240 @@
...
+	var references, referencedBy []string
+
+	// Skip over any edge which isn't between two component nodes. This is a
+	// temporary workaround needed until there's athe concept of configuration
+	// blocks in the API.
+	//
+	// Without this change, the graph fails to render when a configuration
+	// block is referenced in the graph.
+	//
+	// TODO(rfratto): add support for config block nodes in the API and UI.
```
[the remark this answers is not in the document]
rfratto: This was already a limitation; if you look at the removed old code you'll find the same comment.

pkg/flow/flow_components.go
```diff
@@ -0,0 +1,240 @@
...
+		},
+		Arguments: arguments,
+		Exports:   exports,
+		DebugInfo: debugInfo,
+	})
+}
+
+// ComponentDetailOptions is used by [Flow.ListComponents] and
+// [Flow.GetComponent] to determine how much information
+type ComponentDetailOptions struct {
```
mattdurham, line 121: Should the `ComponentDetailOptions`  be passed along with the `ComponentDetail` so that state of what was retrieved doesn't get lost if the `ComponentDetail` is passed around? I am unsure if there is a way to tell a `nil` Exports from a didn't retrieve exports.

tpaschalis (approved):
LGTM, nothing else to add! 🛥️

pkg/flow/flow_components.go
rfratto: IMO this should be the responsibility of the caller to expose what detail was included if they want to propagate that information downstream to _their_ callers.

I'd like to stat with not including the options and revisit later if it's confusing. WDYT?

erikbaranowski (approved):
LGTM

pkg/flow/flow_components.go
rfratto: I'll wrap them in checks to be safe.

pkg/flow/flow_components.go
```diff
@@ -0,0 +1,240 @@
...
+		}
+
+		componentDetailJSON struct {
+			Name         string               `json:"name,omitempty"`
+			Type         string               `json:"type,omitempty"`
+			ID           string               `json:"id,omitempty"`
+			Label        string               `json:"label,omitempty"`
+			References   []string             `json:"referencesTo"`
+			ReferencedBy []string             `json:"referencedBy"`
+			Health       *componentHealthJSON `json:"health"`
```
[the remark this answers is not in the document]
rfratto: No; this matches the existing behavior and matches the expectations from the UI for what fields must always be present.

pkg/flow/flow_components.go
```diff
@@ -0,0 +1,240 @@
...
+	if id.ModuleID == "" {
+		return id.LocalID
+	}
+	return id.ModuleID + "/" + id.LocalID
+}
+
+type ComponentDetail struct {
+	// Component is the instance of the component. Component will be nil if the
+	// component exists in the graph but has not successfully evaluated yet.
+	Component component.Component
```
mattdurham, line 31: Is Component used anywhere?

rfratto: Not at the moment. It will play a critical part in my design for services though.

mattdurham (approved):
LGTM

Issue #4073, opened 2023-06-06
Add documentation for using the config conversion tools for end users


erikbaranowski:
brainstorm
- how defaults are handled
- how unsupported config is handled
- use cases
- recommendations
- mentioning which version of upstream things we are converting from

PR #4106, opened 2023-06-08
Add support to convert azure_sd_configs to Flow

erikbaranowski:
#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/converter/internal/prometheusconvert/prometheusconvert.go
+++ b/converter/internal/prometheusconvert/prometheusconvert.go
@@ -76,7 +76,7 @@ func AppendAll(f *builder.File, promConfig *promconfig.Config) diag.Diagnostics 
 				// targets = append(targets, exports.Targets)
 				diags = append(diags, newDiags...)
 			default:
-				diags.Add(diag.SeverityLevelWarn, fmt.Sprintf("unsupported service discovery %s was provided", sdc.Name()))
+				diags.Add(diag.SeverityLevelWarn, fmt.Sprintf("unsupported service discovery %s was provided", serviceDiscoveryConfig.Name()))
 			}
 		}
 
--- a/converter/internal/prometheusconvert/prometheusconvert.go
+++ b/converter/internal/prometheusconvert/prometheusconvert.go
@@ -69,7 +69,7 @@ func AppendAll(f *builder.File, promConfig *promconfig.Config) diag.Diagnostics 
 		for _, serviceDiscoveryConfig := range scrapeConfig.ServiceDiscoveryConfigs {
 			switch sdc := serviceDiscoveryConfig.(type) {
 			case promdiscover.StaticConfig:
-				continue
+				targets = append(targets, getScrapeTargets(sdc)...)
 			case *promazure.SDConfig:
 				_, newDiags := appendDiscoveryAzure(f, scrapeConfig.JobName, sdc)
 				// exports, newDiags := appendDiscoveryAzure(f, scrapeConfig.JobName, sdc)
@@ -77,13 +77,6 @@ func AppendAll(f *builder.File, promConfig *promconfig.Config) diag.Diagnostics 
 				diags = append(diags, newDiags...)
 			default:
 				diags.Add(diag.SeverityLevelWarn, fmt.Sprintf("unsupported service discovery %s was provided", sdc.Name()))
-			}
-		}
-
-		for _, serviceDiscoveryConfig := range scrapeConfig.ServiceDiscoveryConfigs {
-			switch sdc := serviceDiscoveryConfig.(type) {
-			case promdiscover.StaticConfig:
-				targets = append(targets, getScrapeTargets(sdc)...)
 			}
 		}
 
--- a/converter/internal/prometheusconvert/prometheusconvert.go
+++ b/converter/internal/prometheusconvert/prometheusconvert.go
@@ -5,10 +5,12 @@ import (
 	"fmt"
 
 	"github.com/go-kit/log"
+	"github.com/grafana/agent/component/discovery"
 	"github.com/grafana/agent/converter/diag"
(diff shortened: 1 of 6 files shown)


converter/internal/common/convert_targets.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	for ix, targetMap := range f.Targets {
+		if len(targetMap) != 1 {
+			panic("unexpected number of targets received on a target map")
+		}
+
+		for key, target := range targetMap {
+			if key == "__address__" {
+				toks = append(toks, builder.Token{Tok: token.LBRACK, Lit: ""})
```
erikbaranowski, line 39: I could stack all these on a single append if we prefer. More importantly, if there's a slicker way to create this whole []builder.Token let me know. I don't think it's too awful this way either though.

mattdurham: Would it be easier to parse a string into tokens? The below isnt bad but just a bit verbose. Also a comment showing what the expected output would be super cool so we dont have to fully decipher the tokens meaning.

erikbaranowski: I can add an example for sure, good idea. For the string into tokens suggestion, do we have a clear way to do that already or something new to consider?

converter/internal/prometheusconvert/prometheusconvert.go
```diff
@@ -53,28 +52,35 @@ func Convert(in []byte) ([]byte, diag.Diagnostics) {
...
+			switch sdc := serviceDiscoveryConfig.(type) {
+			case promdiscover.StaticConfig:
+				targets = append(targets, getScrapeTargets(sdc)...)
+			case *promazure.SDConfig:
+				exports, newDiags := appendDiscoveryAzure(f, scrapeConfig.JobName, sdc)
+				targets = append(targets, exports.Targets...)
+				diags = append(diags, newDiags...)
 			default:
-				diags.Add(diag.SeverityLevelWarn, fmt.Sprintf("unsupported service discovery %s was provided", sdConfig.Name()))
+				diags.Add(diag.SeverityLevelWarn, fmt.Sprintf("unsupported service discovery %s was provided", serviceDiscoveryConfig.Name()))
```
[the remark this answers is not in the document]
erikbaranowski: I anticipate a few scenarios with the conversion code.

You can run a mode that starts the agent immediately.
- By defaults errors/warnings will cause it to fail
- There will be a flag to allow warnings to be ignored

You can run a formatter that will output the river similar to our `fmt` command today
- Same idea on a flag about warnings I think

converter/internal/common/convert_targets.go
erikbaranowski: Agreed, I was thinking similar after matt's comment 👍

converter/internal/common/convert_targets.go
```diff
@@ -0,0 +1,67 @@
...
+				toks = append(toks, builder.Token{Tok: token.LCURLY, Lit: ""})
+				toks = append(toks, builder.Token{Tok: token.LITERAL, Lit: "\n"})
+				toks = append(toks, builder.Token{Tok: token.LITERAL, Lit: key})
+				toks = append(toks, builder.Token{Tok: token.ASSIGN, Lit: ""})
+				toks = append(toks, builder.Token{Tok: token.STRING, Lit: `"` + target + `"`})
+				toks = append(toks, builder.Token{Tok: token.COMMA, Lit: ""})
+				toks = append(toks, builder.Token{Tok: token.LITERAL, Lit: "\n"})
+				toks = append(toks, builder.Token{Tok: token.RCURLY, Lit: ""})
+				toks = append(toks, builder.Token{Tok: token.RBRACK, Lit: ""})
```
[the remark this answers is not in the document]
erikbaranowski: I copied other usage but good question. I will test it.

converter/internal/common/convert_targets.go
```diff
@@ -0,0 +1,67 @@
...
+	// We are relying on each targetMap having exactly 1 target which we control
+	// from the upstream converter code. We panic below if not.
+	if len(f.Targets) > 1 {
+		toks = append(toks, builder.Token{Tok: token.LITERAL, Lit: "concat"})
+		toks = append(toks, builder.Token{Tok: token.LPAREN, Lit: ""})
+	}
+
+	for ix, targetMap := range f.Targets {
+		if len(targetMap) != 1 {
+			panic("unexpected number of targets received on a target map")
```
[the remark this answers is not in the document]
erikbaranowski: This should never happen from the controlled converter code and is implementing an interface so I can't necessarily change the method signature. Alternatively, I could make it more robust to handle the looping regardless of targetMap length. It ups the complexity slightly but might be worth it.

converter/internal/prometheusconvert/prometheusconvert.go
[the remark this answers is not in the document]
erikbaranowski: Ya that sounds correct. I assume our fmt command does similar today though I haven't studied it yet.

converter/internal/common/convert_targets.go
erikbaranowski: A series of testing has been added to cover the scenarios.

erikbaranowski:
I've engaged with @clayton-cornell to start us thinking about how we want to organize these docs.

I believe we will have some initial docs when we wire up the command line. 

- #3737
- #3738 

And then the main docs are planned under #4073 

I don't anticipate docs detailed to the point of component specific information that need maintenance as we go.

mattdurham (approved):
LGTM

rfratto (commented):
Sorry, this has apparently been pending since you opened the PR but I forgot to hit submit.

PR #4118, opened 2023-06-09
prometheus.operator.servicemonitors: allow servicemonitors to set bearerTokenFile

### LINKED ISSUE #4099: Support bearerTokenFile property for prometheus.operator.*monitor components
### Request

Following #3915, I would like to be able to set the `bearerTokenFile` property in PodMonitor & ServiceMonitor resources.

As of now, only bearerTokenSecret is supported which requires to create the given secret beforehand.

### Use case

In case we don't want to create a dedicated secret with the bearer token, we want to use the pod token file located at `/var/run/secrets/kubernetes.io/serviceaccount/token` using the `bearerTokenFile` property

captncraig:
Fixes #4099 

This field got missed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,8 @@ Main (unreleased)
 
 - Fix oauth default scope in `loki.source.azure_event_hubs`. (@akselleirv)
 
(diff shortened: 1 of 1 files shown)


rfratto (approved):
LGTM but needs a changelog entry

Issue #3929, opened 2023-05-18
Add registry to helm chart


edwintye:
### Request

I want to have the ability to use a different registry from the default `docker.io`, say a private cache of the image in JFrog or Nexus. Seems like an easy enough change without breaking the default behaviour on [_agent.yaml#3](https://github.com/grafana/agent/blob/main/operations/helm/charts/grafana-agent/templates/containers/_agent.yaml#L3) to

```yaml
image: {{ .Values.image.registry}}/{{ .Values.image.repository }}:{{ include "grafana-agent.imageTag" . }}
```

### Use case

Fetching images from dockerhub has some well known issues these days and will increasingly be more challenging.  We just want to decrease the amount of traffic across the public internet when this problem can be resolved easily.

rfratto:
I don't think a dedicated registry field is necessary; to use a different registry you can add it to the repository: 

```yaml 
image:
  repository: quay.io/REPO/IMAGE_NAME:TAG # (or another registry)
```

edwintye:
Is that the preferred solution? Then you are forcing everyone to treat `repository` as `registry/repository` here which is perfectly fine by me.   Can someone changing the existing value from `grafana/agent` to `docker.io/grafana/agent` to help future readers to see that it is not a missing feature but intended behaviour?

rfratto:
> Is that the preferred solution?

In general I'd prefer not to add new fields if an existing field can be made to work, yeah. 

> Can someone change the existing value from grafana/agent to docker.io/grafana/agent to help future readers see that it is not a missing feature but intended behaviour (since the silent docker.io prepend is unique to docker)?

I'm ok with either making the default explicit to note that it's using docker.io or by documenting it to note that a different registry can be used.

(I think image.repository is a bad name, but that's the default name Helm generates so I'm not sure how common it is to change it to something else like image.name)

QuentinBisson:
`image.registry` and more specifically global.image.registry or `global.imageRegistry` is something that is commonly present in a lot of community helm charts as a default for a lot of different use cases like umbrella charts (using a global property https://github.com/prometheus-community/helm-charts/blob/30ab923a49d7c7b9434e2eabaac03baf6c226b1e/charts/kube-prometheus-stack/values.yaml#L187)

For what it's worth, even the loki or vertical pod autoscaler charts supports it https://github.com/grafana/helm-charts/blob/c756b95494544056e43098b9e8b695a8010c21f2/charts/loki-distributed/values.yaml#L4

[...]

PR #4127, opened 2023-06-11
Add image registry support

QuentinBisson:
#### PR Description

This PR adds support for image registry override and add a global field as well so the registry can be overriden is used in umbrella charts

Fixes https://github.com/grafana/agent/issues/3929

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated


TheoBrigitte (approved):
LGTM

operations/helm/charts/grafana-agent/values.yaml
```diff
@@ -5,6 +5,14 @@ nameOverride: null
 # resource names.
 fullnameOverride: null
 
+global:
```
rfratto, line 8: Can this get a comment explaining what it is and how it works?

I understand that it's convention with a few Helm charts, but I think the way these Helm charts use global sections is the inverse from what I would normally expect; I would normally expect global sections to provide the defaults, and then for other settings to override the global defaults.

Since this works in the opposite direction (globals override other settings), I'd like to add a comment making that clear.

rfratto (approved):
LGTM, will merge once my two final comments are addressed. 

I'll be honest, I'm not sure how I feel about this change. I re-read the conversation in #3929 a few times, but I can't help but feeling this is a Helm chart management issue and not a design issue with the Helm chart itself. 

That being said, my resistance may be because we really don't need this change internally, so maybe I just don't get how it's necessary or alleviates a ton of pain. Because I don't feel like I "get it" and, as you pointed out, other Helm charts use the same pattern, I'm not going to block this getting merged.

I'd like to revisit this decision in the future prior to marking the Helm chart as 1.0 to see how widely used these settings are to identify if we should continue to support them long term.

operations/helm/charts/grafana-agent/ci/local-image-registry-values.yaml
```diff
@@ -0,0 +1,7 @@
+# Test rendering of the chart with the individual image registries explicitly set to another value.
+image:
+  registry: quay.io
+
+configReloader:
+  image:
+    registry: quay.io
```
QuentinBisson, line 7:
```suggestion
    registry: quay.io

```

operations/helm/charts/grafana-agent/ci/global-image-registry-values.yaml
```diff
@@ -0,0 +1,11 @@
...
+global:
+  image:
+    registry: quay.io
+
+image:
+  registry: docker.com # Invalid value by default
+
+configReloader:
+  image:
+    registry: docker.com
```
QuentinBisson, line 11:
```suggestion
    registry: docker.com

```

operations/helm/charts/grafana-agent/values.yaml
QuentinBisson: I added a comment above global. For information, it works that way because in most of the non grafana upstream projects I've seen, the image registry is generally something other than docker like quay so the existing values are  already defined as:

```yaml
image:
  registry: quay.io
```
That means that if the "local" setting was taking precedent then the global value would never be applied.

PR #4138, opened 2023-06-12
Allow traces and metrics to propragate out of modules

### LINKED ISSUE #3107: Flow Modules: Add metrics if needed

mattdurham:
#### PR Description

This is a big PR but it has all been covered by previous merges into `module_staging` ideally this is double checking and then merge. The other primary goal is to add a wall between loading modules and exposing the inner workings of modules/flow. 

Closes #3107

#### Notes to the Reviewer

There is also some work on Mixins that may need to happen or at least be considered for long term.
There is another PR I will do after this one on removing `TestFlowLogger`.

#### PR Checklist

- [X] CHANGELOG updated
- [ ] Documentation added
- [X] Tests updated


mattdurham:
The only functional change from the `module_staging` is fixed a bug with removing the ID. Added a test to handle it.

tpaschalis:
@mattdurham Was this the PR that included the refactor around logging, or are we having any more logging-related changes coming up?

If not, I'm gonna redo #3566 to fit this new approach.

PR #4145, opened 2023-06-13
Support for all prometheus to Flow component conversion

### LINKED ISSUE #4018: Support converting Prometheus configs to supported Flow configs
Add support to the new Convert API for prometheus config to all existing Flow components.

erikbaranowski:
#### PR Description

Implement all of the remaining prometheus config component conversions

Closes #4018 

#### Notes to the Reviewer

Apologies for the large review. I didn't want to do 1 PR per discovery component but then things ballooned a bit. I'm available to discuss anything about the implementation to help clear anything up.

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [x] Tests updated


converter/internal/common/common.go
```diff
@@ -32,3 +39,38 @@ func getValueOverrideHook() builder.ValueOverrideHook {
...
+
+func GetUniqueLabel(label string, currentCount int) string {
+	if currentCount == 1 {
+		return label
+	}
+
+	return fmt.Sprintf("%s_%d", label, currentCount)
+}
+
+// prettyPrint attempts to pretty print the input slice. If prettyPrint fails,
```
mattdurham, line 51: nit: PrettyPrint instead of prettyPrint

converter/internal/common/common.go
```diff
@@ -32,3 +39,38 @@ func getValueOverrideHook() builder.ValueOverrideHook {
...
+	}
+
+	f, err := parser.ParseFile("", in)
+	if err != nil {
+		diags.Add(diag.SeverityLevelWarn, err.Error())
+		return in, diags
+	}
+
+	var buf bytes.Buffer
+	if err := printer.Fprint(&buf, f); err != nil {
```
mattdurham, line 68: This var shadowing looks wierd, can we use `err =`

converter/internal/common/common.go
```diff
@@ -32,3 +39,38 @@ func getValueOverrideHook() builder.ValueOverrideHook {
...
+		diags.Add(diag.SeverityLevelWarn, err.Error())
+		return in, diags
+	}
+
+	// Add a trailing newline at the end of the file, which is omitted by Fprint.
+	_, _ = buf.Write([]byte{'\n'})
```
mattdurham, line 74: WriteString may save a conversion though I suspect its the exact same performance, just a bit clearer.

converter/internal/common/common.go
```diff
@@ -32,3 +39,38 @@ func getValueOverrideHook() builder.ValueOverrideHook {
...
+func GetUniqueLabel(label string, currentCount int) string {
+	if currentCount == 1 {
+		return label
+	}
+
+	return fmt.Sprintf("%s_%d", label, currentCount)
+}
+
+// prettyPrint attempts to pretty print the input slice. If prettyPrint fails,
+// the input slice is returned unmodified.
```
mattdurham, line 52: Instead of saying `input slice` maybe add a note that it parses a river config.

converter/internal/common/convert_targets.go
```diff
@@ -19,17 +19,13 @@ var _ river.Capsule = ConvertTargets{}
...
+	targetCount := len(f.Targets)
 	if targetCount == 0 {
-		toks = append(toks, builder.Token{Tok: token.LBRACK})
-		toks = append(toks, builder.Token{Tok: token.RBRACK})
-		return toks
+		expr.SetValue(f.Targets)
+		return expr.Tokens()
 	}
 
 	if targetCount > 1 {
```
mattdurham, line 31: Is this always true?

converter/internal/common/convert_targets.go
```diff
@@ -38,22 +34,22 @@ func (f ConvertTargets) RiverTokenize() []builder.Token {
...
-				toks = append(toks, builder.Token{Tok: token.LITERAL, Lit: key})
-				toks = append(toks, builder.Token{Tok: token.ASSIGN})
-				toks = append(toks, builder.Token{Tok: token.STRING, Lit: `"` + target + `"`})
-				toks = append(toks, builder.Token{Tok: token.COMMA})
-				toks = append(toks, builder.Token{Tok: token.LITERAL, Lit: "\n"})
-				toks = append(toks, builder.Token{Tok: token.RCURLY})
-				toks = append(toks, builder.Token{Tok: token.RBRACK})
+		keyValMap := map[string]string{}
+		for key, val := range targetMap {
+			if key == "__expr__" {
```
mattdurham, line 39: `__address__` -> `__expr__` ? This looks to be some special handling for conversion but maybe spelling that out a bit more?

converter/internal/prometheusconvert/digitalocean.go
```diff
@@ -0,0 +1,63 @@
...
+		FollowRedirects: sdConfig.HTTPClientConfig.FollowRedirects,
+		EnableHTTP2:     sdConfig.HTTPClientConfig.EnableHTTP2,
+	}, validateDiscoveryDigitalOcean(sdConfig)
+}
+
+func validateDiscoveryDigitalOcean(sdConfig *promdigitalocean.SDConfig) diag.Diagnostics {
+	var diags diag.Diagnostics
+
+	if sdConfig.HTTPClientConfig.BasicAuth != nil {
+		diags.Add(diag.SeverityLevelWarn, "unsupported basic_auth for digitalocean_sd_configs")
```
mattdurham, line 44: This applies to all the checks like this and there might already be discussion on this, but is there a reason this is `warn` instead of `error`?

converter/internal/prometheusconvert/prometheusconvert.go
```diff
@@ -49,38 +42,96 @@ func Convert(in []byte) ([]byte, diag.Diagnostics) {
 		diags.Add(diag.SeverityLevelError, fmt.Sprintf("failed to render Flow config: %s", err.Error()))
 		return nil, diags
 	}
-	return buf.Bytes(), diags
+
+	if len(buf.Bytes()) > 0 {
```
mattdurham, line 46: Thoughts on inverting this to `if len(buf.bytes) == 0 { return nil, diags}`, that would match the pattern above of the ifs enclosing the non-happy path

converter/internal/prometheusconvert/prometheusconvert_test.go
```diff
@@ -36,17 +34,13 @@ func TestConvert(t *testing.T) {
 
 			t.Run(caseName, func(t *testing.T) {
 				actual, diags := prometheusconvert.Convert(inputBytes)
+				os.WriteFile("/home/erik/temp/"+caseName+".river", actual, 0666)
```
mattdurham, line 37: this looks pretty sus ;) Also we may want this to be linux only?

mattdurham (changes requested):
Overall solid, mostly minor nits with one or two legitimate questions. I only gave the testdate a cursory look.

converter/internal/prometheusconvert/prometheusconvert_test.go
erikbaranowski: Haha woops. Tests work in both windows/linux after normalizing the line endings

converter/internal/prometheusconvert/digitalocean.go
erikbaranowski: Unsupported configs are classified as warnings because we can do the conversion but those pieces are ignored. We will default to failing on warnings but allow overriding them by choice.

converter/internal/common/convert_targets.go
erikbaranowski: Will do, this is how we inject the string for other component exports during tokenize for this type.

converter/internal/common/common.go
mattdurham: IMO and outside this scope but I think we should start typedef river configs from []byte to `type RiverConfig []byte`

converter/internal/common/convert_targets.go
erikbaranowski: not if targetCount == 1 we can include with without the `concat()`

converter/internal/prometheusconvert/testdata/azure.river
```diff
@@ -0,0 +1,53 @@
...
+	}
+	follow_redirects = true
+	enable_http2     = true
+}
+
+prometheus.scrape "prometheus1" {
+	targets = concat([{
+		__address__ = "localhost:9090",
+	}],
+		discovery.azure.prometheus1.targets)
```
rfratto, line 34: This looks weird, is this being formatted properly? 

I think maybe if we're concating multiple things, each set of targets should be on their own runs of lines, so that it's this instead:

```river 
targets = concat(
  [{__address__ = "localhost:9090"}], 
  discovery.azure.prometheus1.targets,
)
```

converter/internal/prometheusconvert/testdata/consul.river
```diff
@@ -0,0 +1,45 @@
+prometheus.remote_write "default" {
+	external_labels = {}
```
rfratto, line 2: I don't think this should be set. 

It looks like we always return a non-nil map for this: 

https://github.com/grafana/agent/blob/a95a2bd2eb68a0b6ef692d29f21f89dd1b64bb80/converter/internal/prometheusconvert/remote_write.go#L23C18-L23C64

I think it should be nil if the map is empty so this doesn't print.

converter/internal/prometheusconvert/testdata/discovery.river
```diff
@@ -0,0 +1,69 @@
+prometheus.remote_write "default" {
```
rfratto, line 1: It would be nice if we had a little more control over the order these converted in, so that it's roughly:

1. Discovery component(s) 
2. Discovery relabel component(s) (if any)
3. Prometheus scrape component(s) 
4. Prometheus relabel component(s) (if any) 
5. Prometheus remote_write

If it's too much effort, it's probably not worth it right now.

converter/internal/prometheusconvert/testdata/discovery.river
```diff
@@ -0,0 +1,69 @@
...
+
+	managed_identity {
+		client_id = "client"
+	}
+	follow_redirects = true
+	enable_http2     = true
+}
+
+discovery.relabel "prometheus1" {
+	targets = discovery.azure.prometheus1.targets
```
rfratto, line 47: Static targets also are subject to discovery.relabel, so this should be:

```river 
targets = concat(
  [{ __address__ = "localhost:9090" }],
  discovery.azure.prometheus1.targets,
)
```

converter/internal/prometheusconvert/testdata/discovery.river
```diff
@@ -0,0 +1,69 @@
...
+		regex         = "^(?:(.*))$"
+		target_label  = "__param_target2"
+	}
+}
+
+prometheus.scrape "prometheus1" {
+	targets = concat([{
+		__address__ = "localhost:9090",
+	}],
+		discovery.relabel.prometheus1.targets)
```
rfratto, line 66: See the comment above; this should be: 

```river 
targets = discovery.relabel.prometheus1.targets
```

converter/internal/prometheusconvert/testdata/discovery_relabel.river
```diff
@@ -0,0 +1,78 @@
...
+		}
+
+		metadata_config {
+			max_samples_per_send = 500
+		}
+	}
+}
+
+discovery.relabel "prometheus1" {
+	targets = []
```
rfratto, line 21: This is wrong, but it'll be fixed once you address the static targets issue

converter/internal/prometheusconvert/testdata/discovery_relabel.yaml
```diff
@@ -0,0 +1,26 @@
+scrape_configs:
+  - job_name: "prometheus1"
+    static_configs:
+      - targets: ["localhost:9090"]
+    relabel_configs:
+      - source_labels: [__address1__]
+        target_label: __param_target1
+      - source_labels: [__address2__]
+        target_label: __param_target2
```
rfratto, line 9: It'd be great if we could include a relabel rule with a regular expression which uses escaped characters, especially an escaped double quote. 

I'd like to verify that it won't generate an invalid River string, and that the escaping is retained.

converter/internal/prometheusconvert/testdata/scrape.yaml
```diff
@@ -11,12 +11,16 @@ scrape_configs:
     scrape_timeout: 5s
     static_configs:
       - targets: ["localhost:9090"]
```
rfratto, line 13: Can you make this a list of two targets to make sure that both of them get the app=foo label?

```yaml
static_configs:
- targets: ["localhost:9090", "localhost:9091"] 
  labels:
    app: 'foo'
```

converter/internal/prometheusconvert/testdata/ec2.yaml
```diff
@@ -0,0 +1,19 @@
+scrape_configs:
+  - job_name: "prometheus1"
+    static_configs:
+      - targets: ["localhost:9090"]
+    ec2_sd_configs:
+      - region: 'us-east-1'
+        access_key: 'YOUR_ACCESS_KEY'
+        secret_key: 'YOUR_SECRET_KEY'
+        port: 8080
```
rfratto, line 9: Do we have any of these tests which validate that having multiple instances of the same SD generates a unique component name?

rfratto (approved):
LGTM! Large chunk of feedback though :)

converter/converter.go
```diff
@@ -35,35 +32,10 @@ const (
 func Convert(in []byte, kind Input) ([]byte, diag.Diagnostics) {
 	switch kind {
 	case InputPrometheus:
-		return prettyPrint(prometheusconvert.Convert(in))
+		return prometheusconvert.Convert(in)
```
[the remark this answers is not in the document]
erikbaranowski: I'm going to leave it for now so that tests are using it and consider formatting part of the internal APIs responsibility.

converter/internal/prometheusconvert/kubernetes.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	return &kubernetes.Arguments{
+		APIServer:          config.URL(sdConfig.APIServer),
+		Role:               string(sdConfig.Role),
+		KubeConfig:         sdConfig.KubeConfig,
+		HTTPClientConfig:   *toHttpClientConfig(&sdConfig.HTTPClientConfig),
+		NamespaceDiscovery: *toNamespaceDiscovery(&sdConfig.NamespaceDiscovery),
+		Selectors:          *toSelectorConfig(&sdConfig.Selectors),
+		AttachMetadata:     *toAttachMetadata(&sdConfig.AttachMetadata),
```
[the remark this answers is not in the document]
erikbaranowski: Updated except toHttpClientConfig which is used elsewhere differently.

converter/internal/prometheusconvert/testdata/azure.river
rfratto: Though if there's more than one label for the static object it should probably be

```river 
targets = concat(
  [{
     __address__ = "localhost:9090",
     foo         = "bar",
  }], 
  discovery.azure.prometheus1.targets,
)
```

mattdurham (approved):
Awesome!

converter/internal/prometheusconvert/testdata/discovery.river
erikbaranowski: This has been implemented a couple ways in the past 2 commits. The most recent 1 will organize the blocks into categories so they don't need to be sorted at the end and get appended to the file as we want them.

converter/internal/common/common.go
```diff
@@ -14,13 +14,13 @@ import (
...
-// value override hook to the file we are building.
-func AppendBlockWithOverride(f *builder.File, name []string, label string, args component.Arguments) {
+// AppendNewBlockWithOverride appends Flow component arguments using the convert
+// value override hook to a slice of block pointers.
+func AppendNewBlockWithOverride(blocks []*builder.Block, name []string, label string, args component.Arguments) []*builder.Block {
 	block := builder.NewBlock(name, label)
 	block.Body().SetValueOverrideHook(getValueOverrideHook())
 	block.Body().AppendFrom(args)
-	f.Body().AppendBlock(block)
+	return append(blocks, block)
```
rfratto, line 23: Given the pattern of this now, I'm not sure it makes sense for this to return the new set of blocks rather than just creating the block and having the caller do the append

rfratto (approved):

converter/internal/prometheusconvert/testdata/azure.river
erikbaranowski: I fixed up the multiline concat formatting. The single line address example would require changes to how slice expressions are tokenized in general unless I don't reuse the generic method:

```golang
expr.SetValue([]map[string]string{keyValMap})
toks = append(toks, expr.Tokens()...)
```

rfratto: That's fine, the single line form isn't necessary. The output looks much, much better now. Thanks!

PR #4147, opened 2023-06-13
misc: remove dependency on k3d

rfratto:
This commit removes the dependency on k3d, replacing it with the existing dependency on testcontainers-go and the k3s module.

k3d is a very large module to have a dependency just for tests; it makes makes updating other dependencies more difficult. Switching it with k3d and testcontainers-go ensures that our direct dependencies have less overlap so updating things is easier.


pkg/util/k8s/k8s.go
```diff
@@ -4,45 +4,23 @@ package k8s
...
 // the cluster.
 //
-// Cluster also runs an NGINX ingress controller which is exposed to the host.
-// Call GetHTTPAddr to get the address for making requests against the server.
-//
-// Set K8S_USE_DOCKER_NETWORK in your environment variables if you are
-// running tests from inside of a Docker container. This environment variable
-// configures the k3s Docker container to join the same network as the
-// container tests are running in. When this environment variable isn't set,
-// the exposed ports on the Docker host are used for cluster communication.
```
rfratto, line 44: This is no longer required because testcontainers-go has logic to figure out if it's running in Docker automatically.

ptodev (approved):

spartan0x117 (approved):
LGTM

Issue #3923, opened 2023-05-18
Flow: Display metrics from Collector components in the Agent's metrics endpoint


ptodev:
### Request

In static mode, if we use the Otel batch processor we can see metrics such as the `traces_processor_batch_batch_send_size` histogram. This metric comes from within the Collector's codebase and it is not visible on the Agent's `/metrics` endpoint when the Agent runs in Flow mode.  

According to @rfratto that's because the version of Otel we embed doesn't support using the Otel SDK for processors, and only uses OpenCensus. 

We need to make such metrics visible in Flow mode and this should ideally work out of the box when we add new Otel components into Flow without extra work just to get metrics integrated.

### Use case

Feature parity between Static and Flow mode.

rfratto:
👍 

I'd like to note that this will happen automatically once we upgrade the OpenTelemetry Collector SDK, since newer versions support using the OpenTelemetry SDK for processor telemetry (but not the version we currently import).

ptodev:
I suspect that even after we upgrade to the latest Collector version we will still not see a lot of new metrics. 

I believe most components in the Contrib codebase use OpenCensus for metrics. For example, the Kafka receiver which we added to Flow recently also [uses](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/kafkareceiver/metrics.go) OpenCensus.

rfratto:
That's pretty unfortunate, but I think there's plans for the OTel SDK to be adopted throughout the collector, which would address that issue. 

In the meantime, it looks like it may be possible for us to create a shim to get OpenCensus metrics per component without exposing more information than we'd want. However, it also looks like that would require manual work so it'd be something we'd opt each component into.

rfratto:
@ptodev Is this a blocker for feature parity? Can we remove the feature-parity label?

ptodev:
I think we can close the whole issue, since we no longer use the OpenCensus metrics in static mode. In theory, the Otel metrics in static mode should be the same as the Otel metrics in flow mode.

rfratto:
Closing because OpenCensus metrics are not being used in static mode.

Issue #2295, opened 2022-10-03
Flow: otelcol.processor.spanmetrics component


rfratto:
Expose the OpenTelemtry Collector `spanmetrics` processor as a Flow component called `otelcol.processor.spanmetrics`. This component is from the `otelcol-contrib` distribution.

rfratto:
We should probably _not_ port this over from static mode, as a span2metrics connector component is being planned upstream for OpenTelemetry Collector, and we should probably use that instead.

mar4uk:
@ptodev Was `spanmetrics` component added to Agent Flow components? I can't find it in https://github.com/grafana/agent/tree/main/component/otelcol/processor

ptodev:
Hi @mar4uk 👋  It's not added yet, but it's the next one on my list, after #4261 and #2300
I'm sorry for closing this issue - not sure why this is, it must have been a mistake. Probably got mixed up with #2612

PR #4148, opened 2023-06-13
Add a otelcol.exporter.loadbalancing component

### LINKED ISSUE #4068: Flow: otelcol.exporter.loadbalancing component
### Request

Expose the OpenTelemtry Collector [loadbalancing](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/loadbalancingexporter) exporter as a Flow component called `otelcol.exporter.loadbalancing`. This component is from the otelcol-contrib distribution.

### Use case

Feature parity with Static mode

ptodev:
#### PR Description

This PR is mostly done, with the exception of the items listed in the "notes to the Reviewer" section. 

For now the main thing I'd appreciate feedback on is the `endpoint` attribute in the client block. Could we make it mandatory for all components except this, without copying the whole struct? And would you prefer copying the struct to making it an optional argument, as is done in this PR?

Having it as an optional argument is more in line with the Collector, but I fear that it goes too much agains the Agent way of doing things. It's a "three steps back and one step forward" type situation :)

#4068

#### Notes to the Reviewer

There are a few controversial things that I'm not sure about:

* The `endpoint` attribute in the gRPC client block used to be mandatory and I changed it to optional. I couldn't think of a clean way to have it be optional just for that component. If you can think of one, please let me know because it'd make more sense to have a mandatory `endpoint` for the other components.
* According to the official Collector [docs](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/loadbalancingexporter), this exporter also works for logs? See the table at the top of the Collector's docs. I didn't mention this in our docs. I'll double check if this is the case.
* This component also uses [metrics](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/loadbalancingexporter#metrics). However, I didn't list them in the Agent docs because I'm not sure if they work. I'll try running the collector locally to make sure we can see them.

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,8 @@ Main (unreleased)
     cluster and scrape the targets they reference. (@captncraig)
   - `pyroscope.ebpf` collects system-wide performance profiles from the current
     host (@korniltsev)
+  - `otelcol.exporter.loadbalancing` - export traces and logs to multiple OTLP gRPC 
+    endpoints in a load-balanced way. (@ptodev)
 
 - New Grafana Agent Flow command line utilities:
 
--- a/component/otelcol/config_tls.go
+++ b/component/otelcol/config_tls.go
@@ -70,31 +70,17 @@ func (args *TLSSetting) Convert() *otelconfigtls.TLSSetting {
 		return nil
 	}
 
-	res := &otelconfigtls.TLSSetting{
+	return &otelconfigtls.TLSSetting{
+		CAPem:          configopaque.String(args.CA),
 		CAFile:         args.CAFile,
+		CertPem:        configopaque.String(args.Cert),
 		CertFile:       args.CertFile,
+		KeyPem:         configopaque.String(string(args.Key)),
 		KeyFile:        args.KeyFile,
 		MinVersion:     args.MinVersion,
 		MaxVersion:     args.MaxVersion,
 		ReloadInterval: args.ReloadInterval,
 	}
-
-	// Collector sets the pem attributes to nil by default,
-	// so let's match this.
-
-	if args.CA != "" {
-		res.CAPem = configopaque.String(args.CA)
-	}
-
-	if args.Cert != "" {
-		res.CertPem = configopaque.String(args.Cert)
-	}
-
-	if args.Key != "" {
-		res.KeyPem = configopaque.String(args.Key)
-	}
-
-	return res
 }
 
 // Validate implements river.Validator.
--- a/component/otelcol/config_tls.go
+++ b/component/otelcol/config_tls.go
@@ -70,17 +70,31 @@ func (args *TLSSetting) Convert() *otelconfigtls.TLSSetting {
 		return nil
 	}
 
-	return &otelconfigtls.TLSSetting{
-		CAPem:          configopaque.String(args.CA),
+	res := &otelconfigtls.TLSSetting{
 		CAFile:         args.CAFile,
-		CertPem:        configopaque.String(args.Cert),
 		CertFile:       args.CertFile,
-		KeyPem:         configopaque.String(string(args.Key)),
 		KeyFile:        args.KeyFile,
 		MinVersion:     args.MinVersion,
 		MaxVersion:     args.MaxVersion,
 		ReloadInterval: args.ReloadInterval,
 	}
+
+	// Collector sets the pem attributes to nil by default,
+	// so let's match this.
+
+	if args.CA != "" {
+		res.CAPem = configopaque.String(args.CA)
+	}
+
+	if args.Cert != "" {
+		res.CertPem = configopaque.String(args.Cert)
+	}
+
+	if args.Key != "" {
+		res.KeyPem = configopaque.String(args.Key)
+	}
+
+	return res
 }
 
 // Validate implements river.Validator.
--- a/component/otelcol/exporter/loadbalancing/loadbalancing.go
+++ b/component/otelcol/exporter/loadbalancing/loadbalancing.go
@@ -1,0 +1,250 @@
+// Package loadbalancing provides an otelcol.exporter.loadbalancing component.
+package loadbalancing
+
+import (
+	"time"
+
+	"github.com/alecthomas/units"
+	"github.com/grafana/agent/component"
+	"github.com/grafana/agent/component/otelcol"
+	"github.com/grafana/agent/component/otelcol/auth"
+	"github.com/grafana/agent/component/otelcol/exporter"
+	"github.com/grafana/agent/pkg/river"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/exporter/loadbalancingexporter"
+	otelcomponent "go.opentelemetry.io/collector/component"
+	otelconfigauth "go.opentelemetry.io/collector/config/configauth"
+	otelconfiggrpc "go.opentelemetry.io/collector/config/configgrpc"
+	"go.opentelemetry.io/collector/config/configopaque"
+	"go.opentelemetry.io/collector/exporter/exporterhelper"
+	"go.opentelemetry.io/collector/exporter/otlpexporter"
+	otelextension "go.opentelemetry.io/collector/extension"
+)
+
+func init() {
+	component.Register(component.Registration{
+		Name:    "otelcol.exporter.loadbalancing",
+		Args:    Arguments{},
+		Exports: otelcol.ConsumerExports{},
+
+		Build: func(opts component.Options, args component.Arguments) (component.Component, error) {
+			fact := loadbalancingexporter.NewFactory()
+			return exporter.New(opts, fact, args.(Arguments))
+		},
+	})
+}
+
+// Arguments configures the otelcol.exporter.loadbalancing component.
+type Arguments struct {
+	Protocol   Protocol         `river:"protocol,block"`
+	Resolver   ResolverSettings `river:"resolver,block"`
+	RoutingKey string           `river:"routing_key,attr,optional"`
+}
+
+var (
+	_ exporter.Arguments = Arguments{}
+	_ river.Defaulter    = &Arguments{}
+)
+
+// DefaultArguments holds default values for Arguments.
+var DefaultArguments = Arguments{
+	RoutingKey: "traceID",
+}
+
+// SetToDefault implements river.Defaulter.
+func (args *Arguments) SetToDefault() {
+	*args = DefaultArguments
+}
+
+// Convert implements exporter.Arguments.
+func (args Arguments) Convert() (otelcomponent.Config, error) {
+	return &loadbalancingexporter.Config{
+		Protocol:   args.Protocol.Convert(),
+		Resolver:   args.Resolver.Convert(),
+		RoutingKey: args.RoutingKey,
+	}, nil
+}
+
+// Protocol holds the individual protocol-specific settings. Only OTLP is supported at the moment.
+type Protocol struct {
+	OTLP OtlpConfig `river:"otlp,block"`
+}
+
+func (protocol Protocol) Convert() loadbalancingexporter.Protocol {
+	return loadbalancingexporter.Protocol{
+		OTLP: protocol.OTLP.Convert(),
+	}
+}
+
+// OtlpConfig defines the config for an OTLP exporter
+type OtlpConfig struct {
+	Timeout time.Duration          `river:"timeout,attr,optional"`
+	Queue   otelcol.QueueArguments `river:"queue,block,optional"`
+	Retry   otelcol.RetryArguments `river:"retry,block,optional"`
+	// Most of the time, the user will not have to set anything in the client block.
+	// However, the block should not be "optional" so that the defaults are populated.
+	Client GRPCClientArguments `river:"client,block"`
+}
+
+func (otlpConfig OtlpConfig) Convert() otlpexporter.Config {
+	return otlpexporter.Config{
+		TimeoutSettings: exporterhelper.TimeoutSettings{
+			Timeout: otlpConfig.Timeout,
+		},
+		QueueSettings:      *otlpConfig.Queue.Convert(),
+		RetrySettings:      *otlpConfig.Retry.Convert(),
+		GRPCClientSettings: *otlpConfig.Client.Convert(),
+	}
+}
+
+// ResolverSettings defines the configurations for the backend resolver
+type ResolverSettings struct {
+	Static *StaticResolver `river:"static,block,optional"`
+	DNS    *DNSResolver    `river:"dns,block,optional"`
+}
+
+func (resolverSettings ResolverSettings) Convert() loadbalancingexporter.ResolverSettings {
+	res := loadbalancingexporter.ResolverSettings{}
+
+	if resolverSettings.Static != nil {
+		staticResolver := resolverSettings.Static.Convert()
+		res.Static = &staticResolver
+	}
+
+	if resolverSettings.DNS != nil {
+		dnsResolver := resolverSettings.DNS.Convert()
+		res.DNS = &dnsResolver
+	}
+
+	return res
+}
+
+// StaticResolver defines the configuration for the resolver providing a fixed list of backends
+type StaticResolver struct {
+	Hostnames []string `river:"hostnames,attr"`
+}
+
+func (staticResolver StaticResolver) Convert() loadbalancingexporter.StaticResolver {
+	return loadbalancingexporter.StaticResolver{
+		Hostnames: staticResolver.Hostnames,
+	}
+}
+
+// DNSResolver defines the configuration for the DNS resolver
+type DNSResolver struct {
+	Hostname string        `river:"hostname,attr"`
+	Port     string        `river:"port,attr,optional"`
+	Interval time.Duration `river:"interval,attr,optional"`
+	Timeout  time.Duration `river:"timeout,attr,optional"`
+}
+
+var (
+	_ river.Defaulter = &DNSResolver{}
+)
+
+// DefaultDNSResolver holds default values for DNSResolver.
+var DefaultDNSResolver = DNSResolver{
+	Port:     "4317",
+	Interval: 5 * time.Second,
+	Timeout:  1 * time.Second,
+}
+
+// SetToDefault implements river.Defaulter.
+func (args *DNSResolver) SetToDefault() {
+	*args = DefaultDNSResolver
+}
+
+func (dnsResolver *DNSResolver) Convert() loadbalancingexporter.DNSResolver {
+	return loadbalancingexporter.DNSResolver{
+		Hostname: dnsResolver.Hostname,
+		Port:     dnsResolver.Port,
+		Interval: dnsResolver.Interval,
+		Timeout:  dnsResolver.Timeout,
+	}
+}
+
+// Extensions implements exporter.Arguments.
+func (args Arguments) Extensions() map[otelcomponent.ID]otelextension.Extension {
+	return args.Protocol.OTLP.Client.Extensions()
+}
+
+// Exporters implements exporter.Arguments.
+func (args Arguments) Exporters() map[otelcomponent.DataType]map[otelcomponent.ID]otelcomponent.Component {
+	return nil
+}
+
+// GRPCClientArguments is the same as otelcol.GRPCClientArguments, but without an "endpoint" attribute
+type GRPCClientArguments struct {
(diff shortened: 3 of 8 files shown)


docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
```diff
@@ -0,0 +1,256 @@
+---
+title: otelcol.exporter.loadbalancing
+labels:
+  stage: beta
```
ptodev, line 4: I labeled this as "beta" because the Otel Collector [also](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/loadbalancingexporter) labeled it as beta

docs/sources/flow/reference/components/otelcol.exporter.jaeger.md
```diff
@@ -62,7 +62,7 @@ The following arguments are supported:
 
 Name | Type | Description | Default | Required
 ---- | ---- | ----------- | ------- | --------
-`endpoint` | `string` | `host:port` to send telemetry data to. | | yes
+`endpoint` | `string` | `host:port` to send telemetry data to. | `""` | no
```
ptodev, line 65: I don't like this. IF we can avoid it while keeping the code clean, I'd prefer it. But not sure what is the best way? Copying the GRPCClientArguments struct seems like the simplest way to go.

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
```diff
@@ -0,0 +1,256 @@
...
+## Arguments
+
+`otelcol.exporter.loadbalancing` supports the following arguments:
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`routing_key` | `string` | Routing strategy for load balancing. | `"traceID"` | no
+
+The `routing_key` attribute determines how to route spans across endpoints. Its value could be one of the following:
+* `"service"`: exports spans based on their service name. This is useful when using processors like the span metrics,
```
ptodev, line 48: We don't [yet](https://github.com/grafana/agent/issues/2295) have a span metrics processor :) But I added this comment because it was mentioned in the Collector docs and seems useful

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
```diff
@@ -0,0 +1,256 @@
...
+
+otelcol.exporter.loadbalancing "default" {
+    resolver {
+        static {
+            hostnames = ["localhost:55690", "localhost:55700"]
+        }
+    }
+    protocol {
+        otlp {
+            client {}
```
ptodev, line 252: It's a bit weird how we have to add all these blocks just so `client {}` gets populated with defaults. IMO the user would not expect that. They would expect not having blocks would simply init them to the default values mentioned in the docs. Instead, they'd be zero-initialised. It feels a bit wrong to have to make the blocks mandatory just so they get their default values, and I might raise an issue about this so we can discuss it separately.

The problem is that if they get set to their default values, what should we set the mandatory values to? I guess the argument for not setting defaults is that mandatory attributes are not set either. But then probably nothing should be set and the whole block should be a nil.

docs/sources/flow/reference/components/otelcol.exporter.jaeger.md
rfratto: Yeah, I would personally copy the GRPCClientArguments for now until a better solution shows up; I don't think marking them all as optional for this one component should be the path we take.

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
```diff
@@ -0,0 +1,263 @@
...
+The actual backend load isn't taken into consideration. Even though this load-balancer won't do 
+round-robin balancing of the batches, the load distribution should be very similar among backends 
+with a standard deviation under 5% at the current configuration.
+
+`otelcol.exporter.loadbalancing` is especially useful for backends configured with tail-based samplers
+which make a decision based on the view of the full trace.
+
+When a list of backends is updated, around `1/n` of the space will be changed, so that the same trace ID 
+might be directed to a different backend, where `n` is the number of backends. This should be stable enough 
+for most cases, and the higher the number of backends, the less disruption it should cause.
```
ptodev, line 31: The otel [docs](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.63.0/exporter/loadbalancingexporter) also mention this:

> Still, if routing stability is important for your use case and your list of backends are constantly changing, consider using the groupbytrace processor. This way, traces are dispatched atomically to this exporter, and the same decision about the backend is made for the trace as a whole.

However I don't want to include this, because we don't have a groupbytrace processor and it'd be confusing to mention it imo.

ptodev:
I think all changes are finalised now, and the PR is ready for merging as long as it's approved. A few interesting points:

* As expected, the internal metrics from this component don't work. I didn't mention them in the docs. We have #3923 to fix this for all Flow components.
* The gRPC struct was copied and the `endpoint` attribute removed
* Apparently this exporter does work for logs, so I mentioned that in the docs

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
```diff
@@ -0,0 +1,263 @@
...
+
+## Exported fields
+
+The following fields are exported and can be referenced by other components:
+
+Name | Type | Description
+---- | ---- | -----------
+`input` | `otelcol.Consumer` | A value that other components can use to send telemetry data to.
+
+`input` accepts `otelcol.Consumer` data for telemetry signals of these types:
```
ptodev, line 223: This way of presenting the exported fields is not standard. I customised it to highlight that only logs and traces work. Maybe I also mention unsupported types like this:

> These signal types are not supported:
> * metrics

We also have #4151 for this presentation issue

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
clayton-cornell, line 31: I'm a little fuzzy on what we are trying to say here... 

I'm still not sure about the wording (still passive) but... is this what you're stying?

> When a list of backends is updated, around `1/n` of the space will be changed, where `n` is the number of backends. After the update, it's possible that a trace ID could be directed to a different backend. 

And... what disruption will this cause. What's the impact? Do we need to consider and document the impact of trace IDs moving around?

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
```diff
@@ -0,0 +1,263 @@
...
+
+The `routing_key` attribute determines how to route signals across endpoints. Its value could be one of the following:
+* `"service"`: exports spans based on their service name. This is useful when using processors like the span metrics, 
+so all spans for each service are sent to consistent collector instances for metric collection. 
+Otherwise, metrics for the same services would be sent to different collectors, making aggregations inaccurate.
+* `"traceID"`: exports spans based on their traceID.
+
+## Blocks
+
+The following blocks are supported inside the definition of
```
clayton-cornell, line 64:
```suggestion
The following blocks are supported in the definition of
```

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
```diff
@@ -0,0 +1,263 @@
...
+`balancer_name` | `string` | Which gRPC client-side load balancer to use for requests. | | no
+`auth` | `capsule(otelcol.Handler)` | Handler from an `otelcol.auth` component to use for authenticating requests. | | no
+
+{{< docs/shared lookup="flow/reference/components/otelcol-compression-field.md" source="agent" >}}
+
+The `balancer_name` argument controls what client-side load balancing mechanism
+to use. See the gRPC documentation on [Load balancing][] for more information.
+When unspecified, `pick_first` is used.
+
+An HTTP proxy can be configured through the following environment variables:
```
clayton-cornell, line 158:
```suggestion
You can configure an HTTP proxy with the following environment variables:
```

rfratto:
I'm removing my request for a reviewer because I unfortunately don't have the time to review this right now. I'm asking around if someone else can take a look, though!

component/otelcol/exporter/loadbalancing/loadbalancing.go
```diff
@@ -0,0 +1,246 @@
...
+func (resolverSettings ResolverSettings) Convert() loadbalancingexporter.ResolverSettings {
+	res := loadbalancingexporter.ResolverSettings{}
+
+	if resolverSettings.Static != nil {
+		staticResolver := resolverSettings.Static.Convert()
+		res.Static = &staticResolver
+	}
+
+	if resolverSettings.DNS != nil {
+		dnsResover := resolverSettings.DNS.Convert()
```
mattdurham, line 114: nit: `dnsResolver`

component/otelcol/config_tls.go
```diff
@@ -69,17 +69,31 @@ func (args *TLSSetting) Convert() *otelconfigtls.TLSSetting {
...
+	if args.CA != "" {
+		res.CAPem = []byte(args.CA)
+	}
+
+	if args.Cert != "" {
+		res.CertPem = []byte(args.Cert)
+	}
+
+	if args.Key != "" {
+		res.KeyPem = []byte(string(args.Key))
```
mattdurham, line 93: nit: You shouldnt need the string conversion.

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
```diff
@@ -0,0 +1,263 @@
...
+
+Note that either a trace ID or a service name is used for the decision on which backend to use. 
+The actual backend load isn't taken into consideration. Even though this load-balancer won't do 
+round-robin balancing of the batches, the load distribution should be very similar among backends 
+with a standard deviation under 5% at the current configuration.
+
+`otelcol.exporter.loadbalancing` is especially useful for backends configured with tail-based samplers
+which make a decision based on the view of the full trace.
+
+When a list of backends is updated, around `1/n` of the space will be changed, so that the same trace ID
```
mattdurham, line 29: Agree with Clayton here, I am not sure what this means. What's a space? I think I can infer what it means but more text here would be nice.

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
```diff
@@ -0,0 +1,263 @@
...
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`routing_key` | `string` | Routing strategy for load balancing. | `"traceID"` | no
+
+The `routing_key` attribute determines how to route signals across endpoints. Its value could be one of the following:
+* `"service"`: exports spans based on their service name. This is useful when using processors like the span metrics, 
+so all spans for each service are sent to consistent collector instances for metric collection. 
+Otherwise, metrics for the same services would be sent to different collectors, making aggregations inaccurate.
+* `"traceID"`: exports spans based on their traceID.
```
mattdurham, line 60: Can we get more detail on how this is handled? Or a link to otel docs?

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
```diff
@@ -0,0 +1,263 @@
...
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`hostnames` | `list(string)` | List of endpoints to export to. |  | yes
+
+### dns block
+
+The `dns` block periodically resolves an IP address via the DNS `hostname` attribute. This IP address 
+and the port specified via the `port` attribute will then be used by the gRPC exporter 
+as the endpoint to which to send data to.
```
mattdurham, line 115: There might be some inconsistent verbiage, `export to` and `send to`.

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
```diff
@@ -0,0 +1,263 @@
...
+
+The following arguments are supported:
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`hostname` | `string`   | DNS hostname to resolve. |  | yes
+`port`     | `string`   | Port to be used with the resolved IP address. | `"4317"` | no
```
mattdurham, line 122: Should this be expanded out, because at first glance it looks like the hostname port. `Port to be used with the IP addresses returned from the DNS hostname`?

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
```diff
@@ -0,0 +1,263 @@
...
+The following arguments are supported:
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`compression` | `string` | Compression mechanism to use for requests. | `"gzip"` | no
+`read_buffer_size` | `string` | Size of the read buffer the gRPC client to use for reading server responses. | | no
+`write_buffer_size` | `string` | Size of the write buffer the gRPC client to use for writing requests. | `"512KiB"` | no
+`wait_for_ready` | `boolean` | Waits for gRPC connection to be in the `READY` state before sending data. | `false` | no
+`headers` | `map(string)` | Additional headers to send with the request. | `{}` | no
+`balancer_name` | `string` | Which gRPC client-side load balancer to use for requests. | | no
```
mattdurham, line 149: Isn't the default here `pick_first`?

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
```diff
@@ -0,0 +1,263 @@
...
+## Exported fields
+
+The following fields are exported and can be referenced by other components:
+
+Name | Type | Description
+---- | ---- | -----------
+`input` | `otelcol.Consumer` | A value that other components can use to send telemetry data to.
+
+`input` accepts `otelcol.Consumer` data for telemetry signals of these types:
+* logs
```
mattdurham, line 224: Specify otel logs, since we also support loki?

component/otelcol/exporter/loadbalancing/loadbalancing.go
```diff
@@ -0,0 +1,246 @@
...
+	//TODO: Do we need to set these extensions?
+	return args.Protocol.OTLP.Client.Extensions()
+}
+
+// Exporters implements exporter.Arguments.
+func (args Arguments) Exporters() map[otelconfig.DataType]map[otelconfig.ComponentID]otelcomponent.Exporter {
+	return nil
+}
+
+// GRPCClientArguments is the same as otelcol.GRPCClientArguments, but without an "endpoint" attribute
```
mattdurham, line 176: I think this is fine since it is fundamentally different and we dont have a way to remove parameters.

CHANGELOG.md
```diff
@@ -24,7 +24,7 @@ Main (unreleased)
   - `prometheus.exporter.kafka` collects metrics from Kafka Server (@oliver-zhang)
   - `otelcol.processor.attributes` accepts telemetry data from other `otelcol`
     components and modifies attributes of a span, log, or metric. (@ptodev)
-
+  - `otelcol.exporter.loadbalancing` - export traces to multiple OTLP gRPC endpoints in a load-balanced way. (@ptodev)
```
mattdurham, line 27: and logs

mattdurham (changes requested):

mattdurham:
Overall looks solid, a few nits and requests for more clarification and documentation.

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
```diff
@@ -0,0 +1,263 @@
...
+{{< docs/shared lookup="flow/stability/beta.md" source="agent" >}}
+
+`otelcol.exporter.loadbalancing` accepts logs and/or traces from other `otelcol` components
+and writes them over the network using the OpenTelemetry Protocol (OTLP) protocol. 
+
+> **NOTE**: `otelcol.exporter.loadbalancing` is a wrapper over the upstream
+> OpenTelemetry Collector `loadbalancing` exporter. Bug reports or feature requests will
+> be redirected to the upstream repository, if necessary.
+
+Multiple `otelcol.exporter.loadbalancing` components can be specified by giving them
```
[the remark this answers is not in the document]
ptodev: This is actually a standard line which we put in a lot of docs pages. I'm not against changing it everywhere, but for the sake of consistency I'd prefer to leave it as it is in this page for now. Maybe it can be changed for all components in a separate PR.

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
ptodev: This is also a standard line which we put in many docs pages. If we are to change it, I think it should be done for all pages in a separate PR.

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
ptodev: I agree that it is confusing. I just copy pasted it from the original Otel docs because tbh I'm not entirely sure about their math. I rephrased it differently, in a way that makes more sense to me, and if you think it makes sense then I'll also open a PR upstream to change the Otel docs. I changed their formula a little bit too.

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
clayton-cornell: @ptodev No problem. Leave it as-is then. Better we opt for consistency in the phrasing now and deal with the bigger issues :-)

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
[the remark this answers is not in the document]
clayton-cornell: Then lets leave it as-is for now - save it for a more general "cleanup" PR.

clayton-cornell (commented):
No further chanegs or questions from teh docs side of things. I'd say docs are good to go for now :-)

component/otelcol/exporter/loadbalancing/loadbalancing.go
```diff
@@ -0,0 +1,246 @@
...
+		Hostname: dnsResolver.Hostname,
+		Port:     dnsResolver.Port,
+		Interval: dnsResolver.Interval,
+		Timeout:  dnsResolver.Timeout,
+	}
+}
+
+// Extensions implements exporter.Arguments.
+func (args Arguments) Extensions() map[otelconfig.ComponentID]otelcomponent.Extension {
+	//TODO: Do we need to set these extensions?
```
mattdurham, line 167: Do we need an answer for this for our PR?

ptodev: Sorry, I should have deleted this comment. We do need to set extensions, and it's already done. I just forgot to delete the comment :) It's deleted now.

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
mattdurham: This still feels a bit odd, the functional default is `pick_first` . @clayton-cornell for his thoughts on it. 
`When unspecified, `pick_first` is used.` -> This feels like the definition of a default when applied to a single field.

component/otelcol/config_tls.go
```diff
@@ -70,17 +70,31 @@ func (args *TLSSetting) Convert() *otelconfigtls.TLSSetting {
...
 		CertFile:       args.CertFile,
-		KeyPem:         configopaque.String(string(args.Key)),
 		KeyFile:        args.KeyFile,
 		MinVersion:     args.MinVersion,
 		MaxVersion:     args.MaxVersion,
 		ReloadInterval: args.ReloadInterval,
 	}
+
+	// Collector sets the pem attributes to nil by default,
+	// so let's match this.
```
mattdurham, line 83: nit: nil or ""?

mattdurham (approved):
Looks reasonable with a few comments that arent major

component/otelcol/config_tls.go
ptodev: Sorry, I'm going to revert this change. This used to be a byte array which would default to nil, but then Otel changed it upstream to a string which defaults to "". The unit tests still pass ok when this change is reverted, so I think it's not needed. Thanks for spotting it!

PR #4152, opened 2023-06-14
Upgrade cadvisor dependency to v0.47.0

jcreixell:
#### PR Description

  - I have not found any backwards incompatible changes
  - See https://github.com/google/cadvisor/releases

#### PR Checklist

- [ ] CHANGELOG updated


marctc (approved):
I wonder why we don't get dependabot issues for this dependency when new releases are pushed

PR #4153, opened 2023-06-14
helm: allow custom resources for config reloader container

tpaschalis:
#### PR Description
Making use of the HPA requires both containers to have resource requests set up to calculate a replica count from memory resource utilization.

This PR allows setting the config reloader 

No issue filed.

#### PR Checklist

- [X] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/operations/helm/charts/grafana-agent/values.yaml
+++ b/operations/helm/charts/grafana-agent/values.yaml
@@ -103,6 +103,8 @@ configReloader:
     tag: v0.8.0
   # -- Override the args passed to the container.
   customArgs: []
+  # -- Resource requests and limits to apply to the config reloader container.
+  resources: {}
 
 controller:
   # -- Type of controller to use for deploying Grafana Agent in the cluster.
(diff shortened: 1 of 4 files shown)


operations/helm/charts/grafana-agent/values.yaml
```diff
@@ -103,6 +103,8 @@ configReloader:
     tag: v0.8.0
   # -- Override the args passed to the container.
   customArgs: []
+  # -- Resource requests and limits to apply to the config reloader container.
+  resources: {}
```
rfratto, line 107: Should we provide some defaults here? We should roughly know what the requests should be, and don't expect the load to ever change.

rfratto (approved):

operations/helm/charts/grafana-agent/values.yaml
tpaschalis: Good point; I've added some extra-small defaults to make HPA provisioning easier.

PR #4160, opened 2023-06-14
Part I: converting promtail configs to flow mode

thampiotr:
#### PR Description
This sets up foundations for the promtail configuration conversion to flow mode configuration.

There is some work remaining, but to limit the size of the PR, I am implementing only some features. Specifically, what is remaining is: 
* More SD components
* More log source components
* More pipeline stages support for loki.process

#### Notes to the Reviewer
There are two main kinds of pipelines that this supports:
- Logs source based - see `source_pipeline_example.river` test - for example: `loki.source.journal -> loki.process -> loki.write` with relabel rules directly passed to `loki.source.journal`.
  - Some components will not support taking relabel rules directly, such as `loki.source.cloudflare`, so there is an extra `loki.relabel` added. See `cloudflare_relabel.river` test.
- SD-based - see `sd_pipeline_example.river` test - for example: `discovery.kubernetes -> discovery.relabel -> discovery.file -> loki.source.file -> loki.process -> loki.write`.

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/converter/internal/common/common.go
+++ b/converter/internal/common/common.go
@@ -18,8 +18,14 @@ import (
 // NewBlockWithOverride generates a new [*builder.Block] using a hook to
 // override specific types.
 func NewBlockWithOverride(name []string, label string, args component.Arguments) *builder.Block {
+	return NewBlockWithOverrideFn(name, label, args, getValueOverrideHook())
+}
+
+// NewBlockWithOverrideFn generates a new [*builder.Block] using a hook fn to
+// override specific types.
+func NewBlockWithOverrideFn(name []string, label string, args component.Arguments, fn builder.ValueOverrideHook) *builder.Block {
 	block := builder.NewBlock(name, label)
-	block.Body().SetValueOverrideHook(getValueOverrideHook())
+	block.Body().SetValueOverrideHook(fn)
 	block.Body().AppendFrom(args)
 	return block
 }
--- a/converter/internal/promtailconvert/promtailconvert.go
+++ b/converter/internal/promtailconvert/promtailconvert.go
@@ -1,0 +1,215 @@
+package promtailconvert
+
+import (
+	"bytes"
+	"flag"
+	"fmt"
+
+	"github.com/alecthomas/units"
+	"github.com/grafana/agent/component/common/loki"
+	lokiwrite "github.com/grafana/agent/component/loki/write"
+	"github.com/grafana/agent/converter/diag"
+	"github.com/grafana/agent/converter/internal/common"
+	"github.com/grafana/agent/converter/internal/prometheusconvert"
+	"github.com/grafana/agent/converter/internal/promtailconvert/internal/build"
+	"github.com/grafana/agent/pkg/river/token/builder"
+	"github.com/grafana/dskit/flagext"
+	"github.com/grafana/loki/clients/pkg/promtail/client"
+	promtailcfg "github.com/grafana/loki/clients/pkg/promtail/config"
+	"github.com/grafana/loki/clients/pkg/promtail/limit"
+	"github.com/grafana/loki/clients/pkg/promtail/positions"
+	"github.com/grafana/loki/clients/pkg/promtail/scrapeconfig"
+	lokicfgutil "github.com/grafana/loki/pkg/util/cfg"
+	lokiflag "github.com/grafana/loki/pkg/util/flagext"
+	"gopkg.in/yaml.v2"
+)
+
+type Config struct {
+	promtailcfg.Config `yaml:",inline"`
+}
+
+// Clone takes advantage of pass-by-value semantics to return a distinct *Config.
+// This is primarily used to parse a different flag set without mutating the original *Config.
+func (c *Config) Clone() flagext.Registerer {
+	return func(c Config) *Config {
+		return &c
+	}(*c)
+}
+
+// Convert implements a Promtail config converter.
+func Convert(in []byte) ([]byte, diag.Diagnostics) {
+	var (
+		diags diag.Diagnostics
+		cfg   Config
+	)
+
+	// Set default values first.
+	flagSet := flag.NewFlagSet("", flag.PanicOnError)
+	err := lokicfgutil.Unmarshal(&cfg,
+		lokicfgutil.Defaults(flagSet),
+	)
+	if err != nil {
+		diags.Add(diag.SeverityLevelCritical, fmt.Sprintf("failed to set default Promtail config values: %s", err))
+		return nil, diags
+	}
+
+	// Unmarshall explicitly specified values
+	if err := yaml.UnmarshalStrict(in, &cfg); err != nil {
+		diags.Add(diag.SeverityLevelCritical, fmt.Sprintf("failed to parse Promtail config: %s", err))
+		return nil, diags
+	}
+
+	// Replicate promtails' handling of this deprecated field.
+	if cfg.ClientConfig.URL.URL != nil {
+		// if a single client config is used we add it to the multiple client config for backward compatibility
+		cfg.ClientConfigs = append(cfg.ClientConfigs, cfg.ClientConfig)
+	}
+
+	f := builder.NewFile()
+	diags = AppendAll(f, &cfg.Config, diags)
+
+	var buf bytes.Buffer
+	if _, err := f.WriteTo(&buf); err != nil {
+		diags.Add(diag.SeverityLevelCritical, fmt.Sprintf("failed to render Flow config: %s", err.Error()))
+		return nil, diags
+	}
+
+	if len(buf.Bytes()) == 0 {
+		return nil, diags
+	}
+
+	prettyByte, newDiags := common.PrettyPrint(buf.Bytes())
+	diags = append(diags, newDiags...)
+	return prettyByte, diags
+}
+
+// AppendAll analyzes the entire promtail config in memory and transforms it
+// into Flow components. It then appends each argument to the file builder.
+func AppendAll(f *builder.File, cfg *promtailcfg.Config, diags diag.Diagnostics) diag.Diagnostics {
+	validateTopLevelConfig(cfg, &diags)
+
+	var writeReceivers = make([]loki.LogsReceiver, len(cfg.ClientConfigs))
+	var writeBlocks = make([]*builder.Block, len(cfg.ClientConfigs))
+	// Each client config needs to be a separate remote_write,
+	// because they may have different ExternalLabels fields.
+	for i, cc := range cfg.ClientConfigs {
+		writeBlocks[i], writeReceivers[i] = newLokiWrite(&cc, &diags, i)
+	}
+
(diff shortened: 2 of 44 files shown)


component/common/loki/types.go
```diff
@@ -27,6 +29,14 @@ const finalEntryTimeout = 5 * time.Second
...
 type LogsReceiver chan Entry
 
+// TODO(piotr): Since LogsReceiver is not an interface, we can't extend it...
+func (l LogsReceiver) RiverTokenize() []builder.Token {
+	return []builder.Token{{
+		Tok: token.STRING,
+		Lit: "loki.write.FIX_ME.receiver",
+	}}
+}
+
```
thampiotr, line 39: @erikbaranowski I ran into this issue when trying to follow the pattern you've used for prometheus config conversion... I cannot extend `LogsReceiver` with `RiverTokenize` because it's not an interface.

I am happy to implement a refactor in a separate PR for changing `type LogReceiver chan Entry` into an interface, something like:
```
type LogReceiver interface {
    Chan() chan Entry
}
```

what do you think?

erikbaranowski: we spoke offline about trying it as an interface since other tricks don't line up exactly in this case.

converter/internal/promtailconvert/testdata/kubernetes.river
```diff
@@ -0,0 +1,70 @@
...
+			host = "example",
+			path = "/oauth2/token",
+		}
+
+		tls_config { }
+	}
+}
+
+discovery.file "fun" {
+	path_targets = concat(discovery.kubernetes.fun_0.targets, discovery.kubernetes.fun_1.targets, discovery.kubernetes.fun_2.targets, discovery.kubernetes.fun_3.targets, discovery.kubernetes.fun_4.targets)
```
rfratto, line 64: tiny nit: it would be nice to format these on separate lines for readability:

```suggestion
	path_targets = concat(
		discovery.kubernetes.fun_0.targets, 
		discovery.kubernetes.fun_1.targets, 
		discovery.kubernetes.fun_2.targets, 
		discovery.kubernetes.fun_3.targets, 
		discovery.kubernetes.fun_4.targets,
	)
```

converter/internal/promtailconvert/testdata/source_pipeline_example.river
```diff
@@ -0,0 +1,41 @@
+loki.write "default_0" {
+	endpoint {
+		url              = "http://localhost/loki/api/v1/push"
+		follow_redirects = false
+		enable_http2     = false
+	}
+	external_labels = {}
+}
```
rfratto, line 8: tiny nit: it would be nice to put loki.write at the bottom. I typically see the order of the files as:

1. `discovery.*` component(s)
2. `discovery.relabel` component(s)
3. `loki.source.*` component(s) 
4. `loki.write` component(s)

rfratto (commented):
I looked at this from the output of the conversion; mostly looks good! I only have one real question with a few small nits.

converter/internal/promtailconvert/promtailconvert.go
```diff
@@ -0,0 +1,270 @@
...
+
+	// Append all the components that produce logs directly.
+	// If any relabelling is required, it will be done via a loki.relabel component.
+	// The logs are sent to loki.process if processing is needed, or directly to loki.write components.
+	//TODO(thampiotr): add support for other integrations
+	b.AppendCloudFlareConfig()
+	b.AppendJournalConfig()
+}
+
+func appendLokiWrite(f *builder.File, client *client.Config, diags *diag.Diagnostics, index int) loki.LogsReceiver {
```
erikbaranowski, line 210: consider now or later breaking these functions up into their own files for each flow component. it's not too bad now but I think it may get less manageable as you continue.

erikbaranowski (commented):
Great start to this new converter!

Heads up, I did modify how the prometheus validation is organized so your changes to prometheus converter code may cause a little merge weirdness. Hopefully it's not too bad since you were mostly just making some functionality public.

converter/internal/common/common.go
```diff
@@ -24,6 +24,16 @@ func NewBlockWithOverride(name []string, label string, args component.Arguments)
 	return block
 }
 
+// NewBlockWithOverrideFn generates a new [*builder.Block] using a hook fn to
+// override specific types.
+func NewBlockWithOverrideFn(name []string, label string, args component.Arguments, fn builder.ValueOverrideHook) *builder.Block {
+	block := builder.NewBlock(name, label)
+	block.Body().SetValueOverrideHook(fn)
+	block.Body().AppendFrom(args)
+	block.Body().SetValueOverrideHook(nil)
```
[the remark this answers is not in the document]
thampiotr: Yeah, I'll actually remove this line, so the function does more exactly what its name suggests.

converter/internal/promtailconvert/promtailconvert.go
```diff
@@ -0,0 +1,270 @@
...
+	"gopkg.in/yaml.v2"
+)
+
+type Config struct {
+	promtailcfg.Config `yaml:",inline"`
+}
+
+// Clone takes advantage of pass-by-value semantics to return a distinct *Config.
+// This is primarily used to parse a different flag set without mutating the original *Config.
+func (c *Config) Clone() flagext.Registerer {
```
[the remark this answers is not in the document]
thampiotr: Yeah, we need this to call 
```go
	err := lokicfgutil.Unmarshal(&cfg,
		lokicfgutil.Defaults(flagSet),
	)
```
That's the cleanest way to get defaults from promtail I found.

converter/internal/promtailconvert/promtailconvert_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	promSuffix   = ".yaml"
+	flowSuffix   = ".river"
+	errorsSuffix = ".errors"
+)
+
+func TestConvert(t *testing.T) {
```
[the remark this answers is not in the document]
thampiotr: That's a good idea, we should keep these in sync, thanks!

converter/internal/promtailconvert/testdata/remote_write.yaml
```diff
@@ -0,0 +1,44 @@
...
+      max_period: 10s
+      max_retries: 20
+    external_labels:
+      cluster: backyard-pine-treehouse-1
+      host: strawberry-pi
+    timeout: 30s
+tracing:
+  enabled: false
+server:
+  disable: true
```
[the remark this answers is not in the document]
thampiotr: I've spent more time than I probably should on figuring out that this is actually an issue with the current promtail version that we have, which has been [fixed in this commit](https://github.com/grafana/loki/commit/de937e72cdb62062424e482fce1c87efc38bd9d9).

Once we update loki, this will start failing. It so happens that I am also the author of the PR updating loki, so I'll take care of broken tests at the point of whichever PR goes second.

converter/internal/promtailconvert/promtailconvert.go
```diff
@@ -0,0 +1,270 @@
...
+	diags = append(diags, newDiags...)
+	return prettyByte, diags
+}
+
+// AppendAll analyzes the entire promtail config in memory and transforms it
+// into Flow components. It then appends each argument to the file builder.
+func AppendAll(f *builder.File, cfg *promtailcfg.Config, diags diag.Diagnostics) diag.Diagnostics {
+	// We currently do not support the new global file watch config. It's an error, since setting it indicates
+	// some advanced tuning which the user likely needs.
+	if cfg.Global.FileWatch != file.DefaultWatchConig {
```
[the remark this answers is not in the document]
thampiotr: Sounds good, done!

converter/internal/promtailconvert/promtailconvert.go
thampiotr: makes sense, done!

thampiotr:
> Heads up, I did modify how the prometheus validation is organized so your changes to prometheus converter code may cause a little merge weirdness. Hopefully it's not too bad since you were mostly just making some functionality public.

Yes, I'm aware - that's the reason I didn't move the exported methods to the top of the file as one would typically do - I wanted to reduce merge issues. Feel free to refactor as you see fit though :)

converter/internal/promtailconvert/testdata/cloudflare_relabel.river
```diff
@@ -0,0 +1,31 @@
...
+	zone_id   = "area51"
+	labels    = {
+		identity    = "unidentified",
+		object_type = "flying",
+	}
+	workers     = 42
+	pull_range  = "1h0m0s"
+	fields_type = "all"
+	forward_to  = [loki.relabel.fun.receiver]
+}
```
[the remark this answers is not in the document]
thampiotr: Took a look and I found the following:
* Cloudflare component only adds static labels that it's configured with: https://github.com/grafana/agent/blob/99aba52da1b64250c992c7eb176a4d6d24919fb9/component/loki/source/cloudflare/internal/cloudflaretarget/target.go#L180C1-L186
* The docs also highlight that all the data comes as JSON:
> All incoming Cloudflare log entries are in JSON format. You can make use of the loki.process component and a JSON processing stage to extract more labels or change the log line format. A sample log looks like this:
[source](https://grafana.com/docs/agent/latest/flow/reference/components/loki.source.cloudflare/)
* I also checked in a unit test that if I add a `__prefixed` label to static labels for cloudflare component, they do not get dropped.

So I think that the way this works now is fine - one can only set static labels on cloudflare, but one can also relabel those static labels using the rules... which in theory can be shared for many integrations etc.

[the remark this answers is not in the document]
rfratto: Thanks for checking! Agreed that the way this works now is fine.

converter/internal/promtailconvert/validate.go
```diff
@@ -0,0 +1,72 @@
...
+	// some advanced tuning which the user likely needs.
+	if cfg.Global.FileWatch != file.DefaultWatchConig {
+		diags.Add(diag.SeverityLevelError, "global/file_watch_config is not supported")
+	}
+
+	// The positions global config is not supported in Flow Mode. It's a warning since the Flow Mode likely
+	// provides a good enough alternative through components' positions files in its own data directory.
+	if cfg.PositionsConfig != defaultPositionsConfig() {
+		diags.Add(
+			diag.SeverityLevelWarn,
```
erikbaranowski, line 22: With `SeverityLevelWarn`, we are able to run the agent without bypassing errors in these cases.  I want to be very careful that we are sure we want to be able to do that if the behavior doesn't exactly match promtails for each of these. (same question on other warnings in this validation)

thampiotr: That's a good point. I have made this an error and also reviewed the severity level of other diags.

PR #4162, opened 2023-06-15
helm: make resources required for autoscaling

tpaschalis:
#### PR Description
This PR amends the Helm chart to fail when trying to use autoscaling without having the resources key set for both the agent and configReloader containers.

I'm on the fence whether the added complexity is worth it from a Helm-conventions perspective, let me know what you think. If these were not set, the HPA would be created, but will have no effect.

No issue filed.

#### Notes to the Reviewer
* This will require amending if/when we allow the HPA to act on custom metrics instead of CPU/Memory utilization
* This started out with the generic verification that the `resources.request` field just exists; the last commit adds specific validations for when the user has chosen cpu/memory-utilization based autoscaling. Let me know if you think it's too much and I can revert back to the previous solution. 
* We have to check for both the resources and resources.X field, otherwise we're getting errors like `<.Values.agent.resources.requests.memory>: nil pointer evaluating interface {}.memory`
 
#### PR Checklist

- [X] CHANGELOG updated
- [X] Documentation added
- [X] Tests updated

### DIFF
--- a/operations/helm/charts/grafana-agent/CHANGELOG.md
+++ b/operations/helm/charts/grafana-agent/CHANGELOG.md
@@ -18,6 +18,11 @@ Unreleased
 
 - Don't specify replica count for StatefulSets when autoscaling is enabled (@captncraig)
 
+### Other changes
+
+- Make the agent and config-reloader container resources required when using
+  autoscaling. (@tpaschalis)
+
 0.15.0 (2023-06-08)
 -------------------
 
--- a/operations/helm/charts/grafana-agent/templates/hpa.yaml
+++ b/operations/helm/charts/grafana-agent/templates/hpa.yaml
@@ -1,4 +1,6 @@
 {{- if and (or (eq .Values.controller.type "deployment") (eq .Values.controller.type "statefulset" ))  .Values.controller.autoscaling.enabled }}
+{{- $name := .Values.agent.resources.requests | required ".Values.agent.resources.requests is required when using autoscaling." -}}
+{{- $name := .Values.configReloader.resources.requests | required ".Values.configReloader.resources.requests is required when using autoscaling." -}}
 apiVersion: autoscaling/v2
 kind: HorizontalPodAutoscaler
 metadata:
--- a/operations/helm/charts/grafana-agent/templates/hpa.yaml
+++ b/operations/helm/charts/grafana-agent/templates/hpa.yaml
@@ -1,6 +1,16 @@
 {{- if and (or (eq .Values.controller.type "deployment") (eq .Values.controller.type "statefulset" ))  .Values.controller.autoscaling.enabled }}
-{{- $name := .Values.agent.resources.requests | required ".Values.agent.resources.requests is required when using autoscaling." -}}
-{{- $name := .Values.configReloader.resources.requests | required ".Values.configReloader.resources.requests is required when using autoscaling." -}}
+{{- if not (eq .targetMemoryUtilizationPercentage "0") }} 
+  {{- $name := .Values.agent.resources.requests | required ".Values.agent.resources.requests is required when using autoscaling." -}}
+  {{- $name := .Values.agent.resources.requests.memory | required ".Values.agent.resources.requests.memory is required when using autoscaling based on memory utilization." -}}
+  {{- $name := .Values.configReloader.resources.requests  | required ".Values.configReloader.resources.requests is required when using autoscaling." -}}
+  {{- $name := .Values.configReloader.resources.requests.memory  | required ".Values.configReloader.resources.requests.memory is required when using autoscaling based on memory utilization." -}}
+{{- end}}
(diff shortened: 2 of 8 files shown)


operations/helm/charts/grafana-agent/templates/hpa.yaml
```diff
@@ -1,4 +1,16 @@
 {{- if and (or (eq .Values.controller.type "deployment") (eq .Values.controller.type "statefulset" ))  .Values.controller.autoscaling.enabled }}
+{{- if not (eq .targetMemoryUtilizationPercentage "0") }}
```
tpaschalis, line 2: Initially, I just checked for the presence of the resources.requests field, but then added the more fine-grained checks.

If you think this is too much, I can revert back to the simpler check with a better error message.

operations/helm/charts/grafana-agent/templates/hpa.yaml
```diff
@@ -1,4 +1,16 @@
 {{- if and (or (eq .Values.controller.type "deployment") (eq .Values.controller.type "statefulset" ))  .Values.controller.autoscaling.enabled }}
+{{- if not (empty .targetMemoryUtilizationPercentage) }}
```
tpaschalis, line 2: Initially, I just checked for the presence of the resources.requests field, but then added the more fine-grained checks.

If you think this is too much, I can revert back to the simpler check with a better error message.

operations/helm/charts/grafana-agent/CHANGELOG.md
```diff
@@ -18,6 +18,9 @@ Unreleased
 
 - Don't specify replica count for StatefulSets when autoscaling is enabled (@captncraig)
 
+- Make the agent and config-reloader container resources required when using
+  autoscaling. (@tpaschalis)
```
rfratto, line 22: Is this an enhancement or an "other change?"

operations/helm/charts/grafana-agent/templates/hpa.yaml
```diff
@@ -1,4 +1,16 @@
...
+  {{- $name := .Values.configReloader.resources.requests  | required ".Values.configReloader.resources.requests is required when using autoscaling." -}}
+  {{- $name := .Values.configReloader.resources.requests.memory  | required ".Values.configReloader.resources.requests.memory is required when using autoscaling based on memory utilization." -}}
+{{- end}}
+{{- if not (empty .targetCPUUtilizationPercentage)}} 
+  {{- $name := .Values.agent.resources.requests | required ".Values.agent.resources.requests is required when using autoscaling." -}}
+  {{- $name := .Values.agent.resources.requests.cpu | required ".Values.agent.resources.requests.cpu is required when using autoscaling based on cpu utilization." -}}
+  {{- $name := .Values.configReloader.resources.requests  | required ".Values.configReloader.resources.requests is required when using autoscaling." -}}
+  {{- $name := .Values.configReloader.resources.requests.cpu  | required ".Values.configReloader.resources.requests.cpu is required when using autoscaling based on cpu utilization." -}}
+{{- end}}
```
rfratto, line 13: This is a great idea, I just wish it didn't have to be so long 😢

rfratto (approved):
:shipit:

operations/helm/charts/grafana-agent/templates/hpa.yaml
rfratto:
```go
  {{- .Values.agent.resources.requests | required ".Values.agent.resources.requests is required when using autoscaling." -}}
  {{- .Values.agent.resources.requests.cpu | required ".Values.agent.resources.requests.cpu is required when using autoscaling based on cpu utilization." -}}
  {{- .Values.configReloader.resources.requests | required ".Values.configReloader.resources.requests is required when using autoscaling." -}}
  {{- .Values.configReloader.resources.requests.cpu | required ".Values.configReloader.resources.requests.cpu is required when using autoscaling based on cpu utilization." -}}
```

would still be pretty long but a little bit shorter :)

tpaschalis: Good catch, I probably copied from an Helm docs example; we don't make use of it here. Removing it.

tpaschalis: Hmm, when I remove the variable, it reports the following error. Maybe the example had the local variable to avoid actually rendering the value it if it was present, I'll re-read the docs.

```
$ helm upgrade --install --namespace agent grafana-agent .
Error: UPGRADE FAILED: YAML parse error on grafana-agent/templates/hpa.yaml: error converting YAML to JSON: yaml: line 2: mapping values are not allowed in this context
```

rfratto: Never mind, I see that required returns the value, so you do actually need to bind it to a temporary variable.

Issue #4141, opened 2023-06-13
Support custom fields lookup in geoip stage in loki.process component


akselleirv:
### Request

The [geoip stage](https://grafana.com/docs/agent/next/flow/reference/components/loki.process/#stagegeoip-block) in `loki.process` only support extracting a set of predefined fields from the MMDB database.

A new attribute in the geoip is therefore required. A suggestion:
```
loki.process "example" {
	stage.json {
		expressions = {ip = "clientIP"}
	}
	stage.geoip {
		db      = "/path/to/GeoLite2-City-Enriched.mmdb"
		db_type = "city"
		source  = "ip"
		lookup  = {
			"department"          = "MyCompany.DeptName",
			"parent_vnet"         = "MyCompany.ParentVnet",
			"resource_group_name" = "MyCompany.ResourceGroupName",
			"subnet"              = "MyCompany.SubnetName",
                        "envs"                = "MyCompany.Environments",
 		}
	}

	forward_to = []
}
```

Where lookup is a map of `map[label]*JMESPath{}`. Please feel free to suggest a better name than `lookup`. Naming is hard.

### Use case

Add private IP addresses to the MMDB database, as explained in the article [Enriching MMDB files](https://github.com/maxmind/mmdb-from-go-blogpost), and then enhance the logs with this information by using the merge stage as proposed proposed in #4111.

PR #4166, opened 2023-06-15
loki.process: extract data to shared map in geoip stage

akselleirv:
#### PR Description

Extracting data as labels, how it was done in the port from promtail, makes it more difficult to use with other stages. In addition, extracted values had to be converted to a string. The only extracted non-string values are currently only `latitude` and `longitude`, however, with the #4141 proposal then it would be possible to add complex types. 

#### PR Checklist

- [ ] CHANGELOG updated
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/component/loki/process/internal/stages/geoip.go
+++ b/component/loki/process/internal/stages/geoip.go
@@ -115,7 +115,7 @@ func (g *geoIPStage) Name() string {
 	return StageTypeGeoIP
 }
 
-func (g *geoIPStage) process(labels model.LabelSet, extracted map[string]interface{}) {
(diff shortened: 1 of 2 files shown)


tpaschalis:
@mattdurham do you mean race conditions between multiple loki.process stages running in parallel (like we fixed in https://github.com/grafana/agent/commit/67692153c9037d5892dcec9347973b6ce6354be9) or something different altogether?

tpaschalis (approved):
LGTM, thanks Aksel!

mattdurham:
@tpaschalis yes I did, so likely not an issue if we are handling it in the root. Apologies.

PR #4168, opened 2023-06-15
Docs: Fix broken link to Mimir Get started.

osg-grafana:


osg-grafana:
@brendamuir, does this need to be backported?

clayton-cornell:
> does this need to be backported?

I checked the prev versions. They all 404, so backporting is needed :-)  Added the labels.

PR #4170, opened 2023-06-15
drone: only run Windows tests on merge

rfratto:
The Windows tests historically have bad performance, causing the turnaround time for a PR to be anywhere up to 10-15 minutes, even though all other jobs finish in 3 minutes or less.

This commit takes Windows tests out of the PR testing path to reduce total turnaround time for a PR.

Windows tests will now be run on main commits, and maintainers will need to be on the lookout for test failures so they can follow up if a bad PR gets merged that breaks Windows.


mattdurham (approved):
LGTM

PR #4176, opened 2023-06-15
converter: pretty-print the output

rfratto:
Normally builder.File already pretty-prints the file, but custom tokens aren't pretty-printed. This makes sure that the resulting file is in the standard format.


erikbaranowski (approved):
LGTM, it can always be shuffled later but that looks like a good spot to me

PR #4187, opened 2023-06-16
Fix relabel bug

thampiotr:
#### PR Description

In [this change](https://github.com/grafana/agent/commit/edcc83858ad92f3b96c44fa77142db352dd431de#diff-5f10a07e4c853517c2aad2506e6686e92430e34aaa059f7338c7fcd89df2e0bfL198), the use of `:=` creates a new `relabelled` variable in scope only of the `else` block. This means that even though we call `addToCache`, we will return `nil` from `relabel` method.

It likely means that relabel is not done correctly when we have a first cache miss.

It's not clear if this can lead to issues with caching yet (e.g. memory leak), further testing is required.

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,7 +49,7 @@ Main (unreleased)
 
 - Fix a bug where `prometheus.relabel` would not relabel correctly when there is a cache miss. (@thampiotr)
 
-- Fix a bug where `prometheus.relabel` would not relabel correctly the exemplars or metadata. (@tpaschalis)
+- Fix a bug where `prometheus.relabel` would not correctly relabel exemplars or metadata. (@tpaschalis)
 
 ### Other changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,7 +47,9 @@ Main (unreleased)
 
 - Fix oauth default scope in `loki.source.azure_event_hubs`. (@akselleirv)
 
-- Fix a bug where `prometheus.relabel` would not relabel correctly when there is a cache miss. (@thampiotr) 
+- Fix a bug where `prometheus.relabel` would not relabel correctly when there is a cache miss. (@thampiotr)
+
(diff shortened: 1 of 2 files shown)


tpaschalis (approved):
:shipit:

CHANGELOG.md
```diff
@@ -47,6 +47,10 @@ Main (unreleased)
 
 - Fix oauth default scope in `loki.source.azure_event_hubs`. (@akselleirv)
 
+- Fix a bug where `prometheus.relabel` would not relabel correctly when there is a cache miss. (@thampiotr)
+
+- Fix a bug where `prometheus.relabel` would not relabel correctly the exemplars or metadata. (@tpaschalis)
```
rfratto, line 52:
```suggestion
- Fix a bug where `prometheus.relabel` would not correctly relabel exemplars or metadata. (@tpaschalis)
```

component/prometheus/relabel/relabel.go
```diff
@@ -96,6 +97,10 @@ func New(o component.Options, args Arguments) (*Component, error) {
 		Name: "agent_prometheus_relabel_cache_size",
 		Help: "Total size of relabel cache",
 	})
+	c.cacheDeletes = prometheus_client.NewCounter(prometheus_client.CounterOpts{
+		Name: "agent_prometheus_relabel_cache_deletes",
+		Help: "Total number of cache deletes",
+	})
 
 	var err error
 	for _, metric := range []prometheus_client.Collector{c.metricsProcessed, c.metricsOutgoing, c.cacheMisses, c.cacheHits, c.cacheSize} {
```
rfratto, line 106: You need to add the new metric here so it gets registered with the collector

rfratto (approved):
LGTM! @tpaschalis can you give this a second pair of eyes?

tpaschalis (approved):

PR #4189, opened 2023-06-16
Fix several issues with statsd exporter

jcreixell:
#### PR Description

  - Static and Flow: statsd_exporter failed when specifying mappings. This was caused by an unmarshal/marshal/unmarshal sequence that set some wrong defaults.
  - Flow: Fixes defaults for `listen_udp` and `listen_tcp` arguments and updates documentation accordingly.
  - Flow: Fixes an issue where an undefined `mapping_config_path` argument would cause a load error.

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated


pkg/integrations/statsd_exporter/statsd_exporter.go
```diff
@@ -53,11 +53,11 @@ var DefaultConfig = Config{
...
-	ListenUDP      string               `yaml:"listen_udp,omitempty"`
-	ListenTCP      string               `yaml:"listen_tcp,omitempty"`
-	ListenUnixgram string               `yaml:"listen_unixgram,omitempty"`
-	UnixSocketMode string               `yaml:"unix_socket_mode,omitempty"`
-	MappingConfig  *mapper.MetricMapper `yaml:"mapping_config,omitempty"`
+	ListenUDP      string `yaml:"listen_udp,omitempty"`
+	ListenTCP      string `yaml:"listen_tcp,omitempty"`
+	ListenUnixgram string `yaml:"listen_unixgram,omitempty"`
+	UnixSocketMode string `yaml:"unix_socket_mode,omitempty"`
+	MappingConfig  any    `yaml:"mapping_config,omitempty"`
```
spartan0x117, line 60: Is my understanding correct that the original Unmarshal in `readMappingFile` will make a `map[map[...]...]` (series of maps that eventually at the leaves will be `map[string]string`), and then Marshaling this again _should_ result in essentially the exact yaml that was present in the file (possible ordering of keys on the same level of indentation being the main difference), which should then lead to the `statsdMapper.InitFromYAMLString(string(cfgBytes))` [in the integration code](https://github.com/grafana/agent/blob/main/pkg/integrations/statsd_exporter/statsd_exporter.go#L140) being called with as close to the exact contents of the file as possible?

spartan0x117 (approved):
A question for my own understanding, but LGTM!

clayton-cornell (commented):
Looks good on the docs side

clayton-cornell (approved):

PR #4190, opened 2023-06-16
prometheus.relabel: do a more resilient empty label check

rfratto:
Checking for a labelset to be nil is not always correct; an dropped metric from relabel will return a non-nil, but still empty labelset.


tpaschalis (approved):

mattdurham (approved):
Piling on

PR #4191, opened 2023-06-16
Switch to using capped LRU cache instead of raw map.

mattdurham:
#### PR Description

Instead of using an unbounded `map` for relabel caches instead use an LRU of 100_000 items. Currently its not configurable but likely to become so once we sort out good defaults and performance.

#### PR Checklist

- [X] CHANGELOG updated
- [ ] Documentation added
- [X] Tests updated

### DIFF
--- a/component/prometheus/relabel/relabel.go
+++ b/component/prometheus/relabel/relabel.go
@@ -12,6 +12,7 @@ import (
 	"github.com/grafana/agent/component"
 	flow_relabel "github.com/grafana/agent/component/common/relabel"
 	"github.com/grafana/agent/component/prometheus"
+	lru "github.com/hashicorp/golang-lru"
 	prometheus_client "github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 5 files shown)


component/prometheus/relabel/relabel.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/grafana/agent/component"
 	flow_relabel "github.com/grafana/agent/component/common/relabel"
 	"github.com/grafana/agent/component/prometheus"
+	lru "github.com/hashicorp/golang-lru"
```
rfratto, line 15: Did you meant to use `github.com/hashicorp/golang-lru/v2`, which supports generics?

mattdurham: I didn't explicitly do that one since Loki isn't, but def can switch to it.

rfratto: Seems like it might be nicer for us, but I'll leave the choice up to you

rfratto (approved):

PR #4196, opened 2023-06-16
otelcol.exporter.prometheus: Make histogram counts cumulative for Prometheus

### LINKED ISSUE #4193: otelcol.exporter.prometheus: Histograms are not exported with cumulative counts
### What's wrong?

The following community discussion in OpenTelemetry summarises the problem well https://github.com/open-telemetry/community/discussions/897

Essentially (copied from the discussion above):

In [prometheus](https://prometheus.io/docs/concepts/metric_types/#histogram):
```
A histogram with a base metric name of <basename> exposes multiple time series during a scrape:

cumulative counters for the observation buckets, exposed as <basename>_bucket{le="<upper inclusive bound"}
the total sum of all observed values, exposed as <basename>_sum
the count of events that have been observed, exposed as <basename>_count (identical to <basename>_bucket{le="+Inf"} above)
```

But in OpenTelemetry, buckets only contains count [inside the bucket interval](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/metrics/v1/metrics.proto#L469) (see also [the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/datamodel.md#histogram)).

This means that, when a user creates a histogram bucket through OpenTelemetry and then uses the Agent otelcol.exporter.prometheus, the generated histogram bucket will be invalid and rendering it in Grafana causes really bizarre graphs, making high percentiles show as `infinity`.

### Steps to reproduce

Here's a sample river configuration to expose this problem:

```
otelcol.receiver.otlp "default" {
  grpc {
    endpoint = "localhost:4017"
  }
  http {
    endpoint = "localhost:4018"
  }

  output {
    metrics = [otelcol.processor.batch.default.input]
  }
}
  
otelcol.processor.batch "default" {
  output {
    metrics = [otelcol.exporter.prometheus.default.input]
  }
}
  
otelcol.exporter.prometheus "default" {
  forward_to = [prometheus.remote_write.mimir.receiver]
}
  
prometheus.remote_write "mimir" {
  endpoint {
    url = "http://localhost:9009/api/v1/push"
  }
}
```

Any OTel example in Go will generate the wrong kind of Histogram, for example something like this:

```go
meter.Float64Histogram(HTTPServerDuration, instrument.WithUnit("s"))
```

### System information

Any

### Software version

Latest Grafana Agent compiled from main

### Configuration

_No response_

### Logs

_No response_

grcevski:
#### PR Description

This PR attempts to fix the problem with non-cumulative histograms exported to Prometheus, from the otel metrics collector. The code change follows the approach taken by the opentelemetry-prometheus-exporter as shown here:
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/prometheusexporter/collector.go#LL224C76-L224C76

Given a number of histogram buckets, we first sort the buckets by the bucket bounds value, then we calculate the cumulative value for each bucket, instead of using the exact bucket count provided by OTel. 

Fixes #4193

#### Notes to the Reviewer

I'd appreciate any feedback on where to add tests for this issue.

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/component/otelcol/exporter/prometheus/internal/convert/convert.go
+++ b/component/otelcol/exporter/prometheus/internal/convert/convert.go
@@ -472,7 +472,9 @@ func (conv *Converter) consumeHistogram(app storage.Appender, memResource *memor
 
 		sort.Float64s(bounds)
 
-		// Calculate cumulative count value as Prometheus wants it
+		// Calculate cumulative count values. Prometheus expects cummulative bucket counts for histograms.
+		// This has nothing to do with temporality, it doesn't affect cummulative vs delta histograms, it
+		// simply matches the format of bucket counts expected by Prometheus.
 		var c uint64 = 0
 		for i := 0; i < len(bounds); i++ {
 			bound := bounds[i]
--- a/component/otelcol/exporter/prometheus/internal/convert/convert.go
+++ b/component/otelcol/exporter/prometheus/internal/convert/convert.go
@@ -10,6 +10,8 @@ package convert
 import (
 	"context"
 	"fmt"
+	"math"
+	"sort"
 	"strconv"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


tpaschalis:
> fix the problem with non-cumulative histograms exported to Prometheus, from the otel metrics collector

So to better understand, this _is_ for cumulative temporality histograms (we're dropping delta temporality histograms for now as we haven't implemented the delta-to-cumulative conversion), but you mean we're fixing the cumulative distribution inside of the buckets, right?

component/otelcol/exporter/prometheus/internal/convert/convert.go
```diff
@@ -458,11 +460,34 @@ func (conv *Converter) consumeHistogram(app storage.Appender, memResource *memor
...
+		bounds := make([]float64, bSize)
+		for i := 0; i < dp.ExplicitBounds().Len() && i < dp.BucketCounts().Len(); i++ {
+			bound := dp.ExplicitBounds().At(i)
+			buckets[bound] = dp.BucketCounts().At(i)
+			bounds = append(bounds, bound)
+		}
+
+		sort.Float64s(bounds)
+
+		// Calculate cumulative count value as Prometheus wants it
```
tpaschalis, line 475: Do you think there's a way to reword here so that it's clearer we don't refer to cumulative _temporality_ but so that we make it clear it's about the way that Prometheus uses cumulative buckets for histograms?

tpaschalis (approved):
Surprising from the OTel side, but it is what it is; LGTM at a first glance.

Let's fix the test and I'll leave the final ✅ to @rfratto who wrote the converter

component/otelcol/exporter/prometheus/internal/convert/convert.go
grcevski: Done! Thanks.

grcevski:
Thanks @tpaschalis  and @rfratto! I've updated the changelog and resolved conflicts with main.

PR #4200, opened 2023-06-19
feat: stage jsonmerge in loki.process

### LINKED ISSUE #4111: loki.process stage which merges given labels to the log line
### Request

Discussion started in Slack thread https://grafana.slack.com/archives/C01050C3D8F/p1686237967379259

Add a stage in `loki.process` which takes a set of labels and adds them to the log line. My initial use case was to add it to a JSON object, but it could be more generic add make it possible to use it with logfmt and possibly an unstructured format by appending.

### Use case

The GeoIP stage in `loki.process`  the values to labels, but I want to add it to the log line itself.

An example which add a set of labels to a JSON object: 

```
Log line:  {"msg":"log message","user":"a-user"}
Labels: { "level" = "error", "env" = "dev","user_id" = "f8fas0r"}

stage.merge {
    labels      = [ "env", "user_id" ]
    source_type = "json"
}

transforms to:
Log line: {"msg":"log message","user":"a-user","env":"dev","user_id":"f8fas0r" }
Labels: { "level" = "error" }
```

akselleirv:
#### PR Description

Merges JSON from the extracted map specified in the values list with the JSON data from the `source` value. 

Closes #4111 

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/component/loki/process/internal/stages/jsonmerge.go
+++ b/component/loki/process/internal/stages/jsonmerge.go
@@ -1,0 +1,96 @@
+package stages
+
+import (
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+	json "github.com/json-iterator/go"
+	"github.com/prometheus/common/model"
+)
+
+// ErrEmptyValuesJSONMergeStageConfig error returned if the config is empty.
+var ErrEmptyValuesJSONMergeStageConfig = errors.New("values in jsonmerge config cannot be empty")
+var ErrEmptyOutputJSONMergeStageConfig = errors.New("output in jsonmerge config cannot be empty")
+
+// JSONMergeConfig contains the slice of labels to be dropped.
+type JSONMergeConfig struct {
+	Source string   `river:"source,attr"`
+	Values []string `river:"values,attr"`
+	Output string   `river:"output,attr"`
+}
+
+func newJSONMergeStage(logger log.Logger, config JSONMergeConfig) (Stage, error) {
+	if len(config.Values) < 1 {
+		return nil, ErrEmptyValuesJSONMergeStageConfig
+	}
+
+	if config.Output == "" {
+		return nil, ErrEmptyOutputJSONMergeStageConfig
+	}
+
+	return toStage(&jsonmergeStage{
+		config: config,
+		logger: logger,
+	}), nil
+}
+
+type jsonmergeStage struct {
+	config JSONMergeConfig
+	logger log.Logger
+}
+
+// Process implements Stage.
+func (j *jsonmergeStage) Process(labels model.LabelSet, extracted map[string]interface{}, t *time.Time, entry *string) {
+	obj, err := j.extractSourceObject(extracted, entry)
+	if err != nil {
+		if Debug {
+			level.Debug(j.logger).Log("msg", "unable to extract source value", "error", err.Error())
+		}
+	}
+
+	for _, value := range j.config.Values {
+		extractedValue, ok := extracted[value]
+		if !ok {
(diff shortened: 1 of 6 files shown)


component/loki/process/internal/stages/jsonmerge.go
```diff
@@ -0,0 +1,96 @@
...
+
+	b, err := json.Marshal(obj)
+	if err != nil {
+		if Debug {
+			level.Debug(j.logger).Log("msg", "unable to json marshal merged object", "error", err.Error())
+		}
+		return
+	}
+
+	extracted[j.config.Output] = string(b)
```
akselleirv, line 73: Hello @tpaschalis, I have some confusion with adding data to the extracted map and using it in the `stage.output`. The output stage requires that the value from the extracted map cannot be a complex structure. Therefore I need to marshal the data before adding it to the extracted map.

However, this results in stringed JSON objects in the output line, for example:
```
{"something":"[1,2,3]","extracted_object":"{\"one\":true}"}
```
Any ideas for working around this issue?

rfratto:
Hey @akselleirv! Just so you know, nobody gets notified about draft PRs, so this fell off our radar. 

If you're looking for eyes on a draft PR, you might want to explicitly ping someone in a comment so they get notified.

cc @tpaschalis, since he seems like the appropriate person to help push this forward.

akselleirv:
> Hey @akselleirv! Just so you know, nobody gets notified about draft PRs, so this fell off our radar.
> 
> If you're looking for eyes on a draft PR, you might want to explicitly ping someone in a comment so they get notified.
> 
> cc @tpaschalis, since he seems like the appropriate person to help push this forward.

Hello @rfratto, I have hoping that my [code comment](https://github.com/grafana/agent/pull/4200#discussion_r1233679729) would notify @tpaschalis. I assumed he was busy the last days and therefore haven't had the time to reply. I'll send him a message on slack.

component/loki/process/internal/stages/jsonmerge.go
tpaschalis: So iiuc you're wondering about the stringed `"[1,2,3]"` field that is no longer an array, right?

The reason behind this is that `stage.json` stores [complex objects](https://github.com/grafana/agent/blob/914b965ba30a752682081c12e9a6433e4a4e3bb0/component/loki/process/internal/stages/json.go#L159-L169) back in their marshalled form, so it gets stored as a string from the very start. On the other hand, numbers and booleans get to keep their type.

Using the `json` filter at query time will give you back the non-stringified form, but there's no way to make Loki treat it as a proper array you can index, slice etc. There's been some discussion in allowing such operations at query time but I'm not sure how far it has progressed.

image

Unfortunately there's no other workaround for now; but getting stage.json to handle more types natively might be interesting in the future

akselleirv:
Hi again @tpaschalis, apologies being back and forth on this issue, but I might close this PR as I've found another workaround which works good enough for my use-case. It might look a bit hacky, but it works 😉


I played around with the `stage.replace` and found a good enough solution.

```
stage.replace {
	expression = "(}$)"
	
	replace =  ",\"ip_lookup\":{\"city_name\":\"{{ .geoip_city_name }}\",\"country_name\":\"{{ .geoip_country_name }}\",\"latitude\":\"{{ .geoip_location_latitude }}\",\"longitude\":\"{{ .geoip_location_longitude }}\"}}"
}
```
It basically just appends to the end of the JSON string.

Also the limitations you mentioned in [the comment above](https://github.com/grafana/agent/pull/4200#discussion_r1238186287), then this component feels a bit too limited and thus unnecessary imo. 

Agree that I should close this PR?

PR #4215, opened 2023-06-20
Authentication support for module.git

### LINKED ISSUE #3694: module.git should support authentication
It would be nice if module.git could authenticate so that it can pull from private repositories.

Not sure what auth options the underlying git library allows, but we should probably expose some of them.

djcode:
#### PR Description

Added authentication ( Basic and SSH Key ) support to `module.git`.

Fixes #3694

#### Notes to the Reviewer

- First PR to grafana-agent.
- I don't consider myself much of a Go developer. Please be harsh and thorough.
- I tried to mimic the authentication logic seen in the `discovery.consul` component.
- Tested locally against some private Github repos, but not sure how to write effective test cases for this.

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [ ] Tests updated


component/module/git/internal/vcs/auth/auth.go
```diff
@@ -0,0 +1,67 @@
+package auth
```
rfratto, line 1: I don't think it's necessary to create a dedicated package for auth; you can just put this file in the upper vcs package.

component/module/git/internal/vcs/auth/auth.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	return &http.BasicAuth{
+		Username: b.Username,
+		Password: string(b.Password),
+	}
+}
+
+type SSHKey struct {
+	Username   string            `river:"username,attr,optional"`
+	Keyfile    string            `river:"keyfile,attr,optional"`
```
rfratto, line 53: We typically use snake_case for multi-word attributes:

```suggestion
	Keyfile    string            `river:"key_file,attr,optional"`
```

component/module/git/internal/vcs/auth/auth.go
```diff
@@ -0,0 +1,67 @@
...
+// Convert converts our type to the native prometheus type
+func (s *SSHKey) Convert() (transport.AuthMethod, error) {
+	if s == nil {
+		return nil, nil
+	}
+	publickeys, err := ssh.NewPublicKeysFromFile(s.Username, s.Keyfile, string(s.Passphrase))
+	if err != nil {
+		return nil, fmt.Errorf("Loading SSH keys failed: %s", err.Error())
+	}
+	return publickeys, nil
```
rfratto, line 66: Cool. I'd like to see it possible to specify the key from another component, which would be extra useful to get the SSH keys from Vault. It looks like the ssh package has a method to give it PEM bytes directly: https://pkg.go.dev/github.com/go-git/go-git/v5/plumbing/transport/ssh#NewPublicKeys

Could we make `key` an extra attribute which is mutually exclusive with setting `key_file`, and use NewPublicKeys if the `key` attribute is non-empty, and NewPublicKeysFromFile otherwise?

component/module/git/internal/vcs/auth/auth.go
```diff
@@ -0,0 +1,67 @@
...
+	if b == nil {
+		return nil
+	}
+	return &http.BasicAuth{
+		Username: b.Username,
+		Password: string(b.Password),
+	}
+}
+
+type SSHKey struct {
```
rfratto, line 51: In the context of SSH keys, I don't think these are all optional; at least the key file would be required, right?

component/module/git/internal/vcs/auth/auth.go
```diff
@@ -0,0 +1,67 @@
...
+		key, _ := h.SSHKey.Convert()
+		return key
+	}
+	return nil
+}
+
+type BasicAuth struct {
+	Username string            `river:"username,attr,optional"`
+	Password rivertypes.Secret `river:"password,attr,optional"`
+}
```
rfratto, line 38: I don't think these are both optional; shouldn't they both be required for basic auth?

rfratto (commented):
Thanks for working on this! I have a few comments, but this is headed in a very positive direction.

component/module/git/internal/vcs/auth/auth.go
djcode: The second after I pushed the PR I thought the same thing. Updated it so it's under the vcs package.

component/module/git/internal/vcs/auth/auth.go
djcode: That's a great question. I thought the same thing and I'm happy to make that change, I did just copy this logic from https://github.com/grafana/agent/blob/main/component/common/config/types.go#L130

rfratto: Interesting, we probably had that discussion before and it turned out they truly should be optional. This can stay as-is, then 👍

component/module/git/internal/vcs/auth/auth.go
djcode: Fixed this in the latest commits.

component/module/git/internal/vcs/auth/auth.go
djcode: I really like this idea. I added in a new 'key' attribute that uses `NewPublicKeys` instead and tested it with a `local.file` component. Worked really well in my local tests.

component/module/git/internal/vcs/auth/auth.go
djcode: I marked these optional to match the BasicAuth block. I didn't experiment with what happens marking these as required, but I assumed if the block is optional all it's attributes also must be optional.

rfratto: It's valid for an optional block to have required attributes; it means those attributes are only required if the block is specified in the config file.

djcode: I just looked at how auth is done for remote.vault and saw required attributes in optional blocks. Added a commit to address this.

I think I've fixed everything now. Thank you for your patience with me reviewing this!

clayton-cornell (commented):
Docs look OK

rfratto (approved):
LGTM, thank you for working on this!

PR #4219, opened 2023-06-21
Flow: add Elasticsearch exporter component

### LINKED ISSUE #2308: Flow: Add elasticsearch_exporter
Create `prometheus.integration.elasticsearch_exporter` component

marctc:
#### PR Description

This PR adds a Flow component which embed elasticsearch exporter.

Fixes #2308

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated


docs/sources/flow/reference/components/prometheus.exporter.elasticsearch.md
```diff
@@ -0,0 +1,107 @@
...
+```river
+prometheus.exporter.elasticsearch "example" {
+  address = "localhost:9200"
+}
+
+// Configure a prometheus.scrape component to collect Elasticsearch metrics.
+prometheus.scrape "demo" {
+  targets    = prometheus.exporter.elasticsearch.example.targets
+  forward_to = [ /* ... */ ]
+}
```
spartan0x117, line 104: I think this should be as copy-paste-able as possible, so something like:
```suggestion
// Configure a prometheus.scrape component to collect Elasticsearch metrics.
prometheus.scrape "demo" {
  targets    = prometheus.exporter.elasticsearch.example.targets
  forward_to = [ prometheus.remote_write.example.receiver ]
}

prometheus.remote_write "example" {
  // Send metrics to a locally running Mimir.
  endpoint {
    url = "http://mimir:9090/api/v1/write"

    basic_auth {
      username = "sample-username"
      password = "sample-password"
    }
  }
}
```

component/prometheus/exporter/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,91 @@
...
+var DefaultArguments = Arguments{
+	Address:                   "http://localhost:9200",
+	Timeout:                   5 * time.Second,
+	Node:                      "_local",
+	ExportClusterInfoInterval: 5 * time.Minute,
+	IncludeAliases:            true,
+}
+
+type Arguments struct {
+	Address                   string        `river:"address,attr,optional"`
```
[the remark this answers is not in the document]
marctc: It has a default value to http://localhost:9200, maybe we could change it, but the original integration behaves like this

docs/sources/flow/reference/components/prometheus.exporter.elasticsearch.md
marctc: I haven't seen any example like this in the current docs. We could change it, but for now better to be consistent, no?

component/prometheus/exporter/elasticsearch/elasticsearch.go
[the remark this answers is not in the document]
spartan0x117: You're right, I saw the check and the `Usage` in the docs and forgot about the default.

docs/sources/flow/reference/components/prometheus.exporter.elasticsearch.md
spartan0x117: The guide on writing docs hasn't always been followed (and quite a few things predate it), but the guidance is to make it copy-pastable and a full pipeline, when possible: https://github.com/grafana/agent/blob/main/docs/developer/writing-flow-component-documentation.md?plain=1#L494

spartan0x117: True, I'm not sure there's a stance on exactly _how_ copy-pastable the example needs to be, and it won't be possible to make it work out of the box for every case. But in this case it would be easy for someone to copy-paste the full example into a river config and would just need to replace the url.

It could also be made more generic with something like (which then wouldn't work if basic auth was needed):
```
prometheus.remote_write "example" {
  endpoint {
    url = env("REMOTE_WRITE_URL")
  }
}
```

marctc: I think your point it's valid, but out of the scope of this PR IMHO. Do you want open issue/PR to address this?

spartan0x117: Sure, sounds good to me :smile:

docs/sources/flow/reference/components/prometheus.exporter.elasticsearch.md
```diff
@@ -0,0 +1,107 @@
...
+title: prometheus.exporter.elasticsearch
+---
+
+# prometheus.exporter.elasticsearch
+The `prometheus.exporter.elasticsearch` component embeds
+[elasticsearch_exporter](https://github.com/prometheus-community/elasticsearch_exporter) for
+the collection of metrics from ElasticSearch servers.
+
+Note that currently, an Agent can only collect metrics from a single ElasticSearch server.
+However, the exporter is able to collect the metrics from all nodes through that server configured.
```
clayton-cornell, line 18:
```suggestion
{{% admonition type="note" %}}
Currently, an Agent can only collect metrics from a single ElasticSearch server.
However, the exporter can collect the metrics from all nodes through that server configured.
{{% /admonition %}}
```

docs/sources/flow/reference/components/prometheus.exporter.elasticsearch.md
```diff
@@ -0,0 +1,107 @@
...
+## Usage
+
+```river
+prometheus.exporter.elasticsearch "LABEL" {
+    address = "ELASTICSEARCH_ADDRESS"
+}
+```
+
+## Arguments
+The following arguments can be used to configure the exporter's behavior.
```
clayton-cornell, line 32:
```suggestion
You can use the following arguments to configure the exporter's behavior.
```

clayton-cornell (commented):
A couple of minor suggestions

PR #4223, opened 2023-06-21
prometheus.remote_write: do not fall back to upstream defaults

### LINKED ISSUE #4164: Flow prometheus.remote_write does not seem to pick up the correct defaults for QueueOptions
### What's wrong?

Hey there,
I have tried to get my head around this - and I am not sure where exactly this bug originates, but I think the current behaviour is not intended.

**Scenario**

We are currently optimizing our `remote_write` configuration in some OnPrem prometheus instances and wanted to take a look at our grafana-agent deployments too.

We looked at blogposts like [this one](https://grafana.com/blog/2021/04/12/how-to-troubleshoot-remote-write-issues-in-prometheus/) and wanted to verify that the `queue_config` inside the agent matches our expectations. So after reading the [grafana-agent docs](https://grafana.com/docs/agent/latest/flow/reference/components/prometheus.remote_write/#queue_config-block) we looked at the exported prometheus metrics like `prometheus_remote_storage_shard_capacity` - and were surprised that the default settings did not match the values documented.

By default we see a capacity of `2500` while the documentation states `10000`.
```
# HELP prometheus_remote_storage_shard_capacity The capacity of each shard of the queue used for parallel sending to the remote storage.
# TYPE prometheus_remote_storage_shard_capacity gauge
prometheus_remote_storage_shard_capacity{component_id="prometheus.remote_write.write_samples",remote_name="my-endpoint",url="http://host.k3d.internal:9009/api/prom/push"} 2500
```
This also applies to other metrics like `max_shards` or `max_samples_per_batch`.

After looking into the grafana-agent and prometheus code base we were getting confused:
- grafana/agent defines a capacity of [`10000` in `DefaultQueueOptions`](https://github.com/grafana/agent/blob/371470bc0e58c4f1620c96463f598e7045939d07/component/prometheus/remotewrite/types.go#L27)
  - This value has been increased (based on the upstream changes) in #3453
- prometheus/prometheus also defines a capacity of [`10000` in their `config.go`](https://github.com/prometheus/prometheus/blob/f399f386cef3b19c48157e9678a4f50997db3f41/config/config.go#L187)
  - That value has been increased in prometheus/prometheus#12203

**Possible explanation**

I looked into the `go.mod` to find out which upstream version of prometheus is being used by the agent. Apparently not the official prometheus/prometheus project is used but instead a fork from grafana: https://github.com/grafana/prometheus

Currently the grafana-agent includes version `release-2.42.0-grafana` of the forked repository: https://github.com/grafana/agent/blob/ba679b66dc6b0d706e1b5396439ac058ac70e36b/go.mod#L625

Looking at that forked version one can see that the `config.go` still contains the old defaults for QueueOptions: https://github.com/grafana/prometheus/blob/release-2.42.0-grafana/config/config.go#L185

So I assume that these upstream-fork settings somehow overwrite the `DefaultQueueOptions` inside the grafana/agent repository?

**Custom settings are being exported as prometheus metrics**

During my analysis I suspected that maybe the configuration from grafana-agent never gets exported as a prometheus metric - and that this would be the reason why the "old" (in the fork present) values were showing up.

But when you define a custom value (e.g. `5000`) for `capacity` this value gets successfully exported - inside the agent ui and the metrics endpoint:
```
prometheus.remote_write "write_samples" {
  endpoint {
    name = "my-customized-endpoint"
    url = "http://host.k3d.internal:9009/api/prom/push"
    queue_config {
      capacity = 5000
    }
  }
}
```

```
# HELP prometheus_remote_storage_shard_capacity The capacity of each shard of the queue used for parallel sending to the remote storage.
# TYPE prometheus_remote_storage_shard_capacity gauge
prometheus_remote_storage_shard_capacity{component_id="prometheus.remote_write.write_samples",remote_name="my-customized-endpoint",url="http://host.k3d.internal:9009/api/prom/push"} 5000
```

Screenshot 2023-06-15 123413

**Summary**

One can easily overwrite all default settings for `queue_options` so that they match the official/upstream values. But I guess that is not the intended behaviour - especially because the docs list other default values that are not correct.

### Steps to reproduce

Just deploy the Grafana Agent with the `prometheus.remote_write` integration and take a look at the metrics (e.g. at `localhost:80/metrics`).

Based on the default settings `prometheus_remote_storage_shard_capacity` should have a value of `10000` - but instead shows `2500`. This can be also verified with the other `queue_config` settings like `max_shards` or `max_samples_per_send`.

### System information

Windows 10 22H2 // k3d v5.4.6 // k3s version v1.24.4-k3s1

### Software version

Grafana Agent v0.34.0

### Configuration

```text
discovery.kubernetes "agents" {
  role = "pod"
  namespaces {
    names = ["default"]
  }
}

prometheus.scrape "scrape_agents" {
  targets    = discovery.kubernetes.agents.targets
  forward_to = [prometheus.remote_write.write_samples.receiver]
}

prometheus.remote_write "write_samples" {
  endpoint {
    name = "my-endpoint"
    url = "http://host.k3d.internal:9009/api/prom/push"
  }
}
```

### Logs

_No response_

rfratto:
When the `queue_config` or `metadata_config` blocks were fully unspecified, their values were inherited from the Prometheus package. This is an issue in cases where we purposefully changed the values we used.

A temporary workaround for the user is to specify empty blocks so the correct defaults are used (`queue_config {}`, `metadata_config {}`).

Fixes #4164.


component/prometheus/remotewrite/types.go
```diff
@@ -111,7 +111,7 @@ func (r *QueueOptions) SetToDefault() {
 
 func (r *QueueOptions) toPrometheusType() config.QueueConfig {
 	if r == nil {
-		return config.DefaultQueueConfig
+		return DefaultQueueOptions.toPrometheusType()
```
ptodev, line 114: I feel like something like it would be cleaner to create a new `QueueOptions` instead, call `SetToDefault()` and then call and return the result of `toPrometheusType()`. Otherwise we'd make the same error in the future.

rfratto: Can you show me a code snippet? I'm not sure how that's different than this.

mattdurham (approved):
LGTM

component/prometheus/remotewrite/types.go
ptodev: I meant something like this:
```
q := QueueOptions{}
q.SetToDefault()
return q.toPrometheusType()
```

The difference is that there is only one place in the code where the defaults are applied - in `SetToDefault()`. Having it in more than one place is what caused the original problem. I suppose this the snipped above is a bit ugly and inefficient, but maybe could get better with some refactoring.

Sorry for nitpicking, it's not a strong requirement on my side, just a suggestion 😄  Feel free to ignore if you're busy.

PR #4225, opened 2023-06-21
Added json_path function

### LINKED ISSUE #4171: river: Add jsonpath function
### Request

Add a [`jsonpath`](https://goessner.net/articles/JsonPath/index.html#e2) function similar to json_encode.

### Use case

Azure supports fetch instance metadata tags through [IMDS](https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=linux#sample-3-get-vm-tags). Request tags give a JSON array back.

For example:

```json
[
  {
    "name": "Department",
    "value": "IT"
  },
  {
    "name": "ReferenceNumber",
    "value": "123456"
  },
  {
    "name": "TestStatus",
    "value": "Pending"
  }
]
```

I would like to pass some of the tag values to `external_label` property. To fetch `Department` in this case, I would like to use a JSONPath expression like `$.[?(@.name == "Department")].value`.

## Alternatives

A map or loop function could also archive this, e.g. `[for tag in remote.http.tags.content: tag.value if tag.key == "Department"]` (HCL syntax)

Since JSONPath seems like a complex full features solution, I'm also open to alternatives.

jkroepke:
#### PR Description

This PR add a json_path function to stdlib. See #4171 for use cases.

Fixes #4171 

#### Notes to the Reviewer

Coming from https://cburgmer.github.io/json-path-comparison/, I choice github.com/ohler55/ojg. The library is MIT licensed, seems actively maintained and has zero dependencies.

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated


mattdurham (approved):
This looks good to me pending Clayton checking doc, great work!

docs/sources/flow/reference/stdlib/json_path.md
```diff
@@ -0,0 +1,36 @@
...
+> json_path("[{\"name\": \"Department\",\"value\": \"IT\"},{\"name\":\"TestStatus\",\"value\":\"Pending\"}]", "[?(@.name == \"Department\")].value")
+["IT"]
+
+> json_path("{\"key\": \"value\"}", ".nonexists")
+[]
+
+> json_path("{\"key\": \"value\"}", ".key")[0]
+value
+
+```
```
rfratto, line 36: As an aside: 

Wow, River really needs alternative ways of defining strings so users don't have to escape double quotes like this constantly.

jkroepke: I would recomend backticks here. Backticks are already used in other languages like go or js.

rfratto: Backticks (Go, JS) or single quotes (YAML) seem like fine options. 

Though, backticks in JS indicate an interpolated string, which I currently proposed a different syntax for in #2972. Maybe #2972 should change its recommendation to backticks?

jkroepke: hm.

Is there any downside by keeping the string interpolation for backticks + recommend them to avoid quote escaping?

In JS, the developers are using backtick to avoid escaping, too. (https://stackoverflow.com/questions/53180063/how-do-i-save-a-json-string-in-javascript-file). But JS developers are always a bit crazy.

An alternative would be textblock, like:

```python
i = """
{"hello": "world"}
"""
```

I know this from syntax from java, python and markdown (using 3 backticks)

But I personally like the backtick thing a bit more.

rfratto: In my prototype, I found that there's issues in general with using `"` anywhere in the interpolation character (`"` or `"""`), since the formatter code needs to recognize when it needs to escape inner `"` or not.

For that reason alone, to keep the syntax cleaner, using a backtick for interpolation might make sense, with support for escaping interpolation and escaping backticks with `\`. 

However, Go's version of backticks is much more literal, where you can't even escape backticks themselves. I guess that would be the downside of using backticks for string interpolation and avoiding double-quotes: it would mean that there is still no string character which gives you string literals without the need to escape anything (but maybe that's fine).

erikbaranowski: +1 backticks for me

rfratto: (We should probably move this discussion to #2972 but it does sound like backticks are probably a good choice for implementation simplicity, familiarity, and uniqueness to not need frequent escaping.)

docs/sources/flow/reference/stdlib/json_path.md
```diff
@@ -0,0 +1,36 @@
...
+
+# json_path
+
+The `json_path` function lookup values using [jsonpath](https://goessner.net/articles/JsonPath/) syntax.
+
+The function expects 2 strings. First string is the json string to lookup values from. Second string is the jsonpath expression.
+
+`json_path` always return a list of values. If the jsonpath expression does not match any values, an empty list is returned.
+
+> Remember to escape double quotes when passing JSON string literals to `json_path`.
```
tpaschalis, line 15: It's implied, but could we also mention that we _can_ pass the string export of another component directly if it's valid JSON?

jkroepke: I do a bit copy/paste from json_encode. Is this enough?

tpaschalis (approved):
Thanks, LGTM!

docs/sources/flow/reference/stdlib/json_path.md
```diff
@@ -0,0 +1,41 @@
...
+
+The `json_path` function lookup values using [jsonpath](https://goessner.net/articles/JsonPath/) syntax.
+
+The function expects 2 strings. First string is the json string to lookup values from. Second string is the jsonpath expression.
+
+`json_path` always return a list of values. If the jsonpath expression does not match any values, an empty list is returned.
+
+A common use case of `json_path` is to decode and filter the output of a [`local.file`][] or [`remote.http`][] component to a River value.
+
+> Remember to escape double quotes when passing JSON string literals to `json_path`.
```
clayton-cornell, line 17:
```suggestion
The function expects two strings. The first string is the JSON string used look up values. The second string is the jsonpath expression.

`json_path` always returns a list of values. If the jsonpath expression does not match any values, an empty list is returned.

A common use case of `json_path` is to decode and filter the output of a [`local.file`][] or [`remote.http`][] component to a River value.

> Remember to escape double quotes when passing JSON string literals to `json_path`.
```

clayton-cornell (commented):
Some suggestions for the doc

jkroepke:
@clayton-cornell thanks, commited.

PR #4235, opened 2023-06-22
config-reloader: change to non-root user

rootmout:
#### PR Description

Force the `config-reloader` container to run as root is not practical in platform that are under rigorous security constraints (ex: Openshift). 

Unless I am missing a specific situation, I suggest to setup the security context of this container, the same way the Prometheus Operator does. See [prometheus-operator/prometheus-operator config_reloader.go#L256-L262](https://github.com/prometheus-operator/prometheus-operator/blob/c840b7eaa857fc45f8a9be52cfd8d315f7c88e9f/pkg/operator/config_reloader.go#L256-L262) 

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added (was not a part mentioned in doc)
- [x] Tests updated


pkg/operator/resources_pod_template.go
```diff
@@ -207,14 +207,20 @@ func generatePodTemplate(
...
+	boolFalse := false
+	boolTrue := true
 	operatorContainers := []core_v1.Container{
 		{
 			Name:         "config-reloader",
 			Image:        imagePathConfigReloader,
 			VolumeMounts: volumeMounts,
 			Env:          envVars,
 			SecurityContext: &core_v1.SecurityContext{
-				RunAsUser: pointer.Int64(0),
```
jcreixell, line 217: @rfratto is there a specific reason why this was initially configured to run as root? do you see any problems with this?

rfratto: Probably just me being aggressive with copy and paste, I think it's safe to remove.

rfratto (approved):
Sorry this took so long to merge. Thanks for the contribution!

PR #4238, opened 2023-06-23
Disable blackbox metrics when integration is not enabled

### LINKED ISSUE #2627: blackbox metric is collected when blackbox integration is disabled
`blackbox_exporter_config_last_reload_successful: 0` metric returned by grafana agent 0.29 even if  blackbox_exporter integration is not enabled.

I think there could be two issues with this:
- We try to collect only metrics that are useful for users so, so users don’t pay too much if they use Grafana Cloud. Of course it just one metrics but…
- Also, since it has 0, it could trigger some blackbox exporter related alerts , as 0 actually means there is an issue with reloading config

marctc:
#### PR Description

This PR includes the last changes of blackbox exporter, which has this [PR](https://github.com/prometheus/blackbox_exporter/pull/1060). Those changes are to prevent that some of the metrics of blackbox exporter where automatically added to global metrics registry even if the blackbox integration was not enabled.

Fixes #2627

#### PR Checklist

- [x] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -130,7 +130,7 @@ require (
 	github.com/prometheus-operator/prometheus-operator v0.62.0
 	github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.63.0
 	github.com/prometheus-operator/prometheus-operator/pkg/client v0.63.0
-	github.com/prometheus/blackbox_exporter v0.24.0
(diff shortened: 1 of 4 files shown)


spartan0x117 (approved):
LGTM!

PR #4254, opened 2023-06-23
Add convert CLI command to grafana-agent

erikbaranowski:
#### PR Description

most of this PR is about how we are setting up the CLI and the documentation in/around the CLI. I'm open to feedback and opinions from anyone on what this should look like after you take a look at the PR.

we will need some more extensive docs about this work in a new Migration guide section of our docs covered by https://github.com/grafana/agent/issues/4073

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [ ] Tests updated


docs/sources/flow/reference/cli/convert.md
```diff
@@ -0,0 +1,33 @@
...
+file on disk with the converted results. `--write` can only be provided when
+`grafana-agent convert` is not reading from standard input.
+
+The command fails if the file being converted has syntactically incorrect
+configuration or cannot be converted to a Grafana Agent Flow river file.
+
+The following flags are supported:
+
+* `--write`, `-w`: Write the converted file back to disk when not reading from
+  standard input.
```
tpaschalis, line 29: Have we considered an `--output` flag instead? (and writing to stdout when it's not present)

IMHO overwriting the input configuration file might be confusing to users, cannot be retried, plus we have to keep track where the file came to check whether it can be used or not.

erikbaranowski: I agree, updated

docs/sources/flow/reference/cli/convert.md
```diff
@@ -0,0 +1,32 @@
...
+`grafana-agent convert` is not reading from standard input.
+
+The command fails if the file being converted has syntactically incorrect
+configuration or cannot be converted to a Grafana Agent Flow river file.
+
+The following flags are supported:
+
+* `--output`, `-o`: The filepath where the output is written.
+
+* `--source-format`, `-f`: The format of the source file. Supported formats: 'prometheus'.
```
clayton-cornell, line 30:
```suggestion
* `--source-format`, `-f`: Required. The format of the source file. Supported formats: 'prometheus'.
```

docs/sources/flow/reference/cli/convert.md
```diff
@@ -0,0 +1,32 @@
...
+The `--output` flag can be specified to write the contents of the converted
+file on disk to the specified path. `--output` can only be provided when
+`grafana-agent convert` is not reading from standard input.
+
+The command fails if the file being converted has syntactically incorrect
+configuration or cannot be converted to a Grafana Agent Flow river file.
+
+The following flags are supported:
+
+* `--output`, `-o`: The filepath where the output is written.
```
clayton-cornell, line 28:
```suggestion
* `--output`, `-o`: The filepath and filename where the output is written.
```

clayton-cornell (approved):

clayton-cornell (commented):
Docs are fine as-is. It matches the style/language of fmt. 

Future edit and rework could be done to clean up the language of all Agent CLI docs and clarify some things

docs/sources/flow/reference/cli/convert.md
```diff
@@ -0,0 +1,32 @@
...
+
+Usage: `grafana-agent convert [FLAG ...] FILE_NAME`
+
+If the `FILE_NAME` argument is not provided or if the `FILE_NAME` argument is
+equal to `-`, `grafana-agent convert` converts the contents of standard input. Otherwise,
+`grafana-agent convert` reads and converts the file from disk specified by the argument.
+
+The `--output` flag can be specified to write the contents of the converted
+file on disk to the specified path. `--output` can only be provided when
+`grafana-agent convert` is not reading from standard input.
```
tpaschalis, line 21: I don't think we need this restriction anymore with the `--output` flag, right?

cmd/internal/flowmode/cmd_convert.go
```diff
@@ -0,0 +1,141 @@
...
+				}
+				return fmt.Errorf("encountered errors during formatting")
+			}
+
+			return err
+		},
+	}
+
+	cmd.Flags().StringVarP(&f.output, "output", "o", f.output, "The filepath where the output is written.")
+	cmd.Flags().StringVarP(&f.sourceFormat, "source-format", "f", f.sourceFormat, "The format of the source file. Supported formats: 'prometheus'.")
```
tpaschalis, line 64: Is there an easy way to make this a required parameter? Otherwise it fails with the cryptic message
```
Error: (Error) unrecognized kind "" given to the config converter
```

tpaschalis (approved):
Just adding a couple of nits regarding ease of use.

Otherwise approving so I don't block things, I'm super happy we're having this!

cmd/internal/flowmode/cmd_convert.go
```diff
@@ -0,0 +1,141 @@
...
+	cmd.Flags().BoolVarP(&f.bypassWarnings, "bypass-warnings", "b", f.bypassWarnings, "Enable bypassing warnings when converting")
```
mattdurham, line 65: Do we have documentation on what `warnings` mean?

converter/diag/diagnostics.go
```diff
@@ -62,3 +62,16 @@ func (ds *Diagnostics) Add(severity Severity, message string) {
...
 	})
 }
+
+// Error implements error.
+func (ds Diagnostics) Error() string {
+	var errorMessage string
+	for ix, diag := range ds {
+		errorMessage += diag.Error()
+		if ix+1 < len(ds) {
+			errorMessage += "\n"
```
mattdurham, line 72: nit: How does this behave on `cmd.exe`?

docs/sources/flow/reference/cli/convert.md
```diff
@@ -0,0 +1,32 @@
...
+
+The command fails if the file being converted has syntactically incorrect
+configuration or cannot be converted to a Grafana Agent Flow river file.
```
mattdurham, line 24: nit: Should both uses of file be changed to config to handle stdin/stdout

mattdurham (approved):
Minor nits but looks good

docs/sources/flow/reference/cli/convert.md
erikbaranowski: Can you elaborate on this 1 or was it addressed with my last commit? I'm not sure what it is referring to in the context of this paragraph.

cmd/internal/flowmode/cmd_convert.go
```diff
@@ -0,0 +1,142 @@
...
+The -o flag can be used to write the formatted file back to disk. When -o is not provided, fmt will write the result to stdout.
+
+The -f flag can be used to specify the format we are converting from.
+
+The -b flag can be used to bypass warnings. Warnings are defined as 
+non-blocking issues identified during the conversion for feature[s] on the source file
+not supported in Grafana Agent Flow.`,
+		Args:         cobra.RangeArgs(0, 1),
+		SilenceUsage: true,
+		Aliases:      []string{"convert"},
```
rfratto, line 41: You don't need an alias to the same command name, this is for alternative names

cmd/internal/flowmode/cmd_convert.go
```diff
@@ -0,0 +1,142 @@
...
+
+func convertCommand() *cobra.Command {
+	f := &flowConvert{
+		output:         "",
+		sourceFormat:   "",
+		bypassWarnings: false,
+	}
+
+	cmd := &cobra.Command{
+		Use:   "convert [flags] file",
```
rfratto, line 25: If file is optional, then this should be:

```suggestion
		Use:   "convert [flags] [file]",
```

cmd/internal/flowmode/cmd_convert.go
```diff
@@ -0,0 +1,142 @@
...
+	if err != nil {
+		return err
+	}
+
+	riverBytes, diags := converter.Convert(inputBytes, converter.Input(fc.sourceFormat))
+	hasError := hasErrorLevel(diags, convert_diag.SeverityLevelError)
+	hasWarn := hasErrorLevel(diags, convert_diag.SeverityLevelWarn)
+	if hasError || (!fc.bypassWarnings && hasWarn) {
+		return diags
+	}
```
rfratto, line 113: I find it a bit surprising that the default behavior is that warnings are treated as errors, and you have to specifically bypass them.

IMO warnings should _not_ cause the conversion to fail, but you should instead allow a flag to fail on warnings.

cmd/internal/flowmode/cmd_convert.go
```diff
@@ -0,0 +1,142 @@
...
+
+	if fc.output == "" {
+		_, err := io.Copy(os.Stdout, &buf)
+		return err
+	}
+
+	wf, err := os.OpenFile(fc.output, os.O_WRONLY|os.O_TRUNC|os.O_CREATE, 0666)
+	if err != nil {
+		return err
+	}
```
rfratto, line 126: nit: use `os.Create`, which will give you the same effect as this.

converter/diag/diagnostics.go
```diff
@@ -62,3 +62,16 @@ func (ds *Diagnostics) Add(severity Severity, message string) {
...
+		}
+	}
+
+	return errorMessage
```
rfratto, line 76: nit: use `strings.Builder` instead, which is more efficient and also handles newlines for you:

```suggestion
	var sb strings.Builder
	for ix, diag := range ds {
		fmt.Fprint(&sb, diag.Error())
		if ix+1 < len(ds) {
			fmt.Fprintln(&sb) 
		}
	}

	return sb.String()
```

docs/sources/flow/reference/cli/convert.md
```diff
@@ -0,0 +1,35 @@
+---
+title: grafana-agent convert
+weight: 100
+---
+
+# `grafana-agent convert` command
+
+The `grafana-agent convert` command converts a supported configuration file format
+to a Grafana Agent Flow river file.
```
rfratto, line 9: River is a proper noun:

```suggestion
to a Grafana Agent Flow River file.
```

docs/sources/flow/reference/cli/convert.md
```diff
@@ -0,0 +1,35 @@
...
+
+The `--output` flag can be specified to write the contents of the converted
+file on disk to the specified path.
+
+The `--bypass-warnings` flag can be used to bypass warnings. Warnings are defined as 
+non-blocking issues identified during the conversion for feature[s] on the source file
+not supported in Grafana Agent Flow.
+
+The command fails if the file being converted has syntactically incorrect
+configuration or cannot be converted to a Grafana Agent Flow river file.
```
rfratto, line 27:
```suggestion
configuration or cannot be converted to a Grafana Agent Flow River file.
```

docs/sources/flow/reference/cli/convert.md
```diff
@@ -0,0 +1,35 @@
...
+
+The following flags are supported:
+
+* `--output`, `-o`: The filepath and filename where the output is written.
+
+* `--source-format`, `-f`: Required. The format of the source file. Supported formats: 'prometheus'.
```
rfratto, line 33: I think we're burying the lede here with a lack of explanation for what `prometheus` here means. 

I'd like to see a section for supported source formats, which go into a little more detail about them. For example, we should document:

* What version of Prometheus is used for conversion, potentially even linking to its documentation
* What broad features are/aren't supported (recording rules, alerting rules, etc)
* Anything else you might think is useful to know about the Prometheus conversion

rfratto (approved):
Nice work! I'm excited to start playing around with this.

cmd/internal/flowmode/cmd_convert.go
erikbaranowski: I think it's a fair point on the semantics. That being said, I think it's important that our default behavior for running the agent is not to start up automatically if we have unsupported features in the conversion and I was making this behave the same way for consistency.

I'm open to suggestion here if we can change the semantics in a way that makes sense and gives us the behavior we want. Do you agree on the behavior when running should not start without an explicit flag if something in the source file is unsupported?

docs/sources/flow/reference/cli/convert.md
erikbaranowski: @clayton-cornell what do you think about the best way to organize this information? I was anticipating most of these details would exist in the migration guide for https://github.com/grafana/agent/issues/4073 . If we want to include that work as part of this PR before merge I think that's also fair.

I don't have a strong opinion about the location of this information, but I do agree we want it before I consider the converter 1.0 'done' end to end.

cmd/internal/flowmode/cmd_convert.go
erikbaranowski: An alternative brainstorm is to have them also be errors but introduce an error type to the Diagnostic so we can differentiate and add the right semantic flag for ignoring them.

docs/sources/flow/reference/cli/convert.md
erikbaranowski: Proposal: do we want to make this the staging branch and do the full docs/run CLI changes to this branch then merge it to main all together?

Let's still coalesce on the convert CLI implementation on this PR as written.

cmd/internal/flowmode/cmd_convert.go
rfratto: After talking about this offline, I think we need more time to explore what types of warnings we might generate, and if all of them fall into the same category of needing investigation or being OK for the user to ignore. 

For example, are the warnings for these two the same severity: 

1. Alerting is configured in Prometheus but no equivalent alerting config was generated in Flow
2. The wal_directory argument in static mode is ignored because the WAL storage happens at a higher level in Flow 

To me, the first one seems more fatal than the second, while the second one is more of a actual heads-up style of warning. 

We'll be able to classify these better after we're done with writing all the converters, so my suggestion for now would be to leave the behavior as-is and revisit changing whether warnings are fatal later, since making them non-fatal would be a backwards compatible change, but the inverse wouldn't.

docs/sources/flow/reference/cli/convert.md
erikbaranowski: I'm going to draft some additional docs for this PR and let's go from there

docs/sources/flow/reference/cli/convert.md
mattdurham: IE the input can be stdin and output can be stdout and not just files. 

```
The command fails if the config being converted has syntactically incorrect
configuration or cannot be converted to a Grafana Agent Flow river config.
```

Its a super minor opinion and if you think file is clearer keep it.

erikbaranowski: got it now, thanks for explaining. I'm updating throughout for my next pass with extra docs.

docs/sources/flow/reference/cli/convert.md
clayton-cornell: @erikbaranowski This is something we've chatted a little about in an offline 1:1 and in messages and it's in-0line with my earlier concerns about this not really providing all info and being confusing to me when I read it. I opted to say "it's follwoing the `fmt` doc, so... I'll leave it as-is for now. It could definitely have a rethink - and if we do document this a bit differently, we should revisit (in another PR?) the `fmt` doc... and maybe `run` as well.

For here, maybe this: follow manpage style for the CLI documentation. Stick to the basics... this is the command and syntax, and here are the arguments you can use... (these are required, these are optional) followed by a common usecase example.

Then for more details on implementation and so on, we shift that to a migration guide that explains the "why woudl I use this" in more detail.

For where to include... keep the PRs simple.  We can keep this for code changes and the basic CLI doc content.  This makes reviews a bit more manageable (bite sized).

erikbaranowski: thanks all, @clayton-cornell can you take 1 more pass on the last commit and see if you want to tune anything for now. I appreciate all the feedback and hope to keep iterating on this.

docs/sources/flow/reference/cli/convert.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Usage
+
+Usage: `grafana-agent convert [FLAG ...] FILE_NAME`
+
+If the `FILE_NAME` argument is not provided or if the `FILE_NAME` argument is
+equal to `-`, `grafana-agent convert` converts the contents of standard input. Otherwise,
+`grafana-agent convert` reads and converts the file from disk specified by the argument.
+
+The `--output` flag can be specified to write the contents of the converted
```
ptodev, line 19: Not sure why we list `--output` and `--bypass-warnings` both here and in the list below? I think there's no need to list them here, we could just have them in the list.

docs/sources/flow/reference/cli/convert.md
```diff
@@ -0,0 +1,68 @@
...
+property is left off the Flow output.
+* If a non-provided source config value default matches a Flow default value,
+the property is left off the Flow output.
+* If a non-provided source config value default doesn't match a Flow default
+value, the Flow default value is included in the Flow output.
+
+### Warnings
+
+Warnings are defined as non-blocking issues identified during the conversion
+for feature[s] on the source file not supported in Grafana Agent Flow. A
```
ptodev, line 51: > Warnings are defined as non-blocking issues identified during the conversion
for feature[s] on the source file not supported in Grafana Agent Flow.

If the source config uses a feature which is not available in Flow, then wouldn't that have to be an error? If the feature is missing from the Flow config, wouldn't that mean it has reduced functionality?

cmd/internal/flowmode/cmd_convert.go
ptodev: It might also help to have different categories of errors and warnings and document them. Or we could have different error codes for each type of error/warning. Then the customers could choose which warnings/errors to enable or disable. It'd also make the docs more clear about what kinds of issues could be reported.

ptodev (approved):
I left a few minor comments, but I'm happy for the PR to be merged even if they're not addressed.

Issue #4233, opened 2023-06-22
Add a component that instead of using the Kubernetes API allows utilizing the local API


mattdurham:
### Background

Hitting the API constantly can put significant strains on the Kubeneretes API when the local kubelet instance already has much of the data needed. 

### Proposal

I propose we have a component that mirrors the kubernetes api that allows grabbing that data, there are some concerns over the stability of the API but we can likely limit or make rails on what it works with and doesnt.

rfratto:
I'm not against this, but I'm not sure if it's possible, so it needs investigation. 

Last I checked, Kubelets have an undocumented API, which implies (to me) that there's no guarantees that endpoints will work between releases, or even return the same data.

mattdurham:
It is undocumented but looks stableish, so there are real concerns there. https://github.com/cyberark/kubeletctl seems to have wrappers around the api and likely can be reviewed to see how unstable it actually is.

rfratto:
> I propose we have a component that mirrors the kubernetes api that allows grabbing that data,

What data do we want to grab from Kubelets? I think the only thing they'll have available on them is a list of running pods.

mattdurham:
Pods are the most interesting data and the best first case, though https://github.com/cyberark/kubeletctl/blob/master/API_TABLE.md lists other things that we might find value in.

rfratto:
Thanks, that table is helpful. 

I did some investigation and found the following:

* Code for all the current API endpoints from the server is located in [pkg/kubelet/server/server.go](https://github.com/kubernetes/kubernetes/blob/9325a57125e8502941d1b0c7379c4bb80a678d5c/pkg/kubelet/server/server.go). 
* The `/pods` endpoint returns the list of pods (also listed in the API table link above). This is probably where we want to start; not much else looks immediately useful. 
  * However, in the future, it might make sense for `loki.source.kubernetes` and `loki.source.podlogs` to talk directly to Kubelets for logs too, rather than using the proxy from the API server like they do today. 
* There is no "watch" capabilities for the `/pods` API endpoint. We would have to do polling to get changes over time. 

Things I still don't know:

* How can a container get access to the Kubelet API? I know you can proxy requests to it from the API server, but that's what we're trying to avoid. Is there a way containers can talk directly to their hosting Kubelets? 
* When talking directly to a Kubelet, it's not clear what authentication we would need to use. 

Minus the open questions above, it looks like this component is theoretically possible.

mikemykhaylov:
[...]

PR #4255, opened 2023-06-23
Implement new `discovery.kubelet` integration

gcampbell12:
#### PR Description

Implementation for https://github.com/grafana/agent/issues/4233, uses the kubelet `/pods` endpoint to discover pods running on the local host as scrape targets.

https://github.com/grafana/agent/issues/4233

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated


component/discovery/kubelet/kubelet.go
```diff
@@ -0,0 +1,268 @@
...
+		Exports: discovery.Exports{},
+		Build: func(opts component.Options, args component.Arguments) (component.Component, error) {
+			return New(opts, args.(Arguments))
+		},
+	})
+}
+
+// Arguments configures the discovery.kubernetes component.
+type Arguments struct {
+	KubeletURL       config.URL                        `river:"kubelet_url,attr"`
```
rfratto, line 69: I had some open questions over in https://github.com/grafana/agent/issues/4233#issuecomment-1603281700 where I'm still not sure of the answers:

> * How can a container get access to the Kubelet API? I know you can proxy requests to it from the API server, but that's what we're trying to avoid. Is there a way containers can talk directly to their hosting Kubelets?
> * When talking directly to a Kubelet, it's not clear what authentication we would need to use.

This is stuff I think we should probably end up documenting, so can you help me understand these points? 

How would someone set this URL correctly in Kubernetes? What would they need to do to make the Kubelet accessible (if anything?)

And where would the authentication come from?

rfratto (commented):
Thanks! At a cursory glance, this looks good, but I still have some questions about what a user would need to do to get this component to work.

(I would need to give this a slightly deeper review once it's out of draft to validate the labels and client code)

component/discovery/kubelet/kubelet.go
gcampbell12: Hey @rfratto thanks for taking a look!
- This is going to depend on how you have networking setup between hosts and pods, but the kubelet has an `--address` flag to bind it to an interface and then exposes a REST https server on 10250 by default (the port can be changed with `--port`) so if your agent pod can route to `<host-ip>:10250` and you bind the kubelet to `0.0.0.0` this will work, or if you have your grafana-agent daemonset host networked this also works. 
The `/pods` endpoint on the kubelet I'm using in this PR is essentially what the API server proxies too, this PR bypasses the API server and hits the kubelets directly. Essentially what we need to call out in the docs is that the grafana-agent pod must be able to reach the kubelet port for this to work.
- Authentication wise you can use a normal bearer token from a `serviceaccount` you just need `get` permissions on `nodes/proxy` resource in the grafana-agent cluster role, client certs should also work. Theres an also an option to enable an unauthed read only port on the kubelet, but it's not recommended.

rfratto: Thanks for helping me understand! 

> Authentication wise you can use a normal bearer token from a serviceaccount you just need get permissions on nodes/proxy resource in the grafana-agent cluster role, client certs should also work. Theres an also an option to enable an unauthed read only port on the kubelet, but it's not recommended.

Oh interesting! I was aware of the nodes/proxy resource but I didn't realize that it could also be used to talk to the nodes directly. Good to know 👍

mattdurham:
Going to test this today.

docs/sources/flow/reference/components/discovery.kubelet.md
```diff
@@ -0,0 +1,157 @@
...
+`discovery.kubelet` does not expose any component-specific debug metrics.
+
+## Examples
+
+### Bearer token file authentication
+
+This example uses a bearer token file to authenticate to the Kubernetes API:
+
+```river
+discovery.kubernetes "k8s_pods" {
```
mattdurham, line 141: Same here with needs to be `discovery.kubelet`

gcampbell12: Oops too much copy and paste, fixed!

docs/sources/flow/reference/components/discovery.kubelet.md
```diff
@@ -0,0 +1,147 @@
...
+
+One of the following authentication methods must be provided if kubelet authentication is enabled
+ - [`bearer_token` argument](#arguments).
+ - [`bearer_token_file` argument](#arguments).
+ - [`authorization` block][authorization].
+
+### namespaces
+
+The `namespaces` list limits the namespaces to discover resources in. If
+omitted, all namespaces are searched.
```
rfratto, line 45: It's a little unusual for arguments to have their own header; normally they would just be described in the same header as the arguments or the block they're in:

```suggestion
The `namespaces` argument limits the namespaces to discover resources in. If
omitted, all namespaces are searched.
```

component/discovery/kubelet/kubelet.go
```diff
@@ -0,0 +1,260 @@
...
+		Exports: discovery.Exports{},
+		Build: func(opts component.Options, args component.Arguments) (component.Component, error) {
+			return New(opts, args.(Arguments))
+		},
+	})
+}
+
+// Arguments configures the discovery.kubernetes component.
+type Arguments struct {
+	KubeletURL       config.URL              `river:"kubelet_url,attr"`
```
mattdurham, line 67: nit: I wonder if `kubelet_url` is redundant and go to url instead, and if we should default to `https://localhost:10250`.

docs/sources/flow/reference/components/discovery.kubelet.md
```diff
@@ -0,0 +1,145 @@
...
+
+The following fields are exported and can be referenced by other components:
+
+Name | Type | Description
+---- | ---- | -----------
+`targets` | `list(map(string))` | The set of targets discovered from the Kubelet API.
+
+Each target includes the following labels::
+
+* `__meta_kubernetes_namespace`: The namespace of the pod object.
```
mattdurham, line 76: Should we add  `__address__` should we add it to this list? Also given this is an internal API should there be a disclaimer that these are not guaranteed?

docs/sources/flow/reference/components/discovery.kubelet.md
```diff
@@ -0,0 +1,145 @@
+---
+title: discovery.kubelet
```
mattdurham, line 2: I would like to label this has BETA until we get some more confidence in using it.  Add 
```
labels:
  stage: beta
```
under the title

mattdurham (changes requested):
Fantastic, a few minor things but looks really close.

component/discovery/kubelet/kubelet.go
gcampbell12: Yep I agree this makes sense, added this via the `SetToDefault` method used elsewhere

docs/sources/flow/reference/components/discovery.kubelet.md
gcampbell12: I've added the address label to the list, when you say not guaranteed do you mean we should call out these may be subject to change or something else?

component/discovery/kubelet/kubelet.go
gcampbell12: Ah I think you want to rename the option to `url` too?

gcampbell12: Have renamed, let me know if I misunderstood!

docs/sources/flow/reference/components/discovery.kubelet.md
gcampbell12: Have added the address label and added a note around the Kubelet API being internal

component/discovery/kubelet/kubelet.go
```diff
@@ -0,0 +1,278 @@
...
+		Name:    "discovery.kubelet",
+		Args:    Arguments{},
+		Exports: discovery.Exports{},
+		Build: func(opts component.Options, args component.Arguments) (component.Component, error) {
+			return New(opts, args.(Arguments))
+		},
+	})
+}
+
+// Arguments configures the discovery.kubernetes component.
```
mattdurham, line 70: nit: discovery.kubernetes

mattdurham (changes requested):
One last comment change and I think its good to go

component/discovery/kubelet/kubelet.go
gcampbell12: nice catch, fixed!

mattdurham (approved):
LGTM

Issue #4253, opened 2023-06-23
Proposal: Flow services


rfratto:
> **NOTE**: This is a very long proposal. Please read it in its entirety before commenting.

In the context of Flow, a "service" is a low-level primitive which exposes additional behavior on top of the Flow controller. Today, there are three services that exist in `pkg/flow`: 

* The *telemetry* service, allowing components to create metrics, logs, and traces for component observability.
* The *HTTP* service, allowing components to expose an HTTP endpoint on a shared server for things such as Prometheus exporters. 
* The *clustering* service, allowing components to distribute work amongst themselves. 
	* The clustering service depends on the HTTP service to function properly. 

All of the services today are started prior to the Flow controller. 

One of the remaining items for Flow feature parity is adding support for [configuring TLS settings for the HTTP service at runtime](https://github.com/grafana/agent/issues/2715). This would be added as a new config block into the Flow configuration file. This leads to a minimum set of functional requirements: 

* For security reasons, plaintext HTTP traffic must only be accepted when a root config is loaded that does not configure TLS. 
* If the TLS settings reference a component export (such as `remote.vault`), those components must be evaluated first to be able to evaluate the TLS settings properly. 
* TLS settings must not have a cyclic dependency: it is not valid for the TLS settings to depend on a component which depends on the HTTP service to function.
	* For example, `prometheus.exporter.unix` depends on the HTTP service to create its targets. 

These functional requirements extend to the clustering service, which must only start after the HTTP service is ready. 

To allow services to be configured at runtime, those services (any any service which depends on them) must now exist in the DAG to ensure proper evaluation order; today, only the telemetry service exists in the DAG. 

Additionally, there has been [ongoing work](https://github.com/grafana/agent/pull/3981) to prepare Grafana Agent for a 1.0 release. This included some preliminary offline discussions about moving Flow and River to a separate repository so they may have their own versions separate from Grafana Agent. 

This proposal aims to solve both problems at once: 

1. Allow TLS settings to be configurable at runtime. 
2. Decouple the agent-specific HTTP and clustering services from `pkg/flow` in preparation for Flow and River to be moved to a dedicated Git repository. 

## Goals 

* Allow TLS settings to be configurable at runtime. 
* Ensure that plaintext HTTP traffic is only accepted when TLS is not configured. 
* Prevent cyclic dependencies between services and components. 
* Allow services to depend on other services. 
* Decouple the HTTP and clustering service from  `pkg/flow`. 
[...]

PR #4258, opened 2023-06-24
POC: Implement Auth Basic for Web API

### LINKED ISSUE #3601: Add Auth Basic to Grafana Agent API
Hi,

After working some time with grafana-agent, I identified that all exporter modules (at least in flow mode) are exposing the metrics endpoint trough the internal grafana-agent http server.

I would like to ask if its possible to protect the APIs (and UI) by Auth Basic. This may protects unauthorized access from other contexts on the machine, if grafana-agent is natively installed on a system. If I install grafana-agent on each machine, I have some worries that a potential pen test could raise this issue.

Auth Basic over HTTP is insecure, but more secure compared to no authentication. A https listener for the grafana-agent might be also a good solution to add full protection.

jkroepke:
#### PR Description

This PR introduce Auth Basic option for flow mode.

Until Grafana Agent supports TLS, auth basic may insecure. However, the endpoint runs on location my default. An admin which can intercept the traffic may able to read the password file, too.

The Cluster API is auth basic protected, too. If one cluster node enables auth basic, all cluster members requires the same auth basic username and password combination.

For connection via in-memory listener, auth basic is disabled.

GO Profiling works with auth basic:

> go tool pprof http://user:pass@127.0.0.1:12345/debug/pprof/profile

Fixes #3601 

#### Notes to the Reviewer

Go code may goofy at this time.

DOCS and CHANGELOG is missing at the moment.

For self metrics scrape, the docs should be updated to:

```
prometheus.scrape "example" {
  // Collect metrics from Grafana Agent's default listen address.
  targets = [{
    __address__ = "agent.internal:12345",
  }]

  forward_to = []
}
```

because `agent.internal:12345` goes over the in-memory listener and does not require authentication.

Injecting the middleware on router level would be an better option.

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated


jkroepke:
This cluster mode is not tested and it discussable if it should be protected by auth basic.

rfratto:
I'm in the middle of implementing #4253, which would allow this to be configurable in the config file instead of command line flags. 

Is there any desire to have this configurable in the config file instead of flags? If so, I think we should wait for #4253 so we don't have to have a breaking change for this.

jkroepke:
Config file looks better to me.
I just feel not confident enough to implement a flow component and a reloading mechanismen. 🙂

PR #4262, opened 2023-06-26
Update create-config-file.md

marybelvargas:
I believe the WAL is NOT required but encouraged. If it is required then might be worth to update the wording on this statement. Probably something like:  

> This is required and it ensures that samples will be redelivered in case of failure (e.g., network issues, machine reboot).


clayton-cornell (approved):
Agreed. This should state "not required" just to make the sentence and concept make sense.

clayton-cornell:
@grafana/grafana-agent-core-maintainers Can someone on the team verify that WAL is NOT required (but is useful) with Agent on Windows?

mattdurham:
A WAL is essentially required in the Agent. Technically if you are NOT doing metrics then you don't need a WAL, but given that metrics are our most common use case I would say it is required in Agent Static Mode.

mattdurham:
In rereading this there are some statements that are incorrect with the original text. A WAL currently does not protect against machine reboots/agent restarts. Its an open issue with Prometheus.

docs/sources/static/configuration/create-config-file.md
```diff
@@ -42,7 +42,7 @@ integrations:
 ```
 
 In this example, we first must configure the `wal_directory` which is used to
-store metrics in a Write-Ahead Log. This is required, but ensures that samples
+store metrics in a Write-Ahead Log. This is not required, but ensures that samples
```
clayton-cornell, line 45:
```suggestion
store metrics in a Write-Ahead Log (WAL). WAL is required and ensures that samples
```

docs/sources/static/configuration/create-config-file.md
```diff
@@ -42,7 +42,7 @@ integrations:
...
 will be redelivered in case of failure (e.g., network issues, machine reboot). We
```
clayton-cornell, line 46:
```suggestion
will be redelivered in case of failures, for example if there are network connectivity issues. We
```

@mattdurham would this suggestion cover the point you raised about reboots/restarts not handled by WAL?

mattdurham: Yes, that is primary use case.

clayton-cornell:
@marybelvargas Over to you with the suggestions (combined with Matt's input). If you're happy with the suggestions (and apply them), ping me back and I can finalize and merge.

marybelvargas:
Hi @clayton-cornell,

I'm good with the changes. Thank you!!

clayton-cornell:
Merging

clayton-cornell:
I am closing this PR because the CLA has not been signed by everyone. I cannot merge this PR without the CLA requirement. 

I've included this change in another PR: https://github.com/grafana/agent/pull/4355 that I'm working on so the correction is not lost/forgotten.

Issue #3979, opened 2023-05-25
Flow: Add compression details to any read/write component.


mattdurham:
### Request

Add compression details to any read/write component, ie `remote_write` should specify that it uses snappy compression

### Use case

Better document our components

PR #4264, opened 2023-06-26
Add technical details on compression for primary logs, traces, metrics

mattdurham:
#### PR Description

Quick PR to add technical details to our primary in/outs for logs, traces and metrics.

Addresses the primary use case of #3979 

There are a TON of other protocols and paths but this should represent the primary in and outs that are most commonly used.

#### PR Checklist

- [ ] CHANGELOG updated
- [X] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/flow/reference/components/loki.write.md
+++ b/docs/sources/flow/reference/components/loki.write.md
@@ -154,3 +154,7 @@ loki.write "local" {
     }
 }
 ```
+
+## Technical details
+
+`loki.write` uses [snappy](https://en.wikipedia.org/wiki/Snappy_(compression)) for compression.
--- a/docs/sources/flow/reference/components/otelcol.receiver.otlp.md
+++ b/docs/sources/flow/reference/components/otelcol.receiver.otlp.md
@@ -217,3 +217,7 @@ otelcol.exporter.otlp "default" {
   }
 }
 ```
+
+## Techincal details
+
+`otelcol.receiver.otlp` supports [gzip](https://en.wikipedia.org/wiki/Gzip) for compression.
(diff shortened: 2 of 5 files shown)


mattdurham:
Used verbiage for `uses` to indicate that in our usage it always uses something and supports to indicate it depends on setting/headers.

mattdurham:
Added the technical details at the bottom but totally open to changing that.

docs/sources/flow/reference/components/loki.write.md
```diff
@@ -154,3 +154,7 @@ loki.write "local" {
     }
 }
 ```
+
+## Technical details
```
clayton-cornell, line 158:
```suggestion
## File compression
```

@mattdurham Could we be more specific for the heading (and in all other topics here)?  Or are we planning to add other information beyond just file compression support?

mattdurham: I was thinking it could be more generic and grow over time but that might be to forward looking. I will change `Compression` and we can expand later if needed.

mattdurham: FYI I dont want to use file compression since its really line level compression

clayton-cornell: The title makes perfect sense as just `Compression`.

docs/sources/flow/reference/components/otelcol.receiver.otlp.md
```diff
@@ -217,3 +217,7 @@ otelcol.exporter.otlp "default" {
   }
 }
 ```
+
+## Techincal details
```
clayton-cornell, line 221:
```suggestion
## Compression
```

clayton-cornell (approved):
All good. One small change for a missed heading title change.

PR #4266, opened 2023-06-27
clustering: add basic UI page for peer info

### LINKED ISSUE #3483: Provide UI-level debugging information for clustering
Distributed systems are hard. 

We should provide a way for users to debug and inspect the status of their clusters via the UI and possibly navigate from node to node.
The ability to know the peer addresses and call their endpoints will be helpful here.

tpaschalis:
#### PR Description
This PR adds a first version of a Flow UI page that contains information around Agent Clustering.

* It introduces a new /api/v0/web/peers endpoint that is wired to call the Peers() method of the Flow controller's clusterer.
* It copies the Typescript/React `ComponentList` code and adds the hooks and types to populate with the API response
* It adds a new page under /clustering and adds it to the NavBar

image

One thing that I'd like to do here is to be able to expand the width of the "Node Name" column but so far I've had no luck 🤦 

Another thing that is potentially interesting is to show/hide the page from the NavBar and conditionally render it only when clustering is _enabled_ and we haven't fallen back to the localNode implementation. Do you think it's worth it?

Fixes #3483 

#### Notes to the Reviewer
* It's my first time ever touching our React-based UI, so please advice about things that could/should be done differently.
* This depends on https://github.com/grafana/ckit/pull/22 to be merged. Alternatively we could introduce a lightweight wrapper around it on our end here.
* Ideally, this table should be sortable/filterable in the future. Similarly, we could explore if we could make our helm chart so that it's easy to navigate from one node's UI to another via the UI.

#### PR Checklist

- [x] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/web/api/api.go
+++ b/web/api/api.go
@@ -11,23 +11,26 @@ import (
 
 	"github.com/gorilla/mux"
 	"github.com/grafana/agent/component"
+	"github.com/grafana/agent/pkg/cluster"
 	"github.com/prometheus/prometheus/util/httputil"
 )
 
 // FlowAPI is a wrapper around the component API.
 type FlowAPI struct {
 	flow component.Provider
+	node cluster.Node
 }
 
 // NewFlowAPI instantiates a new Flow API.
-func NewFlowAPI(flow component.Provider) *FlowAPI {
-	return &FlowAPI{flow: flow}
+func NewFlowAPI(flow component.Provider, node cluster.Node) *FlowAPI {
+	return &FlowAPI{flow: flow, node: node}
 }
 
 // RegisterRoutes registers all the API's routes.
 func (f *FlowAPI) RegisterRoutes(urlPrefix string, r *mux.Router) {
 	r.Handle(path.Join(urlPrefix, "/components"), httputil.CompressionHandler{Handler: f.listComponentsHandler()})
 	r.Handle(path.Join(urlPrefix, "/components/{id}"), httputil.CompressionHandler{Handler: f.getComponentHandler()})
+	r.Handle(path.Join(urlPrefix, "/peers"), httputil.CompressionHandler{Handler: f.getClusteringPeersHandler()})
 }
 
 func (f *FlowAPI) listComponentsHandler() http.HandlerFunc {
@@ -65,6 +68,20 @@ func (f *FlowAPI) getComponentHandler() http.HandlerFunc {
 		}
 
 		bb, err := json.Marshal(component)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+		_, _ = w.Write(bb)
+	}
+}
+
+func (f *FlowAPI) getClusteringPeersHandler() http.HandlerFunc {
+	return func(w http.ResponseWriter, _ *http.Request) {
+		// TODO(@tpaschalis) Detect if clustering is disabled and propagate to
+		// the Typescript code (eg. via the returned status code?).
+		peers := f.node.Peers()
+		bb, err := json.Marshal(peers)
 		if err != nil {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return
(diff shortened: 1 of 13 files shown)


web/ui/src/features/clustering/PeerList.tsx
```diff
@@ -0,0 +1,45 @@
...
+          <span className={styles.idName}>{name}</span>
+        </td>
+        <td>
+          <span className={styles.idName}>{addr}</span>
+        </td>
+        <td>
+          <span className={styles.idName}>{state}</span>
+        </td>
+        <td>
+          <span> {isSelf ? '✅' : '🚫'}</span>
```
mattdurham, line 32: Is there anyway to get the distribution of activity? Might be entirely out of scope but I think dskit has something like it.

mattdurham (approved):
Conditionally approve but think there are a few suggestions that might need to be made.

web/ui/src/features/clustering/PeerList.tsx
tpaschalis: Hm, no, since there's no central KV store but each node uses the hashing function whenever it needs to, there's currently no way to estimate what the current distribution is.

A potentially interesting thing here is to be able to list the components that make use of clustering, calculate the percentage of targets owned and navigate to their details page for more fine-grained information.

web/api/api.go
```diff
@@ -11,23 +11,26 @@ import (
...
 // FlowAPI is a wrapper around the component API.
 type FlowAPI struct {
 	flow component.Provider
+	node cluster.Node
 }
 
 // NewFlowAPI instantiates a new Flow API.
-func NewFlowAPI(flow component.Provider) *FlowAPI {
-	return &FlowAPI{flow: flow}
+func NewFlowAPI(flow component.Provider, node cluster.Node) *FlowAPI {
```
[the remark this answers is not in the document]
tpaschalis: You're right; `cluster.Node` is indeed an interface here

What I wasn't sure is if it makes sense to create a new abstraction here, but for now I'm just passing the clusterer implementation we create in main.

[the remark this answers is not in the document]
mattdurham: Ack sorry! I looked at cluster.node and not the big N.

web/ui/src/features/clustering/PeerList.tsx
mattdurham: That's a fair trade off. Might be something to consider in the future having each node expose debug info that's not shared via ckit.

web/ui/src/pages/Clustering.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+import PeerList from '../features/clustering/PeerList';
+import Page from '../features/layout/Page';
+import { usePeerInfo } from '../hooks/peerInfo';
+
+function PageClusteringPeers() {
+  const peers = usePeerInfo();
+
+  return (
+    <Page name="Clustering" desc="List of clustering peers." icon={faCubes}>
```
rfratto, line 11: nit: Don't put a period on an h2:

```suggestion
    <Page name="Clustering" desc="List of clustering peers" icon={faCubes}>
```

web/ui/src/features/clustering/Table.module.css
```diff
@@ -0,0 +1,34 @@
...
+.table th {
+  padding: 8px;
+  text-align: left;
+  background-color: #f4f5f5;
+  color: rgba(36, 41, 46, 0.75);
+}
+
+.table td:first-child {
+  width: 40%;
+}
```
rfratto, line 34: I think this will have problems; we ideally want the table to be a little more fluid than this.

rfratto (approved):
Nice!

web/ui/src/features/clustering/types.ts
```diff
@@ -0,0 +1,9 @@
+export interface PeerInfo {
+  name: string;
+
+  addr: string;
+
+  state: string;
+
+  isSelf: boolean;
```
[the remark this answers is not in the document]
tpaschalis: 📈 my LoC metrics of course!
(they used to have docstrings and spaces when I copied them over, which I never added back 😅)

PR #4278, opened 2023-06-28
[Agent Management] Add Label Management

spartan0x117:
#### PR Description
In order to add HTTP headers to the request, `FetchRemoteConfig` had to be changed to make raw HTTP requests, rather than using the code in the `remote_config` package. The behavior (in terms of metrics exposed and `Retry-After`) should be the same.

Other than that the primary change is here (and in the `RemoteConfiguration` struct to add the new fields):
```go
// Set headers for label management, if enabled
if r.InitialConfig.RemoteConfiguration.LabelManagementEnabled && r.InitialConfig.RemoteConfiguration.AgentID != "" {
	req.Header.Add("X-LabelManagementEnabled", "1")
	req.Header.Add("X-AgentID", r.InitialConfig.RemoteConfiguration.AgentID)
}
```
which adds the headers necessary for label management when (a) it is enabled in the config and (b) an `agent_id` was set.

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [X] Tests updated

### DIFF
--- a/docs/sources/static/configuration/agent-management.md
+++ b/docs/sources/static/configuration/agent-management.md
@@ -48,6 +48,12 @@ agent_management:
     # Set of self-identifying labels used for snippet selection.
     labels:
       [ <labelname>: <labelvalue> ... ]
+
+    # Whether to use labels from the label management service. If enabled, labels from the API supersede the ones configured in the agent.
+    label_management_enabled: <bool>
+
+    # A unique ID for the agent, which is used to identify the agent in the label management service.
+    agent_id: <string>
 ```
 
 ## API (v2)
--- a/docs/sources/static/configuration/agent-management.md
+++ b/docs/sources/static/configuration/agent-management.md
@@ -50,9 +50,9 @@ agent_management:
       [ <labelname>: <labelvalue> ... ]
 
     # Whether to use labels from the label management service. If enabled, labels from the API supersede the ones configured in the agent.
-    label_management_enabled: <bool>
+    label_management_enabled: <bool> | default = false
 
-    # A unique ID for the agent, which is used to identify the agent in the label management service.
+    # A unique ID for the agent, which is used to identify the agent.
     agent_id: <string>
 ```
 
--- a/pkg/config/agentmanagement.go
+++ b/pkg/config/agentmanagement.go
@@ -156,7 +156,7 @@ func (r remoteConfigHTTPProvider) FetchRemoteConfig() ([]byte, error) {
 	}
 
 	client := &http.Client{
-		Timeout: 5*time.Second,
+		Timeout: 5 * time.Second,
 	}
 	response, err := client.Do(req)
 	if err != nil {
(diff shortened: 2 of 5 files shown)


pkg/config/agentmanagement.go
```diff
@@ -114,47 +121,95 @@ func (r remoteConfigHTTPProvider) CacheRemoteConfig(remoteConfigBytes []byte) er
...
 	}
+	cleanedPassword := strings.TrimSpace(string(password))
+	return cleanedPassword, nil
+}
 
-	dir, err := os.Getwd()
+// createHTTPRequest creates an HTTP request based on the values in c. If label management is enabled,
+// the request will include headers to include the agent id and indicate that label management is enabled.
+func createHTTPRequest(c *AgentManagementConfig) (*http.Request, error) {
+	// Create the full url, possibly including labels as query params
```
jcreixell, line 137: I have some concerns about the code duplication and potential drift, WDYT about extending remote config and allowing to pass it headers as options (see POC [here](https://github.com/grafana/agent/commit/5f1136e1b4e7a3b3431d77b6608ef06007959bf1)). If we want to be more protocol agnostic, we could also pass it a `map[string]string` and map it to headers in the http provider.

pkg/config/agentmanagement.go
```diff
@@ -7,10 +7,13 @@ import (
 	"errors"
```
jcreixell, line 7: let's not forget to extend the docs with the new config fields

jcreixell (commented):
please see my comments :eye:

docs/sources/static/configuration/agent-management.md
```diff
@@ -48,6 +48,12 @@ agent_management:
     # Set of self-identifying labels used for snippet selection.
     labels:
       [ <labelname>: <labelvalue> ... ]
+
+    # Whether to use labels from the label management service. If enabled, labels from the API supersede the ones configured in the agent.
+    label_management_enabled: <bool>
```
jcreixell, line 53: let's document that it defaults to false?

docs/sources/static/configuration/agent-management.md
```diff
@@ -48,6 +48,12 @@ agent_management:
...
+
+    # A unique ID for the agent, which is used to identify the agent in the label management service.
```
jcreixell, line 55: we could be more generic here, as it is likely this will be used for more than just label management

pkg/config/agentmanagement.go
```diff
@@ -133,6 +135,13 @@ func (r remoteConfigHTTPProvider) FetchRemoteConfig() ([]byte, error) {
 		HTTPClientConfig: httpClientConfig,
 	}
 
+	if r.InitialConfig.RemoteConfiguration.LabelManagementEnabled && r.InitialConfig.RemoteConfiguration.AgentID != "" {
```
jcreixell, line 138: shall we add some validations to the config to make sure that:

- no labels are defined when `label_management_enabled: true`
- `agent_id` is defined when  `label_management_enabled: true`

?

jcreixell (approved):
overall LGTM, please see my last comments

clayton-cornell (commented):
Looks like the key points were covered in the review. No further suggestions.

PR #4286, opened 2023-06-28
clustering: allow setting the node's name with a CLI flag

tpaschalis:
#### PR Description
This PR allows users to configure the node name used for clustering with a command line flag. Since the empty node name defaults to the node's HOSTNAME, it made it difficult to test clustering on a single machine.

No issue filed.

#### Notes to the Reviewer
The list of clustering-based options has continued growing. This is the last new option I'm adding without some kind of structure to it. Once/if https://github.com/grafana/agent/issues/4253 is up, then we'll probably need to refactor the way we configure clustering either way.

#### PR Checklist

- [X] CHANGELOG updated
- [X] Documentation added
- [ ] Tests updated

### DIFF
--- a/cmd/internal/flowmode/cmd_run.go
+++ b/cmd/internal/flowmode/cmd_run.go
@@ -97,7 +97,7 @@ depending on the nature of the reload error.
 	cmd.Flags().
 		BoolVar(&r.clusterEnabled, "cluster.enabled", r.clusterEnabled, "Start in clustered mode")
 	cmd.Flags().
-		StringVar(&r.clusterAdvAddr, "cluster.node-name", r.clusterNodeName, "The name to use for this node")
+		StringVar(&r.clusterNodeName, "cluster.node-name", r.clusterNodeName, "The name to use for this node")
 	cmd.Flags().
 		StringVar(&r.clusterAdvAddr, "cluster.advertise-address", r.clusterAdvAddr, "Address to advertise to the cluster")
 	cmd.Flags().
--- a/cmd/internal/flowmode/cmd_run.go
+++ b/cmd/internal/flowmode/cmd_run.go
@@ -97,6 +97,8 @@ depending on the nature of the reload error.
 	cmd.Flags().
 		BoolVar(&r.clusterEnabled, "cluster.enabled", r.clusterEnabled, "Start in clustered mode")
 	cmd.Flags().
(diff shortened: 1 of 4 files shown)


cmd/internal/flowmode/cmd_run.go
```diff
@@ -93,6 +93,8 @@ depending on the nature of the reload error.
 		BoolVar(&r.enablePprof, "server.http.enable-pprof", r.enablePprof, "Enable /debug/pprof profiling endpoints.")
 	cmd.Flags().
 		BoolVar(&r.clusterEnabled, "cluster.enabled", r.clusterEnabled, "Start in clustered mode")
+	cmd.Flags().
```
mattdurham, line 96: Does it make sense to use this as the hostname? Feels like there could be drift between metric labels and cluster name

tpaschalis: If the node is unset, this node name defaults to the machine's HOSTNAME.

Is your suggestion that this goes the other way around? That if we set this CLI flag we also override the machine's HOSTNAME?

docs/sources/flow/reference/cli/run.md
```diff
@@ -35,6 +35,7 @@ The following flags are supported:
 * `--storage.path`: Base directory where components can store data (default `data-agent/`).
 * `--disable-reporting`: Disable [usage reporting][] of enabled [components][] to Grafana (default `false`).
 * `--cluster.enabled`: Start the Agent in clustered mode (default `false`).
+* `--cluster.node-name`: The name to use for this node (defaults to the environment's HOSTNAME).
```
rfratto, line 38: Is the all caps semantically meaningful here? You use HOSTNAME here but "hostname" down below. Since all caps usually indicate placeholders, maybe this should be "hostname" instead.

@clayton-cornell WDYT?

rfratto (approved):
LGTM from my side, but you should resolve that open thread with Matt first before merging :)

cmd/internal/flowmode/cmd_run.go
mattdurham: Discussed this and I now understand the use case so a bit of doc on doing this when running multiple nodes on one machine clears this up.

mattdurham (approved):
LGTM

docs/sources/flow/reference/cli/run.md
tpaschalis: Putting this in lowercase. It'd make more sense as `$HOSTNAME` but I don't really like that.

PR #4288, opened 2023-06-28
river/vm: unwrap nonsesitive OptionalSecret before performing binary ops

### LINKED ISSUE #4036: content should be one of [number string] for binop +, got capsule
### What's wrong?

Concat string through `+` operator wont work, with responses from a flow component like http.remote

### Steps to reproduce

1. Create a config file like described below.
2. Start agent

### System information

MacOS 13.4

### Software version

v0.33.0 + main

### Configuration

```text
// Fetch resourceId from IMDS
remote.http "imds_resource_name" {
    url            = "http://169.254.169.254/metadata/instance/compute/name?api-version=2021-02-01&format=text"
    headers = {
        "Metadata" = "true",
    }
    poll_frequency = "86400s"
    poll_timeout   = "1s"
}

remote.http "config_default" {
    url            = "https://exmaple.com/grafana-agent/config/default/"+remote.http.imds_resource_name.content+".river"
}
```

### Logs

```text
remote.http.imds_resource_name.content should be one of [number string] for binop +, got capsule
```

rfratto:
This change unwraps an OptionalSecret (when IsSecret is false) to a string for the purposes of binary operations. This allows OptionalSecret to be compared to strings and concatenated with strings, including with other OptionalSecret values.

This is a hack to fix the observable behavior, but the nature of the fix means that other issues may still occur in the future if there are ever more capsule types which support conversion to a non-capsule type. A longer-term fix is needed once we identify what type of conversions binary operations should do.

Fixes #4036.

### DIFF
--- a/pkg/river/vm/op_binary_test.go
+++ b/pkg/river/vm/op_binary_test.go
@@ -46,6 +46,21 @@ func TestVM_OptionalSecret_Conversion(t *testing.T) {
 			expect: bool(true),
 		},
 		{
+			name:   "capsule (secret) == capsule (secret)",
+			input:  `secret_val == secret_val`,
+			expect: bool(true),
+		},
+		{
+			name:   "capsule (non secret) == capsule (non secret)",
+			input:  `non_secret_val == non_secret_val`,
+			expect: bool(true),
+		},
+		{
+			name:   "capsule (non secret) == capsule (secret)",
+			input:  `non_secret_val == secret_val`,
+			expect: bool(false),
+		},
+		{
 			name:        "secret + string",
 			input:       `secret_val + string_val`,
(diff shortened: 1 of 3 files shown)


pkg/river/vm/op_binary_test.go
```diff
@@ -0,0 +1,79 @@
...
+		},
+		{
+			name:   "capsule + string",
+			input:  `non_secret_val + string_val`,
+			expect: string("worldhello"),
+		},
+		{
+			name:   "string == capsule",
+			input:  `"world" == non_secret_val`,
+			expect: bool(true),
```
mattdurham, line 41: How does this handle `secret_val == secret_val`?

mattdurham (approved):
LGTM

PR #4295, opened 2023-06-28
Add support to the CLI command `run` for converting a config from a different format

### LINKED ISSUE #3737: Wire up the Convert API to the `fmt` command for translating supported configs to river

### LINKED ISSUE #3738: Wire up the Convert API to the `run` command for translating supported configs to river on the fly

erikbaranowski:

Closes #3737
Closes #3738

#### Notes to the Reviewer

We should begin work on https://github.com/grafana/agent/issues/4073 right after this and complete it before release to take a wider view of docs for the converter feature.

Testing notes: I ran a prometheus config and confirmed it was scraping/forwarding metrics to grafana cloud using this. I also reloaded  with the API after updating the config.yaml and saw the update in the debug UI. I tested with an unsupported feature and confirmed the converter.bypass-warnings flag must be provided or the agent won't start up.

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,9 @@ Main (unreleased)
 
 - Flow UI: Add a view for listing the Agent's peers status when clustering is enabled. (@tpaschalis) 
 
-- Add support to the `grafana-agent fmt` CLI for converting a river file from supported formats to river. (@erikbaranowski)
+- Add a new CLI command `grafana-agent convert` for converting a river file from supported formats to river. (@erikbaranowski)
+
+- Add support to the `grafana-agent run` CLI for converting a river file from supported formats to river. (@erikbaranowski)
 
 ### Enhancements
 
--- a/cmd/internal/flowmode/cmd_run.go
+++ b/cmd/internal/flowmode/cmd_run.go
@@ -50,6 +50,7 @@ func runCommand() *cobra.Command {
 		uiPrefix:         "/",
 		disableReporting: false,
 		enablePprof:      true,
+		configFormat:     "flow",
 	}
 
 	cmd := &cobra.Command{
@@ -101,7 +102,7 @@ depending on the nature of the reload error.
 		StringVar(&r.clusterJoinAddr, "cluster.join-addresses", r.clusterJoinAddr, "Comma-separated list of addresses to join the cluster at")
 	cmd.Flags().
 		BoolVar(&r.disableReporting, "disable-reporting", r.disableReporting, "Disable reporting of enabled components to Grafana.")
-	cmd.Flags().StringVar(&r.configFormat, "config.format", r.configFormat, "The format of the source file. Supported formats: 'prometheus'.")
+	cmd.Flags().StringVar(&r.configFormat, "config.format", r.configFormat, "The format of the source file. Supported formats: 'flow', 'prometheus'.")
 	cmd.Flags().BoolVar(&r.configBypassConversionWarnings, "config.bypass-conversion-warnings", r.configBypassConversionWarnings, "Enable bypassing warnings when converting")
 	return cmd
 }
@@ -377,7 +378,7 @@ func loadFlowFile(filename string, converterSourceFormat string, converterBypass
 		return nil, err
 	}
 
-	if converterSourceFormat != "" {
+	if converterSourceFormat != "flow" {
 		var diags convert_diag.Diagnostics
 		bb, diags = converter.Convert(bb, converter.Input(converterSourceFormat))
 		hasError := hasErrorLevel(diags, convert_diag.SeverityLevelError)
--- a/cmd/internal/flowmode/cmd_run.go
+++ b/cmd/internal/flowmode/cmd_run.go
@@ -13,6 +13,8 @@ import (
 	"syscall"
 
 	"github.com/grafana/agent/component"
+	"github.com/grafana/agent/converter"
+	convert_diag "github.com/grafana/agent/converter/diag"
 	"github.com/grafana/agent/web/api"
 	"github.com/grafana/agent/web/ui"
 	"github.com/grafana/ckit/memconn"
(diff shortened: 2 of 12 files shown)


cmd/internal/flowmode/cmd_run.go
```diff
@@ -99,19 +101,23 @@ depending on the nature of the reload error.
 		StringVar(&r.clusterJoinAddr, "cluster.join-addresses", r.clusterJoinAddr, "Comma-separated list of addresses to join the cluster at")
 	cmd.Flags().
 		BoolVar(&r.disableReporting, "disable-reporting", r.disableReporting, "Disable reporting of enabled components to Grafana.")
+	cmd.Flags().StringVar(&r.converterSourceFormat, "converter.source-format", r.converterSourceFormat, "The format of the source file. Supported formats: 'prometheus'.")
+	cmd.Flags().BoolVar(&r.converterBypassWarnings, "converter.bypass-warnings", r.converterBypassWarnings, "Enable bypassing warnings when converting")
```
rfratto, line 105: What do you think about putting these flags in the `config` namespace to make it more clear what's happening, e.g., `--config.format` and `--config.bypass-conversion-warnings`? 

`converter` makes it sound like there's some converter process always running, when really this is just processing done on the config file. 

(When I was playing around with the `convert` subcommand, I found `--source-format` to be weirdly long; `--format` might make enough sense since the target format is always Flow)

docs/sources/flow/reference/cli/run.md
```diff
@@ -75,3 +75,17 @@ other agent nodes; if this is empty it will attempt to find a suitable address
 to advertise from a list of default network interfaces. The agent must be
 reachable over HTTP on this address as communication happens over the agent's
 HTTP server.
+
+## Converter mode
```
rfratto, line 79: "Mode" tends to be reserved for operational modes (static mode, flow mode), so I don't want to overload the term here. 

Can we just call it what it is? Something like "Configuration conversion?"

cmd/internal/flowmode/cmd_run.go
erikbaranowski: done

docs/sources/flow/reference/cli/run.md
erikbaranowski: done

rfratto (approved):
LGTM pending documentation review from Clayton. 

Thinking about this, though, I have a question: should we be calling this out as a beta feature? As it is right now, it's defaulting to stable, but I don't know if we're ready to call it that yet.

(It's also not experimental because I think we're set on supporting this use case)

erikbaranowski:
Good call on calling it beta for at least 1 release. I will addd that to the header here and in the convert command doc tomorrow for this PR

CHANGELOG.md
```diff
@@ -31,7 +31,9 @@ Main (unreleased)
 
 - Flow UI: Add a view for listing the Agent's peers status when clustering is enabled. (@tpaschalis) 
 
-- Add support to the `grafana-agent fmt` CLI for converting a river file from supported formats to river. (@erikbaranowski)
+- Add a new CLI command `grafana-agent convert` for converting a river file from supported formats to river. (@erikbaranowski)
+
+- Add support to the `grafana-agent run` CLI for converting a river file from supported formats to river. (@erikbaranowski)
```
thampiotr, line 36: duplicated line

cmd/internal/flowmode/cmd_run.go
```diff
@@ -99,19 +101,23 @@ depending on the nature of the reload error.
 		StringVar(&r.clusterJoinAddr, "cluster.join-addresses", r.clusterJoinAddr, "Comma-separated list of addresses to join the cluster at")
 	cmd.Flags().
 		BoolVar(&r.disableReporting, "disable-reporting", r.disableReporting, "Disable reporting of enabled components to Grafana.")
+	cmd.Flags().StringVar(&r.configFormat, "config.format", r.configFormat, "The format of the source file. Supported formats: 'prometheus'.")
```
thampiotr, line 104: Isn't default config format `river`? Should we make it the default and a supported option? I think this makes sense for a flag called `config.format`.

Alternatively, if we want this flag to mean "convert config from this format to river", then we could name the flag differently, e.g. `config.convert-format`".

CHANGELOG.md
erikbaranowski: This includes a wording fix for a previous PR on `grafana-agent convert` that was originally `grafana-agent fmt`.

docs/sources/flow/reference/cli/run.md
```diff
@@ -75,3 +77,17 @@ other agent nodes; if this is empty it will attempt to find a suitable address
...
+
+When the `--config.format` command-line argument is provided, Grafana
+Agent will convert the config file from the source format to River and
+immediately start running with it. This leverages the same converter API
+described in the [grafana-agent convert][] docs.
+
+If the `--config.bypass-conversion-warnings` command-line argument is also provided,
+Grafana Agent will ignore any warnings provided by the converter. This should
+be used with caution since the resulting conversion is not equivalent to the
+original config.
```
clayton-cornell, line 91:
```suggestion
If you also use the `--config.bypass-conversion-warnings` command-line argument,
Grafana Agent will ignore any warnings from the converter. Use this
argument with caution because the resulting conversion is not equivalent to the
original configuration.
```

clayton-cornell (approved):
Some small suggestions to shift the doc input to more active voice.

cmd/internal/flowmode/cmd_run.go
erikbaranowski: +1 to `config.convert-format`. @rfratto do we have consensus on this?

rfratto: `-config.format` defaulting to `river` (ie apply no conversion) is a good idea. 

I'm not sure about `config.convert-format`, to me it implies converting TO a format, rather than converting FROM a format. 

I don't want to block on bikeshedding though, final call is yours.

erikbaranowski: I can update the default and code it to handle `river` as existing behavior.

rfratto: > I can update the default and code it to handle river as existing behavior.

I don't think that's necessary, but the CLI could understand river as "no op", and not-river as "do a conversion"


BTW: I don't think it should be `river`, but `flow` instead; River is the language, Flow is the expected schema of a config file written in that language. It's the same reason one of the expected formats is Prometheus (the schema) and not YAML (the language)

erikbaranowski: makes sense. I have to add 1 small conditional to not call the converter if `flow` otherwise it will pass it along.

erikbaranowski: This change has been implemented. I'll merge this tomorrow pending any last minute feedback.

rfratto: LGTM 👍

thampiotr (approved):
LGTM, thanks!

PR #4296, opened 2023-06-28
river/token/builder: don't always skip zero valued optional fields

rfratto:
The logic to check to see if an optional field matches its default takes precedence over the old logic.

This fixes an unreleased issue where converting fields like send_exemplars would never be emitted, because `true` was the default (so it was omitted) and `false` is a zero value (so it was omitted).

The changelog is not updated since this issue is only observable in main.


mattdurham (approved):
LGTM

erikbaranowski (approved):
LGTM

PR #4299, opened 2023-06-29
Ensure __delta__ label is exporter in pyroscope.write

simonswine:
This is important to signal to the backend, that the profile is no longer using commulative values.


cyriltovena (approved):
LGTM

rfratto (approved):

PR #4300, opened 2023-06-29
Fix static mode operational dash for agent-logs namespace

thampiotr:
#### PR Description

When using dashboard mixin, on the "Agent Operational" dashboard, the `all` option for `pod` template expects that pods have `grafana-agent-.*` prefix. This is not always the case for every user. At the same time filtering pods in some way is useful, so this PR changes the regex to `.*agent.*`. 

Also gave some panels more descriptive names as I found some of them quite cryptic.

#### PR Checklist

- [x] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/example/docker-compose/grafana/dashboards/agent-operational.json
+++ b/example/docker-compose/grafana/dashboards/agent-operational.json
@@ -1093,7 +1093,7 @@
             "multi": true,
             "name": "namespace",
             "options": [ ],
-            "query": "label_values(agent_build_info, namespace)",
+            "query": "label_values(agent_build_info{cluster=~\"$cluster\"}, namespace)",
             "refresh": 1,
             "regex": "",
             "sort": 2,
@@ -1117,7 +1117,7 @@
             "multi": true,
             "name": "container",
             "options": [ ],
-            "query": "label_values(agent_build_info, container)",
+            "query": "label_values(agent_build_info{cluster=~\"$cluster\", namespace=\"$namespace\"}, container)",
             "refresh": 1,
             "regex": "",
             "sort": 2,
@@ -1128,7 +1128,7 @@
             "useTags": false
          },
          {
-            "allValue": ".*agent.*",
+            "allValue": ".+",
             "current": {
                "selected": true,
                "text": "All",
@@ -1141,7 +1141,7 @@
             "multi": true,
             "name": "pod",
             "options": [ ],
-            "query": "label_values(agent_build_info{container=~\"$container\", namespace=\"$namespace\"}, pod)",
+            "query": "label_values(agent_build_info{cluster=~\"$cluster\", namespace=\"$namespace\", container=\"$container\"}, pod)",
             "refresh": 1,
             "regex": "",
             "sort": 2,
--- a/example/docker-compose/grafana/dashboards/agent-operational.json
+++ b/example/docker-compose/grafana/dashboards/agent-operational.json
@@ -56,7 +56,7 @@
                "thresholds": [ ],
                "timeFrom": null,
                "timeShift": null,
-               "title": "GCs",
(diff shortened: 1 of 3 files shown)


example/docker-compose/grafana/dashboards/agent-operational.json
```diff
@@ -284,7 +284,7 @@
                "thresholds": [ ],
                "timeFrom": null,
                "timeShift": null,
-               "title": "CPU",
+               "title": "CPU Usage [time/s]",
```
ptodev, line 287: The only problem with including units in the title that I can think of, is that if someone edits an axis on the dashboard to have different units it might be confusing, or they'd have to update the title manually. 

Wouldn't it be better to just set the unit in the Grafana settings for each axis, and then Grafana will just display them on each axis?

tpaschalis (approved):

production/grafana-agent-mixin/debugging.libsonnet
```diff
@@ -6,20 +6,20 @@ local g = import 'grafana-builder/grafana.libsonnet';
     'agent-operational.json':
       utils.injectUtils(g.dashboard('Agent Operational'))
       .addMultiTemplate('cluster', 'agent_build_info', 'cluster')
-      .addMultiTemplate('namespace', 'agent_build_info', 'namespace')
-      .addMultiTemplate('container', 'agent_build_info', 'container')
-      .addMultiTemplateWithAll('pod', 'agent_build_info{container=~"$container", namespace="$namespace"}', 'pod', all='grafana-agent-.*')
+      .addMultiTemplate('namespace', 'agent_build_info{cluster=~"$cluster"}', 'namespace')
+      .addMultiTemplate('container', 'agent_build_info{cluster=~"$cluster", namespace="$namespace"}', 'container')
+      .addMultiTemplate('pod', 'agent_build_info{cluster=~"$cluster", namespace="$namespace", container="$container"}', 'pod')
```
thampiotr, line 11: I think making the templates depending on each other helps better in navigating the hierarchy of k8s resources.

example/docker-compose/grafana/dashboards/agent-operational.json
thampiotr: I think if someone changes the axis, they can also change the title. I didn't put unit everywhere as it's sometimes obvious (e.g. bytes), but I think it's helpful for some less obvious charts, like CPU Usage which is "how many seconds of CPU time was used every second".

example/docker-compose/grafana/dashboards/agent-operational.json
```diff
@@ -916,7 +916,7 @@
                "thresholds": [ ],
                "timeFrom": null,
                "timeShift": null,
-               "title": "Series/Config",
+               "title": "Series per Config",
```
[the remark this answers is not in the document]
thampiotr: It does look a bit worse when we have `Heap Used /Series/Pod` - looks like some kind of resource path or hierarchy. Since `per` fits on the screen, I'd prefer to use that.

example/docker-compose/grafana/dashboards/agent-operational.json
ptodev: Even in those cases I think I'd prefer to have the units on the actual axes instead of in the title. But I'm ok with having this in the title like this, it's still an improvement and I think it'll work ok.

example/docker-compose/grafana/dashboards/agent-operational.json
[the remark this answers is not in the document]
ptodev: That's true, let's try out the "per" notation - I'd be happy with it

ptodev (approved):

PR #4303, opened 2023-06-29
Make loki.LogsReceiver an interface

thampiotr:
#### PR Description

In order to implement config converter we need to override how a components' export is tokenized. 
* There is an example of [how this is done here](https://github.com/grafana/agent/blob/main/converter/internal/prometheusconvert/remote_write.go#L16) for `prometheus.remote_write` - when we create a block for remote_write, we return an `remotewrite.Exports`.
* That export is then set on river components args that use it and since we override the way its tokenized, we control how it is displayed in the final River configuration that is generated. In the example above, the exported receiver will be displayed as `"prometheus.remote_write.default.receiver"` which is what we want.
* However, the trick to override how an export value is tokenized requires us to [extend the exported type](https://github.com/grafana/agent/blob/main/converter/internal/common/convert_appendable.go#L14), which can only be done when it's an interface.
* `loki.LogsReceiver` is not an interface, so I ran into an issue when trying to implement `promtail` config conversion support. The problem can be[ seen in code in this draft PR](https://github.com/grafana/agent/pull/4160#discussion_r1245394586).

Proposed solution: make `loki.LogsReceiver` an interface.

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/component/common/loki/types.go
+++ b/component/common/loki/types.go
@@ -23,9 +23,29 @@ import (
 // to an outage or erroring (such as limits being hit).
 const finalEntryTimeout = 5 * time.Second
 
-// LogsReceiver is an alias for chan Entry which is used for component
+// LogsReceiver is an interface providing `chan Entry` which is used for component
 // communication.
-type LogsReceiver chan Entry
+type LogsReceiver interface {
+	Chan() chan Entry
+}
+
+type logsReceiver struct {
+	entries chan Entry
+}
+
+func (l *logsReceiver) Chan() chan Entry {
+	return l.entries
+}
+
+func NewLogsReceiver() LogsReceiver {
+	return NewLogsReceiverWithChannel(make(chan Entry))
+}
+
+func NewLogsReceiverWithChannel(c chan Entry) LogsReceiver {
+	return &logsReceiver{
+		entries: c,
+	}
+}
 
 // Entry is a log entry with labels.
 type Entry struct {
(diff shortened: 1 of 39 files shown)


component/common/loki/types.go
```diff
@@ -23,9 +23,29 @@ import (
...
+
+func NewLogsReceiver() LogsReceiver {
+	return NewLogsReceiverWithChannel(make(chan Entry))
+}
+
+func NewLogsReceiverWithChannel(c chan Entry) LogsReceiver {
+	return &logsReceiver{
+		entries: c,
+	}
+}
```
thampiotr, line 48: This is the main change - the rest is just making it work using it.

mattdurham:
Might be best for a separate PR, but I am wondering if its possible to hide the underlying implementation and the interface be more generic while we are looking at it. 

```
type LogsReceiver interface {
	Send(Entry)
        Receive() Entry
}
```

mattdurham (approved):
LGTM

erikbaranowski (approved):
LGTM

thampiotr:
> Might be best for a separate PR, but I am wondering if its possible to hide the underlying implementation and the interface be more generic while we are looking at it.
> 
> ```
> type LogsReceiver interface {
> 	Send(Entry)
>         Receive() Entry
> }
> ```

There's also some places where `close(channel)` is called, so not 100% sure we can hide that it's a channel and hence it may not be useful to do that.

PR #4307, opened 2023-06-29
static to flow for prometheus config

erikbaranowski:
#### PR Description

This PR implements the prometheus portion of static to flow using a very minimal implementation in the `staticconvert` package by reusing the `prometheusconvert` package

Related to #4291 

#### Notes to the Reviewer

Another PR will be needed to create warnings for unsupported static to flow features before this gets wired up to the command line.

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [x] Tests updated


thampiotr (approved):
Nothing to add from my side!

converter/internal/prometheusconvert/prometheusconvert.go
```diff
@@ -52,50 +52,41 @@ func Convert(in []byte) ([]byte, diag.Diagnostics) {
...
-		prometheusScrapeBlocks:      []*builder.Block{},
-		prometheusRelabelBlocks:     []*builder.Block{},
-		prometheusRemoteWriteBlocks: []*builder.Block{},
-	}
-}
-
 // AppendAll analyzes the entire prometheus config in memory and transforms it
 // into Flow Arguments. It then appends each argument to the file builder.
 // Exports from other components are correctly referenced to build the Flow
 // pipeline.
```
mattdurham, line 58: Add some documentation around labelPrefix, its not clear what that does from the functional call? Seems like it was jobname and is now label prefix?

converter/internal/prometheusconvert/prometheusconvert.go
```diff
@@ -52,50 +52,41 @@ func Convert(in []byte) ([]byte, diag.Diagnostics) {
...
+		remoteWriteLabel = "default"
+	}
+	remoteWriteExports := appendPrometheusRemoteWrite(pb, promConfig.GlobalConfig, promConfig.RemoteWriteConfigs, remoteWriteLabel)
 	remoteWriteForwardTo := []storage.Appendable{remoteWriteExports.Receiver}
+
 	scrapeForwardTo := remoteWriteForwardTo
 	for _, scrapeConfig := range promConfig.ScrapeConfigs {
-		promMetricsRelabelExports := appendPrometheusRelabel(pb, scrapeConfig.MetricRelabelConfigs, remoteWriteForwardTo, scrapeConfig.JobName)
+		label := scrapeConfig.JobName
+		if labelPrefix != "" {
```
mattdurham, line 73: Should this be remoteWriteLabel, or does default not apply here? Or to prevent confusion should the check for `remoteWriteLabel == ""` be embeded into the `appendPrometheusRetmoreWrite` func? I dont see it being used aside from the call to that func.

converter/internal/prometheusconvert/prometheusconvert_test.go
```diff
@@ -44,6 +44,11 @@ func TestConvert(t *testing.T) {
 					}
 				}
 
+				// If we expect more errors than we got
+				if len(expectedErrors) > len(diags) {
```
mattdurham, line 48: Should this be a `!=` ? Or is diags being less the only valid use case? Would complicate your fails a bit though.

mattdurham (changes requested):
Looking solid, some clarifications and doc on how things are used would be nice.

converter/internal/prometheusconvert/prometheusconvert_test.go
erikbaranowski: the `<` (opposite) case is handled above on line 43

converter/internal/prometheusconvert/prometheusconvert.go
erikbaranowski: I added some details about this

mattdurham (approved):
LGTM

PR #4310, opened 2023-06-30
Spartan0x117/flow exporter module metrics path label fix

spartan0x117:
#### PR Description
I noticed that when defining exporters inside of modules, trying to scrape them resulted in a `404 Not Found` error. After digging a bit, I saw that the exported target's addresses were the same as though the component were not defined inside a module (e.g. `/api/v0/component/prometheus.exporter.redis.example_exporter`) instead of having the module be a part of the path (e.g. `/api/v0/component/module.file.example_module/prometheus.exporter.redis.local_redis`). Debugging a bit further led me to where the value was initially set, and I saw that both `HTTPPath` and `DataPath` were using the controller-local ID rather than the global ID.

I change `DataPath` as well, as I'd assume otherwise two of the same components defined in different modules might conflict where they write to.

Let me know if this is the wrong approach and should be handled a different way :smile: 

#### PR Checklist

- [X] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,6 +86,8 @@ Main (unreleased)
 
 - Fix Grafana Agent mixin's "Agent Operational" dashboard expecting pods to always have `grafana-agent-.*` prefix. (@thampiotr)
 
+- Fix the HTTP Path and Data Path from the controller-local ID to the global ID for components loaded from within a module loader. (@spartan0x117)
+
 v0.34.3 (2023-06-27)
 --------------------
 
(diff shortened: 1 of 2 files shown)


mattdurham (changes requested):
Great catch, can we add a quick test to it?
New file component_test.go

```
package controller

import (
	"github.com/grafana/agent/component"
	"testing"
)

func TestGlobalID(t *testing.T) {
	mo := getManagedOptions(ComponentGlobals{
		DataPath:       "/data/",
		HTTPPathPrefix: "/http/",
		ControllerID:   "module.file",
		NewModuleController: func(id string) component.ModuleController {
			return nil
		},
	}, &ComponentNode{
		nodeID: "local.id",
	})
	require.True(t, mo.HTTPPath == "/http/module.file/local.id/")
	require.True(t, mo.DataPath == "/data/module.file/local.id")
}
```

mattdurham:
I tried to use codespaces to add the file directly but something went sideways.

mattdurham (approved):

PR #4312, opened 2023-06-30
Added write_relabel config for prometheus.remote_write

### LINKED ISSUE #4311: Add write_relabel_configs config to prometheus.remote_write
### Request

The Prometheus configuration allows to define relabel configs on the remote_write configuration.

https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write

I would like to see the configuration on grafana, too:

https://grafana.com/docs/agent/latest/flow/reference/components/prometheus.remote_write/	

### Use case

Instead configure relabel actions on each scrape component, I would like to apply to relabel configuration on a more central approach.

jkroepke:
#### PR Description

Added support relabel_config on remote_write configuration.

Fixes #4311 

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated


mattdurham:
What is the use case for using this versus a relabel component?

jkroepke:
1 Use case: 

I'm fetch IMDS data and put it as external_labels on the prometheus.remote_write component. Then I so some modification with relabel_config, to sanitize them. external_labels are not covered by relabel component (tbh, I did not test this, but looking at the flow, the external_labels are attached after relabel component)

clayton-cornell (commented):
No changes for docs. Looks like everything is covered

mattdurham:
Do you have an example config of what your wanting to do? That might help me visualize it better

jkroepke:
For example

```
remote.http "imds_resource_id" {
    url            = "http://169.254.169.254/metadata/instance/compute/resourceId?api-version=2021-02-01&format=text"
    headers = {
        "Metadata" = "true",
    }
    poll_frequency = "86400s"
    poll_timeout   = "1s"
}

prometheus.remote_write "prometheus" {
    endpoint {
        url = "http://prometheus:8080/api/v1/write"

        basic_auth {
            username = "foo"
            password = "bar"
        }

        queue_config {
            retry_on_http_429 = true
        }

        write_relabel_config {
            source_labels = ["resourceID"]
            target_label = "resourceID"
            action = "lowercase"
        }
    }

    external_labels = {
        "resourceID" = remote.http.imds_resource_id.content,
    }
}

```

mattdurham:
I am a bit hesitant on this because it feels like something that could be accomplished with a relabel component, which granted is a bit uglier. IMO `prometheus.remote_write` is my least favorite component because it already unites the wal and metric http client into one thing. That being said it already has this functionality so it makes porting over a lot easier I imagine. @erikbaranowski would this make your life a ton easier in porting over behavior?

jkroepke:
It can be accomplished with the relabel component, but then I have resourceID to all relabel components.

Another use case is to whitelist metrics to an remote write endpoint.

For example, you have 2 remote endpoints and you want to send different sets of metrics to different upstreams.

The first remote endpoint is the Grafana cloud which has a quota limit of metric count and a second upstream would be a local prometheus instance where you can sent everything.

I mean, if prometheus supports this, why not expose this functionality in Grafana Agent, too?

mattdurham:
Sounds reasonable then. Regarding exposing functionality in the Agent from Prometheus, the approach I would like to take is limiting the scope of a component. I would much rather have a decoupled wal from the remote write and put in other components to fulfill the asks.

My personal opinion on a better world

```
prometheus.scrape "scrape" {
  forward_to = [prometheus.wal.memory.mem1.receiver]
}

prometheus.wal.memory "mem1" {
   forward_to = [prometheus.relabel.relabeling.receiver]
}

prometheus.relabel "relabeling" {
  rule {}
  forward_to = [prometheus.remote.endpoint.receiver]
}

promtheus.remote "endpoint" {
  endpoint {}
}
```

PR #4316, opened 2023-06-30
Re-enable mongodb exporter

jcreixell:
#### PR Description

  - The exporter had been disabled in #4165 due to licensing issues.
  - The licensing issues have been resolved. See https://github.com/percona/mongodb_exporter/commit/6eedce7c9f79932592dad0496b49b2028992ff75
  - Special thanks to @BupycHuk and the percona team.

#### PR Checklist

- [x] CHANGELOG updated


go.mod
```diff
@@ -122,6 +122,7 @@ require (
 	github.com/opentracing/opentracing-go v1.2.0
 	github.com/ory/dockertest/v3 v3.8.1
 	github.com/oschwald/geoip2-golang v1.8.0
+	github.com/percona/mongodb_exporter v0.31.2
```
rfratto, line 125: This is still a version of mongodb_exporter that has the licensing issues, while the version without it hasn't been released yet. 

This should be updated to SHA 6eedce7c9f79932592dad0496b49b2028992ff75 so we're importing the right thing

jcreixell: you are totally right, fixing

jcreixell:
this needs a bit more work, since we are upgrading from v0.31.2 to v0.39+. Seems like there are some compilation errors at the moment related to prometheus. I will also have a look at the changelog and see if there is anything we need to update in the integration itself or docs.

marctc:
MongoDB exporter uses a custom fork of `client_golang`. I opened a [PR upstream](https://github.com/percona/mongodb_exporter/pull/682) removing that fork.

marctc (approved):

PR #4318, opened 2023-06-30
Fix kubelet discovery for deleted pods

gcampbell12:
#### PR Description

I found a bug which causes pods to continue to be scraped when they have been deleted (and the kubelet no longer returns the pod in its response).

The discovery component requires a target group to be in the list of discovered targets with an empty target list when a group should be removed. Therefore we need to cache the previous discovery results from the kubelet and compare them in the next refresh cycle in order to create the 'empty' target groups.

https://github.com/grafana/agent/blob/73cf6f5b188ce7ddbb1683101f63039611545868/component/discovery/discovery.go#L215-L224

#### Notes to the Reviewer

Unsure if we want a changelog for this since it's unreleased still?

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated


mattdurham:
We actually dont need the changelog for the reason you mentioned =)

gcampbell12:
@mattdurham anything I can do to get this over the line? Would be great to have kubelet discovery fully functional in the next release!

mattdurham:
Apologies, merging now

gcampbell12:
Thanks @mattdurham!

mattdurham (approved):
LGTM

PR #4320, opened 2023-06-30
Introduce service interface, port HTTP service

rfratto:
This is the first of many PRs to implement #4253. 

This first PR does three things:

* Introduce the service package API as defined in #4253. 
* Implement the service.Host interface in pkg/flow (even though it's a no-op for now). 
* Port the HTTP service to `service/http`, implementing the appropriate interface. 

Note that the resulting code is a bit hackier now than it will be in the future once services are fully implemented.


rfratto:
I verified that the following still work:

* Scraping exporter components 
* The UI, including with a basic config and a config using modules

service/http/http.go
```diff
@@ -0,0 +1,275 @@
+// Package http implements the HTTP service for Flow.
+package http
```
captncraig, line 2: Is adding another top level directory the right call here? Would it make more sense to go under pkg/flow? Do we have guidelines on what makes something important enough for a tld?

captncraig (commented):
I'd like to see an example here of an existing component being converted to use this. That may help to do a better analysis of whether these are the right level of apis to expose.

service/http/http.go
rfratto: I'm gradually moving away from the pkg/ folder, most Flow things are top-level now (like component). 

I can move if it you want, though.

rfratto:
> I'd like to see an example here of an existing component being converted to use this. That may help to do a better analysis of whether these are the right level of apis to expose.

Because of how everything is interconnected, it's going to take a while before you can see the results of this. I'd prefer to optimize for smaller, more reviewable PRs right now, and then we can revert everything later if the API doesn't work out.

service/http/http.go
captncraig: I'm ok with that.

PR #4321, opened 2023-06-30
component: include running module IDs in Info

rfratto:
This allows clients to query for running modules. 

This introduces an internal API for module controllers for lower-level information, leaving the API exposed to components unchanged.

This is required as part of enabling the component.Provider API to be able to navigate through modules and have a global view of all running components.

### DIFF
--- a/component/component_provider.go
+++ b/component/component_provider.go
@@ -54,6 +54,12 @@ type Info struct {
 	// evaluated yet.
 	Component Component
 
+	// ModuleIDs includes the list of current module IDs that the component is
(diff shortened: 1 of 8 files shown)


mattdurham (approved):
Lgtm, I keep thinking at some point we should alias some of these Id string representations

rfratto:
> Lgtm, I keep thinking at some point we should alias some of these Id string representations

Yeah, I agree, though it might have to wait until we combine the component/flow/service packages so we don't have cyclic dependency issues.

We always could have a package just for IDs, but that feels a bit gross I think.

PR #4325, opened 2023-06-30
[draft] flow: centrally track modules

rfratto:
This adds a central registration list of running modules, enabling an API to look up running components within a module.

The safety of this implementation depends on the behavior that modules are leak free and that they correctly unregister themselves.


rfratto:
Closing. The approach here is a bit wonky because of controllers having pointers to themselves. It would instead be better to traverse down the module tree to get the module you want, which can be done just by updating component controller and module.

PR #4330, opened 2023-07-03
feat(configReloader): set securitycontext

yanehi:
#### PR Description

Users can now set the IDs of the configReloader container in addition to the grafana-agent.

I would like to run the configReloader container with non-root IDs in my cluster. Unfortunately I currently have no way to set these via the values. With my pullrequest it should be possible to set the IDs for the container optionally.

#### PR Checklist

- [x] CHANGELOG updated
- [ ] Documentation added --> Not relevant because it is only a small change?
- [ ] Tests updated --> Not relevant because it is only a small change?

### DIFF
--- a/operations/helm/charts/grafana-agent/CHANGELOG.md
+++ b/operations/helm/charts/grafana-agent/CHANGELOG.md
@@ -23,6 +23,8 @@ Unreleased
 
 - Update Grafana Agent version to v0.34.2. (@captncraig)
 
+- Set securityContext for configReloader container. (@yanehi)
+
 ### Other changes
 
 - Make the agent and config-reloader container resources required when using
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -81,6 +81,8 @@ Main (unreleased)
 
 - Allow setting the node name for clustering with a command-line flag. (@tpaschalis)
 
+- Set securityContext for configReloader container. (@yanehi)
```
rfratto, line 84: The Helm chart has its own CHANGELOG.md in the operations/helm directory, can you move this there instead?

rfratto (approved):
LGTM after the CHANGELOG.md entry is moved, thank you!

CHANGELOG.md
yanehi: Done!

PR #4336, opened 2023-07-04
docs: enhance tooling documentation

theSuess:
Moved the tooling section to where developers would expect it. Also added links to the emacs-mode and tree-sitter grammar


clayton-cornell (approved):
This makes sense to me.

clayton-cornell:
@theSuess Question before we merge this change... in the original text, it was clear that the tools were community developed. Now it makes it look more like it's "official".... yet they exist in a private github repo. Do we want to either make it clear these are still community tools and not Grafana, or should they be moved out of the private repos?

theSuess:
Good point, I've added back the `Experimental` declaration for now and will try to get these extensions under the Grafana umbrella

clayton-cornell:
Looks good to me. Time to merge. :-)

PR #4338, opened 2023-07-04
tools: add Makefile helpers for agent-flow-mixin

tpaschalis:
#### PR Description
As part of writing documentation for clustering, I'd like to simplify the way users interact with our agent-flow-mixin.

This is a first approach in doing so; allowing people to compile our mixin and have its output ready to be imported into their Grafana + Prometheus instances.

No issue filed

#### PR Checklist

- [ ] CHANGELOG updated (N/A)
- [ ] Tests updated (N/A)

### DIFF
--- a/tools/make/mixin.mk
+++ b/tools/make/mixin.mk
@@ -1,0 +1,32 @@
+# Makefile.mixin adds mixin-specific targets.
+
+PARENT_MAKEFILE := $(firstword $(MAKEFILE_LIST))
+
+MIXIN_PATH       ?= operations/agent-flow-mixin
+MIXIN_OUT_PATH   ?= operations/agent-flow-mixin-compiled
+
+format-mixin: ## Format the mixin files.
+	@find $(MIXIN_PATH) -type f -name '*.libsonnet' | xargs jsonnetfmt -i
+
+check-mixin: ## Build, format and check the mixin files.
+check-mixin: build-mixin format-mixin check-mixin-jb check-mixin-mixtool
+	@cd $(MIXIN_PATH) && \
+	jb install && \
+	mixtool lint mixin.libsonnet
+
+check-mixin-jb:
+	@cd $(MIXIN_PATH) && \
+	jb install
+
+check-mixin-mixtool: check-mixin-jb
+	@cd $(MIXIN_PATH) && \
+	mixtool lint mixin.libsonnet
+
+build-mixin: ## Generates the agent-flow-mixin zip file.
+build-mixin: check-mixin-jb
+		mkdir -p "$(MIXIN_OUT_PATH)"; \
+		mixtool generate all --output-alerts "$(MIXIN_OUT_PATH)/alerts.yaml" --directory "$(MIXIN_OUT_PATH)/dashboards" "${MIXIN_PATH}/mixin.libsonnet"; \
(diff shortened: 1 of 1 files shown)


operations/agent-flow-mixin/jsonnetfile.lock.json
```diff
@@ -0,0 +1,5 @@
+{
```
tpaschalis, line 1: Gets created when calling `jb install` but we never checked it in. As far as I can tell, other projects do check it in their repos, so I thought I'd do the same.

tools/make/mixin.mk
```diff
@@ -0,0 +1,32 @@
...
+MIXIN_PATH       ?= operations/agent-flow-mixin
+MIXIN_OUT_PATH   ?= operations/agent-flow-mixin-compiled
+
+format-mixin: ## Format the mixin files.
+	@find $(MIXIN_PATH) -type f -name '*.libsonnet' | xargs jsonnetfmt -i
+
+check-mixin: ## Build, format and check the mixin files.
+check-mixin: build-mixin format-mixin check-mixin-jb check-mixin-mixtool
+	@cd $(MIXIN_PATH) && \
+	jb install && \
```
thampiotr, line 14: I think you get `jb install` already from `check-mixin-jb` - if I understand this right.

tools/make/mixin.mk
```diff
@@ -0,0 +1,32 @@
...
+	mixtool lint mixin.libsonnet
+
+check-mixin-jb:
```
thampiotr, line 17: Nit: I'd call it `mixin-jb-install` or sth like this,

tools/make/mixin.mk
```diff
@@ -0,0 +1,32 @@
...
+	@cd $(MIXIN_PATH) && \
+	jb install
+
+check-mixin-mixtool: check-mixin-jb
```
thampiotr, line 21: Nit: This could be called `lint-mixin` 🤔

thampiotr (approved):
LGTM, thanks! 
Only some small comments, but not blocking.

tools/make/mixin.mk
tpaschalis: You're right, thanks!

tools/make/mixin.mk
```diff
@@ -0,0 +1,32 @@
+# Makefile.mixin adds mixin-specific targets.
+
+PARENT_MAKEFILE := $(firstword $(MAKEFILE_LIST))
+
+MIXIN_PATH       ?= operations/agent-flow-mixin
+MIXIN_OUT_PATH   ?= operations/agent-flow-mixin-compiled
```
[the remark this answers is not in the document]
tpaschalis: My initial use case was to mention the `make` command in the clustering docs PR so that people could easily reuse our dashboards and alerts. TBH I hadn't thought about including this as a release asset.

So you think this should be invoked among the other commands in `make dist`? I think it makes sense; and once we've fixed linting errors we can also make it part of our Drone setup, much like other build artifacts.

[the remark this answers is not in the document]
rfratto: > So you think this should be invoked among the other commands in make dist? I think it makes sense; and once we've fixed linting errors we can also make it part of our Drone setup, much like other build artifacts.

Hm, maybe? At first I thought that was what Mimir was doing, but I see they have the same -compiled folder as you do here. 

I'll leave it up to you, having it as a release asset potentially makes the workflow of installing it easier.

PR #4343, opened 2023-07-05
Add docs for updating Otel dependencies

ptodev:
Putting in writing the lessons learned from #3858. Hopefully this helps us to do upgrades more often and with less effort.

### DIFF
--- a/docs/developer/updating-otel.md
+++ b/docs/developer/updating-otel.md
@@ -1,0 +1,250 @@
+# Updating OpenTelemetry Collector dependencies
+
+The Agent depends on various OpenTelemetry (Otel) modules such as these:
+```
+github.com/open-telemetry/opentelemetry-collector-contrib/exporter/jaegerexporter v0.80.0
+github.com/open-telemetry/opentelemetry-collector-contrib/extension/sigv4authextension v0.80.0
+github.com/open-telemetry/opentelemetry-collector-contrib/processor/spanmetricsprocessor v0.80.0
+go.opentelemetry.io/collector v0.80.0
+go.opentelemetry.io/collector/component v0.80.0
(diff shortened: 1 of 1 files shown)


rfratto (approved):

rfratto (approved):
Rubber stamp; I don't personally have time to dive deep into this, but I don't want you to be blocked on getting docs merged. We can improve them over time.

ptodev:
> Rubber stamp; I don't personally have time to dive deep into this, but I don't want you to be blocked on getting docs merged. We can improve them over time.

@rfratto thank you, but based on [your comment](https://github.com/grafana/agent/pull/4948#pullrequestreview-1604960855) on another PR, I think the section on labeling the stability will have to be changed:

> 4. Update the Agent's code and documentation where needed.
     * Pay particular attention to whether the component's stability has changed. Stability is labeled at the top of the Collector documentation. The stability label in Collector's docs should match the one in the Agent's docs.
    
When someone updates the components, they need to know whether the stability label in the Agent docs is accurate. I'm not sure what is the best way to do this right now? E.g. some components which are beta in Collector might be stable in the Agent, etc.

rfratto:
The stability level can always be increased but should never be decreased. We should be extra cautious when introducing new components. If we introduce it as stable, it's stable.

ptodev:
@rfratto Thank you - I updated the instructions to say that stability lebels should never be decreased. Now they also say that if a stability in Otel has increased, the developer should check with the rest of the team on whether the Agent component's stability should also be increased.

@wildum Would you mind reviewing the documentation please? I updated it just now to reflect some of the points we discussed on the PR.

wildum (approved):

PR #4344, opened 2023-07-05
docs: introduce docs for clustering

tpaschalis:
#### PR Description
This PR introduces the first proper documentation around clustering. It contains general information around the concept, instructions on how to run with using our Helm chart, how to install the mixin, discusses some of the failure modes, and explains the first clustering use case to auto-distribute targets.

It also fixes some of the previous documentation and marks the feature as **Beta**.

No issue filed.

#### PR Checklist

- [ ] CHANGELOG updated (N/A)
- [X] Documentation added
- [ ] Tests updated (N/A)


docs/sources/flow/concepts/clustering.md
```diff
@@ -0,0 +1,306 @@
...
+The behavior of a standalone, non-clustered agent is the same as if it was a
+1-node cluster.
+
+## Usage
+
+The following set of command-line flags are used to configure clustering when
+starting up Grafana Agent:
+- `--server.http.listen-addr`: the address where the agent’s HTTP server
+  listens to.
+- `--cluster.enabled`: enables cluster awareness.
```
mattdurham, line 36: Why do I need this if `The behavior of a standalone, non-clustered agent is the same as if it was a
1-node cluster.` is true?

docs/sources/flow/concepts/clustering.md
```diff
@@ -0,0 +1,306 @@
...
+- `--cluster.node-name`: defines the name used by the cluster node. If not
+  provided, it defaults to the machine's hostname.
+- `--cluster.advertise-address`: defines the address the agent advertises for
```
mattdurham, line 39: Add a comment on where it is inferred from, I am assuming from `--server.http.listen-addr`

docs/sources/flow/concepts/clustering.md
```diff
@@ -0,0 +1,306 @@
...
+  its peers to connect to. If not provided, it is inferred automatically.
+- `--cluster.join-addresses`: accepts a comma-separated list of addresses to
+  join the cluster at. These can be IP addresses with an optional port, or a
+DNS record to lookup.
+
+Cluster communication happens over HTTP/2 on the agents’ HTTP listener. The
+agents must be configured to accept connections from one another on the address
+defined by the `--server.http.listen-addr` flag.
```
mattdurham, line 45: Is using `--server.http.listen-addr` accurate if I am using `--cluster.advertise-address`?

docs/sources/flow/concepts/clustering.md
```diff
@@ -0,0 +1,306 @@
...
+
+Each cluster member’s name must be unique within the cluster. Nodes which try
+to join with a conflicting name are rejected.
```
mattdurham, line 48: What does rejected mean? Not added to the cluster or show up in the cluster diagram but in a bad state?

docs/sources/flow/concepts/clustering.md
```diff
@@ -0,0 +1,306 @@
...
+join or can try to connect to itself.
+
+## Deploy clustering using the Helm chart
+
+The easiest way to deploy agent clustering is by making use of our
+[Helm chart][].
+
+The following `values.yaml` file deploys an agent StatefulSet for metrics
+collection. It makes use of a [headless service][] to retrieve the IPs of the
+agent pods for the `--cluster.join-addresses` argument, as well as an HPA for
```
mattdurham, line 68: Link or expand HPA? A

docs/sources/flow/concepts/clustering.md
```diff
@@ -0,0 +1,306 @@
...
+# Clustering (beta)
+
+Agent Clustering enables Grafana Agent Flow to coordinate a fleet of agents to
+work together for workload distribution and high availability. It makes agent
+deployments horizontally scalable by default, and minimizes the operational
+overhead of managing your agent infrastructure to handle demand.
+
+To achieve this, Grafana Agent makes use of an eventually consistent model and
+the lightweight [grafana/ckit][] framework. All running nodes are assumed to
+use the same configuration and have access to the same type of hardware and
```
mattdurham, line 17: Maybe add somewhere later a bit about convergence around a config when rolling our new configurations, there is some of this in the troubleshooting but from reading this line to me it says I cant do a rolling release but have to do a full up and down when changing configuration.

docs/sources/flow/concepts/clustering.md
tpaschalis: They're never added to the cluster. Nodes that detect conflicted names (eg. during the initial full synchronization when joining the cluster) will bail early and refuse to gossip messages about the new node joining.

docs/sources/flow/concepts/clustering.md
tpaschalis: No, because `--server.http.listen-addr` is likely to be something generic like `0.0.0.0:12345`.

By default, `ckit` [looks into](https://github.com/grafana/ckit/blob/30fb40ef3315f650acdebcba10c338ca4f062693/advertise/addr.go#L14-L18
) the `eth0`, `en0` local network interfaces to find a suitable address

I tried to explain this below, but I can add some more information to make this clearer.

docs/sources/flow/concepts/clustering.md
tpaschalis: Nope, you're right, thanks!

docs/sources/flow/concepts/clustering.md
```diff
@@ -0,0 +1,303 @@
...
+network resources such as PVCs and service discovery APIs.
+
+Clustering is built from the ground-up to be the de-facto answer to "how do I
+scale my Grafana Agent setup" in the future. As an example, in comparison to a
+setup using [hashmod sharding][], clustering with target auto-distribution
+scales and provides resiliency at _half_ the cost in resources and provides
+dynamic resharding without the need to change configuration files.
+
+The behavior of a standalone, non-clustered agent is the same as if it was a
+1-node cluster.
```
rfratto, line 27: [Spell out numerals less than 10](https://developers.google.com/style/numbers):

```suggestion
one-node cluster.
```

rfratto: Alternatively, you could say "single-node cluster" here too.

docs/sources/flow/concepts/clustering.md
```diff
@@ -0,0 +1,303 @@
...
+
+## Usage
+
+The following set of command-line flags are used to configure clustering:
```
rfratto, line 31: This feels like detailed reference documentation to me, plus we already document these flags in the cli page. Should we link to that page here?

docs/sources/flow/concepts/clustering.md
```diff
@@ -0,0 +1,303 @@
...
+
+Peers communicate over HTTP/2 on the agent's built-in HTTP server. Each node
+must be configured to accept connections from one another on the address
+defined or inferred in `--cluster.advertise-address`.
+
+Each cluster member’s name must be unique within the cluster. Nodes which try
+to join with a conflicting name are rejected.
+
+If the advertised address is not explicitly set, the agent tries to infer a
+suitable one from the `eth0` and `en0` local network interfaces.
```
rfratto, line 51: This is missing information on:

* What does the inference do? 
* What happens when the inference fails?

docs/sources/flow/concepts/clustering.md
```diff
@@ -0,0 +1,303 @@
+---
```
rfratto, line 1: I'm a little worried the concept page is doing too much. 

From the Writers' Toolkit on [Concept topics](https://grafana.com/docs/writers-toolkit/writing-guide/topic-types/concept/#concept-topic):

> The following types of content can be included in concepts:
>
> * Detailed overviews of Grafana’s features with benefits and clearly defined terms
> * Diagrams that help users understand the components of a system
> * Process flow diagrams
> * Best practice guidelines
> * An example of how a feature is used. Examples might include screenshots or other supporting visuals
> 
> A concept topic does not include:
>
> * Step-by-step instructions
> * Reference information, such as lookup tables or lists of values

I would suggest splitting this doc up, where this focuses more on the benefits of clustering, and maybe some architectural diagrams, with best practices. 

I would move everything else to either the appropriate reference topic (i.e., documenting command line flags in cli/run.md) or troubleshooting information to the [Debugging](https://grafana.com/docs/agent/latest/flow/monitoring/debugging/) page.

@clayton-cornell Am I off-base here?

docs/sources/flow/concepts/clustering.md
mattdurham: I personally like the notion that agent is always clustering even if one, avoids the multiple apps in one that we got with static mode.

docs/sources/flow/concepts/clustering.md
tpaschalis: I removed the documentation from the `run` command page when I tried to gather everything here.

Since you think it reads too heavy, let's separate the info back and see how it looks like!

docs/sources/flow/concepts/clustering.md
```diff
@@ -0,0 +1,303 @@
...
+
+Clustering enables Grafana Agent Flow to coordinate a fleet of agents to
+work together for workload distribution and high availability. It makes agent
+deployments horizontally scalable by default, and minimizes the operational
+overhead of managing your agent infrastructure to handle demand.
+
+To achieve this, Grafana Agent makes use of an eventually consistent model and
+the lightweight [grafana/ckit][] framework. All running nodes are assumed to
+use the same configuration and have access to the same type of hardware and
+network resources such as PVCs and service discovery APIs.
```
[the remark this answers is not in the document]
tpaschalis: Agreed, I'll try to make it a bit more generic. How does this sound:
```
To achieve this, Grafana Agent makes use of an eventually consistent model that
assumes all participating Agents are interchangeable and converge on using the
same configuration file.
```

docs/sources/flow/concepts/clustering.md
tpaschalis: Ok, I've went through a second iteration of this

a) The command-line flags with their explanation have been moved back to the `run` command docs
b) The debugging information is more high-level now and has been decoupled from mentions of the mixin dashboards; it mentions the new UI page and logs instead which will always be present
c) There's a new page just for the `mixin` with instructions for people who want to install it and showcasing the clustering dashboards
d) The concepts clustering page is now more streamlined with less 'empty' braggadocious statements. It highlights the target auto-distribution usecase first, and shows the Helm section more clearly as an example.

I'll look into whether there's any diagrams to be added or a best practices section, but let me know if there's any other high-level structure suggestions cc @mattdurham

docs/sources/flow/concepts/clustering.md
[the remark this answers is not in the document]
rfratto: That's better, I like it 👍

docs/sources/flow/concepts/clustering.md
```diff
@@ -0,0 +1,306 @@
...
+
+The ports on the join-addresses list default to the port of the node’s HTTP
+listener if not explicitly provided; it’s generally recommended to align the
+port numbers on as many nodes as possible to simplify the deployment process.
+
+Finally, the first node that is used to bootstrap a new cluster (also known as
+the "seed node") can either omit specifying the flag that specifies peers to
+join or can try to connect to itself.
+
+## Deploy clustering using the Helm chart
```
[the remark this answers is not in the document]
tpaschalis: I've amended this to sound more like an example rather than a guide, and moved the actual instructions around command-line flags page.

I tried spinning something up with docker-compose, but since Flow doesn't have a /-/ready endpoint, it made waiting for the seed node to have something to connect to quite ugly so I think we may leave it for now. WDYT?

docs/sources/flow/concepts/clustering.md
```diff
@@ -0,0 +1,303 @@
...
+To achieve this, Grafana Agent makes use of an eventually consistent model and
+the lightweight [grafana/ckit][] framework. All running nodes are assumed to
+use the same configuration and have access to the same type of hardware and
+network resources such as PVCs and service discovery APIs.
+
+Clustering is built from the ground-up to be the de-facto answer to "how do I
+scale my Grafana Agent setup" in the future. As an example, in comparison to a
+setup using [hashmod sharding][], clustering with target auto-distribution
+scales and provides resiliency at _half_ the cost in resources and provides
+dynamic resharding without the need to change configuration files.
```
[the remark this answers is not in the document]
tpaschalis: Yeah, this paragraph wasn't as useful. I've substituted it for a fairer comparison to hashmod sharding in the intro, to reinforce what the benefits of clustering can be.

```
In comparison to [hashmod sharding][] as a scaling mechanism, clustering with
target auto-distribution provides high availability without requiring multiple
replicas for the same shard, and can dynamically reshard targets without
changes to the configuration file.
```

docs/sources/flow/concepts/clustering.md
mattdurham: I would still like to see an example that doesn't depend on helm/k8s, so people know its not just for k8s. Even if its a one liner running the raw executable multiple times.

docs/sources/flow/concepts/clustering.md
tpaschalis: I've added some more details when moving this back to the run command page. Should we put more detail?

docs/sources/flow/concepts/clustering.md
```diff
@@ -0,0 +1,243 @@
...
+assumes all participating Agents are interchangeable and converge on using the
+same configuration file.
+
+The behavior of a standalone, non-clustered agent is the same as if it was a
+single-node cluster.
+
+In comparison to [hashmod sharding][] as a scaling mechanism, clustering with
+target auto-distribution provides high availability without requiring multiple
+replicas for the same shard, and can dynamically reshard targets without
+changes to the configuration file.
```
rfratto, line 25: Hm, we haven't really ever talked about the hashmod sharding approach w/r/t Flwo mode, and we haven't ever linked to static mode docs from Flow mode either. 

Should we drop the comparison and only talk about the benefits of clustering? 

cc @clayton-cornell

docs/sources/flow/concepts/clustering.md
```diff
@@ -0,0 +1,243 @@
...
+
+Whenever a cluster state change is detected, either due to a new node joining
+or an existing node going away, all participating components locally
+recalculate target ownership and rebalance the number of targets they’re
+scraping without explicitly communicating ownership over the network.
+
+The agent makes use of a fully-local consistent hashing algorithm to distribute
+targets, meaning that on average only ~1/N of the targets are redistributed.
+This is in contrast to hashmod sharding where up to 100% of the targets could
+be reassigned to another node and possibly cause system instability.
```
rfratto, line 50: Ditto here with the comparison: I'm not sure if we should be trying to sell people on clustering compared to other approaches here, and instead we could focus solely on the benefits of clustering (esp. since it's the only scaling mechanism we're choosing to document for Flow mode).

docs/sources/flow/concepts/clustering.md
```diff
@@ -0,0 +1,243 @@
...
+number of agents to distribute workload during peaks, but also provides
+resiliency, since in the event of a node going away, its targets are
+automatically picked up by one of their peers. Again, this is in contrast to
+hashmod sharding which requires running multiple replicas of each shard for HA.
+
+The components who can make use of target auto-distribution are the following:
+- [prometheus.scrape][]
+- [pyroscope.scrape][]
+- [prometheus.operator.podmonitors][]
+- [prometheus.operator.servicemonitors][]
```
rfratto, line 62: This list is going to quickly become stale. Should we make it clear that this list is an example of some components which have clustering support, and how readers can know if a component supports clustering based on the reference page for a component?

docs/sources/flow/concepts/clustering.md
```diff
@@ -0,0 +1,243 @@
...
+
+These components can opt-in to auto-distributing targets between nodes by
+defining the `clustering` block like this:
```
rfratto, line 65:
```suggestion
These components can opt-in to auto-distributing targets between nodes by
adding a `clustering` block with the `enabled` argument set to true: 
```

docs/sources/flow/concepts/clustering.md
```diff
@@ -0,0 +1,243 @@
...
+```river
+prometheus.scrape "default" {
+    clustering {
+      enabled = true
+    }
+    ...
+}
+```
+
+## Local clustering example
```
rfratto, line 75: Do examples belong in concept documentation? Should this be a tutorial, or a getting started guide instead? 

cc @clayton-cornell

docs/sources/flow/concepts/clustering.md
```diff
@@ -0,0 +1,243 @@
...
+
+The following commands bootstrap a three-node local cluster
+```
+# node-a (seed node)
+$ ./build/grafana-agent-flow \
```
rfratto, line 80: Readers might not understand where the `./build` folder came from.

docs/sources/flow/concepts/clustering.md
```diff
@@ -0,0 +1,243 @@
...
+The easiest way to deploy an agent cluster is by making use of our
+[Helm chart][]. Here's an example of how to achieve that.
+
+The following `values.yaml` file deploys anStatefulSet for metrics
+collection. It makes use of a [headless service][] to retrieve the IPs of the
+agent pods for the `--cluster.join-addresses` argument, as well as an
+[Horizontal Pod Autoscaler][] (HPA) for dynamically matching demand.
+
+```yaml
+--- clustering-values.yaml ---
```
rfratto, line 119: Users might accidentally paste this in, I think we should take the filename header out of both code blocks.

docs/sources/flow/concepts/clustering.md
```diff
@@ -0,0 +1,243 @@
...
+of its peers and the details page of the each node's `prometheus.scrape`
+component will point to which target is being scraped by that node in its
+Debug Info.
+
+## Helm chart clustering example
+
+The easiest way to deploy an agent cluster is by making use of our
+[Helm chart][]. Here's an example of how to achieve that.
+
+The following `values.yaml` file deploys anStatefulSet for metrics
```
rfratto, line 113:
```suggestion
The following `values.yaml` file deploys a StatefulSet for metrics
```

docs/sources/flow/concepts/clustering.md
```diff
@@ -0,0 +1,243 @@
...
+collection. It makes use of a [headless service][] to retrieve the IPs of the
+agent pods for the `--cluster.join-addresses` argument, as well as an
+[Horizontal Pod Autoscaler][] (HPA) for dynamically matching demand.
+
+```yaml
+--- clustering-values.yaml ---
+agent:
```
rfratto, line 120: There's a lot of fields here which aren't relevant to enabling clustering. Should we trim this down to only what a user needs to include in their custom values.yaml?

docs/sources/flow/concepts/clustering.md
```diff
@@ -0,0 +1,243 @@
...
+    targetMemoryUtilizationPercentage: 80
+
+service:
+  # -- Creates a Headless Service for the controller's pods.
+  enabled: true
+  type: ClusterIP
+  clusterIP: 'None'
+```
+
+Also, here's a simple River config file
```
rfratto, line 167: There's missing instructions here that tell the user what to do with this file w/r/t to the example. The values.yaml you've instructed the user to use doesn't make use of this, either.

docs/sources/flow/concepts/clustering.md
```diff
@@ -0,0 +1,243 @@
...
+	forward_to = [prometheus.remote_write.default.receiver]
+}
+
+prometheus.remote_write "default"{
+	endpoint {
+		url = env("PROMETHEUS_URL")
+		basic_auth {
+			username = env("PROMETHEUS_USERNAME")
+			password = env("PROMETHEUS_API_KEY")
+		}
```
rfratto, line 193: These environment variables aren't set in the values.yaml you've instructed the user to use.

docs/sources/flow/monitoring/debugging.md
```diff
@@ -69,6 +69,17 @@ The component detail page shows the following information for each component:
...
 
+### Clustering page
+
+![](../../../assets/ui_clustering_page.png)
+
+The clustering page shows the following information for each cluster node:
+
+* The node's name.
+* The node's advertised address.
+* The current state of the node (Viewer/Participant/Terminating).
```
rfratto, line 80: I think we should include a brief explanation for what each of these states means w/r/t clustering.

docs/sources/flow/monitoring/debugging.md
```diff
@@ -89,4 +100,34 @@ The location of Grafana Agent's logs is different based on how it is deployed.
...
 system.
 
+## Debugging clustering issues
+
+To debug issues when using [clustering][], check for the following symptoms.
+
+- **Cluster not converging**: The cluster peers are not converging on the same
+  view of their peers' status. This is most likely due to network connectivity
+issues between the cluster nodes. Use the Flow UI to understand which nodes are
+not being picked up by their peers.
```
rfratto, line 110: This is implying you have to look at the Flow UI for every node, right? I think we should note that here.

docs/sources/flow/monitoring/mixin.md
```diff
@@ -0,0 +1,51 @@
+---
+title: Mixin
+weight: 400
+---
+
+# Mixin
```
rfratto, line 6: It looks like this this should be formatted as a [Task page](https://grafana.com/docs/writers-toolkit/writing-guide/topic-types/task/).

docs/sources/flow/reference/cli/run.md
```diff
@@ -63,26 +63,35 @@ reloading.
 
 [component controller]: {{< relref "../../concepts/component_controller.md" >}}
 
-## Clustered mode (experimental)
+## Clustering (beta)
```
rfratto, line 66: Heads up: there should be a changelog entry indicating that clustering in Flow mode has graduated to beta.

docs/sources/flow/concepts/clustering.md
mattdurham: IMO I like the focus on what we are offering and ignoring hashmod/scraping service mode.

tpaschalis: Ok, leaving the comparisons out for now.

docs/sources/flow/monitoring/debugging.md
tpaschalis: I took a stab to add a high-level explanation in the debugging page, let me know if it reads better now.

docs/sources/flow/monitoring/mixin.md
tpaschalis: I've went ahead and done this. The images showing off the clustering dashboards didn't seem to fit so well now, so I've removed them.

Let me know if you think it's okay under `monitoring` for now or if we should find another place for it.
image

docs/sources/flow/concepts/clustering.md
tpaschalis: I'll include the presence of the `clustering` block as the way to figure out whether a component can opt-in and mention these are example components that can use it.

docs/sources/flow/concepts/clustering.md
clayton-cornell: From our discussion... generally you would keep code examples to topics where they are being discussed (generally tasks and tutorials). But... a concept can definitely have an example code snippet if it helps clarify the top level description of the feature.

docs/sources/flow/concepts/clustering.md
rfratto: @mattdurham I agree, but can we make that a follow up task for when Paschalis is back from PTO?

docs/sources/flow/getting-started/configure-agent-clustering.md
```diff
@@ -0,0 +1,59 @@
+---
+title: Configure agent clustering with Helm
+weight: 400
+---
+
+# Configure agent clustering with Helm
```
rfratto, line 6: @clayton-cornell I imagine this page will have more than just Helm instructions in the future, but I'm not sure if we should start out with having Helm be an `h2`. 

(Also: each deployment environment will have different "before you begin" requirements, since Helm has the unique requirement of a StatefulSet.

WDYT?

rfratto (approved):

mattdurham (commented):
LGTM

docs/sources/flow/getting-started/distribute-prometheus-scrape-load.md
```diff
@@ -0,0 +1,56 @@
...
+changes, and may be replaced with equivalent functionality that cover the same
+use case.
+{{%/admonition %}}
+
+[beta]: {{< relref "../../stability.md#beta" >}}
+[Clustering]: {{< relref "../concepts/clustering.md" >}}
+
+## Before you begin
+
+- Familiarize yourself with how to [configure existing installs][].
```
ptodev, line 24:
```suggestion
- Familiarize yourself with how to [configure existing Agent installations][].
```

docs/sources/flow/getting-started/distribute-prometheus-scrape-load.md
```diff
@@ -0,0 +1,56 @@
...
+- [Configure Prometheus metrics collection][].
+- [Configure clustering][] of agents.
+- Ensure that all of your clustered agents have the same config file.
+
+[configure existing installs]: {{< relref "../setup/configure" >}}
```
ptodev, line 29:
```suggestion
[configure existing agent installations]: {{< relref "../setup/configure" >}}
```

docs/sources/flow/reference/cli/run.md
```diff
@@ -63,26 +63,61 @@ reloading.
...
+must be configured to accept connections on `--server.http.listen-addr` and the
+address defined or inferred in `--cluster.advertise-address`.
 
-The agent will advertise its own address as `--cluster.advertise-address` to
-other agent nodes; if this is empty it will attempt to find a suitable address
-to advertise from a list of default network interfaces. The agent must be
-reachable over HTTP on this address as communication happens over the agent's
-HTTP server.
+If the `--cluster.advertise-address` flag is not explicitly set, the agent
+tries to infer a suitable one from the `eth0` and `en0` local network
```
ptodev, line 81: `eth0` and `en0` are Linux-specific terms. Maybe someone will think it doesn't work on windows?

docs/sources/flow/concepts/clustering.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Use cases
+
+### Target auto-distribution
+
+Target auto-distribution is the most basic use case of clustering; it allows
+scraping components running on all peers to distribute scrape load between
+themselves. All nodes must have access to the same service discovery APIs, and
+the set of targets should converge on a timeline comparable to the scrape
+interval.
```
ptodev, line 34: > the set of targets should converge on a timeline comparable to the scrape interval

- This makes is sound that the time to converge goes up and down depending on the length of the scrape interval that is configured. 
- Also, it's not clear whose responsibility it is to make sure this is the case - does the Agent do it automatically, or is there a config option to make sure the converge interval is similar?

docs/sources/flow/concepts/clustering.md
```diff
@@ -0,0 +1,82 @@
+---
+title: Clustering
+weight: 500
+labels:
+  stage: beta
+---
+
+# Clustering (beta)
```
ptodev, line 8: Would be nice to have a comparison of clustering vs hashmod, and when to chose what in different situations?

ptodev (approved):

docs/sources/flow/getting-started/configure-agent-clustering.md
clayton-cornell: @rfratto Yes, we could simplify the heading to be more generic and bump the Helm Chart up one level.

```
# Configure Grafana Agent clustering

Grafana Agent can be configured to run with [clustering][] so that
individual agents can work together for workload distribution and high
availability.

(and any other intro text)

## Configure Grafana Agent clustering with Helm Chart

(Instructions specific to Helm)
````

This leaves it open/easy to add any future cluster config tooling/options.

Comment (rfratto) [reply docs/sources/flow/getting-started/configure-agent-clustering.md]: Done, let me know what you think :)

File: docs/sources/flow/concepts/clustering.md (hunk 34 above)
Comment (rfratto) [reply docs/sources/flow/concepts/clustering.md:8]: For now we don't want to document anything other than clustering, since we want to only recommend clustering for scaling Flow agent installs.

Comment (ptodev) [reply docs/sources/flow/concepts/clustering.md:8]: I was thinking that eventual consistency might be a problem for some customers? Although that probably shouldn't be the case, since Prometheus is designed in a way that it doesn't have to give 100% accurate data. If super accurate data is what customers want, probably Prometheus isn't the way to go. But then, clustering isn't just for Prometheus :) I'm happy if we don't mention hashmod for now, but I think eventually we might have to.

File: docs/sources/flow/reference/cli/run.md (hunk 32 above)
Comment (rfratto) [reply docs/sources/flow/reference/cli/run.md:81]: That's correct, eth0 and en0 are always used as the defaults meaning it won't work on Windows. I've clarified that Windows users need to provide a value for this flag here.

File: docs/sources/flow/concepts/clustering.md (hunk 33 above)
Comment (rfratto) [reply docs/sources/flow/concepts/clustering.md:34]: i've reworked this sentence a bit and dropped the line about converging the targets. I have a rough idea of what Paschalis is trying to mention here, but I think convergence usually happens fast enough, and scrape intervals are usually high enough, that we don't need to document anything just yet.

File: docs/sources/flow/monitoring/debugging.md [hunk 35]
```diff
@@ -89,4 +100,30 @@ The location of Grafana Agent's logs is different based on how it is deployed.
...
+- **Configuration drift**: Clustering assumes that all nodes are running with
+  the same configuration file at roughly the same time. Check the logs for
+issues with the reloaded configuration file as well as the graph page to verify
+changes have been applied.
+- **Node name conflicts**: Clustering assumes all nodes have unique names;
+  nodes with conflicting names are rejected and will not join the cluster. Look
+at the clustering UI page for the list of current peers with their names, and
+check the logs for any reported name conflict events.
+- **Node stuck in terminating state**: The node attempted to gracefully shut
+  down, set its state to Terminating but has not completely gone away. Check
```
clayton-cornell, line 124:
```suggestion
- **Node stuck in terminating state**: The node attempted to gracefully shut 
down and set its state to Terminating, but it has not completely gone away. Check
```

docs/sources/flow/monitoring/debugging.md
```diff
@@ -89,4 +100,30 @@ The location of Grafana Agent's logs is different based on how it is deployed.
...
+the clustering page to view the state of the peers and verify that the
+terminating Agent has indeed been shut down.
```
clayton-cornell, line 126:
```suggestion
terminating Agent has been shut down.
```

docs/sources/flow/reference/cli/run.md
```diff
@@ -63,26 +63,63 @@ reloading.
...
+interfaces. If the advertised address cannot be determined, the agent will
+fail to start. Windows users must explicitly pass a value for
+`--cluster.advertise-address`, since Windows does not have interfaces name
+`eth0` or `en0`.
+
+The comma-separated list of addresses provided in `--cluster.join-addresses`
+can either be IP addresses with an optional port, or DNS records to lookup.
+The ports on the this list of addreses default to the port used for the HTTP
```
clayton-cornell, line 89:
```suggestion
The ports on the list of addresses default to the port used for the HTTP
```

docs/sources/flow/reference/cli/run.md
```diff
@@ -63,26 +63,63 @@ reloading.
...
+listener if not explicitly provided. It’s generally recommended to
+align the port numbers on as many nodes as possible to simplify the deployment
```
clayton-cornell, line 91:
```suggestion
listener if not explicitly provided. We recommend that you
align the port numbers on as many nodes as possible to simplify the deployment
```

clayton-cornell (commented):
Some suggestions added

docs/sources/flow/concepts/clustering.md
```diff
@@ -0,0 +1,82 @@
...
+As such, target auto-distribution not only allows to dynamically scale the
+number of agents to distribute workload during peaks, but also provides
+resiliency, since in the event of a node going away, its targets are
+automatically picked up by one of their peers.
+
+The agent makes use of a fully-local consistent hashing algorithm to distribute
+targets, meaning that on average only ~1/N of the targets are redistributed.
+
+Refer to component reference documentation to discover whether it supports
+clustering, such as:
```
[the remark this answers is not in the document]
rfratto: Does this imply that the list is exhaustive? That's what I'm trying to avoid, since I'm worried this list will quickly become stale and annoying to maintain separately from reference docs.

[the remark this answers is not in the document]
clayton-cornell: Mmmm... yeah good point. I've edited the suggestion to add `for example`. That relieves the pressure for it being a complete list of all cases.

clayton-cornell (approved):

PR #4347, opened 2023-07-05
introduce Critical severity level for converter

erikbaranowski:
#### PR Description

This PR introduces a new diagnostic level for the converter API of Critical. This allows us to bump up existing Warning level diagnostics to Error and creates better clarity for the CLI commands.

#### PR Checklist

- [ ] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/cmd/internal/flowmode/cmd_convert.go
+++ b/cmd/internal/flowmode/cmd_convert.go
@@ -16,9 +16,9 @@ import (
 
 func convertCommand() *cobra.Command {
 	f := &flowConvert{
-		output:         "",
-		sourceFormat:   "",
-		bypassWarnings: false,
+		output:       "",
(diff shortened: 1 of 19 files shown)


rfratto:
This makes more sense to me compared to warning/error. 

When would you suggest that something is treated as critical instead of an error?

erikbaranowski:
Critical to me is when a river output cannot be generated (there are some usages of it you can find in the file diffs). A converter could also decide that something in the original config is so problematic they want to call it Critical as well.

rfratto (approved):
LGTM, I like it.

mattdurham (approved):
This feels much clearer

clayton-cornell (commented):
LGTM

Issue #4337, opened 2023-07-04
Timestamp stage not working on windows


PsySuck:
### What's wrong?

I'm trying to use the timestamp stage to change the time of an event. In promtail, the timestamp stage works, but it does not work when transferring the config to Grafana Agent.
<img width="345" alt="Screenshot 2023-07-04 at 12 35 08" src="https://github.com/grafana/agent/assets/87027160/cc6c8fe3-4aed-401e-8323-117431107188">

Inspect in promtail
```
{stages.Entry}.Entry.Entry.Timestamp:
        -: 2023-07-04 14:09:21.986855 +0300 EEST
        +: 2023-07-01 00:00:01 +0300 EEST
```

### Steps to reproduce

1. Install Agent.
2. Add Timestamp with Location and Custom Format
3. Run agent 

### System information

Windows server 2019

### Software version

agent, version v0.34.3 

### Configuration

```text
Promtail:
- job_name: iis-frontend
  static_configs:
    - targets:
        - localhost
      labels:
        __path__: E:\IIS_Logs\W3SVC1\u_ex23*.log
  pipeline_stages:
    - match:
          # Drop lines that are comments (start with #)
        selector: '{service="exchange"} |~ "^#"'
        action: drop
        drop_counter_reason: iis_comment_line
                
    - regex:
        source:
        expression: "^(?P<time>\\d{4}-\\d{2}-\\d{2} \\d{2}:\\d{2}:\\d{2}) "
            
    - timestamp:
        source: time
        format: 2006-01-02 15:04:05
        location: Europe/Sofia
Grafana Agent:       
 - job_name: iis-frontend
          static_configs:
            - targets:
                - localhost
              labels:
                service: exchange
                log_type: iis-frontend
                __path__: E:\IIS_Logs\W3SVC1\u_ex*.log
          pipeline_stages:
            - match:
          # Drop lines that are comments (start with #)
                selector: '{service="exchange"} |~ "^#"'
                action: drop
                drop_counter_reason: iis_comment_line
                
            - regex:
                source:
                expression: "^(?P<time>\\d{4}-\\d{2}-\\d{2} \\d{2}:\\d{2}:\\d{2}) (?P<csmethod>.+?) (?P<csuristem>.+?) (?P<csuriquery>.+?) (?P<csusername>.+?) (?P<cip>.+?) (?P<useragent>.+?) (?P<cookie>.+?) (?P<scstatus>.+?) (?P<ip>.+?):"
            
            - timestamp:
[...]

PR #4351, opened 2023-07-05
grafana-agent-flow: ensure time/tzdata is imported

rfratto:
time/tzdata needs to be imported to ensure all timezones are embedded into the binary. pkg/logs already imported this for the grafana-agent binary, but the grafana-agent-flow binary did not have it in its import path.

Related to #4337.


tpaschalis (approved):
Good catch, ty!

PR #4353, opened 2023-07-05
flow: Allow searching for components within modules

rfratto:
This changes the implementation of the `component.Provider` interface by the Flow package to successfully retrieve components in other modules, and to navigate the entire hierarchy of modules. 

This will allow a single HTTP handler to invoke the correct handler for a component within a module, without requiring the module loader to implement an HTTP handler itself.


cmd/internal/flowmode/cmd_run.go
```diff
@@ -365,8 +365,7 @@ func (fr *flowRun) Run(configFile string) error {
 // getEnabledComponentsFunc returns a function that gets the current enabled components
 func getEnabledComponentsFunc(f *flow.Flow) func() map[string]interface{} {
 	return func() map[string]interface{} {
-		components := f.ListComponents(component.InfoOptions{})
-
+		components := component.GetAllComponents(f, component.InfoOptions{})
```
rfratto, line 368: This means usage stats will now contain stats across all modules, which previously wasn't happening 🎉

pkg/flow/flow_components.go
```diff
@@ -13,8 +13,15 @@ func (f *Flow) GetComponent(id component.ID, opts component.InfoOptions) (*compo
...
-	// TODO(rfratto): navigate running modules and return a component within a
-	// module.
+	if id.ModuleID != "" {
+		mod, ok := f.modules.Get(id.ModuleID)
+		if !ok {
+			return nil, fmt.Errorf("component %q does not exist", id)
+		}
+
+		// TODO(rfratto): can mod.f be nil here?
+		return mod.f.GetComponent(component.ID{LocalID: id.LocalID}, opts)
```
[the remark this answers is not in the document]
mattdurham: There is a window where `mod.f` can be nil. Quick look seems like moving the `mod.f` instantiation to `newModule` should work fine.

pkg/flow/flow_components.go
```diff
@@ -32,11 +39,19 @@ func (f *Flow) GetComponent(id component.ID, opts component.InfoOptions) (*compo
 }
 
 // ListComponents implements [component.Provider].
-func (f *Flow) ListComponents(opts component.InfoOptions) []*component.Info {
+func (f *Flow) ListComponents(moduleID string, opts component.InfoOptions) ([]*component.Info, error) {
```
mattdurham, line 42: IMO I think we need types/doc around global, module, component IDs. Its far to easy to get confused about each one. This may not be the PR for it but when I see this here I feel I have to think about what type of ID it wants.

pkg/flow/flow.go
```diff
@@ -137,6 +138,10 @@ type Flow struct {
 // New creates and starts a new Flow controller. Call Close to stop
 // the controller.
 func New(o Options) *Flow {
+	return newController(newModuleRegistry(), o)
+}
+
+func newController(modReg *moduleRegistry, o Options) *Flow {
```
mattdurham, line 144: `newController` feels like an overloaded term here.  Also `New` might need to be renamed to `NewRoot`  (that feels like a terrible name) or `NewController`. It feels odd to know when to use which New/newController.

rfratto: It might make more sense to rename it in the future when we combine packages but I don't think we should change public API names because of internal implementation details. 

I can document when to call newController internally.

mattdurham (approved):
LGTM

erikbaranowski (approved):
LGTM

PR #4356, opened 2023-07-06
Decouple the API used by the UI

rfratto:
This PR decouples the UI's API into a single consumer (service/http). The API now serves requests for all components across all modules rather than requiring each module loader to expose the API as part of its HTTP handlers. 

See individual commit messages for more details. 

I have successfully tested these changes by rebuilding the UI and using a multi-module config, with module nesting and components within modules which have references to other components in the same module.

### DIFF
--- a/component/component_provider.go
+++ b/component/component_provider.go
@@ -86,9 +86,10 @@ func (info *Info) MarshalJSON() ([]byte, error) {
 		}
 
 		componentDetailJSON struct {
-			Name         string               `json:"name,omitempty"`
+			Name         string               `json:"name"`
 			Type         string               `json:"type,omitempty"`
-			ID           string               `json:"id,omitempty"`
+			ID           string               `json:"id"`
+			ModuleID     string               `json:"moduleID"`
 			Label        string               `json:"label,omitempty"`
 			References   []string             `json:"referencesTo"`
 			ReferencedBy []string             `json:"referencedBy"`
@@ -97,6 +98,7 @@ func (info *Info) MarshalJSON() ([]byte, error) {
 			Arguments    json.RawMessage      `json:"arguments,omitempty"`
 			Exports      json.RawMessage      `json:"exports,omitempty"`
 			DebugInfo    json.RawMessage      `json:"debugInfo,omitempty"`
+			ModuleIDs    []string             `json:"moduleIDs,omitempty"`
 		}
 	)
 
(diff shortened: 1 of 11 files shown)


rfratto:
This is related to the #4253 work, where I'm first trying to fully remove HTTP code out from pkg/flow. This is the first step of it; the next step is to do the same for the `/api/v0/component` endpoint.

web/api/api.go
```diff
@@ -28,14 +29,26 @@ func NewFlowAPI(flow component.Provider, node cluster.Node) *FlowAPI {
...
+	r.Handle(path.Join(urlPrefix, "/components/{id:.+}"), httputil.CompressionHandler{Handler: f.getComponentHandler()})
 	r.Handle(path.Join(urlPrefix, "/peers"), httputil.CompressionHandler{Handler: f.getClusteringPeersHandler()})
 }
 
 func (f *FlowAPI) listComponentsHandler() http.HandlerFunc {
-	return func(w http.ResponseWriter, _ *http.Request) {
-		components, err := f.flow.ListComponents("", component.InfoOptions{
+	return func(w http.ResponseWriter, r *http.Request) {
+		// moduleID is set from the /modules/{moduleID:.+}/components route above
+		// but not from the /compoents route.
```
mattdurham, line 45: nit: compoents -> components

component/component_provider.go
```diff
@@ -131,7 +133,8 @@ func (info *Info) MarshalJSON() ([]byte, error) {
 	return json.Marshal(&componentDetailJSON{
 		Name:         info.Registration.Name,
 		Type:         "block",
-		ID:           info.ID.LocalID, // TODO(rfratto): support getting component from module.
+		ModuleID:     info.ID.ModuleID,
+		ID:           info.ID.LocalID,
```
mattdurham, line 137: Should we name this `LocalID` to make it clearer?

component/component_provider.go
```diff
@@ -143,6 +146,7 @@ func (info *Info) MarshalJSON() ([]byte, error) {
 		Arguments: arguments,
 		Exports:   exports,
 		DebugInfo: debugInfo,
+		ModuleIDs: info.ModuleIDs,
```
mattdurham, line 149: Thoughts on changing this name to something like `Child(Sub)ModuleIDs`? This feels like you have to think what ModuleID and ModuleIDs are in this context.

mattdurham (approved):
Minor naming nits but looks alright. I gave the typescript a very minor overview.

component/component_provider.go
rfratto: I went for `CreatedModuleIDs` to make it clearer that they're modules created by the component, LMK if you think ChildModuleIDs/SubModuleIDs is preferable.

erikbaranowski (approved):
LGTM

PR #4357, opened 2023-07-06
flow: sanitize tests when testing file paths

rfratto:
Sanitize the filepaths used in test assertions to always use `/` characters.

This fixes an issue introduced by tests added in #4310, where the new tests failed on Windows since the returned paths used the default Windows path separator `\` instead.


spartan0x117 (approved):
Thanks for catching!

PR #4361, opened 2023-07-06
flow: decouple component HTTP API

rfratto:
This PR decouples the HTTP API exposed by component.HTTPComponents (e.g., `/api/v0/component/COMPONENT_ID/`) to be handled centrally by the HTTP service. 

This allows us to remove HTTP handler logic from the Flow controller and means that module loaders no longer have to implement their own HTTP handlers to expose handlers from components within that module.

See individual commits for details.


erikbaranowski (approved):
I noted a design concern but I'm good with moving forward for now.

mattdurham (approved):
Some concerns about future complexity in split path but works for now

PR #4362, opened 2023-07-06
Use `rivertypes.OptionalSecret` for SNMP and Blackbox components inline config

### LINKED ISSUE #4302: "config" argument in prometheus.exporter.blackbox should be an optional secret
### Request

The [config](https://grafana.com/docs/agent/latest/flow/reference/components/prometheus.exporter.blackbox/#arguments) argument contains the full text [config](https://github.com/prometheus/blackbox_exporter/blob/master/CONFIGURATION.md) of a blackbox exporter. This could contain secrets such as passwords. In order to not expose the passwords, we should consider making this field an "optional secret" instead of a "string".

### Use case

Extra security.

marctc:
#### PR Description

This PR changes inline `config` fields of blackbox and snmp exporter to be type `Secret`.

Fixes #4302

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/component/prometheus/exporter/blackbox/blackbox.go
+++ b/component/prometheus/exporter/blackbox/blackbox.go
@@ -13,6 +13,7 @@ import (
 	"github.com/grafana/agent/component/prometheus/exporter"
 	"github.com/grafana/agent/pkg/integrations"
 	"github.com/grafana/agent/pkg/integrations/blackbox_exporter"
+	"github.com/grafana/agent/pkg/river/rivertypes"
 )
 
 func init() {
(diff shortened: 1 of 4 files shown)


component/prometheus/exporter/blackbox/blackbox.go
```diff
@@ -81,10 +82,10 @@ func (t TargetBlock) Convert() []blackbox_exporter.BlackboxTarget {
 }
 
 type Arguments struct {
-	ConfigFile         string        `river:"config_file,attr,optional"`
-	Config             string        `river:"config,attr,optional"`
-	Targets            TargetBlock   `river:"target,block"`
-	ProbeTimeoutOffset time.Duration `river:"probe_timeout_offset,attr,optional"`
+	ConfigFile         string            `river:"config_file,attr,optional"`
+	Config             rivertypes.Secret `river:"config,attr,optional"`
```
rfratto, line 86: Is the config always sensitive? If not, this should be a rivertypes.OptionalSecret so it can optionally be sensitive depending on the River passed to it.

rfratto (approved):
LGTM!

clayton-cornell (approved):
Looks ok to me

PR #4369, opened 2023-07-07
Expose loki.porcess stages package for config conversions

thampiotr:
#### PR Description
In order to implement promtail config conversions, we'll need to create `loki.process` stages, which are currently internal package. 

We need to expose it in order to continue. As an alternative, we could write some conversion logic inside the `process` package, but it may not be a desired way to organise code.


tpaschalis:
If we go ahead with this approach, we should also move the `metrics` package from under `loki/process/internal/, right?

tpaschalis (approved):
After an offline sync, I'm approving this so we don't have to break the convention of where conversion-related code exists for now.

Once this is done however, I'd like to reconsider moving stuff behind an `internal` package; let's add a `TODO` comment here so that we don't forget to do that.

PR #4371, opened 2023-07-07
cluster: transition to StateTerminating before leaving the cluster

tpaschalis:
#### PR Description
This PR ensures that cluster nodes shutting down transition to the StateTerminating before leaving the memberlist, so that they're no longer considered owners for write hash operations.

No issue filed.

#### Notes to the Reviewer
We introduce a new context here, because when in order to call Stop() on the deferred function, it would mean that the context from `interruptContext` would already be cancelled.

#### PR Checklist

- [ ] CHANGELOG updated (N/A)
- [ ] Tests updated (N/A)

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -231,9 +231,13 @@ func (c *Clusterer) Start(ctx context.Context) error {
 }
 
 // Stop stops the Clusterer.
-func (c *Clusterer) Stop() error {
+func (c *Clusterer) Stop(ctx context.Context) error {
 	switch node := c.Node.(type) {
 	case *GossipNode:
+		err := node.ChangeState(ctx, peer.StateTerminating)
+		if err != nil {
+			level.Error(node.log).Log("msg", "failed to change state to Terminating before shutting down", "err", err)
+		}
 		return node.Stop()
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/cluster/cluster.go
```diff
@@ -234,6 +234,15 @@ func (c *Clusterer) Start(ctx context.Context) error {
...
 	case *GossipNode:
+		// The node is going away. We move to the Terminating state to signal
+		// that we should not be owners for write hashing operations anymore.
+		ctx, ccl := context.WithTimeout(context.Background(), 5*time.Second)
+		defer ccl()
+
+		err := node.ChangeState(ctx, peer.StateTerminating)
+		if err != nil {
+			level.Error(node.log).Log("msg", "failed to change state to Terminating before shutting down", "err", err)
+		}
```
rfratto, line 245: Should this happen earlier? i.e., should this happen before stopping the Flow controller?

rfratto: Can you add a TODO for us to investigate this?

rfratto (approved):
LGTM

rfratto:
Hijacking this PR so we can get it merged for v0.35 :)

PR #4372, opened 2023-07-07
helm: add native support for clustering with agent.clustering.enabled value

rfratto:
This adds a new value, `agent.clustering.enabled`, to automatically configure clustering with the appropriate service and flags. 

This makes it easier to deploy clustering by having a dedicated service for clustering, compared to the current approach of disabling the existing service, and then switching it to a headless service. 

Because clustering only discovers peers once on startup, and because StatefulSet is the only controller which can roll out one pod at a time, controller.type=statefulset is required to use clustering. A long-term solution would be to periodically rediscover peers, which would allow other controller types to be used correctly.

### DIFF
--- a/operations/helm/charts/grafana-agent/CHANGELOG.md
+++ b/operations/helm/charts/grafana-agent/CHANGELOG.md
@@ -16,7 +16,8 @@ Unreleased
 ### Features
 
 - Add native support for Flow mode clustering with the
-  `agent.clustering.enabled` flag. (@rfratto)
+  `agent.clustering.enabled` flag. Clustering may only be enabled in Flow mode
+  when deploying a StatefulSet. (@rfratto)
 
 ### Enhancements
 
--- a/operations/helm/charts/grafana-agent/CHANGELOG.md
+++ b/operations/helm/charts/grafana-agent/CHANGELOG.md
@@ -13,6 +13,11 @@ Unreleased
 0.16.0 (2023-06-20)
 -------------------
 
+### Features
+
+- Add native support for Flow mode clustering with the
+  `agent.clustering.enabled` flag. (@rfratto)
+
 ### Enhancements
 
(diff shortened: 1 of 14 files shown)


tpaschalis (approved):
LGTM

tpaschalis (approved):
LGTM with the new condition around StatefulSets.

tpaschalis:
This was released as [v0.17.0](https://github.com/grafana/agent/blob/main/operations/helm/charts/grafana-agent/CHANGELOG.md#0170-2023-07-19) last week!

We'd love to hear your feedback, how clustering works for you, or if you have any questions on how to set it up!

tpaschalis:
@braunsonm This was released as [v0.17.0](https://github.com/grafana/agent/blob/main/operations/helm/charts/grafana-agent/CHANGELOG.md#0170-2023-07-19) last week!

We'd love to hear your feedback, how clustering works for you, or if you have any questions on how to set it up!

braunsonm:
Thanks @tpaschalis it seems to work well. I'm not sure where to leave feedback so I'll drop it here.

The lack of support for Grafana Agent to discover peers at runtime is really killer. Right now I think we have a simple usecase - collect logs from our Kubernetes cluster and send them to Loki, and respect Prometheus Operator CRs to collect metrics and send them to mimir. Unfortunately because I can't run Grafana Agent as a Daemonset I am running it as a statefulset that has antiaffinity so only 1 runs per host. This involves making sure you match the replicas with the number of nodes.

It would be great if either daemonset support was added (discover peers at runtime) or this was implemented: https://github.com/grafana/agent/issues/4502 this way logs for pods can be collected with flow mode without sending duplicates. Thanks for the work!

tpaschalis:
Thank you for the feedback and the new ideas you've been proposing @braunsonm 🙌 

I've opened [an issue](https://github.com/grafana/agent/issues/4596) to keep track of runtime peer discovery.

PR #4384, opened 2023-07-10
introduce a report generation flag to the convert CLI command

erikbaranowski:
#### PR Description

introduce a report generation flag to the convert CLI command
- includes a way to remove a Severity type from the diags especially helpful in testing if we want to skip Info severity testing.
- extendable for future report types without an API change
- fixes a 1 line prometheus bug I noticed on my way

does not in this PR
- include useful info diags for prometheus
- implement a shared testing package for converters so we don't have to repeat as much of the test code everywhere

#### PR Checklist

- [ ] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated


converter/diag/diagnostics.go
```diff
@@ -81,3 +81,14 @@ func (ds Diagnostics) Error() string {
...
+
+func (ds Diagnostics) GenerateReport(reportType string, filename string) error {
+	switch reportType {
+	case Text:
+		return generateTextReport(ds, filename)
+	case HTML:
+		return generateHTMLReport(ds, filename)
+	default:
+		return fmt.Errorf("unsupported diagnostic report type %q", reportType)
+	}
```
rfratto, line 93: Idiomatically, this should accept a `io.Writer` argument instead of always writing the report to a file.

converter/diag/diagnostics.go
```diff
@@ -81,3 +82,24 @@ func (ds Diagnostics) Error() string {
...
 }
+
+func (ds Diagnostics) GenerateReport(reportType string, writer io.Writer) error {
+	switch reportType {
+	case Text:
+		return generateTextReport(ds, writer)
+	default:
+		return fmt.Errorf("unsupported diagnostic report type %q", reportType)
+	}
+}
```
rfratto, line 93: I may be making this up, but I think usually io.Writer is the first argument to a function like this (a la Fprint)

rfratto (approved):

mattdurham (approved):
LGTM

docs/sources/flow/reference/cli/convert.md
```diff
@@ -21,6 +21,9 @@ equal to `-`, `grafana-agent convert` converts the contents of standard input. O
 The `--output` flag can be specified to write the contents of the converted
 config to the specified path.
 
+The `--report` flag can be used to generate a diagnostic report. When
+`--report` is not provided, no report is generated.
+
 The `--bypass-errors` flag can be used to bypass [errors].
```
clayton-cornell, line 27:
```suggestion
There are several different flags available for the `grafana-agent convert` command. You can use the `--output` flag to write the contents of the converted config to a specified path. You can use the `--report` flag to generate a diagnostic report. The `--bypass-errors` flag allows you to bypass any [errors] generated during the file conversion.
```

Maybe we could change this up a little to make it a descriptive paragraph so it doesn't look/read like a simple restate of the bullet list of flags

Issue #3607, opened 2023-04-24
Flow: replace discovery_target_decode with components


rfratto:
discovery_target_decode was introduced as a way to act as a substitute for the functionality of Prometheus' HTTP and file discovery with `local.file` and `remote.http`. 

However, this approach was flawed: 

* Prometheus file discovery supports YAML files, which cannot be done with local.file/discovery_target_decode. 
* I think it was a mistake to introduce component-specific functions. 

I would like to replace `discovery_target_decode` with two components:

* `discovery.file_targets` for Prometheus' file SD. 
  * `discovery.file` is already in-use for file globbing, unless we renamed that component.  
* `discovery.http` for Prometheus' HTTP SD. 

And going forward, I'd like to change the guidelines around the stdlib to _not_ include any component-specific functions, and keep the stdlib generic. 

I'm looking for feedback on this proposal.

zackman0010:
Hi! What's the priority on this kind of functionality? We recently moved to using file-based discovery for our own services, where on startup each service creates its own file in a specified folder. Our Static configuration uses file globbing to find all the files in that folder to dynamically locate the targets. We can't hardcode the files due to their usage of random ports, so the inability to dynamically read multiple files is unfortunately a hard blocker on us moving to use Flow.

zackman0010:
@rfratto Just to share my own opinion, I like the second choice for one reason: the current existing `discovery.file` doesn't actually return the files themselves, it instead returns the file _paths_. So maybe the old could be renamed `discovery.filepath`?

mattdurham:
Some random thoughts:
* A new and maybe not needed `local.files` should return a `map(string)` where the key is the filename and the value is the content. This aligns with `local.file`
* `discovery.file` stays as is
* `discovery.file_targets` handles a more label based Prometheus approach.

mattdurham:
I should add that `local.files` likely is not immediately useful without adding some additional exports and functions.

tpaschalis:
My 2c is to name the new component `discovery.file_targets`.

The thing is, keeping the current `discovery.file` name isn't doing us any favors, and might continue being confusing. Maybe we can _also_ rename the existing component to `discovery.fileglob` and have a clearer distinction of what each is doing?

> `local.files` should return a map(string) where the key is the filename and the value is the content
This could also be a module once/if #2783 is implemented, right?

rfratto:
I'm comfortable with both components being renamed to not be ambiguous, especially if there's going to be two file-based discovery components. 

[...]

PR #4394, opened 2023-07-11
breaking change: rename discovery.file to local.file_match

rfratto:
This commit renames discovery.file to local.file_match so that:

* The purpose of the component is more clear (i.e., to find files on the local filesystem which match a given pattern).

* To remove a point of confusion where `discovery.file` was thought to be Prometheus' file SD.

* To allow a component for Prometheus' file SD to be named discovery.file, which otherwise would've collided with the now-named local.file_match.

A redirect from the old documentation page is explicitly not configured because we anticipate creating a new page called discovery.file, which will have a warning about the rename.

Related to #3607.


captncraig (approved):

jcreixell (approved):

clayton-cornell (approved):
LGTM

PR #4395, opened 2023-07-11
Do not fail on startup when a module loader receives invalid config

marctc:
#### PR Description

This PR changes the behavior of module loaders which currently make the Agent panic on startup if the module loaders
loads a Flow config with errors. In cases like `module.http` (specially using it combined with agent-management) it's important to keep the agents running regardless of the loaded configuration.

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/component/module/file/file.go
+++ b/component/module/file/file.go
@@ -6,6 +6,7 @@ import (
 
 	"go.uber.org/atomic"
 
+	"github.com/go-kit/log/level"
 	"github.com/grafana/agent/component"
 	"github.com/grafana/agent/component/local/file"
 	"github.com/grafana/agent/component/module"
@@ -75,7 +76,8 @@ func New(o component.Options, args Arguments) (*Component, error) {
 		return nil, err
 	}
 	if err := c.Update(args); err != nil {
-		return nil, err
+		level.Error(o.Logger).Log("msg", "loading of component failed", "err", err)
+		return c, nil
 	}
 	return c, nil
 }
--- a/component/module/git/git.go
+++ b/component/module/git/git.go
@@ -87,7 +87,8 @@ func New(o component.Options, args Arguments) (*Component, error) {
 	}
 
 	if err := c.Update(args); err != nil {
-		return nil, err
+		level.Error(o.Logger).Log("msg", "loading of component failed", "err", err)
+		return c, nil
 	}
 	return c, nil
 }
--- a/component/module/http/http.go
+++ b/component/module/http/http.go
@@ -6,6 +6,7 @@ import (
 
 	"go.uber.org/atomic"
 
+	"github.com/go-kit/log/level"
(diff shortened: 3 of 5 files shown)


mattdurham:
Talked with Marc about this and this seems like a bug. If the main config is loaded correctly but a module fails then we should retry since it could be a network error or something that will resolve itself. This happens because New returns an error from the child component. Also adding @erikbaranowski for his thoughts.

component/module/file/file.go
```diff
@@ -75,7 +76,8 @@ func New(o component.Options, args Arguments) (*Component, error) {
 		return nil, err
 	}
 	if err := c.Update(args); err != nil {
-		return nil, err
+		level.Error(o.Logger).Log("msg", "loading of component failed", "err", err)
```
jcreixell, line 79: do we want to add telemetry for module loading failures?

rfratto (changes requested):
This change will inadvertently break some expected behavior about how component evaluation works, and how errors are reported as close to the point of origin as possible. 

To avoid compounding evaluation errors, components _need_ to fail early. For example, imagine this scenario:

```river 
local.file "example" {
  filename = "INVALID_FILE" 
}

remote.http "example" {
  url = json_decode(local.file.example.content).HTTP_URL
}
```

Note that local.file is given an invalid file here. If it _doesn't_ fail, then the error message reported to the user is about remote.http.example instead, even though local.file.example was actually the problematic component. 

This applies to module loaders too: module loaders _must_ fail fast if they're given an invalid configuration, otherwise unexpected errors will be returned to the user when using the exports of a silently-failed component. 

> In cases like module.http (specially using it combined with agent-management) it's important to keep the agents running regardless of the loaded configuration.

If this is the goal, I think this problem is being solved at the wrong level. A different approach could be to remove the restriction that the initial load of the controller must succeed, or make that condition optional:

https://github.com/grafana/agent/blob/da959e0da39d37442751b1685c868ffba47dfd1e/pkg/flow/flow.go#L269-L273

mattdurham:
I had originally chatted with Marc and talked him out of making it optional behavior and now I think I was wrong. 

It feels like there are un-fixable errors and then there are errors that resolve (network) and currently we can't differentiate between the two globally but a component could. Thoughts on allowing the module loaders (or any component) to more closely control their lifecycle, currently the `module.http` could detect a network error and retry x/forever times but not return which would hold off further processing the graph.

There would be an issue with debugging I imagine, though I think the debug info would be available if we were careful on how we handle the mutexes.

mattdurham:
Psuedocode

```
	level.Error(o.Logger).Log("msg", "loading of component failed", "err", err)
        if isRecoveralble(err) {
            LOOP_AND_TRY_AGAIN_X_TIMES
        }
}
``

PR #4400, opened 2023-07-11
Add quick detail on meta labels.

mattdurham:
Add a very small bit of doc on meta labels, I was going to add instructions on how to keep them but this doesn't feel like the right place.

### DIFF
--- a/docs/sources/flow/reference/components/prometheus.scrape.md
+++ b/docs/sources/flow/reference/components/prometheus.scrape.md
@@ -262,7 +262,7 @@ http://blackbox-exporter:9116/probe?target=grafana.com&module=http_2xx
 `prometheus.scrape` supports [gzip](https://en.wikipedia.org/wiki/Gzip) compression.
 
 The following special labels can change the behavior of prometheus.scrape:
-* `__address__` is the name of the label that holds the address of a scrape target.
+* `__address__` is the name of the label that holds the `<host>:<port>` address of a scrape target.
 * `__metrics_path__`   is the name of the label that holds the path on which to scrape a target.
 * `__scheme__` is the name of the label that holds the scheme (http,https) on which to  scrape a target.
 * `__scrape_interval__` is the name of the label that holds the scrape interval used to scrape a target.


docs/sources/flow/reference/components/prometheus.remote_write.md
```diff
@@ -332,3 +332,7 @@ prometheus.scrape "demo" {
 ## Compression
 
 `prometheus.remote_write` uses [snappy](https://en.wikipedia.org/wiki/Snappy_(compression)) for compression.
+
+## Meta labels
+
+Any labels that start with `__` will be removed before sending to the endpoint.
```
rfratto, line 338: Does this belong as a dedicated H2, or should it go closer to the documentation for where data is sent (the `endpoint` block)?

I'm concerned that adding too many h2s breaks the consistency patterns of how we document reference pages for components. 

cc @clayton-cornell

mattdurham: I am fine with that too, I think there is an unknown in where to put these little implementation details.

clayton-cornell: Could this be formatted as a note?

mattdurham: Is there a place for something like rolling up the compression/meta into this? Though it feels very grab bag. 

```
## Implementation details

* `prometheus.remote_write` uses [snappy](https://en.wikipedia.org/wiki/Snappy_(compression)) for compression.
* Any labels that start with `__` will be removed before sending to the endpoint.
```

Its almost like a FAQ or troubleshooting guidelike. The entry point might be why are my labels not being written to prometheus or what compression does remote_write use

mattdurham:
I added technical details has a catchall. I dont think the name is great and very open to ideas though its something we can workshop later if we dont find a concensus.

Also added special labels

docs/sources/flow/reference/components/prometheus.scrape.md
```diff
@@ -256,6 +256,19 @@ http://blackbox-exporter:9115/probe?target=grafana.com&module=http_2xx
 http://blackbox-exporter:9116/probe?target=grafana.com&module=http_2xx
 ```
 
-## Compression
+### Technical details
 
 `prometheus.scrape` supports [gzip](https://en.wikipedia.org/wiki/Gzip) compression.
+
+Special Labels that can change the behavior of prometheus.scrape
```
clayton-cornell, line 264:
```suggestion
The following special labels can change the behavior of prometheus.scrape:
```

docs/sources/flow/reference/components/prometheus.scrape.md
```diff
@@ -256,6 +256,19 @@ http://blackbox-exporter:9115/probe?target=grafana.com&module=http_2xx
...
+* `__address__` is the name of the label that holds the address of a scrape target.
+* `__metrics_path__`   is the name of the label that holds the path on which to scrape a target.
+* `__scheme__` is the name of the label that holds the scheme (http,https) on which to  scrape a target.
+* `__scrape_interval__` is the name of the label that holds the scrape interval used to scrape a target.
+* `__scrape_timeout__` is the name of the label that holds the scrape timeout used to scrape a target.
+* `__param__` is a prefix for labels that provide URL parameters used to scrape a target.
+
+* Special labels added after a scrape
```
clayton-cornell:
```suggestion
The following special labels are added after a scrape:
```

clayton-cornell (commented):
Some suggestions for formatting/syntax

docs/sources/flow/reference/components/prometheus.scrape.md
```diff
@@ -257,6 +257,19 @@ http://blackbox-exporter:9115/probe?target=grafana.com&module=http_2xx
 http://blackbox-exporter:9116/probe?target=grafana.com&module=http_2xx
 ```
 
-## Compression
+### Technical details
 
 `prometheus.scrape` supports [gzip](https://en.wikipedia.org/wiki/Gzip) compression.
+
+The following special labels can change the behavior of prometheus.scrape:
+* `__address__` is the name of the label that holds the address of a scrape target.
```
tpaschalis, line 265:
```suggestion
* `__address__` is the name of the label that holds the `<host>:<port>`  address of a scrape target.
```

WDYT? It'd tripped me up in the past, may be helpful here.

tpaschalis (approved):

docs/sources/flow/reference/components/prometheus.scrape.md
clayton-cornell: Making it clear exactly what address... a good idea. (side thought: Are there other areas where being more explicit like thsi will help?)

There's an extra space between port and address.

docs/sources/flow/reference/components/prometheus.remote_write.md
clayton-cornell: I've been thinking about this, and don't yet have a suggestion.

* What would a user be doing with Agent where they'd need to know this info? Initial config? Fine tuning? Trying to fix a problem with data flow/collection?
* Is this simply nice to know info? or more critical to a successful Agent setup?

mattdurham: Its a common gotcha that people remark on. Usually like where are my labels? The compression one is more from a customer asking details to make sure we arent writing 5x the data we should.

clayton-cornell: @mattdurham This PR is getting old. I suggest we merge it as-is even though it's not 100% what we'd like. Get the info at least published and we can revisit this as it evolves.

PR #4404, opened 2023-07-12
[Flow] Add new discovery.file component

### LINKED ISSUE #3346: Support for file_sd_configs like behavior in flow
The agent in flow is missing support for finding files in the filesystem and parsing them as prometheus targets.

Currently, we can use `discovery.file` to discover different files in our filesystem using double-star patterns. but these files are not read by the agent and are simply passed on as targets with `__path__` labels.
On the other hand, we have the `local.file` component which does read a files content and can be used in combination with the `discovery_target_decode` function to generate scrape targets. unfortunately, this component support passing in only a single path and cannot be combined with the `discovery.file` component to read multiple files.

This means in flow, we currently lack a capability which exists in prometheus natively in the `file_sd_configs` which support mapping files using double-star patterns and parsing target configs from their contents.

would be great to get this capability in the agent in flow mode.

spartan0x117:
#### PR Description
Adds the new `discovery.file` component, which implements Prometheus file-based SD.

Note that there is a banner added to the new `discovery.file` reference docs. Let me know if the wording should be changed or it should be moved elsewhere.

Closes #3346

#### PR Checklist

- [X] CHANGELOG updated
- [X] Documentation added
- [X] Tests updated


docs/sources/flow/reference/components/discovery.file.md
```diff
@@ -0,0 +1,163 @@
+---
+title: discovery.file
+---
+
+# discovery.file
+
+> **NOTE:** In `v0.35.0` of the Grafana Agent, the `discovery.file` component was renamed to [local.file_match][]. From `v0.35.0` onwards, the `discovery.file` 
+> component behaves as documented here.
```
captncraig, line 8: Maybe specify here briefly what local.file_match does, and why they may want to go there instead.

captncraig (approved):
Code looks good to me. I'd say wait for @clayton-cornell to weigh in on docs.

clayton-cornell (approved):
LGTM

PR #4411, opened 2023-07-12
flow: always allow running with a partially evaluated graph

rfratto:
This commit changes two behaviors about the Flow system:

* Components can now opt-in to being run even if their initial evaluation fails by returning a component.Component instance alongside their evaluation error.

  Previously, if the initial evaluation of a component failed, it would not be run until a succesful evaluation of the component occurred.

  This change allows polling-based components (such as `remote.http`) to rely on their native polling mechanisms to auto-resolve health issues if their initial evaluation fails.

  Without this change, a `remote.http` component with no dependencies that fails on the first evaluate would not be re-evaluated again until the config file is reloaded.

* LoadFile will no longer require the first evaluation to complete without error before scheduling components to be run.

  Previously, no components would run until all components evaluated successfully at least once.

  This change allows operating the Flow controller in a partially applied state, where a subset of components are healthy. This is critical in scenarios where a managed agent is being run where a subset of components may not always be immediately healthy on startup.

Implementing the second behavior implies that the first behavior should also be implemented, since users may not understand when it is required to reload the config file to fix polling-based components. 

Callers may emulate the previous behavior by not calling `Run` until `LoadFile` succeeds at least once. For the scope of this PR, this is what `grafana-agent run` has always done, and it hasn't changed here.

Supersedes #4395.


pkg/flow/flow.go
```diff
@@ -259,14 +277,14 @@ func (f *Flow) Run(ctx context.Context) {
...
-func (f *Flow) LoadFile(file *File, args map[string]any) error {
+// If strict is true, the controller will only start running components after
+// Load is called once without any configuration errors.
+func (f *Flow) LoadFile(file *File, args map[string]any, strict bool) error {
 	f.loadMut.Lock()
 	defer f.loadMut.Unlock()
 
 	diags := f.loader.Apply(args, file.Components, file.ConfigBlocks)
-	if !f.loadedOnce.Load() && diags.HasErrors() {
+	if strict && !f.loadedOnce.Load() && diags.HasErrors() {
```
thampiotr, line 287: The issue here I think is that we allow all the components to use the `strict=false` mode, while it only makes sense for some components.

Could we change the component only - to not return any errors when it encounters a recoverable error and keep retrying inside its `Run`? To not hide the problem, it could report it via `CurrentHealth() Health` but continue running.

rfratto: > Could we change the component only - to not return any errors when it encounters a recoverable error and keep retrying inside its Run? To not hide the problem, it could report it via CurrentHealth() Health but continue running.

I'm firmly against this, because not returning any errors means that you will have downstream evaluation errors for any component which depends on the failed component, even though there is only one actually invalid component.

thampiotr: > not returning any errors means that you will have downstream evaluation errors for any component which depends on the failed component, even though there is only one actually invalid component.

If this is combined with making the component that originated the error unhealthy via `CurrentHealth`, then one may be able to see in the UI where the issues start by following the dependency arrows... I get your concern though, having them marked as "unknown" is nicer.

> Can you explain what problem this causes? 

When I set `strict=false` then even unrecoverable errors will allow the graph evaluation to continue, while there may be no way for the component to recover - it may be completely misconfigured or missing some dependency that is not polled. Not every component can recover, in fact most components cannot recover, so I thought it may be too permissive. 

> What does "allow all components to use the strict=false mode" mean for end users?
What I mean is that the setting `strict=false` applies globally. While some components poll and may be able to recover, most components cannot recover.

I am not strongly against the solution in this prototype, but I can see some merits of doing it in the component.

rfratto:
> When I set strict=false then even unrecoverable errors will allow the graph evaluation to continue, while there may be no way for the component to recover - it may be completely misconfigured or missing some dependency that is not polled. Not every component can recover, in fact most components cannot recover, so I thought it may be too permissive.

Based on this, I think there is a misunderstanding about what this PR does. This PR changes two things:

* It is now possible to run components which evaluated successfully, while leaving failed components in the "unhealthy" or "unknown" state. Previously, no components could run until there was at least one successful graph load. 
* Components can now opt-in to being run even if their evaluation fails by returning a component instance alongside the error. Previously, if an evaluation fails, that component would not run until the next time LoadFile is called. 

Prior to this change, a component failing to evaluate did not halt the entire graph evaluation; it only halted the evaluation of downstream components. That doesn't change here. What does change is the ability to allow succeeded components to run without a completely successful load first. 

The second behavior is primarily useful for components which do polling. If a polling-based component fails on the initial load, the component author can determine whether it makes sense to still return a component instance to allow it to run. This enables a pseudo "recovery" mechanism for polling based components, but the controller has no concept of recovery; just the concept of running components and evaluating the graph.

pkg/flow/flow.go
thampiotr: thanks for the reply, it makes more sense now

rfratto, line 287: I don't adding an argument to LoadFile is required; the caller should decide whether or not an error from LoadFile is fatal before running the controller. 

The rest of this code is still necessary though.

rfratto:
@thampiotr PTAL, I updated the PR to remove the "non-strict mode" terminology in favor of just changing the default semantics, with instructions for how callers can emulate the previous behavior.

I've also updated the commit message / PR description to try to make it more clear exactly what behaviors this changes, including when/why it's useful.

rfratto:
I'm taking this out of draft since this is essentially ready for final review and a decision about the direction.

mattdurham:
Do we want/need to add documentation about this behavior?

rfratto:
> Do we want/need to add documentation about this behavior?

I don't think so at this current stage, since nothing changes (yet) from the user's perspective. This only unlocks the capability for Flow to behave correctly if we stopped asserting that the first load is successful.

pkg/flow/internal/controller/node_component.go
```diff
@@ -284,12 +284,14 @@ func (cn *ComponentNode) evaluate(scope *vm.Scope) error {
...
-		if err != nil {
-			return fmt.Errorf("building component: %w", err)
-		}
 		cn.managed = managed
 		cn.args = argsCopyValue
 
+		// We do the error check at the end to allow the component to still return
+		// a component instance with an error to signal that it should still run.
+		if err != nil {
+			return fmt.Errorf("building component: %w", err)
```
erikbaranowski, line 293: what do you think about logging this error in addition to returning it?

I am coming at this from the perspective of someone debugging the agent. We should provide as much telemetry data as possible for failed nodes. This is a small tradeoff for giving the extra flexibility in behavior (not stopping the agent on a single bad node)

jcreixell:
@rfratto what's missing to move forward with this? the solution seems reasonable to me, I will test it locally to make sure it behaves as expected in the context of agent management but other than that LGTM

jcreixell:
just tested this with `remote.http` and `module.string` and getting the following nil pointer error:

```
Error: module.string.agent_management:1:1: Unrecognized component name "local.files"

Error: module.string.agent_management:10:18: component "local.file.metrics_writer.content" does not exist
ts=2023-08-28T10:24:12.51244216Z level=info msg="scheduling loaded components and services"
interrupt received
ts=2023-08-28T10:24:12.512466596Z level=debug msg="flow controller exiting"
ts=2023-08-28T10:24:12.512467964Z level=error msg="failed to start reporter" err="context canceled"
ts=2023-08-28T10:24:12.512483342Z level=info component=remote.http.remoteconfig msg="component exited"
ts=2023-08-28T10:24:12.512738041Z level=info service=http msg="now listening for http traffic" addr=127.0.0.1:12345
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x3ed7dbe]

goroutine 259 [running]:
github.com/grafana/agent/component/module/string.(*Component).Run(0xc002732280?, {0x7f256c8?, 0xc0027090e0?})
	/home/jorge/workspace/src/github.com/rfratto/agent/component/module/string/string.go:63 +0x1e
github.com/grafana/agent/pkg/flow/internal/controller.(*ComponentNode).Run(0xc002732280, {0x7f256c8, 0xc0027090e0})
	/home/jorge/workspace/src/github.com/rfratto/agent/pkg/flow/internal/controller/node_component.go:330 +0x122
github.com/grafana/agent/pkg/flow/internal/controller.newTask.func1()
	/home/jorge/workspace/src/github.com/rfratto/agent/pkg/flow/internal/controller/scheduler.go:140 +0x74
created by github.com/grafana/agent/pkg/flow/internal/controller.newTask
	/home/jorge/workspace/src/github.com/rfratto/agent/pkg/flow/internal/controller/scheduler.go:137 +0x150
```

local agent  config:

```
remote.http "remoteconfig" {
  url = "some_url"
  headers = {
    "Authorization" = "Basic <some_key>",
  }
  poll_frequency = "11s"
}

module.string "agent_management" {
  content = remote.http.remoteconfig.content
}
```

remote config:

```
  // intentional typo in component name
  local.files "metrics_writer" {
    filename = "<some+path>"
  }

  prometheus.remote_write "grafanacloud" {
    endpoint {
      url = "<some_push_url>"
      basic_auth {
        username = "<some_id>"
        password = local.file.metrics_writer.content
      }
   }
  }

```

rfratto:
> @rfratto what's missing to move forward with this? the solution seems reasonable to me, I will test it locally to make sure it behaves as expected in the context of agent management but other than that LGTM

We're still missing general consensus on whether this is a behavioral change we want to make to the Flow controller.

thampiotr:
Seems like this is not needed at the moment / not blocking anything. Should we close this PR?

rfratto:
Yeah, let's close for now. We can revisit if/when needed

PR #4388, opened 2023-07-10
river: Add various string functions to stdlib

jkroepke:
#### PR Description

Add various string functions to stdlib. I don't need them all right now, however have some basic string functions to river might be a good idea.

#### Notes to the Reviewer

docs are heavily inspired by [Terraform](https://developer.hashicorp.com/terraform/language/functions/lower). I hope thats fine.

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated

PR #4414, opened 2023-07-12
flow: remove discovery_target_decode function from the standard library

### LINKED ISSUE #3607: Flow: replace discovery_target_decode with components

### LINKED ISSUE #4305: discovery_target_decode is considered a component rather than standard lib function
### What's wrong?

The following error is raised when using teh std lib function `discovery_target_decode`:

`Error: /etc/agent-config/agent.river:27:14: component "discovery_target_decode" does not exist `

It seems that discovery_target_decode is considered a component rather a standard lib function. I don't have similar issues with the rest of the functions from the std lib.
I'm running version v0.34.3 of the agent in docker.

### Steps to reproduce

The bug can be reproduced by running the example code found in the documentation: https://grafana.com/docs/agent/latest/flow/reference/stdlib/discovery_target_decode/

```
local.file "example" {
  filename = env("TARGETS_FILE")
}

prometheus.scrape "default" {
  targets = discovery_target_decode(local.file.example.content)
}
```

### System information

Docker

### Software version

v0.34.3

### Configuration

_No response_

### Logs

_No response_

rfratto:
Depends on #4404. 

Closes #3607
Closes #4305


pkg/flow/internal/stdlib/stdlib.go
```diff
@@ -1,46 +0,0 @@
...
-	"github.com/prometheus/prometheus/discovery/targetgroup"
-)
-
-// Identifiers holds a list of stdlib identifiers by name. All interface{}
-// values are River-compatible values.
-//
-// Function identifiers are Go functions with exactly one non-error return
-// value, with an optionally supported error return value as the second return
-// value.
-var Identifiers = map[string]interface{}{
```
captncraig, line 18: Should we keep this file around for future usages. I know https://github.com/grafana/agent/pull/4388/files uses it for one thing, unless this is the wrong place to add those. I still think its possible there are flow-specific stdlib functions possible.

rfratto: #4388 is extending pkg/river/internal/stdlib, but this is removing pkg/flow/internal/stdlib. We consider it an anti-pattern now to have Flow-specific stdlib functions, which is why this file is being removed in favor of only having the River one.

captncraig (approved):

mattdurham (approved):
LGTM

clayton-cornell (approved):
LGTM

PR #2654, opened 2022-12-27
Rework Grafana Agent Operator docs

karengermond:
#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

PR #4415, opened 2023-07-12
Remove oauth2 agent config from example in Operator docs

clayton-cornell:
#### PR Description

Oauth2 is mentioned as supported in Operator `remotewrite`. This is incorrect and needs to be removed from the documentation. Oauth2 is supported in static and flow mode `remote_write`, but not in operator.

Reference https://github.com/grafana/agent/pull/2654 that added the information

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated


rfratto (approved):
Thank you!

PR #4425, opened 2023-07-13
agent-flow-mixin: do not fire ClusterLamportClockStuck for single-node clusters

rfratto:
Single-node clusters have no external peers to communicate with, so their lamport times won't increase.


mattdurham (approved):
LGTM

PR #4426, opened 2023-07-13
[Docs] Fix typo in Static mode Docker setup guide

spartan0x117:


clayton-cornell (approved):
Oooops, I need to replace my keyboard. It's starting to double type letters. Thanks for catching this!

PR #4429, opened 2023-07-13
Clarify otelcol.exporter.jaeger deprecation

ptodev:
I added this deprecation recently, but didn't know it needs to be under a "deprecation" section and listed in the upgrade guide.

The exporter is due to be deleted from the Collector in July 2023, so I think v0.38.0 is ok as a target version for removal.


rfratto (approved):
Thanks!

PR #4430, opened 2023-07-13
Draft: repro and fix encoding issues

thampiotr:
#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/river/token/builder/builder_test.go
+++ b/pkg/river/token/builder/builder_test.go
@@ -551,29 +551,18 @@ func TestBuilder_AppendInner_MatchesDefaults(t *testing.T) {
 	expect := format(t, `
 		field = "some_value"
 		
-		block {
-			number = 123
-		}
+		block { }
 		
-		block_slice {
-			number = 123
-		}
+		block_slice { }
 		
-		block_slice {
-			number = 123
-		}
+		block_slice { }
+		
+		block_ptr { }
+		
+		block_ptr_slice { }
+		
+		block_ptr_slice { }
 
-		block_ptr {
-			number = 123
-		}
-		
-		block_ptr_slice {
-			number = 123
-		}
-		
-		block_ptr_slice {
-			number = 123
-		}
 	`)
 
 	require.Equal(t, expect, string(f.Bytes()))
--- a/pkg/river/token/builder/builder_test.go
+++ b/pkg/river/token/builder/builder_test.go
@@ -537,14 +537,14 @@ func TestBuilder_AppendInner_MatchesDefaults(t *testing.T) {
 	f.Body().AppendFrom(Structure{
 		Field: "some_value",
 
-		Block: InnerBlockDefaulter{Number: 123},
+		Block: InnerDefaults,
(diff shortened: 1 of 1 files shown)


pkg/river/token/builder/builder_test.go
```diff
@@ -409,3 +409,266 @@ func TestBuilder_MapBlocks(t *testing.T) {
...
+		BlocksSlice: []InnerBlockDefaulter{},
+
+		BlockPtr:       &InnerBlockDefaulter{},
+		BlocksPtrSlice: []*InnerBlockDefaulter{},
+	})
+
+	expect := format(t, `
+		field = "some_value"
+
+		block_ptr { }
```
thampiotr, line 594: Why are all the values skipped? they all should have `number = 0`. Looks like a bug?

pkg/river/token/builder/builder_test.go
```diff
@@ -409,3 +409,266 @@ func TestBuilder_MapBlocks(t *testing.T) {
...
+
+		Block:       InnerBlockDefaulter{},
+		BlocksSlice: nil,
+
+		BlockPtr:       nil,
+		BlocksPtrSlice: nil,
+	})
+
+	expect := format(t, `
+		field = "some_value"
```
thampiotr, line 621: The `block` differs from its default, but not encoded here. Bug?

thampiotr (commented):
The tests I've added currently pass, but I am not sure this is the behaviour we want. See the comments below.

pkg/river/token/builder/builder_test.go
```diff
@@ -409,3 +409,346 @@ func TestBuilder_MapBlocks(t *testing.T) {
...
+
+	expect := format(t, `
+		inner = {
+			number = 123,
+		}
+		inner_slice = []
+		inner_ptr   = {
+			number = 123,
+		}
+		inner_ptr_slice = []
```
thampiotr, line 704: This test's input data is virtually the same as the test above (`TestBuilder_AppendInnerBlock_InnerDefaultsMatch_OuterDefaultsNil`) with the only difference that instead of `block`, the `OuterAttrDefaulter` uses `attr` for its fields.

I would expect a result that is closer to how we handle a `block`, without printing default values for optional attributes?

pkg/river/token/builder/builder_test.go
thampiotr: Debugging this for the field named `block`:
image
We can see that it's determined to match the default:
image

But that's the default of the `Structure`, not the `InnerBlockDefaulter`. The correct default for the `InnerBlockDefaulter` has `123` in it.

I think that's the issue. For a given field in an inner block we have two possible default values:
* One coming from the outer block's `SetToDefault`
* Another coming from the fields' own `SetToDefault`

They can give conflicting defaults. We also need to be careful with how we treat `nil` returned in defaults.

thampiotr: I've drafted a fix here, but this may not be the best way to tackle it: https://github.com/grafana/agent/pull/4435/files

PR #4431, opened 2023-07-13
Test github action for trivy scanning

captncraig:
This is based off of the github generated action under the security tab for "code scanning".

The idea is that on push to main, it will build the docker image, generate a report, and upload to the security tab. 

Hoping this is a way to centralize the vulnerabilities we know about, and hopefully be a transparent way to communicate status of individual cves.


captncraig:
I'll test this in my fork for now.

PR #4442, opened 2023-07-13
No longer needed since we removed discussions?

mattdurham:
We dont have discussions section anymore


rfratto (commented):
We still have the discussions page and haven't migrated off of it yet, we're waiting to hear if Grafana Agent will have a dedicated section on community.grafana.com first.

PR #4443, opened 2023-07-13
Decode / encode tests

thampiotr:
#### PR Description
This PR adds tests that cover different cases of nested blocks with defaults etc.
* Encodes given block to river
* Verifies river matches the expectations
* Decodes river back to a struct
* Compares the result with original

Furthermore, this PR:
* Fixes one small issue in encoding logic
* Adds docs with details how Defaulter works
* Refactors a couple of Default-related fragments of code (after reviewing all the Defaulters to make sure they agree on defaults between outer and inner blocks)

#### PR Checklist

- [ ] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated


pkg/river/token/builder/builder_defaults_test.go
```diff
@@ -0,0 +1,136 @@
...
+		// NOTE: this is correct, because inner block's defaults will be applied
+		river: `
+			inner { }
+		`,
+	},
+	{
+		name:  "outer struct block / matching defaults / no value set",
+		in:    OuterStructBlockWithMatchingDefault{},
+		river: `inner { }`,
+	},
```
thampiotr, line 81: This test appears to fail, because the value encoded to River does not decode into the same original value:

```
 builder_defaults_test.go:120: 
        	Error Trace:	/Users/piotr/workspace/agent/pkg/river/token/builder/builder_defaults_test.go:120
        	Error:      	Not equal: 
        	            	expected: builder_test.OuterStructBlockWithMatchingDefault{Inner:builder_test.AttrWithDefault{Number:0}}
        	            	actual  : builder_test.OuterStructBlockWithMatchingDefault{Inner:builder_test.AttrWithDefault{Number:123}}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -2,3 +2,3 @@
        	            	  Inner: (builder_test.AttrWithDefault) {
        	            	-  Number: (int) 0
        	            	+  Number: (int) 123
        	            	  }
        	Test:       	TestDefaults/outer_struct_block_/_matching_defaults_/_no_value_set
        	Messages:   	Encoding and decoding the same block gave different results
```

pkg/river/token/builder/builder.go
```diff
@@ -216,13 +216,6 @@ func (b *Body) encodeField(prefix []string, field rivertags.Field, fieldValue re
 		fullName := mergeStringSlice(prefix, field.Name)
 
 		switch {
-		case fieldValue.IsZero():
-			// It shouldn't be possible to have a required block which is unset, but
-			// we'll encode something anyway.
-			inner := NewBlock(fullName, "")
-			inner.body.SetValueOverrideHook(b.valueOverrideHook)
-			b.AppendBlock(inner)
-
```
thampiotr, line 225: Here a zero struct is encoded as empty block in River. When it's decoded, the defaults will be applied, so we wouldn't get the same value as we encoded. 

Luckily all we need to do is remove this case and the other cases will handle it correctly, taking defaults into account.

pkg/river/token/builder/blocks_nesting_test.go
```diff
@@ -0,0 +1,693 @@
...
+		// Seems impossible to encode this case in River, because we would need to somehow explicitly set the inner block
+		// to nil? How can we do that?
+
+		//{
+		//	name: "nil",
+		//	in: OutZeroDefInPtrBlkOptWithDef{
+		//		Inner: nil,
+		//	},
+		//	river: ``,
+		//},
```
thampiotr, line 261: This is another outstanding issue. I'm not sure how we can fix it. 
We could return an error in decode when a block field with defaults is set to nil.
Or we could disallow pointers to have defaults - which would be a bigger change.

pkg/river/token/builder/blocks_nesting_test.go
```diff
@@ -0,0 +1,693 @@
...
+	return []testCase{
+		//{
+		//	//TODO: encoded river doesn't decode correctly due to error: missing required block "inner"
+		//	// seems like we should have failed on encoding to river?
+		//	name: "nil",
+		//	in: OutZeroDefInPtrBlkWithDef{
+		//		Inner: nil,
+		//	},
+		//	river: "",
+		//},
```
thampiotr, line 459: Another issue: we encode happily to River, but decoding errors with "missing required block". 
Seems like we should error on encoding too?

pkg/river/internal/value/decode.go
```diff
@@ -13,6 +13,24 @@ import (
...
+//   - when an inner block is not provided at all, the outer type's defaults
+//     will be used.
+//   - if the defaults are not defined, zero values (for structs) or
+//     nil (for pointers) will be set for fields that are not present.
+//     This means that, for example, if an outer type doesn't define defaults,
+//     but the inner type does, the inner type's defaults will not be used when
+//     the block is not provided at all.
+//
+// To avoid confusion, it is strongly recommended to ensure that the
+// SetToDefault method sets the same defaults for the outer and the inner types. .
```
rfratto, line 33: I think this is a bit long. What do you think about this?

```suggestion
// Defaulter is only used for block and body River types.
//
// If a River block or body wraps a type implementing Defaulter, the 
// wrapping type must also implement Defaulter to propagate the default 
// value of of the wrapped type, otherwise defaults for the wrapped type 
// only apply when the wrapped type is defined in the River config file. 
```

pkg/river/token/builder/blocks_nesting_test.go
```diff
@@ -0,0 +1,694 @@
+package builder_test
```
rfratto, line 1: Unfortunately, I find this file fairly hard to read:

* The names for identifiers are quite long and hard to parse, such as `OutMatchDefInPtrBlkOptMatchDef`. 
* I find the test harness code here hard to grok beacuse of the coupling between test types (i.e., types that will be used for encoding/decoding in tests) and the tests to perform (i.e., the implementation of the `testFactory` method). I would prefer a giant test table in a single test function over what is being done now. 

Can we find a way to condense this file down to make it easier to scan and identify what's being tested? Also, if possible, can we try to build on top of tests to get test coverage rather than trying to test every combination of scenarios? (For example, I am not sure it's necessary for all the pointer tests to exist, since they don't behave any differently than non-pointer examples, minus what the definition of the "zero value" is).

rfratto (commented):
Thanks Piotr!

pkg/river/token/builder/blocks_nesting_test.go
thampiotr: Yeah agreed, it's been hard to name these tests and keep them easy to read, because structs need to have methods so they are declared outside of the test function. I've attempted to at least keep the test cases next to the structs definitions.

What do you propose I call a test case structure that tests `outer with zero value default, inner struct block, optional with a default value`? I could just call them `TestStructN` and keep more descriptive test names. I'm not sure it is going to help though. 

I do think we should have more edge case tests and I wanted to cover a large number of cases, because I found it's easy to get surprising behaviour and pretty time consuming to figure it out in the code.

thampiotr: > Also, if possible, can we try to build on top of tests to get test coverage rather than trying to test every combination of scenarios?

I wouldn't rely on coverage too much here, because we use reflection a lot and for example executing an if statement on `v.IsZero()` using only one type doesn't seem sufficient given the number of allowed inputs. The test coverage was pretty good and we did find a few issues already.

rfratto: > What do you propose I call a test case structure that tests outer with zero value default, inner struct block, optional with a default value? I could just call them TestStructN and keep more descriptive test names. I'm not sure it is going to help though.

If I understand the scenario you just laid out: NonPropagatingBlock? Because it doesn't propagate the defaults from the inner struct.

pkg/river/internal/value/decode.go
thampiotr: I personally find it harder to understand it in this compressed form. Also I would like to clearly state that if we have e.g. `inner {}` in River then the inner type's defaults will be used, but when we have nothing `""` then the outer type's defaults will be used.

> Defaulter is only used for block and body River types.
You can use Defaulter anywhere, but it won't be invoked unless it's on a block or body. But also, the defaulter can be used to set the arguments inside the block and body - so in a way it can be used 'with' arguments, but you shouldn't make your argument types implement it. 

> the wrapping type must also implement Defaulter to propagate the default
It does get funny with pointers though - I think it's fine if the outer type sets the inner field to `nil`, but it can be added with its defaults (from the inner type) if one provides `inner { }`. So I wouldn't say that the wrapping type must implement Defaulter. If it was a requirement, we probably should check that it holds and error when it's violated?

rfratto: > Also I would like to clearly state that if we have e.g. inner {} in River then the inner type's defaults will be used, but when we have nothing "" then the outer type's defaults will be used.

I tried to say that here: 

> defaults for the wrapped type only apply when the wrapped type is defined in the River config file. 

Is there a clearer way of describing this succinctly?

thampiotr: How about something like this?

```
// Defaulter will be called only on block and body river types.
//
// When using nested blocks, the wrapping type should also implement the 
// Defaulter to propagate the defaults of the wrapped type. Otherwise, the 
// wrapping type's defaults will be used when the wrapped field is not defined
// in the River config, or the wrapped type's defaults will be used if the 
// wrapped field is defined in the River config.
```
Heavily inspired by your suggestion, but slightly different.

rfratto: I think we should make this a "must" rather than a "should" w/r/t propagating defaults. 

I'd like to tweak the "Otherwise" part a bit to make it rotate around what happens when the wrapped block is defined compared to when defaults are used:

```
// Defaulter will be called only on block and body river types.
//
// When using nested blocks, the wrapping type must also implement the 
// Defaulter to propagate the defaults of the wrapped type. Otherwise, 
// defaults used for the wrapped type become inconsistent: 
//
// - If the the wrapped block is NOT defined in the River config, the wrapping 
//   type's defaults are used. 
// - If the wrapped block IS defined in the River config, the wrapped type's 
//   defaults are used. 
```

WDYT? I reintroduced the bullet list to hopefully make it easier to scan as the paragraph got longer while still keeping the overall additions to the comment shorter than before.

pkg/river/token/builder/blocks_nesting_test.go
thampiotr: I have re-written tests and now I believe they are more readable. I still opted to cover some `struct` vs. `pointer` cases separately even though they may hit the same code path - otherwise it feels it's too white-box approach and I'm overfitting the tests to implementation.

pkg/river/token/builder/nested_defaults_test.go
```diff
@@ -0,0 +1,229 @@
...
+		input:         PtrPropagatingDefault{Inner: &AttrWithDefault{Number: defaultNumber}},
+		expectedRiver: "",
+	},
+	// TODO: Test fails currently: invariant violated - not clear how to explicitly set the inner block to nil in River
+	//       The builder currently encodes this into empty river block, which decodes into default value.
+	//{
+	//	name:          "pointer propagating default - nil pointer",
+	//	input:         PtrPropagatingDefault{},
+	//	expectedRiver: "",
+	//},
```
thampiotr, line 59: In other words, the issue here is:
* An empty string decodes into `{Inner: &{Number: 123}}`
* But `{Inner: nil}` encodes into empty string

Not sure how we can fix it, since it's not clear how to explicitly set a block to nil in River. Maybe trying to encode a nil block should be an error?

rfratto: This is tricky because I think this is correct, but fails just because of how the tests work right now:

1. `PtrPropagatingDefault{}` _should_ encode to the empty string because the block isn't provided.
2. When decoding, defaults are applied to `PtrPropagatingDefault{}`, which sets `Inner` to `&{Number: 123}`. 
3. This means that the developer is claiming the `Inner: nil` and `Inner: &{Number: 123}` are semantically equivalent from the user's perspective; if the user provides an empty string or `inner { number = 123 }`, they decode the same way. 

I'm not sure how this should be addressed from the test scenario.

rfratto (approved):
LGTM. I'm considering whether we should merge this as-is and deal with the commented out tests in a follow up discussion. WDYT?

pkg/river/token/builder/blocks_nesting_test.go
rfratto: The tests are _much_ better now, thank you!

PR #4199, opened 2023-06-17
operator: allow customization of config-reloader image

alekseybb197:
#### PR Description

Add deployment specification keys for describe Prometheus Config Reloader image.

#### Notes to the Reviewer

All code lines made as code for Grafana Agent abow.

#### PR Checklist

- [+] CHANGELOG updated
- [+] Documentation added
- [+] Tests updated

PR #4447, opened 2023-07-13
Operator: Regenerate crds

captncraig:
It looks like after #4199, the crd definitions and documentation never got regenerated. 

Also adds a CI check for that as well.


jcreixell (approved):

marctc:
> Also adds a CI check for that as well.

Awesome!

marctc (approved):

PR #4450, opened 2023-07-14
Update `make docs` procedure and ensure that it is regularly updated in future.

jdbaldry:
See individual commits.

I'll need an admin to add @grafana/docs-tooling to the repository if you would like me to be able to promptly review and merge the daily automated update PRs.

@ptodev, this also updates the `make docs` behavior to use the current repository as the source of the content and not use the repository inferred from the project name.

### DIFF
--- a/.github/workflows/update-make-docs.yml
+++ b/.github/workflows/update-make-docs.yml
@@ -1,0 +1,27 @@
+name: Update `make docs` procedure
+on:
+  schedule:
+    - cron: '0 7 * * 1-5'
+jobs:
+  main:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Update procedure
+        if: github.repository != 'grafana/writers-toolkit'
+        run: |
+          curl -s -Lo docs/docs.mk https://raw.githubusercontent.com/grafana/writers-toolkit/main/docs/docs.mk
+          curl -s -Lo docs/make-docs https://raw.githubusercontent.com/grafana/writers-toolkit/main/docs/make-docs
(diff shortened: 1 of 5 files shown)


ptodev (approved):
Thank you, this is great! I'm very happy with the change to default to the current repo, and also with the codeowners change.

Btw the docs scripts and the automatic updater seem like a lot of very customized tooling. Maybe it would reduce the docs team's TCO if the tooling is written in Go, and if we use Dependabot for updating it across repos? Then other teams would also have a better chance of contributing, since most people would be more skilled at Go than Bash 😄   This is just a personal observation, but obviously I'm happy for the PR to be merged and everything to stay as it is.

jdbaldry:
> Thank you, this is great! I'm very happy with the change to default to the current repo, and also with the codeowners change.
> 
> Btw the docs scripts and the automatic updater seem like a lot of very customized tooling. Maybe it would reduce the docs team's TCO if the tooling is written in Go, and if we use Dependabot for updating it across repos? Then other teams would also have a better chance of contributing, since most people would be more skilled at Go than Bash 😄 This is just a personal observation, but obviously I'm happy for the PR to be merged and everything to stay as it is.

Thanks for sharing the idea, I think less bespoke tooling would be really valuable. My two priorities are automatic updates and portability. Its actually written in POSIX sh at the moment for that reason. One consideration to rewriting the script in Go would be that contributors would be required to install Go to run the local webserver. At the moment, there is a dependency on a container runtime, so perhaps we could use nested containers to avoid the need for user Go installation. Definitely something to think about in the future as the shell script becomes more complicated.

erikbaranowski (approved):
LGTM, I tested `make docs` and it is confirmed working on this branch

PR #4452, opened 2023-07-14
remote.http: set health after poll on Update

### LINKED ISSUE #4451: remote.http: Component remains in status Unknown unless first background poll
### What's wrong?

Any remote.http component remains in status unknown unless first background poll.

For components which have 24h poll frequency configured, it takes one day to get greeen.

image

### Steps to reproduce

Setup an http.remote components and start the agent

### System information

MacOS 13.4

### Software version

v0.33.3

### Configuration

_No response_

### Logs

_No response_

rfratto:
Previously, the health of the component was only set on background polls, which caused the health of the component to be listed as "Unknown" until a background poll occured.

In cases where the polling interval is extremely infrequent, the component appeared to be permanently in state "Unknown."

Closes #4451.

### DIFF
--- a/component/remote/http/http.go
+++ b/component/remote/http/http.go
@@ -160,11 +160,11 @@ func (c *Component) nextPoll() time.Duration {
 // not be held when calling. After polling, the component's health is updated
 // with the success or failure status.
 func (c *Component) poll() {
-	startTime := time.Now()
 	err := c.pollError()
+	c.updatePollHealth(err)
+}
 
-	// NOTE(rfratto): to prevent the health from being inaccessible for longer
-	// than is needed, only update the health after the poll finished.
+func (c *Component) updatePollHealth(err error) {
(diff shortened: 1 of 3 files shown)


component/remote/http/http.go
```diff
@@ -160,25 +160,25 @@ func (c *Component) nextPoll() time.Duration {
 // not be held when calling. After polling, the component's health is updated
 // with the success or failure status.
 func (c *Component) poll() {
-	startTime := time.Now()
```
rfratto, line 163: I don't think it makes sense for the health time to be set when the poll _started_, rather than when it _finished._ I'm not sure why I did this initially, but I think it makes more sense with the update time being set to when the health was generated.

jkroepke:
Hi, would this PR part of the next 0.35.0?

mattdurham:
Yes,  l am going to cut another rc today

PR #4463, opened 2023-07-14
Converter test library

erikbaranowski:
#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [x] Tests updated


converter/internal/test_common/testing.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/grafana/agent/converter/diag"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	flowSuffix  = ".river"
+	diagsSuffix = ".diags"
+)
+
+// TestDirectory will execute tests for converting from a source configuration
```
erikbaranowski, line 22: This felt warranted due to how much is wrapped up in this test, but if it feels redundant now that the function itself is much more readable I can nuke it.

converter/internal/test_common/testing.go
```diff
@@ -0,0 +1,138 @@
...
+		return normalizeLineEndings(outputBytes)
+	}
+
+	return nil
+}
+
+// validateRiver makes sure the expected river and actual river are a match
+func validateRiver(t *testing.T, expectedRiver []byte, actualRiver []byte) {
+	if len(expectedRiver) > 0 {
+		if !reflect.DeepEqual(expectedRiver, actualRiver) {
```
erikbaranowski, line 130: @thampiotr I added this extra condition before printing because 1 failure was printing out all expected river output and it was confusing to parse which went wrong.

erikbaranowski: I also do like having this print in this case a LOT 🎉 . I'm considering if we should do similar smarter printing for the diags testing as well, but that can come later too if we decide we want it.

PR #4473, opened 2023-07-17
Add canonical URL to all pages

jdbaldry:
Grafana Cloud documentation wants to reuse this documentation as part of its reorganization.

Having duplicate pages can be detrimental to SEO and setting the canonical URL helps by letting search engines only index the canonical page.

We do not index /docs/agent/next/ pages, only /docs/agent/latest/, so there is no need to concern ourselves with how this canonical is managed across versions. Pages that are not "latest", automatically have their canonical set to "latest" by the website.

I will open a similar PR in the latest version branch to update the current "latest" pages.

### DIFF
--- a/docs/sources/_index.md
+++ b/docs/sources/_index.md
@@ -1,4 +1,5 @@
 ---
+canonical: https://grafana.com/docs/agent/latest/
 title: Grafana Agent
 weight: 1
 ---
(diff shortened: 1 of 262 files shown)


clayton-cornell (approved):
LGTM

PR #4474, opened 2023-07-17
fix(helm): use release namespace instead of default

korniltsev:
#### PR Description
I've noticed helm chart did not specify namespace and I wanted to deploy ebpf profiler in  a separate namespace, so it looks like these should be added.
#### PR Checklist

- [x] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated


korniltsev:
Hmmm, maybe I should not do it

korniltsev:
https://github.com/helm/helm/issues/5465

PR #4482, opened 2023-07-17
Add trivy workflow

captncraig:
This will automatically run a trivy scan on each push to main. It will add results to the Security tab.

Related to #4439


mattdurham (approved):
Lgtm

PR #4497, opened 2023-07-18
Update Grafana Agent to use v1 CronJob API.

mplzik:
#### PR Description

The `v1beta1` CronJob API has been removed in Kubernetes 1.25 and its
stable replacement in `v1` exists since Kubernetes 1.21, which is not
supported for a fairly long time anymore.

Signed-off-by: Milan Plzik <milan.plzik@grafana.com>

#### PR Checklist

- [x] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ Main (unreleased)
 ### Enhancements
 
 - Add [godeltaprof](https://github.com/grafana/godeltaprof) profiling types (`godeltaprof_memory`, `godeltaprof_mutex`, `godeltaprof_block`) to `pyroscope.scrape` component
-- Update jsonnet libraries to use the `v1` CronJob API instead of the `v1beta1` that is removed from Kubernetes 1.25. The `v1` CronJob has been around since Kubernetes 1.21, making it a safe change breakage-wise.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -13,7 +13,7 @@ Main (unreleased)
 ### Enhancements
 
 - Add [godeltaprof](https://github.com/grafana/godeltaprof) profiling types (`godeltaprof_memory`, `godeltaprof_mutex`, `godeltaprof_block`) to `pyroscope.scrape` component
-
+- Update jsonnet libraries to use the `v1` CronJob API instead of the `v1beta1` that is removed from Kubernetes 1.25. The `v1` CronJob has been around since Kubernetes 1.21, making it a safe change breakage-wise.
```
marctc, line 16: I don't we need changelog in this case as it not affecting users.

marctc (approved):

CHANGELOG.md
mplzik: Let me revert this change then. This was added as a precaution in case some of the users would run ancient Kubernetes versions.

PR #4506, opened 2023-07-18
Validating all currently unsupported static to flow conversion configurations

erikbaranowski:
#### PR Description

Load and validate all configuration for static to Flow. Much of the now unsupported configurations will become supported in future PRs as we add functionality.

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [x] Tests updated


converter/internal/common/validate.go
```diff
@@ -0,0 +1,26 @@
+package common
+
+import (
+	"fmt"
+	"reflect"
+
+	"github.com/grafana/agent/converter/diag"
+)
+
+func UnsupportedNotDeepEquals(a any, b any, name string) diag.Diagnostics {
```
erikbaranowski, line 10: This file and funcs might evolve in future PRs but this gets the ball rolling

converter/internal/staticconvert/staticconvert.go
```diff
@@ -2,35 +2,35 @@ package staticconvert
...
 	prom_config "github.com/prometheus/prometheus/config"
+
+	_ "github.com/grafana/agent/pkg/integrations/install" // Install integrations
 )
 
 // Convert implements a Static config converter.
 func Convert(in []byte) ([]byte, diag.Diagnostics) {
 	var diags diag.Diagnostics
 
-	var staticConfig config.Config
```
erikbaranowski, line 20: Changes to `Convert` were required for integrations to load into memory correctly.

converter/internal/common/validate.go
rfratto, line 10: It might be easier for consumers if this took diagnostics as a parameter, similar to `require.Equals` taking `*testing.T` as a parameter: 

```go
func UnsupportedNotDeepEquals(diags *diag.Diagnostics, a any, b any, name string)
```

converter/internal/staticconvert/validate.go
```diff
@@ -0,0 +1,123 @@
...
+	diags = append(diags, newDiags...)
+
+	newDiags = validateTraces(staticConfig.Traces)
+	diags = append(diags, newDiags...)
+
+	newDiags = validateLogs(staticConfig.Logs)
+	diags = append(diags, newDiags...)
+
+	newDiags = validateAgentManagement(staticConfig.AgentManagement)
+	diags = append(diags, newDiags...)
```
rfratto, line 35: I think the pattern here is a bit verbose. What do you think about having validate functions accept a pointer to diag.Diagnostics as an argument so they can mutate it? That way this would become:

```go
	_, grpcListenPort, _ := staticConfig.ServerFlags.GRPC.ListenHostPort()

	validateServer(&diags, staticConfig.Server)
	validateMetrics(&diags, staticConfig.Metrics, grpcListenPort)
	validateIntegrations(&diags, staticConfig.Integrations)
	validateTraces(&diags, staticConfig.Traces)
	validateLogs(&diags, staticConfig.Logs)
	validateAgentManagement(&diags, staticConfig.AgentManagement)
```

rfratto (approved):

converter/internal/staticconvert/validate.go
mattdurham: Not a fan of passing in the pointer.  To me passing in diags makes me have to think why am I passing in diags and what are the vehaviors of nil/empty. It feels more stateful. 

if we wanted less verbose we could inline the append.

`newDiags = append(diags,validateLogs(staticConfig.Logs)...)`

thampiotr: You can also `diags.AddAll(validateLogs(staticConfig.Logs))` which mutates your local variable, but doesn't send it to another function.

erikbaranowski: I discussed this same topic with @thampiotr yesterday in a similar conversation about how to implement the validation for converters: https://github.com/grafana/agent/pull/4500#discussion_r1268484303

Due to the variety of implementations and possible paths, we are going to break this out as a separate PR that focuses on just that.

converter/internal/common/validate.go
erikbaranowski: see below discussion about a planned validation discussion and refactor/tune up. https://github.com/grafana/agent/pull/4506#discussion_r1270674998

PR #4527, opened 2023-07-20
agent-flow-mixin: fix display of heatmaps

rfratto:
In newer versions of Grafana, heatmap panels default to calculating the buckets based on the query. However, the heatmaps we're using in the dashboard have precalculated buckets.

This adds a `calculate: false` option so that the heatmaps display as expected.

Before:

image

After:

image


mattdurham (approved):
LGTM

PR #4532, opened 2023-07-20
Converter static logs

erikbaranowski:
#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [x] Tests updated


converter/internal/staticconvert/testdata/promtail_prom.river
```diff
@@ -0,0 +1,130 @@
+discovery.consul "name_jobName" {
```
erikbaranowski, line 1: This component exists twice and the label is an exact match between prometheus and promtail. I want to call this out so we can decide how to move forward.

converter/internal/promtailconvert/promtailconvert.go
```diff
@@ -74,26 +74,27 @@ func Convert(in []byte) ([]byte, diag.Diagnostics) {
...
 	prettyByte, newDiags := common.PrettyPrint(buf.Bytes())
-	diags = append(diags, newDiags...)
+	diags.AddAll(newDiags)
 	return prettyByte, diags
 }
 
 // AppendAll analyzes the entire promtail config in memory and transforms it
 // into Flow components. It then appends each argument to the file builder.
-func AppendAll(f *builder.File, cfg *promtailcfg.Config, diags diag.Diagnostics) diag.Diagnostics {
+func AppendAll(f *builder.File, cfg *promtailcfg.Config, labelPrefix string, diags diag.Diagnostics) diag.Diagnostics {
```
erikbaranowski, line 83: I had to touch a decent amount of promtail to pass through the labelPrefix but also make the label incrementing consistent with prometheus and everywhere in promtail converter.

converter/internal/staticconvert/testdata/promtail_scrape.diags
```diff
@@ -0,0 +1,2 @@
+(Error) global positions configuration is not supported - each Flow Mode's loki.source.file component has its own positions file in the component's data directory
```
erikbaranowski, line 1: These concern me that every single static promtail config will have this error and warning currently. I'm not sure what we can do about it.

converter/internal/staticconvert/testdata/promtail_prom.river
erikbaranowski: Brainstorm:
- I think we may need to do something where when adding blocks we check for uniqueness right them on the builder.File. This would require exposing more of that API publicly than is exposed today.

converter/internal/common/common.go
```diff
@@ -59,6 +59,21 @@ func GetUniqueLabel(label string, currentCount int) string {
...
+	if currentCount != 0 {
+		finalLabel += fmt.Sprintf("_%d", currentCount+1)
+	}
+
+	if labelPrefix != "" {
+		finalLabel = labelPrefix + "_" + finalLabel
+	}
+
+	return finalLabel
+}
```
thampiotr, line 75: I was thinking maybe we do something like this: https://go.dev/play/p/SleUy3-VQEA?v=
LMK what you think. I wanted to:
* support more than two string components in the label
* avoid calling with hard-coded `currentCount` set to 0

converter/internal/promtailconvert/internal/build/cloudflare.go
```diff
@@ -33,9 +33,10 @@ func (s *ScrapeConfigBuilder) AppendCloudFlareConfig() {
 			return val
 		}
 	}
+	compLabel := common.GetLabelWithPrefix(s.globalCtx.LabelPrefix, s.cfg.JobName, 0)
```
thampiotr, line 36: Would be nice to avoid passing `0` in these.

thampiotr (commented):
Adding some early suggestions, will take a deeper look later.

converter/internal/common/common.go
erikbaranowski: I updated the implementation similar to what you suggested with a tweak for indexing and empty prefixes. The pattern is now 1 indexed without an index on the first label:

LABEL
LABEL_2
LABEL_3
etc

Some alternative patterns... I don't feel super strongly pro/against any honestly, but we can easily change it in the 1 place and update test outputs now:

LABEL
LABEL_1
LABEL_2
etc

OR

LABEL_0
LABEL_1
LABEL_2
etc

OR 

LABEL_1
LABEL_2
LABEL_3

converter/internal/promtailconvert/internal/build/file_sd.go
```diff
@@ -13,7 +11,7 @@ func (s *ScrapeConfigBuilder) AppendFileSDs() {
 	}
 	for i, sd := range s.cfg.ServiceDiscoveryConfig.FileSDConfigs {
 		args := prometheusconvert.ToDiscoveryFile(sd)
-		compLabel := fmt.Sprintf("file_sd_%d", i)
```
erikbaranowski, line 16: 🐛 🔨

converter/internal/staticconvert/testdata/promtail_prom.river
erikbaranowski: I added a todo on a test that demonstrates this

erikbaranowski: I appended `metrics_` or `logs_` to labels to prevent collision between the 2 and provide clearer output. There is 1 edge case left that is now documented in code. A detector on the final output might be a way to go to provide a diag if we detect a collision after completed conversion.

converter/internal/staticconvert/staticconvert.go
```diff
@@ -88,9 +90,54 @@ func AppendStaticPrometheus(f *builder.File, staticConfig *config.Config) diag.D
...
+			ServerConfig:    server.Config{Disable: true},
+			ClientConfig:    client.Config{},
+			ClientConfigs:   logConfig.ClientConfigs,
+			PositionsConfig: logConfig.PositionsConfig,
+			ScrapeConfig:    logConfig.ScrapeConfig,
+			TargetConfig:    logConfig.TargetConfig,
+			LimitsConfig:    logConfig.LimitsConfig,
+			Options:         promtail_config.Options{},
+			Tracing:         tracing.Config{Enabled: false},
+			WAL:             wal.Config{Enabled: false},
```
thampiotr, line 119: I see - for the unsupported features, we disable them (which is not always the same as default). 
Nit: Maybe a comment to clarify?

converter/internal/staticconvert/staticconvert.go
```diff
@@ -88,9 +90,54 @@ func AppendStaticPrometheus(f *builder.File, staticConfig *config.Config) diag.D
...
+		}
+
+		// We need to set this when empty so the promtail converter doesn't think it has been overridden
+		if promtailConfig.Global == (promtail_config.GlobalConfig{}) {
```
thampiotr, line 123:
```suggestion
		if promtailConfig.Global == promtail_config.GlobalConfig{} {
```
Do you need the parens?

converter/internal/staticconvert/testdata/prom_scrape.diags
```diff
@@ -1,2 +1,3 @@
 (Error) unsupported global evaluation_interval config was provided
+(Warning) Static mode command line flags which modify the running agent are not supported for the static to flow converter.
```
thampiotr, line 2: Nice! Should we add some call to action / how to resolve it?
e.g. "Please review your promtail command line flags and ensure they are set in Flow mode config file where necessary."

converter/internal/staticconvert/testdata/promtail_scrape.diags
thampiotr: Yeah... it could be a warning, but I do want it to catch attention. It's kinda like with the CLI arguments. Losing this (the CLI args or specific positions file location) may be harmless for many, but probably there is a good 20+% of cases out there for which it could be breaking the setup.

thampiotr (approved):
LGTM! just minor comments, happy for this to merge

converter/internal/common/common.go
```diff
@@ -49,14 +50,23 @@ func getValueOverrideHook() builder.ValueOverrideHook {
...
+	var sParts []string
+	for _, part := range parts {
+		if part != "" {
+			sParts = append(sParts, fmt.Sprintf("%v", part))
+		}
+	}
+	return strings.Join(sParts, "_")
+}
+
+func GetLabelWithIndex(index int, parts ...interface{}) string {
```
rfratto, line 63: Two nits:

1. It's slightly unidiomatic for functions to get Get in the name; these would normally be `LabelForParts` and `LabelWithIndex`
2. These should have some Godoc comments explaining to people writing converters what they do and how to use them :)

rfratto (approved):
Reviewed the code and most of the test cases, looks good!

converter/internal/prometheusconvert/prometheusconvert.go
```diff
@@ -106,34 +106,34 @@ func appendServiceDiscoveryConfigs(pb *prometheusBlocks, serviceDiscoveryConfig
 			targets = append(targets, getScrapeTargets(sdc)...)
 		case *prom_azure.SDConfig:
 			labelCounts["azure"]++
-			exports = appendDiscoveryAzure(pb, common.GetUniqueLabel(label, labelCounts["azure"]), sdc)
+			exports = appendDiscoveryAzure(pb, common.GetLabelWithIndex(labelCounts["azure"]-1, label), sdc)
 		case *prom_consul.SDConfig:
 			labelCounts["consul"]++
-			exports = appendDiscoveryConsul(pb, common.GetUniqueLabel(label, labelCounts["consul"]), sdc)
+			exports = appendDiscoveryConsul(pb, common.GetLabelWithIndex(labelCounts["consul"]-1, label), sdc)
```
mattdurham, line 112: nit: This looks odd with the `++` followed by an immediate `-1`, can we reverse the order and not have to do the -1?

mattdurham (approved):
LGTM

converter/internal/staticconvert/staticconvert.go
erikbaranowski: you do need it to do this inline here or you receive

`expected ';', found '{'`

converter/internal/staticconvert/staticconvert.go
erikbaranowski: I'm adding a helper defaulter func in the pkg/logs for this config and using it in both places for clarity and maintainability.

converter/internal/prometheusconvert/prometheusconvert.go
erikbaranowski: I don't love this either but it saves a line of code for each. I think this code will be refactored and disappear once the dust settles so I'm inclined to leave it for today.

longer explanation:
`labelCounts["consul"]++` will:
- add `consul` to the slice with an initial value of 1
- increment `consul`

Since this is 1 indexed it creates the unfortunate offset.

thampiotr (approved):

PR #4536, opened 2023-07-20
Remove zero width spaces from reference titles

clayton-cornell:
#### PR Description

Zero width spaces in titles is a legacy workaround that's no longer needed. There are still a few titles in the Agent ref docs that contain zero width spaces. This PR removes them 

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated


rfratto (approved):
image

If I squint, I think I can see the difference :)

clayton-cornell:
Limiting this backport to just v0.35. The change isn't visible to end users - it's housekeeping, so there's zero impact to users if it's not backported to v0.34 or v0.33.

PR #4537, opened 2023-07-21
Fix canonical URLs that were added with incorrect path prefix

jdbaldry:
I set the path prefix to /`docs/grafana/agent/latest/` and not `/docs/agent/latest/`


clayton-cornell (approved):
LGTM

jdbaldry:
Ah I'm not sure how well this will backport. It's not strictly necessary for any versions older than "latest". If you do want a similar patch in other branches, I can open those up.

clayton-cornell:
Removed backports since the change is only done on Next and not backported.

PR #4540, opened 2023-07-21
fix(helm): add hostPID value

korniltsev:
#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/flow/reference/components/pyroscope.ebpf.md
+++ b/docs/sources/flow/reference/components/pyroscope.ebpf.md
@@ -83,6 +83,12 @@ can help you pin down a profiling target.
 | `__name__`         | pyroscope metric name. Defaults to `process_cpu`.                                                                                |
 | `__container_id__` | The container ID derived from target.                                                                                            |
 
+
+### Privileges
+
+You are required to run the agent as root and inside host pid namespace in order to `pyroscope.ebpf` component to work.
+See helm example below how to do it with helm.
+
 ### Container ID
 
 Each collected stack trace is then associated with a specified target from the targets list, determined by a
@@ -152,8 +158,8 @@ strip elf -o elf.stripped
 objcopy --add-gnu-debuglink=elf.debug elf.stripped elf.debuglink
 ```
 
-For system libraries, ensure that debug symbols are installed. On Ubuntu, for example, you can install them by
-executing:
+For system libraries, ensure that debug symbols are installed. On Ubuntu, for example, you can install debug symbols 
+for `libc` by executing:
 
 ```bash
 apt install libc6-dbg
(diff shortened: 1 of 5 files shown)


docs/sources/flow/reference/components/pyroscope.ebpf.md
```diff
@@ -244,3 +250,45 @@ pyroscope.ebpf "default" {
   targets      = discovery.relabel.local_containers.output
 }
 ```
+
+### Helm deployment
```
mattdurham, line 254: @clayton-cornell is the below more like the task based doc? As opposed to reference?

clayton-cornell: @mattdurham Yes, this looks like a task... steps to follow to do a thing. A reference topic should only be covering things like functions and parameters, error messages, and return codes. One way it could be included here in a reference topic is if it's written as an example.

rfratto: @clayton-cornell Can the content about what privileges are required for the component to function stay in the reference documentation?

clayton-cornell: @rfratto @mattdurham Reading through this topic, I'd suggest that this information be presented differently. Assuming I understand things correctly, something like this:

1. Remove the new section titled `Privileges`.
1. Move the information about needing root and being inside host pid namespace up to the top and rework the text as a note: `To use the  `pyroscope.ebpf` component you must run Grafana Agent as root and inside host pid namespace.`  This appears to me to be a general requirement (root and in pid namespace) to use the component, correct?
1. The Helm deployment example is contained within the Example heading so it should be able to stay. It needs more description around it though:  `The following example shows you how you can configure Helm to...` with info about how this config is setting up as root and in the right namespace (I am not sure how this is done in that YAML though... which config sets the namespace, etc).  Then we'd need a sentence about what the `heml install` step is doing to round out the example.

I think this would meet the needs of a ref topic - or at least close enough.

rfratto: > This appears to me to be a general requirement (root and in pid namespace) to use the component, correct?

Yes, that's right 👍 

> The Helm deployment example is contained within the Example heading so it should be able to stay. 

I personally feel weird about the Helm deployment example in the reference documentation; I'd prefer if all the examples on reference pages were examples for how to configure Grafana Agent, with deployment tasks kept outside of component references. 

Final call is yours though, Clayton.

clayton-cornell: I'm 100% with you on the Helm deployment example - it's not really the right place for it here. I'm not sure where it can do right now though. I'll put suggestions in for the other changes I proposed.

clayton-cornell: @korniltsev Can we implement the first suggestion - I can't make this change without doing a git checkout on your branch.

Between line 13 and line 14 we insert the following:

```

To use the  `pyroscope.ebpf` component you must run Grafana Agent as root and inside host pid namespace. 
{{% /admonition %}}
````

Then remove the new section on titled ``Privileges` (lines 87-91).

korniltsev: Thanks for feedback and sorry for late response.

I've added a note to the root section, removed "privileges" section. 

I've also removed helmchart example section, I will figure out how/where to put it later.

tpaschalis (approved):

PR #4547, opened 2023-07-21
agent-flow-mixin: change heatmap type from `heatmap-new` to `heatmap`

rfratto:
The migration from `heatmap-new` to `heatmap` (in Grafana 10) seems to force the heatmap to calculate the heatmap buckets regardless of the setting of `calculate`. Changing the type to `heatmap` addresses this issue.

Related to #4527, which didn't fully fix the issue when migrating the dashboard to Grafana 10.


mattdurham (approved):
The new king is dead long live the old one...also LGTM

PR #4548, opened 2023-07-21
Fix nil reference oauth2

mattdurham:
#### PR Description

Fix nil reference with tls not being defined with Oauth2.

#### PR Checklist

- [X] CHANGELOG updated
- [ ] Documentation added
- [X] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@ Main (unreleased)
   `prometheus.operator` component which discovered no custom resources.
   (@rfratto)
 
-- Fix issue with nil tls config reference when using oauth2. (@mattdurham) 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -28,6 +28,8 @@ Main (unreleased)
   `prometheus.operator` component which discovered no custom resources.
   (@rfratto)
 
+- Fix issue with nil tls config reference when using oauth2. (@mattdurham)
```
rfratto, line 31: What is the "issue?" 

Also, I wouldn't mention `nil` since that's a Go specific term that isn't exposed to users, unlike a panic would be.

```suggestion
- Fix panic when using `oauth2` without specifying `tls_config`. (@mattdurham) 
```

rfratto (approved):

PR #4559, opened 2023-07-21
flow: always log when node evaluation completes

rfratto:
Previously, a log line when graph node evaluation completes was only generated when a graph node was evaluated as part of a full walk (meaning when the config file was reloaded).

This change ensures that node evaluation always generates a log line, including for partial walks (meaning when a component updates its exports, causing downstream dependencies to reevaluate).


erikbaranowski (approved):
LGTM

PR #4563, opened 2023-07-24
Improve allocations of the delta profiles appender

cyriltovena:
This improves allocations of the delta profiles appender.

Mostly it doesn't use `ioutil.ReadAll` but instead grows and reuses a buffer. I've also moved the buffer to a pool to avoid having those scale with the amount of targets.

Result on a very small profile.

```
❯ benchstat before.txt after.txt
ComputeDelta-16    31.1µs ± 5%    31.1µs ± 3%     ~     (p=0.842 n=9+10)

ComputeDelta-16      743B ± 0%      169B ± 8%  -77.22%  (p=0.000 n=10+10)

ComputeDelta-16      3.00 ± 0%      1.00 ± 0%  -66.67%  (p=0.000 n=10+10)
```

### DIFF
--- a/component/pyroscope/scrape/delta_profiles.go
+++ b/component/pyroscope/scrape/delta_profiles.go
@@ -2,7 +2,6 @@ package scrape
 
 import (
 	"bytes"
-	"compress/gzip"
 	"context"
 	"encoding/binary"
 	"fmt"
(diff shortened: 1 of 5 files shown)


korniltsev (approved):
great!

mattdurham (approved):
LGTM

PR #4564, opened 2023-07-24
clustering: wire in go-discover for discovering peers

### LINKED ISSUE #3484: Make use of go-discover to bootstrap cluster and add discover extra nodes
Instead of always relying on passing the IP of another node to join the cluster at as a command-line flag, we can make use of [go-discover](https://github.com/hashicorp/go-discover) as a second way of bootstrapping a cluster.

tpaschalis:
#### PR Description
This PR adds a new `cluster.discover-peers` command-line flag, to enable discovering clustering peers using the [go-discover](https://github.com/hashicorp/go-discover) library based on a list of key-value tuples, depending on the used "provider" and the settings it allows for.

We support all the default providers, plus the extra `k8s` provider.

Fixes #3484

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [ ] Tests updated


mattdurham (approved):
LGTM, its cool we already had support for discover peers in the gossip!

PR #4578, opened 2023-07-25
Fix incorrect display of trace IDs in the automatic_logging processor

### LINKED ISSUE #4552: 0.35.0 trace autologging provides incorrect trace IDs
### What's wrong?

Prior to v0.35.0, autologging correctly output loglines with the trace ID that was passed to it, eg:
```bash
  | 2023-07-21 11:21:24.493 | span=requester dur=12181760ns status=STATUS_CODE_OK http.method=GET http.target=unicorn svc=mythical-requester traceId=be8481dd93f1c751bd1ee320e5629197
-- | -- | --
```
From 0.35.0, the trace ID is now converted into what appears to be a decimal only identifer (that is not convertable to a working HEX trace ID hash):
```bash
2023-07-21 11:25:47.392 | span=requester dur=27946240ns status=Ok http.method=GET http.target=beholder svc=mythical-requester traceId=3661626132633865396432653963323538376566646565303064373233613266
-- | --
```

This means that correlation between traces that generate autologged log lines is not longer possible and breaks the general log->traces workflow.

### Steps to reproduce

Install Agent v0.35.0 into any ecosystem and then configure it to receive traces and autogenerate logs from traces. 

You can easily test this out using the [Intro to MLT](https://github.com/grafana/intro-to-mlt/) sandbox, and changing [this line](https://github.com/grafana/intro-to-mlt/blob/main/docker-compose.yml#L9) to `image: grafana/agent:v0.35.0`, then use the `docker-compose.yaml` file to initialise a working system.

Go to [the Explorer](http://localhost:3000/explore?orgId=1&left=%7B%22datasource%22:%22loki%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bjob%3D%5C%22agent%5C%22%7D%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22loki%22%7D,%22editorMode%22:%22builder%22%7D%5D,%22range%22:%7B%22from%22:%22now-5m%22,%22to%22:%22now%22%7D%7D) to see the result and inability to correlate with traces.

### System information

MacOS 12.6.5 (21G531)

### Software version

Grafana Agent v0.35.0

### Configuration

```text
automatic_logging:
        # Use the logs instance defined at the start of the configuration file, specifying a logs instance and the
        # named Promtail instance.
        backend: logs_instance
        logs_instance_name: loki
        # Ensure one log one line per root span (ie. one per trace).
        roots: true
        processes: false
        # Setting spans to `true` (default false) will generate a log line *per span* received. Note that this
        # can produce a huge amount of log information.
        spans: false
        # If present in the span, add the http.method, http.target and http.status_code span attributes to the log line.
        span_attributes:
          - http.method
          - http.target
          - http.status_code
        # Force the trace ID to be set as `traceId`. This ensures correlation from within Grafana between data sources.
        overrides:
          trace_id_key: "traceId"
```

### Logs

```text
2023-07-21 11:25:47.392 | span=requester dur=27946240ns status=Ok http.method=GET http.target=beholder svc=mythical-requester traceId=3661626132633865396432653963323538376566646565303064373233613266
-- | --
```

ptodev:
#### PR Description

This is a bug which was introduced in the Otel upgrade in #3858. I must have gotten confused about the [deprecation](https://github.com/open-telemetry/opentelemetry-collector/pull/6530) of the HexString function. Apologies for this - the bug could have been easily prevented if I'd taken more care reading the functions and the deprecation PR.

Fixes #4552

#### Notes to the Reviewer

There are two other places which use such code, but I think they are not impacted from an external user point of view:
* The `SpanToKeyVal` function in `pkg/integrations/v2/app_agent_receiver/payload.go` appears to be unused.
* In the service graph processor as long as a unique ID is generated (which I think is the case), then the code should work, even if the ID is not the same hex ID.

I will mention the service graph in the changelog anyway so that users are advised to upgrade.

I tested this locally for the automatic_logging processor only.

#### PR Checklist

- [x] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,8 @@ Main (unreleased)
 
 - Fix potential goroutine leak in log file tailing in static mode. (@thampiotr)
 
+- Fix incorrect display of trace IDs in the automatic_logging processor of static mode's traces subsystem. (@ptodev)
+
 ### Other changes
 
 - Compile journald support into builds of `grafana-agentctl` so
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,9 @@ Main (unreleased)
 
 - Fix potential goroutine leak in log file tailing in static mode. (@thampiotr)
 
-- Fix incorrect display of trace IDs in the automatic_logging processor of static mode's traces subsystem. (@ptodev)
+- Fix incorrect display of trace IDs in the automatic_logging processor of static mode's traces subsystem.
+  Users of the static mode's service graph processor are also advised to upgrade, 
+  although the bug should theoretically not affect them. (@ptodev)
 
 ### Other changes
 
--- a/pkg/integrations/v2/app_agent_receiver/payload.go
+++ b/pkg/integrations/v2/app_agent_receiver/payload.go
@@ -1,7 +1,6 @@
 package app_agent_receiver
 
 import (
-	"encoding/hex"
 	"fmt"
 	"sort"
 	"strconv"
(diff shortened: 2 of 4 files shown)


ptodev:
I think it's better to use `.String()` instead of `hex.EncodeToString` for two reasons:
* It is backwards compatible.
* The main use case for the `.String()` is logging, and that is exactly what the automatic_logging processor uses it for. 

For other use cases such as the one in the service graph processor, using `.String()` makes less sense, but since it's an older piece of code I think we could just leave it as it is.

thampiotr (approved):
Looks good to me!

tpaschalis (approved):

PR #4359, opened 2023-07-06
Cleanup after the Otel upgrade

ptodev:
#### PR Description

#3858 updated to version 0.80 of Otel Collector, but some work was still yet to be done. This PR completes the Otel upgrade by doing the following:

* Updates our components to use the latest features of Collector's components
* Disables Otel Collector "process" metrics
* Uses version `v1.0.0-rcv0013` of pdata and featuregate modules instead of the ones from the fork.

This PR completes the upgrade to version 0.80 of the Collector. No more work is expected to be needed.

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated

PR #4579, opened 2023-07-25
Fix upgrade guide

ptodev:
I'd [updated](https://github.com/grafana/agent/pull/4359/files#diff-8f08accdca0d98a1c64aa36075f87267ebe91f6db19ca8074554e58667f73d2f) the upgrade guide with a new "Main (unreleased)" section similar to the one in the changelog, but apparently there was already a 0.35 section underneath which I hadn't noticed. This will PR will add the deprecation information in the right section.

### DIFF
--- a/docs/sources/static/upgrade-guide.md
+++ b/docs/sources/static/upgrade-guide.md
@@ -20,37 +20,6 @@ static mode.
 > [upgrade-guide-operator]: {{< relref "../operator/upgrade-guide.md" >}}
 > [upgrade-guide-flow]: {{< relref "../flow/upgrade-guide.md" >}}
 
-## Main (unreleased)
-
-### Breaking change: Removed and renamed tracing metrics
-
-In the traces subsystem for Static mode some metrics are removed and others are renamed.
-The reason for the removal is a bug which caused the metrics to be incorrect if more than one instance of a traces configuration is specified.
-
-Removed metrics:
-- "blackbox_exporter_config_last_reload_success_timestamp_seconds" (gauge)
-- "blackbox_exporter_config_last_reload_successful" (gauge)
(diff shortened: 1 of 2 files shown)


docs/sources/static/upgrade-guide.md
```diff
@@ -97,6 +66,35 @@ See [Module and Auth Split Migration](https://github.com/prometheus/snmp_exporte
 The experimental feature Dynamic Configuration has been removed. The use case of dynamic configuration will be replaced
 with [Modules](../../concepts/modules/) in Grafana Agent Flow.
 
+### Breaking change: Removed and renamed tracing metrics
+
+In the traces subsystem for Static mode some metrics are removed and others are renamed.
+The reason for the removal is a bug which caused the metrics to be incorrect if more than one instance of a traces configuration is specified.
+
+Removed metrics:
+- "blackbox_exporter_config_last_reload_success_timestamp_seconds" (gauge)
```
marctc, line 75: These metrics are only removed if the integration is not used

ptodev: Even if the integration is used, would this metric show up? We are using a [detached commit](https://github.com/prometheus/blackbox_exporter/tree/bd22efa1c900a2418e9b5ae3f3d3148f33931b60) of the exporter, and I don't see Agent code calling the [NewSafeConfig](https://github.com/prometheus/blackbox_exporter/blob/bd22efa1c900a2418e9b5ae3f3d3148f33931b60/config/config.go#L92) function anywhere?

marctc: You are right, those metrics are not relevant to the agent anymore. Good call!

marctc (approved):

PR #4581, opened 2023-07-25
Cherry pick commits for the 35.1 patch release

ptodev:
These are all the commits which should go into 35.1.
Once this is merged, I'll raise another PR to remove the bugfix entry from the changelog of `main` and will add a new section for 35.1.


mattdurham (approved):
LGTM

PR #4590, opened 2023-07-25
Correct file name in the Agent standalone install and macOS install

clayton-cornell:
#### PR Description

This PR corrects the filename in the Agent standalone binary install topic.

* Change `grafana-agent-flow` to `grafana-agent` in step 3 of the download procedure.

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/flow/setup/install/binary.md
+++ b/docs/sources/flow/setup/install/binary.md
@@ -25,7 +25,7 @@ To download the Grafana Agent as a standalone binary, perform the following step
 
 1. Scroll down to the **Assets** section.
 
-1. Download the `grafana-agent-flow` zip file that matches your operating system and machine's architecture.
+1. Download the `grafana-agent` zip file that matches your operating system and machine's architecture.
 
 1. Extract the package contents into a directory.
 
--- a/docs/sources/flow/setup/install/macos.md
+++ b/docs/sources/flow/setup/install/macos.md
@@ -45,7 +45,7 @@ To upgrade Grafana Agent on macOS, run the following commands in a terminal wind
 1. Upgrade Grafana Agent:
 
    ```shell
-   brew upgrade grafana-agent
+   brew upgrade grafana-agent-flow
    ```
 
 1. Restart Grafana Agent:


docs/sources/flow/setup/install/binary.md
```diff
@@ -25,7 +25,7 @@ To download the Grafana Agent as a standalone binary, perform the following step
 
 1. Scroll down to the **Assets** section.
 
-1. Download the `grafana-agent-flow` zip file that matches your operating system and machine's architecture.
+1. Download the `grafana-agent` zip file that matches your operating system and machine's architecture.
```
ptodev, line 28: This is the "flow" section, so isn't "grafana-agent-flow" more correct? Or is the change because there is no "grafana-agent-flow-darwin" zip file?

tpaschalis: I think that's the reason; we only have `-flow` suffixed assets for deb/rpm and the windows installer.

tpaschalis (approved):

docs/sources/flow/setup/install/binary.md
ptodev: If we advise users to use `grafana-agent`, then we also would have to mention how to enable flow mode explicitly? Otherwise the instructions would be incomplete.

IMO we should just work to add the missing -flow binaries, since it seems like an oversight. I can't think of a reason why they we shouldn't have them.

tpaschalis: > If we advise users to use grafana-agent, then we also would have to mention how to enable flow mode explicitly? Otherwise the instructions would be incomplete.

I think that's covered in the "Next steps" part of the guide, and the "Start Grafana Agent" section.

> IMO we should just work to add the missing -flow binaries, since it seems like an oversight. I can't think of a reason why they we shouldn't have them.

The reason we initially added them as only as system packages was that, with system packages there was no way to run static mode and flow mode in parallel, so it was a way to offer users a workaround. On the other hand, with the binary, it's easy to do so (we just run it twice, with different AGENT_MODE variables).

We certainly _could_ add them as separate binaries, but I'm not sure if there's an immediate benefit here IMHO, what do you think?

clayton-cornell: Just to verify things here. The steps to explicitly enable flow mode on the standalone binaries is included in the Start steps for the binary (the command includes `AGENT_MODE=flow`. This PR fixes up the documented instructions for the state as things are right now.

ptodev: Thanks Clayton, that sounds fine.

PR #4605, opened 2023-07-26
enhancement(agentctl): better validation of config file

fgouteroux:
#### PR Description

Hello, I would like to add some validation steps in `grafana-agentctl config-check` cmd, like this:
```
$ grafana-agentctl config-check bad-config-values.yaml 
invalid metrics config: error validating instance test: scrape interval greater than wal_truncate_frequency for scrape config with job name "agent-3"
```
I'm using the `grafana-agentctl` cmd to check the config file validity before applying changes and restart the agent, so I would expect it fail at this step.

#### PR Checklist

- [x] CHANGELOG updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Main (unreleased)
 
 ### Enhancements
 
+- Better validation of config file with `grafana-agentctl config-check` cmd (@fgouteroux)
 - Add [godeltaprof](https://github.com/grafana/godeltaprof) profiling types (`godeltaprof_memory`, `godeltaprof_mutex`, `godeltaprof_block`) to `pyroscope.scrape` component
 
 - Integrations: make `udev` data path configurable in the `node_exporter` integration. (@sduranc)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Main (unreleased)
 ### Enhancements
 
 - Better validation of config file with `grafana-agentctl config-check` cmd (@fgouteroux)
+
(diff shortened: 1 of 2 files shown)


cmd/grafana-agentctl/main.go
```diff
@@ -132,9 +132,33 @@ the exit code will be 1.`,
...
+				fmt.Fprintf(os.Stderr, "invalid trace config: %s\n", err)
+				os.Exit(1)
+			}
+
+			if cfg.AgentManagement.Enabled {
+				if err := cfg.AgentManagement.Validate(); err != nil {
+					fmt.Fprintf(os.Stderr, "invalid agent management config: %s\n", err)
+					os.Exit(1)
+				}
+			}
```
rfratto, line 159: Have you tried replacing all of this with a call to `cfg.Validate(nil)`? It shouldn't be required to have an actual flagset:

 ```suggestion
			if err := cfg.Validate(nil); err != nil {
				fmt.Fprintf(os.Stderr, "failed to validate config: %s\n", err)
				os.Exit(1)
			}
```

**Review by rfratto — COMMENTED** [review]
Thanks for the PR!

By the way, can you list this change as an enhancement in CHANGELOG.md?

File: cmd/grafana-agentctl/main.go (hunk 1 above)
Comment (fgouteroux) [reply cmd/grafana-agentctl/main.go:159]: ah yes much better !

File: CHANGELOG.md [hunk 2]
```diff
@@ -12,6 +12,7 @@ Main (unreleased)
 
 ### Enhancements
 
+- Better validation of config file with `grafana-agentctl config-check` cmd (@fgouteroux)
```
rfratto, line 15: tiny nit: we try to have spacing between entries

```suggestion
- Better validation of config file with `grafana-agentctl config-check` cmd (@fgouteroux)

```

rfratto (approved):
Thank you!

PR #4608, opened 2023-07-26
clustering: enable refreshing the list of peers on an interval

### LINKED ISSUE #4465: Clustering: re-discover and re-join peers on interval in background
### Request

Add a new flag, `--cluster.rejoin-interval` which specifies how often a node should rediscover peers and rejoin them to address split brain issues. 

`--cluster.rejoin-interval` should default to some reasonable value, such as `60s`. 

When set to `0s`, rediscovery/rejoining is disabled. 

This proposal should be paired with #4464 to avoid overwhelming the network on large clusters, as the state push/pull done on join is more expensive than other gossip traffic. 

### Use case

Today, for nodes to join a cluster successfully, one of the following must be true:

* All nodes must join the same seed node, or 
* The initial cluster cannot be bootstrapped in parallel. 

This is because the set of peers to join is determined once at startup. 

In effect, this means that clustering on Kubernetes must only be enabled via a StatefulSet, and `podManagementPolicy` must not be set to Parallel. If neither of these conditions are met, clustering will end up in a split brain state. 

To avoid these constraints, nodes should rediscover and rejoin the cluster in the background on some timer to address split brain issues.

tpaschalis:
#### PR Description
This PR enables clustering to periodically refresh its list of peers found through one of `--cluster.join-addresses` or `--cluster.discover-peers` and attempt to re-join them using memberlist's Start.

This re-discovery is run by default every 60s, and is configurable via `--cluster.rejoin-interval`.

This allows for a more robust clustering implementation, where a temporary split brain at startup is not fatal, requiring _another_ node to coordinate the cluster.

To do this, the PR
* Decouples `JoinPeers` from its previous triple role (storing hardcoded addresses, used to populate the return of `go-discover` output and being passed to Start. Now JoinPeers is only
* Honors GossipConfig's behavior of not changing during runtime
* Extracts the logic of peer discovery into a new GetPeers method on gossipNode
* Uses the new method to bootstrap the cluster as well as re-discover and re-join peers

Fixes #4465 

#### Notes to the Reviewer

During the refresh, the node receives stale messages about itself for every cycle it does. I'm looking to whether we should do something about this.
```
ts=2023-07-26T17:53:05.244578Z level=debug component=memberlist ts="2023/07/26 20:53:05" caller="[DEBUG] memberlist: Initiating push/pull sync with:  127.0.0.1:12346" msg=
ts=2023-07-26T17:53:05.245871Z level=debug msg="merging remote state" remote_time=12
ts=2023-07-26T17:53:05.245903Z level=debug msg="got stale message about self" msg="node-a @14: participant"
ts=2023-07-26T17:53:06.611919Z level=debug msg="handling state message" msg="node-b @16: participant"
```

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [X] Tests updated

### DIFF
--- a/docs/sources/flow/reference/cli/run.md
+++ b/docs/sources/flow/reference/cli/run.md
@@ -48,6 +48,7 @@ The following flags are supported:
 * `--cluster.node-name`: The name to use for this node (defaults to the environment's hostname).
 * `--cluster.join-addresses`: Comma-separated list of addresses to join the cluster at (default `""`). Mutually exclusive with `--cluster.discover-peers`.
 * `--cluster.discover-peers`: List of key-value tuples for discovering peers (default `""`). Mutually exclusive with `--cluster.join-addresses`.
+* `--cluster.rejoin-interval`: How often to rejoin the list of peers (default `"60s"`).
 * `--cluster.advertise-address`: Address to advertise to other cluster nodes (default `""`).
 * `--config.format`: The format of the source file. Supported formats: 'flow', 'prometheus', 'promtail' (default `"flow"`).
 * `--config.bypass-conversion-errors`: Enable bypassing errors when converting (default `false`).
@@ -109,10 +110,18 @@ on the chosen provider and the filtering key-values it supports. Clustering
 supports the default set of providers available in go-discover and registers
 the `k8s` provider on top.
 
-If either the key or the value in a pair contains a space, a backslash, or
+If either the key or the value in a tuple pair contains a space, a backslash or
 double quotes, then it must be quoted with double quotes. Within this quoted
 string, the backslash can be used to escape double quotes or the backslash
 itself.
+
+The `--cluster.rejoin-interval` flag defines how often each node should
+rediscover peers based on the contents of the `--cluster.join-addresses` and
+`--cluster.discover-peers` flags and try to rejoin them.  This operation
+is useful for addressing split-brain issues if the initial bootstrap is
+unsuccessful and for making clustering easier to manage in dynamic
+environments. To disable this behavior, set the `--cluster.rejoin-interval`
+flag to `"0s"`.
 
 Discovering peers using the `--cluster.join-addresses` and
 `--cluster.discover-peers` flags only happens on startup; after that, cluster
@@ -120,7 +129,7 @@ nodes depend on gossiping messages with each other to converge on the cluster's
 state.
 
 The first node that is used to bootstrap a new cluster (also known as
-the "seed node") can either omit the flag that specifies peers to join or can
+the "seed node") can either omit the flags that specify peers to join or can
 try to connect to itself.
 
 ### Clustering states
--- a/docs/sources/flow/reference/cli/run.md
+++ b/docs/sources/flow/reference/cli/run.md
@@ -110,7 +110,7 @@ on the chosen provider and the filtering key-values it supports. Clustering
 supports the default set of providers available in go-discover and registers
 the `k8s` provider on top.
 
-If either the key or the value in a tuple pair contains a space, a backslash or
+If either the key or the value in a tuple pair contains a space, a backslash, or
 double quotes, then it must be quoted with double quotes. Within this quoted
 string, the backslash can be used to escape double quotes or the backslash
 itself.
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -93,7 +93,7 @@ type Clusterer struct {
 	Node Node
 }
 
-func getJoinAddr(addrs []string, in string) []string {
+func appendJoinAddr(addrs []string, in string) []string {
 	_, _, err := net.SplitHostPort(in)
 	if err == nil {
 		addrs = append(addrs, in)
@@ -117,7 +117,7 @@ func getJoinAddr(addrs []string, in string) []string {
 }
 
 // New creates a Clusterer.
-func New(log log.Logger, reg prometheus.Registerer, clusterEnabled bool, name, listenAddr, advertiseAddr, joinAddr, discoverPeers string) (*Clusterer, error) {
+func New(log log.Logger, reg prometheus.Registerer, clusterEnabled bool, name, listenAddr, advertiseAddr, joinAddr, discoverPeers string, rejoinInterval time.Duration) (*Clusterer, error) {
 	// Standalone node.
 	if !clusterEnabled {
 		return &Clusterer{Node: NewLocalNode(listenAddr)}, nil
@@ -125,13 +125,13 @@ func New(log log.Logger, reg prometheus.Registerer, clusterEnabled bool, name, l
 
 	gossipConfig := DefaultGossipConfig
 
-	defaultPort := 80
 	_, portStr, err := net.SplitHostPort(listenAddr)
 	if err == nil { // there was a port
-		defaultPort, err = strconv.Atoi(portStr)
+		defaultPort, err := strconv.Atoi(portStr)
 		if err != nil {
 			return nil, err
 		}
+		gossipConfig.DefaultPort = defaultPort
 	}
 
 	if name != "" {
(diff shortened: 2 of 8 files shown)


pkg/cluster/gossip.go
```diff
@@ -123,8 +123,17 @@ func (c *GossipConfig) ApplyDefaults() error {
...
 		}
 		c.discoverer = d
+	}
+
+	return c.RefreshPeers()
+}
 
-		addrs, err := d.Addrs(c.DiscoverPeers, stdlog.New(io.Discard, "", 0)) // TODO(rfratto): log to log.Logger?
+// RefreshPeers updates the list of peers the node will look to connect to.
+func (c *GossipConfig) RefreshPeers() error {
```
rfratto, line 141: This requires ApplyDefaults to be called before calling RefreshPeers, otherwise `c.discoverer` may be nil. 

Can you find a way to either:

1. Change the semantics to avoid c.discoverer being nil, or 
2. Document that ApplyDefaults _must_ be called?

(IMO it's a bit weird that the config is responsible for discovering and refreshing peers, so it might be possible to change the semantics to introduce some separation of concerns)

pkg/cluster/gossip.go
```diff
@@ -53,15 +54,18 @@ type GossipConfig struct {
...
-	// exclusive with DiscoverPeers.
+	// List of one or more host:port peer addresses to connect to.
 	//
 	// If an agent connects to no peers, it will form a one-node cluster until a
 	// peer connects to it explicitly.
 	JoinPeers flagext.StringSlice
 
+	// Comma-separated list of hosts, DNS records, or host:port peer addresses
+	// to look up. Mutually exclusive with DiscoverPeers.
+	JoinAddr string
```
rfratto, line 66: This feels off to me; we now have JoinAddr and JoinPeers, which do the same thing except JoinPeers is effectively ignored. 

We shouldn't be introducing a new field here IMO.

pkg/cluster/cluster.go
```diff
@@ -199,6 +201,28 @@ func (c *Clusterer) Start(ctx context.Context) error {
...
 
+		if node.cfg.RejoinInterval > 0 {
+			go func() {
+				t := time.NewTicker(node.cfg.RejoinInterval)
+				for {
+					select {
+					case <-t.C:
+						err := node.cfg.RefreshPeers()
+						if err != nil {
+							level.Error(node.log).Log("msg", "failed to refresh the list of peers", "err", err)
```
rfratto, line 212: Do we want this to continue down to node.Start even if RefreshPeers fails?

pkg/cluster/gossip.go
tpaschalis: The reason for this is that JoinAddr may be a DNS record that will be expanded to multiple addresses. If we use the same field, we'll lose that information during the refresh, and it will only be applicable to the DiscoverPeers.

Then JoinAddr/DiscoverPeers are the way to rediscover and JoinPeers is the actual list of peers (with expanded DNS records, go-discover addresses, the default port added etc).

tpaschalis: Maybe the separation of concerns is a good idea to generate the list of `<host>:<port>` addresses to join in a different way.

rfratto: Can we change the semantics so that JoinPeers is expanded if it contains a DNS record?

tpaschalis: > Can we change the semantics so that JoinPeers is expanded if it contains a DNS record?

Yes, but still the result has to be stored separately so that it has the chance to be expanded on the next tick as well.

rfratto: IMO it definitely feels like we should move some of that state storage to the clustering instance instead of the config.

docs/sources/flow/reference/cli/run.md
```diff
@@ -100,18 +101,26 @@ on the chosen provider and the filtering key-values it supports. Clustering
...
+If either the key or the value in a tuple pair contains a space, a backslash or
 double quotes, then it must be quoted with double quotes. Within this quoted
 string, the backslash can be used to escape double quotes or the backslash
 itself.
 
+The `--cluster.rejoin-interval` flag defines how often each node should
+rediscover peers based on the contents of the `--cluster.join-addresses` and
+`--cluster.discover-peers` flags and try to rejoin them.  This operation
+is useful for addressing split-brain issues if the initial bootstrap was not
+successful, as well as making clustering easier to manage in dynamic
```
clayton-cornell, line 113:
```suggestion
`--cluster.discover-peers` flags and try to rejoin them.  This operation
is useful for addressing split-brain issues if the initial bootstrap is
unsuccessful and for making clustering easier to manage in dynamic
```

pkg/cluster/gossip_test.go
```diff
@@ -117,27 +122,27 @@ func TestConfig_ApplyDefaults(t *testing.T) {
...
+		err := gc.ApplyDefaults()
 		require.NoError(t, err)
-		require.Equal(t, []string{"fizzbuzz:5555"}, []string(gc.JoinPeers))
+		require.Equal(t, []string{"fizzbuzz:5555"}, getPeers(t, gc))
 	})
 
 	t.Run("peers can use default port", func(t *testing.T) {
 		gc := defaultConfig
-		gc.JoinPeers = []string{"foobar"}
+		gc.JoinPeers = []string{"192.168.1.14"}
```
tpaschalis, line 132: With the new behavior, `foobar` was being handled as a DNS record that cannot be looked up, so I changed it to an IP address.

pkg/cluster/gossip.go
tpaschalis: I've documented that ApplyDefaults must be called successfully before using it to create a NewGossipNode.

tpaschalis:
@rfratto I've decoupled JoinPeers's usage from defining static peers, to being overwritten by DiscoverPeers _and_ being used in calls to Start(). Now GossipConfig does not change during runtime

I'd like to add some more tests to increase our confidence, but wanted to see what you thought of this separation of concerns.

tpaschalis:
> I still have some gripes with it having internal state and

What do you mean by 'internal state' here, the discoverer? Because the list of peers is re-generated on every tick from the clusterer's configuration.

rfratto:
> What do you mean by 'internal state' here, the discoverer? Because the list of peers is re-generated on every tick from the clusterer's configuration.

Yeah, the discoverer. If I was writing this from scratch I would:

* Move validation of GossipConfig to `NewGossipNode`
* Have `NewGossipNode` construct and store the `discoverer` implementation.

This would fully decouple config settings from business logic.

clayton-cornell:
The current docs are fine as-is.

pkg/cluster/cluster.go
```diff
@@ -197,16 +187,37 @@ func (c *Clusterer) Start() error {
...
+						level.Error(node.log).Log("msg", "failed to refresh the list of peers", "err", err)
+						continue // we'll try on the next tick
+					}
+					err = node.Start(peers)
+					if err != nil {
+						level.Error(node.log).Log("msg", "failed to rejoin the list of peers", "err", err)
+					}
+				}
+			}()
+		}
```
rfratto, line 219: Tiny nit: this technically leaks and doesn't have a way to ever terminate the goroutine.

rfratto (approved):

pkg/cluster/cluster.go
tpaschalis: Initially I'd passed the parent context to stop the ticker when it's cancelled, but when this happens we'd already be exiting. I'll fix it either way.

pkg/cluster/gossip.go
```diff
@@ -94,46 +101,42 @@ func (c *GossipConfig) ApplyDefaults(defaultPort int) error {
 		if err != nil {
 			return fmt.Errorf("determining advertise address: %w", err)
 		}
-		c.AdvertiseAddr = fmt.Sprintf("%s:%d", addr.String(), defaultPort)
+		c.AdvertiseAddr = fmt.Sprintf("%s:%d", addr.String(), c.DefaultPort)
 	} else {
-		c.AdvertiseAddr = appendDefaultPort(c.AdvertiseAddr, defaultPort)
+		c.AdvertiseAddr = appendDefaultPort(c.AdvertiseAddr, c.DefaultPort)
```
mattdurham, line 106: Is the assumption that AdvertiseAddr does NOT contain the port?

mattdurham (approved):
LGTM

docs/sources/flow/reference/cli/run.md
```diff
@@ -109,18 +110,26 @@ on the chosen provider and the filtering key-values it supports. Clustering
 supports the default set of providers available in go-discover and registers
 the `k8s` provider on top.
 
-If either the key or the value in a pair contains a space, a backslash, or
+If either the key or the value in a tuple pair contains a space, a backslash or
```
clayton-cornell, line 113:
```suggestion
If either the key or the value in a tuple pair contains a space, a backslash, or
```

clayton-cornell (commented):
Missed a comma

clayton-cornell (approved):
LGTM

PR #4612, opened 2023-07-26
Expose WAL configuration in `loki.write`

thepalbi:
#### PR Description

This PR exposes in river settings for configuring WAL support inside the `loki.write` component. By default, WAL will be disabled until we get comfortable enough with running, and provide support for it. If disabled, the component will work as normal. If enabled, received entries will be first written to WAL, and from there read and remote-written by each remote write client.

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [X] Tests updated


component/common/loki/wal/writer.go
```diff
@@ -83,7 +83,7 @@ func NewWriter(walCfg Config, logger log.Logger, reg prometheus.Registerer) (*Wr
 	}
 
 	wrt.reclaimedOldSegmentsSpaceCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
-		Namespace: "promtail",
+		Namespace: "loki_write",
```
thepalbi, line 86: driveby

docs/sources/flow/reference/components/loki.write.md
```diff
@@ -115,6 +117,26 @@ based on a hash of the endpoint settings.
 
 {{< docs/shared lookup="flow/reference/components/tls-config-block.md" source="agent" >}}
 
+### wal block
```
thepalbi, line 120: should we add a note here mentioning this is experimental for the time being?

component/common/loki/wal/config.go
```diff
@@ -20,17 +21,37 @@ type Config struct {
 	//
 	// WAL support is a WIP. Do not enable in production setups until https://github.com/grafana/loki/issues/8197
 	// is finished.
-	Enabled bool `yaml:"enabled"`
+	Enabled bool `yaml:"enabled" river:"enabled,attr,optional"`
 
 	// Path where the WAL is written to.
-	Dir string `yaml:"dir"`
+	Dir string `yaml:"dir" river:"dir,attr,optional"`
```
rfratto, line 27: This is a bit of an anti-pattern for Flow I think; components are given directories unique to that component and they should use that for storage instead of having it be configurable.

docs/sources/flow/reference/components/loki.write.md
rfratto: Yes, mark the header as experimental:

```suggestion
### wal block (experimental)
```

component/common/loki/wal/config.go
thepalbi: Ah you mean `opts.DataPath` right? Do you have an example of a component using it? Is that directory kept the same if one restarts the agent?

rfratto: > Ah you mean opts.DataPath right? Do you have an example of a component using it?

Yeah, for an example, look at this in prometheus.remote_write where the WAL is stored:

https://github.com/grafana/agent/blob/ba235197332fb071a126a3389ce12367ff020850/component/prometheus/remotewrite/remote_write.go#L75-L76

> Is that directory kept the same if one restarts the agent?

Depends on how the agent is configured. All the components get their data paths from the `--storage.path` flag; typically if the user wants it to persist they would set `--storage.path` to something which persists between restarts/reboots.

clayton-cornell (commented):
A few suggestions in the description of the WAL block

component/common/loki/wal/config.go
thepalbi: Applied the same idea used in the prom remote write

docs/sources/flow/reference/components/loki.write.md
```diff
@@ -115,6 +117,26 @@ based on a hash of the endpoint settings.
...
 
+### wal block (experimental)
+
+The optional `wal` block configures the write-ahead log used in the Loki remote-write client. If omitted, the WAL
+will be disabled. When enabled, the log entries sent to the `loki.write` component will be first written to a write-ahead
+log, under the `dir` directory, and then read from there into the remote-write client. This provides durability guarantees
+when an entry reaches this component. The client knows when to read from the WAL using the following two mechansims:
+- The WAL-writer side of the `loki.write` component notifies the reader side that there is new data available.
+- The WAL-reader side check periodically if there is new data, increasing the wait time exponentially, between
+  `min_read_frequency` and `max_read_frequency`.
```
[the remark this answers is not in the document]
thepalbi: applied manually

component/common/loki/wal/config.go
thepalbi: Ok, using a sub-dir under the component data dir.

thepalbi:
@rfratto also created a new `WalArguments` struct to just expose what's needed in the to be configured, rather than coupling this behaviour in the internal `wal.Config` struct.

component/loki/write/write.go
```diff
@@ -32,6 +34,27 @@ type Arguments struct {
...
+// WalArguments holds the settings for configuring the Write-Ahead Log (WAL) used
+// by the underlying remote write client.
+type WalArguments struct {
+	Enabled          bool          `river:"enabled,attr,optional"`
+	MaxSegmentAge    time.Duration `river:"max_segment_age,attr,optional"`
+	MinReadFrequency time.Duration `river:"min_read_frequency,attr,optional"`
+	MaxReadFrequency time.Duration `river:"max_read_frequency,attr,optional"`
+}
+
+func (wa *WalArguments) SetToDefault() {
```
tpaschalis, line 49: nit: Do we also want to implement the Validator interface to eg. make sure that `min_read_frequency < max_read_frequency < max_segment_age`?

docs/sources/flow/reference/components/loki.write.md
```diff
@@ -115,6 +117,32 @@ based on a hash of the endpoint settings.
...
+
+The WAL is located inside a component-specific directory relative to the
+storage path Grafana Agent is configured to use. See the
+[`agent run` documentation][run] for how to change the storage path.
+
+The following arguments are supported:
+
+Name                  | Type       | Description                                                                                                        | Default   | Required
+--------------------- |------------|--------------------------------------------------------------------------------------------------------------------|-----------| --------
+`enabled`                 | `bool`     | Wether to enable or not the WAL.                                                                                   | false     | no
```
tpaschalis, line 139:
```suggestion
`enabled`                 | `bool`     | Whether to enable the WAL.                                                                                   | false     | no
```

tpaschalis (approved):
LGTM!

tpaschalis (approved):

PR #4615, opened 2023-07-27
Add an otelcol.processor.discovery component

### LINKED ISSUE #2299: Flow: otelcol.processor.discovery component
Expose the Grafana Agent-specific custom `promsdprocessor` processor component as a Flow component called `otelcol.processor.discovery`. 

Note that this component does not have an upstream counterpart; it was something created specifically for Grafana Agent to allow for label association with telemetry data coming from Prometheus SD.  

It's likely that we won't want to use the current processor and would instead make a custom one. The current processor runs a Prometheus Service Discovery manager and spawns various SD workers to find targets. This isn't necessary for Flow; we could instead just pass it a list of discovered targets similarly like what you would do for `prometheus.scrape`.

ptodev:
#### PR Description

This component exists in static mode, but not in the Otel Collector. I tried to reuse as much code as possible between static and flow mode.

Fixes #2299

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated


pkg/traces/promsdprocessor/consumer/consumer.go
```diff
@@ -0,0 +1,249 @@
...
+
+func (c *Consumer) SetHostLabels(hostLabels map[string]discovery.Target) {
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+
+	c.hostLabels = hostLabels
+}
+
+func (c *Consumer) GetHostLabels(ip string) (map[string]string, error) {
+	c.mtx.Lock()
```
mattdurham, line 116: nit: RLock/RUnlock instead? Would need to change to RWMutex

component/otelcol/processor/discovery/discovery.go
```diff
@@ -0,0 +1,147 @@
...
+var _ component.Component = (*Component)(nil)
+
+// New creates a new otelcol.exporter.discovery component.
+func New(o component.Options, c Arguments) (*Component, error) {
+	if c.Output.Logs != nil || c.Output.Metrics != nil {
+		level.Warn(o.Logger).Log("msg", "non-trace input detected; this component only works for traces")
+	}
+
+	nextTraces := fanoutconsumer.Traces(c.Output.Traces)
+	consumer, err := promsdconsumer.NewConsumer(nextTraces, c.OperationType, c.PodAssociations, o.Logger)
```
mattdurham, line 94: If I update pod associations how do those get updated?

pkg/traces/promsdprocessor/consumer/consumer.go
ptodev: Actually, I just realised there is a bigger issue 😬  This function returns a map which points to the same values. The map is not deep copied. The original code [doesn't have this issue](https://github.com/grafana/agent/blob/v0.35.1/pkg/traces/promsdprocessor/prom_sd_processor.go#L174-L177) because it locks a mutex for as long as we need to read the values in the map.

I deleted this function and switched back to the original implementation's choice of locking a mutex for as long as we need to read the data.

Thank you for bringing this to attention!

component/otelcol/processor/discovery/discovery.go
ptodev: [consumer.go](https://github.com/grafana/agent/blob/d53425c45bee6737e65a4daa40ab29bac52fb4f1/pkg/traces/promsdprocessor/consumer/consumer.go) has to be updated, as well as functions listing default values for pod associations, such as:

- [static mode](https://github.com/grafana/agent/blob/d53425c45bee6737e65a4daa40ab29bac52fb4f1/pkg/traces/promsdprocessor/prom_sd_processor.go#L54-L62) defaults
- [flow mode](https://github.com/grafana/agent/blob/d53425c45bee6737e65a4daa40ab29bac52fb4f1/component/otelcol/processor/discovery/discovery.go#L49-L55) defaults
- unit tests, such as [here](https://github.com/grafana/agent/blob/d53425c45bee6737e65a4daa40ab29bac52fb4f1/pkg/traces/promsdprocessor/prom_sd_processor.go#L54-L62) and [here](https://github.com/grafana/agent/blob/d53425c45bee6737e65a4daa40ab29bac52fb4f1/pkg/traces/promsdprocessor/consumer/consumer_test.go#L253-L261)

docs/sources/flow/reference/components/otelcol.processor.discovery.md
```diff
@@ -0,0 +1,159 @@
...
+
+    output {
+        traces = [otelcol.exporter.otlp.default.input]
+    }
+}
+```
+
+### Using more than one discovery process
+
+Outputs form more than one discovery process can be combined via the `concat` function.
```
clayton-cornell, line 120:
```suggestion
Outputs from more than one discovery process can be combined via the `concat` function.
```

component/otelcol/processor/discovery/discovery.go
ptodev: Wow, you're absolutely correct that this wasn't working properly. I think it is fixed now - please take a look?

mattdurham: Looks good now!

pkg/traces/promsdprocessor/consumer/consumer.go
mattdurham: Is this all resolved now?

ptodev: Yes. Now we just lock the mutex for [as long as the data is being read](https://github.com/grafana/agent/blob/2299-flow-otelcolprocessordiscovery-component/pkg/traces/promsdprocessor/consumer/consumer.go#L153-L175).

clayton-cornell:
Docs look OK now.

mattdurham (approved):
LGTM

clayton-cornell (approved):

component/otelcol/processor/discovery/discovery.go
```diff
@@ -0,0 +1,151 @@
...
+		}
+		promsdconsumer.CleanupLabels(labels)
+		hostLabels[host] = labels
+	}
+
+	c.consumer.UpdateOptions(promsdconsumer.Options{
+		HostLabels:      hostLabels,
+		OperationType:   cfg.OperationType,
+		PodAssociations: cfg.PodAssociations,
+		NextConsumer:    fanoutconsumer.Traces(cfg.Output.Traces),
```
ptodev, line 147: @mattdurham Apologies, but I had push a commit to fix another issue after the latest review. I wasn't updating the "next consumer".

pkg/traces/promsdprocessor/consumer/consumer.go
```diff
@@ -0,0 +1,254 @@
...
+// UpdateOptionsHostLabels is used in static mode, where only the host labels need to be updated.
+func (c *Consumer) UpdateOptionsHostLabels(hostLabels map[string]discovery.Target) {
+	c.optsMut.Lock()
+	defer c.optsMut.Unlock()
+
+	c.opts.HostLabels = hostLabels
+}
+
+func (c *Consumer) ConsumeTraces(ctx context.Context, td ptrace.Traces) error {
+	c.optsMut.RLock()
```
ptodev, line 130: @mattdurham  I also moved this mutex here instead of in `processAttributes`. This is so that we use a consistent set of "next consumer" and data from the discovery components. Unfortunately this means the mutex is held unnecessarily long for static mode, because out of the four "options" static mode only updates HostLabels:

* HostLabels
* OperationType 
* PodAssociations
* NextConsumer

I think this should be fine because I expect this traces processing to complete in less time than Prometheus' service discovery refresh. But want to give you a heads up.

pkg/traces/promsdprocessor/consumer/consumer.go
```diff
@@ -0,0 +1,254 @@
...
+	if cl.Addr == nil {
+		return ""
+	}
+
+	host := cl.Addr.String()
+	if strings.Contains(host, ":") {
+		var err error
+		splitHost, _, err := net.SplitHostPort(host)
+		if err != nil {
+			// It's normal for this to fail for IPv6 address strings that don't actually include a port.
```
mattdurham, line 220: If its normal should we return an error at all? IE only error if IP4?

mattdurham (changes requested):
Some questions on functionality around the splitting of IPs

pkg/traces/promsdprocessor/consumer/consumer.go
ptodev: I think this is why this is a debug log instead of a warning log. It's supposed to be ok for a failure to happen, but the debug log might be of interest in case it's a genuinely malformed hostname (e.g. `::1]:80` instead of  `[::1]:80`). Btw this function is from [old code](https://github.com/grafana/agent/blob/v0.35.3/pkg/traces/promsdprocessor/prom_sd_processor.go#L122-L141).

ptodev: Btw we are not returning an error. In case of an error, we are returning the original address string.

pkg/traces/promsdprocessor/consumer/consumer.go
```diff
@@ -0,0 +1,254 @@
...
+	if !ok {
+		return "", fmt.Errorf("unable to find address in labels %q", labels.Labels())
+	}
+
+	host := address
+	if strings.Contains(host, ":") {
+		var err error
+		host, _, err = net.SplitHostPort(host)
+		if err != nil {
+			return "", fmt.Errorf("unable to split host and port in address %q: %w", address, err)
```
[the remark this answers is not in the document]
ptodev: The comment from `getConnectionIP` says that the error is expected if no port is present in an IPv6 address. I suppose that `GetHostFromLabels` expects a port to always be present in the Prometheus `__address__` label. Btw `GetHostFromLabels` is [this](https://github.com/grafana/agent/blob/v0.35.3/pkg/traces/promsdprocessor/prom_sd_processor.go#L279-L293) code from the old codebase.

pkg/traces/promsdprocessor/consumer/consumer.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+var _ otelconsumer.Traces = (*Consumer)(nil)
+
+func NewConsumer(opts Options, logger log.Logger) (*Consumer, error) {
+	c := &Consumer{
+		logger: logger,
+	}
+
+	err := c.UpdateOptions(opts)
```
ptodev, line 92: @mattdurham I updated `NewConsumer` to call `UpdateOptions` instead of setting the options directly. This ensures we do the same checks on new options in both `NewConsumer` and in `UpdateOptions`.

mattdurham (approved):
LGTM

PR #4619, opened 2023-07-27
pkg/metrics/wal: always write series records to WAL, even on rollback

rfratto:
If a new series is introduced in a storage.Appender instance, that series should be written to the WAL once the storage.Appender is closed, even on Rollback.

Previously, new series would only be written to the WAL when calling Commit. However, because the series is stored in memory regardless, subsequent calls to Commit may write samples to the WAL which reference a series ID which that was never written.

Port of prometheus/prometheus#12592

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,10 @@ Main (unreleased)
 
 - Fix potential goroutine leak in log file tailing in static mode. (@thampiotr)
 
+- Fix issue where series records would never get written to the WAL if a scrape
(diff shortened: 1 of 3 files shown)


mattdurham:
Is there a good way to add tests to this?

rfratto:
> Is there a good way to add tests to this? 

I added a test case, PTAL

rfratto:
> Is there a good way to add tests to this?

@mattdurham I added a test case, PTAL

mattdurham (approved):
Lgtm

PR #4620, opened 2023-07-27
flow: Add 'write_to' argument to logging block

### LINKED ISSUE #2700: flow: Allow the `logging` block to send the Agent's own logs to`loki.*` components

tpaschalis:
#### PR Description
This is a rehash of #3566 to add a new `write_to` argument to the special logging configuration block, to tee the Agent's own logs to one or more loki.* components' exported receivers.

Fixes #2700

#### Notes to the Reviewer
I'll look into whether I can add some test for this.

#### PR Checklist

- [X] CHANGELOG updated
- [X] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ Main (unreleased)
 
 - Clustering: Enable peer discovery with the go-discover package. (@tpaschalis)
 
+- Flow: Allow the `logging` configuration block to tee the Agent's logs to one
+  or more loki.* components. (@tpaschalis)
+
 - New Grafana Agent Flow components:
 
   - `prometheus.exporter.gcp` - scrape GCP metrics (@tburgessdev)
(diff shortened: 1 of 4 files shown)


rfratto (approved):
🎉 Super exciting! This will really simplify our instructions for how to self-collect telemetry data.

PR #4611, opened 2023-07-26
Make RPM file digests use FIPS-compliant sha256 instead of MD5

andrewimeson:
#### PR Description

Make RPM file digests use FIPS-compliant sha256 instead of MD5

#### Which issue(s) this PR fixes

Closes #4419, relates to #4267

#### PR Checklist

- [X] CHANGELOG updated
- [X] Documentation added
- [X] Tests updated

PR #4622, opened 2023-07-27
Fix issue with fips on behalf of @andrewimeson

mattdurham:
This is wholly a copy of https://github.com/grafana/agent/pull/4611/files so we can get it into a patch asap.


erikbaranowski (approved):
LGTM

PR #4640, opened 2023-07-28
Cleanup static mode after the otel upgrade, fix bug with Jaeger remote sampling

ptodev:
#### PR Description

There is an issue with Jaeger remote sampling not working after #3858 was merged. This PR will fix it.
I also updated the docs, and updated other components in the Static mode's tracing subsystem to use new parameters which were not previously available on the old Otel version.

#### Notes to the Reviewer

* We will need to create a patch release for v0.35 once this PR is merged to `main`.
* I wish I could add more tests - for example, for marshalling to "secret".
* I wish there could be an error if there is an empty `jaeger_remote_sampling:` in the config. However, I could not get that to work.
* This is not the last fix which has to go to 0.35.3. I will also need to fix another bug due to the Otel upgrade, but will do it on a separate PR.

#### PR Checklist

- [x] CHANGELOG updated
- [x] Documentation added
- [x] Tests updated


docs/sources/flow/upgrade-guide.md
```diff
@@ -21,6 +21,12 @@ Grafana Agent Flow.
 
 ## v0.35
 
+### Breaking change: Jaeger remote sampling does not work
+
+Jaeger remote sampling used to be configured via the jaeger receiver. This receiver was updated to a new version, and remote sampling has stopped working.
+
+Jaeger remote sampling will be reintroduced in v35.3. It will be configured separately from the jaeger receiver.
```
rfratto, line 28: I don't know if we should introduce new functionality in a patch release; it goes against semver. We could say this is a known issue in v0.35 and to wait until v0.36 for a fix.

mattdurham: If it is something that is currently broken then fixing it in a patch via new functionality feels reasonable. The other option that is most semver compatible is to roll back the otel upgrade. Given the effort there that is painful to do. @ptodev have we gotten any reports on this or did we find it? 

I guess fundamentally though we could roll out a v0.36 instead of a patch if we wanted to align with semver?

rfratto: Yeah I'm not going to push too hard on this, but it does feel bad to categorize it as a breaking change, since that feels more deliberate/planned, and this is just an artifact of upgrading.

clayton-cornell (approved):
No additional changes for docs. It's fine as-is

ptodev:
Thank you for the reviews. I updated the doc with Robert's suggestions, added example configs to the upgrade guide, and also moved my changes from the flow upgrade guide to the static one (I didn't notice it earlier, apologies 🤦‍♂️  ). I'll merge this when the CI finishes.

Issue #4464, opened 2023-07-14
Clustering: don't try to join all peers on startup


rfratto:
### Request

Add a new flag, `-cluster.max-initial-join-peers` which specifies the number of initial peers to join from the discovered set. 

When `-cluster.max-initial-join-peers` is set to 0, all discovered peers should be joined. 

When `-cluster.max-initial-join-peers` is set to a non-zero number, the subset of peers used for the join should be randomized. 

If `-cluster.max-initial-join-peers` is larger than the number of discovered peers, all discovered peers are used for joining, rather than it being an error.  

### Use case

Currently, when an agent starts with clustering enabled, it performs a push/pull with every discovered peer before the join is considered successful.

This operation can be costly for clusters of significant sizes.

PR #4647, opened 2023-07-31
clustering: wait for graph evaluation before changing to Participant state

### LINKED ISSUE #4639: Flow clustering: only change to participant state after initial config file load
### Background

In clusters with significantly sized WALs (containing tens of millions of active series), WAL replay on startup can take several minutes. 

However, because nodes immediately transition to the participant state, work assigned to these nodes during WAL replay will see no scrapes or incoming metrics until WAL replay completes. This can lead to a gap of metrics. 

There are a few workarounds to this solution:

* Run a higher number of shards, leading to a smaller WAL, and thus a shorter replay time. 
* Truncate the WAL more frequently so there are a smaller number of segments which need replaying.

I propose changing the semantics of when a node enters the participant state so replaying the WAL doesn't delay a subset of targets from being scraped.

### Proposal

Nodes should only change to the participant state after the initial config file load completes (which is the same condition for /-/ready returning HTTP 200).

This means that other ready nodes can continue accepting the load of new or restarted pods while the new or restarted pod is replaying its WAL.

tpaschalis:
#### PR Description
This commit enables delaying the transition of the GossipNode to Participant state to after the graph has been completely evaluated at least once. This guards against missing scrapes when other peers think we're ready to participate to distributing load, but we're still loading all our components.

Fixes #4639.

#### Notes to the Reviewer
A more straighforward solution would be to move the [Start call](https://github.com/grafana/agent/pull/4647/files#diff-a5071380b513bfb7750c3b1d67b1f7b9e34fcd4e86d24c0ac19df8b9dddc6900R267) in  `cmd/internal/flowmode/cmd_run.go` to after the block performing the initial reload. 

On one hand, this would be simpler/synchronous but on the other hand would add another unnecessary waiting time to bootstrap a new cluster, especially if it's a large one. Since we don't have #4464 in place yet, this can be expensive; this solution makes sure that the cluster bootstrapping and the graph evaluation can happen in parallel.

#### PR Checklist

- [x] CHANGELOG updated
- [X] Tests updated

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -192,7 +192,7 @@ func New(log log.Logger, reg prometheus.Registerer, clusterEnabled bool, name, l
 // new cluster of its own.
 // The gossipNode will start out as a Viewer; to participate in clustering,
 // the node needs to transition to the Participant state using ChangeState.
-func (c *Clusterer) Start(ctx context.Context) error {
+func (c *Clusterer) Start() error {
 	switch node := c.Node.(type) {
 	case *localNode:
 		return nil // no-op, always ready
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -190,6 +190,8 @@ func New(log log.Logger, reg prometheus.Registerer, clusterEnabled bool, name, l
 // For the gossipNode implementation, Start will attempt to connect to the
 // configured list of peers; if this fails it will fall back to bootstrapping a
 // new cluster of its own.
+// The gossipNode will start out as a Viewer; to participate in clustering,
+// the node needs to transition to the Participant state using ChangeState.
 func (c *Clusterer) Start(ctx context.Context) error {
 	switch node := c.Node.(type) {
 	case *localNode:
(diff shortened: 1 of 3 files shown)


pkg/cluster/cluster.go
```diff
@@ -206,15 +207,33 @@ func (c *Clusterer) Start(ctx context.Context) error {
...
+		// to verify that the graph has been evaluated to move on to the
+		// Participant state and signal that we wish to participate in reading
 		// or writing data.
-		ctx, ccl := context.WithTimeout(ctx, 5*time.Second)
-		defer ccl()
-		err = node.ChangeState(ctx, peer.StateParticipant)
-		if err != nil {
-			return err
-		}
+		go func(ctx context.Context) {
```
rfratto, line 214: Are there any benefits of doing this instead of extracting the call to ChangeState to a new method which we invoke when we know the load has finished?

tpaschalis: You're right, I was overthinking it, KISS :)

The errors that can return are either due to invalid transitions or not being able to encode the message, both of which wouldn't be solved with a retry.

rfratto (approved):
Much better :) Thanks!

PR #4657, opened 2023-07-31
Promtail converters: add remaining pipeline_stages -> loki.process conversions

thampiotr:
#### PR Description

This adds the remaining stages for the `pipeline_stages` to `loki.process` conversions. 

#### Notes to the Reviewer

Most of them are supported, but some are unsupported an also not documented in promtail docs.

Remaining features to implement conversions for:
* Encoding
* DecompressionCfg

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/converter/internal/promtailconvert/internal/build/stages.go
+++ b/converter/internal/promtailconvert/internal/build/stages.go
@@ -5,6 +5,8 @@ import (
 	"sort"
 	"time"
 
+	"github.com/alecthomas/units"
+
 	promtailmetric "github.com/grafana/loki/clients/pkg/logentry/metric"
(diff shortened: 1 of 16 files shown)


converter/internal/promtailconvert/internal/build/stages.go
thampiotr, line 1: We could split this into files, but I don't think it would improve the developer experience as it would result in many small files instead.

erikbaranowski: makes sense and I agree 👍

erikbaranowski (approved):
LGTM!

PR #4661, opened 2023-07-31
Converter static integration v1 pattern

erikbaranowski:
#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/converter/internal/prometheusconvert/prometheusconvert.go
+++ b/converter/internal/prometheusconvert/prometheusconvert.go
@@ -36,7 +36,7 @@ func Convert(in []byte) ([]byte, diag.Diagnostics) {
 	}
 
 	f := builder.NewFile()
-	diags = AppendAll(f, promConfig, "")
+	diags = AppendAll(f, promConfig, "", []discovery.Target{})
 
 	var buf bytes.Buffer
 	if _, err := f.WriteTo(&buf); err != nil {
@@ -58,7 +58,7 @@ func Convert(in []byte) ([]byte, diag.Diagnostics) {
 // Exports from other components are correctly referenced to build the Flow
 // pipeline. A non-empty labelPrefix can be provided for label uniqueness when
 // calling this function for the same builder.File multiple times.
-func AppendAll(f *builder.File, promConfig *prom_config.Config, labelPrefix string) diag.Diagnostics {
+func AppendAll(f *builder.File, promConfig *prom_config.Config, labelPrefix string, additionalTargets []discovery.Target) diag.Diagnostics {
 	pb := newPrometheusBlocks()
 
 	remoteWriteExports := appendPrometheusRemoteWrite(pb, promConfig.GlobalConfig, promConfig.RemoteWriteConfigs, labelPrefix)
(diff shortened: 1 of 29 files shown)


converter/internal/staticconvert/internal/build/builder.go
```diff
@@ -0,0 +1,84 @@
...
+
+func (b *IntegrationsV1ConfigBuilder) AppendExporter(commonConfig *int_config.Common, name string, extraTargets []discovery.Target) {
+	scrapeConfigs := []*prom_config.ScrapeConfig{}
+	if b.cfg.Integrations.ConfigV1.ScrapeIntegrations {
+		scrapeConfig := prom_config.DefaultScrapeConfig
+		scrapeConfig.MetricsPath = fmt.Sprintf("integrations/%s/metrics", name)
+		scrapeConfig.JobName = fmt.Sprintf("integrations/%s", name)
+		scrapeConfig.RelabelConfigs = commonConfig.RelabelConfigs
+		scrapeConfig.MetricRelabelConfigs = commonConfig.MetricRelabelConfigs
+		// TODO: Add support for scrapeConfig.HTTPClientConfig
```
erikbaranowski, line 62: I think there's something to wire up here but I'm not 100% sure if that's true or where to wire from.

converter/internal/prometheusconvert/prometheusconvert.go
```diff
@@ -36,7 +37,7 @@ func Convert(in []byte) ([]byte, diag.Diagnostics) {
 	}
 
 	f := builder.NewFile()
-	diags = AppendAll(f, promConfig, "")
+	diags = AppendAll(f, promConfig, "", []discovery.Target{}, nil)
```
thampiotr, line 40: Maybe we can have a couple of specialised `Append*` - which would also work better with the suggestion I made below regarding the label function passed as an argument?

For example:
* `AppendAll(file, prom_config)` -> calls `AppendAllNested(....)` with some hard-coded arguments
* `AppendAllNested(file, prom_config, jobNameToCompLabelsFunc, extraTargets, remoteWriteExports)`

converter/internal/staticconvert/internal/build/apache_exporter.go
```diff
@@ -0,0 +1,32 @@
...
+	b.f.Body().AppendBlock(common.NewBlockWithOverride(
+		[]string{"prometheus", "exporter", "apache"},
+		compLabel,
+		args,
+	))
+
+	return prometheusconvert.NewDiscoverExports(fmt.Sprintf("prometheus.exporter.apache.%s.targets", compLabel))
+}
+
+func ToApacheExporter(config *apache_http.Config) *apache.Arguments {
```
thampiotr, line 26: Could be un-exported to start with.

converter/internal/staticconvert/internal/build/builder.go
```diff
@@ -0,0 +1,84 @@
...
+		if !integration.Common.Enabled {
+			continue
+		}
+
+		var exports discovery.Exports
+		switch itg := integration.Config.(type) {
+		case *apache_http.Config:
+			exports = b.AppendApacheExporter(itg, &integration.Common)
+		case *node_exporter.Config:
+			exports = b.AppendNodeExporter(itg, &integration.Common)
```
thampiotr, line 45: Should we have a clause for unsupported integration?
a new one can be added, without adding support in the config converters

converter/internal/staticconvert/internal/build/node_exporter.go
```diff
@@ -0,0 +1,119 @@
...
+	b.f.Body().AppendBlock(common.NewBlockWithOverride(
+		[]string{"prometheus", "exporter", "unix"},
+		"",
+		args,
+	))
+
+	return prometheusconvert.NewDiscoverExports("prometheus.exporter.unix.targets")
+}
+
+func ToNodeExporter(config *node_exporter.Config) *unix.Arguments {
```
thampiotr, line 23: Yikes. But I guess it's a necessary evil - I don't see how we could make it simpler without inventing some magic conversion protocol.

thampiotr (approved):
LGTM! Just a couple of suggestions.

converter/internal/staticconvert/internal/build/builder.go
erikbaranowski: That gets handled in the separate validate.go functionality with a similar case statement for unsupported integrations. This will be up for discussion and refactor if we want to do it inline once the dust settles on the first 3 converters.

converter/internal/staticconvert/internal/build/builder.go
erikbaranowski: This would be for the `prometheus.scrape` component generated by this converter to scrape the exporter.

thampiotr (approved):
LGTM, thanks!

PR #4666, opened 2023-08-01
Fix the type of Password field in kafka config

### LINKED ISSUE #4636: loki.source.kafka should use secret type for password
### What's wrong?

https://github.com/grafana/agent/blob/29131a6cfbe6af4939577a85940836103bf628fb/component/loki/source/kafka/kafka.go#L56 uses `string` for password

### Steps to reproduce

Set `loki.source.kafka` password in the args.
Go to Agent UI.
See the password leaked in the UI.

### System information

_No response_

### Software version

_No response_

### Configuration

_No response_

### Logs

_No response_

harsiddhdave44:
The `Password` field in `kafka.go` was set to string, which caused it to be exposed. This PR fixes it by changing its type to `flagext.Secret`, which ensures that the value is not exposed directly without explicit calling to its `String()` method

- Fixes #4636

#### PR Checklist

- [ ] CHANGELOG updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/component/loki/source/kafka/kafka.go
+++ b/component/loki/source/kafka/kafka.go
@@ -53,7 +53,7 @@ type KafkaAuthentication struct {
 type KafkaSASLConfig struct {
 	Mechanism   string            `river:"mechanism,attr,optional"`
 	User        string            `river:"user,attr,optional"`
-	Password    string            `river:"password,attr,optional"`
+	Password    flagext.Secret    `river:"password,attr,optional"`
 	UseTLS      bool              `river:"use_tls,attr,optional"`
 	TLSConfig   config.TLSConfig  `river:"tls_config,block,optional"`
 	OAuthConfig OAuthConfigConfig `river:"oauth_config,block,optional"`
@@ -192,13 +192,13 @@ func (args *Arguments) Convert() kt.Config {
 }
 
 func (auth KafkaAuthentication) Convert() kt.Authentication {
-	var secret flagext.Secret
-	if auth.SASLConfig.Password != "" {
-		err := secret.Set(auth.SASLConfig.Password)
-		if err != nil {
-			panic("Unable to set kafka SASLConfig password")
-		}
-	}
+	// var secret flagext.Secret
+	// if auth.SASLConfig.Password.String() != "" {
+	// 	err := secret.Set(auth.SASLConfig.Password)
+	// 	if err != nil {
+	// 		panic("Unable to set kafka SASLConfig password")
+	// 	}
+	// }
 
 	return kt.Authentication{
 		Type:      kt.AuthenticationType(auth.Type),
@@ -206,7 +206,7 @@ func (auth KafkaAuthentication) Convert() kt.Authentication {
 		SASLConfig: kt.SASLConfig{
 			Mechanism: sarama.SASLMechanism(auth.SASLConfig.Mechanism),
 			User:      auth.SASLConfig.User,
-			Password:  secret,
+			Password:  auth.SASLConfig.Password,
 			UseTLS:    auth.SASLConfig.UseTLS,
 			TLSConfig: *auth.SASLConfig.TLSConfig.Convert(),
 			OAuthConfig: kt.OAuthConfig{
--- a/component/loki/source/kafka/kafka.go
+++ b/component/loki/source/kafka/kafka.go
@@ -27,16 +27,6 @@ func init() {
 	})
 }
 
-// Takes in a variable of type `rivertypes.Secret`, converts it to an OptionalSecret in order to extract the string value
-func ConvertSecretToString(secret rivertypes.Secret) (string, error) {
-	var optSecret rivertypes.OptionalSecret
-	err := secret.ConvertInto(&optSecret)
-	if err != nil {
-		return "", err
-	}
-	return optSecret.Value, nil
-}
-
 // Arguments holds values which are used to configure the loki.source.kafka
 // component.
 type Arguments struct {
--- a/component/loki/source/kafka/kafka.go
+++ b/component/loki/source/kafka/kafka.go
@@ -203,14 +203,13 @@ func (args *Arguments) Convert() kt.Config {
 }
 
 func (auth KafkaAuthentication) Convert() kt.Authentication {
-	password, _ := ConvertSecretToString(auth.SASLConfig.Password)
 	return kt.Authentication{
 		Type:      kt.AuthenticationType(auth.Type),
 		TLSConfig: *auth.TLSConfig.Convert(),
 		SASLConfig: kt.SASLConfig{
 			Mechanism: sarama.SASLMechanism(auth.SASLConfig.Mechanism),
 			User:      auth.SASLConfig.User,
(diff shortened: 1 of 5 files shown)


marctc (commented):
Thanks for the contribution! You also need to update the type of `Password` in `loki.source.kafka.md` with fie

component/loki/source/kafka/kafka.go
```diff
@@ -53,7 +53,7 @@ type KafkaAuthentication struct {
 type KafkaSASLConfig struct {
 	Mechanism   string            `river:"mechanism,attr,optional"`
 	User        string            `river:"user,attr,optional"`
-	Password    string            `river:"password,attr,optional"`
+	Password    flagext.Secret    `river:"password,attr,optional"`
```
tpaschalis, line 56: Since we don't expose this setting as a flag, I think it should be set as a `rivertypes.Secret` type.

Sorry in advance for the confusion; initially the code was forklifted off Promtail, but we're gradually trying to make it more suited to Flow, and this is one of those cases.

.github/workflows/codesee-arch-diagram.yml
```diff
@@ -0,0 +1,23 @@
+# This workflow was added by CodeSee. Learn more at https://codesee.io/
```
tpaschalis, line 1: I'm not sure if this was accidentally added, but can we remove this?

harsiddhdave44: Yes, I did that by mistake. I'm removing it

component/loki/source/kafka/kafka.go
marctc: `SASLConfig` uses `flagext.Secret`. Should we be consistent with the secret types?

tpaschalis: IMHO, fields with River tags should use `rivertypes`; then we can convert to whatever format the underlying code requires.

In this case, it would work either way because flagext.Secret is just an embedded string, but in the general case it wouldn't be certain that Password could be decoded directly from the River type. WDYT?

harsiddhdave44: Yes, Previously I modified the `Password` field earlier to `flagext.Secret`, but as pointed earlier by @tpaschalis, I need to migrate it to rivertypes.Secret instead

marctc: I was implying to change the type in the other struct as well, sorry. Maybe needs to be done in a separate PR

component/loki/source/kafka/kafka_test.go
```diff
@@ -59,6 +59,8 @@ func TestSASLRiverConfig(t *testing.T) {
 	var args Arguments
 	err := river.Unmarshal([]byte(exampleRiverConfig), &args)
 	require.NoError(t, err)
+
+	require.NotEqual(t, "password", args.Authentication.SASLConfig.Password.String())
```
ptodev, line 63: I personally think a `require.Equal` would give the reader of the test a better understanding of what the expected output is.

harsiddhdave44: Yes, you're right. I realized it after pushing the changes. I'll do it, thanks for pointing it out.

clayton-cornell (commented):
Docs look OK. Nothing else in the docs should need updates.

harsiddhdave44:
@marctc Can you please review the PR now?

docs/sources/flow/reference/components/loki.source.kafka.md
```diff
@@ -107,12 +107,12 @@ you must set the `tls_config` block. If `"sasl"` is used, you must set the `sasl
...
- `user`      | `string` | The user name to use for SASL authentication.                                 | `""`       | no       
- `password`  | `string` | The password to use for SASL authentication.                                  | `""`       | no       
- `use_tls`   | `bool`   | If true, SASL authentication is executed over TLS.                            | `false`    | no       
+ Name        | Type                 | Description                                                                   | Default    | Required 
+-------------|----------------------|--------------------------------------------------------------------|----------|-----------------------
+ `mechanism` | `string`             | Specifies the SASL mechanism the client uses to authenticate with the broker. | `"PLAIN""` | no       
+ `user`      | `string`             | The user name to use for SASL authentication.                                 | `""`       | no       
+ `password`  | `rivertypes.Secret`  | The password to use for SASL authentication.                                  | `""`       | no
```
marctc, line 114: we use `secret` instead of the whole type

component/loki/source/kafka/kafka.go
```diff
@@ -192,21 +203,14 @@ func (args *Arguments) Convert() kt.Config {
...
 func (auth KafkaAuthentication) Convert() kt.Authentication {
-	var secret flagext.Secret
-	if auth.SASLConfig.Password != "" {
-		err := secret.Set(auth.SASLConfig.Password)
-		if err != nil {
-			panic("Unable to set kafka SASLConfig password")
-		}
-	}
-
+	password, _ := ConvertSecretToString(auth.SASLConfig.Password)
```
marctc, line 206: In the previous change, we where panic-ing if the secret couldn't be set. Here we leave the error pass silently. Is this intended?

component/loki/source/kafka/kafka.go
```diff
@@ -26,6 +27,16 @@ func init() {
 	})
 }
 
+// Takes in a variable of type `rivertypes.Secret`, converts it to an OptionalSecret in order to extract the string value
+func ConvertSecretToString(secret rivertypes.Secret) (string, error) {
```
marctc, line 31: I think this function should be unexported. Also, consider moving into `rivertypes` package if could be used in other components.

harsiddhdave44: Yes, it can be used in other components as well as we move to `Secret` type for all sensitive data. I'll move it into the `rivertypes` package

component/loki/source/kafka/kafka.go
harsiddhdave44: I wasn't sure how to handle this error. Can you give me some suggestions?

marctc: If the function doesn't bubble the error, i think using the logger should be enough. If this error is as bad to panic the whole agent, you could use panic

harsiddhdave44: @marctc I'm not sure how to get the logger instance in the `Convert() kt.Authentication` function in `component\loki\source\kafka\kafka.go`. Can you help me out here?

marctc: I think in this case you can propagate the error to args `Convert` and then to `Update`

component/loki/source/kafka/kafka.go
rfratto, line 31: A function like this isn't necessary, you can just cast a `rivertypes.Secret` into a string without any logic:

`string(secretValue)`

wildum:
@harsiddhdave44 Hello, do you have time/want to continue working on this PR?

rfratto:
Hi, It's been a few months so I'm going to take this over so we can get this merged.

PR #4668, opened 2023-08-01
loki.source.file: allow defining the encoding of read files

### LINKED ISSUE #3938: loki.source.file: allow tailing non-UTF8 encoded files
### Request

We should change `loki.source.file` to allow the files encoding to be set to something other than UTF-8. I'm not sure if this should be a target-specific option that can be defined via a label or a component-level option, although I'm leaning towards the latter.

### Use case

This is something that is supported on Promtail and should be brought over for feature-parity. Some applications output their logs in non-UTF8 encodings by default, and changing that might be challenging for end-users in restricted environments.

tpaschalis:
#### PR Description
This PR allows `loki.source.file` to define the encoding of the scraped files as part of the component argument block.

I toyed with the idea of allowing the encoding to be set per-file with a dedicated label (eg. `__filepath_encoding__`) but I'm not sure the added overhead is worth it.

Fixes #3938 

#### Notes to the Reviewer

If the encoding field is not set in the component's arguments, the test would fail with the following error
```
        	Error Trace:	agent/component/loki/source/file/file_test.go:234
        	Error:      	Not equal:
        	            	expected: "hello world!�"
        	            	actual  : "\x00h\x00e\x00l\x00l\x00o\x00 \x00w\x00o\x00r\x00l\x00d\x00!\x00"
```

#### PR Checklist

- [X] CHANGELOG updated
- [X] Documentation added
- [X] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ Main (unreleased)
 - Clustering: Nodes take part in distributing load only after loading their
   component graph. (@tpaschalis)
 
+- Allow `loki.source.file` to define the encoding of files. (@tpaschalis)
+
 - New Grafana Agent Flow components:
 
   - `prometheus.exporter.gcp` - scrape GCP metrics. (@tburgessdev)
--- a/component/loki/source/file/file.go
+++ b/component/loki/source/file/file.go
@@ -35,10 +35,10 @@ const (
 
 // Arguments holds values which are used to configure the loki.source.file
 // component.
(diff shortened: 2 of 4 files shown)


rfratto (approved):
LGTM, thanks!

thampiotr (approved):
LGTM! 
I'll need to change https://github.com/grafana/agent/pull/4671 after this lands, or the other way around.

Issue #4665, opened 2023-07-31
Agent does not resolve DNS short names on Windows


scottamation:
### What's wrong?

Since v0.29.0 (when the project changed to using go 1.19) windows agents are unable to resolve DNS short names in the agent config. This is due go 1.19 adding windows support for the native go network stack. Unfortunately the native go DNS resolution process does not act the same as the native OS DNS which causes unexpected behavior.
[This same issue](https://github.com/prometheus/prometheus/issues/11480) was brought up in the Prometheus project and was resolved by having separate build tags for windows which excludes netgo.

### Steps to reproduce

1. Install an agent v0.29.0 or newer on a windows OS.
2. Create a new static metrics config.
3. Reference the target by shortname (no domain).
4. Access the the Agent API at /agent/api/v1/metrics/targets
5. See an error similar to this:
    "scrape_error":"Get \"http://servername:80/metrics\": dial tcp: lookup servername on 10.235.134.16:53: no such host"
6. Change the agent config to reference the target by FQDN.
7. Observe metrics working as expected
8. Revert the Config and install an agent version older than v0.29.0
9. Observe metrics working as expected

### System information

Windows x86_64

### Software version

Grafana Agent v0.29.0+

PR #4672, opened 2023-08-01
build: do not produce Windows binaries with netgo build tags

rfratto:
Go 1.19 adds Windows support for the native Go network stack, but doesn't include support for resolving DNS short names.

Other projects, including Prometheus, have updated their build process to exclude the netgo build tags when producing Windows binaries to work around this behavior.

Fixes grafana/agent#4665

### DIFF
--- a/tools/make/packaging.mk
+++ b/tools/make/packaging.mk
@@ -69,6 +69,9 @@ dist/grafana-agent-darwin-arm64: generate-ui
 
 # NOTE(rfratto): do not use netgo when building Windows binaries, which
 # prevents DNS short names from being resovable. See grafana/agent#4665.
+#
+# TODO(rfratto): add netgo back to Windows builds if a version of Go is
+# released which natively supports resolving DNS short names on Windows.
 dist/grafana-agent-windows-amd64.exe: GO_TAGS += builtinassets
 dist/grafana-agent-windows-amd64.exe: GOOS    := windows
 dist/grafana-agent-windows-amd64.exe: GOARCH  := amd64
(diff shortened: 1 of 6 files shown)


ptodev:
I think you mean that it fixes #4665 and not #4465 ?

rfratto:
> I think you mean that it fixes https://github.com/grafana/agent/issues/4665 and not https://github.com/grafana/agent/issues/4465 ?

🤦 Yes, thank you

tools/make/packaging.mk
```diff
@@ -31,63 +31,65 @@ dist-agent-binaries: dist/grafana-agent-linux-amd64       \
...
 
-dist/grafana-agent-darwin-arm64: GO_TAGS += builtinassets
+dist/grafana-agent-darwin-arm64: GO_TAGS += netgo builtinassets
 dist/grafana-agent-darwin-arm64: GOOS    := darwin
 dist/grafana-agent-darwin-arm64: GOARCH  := arm64
 dist/grafana-agent-darwin-arm64: generate-ui
 	$(PACKAGING_VARS) AGENT_BINARY=$@ $(MAKE) -f $(PARENT_MAKEFILE) agent
 
+# NOTE(rfratto): do not use netgo when building Windows binaries, which
+# prevents DNS short names from being resovable. See grafana/agent#4665.
```
ptodev, line 71: Do we need to add a TODO to revert this change when the underlying Go issue is resolved?
https://github.com/golang/go/issues/57757

rfratto: I don't think that's the same issue, is it? I don't see any mention of the hosts file in the issue or the Prometheus issue.

ptodev: Ah yes, I think you're right and it's not the same issue. But it's probably another issue which needs to be resolved before we use netgo on windows. Anyway, thank you for adding the TODO.

PR #4677, opened 2023-08-01
Use a github app instead of grafanabot in drone

mattdurham:
The Grafanabot is rate limited and switching to per project github apps. This is a start to get that working.


mattdurham:
Successfully finds the secrets, not sure the best way to actually test it until we release. Thankfully its easy to undo. The website update is still blocked until docs team switches but thats on the scheudule.

mattdurham:
Yes I was going to do the Github Actions in a separate PR.

rfratto (approved):
Rubber stamp, let's see if it works

PR #4685, opened 2023-08-02
flow: fix graceful termination on kubernetes

### LINKED ISSUE #4684: Grafana Agent Flow may not be shutting down gracefully in k8s environment
### What's wrong?

While chasing around a reproduction of a `ClusterNodeUnhealthy` alert I noticed that pods didn't seem to report changing to the the Terminating state when being deleted.

Looking through the code, it seems that the interruptContext we're [creating](https://github.com/grafana/agent/blob/a392061ec7f9eb5edd8ebbcdb716c0c9ccd217da/cmd/internal/flowmode/cmd_run.go#L363) responds on SIGINTs, while Kubernetes [sends SIGTERMs](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/) to containers in the pod for graceful termination instead.

If we exec into the pod and use `kill -2 1` instead, we can see the agent pod being terminated gracefully, with its components shutting down and the clusterer transitioning to the terminating state.

Am I possibly missing something here?

### Steps to reproduce

* Deploy Grafana Agent Flow on Kubernetes
* Use `kubectl delete` to kill one of the pods
* Notice that we're _not_ receiving log messages such as the following
```
ts=2023-08-02T09:38:58.216209812Z level=info service=http msg="http server closed" addr=[::]:80 err="http: Server closed"
ts=2023-08-02T09:38:58.216719782Z level=info service=http msg="http server closed" addr=memory err="http: Server closed"
ts=2023-08-02T09:38:58.216793323Z level=info component=loki.write.grafanacloud msg="component exited"
ts=2023-08-02T09:38:58.216813135Z level=debug component=prometheus.remote_write.grafanacloud msg="closing storage"
ts=2023-08-02T09:38:58.222019573Z component=prometheus.remote_write.grafanacloud subcomponent=rw level=info remote_name=60d445 url=https://prometheus-prod-01-eu-west-0.grafana.net/api/prom/push msg="Stopping remote storage..."
...
```

### System information

_No response_

### Software version

Grafana Agent v0.35.2

### Configuration

_No response_

### Logs

_No response_

tpaschalis:
#### PR Description
The signal is natively implemented in Linux and macOS.

On Windows, _CTRL_SHUTDOWN_EVENT which is used on Windows Kubernetes to signal graceful termination is [mapped](https://github.com/golang/go/blob/5f2e24efb3c7e021308d15a26d93e5a7aa3c05f0/src/runtime/os_windows.go#L1105-L1112) to SIGTERM internally so it should work in the same way.

Fixes #4684

#### Notes to the Reviewer
The SIGKILL signal is used as a backup when the graceful termination period is over. Do we think we should have some extra handling around this?

Also, I find it weird this hasn't caused any other problems recently. Is it possible I'm missing any context or that this was done on purpose?

#### PR Checklist

- [X] CHANGELOG updated
- [ ] Tests updated (N/A)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,9 @@ Main (unreleased)
 
 - Allow `loki.source.file` to define the encoding of files. (@tpaschalis)
 
+- Enable graceful termination when receiving SIGTERM/CTRL_SHUTDOWN_EVENT
+  signals. (@tpaschalis)
+
 - New Grafana Agent Flow components:
 
   - `prometheus.exporter.gcp` - scrape GCP metrics. (@tburgessdev)
--- a/cmd/internal/flowmode/cmd_run.go
+++ b/cmd/internal/flowmode/cmd_run.go
@@ -360,7 +360,7 @@ func interruptContext() (context.Context, context.CancelFunc) {
 	go func() {
 		defer cancel()
 		sig := make(chan os.Signal, 1)
-		signal.Notify(sig, os.Interrupt)
+		signal.Notify(sig, os.Interrupt, syscall.SIGTERM)
 		select {
 		case <-sig:
 		case <-ctx.Done():


rfratto (approved):
Hm, I'm also surprised we haven't ran into other issues because of this. The main problem we'd see is WAL corruption, but we automatically handle that now so it might've been happening and went unnoticed. 

Does static mode respond to SIGTERM properly?

tpaschalis:
> Does static mode respond to SIGTERM properly?

Yes it does; we use a handler generated by `github.com/weaveworks/common/signals` which looks for SIGINT/SIGTERM to exit and SIGQUIT to dump goroutines to the logger.

rfratto:
> Yes it does; we use a handler generated by github.com/weaveworks/common/signals which looks for SIGINT/SIGTERM to exit and SIGQUIT to dump goroutines to the logger.

👍 I guess it's good to know Flow is resilient to forceful terminations, but we should still definitely merge this 😛

PR #4673, opened 2023-08-01
non_indexed_labels stage

vlad-diachenko:
#### PR Description

Added non_indexed_labels stage that attaches the label<>values from extracted data to log line entry.

#### Which issue(s) this PR fixes

it repeats the implementation added to promtail in https://github.com/grafana/loki/pull/9986

#### Notes to the Reviewer
Documentation will be added in a separate PR

#### PR Checklist

- [x] CHANGELOG updated
- [ ] Documentation added
- [x] Tests updated

PR #4687, opened 2023-08-02
The commit the exporter to was originally pinned to doesnt work 

mattdurham:
Due to me amending the PR the commit hash is different and likely only working due to go proxy caching the commit. Switching to using the most recent commit in the multi collector. The date is slightly newer since I included a change that better allows debugging and some minor cleanup.


mattdurham:
Closing since it will be resolved with https://github.com/grafana/agent/pull/4673/commits/e222294ce0e963a35664e95aa9ed68ccbee8f2d8

PR #4689, opened 2023-08-02
Propagate services through modules 

rfratto:
> **NOTE**: I recommend reviewing this PR commit-by-commit. 

This change allows components to propagate services to module controllers.

To avoid cyclic dependencies, only services which module loader depends on can be propagated. This effectively means that module loaders must always depend on all services.

Related to #4253.


rfratto:
I'm taking this out of draft for now. I don't know if this is the best solution, but implementation is complete enough to merge if it's approved and I can address feedback for how else to implement propagation if it's not approved.

pkg/flow/internal/controller/component_registry.go
```diff
@@ -0,0 +1,45 @@
...
+// Get retrieves a component using [component.Get].
+func (m RegistryMap) Get(name string) (component.Registration, bool) {
+	reg, ok := m[name]
+	return reg, ok
+}
+
+// MultiRegistry returns a ComponentRegistry which combines the input
+// registries. Earlier registries take precednece over later registries if a
+// component is registered in both.
+func MultiRegistry(reg ...ComponentRegistry) ComponentRegistry {
```
mattdurham, line 32: Whats the purpose of multi registries? Is this mostly for testing?

rfratto: > Whats the purpose of multi registries? Is this mostly for testing?

Yeah, I'm not using it right now so I could take it out, but it's intended for testing, where you want to overlay custom components on top of the default set without modifying the default set

pkg/flow/flow.go
```diff
@@ -202,45 +218,39 @@ func newController(modReg *moduleRegistry, o Options) *Flow {
 			HTTPListenAddr:  o.HTTPListenAddr,
 			DialFunc:        dialFunc,
 			ControllerID:    o.ControllerID,
-			NewModuleController: func(id string) controller.ModuleController {
+			NewModuleController: func(id string, keepServices []string) controller.ModuleController {
```
mattdurham, line 221: nit: The naming feels a little off for `keepServices`? Should these be `availableServices` or follow `needsServices`?

pkg/flow/internal/controller/component_registry.go
mattdurham: It feels like complexity we don't need right now. That makes me have to consider things like multi registration. Is it possible to make it an interface and then put multi capability around testcomponent?

rfratto: I'm not sure what you're suggesting. I'll just take the code out since it's not being called anywhere right now anyway.

mattdurham (approved):
Minor nits but LGTM

pkg/flow/flow.go
```diff
@@ -145,13 +145,27 @@ type Flow struct {
...
+}
+
+// controllerOptions are internal options used to create both root Flow
+// controller and controllers for modules.
+type controllerOptions struct {
+	Options
+
+	ComponentRegistry controller.ComponentRegistry // Custom component registry used in tests.
+	ModuleRegistry    *moduleRegistry              // Where to register created modules.
+	IsModule          bool                         // Whether this controller is for a module.
```
ptodev, line 162: Minor personal preference: I'd prefer if this is an enum which indicates what the controller is for. It'd answer the question of "what is this, if it's not a module controller?". E.g. it'd be clear that this is either a `ControllerModule` vs `ControllerRoot`). But I'm ok with this staying as it is.

ptodev (commented):
LGTM

PR #4691, opened 2023-08-02
Fix GitHub app

mattdurham:
Grafanabot must have had god rights to directly push to deployment tools, so now based on some other projects, added details to automatically open a repo.


erikbaranowski (approved):
LGTM

PR #4699, opened 2023-08-03
flow: fix bug where GetServiceConsumers didn't search through loaded modules

rfratto:
GetServiceConsumers will be used by services to quickly retrieve a list of components and services which depend on a service for service-related operations.

For example, the clustering service will use GetServiceConsumers to know which components to notify of a cluster update.

GetServiceConsumers needs to return dependant components throughout all modules, otherwise the clustering service will be unable to notify all appropriate components when the clustering state changes.

### DIFF
--- a/pkg/flow/flow_services_test.go
+++ b/pkg/flow/flow_services_test.go
@@ -170,6 +170,76 @@ func TestFlow_GetServiceConsumers(t *testing.T) {
 	require.Equal(t, []any{svcB}, consumers)
 }
 
+func TestFlow_GetServiceConsumers_Modules(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	componentBuilt := util.NewWaitTrigger()
+
+	var (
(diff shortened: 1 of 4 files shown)


pkg/flow/flow_services_test.go
```diff
@@ -170,6 +170,76 @@ func TestFlow_GetServiceConsumers(t *testing.T) {
 	require.Equal(t, []any{svcB}, consumers)
 }
 
+func TestFlow_GetServiceConsumers_Modules(t *testing.T) {
```
ptodev, line 177: Do we need to test more complicated situations too - e.g. a module A, which imports a module B, where module B is the one using the service?

ptodev (approved):

pkg/flow/flow_services_test.go
rfratto: We could, but I'm not sure it's necessary. There's not much difference in the code between one layer of modules and many layers of modules.

tpaschalis (approved):
👌

PR #4700, opened 2023-08-03
Porting decompression support from promtail to loki.source.file

### LINKED ISSUE #4669: Ability to tail compressed files (like promtail does)
### Request

Promtail currently supports reading logs from compressed files:
```
# Defines decompression behavior for the given scrape target.
decompression:
  # Whether decompression should be tried or not.
  [enabled: <boolean> | default = false]

  # Initial delay to wait before starting the decompression.
  # Especially useful in scenarios where compressed files are found before the compression is finished.
  [initial_delay: <duration> | default = 0s]

  # Compression format. Supported formats are: 'gz', 'bz2' and 'z.
  [format: <string> | default = ""]
```

We should add the same functionality to `loki.source.file`.

### Use case

* I want to read logs from a compressed file
* I want to migrate from promtail to agent flow

thampiotr:
#### PR Description

This adds a decompression support in `loki.source.file` that matches the current behaviour of `promtail`.
It is a breaking change, since it was also a breaking change in `promtail`.

Fixes #4669 

#### Notes to the Reviewer

The PR that made equivalent change in promtail is this: https://github.com/grafana/loki/pull/8994

Tested this locally using `gzip`:
```
➜  /tmp echo hello_zip > foo.txt
➜  /tmp gzip foo.txt
```

Used the following River config
```
local.file_match "logs" {
  path_targets = [
    {__path__ = "/tmp/*.gz"},
  ]
}

loki.source.file "tmpfiles" {
  targets    = local.file_match.logs.targets
  forward_to = [loki.echo.print.receiver]
  decompression {
    enabled       = true
    initial_delay = "10s"
    format        = "gz"
  }
}

loki.echo "print" {}
```

Observed these logs:
```
ts=2023-08-03T12:40:36.611169Z level=info controller_id= trace_id=ffae15ec97959d9fb70063116e1fe9aa msg="finished partial graph evaluation" duration=1.850875ms
ts=2023-08-03T12:40:36.611344Z level=info component=loki.source.file.tmpfiles component=decompressor msg="read lines routine: started" path=/tmp/foo.txt.gz
ts=2023-08-03T12:40:36.611353Z level=info component=loki.source.file.tmpfiles component=decompressor msg="sleeping before starting decompression" path=/tmp/foo.txt.gz duration=10s
ts=2023-08-03T12:40:46.613484Z level=info component=loki.source.file.tmpfiles component=decompressor msg="successfully mounted reader" path=/tmp/foo.txt.gz ext=.gz
ts=2023-08-03T12:40:46.613576Z level=info component=loki.source.file.tmpfiles component=decompressor msg="read lines routine finished" path=/tmp/foo.txt.gz
ts=2023-08-03T12:40:46.613633Z level=info component=loki.echo.print receiver=loki.echo.print entry=hello_zip labels="{filename=\"/tmp/foo.txt.gz\"}"
```

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/component/loki/source/file/decompresser.go
+++ b/component/loki/source/file/decompresser.go
@@ -117,10 +117,10 @@ func newDecompressor(
 //
 // The selected reader implementation is based on the extension of the given file name.
 // It'll error if the extension isn't supported.
-func mountReader(f *os.File, logger log.Logger, format string) (reader io.Reader, err error) {
+func mountReader(f *os.File, logger log.Logger, format CompressionFormat) (reader io.Reader, err error) {
 	var decompressLib string
 
-	switch format {
+	switch format.String() {
 	case "gz":
 		decompressLib = "compress/gzip"
 		reader, err = gzip.NewReader(f)
--- a/component/loki/source/file/decompresser.go
+++ b/component/loki/source/file/decompresser.go
@@ -115,8 +115,10 @@ func newDecompressor(
 
 // mountReader instantiate a reader ready to be used by the decompressor.
 //
-// The selected reader implementation is based on the extension of the given file name.
-// It'll error if the extension isn't supported.
+// The reader implementation is selected based on the given CompressionFormat.
+// If the actual file format is incorrect, the reading of the header may fail and return an error - depending on the
+// implementation of the underlying compression library. In any case, when a file is corrupted, the subsequent reading
+// of lines will fail.
 func mountReader(f *os.File, logger log.Logger, format CompressionFormat) (reader io.Reader, err error) {
 	var decompressLib string
 
--- a/component/loki/source/file/decompresser.go
+++ b/component/loki/source/file/decompresser.go
@@ -32,11 +32,10 @@ import (
 
 func supportedCompressedFormats() map[string]struct{} {
 	return map[string]struct{}{
-		".gz":     {},
-		".tar.gz": {},
-		".z":      {},
-		".bz2":    {},
-		// TODO: add support for .zip extension.
+		"gz":  {},
+		"z":   {},
+		"bz2": {},
+		// TODO: add support for zip.
 	}
 }
 
@@ -61,9 +60,20 @@ type decompressor struct {
 
 	position int64
 	size     int64
+	cfg      DecompressionConfig
 }
 
-func newDecompressor(metrics *metrics, logger log.Logger, handler loki.EntryHandler, positions positions.Positions, path string, labels string, encodingFormat string) (*decompressor, error) {
+func newDecompressor(
+	metrics *metrics,
+	logger log.Logger,
+	handler loki.EntryHandler,
+	positions positions.Positions,
+	path string,
+	labels string,
+	encodingFormat string,
+	cfg DecompressionConfig,
+) (*decompressor, error) {
+
 	logger = log.With(logger, "component", "decompressor")
 
 	pos, err := positions.Get(path, labels)
@@ -94,6 +104,7 @@ func newDecompressor(metrics *metrics, logger log.Logger, handler loki.EntryHand
 		done:      make(chan struct{}),
 		position:  pos,
 		decoder:   decoder,
+		cfg:       cfg,
 	}
 
 	go decompressor.readLines()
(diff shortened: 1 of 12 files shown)


docs/sources/flow/reference/components/loki.source.file.md
```diff
@@ -107,4 +133,59 @@ loki.write "local" {
 }
 ```
 
+### File globbing
```
thampiotr, line 136: Based on questions from users, I found that it is useful to add this as an example too.

component/loki/source/file/file.go
```diff
@@ -36,9 +37,32 @@ const (
...
+	for sf := range supportedCompressedFormats() {
+		if d.Format == sf {
+			return nil
+		}
+	}
+	return fmt.Errorf(
+		"unsupported compression format: %q - please use one of %q",
+		d.Format,
+		strings.Join(maps.Keys(supportedCompressedFormats()), ", "),
+	)
```
thampiotr, line 65: Added this to eagerly find errors and give feedback:
image

component/loki/source/file/file.go
```diff
@@ -36,9 +37,32 @@ const (
...
+	Targets             []discovery.Target  `river:"targets,attr"`
+	ForwardTo           []loki.LogsReceiver `river:"forward_to,attr"`
+	Encoding            string              `river:"encoding,attr,optional"`
+	DecompressionConfig DecompressionConfig `river:"decompression,block,optional"`
+}
+
+type DecompressionConfig struct {
+	Enabled      bool          `river:"enabled,attr"`
+	InitialDelay time.Duration `river:"initial_delay,attr,optional"`
+	Format       string        `river:"format,attr"`
```
rfratto, line 49: One alternative way you could do this is by introducing a new type here for the format which implements encoding.TextMarshaler/encoding.TextUnmarshaler for validation. 

This would mean that if someone provides an invalid format, just the `format` attribute is highlighted as being a problem instead of the surrounding block. 

This is used in local.file for the `detector` attribute: https://github.com/grafana/agent/blob/9756a4ccc08fa33ab7006fed9fe5dfb44c0dc8ca/component/local/file/detector.go#L15-L65

I don't think you _have_ to do this, and it tends to be a bit more code, but just wanted to point it out as an option.

rfratto: > So maybe we should consider allowing the Validate to provide a name of attribute or block that is a problem?

I'd definitely be interested in something like this.

docs/sources/flow/reference/components/loki.source.file.md
```diff
@@ -26,19 +26,44 @@ log entries to the list of receivers passed in `forward_to`.
...
+
+ Name            | Type       | Description                                                     | Default | Required 
+-----------------|------------|-----------------------------------------------------------------|---------|----------
+ `enabled`       | `bool`     | Whether decompression is enabled.                               |         | yes      
+ `initial_delay` | `duration` | Time to wait before starting to read from new compressed files. | 0       | no       
+ `format`        | `string`   | Compression format.                                             |         | yes      
+
+If you compress a file under a folder being scraped, `loki.source.file` might
+try to ingest your file before you finish compressing it. To avoid it, pick
+an `initial_delay` that is enough to avoid it.
```
tpaschalis, line 61: Since we cannot specify multiple formats, should we explicitly mention that to handle different compressions, users should add a new component?

docs/sources/flow/reference/components/loki.source.file.md
```diff
@@ -26,19 +26,44 @@ log entries to the list of receivers passed in `forward_to`.
...
+ Hierarchy      | Name               | Description                                   | Required 
+----------------|--------------------|-----------------------------------------------|----------
+ decompresssion | [decompresssion][] | Configure reading logs from compressed files. | no       
+
+[decompresssion]: #decompresssion-block
+
+### decompresssion block
+
+The `decompression` block contains configuration for reading logs from 
+compressed files. The following arguments are supported:
```
tpaschalis, line 51:
```suggestion
The `decompression` block contains configuration for reading logs from 
compressed files found in its `targets`. The following arguments are supported:
```

Possible addition so that it doesn't read that if this is enabled, the component _only_ reads compressed files. WDYT? Any better wording?

component/loki/source/file/decompresser.go
```diff
@@ -106,37 +117,35 @@ func newDecompressor(metrics *metrics, logger log.Logger, handler loki.EntryHand
 //
 // The selected reader implementation is based on the extension of the given file name.
 // It'll error if the extension isn't supported.
```
tpaschalis, line 119: This docstring is not accurate now, let's change this. Maybe mention what happens to files that cannot be decoded correctly?

tpaschalis (approved):
LGTM % some docs-related nits, thanks!

docs/sources/flow/reference/components/loki.source.file.md
thampiotr: > so that it doesn't read that if this is enabled, the component only reads compressed files

The component indeed will _only read from compressed files_, trying to decompress all the targets. I think it could be improved, but I wouldn't want to divert from promtails implementation right now.

tpaschalis: Oh, I see. I'm not particularly fond of that decision, but yeah, let's keep in line with Promtail for now.

PR #4705, opened 2023-08-04
YACE dimension_name_requirements support

### LINKED ISSUE #4381: Add support for dimensionNameRequirements in YACE configuration
### Request

YACE (which grafana agent wraps for Cloudwatch statistics collection) support filtering of the available dimensions for a each service type by the configuration item dimensionNameRequirements (see https://github.com/nerdswords/yet-another-cloudwatch-exporter/blob/master/docs/configuration.md#discovery_job_config). Currently, there is no way to specify this in discovery job configuration at https://grafana.com/docs/agent/latest/static/configuration/integrations/cloudwatch-exporter-config/#discovery_job (which would be the most granular level in configuration).

This is valuable, because AWS charges based on the number of metrics returned - and there are a sizeable number of metrics collected for dimensions that aren't generally interesting. For example, in our test account, looking just at RDS instances, we get metrics aggregated by Engine (eg Postgres/Oracle/Aurora) and Class (eg r5.2xlarge, ...). 

image

For the moment, we are throwing away the dimensions we aren't interested in using metric_relabel_configs to "drop" them which at least avoids the storage cost on the Grafana side, but not collecting them in the first place would be better and avoid uneccesary AW$ spend:

```
    metric_relabel_configs:
      - source_labels: [dimension_DatabaseClass, dimension_EngineName, dimension_SourceRegion]
        regex: ".+"
        action: drop
```

### Use case

Avoid collecting Cloudwatch metrics in dimensions that aren't used.

CVdV-au:

Fixes #4381

#### Notes to the Reviewer

Simple config passthrough

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Documentation added
- [X] Tests updated
- [X] Config converters updated


marctc (commented):
You added the changes to flow component and documented the field in the static integration :smile: Could you add documentation for flow component and add the field to the integration? thanks

marctc (approved):

clayton-cornell (approved):
Docs are OK as-is

PR #4706, opened 2023-08-04
Update otelcol.exporter.otlphttp.md

albertored:
`client.endpoint` should contain protocol, if not given the agent does not start


clayton-cornell:
Can someone from @grafana/grafana-agent-maintainers take a quick look at this doc change to verify for technical accuracy. If it's all good, I can get this one merged right away.

docs/sources/flow/reference/components/otelcol.exporter.otlphttp.md
```diff
@@ -67,7 +67,7 @@ The following arguments are supported:
 
 Name | Type | Description | Default | Required
 ---- | ---- | ----------- | ------- | --------
-`endpoint`           | `string`      | `host:port` to send telemetry data to. | | yes
+`endpoint`           | `string`      | `protocol://host:port` to send telemetry data to. | | yes
```
thampiotr, line 70: This will be translated into this property of otel collector: https://github.com/open-telemetry/opentelemetry-collector/blob/v0.80.0/config/confighttp/confighttp.go#L32

So I think we should document it as URL (scheme but also path).

There's a similar issue in `docs/sources/flow/reference/components/otelcol.exporter.otlp.md` which we should also fix.

clayton-cornell: We should use `for example` instead of `e.g.`  Latin abbreviations should be avoided where possible. We shoudl also wrap the URL with code syntax to make it clear (to noth the doc generation tooling and the reader that this is an example and not some clickable URL

> The target URL to send data to, for example, `http://some.url:9411/v1/push`.

ptodev: Giving examples in these tables is not something we normally do. IMO we should leave the table like this:
```
`endpoint` | `string` | `protocol://host:port` to send telemetry data to. | | yes
```

However, the "Example" section below needs to be updated.

thampiotr: If we say `protocol://host:port`, it will be confusing when users later find they need to provide full URL with the resource path, for example `protocol://host:port/some/path`. I'd prefer we call it URL and have examples that include the path component.

ptodev: @thampiotr sounds good! @albertored would you mind changing the table row to:
```suggestion
`endpoint`           | `string`      | The target URL to send telemetry data to. | | yes
```
And then also changing the "Example" section to include an example?

If you'd like to emphasise that the "protocol" has to be included, then you could add a sentence like this to the `### client block` section:

> `endpoint` must begin with a protocol such as "http://".

albertored: Perfect, will do in the next days

albertored: @ptodev done

clayton-cornell (approved):
LGTM

PR #4713, opened 2023-08-04
Add an `otelcol.connector.spanlogs` flow component

### LINKED ISSUE #4261: Flow: Alternative to Static mode's tracing automatic_logging processor
### Request

The [traces](https://grafana.com/docs/agent/latest/static/configuration/traces-config/) config in Static mode contains an `automatic_logging` processor which can create log lines from traces. This is similar to [otelcol.exporter.logging](https://grafana.com/docs/agent/latest/flow/reference/components/otelcol.exporter.logging/), but more flexible in some ways:

- The log output could be sent to Loki, whereas `otelcol.exporter.logging` can only write to console
- `automatic_logging` could log a line per span, root, or process.
- Overrides for Loki log labels

Additionally, it was [reported](https://github.com/grafana/agent/issues/2213#issuecomment-1578937214) that viewing output of `otelcol.exporter.logging` can be awkward since it outputs `\n` inside the message, which [forces](https://github.com/grafana/agent/issues/2213#issuecomment-1582253769) people to paste the message in an editor and replace the line breaks.

I am not sure how the `automatic_logging` processor is used by customers and how important all those features are. If we want to integrate with Loki or to have the additional filtering on span/root/process, then we could create a new component. The new component could output Otlp-formatted logs, and then this could be piped to `otelcol.exporter.loki` and `loki.write`?

This new component would be a bit like Otel's [spanmetrics](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/connector/spanmetricsconnector/README.md) connector, but instead of metrics it would create logs out of the traces. AFAIK there is no such component planned for the Collector yet.

### Use case

Flow feature parity.

ptodev:
#### PR Description

This is the flow mode equivalent to static mode's `automatic_logging` processor.

Fixes #4261

#### Notes to the Reviewer

* Are we sure `otelcol.processor.spanlogs` is a good name for this?
* Initially I wanted to reuse as much code as possible between static and flow modes. Later I gave up on this. I don't think it's very important.
* The usage of `[]any` in the original code made it look a bit weird. Since this is a purely otel component which doesn't interact with Loki, I decided to use Otel's `pcommon.Map`. This is the data structure which span attributes are normally stored in. However, this data structure is inefficient when reading and writing to it, since it's actually an array and we need to traverse it on every insertion. Maybe I should switch to a Go map?

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


clayton-cornell (commented):
No changes for docs.

component/otelcol/processor/spanlogs/consumer.go
```diff
@@ -0,0 +1,259 @@
...
+		return false
+	})
+
+	return &res, nil
+}
+
+func (c *consumer) processKeyVals(output pcommon.Map, resource pcommon.Resource, svc string) {
+	rsAtts := resource.Attributes()
+
+	// name
```
mattdurham, line 221: I dont understand this comment

component/otelcol/processor/spanlogs/consumer.go
```diff
@@ -0,0 +1,259 @@
...
+
+	// Add the attributes
+	logAttributes := res.Attributes()
+
+	// Add logs instance label
+	logAttributes.PutStr(c.opts.overrides.LogsTag, kind)
+
+	copyOtelMap(logAttributes, keyValues, func(key string, val pcommon.Value) bool {
+		// Check if we have to include this label
+		if _, ok := c.opts.labels[key]; ok {
```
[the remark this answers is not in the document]
ptodev: Yes, thank you, I like this idea!

docs/sources/flow/reference/components/otelcol.processor.spanlogs.md
```diff
@@ -0,0 +1,234 @@
...
+canonical: https://grafana.com/docs/agent/latest/flow/reference/components/otelcol.processor.spanlogs/
+title: otelcol.processor.spanlogs
+---
+
+# otelcol.processor.spanlogs
+
+`otelcol.processor.spanlogs` accepts traces telemetry data from other `otelcol`
+components and outputs logs telemetry data for each span, root, or process.
+This allows for automatically building a mechanism for trace
+discovery.
```
clayton-cornell, line 11:
```suggestion
This allows you to automatically build a mechanism for trace discovery.
```

docs/sources/flow/reference/components/otelcol.processor.spanlogs.md
```diff
@@ -0,0 +1,234 @@
...
+
+> **NOTE**: `otelcol.processor.spanlogs` is a custom component unrelated 
+> to any processors from the OpenTelemetry Collector. It is based on the 
+> `automatic_logging` processor in the 
+> [traces](../../../static/configuration/traces-config.md) 
+> subsystem of the Agent static mode.
+
+Multiple `otelcol.processor.spanlogs` components can be specified by giving them
```
clayton-cornell, line 19:
```suggestion
You can specify multiple `otelcol.processor.spanlogs` components by giving them
```

docs/sources/flow/reference/components/otelcol.processor.spanlogs.md
```diff
@@ -0,0 +1,234 @@
...
+`otelcol.processor.spanlogs` supports the following arguments:
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`spans` | `bool` | Log one line per span. | `false` | no
+`roots` | `bool` | Log one line for every root span of a trace. | `false` | no
+`processes` | `bool` | Log one line for every process. | `false` | no
+`span_attributes` | `list(string)` | Additional span attributes to log. | `[]` | no
+`process_attributes` | `list(string)` | Additional process attributes to log. | `[]` | no
+`labels` | `list(string)` | A lists of keys which will be logged as labels. | `[]` | no
```
clayton-cornell, line 43:
```suggestion
`labels` | `list(string)` | A list of keys that will be logged as labels. | `[]` | no
```

docs/sources/flow/reference/components/otelcol.processor.spanlogs.md
```diff
@@ -0,0 +1,234 @@
...
+--------- | ----- | ----------- | --------
+overrides | [overrides][] | Overrides for keys in the log body. | no
+output | [output][] | Configures where to send received telemetry data. | yes
+
+[output]: #output-block
+[overrides]: #overrides-block
+
+### overrides block
+
+The `overrides` block configures overrides for keys which will be logged in the body of the log line.
```
clayton-cornell, line 64:
```suggestion
The `overrides` block configures overrides for keys that will be logged in the body of the log line.
```

docs/sources/flow/reference/components/otelcol.processor.spanlogs.md
```diff
@@ -0,0 +1,234 @@
...
+
+The following attributes are supported:
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`logs_instance_tag` | `string` | Indicates if the log line is for a span, root or process. | `traces` | no
```
clayton-cornell, line 70:
```suggestion
`logs_instance_tag` | `string` | Indicates if the log line is for a span, root, or process. | `traces` | no
```

docs/sources/flow/reference/components/otelcol.processor.spanlogs.md
```diff
@@ -0,0 +1,234 @@
...
+configuration.
+
+## Debug information
+
+`otelcol.processor.spanlogs` does not expose any component-specific debug
+information.
+
+## Example
+
+The configuration below sends logs derived from spans to Loki.
```
clayton-cornell, line 104:
```suggestion
The following configuration sends logs derived from spans to Loki.
```

clayton-cornell (commented):
Some tweaks to the docs

component/otelcol/processor/spanlogs/consumer.go
```diff
@@ -0,0 +1,259 @@
...
+				//TODO: Use a Go map instead of pcommon.NewMap, and convert to pcommon.NewMap only at the very end?
+				//      Unfortunately pcommon.NewMap can be inefficient during insertions,
+				//      because it traverses the whole map on every insertion.
+				if c.opts.spans {
+					keyValues := pcommon.NewMap()
+
+					c.spanKeyVals(keyValues, span)
+					c.processKeyVals(keyValues, rs.Resource(), svc)
+
+					newLogRecord, err := c.createLogRecord(typeSpan, traceID, keyValues)
```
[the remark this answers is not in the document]
ptodev: Yes, I refactored it now, PTAL.

component/otelcol/connector/spanlogs/consumer.go
```diff
@@ -0,0 +1,282 @@
...
+		if pred(k, v) {
+			newVal := out.PutEmpty(k)
+			v.CopyTo(newVal)
+		}
+		return true
+	})
+}
+
+func convertOtelMapToSlice(m pcommon.Map) []any {
+	res := make([]any, 0, m.Len())
```
mattdurham, line 189: Should `m.len()` be `m.len() * 2`? Since we are flattening the map?

component/otelcol/connector/spanlogs/consumer.go
```diff
@@ -0,0 +1,282 @@
...
+	rsAtts := resource.Attributes()
+
+	// Add an attribute with the service name
+	output.PutStr(c.opts.overrides.ServiceKey, svc)
+
+	for _, name := range c.opts.processAttributes {
+		att, ok := rsAtts.Get(name)
+		if ok {
+			// name/key val pairs
+			val := output.PutEmpty(name)
```
mattdurham, line 251: Not sure what we are doing here. So if we find a value in `rsAtts`, we put an empty value into the output and then copy the empty value to att?

component/otelcol/connector/spanlogs/spanlogs.go
```diff
@@ -0,0 +1,129 @@
...
+type Component struct {
+	consumer *consumer
+}
+
+var _ component.Component = (*Component)(nil)
+
+// New creates a new otelcol.exporter.spanlogs component.
+func New(o component.Options, c Arguments) (*Component, error) {
+	if c.Output.Traces != nil || c.Output.Metrics != nil {
+		level.Warn(o.Logger).Log("msg", "non-log output detected; this component only works for log outputs and trace inputs")
```
mattdurham, line 82: Is this true? Why is Metrics checked if logs are the only output?

mattdurham (commented):
A few comments but looking good.

component/otelcol/connector/spanlogs/consumer.go
ptodev: Yeah. It does look weird. The issue is that [pcommon.Map](https://github.com/open-telemetry/opentelemetry-collector/blob/main/pdata/pcommon/map.go) doesn't have a function to insert a `Value` and a `Key` at the same time, like a `PutValue(k string, v Value)`. These kinds of functions only insert specific types like `PutInt(k string, v int64)`.

I want to insert a Value so that I preserve the original type, and this is the only way to do it as far as I can tell.

In hindsight, I should have just used a `map[string]pcommon.Value` instead of a `pcommon.Map`. It would also have been more efficient, since a `pcommon.Map` is actually just an slice. Inserting to it is slow, because it has to traverse the whole slice to see if the element is already in the map.

I'll add a TODO in the code for this, but for now I hope it's fast enough that a `pcommon.Map` will do.

mattdurham (approved):
Lgtm

PR #4736, opened 2023-08-07
remove add to project

mattdurham:
The project workflow itself has this ability built in so I dont think we need this. Tagging @rfratto to get his eyes on it to double check that assumption. Tested this by creating a few dummy issues and they were auto added to the project.

https://github.com/orgs/grafana/projects/269/workflows/11023218


mattdurham:
I dont think so, or at least can be pushed down the road.

PR #4753, opened 2023-08-08
Prototype of auto-generated compatible components section in docs

thampiotr:
#### PR Description

This is a prototype for the proposal https://github.com/grafana/agent/issues/4754
Please see the proposal for detailed information.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


component/metadata/metadata.go
```diff
@@ -0,0 +1,110 @@
...
+
+	// DataTypeLokiLogs represent logs in Loki format
+	DataTypeLokiLogs = DataType("Loki Logs")
+
+	DataTypeOTELTelemetry     = DataType("OTEL Telemetry")
+	DataTypePromMetrics       = DataType("Prometheus Metrics")
+	DataTypePyroscopeProfiles = DataType("Pyroscope Profiles")
+)
+
+type Metadata struct {
```
thampiotr, line 29: Happy to rename this

component/metadata/metadata.go
```diff
@@ -0,0 +1,110 @@
...
+
+func ForComponent(name string) (Metadata, error) {
+	reg, ok := component.Get(name)
+	if !ok {
+		return Metadata{}, fmt.Errorf("could not find component %q", name)
+	}
+	return inferMetadata(reg.Args, reg.Exports), nil
+}
+
+func inferMetadata(args component.Arguments, exports component.Exports) Metadata {
```
thampiotr, line 46: Support for more types can be added in this function, which is also easy to test and make sure we don't get unexpected relationships.

docs/docs_updated_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/grafana/agent/docs/generator"
+	"github.com/stretchr/testify/require"
+)
+
+// Run the below generate command to automatically update the Markdown docs with generated content
+//go:generate go test -run TestCompatibleComponentsSectionUpdated -fix-tests
+
+var fixTestsFlag = flag.Bool("fix-tests", false, "update the test files with the current generated content")
+
+func TestCompatibleComponentsSectionUpdated(t *testing.T) {
```
thampiotr, line 19: If someone forgets to update the docs, CI will fail. But updating them is also as easy as running `go generate` for this file.

docs/sources/flow/reference/components/loki.echo.md
```diff
@@ -41,6 +41,38 @@ Name | Type | Description
 
 `loki.echo` does not expose any component-specific debug information.
 
+<!-- START GENERATED COMPATIBLE COMPONENTS -->
```
thampiotr, line 44: Here just for fun I put the section in a different place rather than at the end. The auto-updating will add the section at the end if it doesn't exist, but it can be moved around and tests will still detect it and update it in its existing location.

PR #4761, opened 2023-08-08
Update stability to match Grafana standards

clayton-cornell:
#### PR Description

The introductory information in the Agent documentation refers to obsolete release terminology. The only release stages we should be referring to are documented here: https://grafana.com/docs/release-life-cycle/

This PR does the following:

- Deletes the obsolete Stability topic.
- Updates the Stability table in the Introduction topic and adds a reference to the Release life cycle topic.
- Updates the shared headless Experimental topic embedded link to redirect to the Release life cycle topic.
- Updates the shared headless Beta topic to define it at Public preview and updates the embedded link to redirect to the Release life cycle topic.
- A note admonition syntax fix is also included in this PR.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


docs/sources/about.md
```diff
@@ -23,18 +23,21 @@ Grafana Agent is available in three different variants:
...
 
-| Project | Stability |
+| Project | Release stage |
 | ------- | --------- |
-| Static mode | Stable |
-| Static mode Kubernetes operator | Beta |
-| Flow mode | Stable |
+| Static mode | Public preview |
+| Static mode Kubernetes operator | Experimental |
+| Flow mode | Public preview |
```
rfratto, line 32: I'm worried about this because it might seem like we're downgrading the stability of everything. 

Should we wait until 1.0 to synchronize terminology?

clayton-cornell: Do you mean with the change from Beta to Experimental for Operator? Would it be better to just tag all as Public preview?

rfratto: I just don't think public preview is an appropriate term for the project in its current state, but neither is Generally Available (since it is not 'fully supported' by the support team).

Quoting the release life cycle document:

> To drive home my concerns, the release cycle document says this: 

> Products and features in public preview are not meant to be used in production environments, and the risks are medium/low.

I'm very concerned that saying that Static mode and Flow mode are "not meant to be used in production environments" will cause confusion and worry from our existing used.

Eve832: My $0.02 - if we don't have comprehensive support for the feature, we can't recommend that folks start using it in production...we're already seeing Cloud user confusion

clayton-cornell:
@rfratto @Eve832 Just following up on  this PR.  I'll close the PR and we will revisit this at (or very close to) later this year when this change makes more sense.

PR #4765, opened 2023-08-09
Add configurable fields loki source cloudflare

### LINKED ISSUE #4760: `loki.source.cloudflare` requested fields should be configurable
### Request

Currently in [loki.source.cloudflare](https://grafana.com/docs/agent/latest/flow/reference/components/loki.source.cloudflare/#arguments) the requested fields can only be one of the [4 pre-defined subsets](https://github.com/grafana/agent/blob/7e4512eb70a174b8fdba9cfdf2fb5171b9c080ec/component/loki/source/cloudflare/internal/cloudflaretarget/fields.go#L22).

We should make it possible to define your own subset of fields, while keeping the presets as they can be useful.

### Use case

When I want to limit number of fields that I get from Cloudflare, keeping only the ones that I need to lower the cost and verbosity.

wildum:
#### PR Description
Add the property `custom_fields` and the type "custom" to the property `fields_type` to allow the user to define its own field for the loki.source.cloudflare. 
In the case where the `fields_type` is set to "all", then the `custom_fields` is expected to be empty or not set.
In the case where the `fields_type` is set to something else than "all" or "custom", the `custom_fields` will be merged with the one provided by the `fields_type` and the duplicates (if any) will be removed.

Fixes #4760 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/component/loki/source/cloudflare/cloudflare.go
+++ b/component/loki/source/cloudflare/cloudflare.go
@@ -75,38 +75,20 @@ func (c *Arguments) Validate() error {
 	if c.PullRange < 0 {
 		return fmt.Errorf("pull_range must be a positive duration")
 	}
-	fields, err := cft.Fields(cft.FieldsType(c.FieldsType))
+	_, err := cft.Fields(cft.FieldsType(c.FieldsType), c.CustomFields)
 	if err != nil {
 		return fmt.Errorf("invalid fields_type set; the available values are 'default', 'minimal', 'extended', 'custom' and 'all'")
+	}
+
+	if c.FieldsType == "all" && len(c.CustomFields) > 0 {
+		return fmt.Errorf("no need to provide custom fields when fields_type is set to 'all' as it contains already all available fields")
 	}
 
 	if invalidFields := cft.FindInvalidFields(c.CustomFields); len(invalidFields) > 0 {
 		return fmt.Errorf("invalid custom_fields: %v", invalidFields)
 	}
 
-	if commonFields := findCommonElement(fields, c.CustomFields); len(commonFields) > 0 {
-		return fmt.Errorf("duplicated fields; the following fields are already part of %s: %v", c.FieldsType, commonFields)
-	}
-
 	return nil
-}
-
-// Find the common values between two slices
-func findCommonElement(slice1, slice2 []string) []string {
-	elements := make(map[string]bool)
-	var common []string
-
-	for _, element := range slice1 {
-		elements[element] = true
-	}
-
-	for _, element := range slice2 {
-		if elements[element] {
-			common = append(common, element)
-		}
-	}
-
-	return common
 }
 
 // Component implements the loki.source.cloudflare component.
--- a/component/loki/source/cloudflare/cloudflare.go
+++ b/component/loki/source/cloudflare/cloudflare.go
@@ -80,7 +80,7 @@ func (c *Arguments) Validate() error {
 		return fmt.Errorf("invalid fields_type set; the available values are 'default', 'minimal', 'extended', 'custom' and 'all'")
 	}
 
-	if c.FieldsType == "all" && len(c.CustomFields) > 0 {
+	if cft.FieldsType(c.FieldsType) == cft.FieldsTypeAll && len(c.CustomFields) > 0 {
 		return fmt.Errorf("no need to provide custom fields when fields_type is set to 'all' as it contains already all available fields")
 	}
 
--- a/component/loki/source/cloudflare/cloudflare.go
+++ b/component/loki/source/cloudflare/cloudflare.go
@@ -31,13 +31,14 @@ func init() {
 // Arguments holds values which are used to configure the
 // loki.source.cloudflare component.
 type Arguments struct {
-	APIToken   rivertypes.Secret   `river:"api_token,attr"`
-	ZoneID     string              `river:"zone_id,attr"`
-	Labels     map[string]string   `river:"labels,attr,optional"`
-	Workers    int                 `river:"workers,attr,optional"`
-	PullRange  time.Duration       `river:"pull_range,attr,optional"`
-	FieldsType string              `river:"fields_type,attr,optional"`
-	ForwardTo  []loki.LogsReceiver `river:"forward_to,attr"`
+	APIToken     rivertypes.Secret   `river:"api_token,attr"`
+	ZoneID       string              `river:"zone_id,attr"`
+	Labels       map[string]string   `river:"labels,attr,optional"`
+	Workers      int                 `river:"workers,attr,optional"`
+	PullRange    time.Duration       `river:"pull_range,attr,optional"`
+	FieldsType   string              `river:"fields_type,attr,optional"`
+	CustomFields []string            `river:"custom_fields,attr,optional"`
+	ForwardTo    []loki.LogsReceiver `river:"forward_to,attr"`
 }
 
 // Convert returns a cloudflaretarget Config struct from the Arguments.
@@ -47,12 +48,13 @@ func (c Arguments) Convert() *cft.Config {
 		lbls[model.LabelName(k)] = model.LabelValue(v)
 	}
 	return &cft.Config{
-		APIToken:   string(c.APIToken),
-		ZoneID:     c.ZoneID,
-		Labels:     lbls,
-		Workers:    c.Workers,
-		PullRange:  model.Duration(c.PullRange),
-		FieldsType: c.FieldsType,
+		APIToken:     string(c.APIToken),
+		ZoneID:       c.ZoneID,
+		Labels:       lbls,
+		Workers:      c.Workers,
+		PullRange:    model.Duration(c.PullRange),
+		FieldsType:   c.FieldsType,
+		CustomFields: c.CustomFields,
 	}
 }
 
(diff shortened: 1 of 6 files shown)


component/loki/source/cloudflare/internal/cloudflaretarget/fields.go
```diff
@@ -39,10 +40,38 @@ var (
 		"FirewallMatchesActions", "FirewallMatchesRuleIDs", "OriginResponseBytes", "OriginResponseTime", "ClientDeviceType", "WAFFlags", "WAFMatchedVar", "EdgeColoID",
 		"RequestHeaders", "ResponseHeaders",
 	}...)
+	allFieldsMap = buildAllFieldsMap(allFields)
 )
 
-// Fields returns the mapping of FieldsType to the set of fields it represents.
-func Fields(t FieldsType) ([]string, error) {
+// Fields returns the concatenation of set of fields represented by the Fieldtype and the give custom fields without duplicates.
```
thampiotr, line 46:
```suggestion
// Fields returns the concatenation of set of fields represented by the Fieldtype and the given custom fields without duplicates.
```

component/loki/source/cloudflare/cloudflare.go
```diff
@@ -73,10 +75,19 @@ func (c *Arguments) Validate() error {
...
-	_, err := cft.Fields(cft.FieldsType(c.FieldsType))
+	_, err := cft.Fields(cft.FieldsType(c.FieldsType), c.CustomFields)
 	if err != nil {
-		return fmt.Errorf("invalid fields_type set; the available values are 'default', 'minimal', 'extended' and 'all'")
+		return fmt.Errorf("invalid fields_type set; the available values are 'default', 'minimal', 'extended', 'custom' and 'all'")
 	}
+
+	if cft.FieldsType(c.FieldsType) == cft.FieldsTypeAll && len(c.CustomFields) > 0 {
+		return fmt.Errorf("no need to provide custom fields when fields_type is set to 'all' as it contains already all available fields")
+	}
```
thampiotr, line 85: Nit: I think we can remove this check, as it is not harmful to the user and we know how to do the right thing (use all fields). There may be other opinions on this though ;)

component/loki/source/cloudflare/internal/cloudflaretarget/fields.go
```diff
@@ -39,10 +40,38 @@ var (
...
+func Fields(t FieldsType, customFields []string) ([]string, error) {
+	fieldsSubset, err := getFieldSubset(t)
+	if err != nil {
+		return nil, err
+	}
+	return mergeAndRemoveDuplicatedFields(fieldsSubset, customFields), nil
+}
+
+func mergeAndRemoveDuplicatedFields(fieldsSubset, customFields []string) []string {
```
thampiotr, line 55: Nit - a bit shorter name
```suggestion
func mergeAndRemoveDuplicates(fieldsSubset, customFields []string) []string {
```

component/loki/source/cloudflare/internal/cloudflaretarget/fields.go
```diff
@@ -39,10 +40,38 @@ var (
...
+	usedFields := make(map[string]bool)
```
thampiotr, line 56: Here, too, it would be more idiomatic to use `struct{}` instead of `bool`. One of the quirks of Go.

component/loki/source/cloudflare/internal/cloudflaretarget/fields_test.go
```diff
@@ -0,0 +1,56 @@
...
+		t.Run(test.name, func(t *testing.T) {
+			result, err := Fields(test.fieldsType, test.customFields)
+			assert.NoError(t, err)
+			assert.ElementsMatch(t, test.expected, result)
+		})
+	}
+}
+
+func TestFindInvalidFields(t *testing.T) {
+	invalidFields := []string{"InvalidField1", "InvalidField2"}
```
thampiotr, line 52: Should probably add some valid fields there too.

component/loki/source/cloudflare/internal/cloudflaretarget/target.go
thampiotr, line 1: General note: this file is copied from `promtail` - it's okay to change it, but we generally would like to keep the changes here and in `fields.go` to the minimum, so that we can more easily merge with upstream changes from `promtail` if there are any in the future. 

I think that it's unavoidable to make changes for this feature here, so your PR is on a good path - if you have some ideas how to make it easier to merge changes in the future (e.g. mostly add new functions instead of changing existing ones), then it may be useful to do that.

thampiotr (commented):
Looking pretty good! some comments, but nothing major

component/loki/source/cloudflare/cloudflare.go
```diff
@@ -73,10 +75,15 @@ func (c *Arguments) Validate() error {
...
-	_, err := cft.Fields(cft.FieldsType(c.FieldsType))
+	_, err := cft.Fields(cft.FieldsType(c.FieldsType), c.AdditionalFields)
 	if err != nil {
-		return fmt.Errorf("invalid fields_type set; the available values are 'default', 'minimal', 'extended' and 'all'")
+		return fmt.Errorf("invalid fields_type set; the available values are 'default', 'minimal', 'extended', 'custom' and 'all'")
 	}
+
+	if invalidFields := cft.FindInvalidFields(c.AdditionalFields); len(invalidFields) > 0 {
+		return fmt.Errorf("invalid additional_fields: %v", invalidFields)
+	}
```
thampiotr, line 85: Thinking it through, it seems like the cloudflare client library doesn't validate fields against any list. I also see in documentation that the available fields list is provided by a live API: https://developers.cloudflare.com/logs/logpull/requesting-logs#fields 

So I think it's most future proof to remove validation here and always allow any custom fields - that way when new fields are added, we don't need to change it.

docs/sources/flow/reference/components/loki.source.cloudflare.md
```diff
@@ -39,11 +39,11 @@ Name            | Type                 | Description          | Default | Requir
...
 `pull_range`    | `duration`           | The timeframe to fetch for each pull request.      | `"1m"` | no
 `fields_type`   | `string`             | The set of fields to fetch for log entries.        | `"default"` | no
+`additional_fields` | `list(string)`       | The addional list of fields to supplement those provided via `fields_type`. |  | no
 
 
-By default `loki.source.cloudflare` fetches logs with the `default` set of
-fields. Here are the different sets of `fields_type` available for selection,
-and the fields they include:
+By default `loki.source.cloudflare` fetches logs with the `default` set of fields + the fields passed through `additional_fields`. 
+Here are the different sets of `fields_type` available for selection, and the fields they include:
```
thampiotr, line 46: I think below we should say something like:

```
* default includes: 
<list of fields> 
plus any extra fields provided via `additional_fields` argument.
```

And then revert line 45 to what it was before. 

Otherwise I get one version of what default means in line 45 and another in line 50.

docs/sources/flow/reference/components/loki.source.cloudflare.md
```diff
@@ -39,11 +39,11 @@ Name            | Type                 | Description          | Default | Requir
...
```
thampiotr, line 47: I know it was there before, but I've noticed that:
- Line 55 doesn't seem to be finished, missing closing " or maybe fields were cut?
- Line 65 has additional `k` at the end

Could you review if these are correct, matching the code?

component/loki/source/cloudflare/cloudflare.go
wildum: Right, but in this case we would lose consistency with the `allFields`. If you need to pick `allFields` and then add more in the additional_field to select all of them because `allFields` does not actually contains all valid fields, that would be confusing. Maybe one way to solve this would be to rename allFields to something else?

docs/sources/flow/reference/components/loki.source.cloudflare.md
clayton-cornell: Also we shouldn't use `+` unless it's a mathematical operator.

docs/sources/flow/reference/components/loki.source.cloudflare.md
clayton-cornell: Looks like 2 extra chars on line 65....`` `k``

clayton-cornell (commented):
Just 2 additions to existing comments.

component/loki/source/cloudflare/cloudflare.go
thampiotr: all fields is already outdated, e.g. it doesn't have BotDetectionIDs which is a valid field as per their docs: https://developers.cloudflare.com/logs/reference/log-fields/zone/http_requests
The reason I want to remove the validation is that customers can specifically start using these new fields without having to file an issue with Agent. We could rename all, but it would be a breaking change. I think the best we can do here is write in the docs something like this: 

> all fields type includes a snapshot of fields available at the time of writing this. If new fields are made available via Cloudflare API, you can get them by using the additiona_fields argument.

component/loki/source/cloudflare/internal/cloudflaretarget/fields.go
```diff
@@ -52,7 +81,29 @@ func Fields(t FieldsType) ([]string, error) {
 		return extendedFields, nil
 	case FieldsTypeAll:
 		return allFields, nil
+	case FieldsTypeCustom:
```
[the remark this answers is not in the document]
wildum: update: thanks to @thampiotr suggestion I updated the code. It looks more elegant now :)

component/loki/source/cloudflare/internal/cloudflaretarget/fields.go
```diff
@@ -41,18 +44,25 @@ var (
...
+		fields = append(minimalFields, additionalFields...)
 	case FieldsTypeExtended:
-		return extendedFields, nil
+		fields = append(extendedFields, additionalFields...)
 	case FieldsTypeAll:
-		return allFields, nil
+		fields = append(allFields, additionalFields...)
+	case FieldsTypeCustom:
+		fields = make([]string, len(additionalFields))
+		copy(fields, additionalFields)
```
thampiotr, line 61: You can make this into one line with
```suggestion
		fields = append(fields, additionalFields)
```
I would also be fine with returning just the `additionalFields` since it shouldn't be modified anywhere.

thampiotr (approved):
Just one nit, but LGTM! Good work 💯

component/loki/source/cloudflare/internal/cloudflaretarget/fields.go
wildum: still nice to clean up the duplicates. If a user has a long list of additional fields it could happen that they define twice the same field in the list by mistake

tpaschalis (approved):
LGTM 👌

PR #4769, opened 2023-08-09
Work around high cardinality metrics from otelcol receivers

### LINKED ISSUE #4764: High cardinality http_server_* metrics from `otelcol.receiver.zipkin`
### What's wrong?

We've recently migrated to agent flow mode and in that process migrated to a setup with a single dedicated agent for ingesting traces. We ingest traces both via OTLP grpc and zipkin (and sometimes jaeger). We noticed that the scrape job to scrape metrics from the dedicated traces agent started becoming heavy, and found that it was due to some very high cardinality metrics exported by the zipkin receiver.

Example:

```prometheus
http_server_duration_bucket{component_id="otelcol.receiver.zipkin.default",http_client_ip="10.129.128.11",http_flavor="1.1",http_method="POST",http_scheme="http",http_status_code="202",net_host_name="zipkin.monitoring.svc.cluster.local",net_sock_peer_addr="127.0.0.6",net_sock_peer_port="33389",otel_scope_name="go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp",otel_scope_version="0.42.0",le="0"} 0
``` 

Notice the `http_client_ip` and `net_sock_peer_port` labels, which quickly explode. This seems to be due to this upstream issue: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/3765

Even though we've configured our scrape job to drop all `http_server_*` metrics, the act of just parsing the `/metrics` endpoint gradually becomes unmanageable as it grows asymptotically. Just now I tested with curl and found it to be 232Mb on our traces agent.

I'm opening this issue with the hope that a workaround can be implemented in the grafana-agent until this issue has been fixed upstream.

### Steps to reproduce

Run the agent in flow mode with an `otelcol.receiver.zipkin` component and start ingesting zipkin traces, and see the `http_server_*` metrics exposed by the agent explode in cardinality.

### System information

Linux x86; GKE 1.24

### Software version

Grafana Agent v0.35.0

### Configuration

```text
otelcol.receiver.zipkin "default" {
	output {
		metrics = [otelcol.processor.memory_limiter.default.input]
		logs = [otelcol.processor.memory_limiter.default.input]
		traces = [otelcol.processor.memory_limiter.default.input]
	}
}
```

### Logs

_No response_

glindstedt:
#### PR Description

Heavily inspired by https://github.com/open-telemetry/opentelemetry-collector/pull/7543 which was mentioned in the upstream issue. This PR adds views to the `MeterProvider` of the receiver component which filters out some high cardinality attributes.

Fixes #4764

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated
- [x] Config converters updated (N/A)


glindstedt:
I suspect the Test failure was a flake? It succeeded on the commit before the changelog update

mattdurham:
Likely, the OSX and Win tests are unreliable at times. I restarted the tests.

ptodev:
Thank you for your contribution! I agree that we should do something about this problem. I'd have preferred if the Collector code avoids high cardinality metrics by default, but they have to be explicitly disabled via a feature gate. In any case, I believe we should be able to toggle this on and off in the Grafana Agent.

Would you be willing to try to implement a boolean flag called `OtelcolDisableHighCardinalityMetrics` inside a new `metrics` block, similar to the [logging](https://grafana.com/docs/agent/latest/flow/reference/config-blocks/logging/) and [tracing](https://grafana.com/docs/agent/latest/flow/reference/config-blocks/tracing/) blocks which we currently have? If you think this is too much work, I could try to do this myself and we will combine our works later. But I personally consider such a switch to be necessary for merging this change, so that we don't affect users who already depend on this functionality.

By the way, we will also need to document those metrics. I raised a separate issue for this - #4769 - and we could work on this in a separate PR.

glindstedt:
@ptodev Ok! Yeah I think I can take a stab at implementing that :+1:

glindstedt:
@ptodev I've been working on a PoC but ran into an issue. Now I have a `metrics` config block that creates a `metric.MeterProvider` which gets passed down to the otelcol component, the problem is that running the `Update` on only the `metrics` component and replacing the MeterProvider with a different configured one doesn't have any effect on the component that it got passed to originally. Is there a way to declare that a component is dependent on another so that an Update on the "parent" triggers updates on all the dependant components? Or in absence maybe a way to force re-evaluation of all components in case a core config like this is changed?

ptodev:
There is no need for the `metrics` block to own the `metric.MeterProvider`. It's ok to just have a new member in `Options` struct in `component/registry.go`. That new member could be an interface which has a function like `MeterOptions()` which returns a `[]metric.Option`. Then in `component/otelcol/receiver/receiver.go` we can call that function and pass the metric options down to `metric.NewMeterProvider()`.

glindstedt:
Even so if we just pass down the options the problem will be the same, the components that use those options when building their MeterProviders will need to be `Update`-ed for the config change to have any effect.

ptodev:
I'm 95% sure that it won't be a problem - flow is designed in such a way that if a config update is required, the `Update()` function needs to be called. You could test it in this way:
1. Start the agent with this flag switched to some value.
2. Hit the `/metrics` endpoint on the agent to see if you have high cardinality metrics.
3. Update the config file so the flag has the opposite value.
4. Hit the [reload endpoint](https://grafana.com/docs/agent/latest/static/api/#reload-configuration-file-beta). Note that this documentation is for static mode, but I think this endpoint works for flow too. 
5. Hit the `/metrics` endpoint again.

rfratto:
@ptodev The suggestion to have a new top-level block with a setting that only impacts a subset of components is weird to me; that's typically not how we've written things for Flow in the past.

What do you think about adding this attribute to a block in the component instead? This would allow per-component tuning of metrics, and would be more Flow native in that things would be logically separated.

ptodev:
@rfratto I'm used to thinking of these feature gates as global settings, since in theory any Otel component could use them. However, in this case we are not setting an Otel feature gate in the Agent, so I agree that we could have per-component overrides.

The main issue with per-component overrides is that we are not sure which components exactly will be impacted, and how to document them. What do you think about this course of action:
* Create a new `metrics` block which is shared between receiver & exporter otel components. This will be similar to how the `output` block is shared.
* Add an option `disable_high_cardinality_metrics` to the `metrics` block.
* Add this new `metrics` block to the code and docs of all exporter and receiver components.

The downside of the above is that if a component doesn't end up producing high cardinality metrics, it'll still have that setting. But I think that's ok.

rfratto:
I'm fine with that proposal, but I would suggest not calling the block `metrics` since that might be confused with the signals emitted from the component.

Something like `debug` or `debug_metrics` would be slightly more descriptive and less ambiguous here.

ptodev:
@rfratto - thank you, `debug_metrics` sounds good to me!

@glindstedt - would you like to try implementing the proposal from [this comment](https://github.com/grafana/agent/pull/4769#issuecomment-1692166879) please? Thank you for working with us, and for your patience! This change would be very useful. If you are too busy, don't worry, we could also pick it up at some point.

glindstedt:
Thank you both for sorting out a path forward. @ptodev yes I think I'll have some time to implement this in the coming week, but if other priorities come in between I'll let you know!

glindstedt:
I've pushed a commit that adds the `debug_metrics` block now, but I have a question. Right now the block is not added to the prometheus and loki exporters/receivers, they don't implement the `receiver.Arguments`/`exporter.Arguments` interfaces like the other components do so I'm wondering if they are special cases and should be treated as such, or if I should go ahead and add the block anyway?

ptodev (changes requested):
Thank you for the update!

It'd be nice to add a few tests at least to some components to make sure that the field is set as expected.

We would also need a documentation, similar to how the output block is currently documented:
* There is a [shared md file](https://github.com/grafana/agent/blob/main/docs/sources/shared/flow/reference/components/output-block.md) for it.
* It is listed in the docs for each component two times:
  * In the [table](https://github.com/grafana/agent/blob/main/docs/sources/flow/reference/components/otelcol.receiver.kafka.md?plain=1#L83) of configuration blocks
  * As a [paragraph](https://github.com/grafana/agent/blob/main/docs/sources/flow/reference/components/otelcol.receiver.kafka.md?plain=1#L252-L254), which imports the shared md file.

In the shared md, we could also mention that this component may not end up using those views if it doesn't generate such metrics. For example, a shared md could look like this:
```
---
aliases:
- ../../otelcol/debug-metrics-block/
canonical: https://grafana.com/docs/agent/latest/shared/flow/reference/components/debug-metrics-block/
headless: true
---

The `debug_metrics` block configures the metrics which this component generates to monitor its state.

The following arguments are supported:

`disable_high_cardinality_metrics` | `bool` | Whether to disable certain high cardinality metrics. | `false` | no

`disable_high_cardinality_metrics` is the Grafana Agent equivalent to the `telemetry.useOtelForInternalMetrics` feature gate in the OpenTelemetry Collector. It removes attributes which could case high cardinality metrics. 
For example, attributes with IP addresses and port numbers in metrics about HTTP and gRPC connections will be removed.
```

ptodev:
> I've pushed a commit that adds the `debug_metrics` block now, but I have a question. Right now the block is not added to the prometheus and loki exporters/receivers, they don't implement the `receiver.Arguments`/`exporter.Arguments` interfaces like the other components do so I'm wondering if they are special cases and should be treated as such, or if I should go ahead and add the block anyway?

This is correct - those are Agent-specific components which are not in the Collector, and they are not in scope for this change.

docs/sources/flow/reference/components/otelcol.exporter.jaeger.md
```diff
@@ -47,6 +47,7 @@ client > tls | [tls][] | Configures TLS for the gRPC client. | no
 client > keepalive | [keepalive][] | Configures keepalive settings for the gRPC client. | no
 sending_queue | [sending_queue][] | Configures batching of data before sending. | no
 retry_on_failure | [retry_on_failure][] | Configures retry mechanism for failed requests. | no
+debug_metrics | [debug_metrics][] | Configures the metrics which this component generates to monitor its state. | no
```
clayton-cornell, line 50:
```suggestion
debug_metrics | [debug_metrics][] | Configures the metrics that this component generates to monitor its state. | no
```

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
```diff
@@ -82,6 +82,7 @@ protocol > otlp > client > tls | [tls][] | Configures TLS for the gRPC client. |
 protocol > otlp > client > keepalive | [keepalive][] | Configures keepalive settings for the gRPC client. | no
 protocol > otlp > queue | [queue][] | Configures batching of data before sending. | no
 protocol > otlp > retry | [retry][] | Configures retry mechanism for failed requests. | no
+debug_metrics | [debug_metrics][] | Configures the metrics which this component generates to monitor its state. | no
```
clayton-cornell, line 85:
```suggestion
debug_metrics | [debug_metrics][] | Configures the metrics that this component generates to monitor its state. | no
```

docs/sources/shared/flow/reference/components/otelcol-debug-metrics-block.md
```diff
@@ -0,0 +1,20 @@
...
+The `debug_metrics` block configures the metrics which this component generates to monitor its state.
+
+The following arguments are supported:
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`disable_high_cardinality_metrics` | `boolean` | Whether to disable certain high cardinality metrics. | `false` | no
+
+`disable_high_cardinality_metrics` is the Grafana Agent equivalent to the
+`telemetry.useOtelForInternalMetrics` feature gate in the OpenTelemetry
```
ptodev, line 17:
```suggestion
`telemetry.disableHighCardinalityMetrics` feature gate in the OpenTelemetry
```

docs/sources/flow/reference/components/otelcol.exporter.logging.md
```diff
@@ -38,6 +38,24 @@ Name | Type | Description | Default | Required
...
 The `verbosity` argument must be one of `"basic"`, `"normal"`, or `"detailed"`. 
 
+## Blocks
+
+The following blocks are supported inside the definition of
+`otelcol.exporter.logging`:
+
+Hierarchy | Block | Description | Required
+--------- | ----- | ----------- | --------
+debug_metrics | [debug_metrics][] | Configures the metrics which this component generates to monitor its state. | no
```
clayton-cornell, line 48:
```suggestion
debug_metrics | [debug_metrics][] | Configures the metrics that this component generates to monitor its state. | no
```

docs/sources/flow/reference/components/otelcol.exporter.otlp.md
```diff
@@ -45,6 +45,7 @@ client > tls | [tls][] | Configures TLS for the gRPC client. | no
 client > keepalive | [keepalive][] | Configures keepalive settings for the gRPC client. | no
 sending_queue | [sending_queue][] | Configures batching of data before sending. | no
 retry_on_failure | [retry_on_failure][] | Configures retry mechanism for failed requests. | no
+debug_metrics | [debug_metrics][] | Configures the metrics which this component generates to monitor its state. | no
```
clayton-cornell, line 48:
```suggestion
debug_metrics | [debug_metrics][] | Configures the metrics that this component generates to monitor its state. | no
```

docs/sources/flow/reference/components/otelcol.receiver.zipkin.md
```diff
@@ -49,13 +49,15 @@ Hierarchy | Block | Description | Required
 --------- | ----- | ----------- | --------
 tls | [tls][] | Configures TLS for the HTTP server. | no
 cors | [cors][] | Configures CORS for the HTTP server. | no
+debug_metrics | [debug_metrics][] | Configures the metrics which this component generates to monitor its state. | no
```
clayton-cornell, line 52:
```suggestion
debug_metrics | [debug_metrics][] | Configures the metrics that this component generates to monitor its state. | no
```

docs/sources/shared/flow/reference/components/otelcol-debug-metrics-block.md
```diff
@@ -0,0 +1,20 @@
+---
+aliases:
+- /docs/agent/shared/flow/reference/components/otelcol-debug-metrics-block/
+canonical: https://grafana.com/docs/agent/latest/shared/flow/reference/components/otelcol-debug-metrics-block/
+headless: true
+---
+
+The `debug_metrics` block configures the metrics which this component generates to monitor its state.
```
clayton-cornell, line 8:
```suggestion
The `debug_metrics` block configures the metrics that this component generates to monitor its state.
```

clayton-cornell (commented):
Some spelling fixes. Switch `which` for `that`.

ptodev (approved):
Looks good! Thank you very much for working on this! I just left one minor comment that needs to be addressed prior to merging.

Also, please feel free to update the Changelog with an entry such as this under the "Enhancements" section:
```
- Added a `disable_high_cardinality_metrics` configuration flag to `otelcol` 
  exporters and receivers to switch high cardinality debug metrics off.  (@glindstedt)
```
That way the change will be visible in the release notes for the next version, and we can pay tribute to you as well :)

component/otelcol/config_debug_metrics.go
```diff
@@ -0,0 +1,16 @@
+package otelcol
+
+// DebugMetricsArguments configures internal metrics of the components
+type DebugMetricsArguments struct {
+	DisableHighCardinalityMetrics bool `river:"disable_high_cardinality_metrics,attr,optional"`
```
ptodev, line 5: TBH I'm not sure if we should just call this `high_cardinality_metrics`. IMO we can merge this as it is for now, and I'll have a discussion with the other people on the team if they'd be in favour of changing it prior to the next release.

glindstedt:
Thank you for the reviews, suggestions commited and changelog updated :+1:

clayton-cornell (approved):

PR #4775, opened 2023-08-10
otel/prometheus_exporter: disable otel_scope_info by default

gouthamve:
The collector doesn't implement this yet and the metrics are not very useful.

See: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/24248

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated
- [ ] ~Config converters updated~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@ Main (unreleased)
   to enable decompression explicitly. See the [upgrade guide][] for migration
   instructions. (@thampiotr)
 
+- `otelcol.exporter.prometheus`: Set `include_scope_info` to `false` by default. You can set 
+  it to `true` to preserve previous behavior. (@gouthamve)
+
 ### Enhancements
 
 - Integrations: include `direct_connect`, `discovering_mode` and `tls_basic_auth_config_path` fields for MongoDB configuration. (@gaantunes)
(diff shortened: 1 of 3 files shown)


tpaschalis (approved):
LGTM, cc-ing @ptodev who worked on upgrading our OTel dependency this past week to merge.

rfratto:
I'm not necessarily against the change, but I'm not sure if this is going against the compatibility spec.

The compatibility [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#instrumentation-scope-1) states:

> Prometheus exporters SHOULD provide a configuration option to disable the otel_scope_info metric and otel_scope_ labels.

I took this to mean that the configuration option should exist, and should default to true. Did I misunderstand this line?

clayton-cornell (commented):
Docs are OK as-is. 

This doesn't account for the clarification of the default status.

gouthamve:
Kicked off a discussion in OTEL land which led to this: https://github.com/open-telemetry/opentelemetry-specification/issues/3657

Having the metrics switched on by default doesn't provide a lot of value and I think the spec will evolve to disable metrics unless there are attributes.

Should we go ahead and merge this? cc @rfratto

PR #4778, opened 2023-08-10
universal converters prototype

thampiotr:
#### PR Description

This is a prototype / attempt to implement a more universal way of converting between different struct formats.

I've noticed that we sometimes have an external format (e.g. promtail) converted into internal format (flow) and then again into promtail so that the config can be passed to the implementation of the component.

It would be nice to define how to convert from one type to another all in one place, with an automatic way of reversing the conversion. It's certainly much harder to implement in Go than in some other languages, even with generics. After seeing this, I'm not sure if it's winning against writing these by hand.


converter/bijection/bijection_test.go
```diff
@@ -0,0 +1,105 @@
...
+		},
+		BtoA: func(b *float64, a *uint64) error {
+			*a = uint64(*b)
+			return nil
+		},
+	}
+
+	BindField[RiverExample, YamlExample, int32, int64](&bi, PropertyNames{A: "TestRiver", B: "TestYAML"}, int32ToInt64)
+	BindField[RiverExample, YamlExample, uint64, float64](&bi, PropertyNames{A: "UInt", B: "UIntValue"}, uint64ToFloat64)
+	BindField(&bi, PropertyNames{A: "Str", B: "Bytes"}, Copy[string]())
```
thampiotr, line 79: Getting any type safety in Go generics for this use case ends up being very verbose.

converter/auto/auto_examples_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"fmt"
+	"github.com/grafana/agent/component/loki/source/kafka"
+	"github.com/grafana/loki/clients/pkg/promtail/scrapeconfig"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestKafka(t *testing.T) {
```
thampiotr, line 11: This test reveals that in this approach we have plenty of exceptions to add,  making this approach probably more error-prone than manually spelling out the entire conversion.

converter/bijection/example_test.go
```diff
@@ -0,0 +1,206 @@
...
+	"github.com/grafana/agent/component/common/config"
+	"github.com/grafana/agent/component/loki/source/kafka"
+	"github.com/grafana/agent/pkg/river/rivertypes"
+	"github.com/grafana/dskit/flagext"
+	"github.com/grafana/loki/clients/pkg/promtail/scrapeconfig"
+	common "github.com/prometheus/common/config"
+	"github.com/prometheus/common/model"
+)
+
+func TestKafkaBijection(t *testing.T) {
```
thampiotr, line 16: This test verifies a two-way conversion of a more complex structure (kafka logs source config). 

The definition of this bijection (conversion logic) is below the test.

converter/bijection/types.go
```diff
@@ -0,0 +1,6 @@
+package bijection
+
+type Bijection[A interface{}, B interface{}] interface {
+	ConvertAToB(a *A, b *B) error
+	ConvertBToA(b *B, a *A) error
+}
```
thampiotr, line 6: This is the idea in essence - a definition of a conversion that can be done in both directions.

PR #4780, opened 2023-08-11
wal: fix detection of corruption errors

tpaschalis:
#### PR Description
This PR fixes a sneaky bug that makes WAL loading more brittle in cases of corruption.

I saw this initially when the Agent was crashlooping on building the component with a message like:
```
Failed to build component: building component: read records: %!w(<nil>)
```

The underlying issue is that we're returning a new `errors.(*errorString)` but we're never wrapping the original `r.Err()` of type `*wlog.CorruptionErr`.

When the return value of loadWAL bubbles up to wal.NewStorage, it never has the chance of being detected as the correct error type to avoid failure to build the storage.

https://github.com/grafana/agent/blob/dc3cdde444ad605c68a0728edcb0a926e04ba50c/pkg/metrics/wal/wal.go#L179-L185

No issue filed.

#### Notes to the Reviewer
I also noticed that the agent would crashloop in case a segment was missing. If we have the following segments and delete the first one
```
data-agent/prometheus.remote_write.default/wal/00000000
data-agent/prometheus.remote_write.default/wal/00000001
data-agent/prometheus.remote_write.default/wal/00000002
data-agent/prometheus.remote_write.default/wal/00000003
```

the agent will still crash loop with the following. I think we should also work for a fix on this.
```
Error: /Users/tpaschalis/river-configs/component-logger.river:30:1: Failed to build component: building component: open WAL segment 0: open data-agent/prometheus.remote_write.default/wal/00000000: no such file or directory
```

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,8 @@ Main (unreleased)
 
 - Fix issue on Windows where DNS short names were unresolvable. (@rfratto)
 
+- Fix issue where corrupt WAL segments lead to crash looping. (@tpaschalis)
+
 v0.35.3 (2023-08-09)
 --------------------
 
--- a/pkg/metrics/wal/wal.go
+++ b/pkg/metrics/wal/wal.go
@@ -451,7 +451,7 @@ func (w *Storage) loadWAL(r *wlog.Reader, multiRef map[chunks.HeadSeriesRef]chun
 		return err
 	default:
 		if r.Err() != nil {
-			return fmt.Errorf("read records: %w", err)
+			return fmt.Errorf("read records: %w", r.Err())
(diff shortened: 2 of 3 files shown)


rfratto (approved):
Can you check to see if this same bug exists in Prometheus Agent too?

tpaschalis:
> Can you check to see if this same bug exists in Prometheus Agent too?

I checked, [it doesn't](https://github.com/prometheus/prometheus/blob/4a56a6bf592ad032714ddedaee1b5e4fc77aa93e/tsdb/agent/db.go#L569-L572).

PR #4790, opened 2023-08-14
flow: change clustering interface to be a call to NotifyClusterChange

rfratto:
Previously, components would opt in to clustering notifications through a call to ClusterUpdatesRegistration. Then, if that method returned true, they would be re-evaluated.

In preparation for the move to services, the clustering service would not be able to function the same way, as services are not permitted to trigger a re-evaluation of a component.

The new interface is a NotifyClusterChange method, which allows components to opt-in to whether they want to ignore that signal or not.

This is not reflected in the CHANGELOG as it is a no-op for users.

### DIFF
--- a/component/prometheus/operator/common/component.go
+++ b/component/prometheus/operator/common/component.go
@@ -11,7 +11,7 @@ import (
 )
 
 type Component struct {
-	mut     sync.Mutex
+	mut     sync.RWMutex
 	config  *operator.Arguments
 	manager *crdManager
 
@@ -108,8 +108,8 @@ func (c *Component) Update(args component.Arguments) error {
 
 // NotifyClusterChange implements component.ClusterComponent.
 func (c *Component) NotifyClusterChange() {
-	c.mut.Lock()
-	defer c.mut.Unlock()
+	c.mut.RLock()
+	defer c.mut.RUnlock()
 
 	if !c.config.Clustering.Enabled {
 		return // no-op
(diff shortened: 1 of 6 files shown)


component/prometheus/operator/common/component.go
```diff
@@ -106,18 +106,27 @@ func (c *Component) Update(args component.Arguments) error {
 	return nil
 }
 
+// NotifyClusterChange implements component.ClusterComponent.
+func (c *Component) NotifyClusterChange() {
+	c.mut.Lock()
```
mattdurham, line 111: nit: is there any benefit to changing this to a RW mutex and using RLock/RUnlock?

component/prometheus/operator/common/component.go
```diff
@@ -106,18 +106,27 @@ func (c *Component) Update(args component.Arguments) error {
...
+	defer c.mut.Unlock()
+
+	if !c.config.Clustering.Enabled {
+		return // no-op
+	}
+
+	// Schedule a reload so targets get redistributed.
+	select {
+	case c.onUpdate <- struct{}{}:
```
captncraig, line 120: This is a case where this change can probably simplify things a lot. Trying to differentiate a clustering update vs an argument update was a real pain. We can probably not send on this channel, and just call `c.manager.ClusteringUpdated()` directly, and simplify the logic in Run by a bit.

captncraig (approved):

PR #4813, opened 2023-08-15
prometheus.operator.*: fix panic when port not defined in CRD

captncraig:
If you didn't define a port at all in your CRD, this would panic. That is not good.


petewall (approved):
Makes sense. Thanks!

PR #4814, opened 2023-08-15
Update cross reference syntax in Agent static mode documentation

clayton-cornell:
#### PR Description

Topic cross references need to be updated to allow for the fact that the Agent documentation is being reused in the Grafana Cloud documentation.

See https://grafana.com/docs/writers-toolkit/write/shortcodes/#docsreference-shortcode for details of the shortcode used to accomplish this cross referencing from different documentation trees.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


rfratto (approved):
Rubber stamp :)

clayton-cornell:
> When do we have to use this new syntax? The syntax is pretty intense.

I've asked for clarification...

docs/sources/static/release-notes.md
```diff
@@ -15,16 +15,18 @@ The release notes provide information about deprecations and breaking changes in
...
+> Other release notes for the different Grafana Agent variants are contained on separate pages:
+>
+> * [Static mode Kubernetes operator release notes][release-notes-operator]
+> * [Flow mode release notes][release-notes-flow]
+
+{{% docs/reference %}}
+[release-notes-operator]: "/docs/agent/ -> /docs/agent/<AGENT VERSION>/operator/release-notes"
+[release-notes-operator]: "/docs/grafana-cloud/ -> /docs/grafana-cloud/monitor-infrastructure/integrations/agent/operator/release-notes"
+[release-notes-flow]: "/docs/agent/ -> /docs/agent/<AGENT VERSION>/flow/release-notes"
+[release-notes-flow]: "/docs/grafana-cloud/ -> /docs/grafana-cloud/monitor-infrastructure/integrations/agent/flow/release-notes"
```
jdbaldry, line 28: So I believe the intent is that we will unmount the Flow documentation from Grafana Cloud which means this page won't exist. Do we want to link back to the Agent docs instead?

Alternatively, we can leave Flow release notes in Grafana Cloud and exclude everything else.

jdbaldry (commented):
One nit and one question regarding the related work in Grafana Cloud documentation.

As discussed outside the PR, you can also consider replacing the absolute paths in Grafana Cloud link destinations with relative ones if you want to avoid them breaking in the case that the directory is moved in Grafana Cloud again.

For example, https://github.com/grafana/agent/pull/4814/files#diff-3c73d39c94bdb1e8a9790eb0d6b9f4fefc552990451a303a5a69302c5ddd16a0R89, could use:
```
[set up]: "/docs/grafana-cloud/ -> ./set-up"
```

docs/sources/static/_index.md
```diff
@@ -1,6 +1,7 @@
 ---
 canonical: https://grafana.com/docs/agent/latest/static/
 title: Static mode
+menuTitle: Static mode
```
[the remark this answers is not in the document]
clayton-cornell: I'll remove for now. I added it because I had an initial thought of expanding the metadata at the same time as the other work here. Better that I have a dedicated PR for this metadata rework.

clayton-cornell:
> ```
> [set up]: "/docs/grafana-cloud/ -> ./set-up"
> ```

Done and committed to this branch.

docs/sources/static/release-notes.md
clayton-cornell: Can we direct back out of Cloud into the Agent docs for this link? Maybe not a good idea since it'll push the reader out of the Cloud doc tree if they click the link... but it's a short term solution either way.

PR #4822, opened 2023-08-16
Add ability to delete WAL on startup

thampiotr:
#### PR Description

While we should address the root causes of issues with WAL, it often happens that such issues lead to the Agent failing to start up. 

In such situations often deleting the WAL is the right solution. However, when for example a container is crashlooping with high frequency or when there are 100s of agents crashing on startup, it can become time-consuming to delete the WAL at a correct path. 

We should make it easier to delete WAL so that users can recover more quickly. This PR adds an environment variable, which when it's set, the WAL will be deleted at startup. 

#### Notes to the Reviewer

Tested on flow:

```
ts=2023-08-16T09:42:02.153807Z level=info controller_id= trace_id=02e15d24bff0dfd990a5cbe398dde72d msg="starting complete graph evaluation"
ts=2023-08-16T09:42:02.154519Z level=info component=prometheus.remote_write.cloud msg="Deleted WAL directory on startup" path=data-agent/prometheus.remote_write.cloud/wal err=null
ts=2023-08-16T09:42:02.155831Z level=info component=prometheus.remote_write.cloud subcomponent=wal msg="replaying WAL, this may take a while" dir=data-agent/prometheus.remote_write.cloud/wal
```

#### PR Checklist

- [ ] Make the same change for static mode
- [ ] Tests updated

### DIFF
--- a/component/prometheus/remotewrite/remote_write.go
+++ b/component/prometheus/remotewrite/remote_write.go
@@ -72,6 +72,15 @@ func New(o component.Options, c Arguments) (*Component, error) {
 	oldDataPath := filepath.Join(o.DataPath, "wal", o.ID)
 	_ = os.RemoveAll(oldDataPath)
 
+	// When experimental delete WAL on startup is enabled, we delete the WAL. This is a useful workaround for a range
+	// of issues where processing WAL leads to a failure of the agent. This should be used sparingly and the root cause
+	// of each issue should be investigated.
+	if os.Getenv("EXPERIMENTAL_DELETE_WAL_ON_STARTUP") == "true" {
+		walPath := wal.SubDirectory(o.DataPath)
+		err := os.RemoveAll(walPath)
+		level.Info(o.Logger).Log("msg", "Deleted WAL directory on startup", "path", walPath, "err", err)
+	}
+
 	walLogger := log.With(o.Logger, "subcomponent", "wal")
 	walStorage, err := wal.NewStorage(walLogger, o.Registerer, o.DataPath)
 	if err != nil {


component/prometheus/remotewrite/remote_write.go
```diff
@@ -72,6 +72,15 @@ func New(o component.Options, c Arguments) (*Component, error) {
 	oldDataPath := filepath.Join(o.DataPath, "wal", o.ID)
 	_ = os.RemoveAll(oldDataPath)
 
+	// When experimental delete WAL on startup is enabled, we delete the WAL. This is a useful workaround for a range
+	// of issues where processing WAL leads to a failure of the agent. This should be used sparingly and the root cause
+	// of each issue should be investigated.
+	if os.Getenv("EXPERIMENTAL_DELETE_WAL_ON_STARTUP") == "true" {
```
ptodev, line 78: Shouldn't this be a config option in the component config?

thampiotr: I think a knob like this is a kind of an insurance policy that may come very useful one day (or even today), but in the ideal world it would not have to exist. So I would prefer this to be a special thing, perhaps even not documented. But I'm open to hear others' opinions and change this approach if needed.

buro9:
I have concerns.

Not withstanding that work may be needed to make the WAL higher quality, nor that whilst data is critical and a loss of which is viewed as unacceptable by most customers... I do accept that a minority of customers would prefer a managed data loss in some scenarios to better manage a speedy recovery from an incident.

The concerns I have relate to the effects of adding any delete mechanism that can be left permanently engaged. By doing so we make it far more likely that data loss can now occur more routinely, and it would go against the goal of this PR by extending MTTR as you the customer now does not have evidence that can help with diagnosis of an issue. One of the key goals of all incident management (see Google SRE book as well as our internal process) is to preserve evidence whilst stopping the bleeding, but deleting the WAL destroys the evidence.

If deleting the WAL can still be justified as a step to take, a one-time delete switch rather than a permanent operational mode (by leaving the proposed ENV var in place for multiple startups of the Agent) is surely preferred.

I would caution against choosing to permanently disable or delete the WAL as it wouldn't help with the scenario describe, all options to address WAL reliability would be preferred first.

thampiotr:
Abandoning this draft after discussion.

PR #4825, opened 2023-08-16
Fix broken admonition

thampiotr:


clayton-cornell (approved):
Oooops, looks good to me

PR #4830, opened 2023-08-16
Fix support bundle generation

### LINKED ISSUE #4803: Agent 0.35.2 and 0.35.3 cannot build support bundle
### What's wrong?

As reported in Slack [here](https://raintank-corp.slack.com/archives/CSN5HV0CQ/p1692051807656359):

> on both my Windows and Linux agents 
> (which are on 0.35.2 or .3)
> I get this error now when trying to hit the support bundle API endpoint:
> 
> ```
> curl http://localhost:9090/-/support
> failed to get Agent logs targets: Get "http://agent.internal:12345/agent/api/v1/logs/targets": dial tcp: lookup agent.internal on 45.33.124.75:53: no such host
> ```
> 
> and no support bundle is generated
> but if I go to the actual logs targets endpoint (`curl http://localhost:9090/agent/api/v1/logs/targets`), I get the expected response
> so maybe the problem is the `agent.internal` in that URL?
> Possible Agent bug?
> This is a bit of a blocker for working some support tickets

Response from @mattdurham:

> Def looks like a bug, it should not be using the agent.internal link since that is for integrations
> Can you a file a bug with the details?

### Steps to reproduce

1. make sure Agent is started with `-config.enable-read-api`
2. confirm what port number Agent is running its HTTP server on (e.g., by looking in `/etc/default/grafana-agent`)
3. try to generate a support bundle. Example: `curl http://localhost:9090/-/support`
4. see error like `failed to get Agent logs targets: Get "http://agent.internal:12345/agent/api/v1/logs/targets": dial tcp: lookup agent.internal on 45.33.115.76:53: no such host`

### System information

Windows 10 and Linux (Debian 11)

### Software version

Grafana Agent 0.35.2 & 0.35.3

### Configuration

```text
make sure Agent is started with `-config.enable-read-api`
```

### Logs

```text
Aug 14 23:37:02 linode-debian11 grafana-agent[12387]: ts=2023-08-14T23:37:02.601940789Z caller=gokit.go:79 level=warn msg="GET /-/support (500) 43.856327ms Response: \"failed to get Agent logs targets: Get \\\"http://agent.internal:12345/agent/api/v1/logs/targets\\\": dial tcp: lookup agent.internal on 45.33.115.76:53: no such host\\n\" ws: false; Accept: */*; User-Agent: curl/7.74.0; "
Aug 14 23:42:30 linode-debian11 grafana-agent[12387]: ts=2023-08-14T23:42:30.723903349Z caller=gokit.go:79 level=warn msg="GET /-/support (500) 22.794825ms Response: \"failed to get Agent logs targets: Get \\\"http://agent.internal:12345/agent/api/v1/logs/targets\\\": dial tcp: lookup agent.internal on 198.58.107.5:53: no such host\\n\" ws: false; Accept: */*; User-Agent: curl/7.74.0; "
```

spartan0x117:
#### PR Description
When trying to get the logs targets, a DefaultHTTPClient was being used, which is not able to access in-memory HTTP addresses. This fixes it to use the correct HTTP Client.

Fixes #4803

#### PR Checklist

- [X] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,8 @@ Main (unreleased)
 
 - Fix panic in `prometheus.operator.*` when no Port supplied in Monitor crds. (@captncraig)
 
+- Fix issue where getting the support bundle failed due to using an HTTP Client that was not able to access the agent in-memory address. (@spartan0x117)
+
 v0.35.4 (2023-08-14)
 --------------------
 
--- a/cmd/grafana-agent/entrypoint.go
+++ b/cmd/grafana-agent/entrypoint.go
@@ -299,7 +299,7 @@ func (ep *Entrypoint) supportHandler(rw http.ResponseWriter, r *http.Request) {
 	ep.mut.Lock()
 	var (
 		enabledFeatures = ep.cfg.EnabledFeatures
-		httpSrvAddress  = ep.cfg.ServerFlags.HTTP.InMemoryAddr
+		httpSrvAddress  = ep.cfg.ServerFlags.HTTP.ListenAddress
 	)
 	ep.mut.Unlock()
 
--- a/pkg/supportbundle/supportbundle.go
+++ b/pkg/supportbundle/supportbundle.go
@@ -119,7 +119,7 @@ func Export(ctx context.Context, enabledFeatures []string, cfg []byte, srvAddres
 		return nil, fmt.Errorf("failed to read Agent logs instances: %s", err)
 	}
 
-	resp, err = http.DefaultClient.Get("http://" + srvAddress + "/agent/api/v1/logs/targets")
+	resp, err = httpClient.Get("http://" + srvAddress + "/agent/api/v1/logs/targets")
 	if err != nil {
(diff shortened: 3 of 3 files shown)


tpaschalis:
Can you verify that this also fixes the issue when we run the Agent on a non-default HTTP port?

The other solution would be to not use the internal address (it was designed to be used by integrations and most likely it's not worth it as an improvement), but use the Agent's proper HTTP server address.

spartan0x117:
I tested with:
```
./grafana-agent -config.file ./test.yaml -server.http.address 127.0.0.1:9999
```
and was able to get a support bundle. Agreed that the internal memory address is not needed as an optimization since support bundles are not getting fetched constantly. I'll test out changing the address to the configured listen address now :smile:

tpaschalis (approved):
LGTM, thanks for the quick fix 🙌
I think any of the two addresses will work, feel free to use whatever you think is best!

PR #4852, opened 2023-08-17
prometheus.operator.*, simplify cluster update logic.

captncraig:
We previously had to discriminate between "argument" updates and "clustering" updates. Now that the notification mechanism for clustering is changed, this gets a bit simpler.

I don't think this needs a changelog.

### DIFF
--- a/component/prometheus/operator/common/component.go
+++ b/component/prometheus/operator/common/component.go
@@ -64,8 +64,7 @@ func (c *Component) Run(ctx context.Context) error {
 			c.reportHealth(err)
 		case <-c.onUpdate:
 			c.mut.Lock()
-			nextConfig := c.config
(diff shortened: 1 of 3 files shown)


rfratto (commented):
I don't feel like I know enough about these components to say whether this is correct or not. I think someone from @grafana/grafana-agent-infrastructure-maintainers should be reviewing this.

(I do agree it doesn't need a CHANGELOG entry though)

component/prometheus/operator/common/component.go
```diff
@@ -64,28 +63,20 @@ func (c *Component) Run(ctx context.Context) error {
 		case err := <-errChan:
 			c.reportHealth(err)
 		case <-c.onUpdate:
-
 			c.mut.Lock()
 			nextConfig := c.config
```
marctc, line 67: nit: we don't need this var now

marctc (approved):

PR #4853, opened 2023-08-17
docs: extract exporter Exports section to a shared snippet

tpaschalis:
A community member reported an issue with the following documentation snippet on exporters 

```
For example, the `targets` can either be passed to a `prometheus.relabel`
component to rewrite the metric's label set, or to a `prometheus.scrape`
component that collects the exposed metrics.
```

Targets should be passed to `discovery.relabel`  to rewrite the targets' label sets; any metric-level relabeling should be done _after_ prometheus.scrape.

Since this change was present in a lot of places, I also extracted it to a shared snippet in `docs/sources/shared/flow/reference/components/exporters-components-exports.md`.

The drawback is that the description of the targets field no longer explicitly mentions _which_ exporter it is, but is a more generic phrase. This could be fixed by parameterizing Hugo shortcodes, but it's more effort than required right now IMHO.

```
> old    The targets that can be used to collect `github` metrics
< new    The targets that can be used to collect exporter metrics
```

- [X] Documentation added
- [ ] Tests updated (N/A)

### DIFF
--- a/docs/sources/flow/reference/components/prometheus.exporter.apache.md
+++ b/docs/sources/flow/reference/components/prometheus.exporter.apache.md
@@ -25,21 +25,8 @@ Name | Type | Description | Default | Required
 `insecure`      | `bool`   | Ignore server certificate if using https. | false | no
 
 ## Exported fields
-The following fields are exported and can be referenced by other components.
 
-Name      | Type                | Description
---------- | ------------------- | -----------
-`targets` | `list(map(string))` | The targets that can be used to collect `apache` metrics.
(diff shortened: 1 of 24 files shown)


clayton-cornell:
> To me, extracting it to a shared location makes sense anyway. As for the specific wording, feel free to propose a different one if you wish.

I think it's fine as-is.

Shoudl we backport this to v0.33 through v0.35? or leave it in `next` for v.036 and newer?

clayton-cornell (approved):

PR #4859, opened 2023-08-18
Add exemplar support otelcol exporter prometheus

### LINKED ISSUE #2485: otelcol.exporter.prometheus: Add exemplar support
`otelcol.exporter.prometheus` was released without support for exemplars at first; we should add it.

wildum:
#### PR Description

Add exemplars support for the otelcol.exporter.prometheus. The exemplars are available for the histograms and for the sums of type "counter" (cumulative and monotonic).

Fixes #2485 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/component/otelcol/exporter/prometheus/internal/convert/cache.go
+++ b/component/otelcol/exporter/prometheus/internal/convert/cache.go
@@ -110,8 +110,7 @@ func (series *memorySeries) WriteExemplarsTo(app storage.Appender, e exemplar.Ex
 	series.Lock()
 	defer series.Unlock()
 
-	_, err := app.AppendExemplar(series.id, series.labels, e)
-	if err != nil {
+	if _, err := app.AppendExemplar(series.id, series.labels, e); err != nil {
 		return err
 	}
 
--- a/component/otelcol/exporter/prometheus/internal/convert/convert.go
+++ b/component/otelcol/exporter/prometheus/internal/convert/convert.go
@@ -292,7 +292,7 @@ func (conv *Converter) consumeGauge(app storage.Appender, memResource *memorySer
 		Help: m.Description(),
 	})
 	if err := metricMD.WriteTo(app, time.Now()); err != nil {
-		level.Warn(conv.log).Log("msg", "failed to write metric family metadata", "err", err)
+		level.Warn(conv.log).Log("msg", "failed to write metric family metadata, metric name", metricName, "err", err)
 	}
 
 	for dpcount := 0; dpcount < m.Gauge().DataPoints().Len(); dpcount++ {
@@ -300,7 +300,7 @@ func (conv *Converter) consumeGauge(app storage.Appender, memResource *memorySer
 
 		memSeries := conv.getOrCreateSeries(memResource, memScope, metricName, dp.Attributes())
 		if err := writeSeries(app, memSeries, dp, getNumberDataPointValue(dp)); err != nil {
-			level.Error(conv.log).Log("msg", "failed to write metric sample", "err", err)
+			level.Error(conv.log).Log("msg", "failed to write metric sample", metricName, "err", err)
 		}
 	}
 }
@@ -420,7 +420,7 @@ func (conv *Converter) consumeSum(app storage.Appender, memResource *memorySerie
 		Help: m.Description(),
 	})
 	if err := metricMD.WriteTo(app, time.Now()); err != nil {
-		level.Warn(conv.log).Log("msg", "failed to write metric family metadata", "err", err)
+		level.Warn(conv.log).Log("msg", "failed to write metric family metadata, metric name", metricName, "err", err)
 	}
 
 	for dpcount := 0; dpcount < m.Sum().DataPoints().Len(); dpcount++ {
@@ -430,13 +430,13 @@ func (conv *Converter) consumeSum(app storage.Appender, memResource *memorySerie
 
 		val := getNumberDataPointValue(dp)
 		if err := writeSeries(app, memSeries, dp, val); err != nil {
-			level.Error(conv.log).Log("msg", "failed to write metric sample", "err", err)
+			level.Error(conv.log).Log("msg", "failed to write metric sample", metricName, "err", err)
 		}
 
 		if convType == textparse.MetricTypeCounter {
 			for i := 0; i < dp.Exemplars().Len(); i++ {
 				if err := conv.writeExemplar(app, memSeries, dp.Exemplars().At(i)); err != nil {
-					level.Error(conv.log).Log("msg", "failed to write exemplar for metric sample", "err", err)
+					level.Error(conv.log).Log("msg", "failed to write exemplar for metric sample", metricName, "err", err)
 				}
 			}
 		}
@@ -459,7 +459,7 @@ func (conv *Converter) consumeHistogram(app storage.Appender, memResource *memor
 		Help: m.Description(),
 	})
 	if err := metricMD.WriteTo(app, time.Now()); err != nil {
-		level.Warn(conv.log).Log("msg", "failed to write metric family metadata", "err", err)
+		level.Warn(conv.log).Log("msg", "failed to write metric family metadata, metric name", metricName, "err", err)
 	}
 
 	for dpcount := 0; dpcount < m.Histogram().DataPoints().Len(); dpcount++ {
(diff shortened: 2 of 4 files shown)


component/otelcol/exporter/prometheus/internal/convert/convert.go
```diff
@@ -524,10 +563,44 @@ func (conv *Converter) consumeHistogram(app storage.Appender, memResource *memor
...
 		}
 	}
 }
 
+// Convert Otel Exemplar to Prometheus Exemplar.
+func (conv *Converter) convertExemplar(otelExemplar pmetric.Exemplar, ts time.Time) exemplar.Exemplar {
+	exemplarLabels := make(labels.Labels, 0)
+
+	if traceID := otelExemplar.TraceID(); !traceID.IsEmpty() {
+		exemplarLabels = append(exemplarLabels, labels.Label{Name: "trace_id", Value: hex.EncodeToString(traceID[:])})
```
thampiotr, line 582: I don't think you need to create a new slice here with `[:]` cause encoding is read-only? LMK if I'm missing something.

component/otelcol/exporter/prometheus/internal/convert/convert.go
```diff
@@ -524,10 +563,44 @@ func (conv *Converter) consumeHistogram(app storage.Appender, memResource *memor
...
+	}
+
+	if spanID := otelExemplar.SpanID(); !spanID.IsEmpty() {
+		exemplarLabels = append(exemplarLabels, labels.Label{Name: "span_id", Value: hex.EncodeToString(spanID[:])})
```
thampiotr, line 586: Same as above.

component/otelcol/exporter/prometheus/internal/convert/convert.go
```diff
@@ -524,10 +563,44 @@ func (conv *Converter) consumeHistogram(app storage.Appender, memResource *memor
...
+	}
+
+	var value float64
+	switch otelExemplar.ValueType() {
```
thampiotr, line 590: Technically it can also be `ExemplarValueTypeEmpty` 🤔

component/otelcol/exporter/prometheus/internal/convert/convert_test.go
```diff
@@ -110,7 +118,27 @@ func TestConverter(t *testing.T) {
 									"count": 333,
 									"sum": 100,
 									"bucket_counts": [0, 111, 0, 222],
-									"explicit_bounds": [0.25, 0.5, 0.75, 1.0]
+									"explicit_bounds": [0.25, 0.5, 0.75, 1.0],
```
thampiotr, line 121: As mentioned above - are bounds always guaranteed to be provided sorted or must the code sort them?

component/otelcol/exporter/prometheus/internal/convert/convert_test.go
```diff
@@ -110,7 +118,27 @@ func TestConverter(t *testing.T) {
...
+										},
+										{
+											"time_unix_nano": 1000000002,
+											"as_double": 0.5,
+											"span_id": "bbbbbbbbbbbbbbbb",
+											"trace_id": "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb"
+										},
+										{
+											"time_unix_nano": 1000000003,
+											"as_double": 1.5,
```
thampiotr, line 137: These are conveniently sorted already too. We should have a test that requires them to be sorted.

docs/sources/flow/reference/components/otelcol.exporter.prometheus.md
```diff
@@ -66,7 +66,6 @@ are forwarded to the `forward_to` argument.
 The following are dropped during the conversion process:
 
 * Metrics that use the delta aggregation temporality
-* Exemplars on OpenTelemetry cumulative sums and OpenTelemetry Histograms
```
thampiotr, line 69: Nice catch!

pkg/util/testappender/internal/dtobuilder/dtobuilder.go
```diff
@@ -398,6 +398,11 @@ func convertExemplar(mt dto.MetricType, e exemplar.Exemplar) *dto.Exemplar {
 
 func findBucket(h *dto.Histogram, bound float64) *dto.Bucket {
 	for _, b := range h.GetBucket() {
+		// Special handling because Inf - Inf returns NaN.
+		if bound == math.Inf(1) && b.GetUpperBound() == math.Inf(1) {
+			return b
+		}
```
thampiotr, line 404: 👍

thampiotr (commented):
Looking good!

component/otelcol/exporter/prometheus/internal/convert/cache.go
```diff
@@ -101,6 +102,19 @@ func (series *memorySeries) WriteTo(app storage.Appender, ts time.Time) error {
...
 	}
+
+	return nil
+}
+
+func (series *memorySeries) WriteExemplarsTo(app storage.Appender, e exemplar.Exemplar) error {
+	series.Lock()
+	defer series.Unlock()
+
+	_, err := app.AppendExemplar(series.id, series.labels, e)
```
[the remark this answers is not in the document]
wildum: that was also my guess but I can have a deeper look on how the reference numbers are managed

component/otelcol/exporter/prometheus/internal/convert/cache.go
```diff
@@ -101,6 +102,19 @@ func (series *memorySeries) WriteTo(app storage.Appender, ts time.Time) error {
...
+	if err != nil {
+		return err
+	}
+
```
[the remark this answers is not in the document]
wildum: Here I would actually suggest a different change:
```
	if _, err := app.AppendExemplar(series.id, series.labels, e); err != nil {
		return err
	}
	return nil
```
Although this is a bit more verbose, I found returning "err" misleading because in most areas when "err" is returned you expect that it is != nil

component/otelcol/exporter/prometheus/internal/convert/convert.go
```diff
@@ -488,6 +507,17 @@ func (conv *Converter) consumeHistogram(app storage.Appender, memResource *memor
 			buckets[bound] = c
 		}
 
+		// Sort the exemplars by value.
```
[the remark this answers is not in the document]
wildum: Wow good point, I missed that it was actually not sorting. I found this though: 
```
	// explicit_bounds specifies buckets with explicitly defined bounds for values.
	//
	// The boundaries for bucket at index i are:
	//
	// (-infinity, explicit_bounds[i]] for i == 0
	// (explicit_bounds[i-1], explicit_bounds[i]] for 0 < i < size(explicit_bounds)
	// (explicit_bounds[i-1], +infinity) for i == size(explicit_bounds)
	//
	// The values in the explicit_bounds array must be strictly increasing.
	//
	// Histogram buckets are inclusive of their upper boundary, except the last
	// bucket where the boundary is at infinity. This format is intentionally
	// compatible with the OpenMetrics histogram definition.
	ExplicitBounds []float64 `protobuf:"fixed64,7,rep,packed,name=explicit_bounds,json=explicitBounds,proto3" json:"explicit_bounds,omitempty"`
```
"// The values in the explicit_bounds array must be strictly increasing." -> can we trust this and assume that the bounds are indeed correctly ordered?

component/otelcol/exporter/prometheus/internal/convert/convert.go
wildum: traceID is an array (type TraceID [16]byte) and EncodeToString is expecting a slice so I believe that we need to convert it into a slice this way (cant pass traceID directly

component/otelcol/exporter/prometheus/internal/convert/convert.go
wildum: should we explicitly handle it? I took this function from the otel collector. It also says that "// Calling this function on zero-initialized Exemplar will cause a panic." on the ValueType function.

component/otelcol/exporter/prometheus/internal/convert/convert.go
thampiotr: ah right, it has a fixed size `[16]` 👍

thampiotr (approved):
LGTM! thanks :)

PR #4865, opened 2023-08-21
Add support for retry on http429 to loki write

### LINKED ISSUE #4335: Support promtail's `drop_rate_limited_batches` in `loki.write`
### Request

`promtail` currently supports the following setting for Loki clients:
```
	// When enabled, Promtail will not retry batches that get a
	// 429 'Too Many Requests' response from the distributor. Helps
	// prevent HOL blocking in multitenant deployments.
	DropRateLimitedBatches bool `yaml:"drop_rate_limited_batches"`
```

This is not supported in `loki.write` component.

### Use case

Helps prevent HOL blocking in multitenant deployments.

wildum:
#### PR Description
Add support for retry on http429 to loki write for static and flow mode. It was decided in the issue to go with the property RetryOnHTTP429 like in prometheus remote write rather than going for the DropRateLimitedBatches used in promtail.

Fixes #4335

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/converter/internal/promtailconvert/testdata/loki_write.yaml
+++ b/converter/internal/promtailconvert/testdata/loki_write.yaml
@@ -1,0 +1,17 @@
+clients:
+  - url: http://localhost/loki/api/v1/push
+    headers:
+      fat: ball
+      bird: house
+    tenant_id: sparrow
+    batchwait: 10s
+    batchsize: 1024
+    proxy_url: http://proxy.example.com
+    backoff_config:
+      min_period: 100ms
+      max_period: 10s
+      max_retries: 20
+    drop_rate_limited_batches: true
+    timeout: 30s
+tracing: {enabled: false}
+server: {register_instrumentation: false}
--- a/converter/internal/promtailconvert/testdata/loki_write.yaml
+++ b/converter/internal/promtailconvert/testdata/loki_write.yaml
@@ -13,5 +13,7 @@ clients:
       max_retries: 20
     drop_rate_limited_batches: true
     timeout: 30s
+  - url: http://localhost/loki/api/v1/push
+    drop_rate_limited_batches: false
 tracing: {enabled: false}
 server: {register_instrumentation: false}
--- a/converter/internal/promtailconvert/testdata/unsupported.diags
+++ b/converter/internal/promtailconvert/testdata/unsupported.diags
@@ -10,7 +10,6 @@
 (Warning) server.log_level is not supported - Flow mode components may produce different logs
 (Error) server.http_path_prefix is not supported
 (Warning) server.health_check_target disabling is not supported in Flow mode
-(Error) DropRateLimitedBatches is currently not supported in Grafana Agent Flow
 (Warning) stream_lag_labels is deprecated and the associated metric has been removed
 (Error) dockerswarm_sd_configs is not supported
 (Error) serverset_sd_configs is not supported
--- a/docs/sources/flow/reference/components/loki.write.md
+++ b/docs/sources/flow/reference/components/loki.write.md
@@ -102,6 +102,11 @@ Endpoints can be named for easier identification in debug metrics by using the
 `name` argument. If the `name` argument isn't provided, a name is generated
 based on a hash of the endpoint settings.
 
+The `retry_on_http_429` argument specifies whether `HTTP 429` status code
+responses should be treated as recoverable errors; other `HTTP 4xx` status code
+responses are never considered recoverable errors. When `retry_on_http_429` is
+enabled, the retry mechanism will be governed by the specified `backoff` configuration.
+
 ### basic_auth block
 
 {{< docs/shared lookup="flow/reference/components/basic-auth-block.md" source="agent" >}}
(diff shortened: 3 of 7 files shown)


converter/internal/promtailconvert/testdata/unsupported.diags
```diff
@@ -10,7 +10,6 @@
 (Warning) server.log_level is not supported - Flow mode components may produce different logs
 (Error) server.http_path_prefix is not supported
 (Warning) server.health_check_target disabling is not supported in Flow mode
-(Error) DropRateLimitedBatches is currently not supported in Grafana Agent Flow
```
thampiotr, line 13: Would be good to test a promtail to flow conversion, which can be done by modifying the `testdata` files.

docs/sources/flow/reference/components/loki.write.md
```diff
@@ -80,6 +80,7 @@ Name                  | Type          | Description                           |
 `proxy_url`           | `string`      | HTTP proxy to proxy requests through. | | no
 `follow_redirects`    | `bool`        | Whether redirects returned by the server should be followed. | `true` | no
 `enable_http2`        | `bool`        | Whether HTTP2 is supported for requests. | `true` | no
+`retry_on_http_429`   | `bool`        | Retry when an HTTP 429 status code is received. | `true` | no
```
thampiotr, line 83: There is a comment in `prometheus.remote_write` which I think we should port:

> The retry_on_http_429 argument specifies whether HTTP 429 status code responses should be treated as recoverable errors; other HTTP 4xx status code responses are never considered recoverable errors. When retry_on_http_429 is enabled, Retry-After response headers from the servers are honored.

Can you also check if this is accurate with regards to `Retry-After` header comment? Want to make sure we're documenting the actual behaviour.

wildum: Unless Im missing something, it seems that the code is not using the Retry-After header value in this case. It is using the backoff parameters:
`		level.Warn(c.logger).Log("msg", "error sending batch, will retry", "status", status, "tenant", tenantID, "error", err)
		c.metrics.batchRetries.WithLabelValues(c.cfg.URL.Host, tenantID).Inc()
		backoff.Wait()
`
The backoff parameters can be configured by the users with MinBackoff, MaxBackoff and MaxRetries.

converter/internal/promtailconvert/testdata/unsupported.diags
wildum: Added a test in https://github.com/grafana/agent/pull/4865/commits/89a101688c68c3354b531dddf1b8e69f05018f69. I did not add the auth tests because I would need to create several files to try the different types of auth and I think that it should be done in another PR (if we really need full coverage of the params)

thampiotr: Sounds good!

converter/internal/promtailconvert/testdata/loki_write.yaml
```diff
@@ -0,0 +1,17 @@
...
+      bird: house
+    tenant_id: sparrow
+    batchwait: 10s
+    batchsize: 1024
+    proxy_url: http://proxy.example.com
+    backoff_config:
+      min_period: 100ms
+      max_period: 10s
+      max_retries: 20
+    drop_rate_limited_batches: true
```
thampiotr, line 14: Nit: Since we're setting this to default value, maybe we should have two clients one with drop_rate_limited_batches set to true and the other set to false?

docs/sources/flow/reference/components/loki.write.md
thampiotr: Oh, I see. Thanks for checking! Indeed, I can see that we use backoff config. And prometheus.remote_write uses "Retry-After" header. 

We should adapt the comment and still add it to docs.

docs/sources/flow/reference/components/loki.write.md
```diff
@@ -101,6 +102,11 @@ Endpoints can be named for easier identification in debug metrics by using the
 `name` argument. If the `name` argument isn't provided, a name is generated
 based on a hash of the endpoint settings.
 
+The `retry_on_http_429` argument specifies whether `HTTP 429` status code
+responses should be treated as recoverable errors; other `HTTP 4xx` status code
+responses are never considered recoverable errors. When `retry_on_http_429` is
+enabled, the retry mechanism will be governed by the specified `backoff` configuration.
```
thampiotr, line 108: What is the `backoff` configuration? We don't have a block or argument with this name ;) 
```suggestion
enabled, the retry mechanism will be governed by the backoff configuration specified through `min_backoff_period`, `max_backoff_period ` and `max_backoff_retries` attributes.
```

thampiotr (approved):
LGTM % one tweak to the docs

docs/sources/flow/reference/components/loki.write.md
wildum: True, thought it was clear enough with the different backoff attributes in the table above but that makes it clearer

PR #3861, opened 2023-05-11
Make sure labels are cloned from loki.source.kubernetes

tpaschalis:
#### PR Description
We've gotten some reports of the agent crashing when using loki.source.kubernetes.

Here's a stack trace of this happening:
<details>
```
grafana-agent ts=2023-05-11T15:57:01.398384003Z level=info trace_id=559f5d0626cd43cb04fc351bc95144b3 msg="starting complete graph evaluation"
grafana-agent ts=2023-05-11T15:57:01.398661803Z level=info msg="now listening for http traffic" addr=0.0.0.0:80
grafana-agent ts=2023-05-11T15:57:01.398753303Z level=info msg="running usage stats reporter"
grafana-agent ts=2023-05-11T15:57:01.398922603Z component=discovery.kubernetes.services level=info msg="Using pod service account via in-cluster config"
grafana-agent ts=2023-05-11T15:57:01.399386203Z level=info trace_id=559f5d0626cd43cb04fc351bc95144b3 msg="finished node evaluation" node_id=discovery.kubernetes.services duration=676.4µs
grafana-agent ts=2023-05-11T15:57:01.399569803Z level=info trace_id=559f5d0626cd43cb04fc351bc95144b3 msg="finished node evaluation" node_id=discovery.relabel.ksm duration=161.6µs
grafana-agent ts=2023-05-11T15:57:01.399655103Z level=info trace_id=559f5d0626cd43cb04fc351bc95144b3 msg="finished node evaluation" node_id=discovery.relabel.agent duration=70.5µs
grafana-agent ts=2023-05-11T15:57:01.399780503Z level=info trace_id=559f5d0626cd43cb04fc351bc95144b3 msg="finished node evaluation" node_id=discovery.relabel.node_exporter duration=112.2µs
grafana-agent ts=2023-05-11T15:57:01.399868103Z level=info trace_id=559f5d0626cd43cb04fc351bc95144b3 msg="finished node evaluation" node_id=discovery.relabel.webapp duration=75µs
grafana-agent ts=2023-05-11T15:57:01.399981303Z level=info trace_id=559f5d0626cd43cb04fc351bc95144b3 msg="finished node evaluation" node_id=discovery.relabel.kubelet duration=100.3µs
grafana-agent ts=2023-05-11T15:57:01.400305203Z level=info trace_id=559f5d0626cd43cb04fc351bc95144b3 msg="finished node evaluation" node_id=loki.write.grafana_cloud_loki duration=308.5µs
[...]

PR #4866, opened 2023-08-21
loki.source.docker: reduce labels-related resource usage

### LINKED ISSUE #4777: CPU consumption for grafana-agent-flow loki scraping
### What's wrong?

I'm using grafana-agent-flow to fetch logs from docker containers and upload them to loki. Grafana is running on host with just 2 docker containers and generates ~ 10 messages per second, but CPU usage looks to high (especially if you will compare it to promtail). Result for ps aux

```
grafana+ 1642768 82.3  4.2 1563584 83032 ?       Ssl  09:59 135:37 /usr/bin/grafana-agent-flow run --server.http.listen-addr=10.20.1.46:12345 --storage.path=/var/lib/grafana-agent-flow /etc/grafana-agent-flow.river
```

Promtail on host with same containers:
```
root        8871  9.5  0.0 6024040 55496 ?       Ssl  Apr13 16394:29 /usr/bin/promtail -config.file=/etc/promtail/config.yml
```

### Steps to reproduce

Tried same config file on 5 hosts in our system with different containers number - grafana-agent-flow consumes more CPU than promtail instance.

### System information

Linux 5.15.0-56-generic, Ubuntu 22.04.1 LTS

### Software version

v0.35.3

### Configuration

```text
logging {
    level = "warn"
}


// we scrape only containers with promtail=enabled label
discovery.docker "containers" {
    host = "unix:///var/run/docker.sock"
    filter {
        name   = "label"
        values = [ "promtail=enabled" ]
    }
}

discovery.relabel "docker" {
    targets = discovery.docker.containers.targets
}

loki.source.docker "default" {
    host       = "unix:///var/run/docker.sock"
    targets    = discovery.relabel.docker.output

    forward_to = [ loki.process.preprocess.receiver ]
}


loki.process "preprocess" {
    forward_to = [ loki.write.loki.receiver ]

    stage.metrics {
        metric.counter {
            name        = "log_lines_read"
            description = "total number of log lines"
            prefix      = "zm_"

            match_all         = true
            action            = "inc"
            max_idle_duration = "24h"
        }
    }

    stage.drop {
        older_than  = "1h"
        drop_counter_reason = "record is too old"
    }

    stage.drop {
        longer_than = "3KB"
        drop_counter_reason = "size is too big"
    }


    stage.metrics {
        metric.counter {
            name        = "log_lines_processed"
            description = "total number of log lines"
            prefix      = "zm_"

            match_all         = true
            action            = "inc"
            max_idle_duration = "24h"
        }
    }
}

loki.write "loki" {
    endpoint {
        url = "http://10.20.1.100:3100/loki/api/v1/push"
    }
}
```

### Logs

_No response_

tpaschalis:
#### PR Description
This PR refactors the usage of labels on the `loki.source.docker` containers to reduce CPU usage by ~84% and memory usage by ~91% (`main` vs `fix`)

Screenshot 2023-08-21 at 16 29 26
Screenshot 2023-08-21 at 16 29 03

Fixes #4777

#### Notes to the Reviewer
I've verified this by using our pyroscope components and checking the profiles on Grafana Cloud with a config like the following

```
pyroscope.scrape "local" {
  targets    = [
    {"__address__" = "localhost:12346", "service_name"="main"},
    {"__address__" = "localhost:12347", "service_name"="fix"},
  ]
  forward_to = [pyroscope.write.grafanacloud.receiver]

  profiling_config {
    profile.process_cpu {
      enabled = true
    }
  }
}

pyroscope.write "grafanacloud" {
  endpoint {
    url = "<url>"
    basic_auth {
      username = "<user>"
      password = "<pass>"
    }
  }
  external_labels = {
    "env" = "tpaschalis-comparison",
  }
}
```

#### PR Checklist
- [X] CHANGELOG.md updated
- [ ] Tests updated

### DIFF
--- a/component/loki/source/docker/internal/dockertarget/target.go
+++ b/component/loki/source/docker/internal/dockertarget/target.go
@@ -43,6 +43,7 @@ type Target struct {
 	positions     positions.Positions
 	containerName string
 	labels        model.LabelSet
+	labelsStr     string
 	relabelConfig []*relabel.Config
 	metrics       *Metrics
 
@@ -55,7 +56,8 @@ type Target struct {
 
 // NewTarget starts a new target to read logs from a given container ID.
 func NewTarget(metrics *Metrics, logger log.Logger, handler loki.EntryHandler, position positions.Positions, containerID string, labels model.LabelSet, relabelConfig []*relabel.Config, client client.APIClient) (*Target, error) {
-	pos, err := position.Get(positions.CursorKey(containerID), labels.String())
+	labelsStr := labels.String()
+	pos, err := position.Get(positions.CursorKey(containerID), labelsStr)
 	if err != nil {
 		return nil, err
 	}
@@ -71,6 +73,7 @@ func NewTarget(metrics *Metrics, logger log.Logger, handler loki.EntryHandler, p
 		positions:     position,
 		containerName: containerID,
 		labels:        labels,
+		labelsStr:     labelsStr,
 		relabelConfig: relabelConfig,
 		metrics:       metrics,
 
@@ -231,7 +234,7 @@ func (t *Target) process(r io.Reader, logStream string) {
 		// problematic if we have the same container with a different set of
 		// labels (e.g. duplicated and relabeled), but this shouldn't be the
 		// case anyway.
-		t.positions.Put(positions.CursorKey(t.containerName), t.labels.String(), ts.Unix())
+		t.positions.Put(positions.CursorKey(t.containerName), t.labelsStr, ts.Unix())
 	}
 }
 
@@ -259,9 +262,9 @@ func (t *Target) Ready() bool {
 	return t.running.Load()
 }
 
-// Labels reports the target's labels.
-func (t *Target) Labels() model.LabelSet {
-	return t.labels
+// LabelsStr returns the target's original labels string representation.
+func (t *Target) LabelsStr() string {
+	return t.labelsStr
 }
 
 // Name reports the container name.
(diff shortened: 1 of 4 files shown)


component/loki/source/docker/internal/dockertarget/target.go
```diff
@@ -200,24 +203,8 @@ func (t *Target) process(r io.Reader, logStream string) {
...
-		for _, lbl := range processed {
-			if strings.HasPrefix(lbl.Name, "__") {
-				continue
-			}
-			filtered[model.LabelName(lbl.Name)] = model.LabelValue(lbl.Value)
-		}
-
 		t.handler.Chan() <- loki.Entry{
-			Labels: filtered,
+			Labels: logStreamLset,
```
thampiotr, line 207: The only concern I'd have here is that may be we should not give away a shared, mutable copy of this. Could use `maps.Copy()`.

tpaschalis: To be honest, we've had a similar [issue](https://github.com/grafana/agent/pull/3861) in the past, but given the overall [promise](https://github.com/grafana/agent/blob/4b8807a517071556c7fd520d7ddd792fbcd311f7/component/component.go#L36-L45) to treat arguments as immutable we decided to only copy/clone on the downstream components that could change the label set (today, only loki.process and loki.relabel), so I don't think it's necessary to do it here as well. WDYT?

tpaschalis: Also, the idea is that we could further reduce the copying-related resource usage with a copy-on-write model eg. only for actual relabeling rules or processing stages that _do_ change the labels, instead of doing it preemptively on every entry passing through.
https://github.com/grafana/agent/blob/8bcc595ecea6b41a83193065a91be17395bb5b91/component/loki/process/process.go#L139-L144

thampiotr: Fair enough - I'm happy with this approach if that's our deliberate choice :) 

One day maybe we could have an immutable wrapper that would not copy? 🤔 not sure if that would be more efficient than copying in our case and how would the implementation look in Go.

PR #4870, opened 2023-08-21
service/cluster: migrate pkg/cluster to a Flow-native service

rfratto:
> **NOTE**: As with all service-related PRs, I recommend reviewing this PR commit-by-commit. 

_This PR is currently opened as a draft to get early feedback._

This PR introduces `service/cluster`, a Flow-focused rewrite of the previous `pkg/cluster` implementation with the following changes:

* service/cluster always starts a ckit.Node for consistency around the API. When clustering is disabled, service/cluster never joins any peers, and never exposes its cluster HTTP handlers to peers in the cluster.

* Much of the behavior previously provided by pkg/cluster is now deferred to the caller:

* The caller is expected to determine their own advertise address, including defaulting to a specific interface.

* The caller is expected to provide their own function to discover peers, even if that function returns a static list.

* The new implementation, as a Flow service, directly informs components about changes to the cluster state instead of having the Flow controller inform components.

While the new implementation puts more work on the caller, its API is simplified and easier to understand. For example, mutually exclusive mechanisms of the previous API (such as configuring JoinPeers and DiscoverPeers) have been simplified into a single, unified mechanism.


rfratto:
I'm opening this up for review but tomorrow I need to do some manual smoke testing to make sure everything still works as we expect.

service/cluster/cluster.go
```diff
@@ -0,0 +1,356 @@
...
+	var wg sync.WaitGroup
+	defer wg.Wait()
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	s.node.Observe(ckit.FuncObserver(func(peers []peer.Peer) (reregister bool) {
+		if ctx.Err() != nil {
+			return false
+		}
```
tpaschalis, line 198: Is this and the other call to `ctx.Err()` below to avoid calling a NotifyClusterChange once we're setting our state to terminating, or just to clean up the Observer once the parent context is exiting?

rfratto: Two reasons:

* We allow ourselves to clean up the Observer once the cluster service is no longer running.
* We allow ourselves to stop early instead of forcing iterating over all components so we can terminate sooner.

Do you want me to add this as comments?

service/cluster/cluster.go
```diff
@@ -0,0 +1,356 @@
...
+//
+// Distribution accuracy measures how close a node was to being responsible for
+// exactly 1/N keys during simulation. Simulation tests used a cluster of 10
+// nodes and hashing 100,000 random keys:
+//
+//	512 tokens per node: min 96.1%, median 99.9%, max 103.2% (stddev: 197.9 hashes)
+const tokensPerNode = 512
+
+// ServiceName defines the name used for the cluster service.
+const ServiceName = "cluster"
```
tpaschalis, line 45: IIRC we should also add this as a dependency to all module loaders, right?

service/cluster/cluster.go
tpaschalis: Yeah, it wouldn't hurt!

tpaschalis (approved):
LGTM from a code perspective, just the one nit I posted.

I will sanity-check with this tomorrow as well, but it looks much cleaner and pleasant to use 🙌

service/cluster/cluster.go
```diff
@@ -0,0 +1,356 @@
...
+	if err := s.node.ChangeState(ctx, peer.StateTerminating); err != nil {
+		level.Error(s.log).Log("msg", "failed to change state to Terminating", "err", err)
+	}
+
+	if err := s.node.Stop(); err != nil {
+		level.Error(s.log).Log("msg", "failed to gracefully stop node", "err", err)
+	}
+}
+
+// Update implements [service.Service]. It is a no-op since the cluster service
```
ptodev, line 306: I don't think an error counts as a no-op :) I agree we should have an error though.

cmd/internal/flowmode/cluster_builder.go
```diff
@@ -0,0 +1,179 @@
...
+
+	if config.AdvertiseAddress == "" {
+		// TODO(rfratto): allow advertise interfaces to be configurable.
+		addr, err := advertise.FirstAddress(advertise.DefaultInterfaces)
+		if err != nil {
+			return nil, fmt.Errorf("determining advertise address: %w", err)
+		}
+		config.AdvertiseAddress = fmt.Sprintf("%s:%d", addr.String(), listenPort)
+	} else {
+		config.AdvertiseAddress = appendDefaultPort(config.AdvertiseAddress, listenPort)
```
ptodev, line 64: I didn't see any place where we log the `AdvertiseAddress`? Seems like a useful thing to log.

ptodev (approved):
LGTM, I added a few minor comments.

cmd/internal/flowmode/cluster_builder.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	switch {
+	case len(opts.JoinPeers) > 0 && opts.DiscoverPeers != "":
+		return nil, fmt.Errorf("at most one of join peers and discover peers may be set")
```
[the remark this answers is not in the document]
rfratto: Yes. That would be desirable if you are creating a seed node.

PR #4877, opened 2023-08-22
Set retry_on_http_429 to true by default prometheus.remote_write queue

### LINKED ISSUE #4441: Make retry on 429 default to true
### Request

Set the retry on 429 to default to true so that the agent can retry  sending and avoid data being lost.

### Use case

We should default to data safety

wildum:
#### PR Description
Set retry_on_http_429 to true by default in the queue config block of prometheus.remote_write.

Fixes #4441 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@ Main (unreleased)
 - `otelcol.exporter.prometheus`: Set `include_scope_info` to `false` by default. You can set 
   it to `true` to preserve previous behavior. (@gouthamve)
 
+- `prometheus.remote_write`: Set `retry_on_http_429` to `true` by default in the `queue_config` block.
+  You can set it to `false` to preserve previous behavior. (@wildum)
+
 ### Enhancements
 
 - Integrations: include `direct_connect`, `discovering_mode` and `tls_basic_auth_config_path` fields for MongoDB configuration. (@gaantunes)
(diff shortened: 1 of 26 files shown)


CHANGELOG.md
```diff
@@ -23,6 +23,9 @@ Main (unreleased)
 - `otelcol.exporter.prometheus`: Set `include_scope_info` to `false` by default. You can set 
   it to `true` to preserve previous behavior. (@gouthamve)
 
+- `prometheus.remote_write`: Set `retry_on_http_429` to `true` by default in the `queue_config` block.
+  You can set it to `false` to preserve previous behavior. (@wildum)
```
thampiotr, line 27: I think we always add migration instructions to the upgrade guide as described above. Could you add some there? I think that we should just say that no action is required if you were setting the `retry_on_http_429` explicitly. If you did not set it explicitly, but you would like to _not_ retry on HTTP 429, you should set it to false as you upgrade to this version.

tpaschalis:
I think it makes sense to also have the setting default to `true` on static mode, WDYT?

wildum:
> I think it makes sense to also have the setting default to `true` on static mode, WDYT?

The [default in prometheus is false](https://prometheus.io/docs/prometheus/2.42/configuration/configuration/#remote_write). Is it possible to set a different default value in static mode?

PR #4884, opened 2023-08-22
service/http: add support for TLS

### LINKED ISSUE #2715: Flow: support TLS config similar to static agent

rfratto:
This commit allows the http service to be configured with TLS.

TLS may be dynamically enabled by configuring the `tls` block in the `http` service:

```river
http {
  tls {
    ...
  }
}
```

Removing this block will dynamically disable TLS.

Services are guaranteed to evaluate their configuration before running, so if the initial configuration file configures TLS, then TLS will always be running prior to the HTTP server starting.

Existing open connections will continue to communicate over plaintext if TLS is enabled on an existing Flow process where TLS was previously not enabled. These connections will eventually be closed after the maximum connection lifetime is reached.

Closes #2715.

## Notes

* This does not include support for the Windows certificate store, which will be added in a follow-up commit.
* Enabling TLS will break clustering, since clustering has no way of creating a TLS client. This will be resolved in a follow up after #4870 is merged.

## TODO 

- [x] Documentation
- [x] Update changelog 
- [x] Determine whether or not we need to flush existing connections when toggling between TLS/not TLS.


rfratto:
> Determine whether or not we need to flush existing connections when toggling between TLS/not TLS.

This was discussed offline, and we decided (at least for now) to document this as an edge case, since it will be unlikely to turn on TLS while the agent is running. 

Just to get a feel for a potential solution, I prototyped flushing connections when changing listeners. It seemed to improve the behavior but not fully fix it when looking at the flow UI. It's probably something I did wrong, though.

rfratto:
I'm on PTO tomorrow so I won't be able to respond to feedback, but I don't plan to include this in the current release cycle, since I also want to make sure it gets released alongside TLS support for cluster clients too.

thampiotr:
> Existing open connections will continue to communicate over plaintext if TLS is enabled on an existing Flow process where TLS was previously not enabled.

Just an idea: maybe we can return a `RecoverableByRestartError` from `Update` which would signal to Flow vm to shutdown and recreate & restart the component. In the medium term I'm happy for this to be documented as an exception though.

service/http/http.go
```diff
@@ -47,12 +48,26 @@ type Options struct {
...
 	opts     Options
 
+	// lazyLis and lazyNetLis are used to lazily enable TLS.
+	//
+	// lazyNetLis eventually wraps around the real network listener once the
+	// Service is running.
+	//
+	// lazyLis either wraps around a TLS listener or lazyNetLis, depending on
+	// whether TLS is currently enabled.
+	lazyLis, lazyNetLis *lazyListener
```
thampiotr, line 69: Maybe we can try some different names to make this easier to follow 🤔 
* lazyLis -> publicLis or outerLis
* lazyNetLis -> tcpLis or innerLis

service/http/http.go
```diff
@@ -85,14 +100,24 @@ func New(opts Options) *Service {
 		n = opts.Clusterer.Node
 	}
 
+	var (
+		lazyNetLis = &lazyListener{}
+		lazyLis    = &lazyListener{}
+	)
+
+	// lazyLis should default to wrapping around lazyNetLis.
+	_ = lazyLis.SetInner(lazyNetLis)
```
thampiotr, line 109: The `lazyLis` points to `lazyNetLis` which has `nil` `inner`. 

So I don't think that any of them is really usable, since the inner-most listener is `nil`. 

Since they are not usable until the `Update` is called, maybe we need only one `lazyListener` - the `lazyLis` and a simple reference to `Listener` that is nil in the beginning until `Run` is called where we would set it to the `netLis` from line 144 🤔 

In other words, I'm just wondering if we need two lazyListeners if one of them has `SetInner` called only once.

service/http/http.go
```diff
@@ -299,3 +348,78 @@ type Component interface {
 	// will receive a request to just `/metrics`.
 	Handler() http.Handler
 }
+
+// lazyListener is a [net.Listener] which lazily initializes the underlying
+// listener.
```
thampiotr, line 353: Should we put it in its own file?

thampiotr (commented):
I think a few tests would be a good idea for this change.

docs/sources/flow/reference/config-blocks/http.md
```diff
@@ -0,0 +1,140 @@
...
+tls | [tls][] | Define TLS settings for the HTTP server. | no
+
+[tls]: #tls-block
+
+### tls block
+
+The `tls` block configures TLS settings for the HTTP server.
+
+{{% admonition type="warning" %}}
+If the `tls` block is provided in a reload after Grafana Agent is already
```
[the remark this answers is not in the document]
clayton-cornell: Missed a word in the suggestion. It probably should be `the configuration is reloaded`

clayton-cornell (commented):
This ref doc looks OK. Nothing is jumping out at me to fix or update.

service/http/http.go
```diff
@@ -299,3 +348,78 @@ type Component interface {
...
+
+func (lis *lazyListener) Addr() net.Addr {
+	lis.mut.RLock()
+	defer lis.mut.RUnlock()
+
+	if lis.inner == nil {
+		// TODO(rfratto): it's not sure if this will cause problems. If this is an
+		// issue, we can do one of two things to address this:
+		//
+		// 1. Return a fake address.
```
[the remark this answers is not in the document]
rfratto: I'm just not sure this code ever gets called, which is why I'm leaving it as a TODO. Option 2 would take more time to implement since it's not simple, so I'd prefer to leave this as-is and we can revisit if it turns out we do hit panics.

service/http/http.go
rfratto: I'm not sure it's worth it, it's only 70 lines of code.

service/http/http.go
rfratto: > In other words, I'm just wondering if we need two lazyListeners if one of them has SetInner called only once.

The main reason is so we don't start listening for network traffic until the service runs. Doing otherwise feels like it breaks the conventions of a Service: don't do anything until Run is called.

service/http/http.go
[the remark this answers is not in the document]
thampiotr: I see. Since its a new feature, it's probably fine as-is. Would be good idea to start using it right away in some test deployments.

service/http/http_test.go
```diff
@@ -0,0 +1,216 @@
...
+
+		resp, err := cli.Do(req)
+		require.NoError(t, err)
+		defer resp.Body.Close()
+
+		require.Equal(t, http.StatusOK, resp.StatusCode)
+	})
+}
+
+func Test_Toggle_TLS(t *testing.T) {
```
thampiotr, line 82: Nice!

thampiotr (approved):
LGTM

docs/sources/flow/reference/config-blocks/http.md
```diff
@@ -0,0 +1,143 @@
...
+is used. The set of elliptic curves specified may be from the following:
+
+* `CurveP256`
+* `CurveP384`
+* `CurveP521`
+* `X25519`
+
+The `min_version` and `max_version` arguments determine the oldest and newest
+TLS version that is acceptable from clients. If not provided, a default value
+is used.
```
tpaschalis, line 136: Which is the default value used?

docs/sources/flow/reference/config-blocks/http.md
```diff
@@ -0,0 +1,143 @@
...
+To ensure all connections use TLS, start Grafana Agent with the `tls` block
+already configured.
+{{% /admonition %}}
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`cert_pem` | `string` | PEM data of the server TLS certificate. | `""` | conditionally
+`cert_file` | `string` | Path to the server TLS certificate on disk. | `""` | conditionally
+`key_pem` | `string` | PEM data of the server TLS key. | `""` | conditionally
+`key_file` | `string` | Path to the server TLS key on disk. | `""` | conditionally
```
mattdurham, line 67: Is the reason we are enabling file is so that it can check it on each access like it does today? Normally I would say use local.file or remote.file and only use expose *_pem.

rfratto: That, plus parity and alignment in general: all of our other TLS-related settings support files plus strings.

docs/sources/flow/reference/config-blocks/http.md
```diff
@@ -0,0 +1,143 @@
...
+* `VerifyClientCertIfGiven`: requests clients to send an optional certificate. If a certificate is sent, it must be valid.
+* `RequireAndVerifyClientCert`: requires clients to send a valid certificate.
+
+The `client_ca_pem` or `client_ca_file` arguments may be used to perform client
+certificate validation. These arguments may only be provided when `client_auth`
+is not set to `NoClientCert`.
+
+The `cipher_suites` argument determines what cipher suites to use. If not
+provided, a default list is used. The set of cipher suites specified may be
+from the following:
```
mattdurham, line 105: There is probably some nuance about boringcrypto here. Not sure if it needs a call out though. IE boring crypto may restrict this further.

mattdurham (approved):
LGTM with some minor questions

docs/sources/flow/reference/config-blocks/http.md
```diff
@@ -0,0 +1,143 @@
...
+The following pairs of arguments are mutually exclusive, and only one may be
+configured at a time:
+
+* `cert_pem` and `cert_file`
+* `key_pem` and `key_file`
+* `client_ca_pem` and `client_ca_file`
+
+The `client_auth` argument determines whether to validate client certificates.
+The default value, `NoClientCert`, indicates that no client certificate
+validation is performed.
```
tpaschalis, line 89: We could maybe explicitly disallow `client_ca_pem/client_ca_file` if `client_auth` is not set, or set to `NoClientCert`?

tpaschalis (approved):
LGTM ⛵

docs/sources/flow/reference/config-blocks/http.md
tpaschalis: Oh, I see. Nevermind, not a big deal either way.

docs/sources/flow/reference/config-blocks/http.md
```diff
@@ -0,0 +1,148 @@
...
+The following pairs of arguments are mutually exclusive, and only one may be
+configured at a time:
+
+* `cert_pem` and `cert_file`
+* `key_pem` and `key_file`
+* `client_ca_pem` and `client_ca_file`
+
+The `client_auth` argument determines whether to validate client certificates.
+The default value, `NoClientCert`, indicates that no client certificate
+validation is performed. The `client_ca_pem` and `client_ca_file` may only be
```
clayton-cornell, line 89:
```suggestion
The default value, `NoClientCert`, indicates that the client certificate
is not validated. The `client_ca_pem` and `client_ca_file` may only be
```

Is this still correct with this small change?

docs/sources/flow/reference/config-blocks/http.md
```diff
@@ -0,0 +1,148 @@
...
+configured when `client_auth` is not `NoClientCert`.
+
+The following values are accepted for `client_auth`:
+
+* `NoClientCert`: client certificates are neither requested nor verified.
+* `RequestClientCert`: requests clients to send an optional certificate.
+* `RequireAnyClientCert`: requires at least one certificate from clients that is not checked for validity.
```
clayton-cornell, line 96: Does this mean that you MUST use a cert that has not been validated? Or that any N certs from the clients are allowed to be used without checking the validity?

clayton-cornell (commented):
Some doc suggestions and questions

docs/sources/flow/reference/config-blocks/http.md
rfratto: Close enough. Users can go to the detailed descriptions for full information.

docs/sources/flow/reference/config-blocks/http.md
rfratto: Made some changes in #4997, can we move this discussion over there?

PR #4886, opened 2023-08-22
fix bug with integer targetPort in prometheus.operator.*

### LINKED ISSUE #4707: Flow mode creates duplicate scrape configs for `ServiceMonitor` in some scenarios
### What's wrong?

It seems that when you use `targetPort` with the flow component `prometheus.operator.servicemonitor`, you will get duplicate scrape configurations added which point to container ports that do not expose metrics.

You can work around this by changing your `ServiceMonitor` to use `port` with a string value instead of `targetPort` with a integer value.

### Steps to reproduce

You can deploy the `cert-manager` helm chart or create a `ServiceMonitor` which uses `targetPort` targeting a container that has more than one container port

```yaml
apiVersion: monitoring.coreos.com/v1                                                                                                                         
kind: ServiceMonitor                                                                                                                                         
metadata:                                                                                                                                                                                                                                                                   
  name: cert-manager                                                                                                                                         
  namespace: cert-manager                                                                                                                                
spec:                                                                                                                                                        
  endpoints:                                                                                                                                                 
  - honorLabels: false                                                                                                                                       
    interval: 60s                                                                                                                                            
    path: /metrics                                                                                                                                           
    targetPort: 9402                                                                                                              
    scrapeTimeout: 30s                                                                                                                                       
  jobLabel: cert-manager                                                                                                                                     
  selector:                                                                                                                                                  
    matchLabels:                                                                                                                                             
      app.kubernetes.io/component: controller                                                                                                                
      app.kubernetes.io/instance: cert-manager                                                                                                               
      app.kubernetes.io/name: cert-manager    
```

What you'll see is even though we are targeting port `9402` specifically, if the container has any other container ports exposed they will also be added as scrape configuration to Mimir. You will see duplicate endpoints when you query and of course one will be down (in this change the one that has instance `9403`, a container port) because no metrics are meant to be scraped there.

![image](https://github.com/grafana/agent/assets/5280764/921876cc-91ec-438e-a832-f8a806c8a955)


### System information

_No response_

### Software version

0.35.2

### Configuration

```text
N/A
```


### Logs

```text
N/A
```

captncraig:
Fixes #4707

### DIFF
--- a/component/prometheus/operator/configgen/config_gen_podmonitor_test.go
+++ b/component/prometheus/operator/configgen/config_gen_podmonitor_test.go
@@ -21,6 +21,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"gopkg.in/yaml.v3"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
 func TestGeneratePodMonitorConfig(t *testing.T) {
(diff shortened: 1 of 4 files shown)


component/prometheus/operator/configgen/config_gen_podmonitor_test.go
```diff
@@ -247,15 +361,15 @@ func TestGeneratePodMonitorConfig(t *testing.T) {
...
+	for _, tc := range suite {
 		t.Run(tc.name, func(t *testing.T) {
 			cg := &ConfigGenerator{
 				Client: &kubernetes.ClientArguments{},
 				AdditionalRelabelConfigs: []*flow_relabel.Config{
 					{TargetLabel: "__meta_foo", Replacement: "bar"},
 				},
 			}
-			cfg, err := cg.GeneratePodMonitorConfig(tc.m, tc.ep, i)
+			cfg, err := cg.GeneratePodMonitorConfig(tc.m, tc.ep, 1)
```
marctc, line 372: Is this change intended? If that's the case, I would like to understand why

captncraig: The number is just a counter in case there are multiple endpoints in a monitor, so they all have unique job names. Each test case incremented it for some reason, but that gets annoying if you change the order of the tests, and have to change the expected values. This makes it so each test uses the same value, but they don't depend on each other.

marctc: Ah, this is a test :facepalm:! Thanks for the explanation anyways!

marctc (approved):

PR #4872, opened 2023-08-21
Upgrade to YACE 0.54.0

ashrayjain:
#### PR Description
Upgrade YACE to 0.54.0.

#### PR Checklist

- [x] CHANGELOG.md updated

PR #4891, opened 2023-08-22
Handle cache refreshing and clearing for clients passed into YACE

ashrayjain:
This was broken here: https://github.com/nerdswords/yet-another-cloudwatch-exporter/pull/974 when we upgrade to YACE 0.54.0

> I think this is technically a breaking change to the library entrypoint since any consumer providing a cache based implementation is now required to handle that before calling the entrypoint.

### DIFF
--- a/pkg/integrations/cloudwatch_exporter/cloudwatch_exporter.go
+++ b/pkg/integrations/cloudwatch_exporter/cloudwatch_exporter.go
@@ -16,12 +16,20 @@ import (
 	"github.com/grafana/agent/pkg/integrations/config"
 )
 
+type cachingFactory interface {
+	yaceClients.Factory
+	Refresh()
+	Clear()
+}
+
+var _ cachingFactory = &yaceClientsV1.CachingFactory{}
+
 // exporter wraps YACE entrypoint around an Integration implementation
 type exporter struct {
-	name         string
-	logger       yaceLoggerWrapper
-	sessionCache yaceClients.Factory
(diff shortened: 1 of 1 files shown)


pkg/integrations/cloudwatch_exporter/cloudwatch_exporter.go
```diff
@@ -43,13 +51,19 @@ func (e *exporter) MetricsHandler() (http.Handler, error) {
 	h := http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 		e.logger.Debug("Running collect in cloudwatch_exporter")
 
+		// since we have called refresh, we have loaded all the credentials
+		// into the clients and it is now safe to call concurrently. Defer the
+		// clearing, so we always clear credentials before the next scrape
+		e.cachingClientFactory.Refresh()
+		defer e.cachingClientFactory.Clear()
```
ashrayjain, line 58: This matches what was previously being done in YACE itself https://github.com/nerdswords/yet-another-cloudwatch-exporter/blob/v0.51.0/pkg/job/scrape.go#L32

ashrayjain:
@tpaschalis tagging you since you also reviewed https://github.com/grafana/agent/pull/4872, which introduced this break.

tpaschalis:
Oops, seems easy to miss. LGTM, but cc-ing @thepalbi who has more context on the exporter for a second pair of eyes.

PR #4893, opened 2023-08-23
Fix the CI step blocking the CD pipeline to publish technical documentation

jdbaldry:
The PR CI was already fixed in https://github.com/grafana/agent/pull/4814 but I neglected to update the CD pipelines.

Reviewers should be able to see that the publish build steps match the ones in PR CI again: https://github.com/grafana/agent/blob/main/.github/workflows/check_docs.yml#L9-L15


tpaschalis (approved):
Thanks!

PR #4897, opened 2023-08-23
Add updating OTEL docs as part of the release

mattdurham:
Grafana Agent is now listed as an Otel distribution and we should update the docs there to reflect that. The reason I think this needs to be a release and not a normal PR, is that we dont want to advertise that the Agent has the components until its released in at least an RC.


ptodev (approved):

docs/developer/release/10-update-otel.md
```diff
@@ -0,0 +1,13 @@
+# Update Open Telemetry Contrib
+
+Grafana Agent is listed has a distribution for Open Telemetry. If there are any new OTel components then Grafana Agent needs to be associated with then open a PR in [OpenTelemetry Contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib) and add the Agent to the list of distributions. [Example](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/653ab064bb797ed2b4ae599936a7b9cfdad18a29/receiver/kafkareceiver/README.md?plain=1#L7)
```
jpkrohling, line 3:
```suggestion
Grafana Agent is listed as a distribution of the OpenTelemetry Collector. If there are any new OTel components that Grafana Agent needs to be associated with, then open a PR in [OpenTelemetry Contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib) and add the Agent to the list of distributions. [Example](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/653ab064bb797ed2b4ae599936a7b9cfdad18a29/receiver/kafkareceiver/README.md?plain=1#L7)
```

jpkrohling (approved):

Issue #4251, opened 2023-06-23
Support decoupled scraping in the cloudwatch exporter integration


dtrejod:
### Request

The grafana-agent `cloudwatch_exporter` integration vendors the OSS YACE:
https://github.com/nerdswords/yet-another-cloudwatch-exporter

YACE supports a concept of decoupled-scraping. Decoupled-scraping is the default, and instructs YACE to spin up an async [go-routine](https://github.com/nerdswords/yet-another-cloudwatch-exporter/blob/be9950213af29f00465867983cb2e897350f6063/cmd/yace/scraper.go#L37) that periodically (every 5m) calls `UpdateMetrics()` instead of calling this func every time a request comes into the sever handler. The implementation in grafana-agent makes it so `UpdateMetrics()` is [synchronous](https://github.com/grafana/agent/blob/92f9ee8cc8d16e6cfcd83c1b6e03a370eccc2495/pkg/integrations/cloudwatch_exporter/cloudwatch_exporter.go#L46) with the server handler. 

I'd like to see grafana-agent also support the decoupled-scraping technique used by YACE. 

### Use case

Decoupled-scraping ensures the number of requests that hit Cloudwatch APIs is not tied to the `scrape_interval` set in the grafana-agent. The number of API calls to cloudwatch APIs is linked to the billing calling these APIs too frequently.

marctc:
cc @thepalbi

dtrejod:
Closed by https://github.com/grafana/agent/pull/4914

PR #4914, opened 2023-08-23
Support decoupled scraping in cloudwatch_exporter integration.

dtrejod:
#### PR Description

Decouple the collection of cloudwatch metrics from the actual scrape on the `/metrics` endpoint. Decouple scraping is the default behavior in the upstream [YACE](https://github.com/nerdswords/yet-another-cloudwatch-exporter/blob/master/README.md) repo. 

In the grafana-agent implementation of decouple-scraping, I made the feature opt-in. The previous behavior is preserved unless users add to the `cloudwatch_exporter` integration the following config:
```
integrations:
    cloudwatch_exporter: 
        decoupled_scraping:
            enabled: true
```

With the above configuration decouple-scraping will be enabled. Scrapes from cloudwatch will occur every 5m by default unless overriden by `integrations.cloudwatch_exporter.decoupled_scraping.scraping_interval`.

https://github.com/grafana/agent/issues/4251

#### Notes to the Reviewer

The existing integration didn't have any tests. I was able to validate the PR here runs in a dev environment I have. Happy to add tests if given guidance on what those would look like in this repo. 

Also not sure what is meant by `Documentation added`? I think docs are auto-generated from docstrings? If so there is documentation for the new feature. 

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


pkg/integrations/cloudwatch_exporter/config.go
```diff
@@ -17,10 +17,11 @@ import (
...
 const (
-	metricsPerQuery       = 500
-	cloudWatchConcurrency = 5
-	tagConcurrency        = 5
-	labelsSnakeCase       = false
+	metricsPerQuery                  = 500
+	cloudWatchConcurrency            = 5
+	tagConcurrency                   = 5
+	labelsSnakeCase                  = false
+	defaultDecoupledScrapingInterval = time.Minute * 5
```
dtrejod, line 24: 5m is the default scrape interval in YACE:
https://github.com/nerdswords/yet-another-cloudwatch-exporter/blob/master/cmd/yace/main.go#L133

marctc:
> Also not sure what is meant by `Documentation added`? I think docs are auto-generated from docstrings? If so there is documentation for the new feature.

`Documentation added` checkbox it's a reminder in the case that PR needs to included changes in the docs. In this case, it's applicable as you added a new field, which would be needed to be documented.

I'd defer the review to @thepalbi, but this PR needs to add the same fields for the flow component. Could you do that as well? thanks!

mattdurham:
@dtrejod I cannot add this since its a fork but can you add this to the doc in cloudwatch-exporter-config?

```
# Instead of retrieving metrics on request, decouple scraping retrieves the metrics on a schedule and returns the cached metrics.
  decoupled_scraping:
    # Enable decoupled scraping.
    [enabled: <boolean> | default = false ]

    # How often to scrape the endpoint.
    [scrape_interval: <duration> | default = "5m"]
```

dtrejod:
> Documentation added checkbox it's a reminder in the case that PR needs to included changes in the docs. In this case, it's applicable as you added a new field, which would be needed to be documented.
> I cannot add this since its a fork but can you add this to the doc in cloudwatch-exporter-config?

Thanks both. I missed the "docs" pkg when looking at the root dir structure of this repo. Added docs to both static and flow integrations. 

Also updated the flow integration to also have the `decoupled scraping` feature. 

> Can I ask what's the driver for preferring the scrape to be asynchronous (like yace does natively) vs. during the /metrics call?

The main driver is to fix `write i/o timeout` errors we are seeing in some of our largest AWS cloud accounts. Today we get errors such as:
```
GET /integrations/cloudwatch_exporter/metrics 53.47008076s, error: write tcp 127.0.0.1:5080->127.0.0.1:55386: i/o timeout
```

When tried a few idea to work around the issue such as bumping the `scrape_interval`/`scrape_timeout` for the component integration to `5m`. However this causes the cloudwatch metrics collected by grafana-agent to be sampled at `5m` vs `1m` for all other integrations. The differing sample frequency futher causes issues when viewing these metrics in Grafana since with a `5m` scrape interval we couldn't use `$rate_interval` and other built-in Grafana variables with our `rate()` funcs. 

The decoupled scraping solves the io timeouts while also allowing us to keep the sample frequency at `1m`.

docs/sources/flow/reference/components/prometheus.exporter.cloudwatch.md
```diff
@@ -323,6 +326,21 @@ is exported to CloudWatch.
...
+of synchronously when the cloudwatch component is scraped.
+
+The decoupled scraping feature is useful to avoid against abuse of API requests that can caused extra billing in an AWS
+account. It is also useful to avoid scrape timeouts when running in very large AWS accounts where the number of
+cloudwatch metrics being gathered can not completed before the component scrape timeout.
+
+| Name              | Type     | Description                                                             | Default | Required |
+|-------------------|----------|-------------------------------------------------------------------------|---------|----------|
+| `enabled`         | `bool`   | Controls whether the decoupled scraping featured is enabled             | false   | no       |
+| `scrape_interval` | `string` | Controls how frequently to asynchronously gather new cloudwatch metrics | 5m      | no       |
```
clayton-cornell, line 342:
```suggestion
| `scrape_interval` | `string` | Controls how frequently to asynchronously gather new CloudWatch metrics | 5m      | no       |
```

docs/sources/static/configuration/integrations/cloudwatch-exporter-config.md
```diff
@@ -132,6 +132,16 @@ Configuration reference:
...
   [fips_disabled: <boolean> | default = false]
 
+
+  # Instead of retrieving metrics on request, decouple scraping retrieves the
+  # metrics on a schedule and returns the cached metrics.
+  decoupled_scraping:
+    # Enable decoupled scraping.
+    [enabled: <boolean> | default = false ]
+
+    # How often to scrape for AWS metrics.
```
clayton-cornell, line 142: Which is it? AWS metrics? or CloudWatch? The rest of the docs refer to it as CloudWatch metrics.

clayton-cornell (commented):
Some comments and questions for docs.

thepalbi:
> > Documentation added checkbox it's a reminder in the case that PR needs to included changes in the docs. In this case, it's applicable as you added a new field, which would be needed to be documented.
> > I cannot add this since its a fork but can you add this to the doc in cloudwatch-exporter-config?
> 
> Thanks both. I missed the "docs" pkg when looking at the root dir structure of this repo. Added docs to both static and flow integrations.
> 
> Also updated the flow integration to also have the `decoupled scraping` feature.
> 
> > Can I ask what's the driver for preferring the scrape to be asynchronous (like yace does natively) vs. during the /metrics call?
> 
> The main driver is to fix `write i/o timeout` errors we are seeing in some of our largest AWS cloud accounts. Today we get errors such as:
> 
> ```
> GET /integrations/cloudwatch_exporter/metrics 53.47008076s, error: write tcp 127.0.0.1:5080->127.0.0.1:55386: i/o timeout
> ```
> 
> When tried a few idea to work around the issue such as bumping the `scrape_interval`/`scrape_timeout` for the component integration to `5m`. However this causes the cloudwatch metrics collected by grafana-agent to be sampled at `5m` vs `1m` for all other integrations. The differing sample frequency futher causes issues when viewing these metrics in Grafana since with a `5m` scrape interval we couldn't use `$rate_interval` and other built-in Grafana variables with our `rate()` funcs.
> 
> The decoupled scraping solves the io timeouts while also allowing us to keep the sample frequency at `1m`.

The setup you are working with, is it grafana agent configured with the cloudwatch integration, and a destination prometheus to write on; or are you using the integration, and making prometheus scrape the `/integrations/cloudwatch_exporter/metrics` endpoint?

As shown in the docs, you can configure the `scrape_interval` and `scrape_timeout` just for the CloudWatch integrations. In that case, have you tried keeping the `1m` global scrape interval, and just bumping the `scrape_timeout` on the CloudWatch integration to sth that allows the scrape to run? 

For the record, we've seen internally in YACE (which is basically what the agent uses underneath) that for big big jobs (like you mention big AWS accounts) some scrape times are in the order of 1/2 mins.

docs/sources/flow/reference/components/prometheus.exporter.cloudwatch.md
```diff
@@ -323,6 +326,21 @@ is exported to CloudWatch.
...
 
+## decoupled scraping block
+
+The decoupled scraping block configures an optional feature which scrapes cloudwatch metrics in the background at a
+fixed interval. When enabled, cloudwatch metrics are gathered asynchronously at the decoupled scraped interval instead
+of synchronously when the cloudwatch component is scraped.
+
+The decoupled scraping feature is useful to avoid against abuse of API requests that can caused extra billing in an AWS
+account. It is also useful to avoid scrape timeouts when running in very large AWS accounts where the number of
+cloudwatch metrics being gathered can not completed before the component scrape timeout.
```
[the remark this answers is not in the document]
dtrejod: It's unclear what the exact number of AWS resources causes the scrape timeout to be hit. It's also not just tied to the size of the AWS account, but also what the user has configured to be scraped via this cloudwatch_exporter integration. 

I'm inclined to leave it purposefully vague by just saying "high"?

docs/sources/static/configuration/integrations/cloudwatch-exporter-config.md
dtrejod: Sorry this was written before the flow docs below. I think CloudWatch is more consistent, so I'll update this.

thepalbi:
Not against this change, but just want to make sure if you tried this other suggestion. Have you tried:
- increasing the `scrape_timeout` on the CW integration, to something larger that `1m`, that it allows your big scrapes to complete
- leaving the `scrape_interval` at `1m` for every integration

In that case you'll keep the `1m` scrapes everywhere, but allowing more time for the big ones, so that they are able to finish. This won't give you the caching mechanism async does, which is useful if you are scraping for example ec2 metrics that are produced every 5m, but still need the 1m scrape interval to keep everything aligned, but might work.

mattdurham (approved):
Thank you for the PR, LGTM after the feedback

dtrejod:
> increasing the scrape_timeout on the CW integration, to something larger that 1m, that it allows your big scrapes to complete leaving the scrape_interval at 1m for every integration

Yes we tried this and still saw consistent timeouts. We weren't able to diagnose the exact reason why these cloudwatch timeouts would happen even with very large scrape_timeout config. The below full structured log line shows you some more information as to the weirdness we saw. You can see in the headers that clearly show our scrape-timoeut was set to 2min yet we still timed out at 1m.. 

```
{"caller":"gokit.go:79","level":"warn","msg":"GET /integrations/cloudwatch_exporter/metrics 53.47008076s, error: write tcp 127.0.0.1:5080->127.0.0.1:55386: i/o timeout ws: false; Accept: application/openmetrics-text;version=1.0.0,application/openmetrics-text;version=0.0.1;q=0.75,text/plain;version=0.0.4;q=0.5,*/*;q=0.1; Accept-Encoding: gzip; User-Agent: GrafanaAgent/v0.35.1; X-Prometheus-Scrape-Timeout-Seconds: 120; ","ts":"2023-08-17T15:25:07.33575929Z"}
```

Overall this decoupled-scraping feature is definitely a solution that is working for us and one we'll start using right away. 

Thanks for the quick reviews!

thepalbi:
Thanks to you for the work @dtrejod !

PR #4920, opened 2023-08-24
Fix label normalisation in otelcol.exporter.loki

### LINKED ISSUE #3928: Document and verify Otlp -> Loki label translation
### Request

The Collector's Loki exporter documents its translation rules [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/lokiexporter) and [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/translator/prometheus/README.md#Labels). At the moment [otelcol.exporter.loki](https://grafana.com/docs/agent/latest/flow/reference/components/otelcol.exporter.loki/) does not have such a documentation. We need to add it.

In addition, we need to:
* Document the importance of hints such `loki.attribute.labels`
* Unit test the translation to make sure it works. I'm not 100% sure it works at the moment, because when I added a unit test just now it didn't pass in the way I thought it should do.

### Use case

The expected behaviour of the component will be more clear to the users.

ptodev:
#### PR Description

A few months ago we upgraded the version of Otel Collector which the Agent uses. This introduced a bug where `otelcol.exporter.loki` didn't normalize labels properly anymore.

Fixes #3928

#### Notes to the Reviewer

* I think there is a metric which should be documented, but I don't have time to make sure we expose it. I'll document it later for all otelcol components via #4788
* After we upgrade to a version of collector which contains [a fix for LogToLokiEntry](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/26093), we can remove the code on our side which applies the labels. We should retain the tests though.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


thampiotr (approved):
LGTM

PR #4927, opened 2023-08-24
Add backoff config to loki source file

### LINKED ISSUE #3606: Flow: support configurable backoff for file polling
#3560 bumped our dependency on Promtail to include grafana/loki#8874, but this was only added into static mode. 

`loki.source.file` should be updated to support being configured with explicit file backoff settings: 

```river 
loki.source.file "example" {
  targets    = TARGET_LIST
  forward_to = FORWARD_LIST  
  
  backoff {
    min_backoff = "250ms" 
    max_backoff = "5s" 
  }
}
```

The default for min and max backoff should be 250ms to match the existing behavior.

wildum:
#### PR Description

Added a backoff block in loki.source.file to configure how often to poll files from disk for changes.

Fixes #3606 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


thampiotr:
Doesn't this need the converters update too?

wildum:
>

wildum:
> Doesn't this need the converters update too?

Thanks for pointing it out, in https://github.com/grafana/agent/issues/3606 it says that it is already handled is static mode and it's only missing in flow mode but the code changed in between and it appears indeed that it required the converters to be updated

docs/sources/flow/reference/components/loki.source.file.md
```diff
@@ -72,6 +74,20 @@ Currently supported compression formats are:
 The component can only support one compression format at a time, in order to
 handle multiple formats, you will need to create multiple components.
 
+### backoff block
```
thampiotr, line 77: Why call it backoff?

Is there a reason to deviate from promtails naming?

image

wildum: I just followed the naming suggested by @rfratto in the issue. What was the initial reason to change the naming Robert?

clayton-cornell: Is `backoff` a predictable or understandable block name in context? As an outsider to the nuances of why the names are chosen, `file_watch` is more relatable to the function than `backoff`.

rfratto: Good question. I'm so used to seeing "backoff settings" w/r/t scraping and sending data that I don't know if I have a bias here. 

Either way, I'm fine with either of the two names, as long as we don't add the redundant `_config` in the name :)

thampiotr: I associate backoff with backing off in case of errors. I like the promtails naming more. And we can drop the _config suffix of course, I was talking about a high level naming here of concepts.

docs/sources/flow/reference/components/loki.source.file.md
```diff
@@ -72,6 +74,20 @@ Currently supported compression formats are:
...
+The following arguments are supported:
+
+ Name                 | Type       | Description                               | Default | Required 
+----------------------|------------|-------------------------------------------|---------|----------
+ `min_poll_frequency` | `duration` | Minimum frequency to poll for files.      |  250ms  | no      
+ `max_poll_frequency` | `duration` | Maximum frequency to poll for files.      |  250ms  | no       
+
+Any time file changes are detected, the poll frequency gets reset to `min_poll_frequency`.
+
+Any time no files changes are detected, the poll frequency doubles in value to the maximum duration specified by `max_poll_frequency`.
```
clayton-cornell, line 89:
```suggestion
If no file changes are detected, the poll frequency doubles until a file change is detected or the poll frequency reaches the `max_poll_frequency`.

If file changes are detected, the poll frequency is reset to `min_poll_frequency`.
```

clayton-cornell (commented):
A suggestion for how the poll freq is documented.

component/loki/source/file/file_test.go
```diff
@@ -76,6 +76,63 @@ func Test(t *testing.T) {
...
+			MaxPollFrequency: time.Millisecond * 500,
+		},
+	}
+
+	go func() {
+		err := ctrl.Run(ctx, args)
+		require.NoError(t, err)
+	}()
+
+	ctrl.WaitRunning(time.Minute)
```
thampiotr, line 110: Ignored return error value

component/loki/source/file/file_test.go
```diff
@@ -76,6 +76,63 @@ func Test(t *testing.T) {
...
+
+	// Sleep for 600ms to miss the first poll, the next poll should be 2*MinPollFrequency later.
+	time.Sleep(time.Millisecond * 600)
+
+	_, err = f.Write([]byte("writing some text\n"))
+	require.NoError(t, err)
+
+	select {
+	case logEntry := <-ch1.Chan():
+		require.Greater(t, time.Since(timeBeforeWriting), 1*time.Second)
```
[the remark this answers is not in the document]
wildum: yeah if my logic is correct, by setting the `MinPollFrequency` to 500ms, we will miss the first tick by sleeping for 600ms. That means that the next tick will be 1 second later which in total is 1.5 second after the write.
If the value of `MinPollFrequency` was not set correctly, it would be at 250ms. That means that the first tick would be missed by sleeping for 600ms, so the next tick would be 500ms later which in total is 750ms after the write. 750ms < 1 second, so the test would fail

CHANGELOG.md
```diff
@@ -62,6 +62,9 @@ v0.36.0 (2023-08-30)
 
 - Add a `labels` argument in `loki.source.windowsevent` to associate additional labels with incoming logs. (@wildum)
 
+- Add a `file_watch` block in `loki.source.file` to configure how often to poll files from disk for changes via `min_poll_frequency` and `max_poll_frequency`.
+  In static mode it can be configured in the global `file_watch_config` via `min_poll_frequency` and `max_poll_frequency`.  (@wildum)
+
```
thampiotr, line 67: Note that this must be moved to unreleased section now.

component/loki/source/file/file_test.go
```diff
@@ -76,6 +76,63 @@ func Test(t *testing.T) {
...
+		require.Equal(t, "writing some text", logEntry.Line)
+	case <-time.After(5 * time.Second):
+		require.FailNow(t, "failed waiting for log line")
+	}
+
+	// Shut down the component.
+	cancel()
+
+	// Wait to make sure that all go routines stopped.
+	time.Sleep(time.Millisecond * 100)
```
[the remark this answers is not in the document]
thampiotr: It seems like you're hitting this line: https://github.com/grafana/tail/blob/master/watch/polling.go#L107 

From what I can tell, the goroutine will exit after at most `MaxPollFrequency`, so can you use that timing instead? A random `100ms` is a bit of a smell and is often quite brittle and can become flaky, because there's no guarantee that it won't take `101ms`. 

We can also make the test a bit faster, e.g. divide all the timings by 2. Helps when trying to test for flakiness using `go test -count 10 github.com/grafana/agent/component/loki/source/file -run TestFileWatch`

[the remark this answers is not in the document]
wildum: the problem is that if I would divide all the timings by 2, then I would set 250ms which is the actual default. The test needs to check that the new values are set correctly so that would not test correctly

thampiotr (approved):

PR #4938, opened 2023-08-25
Clustering: make advertise interfaces configurable

### LINKED ISSUE #4910: clustering: make advertise interfaces configurable
Allowing users to define the network interfaces through which clustering will use to deduce a suitable advertise address will mitigate issues such as #4902

wildum:
#### PR Description

Make advertise interfaces configurable for clustering.
Before windows users had to explicitly set the advertise address but now they can instead configure the network interface available on their machine.

Fixes #4910 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added


docs/sources/flow/reference/cli/run.md
```diff
@@ -90,11 +91,12 @@ must be configured to accept connections on `--server.http.listen-addr` and the
...
-tries to infer a suitable one from the `eth0` and `en0` local network
-interfaces. If the advertised address cannot be determined, the agent will
-fail to start. Windows users must explicitly pass a value for
-`--cluster.advertise-address`, since Windows does not have interfaces name
-`eth0` or `en0`.
+tries to infer a suitable one from `--cluster.advertise-interfaces`.
+If `--cluster.advertise-interfaces` is not explicitly set, the agent will
+infer one from the `eth0` and `en0` local network interfaces.
+If the advertised address cannot be determined, the agent will fail to start.
+Since Windows does not have interfaces name `eth0` or `en0`, Windows users must explicitly pass
```
clayton-cornell, line 98:
```suggestion
Since Windows does not use the interface names `eth0` or `en0`, Windows users must explicitly pass
```

docs/sources/flow/reference/cli/run.md
```diff
@@ -90,11 +91,12 @@ must be configured to accept connections on `--server.http.listen-addr` and the
...
 If the `--cluster.advertise-address` flag is not explicitly set, the agent
-tries to infer a suitable one from the `eth0` and `en0` local network
-interfaces. If the advertised address cannot be determined, the agent will
-fail to start. Windows users must explicitly pass a value for
-`--cluster.advertise-address`, since Windows does not have interfaces name
-`eth0` or `en0`.
+tries to infer a suitable one from `--cluster.advertise-interfaces`.
+If `--cluster.advertise-interfaces` is not explicitly set, the agent will
+infer one from the `eth0` and `en0` local network interfaces.
+If the advertised address cannot be determined, the agent will fail to start.
```
clayton-cornell, line 97:
```suggestion
The agent will fail to start if it can't determine the advertised address.
```

We could rework this sentence to make it active instead of passive.

clayton-cornell (commented):
Some small suggestions

docs/sources/flow/reference/cli/run.md
```diff
@@ -50,6 +50,7 @@ The following flags are supported:
 * `--cluster.discover-peers`: List of key-value tuples for discovering peers (default `""`). Mutually exclusive with `--cluster.join-addresses`.
 * `--cluster.rejoin-interval`: How often to rejoin the list of peers (default `"60s"`).
 * `--cluster.advertise-address`: Address to advertise to other cluster nodes (default `""`).
+* `--cluster.advertise-interfaces`: List of interfaces used to infer an address to advertise. The first one available in the list will be selected (default `["eth0", "en0"]`).
```
rfratto, line 53: IMO the default shown here should be specified as the user would specify it (either comma-delimited or with multiple instances of the flag):

```suggestion
* `--cluster.advertise-interfaces`: List of interfaces used to infer an address to advertise. The first one available in the list will be selected (default `"eth0,en0"`).
```

cmd/internal/flowmode/cmd_run.go
```diff
@@ -102,6 +102,8 @@ depending on the nature of the reload error.
 		StringVar(&r.clusterJoinAddr, "cluster.join-addresses", r.clusterJoinAddr, "Comma-separated list of addresses to join the cluster at")
 	cmd.Flags().
 		StringVar(&r.clusterDiscoverPeers, "cluster.discover-peers", r.clusterDiscoverPeers, "List of key-value tuples for discovering peers")
+	cmd.Flags().
+		StringSliceVar(&r.clusterAdvInterfaces, "cluster.advertise-interfaces", r.clusterAdvInterfaces, "List of interfaces used to infer an address to advertise")
```
tpaschalis, line 106: nit: Can we set the default right on the `flowRun` struct on line 42 instead of implicitly setting it on the code later on?

This way the default value will still be available in case we use it elsewhere and can just call `opts.AdvertiseInterfaces` directly as an argument to FirstAddress.

tpaschalis (approved):
LGTM, thanks!

PR #4966, opened 2023-08-28
Fix elasticsearch flow component usage example

marctc:
#### PR Description

`address` field for Elasticsearch component needs to have `http` as prefix, otherwise
doesn't work. This PR fixes the usage  example in the docs of the flow component.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [x] Documentation added
- [ ] Tests updated


clayton-cornell (approved):
Looks good to me.

PR #4967, opened 2023-08-28
Re-enable stale bot (without closing issues or PRs)

marctc:
#### PR Description

This PR re-enables stale bot but without closing issues or PR marked
as stale. This would help us to identify if we have to follow up on issues or
PR and help contributors to follow up on opened issues or PR.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -14,7 +14,7 @@ jobs:
         with:
           close-issue-reason: not_planned
           days-before-stale: 30
-          days-before-close: -1
+          days-before-close: -1 # never close automatically
           stale-issue-message: >
             This issue has been automatically marked as stale because it has
             not had any activity in the past 30 days. The issue will not be closed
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,40 @@
...
+            Thank you for your contributions!
+          stale-pr-message: >
+            This PR has been automatically marked as stale because it has not
+            had any activity in the past 30 days.
+
+            If the PR could not move forward because it has not enough reviewers, please ping
+            the maintainers in order to get more reviews. If you do not have enough
+            time to follow up on this PR, consider closing it and let us know.
+
+            Thank you for your contributions!
```
rfratto, line 36: I think these should be changed to be slightly more clear that no action is needed from the author of the issue/PR; something that indicates the label is being applied for tracking but their issue/PR will _not_ be automatically closed. 

In particular, I'm concerned that the calls to ping the maintainers has the same annoying effect on authors: they'll think they have to fight the stale bot to remove the label. I think if the messages make it clear that the label is applied for tracking purposes, then we might not need to include this call to action here, but we can still request that the author close their issue/Pr if it's no longer relevant.

rfratto (approved):
LGTM % some tweaks. Thank you!

PR #4988, opened 2023-08-29
Clustering: add cluster.name flag to prevent accidental merges of clusters

### LINKED ISSUE #4466: Clustering: support "cluster names" to prevent two clusters from accidentally merging
### Request

Add a new optional flag, `-cluster.name`, to help prevent two clusters from accidentally merging. When `-cluster.name` is set, nodes will be prevented from joining that cluster if they do not have the same value for `-cluster.name`. 

This change requires changes to grafana/ckit to implement, primarily to expose [memberlist.Config.Label](https://pkg.go.dev/github.com/hashicorp/memberlist#Config).

Note that this field is **not** meant for security protection, and is only meant as a preventative measure from merging two clusters accidentally. Malicious users could still cause two clusters to merge. 

If cluster security is needed, it should be implemented by encrypting cluster traffic over TLS and doing mTLS authentication. 

### Use case

It is currently possible for two distinct clusters to accidentally merge if a new peer attempts to join both clusters simultaneously.

wildum:
#### PR Description

Add a new flag --cluster.name to prevent nodes without this identifier from joining the cluster. This will prevent clusters from accidentally merging.

Fixes #4466 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added


wildum:
The error msg returned by memberlist is not very good when you try to join a cluster with the wrong cluster.name: 
`ts=2023-08-29T08:55:35.700588Z level=warn msg="failed to connect to peers; bootstrapping a new cluster" err="failed to join memberlist: 2 errors occurred:\n\t* Failed to join [::1]:12345: EOF\n\t* Failed to join 127.0.0.1:12345: EOF\n\n"`

wildum:
> > The error msg returned by memberlist is not very good when you try to join a cluster with the wrong cluster.name:
> 
> I'd like for us to list and evaluate our options for improving this; from doing nothing, to forking memberlist and options in between.

I discussed this issue with @rfratto during our 1:1. He suggested to document that a reason for an EOF error could be a mismatch in the cluster name config in both the ckit code and the agent doc. I would say that this is better than nothing.

We could modify the error msg on the agent side but the error is very generic and could be triggered on different scenarios.

Forking memberlist would put additional maintenance effort on the team. That sounds like a heavy price for a bad error message. If we go down the road on modifying the memberlist code we could make a PR on the hashicorp memberlist repo but the changes might not be obvious and the PR might not be accepted.

Any other options?

I would just vote with the doc solution for now.

service/cluster/cluster.go
```diff
@@ -98,6 +99,7 @@ func New(opts Options) (*Service, error) {
 		AdvertiseAddr: opts.AdvertiseAddress,
 		Log:           l,
 		Sharder:       shard.Ring(tokensPerNode),
+		Label:         opts.ClusterName,
```
tpaschalis, line 102: One more thing to point out: in case we use this label with the current [mixin alerts](https://github.com/grafana/agent/blob/e966a94511e0dab32137227bebd54b24c9c35787/operations/agent-flow-mixin/alerts/clustering.libsonnet#L9) and deploy two different Agent clusters on the same K8S cluster and namespace, it would always trigger `ClusterNotConverging`.

Is there a way to avoid this by passing ClusterName as an extra label on metrics? I'm not even sure how the query would work when a cluster has a label but another doesn't (eg. the behavior of `by` clauses for non-existing labels).

It's not think it's necessarily a blocker for now, but we should at least investigate and open an issue to track and fix before the next release.

wildum: can't we just ignore the clusterName the same way we ignore the state in the query?

tpaschalis (approved):
LGTM, looking to merge once CI is green again.

docs/sources/flow/reference/cli/run.md
```diff
@@ -135,9 +135,9 @@ The first node that is used to bootstrap a new cluster (also known as
 the "seed node") can either omit the flags that specify peers to join or can
 try to connect to itself.
 
-The `--cluster.name` can be used to prevent two clusters from accidentally merging.
-When `--cluster.name` is set, nodes will be prevented from joining that cluster if 
-they do not have the same value for `--cluster.name`.
+The `--cluster.name` flag can be used to prevent clusters from accidentally merging.
+When `--cluster.name` is provided, nodes will only join peers who share the same cluster name value.
+By default, the cluster name is empty, and any node that doesn't set the flag can join
```
clayton-cornell:
```suggestion
By default, the cluster name is empty, and any node that doesn't set the flag can join.
```

clayton-cornell (commented):
Missing a full stop

docs/sources/flow/reference/cli/run.md
wildum: yeah I caught it earlier, I believe that you had an outdated version of the code

docs/sources/flow/reference/cli/run.md
```diff
@@ -134,6 +135,12 @@ The first node that is used to bootstrap a new cluster (also known as
 the "seed node") can either omit the flags that specify peers to join or can
 try to connect to itself.
 
+The `--cluster.name` flag can be used to prevent clusters from accidentally merging.
+When `--cluster.name` is provided, nodes will only join peers who share the same cluster name value.
+By default, the cluster name is empty, and any node that doesn't set the flag can join.
+Attempting to join a cluster with a wrong `--cluster.name` will result in a "failed to join memberlist" error
+with a `EOF` reason.
```
clayton-cornell, line 142:
```suggestion
Attempting to join a cluster with a `--cluster.name` that doesn't match will result in a "failed to join memberlist" error
with an `EOF` reason.
```

That reason isn't very intuitive. Is it possible to make it more explicit - something like "Cluster name not matched"?

What does the actual error message look like?

wildum: The problem is that the check is done in memberlist (hashicorp repo) and that they don't return the error. The actual message is `ts=2023-08-29T08:08:59.457373Z level=warn msg="failed to connect to peers; bootstrapping a new cluster" err="failed to join memberlist: 2 errors occurred:\n\t* Failed to join [::1]:12345: EOF\n\t* Failed to join 127.0.0.1:12345: EOF\n\n"`.
So in the agent and in ckit we don't have the control over it, we just know that the node tried to join the memberlist and that something went wrong

clayton-cornell: Ah ok. Then is it necessary to mention EOF?  Can we just note `failed to join memberlist`?  There's other info that we don't note like the specific nodes...

wildum: you're right, it's probably better to just keep `failed to join memberlist`, I will update it

rfratto:
Merging since there are approvals from code and docs maintainers. Thanks William!

PR #4993, opened 2023-08-29
helm: enable clustering support for deployments and daemonsets

### LINKED ISSUE #4995: [Flow mode] support clustering for daemonset controller type
### Request

Please consider support for clustering also for `controller.type=daemonset` when implementing https://github.com/grafana/agent/issues/4598 as it would help with components that need to run:
- only one instance to avoid duplication of records https://github.com/grafana/agent/issues/4582
- determined number of instances to split the workload

With Helm installation it is explicitly disabled for `controller.type` other than `statefulset` here: https://github.com/grafana/agent/blob/8a5fe1484e8192f72104562c2585a26208c4c036/operations/helm/charts/grafana-agent/templates/controllers/statefulset.yaml#L2

### Use case

To avoid duplicate records when collecting kubernetes events via `loki.source.kubernetes_events`.

tpaschalis:
Since we've merged support for periodically rejoining the list of peers, this enables clustering to work with deployments, daemonsets as well as statefulsets with maxUnavailablePods > 1.

We've already identified this can be noisy in large environments, but this will be mitigated with #4464 or similar.
As a drive-by, we fix the changelog of a previous release.

Fixes #4995.

### DIFF
--- a/operations/helm/charts/grafana-agent/CHANGELOG.md
+++ b/operations/helm/charts/grafana-agent/CHANGELOG.md
@@ -8,8 +8,11 @@ changes that impact end-user behavior are listed; changes to documentation or
 internal API changes are not present.
 
 Unreleased
-
 ----------
+
+### Enhancements
+
+- Enable clustering for deployments and daemonsets. (@tpaschalis)
 
 0.22.0 (2023-08-30)
 -------------------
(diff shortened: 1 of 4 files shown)


rfratto (approved):
🎉

PR #4998, opened 2023-08-29
add more complete river identifier validation and sanitization

erikbaranowski:
#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/converter/internal/common/common.go
+++ b/converter/internal/common/common.go
@@ -4,9 +4,12 @@ import (
 	"bytes"
 	"fmt"
 	"strings"
+	"unicode"
 
 	"github.com/grafana/river/parser"
 	"github.com/grafana/river/printer"
+	"github.com/grafana/river/scanner"
+	"github.com/grafana/river/token"
 
 	"github.com/grafana/agent/component"
 	flow_relabel "github.com/grafana/agent/component/common/relabel"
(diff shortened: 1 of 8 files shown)


converter/internal/common/common.go
```diff
@@ -106,3 +109,41 @@ func PrettyPrint(in []byte) ([]byte, diag.Diagnostics) {
...
+		if !unicode.IsLetter(c) && !unicode.IsNumber(c) && c != '_' {
+			newValue += "_"
+			diags.Add(diag.SeverityLevelWarn, fmt.Sprintf("invalid character %q. replaced with \"+\" since identifiers must only contain letters, numbers, and underscores.", c))
+			continue
+		}
+
+		newValue += string(c)
+	}
+
+	return newValue, diags
```
rfratto, line 141: Rather than reimplementing identifier checking, I would instead suggest using River's scanner. See this code here: 

https://github.com/grafana/river/blob/93eb6c45de98ba6b39f5a33690ff82291533f642/parser/internal.go#L711-L715

erikbaranowski: this functionality will be implemented in the river project and used here once complete instead.

rfratto (approved):

PR #5001, opened 2023-08-30
Clustering: add max initial join peers flag

### LINKED ISSUE #4464: Clustering: don't try to join all peers on startup

wildum:
#### PR Description

On startup an agent with clustering enabled will try to join every peers from the discovered set. This action can be costly for clusters of significant sizes.

The flag max-initial-join-peers will allow the users to limit the number of peers an agent should join upon initialisation. The subset of peers is randomly selected amongst the discovered set. 

 Fixes #4464 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/service/cluster/cluster.go
+++ b/service/cluster/cluster.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
+	"math/rand"
 	"net"
 	"net/http"
 	"strings"
@@ -56,13 +57,18 @@ type Options struct {
 	// possible for other nodes to join the cluster.
 	EnableClustering bool
 
-	NodeName         string        // Name to use for this node in the cluster.
-	AdvertiseAddress string        // Address to advertise to other nodes in the cluster.
-	RejoinInterval   time.Duration // How frequently to rejoin the cluster to address split brain issues.
+	NodeName                string        // Name to use for this node in the cluster.
+	AdvertiseAddress        string        // Address to advertise to other nodes in the cluster.
+	RejoinInterval          time.Duration // How frequently to rejoin the cluster to address split brain issues.
+	ClusterMaxInitJoinPeers int           // Number of initial peers to join from the discovered set
 
 	// Function to discover peers to join. If this function is nil or returns an
 	// empty slice, no peers will be joined.
 	DiscoverPeers func() ([]string, error)
+}
+
+type RandShuffler interface {
+	Shuffle(n int, swap func(i, j int))
 }
 
 // Service is the cluster service.
@@ -73,6 +79,7 @@ type Service struct {
 
 	sharder shard.Sharder
 	node    *ckit.Node
+	randGen RandShuffler
 }
 
 var (
@@ -136,6 +143,7 @@ func New(opts Options) (*Service, error) {
 
 		sharder: ckitConfig.Sharder,
 		node:    node,
+		randGen: rand.New(rand.NewSource(time.Now().UnixNano())),
 	}, nil
 }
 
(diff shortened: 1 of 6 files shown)


service/cluster/cluster.go
```diff
@@ -290,7 +298,22 @@ func (s *Service) getPeers() ([]string, error) {
...
+	// Here we return the entire list because we can't take a subset.
+	if s.opts.ClusterMaxInitJoinPeers == 0 || len(peers) < s.opts.ClusterMaxInitJoinPeers {
+		return peers, nil
+	}
+
+	// We shuffle the list and return only a subset of the peers.
+	s.randGen.Shuffle(len(peers), func(i, j int) {
+		peers[i], peers[j] = peers[j], peers[i]
+	})
+	return peers[:s.opts.ClusterMaxInitJoinPeers], nil
```
tpaschalis, line 316: So this change would not only affect the _initial_ set of peers joined, but also the peers we try to join during a refresh (which is what we want).

Could we update the parameter name, docs and PR title to refer to that?

service/cluster/cluster.go
```diff
@@ -56,15 +57,20 @@ type Options struct {
...
 
 	// Function to discover peers to join. If this function is nil or returns an
 	// empty slice, no peers will be joined.
 	DiscoverPeers func() ([]string, error)
 }
 
+type RandShuffler interface {
+	Shuffle(n int, swap func(i, j int))
+}
+
```
tpaschalis, line 73: Is the reason we're adding this only for the tests?

I'm asking because, 'math/rand' already contains a [rand.Shuffle function](https://pkg.go.dev/math/rand#Shuffle) we can use directly with the default rand source? Then in the tests we can reliably seed it with the same number.

service/cluster/cluster.go
wildum: thanks for pointing it out, at first I made the change only for the startup but then changed the implementation and overlooked the refresh. Good that we also want it in the refresh then, I will update everything to refer to that

service/cluster/cluster.go
wildum: yes this was only for the test. You are right, we could make it work with the seeding instead of mocking it. Let's go with your suggestion because it's less code and I don't really see any cons

clayton-cornell (commented):
Docs are OK as-is.

service/cluster/cluster_test.go
```diff
@@ -0,0 +1,59 @@
...
+			s := &Service{
+				opts:    test.opts,
+				randGen: rand.New(rand.NewSource(1)), // Seeded random generator to have consistent results in tests.
+			}
+
+			peers, _ := s.getPeers()
+
+			if !reflect.DeepEqual(peers, test.expectedPeers) {
+				t.Errorf("Expected peers %v, got %v", test.expectedPeers, peers)
+			}
```
rfratto, line 56: tip: you can use [require.ElementsMatch](https://pkg.go.dev/github.com/stretchr/testify/require#ElementsMatch) to assert that two slices are equal regardless of sort order, which can be useful for tests like these where the output order isn't guaranteed.

rfratto (approved):
LGTM, thank you!

clayton-cornell (approved):
Docs are OK as-is

PR #5002, opened 2023-08-30
Fix changelog conflict

wildum:
#### PR Description

Fix a conflict which accidentally made its way on main.

#### PR Checklist

- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,25 +100,7 @@ v0.36.0-rc.3 (2023-08-28)
 - Update `memcached_exporter` to `v0.13.0`, which includes bugfixes, new metrics,
   and the option to connect with TLS. (@spartan0x117)
 
-<<<<<<< HEAD
 - `loki.write` now supports configuring retries on HTTP status code 429. (@wildum)
-=======
-- Flow: Added exemplar support for the `otelcol.exporter.prometheus`. (@wildum)
-
-- New Grafana Agent Flow components:
-
-  - `prometheus.exporter.gcp` - scrape GCP metrics. (@tburgessdev)
-  - `otelcol.processor.span` - accepts traces telemetry data from other `otelcol`
-  components and modifies the names and attributes of the spans. (@ptodev)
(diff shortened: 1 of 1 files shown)


tpaschalis:
cc @jdbaldry sorry for the second ping in a row 😅 

I saw [we recently](https://hub.docker.com/r/grafana/docs-base/tags) updated our docs-base image right? All of a sudden we're getting a lot of these warnings and a couple of errors. Where should we apply the fix?

```
# a LOT of these alerts
WARN  "/hugo/content/docs/agent/latest/flow/reference/components/remote.http.md:100:1": [docs/shared] shared directory page "docs/agent/monitor-infrastructure/shared" not found, is it missing an 'index.md' file?

# And these errors
ERROR [en] REF_NOT_FOUND: Ref "./configuration" from page "docs/grafana-cloud/monitor-infrastructure/integrations/agent/static/configuration/create-config-file.md": page not found
ERROR [en] REF_NOT_FOUND: Ref "./configuration" from page "docs/grafana-cloud/monitor-infrastructure/integrations/agent/static/configuration/flags.md": page not found
ERROR [en] REF_NOT_FOUND: Ref "./install-agent-mac-os" from page "docs/grafana-cloud/monitor-infrastructure/integrations/agent/static/set-up/install/install-agent-binary.md": page not found
```

jdbaldry:
No problem at all. Thanks for flagging this!

I'll take a look at the warnings and errors this morning and confirm what is happening and how to resolve it.

jdbaldry:
Ah, I've found the problem to be a directory that shouldn't have been included in the container image. Because the directory had no files, it was not tracked by git and it wasn't obvious to me that my workspace wasn't clean. 

The result was that your docs were mounted into the Grafana Cloud documentation as part of the Hugo pre-build virtual file system but since all the rest of the Grafana Cloud documentation was absent from the build, it caused these errors.

I've pushed an updated image tag for the revision and `latest` and your builds should now work again.

PR #5036, opened 2023-08-30
add support for discovery.triton flow component and converter

### LINKED ISSUE #2282: Flow: discovery.triton component
For feature parity, we should bring over the upstream Prometheus SD mechanism and introduce it as a Flow component.

erikbaranowski:

Closes #2282

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/component/discovery/triton/triton.go
+++ b/component/discovery/triton/triton.go
@@ -1,0 +1,79 @@
+package triton
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/grafana/agent/component"
+	"github.com/grafana/agent/component/common/config"
+	"github.com/grafana/agent/component/discovery"
+	"github.com/prometheus/common/model"
+	prom_discovery "github.com/prometheus/prometheus/discovery/triton"
+)
+
+func init() {
+	component.Register(component.Registration{
+		Name:    "discovery.triton",
+		Args:    Arguments{},
+		Exports: discovery.Exports{},
+
+		Build: func(opts component.Options, args component.Arguments) (component.Component, error) {
+			return New(opts, args.(Arguments))
+		},
+	})
+}
+
+type Arguments struct {
+	Account         string           `river:"account,attr"`
+	Role            string           `river:"role,attr,optional"`
+	DNSSuffix       string           `river:"dns_suffix,attr"`
+	Endpoint        string           `river:"endpoint,attr"`
+	Groups          []string         `river:"groups,attr,optional"`
+	Port            int              `river:"port,attr,optional"`
+	RefreshInterval time.Duration    `river:"refresh_interval,attr,optional"`
+	Version         int              `river:"version,attr,optional"`
+	TLSConfig       config.TLSConfig `river:"tls_config,block,optional"`
+}
+
+var DefaultArguments = Arguments{
+	Role:            "container",
+	Port:            9163,
+	RefreshInterval: 60 * time.Second,
+	Version:         1,
+}
+
+// SetToDefault implements river.Defaulter.
+func (args *Arguments) SetToDefault() {
+	*args = DefaultArguments
+}
+
+// Validate implements river.Validator.
+func (args *Arguments) Validate() error {
+	if args.RefreshInterval <= 0 {
+		return fmt.Errorf("refresh_interval must be greater than 0")
+	}
+
+	return nil
+}
+
+func (args *Arguments) Convert() *prom_discovery.SDConfig {
+	return &prom_discovery.SDConfig{
+		Account:         args.Account,
+		Role:            args.Role,
+		DNSSuffix:       args.DNSSuffix,
+		Endpoint:        args.Endpoint,
+		Groups:          args.Groups,
+		Port:            args.Port,
+		RefreshInterval: model.Duration(args.RefreshInterval),
+		TLSConfig:       *args.TLSConfig.Convert(),
+		Version:         args.Version,
+	}
+}
+
+// New returns a new instance of a discovery.triton component.
+func New(opts component.Options, args Arguments) (*discovery.Component, error) {
+	return discovery.New(opts, args, func(args component.Arguments) (discovery.Discoverer, error) {
+		newArgs := args.(Arguments)
+		return prom_discovery.New(opts.Logger, newArgs.Convert())
+	})
+}
--- a/converter/internal/prometheusconvert/triton.go
+++ b/converter/internal/prometheusconvert/triton.go
@@ -1,0 +1,41 @@
+package prometheusconvert
+
+import (
+	"time"
+
+	"github.com/grafana/agent/component/discovery"
+	"github.com/grafana/agent/component/discovery/triton"
+	"github.com/grafana/agent/converter/diag"
+	"github.com/grafana/agent/converter/internal/common"
+	prom_triton "github.com/prometheus/prometheus/discovery/triton"
+)
+
+func appendDiscoveryTriton(pb *prometheusBlocks, label string, sdConfig *prom_triton.SDConfig) discovery.Exports {
+	discoveryTritonArgs := ToDiscoveryTriton(sdConfig)
+	name := []string{"discovery", "triton"}
+	block := common.NewBlockWithOverride(name, label, discoveryTritonArgs)
+	pb.discoveryBlocks = append(pb.discoveryBlocks, newPrometheusBlock(block, name, label, "", ""))
+	return NewDiscoveryExports("discovery.triton." + label + ".targets")
+}
+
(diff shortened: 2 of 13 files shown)


component/discovery/triton/triton.go
```diff
@@ -0,0 +1,79 @@
...
+type Arguments struct {
+	Account         string           `river:"account,attr"`
+	Role            string           `river:"role,attr,optional"`
+	DNSSuffix       string           `river:"dns_suffix,attr"`
+	Endpoint        string           `river:"endpoint,attr"`
+	Groups          []string         `river:"groups,attr,optional"`
+	Port            int              `river:"port,attr,optional"`
+	RefreshInterval time.Duration    `river:"refresh_interval,attr,optional"`
+	Version         int              `river:"version,attr,optional"`
+	TLSConfig       config.TLSConfig `river:"tls_config,block,optional"`
```
captncraig, line 35: Most of the discovery has the full HttpClientConfig. Is there a reason this only has tls?

erikbaranowski: ya, that's what prometheus supports for this discovery. if I should still be using HttpClientConfig let me know but I'm not sure how it would work in this case.

https://prometheus.io/docs/prometheus/latest/configuration/configuration/#triton_sd_config

https://github.com/prometheus/prometheus/blob/c5e1498ae243bbf0b8363786de065bf98acd1edb/discovery/triton/triton.go#L68

docs/sources/flow/reference/components/discovery.triton.md
```diff
@@ -0,0 +1,125 @@
...
+## Arguments
+
+The following arguments are supported:
+
+Name               | Type           | Description                                                         | Default       | Required
+------------------ | -------------- | ------------------------------------------------------------------- | ------------- | --------
+`account`          | `string`       | The account to use for discovering new targets.                     |               | yes
+`role`             | `string`       | The type of targets to discover.                                    | `"container"` | no
+`dns_suffix`       | `string`       | The DNS suffix which should be applied to target.                   |               | yes
+`endpoint`         | `string`       | The Triton discovery endpoint (e.g. 'cmon.us-east-3b.triton.zone'). |               | yes
```
clayton-cornell, line 31:
```suggestion
`endpoint`         | `string`       | The Triton discovery endpoint (e.g. 'cmon.us-east-3b.triton.zone'). |               | yes
```

Should we include examples in the argument descriptions? This has come up before and we've generally opted for no examples here. If we do include the example, use `for example` instead of `e.g.`.

docs/sources/flow/reference/components/discovery.triton.md
```diff
@@ -0,0 +1,125 @@
...
+`groups`           | `list(string)` | A list of groups for which targets are retrieved.                   |               | no
```
clayton-cornell, line 32:
```suggestion
`groups`           | `list(string)` | A list of groups to retrieve targets from.                   |               | no
```

You can end a sentence with a preposition. This description is not wrong the way it was written but it's not conversational or natural. You can leave it or change it.. up to you.

docs/sources/flow/reference/components/discovery.triton.md
```diff
@@ -0,0 +1,125 @@
...
+`port`             | `int`          | The port to use for discovery and metrics scraping.                 | `9163`        | no
+`refresh_interval` | `duration`     | Interval at which to refresh the list of targets.                   | `60s`         | no
```
clayton-cornell, line 34:
```suggestion
`refresh_interval` | `duration`     | The refresh interval for the list of targets.                   | `60s`         | no
```

Could we simplify here? Or does this change the meaning from what you intended?

docs/sources/flow/reference/components/discovery.triton.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Exported fields
+
+The following fields are exported and can be referenced by other components:
+
+Name      | Type                | Description
+--------- | ------------------- | -----------
+`targets` | `list(map(string))` | The set of targets discovered from the Triton API.
+
+When `role` is set to `"container"` each target includes the following labels:
```
clayton-cornell, line 66:
```suggestion
When `role` is set to `"container"`, each target includes the following labels:
```

clayton-cornell (commented):
A few doc suggestions

docs/sources/flow/reference/components/discovery.triton.md
erikbaranowski: example removed

converter/internal/prometheusconvert/triton.go
```diff
@@ -0,0 +1,41 @@
...
+func appendDiscoveryTriton(pb *prometheusBlocks, label string, sdConfig *prom_triton.SDConfig) discovery.Exports {
+	discoveryTritonArgs := ToDiscoveryTriton(sdConfig)
+	name := []string{"discovery", "triton"}
+	block := common.NewBlockWithOverride(name, label, discoveryTritonArgs)
+	pb.discoveryBlocks = append(pb.discoveryBlocks, newPrometheusBlock(block, name, label, "", ""))
+	return NewDiscoveryExports("discovery.triton." + label + ".targets")
+}
+
+func validateDiscoveryTriton(sdConfig *prom_triton.SDConfig) diag.Diagnostics {
+	return nil
```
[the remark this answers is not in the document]
erikbaranowski: If using HTTPClientConfig we have a prebuilt validator where things don't align and the other thing I've seen is with digital ocean we had to do some additional specific validations.

https://github.com/grafana/agent/blob/597bb6b01207b39b7e51ba315e6116bbab289f3a/converter/internal/prometheusconvert/digitalocean.go#L25

We don't technically need to even create and wire a validate function in the case of this component, but I have been creating them for consistency. We have discussed coming up with a common pattern for all converters so the pattern you see here may be improved down the road.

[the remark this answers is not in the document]
erikbaranowski: To summarize, we need to validate if there is something supported in prometheus that is not supported in flow.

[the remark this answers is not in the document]
tpaschalis: Great, thanks!

thampiotr:
I think there is an unsupported `if` statement somewhere with `triton_sd_configs is not supported`. @erikbaranowski

docs/sources/flow/reference/components/discovery.triton.md
```diff
@@ -0,0 +1,125 @@
...
+
+Hierarchy | Block | Description | Required
+--------- | ----- | ----------- | --------
+tls_config | [tls_config][] | TLS configuration for requests to the Triton API. | no
+
+[tls_config]: #tls_config-block
+
+### tls_config block
+
+{{< docs/shared lookup="flow/reference/components/tls-config-block.md" source="agent" >}}
```
clayton-cornell, line 56:
```suggestion
{{< docs/shared lookup="flow/reference/components/tls-config-block.md" source="agent" version="<AGENT VERSION>" >}}
```

clayton-cornell (commented):
Add missing share syntax

component/discovery/triton/triton.go
```diff
@@ -0,0 +1,79 @@
...
+		Role:            args.Role,
+		DNSSuffix:       args.DNSSuffix,
+		Endpoint:        args.Endpoint,
+		Groups:          args.Groups,
+		Port:            args.Port,
+		RefreshInterval: model.Duration(args.RefreshInterval),
+		TLSConfig:       *args.TLSConfig.Convert(),
+		Version:         args.Version,
+	}
+}
```
wildum, line 71: I would suggest to always cover the Convert function with a test to make sure that the args are correctly passed to the prometheus discovery. Last time I assigned the TLS config without converting it and it compiled and ran without error. I only realised that the conversion was missing when I ran a test

thampiotr:
After https://github.com/grafana/agent/pull/5069 was merged, I've rebased this PR and adapted to work with the new approach.

converter/internal/promtailconvert/testdata/unsupported.diags
```diff
@@ -14,5 +14,4 @@
 (Error) unsupported service discovery marathon was provided
 (Error) unsupported service discovery nerve was provided
 (Error) unsupported service discovery openstack was provided
-(Error) unsupported service discovery serverset was provided
-(Error) unsupported service discovery triton was provided
\ No newline at end of file
+(Error) unsupported service discovery serverset was provided
```
tpaschalis, line 17: Off-topic question. One of the discovery mechanisms I'm doing (`discovery.kuma`) does not have an entry here. Is it just an omission?

thampiotr (approved):

PR #5039, opened 2023-08-31
Fix changelog: enhancement added under the wrong version

wildum:
#### PR Description

The change was added under v0.36 but the code itself was not part of v0.36.
This PR moves it to the right part in changelog.md

#### PR Checklist

- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@ Main (unreleased)
 - Flow: allow the HTTP server to be configured with TLS in the config file
   using the new `http` config block. (@rfratto)
 
+### Enhancements
+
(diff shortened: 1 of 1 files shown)


tpaschalis (approved):

tpaschalis:
@wildum If this change is present on the `release-v0.36` branch, you can use the `backport *` labels to automatically open a PR there as well.

wildum:
> @wildum If this change is present on the `release-v0.36` branch, you can use the `backport *` labels to automatically open a PR there as well.

Thanks for the tip :) In this case the problem is only on main I believe, unless Im mistaken it's not on https://github.com/grafana/agent/blob/release-v0.36/CHANGELOG.md

PR #5045, opened 2023-08-31
Enable Collector Otel metrics in Flow

### LINKED ISSUE #3923: Flow: Display metrics from Collector components in the Agent's metrics endpoint

ptodev:
#### PR Description

The Agent already had some OTel metrics such as `rpc_server_duration_milliseconds`. However, apparently a special feature gate needs to be enabled so that Agent can receive most other metrics from Otel Collector components.

Fixes #3923

#### Notes to the Reviewer

Some metrics have empty labels like `receiver=""`, `exporter=""`, and `processor=""`, e.g.:
```
exporter_send_failed_spans_ratio_total{component_id="otelcol.exporter.otlp.grafana_cloud_tempo",exporter="",otel_scope_name="go.opentelemetry.io/collector/obsreport/exporter",otel_scope_version=""} 0
processor_batch_timeout_trigger_send_ratio_total{component_id="otelcol.processor.batch.default",otel_scope_name="go.opentelemetry.io/collector/processor/batchprocessor",otel_scope_version="",processor=""} 99
```

I think this is because we don't set the component name in the Otel component factory methods. We have a few options:
* Let these labels be empty. This may be confusing because they will have the same empty value for multiple components. But it's probably not a big deal for now.
* Set the component name via the factory function. It should be a unique name, so that no two components have the same one. The downside of this is that we already have a `component_id` label.
* Create a view which drops these labels.

For now I think we can merge this PR as it is, but I could create a separate issue to resolve the situation with empty labels?

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [ ] Tests updated


clayton-cornell (commented):
Docs are OK. No changes.

cmd/internal/flowmode/cmd_run.go
```diff
@@ -154,6 +168,17 @@ func (fr *flowRun) Run(configFile string) error {
...
+	// Enable the "telemetry.useOtelForInternalMetrics" Collector feature gate.
+	// Currently, Collector components uses OpenCensus metrics by default.
+	// Those metrics cannot be integrated with Agent Flow,
+	// so we need to always use OpenTelemetry metrics.
+	err = enableOtelFeatureGates(
+		"telemetry.useOtelForInternalMetrics",
+	)
+	if err != nil {
+		return err
+	}
```
rfratto, line 180: Have you seen https://github.com/grafana/agent/blob/main/component/otelcol/internal/featuregate/install.go? Why do we need this again here?

rfratto (changes requested):
Some questions about why there's duplicated code for the featuregate.

cmd/internal/flowmode/cmd_run.go
ptodev: Thanks for letting me know! I had no idea about this. I suppose that package didn't work for me since it relied on the `AGENT_MODE` env var, which the "agent-flow" binary doesn't use. I deleted the package since I think my code would work for both "agent" and "agent-flow" binaries. A bonus about deleting it is that we also don't rely on `init()` as much.

I also pushed a minor change to move my function for setting the feature gate to the util package, so that there is less repetition.

cmd/internal/flowmode/cmd_run.go
```diff
@@ -157,6 +158,17 @@ func (fr *flowRun) Run(configFile string) error {
...
+	// Enable the "telemetry.useOtelForInternalMetrics" Collector feature gate.
+	// Currently, Collector components uses OpenCensus metrics by default.
+	// Those metrics cannot be integrated with Agent Flow,
+	// so we need to always use OpenTelemetry metrics.
+	err = util.EnableOtelFeatureGates(
+		"telemetry.useOtelForInternalMetrics",
+	)
+	if err != nil {
+		return err
+	}
```
rfratto, line 170: nit: I don't particularly love having component-specific logic in cmd/internal/flowmode/cmd_run.go, which is otherwise agnostic to all the different component types. It feels harder to discover where things are happening when it's spread out beyond its package.

cmd/internal/flowmode/cmd_run.go
```diff
@@ -157,6 +158,17 @@ func (fr *flowRun) Run(configFile string) error {
 		return fmt.Errorf("building tracer: %w", err)
 	}
 
+	// Enable the "telemetry.useOtelForInternalMetrics" Collector feature gate.
+	// Currently, Collector components uses OpenCensus metrics by default.
+	// Those metrics cannot be integrated with Agent Flow,
+	// so we need to always use OpenTelemetry metrics.
```
rfratto, line 164: nit: this removes the TODO comment which mentions that the feature gate will eventually need to be removed once it's removed from OpenTelemetry.

component/otelcol/internal/featuregate/install_test.go
```diff
@@ -1,13 +0,0 @@
...
-	"testing"
-
-	"github.com/stretchr/testify/require"
-	"go.opentelemetry.io/collector/featuregate"
-)
-
-func Test_enableFeatureGates(t *testing.T) {
-	err := enableFeatureGates(featuregate.GlobalRegistry())
-	require.NoError(t, err, "enableFeatureGates should not have a failed. Did a feature gate get removed from upstream?")
-}
```
rfratto, line 13: nit: it's unfortunate to remove this test. We should bring it back.

ptodev: Ok sure, I added it in a slightly different way so that it tests more. I hope you're ok with it?

cmd/internal/flowmode/cmd_run.go
ptodev: I changed it back to an `init()` function for flow. I'm personally not a big fan of these init functions, but I guess it's ok here :)

ptodev: I think my solution with "init()" won't work, because the flow "init()" will also get called during static mode (because the "agent" binary also imports the otelcol files). Let me know what you think about it. If I can't think of other ideas I'll probably switch it to my non-init version.

mattdurham: Services might be a good spot? Something like an Otel service?

rfratto: I agree that a service seems like a better place for it 👍

ptodev: Thanks for the feedback! I updated the PR to use a service, PTAL.

rfratto (approved):
LGTM, thank you!

PR #5054, opened 2023-08-31
add tracking and comparison of latest healthy content before loading a module

### LINKED ISSUE #5028: module.git should not reload flow components every time it polls
### What's wrong?

module loaders commonly poll but they should not reload flow components unless a change is detected to the content of the module loader.

### Steps to reproduce

use module.git and observe that components are restarted without modification each time it polls.

https://github.com/grafana/agent/blob/2c12db25affbe919e1037005dbb938459b79fb0d/component/module/git/git.go#L231

### System information

_No response_

### Software version

_No response_

### Configuration

_No response_

### Logs

_No response_

erikbaranowski:

Fixes #5028 

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,12 @@ Main (unreleased)
 - Read contextual attributes from Faro measurements (@codecapitano)
 - Rename Grafana Agent service in windows app and features to not include the description
 
+### Bugfixes
+
+- Restart managed components of a module loader only on if module content
+  changes or the last load failed. This was specifically impacting `module.git`
+  each time it pulls. (@erikbaranowski)
+
 v0.36.0 (2023-08-30)
 --------------------
 
--- a/component/module/module.go
+++ b/component/module/module.go
@@ -3,6 +3,7 @@ package module
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"sync"
 	"time"
 
@@ -17,6 +18,7 @@ type ModuleComponent struct {
 	mut           sync.RWMutex
 	health        component.Health
 	latestContent string
+	latestArgs    map[string]any
 }
 
 // Exports holds values which are exported from the run module.
@@ -42,7 +44,7 @@ func NewModuleComponent(o component.Options) (*ModuleComponent, error) {
 // can rely on either or both. If the content is the same as the last time it was
 // successfully loaded, it will not be reloaded.
 func (c *ModuleComponent) LoadFlowContent(args map[string]any, contentValue string) error {
-	if contentValue == c.getLatestContent() {
+	if reflect.DeepEqual(args, c.getLatestArgs()) && contentValue == c.getLatestContent() {
 		return nil
 	}
 
@@ -57,6 +59,7 @@ func (c *ModuleComponent) LoadFlowContent(args map[string]any, contentValue stri
 		return err
 	}
 
+	c.setLatestArgs(args)
 	c.setLatestContent(contentValue)
 	c.setHealth(component.Health{
 		Health:     component.HealthTypeHealthy,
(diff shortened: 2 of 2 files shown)


component/module/module.go
```diff
@@ -38,10 +39,16 @@ func NewModuleComponent(o component.Options) (*ModuleComponent, error) {
...
+// can rely on either or both. If the content is the same as the last time it was
+// successfully loaded, it will not be reloaded.
 func (c *ModuleComponent) LoadFlowContent(args map[string]any, contentValue string) error {
+	if contentValue == c.getLatestContent() {
+		return nil
+	}
+
 	err := c.mod.LoadConfig([]byte(contentValue), args)
 	if err != nil {
+		c.setLatestContent("")
```
erikbaranowski, line 51: reset the latestContent so if content fails to load we are able to retry it

CHANGELOG.md
```diff
@@ -30,6 +30,12 @@ Main (unreleased)
 - Read contextual attributes from Faro measurements (@codecapitano)
 - Rename Grafana Agent service in windows app and features to not include the description
 
+### Bugfixes
+
+- Restart managed components of a module loader only on if module content
+  changes or the last load failed. This was specifically impacting `module.git`
```
rfratto, line 36: Well, it impacts module.http too.

component/module/module.go
rfratto: Maybe add a comment for this?

I'd personally find this surprising, since based on Flow's guarantees, the last config continues to be used. 

Since you only call `setLatestContent` when there's not an error, I'm not sure this is necessary; you can still try with content that hasn't been used before.

CHANGELOG.md
erikbaranowski: Check my understanding: I don't think it is the case today that this happens for module.http or module.file since LoadFlowContent is called inside Update (means the module loader definition changed) and inside the OnStateChange of the embedded component which will only fire if the content is in fact updated.

https://github.com/grafana/agent/blob/33b7b61ed2421f3168cd9349a499201409e54284/component/module/http/http.go#L93

rfratto: Nope, you're right, my bad.

component/module/module.go
erikbaranowski: o ya actually that makes sense. Let me not set it in this case so the latest config is still correctly stored here.

component/module/module.go
```diff
@@ -38,8 +39,13 @@ func NewModuleComponent(o component.Options) (*ModuleComponent, error) {
 
 // LoadFlowContent loads the flow controller with the current component content. It
 // will set the component health in addition to return the error so that the consumer
-// can rely on either or both.
+// can rely on either or both. If the content is the same as the last time it was
+// successfully loaded, it will not be reloaded.
 func (c *ModuleComponent) LoadFlowContent(args map[string]any, contentValue string) error {
+	if contentValue == c.getLatestContent() {
```
mattdurham, line 45: Do we need to check args also?

mattdurham: Thinking of if an argument with a default was not specified and then the argument was added and the content is the same.

rfratto: This sounds right to me, LoadFlowContent should be a no-op if _both_ the content or the args change. If either of them changed, then it should go through.

rfratto (approved):
LGTM % Matt's comment that needs to be resolved.

component/module/module.go
erikbaranowski: @mattdurham, take a peek at what I did in the latest. I'm a little uncomfortable with the arguments being a `map[string]any` for correct comparison in all cases. do you think we need to do something more significant in how arguments are handled or do you think this covers it for now?

mattdurham: That should work, since DeepEqual errs on the side of return false.

mattdurham (approved):
LGTM

PR #5057, opened 2023-09-01
Sync cri stage in loki.process with promtail.

### LINKED ISSUE #4883: `loki.process` `cri` stage should support max line sizes limits like `promtail` does
### Request

Currently, one can configure max partial line sizes in promtail's `cri` stage: https://github.com/grafana/loki/blob/8abe080fd6d04dc5efb074ee0764a9bfd39c82d2/clients/pkg/logentry/stages/extensions.go#L125-L129

for example:
```
type CriConfig struct {
	MaxPartialLines            int              `mapstructure:"max_partial_lines"`
	MaxPartialLineSize         flagext.ByteSize `mapstructure:"max_partial_line_size"`
	MaxPartialLineSizeTruncate bool             `mapstructure:"max_partial_line_size_truncate"`
}
```

This should be configurable in `loki.process` too, since it's a useful protection from OOM when log files are malformed or too large. 

### Use case

It's a useful protection from OOM when log files are malformed or too large.

ptodev:
#### PR Description

There is a [new feature](https://github.com/grafana/loki/pull/9508) in promtail - this PR is adding it to `loki.process`.

Fixes #4883

#### Notes to the Reviewer

* I updated the promtail config converter test file, but I'm not sure if that's all I need to do for the converters?
* I updated the docs to say "number" instead of "int", to make it more consistent with the River [Types and values](https://grafana.com/docs/agent/latest/flow/config-language/expressions/types_and_values/) documentation.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/component/loki/process/stages/extensions.go
+++ b/component/loki/process/stages/extensions.go
@@ -13,8 +13,6 @@ import (
 	"github.com/grafana/river"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/model"
-
-	"github.com/grafana/loki/pkg/util/flagext"
 )
 
 const (
@@ -28,9 +26,9 @@ type DockerConfig struct{}
 // CRIConfig is an empty struct that is used to enable a pre-defined pipeline
 // for decoding entries that are using the CRI logging format.
 type CRIConfig struct {
-	MaxPartialLines            int              `river:"max_partial_lines,attr,optional"`
-	MaxPartialLineSize         flagext.ByteSize `river:"max_partial_line_size,attr,optional"`
-	MaxPartialLineSizeTruncate bool             `river:"max_partial_line_size_truncate,attr,optional"`
+	MaxPartialLines            int    `river:"max_partial_lines,attr,optional"`
+	MaxPartialLineSize         uint64 `river:"max_partial_line_size,attr,optional"`
+	MaxPartialLineSizeTruncate bool   `river:"max_partial_line_size_truncate,attr,optional"`
 }
 
 var (
@@ -163,8 +161,8 @@ func (c *cri) Run(entry chan Entry) chan Entry {
 }
 
 func (c *cri) ensureTruncateIfRequired(e *Entry) {
-	if c.cfg.MaxPartialLineSizeTruncate && len(e.Line) > c.cfg.MaxPartialLineSize.Val() {
-		e.Line = e.Line[:c.cfg.MaxPartialLineSize.Val()]
+	if c.cfg.MaxPartialLineSizeTruncate && len(e.Line) > int(c.cfg.MaxPartialLineSize) {
+		e.Line = e.Line[:c.cfg.MaxPartialLineSize]
 	}
 }
 
--- a/component/loki/process/stages/extensions.go
+++ b/component/loki/process/stages/extensions.go
@@ -5,17 +5,20 @@ package stages
 // new code without being able to slowly review, examine and test them.
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
+	"github.com/grafana/river"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/model"
+
+	"github.com/grafana/loki/pkg/util/flagext"
 )
 
 const (
-	RFC3339Nano         = "RFC3339Nano"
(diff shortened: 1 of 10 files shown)


thampiotr (commented):
Looking good, thanks!

docs/sources/flow/reference/components/loki.process.md
```diff
@@ -363,11 +368,11 @@ The following arguments are supported:
 
 | Name                  | Type     | Description                                                                      | Default | Required |
 | --------------------- | -------- | -------------------------------------------------------------------------------- | ------- | -------- |
-| `rate`                | `int`    | The maximum rate of lines per second that the stage forwards.                    |         | yes      |
-| `burst`               | `int`    | The cap in the quantity of burst lines that the stage forwards.                  |         | yes      |
+| `rate`                | `number` | The maximum rate of lines per second that the stage forwards.                    |         | yes      |
+| `burst`               | `number` | The cap in the quantity of burst lines that the stage forwards.                  |         | yes      |
```
clayton-cornell, line 372:
```suggestion
| `burst`               | `number` | The maximum number of burst lines that the stage forwards.                  |         | yes      |
```
Could we rephrase to be consistent with the other descriptions?

clayton-cornell (commented):
Agree with Piotr's commens.. plus one more :-)

component/loki/process/stages/extensions.go
```diff
@@ -24,7 +27,37 @@ type DockerConfig struct{}
...
+var (
+	_ river.Defaulter = (*CRIConfig)(nil)
+	_ river.Validator = (*CRIConfig)(nil)
+)
+
+// DefaultCRIConfig applies the default values on
+var DefaultCRIConfig = CRIConfig{
+	MaxPartialLines:            100,
+	MaxPartialLineSize:         0,
+	MaxPartialLineSizeTruncate: false,
```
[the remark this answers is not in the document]
thampiotr: Ah, I was under impression that you can set `max_partial_line_size_truncate` to `false` and then the lines that are longer than `max_partial_line_size` will be just dropped instead of truncated. That was the only logical reason to have two config options, but that's not how this works and indeed they are redundant as you've mentioned in the comment above.

component/loki/process/stages/extensions.go
```diff
@@ -24,7 +27,37 @@ type DockerConfig struct{}
...
+
+var (
+	_ river.Defaulter = (*CRIConfig)(nil)
+	_ river.Validator = (*CRIConfig)(nil)
+)
+
+// DefaultCRIConfig applies the default values on
+var DefaultCRIConfig = CRIConfig{
+	MaxPartialLines:            100,
+	MaxPartialLineSize:         0,
```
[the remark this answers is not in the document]
thampiotr: That was my misunderstanding. I would be in favour of only having one configuration value though. A logic in config converter could be implemented to do the right thing. But I'll leave it up to you :)

component/loki/process/stages/extensions.go
```diff
@@ -24,7 +27,37 @@ type DockerConfig struct{}
 
 // CRIConfig is an empty struct that is used to enable a pre-defined pipeline
 // for decoding entries that are using the CRI logging format.
-type CRIConfig struct{}
+type CRIConfig struct {
+	MaxPartialLines            int              `river:"max_partial_lines,attr,optional"`
+	MaxPartialLineSize         flagext.ByteSize `river:"max_partial_line_size,attr,optional"`
```
rfratto, line 32: Using flagext.ByteSize here doesn't work because it doesn't implement encoding.TextMarshaler/encoding.TextUnmarshaler and a uint64 can't be directly assigned to it. 

While we could use the underlying ByteSize type, I think we should swap this out for the underlying uint64 since the Loki docs don't mention that this _could_ be a string. 

Since I'm working on helping the tests pass, I'll push this code for you and you can decide if it's OK tomorrow.

rfratto: > Since I'm working on helping the tests pass, I'll push this code for you and you can decide if it's OK tomorrow.

Also, sorry for jumping in. I saw the PR was approved but a test was failing and thought it'd be a quick fix so we can get this merged, but it turned out to be a bit more than I had anticipated.

ptodev: Thanks for looking at this, and no need to apologise :)
I was looking at the same thing yesterday, reached the same conclusion, and had prepared a very similar change which wasn't pushed. Yours looks a bit nicer, so I'm happy to go with your suggestion.

PR #5062, opened 2023-09-01
Flow discovery consulagent component

### LINKED ISSUE #2261: Flow: discovery.consulagent component
For feature parity, we should bring over the `consulagent` SD mechanism from Promtail and introduce it as a Flow component.

Note that unlike most of the SD mechanisms used in the static mode agent, this is the only mechanism that doesn't come from Prometheus.

wildum:
#### PR Description

Add the component discovery.consulagent. The implementation is taken from Promtail's code: https://github.com/grafana/loki/tree/main/clients/pkg/promtail/discovery/consulagent

Fixes #2261 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


docs/sources/flow/reference/components/discovery.consulagent.md
```diff
@@ -0,0 +1,123 @@
+---
+canonical: https://grafana.com/docs/agent/latest/flow/reference/components/discovery.consulagent/
+title: discovery.consulagent
+---
+
+# discovery.consulagent
+
+`discovery.consulagent` allows retrieving scrape targets from [Consul's Agent API][].
```
clayton-cornell, line 8:
```suggestion
`discovery.consulagent` allows you to retrieve scrape targets from [Consul's Agent API][].
```

clayton-cornell (commented):
Minor suggestion for docs. Otherwise good enough as-is

component/discovery/consulagent/consulagent.go
```diff
@@ -0,0 +1,85 @@
...
+	Token           rivertypes.Secret `river:"token,attr,optional"`
+	Datacenter      string            `river:"datacenter,attr,optional"`
+	TagSeparator    string            `river:"tag_separator,attr,optional"`
+	Scheme          string            `river:"scheme,attr,optional"`
+	Username        string            `river:"username,attr,optional"`
+	Password        rivertypes.Secret `river:"password,attr,optional"`
+	RefreshInterval time.Duration     `river:"refresh_interval,attr,optional"`
+	Services        []string          `river:"services,attr,optional"`
+	ServiceTags     []string          `river:"tags,attr,optional"`
+	TLSConfig       config.TLSConfig  `river:"tls_config,block,optional"`
```
tpaschalis, line 38: Do we also need to configure the `node_meta` field that exist in the [upstream config](https://grafana.com/docs/loki/v2.8.x/clients/promtail/configuration/)?

component/discovery/consulagent/consulagent.go
```diff
@@ -0,0 +1,85 @@
...
+
+type Arguments struct {
+	Server          string            `river:"server,attr,optional"`
+	Token           rivertypes.Secret `river:"token,attr,optional"`
+	Datacenter      string            `river:"datacenter,attr,optional"`
+	TagSeparator    string            `river:"tag_separator,attr,optional"`
+	Scheme          string            `river:"scheme,attr,optional"`
+	Username        string            `river:"username,attr,optional"`
+	Password        rivertypes.Secret `river:"password,attr,optional"`
+	RefreshInterval time.Duration     `river:"refresh_interval,attr,optional"`
```
tpaschalis, line 35: I don't see a use for this field in the upstream configuration, do we need this?

docs/sources/flow/reference/components/discovery.consulagent.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Arguments
+
+The following arguments are supported:
+
+| Name               | Type           | Description                                                                                                    | Default          | Required |
+| ------------------ | -------------- | -------------------------------------------------------------------------------------------------------------- | ---------------- | -------- |
+| `server`           | `string`       | Host and port of the Consul Agent API.                                                                         | `localhost:8500` | no       |
+| `token`            | `secret`       | Secret token used to access the Consul Agent API.                                                              |                  | no       |
+| `datacenter`       | `string`       | Datacenter in which the Consul Agent is configured to run. If not provided, the default is used.               |                  | no       |
```
tpaschalis, line 30: > If not provided, the default is used

What is the default here?

converter/internal/promtailconvert/testdata/consulagent.yaml
```diff
@@ -0,0 +1,35 @@
+scrape_configs:
+  - job_name: fun
+    consulagent_sd_configs:
+      - server: 'localhost:8500'
+        datacenter: bigdata
+        scheme: sketchy
+        username: toby
+        password: this_password_is_safe_innit?
```
tpaschalis, line 8: Undoubtedly! :D

converter/internal/promtailconvert/testdata/consulagent.yaml
```diff
@@ -0,0 +1,35 @@
...
+        tags:
+          - better
+          - faster
+          - stronger
+        services: [ 'myapp' ]
+        node_meta:
+          where: here
+          what: this
+        tag_separator: ";"
+        allow_stale: false
```
tpaschalis, line 19: I'm also not sure what this field is; should we have it as configurable?

component/discovery/consulagent/promtail_consulagent.go
```diff
@@ -0,0 +1,563 @@
+// This code is copied from Promtail (https://https://github.com/grafana/loki/tree/cff8324f241cdc13318ff4b834be750a15ce05fc/clients/pkg/promtail/discovery/consulagent).
+// Some changes have been made to improve the component:
+// - the init function has been removed
+// - the unused config "AllowStale" and "NodeMeta" have been removed
```
tpaschalis, line 4: I just now saw the comment on the two fields.

component/discovery/consulagent/consulagent.go
wildum: Do you mean RefreshInterval? In the promtail implementation the RefreshInterval is used in the Run function to create a ticker which checks repeatedly for new services to watch

component/discovery/consulagent/consulagent.go
wildum: like allow_stale, node_meta is not needed for the consul agent, it a copy paste mistake from the prometheus code that they made in promtail. I removed in the agent because the field was not used

docs/sources/flow/reference/components/discovery.consulagent.md
wildum: according to the code if not set explicitly the datacenter is retrieved from the config of the consulagent and would represent the local node's datacenter. I could improve the doc here.

docs/sources/flow/reference/components/discovery.consulagent.md
```diff
@@ -0,0 +1,125 @@
...
+```river
+discovery.consulagent "example" {
+  server = "localhost:8500"
+  services = [
+    "service1",
+    "service2",
+  ]
+}
+
+prometheus.scrape "demo" {
```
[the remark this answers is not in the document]
wildum: I tried this demo locally with a consul agent running and a sample service and in grafana cloud I could see the metric telling me when my service was up. But I don't have experience with consul so if you think that an example with loki would be more helpful for users Im happy to update it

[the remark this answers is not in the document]
wildum: would it be loki.source.file instead of prometheus.scrape and loki.write instead of prometheus.remote_write?

docs/sources/flow/reference/components/discovery.consulagent.md
```diff
@@ -0,0 +1,125 @@
...
+configuration. In those cases, exported fields retain their last healthy
+values.
+
+## Debug information
+
+`discovery.consulagent` does not expose any component-specific debug information.
+
+### Debug metrics
+
+`discovery.consulagent` does not expose any component-specific debug metrics.
```
tpaschalis, line 89: We should document the component's internal metrics here, too.

docs/sources/flow/reference/components/discovery.consulagent.md
wildum: Actually, I'm not even sure that it makes sense to have this field exposed here. Why would someone need to set the datacenter when we can retrieve it from the agent directly?

docs/sources/flow/reference/components/discovery.consulagent.md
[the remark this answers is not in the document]
tpaschalis: Yes, exactly. It's more in-line with what someone who's looking for `consulagent` will look for.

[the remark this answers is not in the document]
tpaschalis: Actually, they'd need still need to somehow construct the special `__path__` label, which I'm not sure would work on their case. Let's leave this example as-is for now and we can revisit later. 

Can you add an HTML-style comment `<!-- TODO: Include a logging example -->` on the markdown so we add it in the future?

docs/sources/flow/reference/components/discovery.consulagent.md
tpaschalis: > Why would someone need to set the datacenter when we can retrieve it from the agent directly?

Oh you mean the _Consul_ agent here. Yeah, since the upstream struct allows to configure this, I think we should expose this. It's also a less-surprising UX from the converters perspective.

>  I could improve the doc here.
Yeah, let's do that and say where the default comes from.

component/discovery/consulagent/consulagent.go
wildum: I will put a warning for the node_meta. The allow_stale is not in the doc of promtail and also it's a bool which makes it a bit awkward to check whether it was set or not so maybe I don't put a warning for this one?

tpaschalis: Yep, I'm okay with this.

tpaschalis (approved):
LGTM, nice!

PR #5065, opened 2023-09-01
loki.process: fix deadlock in case of frequent Updates

tpaschalis:
#### PR Description
This PR fixes a deadlock candidate in `loki.process` when Update was concurrently called with a log line being fanned out.

We've already fixed similar problems in other components, as #5061 mentions we should go through all of the loki components to look for similar issues.

No issue filed

#### Notes to the Reviewer
Let me know if you know of a better way to detect partial deadlocks here.

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Tests updated

### DIFF
--- a/component/loki/process/process.go
+++ b/component/loki/process/process.go
@@ -85,7 +85,6 @@ func (c *Component) Run(ctx context.Context) error {
 		if c.entryHandler != nil {
 			c.entryHandler.Stop()
 		}
-		close(c.processOut)
 		close(c.processIn)
 		c.mut.RUnlock()
 	}()
--- a/component/loki/process/process.go
+++ b/component/loki/process/process.go
@@ -49,11 +49,13 @@ type Component struct {
 
 	mut          sync.RWMutex
 	receiver     loki.LogsReceiver
-	fanout       []loki.LogsReceiver
 	processIn    chan<- loki.Entry
 	processOut   chan loki.Entry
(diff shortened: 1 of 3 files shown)


component/loki/process/process_test.go
```diff
@@ -387,3 +387,100 @@ stage.static_labels {
...
+	go func() {
+		for {
+			c.Update(args1)
+			c.Update(args2)
+		}
+	}()
+
+	// Run everything for a while
+	time.Sleep(1 * time.Second)
+	require.WithinDuration(t, time.Now(), lastSend, 5*time.Millisecond)
```
tpaschalis, line 485: As a team we've tried to avoid using time.Sleep in tests, but in this case I'm not sure what's the better option.

Adding a backoff and looking in case a partial deadlock is detected earlier on? LMK what you think.

component/loki/process/process.go
```diff
@@ -155,16 +161,17 @@ func (c *Component) handleOut(ctx context.Context, wg *sync.WaitGroup) {
 		case <-ctx.Done():
 			return
 		case entry := <-c.processOut:
-			c.mut.RLock()
-			for _, f := range c.fanout {
+			c.fanoutMut.RLock()
+			fanout := c.fanout
+			c.fanoutMut.RUnlock()
```
mattdurham, line 166: This feels like a band-aid, but if it solves the short term I am all for it.

PR #5066, opened 2023-09-01
Remove dependency on cortexproject/cortex

### LINKED ISSUE #5064: Remove dependency on cortex
### Request

The Agent only uses cortex for `"github.com/cortexproject/cortex/pkg/util/test"`. We should duplicate this so we don't have a dependency on cortex. 

### Use case

Reduce dependencies and surface area to issues.

rfratto:
Fixes #5064


mattdurham (approved):
LGTM

rfratto:
Waiting to see if the new settings work with the macOS runner, and then merging if they do.

PR #5068, opened 2023-09-01
pkg/metrics/wal: remove deleted series from WAL sooner

rfratto:
This change causes deleted series to be deleted earlier based on the last segment involved in the checkpoint rather than the first segment involved in the checkpoint.

This will then cause the next truncate to be more aggressive with dropped series, which should help free memory associated with deleted series in remote_write sooner.

Mirrors prometheus/prometheus#12297 from @bboreham.


tpaschalis (approved):
LGTM

PR #5069, opened 2023-09-01
refactor promtail converter sd configs so they reuse prometheus conve…

erikbaranowski:
…rter functionality

#### PR Description

Use higher level functions from the prometheusconvert API for generating the service discovery river in the promtail converter. Some future work may be done to improve consuming the prometheusconvert API but this gets it all wired up in one place and saves us from modifying the promtail converter when implementing new flow discovery components.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated


converter/internal/prometheusconvert/prometheusconvert.go
```diff
@@ -154,7 +154,7 @@ func appendServiceDiscoveryConfigs(pb *prometheusBlocks, serviceDiscoveryConfig
 			exports = appendDiscoveryLightsail(pb, common.LabelWithIndex(labelCounts["lightsail"]-1, label), sdc)
 		}
 
-		targets = append(exports.Targets, targets...)
+		targets = append(targets, exports.Targets...)
```
erikbaranowski, line 157: This fixes a backwards ordering that was not a functional problem but weird. Discovered when leveraging this for promtail.

converter/internal/promtailconvert/internal/build/service_discovery.go
```diff
@@ -0,0 +1,95 @@
...
+	if len(sdConfigs) == 0 {
+		return
+	}
+
+	pb := prometheusconvert.NewPrometheusBlocks()
+	targets := prometheusconvert.AppendServiceDiscoveryConfigs(pb, sdConfigs, common.LabelForParts(s.globalCtx.LabelPrefix, s.cfg.JobName))
+	pb.AppendToFile(s.f)
+
+	for _, target := range targets {
+		s.allTargetsExps = append(s.allTargetsExps, target["__expr__"])
```
erikbaranowski, line 21: maybe promtail should be using the targets in the same way as prometheus but I'm not going to tackle that in this PR.

converter/internal/promtailconvert/internal/build/service_discovery.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	s.diags.AddAll(prometheusconvert.ValidateServiceDiscoveryConfigs(sdConfigs))
+}
+
+func toDiscoveryConfig(cfg *scrapeconfig.Config) prom_discover.Configs {
```
erikbaranowski, line 27: let's turn the promtail SD configs into a type compatible with the current prometheus API

converter/internal/promtailconvert/promtailconvert.go
```diff
@@ -132,20 +132,11 @@ func appendScrapeConfig(
 ) {
 
 	b := build.NewScrapeConfigBuilder(f, diags, cfg, gctx)
-	b.Validate()
 	b.Sanitize()
 
 	// Append all the SD components
-	b.AppendKubernetesSDs()
-	b.AppendDockerSDs()
 	b.AppendStaticSDs()
```
erikbaranowski, line 138: static SD configs didn't 'just work' and requires a little more digging to wire up. aiming for a future PR if we decide to do it.

converter/internal/promtailconvert/internal/build/consul_agent.go
wildum, line 1: consulagent is not taken from prometheus but from promtail, I'm implementing it in https://github.com/grafana/agent/pull/5062. Maybe it's better not to delete the file and remove the part in toDiscoveryConfig

converter/internal/promtailconvert/promtailconvert.go
tpaschalis: Let's add a TODO comment to remember why we're leaving this here, WDYT?

thampiotr (approved):
LGTM! Thanks for this, I'm very happy to see the boilerplate get lighter :)

tpaschalis (approved):
LGTM, let's merge this once @thampiotr had a chance to take a look.

converter/internal/promtailconvert/promtailconvert.go
thampiotr: I have figured out what was not working and implemented this in my last commit.

converter/internal/prometheusconvert/scrape.go
```diff
@@ -68,7 +69,9 @@ func getScrapeTargets(staticConfig prom_discovery.StaticConfig) []discovery.Targ
 		for _, labelSet := range target.Targets {
 			for labelName, labelValue := range labelSet {
 				targetMap[string(labelName)] = string(labelValue)
-				targets = append(targets, targetMap)
+				newMap := map[string]string{}
+				maps.Copy(newMap, targetMap)
+				targets = append(targets, newMap)
```
thampiotr, line 74: This was a subtle bug where the `targetMap` was being mutated. We need a copy here. This is why some test cases have the `__address__` label changed. I believe this is the correct behaviour now.

converter/internal/promtailconvert/internal/build/consul_agent.go
thampiotr: Discussed offline, there are some considerations on how to do this, so we'll move the discussion to @wildum's PR.

thampiotr:
Instead of adding a TODO for the static SD configs, I've implemented the conversion for them - this should work now. 
Also fixed one subtle issue with map being mutated. PTAL and I'm happy for this to land @tpaschalis

PR #5072, opened 2023-09-03
feat(exporter): add Flow component for agent Integration

### LINKED ISSUE #5021: Support a Flow equivalent to the agent integration in static mode
### Request

Add some kind of Flow component for easily self-collecting agent metrics in Flow mode.

### Use case

Feature parity with static mode, where the agent integration is commonly used on bare metal installs.

hainenber:
#### PR Description

Add a Flow component, equivalent to agent's integration in static mode

#### Which issue(s) this PR fixes/closes

Close #5021 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- ~~[ ] Tests updated~~
- ~~[ ] Config converters updated~~

### DIFF
--- a/component/prometheus/exporter/agent/agent.go
+++ b/component/prometheus/exporter/agent/agent.go
@@ -1,0 +1,42 @@
+package agent
+
+import (
+	"github.com/grafana/agent/component"
+	"github.com/grafana/agent/component/prometheus/exporter"
+	"github.com/grafana/agent/pkg/integrations"
+	"github.com/grafana/agent/pkg/integrations/agent"
+)
+
+func init() {
+	component.Register(component.Registration{
+		Name:          "prometheus.exporter.agent",
+		Args:          Arguments{},
+		Exports:       exporter.Exports{},
+		Singleton:     true,
+		NeedsServices: exporter.RequiredServices(),
+		Build:         exporter.New(createExporter, "agent"),
+	})
+}
+
+func createExporter(opts component.Options, args component.Arguments) (integrations.Integration, error) {
+	a := args.(Arguments)
+	return a.Convert().NewIntegration(opts.Logger)
+}
+
+// Arguments holds values which are used to configured the local.agent component.
+type Arguments struct{}
+
+// Exports holds the values exported by the local.agent component.
+type Exports struct{}
+
+// DefaultArguments defines the default settings
+var DefaultArguments = Arguments{}
+
+// SetToDefault implements river.Defaulter
+func (args *Arguments) SetToDefault() {
+	*args = DefaultArguments
+}
+
+func (a *Arguments) Convert() *agent.Config {
+	return &agent.Config{}
+}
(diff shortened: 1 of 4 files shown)


docs/sources/flow/reference/components/prometheus.exporter.agent.md
```diff
@@ -0,0 +1,70 @@
+---
+canonical: https://grafana.com/docs/agent/latest/flow/reference/components/prometheus.exporter.agent/
+title: prometheus.exporter.agen
+---
+
+# prometheus.exporter.agent
+The `prometheus.exporter.agent` component collects and expose metrics about the agent itself.
```
clayton-cornell, line 7:
```suggestion
The `prometheus.exporter.agent` component collects and exposes metrics about the agent itself.
```

clayton-cornell (commented):
Minor tweak for the docs

docs/sources/flow/reference/components/prometheus.exporter.agent.md
```diff
@@ -0,0 +1,70 @@
+---
+canonical: https://grafana.com/docs/agent/latest/flow/reference/components/prometheus.exporter.agent/
+title: prometheus.exporter.agen
```
marctc, line 3:
```suggestion
title: prometheus.exporter.agent
```

captncraig:
I think this may be more complicated than is needed for this to work. The common prometheus.exporter.* code is useful for proxying static mode integrations, but in this case we don't really need the complexity that adds. We really just need to export a list of arguments including the agent's existing `/metrics` endpoint. The only complication is using the `agent.internal` address instead of `localhost`. I don't know what the existing integration really gives us there, other than another package.

tpaschalis:
@captncraig could you expand on how this alternative could look like? It would help move this PR ahead with a different direction.

component/prometheus/exporter/agent/agent.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/grafana/agent/pkg/integrations"
+	"github.com/grafana/agent/pkg/integrations/agent"
+)
+
+func init() {
+	component.Register(component.Registration{
+		Name:          "prometheus.exporter.agent",
+		Args:          Arguments{},
+		Exports:       exporter.Exports{},
+		Singleton:     true,
```
captncraig, line 15: We should remove this singleton flag. It is going to be going away soon, and there isn't a hard technical reason you can't have 2 of these (even if it is silly :))

captncraig:
I tested this, and it works fine. Any simplifications I want can be addressed in the future. The only change I want is to remove the singleton flag, and fix the docs to reflect that.

Once that is done and the conflicts fixed, I am happy to merge this.

clayton-cornell (commented):
No changes for docs. OK as-is.

component/prometheus/exporter/agent/agent.go
hainenber: Singleton removed. Thanks for the context! 😄

captncraig (approved):

PR #5100, opened 2023-09-05
integrations/eventhandler: remove double-logging of events

rfratto:
This removes the info-level log line indicating that an event is logged. When agent logs are being collected, this makes the event appear as if it was double-collected, as the event appears both in the logs sent directly to Loki and in the agent logs.


erikbaranowski (approved):
LGTM

PR #5107, opened 2023-09-06
flow/logging: use log/slog internally for log handling

rfratto:
This commit updates the Flow logger to expose a log/slog.Handler. The slog.Handler implementation always reflects the current logging settings, even when calling WithAttrs or WithGroup. Please read the extensive comment at the top of handler.go for context and rationale behind the complexity of the implementation.

As of this commit, flow/logging.Logger continues to implement the log.Logger interface, but this will be removed once all references to go-kit/log are removed.

Note that with this change, slog [always](https://cs.opensource.google/go/go/+/refs/tags/go1.21.0:src/log/slog/handler.go;l=264) puts the log message at the _end_ of the set of builtins (timestamp, level, source, message). Previously, there may have been other fields which appeared before the message. This may be particularly jarring, since now fields like `component_id` appear _after_ the message:

Before log/slog:

```
ts=2023-09-06T00:51:53.908392Z level=info component=prometheus.remote_write.default subcomponent=rw remote_name=0a91bf url=http://localhost:9009/api/prom/push msg="Stopping remote storage..."
```

After log/slog: 

```
ts=2023-09-06T00:51:53.908392Z level=info msg="Stopping remote storage..." component=prometheus.remote_write.default subcomponent=rw remote_name=0a91bf url=http://localhost:9009/api/prom/push
```

While I wish the message appeared after all the other attributes, there's no way for us to change this without using a custom backend for handling log entries, which I'm not sure is worthwhile at the moment.


mattdurham:
Should this be in the changelog change due to the reordering of the fields?

mattdurham (commented):
LGTM

pkg/flow/logging/logger.go
```diff
@@ -3,77 +3,84 @@ package logging
...
+		format  formatVar
+		writer  writerVar
+	)
+
+	l := &Logger{
+		inner: w,
+
+		level:  &leveler,
+		format: &format,
+		writer: &writer,
```
tpaschalis, line 40: We only ever use these as a way to set the pointer's value, which is mirrored to the handler.

Should we remove these top-level fields and use the setters directly on the handler's fields?

tpaschalis (approved):
LGTM!

pkg/flow/logging/logger.go
```diff
@@ -96,3 +103,42 @@ func (fw *lokiWriter) Write(p []byte) (int, error) {
...
+	defer w.mut.Unlock()
+	w.w = inner
+}
+
+func (w *writerVar) Write(p []byte) (n int, err error) {
+	w.mut.RLock()
+	defer w.mut.RUnlock()
+
+	if w.w == nil {
+		return len(p), nil
```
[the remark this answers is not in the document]
rfratto: It matters a little; implementations of io.Writer are expected to either return `len(p)` or an error if they return less than `len(p)`. 

You can even see this behavior in the implementation of io.MultiWriter: https://cs.opensource.google/go/go/+/refs/tags/go1.21.1:src/io/multi.go;l=113-116

So in this case, since we've fully consumed the write (we just didn't do anything with it), we should return `len(p)`. I'll explain this in a comment.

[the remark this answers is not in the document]
rfratto: Actually, maybe it should be an error. I'll start with that and we'll see what happens.

pkg/flow/logging/logger.go
rfratto: I'd prefer not to; since WithAttrs and WithGroup are expected to propagate down the pointers down to their caller, it feels cleaner to me if they originate from something which isn't a handler and that the handlers are purely pulling the values rather than being set.

tpaschalis: Ok, sounds fair!

PR #5109, opened 2023-09-06
docs: fix stage.drop block example of loki.process component

sboschman:
#### PR Description

Just updated the stage.drop block example to use the `drop_counter_reason` argument instead of the (old?) `drop_reason` argument.


tpaschalis (approved):
Thank you for the PR!

PR #5119, opened 2023-09-06
feat(comp/loki): expose msg's offset as internal label

### LINKED ISSUE #5079: loki.kafka.source should provide the message offset
### Request

Currently the loki.source.kafka provides information about the partition of the message, but not the offset.
This information is crucial when working with any other tool against kafka and should be accessible through the loki.source.kafka.

So in my opinion there should be an internal label for the offset of the message, just like the partition.

### Use case

Some examples:
- Offset is required to interact through the kafka scripts with the message for example when using kafka-delete-records.sh
- Depending on the amount of messages published to a topic just using a timestamp is not enough

hainenber:
#### PR Description
Expose message's offset in `loki.source.kafka` as internal label in the format of `__meta_kafka_message_offset`

Fixes #5079 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,9 @@ Main (unreleased)
 - Added a `disable_high_cardinality_metrics` configuration flag to `otelcol`
   exporters and receivers to switch high cardinality debug metrics off.  (@glindstedt)
 
+- `loki.source.kafka` component now exposes internal label `__meta_kafka_offset`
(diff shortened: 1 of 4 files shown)


clayton-cornell (commented):
No doc changes. OK as-is

rfratto (approved):
LGTM! 

Looks like there's a lint error. Once you fix that, I'll merge this :)

PR #5121, opened 2023-09-06
Clustering: bump ckit version to add ipv6 support

### LINKED ISSUE #4902: Grafana agent fails on IPv6 only cluster with clustering enabled.
### What's wrong?

We're running an IPv6 only Kubernetes (AWS EKS) cluster.

This basically means that only the Kubernetes nodes have IPv4 addresses, all pods only have IPv6 addresses. All outgoing traffic to IPv4 addresses is NAT'ed by the nodes to their IPv4 address using NAT64. (And DNS64)
When running grafana-agent in clustering mode, the code tries to detect what IP address is should advertise and this code seems to be unable to detect an IPv6 address, causing clustering mode to fail.

This caused deployment using the grafana k8s-monitoring helm chart to fail with these errors:

```
grafana-agent ts=2023-08-23T14:00:27.107335793Z level=info boringcryptoenabled=false
grafana-agent Error: building clusterer: determining advertise address: 2 errors occurred:
config-reloader 2023/08/23 14:00:25 Watching directory: "/etc/agent"
grafana-agent     * interface "eth0" has no suitable addresses
grafana-agent     * interface "en0": route ip+net: no such network interface
```

If I disable clustering everything works.
Most likely the reason is that there is only an IPv6 in the ip list.

Presumably, if I could predict the pod address, I could probably override the cluster address with "--cluster.advertise-address" and it might just work.

### Steps to reproduce

- Setup an IPv6 only kubernetes cluster (AWS provides this option more or less out of the box.)
- Install the k8s-monitoring helm chart with default configuration parameters (as provided by grafana-cloud).

### System information

Linux 5.15.117 x86_64

### Software version

v0.35.2

### Configuration

```text
agent:
  clustering:
    enabled: true
  resources:
    limits:
      cpu: "500m"
      memory: "512Mi"
    requests:
      cpu: "50m"
      memory: "512Mi"
```

### Logs

```text
grafana-agent ts=2023-08-23T14:00:27.107335793Z level=info boringcryptoenabled=false
grafana-agent Error: building clusterer: determining advertise address: 2 errors occurred:
config-reloader 2023/08/23 14:00:25 Watching directory: "/etc/agent"
grafana-agent     * interface "eth0" has no suitable addresses
grafana-agent     * interface "en0": route ip+net: no such network interface
```

### LINKED ISSUE #4956: Add IPv6 support to ckit
ckit currently supports only IPv4 addresses. That means that users can't use the clustering feature with the agent in a IPv6 only cluster (see [#4902](https://github.com/grafana/agent/issues/4902)).

dskit supports IPv6 and its implementation can be used to add this feature to ckit https://github.com/grafana/dskit/commit/1725bcc1e9a19bcdc10c18df5ad433e3c5b22f0b

wildum:
#### PR Description

Clustering: bump ckit version to add ipv6 support. Also add support to select all network interfaces available.

Fixes #4902 
Fixes #4956 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [ ] Tests updated


cmd/internal/flowmode/cluster_builder.go
```diff
@@ -62,12 +62,20 @@ func buildClusterService(opts clusterOptions) (*cluster.Service, error) {
...
-			addr, err := advertise.FirstAddress(opts.AdvertiseInterfaces)
+			advertiseInterfaces := opts.AdvertiseInterfaces
+			if useAllInterfaces(advertiseInterfaces) {
+				advertiseInterfaces = nil
+			}
+			addr, err := advertise.FirstAddress(advertiseInterfaces)
 			if err != nil {
-				level.Warn(opts.Log).Log("msg", "could not find advertise address using default interface names, "+
+				level.Warn(opts.Log).Log("msg", "could not find advertise address using network interfaces", advertiseInterfaces,
 					"falling back to localhost", "err", err)
```
wildum, line 72: @rfratto @tpaschalis why do we fall back to localhost in this case? As a user, if I specify some advertise interfaces and the agent fails to retrieve a valid IP address from it, I would prefer that it returns an error rather than falling back to localhost, which was not an address I wanted. Also, the user might not directly realize that the IP is wrong because the log in the terminal is drowned in startup logs

rfratto: @thampiotr added this code and would have the most context here.

wildum: oops sorry, since that it was clustering I just pinged both of you without actually looking at the history

docs/sources/flow/reference/cli/run.md
```diff
@@ -50,7 +50,7 @@ The following flags are supported:
 * `--cluster.discover-peers`: List of key-value tuples for discovering peers (default `""`). Mutually exclusive with `--cluster.join-addresses`.
 * `--cluster.rejoin-interval`: How often to rejoin the list of peers (default `"60s"`).
 * `--cluster.advertise-address`: Address to advertise to other cluster nodes (default `""`).
-* `--cluster.advertise-interfaces`: List of interfaces used to infer an address to advertise. The first one available in the list will be selected (default `"eth0,en0"`).
+* `--cluster.advertise-interfaces`: List of interfaces used to infer an address to advertise. It can be set to `all` to use all available network interfaces on the system. (default `"eth0,en0"`).
```
clayton-cornell, line 53:
```suggestion
* `--cluster.advertise-interfaces`: List of interfaces used to infer an address to advertise. Set to `all` to use all available network interfaces on the system. (default `"eth0,en0"`).
```

clayton-cornell (commented):
Small simplification of the description.

cmd/internal/flowmode/cluster_builder.go
tpaschalis: I'll look into the overall PR tomorrow, but here's some context for your question. We landed this fix with @thampiotr to fix `main-windows` from when the cluster service was first introduced.

At that point, there was no way to configure the advertise interfaces, so there was no way for the Agent to ever start up (since there was no `eth0` or `en0` by default).

Now that interfaces are configurable we can look into this again, but in any case, we shouldn't fail to start the Agent for something that the user might not care about.

wildum: Ok makes sense, thanks.
I still prefer a missconfigured program to crash on startup rather than falling back on an alternative solution, especially that fixing the problem is easy now. But you made a point that some users might not care and be fine with the fall back. If you believe that the fall back is the best option then let's keep it

tpaschalis: I'm still of the opinion that falling back to localhost is the correct option here.

It's unlikely that someone would deploy a cluster on their _first_ interaction with the Agent, and even if they did, they'd read the docs about the flags and the default interfaces on Windows.

In general, we shouldn't surprise people by having the Agent fail and having to refer them to the docs so they can learn about a parameter they don't even need at that time.

cmd/internal/flowmode/cluster_builder.go
```diff
@@ -62,12 +62,20 @@ func buildClusterService(opts clusterOptions) (*cluster.Service, error) {
...
 		if opts.EnableClustering {
-			addr, err := advertise.FirstAddress(opts.AdvertiseInterfaces)
+			advertiseInterfaces := opts.AdvertiseInterfaces
+			if useAllInterfaces(advertiseInterfaces) {
+				advertiseInterfaces = nil
+			}
+			addr, err := advertise.FirstAddress(advertiseInterfaces)
 			if err != nil {
-				level.Warn(opts.Log).Log("msg", "could not find advertise address using default interface names, "+
+				level.Warn(opts.Log).Log("msg", "could not find advertise address using network interfaces", advertiseInterfaces,
```
tpaschalis, line 71: nit: let's log the opts.AdvertiseInterfaces parameter instead. That way, they will see `all` in logs (or whatever custom handling we might add in the future.

```suggestion
				level.Warn(opts.Log).Log("msg", "could not find advertise address using network interfaces", opts.AdvertiseInterfaces,
```

tpaschalis (approved):
LGTM, thanks for this!

PR #5132, opened 2023-09-07
validate unsupported disabled integrations in the static converter

erikbaranowski:
#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/converter/internal/staticconvert/validate.go
+++ b/converter/internal/staticconvert/validate.go
@@ -98,11 +98,12 @@ func validateMetrics(metricsConfig metrics.Config, grpcListenPort int) diag.Diag
 func validateIntegrations(integrationsConfig config.VersionedIntegrations) diag.Diagnostics {
 	var diags diag.Diagnostics
 
-	if len(integrationsConfig.EnabledIntegrations()) == 0 {
-		return diags
-	}
-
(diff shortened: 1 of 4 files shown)


converter/internal/staticconvert/validate.go
```diff
@@ -98,11 +98,12 @@ func validateMetrics(metricsConfig metrics.Config, grpcListenPort int) diag.Diag
 func validateIntegrations(integrationsConfig config.VersionedIntegrations) diag.Diagnostics {
 	var diags diag.Diagnostics
 
-	if len(integrationsConfig.EnabledIntegrations()) == 0 {
-		return diags
-	}
-
 	for _, integration := range integrationsConfig.ConfigV1.Integrations {
+		if !integration.Common.Enabled {
+			diags.AddWithDetail(diag.SeverityLevelError, fmt.Sprintf("unsupported disabled integration %s.", integration.Name()), "")
```
wildum, line 103: why not just "Add" instead of "AddWithDetail" since you set "" for the details?
`func (ds *Diagnostics) Add(severity Severity, message string)`

erikbaranowski: my main reason is that I was not being smart. updated, good catch

wildum (approved):

PR #5139, opened 2023-09-08
`prometheus.operator.*`: Allow Customization of Default `scrape_interval` and `scrape_timeout`

### LINKED ISSUE #4806: Possibility to adjust ScrapeInterval and ScrapeTimeout for prometheus.operator.servicemonitors
### Request

Currently the values for ScrapeInterval and ScrapeTimeout for the Prometheus Operator ServiceMonitor are set to the global default values. We would like to be able to adjust those values when deploying grafana-agent using the helm chart.

### Use case

We have recording rules defined, which are based on a time interval shorter than the ScrapeInterval of the ServiceMonitor. This leads to a situation, where we can't evaluate the rule.

sberz:
#### PR Description
This PR adds `scrape_interval` and `scrape_timeout` arguments to customize the default behavior of `prometheus.operator.podmonitors`, `prometheus.operator.probes`, and `prometheus.operator.servicemonitors`.
Prometheus Operator fell back to the `global.scrape_interval` and `global.scrape_timeout` when no override was provided in the Pod/ServiceMonitor CRDs. 

During implementation of this feature, I noticed `prometheus.operator.probes` did not pass `RelabelConfigs` to `ConfigGenerator`. The second commit fixes this bug.

Fixes #4806 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


component/prometheus/operator/types.go
```diff
@@ -21,6 +21,12 @@ type Arguments struct {
 	// Namespaces to search for monitor resources. Empty implies All namespaces
 	Namespaces []string `river:"namespaces,attr,optional"`
 
+	// ScrapeInterval is the default interval to scrape targets.
+	ScrapeInterval time.Duration `river:"scrape_interval,attr,optional"`
```
captncraig, line 25: I'd like to call this `DefaultScrapeInterval` and `DefaultScrapeTimeout`.

captncraig: I think it might make sense to make this an additional struct type called `scrape`, and include it as a block.

I will shortly be adding other things from the `Prometheus` crd like enforced honorLabels or limits, so there will be more stuff in there.

So it will be something like:

```
scrape {
  default_scrape_interval = "15s"
}
```

Then you can just pass that whole thing to the manager stuff.

captncraig (changes requested):
This looks really good so far, thank you. The only thing I want to change is making the fields for this into a dedicated struct, since I plan to add more similar settings.

component/prometheus/operator/types.go
sberz: Addressed in 36ab5ccdb2ccc566e4c4c4effa0775d8da41adb1

docs/sources/shared/flow/reference/components/prom-operator-scrape.md
```diff
@@ -0,0 +1,11 @@
...
+aliases:
+- /docs/agent/shared/flow/reference/components/prom-operator-scrape/
+canonical: https://grafana.com/docs/agent/latest/shared/flow/reference/components/prom-operator-scrape/
+headless: true
+---
+
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`default_scrape_interval` | `duration` | How frequently to scrape targets. Used as default if no value is provided by the target resource. | `1m` | no
+`default_scrape_timeout` | `duration` | How long until a scrape request times out. Used as default if no value is provided by the target resource.| `10s` | no
```
clayton-cornell, line 11: What do you mean by `Used as default`? Are the arguments used as default or are the 1m and 10s values used as default if you don't specify a value for `default_scrape_interval` and default_scrape_timeout`?

captncraig: Those values will be used if a given Service Monitor does not specify the interval. Currently a single hard-coded default is used in that case.

clayton-cornell:
```suggestion
`default_scrape_interval` | `duration` | The default interval between scraping targets. Used as the default if the target resource does not provide a scrape interval. | `1m` | no
`default_scrape_timeout` | `duration` | The default timeout for scrape requests. Used as the default if the target resource does not provide a scrape timeout. | `10s` | no
```

Does this suggestion clarify this in the description?

sberz: This sounds a lot better. I updated the descriptions in 28550c2792f18b6c13f12758518f58d17aefce8d and rebased everything to resolve the merge conflicts

clayton-cornell (approved):
Approved for docs
@captncraig over to you for final check/approvals?

sberz:
Rebased everything on main to resolve conflicts in `CHANGELOG.md`

captncraig (approved):

PR #5145, opened 2023-09-10
Allow using image digest in place of tag in `grafana-agent` Helm chart

### LINKED ISSUE #5134: Allow grafana-agent helm chart to use image digest instead of tag
### Request

Hi team, our company mandatory requirement is to use `digests` instead of `tags`. 

But grafana-agent helm chart accepts only tags:
https://github.com/grafana/agent/blob/498e414dff52b6fe97c73ae1e667dd4067bff124/operations/helm/charts/grafana-agent/templates/containers/_agent.yaml#L3C150-L3C150

Can it be implemented similar to bitnami charts? E.g.: https://github.com/bitnami/charts/blob/a26d06f91b80951dff888035e9436ca51d2f95cf/bitnami/nginx/values.yaml#L74C1-L78C13?

Thanks in advance!

### Use case

Tag mutability can cause security issues like bypassing image scanning checks. 
To get deterministic and repeatable deployments, use digests instead of tags.

hainenber:

Fixes #5134 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added


operations/helm/charts/grafana-agent/values.yaml
```diff
@@ -99,6 +99,8 @@ image:
   # -- (string) Grafana Agent image tag. When empty, the Chart's appVersion is
   # used.
   tag: null
+  # -- Grafana Agent image's digest. When set, will override `image.tag`.
+  digest: null
```
tpaschalis, line 103: If you check one of the Agent images on [Dockerhub](https://hub.docker.com/layers/grafana/agent/main/images/sha256-b60321251257feddb85df35faac9e09dd07de2b628bfcf2c6e6449ee5ae56411?context=explore), the digest is reported on the form of `sha256:b6032...` which wouldn't work if copy/pasted here as it would end up as `@sha256:sha256:b6032...`

I think we should either a) document the format we expect, or b) use the `substr` function to detect whether the `sha256:` prefix already exists and only add it then. WDYT?

hainenber: Im thinking of a mixture of both. We clarify acceptable formats for commit I'd: either "sha256:XYZ" or "XYZ", and also perform input modification with `substr` for first format

hainenber: hi there, I've pushed new commits to illustrate. PTAL when you have time. Thanks!

tpaschalis (approved):
Thank you for the contribution! 🙌

tpaschalis:
@hainenber could you please also add an entry to the Helm chart's changelog as well? Then I'll merge this.

hainenber:
Hi there, new entry to Helm chart's CHANGELOG added. Thanks for the heads-up!

rfratto:
@hainenber There's a different CHANGELOG.md for just the Helm chart since it has a different release cycle and different version numbers :) I've moved your entry to the correct file. Thanks for the contribution!

hainenber:
Thank you for the fix! 🙇

rfratto:
@hainenber The folk at Grafana are really grateful about all the PRs you've opened lately (10 in two weeks)! If you're interested, we'd like to send you some swag :) If you're interested, reach out to me at robert.fratto@grafana.com and I can let you know what information we'd need to send you things.

Issue #4652, opened 2023-07-31
Always use retry after header even if response code is not 429


mattdurham:
### Request

Always use retry after header even if response code is not 429

### Use case

If the server isn't properly sending a 429 but is sending a retry header we should honor that, so that load shedding can occur.

ptodev:
Is this for any HTTP status code, or just for 5xx ones?

mattdurham:
Any status code, has long as it sends back a retry header.

ptodev:
I have a couple of concerns if this has to apply for any status code:
* I'm not sure if it makes sense to retry status codes such as 1xx and 2xx.
* Given the context around why this issue was raised, I'm not sure if it's necessary to retry anything other than 429 and 5xx.

mattdurham:
I think the underlying assumption it is easier to override the status code incorrectly than applying a header. So in the case where you have an internal system that is wrapped by a gateway and the internal system adds a header to retry but the gateway doesnt wrap the status code correctly.

tpaschalis:
Closing this as completes; we have merged the Prometheus PR and brought it back to the Agent.

PR #5152, opened 2023-09-11
Chore: Respects Retry-After header for both 429 and 5xx status codes

ying-jeanne:
#### Notes to the Reviewer
Agent will now consider and adhere to the Retry-After header in response to both 429 and 5xx status codes.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,7 @@ Main (unreleased)
   have been standardized. The first fields will always be `ts`, `level`, and
   `msg`, followed by non-common fields. Previously, the position of `msg` was
   not consistent. (@rfratto)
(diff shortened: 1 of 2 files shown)


tpaschalis:
Hey there 👋

We have a tracking issue for this behaviour on the remote_write side on https://github.com/grafana/agent/issues/4652, which should be contributed back to Prometheus.

For the Agent Management-related use case, I'll cc @jcreixell for his view.

spartan0x117:
Hello! :wave:

This change would modify the behavior of fetching configurations over HTTP, and unfortunately not the behavior of `remote_write`. I'm closing this for now, but as @tpaschalis mentioned we have a ticket to honor `Retry-After` headers for more than just `HTTP 429` status codes :smiley_cat:

PR #5157, opened 2023-09-11
Add extra_labels object to BlackboxTarget block

spartan0x117:
#### PR Description
Adds a `labels` object to `prometheus.exporter.blackbox`'s `target` block. This allows setting labels individually per target.

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Documentation added
- [X] Tests updated

### DIFF
--- a/component/prometheus/exporter/blackbox/blackbox.go
+++ b/component/prometheus/exporter/blackbox/blackbox.go
@@ -38,10 +38,11 @@ func buildBlackboxTargets(baseTarget discovery.Target, args component.Arguments)
 	a := args.(Arguments)
 	for _, tgt := range a.Targets {
 		target := make(discovery.Target)
-		for k, v := range baseTarget {
+		// Set extra labels first, meaning that any other labels will override
+		for k, v := range tgt.ExtraLabels {
 			target[k] = v
 		}
-		for k, v := range tgt.ExtraLabels {
+		for k, v := range baseTarget {
 			target[k] = v
 		}
 
--- a/component/prometheus/exporter/blackbox/blackbox.go
+++ b/component/prometheus/exporter/blackbox/blackbox.go
@@ -39,7 +39,7 @@ func buildBlackboxTargets(baseTarget discovery.Target, args component.Arguments)
 	for _, tgt := range a.Targets {
 		target := make(discovery.Target)
 		// Set extra labels first, meaning that any other labels will override
-		for k, v := range tgt.ExtraLabels {
+		for k, v := range tgt.Labels {
 			target[k] = v
 		}
 		for k, v := range baseTarget {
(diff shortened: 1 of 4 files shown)


component/prometheus/exporter/blackbox/blackbox_test.go
```diff
@@ -170,3 +170,36 @@ func TestBuildBlackboxTargets(t *testing.T) {
...
+	args := component.Arguments(baseArgs)
+	targets := buildBlackboxTargets(baseTarget, args)
+	require.Equal(t, 1, len(targets))
+	require.Equal(t, "integrations/blackbox/target_a", targets[0]["job"])
+	require.Equal(t, "http://example.com", targets[0]["__param_target"])
+	require.Equal(t, "http_2xx", targets[0]["__param_module"])
+
+	require.Equal(t, "test", targets[0]["env"])
+	require.Equal(t, "bar", targets[0]["foo"])
+}
```
wildum, line 205: Maybe the previous test could have been extended instead of creating a new one because you are only setting a new attribute

docs/sources/flow/reference/components/prometheus.exporter.blackbox.md
```diff
@@ -167,4 +167,50 @@ Replace the following:
...
+  }
+}
+```
+
+Replace the following:
+
+- `PROMETHEUS_REMOTE_WRITE_URL`: The URL of the Prometheus remote_write-compatible server to send metrics to.
+- `USERNAME`: The username to use for authentication to the remote_write API.
+- `PASSWORD`: The password to use for authentication to the remote_write API.
+
```
wildum, line 214: Similar idea here, I wonder if the extra_labels field is worth a new part with a new example

spartan0x117: Good point, since there are multiple targets it makes it shows how to do it both ways :smile:

component/prometheus/exporter/blackbox/blackbox.go
```diff
@@ -41,6 +41,9 @@ func buildBlackboxTargets(baseTarget discovery.Target, args component.Arguments)
 		for k, v := range baseTarget {
 			target[k] = v
 		}
+		for k, v := range tgt.ExtraLabels {
+			target[k] = v
+		}
```
rfratto, line 46: Should these be allowed to override existing labels?

docs/sources/flow/reference/components/prometheus.exporter.blackbox.md
```diff
@@ -83,7 +83,7 @@ debug metrics.
 ### Collect metrics using a blackbox exporter config file
 
 This example uses a [`prometheus.scrape` component][scrape] to collect metrics
-from `prometheus.exporter.blackbox`:
+from `prometheus.exporter.blackbox`. It adds an extra label `env="dev"` to the metrics emitted by the `grafana` target. The `example` target will not have any extra labels.
```
clayton-cornell, line 86:
```suggestion
from `prometheus.exporter.blackbox`. It adds an extra label, `env="dev"`, to the metrics emitted by the `grafana` target. The `example` target does not have any extra labels.
```

component/prometheus/exporter/blackbox/blackbox.go
spartan0x117: Good point, moved this to set first so that any subsequently set labels will override what's set in `extra_labels`.

rfratto (approved):
I'm not 100% sure if `extra_labels` is the right name (should it just be `labels`? or something else) but I'll leave the decision up to you.

LGTM, I think this is a reasonable addition for a quality of life change.

clayton-cornell (approved):

PR #5165, opened 2023-09-12
Support starting log tailers at the end of file

### LINKED ISSUE #3102: Support starting log tailers at the end of files
Some Grafana Agent users may try to use Grafana Agent to collect historical log files which have tons of data on them. This can lead to increased resource consumption for logs which will never be stored (because they're outside of the acceptable time range), but still need to be read line-by-line since files get tailed starting at the beginning. 

Ideally, Flow components support a parameter for a maximum log age (such as only tailing Kubernetes logs which are newer than 4 hours old), but this isn't possible for files, where byte offsets in the file are used instead.

I would like to see an option added to `loki.source.file` which supports starting a log tailer at the end of files when a cached position is not found.

wildum:
#### PR Description

This PR adds the option to start tailing a file from the end. This can be valuable in case the user wants to skip the content of a large file and only read the new logs.

It is possible that when seeking to the end of the file we end up in the middle of a partial log. To avoid this problem the code will seek for the last newline character and start tailing from there.

Fixes #3102 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


docs/sources/flow/reference/components/loki.source.file.md
```diff
@@ -25,51 +25,57 @@ loki.source.file "LABEL" {
...
+| `forward_to`     | `list(LogsReceiver)` | List of receivers to send log entries to.                                         |         | yes      |
+| `encoding`       | `string`             | The encoding to convert from when reading files.                                  | `""`    | no       |
+| `start_from_eof` | `bool`               | Whether the file should be tailed from the end if a cached position is not found. | `false` | no       |
 
 The `encoding` argument must be a valid [IANA encoding][] name. If not set, it
 defaults to UTF-8.
 
+The `start_from_eof` argument is valuable when you want to tail a large file without reading its entire content.
+When set to true, only new logs will be read, ignoring the existing ones.
+
```
wildum, line 46: I formatted this document in a specific commit, I hope this does not make the review too annoying. I find the doc easier to read/modify when it is nicely formatted,

component/loki/source/file/tailer.go
```diff
@@ -61,6 +64,14 @@ func newTailer(metrics *metrics, logger log.Logger, handler loki.EntryHandler, p
 		positions.Remove(path, labels)
 	}
 
+	// If no cached position is found and the startFromEOF option is enabled.
+	if pos == 0 && startFromEOF {
+		pos, err = getLastLinePosition(path)
```
tpaschalis, line 69: Do you think we should also write the offset in the positions file? It will be written after the SyncPeriod ticker fires (every 10s), but I think it's a good idea to update it here as well.

wildum: This way we would avoid recomputing the offset and we would read the logs which were added since the previous offset computation. I think that this is a minor improvement, but this is easy to add and if the user would rather start fresh from the end he can always delete the position

CHANGELOG.md
```diff
@@ -57,6 +57,9 @@ Main (unreleased)
 - `loki.source.kafka` component now exposes internal label `__meta_kafka_offset`
   to indicate offset of consumed message. (@hainenber)
 
+- Add a`start_from_eof` attribute in `loki.source.file` to have the option to start tailing a file from the end if a cached position is not found.
```
rfratto, line 60: Nit: do we believe eof is a widely understood term for non-developers? If we do, then this is fine, otherwise we might want to pick a different attribute name.

rfratto (commented):
Nice job, thank you for searching for the newline in 1K chunks; that'll definitely help in scenarios where there are huge log files.

CHANGELOG.md
rfratto: @clayton-cornell would you happen to have any naming suggestions here?

tpaschalis: Some suggestions:

```
- ignore_previous_lines
- ignore_past_content
- skip_past_content
```

rfratto: Another way of doing it: 

```
initial_position = "end"  // or "start" for current behavior 
```

clayton-cornell: The audience we target at this level of configuration has to be fairly knowledgeable. Using `start_from_eof` doesn't raise a big flag for me.  Plus... it is defined in the **Arguments** section of the doc. We have lots of obscure references to very specific tooling throughout Agent.

If we really want to change it, it should be descriptive. Maybe `start_from_end_of_log`?  Or how about `tail_from_end` since it relates directly to the function... the user is tailing the logs`.

docs/sources/flow/reference/components/loki.source.file.md
```diff
@@ -25,51 +25,57 @@ loki.source.file "LABEL" {
...
- `forward_to` | `list(LogsReceiver)` | List of receivers to send log entries to.                  |         | yes      
- `encoding`   | `string`             | The encoding to convert from when reading files.           | `""`    | no       
+| Name             | Type                 | Description                                                                       | Default | Required |
+| ---------------- | -------------------- | --------------------------------------------------------------------------------- | ------- | -------- |
+| `targets`        | `list(map(string))`  | List of files to read from.                                                       |         | yes      |
+| `forward_to`     | `list(LogsReceiver)` | List of receivers to send log entries to.                                         |         | yes      |
+| `encoding`       | `string`             | The encoding to convert from when reading files.                                  | `""`    | no       |
+| `start_from_eof` | `bool`               | Whether the file should be tailed from the end if a cached position is not found. | `false` | no       |
```
clayton-cornell, line 39:
```suggestion
| `start_from_eof` | `bool`               | Whether a log file should be tailed from the end if a cached position is not found. | `false` | no       |
```

I assume there could be more than one log file?

docs/sources/flow/reference/components/loki.source.file.md
```diff
@@ -25,51 +25,57 @@ loki.source.file "LABEL" {
...
 
 The `encoding` argument must be a valid [IANA encoding][] name. If not set, it
 defaults to UTF-8.
 
+The `start_from_eof` argument is valuable when you want to tail a large file without reading its entire content.
```
clayton-cornell, line 44:
```suggestion
You can use the `start_from_eof` argument when you want to tail a large file without reading its entire content.
```

CHANGELOG.md
wildum: I'm gonna go with tail_from_end then, unless someone is against it

clayton-cornell (commented):
Docs look ok to me now

docs/sources/flow/reference/components/loki.source.file.md
```diff
@@ -104,26 +110,30 @@ configuration.
...
-* `loki_source_file_encoding_failures_total` (counter): Number of encoding failures.
-* `loki_source_file_files_active_total` (gauge): Number of active files.
+
+- `loki_source_file_read_bytes_total` (gauge): Number of bytes read.
+- `loki_source_file_file_bytes_total` (gauge): Number of bytes total.
+- `loki_source_file_read_lines_total` (counter): Number of lines read.
+- `loki_source_file_encoding_failures_total` (counter): Number of encoding failures.
+- `loki_source_file_files_active_total` (gauge): Number of active files.
 
 ## Component behavior
```
thampiotr, line 126: I think that this section should be amended to mention the option that you can tail?

docs/sources/flow/reference/components/loki.source.file.md
```diff
@@ -29,51 +29,57 @@ loki.source.file "LABEL" {
...
- `forward_to` | `list(LogsReceiver)` | List of receivers to send log entries to.                  |         | yes      
- `encoding`   | `string`             | The encoding to convert from when reading files.           | `""`    | no       
+| Name            | Type                 | Description                                                                         | Default | Required |
+| --------------- | -------------------- | ----------------------------------------------------------------------------------- | ------- | -------- |
+| `targets`       | `list(map(string))`  | List of files to read from.                                                         |         | yes      |
+| `forward_to`    | `list(LogsReceiver)` | List of receivers to send log entries to.                                           |         | yes      |
+| `encoding`      | `string`             | The encoding to convert from when reading files.                                    | `""`    | no       |
+| `tail_from_end` | `bool`               | Whether a log file should be tailed from the end if a cached position is not found. | `false` | no       |
```
thampiotr, line 43: Will users know what 'cached position' means? This is the first mention of it in this page. We should probably use similar vocabulary like the one in 'component behaviour' section.

wildum: changed to 'stored' to match the vocabulary of the section in the component behavior

component/loki/source/file/tailer.go
```diff
@@ -108,6 +121,62 @@ func newTailer(metrics *metrics, logger log.Logger, handler loki.EntryHandler, p
...
+			return 0, err
+		}
+
+		bytesRead, err := file.Read(buf)
+		if err != nil {
+			return 0, err
+		}
+
+		idx := bytes.LastIndexByte(buf[:bytesRead], '\n')
+		// newline found
```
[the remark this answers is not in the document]
wildum: Yes I just tested it and it works fine: saved the file with the CRLF ending and it started at the same line as with the LF ending, with the correct byte difference.
Thanks for pointing this out, I did not think about it before

component/loki/source/file/tailer.go
```diff
@@ -61,6 +64,17 @@ func newTailer(metrics *metrics, logger log.Logger, handler loki.EntryHandler, p
 		positions.Remove(path, labels)
 	}
 
+	// If no cached position is found and the tailFromEnd option is enabled.
+	if pos == 0 && tailFromEnd {
+		pos, err = getLastLinePosition(path)
```
tpaschalis, line 69: Why do we only do this if no cached position is found?

I'd guess the user knows better if they want to set up this parameter. Imagine someone setting `tail_from_end = true` and then immediately stopping the Agent. They come back two days later and re-start the Agent, and the option will not be respected, but it will replay the last two days of logs.

tpaschalis: cc @rfratto who opened the original issue.

rfratto: That's tricky. 

If the agent crashes, you wouldn't want it to skip straight to the end on the file after restarting, and unfortunately I don't think there's any way to use just the position information to determine whether a cached position should be ignored.

I think this is safe behavior for now, but potentially something we can address in the future. 

(And to be fair, "I ran the agent for a few minutes two days ago" is likely an extreme edge case)

tpaschalis: I see the counterpoint, makes sense. Since we've documented this, I think we're good to go.

tpaschalis (approved):
⛵

PR #5169, opened 2023-09-12
prometheus.exporter: standardize minimum set of labels, instance key

### LINKED ISSUE #2424: Flow: What labels should `prometheus.exporter.unix` export?

tpaschalis:
This PR proposes to standardize the minimum set of labels for prometheus.exporter.* to the following set, and making better use of the `instance` label.

```
			model.AddressLabel:      httpData.MemoryListenAddr,
			model.SchemeLabel:       "http",
			model.MetricsPathLabel:  path.Join(httpData.HTTPPathForComponent(opts.ID), "metrics"),
			"instance":              instance,
			"job":                   jobName,
			"__meta_component_name": componentName,
			"__meta_component_id":   opts.ID,
```

This has some changes
a) Removes the `__meta_agent_integration*` labels which were a leftover from the static mode implementation
b) Adds a pair of `__meta_component_name` and `__meta_component_id` labels. The component name defines the type of component being used, while the component ID is the fully quantified unique identifier that includes the component's label
c) The `instance` label still defaults to the Agent's hostname, but each integration has the opportunity to set it to something more meaningful. This means that two `prometheus.exporter.mysql` components will _not_ output the exact same metrics, but populate the `instance` label based on the DSN of the server they're trying to monitor.
d) Specifically for the cloudwatch integration, I copied what the static mode is doing and set the instance label to a hash of the configuration being used.

Fixes #2424

#### Notes to the Reviewer
Do you think there's a better way to achieve this than changing the Creator signature? How do you feel this abstraction has been working out for you, are you satisfied with it? On the long term, and as Flow mode is becoming _the_ way to run the Agent, I suspect the integrations code will be better off if it's embedded in the package that defines the components themselves.

Until `prometheus.exporter.unix` becomes a non-singleton component and can set a label, it needs special handling for the `__meta_component_name` and `__meta_component_id` labels to be set correctly.

#### PR Checklist

- [X] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


component/prometheus/exporter/exporter.go
```diff
@@ -133,16 +137,19 @@ func newExporter(creator Creator, name string, targetBuilderFunc func(discovery.
 		}
 		httpData := data.(http_service.Data)
 
+		componentName := opts.ID[:strings.LastIndex(opts.ID, ".")]
```
captncraig, line 140: What is the reasoning for this change just for `unix`? Can you add that to the PR description?

CHANGELOG.md
```diff
@@ -81,6 +86,7 @@ Main (unreleased)
 
 - Flow: improve river config validation step in `prometheus.scrape` by comparing `scrape_timeout` with `scrape_interval`. (@wildum)
 
+<<<<<<< HEAD
```
captncraig, line 89: merge conflict?

component/prometheus/exporter/cloudwatch/config.go
```diff
@@ -240,3 +243,13 @@ func toYACEDiscoveryJob(rj DiscoveryJob) *yaceConf.Job {
 	}
 	return job
 }
+
+// getHash calculates the MD5 hash of the river representation of the config.
+func getHash(a Arguments) string {
```
captncraig, line 248: I understand this is copying static mode, but this seems utterly ridiculous to me. I wish there was some more usable version of this. Would the component ID be preferable in this case?

component/prometheus/exporter/consul/consul.go
```diff
@@ -21,9 +21,9 @@ func init() {
 	})
 }
 
-func createExporter(opts component.Options, args component.Arguments) (integrations.Integration, error) {
+func createExporter(opts component.Options, args component.Arguments, defaultInstanceKey string) (integrations.Integration, string, error) {
 	a := args.(Arguments)
-	return a.Convert().NewIntegration(opts.Logger)
+	return integrations.NewIntegrationWithInstanceKey(opts.Logger, a.Convert(), defaultInstanceKey)
```
captncraig, line 26: I'm confused by the API here. There is already a mechanism to customize target by using `NewWithTargetBuilder` as your Build function. How is this different form that? I don't love that a component that is fine with the default now has to accept and pass the default through. It feels like an unnecessary detail that should be hidden until they decide to override it.

component/prometheus/exporter/exporter.go
tpaschalis: The reason is that `prometheus.exporter.unix` doesn't support labels, so we cannot correctly set the component name and ID labels correctly. 

I'll write this on the PR description so it's clearer.

CHANGELOG.md
tpaschalis: Oops, fixing this.

component/prometheus/exporter/cloudwatch/config.go
tpaschalis: Maybe yes, I don't have a preference, but I'd like to set it to something more representative of what's being monitored instead of just the component definition.

component/prometheus/exporter/consul/consul.go
tpaschalis: The use of the new `integrations.NewIntegrationWithInstanceKey` helper is to standardize calling the two `InstanceKey` and `NewIntegration` methods from the integrations' Config interface.

Since I'd like _all_ integrations to be able to call `InstanceKey()` to set their own instance label, I wanted to avoid using NewWithTargetBuilder everywhere, but I'm not against this approach if you think it's for the best.

Alternatively, how would you feel about removing the defaultInstance argument and always pass the result of `defaultInstance()` that resolves to the hostname?

captncraig (approved):

PR #5170, opened 2023-09-12
otelcol.processor.discovery should not modify its targets attribute

### LINKED ISSUE #5164: otelcol.processor.discovery causes lose of target data in input components
### What's wrong?

when using the `otelcol.processor.discovery` component, all components passed in the targets argument of the component lose their exported target data. 

so for example with this configuration:
```
discovery.docker "containers" {
  host = "unix:///tmp/docker.sock"
}
```
the UI will show that the discovery component has a number of containers with all of their metadata attributes. but once i change the config to this:
```
discovery.docker "containers" {
  host = "unix:///tmp/docker.sock"
}


otelcol.processor.discovery "traces_docker_attributes" {
	targets = discovery.docker.containers.targets

	output {
		traces = [
			...
		]
	}
}
```

when i reload the configuration the discovery components UI not shows me the exports section as `[{},{},{}]`. important to note the number of empty objects in the map matches the number of running containers but there is no data anymore.

### Steps to reproduce

running agent in flow mode at `0.36.1` with the following configs.
```
discovery.docker "containers" {
  host = "unix:///tmp/docker.sock"
}


otelcol.processor.discovery "traces_docker_attributes" {
	targets = discovery.docker.containers.targets

	output {
		traces = [
			...
		]
	}
}
```

tried the same with an agent runnig on the same version in a kubernetes cluster with:
```
discovery.kubernetes "pods" {
  role= "pod"
}


otelcol.processor.discovery "traces_pod_attributes" {
	targets = discovery.docker.pods.targets

	output {
		traces = [
			...
		]
	}
}

```

### System information

running the grafana agent container from docker hub for linux

### Software version

Grafana agent 0.36.1

### Configuration

```text
discovery.docker "containers" {
  host = "unix:///tmp/docker.sock"
}


otelcol.processor.discovery "traces_docker_attributes" {
	targets = discovery.docker.containers.targets

	output {
		traces = [
			...
		]
	}
}
```

### Logs

```text
only log line i could find that's relevant:
{"controller_id":"","duration":"196.955µs","level":"info","msg":"finished node evaluation","node_id":"discovery.docker.all_containers","trace_id":"4c7a9f4d3c40c39642d7ee408e0c6c08","ts":"2023-09-12T07:19:55.154350064Z"}
```

ptodev:
#### PR Description

Fixes a bug where `otelcol.processor.discovery` would modify one of its config attributes. This seems to have an effect on the component which is passing this attribute to `otelcol.processor.discovery`, because the attribute contains a map and is not deep copied.

Static mode is not affected because it deep copies the targets.

Fixes #5164

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,11 @@ Main (unreleased)
   `msg`, followed by non-common fields. Previously, the position of `msg` was
   not consistent. (@rfratto)
 
+### Bugfixes
+
+- Fixing a bug in `otelcol.processor.discovery` whose `targets` attribute would be modified in a way 
+  which impacted the component passing the `targets` to `otelcol.processor.discovery`. (@ptodev)
+
 v0.36.1 (2023-09-06)
 --------------------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,8 +73,7 @@ Main (unreleased)
 
 ### Bugfixes
 
-- Fixing a bug in `otelcol.processor.discovery` whose `targets` attribute would be modified in a way 
-  which impacted the component passing the `targets` to `otelcol.processor.discovery`. (@ptodev)
+- Fixed a bug where `otelcol.processor.discovery` could modify the `targets` passed by an upstream component. (@ptodev)
 
 v0.36.1 (2023-09-06)
 --------------------
--- a/component/otelcol/processor/discovery/discovery.go
+++ b/component/otelcol/processor/discovery/discovery.go
@@ -77,7 +77,6 @@ func (args *Arguments) Validate() error {
 
 // Component is the otelcol.exporter.discovery component.
 type Component struct {
-	cfg      Arguments
 	consumer *promsdconsumer.Consumer
 	logger   log.Logger
 }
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -71,6 +71,11 @@ Main (unreleased)
   `msg`, followed by non-common fields. Previously, the position of `msg` was
   not consistent. (@rfratto)
 
+### Bugfixes
+
+- Fixing a bug in `otelcol.processor.discovery` whose `targets` attribute would be modified in a way 
+  which impacted the component passing the `targets` to `otelcol.processor.discovery`. (@ptodev)
```
tpaschalis, line 77: nit: what do you think about rephrasing to this?

```suggestion
- Fixed a bug where `otelcol.processor.discovery` could modify the `targets` passed by an upstream component. (@ptodev)
```

tpaschalis (approved):
LGTM; we should be more careful around the [assumption](https://github.com/grafana/agent/blob/eebc15adf7ad05490e3d6eede42d96627103b0f0/component/component.go#L36-L45) that components treat their Arguments and Exports as immutable and freely share memory.

clayton-cornell (approved):
Docs changes are all fixing missing syntax. No additional changes/suggestions there.

I agree with the suggestion to the CHANGELOG. It's much clearer in the rewrite

PR #5178, opened 2023-09-13
Clarify that attribute processor does not drop signals

ptodev:
#### PR Description

There has been some confusion regarding the attributes processor. It contains an `exclude` block, and people sometimes get confused that this means the data will be excluded in downstream components. This is not the case.

I mention the tail sampling processor as a processor which can do this. However, it only works for traces. It would have been nice to also point to similar processors for logs and metrics, but I don't think we have such components in Flow at the moment. I think the otel components for dropping metrics and logs are [metricstransform processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/metricstransformprocessor) and [probabilisticsampler processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/probabilisticsamplerprocessor).

#### PR Checklist

- [ ] CHANGELOG.md updated
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/flow/reference/components/otelcol.processor.attributes.md
+++ b/docs/sources/flow/reference/components/otelcol.processor.attributes.md
@@ -164,9 +164,11 @@ For example, adding a `span_names` filter could cause the component to error if 
 
 The `exclude` block provides an option to exclude data from being fed into the [action] blocks based on the properties of a span, log, or metric records.
 
-> **NOTE**: Signals excluded by the `exclude` block will still be propagated to downstream components.
-> If you would like to not propagate certain signals to downstream components, 
-> consider a processor such as [otelcol.processor.tail_sampling](../otelcol.processor.tail_sampling/).
+{{% admonition type="note" %}}
+Signals excluded by the `exclude` block will still be propagated to downstream components as-is.
+If you would like to not propagate certain signals to downstream components, 
+consider a processor such as [otelcol.processor.tail_sampling](../otelcol.processor.tail_sampling/).
+{{% /admonition %}}
 
 {{< docs/shared lookup="flow/reference/components/match-properties-block.md" source="agent" version="<AGENT VERSION>" >}}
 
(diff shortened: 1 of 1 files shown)


docs/sources/flow/reference/components/otelcol.processor.attributes.md
```diff
@@ -164,6 +164,10 @@ For example, adding a `span_names` filter could cause the component to error if
 
 The `exclude` block provides an option to exclude data from being fed into the [action] blocks based on the properties of a span, log, or metric records.
```
tpaschalis, line 165: Is this true?
```suggestion
The `exclude` block provides an option to exclude data from being fed into the [action] blocks based on the properties of a span, log, or metric records. Excluded data will be forwarded as-is.
```

docs/sources/flow/reference/components/otelcol.processor.attributes.md
```diff
@@ -164,6 +164,10 @@ For example, adding a `span_names` filter could cause the component to error if
...
 
+> **NOTE**: Signals excluded by the `exclude` block will still be propagated to downstream components.
+> If you would like to not propagate certain signals to downstream components, 
+> consider a processor such as [otelcol.processor.tail_sampling](../otelcol.processor.tail_sampling/).
```
clayton-cornell, line 169:
```suggestion
{{% admonition type="note" %}}
Signals excluded by the `exclude` block are still be sent to downstream components.
To prevent specific signals from being sent to downstream components,  you can use a processor
such as [otelcol.processor.tail_sampling](../otelcol.processor.tail_sampling/).
{{% /admonition %}}
```

Better to use the admonition shortcode for notes where we can.

docs/sources/flow/reference/components/otelcol.processor.attributes.md
ptodev: It is true, but I will add it to the new block that I'm adding now, since clarifying this is the whole purpose of the block. I'm also not sure if readers would know what is meant by "forwarded". I'll change this:

> Signals excluded by the `exclude` block will still be propagated to downstream components.

To:

> Signals excluded by the `exclude` block will still be propagated to downstream components as-is.

clayton-cornell (approved):
@ptodev Docs look ok. Are there more changes pending?

ptodev:
> @ptodev Docs look ok. Are there more changes pending?

Nope! This is all I wanted to add.

PR #5188, opened 2023-09-13
otelcol: fix retry mechanism

rfratto:
The v0.80 upgrade introduced two new settings to the common retry mechanism shared between components. One of these new settings, `multiplier`, specifies the rate at which a backoff is applied.

We did not convert this new field after upgrading, leading the multiplier to be set to the Go default (zero). After the first retry, the backoff period would be multiplied by zero for each subsequent retry, causing retries to happen in a hot loop.

A follow up PR should expose these new settings to the user rather than hard coding them at the new upstream defaults. I've split the additions of the new fields and fixing the defaults across two separate PRs in case we want to release this in a patch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,8 @@ Main (unreleased)
 
 - Fixed a bug where `otelcol` components with a retry mechanism would not wait after the first retry. (@rfratto)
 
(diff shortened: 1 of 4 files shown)


erikbaranowski (approved):
LGTM

PR #5194, opened 2023-09-14
add `nil_to_zero` YACE config

### LINKED ISSUE #5075: grafana-agent forces nilToZero:true for YACE causing misfiring of BurstBalance monitors on AWS volume size changes
### What's wrong?

YACE default for nilToZero is false - for good reason. AWS BurstBalance metrics stop being produced when a volume goes from less than 3TB more than 3TB, as BurstBalance is only relevant for smaller volumes sizes.

With grafana-agent as is, this means changing the volume size across this transition causes burst balance monitors to fire, as the burst balance appears to drops to zero (though in practise it is actually infinite).

The comment attached to nilToZero=true says "Avoid producing absence of values in metrics" - I'm not clear what the implications are if we flipped grafana-agent to having the same behaviour OOTB as YACE and producing NaN for this metric. Maybe this should actually be configurable?

### Steps to reproduce

Setup grafana to retrieve the metric aws_rds_burst_balance_average for a small volume (<3TB) - if volume is idle, will report 100.
Increase volume size to >3TB - metric will now report as 0.

... or just inspect the code.

https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using_cloudwatch_ebs.html notes that burst balance is not reporting "if the baseline performance of volume exceeds the maximum burst performance". Admittedly, this is only a problem on ye-olde world gp2 storage (as well as the now fairly obscure st1/sc1) but we rely on gp2 as gp3 is not yet available in non-public AWS which are our main targets.

### System information

N/A

### Software version

Latest

### Configuration

_No response_

### Logs

_No response_

berler:
#### PR Description

Adds `nil_to_zero` config flag so `YACE` can be configured instead of hard-coding `true`.

Fixes #5075

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/flow/reference/components/prometheus.exporter.cloudwatch.md
+++ b/docs/sources/flow/reference/components/prometheus.exporter.cloudwatch.md
@@ -267,13 +267,13 @@ metrics.
 Follow [this guide](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/viewing_metrics_with_cloudwatch.html)
 on how to explore metrics, to easily pick the ones you need.
 
-| Name          | Type           | Description                                                        | Default                                                                                                             | Required |
-| ------------- | -------------- | ------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------- | -------- |
-| `name`        | `string`       | Metric name.                                                       |                                                                                                                     | yes      |
-| `statistics`  | `list(string)` | List of statistics to scrape. Ex: `"Minimum"`, `"Maximum"`, etc.   |                                                                                                                     | yes      |
-| `period`      | `duration`     | See [period][] section below.                                      |                                                                                                                     | yes      |
-| `length`      | `duration`     | See [period][] section below.                                      | Calculated based on `period`. See [period][] for details.                                                           | no       |
-| `nil_to_zero` | `bool`         | When `true`, `NaN` metric values are converted to 0.               | The value of `nil_to_zero` in the parent [static][] or [discovery][] block (`true` if not set in the parent block). | no       |
+| Name          | Type           | Description                                                               | Default                                                                                                             | Required |
+| ------------- | -------------- | ------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------- | -------- |
+| `name`        | `string`       | Metric name.                                                              |                                                                                                                     | yes      |
+| `statistics`  | `list(string)` | List of statistics to scrape. For example, `"Minimum"`, `"Maximum"`, etc. |                                                                                                                     | yes      |
+| `period`      | `duration`     | See [period][] section below.                                             |                                                                                                                     | yes      |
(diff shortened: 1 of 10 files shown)


mattdurham:
Overall looks good but we should add the conversion from static -> flow mode in `cloudwatch_exporter.go`

berler:
> Overall looks good but we should add the conversion from static -> flow mode in `cloudwatch_exporter.go`

Thanks, I think I've fixed that now. I'm not as familiar with this code so I'm not sure if I could still be missing some places.

berler:
@mattdurham I think the staticconvert code currently skips converting static jobs: https://github.com/grafana/agent/blob/0c708da54e5b0d6bc81119dbf3a72a9b88d99641/converter/internal/staticconvert/internal/build/cloudwatch_exporter.go#L32 -- should I fix this too?

berler:
@mattdurham Is this good to merge now? I fixed the converter so it now correctly converts static jobs.

berler:
Rebased to fix merge conflicts.

docs/sources/flow/reference/components/prometheus.exporter.cloudwatch.md
```diff
@@ -265,12 +267,13 @@ metrics.
...
-| `length`     | `duration`     | See [period][] section below.                                    | Calculated based on `period`. See [period][] for details. | no       |
+| Name          | Type           | Description                                                        | Default                                                                                                             | Required |
+| ------------- | -------------- | ------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------- | -------- |
+| `name`        | `string`       | Metric name.                                                       |                                                                                                                     | yes      |
+| `statistics`  | `list(string)` | List of statistics to scrape. Ex: `"Minimum"`, `"Maximum"`, etc.   |                                                                                                                     | yes      |
```
clayton-cornell, line 273:
```suggestion
| `statistics`  | `list(string)` | List of statistics to scrape. For example, `"Minimum"`, `"Maximum"`, etc.   |                                                                                                                     | yes      |
```

clayton-cornell (commented):
One small change for docs

docs/sources/flow/reference/components/prometheus.exporter.cloudwatch.md
berler: @clayton-cornell I'm fine committing this, but I didn't actually change this text. I had to reformat because the table columns needed to get bigger to support the new row that I added.

clayton-cornell: @berler That's fine. It's just some general cleanup we can do in the doc content - I try to catch these (the e.g. or Ex) when I spot them.

clayton-cornell:
@mattdurham Over to you.

clayton-cornell (approved):

mattdurham (approved):
LGTM

PR #5200, opened 2023-09-14
Rename component stability from "alpha" to "experimental".

ptodev:
#### PR Description

#4934 and #4936 named the component stability of `otelcol.processor.span` and `otelcol.connector.spanmetrics` as "alpha". This is an Otel Collector term, and should be replaced by the [Agent's](https://grafana.com/docs/agent/latest/stability/) "experimental" label.

The Agent has a policy of not downgrading stability. I think this change doesn't count as downgrading, because it's reasonable to expect that something labeled "alpha" is not stable.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [x] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/flow/reference/components/otelcol.connector.spanmetrics.md
+++ b/docs/sources/flow/reference/components/otelcol.connector.spanmetrics.md
@@ -11,6 +11,8 @@ title: otelcol.connector.spanmetrics
 
 # otelcol.connector.spanmetrics
 
+{{< docs/shared lookup="flow/stability/experimental.md" source="agent" version="<AGENT VERSION>" >}}
+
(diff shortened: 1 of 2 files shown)


rfratto (approved):
Thanks! These components should also include this shared snippet after the title: https://github.com/grafana/agent/blob/main/docs/sources/shared/flow/stability/experimental.md

See other beta/experimental components for an example of how this is done.

ptodev:
> Thanks! These components should also include this shared snippet after the title: https://github.com/grafana/agent/blob/main/docs/sources/shared/flow/stability/experimental.md
> 
> See other beta/experimental components for an example of how this is done.

Sorry, I knew about this but totally forgot about it. Thanks for the reminder!

PR #5208, opened 2023-09-14
Add openstack config converter

### LINKED ISSUE #5177: Add openstack config converter
### Request

The config converter is missing for openstack_sd_configs. We should implement it.

### Use case

The config converters are useful for migrating from static to flow.

wildum:
#### PR Description

Although we already had the discovery.openstack flow component, the config converter was missing. This PR is adding it.

Fixes #5177 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated


converter/internal/promtailconvert/testdata/openstack.river
```diff
@@ -0,0 +1,39 @@
...
+	domain_id                     = "domain_id"
+	application_credential_name   = "application_credential_name"
+	application_credential_id     = "application_credential_id"
+	application_credential_secret = "application_credential_secret"
+	role                          = "instance"
+	region                        = "region"
+	refresh_interval              = "20s"
+	port                          = 81
+	all_tenants                   = true
+	availability                  = "internal"
```
wildum, line 18: @erikbaranowski @thampiotr I wanted to test the conversion of the TLS config as well but if I set it in the yaml config, the converter will expect all the fields in the river file event if I set only some of them (I believe that this is because the prometheus OpenStack code creates a new TLS config from the given one) :
	tls_config                    = {
		ca_pem               = "",
		ca_file                  = "/home/robin/.ca",
		cert_pem             = "",
		cert_file               = "/home/robin/.cert",
		key_pem              = (secret),
		key_file                 = "/home/robin/.key",
		server_name          = "example.local",
		insecure_skip_verify = true,
		min_version          = "",
	}
Already this is not good because ca, cert and key are duplicated, either the _pem or the _ca should be configured.

And whether I set a value or nothing to the key in the yaml file, it will be converted to (secret):
                                -               key_pem              = "secretKey",
                                +               key_pem              = (secret),

If I try to set the value to (secret) it will not load the river file:
--- FAIL: TestConvert (0.12s)
    --- FAIL: TestConvert/openstack (0.00s)
        testing.go:212: 
                Error Trace:    /Users/williamdumont/agent/converter/internal/test_common/testing.go:212
                                                        /Users/williamdumont/agent/converter/internal/test_common/testing.go:166
                                                        /Users/williamdumont/agent/converter/internal/test_common/testing.go:66
                Error:          Received unexpected error:
                                TestConvert/openstack:23:27: component "secret" does not exist
                Test:           TestConvert/openstack
                Messages:       failed to load the River config: discovery.openstack "prometheus1" {

thampiotr: Have you tried/checked that your code uses this override hook? https://github.com/grafana/agent/blob/main/converter/internal/common/river_utils.go#L40-L41

wildum: it's not being called for the key_pem because I just realised that the tls_config is not being treated as a block but as an attribute: 

                                -       tls_config {
                                +       tls_config                    = {

wildum: of course, this is because I set it as an `attr`, my bad

erikbaranowski (approved):
LGTM, good catch on the `tls_config` attr vs block

PR #5219, opened 2023-09-14
component/otelcol: expose randomization_factor and multiplier in retry blocks

rfratto:
Follow up on #5188.

I'm not super sure on how to document these new fields clearly, help from @clayton-cornell would be appreciated on this.

### DIFF
--- a/docs/sources/shared/flow/reference/components/otelcol-retry-block.md
+++ b/docs/sources/shared/flow/reference/components/otelcol-retry-block.md
@@ -14,14 +14,21 @@ Name | Type | Description | Default | Required
 ---- | ---- | ----------- | ------- | --------
 `enabled` | `boolean` | Enables retrying failed requests. | `true` | no
 `initial_interval` | `duration` | Initial time to wait before retrying a failed request. | `"5s"` | no
+`randomization_factor` | `number` | Factor to randomize wait time before retrying. | `0.5` | no
+`multiplier` | `number` | Factor to grow wait time before retrying. | `1.5` | no
 `max_interval` | `duration` | Maximum time to wait between retries. | `"30s"` | no
 `max_elapsed_time` | `duration` | Maximum amount of time to wait before discarding a failed batch. | `"5m"` | no
 
 When `enabled` is `true`, failed batches are retried after a given interval.
 The `initial_interval` argument specifies how long to wait before the first
 retry attempt. If requests continue to fail, the time to wait before retrying
-increases exponentially. The `max_interval` argument specifies the upper bound
-of how long to wait between retries.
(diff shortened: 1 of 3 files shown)


tpaschalis (approved):
Leaving to @clayton-cornell for the final say, but at least to me it reads well enough.

docs/sources/shared/flow/reference/components/otelcol-retry-block.md
```diff
@@ -14,14 +14,21 @@ Name | Type | Description | Default | Required
...
 retry attempt. If requests continue to fail, the time to wait before retrying
-increases exponentially. The `max_interval` argument specifies the upper bound
-of how long to wait between retries.
+increases by the factor specified by the `multiplier` argument, which must be
+greater than `1.0`. The `max_interval` argument specifies the upper bound of
+how long to wait between retries.
+
+If `randomization_factor` is greater than `0`, the wait time before retries is
+multiplied by a random factor up to `1 ± randomization_factor`, allowing for
+jitter between retrying agents.
```
ptodev, line 31:
```suggestion
The `randomization_factor` argument is useful for adding jitter between retrying agents.
If `randomization_factor` is greater than `0`, the wait time before retries is
multiplied by a random factor in the range 
`[ I - randomization_factor * I, I + randomization_factor * I]`, 
where `I` is the current interval.
```

I was a bit confused about what this means:

> up to `1 ± randomization_factor`

I read the code and I rephrased it in a way that makes more sense to me personally. It makes it more clear that this is a random value in a certain range. Not sure if you like this though?

Btw, what if `randomization_factor` is less than 1? Would the config still be used? Or would it fail validation?

rfratto: > Btw, what if randomization_factor is less than 1? Would the config still be used? Or would it fail validation?

`randomization_factor` is less than 1 by default (it's 0.5). `0` means no randomization.

ptodev: Sorry, I made a typo. I meant to ask "what if `randomization_factor` is less than 0?"

rfratto: I'll add a check for it, but I think a negative randomization_factor also acts as if it were 0.

ptodev:
I find it weird that these two parameters are not documented in the Otel docs either. I asked a [question](https://cloud-native.slack.com/archives/C01N6P7KR6W/p1695051113186629) on the CNCF Slack channel whether this is intentional.

clayton-cornell (commented):
The rand factor is a bit complicated, but I think the rewrite by @ptodev makes sense in context.

So docs are OK as-is.

PR #5233, opened 2023-09-18
Expand documentation to demonstrate label options

### LINKED ISSUE #5232: Update flow docs to demonstrate and link to additional label options
### Request

When getting started with Flow, it is not immediately obvious that you can have additional labels as part of `discovery.http` to configure the scheme, path, timeout, and interval for the `prometheus.scrape` block that targets are passed to.

It would be good to update the docs with an example of how the labels should be presented to `discovery.http` and a link through to the "Technical Details" part of `prometheus.scrape` so users are more aware of the possibilites!

### Use case

It improves the user experience and eases adoption for targets that are outside the k8s or other discovery environments.

proffalken:

Fixes #5232 

#### PR Checklist

- [ x ] Documentation added

### DIFF
--- a/docs/sources/flow/reference/components/discovery.http.md
+++ b/docs/sources/flow/reference/components/discovery.http.md
@@ -27,6 +27,26 @@ Example response body:
 ]
 ```
 
+Example response setting the metrics path and port for [prometheus.scrape](../prometheus.scrape#technical-details) additional config:
+
+```json
+[
+   {
+      "labels" : {
+         "__metrics_path__" : "/api/prometheus",
+         "__scheme__" : "https",
+         "__scrape_interval__" : "60s",
+         "__scrape_timeout__" : "10s",
+         "service" : "custom-api-service"
+      },
+      "targets" : [
+         "custom-api:443"
+      ]
+   },
+]
+
+```
+
 ## Usage
 
 ```river
--- a/docs/sources/flow/reference/components/discovery.http.md
+++ b/docs/sources/flow/reference/components/discovery.http.md
@@ -27,7 +27,9 @@ Example response body:
 ]
 ```
 
-Example response setting the metrics path and port for [prometheus.scrape](../prometheus.scrape#technical-details) additional config:
+It is possible to use additional fields in the JSON to pass parameters to [prometheus.scrape](../prometheus.scrape#technical-details) such as the `metricsPath` and `scrape_interval`.
+
+As an example, the following will provide a target with a custom `metricsPath`, scrape interval, and timeout value:
 
 ```json
 [
@@ -46,6 +48,31 @@ Example response setting the metrics path and port for [prometheus.scrape](../pr
 ]
 
 ```
+
+It is also possible to append query parameters to the metrics path with the `__param_<name>` syntax.
+
+For example, the following will call a metrics path of `/health?target_data=prometheus`:
+
+```json
+[
+   {
+      "labels" : {
+         "__metrics_path__" : "/health",
+         "__scheme__" : "https",
+         "__scrape_interval__" : "60s",
+         "__scrape_timeout__" : "10s",
+         "__param_target_data": "prometheus",
+         "service" : "custom-api-service"
+      },
+      "targets" : [
+         "custom-api:443"
+      ]
+   },
+]
+
+```
+
+For more information on the potential labels you can use, see the [prometheus.scrape technical details](../prometheus.scrape#technical-details) section, or the [Prometheus Configuration](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config) documentation.
 
 ## Usage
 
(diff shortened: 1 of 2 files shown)


docs/sources/flow/reference/components/discovery.http.md
```diff
@@ -27,6 +27,26 @@ Example response body:
...
+Example response setting the metrics path and port for [prometheus.scrape](../prometheus.scrape#technical-details) additional config:
+
+```json
+[
+   {
+      "labels" : {
+         "__metrics_path__" : "/api/prometheus",
+         "__scheme__" : "https",
+         "__scrape_interval__" : "60s",
+         "__scrape_timeout__" : "10s",
```
tpaschalis, line 39: I'm leaving this to @clayton-cornell for feedback, but since we're doing this, can we also a) include a `"__param_<name>": "val"` label which will populate a `?name=val` query parameter and b) briefly explain what these parameters do in the above sentence?

proffalken: Happy to tidy up the sentence, didn't realise that you could pass more things to `prometheus.scrape` via labels - is there a full list anywhere?

Main reason for asking is that quite often if `metricsPath` isn't the default, the chances are we'll want to pass a Bearer token or similar as part of the request, but it's not obvious from the docs how to do that on a "per target" basis!

tpaschalis: Yes; the canonical reference is the upstream Prometheus docs which mention all the special "internal" labels that can be used this way.

https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config

tpaschalis: > quite often if metricsPath isn't the default, the chances are we'll want to pass a Bearer token or similar as part of the request, but it's not obvious from the docs how to do that on a "per target" basis!

I think that's not available today. The HTTP client configuration is done on a per-component basis, so you'd need to set up multiple `prometheus.scrape` component to authenticate differently for different targets.

proffalken: >I think that's not available today. The HTTP client configuration is done on a per-component basis, so you'd need to set up multiple prometheus.scrape component to authenticate differently for different targets.

Shame - I may file an enhancement for that shortly as well then! 😄 

I've written a small microservice in Django that I'm running with this on my local network, and I keep all the services that aren't in K8s listed in that service:

image

The output of the service is a JSON array of hosts, metricsPaths, ports, etc. that can be consumed by `discovery.http` but as soon as any of those endpoints need an auth token, I have to add a new scraper with the token.

The output is as follows at present:

```json
[
   {
      "labels" : {
         "__metrics_path__" : "/metrics",
         "__scheme__" : "http",
         "__scrape_interval__" : "60s",
         "__scrape_timeout__" : "10s",
         "service" : "PiHole"
      },
      "targets" : [
         "192.168.12.1:9100"
      ]
   },
   {
      "labels" : {
         "__metrics_path__" : "/api/prometheus",
         "__scheme__" : "https",
         "__scrape_interval__" : "60s",
         "__scrape_timeout__" : "10s",
         "service" : "HomeAssistant"
      },
      "targets" : [
         "192.168.12.10:443"
      ]
   }
]
```

And this works well for `discovery.http` and a generic `prometheus.scrape` apart from the auth tokens.

Ideally, I'd be able to do something like this:

```json
[
   {
      "labels" : {
         "__metrics_path__" : "/metrics",
         "__scheme__" : "http",
         "__scrape_interval__" : "60s",
         "__scrape_timeout__" : "10s",
         "service" : "PiHole",
         "bearer_token": <MY TOKEN>
      },
      "targets" : [
         "192.168.12.1:9100"
      ]
   },
   {
      "labels" : {
         "__metrics_path__" : "/api/prometheus",
         "__scheme__" : "https",
         "__scrape_interval__" : "60s",
         "__scrape_timeout__" : "10s",
         "service" : "HomeAssistant"
      },
      "targets" : [
         "192.168.12.10:443"
      ]
   }
]
```

and for those targets that support it, prom.scrape would pick that up and use it.

Anyway, I'll log a feature request for that in a bit, but that's the context of how I'm using `discovery.http` 😄

tpaschalis: Yes, let's open a separate issue for this!

In the meantime, let's also document `__param_<name>` and use the above sentence to explain how the final target will look/behave like to merge this 🙌

proffalken: @tpaschalis - latest push contains additional content, let me know what needs changing?

tpaschalis (approved):
LGTM, thanks!

docs/sources/flow/reference/components/discovery.http.md
tpaschalis: I hadn't seen there was an additional example below, thanks.

PR #5243, opened 2023-09-19
Flow: Add app_agent_receiver integration components

megumish:
#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/component/all/all.go
+++ b/component/all/all.go
@@ -2,6 +2,7 @@
 package all
 
 import (
+	_ "github.com/grafana/agent/app_agent_receiver"                                 // Import integrations.v2.app_agent_receiver
 	_ "github.com/grafana/agent/component/discovery/aws"                            // Import discovery.aws.ec2 and discovery.aws.lightsail
 	_ "github.com/grafana/agent/component/discovery/azure"                          // Import discovery.azure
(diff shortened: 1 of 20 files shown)


rfratto (commented):
Hi! Thanks for your work on this. I've taken some time to review the original app_agent_receiver code to make sure I understand how it works prior to reviewing the current state of this PR.

I've noticed that the original code relies on some mechanisms specific to the static mode of the agent, particularly the concept of logging/tracing "subsystems." 

As I'm sure you've noticed, that means to wrap around the component like other integrations do, a lot of capabilities required to have equivalents in Flow mode. 

Rather than doing that, I believe it will be simpler to copy/paste the existing integration and modify its code for Flow mode so it's a better fit without needing to introduce all these new config packages. That will also mean there's less overall work to do, and we only need to model the config for app_agent_receiver. 

I will begin working on these refactorings tomorrow.

rfratto:
Because this PR came from `main` and not a branch, we can't push directly to it; I branched off of your main and copied it over to a new PR (#5314) so we can make changes to it. 

I'm going to close this PR in favor of #5314, but we will still be giving you coauthorship over the work.

PR #5246, opened 2023-09-19
OTLP -> Prom otel_scope metrics and labels updated to spec

### LINKED ISSUE #5097: otlp-->prom: Always include otel_scope_name and otel_scope_version
### What's wrong?

We should always include `otel_scope_name` and `otel_scope_version` in all the metrics according to [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#instrumentation-scope-1).

> Prometheus exporters SHOULD generate an [Info](https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#info)-typed metric named otel_scope_info. If present, Instrumentation Scope name and version MUST be added as otel_scope_name and otel_scope_version labels. Scope attributes MUST also be added as labels following the rules described in the [Metric Attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#metric-attributes) section below.

> Prometheus exporters MUST add the scope name as the otel_scope_name label and the scope version as the otel_scope_version label on all metric points by default, based on the scope the original data point was nested in.

> Prometheus exporters SHOULD provide a configuration option to disable the otel_scope_info metric and otel_scope_ labels.

While we disable `otel_scope_info`, we should always include the two labels imo. Otherwise there will be clashes in metrics.

See: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/24248 for how the upstream is fixing it.

### Steps to reproduce

.

### System information

_No response_

### Software version

_No response_

### Configuration

_No response_

### Logs

_No response_

erikbaranowski:
#### PR Description

- rename name and version labels to match otel spec for otel_scope_info metric
- always include otel_scope_name and otel_scope_version labels for otel metrics converted to prometheus

Closes #5097

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


component/otelcol/exporter/prometheus/internal/convert/convert_test.go
```diff
@@ -310,7 +310,7 @@ func TestConverter(t *testing.T) {
 			includeScopeInfo: true,
 			expect: `
 				# TYPE otel_scope_info gauge
-				otel_scope_info{name="a-name",version="a-version",something_extra="zzz-extra-value"} 1.0
+				otel_scope_info{otel_scope_name="a-name",otel_scope_version="a-version",something_extra="zzz-extra-value"} 1.0
```
rfratto, line 313: I think this is a bug fix, can you list it as such?

component/otelcol/exporter/prometheus/internal/convert/convert.go
```diff
@@ -344,18 +344,15 @@ func (conv *Converter) getOrCreateSeries(res *memorySeries, scope *memorySeries,
 		model.MetricNameLabel, name,
 		model.JobLabel, res.metadata[model.JobLabel],
 		model.InstanceLabel, res.metadata[model.InstanceLabel],
+		scopeNameLabel, scope.metadata[scopeNameLabel],
+		scopeVersionLabel, scope.metadata[scopeVersionLabel],
```
rfratto, line 348: We should have a new config option to add these labels rather than always adding them.

component/otelcol/exporter/prometheus/internal/convert/convert_test.go
rfratto: @gouthamve Is this the right set of labels for when the `otel_scope_info` metric _is_ written?

rfratto (commented):
LGTM except whether the scope name and version are written should be optional rather than enforced.

component/otelcol/exporter/prometheus/internal/convert/convert_test.go
gouthamve: This is the right set of labels. 

See example here: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/compatibility/prometheus_and_openmetrics.md#instrumentation-scope

component/otelcol/exporter/prometheus/internal/convert/convert.go
```diff
@@ -61,6 +61,9 @@ type Options struct {
 	// IncludeScopeInfo includes the otel_scope_info metric and adds
 	// otel_scope_name and otel_scope_version labels to data points.
 	IncludeScopeInfo bool
+	// IncludeScopeLabels includes the otel_scope_name and otel_scope_version
+	// labels from the scope in the metrics.
+	IncludeScopeLabels bool
```
rfratto, line 66: Technically this is a breaking change. Previously, scope labels would be added when setting `include_scope_info`. Now, users have to set `include_scope_labels` separately.

Can we document that as a breaking change?

rfratto (approved):
Thanks Erik! LGTM minus my comment and pending docs review from Clayton.

docs/sources/flow/reference/components/otelcol.exporter.prometheus.md
```diff
@@ -48,9 +49,8 @@ OpenTelemetry instrumentation scopes are converted into `otel_scope_info`
 metrics. Set the `include_scope_info` and `include_target_info` arguments to
 `false`, respectively, to disable the custom metrics.
 
-When `include_scope_info` is `true`, the instrumentation scope name and version
-are added as `otel_scope_name` and `otel_scope_version` labels to every
-converted metric sample.
+`include_scope_labels` is used to enable including the `otel_scope_name` and
+`otel_scope_version` labels to every converted metric sample.
```
clayton-cornell, line 53:
```suggestion
When `include_scope_labels` is `true`  the `otel_scope_name` and
`otel_scope_version` labels are added to every converted metric sample.
```

I am guessing that you need to set this new arg to true?

erikbaranowski: ya that sounds better

clayton-cornell (approved):

PR #5249, opened 2023-09-19
Agent Management: Honor 503 ServiceUnavailable Retry-After header

jcreixell:
#### PR Description
  
In addition to `429 TooManyRequests`, remote servers may return `503 ServiceUnavailable` with a `Retry-After` header. This allows overloaded servers to apply back-pressure for a quick recovery.

#### PR Checklist

- [x] CHANGELOG.md updated


rfratto (approved):
Do we anticipate needing to do this for any other codes outside of 429 and 503?

jcreixell:
@rfratto these two should suffice for now IMHO (one of them is client related, the other server related; this should cover most use cases).

PR #5254, opened 2023-09-20
remove/update cortex replace directives

mattdurham:
#### PR Description

Remove the cortex replace directives. Note I did not remove the Prometheus cortex replace since that felt like a larger scope than this. Specifically the 1.2.0 version has a [CVE](https://nvd.nist.gov/vuln/detail/CVE-2021-3538) where the 17b version does not. We were not using this directly and unlikely to be affected but it was picked up by CVE scanners.


tpaschalis (approved):
LGTM, nice catch!

PR #5260, opened 2023-09-20
[Agent Management] Add Handling for HTTP 304 Not Modified Response

spartan0x117:
#### PR Description
Add `notModifiedError`, return response headers from `httpProvider`, and change agent management to handle `notModifiedError` and fall back to cached config.

The agent will store the last `X-AgentNamespaceVersion` and `X-AgentInfoVersion` headers received from the server and send them along with the next request. If the server responds with an `HTTP 304 Not Modified`, the cached copy is used and the headers are saved.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [X] Documentation added
- [X] Tests updated


docs/sources/static/configuration/agent-management.md
```diff
@@ -68,6 +68,9 @@ agent_management:
     # Whether to use labels from the label management service. If enabled, labels from the API supersede the ones configured in the agent.
     label_management_enabled: <bool> | default = false
 
+    # Whether to accept HTTP Not Modified (304) responses from the API server. If enabled, the agent will use the cached configuration if the API server responds with a 304. This can be disabled for debugging or testing, but will save network bandwidth and CPU time if enabled.
+    accept_http_not_modified: <bool> | default = false
```
jcreixell, line 72: any reason not to make this default to `true`?

clayton-cornell: I was going to ask the same. A default to `false` tells me that if a 304 is received, requested resources are sent again instead of redirecting to the cached resources. That seems backwards to me.

docs/sources/static/configuration/agent-management.md
```diff
@@ -68,6 +68,9 @@ agent_management:
     # Whether to use labels from the label management service. If enabled, labels from the API supersede the ones configured in the agent.
     label_management_enabled: <bool> | default = false
 
+    # Whether to accept HTTP Not Modified (304) responses from the API server. If enabled, the agent will use the cached configuration if the API server responds with a 304. This can be disabled for debugging or testing, but will save network bandwidth and CPU time if enabled.
```
clayton-cornell, line 71:
```suggestion
    # Whether to accept HTTP 304 Not Modified responses from the API server. If enabled, the agent will use the cached configuration if the API server responds with HTTP 304 Not Modified. You can set this argument to `false` for debugging or testing.
```

I'm not sure that we need to state `Using the cached configuration saves network bandwidth and CPU time. ` This is implied by the HTTP 304 and redirecting to the cache. Also this statement isn't quantified in any way here.

jcreixell (approved):
LGTM

clayton-cornell (approved):

PR #5261, opened 2023-09-21
Module id fail

### LINKED ISSUE #4702: Module ID registration failure results in permanent error for that Module ID until restart
### What's wrong?

Once a module component fails to build it will continually fail on that ID until the agent is restarted.

### Preperation

1. Clone https://github.com/grafana/agent-modules
2. Create a new file in a new folder `example/metrics/nested/module.river` with the following contents where FILEPATH is the path to any file in your system.

```river
export "content2" {
	value = local.file.testing.content
}

local.file "testing" {
	filename = "FILEPATH"
}
```
3. Setup env vars so you can run https://github.com/grafana/agent-modules/blob/main/example/file.river

### Testing
1. Start the agent with the following config https://github.com/grafana/agent-modules/blob/main/example/file.river

`go run ./cmd/grafana-agent run ${AGENT_CONFIG_FOLDER}/file.river 2>${LOG_FILEPATH}`

2. Confirm the agent is running in the UI and everything is healthy.
3. Add the following to the bottom of `example/metrics/prometheus_receiver/module.river`

```river
module.file "nested" {
	filename = env("AGENT_CONFIG_FOLDER") + "/metrics/nested/module.river2"
	poll_frequency = "30s"
}
```

THE REMAINING STEPS ARE TIME SENSITIVE BASED ON POLL FREQUENCIES SO YOU MAY NEED TO REPEAT STEPS 1-3 IF YOU MISS AN ERROR STATE

4. Keep refreshing the UI. Once `module.file.metrics_prometheus_receiver` polls you should now see it unhealthy with the following error:

`failed to load module content: module.file.nested:37:1: Failed to build component: building component: failed to read file: open /mnt/c/workspace/agent-modules/example/metrics/nested/module.river2: no such file or directory`

5. Keep refreshing the UI. Once `module.file.nested` polls, you should now see it unhealthy with:

`failed to load module content: module.file.metrics_prometheus_receiver:37:1: Failed to build component: building component: id already exists`

6. Update the bottom of `example/metrics/prometheus_receiver/module.river` so it points to the file from Preperation step 2

```river
module.file "nested" {
	filename = env("AGENT_CONFIG_FOLDER") + "/metrics/nested/module.river"
	poll_frequency = "30s"
}
```

7. Keep refreshing the UI. The error message will not go away even though you are pointed at a valid module.
8. Hit the reload endpoint. The error message will not go away.
9. Restart the agent with the updated config from step 6. It is now healthy!

### System information

_No response_

### Software version

v0.35

### Configuration

_No response_

### Logs

```text
ts=2023-08-03T14:50:27.174855274Z level=error controller_id=module.file.metrics_prometheus_receiver trace_id=2b591a953792494ca81c5b014a729971 msg="failed to evaluate config" node=module.file.nested err="building component: id  already exists"
```

mattdurham:
#### PR Description

This handles the use case where a module gets added but with an invalid configuration, lets say a module.file with invalid config. Then gets resolved. Since the first module failed without running, the ID was added to the registry on NewModule but since Run was never called it never gets cleaned up.

Closes #4702

#### Notes to the Reviewer

I am not a huge fan of this solution. I tried an approach to automatically clean it up but since the error happens deep in the loader by the time the error bubbled up to the parent module it was to late.

 I also thought about and scaffolded a cleanup lifecycle step but that felt to large of a change.

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Tests updated


erikbaranowski (approved):
LGTM, nice test case to validate this 💯

rfratto (commented):
I'm worried about this specific implementation too. 

Ideally, the implementation of a component doesn't need to worry about previous iteration of that same component. 

Does the module creation error still happen if ComponentNode is responsible for terminating/cleaning created modules on an evaluation failure or when the component shuts down?

mattdurham:
The root is at https://github.com/grafana/agent/blob/main/pkg/flow/module.go#L48 , that adds the path to the registry/modules which is called from https://github.com/grafana/agent/blob/main/component/module/file/file.go#L64 then Update fails since the file doesn't exist so it returns nil for the component and the code at https://github.com/grafana/agent/blob/main/pkg/flow/module.go#L137C1-L137C1 never gets called to remove it. 

This may also be an outgrowth of Run vs Update debate of what we should do in either.

IMO the cleanest step would be adding a Cleanup lifecycle that is called regardless if Run is called for a component, though that would require always returning a component.

rfratto:
> IMO the cleanest step would be adding a Cleanup lifecycle that is called regardless if Run is called for a component, though that would require always returning a component.

Who would be the caller of this?

In my mind, ComponentNode has the most control over doing this transparently to component implementations:

1. ComponentNode has access to the module controller instance and can remove modules on behalf of a component. 
2. ComponentNode knows when it's constructing a component, and if that construction fails. 
3. ComponentNode knows when a component exits (i.e., when Run exits). 

So, ComponentNode could unregister components using the module controller directly in one of these cases:

1. The initial component construction fails. 
2. The Run method exits. 

Neither of those would require access to the constructed component, so it would still allow for component constructors to return `nil` (which is also important behavior for #4411 to work properly).

mattdurham:
Sounds solid though using the ComponentNode would mean we would need to check for an interface or flag since it would make sense to move the Removal and Addition of module ids to the component node so only one thing is controlling the add/remove id lifecycle and it is one place.

mattdurham:
Though after giving it a second thought using ComponentNode breaks the moment we add multiple modules being loaded.

rfratto:
Can you explain what you mean a little bit more? I'm confused by why multiple modules complicates the situation, and I'm not sure I'm following the "we would need to check for an interface or flag" bit either.

Maybe it's worth having a prototype of that approach so we can get on the same page?

mattdurham:
Will post up some psuedocode later in the week to go over what I am thinking.

pkg/flow/internal/controller/node_component.go
```diff
@@ -254,6 +254,7 @@ func (cn *ComponentNode) Evaluate(scope *vm.Scope) error {
 	case nil:
 		cn.setEvalHealth(component.HealthTypeHealthy, "component evaluated")
 	default:
+		cn.moduleController.ClearModuleIDs()
```
rfratto, line 257: There's a bug here. 

1. A module loader is constructed successfully. 
2. The user updates the module loader with an invalid configuration (i.e., something that will fail to evaluate)
3. Loaded modules are incorrectly cleared. 

I think you only want to do this if the _first_ evaluation fails.

pkg/flow/module.go
```diff
@@ -136,7 +138,6 @@ func (c *module) LoadConfig(config []byte, args map[string]any) error {
 //
 // Run blocks until the provided context is canceled.
 func (c *module) Run(ctx context.Context) {
-	defer c.o.parent.removeID(c.o.ID)
```
rfratto, line 139: IMO I think we should keep this line; it's a good safety precaution to remove modules once they stop running.

pkg/flow/internal/controller/node_component.go
```diff
@@ -312,6 +313,8 @@ func (cn *ComponentNode) Run(ctx context.Context) error {
 	cn.mut.RLock()
 	managed := cn.managed
 	cn.mut.RUnlock()
+	// When finished running clear the module ids.
+	defer cn.moduleController.ClearModuleIDs()
```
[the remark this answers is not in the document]
mattdurham: I might be misunderstanding, if I load this up I can see the modules from the UI. The above bug may cause them to disappear inappropriately though. Each modulecontroller is instanced to a component.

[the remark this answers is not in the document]
rfratto: Ugh, sorry, I confused this with the evaluate function. The other comment is still relevant, but I was wrong on this one. Resolving :)

pkg/flow/internal/controller/node_component.go
mattdurham: I am thinking this might depend on the behavior of the module loader. IE is it an error with the config passed in for example an invalid file path, or is it an error with the content. Which leads me to think going back to the original commit of having the creator of the module know when to remove is the safest path even if it requires more vigilance.

rfratto: The original issue was specifically that modules leak when constructing a component fails, and how that leaks to the next instance of the component. 

I'm personally less worried about leaking modules when a subsequent evaluation fails, especially with the change you've made to Run, since it guarantees that a module registration will never leak into the next instance of the component. 

If a module leaks within the same component between calls to Update, I'd say that's a component bug which should be fixed at the module level; but I think the bug you're fixing here is a controller bug.

rfratto: > I think the bug you're fixing here is a controller bug.

To rephrase this: I think it's wrong for the Flow controller to retain state across two instances of the same component, which you're now fixing.

pkg/flow/internal/controller/node_component.go
```diff
@@ -257,6 +258,13 @@ func (cn *ComponentNode) Evaluate(scope *vm.Scope) error {
 		msg := fmt.Sprintf("component evaluation failed: %s", err)
 		cn.setEvalHealth(component.HealthTypeUnhealthy, msg)
 	}
+	// Check for failure on initial loading, if not then we need to remove any modules it the component may habe created.
+	// In the case where the component is not a module loader this is a noop.
+	cn.moduleFailureCheck.Do(func() {
+		if err != nil {
+			cn.moduleController.ClearModuleIDs()
+		}
+	})
```
rfratto, line 267: There's still a problem here: what if a component fails to be constructed more than once? 

Rather than using a sync.Once, this check could be moved down to line 289 (i.e., if `cn.reg.Build` fails) so you only clear module IDs for components which can't be built, no matter how many times the construction fails.

rfratto:
Given that we found two rounds of bug in the changes to the controller, can we add more behavioral tests to the controller for the issues we've found? At least these cases:

* Module cache must not persist between multiple instances of the same component.
* Module cache must not persist when a component is terminated.  

The controller is one of our critical pieces of code, so it should be harder to introduce bugs than it currently is and warrants caution in cases where coverage is low.

mattdurham:
After reviewing feedback and trying a few different approaches, I really wanted to avoid touching the controller/node_component/loader code. Instead I moved the id check to the Run method. This means we have a slight delay on modules being registered/visible but I havent found an issue with that yet. I also think that is more representative of the running system. 

Duplicate IDs don't happen anymore but left the check in. Duplicate registry metrics trigger first in the loader. This also means the removal and insertion of IDs are within the same code block which seems simpler. Granted a module loader could now queue multiple is ids of the same sort but no loader loads multiple ids at the moment. When we add support for loading multiple ids, that would fall upon the individual module loader itself to ensure that aren't loader. With the registry duplicate being the catch all. Ideally this would be a different error since its not pointing out the exact problem. 

Going to set this as WIP while I poke at it a bit more in the morning.

component/registry.go
```diff
@@ -49,7 +49,7 @@ type Module interface {
 	//
 	// Run blocks until the provided context is canceled. The ID of a module as defined in
 	// ModuleController.NewModule will not be released until Run returns.
-	Run(context.Context)
+	Run(context.Context) error
```
rfratto, line 52: Just as an aside: this would be considered a breaking change to the API, which is re-enforcing the idea for me that everything should be moved to internal for 1.0 until we're ready to start exposing parts of our code as stable APIs.

pkg/flow/module.go
```diff
@@ -135,9 +140,14 @@ func (c *module) LoadConfig(config []byte, args map[string]any) error {
 // will be run until Run is called.
 //
 // Run blocks until the provided context is canceled.
-func (c *module) Run(ctx context.Context) {
-	defer c.o.parent.removeID(c.o.ID)
+func (c *module) Run(ctx context.Context) error {
+	if err := c.o.parent.addModule(c); err != nil {
+		return err
+	}
+	defer c.o.parent.removeModule(c)
```
rfratto, line 147: We should get this merged, but it's standing out as potentially concerning to me that this changes things such that the lifetime of a component and module are now different: a component exists within the controller whenever it's defined in a file, even if it's not running, but a module doesn't exist until it starts running. 

I'm not sure if this will introduce any problems, so let's keep an eye for related issues once this is merged.

pkg/flow/module_fail_test.go
```diff
@@ -0,0 +1,108 @@
...
+		`test.fail.module "int" { 
+content=""
+fail=false
+}`
+	err = t1.updateContent(goodContent)
+	require.NoError(t, err)
+	cnc()
+}
+
+func init() {
```
rfratto, line 52: Should this be in testcomponents instead?

rfratto (approved):
LGTM, some remarks (that don't need to be addressed) and one that needs an answer, but let's get this merged!

rfratto:
> At least these cases:
>
>   Module cache must not persist between multiple instances of the same component.
>   Module cache must not persist when a component is terminated.

I looked through the tests but it wasn't obvious if these cases were being covered. If they're not, can you add tests for them?

pkg/flow/module.go
mattdurham: IMO this feels alright, the component is the parent of module(s), so the component life span should be greater than the modules it controls. Def something to keep an eye on.

mattdurham:
Added additional checks for the two use cases above.

PR #5263, opened 2023-09-21
Fix scraped targets metric

thampiotr:
#### PR Description

* When hashmod sharding was on, the metric was reporting # of targets after the ones that don't belong to local instance are dropped - because the targets were dropped in the relabel component
* When clustering is enabled, the metric will always report all the discovered targets, not just the ones that belong to local instance. Tracking whether targets are evenly balanced across instances is not possible using this metric.

#### PR Checklist

- [x] CHANGELOG.md updated


rfratto (approved):
LGTM, thanks!

tpaschalis (approved):
Good catch, thanks!

PR #5264, opened 2023-09-21
slogadapter: check for log level being enabled

rfratto:
The custom log.Logger implementation wrapping around a slog.Handler must check to see if a level is enabled before logging log lines.


mattdurham (approved):
LGTM

thampiotr (approved):

PR #5274, opened 2023-09-22
Replace weaveworks common by dskit

### LINKED ISSUE #5273: Replace weaveworks/common by grafana/dskit
In https://github.com/grafana/dskit/pull/342 weaveworks/common was moved to dskit.

In the agent we currently use both github.com/weaveworks/common and github.com/grafana/dskit which means that some pkg are duplicated resulting in this error:

2023/09/22 09:38:56 proto: duplicate proto type registered: httpgrpc.HTTPRequest
2023/09/22 09:38:56 proto: duplicate proto type registered: httpgrpc.HTTPResponse
2023/09/22 09:38:56 proto: duplicate proto type registered: httpgrpc.Header

We should get rid of github.com/weaveworks/common

wildum:
#### PR Description

Weaveworks code was copied in dskit so we can get rid of the weaveworks import and keep only dskit.

Fixes #5273 

#### Notes to the Reviewer

- the logging part is a bit different in dskit, I needed to adapt some of the changes in the code. Please pay attention to this part
- the HTTP client was not added in dskit so I copied the file in our repo

#### PR Checklist


tpaschalis:
I've still to look into the code, but I noticed the original dskit PR mentions "replaced the use of prometheus with promauto".

In the past `promauto` has been the cause of panics with duplicate metrics registration (eg, during an agent reload, or when a component went away and then back again). Can we quickly check which metrics the author refers to and whether it might affect us?

wildum:
> I've still to look into the code, but I noticed the original dskit PR mentions "replaced the use of prometheus with promauto".
> 
> In the past `promauto` has been the cause of panics with duplicate metrics registration (eg, during an agent reload, or when a component went away and then back again). Can we quickly check which metrics the author refers to and whether it might affect us?

This is the commit: https://github.com/grafana/dskit/commit/273c830bd215c73d630cbdfa82790e1ba17af55e#diff-b1ac7164e3864ace829b09b1f1ea0e34f82cde3998169273719ea0b98cd49eb3

According to it, the affected metrics are:
- tcp_connections
- tcp_connections_limit
- request_duration_seconds
- request_message_bytes
- response_message_bytes
- inflight_requests

These are server metrics and we are using those through the common component net.server.go (which is used by prometheus and loki in the agent)

component/common/net/config.go
```diff
@@ -1,13 +1,13 @@
-// Package http contains a River serializable definition of the weaveworks weaveworks config in
-// https://github.com/weaveworks/common/blob/master/server/server.go#L62.
+// Package http contains a River serializable definition of the dskit dskit config in
```
rfratto, line 1: Haha, we can probably fix the duplicated word here:

```suggestion
// Package http contains a River serializable definition of the dskit config in
```

rfratto (approved):
:shipit:! 

> In the past promauto has been the cause of panics with duplicate metrics registration (eg, during an agent reload, or when a component went away and then back again). Can we quickly check which metrics the author refers to and whether it might affect us?

I looked through the code; while they do use promauto, they use promauto with a custom registry so we should be fine here.

PR #5280, opened 2023-09-22
Parallelise nodes evaluation

thampiotr:
Proposed implementation of a fix for the case covered by tests in https://github.com/grafana/agent/pull/5278

Before the changes:
- Components that update their exports are added to `controller.Queue`, which allows only one instance of a component at a time to be queued. The order of popping from the queue is random.
- The `controller.Queue` is drained by a `Flow.Run()` - for each node popped from the queue we call `EvaluateDependencies` - all of its direct and indirect incoming nodes are evaluated. 
- The above evaluation temporarily disables the ability to add components to `controller.Queue` when their exports change, since we walk all the node's dependencies anyway and it would generate redundant work. 

After the changes:
- Components that update their exports are (still) added to `controller.Queue`, which allows only one instance of a component at a time to be queued.  This PR changes the order of popping from the queue to match the insertion order.
- The `controller.Queue` is still drained by a `Flow.Run()`, but instead of the previous `EvaluateDependencies` logic we have the following:
  - We process only direct incoming nodes, but we allow them to add themselves to `controller.Queue` when their exports change - this way the update will cover the entire subgraph, but will stop if the exports are not updated.
  - Instead of evaluating nodes synchronously, we add a task to evaluate each node in a worker pool, which does so concurrently. The worker pool also prevents queuing multiple instances of the same node similar to what `controller.Queue` does. 
- The above mentioned worker pool is shared with modules controllers. It also shards the incoming work across the goroutines using a provided key (component ID), so that we don't have one very slow task hog all the goroutines. 
- We add tests which cover scenarios when the current implementation can lead to one slow component blocking other components, even if they do not appear to depend on the slow one. See https://github.com/grafana/agent/pull/5278

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


pkg/flow/internal/controller/node_component.go
```diff
@@ -394,16 +388,6 @@ func (cn *ComponentNode) setExports(e component.Exports) {
...
 
-	if cn.doingEval.Load() {
-		// Optimization edge case: some components supply exports when they're
-		// being evaluated.
-		//
-		// Since components that are being evaluated will always cause their
-		// dependencies to also be evaluated, there's no reason to call
-		// onExportsChange here.
-		return
-	}
```
thampiotr, line 405: This is an important change, which allows us to only evaluate direct incoming nodes in the graph.

pkg/flow/worker/worker_pool.go
```diff
@@ -0,0 +1,112 @@
...
+
+import (
+	"fmt"
+	"hash/fnv"
+	"math/rand"
+	"runtime"
+	"sync"
+)
+
+type Pool interface {
```
wildum, line 11: This worker_pool implementation is very generic, wasnt there an available solution that would fit our use case instead of implementing a custom one? This one for example: https://pkg.go.dev/github.com/gammazero/workerpool ?

thampiotr: This pool is actually not a typical, generic pool, there are some subtle but important differences. For example, the sharding of the tasks by the key or the fact that the queue is a set, allowing only one instance of given key to be queued at a time. Even the fact that we dequeue just before executing a task will make a difference and would break the tests.

wildum: got it, thanks!

pkg/flow/internal/controller/loader.go
```diff
@@ -586,67 +590,81 @@ func (l *Loader) EvaluateDependencies(c *ComponentNode) {
 	defer span.End()
 
 	logger := log.With(l.log, "trace_id", span.SpanContext().TraceID())
-	level.Info(logger).Log("msg", "starting partial graph evaluation")
+	level.Info(logger).Log("msg", "starting partial graph evaluation", "originator", c.NodeID())
```
thampiotr, line 593: This log line probably needs to change, since this is now just queueing nodes for evaluation really.

pkg/flow/module.go
```diff
@@ -171,4 +173,8 @@ type moduleControllerOptions struct {
 	// ServiceMap is a map of services which can be used in the module
 	// controller.
 	ServiceMap controller.ServiceMap
+
+	// WorkerPool is a worker pool that can be used to run tasks asynchronously. A default pool will be created if this
+	// is nil.
+	WorkerPool worker.Pool
```
tpaschalis, line 179: (premature question, still haven't read through the entire thing)
Is this docstring valid here as well? Or is a nil worker pool replaced only in the root controller?

thampiotr: AFAICT, both modules' controllers and the root controller are created via `newController` in `flow.go` which is where the `nil` check is performed. Open to suggestions if there's a problem with this :)

component/module/tests/updates_test.go
```diff
@@ -0,0 +1,162 @@
+package tests
```
thampiotr, line 1: I've created these tests, because I've noticed there was non-trivial logic in the `loader.go` for updating components' exports, but I didn't see tests.

pkg/flow/internal/controller/loader.go
```diff
@@ -564,80 +568,103 @@ func (l *Loader) OriginalGraph() *dag.Graph {
...
+	// Report queue size metric.
+	l.cm.evaluationQueueSize.Set(float64(l.workerPool.QueueSize()))
+}
+
+// concurrentEvalFn returns a function that evaluates a node and updates the cache. This function can be submitted to
+// a worker pool for asynchronous evaluation.
+func (l *Loader) concurrentEvalFn(n dag.Node, spanCtx context.Context, tracer trace.Tracer, enqueueTime time.Time) func() {
+	return func() {
+		start := time.Now()
+		l.cm.componentWaitTime.Observe(time.Since(enqueueTime).Seconds())
```
thampiotr, line 629: Typical wait time will look something like this:
image

The longer waits are very rare and from what I gather, it is because of component being unlucky to be queued on a worker that has another expensive component to evaluate.

pkg/flow/internal/controller/loader.go
```diff
@@ -564,80 +568,103 @@ func (l *Loader) OriginalGraph() *dag.Graph {
...
+		err := l.workerPool.SubmitWithKey(n.NodeID(), l.concurrentEvalFn(n, dependantCtx, tracer, enqueueTime))
+		if err != nil {
+			level.Error(l.log).Log("msg", "failed to submit node for evaluation", "node_id", n.NodeID(), "err", err)
+			span.SetStatus(codes.Error, err.Error())
 		}
+		span.End()
+	}
+
+	// Report queue size metric.
+	l.cm.evaluationQueueSize.Set(float64(l.workerPool.QueueSize()))
```
thampiotr, line 621: Queue size during my tests remained low. Only sometimes on startup it could get higher (e.g. 30), but would drain quickly.

image

pkg/flow/worker/worker_pool.go
```diff
@@ -0,0 +1,124 @@
...
+func NewDefaultWorkerPool() Pool {
+	return NewShardedWorkerPool(runtime.NumCPU(), 100)
+}
+
+// NewShardedWorkerPool creates a new worker pool with the given number of workers and queue size for each worker.
+// The queued tasks are sharded across the workers using a hash of the key. The pool is automatically started and
+// ready to accept work. To prevent resource leak, Stop() must be called when the pool is no longer needed.
+func NewShardedWorkerPool(workersCount int, queueSize int) Pool {
+	if workersCount <= 0 {
+		panic(fmt.Sprintf("workersCount must be positive, got %d", workersCount))
```
ptodev, line 47: Can we return an error instead of panicking?

pkg/flow/worker/worker_pool.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func (w *shardedWorkerPool) QueueSize() int {
+	w.lock.Lock()
+	defer w.lock.Unlock()
+	return len(w.set)
+}
+
+func (w *shardedWorkerPool) Stop() {
+	close(w.quit)
```
ptodev, line 101: Do we want to wait until all the goroutines have stopped?

pkg/flow/worker/worker_pool.go
```diff
@@ -0,0 +1,124 @@
...
+
+type Pool interface {
+	// Stop stops the worker pool. It does not wait to drain any internal queues. It must only be called once.
+	Stop()
+	// Submit submits a function to be executed by the worker pool on a random worker. Error is returned if the pool
+	// is unable to accept extra work.
+	Submit(func()) error
+	// SubmitWithKey submits a function to be executed by the worker pool, ensuring that only one
+	// job with given key can be queued at the time. Adding a job with a key that is already queued is a no-op.
+	// Error is returned if the pool is unable to accept extra work.
```
ptodev, line 19: What consequences does returning an error instead of blocking have? It seems the default queue size is 100. Is it possible that if we have too many components, this returns an error and the graph is not evaluated ok? Is the caller expected to retry in case of an error? It would be nice if the comment includes more information about this.

pkg/flow/internal/controller/loader.go
```diff
@@ -564,80 +568,103 @@ func (l *Loader) OriginalGraph() *dag.Graph {
...
+			return nil
+		})
+	}
 
-	// Make sure we're in-sync with the current exports of c.
-	l.cache.CacheExports(c.ID(), c.Exports())
+	// Submit all dependencies for asynchronous evaluation.
+	// During evaluation, if a node's exports change, Flow will add it to updated nodes queue (controller.Queue) and
+	// the Flow controller will call EvaluateDependencies on it again. This results in a concurrent breadth-first
+	// traversal of the nodes that need to be evaluated.
```
ptodev, line 604: Is there a situation in which the exports change, but we have already scheduled something that depends on those exports to run, and hence there could be a race condition where a downstream component reads a value which is more recent than the one it's supposed to read?

pkg/flow/worker/worker_pool.go
thampiotr: This would cascade to many calling functions, so I would prefer not to do that. This should never happen anyway unless we make a coding error. Note that this is not that unusual as it seems, the standard library does it too, e.g. creating a channel with negative size will panic too.

pkg/flow/worker/worker_pool.go
thampiotr: I don't think this is necessary and our use case does not require this.

pkg/flow/internal/controller/loader.go
thampiotr: Reading more recent value is okay and it could already happen currently, because `component.Queue` also accepts only one instance of a component to be queued at the time. I was able to reproduce some intermediate updated being missed in https://github.com/grafana/agent/pull/5278 - but that is okay as long as a dependency always eventually evaluates after its parent node has updated and the latest value propagates to all the necessary nodes.

thampiotr (commented):
Thanks @ptodev for the comments, I've responded to them now.

tpaschalis (approved):
Nice work! Nothing jumps out immediately after a couple of reads, and given the dogfooding effort, I'm ✅ for merging this.

Let's leave some time for others to chime in and we can move ahead! 🙌

CHANGELOG.md
```diff
@@ -130,6 +130,12 @@ Main (unreleased)
 - Add new `agent_component_dependencies_wait_seconds` histogram metric and a dashboard panel 
   that measures how long components wait to be evaluated after their dependency is updated (@thampiotr)
 
+- Components evaluation is now performed in parallel, reducing the impact of 
+  slow components potentially blocking the entire telemetry pipeline. 
+  The `agent_component_evaluation_seconds` metric now measures evaluation time 
+  of each node separately, instead of all the directly and indirectly 
+  dependant nodes. (@thampiotr)
```
rfratto, line 137: Controller metrics are documented at `docs/flow/monitoring/controller_metrics.md`; can you update that?

pkg/flow/internal/controller/loader.go
```diff
@@ -564,84 +568,106 @@ func (l *Loader) OriginalGraph() *dag.Graph {
...
-			// Skip over the starting component; the starting component passed to
-			// EvaluateDependencies had its exports changed and none of its input
-			// arguments will need re-evaluation.
-			return nil
+		// Submit the node for asynchronous evaluation.
+		err := l.workerPool.SubmitWithKey(n.NodeID(), l.concurrentEvalFn(n, dependantCtx, tracer, parent))
+		if err != nil {
+			// The error typically means that the workerPool queue is full. This could mean we have too many components
+			// and the agent cannot keep up with the evaluation. To degrade gracefully, we log the error and continue.
+			// Some evaluations may be missed.
```
ptodev, line 615: > Some evaluations may be missed.

Is this a problem? Do we want to make sure we update the whole graph occasionally, on a timed tick, just to guarantee that it's updated in full at least once in a while? E.g. is it possible to have a situation where we miss an evaluation, and then the next evaluation happens after a very long time (because no updates ever came on upstream components)?

component/module/tests/updates_test.go
rfratto: Tests for exports might be located somewhere else (due to some historical artifacts about the iteration of modules). Check with @erikbaranowski to see if we can remove/deduplicate some tests now that we're adding these.

operations/agent-flow-mixin/dashboards/controller.libsonnet
```diff
@@ -188,7 +188,7 @@ local filename = 'agent-flow-controller.json';
         } +
         panel.withUnit('ops') +
         panel.withDescription(|||
-          The frequency in which the component graph gets updated.
+          The frequency in which the components gets updated.
```
rfratto, line 191:
```suggestion
          The frequency at which components get updated.
```

pkg/flow/flow.go
```diff
@@ -155,6 +160,11 @@ func newController(o controllerOptions) *Flow {
 		}
 	}
 
+	if workerPool == nil {
+		level.Info(log).Log("msg", "no worker pool provided, creating a default pool")
+		workerPool = worker.NewDefaultWorkerPool()
```
ptodev, line 165: I wonder if we should provide a way for users to configure this worker pool? Maybe if they run very agents with loads of components they'd a want larger pool size?

pkg/flow/internal/controller/loader.go
```diff
@@ -19,7 +20,7 @@ import (
 	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/trace"
 
-	_ "github.com/grafana/agent/pkg/flow/internal/testcomponents" // Include test components
+	_ "github.com/grafana/agent/pkg/util/testcomponents" // Include test components
```
rfratto, line 23: Why did we move this package? 🤔 It doesn't look like anything outside of pkg/flow needs it, right?

pkg/flow/internal/controller/loader.go
```diff
@@ -564,84 +568,106 @@ func (l *Loader) OriginalGraph() *dag.Graph {
...
+		// Submit the node for asynchronous evaluation.
+		err := l.workerPool.SubmitWithKey(n.NodeID(), l.concurrentEvalFn(n, dependantCtx, tracer, parent))
+		if err != nil {
+			// The error typically means that the workerPool queue is full. This could mean we have too many components
+			// and the agent cannot keep up with the evaluation. To degrade gracefully, we log the error and continue.
+			// Some evaluations may be missed.
+			level.Error(l.log).Log("msg", "failed to submit node for evaluation - the agent is likely "+
+				"overloaded and cannot keep up with evaluating components - some updates may be dropped",
+				"node_id", n.NodeID(), "err", err)
+			span.SetStatus(codes.Error, err.Error())
```
rfratto, line 619: nit: should we set the span status to Ok if there's no error? Otherwise the status will only ever be "unknown" or "error" with the way things are now.

pkg/flow/internal/controller/loader.go
```diff
@@ -564,84 +568,106 @@ func (l *Loader) OriginalGraph() *dag.Graph {
...
+		span.End()
+	}
+
+	// Report queue size metric.
+	l.cm.evaluationQueueSize.Set(float64(l.workerPool.QueueSize()))
+}
 
+// concurrentEvalFn returns a function that evaluates a node and updates the cache. This function can be submitted to
+// a worker pool for asynchronous evaluation.
+func (l *Loader) concurrentEvalFn(n dag.Node, spanCtx context.Context, tracer trace.Tracer, parent *ComponentNode) func() {
```
rfratto, line 630: style nit: It feels weird to me for this to return a function, instead of the caller wrapping it to make it async:

```go
// Submit the node for asynchronous evaluation.
err := l.workerPool.SubmitWithKey(n.NodeID(), func() {
  l.concurrentEvalFn(n, dependantCtx, tracer, parent)
})
```

I think it's slightly more idiomatic in Go for the caller to decide whether something is a blocking call or not.

pkg/flow/internal/controller/loader.go
```diff
@@ -564,84 +568,106 @@ func (l *Loader) OriginalGraph() *dag.Graph {
...
+			defer l.mut.RUnlock()
+			err = l.postEvaluate(l.log, n, evalErr)
+
+			// Additional post-evaluation steps necessary for module exports.
 			if exp, ok := n.(*ExportConfigNode); ok {
 				l.cache.CacheModuleExportValue(exp.Label(), exp.Value())
 			}
+			if l.globals.OnExportsChange != nil && l.cache.ExportChangeIndex() != l.moduleExportIndex {
+				l.globals.OnExportsChange(l.cache.CreateModuleExports())
+				l.moduleExportIndex = l.cache.ExportChangeIndex()
```
rfratto, line 661: Is it safe to modify this value with only a read lock?

pkg/flow/internal/worker/worker_pool.go
```diff
@@ -0,0 +1,125 @@
...
+	"fmt"
+	"hash/fnv"
+	"math/rand"
+	"runtime"
+	"sync"
+)
+
+type Pool interface {
+	// Stop stops the worker pool. It does not wait to drain any internal queues. It must only be called once.
+	Stop()
```
rfratto, line 13: As far as I can tell, Stop is never called by the Flow controllers. It may be tricky to solve this one, since currently pools are shared between modules and the root controller. 

But also because there's currently only one Pool, it might not be a big deal that it leaks, since the root controller persists through the lifetime of the process.

pkg/flow/flow.go
```diff
@@ -192,6 +202,7 @@ func newController(o controllerOptions) *Flow {
 					DataPath:          o.DataPath,
 					ID:                id,
 					ServiceMap:        serviceMap.FilterByName(availableServices),
+					WorkerPool:        workerPool,
```
rfratto, line 205: Is it intentional that modules share a worker pool with the root module? We'll have to see if this causes contention down the line as the number of total components grow across many, many modules.

rfratto (approved):
LGTM, great work! I left some comments for you to consider before merging.

ptodev (approved):
LGTM as far as I can tell. Thank you for this great improvement! I left a few comments, but they are minor ones. 

In general, my main concern is around mutability and whether a component could edit an exported value at the same time when another component is reading it. It seems this won't be happening with this implementation as far as I can tell, but I can't say I understand the code in full.

component/module/tests/updates_test.go
thampiotr: Thanks! I've discussed with Erik and we've moved this test to `pkg/flow/module_caching_test.go`, as it mainly focuses on verifying the module-related caching in `loader.go`, especially in presence of concurrent updates.

erikbaranowski: we decided to move these to the same folder as other module tests. I think they have value in testing some additional things and should be kept.

- loading more than just a module config by loading a config plus module config together. `arguments` and `exports` are passed between them
- testing a real module loader `module.string`. this is more a side effect of the testing and maybe module loaders should have their own testing eventually in `component/module`. Either way, it doesn't hurt to test more code along the way.

pkg/flow/internal/controller/loader.go
thampiotr: That was because of the test in `component/module/file` to prevent circular dependency, but it's no longer needed after I've moved the test, so putting the tetcomponents back where they were.

pkg/flow/internal/controller/loader.go
thampiotr: Great question! I do _think_ it was safe, but certainly not obvious and it could generate unnecessary calls to `OnExportsChange` when a stale `moduleExportIndex` was read. So, to improve clarity, I've modified this code to upgrade to a write lock when this update is needed. I will be doing some of the tests again to verify.

pkg/flow/internal/controller/loader.go
thampiotr: Yep, I agree it's more idiomatic and less indentation, so I've updated to it.

pkg/flow/flow.go
thampiotr: Yes, this was intentional. Having a worker pool per module introduces a problem of adequate pool sizing - e.g. one module may have 100s of components, while another just 1. It can also lead to too many goroutines if we have a lot of small modules. 

If we can use a single pool, and we run into the problem with too many modules, it may still be preferred to just expose a config for shared pool's thread count and queue sizes.

rfratto (approved):
LGTM, but not sure we need the WaitDone method

pkg/flow/internal/controller/loader.go
thampiotr: I wouldn't expect the queue to raise to this level normally and I did consider this as a failure mode... however, having thought more about this, I came up with a way to easily retry the evaluation of components that failed to enqueue and I think it's safe to add it. I wrote a test for it as well and you can check it out in this commit: https://github.com/grafana/agent/pull/5280/commits/5d0862d6e94a6914743f3d7873141e0af1dbf2c6

```
ts=2023-10-03T16:25:15.77487Z level=error msg="failed to submit node for evaluation - the agent is likely overloaded and cannot keep up with evaluating components" controller_id="" err="worker queue is full" node_id=testcomponents.passthrough.inc_dep_2 originator_id=testcomponents.count.inc
```

Thanks for questioning this approach :) My only fear is a retry storm which can lead to some kind of deadlock scenario, but that should be pretty obvious from logs.

pkg/flow/flow.go
thampiotr: It is possible. Previously, however, we were running with one thread doing all this work, so I don't think this will be something we hit soon. I would be open to making it configurable in the future, but we may not need it for a while.

pkg/flow/flow.go
```diff
@@ -300,3 +303,8 @@ func (f *Flow) LoadSource(source *Source, args map[string]any) error {
 func (f *Flow) Ready() bool {
 	return f.loadedOnce.Load()
 }
+
+// WaitDone blocks until flow controller's Run method finishes.
+func (f *Flow) WaitDone() {
+	<-f.runFinished
+}
```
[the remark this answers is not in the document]
thampiotr: ah, definitely! it didn't occur to me 🤦 thanks

PR #5282, opened 2023-09-22
evaluate env vars in static converter instead of crashing

erikbaranowski:
#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/converter/internal/staticconvert/staticconvert.go
+++ b/converter/internal/staticconvert/staticconvert.go
@@ -28,8 +28,8 @@ func Convert(in []byte) ([]byte, diag.Diagnostics) {
 	var diags diag.Diagnostics
 
 	fs := flag.NewFlagSet("convert", flag.ExitOnError)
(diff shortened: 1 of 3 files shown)


tpaschalis (approved):
LGTM ⛵ 

Do we want to use this as part of our migration instructions in the future (eg. help/instruct users to set some specific env vars to build the correct configuration), or do you see this just as a bandaid to not crash?

Also, small nit, the `expandArgs` argument on the `loaderFunc` type was a bit confusing, it's always used to expand the env vars instead, might be helpful to rename as well.

erikbaranowski:
@tpaschalis Good call on the loaderFunc argument naming. I updated it everywhere applicable

I intend to wire this up and write the usage docs next and will include this type of information. This PR for people using env vars in their static config so that:
- it doesn't crash the converter
- the env var evaluates and is put into the resulting river config

I think ideally when running the converter, we would translate the static mode env var to a flow env var (ie. `env("ENV_VAR_NAME")`) but this might be good enough for now without having to do major changes in the static mode code to allow for this. This will work entirely as expected when using the converter to run a static mode config in flow mode since the env vars would need to be evaluated anyway.

PR #5283, opened 2023-09-22
Update redis exporter to 1.54.0

### LINKED ISSUE #5106: Update github.com/oliver006/redis_exporter to v1.54.0
An update for `github.com/oliver006/redis_exporter` (version `v1.54.0`) is now available. Version `v1.51.0` is currently in use.

spartan0x117:
#### PR Description
Updates `redis_exporter` from `v1.51.0` --> `v1.54.0`. Among some bugfixes and added metrics for streams, there is a new `disable_exporting_key_values` option which stops string key values from being used as labels for `check_keys` and `check_single_keys` metrics.

Fixes #5106 

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Documentation added
- [X] Tests updated
- [X] Config converters updated


docs/sources/flow/reference/components/prometheus.exporter.redis.md
```diff
@@ -41,6 +41,7 @@ Omitted fields take their default values.
 | `check_single_keys`           | `list(string)` | List of single keys to export value and length/size.                                                                    |            | no       |
 | `check_streams`               | `list(string)` | List of stream-patterns to export info about streams, groups, and consumers to search for with SCAN.                    |            | no       |
 | `check_single_streams`        | `list(string)` | List of single streams to export info about streams, groups, and consumers.                                             |            | no       |
+| `disable_exporting_key_values`| `bool`         | Whether to disable using values of keys stored in Redis as labels when using `check_keys` or `check_single_keys`.       |            | no       |
```
clayton-cornell, line 44:
```suggestion
| `disable_exporting_key_values`| `bool`         | Whether to disable using key values stored in Redis as labels when using `check_keys` or `check_single_keys`.       |            | no       |
```

Does this have a default? `false`? Or no default because it's not required?

Would it be better to approach it from the positive side and have it be:

`export_key_values` and default to `true`? And set the bool to `false` when you want to disable using stored key values?

spartan0x117: It was originally to match the upstream exporter, but updated to be `export_key_values`!

clayton-cornell: Ah OK. Then whichever way makes sense for the implementation :-)

clayton-cornell (commented):
Looks good from a docs standpoint

converter/internal/staticconvert/testdata/integrations.yaml
```diff
@@ -159,6 +159,7 @@ integrations:
   redis_exporter:
     enabled: true
     redis_addr: "redis-2:6379"
+    export_key_values: true
```
erikbaranowski, line 162: If you want to see this show in `integrations.river` set this to false so it isn't the default value.

erikbaranowski (approved):
LGTM, 1 comment on converter tests to check out.

converter/internal/staticconvert/testdata/integrations.yaml
spartan0x117: Good point, updated!

PR #5286, opened 2023-09-23
feat(tracing): inject component ID as instrumentation scope attribute

### LINKED ISSUE #5285: Traces emitted by Flow components should set component ID as a instrumentation scope attribute
### Request

The instrumentation scope of traces emitted by Flow components should always include the ID of the Flow component as an instrumentation scope attribute.

Currently, `grafana_agent.component_id` and `grafana_agent.controller_id` is injected as a span attribute, but this is a bit weird and feels against the normal patterns expected by OpenTelemetry. 

### Use case

More clearly identify which component identifies spans in a way that corresponds to the best practices of OpenTelemetry:

> A logical unit of the application code with which the emitted telemetry can be associated. It is typically the developer’s choice to decide what denotes a reasonable instrumentation scope. The most common approach is to use the [instrumentation library](https://opentelemetry.io/docs/specs/otel/glossary/#instrumentation-library) as the scope, however other scopes are also common, e.g. a module, a package, or a class can be chosen as the instrumentation scope.

hainenber:

 Fixes #5285 

#### PR Checklist

- [x] CHANGELOG.md updated

### DIFF
--- a/pkg/flow/tracing/wrap_tracer.go
+++ b/pkg/flow/tracing/wrap_tracer.go
@@ -2,6 +2,8 @@ package tracing
 
 import (
 	"context"
+	"path/filepath"
+	"strings"
 
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
@@ -41,6 +43,12 @@ type wrappedProvider struct {
 var _ trace.TracerProvider = (*wrappedProvider)(nil)
 
 func (wp *wrappedProvider) Tracer(name string, options ...trace.TracerOption) trace.Tracer {
+	// Inject the component name as instrumentation scope attribute.
+	// This would not have component's exact ID, aligning with OTEL's definition
+	if wp.id != "" {
+		otelComponentName := strings.TrimSuffix(wp.id, filepath.Ext(wp.id))
+		options = append(options, trace.WithInstrumentationAttributes(attribute.String(wp.spanName, otelComponentName)))
+	}
 	innerTracer := wp.inner.Tracer(name, options...)
 	return &wrappedTracer{
 		inner:    innerTracer,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -179,6 +179,8 @@ v0.37.0-rc.1 (2023-10-06)
 - Fix initialization of the RAPL collector for the node_exporter integration
   and the prometheus.exporter.unix component. (@marctc)
 
+- Set instrumentation scope attribute for traces emitted by Flow component. (@hainenber)
+
 ### Other changes
 
 - Use Go 1.21.1 for builds. (@rfratto)


pkg/flow/tracing/wrap_tracer.go
```diff
@@ -41,29 +39,11 @@ type wrappedProvider struct {
...
 
 func (wp *wrappedProvider) Tracer(name string, options ...trace.TracerOption) trace.Tracer {
-	innerTracer := wp.inner.Tracer(name, options...)
-	return &wrappedTracer{
-		inner:    innerTracer,
-		id:       wp.id,
-		spanName: wp.spanName,
+	// Inject the component ID as instrumentation scope attribute.
+	if wp.id != "" {
+		options = append(options, trace.WithInstrumentationAttributes(attribute.String(wp.spanName, wp.id)))
```
ptodev, line 44: Does the attribute include the component name? E.g. if the component is `otelcol.receiver.otlp "default"`, then would the instrumentation scope name be `otelcol.receiver.otlp.default` or `otelcol.receiver.otlp`? I think the convention is to not have the instance name, so then it would be `otelcol.receiver.otlp`. But I'm not sure what the Otel best practices are.

clayton-cornell (commented):
No docs changes included, so I'll step away from this one :-)

pkg/flow/tracing/wrap_tracer.go
rfratto: Changing the attributes feels like a separate discussion to me, IMO. This PR does what #5285 asked for, and I'm wary about increasing the scope.

ptodev: That's ok, I'm happy for it to be merged as it is. Maybe it is even better that way, since it's more specific. If someone wants to narrow down to `otelcol.receiver.otlp.` they could probably use a TraceQL query to do it, and they could probably filter by `otelcol.receiver.` and `otelcol.` too.

rfratto:
I'm not able to see the instrumentation scope in Tempo; trying to figure out why.

rfratto:
I chatted with the Tempo team and apparently Tempo doesn't support instrumentation scope attributes yet, so setting this would be a no-op for Tempo. For now, we have to decide if this attribute belongs more on the span or on the resource (which Tempo would support today).

WDYT @ptodev @hainenber?

ptodev:
> We have to decide if this attribute belongs more on the span or on the resource (which Tempo would support today).

I suppose a span attribute would be more appropriate. As defined in the [spec](https://github.com/open-telemetry/opentelemetry-proto/blob/v0.20.0/opentelemetry/proto/trace/v1/trace.proto#L68), a "resource" is supposed to combine multiple instrumentation scopes. Hence it doesn't seem appropriate to assume that only one instrumentation scope is part of the resource.

I think we should proceed like this:
* Add span attributes with the full componentID/controllerID as we do now.
  * Those attributes will include the component instance name. E.g. they could look like `otelcol.receiver.otlp.default`.
  * We could keep those attributes in the long term, because specifying the component instance makes more sense as a span attribute than as an instrumentation scope.
* Add instrumentation scope which does not include the component instance name
  * Those attributes will not include the component instance name. E.g. they could look like `otelcol.receiver.otlp`. I think this is more in line with [Otel's expectations](https://opentelemetry.io/docs/specs/otel/glossary/#instrumentation-scope).
  * Even though Tempo doesn't support instrumentation scope, I see no harm in adding it. I guess Tempo will just ignore it? And Tempo will presumably support it soon anyway?

In the long run, after Tempo supports instrumentation scope, we could even optimise by deprecating the `grafana_agent.component_id` span attribute and instead having just a span attribute with the component instance name (e.g. `"default"` instead of `"otelcol.receiver.otlp.default"`). Maybe this would speed up some queries?

BTW I wonder whether we should use the [source code attribute conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/attributes.md#source-code-attributes) as well. I cannot think of a clean way to use it. We could have `"code.namespace" = "grafana_agent.component.otelcol.exporter.loadbalancing"`, but I think it's easier just to have a `"grafana_agent.component_id" = "otelcol.receiver.otlp.default"`. That way we also specify that "default" is the component instance name.

hainenber:
hi there, long week for me, hence the long time for reply :)

First of all, thanks @rfratto for going to extended length to reach out Tempo team to test it out. That was awesome!

Basically, I agree with  @ptodev's proposal here. We can keep the span attribute as is and try to add instrumentation scope without the component's instance name (I suppose that means we need to omit the last dot-separated part, e.g. "a.b.c" -> "a.b")

Although Tempo might not support instrumentation scope attribute, other OTEL-compatible tracing backend might already support it and so it could be more than just no-op ;) 

Let me implement and test the proposal. Will keep you posted

hainenber:
Hi folks, I've made a draft implementation as suggested above. PTAL when you have time. Thanks!

pkg/flow/tracing/wrap_tracer.go
```diff
@@ -41,6 +43,12 @@ type wrappedProvider struct {
 var _ trace.TracerProvider = (*wrappedProvider)(nil)
 
 func (wp *wrappedProvider) Tracer(name string, options ...trace.TracerOption) trace.Tracer {
+	// Inject the component name as instrumentation scope attribute.
+	// This would not have component's exact ID, aligning with OTEL's definition
+	if wp.id != "" {
+		otelComponentName := strings.TrimSuffix(wp.id, filepath.Ext(wp.id))
+		options = append(options, trace.WithInstrumentationAttributes(attribute.String(wp.spanName, otelComponentName)))
+	}
```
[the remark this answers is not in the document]
hainenber: Yes, the old [function](https://github.com/grafana/agent/blob/6f30a7dd816ea7cd364a56b8bdb1666869a5feac/pkg/flow/tracing/wrap_tracer.go#L68-L77) isn't removed, yet ;)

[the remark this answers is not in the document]
rfratto: Cool, just double checking :)

rfratto (approved):
Thanks for your work!

Issue #4957, opened 2023-08-28
ui: incorect peer dependencies


flokli:
### What's wrong?

I see error messages during build, yarn complains about incorrect peer dependencies.

It seems this project wants a react 17-ish, while the used `@grafana/ui` still wants a react 16-ish. Same applies for `react-dom`.

### Steps to reproduce

Invoke `yarn install` from `web/ui`

### System information

Linux, x86_64

### Software version

main branch

### Logs

```text
❯ yarn install
yarn install v1.22.19
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
warning " > @grafana/ui@9.5.2" has incorrect peer dependency "react@^16.8.0 || ^17.0.0".
warning " > @grafana/ui@9.5.2" has incorrect peer dependency "react-dom@^16.8.0 || ^17.0.0".
warning "@grafana/ui > @grafana/data@9.5.2" has incorrect peer dependency "react@^16.8.0 || ^17.0.0".
warning "@grafana/ui > @grafana/data@9.5.2" has incorrect peer dependency "react-dom@^16.8.0 || ^17.0.0".
warning "@grafana/ui > react-hook-form@7.5.3" has incorrect peer dependency "react@^16.8.0 || ^17".
warning "eslint-config-react-app > eslint-plugin-flowtype@8.0.3" has unmet peer dependency "@babel/plugin-syntax-flow@^7.14.5".
warning "eslint-config-react-app > eslint-plugin-flowtype@8.0.3" has unmet peer dependency "@babel/plugin-transform-react-jsx@^7.14.9".
[4/4] Building fresh packages...
Done in 33.13s.
```
```

tpaschalis:
I'm closing this one since after updating our dependencies we no longer get messages about incorrect or unmet peer dependencies.

PR #5287, opened 2023-09-23
feat(build): auto-generate PRs to bump agent's web UI dependencies

hainenber:
#### PR Description
JS dependencies in agent's UI are not kept up-to-date and can cause issue like #4957. We should have a way to automate version upgrade for the agent's UI dependencies.


tpaschalis (approved):
It looks alright at a glance, let's get this in and see how it works.

PR #5297, opened 2023-09-26
ensure syslog parser gets an EOF-terminated reader on udp receive

### LINKED ISSUE #5197: loki.source.syslog udp protocol not working
### What's wrong?

I set up a simple docker compose with minio, loki, grafana and grafana-agent with the syslog source for the protocols udp and tcp to ingest some firewall-logs. tcp is working fine, but over udp I never get any logs ingested into loki. I cannot see, what should be wrong with the config in this case. As a workaround, I added an rsyslog as a forwarder from udp to the tcp port of grafana agent and that's working, but the solution would be simpler, if the udp input of grafana agent would just be working as documented.

### Steps to reproduce

**Docker compose:**

```
version: '3.8'
services:
  grafana:
    container_name: grafana
    environment:
      GF_AUTH_ANONYMOUS_ENABLED: "true"
      GF_AUTH_ANONYMOUS_ORG_ROLE: Admin
      GF_AUTH_BASIC_ENABLED: "false"
      GF_AUTH_DISABLE_LOGIN_FORM: "true"
      GF_DASHBOARDS_DEFAULT_HOME_DASHBOARD_PATH: /var/lib/grafana/dashboards/no_folder/loki_syslog_aio_overview.json
    image: grafana/grafana:latest
    depends_on:
      - loki
    ports:
      - 3000:3000
    restart: unless-stopped
    volumes:
      - ./config/grafana/provisioning/datasources:/etc/grafana/provisioning/datasources:ro
      - ./config/grafana/provisioning/dashboards:/etc/grafana/provisioning/dashboards:ro
      - ./config/grafana/dashboards:/var/lib/grafana/dashboards:ro
      - /etc/localtime:/etc/localtime:ro
      - /etc/timezone:/etc/timezone:ro
  minio:
    command: -c 'mkdir -p /data/loki && /usr/bin/docker-entrypoint.sh minio server /data'
    volumes:
      - /data:/data
      - /etc/localtime:/etc/localtime:ro
      - /etc/timezone:/etc/timezone:ro
    container_name: minio
    entrypoint: sh
    environment:
      MINIO_ROOT_USER: 'minio-user'
      MINIO_ROOT_PASSWORD: 'minio-pass'
      MINIO_ADDRESS: ':9000'
      MINIO_CONSOLE_ADDRESS: ':9001'
      MINIO_ACCESS_KEY: minio123
      MINIO_PROMETHEUS_AUTH_TYPE: public
      MINIO_SECRET_KEY: minio456
    healthcheck:
      interval: 30s
      retries: 3
      test:
        - CMD
        - curl
        - -f
        - http://localhost:9000/minio/health/live
      timeout: 20s
    image: quay.io/minio/minio:latest
    ports:
      - 9000:9000
      - 9001:9001
    restart: unless-stopped
  loki:
    command: -config.file=/etc/loki/loki-config.yml
    container_name: loki
    depends_on:
      - minio
    image: grafana/loki:latest
    ports:
      - 3100:3100
    restart: unless-stopped
    volumes:
      - ./config/loki-config.yml:/etc/loki/loki-config.yml:ro
      - /etc/localtime:/etc/localtime:ro
      - /etc/timezone:/etc/timezone:ro
  agent:
    command: run --server.http.listen-addr=0.0.0.0:12345 /etc/agent/config.river
    container_name: agent
    environment:
      AGENT_MODE: 'flow'
    depends_on:
      - loki
    image: grafana/agent:latest
    ports:
      - 12345:12345
      - 10514:10514/tcp
      - 10514:10514/udp
    restart: unless-stopped
    volumes:
      - ./config/config.river:/etc/agent/config.river:ro
      - /etc/localtime:/etc/localtime:ro
      - /etc/timezone:/etc/timezone:ro
```

### System information

Ubuntu 22.04.3 LTS

### Software version

Grafana Agent v0.36.1

### Configuration

```text
logging {
  level  = "info"
  format = "logfmt"
}

loki.source.syslog "syslogudp" {
  listener {
    address  = "0.0.0.0:10514"
    protocol = "udp"
    labels   = { component = "loki.source.syslog", protocol = "udp"}
  }

  forward_to = [loki.write.local.receiver]
}

loki.source.syslog "syslogtcp" {
  listener {
    address  = "0.0.0.0:10514"
    protocol = "tcp"
    labels   = { component = "loki.source.syslog", protocol = "tcp" }
  }

  forward_to = [loki.write.local.receiver]
}

loki.write "local" {
  endpoint {
    url = "http://loki:3100/loki/api/v1/push"
  }
}
```

### Logs

_No response_

joshuapare:
#### PR Description

Simple fix to convert the incoming datagram into a bytes reader that the ParseStream can handle. currently it freezes up because the datagrams are sent as is without separation so the reader routine blocks until an interrupt.

Fixes #5197

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Tests updated


tpaschalis:
cc @erikbaranowski do I correctly remember you'd investigated this issue in the past? Could you take a look here?

erikbaranowski:
This looks good. @joshuapare can we add a test to syslogtarget_test.go to demonstrate the scenario that is now working with this fix?

component/loki/source/syslog/internal/syslogtarget/transport.go
```diff
@@ -430,16 +431,32 @@ func (t *UDPTransport) handleRcv(c *ConnPipe) {
 	defer t.openConnections.Done()
 
 	lbs := t.connectionLabels(c.addr.String())
-	err := syslogparser.ParseStream(c, func(result *syslog.Result) {
-		if err := result.Error; err != nil {
-			t.handleMessageError(err)
-		} else {
-			t.handleMessage(lbs.Copy(), result.Message)
+
+	for {
```
rfratto, line 435: Thanks for the fix! Can you add a comment here so we don't accidentally revert this code? We've been gradually making sure we've synchronized our code with Promtail, so this is at risk of being reverted if we don't add a warning like this.

```suggestion
	// NOTE: The below code purposefully desyncs from Promtail to fix a bug where 
	// UDP messages were never processed. See grafana/agent#5197 for more context.
	for {
```

joshuapare: @rfratto I actually made a simultaneous PR over to the loki repo, since after investigating it appeared to be in both places for the same reason. The next sync should bring that change over, given it's merged in (once I add those tests @erikbaranowski requested) https://github.com/grafana/loki/pull/10708 (though appears the build is failing due to an unrelated test, I can investigate once I finish these)

rfratto: Ah, gotcha, thanks!

rfratto:
@joshuapare Do you still have time to work on this? It looks like all we need is a test for the fix, and then this is good to be merged :) 

> This looks good. @joshuapare can we add a test to syslogtarget_test.go to demonstrate the scenario that is now working with this fix?

rfratto (approved):
I looked through the existing code, and I think there's enough tests to prove that this doesn't break anything, even if we don't have a test for testing this specific case, I think we can just merge this for now. If it breaks again we'll want a regression test for sure.

erikbaranowski (approved):
seconded @rfratto. let's get this fix out and give it a try.

PR #5299, opened 2023-09-26
Fix bug in metric reporting

thampiotr:
#### PR Description

In https://github.com/grafana/agent/pull/5263/files I have used `len` on a wrong structure, resulting in the targets gauge being always 1.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -122,6 +122,9 @@ Main (unreleased)
 
 - Fixed a bug where converting `YACE` cloudwatch config to river skipped converting static jobs. (@berler)
 
+- Fixed the `agent_prometheus_scrape_targets_gauge` incorrectly reporting all discovered targets
+  instead of targets that belong to current instance when clustering is enabled. (@thampiotr)
+
 ### Other changes
 
 - Use Go 1.21.1 for builds. (@rfratto)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -162,9 +165,6 @@ v0.36.2 (2023-09-22)
 
 - Fix `loki.source.file` race condition in cleaning up metrics when stopping to tail files. (@thampiotr)
 
-- Fixed the `agent_prometheus_scrape_targets_gauge` incorrectly reporting all discovered targets
-  instead of targets that belong to current instance when clustering is enabled. (@thampiotr)
-
```
thampiotr, line 167: I think it's best to retract this statement from `0.36.2` - the bugfix was incorrect, so the metric is still not correct.

CHANGELOG.md
```diff
@@ -122,6 +122,9 @@ Main (unreleased)
 
 - Fixed a bug where converting `YACE` cloudwatch config to river skipped converting static jobs. (@berler)
 
+- Fixed the `agent_prometheus_scrape_targets_gauge` incorrectly reporting all discovered targets
```
tpaschalis, line 125: I think the bug wasn't included in the previous release (right?), so we may not need the entry here.

thampiotr: The old bug is present in all previous versions (but only affecting users when they have clustering on). 
The new bug I've added in https://github.com/grafana/agent/pull/5263 is present in `0.36.2` unfortunately.

tpaschalis (approved):
LGTM, good catch.

Issue #3162, opened 2023-03-02
Flow component for cadvisor integration


rfratto:
Bring over the static mode cadvisor integration as a Flow component.

This one needs caution because it's currently a singleton integration in static mode, and needs changes to be brought over as not-a-singleton in Flow. See #1861.

PR #5300, opened 2023-09-26
pkg/integrations/cadvisor: minimize runner responsibilities

tpaschalis:
#### PR Description
The current iteration of the cadvisor integration is not compatible with the way we're building `prometheus.exporter.*` components today.

The current code gets the HTTP handler for the metrics endpoint once at the start (after the exporter is created), and then calls to run the exporter itself.

https://github.com/grafana/agent/blob/363b169217812dec56dc1f2e9f27998fc7e07644/component/prometheus/exporter/exporter.go#L73-L80

The cadvisor integration behaves somewhat differently and updates its collectors within its custom `Run` method

https://github.com/grafana/agent/blob/363b169217812dec56dc1f2e9f27998fc7e07644/pkg/integrations/cadvisor/cadvisor.go#L148-L150

This means at the point where the component would fetch the handler, no collectors are registered and no metrics are available.

This PR serves to minimize the runner's responsibilities by only watching for the shutdown of the context to close the manager. As a tradeoff, it will spend some more time in the integration's initialization.

No issue filed. Updates https://github.com/grafana/agent/issues/3162

#### Notes to the Reviewer

I've diffed the metrics produced by the `cadvisor` integration on main and the current branch, and diffed them with a Python script created by @spartan0x117. As expected, there's no difference in the metrics being reported.

```
$ wc -l forked_metrics main_metrics
   2274 forked_metrics
   2274 main_metrics
   4548 total
$ python3 mischa.py
Found 1 keys not in main_metrics
Found 1 keys not in forked_metrics
Total: 2 keys not present both files
---------------------------------
Keys not in first file:
cadvisor_build_info{branch="refactor-cadvisor-integration",goarch="amd64",goos="linux",goversion="go1.21.1",revision="97ea39a0b",tags="unknown",version="refactor-cadvisor-integration-97ea39a"}
---------------------------------
Keys not in second file:
cadvisor_build_info{branch="main",goarch="amd64",goos="linux",goversion="go1.21.1",revision="363b16921",tags="unknown",version="main-363b169"}
```

I haven't performed more extensive testing, but the behavior should not be changing here.

#### PR Checklist

- [ ] Tests updated (N/A)

### DIFF
--- a/pkg/integrations/cadvisor/cadvisor.go
+++ b/pkg/integrations/cadvisor/cadvisor.go
@@ -116,12 +116,12 @@ func New(logger log.Logger, c *Config) (integrations.Integration, error) {
 
 	var collectorHTTPClient http.Client
 
-	cludedMetrics, err := c.GetIncludedMetrics()
+	includedMetrics, err := c.GetIncludedMetrics()
 	if err != nil {
-		return nil, fmt.Errorf("unable to determine cluded metrics: %w", err)
+		return nil, fmt.Errorf("unable to determine included metrics: %w", err)
 	}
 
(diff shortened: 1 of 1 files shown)


jcreixell (approved):
LGTM

PR #5311, opened 2023-09-26
`prometheus.operator.*` add debug endpoint to view generated scrape configs.

captncraig:
This just makes the operator components a little easier to debug when we have concerns about relabel rules or things like that.

- Adds a handler for $COMPONENT/scrapeConfig/$NS/$NAME that will serve the generated scrape config in yaml form.
- Add link to that url from UI debug info for each monitor crd.

The handler itself is pretty gross since there is not much routing or anything, I didn't want to pull in too many libraries or anything. But this is not how I'd want to do this if we scaled the concept to more complex endpoints.


spartan0x117 (approved):
LGTM!

PR #5325, opened 2023-09-28
docs: add page about deployment topologies

tpaschalis:
This PR adds a documentation page about common Grafana Agent deployment topologies, their Kubernetes equivalent, high-level pros and cons and when users should consider using each.

Closes #4951

PR #5327, opened 2023-09-28
docs: discussion of different kubernetes controller options.

captncraig:
#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


docs/sources/flow/getting-started/controllers.md
```diff
@@ -0,0 +1,122 @@
...
+    role = "pod"
+    selectors {
+        role = "pod"
+        field = "spec.nodeName=" + constants.hostname
+    }
+}
+discovery.kubernetes "endpoints" {
+    role = "endpoint"
+    selectors {
+        role = "pod"
```
spartan0x117, line 50: This is already documented in the reference docs, so not sure if it's worth calling out again here, but I imagine the two `role` arguments could be a bit confusing to newer folks?

Maybe a sentence like:
```markdown
The top level `role` argument specifies the type of Kubernetes resource to query, and the selectors' `role` specifies the role of the selector.
```

docs/sources/flow/getting-started/controllers.md
```diff
@@ -0,0 +1,122 @@
...
+## DaemonSet
+
+A `DaemonSet` will deploy a Grafana Agent pod to each Node in your Kubernetes cluster. This is needed when you need access to something that is present on every Node in your cluster, such as:
+
+- Local access to log files.
+- Exporters, such as `node_exporter`, which run on every physical node.
+
+There are several downsides to a `DaemonSet` which need to be considered:
+
+- You cannot attach Persistent Volumes to `DaemonSet` pods. This means things like the Write-Ahead Log will use the Node's local disk, which may not be desired in shared environments.
```
spartan0x117, line 29: Depending on the audience's expected familiarity, it might be good to note an example of why it wouldn't be desired (e.g. noisy neighbor; if any other pods on the node really need the disk there could be contention, slow, etc...)?

tpaschalis:
This seems to cover _some_ common ground with #5325.

Should we link the two documents or even merge them into one? cc @clayton-cornell

clayton-cornell:
> This seems to cover _some_ common ground with #5325.
> Should we link the two documents or even merge them into one? 

Looking at the topics as they are in Draft state... I'd keep the separate for now.. but limit #5325 to defining the concepts/pros/cons and use this one to detail how to do things?  That's kind of how it is now... with this PR containing the River configs. Better not to cram everything into a one-size-fits-all topic :-)

captncraig:
Too much overlap with existing deploy page. I will propose additions on top of that, as well as the "Distribute Prometheus metrics scrape load
 pages.

PR #5333, opened 2023-09-28
Templatize version updates

spartan0x117:
#### PR Description
Adds templates of all of the files that reference the latest release directly by version and adds a script to update and generate files with the version injected in the right place.

A few alternative approaches were considered:
1. Using `gomplate`. It is possible to switch to it, if desired, but requires a decent amount of working around existing elements (e.g. things like `{{< relref "./_index.md" >}}`).
2. Using `envsubst`. This has conflicts with a few places, e.g. https://github.com/grafana/agent/blob/548259eceb08efc1de3c4c9598f5ad4dae912692/production/kubernetes/agent-loki.yaml#L5, leading to generated files that have these values replaced with empty strings, which we don't want.

If the `.t` files are not desired, an alternative would be to add the templates all within the `tools` directory and maintain a mapping of where they need to be added. The downside of this approach is that it is more hidden and likely to desync (e.g. if we move a directory to `internal/`, if the template exists with the generated file we don't need to do anything. Otherwise, the output mapping needs to be updated or it will fail the next time someone runs it or put something in the wrong place).

#### Note(s) to the reviewer
The `.t` files are copies of the corresponding file. Going forward, only these should be modified so the generated files are correct. 

The only added code is in `tools/generate-version-files.sh`

#### PR Checklist

- [X] Documentation added


mattdurham:
This is fantastic, thoughts on adding a test to ensure its all happy?

tpaschalis:
Thanks for taking up on this initiative, it's nice to finally not have to do this manually! 🎉 

One thing I'm worried about is that people might accidentally only edit the non-template file, either due to force of habit or going to our documentation page and hitting the "Suggest an edit" button.

How hard do you think it would be to add a Github Action to guard against this?

We could reuse the `./tools/generate-version-files.sh` script by either running it and verifying that nothing changed with a Git command, or make it return a special status code that could tell us if some templating took place or not.

spartan0x117:
Well, the new workflow did correctly find that some of the non-template files were updated in main between me creating the branch and now :sweat_smile: 
```
-looking to use. To learn more about the custom resources Agent Operator provides and their hierarchy, see [Grafana Agent Operator architecture]({{< relref "./architecture" >}}).
+looking to use. To learn more about the custom resources Agent Operator provides and their hierarchy, see [Grafana Agent Operator architecture]({{< relref "./architecture/" >}})
```

spartan0x117:
I update the template files to be of the form `path/to/file.t.ext` rather than `path/to/file.ext.t` because it breaks any syntax highlighting (_technically_ fixable in most text editors by adding custom checks, but I'd rather not force folks to do that) and therefore makes editing the templates more painful. 

This way editing templates should be roughly as easy as before, except perhaps in cases where `$AGENT_VERSION` doesn't play nicely :sweat_smile:

spartan0x117:
> I update the template files to be of the form `path/to/file.t.ext` rather than `path/to/file.ext.t` because it breaks any syntax highlighting (_technically_ fixable in most text editors by adding custom checks, but I'd rather not force folks to do that) and therefore makes editing the templates more painful.
> 
> This way editing templates should be roughly as easy as before, except perhaps in cases where `$AGENT_VERSION` doesn't play nicely 😅

Turns out this approach may not be a good idea, since other tools may pick up those files (for example `pkg/operator/defaults.t.go` doesn't get along with the generated `pkg/operator/defaults.go`), so I reverted the commit.

jdbaldry:
For docs, I think we can avoid templating every file that has reference to the version by setting the version as cascading front matter in `docs/sources/_index.md`. 

By adding the front matter:

```yaml
cascade:
  AGENT_RELEASE: $AGENT_VERSION
```

Any child page can access that parameter at build time with the [Hugo `param` shortcode](https://gohugo.io/content-management/shortcodes/#param), for example: `The current Grafana Agent release version is {{< param "AGENT_RELEASE" >}}`.

In this case we only template that one file and let any other page use the variable wherever they need.

spartan0x117:
@jdbaldry That's perfect! Added it and it works like a charm for all of the markdown docs :smiley:

jdbaldry:
Thanks for making those changes!

Release information is super handy to have available to the pages and minimizing the number of pages that actually have to be fully templated is also nice.

I'll let the rest of the maintainers pick up the PR for full review but I'm happy with the docs side of things for sure.

clayton-cornell (approved):
The docs changes look great. I hope this reduces the effort on the run-up to each release now. Changing one place and cascading... is really useful.

Approving doc changes

production/kubernetes/agent-bare.yaml.t
```diff
@@ -0,0 +1,115 @@
+apiVersion: v1
```
rfratto, line 1: The files in production/kubernetes/*.yaml are generated from Jsonnet, so we shouldn't template these, but rather the Jsonnet that they're generated from :)

production/kubernetes/install-bare.sh.t
```diff
@@ -0,0 +1,34 @@
...
+check_installed curl
+check_installed envsubst
+
+MANIFEST_BRANCH=$AGENT_VERSION
+MANIFEST_URL=${MANIFEST_URL:-https://raw.githubusercontent.com/grafana/agent/${MANIFEST_BRANCH}/production/kubernetes/agent-bare.yaml}
+NAMESPACE=${NAMESPACE:-default}
+
+export NAMESPACE
+
+curl -fsSL "$MANIFEST_URL" | envsubst
```
rfratto, line 34: It would be good for us to remove the production/kubernetes folder entirely, honestly. We don't maintain it anymore. I only see one remaining reference to it that uses a recent version: 

https://grafana.com/docs/agent/latest/static/set-up/install/install-agent-kubernetes/ 

This guide can be changed to use our Helm chart instead for static mode, with some opinionated values.yaml files.

cc @clayton-cornell for a future goal

production/operator/templates/agent-operator.yaml.t
```diff
@@ -0,0 +1,645 @@
+apiVersion: v1
```
rfratto, line 1: Ditto with this file: we may be ale to delete it in favor of the operator Helm chart; I'm not sure where this is used at all, actually. I can't find any references to it in the website or in this repo, so it might be safe to remove it.

production/tanka/grafana-agent/v1/main.libsonnet.t
```diff
@@ -0,0 +1,142 @@
+local agent = import './internal/agent.libsonnet';
```
rfratto, line 1: Cleanup task: we can remove the production/tanka folder completely once we remove production/kubernetes

rfratto (approved):
LGTM! There's a few files which shouldn't be templated, since they're already generated files. 

I also found some instances where we can reduce the number of templated files by cleaning up our documentation and removing stale references to things. That would be helpful long term. 

Either way, not blocking this, it's a good improvement!

production/kubernetes/install-bare.sh.t
clayton-cornell: Opened https://github.com/grafana/agent/issues/5380 so this doesn't get lost.

PR #5334, opened 2023-09-29
Set MinBackoff for Tailer config to one

### LINKED ISSUE #5322: Agent tailer backoff wait time always 0
### What's wrong?

When encountering containers which are not started, for example containers in `ImagePullBackoff` or `CreateContainerConfigError`, grafana-agent logs messages like these almost every second for every failing container:

```
ts=2023-09-28T07:33:52.129111985Z level=warn target=namespace/container_name:pod_id component=module.file.loki_config/loki.source.podlogs.default msg="tailer stopped; will retry" err="container \"container_name\" in pod \"pod_id\" is waiting to start: CreateContainerConfigError"
```

The log itself is not a problem, as it is expected. However, it seems that there is a bug in the code for producing exponential back-off times in `component/loki/source/kubernetes/kubetail/tailer.go`.

StartConfig:
```go
var retailBackoff = backoff.Config{
	MinBackoff: 0,
	MaxBackoff: time.Minute,
}
```

Creating the backoff at L#97:
```go
bo := backoff.New(ctx, retailBackoff)
```

When calculating `nextDelayMin` and `nextDelayMax` in the `backoff` package, they are set to the same value, 0.  This causes the calculation of the next delay of the backoff to be 0 as well:
```go
if b.nextDelayMin >= b.nextDelayMax {
     return b.nextDelayMin
}
```

Presumably, the MinBackoff has to be set to some other value than 0, or the `backoff` package should handle `nextDelayMax/Min == 0` in a different manner. 

### Steps to reproduce

Have a failing container which is being monitored by the grafana agent.

### System information

Darwin ARM64

### Software version

grafana-agent helm chart v0.21.0

### Configuration

_No response_

### Logs

```text
ts=2023-09-28T07:33:52.129111985Z level=warn target=namespace/container_name:pod_id component=module.file.loki_config/loki.source.podlogs.default msg="tailer stopped; will retry" err="container \"container_name\" in pod \"pod_id\" is waiting to start: CreateContainerConfigError"
```

anderssonw:

Fixes #5322 

#### Notes to the Reviewer

Certainly should be looked at from the upstream library's perspective

#### PR Checklist

- [x] CHANGELOG.md updated

### DIFF
--- a/component/loki/source/kubernetes/kubetail/tailer.go
+++ b/component/loki/source/kubernetes/kubetail/tailer.go
@@ -81,9 +81,9 @@ func newLabelSet(l labels.Labels) model.LabelSet {
 var retailBackoff = backoff.Config{
 	// Since our tailers have a maximum lifetime and are expected to regularly
 	// terminate to refresh their connection to the Kubernetes API, the minimum
-	// backoff starts at zero so there's minimum delay between expected
+	// backoff starts at one so there's minimum delay between expected
 	// terminations.
-	MinBackoff: 0,
+	MinBackoff: 1,
 	MaxBackoff: time.Minute,
 }
 
--- a/component/loki/source/kubernetes/kubetail/tailer.go
+++ b/component/loki/source/kubernetes/kubetail/tailer.go
@@ -81,9 +81,9 @@ func newLabelSet(l labels.Labels) model.LabelSet {
 var retailBackoff = backoff.Config{
 	// Since our tailers have a maximum lifetime and are expected to regularly
 	// terminate to refresh their connection to the Kubernetes API, the minimum
-	// backoff starts at one so there's minimum delay between expected
+	// backoff starts at 10ms so there's a small delay between expected
 	// terminations.
-	MinBackoff: 1,
+	MinBackoff: 10 * time.Millisecond,
 	MaxBackoff: time.Minute,
 }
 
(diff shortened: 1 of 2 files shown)


component/loki/source/kubernetes/kubetail/tailer.go
```diff
@@ -81,9 +81,9 @@ func newLabelSet(l labels.Labels) model.LabelSet {
 var retailBackoff = backoff.Config{
 	// Since our tailers have a maximum lifetime and are expected to regularly
 	// terminate to refresh their connection to the Kubernetes API, the minimum
-	// backoff starts at zero so there's minimum delay between expected
+	// backoff starts at one so there's minimum delay between expected
```
tpaschalis, line 84:
```suggestion
	// backoff starts at 1s so there's a small delay between expected
```

component/loki/source/kubernetes/kubetail/tailer.go
```diff
@@ -81,9 +81,9 @@ func newLabelSet(l labels.Labels) model.LabelSet {
...
 	// terminations.
-	MinBackoff: 0,
+	MinBackoff: 1,
```
tpaschalis, line 86: The golang `time.Duration` value corresponds to nanosecond int64 values, so we should explicitly define the unit as well.

Also, after reading through the code could we please make this a bit smaller to avoid missing short-lived pods?

```suggestion
	MinBackoff: 10 * time.Millisecond,
```

anderssonw:
Makes sense, @tpaschalis :)

Also added to the changelog that I forgot to push

tpaschalis (approved):
LGTM, thanks!

PR #5335, opened 2023-09-29
prometheus.scrape: add options for enabling native histograms scraping

### LINKED ISSUE #5136: Prometheus native histograms are not being scraped
It looks like the Agent does not scrape Prometheus' native histograms. We should be setting the EnableProtobufNegotiation; otherwise we only ever scrape the non-gRPC /metrics endpoint, so we never get the native histograms payload.

The Agent should be able to scrape both the HTTP and gRPC endpoints (in case the remote target supports it) and receive data from both histogram types at the same time.

tpaschalis:
#### PR Description
This PR adds a couple of new Arguments to the `prometheus.scrape` component to toggle scraping of Prometheus' native histograms over gRPC and conditionally allowing classic histograms being scraped in parallel.

Fixes #5136 

#### Notes to the Reviewer
Nothing for now.

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,8 @@ Main (unreleased)
   This will load all River files in the directory as a single configuration;
   component names must be unique across all loaded files. (@rfratto, @hainenber)
 
+- Flow: the `prometheus.scrape` component can now configure the scraping of
+  Prometheus native histograms. (@tpaschalis)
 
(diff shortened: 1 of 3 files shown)


erikbaranowski (approved):
LGTM

PR #5338, opened 2023-09-29
Promtail converter will now treat `global positions configuration is …

### LINKED ISSUE #4835: Promtail converter should not fail with a valid positions block
### What's wrong?

If you try to convert a Promtail configuration that doesn't omit its `positions` block, then the converter will fail with a cryptic message. To successfully make the conversion people would have to comment out or remove the block.

I believe we shouldn't outright fail conversions from valid configuration files, and try to provide some better guidance here. Maybe make this into a warning instead?

Tangentially related;  is there a chance we also want to allow converting the positions file itself and providing instructions on _where_ to copy it to continue playback from the same point?

### Steps to reproduce

* Use a static mode config like the following
* Run the converter `./grafana-agent-flow convert --source-format=promtail --output converter-output.river promtail-config.yaml`
* The converter fails with the following message
```
Error: (Error) global positions configuration is not supported - each Flow Mode's loki.source.file component has its own positions file in the component's data directory
...
```

### System information

macOS

### Software version

Grafana Agent v0.35.3

### Configuration

```text
server:
  log_level: info
  profiling_enabled: true
  http_listen_address: "0.0.0.0"
  http_listen_port: 9999
  enable_runtime_reload: true
 
clients:
  - url: "<url>
    basic_auth:
      username: "<user>"
      password: "<pass>"

# Describes how to save read file offsets to disk
positions:
  filename: "/Users/tpaschalis/Workspace/compare-agent-promtail/promtail-positions.yaml"
  sync_period: "10s"


scrape_configs:
  - job_name: system
    static_configs:
      - targets:
          - localhost
        labels:
          job: varlogs  # A `job` label is fairly standard in prometheus and useful for linking metrics and logs.
          host: tpaschalis-macbook  # A `host` label will help identify logs from this machine vs others
          __path__: /var/log/*.log  # The path matching uses a third party library: https://github.com/bmatcuk/doublestar

# Configures how tailed targets will be watched.
target_config:
  sync_period: "10s"
```

### Logs

```text
Error: (Error) global positions configuration is not supported - each Flow Mode's loki.source.file component has its own positions file in the component's data directory
(Warning) If you have a tracing set up for Promtail, it cannot be migrated to Flow Mode automatically. Refer to the documentation on how to configure tracing in Flow Mode.
(Warning) server.profiling_enabled is not supported - use Agent's main HTTP server's profiling endpoints instead
(Warning) The Agent's Flow Mode metrics are different from the metrics emitted by Promtail. If you rely on Promtail's metrics, you must change your configuration, for example, your alerts and dashboards.
```
```

erikbaranowski:
…not supported` as a Warning instead of Error.

Closes #4835

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -125,6 +125,8 @@ Main (unreleased)
 - The `instance` label of targets exposed by `prometheus.exporter.*` components
   is now more representative of what is being monitored. (@tpaschalis)
 
+- Promtail converter will now treat `global positions configuration is not supported` as a Warning instead of Error. (@erikbaranowski)
+
 ### Bugfixes
 
 - Fixed `otelcol.exporter.prometheus` label names for the `otel_scope_info`
--- a/converter/internal/promtailconvert/testdata/unsupported.diags
+++ b/converter/internal/promtailconvert/testdata/unsupported.diags
@@ -1,4 +1,4 @@
-(Error) global positions configuration is not supported - each Flow Mode's loki.source.file component has its own positions file in the component's data directory
+(Warning) global positions configuration is not supported - each Flow Mode's loki.source.file component has its own positions file in the component's data directory
 (Error) Promtail's WAL is currently not supported in Flow Mode
 (Error) limits_config is not yet supported in Flow Mode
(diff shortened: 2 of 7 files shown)


mattdurham (approved):
LGTM

PR #5340, opened 2023-09-30
fix(pkg/flow): cmt stmt in arg block is now parseable

### LINKED ISSUE #5321: argument block "comment" argument fails to parse
### What's wrong?

The docs specify that the argument block has a "comment" argument that can be used to describe the argument, but when loading the module, grafana-agent-flow logs an error saying that it is unrecognized. I see this field commented out in agent-modules so I assume this is known, but I don't see an existing issue for this and the docs are currently misleading if it is not currently implemented

### Steps to reproduce

Create (and use) a river module file with the configuration below

### System information

Linux 4.14.314-238.539.amzn2.x86_64

### Software version

v0.36.1

### Configuration

```text
argument "foo" {
  comment = "A description of foo"
  optional = true
}
```

### Logs

```text
Failed to evaluate node for config block: decoding River: <module path>: unrecognized attribute name "comment"
```

hainenber:

Fixes #5321 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -142,6 +142,8 @@ Main (unreleased)
 
 - Fixed a bug where the BackOffLimit for the kubernetes tailer was always set to zero. (@anderssonw)
 
+- Fixed a bug where Flow agent fails to load `comment` statement in `argument` block. (@hainenber)
+
 ### Other changes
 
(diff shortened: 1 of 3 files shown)


erikbaranowski (approved):
LGTM, thank you for the contribution!

PR #5342, opened 2023-10-01
feat(ci): switch to GH action for testing CI build image

hainenber:
#### PR Description

From what I've checked, `drone.io`'s configuration [no longer supports triggering builds based on changed paths](https://docs.drone.io/pipeline/exec/syntax/trigger/). We should switch to GitHub Actions for this feature.

There should be no risks as the migrated parts are only for building throw-away test images, no pushing to any remote registry. 

#### PR Checklist

- [x] CHANGELOG.md updated


tpaschalis:
Thanks for opening that PR! Even though Drone itself does not support paths for its trigger syntax, there are third-party extensions that can be used for this.

For this reason, and since I think it has been working fine, I think it's best we continue using Drone for the build-image CI step. WDYT?

hainenber:
Yes, I noticed there's a 3rd-party [extension](https://github.com/meltwater/drone-convert-pathschanged) that would help our issue atm but as far as I'm aware, this would require the infrastructure complexity of hosting said extension as well as secrets handling/storage to connect Drone endpoint with it.

If we migrate to GH Actions, it'd be a hassle-(free) :)

rfratto:
For security reasons, we're currently unable to use Drone on PRs, and can only do it after merge. Given that, I think it makes sense to test the CI build images using GitHub Actions, so that we have validations earlier.

cc @tpaschalis WDYT?

tpaschalis:
> cc @tpaschalis WDYT?

Yeah, I'm okay with it (sorry missed your previous message @hainenber)

rfratto (approved):
Thanks @hainenber! I'm going to clean this PR up a little bit (to remove your test changes and resign the Drone config) and then get this merged. Nice work!

CHANGELOG.md
```diff
@@ -204,10 +204,13 @@ v0.37.0-rc.0 (2023-10-05)
 
 - `prometheus.operator.probes` no longer ignores relabeling `rule` blocks. (@sberz)
 
+- Migrate `Check Linux/Windows build image` to GitHub Actions. (@hainenber)
```
rfratto, line 207: While we appreciate your work here, the CHANGELOG is only for features that impact end users of the agent, so I'll be removing this line, sorry 😞

hainenber: Yeah, no worries :D

PR #5361, opened 2023-10-03
Remove singleton and allow unix exporter to be used multiple times

### LINKED ISSUE #3155: Proposal: remove singleton flag from `prometheus.exporter.unix`

mattdurham:
#### PR Description

This fixes the long standing issue of our only singleton component and allows prometheus.exporter.unix to be defined multiple times. It does this by using a custom fork of node_exporter that moves away from kingpin. Tested the default set and every metric and the number of series differed by 3 (number of metrics were the same) which seems to be from slight changes done to the node_exporter.

Closes #3155

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Documentation added
- [X] Tests updated


mattdurham:
Tomorrow going to test Darwin but figured would go ahead and start the review.

CHANGELOG.md
```diff
@@ -81,6 +81,8 @@ Main (unreleased)
 - Flow: the `prometheus.scrape` component can now configure the scraping of
   Prometheus native histograms. (@tpaschalis)
 
+- Flow: Allow `prometheus.exporter.unix` to be specified multiple times and used in modules. (@mattdurham)
```
rfratto, line 84: This is a breaking change (though a breaking change we will all appreciate); can we document it as such?

rfratto: I'd suggest there being two changelog entries here:

* The breaking change that `prometheus.exporter.unix` now requires a label.
* The new feature (or enhancement, however it's categorized now) that `prometheus.exporter.unix` can now be specified multiple times and used in modules.

go.mod
```diff
@@ -728,3 +729,6 @@ exclude (
 )
 
 replace github.com/github/smimesign => github.com/grafana/smimesign v0.2.1-0.20220408144937-2a5adf3481d3
+
+// Replace node_export with custom fork for multi usage.
+replace github.com/prometheus/node_exporter => github.com/grafana/node_exporter v0.18.1-grafana-r01.0.20231003194228-26e058a40f9a
```
marctc, line 734: Could we move where the rest of replace of exporter forks are and also add the link of the PR?

CHANGELOG.md
mattdurham: I thought it would be a breaking change but as long as its the only `prometheus.exporter.unix` no label is needed.  Maybe this is a bug but in this case it means people can upgrade cleanly until they add another one.

This works fine if its the only one.
```
prometheus.exporter.unix {

}
```

rfratto: @mattdurham that's odd, and it _does_ seem like a bug. Labels are supposed to be required on non-singleton components, did that logic get removed?

pkg/flow/internal/controller/loader.go
```diff
@@ -445,36 +445,6 @@ func (l *Loader) populateComponentNodes(g *dag.Graph, componentBlocks []*ast.Blo
...
-
-			if !registration.Singleton && block.Label == "" {
-				diags.Add(diag.Diagnostic{
-					Severity: diag.SeverityLevelError,
-					Message:  fmt.Sprintf("Component %q must have a label", componentName),
-					StartPos: block.NamePos.Position(),
-					EndPos:   block.NamePos.Add(len(componentName) - 1).Position(),
-				})
-				continue
-			}
```
rfratto, line 466: @mattdurham I think we should keep this check; all components must have labels now.

CHANGELOG.md
mattdurham: Seems to be specific to the exporter which leads me to believe something is hanging around that is making it behave differently. Looking into it.

pkg/flow/internal/controller/loader.go
mattdurham: How would you feel about adding a carve out for unix exporter here? If this rolls out its going to be a huge breaking change.

rfratto: Personally, I'd rather make the change now so we're consistent across the board. We have a few more opportunities to do breaking changes before we need to settle on things for 1.0, at which point we're stuck for a while, so now's the time to make these kinds of changes.

If you're uneasy about this, we should probably have a discussion between the maintainers to figure out what we're going to do.

CHANGELOG.md
mattdurham: Ignore the above, you found the issue in loader where I was overzealous

pkg/flow/internal/controller/loader.go
mattdurham: Followup going to leave the  breaking change in.

converter/internal/staticconvert/internal/build/node_exporter.go
```diff
@@ -12,11 +12,11 @@ func (b *IntegrationsV1ConfigBuilder) appendNodeExporter(config *node_exporter.C
 	args := toNodeExporter(config)
 	b.f.Body().AppendBlock(common.NewBlockWithOverride(
 		[]string{"prometheus", "exporter", "unix"},
-		"",
+		"u",
```
rfratto, line 15: `u` feels like a strange label to me; I feel like `default` is more idiomatic based on our examples.

docs/sources/flow/release-notes.md
```diff
@@ -62,6 +62,11 @@ to `true` and controlled via the `include_scope_labels` argument.
 `name` to `otel_scope_name` and `version` to `otel_scope_version`. This is
 now correct with the OTLP Instrumentation Scope specification.
 
+### Breaking change: `prometheus.exporter.unix` now requires a label.
+
+Previously the exporter was a singleton and did not require a label. The exporter now can be used multiple times and
+needs a label, `prometheus.exporter.unix "example"`.
```
rfratto, line 68: Aligning with how other breaking changes are documented (with before/after examples):

``````suggestion
Previously the exporter was a singleton and did not require a label. The exporter now can be used multiple times and
needs a label.

Old configuration example:

```river 
prometheus.exporter.unix { /* ... */ } 
```

New configuration example:

```river 
prometheus.exporter.unix "example" { /* ... */ } 
```
``````

**Review by rfratto — APPROVED** [review]

**Review by marctc — APPROVED** [review]

File: pkg/integrations/node_exporter/config.go [hunk 6]
```diff
@@ -300,209 +302,230 @@ func MapConfigToNodeExporterFlags(c *Config) (accepted []string, ignored []strin
...
-			"--collector.ntp.protocol-version", fmt.Sprintf("%d", c.NTPProtocolVersion),
-			"--collector.ntp.ip-ttl", fmt.Sprintf("%d", c.NTPIPTTL),
-			"--collector.ntp.max-distance", c.NTPMaxDistance.String(),
-			"--collector.ntp.local-offset-tolerance", c.NTPLocalOffsetTolerance.String(),
-		)
-
-		flags.addBools(map[*bool]string{
-			&c.NTPServerIsLocal: "collector.ntp.server-is-local",
-		})
+	defaultPort := 123
```
marctc, line 420: I feel that all these default should be imported directly from node exporter itself. Maybe we should add a TODO or something?

PR #5364, opened 2023-10-04
Prevent hot loop retry on full worker queue

thampiotr:
#### PR Description

This prevents a hot retry loop in the unlikely event of worker queue getting full. The symptom of hot retry loop would be a lot of log lines saying `failed to submit node for evaluation - the agent is likely overloaded and cannot keep up with evaluating components` - as it could be observed when running a test dedicated to reproducing this situation - `TestController_Updates_WithQueueFull`. 

#### Notes to the Reviewer

* `Flow` passes updated components to `loader`'s `EvaluateDependencies`, which submits the component's direct dependencies for evaluation to the worker pool.
* In unlikely event that the pool is full, e.g. we have a component that runs into a deadlock, we would currently log an error and return the parent component back to `Flow` for a retry. But `Flow` would attempt retry right away, leading to a hot loop using CPU and printing a lot of logs.
* In this PR I add a `timeout` parameter to worker pool's submit functions, which will attempt to queue the task until the `timeout` duration passes. As a result, we'll block for some time before failing when the queue is completely full, and the hot retry loop will be prevented, as well as the error log line will be printed less frequently.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/flow/internal/worker/worker_pool.go
+++ b/pkg/flow/internal/worker/worker_pool.go
@@ -6,31 +6,35 @@ import (
 	"math/rand"
 	"runtime"
 	"sync"
+	"time"
 )
 
 type Pool interface {
 	// Stop stops the worker pool. It does not wait to drain any internal queues, but it does wait for the currently
 	// running tasks to complete. It must only be called once.
 	Stop()
-	// Submit submits a function to be executed by the worker pool on a random worker. Error is returned if the pool
-	// is unable to accept extra work.
-	Submit(func()) error
-	// SubmitWithKey submits a function to be executed by the worker pool, ensuring that only one
+	// TrySubmit submits a function fn to be executed by the worker pool on a random worker. Error is returned if the
+	// pool is unable to accept extra work within the provided timeout.
+	TrySubmit(fn func(), timeout time.Duration) error
+	// TrySubmitWithKey submits a function to be executed by the worker pool, ensuring that only one
 	// job with given key can be queued at the time. Adding a job with a key that is already queued is a no-op (even if
-	// the submitted function is different). Error is returned if the pool is unable to accept extra work - the caller
-	// can decide how to handle this situation.
-	SubmitWithKey(string, func()) error
+	// the submitted function is different). Error is returned if the pool is unable to accept extra work within the
+	// provided timeout - the caller can decide how to handle this situation.
+	TrySubmitWithKey(key string, fn func(), timeout time.Duration) error
 	// QueueSize returns the number of tasks currently queued.
 	QueueSize() int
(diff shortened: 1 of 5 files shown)


pkg/flow/internal/worker/worker_pool.go
```diff
@@ -78,19 +85,30 @@ func (w *shardedWorkerPool) SubmitWithKey(key string, f func()) error {
...
-		return nil // only queue one job for given key
-	}
-
-	select {
-	case queue <- wrapped:
-		w.set[key] = struct{}{}
-		return nil
-	default:
-		return fmt.Errorf("worker queue is full")
+	for {
```
thampiotr, line 90: We need a loop where we occasionally release the `w.lock`, because otherwise we could block the execution of the tasks themselves and lead to a deadlock.

pkg/flow/internal/worker/worker_pool.go
```diff
@@ -6,46 +6,52 @@ import (
...
-	// the submitted function is different). Error is returned if the pool is unable to accept extra work - the caller
-	// can decide how to handle this situation.
-	SubmitWithKey(string, func()) error
+	// the submitted function is different). Error is returned if the pool is unable to accept extra work within the
+	// provided timeout - the caller can decide how to handle this situation.
+	TrySubmitWithKey(key string, fn func(), timeout time.Duration) error
 	// QueueSize returns the number of tasks currently queued.
 	QueueSize() int
+	// DefaultTimeout returns the default timeout that can be used with TrySubmit and TrySubmitWithKey.
+	DefaultTimeout() time.Duration
```
thampiotr, line 27: I've opted to keeping the timeout configuration as part of the `Pool` so that we don't need to pass a separate timeout field around - and there are quite a few places between controller/loader/modules which would need it.

pkg/flow/internal/worker/worker_pool.go
```diff
@@ -6,46 +6,52 @@ import (
...
+	defaultTimeout time.Duration
 
 	lock sync.Mutex
 	set  map[string]struct{}
 }
 
 var _ Pool = (*shardedWorkerPool)(nil)
 
+// NewDefaultWorkerPool creates a new worker pool suitable for use in Flow controllers. Since Flow retries component
+// dependency updates, the timeout used will mostly determine how often we'll log an error when the queue is full.
```
thampiotr, line 46: Because of that, I don't think we'll need to make it configurable.

thampiotr:
I came up with a simpler solution in https://github.com/grafana/agent/pull/5370

PR #5369, opened 2023-10-04
Proposal: integration testing

wildum:
#### PR Description

This PR introduces integration testing to the agent for flow.

With a solid test coverage, the agent should be more stable and it should be easier to update code affecting several components without having to go through manual testing. This would be particularly valuable when updating important dependencies like OpenTelemetry or Prometheus. As a result, we might be able to accelerate the release process which is currently quite long.

Another benefit of these tests is to provide examples of pipelines to the users to help them understand how the components can interact together. With some luck, a user might be able to find a test with his exact use case and be able to just copy the config.river file.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added


integration-tests/run-integration-tests.sh
```diff
@@ -0,0 +1,78 @@
...
+        kill $AGENT_PID || true
+        popd
+    ) &
+
+    counter=$((counter+1))
+    if [ "$counter" -eq "$parallel_runs" ]; then
+        wait
+        counter=0
+    fi
+done < <(find ./tests -maxdepth 1 -type d ! -path ./tests)
```
thampiotr, line 65: Can we use something other than `bash` for this? I think with just a bit extra effort we can have a Go implementation that would be easier to maintain and debug in the long run.

integration-tests/configs/otel-gen-client/go.mod
```diff
@@ -0,0 +1,39 @@
+module github.com/open-telemetry/opentelemetry-collector-contrib/examples/demo/client
```
thampiotr, line 1: Can we avoid the `go.mod` files altogether? Ideally we would bring in this dependency in an idiomatic way in our main go.mod. Is there a reason why it cannot be done?

integration-tests/run-integration-tests.sh
thampiotr: Say I have an issue with one test only, how do I run a single test?

integration-tests/run-integration-tests.sh
```diff
@@ -0,0 +1,78 @@
...
+counter=0
+
+# Run tests in parallel
+while read -r test_dir; do
+    (
+        pushd "$test_dir"
+        dir_name=$(basename "$test_dir")
+        agent_logfile="../../${tmp_dir}/${dir_name}_agent.log"
+        test_logfile="../../${tmp_dir}/${dir_name}_test.log"
+        "$AGENT_BINARY_PATH" run config.river > "$agent_logfile" 2>&1 &
```
thampiotr, line 45: Don't all the agents attempt to open on the same port 12345? How do you avoid this problem?

integration-tests/run-integration-tests.sh
```diff
@@ -0,0 +1,78 @@
...
+        rm -rf data-agent
+        kill $AGENT_PID || true
+        popd
+    ) &
+
+    counter=$((counter+1))
+    if [ "$counter" -eq "$parallel_runs" ]; then
+        wait
+        counter=0
+    fi
```
thampiotr, line 64: In other language than Bash you could also start 1 new test as 1 test finishes, instead of waiting for all of them to finish to start a new batch.

integration-tests/run-integration-tests.sh
tpaschalis: Also, a go implementation would allow us to shuffle the order execution, and use build tags to allow for simpler `go test` commands. I agree it's worth exploring.

wildum: I changed the script to use Go and added command options to:
- run without building the agent, useful when running it locally
- run one test in particular

integration-tests/run-integration-tests.sh
wildum: switched to Go

integration-tests/configs/otel-gen-client/go.mod
wildum: I removed it to use the one from the project

integration-tests/run-integration-tests.sh
wildum: I guess that the first one succeed and the other ones don't but that does not seem to create any problem. I will need to dig into that

wildum:
Draft Closed, PR opened here: https://github.com/grafana/agent/pull/5475

PR #5370, opened 2023-10-04
Prevent hot loop retry on full worker queue

thampiotr:
#### PR Description

This prevents a hot retry loop in the unlikely event of worker queue getting full. The symptom of hot retry loop would be a lot of log lines saying `failed to submit node for evaluation - the agent is likely overloaded and cannot keep up with evaluating components` - as it could be observed when running a test dedicated to reproducing this situation - `TestController_Updates_WithQueueFull`. 

#### Notes to the Reviewer

* `Flow` passes updated components to `loader`'s `EvaluateDependencies`, which submits the component's direct dependencies for evaluation to the worker pool.
* In unlikely event that the pool is full, e.g. we have a component that runs into a deadlock, we would currently log an error and return the parent component back to `Flow` for a retry. But `Flow` would attempt retry right away, leading to a hot loop using CPU and printing a lot of logs.
* In this PR I add a `backoff` and retry mechanism to `loader`'s `EvaluateDependencies` function, so that we will retry submitting to worker queue until we succeed with a backoff between attempts. As a result, we don't need a retry mechanism in `Flow` anymore, we won't be retrying on a hot loop, we won't spam the logs and we can cope with temporary slowness of components. 
* If there is a deadlock or a component takes forever to evaluate, we will log an error message which can help with investigations, but we're not automatically recovering from it. 

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/flow/internal/controller/loader.go
+++ b/pkg/flow/internal/controller/loader.go
@@ -586,7 +586,8 @@ func (l *Loader) OriginalGraph() *dag.Graph {
 // workerPool. It should be called whenever components update their exports.
 // It is beneficial to call EvaluateDependencies with a batch of components, as it will enqueue the entire batch before
 // the worker pool starts to evaluate them, resulting in smaller number of total evaluations when
-// node updates are frequent.
+// node updates are frequent. If the worker pool's queue is full, EvaluateDependencies will retry with a backoff until
+// it succeeds or until the ctx is cancelled.
 func (l *Loader) EvaluateDependencies(ctx context.Context, updatedNodes []*ComponentNode) {
 	if len(updatedNodes) == 0 {
 		return
--- a/pkg/flow/internal/controller/loader.go
+++ b/pkg/flow/internal/controller/loader.go
@@ -13,6 +13,7 @@ import (
 	"github.com/grafana/agent/pkg/flow/internal/worker"
 	"github.com/grafana/agent/pkg/flow/tracing"
 	"github.com/grafana/agent/service"
+	"github.com/grafana/dskit/backoff"
 	"github.com/grafana/river/ast"
 	"github.com/grafana/river/diag"
 	"github.com/hashicorp/go-multierror"
@@ -30,6 +31,11 @@ type Loader struct {
 	host         service.Host
 	componentReg ComponentRegistry
 	workerPool   worker.Pool
+	// backoffConfig is used to backoff when an updated component's dependencies cannot be submitted to worker
+	// pool for evaluation in EvaluateDependencies, because the queue is full. This is an unlikely scenario, but when
+	// it happens we should avoid retrying too often to give other goroutines a chance to progress. Having a backoff
+	// also prevents log spamming with errors.
+	backoffConfig backoff.Config
 
 	mut               sync.RWMutex
 	graph             *dag.Graph
@@ -76,6 +82,13 @@ func NewLoader(opts LoaderOptions) *Loader {
 		host:         host,
 		componentReg: reg,
 		workerPool:   opts.WorkerPool,
+
+		// This is a reasonable default which should work for most cases. If a component is completely stuck, we would
+		// retry and log an error every 10 seconds, at most.
+		backoffConfig: backoff.Config{
+			MinBackoff: 1 * time.Millisecond,
+			MaxBackoff: 10 * time.Second,
+		},
 
 		graph:         &dag.Graph{},
 		originalGraph: &dag.Graph{},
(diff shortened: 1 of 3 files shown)


pkg/flow/internal/controller/loader.go
```diff
@@ -570,15 +583,13 @@ func (l *Loader) OriginalGraph() *dag.Graph {
...
 // EvaluateDependencies sends components which depend directly on components in updatedNodes for evaluation to the
-// workerPool. It should be called whenever components update their exports. It returns a list of components for which
-// the dependencies failed to be enqueued for async evaluation and should be retried in the future.
-//
+// workerPool. It should be called whenever components update their exports.
 // It is beneficial to call EvaluateDependencies with a batch of components, as it will enqueue the entire batch before
 // the worker pool starts to evaluate them, resulting in smaller number of total evaluations when
 // node updates are frequent.
-func (l *Loader) EvaluateDependencies(updatedNodes []*ComponentNode) []*ComponentNode {
+func (l *Loader) EvaluateDependencies(ctx context.Context, updatedNodes []*ComponentNode) {
```
ptodev, line 590: In the comment above it'd be good to mention what `ctx` is used for. Some functions use it as a key/value pair, while others just use the "cancel" feature.. It has many uses and if the comment above the function doesn't say what it's for, people have to browse the code to find out.

pkg/flow/internal/controller/loader.go
```diff
@@ -607,28 +618,37 @@ func (l *Loader) EvaluateDependencies(updatedNodes []*ComponentNode) []*Componen
...
+		var (
+			nodeRef, parentRef = n, parent
+			retryBackoff       = backoff.New(ctx, l.backoffConfig)
+			err                error
+		)
+		for retryBackoff.Ongoing() {
+			err = l.workerPool.SubmitWithKey(nodeRef.NodeID(), func() {
+				l.concurrentEvalFn(nodeRef, dependantCtx, tracer, parentRef)
+			})
+			if err != nil {
```
ptodev, line 636: Is being overloaded the only reason for an error? What if there is an error which is not due to overloading, for which there is no point in retrying?

pkg/flow/internal/controller/loader.go
```diff
@@ -77,6 +83,13 @@ func NewLoader(opts LoaderOptions) *Loader {
 		componentReg: reg,
 		workerPool:   opts.WorkerPool,
 
+		// This is a reasonable default which should work for most cases. If a component is completely stuck, we would
+		// retry and log an error every 10 seconds, at most.
+		backoffConfig: backoff.Config{
+			MinBackoff: 1 * time.Millisecond,
+			MaxBackoff: 10 * time.Second,
+		},
```
ptodev, line 91: I had a look at the backoff config, and I think we need to set `MaxRetries` to something other than 0 for `Ongoing()` to ever return `false`. Otherwise I think it'll keep retrying infinitely. IMO there should be a retry limit, but I don't know what's a reasonable one.

ptodev (approved):
Looks good! I like it. My main concern is that the default config may not work for all the customers, and that some may need it to configure it:
* What if a 10 second backoff is too long for some customers?
* Presumably the Agent will drop data on the floor if it retries too many times? Some customers may want it to try harder to send the data before giving up.

That said, I think it's ok to put in some reasonable defaults for now, and change this to be customisable in a future Agent release. In fact, I think this is preferable - it would be good to get customer feedback before we expose ways to configure internals, especially since the way we implement retries may change.

I think we don't have to expose the exact settings. If we don't want to expose internals, then maybe we can have a "pipeline_throttle" setting which customers can set from 1-5:
*  "1" would mean more retries, leading to potentially a slower evaluation but less data being dropped.
*  "5" would be less retries and more speed, but also more data potentially dropped on the floor.
* Values between 1 and 5 would correspond to some in-between settings.

Then we don't have to expose internals, just this one "pipeline_throttle" heuristic.

pkg/flow/internal/controller/loader.go
thampiotr: This is actually intentional - we will be retrying forever to ensure we don't drop any data.
The `Ongoing()` will stop when the parent `context` is cancelled, which is when the `Flow` controller shuts down when the agent is shutting down. 

So we have infinite retry loop that can end only if: a) the worker pool is unclogged somehow, b) the agent shuts down.

pkg/flow/internal/controller/loader.go
thampiotr: Yes, being overloaded is the only possible reason for the error here. It will always make sense to retry it, because we don't want to miss evaluating that component, otherwise it will have stale data.

thampiotr:
> * What if a 10 second backoff is too long for some customers?

As discussed offline, the agent doesn't drop the data but retries forever. The max 10s backoff is just the backoff when adding more work to the queue, it doesn't mean that the work will be getting done any slower. Therefore we believe that this is a good default and customers will not need to adjust it. Also, this is a setting relevant only during a failure mode where components do not evaluate in a reasonable time.

> * Presumably the Agent will drop data on the floor if it retries too many times? Some customers may want it to try harder to send the data before giving up.

As mentioned above, we are not dropping data, we will retry forever.

PR #5384, opened 2023-10-06
build: use consistent versioning

rfratto:
This change enables module.http servers to use the User-Agent header to reliably return different content based on the reported agent version for non release builds.

Previously, module.http servers could only make use of the User-Agent header in release or release candidate versions. Non-release builds were previously tagged as BRANCH-SHA, which do not include any information about what components are available.

Now, non-release builds always include a version, specifically the upcoming version. For example, since v0.38 is the current version under development, main branch builds would be tagged as v0.38.0-devel+SHA.

As part of this change, the release process now includes updating the newly added VERSION file after a release branch is cut; this only needs to be done for the main branch, and only done _after_ a release branch is cut (that is: at the start of a new development cycle for a new minor release).


docs/developer/release/1-create-release-branch.md
```diff
@@ -23,4 +23,8 @@ Patch Releases for that major pr minor version of the agent.
     - If the consensus commit is not the latest commit from main, branch from that instead.
 
     > **NOTE**: Don't create any other branches that are prefixed with `release` when creating PRs or
-    those branches will collide with our automated release build publish rules.
\ No newline at end of file
+    those branches will collide with our automated release build publish rules.
+
+3. Open a PR against the `main` branch to update the VERSION file to the _next_ release.
+
+    For example, if working on release-v0.31, the VERSION file in main should be updated to v0.32.0.
```
thampiotr, line 30: I like the idea in general. I'm just not a fan of creating another manual step. PRs like this https://github.com/grafana/agent/pull/5373 may accidentally change this to wrong value since there is 17+ places to update already and it'll be easy to mix up. 

Can we write a short program that will grab one of the versions that we already have in repo, (like these ones https://github.com/grafana/agent/pull/5373) and use semver to increment its minor version? A major release with breaking changes would require some extra manual steps though.

Alternatively, to make it crystal clear what is inside `VERSION`, maybe we should change it to `yaml` or something and have clearly named fields with `latestReleasedVersion` and `nextReleaseVersion`. And then we could instead build some automation where possible to read version from that file and avoid 17+ places to update version in PRs like this https://github.com/grafana/agent/pull/5373.

tpaschalis: I think it might be a good opportunity for #5333 handle this in addition to all the different places we have to change version.

tools/image-tag
```diff
@@ -9,13 +9,13 @@ fi
...
-BRANCH=$(git rev-parse --abbrev-ref HEAD | sed 's#/#-#g')
+VERSION=$(cat VERSION)
 
-# When 7 chars are not enough to be unique, git automatically uses more.
-# We are forcing to 7 here, as we are doing for grafana/grafana as well.
-SHA=$(git rev-parse --short=7 HEAD | head -c7)
-
-# If this is a tag, use it, otherwise use <branch>-<hash>
-TAG=$(git describe --exact-match 2> /dev/null || echo "${BRANCH}-${SHA}")
-echo ${TAG}${WIP}
```
tpaschalis, line 21: Can we keep the new behavior of using the actual tag if it matches?

docs/developer/release/1-create-release-branch.md
thampiotr: yay! #5333 is awesome

rfratto: I thought about this some last night and there's definitely automation we can introduce to alleviate this extra step. 

I think it should be possible to automatically update this file in main whenever a new release branch is created.

rfratto:
I'm going to close this for now in favor of a proposal (#6100). I'll reopen this depending on the outcome of that proposal.

PR #5387, opened 2023-10-06
Remove duplicated entry in changelog

wildum:


tpaschalis:
Can we also alphabetically sort the list of components as we used to do in previous releases?

tpaschalis (approved):

PR #5406, opened 2023-10-06
reorganize prometheus converter code and limit surface area of dependencies between converters

erikbaranowski:
#### PR Description

This does some things
- organize prometheus component code into a subfolder
- minimize surface area between converter dependencies

BEFORE
image

AFTER
image

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


converter/internal/prometheusconvert/build/prometheus_blocks.go
thampiotr, line 1: Should this be in `internal/build` and similarly, should we have `internal/component` where possible? So that using unintended parts of `prometheusconvert` package is not possible without deliberately exposing them?

thampiotr (approved):
Just one nit, LGTM otherwise!

PR #5407, opened 2023-10-06
Add new options to spanmetrics processor in static mode:

### LINKED ISSUE #4614: spanmetrics: wire in new metrics_flush_interval field
### Request

Since we've updated our OpenTelemetry Collector dependency, we can now wire in a new field available in the spanmetrics processor, `metrics_flush_interval`.

### Use case

This would help with limiting the rate of metrics generation, from every time a trace was received to a preset interval and help users manage their data throughput.

ptodev:
#### PR Description
Adding new config options to spanmetrics processor in static mode:
  - `aggregation_temporality`: configures whether to reset the metrics after flushing.
  - `metrics_flush_interval`: configures how often to flush generated metrics.

Fixes #4614

#### Notes to the Reviewer

This change is ok from a feature parity point of view because both of these config options are present in [otelcol.connector.spanmetrics](https://grafana.com/docs/agent/latest/flow/reference/components/otelcol.connector.spanmetrics/).

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,10 @@ v0.37.2 (2023-10-16)
 
 - Allow Out of Order writing to the WAL for metrics. (@mattdurham)
 
+- Added new config options to spanmetrics processor in static mode (@ptodev):
+  - `aggregation_temporality`: configures whether to reset the metrics after flushing.
+  - `metrics_flush_interval`: configures how often to flush generated metrics.
+
 ### Other changes
 
 - Use Go 1.21.3 for builds. (@tpaschalis)
--- a/docs/sources/static/configuration/traces-config.md
+++ b/docs/sources/static/configuration/traces-config.md
@@ -157,7 +157,7 @@ remote_write:
 automatic_logging:
   # Indicates where the stream of log lines should go. Either supports writing
   # to a logs instance defined in this same config or to stdout.
-  [ backend: <string> | default = "stdout" | supported "stdout", "logs_instance" ]
+  [ backend: <string> | default = "stdout" | supported = "stdout", "logs_instance" ]
   # Indicates the logs instance to write logs to.
   # Required if backend is set to logs_instance.
   [ logs_instance_name: <string> ]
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -159,6 +159,10 @@ v0.37.0-rc.1 (2023-10-06)
 - The `client_secret` config argument in the `otelcol.auth.oauth2` component is 
   now of type `secret` instead of type `string`. (@ptodev)
 
+- Added new config options to spanmetrics processor in static mode:
+  - `aggregation_temporality`: configures whether to reset the metrics after flushing.
+  - `metrics_flush_interval`: configures how often to flush generated metrics.
+
```
tpaschalis, line 165: Let's rebase and relocate these to the Main (unreleased) header.

docs/sources/static/configuration/traces-config.md
```diff
@@ -264,8 +264,16 @@ spanmetrics:
   [ metrics_instance: <string> ]
   # handler_endpoint defines the endpoint where the OTel prometheus exporter will be exposed.
   [ handler_endpoint: <string> ]
-  # dimensions_cache_size defines the size of cache for storing Dimensions
+  # dimensions_cache_size defines the size of cache for storing Dimensions.
+  # Default: 1000
   [ dimensions_cache_size: <int> ]
```
tpaschalis, line 269: nit: can we document the default like we do in other places for static mode?
```suggestion
  [ dimensions_cache_size: <int> | default = 1000 ] 
```

docs/sources/static/configuration/traces-config.md
```diff
@@ -264,8 +264,16 @@ spanmetrics:
...
+  # aggregation_temporality configures whether to reset the metrics after flushing.
+  # It can be either AGGREGATION_TEMPORALITY_CUMULATIVE or AGGREGATION_TEMPORALITY_CUMULATIVE.
+  # Default: AGGREGATION_TEMPORALITY_CUMULATIVE
+  [ aggregation_temporality: <int> ]
+  # metrics_flush_interval configures how often to flush generated metrics.
+  # Default: 15s
+  [ metrics_flush_interval: <int> ]
```
tpaschalis, line 276: Same here, should this be a `<duration>` instead?

ptodev:
@tpaschalis thank you for the review, and apologies for the silly errors! Btw I see that we seem to have a practice to list supported values like this:

```
[ backend: <string> | default = "stdout" | supported = "stdout", "logs_instance" ]
```

I didn't do this for `aggregation_temporality` because it'd make the line too long and it looks awkward. But let me know if you feel strongly about it and I'll add this in. There are also a few other places which don't use this syntax to list supported values, so we could add it there too.

clayton-cornell (commented):
Doc input is minimal. OK as-is.

tpaschalis (approved):
LGTM, thanks!

clayton-cornell (approved):

PR #5408, opened 2023-10-06
Enable compression in logs WAL

thepalbi:
#### PR Description

This PR enables compression the `loki.write` WAL. Both snappy and zstd are supported by the reader/writer side, and the reader side checks the compression of each record read (so no compat issues). Snappy was picked because since this it will be used in a WAL, rather than a wire-protocol, we strive more for cpu efficiency (snappy is cpu lighter) rather higher compression rate (which zstd does better, but at a higher CPU cost).

Enabling compression in WAL allowed us in our agents deployment to have a compression rate of ~2.

image

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ internal API changes are not present.
 Main (unreleased)
 -----------------
 
+### Enhancements
+
+- The `loki.write` WAL now has snappy compression enabled by default. (@thepalbi)
+
+
 v0.37.0-rc.0 (2023-10-05)
 -----------------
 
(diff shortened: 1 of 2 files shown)


tpaschalis (approved):
I'm okay with this change.

But before we merge, can we check how this affects people who were _already_ using the WAL without compression? Do they need to delete their previous WAL? Do they continue operation normally? Any errors/logs they might see?

In case there's breaking behavior here we'd do well to document it.

thepalbi:
@tpaschalis no breaking behaviour expected, and WAL will keep working as is after an agent is restart with compression turned on. 

Basically, when an agent restarts, on the writer side, the wlog will start writing in a new segment, so the existing segments with compression disabled will co-exist with the compressed one. 

On the other side, from the reader end, the `LiveReader` [checks if each wal entry is compressed or not individually](https://github.com/prometheus/prometheus/blob/a5a4eab679ccf2d432ab34b5143ef9e1c6482839/tsdb/wlog/live_reader.go#L216-L234), hence it will handle compressed and uncompressed ones gracefully.

thepalbi:
@tpaschalis is this good to go?

PR #5410, opened 2023-10-06
helm: make installing CRDs optional through the new crds.create value

rfratto:
Currently, CRDs can only be disabled through the `--skip-crds` flag, which may not be available for all users interacting with Helm.

This PR moves the CRDs to a subchart, allowing them to be dynamically disabled through a new setting (`crds.create`).

This follows the pattern established by https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus-operator-crds, thanks to the advice given by @bentonam.


tpaschalis (approved):
LGTM!

PR #5412, opened 2023-10-09
docs: fix 'Flow mode' link

d4nyll:
#### PR Description
Fix link to [Flow Mode](https://grafana.com/docs/agent/latest/) in docs:

Screenshot 2023-10-09 at 09 49 55


tpaschalis (approved):
I'm preemptively approving, but leaving it to @clayton-cornell for the final decision.

I'm not sure why the link is not working here; does it have to do with it being inside of the admonition block?

docs/sources/_index.md
```diff
@@ -21,7 +21,7 @@ form programmable observability **pipelines** for telemetry collection,
 processing, and delivery.
 
 {{% admonition type="note" %}}
-This page focuses mainly on [Flow mode][], the Terraform-inspired variant of Grafana Agent.
+This page focuses mainly on [Flow mode]({{< relref "./flow/_index.md" >}}), the Terraform-inspired variant of Grafana Agent.
```
clayton-cornell, line 24:
```suggestion
This page focuses mainly on Flow mode, the Terraform-inspired variant of Grafana Agent.
```

I'm going to just remove the link here. We link to the Flow mode docs in several places throughout this topic, so we aren't leaving the readers hanging.

We need to avoid using relref where possible because these topics are mounted in the Cloud docs and relrefs like thsi will fail in Cloud (because there's no destination for this relref when the topic is viewed in Cloud docs)

clayton-cornell:
@d4nyll Thanks for spotting this one. It slipped through on me. 

The link is failing to render correctly because it's inside a note block, and the link definition is set in a `doc/reference` ta the bottom of the topic. We are using the `doc/reference` link definitions here because we mount the Static topics in the OSS documentation and in the Grafana Cloud documentation... and we need some special handling for linking so everything work in both doc locations.

The simplest fix here is to simply remove the link from the Note. We have this link elsewhere within the same topic so we aren't loosing anything.

clayton-cornell:
Hmmm... looking deeper into this one, and it looks like the fix is a bit complicated because it happened right around when we templated this `_index.md` file.  So the mistake exists in `_index.md` in the v0.37.1 release and then it was carried over in the conversion to the templated file in `main` (tagged to `next`). So I'll close this PR and fix it in the template in main and then manually change the original file in v0.37.1 in a second PR.

clayton-cornell:
Fixed in:
* https://github.com/grafana/agent/pull/5464
* https://github.com/grafana/agent/pull/5465

Closing this PR.  Thanks for finding this mistake @d4nyll  The fix took a little fiddling and two separate PRs to get it all right. :-)

PR #5424, opened 2023-10-10
Fix: Initialise default namespaces map config

wildum:
#### PR Description

Initialise a map to prevent panic

### DIFF
--- a/component/loki/source/kubernetes_events/event_controller.go
+++ b/component/loki/source/kubernetes_events/event_controller.go
@@ -97,7 +97,9 @@ func (ctrl *eventController) runError(ctx context.Context) error {
 	}
 
 	defaultNamespaces := map[string]cache.Config{}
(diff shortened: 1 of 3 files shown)


component/loki/source/kubernetes_events/event_controller.go
```diff
@@ -97,7 +97,9 @@ func (ctrl *eventController) runError(ctx context.Context) error {
 	}
 
 	defaultNamespaces := map[string]cache.Config{}
-	defaultNamespaces[ctrl.task.Namespace] = cache.Config{}
+	if ctrl.task.Namespace != "" {
```
tpaschalis, line 100: Should we _not_ initialise the map if the string is empty (and move the initialization inside this if clause?

tpaschalis (approved):
🛥️

PR #5425, opened 2023-10-10
helm: move initContainers to inside controller for consistency with pod settings

rfratto:
Other pod-level settings are determined by the `controller` group of settings, so initContainers should be inside of controller (`controller.initContainers`) rather than a top-level field.


tpaschalis (approved):
LGTM, let's get this in for the next Helm chart release.

Issue #4836, opened 2023-08-17
Sync loki.process with Promtail 2.8.4


tpaschalis:

tpaschalis:
Looking at the list of commits since October 2022 in Loki's `clients/pkg/logentry` package and the original inception of the loki.process components, here are the commits that _may_ want some extra work to be ported over.

- [X] 7d5475541 promtail: enable structured_metadata convert labels (#10752)
- [ ] f51ee849b fix(promtail): correctly parse list of drop stage sources from YAML (#10848)
- [X] 00a3c5b02 renamed `non indexed labels` feature to `structured metadata` (#10432)
- [X] bde65667f Upgrade dskit to use packages migrated from weaveworks/common (#10164)
- [X] 0fc934e4e non_indexed_labels stage: additional test-case (#10170)
- [X] 416e01ca8 [promtail] non_indexed_labels stage (#9986)
- [ ] 2dfac7283 Clarify that label filter expressions are not supported in promtail. (#9894)
- [X] 8abe080fd Promtail: improve behavior of partial lines (#9508)
- [X] 4a5644568 Upgrade `golangci-lint` and fix linting errors (#9601)
- [ ] 520e43444 (origin/k149) fix function names in comments (#9415)
- [X] 0295fd469 [new feature] promtail: Add a Promtail stage for probabilistic sampling (#7127)
- [ ] 540380fa9 Promtail: fix GeoIP typo (#8783)
- [X] e2ac2d50e Changes from linter (#8769)
- [X] b94c8aaf1 feature: geoip stage in promtail (#3493)
- [X] 43a21eac0 fix(promtail): Make `cri` tags streams aware. (#8497)
- [X] 85b3e5d4a Promtail: Add event log message stage (#8382)
- [ ] ee1aee0af feat(promtail): add more complex options of dropping logs at the drop stage (#6371)
- [ ] 02e0b3ae8 Fix logentry drop to more intuitively match numeric fields (fixes #7011) (#7292)
- [ ] d5b68c025 fix(promtail): Fix cri tags extra new lines. (#7997)
- [X] b773a7a3c promtail: ratelimiting by label (#7597)
- [X] c5911d534 Loki: implement decolorize filter (#7602)

tpaschalis:
Continuing on the items left above

[...]

Issue #5060, opened 2023-09-01
CPU consumption still increased


alxark:
### What's wrong?

Updated to 0.36.0 version but still faces high CPU consumption on nodes without high load. 

```
grafana+ 1854009 31.5  2.9 1567504 57952 ?       Ssl  07:37  17:57 /usr/bin/grafana-agent-flow run --server.http.listen-addr=10.20.1.72:12345 --storage.path=/var/lib/grafana-agent-flow /etc/grafana-agent-flow.river
```

This host contains only 3 monitored containers. Less than 10 messages / second (during dump time it was almost 1 message / sec).

Metrics after several minutes of uptime:
```
# HELP zm_container_lines total number of log lines for container
# TYPE zm_container_lines counter
zm_container_lines{component_id="loki.process.preprocess"} 132
```

### Steps to reproduce

Start several docker containers with message generating.

### System information

Linux 5.15.0-73-generic #80-Ubuntu SMP Mon May 15 15:18:26 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

### Software version

v0.36.0

### Configuration

```text
discovery.docker "containers" {
    host = "unix:///var/run/docker.sock"
}

loki.source.docker "default" {
    host       = "unix:///var/run/docker.sock"

    targets    = discovery.relabel.logs.output

    forward_to = [ loki.process.preprocess.receiver ]
}

discovery.relabel "logs" {
    targets = discovery.docker.containers.targets

    rule {
        source_labels = ["__meta_docker_container_label_job" ]
        separator     = "/"
        target_label  = "job"
        action        = "replace"
    }
}

loki.process "preprocess" {
    forward_to = [ loki.write.loki.receiver ]

    stage.drop {
        older_than  = "10m"
        drop_counter_reason = "record is too old"
    }

    stage.metrics {
        metric.counter {
            name        = "container_lines"
            description = "total number of log lines for container"
            prefix      = "zm_"
            source      = "job"

            match_all         = true
            action            = "inc"
            max_idle_duration = "24h"
        }
    }
}

loki.write "loki" {
    endpoint {
        url = "http://grafana.htz:3100/loki/api/v1/push"
    }
}
```

alxark:
CPU profiling added: 

[cpu.pprof.zip](https://github.com/grafana/agent/files/12497544/cpu.pprof.zip)

rfratto:
[...]

PR #5432, opened 2023-10-10
Add loki.process stage.eventlogmessage

thampiotr:
#### PR Description

Part of https://github.com/grafana/agent/issues/4836
This fell through the cracks for feature parity. 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


component/loki/process/stages/eventlogmessage.go
```diff
@@ -0,0 +1,134 @@
...
+	}()
+	return out
+}
+
+// Process a event log message from extracted with the specified key, adding additional
+// entries into the extracted map
+func (m *eventLogMessageStage) processEntry(extracted map[string]interface{}, key string) error {
+	value, ok := extracted[key]
+	if !ok {
+		if Debug {
```
tpaschalis, line 67: When we initially ported over the Loki components we made an effort to stop using this pattern.

After a while, we saw users report _significant_ CPU % increase out of the blue, and turns out it came from debug logs in busy logs pipelines. 

Let's keep it as a discussion point for the future, but for now, let's remove this to allow debug-level logs to work properly (since IIRC there's no way wired up to set Debug to true in Flow mode).

component/loki/process/stages/eventlogmessage.go
```diff
@@ -0,0 +1,134 @@
...
+type eventLogMessageStage struct {
+	cfg    *EventLogMessageConfig
+	logger log.Logger
+}
+
+// Create a event log message stage, including validating any supplied configuration
+func newEventLogMessageStage(logger log.Logger, cfg *EventLogMessageConfig) Stage {
+	return &eventLogMessageStage{
+		cfg:    cfg,
+		logger: log.With(logger, "component", "stage", "type", "event_log_message"),
```
tpaschalis, line 42: nit: 

```suggestion
		logger: log.With(logger, "component", "stage", "type", "eventlogmessage"),
```

docs/sources/flow/reference/components/loki.process.md
```diff
@@ -259,6 +261,61 @@ stage.drop {
...
+
+stage.eventlogmessage {
+    source = "message"
+    overwrite_existing = true
+}
+```
+
+Given the following log line:
+```
+{"event_id": 1, "Overwritten": "old", "message": "Message type:\r\nOverwritten: new\r\nImage: C:\\Users\\User\\promtail.exe"}
```
tpaschalis, line 303: nit: can we change the docs to avoid mentioning Promtail?

tpaschalis (approved):
LGTM, couple of tiny nits

component/loki/process/stages/eventlogmessage.go
thampiotr: I think that @rfratto wanted to use this new behaviour actually: https://github.com/grafana/agent/issues/5060#issuecomment-1710349195 

And I can see why - we have reports of CPU usage being a problem.

To solve this I think we should use a different approach altogether. I think the issue should be solved by getting rid of `Log(fmt.Sprintf(...))` and instead use `Log("msg", <message>, "sth", <sth>, ...)` so that the call to `Log` function doesn't require to build a new string. And then the very first thing that our loggers should do is to check for `level` and skip the log line if it's below the logger's level. That way the log function can be inlined by compiler and we don't need to surround every call with `if Debug { ... }`. I believe this is how performant logging frameworks do it.

tpaschalis: Yeah, let's not block this just for the debug statement.

clayton-cornell:
The docs are OK as-is (needs work on the passive voice, but all ref docs are this way so not changing here). 

This doesn't account for @tpaschalis ask about removing Promtail from the example.

PR #5434, opened 2023-10-10
Queue-based client for loki.write

thepalbi:
#### PR Description

Right now, the `loki.write` under the hood uses loki remote write, which has two modes: WAL enabled or disabled. WAL needs some work (better segment cleanup, replay and a better batches sending mechanism), however, this effort is **not easy since the both modes share the same remote write client**.

The main purpose of this PR / effort is to have two different remote-write clients, one specific for WAL, and the existing one for non WAL. This will allow us to easily implement the following upcoming changes:
- a marker based old segments cleanup mechanism
- WAL replay
- better send semantics, using something like the queue based metrics remote write
- while not affecting the non-WAL-enabled component, which is the default and non-experimental use case

The diagrams below show, first, the current remote write implementation for when WAL is enabled.

image

The proposed change introduces a "queueing" based client. Fist, by making the remote write client implement [`wal.WriteTo`](https://github.com/grafana/agent/blob/main/component/common/loki/wal/watcher.go#L52), WAL-specific information bleeds into the client layer, making it easy to track from which WAL segment an entry belongs to. This allows us to easily track when a segment's entries are batched, and then sent, so we can more intelligently keep track of consumed WAL segments. 
Also, right now the same remote write client routine does different tasks: builds segments, checks wether they are ready to be sent, encodes them and sends them. This should be split apart. By introducing a queue in the middle, with buffering, that holds batches ready to be sent, we can separate the batch building from the actual send operation into two different routines. Also, in the future, this allows us to easily add more queues, mimicking the `QueueManager` metrics remote write implementation.

image

The idea of this PR (1 in the list below) is to work as the stepping stone for:
1. PR that introduces the queue-based client
2. PR that implement the marker based segment cleanup, and WAL replay https://github.com/grafana/agent/pull/5590
3. Better send semantics through better queueing

To note there's no significant performance difference between the original client, and the queue client, the following benchmark was run writing 100 log entries with similar labels:

```
➜  client git:(pablo/wal-replay) benchstat ~/work/benches/queue-client-1.txt ~/work/benches/simple-client-1.txt
/Users/pablo/work/benches/queue-client-1.txt:4: missing iteration count
/Users/pablo/work/benches/simple-client-1.txt:4: missing iteration count
goos: darwin
goarch: arm64
pkg: github.com/grafana/agent/component/common/loki/client
               │ /Users/pablo/work/benches/queue-client-1.txt │ /Users/pablo/work/benches/simple-client-1.txt │
               │                    sec/op                    │            sec/op             vs base         │
QueueClient-10                                     1.017 ± 0%                     1.017 ± 0%  ~ (p=0.485 n=6)
```

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated
- [x] Add benchmark


component/common/loki/client/client_writeto.go
```diff
@@ -18,16 +18,9 @@ import (
...
-	seriesLock sync.RWMutex
-
-	// seriesSegment keeps track of the last segment in which the series pointed by each key in this map was seen. Keeping
-	// this in a separate map avoids unnecessary contention.
-	//
-	// Even though it doesn't present a difference right now according to benchmarks, it will help when we introduce other
-	// calls from the wal.Watcher to the wal.WriteTo like `UpdateSeriesSegment`.
-	seriesSegment     map[chunks.HeadSeriesRef]int
-	seriesSegmentLock sync.RWMutex
+	series        map[chunks.HeadSeriesRef]model.LabelSet
```
thepalbi, line 21: drive-by: Cleaning up the `seriesCache` locking to use a single lock, since we are not gonna implement `UpdateSeriesSegment` for the time being.

component/common/loki/client/queue_client.go
```diff
@@ -0,0 +1,486 @@
...
+				if b.age() < c.cfg.BatchWait {
+					continue
+				}
+
+				c.enqueue(tenantID, b)
+				// deleting assuming that since the batch expired the wait time, it
+				// hasn't been written for some time
+				delete(c.batches, tenantID)
+			}
+			c.batchesMtx.Unlock()
```
kgeckhart, line 296: Similar question here, would it be safe to put the items to enqueue in to a slice, release the batch mutex, and then enqueue the items from the slice to avoid blocking with the batch mutex held?

component/common/loki/client/queue_client.go
```diff
@@ -0,0 +1,486 @@
...
+		}
+	}
+}
+
+func (c *queueClient) drain(ctx context.Context) {
+	// auxiliary go-routine will enqueue partial batches to use the same sending routine
+	// this helps constraint the whole drain under a single timeout
+	go func() {
+		c.batchesMtx.Lock()
+		defer c.batchesMtx.Unlock()
```
kgeckhart, line 306: Just wondering if this is necessary, I can see how it's generally just for safety purposes. But if `drain` is called and something is still writing to batches that requires this lock then there's a bug right?

component/common/loki/client/queue_client.go
```diff
@@ -0,0 +1,486 @@
...
+		// TODO(thepalbi): Add metric here
+		level.Debug(c.logger).Log("msg", "series for entry not found")
+	}
+	return nil
+}
+
+func (c *queueClient) appendSingleEntry(lbs model.LabelSet, e logproto.Entry, segment int) {
+	// TODO: can I make this locking more fine grained?
+	c.batchesMtx.Lock()
+	defer c.batchesMtx.Unlock()
```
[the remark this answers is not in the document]
thepalbi: true!

component/common/loki/client/queue_client.go
thepalbi: Mmm this is an interesting question. Two other possible actors interact with `c.batches`:
1. The old batch routine, handled by the wait group above the `drain` call, therefore safe.
2. Someone from outside of the queue client scope, calling `AppendEntries`. We have no control over this since it's the watcher that does the call, hence, we need first stop the watcher and then the client to be sure it ain't gonna call `AppendEntries` in the stopping-queue-client.

We can definitely assume this stopping order, or leave the block for safety.

component/common/loki/client/queue_client.go
thepalbi: absolutely yes!

component/common/loki/client/queue_client.go
```diff
@@ -0,0 +1,486 @@
...
+		_ = nb.addFromWAL(lbs, e, segment)
+		c.batches[tenantID] = nb
+		c.initBatchMetrics(tenantID)
+		return
+	}
+
+	// If adding the entry to the batch will increase the size over the max
+	// size allowed, we do send the current batch and then create a new one
+	if batch.sizeBytesAfter(e.Line) > c.cfg.BatchSize {
+		c.enqueue(tenantID, batch)
```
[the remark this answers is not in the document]
thepalbi: Yes we can, just keep in mind there's a single caller to `queueClient.AppendEntries`, which is the watcher. Therefore:
- There can't be two simultaneous calls to `AppendEntries`
- The only other actor we are unblocking by releasing this lock early is the old batches sending go routine

Now I'm wondering if we could remove the concurrency of having to check over all batches if they are old enough 🤔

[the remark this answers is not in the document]
kgeckhart: Ah I see what you are saying, the WAL watcher is single threaded so releasing the batch mutex earlier doesn't unblock the watcher. If `runSendOldBatches` is no longer holding the mutex while it calls `enqueue` for the old batches then the watcher blocking with the mutex might not be hurting anything. Generally though, it feels less than ideal to call something that's going to block while holding a mutex if you don't really need to.

component/common/loki/client/manager_test.go
```diff
@@ -304,23 +307,3 @@ func TestManager_WALDisabled_MultipleConfigs(t *testing.T) {
 	}
 	require.Len(t, seenEntries, expectedTotalLines)
 }
-
-func TestManager_StopClients(t *testing.T) {
```
thepalbi, line 308: drive-by: test removed since it was asserting / testing through the internal representation of the manager, which is a smell

ptodev (commented):
Thank you for this change! It feels like a step in the right direction. I have a few questions:
* If the Agent config gets reloaded and we switch the `wal` block from `enabled = true` to `enabled = false` (or vice versa), I guess there are no issues?
* I suppose the previous implementation was more synchronous? Hence, if the channel in the new implementation is full, there are no extra delays and  having the channel can only help?
* I guess the idea behind the "send batch work pool" is to have something similar to the Prometheus remote write sharding, except that here we would shard based on log streams?

component/common/loki/client/queue_client.go
```diff
@@ -0,0 +1,584 @@
...
+
+	// TODO: can I make this locking more fine grained?
+	c.batchesMtx.Lock()
+
+	batch, ok := c.batches[tenantID]
+
+	// If the batch doesn't exist yet, we create a new one with the entry
+	if !ok {
+		nb := newBatch(c.maxStreams)
+		_ = nb.addFromWAL(lbs, e)
```
mattdurham, line 309: The assumption here is that a newly created batch should never error when adding a labelset? Maybe a comment about that?

component/common/loki/client/queue_client.go
```diff
@@ -0,0 +1,584 @@
...
+
+		if err == nil {
+			c.metrics.sentBytes.WithLabelValues(c.cfg.URL.Host).Add(bufBytes)
+			c.metrics.sentEntries.WithLabelValues(c.cfg.URL.Host).Add(float64(entriesCount))
+
+			return
+		}
+
+		// Only retry 429s, 500s and connection-level errors.
+		if status > 0 && !batchIsRateLimited(status) && status/100 != 5 {
```
mattdurham, line 460: Does loki send a retry after header we can use?

component/common/loki/client/batch.go
```diff
@@ -110,8 +136,8 @@ func (b *batch) sizeBytes() int {
 
 // sizeBytesAfter returns the size of the batch after the input entry
 // will be added to the batch itself
-func (b *batch) sizeBytesAfter(entry loki.Entry) int {
-	return b.bytes + len(entry.Line)
+func (b *batch) sizeBytesAfter(line string) int {
+	return b.bytes + len(line)
```
mattdurham, line 140: nit: b.bytes sounds to me like a byte array as opposed to lengthOfBytes

component/common/loki/client/client_writeto.go
```diff
@@ -70,7 +59,7 @@ func (c *clientWriteTo) SeriesReset(segmentNum int) {
 	}
 }
 
-func (c *clientWriteTo) AppendEntries(entries wal.RefEntries) error {
+func (c *clientWriteTo) AppendEntries(entries wal.RefEntries, segment int) error {
```
mattdurham, line 62: segment is unused

mattdurham (changes requested):
A few questions and clarifications but overall looks reasonable. Would love a benchmark to.

component/common/loki/client/manager.go
```diff
@@ -179,7 +187,38 @@ func (m *Manager) Stop() {
...
 	}
 	// close clients
-	for _, c := range m.clients {
+	for _, c := range m.stoppableClients {
 		c.Stop()
 	}
 }
+
+// GetClientName computes the specific name for each client config. The name is either the configured Name setting in Config,
+// or a hash of the config as whole, this allows us to detect repeated configs.
```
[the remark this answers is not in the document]
thepalbi: Yes, this is actually a refactor moving the naming logic out of the previous client . The [current implementation](https://github.com/grafana/agent/blob/main/component/common/loki/client/client.go#L197) uses this hashing approach, so I guess it's ok.

component/loki/write/types.go
```diff
@@ -70,6 +71,21 @@ func (r *EndpointOptions) Validate() error {
 	return nil
 }
 
+// QueueConfig controls how the queue logs remote write client is configured. Note that this client is only used when the
+// loki.write component has WAL support enabled.
+type QueueConfig struct {
+	Capacity     int           `river:"capacity,attr,optional"`
+	DrainTimeout time.Duration `river:"drain_timeout,attr,optional"`
```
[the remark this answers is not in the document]
thepalbi: Yes! My plan is to add them once we are ok with the impl

component/loki/write/types.go
```diff
@@ -70,6 +71,21 @@ func (r *EndpointOptions) Validate() error {
...
+}
+
+// SetToDefault implements river.Defaulter.
+func (q *QueueConfig) SetToDefault() {
+	*q = QueueConfig{
+		Capacity:     10,
```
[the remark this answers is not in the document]
thepalbi: Mmm true, for some reason I was assuming the default in `prometheus.remote_write` was this, but it's not. I think an initial heuristic could be something like `capacity = buffer capacity in bytes / batch size`, with batch size the one configured [here](https://grafana.com/docs/agent/next/flow/reference/components/loki.write/#endpoint-block). In that way the user could configure this channel size thinking in bytes in memory (obviously assuming a worst case where each batch enqueued is full). wdyt?

component/common/loki/client/queue_client.go
```diff
@@ -0,0 +1,584 @@
...
+		return nil, errors.New("client needs target URL")
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+
+	c := &queueClient{
+		logger:       log.With(logger, "component", "client", "host", cfg.URL.Host),
+		cfg:          cfg,
+		metrics:      metrics,
+		drainTimeout: cfg.Queue.DrainTimeout, // make this configurable
```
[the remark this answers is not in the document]
thepalbi: Yes! removing comment!

component/common/loki/client/client_writeto.go
thepalbi: Yes, but it's part of the interface that the method attests to be a `wal.WriteTo`. Also, in the interface, the segments in the `AppendEntries` is needed for https://github.com/grafana/agent/pull/5590

thepalbi:
@mattdurham added a benchmark (the one in queue_client_test.go), which I ran with the queue client and the old one. Results here:
```
➜  client git:(pablo/wal-replay) benchstat ~/work/benches/queue-client-1.txt ~/work/benches/simple-client-1.txt
/Users/pablo/work/benches/queue-client-1.txt:4: missing iteration count
/Users/pablo/work/benches/simple-client-1.txt:4: missing iteration count
goos: darwin
goarch: arm64
pkg: github.com/grafana/agent/component/common/loki/client
               │ /Users/pablo/work/benches/queue-client-1.txt │ /Users/pablo/work/benches/simple-client-1.txt │
               │                    sec/op                    │            sec/op             vs base         │
QueueClient-10                                     1.017 ± 0%                     1.017 ± 0%  ~ (p=0.485 n=6)
```

Outcome, no significant difference. Mainly because I think the improvements in this PR will help in worst case scenarios with multi tenancy, and failure scenarios on the remote write destination.

component/common/loki/client/client_writeto.go
mattdurham: Gotcha, I think  some linkers will pick it up if its the name is not an underscore. I think my local linter did but not what we run for drone.

component/loki/write/types.go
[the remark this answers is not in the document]
mattdurham: I think if they can configure it based on overall memory usage thats ideal.

component/common/loki/client/queue_client.go
thepalbi: Just asked, no header from Loki [thread](https://raintank-corp.slack.com/archives/C9T1FLN9K/p1698243835822239)

thepalbi:
@ptodev  responding inline:

> Thank you for this change! It feels like a step in the right direction. I have a few questions:
> 
> * If the Agent config gets reloaded and we switch the `wal` block from `enabled = true` to `enabled = false` (or vice versa), I guess there are no issues?

No issues, but the following would happen:
1. the manager would be stopped, stopping and draining the queue client
2. the non wal client would be enabled

The only thing I can imagine would happen here, is if the watcher is not completely over the top of the current segment, some data would be missed in that segment. But since turning on and off WAL is not an expected use case, I think we are good with that behaviour.  Regardless, with https://github.com/grafana/agent/pull/5590 WAL would be replayed and this could be handled ok.

> * I suppose the previous implementation was more synchronous? Hence, if the channel in the new implementation is full, there are no extra delays and  having the channel can only help?

Exactly, the previous implementation was all unbuffered channels connected together, therefore, synchronous. Having a buffered channel here allows us to add more asyncronicity which unblock the WAL reader end from the remote write client.

> * I guess the idea behind the "send batch work pool" is to have something similar to the Prometheus remote write sharding, except that here we would shard based on log streams?

Yes, the idea is on the third iteration mentioned  in the PR description (after wal replay is implemented, which is the other missing piece), implement something like sharding. Now, if the sharding would be done based on log streams, tenants, or a combination of both is to be decided. I think a combination of both makes sense, but we can leave that for later.

component/loki/write/types.go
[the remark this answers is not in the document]
thepalbi: Awesome, will implement that then. `loki.write` has a default batch size of 1MiB. I think a default capacity of 100 MiB (which translated in the default case to a channel of size 100) is good. wdyt?

thepalbi:
@mattdurham @ptodev added more test cases (different num of entries sent, with our without batching). Also added cases to the benchmark, with default batching settings and 100k entries, and a small test case, and re-run with `benchmem`:

```
goos: darwin
goarch: arm64
pkg: github.com/grafana/agent/component/common/loki/client
                                                                    │   queue    │             regular              │
                                                                    │   sec/op   │   sec/op    vs base              │
ClientImplementations/100k_entries,_100_series,_default_batching-10   4.434 ± 2%   4.716 ± 6%  +6.36% (p=0.004 n=6)
ClientImplementations/100_entries,_single_series,_no_batching-10      1.001 ± 0%   1.016 ± 0%  +1.50% (p=0.002 n=6)
geomean                                                               2.107        2.189       +3.90%

                                                                    │    queue     │              regular               │
                                                                    │     B/op     │     B/op      vs base              │
ClientImplementations/100k_entries,_100_series,_default_batching-10   771.9Mi ± 0%   760.1Mi ± 0%  -1.53% (p=0.002 n=6)
ClientImplementations/100_entries,_single_series,_no_batching-10      842.7Ki ± 5%   830.0Ki ± 2%       ~ (p=0.058 n=6)
geomean                                                               25.20Mi        24.82Mi       -1.52%

                                                                    │    queue    │              regular              │
                                                                    │  allocs/op  │  allocs/op   vs base              │
ClientImplementations/100k_entries,_100_series,_default_batching-10   10.80M ± 0%   10.50M ± 0%  -2.77% (p=0.002 n=6)
ClientImplementations/100_entries,_single_series,_no_batching-10      11.25k ± 0%   10.94k ± 0%  -2.75% (p=0.002 n=6)
geomean                                                               348.7k        339.0k       -2.76%
```

I think at little memory expense (with the 100MiB capacity setting) we get some less CPU stress, and better semantics for write failure scenarios considering we have the buffered send queue.

component/loki/write/types.go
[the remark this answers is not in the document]
thepalbi: Added docs

thepalbi:
@mattdurham @ptodev I think all comments are addressed, can you take another look?

component/loki/write/types.go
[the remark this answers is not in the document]
mattdurham: That feels a bit high for generic usage. IE the most common case is colocated on a server with other noisy processes. Since it is disk backed what about pulling that back to 10 instead of 100?

mattdurham:
Looking good the only comment I have is on the default, I would rather a lower default memory usage. In the case the memory buffer fills up ogs are still written to the wal and once the batches are enqueued in the case of a network issue then logs continue going to the disk until the outage is resolved?

component/loki/write/types.go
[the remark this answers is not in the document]
thepalbi: Yes, for sure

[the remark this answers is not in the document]
thepalbi: Changed to `10MiB`. Once callout is if `buffer capacity in bytes < batch size`, there will be no buffering, and the queue will act similar to the regular client's behaviour (blocking), but I think that's ok. Maybe worth adding a callout in the docs?

thepalbi:
@mattdurham 
> In the case the memory buffer fills up ogs are still written to the wal and once the batches are enqueued in the case of a network issue then logs continue going to the disk until the outage is resolved

Yes, since the wal writer is completely decoupled from the sending end, it will keep writing into wal even though the sender end is completely blocked.

mattdurham (approved):
Awesome work.

thepalbi:
@clayton-cornell if you have the time, can you take a look at the edited docs?

docs/sources/flow/reference/components/loki.write.md
```diff
@@ -128,6 +130,18 @@ enabled, the retry mechanism will be governed by the backoff configuration speci
...
+### queue_config block (experimental)
+
+The optional `queue_config` block configures, when WAL is enabled (see [Write-Ahead block](#wal-block-experimental)), how the
+underlying client queues batches of logs to be sent to Loki.
+
+The following arguments are supported:
+
+| Name            | Type       | Description                                                                                                                                                                      | Default | Required |
+| --------------- | ---------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------- | -------- |
+| `capacity`      | `string`   | Controls the size of the underlying send queue buffer. This setting should be considered as a worst case scenario of memory consumption, in which all enqueued batches are full. | `10MiB`  | no       |
```
clayton-cornell, line 142:
```suggestion
| `capacity`      | `string`   | Controls the size of the underlying send queue buffer. This setting should be considered a worst-case scenario of memory consumption, in which all enqueued batches are full. | `10MiB`  | no       |
```

docs/sources/flow/reference/components/loki.write.md
```diff
@@ -128,6 +130,18 @@ enabled, the retry mechanism will be governed by the backoff configuration speci
...
+| `drain_timeout` | `duration` | Configures the maximum time the client can take to drain the send queue, upon shutdown. During that time it will enqueue pending batches, and drain the send queue sending each. | `"1m"`  | no       |
```
clayton-cornell, line 143:
```suggestion
| `drain_timeout` | `duration` | Configures the maximum time the client can take to drain the send queue upon shutdown. During that time, it will enqueue pending batches and drain the send queue sending each. | `"1m"`  | no       |
```

clayton-cornell (approved):
Some really minor tweaks to the docs.

mattdurham:
@thepalbi ping when you make the docs changes and I will merge.

PR #5452, opened 2023-10-12
Improve loggers performance

thampiotr:
#### PR Description

It's recommended too review the first two commits of this PR separately:
* [First one](https://github.com/grafana/agent/pull/5452/commits/c9d638b3917ecac138afe53c60339100842a00ce) adds benchmarks and includes the results in a comment
* [Second one](https://github.com/grafana/agent/pull/5452/commits/3ed08cfd5598a7d3f336f7af529083d314b5db62) adds a few optimisations for the logging and includes the new benchmark results in a comment.

Summary of optimisations:
* Look for the log level first in log adapter to avoid costly allocations if the record will not be logged
* Implement our own `level` package that can replace gokit's `level.Debug(logger)...` - our `level` package is level-aware and will be checking whether given level is enabled in the underlying logger (if it can, opportunistically. The idea is that we can replace the import with our own `level` and should get performance boost. 

In the long run we should still move to `slog` as it's more performant and simpler.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/flow/logging/level/level.go
+++ b/pkg/flow/logging/level/level.go
@@ -1,0 +1,51 @@
+package level
+
+import (
+	"context"
+	"log/slog"
+
+	"github.com/go-kit/log"
+	"github.com/grafana/agent/pkg/flow/logging"
+)
+
+const (
+	levelKey = "level"
+)
+
+// Error returns a logger that includes a Key/ErrorValue pair.
+func Error(logger log.Logger) log.Logger {
+	return toLevel(logger, "error", slog.LevelError)
+}
+
+// Warn returns a logger that includes a Key/WarnValue pair.
+func Warn(logger log.Logger) log.Logger {
+	return toLevel(logger, "warn", slog.LevelWarn)
+}
+
+// Info returns a logger that includes a Key/InfoValue pair.
+func Info(logger log.Logger) log.Logger {
+	return toLevel(logger, "info", slog.LevelInfo)
+}
+
+// Debug returns a logger that includes a Key/DebugValue pair.
+func Debug(logger log.Logger) log.Logger {
+	return toLevel(logger, "debug", slog.LevelDebug)
+}
+
+func toLevel(logger log.Logger, level string, slogLevel slog.Level) log.Logger {
+	switch l := logger.(type) {
+	case logging.EnabledAware:
+		if !l.Enabled(context.Background(), slogLevel) {
+			return disabledLogger
+		}
+	}
+	return log.WithPrefix(logger, levelKey, level)
+}
+
+var disabledLogger = &noopLogger{}
+
+type noopLogger struct{}
+
+func (d *noopLogger) Log(_ ...interface{}) error {
+	return nil
+}
--- a/internal/slogadapter/gokit.go
+++ b/internal/slogadapter/gokit.go
@@ -19,25 +19,58 @@ type slogAdapter struct {
 	h slog.Handler
 }
 
+var (
+	_ log.Logger = (*slogAdapter)(nil)
+)
+
+// Enabled implements logging.EnabledAware interface.
+func (sa slogAdapter) Enabled(ctx context.Context, l slog.Level) bool {
+	return sa.h.Enabled(ctx, l)
+}
+
 func (sa slogAdapter) Log(kvps ...interface{}) error {
-	// We don't know what the log level or message are yet, so we set them to
-	// defaults until we iterate through kvps.
-	//
+	// Find the log level first, starting with the default.
+	recordLevel := slog.LevelInfo
+	for i := 0; i < len(kvps); i += 2 {
+		if kvps[i] == level.Key() {
+			if i+1 < len(kvps) {
+				value := kvps[i+1]
+				levelValue, _ := value.(level.Value)
+				switch levelValue {
+				case level.DebugValue():
+					recordLevel = slog.LevelDebug
+				case level.InfoValue():
+					recordLevel = slog.LevelInfo
+				case level.WarnValue():
+					recordLevel = slog.LevelWarn
+				case level.ErrorValue():
+					recordLevel = slog.LevelError
+				}
+				break
+			}
+		}
+	}
+
+	// Do not build the record if the level is not enabled.
+	if !sa.h.Enabled(context.Background(), recordLevel) {
+		return nil
+	}
+
 	// Since there's a pattern of wrapping loggers at different depths in go-kit,
 	// we can't consistently know who the caller is, so we set no value for pc
 	// here.
-	rec := slog.NewRecord(time.Now(), slog.LevelInfo, "", 0)
+	rec := slog.NewRecord(time.Now(), recordLevel, "", 0)
 
 	for i := 0; i < len(kvps); i += 2 {
 		var key, value any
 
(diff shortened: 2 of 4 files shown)


thampiotr:
Here's a `benchstat` comparison - thanks @rfratto for tool suggestion
```
benchstat old.txt new.txt
goos: darwin
goarch: arm64
pkg: github.com/grafana/agent/pkg/flow/logging
                                   │   old.txt    │               new.txt               │
                                   │    sec/op    │   sec/op     vs base                │
Logging_NoLevel_Prints-8              1.552µ ± 1%   1.542µ ± 0%   -0.64% (p=0.002 n=10)
Logging_NoLevel_Drops-8              334.20n ± 2%   47.18n ± 2%  -85.88% (p=0.000 n=10)
Logging_GoKitLevel_Drops_Sprintf-8    484.1n ± 0%   376.7n ± 5%  -22.19% (p=0.000 n=10)
Logging_GoKitLevel_Drops-8            503.8n ± 0%   201.2n ± 1%  -60.06% (p=0.000 n=10)
Logging_GoKitLevel_Prints-8           1.735µ ± 0%   1.741µ ± 0%   +0.35% (p=0.003 n=10)
Logging_Slog_Drops-8                  50.51n ± 1%   37.04n ± 0%  -26.67% (p=0.000 n=10)
Logging_Slog_Prints-8                 1.194µ ± 1%   1.169µ ± 1%   -2.01% (p=0.000 n=10)
Logging_FlowLevel_Drops-8                           74.12n ± 1%
Logging_FlowLevel_Prints-8                          1.740µ ± 0%
geomean                               539.1n        335.3n       -39.00%

                                   │  old.txt   │               new.txt                │
                                   │    B/op    │    B/op     vs base                  │
Logging_NoLevel_Prints-8             408.0 ± 0%   392.0 ± 0%   -3.92% (p=0.000 n=10)
Logging_NoLevel_Drops-8              64.00 ± 0%   16.00 ± 0%  -75.00% (p=0.000 n=10)
Logging_GoKitLevel_Drops_Sprintf-8   464.0 ± 0%   352.0 ± 0%  -24.14% (p=0.000 n=10)
Logging_GoKitLevel_Drops-8           528.0 ± 0%   480.0 ± 0%   -9.09% (p=0.000 n=10)
Logging_GoKitLevel_Prints-8          873.0 ± 0%   873.0 ± 0%        ~ (p=1.000 n=10) ¹
Logging_Slog_Drops-8                 32.00 ± 0%   16.00 ± 0%  -50.00% (p=0.000 n=10)
Logging_Slog_Prints-8                64.00 ± 0%   47.00 ± 0%  -26.56% (p=0.000 n=10)
Logging_FlowLevel_Drops-8                         176.0 ± 0%
Logging_FlowLevel_Prints-8                        873.0 ± 0%
geomean                              196.8        168.1       -32.97%
¹ all samples are equal

                                   │   old.txt   │               new.txt                │
                                   │  allocs/op  │ allocs/op   vs base                  │
Logging_NoLevel_Prints-8              13.00 ± 0%   12.00 ± 0%   -7.69% (p=0.000 n=10)
Logging_NoLevel_Drops-8               7.000 ± 0%   1.000 ± 0%  -85.71% (p=0.000 n=10)
Logging_GoKitLevel_Drops_Sprintf-8   11.000 ± 0%   9.000 ± 0%  -18.18% (p=0.000 n=10)
Logging_GoKitLevel_Drops-8           11.000 ± 0%   6.000 ± 0%  -45.45% (p=0.000 n=10)
Logging_GoKitLevel_Prints-8           17.00 ± 0%   17.00 ± 0%        ~ (p=1.000 n=10) ¹
Logging_Slog_Drops-8                  2.000 ± 0%   1.000 ± 0%  -50.00% (p=0.000 n=10)
Logging_Slog_Prints-8                 5.000 ± 0%   3.000 ± 0%  -40.00% (p=0.000 n=10)
Logging_FlowLevel_Drops-8                          3.000 ± 0%
Logging_FlowLevel_Prints-8                         17.00 ± 0%
geomean                               7.871        4.919       -43.82%
¹ all samples are equal
```

pkg/flow/logging/level/level.go
```diff
@@ -0,0 +1,50 @@
...
+
+// Debug returns a logger that includes a Key/DebugValue pair.
+func Debug(logger log.Logger) log.Logger {
+	return toLevel(logger, "debug", slog.LevelDebug)
+}
+
+func toLevel(logger log.Logger, level string, slogLevel slog.Level) log.Logger {
+	if adapter, ok := logger.(logging.EnabledAware); ok {
+		if !adapter.Enabled(context.Background(), slogLevel) {
+			return disabledLogger
```
thampiotr, line 38: The edge case here is that: `level.Error(level.Debug(logger))` (i.e. setting level twice) will not work. I don't think that should be a problem given how we use the loggers right now.

pkg/flow/logging/level/level.go
```diff
@@ -0,0 +1,50 @@
...
+	return toLevel(logger, "info", slog.LevelInfo)
+}
+
+// Debug returns a logger that includes a Key/DebugValue pair.
+func Debug(logger log.Logger) log.Logger {
+	return toLevel(logger, "debug", slog.LevelDebug)
+}
+
+func toLevel(logger log.Logger, level string, slogLevel slog.Level) log.Logger {
+	if adapter, ok := logger.(logging.EnabledAware); ok {
```
tpaschalis, line 36: IIRC each time an interface gets resolved it needs an allocation (maybe that's why even the `Logging_Slog_Drops` case allocates).

I'm curious, what does a `switch logger.(type)` performance look like? Is that the same here?

pkg/flow/logging/level/level.go
```diff
@@ -0,0 +1,50 @@
+package level
```
tpaschalis, line 1: We're not using this `level` package anywhere yet, right? Is the intent to wire it as a substitute to the go-kit level package?

thampiotr: Yep! I want to make it a drop-in replacement so we just change the import statement.

tpaschalis: Are we changing it in this PR, or do you want to leave it as a follow-up?

pkg/flow/logging/level/level.go
thampiotr: I have switched to the implementation you've suggested, but there are no improvements in performance. I will leave it, however, since I like the switch statement more :) 

Re: `Logging_Slog_Drops` I did manage to reduce one allocation in the benchmark code due to `time.Since(now)`, so now it does 0 allocations.

tpaschalis: Hm, interesting tidbit to keep in mind!

I wouldn't expect that `x.(type)` would be the same performance if it didn't have to look up an interface value. No personal preference on the stylistic front, so feel free to use whichever you prefer.

tpaschalis (approved):

pkg/flow/logging/level/level.go
thampiotr: I'll do a follow-up. The changes here also benefit the code that doesn't use the new `level` package, thanks to the log adapter checking for the level first.

PR #5453, opened 2023-10-12
Update Prometheus dependency to v2.47.2

tpaschalis:
This PR reuses our fork to update our dependency to Prometheus v2.47.1, plus some cherry-picked commits that a) improve the retrying logic within Prometheus' remote write and b) fix a panic with Native Histograms handling.

Once these are available in the upcoming minor Prometheus release, we should remove the fork once more..


krajorama (approved):
LGTM, I see my fixes in there. But otherwise hard to judge without vendoring.

tpaschalis:
Bumping this to the next patch version v2.47.2 since it contains the fix on prometheus/prometheus#12838.

Issue #4973, opened 2023-08-28
Clustering: Implement a smoke test


tpaschalis:

tpaschalis:
> Are you thinking about a smoke test in the style of our old smoke tests, or a set of Go unit tests?

I missed your comment, sorry.

No, I wasn't thinking about our old smoke tests. First, a high-level Go-orchestrated test that starts some Flow controllers and makes sure that the target distribution between them is correct, and then more tests covering other behavior.

PR #5454, opened 2023-10-12
clustering: fix handling of empty join-addresses flag

### LINKED ISSUE #5449: clustering: fix condition for mutual exclusivity of join-addresses and discover-peers
To populate the JoinPeers field we're using an expression like `strings.Split(fr.clusterJoinAddr, ",")`. In case of an empty string, this will lead to a slice with an empty element (instead of an empty slice as was the original intent).

This breaks the mutually exclusive comparison between JoinPeers and DiscoverPeers.

https://github.com/grafana/agent/blob/a0c55a76a467c957a0ea2419d806aab076f3e5bd/cmd/internal/flowmode/cluster_builder.go#L86-L88

Leading to errors like the following
```
ts=2023-10-12T11:02:01.653094Z level=info "boringcrypto enabled"=false
Error: at most one of join peers and discover peers may be set
```

tpaschalis:
#### PR Description

We recently almost had this bite us when porting the cadvisor integration to a new component. 

`strings.Split` returns the input string itself if the separator is not found; this meant that for an empty flag, `strings.Split("", ",")` returned a 1-element slice with the empty string, and _not_ an empty slice.

Fixes #5449 

#### Notes to the Reviewer

Nothing to note.

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Tests updated

### DIFF
--- a/cmd/internal/flowmode/cluster_builder_test.go
+++ b/cmd/internal/flowmode/cluster_builder_test.go
@@ -1,0 +1,18 @@
+package flowmode
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestBuildClusterService(t *testing.T) {
+	opts := clusterOptions{
+		JoinPeers:     []string{"foo", "bar"},
+		DiscoverPeers: "provider=aws key1=val1 key2=val2",
+	}
+
+	cs, err := buildClusterService(opts)
+	require.Nil(t, cs)
+	require.EqualError(t, err, "at most one of join peers and discover peers may be set")
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@ Main (unreleased)
 
 - Fixed an issue where `loki.process` validation for stage `metric.counter` was 
   allowing invalid combination of configuration options. (@thampiotr)
+
+- Fix the handling of the `--cluster.join-addresses` flag causing an invalid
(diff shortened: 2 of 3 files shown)


cmd/internal/flowmode/cluster_builder_test.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestBuildClusterService(t *testing.T) {
+	opts := clusterOptions{
+		JoinPeers:     []string{"foo", "bar"},
```
thampiotr, line 11: Shouldn't we have a test with `JoinPeers` empty string somewhere, or setting a flag to it?

tpaschalis: Yes, pushing it.

tpaschalis: (sorry, didn't post the comment beforehand).

After taking a look we don't have a good place for putting tests related to flag parsing. I'd like to (sooner or later) move ahead with #4973 which would cover this. I'm merging this to move ahead with the patch release.

juyongli:
Hi, I am the one who reported this defect #108849 grafana-agent-flow clustering mode error. Seems you guys have fixed it. Thanks.

tpaschalis:
You're welcome @juyongli! Let us know how clustering works for you and any other feedback you may have!

PR #5456, opened 2023-10-12
docs: small typo in an example

mlcdf:
Hello :)

The exported field for local.file is `content` and not `contents`. :)

https://grafana.com/docs/agent/latest/flow/reference/components/local.file/#exported-fields


spartan0x117 (approved):
Thanks for catching!

PR #5463, opened 2023-10-12
upgrade packages in dockerfile

captncraig:
This resolves a number of (afaik unexploitable) cves.

Mantic does not have definitions yet, but there are updated libc6 packages. This runs `apt-get upgrade -y` in the dockerfile.


ptodev (approved):

ptodev:
My only concern with this is that it could update the packages to different versions every time, making the build less reproducible. If we can update just the one package to the version we need, maybe it'd be better.

captncraig:
That's a valid point, but not as long as we are building off plain `ubuntu:lunar` which also receives periodic updates. And these CVE reports come in at random intervals too. I'd be open to solving the problem, but I'd rather close these cve issues first.

PR #5466, opened 2023-10-12
helm: add secrets and configmaps to rbac

captncraig:
cc @petewall


petewall (approved):
Thanks! This is great.

PR #5471, opened 2023-10-13
"apt-get upgrade" in agentctl and operator docker images

ptodev:
#5463 updated the `grafana-agent` dockerfile, but not the other two docker files. This PR also updates the other two files for consistency, and also to make sure they are not flagged for CVEs.


tpaschalis (approved):
LGTM for now, let's roll this back once Ubuntu vendors the package versions that we want.

PR #5474, opened 2023-10-13
WIP: Pipeline tests

thampiotr:
#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/component/prometheus/remotewrite/remote_write.go
+++ b/component/prometheus/remotewrite/remote_write.go
@@ -107,6 +107,9 @@ func New(o component.Options, c Arguments) (*Component, error) {
 			if localID == 0 {
 				prometheus.GlobalRefMapping.GetOrAddLink(res.opts.ID, uint64(newRef), l)
 			}
+			//TODO(thampiotr): We can expose this as `notify_on_append` option - it can be used to reduce latency in
+			// the pipeline if users want to, with a trade-off being more CPU and I/O.
+			remoteStore.Notify()
 			return globalRef, nextErr
 		}),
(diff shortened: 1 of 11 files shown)


component/prometheus/remotewrite/remote_write.go
```diff
@@ -107,6 +107,9 @@ func New(o component.Options, c Arguments) (*Component, error) {
 			if localID == 0 {
 				prometheus.GlobalRefMapping.GetOrAddLink(res.opts.ID, uint64(newRef), l)
 			}
+			//TODO(thampiotr): We can expose this as `notify_on_append` option - it can be used to reduce latency in
+			// the pipeline if users want to, with a trade-off being more CPU and I/O.
+			remoteStore.Notify()
```
thampiotr, line 112: Adding this shortens the time it takes to run prometheus tests from 15s to 5s. But I don't see a reason why these tests wouldn't run in 200ms each, we just may need to do some digging.

cmd/internal/pipelinetests/testdata/target_scrape_and_write_otel_conversion.river
```diff
@@ -0,0 +1,47 @@
+logging {
+  level  = "debug"
+  format = "logfmt"
+}
+
+// Scrape test target and send to OTEL converter
```
thampiotr, line 6: Example of a test that touches on quite a lot of components from `prometheus` and `otelcol` namespaces, testing the agent as a whole. We can quite easily go further with relabels etc. 

The goal is not to make an exhaustive test, but to verify that the pipeline works and data flows through it really.

thampiotr:
This is not worked on right now, but we may bring it back in the future. Closing for now.

PR #5475, opened 2023-10-13
Integration tests proposal

wildum:
#### PR Description

This PR introduces integration testing to the agent for flow.

#### PR Checklist

- [x] Documentation added


integration-tests/docker-compose.yaml
ptodev, line 1: Why do we need to run Otel Collector? I also don't see why we need `demo-client` and `demo-server`? Can we not just use tools like avalanche and [xk6-client-tracing](https://github.com/grafana/xk6-client-tracing) for generating fake telemetry, and then use the Agent for everything else?

ptodev (commented):
Thank you for working on this. I added a few comments. It would also be nice to have documentation on when we should use unit tests vs when we should use integration tests.

In the future, I also wonder if we can use k6 for any of the tests. We don't have to do it for this PR though.

integration-tests/docker-compose.yaml
wildum: I initially wanted to test exemplars but I haven't found a way to correctly query it from Mimir yet.

Makefile
```diff
@@ -172,7 +172,7 @@ lint: agentlint
 # We have to run test twice: once for all packages with -race and then once
 # more without -race for packages that have known race detection issues.
 test:
-	$(GO_ENV) go test $(GO_FLAGS) -race ./...
+	$(GO_ENV) go test $(GO_FLAGS) -race $(shell go list ./... | grep -v /integration-tests/)
```
thampiotr, line 175: IMO `make test` should run all the tests to help me diagnose a failing PR check? 

We could have a `make integration-test` target if the above is not a good idea. The only thing is that I'd like the CI job to clearly state what to run locally to repro it.

integration-tests/common/log.go
```diff
@@ -0,0 +1,16 @@
...
+
+import "encoding/json"
+
+type LogResponse struct {
+	Status string    `json:"status"`
+	Data   []LogData `json:"data"`
+}
+
+type LogData struct {
+	Filename string `json:"filename"`
```
thampiotr, line 11: Should we go as far as checking the log line contents? 🤔 
I think that would be okay, but I wouldn't go any further than that.

integration-tests/common/metric.go
```diff
@@ -0,0 +1,26 @@
...
+	Result     []MetricResult `json:"result"`
+}
+
+type MetricResult struct {
+	Metric Metric `json:"metric"`
+}
+
+type Metric struct {
+	Job  string `json:"job"`
+	Name string `json:"__name__"`
```
thampiotr, line 21: Similar, seems like a good idea to also check the metric value just for completeness. We can assert with something like `GreaterThan`.

integration-tests/configs/mimir/mimir.yaml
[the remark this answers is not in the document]
thampiotr: That sounds like a good idea.

integration-tests/configs/otel-gen-client/Dockerfile
```diff
@@ -0,0 +1,11 @@
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+FROM golang:1.21 as build
+WORKDIR /app/
+COPY go.mod go.sum ./
+RUN go mod download
+COPY ./integration-tests/configs/otel-gen-client/ ./
+RUN CGO_ENABLED=0 go build -o main main.go
```
thampiotr, line 8: Nit: do you need `CGO_ENABLED=0` here? I don't see immediately why we'd need that.

integration-tests/configs/otel-gen-client/Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+FROM alpine:3.18
+COPY --from=build /app/main /app/main
+CMD ["/app/main"]
```
thampiotr, line 11: Nit: no new line at the end of file

integration-tests/configs/otel-gen-client/main.go
```diff
@@ -0,0 +1,201 @@
...
+		requestLatency.Record(ctx, latencyMs, metric.WithAttributes(commonLabels...))
+		requestCount.Add(ctx, 1, metric.WithAttributes(commonLabels...))
+
+		fmt.Printf("Latency: %.3fms\n", latencyMs)
+		time.Sleep(time.Duration(1) * time.Second)
+	}
+}
+
+func makeRequest(ctx context.Context) {
+	demoServerAddr, ok := os.LookupEnv("DEMO_SERVER_ENDPOINT")
```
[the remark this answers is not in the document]
thampiotr: +1 we should do that so the contract of what vars are expected is clear

integration-tests/configs/otel-gen-client/main.go
```diff
@@ -0,0 +1,201 @@
+// Copyright The OpenTelemetry Authors
```
thampiotr, line 1: This is copied from somewhere, right? For simplicity of updating these files in the future, can you add a comment with a link to where this is copied from, with a commit sha? Same for the other files.

integration-tests/configs/otel-gen-server/main.go
```diff
@@ -0,0 +1,169 @@
...
+			sleep = rng.Int63n(2000)
+		case 1:
+			sleep = rng.Int63n(15)
+		case 2:
+			sleep = rng.Int63n(917)
+		case 3:
+			sleep = rng.Int63n(87)
+		case 4:
+			sleep = rng.Int63n(1173)
+		}
```
thampiotr, line 139: Hmm, any idea why they do that? Is that to have a specific distribution?

integration-tests/main.go
```diff
@@ -0,0 +1,56 @@
...
+	} else {
+		testDirs, err := filepath.Glob("./tests/*")
+		if err != nil {
+			panic(err)
+		}
+		logChan = make(chan TestLog, len(testDirs))
+		runAllTests()
+	}
+
+	cleanUpEnvironment()
```
thampiotr, line 53: As a general practice, I think it's better to run this in a defer statement so in case of panics it can be executed. Or when someone adds a new exit point from the function.

integration-tests/tests/otel-to-prom-metrics/config.river
```diff
@@ -0,0 +1,36 @@
...
+  endpoint {
+    url = "http://localhost:9009/api/v1/push"
+    metadata_config {
+        send_interval = "1s"
+    }
+    queue_config {
+        max_samples_per_send = 100
+    }
+  }
+}
```
thampiotr, line 36: missing new line - similar to other files

integration-tests/utils.go
```diff
@@ -0,0 +1,134 @@
...
+			defer wg.Done()
+			runSingleTest(td)
+		}(testDir)
+	}
+	wg.Wait()
+}
+
+func cleanUpEnvironment() {
+	fmt.Println("Cleaning up Docker environment...")
+	err := exec.Command(dockerComposeCmd, "down").Run()
```
thampiotr, line 112: Should we consider running `docker-compose down --volumes --rmi all` in order to not clog up the CI agents? I know that occasionally on my macOS, the Docker Desktop runs out of space due to number of images.

integration-tests/docker-compose.yaml
thampiotr: I think testing a simple end-to-end flow of each telemetry data type (logs, metrics, traces, profiles) to each major ecosystem (prom, loki, otel) would be within the scope of these tests and is a good idea to add. 

Anything more complicated and detailed we should probably test in unit-tests or something like the pipeline tests I've proposed here: https://github.com/grafana/agent/pull/5474

thampiotr (commented):
Few comments. I would also like to see a README.md where we could describe what tests are appropriate for this framework and what are the alternatives. 

In general, I would be inclined to merge this and see how reliable it is in practice. We may tweak or revert it later, but either way it's a worthwhile experiment and good work driving and implementing this @wildum ! 🎉 👏

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,25 @@
+name: Integration Tests
```
thampiotr, line 1: Forgot to ask - why GitHub Actions over using Drone? 
We may have more powerful machines in Drone. Also would be good to have decent statistics about runtime and success rates. Not sure which option does it better.

integration-tests/docker-compose.yaml
wildum: I dont think that avalanche can generate otel metrics. I used the demo-client and demo-server from the demo of the opentelemetry collector. @ptodev do you know about a good otel metrics generator to replace these demo app + the otel collector?

Makefile
wildum: make test is used in drone pipelines where the integration tests cannot run. I will create a make integration-test for now

integration-tests/configs/otel-gen-server/main.go
wildum: they use it to simulate different latencies, to create more interesting traces. I can remove this part to avoid long sleep

integration-tests/configs/otel-gen-client/Dockerfile
wildum: yeah because if I build on a mac and then try to run it on Alpine there are problems with C library (apparently musl vs glibc). I tried removing this flag and the binary would not run in docker

integration-tests/docker-compose.yaml
wildum: Actually I think that the demo-client / demo-server / opentelemetry collector is not so bad because it gives us a lot of flexibility for our testing. I think that it's valuable in our case because opentelemetry is moving fast and we always try to keep up with the latest updates

ptodev: > do you know about a good otel metrics generator to replace these demo app + the otel collector?

Nope. You could try asking on the internal opentelemetry channel. Alternatively, we could use Avalanche and have the Agent convert it to OTLP.

ptodev: > I used the demo-client and demo-server from the demo of the opentelemetry collector.

If this code is copied from somewhere, can we import it instead? Id doesn't have to be done for this issue, but it'd be nice if you could raise another issue for it? I guess it'd require an upstream contribution too?

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,22 @@
...
+on:
+  # Run tests on main just so the module and build cache can be saved and used
+  # in PRs. This speeds up the time it takes to test PRs dramatically.
```
ptodev, line 4: So the build cache of main is used when building PRs, but the build cache of one PR is not used when building other PRs? I thought that the cache would be something stored on a build server, and it'd be used regardless of which branch is being built.

integration-tests/common/metric.go
```diff
@@ -0,0 +1,48 @@
...
+	Status string     `json:"status"`
+	Data   MetricData `json:"data"`
+}
+
+type MetricData struct {
+	ResultType string         `json:"resultType"`
+	Result     []MetricResult `json:"result"`
+}
+
+type MetricResult struct {
```
ptodev, line 18: Do we need to check the metric type (counter, sum, etc.)? You don't have to do it now, but maybe as a todo

integration-tests/docker-compose.yaml
```diff
@@ -0,0 +1,62 @@
...
+    ports:
+      - "9009:9009"
+
+  loki:
+    image: grafana/loki:2.8.3
+    command: -config.file=/etc/loki/local-config.yaml
+    ports:
+      - "3100:3100"
+
+  otel-collector:
```
ptodev, line 20: I'm not sure if we should be running a Collector. I think we should just test the Agent. But I'm ok to leave this here for now.

ptodev (approved):
Thanks, looks good! I'm happy for it to be merged once @thampiotr also approves. 
Btw I wonder if we can also test kubernetes components one day too, using something like k3d. Testing kubernetes-related things locally has been a bit of a pain for me personally.

.github/workflows/integration-tests.yml
wildum: "I thought that the cache would be something stored on a build server, and it'd be used regardless of which branch is being built." -> I believe that this is correct. I think that the point is to always have a fresh cache available from the main branch that PRs can use when building for the first time

integration-tests/docker-compose.yaml
wildum: Yeah I wanted to use it for exemplars but I went for a more straightforward test. I would be happy to reconsider it in another issue

integration-tests/docker-compose.yaml
wildum: I created a tracking issue for the next steps.

.github/workflows/integration-tests.yml
ptodev: When is this workflow ran on the `main` branch? Is it when a PR is merged to it? If the test already ran during the PR, I'm not sure if there is any point in having them also run after the PR is merged? I guess the cache would be the same.

wildum: I agree but @rfratto removed it and they put it back for the unit tests in https://github.com/grafana/agent/pull/2921. Robert in the PR you say that "it turns out commits to main are responsible for storing the module and build cache", but aren't PRs also building the same cache?

ptodev: Hmm ok. I am not sure why this is, but if you could clarify in the comment in `.github/workflows/integration-tests.yml` that the cache will only be populated if we run this on `main` and not just the CI of any other branch? It'd be nice if @rfratto could help clarify further why this is, if he happens to remember the reason.

wildum: I think we cant try without it and see how it behaves. If we see that the PRs don't hit the cache we can investigate further

ptodev: > If we see that the PRs don't hit the cache

Do you know how to check for this? IS there a log line?

wildum: If the tests run under 4 mins then the cache was hit, if it runs in around 14mins that means that it spent 10mins to download the dependencies instead of using the cache

PR #5485, opened 2023-10-16
Use types from discovery Hetzner when converting configs

marctc:
#### PR Description

The type `hetzner.Role` and constants `HetznerRoleRobot` and `HetznerRoleHcloud` where not exported.
After upgrading Prometheus to last version, those are now exported, therefore we can cleanup the current
code to use the types from Prometheus.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


spartan0x117 (approved):
LGTM!

PR #5488, opened 2023-10-16
Make namespace and server optional in discovery nomad

### LINKED ISSUE #5479: Docs feedback: /docs/sources/flow/reference/components/discovery.nomad.md
Looks like `url` and `namespace` parameters are now required in `discover.nomad` while docs still say that they aren't.

I get `Failed to build component: decoding River: missing required attribute "namespace"` error when having just the url set.

marctc:
#### PR Description

Fields `namespace` and `server` from `discovery.nomad` have a default value but
the fields where marked as required. This PR fixes the issue.

Fixes #5479

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


spartan0x117 (approved):
I think the `docs/sources/flow/reference/components/discovery.nomad.md`'s `Usage` section may also need to be updated to remove the non-required `server` argument?

But LGTM!

PR #5489, opened 2023-10-16
Move GlobalRefMap into service instead of global var.

mattdurham:
#### PR Description

This is mostly a shift from global var to service with ideally no change in behavior. 

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


mattdurham:
ugh fixing some things when i blindly hit update branch from github ui

component/otelcol/exporter/prometheus/prometheus.go
```diff
@@ -75,7 +76,12 @@ var _ component.Component = (*Component)(nil)
 
 // New creates a new otelcol.exporter.prometheus component.
 func New(o component.Options, c Arguments) (*Component, error) {
-	fanout := prometheus.NewFanout(nil, o.ID, o.Registerer)
+	service, err := o.GetServiceData(labelstore.ServiceName)
+	if err != nil {
+		return nil, err
+	}
+	ls := service.(labelstore.LabelStore)
```
ptodev, line 83: Is this label store still valid even if the config is reloaded? Should we use a new label store instead, and update it in the component's `Update` function?

service/labelstore/service.go
```diff
@@ -0,0 +1,226 @@
...
+		m = &remoteWriteMapping{
+			RemoteWriteID: componentID,
+			localToGlobal: make(map[uint64]uint64),
+			globalToLocal: make(map[uint64]uint64),
+		}
+		s.mappings[componentID] = m
+	}
+
+	labelHash := lbls.Hash()
+	globalID, found := s.labelsHashToGlobal[labelHash]
```
ptodev, line 105: Does this handle hash collisions?

ptodev (commented):
I'm surprised that something like this is global. Isn't there going to be interference between components in certain edge cases? For example, if we have two different scrape -> remote_write pipelines, where there we scrape different endpoints but the metric names and labels are the same? In that case, for example staleness marker won't be applied properly if one series goes stale but the other isn't, and both are series with the same hash but actually come from different endpoints?

component/otelcol/exporter/prometheus/prometheus.go
mattdurham: Services require that items returned from service.data are consistent throughout the lifetime of the process.

service/labelstore/service.go
mattdurham: It does not but its what we have been running for a year or so.

component/otelcol/exporter/prometheus/prometheus.go
ptodev: I wonder if there are situations when this would fail. For example, this scenario might be an issue:
1. An Agent does scrape -> remote_write.
2. The config is reloaded to scrape a different endpoint, but this other endpoint has the same labels and metrics (so the series look the same).
3. At that point, the map would get confused and think that it's still tracking the old series.

I'm not sure if this has any adverse effects, but seems like a weird side-effect. There might be similar side-effects if we reload the config and point it to a different remote_write endpoint too.

mattdurham:
There should be no change to any underlying behavior. Beforehand the globalrefmap was a global var but instead moving it into a service. Its possible in the above scenario with two scrape loops for there to be overlap, though its not an issue today because we haven't wired in staleness markers. Its possible it would be an issue when we do wire them in.

component/otelcol/exporter/prometheus/prometheus.go
mattdurham: That would be fine since in the current iteration they are the same to it and should be. At the heart its a int to string mapping and a int to local remote write int mapping.

service/labelstore/service.go
```diff
@@ -0,0 +1,226 @@
...
+
+// RemoveStaleMarker removes a stale marker
+func (s *service) RemoveStaleMarker(globalRefID uint64) {
+	s.mut.Lock()
+	defer s.mut.Unlock()
+
+	delete(s.staleGlobals, globalRefID)
+}
+
+// staleDuration determines how often we should wait after a stale value is received to GC that value
```
thampiotr, line 185:
```suggestion
// staleDuration determines how long we should wait after a stale value is received to GC that value
```

service/labelstore/service.go
```diff
@@ -0,0 +1,226 @@
...
+var staleDuration = time.Minute * 10
```
thampiotr, line 186: Should we take this opportunity and make this configurable?

service/labelstore/service.go
```diff
@@ -0,0 +1,226 @@
...
+
+// CheckStaleMarkers is called to garbage collect and items that have grown stale over stale duration (10m)
+func (s *service) CheckStaleMarkers() {
```
thampiotr, line 189: `CheckStaleMarkers` doesn't sound like a good name IMO - doesn't reflect that this will delete old markers and garbage-collect effectively. We can call it just `GarbageCollection()` as it may serve multiple purposes.

thampiotr, line 189: This seems to only be called from a test. Shouldn't we have something calling this?

component/otelcol/exporter/prometheus/prometheus.go
thampiotr: > That would be fine since in the current iteration they are the same to it and should be.

I don't follow. Why using a stale LocalRefID, from the previous configuration of a remote_write is not a problem?

component/prometheus/remotewrite/remote_write.go
```diff
@@ -103,10 +112,10 @@ func New(o component.Options, c Arguments) (*Component, error) {
 				return 0, fmt.Errorf("%s has exited", o.ID)
 			}
 
-			localID := prometheus.GlobalRefMapping.GetLocalRefID(res.opts.ID, uint64(globalRef))
+			localID := ls.GetLocalRefID(res.opts.ID, uint64(globalRef))
```
thampiotr, line 115: This `localID` may be for a previous configuration of the component and may no longer be valid. Is that okay?

service/labelstore/service.go
mattdurham: We should but avoiding it for this PR since this is fork lifting functionality with ideally zero change.

component/otelcol/exporter/prometheus/prometheus.go
mattdurham: In the above scenario its the scraper that is changing not the remote_write. In the case of remote_write changing, I don't see anywhere we reset the walStorage in remote_write component on the updating so the local id cache should stay the same.

pkg/flow/componenttest/componenttest.go
```diff
@@ -158,6 +159,9 @@ func (c *Controller) buildComponent(dataPath string, args component.Arguments) (
 		DataPath:      dataPath,
 		OnStateChange: c.onStateChange,
 		Registerer:    prometheus.NewRegistry(),
+		GetServiceData: func(name string) (interface{}, error) {
+			return labelstore.New(nil), nil
+		},
```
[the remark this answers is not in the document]
mattdurham: sounds good

service/labelstore/service.go
mattdurham: Since it the code path is unused and trying to avoid changing behavior would like to push this to another pr.

component/prometheus/remotewrite/remote_write.go
mattdurham: There might be an issue with queue manager and seriesLabels but I think to fix that would require some Prometheus changes. Worst case a series would drop until the wal watcher reread a segment with them. This would also be a bug in our current incarnation.

service/labelstore/service.go
mattdurham: How do you feel about CheckAndRemoveStaleMarkers, I feel GarbageCollection is a bit to general at the moment.

component/prometheus/remotewrite/remote_write.go
thampiotr: True, that would already be a bug :) Maybe we could prevent it by adding a component 'incarnation' ID to the current component ID, so they don't reuse labels when config changes. Let's park this discussion for the future.

service/labelstore/data.go
```diff
@@ -0,0 +1,13 @@
...
+
+type LabelStore interface {
+	GetOrAddLink(componentID string, localRefID uint64, lbls labels.Labels) uint64
+	GetOrAddGlobalRefID(l labels.Labels) uint64
+	GetGlobalRefID(componentID string, localRefID uint64) uint64
+	GetLocalRefID(componentID string, globalRefID uint64) uint64
+	AddStaleMarker(globalRefID uint64, l labels.Labels)
+	RemoveStaleMarker(globalRefID uint64)
+	CheckStaleMarkers()
+}
```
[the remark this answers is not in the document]
mattdurham: Sounds good will do.

mattdurham:
FYI created https://github.com/grafana/agent/issues/5585 to handle wiring up garbage collection.

PR #5515, opened 2023-10-17
fix issue with non utf8 encodings and environment variable expansion

### LINKED ISSUE #4519: Issue with windows encoding and expand vars
### What's wrong?

If using an encoding that is not windows 1252 on a file (lets say UTF-16E) when the agent tries to expand vars it seems to write a UTF8 string into UTF16. This gives an error.

The simplest solution would be to either convert to UTF8 beforehand or return an error if the file is not utf8 or windows 1252. 

### Steps to reproduce

TBD

### System information

_No response_

### Software version

_No response_

### Configuration

_No response_

### Logs

_No response_

mattdurham:
#### PR Description

This converts all configs to UTF8. It doesnt attempt to try every combo of encodings but the most common ones: ascii, utf8bom/nobom, utf16le bom/nobom, and utf32bom/nobom.

Closes #4519

#### PR Checklist

- [X] CHANGELOG.md updated
- [ ] Documentation added
- [X] Tests updated


cmd/internal/flowmode/cmd_run.go
```diff
@@ -383,6 +384,10 @@ func loadFlowSource(path string, converterSourceFormat string, converterBypassEr
 			}
 
 			bb, err := os.ReadFile(curPath)
+			// Retain the original error and don't try to convert if an error occurred.
+			if err != nil {
+				bb, err = config.EnsureUTF8(bb)
+			}
```
rfratto, line 390: Technically this is a spec change for River files: https://grafana.com/docs/agent/latest/flow/config-language/files/

> River files are required to be UTF-8 encoded, and are permitted to contain Unicode characters. River files can use both Unix-style line endings (LF) and Windows-style line endings (CRLF), but formatters may replace all line endings with Unix-style ones.

pkg/config/config.go
```diff
@@ -356,14 +359,51 @@ func performEnvVarExpansion(buf []byte, expandEnvVars bool) ([]byte, error) {
 // applied to the file and must be done manually if LoadBytes
 // is called directly.
 func LoadBytes(buf []byte, expandEnvVars bool, c *Config) error {
-	expandedBuf, err := performEnvVarExpansion(buf, expandEnvVars)
+	utf8Buf, err := EnsureUTF8(buf)
```
erikbaranowski, line 362: nit: maybe this should be in performEnvVarExpansion so the code at this level doesn't change

cmd/internal/flowmode/cmd_run.go
mattdurham: How would you feel about adding a check that if the config file is not utf8bom/nobom, we return an error stating that it needs to be utf8? Right now if you pass in an utf16 you get a a slew of invalid character errors.

rfratto: Sure, I think that would be OK

pkg/config/config.go
mattdurham: Sounds reasonable.

cmd/internal/flowmode/cmd_run.go
mattdurham: Error on startup is now similar to `Error: reading config path "/home/mdurham/utils/flow/rw.river": configuration is encoded with UTF16LittleEndian but must be utf8`

PR #5516, opened 2023-10-17
Update migrating-from-static.md

mikempx:
Running existing static to flow conversion with "grafana-agent" insinuated that I still needed the old agent. By default, the executable is named, "grafana-agent-flow" in the documentation and doesn't mention a need to rename the executable.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/flow/getting-started/migrating-from-static.md
+++ b/docs/sources/flow/getting-started/migrating-from-static.md
@@ -83,13 +83,13 @@ features available in Grafana Agent Flow mode.
    {{% admonition type="caution" %}}If you bypass the errors, the behavior of the converted configuration may not match the original [Static] mode configuration. Make sure you fully test the converted configuration before using it in a production environment.{{% /admonition %}}
 
     ```bash
-    AGENT_MODE=flow; grafana-agent convert --source-format=static --bypass-errors --output=OUTPUT_CONFIG_PATH INPUT_CONFIG_PATH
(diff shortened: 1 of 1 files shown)


clayton-cornell:
@erikbaranowski Can you take a look at this? Is this the right solution or is there something else going on here that we need to account for?

clayton-cornell:
@mikempx Can you sign the CLA whenyou get a chance :-)

erikbaranowski:
@clayton-cornell the converter will work with either binary. It requires the env var only for the grafana-agent binary since Flow mode is not the default. I'm comfortable with documenting it in the converter guides however makes sense for clarity and consistency.

clayton-cornell:
@erikbaranowski maybe we document it both ways? 

If you're using Grafana Agent static mode: `AGENT_MODE=flow grafana-agent convert --source-format=static --bypass-errors --output=OUTPUT_CONFIG_PATH INPUT_CONFIG_PATH`

If you're using Grafana Agent flow mode: `grafana-agent-flow convert --source-format=static --bypass-errors --output=OUTPUT_CONFIG_PATH INPUT_CONFIG_PATH`

erikbaranowski:
I'm good with documenting both since we have done so in places like here: https://grafana.com/docs/agent/latest/flow/reference/cli/run/.

Let's update all 3 migration guides if we are making this change for consistency.

clayton-cornell:
@mikempx I'm going to close this PR. Your request resulted in a larger change to the docs than was implemented here and it was easier/faster and cleaner to open a new PR to take care of the change.

The new PR is: https://github.com/grafana/agent/pull/5544

PR #5526, opened 2023-10-18
Agent Management: Small doc tweaks

jcreixell:
#### PR Description

- Remove references to v2 in the API (this does not have much meaning to users).
- Specify that `agent_id` is required when label management is enabled.


spartan0x117 (approved):
LGTM, what do you think about moving the `agent_id` section right underneath the `label_management_enabled` section to make it easier to find after reading the requirement?

PR #5527, opened 2023-10-18
docs: remove mention of using statefulset for clustering

rfratto:
Clustering has not required a StatefulSet since v0.36.


rfratto:
@tpaschalis Is there anything else that needs to be configured if someone is using clustering without a StatefulSet?

clayton-cornell (approved):
Docs are OK as-is.

tpaschalis (approved):
No, there's no other prerequisite.

Issue #5225, opened 2023-09-15
[RFC] Grafana Agent flow mode plugins


rfratto:
**The most recent copy of this proposal can be found on [Google docs](https://docs.google.com/document/d/1ntN862VDFvvj9am2KFR9pNsHtjIN6f3luTYv6ZoGRMk/edit). The below is the original version of this proposal for posterity.**

> **NOTE**: _This proposal is likely being written 8-12 months too soon_. However, plugins are likely to be the last foundational piece of the future of Grafana Agent (following flow mode and modules). As the last foundational piece, it is important to get alignment on future plans early so we do not take any actions that goes against the long-term goals. 
>
> If you are reading this and are excited for plugins, do not expect rapid movement or delivery any time soon. The best case scenario is that plugins are made generally available in a 2025 release. 

## Background

Currently, new capabilities to Grafana Agent Flow can only be added by contributing a new component to the official Git repository (https://github.com/grafana/agent). 

Having a centralized repository of components causes issues for an open source project to thrive: 

* Grafana Agent maintainers may disagree with the value or implementation of a new component, preventing others from using it.
* Accepting a new component implies that it is "official" and receives the same amount of support as all other components.
* All components must share the same Apache 2.0 licensing; preventing components which can not follow this license, such as a component to launch a Loki server. 
* The project ends up with a massive amount of dependencies, causing dependency hell issues.
* All components are bound to the same release cycle, even if newer components would benefit from more frequent releases. 

Other projects, such as the OpenTelemetry Collector, solve this problem by having different _distributions_ of the collector. While distribution solves the issues above, it also fragments the community, as different distributions may have different subsets of components, making migration between distributions difficult. 

I propose that we support a plugin system for Grafana Agent flow mode, allowing sets of components to be provided by external plugins which can be loaded at runtime into the Grafana Agent process. 

This proposal serves as a high-level proposal of plugins to achieve maintainer and community consensus on the long-term goals. 

## Goals
 
* Get consensus on the need for plugins. 
* Establish baseline requirements for plugins. 
* Establish proposals to follow this proposal. 
* Establish phases for how plugins will be introduced. 

## Non-goals 

* Establish any technical information about plugins. 

## Proposal 

Flow mode should introduce the concept of a "plugin," where a plugin is some loadable code that provides one or more components that can be defined in a Flow configuration. 

The mechanism through which plugins are created, retrieved and defined are not in scope for this proposal:

[...]

PR #5528, opened 2023-10-18
Improves Pyroscope's scrape sync

cyriltovena:
This improves the sync method of Pyroscope scrapes based on some investigation.

The change is basically caching values of labels, url and hash of targets to avoid regenerating them when we sync and scrape profiles.

To ensure we're improving something I've added a benchmark.

Result:

```
benchstat before.txt after.txt
name     old time/op    new time/op    delta
Sync-16     796µs ±13%     167µs ± 3%  -79.06%  (p=0.008 n=5+5)

name     old alloc/op   new alloc/op   delta
Sync-16     609kB ± 0%     162kB ± 0%  -73.35%  (p=0.008 n=5+5)

name     old allocs/op  new allocs/op  delta
Sync-16     8.26k ± 0%     1.55k ± 0%  -81.29%  (p=0.008 n=5+5)
```

I'm sure there's more we can do.

### DIFF
--- a/component/pyroscope/scrape/scrape.go
+++ b/component/pyroscope/scrape/scrape.go
@@ -97,12 +97,21 @@ type ProfilingConfig struct {
 	Custom            []CustomProfilingTarget `river:"profile.custom,block,optional"`
 
 	PprofPrefix string `river:"path_prefix,attr,optional"`
+
+	allTargets map[string]ProfilingTarget
 }
 
 // AllTargets returns the set of all standard and custom profiling targets,
 // regardless of whether they're enabled. The key in the map indicates the name
 // of the target.
-func (cfg ProfilingConfig) AllTargets() map[string]ProfilingTarget {
+func (cfg *ProfilingConfig) AllTargets() map[string]ProfilingTarget {
+	if cfg.allTargets == nil {
+		cfg.allTargets = cfg.buildAllTargets()
+	}
+	return cfg.allTargets
+}
+
+func (cfg ProfilingConfig) buildAllTargets() map[string]ProfilingTarget {
 	targets := map[string]ProfilingTarget{
 		pprofMemory:            cfg.Memory,
 		pprofBlock:             cfg.Block,
@@ -376,7 +385,7 @@ func (c *Component) DebugInfo() interface{} {
 			if st != nil {
 				res = append(res, scrape.TargetStatus{
 					JobName:            job,
-					URL:                st.URL().String(),
+					URL:                st.URL(),
 					Health:             string(st.Health()),
 					Labels:             st.discoveredLabels.Map(),
 					LastError:          lastError,
--- a/component/pyroscope/scrape/scrape.go
+++ b/component/pyroscope/scrape/scrape.go
@@ -97,21 +97,12 @@ type ProfilingConfig struct {
 	Custom            []CustomProfilingTarget `river:"profile.custom,block,optional"`
 
 	PprofPrefix string `river:"path_prefix,attr,optional"`
-
-	allTargets map[string]ProfilingTarget
 }
 
 // AllTargets returns the set of all standard and custom profiling targets,
 // regardless of whether they're enabled. The key in the map indicates the name
 // of the target.
 func (cfg *ProfilingConfig) AllTargets() map[string]ProfilingTarget {
-	if cfg.allTargets == nil {
-		cfg.allTargets = cfg.buildAllTargets()
-	}
-	return cfg.allTargets
-}
-
-func (cfg ProfilingConfig) buildAllTargets() map[string]ProfilingTarget {
 	targets := map[string]ProfilingTarget{
 		pprofMemory:            cfg.Memory,
 		pprofBlock:             cfg.Block,
--- a/component/pyroscope/scrape/scrape_loop.go
+++ b/component/pyroscope/scrape/scrape_loop.go
@@ -56,12 +56,12 @@ func newScrapePool(cfg Arguments, appendable pyroscope.Appendable, logger log.Lo
 func (tg *scrapePool) sync(groups []*targetgroup.Group) {
 	tg.mtx.Lock()
 	defer tg.mtx.Unlock()
-
+	allTarget := tg.config.ProfilingConfig.AllTargets()
 	level.Info(tg.logger).Log("msg", "syncing target groups", "job", tg.config.JobName)
 	var actives []*Target
 	tg.droppedTargets = tg.droppedTargets[:0]
 	for _, group := range groups {
-		targets, dropped, err := targetsFromGroup(group, tg.config)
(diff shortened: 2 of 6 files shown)


component/pyroscope/scrape/scrape.go
```diff
@@ -97,12 +97,21 @@ type ProfilingConfig struct {
 	Custom            []CustomProfilingTarget `river:"profile.custom,block,optional"`
 
 	PprofPrefix string `river:"path_prefix,attr,optional"`
+
+	allTargets map[string]ProfilingTarget
```
rfratto, line 101: Can we cache the targets elsewhere? IMO it's an anti-pattern for a config field to contain state, and it'll eventually collide with our long-term plans to serialize these objects (see #5225).

cyriltovena: Moved the new benchmark is still pretty good:

```
benchstat before.txt after.txt
name     old time/op    new time/op    delta
Sync-16     796µs ±13%     182µs ± 4%  -77.13%  (p=0.008 n=5+5)

name     old alloc/op   new alloc/op   delta
Sync-16     609kB ± 0%     171kB ± 0%  -71.90%  (p=0.008 n=5+5)

name     old allocs/op  new allocs/op  delta
Sync-16     8.26k ± 0%     1.55k ± 0%  -81.19%  (p=0.008 n=5+5)
```

component/pyroscope/scrape/target.go
```diff
@@ -45,10 +45,13 @@ const (
 type Target struct {
 	// Labels before any processing.
 	discoveredLabels labels.Labels
-	// Any labels that are added to this target and its metrics.
-	labels labels.Labels
+	// Any privateLabels that are added to this target and its metrics.
+	privateLabels labels.Labels
+	labels        labels.Labels
```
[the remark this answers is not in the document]
thampiotr: It seems like we don't want the disjoint private/public labels, because I think it would break, for example:
```
t.appender.Append(context.Background(), t.privateLabels, []*pyroscope.RawSample{{RawProfile: b}});
```
I think you'll want to have `allLabels` and `publicLabels`, no need for private ones. I can take a look tomorrow.

component/pyroscope/scrape/target.go
```diff
@@ -107,13 +150,7 @@ func (t *Target) Params() url.Values {
 
 // Labels returns a copy of the set of all public labels of the target.
 func (t *Target) Labels() labels.Labels {
```
thampiotr, line 152: We should update the comment and check if no longer returning a copy is okay.

component/pyroscope/scrape/scrape_loop.go
```diff
@@ -238,7 +238,7 @@ func (t *scrapeLoop) scrape() {
 	if len(b) > 0 {
 		t.lastScrapeSize = len(b)
 	}
-	if err := t.appender.Append(context.Background(), t.labels, []*pyroscope.RawSample{{RawProfile: b}}); err != nil {
+	if err := t.appender.Append(context.Background(), t.privateLabels, []*pyroscope.RawSample{{RawProfile: b}}); err != nil {
```
thampiotr, line 241: Maybe I don't have a full context here, but shouldn't we call `Append` with either just public labels, or with all public + private labels (which seems to be the case previously?)

thampiotr: Traced the code and I see that this matches the previous behaviour now.

thampiotr (approved):

component/pyroscope/scrape/target.go
```diff
@@ -68,24 +69,24 @@ func NewTarget(lbls, discoveredLabels labels.Labels, params url.Values) *Target
 			publicLabels = append(publicLabels, l)
 		}
 	}
-	url := URLFromTarget(lbls, params)
+	url := urlFromTarget(lbls, params)
 
 	h := fnv.New64a()
-	_, _ = h.Write([]byte(fmt.Sprintf("%016d", publicLabels.Hash())))
+	_, _ = h.Write([]byte(strconv.FormatUint(publicLabels.Hash(), 16)))
```
thampiotr, line 75: fmt.Sprintf leads to an allocation, so this reduces it just a bit

component/pyroscope/scrape/target.go
```diff
@@ -58,25 +62,65 @@ type Target struct {
...
+		copy(newParams[k], v)
+	}
+	for _, l := range lbls {
+		if !strings.HasPrefix(l.Name, model.ParamLabelPrefix) {
+			continue
+		}
+		ks := l.Name[len(model.ParamLabelPrefix):]
+
+		if len(newParams[ks]) > 0 {
+			newParams[ks][0] = l.Value
```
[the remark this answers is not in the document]
thampiotr: It's the same behaviour as the previous implementation - see line 159. If this is still a bug, we might want to open this in a separate PR. I'll let @cyriltovena comment.

component/pyroscope/scrape/target.go
```diff
@@ -43,12 +43,16 @@ const (
 
 // Target refers to a singular HTTP or HTTPS endpoint.
 type Target struct {
+	// All labels of this target - public and private
+	allLabels labels.Labels
+	// Only public labels that are added to this target and its metrics.
+	publicLabels labels.Labels
```
tpaschalis, line 49: small nit: do we need to store a copy of the public labels?

AFAICT we only use it on the t.Labels() method. The method in turn is used a) In some logging contexts, b) when calling Clone() which will re-filter for the non-public labels, and c) in scrape_loop.go when checking whether to add targets in the `actives` slice, which should be already filtered inside of `populateLabels` in `targetsFromGroup`.

I think this may need a second look, but isn't a blocker in any case.

thampiotr: +1 - it also seems to me we can get rid of `.Clone()` method - which would allow to remove even more stuff and simplify. Is there some magic that requires the `Target` to have `Clone()`?

The `t.Labels()` is [also used in here, btw, to determine which targets are active](https://github.com/grafana/agent/blob/main/component/pyroscope/scrape/scrape_loop.go#L70).

And speaking of logging, [this debug line will](https://github.com/grafana/agent/blob/main/component/pyroscope/scrape/scrape_loop.go#L274) always build the `Labels().String()` which adds to allocation. 

Maybe we should do those improvements in a follow-up PR too?

tpaschalis: Yeah I saw the `c)` use-case, but I _think_ that this will always be true since `targetsFromGroup` will always filter out targets that only have internal labels, and they'd be bundled into `dropped` instead

As I said, not a blocker from me, happy to see this land and see the benefits.

thampiotr:
Merging this as it provides a lot of value by addressing the core problem. We can follow-up with more clean up work afterwards.

PR #5532, opened 2023-10-18
helm: fix issue where the PodLogs CRD was installed with Helm annotations 

### LINKED ISSUE #5531: PodLog CRD in grafana-agent helm subchart incompatible with PodLog CRD in another part of the project
### What's wrong?

A recent change to the `grafana-agent` Helm chart added a `crd` subchart that contains a PodLog CRD that is incompatible with the PodLog CRD located in `production/operator/crds`. The [change](https://github.com/grafana/agent/pull/5410) shows that a CRD with only `v1alpha2` version listed while [another copy of the CRD](https://github.com/grafana/agent/blob/f2b2456503338af3a5d00cbdf628042fe933189c/production/operator/crds/monitoring.grafana.com_podlogs.yaml) contains the `v1alpha1` version.

If a user sets `crds.create` to true for this Helm chart and then installs another Grafana product's Helm cart such as [mimir](https://github.com/grafana/mimir/blob/2c11a34ae4531a4c280f51ab9133c855dc523d0f/operations/helm/charts/mimir-distributed/templates/metamonitoring/pod-logs.yaml)  with a `PodLog` template that references api version `v1alpha1`, that install will fail.

### Steps to reproduce

- Install `grafana-agent` helm chart per these instructions https://artifacthub.io/packages/helm/grafana/grafana-agent#usage and override `crds.create` value to true
- Install `mimir-distributed` helm chart with meta monitoring enabled per these instructions https://grafana.com/docs/helm-charts/mimir-distributed/latest/get-started-helm-charts/#generate-some-metrics-for-testing
- Observe install failure for the `PodLog` resource because the CRD that was installed from the `grafana-agent` Helm chart is incompatible.

### System information

Linux 5.10.192-183.736.amzn2.x86_64

### Software version

Grafana Agent: 0.37.2 Helm chart version: grafana-agent-0.27.0

### Configuration

```text
grafana-agent values at default except for `crds.create` which should be set to true

mimir-distributed values for meta-monitoring described in https://grafana.com/docs/helm-charts/mimir-distributed/latest/get-started-helm-charts/#generate-some-metrics-for-testing
```

### Logs

_No response_

rfratto:
#5410 moved CRDs to a subchart to allow the values.yaml to determine whether CRDs were installed. This was primarily intended to be used by users where passing `--skip-crds` is not possible, due to limited infrastructure tooling. 

However, this caused a problem with upgrading, as the CRDs were now installed as a template, causing the normal Helm annotations to be added. 

This PR moves the CRDs to the `crds` folder of the subchart, allowing them to be managed as normal CRDs by Helm but still optionally installing them based on the values.yaml file.

Fixes #5531.


ptodev (approved):
I'm not an expert, but it makes sense to me on a high level. 

Do you think we should add extra notes in the testing section in `operations/helm/README.md` so that we can prevent such issues with future updates?

tpaschalis (approved):
I hit that issue on some point after trying to upgrade with v0.27.0, thanks for taking care of it! 🙌

hakuno:
Same here.
Thank you!

PR #5541, opened 2023-10-19
fix(prom/scrape): recreate internal scraper after reloaded

### LINKED ISSUE #5428: Grafana agent Flow module returning Healthy status when child component is not.
### Request

Grafana agent  to report consistent and real status for modules in the UI

Grafana agent Flow module is returning Healthy status when dependant part of it is not.
It would be nice to report the module as NOT HEALTHY when some of the child components are unhealthy

Grafana Agent version FLOW Mode
Version 0.36.2

### Use case

For consitency and observability of the real situation

### LINKED ISSUE #5536: Enabling enable_protobuf_negotiation in prometheus.scrape component requires restart
### What's wrong?

Set the configuration option `enable_protobuf_negotiation` in agent flow config.
Config was reloaded , but the option did not become effective.

E.g. DT, PR 98147.

### Steps to reproduce

1. Start Grafana Agent in Flow more without the option and start scraping an application that has native histograms.
2. Set the option to true and reload configuration.

### System information

_No response_

### Software version

_No response_

### Configuration

_No response_

### Logs

_No response_

hainenber:
#### PR Description
Recreate internal scrapeManager of  `prometheus.scrape` Flow component and hence multiple options will be applied once getting reload signal

Fixes #5536 

#### Notes to the Reviewer
Let me know if a unit test is required or current setup is enough to ensure no regression since this seems pretty convoluted to write a straightforward test.

#### PR Checklist

- [x] CHANGELOG.md updated
- [] Documentation added
- [ ] Tests updated

### DIFF
--- a/component/prometheus/scrape/scrape.go
+++ b/component/prometheus/scrape/scrape.go
@@ -140,27 +140,11 @@ var (
 
 // New creates a new prometheus.scrape component.
 func New(o component.Options, args Arguments) (*Component, error) {
-	data, err := o.GetServiceData(http.ServiceName)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get information about HTTP server: %w", err)
-	}
-	httpData := data.(http.Data)
-
-	data, err = o.GetServiceData(cluster.ServiceName)
+	data, err := o.GetServiceData(cluster.ServiceName)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get information about cluster: %w", err)
 	}
 	clusterData := data.(cluster.Cluster)
-
-	flowAppendable := prometheus.NewFanout(args.ForwardTo, o.ID, o.Registerer)
-	scrapeOptions := &scrape.Options{
-		ExtraMetrics: args.ExtraMetrics,
-		HTTPClientOptions: []config_util.HTTPClientOption{
-			config_util.WithDialContextFunc(httpData.DialFunc),
-		},
-		EnableProtobufNegotiation: args.EnableProtobufNegotiation,
-	}
-	scraper := scrape.NewManager(scrapeOptions, o.Logger, flowAppendable)
 
 	targetsGauge := client_prometheus.NewGauge(client_prometheus.GaugeOpts{
 		Name: "agent_prometheus_scrape_targets_gauge",
@@ -174,10 +158,18 @@ func New(o component.Options, args Arguments) (*Component, error) {
 		opts:          o,
 		cluster:       clusterData,
 		reloadTargets: make(chan struct{}, 1),
-		scraper:       scraper,
-		appendable:    flowAppendable,
 		targetsGauge:  targetsGauge,
 	}
+
+	// Update created component with prometheus.Fanout and prometheus.ScrapeManager
+	data, err = o.GetServiceData(http.ServiceName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get information about HTTP server: %w", err)
+	}
+	httpData := data.(http.Data)
+	flowAppendable, scraper := c.createPromScrapeResources(httpData, args)
+	c.appendable = flowAppendable
+	c.scraper = scraper
 
 	// Call to Update() to set the receivers and targets once at the start.
 	if err := c.Update(args); err != nil {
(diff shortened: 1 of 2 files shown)


component/prometheus/scrape/scrape.go
```diff
@@ -174,11 +158,19 @@ func New(o component.Options, args Arguments) (*Component, error) {
...
 
+	// Update created component with prometheus.Fanout and prometheus.ScrapeManager
+	data, err = o.GetServiceData(http.ServiceName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get information about HTTP server: %w", err)
+	}
+	httpData := data.(http.Data)
+	flowAppendable, scraper := c.createPromScrapeResources(httpData, args)
+	c.appendable = flowAppendable
+	c.scraper = scraper
```
rfratto, line 172: Is this necessary, if the same code will run in Update on line 175?

component/prometheus/scrape/scrape.go
```diff
@@ -236,13 +228,22 @@ func (c *Component) Update(args component.Arguments) error {
...
 	c.args = newArgs
 
+	// Update scraper with scrape options
+	data, err := c.opts.GetServiceData(http.ServiceName)
+	if err != nil {
+		return fmt.Errorf("failed to get information about HTTP server: %w", err)
+	}
+	newFlowAppendables, newScraper := c.createPromScrapeResources(data.(http.Data), newArgs)
+	c.appendable = newFlowAppendables
+	c.scraper = newScraper
```
rfratto, line 238: There's an issue here: the Run method runs the original `c.scraper`, and will never terminate that scraper/run the new one that's created. This means the changes here on subsequent updates will never truly apply.

component/prometheus/scrape/scrape.go
```diff
@@ -236,13 +228,22 @@ func (c *Component) Update(args component.Arguments) error {
 	defer c.mut.Unlock()
 	c.args = newArgs
 
+	// Update scraper with scrape options
+	data, err := c.opts.GetServiceData(http.ServiceName)
+	if err != nil {
+		return fmt.Errorf("failed to get information about HTTP server: %w", err)
+	}
+	newFlowAppendables, newScraper := c.createPromScrapeResources(data.(http.Data), newArgs)
+	c.appendable = newFlowAppendables
```
rfratto, line 237: I don't think we have to recreate the appendables; the call to UpdateChildren on line 240 should be sufficient, and we can return to having only one instance of `c.appendable` for the lifetime of the component.

rfratto (commented):
Thanks! This is heading in the right direction but needs some more work before it's mergeable. 

Can you write a unit test for this? It would be helpful for us to validate the fix with automation instead of reviewing by eye.

hainenber:
Alrighty, I'll try writing one 😄

component/prometheus/scrape/scrape.go
hainenber: I was trying to retain the previous setup. Now you're saying, it does look extra. Let see if the upcoming unit test can invalidate this step and I'll remove it

component/prometheus/scrape/scrape.go
hainenber: Sounds great, -1 work for me then 😄

rfratto:
@hainenber You might find the [componenttest](https://github.com/grafana/agent/tree/main/pkg/flow/componenttest) package useful here to test components (Look for other unit tests which use it to get examples).

This will handle running/updating the component for you so you can focus your unit test on creating a fake scrape target and determining whether protobuf negotiation is actually enabled. You'll know protobuf negotation is being used by looking at the `Accept` header sent to scrape targets; if it includes `vnd.google.protobuf`, then protobuf negotiation is enabled.

component/prometheus/scrape/scrape.go
hainenber: I totally missed this one out 🫣 

But I do think we can rewrite the scraper's Run goroutine into a for-select loop to handle scraper restart. 

Let me try out my luck on this.

hainenber:
Hi there, I've found making the goroutine in Run() function reloadable pretty tough and think it's much better to fix this on upstream, namely `prometheus/prometheus`.

I've made a [PR](https://github.com/prometheus/prometheus/pull/13017) to address it. Once the patch got merged, hopefully someone can cherry-pick it into a new `grafana/prometheus` 😄 

For now, this PR will be closed in anticipation.

hainenber:
Turns out [the `enable_protobuf_negotiation` option has been deprecated in latest `prometheus`](https://github.com/hainenber/prometheus/commit/624b973ebf4d0d7ed863b5538d7940c9c66c5cd0#diff-344d08b4acfbe19edf8173007073c1f12f126c10830855eca8d47936e53f9371).

The new option, `scrape_protocol`, can be specified in `scrape_configs` and normal agent reload will help fixing original issue.

My PR on `prometheus/prometheus` is still opened to allow reloading the manager with less diffused options, such as `extra_metrics`

PR #5542, opened 2023-10-19
docs: Add label to prometheus.exporter.unix

jkroepke:
Ref: #5361

#### PR Description

Missing label for prometheus.exporter.unix

See #5361

#### PR Checklist

- [ ] CHANGELOG.md updated
- [x] Documentation added
- [ ] Tests updated
- [ ] Config converters updated

PR #5543, opened 2023-10-19
docs: fix old instances of prometheus.exporter.unix not needing a label

rfratto:
PR #5361 removed the concept of "singleton components" (components which must not have a user-specified label), and thus made prometheus.exporter.unix require a label.

However, documentation referring to the old form of this component was not updated to include a label.

Related to #5542.

### DIFF
--- a/docs/sources/flow/concepts/configuration_language.md
+++ b/docs/sources/flow/concepts/configuration_language.md
@@ -73,10 +73,6 @@ The most common expression is to reference the exports of a component like
 formed by merging the component's name (e.g., `local.file`), label (e.g.,
 `password_file`), and export name (e.g., `content`), delimited by period.
 
-For components that don't use labels, like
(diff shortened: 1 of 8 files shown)


docs/sources/flow/concepts/configuration_language.md
```diff
@@ -73,10 +73,6 @@ The most common expression is to reference the exports of a component like
 formed by merging the component's name (e.g., `local.file`), label (e.g.,
 `password_file`), and export name (e.g., `content`), delimited by period.
 
-For components that don't use labels, like
-`prometheus.exporter.unix`, only combine the component name with
-export name: `prometheus.exporter.unix.targets`.
```
rfratto, line 78: There is no longer any component which doesn't use a label; all of them require specifying a label.

tpaschalis (approved):

PR #5556, opened 2023-10-20
docs: remove implication that flow mode doesn't have feature parity from about page

rfratto:
Flow mode has feature parity with static mode and operator in v0.37. This removes references implying it doesn't.


clayton-cornell (approved):
I forgot to fix that one!

Issue #5567, opened 2023-10-23
Add option add_metrics_suffixes to otelcol.exporter.prometheus component


mar4uk:
### Request

I want to be able to configure whether Prometheus metric names should include type and unit suffixes

### Use case

In some cases in order `otelcol.exporter.prometheus` could be used as an alternative to [prometheus exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/prometheusexporter) component of the opentelemetry collector option `add_metrics_suffixes` should be supported

PR #5568, opened 2023-10-23
otelcol.exporter.prometheus add_metrics_suffixes option

mar4uk:
#### PR Description
Added option add_metrics_suffixes. If set to false, type and unit suffixes will not be added to metrics. Default: true.
This option is needed for compatibility with https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/prometheusexporter

https://github.com/grafana/agent/issues/5567

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,6 +88,9 @@ Main (unreleased)
   - `otelcol.receiver.kafka` has a new `version` argument to change the version of 
     the SASL Protocol for SASL authentication.
 
+- Added new config options to `otelcol.exporter.prometheus` in flow mode (@mar4uk):
+  - `add_metrics_suffixes`: configures whether to add type and unit suffixes to metrics names.
+
 v0.37.2 (2023-10-16)
 -----------------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -89,7 +89,7 @@ Main (unreleased)
     the SASL Protocol for SASL authentication.
 
 - Added new config options to `otelcol.exporter.prometheus` in flow mode (@mar4uk):
-  - `add_metrics_suffixes`: configures whether to add type and unit suffixes to metrics names.
+  - `add_metric_suffixes`: configures whether to add type and unit suffixes to metrics names.
 
 v0.37.2 (2023-10-16)
 -----------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,9 +88,9 @@ Main (unreleased)
   - `otelcol.receiver.kafka` has a new `version` argument to change the version of 
     the SASL Protocol for SASL authentication.
 
-- Added new config options to `otelcol.exporter.prometheus` in flow mode (@mar4uk):
-  - `add_metric_suffixes`: configures whether to add type and unit suffixes to metrics names.
-
+- Added an `add_metric_suffixes` option to `otelcol.exporter.prometheus` in flow mode, 
+  which configures whether to add type and unit suffixes to metrics names. (@mar4uk)
+  
 v0.37.2 (2023-10-16)
 -----------------
 
--- a/component/otelcol/exporter/prometheus/internal/convert/convert.go
+++ b/component/otelcol/exporter/prometheus/internal/convert/convert.go
@@ -288,7 +288,6 @@ func (conv *Converter) consumeMetric(app storage.Appender, memResource *memorySe
 }
 
 func (conv *Converter) consumeGauge(app storage.Appender, memResource *memorySeries, memScope *memorySeries, m pmetric.Metric) {
-	// TODO: should we make the param addMetricSuffixes configurable? For now we set the default value (true) to keep the same behavior as before
 	metricName := prometheus.BuildCompliantName(m, "", conv.opts.AddMetricSuffixes)
 
 	metricMD := conv.createOrUpdateMetadata(metricName, metadata.Metadata{
--- a/component/otelcol/exporter/prometheus/internal/convert/convert.go
+++ b/component/otelcol/exporter/prometheus/internal/convert/convert.go
@@ -64,7 +64,7 @@ type Options struct {
 	// labels from the scope in the metrics.
 	IncludeScopeLabels bool
 	// AddMetricSuffixes controls whether suffixes are added to metric names. Defaults to true.
-	AddMetricSuffixes bool `mapstructure:"add_metric_suffixes"`
+	AddMetricSuffixes bool
 }
 
 var _ consumer.Metrics = (*Converter)(nil)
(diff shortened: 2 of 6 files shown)


component/otelcol/exporter/prometheus/prometheus.go
```diff
@@ -80,6 +82,7 @@ func New(o component.Options, c Arguments) (*Component, error) {
 	converter := convert.New(o.Logger, fanout, convert.Options{
 		IncludeTargetInfo: true,
 		IncludeScopeInfo:  false,
+		AddMetricSuffixes: true,
```
thampiotr, line 85: Shouldn't this be `c.AddMetricSuffixes` here?
This is something that we, as a team, should probably cover with better tests.

docs/sources/flow/reference/components/otelcol.exporter.prometheus.md
```diff
@@ -37,13 +37,14 @@ otelcol.exporter.prometheus "LABEL" {
...
-`include_target_info` | `boolean` | Whether to include `target_info` metrics. | `true` | no
-`include_scope_info` | `boolean` | Whether to include `otel_scope_info` metrics. | `false` | no
+Name | Type | Description                                               | Default | Required
+---- | ---- |-----------------------------------------------------------| ------- | --------
+`include_target_info` | `boolean` | Whether to include `target_info` metrics.                 | `true` | no
+`include_scope_info` | `boolean` | Whether to include `otel_scope_info` metrics.             | `false` | no
 `include_scope_labels` | `boolean` | Whether to include additional OTLP labels in all metrics. | `true` | no
-`gc_frequency` | `duration` | How often to clean up stale metrics from memory. | `"5m"` | no
-`forward_to` | `list(receiver)` | Where to forward converted Prometheus metrics. | | yes
+`add_metrics_suffixes` | `boolean` | Whether to add type and unit suffixes to metrics. | `true` | no
```
thampiotr, line 45:
```suggestion
`add_metrics_suffixes` | `boolean` | Whether to add type and unit suffixes to metrics names. | `true` | no
```

thampiotr (commented):
Thanks for working on this! The docs may need some input from @clayton-cornell

thampiotr (commented):
I think we'll also need an entry in CHANGELOG.md. You can see how to add it following the pattern or our `docs/developer/contributing.md`

component/otelcol/exporter/prometheus/internal/convert/convert.go
```diff
@@ -63,6 +63,8 @@ type Options struct {
 	// IncludeScopeLabels includes the otel_scope_name and otel_scope_version
 	// labels from the scope in the metrics.
 	IncludeScopeLabels bool
+	// AddMetricSuffixes controls whether suffixes are added to metric names. Defaults to true.
+	AddMetricSuffixes bool `mapstructure:"add_metric_suffixes"`
```
tpaschalis, line 67:
```suggestion
	// AddMetricSuffixes controls whether suffixes are added to metric names. Defaults to true.
	AddMetricSuffixes bool
```

component/otelcol/exporter/prometheus/prometheus.go
mar4uk: Good catch!
Here the default value is set. And it is later overridden by a value from `c.AddMetricSuffixes` there:
```
if err := res.Update(c); err != nil {
	return nil, err
}
```
After your comment I realized that I forgot to update the method `Update` so it takes the value from `cfg. AddMetricSuffixes`

I actually am not quite sure why we set default values twice: once in `SetToDefault` method, and the second time in the `New` function. I'll write test to check config is unmarshalled correctly

component/otelcol/exporter/prometheus/internal/convert/convert.go
mar4uk: Thanks! fixed

component/otelcol/exporter/prometheus/prometheus.go
```diff
@@ -35,6 +35,7 @@ type Arguments struct {
 	IncludeScopeLabels bool                 `river:"include_scope_labels,attr,optional"`
 	GCFrequency        time.Duration        `river:"gc_frequency,attr,optional"`
 	ForwardTo          []storage.Appendable `river:"forward_to,attr"`
+	AddMetricSuffixes  bool                 `river:"add_metrics_suffixes,attr,optional"`
```
rlankfo, line 38: can we change this tag to be `add_metric_suffixes` instead of `add_metrics_suffixes` so we can match what the collector is doing?

rlankfo:
```suggestion
	AddMetricSuffixes  bool                 `river:"add_metric_suffixes,attr,optional"`
```

mar4uk: you're right, I made a typo, will fix it to align with collector 👍

component/otelcol/exporter/prometheus/prometheus.go
thampiotr: Right, the `Update` seems to be the correct place to use `c.AddMetricSuffixes`. Thanks for fixing.

In `New` we call `converter.New` and then immediately we'll call `converter.UpdateOptions` inside the `Update`, so what you pass in the `New` may not really matter. 

However, I think for readability of the code it's best we still make the values in both places to be the correct set of `convert.Options`. We can add a function that allows us to convert `Arguments` to `convert.Options` and use it in both places.

mar4uk: sounds good! Did this

CHANGELOG.md
```diff
@@ -69,6 +69,9 @@ Main (unreleased)
 
 - Improved performance of `pyroscope.scrape` component when working with a large number of targets. (@cyriltovena)
 
+- Added new config options to `otelcol.exporter.prometheus` in flow mode (@mar4uk):
+  - `add_metric_suffixes`: configures whether to add type and unit suffixes to metrics names.
```
thampiotr, line 73:
```suggestion
- Added an `add_metric_suffixes` option to `otelcol.exporter.prometheus` in flow mode, which configures whether to add type and unit suffixes to metrics names. (@mar4uk)
```

thampiotr (approved):
LGTM % one suggestion. Thanks!

clayton-cornell (approved):
Docs as they are look OK. I can't tell if we need more docs. I'm thinking no since this is one option on existing functionality.

tpaschalis (approved):
LGTM!

rlankfo (approved):

PR #5572, opened 2023-10-23
Replace relrefs with docrefs in Agent install and start topics

clayton-cornell:
#### PR Description

The relref shortcode is being deprecated in favour of the docs/reference shortcode. The new shortcode allows us to have more control over the references and links and makes sure that topics mounted into more than one project can have working links within that project.

This PR continues the process of migrating the Flow documentation to use the docs/reference shortcodes as preparation for mounting Flow topics into the Cloud docs

#### Notes to the Reviewer

I removed the Start Agent link from the Deploy Kubernetes topic - seemed to be redundant since the deployment doesn't need a start step (if I understand the K8s install correctly)

A couple of Notes were converted to block quote styled notes to allow doc/reference links within the Note. The doc/reference links do not work when embedded into other shortcodes. The linking works properly in a block quote and the published result is the same.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


docs/sources/flow/setup/install/kubernetes.md
```diff
@@ -59,5 +57,14 @@ For more information on the Grafana Agent Helm chart, refer to the Helm chart do
 
 ## Next steps
 
-- [Start Grafana Agent]({{< relref "../start-agent" >}})
-- [Configure Grafana Agent]({{< relref "../configure/configure-kubernetes" >}})
+<!--- [Start Grafana Agent][]-->
```
erikbaranowski, line 60: should we remove the commented out unused link here and below?

erikbaranowski (approved):
LGTM, 1 question to check out

docs/sources/flow/setup/install/kubernetes.md
clayton-cornell: I was going to, but I wanted to double check that Agent in K8s deployments doesn't need an explicit start - in my local test it doesn't.

clayton-cornell: Removed it :-)

PR #5580, opened 2023-10-24
pkg/metrics/wal: drop out-of-order exemplars

tpaschalis:
#### PR Description

During my second or third week at Grafana, I wrote #1316 to avoid sending duplicate exemplars and putting extra load on remote write systems Now, almost two years after, it's fun having to revisit this code and reminiscing all the steps we've made to be here today. :) Just wanted to say thanks to everyone who helped on this journey. <3

Anyway, Native Histogram can have more than one exemplar on the same datapoint; if they don't have an explicit timestamp they inherit the scrape's timestamp, and this triggers the out-of-order warning on the remote write side.

Since OOO exemplars are not in the 'normal' code path, having remote writes reject them with a 4xx code makes the agent think that _all_ datapoints in that request were discarded (as there's no way for the protocol to relay partial success) leading to metrics such as `prometheus_remote_storage_samples_failed_total` report a false positive with a steady rate of failed samples and potentially masking _other_ issues with remote write connections.

Until there's consensus on how to handle this from the Prometheus side, we can put this workaround in place; instead of dropping only duplicate exemplars, we can also filter out the OOO ones.

Relates to: https://github.com/prometheus/prometheus/issues/12971

#### Notes to the Reviewer
Not adding a CHANGELOG entry as this shouldn't be a user-facing change.

#### PR Checklist

- [X] Tests updated

### DIFF
--- a/pkg/metrics/wal/wal_test.go
+++ b/pkg/metrics/wal/wal_test.go
@@ -172,7 +172,7 @@ func TestStorage_DuplicateExemplarsIgnored(t *testing.T) {
 	replayer := walReplayer{w: &collector}
 	require.NoError(t, replayer.Replay(s.wal.Dir()))
 
-	// We had 9 calls to AppendExemplar but only 4 of those should have gotten through
+	// We had 11 calls to AppendExemplar but only 4 of those should have gotten through.
 	require.Equal(t, 4, len(collector.exemplars))
 }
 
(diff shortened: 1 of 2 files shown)


pkg/metrics/wal/wal_test.go
```diff
@@ -163,12 +164,16 @@ func TestStorage_DuplicateExemplarsIgnored(t *testing.T) {
 	_, _ = app.AppendExemplar(sRef, nil, e)
 	_, _ = app.AppendExemplar(sRef, nil, e)
 
+	e.Ts = 24
+	_, _ = app.AppendExemplar(sRef, nil, e)
+	_, _ = app.AppendExemplar(sRef, nil, e)
+
```
tpaschalis, line 170: Adding an OOO exemplar doesn't change the number of exemplars stored in line 177.

pkg/metrics/wal/wal_test.go
```diff
@@ -452,14 +456,11 @@ func TestDBAllowOOOSamples(t *testing.T) {
...
 	require.NoError(t, app.Commit())
+
 	for _, metric := range payload {
-		for _, sa := range metric.samples {
-			// We want to set the timestamp to before. This should no longer trigger an out of order.
-			sa.ts = sa.ts - 10_000
-		}
-	}
-	for _, metric := range payload {
-		metric.Write(t, app)
```
tpaschalis, line 462: The Write helper was also trying to re-append exemplars, which now fails.

I've extracted this to a new helper method that includes the timestamp offset and does _not_ add exemplars.

mattdurham (approved):
LGTM

PR #5584, opened 2023-10-24
update oracledb-exporter to latest

schmikei:
#### PR Description

Mostly pulling it in to benefit from this change: 

https://github.com/iamseth/oracledb_exporter/pull/344

Where there is better logging when users get an unsupported value type error message.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


marctc (approved):
Thanks!

PR #5588, opened 2023-10-24
Fix issue with monitors

mattdurham:
#### PR Description

Fixes an issue with pod monitor that was introduced today. Will get something with unit tests out asap.


captncraig (approved):
I trust you.

rfratto:
@mattdurham The service needs to be listed as dependencies in module loaders too. Can you follow up with a PR to include it there?

PR #5590, opened 2023-10-24
Implement WAL replay and markers for `loki.write`

thepalbi:
#### PR Description

This PR branches of https://github.com/grafana/agent/pull/5434, and applies the work done in https://github.com/ptodev/prometheus/pull/1 to improve two things:
1. WAL replay is implemented for `loki.write`, so upon an agent restart, it will replay "not sent" data saved in the WAL segments that are alive
2. Implement a "markers" mechanism that keeps track of which was the last consumed segment, consumed being read and delivered to the RW destination. This mechanisms works by creating a feedback loop between the side that reads from the WAL, and the side that sends batches.

#### TODOs
- [ ] Add tests for corrupted marker files https://github.com/grafana/agent/pull/5590#discussion_r1372979663

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/component/common/loki/client/batch.go
+++ b/component/common/loki/client/batch.go
@@ -30,6 +30,9 @@ type batch struct {
 	createdAt  time.Time
 
 	maxStreams int
+
+	// segmentCounter tracks the amount of entries for each segment present in this batch.
+	segmentCounter map[int]int
 }
 
 func newBatch(maxStreams int, entries ...loki.Entry) *batch {
@@ -73,13 +76,14 @@ func (b *batch) add(entry loki.Entry) error {
 }
 
 // add an entry to the batch
-func (b *batch) addFromWAL(lbs model.LabelSet, entry logproto.Entry) error {
+func (b *batch) addFromWAL(lbs model.LabelSet, entry logproto.Entry, segmentNum int) error {
 	b.totalBytes += len(entry.Line)
 
 	// Append the entry to an already existing stream (if any)
 	labels := labelsMapToString(lbs, ReservedLabelTenantID)
 	if stream, ok := b.streams[labels]; ok {
 		stream.Entries = append(stream.Entries, entry)
+		b.countForSegment(segmentNum)
 		return nil
 	}
 
@@ -93,6 +97,7 @@ func (b *batch) addFromWAL(lbs model.LabelSet, entry logproto.Entry) error {
 		Labels:  labels,
 		Entries: []logproto.Entry{entry},
 	}
+	b.countForSegment(segmentNum)
 
 	return nil
 }
@@ -170,4 +175,12 @@ func (b *batch) createPushRequest() (*logproto.PushRequest, int) {
 		entriesCount += len(stream.Entries)
 	}
 	return &req, entriesCount
+}
+
+func (b *batch) countForSegment(segment int) {
+	if curr, ok := b.segmentCounter[segment]; ok {
+		b.segmentCounter[segment] = curr + 1
+		return
+	}
+	b.segmentCounter[segment] = 1
 }
--- a/component/common/loki/client/batch.go
+++ b/component/common/loki/client/batch.go
@@ -19,6 +19,12 @@ const (
 	errMaxStreamsLimitExceeded = "streams limit exceeded, streams: %d exceeds limit: %d, stream: '%s'"
 )
 
+// SentDataMarkerHandler is a slice of the MarkerHandler interface, that the batch interacts with to report the event that
+// all data in the batch has been delivered or a client failed to do so.
+type SentDataMarkerHandler interface {
+	UpdateSentData(segmentId, dataCount int)
+}
+
 // batch holds pending log streams waiting to be sent to Loki, and it's used
 // to reduce the number of push requests to Loki aggregating multiple log streams
 // and entries in a single batch request. In case of multi-tenant Promtail, log
@@ -53,16 +59,6 @@ func newBatch(maxStreams int, entries ...loki.Entry) *batch {
 	return b
 }
 
-type SendDataMarkerHandler interface {
-	UpdateSentData(segmentId, dataCount int)
-}
-
-func (b *batch) reportAsSentData(h SendDataMarkerHandler) {
-	for seg, data := range b.segmentCounter {
-		h.UpdateSentData(seg, data)
-	}
-}
-
 // add an entry to the batch
 func (b *batch) add(entry loki.Entry) error {
 	b.totalBytes += len(entry.Line)
@@ -86,7 +82,8 @@ func (b *batch) add(entry loki.Entry) error {
 	return nil
 }
 
-// add an entry to the batch
+// addFromWAL adds an entry to the batch, tracking that the data being added comes from segment segmentNum read from the
+// WAL.
 func (b *batch) addFromWAL(lbs model.LabelSet, entry logproto.Entry, segmentNum int) error {
 	b.totalBytes += len(entry.Line)
 
@@ -188,10 +185,19 @@ func (b *batch) createPushRequest() (*logproto.PushRequest, int) {
 	return &req, entriesCount
 }
 
-func (b *batch) countForSegment(segment int) {
-	if curr, ok := b.segmentCounter[segment]; ok {
-		b.segmentCounter[segment] = curr + 1
+// countForSegment tracks that one data item has been read from a certain WAL segment.
+func (b *batch) countForSegment(segmentNum int) {
+	if curr, ok := b.segmentCounter[segmentNum]; ok {
+		b.segmentCounter[segmentNum] = curr + 1
 		return
 	}
-	b.segmentCounter[segment] = 1
+	b.segmentCounter[segmentNum] = 1
+}
+
+// reportAsSentData will report for all segments whose data is part of this batch, the amount of that data as sent to
+// the provided SentDataMarkerHandler
+func (b *batch) reportAsSentData(h SentDataMarkerHandler) {
+	for seg, data := range b.segmentCounter {
+		h.UpdateSentData(seg, data)
+	}
 }
--- a/component/common/loki/client/batch.go
+++ b/component/common/loki/client/batch.go
@@ -37,10 +37,11 @@ type batch struct {
 
 func newBatch(maxStreams int, entries ...loki.Entry) *batch {
 	b := &batch{
-		streams:    map[string]*logproto.Stream{},
-		totalBytes: 0,
-		createdAt:  time.Now(),
-		maxStreams: maxStreams,
+		streams:        map[string]*logproto.Stream{},
+		totalBytes:     0,
+		createdAt:      time.Now(),
+		maxStreams:     maxStreams,
+		segmentCounter: map[int]int{},
 	}
 
 	// Add entries to the batch
--- a/component/common/loki/client/batch.go
+++ b/component/common/loki/client/batch.go
@@ -53,6 +53,16 @@ func newBatch(maxStreams int, entries ...loki.Entry) *batch {
 	return b
 }
 
+type SendDataMarkerHandler interface {
+	UpdateSentData(segmentId, dataCount int)
+}
+
+func (b *batch) reportAsSentData(h SendDataMarkerHandler) {
+	for seg, data := range b.segmentCounter {
+		h.UpdateSentData(seg, data)
+	}
+}
+
 // add an entry to the batch
 func (b *batch) add(entry loki.Entry) error {
 	b.totalBytes += len(entry.Line)
--- a/component/common/loki/client/internal/marker_encoding.go
+++ b/component/common/loki/client/internal/marker_encoding.go
@@ -30,7 +30,7 @@ func EncodeMarkerV1(segment uint64) ([]byte, error) {
 	return bs, nil
 }
 
-// DecodeMarkerV1
+// DecodeMarkerV1 decodes the segment number from a segment marker, encoded with EncodeMarkerV1.
 func DecodeMarkerV1(bs []byte) (uint64, error) {
 	// first check that read byte stream has expected length
 	if len(bs) != 14 {
--- a/component/common/loki/client/internal/marker_encoding.go
+++ b/component/common/loki/client/internal/marker_encoding.go
@@ -1,0 +1,55 @@
+package internal
+
+import (
+	"encoding/binary"
+	"fmt"
+	"hash/crc32"
+)
+
+var (
+	markerHeaderV1 = []byte{'0', '1'}
+)
+
+// EncodeMarkerV1 encodes the segment number, from whom we need to create a marker, in the marker file format,
+// which in v1 includes the segment number and a trailing CRC code of the first 10 bytes.
+func EncodeMarkerV1(segment uint64) ([]byte, error) {
+	// marker format v1
+	// marker [ 0 , 1 ] - HEADER, which is used to track version
+	// marker [ 2 , 9 ] - encoded unit 64 which is the content of the marker, the last "consumed" segment
+	// marker [ 10, 13 ] - CRC32 of the first 10 bytes of the marker, using IEEE polynomial
+	bs := make([]byte, 14)
+	// write header with marker format version
+	bs[0] = markerHeaderV1[0]
+	bs[1] = markerHeaderV1[1]
+	// write actual marked segment number
+	binary.BigEndian.PutUint64(bs[2:10], segment)
+	// checksum is the IEEE CRC32 checksum of the first 10 bytes of the marker record
+	checksum := crc32.ChecksumIEEE(bs[0:10])
+	binary.BigEndian.PutUint32(bs[10:], checksum)
+
+	return bs, nil
+}
+
+// DecodeMarkerV1
+func DecodeMarkerV1(bs []byte) (uint64, error) {
+	// first check that read byte stream has expected length
+	if len(bs) != 14 {
+		return 0, fmt.Errorf("bad length %d", len(bs))
+	}
+
+	// check CRC first
+	expectedCrc := crc32.ChecksumIEEE(bs[0:10])
+	gotCrc := binary.BigEndian.Uint32(bs[len(bs)-4:])
+	if expectedCrc != gotCrc {
+		return 0, fmt.Errorf("corrupted WAL marker")
+	}
+
+	// check expected version header
+	header := bs[:2]
+	if !(header[0] == markerHeaderV1[0] && header[1] == markerHeaderV1[1]) {
+		return 0, fmt.Errorf("wrong WAL marker header")
+	}
+
+	// lastly, decode marked segment number
+	return binary.BigEndian.Uint64(bs[2:10]), nil
+}
--- a/component/common/loki/client/internal/marker_file_handler.go
+++ b/component/common/loki/client/internal/marker_file_handler.go
@@ -46,12 +46,6 @@ func NewMarkerFileHandler(logger log.Logger, walDir string) (MarkerFileHandler, 
 	if err := os.MkdirAll(markerDir, MarkerFolderMode); err != nil {
 		return nil, fmt.Errorf("error creating segment marker folder %q: %w", markerDir, err)
 	}
-	// If marker folder exists, and has other perms, change them
-	if stat, err := os.Stat(markerDir); err == nil && stat.Mode().Perm() != MarkerFolderMode {
-		if err := os.Chmod(markerDir, MarkerFolderMode); err != nil {
-			return nil, err
-		}
-	}
 
 	mfh := &markerFileHandler{
 		logger:                    logger,
--- a/component/common/loki/client/internal/marker_file_handler.go
+++ b/component/common/loki/client/internal/marker_file_handler.go
@@ -1,15 +1,14 @@
 package internal
 
 import (
+	"bytes"
 	"fmt"
-	"github.com/grafana/agent/component/common/loki/wal"
-	"os"
-	"path/filepath"
-	"strconv"
-
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
-	"github.com/prometheus/prometheus/tsdb/fileutil"
+	"github.com/grafana/agent/component/common/loki/wal"
+	"github.com/natefinch/atomic"
+	"os"
+	"path/filepath"
 )
 
 const (
@@ -28,6 +27,7 @@ type MarkerFileHandler interface {
 
 type markerFileHandler struct {
 	logger                    log.Logger
+	lastMarkedSegmentDir      string
 	lastMarkedSegmentFilePath string
 }
 
@@ -45,6 +45,7 @@ func NewMarkerFileHandler(logger log.Logger, walDir string) (MarkerFileHandler, 
 
 	mfh := &markerFileHandler{
 		logger:                    logger,
+		lastMarkedSegmentDir:      filepath.Join(markerDir),
 		lastMarkedSegmentFilePath: filepath.Join(markerDir, MarkerFileName),
 	}
 
@@ -53,7 +54,7 @@ func NewMarkerFileHandler(logger log.Logger, walDir string) (MarkerFileHandler, 
 
 // LastMarkedSegment implements wlog.Marker.
 func (mfh *markerFileHandler) LastMarkedSegment() int {
-	bb, err := os.ReadFile(mfh.lastMarkedSegmentFilePath)
+	bs, err := os.ReadFile(mfh.lastMarkedSegmentFilePath)
 	if os.IsNotExist(err) {
 		level.Warn(mfh.logger).Log("msg", "marker segment file does not exist", "file", mfh.lastMarkedSegmentFilePath)
 		return -1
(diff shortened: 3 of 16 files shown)


component/common/loki/client/internal/marker_handler.go
```diff
@@ -0,0 +1,121 @@
...
+				continue
+			}
+
+			// we know (segment, 0) is in the map
+
+			// TODO: Is it safe to assume that just because a segment is 0 inside the map,
+			//      all samples from it have been processed?
+			if segment > markableSegment {
+				markableSegment = segment
+			}
```
ptodev, line 102: I'm not sure if this is correct. For example:
* Let's have two batches - `batch_1` and `batch_2`.
* `batch_1` is at segment 3, whereas `batch_2` is at segment 5.
* `markableSegment` should be set at 3, whereas atm I think we will set it at 5

component/common/loki/client/internal/marker_handler_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+func (m *mockMarkerFileHandler) LastMarkedSegment() int {
+	return m.lastMarkedSegment
+}
+
+func (m *mockMarkerFileHandler) MarkSegment(segment int) {
+	m.lastMarkedSegment = segment
+}
+
+func TestMarkerHandler(t *testing.T) {
```
ptodev, line 22: It'd be nice to add some fail-safety for corrupted marker files and also to have tests to check it.

component/common/loki/client/queue_client.go
```diff
@@ -0,0 +1,601 @@
...
+		logger: logger,
+	}
+
+	q.wg.Add(1)
+	go q.run()
+
+	return &q
+}
+
+// enqueue is a blocking operation to add to the send queue a batch ready to be sent.
```
ptodev, line 71: I think it "may be blocking"? The channel is buffered, so it may or may not be full.

component/common/loki/wal/watcher.go
```diff
@@ -58,7 +62,18 @@ type WriteTo interface {
...
 
-	AppendEntries(entries wal.RefEntries) error
+	AppendEntries(entries wal.RefEntries, segmentNum int) error
+}
+
+// Marker allows the Watcher to start from a specific segment in the WAL.
+// Implementers can use this interface to save and restore save points.
+type Marker interface {
+	// LastMarkedSegment should return the last segment stored in the marker.
+	// Must return nil if there is no mark.
```
ptodev, line 72: > Must return nil if there is no mark.

Is this comment valid? `LastMarkedSegment` doesn't return a pointer.

component/common/loki/wal/watcher.go
```diff
@@ -150,10 +184,27 @@ func (w *Watcher) run() error {
...
+	}
+
+	return -1, errors.New("failed to find segment for index")
+}
+
 // watch will start reading from the segment identified by segmentNum. If an EOF is reached, it will keep
 // reading for more WAL records with a wlog.LiveReader. Periodically, it will check if there's a new segment, and if positive
 // read the remaining from the current one and return.
-func (w *Watcher) watch(segmentNum int) error {
+func (w *Watcher) watch(segmentNum int, tail bool) error {
```
ptodev, line 207: It'd be good to mention in the comment above what `tail` is for.

component/common/loki/wal/watcher.go
```diff
@@ -212,7 +275,18 @@ func (w *Watcher) watch(segmentNum int) error {
...
+
+		// Ignore all errors reading to end of segment whilst replaying the WAL. If error, log a warning accordingly. If not,
+		// return with an ignorable error.
+		if !tail {
+			if err != nil && errors.Unwrap(err) != io.EOF {
+				level.Warn(w.logger).Log("msg", "Ignoring error reading to end of segment, may have dropped data", "segment", segmentNum, "err", err)
+			} else if reader.Offset() != size {
+				level.Warn(w.logger).Log("msg", "Expected to have read whole segment, may have dropped data", "segment", segmentNum, "read", reader.Offset(), "size", size)
+			}
+			return ErrIgnorable
```
ptodev, line 289: Aren't ignorable errors called "warnings"? 😄 Is there something special about these ones which makes them more important?

component/loki/write/types.go
```diff
@@ -70,6 +71,21 @@ func (r *EndpointOptions) Validate() error {
 	return nil
 }
 
+// QueueConfig controls how the queue logs remote write client is configured. Note that this client is only used when the
+// loki.write component has WAL support enabled.
+type QueueConfig struct {
+	Capacity     int           `river:"capacity,attr,optional"`
+	DrainTimeout time.Duration `river:"drain_timeout,attr,optional"`
+}
```
ptodev, line 79: If these parameters are something users can configure, then we should list them in the component docs.

component/common/loki/client/queue_client.go
```diff
@@ -0,0 +1,601 @@
...
+
+	// keep reading messages from sendQueue until all have been consumed, or timeout is exceeded
+	for {
+		select {
+		case qb := <-q.q:
+			// drain uses the same timeout, so if a timeout was applied to the parent context, it can cancel the underlying
+			// send operation preemptively.
+			q.client.sendBatch(ctx, qb.TenantID, qb.Batch)
+			// mark segment data for that batch as sent, even if the send operation failed
+			qb.Batch.reportAsSentData(q.client.markerHandler)
```
ptodev, line 123: If this is doing the same as the `run()` function, then I wonder if we should write a special function which just does `sendBatch` + `reportAsSentData`, and that function can then be used both by `run` and `closeAndDrain`. I'm worried about someone modifying the `run()` function and forgetting to modify this one.

ptodev (commented):
Thank you for this! I am super happy for this change. IMO it doesn't have to be super polished in order to be merged, but I just have a few main concerns:
* I'd like us to protect ourselves from corrupted marker files.
* Some of the new interfaces could use a few more comments :)
* I guess performance isn't impacted much, and we don't have to run any special benchmarks?

component/common/loki/client/batch.go
```diff
@@ -48,6 +52,16 @@ func newBatch(maxStreams int, entries ...loki.Entry) *batch {
 	return b
 }
 
+type SendDataMarkerHandler interface {
```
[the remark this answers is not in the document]
thepalbi: Yes for sure, will work through them. Didn't add much docs on top of those since this PR is branched of the other one. Now that https://github.com/grafana/agent/pull/5434 is approved I'll rebase and clean up this one (docs, tests, benches, etc).

component/common/loki/client/internal/marker_file_handler.go
```diff
@@ -0,0 +1,87 @@
...
+	mfh := &markerFileHandler{
+		logger:                    logger,
+		lastMarkedSegmentFilePath: filepath.Join(markerDir, "segment"),
+	}
+
+	return mfh, nil
+}
+
+// LastMarkedSegment implements wlog.Marker.
+func (mfh *markerFileHandler) LastMarkedSegment() int {
```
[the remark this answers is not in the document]
thepalbi: Yes, I agree when we introduce offsets into the segment (so we can keep more granular markers, and seek directly to that byte) we might want to change the file structure. To keep this PR as small as possible, I'm down for leaving this as is, and later we can introduce some file headers in case we need to change the marker files after offsets as well.

component/common/loki/client/internal/marker_file_handler_test.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestMarkerFileHandler(t *testing.T) {
+	dir := t.TempDir()
+	fh, err := NewMarkerFileHandler(log.NewNopLogger(), dir)
+	require.NoError(t, err)
+
+	// write first something to marker
+	markerFile := path.Join(dir, "remote", "segment")
```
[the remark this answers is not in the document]
thepalbi: I inherited the `remote` folder name from the PR you did for prometheus. Maybe we can re-name the folder to either something like:
- `internal` to give the sense that we store internal state stuff related to the WAL inner workings
- `state` to give the sense we track state stuff
- `markers`?

Apart from that, yes for making the constants.

component/common/loki/client/internal/marker_handler.go
```diff
@@ -0,0 +1,121 @@
...
+func (mh *markerHandler) UpdateSentData(segmentId, dataCount int) {
+	mh.dataIOUpdate <- data{
+		segmentId: segmentId,
+		dataCount: -1 * dataCount,
+	}
+}
+
+func (mh *markerHandler) Stop() {
+	// Firstly stop the Marker Handler, because it might want to use the Marker File Handler.
+	mh.quit <- struct{}{}
```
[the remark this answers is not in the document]
thepalbi: Yess!

component/loki/write/types.go
thepalbi: Handled in https://github.com/grafana/agent/pull/5434

component/common/loki/wal/watcher.go
thepalbi: Clarified

component/common/loki/wal/watcher.go
thepalbi: You are right actually, we can just log the warnings, and then return nil. Changing to that.

thepalbi:
@ptodev went through ~half the comments, and rebased over the latest changes in https://github.com/grafana/agent/pull/5434, which is already approved.

component/common/loki/client/internal/marker_file_handler.go
```diff
@@ -0,0 +1,87 @@
...
+	var (
+		segmentText = strconv.Itoa(segment)
+		tmp         = mfh.lastMarkedSegmentFilePath + ".tmp"
+	)
+
+	if err := os.WriteFile(tmp, []byte(segmentText), 0o666); err != nil {
+		level.Error(mfh.logger).Log("msg", "could not create segment marker file", "file", tmp, "err", err)
+		return
+	}
+	if err := fileutil.Replace(tmp, mfh.lastMarkedSegmentFilePath); err != nil {
```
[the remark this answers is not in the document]
thepalbi: Ok, just did a read over atomicity in renaming files. Let me split the answer in:
1. atomic marker swap
2. marker corruption
3. appending

For 1), just found here that if we use `os.Rename` the operation is [atomic in POSIX OSes](https://pkg.go.dev/os#Rename), and in Windows, according to [this](https://stackoverflow.com/questions/30385225/is-there-an-os-independent-way-to-atomically-overwrite-a-file), should be atomic if the source and destination are the same media (which they are in our case). I think we could use `os.Rename` and `os.CreateTemp` to implement the "atomic write" of the marker. In the Windows case, I think we could have a "this should work" for now, and then with some more digging and testing harden the implementation.

As for 2), agree! We should add either some footer or some checksum to the file to handle cases when a crash occurs during writing. However, if we introduce either a checksum or header, I think we should take the opportunity to define a format for the marker file then:

image

[source](https://app.diagrams.net/?src=about#G1TfBPXEcZY7sgOJXyfdaKZfI9vnfKvD_3)

For the error checking code, we can use [CRC](https://pkg.go.dev/hash/crc32) which is supported in Go. Also, if we were to change to an append only marker store, 12 byte is a not too big size of appended entries. Wdyt?

For 3) I guess we could, but this introduces the problem of the append-only marker to grow and grow, which long term will introduce us problems we need to tackle like marker rotation, or cropping it when it grows too large. If we can, I'll start with a simple swap marker like we have, and then since it's hidden behind the `MarkerHandler` we can rethink it.

component/common/loki/client/internal/marker_handler_test.go
thepalbi: Yes, I'll add this once we agree on how the marker file should look like 😀

component/common/loki/client/queue_client.go
thepalbi: Fixed comment, I was referring that it might block if the backing channel has no capacity.

component/common/loki/client/queue_client.go
thepalbi: Agree!

component/common/loki/client/internal/marker_file_handler.go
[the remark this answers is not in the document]
thepalbi: @ptodev wdyt of https://github.com/grafana/agent/pull/5655?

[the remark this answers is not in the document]
mattdurham: os.Rename isn't atomic with the defaults it uses, you need to use something like https://github.com/natefinch/atomic. Specifically it needs to pass in MOVEFILE_WRITE_THROUGH which golang doesn't. https://cs.opensource.google/go/go/+/master:src/cmd/vendor/golang.org/x/sys/windows/syscall_windows.go?q=windows.rename&ss=go%2Fgo

component/common/loki/client/internal/marker_file_handler.go
```diff
@@ -0,0 +1,96 @@
...
+
+var (
+	_ MarkerFileHandler = (*markerFileHandler)(nil)
+)
+
+// NewMarkerFileHandler creates a new markerFileHandler.
+func NewMarkerFileHandler(logger log.Logger, walDir string) (MarkerFileHandler, error) {
+	markerDir := filepath.Join(walDir, MarkerFolderName)
+	// attempt to create dir if doesn't exist
+	if err := os.MkdirAll(markerDir, 0o777); err != nil {
```
mattdurham, line 42: 777 feels a bit excessive.

component/common/loki/client/internal/marker_file_handler_test.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestMarkerFileHandler(t *testing.T) {
+	dir := t.TempDir()
+	fh, err := NewMarkerFileHandler(log.NewNopLogger(), dir)
+	require.NoError(t, err)
+
+	// write first something to marker
+	markerFile := path.Join(dir, MarkerFolderName, MarkerFileName)
```
mattdurham, line 18: should be filepath instead of file so the test runs on windows.

component/common/loki/client/internal/marker_handler.go
thepalbi: @ptodev responding on the comment above

component/common/loki/client/internal/marker_file_handler.go
thepalbi: what do you think should be a sensible perm for the folder?

component/common/loki/client/internal/marker_file_handler.go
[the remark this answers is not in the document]
thepalbi: Aha, I think that lib is small enough and with a simple enough API that we can add it. For POSIX it just calls `os.Rename`, and for windows it uses a diff syscall. On it!

component/common/loki/client/internal/marker_file_handler.go
mattdurham: I tend to go with 600/666, giving arbitrary files execute makes me shiver a bit.

component/common/loki/client/internal/marker_file_handler.go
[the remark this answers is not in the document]
mattdurham: Agreed

component/common/loki/client/internal/marker_file_handler.go
thepalbi: Let's go with 600, which is the most restrictive

component/common/loki/client/internal/marker_handler.go
```diff
@@ -0,0 +1,178 @@
...
+// The algorithm will find the highest numbered segment that is considered as "consumed", with its all predecessors
+// "consumed" as well.
+//
+// A consumed segment is one with data count of zero, meaning that there's no data left in flight for it, or it hasn't
+// received any updates for tooOldThreshold time.
+//
+// Also, while reviewing the data items in segmentDataCount, those who are consumed will be deleted to clean up space.
+//
+// This algorithm runs in O(N log N), being N the size of segmentDataCount, and allocates O(N) memory.
+func FindMarkableSegment(segmentDataCount map[int]countDataItem, tooOldThreshold time.Duration) int {
```
thepalbi, line 145: @ptodev one thing we can do to improve this after this first version, is use for `segmentDataCount` a data structure that is:
- a map
- support iteration over it's keys

I imagine something like a tree-based map that allows an in-order walk over it's items would work. But I repeat, for later and to minimize the N log N, which might not be even needed, since my idea is to run the mark segment not for each update we receive, but only every so much time (like limited by a ticker).

component/common/loki/client/internal/marker_file_handler.go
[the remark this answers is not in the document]
ptodev: Ok - I'm happy to leave it as it is for now. I agree that this is a reasonable first implementation.

component/common/loki/client/internal/marker_file_handler_test.go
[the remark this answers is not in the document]
ptodev: I inherited those names from previous PRs which tried to implement this :) I suppose the reason behind the name "remote" is that it indicates it contains files for the remote write package. I think it does make sense to call the folder something which is related to remote write. 

I think the file name "segment" is ambiguous though. Maybe something like "wal_segment" would be more clear.

component/common/loki/client/internal/marker_file_handler.go
ptodev, line 1: If someone deletes the WAL, but does not delete the marker file, would the WAL replay code work ok? I suppose in that case the code should know to ignore the segment ID in the marker file?

component/common/loki/client/internal/marker_handler.go
ptodev: Thank you, the new algorithm makes sense. I'm just not sure whether the `tooOldThreshold` should be configurable. Atm it is hardcoded to one hour - I wonder why this value was chosen? It'd be good to clarify it in a comment please.

component/common/loki/client/internal/marker_file_handler.go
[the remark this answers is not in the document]
thepalbi: Closing this thread with agreement on marker file format and atomic swap, which is merged in https://github.com/grafana/agent/pull/5590/commits/97e87dee9216164c7364acae344f050f8ecbab3c

component/common/loki/client/internal/marker_file_handler_test.go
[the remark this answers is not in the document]
thepalbi: Agree on remote, and let's use `marked_segment`, to make it explicit it's the marker

component/common/loki/client/internal/marker_handler.go
thepalbi: Oh yhea forgot to parametrize the value, added `time.Hour` for testing.

thepalbi: Used [`MaxSegmentAge`](https://github.com/grafana/agent/blob/main/component/common/loki/wal/config.go#L31) to configure the tooOldThreshold

component/common/loki/client/internal/marker_file_handler.go
thepalbi: So the marking logic is not connected to the WAL, so it can tell if it was deleted or not. I guess for WAL deletions we should advice that the marker has to be deleted as well. 

On the other hand, the watcher, once it reads from the marker, tries to find the following segment by looking into disk, so since the WAL doesn't exists it will enter a retry loop until the WAL get's re-created.

thepalbi: I think we could add an additional logic to the marker interface, so the WAL can force write to the marker file. Like if it detects WAL deletion, force mark a segment

component/common/loki/client/internal/marker_file_handler.go
```diff
@@ -0,0 +1,107 @@
...
+	if err := os.MkdirAll(markerDir, MarkerFolderMode); err != nil {
+		return nil, fmt.Errorf("error creating segment marker folder %q: %w", markerDir, err)
+	}
+	// do we need this bit below?
+	//if stat, err := os.Stat(markerDir); err == nil && stat.Mode().Perm() != MarkerFileMode {
+	//	if err := os.Chmod(markerDir, MarkerFileMode); err != nil {
+	//		return nil, err
+	//	}
+	//}
+
```
thepalbi, line 55: @mattdurham do you think we need this bit? Like if the folder exists, double check it has the perms we want, and if not, do a chmod?

component/common/loki/client/internal/marker_handler.go
```diff
@@ -0,0 +1,211 @@
...
+		}
+
+		if !shouldRunFind {
+			continue
+		}
+
+		markableSegment := FindMarkableSegment(segmentDataCount, mh.maxSegmentAge)
+		level.Debug(mh.logger).Log("msg", fmt.Sprintf("found as markable segment %d", markableSegment))
+		if markableSegment > mh.lastMarkedSegment {
+			mh.markerFileHandler.MarkSegment(markableSegment)
```
thepalbi, line 155: @thepalbi to add metric tracking this here

thepalbi:
@ptodev @mattdurham just added on top of the previous queue client benchmarks, once comparing the nil marker and real marker implementations. No significant diff whatsoever:
```
goos: darwin
goarch: arm64
pkg: github.com/grafana/agent/component/common/loki/client
                                                                    │ queue_nil_marker_handler │       queue_marker_handler       │
                                                                    │          sec/op          │   sec/op    vs base              │
ClientImplementations/100_entries,_single_series,_no_batching-10                    1.001 ± 0%   1.001 ± 0%       ~ (p=0.394 n=6)
ClientImplementations/100k_entries,_100_series,_default_batching-10                 4.292 ± 1%   4.336 ± 1%  +1.02% (p=0.026 n=6)
geomean                                                                             2.073        2.084       +0.50%

                                                                    │ queue_nil_marker_handler │        queue_marker_handler        │
                                                                    │           B/op           │     B/op      vs base              │
ClientImplementations/100_entries,_single_series,_no_batching-10                  861.0Ki ± 5%   861.6Ki ± 5%       ~ (p=0.615 n=6)
ClientImplementations/100k_entries,_100_series,_default_batching-10               788.9Mi ± 0%   788.0Mi ± 0%  -0.12% (p=0.002 n=6)
geomean                                                                           25.75Mi        25.75Mi       -0.02%

                                                                    │ queue_nil_marker_handler │       queue_marker_handler        │
                                                                    │        allocs/op         │  allocs/op   vs base              │
ClientImplementations/100_entries,_single_series,_no_batching-10                   11.26k ± 1%   11.28k ± 0%       ~ (p=0.058 n=6)
ClientImplementations/100k_entries,_100_series,_default_batching-10                10.80M ± 0%   10.80M ± 0%  -0.01% (p=0.002 n=6)
geomean                                                                            348.8k        349.1k       +0.09%
```

component/common/loki/client/internal/marker_file_handler.go
mattdurham: No, since we are within the agents datapath we can assume we own it.

component/common/loki/client/batch.go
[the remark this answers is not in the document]
thepalbi: Tests, comments and bench added

component/common/loki/client/internal/marker_file_handler.go
thepalbi: Just added the logic, so will remove it!

component/common/loki/client/internal/marker_handler.go
thepalbi: Added

component/common/loki/wal/watcher_metrics.go
```diff
@@ -58,6 +59,15 @@ func NewWatcherMetrics(reg prometheus.Registerer) *WatcherMetrics {
 			},
 			[]string{"id"},
 		),
+		replaySegment: prometheus.NewGaugeVec(
```
mattdurham, line 62: I have some additional metrics that would be handy but if you want to do them in another more focused PR thats fine. WAL size in bytes, and WAL oldest timestamp and current timestamp. Kind of like segment but its more actionable for alerts.

thepalbi: Down for checking that in another PR, but since our WAL writer uses prometheus'es wlog implementation, we inherit all this metrics https://github.com/prometheus/prometheus/blob/3450572840881c870012ef80bff1877568698638/tsdb/wlog/wlog.go#L196-L205. I think we don't have timestamps, so that one might be interesting for alerting on WAL not being written or too far behind (like to measure lag from wal write to watcher read).

thepalbi:
@mattdurham @ptodev build is green, and all convos are resolved, can you take a final pass when you have some time?

component/common/loki/client/internal/marker_encoding.go
```diff
@@ -0,0 +1,55 @@
...
+var (
+	markerHeaderV1 = []byte{'0', '1'}
+)
+
+// EncodeMarkerV1 encodes the segment number, from whom we need to create a marker, in the marker file format,
+// which in v1 includes the segment number and a trailing CRC code of the first 10 bytes.
+func EncodeMarkerV1(segment uint64) ([]byte, error) {
+	// marker format v1
+	// marker [ 0 , 1 ] - HEADER, which is used to track version
+	// marker [ 2 , 9 ] - encoded unit 64 which is the content of the marker, the last "consumed" segment
```
mattdurham, line 18: nit: unit -> uint

component/common/loki/wal/watcher_test.go
```diff
@@ -52,6 +52,26 @@ func (t *testWriteTo) AppendEntries(entries wal.RefEntries, _ int) error {
...
+	for _, e := range t.ReadEntries.StartIterate() {
+		if _, ok := seen[e.Line]; ok {
+			seen[e.Line] = true
+		}
+	}
+	t.ReadEntries.DoneIterate()
+
+	allSeen := true
+	for _, wasSeen := range seen {
+		allSeen = allSeen && wasSeen
```
mattdurham, line 69: nit: simplify to moving the require.True here for wasSeen, could add the line that was missing for better context.

mattdurham (changes requested):

mattdurham (approved):
There are some small nits and a few side conversations but looking solid.

PR #5603, opened 2023-10-25
Replace relrefs with docrefs in Agent concepts and Agent monitoring topics

clayton-cornell:
#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/flow/concepts/clustering.md
+++ b/docs/sources/flow/concepts/clustering.md
@@ -28,8 +28,6 @@ single-node cluster.
 You configure clustering by passing `cluster` command-line flags to the [run][]
 command.
 
-[run]: {{< relref "../reference/cli/run.md#clustering-beta" >}}
-
 ## Use cases
 
 ### Target auto-distribution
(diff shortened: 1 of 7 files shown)


ptodev (approved):
I'm rubber stamping the PR, but I have a few concerns regarding content duplication:
* Are the Grafana Cloud docs page just a copy pasted version of the Agent docs? Or are they imported from the Agent repo?
* If the GC docs are copy pasted, how do we keep them in sync?
* How does GC decide which version of the Agent to use for docs?
* I don't see a Flow section in the GC Agent docs? I'm not sure why we are adding GC links even though they don't exist? Is it for future proofing?

I personally really think GC should just link to the Agent docs instead of duplicating the content... I can't see an advantage in duplication. Maybe the intent is to make it easier for GC users to find. But I think it's even confusing to the users because when they search our docs they see duplicate pages with the same information.

clayton-cornell:
@ptodev Some answers

1. The Agent docs are mounted in Cloud, not copy/pasted. That means a single source and multiple publish locations.
2. They stay in sync because we have one source of information :-)
3. Cloud uses latest (as far as I know)
4. Flow is not yet mounted in Cloud docs. It was, and we disconnected it. Currently only Static is mounted there.
5. The work here is preparing when we actually do mount Flow docs into Cloud. Relrefs work with a single mount point - currently just the OSS docs. Things can get weird or broken when you have single source docs mounted in multiple locations with non-identical pathing.

ptodev:
@clayton-cornell thank you for clarifying! IMO the approach of always using the latest version works well for Mimir/Loki/Tempo, but it doesn't work so well for the Agent. It's because the databases on the Cloud are naturally always on the latest version, but the Agent runs on the client's machines and the clients may run any version of it.

PR #5618, opened 2023-10-26
Add WAL corruption information and fix relrefs

clayton-cornell:
#### PR Description

This PR adds information about deleting corrupted WALs.

Fixes [#7866](https://github.com/grafana/support-escalations/issues/7866)

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/shared/wal-data-retention.md
+++ b/docs/sources/shared/wal-data-retention.md
@@ -27,7 +27,7 @@ for remote writing. If that data has not yet been pushed to the remote
 endpoint, it is lost.
 
 This behavior dictates the data retention for the `prometheus.remote_write`
-component. It also means that it is impossible to directly correlate data
+component. It also means that it's impossible to directly correlate data
 retention directly to the data age itself, as the truncation logic works on
 _segments_, not the samples themselves. This makes data retention less
 predictable when the component receives a non-consistent rate of data.
@@ -48,7 +48,7 @@ removed. The `max_keepalive_time` or `max_wal_time` controls the maximum age of
 samples that can be kept in the WAL. Samples older than
 `max_keepalive_time` are forcibly removed.
 
-### In cases of `remote_write` outages
+### Extended `remote_write` outages
 When the remote write endpoint is unreachable over a period of time, the most
 recent successfully sent timestamp is not updated. The
 `min_keepalive_time` and `max_keepalive_time` arguments control the age range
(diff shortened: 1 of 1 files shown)


docs/sources/shared/wal-data-retention.md
```diff
@@ -74,5 +74,32 @@ If the component falls behind more than one third of the data written since the
...
+and crashes before you can stop Grafana Agent and other running services. When you restart Grafana
+Agent, the Agent verifies the WAL, removing any corrupt segments it finds. Sometimes, this repair
+is unsuccessful, and you must manually delete the corrupted WAL to continue.
+
+When the WAL becomes corrupted, Grafana Agent writes error messages such as
+`err="failed to find segment for index"` to the log file.
+
+{{% admonition type="caution" %}}
+Deleting a WAL segment or a WAL file permanently deletes the stored WAL data.
+{{% /admonition %}}
```
rfratto, line 90: Stored WAL data isn't replayed on restart anyway, so this caution isn't that severe (at least until this behavior is changed)

docs/sources/shared/wal-data-retention.md
```diff
@@ -74,5 +74,32 @@ If the component falls behind more than one third of the data written since the
...
+
+To delete the corrupted WAL:
+
+1. [Stop][] Grafana Agent.
```
rfratto, line 94: an aside: I'm actually not sure it's necessary to stop the agent before deleting the WAL. We regularly delete it against running agents, but I think we typically restart anyway "just to be safe."

It's probably better to leave this in just so things are more predictable.

docs/sources/shared/wal-data-retention.md
```diff
@@ -74,5 +74,32 @@ If the component falls behind more than one third of the data written since the
...
+1. Delete the WAL from disk. By default, the WAL is in the `data-agent` directory located in
```
rfratto, line 95: They should be careful to not delete the _entire_ directory, especially in Flow mode, since it has other things. 

They should look for the subdirectory called `wal` (which may be nested inside several other folders in Flow mode) and delete that.

rfratto (approved):
LGTM with some comments

docs/sources/shared/wal-data-retention.md
clayton-cornell: @rfratto I could convert it to a note... or remove it altogether.

docs/sources/shared/wal-data-retention.md
clayton-cornell: @rfratto So something like 

`Find and delete the contents of the wal directory.  By default the wal directory is a subdirectory of data-agent located in the Grafana Agent working directory.`

Might need a little tweaking but something along these lines?

rfratto: Yep. There will be one `wal` directory per:

* (Static mode) Metrics instance running 
* (Flow mode) `prometheus.remote_write` component running

PR #5619, opened 2023-10-26
Promtail sync loki source api

### LINKED ISSUE #4839: Sync loki.source.api with Promtail 2.8.4

wildum:
#### PR Description

Nothing new since the component was added (https://github.com/grafana/agent/pull/3648): https://github.com/grafana/loki/commits/main/clients/pkg/promtail/targets/lokipush

The only change would be the one coming from the last commit but we would need to update the loki version to match it.

Fixes #4839

### DIFF
--- a/component/loki/source/api/internal/lokipush/push_api_server.go
+++ b/component/loki/source/api/internal/lokipush/push_api_server.go
@@ -132,7 +132,7 @@ func (s *PushAPIServer) getRelabelRules() []*relabel.Config {
 	return newRules
 }
 
-// NOTE: This code is copied from Promtail (3478e180211c17bfe2f3f3305f668d5520f40481) with changes kept to the minimum.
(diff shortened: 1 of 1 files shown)


erikbaranowski:
@tpaschalis was there something specific we were looking to sync here or was that simply a maintenance issue for research and updates? https://github.com/grafana/agent/issues/4839

tpaschalis:
> was there something specific we were looking to sync here or was that simply a maintenance issue for research and updates

No, I didn't have something specific in mind, just opened an issue for all `loki.*` components so we look in case there's any updates.

Since `loki.source.api` reuses the extracted weaveworks server code and there doesn't seem to be any new updates to the `lokipush` package, I think we can merge this as-is and close the original issue.

tpaschalis:
> The only change would be the one coming from the last commit but we would need to update the loki version to match it.


cc @wildum this looks like it's an interesting feature (ingest OTLP directly from loki.source.api without needing to have a new OTLP receiver component defined).

I'm worried this might slip through the cracks, do you mind opening a tracking issue to look into how we want to expose this to Flow users?

tpaschalis (approved):

PR #5623, opened 2023-10-26
Flow: Improve Kubernetes log collection 

### LINKED ISSUE #4502: Make `loki.source.kubernetes` clusterable in flow-mode
### Request

I would like the loki source for kubernetes to be able to distribute the targets across nodes in clustered mode on the Grafana-Agent when deployed with Flow mode.

I'm not entirely sure if this is supported already. There is mentioned on `loki.source.podlogs` that `loki.source.kubernetes` can distribute load. However, I can't find any reference in the code or the docs to how the `loki.source.kubernetes` distributes the targets when in clustered mode.

### Use case

Due to the number of components, a single Grafana Agent is not strong enough to handle all logging concerns. We need to be able to scale the agent without duplicate log ingestion.

### LINKED ISSUE #5040: `loki.source.kubernetes` should handle log rotation
### Background

Over on the k8s monitoring helm repo, we found that [`loki.source.kubernetes` stops tailing logs after log file rotation](https://github.com/grafana/k8s-monitoring-helm/issues/61).

### Proposal

Test again once the underlying issue in Kubernetes being tracked here is resolved:

- https://github.com/kubernetes/kubernetes/pull/118500

rfratto:
> **NOTE**: This PR is a combination of four smaller PRs, made in collaboration with @slim-bean. It'll be easier to review this PR commit-by-commit. 

### component/prometheus: fix panic in interceptor when child isn't set

This commit fixes a panic in prometheus.Interceptor where an interceptor which doesn't forward samples to another appendable panics when appending data.

Authored by @slim-bean 

### loki.source.kubernetes: improve detection of rolled log files

Versions of Kubernetes that do not contain kubernetes/kubernetes#115702 will fail to detect rolled log files, causing the API to stop sending logs to the agent for processing.

To work around this, this commit introduces a rolling average calculator to determine the average delta between log entries per target. If 3x the normal delta time has elapsed since the last entry, the tailer is restarted.

False positives here are acceptable, but false negatives mean that log lines may not appear for an extended period of time until the rolling detection succeeds.

Authored by @slim-bean 

Closes #5040

### loki.source.kubernetes: support clustering

Add support for loki.source.kubernetes to distribute targets using clustering.

Closes #4502

Authored by @slim-bean 

### loki.source.podlogs: support clustering 

Add support for loki.source.podlogs to distribute targets using clustering.

Authored by @rfratto

### DIFF
--- a/component/loki/source/kubernetes/kubernetes.go
+++ b/component/loki/source/kubernetes/kubernetes.go
@@ -43,13 +43,7 @@ type Arguments struct {
 	// Client settings to connect to Kubernetes.
 	Client commonk8s.ClientArguments `river:"client,block,optional"`
 
-	Clustering Clustering `river:"clustering,block,optional"`
-}
-
-// Clustering holds values that configure clustering-specific behavior.
-type Clustering struct {
-	// TODO(@tpaschalis) Move this block to a shared place for all components using clustering.
-	Enabled bool `river:"enabled,attr"`
+	Clustering cluster.ComponentBlock `river:"clustering,block,optional"`
 }
 
 // DefaultArguments holds default settings for loki.source.kubernetes.
--- a/component/loki/source/kubernetes/kubernetes.go
+++ b/component/loki/source/kubernetes/kubernetes.go
@@ -18,13 +18,15 @@ import (
 	"github.com/grafana/agent/component/discovery"
 	"github.com/grafana/agent/component/loki/source/kubernetes/kubetail"
 	"github.com/grafana/agent/pkg/flow/logging/level"
+	"github.com/grafana/agent/service/cluster"
 	"k8s.io/client-go/kubernetes"
 )
 
 func init() {
 	component.Register(component.Registration{
-		Name: "loki.source.kubernetes",
-		Args: Arguments{},
+		Name:          "loki.source.kubernetes",
+		Args:          Arguments{},
+		NeedsServices: []string{cluster.ServiceName},
 
 		Build: func(opts component.Options, args component.Arguments) (component.Component, error) {
 			return New(opts, args.(Arguments))
@@ -40,6 +42,14 @@ type Arguments struct {
 
 	// Client settings to connect to Kubernetes.
 	Client commonk8s.ClientArguments `river:"client,block,optional"`
+
+	Clustering Clustering `river:"clustering,block,optional"`
+}
+
+// Clustering holds values that configure clustering-specific behavior.
+type Clustering struct {
+	// TODO(@tpaschalis) Move this block to a shared place for all components using clustering.
+	Enabled bool `river:"enabled,attr"`
 }
 
 // DefaultArguments holds default settings for loki.source.kubernetes.
@@ -57,6 +67,7 @@ type Component struct {
 	log       log.Logger
 	opts      component.Options
 	positions positions.Positions
+	cluster   cluster.Cluster
 
 	mut         sync.Mutex
 	args        Arguments
@@ -72,6 +83,7 @@ type Component struct {
 var (
 	_ component.Component      = (*Component)(nil)
 	_ component.DebugComponent = (*Component)(nil)
+	_ cluster.Component        = (*Component)(nil)
 )
 
 // New creates a new loki.source.kubernetes component.
@@ -89,7 +101,13 @@ func New(o component.Options, args Arguments) (*Component, error) {
 		return nil, err
 	}
 
+	data, err := o.GetServiceData(cluster.ServiceName)
+	if err != nil {
+		return nil, err
+	}
+
 	c := &Component{
+		cluster:   data.(cluster.Cluster),
 		log:       o.Logger,
 		opts:      o,
 		handler:   loki.NewLogsReceiver(),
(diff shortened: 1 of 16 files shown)


component/loki/source/kubernetes/kubernetes.go
```diff
@@ -168,28 +186,43 @@ func (c *Component) Update(args component.Arguments) error {
...
 		if err != nil {
 			// TODO(rfratto): should this set the health of the component?
 			level.Error(c.log).Log("msg", "failed to process input target", "target", lset.String(), "err", err)
 			continue
 		}
-		targets = append(targets, kubetail.NewTarget(lset, processed))
+		tailTargets = append(tailTargets, kubetail.NewTarget(lset, processed))
 	}
 
 	// This will never fail because it only fails if the context gets canceled.
```
captncraig, line 210: Since this is called while holding the component mutex, I'd say yes, we should have a timeout. Even if we can only log it.

component/loki/source/kubernetes/kubernetes.go
```diff
@@ -168,28 +186,43 @@ func (c *Component) Update(args component.Arguments) error {
...
+	targets = distTargets.Get()
 
-	for _, inTarget := range newArgs.Targets {
-		lset := inTarget.Labels()
+	tailTargets := make([]*kubetail.Target, 0, len(targets))
+	for _, target := range targets {
+		lset := target.Labels()
 		processed, err := kubetail.PrepareLabels(lset, c.opts.ID)
 		if err != nil {
 			// TODO(rfratto): should this set the health of the component?
```
captncraig, line 203: In what situations will PrepareLabels error? Is there a case where we discard a target and increment a counter or something?

component/loki/source/kubernetes/kubetail/tailer.go
```diff
@@ -167,9 +169,51 @@ func (t *tailer) tail(ctx context.Context, handler loki.EntryHandler) error {
...
+				// False positives here are acceptable, but false negatives mean that
+				// we'll have a larger spike of missing logs until we detect a rolled
+				// file.
+				avg := calc.GetAverage()
+				last := calc.GetLast()
+				if last.IsZero() {
+					continue
+				}
+				s := time.Since(last)
+				if s > avg*3 {
```
captncraig, line 211: We could potentially add a "sensitivity" arg for this to let them be more or less reactive. Would that be useful, or just confusing?

component/loki/source/kubernetes/kubetail/tailer.go
```diff
@@ -167,9 +169,51 @@ func (t *tailer) tail(ctx context.Context, handler loki.EntryHandler) error {
...
+
+	// Create a new rolling average calculator to determine the average delta
+	// time between log entries.
+	//
+	// Here, we track the most recent 10,000 delta times to compute a fairly
+	// accurate average. If there are less than 100 deltas stored, the average
+	// time defaults to 1h.
+	//
+	// The computed average will never be less than the minimum of 2s.
+	calc := newRollingAverageCalculator(10000, 100, 2*time.Second, maxTailerLifetime)
```
captncraig, line 181: Does that mean the max refresh interval is 6 seconds? I think we won't thrash frequent updates since this makes a new tailer each time, and that takes time to reach stable averages?

component/loki/source/kubernetes/kubetail/tailer.go
rfratto: I'd prefer to keep this locked down for now, but we can expose it as an argument to end users later if it turns out to be necessary.

component/loki/source/kubernetes/kubetail/tailer.go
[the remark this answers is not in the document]
rfratto: > thats a good chunk of memory for each target. 

It's 80KB per target, right? A time.Duration is a uint64 (8 bytes), meaning this is 80,000 bytes per target. 

I'm not sure I'm convinced yet that 80KB is a huge overhead for a single target, especially since once you have thousands of targets I'm not sure this would be the largest portion of the agent's memory usage. 

We can also tune this over time if the overhead does appear to be significant; I'm not sure if there was any particular reason @slim-bean chose these specific values. 

> If there was a good way to detect k8s version or something, it could reduce memory at the cost of complexity. Maybe not worth it?

I'd be happy to do this, but I'm not sure that the fix is released yet?

component/loki/source/kubernetes/kubernetes.go
```diff
@@ -40,6 +42,14 @@ type Arguments struct {
 
 	// Client settings to connect to Kubernetes.
 	Client commonk8s.ClientArguments `river:"client,block,optional"`
+
+	Clustering Clustering `river:"clustering,block,optional"`
+}
+
+// Clustering holds values that configure clustering-specific behavior.
+type Clustering struct {
```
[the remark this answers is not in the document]
rfratto: @tpaschalis Did you have a specific idea for how to deduplicate this work? 

I'm happy to do it in this PR if you want to see it but I'm also worried about blocking this with more rounds of review.

component/loki/source/kubernetes/kubernetes.go
rfratto: I'm nervous about that; maybe it should be detached from the mutex instead. 

If it times out and fails, that means that the targets won't ever converge unless there's another discovery update or the cluster updates again. While discovery updates are common for us, it might not be common in more stable environments. 

So, the risk of a timeout causing targets to never converge is worrying me. WDYT?

component/loki/source/kubernetes/kubernetes.go
rfratto: Looks like it fails based on validations:

> // Validation of lset fails if there is no label indicating the pod namespace,
> // pod name, or container name.

In this particular case, the component would've been given a bad target.

component/loki/source/kubernetes/kubetail/tailer.go
rfratto: No, the _minimum_ refresh interval is 6 seconds; the default maximum is 3h, and with the way this is written it's not possible for that to get any higher. 

It's not possible to go above 3h because you need at least 100 log lines for the average to be calculated; logging less frequently than once every 3 hours means the tailer would've already been reset (including resetting the calculator) before the true average was detected.

captncraig (approved):
Anything else I put is really optional. Super excited to play with this.

mattdurham (approved):
Fantastic!

clayton-cornell (approved):
Docs are OK as-is

component/loki/source/kubernetes/kubernetes.go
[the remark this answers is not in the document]
tpaschalis: I think this type should probably reside in the cluster service package.

I'm not against doing it on this PR, and it sounds like an easy-enough change so that it won't be a blocker.

component/loki/source/podlogs/reconciler.go
```diff
@@ -101,6 +126,31 @@ func (r *reconciler) Reconcile(ctx context.Context, cli client.Client) error {
...
 
+func distributeTargets(c cluster.Cluster, targets []*kubetail.Target) []*kubetail.Target {
+	if c == nil {
+		return targets
+	}
+
+	res := make([]*kubetail.Target, 0, (len(targets)+1)/len(c.Peers()))
+
+	// TODO(@tpaschalis): Make sure OpReadWrite is the correct operation;
+	// eg. this determines how clustering behaves when nodes are shutting down.
```
tpaschalis, line 137: OpReadWrite _is_ the correct operation here, I think we can remove this TODO comment

component/loki/source/kubernetes/kubetail/tailer.go
```diff
@@ -167,9 +169,51 @@ func (t *tailer) tail(ctx context.Context, handler loki.EntryHandler) error {
...
+				// file.
+				avg := calc.GetAverage()
+				last := calc.GetLast()
+				if last.IsZero() {
+					continue
+				}
+				s := time.Since(last)
+				if s > avg*3 {
+					level.Info(t.log).Log("msg", "have not seen a log line in 3x average time between lines, closing and re-opening tailer", "rolling_average", avg, "time_since_last", s)
+					return
```
tpaschalis, line 213: Just to make sure I get this correctly, this will refresh the tailer, because the deferred function will call stream.Close(), causing line 263 below to return and the `retailBackoff` triggering another tailer to start, right?

tpaschalis (approved):
LGTM, looking forward to make use of this!

component/loki/source/kubernetes/kubetail/tailer.go
rfratto: Yes 👍

PR #5634, opened 2023-10-27
Add an "exclude_event_message" argument to loki.source.windowsevent

### LINKED ISSUE #4851: Sync loki.source.windowsevent with Promtail 2.8.4

ptodev:
#### PR Description

Syncs the `loki.source.windowsevent` flow component with a [change](https://github.com/grafana/loki/pull/7462) done in promtail a few months ago. Interestingly, this config entry exists in the Agent codebase but was always set to false. I don't know why that is the case. Probably it was an oversight.

Fixes #4851

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,9 @@ Main (unreleased)
 - Added an `add_metric_suffixes` option to `otelcol.exporter.prometheus` in flow mode, 
   which configures whether to add type and unit suffixes to metrics names. (@mar4uk)
 
+- Added an `exclude_event_message` option to `loki.source.windowsevent` in flow mode,
+  which excludes the human-friendly event message from Windows event logs. (@ptodev)
+
 v0.37.3 (2023-10-26)
 -----------------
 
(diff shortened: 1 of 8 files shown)


wildum (approved):
LGTM, good catch!

tpaschalis (approved):
LGTM. This looks like it's the only change since the original introduction of the component.

Coule you please add a `// NOTE:` header so we keep track of the latest Promtail commit we've checked for changes against (in this case, bde65667f7c88af17b7729e3621d7bd5d1d3b45f) to make our job easier if we have to re-sync these components?

ptodev:
> Coule you please add a `// NOTE:` header so we keep track of the latest Promtail commit we've checked for changes against (in this case, bde65667f7c88af17b7729e3621d7bd5d1d3b45f) to make our job easier if we have to re-sync these components?

@tpaschalis  Shouldn't it be 07cbef92268a? That's the version of `github.com/grafana/loki` which we have in the go.mod.

I'm happy to add such a comment, but tbh I'm not sure how much I would trust it in the future. I think this approach to updating Promtail dependencies definitely needs a bigger reform. E.g. it'd be nice if we could at least use specific versions of Promtail instead of commits.

clayton-cornell (approved):
Docs look good

PR #5642, opened 2023-10-28
Sortable header for component list in web UI

hainenber:

Implements #5616 

#### Notes to the Reviewer
Please check if the screen recording is as expected or something more formal needs to be done, i.e. tests

https://github.com/grafana/agent/assets/41283691/b037e1e4-efa2-47ad-b6f2-89012f3a6652

#### PR Checklist

- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -118,9 +118,6 @@ Main (unreleased)
 
 - Make component list sortable in web UI. (@hainenber)
 
-- Added an `exclude_event_message` option to `loki.source.windowsevent` in flow mode,
-  which excludes the human-friendly event message from Windows event logs. (@ptodev)
-
 - Improve detection of rolled log files in `loki.source.kubernetes` and
   `loki.source.podlogs` (@slim-bean).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -116,8 +116,6 @@ Main (unreleased)
 - Added an `exclude_event_message` option to `loki.source.windowsevent` in flow mode,
   which excludes the human-friendly event message from Windows event logs. (@ptodev)
 
-- Make component list sortable in web UI. (@hainenber)
-
 - Improve detection of rolled log files in `loki.source.kubernetes` and
   `loki.source.podlogs` (@slim-bean).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -122,6 +122,8 @@ Main (unreleased)
 
 - Support clustering in `loki.source.podlogs` (@rfratto).
 
+- Make component list sortable in web UI. (@hainenber)
+
 v0.37.3 (2023-10-26)
 -----------------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -115,6 +115,12 @@ Main (unreleased)
 
 - Added an `exclude_event_message` option to `loki.source.windowsevent` in flow mode,
   which excludes the human-friendly event message from Windows event logs. (@ptodev)
+
+- Make component list sortable in web UI. (@hainenber)
+
+- Added an `exclude_event_message` option to `loki.source.windowsevent` in flow mode,
+  which excludes the human-friendly event message from Windows event logs. (@ptodev)
+
 - Improve detection of rolled log files in `loki.source.kubernetes` and
   `loki.source.podlogs` (@slim-bean).
 
--- a/web/ui/src/features/component/Table.module.css
+++ b/web/ui/src/features/component/Table.module.css
@@ -33,10 +33,6 @@
   width: 2%;
 }
 
-.table tr th[data-sort-order="default"] {
-  cursor: pointer;
-}
-
 .table tr th[data-sort-order="desc"] {
   cursor: pointer;
   box-shadow: none !important;
--- a/web/ui/src/features/component/Table.module.css
+++ b/web/ui/src/features/component/Table.module.css
@@ -32,3 +32,25 @@
 .table td:first-child {
   width: 2%;
 }
+
+.table tr th[data-sort-order="default"] {
+  cursor: pointer;
+}
+
+.table tr th[data-sort-order="desc"] {
+  cursor: pointer;
+  box-shadow: none !important;
+    &:before {
+      content: "▼";
+      float: right;
+    }
+}
+
+.table tr th[data-sort-order="asc"] {
+  cursor: pointer;
+  box-shadow: none !important;
+    &:before {
+      content: "▲";
+      float: right;
+    } 
+}
--- a/web/ui/src/features/component/TableHead.tsx
+++ b/web/ui/src/features/component/TableHead.tsx
@@ -15,7 +15,7 @@ const TableHead = ({ headers, handleSorting }: Props) => {
     // User clicks on the new header, use default ASC sort order
     let sortOrder = SortOrder.ASC;
 
-    // User clicks again on the header we toggled the sort order previously
+    // User clicks again on the header, we toggle the previous sort order
     if (header === sortField) {
       sortOrder = order === SortOrder.ASC ? SortOrder.DESC : SortOrder.ASC;
     }
--- a/web/ui/src/features/component/TableHead.tsx
+++ b/web/ui/src/features/component/TableHead.tsx
@@ -23,12 +23,11 @@ const TableHead = ({ headers, handleSorting }: Props) => {
   return (
     <tr>
       {headers.map((header) => {
-        const sortOrderHeaderAttribute = sortField === header ? order : 'default';
         return (
           <th
             key={header}
             onClick={() => handleSortingChange(header)}
-            data-sort-order={handleSorting ? sortOrderHeaderAttribute : undefined}
+            data-sort-order={handleSorting && sortField === header ? order : undefined}
           >
             {header}
           </th>
--- a/web/ui/src/features/component/TableHead.tsx
+++ b/web/ui/src/features/component/TableHead.tsx
@@ -1,0 +1,39 @@
+import { useState } from 'react';
+
+interface Props {
+  headers: string[];
+  handleSorting?: (header: string, sortOrder: string) => void;
+}
+
+const TableHead = ({ headers, handleSorting }: Props) => {
+  const [sortField, setSortField] = useState('');
+  const [order, setOrder] = useState('asc');
+
+  const handleSortingChange = (header: string) => {
+    const sortOrder = header === sortField && order === 'asc' ? 'desc' : 'asc';
+    setSortField(header);
+    setOrder(sortOrder);
+    if (handleSorting !== undefined) {
+      handleSorting(header, sortOrder);
+    }
+  };
+
+  return (
+    <tr>
+      {headers.map((header) => {
+        const sortOrderHeaderAttribute = sortField === header ? order : 'default';
+        return (
+          <th
+            key={header}
+            onClick={() => handleSortingChange(header)}
+            data-sort-order={handleSorting ? sortOrderHeaderAttribute : undefined}
+          >
+            {header}
+          </th>
+        );
+      })}
+    </tr>
+  );
+};
+
+export default TableHead;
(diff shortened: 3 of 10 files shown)


web/ui/src/features/component/TableHead.tsx
```diff
@@ -0,0 +1,39 @@
+import { useState } from 'react';
+
+interface Props {
+  headers: string[];
+  handleSorting?: (header: string, sortOrder: string) => void;
+}
+
+const TableHead = ({ headers, handleSorting }: Props) => {
+  const [sortField, setSortField] = useState('');
+  const [order, setOrder] = useState('asc');
```
wildum, line 10: I think that the order could be either a bool that you could just toggle with setOrder(!order). But if you want to keep an explicit state then I would suggest using an enum rather than a string

web/ui/src/features/component/TableHead.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+  const handleSortingChange = (header: string) => {
+    const sortOrder = header === sortField && order === 'asc' ? 'desc' : 'asc';
+    setSortField(header);
+    setOrder(sortOrder);
+    if (handleSorting !== undefined) {
```
wildum, line 16: the setters should be inside of the if statement because if handleSorting is undefined then there is not point of changing the value of sortField and order.

web/ui/src/pages/PageComponentList.tsx
```diff
@@ -5,11 +5,32 @@ import Page from '../features/layout/Page';
...
+
+  // TODO: make this sorting logic reusable
+  const handleSorting = (sortField: string, sortOrder: string): void => {
+    if (sortField) {
+      const sorted = [...components].sort((a, b) => {
+        const sortValueA = sortField === 'Health' ? a.health.state.toString() : a.localID;
+        const sortValueB = sortField === 'Health' ? b.health.state.toString() : b.localID;
+        if (sortValueA === null) return 1;
+        if (sortValueB === null) return -1;
+        if (sortValueA === null && sortValueB == null) return 0;
```
wildum, line 18:
```suggestion
```
this line can be removed because you are checking the null values above already

web/ui/src/pages/PageComponentList.tsx
```diff
@@ -5,11 +5,32 @@ import Page from '../features/layout/Page';
 import { useComponentInfo } from '../hooks/componentInfo';
 
 function PageComponentList() {
-  const components = useComponentInfo('');
+  const [components, setComponents] = useComponentInfo('');
+
+  // TODO: make this sorting logic reusable
+  const handleSorting = (sortField: string, sortOrder: string): void => {
+    if (sortField) {
```
wildum, line 12: to avoid nesting you could have if (!sortfield) return;

web/ui/src/pages/PageComponentList.tsx
```diff
@@ -5,11 +5,32 @@ import Page from '../features/layout/Page';
...
+        if (sortValueA === null) return 1;
+        if (sortValueB === null) return -1;
+        if (sortValueA === null && sortValueB == null) return 0;
+        return (
+          sortValueA.localeCompare(sortValueB, 'en', {
+            numeric: true,
+          }) * (sortOrder === 'asc' ? 1 : -1)
+        );
+      });
+      if (sortOrder !== null) {
```
wildum, line 25: maybe you could have the check also at the beginning. so it woud then be if (!sortfield || !sortOrder) return;

wildum (changes requested):
Thanks once again for your contribution! It's nice to see improvements for the UI and the little video that you added in description is very helpful. My knowledge of react is very limited but I left some suggestions

hainenber:
hi there, all of your proposals are implemented, I wholeheartedly agreed with all of them. The added changes don't change the sorting logic and visuals, so the video still stays valid. PTAL if you have time again. Thank you!

web/ui/src/features/component/Table.module.css
```diff
@@ -32,3 +32,25 @@
...
+}
+
+.table tr th[data-sort-order="asc"] {
+  cursor: pointer;
+  box-shadow: none !important;
+    &:before {
+      content: "▲";
+      float: right;
+    } 
+}
```
wildum, line 56: is "box-shadow: none !important;" needed here?
can we simplify the CSS with:

.table tr th {
  cursor: pointer;
}

.table tr th[data-sort-order="desc"]:before {
  content: "▼";
  float: right;
}

.table tr th[data-sort-order="asc"]:before {
  content: "▲";
  float: right;
}

web/ui/src/features/component/TableHead.tsx
```diff
@@ -0,0 +1,41 @@
...
+  headers: string[];
+  handleSorting?: (header: string, sortOrder: SortOrder) => void;
+}
+
+const TableHead = ({ headers, handleSorting }: Props) => {
+  const [sortField, setSortField] = useState('');
+  const [order, setOrder] = useState(SortOrder.ASC);
+
+  const handleSortingChange = (header: string) => {
+    const sortOrder = header === sortField && order === SortOrder.ASC ? SortOrder.DESC : SortOrder.ASC;
```
wildum, line 15: I found the logic a little bit hard to understand. Maybe we could refactor the code a bit to make the two following scenarios more obvious:
- the user clicks on a new header
- the user clicks on the same header another time

I would suggest this, although it is a bit verbose it clearly handles both scenarios:
```
if (header === sortField) {
  // If the user clicked again on the same header we toggle the sort order 
  sortOrder = order === SortOrder.ASC ? SortOrder.DESC : SortOrder.ASC;
} else {
  // If the user clicks on a new header we sort the entries by default in ASC order
  sortOrder = SortOrder.ASC;
}
```

web/ui/src/features/component/TableHead.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  return (
+    <tr>
+      {headers.map((header) => {
+        const sortOrderHeaderAttribute = sortField === header ? order : 'default';
+        return (
+          <th
+            key={header}
+            onClick={() => handleSortingChange(header)}
+            data-sort-order={handleSorting ? sortOrderHeaderAttribute : undefined}
```
wildum, line 31: can we change it to:
`data-sort-order={sortField === header ? order : undefined}`
and get rid of `const sortOrderHeaderAttribute = sortField === header ? order : 'default';` ?

web/ui/src/features/component/Table.module.css
wildum: if you remove the 'default' as stated in my other comment then I believe that this should be ok

web/ui/src/pages/PageComponentList.tsx
```diff
@@ -1,15 +1,33 @@
...
+      if (sortValueA === null) return 1;
+      if (sortValueB === null) return -1;
+      return (
+        sortValueA.localeCompare(sortValueB, 'en', {
+          numeric: true,
+        }) * (sortOrder === SortOrder.ASC ? 1 : -1)
+      );
+    });
+    setComponents(sorted);
+  };
```
wildum, line 26: To make the sorting easier to extend you could use a map:

```
const fieldMappings: { [key: string]: (comp: any) => string | undefined } = {
    Health: comp => comp.health?.state?.toString(),
    localID: comp => comp.localID,
   // Add new fields if needed here.
};

function getSortValue(component: any, field: string): string | undefined {
    const valueGetter = fieldMappings[field];
    return valueGetter ? valueGetter(component) : undefined;
}
```

And you could use it as follow in the sorting function:
(sortValueA and sortValueB can now be 'undefined', that's why I changed the if statement as well)
```
    const sortValueA = getSortValue(a, sortField);
    const sortValueB = getSortValue(b, sortField);

    if (!sortValueA) return 1;
    if (!sortValueB) return -1;
```

wildum (changes requested):
Thanks for adapting the code, I did a second round of review with a few more suggestions

wildum:
One more thing, after discussing with our new UX designer, we would like to have by default the sorting enabled with ID "asc". This way the arrow will already be shown and it's easier for the user to find the sorting functionality

web/ui/src/features/component/Table.module.css
hainenber: CSS block for `default` table header removed. I believe this can be resolved now.

web/ui/src/features/component/TableHead.tsx
hainenber: Done this one too in latest commit.

web/ui/src/pages/PageComponentList.tsx
hainenber: Done this one too in latest commit.

web/ui/src/features/component/TableHead.tsx
hainenber: Done this one too in latest commit.

hainenber:
Hey, thanks for another review. I've made the refactor to simplify the sorting logic here and there as well as making ID column to be sorted ascendingly by default. PTAL again

CHANGELOG.md
```diff
@@ -113,13 +113,21 @@ Main (unreleased)
 
 - Added an `exclude_event_message` option to `loki.source.windowsevent` in flow mode,
   which excludes the human-friendly event message from Windows event logs. (@ptodev)
+
+- Make component list sortable in web UI. (@hainenber)
+
+- Added an `exclude_event_message` option to `loki.source.windowsevent` in flow mode,
+  which excludes the human-friendly event message from Windows event logs. (@ptodev)
+
```
wildum, line 121: duplicated entries to remove

web/ui/src/features/component/TableHead.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+const TableHead = ({ headers, handleSorting }: Props) => {
+  const [sortField, setSortField] = useState('ID');
+  const [order, setOrder] = useState(SortOrder.ASC);
+
+  const handleSortingChange = (header: string) => {
+    // User clicks on the new header, use default ASC sort order
+    let sortOrder = SortOrder.ASC;
+
+    // User clicks again on the header we toggled the sort order previously
```
wildum, line 18:
```suggestion
    // User clicks again on the header, we toggle the previous sort order
```

wildum (changes requested):
Thanks for making the changes. I added two little comments but it looks good, nice job :)

web/ui/src/pages/PageComponentList.tsx
```diff
@@ -1,15 +1,44 @@
...
 import Page from '../features/layout/Page';
 import { useComponentInfo } from '../hooks/componentInfo';
 
+const fieldMappings: { [key: string]: (comp: any) => string | undefined } = {
+  Health: (comp) => comp.health?.state?.toString(),
+  ID: (comp) => comp.localID,
+  // Add new fields if needed here.
+};
+
+function getSortValue(component: any, field: string): string | undefined {
```
wildum, line 14: nit: we actually should replace any (+ the one ligne 8) with Component. This was my mistake in the code I suggested and was caught by the linter: " Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any"

hainenber: Fixed in latest commit

CHANGELOG.md
hainenber: Duplicate entry removed. Thanks for pointing this one out!

CHANGELOG.md
```diff
@@ -115,13 +115,18 @@ Main (unreleased)
 
 - Added an `exclude_event_message` option to `loki.source.windowsevent` in flow mode,
   which excludes the human-friendly event message from Windows event logs. (@ptodev)
+
+- Make component list sortable in web UI. (@hainenber)
```
wildum, line 119: this entry is still duplicated

hainenber: Removed in latest commit

wildum (approved):
Good work, it looks good to me, you will still need to update with main because the changelog has conflicts. I'm not an official maintainer yet so I can't merge it myself. I will ping my colleagues to take over from here

PR #5643, opened 2023-10-30
Remove integration-tests from windows unit tests in drone

wildum:
Looking at recent builds on main I noticed that the windows unit tests were trying to run the integration tests (https://drone.grafana.net/grafana/agent/14196/2/2).
Integration tests are not meant to run in drone for now but in github actions. This PR should fix the problem.

### DIFF
--- a/.drone/drone.yml
+++ b/.drone/drone.yml
@@ -181,7 +181,7 @@ platform:
   version: "1809"
 steps:
 - commands:
-  - go test -tags="nodocker,nonetwork" ./...
+  - go test -tags="nodocker,nonetwork" $(go list ./... | grep -v integration-tests)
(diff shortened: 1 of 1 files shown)


.drone/drone.yml
```diff
@@ -181,7 +181,7 @@ platform:
   version: "1809"
 steps:
 - commands:
-  - go test -tags="nodocker,nonetwork" ./...
+  - go test -tags="nodocker,nonetwork" $(go list ./... | grep -v integration-tests)
```
tpaschalis, line 184: Would it be better if we had a Go build tag on the integration tests instead?

tpaschalis (approved):

PR #5649, opened 2023-10-30
feat: add controller annotations

mcanevet:
#### PR Description

This PR allows setting annotations to the controller.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added


operations/helm/charts/grafana-agent/values.yaml
```diff
@@ -152,6 +152,9 @@ controller:
   # -- Number of pods to deploy. Ignored when controller.type is 'daemonset'.
   replicas: 1
 
+  # -- Annotations to add to controller.
+  annotations: {}
```
rfratto, line 156: For future flexibility, should this be `extraAnnotations`? (For example, if we eventually add a default set of annotations, this should likely append instead of overwrite)

mcanevet: every time I've seen this parameter exposed in a helm chart it was just named "annotations".

mcanevet: For example: https://github.com/grafana/helm-charts/blob/main/charts/grafana/values.yaml#L190-L191

rfratto: Looking through this chart, we do currently use the `extra*` terminology a few times in our chart:

* `agent.extraEnv`
* `agent.extraArgs`
* `agent.extraPorts`
* `agent.mounts.extra`
* `controller.volumes.extra` 
* `ingress.extraPaths`

There are a few cases where, like you mentioned, it's not used:

* `controller.podAnnotations`
* `controller.podLabels` 

I'm wondering if it makes sense to be consistent here, especially since the current chart has an implication of being additive whenever "extra" is mentioned.

mcanevet: I don't mind to change `annotations` to `extraAnnotations`. @rfratto could you please discuss that with other maintainers of the Helm Chart so that I can make a change if needed? Or maybe you are the only one to decide, then please tell me what is your preference.

captncraig: In this case, we are not adding any annotations of our own, so I don't know that `extra` is needed. But one common pitfal is that people put annotations on the controller, expecting them to show up on the pod as well. Whatever we end up with should help people with that distinction. I am ok with `annotations`, with a reasonable comment.

rfratto: > In this case, we are not adding any annotations of our own, so I don't know that extra is needed

It's not in this particular case, but my original comment is concerned about future extension, to allow us flexibility in the future to start adding pod annotations if we need to.

tpaschalis: I like the `extra*` convention for things we extend/append to, and make sure we don't use it for things we replace instead.

mcanevet: @rfratto @captncraig @tpaschalis PR amended

tpaschalis:
Hey there 👋 Do yo mind taking a look at the failed CI job? Looks like you'll have to run `make generate-helm-docs` to fix the documentation

rfratto:
There seems to be a linting error unrelated to your PR that we haven't caught yet, since this is the first Helm PR since dependabot updated our dependencies for the lint job. 

I'll push a fix for that error so we can get this merged.

rfratto (approved):
Thanks for the contribution!

PR #5663, opened 2023-10-31
feat: Add prometheus targets dashboard

### LINKED ISSUE #5662: Grafana Agent dashboards do not show Prometheus Discovered targets
### Request

When running Grafana Agent in clustering mode (and clustering is enabled in Podmonitors / ServiceMonitor components) then there isn't a single overview of the discovered targets any more making it difficult to find out if targets have scrape errors.

Some metrics are made available by Grafana Agent that can be used to setup a dashboard. Some could be reused from the static mode dashboards. 

### Use case

This feature is required to monitor state of discovered targets.

Paul424:
#### PR Description

Adds a prometheus target dashboard to the agent flow mixin; helpful to "see" if targets are failing and on which instance they run when in clustering mode.

#### PR Checklist

- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,8 @@ Main (unreleased)
  
 - Add queueing logs remote write client for `loki.write` when WAL is enabled. (@thepalbi)
 
+- Add a Prometheus Target dashboard to agent-flow-mixin (@Paul424)
+
 ### Bugfixes
 
 - Fixed an issue where `loki.process` validation for stage `metric.counter` was
--- a/operations/agent-flow-mixin/dashboards.libsonnet
+++ b/operations/agent-flow-mixin/dashboards.libsonnet
@@ -2,6 +2,7 @@
   grafanaDashboards+:
     (import './dashboards/controller.libsonnet') +
     (import './dashboards/resources.libsonnet') +
+    (import './dashboards/prometheus-target.libsonnet') +
     (import './dashboards/prometheus.remote_write.libsonnet') +
     (import './dashboards/cluster-node.libsonnet') +
     (import './dashboards/cluster-overview.libsonnet'),
--- a/operations/agent-flow-mixin/dashboards/prometheus-target.libsonnet
+++ b/operations/agent-flow-mixin/dashboards/prometheus-target.libsonnet
@@ -6,11 +6,7 @@ local filename = 'agent-flow-prometheus-target.json';
   [filename]:
     dashboard.new(name='Grafana Agent Flow / Prometheus Target') +
     dashboard.withDocsLink(
-      url='https://grafana.com/docs/agent/latest/flow/reference/components/prometheus.operator.servicemonitors/',
-      desc='Component documentation',
-    ) +
-    dashboard.withDocsLink(
-      url='https://grafana.com/docs/agent/latest/flow/reference/components/prometheus.operator.podmonitors/',
+      url='https://grafana.com/docs/agent/latest/flow/reference/components/prometheus.scrape/',
       desc='Component documentation',
     ) +
     dashboard.withDashboardsLink() +
@@ -19,17 +15,17 @@ local filename = 'agent-flow-prometheus-target.json';
       dashboard.newTemplateVariable('cluster', |||
         label_values(agent_component_controller_running_components, cluster)
       |||),
-      dashboard.newTemplateVariable('namespace', |||
+      dashboard.newMultiTemplateVariable('namespace', |||
         label_values(agent_component_controller_running_components{cluster="$cluster"}, namespace)
       |||),
       dashboard.newMultiTemplateVariable('job', |||
-        label_values(agent_component_controller_running_components{cluster="$cluster", namespace="$namespace"}, job)
+        label_values(agent_component_controller_running_components{cluster="$cluster", namespace=~"$namespace"}, job)
       |||),
       dashboard.newMultiTemplateVariable('instance', |||
-        label_values(agent_component_controller_running_components{cluster="$cluster", namespace="$namespace"}, instance)
+        label_values(agent_component_controller_running_components{cluster="$cluster", namespace=~"$namespace"}, instance)
       |||),
       dashboard.newMultiTemplateVariable('scrape_job', |||
-        label_values(prometheus_target_scrape_pool_targets{namespace="$namespace"}, scrape_job)
+        label_values(prometheus_target_scrape_pool_targets{namespace=~"$namespace"}, scrape_job)
       |||),
     ]) +
     dashboard.withPanelsMixin([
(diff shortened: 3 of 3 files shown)


rfratto:
Hi, thanks for the contribution! 

I think you've identified some useful things to track here, but I'm worried that the dashboard is too general to be useful for every user of Grafana Agent. In particular, it requires three things:

* Using `prometheus.scrape` or `prometheus.operator.*`
* Using `prometheus.remote_write` 
* Collecting conntrack stats 

Any subset of these will be true for a given user, so I'm nervous about building a dashboard that uses all three, where some users might see empty panels and be confused. 

My inclination is to do some mix of the following: 

* Generally make the dashboard more narrowly focused on the Prometheus scraper; removing the dependency on conntrack (potentially moving that out to a new dashboard) and prometheus.remote_write. 
* Find another general way of collecting the metrics to reduce dependencies; for example, can we add new metrics to track outgoing samples so that the dashboard doesn't need to rely on `prometheus.remote_write`

Paul424:
> Hi, thanks for the contribution!
> 
> I think you've identified some useful things to track here, but I'm worried that the dashboard is too general to be useful for every user of Grafana Agent. In particular, it requires three things:
> 
> * Using `prometheus.scrape` or `prometheus.operator.*`
> * Using `prometheus.remote_write`
> * Collecting conntrack stats
> 
> Any subset of these will be true for a given user, so I'm nervous about building a dashboard that uses all three, where some users might see empty panels and be confused.
> 
> My inclination is to do some mix of the following:
> 
> * Generally make the dashboard more narrowly focused on the Prometheus scraper; removing the dependency on conntrack (potentially moving that out to a new dashboard) and prometheus.remote_write.
> * Find another general way of collecting the metrics to reduce dependencies; for example, can we add new metrics to track outgoing samples so that the dashboard doesn't need to rely on `prometheus.remote_write`

@rfratto thanks for reviewing and your feedback, i'm fairly new to GrafanaAgent but eager to learn and improve.

1. I'm not sure about the dependency with remote-write; i'm guessing it's one of the metrics i have been using? But which one?
2. The conntrack metrics was the only metric i could find (with my config) that actually shows failing scrapes. The others (prometheus_target_scrapes_*) remain silent on network connection issue's like firewall/nwpols or TLS issue's.
3. I think given the title of the new dashboard it will be clear to users what the panels are addressing; it's only relevant to configs that have a prometheus scrape configured, i could further elaborate in the dashboard description maybe?

Paul

rfratto:
> I'm not sure about the dependency with remote-write; i'm guessing it's one of the metrics i have been using? But which one?

It was `agent_wal_samples_appended_total`, that's only ever available when `prometheus.remote_write` is used. (There's a `component_id` label on all of these metrics to let you know which component they come from)

> The conntrack metrics was the only metric i could find (with my config) that actually shows failing scrapes. The others (prometheus_target_scrapes_*) remain silent on network connection issue's like firewall/nwpols or TLS issue's.

Right, I think it's valuable, but conntrack isn't available for macOS or Windows users, so it might need its own dedicated dashboard. 

> I think given the title of the new dashboard it will be clear to users what the panels are addressing; it's only relevant to configs that have a prometheus scrape configured, i could further elaborate in the dashboard description maybe?

I did some brainstorming offline, and my gut feeling at the moment is that we have three types of overall dashboards:

* General dashboards not specific to components (such as the clustering dashboards)
* Component-specific dashboards (like the `prometheus.remote_write` dashboard. 
* Namespace-based dashboard (like a dashboard for all `prometheus.operator.*` components) 

I think we should avoid dashboards that cover specific pipelines that cross through multiple components, as pipelines are so flexible that I'm not sure if we can ever have one that's useful for the vast majority of users.

I think this dashboard falls somewhere in between component-specific (since it's primarily about scraping) and namespace-based (since it also covers multiple `prometheus.operator.*` components). 

Currently, I'm of the opinion that we should split this up into two dashboards, even if they're copies of each other for a moment:

1. A component-specific dashboard just for `prometheus.scrape`. 
2. A namespace-based dashboard for `prometheus.operator.*` components. 

The important bit here is that they're operating on components, so you shouldn't see metrics for a non-prometheus.scrape component in the prometheus.scrape dashboard, and you shouldn't see metrics for a non-prometheus.operator component in the prometheus.operator dashboard. 

I think conntrack, given that it's not metrics emitted by the agent and that not all agent users will be ale to collect those metrics, should be out of scope for now.

Paul424:
Hello Robert, thanks for the feedback, I already see I agree on most items but give me some time to think about this...

Paul424:
> > I'm not sure about the dependency with remote-write; i'm guessing it's one of the metrics i have been using? But which one?
> 
> It was `agent_wal_samples_appended_total`, that's only ever available when `prometheus.remote_write` is used. (There's a `component_id` label on all of these metrics to let you know which component they come from)
> 
> > The conntrack metrics was the only metric i could find (with my config) that actually shows failing scrapes. The others (prometheus_target_scrapes_*) remain silent on network connection issue's like firewall/nwpols or TLS issue's.
> 
> Right, I think it's valuable, but conntrack isn't available for macOS or Windows users, so it might need its own dedicated dashboard.
> 
> > I think given the title of the new dashboard it will be clear to users what the panels are addressing; it's only relevant to configs that have a prometheus scrape configured, i could further elaborate in the dashboard description maybe?
> 
> I did some brainstorming offline, and my gut feeling at the moment is that we have three types of overall dashboards:
> 
> * General dashboards not specific to components (such as the clustering dashboards)
> * Component-specific dashboards (like the `prometheus.remote_write` dashboard.
> * Namespace-based dashboard (like a dashboard for all `prometheus.operator.*` components)
> 
> I think we should avoid dashboards that cover specific pipelines that cross through multiple components, as pipelines are so flexible that I'm not sure if we can ever have one that's useful for the vast majority of users.
> 
> I think this dashboard falls somewhere in between component-specific (since it's primarily about scraping) and namespace-based (since it also covers multiple `prometheus.operator.*` components).
> 
> Currently, I'm of the opinion that we should split this up into two dashboards, even if they're copies of each other for a moment:
> 
> 1. A component-specific dashboard just for `prometheus.scrape`.
> 2. A namespace-based dashboard for `prometheus.operator.*` components.
> 
> The important bit here is that they're operating on components, so you shouldn't see metrics for a non-prometheus.scrape component in the prometheus.scrape dashboard, and you shouldn't see metrics for a non-prometheus.operator component in the prometheus.operator dashboard.
> 
> I think conntrack, given that it's not metrics emitted by the agent and that not all agent users will be ale to collect those metrics, should be out of scope for now.

Hello @rfratto,

I recognize the issue's you describe and also i agree on the "scoped" dashboards to add.

For now/short term i will take the dashboard additions to our own repo and set them next to the agent dashboards, this way we can iterate first to see if we can provide useful monitoring, some observations so far:
- conntrack is crucial to find failing scrapes at network level; setting these in a separate dashboard could work but it's always going to be joined with scrape metrics; otherwise you'll see any/all kinds of conntrack usage.
- some prometheus metrics are missing (at least like: prometheus_tsdb_head_samples_appended_total) this prevents us to reuse existing panels from kube-prom-stack (hence use of remote write metrics).

Paul

rfratto:
I'm going to close this PR for now since you're planning on keeping your dashboard changes downstream as you iterate on them.

Thanks for the work here though! If you want to eventually upstream your contributions reach out to us in another issue or PR and we can revisit the discussion then.

PR #5664, opened 2023-10-31
chore: fix build agent under Windows

jkroepke:
#### PR Description

I'm trying to build agent under Windows in powershell.exe

Windows does not support shebang. Putting bash in front archive the same (running bash from path).

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


Makefile
```diff
@@ -135,7 +135,7 @@ PROPAGATE_VARS := \
 
 GO_ENV := GOOS=$(GOOS) GOARCH=$(GOARCH) GOARM=$(GOARM) CGO_ENABLED=$(CGO_ENABLED)
 
-VERSION      ?= $(shell ./tools/image-tag)
+VERSION      := $(shell bash ./tools/image-tag)
```
rfratto, line 138: `?=` is significant here; we don't want to override VERSION if it's set via the environment; this is used in our CI scripts. 

```suggestion
VERSION      ?= $(shell bash ./tools/image-tag)
```

packaging/grafana-agent-flow/windows/install_script.nsis
```diff
@@ -10,7 +10,7 @@ Unicode true
 
 !include nsDialogs.nsh
 !include FileFunc.nsh
-!include ./macros.nsis
```
jkroepke, line 13: I got the error on my Windows machines without using docker

```
Changing directory to: "C:\Users\jok\GolandProjects\grafana-agent\packaging\grafana-agent-flow\windows"

Processing script file: "./packaging/grafana-agent-flow/windows/install_script.nsis" (ACP)
Unicode: true
Processing default plugins: "C:\Program Files (x86)\NSIS\Plugins\x86-unicode\*.dll"

!include: "C:\Program Files (x86)\NSIS\Include\nsDialogs.nsh" (UTF8)
!include: closed: "C:\Program Files (x86)\NSIS\Include\nsDialogs.nsh"
!include: "C:\Program Files (x86)\NSIS\Include\FileFunc.nsh" (UTF8)
!include: closed: "C:\Program Files (x86)\NSIS\Include\FileFunc.nsh"
!include: could not find: "./macros.nsis"
Error in script "./packaging/grafana-agent-flow/windows/install_script.nsis" on line 13 -- aborting creation process
make: *** [dist-agent-flow-installer] Fehler 1

```

tools/make/packaging.mk
```diff
@@ -6,7 +6,7 @@ PARENT_MAKEFILE := $(firstword $(MAKEFILE_LIST))
 dist: dist-agent-binaries dist-agent-flow-binaries dist-agentctl-binaries dist-agent-packages dist-agent-flow-packages dist-agent-installer dist-agent-flow-installer
 
 clean-dist:
-	rm -rf dist dist.temp
```
jkroepke, line 9: That fixes

```
PS C:\Users\jok\GolandProjects\grafana-agent> make clean
process_begin: CreateProcess(NULL, which docker, ...) failed.
rm -rf dist dist.temp
process_begin: CreateProcess(NULL, rm -rf dist dist.temp, ...) failed.
make (e=2): The system cannot find the file specified.
make: *** [clean-dist] Fehler 2

```

jkroepke (commented):
See comment changed below

jkroepke:
@rfratto Does the CI cover the execution of NSIS? 

I have some issues building the agent under Mac for Windows, including the NSIS installer

packaging/grafana-agent-flow/windows/install_script.nsis
```diff
@@ -49,8 +49,8 @@ Section "install"
 
   # Configure the out path and copy files to it.
   SetOutPath "$INSTDIR"
-  File "../../../dist.temp/grafana-agent-flow-windows-amd64.exe"
-  File "../../../dist.temp/grafana-agent-service-windows-amd64.exe"
+  File "..\..\..\dist.temp\grafana-agent-flow-windows-amd64.exe"
```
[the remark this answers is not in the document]
rfratto: > Need a revalidation under UNIX.

I tested this with `USE_CONTAINER=1 make dist-agent-flow-installer` and it still works under UNIX :)

tools/make/packaging.mk
```diff
@@ -374,6 +374,7 @@ ifeq ($(USE_CONTAINER),1)
 else
 	cp ./dist/grafana-agent-windows-amd64.exe ./packaging/grafana-agent/windows
 	cp LICENSE ./packaging/grafana-agent/windows
+	"mkdir" -p dist
```
rfratto, line 377: Does `mkdir` really have to be in quotes here? It's always a single word with no spaces so it shouldn't need it.

Ditto on line 386 below.

rfratto (commented):
Thank you! I only have one question about why we're using `"mkdir"` instead of `mkdir`, but otherwise this looks good!

rfratto:
> @rfratto Does the CI cover the execution of NSIS?

No, but I tested it locally; you're fine here.

tools/make/packaging.mk
jkroepke: Without quotes, there will be an failure under Windows.

It seems like the cmd.exe builtin mkdir command does not support the `-p` switch while the mkdir.exe (C:/Windows/system32/) support it.

The quotes forces to use the external exe instead the builtin command.

I found the solution here: https://superuser.com/questions/1253369/gnuwin32-makefile-mkdir-p

rfratto: That’s odd, and something we won’t remember :). Can you add a comment for each instance of this so we don’t accidentally remove the quotes in a future commit?

rfratto (approved):
Thanks!

PR #5678, opened 2023-11-01
Added Kubernetes service resolver to loadbalancing exporters.

ptodev:
#### PR Description

This feature was [added](https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/22776/files) to Collector version [v0.82.0](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.82.0), but it looks like we didn't update the components on our end to use it.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/component/otelcol/exporter/loadbalancing/loadbalancing.go
+++ b/component/otelcol/exporter/loadbalancing/loadbalancing.go
@@ -213,12 +213,12 @@ type KubernetesResolver struct {
 
 var _ river.Defaulter = &KubernetesResolver{}
 
-// DefaultDNSResolver holds default values for K8sSvcResolver.
-var DefaultK8sSvcResolver = KubernetesResolver{}
-
 // SetToDefault implements river.Defaulter.
 func (args *KubernetesResolver) SetToDefault() {
-	*args = DefaultK8sSvcResolver
+	if args == nil {
+		args = &KubernetesResolver{}
+	}
+	args.Ports = []int32{4317}
 }
 
 func (k8sSvcResolver *KubernetesResolver) Convert() loadbalancingexporter.K8sSvcResolver {
--- a/component/otelcol/exporter/loadbalancing/loadbalancing.go
+++ b/component/otelcol/exporter/loadbalancing/loadbalancing.go
@@ -224,7 +224,7 @@ func (args *KubernetesResolver) SetToDefault() {
 func (k8sSvcResolver *KubernetesResolver) Convert() loadbalancingexporter.K8sSvcResolver {
 	return loadbalancingexporter.K8sSvcResolver{
 		Service: k8sSvcResolver.Service,
-		Ports:   k8sSvcResolver.Ports,
+		Ports:   append([]int32{}, k8sSvcResolver.Ports...),
 	}
 }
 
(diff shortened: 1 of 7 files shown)


docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
```diff
@@ -136,6 +138,26 @@ Name | Type | Description | Default | Required
 `timeout`  | `duration` | Resolver timeout. | `"1s"`  | no
 `port`     | `string`   | Port to be used with the IP addresses resolved from the DNS hostname. | `"4317"` | no
 
+### k8s block
+
+The `k8s` block provides a way to load balance across the pods of a Kubernetes service. The Agent will be notified
```
clayton-cornell, line 143:
```suggestion
You can use the `k8s` block to load balance across the pods of a Kubernetes service. The Agent will be notified
```

clayton-cornell (commented):
One minor suggestion for the docs

component/otelcol/exporter/loadbalancing/loadbalancing.go
```diff
@@ -139,6 +139,7 @@ func (otlpConfig OtlpConfig) Convert() otlpexporter.Config {
 type ResolverSettings struct {
 	Static *StaticResolver `river:"static,block,optional"`
 	DNS    *DNSResolver    `river:"dns,block,optional"`
+	K8sSvc *K8sSvcResolver `river:"k8s,block,optional"`
```
rfratto, line 142: Why `k8s` and not `kubernetes`? I know `k8s` isn't that uncommmon but I think it's better to be explicit and spell it out here.

Even if upstream uses `k8s`, I think we should consider whether we want to explicitly follow that pattern.

clayton-cornell: I wondered that in the docs (the k8s block). I should have asked there too. We definitely should spell out all 3rd party product names even when shortcuts are in common use.

ptodev: I also think "kubernetes" is more clear than "k8s", and I was wondering about the same thing when I was implementing this. Ultimately I decided to keep using "k8s" for consistency with Collector:
* If we retain the same argument names, it is much easier for people to translate their configs form Collector to Flow and vice versa.
* If we start changing names of config arguments, we could theoretically run into issues if Collector adds new arguments with the same (or similar) name. It's unlikely that they will add a new argument called "kubernetes" here, but I think it's better to be on the safe side and avoid future confusion.

For those two reasons I am generally very hesitant to change argument names of components which are directly ported from the Collector.

rfratto: Personally, I believe we should feel comfortable taking our wrapped components in our own direction; we don't guarantee that our wrapped otelcol components are exactly the same (after all, we don't even use the same config language), and we've already drifted from upstream by using our own names and types for fields which indicate byte size (so users can provide a string like `10MiB`). 

I'd recommend asking the wider set of maintainers for their opinion so we know what do to in this specific case.

ptodev: Do you mind if I just change the attribute from "k8s" to "kubernetes"? I am personally happy to change it, I do agree that it makes it more clear, and I doubt that there will be an issue due to OTel adding extra parameters in the future. I really doubt anyone will disagree strongly about it, but I can ask if you insist.

rfratto (approved):
LGTM, but personally I still think the block should be called `kubernetes` instead of `k8s`.

ptodev:
> LGTM, but personally I still think the block should be called `kubernetes` instead of `k8s`.

I agree, and I've happily added it to the PR. Thank you for the review! 🚢

PR #5692, opened 2023-11-02
misc: remove references to milestones

rfratto:
As part of improving our work process, we are no longer using milestones for tracking when items get released.

### DIFF
--- a/docs/sources/_index.md.t
+++ b/docs/sources/_index.md.t
@@ -101,8 +101,6 @@ one minor release is moved.
 
 Patch and security releases may be created at any time.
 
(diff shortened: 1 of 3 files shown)


docs/sources/_index.md.t
```diff
@@ -101,8 +101,6 @@ one minor release is moved.
 
 Patch and security releases may be created at any time.
 
-[Milestones]: https://github.com/grafana/agent/milestones
```
rfratto, line 104: This was a stale link; it was never actually referenced anywhere so this is a no-op change to docs.

tpaschalis (approved):

PR #5699, opened 2023-11-02
Added last read/wrote timestamp metrics to loki.write

thepalbi:
#### PR Description

This PR adds two new metrics to the WAL enabled `loki.write`:
- last written timestamp
- last read timestamp

This will allow us to track the delay between the WAL writing and reading side, in actual time from the entries flowing through the WAL.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/component/common/loki/client/queue_client.go
+++ b/component/common/loki/client/queue_client.go
@@ -150,10 +150,11 @@ func (q *queue) closeNow() {
 // which allows it to be injected in the wal.Watcher as a destination where to write read series and entries. As the watcher
 // reads from the WAL, batches are created and dispatched onto a send queue when ready to be sent.
 type queueClient struct {
-	metrics *Metrics
-	logger  log.Logger
-	cfg     Config
-	client  *http.Client
+	metrics   *Metrics
+	qcMetrics *QueueClientMetrics
+	logger    log.Logger
+	cfg       Config
+	client    *http.Client
 
 	batches      map[string]*batch
 	batchesMtx   sync.Mutex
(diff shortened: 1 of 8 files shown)


component/common/loki/client/queue_client.go
```diff
@@ -283,16 +285,26 @@ func (c *queueClient) AppendEntries(entries wal.RefEntries, segment int) error {
...
 		c.markerHandler.UpdateReceivedData(segment, len(entries.Entries))
 	} else {
 		// TODO(thepalbi): Add metric here
 		level.Debug(c.logger).Log("msg", "series for entry not found")
 	}
+
+	// emit metric tracking max seen timestamp in WAL entry
+	if maxSeenTimestamp != -1 {
+		c.qcMetrics.lastReadTimestamp.WithLabelValues().Set(float64(maxSeenTimestamp))
+	}
```
tpaschalis, line 306: nit: won't this _always_ be the case when there's at least 1 entry?

pkg/util/metrics.go
```diff
@@ -0,0 +1,16 @@
+package util
+
+import "github.com/prometheus/client_golang/prometheus"
+
+// MustRegisterOrGet will attempt to register the supplied collector into the register. If it's already
+// registered, it will return that one.
+// In case that the register procedure fails with something other than already registered, this will panic.
+func MustRegisterOrGet(reg prometheus.Registerer, c prometheus.Collector) prometheus.Collector {
```
tpaschalis, line 8: Good call; we can reuse this in other places inside of the Loki components

tpaschalis (approved):
LGTM with a small nit

component/common/loki/client/queue_client.go
thepalbi: Yes! Upon an `AppendEntries` call there will always be at least one entry, if not it wouldn't have been decoded. Removing if.

thepalbi:
@tpaschalis all fixed

PR #5704, opened 2023-11-03
Fixing example in remote.kubernetes.secret.md to include nonsensitive()

nicolai-hornung-bl:
#### PR Description

The prometheus.remote_write > basic_auth > username is a string type and the agent will non start if the username is provided directly as a secret out of remote.kubernetes.secret as done in the example. nonsensitive() needs to be added to it for the config to be valid.

#### PR Checklist

- [x] Documentation added


clayton-cornell (approved):
Looks good to me. Thank you for your input here!

Issue #5411, opened 2023-10-09
azure_exporter doesn't support multiple dimensions


mtrin:
### What's wrong?

In a instance where we have to use a metric namespace with multiple dimension combinations, the only way to make it work seemed to be configuring one exporter per combination.
My use case is [applicationGateway](https://github.com/MicrosoftDocs/azure-reference-other/blob/main/azure-monitor-ref/supported-metrics/includes/Microsoft-Network-applicationgateways-metrics-include.md)

### Steps to reproduce

Attempt to configure everything in one exporter will fail with the described error.
It seems to be related to upstream version of the webdevops module and if I understand it right, just by using [this ](https://github.com/webdevops/azure-metrics-exporter/pull/53) version of the module `v0.0.0-20230721124911-34985800463f` would at least validate the dimensions by default and might resolve the problem. 
Is there anything that I'm missing?
I could not test this on the static version because I found it impossible to make it work accordingly to the docs using multiple instances.

### System information

Kubernetes

### Software version

v0.37.0-rc.0

### Configuration

```text
prometheus.exporter.azure "example" {
	subscriptions = ["<...>"]
	resource_type = "Microsoft.Network/applicationGateways"

	included_dimensions = ["BackendHttpSetting", "BackendPool", "BackendServer", "HttpStatusGroup"]
	metrics             = [
		"BackendResponseStatus",
		"Throughput",
	]
	timespan = "PT1M"
}

## workaround is to use two separate exporters for the above, then the error goes away. My case I'm having to use several exporters.
```

### Logs

```text
GET https://management.azure.com/subscriptions/<subscription>/resourceGroups/<rg>/providers/Microsoft.Network/applicationGateways/<agwname>/providers/Microsoft.Insights/metrics
--------------------------------------------------------------------------------
RESPONSE 400: 400 Bad Request
ERROR CODE: BadRequest
--------------------------------------------------------------------------------
{
  "error": {
    "additionalInfo": [
      {
        "type": "string",
        "info": "TraceId={<...>}"
      },
      {
        "type": "string",
        "info": "ExceptionType=Microsoft.Online.Metrics.MetricsMP.Utilities.RPRequestFormatException"
      }
    ],
    "code": "BadRequest",
    "message": "Metric: Throughput does not support requested dimension combination: backendhttpsetting,backendpool,backendserver,httpstatusgroup, supported ones are: , TraceId: {<...>}"
  }
}
```

marctc:
cc @thepalbi

[...]

PR #5707, opened 2023-11-03
azure exporter: Enable features from upstream exporter

kgeckhart:
#### PR Description
This PR adds two features from the upstream exporter which should help reduce the complexity when pull metrics from azure,

1. A code path to call a different azure API to gather metrics at the subscription level which can drastically reduce the likelihood of rate limiting
2. Disabling dimension validation reducing the number of exporter instances needed

I did as much as I could to add further context around these in the code/docs.

Fixes: #5411

Azure API rate limiting has come up internally but I couldn't find a corresponding issue to close. I can make one if needed

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


docs/sources/flow/reference/components/prometheus.exporter.azure.md
```diff
@@ -119,8 +137,11 @@ prometheus.exporter.azure "example" {
 		"SuccessServerLatency",
 		"Transactions",
 	]
+	included_dimensions = [
+        "ApiName",
+        "TransactionType",
```
thepalbi, line 142: nit: indent

docs/sources/flow/reference/components/prometheus.exporter.azure.md
```diff
@@ -77,6 +90,8 @@ Tags in `included_resource_tags` will be added as labels with the name `tag_<tag
 
 Valid values for `azure_cloud_environment` are `azurecloud`, `azurechinacloud`, `azuregovernmentcloud` and `azurepprivatecloud`.
 
+`validate_dimensions` is disabled by default to reduce the number of azure exporter instances requires when a `resource_type` has metrics with varying dimensions. When `validate_dimensions` is enabled you will need 1 exporter instance per metric + dimension combination which is more tedious to maintain.  
+
```
thepalbi, line 94: is this a breaking change?

kgeckhart: Maybe? This used to default to `true` so any existing configs won't break but the default behavior will change. I could flip the default to `true` to match the existing behavior or add it in the breaking changes section.

thepalbi: considering the feature, I think it's better to leave the false default and do a breaking change. wdyt @tpaschalis @mattdurham ?

mtrin: From what I understand, it would only mean existing configs could potentially start accepting more dimensions although they would be manually specified anyway

mattdurham: Whats the scope of the change between the two? Feels like we are getting more metrics, are we talking 2x,10x,100x? Consider me very ignorant of the internal workings of azure :)

mattdurham: For some additional context, if it makes sense to do a breaking change then I am onboard with it but it would need to be documented.

docs/sources/static/configuration/integrations/azure-exporter-config.md
```diff
@@ -11,9 +11,16 @@ description: Learn about azure_exporter_config
...
+collect metrics from [Azure Monitor](https://azure.microsoft.com/en-us/products/monitor). 
+
+The exporter offers two options for gathering metrics,
+1. (Default) Use an [Azure Resource Graph](https://azure.microsoft.com/en-us/get-started/azure-portal/resource-graph/#overview) query to identify resources for gathering metrics
+   1. This will make 1 API call per resource identified
+   1. Subscriptions with a reasonable amount of resources are liable to hit the [12000 requests per hour rate limit](https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/request-limits-and-throttling#subscription-and-tenant-limits) azure enforces 
+1. Set the regions to gather metrics from and get metrics for all resources across those regions
+   1. This will make an API call per subscription reducing the number of API calls dramatically 
+   1. This approach does not work with all resource types and Azure's does not document which resource types do/do not work
+   1. A resource type which is not support will produce errors which look like `Resource type: microsoft.containerservice/managedclusters not enabled for Cross Resource metrics`
```
mattdurham, line 23: Love the detail. What is the recourse if a user hits this error?

docs/sources/flow/reference/components/prometheus.exporter.azure.md
mtrin: @mattdurham I don't know much either about azure internals but I'm thinking the "break" is that wrong configs would start working. Say if someone has something like this and the agent is running, it would start working catching the default dimensions for each metric.
```
prometheus.exporter.azure "example" {
	subscriptions = ["<...>"]
	resource_type = "Microsoft.Network/applicationGateways"

	included_dimensions = []  
	metrics             = [
		"BackendResponseStatus",
		"Throughput",
	]
	timespan = "PT1M"
}
```

kgeckhart: This might be more of a behavioral change vs a breaking change. It won't change any currently working configuration someone will be running. They will still get the same amount of data. 

On main a config which requests this data from azure fails but runs without issue on this branch,
```yaml
            resource_type: ["microsoft.containerservice/managedclusters"]
            metrics:
              - node_cpu_usage_millicores
              - node_cpu_usage_percentage
              - node_disk_usage_bytes
              - node_disk_usage_percentage
              - node_memory_rss_bytes
              - node_memory_rss_percentage
              - node_memory_working_set_bytes
              - node_memory_working_set_percentage
              - node_network_in_bytes
              - node_network_out_bytes
            included_resource_tags:
              - environment
            included_dimensions:
              - node
              - nodepool
              - device
```
The `device` dimension is only valid for some of the requested metrics, https://learn.microsoft.com/en-us/azure/azure-monitor/reference/supported-metrics/microsoft-containerservice-managedclusters-metrics. The only valid way to do this on main was to split the config in to multiple definitions (do a self scrape in static mode or multiple exporter definitions in flow mode). 

Without validation though there's no safe guards for mistyping/misconfiguration so accidentally doing something like `devie` would not give you an error where it would previously.

docs/sources/static/configuration/integrations/azure-exporter-config.md
kgeckhart: That's a great question, I think the only option is to fall back to the resource graph API based approach which I can add in this list to be clear.

docs/sources/flow/reference/components/prometheus.exporter.azure.md
```diff
@@ -10,19 +10,28 @@ description: Learn about prometheus.exporter.azure
...
 the Egress metric for BlobService would be exported as `azure_microsoft_storage_storageaccounts_blobservices_egress_total_bytes`.
 
+The exporter offers two options for gathering metrics,
+1. (Default) Use an [Azure Resource Graph](https://azure.microsoft.com/en-us/get-started/azure-portal/resource-graph/#overview) query to identify resources for gathering metrics
+   1. This will make 1 API call per resource identified
+   1. Subscriptions with a reasonable amount of resources are liable to hit the [12000 requests per hour rate limit](https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/request-limits-and-throttling#subscription-and-tenant-limits) azure enforces 
+1. Set the regions to gather metrics from and get metrics for all resources across those regions
+   1. This will make an API call per subscription reducing the number of API calls dramatically 
+   1. This approach does not work with all resource types and Azure's does not document which resource types do/do not work
+   1. A resource type which is not support will produce errors which look like `Resource type: microsoft.containerservice/managedclusters not enabled for Cross Resource metrics`
```
clayton-cornell, line 26:
```suggestion
The exporter offers the following two options for gathering metrics.

1. (Default) Use an [Azure Resource Graph](https://azure.microsoft.com/en-us/get-started/azure-portal/resource-graph/#overview) query to identify resources for gathering metrics.
   1. This query will make one API call per resource identified.
   1. Subscriptions with a reasonable amount of resources can hit the [12000 requests per hour rate limit](https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/request-limits-and-throttling#subscription-and-tenant-limits) Azure enforces.
1. Set the regions to gather metrics from and get metrics for all resources across those regions.
   1. This option will make one API call per subscription, dramatically reducing the number of API calls.
   1. This approach does not work with all resource types, and Azure does not document which resource types do or do not work.
   1. A resource type that is not supported produces errors that look like `Resource type: microsoft.containerservice/managedclusters not enabled for Cross Resource metrics`.
```

docs/sources/static/configuration/integrations/azure-exporter-config.md
mattdurham: Awesome I think that would help!

docs/sources/flow/reference/components/prometheus.exporter.azure.md
mattdurham: Sounds reasonable then and pretty fine with this then

docs/sources/flow/reference/components/prometheus.exporter.azure.md
```diff
@@ -10,19 +10,28 @@ description: Learn about prometheus.exporter.azure
...
+
 ## Authentication
 
 Grafana agent must be running in an environment with access to Azure. The exporter uses the Azure SDK for go and supports [authentication](https://learn.microsoft.com/en-us/azure/developer/go/azure-sdk-authentication?tabs=bash#2-authenticate-with-azure).
 
 The account used by Grafana Agent needs:
 
-- [Read access to the resources that will be queried by Resource Graph](https://learn.microsoft.com/en-us/azure/governance/resource-graph/overview#permissions-in-azure-resource-graph)
+- When using an Azure Resoure Graph query, [read access to the resources that will be queried by Resource Graph](https://learn.microsoft.com/en-us/azure/governance/resource-graph/overview#permissions-in-azure-resource-graph)
```
clayton-cornell, line 34:
```suggestion
- [Read access to the resources that will be queried by Resource Graph](https://learn.microsoft.com/en-us/azure/governance/resource-graph/overview#permissions-in-azure-resource-graph) if you are using an Azure Resource Graph query.
```

docs/sources/flow/reference/components/prometheus.exporter.azure.md
```diff
@@ -50,23 +59,27 @@ prometheus.exporter.azure LABEL {
...
-| `metric_help_template`        | `string`       | Description of the metric.                                           | `"Azure metric {metric} for {type} with aggregation {aggregation} as {unit}"` | no       |
+| Name                          | Type           | Description                                                                                                                                                            | Default                                                                       | Required |
+|-------------------------------|----------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------|----------|
+| `subscriptions`               | `list(string)` | List of subscriptions to scrap metrics from.                                                                                                                           |                                                                               | yes      |
```
clayton-cornell, line 64:
```suggestion
| `subscriptions`               | `list(string)` | List of subscriptions to scrape metrics from.                                                                                                                           |                                                                               | yes      |
```

docs/sources/flow/reference/components/prometheus.exporter.azure.md
```diff
@@ -50,23 +59,27 @@ prometheus.exporter.azure LABEL {
...
+| `resource_type`               | `string`       | The Azure Resource Type to scrape metrics for.                                                                                                                         |                                                                               | yes      |
+| `metrics`                     | `list(string)` | The metrics to scrape from resources.                                                                                                                                  |                                                                               | yes      |
+| `resource_graph_query_filter` | `string`       | The [Kusto query][] filter to apply when searching for resources. Cannot be used if `regions` is set.                                                                  |                                                                               | no       |
```
clayton-cornell, line 67:
```suggestion
| `resource_graph_query_filter` | `string`       | The [Kusto query][] filter to apply when searching for resources. Can't be used if `regions` is set.                                                                  |                                                                               | no       |
```

docs/sources/flow/reference/components/prometheus.exporter.azure.md
```diff
@@ -50,23 +59,27 @@ prometheus.exporter.azure LABEL {
...
+| `regions`                     | `list(string)` | The list of regions for gathering metrics and enables gathering metrics for all resources in the subscription. Cannot be used if `resource_graph_query_filter` is set. |                                                                               | no       |
```
clayton-cornell, line 68:
```suggestion
| `regions`                     | `list(string)` | The list of regions for gathering metrics and enables gathering metrics for all resources in the subscription. Can't be used if `resource_graph_query_filter` is set. |                                                                               | no       |
```

docs/sources/static/configuration/integrations/azure-exporter-config.md
clayton-cornell, line 23:
```suggestion
The exporter offers the following two options for gathering metrics.

1. (Default) Use an [Azure Resource Graph](https://azure.microsoft.com/en-us/get-started/azure-portal/resource-graph/#overview) query to identify resources for gathering metrics.
   1. This query will make one API call per resource identified.
   1. Subscriptions with a reasonable amount of resources can hit the [12000 requests per hour rate limit](https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/request-limits-and-throttling#subscription-and-tenant-limits) Azure enforces.
1. Set the regions to gather metrics from and get metrics for all resources across those regions.
   1. This option will make one API call per subscription, dramatically reducing the number of API calls. 
   1. This approach does not work with all resource types, and Azure does not document which resource types do or do not work.
   1. A resource type that is not supported produces errors that look like `Resource type: microsoft.containerservice/managedclusters not enabled for Cross Resource metrics`.
```

docs/sources/static/configuration/integrations/azure-exporter-config.md
```diff
@@ -93,7 +100,14 @@ The account used by Grafana Agent needs:
 
   # Optional: The [kusto query](https://learn.microsoft.com/en-us/azure/data-explorer/kusto/query/) filter to apply when searching for resources
   # This value will be embedded in to a template query of the form `Resources | where type =~ "<resource_type>" <resource_graph_query_filter> | project id, tags`
+  # Cannot be used if `regions` is set.
   [resource_graph_query_filter: <string>]
+  
+  # Optional: The list of regions for gathering metrics. Enables gather metrics for all resources in the subscription.
+  # The list of available `regions` to your subscription can be found by running the azure CLI command `az account list-locations --query '[].name'`.
```
clayton-cornell, line 107:
```suggestion
  # The list of available `regions` to your subscription can be found by running the Azure CLI command `az account list-locations --query '[].name'`.
```

docs/sources/static/configuration/integrations/azure-exporter-config.md
```diff
@@ -93,7 +100,14 @@ The account used by Grafana Agent needs:
 
   # Optional: The [kusto query](https://learn.microsoft.com/en-us/azure/data-explorer/kusto/query/) filter to apply when searching for resources
   # This value will be embedded in to a template query of the form `Resources | where type =~ "<resource_type>" <resource_graph_query_filter> | project id, tags`
+  # Cannot be used if `regions` is set.
   [resource_graph_query_filter: <string>]
+  
+  # Optional: The list of regions for gathering metrics. Enables gather metrics for all resources in the subscription.
```
clayton-cornell, line 106:
```suggestion
  # Optional: The list of regions for gathering metrics. Enables gathering metrics for all resources in the subscription.
```

docs/sources/static/configuration/integrations/azure-exporter-config.md
```diff
@@ -93,7 +100,14 @@ The account used by Grafana Agent needs:
...
+  # The list of available `regions` to your subscription can be found by running the azure CLI command `az account list-locations --query '[].name'`.
+  # Cannot be used if `resource_graph_query_filter` is set.
```
clayton-cornell, line 108:
```suggestion
  # Can't be used if `resource_graph_query_filter` is set.
```

docs/sources/static/configuration/integrations/azure-exporter-config.md
```diff
@@ -93,7 +100,14 @@ The account used by Grafana Agent needs:
 
   # Optional: The [kusto query](https://learn.microsoft.com/en-us/azure/data-explorer/kusto/query/) filter to apply when searching for resources
   # This value will be embedded in to a template query of the form `Resources | where type =~ "<resource_type>" <resource_graph_query_filter> | project id, tags`
+  # Cannot be used if `regions` is set.
```
clayton-cornell, line 103:
```suggestion
  # Can't be used if `regions` is set.
```

docs/sources/static/configuration/integrations/azure-exporter-config.md
```diff
@@ -137,6 +151,10 @@ The account used by Grafana Agent needs:
 
   # Optional: Which azure cloud environment to connect to, azurecloud, azurechinacloud, azuregovernmentcloud, or azurepprivatecloud
   [azure_cloud_environment: <string> | default = "azurecloud"]
+  
+  # Optional: validation is disabled by default to reduce the number of azure exporter instances required when a `resource_type` has metrics with varying dimensions.
```
clayton-cornell, line 155:
```suggestion
  # Optional: Validation is disabled by default to reduce the number of Azure exporter instances required when a `resource_type` has metrics with varying dimensions. 
```

docs/sources/flow/reference/components/prometheus.exporter.azure.md
```diff
@@ -77,6 +90,8 @@ Tags in `included_resource_tags` will be added as labels with the name `tag_<tag
 
 Valid values for `azure_cloud_environment` are `azurecloud`, `azurechinacloud`, `azuregovernmentcloud` and `azurepprivatecloud`.
 
+`validate_dimensions` is disabled by default to reduce the number of azure exporter instances requires when a `resource_type` has metrics with varying dimensions. When `validate_dimensions` is enabled you will need 1 exporter instance per metric + dimension combination which is more tedious to maintain.
```
clayton-cornell, line 93:
```suggestion
`validate_dimensions` is disabled by default to reduce the number of Azure exporter instances requires when a `resource_type` has metrics with varying dimensions. When `validate_dimensions` is enabled you will need one exporter instance per metric + dimension combination which is more tedious to maintain.  
```

docs/sources/static/configuration/integrations/azure-exporter-config.md
```diff
@@ -137,6 +151,10 @@ The account used by Grafana Agent needs:
 
   # Optional: Which azure cloud environment to connect to, azurecloud, azurechinacloud, azuregovernmentcloud, or azurepprivatecloud
   [azure_cloud_environment: <string> | default = "azurecloud"]
+  
+  # Optional: validation is disabled by default to reduce the number of azure exporter instances required when a `resource_type` has metrics with varying dimensions. 
+  # Choosing to enable `validate_dimensions` will require 1 exporter instance per metric + dimension combination which can be very tedious to maintain.
```
clayton-cornell, line 156:
```suggestion
  # Choosing to enable `validate_dimensions` will require one exporter instance per metric + dimension combination which can be very tedious to maintain.
```

docs/sources/flow/reference/components/prometheus.exporter.azure.md
kgeckhart: Oddly, locally this is all spaced properly and renders as I would expect. I'm really confused about why it's off when pushed.

kgeckhart:
@mattdurham and @clayton-cornell I think I covered all the PR feedback if you wanted to take another pass

clayton-cornell:
@mattdurham This one fell off my ToDo list. There's nothing more I have to add for docs input. Over to you for a final look/approval?

kgeckhart:
I'll happily resolve conflicts after getting a 👍 on the changes. The `go.mod` + `CHANGELOG.md` are going to keep drifting really quickly.

captncraig (approved):

PR #5712, opened 2023-11-04
feat(module/git): allow module.git use cached content when failed to fetch

### LINKED ISSUE #5708: Agent fails to start when if can't pull the git repo even if the repos is present locally already
### What's wrong?

When the Grafana Agent is started using the [git module](https://grafana.com/docs/agent/latest/flow/reference/components/module.git/) the agent will clone the repo and store it locally. Then after the first startup was successful, if the agent is restarted and the git server is unavailable the agent fails to start even tough the repo is stored on disk and the module file would be accessible. 

### Steps to reproduce

1. Configure the agent to use a git a repository
2. Start the Agent (wait for it to fully start and check that the repository is present in the agent data folder)
3. Check that the repository and files are cached locally:
```
root@bc82d36f2915:/# ls -la /var/lib/grafana-agent-flow/module.git.test/repo/
total 4
drwxr-xr-x. 1 root root  40 Nov  3 16:22 .
drwxr-xr-x. 1 root root  70 Nov  3 16:22 ..
drwxr-xr-x. 1 root root  52 Nov  3 16:22 .git
-rw-r--r--. 1 root root 267 Nov  3 16:22 prometheus.river
```
5. Switch the repository to private to make it inaccessible 
6. Start the Agent and observe that it fails to start

### System information

Windows Server 2022 & Ubuntu 22.04

### Software version

Grafana Agent v0.37.1

### Configuration

```text
module.git "test" {
  repository = "https://github.com/ntimo/grafana-agent-git-module-test.git"
  path       = "prometheus.river"
}
```

### Logs

```text
Error: /etc/config.river:1:1: Failed to build component: building component: failed to update repository "https://github.com/ntimo/grafana-agent-git-module-test.git": authentication required

1 |   module.git "test" {
  |  _^^^^^^^^^^^^^^^^^^^
2 | |   repository = "https://github.com/ntimo/grafana-agent-git-module-test.git"
3 | |   path       = "prometheus.river"
4 | | }
  | |_^
5 |   

Error: could not perform the initial load successfully
```

hainenber:
#### PR Description
Allow `module.git` loader to use cached contents when failed to fetch during startup/update. This should be transient as the next poll interval, if fetch failure got resolved, the Git content will be up-to-date.

Fixes #5708

#### Notes to the Reviewer
Do let me know if something more format is needed, like a unit test or so.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Tests updated

### DIFF
--- a/component/module/git/git.go
+++ b/component/module/git/git.go
@@ -92,8 +92,16 @@ func New(o component.Options, args Arguments) (*Component, error) {
 		argsChanged: make(chan struct{}, 1),
 	}
 
+	// Only acknowledge the error from Update if it's not a
+	// vcs.UpdateFailedError; vcs.UpdateFailedError means that the Git repo
+	// exists but we were just unable to update it.
 	if err := c.Update(args); err != nil {
-		return nil, err
+		if errors.As(err, &vcs.UpdateFailedError{}) {
+			level.Error(c.log).Log("msg", "failed to update repository", "err", err)
+			c.updateHealth(err)
+		} else {
+			return nil, err
+		}
 	}
 	return c, nil
 }
@@ -195,11 +203,10 @@ func (c *Component) Update(args component.Arguments) (err error) {
 
 	// Create or update the repo field.
 	// Failure to update repository makes the module loader temporarily use cached contents on disk
-	var updateVcsErr vcs.UpdateFailedError
 	if c.repo == nil || !reflect.DeepEqual(repoOpts, c.repoOpts) {
 		r, err := vcs.NewGitRepo(context.Background(), repoPath, repoOpts)
 		if err != nil {
-			if errors.As(err, &updateVcsErr) {
+			if errors.As(err, &vcs.UpdateFailedError{}) {
 				level.Error(c.log).Log("msg", "failed to update repository", "err", err)
 				c.updateHealth(err)
 			} else {
@@ -228,14 +235,8 @@ func (c *Component) Update(args component.Arguments) (err error) {
 // controller. pollFile must only be called with c.mut held.
 func (c *Component) pollFile(ctx context.Context, args Arguments) error {
 	// Make sure our repo is up-to-date.
-	var updateVcsErr *vcs.UpdateFailedError
 	if err := c.repo.Update(ctx); err != nil {
-		if errors.As(err, &updateVcsErr) {
-			level.Error(c.log).Log("msg", "failed to update repository", "err", err)
-			c.updateHealth(err)
-		} else {
-			return err
-		}
+		return err
 	}
 
 	// Finally, configure our controller.
--- a/component/module/git/git.go
+++ b/component/module/git/git.go
@@ -198,11 +198,12 @@ func (c *Component) Update(args component.Arguments) (err error) {
 	var updateVcsErr vcs.UpdateFailedError
 	if c.repo == nil || !reflect.DeepEqual(repoOpts, c.repoOpts) {
 		r, err := vcs.NewGitRepo(context.Background(), repoPath, repoOpts)
-		if errors.As(err, &updateVcsErr) {
-			level.Error(c.log).Log("msg", "failed to update repository", "err", err)
-		}
 		if err != nil {
-			return err
+			if errors.As(err, &updateVcsErr) {
+				level.Error(c.log).Log("msg", "failed to update repository", "err", err)
+			} else {
+				return err
+			}
 		}
 		c.repo = r
 		c.repoOpts = repoOpts
--- a/component/module/git/git.go
+++ b/component/module/git/git.go
@@ -201,6 +201,7 @@ func (c *Component) Update(args component.Arguments) (err error) {
 		if err != nil {
 			if errors.As(err, &updateVcsErr) {
 				level.Error(c.log).Log("msg", "failed to update repository", "err", err)
+				c.updateHealth(err)
 			} else {
 				return err
 			}
@@ -231,6 +232,7 @@ func (c *Component) pollFile(ctx context.Context, args Arguments) error {
 	if err := c.repo.Update(ctx); err != nil {
 		if errors.As(err, &updateVcsErr) {
 			level.Error(c.log).Log("msg", "failed to update repository", "err", err)
+			c.updateHealth(err)
 		} else {
 			return err
 		}
--- a/component/module/git/git.go
+++ b/component/module/git/git.go
@@ -98,7 +98,6 @@ func New(o component.Options, args Arguments) (*Component, error) {
 	if err := c.Update(args); err != nil {
 		if errors.As(err, &vcs.UpdateFailedError{}) {
 			level.Error(c.log).Log("msg", "failed to update repository", "err", err)
-			c.updateHealth(err)
 		} else {
 			return nil, err
 		}
--- a/component/module/git/git.go
+++ b/component/module/git/git.go
@@ -3,6 +3,7 @@ package git
 
 import (
 	"context"
+	"errors"
 	"path/filepath"
 	"reflect"
 	"sync"
@@ -193,8 +194,13 @@ func (c *Component) Update(args component.Arguments) (err error) {
 	}
 
 	// Create or update the repo field.
+	// Failure to update repository makes the module loader temporarily use cached contents on disk
+	var updateVcsErr vcs.UpdateFailedError
 	if c.repo == nil || !reflect.DeepEqual(repoOpts, c.repoOpts) {
 		r, err := vcs.NewGitRepo(context.Background(), repoPath, repoOpts)
+		if errors.As(err, &updateVcsErr) {
+			level.Error(c.log).Log("msg", "failed to update repository", "err", err)
+		}
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


component/module/git/git.go
```diff
@@ -220,8 +227,13 @@ func (c *Component) Update(args component.Arguments) (err error) {
...
 func (c *Component) pollFile(ctx context.Context, args Arguments) error {
 	// Make sure our repo is up-to-date.
+	var updateVcsErr *vcs.UpdateFailedError
 	if err := c.repo.Update(ctx); err != nil {
-		return err
+		if errors.As(err, &updateVcsErr) {
+			level.Error(c.log).Log("msg", "failed to update repository", "err", err)
+		} else {
+			return err
+		}
```
rfratto, line 236: I'm worried this fix is too broad, and changes behavior we don't want to change: this (and the other change) is now hiding the error from the user even if the repo has never been cloned before.

Rather than hiding the error, I'd like to see the behavior changed to the following:

* Always return an error if `git clone` fails, since `git clone` is only used when the cache does not exist. 
* If a `git pull` fails while building the component via New, do not return an error via New, but still make sure the component's health gets marked as unhealthy. 
* In all other cases, if `git pull` fails, return an error. 

These changes will fix #5708 while never hiding the failure from the component health or the Flow controller.

rfratto (commented):
Thanks! I have a comment about the approach but I think you only need a few small tweaks to get this ready for merging.

component/module/git/git.go
hainenber: Correct me if I'm wrong but I assume my changes would only hide `git pull` error from being returned and for other errors, they'll be returned as normal. `git clone` error is always returned due to its criticality for the component and my changes assumedly doesn't change this behavior

You're right on hiding `git pull` error from Flow controller isn't ideal and I should mark it the component unhealthy whenever a failed `git pull` error occurs. Let me know if this resolves your concern 👀

rfratto: > Correct me if I'm wrong but I assume my changes would only hide `git pull` error from being returned and for other errors, they'll be returned as normal. 

You're right, sorry for the misunderstanding here; I didn't catch that there was two different error types depending on the operation that failed.

component/module/git/internal/vcs/git.go
```diff
@@ -58,15 +58,15 @@ func NewGitRepo(ctx context.Context, storagePath string, opts GitRepoOptions) (*
...
 		Force:      true,
 		Auth:       opts.Auth.Convert(),
 	})
-	if err != nil && !errors.Is(err, git.NoErrAlreadyUpToDate) {
-		return nil, UpdateFailedError{
+	if fetchRepoErr != nil && !errors.Is(fetchRepoErr, git.NoErrAlreadyUpToDate) {
+		err = UpdateFailedError{
 			Repository: opts.Repository,
-			Inner:      err,
+			Inner:      fetchRepoErr,
```
rfratto, line 69: Hm, I'd rather the error handler here return instead of falling back to Reset, that way we don't touch the repo at all in case a pull fails; something like

```go
if err != nil && !errors.Is(err, git.NoErrAlreadyUpToDate) {
  workTree, workTreeErr := repo.Worktree()
  if workTreeErr != nil {
    return nil, workTreeErr  
  }

  return &GitRepo{
    opts: opts, 
    repo: repo, 
    workTree: workTree, 
  }, UpdateFailedError{
    Repository: opts.Repository,
    Inner: err, 
  }
}
```

(This can probably be refactored a bit so the GitRepo object is constructed before doing a pull to reduce duplication with the successful path)

component/module/git/internal/vcs/git.go
```diff
@@ -103,15 +103,16 @@ func isRepoCloned(dir string) bool {
...
 		Force:      true,
 		Auth:       repo.opts.Auth.Convert(),
 	})
-	if err != nil && !errors.Is(err, git.NoErrAlreadyUpToDate) {
-		return UpdateFailedError{
+	if fetchRepoErr != nil && !errors.Is(fetchRepoErr, git.NoErrAlreadyUpToDate) {
+		err = UpdateFailedError{
 			Repository: repo.opts.Repository,
-			Inner:      err,
+			Inner:      fetchRepoErr,
```
rfratto, line 115: Ditto here: I'm not sure we should be falling back to resetting the git repo when the update fails, and I think we should just return directly here. 

(Also, this instance of `err` is always overwritten by the assignment on line 125 below)

component/module/git/git.go
```diff
@@ -193,10 +194,17 @@ func (c *Component) Update(args component.Arguments) (err error) {
...
 	if c.repo == nil || !reflect.DeepEqual(repoOpts, c.repoOpts) {
 		r, err := vcs.NewGitRepo(context.Background(), repoPath, repoOpts)
 		if err != nil {
-			return err
+			if errors.As(err, &updateVcsErr) {
+				level.Error(c.log).Log("msg", "failed to update repository", "err", err)
+				c.updateHealth(err)
+			} else {
+				return err
+			}
```
rfratto, line 207: Hm, I'm still not sure if the error check needs to happen here in the Update or in the call to pollFile. 

It seems like it would be sufficient if New made the check instead, since we're really only interested about fixing component construction to use a cache:

```go
// New creates a new module.git component.
func New(o component.Options, args Arguments) (*Component, error) {
	m, err := module.NewModuleComponent(o)
	if err != nil {
		return nil, err
	}
	c := &Component{
		opts: o,
		log:  o.Logger,

		mod: m,

		argsChanged: make(chan struct{}, 1),
	}

	// Only acknowledge the error from Update if it's not a
	// vcs.UpdateFailedError; vcs.UpdateFailedError means that the Git repo
	// exists but we were just unable to update it.
	if err := c.Update(args); err != nil && !errors.Is(err, &vcs.UpdateFailedError{}) {
		return nil, err 
	} 
	return c, nil
}
```

I think adding that check in New will allow you to leave Update and pollFile unchanged. WDYT?

component/module/git/internal/vcs/git.go
hainenber: Good point here. I've amended in latest commit.

component/module/git/internal/vcs/git.go
hainenber: Good point here. I've amended in latest commit.

component/module/git/git.go
hainenber: I was implementing this logic until finding out a crash due to nil pointer as `repo` attribute isn't populated due to removal of assignment at `Update` method

image

To proceed, I think we can remove changes at `pollFile` and keep the change at `Update`.

hainenber: There's no panic when there's only changes at `Update` method and the module's debug info is pretty clear

image

hainenber: My bad, `pollFile` also returns `vcs.UpdateFailedError` so the `New` method also has to have the changes as well. I've finalized in the latest commit

hainenber:
GH checks is failing due to [degraded performance](https://www.dockerstatus.com/) on DockerHub's side. Please help rerunning when it's better, Robert. Thanks!

component/module/git/git.go
```diff
@@ -91,8 +92,16 @@ func New(o component.Options, args Arguments) (*Component, error) {
...
 	}
 
+	// Only acknowledge the error from Update if it's not a
+	// vcs.UpdateFailedError; vcs.UpdateFailedError means that the Git repo
+	// exists but we were just unable to update it.
 	if err := c.Update(args); err != nil {
-		return nil, err
+		if errors.As(err, &vcs.UpdateFailedError{}) {
+			level.Error(c.log).Log("msg", "failed to update repository", "err", err)
+			c.updateHealth(err)
```
rfratto, line 101: This shouldn't be necessary since the call to `Update` will already update the health for you if it returns an error.

component/module/git/git.go
```diff
@@ -193,10 +202,16 @@ func (c *Component) Update(args component.Arguments) (err error) {
...
 	if c.repo == nil || !reflect.DeepEqual(repoOpts, c.repoOpts) {
 		r, err := vcs.NewGitRepo(context.Background(), repoPath, repoOpts)
 		if err != nil {
-			return err
+			if errors.As(err, &vcs.UpdateFailedError{}) {
+				level.Error(c.log).Log("msg", "failed to update repository", "err", err)
+				c.updateHealth(err)
+			} else {
+				return err
+			}
```
rfratto, line 214: I'm not sure I'm following whether this is necessary still; with the latest change, it's currently the callers responsibility for whether UpdateFailedError needs separate handling or not. 

Am I missing something that this check is changing?

rfratto (commented):
Sorry for the delay between reviews, there was a holiday yesterday. This is getting much closer!

rfratto:
I can't run the CI because of merge conflicts, can you resolve those?

component/module/git/git.go
hainenber: Oh, I wasn't aware of this internal detail :D

Removed `updateHealth` in latest commit

component/module/git/git.go
hainenber: The err check for `UpdateFailedError` in `New` method is to check for returned error from `pollFile` method.

This check right here is to inform Flow controller component's unhealthiness from failing to update Git repo and to continue assigning the Repo object and options to internal attributes. Returning early will leave the attributes, namely `r.repo` empty and can cause null pointer exception, I'd wager.

hainenber:
> I can't run the CI because of merge conflicts, can you resolve those?

Thanks for the heads-up, done resolving on my side!

PR #4948, opened 2023-08-25
Label the stability of otelcol flow components

ptodev:
#### PR Description

The stability level of otelcol components was not always present in the Agent docs. I updated them to match the stability indicated in the otel docs. However, a few things are not very clear:

* We [don't have](https://grafana.com/docs/agent/latest/stability/) a "deprecated" stability level. How should we label `otelcol.exporter.jaeger`?
* How should we label components with different stabilities for different signals (e.g. `otelcol.exporter.otlp`)?
* Would it be better to link to the Agent's stability page, simiarly to how Otel does it? E.g. in the [basic auth extension](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/basicauthextension#basic-authenticator), otel links to  [beta](https://github.com/open-telemetry/opentelemetry-collector#beta) stability description. Our links would go to [experimental](https://grafana.com/docs/agent/latest/stability/#experimental), [beta](https://grafana.com/docs/agent/latest/stability/#beta) and [stable](https://grafana.com/docs/agent/latest/stability/#stable). Personally, I think having the links is nice.

Links to otel docs:

- [basicauthextension](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.80.0/extension/basicauthextension)
- [bearertokenauthextension](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.80.0/extension/bearertokenauthextension)
- [headerssetterextension](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.80.0/extension/headerssetterextension)
- [oauth2clientauthextension](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.80.0/extension/oauth2clientauthextension)
- [sigv4authextension](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.80.0/extension/sigv4authextension)
- [spanmetricsconnector](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.80.0/connector/spanmetricsconnector)
- [jaegerexporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.80.0/exporter/jaegerexporter)
- [loggingexporter](https://github.com/open-telemetry/opentelemetry-collector/tree/v0.80.0/exporter/loggingexporter)
- [lokiexporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.80.0/exporter/lokiexporter)
[...]

PR #5715, opened 2023-11-06
flow: Add otelcol.receiver.vcenter component

marctc:
#### PR Description

This PR adds `otelcol.receiver.vcenter` which enables collection of metrics of vShere 
machines.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


docs/sources/flow/reference/components/otelcol.receiver.vcenter.md
```diff
@@ -0,0 +1,129 @@
...
+canonical: https://grafana.com/docs/agent/latest/flow/reference/components/otelcol.receiver.vcenter/
+title: otelcol.receiver.vcenter
+description: Learn about otelcol.receiver.vcenter
+---
+
+# otelcol.receiver.vcenter
+
+`otelcol.receiver.vcenter` accepts telemetry data from a 
+vCenter or ESXi host running VMware vSphere APIs and
+forwards it to other `otelcol.*` components.
```
ptodev, line 15: I think this paragraph needs a few more things:
* We need to make it clear that this component only works with metrics.
* Which metrics are produced? Seems it is [these](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.87.0/receiver/vcenterreceiver/documentation.md) ones? If we are not going to list them in our docs, it would be good to link to the OTel page instead. However, I'd prefer if we don't link to the OTel page because it contains yaml config for the collector which our users will find confusing. Could you try to list them here please?
* Do we need to list the [prerequisites](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.87.0/receiver/vcenterreceiver/README.md#prerequisites)? They seem useful.

docs/sources/flow/reference/components/otelcol.receiver.vcenter.md
```diff
@@ -0,0 +1,129 @@
...
+`endpoint` has the following format: `<protocol>://<hostname>` (e.g. `https://vcsa.hostname.localnet`)
+
+## Blocks
+
+The following blocks are supported inside the definition of
+`otelcol.receiver.vcenter`:
+
+Hierarchy | Block | Description | Required
+--------- | ----- | ----------- | --------
+tls | [tls][] | Configures TLS for the server. | no
```
ptodev, line 62: I think this component has a client and not a servier?
```suggestion
tls | [tls][] | Configures TLS for the HTTP client. | no
```

docs/sources/flow/reference/components/prometheus.exporter.vsphere.md
```diff
@@ -5,13 +5,23 @@ aliases:
...
 The `prometheus.exporter.vsphere` component embeds [`vmware_exporter`](https://github.com/grafana/vmware_exporter) to collect vSphere metrics
 
+> **BETA**: This is a [beta][] component. Beta components are subject to breaking
+> changes, and may be replaced with equivalent functionality that cover the
+> same use case. We recommend to use [otelcol.receiver.vcenter][] instead.
+
+[beta]: {{< relref "../../../stability.md#beta" >}}
+[otelcol.receiver.vcenter]: {{< relref "./otelcol.receiver.vcenter.md" >}}
+
+
```
ptodev, line 24:
```suggestion
```

docs/sources/flow/reference/components/prometheus.exporter.vsphere.md
```diff
@@ -5,13 +5,23 @@ aliases:
 - /docs/grafana-cloud/monitor-infrastructure/integrations/agent/flow/reference/components/prometheus.exporter.vsphere/
 canonical: https://grafana.com/docs/agent/latest/flow/reference/components/prometheus.exporter.vsphere/
 title: prometheus.exporter.vsphere
+labels:
+  stage: beta
```
ptodev, line 9:
```suggestion
```
We cannot do this :) Once we have said a component is stable, we cannot label it as non-stable. Check out [this comment](https://github.com/grafana/agent/pull/4948#pullrequestreview-1604960855) for a previous discussion.

docs/sources/flow/reference/components/otelcol.receiver.vcenter.md
```diff
@@ -0,0 +1,129 @@
...
+## Example
+
+This example forwards received telemetry data through a batch processor before
+finally sending it to an OTLP-capable endpoint:
+
+```river
+otelcol.receiver.vcenter "default" {
+  endpoint         = "http://localhost:15672"
+  username		   = "otelu"
+  password         = "password"
```
ptodev, line 108:
```suggestion
  endpoint = "http://localhost:15672"
  username = "otelu"
  password = "password"
```

docs/sources/flow/reference/components/otelcol.receiver.vcenter.md
```diff
@@ -0,0 +1,129 @@
...
+`endpoint` | `string` | Endpoint to the vCenter Server or ESXi host that has the sdk path enabled. | | yes
+`username` | `string` | Username to use for authentication. | | yes
+`password` | `string` | Password to use for authentication. | | yes
+`collection_interval` | `duration` | This receiver collects metrics on an interval. | `"1m"` | no
+`initial_delay` | `duration` | Defines how long this receiver waits before starting.. | `"1s"` | no
+`timeout` | `duration` | Defines the timeout for the underlying HTTP client. | `"0s"` | no
+
+`endpoint` has the following format: `<protocol>://<hostname>` (e.g. `https://vcsa.hostname.localnet`)
+
+## Blocks
```
ptodev, line 55: Are we not boing to have config for vcenter resource attributes?
```
// ResourceAttributesConfig provides config for vcenter resource attributes.
type ResourceAttributesConfig struct {
	VcenterClusterName               ResourceAttributeConfig `mapstructure:"vcenter.cluster.name"`
	VcenterDatastoreName             ResourceAttributeConfig `mapstructure:"vcenter.datastore.name"`
	VcenterHostName                  ResourceAttributeConfig `mapstructure:"vcenter.host.name"`
	VcenterResourcePoolInventoryPath ResourceAttributeConfig `mapstructure:"vcenter.resource_pool.inventory_path"`
	VcenterResourcePoolName          ResourceAttributeConfig `mapstructure:"vcenter.resource_pool.name"`
	VcenterVMID                      ResourceAttributeConfig `mapstructure:"vcenter.vm.id"`
	VcenterVMName                    ResourceAttributeConfig `mapstructure:"vcenter.vm.name"`
}
```

docs/sources/flow/reference/components/otelcol.receiver.vcenter.md
marctc: Do we need to include everything that it's in the component readme? Or linking to it would provide users some understanding what are they using?

component/otelcol/receiver/vcenter/vcenter.go
```diff
@@ -0,0 +1,104 @@
...
+	Output *otelcol.ConsumerArguments `river:"output,block"`
+}
+
+var _ receiver.Arguments = Arguments{}
+
+var (
+	// DefaultArguments holds default values for Arguments.
+	DefaultArguments = Arguments{
+		ScraperControllerArguments: otelcol.DefaultScraperControllerArguments,
+	}
```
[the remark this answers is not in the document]
marctc: I don't know if there's a better way to do this, but the defaults were not applied if didn't add this (the tests were failing)

docs/sources/flow/reference/components/otelcol.receiver.vcenter.md
ptodev: > Do we need to include everything that it's in the component readme?

No - we might include more information, or less. We need to assume that the users will not read the Collector docs and that our doc page is the first page they are seeing for this component. 

> Or linking to it would provide users some understanding what are they using?

In general we avoid linking to OTel docs because they tend to contain Collector-specific yaml config which confuses the users. It's best to include all the relevant info in our own docs, unless there is a strong reason not to.

component/otelcol/receiver/vcenter/vcenter.go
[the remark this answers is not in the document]
ptodev: Ok, I think it's ok to keep this code as it is here.

docs/sources/flow/reference/components/otelcol.receiver.vcenter.md
```diff
@@ -0,0 +1,133 @@
...
+
+> **NOTE**: `otelcol.receiver.vcenter` is a wrapper over the upstream
+> OpenTelemetry Collector `vcenter` receiver from the `otelcol-contrib`
+> distribution. Bug reports or feature requests will be redirected to the
+> upstream repository, if necessary.
+
+Multiple `otelcol.receiver.vcenter` components can be specified by giving them
+different labels.
+
+Make sure that you fulfill the [prerequisites](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/vcenterreceiver/README.md#prerequisites) before using this component.
```
ptodev, line 25: There are a couple of issues with linking to the OTEL docs like this:
* This link goes to the main branch. It'd ideally go to the version of OTEL which we use - see the otelcol.processor.transform for example URLs.
* There are various things listed in the page, including Collector yampl config examples and Collector feature gates. This could be confusing to the Agent users. 

I'd prefer if we just copy over the relevant information:

> This receiver has been built to support ESXi and vCenter versions:
> 
> * 7.5
> * 7.0
> * 6.7
> 
> A “Read Only” user assigned to a vSphere with permissions to the vCenter server, cluster and all subsequent resources being monitored must be specified in order for the receiver to retrieve information about them.

docs/sources/flow/reference/components/otelcol.receiver.vcenter.md
```diff
@@ -0,0 +1,133 @@
...
+
+The full list of metrics that can be collected can be found [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/vcenterreceiver/documentation.md)
```
ptodev, line 27: I had a look at this page, and a few things are not clear to me:
* There are [optional metrics](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/vcenterreceiver/documentation.md#optional-metrics). How are these enabled in the Agent? There is a yaml config in the Collector docs, but I don't see how this translates to Agent config.
* Are the [resource attributes](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/vcenterreceiver/documentation.md#resource-attributes) enabled in the Agent by default as well? Can they be disabled?

component/otelcol/receiver/vcenter/vcenter.go
```diff
@@ -0,0 +1,345 @@
...
+type ResourceAttributeConfig struct {
+	Enabled bool `river:"enabled,attr"`
+}
+
+// SetToDefault implements river.Defaulter.
+func (args *ResourceAttributeConfig) SetToDefault() {
+	*args = ResourceAttributeConfig{
+		Enabled: true,
+	}
+}
```
ptodev, line 152: I think we can remove this because "enabled" is mandatory. Hence, there is no need for a default value.

component/otelcol/receiver/vcenter/vcenter.go
```diff
@@ -0,0 +1,345 @@
...
+		"enabled": r.Enabled,
+	}
+}
+
+// SetToDefault implements river.Defaulter.
+func (args *MetricConfig) SetToDefault() {
+	*args = MetricConfig{
+		Enabled: true,
+	}
+}
```
ptodev, line 52: I think we can remove this because "enabled" is mandatory. Hence, there is no need for a default value.

docs/sources/flow/reference/components/otelcol.receiver.vcenter.md
```diff
@@ -0,0 +1,143 @@
...
+
+> **NOTE**: `otelcol.receiver.vcenter` is a wrapper over the upstream
+> OpenTelemetry Collector `vcenter` receiver from the `otelcol-contrib`
+> distribution. Bug reports or feature requests will be redirected to the
+> upstream repository, if necessary.
+
+Multiple `otelcol.receiver.vcenter` components can be specified by giving them
+different labels.
+
+The full list of metrics that can be collected can be found [here][vcenter metrics].
```
ptodev, line 29: We avoid using the word "here" in links:
https://developers.google.com/style/link-text#write-link-text

ptodev (approved):
Thanks you, I'm happy for the PR to be merged. I do think we need to address the currently open comments on a follow-up PR though. 

Also, we need more tests in the follow-up PR:
* tests for unmarshaling config
* a test which starts a mock http server and sends a mock payload, so that we can test that all metrics and attributes are streamed ok

component/otelcol/receiver/vcenter/vcenter.go
```diff
@@ -0,0 +1,345 @@
...
+		"vcenter.vm.name":                      args.VcenterVMName.Convert(),
+	}
+
+	return res
+}
+
+type MetricsBuilderConfig struct {
+	Metrics            MetricsConfig            `river:"metrics,block,optional"`
+	ResourceAttributes ResourceAttributesConfig `river:"resource_attributes,block,optional"`
+}
```
ptodev, line 195: In the follow-up PR we also need to document these options in our docs.

PR #5719, opened 2023-11-06
Clarify blackbox exporter documentation

### LINKED ISSUE #5007: prometheus.exporter.blackbox: issues with documentation and examples
I've had some issues with our current [documentation](https://grafana.com/docs/agent/latest/flow/reference/components/prometheus.exporter.blackbox/#examples) of the prometheus.exporter.blackbox component.

1) As a convention the first Usage section contains the minimum set of parameters required to run the component. In this case, at least _one_ target block is required, which doesn't exist. Maybe we can at least add a dummy `target` block and signify that there can be more, like we do for [prometheus.relabel](https://grafana.com/docs/agent/next/flow/reference/components/prometheus.relabel/#usage)
```
Error: /Users/tpaschalis/river-configs/clustering_scheduling.river:1:1: Failed to build component: decoding River: missing required block "target"

1 |   prometheus.exporter.blackbox "all" {
  |  _^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2 | | }
  | |_^
3 |
```

2) If I define a dummy `target` block, the component runs, but its endpoint reports `Unknown module "http_2xx"`, which is fair, but could use a better UX by erroring out early if neither `config` or `config_file` were passed.

3) The first example for "Collect metrics using a blackbox exporter config file ". Can we lead users where to find a valid `blackbox_modules.yml` file, maybe the one available on the [upstream repo](https://github.com/prometheus/blackbox_exporter/blob/master/example.yml)?

4) The second example for "Collect metrics using an embedded configuration" doesn't seem to work. Hitting the exporter's endpoint results in a cryptic message
```
$ curl localhost:12345/api/v0/component/prometheus.exporter.blackbox.example/metrics
Target parameter is missing
```

marctc:
#### PR Description

- Documentation for blackbox exporter contained some broken config examples. This PR fixes
them and clarifies some areas.
- Added validation for either specifying config file or blackbox module config content.
 
Fixes #5007 

#### PR Checklist

- [ ] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/component/prometheus/exporter/blackbox/blackbox.go
+++ b/component/prometheus/exporter/blackbox/blackbox.go
@@ -106,10 +106,14 @@ func (a *Arguments) Validate() error {
 		return errors.New("config and config_file are mutually exclusive")
 	}
 
+	if a.ConfigFile == "" && a.Config.Value == "" {
+		return errors.New("config or config_file must be set")
+	}
+
 	var blackboxConfig blackbox_config.Config
 	err := yaml.UnmarshalStrict([]byte(a.Config.Value), &blackboxConfig)
 	if err != nil {
-		return fmt.Errorf("invalid backbox_exporter config: %s", err)
+		return fmt.Errorf("invalid blackbox_exporter config: %s", err)
 	}
 
 	return nil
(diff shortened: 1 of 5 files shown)


tpaschalis (approved):
LGTM!

clayton-cornell (approved):
Docs are good as-is

Issue #4598, opened 2023-07-26
clustering: allow cluster-aware component scheduling


tpaschalis:
### Request

We've discussed about introducing a second use case for clustering; fine-grained component scheduling.

This would allow components to opt-in to a new mechanism which would enable running N replicas of a component inside of a cluster, rebalancing whenever the cluster state changes.

### Use case

* Running exporters only on a subset of peers
* Running service discovery centrally and distributing the results within other peers

PR #5725, opened 2023-11-06
simple clustering support for loki.source.kubernetes_events

captncraig:
This implements clustering by simply distributing each namespace (or the only namespace for all when `""` is given) to a different node.

As I submit this I realize there could be a problem since the positions will be stored locally, so if the clustering changes, the new owner could try to catch up. That may need a little thought.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -98,6 +98,8 @@ Main (unreleased)
 - Grafana Agent Operator: `config-reloader` container no longer runs as root.
   (@rootmout)
 
+- `loki.source.kubernetes_events` now supports clustering. (@captncraig)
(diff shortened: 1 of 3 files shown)


rfratto:
As discussed offline, this is a case where #4598 would be a more appropriate solution rather than adding a clustering block to `loki.source.kubernetes_events`; it's not clear if even significantly large clusters have so much event volume that it makes sense to shard event handling across multiple nodes. 

For now, we'll continue to defer deduplication of logs to Loki while we wait for #4598 to be available. 

I'm closing this for now, but we can revisit later if our knowledge about loki.source.kubernetes_events changes.

PR #5726, opened 2023-11-07
Add fallback cache to remote.http

### LINKED ISSUE #5506: Implement configuration cache for `module.*` components
### Request

If the agent is running and the remote endpoint for module.http goes down, the agent continuously running with the lastest functional configuration.

However, if the agent will be restart, the agent refuse to restart, since external endpoints are not available which will be result in losing metrics.

I'm having my eye on the Remote Agent Configuration. 
https://github.com/grafana/agent/blob/8186196db6aa3931a0197ecdd163265db30aeac0/pkg/config/agentmanagement.go#L229-L232C1 I'm aware that is only for internal experimenting, however the concept of caching valid configuration is something that i'm looking for. 

### Use case

This would increase the reliability of the Agent, if the configuration endpoint is not available. At the moment, the configuration data plane must be considered as critical component. If the configuration endpoints are gone and agents are restarting for various reasons, the are out of the monitoring.

spartan0x117:
#### PR Description
Adds a `fallback_cache` configuration block to `remote.http` (which `module.http` inherits). Also adds new `DebugInfo` to `remote.http` and `module.http`. Secrets _can_ be cached. Sets the file permissions for the cache to `600`. The filename is a `sha1` hash of the url, to both hide any secrets that could be in the url itself and allow for switching between urls dynamically during runtime by hot-reloading the config.

Example config:
```river
module.http "remote_config" {
  url            = "http://localhost:8000/demo.river"
  poll_frequency = "10s"

  fallback_cache {
    enabled = true
  }
}
```

Example of new `DebugInfo` for `remote.http` and `module.http`:
image

Fixes grafana/alloy#332

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Documentation added
- [X] Tests updated


rfratto (commented):
Especially going into v1, I'm increasingly weary about adding new configuration options, without knowing whether we intend to keep them forever. In particular, my thoughts are:

* I'm not sure if it ever makes sense for someone to disable this functionality. 

* I'm not sure when it makes sense to configure a max age, and what advise we should give a user who is trying to tune the max age. 

* I'm not sure if there's any risk of storing secrets on disk, as long as the cache files have the permissions of `600` and we document that files are cached on disk. 

Until we have more data showing that we need all three of these options, what do you think about doing one of the following:

* Marking this specific block as "experimental" or "beta" to allow us to more comfortably make breaking changes to it, and to communicate that we're not sure about what the final form of caching would take. 

* Removing the configuration options entirely and choosing fixed options for the user, until we get more data showing that the options are required (i.e., always cache with no max age / or a reasonable max age). 

WDYT? (If you come up with other ways that allow us to retain the ability to experiment with this or change things prior to 1.0, feel free to propose them 🙂)

rfratto:
> Especially going into v1, I'm increasingly weary about adding new configuration options, without knowing whether we intend to keep them forever. In particular, my thoughts are:
> 
> * I'm not sure if it ever makes sense for someone to disable this functionality.
> * I'm not sure when it makes sense to configure a max age, and what advise we should give a user who is trying to tune the max age.
> * I'm not sure if there's any risk of storing secrets on disk, as long as the cache files have the permissions of `600` and we document that files are cached on disk.
> 
> Until we have more data showing that we need all three of these options, what do you think about doing one of the following:
> 
> * Marking this specific block as "experimental" or "beta" to allow us to more comfortably make breaking changes to it, and to communicate that we're not sure about what the final form of caching would take.
> * Removing the configuration options entirely and choosing fixed options for the user, until we get more data showing that the options are required (i.e., always cache with no max age / or a reasonable max age).
> 
> WDYT? (If you come up with other ways that allow us to retain the ability to experiment with this or change things prior to 1.0, feel free to propose them 🙂)

And, $0.02: personally, I prefer hard-coding things and keeping the configuration surface small and adding configuration options later once we know they're necessary. That being said, I'm OK with whatever you decide here.

spartan0x117:
That is reasonable to me! I do think it makes sense to be able to toggle the caching on/off at least, in case a user doesn't want a secret to be cached on disk at all but still wants to use components that embed `remote.http` (or use it directly). 

In light of that, I don't recall there being a concept of individual arguments/features having a stability, do you have any issue with just adding a single new argument: `enable_fallback_cache`? I could also add the block with only the one argument and give it a stability of `beta / experimental`, but that seems a bit much, IMO.

rfratto:
> In light of that, I don't recall there being a concept of individual arguments/features having a stability

We've been making use of it for [clustering](https://grafana.com/docs/agent/latest/flow/reference/components/prometheus.scrape/#clustering-beta), since clustering is in beta but is configured in components which are stable. 

> do you have any issue with just adding a single new argument: enable_fallback_cache? I could also add the block with only the one argument and give it a stability of beta / experimental, but that seems a bit much, IMO.

Given the same argument as before, adding this as a block to allow for additional future configuration options makes sense to me. (Even if the block doesn't get a stability label.):

```
fallback_cache { enabled = true }
```

component/remote/http/http.go
```diff
@@ -34,10 +38,11 @@ func init() {
...
 type Arguments struct {
-	URL           string        `river:"url,attr"`
-	PollFrequency time.Duration `river:"poll_frequency,attr,optional"`
-	PollTimeout   time.Duration `river:"poll_timeout,attr,optional"`
-	IsSecret      bool          `river:"is_secret,attr,optional"`
+	URL                  string        `river:"url,attr"`
+	PollFrequency        time.Duration `river:"poll_frequency,attr,optional"`
+	PollTimeout          time.Duration `river:"poll_timeout,attr,optional"`
+	IsSecret             bool          `river:"is_secret,attr,optional"`
+	FallbackCacheEnabled bool          `river:"fallback_cache_enabled,attr,optional"`
```
rfratto, line 45: Would calling this `cache` instead of `fallback_cache` lose meaning? (I'm concerned about the length of the name being annoying to type)

spartan0x117: Yeah, I mulled over this one a bit. My reasoning was that it might be confused with a cache that optimizes queries?

spartan0x117: Perhaps just a `fallback` block?
```river
fallback { enabled = true }
```

rfratto: Maybe @clayton-cornell would have opinions for what name be commonly understood here?

docs/sources/shared/flow/reference/components/remote-http-fallback-cache-block.md
jkroepke, line 1: Thanks for pushing this feature. About the fallback_cache/enable/max_age discussion, I would like the ask if it would make sense to respect HTTP [Cache-Control](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control?retiredLocale=de) header? 

I have no pro or counter arguments except it would increase the complexity a little bit. Explicit configuration should be preferred.

jkroepke:
If the remote endpoint is not reasonable, but the fallback cache is present. Whats the components health status? OK? Error?

spartan0x117:
> If the remote endpoint is not reasonable, but the fallback cache is present. Whats the components health status? OK? Error?

The component's health would be `Healthy`, but the Agent would have logged an `error` for the failure to reach the endpoint and a `warn` when it fell-back to the cache (or another `error` if the fallback failed). I could also possibly add a metric for the requests by response code and one for outright errors (i.e. the remote is entirely unreachable).

jkroepke:
There is only Healthy and Unhealthy status? A warning statue does not exists for an components?

It would be great to have an prometheus metric, if the origin source is used or an cached configuration. As SRE, I would like to know if cache is in use, because than something is wrong.

spartan0x117:
> There is only Healthy and Unhealthy status? A warning statue does not exists for an components?

There is `Healthy`, `Unhealthy`, and `Unknown`. Since the component itself is still technically functioning as expected, I kept the status as healthy. I'll add some metrics to make it detectable, though.

spartan0x117:
I've added `remote_http_fallbacks_total` and `remote_http_fallbacks_failed_total` as metrics for being able to determine when the component is unable to reach the remote and also when it is unable to read from the local cache (doesn't exist, for example).

component/remote/http/http.go
clayton-cornell: I missed this one.

Do we need `Enabled` in the bool name? Id' probably opt for either `Fallback` or `FallbackCache`. It's a bool, so if it's set to t`true` it should be clear it's enabled.... or it seems that way to me.

docs/sources/shared/flow/reference/components/remote-http-fallback-cache-block.md
```diff
@@ -0,0 +1,18 @@
...
+- /docs/agent/shared/flow/reference/components/remote-http-fallback-cache-block/
+canonical: https://grafana.com/docs/agent/latest/shared/flow/reference/components/remote-http-fallback-cache-block/
+description: Shared content, remote http fallback cache
+headless: true
+---
+
+The `fallback_cache` block configures a local cache that is used as a fallback if
+the remote endpoint is unavailable or returns an error.
+
+Cache files are created with `600` permissions and owned by the user running the
```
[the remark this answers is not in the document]
mattdurham: Go will attempt to map unix permission to windows it works well enough until you get to edge cases but this works as expected on windows. Owned by the user is the key thing here.

spartan0x117:
Just wanted to bump here to see if I should work towards merging or close this PR :smile:

clayton-cornell:
@spartan0x117 I have no further input on the doc side.

clayton-cornell:
@spartan0x117 This PR is a bit stale... is it still valid?

rfratto:
I'm going to close this for now as stale. If this will be picked back up, I think it should be moved to the Alloy repo and then we can decide whether it makes sense to backport.

PR #5732, opened 2023-11-07
Add a table to the flow description showing benefits and strengths of flow

clayton-cornell:
#### PR Description

This PR adds 3 tables to the Static/Flow decision topic. This is groundwork information for future flow topic refinement.

Some minor tidy and cleanup is included in this PR. This stats the alignment between the various entry topics. Longer term plan here is to reduce duplication and simplify the message (future PRs to focus on the cleanup and merging of information).

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


docs/sources/about.md
```diff
@@ -76,14 +84,12 @@ You should run the Static mode Kubernetes operator when:
...
 September 29, 2022.
 
-Flow mode was introduced as a re-imagining of Grafana Agent with a focus on
-vendor neutrality, ease-of-use, improved debuggability, and ability to adapt to
-the needs of power users by adopting a configuration-as-code model.
-
-Flow mode is considered to be the future of the Grafana Agent project.
+Grafana Agent Flow mode focuses on vendor neutrality, ease-of-use,
+improved debuggability, and ability to adapt to the needs of power
+users by adopting a configuration-as-code model.
```
jdbaldry, line 92: nit:

Vale doesn't recognize "debuggability" as a word. Perhaps "easier debugging"? I guess this is a comparison to static mode. Should that be more clear?

nit:

[Semantic line breaks](https://sembr.org/) are better for future line based diffing in the GitHub UI and result in the same HTML output.

docs/sources/about.md
```diff
@@ -14,15 +14,14 @@ weight: 100
 
 # Introduction to Grafana Agent
 
-Grafana Agent is a vendor-neutral, batteries-included telemetry collector. It
-is designed to be flexible, performant, and compatible with multiple ecosystems
-such as Prometheus and OpenTelemetry.
+Grafana Agent is a flexible, high performance, vendor-neutral telemetry collector. It's fully compatible with the most popular
+open source observability standards such as OpenTelemetry (OTel) and Prometheus.
```
jdbaldry, line 18: nit:

[Semantic line breaks](https://sembr.org/) are better for future line based diffing in the GitHub UI and result in the same HTML output.

```suggestion
Grafana Agent is a flexible, high performance, vendor-neutral telemetry collector. It's fully compatible with the most popular open source observability standards such as OpenTelemetry (OTel) and Prometheus.
```

docs/sources/about.md
clayton-cornell: Yeah I didn't clean up much in this pass (like debuggability). There's a lot of rewrite that's needed.

The semantic linebreaks are... difficult to manage because everyone has different linebreaks based on what tool they are using. Every edit ends up with wildly different linebreaks. Not sure how to manage/enforce this long term.

jdbaldry: Yeah, the argument of semantic line breaks is that they provide the most meaningful diff when using a line based diff. GitHub primarily works with line based diffing and that means that semantic line breaks provide the most easy to read diff in the primary place where we view patches.

Converesely, arbitrary line lengths are maybe sympathetic to those reading in a terminal but I would expect that any modern text editor or other presentation layer should be able to perform soft wrapping of long lines.

docs/sources/about.md
```diff
@@ -61,12 +68,10 @@ You should run Static mode when:
...
 
-The [Static mode Kubernetes operator][] is a variant of Grafana Agent first
-introduced on June 17, 2021. It is currently in beta.
+The [Static mode Kubernetes operator][] is a variant of Grafana Agent introduced on June 17, 2021. It's currently in beta.
 
-The Static mode Kubernetes operator was introduced for compatibility with
-Prometheus Operator, allowing static mode to support resources from Prometheus
-Operator, such as ServiceMonitors, PodMonitors, and Probes.
+The Static mode Kubernetes operator provides compatibility with Prometheus Operator,
+allowing static mode to support resources from Prometheus Operator, such as ServiceMonitors, PodMonitors, and Probes.
```
rfratto, line 74: Sorry for scope creep, but the "you should run Static mode Kubernetes operator" section doesn't feel right anymore, since Flow mode does the one thing listed in the bullet point. 

Should we change this to something like "It is no longer recommended to use Static mode Kubernetes operator for new deployments of Grafana Agent?"

cc @richardqlam

docs/sources/about.md
```diff
@@ -76,33 +81,51 @@ You should run the Static mode Kubernetes operator when:
...
+|                          | Prometheus Agent mode | Grafana Agent Static mode | Grafana Agent Operator | OpenTelemetry Collector | Grafana Agent Flow mode |
+|--------------------------|-----------------------|---------------------------|------------------------|-------------------------|-------------------------|
+| **Kubernetes native**    | No                    | No                        | Yes                    | Yes                     | Yes                     |
+| **Clustering**           | No                    | No                        | No                     | No                      | No                      |
+| **Prometheus rules**     | No                    | No                        | No                     | No                      | Yes                     |
+| **Native Vault support** | No                    | No                        | No                     | No                      | Yes                     |
 
-  * **Ecosystem transformation**: You need to be able to convert Prometheus and
-    Loki pipelines to and from OpenTelmetry Collector pipelines.
+#### Grafana Cloud solutions
```
rfratto, line 121: It feels strange to me for these to be subheaders of Flow mode, since it also describes the features of all other nodes. Should there be a dedicated `h2` section for these?

docs/sources/about.md
```diff
@@ -76,33 +81,51 @@ You should run the Static mode Kubernetes operator when:
...
+  * **Grafana Pyroscope support**: Support for collecting profiles for Grafana Pyroscope.
+
+#### Core telemetry
+
+|              | Prometheus Agent mode | Grafana Agent Static mode | Grafana Agent Operator | OpenTelemetry Collector | Grafana Agent Flow mode  |
+|--------------|-----------------------|---------------------------|------------------------|-------------------------|--------------------------|
+| **Metrics**  | Prometheus            | Prometheus                | Prometheus             | OTel                    | [Prometheus][], [OTel][] |
+| **Logs**     | No                    | Loki                      | Loki                   | OTel                    | [Loki][], [OTel][]       |
+| **Traces**   | No                    | OTel                      | OTel                   | OTel                    | [OTel][]                 |
+| **Profiles** | No                    | No                        | No                     | No                      | Pyroscope                |
```
thampiotr, line 110: Should we make Flow the first column in these tables? It seems to me more eye traffic will be on the left and also in case of narrow browser windows.

docs/sources/about.md
richardqlam: I would leave the existing language and then add what @rfratto recommended below. So something like, "As of xyz date it is recommended to use Grafana agent Flow mode to support xyz use cases"

clayton-cornell: @rfratto @richardqlam How about a note that says something like:

```
Grafana Agent version 0.37 and newer provides Prometheus Operator compatibility in Flow mode. You should run Grafana Agent Flow mode for all new Grafana Agent deployments.
```

We could add something similar to the Static section as well - the first step in nudging people to use Flow over Static or Operator.

rfratto: I like that 👍

clayton-cornell: Added in a new commit

docs/sources/about.md
clayton-cornell: I moved the tables up to the top of the document and added a simple lead-in

docs/sources/about.md
clayton-cornell: @thampiotr I switched the order of the columns going with Flow, Static, Operator, OTel, Prometheus.

thampiotr (approved):

PR #5747, opened 2023-11-09
Flow: syslog: use best-effort, partially parsed message

thampiotr:


thampiotr:
This wouldn't prevent closing of the syslog connection though, so may not be worth doing. See https://github.com/grafana/loki/issues/4270

PR #5751, opened 2023-11-10
Make worker pool more resilient to slow tasks

thampiotr:
#### PR Description

The current worker pool distributes tasks among goroutines using hash of the task's key. In presence of a very slow task, e.g. due to a bug, we can have one goroutine that will be effectively stuck. All other tasks that are unfortunate to hash to the same goroutine will also be affected.

We could prevent this if we were to not rely on hash, but instead assign tasks randomly to any goroutine. The problem with this approach, however, is that a slow task may come in again and again and soon we will have every goroutine in the pool trying to execute the very slow task. So, in order to prevent this, we must only allow one task with given key to be executing at any given time. 

In this PR changes the worker pool implementation:
* We _maintain_ the existing behaviour of only one task with given key can be waiting to be executed at a time - this reduces duplicate work.
* We add a mechanism that tracks the task keys that are currently running and only allow the goroutines to run one task with a given key at a time, to prevent the problem described above.
* We assign tasks eligible to run on a random available goroutine
As a result a very slow task will a) not be able to take over all the goroutines b) will not block other tasks from executing.

As discussed in previous PRs - the worker pool we need for the Agent requires a quite unique behaviour as described above and we are not able to use an existing library to fulfil these requirements.

#### Notes to the Reviewer

Tested this in a dev cluster, where, as expected, I can see a drop in times that tasks had to wait for a goroutine to pick them up:

image
(left half - before change, right half - after the change)

The size of this effect is expected to be larger in other large clusters that have slower components.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated

### DIFF
--- a/pkg/flow/internal/worker/worker_pool.go
+++ b/pkg/flow/internal/worker/worker_pool.go
@@ -2,8 +2,6 @@ package worker
 
 import (
 	"fmt"
-	"hash/fnv"
-	"math/rand"
 	"runtime"
 	"sync"
 )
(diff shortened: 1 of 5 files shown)


pkg/flow/internal/worker/worker_pool_test.go
```diff
@@ -177,21 +180,102 @@ func TestWorkerPool(t *testing.T) {
...
-			require.Equal(t, 1, pool.QueueSize())
+			require.Equal(t, 2, pool.QueueSize())
 
 			// Third task cannot be accepted, because the queue is full
 			err = pool.SubmitWithKey("k3", func() { tasksDone.Inc() })
 			require.ErrorContains(t, err, "queue is full")
+			require.Equal(t, 2, pool.QueueSize())
+		})
+
+		t.Run("should not block when one task is stuck", func(t *testing.T) {
```
thampiotr, line 200: In the previous implementation, this test would fail.

pkg/flow/internal/worker/worker_pool_test.go
```diff
@@ -90,17 +91,18 @@ func TestWorkerPool(t *testing.T) {
 				tasksDone.Inc()
 			})
 			require.NoError(t, err)
+			defer func() { close(blockFirstTask) }()
 
 			// Wait for the first task to be running already and blocking the worker
 			<-firstTaskRunning
-			require.Equal(t, 0, pool.QueueSize())
+			require.Equal(t, 1, pool.QueueSize())
```
thampiotr, line 98: The definition of QueueSize changed to also include the tasks that are currently executing. This makes things a bit simpler in the implementation, while still being able to perform its function: tell us when queue is building up.

pkg/flow/internal/worker/worker_pool.go
```diff
@@ -12,119 +10,180 @@ type Pool interface {
...
+		}
+	}
+
+	// Return if we didn't find any task ready to run
+	if !found {
+		return
+	}
+
+	// Remove the task from waiting and add it to running set
+	w.waitingOrder = append(w.waitingOrder[:index], w.waitingOrder[index+1:]...)
```
rfratto, line 170: What's the behavior of append when removing an element in this way; is it copying the whole slice? I'm curious if this has any performance implications for allocations long term, and if a traditional linked list might be more appropriate.

rfratto (approved):
LGTM, nice improvement! I have one question about performance implications, but I think this is good to go.

pkg/flow/internal/worker/worker_pool.go
thampiotr: It's a good point as a linked list should be a natural choice for this algorithm. 

The above code will copy the second argument element-by-element, as per https://go.dev/blog/slices-intro, but it will NOT allocate since we're reducing the size of the slice. This will be slower than a linked-list approach, where removal is O(1) instead of O(N). However, this code is not a bottleneck for two reasons: 1) this slice copying is dominated by other, slower processes - like evaluating components that can take multiple milliseconds, compared to slice copy that takes few nanoseconds. 2) the size of the queue is usually less than 10 elements even in our largest clusters. 

If we had a linked list in standard Go library, I'd use it, but we don't and it's not a bottleneck, so I'm proposing we stick to this more idiomatic implementation.

thampiotr: > If we had a linked list in standard Go library, I'd use it, but we don't and it's not a bottleneck, so I'm proposing we stick to this more idiomatic implementation.

There _is_ linked list in [Go standard library](https://pkg.go.dev/container/list) (thanks!) but it allocates memory, so it would not be more performant here... and it's still not a bottleneck :) I have some benchmarks here: https://github.com/grafana/agent/pull/5765 which we can merge if we want to keep it for the posteriority.

pkg/flow/internal/worker/worker_pool.go
```diff
@@ -12,119 +10,180 @@ type Pool interface {
...
+}
+
+func (w *workQueue) tryEnqueue(key string, f func()) (bool, error) {
+	w.lock.Lock()
+	defer w.lock.Unlock()
+
+	// Don't enqueue if same task already waiting
+	if _, exists := w.waiting[key]; exists {
+		return false, nil
+	}
```
tpaschalis, line 120: cc @thampiotr

In this case, where we try to enqueue the same task key is already waiting, don't we want to replace the waiting func with the new incoming one instead?

tpaschalis (commented):
Reviewing after the fact, but LGTM; I like the new semantics and the readability 🙂 

Adding a small nit about how `tryEnqueue` semantics below and whether they matter in the context of Flow graph updates.

pkg/flow/internal/worker/worker_pool.go
thampiotr: We can, but we don't need to do that. The `worker_pool` is a pretty unique implementation that I only ever see being used by the `loader` and the `loader` always submits an equivalent function with a given key. Whether we replace the one that is already queued or not, doesn't really matter.

PR #5752, opened 2023-11-10
Fix logging issue

thampiotr:
#### PR Description

The recent _unreleased_ change to the logging library had a bug where in some cases the log lines would be reported as INFO, regardless of the level set.

This PR fixes this issue and adds tests.

#### PR Checklist

- [x] Tests updated


rfratto (approved):
Thanks for fixing this!

PR #5753, opened 2023-11-10
Improve otel integration tests

wildum:
#### PR Description

I simplified the way the otel metrics are generated and added more metrics including exponential histograms.


integration-tests/tests/otlp-metrics/otlp_metrics_test.go
```diff
@@ -1,23 +1,80 @@
...
+	return fmt.Sprintf("%s%s{test_name='otlp_metrics'}", promURL, metricName)
+}
+
+func TestOTLPMetrics(t *testing.T) {
+	tests := []struct {
+		query  string
+		metric string
+	}{
+		// TODO: better differentiate these metric types?
+		{metricQuery("example_counter"), "example_counter"},
```
thampiotr, line 23: Could we remove duplicated strings here and in other places? I think you can find a way to only write "example_counter" etc only once per line.

integration-tests/tests/otlp-metrics/otlp_metrics_test.go
```diff
@@ -1,23 +1,80 @@
...
+	assert.EventuallyWithT(t, func(c *assert.CollectT) {
+		err := common.FetchDataFromURL(query, &metricResponse)
+		assert.NoError(c, err)
+		if assert.NotEmpty(c, metricResponse.Data.Result) {
+			assert.Equal(c, metricResponse.Data.Result[0].Metric.Name, expectedMetric)
+			assert.Equal(c, metricResponse.Data.Result[0].Metric.TestName, "otlp_metrics")
+			if assert.NotNil(c, metricResponse.Data.Result[0].Histogram) {
+				histogram := metricResponse.Data.Result[0].Histogram
+				assert.NotEmpty(c, histogram.Data.Count)
+				assert.NotEmpty(c, histogram.Data.Sum)
```
thampiotr, line 60: Is `Count` and `Sum` a number? What does checking that they're empty mean?

Do we want to instead check they're greater than 0, or maybe greater than 10? (they will eventually be greater than 10 from what I gather).

wildum: count and sum are numbers represented as strings in the answer. Checking that the string is not empty means that there is a number. I can convert it as a number and check that it is > 10 for better precision.

thampiotr: ah, I see - yeah, I guess you can can check first if empty, then try convert and pick some number that will be eventually reached :)

integration-tests/docker-compose.yaml
```diff
@@ -2,7 +2,7 @@ version: "3"
 services:
 
   mimir:
-    image: grafana/mimir:latest
```
wildum, line 5: got pranked here, the tag points now to an older version that does not support exponential histograms (it went back to 2.9.3)

thampiotr (approved):

Issue #5733, opened 2023-11-07
config: Support Arguments on root module


jkroepke:
### Request

Grafana Agent have [arguments](https://grafana.com/docs/agent/latest/flow/reference/config-blocks/argument/) block which only exists on module.

I had the idea to support arguments on the root context, too. arguments can be defined via CLI argument, e.g. `--set arguments=value`.

### Use case

It would be an alternative in addition to environment variables parameterize static configuration.

tpaschalis:
That is an interesting idea to explore, and an alternative to the [debug.variables](https://github.com/grafana/agent/issues/2772) component idea we'd discussed in the past, with a tradeoff between dynamic behavior and more understandable configuration.

I'm not 100% sold on it yet, but it can be discussed as part of #5547.

jkroepke:
Origin: https://docs.docker.com/engine/reference/commandline/buildx_bake/#set

I found that pattern on the docker buildx plugin. At the moment, I have to use environment variables which is also fine.

PR #5762, opened 2023-11-13
Flow Windows service: Support environment variables

jkroepke:
#### PR Description

Windows support environments for global, per user and per service. All three os native methods have downsides

* Global scope exposing environment variables to all program
* Per User: It's easy for human users to change own environment variables, but hard too for system users.
* Per Service: If the service is using a local system account, then environment variables will be read once from registry. Changes requires a reboot.

I copied the replicate the Arguments logic for Environment. Environment are needed to inject configuration from outside. Define configuration via CLI argument is not supported (see: https://github.com/grafana/agent/issues/5733)

#### Notes to the Reviewer

Tested on Windows Server 2022.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


docs/sources/flow/setup/install/windows.md
```diff
@@ -58,6 +58,13 @@ To do a silent install of Grafana Agent on Windows, perform the following steps.
 * `/DISABLEREPORTING=<yes|no>` Disable [data collection][]. Default: `no`
 * `/DISABLEPROFILING=<yes|no>` Disable profiling endpoint. Default: `no`
 
+## Service Configuration
+
+Grafana Agent is using the Windows Registry `HKLM\Software\Grafana\Grafana Agent Flow` for service configuration.
+
+* `Arguments` (Type `REG_MULTI_SZ`) Each value represents a binary argument for grafana-agent-flow binary.
```
mattdurham, line 65: Do we need any sort of documentation around escaping the values here? Thinking quotes ,double quotes etc

jkroepke: I think no. `Arguments` is a multi string value which is natively passed to an go slice. 

Of couse you have to quote think, if you are running reg.exe from CLI. But if you using RegEdit, then no additional quoting is required.

It's could be important to mention that there is no shell parsing. If you quote path, like `"C:\file.json"`, then the Agent complains that the path `"C:\file.json"` does not exists (because the quotes are not stripped)

docs/sources/flow/setup/install/windows.md
```diff
@@ -57,6 +57,14 @@ To do a silent install of Grafana Agent on Windows, perform the following steps.
...
 * `/DISABLEREPORTING=<yes|no>` Disable [data collection][]. Default: `no`
 * `/DISABLEPROFILING=<yes|no>` Disable profiling endpoint. Default: `no`
+* `/ENVIRONMENT="KEY=VALUE\0KEY2=VALUE2"` Define environment variables for Windows Service. Default: ``
+
+## Service Configuration
+
+Grafana Agent is using the Windows Registry `HKLM\Software\Grafana\Grafana Agent Flow` for service configuration.
+
+* `Arguments` (Type `REG_MULTI_SZ`) Each value represents a binary argument for grafana-agent-flow binary.
+* `Environment` (Type `REG_MULTI_SZ`) Each value represents a environment value `KEY=VALUE` for grafana-agent-flow binary.
```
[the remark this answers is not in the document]
jkroepke: Do you mean the output of `reg query` ?

cmd/grafana-agent-service/config_windows.go
```diff
@@ -16,6 +16,10 @@ type config struct {
 	// not included.
 	Args []string
 
+	// Environment holds arguments to pass to the Grafana Agent binary. os.Environ is
+	// included.
```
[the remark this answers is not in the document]
jkroepke: @ptodev what did you think about this?

```suggestion
	// Environment holds arguments to pass to the Grafana Agent binary. 
	// Each item represents an environment variable in form "key=value".
	// All environments variables from the current process with be merged into Environment
```

packaging/grafana-agent-flow/windows/install_script.nsis
```diff
@@ -51,6 +52,7 @@ Section "install"
   ${GetParameters} $PassedInParameters
   ${GetOptions} $PassedInParameters "/DISABLEPROFILING=" $DisableProfiling
   ${GetOptions} $PassedInParameters "/DISABLEREPORTING=" $DisableReporting
+  ${GetOptions} $PassedInParameters "/ENVIRONMENT=" $Environment
```
rfratto, line 55: Is it possible to pass more than one environment variable with this? It seems like this is restricted to a single variable, which is a bit weird to me.

cmd/grafana-agent-service/config_windows.go
```diff
@@ -42,9 +46,15 @@ func loadConfig() (*config, error) {
 		return nil, fmt.Errorf("failed to retrieve key Arguments: %w", err)
 	}
 
+	env, _, err := agentKey.GetStringsValue("Environment")
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve key Arguments: %w", err)
```
rfratto, line 51:
```suggestion
		return nil, fmt.Errorf("failed to retrieve key Environment: %w", err)
```

cmd/grafana-agent-service/config_windows.go
```diff
@@ -42,9 +46,15 @@ func loadConfig() (*config, error) {
 		return nil, fmt.Errorf("failed to retrieve key Arguments: %w", err)
 	}
 
+	env, _, err := agentKey.GetStringsValue("Environment")
```
rfratto, line 49: Since this amends additional environment variables (rather than replacing the entire environment), I feel this should be named to indicate that. something like "AdditionalEnvironmentVars" or "ExtraEnvironmentVars"

rfratto (commented):
Thank you!

packaging/grafana-agent-flow/windows/install_script.nsis
jkroepke: Yes, see docs above

`/ENVIRONMENT="KEY=VALUE\0KEY2=VALUE2"`

The REG_MULTI_SZ typ is a string with null byte separator. The value is passed one by one to [`reg add`](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/reg-add#examples) command. 


I agree, this feels a bit unintuitive on first view, but from technical point of view it has some benefits. If we decide to use `;` as separator, then an additional logic is required to escape separator. And the logic needs to be written in NSIS script.

cmd/grafana-agent-service/config_windows.go
ptodev: They're not necessary additional - they might shadow an already existing one.

ptodev: Actually, I'm not sure if the current implementation will shadow env vars properly, but if it doesn't I'd consider that a bug.

cmd/grafana-agent-service/config_windows.go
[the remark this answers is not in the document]
ptodev: Do you think this would be more clear (and correct?):

`// Environment holds environment variables for the Grafana Agent service.`

docs/sources/flow/setup/install/windows.md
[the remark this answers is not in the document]
ptodev: I guess so? I've only used `reg export` and I suppose `reg query` does a similar job. As long as people can see the values in a format they can recognise, it's ok.

cmd/grafana-agent-service/config_windows.go
jkroepke: I would also recommend to keep it Environment. I feel it natural that the current environment will be enriched instead replaced.

docs/sources/flow/setup/install/windows.md
```diff
@@ -57,6 +57,14 @@ To do a silent install of Grafana Agent on Windows, perform the following steps.
 * `/CONFIG=<path>` Path to the configuration file. Default: `$INSTDIR\config.river`
 * `/DISABLEREPORTING=<yes|no>` Disable [data collection][]. Default: `no`
 * `/DISABLEPROFILING=<yes|no>` Disable profiling endpoint. Default: `no`
+* `/ENVIRONMENT="KEY=VALUE\0KEY2=VALUE2"` Define environment variables for Windows Service. Default: ``
+
+## Service Configuration
+
+Grafana Agent is using the Windows Registry `HKLM\Software\Grafana\Grafana Agent Flow` for service configuration.
```
clayton-cornell, line 64:
```suggestion
Grafana Agent uses the Windows Registry `HKLM\Software\Grafana\Grafana Agent Flow` for service configuration.
```

jkroepke:
Is there anything from my side todo here?

mattdurham (approved):
Looking good from my end

jkroepke:
I fixed the Changelog entry, since 0.38 is released.

PR #5764, opened 2023-11-14
Add an otelcol.processor.resourcedetection component

ptodev:
This is a draft PR to port over the [resource detection processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.87.0/processor/resourcedetectionprocessor/README.md). There are a few things which we need to agree on before the PR is taken out of "draft" stage. I will open comments on the specific lines where we need consensus.


component/otelcol/processor/resourcedetection/resourcedetection.go
```diff
@@ -0,0 +1,204 @@
...
+
+	// DetectorConfig is a list of settings specific to all detectors
+	DetectorConfig DetectorConfig `river:",squash"`
+
+	// HTTP client settings for the detector
+	// Timeout default is 5s
+	// Client otelcol.HTTPClientArguments `river:",squash"`
+	Timeout time.Duration `river:"timeout,attr,optional"`
+	//TODO: Uncomment this later? Can we just get away with a timeout, or do we need all the http client settings?
+	//      It seems that HTTP client settings are only used in the ec2 detection via ClientFromContext.
```
ptodev, line 64: The HTTP settings in the Collector expose [HTTP client settings](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.87.0/processor/resourcedetectionprocessor/config.go#L37-L39) for two reasons:
* The timeout is used to cancel resource detections if they take too long. This use case has nothing to do with HTTP at all.
* The ec2 detection [uses the HTTP settings](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.87.0/processor/resourcedetectionprocessor/internal/aws/ec2/ec2.go#L101-L108) if they are provided. This feature is not documented in the OTel docs.

For now I would like to expose just the `timeout` parameter:
* This is because it would be super confusing to document `timeout` as both a timeout for HTTP and for ec2.
* I'll leave a TODO to implement it in the future.

I wish the OTel Collector worked differently:
* There should be a non-HTTP `timeout` settings for all detectors
* There should be HTTP settings specific to the ec2 detector. That way ec2 won't use `context`, and it will be possible to have different HTTP settings for different detectors.

Maybe it'll be work raising these issues with the upstream maintainers.

docs/sources/flow/reference/components/otelcol.processor.resourcedetection.md
```diff
@@ -0,0 +1,285 @@
...
+`override`  | `bool`         | Configures whether existing resource attributes should be overriden or preserved.      | `true`  | no
+`timeout`   | `duration`     | Timeout by which all specified detectors must complete.                                | `"5s"`  | no
+
+## Blocks
+
+The following blocks are supported inside the definition of
+`otelcol.processor.resourcedetection`:
+
+Hierarchy                              | Block                                                       | Description                                       | Required
+-------------------------------------- | ----------------------------------------------------------- | ------------------------------------------------- | --------
```
ptodev, line 52: I'd like to rename this section to "Configuration" and remove the table with blocks. In its place I could put a table with "root level" attributes and a table with "root level" blocks. This was already described in #5390

component/otelcol/processor/resourcedetection/internal/resource_attribute_config/resource_attribute_config.go
```diff
@@ -0,0 +1,28 @@
+package resource_attribute_config
+
+// Configures whether a resource attribute should be enabled or not.
```
ptodev, line 3: Note that `otelcol.processor.resourcedetection` will work differently from the Collector's original processor:
* In the Collector, many resource attributes are enabled by default. But some resource attributes are not.
* In the Agent, all resource attributes will be disabled by default. Users will have to enable them explicitly.

There are a few reasons why I decided to do it this way:
* If the Agent sets various resource attributes by default, the Collector might print warnings which are confusing to the user. This is the `enabledSetByUser` parameter mentioned in the comment below.
* So that we don't have to waste time keeping the Agent's defaults in sync with the Collector's.

For example, this is a labda resource detection with all resource attributes enabled:
```
otelcol.processor.resourcedetection "default"  {
  detectors = ["lambda"]
  lambda {
    resource_attributes {
      aws.log.group.names { enabled = true }
      aws.log.stream.names { enabled = true }
      cloud.platform { enabled = true }
      cloud.provider { enabled = true }
      cloud.region { enabled = true }
    }
  }
  output {
    logs = [otelcol.exporter.otlp.default.input]
  }
}
```

If a user wants to disable a resource attribute, all they need to do is just not list it:
```
otelcol.processor.resourcedetection "default"  {
  detectors = ["lambda"]
  lambda {
    resource_attributes {
      aws.log.group.names { enabled = true }
    }
  }
  output {
    logs = [otelcol.exporter.otlp.default.input]
  }
}
```

They could also explicitly set it to `enabled = false`:
```
otelcol.processor.resourcedetection "default"  {
  detectors = ["lambda"]
  lambda {
    resource_attributes {
      aws.log.group.names { enabled = true }
      cloud.region { enabled = false }
    }
  }
  output {
    logs = [otelcol.exporter.otlp.default.input]
  }
}
```

docs/sources/flow/reference/components/otelcol.processor.resourcedetection.md
```diff
@@ -0,0 +1,261 @@
...
+
+### ecs > resource_attributes
+
+### eks
+
+### eks > resource_attributes
+
+### elasticbeanstalk
+
+### elasticbeanstalk > resource_attributes
```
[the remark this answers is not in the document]
clayton-cornell: @ptodev I know there's been no visible movement on PR #5381 but.... it's a good proposal, and I've been wanting to get started on implementing it across all Ref topics... just haven't got to it yet.

Also.. guidelines for docs layouts should be that.. guidelines. If following the guideline results in a mess, we should be flexible and adjust the approach. If there's a better way (even when that means adjusting existing docs) that makes things more clear and understandable, then I'd say go for it. In this case there's already a PR open suggesting the same change so this PR is not a change that exists in isolation.

component/otelcol/processor/resourcedetection/internal/resource_attribute_config/resource_attribute_config.go
mar4uk: I completely understand the approach you are suggesting and support it. The only concern I have is will this change prevent the `otelcol.processor.resourcedetection` component from becoming a distribution of otel collector resourcedetectionprocessor?

ptodev: @mar4uk I decided to backtrack on this, and instead I made the Agent component behave similarly to the Collector one. It will default certain resource attributes. I hope this makes things easier for users. ATM it defaults the same ones which the Collector defaults, but in the future it might diverge a bit. That should be ok.

docs/sources/flow/reference/components/otelcol.processor.resourcedetection.md
```diff
@@ -0,0 +1,940 @@
...
+If multiple detectors are inserting the same attribute name, the first detector to insert wins. 
+For example if you had `detectors = ["eks", "ec2"]` then `cloud.platform` will be `aws_eks` instead of `ec2`. 
+
+For AWS, the below ordering is recommended:
+  1. [lambda][]
+  1. [elasticbeanstalk][]
+  1. [eks][]
+  1. [ecs][]
+  1. [ec2][]
+<!-- TODO: Why do the otel docs also list a GCP ordering? There is only one GCP detector. -->
```
ptodev, line 88: I need to check this:
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.87.0/processor/resourcedetectionprocessor/README.md#gcp

docs/sources/flow/reference/components/otelcol.processor.resourcedetection.md
```diff
@@ -0,0 +1,940 @@
...
+`tags` can be used to gather tags for the EC2 instance which {{< param "PRODUCT_ROOT_NAME" >}} is running on. 
+Note that in order to fetch EC2 tags, the IAM role assigned to the EC2 instance must have a policy that includes the `ec2:DescribeTags` permission.
+
+The `ec2` block supports the following blocks:
+
+Block                                          | Description                                       | Required
+---------------------------------------------- | ------------------------------------------------- | --------
+[resource_attributes](#ec2--resource_attributes) | Configures which resource attributes to add.      | no
+
+##### ec2 > resource_attributes
```
ptodev, line 163: These `... > resource_attributes` headings have a small weight of `####` so that they don't appear in the ToC. This makes the ToC much cleaner.

docs/sources/flow/reference/components/otelcol.processor.resourcedetection.md
```diff
@@ -0,0 +1,940 @@
...
+
+Block                                          | Description                                       | Required
+---------------------------------------------- | ------------------------------------------------- | --------
+[resource_attributes](#consul--resource_attributes) | Configures which resource attributes to add.      | no
+
+#### consul > resource_attributes
+
+The `resource_attributes` block supports the following blocks:
+
+<!-- TODO: Why are all these azure resource attributes here? -->
```
ptodev, line 396: I need to check this too:
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.87.0/processor/resourcedetectionprocessor/internal/consul/internal/metadata/generated_config.go#L28-L37

docs/sources/flow/reference/components/otelcol.processor.resourcedetection.md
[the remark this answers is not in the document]
ptodev: I updated the PR to hide the `... > resource_attributes` headings in the ToC:
https://github.com/grafana/agent/pull/5764#discussion_r1447975010

component/otelcol/processor/resourcedetection/resourcedetection.go
```diff
@@ -0,0 +1,223 @@
...
+		Exports: otelcol.ConsumerExports{},
+
+		Build: func(opts component.Options, args component.Arguments) (component.Component, error) {
+			fact := resourcedetectionprocessor.NewFactory()
+			return processor.New(opts, fact, args.(Arguments))
+		},
+	})
+}
+
+// Arguments configures the otelcol.processor.probabilistic_sampler component.
```
rlankfo, line 44: I assume this should be `otelcol.processor.resourcedetection` here

rlankfo:
Does this provide support for the env resource detector? e.g. setting resource attributes in `OTEL_RESOURCE_ATTRIBUTES` environment variable?

ptodev:
> Does this provide support for the env resource detector? e.g. setting resource attributes in `OTEL_RESOURCE_ATTRIBUTES` environment variable?

@rlankfo - Yes. It is mentioned in the [docs](https://github.com/grafana/agent/pull/5764/files/4bb42f67714707d5c34b212e54c465417fb3e396#diff-ba78575ddc15f488c5e89aedeefea2e9ba4cfe5b2391a90edc3f972114c957aeR70-R73) and there is a [unit test](https://github.com/grafana/agent/pull/5764/files/4bb42f67714707d5c34b212e54c465417fb3e396#diff-5fa1e129bac5f26513f330cf291e03329447952192d0262f8b21b933cc720603R1457-R1484).

docs/sources/flow/reference/components/otelcol.processor.resourcedetection.md
mar4uk: the resourcedetection readme needs a fix probably. I found that gce and gke were deprecated to use only one gcp detector https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/10347, ordering for gcp is not needed anymore if I understood correctly

component/otelcol/processor/resourcedetection/internal/openshift/config.go
```diff
@@ -0,0 +1,66 @@
...
+func (args Config) Convert() map[string]interface{} {
+	return map[string]interface{}{
+		"address":             args.Address,
+		"token":               args.Token,
+		"tls":                 args.TLSSettings.Convert(),
+		"resource_attributes": args.ResourceAttributes.Convert(),
+	}
+}
+
+// ResourceAttributesConfig provides config for openshift resource attributes.
```
mar4uk, line 51: nit: in all other config files `resourcedetectionprocessor/` prefix was added. Shall it be `resourcedetectionprocessor/openshift` instead of `openshift`. That's super minor

ptodev: Thanks for spotting this! I'll remove the `resourcedetectionprocessor/` because it feels unnecessary.

docs/sources/flow/reference/components/otelcol.processor.resourcedetection.md
```diff
@@ -0,0 +1,939 @@
...
+* `kubernetes_node`
+
+Note that `env` is the only detector which is not configured through a River block.
+The `env` detector reads resource information from the `OTEL_RESOURCE_ATTRIBUTES` environment variable. 
+This is expected to be in the format `<key1>=<value1>,<key2>=<value2>,...`, 
+the details of which are currently pending confirmation in the OpenTelemetry specification.
+
+If a detector other than `env` is needed, it can be customized by using the relevant River block for it.
+For example, the `ec2` detector can be customized using the [ec2][] block.
+If the [ec2][] block is omitted, the defaults specified in the [ec2][] block documentation will be used.
```
clayton-cornell, line 77:
```suggestion
If you omit the [ec2][] block, the defaults specified in the [ec2][] block documentation are used.
```

docs/sources/flow/reference/components/otelcol.processor.resourcedetection.md
```diff
@@ -0,0 +1,939 @@
...
+
+If multiple detectors are inserting the same attribute name, the first detector to insert wins. 
+For example if you had `detectors = ["eks", "ec2"]` then `cloud.platform` will be `aws_eks` instead of `ec2`.
```
clayton-cornell, line 80:
```suggestion
For example, if you had `detectors = ["eks", "ec2"]` then `cloud.platform` will be `aws_eks` instead of `ec2`. 
```

docs/sources/flow/reference/components/otelcol.processor.resourcedetection.md
```diff
@@ -0,0 +1,939 @@
...
+the details of which are currently pending confirmation in the OpenTelemetry specification.
+
+If a detector other than `env` is needed, it can be customized by using the relevant River block for it.
+For example, the `ec2` detector can be customized using the [ec2][] block.
+If the [ec2][] block is omitted, the defaults specified in the [ec2][] block documentation will be used.
+
+If multiple detectors are inserting the same attribute name, the first detector to insert wins. 
+For example if you had `detectors = ["eks", "ec2"]` then `cloud.platform` will be `aws_eks` instead of `ec2`. 
+
+For AWS, the below ordering is recommended:
```
clayton-cornell, line 82:
```suggestion
The following order is recommended for AWS:
```

docs/sources/flow/reference/components/otelcol.processor.resourcedetection.md
```diff
@@ -0,0 +1,939 @@
...
+
+If the instance is part of AWS ParallelCluster and the detector is failing to connect to the metadata server, 
+check the iptable and make sure the chain `PARALLELCLUSTER_IMDS` contains a rule that allows the {{< param "PRODUCT_ROOT_NAME" >}} user to access `169.254.169.254/32`.
+
+[AWS SDK for Go]: https://docs.aws.amazon.com/sdk-for-go/api/aws/ec2metadata/
+[EC2 instance metadata API]: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html
+[AWS cli user guide]: https://github.com/awsdocs/aws-cli-user-guide/blob/a2393582590b64bd2a1d9978af15b350e1f9eb8e/doc_source/cli-configure-proxy.md#using-a-proxy-on-amazon-ec2-instances
+
+`tags` can be used to gather tags for the EC2 instance which {{< param "PRODUCT_ROOT_NAME" >}} is running on. 
+Note that in order to fetch EC2 tags, the IAM role assigned to the EC2 instance must have a policy that includes the `ec2:DescribeTags` permission.
```
clayton-cornell, line 154:
```suggestion
To fetch EC2 tags, the IAM role assigned to the EC2 instance must have a policy that includes the `ec2:DescribeTags` permission.
```

We should avoid using "note that". If it is actually a note, then we should use the note admonition shortcode.

The text "in order to" can almost always be simplified to "to".

docs/sources/flow/reference/components/otelcol.processor.resourcedetection.md
```diff
@@ -0,0 +1,939 @@
...
+* `cloud.platform`: `"gcp_kubernetes_engine"`
+* `cloud.account.id`: project id
+* `cloud.region`: only for regional GKE clusters; e.g. `"us-central1"`
+* `cloud.availability_zone`: only for zonal GKE clusters; e.g. `"us-central1-c"`
+* `k8s.cluster.name`
+* `host.id`: instance id
+* `host.name`: instance name; only when workload identity is disabled
+
+One known issue is when GKE workload identity is enabled, the GCE metadata endpoints won't be available, 
+thus the GKE resource detector won't be able to determine `host.name`.
```
clayton-cornell, line 496:
```suggestion
One known issue happens when GKE workload identity is enabled. The GCE metadata endpoints won't be available, 
and the GKE resource detector won't be able to determine `host.name`. 
```

docs/sources/flow/reference/components/otelcol.processor.resourcedetection.md
```diff
@@ -0,0 +1,939 @@
...
+* `cloud.account.id`: project id
+* `cloud.region`: only for regional GKE clusters; e.g. `"us-central1"`
+* `cloud.availability_zone`: only for zonal GKE clusters; e.g. `"us-central1-c"`
+* `k8s.cluster.name`
+* `host.id`: instance id
+* `host.name`: instance name; only when workload identity is disabled
+
+One known issue is when GKE workload identity is enabled, the GCE metadata endpoints won't be available, 
+thus the GKE resource detector won't be able to determine `host.name`. 
+In that case, users are encouraged to set `host.name` from either:
```
clayton-cornell, line 497:
```suggestion
If this happens, you can set `host.name` from one of the following resources:
```

docs/sources/flow/reference/components/otelcol.processor.resourcedetection.md
```diff
@@ -0,0 +1,939 @@
...
+- `node.name` through the downward API with the `env` detector.
+- Obtaining the Kubernetes node name from the Kubernetes API (with `k8s.io/client-go`).
```
clayton-cornell, line 499:
```suggestion
- Get the `node.name` through the downward API with the `env` detector.
- Get the Kubernetes node name from the Kubernetes API (with `k8s.io/client-go`).
```

What is a downward API.

docs/sources/flow/reference/components/otelcol.processor.resourcedetection.md
```diff
@@ -0,0 +1,939 @@
...
+[gcp.cloud_run.job.task_index][res-attr-cfg]        | Toggles the `gcp.cloud_run.job.task_index` resource attribute. <br> Sets `enabled` to `true` by default.        | no
+[gcp.gce.instance.hostname][res-attr-cfg]        | Toggles the `gcp.gce.instance.hostname` resource attribute. <br> Sets `enabled` to `false` by default.        | no
+[gcp.gce.instance.name][res-attr-cfg]        | Toggles the `gcp.gce.instance.name` resource attribute. <br> Sets `enabled` to `false` by default.        | no
+[host.id][res-attr-cfg]        | Toggles the `host.id` resource attribute. <br> Sets `enabled` to `true` by default.        | no
+[host.name][res-attr-cfg]        | Toggles the `host.name` resource attribute. <br> Sets `enabled` to `true` by default.        | no
+[host.type][res-attr-cfg]        | Toggles the `host.type` resource attribute. <br> Sets `enabled` to `true` by default.        | no
+[k8s.cluster.name][res-attr-cfg]        | Toggles the `k8s.cluster.name` resource attribute. <br> Sets `enabled` to `true` by default.        | no
+
+#### Google Compute Engine (GCE) Metadata
```
clayton-cornell, line 471: Are all these headings supposed to be initial caps? Google Compute Engine should be (it defines the GCE acronym) but is `Metadata` also supposed to be initial caps?

docs/sources/flow/reference/components/otelcol.processor.resourcedetection.md
```diff
@@ -0,0 +1,939 @@
...
+  name: grafana-agent
+rules:
+- apiGroups: ["config.openshift.io"]
+  resources: ["infrastructures", "infrastructures/status"]
+  verbs: ["get", "watch", "list"]
+```
+
+By default, the API address is determined from the environment variables `KUBERNETES_SERVICE_HOST`, 
+`KUBERNETES_SERVICE_PORT` and the service token is read from `/var/run/secrets/kubernetes.io/serviceaccount/token`. 
+If TLS is not explicit disabled and no `ca_file` is configured, `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` is used.
```
clayton-cornell, line 663:
```suggestion
If TLS is not explicitly disabled and no `ca_file` is configured, `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` is used. 
```

docs/sources/flow/reference/components/otelcol.processor.resourcedetection.md
```diff
@@ -0,0 +1,939 @@
...
+rules:
+- apiGroups: ["config.openshift.io"]
+  resources: ["infrastructures", "infrastructures/status"]
+  verbs: ["get", "watch", "list"]
+```
+
+By default, the API address is determined from the environment variables `KUBERNETES_SERVICE_HOST`, 
+`KUBERNETES_SERVICE_PORT` and the service token is read from `/var/run/secrets/kubernetes.io/serviceaccount/token`. 
+If TLS is not explicit disabled and no `ca_file` is configured, `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` is used. 
+The determination of the API address, `ca_file` and the service token is skipped if they are set in the configuration.
```
clayton-cornell, line 664:
```suggestion
The determination of the API address, `ca_file`, and the service token is skipped if they are set in the configuration.
```

docs/sources/flow/reference/components/otelcol.processor.resourcedetection.md
```diff
@@ -0,0 +1,939 @@
...
+kind: ClusterRole
+metadata:
+  name: grafana-agent
+rules:
+  - apiGroups: [""]
+    resources: ["nodes"]
+    verbs: ["get", "list"]
+```
+
+`auth_type` can be set either of the following:
```
clayton-cornell, line 715:
```suggestion
`auth_type` can be set to one of the following:
```

docs/sources/flow/reference/components/otelcol.processor.resourcedetection.md
```diff
@@ -0,0 +1,939 @@
...
+    logs    = [otelcol.exporter.otlp.default.input]
+    metrics = [otelcol.exporter.otlp.default.input]
+    traces  = [otelcol.exporter.otlp.default.input]
+  }
+}
+```
+
+### env and ec2
+
+There is no need to put in a `ec2 {}` River block.
```
clayton-cornell, line 803:
```suggestion
There is no need to put in an `ec2 {}` River block.
```

docs/sources/flow/reference/components/otelcol.processor.resourcedetection.md
```diff
@@ -0,0 +1,939 @@
...
+
+  output {
+    logs    = [otelcol.exporter.otlp.default.input]
+    metrics = [otelcol.exporter.otlp.default.input]
+    traces  = [otelcol.exporter.otlp.default.input]
+  }
+}
+```
+
+You need to also add this to your workload:
```
clayton-cornell, line 882:
```suggestion
You need to add this to your workload:
```

docs/sources/flow/reference/components/otelcol.processor.resourcedetection.md
```diff
@@ -0,0 +1,939 @@
...
```
clayton-cornell, line 911:
```suggestion
You need to add this to your workload:
```

docs/sources/flow/reference/components/otelcol.processor.resourcedetection.md
```diff
@@ -0,0 +1,939 @@
...
+
+- Components that export [OpenTelemetry `otelcol.Consumer`]({{< relref "../compatibility/#opentelemetry-otelcolconsumer-exporters" >}})
+
+`otelcol.processor.resourcedetection` has exports that can be consumed by the following components:
+
+- Components that consume [OpenTelemetry `otelcol.Consumer`]({{< relref "../compatibility/#opentelemetry-otelcolconsumer-consumers" >}})
+
+{{% admonition type="note" %}}
+
+Connecting some components may not be sensible or components may require further configuration to make the
```
clayton-cornell, line 934:
```suggestion
Connecting some components may not be sensible, or components may require further configuration to make the 
```

clayton-cornell (commented):
Some doc suggestions and questions

component/otelcol/processor/resourcedetection/internal/resource_attribute_config/resource_attribute_config.go
```diff
@@ -0,0 +1,12 @@
...
+type ResourceAttributeConfig struct {
+	Enabled bool `river:"enabled,attr"`
+}
```
tpaschalis, line 6: I don't mean to block this PR, since this is already a big effort, but I'm curious, did we consider other possibilities to make this easier to configure and/or maintain in the future?

For example, one possibility was to consider making this a bool type alias, or an enum-like field

So instead of
```
cloud.account.id { enabled = true }
```

We either get
```
cloud.account.id = true
// or
enabled = ["cloud.account.id", "foo", "bar"]
```

Also, AFAICT in the upstream OTEL collector they use some kind of [YAML metadata](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/resourcedetectionprocessor/internal/aws/ec2/metadata.yaml) to generate all these config options, maybe we could use a similar approach?

tpaschalis (approved):
LGTM % some nits; I don't mean to block this PR as it's been a large effort, but I have a potential concern about the long-term maintainability of it and whether we could re-think how we configure these blocks and keep track of new options.

GIven that would be a big change in of itself, I'm fine with re-considering in a future PR.

docs/sources/flow/reference/components/otelcol.processor.resourcedetection.md
```diff
@@ -0,0 +1,939 @@
...
+}
+```
+
+## Arguments
+
+`otelcol.processor.resourcedetection` supports the following arguments:
+
+Name        | Type           | Description                                                                            | Default   | Required
+----------- | -------------- | -------------------------------------------------------------------------------------- |---------- | --------
+`detectors` | `list(string)` | An ordered list of named detectors which should be ran to detect resource information. | `["env"]` | no
```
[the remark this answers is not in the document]
ptodev: The detectors are ran by `otelcol.processor.resourcedetection`. For example, `ecs` and `consul` are detectors. I think `should be ran` is ok because it indicates that if a detector is no listed in `detectors`, it will not be ran at all.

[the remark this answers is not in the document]
clayton-cornell: "ran" is past tense, so it's probably the wrong word here. "should" implies it's optional.. you should do it.. but you don't have to.

`An ordered list of named detectors used to detect resource information.`  This leaves it open to whichever process "owns" and runs the detectors.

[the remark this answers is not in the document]
ptodev: Ok, thank you, I'll change it to that!

docs/sources/flow/reference/components/otelcol.processor.resourcedetection.md
ptodev: There is no reason for `Metadata` that I can think of, so I will lower it.

docs/sources/flow/reference/components/otelcol.processor.resourcedetection.md
ptodev: It is a way for a pod to get more information about the environment it is running in:
https://kubernetes.io/docs/concepts/workloads/pods/downward-api/

I'll change "downward API" to be a link to the k8s docs.

clayton-cornell (approved):
Docs look good to me

component/otelcol/processor/resourcedetection/resourcedetection.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+func (args Arguments) ConvertDetectors() []string {
+	if args.Detectors == nil {
+		return nil
+	}
+
+	res := make([]string, 0, len(args.Detectors))
+	for _, detector := range args.Detectors {
+		//TODO(ptodev): Check if the detector name is valid
```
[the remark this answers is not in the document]
ptodev: Thanks for spotting this! I think the Collector code would throw an error, but it would be more correct for River to do it in the Validate() function. I updated the code to do just that, and added tests.

docs/sources/flow/reference/components/otelcol.processor.resourcedetection.md
ptodev: I raised a PR to fix this upstream:
https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/30727

mar4uk:
🎉

Issue #2051, opened 2022-08-19
Officically Document proxy_url support


metzgertie:
The documentation for the agent is currently out of date and does not mention the support for proxy_url. You'll need to search the supported prometheus drop-in options or the link pointing at it.

https://community.grafana.com/t/grafana-agent-behind-network-firewall-cant-specify-proxy/65044

rfratto:
Hi, thanks for the report. 

Unfortunately it's unclear where proxy_url is supported. Our [docs for metrics_config](https://grafana.com/docs/agent/latest/configuration/metrics-config/) link out to Prometheus for remote_write, which is where the proxy_url field is defined. 

This is something we've struggled with historically and are trying to find a good middle ground for: if we're directly exposing Prometheus fields to users, duplicating information is risky because you can become out of date. On the other hand, linking out to information can make it become unclear what is actually supported. 

We'll spend some time thinking about what to do about this and improve our documentation.

metzgertie:
@rfratto Thanks for the answer. I just noticed it because I suddendly needed to setup a agent in a third-party enviroment with a custom proxy. 
I'd like to see a clarified stance on proxy support in the agent.

PR #5767, opened 2023-11-14
Add HTTPClientConfig to discovery.ec2 and add a simple test

### LINKED ISSUE #5766: Allow customising HTTPClientConfig in discovery.ec2 component
### Request

Some enterprise environments require using a proxy in order to connect to the AWS API. Without being able to customise HTTPClientConfig the discovery.ec2 component cannot be used as the API is inaccessible.

Example configuration:

```river
discovery.ec2 "ec2" {
  region = "ap-southeast-2"

  proxy_url = "http://my-great-proxy.local.home/"
}
```

A potential alternative is using environment variables, but this currently does not work either (and can have potential other negative impacts anyway).

```bash
export http_proxy=http://my-great-proxy.local.home/
export https_proxy=http://my-great-proxy.local.home/
```

### Use case

The key field to be customised is `proxy_url` so that the `discovery.ec2` component can be used when the AWS API is only accessible via a corporate proxy.

cmbrad:
#### PR Description

Updates `discovery.ec2` to allow specifying HTTPClientConfig fields such as `proxy_url`. This allows `discovery.ec2` to work behind a corporate proxy.

Fixes #5766

#### Notes to the Reviewer

There are currently no tests - I've added a very basic test. It's not entirely clear to me how/where to update the docs. #2051 seems to relate to this.

It appears various discovery components either implement this functionality by injecting the whole `HTTPClientConfig` or just specific fields. I've gone with the simple implementation here.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated


spartan0x117:
:wave: Thanks for the contribution! The docs changes would go in `docs/sources/flow/reference/components/discovery.ec2.md`, if you could add those.

You would probably want to add a new block section with:
```
{{< docs/shared lookup="flow/reference/components/http-client-config-block.md" source="agent" version="<AGENT VERSION>" >}}
```
And you don't need to rewrite the contents :smiley:

cmbrad:
Thanks for the pointer! By the look of the existing codebase (ie `docs/sources/flow/reference/components/remote.http.md`) that shared block works well when the options are in a separate `client` block - but currently I think this implementation just squashes arguments rather than using a separate block (to match the majority of other discovery component implementations).

Instead I chose to match `docs/sources/flow/reference/components/discovery.consul.md` (and other similar ones) which put a few select arguments directly in-line. Let me know what you think

spartan0x117 (approved):
LGTM! Pending review from @clayton-cornell I can get this merged

clayton-cornell:
@spartan0x117 Looks OK to me. I haven't looked into the conflicts.

PR #5771, opened 2023-11-14
ci: use Apple silicon macOS runners

rfratto:
Use the new Apple silicon macOS runner when running macOS tests.

While these new runners are twice as expensive as the normal macOS runners, it is expected to be cost-neutral thanks to the speed increase.

For more information, see GitHub's announcement about the new runners: https://github.blog/2023-10-02-introducing-the-new-apple-silicon-powered-m1-macos-larger-runner-for-github-actions/

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,7 +20,7 @@ jobs:
     - name: Set up Go 1.21
       uses: actions/setup-go@v4
       with:
-        go-version: "1.21.0"
+        go-version: "1.21"
         cache: true
     - name: Test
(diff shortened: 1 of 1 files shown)


rfratto:
The [test run](https://github.com/grafana/agent/actions/runs/6868828933/job/18680306673?pr=5771) for this PR took 12m46s, compared to a [previous test run](https://github.com/grafana/agent/actions/runs/6868444034/job/18678991777) that took 23m5s. It does seem like the assumption that this is roughly twice as fast as the old runner holds true.

mattdurham (approved):
Woot Woot!

PR #5772, opened 2023-11-14
For Loki Sources position files: Use a best effort atomic rename

### LINKED ISSUE #4916: Agend Flow crash after client reboot
### What's wrong?

after rebooting windows 10 clients the grafana agent flow service wont start anymore

when deleting `C:\ProgramData\Grafana Agent Flow\data\module.file.logs\loki.source.windowsevent.application` the service can start again

### Steps to reproduce

using below config on windows and reboot the client

### System information

WIN10 x64

### Software version

v0.34 &  v0.35.4

### Configuration

```text
/********************************************
 * LOKI SCRAPE
 ********************************************/
loki.source.windowsevent "application"  {
    eventlog_name = "Application"
    forward_to = [loki.process.extract_labels.receiver]
}

/********************************************
 * LOKI REGEX LABELS
 ********************************************/
loki.process "extract_labels" {
  forward_to = [loki.relabel.hostname.receiver]
  stage.regex {
    expression = "\"event_id\":(?P<event_id>\\d+),\"level\":(?P<severity>\\d+)"
  }
  stage.labels {
    values = {
      "level"  = "severity",
      "event_id" = "event_id",
    }
  }
}

loki.relabel "hostname" {

  rule {
    action       = "replace"
    target_label = "computer"
    replacement  = constants.hostname
  }
}

/********************************************
 * LOKI WRITE
 ********************************************/
loki.write "default" {
    endpoint {
        url ="https://redacted"
        tenant_id = "redacted"
    }
}
```

### Logs

```text
runtime error: index out of range [0] with length 0

goroutine 207 [running]:
github.com/grafana/loki/clients/pkg/promtail/targets/windows/win_eventlog.formatEventString(0x1, 0xc001a7bd10?, 0x200?)
	/go/pkg/mod/github.com/grafana/loki@v1.6.2-0.20230414223651-220cbdd4f172/clients/pkg/promtail/targets/windows/win_eventlog/win_eventlog.go:532 +0x1db
github.com/grafana/loki/clients/pkg/promtail/targets/windows/win_eventlog.(*EventFetcher).renderEvent(_, 

level=error msg="service exited with error" err="exit status 2" exit_code=2
```

mattdurham:
#### PR Description

This switches the bookmark and positions file in loki.sources to use a more resilient method. This is primarily an issue on windows where a reboot on writing to the file causes an issue. The default os.rename in windows is not reliable. The atomic package uses a much more reliable method though there are issues where it doesnt work. These are generally limited to network file systems and non-ntfs. This should cover 99.9% of our cases without problem.

Most of the changes are fork lifting from promtail.

Closes #4916

#### PR Checklist

- [x] CHANGELOG.md updated


component/loki/source/windowsevent/target.go
```diff
@@ -0,0 +1,230 @@
+//go:build windows
+// +build windows
+
+// This code is copied from Promtail with minor changes.
```
thampiotr, line 4: Can we add a commit sha of what promtail it's copied from? Makes it easier to refresh with changes in the future.

component/loki/source/windowsevent/bookmark.go
```diff
@@ -0,0 +1,83 @@
...
+	bm, err := win_eventlog.CreateBookmark(fileString)
+	if err != nil {
+		return nil, err
+	}
+	return &bookMark{
+		handle: bm,
+		path:   path,
+		isNew:  fileString == "",
+		buf:    buf,
+	}, nil
```
thampiotr, line 73: In this fragment of code few things can still go wrong and possibly the agent will be crashlooping until someone manually repairs the bookmark file or, more likely, deletes it. 

Should we instead have on a high-level something like this?

```
func newBookmark() {
  if not fileExists {
    createNewOne()
  } else {
    err := useExistingFile()
    if err != nil {
      renameCorruptedFile()
      createNewOne()
    }
  }
}
```

The idea being that if we detect a corrupted file, we attempt to recover.

thampiotr (commented):
Couple of questions. Would be nice if we could test this somehow too, although I realise it's on Windows so would be limited utility.

component/loki/source/windowsevent/target.go
mattdurham: Sure, I can add that in.

component/loki/source/windowsevent/bookmark.go
mattdurham: If we error on the create bookmark then we could call it again with a blank string to force a new bookmark.

mattdurham:
Unit tests are hard since it would need to stop during the write to disk.

thampiotr (approved):

PR #5773, opened 2023-11-14
helm chart: align k8s version label with Values

tristanburgess:
#### PR Description
Use helm chart template functions to supply `app.kubernetes.io/version` with value as configured by any possible override for the Agent image tag version to the Helm chart Values.

#### PR Checklist

- [x] CHANGELOG.md updated


operations/helm/charts/grafana-agent/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: grafana-agent
 description: 'Grafana Agent'
 type: application
-version: 0.27.2
```
tpaschalis, line 5: Bumping this will automatically publish a new chart version. IMHO it's not required straight away, let's leave it as-is.

Otherwise, we'll have to update the CHANGELOG headers to mirror that.

operations/helm/charts/grafana-agent/templates/_helpers.tpl
```diff
@@ -55,9 +55,7 @@ helm.sh/chart: {{ include "grafana-agent.chart" . }}
 app.kubernetes.io/version: "vX.Y.Z"
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 {{- else }}
-{{- if .Chart.AppVersion }}
-app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
-{{- end }}
+app.kubernetes.io/version: {{ substr 1 -1 (include "grafana-agent.imageId" .) }}
```
tpaschalis, line 58: What's the substr call here for, getting rid of quotes? Do we need it?

tristanburgess: it's to get rid of the delimiter character that constitutes the first character of the output of the `"grafana-agent.imageId"` template function. For example, for image tags, instead of `v0.37.4`, `"grafana-agent.imageId"` alone would output `:v0.37.4`. For digests, the output would be `@sha256:<digest>` instead of `sha256:<digest>`.

tpaschalis: Makes sense!

tpaschalis (approved):
LGTM. Let's re-run `make generate-helm-docs` to fix the README and we're good to go.

PR #5774, opened 2023-11-14
Add metadata to User-Agent header

captncraig:
- Refactors User-Agent code into a shared package, instead of the same copy/paste snippet everywhere.
- Add additional metadata in parenthesis at end of user agent, semicolon delimeted.
- Current metadata added:
  - agent mode (`flow` or `static`)
  - GOOS
  - `operator` if running in operator generated pod.
- Operator adds `AGENT_OPERATOR=1` to agent pods it creates to facilitate the above.

Example User Agents:

- `GrafanaAgent/v1.2.3 (static; linux; operator)`
- `GrafanaAgent/v1.2.3 (flow; windows)`
- `GrafanaAgent/v1.2.3 (static; linux)`

### Why?

This will allow database operators (mimir,loki,pyroscope, etc.) to better understand agent traffic patterns sending them data.


pkg/useragent/useragent.go
```diff
@@ -0,0 +1,61 @@
+// package useragent provides a consistent way to get a user agent for outbound http requests from Grafana Agent.
+// The default User-Agent is `GrafanaAgent/$VERSION($MODE)`
+// Where version is the build version of the agent and MODE is one of "static" or "flow".
+package useragent
```
rfratto, line 4: Can we put this in `internal/useragent` instead of `pkg/useragent`? Since we're heading towards 1.0 where we're unlikely to have a large public API (or any public API at all), I'd like to get in the habit of only exposing what absolutely needs to be made public.

pkg/useragent/useragent.go
```diff
@@ -0,0 +1,61 @@
...
+	metadata := []string{}
+	if mode := getRunMode(); mode != "" {
+		metadata = append(metadata, mode)
+	}
+	metadata = append(metadata, goos)
+	if op := getOperator(); op != "" {
+		metadata = append(metadata, op)
+	}
+	if len(metadata) > 0 {
+		parenthesis = fmt.Sprintf(" (%s)", strings.Join(metadata, ";"))
```
rfratto, line 34: What do you think about using `strings.Join(metadata, "; ")` here (with the space after the semicolon)? That would help make the metadata slightly more human-readable:

```GrafanaAgent/v1.2.3 (static; linux; operator)```

vs 

```GrafanaAgent/v1.2.3 (static;linux;operator)```

File: pkg/useragent/useragent.go [hunk 3]
```diff
@@ -0,0 +1,61 @@
...
+
+const (
+	operatorEnv = "AGENT_OPERATOR"
+	modeEnv     = "AGENT_MODE"
+)
+
+// settable by tests
+var goos = runtime.GOOS
+
+func UserAgent() string {
```
rfratto, line 23: Calling this `Get()` might be less repetitive so callers use `useragent.Get()` instead of `useragent.UserAgent()`

rfratto (approved):
LGTM! Just a few small comments about style

rfratto (approved):

PR #5780, opened 2023-11-15
Remove NeedsServices from components

### LINKED ISSUE #5777: Remove requirement that components must depend on consumed services
### Background

Currently, all components that use a service must explicitly declare a dependency on that service. This ensures that the service is guaranteed to evaluate its config (if any) before the dependent component is evaluated. 

Because module loader components create sub-controllers, module loader components must always explicitly depend on all services. 

This requirement is very error-prone, and we've had several instances where we've forgotten to update the list of dependencies for module loaders when building a new service, preventing some components in modules from ever loading properly. 

This requirement is also not necessary; there is no service today whose exposed Data depends on the configuration of the service. 

### Proposal

I propose three changes:

* Add a soft requirement that the data of a service must not depend on its runtime configuration. 
* Remove the `NeedsServices` field from `component.Registration` and allow components to access any service in the controller. 
* Always propagate all services down to module controllers. 

Services may still define a dependency on other services; this is not being changed at this time since we haven't noticed any pain points there yet.

rfratto:
This PR removes the NeedsServices field from component registration. See #5777 for context. 

I've split the PR across several commits to make it more readable. 

Closes #5777.


clayton-cornell (commented):
Doc changes are small - looks good as-is

wildum (approved):
LGTM

thampiotr (approved):

PR #5788, opened 2023-11-15
Agent Management: Introduce support for template variables

jcreixell:
#### PR Description

  - This change allows managing template variables for remote configuration on a per-agent basis.
  - Both base configurations and snippets can be interpreted as templates and evaluated at load time with the provided template variables.
  - Templates must follow go's `text/template` syntax.
  - This greatly increases the flexibility and reusability of snippets.
  - Template evaluation has been tested in different scenarios and seems really robust. If the variables defined in the template cannot be resolved (even nested ones), and empty string is rendered instead.
  - Note: templates are only evaluated when the `template_variables` field within the `agent_metadata` remote config field is non-empty.
  - Note: this feature only applies to static mode.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated


marctc (commented):
do we also want to explain this behavior in the docs?

marctc (approved):

tpaschalis (approved):
LGTM, looks pretty straightforward!

clayton-cornell (approved):

PR #5798, opened 2023-11-17
fix: partial docker syncTargets

### LINKED ISSUE #5797: loki.source.docker: positions removed on transient docker error
### What's wrong?

When a transient docker error occurs the position will be removed from the positions file leading to the agent re-ingesting the docker logs from the beginning.

According to [this comment](https://github.com/grafana/agent/blob/55b34460859a1b64f5e698aeb5abd045bf39b23d/component/loki/source/docker/runner.go#L156) the position is removed from the file if it is missing from the "new targets".

This function is called from [here](https://github.com/grafana/agent/blob/950870ad82635c87f102ec0ae2aac0097d965d77/component/loki/source/docker/docker.go#L247) where the targets are being created based on the target input - the targets array is filled within the loop where we call syncTargets as well. This obviously causes the targets in syncTargets to be incomplete and therefore cause the targets to be deregistered - and restarts the entire log ingestion as this automatically resets the position file.

I would expect that [this line](https://github.com/grafana/agent/blob/950870ad82635c87f102ec0ae2aac0097d965d77/component/loki/source/docker/docker.go#L247) should instead be called after the for loop - when the targets slice is fully populated.

### Steps to reproduce

Unclear how to reproduce.  Also not entirely sure why this was never reported as a bug as I would expect that anybody using the docker source would have to experience this issue of regular restarts of the docker log ingestion.

### System information

_No response_

### Software version

at leat from v0.37.1

### Configuration

_No response_

### Logs

```text
ts=2023-11-16T15:22:16.382180877Z level=error msg="could not set up a wait request to the Docker client" target=b50f59e857184efafc04d640a74b10ee26a854e311c39a80c0fee6baa40ee251 component=loki.source.docker.default error=""
ts=2023-11-16T15:22:16.382302835Z level=warn msg="could not transfer logs" component=loki.source.docker.default target=docker/b50f59e857184efafc04d640a74b10ee26a854e311c39a80c0fee6baa40ee251 written=0 container=b50f59e857184efafc04d640a74b10ee26a854e311c39a80c0fee6baa40ee251 err="context canceled"
ts=2023-11-16T15:22:16.382410919Z level=info msg="removing entry from positions file" component=loki.source.docker.default path=cursor-b50f59e857184efafc04d640a74b10ee26a854e311c39a80c0fee6baa40ee251 labels="{__address__=\"172.17.0.2\", __meta_docker_container_id=\"b50f59e857184efafc04d640a74b10ee26a854e311c39a80c0fee6baa40ee251\", __meta_filepath=\"/etc/agent/loki.json\"}"
ts=2023-11-16T15:22:16.382527627Z level=info msg="finished node evaluation" controller_id="" node_id=loki.source.docker.default duration=712.333µs
```

FerdinandvHagen:
syncTargets for docker is called within a loop which leads to an incomplete targets slice to be sent to the manager.

Moving the syncTargets after the loop ensures that always all targets are correctly registered.

Fixes #5797 

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


rfratto (commented):
LGTM! 

Can you add this as a bug fix entry to CHANGELOG.md before we merge?

tpaschalis (approved):
LGTM! FYI @FerdinandvHagen I've rebased with main to avoid conflicts with the changelog and added a new bugfix entry.

Thanks for your contribution!

PR #5804, opened 2023-11-17
Add graceful shutdown to `loki.write` draining the WAL

thepalbi:
#### PR Description

This PR is a follow up after some discussions in https://github.com/grafana/agent/pull/5770.

The idea of this PR is to allow `loki.write` to do a graceful shutdown when the component is stopped. This graceful shutdown will attempt to drain the WAL, making the watcher consume all of the segments it can, before a timeout occurs. Note that two timeouts will managet the WAL enabled `loki.write` component: the WAL drain timeout, and the queue client drain timeout.

Also, the Watcher is refactored a bit to simlify the "possible states" in which this operates.

Related to https://github.com/grafana/cloud-onboarding/issues/5407

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


docs/sources/flow/reference/components/loki.write.md
```diff
@@ -162,10 +162,11 @@ The following arguments are supported:
...
-`enabled`                 | `bool`     | Whether to enable the WAL.                                                                                   | false     | no
+`enabled`                 | `bool`     | Whether to enable the WAL.                                                                                         | false     | no
 `max_segment_age`             | `duration` | Maximum time a WAL segment should be allowed to live. Segments older than this setting will be eventually deleted. | `"1h"`    | no
 `min_read_frequency`          | `duration` | Minimum backoff time in the backup read mechanism.                                                                 | `"250ms"` | no
 `max_read_frequency`          | `duration` | Maximum backoff time in the backup read mechanism.                                                                 | `"1s"`    | no
+`drain_timeout`          | `duration` | Maximum time the WAL drain procedure can take, before being forcefully stopped.                                    | `"30s"`   | no
```
clayton-cornell, line 169:
```suggestion
`drain_timeout`          | `duration` | Maximum time the WAL drain procedure can take before being forcefully stopped.                                    | `"30s"`   | no
```

clayton-cornell (commented):
Doc change is :+1:

component/common/loki/wal/config.go
```diff
@@ -12,6 +12,7 @@ const (
 var DefaultWatchConfig = WatchConfig{
 	MinReadFrequency: time.Millisecond * 250,
 	MaxReadFrequency: time.Second,
+	DrainTimeout:     time.Second * 30,
```
mattdurham, line 15: Is this based on the default grace termination timer?

component/common/loki/wal/watcher.go
```diff
@@ -331,9 +349,24 @@ func (w *Watcher) decodeAndDispatch(b []byte, segmentNum int) (bool, error) {
...
 }
 
-func (w *Watcher) Stop() {
-	// first close the quit channel to order main mainLoop routine to stop
-	close(w.quit)
+// Stop stops the Watcher, draining the WAL until the end of the last segment if drain is true. Since the writer of the WAL
+// is expected to have stopped before the Watcher, no further writes are expected, so segments can be safely consumed.
+//
+// Note if drain is enabled, the caller routine of Stop will block executing the drain procedure.
+func (w *Watcher) Stop(drain bool) {
```
mattdurham, line 356: Part of me think this is combining two state transitions in one call which might be confusing?

component/common/loki/wal/watcher.go
```diff
@@ -331,9 +349,24 @@ func (w *Watcher) decodeAndDispatch(b []byte, segmentNum int) (bool, error) {
...
+	if drain {
+		level.Info(w.logger).Log("msg", "Draining Watcher")
+		w.state.Transition(internal.StateDraining)
+		// wait for drain timeout, or stopping state, in case the Watcher does the transition itself promptly
+		select {
+		case <-time.NewTimer(w.drainTimeout).C:
```
mattdurham, line 362: Does this mean each wal watcher gets the grace period which could become quite long? IE if you had 4 it would be 2 minutes.

component/loki/write/write_test.go
```diff
@@ -90,19 +91,22 @@ func TestUnmarshallWalAttrributes(t *testing.T) {
 				MaxSegmentAge:    wal.DefaultMaxSegmentAge,
 				MinReadFrequency: wal.DefaultWatchConfig.MinReadFrequency,
 				MaxReadFrequency: wal.DefaultWatchConfig.MaxReadFrequency,
+				DrainTimeout:     wal.DefaultWatchConfig.DrainTimeout,
```
ramcharan-tech, line 94: Can we have one more config , For Ex: MaxSegmentSize which will deletes the logs when it touches the disk space size. which will be useful in some situations where the following error throws
msg=failed to write entry component=loki.write.remote err=write /tmp/agent/loki.write.remote/wal/00000569: no space left on device.

component/common/loki/wal/watcher.go
thepalbi: Mmm actually you right, it's almost identical to split apart into a `Stop` and `Drain` methods, let me try that

component/common/loki/wal/config.go
thepalbi: So when stopping a WAL enabled `loki.write`, there's two graceful terminations occurring:
1. first the watchers, which assuming we have a single endpoint config, it will be `Watcher.DrainTimeout`
2. then the clients, which assuming we have a single endpoint config, it will be `QueueConfig.DrainTimeout`

That said, the worst total wait is `Watcher.DrainTimeout + QueueConfig.DrainTimeout`. Since the default graceful termination [is 30 seconds](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/) as you mentioned, let's default to 15 s each so the total sum is 30s.

component/common/loki/wal/watcher.go
thepalbi: Ok splitted apart into `Stop` and `Drain`

component/common/loki/wal/watcher.go
thepalbi: Mmm as for this, in theory and how is implemented yes. All watchers are stopped sequentially, and all clients after. Which taking into account what I mentioned [here](https://github.com/grafana/agent/pull/5804#discussion_r1403440247), could lead to `N * (Watcher.DrainTimeout + QueueConfig.DrainTimeout)`, which if one is using many `EndpointConfigs` could be big. 

Technically, the only relation on stopping is that the `(watcher, client)` pairs are stopped in order. wdyt of changing a bit the structure for keeping them in the manager, and doing something like:

1. watchers and client when creating are kept as tuples, like stoppables  = list(<watcher, client>)

```
def stop:
  var wg sync.WaitGroup
  for watch, client in stoppables:
    wg.Add(1)
    go func() {
      watch.Stop()
      client.Stop()
    }

  wg.Wait()
```

in that case we are only spending `Watcher.DrainTimeout + QueueConfig.DrainTimeout` over all endpoint configs.

component/loki/write/write_test.go
thepalbi: In my opinion, the device that backs the WAL running out of space is a smell that something is happening, between:
- the agent scraping more that it can send
- the WAL volume is too small
- one should be able to make the WAL delete all segments more rapidly with `max_segment_age`, which makes sense since the WAL needs to be thought more of keeping __the last X minutes of data__ rather than __X bytes of data__

Can you share a bit more of your use case, to understand why the WAL volume is reaching it's capacity?

component/common/loki/wal/watcher.go
thepalbi: Ok, changed the manager so that all `watcher -> client` pairs are stopped concurrecntly, reducing the graceful stop time by the number of client config.s

thepalbi:
@mattdurham all comments should be resolved

mattdurham (approved):
LGTM

PR #5832, opened 2023-11-21
Update windows defaults to use upstream defaults except for enabled collectors.

### LINKED ISSUE #5831: prometheus.exporter.windows logical_disk does not match defaults
### What's wrong?

The documentation for prometheus.exporter.windows [claims](https://grafana.com/docs/agent/latest/flow/reference/components/prometheus.exporter.windows/#logical_disk-block) that logical_disk includes all disks by default, but this is not true.

To collect logical_disk stats, I had to explicitly configure the block with the documented defaults. 

### Steps to reproduce

1. Run config with a `prometheus.exporter.windows` component 
2. Observe metrics exported from component
3. Note that logical_disk metrics are missing.

### System information

Windows 11

### Software version

Grafana Agent v0.38.0

### Configuration

```text
prometheus.exporter.windows "default" {
	enabled_collectors = ["cpu","cs","logical_disk","net","os","service","system","time","diskdrive"]
}
```

### Logs

_No response_

mattdurham:
#### PR Description

This aligns the defaults of the windows exporter with the upstream, sadly its basically in two places yaml and river.

Fixes #5831 

#### PR Checklist

- [x] CHANGELOG.md updated
- [ x] Tests updated

### DIFF
--- a/pkg/integrations/windows_exporter/config_windows.go
+++ b/pkg/integrations/windows_exporter/config_windows.go
@@ -1,6 +1,9 @@
 package windows_exporter
 
-import "github.com/prometheus-community/windows_exporter/pkg/collector"
+import (
+	"github.com/prometheus-community/windows_exporter/pkg/collector"
+	col "github.com/prometheus-community/windows_exporter/pkg/collector"
+)
 
 func (c *Config) ToWindowsExporterConfig() collector.Config {
 	cfg := collector.ConfigDefaults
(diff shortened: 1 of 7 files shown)


pkg/integrations/windows_exporter/config_windows.go
```diff
@@ -1,6 +1,9 @@
 package windows_exporter
 
-import "github.com/prometheus-community/windows_exporter/pkg/collector"
+import (
+	"github.com/prometheus-community/windows_exporter/pkg/collector"
+	col "github.com/prometheus-community/windows_exporter/pkg/collector"
```
rfratto, line 5: nit: importing the same package twice with a different alias; can we remove one?

rfratto (commented):
I'm not sure I understand what the problem was and why this fixes it 🤔 Can you explain?

mattdurham:
Before we were [setting](https://github.com/grafana/agent/blob/v0.37.4/pkg/integrations/windows_exporter/config.go) the configs to "". When I made the change that stayed in, but we were depending on the behavior of [registering the kingpin](https://github.com/grafana/agent/blob/v0.37.4/pkg/integrations/windows_exporter/windows_exporter_windows.go#L22) flags to set the default. Those values then got passed to the janky initializer that set the values that were then passed to the individual collectors.

PR #5840, opened 2023-11-23
pyroscope.scrape: change error log level to not swallow errors

korniltsev:
#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/component/pyroscope/scrape/scrape_loop.go
+++ b/component/pyroscope/scrape/scrape_loop.go
@@ -228,7 +228,7 @@ func (t *scrapeLoop) scrape() {
 		}
 	}
 	if err := t.fetchProfile(scrapeCtx, profileType, buf); err != nil {
-		level.Debug(t.logger).Log("msg", "fetch profile failed", "target", t.Labels().String(), "err", err)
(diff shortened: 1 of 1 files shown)


ptodev (approved):

component/pyroscope/scrape/scrape_loop.go
```diff
@@ -228,7 +228,7 @@ func (t *scrapeLoop) scrape() {
 		}
 	}
 	if err := t.fetchProfile(scrapeCtx, profileType, buf); err != nil {
-		level.Debug(t.logger).Log("msg", "fetch profile failed", "target", t.Labels().String(), "err", err)
+		level.Error(t.logger).Log("msg", "fetch profile failed", "target", t.Labels().String(), "err", err)
```
ptodev, line 231: It makes sense to report this, but if the log line is expected to be printed a lot maybe this should be a metric? That way we can have an alert as well.

korniltsev: lets try like this in dev and then revert and make it metric when we have them too much.

Also do you know if scraping metric produces error? It feels they should be aligned and have similar experience.

PR #5842, opened 2023-11-23
New declare block for modules 2.0

wildum:
#### PR Description

This PR is the first part of the implementation of the new modules defined in this [doc](https://docs.google.com/document/d/1_DVeuoY89HcCUuI8At-WU5fx4uApgMOYRVm-5B4GbgQ/edit#heading=h.bjbz01hebx0). It only adds the implementation of the new "declare" block).

For every declare block the loader will build a graph and store it in a tree.
When the loader populates the components, it will check for a match in the tree and if successful it will clone the stored graph and merge it with the main graph. To avoid nodeID collisions and provide scope, a prefix with the nodeID of the declareComponentNode is added to all nodes of the graph.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


rfratto (changes requested):
Thanks for this! I'm really excited about the new modules design and I can't wait to start playing around with it. 

I need some more time to think about the implications of the DAGs being merged together rather than being fully isolated. 

In the meantime, there are three things I've noticed with the current behavior of this PR:

#### Declare components leak on subsequent reloads

This config will leak `prometheus.remote_write` components every subsequent reload of the config file: 

```river
declare "example" {
		endpoint {
		}
	}
}

example "default" { }
```

To reproduce, `curl http://localhost:12345/-/reload` a few times and then take a look at <http://localhost:12345/debug/pprof/goroutines?debug=2>; you'll notice that there's a bunch of goroutines for running the prometheus.remote_write component; there should only be one. 

Additionally, this also means that the component is being reconstructed between reloads. This doesn't align with the behavior of not using declare blocks, where components are only constructed if they were not found in the previous config file, and the underlying components persist between reloads, being evaluated if relevant. 

#### Required arguments are being ignored 

You've mentioned that you removed the check for required arguments because it breaks the new modules; we shouldn't be changing any behavior in old modules, including this. 

Additionally, new modules _should_ check for required arguments; they're marked as required for a reason.

#### Declare components aren't represented in the UI

When I went to the UI, I expected there to be an entry for my declared component node; the intent is that declare components act just like native components (to the end user). However, I only saw the components created inside the declare, and no representation for the declare component itself. Not being able to see the declare component also means that I'm unable to inspect the current arguments or exports. 

I would personally expect to only see the declare component itself, and not any of the components that are part of the template; I would expect, similarly to current modules, that I would have to navigate to the declare component to see the components that are encapsulated inside of it.

wildum:
Thanks for your feedback! I addressed the two first points.
I admit that so far I had completely ignored the UI. I will make changes to replicate the existing behavior. What about the graph display? Old modules behavior only shows the root modules. Now we have the possibility the show more, should we?

rfratto:
> Thanks for your feedback! I addressed the two first points.
I admit that so far I had completely ignored the UI. I will make changes to replicate the existing behavior. What about the graph display? Old modules behavior only shows the root modules. Now we have the possibility the show more, should we?

@wildum I think for now, at least to avoid scope creep, we can start with the following:

* The UI should show instances of declared components, but not any of the components inside of the declare block.

Then, later on, we can expand it find a way to show the components inside of the declare block (likely using the same method that the current modules do)

rfratto:
@wildum I spent some more time thinking about the proposed implementation here and the implications of having the DAGs be merged into a unified DAG. I've also talked to @thampiotr to get a second opinion to make sure I'm not being stubborn :) 

Merging the DAGs together makes me worry about three things in particular: 

* Declare components are meant to be an abstraction to hide details from the user. By merging the nested graphs together and exposing the inner components to the outer graph, the abstraction leaks, and becomes visible in the UI. While we could introduce new logic to hide those inner components in the UI, it would be generally more complex than having fully isolated DAGs. 

* The blast radius of merging graphs ended up being larger than I would like it to be. It seems like we've added a lot of special checks for declare nodes / declare components, and some logic that works for non-merged graphs had to be duplicated (such as reusing component nodes when a new graph is being constructed). 

* Personally, I find that there's additional cognitive overhead with the namespacing/cloning functionality that wouldn't be present if the DAGs were fully isolated. 

Given the branch is called `experiment-with-modules`: your experiment was definitely successful, and you've helped me better understand what we should aim for with our next iteration. In particular: reusing existing modules logic as much as possible, keeping the DAGs separate, and keeping the amount of added code for declare to a minimum. 

I think our next steps should be to chat with @thampiotr and get consensus on a rough design for how declare/import will generally work in a way that meets the criteria above. WDYT?

wildum:
@rfratto thanks for your feedback. I also had a discussion with @thampiotr afterward to discuss the topic. I agree on the fact that this is an important topic to implement in the core of Flow and for that it is worth it to explore different paths. I'd be happy to experiment with different approaches before we decide on which one to commit.

But I don't agree with some of the points you and Piotr brought up. The way I see things is that there is always one graph. The "isolated" graphs are actually glued together via the value_cache which in the end imitates more or less the equivalent of having a double edges connection. I personally find this glue code not intuitive and hard to follow.

I also would like to point out that this is a rough implementation. A few iterations down the line and we could end up with a solution with one flow controller, one graph, old modules gone, no cache, only blocknodes are stored for the declare templates instead of the full component nodes to avoid all the cloning part, and better namespacing. The eval would run seamlessly through the graph and the scope of every node would be built by checking the export values of the nodes connected to its outgoing edges. I actually believe that a lot of cognitive overhead would be removed. For example the eval part would be trivial.

I also don't see why the UI would be a problem. It's actually really intuitive to adapt: fetch only the components (including declare) that have no namespace. When you click on a declare the UI would fetch the nodes corresponding to the namespace of the declare.

That being said, going down this path would be committing for a re-write of core working code. And although now might be the opportunity to take such actions before the project reaches a new level of maturity, I understand the risks and efforts that it would require. We also have to take into account that this solution does not handle the "import" for now and that it might not be a piece of cake to implement, especially if we want to keep the changes to the minimum.

So let's put this one on the side for now and experiment with a different path :)

PR #5843, opened 2023-11-23
Improve visibility of slow component evaluations & failed components in general

thampiotr:
#### PR Description

As a learning from a recent investigation, we want to improve debugging experience when:
* components are slow to evaluate
* config reloads fail

This PR achieves the above goals by:
* Adding a graph that shows which components take longer than 1m to evaluate
* Setting the histogram buckets to a more useful range, so the existing charts become more useful
* Adding an alert for slow component evaluation if there are more than 1 in 10m window for 15m
* Adding an alert for components in unhealthy state if there are unhealthy components for more than 15m

#### Notes to the Reviewer

Examples of how this looks:

image

Healthy cluster:
image

Some unhealthy components:
image

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


rfratto (approved):
Nice, thanks!

PR #5844, opened 2023-11-23
5679 flow mode service discovery deletes jobs with same prefix

### LINKED ISSUE #5679: GrafanaAgent self-monitoring using ServiceMonitor is dropped for a single instance
### What's wrong?

In our setup we implemented GrafanaAgent self-monitoring using ServiceMonitors which are picked up by the prometheus service monitor component (https://grafana.com/docs/agent/latest/flow/reference/components/prometheus.operator.servicemonitors/). Then we noticed the flow mode dashboard shows the cluster isn't converged; it shows 22 instances (out of 23 nodes).

A quick check on the metric being used shows that at a given moment that 23th metric just disappears.

image

We checked the debugging-clustering-issues guideline (https://grafana.com/docs/agent/latest/flow/monitoring/debugging/#debugging-clustering-issues) but that didn't help:
- The grafana-agent service is health with all 23 instances attached to it.
- Nodes are up
- All 23 instances registered with the clustering mode
- We checked all 23 UI's but did not find any scrape config for that missing ip-address (.5.141)
- Logging didn't show anything unusual
- No crashes or anything

We have now added the hostname of the grafana-agent instance so we can at least detect which instance was scraping the endpoint until the issue happened.

Are there any additional metrics or logging categories we can enable to further debug this issue when it's occuring?

### Steps to reproduce

Setup clustering mode and self-monitoring using prometheus.service-monitor component.

### System information

GKE

### Software version

grafana/agent:v0.37.2

### Configuration

_No response_

### Logs

_No response_

Paul424:

Fixes #5679

#### Notes to the Reviewer
Two interfaces had to be inserted to enable mocking the actual discovery-manager and scrape-manager; if this should be done differently please let me know through the PR so i can rework.

#### PR Checklist
- [x] CHANGELOG.md updated
- [x] Tests updated

### DIFF
--- a/component/prometheus/operator/common/crdmanager.go
+++ b/component/prometheus/operator/common/crdmanager.go
@@ -39,6 +39,68 @@ import (
 // Generous timeout period for configuring all informers
 const informerSyncTimeout = 10 * time.Second
 
+type DiscoveryManager interface {
+	Run() error
+	SyncCh() <-chan map[string][]*targetgroup.Group
+	ApplyConfig(cfg map[string]discovery.Configs) error
+}
+
+type discoveryManager struct {
+	discoveryManager *discovery.Manager
+}
+
+func newDiscoveryManager(m *discovery.Manager) *discoveryManager {
+	return &discoveryManager{
+		discoveryManager: m,
+	}
+}
+
+func (m *discoveryManager) Run() error {
+	return m.discoveryManager.Run()
+}
+
+func (m *discoveryManager) SyncCh() <-chan map[string][]*targetgroup.Group {
+	return m.discoveryManager.SyncCh()
+}
+
+func (m *discoveryManager) ApplyConfig(cfg map[string]discovery.Configs) error {
+	return m.discoveryManager.ApplyConfig(cfg)
+}
+
+type ScrapeManager interface {
+	Run(tsets <-chan map[string][]*targetgroup.Group) error
+	Stop()
+	TargetsActive() map[string][]*scrape.Target
+	ApplyConfig(cfg *config.Config) error
+}
+
+type scrapeManager struct {
+	scrapeManager *scrape.Manager
+}
+
+func newScrapeManager(m *scrape.Manager) *scrapeManager {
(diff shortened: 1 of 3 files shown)


component/prometheus/operator/common/crdmanager.go
```diff
@@ -39,15 +39,77 @@ import (
 // Generous timeout period for configuring all informers
 const informerSyncTimeout = 10 * time.Second
 
+type DiscoveryManager interface {
```
captncraig, line 42: Can you put these interfaces in a seperate file. I understand why you added them, and I do believe they are still useful, but this file is large enough as it is.

captncraig (approved):

captncraig:
Oh, I think I see one complication here. For podmonitors and servicemonitors, the job name is `podMonitor/ns/name/1`, but for probes it is just `probe/ns/name`, so we do a more generous prefix match to cover both cases. So we need to do something different for probes.

Maybe if `c.kind` is probe, clearConfigs should do something different, like a full equality match?

Or maybe the string matching on job name is silly, and we should create a full lookup of monitor names to a list of map keys, so we rely on fewer assumptions?

captncraig (changes requested):
I think we need to address the issue with probes before merging.

captncraig:
I'm thinking something like this: https://github.com/grafana/agent/commit/7268b5f856f54de3aba01b2489f1f2d3d73b951d, although I haven't tested it or merged with your tests. What do you think?

paul-bormans:
> I'm thinking something like this: [7268b5f](https://github.com/grafana/agent/commit/7268b5f856f54de3aba01b2489f1f2d3d73b951d), although I haven't tested it or merged with your tests. What do you think?

Let me check (tomorrow); perhaps I could write tests for probes first to make sure we don't break anything.

captncraig:
Thank you. I've incorporated your changes into #5862 since we would like to get this fix merged asap and possibly into a patch release. Thank you so much for tracking this one down.

PR #5846, opened 2023-11-24
fix: use official hashicorp consul image

tranngoclam:
#### PR Description

- Use the official HashiCorp Consul image

from https://hub.docker.com/_/consul:

> Upcoming in Consul 1.16, we will stop publishing official Dockerhub images and publish only our Verified Publisher images. Users of Docker images should pull from [hashicorp/consul](https://hub.docker.com/r/hashicorp/consul) instead of [consul](https://hub.docker.com/_/consul). Verified Publisher images can be found at https://hub.docker.com/r/hashicorp/consul.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,8 @@ Main (unreleased)
 
 - Fix issue with windows_exporter defaults not being set correctly. (@mattdurham)
 
+- Use official HashiCorp Consul image. (@tranngoclam)
+
 v0.38.0 (2023-11-21)
 --------------------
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -35,6 +35,8 @@ Main (unreleased)
 
 - Fix issue with windows_exporter defaults not being set correctly. (@mattdurham)
 
+- Use official HashiCorp Consul image. (@tranngoclam)
+
```
rfratto, line 39: This isn't a user facing change (i.e., something that's a part of the binary and is exposed to end users), so the changelog entry should be removed.

We still appreciate the contribution though! 

```suggestion
```

PR #5847, opened 2023-11-24
Improve the docs of pyroscope.scrape

ptodev:
Adding extra information which me and @thampiotr recently learned while tweaking Agents running `pyroscope.scrape`.

The PR also cleans up the docs in general.

I am not sure who should review this PR from the Docs squad. IIUC @clayton-cornell is normally helping with Agent docs whereas @knylander-grafana usually helps with Pyroscope docs :)

### DIFF
--- a/docs/sources/flow/reference/components/pyroscope.scrape.md
+++ b/docs/sources/flow/reference/components/pyroscope.scrape.md
@@ -17,7 +17,7 @@ title: pyroscope.scrape
 
 `pyroscope.scrape` collects [pprof] performance profiles for a given set of HTTP `targets`. 
 
-`pyroscope.scrape` borrows the scraping behavior of `prometheus.scrape`.
+`pyroscope.scrape` mimcks the scraping behavior of `prometheus.scrape`.
 Similarly to how Prometheus scrapes metrics via HTTP, `pyroscope.scrape` collects profiles via HTTP requests.
 
 Unlike Prometheus, which usually only scrapes one `/metrics` endpoint per target, 
--- a/docs/sources/flow/reference/components/pyroscope.scrape.md
+++ b/docs/sources/flow/reference/components/pyroscope.scrape.md
@@ -20,7 +20,7 @@ title: pyroscope.scrape
 `pyroscope.scrape` borrows the scraping behavior of `prometheus.scrape`.
 Similarly to how Prometheus scrapes metrics via HTTP, `pyroscope.scrape` collects profiles via HTTP requests.
 
-Unlike how Prometheus usually only scrapes one `/metrics` endpoint per target, 
+Unlike Prometheus, which usually only scrapes one `/metrics` endpoint per target, 
 `pyroscope.scrape` may need to scrape multiple endpoints for the same target.
 This is because different types of profiles are scraped on different endpoints. 
 For example, "mutex" profiles may be scraped on a `/debug/pprof/delta_mutex` HTTP endpoint, whereas 
@@ -130,14 +130,16 @@ so that they can be linked to a scrape target:
 
 If `scrape_interval` is short:
 * Advantages:
-  * Less profiles may be lost if the application being scraped crashes.
+  * Fewer profiles may be lost if the application being scraped crashes.
 * Disadvantages:
-  * Greater network load during scrapes and remote writes.
+  * Greater consumption of CPU, memory, and network resources during scrapes and remote writes.
+  * The backend database (Pyroscope) will consume more storage space.
 
 If `scrape_interval` is long:
 * Advantages:
-  * More profiles may be lost if the application being scraped crashes.
+  * Lower resource consumption.
 * Disadvantages:
+  * More profiles may be lost if the application being scraped crashes.
   * If the [delta argument][] is set to `true`, the batch size of 
     each remote write to Pyroscope may be bigger.
     The Pyroscope database may need to be tuned with higher limits.
@@ -152,7 +154,7 @@ For example, consider this situation:
 
 #### `scrape_timeout` argument
 
-`scrape_timeout` should be larger than `scrape_interval`.
+`scrape_timeout` must be larger than `scrape_interval`.
 
 ## Blocks
 
--- a/docs/sources/flow/reference/components/pyroscope.scrape.md
+++ b/docs/sources/flow/reference/components/pyroscope.scrape.md
@@ -42,7 +42,7 @@ The scraped performance profiles can be forwarded to components such as
 
 Multiple `pyroscope.scrape` components can be specified by giving them different labels.
 
-[debug UI]: {{< relref "../../monitoring/debugging/" >}}
+[debug UI]: {{< relref "../../tasks/debug.md" >}}
 
 ## Usage
 
--- a/docs/sources/flow/reference/components/pyroscope.scrape.md
+++ b/docs/sources/flow/reference/components/pyroscope.scrape.md
@@ -34,7 +34,7 @@ responded with an HTTP `200 OK` status code and returned the body of a valid [pp
 
 If a scrape request fails, the [debug UI][] for `pyroscope.scrape` will show:
 * Detailed information about the failure.
-* The last successful scrape.
+* The time of the last successful scrape.
 * The labels last used for scraping.
 
 The scraped performance profiles can be forwarded to components such as 
@@ -75,7 +75,7 @@ Name                | Type                     | Description                    
 `job_name`          | `string`                 | The job name to override the job label with.                       | component name | no
 `params`            | `map(list(string))`      | A set of query parameters with which the target is scraped.        |                | no
 `scrape_interval`   | `duration`               | How frequently to scrape the targets of this scrape configuration. | `"15s"`        | no
-`scrape_timeout`    | `duration`               | The timeout for scraping targets of this configuration.            | `"18s"`        | no
+`scrape_timeout`    | `duration`               | The timeout for scraping targets of this configuration. Must be larger than `scrape_interval`. | `"18s"`        | no
 `scheme`            | `string`                 | The URL scheme with which to fetch metrics from targets.           | `"http"`       | no
 `bearer_token`      | `secret`                 | Bearer token to authenticate with.                                 |                | no
 `bearer_token_file` | `string`                 | File containing a bearer token to authenticate with.               |                | no
@@ -96,7 +96,7 @@ At most one of the following authentication mechanisms can be provided:
 
 `job_name` defaults to the component's unique identifier.
 
-For example, the `job_name` of `pyroscope.scrape "local" { ... }` will be `"pyroscope.scrape/local"`.
+For example, the `job_name` of `pyroscope.scrape "local" { ... }` will be `"pyroscope.scrape.local"`.
 
 #### `targets` argument
 
@@ -128,6 +128,10 @@ so that they can be linked to a scrape target:
 
 #### `scrape_interval` argument
 
+The `scrape_interval` typically refers to the frequency with which {{< param "PRODUCT_NAME" >}} collects performance profiles from the monitored targets. 
+It represents the time interval between consecutive scrapes or data collection events. 
+This parameter is important for controlling the trade-off between resource usage and the freshness of the collected data.
+
 If `scrape_interval` is short:
 * Advantages:
   * Fewer profiles may be lost if the application being scraped crashes.
@@ -151,10 +155,6 @@ For example, consider this situation:
 * The application being scraped is running an HTTP server with a timeout of 30 seconds.
 * Any scrape HTTP requests where the [delta argument][] is set to `true` will fail, 
   because they will attempt to run for 59 seconds.
-
-#### `scrape_timeout` argument
-
-`scrape_timeout` must be larger than `scrape_interval`.
 
 ## Blocks
 
--- a/docs/sources/flow/reference/components/pyroscope.scrape.md
+++ b/docs/sources/flow/reference/components/pyroscope.scrape.md
@@ -15,11 +15,34 @@ title: pyroscope.scrape
 
 {{< docs/shared lookup="flow/stability/beta.md" source="agent" version="<AGENT_VERSION>" >}}
 
-`pyroscope.scrape` configures a [pprof] scraping job for a given set of
-`targets`. The scraped performance profiles are forwarded to the list of receivers passed in
-`forward_to`.
+`pyroscope.scrape` collects [pprof] performance profiles for a given set of HTTP `targets`. 
+
+`pyroscope.scrape` borrows the scraping behavior of `prometheus.scrape`.
+Similarly to how Prometheus scrapes metrics via HTTP, `pyroscope.scrape` collects profiles via HTTP requests.
+
+Unlike how Prometheus usually only scrapes one `/metrics` endpoint per target, 
+`pyroscope.scrape` may need to scrape multiple endpoints for the same target.
+This is because different types of profiles are scraped on different endpoints. 
+For example, "mutex" profiles may be scraped on a `/debug/pprof/delta_mutex` HTTP endpoint, whereas 
+memory consumption may be scraped on a `/debug/pprof/allocs` HTTP endpoint.
+
+The profile paths, protocol scheme, scrape interval, scrape timeout,
+query parameters, as well as any other settings can be configured within `pyroscope.scrape`.
+
+The `pyroscope.scrape` component regards a scrape as successful if it
+responded with an HTTP `200 OK` status code and returned the body of a valid [pprof] profile.
+
+If a scrape request fails, the [debug UI][] for `pyroscope.scrape` will show:
+* Detailed information about the failure.
+* The last successful scrape.
+* The labels last used for scraping.
+
+The scraped performance profiles can be forwarded to components such as 
+`pyroscope.write` via the `forward_to` argument.
 
 Multiple `pyroscope.scrape` components can be specified by giving them different labels.
+
+[debug UI]: {{< relref "../../monitoring/debugging/" >}}
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


docs/sources/flow/reference/components/pyroscope.scrape.md
```diff
@@ -15,12 +15,35 @@ title: pyroscope.scrape
...
-`pyroscope.scrape` configures a [pprof] scraping job for a given set of
-`targets`. The scraped performance profiles are forwarded to the list of receivers passed in
-`forward_to`.
+`pyroscope.scrape` collects [pprof] performance profiles for a given set of HTTP `targets`. 
+
+`pyroscope.scrape` borrows the scraping behavior of `prometheus.scrape`.
+Similarly to how Prometheus scrapes metrics via HTTP, `pyroscope.scrape` collects profiles via HTTP requests.
+
+Unlike how Prometheus usually only scrapes one `/metrics` endpoint per target, 
+`pyroscope.scrape` may need to scrape multiple endpoints for the same target.
```
thampiotr, line 24:
```suggestion
Unlike Prometheus, which usually only scrapes one `/metrics` endpoint per target, 
`pyroscope.scrape` may need to scrape multiple endpoints for the same target.
```

docs/sources/flow/reference/components/pyroscope.scrape.md
```diff
@@ -32,45 +55,104 @@ pyroscope.scrape "LABEL" {
...
-
- At most one of the following can be provided:
+Name                | Type                     | Description                                                  | Default        | Required
+------------------- | ------------------------ | ------------------------------------------------------------ | -------------- | --------
+`targets`           | `list(map(string))`      | List of targets to scrape.                                   |                | yes
+`forward_to`        | `list(ProfilesReceiver)` | List of receivers to send scraped profiles to.               |                | yes
+`job_name`          | `string`                 | The job name to override the job label with.                 | component name | no
+`params`            | `map(list(string))`      | A set of query parameters with which the target is scraped.  |                | no
+`scrape_interval`   | `duration`               | How frequently to scrape the targets of this scrape config.  | `"15s"`        | no
+`scrape_timeout`    | `duration`               | The timeout for scraping targets of this config.             | `"18s"`        | no
```
thampiotr, line 78: ah, thanks for fixing

docs/sources/flow/reference/components/pyroscope.scrape.md
```diff
@@ -322,6 +404,19 @@ If the agent is _not_ running in clustered mode, this block is a no-op.
 
 [using clustering]: {{< relref "../../concepts/clustering.md" >}}
 
+## Common configuration
+
+### `delta` argument
+
+When the `delta` argument is `false`, the [pprof][] HTTP query will be instantaneous.
```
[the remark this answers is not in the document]
ptodev: "Delta" to me means that the difference between the first sample and the last sample. 
Is that the case here? I was under the impression that all samples for the duration of seconds=14 will be taken into account.

docs/sources/flow/reference/components/pyroscope.scrape.md
```diff
@@ -32,45 +55,104 @@ pyroscope.scrape "LABEL" {
...
+|------------------|----------------------------------------------------------------- |
+| `"job"`          | The `job_name` that the target belongs to.                       |
+| `"instance"`     | The `__address__` or `<host>:<port>` of the scrape target's URL. |
+| `"service_name"` | The inferred Pyroscope service name.                             |
+
+#### `scrape_interval` argument
+
+If `scrape_interval` is short:
+* Advantages:
+  * Less profiles may be lost if the application being scraped crashes.
```
thampiotr, line 133:
```suggestion
  * Fewer profiles may be lost if the application being scraped crashes.
```

docs/sources/flow/reference/components/pyroscope.scrape.md
```diff
@@ -32,45 +55,104 @@ pyroscope.scrape "LABEL" {
...
+* Disadvantages:
+  * Greater network load during scrapes and remote writes.
```
thampiotr, line 135: I'd say greater resource usage in general. Also more data stored in the DB.

docs/sources/flow/reference/components/pyroscope.scrape.md
```diff
@@ -32,45 +55,104 @@ pyroscope.scrape "LABEL" {
...
+| Label            | Description                                                      |
+|------------------|----------------------------------------------------------------- |
+| `"job"`          | The `job_name` that the target belongs to.                       |
+| `"instance"`     | The `__address__` or `<host>:<port>` of the scrape target's URL. |
+| `"service_name"` | The inferred Pyroscope service name.                             |
+
+#### `scrape_interval` argument
+
+If `scrape_interval` is short:
+* Advantages:
```
thampiotr, line 132: One advantage is also being able to catch events such as OOM when they happen quickly. So maybe something like "Ability to capture rapid changes in profiles".

docs/sources/flow/reference/components/pyroscope.scrape.md
```diff
@@ -32,45 +55,104 @@ pyroscope.scrape "LABEL" {
...
+  * Less profiles may be lost if the application being scraped crashes.
+* Disadvantages:
+  * Greater network load during scrapes and remote writes.
+
+If `scrape_interval` is long:
+* Advantages:
+  * More profiles may be lost if the application being scraped crashes.
```
thampiotr, line 139: That's a disadvantage.

docs/sources/flow/reference/components/pyroscope.scrape.md
```diff
@@ -32,45 +55,104 @@ pyroscope.scrape "LABEL" {
...
+#### `scrape_interval` argument
+
+If `scrape_interval` is short:
+* Advantages:
+  * Less profiles may be lost if the application being scraped crashes.
+* Disadvantages:
+  * Greater network load during scrapes and remote writes.
+
+If `scrape_interval` is long:
+* Advantages:
```
thampiotr, line 138: Lower resource consumption.

docs/sources/flow/reference/components/pyroscope.scrape.md
```diff
@@ -32,45 +55,104 @@ pyroscope.scrape "LABEL" {
...
+
+For example, consider this situation:
+* `pyroscope.scrape` is configured with a `scrape_interval` of `"60s"`.
+* The application being scraped is running an HTTP server with a timeout of 30 seconds.
+* Any scrape HTTP requests where the [delta argument][] is set to `true` will fail, 
+  because they will attempt to run for 59 seconds.
+
+#### `scrape_timeout` argument
+
+`scrape_timeout` should be larger than `scrape_interval`.
```
thampiotr, line 155:
```suggestion
`scrape_timeout` must be larger than `scrape_interval`.
```

docs/sources/flow/reference/components/pyroscope.scrape.md
thampiotr, line 78: I think it's best to add it right here: "It must be larger than `scrape_interval`".

thampiotr (approved):
Few minor comments, otherwise LGTM

docs/sources/flow/reference/components/pyroscope.scrape.md
```diff
@@ -32,45 +55,104 @@ pyroscope.scrape "LABEL" {
...
- [arguments]: #arguments
+[arguments]: #arguments
+
+#### `job_name` argument
+
+`job_name` defaults to the component's unique identifier.
+
+For example, the `job_name` of `pyroscope.scrape "local" { ... }` will be `"pyroscope.scrape/local"`.
+
+#### `targets` argument
```
ptodev, line 101: It's not normal to add a section for an argument inside the `## Arguments` section. However, I think it really helps here because there's lots to discuss. I added the sections using `####` and not `###` so that they don't stand out too much and so that they are not listed in the ToC:
Screenshot 2023-11-24 at 13 28 50

docs/sources/flow/reference/components/pyroscope.scrape.md
```diff
@@ -92,35 +174,41 @@ The following blocks are supported inside the definition of `pyroscope.scrape`:
...
+[profile.godeltaprof_mutex]: #profilegodeltaprof_mutex-block
+[profile.godeltaprof_block]: #profilegodeltaprof_block-block
+[profile.custom]: #profilecustom-block
 [pprof]: https://github.com/google/pprof/blob/main/doc/README.md
 [clustering]: #clustering-beta
 
 [fgprof]: https://github.com/felixge/fgprof
 [godeltaprof]: https://github.com/grafana/godeltaprof
 
+[delta argument]: #delta-argument
```
ptodev, line 210: > [godeltaprof]: https://github.com/grafana/godeltaprof

This is pointing to an archived repo?

docs/sources/flow/reference/components/pyroscope.scrape.md
```diff
@@ -288,7 +370,7 @@ profile.custom "PROFILE_TYPE" {
 Multiple `profile.custom` blocks can be specified. Labels assigned to
 `profile.custom` blocks must be unique across the component.
 
-The `profile.custom` block accepts the following arguments:
+The following arguments are supported:
```
ptodev, line 373: I'd be nice to have an example with `profile.custom` in the Examples section?

docs/sources/flow/reference/components/pyroscope.scrape.md
ptodev, line 1: @clayton-cornell - The changes in the PR make this page more in line with our [Writing documentation for Flow components](https://github.com/grafana/agent/blob/main/docs/developer/writing-flow-component-documentation.md) guide, but I have taken a few liberties which are noted in comments.

@knylander-grafana - there is a [Go (pull mode)](https://grafana.com/docs/pyroscope/latest/configure-client/grafana-agent/go_pull/) page on the Pyroscope website which contains lots of duplicate information. Should we remove the duplications?

docs/sources/flow/reference/components/pyroscope.scrape.md
ptodev: I can see how this is true if `duration = false`, because then profiles would work similarly to Prometheus metrics.

However, if `duration = true`, wouldn't a lower scrape frequency lead to less catched events?

E.g. if `scrape_frequency = "15s"`, then each observation will last 14 seconds. For every 60 seconds we will lose 4 seconds of observations - this is the time between scrapes. On the other hand, for `scrape_frequency = "60s"` we will only lose 1 second worth of samples for every minute.

It'd be great if someone from the Pyroscope team confirms this.

thampiotr: The problem is when a bug leads to e.g. a rapid growth in memory and within 20 seconds the instance OOMs. This is much harder to catch with 1m scrape interval, as you'd need to be lucky to finish the profile just a few seconds before OOM event happens.

docs/sources/flow/reference/components/pyroscope.scrape.md
ptodev: True, I'll fix it :)

docs/sources/flow/reference/components/pyroscope.scrape.md
ptodev: I'm not too sure about this one. We usually document constraints under the table... I'll let the comment sit for now in case someone else wants to opine.

docs/sources/flow/reference/components/pyroscope.scrape.md
ptodev: I agree that when analysing OOMs, it's better to have a short scrape frequency. However, in case the application is not crashing and `duration = true`, I think that longer scrapes will actually result in more samples and will better detect rapid changes.

thampiotr: What do you refer to in `duration = false|true`? I don't see such setting on pyroscope.scrape.

ptodev: Sorry, I meant `delta = true`. I'll edit my comment above.

docs/sources/flow/reference/components/pyroscope.scrape.md
```diff
@@ -15,12 +15,35 @@ title: pyroscope.scrape
 
 {{< docs/shared lookup="flow/stability/beta.md" source="agent" version="<AGENT VERSION>" >}}
 
-`pyroscope.scrape` configures a [pprof] scraping job for a given set of
-`targets`. The scraped performance profiles are forwarded to the list of receivers passed in
-`forward_to`.
+`pyroscope.scrape` collects [pprof] performance profiles for a given set of HTTP `targets`. 
+
```
Rperry2174, line 19: I usually share this picture to show visually what the grafana agent does... may be worth adding in here:

image

docs/sources/flow/reference/components/pyroscope.scrape.md
Rperry2174: We created that page in response to feedback that this page was too verbose and hard to get started quickly. We will definitely keep the other page for now and I'll just manage the duplicate work of making sure changes made here make it over there

docs/sources/flow/reference/components/pyroscope.scrape.md
```diff
@@ -15,12 +15,35 @@ title: pyroscope.scrape
...
+`pyroscope.scrape` borrows the scraping behavior of `prometheus.scrape`.
```
Rperry2174, line 20:
```suggestion
`pyroscope.scrape` mimcks the scraping behavior of `prometheus.scrape`.
```

docs/sources/flow/reference/components/pyroscope.scrape.md
Rperry2174: cc @cyriltovena curious your thoughts here

docs/sources/flow/reference/components/pyroscope.scrape.md
```diff
@@ -32,45 +55,104 @@ pyroscope.scrape "LABEL" {
...
  - [`oauth2` block][oauth2].
 
- [arguments]: #arguments
+[arguments]: #arguments
+
+#### `job_name` argument
+
+`job_name` defaults to the component's unique identifier.
+
+For example, the `job_name` of `pyroscope.scrape "local" { ... }` will be `"pyroscope.scrape/local"`.
```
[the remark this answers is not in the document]
Rperry2174: cc @korniltsev

docs/sources/flow/reference/components/pyroscope.scrape.md
```diff
@@ -32,45 +55,104 @@ pyroscope.scrape "LABEL" {
...
+
+The list of `targets` can be provided [statically][example_static_targets], [dynamically][example_dynamic_targets], 
+or a [combination of both][example_static_and_dynamic_targets].
+
+The special `__address__` label _must always_ be present and corresponds to the
+`<host>:<port>` that is used for the scrape request.
+
+Labels starting with a double underscore (`__`) are treated as _internal_, and are removed prior to scraping.
+
+The special label `service_name` is required and must always be present.
```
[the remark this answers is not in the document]
Rperry2174: yes.. much of the ui and backend relies on this existing... technically I believe explore page will work if this doesn't exist, but for all intents and purposes this is required

docs/sources/flow/reference/components/pyroscope.scrape.md
```diff
@@ -32,45 +55,104 @@ pyroscope.scrape "LABEL" {
...
+The following labels are automatically injected to the scraped profiles 
+so that they can be linked to a scrape target:
+
+| Label            | Description                                                      |
+|------------------|----------------------------------------------------------------- |
+| `"job"`          | The `job_name` that the target belongs to.                       |
+| `"instance"`     | The `__address__` or `<host>:<port>` of the scrape target's URL. |
+| `"service_name"` | The inferred Pyroscope service name.                             |
+
+#### `scrape_interval` argument
```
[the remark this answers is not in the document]
Rperry2174: cc @cyriltovena we should give a straightforward explanation similar to that given in slack that explains how this interval works

docs/sources/flow/reference/components/pyroscope.scrape.md
```diff
@@ -152,98 +240,92 @@ Name | Type | Description | Default | Required
...
-When the `delta` argument is `true`, a `seconds` query parameter is
-automatically added to requests.
+For more information about the `delta` argument, see the [delta argument][] section.
 
 ### profile.block block
 
 The `profile.block` block collects profiles on process blocking.
 
-It accepts the following arguments:
+The following arguments are supported:
```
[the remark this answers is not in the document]
clayton-cornell: Working on it :-)  I dont' have a PR ready yet, but yes... block is not necessary in the child headings under the `Blocks` parent.

docs/sources/flow/reference/components/pyroscope.scrape.md
ptodev: Since the picture shows the interactions between different components, I'd prefer not to place it in the reference docs for pyroscope.scrape. It'd be great to have a new "Collecting Profiling Go Applications" task-based doc in the [Getting Started](https://grafana.com/docs/agent/next/flow/getting-started/) section. That doc can walk people through installing the Agent for profiling and we could put the picture there.

I think having such an architecture picture would be great for most task-based docs, and it would be even better if they are all consistent.
cc @clayton-cornell

docs/sources/flow/reference/components/pyroscope.scrape.md
ptodev: We could look into converting [Go (pull mode)](https://grafana.com/docs/pyroscope/latest/configure-client/grafana-agent/go_pull/) and [eBPF](https://grafana.com/docs/pyroscope/latest/configure-client/grafana-agent/ebpf/) into new task-based docs under Agent's [Getting Started](https://grafana.com/docs/agent/next/flow/getting-started/) section called "Profiling Go Applications" and "Profiling with eBPF".

docs/sources/flow/reference/components/pyroscope.scrape.md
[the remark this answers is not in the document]
korniltsev: it defaults to Options.ID which should be "pyroscope.scrape.local"

docs/sources/flow/reference/components/pyroscope.scrape.md
korniltsev: I think it is a bug to require scrape_timeout longer that scrape_interval.
It may make sense for cpu, but does not makes sense for other (not delta profiles) requests which should not last long

docs/sources/flow/reference/components/pyroscope.scrape.md
ptodev: The comment is referring to https://github.com/grafana/godeltaprof
Should we change the link to: https://github.com/grafana/pyroscope-go/tree/main/godeltaprof ?

ptodev: I'll just change it

docs/sources/flow/reference/components/pyroscope.scrape.md
thampiotr: I see. Shouldn't the docs therefore explain this - that CPU profiles may need different timeout & interval than the remaining profile types?

docs/sources/flow/reference/components/pyroscope.scrape.md
```diff
@@ -15,12 +15,35 @@ title: pyroscope.scrape
...
+
+The profile paths, protocol scheme, scrape interval, scrape timeout,
+query parameters, as well as any other settings can be configured within `pyroscope.scrape`.
+
+The `pyroscope.scrape` component regards a scrape as successful if it
+responded with an HTTP `200 OK` status code and returned the body of a valid [pprof] profile.
+
+If a scrape request fails, the [debug UI][] for `pyroscope.scrape` will show:
+* Detailed information about the failure.
+* The last successful scrape.
```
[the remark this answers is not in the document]
cyriltovena: yes I supposed so

docs/sources/flow/reference/components/pyroscope.scrape.md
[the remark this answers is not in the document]
cyriltovena: Sure we can describe what this setting changes.

_the "scrape_interval" typically refers to the frequency at which the agents collects performance profiles from the monitored targets. It represents the time interval between consecutive scrapes or data collection events. This parameter is important for controlling the trade-off between resource usage and the freshness of the collected data._

docs/sources/flow/reference/components/pyroscope.scrape.md
```diff
@@ -142,7 +230,7 @@ an `oauth2` block.
 The `profiling_config` block configures the profiling settings when scraping
 targets.
 
-The block contains the following attributes:
+The following arguments are supported:
```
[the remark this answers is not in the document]
cyriltovena: Might be a bug yes. Worth confirming it doesn't work and open an issue.

clayton-cornell:
@ptodev Does this make this one ready for merge? There's a lot of unresolved comments throughout the doc.

docs/sources/flow/reference/components/pyroscope.scrape.md
ptodev: For now I will just state "must be larger than `scrape_interval`" in the table as @thampiotr  suggested. We usually state such things under the table, but I think it improves readability in this case and it's worth making an exception.

> Shouldn't the docs therefore explain this - that CPU profiles may need different timeout & interval than the remaining profile types?

I think for now we can just explain how the component behaves, and we can improve with such advice later on. I think what @korniltsev states applies not just for CPU, but any "delta" profile.

Btw I think the reason for timing out delta profiles before the next query is started is that there can't be more than one active pprof for a certain type at a time. However, even if we have a very slow non-delta pprof query, I suppose we still won't be able to send a new query until the old one has finished. This is just a guess though.

ptodev:
> @ptodev Does this make this one ready for merge? There's a lot of unresolved comments throughout the doc.

@clayton-cornell I am happy to merge it. It is a good improvement over what we have atm, and most of the discussions are closed now. The only remaining thing I'm curious about is how [delta vs-non-delta profiles are displayed in the UI](https://github.com/grafana/agent/pull/5847#discussion_r1404386773), but we could ask about that some other time. Would you mind if I merge it please? I'm not sure if you want to review further.

clayton-cornell:
@ptodev Lets get this one published. We can open a new PR when needed ;-)

PR #5850, opened 2023-11-24
Fixed the url links

htcosta:
#### PR Description
Links not displaying correctly, due to concatenation of two markdown elements.
 
Backticks should be used inside the link text, to properly be displayed as inline code style, while still preserving the link element.
This PR fixes these two entries.


clayton-cornell (approved):
Thank you for catching this!  These things slip through from time to time.

PR #5858, opened 2023-11-27
discovery.process and pyroscope.java

korniltsev:
#### PR Description
`discovery.process` is pretty simple
There are couple of things worth noting for `pyroscope.java`

We bundle two versions of async-profiler one for musl, second for glibc. The musl one is compiled with static stdc++ because alpine often does not have stdc++ installed by default. `asprof` is compiled statically, so that we have 1 binary instead of 2.
`libAsyncProfiler.so ` absolute path should be exactly the same on host as inside container.

This is the layout of the bundled `.tar.gz`
```
// asprof
// glibc / libasyncProfiler.so
// musl / libasyncProfiler.so
```

Extracted to  this layout on the host. And shared library is copied to the target process at the same path via `/proc/pid/root/...`
```
./grafana-agent-asprof-musl-f97eb5b45a225752d7b26bb0fd4f2e37d1ddf5d5
./grafana-agent-asprof-musl-f97eb5b45a225752d7b26bb0fd4f2e37d1ddf5d5/bin
./grafana-agent-asprof-musl-f97eb5b45a225752d7b26bb0fd4f2e37d1ddf5d5/bin/asprof
./grafana-agent-asprof-musl-f97eb5b45a225752d7b26bb0fd4f2e37d1ddf5d5/lib
./grafana-agent-asprof-musl-f97eb5b45a225752d7b26bb0fd4f2e37d1ddf5d5/lib/libasyncProfiler.so
./grafana-agent-asprof-glibc-f97eb5b45a225752d7b26bb0fd4f2e37d1ddf5d5
./grafana-agent-asprof-glibc-f97eb5b45a225752d7b26bb0fd4f2e37d1ddf5d5/bin
./grafana-agent-asprof-glibc-f97eb5b45a225752d7b26bb0fd4f2e37d1ddf5d5/bin/asprof
./grafana-agent-asprof-glibc-f97eb5b45a225752d7b26bb0fd4f2e37d1ddf5d5/lib
./grafana-agent-asprof-glibc-f97eb5b45a225752d7b26bb0fd4f2e37d1ddf5d5/lib/libasyncProfiler.so

```

It is loaded from `bin/apsorf` binary as 'bin/../lib/libasyncProfiler.so`

asprof is built from upstream main branch with these changes to makefile https://gist.github.com/korniltsev/972f8ee37f3280af6757d7b8fec445d0

I had one interesting bug when my unpacking code was overwriting shared libraries in the target process root fs causing SIGSEGV in `jvm->dlopen("libAsyncProfiler.so")` so now it does not overwrite files.
#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


component/discovery/process/process.go
```diff
@@ -0,0 +1,99 @@
...
+	joinUpdates chan []discovery.Target
+}
+
+func (c *Component) Run(ctx context.Context) error {
+	processes, err := discover(c.l, c.procFS)
+	if err != nil {
+		return err
+	}
+	c.processes = convertProcesses(processes)
+	c.changed()
```
cyriltovena, line 69: nit: This could be a function

```
sync = func() {....}
```

So line 77 you don't repeat

cyriltovena (commented):
Really simple, love it so far !

component/pyroscope/java/java_stub.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/grafana/agent/pkg/flow/logging/level"
+)
+
+func init() {
+	component.Register(component.Registration{
+		Name: "pyroscope.java",
+		Args: Arguments{},
+
+		Build: func(opts component.Options, args component.Arguments) (component.Component, error) {
+			level.Warn(opts.Logger).Log("msg", "the pyroscope.java component only works on linux; enabling it otherwise will do nothing")
```
tpaschalis, line 18:
```suggestion
			level.Warn(opts.Logger).Log("msg", "the pyroscope.java component only works on linux for amd64 and arm64; enabling it otherwise will do nothing")
```

tpaschalis (commented):
Some high-level drive-by comments, didn't look so closely to the actual profiling bits.

component/pyroscope/java/args.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/grafana/agent/component/discovery"
+	"github.com/grafana/agent/component/pyroscope"
+)
+
+type Arguments struct {
+	Targets   []discovery.Target     `river:"targets,attr"`
+	ForwardTo []pyroscope.Appendable `river:"forward_to,attr"`
+
+	TmpDir          string          `river:"tmp_dir,attr,optional"`
```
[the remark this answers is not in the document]
cyriltovena: Yeah that's a good idea !

component/discovery/process/discover.go
```diff
@@ -0,0 +1,168 @@
...
+		res = append(res, process{
+			pid:         spid,
+			exe:         exe,
+			cwd:         cwd,
+			commandline: commandline,
+			containerID: containerID,
+			username:    username,
+			uid:         uid,
+		})
+		//_ = level.Debug(l).Log("msg", "found process", "pid", p.Pid, "exe", exe, "cwd", cwd, "container_id", containerID, "commandline", commandline)
```
cyriltovena, line 149:
```suggestion

```

component/pyroscope/java/asprof/asprof_test.go
```diff
@@ -0,0 +1,155 @@
...
+//			if err != nil {
+//				return err
+//			}
+//		}
+//	}
+//	return nil
+//}
+
+func tempDir(t *testing.T) string {
+	dir, err := os.MkdirTemp("", "asprof-test")
```
cyriltovena, line 149: nit.

```suggestion
        t.Helper()
	dir, err := os.MkdirTemp("", "asprof-test")
```

component/pyroscope/java/java.go
```diff
@@ -0,0 +1,125 @@
...
+)
+
+func init() {
+	component.Register(component.Registration{
+		Name: "pyroscope.java",
+		Args: Arguments{},
+
+		Build: func(opts component.Options, args component.Arguments) (component.Component, error) {
+			if os.Getuid() != 0 {
+				return nil, fmt.Errorf("java profiler: must be run as root")
```
cyriltovena, line 30: Nice !

component/pyroscope/java/args.go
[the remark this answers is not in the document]
cyriltovena: @tpaschalis Does the `options.DataPath` is unique per component or do we need to create a folder for our component ?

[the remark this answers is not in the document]
korniltsev: The path should be the same on host and container, so the `pyroscope.java` component may recreate directories and I dont really want it to touch anything except /tmp. I already dont like the fact that if `/tmp` does not exist in the target container, the `pyroscope.java` component is currently creating it with 0755 permissions and without sticky bit.
It also should have read permissions for unprivileged users on host, so that unprivileged java process on host (not container, same mount ns) can read and load `libAsyncProfiler.so`.

The path should also be not writable by unprivileged user. We unpack an elf binary to `/tmp/grafana-agent-.../bin/asprof` and then invoke it as privileged user.  So we better make sure unprivileged user can not modify or manipulated it in any way(symlinks or incorrect permissions)

I've tried to implement these checks with `/tmp` having sticky bit, but still not 100% sure it is fully correct.

component/pyroscope/java/loop.go
```diff
@@ -0,0 +1,253 @@
...
+}
+
+func (p *profilingLoop) loop(ctx context.Context) {
+	if err := p.profiler.CopyLib(p.dist, p.pid); err != nil {
+		p.onError(fmt.Errorf("failed to copy libasyncProfiler.so: %w", err))
+		return
+	}
+	defer p.stop()
+	sleep := func() bool {
+		timer := time.NewTimer(p.interval())
```
cyriltovena, line 81:
```suggestion
		timer := time.NewTimer(p.interval())
		defer timer.Stop()
```

component/pyroscope/java/loop.go
```diff
@@ -0,0 +1,253 @@
...
+	for {
+		err := p.start()
+		if err != nil {
+			//  could happen when agent restarted - [ERROR] Profiler already started\n
+			alive := p.onError(fmt.Errorf("failed to start: %w", err))
+			if !alive {
+				return
+			}
+		}
+		done := sleep()
```
cyriltovena, line 98: you actually don't need done here. you could just do if ctx.Err() != nil { return }

component/discovery/process/args.go
```diff
@@ -0,0 +1,37 @@
...
+	Exe         bool `river:"exe,attr,optional"`
+	Commandline bool `river:"commandline,attr,optional"`
+	Username    bool `river:"username,attr,optional"`
+	UID         bool `river:"uid,attr,optional"`
+	ContainerID bool `river:"container_id,attr,optional"`
+}
+
+var DefaultConfig = Arguments{
+	Join:            nil,
+	RefreshInterval: 14 * time.Second,
```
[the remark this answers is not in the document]
korniltsev: I've set it to a different value than 15s, which is profiling interval for java component.
I just wanted it to be not in sync with java component for debugging purposes. I dont have an opinion on what could be a good default.

component/discovery/process/process.go
```diff
@@ -0,0 +1,85 @@
...
+
+		case jt := <-c.joinUpdates:
+			c.join = jt
+			c.changed()
+		}
+	}
+}
+
+func (c *Component) Update(args component.Arguments) error {
+	c.joinUpdates <- args.(Arguments).Join
```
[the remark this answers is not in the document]
korniltsev: yep, done

component/discovery/process/process.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+		c.processes = convertProcesses(processes)
+		c.changed()
+		return nil
+	}
+	if err := doDiscover(); err != nil {
+		return err
+	}
+
+	t := time.NewTicker(c.args.RefreshInterval)
```
cyriltovena, line 56:
```suggestion
	t := time.NewTicker(c.args.RefreshInterval)
	defer t.Stop()
```

cyriltovena: FYI here this is not a big deal, but it's good habit to release ticker and timer since otherwise they leak resource.

component/pyroscope/java/asprof/asprof_test.go
```diff
@@ -0,0 +1,156 @@
...
+			err = p.ExtractDistributions()
+			t.Logf("expected %s", err)
+			assert.Error(t, err)
+		})
+	}
+}
+
+//this one requires root and docker(or mount namespace)
+//func TestDistSymlinkIntoProcRoot(t *testing.T) {
+//	if os.Getuid() != 0 {
```
cyriltovena, line 82: Should you keep those but use `t.Skip()`

cyriltovena: This way you can still run it without uncommenting :)

component/pyroscope/java/loop.go
```diff
@@ -0,0 +1,254 @@
...
+		_ = level.Debug(p.logger).Log("err", err)
+	}
+	p.mutex.Lock()
+	defer p.mutex.Unlock()
+	p.error = err
+	return alive
+}
+
+func (p *profilingLoop) interval() time.Duration {
+	return time.Second * 15 // todo
```
cyriltovena, line 239: I think this should come from the config ?

component/pyroscope/java/loop.go
```diff
@@ -0,0 +1,254 @@
...
+		}
+		_ = l.Log("msg", "pushed jfr-pprof")
+	}
+	return nil
+}
+
+func (p *profilingLoop) start() error {
+	p.startTime = time.Now()
+	argv := make([]string, 0, 14)
+	argv = append(argv,
```
cyriltovena, line 170:
```suggestion
        // asprof cli reference: https://github.com/async-profiler/async-profiler?tab=readme-ov-file#profiler-options
	argv = append(argv,
```

component/discovery/process/args.go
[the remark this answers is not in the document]
korniltsev: I set it to 60s now

component/pyroscope/java/loop.go
```diff
@@ -0,0 +1,254 @@
...
+	p.startTime = endTime
+	defer func() {
+		os.Remove(jfrFile)
+	}()
+
+	err := p.stop()
+	if err != nil {
+		return fmt.Errorf("failed to stop : %w", err)
+	}
+	jfrBytes, err := os.ReadFile(jfrFile)
```
cyriltovena, line 126: Just an observation that we might want to use a pool for the slice of bytes in the future. If many `profilinLoop` exists we could re-use memory.

I don't think it's worth the effort until we see this as a problem.

component/pyroscope/java/asprof/asprof_test.go
korniltsev: I removed them. They are incomplete and used to require some manual steps like launching a docker container

component/pyroscope/java/loop.go
korniltsev: YES!

docs/sources/flow/reference/components/pyroscope.java.md
```diff
@@ -0,0 +1,149 @@
...
+canonical: https://grafana.com/docs/agent/latest/flow/reference/components/pyroscope.java/
+description: Learn about pyroscope.java
+title: pyroscope.java
+---
+
+# pyroscope.java
+
+{{< docs/shared lookup="flow/stability/beta.md" source="agent" version="<AGENT_VERSION>" >}}
+
+`pyroscope.java` continuously profiles Java processes running on the local Linux OS using async-profiler.
```
cyriltovena, line 16: I think we should link back to asprofiler github.

cyriltovena (approved):
Impressive ! This is really looking great 🚢

component/pyroscope/java/loop.go
korniltsev: fixed

korniltsev:
sorry for much noise

I'd appreciate docs review/ cc @clayton-cornell  @cyriltovena

cyriltovena:
Docs are good to me.

docs/sources/flow/reference/components/pyroscope.java.md
```diff
@@ -0,0 +1,177 @@
...
+| `forward_to` | `list(ProfilesReceiver)` | List of receivers to send collected profiles to. |         | yes      |
+| `tmp_dir`    | `string`                 | Temporary directory to store async-profiler      | `/tmp`  | no       |
+
+## Profiling behavior
+
+The special label `__process_pid__` _must always_ be present in each target of `targets` and corresponds to the `PID` of
+the process to profile.
+
+Upon component startup, `pyroscope.java` will create a temporary directory under `tmp_dir` and extract the
+async-profiler binaries for both glibc and musl into it with the following layout.
```
clayton-cornell, line 48:
```suggestion
async-profiler binaries for both glibc and musl into the directory with the following layout.
```

docs/sources/flow/reference/components/pyroscope.java.md
```diff
@@ -0,0 +1,177 @@
...
+
+```
+/tmp/grafana-agent-asprof-glibc-{SHA1}/bin/asprof
+/tmp/grafana-agent-asprof-glibc-{SHA1}/lib/libasyncProfiler.so
+/tmp/grafana-agent-asprof-musl-{SHA1}/bin/asprof
+/tmp/grafana-agent-asprof-musl-{SHA1}/lib/libasyncProfiler.so
+```
+
+Upon process profiling startup, the component detects libc type and copies according `libAsyncProfiler.so` into the
```
clayton-cornell, line 57:
```suggestion
After process profiling startup, the component detects libc type and copies according `libAsyncProfiler.so` into the
```

docs/sources/flow/reference/components/pyroscope.java.md
```diff
@@ -0,0 +1,177 @@
...
+
+The following arguments are supported:
+
+| Name          | Type       | Description                                                                                             | Default | Required |
+|---------------|------------|---------------------------------------------------------------------------------------------------------|---------|----------|
+| `interval`    | `duration` | How frequently to collect profiles from the targets.                                                    | "60s"   | no       |
+| `cpu`         | `bool`     | A flag to enable cpu profiling, using `itimer` async-profiler event.                                    | true    | no       |
+| `sample_rate` | `int`      | CPU profiling sample rate. It is converted from Hz to interval and passed as `-i` arg to async-profiler | 100     | no       |
+| `alloc`       | `string`   | Allocation profiling sampling configuration  It is passed as `--alloc` arg to async-profiler            | "512k"  | no       |
+| `lock`        | `string`   | Lock profiling sampling configuration. It is passed as `--lock` arg to async-profiler                   | "10ms"  | no       |
```
clayton-cornell, line 92:
```suggestion
| `sample_rate` | `int`      | CPU profiling sample rate. It is converted from Hz to interval and passed as `-i` arg to async-profiler. | 100     | no       |
| `alloc`       | `string`   | Allocation profiling sampling configuration  It is passed as `--alloc` arg to async-profiler.            | "512k"  | no       |
| `lock`        | `string`   | Lock profiling sampling configuration. It is passed as `--lock` arg to async-profiler.                   | "10ms"  | no       |
```

docs/sources/flow/reference/components/pyroscope.java.md
```diff
@@ -0,0 +1,177 @@
...
+{{% /admonition %}}
+
+## Blocks
+
+The following blocks are supported inside the definition of
+`pyroscope.java`:
+
+| Hierarchy        | Block                | Description                            | Required |
+|------------------|----------------------|----------------------------------------|----------|
+| profiling_config | [profiling_config][] | Describes java profiling configuration | no       |
```
clayton-cornell, line 76:
```suggestion
| profiling_config | [profiling_config][] | Describes java profiling configuration. | no       |
```

docs/sources/flow/reference/components/pyroscope.java.md
```diff
@@ -0,0 +1,177 @@
...
+/tmp/grafana-agent-asprof-musl-{SHA1}/bin/asprof
+/tmp/grafana-agent-asprof-musl-{SHA1}/lib/libasyncProfiler.so
+```
+
+Upon process profiling startup, the component detects libc type and copies according `libAsyncProfiler.so` into the
+target process file system at the exact same path.
+
+{{% admonition type="note" %}}
+The `asprof` binary is executed as privileged user.
+If user changes `tmp_dir` configuration to something other than `/tmp`, then it is up to the user to ensure that the
```
clayton-cornell, line 62:
```suggestion
If you change the `tmp_dir` configuration to something other than `/tmp`, then you must ensure that the
```

docs/sources/flow/reference/components/pyroscope.java.md
```diff
@@ -0,0 +1,177 @@
...
+directory is not writable by unprivileged users.
+{{% /admonition %}}
+
+{{% admonition type="note" %}}
+{{% /admonition %}}
+
```
clayton-cornell, line 68:
```suggestion

```

docs/sources/flow/reference/components/pyroscope.java.md
```diff
@@ -0,0 +1,177 @@
...
+/tmp/grafana-agent-asprof-glibc-{SHA1}/lib/libasyncProfiler.so
+/tmp/grafana-agent-asprof-musl-{SHA1}/bin/asprof
+/tmp/grafana-agent-asprof-musl-{SHA1}/lib/libasyncProfiler.so
+```
+
+Upon process profiling startup, the component detects libc type and copies according `libAsyncProfiler.so` into the
+target process file system at the exact same path.
+
+{{% admonition type="note" %}}
+The `asprof` binary is executed as privileged user.
```
clayton-cornell, line 61:
```suggestion
The `asprof` binary must be executed by a privileged user.
```

clayton-cornell (commented):
Some doc suggestions

docs/sources/flow/reference/components/pyroscope.java.md
korniltsev: just to clarify: it is the  `pyroscope.java` compoenents which executes `asprof` binary.

`pyroscope.java` should be run as privileged. Users do not execute `asprof` manually.

Is suggestion aligned with it?

korniltsev: cc @clayton-cornell

clayton-cornell: @korniltsev Ah that wasn't clear to me from the context. 

What about something like this:

```suggestion
The `asprof` binary runs with elevated permissions.
```

docs/sources/flow/reference/components/pyroscope.java.md
```diff
@@ -0,0 +1,173 @@
...
+/tmp/grafana-agent-asprof-musl-{SHA1}/lib/libasyncProfiler.so
+```
+
+After process profiling startup, the component detects libc type and copies according `libAsyncProfiler.so` into the
+target process file system at the exact same path.
+
+{{% admonition type="note" %}}
+The `asprof` binary is executed as privileged user.
+If you change the `tmp_dir` configuration to something other than `/tmp`, then you must ensure that the
+directory is not writable by unprivileged users.
```
clayton-cornell, line 63:
```suggestion
directory is only writable by root.
```

It might be better to state specifically what user can write to the dir vs the more ambiguous "unprivileged".

PR #5861, opened 2023-11-27
Replace docker compose by testcontainers

wildum:
#### PR Description

Test running integration tests via testcontainers instead of docker compose.


wildum:
Closing for now since that we might be moving away from Drone and that docker compose is quite useful for local testing

PR #5869, opened 2023-11-28
max_cache_size was being set to 0

mattdurham:
#### PR Description

MaxCacheSize doesnt exist in static mode but needs to be non zero in flow. The existing logic compares the iszero with the default. Since the input is a zero, and thats not the default it persists 0. This value cannot be 0 and needs to be a positive value. By setting it to the default manually it means it no longer needs  to be emitted which is the best case scenario.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated


converter/internal/promtailconvert/internal/build/scrape_builder.go
```diff
@@ -105,6 +105,10 @@ func (s *ScrapeConfigBuilder) getOrNewLokiRelabel() string {
 			RelabelConfigs: component.ToFlowRelabelConfigs(s.cfg.RelabelConfigs),
 		}
 		compLabel := common.LabelForParts(s.globalCtx.LabelPrefix, s.cfg.JobName)
+		// max_cache_size doesnt exist in static, and we need to manually set it  to default.
+		// Else since the go default value will be 0, the IsZero is also 0, so it will assume it should be 0 when comparing
+		// in appendblock. The check against the default value of 10_000 ensures it will always be written as 0.
+		args.MaxCacheSize = 10_000
```
erikbaranowski, line 111: Can this be set to `relabel.DefaultArguments.MaxCacheSize` instead? From `component/loki/relabel`

erikbaranowski: also can we just set this between line 105 and 106 as a 1 liner?

mattdurham: Will do and cleaned up the comment a bit.

erikbaranowski (approved):

PR #5872, opened 2023-11-28
build(packaging): provide zipped Flow binaries at release

### LINKED ISSUE #5728: Standalone executable binary for flow mode?
I just checked releases and I don't see an option for a standalone executable (or ZIP) for grafana agent flow for Linux amd64.

Maybe I'm sleep deprived and I just don't see it, but if it truly is not available, then why not? Isn't it written in go and thus be relatively simple to provide such a release?

If not, then how for example do you suggest we run it on alpine linux?

hainenber:

Fixes #5728 

#### PR Checklist

- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,10 @@ Main (unreleased)
 
 - Fix issue in `prometheus.operator.*` where targets would be dropped if two crds share a common prefix in their names. (@Paul424, @captncraig)
 
+### Other changes
+
+- Provide standalone Flow binaries on new version release. (@hainenber)
+
 v0.38.0 (2023-11-21)
 --------------------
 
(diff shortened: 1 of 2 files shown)


hainenber:
I think for starters, we can start providing Flow binaries for x86_64 and arm/64 only, given that they are mostly used. There would be 2 more binaries in the asset

Would there be any tradeoffs?

rfratto:
The tradeoff is really just bloating the release asset list, which could make it harder to find what you're looking for. We already have 39 assets as part of the release 😬

There's some minor tradeoff with increasing automation time to publish a release, but I'm not too concerned about that.

hainenber:
I see. OP's issue can also be addressed by running normal binaries in Flow mode. If anything, maybe we can have some kind of notes in GH release to inform folks of that fact 😄 

Close this one off for now

PR #5877, opened 2023-11-28
[docs] Fix Helm values indentation in migrating-from-operator.md

emirisman:
Here is another one :)


clayton-cornell (approved):
Looks good! Thanks for catching this one too. Incorrect indents slips through.

PR #5880, opened 2023-11-29
Add Deploy Mode to usage stats.

captncraig:
With 0.38.0 we added a deploy mode to the user agent, and added ways to set it, but we did not add it to the usage stats, which would be very useful. This adds it.

Also makes a best-effort attempt to detect homebrew installs, by looking t the executable path.

### DIFF
--- a/docs/sources/data-collection.md
+++ b/docs/sources/data-collection.md
@@ -30,6 +30,7 @@ The usage information includes the following details:
 * List of enabled feature flags ([Static] mode only).
 * List of enabled integrations ([Static] mode only).
 * List of enabled [components][] ([Flow] mode only).
+* Method used to deploy Grafana Agent (docker, helm, rpm, operator, etc.)
 
 This list may change over time. All newly reported data is documented in the CHANGELOG.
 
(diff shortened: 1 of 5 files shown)


jcreixell (commented):
could you please add a test [here](https://github.com/grafana/agent/blob/main/internal/useragent/useragent_test.go)? other than that, LGTM

rfratto (commented):
You'll need to update this page: https://grafana.com/docs/agent/latest/data-collection/

And add this in the CHANGELOG, since we promised we'd do that any time list of collected data changes.

captncraig:
Added tests and transparency.

docs/sources/data-collection.md
```diff
@@ -30,6 +30,7 @@ The usage information includes the following details:
 * List of enabled feature flags ([Static] mode only).
 * List of enabled integrations ([Static] mode only).
 * List of enabled [components][] ([Flow] mode only).
+* Method used to deploy Grafana Agent (docker, helm, rpm, operator, etc.)
```
clayton-cornell, line 33:
```suggestion
* Method used to deploy Grafana Agent, for example Docker, Helm, RPM, or Operator.
```

rfratto (approved):
LGTM once Clayton approves docs

PR #5894, opened 2023-11-30
Rename EvaluateDependencies to EvaluateDependants

wildum:
This PR just renames one function.

In this case we don't evaluate the dependencies, we evaluate the dependants of the updated nodes.

From the dag definition:

// Dependants returns the list of Nodes that depend on n: all Nodes for which an edge to n is defined.
func (g *Graph) Dependants(n Node) []Node

// Dependencies returns the list of Nodes that n depends on: all Nodes for which an edge from n is defined.
func (g *Graph) Dependencies(n Node) []Node


thampiotr (approved):
thanks!

PR #5895, opened 2023-11-30
Move changelog entry to correct place.

mattdurham:
I put the changelog for 5869 in the wrong place , it is now in 38.1 where it belongs.

- [x] CHANGELOG.md updated


rfratto (approved):
Thanks!

PR #5904, opened 2023-12-01
Explain tracing load balancing in more detail

ptodev:
When load balancing traces over multiple Agent instances, there are lots of potential issues that our users need to watch out for. These docs updates go into detail and provide examples. 

Some things are still not complete:

- I still have to add some more info regarding Static mode.
- Ideally we could also add pictures with the architecture.
- I need to make a working Static config which uses the Kubernetes resolver.
- I'm still not sure if the locations in the docs which I edited are the most appropriate ones. Especially the "Deploy Grafana Agent" page.

### DIFF
--- a/docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
+++ b/docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
@@ -297,9 +297,9 @@ All spans for a given `service.name` must go to the same spanmetrics Agent.
   * At worst, it could lead to an inaccurate data due to overlapping samples for the metric series.
 
 However, there are ways to scale `otelcol.connector.spanmetrics` without the need for a load balancer:
-1. Each Agent could add an attribute such as `agent.id` in order to make its series unique.
+1. Each Agent could add an attribute such as `collector.id` in order to make its series unique.
   * A `sum by` PromQL query could be used to aggregate the metrics from different Agents.
-  * An extra `agent.id` attribute has a downside that the metrics stored in the database will have higher {{< term "cardinality" >}}cardinality{{< /term >}}.
+  * An extra `collector.id` attribute has a downside that the metrics stored in the database will have higher {{< term "cardinality" >}}cardinality{{< /term >}}.
 2. Spanmetrics could be generated in the backend database instead of the Agent.
   * For example, span metrics can be [generated][tempo-spanmetrics] in Grafana Cloud by the Tempo traces database.
 
@@ -320,10 +320,10 @@ It is challenging to scale `otelcol.connector.servicegraph` over multiple Agent 
     they will generate the same service graph metric series.
   * If the series from two Agents are the same, this will lead to issues 
     when writing them to the backend database.
-  * Users could differentiate the series by adding an attribute such as `"agent.id"`.
+  * Users could differentiate the series by adding an attribute such as `"collector.id"`.
       * Unfortunately, there is currently no method in the Agent to aggregate those series from different Agents and merge them into one series.
     * A PromQL query could be used to aggregate the metrics from different Agents.
-    * If the metrics are stored in Grafana Mimir, cardinality issues due to `"agent.id"` labels can be solved using [Adaptive Metrics][adaptive-metrics].
+    * If the metrics are stored in Grafana Mimir, cardinality issues due to `"collector.id"` labels can be solved using [Adaptive Metrics][adaptive-metrics].
 
 A simpler, more scalable alternative to generating service graph metrics in the Agent is to generate them entirely in the backend database. 
 For example, service graphs can be [generated][tempo-servicegraphs] in Grafana Cloud by the Tempo traces database.
--- a/docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
+++ b/docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
@@ -291,16 +291,16 @@ All spans for a given `service.name` must go to the same spanmetrics Agent.
 * This can be done by configuring `otelcol.exporter.loadbalancing` with `routing_key = "service"`.
 * If this is not done, metrics generated from spans might be incorrect.
   * For example, if similar spans for the same `service.name` end up on different Agent instances,
-    the two Agents will have identical metic series for calculating span latency, errors, and number of requests.
+    the two Agents will have identical metric series for calculating span latency, errors, and number of requests.
   * When both Agents attempt to remote write the metrics to a database such as Mimir,
-    the series with clash with each other.
-  * At best this will lead to an error in the Agent and a rejected remote write. 
-  * At worst, it could lead to an inaccurate data dure to overlapping samples for the metric series.
+    the series may clash with each other.
+  * At best this will lead to an error in the Agent and a rejected write to the metrics database. 
+  * At worst, it could lead to an inaccurate data due to overlapping samples for the metric series.
 
 However, there are ways to scale `otelcol.connector.spanmetrics` without the need for a load balancer:
 1. Each Agent could add an attribute such as `agent.id` in order to make its series unique.
   * A `sum by` PromQL query could be used to aggregate the metrics from different Agents.
-  * An extra `agent.id` attribute has a downside that the metrics stored in the database will have extra cardinality.
+  * An extra `agent.id` attribute has a downside that the metrics stored in the database will have higher cardinality.
 2. Spanmetrics could be generated in the backend database instead of the Agent.
   * For example, span metrics can be [generated][tempo-spanmetrics] in Grafana Cloud by the Tempo traces database.
 
@@ -329,7 +329,7 @@ Unfortunately, there is generally no reliable way to scale `otelcol.connector.se
 There are ways to work around this:
   1. Each Agent could add an attribute such as `agent.id` in order to make its series unique.
     * A PromQL query could be used to aggregate the metrics from different Agents.
-    * An extra `agent.id` attribute has a downside that the metrics stored in the database will have extra cardinality.
+    * An extra `agent.id` attribute has a downside that the metrics stored in the database will have higher cardinality.
   2. A simpler, more scalable alternative to generating service graph metrics 
     in the Agent is to do them in the backend database. 
     * For example service graphs can be [generated][tempo-servicegraphs] in Grafana Cloud by the Tempo traces database.
@@ -419,7 +419,7 @@ otelcol.exporter.loadbalancing "default" {
 }
 ```
 
-Below is an example Kubernetes configuration which sets up two sets of Agents:
+Below is an example Kubernetes configuration which configures two sets of Agents:
 * A pool of "load balancer" Agents:
   * Spans are received from instrumented applications via `otelcol.receiver.otlp`
   * Spans are then exported via `otelcol.exporter.loadbalancing`.
--- a/docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
+++ b/docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
@@ -143,9 +143,9 @@ Name | Type | Description | Default | Required
 
 ### kubernetes block
 
-You can use the `kubernetes` block to load balance across the pods of a Kubernetes service. The Agent will be notified
-by the Kubernetes API whenever a new pod is added or removed from the service. The `kubernetes` resolver has a
-much faster response time than the `dns` resolver, because it doesn't require polling.
+You can use the `kubernetes` block to load balance across the pods of a Kubernetes service. 
+The Agent will be notified by the Kubernetes API whenever a new pod is added or removed from the service. 
+The `kubernetes` resolver has a much faster response time than the `dns` resolver, because it doesn't require polling.
 
 The following arguments are supported:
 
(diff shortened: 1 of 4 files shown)


docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
```diff
@@ -264,6 +267,95 @@ Name | Type | Description
...
+  * A `sum by` PromQL query could be used to aggregate the metrics from different Agents.
+  * An extra `agent.id` attribute has a downside that the metrics stored in the database will have extra cardinality.
+2. Spanmetrics could be generated in the backend database instead of the Agent.
+  * For example, span metrics can be [generated][tempo-spanmetrics] in Grafana Cloud by the Tempo traces database.
+
+[tempo-spanmetrics]: https://grafana.com/docs/tempo/latest/metrics-generator/span_metrics/
+
+### otelcol.connector.servicegraph
+Unfortunately, there is generally no reliable way to scale `otelcol.connector.servicegraph` over multiple Agent instance.
+<!-- Or is there? Can we fix the issue below using a PromQL query? There may still be higher than normal cardinality though? -->
```
thampiotr, line 311: Just curious: Would the service graph be only partially complete? If we have enough traces and not a very big cluster, would we have a high chance of catching all the edges in the graph anyway? Or would it always miss some edges consistently?

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
```diff
@@ -264,6 +267,95 @@ Name | Type | Description
...
+  then no single Agent will be able to pair the spans and a metric won't be generated.
+
+`otelcol.exporter.loadbalancing` can solve this problem partially if it is configured with `routing_key = "traceID"`.
+  * Each Agent will then be able to calculate service graph for each client/server pair in a trace.
+  * However, it is possible to have a span with similar "server"/"client" values 
+    in a different trace, processed by another Agent.
+  * If two different Agents, process similar "server"/"client" spans, 
+    they will generate the same service graph metric series.
+  * If the series from two Agents are the same, this will lead to issues 
+    when remote writing them in the backend database.
```
thampiotr, line 324:
```suggestion
  * If the series from two Agents are the same, this will lead to issues 
    when writing them in the backend database.
```
It's just writing to a DB...

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
```diff
@@ -264,6 +267,95 @@ Name | Type | Description
...
+
+  * Users could differentiate the series by adding a label such as `"agent_instance"`, 
+    but there is currently no method in the Agent to aggregate those series from different Agents and merge them into one series.
+
+There are ways to work around this:
+  1. Each Agent could add an attribute such as `agent.id` in order to make its series unique.
+    * A PromQL query could be used to aggregate the metrics from different Agents.
+    * An extra `agent.id` attribute has a downside that the metrics stored in the database will have extra cardinality.
```
thampiotr, line 332:
```suggestion
    * An extra `agent.id` attribute has a downside that the metrics stored in the database will have higher cardinality.
```

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
```diff
@@ -264,6 +267,95 @@ Name | Type | Description
...
+  2. A simpler, more scalable alternative to generating service graph metrics
```
thampiotr, line 333: Ah, so there is a way! maybe we should amend the text above then to hint that there is a workaround below.

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
thampiotr, line 1: All this can be complex, but some basic setups could be encapsulated in a reusable agent module(s) I think. Maybe worth considering in the future.

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
```diff
@@ -264,6 +267,95 @@ Name | Type | Description
...
+Different Flow components may require a different `routing_key` for `otelcol.exporter.loadbalancing`.
+* For example, `otelcol.processor.tail_sampling` requires `routing_key = "traceID"` 
+  whereas `otelcol.connector.spanmetrics` requires `routing_key = "service"`.
+* Therefore, it is not possible to use the same load balancer for both tail sampling and span metrics.
+* Two different sets of load balancers have to be set up:
+  * One set of `otelcol.exporter.loadbalancing` with `routing_key = "traceID"`, sending spans to Agents doing tail sampling and no span metrics.
+  * Another set of `otelcol.exporter.loadbalancing` with `routing_key = "service"`, sending spans to Agents doing span metrics and no service graphs.
+* Unfortunately, this can also lead to side effects. For example, if `otelcol.connector.spanmetrics` is configured to generate exemplars,
+  the tail sampling Agents might drop the trace which the exemplar points to. There is no coordination between the tail sampling Agents and
+  the span metrics Agents to make sure trace IDs for exemplars are kept.
```
thampiotr, line 350: oh dear... good callout tho

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
```diff
@@ -264,6 +267,95 @@ Name | Type | Description
...
+
+<!-- 
+TODO: Add a troubleshooting section?
+1. Use GODEBUG for DNS resolver logging
```
thampiotr, line 354: Nice trick!

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
```diff
@@ -264,6 +267,95 @@ Name | Type | Description
 * logs
 * traces
 
+## Choosing a load balancing strategy
```
thampiotr, line 270: I'm on the fence if this could become an article or if it belongs in this reference. One could argue that e.g. `spanmetrics` related comments should be in the `spanmetrics` reference page. A fair solution would be to have a task-based doc about scaling and link to it from everywhere.

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
```diff
@@ -301,3 +395,565 @@ otelcol.exporter.loadbalancing "default" {
...
+    }
+    protocol {
+        otlp {
+            client {}
+        }
+    }
+}
+```
+
+Below is an example Kubernetes configuration which sets up two sets of Agents:
```
thampiotr, line 422:
```suggestion
Below is an example Kubernetes configuration which configures two sets of Agents:
```

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
```diff
@@ -301,3 +395,565 @@ otelcol.exporter.loadbalancing "default" {
...
+* A pool of "load balancer" Agents:
```
thampiotr, line 423: Nit: I would consider calling them `load-balancing Agents` and `backing Agents` to avoid the quotation marks... gives the impression it's a temporary name, but in fact it seems like a perfectly good name.

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
```diff
@@ -301,3 +395,565 @@ otelcol.exporter.loadbalancing "default" {
...
+    protocol: TCP
+    targetPort: agent-lb
+  selector:
+    name: agent-traces-sampling
+  type: ClusterIP
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: agent-traces-sampling
```
thampiotr, line 544: I'd think this should match the text description, so it'd be `agent-traces-backing`.

docs/sources/static/configuration/traces-config.md
```diff
@@ -308,13 +308,27 @@ tail_sampling:
...
-# different agent instances can receive spans for the same trace.
+# Enabling this feature is required for "tail_sampling", "spanmetrics", and "service_graphs"
+# to correctly work when spans are ingested by multiple agent instances.
 #
 # Load balancing works by layering two pipelines and consistently exporting
 # spans belonging to a trace to the same agent instance.
 # Agent instances need to be able to communicate with each other via gRPC.
 #
+# When load_balancing is enabled:
+# 1. When an Agent receives spans from the configures "receivers".
```
thampiotr, line 319:
```suggestion
# 1. When an Agent receives spans from the configured "receivers".
```

docs/sources/static/set-up/deploy-agent.md
```diff
@@ -11,3 +11,190 @@ weight: 300
 
 {{< docs/shared source="agent" lookup="/deploy-agent.md" version="<AGENT VERSION>" >}}
 
+## For scalable ingestion of traces
```
thampiotr, line 14: I suspect some of the comments from flow mode will apply here too.

thampiotr (commented):
First pass, this is pretty large, would be good to figure out how to make it more brief and less repetitive, because as the way things work in OTEL change, we'd have more and more docs to properly maintain.

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
```diff
@@ -264,6 +267,95 @@ Name | Type | Description
...
+    the two Agents will have identical metic series for calculating span latency, errors, and number of requests.
+  * When both Agents attempt to remote write the metrics to a database such as Mimir,
+    the series with clash with each other.
+  * At best this will lead to an error in the Agent and a rejected remote write. 
+  * At worst, it could lead to an inaccurate data dure to overlapping samples for the metric series.
+
+However, there are ways to scale `otelcol.connector.spanmetrics` without the need for a load balancer:
+1. Each Agent could add an attribute such as `agent.id` in order to make its series unique.
+  * A `sum by` PromQL query could be used to aggregate the metrics from different Agents.
+  * An extra `agent.id` attribute has a downside that the metrics stored in the database will have extra cardinality.
```
ptodev, line 303: I think adaptive metrics might alleviate the cardinality issues, but I'm not sure exactly how. We could add more details about it in the future.

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
ptodev: TBH I don't like word "backing", but I use it here because I have heard it mentioned in OTel discussions. I think "sampling" makes it more clear what the Agents are doing. WDYT? I'll also ask the OTel experts if there is a convention we must follow.

docs/sources/flow/setup/deploy-agent.md
```diff
@@ -13,3 +13,58 @@ weight: 900
...
+
+However, if the load on the Agents is big, it may be beneficial to
+process different telemetry signals in different sets of Agents:
+* This provides bettor stability due to the isolation between processes.
+  * For example, an overloaded Agent processing traces won't impact an Agent processing metrics.
+* Different types of signal collection require different methods for scaling:
+  * "Pull" components such as `prometheus.scrape` and `pyroscope.scrape` are scaled using hashmod sharing or clustering. 
+  * "Push" components such as `otelcol.receiver.otlp` are scaled by placing a load balancer in front of them.
+
+### Traces
```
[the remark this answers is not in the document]
ptodev: Having a section for traces helps in two ways:
* It prevents us from having to mention traces in every heading underneath, e.g. "When to scale traces" or "Stateful and stateless components traces components".
* It is more clear to the user what the whole of this section is about.
* It is easier to see how to extend the doc page with specifics for other telemetry signals - just add a "Metrics", "Logs", "Profiles" sections etc.

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
ptodev: It's a good idea, but because LB deployments span multiple Agents there will always be a need to explain to people that "this set of Agents does this", and "that set of Agents does that".

Modules are for cases when multiple components are composed within the same Agent. In the Traces LB case, we only really care about the `otelcol.exporter.loadbalancing` component since that's the only component present in all traces LB deployment scenarios.

When we port the "group by trace" processor we could make a module which contains it + `otelcol.exporter.loadbalancing`, but that feels like a very simple module and I'm not sure if it's worth it. But we can consider it in the future.

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
```diff
@@ -264,6 +267,95 @@ Name | Type | Description
...
+2. Spanmetrics could be generated in the backend database instead of the Agent.
+  * For example, span metrics can be [generated][tempo-spanmetrics] in Grafana Cloud by the Tempo traces database.
+
+[tempo-spanmetrics]: https://grafana.com/docs/tempo/latest/metrics-generator/span_metrics/
+
+### otelcol.connector.servicegraph
+Unfortunately, there is generally no reliable way to scale `otelcol.connector.servicegraph` over multiple Agent instance.
```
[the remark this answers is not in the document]
ptodev:
```suggestion
Scaling `otelcol.connector.servicegraph` over multiple Agent instances is challenging.
```
Do you mind if I replace the whole sentence like this?

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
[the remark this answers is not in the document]
ptodev: On Mimir we can do read queries which aggregate data from multiple Agent instances. However, the data will still have high cardinality because Mimir has to store the "agent_id" label. I don't think there is a way to reduce it unless people use adaptive metrics somehow.

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
ptodev: We could have step by step task based docs in addition to the current explanations. Or, one day we could just convert the current explanations to task based docs. But atm I just don't have the time to explain step by step all the possible scenarios in a task based doc.

docs/sources/flow/setup/deploy-agent.md
```diff
@@ -13,3 +13,58 @@ weight: 900
...
+Examples of stateful components:
+
+* `otelcol.processor.tail_sampling`
+* `otelcol.connector.spanmetrics`
+* `otelcol.connector.servicegraph`
+
+<!-- TODO: link to the otelcol.exporter.loadbalancing docs for more info -->
+
+A "stateless component" does not need to aggregate specific spans in order to work correctly - 
+it can work correctly even if it only has some of the spans of a trace.
```
clayton-cornell, line 61:
```suggestion
A "stateless component" does not need to aggregate specific spans to work correctly - 
it can work correctly even if it only has some of the spans of a trace.
```

docs/sources/flow/setup/deploy-agent.md
```diff
@@ -13,3 +13,58 @@ weight: 900
...
+
+Stateless Agents can be scaled without using `otelcol.exporter.loadbalancing`.
+You could use an off-the-shelf load balancer to, for example, do a round-robin load balancing.
```
clayton-cornell, line 64:
```suggestion
Stateless {{< param "PRODUCT_ROOT_NAME" >}} can be scaled without using `otelcol.exporter.loadbalancing`.
For example, you could use an off-the-shelf load balancer to do a round-robin load balancing.
```

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
```diff
@@ -264,6 +267,90 @@ Name | Type | Description
...
+
+### otelcol.connector.spanmetrics
+All spans for a given `service.name` must go to the same spanmetrics Agent.
+* This can be done by configuring `otelcol.exporter.loadbalancing` with `routing_key = "service"`.
+* If this is not done, metrics generated from spans might be incorrect.
+  * For example, if similar spans for the same `service.name` end up on different Agent instances,
+    the two Agents will have identical metric series for calculating span latency, errors, and number of requests.
+  * When both Agents attempt to write the metrics to a database such as Mimir, the series may clash with each other.
+  * At best this will lead to an error in the Agent and a rejected write to the metrics database. 
+  * At worst, it could lead to an inaccurate data due to overlapping samples for the metric series.
```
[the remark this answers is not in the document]
ptodev: This is a very dense and hard to understand subject. The reason I wrote them as bullets is that I felt it divides this information in bullets makes it easier to digest than having one big paragraph. But if you think this makes more sense from a writing point of view, I'm happy to change it.

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
```diff
@@ -264,6 +267,90 @@ Name | Type | Description
...
+  have a full view of the trace when making a sampling decision.
+  * For example, a `rate_limiting` tail sampling strategy may incorrectly pass through 
+    more spans than expected if the spans for the same trace are spread out to more than one Agent instance.
+
+<!-- Make "rate limiting" a URL to the tail sampler doc -->
+
+### otelcol.connector.spanmetrics
+All spans for a given `service.name` must go to the same spanmetrics Agent.
+* This can be done by configuring `otelcol.exporter.loadbalancing` with `routing_key = "service"`.
+* If this is not done, metrics generated from spans might be incorrect.
```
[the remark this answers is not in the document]
ptodev: Thanks, that's great feed back - I'll change the line to:

> If you do not configure `routing_key = "service"`, metrics generated from spans might be incorrect.

ptodev:
Hi, @clayton-cornell, thank you for the comments! I updated this PR following your feedback, and it is ready for review again.

docs/sources/flow/reference/components/otelcol.exporter.loadbalancing.md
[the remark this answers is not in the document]
clayton-cornell: Yeah I agree, it's a complex topic. The problem with the bullet points is that they should be "in agreement" or to put it another way, they should all follow the same flow.... which these do not. Instead of one big paragraph, try out the double LF which spaces out the sentences vs clumping it all int one big lump.

Either way will work... most important to get the info published without being overly pedantic on my side :-)

PR #5905, opened 2023-12-01
Support for NS records to discovery.dns

djcode:
#### PR Description
I added support for NS records to DNS discovery in Prometheus. https://github.com/prometheus/prometheus/pull/13219
This PR adds that support to the `discovery.dns` component.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [ ] Tests updated


ptodev (approved):

djcode:
> The change looks good, but I think we need to update our Prometheus dependency before merging it. Do you know when the next version of Prometheus with this change will be released?

That's a good point. I'm not sure when the next release will be nor if it will include my change. I'm happy to keep an eye on when the Prometheus change is properly released and update here when it happens.

djcode:
My upstream change was merged in 2.49.0.
I looked at how to update the Prometheus library inside Grafana Agent and it seems a bit more involved than I'd like for my little change. 

https://github.com/grafana/agent/blob/main/go.mod#L690 has been updated regularly since the beginning of my PR, so I'm happy to wait until that is version based off a Prometheus 2.49.0+ build and my change is the only thing that needs to be changed.

ptodev:
Hi, @djcode! I opened a [PR](https://github.com/grafana/alloy/pull/942) to add this feature to Alloy. Thank you for contributing it. The Alloy PR is exactly the same as yours. 

I am going to close this Agent PR because it might take a while for us to upgrade the Prometheus dependency in Grafana Agent. Even if we do, we are mostly focused on bugfixes with Agent releases now, and the NS DNS support won't take much priority unless someone mentioned that it's an issue for them. Please let me know if this is indeed a big issue for you and I'd be happy to reopen the Agent PR.

PR #5916, opened 2023-12-05
Fix windows tests that are failing CI on main

erikbaranowski:
#### PR Description

Fixes broken windows tests possibly coming from: https://github.com/grafana/agent/pull/5590 & https://github.com/grafana/agent/pull/5753

We will want to follow up on the second one which is disabled for windows in this PR to get CI working again.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/integration-tests/tests/scrape-prom-metrics/scrape_prom_metrics_test.go
+++ b/integration-tests/tests/scrape-prom-metrics/scrape_prom_metrics_test.go
@@ -1,3 +1,5 @@
+//go:build !windows
+
 package main
 
 import (
--- a/component/common/loki/client/internal/marker_file_handler.go
+++ b/component/common/loki/client/internal/marker_file_handler.go
@@ -16,8 +16,10 @@ const (
 	MarkerFolderName = "remote"
 	MarkerFileName   = "segment_marker"
 
-	MarkerFolderMode os.FileMode = 0o700
-	MarkerFileMode   os.FileMode = 0o600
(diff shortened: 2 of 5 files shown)


integration-tests/tests/scrape-prom-metrics/scrape_prom_metrics_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build !windows
```
wildum, line 1: The problem with the integration tests is that they cannot run in Drone. They only run in github Actions (because it uses docker compose which is not possible in Drone). I was planning to use test-containers (see https://github.com/grafana/agent/pull/5861) but now that we are moving away from Drone I will probably keep them in Github Actions. The reason why the tests are running is because the Windows pipeline is calling "go test" at the root of the pipeline and that triggers the integration tests (https://github.com/grafana/agent/blob/main/.drone/drone.yml#L184). They dont run when you run make test because I made a special exclusion (https://github.com/grafana/agent/blob/main/Makefile#L176). I wanted to do the same thing to fix the windows pipeline but I modified directly the yml file not knowing that it was a generated file (I just realised now) https://github.com/grafana/agent/pull/5643/files. Now Im not sure whether this fix actually work or if that's the right approach (@tpaschalis suggested also to use build tags but you need to make sure that the nothing runs in the integration tests folder)

wildum (approved):
Approval for the integration tests part

PR #5928, opened 2023-12-06
[draft] Demo how we could pass targets via receivers

thampiotr:
#### PR Description

This PR demonstrates how we could implement `TargetsReceiver` so that Targets can be sent between components in the same way that we currently send Logs, Metrics, Traces etc. 

This has a benefit of a more intuitive graph that more closely matches the data flow. The disadvantage is more complex code to handle components updates/shutdown. 

Below is a before/after of a simple example pipeline:

image

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


thampiotr:
This was a demo of the idea that is being discussed, but this PR will not be merged.

PR #5930, opened 2023-12-07
Add discovery.relabel_script to relabel with scripting

thampiotr:
#### PR Description
Adds a new `discovery.relabel_script` component that allows to relabel targets using a Python script (to be exact, it's a limited [Starlark dialect](https://github.com/google/starlark-go), that is used, for example, in Bazel).
* This allows to combine different discovery targets into one target, so we can, for example, enrich targets we discover in one place with extra labels discovered in another place if we join them in a script.
* It makes some target manipulations much simpler to read and understand.
* Thanks to multiline strings support in River, the config files look pretty decent. Could also add support for pulling a script from a file.
* Does not require users to learn some new (future) River scripting syntax, Python is a popular language many users will know
* Does not require the Agent to implement complex River features to provide more built-in functions and closures. However, it also does not prevent us from doing it in the future.

#### Notes to the Reviewer
* Docs are TODO
* With VSCode plugin it looks like this:
   image
   I may have old version that doesn't support multiline strings? But it also doesn't break entirely.

Some more info about [Starlark can be found here](https://github.com/bazelbuild/starlark/). The important part are Starlark's design principles, which I believe align well with our use case:
image

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/component/all/all.go
+++ b/component/all/all.go
@@ -26,6 +26,7 @@ import (
 	_ "github.com/grafana/agent/component/discovery/openstack"                      // Import discovery.openstack
 	_ "github.com/grafana/agent/component/discovery/puppetdb"                       // Import discovery.puppetdb
 	_ "github.com/grafana/agent/component/discovery/relabel"                        // Import discovery.relabel
+	_ "github.com/grafana/agent/component/discovery/relabel_script"                 // Import discovery.relabel_script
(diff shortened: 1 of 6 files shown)


thampiotr:
This is being discussed, but the PR will not be merged without more discussion or design, so I'm closing it for now.

PR #5933, opened 2023-12-07
chore(deps): remove minio dependency

domasx2:
#### PR Description

Turns out vendored minio dep used for faro receiver has AGPL license. This PR removes the dependency, instead copying in relevant code from older Apache licenesed version of minio.  

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/pkg/util/wildcard/match.go
+++ b/pkg/util/wildcard/match.go
@@ -1,0 +1,73 @@
+/*
+ * MinIO Cloud Storage, (C) 2015, 2016 MinIO, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package wildcard
+
+// MatchSimple - finds whether the text matches/satisfies the pattern string.
+// supports only '*' wildcard in the pattern.
+// considers a file system path as a flat name space.
+ func MatchSimple(pattern, name string) bool {
+	 if pattern == "" {
+		 return name == pattern
+	 }
+ 
+	 if pattern == "*" {
+		 return true
+	 }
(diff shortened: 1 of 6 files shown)


pkg/util/wildcard/match.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	// Does only wildcard '*' match.
+	return deepMatchRune([]rune(name), []rune(pattern), true)
+}
+
+// Match -  finds whether the text matches/satisfies the pattern string.
+// supports  '*' and '?' wildcards in the pattern string.
+// unlike path.Match(), considers a path as a flat name space while matching the pattern.
+// The difference is illustrated in the example here https://play.golang.org/p/Ega9qgD4Qz .
+func Match(pattern, name string) (matched bool) {
```
ptodev, line 38: The comment above talks about a "text", whereas the function's input variable is "name". It'd be nice if both the comment and the function are using the same name.

pkg/util/wildcard/match.go
```diff
@@ -0,0 +1,71 @@
...
+			if len(str) == 0 && !simple {
+				return false
+			}
+		case '*':
+			return deepMatchRune(str, pattern[1:], simple) ||
+				(len(str) > 0 && deepMatchRune(str[1:], pattern, simple))
+		}
+
+		str = str[1:]
+		pattern = pattern[1:]
```
ptodev, line 67: This seems a bit inefficient? I suppose performance is not important in this case?

pkg/util/wildcard/match_test.go
```diff
@@ -0,0 +1,542 @@
...
+			text:    "my-bucket/mnopqanda",
+			matched: false,
+		},
+	}
+	// Iterating over the test cases, call the function under test and asert the output.
+	for i, testCase := range testCases {
+		actualResult := wildcard.Match(testCase.pattern, testCase.text)
+		if testCase.matched != actualResult {
+			t.Errorf("Test %d: Expected the result to be `%v`, but instead found it to be `%v`", i+1, testCase.matched, actualResult)
+		}
```
ptodev, line 377:
```suggestion
		t.Run(fmt.Sprintf("%d", i+1), func(t *testing.T) {
			actualResult := wildcard.Match(testCase.pattern, testCase.text)
			require.Equal(t, actualResult, testCase.matched)
		})
```
It's a personal preference, but I'd prefer something like this.

ptodev (approved):
Thanks! I know it's just copied code, but I couldn't help but put in some comments :) I'm happy to just merge it though.

PR #5943, opened 2023-12-09
feat(pkg/operator): sort endpoints in reconciler for better perf

### LINKED ISSUE #5934: operator: kubelet endpoints are non-deterministic
### What's wrong?

https://github.com/grafana/agent/blob/99aba52da1b64250c992c7eb176a4d6d24919fb9/pkg/operator/kubelet.go#L108

In this code, we generate Endpoints based on a list of nodes. The node order is non-deterministic, so we end up with flip-flopping endpoints. This in turn has some cost in the API server, and possible other things watching endpoints

### Steps to reproduce

```
for i in {0..100}; do kubectl label node NODE --overwrite a=$i; done
```

is an easy way to trigger a bunch of reconciles, although it happens on its own naturally (I think every 5min from resync, but not certain; could also be node heartbeats, etc).

Then you can `kubectl get endpoints kubelet -w` and see it updates

### System information

_No response_

### Software version

v0.38.1

### Configuration

_No response_

### Logs

_No response_

hainenber:
#### PR Description
Sort Kubelet endpoints in agent operator to reduce resource cost on API server and endpoint watchers

Fixes #5934 

#### PR Checklist

- [x] CHANGELOG.md updated


tpaschalis (approved):
LGTM; leaving to @captncraig to see if he has any more input so we can close the original issue.

Thank you @hainenber for the continued efforts to improving the Agent 🙌

PR #5945, opened 2023-12-09
feat!(exporter/blackbox): use required name attr instead label for target block

### LINKED ISSUE #5940: Allow prometheus.exporter.blackbox to provide target name via a string instead of river identifier
### Request

The current implementation of `prometheus.exporter.blackbox` uses the target block label as the target name as documented [here](https://grafana.com/docs/agent/next/flow/reference/components/prometheus.exporter.blackbox/#target-block). For example,

```river
prometheus.exporter.blackbox "example" {
  config_file = "blackbox_modules.yml"

  target "grafana" {
    address = "http://grafana.com"
    module  = "http_2xx"
    labels = {
      "env": "dev",
    }
  }
}
```

Instead, use an argument inside the block for less restrictive naming. For example,

```river
prometheus.exporter.blackbox "example" {
  config_file = "blackbox_modules.yml"

  target {
    name     = "grafana"
    address = "http://grafana.com"
    module  = "http_2xx"
    labels = {
      "env": "dev",
    }
  }
}
```

### Use case

River identifiers are fairly restrictive. https://grafana.com/docs/agent/next/flow/config-language/syntax/#identifiers

This prevents someone from having an otherwise valid target name like `job-name` and is a stumbling block when converting from static mode to flow mode since static mode does not have this restriction.

hainenber:

Fixes #5940 

#### Notes to the Reviewer
Should this be a breaking change? I gave it some thoughts and unless we specify an interface to allow interchangeability (which is awkward, IMO), breaking change it is since `label` block is mandatory atm.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ internal API changes are not present.
 Main (unreleased)
 -----------------
 
+### Breaking changes
+
+- `target` block in `prometheus.exporter.blackbox` requires mandatory `name`
+  attribute instead of label block previously. (@hainenber)
+
 ### Enhancements
 
 - Flow Windows service: Support environment variables. (@jkroepke)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,8 @@ Main (unreleased)
 
 ### Breaking changes
 
-- `target` block in `prometheus.exporter.blackbox` requires mandatory `name`
-  attribute instead of label block previously. (@hainenber)
+- The `target` block in `prometheus.exporter.blackbox` requires a mandatory `name`
+  argument instead of a block label. (@hainenber)
 
 ### Enhancements
 
--- a/docs/developer/writing-exporter-flow-components.md
+++ b/docs/developer/writing-exporter-flow-components.md
@@ -40,7 +40,7 @@ prometheus.exporter.blackbox "example" {
 	config_file = "blackbox_modules.yml"
 	
 	target {
-    name    = "example"
+		name    = "example"
 		address = "http://example.com"
 		module  = "http_2xx"
 	}
--- a/docs/developer/writing-exporter-flow-components.md
+++ b/docs/developer/writing-exporter-flow-components.md
@@ -39,7 +39,8 @@ The river config would look like this:
 prometheus.exporter.blackbox "example" { 
 	config_file = "blackbox_modules.yml"
 	
-	target "example" {
+	target {
+    name    = "example"
 		address = "http://example.com"
 		module  = "http_2xx"
 	}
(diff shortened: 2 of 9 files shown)


rfratto:
I'm on PTO so I shouldn't be on GitHub, but I saw the recommendation for a breaking change in #5940 and wanted to offer an alternative: 

Another way of doing this change that wouldn't be a breaking change is to have an optional attribute to override the target name, similar to the `job_name` attribute in prometheus.scrape:

```river 
target "my_target" { // default name: my_target 
  name = "my-target-with-custom-characters" // override default 
  ... 
}
```

cc @erikbaranowski

hainenber:
Oh that's a good alternative right there. But I couldn't help but wonder if this is some kind of debt that can be cleared before v1.0.0 😄 

If Erik's cool with this, I'm cool as well.

CHANGELOG.md
```diff
@@ -10,6 +10,11 @@ internal API changes are not present.
 Main (unreleased)
 -----------------
 
+### Breaking changes
+
+- `target` block in `prometheus.exporter.blackbox` requires mandatory `name`
+  attribute instead of label block previously. (@hainenber)
```
ptodev, line 16:
```suggestion
- The `target` block in `prometheus.exporter.blackbox` requires a mandatory `name`
  argument instead of a block label. (@hainenber)
```

ptodev (commented):
If we go ahead with a breaking change, it will need to be documented in [docs/sources/flow/release-notes.md](https://github.com/grafana/agent/blob/main/docs/sources/flow/release-notes.md), which corresponds to [Release notes for Grafana Agent Flow](https://grafana.com/docs/agent/latest/flow/release-notes/).

My personal preference would be to just do it as a breaking change, because this seems very minor. However, we should get some team consensus on this before proceeding with a specific approach.

docs/developer/writing-exporter-flow-components.md
```diff
@@ -39,7 +39,8 @@ The river config would look like this:
 prometheus.exporter.blackbox "example" { 
 	config_file = "blackbox_modules.yml"
 	
-	target "example" {
+	target {
+    name    = "example"
```
erikbaranowski, line 43:
```suggestion
		name    = "example"
```

erikbaranowski (approved):
looks great, thanks for taking care of this!

I'll have a follow up PR likely to address some similar components we have identified.

erikbaranowski:
My preference is to go with the breaking change before 1.0 instead of leaving a potential stumbling block behind. If someone uses a label they may waste time figuring out why `job-name` is invalid.

erikbaranowski:
I'm poking at the go test issue to see what is going on

erikbaranowski:
A totally unrelated test had failed in the CI but I can't recreate it locally. I restarted the CI and hopefully it doesn't keep happening.

PR #5957, opened 2023-12-11
update restrictive label arguments to attributes

### LINKED ISSUE #5940: Allow prometheus.exporter.blackbox to provide target name via a string instead of river identifier

erikbaranowski:
…bute

Fixes #5940 

#### PR Checklist

- [ ] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/component/prometheus/exporter/blackbox/blackbox.go
+++ b/component/prometheus/exporter/blackbox/blackbox.go
@@ -67,7 +67,7 @@ var DefaultArguments = Arguments{
 
 // BlackboxTarget defines a target to be used by the exporter.
 type BlackboxTarget struct {
-	Name   string            `river:",label"`
+	Name   string            `river:"name,attr"`
 	Target string            `river:"address,attr"`
 	Module string            `river:"module,attr,optional"`
 	Labels map[string]string `river:"labels,attr,optional"`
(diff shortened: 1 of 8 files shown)


erikbaranowski:
a PR was already opened, I will work with the author

Issue #3244, opened 2023-03-12
UI API fails when viewing component which exports a map block


rfratto:
@erikbaranowski and I ran into an issue with #3227 where, if the exports map was set to a block instead of an attribute, it would fail to show in the UI.

The UI API currently [assumes](https://github.com/grafana/agent/blob/3afc90d370a45dacb93bdcb841e81801a7df272f/pkg/river/encoding/block.go#L41) that all blocks are structs, which is no longer true with #3232.

rfratto:
I'm reopening this one; even though grafana/river#41 got merged, the dependency hasn't been updated in this repo yet so the issue is still observable in main.

Issue #3844, opened 2023-05-10
River config file linter


petewall:
### Request

I often write my river files on one machine before deploying them to where the agent is running (especially when running the agent on Kubernetes, using a ConfigMap to store the config). I would like a way to check the config for syntax errors before deploying the config, only to find the agent doesn't start.

Ideally, I could envision something like:
```
grafana-agent config lint my-config.river
```
which would do a lightweight syntax check and any other offline checking. And a:

```
grafana-agent config validate my-config.river
```
which could do a more thorough test (e.g. `local.file` components actually point to a file on the filesystem)

### Use case

This would improve the delivery speed and quality of my config files.

mattdurham:
There is https://grafana.com/docs/agent/latest/flow/reference/cli/fmt/ does that suffice for the basic needs?

rfratto:
The challenge with this is to what level things should be checked, and whether it's even possible for an editor to correctly validate things before a config file is deployed. 

For example, let's imagine this config file:

```
local.file "example" {
  filename = "/etc/data.json"
}

prometheus.remote_write "example" {
  external_labels = json_decode(local.file.example.content) 
}
```

To what degree should a validator be able to assert this config file is correct? Does it need to validate `json_decode(local.file.example.content)`? If so, the validator needs to have access to `/etc/data.json` for the validation to succeed, but that file may only exist on the target environment. 

I see a few different tiers of validation here:

* Validate language syntax only (`grafana-agent fmt` does this today)
* Validate that component names, block names, and argument names are valid (the proposed `grafana-agent lint`)  
* Validate that expressions are valid (the proposed `grafana-agent validate`) 

Because `grafana-agent validate` requires constructing the component graph and doing a full evaluation, and also requires being checked in the same environment that agents run on, I'm skeptical whether it will provide as much value as it seems. 

IMO, `grafana-agent lint` is lower hanging fruit, and `grafana-agent validate` should ideally be unnecessary if the agent has powerful enough runtime debugging/reloading utilities to produce a working config file.

PR #5968, opened 2023-12-13
Module with declare and import

### LINKED ISSUE #5547: [RFC] Making modules more like components
> **NOTE**: For the most recent version of this proposal, see the RFC in [Google Docs](https://docs.google.com/document/d/1_DVeuoY89HcCUuI8At-WU5fx4uApgMOYRVm-5B4GbgQ). A snapshot of the proposal at the time of opening this issue is provided below for posterity. 

## Background 

[Modules](https://github.com/grafana/agent/blob/main/docs/rfcs/0007-flow-modules.md), first introduced in Grafana Agent v0.33.0, are a beta feature which allow users to encapsulate a Flow configuration file as a component. Today, somewhere between 8% and 11% of all Flow installations use at least one module. Modules are composed of two parts:

* The module content: A Flow config containing a pipeline to encapsulate as a component. Module content is published by module authors. 
* The module loader: A Flow component that retrieves module content from a specific origin (Git repository, HTTP, etc.) and runs it as a pipeline. Module loaders are configured by module consumers.  

The current implementation of modules has led to issues for both module authors and module consumers: 

* Module consumers struggle with:
  * Ambiguous terminology: Module loaders use "arguments" to refer to both configuring the module content and configuring the module loader, which can be hard to follow for new users.
  * Module loaders are repetitive: The same module loader arguments must be repeated each time a module from the same origin is used.
  
* Module authors struggle with: 
  * Complex composability: Module authors are required to [propagate module loader settings down to inner modules](https://github.com/grafana/agent-modules/blob/a02c9e370e994247a5d974369e1017bb91a57f3c/modules/kubernetes/logs/all.river#L55-L68) to compose smaller modules into larger ones.

This document proposes options to iterate on Flow modules to address these problems.

## Goals

* Make modules easier to understand for new users. 
* Reduce repetition for importing multiple modules from the same origin.
* Make it easier to compose smaller modules into larger modules. 
* Retain the ability for modules to be used for remote configuration. 
* Retain the ability to load a dynamic set of modules. 

### Allow Flow configs to declare components
(Credit to @thampiotr for this idea)

In this proposal, Flow configs will be allowed to declare components, and users will be able to import components from other configs. The following new config blocks will be introduced: 

* `declare`: A labeled block which declares a new component, where the component name is taken from the block label.  
* `import.*`: A labeled block which imports declared components from a module origin to the namespace specified by the block label. Imported components are then usable as normal components. 
  * Represented as `import.string`, `import.file`, `import.git`, `import.http`.

#### Example

```
--- module.river --- 

// Declare a new component called "add."
declare "add" {
	argument "a" { }
	argument "b" { }

	export "sum" {
		value = argument.a.value + argument.b.value 
	}
}

// Declare a new component called "subtract." 
declare "subtract" {
	argument "a" { }
	arugment "b" { }

	export "difference" {
		value = argument.a.value - argument.b.value
	}
}

--- config.river ---

// Import components from our module into the namespace "my_module".
import.file "my_module" { 
path = "./module.river"
}

// Use the "add" component from my_module.
my_module.add "example" {
	a = 5 
	b = 6 
}

// my_module.add.example.sum == 11 
```  

##### Declare 

The `declare` block is a labeled block which declares a new component. The name of the component is taken from the block's label. The content of the block is what would traditionally compose a module source today: argument blocks, export blocks, and components forming a pipeline. 

Expressions inside a declare block may reference components outside of that declare block:

```river 
discovery.kubernetes "pods" {
	role = "pod"
}

declare "scrape_pods" {
	argument "forward_to" {} 

	prometheus.scrape "default" {
		targets    = discovery.kubernetes.pods.targets 
		forward_to = argument.forward_to.value 
	}
}
```

A config can use components that it declares. When thai is done, there is no namespace for the declared components: 

```river
declare "add" {
	argument "a" { }
	argument "b" { }

	export "sum" {
		value = argument.a.value + argument.b.value 
	}
}

add "example" {
	a = 15 
	b = 17
}

// add.example.sum == 32 
```

##### Import

The `import` block is a labeled block which retrieves a module source from a specific origin and loads declared components from it. Declared components are imported into the namespace as specified by the import block's label. The namespace must be unique across the module, even across imports of a different kind. 

Only components declared directly in the dependency are exposed as components; components from transitive imports are not exposed. 

For backwards compatibility with existing modules, the import config blocks accept an arguments block to give to the module source and may export custom values from the module source. Additionally, imported modules do not have to declare components, allowing importing a module to have side effects, such as starting its own pipelines. 

If the import namespace collides with an existing namespace, the existing namespace is shadowed by the import. This keeps forward compatibility in case new namespaces are introduced which collide with an import. 

The `import.file` and `import.git` config blocks will support folders. The behavior of folders matches the behavior of passing a folder to `grafana-agent run`: all `.river` files in that folder (but not subfolders) are loaded as a single translation unit. 

Folder support improves component composability, as a folder can be used to define a set of components which are then composed together into a larger exposed component. 

##### Flow controller 

The Flow controller needs to be updated to properly understand that components are specified dynamically, and that the schema of imported components depends on the import block being evaluated. 

It is highly recommended to implement declare and import as config blocks and not components to minimize API changes and disallow components from defining other components. 

##### Support for dynamic modules

New `import` blocks can be created in the future to import a dynamic set of components, such as the Kubernetes capabilities proposed by [#5477](https://github.com/grafana/agent/issues/5477) and [#3598](https://github.com/grafana/agent/issues/3598).  

#### Pros/Cons

* Pros:
  * Aligns components and modules more closely: modules expose new actual components.
  * No repetition of module loader configuration when using two components from the same origin, since loading modules and using modules are now decoupled.
  * Improves composability: one file can declare multiple components and build them into larger components.
  * Has backwards compatibility with existing modules. 
  * As a bonus, it's possible to make this architecture extend to [plugins](https://github.com/grafana/agent/issues/5225), which are also intended to import new components. 

* Cons:
  * More work. 
  * Magic behavior around declare and import that need to be documented. 
  * Encourages module authors to design/structure modules differently than today to take advantage of the new capabilities. 

### Keep module architecture the same

Instead of making changes to how modules work, we will continue on with the current implementation of modules. 

With this proposal, we will add onto the design of modules organically to address problems as they arise, such as implementing some or all of these items:

* [Proposal: Introduce submodule component](https://github.com/grafana/agent/issues/5383)
* [Support specifying module filenames relative to current file](https://github.com/grafana/agent/issues/5320)
* Add a git service to reduce the total number of Git clones. 

#### Pros/Cons

* Pros:
  * Smaller blast radius.
  * Potentially less work.  

* Cons:
  * Doesn't immediately address the problems as defined in this document. 
  * Problems like "Module loaders are repetitive" can't be fully solved by additive additions to the current module loader architecture.

wildum:
New modules with import and declare keywords.

Old modules are left untouched by the changes.

Fixes #5547

#### Note for reviewers

This PR is pretty big and not easy to review but we wanted to have most of the functionalities on one branch to be sure that the approach was the correct one. Follow-up PRs will be made to handle nits, missing functionalities, and refactors which are no blockers for the PR.

#### Description

A description of the functionality can be found [here](https://github.com/grafana/agent/issues/5547).

Most relevant code changes:

- ComponentNode has been renamed BuiltinComponentNode
- git logic of the module.git has been moved to a util package to be used by new modules
- New nodes have been introduced:
  - Declare node: holds a static description of a module that has a string
  - Import node: retrieve content from a source that can be defined (HTTP, git, file). The content should contain only import and/or declare blocks. When an import node imports content containing import blocks, it will create importNode children. These children are NOT part of the graph, they are evaluated and they run within their import node parent. When a child has new content, it will notify its parent recursively till the new content reaches the root import node. The root importConfigNode holds all the declare definitions from its children and is responsible for notifying its dependants via the graph for re-evaluation.
  - Custom component node: manages a module. The content of the module will be given by an import node or a declare node.
-Previously only componentNodes could have dependants. Now customComponentNode and importConfigNodes can also have dependants. They implement the interface NodeWithDependants.
- The interface ComponentNode is implemented by the BuiltInComponentNode and the CustomComponentNode to handle both types together in the loader
- Another new interface is the ComponentInfo interface which is used by the flow_components.go to provide information on components (previously this was restricted to the struct ComponentNode). This allows other nodes like declareComponentNode and importConfigNode to appear in the UI.
- I added a few tests via module_declare_test.go, module_import_test.go and the integration tests. This is a start, we probably want to add more, especially on the error-handling part. But this could come via a separate PR.
- The old module code is left mostly untouched. I don't expect this PR to bring any harm to existing configs.

#### TODO (possibly in following PRs)

- import string
- import directory for import file and import git
- more/better testing
- refactor the integration tests to reduce redundancy
- explore the possibility to add a Content function to the BlockStmt in the river repo to avoid redundancy in the Declare struct
- docs
- the agent panics when the user clicks on the info button in the UI of a custom component because "river/encoding/riverjson: can only encode struct values to bodies, got map". This is due to the fact that the new modules use the map directly and not the export type. It seems to refer to "https://github.com/grafana/agent/issues/3244"

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated


tpaschalis:
Hey there 👋  Looks like this is going to require a new cup of coffee for the day 😅 

I realize it might not be a simple task to split this in logical increments, but could we offer some guidance here to help people who haven't followed development take a closer look?

* What are the high-level parts to review?
* Anything to pay special attention to or big ideas to keep in mind?
* Any areas you think warrant a closer look so we don't miss anything crucial?
* Any starting points for people wanting to try this out?

mattdurham:
Added this to my todo today.

pkg/flow/internal/controller/module_references.go
```diff
@@ -0,0 +1,77 @@
...
+	importNodes map[string]*ImportConfigNode,
+	declareNodes map[string]*DeclareNode,
+	uniqueReferences map[string]ModuleReference,
+	parentModuleDefinitions map[string]string,
+) error {
+
+	node, err := parser.ParseFile("", []byte(content))
+	if err != nil {
+		return err
+	}
```
rfratto, line 51: I'm very uneasy with how much code is related to passing around strings and parsing them later. I would vastly prefer us to change things so that we passed around parsed AST blocks.

pkg/flow/internal/controller/node_config_import.go
```diff
@@ -0,0 +1,396 @@
...
+	lastUpdateTime     atomic.Time
+
+	healthMut  sync.RWMutex
+	evalHealth component.Health // Health of the last evaluate
+	runHealth  component.Health // Health of running the component
+}
+
+var _ NodeWithDependants = (*ImportConfigNode)(nil)
+var _ RunnableNode = (*ImportConfigNode)(nil)
+var _ ComponentNode = (*ImportConfigNode)(nil)
```
rfratto, line 51: It feels off to me that ImportConfigNode implements the ComponentNode interface; it's not a component.

pkg/flow/internal/controller/node_declare_component.go
```diff
@@ -0,0 +1,361 @@
...
+	"github.com/grafana/agent/component/module"
+	"github.com/grafana/agent/pkg/flow/logging/level"
+	"github.com/grafana/agent/pkg/flow/tracing"
+	"github.com/grafana/river/ast"
+	"github.com/grafana/river/vm"
+	"github.com/prometheus/client_golang/prometheus"
+	"go.uber.org/atomic"
+)
+
+// DeclareComponentNode is a controller node which manages a module.
```
rfratto, line 24: > manages a module

I think our terminology here (across the set of maintainers, not just you and me) is misaligned, and we should spend a little bit of time to fix that. 

Here's my current view of things:

* Module: a unit of configuration.
* Root module: the configuration directly loaded by `grafana-agent run`. 
* Builtin component: a component provided by the Grafana Agent binary.
* Declare: a configuration block which declares a new component. 
* Declared component: a component provided by a declare block.
* Import: a configuration block which imports component definitions from another module. It is not valid to import a module which contains blocks other than import and declare. 

This is just what I'm suggesting; I think we need to have a offline conversation to get aligned on all the terminology, and then make sure all the comments and type names are aligned with that.

pkg/flow/internal/controller/node_declare_component.go
```diff
@@ -0,0 +1,361 @@
...
+	// If this is a local declare.
+	importLabel := ""
+	declareLabel := parts[0]
+	// If this is an imported module.
+	if len(parts) > 1 {
+		importLabel = parts[0]
+		declareLabel = parts[1]
+	}
+	return importLabel, declareLabel
+}
```
rfratto, line 77: This is making an assumption about the name of the component indicating its origin, and I don't think that'll hold true forever; it's likely that in the near future we'll be asked to support declaring components with `.` in the name so they feel even more like native components. 

Can we make this code more flexible for future-proofing or find a way to avoid it entirely?

pkg/flow/internal/controller/node_declare_component.go
```diff
@@ -0,0 +1,361 @@
...
+func (cn *DeclareComponentNode) ModuleIDs() []string {
+	return cn.moduleController.ModuleIDs()
+}
+
+// BlockName returns the name of the block.
+func (cn *DeclareComponentNode) BlockName() string {
+	return cn.componentName
+}
+
+// This node does not manage any component.
```
rfratto, line 358: For consistency, I wish it did; it feels important that anything claiming to be a ComponentNode has a non-nil return value for its managed component instance. 

It's not a difficult interface to implmement (only two methods) so it might be worth doing here. 

Alternatively, we can change the interface to remove this method from component node and make it unique to native components. However, I would still prefer that this return a non-nil value.

pkg/flow/internal/controller/node_native_component.go
```diff
@@ -90,7 +90,7 @@ type ComponentNode struct {
 	registry          *prometheus.Registry
 	exportsType       reflect.Type
 	moduleController  ModuleController
-	OnComponentUpdate func(cn *ComponentNode) // Informs controller that we need to reevaluate
+	OnComponentUpdate func(cn NodeWithDependants) // Informs controller that we need to reevaluate
```
rfratto, line 93: It feels off to me that we're not passing the ComponentNode interface to a function called OnComponentUpdate. Can we change the function name or the type passed to the function so they align?

pkg/flow/internal/controller/node_native_component.go
```diff
@@ -111,11 +111,12 @@ type ComponentNode struct {
...
 }
 
-var _ BlockNode = (*ComponentNode)(nil)
+var _ NodeWithDependants = (*NativeComponentNode)(nil)
+var _ ComponentNode = (*NativeComponentNode)(nil)
 
 // NewComponentNode creates a new ComponentNode from an initial ast.BlockStmt.
 // The underlying managed component isn't created until Evaluate is called.
-func NewComponentNode(globals ComponentGlobals, reg component.Registration, b *ast.BlockStmt) *ComponentNode {
+func NewComponentNode(globals ComponentGlobals, reg component.Registration, b *ast.BlockStmt) *NativeComponentNode {
```
rfratto, line 119: The function name here should be updated to NewNativeComponentNode.

pkg/flow/internal/import-source/import_file.go
rfratto, line 1: It's unidiomatic for the folder name to not match the package name; this should be `pkg/flow/internal/importsource/*.go`.

pkg/flow/internal/import-source/import_source.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/grafana/agent/component"
+	"github.com/grafana/river/vm"
+)
+
+type SourceType int
+
+const (
+	FILE SourceType = iota
+	HTTP
+	GIT
```
rfratto, line 16: It's unidiomatic for constants to be all uppercase in Go; these should be `File`, `HTTP`, and `Git` respectively. (There's an exception for HTTP as it's an acronym).

pkg/util/git/auth.go
rfratto, line 1: Two comments:

1. Can we put this in `internal/` instead of `pkg/util`? I don't think this need to be a part of the agent's public API. 
2. Can you keep the folder name matching the package name rather than renaming it to git? i.e., this should be `internal/vcs` and not `internal/git`

pkg/flow/internal/controller/module_info.go
rfratto, line 1: I find this file to be the hardest for me to understand across the whole PR:

1. It's not clear to me what "ModuleInfo" is, and whether we have the same definition of module (see my other comment for the need for us to have the maintainers aligned on a set of definitions). Why does ModuleInfo encapsulate both content and definitions? 
2. Every function here takes a lot of arguments, and it's not entirely clear from reading the code what they should be. 

I think we need to invest more effort in making the utility of this code clearer for readers, and introduce a minimal set types/concepts/etc as needed so it's easier to fit inside a reader's mental model of the controller.

rfratto (commented):
First off, thank you for working on this! This is a lot closer to being mergeable compared to the first few efforts, but I still have some comments that I would like to see resolved, and I've left the first round of comments below. 

A brief summary of my findings:

* I'm concerned that everyone is working with a different set of definitions for what a new module is, and I think we need to have a chat offline as a group to get aligned on this. 

* I'm worried there's too many differences in between native/non-native components and local/non-local modules that is leaking into the Loader implementation. Refactoring to make these constructs behave more identically and remove specialized knowledge from the Loader is likely in order.  

* I find the amount of code added for passing around Flow config strings fairly alarming; we should make changes to convert those strings into ASTs as soon as we can and pass the AST around instead. 

More detail about the above (and some other small comments and nits) are in the comments below.

pkg/flow/internal/controller/node_config_import.go
wildum: A ComponentNode is not a component, it's a node that manages a component. ImportConfigNode manages a component to import the content

pkg/flow/internal/controller/node_declare_component.go
wildum: If we allow "." in the names, what happens when I declare a component "A.B" and I have an import statement "A" that imports a declare named "B"? Both should be instantiated via "A.B" according to the current design.

pkg/flow/internal/controller/node_native_component.go
wildum: good catch, missed it during the renaming

component/component_provider.go
```diff
@@ -93,8 +93,8 @@ type Info struct {
 	// this component depends on, or is depended on by, respectively.
 	References, ReferencedBy []string
 
-	Registration Registration // Component registration.
```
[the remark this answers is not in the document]
wildum: But the registration does not have the types? It has "Args Arguments" and "Exports Exports" which are already inside of the Info struct as "Arguments Arguments" and "Exports   Exports". Removing the Registration field removes this repetition and the "Build" function (which should not be there because retrieving info about a particular instance of a component to create a new component of the same type should not be a correct way to create a component)

pkg/flow/internal/controller/module_references.go
wildum: Good idea, in this [commit](https://github.com/grafana/agent/pull/5968/commits/8f50ca1839f9e806aa99f30610711ab3a84cf070) I changed the code to use the Declare type instead of the plain content string. This way I can re-use the AST when searching for module references within a declare block. I also think that it makes the code clearer

component/component_provider.go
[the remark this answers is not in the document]
rfratto: > But the registration does not have the types? 

The registration _does_ have the types; it holds the zero value for the arguments and exports types of a given component, and it's used for unmarshaling. In the future, those fields could also be used for additional validity checking without instantiating components (e.g., #3844). 

> (which should not be there because retrieving info about a particular instance of a component to create a new component of the same type should not be a correct way to create a component

I agree, but I also didn't suggest that the registration from this API would be the way to construct new components; just for the full set of information associated with a component. 

Let me think about this a bit. This is a public API, which I think we need to be cautious about changing. I need to consider what the future implications are for _not_ including the registration in this API, particularly around whether that limits us with the web UI or any other future services.

We should change APIs deliberately and not to make another feature easier. So if we're going to change this API, we should make sure it's the right change.

pkg/flow/internal/controller/module_references.go
wildum: The declares are passed as pointers to avoid copying the string. I don't check for nil pointers because in the case when this would happen (which is not possible now), then it would be ok to panic because there would have been a programming error that would have allowed it and it should be fixed (no point to recover).

pkg/flow/internal/controller/node_config_import.go
wildum: I will rename the ComponentNode interface implemented by NativeComponentNode / DeclareComponentNode / ImportConfigNode to something else (don't know what yet) and I will create a new interface ComponentNode that is only implemented by NativeComponentNode and DeclareComponentNode

rfratto: > A ComponentNode is not a component

I'm sorry, but this isn't true. The documentation of a ComponentNode states:

> ComponentNode is a generic representation of a Flow component. 

The original intent was that ComponentNodes are components in the graph, wrapped by some node-specific logic. A node can create and manage an instance of a component, but unless it's _acting_ as a component itself, I don't think it should implement ComponentNode.

rfratto: Following up on this: I'm wrong; I had a branch for experimenting with plugins where I _also_ created a ComponentNode interface (with the definition above), and I got confused about what's in main. I even looked at my local code, but I was still checked out to that experimental branch :facepalm: Sorry for the confusion here.

In any case, I think the definition I had come up with makes logical sense: something in the DAG which implements an interface used by component is intending to act as a component from the perspective of a DAG.

I'd be OK with finding a more appropriate name or restricting the interface to that definition.

component/module/module.go
```diff
@@ -16,20 +16,34 @@ type ModuleComponent struct {
...
+	}
+	var err error
+	c.mod, err = o.ModuleController.NewModule("", func(exports map[string]any) {
+		c.opts.OnStateChange(exports)
+	})
+	return c, err
+}
+
+// TODO: Remove when getting rid of old modules
+func NewModuleComponentDeprecated(o component.Options) (*ModuleComponent, error) {
```
tpaschalis, line 46: I'm not sure what the difference between `NewModuleComponent/NewModuleComponentDeprecated` is, can we expand on this?


To minimize breaking changes to the API, the Go convention is to mark the current function with a deprecation notice in the docstring and add the new one under a different name. We're still pre v1.0, but it'd be a good exercise to figure out how you'd handle this instead.

component/module/module.go
```diff
@@ -43,12 +57,12 @@ func NewModuleComponent(o component.Options) (*ModuleComponent, error) {
 // LoadFlowSource loads the flow controller with the current component source.
 // It will set the component health in addition to return the error so that the consumer can rely on either or both.
 // If the content is the same as the last time it was successfully loaded, it will not be reloaded.
-func (c *ModuleComponent) LoadFlowSource(args map[string]any, contentValue string) error {
-	if reflect.DeepEqual(args, c.getLatestArgs()) && contentValue == c.getLatestContent() {
+func (c *ModuleComponent) LoadFlowSource(args map[string]any, contentValue string, parentDeclareContents map[string]string) error {
+	if reflect.DeepEqual(args, c.getLatestArgs()) && contentValue == c.getLatestContent() && reflect.DeepEqual(args, c.getLatestParentDeclareContents()) {
```
tpaschalis, line 61: I _believe_ that's what you meant to write here, right?
```suggestion
func (c *ModuleComponent) LoadFlowSource(args map[string]any, contentValue string, parentDeclareContents map[string]string) error {
	if reflect.DeepEqual(args, c.getLatestArgs()) && contentValue == c.getLatestContent() && reflect.DeepEqual(parentDeclareContents, c.getLatestParentDeclareContents()) {
```


As a note, one of the saving graces of this package was the abundant docstrings, documenting the author's intentions for the future readers. Upon reading this, I have zero clue what a parent is, what its contents are supposed to be, and whether it's stuff that the parent declares, or what the content refers to. This ties in to the discussion about LoadConfig and whether we _need_ to change the signature here.

pkg/flow/internal/controller/node_with_component.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"context"
+
+	"github.com/grafana/agent/component"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// NodeWithComponent is ?
+// TODO we need a better name
+type NodeWithComponent interface {
```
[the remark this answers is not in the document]
tpaschalis: Similar to the previous comment about deprecating, it's a little weird renaming this just for an extension interface to take its place, and _another_ similarly named interface `NodeWithDependants` to have nothing to do with it.

So far I've seen both
a) ComponentNode keeping its own semantics and continuing to be handled as one.
b) ComponentNode being substituted by NodeWithComponent
c) ComponentNode being substituted by NodeWithDependants
d) ComponentNode being substituted by BuiltinComponentNode

I don't mean to nitpick names, but I think it's genuinely confusing.

pkg/flow/internal/controller/component_node.go
```diff
@@ -0,0 +1,34 @@
+package controller
+
+import (
+	"github.com/grafana/agent/component"
+	"github.com/grafana/agent/pkg/flow/internal/dag"
+)
+
+type ComponentNode interface {
```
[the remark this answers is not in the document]
tpaschalis: As a nitpick, I'm personally not a fan of breaking up functionality into different files; it makes browsing and refactoring the codebase harder.

But could we make some more effort on here? I'm not sure why this is an extension interface, what it's use is, or what to expect from implementations.

pkg/flow/internal/controller/node_custom_component.go
```diff
@@ -0,0 +1,366 @@
...
+func (cn *CustomComponentNode) ModuleIDs() []string {
+	return cn.moduleController.ModuleIDs()
+}
+
+// BlockName returns the name of the block.
+func (cn *CustomComponentNode) BlockName() string {
+	return cn.componentName
+}
+
+// This node does not manage any component.
```
tpaschalis, line 358: Is that true here?

component/module/module.go
wildum: oof that's a good catch, thanks!  I will also make some changes to the doc and the signature

pkg/flow/internal/controller/node_custom_component.go
wildum: I had a different definition of module in mind and I thought that this node was managing a module and not a component. I plan on following what Robert said above and to have this function return something

component/module/module.go
wildum: I added the depretation notice and more doc

tpaschalis: Do we _need_ two separate functions here? They look identical to.

wildum: The callback passed to the NewModule is different: in the deprecated version it passes the Exports type while in the new version it passes the map[string]any. As a result, when using new modules you don't need to write ".exports" in you config. It makes it less verbose and it acts like the other components.
I decided to use two functions because:
- it does not touch the old logic which reduces the risk of regression for the old modules when we merge this
- when we will get rid of the old modules, we can just remove the deprecated part without having to touch the new modules logic (less risk of regression for the new modules)
But I'm also open to suggestions if you have another solution in mind

pkg/flow/internal/controller/loader.go
```diff
@@ -117,7 +127,7 @@ func NewLoader(opts LoaderOptions) *Loader {
 // The provided parentContext can be used to provide global variables and
 // functions to components. A child context will be constructed from the parent
 // to expose values of other components.
```
thampiotr, line 120: For code readability, the comments on `Apply` method should mention what is going on with `configBlocks`, `declares` and what is `parentModuleDefinitions` for.

pkg/flow/internal/controller/loader.go
```diff
@@ -252,9 +279,12 @@ func (l *Loader) Cleanup(stopWorkerPool bool) {
 }
 
 // loadNewGraph creates a new graph from the provided blocks and validates it.
-func (l *Loader) loadNewGraph(args map[string]any, componentBlocks []*ast.BlockStmt, configBlocks []*ast.BlockStmt) (dag.Graph, diag.Diagnostics) {
+func (l *Loader) loadNewGraph(args map[string]any, componentBlocks []*ast.BlockStmt, configBlocks []*ast.BlockStmt, declares []*Declare) (dag.Graph, diag.Diagnostics) {
 	var g dag.Graph
 
+	// Reset the module cache before loading a new graph. This step is crucial to ensure that references to outdated modules, not included in the new configuration, are removed.
+	l.moduleReferences = make(map[string][]ModuleReference)
```
thampiotr: The comment says "module cache", while the field is called `moduleReferences`. Let's make sure we get the naming consistent and update the comment when needed. I would also like to know what's the purpose of this map.

pkg/flow/internal/controller/loader.go
```diff
@@ -252,9 +279,12 @@ func (l *Loader) Cleanup(stopWorkerPool bool) {
...
+
 	// Split component blocks into blocks for components and services.
 	componentBlocks, serviceBlocks := l.splitComponentBlocks(componentBlocks)
```
thampiotr: I know this is not added by you, but this looks weird - the naming is not consistent here. It's like saying "split apples into apples and bananas".

pkg/flow/internal/controller/node_declare.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"sync"
+
+	"github.com/grafana/river/ast"
+	"github.com/grafana/river/vm"
+)
+
+type DeclareNode struct {
+	label         string
+	nodeID        string
```
thampiotr, line 12: Isn't nodeID redundant? it seems to be always `label + componentName`.

pkg/flow/internal/controller/node_declare.go
```diff
@@ -0,0 +1,51 @@
...
+	componentName string
+	declare       *Declare
+	mut           sync.RWMutex
```
thampiotr, line 15: `mut` should be grouped with the fields that it guards. For example, if `mut` is guarding `declare` then:

```
type DeclareNode struct {
	label         string
	nodeID        string
	componentName string

	mut           sync.RWMutex
	declare       *Declare
}
```

pkg/flow/internal/controller/node_declare.go
```diff
@@ -0,0 +1,51 @@
...
+		nodeID:        BlockComponentID(declare.Block).String(),
+		componentName: declare.Block.GetBlockName(),
+		declare:       declare,
+	}
+}
+
+func (cn *DeclareNode) Declare() *Declare {
+	cn.mut.Lock()
+	defer cn.mut.Unlock()
+	return cn.declare
```
thampiotr, line 33: Why do we need a mutex for Declare? 
If we do need it, why is it not inside the Declare that should encapsulate and guard the access to its own fields?

thampiotr: I spent a lot of time trying to see if the fields inside `*Declare` can change and thus need to be protected with a mutex. So far it seems to me that all the `Declare` structs are created inside `flow.ParseSource` and are not modified on updates. We must make it easier to read this code with better naming (or comments if we fail at naming) and remove any misleading / dead code.

pkg/flow/internal/controller/loader.go
```diff
@@ -464,17 +512,57 @@ func (l *Loader) populateComponentNodes(g *dag.Graph, componentBlocks []*ast.Blo
...
+			}
 		}
-
-		g.Add(c)
 	}
 
 	return diags
 }
 
+func (l *Loader) shouldAddDeclareComponentNode(firstPart, componentName string) bool {
```
thampiotr: This method's name describes more the context where it is called instead of what it does. It's better for methods to describe what they do. This one checks whether a component with given name has been defined via declare - either directly in this config or via imports or in a parent. We should call this something like `isCustomComponent` once we switch to the new naming conventions.

pkg/flow/internal/controller/loader.go
```diff
@@ -464,17 +512,57 @@ func (l *Loader) populateComponentNodes(g *dag.Graph, componentBlocks []*ast.Blo
...
+	_, declareExists := l.declareNodes[firstPart]
+	_, importExists := l.importNodes[firstPart]
+	_, moduleDepExists := l.parentModuleDefinitions[componentName]
```
thampiotr: Do we need this? I thought that we don't need to support instantiating components that were declared in parent modules?

thampiotr (commented):
I went through only part of this PR so far. I think we need to put more work into making this readable - I'm worried it will make maintenance in the future much harder. Let's start with adopting the new naming convention and adjusting the logic around nested imports as discussed recently. Let me know when it's ready to review further!

component/registry.go
```diff
@@ -44,7 +44,7 @@ type Module interface {
 	// LoadConfig parses River config and loads it into the Module.
 	// LoadConfig can be called multiple times, and called prior to
 	// [Module.Run].
-	LoadConfig(config []byte, args map[string]any) error
+	LoadConfig(config []byte, args map[string]any, moduleDefinitions map[string]string) error
```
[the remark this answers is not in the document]
wildum: I introduced a new type LoaderConfigOptions that can be used when loading a new config to pass specific options. In this case it is used to pass the additional declare contents but in the future it can be extended if needed.

component/module/module.go
wildum: I made some changes and introduced a new LoaderConfigOptions type. Hopefully, things are more clear now

component/module/module.go
```diff
@@ -43,12 +57,12 @@ func NewModuleComponent(o component.Options) (*ModuleComponent, error) {
 // LoadFlowSource loads the flow controller with the current component source.
 // It will set the component health in addition to return the error so that the consumer can rely on either or both.
 // If the content is the same as the last time it was successfully loaded, it will not be reloaded.
-func (c *ModuleComponent) LoadFlowSource(args map[string]any, contentValue string) error {
-	if reflect.DeepEqual(args, c.getLatestArgs()) && contentValue == c.getLatestContent() {
+func (c *ModuleComponent) LoadFlowSource(args map[string]any, contentValue string, parentModuleDefinitions map[string]string) error {
```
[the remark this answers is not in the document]
wildum: I introduced a new type LoaderConfigOptions which hopefully makes this more clear

pkg/flow/internal/controller/loader.go
wildum: It's now handled by the component_node_manager as customComponentDependencies. It's being initialized by FindCustomComponentDependencies (// FindCustomComponentDependencies traverses the AST of the provided declare and collects references to known custom components.). I can also add a comment on the field if needed

pkg/flow/internal/controller/loader.go
wildum: We do need it. One example (there are other scenarios):

```
import A {}

declare b { A.a {} }

b {}
```

In this case you need to pass the additional declare "A.a" to b.

component/module/module.go
```diff
@@ -16,20 +16,34 @@ type ModuleComponent struct {
...
-	latestArgs    map[string]any
+	mut                           sync.RWMutex
+	health                        component.Health
+	latestContent                 string
+	latestArgs                    map[string]any
+	latestParentModuleDefinitions map[string]string
 }
 
 // Exports holds values which are exported from the run module.
+// This export type is deprecated.
```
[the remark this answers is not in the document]
wildum: Done, as a side note I need to add a nolint when the type is used else the lint fails the build

pkg/flow/internal/controller/loader.go
thampiotr: Right. I guess you wouldn't really call `A.a` a parent module here though? 
Let's make sure that the naming convention we adopt works here without introducing more confusion.

wildum: In the example, the root controller imports the declare and passes it down to the child controller spawn by the custom component b. I think parent is ok but not so good because it can be confusing if you don't have the full context. That's why I remaned it to additionalDeclareContents which is provided via LoaderConfigOptions.

pkg/flow/internal/controller/loader.go
```diff
@@ -464,17 +512,57 @@ func (l *Loader) populateComponentNodes(g *dag.Graph, componentBlocks []*ast.Blo
 				})
 				continue
 			}
-
-			// Create a new component
-			c = NewComponentNode(l.globals, registration, block)
+			firstPart := strings.Split(componentName, ".")[0]
```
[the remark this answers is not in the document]
thampiotr: I think that the import label or declare label become a custom component's namespace. So you can maybe call this a `namespace` and then check if this namespace has been shadowed by an import or declare, as it is allowed.

integration-tests/tests/scrape-prom-metrics-module-git/scrape_prom_metrics_module_git_test.go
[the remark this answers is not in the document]
wildum: Yeah with scrape_prom_metrics_test and module_file as well. I copy pasted the tests to quickly check if the imports were behaving as expected. I would like to refactor these in a separate PR

pkg/flow/internal/controller/loader.go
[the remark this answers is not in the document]
wildum: ok namespace is good, the shadowing happenr via the order in findCustomComponentDependencies and getCustomComponentConfig

pkg/flow/internal/controller/node_declare_component.go
wildum: the code has been adapted to the terminology decided by the team

pkg/flow/internal/controller/node_declare.go
wildum: thanks for bringing my attention to this. This branch has been through a lot of changes with the different design and renaming and feedbacks, and while I'm trying to clean it up there might be artifacts of previous implementations or copy pasta that I need to clean. I will take some time tomorrow to go deeper into details (fields that can be private, fields that don't need mutex, areas where comments are missing etc...)

pkg/flow/internal/controller/node_declare.go
wildum: it's `componentName+"."+label`. This is common to every node and it is used for optimization (from builtin component node: "nodeID string // Cached from id.String() to avoid allocating new strings every time NodeID is called.")

pkg/flow/internal/controller/component_node.go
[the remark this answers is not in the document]
wildum: Added docs and improved naming. The interfaces are used to give specific roles to nodes so that they can be handled as groups. The current setup feels still far from perfect but I'm struggling to find something better without increasing the blast radius which is already important with this PR. I'm open to suggestions

pkg/flow/internal/controller/node_with_component.go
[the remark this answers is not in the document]
wildum: As explained by the comment in the other thread, I renamed and documented the interfaces. It's not perfect but it's very useful in the current setup to handle the nodes as groups. I can't find a better way without increasing the blast radius which is already big. I'm open to suggestions

pkg/flow/internal/controller/node_declare_component.go
wildum: Also I believe that rivers currently does not support "." in labels?

component/component_provider.go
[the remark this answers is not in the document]
wildum: In builtin component, the args and exports are initialized by the registration when the node is created. That means that they also have the types and that we don't need the registration object.
Also registration is currently something specific for the builtin components. It is not used by the custom components so that would require some workaround to keep it in the component_provider info

pkg/flow/internal/controller/node_config_import.go
```diff
@@ -0,0 +1,459 @@
...
+		if err != nil {
+			return fmt.Errorf("imported node %s failed to evaluate, %v", child.label, err)
+		}
+	}
+	return nil
+}
+
+// runChildren run the import nodes managed by this import node.
+// The children list can be updated onContentUpdate. In this case we need to stop the running children and run the new set of children.
+func (cn *ImportConfigNode) runChildren(parentCtx context.Context) error {
```
wildum, line 243: This function is not trivial, I added a check for leaky goroutines and it appears that every few test that one go routine inside of start children is leaky. Not sure whether it is a test error or if there is a real leak. I'd be happy to get feedback on how this area can be improved

PR #5970, opened 2023-12-13
Add code to check for staleness in labelstore

### LINKED ISSUE #5585: GlobalRefMap/LabelStore should check staleness markers and/or other garbage collections
### Request

GlobalRefMap can leak 16 bytes +  label per label. This can cause higher than expected memory usage in high cardinality environments.

### Use case

Reduce memory usage of global ref map. Add staleness checks, though this is likely not a full solution since honor timestamps and metrics receiver can prevent this.

mattdurham:
#### PR Description

This wires up the staleness check from the labelstore that will ideally reduce memory usage in high cardinality or high churn environments.

Closes #5585

#### PR Checklist

- [X] CHANGELOG.md updated
- [x] Tests updated


component/prometheus/interceptor.go
```diff
@@ -102,6 +103,13 @@ func (a *interceptappender) Append(ref storage.SeriesRef, l labels.Labels, t int
 		ref = storage.SeriesRef(a.ls.GetOrAddGlobalRefID(l))
 	}
 
+	if value.IsStaleNaN(v) {
+		a.ls.AddStaleMarker(uint64(ref), l)
+	} else {
+		// This feels excessive to call this each time.
+		a.ls.RemoveStaleMarker(uint64(ref))
```
mattdurham, line 110: I really really dislike this and wish we could batch the checking.

service/labelstore/service.go
```diff
@@ -6,22 +6,26 @@ import (
...
-	labelsHashToGlobal map[uint64]uint64
-	staleGlobals       map[uint64]*staleMarker
+	log                 log.Logger
+	mut                 sync.Mutex
+	globalRefID         uint64
+	mappings            map[string]*remoteWriteMapping
+	labelsHashToGlobal  map[uint64]uint64
+	staleGlobals        map[uint64]*staleMarker
+	totalIDs            *prometheus.Desc
+	idsInRemoteWrapping *prometheus.Desc
```
thampiotr, line 26: Could these be constants in the package?

service/labelstore/service.go
```diff
@@ -56,12 +69,33 @@ func (s *service) Definition() agent_service.Definition {
...
+	m <- s.idsInRemoteWrapping
+}
+func (s *service) Collect(m chan<- prometheus.Metric) {
+	s.mut.Lock()
+	defer s.mut.Unlock()
+
+	m <- prometheus.MustNewConstMetric(s.totalIDs, prometheus.CounterValue, float64(len(s.labelsHashToGlobal)))
+	for name, rw := range s.mappings {
+		m <- prometheus.MustNewConstMetric(s.idsInRemoteWrapping, prometheus.CounterValue, float64(len(rw.globalToLocal)), name)
+	}
```
thampiotr, line 83: Should these metrics be Gauge since they can go up and down?

component/prometheus/interceptor.go
```diff
@@ -102,6 +103,13 @@ func (a *interceptappender) Append(ref storage.SeriesRef, l labels.Labels, t int
...
+	}
```
thampiotr, line 111: Hmm, we did appear sometimes to run into bottleneck (WAL delay was growing) without hitting any obvious resource limitation. Could it be contention on locks like this? (of course not here, cause it's new, but similar calls to labelstore).

So I think there are two main questions I have here:
* Are there metrics that we could use to measure the overhead this could introduce?
* Could we test this branch out in dev first?

If we find there is a bottleneck here, we perhaps should consider having using a striped implementation of labelcache, where we have N locks and use ref hash to pick which lock is needed 🤔 increasing effectively number of possible concurrent operations on the labelstore. Would hopefully be able to find an existing library for it?

service/labelstore/service.go
```diff
@@ -56,12 +69,33 @@ func (s *service) Definition() agent_service.Definition {
...
+}
+
 // Run starts a Service. Run must block until the provided
 // context is canceled. Returning an error should be treated
 // as a fatal error for the Service.
 func (s *service) Run(ctx context.Context, host agent_service.Host) error {
-	<-ctx.Done()
-	return nil
+	staleCheck := time.NewTicker(10 * time.Minute)
```
thampiotr, line 90: We probably don't need to do this now, but do you think there may be a need to configure this in the future?

mattdurham: Since staleness markers arent configurable and this was meant to be double the staleness marker length I dont think it will need to be configurable. Though that is famous last words.

component/prometheus/interceptor.go
mattdurham: The locks do show up on profiles, I though about using a bucketed approach but hesitant to add more complexity to the store. I would be more pro seeing how this behaves and then adjusting again in a more focused PR. This should be less overhead than the other calls at least.

mattdurham: This ended up not having a significant impact on cpu after running it for 12 hours.

thampiotr: Amazing, thanks for testing!

thampiotr (approved):

component/prometheus/interceptor.go
mattdurham: Going to update my comment :) so its clear we tested it

PR #5845, opened 2023-11-23
Update shared content topics

clayton-cornell:
#### PR Description

This PR is the first of several focused on a cleanup of the Agent Flow component reference documentation.

This PR updates the shared content topics (Flow reference docs).  Changes include:
- Adding/removing punctuation
- Reformatting and alphabetising tables
- Removing some (not all) of the passive voice
- Correcting subject/verb agreement
- Correcting singular/plural
- Adding missing words or removing extra words (depending on the sentence)
- Reworking to use semantic line breaks instead of the arbitrary +/- 80 columns

NOT included:
- Fixing ALL cases of passive voice (small/no gain for high effort)
- Adding lead-in sentences. It's hit/miss on these topics if they have a description before the table of attributes. I left it as-is in this area.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated
- [ ] Config converters updated

PR #5972, opened 2023-12-13
Remove replace directive for golang.org/x/exp

### LINKED ISSUE #5921: Add `track_timestamps_staleness` to prometheus.scrape
### Request

The `track_timestamps_staleness` feature was [added](https://github.com/prometheus/prometheus/pull/13060) to Prometheus recently. To pick it up, we would need to upgrade to Prometheus v2.48 or above.

`track_timestamps_staleness` is `false` by default in Prometheus, because it is considered a breaking change. We should consider whether it could be set to `true` in the Agent by default. For example, does setting this to `false` have any benefits? Are there any implications to alerts which rely on these metrics, and would that be s serious breaking change to users?

Also, we need to go into more detail in the docs about what metrics with explicit timestamps are, and what staleness markers are, so that users can make a decision how to configure `track_timestamps_staleness`. This is what the [Prometheus docs](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config) say for `track_timestamps_staleness`:

```
# track_timestamps_staleness controls whether Prometheus tracks staleness of
# the metrics that have an explicit timestamps present in scraped data.
#
# If track_timestamps_staleness is set to "true", a staleness marker will be
# inserted in the TSDB when a metric is no longer present or the target
# is down.
[ track_timestamps_staleness: <boolean>  | default = false ]
```

I personally didn't even know there could be "metrics that have an explicit timestamps present in scraped data". It is also not clear to me what the implications of adding a staleness marker are.

### Use case

Setting `track_timestamps_staleness` to `true` could fix a bug where container metrics linger for 5 minutes after a container dies.

ptodev:
Removing a replace directive for `golang.org/x/exp` form the Agent's go.mod file.

This is necessary because on a separate branch I am upgrading Agent to a new OpenTelemetry version, and it requires a new version of `github.com/grafana/loki/pkg/push` which needs the latest `golang.org/x/exp`.

The reason why `golang.org/x/exp` has been problematic is because the return type of `SortFunc` changed from `bool` (in the [old version](https://pkg.go.dev/golang.org/x/exp@v0.0.0-20230711153332-06a737ee72cb/slices#SortFunc)) to `int` (in the [new version](https://pkg.go.dev/golang.org/x/exp@v0.0.0-20231206192017-f3f8817b8deb/slices#SortFunc)). Apparently some packages like to use `golang.org/x/exp` because it's more performant.

Fixes #5921


ptodev:
The `replace` directive for `prometheus/prometheus` can be removed, because [v2.48.0](https://github.com/prometheus/prometheus/releases/tag/v2.48.0) contains the two PRs we need:
https://github.com/prometheus/prometheus/pull/12677
https://github.com/prometheus/prometheus/pull/12729

mattdurham (approved):
Looks like we got rid of a lot of cruft! Awesome

ptodev:
@mattdurham unfortunately the Linux build doesn't work because Pyroscope's eBPF module has a [replace directive](https://github.com/grafana/pyroscope/blob/main/ebpf/go.mod#L51-L54) for the `exp` module 🙀  I'll try to change their go.mod.

ptodev:
I raised a [PR](https://github.com/grafana/pyroscope/pull/2848) for Pyroscope.

ptodev:
I had to upgrade our Prometheus dependency, and this became a much bigger PR than expected.

docs/sources/flow/reference/components/prometheus.scrape.md
```diff
@@ -44,30 +44,31 @@ The following arguments are supported:
...
-`label_name_length_limit`  | `uint`     | More than this label name length post metric-relabeling causes the scrape to fail. | | no
-`label_value_length_limit` | `uint`     | More than this label value length post metric-relabeling causes the scrape to fail. | | no
-`bearer_token` | `secret` | Bearer token to authenticate with. | | no
-`bearer_token_file` | `string` | File containing a bearer token to authenticate with. | | no
-`proxy_url` | `string` | HTTP proxy to proxy requests through. | | no
-`follow_redirects` | `bool` | Whether redirects returned by the server should be followed. | `true` | no
-`enable_http2` | `bool` | Whether HTTP2 is supported for requests. | `true` | no
+`honor_labels`                | `bool`     | Indicator whether the scraped metrics should remain unmodified. | `false` | no
+`honor_timestamps`            | `bool`     | Indicator whether the scraped timestamps should be respected. | `true` | no
+`track_timestamps_staleness`  | `bool`     | Indicator whether to track the staleness of the scraped timestamps. | `false` | no
```
ptodev, line 54: @bboreham Regarding #5921 - for now I intend to default this to `false` for a few reasons:
* IIUC, `track_timestamps_staleness = true` only makes sense if out of order ingestion is allowed in the back end database.
* Apparently there are implications to querying and [alerting](https://www.robustperception.io/alerting-on-gauges-in-prometheus-2-0/).

Please let me know if you disagree. I'm open to changing this to default to `true` prior to merging. If it causes a backwards-incompatible change for the users, we'll have to mention it explicitly in our changelog and write docs with steps on what users need to change to get the behaviour they need (e.g. how to change their alerts or dashboards). 

We usually list backwards incompatible changes here: 
https://grafana.com/docs/agent/latest/flow/release-notes/

bboreham:
> IIUC, track_timestamps_staleness = true only makes sense if out of order ingestion is allowed in the back end database.

Not as far as I know.  This is to fix the long-standing irritation where cAdvisor metrics linger on for 5 minutes after the pod has gone.

The blog you cite relates to the last change in staleness handling 5 years ago; I don't think it is relevant here.

docs/sources/flow/reference/components/prometheus.scrape.md
ptodev: > > IIUC, track_timestamps_staleness = true only makes sense if out of order ingestion is allowed in the back end database.
> 
> Not as far as I know. This is to fix the long-standing irritation where cAdvisor metrics linger on for 5 minutes after the pod has gone.
> 
> The blog you cite relates to the last change in staleness handling 5 years ago; I don't think it is relevant here.

@bboreham I think the assumption behind setting `track_timestamps_staleness = true` is that if the scraper didn't get any samples then that must be because there aren't any. But is this a good assumption to make in the general case? If a sample is exposed via explicit timestamps, and if there is no new value to report, then what sample should be exposed for its series the next time a scrape happens? Is the convention to just report the same value with a new timestamp?

I think it makes sense to default `track_timestamps_staleness` to `true` if:
* The convention is that the absence of a sample is considered enough evidence to decide that the series is stale.
* There is no need to enable out of order ingestion. This is so that we prevent a situation where a timestamp arrives late (e.g. because it takes a long time to generate) but it can't be ingested in the TSDB because there is already a staleness marker with a more recent timestamp.

This explicit timestamp feature seems like a way to "push" metrics.... so I'm not sure what assumptions are ok to make. I suspect that "push" systems like OTel don't just declare a series as stale if no samples were pushed in a certain time.

component/discovery/aws/lightsail.go
```diff
@@ -35,26 +36,30 @@ type LightsailArguments struct {
 	RoleARN         string            `river:"role_arn,attr,optional"`
 	RefreshInterval time.Duration     `river:"refresh_interval,attr,optional"`
 	Port            int               `river:"port,attr,optional"`
+
```
wildum, line 39: nit: remove newline

component/discovery/ovhcloud/ovhcloud.go
```diff
@@ -0,0 +1,94 @@
...
+		return fmt.Errorf("application_key can not be empty")
+	}
+
+	if args.ApplicationSecret == "" {
+		return fmt.Errorf("application_secret can not be empty")
+	}
+
+	if args.ConsumerKey == "" {
+		return fmt.Errorf("consumer_key can not be empty")
+	}
```
wildum, line 64: nit:  in the error messages above, it should be 'cannot' in one word

docs/sources/flow/reference/components/discovery.ovhcloud.md
```diff
@@ -0,0 +1,165 @@
...
+```
+
+Replace the following:
+  - `APPLICATION_KEY`: The OVHcloud [API][] application key.
+  - `APPLICATION_SECRET`: The OVHcloud [API][] application secret.
+  - `CONSUMER_KEY`: The OVHcloud [API][] consumer key.
+  - `SERVICE`: The OVHcloud service of the targets to retrieve.
+  - `PROMETHEUS_REMOTE_WRITE_URL`: The URL of the Prometheus remote_write-compatible server to send metrics to.
+  - `USERNAME`: The username to use for authentication to the remote_write API.
+  - `PASSWORD`: The password to use for authentication to the remote_write API.
```
wildum, line 147: nit: for the example I would suggest to also set refresh_interval and endpoint + I would directly set some real looking data instead of the placeholders (the placeholders are already used in the ##usage part)

docs/sources/flow/reference/components/prometheus.scrape.md
```diff
@@ -76,6 +77,20 @@ Name | Type | Description | Default | Required
  - [`authorization` block][authorization].
  - [`oauth2` block][oauth2].
 
+`track_timestamps_staleness` controls whether Prometheus tracks [staleness][prom-staleness] of metrics which with an explicit timestamp present in scraped data.
```
wildum, line 80:
```suggestion
`track_timestamps_staleness` controls whether Prometheus tracks [staleness][prom-staleness] of metrics which have an explicit timestamp present in scraped data.
```

wildum (commented):
looks good, just a few nits, thanks for taking care of this :)

go.mod
```diff
@@ -674,14 +677,6 @@ replace (
 	k8s.io/klog/v2 => github.com/simonpasquier/klog-gokit/v3 v3.3.0
 )
 
-// TODO(tpaschalis): remove replace directive once:
-//
-// * There is a release of Prometheus which contains
-// prometheus/prometheus#12677 and prometheus/prometheus#12729.
-// We use the last v1-related tag as the replace statement does not work for v2
-// tags without the v2 suffix to the module root.
-replace github.com/prometheus/prometheus => github.com/grafana/prometheus v1.8.2-0.20231016083943-46550094220d // grafana:prometheus:v0.47.2-retry-improvements
```
tpaschalis, line 683: Thanks! 🙏

docs/sources/flow/reference/components/discovery.ovhcloud.md
```diff
@@ -0,0 +1,165 @@
...
+}
+```
+
+## Arguments
+
+The following arguments are supported:
+
+Name               | Type           | Description                                                    | Default       | Required
+------------------ | -------------- | -------------------------------------------------------------- | ------------- | --------
+endpoint           | `string`       | [API][] endpoint.                                              | "ovh-eu"      | no
```
tpaschalis, line 42: Personal nit; feel free to disregard.

I find these easier to read if they're listed in order, first the required field, then the optional ones.

tpaschalis (approved):
LGTM, approving this to unblock you so we can move ahead once CI is green, and the discussion with Bryan has settled.

component/discovery/aws/lightsail.go
```diff
@@ -28,15 +28,14 @@ func init() {
...
-	Profile         string            `river:"profile,attr,optional"`
-	RoleARN         string            `river:"role_arn,attr,optional"`
-	RefreshInterval time.Duration     `river:"refresh_interval,attr,optional"`
-	Port            int               `river:"port,attr,optional"`
-
+	Endpoint         string                  `river:"endpoint,attr,optional"`
+	Region           string                  `river:"region,attr,optional"`
+	AccessKey        string                  `river:"access_key,attr,optional"`
+	SecretKey        rivertypes.Secret       `river:"secret_key,attr,optional"`
+	Profile          string                  `river:"profile,attr,optional"`
```
tpaschalis, line 35: Sorry to leave another stray comment. Should some of these be non-optional since they don't have a default value set? https://prometheus.io/docs/prometheus/latest/configuration/configuration/#lightsail_sd_config

docs/sources/flow/reference/components/discovery.ovhcloud.md
ptodev: Yes, I agree. I also don't like how here we repeat the definitions of the arguments. I did it this way because it's consistent with other discovery components, but I agree we should change this for all discovery components at a later point.

component/discovery/aws/lightsail.go
ptodev: I think they are truly optional. It is not possible to state what the default value will be, because the defaults come from environment variables.

docs/sources/flow/reference/components/discovery.ovhcloud.md
ptodev: The latest consensus is to sort these alphabetically: https://github.com/grafana/agent/pull/5845#discussion_r1404487210

TBH I didn't try to do it here because I wasn't sure how to do it easily. I just tried it with Copilot. It worked, and by chance it turns out the optional fields got grouped together, almost at the end of the table :)

tpaschalis: 😅 didn't know about the convention, thanks for letting me know!

wildum (approved):

docs/sources/flow/reference/components/prometheus.scrape.md
ptodev: Hi @bboreham, would you mind getting back to us on the comment above please?

mattdurham (approved):
lgtm

docs/sources/flow/reference/components/prometheus.scrape.md
bboreham: If no sample is supplied for a timestamp, PromQL (at query time) will use the preceding value up to 5 minutes old.
This creates a long-standing issue with cAdvisor (i.e. Kubelet container metrics).

It's got nothing to do with "push".

There is no expectation that explicit timestamps come out of order.  This never worked historically in Prometheus, and there is no reason to suppose people started sending them.

> But is this a good assumption to make in the general case?

Yes, it is the standard behaviour when exporters do not supply the timestamp.  Which is the vast majority.

bboreham: By the way, I don't think we should change this default in the middle of an 800-line PR doing other things.
I can make a separate PR to fix it.

ptodev: Ok, thank you, I'll take the `track_timestamps_staleness` parameter out of this PR. We can introduce it in a different PR. I don't want to add it now and change its default value later, because `prometheus.scrape` is a [stable](https://grafana.com/docs/agent/latest/stability/#stable) component and its defaults aren't meant to change often.

ptodev:
@tpaschalis @mattdurham @wildum I am re-requesting a review, because I rebased the branch and as [discussed](https://github.com/grafana/agent/pull/5972#discussion_r1441980155) I removed `track_timestamps_staleness ` from `prometheus.scrape`. The `track_timestamps_staleness` argument was added to Prometheus recently. Not having it would mean that Flow doesn't have feature parity with Static mode anymore. However, I think this is ok for two reasons:
* `track_timestamps_staleness` is a new argument which is probably not widely used yet.
* There is already a precedent for not supporting some arguments in Flow. For example, `no_proxy` in the HTTP client config.

If you agree that it's ok to not support this argument for now, please feel free to approve the PR again.

tpaschalis (approved):
LGTM! Let's follow up with discussion for the new argument in a new PR.

PR #5980, opened 2023-12-14
Clarify usage of spanmetrics connector with a prometheus remote write

ptodev:
This has been a source of confusion for several customers.

### DIFF
--- a/docs/sources/flow/reference/components/otelcol.connector.spanmetrics.md
+++ b/docs/sources/flow/reference/components/otelcol.connector.spanmetrics.md
@@ -259,42 +259,420 @@ otelcol.exporter.otlp "production" {
 
 ### Sending metrics via a Prometheus remote write
 
+The generated metrics can be sent to a Prometheus-compatible database such as Grafana Mimir.
+However, extra steps are required in order to make sure all metric samples are received.
+
+`otelcol.connector.spanmetrics` is an OTLP-native component. As such, it aims to preserve the resource attributes of spans.
+
+1. For example, let's assume that there are two incoming resources spans with the same `service.name` and `k8s.pod.name` resource attributes.
+    {{< collapse title="Example JSON of two incoming spans." >}}
+    ```json
+    {
+    	"resourceSpans": [
+    		{
+    			"resource": {
+    				"attributes": [
+    					{
+    						"key": "service.name",
+    						"value": { "stringValue": "TestSvcName" }
+    					},
+    					{
+    						"key": "k8s.pod.name",
+    						"value": { "stringValue": "first" }
+    					}
+    				]
+    			},
+    			"scopeSpans": [
+    				{
+    					"spans": [
+    						{
+    							"trace_id": "7bba9f33312b3dbb8b2c2c62bb7abe2d",
+    							"span_id": "086e83747d0e381e",
+    							"name": "TestSpan",
+    							"attributes": [
+    								{
+    									"key": "attribute1",
+    									"value": { "intValue": "78" }
+    								}
+    							]
+    						}
+    					]
+    				}
+    			]
+    		},
+    		{
+    			"resource": {
+    				"attributes": [
+    					{
+    						"key": "service.name",
+    						"value": { "stringValue": "TestSvcName" }
+    					},
+    					{
+    						"key": "k8s.pod.name",
+    						"value": { "stringValue": "first" }
+    					}
+    				]
+    			},
+    			"scopeSpans": [
+    				{
+    					"spans": [
+    						{
+    							"trace_id": "7bba9f33312b3dbb8b2c2c62bb7abe2d",
+    							"span_id": "086e83747d0e381b",
+    							"name": "TestSpan",
+    							"attributes": [
+    								{
+    									"key": "attribute1",
+    									"value": { "intValue": "78" }
+    								}
+    							]
+    						}
+    					]
+    				}
+    			]
+    		}
+    	]
+    }
+    ```
+    {{< /collapse >}}
+
+1. `otelcol.connector.spanmetrics` will preserve the incoming `service.name` and `k8s.pod.name` resource attributes by attaching them to the output metrics resource.
+    {{< collapse title="Example JSON of one outgoing metric resource." >}}
+    ```json
+    {
+        "resourceMetrics": [
+            {
+                "resource": {
+                    "attributes": [
+                        {
+                            "key": "service.name",
+                            "value": { "stringValue": "TestSvcName" }
+                        },
+                        {
+                            "key": "k8s.pod.name",
+                            "value": { "stringValue": "first" }
+                        }
+                    ]
+                },
+                "scopeMetrics": [
+                    {
+                        "scope": { "name": "spanmetricsconnector" },
+                        "metrics": [
+                            {
+                                "name": "calls",
+                                "sum": {
+                                    "dataPoints": [
+                                        {
+                                            "attributes": [
+                                                {
+                                                    "key": "service.name",
+                                                    "value": { "stringValue": "TestSvcName" }
+                                                },
+                                                {
+                                                    "key": "span.name",
+                                                    "value": { "stringValue": "TestSpan" }
+                                                },
+                                                {
+                                                    "key": "span.kind",
+                                                    "value": { "stringValue": "SPAN_KIND_UNSPECIFIED" }
+                                                },
+                                                {
+                                                    "key": "status.code",
+                                                    "value": { "stringValue": "STATUS_CODE_UNSET" }
+                                                }
+                                            ],
+                                            "startTimeUnixNano": "0",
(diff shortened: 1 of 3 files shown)


docs/sources/flow/reference/components/otelcol.connector.spanmetrics.md
```diff
@@ -259,42 +274,490 @@ otelcol.exporter.otlp "production" {
...
 
-In order for a `target_info` metric to be generated, the incoming spans resource scope
-attributes must contain `service.name` and `service.instance.id` attributes.
+The generated metrics can be sent to a Prometheus-compatible database such as Grafana Mimir.
+However, extra steps are required in order to make sure all metric samples are received.
+
+`otelcol.connector.spanmetrics` is an OTLP-native component. As such, it aims to preserve the resource attributes of spans.
+
+1. For example, let's assume that there are two incoming resources spans with the same `service.name` and `k8s.pod.name` resource attributes.
+   {{< collapse title="Example JSON of two incoming spans." >}}
```
ptodev, line 283: This PR is notable for using these `collapse` features inside the Examples section of Flow component docs. AFAIK we haven't done this before. I think it could really help illustrate this issue to people who are not familiar with OTLP.

Screenshot 2024-01-05 at 18 00 45

docs/sources/flow/reference/components/otelcol.connector.spanmetrics.md
```diff
@@ -259,42 +274,490 @@ otelcol.exporter.otlp "production" {
...
+   {{< /collapse >}}
+
+[prom-data-model]: https://prometheus.io/docs/concepts/data_model/
+
+Unfortunately, the [Prometheus data model][prom-data-model] has no notion of resource attributes.
+This means that if `otelcol.connector.spanmetrics` outputs metrics with identical metric attributes,
+but different resource attributes, `otelcol.exporter.prometheus` will convert the metrics into the same metric series.
+This problem can be solved by doing **either** of the following:
+
+- Recommended approach: Prior to `otelcol.connector.spanmetrics`, remove all resource attributes from the incoming spans which are not needed by `otelcol.connector.spanmetrics`.
```
thampiotr, line 625: Thanks for writing this up and coming up with solutions to this problem. I worry that for users this all seems complex and the Agent is doing a poor job in simplifying this for the end user. 

Ideally, I would like to have a brief description of the problem and the ability to choose between the two recommended solutions using a one-line of config. This could be a module or a component that can hide this complexity away from the user. 

Can we consider putting this problem/solutions description in a different document as a short-term way to unblock users, but open an issue to build something with a better user experience?

ptodev: If there is another module/component which can abstract away these things, then that'd be great. However, this is the reference page for `otelcol.connector.spanmetrics`, so it feels like the right place to explain why it behaves the way it does and how to use it. 

I agree that we can try to be more straight to the point in this example section. I could make a new section outside of the Examples section where I explain the tendency to preserve resource attributes? And I could link to it from the Examples section, for users who want a deeper dive into why things are the way they are?

docs/sources/flow/reference/components/otelcol.connector.spanmetrics.md
```diff
@@ -259,42 +274,490 @@ otelcol.exporter.otlp "production" {
...
 ### Sending metrics via a Prometheus remote write
 
-In order for a `target_info` metric to be generated, the incoming spans resource scope
-attributes must contain `service.name` and `service.instance.id` attributes.
+The generated metrics can be sent to a Prometheus-compatible database such as Grafana Mimir.
+However, extra steps are required in order to make sure all metric samples are received.
+
+`otelcol.connector.spanmetrics` is an OTLP-native component. As such, it aims to preserve the resource attributes of spans.
+
+1. For example, let's assume that there are two incoming resources spans with the same `service.name` and `k8s.pod.name` resource attributes.
```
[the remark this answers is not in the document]
ptodev: As mentioned in the [other comment](https://github.com/grafana/agent/pull/5980/files#r1444833913), I could move the explanation about resource attributes to another section in this doc? IMO this reference doc should the the most detailed doc about this component and should explain in detail how it works. The resource attributes explanation does describe "what the component will do". 

Btw the first time I heard about the way this components threats resource attributes, I thought it was a bug 😆 I wanted to add this explanation to illustrate why it's not.

docs/sources/flow/reference/components/otelcol.connector.spanmetrics.md
```diff
@@ -22,10 +22,25 @@ aggregates Request, Error and Duration (R.E.D) OpenTelemetry metrics from the sp
   including Errors. Multiple metrics can be aggregated if, for instance, a user wishes to
   view call counts just on `service.name` and `span.name`.
 
-- **Error** counts are computed from the Request counts which have an `Error` status code metric dimension.
+  To track requests, a `calls` metric will be created with a `status.code` metric datapoint attribute set to `Ok`:
```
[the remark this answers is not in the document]
ptodev: Yes, both requests and errors are tracked using a metric with the same name - `calls`.  I agree with you that the phrasing here could be improved! I'll change it to this:

>  Requests are tracked using a `calls` metric with a `status.code` datapoint attribute set to `Ok`:

> Errors are tracked using a `calls` metric with a `status.code` datapoint attribute set to `Error`:

Let me know if you'd like me to change it further

docs/sources/flow/reference/components/otelcol.connector.spanmetrics.md
thampiotr: > it feels like the right place to explain why it behaves the way it does

yeah, that certainly :) I just have doubts if this is the right place for what looks like a bit of a workaround... as mentioned, I'd prefer we had a way to address this with a single line of config, where we just say which of the two strategies to address the problem should be used. In practice it will be a few lines, of course, but we should perhaps consider a module or a change to the component in some way that will allow to resolve this problem in a more user-friendly way. 

Is there a way we can make this task simpler for the users?

ptodev: We could make it simpler for users if we offer a module which contains three components:
1. A transform processor which removes unnecessary resource attributes
1. A spanmetrics connector
1. An otelcol.prometheus.exporter

I could raise an issue for such a module to be created? This sort of thing is why modules are created, so I suppose we should be able to do it. But it'd be better wait and to do it with Modules v2 rather than with v1.

thampiotr: Yeah, I agree with the module approach 👍 
Migration to a new module style shouldn't be too hard, but we can prioritise this independently nevertheless.

docs/sources/flow/reference/components/otelcol.connector.spanmetrics.md
```diff
@@ -184,6 +199,348 @@ The following fields are exported and can be referenced by other components:
...
+               }
+             ]
+           }
+         ]
+       }
+     ]
+   }
+   ```
+   {{< /collapse >}}
+
```
thampiotr, line 543: Should we link at the bottom to `Sending metrics via a Prometheus remote write` saying that there's a way to ensure all samples are received?

thampiotr (approved):
Looks more clean now I believe :)

docs/sources/flow/reference/components/otelcol.connector.spanmetrics.md
ptodev: It is not necessary, because using `otelcol.connector.spanmetrics` only with other OTLP-native components is totally ok. There is only a problem when converting the OTLP metrics to Prometheus format.

This "Handling of resource attributes" section simply describes the OTLP features.

ptodev:
Hi @clayton-cornell - I intended to merge this one without your approval, while you were away :) Since your back, would you like to take a look? If you're busy I could merge it anyway. I think we will need to backport this to the 0.39 release too, once it's available.

docs/sources/flow/reference/components/otelcol.connector.spanmetrics.md
```diff
@@ -184,6 +199,348 @@ The following fields are exported and can be referenced by other components:
...
+           }
+         ]
+       }
+     ]
+   }
+   ```
+
+   {{< /collapse >}}
+
+1. In order to preserve the values of all resource attributes, `otelcol.connector.spanmetrics` will produce two resource metrics.
```
clayton-cornell, line 422:
```suggestion
1. To preserve the values of all resource attributes, `otelcol.connector.spanmetrics` will produce two resource metrics.
```

clayton-cornell (commented):
One quick tweak. Otherwise, lets get this merged and revisit once it's live (the content is ok as-is... it could use a cleanup, but that applies to the whole topic, not just this input).

clayton-cornell (approved):

PR #5984, opened 2023-12-15
helm: reload controller with configmap

charlie-haley:
#### PR Description

This PR adds the [checksum annotation](https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments) so that workload resources are rolled out when the configmap changes without needing to rely on external controllers.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/operations/helm/charts/grafana-agent/templates/controllers/_pod.yaml
+++ b/operations/helm/charts/grafana-agent/templates/controllers/_pod.yaml
@@ -3,6 +3,7 @@ metadata:
   {{- with .Values.controller.podAnnotations }}
   annotations:
     {{- toYaml . | nindent 4 }}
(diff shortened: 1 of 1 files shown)


tpaschalis (changes requested):
Hey there @charlie-haley, thanks for the PR, sorry for taking so long to get back to you.

Looking at this change, I'd like to hold off on it for now.

The Helm chart already contains the `watch-container` for reloading the configuration; we're also interested in [making the Agent itself](https://github.com/grafana/agent/issues/2691) watch and reload the configuration. In that spirit, I think that adding a third magic way of achieving the same thing might be confusing to users, especially since config changes don't mean you have to re-roll your pods.

charlie-haley:
Thanks @tpaschalis, that makes sense, having the agent reload the config would definitely be better solution, I'll close this PR!

PR #5992, opened 2023-12-17
fix(flow): ensure format consistency for all emitted logs

### LINKED ISSUE #5969: Agent not honoring  log format
### What's wrong?

Trying to logout json format using docker agent, but the first lines come up with logfmt format.

### Steps to reproduce

Every time I start (create or restart, it is the same)  the container the first lines arre coming with default logfmt format. This breaks my parsing.

### System information

Docker 24.0.6-ce

### Software version

grafana/agent:v0.38.1

### Configuration

river file:

```
logging {
  level  = "info"
  format = "json"
}
```

### Logs

```
ts=2023-12-13T14:03:37.051559331Z level=info "boringcrypto enabled"=false
ts=2023-12-13T14:03:37.052314451Z level=info msg="starting complete graph evaluation" controller_id="" trace_id=225dca1a0a36b1391f0c7bb4d252b2fa
ts=2023-12-13T14:03:37.052337881Z level=info msg="applying non-TLS config to HTTP server" service=http
ts=2023-12-13T14:03:37.052343341Z level=info msg="finished node evaluation" controller_id="" trace_id=225dca1a0a36b1391f0c7bb4d252b2fa node_id=http duration=12.88µs
ts=2023-12-13T14:03:37.052353692Z level=info msg="finished node evaluation" controller_id="" trace_id=225dca1a0a36b1391f0c7bb4d252b2fa node_id=ui duration=360ns
ts=2023-12-13T14:03:37.052359882Z level=info msg="finished node evaluation" controller_id="" trace_id=225dca1a0a36b1391f0c7bb4d252b2fa node_id=cluster duration=310ns
ts=2023-12-13T14:03:37.052794767Z level=info msg="finished node evaluation" controller_id="" trace_id=225dca1a0a36b1391f0c7bb4d252b2fa node_id=loki.write.default duration=428.985µs
ts=2023-12-13T14:03:37.05372307Z level=info msg="finished node evaluation" controller_id="" trace_id=225dca1a0a36b1391f0c7bb4d252b2fa node_id=loki.process.node duration=913.372µs
ts=2023-12-13T14:03:37.055419413Z level=info msg="finished node evaluation" controller_id="" trace_id=225dca1a0a36b1391f0c7bb4d252b2fa node_id=loki.process.docker_sd_promtail duration=1.684233ms
ts=2023-12-13T14:03:37.055466433Z level=info msg="finished node evaluation" controller_id="" trace_id=225dca1a0a36b1391f0c7bb4d252b2fa node_id=local.file_match.node duration=32.65µs
ts=2023-12-13T14:03:37.055623445Z level=info msg="finished node evaluation" controller_id="" trace_id=225dca1a0a36b1391f0c7bb4d252b2fa node_id=loki.source.file.node duration=149.491µs
ts=2023-12-13T14:03:37.055639946Z level=info msg="finished node evaluation" controller_id="" trace_id=225dca1a0a36b1391f0c7bb4d252b2fa node_id=otel duration=5.97µs
ts=2023-12-13T14:03:37.055865689Z level=info msg="finished node evaluation" controller_id="" trace_id=225dca1a0a36b1391f0c7bb4d252b2fa node_id=discovery.relabel.docker_sd_promtail duration=219.733µs
ts=2023-12-13T14:03:37.055909149Z level=info msg="finished node evaluation" controller_id="" trace_id=225dca1a0a36b1391f0c7bb4d252b2fa node_id=discovery.relabel.cadvisor duration=33.47µs
ts=2023-12-13T14:03:37.05597643Z level=info msg="finished node evaluation" controller_id="" trace_id=225dca1a0a36b1391f0c7bb4d252b2fa node_id=discovery.docker.docker_sd duration=57.19µs
ts=2023-12-13T14:03:37.056167423Z level=info msg="finished node evaluation" controller_id="" trace_id=225dca1a0a36b1391f0c7bb4d252b2fa node_id=discovery.relabel.docker_sd duration=182.223µs
{"ts":"2023-12-13T14:03:37.056196673Z","level":"info","msg":"finished node evaluation","controller_id":"","trace_id":"225dca1a0a36b1391f0c7bb4d252b2fa","node_id":"logging","duration":19210}
{"ts":"2023-12-13T14:03:37.056227434Z","level":"info","msg":"finished node evaluation","controller_id":"","trace_id":"225dca1a0a36b1391f0c7bb4d252b2fa","node_id":"labelstore","duration":13950}
{"ts":"2023-12-13T14:03:37.058708947Z","level":"info","msg":"replaying WAL, this may take a while","component":"prometheus.remote_write.default","subcomponent":"wal","dir":"/data/prometheus.remote_write.default/wal"}
{"ts":"2023-12-13T14:03:37.06043596Z","level":"info","msg":"WAL checkpoint loaded","component":"prometheus.remote_write.default","subcomponent":"wal"}
{"ts":"2023-12-13T14:03:37.063681393Z","level":"info","msg":"WAL segment loaded","component":"prometheus.remote_write.default","subcomponent":"wal","segment":3,"maxSegment":41}
{"ts":"2023-12-13T14:03:37.076268551Z","level":"info","msg":"WAL segment loaded","component":"prometheus.remote_write.default","subcomponent":"wal","segment":4,"maxSegment":41}
{"ts":"2023-12-13T14:03:37.084223967Z","level":"info","msg":"WAL segment loaded","component":"prometheus.remote_write.default","subcomponent":"wal","segment":5,"maxSegment":41}
```

The amount of the logfmt entries varies, but there are always some.

hainenber:
#### PR Description
Ensure log format consistency for all logs emitted by the agent by ensuring `logging` block got evaluated firstly and re-order possibly inconsistent log sources.

Fixes #5969 

#### Notes to the Reviewer
Is there any harm in evaluating the `logging` block twice? My naive assumption is No but would like you guys got more ctx on me on this :D

#### PR Checklist

- [x] CHANGELOG.md updated


pkg/flow/internal/controller/loader.go
```diff
@@ -154,6 +153,34 @@ func (l *Loader) Apply(args map[string]any, componentBlocks []*ast.BlockStmt, co
...
+				EndPos:   ast.EndPos(n.Block()).Position(),
+			})
+			span.SetStatus(codes.Error, err.Error())
+		}
+		if exp, ok := n.(*ExportConfigNode); ok {
+			l.cache.CacheModuleExportValue(exp.Label(), exp.Value())
+		}
+		level.Info(logger).Log("msg", "finished node evaluation", "node_id", n.NodeID(), "duration", time.Since(start))
+		span.SetStatus(codes.Ok, "")
+	}
```
wildum, line 180: Hey, thanks for contributing again :) 
The problem with this approach is that the logging node might not be a leaf. It might actually depend on another node like in this config:

```
local.file "LABEL" {
  filename = "tmp.txt"
}

logging {
  level  = "info"
  format = local.file.LABEL.content
}
```

In this case you would first need to evaluate the local.file node before being able to evaluate the logging node, which would produce some logs in the wrong format.

hainenber: Oh shoot, thats the edge case I havent thought about. Maybe we can apply the evaluation for leaf-only logging node?

I think for logging cfg, most ppl would tend to be explicit. But thats my naive assumption 😁

wildum: Maybe there is a way to buffer the logs till the logging node is evaluated so that they can be converted in the right format before being written? @rfratto what do you think?

rfratto: IMO, I think buffering logs until the first evaluation of the block will likely be easier to understand and maintain long term.

hainenber: Hi folks, when I'm implementing the buffered logger, I just discovered that there are components using modules that also emits logs and we need to buffer them into account. If we decide to tackle this, the scope will vastly increase

For example, [`prometheus.remote_write` component with its usage of `go-kit/level/log`](https://github.com/prometheus/prometheus/blob/6150e1ca0ede508e56414363cc9062ef522db518/tsdb/wlog/watcher.go#L222)

image

I think my PR and a little bit of guidance, i.e. informing folks to explicitly declare `logging` node, would be fine to resolve to this. 

Wdyt?

hainenber: Sure, it's a small one

```
logging {
  format = "json"
}

module.git "test" {
  repository = "https://github.com/ntimo/grafana-agent-git-module-test"
  path       = "prometheus.river"
}
```

wildum: Thanks, I had a quick look at it, in logger.go, all the logs are passing through `Log(kvps ...interface{})`. There it's possible to store the logs in a buffer instead of passing them to `slogadapter.GoKit(l.handler).Log(kvps...)`. The Update function is called twice, once on New and once on Evaluate. I think that we could actually not run it in the New function. That way, when the first Update comes you can empty the buffer and stop intercepting the calls in the `Log(kvps ...interface{})` function. I'm out of office now but I can help with it beginning next week if needed.

rfratto: @hainenber Can you share what you have for the buffering approach? I'd like to see what it's looking like and identify whether I can help get it finished and merged.

hainenber: Hi there, I've got a busy schedule last week. I made a draft [PR](https://github.com/hainenber/agent/pull/1) here for the buffering approach I was working on previously.

wildum: Thanks for taking a shot at it :) I don't believe that the code you shared in your PR is the right approach. I think that the problem can be solved by adding the buffering inside of the pkg/flow/logging/logger.go with the solution I described above. Only a few changes within this file should be enough to solve the issue.
Please don't feel pressured, if you have too much on your plate I'd be happy to take over this issue. But if you want to continue working on it and the approach I described is unclear don't hesitate to ask questions.

hainenber: hi there, I'll give another shot using your suggestion, thanks!

hainenber: Holy moly, your suggestion works @wildum! 

image

CHANGELOG.md
```diff
@@ -69,6 +69,9 @@ Main (unreleased)
 - Fix an issue with static integrations-next marshaling where non singletons
   would cause `/-/config` to fail to marshal. (@erikbaranowski)
 
+- Fix an issue where agent logs are emitted before correctly determine 
+  logging format. (@hainenber)
```
wildum, line 73: Fix an issue where agent logs are emitted before the logging format is correctly determined.

pkg/flow/logging/logger.go
```diff
@@ -85,11 +92,25 @@ func (l *Logger) Update(o Options) error {
 	}
 	l.writer.Set(newWriter)
 
+	// Print out the buffered logs since we determined the log format already
+	for _, bufferedLogChunk := range l.buffer {
+		if err := slogadapter.GoKit(l.handler).Log(bufferedLogChunk...); err != nil {
+			return err
+		}
+	}
```
wildum, line 100: the buffer should be emptied here, else every call to Update() will print the content of the buffer
`l.buffer = nil`

pkg/flow/logging/logger.go
```diff
@@ -85,11 +92,25 @@ func (l *Logger) Update(o Options) error {
...
 	return nil
 }
 
 // Log implements log.Logger.
 func (l *Logger) Log(kvps ...interface{}) error {
+	// Buffer logs before confirming log format is configured in `logging` block
+	l.bufferMut.Lock()
+	defer l.bufferMut.Unlock()
```
rfratto, line 108: This adds a bottleneck to logging where only one goroutine can log at a time; I'd recommend changing `l.bufferMut` to a RWMutex and grab a read lock here instead.

wildum (changes requested):

pkg/flow/logging/logger.go
```diff
@@ -23,6 +23,10 @@ type EnabledAware interface {
 type Logger struct {
 	inner io.Writer // Writer passed to New.
 
+	buffer       [][]interface{} // Store logs before correctly determine the log format
+	hasLogFormat bool            // Confirmation whether log format has been determined
+	bufferMut    sync.Mutex
```
wildum, line 28: nit: as a good practice we try to put the mutex field above the fields that it protects

pkg/flow/logging/logger.go
hainenber: I've addressed. PTAL

CHANGELOG.md
hainenber: I've addressed. PTAL

pkg/flow/logging/logger.go
hainenber: I've addressed. PTAL

pkg/flow/logging/logger.go
hainenber: Sadly there's a [data race](https://github.com/grafana/agent/actions/runs/7786596192/job/21231781964) when I switched to read lock usage since we're growing the buffer below as well. I've reverted back to vanilla mutex

pkg/flow/logging/logger_test.go
```diff
@@ -139,7 +222,14 @@ func TestLevels(t *testing.T) {
...
 			logger: func(w io.Writer) (log.Logger, error) {
-				logger, err := logging.New(w, debugLevel())
+				o := debugLevel()
+				logger, err := logging.New(w, o)
+				if err != nil {
+					return nil, err
+				}
+				if err := logger.Update(o); err != nil {
+					return nil, err
+				}
```
wildum, line 232: nit: to reduce redundancy you can add a function that takes in the options, creates the logger, updates it and returns it

hainenber: Good callout. I've amended

pkg/flow/logging/logger.go
rfratto: Yep, you're right about the datarace. I still think this is a performance bottleneck though, so we need to be a little more careful about how we write this code. 

For example, something like this could work, but other solutions might be appropriate too:

```go
l.bufferMut.RLock() 
if !l.hasLogFormat {
  l.bufferMut.RUnlock() 
  l.bufferMut.Lock() 
  // Check hasLogFormat again; could have changed since the unlock. 
  if !l.hasLogFormat {
    l.buffer = append(l.buffer, kvps) 
    return nil 
  }
  l.bufferMut.Unlock() 
} else {
  l.bufferMut.RUnlock()
}
```

It's quite ugly, but it would work. If you can find a cleaner way of implementing this it would be appreciated 🙂

pkg/flow/logging/logger.go
```diff
@@ -57,9 +64,6 @@ func New(w io.Writer, o Options) (*Logger, error) {
 		},
 	}
 
-	if err := l.Update(o); err != nil {
-		return nil, err
-	}
```
wildum, line 62: There is a problem with removing this (my bad I was the one suggesting it). It ignores the provided options and as a result it seems to create some obscure bugs in tests (the logger is used in many tests).
One way to fix it would be to keep this constructor with the update and to create a new constructor without the update call and without options. Something like "NewDeffered"? This would be used in the cmd_run.go and all the tests that depends on the loader would use the other constructor.

pkg/flow/logging/logger.go
```diff
@@ -60,6 +67,35 @@ func New(w io.Writer, o Options) (*Logger, error) {
 	if err := l.Update(o); err != nil {
 		return nil, err
 	}
+
+	return l, nil
+}
+
+// NewDeferred creates a new logger with the default log level and format.
+// The logger is not updated during initialization.
+func NewDeferred(w io.Writer, o Options) (*Logger, error) {
```
wildum, line 76: "o Options" is not used in this function and can be removed

hainenber: Removed :D

pkg/flow/logging/logger.go
wildum: there is a little bug here: there should be an `l.bufferMut.Unlock()` right before the `return nil`, else the mutex stays locked forever

hainenber: Great catch! Amended :D

wildum (approved):
LGTM, thanks for taking care of it :)

PR #5999, opened 2023-12-18
Attach Agent ID to remote-write requests

captncraig:
**Refactorings to agent seed handling**

- In flow mode, agent_seed.json should be stored in the data path now so it is more likely to persist across restarts in more environments.
- Created dedicated `internal/agentseed` package that can be used anywhere.
- There is no error case in the `Get` api. If we cannot read or write the file, worst case is we generate a new uid and use it only for the lifetime of this process.

**UID to outbound headers**

- Adds a `X-Agent-ID` header to outbound reequests from
  - `prometheus.remote_write`
  - `loki.write`
  - `pyroscope.write`
  - Static mode metrics and logs remote writes.

This header will be sent in the http headers only to remote writes you have configured.

### DIFF
--- a/component/loki/write/write.go
+++ b/component/loki/write/write.go
@@ -12,6 +12,7 @@ import (
 	"github.com/grafana/agent/component/common/loki/client"
 	"github.com/grafana/agent/component/common/loki/limit"
 	"github.com/grafana/agent/component/common/loki/wal"
+	"github.com/grafana/agent/internal/agentseed"
 )
 
 func init() {
@@ -159,6 +160,13 @@ func (c *Component) Update(args component.Arguments) error {
 	}
 
 	cfgs := newArgs.convertClientConfigs()
+	uid := agentseed.Get().UID
+	for _, cfg := range cfgs {
+		if cfg.Headers == nil {
+			cfg.Headers = map[string]string{}
+		}
+		cfg.Headers["X-Agent-UID"] = uid
+	}
 	walCfg := wal.Config{
 		Enabled:       newArgs.WAL.Enabled,
 		MaxSegmentAge: newArgs.WAL.MaxSegmentAge,
--- a/component/loki/write/write.go
+++ b/component/loki/write/write.go
@@ -165,7 +165,7 @@ func (c *Component) Update(args component.Arguments) error {
 		if cfg.Headers == nil {
 			cfg.Headers = map[string]string{}
 		}
-		cfg.Headers["X-Agent-UID"] = uid
+		cfg.Headers[agentseed.HeaderName] = uid
 	}
 	walCfg := wal.Config{
 		Enabled:       newArgs.WAL.Enabled,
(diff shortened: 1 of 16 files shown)


internal/agentseed/agentseed.go
```diff
@@ -0,0 +1,135 @@
...
+
+var dataDir = ""
+var logger log.Logger
+
+var savedSeed *AgentSeed
+
+// Init should be called by an app entrypoint as soon as it can to configure where the unique seed will be stored.
+// dir is the directory where we will read and store agent_seed.json
+// If left empty it will default to $APPDATA or /tmp
+func Init(dir string, log log.Logger) {
```
tpaschalis, line 34: nit: Is this a sync.Once type of thing? Otherwise, let's call it Set?

internal/agentseed/agentseed.go
```diff
@@ -0,0 +1,135 @@
...
+	dataDir = dir
+	logger = log
+}
+
+// Get will return a unique uuid for this agent.
+// Seed will be saved in agent_seed.json
+// If path is not empty, that will be the "preferred" place to read and save it.
+// If it is empty, we will fall back to $APPDATA on windows or /tmp on *nix systems to read the file.
+func Get() (seed *AgentSeed) {
```
tpaschalis, line 43: I think the docstring is a bit off.

internal/agentseed/agentseed.go
```diff
@@ -0,0 +1,135 @@
+package agentseed
```
tpaschalis, line 1: Is there a way to make this work for otelcol.exporter.* components?

component/loki/write/write.go
```diff
@@ -159,6 +160,13 @@ func (c *Component) Update(args component.Arguments) error {
 	}
 
 	cfgs := newArgs.convertClientConfigs()
+	uid := agentseed.Get().UID
+	for _, cfg := range cfgs {
+		if cfg.Headers == nil {
+			cfg.Headers = map[string]string{}
+		}
+		cfg.Headers["X-Agent-UID"] = uid
```
jcreixell, line 168: can we define the header name somewhere as a constant? (seems to be repeated in multiple places)

jcreixell, line 168: nit: let's call it `X-Agent-ID`? (is is a unique id, but that's not relevant to the transport layer)

internal/agentseed/agentseed.go
captncraig: Unclear at present. I'd assume there's some mechanism for including metadata, even though it may not be over http. Needs more research.

internal/agentseed/agentseed.go
jcreixell, line 43: `Get` can be understood as a side-effect operation, but in this case we initialize and store the seed. How about a more descriptive name? (`Fetch`, `FindOrCreate`, etc)

jcreixell (approved):

PR #6018, opened 2023-12-22
feat(server/http): allow Basic Auth to be configurable for web UI

### LINKED ISSUE #3601: Add Auth Basic to Grafana Agent API

### LINKED ISSUE #6008: Basic Authentication and TLS options for Grafana Flow UI
### Request

The Grafana Flow UI is very useful when it comes to troubleshooting and debugging components and also verify if flow is as per our expectation.
We would like to have a basic authentication protecting the UI similar to the webtoolkit options available in Prometheus or Node exporter.

### Use case

Even though there is no sensitive information available in the UI it also serves the agent meta monitoring metrics which contains the IP address, version of the Grafana agent, components used, remote write system info etc. 

Since the basic auth is not available we are not able to use the UI and restricting the server listening address only to localhost.

We hope to see the support for basic auth along with other TLS parameters  such as TLS version, cipher suites etc provided by webtoolkit used by  softwares such as prometheus and node exporter to be available for the Grafana Flow UI as well.

https://github.com/prometheus/exporter-toolkit/blob/master/docs/web-configuration.md

hainenber:

Fixes #3601 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


docs/sources/flow/reference/cli/run.md
```diff
@@ -50,6 +50,7 @@ The following flags are supported:
   (default `agent.internal:12345`).
 * `--server.http.listen-addr`: Address to listen for HTTP traffic on (default `127.0.0.1:12345`).
 * `--server.http.ui-path-prefix`: Base path where the UI is exposed (default `/`).
+* `server.http.basic-auth-filepath`: Absolute path to file containing user and hashed password for Basic Auth (default `""`, signifies disabled).
```
tunguyen9889, line 53:
```suggestion
* `--server.http.basic-auth-filepath`: Absolute path to file containing user and hashed password for Basic Auth (default `""`, signifies disabled).
```

hainenber:
Basic Auth creds are prompted when trying to access agent's `web` UI, including component routes

image
image

Operational routes, i.e. `/debug/pprof`, `/metrics`, `/-/ready` and alikes are left alone
image
image
image

clayton-cornell (commented):
Doc changes are really small in this PR. OK as-is

clayton-cornell:
@grafana/grafana-agent-maintainers This probably needs a once-over by the dev team

tpaschalis (changes requested):
Hey there 👋  While this looks promising, we'd like to hold off of it for now after some discussion.

The rationale is that for common deployment methods, it's easy enough to achieve the same functionality with a separate layer (eg. using some authentication proxy or mLTS). 

I like how we've contained everything to the UI service in this PR, but maintaining this Agent-side would eventually mean adding more authentication methods and/or extending it to the HTTP service which requires some more care.

If we decide to revisit, we could certainly revive this PR and use it as the base!

PR #6021, opened 2023-12-26
Adding more option to enable required mongo stats to monito

Nitesh-vaidyanath:
#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Documentation added
- [X] Tests updated
- [] Config converters updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2317,7 +2317,7 @@ v0.24.0 (2022-04-07)
 
 ### Features
 
-- Adding more option to enable required mongo stats to monitor. (@Nitesh0-vaidyanath)
+- Adding more option to enable required mongo stats to monitor. (@Nitesh-vaidyanath)
 
 - Added config read API support to GrafanaAgent Custom Resource Definition.
   (@shamsalmon)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2317,6 +2317,8 @@ v0.24.0 (2022-04-07)
 
 ### Features
 
+- Adding more option to enable required mongo stats to monitor. (@Nitesh0-vaidyanath)
+
 - Added config read API support to GrafanaAgent Custom Resource Definition.
   (@shamsalmon)
 
--- a/docs/sources/flow/reference/components/prometheus.exporter.mongodb.md
+++ b/docs/sources/flow/reference/components/prometheus.exporter.mongodb.md
@@ -39,6 +39,12 @@ Omitted fields take their default values.
 | `direct_connect`             | `boolean` | Whether or not a direct connect should be made. Direct connections are not valid if multiple hosts are specified or an SRV URI is used. | false   | no       |
 | `discovering_mode`           | `boolean` | Wheter or not to enable autodiscover collections.                                                                                       | false   | no       |
 | `tls_basic_auth_config_path` | `string`  | Path to the file having Prometheus TLS config for basic auth. Only enable if you want to use TLS based authentication.                  |         | no       |
+| `enable_db_stats`            | `boolean` | Enable mongo database stats | false   | no       |
+| `enable_diagnostic_data`     | `boolean` | Enable mongo diagonstic data | false   | no       |
+| `enable_replicaset_status`   | `boolean` | Enable mongo replica set status stats | false   | no       |
+| `enable_top_metrics`         | `boolean` | Enable mongo top metrics | false   | no       |
+| `enable_index_stats`         | `boolean` | Enable mongo index stats | false   | no       |
+| `enable_coll_stats`          | `boolean` | Enable mongo collection stats | false   | no       |
 
 MongoDB node connection URI must be in the [`Standard Connection String Format`](https://docs.mongodb.com/manual/reference/connection-string/#std-label-connections-standard-connection-string-format)
 
--- a/docs/sources/static/configuration/integrations/mongodb_exporter-config.md
+++ b/docs/sources/static/configuration/integrations/mongodb_exporter-config.md
@@ -79,6 +79,34 @@ Besides that, there's not much to configure. Please refer to the full reference 
   # MongoDB node connection URL, which must be in the [`Standard Connection String Format`](https://docs.mongodb.com/manual/reference/connection-string/#std-label-connections-standard-connection-string-format)
   [mongodb_uri: <string>]
 
+  # By default all the below stats will be
+  # collected by mongo exporter.
+  # EnableDBStats=true
+  # EnableCollStats=true
+  # EnableTopMetrics=true
+  # EnableReplicasetStatus=true
+  # EnableIndexStats=true
+  # EnableCurrentopMetrics=true
+  # EnableProfile=true
+  # EnableDiagnosticData=true
+
+  # To disable unwanted stats and to 
+  # collect only required stats set parameter
+  # with prefix enable to true.  
+
+  # Mongo exporter variable EnableDBStats 
+  [enable_db_stats: <boolean> | default = false]
+  # Mongo exporter variable  EnableDiagnosticData
+  [enable_diagnostic_data: <boolean> | default = false]
+  # Mongo exporter variable EnableReplicasetStatus
+  [enable_replicaset_status: <boolean> | default = false]
+  # Mongo exporter variable EnableTopMetrics
+  [enable_top_metrics: <boolean> | default = false]
+  # Mongo exporter variable EnableIndexStats
+  [enable_index_stats: <boolean> | default = false]
+  # Mongo exporter variable EnableCollStats
+  [enable_coll_stats: <boolean> | default = false]
+
   # Whether or not a direct connect should be made. Direct connections are not valid if multiple hosts are specified or an SRV URI is used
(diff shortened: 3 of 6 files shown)


pkg/integrations/mongodb_exporter/mongodb_exporter.go
```diff
@@ -17,15 +17,29 @@ import (
...
 	DirectConnect: true,
+	EnableDBStats: false,
+	EnableDiagnosticData: false,
+	EnableReplicasetStatus: false,
+	EnableTopMetrics: false,
+	EnableIndexStats: false,
+	EnableCollStats: false,
 }
 
+var collectAll = true
```
marctc, line 28: Is there any reason to make this variable global?

pkg/integrations/mongodb_exporter/mongodb_exporter.go
```diff
@@ -17,15 +17,29 @@ import (
 
 var DefaultConfig = Config{
 	DirectConnect: true,
+	EnableDBStats: false,
```
marctc, line 20: wouldn't be better to have all of this enabled by default instead of setting them to true later (from line 92).

docs/sources/static/configuration/integrations/mongodb_exporter-config.md
```diff
@@ -79,6 +79,34 @@ Besides that, there's not much to configure. Please refer to the full reference
   # MongoDB node connection URL, which must be in the [`Standard Connection String Format`](https://docs.mongodb.com/manual/reference/connection-string/#std-label-connections-standard-connection-string-format)
   [mongodb_uri: <string>]
 
+  # By default all the below stats will be
```
marctc, line 82: This explanation should be also added to flow component doc. But if we change the default behavior, we can remove them altogether.

Nitesh-vaidyanath: Hi @marctc, there is no change in default behavior. I have just added more configurable parameter.

Nitesh-vaidyanath: Yes i have already added this to [docs/sources/flow/reference/components/prometheus.exporter.mongodb.md](https://github.com/grafana/agent/pull/6021/files#diff-57174b3cfd9003a73fbd9354096a52c6890fa311a439e0082f50cba8129e281c)

pkg/integrations/mongodb_exporter/mongodb_exporter.go
marctc: Maybe I'm missing something, but I still don't understand why these flags are not enabled by default and we need a `collectAll` variable (which is the default behavior).

docs/sources/static/configuration/integrations/mongodb_exporter-config.md
marctc: By default behavior I meant my comment in https://github.com/grafana/agent/pull/6021/files#r1437078040. The component docs specifies that those flags are disabled by default but then the code does the opposite (as it reflected here).

pkg/integrations/mongodb_exporter/mongodb_exporter.go
Nitesh-vaidyanath: Yeah it is bit confussing but setting collectAll to true and enabling all other other flags doesn't make any differnce. In the comment i added for the sake of understanding, collectAll=ture is equaliant to enabling all the other flags  however, if we enable any one or more flags it will diasable collectAll and other flags too.

CHANGELOG.md
```diff
@@ -2317,6 +2317,8 @@ v0.24.0 (2022-04-07)
 
 ### Features
 
+- Adding more option to enable required mongo stats to monitor. (@Nitesh-vaidyanath)
```
marctc, line 2320: This changelog entry is placed in a old release section. Please, move it to unreleased section.

marctc (changes requested):
Please see comments. Once comments are addressed the PR would be ready to be merged.

pkg/integrations/mongodb_exporter/mongodb_exporter.go
marctc: If it's confusing for you, it will be confusing for the rest of users. Would you mind refactor the configuration to be easier to understand?

jcreixell: I think I understand the point that, because `CollectAll` was hardcoded to `true`, the new flags won't have any effect unless we expose the flag and users manually disable it.
 
In order to mirror the actual [collector's logic](https://github.com/percona/mongodb_exporter/blob/cb3db8922fb068bba6ecd4c12c231cb63d970515/exporter/exporter.go#L153-L165) as closely as possible and avoid misunderstanding from users, I would suggest fully exposing `CollectAll` as a new flag, and setting it to `true` in `DefaultConfig` (as @marctc suggested).

Rationale: most users won't need that level of granularity, but if they do, they can disable `CollectAll` and set each desired flag individually.

We should also make sure that we are exposing all the flags that would be enabled by `CollectAll` (afaik we are missing [some](https://github.com/percona/mongodb_exporter/blob/cb3db8922fb068bba6ecd4c12c231cb63d970515/exporter/exporter.go#L157-L165)).

This is also a potential argument for not exposing any flags at all: if the collector dependency is updated and some new flags are added and missed as part of the update, users who manually set the list of collectors to enable may miss the new ones, so there is some maintenance cost involved in providing such level of granularity.

tpaschalis:
@Nitesh-vaidyanath I think we're on a good path, do you have the time to address @marctc's feedback here?

docs/sources/flow/reference/components/prometheus.exporter.mongodb.md
```diff
@@ -39,6 +39,12 @@ Omitted fields take their default values.
...
 | `discovering_mode`           | `boolean` | Wheter or not to enable autodiscover collections.                                                                                       | false   | no       |
 | `tls_basic_auth_config_path` | `string`  | Path to the file having Prometheus TLS config for basic auth. Only enable if you want to use TLS based authentication.                  |         | no       |
+| `enable_db_stats`            | `boolean` | Enable mongo database stats | false   | no       |
+| `enable_diagnostic_data`     | `boolean` | Enable mongo diagonstic data | false   | no       |
+| `enable_replicaset_status`   | `boolean` | Enable mongo replica set status stats | false   | no       |
+| `enable_top_metrics`         | `boolean` | Enable mongo top metrics | false   | no       |
+| `enable_index_stats`         | `boolean` | Enable mongo index stats | false   | no       |
+| `enable_coll_stats`          | `boolean` | Enable mongo collection stats | false   | no       |
```
clayton-cornell, line 47:
```suggestion
| `enable_db_stats`            | `boolean` | Enable MongoDB database statistics. | false   | no       |
| `enable_diagnostic_data`     | `boolean` | Enable MongoDB diagnostic data. | false   | no       |
| `enable_replicaset_status`   | `boolean` | Enable MongoDB replica set status statistics. | false   | no       |
| `enable_top_metrics`         | `boolean` | Enable MongoDB top metrics. | false   | no       |
| `enable_index_stats`         | `boolean` | Enable MongoDB index statistics. | false   | no       |
| `enable_coll_stats`          | `boolean` | Enable MongoDB collection statistics. | false   | no       |
```

docs/sources/static/configuration/integrations/mongodb_exporter-config.md
```diff
@@ -79,6 +79,34 @@ Besides that, there's not much to configure. Please refer to the full reference
...
+  [enable_db_stats: <boolean> | default = false]
+  # Mongo exporter variable  EnableDiagnosticData
+  [enable_diagnostic_data: <boolean> | default = false]
+  # Mongo exporter variable EnableReplicasetStatus
+  [enable_replicaset_status: <boolean> | default = false]
+  # Mongo exporter variable EnableTopMetrics
+  [enable_top_metrics: <boolean> | default = false]
+  # Mongo exporter variable EnableIndexStats
+  [enable_index_stats: <boolean> | default = false]
+  # Mongo exporter variable EnableCollStats
```
clayton-cornell, line 107:
```suggestion
  # MongoDB exporter variable EnableDBStats 
  [enable_db_stats: <boolean> | default = false]
  # MongoDB exporter variable  EnableDiagnosticData
  [enable_diagnostic_data: <boolean> | default = false]
  # MongoDB exporter variable EnableReplicasetStatus
  [enable_replicaset_status: <boolean> | default = false]
  # MongoDB exporter variable EnableTopMetrics
  [enable_top_metrics: <boolean> | default = false]
  # MongoDB exporter variable EnableIndexStats
  [enable_index_stats: <boolean> | default = false]
  # MongoDB exporter variable EnableCollStats
```

CHANGELOG.md
clayton-cornell, line 2320:
```suggestion
- Adding more option to enable required MongoDB stats to monitor. (@Nitesh-vaidyanath)
```

We also need to correct the MongoDB name to match their product name/capitalization

clayton-cornell (commented):
Some doc input

PR #6023, opened 2023-12-27
Add additional service labels

m4r1u2:
#### PR Description

Add helm option to set extra service lables.

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Documentation added
- [ ] Tests updated


tpaschalis:
Hey there, thanks for the contribution. Could you explain the rationale behind service layers? Is there a difference than using the normal `grafana-agent.labels` field here?

PR #6026, opened 2023-12-27
add max_cache_size

mattdurham:
#### PR Description

Allow configuring the max_relabel_cache, there are many cases where 100k is far to low.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added

### DIFF
--- a/component/prometheus/relabel/relabel.go
+++ b/component/prometheus/relabel/relabel.go
@@ -46,15 +46,15 @@ type Arguments struct {
 	MetricRelabelConfigs []*flow_relabel.Config `river:"rule,block,optional"`
 
 	// Cache size to use for LRU cache.
-	//CacheSize int `river:"cache_size,attr,optional"`
+	CacheSize int `river:"max_cache_size,attr,optional"`
 }
 
 // SetToDefault implements river.Defaulter.
-/*func (arg *Arguments) SetToDefault() {
+func (arg *Arguments) SetToDefault() {
 	*arg = Arguments{
-		CacheSize: 500_000,
+		CacheSize: 100_000,
 	}
-}*/
+}
 
 // Exports holds values which are exported by the prometheus.relabel component.
 type Exports struct {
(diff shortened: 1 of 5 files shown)


component/prometheus/relabel/relabel.go
```diff
@@ -46,15 +46,15 @@ type Arguments struct {
...
-	//CacheSize int `river:"cache_size,attr,optional"`
+	CacheSize int `river:"max_cache_size,attr,optional"`
 }
 
 // SetToDefault implements river.Defaulter.
-/*func (arg *Arguments) SetToDefault() {
+func (arg *Arguments) SetToDefault() {
 	*arg = Arguments{
-		CacheSize: 500_000,
+		CacheSize: 100_000,
```
thampiotr, line 55: Just checking - the 500k had no effect previously?

component/prometheus/relabel/relabel.go
```diff
@@ -46,15 +46,15 @@ type Arguments struct {
 	MetricRelabelConfigs []*flow_relabel.Config `river:"rule,block,optional"`
 
 	// Cache size to use for LRU cache.
-	//CacheSize int `river:"cache_size,attr,optional"`
+	CacheSize int `river:"max_cache_size,attr,optional"`
```
thampiotr, line 49: We probably want to validate this is not negative. I wonder if 0 has a meaning of unlimited and if we should allow it at all? Seems like better not to allow it, but I could be convinced otherwise.

thampiotr (approved):
LGTM % comments and converter test fixes :)

component/prometheus/relabel/relabel.go
mattdurham: It was commented out, the real default was 100k.

component/prometheus/relabel/relabel.go
mattdurham: 0 and below basically means the lru cache always evicts after adding.

mattdurham: will add a check for 0

PR #6028, opened 2023-12-28
Add resource usage guidelines docs

### LINKED ISSUE #5950: Provide better guidance on agent resource requirements

thampiotr:
#### PR Description

Add resource usage guidelines based on our experiences.

Fixes #5950

#### Notes to the Reviewer

Note: the traces usage is too low to provide any reasonable guidelines. We will need to come back to this.

#### PR Checklist

- [x] Documentation added

### DIFF
--- a/docs/sources/flow/monitoring/agent-resource-usage.md
+++ b/docs/sources/flow/monitoring/agent-resource-usage.md
@@ -1,0 +1,82 @@
+---
+aliases:
+- /docs/agent/flow/monitoring/resource-usage/
+- /docs/grafana-cloud/agent/flow/monitoring/resource-usage/
+- /docs/grafana-cloud/monitor-infrastructure/agent/flow/monitoring/resource-usage/
+- /docs/grafana-cloud/monitor-infrastructure/integrations/agent/flow/monitoring/resource-usage/
+- /docs/grafana-cloud/send-data/agent/flow/monitoring/resource-usage/
+canonical: https://grafana.com/docs/agent/latest/flow/monitoring/resource-usage/
+description: Guidance for expected Agent resource usage
+headless: true
+title: Resource usage
+---
+
+# {{% param "PRODUCT_NAME" %}} resource usage
+This page provides guidance for expected resource usage of {{% param "PRODUCT_NAME" %}}
+for each telemetry type. The information on this page is based on operational
+experience of some of the {{% param "PRODUCT_NAME" %}} maintainers.
+
+{{% admonition type="note" %}}
+
+The resource usage depends on the workload, hardware and the configuration used.
+The information provided on this page is intended to be a starting point for 
+most users and your actual resource usage may be different.
+
+{{% /admonition %}}
+
+## Prometheus metrics
+
+The Prometheus metrics resource usage depends mainly on the number of active series
+that need to be scraped.
+
+As a rule of thumb, **per each 1 million active series** with 15s scrape interval,
+you can expect to use approximately:
+* 1.1 CPU cores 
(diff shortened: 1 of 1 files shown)


docs/sources/flow/monitoring/agent-resource-usage.md
```diff
@@ -0,0 +1,72 @@
...
+most users, but your actual usage may be different.
+
+{{% /admonition %}}
+
+## Prometheus metrics
+
+The Prometheus metrics resource usage depends mainly on the number of active series
+that need to be scraped.
+
+As a rule of thumb, **per each 1 million active series** with 15s scrape interval,
```
mattdurham, line 32: We should use 60s here since that is what we recommend everywhere.

thampiotr: Hmm, in that case we don't have a data from production... unless we change our prod to match what we recommend. Any idea why do we not use the recommended interval?

jcreixell: AFAIK 60s is optimized for Grafana Cloud and is documented here https://grafana.com/docs/grafana-cloud/cost-management-and-billing/understand-your-invoice/metrics-invoice/#data-points-per-minute-dpm. It is possible to tweak it via scrape config but it has a direct impact on cost

thampiotr: I guess the best way forward is for me to set up an experiment where we do a 60s scrape interval for some time to collect data?

jcreixell: We can try and see what we get, I assume most OSS users will use the 15s interval and Cloud users would prefer the 60s interval. My guess is that memory usage should be similar, but CPU and memory will change (more data points to filter and send)

docs/sources/flow/monitoring/agent-resource-usage.md
```diff
@@ -0,0 +1,78 @@
...
+
+The Prometheus metrics resource usage depends mainly on the number of active
+series that need to be scraped.
+
+As a rule of thumb, **per each 1 million active series** 
+you can expect to use approximately:
+
+* 1.5 CPU cores
+* 11 GiB of memory
+* 1.5 MiB/s of total network bandwidth, send and receive
```
thampiotr, line 38: @mattdurham @jcreixell - this is based on experimental data from setting scrape interval to 1m. I think it's the best data we have right now until we start doing benchmarks more systematically and I'd recommend we go with this for now. LMK if you have any objections :)

docs/sources/flow/monitoring/agent-resource-usage.md
```diff
@@ -0,0 +1,78 @@
...
+The resource usage depends on the workload, hardware and the configuration used.
+The information on this page is a good starting point for most users, but your
+actual usage may be different.
+
+{{% /admonition %}}
+
+## Prometheus metrics
+
+The Prometheus metrics resource usage depends mainly on the number of active
+series that need to be scraped.
```
jcreixell, line 31: nit: The Prometheus metrics resource usage depends mainly on the number of active
series that need to be scraped **, and the scrape interval**.

docs/sources/flow/monitoring/agent-resource-usage.md
```diff
@@ -0,0 +1,78 @@
...
+
+As a rule of thumb, **per each 1 million active series**
```
jcreixell, line 33: nit: mention here "with default settings"

jcreixell (approved):
added some nits, LGTM overall

Issue #5949, opened 2023-12-11
Create a clear space in docs hierarchy for the task-based docs


thampiotr:
Currently, we have the following top-level hierarchy:

<img width="336" alt="image" src="https://github.com/grafana/agent/assets/17101802/05d8593a-6077-4a24-afd7-be5cd70538c6">

We don't have a dedicated place to put the task-based documentation. Existing top-level sections seem to overlap.

This issue proposes to have the following top-level sections instead:
- Get started - should include how to quickly install the agent and get hands-on experience. There may be some overlap or links to other pages, but the difference is that "get started" here is dedicated for the new users who are onboarding. We may go less in-depth in these pages and showcase the features and advantages of the Agent.
- Concepts - learn the concepts of the Agent used throughout the documentation. 
- Tasks - pages dedicated to performing specific tasks and achieving a desired outcome. There can be many of them, these should be production-ready and contain best practices and recommendation. They are low level of detail, with Reference pages being the only ones more detailed.
- Tutorials - pages dedicated to learning - usually covering more ground than Tasks, may not need to be production-ready. It's a good place to go to after you've done Get Started section and want to prototype something. After completing a tutorial, the user should be able to dive deeper using Tasks section or Reference. 
- Reference - remains unchanged. The most detailed docs we offer.
- Release notes - unchanged.

Once we have a broad consensus on what sections we want to have, we can assign the existing pages to correct sections.

thampiotr:
@clayton-cornell and @ptodev, I would very much like to hear your opinion on this proposal and what alternative do you think we should consider.

mattdurham:
Would this create a more nested hierarchy?

clayton-cornell:
Another thought... split off or at least separate the River-specific concepts from how to use Rive in Agent.  Iv'e been tinkering with this already (but no concrete results). There's lot of overlap and "mushing together" of River and Agent in the current Concepts and Configuration language sections.

thampiotr:
> Would this create a more nested hierarchy? 
@mattdurham not really, because we mostly move things around on the same level where they are. I think maybe the "Monitoring" section would become Task-based topics instead.

> split off or at least separate the River-specific concepts from how to use Rive in Agent
@clayton-cornell that would make sense, I agree - we could start by just splitting it into separate pages and making sure we talk about River more generally and then have a page on how Agent uses River. I think that maybe that change can be done separately from this issue.

ptodev:
[...]

PR #6031, opened 2024-01-02
Docs reorg: part 1: add section definitions

thampiotr:
#### PR Description

Adds a description of the new docs hierarchy that matches the input from discussion on https://github.com/grafana/agent/issues/5949

After merging, the next step will be to actually do the reorg using the existing documentation pages. We'll put existing pages in their best fitting new categories, even if they're not 100% compliant to start with - we'll be cleaning them up afterwards. And we'll be able to add new content, e.g. in the Tasks section.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,14 +1,15 @@
 # Grafana Agent Documentation
 
-This directory contains documentation for Grafana Agent. It is split into two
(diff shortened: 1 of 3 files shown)


rfratto (approved):
LGTM :+1:

PR #6042, opened 2024-01-03
helm: clean up logic for autodelete pvcs

captncraig:
#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/operations/helm/charts/grafana-agent/CHANGELOG.md
+++ b/operations/helm/charts/grafana-agent/CHANGELOG.md
@@ -17,7 +17,6 @@ Unreleased
 ### Enhancements
 
 - Update `rbac` to include necessary rules for the `otelcol.processor.k8sattributes` component. (@rlankfo)
-- Add value `.controller.minK8sVersionForAutoDelete` to support earlier k8s versions with beta feature gates enabled. (@captncraig)
 
 0.29.0 (2023-11-30)
 -------------------
--- a/operations/helm/charts/grafana-agent/CHANGELOG.md
+++ b/operations/helm/charts/grafana-agent/CHANGELOG.md
@@ -10,9 +10,14 @@ internal API changes are not present.
 Unreleased
 ----------
 
+### Bugfixes
+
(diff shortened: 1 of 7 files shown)


operations/helm/charts/grafana-agent/CHANGELOG.md
```diff
@@ -10,9 +10,14 @@ internal API changes are not present.
 Unreleased
 ----------
 
+### Bugfixes
```
rfratto, line 13: Bugfixes goes below enhancements, doesn't it? https://github.com/grafana/agent/blob/main/docs/developer/contributing.md#updating-the-changelog

operations/helm/charts/grafana-agent/values.yaml
```diff
@@ -189,6 +189,9 @@ controller:
   # -- Whether to enable automatic deletion of stale PVCs due to a scale down operation, when controller.type is 'statefulset'.
   enableStatefulSetAutoDeletePVC: false
 
+  # -- The StatefulSetAutoDeletePVC feature gate went into alpha status with k8s 1.27. If you have it enabled on an earlier version, you can lower this value as far as 1.23.
+  minK8sVersionForAutoDelete: ">= 1.27"
```
rfratto, line 193: I'm looking at what other Helm charts do here to remove the need for another option. It looks like the Loki chart [only checks for 1.23](https://github.com/grafana/loki/blob/5c502e497fcf1e53030a885046098ee9981bf07b/production/helm/loki/templates/read/statefulset-read.yaml#L27). I see a similar check in the [Consul Helm chert](https://github.com/hashicorp/consul-k8s/blob/5ed8514254e026f562d703e2acc7ae7595ad6610/charts/consul/templates/server-statefulset.yaml#L47).

Ideally we'd avoid a new config option here, and rely on documentation on `enableStatefulSetAutoDeletePVC`, where we mention it's a feature gate from 1.23 to 1.26, and generally available afterwards.

rfratto (approved):
Overall LGTM. I'm not sure about adding a Kube version setting since other charts don't seem to do the same thing (see comment) and it seems like we could get away with a more lax check with documentation warning users about when it's safe to enable.

PR #6043, opened 2024-01-04
Move 'Getting started' to 'Tasks'

thampiotr:
#### PR Description
This PR moves the pages in current "getting started" to a new "Tasks" section and maintains the page aliases to keep the links working. 

The current pages inside the "getting started" appear to best match the guidelines for Tasks docs. Even though they are not 100% compliant, this is a first step towards implementing a new docs structure as [described here](https://github.com/grafana/agent/blob/main/docs/developer/writing-docs.md) and [discussed here](https://github.com/grafana/agent/issues/5949). 

Next steps to implement the new structure:
- Move the "Monitoring" section into "Tasks" - this may require some small tweaks to content to make it more task-oriented
- Create a new "Getting started" section and move to it the current installation pages.

PR #6051, opened 2024-01-05
Create migrate tasks subdirectory

thampiotr:
#### PR Description

This just creates a Migrate tasks subdirectory to better organise them.

image

* Checked for warnings and there are no broken links.
* Not creating new backwards-compatible aliases for previous location, because the pages I'm moving here were at their new location only for about 1h as of this writing: https://github.com/grafana/agent/pull/6043.


thampiotr:
Discussed offline with Paulin. Given that these links were short lived and recent changes, we prefer to merge this rather sooner than later. I'll merge now. If any issues come up, we'll clean up when Clayton is back. Thanks for review!

PR #6052, opened 2024-01-05
Add `sample_age_limit` to remote_write config to drop old samples

### LINKED ISSUE #4651: Improve remote write outage recovery by discarding too-old samples
During prolonged outages of a remote write endpoint, Grafana Agents will continue accumulating data and can potentially fall behind by a significant period of time.

Once an outage ends, catching up (if possible at all) could take hours, since the WAL contents are read and pushed to the remote_write endpoint in sequence.

Moreover, in cases where the remote_write endpoint has a configured timeframe as to how old data it can accept, most of this data will be discarded at the database layer, putting unnecessary load  on both the agent and the remote write backend.

An interesting idea that can generally help with recovering from outages would be to discard datapoints that we know cannot be accepted by the remote earlier on. A way to achieve this is the following Prometheus proposal, where some of the legwork has already been done for us.

https://github.com/prometheus/prometheus/issues/7912

marctc:
#### PR Description

This PR adds the feature of adding a limit of the age of samples to be sent to remote write.
It includes the Prometheus 2.48.1 release with the changes from this [PR](https://github.com/prometheus/prometheus/pull/13002).

Fixes #4651

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [ ] Tests updated


docs/sources/flow/reference/components/prometheus.remote_write.md
```diff
@@ -191,6 +192,10 @@ responses should be treated as recoverable errors; other `HTTP 4xx` status code
 responses are never considered recoverable errors. When `retry_on_http_429` is
 enabled, `Retry-After` response headers from the servers are honored.
 
+The `sample_age_limit` argument specifies the maximum age of samples to send. Any
+samples older than the limit are dropped and won't be sent to the remote storage.
```
tpaschalis, line 196: Can we say that by default, or when it is set to zero the feature is disabled?

tpaschalis (approved):

docs/sources/flow/reference/components/prometheus.remote_write.md
tpaschalis: Also, is there anything to do for the config converters here?

marctc: I think so, some tests are failing.

ptodev (commented):
Are we confident that this change will be merged in Prometheus? Could we be stuck on a fork? Or can we remove this from the Agent, even though prometheus.remote_write is a stable component?

marctc:
> Are we confident that this change will be merged in Prometheus? 

Yes, the PR got enough feedback and it will be merged anytime soon

PR #6055, opened 2024-01-05
loki.source.docker: deduplicate matching container IDs

### LINKED ISSUE #4403: loki.source.docker error: could not set up a wait request to the Docker client
### What's wrong?

I have a very basic Docker log scraping config with the agent running in flow mode. The logs seem to be transmitted normally, however I get constant error/warning messages from the `grafana-agent` complaining about the inability to read container logs. The error messages are produced and repeated for all targeted containers on a regular basis.

### Steps to reproduce

Simply run `AGENT_MODE=flow grafana-agent --server.http.listen-addr=0.0.0.0:12345 run /etc/grafana-agent.river` with the provided config on the affected host.

Agent is running as user `grafana-agent` who is a member of the following groups: `grafana-agent adm systemd-journal docker`

### System information

Linux 4.4.0-92-generic x86_64

### Software version

Grafana Agent v0.34.3

### Configuration

```text
logging {
  level  = "warn"
  format = "logfmt"
}

prometheus.scrape "default" {
	targets    = [{"__address__" = "localhost:12345"}]
	forward_to = [prometheus.remote_write.prom.receiver]
}

prometheus.exporter.unix {
	include_exporter_metrics = true
	disable_collectors       = ["mdadm"]
}

prometheus.remote_write "prom" { }

discovery.docker "linux" {
	host = "unix:///var/run/docker.sock"
	filter {
		name = "status"
		values = ["running"]
	}
}

loki.source.docker "default" {
	host    = "unix:///var/run/docker.sock"
	targets = discovery.relabel.docker.output
	forward_to = [loki.process.final.receiver]
}

discovery.relabel "docker" {
	targets = discovery.docker.linux.targets

	rule {
		source_labels = ["__meta_docker_container_name"]

		target_label = "container_name"
	}

	rule {
		source_labels = ["__meta_docker_container_id"]
		target_label  = "container_id"
	}
}

loki.relabel "journal" {
	forward_to = []

	rule {
		source_labels = ["__journal__systemd_unit"]
		target_label  = "unit"
	}

	rule {
		source_labels = ["__journal_priority_keyword"]
		target_label  = "level"
	}

	rule {
		source_labels = ["__journal__transport"]
		target_label  = "transport"
	}

	rule {
		source_labels = ["__journal__boot_id"]
		target_label  = "boot_id"
	}
}

loki.source.journal "read" {
	forward_to    = [loki.process.final.receiver]
	relabel_rules = loki.relabel.journal.rules
}

loki.process "final" {
	forward_to = [loki.write.grafana_cloud.receiver]

	stage.drop {
		older_than = "1h"
	}
}

loki.write "grafana_cloud" {
	external_labels = {instance = constants.hostname}

	endpoint {
		url = "https://USER:SECRET@logs-prod3.grafana.net/loki/api/v1/push"
	}
}
```

### Logs

```text
# the following logs repeat many times, once for each container; then repeat every few minutes

2023-07-11 19:00:04	ts=2023-07-11T23:00:04.735108377Z component=loki.source.docker.default level=warn target=docker/4f8edb61c3d75c606ddf0159ba4892aee758a53a0a58e9b4faec13fa63457538 msg="could not transfer logs" written=0 container=4f8edb61c3d75c606ddf0159ba4892aee758a53a0a58e9b4faec13fa63457538 err="context canceled"

2023-07-11 19:00:04	ts=2023-07-11T23:00:04.735116325Z component=loki.source.docker.default level=error target=e2a20d798dbb394b802179ed42fd1001ab4a5d0ed66c69649e65df3a65be0221 msg="could not set up a wait request to the Docker client" error=
```

tpaschalis:
#### PR Description
The `discovery.docker` component was initially built with Prometheus service discovery in mind; this means that it will report a different target for each port, network and network interface that a container exposes/belongs to, so that it can be scraped accordingly.

This means, that when `discovery.docker` is used as an input to loki.source.docker, it will set up unnecessary load on the Docker API by trying to fetch logs for the same container ID multiple times. This PR makes it so that during a component Update, the targets are deduplicated based on the container ID, keeping only the first one.

Related to: #4403

I'd like to gather some more evidence until we say this fixes #4403 outright, so keeping this PR in draft in the meantime.

#### Notes to the Reviewer

I've run a simplistic benchmark with a chatty container that exposes two ports, and as expected, the CPU time spent on reading logs is ~half here.
image

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Documentation added
- [X] Tests updated

### DIFF
--- a/docs/sources/flow/reference/components/loki.source.docker.md
+++ b/docs/sources/flow/reference/components/loki.source.docker.md
@@ -131,11 +131,11 @@ fully qualified name) to store its _positions file_. The positions file
 stores the read offsets so that if there is a component or Agent restart,
 `loki.source.docker` can pick up tailing from the same spot.
 
-In case the targets argument contains multiple entries with the same container
+If the target's argument contains multiple entries with the same container
 ID (for example as a result of `discovery.docker` picking up multiple exposed
-ports or networks) `loki.source.docker` will deduplicate them, and only keep
+ports or networks), `loki.source.docker` will deduplicate them, and only keep
 the first of each container ID instances, based on the
-`__meta_docker_container_id` label.  As such, the Docker daemon will be queried
+`__meta_docker_container_id` label.  As such, the Docker daemon is queried
 for each container ID only once, and only one target will be available in the
 component's debug info.
 
--- a/docs/sources/flow/reference/components/loki.source.docker.md
+++ b/docs/sources/flow/reference/components/loki.source.docker.md
@@ -131,9 +131,9 @@ fully qualified name) to store its _positions file_. The positions file
 stores the read offsets so that if there is a component or Agent restart,
 `loki.source.docker` can pick up tailing from the same spot.
 
-In case the targets Argument contains multiple targets with the same container
-ID, for example as a result of `discovery.docker` picking up multiple exposed
-ports or networks, `loki.source.docker` will deduplicate these, keeping only
+In case the targets argument contains multiple entries with the same container
+ID (for example as a result of `discovery.docker` picking up multiple exposed
+ports or networks) `loki.source.docker` will deduplicate them, and only keep
 the first of each container ID instances.
 
 ## Example
--- a/docs/sources/flow/reference/components/loki.source.docker.md
+++ b/docs/sources/flow/reference/components/loki.source.docker.md
@@ -131,6 +131,11 @@ fully qualified name) to store its _positions file_. The positions file
 stores the read offsets so that if there is a component or Agent restart,
 `loki.source.docker` can pick up tailing from the same spot.
 
+In case the targets Argument contains multiple targets with the same container
+ID, for example as a result of `discovery.docker` picking up multiple exposed
+ports or networks, `loki.source.docker` will deduplicate these, keeping only
+the first of each container ID instances.
+
 ## Example
 
 This example collects log entries from the files specified in the `targets`
--- a/docs/sources/flow/reference/components/loki.source.docker.md
+++ b/docs/sources/flow/reference/components/loki.source.docker.md
@@ -134,7 +134,10 @@ stores the read offsets so that if there is a component or Agent restart,
 In case the targets argument contains multiple entries with the same container
 ID (for example as a result of `discovery.docker` picking up multiple exposed
 ports or networks) `loki.source.docker` will deduplicate them, and only keep
-the first of each container ID instances.
+the first of each container ID instances, based on the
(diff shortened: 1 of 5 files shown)


rfratto (approved):
Ultimately I think this makes sense, but I think we need to document it, since it might be surprising if you effectively input three targets (with the same ID) into loki.source.docker but the debug info only shows one.

tpaschalis:
> I think we need to document it, since it might be surprising if you effectively input three targets (with the same ID) into loki.source.docker but the debug info only shows one.

I've added a paragraph to document this new behavior, hopefully it clears things up.

rfratto (approved):
LGTM but the docs change should get a review by @clayton-cornell

docs/sources/flow/reference/components/loki.source.docker.md
```diff
@@ -131,6 +131,14 @@ fully qualified name) to store its _positions file_. The positions file
 stores the read offsets so that if there is a component or Agent restart,
 `loki.source.docker` can pick up tailing from the same spot.
 
+In case the targets argument contains multiple entries with the same container
```
clayton-cornell, line 134:
```suggestion
If the target's argument contains multiple entries with the same container
```

docs/sources/flow/reference/components/loki.source.docker.md
```diff
@@ -131,6 +131,14 @@ fully qualified name) to store its _positions file_. The positions file
...
+ID (for example as a result of `discovery.docker` picking up multiple exposed
+ports or networks) `loki.source.docker` will deduplicate them, and only keep
```
clayton-cornell, line 136:
```suggestion
ID (for example as a result of `discovery.docker` picking up multiple exposed
ports or networks), `loki.source.docker` will deduplicate them, and only keep
```

Generally, we should avoid using brackets in the text. I'm leaving this one in for now because removing it makes for a VERY long sentence (or forces a big rewrite).

docs/sources/flow/reference/components/loki.source.docker.md
```diff
@@ -131,6 +131,14 @@ fully qualified name) to store its _positions file_. The positions file
...
+the first of each container ID instances, based on the
+`__meta_docker_container_id` label.  As such, the Docker daemon will be queried
```
clayton-cornell, line 138:
```suggestion
`__meta_docker_container_id` label.  As such, the Docker daemon is queried
```

clayton-cornell (commented):
Some minor changes to the doc input

PR #6059, opened 2024-01-06
chore(build): remove seemingly obsolete exclude directive

hainenber:
#### PR Description
Remove seemingly obsolete `exclude` directive. After removal, I was able to run `go mod tidy` but I still don't know what test got failed due to this directive back then. @rfratto can you help recall this one? :D


rfratto (approved):
I don't remember, but if it works now that's all that matters. LGTM :)

PR #6062, opened 2024-01-06
fix: ensure the chart works when using image digests

claudioscalzo:
#### Notes to the Reviewer

- Fix a bug preventing the `.Values.configReloader.image.digest` Helm value to be correctly retrieved.
- Fix a bug preventing digests to be used as labels because of their length.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated
- [ ] Documentation added


tpaschalis:
Hey Claudio, thank you for your contribution to Grafana Agent!

Could we add a new test for this by creating a new values.yaml file on `operations/helm/charts/grafana-agent/ci` and then running `make generate-helm-tests`?

I think it's going to make it clearer what the fix is about, and avoid breaking it accidentally in the future.

operations/helm/charts/grafana-agent/CHANGELOG.md
```diff
@@ -10,6 +10,11 @@ internal API changes are not present.
 Unreleased
 ----------
 
+### Bugfixes
+
+- Fix a bug preventing the `.Values.configReloader.image.digest` Helm value to be correctly retrieved. (@claudioscalzo)
+- Fix a bug preventing digests to be used as labels because of their length. (@claudioscalzo)
```
tpaschalis, line 16: Is there a documented limit for the length of this label that we can point to? Why choose 7 characters here?

claudioscalzo: Yes, added the documented limit to the changelog entry ✅ I chose a 7-character string because it's the standard for short SHAs

claudioscalzo:
Hi @tpaschalis 👋🏻 Thanks!
I've added the test values and expanded the changelog ✅

tpaschalis:
I think we're on the right path @claudioscalzo!

Two things: The change broke one of the previous test. Also, the new test case doesn't seem to populate `app.kubernetes.io/version` with the digest? I'll have to take a closer look though.

claudioscalzo:
Hi @tpaschalis, thanks, the CI tests should now pass (a typo prevented them to be successful)!

Regarding the missing digests on the pre-rendered tests, the behaviour you're noticing is due to [this existing piece of code](https://github.com/grafana/agent/blob/28846b29f3e244f0c70030e9643a55f7521e3aeb/operations/helm/charts/grafana-agent/templates/_helpers.tpl#L54-L57) which always sets a static string for those labels inside test templates. I'm not aware of the reasons behind this choice but I wouldn't change this behavior inside this PR. Let me know what you think!

PR #6068, opened 2024-01-08
Fix windows tests 3

erikbaranowski:
#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/component/common/loki/wal/watcher_test.go
+++ b/component/common/loki/wal/watcher_test.go
@@ -685,9 +685,9 @@ func TestWatcher_StopAndDrainWAL(t *testing.T) {
 		watcher.Drain()
 		watcher.Stop()
 
-		// expecting 15s (missing 15 entries * 1 sec delay in AppendEntries) +/- 1.1s (taking into account the drain timeout
+		// expecting 15s (missing 15 entries * 1 sec delay in AppendEntries) +/- 2.0s (taking into account the drain timeout
 		// has one extra second.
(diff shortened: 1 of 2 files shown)


component/common/loki/wal/watcher_test.go
```diff
@@ -685,9 +685,9 @@ func TestWatcher_StopAndDrainWAL(t *testing.T) {
 		watcher.Drain()
 		watcher.Stop()
 
-		// expecting 15s (missing 15 entries * 1 sec delay in AppendEntries) +/- 1.1s (taking into account the drain timeout
+		// expecting 15s (missing 15 entries * 1 sec delay in AppendEntries) +/- 2.0s (taking into account the drain timeout
 		// has one extra second.
-		require.InDelta(t, time.Second*15, time.Since(now), float64(time.Millisecond*1100), "expected the drain procedure to take around 15s")
+		require.InDelta(t, time.Second*15, time.Since(now), float64(time.Millisecond*2000), "expected the drain procedure to take around 15s")
```
erikbaranowski, line 690: @thepalbi These failures are sporadic for windows testing. If you feel this should be done a different way to make them more resilient let me know.

mattdurham: file io tests on windows are always a flaky. Sounds reasonable but if it gets to flaky we may have to make it linux only.

mattdurham (approved):
lgtm

component/common/loki/wal/watcher_test.go
thepalbi: I'll take a look after to see if there's a way to make them more resilient

PR #6073, opened 2024-01-09
Create remote git component from git module

wildum:
#### PR Description

With this PR we move the git logic out of the git module to create a new component remote.git.
Both the current git module and the upcoming new git module will use this component to retrieve the module content, in a similar manner as the remote.http.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


component/remote/git/git.go
```diff
@@ -0,0 +1,287 @@
...
+		tickerC <-chan time.Time
+	)
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+
+		case <-c.argsChanged:
+			c.mut.Lock()
```
thampiotr, line 131: For readability, you can just extract the `PullFrequency` into local variable while holding the mutex, and then the rest of the code doesn't need to hold `mut`.

docs/sources/flow/reference/components/remote.git.md
```diff
@@ -0,0 +1,120 @@
...
+`htts://github.com/grafana/agent.git`.
+
+The `revision` attribute, when provided, must be set to a valid branch, tag, or
+commit SHA within the repository.
+
+The `path` attribute must be set to a path which is accessible from the root of
+the repository, such as `FILE_NAME.river` or `FOLDER_NAME/FILE_NAME.river`.
+
+If `pull_frequency` is not `"0s"`, the Git repository will be pulled for
+updates at the frequency specified.
```
thampiotr, line 48: If it is `0s`, will it be pulled once only? or never?

docs/sources/flow/reference/components/remote.git.md
```diff
@@ -0,0 +1,120 @@
...
+- /docs/grafana-cloud/send-data/agent/flow/reference/components/remote.git/
+canonical: gits://grafana.com/docs/agent/latest/flow/reference/components/remote.git/
+description: Learn about remote.git
+title: remote.git
+---
+
+# remote.git
+
+`remote.git` retrieves the content of file stored in a git repository and exposes it to other components. The repo
+is polled for changes so that the most recent content is always available.
```
thampiotr, line 15:
```suggestion
is polled for changes so that the most recent content is eventually available.
```

docs/sources/flow/reference/components/remote.git.md
```diff
@@ -0,0 +1,120 @@
...
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`repository` | `string` | The Git repository address to retrieve the content from. | | yes
+`revision` | `string` | The Git revision to retrieve the content from. | `"HEAD"` | no
+`path` | `string` | The path in the repository where the content is stored. | | yes
+`pull_frequency` | `duration` | The frequency to pull the repository for updates. | `"60s"` | no
+
+The `repository` attribute must be set to a repository address that would be
+recognized by Git with a `git clone REPOSITORY_ADDRESS` command, such as
+`htts://github.com/grafana/agent.git`.
```
thampiotr, line 39:
```suggestion
`https://github.com/grafana/agent.git`.
```

We should probably also use `git` protocol URL to be super clear.

thampiotr (commented):
Got a few comments, so will stop here for the first pass. 

Have you tried to run this locally and tested updating the config, wrong URLs, etc?

component/module/git/git.go
```diff
@@ -80,190 +63,108 @@ func New(o component.Options, args Arguments) (*Component, error) {
...
 	return c, nil
 }
 
+// NewManagedLocalComponent creates the new remote.git managed component.
+func (c *Component) newManagedLocalComponent(o component.Options) (*remote_git.Component, error) {
+	remoteGitOpts := o
+	remoteGitOpts.OnStateChange = func(e component.Exports) {
+		c.setContent(e.(remote_git.Exports).Content)
+
+		if !c.inUpdate.Load() && c.isCreated.Load() {
```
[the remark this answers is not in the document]
wildum: This is part of the old modules and it will be deprecated once the new modules are ready, I think that we can keep it like this

docs/sources/flow/reference/components/remote.git.md
```diff
@@ -0,0 +1,120 @@
...
+
+The `repository` attribute must be set to a repository address that would be
+recognized by Git with a `git clone REPOSITORY_ADDRESS` command, such as
+`https://github.com/grafana/agent.git`.
+
+The `revision` attribute, when provided, must be set to a valid branch, tag, or
+commit SHA within the repository.
+
+The `path` attribute must be set to a path which is accessible from the root of
+the repository, such as `FILE_NAME.river` or `FOLDER_NAME/FILE_NAME.river`.
```
tpaschalis, line 45: (Haven't read the code yet, apologies, but could be worth adding here).

What happens when the path or revision are invalid? Does the component error out and exit the Agent on startup? Will it retry at the specified interval? If the component does run, is the content Export left empty or null?

component/module/git/git.go
```diff
@@ -28,43 +24,30 @@ func init() {
...
+	args    Arguments
+	content string
 
-	argsChanged chan struct{}
-
-	healthMut sync.RWMutex
-	health    component.Health
+	managedRemoteGit *remote_git.Component
+	inUpdate         atomic.Bool
+	isCreated        atomic.Bool
```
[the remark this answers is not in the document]
wildum: I was missing the `defer c.isCreated.Store(true)`, I took the same pattern as in the HTTP remote to keep it consistent. I'm not looking for a refactoring because this code should be soon deprecated

component/module/git/git.go
```diff
@@ -80,190 +63,108 @@ func New(o component.Options, args Arguments) (*Component, error) {
...
+	// on initial load.
+	return c.mod.LoadFlowSource(newArgs.Arguments, c.getContent())
 }
 
 // CurrentHealth implements component.HealthComponent.
 func (c *Component) CurrentHealth() component.Health {
-	c.healthMut.RLock()
-	defer c.healthMut.RUnlock()
+	return component.LeastHealthy(c.managedRemoteGit.CurrentHealth(), c.mod.CurrentHealth())
+}
```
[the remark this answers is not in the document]
wildum: ok I see, that's a very weird pattern but I'm fine doing the same for this file

component/remote/git/git.go
```diff
@@ -0,0 +1,287 @@
...
+			Message:    "component started",
+			UpdateTime: time.Now(),
+		},
+	}
+
+	// Only acknowledge the error from Update if it's not a
+	// vcs.UpdateFailedError; vcs.UpdateFailedError means that the Git repo
+	// exists but we were just unable to update it.
+	if err := c.Update(args); err != nil {
+		if errors.As(err, &vcs.UpdateFailedError{}) {
```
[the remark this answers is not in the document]
wildum: it can hit this line via Update > pollFile > repo.Update

docs/sources/flow/reference/components/remote.git.md
wildum: If the agent does not have a local copy of the repo corresponding to the config in the river file, then it will error out and exit. Else, it will use the local copy content and mark itself as unhealthy because it could not update it with the latest content

docs/sources/flow/reference/components/remote.git.md
wildum: it is pulled once

component/remote/git/git.go
```diff
@@ -0,0 +1,287 @@
...
+			} else {
+				return err
+			}
+		}
+		c.repo = r
+		c.repoOpts = repoOpts
+	}
+
+	if err := c.pollFile(context.Background(), newArgs); err != nil {
+		return err
```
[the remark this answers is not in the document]
wildum: fixed by moving the error handling previously in the New function

component/remote/git/git.go
```diff
@@ -0,0 +1,287 @@
...
+			UpdateTime: time.Now(),
+		},
+	}
+
+	// Only acknowledge the error from Update if it's not a
+	// vcs.UpdateFailedError; vcs.UpdateFailedError means that the Git repo
+	// exists but we were just unable to update it.
+	if err := c.Update(args); err != nil {
+		if errors.As(err, &vcs.UpdateFailedError{}) {
+			level.Error(c.log).Log("msg", "failed to update repository", "err", err)
```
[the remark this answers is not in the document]
wildum: it will be retried only if the pullFrequency is != 0. If so, this log line will trigger right after: 	level.Info(c.log).Log("msg", "updating repository pull frequency, next pull attempt planned in:", "new_frequency", pullFrequency). I think that this is ok?

wildum:
I tried the remote.git component locally in various scenarios:
- wrong url
- wrong revision
- wrong path
- wrong pull frequency
- correct config > remove file from repo to check unhealthy state > put file back to check recovery
For the module.git I added an integration test

docs/sources/flow/reference/components/remote.git.md
```diff
@@ -0,0 +1,120 @@
...
+
+If `pull_frequency` is not `"0s"`, the Git repository will be pulled for
```
erikbaranowski, line 47: Should we add something here about how pulling hosted git repositories too often can result in throttling? This is a pitfall several folks have run into.

wildum: good idea, I added a warning in this doc and in the module doc

component/remote/git/git.go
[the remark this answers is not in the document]
thampiotr: Yeah. Or just "will retry in: ...".

docs/sources/flow/reference/components/remote.git.md
thampiotr: sorry, I wasn't asking for an answer, I wanted to suggest adding this to docs

component/remote/git/git.go
```diff
@@ -0,0 +1,287 @@
...
+	Repository    string            `river:"repository,attr"`
+	Revision      string            `river:"revision,attr,optional"`
+	Path          string            `river:"path,attr"`
+	PullFrequency time.Duration     `river:"pull_frequency,attr,optional"`
+	GitAuthConfig vcs.GitAuthConfig `river:",squash"`
+}
+
+// Exports holds settings exported by remote.git.
+type Exports struct {
+	Content string `river:"content,attr"`
```
thampiotr, line 43: Should this also be an OptionalSecret like in remote.http?
It's not good to keep secrets in git, but it's not unheard of... and better if we don't expose it in the UI? 🤔

rfratto:
Personally, I think it's too early for us to add a remote.git component. Modules are still in active development, and modules are the primary intended usage of Git, where we still have to figure out things like what happens when using a folder instead of a single file, and changes there might influence how a remote.git component could _eventually_ work. 

If I'm alone in these concerns, I would instead encourage this component to be introduced as an experimental component rather than stable like it is now. But I would personally prefer deferring a remote.git component until we've fleshed out the Git use case for modules more.

cc @tpaschalis @thampiotr

rfratto:
(cc @erikbaranowski too as the other reviewer; sorry, missed you on the first pass)

erikbaranowski:
I am personally in favor of supporting a remote component for git outside of the modules use case. If we feel the timing is wrong or risk too high to make this change I am ok holding off since I also don't see it has a high priority need.

I also agree that remote.git should look to solve the problem in a generic way and that should mean eventually handling multiple files in the repo and caching in case the remote is down.

docs/sources/flow/reference/components/module.git.md
```diff
@@ -60,7 +60,9 @@ the repository, such as `FILE_NAME.river` or `FOLDER_NAME/FILE_NAME.river`.
 
 If `pull_frequency` is not `"0s"`, the Git repository will be pulled for
 updates at the frequency specified, causing the loaded module to update with
-the retrieved changes.
+the retrieved changes. If it is set to `"0s"`, the Git repository will be pulled once on init.
+
+**_WARNING:_** Pulling hosted git repositories too often can result in throttling.
```
clayton-cornell, line 65:
```suggestion
{{% admonition type="warning" %}}
Pulling hosted Git repositories too often can result in throttling.
{{% /admonition %}}
```

docs/sources/flow/reference/components/remote.git.md
```diff
@@ -0,0 +1,122 @@
...
+- /docs/grafana-cloud/monitor-infrastructure/integrations/agent/flow/reference/components/remote.git/
+- /docs/grafana-cloud/send-data/agent/flow/reference/components/remote.git/
+canonical: gits://grafana.com/docs/agent/latest/flow/reference/components/remote.git/
+description: Learn about remote.git
+title: remote.git
+---
+
+# remote.git
+
+`remote.git` retrieves the content of file stored in a git repository and exposes it to other components. The repo
```
clayton-cornell, line 14:
```suggestion
`remote.git` retrieves the content of a file stored in a Git repository and exposes it to other components. The repository
```

docs/sources/flow/reference/components/remote.git.md
```diff
@@ -0,0 +1,122 @@
...
+Name | Type | Description | Default | Required
+---- | ---- | ----------- | ------- | --------
+`repository` | `string` | The Git repository address to retrieve the content from. | | yes
+`revision` | `string` | The Git revision to retrieve the content from. | `"HEAD"` | no
+`path` | `string` | The path in the repository where the content is stored. | | yes
+`pull_frequency` | `duration` | The frequency to pull the repository for updates. | `"60s"` | no
+
+The `repository` attribute must be set to a repository address that would be
+recognized by Git with a `git clone REPOSITORY_ADDRESS` command, such as
+`https://github.com/grafana/agent.git`.
```
clayton-cornell, line 39:
```suggestion
You must set the `repository` attribute to a repository address that Git would recognize
with a `git clone REPOSITORY_ADDRESS` command, such as `https://github.com/grafana/agent.git`.
```

docs/sources/flow/reference/components/remote.git.md
```diff
@@ -0,0 +1,122 @@
...
+
+The `revision` attribute, when provided, must be set to a valid branch, tag, or
+commit SHA within the repository.
```
clayton-cornell, line 42:
```suggestion
When provided, the `revision` attribute must be set to a valid branch, tag, or
commit SHA within the repository.
```

docs/sources/flow/reference/components/remote.git.md
```diff
@@ -0,0 +1,122 @@
...
+
+The `path` attribute must be set to a path which is accessible from the root of
+the repository, such as `FILE_NAME.river` or `FOLDER_NAME/FILE_NAME.river`.
```
clayton-cornell, line 45:
```suggestion
You must set the `path` attribute to a path accessible from the repository's root,
such as `FILE_NAME.river` or `FOLDER_NAME/FILE_NAME.river`.
```

docs/sources/flow/reference/components/remote.git.md
```diff
@@ -0,0 +1,122 @@
...
+
+If `pull_frequency` is not `"0s"`, the Git repository will be pulled for
+updates at the frequency specified. If it is set to `"0s"`, the Git repository will be pulled once on init.
+
+**_WARNING:_** Pulling hosted git repositories too often can result in throttling.
```
clayton-cornell, line 50:
```suggestion
{{% admonition type="warning" %}}
Pulling hosted Git repositories too often can result in throttling.
{{% /admonition %}}
```

docs/sources/flow/reference/components/remote.git.md
```diff
@@ -0,0 +1,122 @@
...
+recognized by Git with a `git clone REPOSITORY_ADDRESS` command, such as
+`https://github.com/grafana/agent.git`.
+
+The `revision` attribute, when provided, must be set to a valid branch, tag, or
+commit SHA within the repository.
+
+The `path` attribute must be set to a path which is accessible from the root of
+the repository, such as `FILE_NAME.river` or `FOLDER_NAME/FILE_NAME.river`.
+
+If `pull_frequency` is not `"0s"`, the Git repository will be pulled for
```
clayton-cornell, line 47:
```suggestion
If `pull_frequency` is not `"0s"`, the Git repository is pulled for
```

docs/sources/flow/reference/components/remote.git.md
```diff
@@ -0,0 +1,122 @@
...
+updates at the frequency specified. If it is set to `"0s"`, the Git repository will be pulled once on init.
```
clayton-cornell, line 48:
```suggestion
updates at the frequency specified. If it is set to `"0s"`, the Git repository is pulled once on init.
```

clayton-cornell (commented):
Some doc suggestions

thampiotr:
To add my 2p to the discussion on whether we need this: 

* I agree it's best to mark it as experimental. We should probably always start with experimental or beta stability label.
* Even if we refactor to have a git service, I think it's still useful to have remote.git? We can refactor it then. The code duplication is smaller.
* @wildum has raised this offline to discuss on 15th Dec and it looked like it had more voices in favour than against. But of course, we can discuss further. 

Is there a known issue or re-work that would need to be done if we switched to using a git service? Or is the concern that this is just more things using git that would need switching to use the new service?

component/remote/git/git.go
thampiotr: I would want it to be documented in `remote.git`, just the way that it's implemented in [remote.http](https://grafana.com/docs/agent/next/flow/reference/components/remote.http/). I don't think it would impact `module.git` docs.

tpaschalis:
I'm personally in favor of potentially waiting for the new modules to hit the floor and _then_ implementing this new component.

Since most popular Git SaaS implementations (GitHub, Gitlab, Gitea) seem to support fetching the raw contents of a file both for private and public repos, the simplest of use cases are well handled with remote.http as-is, and we could look for more advanced ones to be covered by this new component.

rfratto:
> Is there a known issue or re-work that would need to be done if we switched to using a git service? Or is the concern that this is just more things using git that would need switching to use the new service?

I'm primarily concerned that there's other higher priority things to do for 1.0 right now that will influence the design of a generalized component which gets content from Git. 

In other words, two things concern me:

* Whether merging and supporting this takes precedence over other priorities 
* Whether this is prone to breaking changes given the uncertainty around how our interaction with Git will change over time

wildum:
Thank you all for your feedback. This feature was not part of a specific request, I just thought it might be a good idea but I understand the concerns. I will close the PR for now, we can re-open the discussions and come up with a stronger design when the timing feels right :)

Issue #6077, opened 2024-01-09
Remove the `production/` folder in favor of `operations/` 


rfratto:
### Background

The `production/` folder is composed of four pieces:

* `grafana-agent-mixin/`: A Jsonnet mixin for static mode monitoring
* `kubernetes/`: Kubernetes manifests to deploy the agent in static mode
* `operator/`: CRDs and manifests for deploying the operator 
* `tanka/`: Tanka configs for deploying both static mode and the operator 

The `production/` folder has been effectively abandoned over time in favor of the new `operations/` folder, which contains:

* `agent-flow-mixin/`: A Jsonnet mixin for flow mode monitoring.
* `helm/`: Helm chart(s) for Grafana Agent.
* `river-jsonnet/`: A Jsonnet library for writing River in Jsonnet. 

Having both of these directories is confusing, as it's not obvious (unless you look at Git history) which is newer. Additonally, some of the subdirectories in `production/` are effectively abandoned. 

`production/` should be merged into `operations/`, and abandoned subdirectories should be removed. 

### Proposal

Merge non-abandoned subdirectories from `production/` into `operations/`: 

- [ ] Move `production/grafana-agent-mixin` to `operations/agent-static-mixin` 
- [ ] Remove `production/kubernetes` (deprecated by `operations/helm`)
- [ ] Move `production/operator` to `operations/agent-static-operator`, and update the Makefile at the root of the repository to generate CRD manifests in the new directory.
- [ ] Remove `production/tanka` (deprecated by `operations/helm`) 

Care must be taken before removing these directories to ensure that no documentation points at this directory either from this repo or within Grafana Cloud instructions.

rfratto:
I poked through everything, and here are the main issues:

External references to `production/kubernetes`: 

* Our own documentation for [Deploy Grafana Agent in static mode on Kubernetes](https://grafana.com/docs/agent/latest/static/set-up/install/install-agent-kubernetes/)
* Grafana Cloud documentation for [Upgrade Kubernetes Monitoring](https://grafana.com/docs/grafana-cloud/monitor-infrastructure/kubernetes-monitoring/configuration/upgrade-k8s/)

External references to `production/operator`:

* [Operator Helm chart](https://github.com/grafana/helm-charts/blob/7d2030e19ab7f2e253f102ce83de3fade8fe5b80/charts/agent-operator/README.md.gotmpl#L19) references this directory. 
[...]

PR #6079, opened 2024-01-09
docs: remove reference to production/kubernetes

rfratto:
This commit modifies the Kubernetes installation instructions for static mode to use the Grafana Agent Helm chart instead of the abandoned production/kubernetes manifests.

This change needs to be backported pretty far back. References to the old folder exist as far back as v0.25, but we should only backport this as far back as the introduction of Flow mode, and find an alternative for versions prior to that (such as version pinning the old directory instead of having it use the main branch). 

Related to #6077.

### DIFF
--- a/docs/sources/static/set-up/install/install-agent-kubernetes.md
+++ b/docs/sources/static/set-up/install/install-agent-kubernetes.md
@@ -11,7 +11,7 @@ weight: 300
 
 # Deploy Grafana Agent in static mode on Kubernetes
 
-Grafana Agent in static mode can be deployed on Kubernetes by using the Helm chart for Grafana Agent.
+You can use the Helm chart for Grafana Agent to deploy Grafana Agent in static mode on Kubernetes.
 
 ## Before you begin
 
--- a/docs/sources/static/set-up/install/install-agent-kubernetes.md
+++ b/docs/sources/static/set-up/install/install-agent-kubernetes.md
@@ -11,40 +11,53 @@ weight: 300
 
 # Deploy Grafana Agent in static mode on Kubernetes
 
-You can deploy Grafana Agent in static mode on Kubernetes.
+Grafana Agent in static mode can be deployed on Kubernetes by using the Helm chart for Grafana Agent.
+
+## Before you begin
+
+* Install [Helm][] on your computer.
+* Configure a Kubernetes cluster that you can use for Grafana Agent.
+* Configure your local Kubernetes context to point to the cluster.
+
+[Helm]: https://helm.sh
 
 ## Deploy
 
-To deploy Grafana Agent in static mode on Kubernetes, perform the following steps.
-
-1. Download one of the following manifests from GitHub and save it as `manifest.yaml`:
-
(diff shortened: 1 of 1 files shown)


tpaschalis:
Also, could maybe just backporting a symlink partly help with this?

rfratto:
The risk here is that every instance of our versioned documentation (for this page) points to the main branch, so old documentation will no longer work unless we go back to pin the versions or update it. I'd leave it @clayton-cornell to decide whether that's a risk.

> Also, could maybe just backporting a symlink partly help with this?

If we do a symlink then we can't get rid of the folder 😞

tpaschalis (approved):

clayton-cornell:
Generally, we only support three versions back from `next`. There's a lot of things in the older docs that no longer work. The risk with not backporting is only realised if we have customers on really old versions of Agent who will be trying to use these instructions. Do we have any idea what the oldest Agent is out there in the wild?

Automatic backporting will start to fail pretty quickly as we move back through the versions, and it'll become a fairly manual process... to the point of being a rather big task to untangle all the broken bits.

What is actually supported within Agent? Do we really support all versions of Agent back to 0.25? or back to the introduction of Flow?

rfratto:
> Do we have any idea what the oldest Agent is out there in the wild?

Based on usage stats, there's people running v0.25.1. I don't know when usage stats were introduced, so there could be users on versions older than these. 

> Automatic backporting will start to fail pretty quickly as we move back through the versions, and it'll become a fairly manual process... to the point of being a rather big task to untangle all the broken bits.

As an alternative, the older versions can be updated to point at a release tag instead of the main branch so the old links continue working. This is still some work compared to not changing older versions at all, but it might be a good middle ground. 

> Generally, we only support three versions back from next.
> 
> What is actually supported within Agent? Do we really support all versions of Agent back to 0.25? or back to the introduction of Flow?

I don't know if we have an official stance. People run very old versions, and we sometimes offer limited support for it, but we always recommend upgrading. The most 2 or 3 recent versions is where I'd usually draw the line. 

I'm personally comfortable with this only being backported to the most 3 recent versions. If you want me to do the workaround above (i.e., don't point at main in docs versions older then v0.36), I'm also comfortable doing that. 

WDYT @clayton-cornell?

clayton-cornell:
@rfratto I'm good with either focusing on the previous 3 release or the middle ground idea. I guess it's a case of work to do vs benefit. End of the day, if it helps the user base then we should do it :-)

docs/sources/static/set-up/install/install-agent-kubernetes.md
```diff
@@ -11,40 +11,53 @@ weight: 300
 
 # Deploy Grafana Agent in static mode on Kubernetes
 
-You can deploy Grafana Agent in static mode on Kubernetes.
+Grafana Agent in static mode can be deployed on Kubernetes by using the Helm chart for Grafana Agent.
```
clayton-cornell, line 14:
```suggestion
You can use the Helm chart for Grafana Agent to deploy Grafana Agent in static mode on Kubernetes.
```

clayton-cornell (commented):
One small suggestion ti try and make the sentence a little more active.

clayton-cornell (approved):

PR #6080, opened 2024-01-09
misc: move folders to operations/ and create temporary symlinks

rfratto:
This change copies folders from `production/` to `operations/` that we intend to keep (the static mode mixin and operator related manifests).

I tried to originally make symlinks and move the folders, but this doesn't work in the GitHub UI, where it fails to navigate to a file within a symlinked directory. Since there are references to such files, symlinks won't work for us here. 

Once references to the productions folder are completely eliminated, the folder along with its copies will be removed.

Related to #6077.


tpaschalis (approved):
LGTM

rfratto:
Actually, this won't work. GitHub 404s if you try to navigate to https://github.com/rfratto/agent/blob/production-migrate-folders/production/kubernetes/agent-bare.yaml. It appears GitHub doesn't know how to navigate symlinks.

rfratto:
It might be worthwhile to make a copy instead of symlinking for the brief period where the old `production/` folder still exists.

PR #6082, opened 2024-01-09
loki.process: document limitation of stage.json

### LINKED ISSUE #5838: stage.json of loki.process cannot have hyphen (`-`) in the expressions value
### What's wrong?

This is actually a bug in the `jmespath/go-jmespath` - https://github.com/jmespath/go-jmespath/issues/58
but I wanted to bring it to your attention as well, as it was hard to debug

Workaround for me is to define it like this:
```
    stage.json {
      expressions = {
        http_user_agent = "\"request_User-Agent\"",
      }
    }
```

### Steps to reproduce

Try to load the provided config.

### System information

6.5.10-200.fc38.aarch64

### Software version

grafana-agent-flow-0.38.0-1.aarch64

### Configuration

```text
loki.process "enrich" {

  stage.match {
    selector = "{container=\"traefik\"}"

    stage.json {
      expressions = {
        http_user_agent = "request_User-Agent",
      }
    }
  }
}
```

### Logs

```text
Error: /etc/grafana-agent-flow.river:308:1: Failed to build component: building component: invalid stage config invalid stage config could not compile JMES expression: SyntaxError: Unexpected token at the end of the expression: tNumber: match stage failed to create pipeline from config: {{container="traefik"} [{<nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> 0x4002ab7618 <nil> <nil> <nil> <nil> <nil>} {<nil> <nil> <nil> <nil> <nil> <nil> 0x4002bf4768 <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil>} {<nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> 0x4002ab7638 <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil> <nil>}] keep  }
```

tpaschalis:
#### PR Description
This PR documents a limitation of the jmespath library used to parse JSON expressions for the json stage of the loki.process component.

Fixes #5838

#### PR Checklist

- [X] Documentation added
- [ ] Tests updated (N/A)

### DIFF
--- a/docs/sources/flow/reference/components/loki.process.md
+++ b/docs/sources/flow/reference/components/loki.process.md
@@ -382,6 +382,27 @@ following key-value pair to the set of extracted data.
 username: agent
 ```
 
+{{% admonition type="note" %}}
+Due to a limitation of the upstream jmespath library, defining an expression
+that contains a hyphen `-` needs to be wrapped in quotes so that it's not
+considered a numeric expression.
+{{% /admonition %}}
+
+In this case, you can use one of two syntaxes to circumvent this issue:
+```
+stage.json {
+  expressions = {
+    http_user_agent = "\"request_User-Agent\"",
+  }
+}
+
+stage.json {
+  expressions = {
+    http_user_agent = `"request_User-Agent"`,
+  }
+}
+```
+
 ### stage.label_drop block
 
 The `stage.label_drop` inner block configures a processing stage that drops labels
--- a/docs/sources/flow/reference/components/loki.process.md
+++ b/docs/sources/flow/reference/components/loki.process.md
@@ -383,9 +383,12 @@ username: agent
 ```
 
 {{% admonition type="note" %}}
-Due to a limitation of the upstream jmespath library, defining an expression
-that contains a hyphen `-` needs to be wrapped in quotes so that it's not
-considered a numeric expression.
+Due to a limitation of the upstream jmespath library, any string that contains
+a hyphen `-` needs to be wrapped in quotes so that it's not considered a
+numerical expression.
+
+Otherwise, you might get errors like:
+`Unexpected token at the end of the expression: tNumber`
 {{% /admonition %}}
 
 In this case, you can use one of two syntaxes to circumvent this issue:


docs/sources/flow/reference/components/loki.process.md
```diff
@@ -382,6 +382,30 @@ following key-value pair to the set of extracted data.
...
+{{% admonition type="note" %}}
+Due to a limitation of the upstream jmespath library, any string that contains
+a hyphen `-` needs to be wrapped in quotes so that it's not considered a
+numerical expression.
+
+Otherwise, you might get errors like:
+`Unexpected token at the end of the expression: tNumber`
+{{% /admonition %}}
+
+In this case, you can use one of two syntaxes to circumvent this issue:
```
clayton-cornell, line 394: @tpaschalis which issue is being dealt with here? The one in the note?

tpaschalis: Yes. Is it clear, or should I also include it on the note?

docs/sources/flow/reference/components/loki.process.md
```diff
@@ -382,6 +382,30 @@ following key-value pair to the set of extracted data.
 username: agent
 ```
 
+Due to a limitation of the upstream jmespath library, any string that contains
+a hyphen `-` needs to be wrapped in quotes so that it's not considered a
+numerical expression.
+
+Otherwise, you might get errors like:
+`Unexpected token at the end of the expression: tNumber`
```
clayton-cornell, line 391:
```suggestion
Due to a limitation of the upstream jmespath library, you must wrap any string that contains a hyphen `-` in quotes so that it's not considered a numerical expression.

If you don't use quotes to wrap a string that contains a hyphen, you will get errors like: `Unexpected token at the end of the expression: tNumber`

You can use one of two options to circumvent this issue:

1. An escaped double quote. For example: `http_user_agent = "\"request_User-Agent\""`
1. A backtick quote. For example: ``http_user_agent = `"request_User-Agent"` ``
```

Maybe we could rework it like this?  

Do we need the full stage.json in the examples? or is it enough to show a quoted string?

tpaschalis: Yeah, I like this suggestion, will fix this

clayton-cornell (approved):

PR #6087, opened 2024-01-09
Split 'Start agent' into per-platform pages

thampiotr:
#### PR Description

In order to implement the new docs structure, we want to move some pages around. But there is a slight problem with the `start-agent.md`:
* It is for all platforms in one page, while other pages are per-platform
* Keeping it in current shape wouldn't fit nicely into the new tasks nor into get-started section.

So this PR does the following:
* Split the `start-agent.md` into per-platform instructions pages
* Fix the links and add backwards-compatible aliases

Checked with `make docs` that there are no dead links.
Verified that the alias for the removed page still works with `curl`.

image

Next steps - roughly speaking - rename the `setup` into the new `get-started` folder.


docs/sources/flow/setup/start-agent.md
thampiotr, line 1: I was considering making this page `headless: true`, to hide it from the menu. But I think it may actually help a few people find what they're looking for, so opted to leave as-is.

clayton-cornell:
I'm not yet convinced this is the way to go here.  I don't want to block this, but I need to understand what the purpose and reasoning is before we move forward.

Generally, tasks should be focused on a single thing to do... a specific outcome and **install** and **run** are two different actions or outcomes.

Would it make more sense to arrange the TOC like this instead?

image

NOTE: I've stripped out the product name in this example TOC as part of a look forward where it really won't be necessary to include the specific product at child levels in the TOC.

I've also moved the **Run in Standalone** back into the **Start** section. It feels very out of place in a **Tasks** section - a place I'd expect to find things to do once my software is installed and running.

This example is more-or-less in alignment with the overall Grafana doc architecture - refer to this TOC for example: https://grafana.com/docs/grafana/latest/setup-grafana/start-restart-grafana/
It could certainly use more refinement... but serves as an example for now.

Yes, it makes sense to test your installation but running/starting the software, but... as a separate task, you can link to it as a next step. We can't, and shouldn't overload the tasks with everything in one page. :-)

thampiotr:
@clayton-cornell that's fine with me, I was under impression that another `Start` "folder" in the menu is undesirable, but I'm okay with it if we want it. Will change to this approach.

docs/sources/flow/setup/install/binary.md
```diff
@@ -45,12 +45,9 @@ To download {{< param "PRODUCT_NAME" >}} as a standalone binary, perform the fol
 
 ## Next steps
 
-- [Start {{< param "PRODUCT_NAME" >}}[Start]
-- [Configure {{< param "PRODUCT_NAME" >}}[Configure]
```
thampiotr, line 49: I've removed `Configure` link, because there's no configure page for standalone binary. In fact, the instructions on this page already describe how to specify a path to your config file. This is similar to what we have for docker and kubernetes install instructions. We can change this in the future, but I'd rather not do it in this PR.

docs/sources/flow/setup/start-agent.md
thampiotr, line 1: All the content from this file was copied to `start/*.md` files.

docs/sources/flow/setup/start/_index.md
```diff
@@ -0,0 +1,33 @@
...
+# Start {{% param "PRODUCT_NAME" %}}
+
+Use the following pages to learn how to start, restart, and stop {{< param "PRODUCT_NAME" >}} after it is installed.
+For installation instructions, refer to [Install {{< param "PRODUCT_NAME" >}}][Install].
+
+{{< section >}}
+
+{{% docs/reference %}}
+[data collection]: "/docs/agent/ -> /docs/agent/<AGENT_VERSION>/data-collection.md"
+[data collection]: "/docs/grafana-cloud/ -> /docs/grafana-cloud/send-data/agent/data-collection.md"
```
thampiotr, line 30: TODO: this needs to remove it

clayton-cornell (approved):

PR #6090, opened 2024-01-09
Fix handling of partial reads in remote.s3

### LINKED ISSUE #4704: remote.s3 not pulling in yaml file - empty string
### What's wrong?

Grafana Agent in flow mode installed in K8S.

I have a snmp.yml for the prometheus.exporter.snmp that I have stored in an S3 bucket.   I generated the yaml file since I needed to add a module that isn't in the default.  The UI shows that the S3 component is healthy, but the contents is an empty string "".   I validated the yaml file is valid.  

To see if it's a bucket permission issue, I uploaded a plain text file, with a simple string in it.  That file load successfully and displays it's contents in the UI.

The prometheus scrape of the snmp exporter is failing (even though it says healthy), due to "unknown module", which to me says it either isn't using the S3 file, or the S3 file isn't being loaded and therefore the module is missing in the default.

The agent console shows no errors.  There are no debug options and no ability to check the contents returned that I can tell as it's in memory.

### Steps to reproduce

Compile a new snmp.yml file with an additional module.
Upload the file to an S3 bucket.
Configure the river file to use remote.s3 to pull the file.
In the UI, it shows the contents as "" and the remote.s3 as healthy.

### System information

_No response_

### Software version

Grafana Agent v0.35.2 in flow mode

### Configuration

```text
remote.s3 "snmpconfig" {
      path = "s3://grafana-agent-bucket/snmp.yml"
      poll_frequency = "5m"
      is_secret = false
    }  <-- empty string returned

    remote.s3 "hi_folder" {
      path = "s3://grafana-agent-bucket/monitoring-cluster/hi.txt"
      poll_frequency = "5m"
      is_secret = false
    } <-- works fine

    remote.s3 "hi_root" {
      path = "s3://grafana-agent-bucket/hi.txt"
      poll_frequency = "5m"
      is_secret = false
    }  <-- works fine

    prometheus.exporter.snmp "netscaler" {
      config = remote.s3.snmpconfig.content

      target "publicProdNetscaler" {
        address = "172.0.0.123"
        module = "netscaler"
        auth = "network"
      }
      target "dmzNetscaler" {
        address = "netscalerADChostnameorip"
        module = "netscaler"
        auth = "network"
      }
    }

    prometheus.scrape "netscaler" {
      targets = prometheus.exporter.snmp.netscaler.targets
      forward_to = [prometheus.relabel.snmp.receiver]
    }



--------
Truncated snmp.yml

auths:
modules:
  if_mib:
  keepalived:
  netscaler:
```

### Logs

```text
Name	Value
job	
"prometheus.scrape.netscaler"
url	
"http://agent.internal:12345/api/v0/component/prometheus.exporter.snmp.netscaler/metrics?auth=network&module=netscaler&target=172.0.0.123"
health	
"down"
labels	
{
	instance = "k8s-workernode",
	job      = "integrations/snmp/publicProdNetscaler",
}
last_error	
"server returned HTTP status 400 Bad Request"
last_scrape	
"2023-08-03T16:47:17.021091026Z"
last_scrape_duration	
"373.166µs"

When I portfoward and hit that URL endpoint:

Unknown module 'netscaler'
```

spartan0x117:
#### PR Description
Properly error on unexpected EOF errors by attempting to read the full Content-Length's worth of data.

Fixes #4704 

#### PR Checklist

- [X] CHANGELOG.md updated


rfratto (approved):
Looks familiar!

component/remote/s3/watcher.go
```diff
@@ -100,10 +99,15 @@ func (w *watcher) getObject(ctx context.Context) ([]byte, error) {
 	if err != nil {
 		return []byte{}, err
 	}
+	defer output.Body.Close()
+
 	buf := make([]byte, output.ContentLength)
-	_, err = output.Body.Read(buf)
-	if !errors.Is(err, io.EOF) {
```
rfratto, line 105: I remember what the problem with this is: the issue isn't an unexpected EOF, but rather a partial read.

Implementation of io.Reader are permitted to do partial reads; which is especially common over the network where files may be streaming. The documentation for io.Reader states:

> Callers should always process the n > 0 bytes returned before considering the error err. Doing so correctly handles I/O errors that happen after reading some bytes and also both of the allowed EOF behaviors. 

That wasn't happening here; we were assuming the first call to Read would always read the full bytes, which isn't required by the io.Reader implementation. Partial reads don't return errors, which means that on a partial read here we were hitting this condition (as `!errors.Is(nil, io.EOF)` is true). 

The new code guarantees that we'll fully consume the reader and completely fill our buffer to its length or return an error if we received EOF before filling up the buffer.

CHANGELOG.md
```diff
@@ -19,8 +19,8 @@ Main (unreleased)
 
 ### Bugfixes
 
-- Fixes handling of an unexpeced EOF from S3 in `remote.s3` by erroring if an EOF
-  is encountered before Content-Length number of bytes have been read. (@spartan0x117)
+- Fixes `remote.s3` to always read the full Content-Length of the returned object, 
+  additionally erroring if an unexpected EOF is encountered before that many bytes are read. (@agent-squad)
```
rfratto: I think this should say what the fix is from a user's perspective, something like:

> Fix an issue in `remote.s3` where the exported content of an object would be an empty string if `remote.s3` failed to fully retrieve the file in a single read call.

WDYT?

PR #6112, opened 2024-01-11
Docs: Rename 'Setup' to 'Get started' and 'Start' to 'Run'

### LINKED ISSUE #5949: Create a clear space in docs hierarchy for the task-based docs

thampiotr:
#### PR Description

As part of improving docs structure, this PR moves the `flow/setup/` folder to `flow/get-started`. This is following the agreed new guidelines from `writing-docs.md`.

This also renames "Start" to "Run" as the former was clashing with "Get started".

Checked with `make docs` that there are no dead links.
Verified that the aliases for moved pages still works with `curl`.

image

This is the last step 🎉 
Fixes #5949


docs/sources/flow/get-started/start/_index.md
thampiotr, line 1: Pages in `start/` folder are new, created in https://github.com/grafana/agent/pull/6087, were not released yet anywhere other than `next` docs preview. So we don't need to add extra backwards compatibility aliases here.

rfratto:
Let me know if I'm nitpicking too hard here, but I find the combination of a folder called "get started" and a subfolder called "start" potentially confusing; people might think that "start" means "start here" (even if it's the second item in the list). 

Should we rename start to another verb to avoid this potential confusion? Or is this too much of a stretch?

(cc @clayton-cornell)

thampiotr:
@rfratto yeah, it could be better. Maybe "Run"? The instructions inside the `start/` folder actually talk about start/stop/restart. Let me know what you think about "Run" - I can make this changes now to reduce the back-n-forth.
Those are still new pages, so should be easier to rename.

rfratto:
run makes sense to me personally 👍

thampiotr:
@rfratto it's done now :) I think we're ready to merge this and I can address any follow-up comments from @clayton-cornell if needed.

clayton-cornell:
Run is good from my side... we might need to tweak the topic titles/headings here and there but it makes sense (and was something I was considering suggesting late yesterday as well... just hadn't got to it)

PR #6114, opened 2024-01-11
feat(loki.source.awsfirehose): add authentication support

sberz:
#### PR Description
Add an optional `access_key` attribute to enforce authentication for AWS Firehose requests. Unlike other exteral api requests AWS Firehose uses a different authentication header (`X-Amz-Firehose-Access-Key`), so using simple external proxy authentication (e.g ingress-nginx basic auth) is not possible here. Let's add a simple access key check, instead of requiring each user to build their own custom solution. 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


clayton-cornell:
@grafana/grafana-agent-maintainers this has code changes that needs a review.

component/loki/source/aws_firehose/component.go
```diff
@@ -159,7 +161,7 @@ func (c *Component) Update(args component.Arguments) error {
 
 	if err = c.server.MountAndRun(func(router *mux.Router) {
 		// re-create handler when server is re-computed
-		handler := internal.NewHandler(c, c.logger, c.handlerMetrics, c.rbs, newArgs.UseIncomingTimestamp)
+		handler := internal.NewHandler(c, c.logger, c.handlerMetrics, c.rbs, newArgs.UseIncomingTimestamp, string(newArgs.AccessKey))
```
mattdurham, line 164: I think we need to also set handleNeedsUpdate if the access key changes above.

mattdurham (changes requested):
Need to  update handler if key changes

component/loki/source/aws_firehose/component.go
sberz: Good catch! I added the check here: https://github.com/grafana/agent/pull/6114/files#diff-85fbd6d7bc146be7386366fda664cc67e006325f5e38a18f4fcc8af4ebb35c63R136-R138
 I also rebased the PR on the latest main version

mattdurham (approved):
Looks good

PR #6116, opened 2024-01-11
flow mixin: group "RW by component" panel by component

tristanburgess:
#### PR Description

Break up agent flow mixin Remote Write Active Series (by component) panel into two queries

1. one query for active series with non-empty component ID, summed in total for each component ID.
2. one query for any active series which happen to have an empty component ID (e.g. hybrid agent environments).

Maybe 2. is not needed, but the main point of this PR is that, if we have All instances selected, we should show the total series for each component across all instances. If we select a specific instance, specific active series for that specific instance will still be shown.

*Before*
Screenshot 2024-01-10 150917

*After*
Screenshot 2024-01-11 102619

#### PR Checklist

- [x] CHANGELOG.md updated


tristanburgess:
@tpaschalis updated as discussed to create two panels, one for component + instance, one for just component

image

tpaschalis (approved):
Thanks you for this Tristan!

PR #6117, opened 2024-01-11
example: remove k3d sample environment

rfratto:
The k3d example environment has been unused in a long time, and no longer works as the Tanka library it depended on has been removed.

The docker-compose environment supersedes the k3d environment for simple use cases.

I have left `example/docker-compose` untouched, but I will look into revamping it and making it work with Flow in a follow-up PR.


tpaschalis (approved):
📉

PR #6130, opened 2024-01-11
feat(operations/helm) add topology spread constraints

### LINKED ISSUE #5939: Add topologySpreadConstraints to helm chart
### Request

I would like to have a topologySpreadConstraints configurable setting added to the Grafana Agent helm chart. Settings must be configurable through the values.yaml.

### Use case

When deploying clustered Grafana Agents with a Deployment or StatefulSet, it would be useful to have topologySpreadConstraints to the pod specs so that pods get distributed evenly across multiple availability zones. 

This is especially important when using karpenter to provision nodes because pods might all get scheduled on a single instance, which isn't great for system resiliency if that one node goes away.

etiennep:
#### PR Description

Adds configuration to support topology spread constraints in the helm chart useful for deployment or statefulset controller types.

This allows pods to be distributed evenly across availability zones for example.

Fixes #5939

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/operations/helm/charts/grafana-agent/templates/controllers/_pod.yaml
+++ b/operations/helm/charts/grafana-agent/templates/controllers/_pod.yaml
@@ -58,6 +58,10 @@ spec:
   tolerations:
     {{- toYaml . | nindent 4 }}
   {{- end }}
+  {{- with .Values.controller.topologySpreadConstraints }}
+  topologySpreadConstraints:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
   volumes:
     - name: config
       configMap:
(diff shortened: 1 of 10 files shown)


operations/helm/charts/grafana-agent/templates/controllers/_pod.yaml
```diff
@@ -57,6 +57,10 @@ spec:
   tolerations:
     {{- toYaml . | nindent 4 }}
   {{- end }}
+  {{- with .Values.controller.topologySpreadConstraints }}
+  topologySpreadConstraints:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
```
tpaschalis, line 63: Should we only include this field if the value is set? (eg. similar to what we do with other fields like priorityClassName a few lines above)

Otherwise it looks like pods are failing to start as they're waiting for the empty conditions to be met.

etiennep: I looked at the tests after re-generating, and it looks like that whole section gets omitted unless the array is set to a non-empty value. In any case, I've set the condition to `ScheduleAnyway` in case there's some test I didn't know was being run during CI. That should get around any scheduling failure.

tpaschalis (approved):
Thank you for the contribution Etienne!

PR #6137, opened 2024-01-14
chore: migrate away from `aws-sdk-go` v1

hainenber:
#### PR Description

`github.com/aws-sdk-go` has been announced to be in maintenance mode July 31, 2024. It's better to future-proof by moving to `aws-sdk-go-v2` and also 1 less dependency. 

#### PR Checklist

- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -180,6 +180,8 @@ v0.40.0 (2024-02-27)
 - `grafana-agent` and `grafana-agent-flow` fallback to default X.509 trusted root certificates
   when the `GODEBUG=x509usefallbackroots=1` environment variable is set. (@hainenber)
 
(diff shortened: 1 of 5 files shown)


tpaschalis:
Hey there, sorry for the noise but could you please rebase with main and fix the conflicts?

I think this is on a good path; we'll have to look into their [migration guide](https://aws.github.io/aws-sdk-go-v2/docs/migrating/) to make sure we haven't missed anything else, but the fact that everything compiles is encouraging.

hainenber:
Thank you so much Paschalis for your review effort. It's not noise to me at all, in fact I like these GH notifications :D

I've rebased and resolved the conflict. PTAL again. Thanks!

tpaschalis (approved):
LGTM, thanks! 🙌

PR #6154, opened 2024-01-15
fix(otelcol): fix Prometheus exporter not collecting metrics after reload

### LINKED ISSUE #6140: "An error has occurred while serving metrics" after a reload
### What's wrong?

After I deployed a change to the configuration of my agent to add an extra trace relabeling, my agents reloaded fine and since the metrics fail to collect due to an error as shown below:

```
An error has occurred while serving metrics:
16 error(s) occurred:   
collected metric "traces_service_graph_unpaired_spans_total" { label:{name:"client"  value:""}  label:{name:"server"  value:"thanos-sidecar"}  label:{name:"traces_config"  value:"default"}  counter:{value:63862}} was collected before with the same name and label values
```
The above repeats multiple times for each metric
Restarting the pod fixes the issue
I can reproduce the issue without changing the config by simply trigerring a reload using the `/-/reload` endpoint

### Steps to reproduce

- start Grafana Agent (from official helm chart, with custom config file using static mode)
- collect metrics
- make a change to the configmap (I added a relabel_configs entry)
- check the metrics endpoint

### System information

_No response_

### Software version

v0.39.0

### Configuration

_No response_

### Logs

_No response_

hainenber:
#### PR Description

Fixes OTEL metrics not getting collected after the Agent is reloaded. The fix has already arrived in [`go.opentelemetry.io/otel/sdk/metric@1.20.0`](https://github.com/open-telemetry/opentelemetry-go/blob/main/CHANGELOG.md#fixed-2)

Fixes #6140 

#### PR Checklist

- [x] CHANGELOG.md updated


tpaschalis (approved):
Thank you once more @hainenber 🙌 

For posterity, I think the upstream fix was https://github.com/open-telemetry/opentelemetry-go/pull/4648 and the original report was https://github.com/open-telemetry/opentelemetry-go/issues/4032

PR #6178, opened 2024-01-17
Draft: Skeleton / plan for future k8s docs

thampiotr:
This draft PR outlines what pages with roughly what content we could add as one of the ways to improve users' experience when working with k8s.

### DIFF
--- a/docs/sources/flow/tasks/configure/configure-kubernetes.md
+++ b/docs/sources/flow/tasks/configure/configure-kubernetes.md
@@ -21,6 +21,12 @@ weight: 200
 
 To configure {{< param "PRODUCT_NAME" >}} on Kubernetes, perform the following steps:
 
+TODO(thampiotr): The values.yaml is quite dense and may not be suitable for new users. Instead, we should
+turn this into more step-by-step instruction to changing Grafana Agent configuration when helm chart is used. This will 
+put it more in-line with the other docs inside this "configure" section.
+For changing the Helm chart config we can refer to our other task-based docs (to be added in the future) that cover this and to the Helm 
+chart configuration docs for full details.
+
 1. Download a local copy of [values.yaml][] for the Helm chart.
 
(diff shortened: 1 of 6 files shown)


docs/sources/flow/tasks/configure/configure-kubernetes.md
```diff
@@ -21,6 +21,12 @@ weight: 200
 
 To configure {{< param "PRODUCT_NAME" >}} on Kubernetes, perform the following steps:
 
+TODO(thampiotr): The values.yaml is quite dense and may not be suitable for new users. Instead, we should
```
mattdurham, line 24: Is this description more of a tutorial than task? It also might be to broad, when saying configure kubernetes, is that for metrics, logs, traces ,pyroscope , mimir rules? Each of the them should be a different task since they require substantially different approaches.

mattdurham: Basically I think the below `collect-` are good and this one is a tutorial

thampiotr: Yeah, things are getting more complicated when I realise that to deploy as a useful DaemonSet for logs, one needs to install the helm chart together with the correct volume mounts etc. 

Having it all on a single `collect-` page may be too much though, I was wondering if I can split it into two pages: one to install a DaemonSet with Helm, and second one to start using the kubernetes module from agent-modules repo for logs. 

I'll maybe start with a single large page and we'll split this later if needed. Will be easier to decide once we have a full doc.

thampiotr: Thanks @mattdurham ! I've updated this now to reflect this slightly different, but I think better approach.

thampiotr:
Closed this one as superseded by this: https://github.com/grafana/agent/pull/6187

PR #6180, opened 2024-01-17
Add OpenTelemetry dashboards to the Flow mixin

ptodev:
The mixin for Static mode already has a few dashboards. This PR adds the same ones + a few more to the Flow mixin.

For now the dashboards are mostly tracing-specific. Over time we will add logs and metrics support. We will also add dashboards for more components.

### DIFF
--- a/operations/agent-flow-mixin/dashboards/opentelemetry.libsonnet
+++ b/operations/agent-flow-mixin/dashboards/opentelemetry.libsonnet
@@ -112,7 +112,7 @@ local stackedPanelMixin = {
         panel.withQueries([
           panel.newQuery(
             expr=|||
-             sum(processor_batch_metadata_cardinality_ratio{cluster="$cluster", namespace="$namespace", instance=~"$instance"}) or vector(0)
+             processor_batch_metadata_cardinality_ratio{cluster="$cluster", namespace="$namespace", instance=~"$instance"}
             |||,
             legendFormat='{{ pod }}',
           ),
--- a/operations/agent-flow-mixin/dashboards/opentelemetry.libsonnet
+++ b/operations/agent-flow-mixin/dashboards/opentelemetry.libsonnet
@@ -1,0 +1,174 @@
+local dashboard = import './utils/dashboard.jsonnet';
+local panel = import './utils/panel.jsonnet';
+local filename = 'agent-flow-opentelemetry.json';
+
+local stackedPanelMixin = {
+  fieldConfig+: {
+    defaults+: {
+      custom+: {
+        fillOpacity: 20,
+        gradientMode: 'hue',
+        stacking: { mode: 'normal' },
+      },
+    },
+  },
+};
+
+{
+  [filename]:
+    dashboard.new(name='Grafana Agent Flow / OpenTelemetry') +
+    dashboard.withDashboardsLink() +
+    dashboard.withUID(std.md5(filename)) +
+    dashboard.withTemplateVariablesMixin([
+      dashboard.newTemplateVariable('cluster', |||
+        label_values(agent_component_controller_running_components, cluster)
+      |||),
+      dashboard.newTemplateVariable('namespace', |||
+        label_values(agent_component_controller_running_components{cluster="$cluster"}, namespace)
+      |||),
+      dashboard.newMultiTemplateVariable('instance', |||
+        label_values(agent_component_controller_running_components{cluster="$cluster", namespace="$namespace"}, instance)
+      |||),
+    ]) +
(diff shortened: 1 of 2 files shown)


operations/agent-flow-mixin/dashboards/opentelemetry.libsonnet
```diff
@@ -0,0 +1,174 @@
...
+      (
+        panel.newHeatmap('RPC server duration (traces)') +
+        panel.withUnit('milliseconds') +
+        panel.withDescription(|||
+          The duration of inbound RPCs.
+        |||) +
+        panel.withPosition({ x: 16, y: 0, w: 8, h: 10 }) +
+        panel.withQueries([
+          panel.newQuery(
+            expr='sum by (le) (increase(rpc_server_duration_milliseconds_bucket{cluster="$cluster", namespace="$namespace", instance=~"$instance", rpc_service="opentelemetry.proto.collector.trace.v1.TraceService"}[$__rate_interval]))',
```
ptodev, line 80: I'm not very experience with Prometheus - can I ask someone to please double check that this PromQL makes sense?

Also, for this heatmap sometimes there are negative values in the most recent buckets:
Screenshot 2024-01-17 at 16 28 53

The negative values disappear once the bucket is older. I don't know why this happens?

operations/agent-flow-mixin/dashboards/opentelemetry.libsonnet
```diff
@@ -0,0 +1,174 @@
...
+      ),
+      (
+        panel.newHeatmap('Number of units in the batch') +
+        panel.withDescription(|||
+          Number of units in the batch
+        |||) +
+        panel.withPosition({ x: 0, y: 10, w: 8, h: 10 }) +
+        panel.withQueries([
+          panel.newQuery(
+            expr='sum by (le) (increase(processor_batch_batch_send_size_ratio_bucket{cluster="$cluster", namespace="$namespace", instance=~"$instance"}[$__rate_interval]))',
```
ptodev, line 100: Here we also get negative values:
Screenshot 2024-01-17 at 16 29 09

operations/agent-flow-mixin/dashboards/opentelemetry.libsonnet
ptodev: I am not very confident in the PromQL query because I copied it from another place in this mixin. I can understand the non-histogram queries in my dashboards, but this one about histograms is a bit over my head.

operations/agent-flow-mixin/dashboards/opentelemetry.libsonnet
```diff
@@ -0,0 +1,174 @@
...
+      (
+        panel.new(title='Distinct metadata values', type='timeseries') +
+        panel.withDescription(|||
+          Number of distinct metadata value combinations being processed
+        |||) +
+        panel.withPosition({ x: 8, y: 10, w: 8, h: 10 }) +
+        panel.withQueries([
+          panel.newQuery(
+            expr=|||
+             sum(processor_batch_metadata_cardinality_ratio{cluster="$cluster", namespace="$namespace", instance=~"$instance"}) or vector(0)
```
tpaschalis, line 115: As discussed offline: this probably makes more sense per-pod, similar to other queries instead of an aggregated metric.

operations/agent-flow-mixin/dashboards/opentelemetry.libsonnet
tpaschalis: As discussed offline: we're still a little unclear on what the actual underlying metric is about (`ratio`? of what?). Does the 220 upper-limit value here make sense compared to the rate of data we're seeing in the actual environment>

We can take a look, but maybe we could remove this panel for now and put it back on after we've got a better understanding of what we're looking at here.

jcreixell:
side note: given that for the OTEL use case (mostly stateless, supposedly fair load balancing) individual agent failures are probably less common than when using a WAL, does the per agent partition make sense? (instead of having aggregated metrics across all instances and a dropdown to select individual ones). For example, aggregated accepted/refused spans, etc.

jcreixell (approved):

operations/agent-flow-mixin/dashboards/opentelemetry.libsonnet
ptodev: I think those weird buckets are only visible in my locally running Grafana and not in the actual environment (even for the same data in Mimir), so I'll assume that the dashboard is ok for now and will merge it as is. We can improve it over time.

tpaschalis (approved):

PR #6182, opened 2024-01-18
Docs to clarify what various components are used for

ptodev:
I'd like the new docs to be the place we refer people when someone asks whether the Agent can do something, or whether it can be used in a certain situation with a certain telemetry. We currently don't have a doc which lists all of this info, and we need to rephrase it manually in conversations.

This PR is a draft because I am not sure of the best location for this, and I'm not sure how it is best represented.

### DIFF
--- a/docs/sources/flow/reference/components/_index.md
+++ b/docs/sources/flow/reference/components/_index.md
@@ -14,6 +14,96 @@ weight: 300
 
 This section contains reference documentation for all recognized [components][].
 
+## Which component do I need?
+
+### Receiving telemetry
+
+##### Logs
+
+* In the Loki format: `loki.source` components such as `loki.source.file`.
+  * If you need to convert from the Loki format to OTLP, you can use `otelcol.receiver.loki`.
+* In the OTLP format: some `otelcol.receiver` components, such as `otelcol.receiver.otlp`.
+
+##### Metrics
+
+`prometheus.scrape` scrapes HTTP endpoints using the Prometheus exposition format.
+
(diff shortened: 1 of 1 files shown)


docs/sources/flow/reference/components/_index.md
ptodev, line 1: I haven't mentioned some components yet: 
* `discovery.*`
* `remote.*`
* [pyroscope.java](https://grafana.com/docs/agent/next/flow/reference/components/pyroscope.java/)
* [pyroscope.ebpf](https://grafana.com/docs/agent/next/flow/reference/components/pyroscope.ebpf/)
* [faro.receiver](https://grafana.com/docs/agent/next/flow/reference/components/faro.receiver/)

I'm worried that this page is starting to get really long. Maybe we could make it a totally separate page?

ptodev:
I think the ToC looks nice with the proposed titles:
Screenshot 2024-01-18 at 11 52 09

docs/sources/flow/reference/components/_index.md
clayton-cornell: I would start with the single page in draft and split only if needed. As it is, each section is fairly small

clayton-cornell: But... as I read it we are getting really deep with the heading levels. I would suggest we break this off the index and just have it as its own standalone page. That might help reduce the empty heading overload that the current draft has (we should avoid having a heading immediately followed by another heading with no text in between the headings).

ptodev:
I'll close this because Agent is close to EOL now...

PR #6191, opened 2024-01-18
pyroscope: change demangle default value from `none` to `full`

petethepig:
See this thread for more info: https://raintank-corp.slack.com/archives/C048VSWPZ4P/p1705604141660129

cc @korniltsev — WDYT?


korniltsev (approved):
I've made it none by default as a premature optimization.

I dont like changing it to full as current implementation will process each symbol even when we dont need any processing, even if it is golang or python.

For example grafana workloads doe not need this default.

korniltsev:
You probably need to update changlog

korniltsev:
Other things to consider in the future: 
- demangling should be done only for C++/Rust. currently it is done for all
- It can happen on the storage side at distributor or maybe query

korniltsev:
You probably need to update docs with the new default

PR #6202, opened 2024-01-19
prototype from grafana.cloud components DO NOT MERGE

erikbaranowski:
THESE ARE PROTOTYPES DO NOT MERGE

DOES NOT INCLUDE:
- well written code
- components we definitely want to support

Notes:
- Each component could also be implemented at the stack level instead of org level
- grafana.cloud.instances should probably export the token too so you don't need to set it multiple places
- grafana.cloud.receivers
  - Maybe should group the exports by receiver type rather than stack
  - Only implemented for prometheus to start
  - Component arguments could include all promtheus.remote_write args if we wanted

Usage:

Set up an access policy with a token like this
image

- grafana.cloud.instances

```river
grafana.cloud.instances "default" {
	token = "TOKEN"
	org   = "ORG"
}

	endpoint {
		url = grafana.cloud.instances.default.stacks["STACK"]["metrics_url"]

		basic_auth {
			username = grafana.cloud.instances.default.stacks["STACK"]["metrics_username"]
			password = "TOKEN"
		}
	}
}

prometheus.scrape "default" {
	targets         = [{"__address__" = "0.0.0.0:12345"}]
	forward_to      = [prometheus.remote_write.default.receiver]
	scrape_interval = "10s"
}
```

- grafana.cloud.receivers

```river
grafana.cloud.receivers "default" {
	token = "TOKEN"
	org   = "ORG"
}

prometheus.scrape "default" {
	targets         = [{"__address__" = "0.0.0.0:12345"}]
	forward_to      = [
		grafana.cloud.receivers.default.stacks["STACK1"].prometheus_receiver,
		grafana.cloud.receivers.default.stacks["STACK2"].prometheus_receiver,
	]
	scrape_interval = "10s"
}
```

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/component/all/all.go
+++ b/component/all/all.go
@@ -33,6 +33,8 @@ import (
 	_ "github.com/grafana/agent/component/discovery/triton"                         // Import discovery.triton
 	_ "github.com/grafana/agent/component/discovery/uyuni"                          // Import discovery.uyuni
 	_ "github.com/grafana/agent/component/faro/receiver"                            // Import faro.receiver
+	_ "github.com/grafana/agent/component/grafana/cloud/instances"                  // Import grafana.cloud.instances
+	_ "github.com/grafana/agent/component/grafana/cloud/receivers"                  // Import grafana.cloud.receivers
 	_ "github.com/grafana/agent/component/local/file"                               // Import local.file
 	_ "github.com/grafana/agent/component/local/file_match"                         // Import local.file_match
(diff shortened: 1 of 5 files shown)


erikbaranowski:
No plans to move forward with this at this time.

PR #6212, opened 2024-01-22
Fix non-unique ID used in worker pool

### LINKED ISSUE #6199: Nested modules sometimes fail to propagate exports
### What's wrong?

This branch contains the test that sometimes triggers the bug and a hacky fix: https://github.com/grafana/agent/compare/experiment-modules-bug?expand=1

More context from @rfratto after a quick investigation: the worker pool is shared between all controllers, and is used for evaluating updated components. items are queued into the worker pool using [SubmitWithKey](https://github.com/grafana/agent/blob/afd592d96c772c09cea3ef2f5518be70ef19dd9c/pkg/flow/internal/worker/worker_pool.go#L13-L24).
note this lne:
Adding a job with a key that is already queued is a no-op (even if the submitted function is different).
the bug is that the key is the local ID of a component, not the global ID. because you have multiple components with the local ID of [testcomponents.passthrough.pt](http://testcomponents.passthrough.pt/), the worker queue is getting confused and is dropping events for one/both of them.
it's essentially a race condition which depends on whether the worker queue processes events from both components quickly enough. if it doesn't, events for one of them get dropped.

Without the fix, you can avoid the problem by having a different label on every component. **It also seems that you need a different label on the export blocks. We don't know yet why this is the case for the exports block, this part should be investigated as it might reveal another bug.**

### Steps to reproduce

To trigger the bug you can modify the test of the branch above and put the same label on the same components (for example put the label "pt" on both "testcomponents.passthrough". If you run the test without the hacky fix several time you will see that it sometimes either fail or it pauses for about a second before succeeding.

### System information

_No response_

### Software version

v0.39.0

### Configuration

_No response_

### Logs

```text
Here is paused for one second:
ts=2024-01-19T17:34:14.091856Z level=info msg="finished node evaluation" controller_id=module.file.default/module.file.default node_id=export.output duration=5.708µs
ts=2024-01-19T17:34:15.231655Z level=info msg="starting complete graph evaluation" controller_id=module.file.default trace_id=5b53b50575f1983e1f94ecbcd863632f
```

thampiotr:
#### PR Description

* Add a test in the first commit
* Add a fix in the second commit

Fixes #6199

#### Notes to the Reviewer

Wrote a slightly different, smaller test than the draft branch mentioned on the issue.

Before the fix, running `go test ./pkg/flow/ -count 10 -test.run "TestUpdates_TwoModules*"` would usually catch the bug:
```
--- FAIL: TestUpdates_TwoModules_SameCompNames (3.01s)
    module_caching_test.go:200:
        	Error Trace:	/Users/piotr/workspace/agent/pkg/flow/module_caching_test.go:200
        	Error:      	Condition never satisfied
        	Test:       	TestUpdates_TwoModules_SameCompNames
FAIL
FAIL	github.com/grafana/agent/pkg/flow	19.395s
FAIL
```

After the fix, it doesn't fail.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,9 @@ Main (unreleased)
 
 - Fix issue causing duplicate logs when a docker target is restarted. (@captncraig)
 
+- Fix an issue where blocks having the same type and the same label across
+  modules could result in missed updates. (@thampiotr)
+
 ### Other changes
 
 - Removed support for Windows 2012 in line with Microsoft end of life. (@mattdurham)
--- a/pkg/flow/internal/controller/loader.go
+++ b/pkg/flow/internal/controller/loader.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"path"
 	"sync"
 	"time"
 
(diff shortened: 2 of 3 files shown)


tpaschalis (approved):
LGTM.

PR #6213, opened 2024-01-22
feat(operations/helm) add support for extraObjects

timtalbot:
#### PR Description
Adds support for an `extraObjects` field in chart values to add additional manifests with chart installation.

#### Notes to the Reviewer
Following the model used by many charts including other Grafana ones to deploy extra objects with the chart. Thanks for your time!
#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/operations/helm/charts/grafana-agent/CHANGELOG.md
+++ b/operations/helm/charts/grafana-agent/CHANGELOG.md
@@ -15,6 +15,8 @@ Unreleased
 
 ### Enhancements
 
+- Add support for `extraObjects` to define additional manifests. (@timtalbot)
+
 - Add `kubectl.kubernetes.io/default-container: grafana-agent` annotation to allow various tools to choose `grafana-agent` container as default target (@aerfio)
 
 - Add support for topology spread constraints in helm chart. (@etiennep)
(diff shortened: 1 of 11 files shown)


operations/helm/charts/grafana-agent/CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ Unreleased
 
 ### Enhancements
 
+- Add support for `extraObjects` to define additional manifests. (@timtalbot)
```
tpaschalis, line 18: Let's move the entry to a new Enhancements header under Unreleased instead.

timtalbot: Oops! Good catch.

tpaschalis (dismissed):
Could you describe a couple of use cases where this could be useful?

This has come up in the past, but I'd like for us to pinpoint some real-world problems this is solving before introducing it simply because other charts do the same.

IMHO, this might be a footgun to users. How do we debug issues arising from people reusing values.yaml files they found online, or how do we deal with security concerns around having random objects reuse the agent's RBAC resources?

tpaschalis:
Hey there, apologies for the belated response. For the time being, I'd like to hold off on this and reconsider if some other use case with no workaround.

QuentinBisson:
@tpaschalis The simplest use case I have is this:
How can I deploy a secret with this chart for let's say, connect to a secured endpoint like let's say grafana cloud?
Without creating the secret outside of the chart, how would you create this secret? Without this or another mechanism, then the chart is not self-sufficient

PR #6222, opened 2024-01-23
Rename prometheus.remote_write dashboard & add scrape metrics

thampiotr:
#### PR Description

This adds visibility of scrape performance to our main Prometheus dashboard (currently called `prometheus.remote_write`)

image

We've agreed to have one dashboard per component namespace, with metrics of each component in separate collapsible rows on the dashboard. So also doing the following: 
* Renaming `prometheus.remote_write` dashboard to "Prometheus Components"
* Adding rows for each component
* Putting prometheus.scrape on top as discussed, to reflect better the pipeline order.

(internal issue)

#### Notes to the Reviewer

This is a follow-up from an incident where scrape performance was badly affected, but it was hard to notice this using our dashboards. 

During the incident, the query we're adding here would clearly highlight the issue:
image

#### PR Checklist

- [x] CHANGELOG.md updated


operations/agent-flow-mixin/dashboards/prometheus.remote_write.libsonnet
```diff
@@ -347,6 +347,70 @@ local stackedPanelMixin = {
         ])
       ),
 
+      // Scrape success rate
```
rfratto, line 350: The reason I called this dashboard `prometheus.remote_write` was to have component-specific dashboards; this came from the understanding that not all users will use the same set of components in a pipeline, so combining metrics from two components in a single dashboard wouldn't necessarily be useful for all users.

In this particular case, not everyone using a `prometheus.remote_write` component also has a `prometheus.scrape` component in the same pipeline (i.e., users using OpenTelemetry components but exporting metrics to Prometheus). 

Can we make a new dashboard for `prometheus.scrape`?

thampiotr: I would prefer we instead rename this dashboard to "Prometheus Metrics" and I could make separate collapsible rows for different components. 

The reason is that we have a lot of components and it'd be quite frustrating to have to jump between tabs to correlate the issues in the pipeline that happen in different components. Poor user experience overall. When assessing system health, a high level dashboard is a better starting point for identifying where the issue is and I believe we should provide this in the first instance. 

For the OTEL-based pipelines, we could also have a dashboard with rows for different components or telemetry signals.

rfratto: Can this be discussed with the rest of the maintainers offline so we can get consensus on an approach for how to handle and expand these dashboards?

thampiotr: I've addressed the feedback as discussed.

rfratto: For public record, the offline consensus was to create dashboards aligned with component namespaces, with each row being for a different component, and the order of the rows roughly reflecting the order on the write path for that namespace.

rfratto (approved):

PR #6223, opened 2024-01-23
Upgrade river dependency to main

### LINKED ISSUE #3244: UI API fails when viewing component which exports a map block

wildum:
#### PR Description

This upgrade bring this one [change](https://github.com/grafana/river/pull/41) which is needed for the UI of the new modules .

Closes #3244.


rfratto (approved):
Thanks!

PR #6229, opened 2024-01-24
New interface to retrieve info about running components inside of nodes.

wildum:
Before ComponentNodes were the only nodes managing a component. The [new modules](https://github.com/grafana/agent/issues/5547) will introduce two new nodes that will have running components inside them:
- the CustomComponentNode will be running a custom component defined by a declare.
- the ImportConfigNode will be running a component to retrieve a module from a given location (the component depends on the source, if the module is stored in a file then the node will run a local_file component).

The component provider API is used to retrieve information about running components. It's implemented by Flow to access the components within the graph. Using the ComponentInfo interface instead of the ComponentNode struct will allow Flow to retrieve information about components running inside any nodes.

Flow services using the API will properly retrieve info about all the components involved in the pipeline. That means, for example in our case, that the UI will be able to show the ImportConfigNode and the CustomComponentNode.

API change:
The Registration field has been replaced by simply the name of the block as a string for the following reasons:
- Registration is used to retrieve information about built-in components. It's not a concept that CustomComponents have.
- Registration provides a "Build" function that is used to build the component. The API is used to retrieve information about running components so this is not relevant information to expose.
- Registration provides Arguments and Exports that are already available in the Info struct. This information is redundant because these two fields are initialized by the Registration when the node is created.
The only field used (and the only relevant one according to the reasons above) is the name of the block. That's why this PR suggests replacing it with BlockName.

The MarshalJSON has the warning "The format of the representation is not stable and is subject to change." which I believe makes the above change ok.


pkg/flow/internal/controller/component_node.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"context"
+
+	"github.com/grafana/agent/component"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// ComponentInfo is an interface that encapsulates methods for managing and retrieving detailed information about a component within a BlockNode.
+type ComponentInfo interface {
```
rfratto, line 11: Unfortunately I also have some issues with this interface that I think should be resolved:

* I think the name and description doesn't really represent my mental modal of how nodes in the DAG work. Rather than a component being "within" a BlockNode, I see a node presenting itself _as_ a component. 

* I think this interface is very broad, and probably inclusive of things that don't make sense for non-builtin components: at first glance `DebugInfo()`, `Component()`, `ModuleIDs()`, and `Registry()` seem to be specific to builtin components. 

* `BlockName()` throws me off a little. I know it's literally what is being returned, but it can be seen as a leaky abstraction for the configuration language and where the component name may come from. I would rather this be called `ComponentName()`, which more representative of the data being returned and is named agnostically to how that data gets populated in the node. 

I'd like to see these changes:

1. Rename the `ComponentNode` struct to `BuiltinComponentNode`, and allow this interface to be `ComponentNode`. 

2. Minimize the interface as much as possible. Only methods which make sense for _all_ possible implementations of a component should remain. 

3. Rename `BlockName()` to `ComponentName()`. 

WDYT?

rfratto (changes requested):

pkg/flow/internal/controller/component_node.go
wildum: In the current implementation I already have an interface for ComponentNode. I wanted to bring it with the rename of the ComponentNode to BuiltinComponentNode via a different PR.
- BuiltinComponentNode / CustomComponentNode implement Component Node (used by the Loader to handle both as ComponentNode)
- BuiltinComponentNode / CustomComponentNode / ImportConfigNode implement Component Info (used by the component provider to retrieve information about running components)

How should we handle the naming?

About removing the functions I think that makes sense, I can use type switch in the getComponentDetail to pass the builtin component info to build the component.Info

wildum: But do we want the Import node to appear in the UI? It was a bit of a stretch to pass it as Component because it contains a component but at least it would reuse the current logic to pass the data through the API

rfratto: > But do we want the Import node to appear in the UI? It was a bit of a stretch to pass it as Component because it contains a component but at least it would reuse the current logic to pass the data through the API


For now I'm OK with the import node _not_ appearing in the UI. The UI needs improvements in general (such as showing configuration blocks), so we'll find the best way to expose other non-component nodes when it's time to expand the UI to cover more than components.

wildum: I made the changes and I also put the block node interface inside of the runnable interface to follow the schema that you put above. I think that this hierarchy makes sense and makes it easier to follow

pkg/flow/flow_components.go
```diff
@@ -119,11 +113,20 @@ func (f *Flow) getComponentDetail(cn *controller.ComponentNode, graph *dag.Graph
...
 		ReferencedBy: referencedBy,
 
-		Registration: cn.Registration(),
-		Health:       health,
+		ComponentName: cn.ComponentName(),
+		Health:        health,
 
 		Arguments: arguments,
 		Exports:   exports,
 		DebugInfo: debugInfo,
```
rfratto, line 121: `debugInfo` is never set now; can you remove this and the declaration above?

pkg/flow/internal/controller/component_node.go
```diff
@@ -0,0 +1,33 @@
...
+	"context"
+
+	"github.com/grafana/agent/component"
+)
+
+// ComponentNode is a generic representation of a Flow component.
+type ComponentNode interface {
+	RunnableNode
+
+	// CurrentHealth returns the current health of the node.
```
rfratto, line 13:
```suggestion
	// CurrentHealth returns the current health of the component.
```

pkg/flow/internal/controller/component_node.go
```diff
@@ -0,0 +1,33 @@
...
+	CurrentHealth() component.Health
+
+	// Arguments returns the current arguments of the managed component.
+	Arguments() component.Arguments
+
+	// Exports returns the current set of exports from the managed component.
+	Exports() component.Exports
+
+	// Label returns the label for the block or "" if none was specified.
```
rfratto, line 22: All components have labels now, right? 

```suggestion
	// Label returns the component label.
```

pkg/flow/internal/controller/component_node.go
```diff
@@ -0,0 +1,33 @@
...
+	Label() string
+
+	// ComponentName returns the name of the block.
```
rfratto, line 25:
```suggestion
	// ComponentName returns the name of the component.
```

pkg/flow/internal/controller/component_node.go
```diff
@@ -0,0 +1,33 @@
...
+	ComponentName() string
+
+	// Run runs the managed component in the calling goroutine until ctx is canceled.
+	Run(ctx context.Context) error
+
```
rfratto, line 30: This is defined in RunnableNode, it doesn't need to be redefined here:

```suggestion
```

pkg/flow/internal/controller/loader.go
```diff
@@ -452,7 +452,7 @@ func (l *Loader) populateComponentNodes(g *dag.Graph, componentBlocks []*ast.Blo
 		// Check the graph from the previous call to Load to see we can copy an
 		// existing instance of ComponentNode.
 		if exist := l.graph.GetByID(id); exist != nil {
-			c = exist.(*ComponentNode)
+			c = exist.(*BuiltinComponentNode)
```
rfratto, line 455: Let's be a little protective and only cast this to the minimum interface we need:

```suggestion
			c = exist.(ComponentNode)
```

pkg/flow/internal/controller/loader.go
```diff
@@ -435,7 +435,7 @@ func (l *Loader) populateComponentNodes(g *dag.Graph, componentBlocks []*ast.Blo
 		blockMap = make(map[string]*ast.BlockStmt, len(componentBlocks))
 	)
 	for _, block := range componentBlocks {
-		var c *ComponentNode
+		var c *BuiltinComponentNode
```
rfratto, line 438: It seems likely that we'll want this to be the interface instead of the concrete type:

```suggestion
		var c ComponentNode
```

pkg/flow/internal/controller/metrics.go
```diff
@@ -110,9 +110,11 @@ func (cc *controllerCollector) Collect(ch chan<- prometheus.Metric) {
 	componentsByHealth := make(map[string]int)
 
 	for _, component := range cc.l.Components() {
-		health := component.CurrentHealth().Health.String()
-		componentsByHealth[health]++
-		component.registry.Collect(ch)
+		if builtinComponent, ok := component.(*BuiltinComponentNode); ok {
+			health := builtinComponent.CurrentHealth().Health.String()
+			componentsByHealth[health]++
```
rfratto, line 115: The health applies to all ComponentNodes, so I think this needs to be done outside of the type check so we collect health information on non-builtin components.

Only the call to `builtinComponent.registry.Collect(ch)` is unique to builtin components.

pkg/flow/internal/controller/node_builtin_component.go
```diff
@@ -73,12 +73,12 @@ type ComponentGlobals struct {
 	GetServiceData      func(name string) (interface{}, error) // Get data for a service.
 }
 
-// ComponentNode is a controller node which manages a user-defined component.
+// BuiltinComponentNode is a controller node which manages a user-defined component.
```
rfratto, line 76: "user-defined component" is ambiguous now that we're going to introduce the concepts of custom components. 

```suggestion
// BuiltinComponentNode is a controller node which manages a builtin component.
```

rfratto (commented):
LGTM after this final set of comments :)

rfratto (approved):
Let's go!

PR #6236, opened 2024-01-25
Tweak clustering alerts

thampiotr:
#### PR Description

- Removed `ClusterLamportClockDrift` and `ClusterLamportClockStuck` alerts from
  Grafana Agent Mixin to focus on alerting on symptoms.

- Increased clustering alert periods to 10 minutes to improve the
  signal-to-noise ratio in Grafana Agent Mixin.

#### Notes to the Reviewer

Based on internal team discussion.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/operations/agent-flow-mixin/alerts/clustering.libsonnet
+++ b/operations/agent-flow-mixin/alerts/clustering.libsonnet
@@ -8,7 +8,7 @@ alert.newGroup(
       'ClusterNotConverging',
       'stddev by (cluster, namespace) (sum without (state) (cluster_node_peers)) != 0',
       'Cluster is not converging.',
-      '5m',
+      '10m',
     ),
 
     // Cluster has entered a split brain state.
@@ -23,15 +23,7 @@ alert.newGroup(
         count by (cluster, namespace) (cluster_node_info)
       |||,
       'Cluster nodes have entered a split brain state.',
-      '5m',
-    ),
-
-    // Standard Deviation of Lamport clock time between nodes is too high.
-    alert.newRule(
-      'ClusterLamportClockDrift',
-      'stddev by (cluster, namespace) (cluster_node_lamport_time) > 4 * sqrt(count by (cluster, namespace) (cluster_node_info))',
-      "Cluster nodes' lamport clocks are not converging.",
-      '5m'
+      '10m',
     ),
 
     // Nodes health score is not zero.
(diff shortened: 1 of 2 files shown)


operations/agent-flow-mixin/alerts/clustering.libsonnet
```diff
@@ -41,22 +33,7 @@ alert.newGroup(
...
-    // with no external peers will not increase their lamport time because
-    // there is no cluster networking traffic.
-    alert.newRule(
-      'ClusterLamportClockStuck',
-      |||
-        sum by (cluster, namespace, instance) (rate(cluster_node_lamport_time[2m])) == 0
-        and on (cluster, namespace, instance) (cluster_node_peers > 1)
-      |||,
-      "Cluster nodes's lamport clocks is not progressing.",
-      '5m',
```
tpaschalis, line 59: I think we can keep this alert; this does not depend on peer timing but is purely for a specific node. WDYT?

operations/agent-flow-mixin/alerts/clustering.libsonnet
```diff
@@ -84,10 +61,7 @@ alert.newGroup(
         ) > 1
       |||,
       'Cluster nodes are not using the same configuration file.',
-      '5m',
+      '10m',
```
tpaschalis, line 64: I believe it doesn't hurt to have the 5m interval; deploying an erroneous configuration file or being stuck in terminating state doesn't sound something that can be flappy.

tpaschalis (approved):
LGTM; left a couple of thoughts but leaving it up to you as to what you believe is best.

operations/agent-flow-mixin/alerts/clustering.libsonnet
thampiotr: It would still be a cause and not a symptom. If the clock is stuck, we will eventually get a cluster transition and the other alert will fire. 

If we try to predict all possible causes, we'll likely fail and we'll also have a ton of alerts and they will misfire frequently, so I'm feeling quite strongly against alerts on causes.

tpaschalis: Ok, let's move ahead with this and see how it plays out.

operations/agent-flow-mixin/alerts/clustering.libsonnet
thampiotr: Happy to revert for this one alert I guess, the config map reloader not kicking in for 5min could be a permanent issue, yeah.

PR #6242, opened 2024-01-25
Fix heading remove variable

clayton-cornell:
#### PR Description

This PR removes the `{{< param "PRODUCT_NAME" >}}` Hugo shortcode from headings. If we use this shortcode in the heading text Hugo rewrites the anchor to say `#hahahugoshortcodes3hbhb` wherever the shortcode is used.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/sources/flow/_index.md
+++ b/docs/sources/flow/_index.md
@@ -14,7 +14,7 @@ cascade:
   PRODUCT_ROOT_NAME: Grafana Agent
 ---
 
-# {{% param "PRODUCT_NAME" %}}
+# Grafana Agent Flow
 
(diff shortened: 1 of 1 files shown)


jdbaldry:
I thought this worked with `{{%` which happens pre rendering?

(https://grafana.com/docs/writers-toolkit/write/shortcodes/#param)

jdbaldry:
Yeah in https://grafana.com/docs/agent/v0.39/flow/#grafana-agent-flow, the first heading is working and the second isn't. 

The difference is the use of `{{%` vs. `{{<`

clayton-cornell:
Yeah, you're right... I forgot the fix even though I recently documented the fix in the Writer's Toolkit :-(

PR #6252, opened 2024-01-26
pyroscope.scrape: Adjust handling of scrape_timeout

simonswine:
This automatically adapts the timeout when a ProfilingTarget is fetched
with Delta=true. In those cases the HTTP calll is block for
ScrapeInterval - 1s while the target profiles itself.

This also updates the ScrapeTimeout to 10s (default in prometheus).


korniltsev (approved):
LGTM

tpaschalis (approved):
Looks reasonable

PR #6254, opened 2024-01-26
Fix wrong Y axis on new chart

thampiotr:
This is an unreleased feature, so no need to update the changelog. 
* The % Y axis used a fraction not multiplied by 100.
* Added the `in <cluster>` to make it a bit more clear how the chart doesn't take into account other dash template variables from drop-downs.

image

### DIFF
--- a/operations/agent-flow-mixin/dashboards/prometheus.libsonnet
+++ b/operations/agent-flow-mixin/dashboards/prometheus.libsonnet
@@ -20,7 +20,7 @@ local scrapePanels(y_offset) = [
   // Scrape success rate
   (
     panel.new(title='Scrape success rate in $cluster', type='timeseries') +
-    panel.withUnit('percent') +
+    panel.withUnit('percentunit') +
     panel.withDescription(|||
       Percentage of targets successfully scraped by prometheus.scrape
       components.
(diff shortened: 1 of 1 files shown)


operations/agent-flow-mixin/dashboards/prometheus.libsonnet
```diff
@@ -19,7 +19,7 @@ local scrapePanels(y_offset) = [
 
   // Scrape success rate
   (
-    panel.new(title='Scrape success rate', type='timeseries') +
+    panel.new(title='Scrape success rate in $cluster', type='timeseries') +
```
rfratto, line 22: TIL you could do this 😛

operations/agent-flow-mixin/dashboards/prometheus.libsonnet
```diff
@@ -19,7 +19,7 @@ local scrapePanels(y_offset) = [
...
     panel.withUnit('percent') +
```
rfratto, line 23: Would it make more sense to change the unit than the query?

rfratto (approved):

operations/agent-flow-mixin/dashboards/prometheus.libsonnet
rfratto: Yeah, I think it's `percentunit` for a scale of 0-1

thampiotr: TIL :D thanks!

PR #6255, opened 2024-01-26
Change the default value of disable_high_cardinality_metrics to "true".

ptodev:
In #4769, I suggested that `disable_high_cardinality_metrics` default to `false` for a few reasons:

- Agent Flow was already showing high cardinality metrics, and removing them would have been a breaking change.
- OTel Collector enables these high cardinality metrics by default.

Unfortunately, it seems that those high cardinality metrics seem to really cause an understandable inconvenience for users. In the spirit of providing good defaults, it'd be best to exclude such metrics by default.

This will also simplify Static mode -> Flow converters, because high cardinality metrics are already disabled by default in Static mode.

I suspect that one day OTel Collector will have to also exclude these metrics by default. At the moment it's not a huge issue for the Collector, because you'd have to have a `telemetry.useOtelForInternalMetrics` feature gate enabled first.

cc @glindstedt

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ Main (unreleased)
 
 - Prohibit the configuration of services within modules. (@wildum)
 
+- For `otelcol.exporter` components, change the default value of `disable_high_cardinality_metrics` to `true`. (@ptodev)
+
 ### Features
 
 - A new `discovery.process` component for discovering Linux OS processes on the current host. (@korniltsev)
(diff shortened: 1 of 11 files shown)


clayton-cornell:
Doc change is small... looks Ok from my side.

jcreixell (approved):

clayton-cornell (approved):

clayton-cornell:
@jcreixell Is this one ready to merge?

PR #6269, opened 2024-01-29
add initial integration tests for exporters

marctc:
#### PR Description

Add integration test to check that redis and node exporter work properly.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/integration-tests/common/metrics_assert.go
+++ b/integration-tests/common/metrics_assert.go
@@ -98,8 +98,6 @@ func AssertMetricData(t *testing.T, query, expectedMetric string, testName strin
 		if assert.NotEmpty(c, metricResponse.Data.Result) {
 			assert.Equal(c, metricResponse.Data.Result[0].Metric.Name, expectedMetric)
 			assert.Equal(c, metricResponse.Data.Result[0].Metric.TestName, testName)
-			fmt.Println(metricResponse.Data.Result[0])
-			fmt.Println(metricResponse.Data.Result[0].Value)
 			assert.NotEmpty(c, metricResponse.Data.Result[0].Value.Value)
 			assert.Nil(c, metricResponse.Data.Result[0].Histogram)
 		}


integration-tests/tests/redis/redis_metrics_test.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"testing"
+
+	"github.com/grafana/agent/integration-tests/common"
+)
+
+func TestRedisMetrics(t *testing.T) {
+	var redisMetrics = []string{
+		"redis_up",
+		"redis_memory_used_bytes",
```
ptodev, line 14: I wonder if we should start documenting the metrics which exporters create? We could add a new section to the component docs, or could mention them in the description.
Given that we have tests to prove that they work, it should be especially easy.
cc @wildum

ptodev (approved):

integration-tests/tests/unix/unix_metrics_test.go
```diff
@@ -0,0 +1,163 @@
...
+		"node_timex_offset_seconds",
+		"node_timex_sync_status",
+		"node_uname_info",
+		"node_vmstat_oom_kill",
+		"node_vmstat_pgfault",
+		"node_vmstat_pgmajfault",
+		"node_vmstat_pgpgin",
+		"node_vmstat_pgpgout",
+		"node_vmstat_pswpin",
+		"node_vmstat_pswpout",
```
wildum, line 160: That's about 150 metrics. If they all fail that will spawn a HUGE block of text in the terminal which would make debugging hard. We should find a way to return either "all expected metrics are missing" or "the following metrics are missing:(list of the missing metrics)" if only a subset failed. You don't need to solve this in your branch but if you could add a "TODO" that would be nice

wildum (approved):

integration-tests/docker-compose.yaml
```diff
@@ -29,4 +29,8 @@ services:
       dockerfile: ./integration-tests/configs/prom-gen/Dockerfile
       context: ..
     ports:
-      - "9001:9001"
\ No newline at end of file
+      - "9001:9001"
+  redis:
+    image: redis:6.0.9-alpine
```
Dasomeone, line 34: This is a 3 year old image, maybe we want to use something newer? Especially concerning the recent news about security issues with older versions of alpine?

integration-tests/tests/redis/redis_metrics_test.go
Dasomeone: Similarly, maybe an idea here would be in the future to ingest this from a metrics file rather than have it hardcoded, similar to what we now export for the cloud integrations?

Dasomeone (commented):
Looks great to my eyes, just two minor comments we can talk about in future PRs!

PR #6270, opened 2024-01-29
Add an error diag to the converter from static to flow when `scrape_i…

erikbaranowski:
…ntegration` is set to true but no remote_write is defined.

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,9 @@ Main (unreleased)
 - Fix an issue where blocks having the same type and the same label across
   modules could result in missed updates. (@thampiotr)
 
+- Add an error diag to the converter from static to flow when `scrape_integration` is set
+  to true but no remote_write is defined. (@erikbaranowski)
+
 ### Other changes
 
 - Removed support for Windows 2012 in line with Microsoft end of life. (@mattdurham)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,8 +62,8 @@ Main (unreleased)
 - Fix an issue where blocks having the same type and the same label across
   modules could result in missed updates. (@thampiotr)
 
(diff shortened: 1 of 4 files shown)


converter/internal/staticconvert/testdata/integrations_no_rw.yaml
```diff
@@ -0,0 +1,4 @@
+integrations:
+  node_exporter:
+    scrape_integration: true
```
ptodev, line 3: Interesting feature. If we don't support it, then this is technically a feature parity issue because customers might want to use it to control their DPM.

That said, I wouldn't want to support it in Flow in the same way Static mode does it. We need to make a new component which can expose metrics on an HTTP endpoint using the Prometheus exposition format.

CHANGELOG.md
```diff
@@ -62,6 +62,9 @@ Main (unreleased)
 - Fix an issue where blocks having the same type and the same label across
   modules could result in missed updates. (@thampiotr)
 
+- Add an error diag to the converter from static to flow when `scrape_integration` is set
+  to true but no remote_write is defined. (@erikbaranowski)
```
ptodev, line 66:
```suggestion
- An error will be returned in the converter from Static to Flow when `scrape_integration` is set
  to `true` but no `remote_write` is defined. (@erikbaranowski)
```
I think this is a bit more clear.

ptodev (approved):

rfratto (approved):
Personally I think this is a bug in static mode; I don't believe it makes sense to configure an integration to scrape itself without having any remote_writes configured.

I'm OK with this fix as-is, but I do think it should result in an error in static mode.

PR #6277, opened 2024-01-30
Introduce remotecfg service

tpaschalis:
#### PR Description
This PR introduces a new service that makes it possible for Grafana Agent to run configuration fetched via a remote endpoint.

The PR is more easily reviewable going commit-by-commit.

No issue filed.

#### Notes to the Reviewer

My main concern is whether the instantiation of the new controller makes sense, both now and long term with the upcoming redesign of Modules.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/cmd/internal/flowmode/cmd_run.go
+++ b/cmd/internal/flowmode/cmd_run.go
@@ -33,6 +33,7 @@ import (
 	httpservice "github.com/grafana/agent/service/http"
 	"github.com/grafana/agent/service/labelstore"
 	otel_service "github.com/grafana/agent/service/otel"
+	remotecfgservice "github.com/grafana/agent/service/remotecfg"
 	uiservice "github.com/grafana/agent/service/ui"
 	"github.com/grafana/ckit/advertise"
 	"github.com/grafana/ckit/peer"
@@ -243,6 +244,14 @@ func (fr *flowRun) Run(configPath string) error {
 		EnablePProf:      fr.enablePprof,
 	})
 
+	remoteCfgService, err := remotecfgservice.New(remotecfgservice.Options{
+		Logger:      log.With(l, "service", "remotecfg"),
+		StoragePath: fr.storagePath,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to create the remotecfg service: %w", err)
+	}
+
 	uiService := uiservice.New(uiservice.Options{
 		UIPrefix: fr.uiPrefix,
 		Cluster:  clusterService.Data().(cluster.Cluster),
@@ -267,6 +276,7 @@ func (fr *flowRun) Run(configPath string) error {
 			clusterService,
 			otelService,
 			labelService,
+			remoteCfgService,
 		},
 	})
 
--- a/docs/sources/flow/reference/config-blocks/remotecfg.md
+++ b/docs/sources/flow/reference/config-blocks/remotecfg.md
@@ -13,7 +13,7 @@ title: remotecfg block
 # remotecfg block
 
 `remotecfg` is an optional configuration block that enables {{< param "PRODUCT_NAME" >}}
-to fetch and load configuration from a remote endpoint.
+to fetch and load the configuration from a remote endpoint.
 `remotecfg` is specified without a label and can only be provided once per
 configuration file.
 
@@ -44,13 +44,13 @@ The following arguments are supported:
 
 Name             | Type                 | Description                                      | Default     | Required
 -----------------|----------------------|--------------------------------------------------|-------------|---------
-`url`            | `string`             | The address of the API to poll for configuration | `""`        | no
-`id`             | `string`             | A self-reported ID                               | `see below` | no
-`metadata`       | `map(string)`        | A set of self-reported metadata                  | `{}`        | no
-`poll_frequency` | `duration`           | How often to poll the API for new configuration  | `"1m"`      | no
+`url`            | `string`             | The address of the API to poll for configuration. | `""`        | no
+`id`             | `string`             | A self-reported ID.                               | `see below` | no
+`metadata`       | `map(string)`        | A set of self-reported metadata.                  | `{}`        | no
+`poll_frequency` | `duration`           | How often to poll the API for new configuration.  | `"1m"`      | no
 
-If not set, the self-reported `id` that the Agent uses is a randomly generated,
-anonymous unique ID (UUID) that is stored on the Agent's storage path as
+If not set, the self-reported `id` that {{< param "PRODUCT_ROOT_NAME" >}} uses is a randomly generated,
+anonymous unique ID (UUID) that is stored in {{< param "PRODUCT_ROOT_NAME" >}}'s storage path as
 `agent_seed.json` so that it can persist across restarts.
 
 ## Blocks
--- a/docs/sources/flow/reference/config-blocks/remotecfg.md
+++ b/docs/sources/flow/reference/config-blocks/remotecfg.md
@@ -28,8 +28,8 @@ The [API definition][] for managing and fetching configuration that the
 remotecfg {
 	url = "SERVICE_URL"
 	basic_auth {
-		username      = USERNAME
-		password_file = PASSWORD_FILE
+		username      = "USERNAME"
+		password_file = "PASSWORD_FILE"
 	}
 
 	id             = constants.hostname
@@ -42,8 +42,8 @@ remotecfg {
 
 The following arguments are supported:
 
-Name             | Type                 | Description                                      | Default     | Required
------------------|----------------------|--------------------------------------------------|-------------|---------
+Name             | Type                 | Description                                       | Default     | Required
+-----------------|----------------------|---------------------------------------------------|-------------|---------
 `url`            | `string`             | The address of the API to poll for configuration. | `""`        | no
 `id`             | `string`             | A self-reported ID.                               | `see below` | no
 `metadata`       | `map(string)`        | A set of self-reported metadata.                  | `{}`        | no
@@ -54,6 +54,9 @@ If the `url` is not set, then the service block is a no-op.
 If not set, the self-reported `id` that the Agent uses is a randomly generated,
 anonymous unique ID (UUID) that is stored as an `agent_seed.json` file in the
 Agent's storage path so that it can persist across restarts.
+
+The `id` and `metadata` fields are used in the periodic request sent to the
+remote endpoint so that the API can decide what configuration to serve.
 
 ## Blocks
 
--- a/docs/sources/flow/reference/config-blocks/remotecfg.md
+++ b/docs/sources/flow/reference/config-blocks/remotecfg.md
@@ -33,8 +33,8 @@ remotecfg {
 	}
 
 	id             = constants.hostname
-	metadata       = {"cluster" = "dev-us-central-0", "namespace" = "agent-otlp"}
-	poll_frequency = "10m"
+	metadata       = {"cluster" = "dev", "namespace" = "otlp-dev"}
+	poll_frequency = "5m"
 }
 ```
 
@@ -49,9 +49,11 @@ Name             | Type                 | Description                           
 `metadata`       | `map(string)`        | A set of self-reported metadata.                  | `{}`        | no
 `poll_frequency` | `duration`           | How often to poll the API for new configuration.  | `"1m"`      | no
 
-If not set, the self-reported `id` that {{< param "PRODUCT_ROOT_NAME" >}} uses is a randomly generated,
-anonymous unique ID (UUID) that is stored in {{< param "PRODUCT_ROOT_NAME" >}}'s storage path as
-`agent_seed.json` so that it can persist across restarts.
+If the `url` is not set, then the service block is a no-op.
+
+If not set, the self-reported `id` that the Agent uses is a randomly generated,
+anonymous unique ID (UUID) that is stored as an `agent_seed.json` file in the
+Agent's storage path so that it can persist across restarts.
 
 ## Blocks
 
--- a/docs/sources/flow/reference/config-blocks/remotecfg.md
+++ b/docs/sources/flow/reference/config-blocks/remotecfg.md
@@ -47,7 +47,7 @@ Name             | Type                 | Description                           
 `url`            | `string`             | The address of the API to poll for configuration | `""`        | no
 `id`             | `string`             | A self-reported ID                               | `see below` | no
 `metadata`       | `map(string)`        | A set of self-reported metadata                  | `{}`        | no
-`poll_frequency` | `duration`           | How often to poll the API for new configuration  | `"1m"       | no
+`poll_frequency` | `duration`           | How often to poll the API for new configuration  | `"1m"`      | no
 
 If not set, the self-reported `id` that the Agent uses is a randomly generated,
 anonymous unique ID (UUID) that is stored on the Agent's storage path as
--- a/docs/sources/flow/reference/config-blocks/remotecfg.md
+++ b/docs/sources/flow/reference/config-blocks/remotecfg.md
@@ -1,0 +1,92 @@
+---
+aliases:
+- /docs/grafana-cloud/agent/flow/reference/config-blocks/remotecfg/
+- /docs/grafana-cloud/monitor-infrastructure/agent/flow/reference/config-blocks/remotecfg/
+- /docs/grafana-cloud/monitor-infrastructure/integrations/agent/flow/reference/config-blocks/remotecfg/
+- /docs/grafana-cloud/send-data/agent/flow/reference/config-blocks/remotecfg/
+canonical: remotecfgs://grafana.com/docs/agent/latest/flow/reference/config-blocks/remotecfg/
+description: Learn about the remotecfg configuration block
+menuTitle: remotecfg
+title: remotecfg block
+---
+
+# remotecfg block
+
+`remotecfg` is an optional configuration block that enables {{< param "PRODUCT_NAME" >}}
+to fetch and load configuration from a remote endpoint.
+`remotecfg` is specified without a label and can only be provided once per
+configuration file.
+
+The [API definition][] for managing and fetching configuration that the
+`remotecfg` block uses is available under the Apache 2.0 license.
+
+[API definition]: https://github.com/grafana/agent-remote-config
+
+## Example
+
+```river
+remotecfg {
+	url = "SERVICE_URL"
+	basic_auth {
+		username      = USERNAME
+		password_file = PASSWORD_FILE
+	}
+
+	id             = constants.hostname
+	metadata       = {"cluster" = "dev-us-central-0", "namespace" = "agent-otlp"}
+	poll_frequency = "10m"
+}
+```
+
+## Arguments
+
+The following arguments are supported:
+
+Name             | Type                 | Description                                      | Default     | Required
+-----------------|----------------------|--------------------------------------------------|-------------|---------
+`url`            | `string`             | The address of the API to poll for configuration | `""`        | no
+`id`             | `string`             | A self-reported ID                               | `see below` | no
+`metadata`       | `map(string)`        | A set of self-reported metadata                  | `{}`        | no
+`poll_frequency` | `duration`           | How often to poll the API for new configuration  | `"1m"       | no
+
+If not set, the self-reported `id` that the Agent uses is a randomly generated,
+anonymous unique ID (UUID) that is stored on the Agent's storage path as
+`agent_seed.json` so that it can persist across restarts.
+
+## Blocks
+
+The following blocks are supported inside the definition of `remotecfg`:
+
+Hierarchy | Block | Description | Required
+--------- | ----- | ----------- | --------
+basic_auth | [basic_auth][] | Configure basic_auth for authenticating to the endpoint. | no
+authorization | [authorization][] | Configure generic authorization to the endpoint. | no
+oauth2 | [oauth2][] | Configure OAuth2 for authenticating to the endpoint. | no
+oauth2 > tls_config | [tls_config][] | Configure TLS settings for connecting to the endpoint. | no
+tls_config | [tls_config][] | Configure TLS settings for connecting to the endpoint. | no
+
+The `>` symbol indicates deeper levels of nesting. For example,
+`oauth2 > tls_config` refers to a `tls_config` block defined inside
+an `oauth2` block.
+
+[basic_auth]: #basic_auth-block
+[authorization]: #authorization-block
+[oauth2]: #oauth2-block
+[tls_config]: #tls_config-block
+
+### basic_auth block
+
+{{< docs/shared lookup="flow/reference/components/basic-auth-block.md" source="agent" version="<AGENT_VERSION>" >}}
+
+### authorization block
+
+{{< docs/shared lookup="flow/reference/components/authorization-block.md" source="agent" version="<AGENT_VERSION>" >}}
+
+### oauth2 block
+
+{{< docs/shared lookup="flow/reference/components/oauth2-block.md" source="agent" version="<AGENT_VERSION>" >}}
+
+### tls_config block
+
+{{< docs/shared lookup="flow/reference/components/tls-config-block.md" source="agent" version="<AGENT_VERSION>" >}}
+
--- a/pkg/flow/flow_services.go
+++ b/pkg/flow/flow_services.go
@@ -3,6 +3,7 @@ package flow
 import (
 	"github.com/grafana/agent/pkg/flow/internal/controller"
 	"github.com/grafana/agent/pkg/flow/internal/dag"
+	"github.com/grafana/agent/pkg/flow/internal/worker"
 	"github.com/grafana/agent/service"
 )
 
@@ -51,4 +52,23 @@ func serviceConsumersForGraph(graph *dag.Graph, serviceName string, includePeerS
 	}
 
 	return consumers
+}
+
+// NewController returns a new, unstarted, isolated Flow controller so that
+// services can instantiate their own components.
+func (f *Flow) NewController(id string) service.Controller {
+	return newController(controllerOptions{
+		Options: Options{
+			ControllerID:    id,
+			Logger:          f.opts.Logger,
+			Tracer:          f.opts.Tracer,
+			DataPath:        f.opts.DataPath,
+			Reg:             f.opts.Reg,
+			OnExportsChange: f.opts.OnExportsChange,
+			Services:        f.opts.Services,
+		},
+		IsModule:       true,
+		ModuleRegistry: newModuleRegistry(),
+		WorkerPool:     worker.NewDefaultWorkerPool(),
+	})
 }
--- a/pkg/flow/flow_services.go
+++ b/pkg/flow/flow_services.go
@@ -82,7 +82,11 @@ type serviceController struct {
 }
 
 func (sc serviceController) Run(ctx context.Context) { sc.f.Run(ctx) }
-func (sc serviceController) LoadSource(source any, args map[string]any) error {
-	return sc.f.LoadSource(source.(*Source), args)
+func (sc serviceController) LoadSource(b []byte, args map[string]any) error {
+	source, err := ParseSource("", b)
+	if err != nil {
+		return err
+	}
+	return sc.f.LoadSource(source, args)
 }
 func (sc serviceController) Ready() bool { return sc.f.Ready() }
--- a/pkg/flow/flow_services.go
+++ b/pkg/flow/flow_services.go
@@ -67,8 +67,8 @@ func (f *Flow) NewController(id string) service.Controller {
 				Tracer:          f.opts.Tracer,
 				DataPath:        f.opts.DataPath,
 				Reg:             f.opts.Reg,
-				OnExportsChange: f.opts.OnExportsChange,
 				Services:        f.opts.Services,
+				OnExportsChange: nil, // NOTE(@tpaschalis, @wildum) The isolated controller shouldn't be able to export any values.
 			},
 			IsModule:       true,
 			ModuleRegistry: newModuleRegistry(),
--- a/pkg/flow/flow_services.go
+++ b/pkg/flow/flow_services.go
@@ -1,6 +1,8 @@
 package flow
 
 import (
+	"context"
+
 	"github.com/grafana/agent/pkg/flow/internal/controller"
 	"github.com/grafana/agent/pkg/flow/internal/dag"
 	"github.com/grafana/agent/pkg/flow/internal/worker"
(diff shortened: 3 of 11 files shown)


pkg/flow/flow.go
```diff
@@ -277,10 +277,12 @@ func (f *Flow) Run(ctx context.Context) {
 //
 // The controller will only start running components after Load is called once
 // without any configuration errors.
-func (f *Flow) LoadSource(source *Source, args map[string]any) error {
+func (f *Flow) LoadSource(src any, args map[string]any) error {
```
erikbaranowski, line 280: was the type changed here to prevent some dependency? It's a little weird to my eyes to have an `any` param that is casted to another single type before use.

service/service.go
```diff
@@ -54,6 +54,17 @@ type Host interface {
 	// GetServiceConsumers gets the list of services which depend on a service by
 	// name.
 	GetServiceConsumers(serviceName string) []Consumer
+
+	// NewController returns an unstarted, isolated Controller that a Service
+	// can use to instantiate its own components.
+	NewController(id string) Controller
```
erikbaranowski, line 60: I wonder about this one and if we need to do some additional work to make running multiple flow controllers a smooth experience if services can also start them. I know there's already some limitations in the debug UI for graphing modules 1.0.

service/remotecfg/remotecfg.go
```diff
@@ -0,0 +1,281 @@
...
+	// Reading from the API succeeded, let's try to load the contents.
+	if getErr == nil && parseErr == nil {
+		err := s.ctrl.LoadSource(src, nil)
+		if err != nil {
+			level.Error(s.opts.Logger).Log("msg", "could not load the API response contents", "err", err)
+		} else {
+			s.currentConfigHash = getHash(gcr.Msg.Content)
+		}
+	} else {
+		// Either the API call or parsing its contents failed, let's try the
```
erikbaranowski, line 168: I like having the cache idea built in up front. Very nice ❤️

pkg/flow/flow.go
tpaschalis: Yes, this is a way I tried to break a cyclic dependency.

I can write up some more details tomorrow so we can brainstorm if there's a better way to deal with this.

service/remotecfg/remotecfg.go
```diff
@@ -0,0 +1,281 @@
...
+		// on-disk cache.
+		level.Info(s.opts.Logger).Log("msg", "falling back to the on-disk cache")
```
spartan0x117, line 170: Doesn't have to be this PR, but should we add metrics for falling back to the cache and potentially also failing to fall back (doesn't exist or invalid)?

service/remotecfg/remotecfg.go
```diff
@@ -0,0 +1,281 @@
...
+		b, err := os.ReadFile(s.dataPath)
+		if err != nil {
+			level.Error(s.opts.Logger).Log("msg", "could not read from the on-disk cache", "err", err)
+		}
+		if len(b) > 0 {
```
spartan0x117, line 175: Should this section still run if there was an error reading the file?
```suggestion
		} else if len(b) > 0 {
```

service/remotecfg/remotecfg.go
```diff
@@ -0,0 +1,281 @@
...
+			}
+		}
+	}
+	s.mut.RUnlock()
+
+	for {
+		select {
+		case <-s.ticker.C:
+			s.mut.RLock()
+			gcr, err := s.asClient.GetConfig(ctx, s.getConfigRequest)
```
spartan0x117, line 195: Would it be better to create a new `agentv1.GetConfigRequest` on each request? They contain some amount of state and it may not be safe to reuse one. It is also unlikely to be a bottleneck in the grand scheme of things.

service/remotecfg/remotecfg.go
```diff
@@ -0,0 +1,281 @@
...
+	gcr, getErr = s.asClient.GetConfig(ctx, s.getConfigRequest)
+	if getErr == nil {
+		src, parseErr = flow.ParseSource(ServiceName, []byte(gcr.Msg.GetContent()))
+	}
+
+	// Reading from the API succeeded, let's try to load the contents.
+	if getErr == nil && parseErr == nil {
+		err := s.ctrl.LoadSource(src, nil)
+		if err != nil {
+			level.Error(s.opts.Logger).Log("msg", "could not load the API response contents", "err", err)
```
marctc, line 163: should we add telemetry for some of the failure cases? I believe @spartan0x117 added some in static mode that can be reused (and therefore the alerts).

cmd/internal/flowmode/cmd_run.go
```diff
@@ -262,6 +271,7 @@ func (fr *flowRun) Run(configPath string) error {
 			clusterService,
 			otelService,
 			labelService,
+			remoteCfgService,
```
wildum, line 274: the service is created and will run for all agents? That would mean that we create and run a controller without any config, can this be a problem?

service/remotecfg/remotecfg.go
wildum, line 168: The risk with retrieving the cached config from the disk is that it may run an old config that I don't want to run anymore just because I misconfigured the connection to my new API or my new config is not correct. In this case I would rather have my agent not run anything than running something I don't want it to run. Maybe there should be an argument to let the user decide what behavior he wants?

service/remotecfg/remotecfg.go
tpaschalis: Yes, I'd like to add them in a follow-up PR.

cmd/internal/flowmode/cmd_run.go
tpaschalis: I don't think there would be a problem here and toggling the service on and off dynamically would increase the complexity of the feature.

Do you have any specific concerns in mind?

wildum: If a user doesn't configure it via the corresponding river block, then there is no use to run the service because there is no URL to retrieve the config from, right? If yes then we can just create the service in the loader when we populate the nodes and only if there is a block for it.
What happens currently when you don't use the service but it still runs? Would it pick up a cached config if it finds one? It might also log things and spawn go routines (like the worker pool) which may bring confusions for the users

wildum: Sorry I was writing the comment above before refreshing the page so i did not see the last comment

service/remotecfg/remotecfg.go
```diff
@@ -118,12 +133,93 @@ var _ service.Service = (*Service)(nil)
 func (s *Service) Run(ctx context.Context, host service.Host) error {
 	s.ctrl = host.NewController(ServiceName)
 
+	// Run the service's own controller.
 	go func() {
 		s.ctrl.Run(ctx)
 	}()
+
+	// We're on the initial start-up of the service.
```
rfratto, line 143: I'm not a huge fan of how this always executes, even if the user hasn't configured the service. It feels like we need to refactor this to account for the service running when the user hasn't configured it.

service/remotecfg/remotecfg.go
```diff
@@ -118,12 +133,93 @@ var _ service.Service = (*Service)(nil)
...
+	}
+
+	// Reading from the API succeeded, let's try to load the contents.
+	if getErr == nil && parseErr == nil {
+		err := s.ctrl.LoadSource(src, nil)
+		if err != nil {
+			level.Error(s.opts.Logger).Log("msg", "could not load the API response contents", "err", err)
+		} else {
+			s.currentConfigHash = getHash(gcr.Msg.Content)
+		}
```
rfratto, line 174: Why doesn't the initial load also cache to disk?

service/remotecfg/remotecfg.go
tpaschalis: I wasn't able to find anything stateful with a glance, but I'm OK with trying this.

service/remotecfg/remotecfg.go
```diff
@@ -118,12 +133,93 @@ var _ service.Service = (*Service)(nil)
...
+				} else {
+					s.currentConfigHash = getHash(string(b))
+				}
+			}
+		}
+	}
+
 	for {
 		select {
+		case <-s.ticker.C:
```
rfratto, line 201: The code here feels a bit repetitive, and it looks like it could be made more DRY. I'd like for the Run function to be refactored a bit to carve out reusable bits of logic.

service/remotecfg/remotecfg.go
```diff
@@ -133,6 +229,36 @@ func (s *Service) Run(ctx context.Context, host service.Host) error {
...
+		// TODO: We either never set the block on the first place,
+		// or recently removed it. Make sure we stop everything gracefully
+		// before returning.
+		return nil
+	}
+
+	// TODO: Should we also include the username, or some other property on the
+	// hash for the directory name?
+	s.dataPath = filepath.Join(s.opts.StoragePath, ServiceName, getHash(newArgs.URL))
+	s.ticker.Reset(newArgs.PollFrequency)
```
rfratto, line 271: The way this works means that you won't get the new config until after the `PollFrequency` elapses, which might be confusing to users who reloaded their config (imagine changing the frequency to `1h` but it takes one hour to get the first config after the reload).

I think calling Update should immediately trigger a new retrieval.

service/remotecfg/remotecfg.go
```diff
@@ -0,0 +1,286 @@
...
+
+	s.mut.RLock()
+	req := connect.NewRequest(&agentv1.GetConfigRequest{
+		Id:       s.args.ID,
+		Metadata: s.args.Metadata,
+	})
+	gcr, err := s.asClient.GetConfig(ctx, req)
+	if err != nil {
+		return err
+	}
```
rfratto, line 161: Returning an error from Run permanently terminates it for the remainder of the process lifetime. Is that what you want?

rfratto (changes requested):
Looks mostly good, just have a few pieces of feedback.

service/remotecfg/remotecfg.go
tpaschalis: Yes, I've found this is buggy. Similarly, there's a race condition depending on whether Update has had a chance to be called before this.

I'll refactor and synchronize this properly.

service/remotecfg/remotecfg.go
```diff
@@ -0,0 +1,281 @@
...
+	// won't run otherwise
+	if a.HTTPClientConfig != nil {
+		return a.HTTPClientConfig.Validate()
+	}
+
+	return nil
+}
+
+// Data includes information associated with the remotecfg service.
+type Data struct {
```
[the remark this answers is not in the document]
tpaschalis: You're right; I mainly left it to conform with the other services, but I'm removing it.

service/remotecfg/remotecfg.go
tpaschalis: I'm resolving this comment as after some refactoring I believe the logic is more straightforward now.

service/remotecfg/remotecfg.go
```diff
@@ -0,0 +1,289 @@
...
+
+// Run implements [service.Service] and starts the remotecfg service. It will
+// run until the provided context is canceled or there is a fatal error.
+func (s *Service) Run(ctx context.Context, host service.Host) error {
+	s.ctrl = host.NewController(ServiceName)
+
+	// Let's try to read from the API or the local cache as a fallback.
+	var b1, b2 []byte
+	b1, err := s.getAPIConfig()
+	b2, err2 := s.getCachedConfig()
```
rfratto, line 133: I've been looking at this PR again after reading @wildum's [comment](https://github.com/grafana/agent/pull/6277#discussion_r1472607048), trying to make sense of when cache invalidation should occur. 

I came up with two cases that I think are the most important:

* The existing cache should be invalidated when the `remotecfg` block is no longer being configured. 
* The existing cache should be invalidated when the `remotecfg` block is changed to point to a different server, or with a different authentication username, or potentially if the agent's identifying labels have changed. 

I don't think I have a fully exhaustive list for the second case, but it seems like the items I listed all fall into the category of "the endpoint is different or the agent's identity is different."

pkg/flow/flow_services.go
```diff
@@ -52,3 +53,22 @@ func serviceConsumersForGraph(graph *dag.Graph, serviceName string, includePeerS
...
+// services can instantiate their own components.
+func (f *Flow) NewController(id string) service.Controller {
+	return newController(controllerOptions{
+		Options: Options{
+			ControllerID:    id,
+			Logger:          f.opts.Logger,
+			Tracer:          f.opts.Tracer,
+			DataPath:        f.opts.DataPath,
+			Reg:             f.opts.Reg,
+			OnExportsChange: f.opts.OnExportsChange,
```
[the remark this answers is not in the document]
tpaschalis: Why should it always be nil here?

service/remotecfg/remotecfg.go
```diff
@@ -0,0 +1,323 @@
...
+
+	if err != nil && err2 != nil {
+		level.Debug(s.opts.Logger).Log("msg", "failed to get any configuration during startup", "apiErr", err, "cacheErr", err2)
+		return
+	}
+
+	err = s.parseAndLoad(b1)
+	if err != nil {
+		err = s.parseAndLoad(b2)
```
wildum, line 252: this should probably not be called if err2 != nil

tpaschalis: I tried to tradeoff the check for some readability here; it should be OK to call it since an error in err2 means that b2 will be empty, but will check if I can improve it.

service/remotecfg/remotecfg.go
tpaschalis: I've refactored the fetch-from-api-and-fallback-to-cache logic in a new method that's called from Update.

service/remotecfg/remotecfg.go
tpaschalis: This was a little confusing since we both want to execute the fetching logic on an Update _and_ on the startup of `Run`, but the conditions to it _also_ come from Update.

As a tradeoff, I've kept the initial call here, but added an early exit on `fetch()` so it shouldn't have any effect when the service is not running.

service/remotecfg/remotecfg.go
```diff
@@ -0,0 +1,139 @@
...
+	go func() {
+		s.ctrl.Run(ctx)
+	}()
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		default:
+		}
+	}
```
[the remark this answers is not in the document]
tpaschalis: Thanks for the input. I'm trying to become better at this.

I feel the order in which I'm building a feature isn't necessarily the same order that's most useful for reviewing; I'll keep this in mind for the future.

service/service.go
```diff
@@ -54,6 +54,17 @@ type Host interface {
...
+}
+
+// Controller is implemented by flow.Flow.
+type Controller interface {
+	Run(ctx context.Context)
+	LoadSource(source any, args map[string]any) error
```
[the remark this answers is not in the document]
tpaschalis: Thanks. I've still kept the service.Controller interface the same ("Go interfaces generally belong in the package that uses values of the interface type"), but used this to not bleed any changes to other parts of the codebase.

service/remotecfg/remotecfg.go
tpaschalis: I'd like to avoid multiple configuration options and try to offer a more opinionated way of people using the service and we've identified this as a necessary feature.

In any case, the hash currently takes into account the entirety of the Arguments block, so it's harder to accidentally reuse an erroneous value.

cmd/internal/flowmode/cmd_run.go
tpaschalis: We've caught up offline about a potential way to conditionally run pipelines. Since there isn't a pressing need for that (the overhead is just an extra running goroutine), and that it'd work differently for other services, we might revisit in the future but decided to not do anything for now.

pkg/flow/flow_services.go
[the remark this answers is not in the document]
tpaschalis: Thanks. After catching up offline, I've set this to nil so to expliticly disallow any `export` blocks to be configured here.

docs/sources/flow/reference/config-blocks/remotecfg.md
```diff
@@ -0,0 +1,92 @@
...
+- /docs/grafana-cloud/send-data/agent/flow/reference/config-blocks/remotecfg/
+canonical: remotecfgs://grafana.com/docs/agent/latest/flow/reference/config-blocks/remotecfg/
+description: Learn about the remotecfg configuration block
+menuTitle: remotecfg
+title: remotecfg block
+---
+
+# remotecfg block
+
+`remotecfg` is an optional configuration block that enables {{< param "PRODUCT_NAME" >}}
```
mattdurham, line 15: Does this inherently mean that they cannot define other config blocks in the remote cfg? ie http,logging ect?

docs/sources/flow/reference/config-blocks/remotecfg.md
```diff
@@ -0,0 +1,92 @@
...
+}
+```
+
+## Arguments
+
+The following arguments are supported:
+
+Name             | Type                 | Description                                      | Default     | Required
+-----------------|----------------------|--------------------------------------------------|-------------|---------
+`url`            | `string`             | The address of the API to poll for configuration | `""`        | no
```
spartan0x117, line 47: suggestion: this is marked as not required because the service itself isn't required for the agent to run, right? But for the service to function, a URL does need to be provided. Perhaps there should be a note below the table that the URL _is_ functionally required if the service is to run.

docs/sources/flow/reference/config-blocks/remotecfg.md
```diff
@@ -0,0 +1,92 @@
...
+`id`             | `string`             | A self-reported ID                               | `see below` | no
+`metadata`       | `map(string)`        | A set of self-reported metadata                  | `{}`        | no
+`poll_frequency` | `duration`           | How often to poll the API for new configuration  | `"1m"       | no
```
spartan0x117, line 50:
```suggestion
`poll_frequency` | `duration`           | How often to poll the API for new configuration  | `"1m"`      | no
```

docs/sources/flow/reference/config-blocks/remotecfg.md
tpaschalis: Yes; right now services blocks are always implicitly constructed (so there's no way to have an required field).

We can expand on this with a note below.

docs/sources/flow/reference/config-blocks/remotecfg.md
```diff
@@ -0,0 +1,92 @@
...
+
+If not set, the self-reported `id` that the Agent uses is a randomly generated,
+anonymous unique ID (UUID) that is stored on the Agent's storage path as
+`agent_seed.json` so that it can persist across restarts.
```
spartan0x117, line 54: nit:
```suggestion
If not set, the self-reported `id` that the Agent uses is a randomly generated,
anonymous unique ID (UUID) that is stored in the Agent's storage path as
`agent_seed.json` so that it can persist across restarts.
```

spartan0x117: Actually looking at this a second time this may not be any clearer? :thinking:

docs/sources/flow/reference/config-blocks/remotecfg.md
wildum: It cannot contain tracing and logging blocks, but it technically can contain argument and export blocks. I think that we should forbid it as it makes no sense to have them (the controller is marked as isolated)

service/remotecfg/remotecfg.go
```diff
@@ -0,0 +1,318 @@
...
+
+	err = s.parseAndLoad(b1)
+	if err != nil && err2 == nil {
+		err = s.parseAndLoad(b2)
+	}
+
+	if err != nil {
+		level.Error(s.opts.Logger).Log("msg", "failed to load remote cfg during startup", "err", err)
+	}
```
wildum, line 256: I find the logic a bit hard to follow, especially because we are keeping track of two errors. I would like to suggest the following approach (not tested):

```
b, err := s.getAPIConfig()
if err != nil {
  err = s.parseAndLoad(b)
  if err == nil {
    return
  }
}
level.Error(s.opts.Logger).Log("msg", "failed to load remote cfg during startup, falling back to cache", "err", err)
b, err = s.getCachedConfig()
if err != nil {
    err = s.parseAndLoad(b)
    if err == nil {
      return
   }
}
level.Error(s.opts.Logger).Log("msg", "failed to load cached remote cfg during startup", "err", err)
```

docs/sources/flow/reference/config-blocks/remotecfg.md
```diff
@@ -0,0 +1,92 @@
...
+canonical: remotecfgs://grafana.com/docs/agent/latest/flow/reference/config-blocks/remotecfg/
+description: Learn about the remotecfg configuration block
+menuTitle: remotecfg
+title: remotecfg block
+---
+
+# remotecfg block
+
+`remotecfg` is an optional configuration block that enables {{< param "PRODUCT_NAME" >}}
+to fetch and load configuration from a remote endpoint.
```
clayton-cornell, line 16:
```suggestion
to fetch and load the configuration from a remote endpoint.
```

docs/sources/flow/reference/config-blocks/remotecfg.md
clayton-cornell, line 47:
```suggestion
`url`            | `string`             | The address of the API to poll for configuration. | `""`        | no
```

docs/sources/flow/reference/config-blocks/remotecfg.md
```diff
@@ -0,0 +1,92 @@
...
+## Arguments
+
+The following arguments are supported:
+
+Name             | Type                 | Description                                      | Default     | Required
+-----------------|----------------------|--------------------------------------------------|-------------|---------
+`url`            | `string`             | The address of the API to poll for configuration | `""`        | no
+`id`             | `string`             | A self-reported ID                               | `see below` | no
+`metadata`       | `map(string)`        | A set of self-reported metadata                  | `{}`        | no
+`poll_frequency` | `duration`           | How often to poll the API for new configuration  | `"1m"`      | no
```
clayton-cornell, line 50:
```suggestion
`id`             | `string`             | A self-reported ID.                               | `see below` | no
`metadata`       | `map(string)`        | A set of self-reported metadata.                  | `{}`        | no
`poll_frequency` | `duration`           | How often to poll the API for new configuration.  | `"1m"`      | no
```

docs/sources/flow/reference/config-blocks/remotecfg.md
clayton-cornell: We should not call it `the Agent`. It's a collector. Call it by its name: {{< param "PRODUCT_ROOT_NAME" >}} for Grafana Agent or {{< param "PRODUCT_NAME" >}} for Grafana Agent Flow.

docs/sources/flow/reference/config-blocks/remotecfg.md
```diff
@@ -0,0 +1,92 @@
...
+
+If not set, the self-reported `id` that the Agent uses is a randomly generated,
```
clayton-cornell, line 52:
```suggestion
If not set, the self-reported `id` that {{< param "PRODUCT_ROOT_NAME" >}} uses is a randomly generated,
```

docs/sources/flow/reference/config-blocks/remotecfg.md
```diff
@@ -0,0 +1,92 @@
...
+anonymous unique ID (UUID) that is stored on the Agent's storage path as
+`agent_seed.json` so that it can persist across restarts.
```
clayton-cornell, line 54:
```suggestion
anonymous unique ID (UUID) that is stored on t{{< param "PRODUCT_ROOT_NAME" >}}'s storage path as
`agent_seed.json` so that it can persist across restarts.
```

clayton-cornell (commented):
Some suggestions for docs

service/remotecfg/remotecfg.go
```diff
@@ -0,0 +1,318 @@
...
+
+// fetch attempts to read configuration from the API and the local cache
+// and then parse/load in order of preference.
+func (s *Service) fetch() {
+	if !s.isEnabled() {
+		return
+	}
+
+	var b1, b2 []byte
+	b1, err := s.getAPIConfig()
```
erikbaranowski, line 241: I wonder if the reason the multiple errors are needed is due to the algorithm order. I think what you have is working, but this may clear it up:

1) getAPIConfig
2) if 1 succeeds, attempt parseAndLoad the result from 1
3) if 1 or 2 fails, getCachedConfig
4) if 3 succeeds, parseAndLoad the result from 3

We could optimize that getAPIConfig and getCachedConfig are called in parallel with go routines if we want.

erikbaranowski: @wildum wrote a version of this below*

https://github.com/grafana/agent/pull/6277/files#r1481818052

service/remotecfg/remotecfg.go
```diff
@@ -0,0 +1,318 @@
...
+
+	// Run the service's own controller.
+	go func() {
+		s.ctrl.Run(ctx)
+	}()
+
+	for {
+		select {
+		case <-s.ch:
+			b, err := s.getAPIConfig()
```
erikbaranowski, line 151: Is there a way we can make s.fetch() robust enough to be used here in addition to the other 2 references? For example, the call to s.fetch() in this func does not read/write a cache file. In practice probably not a huge deal on startup but those little nuances could be tough to track down in the future.

service/service.go
tpaschalis: Ideally yes. I'd like at least to explore with what can be displayed in the UI (even if it's eg. a new page just for remotecfg).

service/remotecfg/remotecfg.go
```diff
@@ -0,0 +1,318 @@
...
+		s.mut.Unlock()
+		return nil
+	}
+
+	s.mut.Lock()
+	hash, err := newArgs.Hash()
+	if err != nil {
+		return err
+	}
+	s.dataPath = filepath.Join(s.opts.StoragePath, ServiceName, hash)
```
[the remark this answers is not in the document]
tpaschalis: I think we _do_ need to add more than just the URL for the hash location though, eg. in case the username changes we'll want to not try to load configuration from the previous user.

Do you suggest just manually hash the fields instead of the entire struct in that helper instead?

service/remotecfg/remotecfg.go
tpaschalis: I've specifically avoided falling back to a cache here; the flow controller should be rejecting any invalid config coming its way, so we don't need to fall back on anything here.

Do you believe it would make this easier to read if we extracted the flushing to disk in another helper or added an option on fetch to conditionally flush to disk instead?

service/remotecfg/remotecfg.go
tpaschalis: Readability matters, right? 😅

The reason I tried the current way is that we're all used to `if err != nil` conditions in 90% of the code we read/write, so it's be easy to skim over and misunderstand what's happening here, but I guess the added implicitness isn't worth it.

tpaschalis: The reason I tried the current way is that we're all used to `if err != nil` conditions in 90% of the code we read/write, so it's be easy to skim over and misunderstand what's happening here, but I guess the added implicitness isn't worth it.

wildum: agreed that the `if err == nil` is a bit special but imo it's one of the rare cases where it makes things easier to follow

tpaschalis: In your above example, _all_ conditions have to be `err == nil`, so it's indeed hard to write/read even as a small example.

In any case, it doesn't matter for how the feature works, so I'm ok with changing it.

service/remotecfg/remotecfg.go
tpaschalis: I'm ok with either way, but I don't think it makes sense to introduce more complex optimizations at this stage.

docs/sources/flow/reference/config-blocks/remotecfg.md
tpaschalis: Hmm, that's a good point, should likely be changed.

service/remotecfg/remotecfg.go
```diff
@@ -0,0 +1,318 @@
...
+
+	return &Service{
+		opts:   opts,
+		ticker: time.NewTicker(math.MaxInt64),
+	}, nil
+}
+
+// Data is a no-op for the remotecfg service.
+func (s *Service) Data() any {
+	return nil
```
mattdurham, line 122: Should this be a panic instead? Normally we dont panic but this feels like the same category as duplicate component ids.

service/remotecfg/remotecfg.go
```diff
@@ -0,0 +1,318 @@
...
+				level.Error(s.opts.Logger).Log("msg", "failed to load configuration from the API", "err", err)
+				continue
+			}
+
+			level.Info(s.opts.Logger).Log("msg", "new remote configuration loaded successfully")
+
+			// If successful, flush to disk and keep a copy.
+			err = os.WriteFile(s.dataPath, b, 0750)
+			if err != nil {
+				level.Error(s.opts.Logger).Log("msg", "failed to flush remote configuration contents the on-disk cache", "err", err)
```
mattdurham, line 176: To be clear we are fine with the configuration it is running being different from the value on the filesystem?

tpaschalis: Sorry, I didn't get what you mean here.

mattdurham: If it fails writing the new config it will still load the remote config so now the cache and running configs are different. The only big use case I could see is if we needed them to investigate the cache, but I think that is pretty far fetched.

service/remotecfg/remotecfg.go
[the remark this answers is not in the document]
erikbaranowski: Since this service controls the folder `[storage_path]/remotecfg` do we need to use a hash at all for where the config is cached? Or can we avoid that altogether?

(This is not to say we shouldn't hash the config contents and use that for the reloading and the caching logic)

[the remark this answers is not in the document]
erikbaranowski: To be more specific, what are the downsides of storing the cached config at the full path `[storage_path]/remotecfg/cache.river`

pkg/flow/flow_services.go
```diff
@@ -52,3 +55,38 @@ func serviceConsumersForGraph(graph *dag.Graph, serviceName string, includePeerS
...
+			Options: Options{
+				ControllerID:    id,
+				Logger:          f.opts.Logger,
+				Tracer:          f.opts.Tracer,
+				DataPath:        f.opts.DataPath,
+				Reg:             f.opts.Reg,
+				Services:        f.opts.Services,
+				OnExportsChange: nil, // NOTE(@tpaschalis, @wildum) The isolated controller shouldn't be able to export any values.
+			},
+			IsModule:       true,
```
tpaschalis, line 73: cc @rfratto as @mattdurham pointed out, we likely _don't_ want this new controller to run as a module controller, so that remote configuration can manage its own logging/tracing block for example.

Any concerns in regards to that?

service/remotecfg/remotecfg.go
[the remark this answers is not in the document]
tpaschalis: The downside is potentially running with a new username/password, failing to authenticate and then falling back to the cached configuration that already existed for another user, or another URL.

pkg/flow/flow_services.go
tpaschalis: Eh, I see. Then I think it's something we can work out later on (and after modules v2.0 are out).

wildum: Yeah these isModule conditions spread around the controller are scary. This is something that we should be able to address in the refactor of the controller

service/remotecfg/remotecfg.go
tpaschalis: Yeah, I don't see this as a major issue TBH. The cache is a best-effort thing only, the service should run fine even if storage-path is not writeable.

service/remotecfg/remotecfg.go
tpaschalis: I don't mind either way, but I don't see a benefit for panicking here.

service/remotecfg/remotecfg.go
tpaschalis: We've talked offline and reworked the logic into smaller, reusable pieces between fetching remotely and locally only; I personally think it does look cleaner that way. Thanks @erikbaranowski !

erikbaranowski (approved):
LGTM, nice work 🎉

pkg/flow/flow_services.go
mattdurham: How do we enforce what they can or cannot run in the main config when using remote config? We dont necessarily need it for this pr but I think it needs to be addressed before we productionize it.

tpaschalis: Sorry for jumping the gun here, I should have made sure we'd finalized the discussion first.

Since we mark the new controller with `IsModule: true`, then controller code will inherit the same services as run by the root controller. As such, if the remotecfg controller detects a config block, it will reject the new config being applied.

PR #6279, opened 2024-01-30
Docs/update shortcode syntax

clayton-cornell:
#### PR Description

This PR updates the shortcode syntax used for Note, Warning, Caution, and Tip, and the syntax used for `param` in the body text.

The `%` and `<` characters in the shortcode control the stage that the shortcode is parsed during the site generation. 

* We should use `{{% shortcode %}}` for content that needs to be parsed in early stages (for example `{{% parm VARIABLE %}}` in headings and all `docs\reference` shortcodes. 
* We should use `{{< shortcode >}}` for all body text content (for example Note, Tip, Warning and Caution as well as `{{< parm VARIABLE >}}`)

Changes in this PR:
* Updated the Note definition in `links_to_types.go` and regenerated the reference topics
* Used find and sed to search and replace all other instances where the wrong syntax was used in Note, Warning Tip, Caution:
  * `find . -type f -exec sed -i "s/{{% admonition type=\"note\" %}}/{{< admonition type=\"note\" >}}/g" {} +`
  * `find . -type f -exec sed -i "s/{{% admonition type=\"caution\" %}}/{{< admonition type=\"caution\" >}}/g" {} +`
  * `find . -type f -exec sed -i "s/{{% admonition type=\"warning\" %}}/{{< admonition type=\"warning\" >}}/g" {} +`
  * `find . -type f -exec sed -i "s/{{%\/admonition %}}/{{< \/admonition >}}/g" {} +`
  * `find . -type f -exec sed -i "s/{{% \/admonition %}}/{{< \/admonition >}}/g" {} +`
* Used grep to find any other use of `{{%`:
  * `grep -Rn '{{%'`

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/docs/generator/links_to_types.go
+++ b/docs/generator/links_to_types.go
@@ -39,8 +39,8 @@ func (l *LinksToTypesGenerator) Generate() (string, error) {
 
 	note := `
 {{< admonition type="note" >}}
-Connecting some components may not be sensible or components may require further configuration to make the 
-connection work correctly. Refer to the linked documentation for more details.
+Connecting some components may not be sensible or components may require further configuration to make the connection work correctly.
+Refer to the linked documentation for more details.
 {{< /admonition >}}
 `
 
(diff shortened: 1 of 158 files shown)


jdbaldry (approved):
Nice!

PR #6284, opened 2024-01-31
fix for integrations v2 marshaling multiplexes

### LINKED ISSUE #3627: static: multiplex integrations-next instances cannot have their config marshalled
As an example, the mongodb integration uses the shim to register itself to the v2 'integrations-next' functionality as a multiplex integration.

When using a config file like the following and starting the Agent, we can indeed monitor multiple mongodb instances.
The issue presents itself when using `-config.enable-read-api`. The `/-/config` endpoint produces an error as it's somehow still expected to be a single key instead of an array.

```yaml
metrics:
  global:
    scrape_interval: 5s
    remote_write:
      - url: https://prometheus-prod-01-eu-west-0.grafana.net/api/prom/push
        basic_auth:
          username: user
          password: pass
  configs:
    - name: default
      scrape_configs:
        - job_name: agent
          static_configs:
            - targets: ['127.0.0.1:12345']

integrations:
  mongodb_configs:
    - mongodb_uri: "mongodb://mongodb0.example.com:27017"
    - mongodb_uri: "mongodb://mongodb0.example.com:27018"
    - mongodb_uri: "mongodb://mongodb0.example.com:27019"
```

```
$ ./build/grafana-agent -config.file ~/integrations-next-multiplex.yaml -config.enable-read-api -enable-features=integrations-next
... 

$ curl localhost:12345/-/config
failed to marshal config: integration "mongodb" may not be defined more than once
```

erikbaranowski:

Fixes #3627

#### Notes to the Reviewer

Output now shows for the integrations in the linked bug issue:
image

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/pkg/integrations/v2/register.go
+++ b/pkg/integrations/v2/register.go
@@ -229,7 +229,7 @@ func MarshalYAML(v interface{}) (interface{}, error) {
 		}
 
 		if _, exists := uniqueSingletons[fieldName]; exists && integrationType == TypeSingleton {
-				return nil, fmt.Errorf("integration %q may not be defined more than once", fieldName)
+			return nil, fmt.Errorf("integration %q may not be defined more than once", fieldName)
 		}
 		uniqueSingletons[fieldName] = struct{}{}
 
--- a/pkg/integrations/v2/register.go
+++ b/pkg/integrations/v2/register.go
@@ -228,10 +228,12 @@ func MarshalYAML(v interface{}) (interface{}, error) {
 			panic(fmt.Sprintf("config not registered: %T", data))
 		}
 
-		if _, exists := uniqueSingletons[fieldName]; exists {
-			return nil, fmt.Errorf("integration %q may not be defined more than once", fieldName)
+		if c.IsSingleton() {
+			if _, exists := uniqueSingletons[fieldName]; exists {
+				return nil, fmt.Errorf("integration %q may not be defined more than once", fieldName)
+			}
+			uniqueSingletons[fieldName] = struct{}{}
 		}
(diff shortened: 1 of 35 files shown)


pkg/integrations/v2/register.go
```diff
@@ -228,10 +228,12 @@ func MarshalYAML(v interface{}) (interface{}, error) {
...
 		}
 
-		if _, exists := uniqueSingletons[fieldName]; exists {
-			return nil, fmt.Errorf("integration %q may not be defined more than once", fieldName)
+		if integrationType == TypeSingleton {
+			if _, exists := uniqueSingletons[fieldName]; exists {
+				return nil, fmt.Errorf("integration %q may not be defined more than once", fieldName)
+			}
+			uniqueSingletons[fieldName] = struct{}{}
 		}
```
rfratto, line 236: I don't care if this changes, but I do want to point out that we don't need to introduce nested ifs here:

```suggestion
		if _, exists := uniqueSingletons[fieldName]; exists && integrationType == TypeSingleton {
				return nil, fmt.Errorf("integration %q may not be defined more than once", fieldName)
		}
		uniqueSingletons[fieldName] = struct{}{}
```

rfratto (approved):
Thanks Erik!

pkg/integrations/v2/register.go
erikbaranowski: ya I was trying to isolate the cost of resetting the value in the map but I think the legibility is worth changing it as you suggest

PR #6289, opened 2024-02-01
Demonstrate one potential implementation for `declare` 

rfratto:
**DO NOT MERGE; DEMONSTRATIVE PR ONLY**

> **NOTE**: This PR is meant to be reviewed commit-by-commit. 

This PR demonstrates one potential implementation for `declare` blocks as defined in #5547. The purpose of this PR is to compare/contrast approaches towards modelling the way declare blocks are scoped and how components are retrieved, with the other approach being in the [declare-new-modules branch](https://github.com/grafana/agent/tree/declare-new-modules) (which had a PR at #6253). 

The architecture of this implementation relies on the `ComponentRegistry` interface being used to look up built-in and custom components. The ComponentRegistry is passed around to child controllers, and child controllers create an interface implementation which extends the provided registry with custom component definitions in their local scope. 

Keep these things in mind when reviewing this PR: 

1. There is a [bug](https://github.com/rfratto/agent/blob/d56eb0096d855aeea38d4a971decc5fac959b2b5/pkg/flow/internal/controller/node_custom_component.go#L143-L147) left in the code. I don't think the bug detracts away from the discussion of evaluating this design, so I'm leaving it in for now. 

2. For this PR, `declare` blocks are not able to access top-level component exports, which is a behaviour described in #5547. This detail should likely not be implemented in any initial implementation for `declare`, and we should revisit the RFC to see if that capability should be scoped out entirely for now. 

3. The final commit copies the unit tests from #6253 to ensure that the same test cases pass.

### DIFF
--- a/component/registry.go
+++ b/component/registry.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/go-kit/log"
 	"github.com/grafana/regexp"
+	"github.com/grafana/river/ast"
 	"github.com/prometheus/client_golang/prometheus"
 	"go.opentelemetry.io/otel/trace"
 	"golang.org/x/exp/maps"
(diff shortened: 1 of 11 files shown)


rfratto:
Closing; the PR exists which is enough for reviewing it as a potential implementation, but it's still never intended to be merged.

PR #6291, opened 2024-02-02
Declare new modules (second attempt)

wildum:
#### PR Description

This PR introduces the declare functionality defined by the https://github.com/grafana/agent/issues/5547.

#### Notes to the Reviewer

The PR can be reviewed commit by commit.
It contains some parts of the import logic to help the reviewers understand how the import will fit into the current design. The missing parts are left as "TODO" and will come via another PR. (it has already been implemented and tested with this design)

The structure of the design relies on a ComponentNodeManager which holds a builtinComponentRegistry and a customComponentRegistry. It is used by the loader to create ComponentNodes and provide the data that they need to run.

#### PR Checklist

- [x] Tests updated


pkg/flow/internal/controller/node_config_import.go
```diff
@@ -0,0 +1,40 @@
+package controller
+
+import (
+	"github.com/grafana/river/ast"
+	"github.com/grafana/river/vm"
+)
+
+// TODO: This struct is an empty shell for now, its implementation will come via another PR.
+
+type ImportConfigNode struct {
```
rfratto, line 10: I would ask that we isolate things out more. The ask was for the larger PR to be split into smaller, isolated PRs, so I wouldn't expect to see Import referenced at all here. 

I also don't understand what "ImportConfigNodesChildren" means yet, so I would ask that all code related to import be removed from this PR.

pkg/flow/internal/controller/loader.go
```diff
@@ -743,3 +783,30 @@ func (l *Loader) isModule() bool {
...
+				l.collectCustomComponentReferences(stmt.Body, uniqueReferences)
+			default:
+				if declareNode, ok := l.declareNodes[stmt.Name[0]]; ok {
+					uniqueReferences[declareNode] = struct{}{}
+				} else if importNode, ok := l.importConfigNodes[stmt.Name[0]]; ok {
+					uniqueReferences[importNode] = struct{}{}
+				}
+			}
+		}
+	}
```
rfratto, line 841: nit: idiomatically we try to avoid deep scope nesting like what exists here. Can you rework this a little to avoid the nesting more?

pkg/flow/internal/controller/loader.go
```diff
@@ -507,6 +535,18 @@ func (l *Loader) wireGraphEdges(g *dag.Graph) diag.Diagnostics {
 
 				g.AddEdge(dag.Edge{From: n, To: dep})
 			}
+		case *DeclareNode:
+			// Although they do nothing on evaluation, DeclareNodes are wired
+			// to detect cyclic dependencies. If a declare "a" block contains an instance
+			// of a declare "b" which contains an instance of the declare "a", both DeclareNodes
+			// will depend on each others, creating a cycle in the graph which will be detected later.
```
rfratto, line 550: This comment is fine, but it would be easier to understand with an inline code example; I'd recommend writing one to help illustrate the cyclic dependency you're describing.

pkg/flow/internal/controller/custom_component_registry.go
```diff
@@ -0,0 +1,93 @@
...
+		}
+	}
+
+	for k, v := range s.imports {
+		if v != nil {
+			newScope.imports[k] = v.deepCopy()
+		}
+	}
+	return newScope
+}
```
rfratto, line 93: Why do we need to create a deep copy? Code for producing copies is prone to become stale as new functionality is added, leading to obscure bugs. If possible, I recommend avoiding the need to make copies as much as possible.

pkg/flow/internal/controller/loader.go
```diff
@@ -123,7 +131,7 @@ func NewLoader(opts LoaderOptions) *Loader {
 // to expose values of other components.
 //
 // declareBlocks are pieces of config that can be used as templates to instantiate custom components.
-func (l *Loader) Apply(args map[string]any, componentBlocks []*ast.BlockStmt, configBlocks []*ast.BlockStmt, declareBlocks []*ast.BlockStmt) diag.Diagnostics {
+func (l *Loader) Apply(args map[string]any, componentBlocks []*ast.BlockStmt, configBlocks []*ast.BlockStmt, declareBlocks []*ast.BlockStmt, options LoadOptions) diag.Diagnostics {
```
rfratto, line 135: The parameter list here is getting out of hand. Can you move all of the parameters to LoadOptions?

pkg/flow/internal/controller/loader.go
```diff
@@ -109,6 +109,14 @@ func NewLoader(opts LoaderOptions) *Loader {
 	return l
 }
 
+// LoadOptions are options that can be provided when loading a new River config.
+type LoadOptions struct {
```
rfratto, line 114: nit: this is passed to Apply, so ApplyOptions (or LoaderApplyOptions) would be a more descriptive name.

pkg/flow/module.go
```diff
@@ -58,6 +59,29 @@ func (m *moduleController) NewModule(id string, export component.ExportFunc) (co
...
+
+	mod := newModule(&moduleOptions{
+		ID:                      fullPath,
+		export:                  export,
+		moduleControllerOptions: m.o,
+		parent:                  m,
+	})
+
+	return mod, nil
+}
```
rfratto, line 83: I don't understand why this method is being added? It looks to be exactly the same as NewModule. Can we avoid adding this, please? If you want to rename NewModule to make it more fitting to both cases, please feel free do so.

pkg/flow/internal/controller/node_custom_component.go
```diff
@@ -0,0 +1,325 @@
...
+	// If this is a local declare.
+	importNamespace := ""
+	customComponentName := parts[0]
+	// If this is an imported declare.
+	if len(parts) > 1 {
+		importNamespace = parts[0]
+		customComponentName = parts[1]
+	}
+	return importNamespace, customComponentName
+}
```
rfratto, line 73: My [comment](https://github.com/grafana/agent/pull/5968#discussion_r1450780874) from the original PR (#5968) still stands here: 

> This is making an assumption about the name of the component indicating its origin, and I don't think that'll hold true forever; it's likely that in the near future we'll be asked to support declaring components with . in the name so they feel even more like native components.
>
> Can we make this code more flexible for future-proofing or find a way to avoid it entirely?

That being said, I don't want to spin our wheels on a discussion here and prevent this from getting merged any longer. Can you add a TODO statement for future proofing this?

pkg/flow/internal/controller/node_custom_component.go
```diff
@@ -0,0 +1,325 @@
...
+	}
+}
+
+// CurrentHealth returns the current health of the CustomComponentNode.
+//
+// The health of a CustomComponentNode is determined by combining:
+//
+//  1. Health from the call to Run().
+//  2. Health from the last call to Evaluate().
+//  3. Health reported from the custom component.
```
rfratto, line 283: This comment is stale; health from the custom component isn't used here.

pkg/flow/internal/controller/component_node_manager.go
```diff
@@ -0,0 +1,99 @@
...
+	return &ComponentNodeManager{
+		globals:             globals,
+		builtinComponentReg: componentReg,
+	}
+}
+
+// CreateComponentNode creates a new builtin component or a new custom component.
+func (m *ComponentNodeManager) createComponentNode(componentName string, block *ast.BlockStmt) (ComponentNode, error) {
+	// firstPart may correspond either to a namespace or to a componentName for custom components
+	firstPart := strings.Split(componentName, ".")[0]
```
rfratto, line 34: The componentName argument is set by calling `block.GetBlockName()`, so it seems redundant to split it into parts again. Can this be replaced with `block.Name[0]`?

pkg/flow/internal/controller/loader.go
```diff
@@ -486,39 +499,26 @@ func (l *Loader) populateComponentNodes(g *dag.Graph, componentBlocks []*ast.Blo
...
-				diags.Add(diag.Diagnostic{
-					Severity: diag.SeverityLevelError,
-					Message:  fmt.Sprintf("Unrecognized component name %q", componentName),
-					StartPos: block.NamePos.Position(),
-					EndPos:   block.NamePos.Add(len(componentName) - 1).Position(),
-				})
-				continue
-			}
-
-			if block.Label == "" {
```
rfratto, line 470: Where did this condition for a missing label get moved to? I can't find it anymore.

pkg/flow/declare_test.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+}
+
+type errorTestCase struct {
+	name          string
+	config        string
+	expectedError *regexp.Regexp
+}
+
+func TestDeclareError(t *testing.T) {
```
rfratto, line 234: I'd like to see test cases for at least:

* Ensuring that a custom component can't reference top-level declares, since its being scoped out for now. 

* Ensuring that a nested declare can't access components out of scope:

  ```river
  declare "a" {
    b_1 "default" { } // this should error 
  }
  declare "b" {
    declare "b_1" {}
  }
  ```

* What happens when you declare a component called `declare`? (IMO this should be invalid, and should probably be a reserved name alongside import.)

If you can come up with any other test cases for where declare can/should fail, please add them as well.

pkg/flow/declare_test.go
```diff
@@ -0,0 +1,308 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+type testCase struct {
+	name     string
+	config   string
+	expected int
+}
+
+func TestDeclare(t *testing.T) {
```
rfratto, line 23: I'd like to see test cases for:

* Ensuring that updating the configuration and changing the definition of a declare block works with the new definition. 

* Ensuring that it's possible to shadow builtin components with custom ones without an error.

rfratto (commented):
This is my only round of feedback before approving. I have problems following the code, particularly tracing where custom component registries are created and passed around, as the combined piece of logic spread across a few files. 

But at this point, where we need modules done, I'm not interested in blocking this PR and trying to rearchitect how data is made available to nested controllers. Let's disagree and commit following my last bits of feedback here.

pkg/flow/module.go
wildum: This new function returns a "controller.CustomComponent" while the other one returns a "component.Module". The idea is that the old one could be removed when we get rid of the old modules. I couldn't reuse the component.Module interface because the LoadOptions struct is defined in the controller package (I could use any but I did not like it). Do you have a suggestion to handle this better?

rfratto: I see. That's an artifact of the registry being passed at apply time instead of at construction time. 

Let's just leave it.

pkg/flow/declare_test.go
wildum: > Ensuring that a custom component can't reference top-level declares, since its being scoped out for now.

What do you mean by this?
The following case is allowed:
```
declare "a" {}
declare "b" { a {} }
b {}
```
Or do you mean that it should not be supported to reference the exports of a node declare outside of the declare?

pkg/flow/declare_test.go
rfratto: It will be possible to fully shadow with import. For now, the test can check for shadowing a namespace.

wildum: something like this?
```
declare "a" {}
declare "b" {
  declare "a" {}
  a {} // no error, use the template defined locally (inside of "b")
}
b {}
```

rfratto: That's another case that should work, so that should be a test as well. But no, I was talking about declaring a component called `prometheus` or `local` or the name of another builtin namespace.

wildum: Ok, will test both

pkg/flow/internal/controller/loader.go
wildum: There are a few ways we can go if we put these params in the options.
Currently the functions are as follow:
```
func (c *module) LoadBody(body ast.Body, args map[string]any, options ApplyOptions) error
func (f *Flow) loadSource(source *Source, args map[string]any, options controller.ApplyOptions) error
func (l *Loader) Apply(args map[string]any, componentBlocks []*ast.BlockStmt, configBlocks []*ast.BlockStmt, declareBlocks []*ast.BlockStmt, options LoadOptions) diag.Diagnostics 
```
One way to change it could be:
```
func (c *module) LoadBody(body ast.Body, args map[string]any, customComponentRegistry *CustomComponentRegistry) error
func (f *Flow) loadSource(source *Source, args map[string]any, customComponentRegistry *CustomComponentRegistry)
func (l *Loader) Apply(options ApplyOptions)
```
What do you think?

rfratto: That's a lot better, I'm in favor 👍

pkg/flow/internal/controller/loader.go
wildum: One way to do it could be to add a function "processBlockStmt" that contains the logic 
```
        componentName := strings.Join(stmt.Name, ".")
	switch componentName {
	case "declare":
		l.collectCustomComponentReferences(stmt.Body, uniqueReferences)
	default:
		if declareNode, ok := l.declareNodes[stmt.Name[0]]; ok {
			uniqueReferences[declareNode] = struct{}{}
		} else if importNode, ok := l.importConfigNodes[stmt.Name[0]]; ok {
			uniqueReferences[importNode] = struct{}{}
		}
	}
```
The default part could also be in a function.
Else I don't see a way to put a continue or a return that could help reduce the nesting in a way that would make it more readable. Do you have a suggestion in mind?

rfratto: There's a few ways you can go about this, but if you're interested in having a switch statement, then relying on early continues helps flatten out the hierarchy: 

```go
for _, stmt := range stmts {
  blockStmt, ok := stmt.(*ast.BlockStmt)
  if !ok {
    continue 
  }

  var (
    componentName = strings.Join(blockStmt.Name, ".")

    declareNode, foundDeclare = l.declareNodes[blockStmt.Name[0]] 
    importNode, foundImport = l.importNodes[blockStmt.Name[0]]  
  )

  switch {
  case componentName == "declare":
    l.collectCustomComponentReferences(blockStmt.Body, uniqueReferences) 
  case foundDeclare:
    uniqueReferences[declareNode] = struct{}{}
  case foundImport:
    uniqueReferences[importNode] = struct{}{}
  }
}
```

If you find this to be overkill, I would be happy with just removing the outer switch statement (so that instead of the current `for -> switch -> switch -> if` it's `for -> switch -> if`).

wildum: Looks good to me, I will go forward with this option, thanks :)

pkg/flow/internal/controller/custom_component_registry.go
wildum: My concern is that because we share the registry through all modules, a parent may modify it while the child is accessing it. I'm not sure if the mutex is enough in this case because when it will be released by the parent, the child might not find the content that it was looking for. The change will trigger a reload of the child, but you have this small window where the behavior of the child is undefined (may not find the content or may find a content that does not work with the current state of the child). I remove it now because I think that such situation should occur only with imports (with declare only we would create a new pointer on reload). What do you think?

pkg/flow/internal/controller/node_custom_component.go
wildum: This code is not needed with the declare part. I will bring it back with a TODO when adding the import part in another PR, unless a better idea comes up

pkg/flow/internal/controller/loader.go
wildum: good catch, brought it back and added a test

pkg/flow/declare_test.go
wildum: Thanks for the suggestions, although the code logic matches the expected outcome, the error messages are not ideal for the tests: 
- OutOfScopeReference: error: component "testcomponents.count.inc.count" does not exist
```
       testcomponents.count "inc" {
		frequency = "10ms"
		max = 10
	}

	declare "example_a" {
		testcomponents.summation "sum" {
			input = testcomponents.count.inc.count // invalid reference
		}
	}
	example_a "test" {}
```
- OutOfScopeDefinition: error: unrecognized component name "b_1"
```
	declare "a" {
		b_1 "default" { } // this should error 
	}
	declare "b" {
		declare "b_1" {}
	}
	a "example" {}
```
In both cases, the declare cannot tell the difference between a reference that does not exist or a reference that is out of scope because it does not have access to the full scope. 
I suggest that we either keep it like this, or we change the messages to smth like "component "testcomponents.count.inc.count" does not exist or is out of scope".

rfratto: @wildum the new error message you proposed ("X does not exist or is out of scope") makes sense to me, let's move to that 👍

pkg/flow/internal/controller/custom_component_registry.go
rfratto: Let's leave it for now (unless you've already removed it), I'd like to avoid any larger changes, especially if you think they might be brought right back

rfratto (approved):
Thanks William! Feel free to merge after [this last comment](https://github.com/grafana/agent/pull/6291#discussion_r1481553502) is addressed.

pkg/flow/declare_test.go
rfratto: @wildum my gut feeling for that one is to change up the error message a little bit to something like:

> component name X is not defined in this scope

I'm not sure why this is my gut feeling; I think I want to have the phrasing different to more easily differentiate between errors for invalid references and invalid declarations

wildum: users that only use built-in components and are not aware of custom components may find this message a bit obscure if they just mistyped the name of a component. In this context it's not very clear what "scope" refers to.
Maybe "cannot retrieve the definition of component name X"?

rfratto: `cannot retrieve the definition of component name X` works for me, :shipit:

PR #6296, opened 2024-02-05
fix(otelcol/comp/svcgraph): measure span latency in second instead of milisecond

### LINKED ISSUE #6280: otelcol.connector.servicegraph reporting metrics in milliseconds
### What's wrong?

otelcol.connector.servicegraph is reporting the duration of spans in milliseconds in its exported histogram metrics (`traces_service_graph_request_client_seconds` and `traces_service_graph_request_server_seconds`). This is due to a bug in the upstream otel processor, which has been [fixed in version 0.89](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.89.0/CHANGELOG.md#-bug-fixes-). This does not lead to erroneous data (the buckets are also in milliseconds), but does lead to the service graph showing request durations 1000x higher than their true values.

We should be able to fix this bug by updating the dependency version for `opentelemetry-collector-contrib` in go.mod.

### Steps to reproduce

Initialize `otelcol.connector.servicegraph` with any configuration, inspect exported metrics.

### System information

Affects all systems

### Software version

Affects latest Grafana Agent (v0.39.1)

### Configuration

```text
N/A
```

### Logs

```text
N/A
```

hainenber:

 Fixes #6280

#### PR Checklist

- [x] CHANGELOG.md updated


go.mod
```diff
@@ -171,46 +171,46 @@ require (
...
-	go.opentelemetry.io/collector/config/confighttp v0.87.0
-	go.opentelemetry.io/collector/config/confignet v0.87.0
-	go.opentelemetry.io/collector/config/configopaque v0.87.0
-	go.opentelemetry.io/collector/config/configtelemetry v0.87.0
-	go.opentelemetry.io/collector/config/configtls v0.87.0
-	go.opentelemetry.io/collector/confmap v0.87.0
-	go.opentelemetry.io/collector/connector v0.87.0
-	go.opentelemetry.io/collector/consumer v0.87.0
-	go.opentelemetry.io/collector/exporter v0.87.0
+	go.opentelemetry.io/collector v0.89.0
```
ptodev, line 174: Unfortunately it's not possible to merge this PR, because updating shared OTel dependencies could have a knock-on effect on various OTel components other than the service graph connector. We usually update all OTel dependencies in bulk:
https://github.com/grafana/agent/blob/main/docs/developer/updating-otel.md

Unfortunately it's a very time consuming and error prone process, and maybe not the best candidate for a community contribution :/ I hope the process will be easier in the future.

ptodev:
Thank you for the submission, but unfortunately I need to close it due to [this comment](https://github.com/grafana/agent/pull/6296/files#r1478018402) :/

go.mod
hainenber: Noprob, thanks for clarifying ✌️

PR #6311, opened 2024-02-05
Set mutex and block profiling

mattdurham:

I am explicitly not listing the settings in the docs at this time, since we may figure out the default is good enough and can ignore using them.

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Documentation added

### DIFF
--- a/cmd/internal/flowmode/cmd_run.go
+++ b/cmd/internal/flowmode/cmd_run.go
@@ -466,6 +466,7 @@ func setMutexBlockProfiling(l log.Logger) {
 			runtime.SetMutexProfileFraction(100 / rate)
 		} else {
 			level.Error(l).Log("msg", "error setting MUTEX_PROFILING_PERCENT", "err", err)
+			runtime.SetMutexProfileFraction(1000)
 		}
 	} else {
 		// Why 1000 because that is what istio defaults to and that seemed reasonable to start with.
@@ -478,6 +479,7 @@ func setMutexBlockProfiling(l log.Logger) {
 			runtime.SetBlockProfileRate(100 / rate)
 		} else {
 			level.Error(l).Log("msg", "error setting BLOCK_PROFILING_PERCENT", "err", err)
+			runtime.SetBlockProfileRate(1000)
 		}
 	} else {
 		runtime.SetBlockProfileRate(1000)
--- a/cmd/internal/flowmode/cmd_run.go
+++ b/cmd/internal/flowmode/cmd_run.go
@@ -477,7 +477,7 @@ func setMutexBlockProfiling(l log.Logger) {
 		if err == nil && rate > 0 {
 			runtime.SetBlockProfileRate(100 / rate)
 		} else {
-			level.Error(l).Log("msg", "error setting MUTEX_PROFILING_PERCENT", "err", err)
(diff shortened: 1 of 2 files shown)


cmd/internal/flowmode/cmd_run.go
```diff
@@ -450,3 +455,32 @@ func splitPeers(s, sep string) []string {
...
+		if err == nil && rate > 0 {
+			// The 100/rate is because the value is interpreted as 1/rate. So 50 would be 100/50 = 2 and become 1/2 or 50%.
+			runtime.SetMutexProfileFraction(100 / rate)
+		} else {
+			level.Error(l).Log("msg", "error setting MUTEX_PROFILING_PERCENT", "err", err, "value", mutexProfileFraction)
+			runtime.SetMutexProfileFraction(1000)
+		}
+	} else {
+		// Why 1000 because that is what istio defaults to and that seemed reasonable to start with.
+		runtime.SetMutexProfileFraction(1000)
```
rfratto, line 472: Also, 1000 for a rate of MutexProfileFraction is a sampling rate of 0.1%. Is that enough? (Probably, but just want to be explicit that it's <1%).

cmd/internal/flowmode/cmd_run.go
```diff
@@ -450,3 +455,32 @@ func splitPeers(s, sep string) []string {
 	}
 	return strings.Split(s, sep)
 }
+
+func setMutexBlockProfiling(l log.Logger) {
+	mutexProfileFraction := os.Getenv("MUTEX_PROFILING_PERCENT")
```
rfratto, line 460: nit: feels weird to call these variables *Fraction when they're percentages and made into fractions later.

cmd/internal/flowmode/cmd_run.go
```diff
@@ -450,3 +455,32 @@ func splitPeers(s, sep string) []string {
...
+		}
+	} else {
+		// Why 1000 because that is what istio defaults to and that seemed reasonable to start with.
+		runtime.SetMutexProfileFraction(1000)
+	}
+	blockProfileFraction := os.Getenv("BLOCK_PROFILING_PERCENT")
+	if blockProfileFraction != "" {
+		rate, err := strconv.Atoi(blockProfileFraction)
+		if err == nil && rate > 0 {
+			runtime.SetBlockProfileRate(100 / rate)
```
rfratto, line 478: This doesn't seem right; block profiling rate is _not_ the same as mutex profiling fraction; the docs for SetBlockProfileRate say that the value given to it is nanoeseconds, so here `BLOCK_PROFILING_PERCENT=50` would get block profiles every 2 nanoseconds, which is almost definitely not what the user wants.

rfratto (changes requested):

rfratto:
(To be clear, my "request changes" was _only_ for my [comment](https://github.com/grafana/agent/pull/6311#discussion_r1478854212) about SetBlockProfilingRate; everything else is a nit)

rfratto (approved):
LGTM. If we see that the impact of the defaults are more than we want, we might want to consider changing the defaults to align with Go (i.e., disabled by default). But we won't really know until we try so :shipit:

PR #6313, opened 2024-02-05
Adjust the latency_fanout with some very unscientific values

mattdurham:

The default buckets ended at 10 seconds and the write path can take much longer than that. Chose 300s (5m) as the max since anything above that hits the staleness marker and would be a really bad indicator.

#### PR Checklist

- [x] CHANGELOG.md updatedad

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,7 +51,7 @@ Main (unreleased)
 
 - `service_name` label is inferred from discovery meta labels in `pyroscope.java` (@korniltsev)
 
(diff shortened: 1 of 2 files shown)


rfratto (commented):
Haha I don't want to have to change this again after we find out we need more specific buckets. Can you ask around how we should distribute buckets for a histogram? It would be good knowledge for us to have in general going forward.

(Or is this an opportunity for us to use native histograms, even? 👀)

rfratto:
I'm going to close this for now as stale. If we want to pick this back up, I think we should move the PR to the Alloy repo and then consider whether it should be backported here.

PR #6317, opened 2024-02-06
Add track_timestamps_staleness

### LINKED ISSUE #5921: Add `track_timestamps_staleness` to prometheus.scrape

bboreham:
#### PR Description

Added upstream 4 months ago: https://github.com/prometheus/prometheus/pull/13060

#5921

#### Notes to the Reviewer

I have cherry-picked the commit originally in #5972; I did not yet change the default to true, because I'm unsure where that would go.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated
- NA Config converters updated - this was done previously

### DIFF
--- a/component/prometheus/scrape/scrape.go
+++ b/component/prometheus/scrape/scrape.go
@@ -51,6 +51,8 @@ type Arguments struct {
 	HonorLabels bool `river:"honor_labels,attr,optional"`
 	// Indicator whether the scraped timestamps should be respected.
 	HonorTimestamps bool `river:"honor_timestamps,attr,optional"`
+	// Indicator whether to track the staleness of the scraped timestamps.
+	TrackTimestampsStaleness bool `river:"track_timestamps_staleness,attr,optional"`
 	// A set of query parameters with which the target is scraped.
 	Params url.Values `river:"params,attr,optional"`
 	// Whether to scrape a classic histogram that is also exposed as a native histogram.
@@ -94,13 +96,14 @@ type Arguments struct {
 // SetToDefault implements river.Defaulter.
 func (arg *Arguments) SetToDefault() {
 	*arg = Arguments{
-		MetricsPath:      "/metrics",
-		Scheme:           "http",
-		HonorLabels:      false,
-		HonorTimestamps:  true,
-		HTTPClientConfig: component_config.DefaultHTTPClientConfig,
-		ScrapeInterval:   1 * time.Minute,  // From config.DefaultGlobalConfig
-		ScrapeTimeout:    10 * time.Second, // From config.DefaultGlobalConfig
+		MetricsPath:              "/metrics",
+		Scheme:                   "http",
+		HonorLabels:              false,
+		HonorTimestamps:          true,
+		TrackTimestampsStaleness: false,
+		HTTPClientConfig:         component_config.DefaultHTTPClientConfig,
+		ScrapeInterval:           1 * time.Minute,  // From config.DefaultGlobalConfig
+		ScrapeTimeout:            10 * time.Second, // From config.DefaultGlobalConfig
 	}
 }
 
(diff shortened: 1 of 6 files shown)


component/prometheus/scrape/scrape.go
```diff
@@ -94,13 +96,14 @@ type Arguments struct {
...
-		HonorLabels:      false,
-		HonorTimestamps:  true,
-		HTTPClientConfig: component_config.DefaultHTTPClientConfig,
-		ScrapeInterval:   1 * time.Minute,  // From config.DefaultGlobalConfig
-		ScrapeTimeout:    10 * time.Second, // From config.DefaultGlobalConfig
+		MetricsPath:              "/metrics",
+		Scheme:                   "http",
+		HonorLabels:              false,
+		HonorTimestamps:          true,
+		TrackTimestampsStaleness: false,
```
rfratto, line 103: @bboreham If you want to set the default to true, this is the place to do it.

docs/sources/flow/reference/components/prometheus.scrape.md
```diff
@@ -51,6 +51,7 @@ Name | Type | Description | Default | Required
 `enable_protobuf_negotiation` | `bool`     | Whether to enable protobuf negotiation with the client. | `false` | no
 `honor_labels`                | `bool`     | Indicator whether the scraped metrics should remain unmodified. | `false` | no
 `honor_timestamps`            | `bool`     | Indicator whether the scraped timestamps should be respected. | `true` | no
+`track_timestamps_staleness`  | `bool`     | Indicator whether to track the staleness of the scraped timestamps. | `false` | no
```
rfratto, line 54: @bboreham If you want to change the default to true, this also needs to be updated.

rfratto (approved):
LGTM, left a few comments about how to set it to true by default if that's what you think it should be.

bboreham:
> If you want to set the default to true

Yes I absolutely do, this is the correct setting for most people.

There was a lot of debate at https://github.com/grafana/agent/pull/5972#pullrequestreview-1787355744

bboreham:
However it may be appropriate to merge this PR, which is just allowing the config, then do a separate PR where we start arguing again.

rfratto:
> However it may be appropriate to merge this PR, which is just allowing the config, then do a separate PR where we start arguing again.

Sure, if you think that's easier 👍 merging

clayton-cornell (commented):
Docs are fine as-is

Issue #6003, opened 2023-12-19
Support/embed go x509 fallback roots


jkroepke:
### Request

Go 1.20 and higher provides an fallback for x509 certificate validation, if system pool does not provide the correct certificates. 

This can be enabled by import

```
import _ "golang.org/x/crypto/x509roots/fallback"
```

Ref: https://pkg.go.dev/golang.org/x/crypto/x509roots/fallback

The usage for the fallback root can be enforced with [`GODEBUG=x509usefallbackroots=1`](https://pkg.go.dev/crypto/x509#SetFallbackRoots). 

### Use case

I have some windows servers where I'm unable to use the system verifier to validate ssl connections.

Ref: https://github.com/golang/go/issues/63238

Define an own Truststore is the workaround for the issue above. It sounds great, if go directly maintains an own embedded trust store out of the box.

tpaschalis:
Hello there Jan-Otto 👋  Happy new year!

I'm looking forward to see how the upstream golang/go issue progresses, but wanted to ask if you're still blocked on this, and what the fix could look like. Would it be embedding the x509roots so you can build the Agent using the custom GODEBUG flag?

jkroepke:
Hey!

reading https://go.googlesource.com/go/+/master/src/crypto/x509/root.go?autodive=0%2F%2F#38 ,

```go
import _ "golang.org/x/crypto/x509roots/fallback"
```

Should sufficient to resolve the issue. No more settings are required to expose. `GODEBUG=x509usefallbackroots=1` is a run time variable, not a build-time variable. If I set `GODEBUG=x509usefallbackroots=1` on agent start, the go runtime should not use the system verifiers anymore, instead it would use the embedded certificate chain provided by `x509roots`.

jkroepke:
unstable

tpaschalis:
Hey there Jan-Otto! Since we've merged https://github.com/grafana/agent/pull/6340, I'm closing this as completed. Let us know if there's anything else around this issue!

PR #6340, opened 2024-02-09
feat(cmd/agent): add fallback X.509 trusted roots

hainenber:
#### PR Description

Add fallback X.509 trusted roots to `grafana-agent` and `grafana-agent-flow` binaries. 

Risk is mild to empty as the functionality is only activated via injecting env var [GODEBUG=x509usefallbackroots=1](https://pkg.go.dev/crypto/x509#SetFallbackRoots) to the binary during runtime.

Implements #6003 

#### PR Checklist

- [x] CHANGELOG.md updated


ptodev (approved):
Thanks @hainenber, LGTM. I suppose the main worry with this is that those certificates may no longer be trusted one day. If the users update their Agent binary regularly, this shouldn't be a problem.

Also, apparently the GODEBUG env var [has to be set](https://cs.opensource.google/go/go/+/refs/tags/go1.22.0:src/crypto/x509/root.go;l=51) for any of this to take effect, so I think we can assume that we won't impact users negatively if they don't use the feature.

cc @clayton-cornell it might be good to document interesting uses of GODEBUG and other Go env vars (such as the ones for proxying). Maybe we can add a new section under [Command-line interface](https://grafana.com/docs/agent/latest/flow/reference/cli/)? I could raise an issue for this if you'd like me to.

CHANGELOG.md
```diff
@@ -87,6 +87,8 @@ Main (unreleased)
 
 - Updated docs for MSSQL Integration to show additional authentication capabilities. (@StefanKurek)
 
+- Embed fallback X.509 trusted roots in `grafana-agent` and `grafana-agent-flow` binaries. (@hainenber)
```
ptodev, line 90:
```suggestion
- `grafana-agent` and `grafana-agent-flow` fallback to default X.509 trusted root certificates
  when the `GODEBUG=x509usefallbackroots=1` environment variable is set. (@hainenber)
```

CHANGELOG.md
```diff
@@ -87,6 +87,9 @@ Main (unreleased)
 
 - Updated docs for MSSQL Integration to show additional authentication capabilities. (@StefanKurek)
 
+- `grafana-agent` and `grafana-agent-flow` fallback to default X.509 trusted root certificates
+  when the `GODEBUG=x509usefallbackroots=1` environment variable is set. (@hainenber)
```
clayton-cornell, line 91: This isn't documented anywhere (that I can find). Is this something we should add to the HTTP or TLS docs?

ptodev: In the block configuration we normally only document the arguments which are in the block. I haven't seen documentation about environment variables in block documentation. I think we should make a special page for the env vars and link to it from the block docs.

@jkroepke could you please let us know which Agent components you are using in this case? And why is it not sufficient to use an argument such as `ca_pem` in the [tls config block](https://grafana.com/docs/agent/latest/flow/reference/components/discovery.http/#tls_config-block)?

jkroepke: docs for 

> GODEBUG=x509usefallbackroots=1

https://pkg.go.dev/crypto/x509#SetFallbackRoots

I'm plan to use it for

* module.http
* remote.http
* prometheus.remote_write
* loki.write

> And why is it not sufficient to use an argument such as ca_pem in the [tls config block](https://grafana.com/docs/agent/latest/flow/reference/components/discovery.http/#tls_config-block)?

The `ca_pem` block is sufficient. But then I have to ship the whole ca-certificates pem file together with the setup. Using the fallback X.509 trusted roots from go also fits well and would eliminate the requirement of shipping a own ca pem file.

An other benefit is if the root CAs are embedded:
* grafana-agent could run in a `FROM scratch` docker image which reduce the attack surface to a minimum. Since grafana may has very high permissions in a Kubernetes cluster, reducing the dependencies to a minimum would increase the security as well.

Configure the `ca_pem`, but using the trust ca-certificates sounds silly for first time, but using a custom ca certificates file skips the system verifier. The issue with the system verifier under Windows is that go has no capabilities to configure flags for the Windows Crypto API.

Context: https://github.com/golang/go/issues/63238#issuecomment-1807292976

ptodev: @jkroepke Thanks for clarifying - this also confirms what I inferred from the issue so far.

@clayton-cornell If we are to document such env vars, we would also need to document how they could interfere with arguments in the River config, and which one takes precedence. I could raise a separate issue for this? I don't think it's a must for now. In the past, I've seen this issue mostly with the `HTTP_PROXY` env var.

clayton-cornell: @ptodev I'd say drop it into an issue and backlog it for a future doc improvement. It's not critical right now... it's just something I wondered about as I read through the current change, and it would certainly be helpful to have this added at some point.

PR #6342, opened 2024-02-09
pyroscope.scrape: godeltaprof probing

korniltsev:
#### PR Description

This PR changes pyroscope.scrape behavior to do "godeltaprof probing" at the very beginning of target scrape.
What godeltaprof probing mean is when a component configured to scrape memory/block/mutex, it will attempt to scrape default godeltaprof path for godeltaprof_memory/godeltaprof_block/godeltaprof_mutex and if it succeeds , the scrape loop will keep using godeltaprof paths. If it fails it will fallback to scraping regular profiles.

#### Notes to the Reviewer
Should we have a flag to disable it?
#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated


ptodev:
> Should we have a flag to disable it?

Is there a reason why the user would prefer using regular profiles? Or a reason to avoid godeltaprof? Maybe regular profiles work better for some use cases? Or some customers might experience issues with godeltaprof? Given that godeltaprof is a different technology, I suppose it comes with its own pros and cons.

If we're not absolutely certain that regular profiles are not necessary when godeltaprof works, then IMO it'd be best to make this feature configurable.

korniltsev:
> Is there a reason why the user would prefer using regular profiles?   Or a reason to avoid godeltaprof? 

- A user has not imported godeltaprof library or exposed godeltaprof endpoints available for scraping.
- Some one may not trust/like it ( golang runtime profiler vs less mature fork of golang runtime profiler)


> Or a reason to avoid godeltaprof? Maybe regular profiles work better for some use cases? Or some customers might experience issues with godeltaprof? Given that godeltaprof is a different technology, I suppose it comes with its own pros and cons.

I am not aware of such reasons/issues/cons. As far as we observe it is better both for targets and scrapers. If there are any cons found I'd like to know them, document and fix them.

> If we're not absolutely certain that regular profiles are not necessary when godeltaprof works, then IMO it'd be best to make this feature configurable.

One reason to make it configurable is that for users who has not enabled godeltaprof, they may see a 404 request to `/debug/pprof/delta_heap` and it may be confusing.

ptodev:
I see, thank you. I think it makes sense to have functionality which tries one endpoint, and if it doesn't work, tries another. But this needs to be made clear in the docs, so that the users are not confused why the Agent is not scraping the endpoint that they thought it would do.

WDYT about making the `path` argument inside blocks such as [profile.block](https://grafana.com/docs/agent/latest/flow/reference/components/pyroscope.scrape/#profileblock-block) a slice of strings? By default we could list two endpoints there, where the first one is for godeltaprof. The rule could be that first endpoint in the slice that returns a sensible result will be the one that `pyroscope.scrape` uses?

korniltsev:
I dont like the code I submited. It complicates the code. It complicates the documentation. Lets not merge it. Sorry for confusion.

PR #6343, opened 2024-02-09
POC: hostmetrics connector

jcreixell:
- Host metrics connector reads resource attributes from traces (and potentially other telemetry signals) and extracts the host information based on a list of predefined attributes, listed in order of priority.
  - It includes an otel collector connector and a wrapping flow component.

### DIFF
--- a/component/otelcol/connector/spanhostmetrics/spanhostmetricsconnector.go
+++ b/component/otelcol/connector/spanhostmetrics/spanhostmetricsconnector.go
@@ -1,0 +1,126 @@
+package spanhostmetrics
+
+import (
+	"context"
+	"time"
+
+	"go.uber.org/zap"
+
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/connector"
+	"go.opentelemetry.io/collector/consumer"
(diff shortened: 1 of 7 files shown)


component/otelcol/connector/spanhostmetrics/spanhostmetricsconnector.go
```diff
@@ -0,0 +1,126 @@
...
+						dpCalls.SetIntValue(int64(1))
+
+						// TODO: store the metric in a map and flush it after a certain time instead of on every span
+						err := c.metricsConsumer.ConsumeMetrics(ctx, metrics)
+						if err != nil {
+							return err
+						}
+						break
+					}
+				}
```
jcreixell, line 121: we probably need some telemetry to detect traces with no host info (or at least some logging)

rlankfo:
Awesome, thanks for providing this. One thing to note is this connector could be expanded in the future to include other signals like logs and traces. This makes me think we shouldn't tie the name to a certain signal.

jcreixell:
@rlankfo good point, I thought we could have a separate metric per signal, but the name of the connector should be generic :+1:

jcreixell:
Superseded by #6410

PR #6345, opened 2024-02-09
discovery.process: Accept failing username detection

simonswine:
Quite common with docker container UIDs not resolving on the host or in  another container.

This will basically move on in those cases, rather than dropping the whole process.

Unsure if this should also be a lower level log message, potentially only debug.

### DIFF
--- a/component/discovery/process/discover.go
+++ b/component/discovery/process/discover.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"os/user"
 	"path"
 	"runtime"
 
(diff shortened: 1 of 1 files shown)


ptodev:
> Unsure if this should also be a lower level log message, potentially only debug.

Maybe a warning? I'm actually surprised that if any of those fails, the discovery for that process as a whole gives up. I thought it'd be all best effort? It'd be good to document this in the [discover_config block docs](https://grafana.com/docs/agent/next/flow/reference/components/discovery.process/#discover_config-block) too.

component/discovery/process/discover.go
```diff
@@ -118,7 +118,6 @@ func discover(l log.Logger, cfg *DiscoverConfig) ([]process, error) {
 			username, err = p.Username()
 			if err != nil {
 				loge(int(p.Pid), err)
-				continue
```
ptodev, line 121: Does this mean that a `__meta_process_username` label will still be added, but with a value of `""`?

simonswine: No the convertProcess has filters for the null values

https://github.com/grafana/agent/blob/1f3943dddbf9f1b8a4146bcd95f88be8aadbfa38/component/discovery/process/discover.go#L67

simonswine:
> Maybe a warning? I'm actually surprised that if any of those fails, the discovery for that process as a whole gives up. I thought it'd be all best effort? It'd be good to document this in the [discover_config block docs](https://grafana.com/docs/agent/next/flow/reference/components/discovery.process/#discover_config-block) too.

@tpaschalis In the latest revision I am actually filtering this error all together. 

We need to continue for missing executable, cwd, and cmdline, as this is a sign that we have a kernel level process.

ptodev (approved):

PR #6347, opened 2024-02-09
Fix goroutine leak in local.file

### LINKED ISSUE #6341: Memory leak if initial eval process fail for some components
### What's wrong?

local.file is starting some go routines on creation that are cleaned up when Run is canceled. The problem is that if an error happens in between (for example if the evaluation process fails), the goroutines are left hanging.
Other components might have the same pattern. This is not a big problem in general because the process will exit and the OS will clean up the leaky goroutines but for testing it's a bit annoying.

### Steps to reproduce

Run a test with a config that has a local.file component and another component that depends on it. The other component should have a wrong config that will fail the evaluation process.
Use verifyNoGoroutineLeaks at the end of the test to see the leaky go routines

### System information

_No response_

### Software version

v39.2

### Configuration

_No response_

### Logs

_No response_

wildum:
#### PR Description

Call UpdateArgs instead of Update in the New() function to avoid starting goroutines before Run() is called

Fixes grafana/alloy#247 

#### PR Checklist

- [ ] CHANGELOG.md updated

### DIFF
--- a/component/local/file/file.go
+++ b/component/local/file/file.go
@@ -107,7 +107,8 @@ func New(o component.Options, args Arguments) (*Component, error) {
 	}
 	// Perform an update which will immediately set our exports to the initial
 	// contents of the file.
-	if err = c.Update(args); err != nil {
+	// UpdateArgs instead of Update to prevent starting goroutines if Run() is not called.
+	if err = c.UpdateArgs(args); err != nil {
 		return nil, err
 	}
 	return c, nil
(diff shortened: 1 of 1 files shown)


rfratto (commented):
Thanks for taking a crack at this! I find this a little hard to reason about, as now I have to track down to make sure no fields set in Update are needed at construction time, and that they're all set once Run is fully in motion.

Instead of creating two updates, can we try to prevent the goroutine from being started until Run is called?

component/local/file/file.go
```diff
@@ -200,6 +200,16 @@ func (c *Component) Update(args component.Arguments) error {
...
+	return nil
+}
+
+// Update implements component.Component.
+func (c *Component) Update(args component.Arguments) error {
+
+	err := c.UpdateArgs(args)
+	if err != nil {
+		return err
+	}
```
ptodev, line 212: I think `Run()` could be called without calling `Update()`. In that case, is the fact that we didn't call `c.configureDetector()` prior to `Run()` a problem?

PR #6348, opened 2024-02-09
Allow the Host interface to return services.

tpaschalis:
#### PR Description
This PR enables the Host interface to access other running services. Currently, this simplifies the handling around the UI service and the cluster.Cluster interface it calls to, but in the future this can also help the remotecfg service populate some UI interface.

No issue filed.

#### Notes to the Reviewer
cc @rfratto from a TODO comment as he's already thought about this in the past.

#### PR Checklist

- [ ] Tests updated (N/A)

### DIFF
--- a/pkg/flow/flow_services.go
+++ b/pkg/flow/flow_services.go
@@ -26,6 +26,16 @@ func (f *Flow) GetServiceConsumers(serviceName string) []service.Consumer {
 	return consumers
 }
 
+func (f *Flow) GetService(name string) (service.Service, bool) {
+	for _, svc := range f.opts.Services {
+		if svc.Definition().Name == name {
+			return svc, true
+		}
+	}
+	return nil, false
+
+}
+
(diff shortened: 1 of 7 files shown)


pkg/flow/flow_services.go
```diff
@@ -26,6 +26,15 @@ func (f *Flow) GetServiceConsumers(serviceName string) []service.Consumer {
 	return consumers
 }
 
+func (f *Flow) GetService(name string) (service.Service, bool) {
```
rfratto, line 29: Godoc?

service/service.go
```diff
@@ -37,6 +37,12 @@ type Definition struct {
 	DependsOn []string
 }
 
+// A Provider exposes information around services.
+type Provider interface {
+	// GetService looks up a running service using its name.
+	GetService(name string) (Service, bool)
+}
```
rfratto, line 44: I'm guessing we need this and can't just use service.Host?

rfratto (approved):
Thanks!

PR #6350, opened 2024-02-09
Import - new modules

wildum:
#### PR Description

This PR adds the import functionality of the new modules with import.string and import.file.

The functionality is described in #5547 

The change introduces a new type of node: ImportConfigNode.
It imports modules from an ImportSource to retrieve declare and import blocks.
For every import block, it creates an ImportConfigNode child. The child is evaluated and runs inside of its parent.
When the ImportSource retrieves new content, the ImportConfigNode updates its declares and import node children. Then it propagates an update to the root ImportConfigNode which will notify the controller that it has been updated.
The loader will then update the CustomComponentRegistry with new content and will trigger the dependants of the ImportConfigNode for reevaluation.

#### Notes to the Reviewer

The PR can be reviewed commit by commit.

#### PR Checklist

- [x] Tests updated


pkg/flow/internal/controller/node_config_import.go
```diff
@@ -0,0 +1,328 @@
...
+	// trigger to stop previous children from running and to start running the new ones.
+	if cn.importChildrenRunning {
+		cn.importChildrenUpdateChan <- struct{}{}
+	}
+
+	cn.OnBlockNodeUpdate(cn)
+}
+
+// processImportedContent processes declare and import blocks of the provided ast content.
+func (cn *ImportConfigNode) processImportedContent(content *ast.File) {
```
rfratto, line 143: similar nit from the last PR, can we make this nesting a little less deep?

pkg/flow/internal/controller/node_config_import.go
```diff
@@ -0,0 +1,328 @@
...
+	// evaluate the importConfigNodesChildren that have been created
+	err = cn.evaluateChildren()
+	if err != nil {
+		level.Error(cn.logger).Log("msg", "failed to evaluate nested import", "err", err)
+		return
+	}
+
+	// trigger to stop previous children from running and to start running the new ones.
+	if cn.importChildrenRunning {
+		cn.importChildrenUpdateChan <- struct{}{}
```
rfratto, line 136: Should this handle the case where the channel is already full?

```go
select {
case cn.importChildrenUpdateChan <- struct{}{}: // queued trigger
default: // trigger already queued; no-op 
}
```

(This only applies if you decide to buffer importChildrenUpdateChan, which it isn't right now)

pkg/flow/internal/controller/node_config_import.go
```diff
@@ -0,0 +1,328 @@
...
+
+	cn := &ImportConfigNode{
+		nodeID:                   nodeID,
+		globalID:                 globalID,
+		label:                    block.Label,
+		componentName:            block.GetBlockName(),
+		globals:                  globals,
+		block:                    block,
+		OnBlockNodeUpdate:        globals.OnBlockNodeUpdate,
+		importChildrenUpdateChan: make(chan struct{}),
```
rfratto, line 73: Should this be a buffered channel so writes don't block? `make(chan struct{}, 1)`

pkg/flow/internal/controller/node_config_import.go
```diff
@@ -0,0 +1,328 @@
...
+	}
+	cn.importedDeclares[stmt.Label] = stmt.Body
+}
+
+// processDeclareBlock creates an ImportConfigNode child from the provided import block.
+func (cn *ImportConfigNode) processImportBlock(stmt *ast.BlockStmt, fullName string) {
+	sourceType := importsource.GetSourceType(fullName)
+	if _, ok := cn.importConfigNodesChildren[stmt.Label]; ok {
+		level.Error(cn.logger).Log("msg", "import block redefined", "name", stmt.Label)
+		return
```
rfratto, line 176: Should this error be raised up higher? Should this cause something else to fail rather than being constrained to logs?

pkg/flow/internal/controller/node_config_import.go
```diff
@@ -0,0 +1,328 @@
...
+	newCtx, cancel := context.WithCancel(ctx)
+	defer cancel() // This will stop the children and the managed source.
+
+	errChan := make(chan error, 1)
+
+	if importChildrenCount > 0 {
+		go func() {
+			errChan <- cn.runChildren(newCtx)
+		}()
+	}
```
rfratto, line 230: There seems to be a potential bug here: if the length of `cn.importConfigNodesChildren` is zero when Run is first called, but then is updated to be non-zero afterwards, the `cn.runChildren` goroutine won't be running to trigger on the write to `cn.importChildrenUpdateChan` to run the newly discovered children.

pkg/flow/internal/controller/node_config_import.go
```diff
@@ -0,0 +1,328 @@
...
+	importChildrenCount := len(cn.importConfigNodesChildren)
+	cn.mut.Unlock()
+	if cn.source == nil {
+		return ErrUnevaluated
+	}
+
+	newCtx, cancel := context.WithCancel(ctx)
+	defer cancel() // This will stop the children and the managed source.
+
+	errChan := make(chan error, 1)
```
rfratto, line 224: tip: use [run.Group](https://pkg.go.dev/github.com/oklog/run#Group) here to launch multiple goroutines which can fail; there's examples of this being done elsewhere in our code.

pkg/flow/internal/controller/node_config_import.go
```diff
@@ -0,0 +1,328 @@
...
+
+	if err != nil {
+		level.Error(cn.logger).Log("msg", "import exited with error", "err", err)
+	} else {
+		level.Info(cn.logger).Log("msg", "import exited")
+	}
+	return err
+}
+
+// runChildren run the import nodes managed by this import node.
```
rfratto, line 246: Is it possible to replace the implementation here to use [our runner package](https://github.com/grafana/agent/blob/main/pkg/runner/runner.go)? I've noticed a few bugs in the code, but rather than point them out and check the rest of the code, I think it makes sense to align on the existing concept we have for managing a dynamic set of things to run.

pkg/flow/internal/controller/metrics.go
```diff
@@ -109,6 +109,7 @@ func newControllerCollector(l *Loader, id string) *controllerCollector {
 func (cc *controllerCollector) Collect(ch chan<- prometheus.Metric) {
 	componentsByHealth := make(map[string]int)
 
+	// Should we also collect metrics from components running in import config nodes?
```
rfratto, line 112: Yes 😛

pkg/flow/internal/controller/node_custom_component.go
```diff
@@ -54,6 +58,29 @@ type CustomComponentNode struct {
...
 
+// ExtractImportAndDeclare extracts an importNamespace and a customComponentName from a componentName.
+// TODO: this function assumes that CustomComponentNames and ImportNamespaces don't contain "."
+// In other words, there are two possible scenarios:
+// - [customComponentName] LABEL ->  instance of a local declare
+// - [importNamespace].[customComponentName] LABEL -> instance of an imported declare
+// To address this limitation in the future, we will need to find a different approach to
+// identify the importNamespaces and CustomComponentNames.
+func ExtractImportAndDeclare(componentName string) (string, string) {
+	parts := strings.Split(componentName, ".")
```
rfratto, line 69: At least for now, this will be a little more prepared for the possibility of future changes to custom components if this is changed to use `strings.SplitN(componentName, ".", 2)`, that way only the first `.` is considered for the import namespace, and other `.` can be a part of the declare name.

pkg/flow/import_test.go
```diff
@@ -0,0 +1,769 @@
+package flow_test
```
rfratto, line 1: The River code in the tests don't render very nicely on GitHub (the indentation is inconsistent); can you format them so they're easier to follow?

rfratto, line 1: This file is very big. You might want to consider using testdata and [txttar](https://pkg.go.dev/golang.org/x/tools/txtar) as a way to better organize this. For other testing tips, I'd recommend watching the talk [Go Testing By Example](https://research.swtch.com/testing).

pkg/flow/import_test.go
```diff
@@ -0,0 +1,769 @@
...
+			// Check for initial condition
+			require.Eventually(t, func() bool {
+				export := getExport[testcomponents.SummationExports](t, ctrl, "", "testcomponents.summation.sum")
+				return export.LastAdded == 10
+			}, 3*time.Second, 10*time.Millisecond)
+		})
+	}
+}
+
+func TestImport(t *testing.T) {
```
rfratto, line 91: I'm interested in seeing a test case for a deeply nested import where the definition of the import changes and is used to propagate something up to the root controller; I can't tell if you already have a test case that does something like this.

rfratto (commented):
Thanks William! I don't have any huge comments, just a few nits, questions, and recommendations for helping making the tests shine.

pkg/flow/internal/controller/node_config_import.go
wildum: thanks for catching it, I use the runner pkg now

pkg/flow/import_test.go
wildum: I had some tests but it was unreadable. I refactored it with txtar and it should now be much clearer. I do have tests where deeply nested modules are being updated. Let me know what you think about the new structure of the tests and there are more cases that you would like me to cover

pkg/flow/internal/controller/node_config_import.go
```diff
@@ -0,0 +1,425 @@
...
+
+type childRunner struct {
+	node    *ImportConfigNode
+	errChan chan error
+}
+
+func (cr *childRunner) Run(ctx context.Context) {
+	err := cr.node.Run(ctx)
+	if err != nil {
+		cr.errChan <- err
```
rfratto, line 405: I'm a little worried about the possibility of a goroutine leak here; if multiple children have fatal errors, then the third write will block forever here (the first write will be read in the `run` method; the second write will be written to the buffer; the third write will block because the buffer is full).

Do we need to be sending the failed signal back to the caller, or can we do something else here, such as logging errors or setting the health of the overall node?

pkg/flow/internal/controller/node_config_import.go
```diff
@@ -0,0 +1,425 @@
...
+func (cn *ImportConfigNode) run(errChan chan error, updateTasks func() error) error {
+	for {
+		select {
+		case <-cn.importChildrenUpdateChan:
+			err := updateTasks()
+			if err != nil {
+				return err
+			}
+		case err := <-errChan:
+			return err
```
rfratto, line 370: The Flow controller expects that the `Run` method only exits in two scenarios:

1. The provided context is cancelled (in which case no error should be returned) 
2. A fatal unrecoverable error occurs, where calling `Run` again would not resolve it. 

Because of these assumptions, the controller will never call Run again for the same node once it returns. Returning early here is signaling a fatal error, but I'm not sure that it really is one for all writes to run. 

Can you put more thought into the failure modes and try to reduce the blast radius of failure here?

rfratto (commented):
Partial review; I wanted to leave a comment on something I noticed, but I still need to spend more time doing a full review of the current state.

pkg/flow/import_test.go
```diff
@@ -0,0 +1,362 @@
...
+	_ "github.com/grafana/agent/component/module/string"
+)
+
+// updateFile is used to change the content of a file used by a module
+type updateFile struct {
+	name         string   // name of the file which should be updated (module | nested_module)
+	updateConfig []string // new module config which should be used
+}
+
+func TestImport(t *testing.T) {
```
rfratto, line 27: Nice! This is much easier to follow than before. I have some light feedback though:

As it is now, you have to modify two files (the txtar file and this file) to add a new test case, which can be slightly burdensome. Have you considered an approach similar to what the converters test use, where the test automatically loads files in a directory, and uses individual files (or pairs of files) as the test cases? 

If you move to multiple files, you can use some conventions for defining new test cases. For example:


```txtar
This file tests import passthrough module. 

-- main.river -- 
testcomponents.count "inc" {
	frequency = "10ms"
	max = 10
}

import.file "testImport" {
	filename = "module.river"
}

testImport.a "cc" {
	input = testcomponents.count.inc.count
}

testcomponents.summation "sum" {
	input = testImport.a.cc.output
}

-- module.river -- 
declare "a" {
	argument "input" {}

	testcomponents.passthrough "pt" {
		input = argument.input.value
		lag = "1ms"
	}

	export "output" {
		value = testcomponents.passthrough.pt.output
	}
}

-- update/module.river -- 
declare "a" {
	argument "input" {}

	export "output" {
		value = -argument.input.value
	}
}
```

In the example above, the following conventions are used:

* The description of the txtar explains the test and what it's doing. 
* The primary configuration file to load is always called `main.river`. 
* Other files in the same directory are unpacked on disk in a temporary directory (so `main.river` and `module.river` are unpacked together) where the module is ran.
* Files in the `update/` directory overwrite any existing files on demand; such that when you want to update files, `update/module.river` overwrites `module.river`. 

I don't know if these conventions make the most sense for these tests (I only came up with the first example that came to mind), but it just serves as an example for some improvements that can be made. The main point of my feedback is to try to find a way to create new test cases in a single place without having to touch multiple files.

pkg/flow/import_test.txtar
```diff
@@ -0,0 +1,283 @@
+-- declare_passthrough --
```
rfratto, line 1: nit: typically files like these belong in a `testdata/` subdirectory (relative to the package being tested) to clearly indicate its assets only used in testing.

pkg/flow/internal/controller/component_node_manager.go
```diff
@@ -43,15 +43,25 @@ func (m *ComponentNodeManager) createComponentNode(componentName string, block *
...
+
+	if namespace == "" {
+		template, customComponentRegistry = findLocalDeclare(m.customComponentReg, componentName)
+	} else {
+		template, customComponentRegistry = findImportedDeclare(m.customComponentReg, namespace, componentName)
+	}
 
 	if customComponentRegistry == nil || template == nil {
-		return nil, nil, fmt.Errorf("custom component config not found in the registry, componentName: %s", componentName)
+		return nil, nil, fmt.Errorf("custom component config not found in the registry, namespace: %s, componentName: %s", namespace, componentName)
```
rfratto, line 62: nit: I encourage using `%q` over `%s` for strings wherever possible; `%q` produces a quoted string and makes it much easier to notice when an empty string is being printed. 

(sorry I missed this before)

pkg/flow/internal/controller/custom_component_registry.go
```diff
@@ -29,3 +34,40 @@ func (s *CustomComponentRegistry) registerDeclare(declare *ast.BlockStmt) {
...
+}
+
+// updateImportContentChildren recurse through the children of an import node
+// and update their scope with the imported declare blocks.
+func (s *CustomComponentRegistry) updateImportContentChildren(importNode *ImportConfigNode) {
+	for _, child := range importNode.ImportConfigNodesChildren() {
+		childScope := NewCustomComponentRegistry(nil)
+		childScope.declares = child.ImportedDeclares()
+		childScope.updateImportContentChildren(child)
+		s.imports[child.label] = childScope
```
rfratto, line 71: Does `s.mut` need to be held here?

pkg/flow/internal/controller/node_config_import.go
```diff
@@ -0,0 +1,425 @@
...
+// TODO: is nodeID enough?
+func (cn *ImportConfigNode) Equals(other runner.Task) bool {
+	otherTask := other.(*ImportConfigNode)
+
+	if cn == otherTask {
+		return true
+	}
+
+	return cn.NodeID() == otherTask.NodeID()
+}
```
rfratto, line 425: Yes; nodeID should be enough in both cases here. `Hash` and `Equals` are used to find identical runner tasks; since node IDs are unique per graph, it is safe to use it for hashing and equality checking here. 

(Also, the runner package handles hash collision gracefully, so it's OK if the hash is imperfect)

pkg/flow/internal/controller/node_custom_component.go
```diff
@@ -54,6 +58,29 @@ type CustomComponentNode struct {
...
+	// If this is a local declare.
+	importNamespace := ""
+	customComponentName := parts[0]
+	// If this is an imported declare.
+	if len(parts) > 1 {
+		importNamespace = parts[0]
+		customComponentName = parts[1]
+	}
+	return importNamespace, customComponentName
+}
```
rfratto, line 82: In cases where a function returns two values of the same type, I'd recommend either naming the return values so it's obvious to callers which value is which or creating a struct which names the values. 

Here's an example of the first case (also introducing a switch statement to maybe make things more clear):

```go
func ExtractImportAndDeclare(componentName string) (importNamespace, customComponentName string) {
  parts := strings.SplitN(componentName, ".", 2)
  switch len(parts) {
  case 1: // [customComponentName] 
    customComponentName = parts[0] 
  case 2: // [importNamespace].[customComponentName] 
    importNamespace = parts[0]
    customComponentName = parts[1] 
  }

  return importNamespace, customComponentName 
}
```

rfratto (approved):
Thanks William! I left some final comments. All my comments here in this review are nits, but my feedback from before about the `errChan` is a little more firm and I'd like to ideally see that resolved at least before merging.

pkg/flow/internal/controller/custom_component_registry.go
wildum: updateImportContentChildren is meant to be only called by updateImportContent which is already holding the mut

pkg/flow/internal/controller/node_config_import.go
wildum: There is actually a problem with this:
- root imports "module" that imports "nested_module_1"
- "module" is updated to import "nested_module_2" instead, but the nodeID of the import config block is unchanged
- the runner pkg will think that the exact same task is already running (because the nodeID is the same) and will ignore the task

=> the runner will keep running the old child and the new child won't run

In our case I believe that we don't want this optimization, we always want to stop the previous workers and start new ones because we are creating new nodes.

the easy way to do this would be to have Equals return false (hash doesn't matter anyway because if there is a collision it will check with Equals)

I'm not worried about performance because module updates are very rare (there are checks to make sure that if the imported content is the same we don't do anything, that means that we would need a real change to update the children)

wdyt?

wildum: (we also don't have the risk of running twice the same task because we check that the imported blocks are not duplicated)

pkg/flow/internal/controller/node_custom_component.go
wildum: Nice, I did not know this was possible in Go, that works really well in this case, thanks

pkg/flow/import_test.go
wildum: tha's a good suggestion. I initially used one file to reduce redundancy and allow modules to be reused and concatenated to form new usecases but I agree that it's not the best when it comes to add new tests and we also end up with a big file that is hard to navigate. Your approach makes it much easier to follow, I will make the changes

pkg/flow/internal/controller/node_config_import.go
rfratto: @wildum Normally we'd expect nodes to be dynamically updated instead of recycled/replaced, but let's change `Equals` to be a more basic pointer check for now and we can revisit this later once the refactorings for the controller start coming in.

PR #6355, opened 2024-02-12
Staleness batch

mattdurham:
#### PR Description

This moves the staleness tracker to use batch operation. Using the two fanout options should ensure that tracking is always called.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated


service/labelstore/service_test.go
thampiotr, line 1: Can we write some benchmarks and share the results?

service/labelstore/service.go
```diff
@@ -196,26 +198,23 @@ func (s *service) GetLocalRefID(componentID string, globalRefID uint64) uint64 {
...
 
-	s.staleGlobals[globalRefID] = &staleMarker{
-		lastMarkedStale: time.Now(),
-		labelHash:       l.Hash(),
-		globalID:        globalRefID,
+	for _, id := range ids {
+		if value.IsStaleNaN(id.Value) {
+			s.staleGlobals[id.GlobalRefID] = &staleMarker{
+				lastMarkedStale: time.Now(),
+				labelHash:       id.Labels.Hash(),
```
thampiotr, line 209: Let's make sure we don't call `id.Labels.Hash()` inside the critical section. Something like this:
```
func (s *service) TrackStaleness(ids []StalenessTracker) {
	var (
		toAdd    = make([]*staleMarker, 0)
		toRemove = make([]uint64, 0)
		now      = time.Now()
	)

	for _, id := range ids {
		if value.IsStaleNaN(id.Value) {
			toAdd = append(toAdd, &staleMarker{
				globalID:        id.GlobalRefID,
				lastMarkedStale: now,
				labelHash:       id.Labels.Hash(),
			})
		} else {
			toRemove = append(toRemove, id.GlobalRefID)
		}
	}

	s.mut.Lock()
	defer s.mut.Unlock()

	for _, marker := range toAdd {
		s.staleGlobals[marker.globalID] = marker
	}
	for _, id := range toRemove {
		delete(s.staleGlobals, id)
	}
}
```

component/prometheus/interceptor.go
```diff
@@ -102,13 +103,11 @@ func (a *interceptappender) Append(ref storage.SeriesRef, l labels.Labels, t int
...
 		ref = storage.SeriesRef(a.ls.GetOrAddGlobalRefID(l))
 	}
-
-	if value.IsStaleNaN(v) {
-		a.ls.AddStaleMarker(uint64(ref), l)
-	} else {
-		// Tested this to ensure it had no cpu impact, since it is called so often.
-		a.ls.RemoveStaleMarker(uint64(ref))
-	}
+	a.stalenessTrackers = append(a.stalenessTrackers, labelstore.StalenessTracker{
```
thampiotr, line 106: We create a `labelstore.StalenessTracker` here, but later in `func (s *service) TrackStaleness(ids []StalenessTracker)` we convert these to `&staleMarker{}` and calculate the labels hash. Could we instead create the `&staleMarker{}` type right away here and calculate the hash here? 

That way there will be less work and structs to create, the code will get simpler too I think.

service/labelstore/service.go
```diff
@@ -196,26 +198,23 @@ func (s *service) GetLocalRefID(componentID string, globalRefID uint64) uint64 {
...
 	s.mut.Lock()
 	defer s.mut.Unlock()
 
-	s.staleGlobals[globalRefID] = &staleMarker{
-		lastMarkedStale: time.Now(),
-		labelHash:       l.Hash(),
-		globalID:        globalRefID,
+	for _, id := range ids {
+		if value.IsStaleNaN(id.Value) {
+			s.staleGlobals[id.GlobalRefID] = &staleMarker{
```
thampiotr, line 207: Here we have a map of pointers, but in the fanout we use a slice of structs `stalenessTrackers []labelstore.StalenessTracker`. We also use slice of structs for labels. So it's a bit inconsistent and I'm not sure what's the thinking behind it. Do we have benchmarks for what performs better? Did this come up in allocation profiles somewhere?

service/labelstore/service.go
```diff
@@ -196,26 +198,23 @@ func (s *service) GetLocalRefID(componentID string, globalRefID uint64) uint64 {
 	return local
 }
 
-// AddStaleMarker adds a stale marker
-func (s *service) AddStaleMarker(globalRefID uint64, l labels.Labels) {
+func (s *service) TrackStaleness(ids []StalenessTracker) {
 	s.mut.Lock()
 	defer s.mut.Unlock()
```
thampiotr, line 203: Just a thought: could we use a different mutex for `s.staleGlobals` and thus avoid contention with other methods like GetLocalRefID, GetGlobalRefID, GetOrAddGlobalRefID, etc? seems possible at a glance

thampiotr (commented):
Neat idea to batch it! Would really like to see some benchmarks so we can know for sure how much better this is :)

component/prometheus/fanout.go
```diff
@@ -110,12 +111,11 @@ func (a *appender) Append(ref storage.SeriesRef, l labels.Labels, t int64, v flo
 	if ref == 0 {
 		ref = storage.SeriesRef(a.ls.GetOrAddGlobalRefID(l))
 	}
-	if value.IsStaleNaN(v) {
-		a.ls.AddStaleMarker(uint64(ref), l)
-	} else {
-		// Tested this to ensure it had no cpu impact, since it is called so often.
-		a.ls.RemoveStaleMarker(uint64(ref))
-	}
+	a.stalenessTrackers = append(a.stalenessTrackers, labelstore.StalenessTracker{
```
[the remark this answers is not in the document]
mattdurham: We do but it never had it so keeping the pr small. I will have 2-3 more prs incoming.

service/labelstore/service.go
mattdurham: Sure that makes sense.

component/prometheus/interceptor.go
mattdurham: I feel those are owned by two different things. Mainly the last marked state should really only be set by the labelstore itself and exposing that field feels off. This gets cleaned up slightly in the next PR.

service/labelstore/service.go
mattdurham: You dont have the traverse the map to check if you need to add or update the value like you would an array. Now its unlikely you would need to add staleness markers mutliple times, but I would rather not assume that.

service/labelstore/service.go
mattdurham: I fiddled with that a bit but it felt a bit more error prone in developing. I would prefer to split that out to another pr if we wanted to go that route.

service/labelstore/service_test.go
mattdurham: Sure, added a benchmark. TrackStaleness with concurrent calls averages around **2.057315 ms** per handling 100k entries. 

Backporting the benchmark to main takes **14ms** per 100k entries. So 7x better.

thampiotr (approved):

PR #6364, opened 2024-02-13
Fix flaky test metrics pkg

wildum:
This test would sometimes not pass on windows, making our main build fail (see https://drone.grafana.net/grafana/agent/16444/2/2).

When testing I also noticed that when running it many times on Mac it would also fail.

The fix makes sure that the context is properly canceled before exiting the test

### DIFF
--- a/pkg/metrics/instance/instance_integration_test.go
+++ b/pkg/metrics/instance/instance_integration_test.go
@@ -193,9 +193,14 @@ remote_write: []
 	require.NoError(t, err)
 
 	instCtx, cancel := context.WithCancel(context.Background())
-	defer cancel()
+	done := make(chan struct{})
+	defer func() {
+		cancel()
+		<-done
+	}()
 	go func() {
 		err := inst.Run(instCtx)
+		close(done)
 		require.NoError(t, err)
 	}()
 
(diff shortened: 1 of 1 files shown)


mattdurham (approved):
lgtm

pkg/metrics/instance/instance_integration_test.go
```diff
@@ -73,9 +73,14 @@ remote_write: []
 	require.NoError(t, err)
 
 	instCtx, cancel := context.WithCancel(context.Background())
-	defer cancel()
+	done := make(chan struct{})
+	defer func() {
+		cancel()
+		<-done
+	}()
```
rfratto, line 80: A slightly more idiomatic way of doing this is with [sync.WaitGroup](https://pkg.go.dev/sync#WaitGroup):

```go 
var wg sync.WaitGroup
defer func() {
  cancel() 
  wg.Wait() 
}

wg.Add(1) // Do this before launching the goroutine 
go func() {
  defer wg.Done() 
  // business logic here 
}
```

It tends to be slightly more code than what you have (because you have to explicitly add deltas to the waitgroup), but it doesn't have the risk of closing the same channel twice, and it works across multiple goroutines.

What you have now is perfectly fine, but I just wanted to give you a heads up about sync.WaitGroup :)

rfratto (approved):
Thanks William!

pkg/metrics/instance/instance_integration_test.go
wildum: Thanks, I made the change :)

PR #6367, opened 2024-02-14
pyroscope.scrape: Standarize profile names

simonswine:
When a user adopts godeltaprof_, currently the resulting profile type
name is shown separately from the default ones. That makes is harder
than it should be to adopt godeltaprof.

I am proposing that we send the profiles still under the name without
the godeltaprof prefix.


tpaschalis (approved):
LGTM personally, if @korniltsev is OK with it then we can merge.

korniltsev (approved):
LGTM with some notes

There is a piece of code determining if we should compute delta based on labels
https://github.com/grafana/agent/blob/6eb188959fe1dafda56ed34a8fb7c7fb9832ec99/component/pyroscope/scrape/scrape_loop.go#L188-L187

It looks like it is using `allLabels` , not the modified `publicLabels` so we should be ok there. I would maybe write a test to make sure delta appender is never used for godeltaprof renamed targets.

Another issue we could have is hash collision in case when there are 2 targets with memory and godeltaprof memory, the hash is calculated on public labels and they may end up the same after renaming. Not sure what to do best here. Maybe do not allow scraping godelttaprof_memory and memory at the same time from the same component?

simonswine:
This is now covered by: https://github.com/grafana/alloy/pull/662

PR #6369, opened 2024-02-14
Add build image for windows boringcrypto executable.

mattdurham:
#### PR Description

This lays the groundwork for building windows agent flow using boring crypto. The important bit for this is `FROM mcr.microsoft.com/oss/go/microsoft/golang:1.22-bullseye as golang` which tells it to use the fips compatible windows golang runtime.

Not adding a changelog until the second part of the pr to use the built image.

At some point may switch the linux version to using this so its cleaner on what builds what.


mattdurham:
Note I have built a fips windows version using my docker account mrdgrafana build image.

mattdurham:
I originally though about importing the core build image and only overwriting the go parts but then the build image needs to have another step and that seems painful. Basically would have to trigger a build on the normal build images, change the code for fips to use them, then trigger another build for the fips image.

build-image/boringcrypto/Dockerfile
```diff
@@ -0,0 +1,107 @@
+# syntax=docker/dockerfile:1.4
+
+# NOTE: The main difference in this file vs the parent Dockerfile is using mcr.microsoft.com/oss/go/microsoft/golang:1.22-bullseye which is a microsoft
```
rfratto, line 3: I'm worried about this getting desynced over time with the other Linux Dockerfile. Can we find a way to templatize the existing Linux Dockerfile instead of creating a new one? IIUC it's possible to have `ARG` statements which are controllable via the `docker build` command.

mattdurham: That worked like a charm, thanks!

build-image/Dockerfile
```diff
@@ -4,6 +4,10 @@
 # default when running `docker buildx build` or when DOCKER_BUILDKIT=1 is set
 # in environment variables.
 
+# NOTE: The GO_RUNTIME is used to switch between the default google go runtime and  mcr.microsoft.com/oss/go/microsoft/golang:1.22-bullseye which is a microsoft
+# fork of go that allows using windows crypto instead of boring crypto. Details at https://github.com/microsoft/go/tree/microsoft/main/eng/doc/fips
+ARG GO_RUNTIME=golang:1.22.0-bullseye
+
```
rfratto, line 10: nit: should we remove the default so we're forced to update it in the Jsonnet consistently?

.drone/drone.yml
```diff
@@ -1349,8 +1350,3 @@ get:
   path: infra/data/ci/github/updater-app
 kind: secret
 name: updater_private_key
----
-kind: signature
-hmac: d66a28c3489b7a19b64e82590630ceed4fc71789a76b3da8564c2aae604253ad
-
-...
```
rfratto, line 1322: Looks like the signature got removed, you'll have to resign I think

rfratto (approved):
LGTM

PR #6370, opened 2024-02-15
Fix two data races in declare tests

wildum:
#### PR Description

Fix two data races with mutex.

### DIFF
--- a/pkg/flow/internal/controller/component_node_manager.go
+++ b/pkg/flow/internal/controller/component_node_manager.go
@@ -74,3 +74,10 @@ func findLocalDeclare(reg *CustomComponentRegistry, componentName string) (ast.B
 	}
 	return nil, nil
 }
+
+func (m *ComponentNodeManager) SetCustomComponentRegistry(reg *CustomComponentRegistry) {
+	m.mut.Lock()
+	defer m.mut.Unlock()
+
+	m.customComponentReg = reg
+}
--- a/pkg/flow/internal/controller/component_node_manager.go
+++ b/pkg/flow/internal/controller/component_node_manager.go
@@ -75,7 +75,7 @@ func findLocalDeclare(reg *CustomComponentRegistry, componentName string) (ast.B
 	return nil, nil
 }
 
-func (m *ComponentNodeManager) SetCustomComponentRegistry(reg *CustomComponentRegistry) {
+func (m *ComponentNodeManager) setCustomComponentRegistry(reg *CustomComponentRegistry) {
 	m.mut.Lock()
 	defer m.mut.Unlock()
 
--- a/pkg/flow/internal/controller/custom_component_registry.go
+++ b/pkg/flow/internal/controller/custom_component_registry.go
@@ -1,13 +1,17 @@
 package controller
 
 import (
+	"sync"
+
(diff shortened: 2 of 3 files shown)


pkg/flow/internal/controller/component_node_manager.go
```diff
@@ -74,3 +74,10 @@ func findLocalDeclare(reg *CustomComponentRegistry, componentName string) (ast.B
 	}
 	return nil, nil
 }
+
+func (m *ComponentNodeManager) SetCustomComponentRegistry(reg *CustomComponentRegistry) {
```
thampiotr, line 78: This doesn't need to be exported, right?

thampiotr (approved):
LGTM % the exporting of method

pkg/flow/internal/controller/component_node_manager.go
wildum: nop, it's used only in the controller pkg right now. Feels weird not to export a setter but I guess I should stick with "if not used outside of pkg then should not export"

thampiotr: definitely! do not export something that is not used outside - keep classes well encapsulated. 

When reading this code, it would be easy to wonder what else might be calling this Set from outside the package, then one would need to search and notice that it's not used anywhere else and it would be a waste of few seconds and increase on a cognitive load on the reader (and maybe annoyance). Code that is easy to read is well encapsulated.

tpaschalis (commented):
:shipit:

tpaschalis (approved):

pkg/flow/internal/controller/component_node_manager.go
wildum: yeah, but this pkg is so big (controller) that we don't have much encapsulation and not exporting something does not make much different since that it's by default available to a very wide scope

thampiotr: > not exporting something does not make much different 

It does extend the scope in which the function is available by almost 3x if you look at the lines of code:
```
➜  agent git:(main) cloc pkg/flow/
-------------------------------------------------------------------------------
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
Go                              77           1763           1402           8318
-------------------------------------------------------------------------------
➜  agent git:(main) cloc pkg/flow/internal/controller
-------------------------------------------------------------------------------
Language                     files          blank        comment           code
-------------------------------------------------------------------------------
Go                              27            628            558           2825
-------------------------------------------------------------------------------
```

And the argument of the less encapsulated code having higher cognitive load and thus being harder to read still stands.

FWIW encapsulation is part of numerous best-practices and I don't see why we would deviate from it here?

wildum: sorry for the confusion, I agreed that it's better not to export it. I just wanted to add the fact that the scope of the pkg is already quite big that I don't find it well encapsulated, but better have a little bit more encapsulation if we can. I just wonder whether this is how Go works and that's ok or if we should split packages better in this area. I find it weird that Im adding a setter, but the field can be modified anyway in the whole pkg. I encountered some areas where there are getters and setters but the field is being accessed/modified directly and it's a risk

PR #6379, opened 2024-02-16
Add framework for converting OpenTelemetry Collector configs into Flow mode configs

rfratto:
> **NOTE**: This PR is intended to be reviewed commit-by-commit.

This PR adds the initial framework for converting OpenTelemetry Collector configs into Flow mode configs. 

As this converter is likely to be updated the most frequently, I have invested effort into trying to reduce the development effort required to support conversion for new components. 

The framework relies on implementations of `componentConverter` interface implementations, each of which handles the conversion of a single OpenTelemetry Collector component into one or more Flow mode components. 

For this initial PR, only conversion for receivers, processors, and exporters are supported. Future work will be needed to support converting connectors and extensions (auth extensions in particular).


tpaschalis (approved):
LGTM, I was able to follow things around and think that's a great start for us to add more converters gradually. I've only added some docs-related nits and clarification questions here and there, not meant to be something blocking.

converter/internal/otelcolconvert/converter.go
```diff
@@ -0,0 +1,196 @@
...
+	//
+	// * Use [state.FlowComponentLabel] to get a unique Flow label for the component
+	//   being generated.
+	//
+	// Implementations of componentConverter should use [state.FlowComponentLabel]
+	// to retrieve a unique Flow label for the component being converted.
+	//
+	// ConvertAndAppend may be called more than once with the same component used
+	// in different pipelines. Use [state.FlowComponentLabel] to get a guaranteed
+	// unique Flow component label for the current state.
```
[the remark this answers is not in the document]
rfratto: haha, I rewrote this three times and forgot to remove the older instances 😅

converter/internal/otelcolconvert/converter_otlpreceiver.go
```diff
@@ -0,0 +1,221 @@
...
+	converters = append(converters, otlpReceiverConverter{})
+}
+
+type otlpReceiverConverter struct{}
+
+func (otlpReceiverConverter) Factory() component.Factory { return otlpreceiver.NewFactory() }
+
+func (otlpReceiverConverter) InputComponentName() string { return "" }
+
+func (otlpReceiverConverter) ConvertAndAppend(state *state, id component.InstanceID, cfg component.Config) diag.Diagnostics {
```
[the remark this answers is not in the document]
rfratto: Yes, I added a comment for this. In most of our converters, the defaults are coming from the upstream tool and not the Flow component though; this is by design.

converter/internal/otelcolconvert/otelcolconvert.go
```diff
@@ -0,0 +1,255 @@
...
+			facts.Processors[fact.Type()] = fact
+		case exporter.Factory:
+			facts.Exporters[fact.Type()] = fact
+		case extension.Factory:
+			facts.Extensions[fact.Type()] = fact
+		case connector.Factory:
+			facts.Connectors[fact.Type()] = fact
+
+		default:
+			panic(fmt.Sprintf("unknown component factory type %T", fact))
```
erikbaranowski, line 135: WDYT about a diag on this rather than a panic? This seems like an unsupported ERROR diag ideally where the rest of the config can be potentially converted. If not possible to continue conversion here for otel, maybe a CRITICAL?

converter/internal/otelcolconvert/otelcolconvert.go
```diff
@@ -0,0 +1,255 @@
...
+		return nil, diags
+	}
+
+	cfg, err := readOpentelemetryConfig(in)
+	if err != nil {
+		diags.Add(diag.SeverityLevelCritical, err.Error())
+		return nil, diags
+	}
+	if err := cfg.Validate(); err != nil {
+		diags.Add(diag.SeverityLevelError, fmt.Sprintf("failed to validate config: %s", err))
```
erikbaranowski, line 60: Anything where a config isn't generated should be a CRITICAL diag.

converter/internal/otelcolconvert/otelcolconvert.go
```diff
@@ -0,0 +1,255 @@
...
+}
+
+// appendConfig converts the provided OpenTelemetry config into an equivalent
+// Flow config and appends the result to the provided file.
+func appendConfig(file *builder.File, cfg *otelcol.Config) diag.Diagnostics {
+	var diags diag.Diagnostics
+
+	groups, err := createPipelineGroups(cfg.Service.Pipelines)
+	if err != nil {
+		diags.Add(diag.SeverityLevelCritical, fmt.Sprintf("failed to interpret config: %s", err))
```
erikbaranowski, line 149: I'm marking this comment on one of them but it would be nice to see some testing where various diags get hit. It's not 100% obvious what will happen for each of them in terms of whether we get a partially correct .river or not

converter/internal/otelcolconvert/converter_otlpreceiver.go
```diff
@@ -0,0 +1,221 @@
...
+
+	return &otelcol.CORSArguments{
+		AllowedOrigins: cfg.AllowedOrigins,
+		AllowedHeaders: cfg.AllowedHeaders,
+
+		MaxAge: cfg.MaxAge,
+	}
+}
+
+type tokenizedConsumer struct {
```
erikbaranowski, line 208: I don't love seeing new types buried this deep into a file. WDYT about making a new file for these tokenization types for otel?

converter/internal/otelcolconvert/otelcolconvert.go
erikbaranowski: There are some other panics throughout I think it is worth a pass to see if you think they can become diags or not.

erikbaranowski (commented):
Solid work 🚀 I've added some thoughts and nits. I think this is super close to merging.

converter/internal/otelcolconvert/otelcolconvert.go
rfratto: Generally speaking, I use panics to indicate that the code is fundamentally broken; users shouldn't be able to run into these things. I can change these out with diags, but I also think that comes at a cost of complicating the code, as now I have to wire in more `diags.AddAll` calls and also check to see if a critical error has been generated yet. 

I see a few options here:

* Replace panics with critical diagnostics as you suggested
* Catch panics at the call to `Convert` and turn them into critical diagnostics to avoid complicating code 
* Leave the panics as-is with no changes

What would you prefer? The converters are your code so I'd rather follow your lead here

converter/internal/otelcolconvert/otelcolconvert.go
rfratto: I've made this critical. The differences between error and critical are a bit lost on me, so if I've missed any other cases where you think the severity level is wrong please let me know and I'll happily change it :)

converter/internal/otelcolconvert/otelcolconvert.go
rfratto: I added some tests for critical errors that can be hit; not all of them are possible to hit and are written defensively. 

I put these in a separate folder so they don't get in the way of the component conversion tests.

converter/internal/otelcolconvert/converter_otlpreceiver.go
rfratto: Moved to an converter_helpers.go file

rfratto:
@erikbaranowski Ready for another round of reviews; I addressed all of your comments except for the first one, after you let me know what you'd prefer I'll resolve that too

converter/internal/test_common/testing.go
```diff
@@ -86,12 +87,17 @@ func getExpectedDiags(t *testing.T, diagsFile string) []string {
 	if _, err := os.Stat(diagsFile); err == nil {
 		errorBytes, err := os.ReadFile(diagsFile)
 		require.NoError(t, err)
-		errorsString := string(normalizeLineEndings(errorBytes))
```
rfratto, line 89: This has been removed from the new code because `bufio.NewScanner` gracefully handles \r\n and \n out of the box.

converter/internal/otelcolconvert/otelcolconvert.go
erikbaranowski: I don't think this is a blocking thing if it complicates the code. We can always come back to it later if it becomes a stumbling point for users. Catching panics is interesting but I'm worried it will catch other panics and not display in a useful way for debugging if we start catching them and not display the whole error stack which kind of defeats the purpose.

erikbaranowski (approved):
Nice work, this sets us up really nicely for this converter

PR #6381, opened 2024-02-16
Add OpenTelemetry alerts

ptodev:
For now I'm just adding alerts for spans. In the near future we could also add alerts for metrics and logs.


tpaschalis (approved):
Thanks for doing this, I'd really like to see our mixin useful for all telemetry signals and this is a good start!

PR #6385, opened 2024-02-17
fix(loki/src/k8s_events): correctly update health state for the component during operation

### LINKED ISSUE #6181: Metric agent_component_controller_running_components when  component fails
### What's wrong?

When the `loki.source.kubernetes_events` component fails to load and exits, the value of the `agent_component_controller_running_components` metric does not change. All GA components are reported as healthy.

When with the enclosed configuration, the `Source:EventSource{Component:,Host:,}` component fails to load the `loki.source.kubernetes_events` reports 6 "healthy" components - the same value as with successfully loaded components. The `health_type` is always `"healthy"`.

The failure to load the component is reported only by this logline:

image

### Steps to reproduce

The Kubernetes Events source fails to load only occasionally and only in overloaded clusters. 
Maybe, to reproduce the problem with a dev version, the [informerSyncTimeout](https://github.com/grafana/agent/blob/main/component/loki/source/kubernetes_events/kubernetes_events.go#L26) can be set to a very low value

### System information

_No response_

### Software version

v0.37.2

### Configuration

```text
Full `config.river`


    logging {
      level  = "info"
      format = "json"
    }

    otelcol.receiver.otlp "otlp" {
      http {}

      output {
        metrics= [otelcol.exporter.prometheus.prom.input]
      }
    }

    otelcol.exporter.prometheus "prom" {
      forward_to  = [ prometheus.remote_write.default.receiver ]
    }

      endpoint {
        url = "http://prometheus-server.monitoring-system.svc.cluster.local/api/v1/write"
      }
    }

    loki.write "obs" {
      external_labels = { cluster = "staging" }
      endpoint {
        url = "https://loki.platform-staging.internal.xxx.yyy/loki/api/v1/push"
      }
    }

    loki.relabel "drop_instance" {
      forward_to = [loki.write.obs.receiver]
      rule {
        action = "labeldrop"
        regex  = "instance"
      }
    }

    loki.source.kubernetes_events "default" {
      forward_to = [loki.relabel.drop_instance.receiver]
      job_name = "kubernetes_events"
      log_format = "json"
    }
```
```


### Logs

_No response_

### PR DESCRIPTION
**Pull request by hainenber**

Fixes #6181 

#### PR Checklist

- [x] CHANGELOG.md updated

### CODE REVIEW

File: component/loki/source/kubernetes_events/kubernetes_events.go [hunk 1]
```diff
@@ -152,6 +156,7 @@ func (c *Component) Run(ctx context.Context) error {
 				c.tasksMut.RUnlock()
 
 				if err := c.runner.ApplyTasks(ctx, tasks); err != nil {
+					c.setHealth(err)
```
tpaschalis, line 159: I don't think this solves the issue in #6181 entirely.

ApplyTasks will return an error if the runner is closed, or if the context is already cancelled/exceeded via ctx.Err; we don't have a way to propagate errors directly from the underlying tasks as the worker interface is `Run(ctx context.Context)`.

To make this happen we'd need to have a goroutine-safe mechanism of bubbling up the errors and sieving through them.

hainenber: I've changed my approach and now choose one with smaller blast radius. That is, the `eventController` retain any error when executing `Run` method and there's a recurring actor in the component to set health based on the returned results.

Not exactly ideal since the component's health is not reflected right away but I think it's a worthwhile tradeoff

PR #6388, opened 2024-02-19
Add a namespace allowlist to loki.source.podlogs

### LINKED ISSUE #2867: Flow: add ability to restrict namespaces of loki.source.podlogs
Currently, `loki.source.podlogs` watches all namespaces for PodLogs, which requires giving the agent a ClusterRole binding. If a user wanted to restrict the permissions of the agent to a specific namespace, there's currently no way to do this; `loki.source.podlogs` always watches for PodLogs at the cluster scope. 

`loki.source.podlogs` should be configurable with an allowlist of namespaces to watch.

ptodev:
Fixes grafana/alloy#496. I think the solution should be something along those lines, but I'm not sure what is the best way to test my change? 

I don't think I could add unit tests without bootstrapping some sort of k8s environment? In that case, I could either do a one-off local test, or I could try to add something to the integration test suite? I don't think we currently have k8s integration tests.

cc @rfratto, @captncraig, @wildum

### DIFF
--- a/component/loki/source/podlogs/controller.go
+++ b/component/loki/source/podlogs/controller.go
@@ -47,7 +47,7 @@ func newController(l log.Logger, reconciler *reconciler) *controller {
 	}
 }
 
-func (ctrl *controller) UpdateConfig(cfg *rest.Config) error {
+func (ctrl *controller) UpdateConfig(cfg *rest.Config, namespaces []string) error {
 	scheme := runtime.NewScheme()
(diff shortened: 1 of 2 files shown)


ptodev:
Hi @dbluxo, indeed we don't use Agent anymore so I will close this issue. We will need a new PR ot be  raised for Alloy, but AFAIK no one is working on this issue at the moment.

PR #6395, opened 2024-02-20
docs: fix write_relabel_config block reference

### LINKED ISSUE #5741: Docs feedback: /docs/sources/flow/reference/components/prometheus.remote_write.md
The `write_relabel_config` section is an embedded link to `{{< docs/shared lookup="flow/reference/components/rule-block.md" source="agent" version="" >}}`.  However this results in a confusing final render as the documentation mentions "The rule block contains the definition..." instead of "The write_relabel_config block contains..."

tpaschalis:
image

The documentation of the write_relabel_config block reuses the rule.md shared snippet which results in confusing documentation. Since there's no way to templatize these to appear with a different block names, we had some options:

a) Rename the `write_relabel_config` block to `rule`. This makes relabeling rules consistent across all components, but is a breaking change.
b) Inline the documentation for this block since it's not being used anywhere else.
c) Create a copy of the shared block with minimal changes so it's easier to compare and keep in sync.

I think that the short-term solution c) might be a good tradeoff while we think about renaming the block itself.

Fixes #5741

### DIFF
--- a/docs/sources/flow/reference/components/prometheus.remote_write.md
+++ b/docs/sources/flow/reference/components/prometheus.remote_write.md
@@ -211,7 +211,7 @@ Name | Type | Description | Default | Required
 
 ### write_relabel_config block
 
-{{< docs/shared lookup="flow/reference/components/rule-block.md" source="agent" version="<AGENT_VERSION>" >}}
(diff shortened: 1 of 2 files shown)


clayton-cornell (approved):
LGTM

clayton-cornell:
Do we leave this in `next` (v0.40)? or backport?

PR #6401, opened 2024-02-20
flowtest: introduce script-based testing framework for Flow

rfratto:
This commit introduces a new package, flow/flowtest, which can be used for running script-based test using Flow components.

flowtest accepts a txtar archive, where the comment of the txtar archive denotes the script to run. Files in the txtar archive are unpacked to a temporary directory as part of the test; that temporary directory is also used as the working directory of the Flow controller.

See https://rsc.io/script for more information about the scripting language and available default commands.

The true value of this package is unknown, but it is expected to be useful in some cases as an alternative to writing pure Go tests. It is likely that more commands will be needed over time to cover more use cases, and we may find ourselves replacing existing tests with script-based tests where appropriate. 

## Flow-specific script commands

In addition to the default commands (`cd`, `cat`, `exec`, etc.), the following Flow-specific commands are made available to scripts:

- `controller.load [file]`: Load a file into the Flow controller.
- `controller.start`: Start the Flow controller.
- `assert.health [component] [health]`: Assert the health of a specific component.

Note that `controller.start` should almost always be run as a background command to allow the controller to run for the duration of the test.

Users of flowtest can provide custom commands on top of these by passing `WithExtraCommands` to `TestScript`.

## Example test script

The following is an example script using these commands:

```texture
# This file performs a basic test of loading a component and asserting its
# health.

controller.load main.river
controller.start &

assert.health local.file.example healthy

-- main.river --
local.file "example" {
  filename = "./hello.txt"
}

-- hello.txt --
Hello, world!
```

See the testdata/ directory for other test examples.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -583,7 +583,7 @@ require (
 	golang.org/x/mod v0.14.0 // indirect
 	golang.org/x/sync v0.5.0 // indirect
 	golang.org/x/term v0.16.0 // indirect
-	golang.org/x/tools v0.15.0 // indirect
+	golang.org/x/tools v0.15.0
 	golang.org/x/xerrors v0.0.0-20220907171357-04be3eba64a2 // indirect
 	gomodules.xyz/jsonpatch/v2 v2.4.0 // indirect
 	gonum.org/v1/gonum v0.14.0 // indirect
@@ -603,7 +603,7 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.3.0 // indirect
 )
 
-require github.com/ianlancetaylor/demangle v0.0.0-20230524184225-eabc099b10ab
+require github.com/ianlancetaylor/demangle v0.0.0-20230524184225-eabc099b10ab // indirect
 
 require (
 	connectrpc.com/connect v1.14.0
@@ -618,6 +618,7 @@ require (
 	go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetrichttp v0.42.0
 	golang.org/x/crypto/x509roots/fallback v0.0.0-20240208163226-62c9f1799c91
 	k8s.io/apimachinery v0.28.3
+	rsc.io/script v0.0.2-0.20231205190631-334f6c18cff3
 )
 
 require (
--- a/pkg/flow/flowtest/commands.go
+++ b/pkg/flow/flowtest/commands.go
@@ -1,0 +1,136 @@
+package flowtest
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/grafana/agent/component"
+	"github.com/grafana/agent/pkg/flow"
+	"github.com/grafana/dskit/backoff"
+	"rsc.io/script"
+)
+
+func createCommands(f *flow.Flow, opts *options) map[string]script.Cmd {
+	cmds := script.DefaultCmds()
+	registerControllerCommands(f, cmds)
+
+	for name, cmd := range opts.cmds {
+		cmds[name] = cmd
+	}
+
+	return cmds
+}
+
+func registerControllerCommands(f *flow.Flow, cmds map[string]script.Cmd) {
+	cmds["controller.load"] = script.Command(
+		script.CmdUsage{
+			Summary: "load a file into the Flow controller",
+			Args:    "file",
+		},
+		func(s *script.State, args ...string) (script.WaitFunc, error) {
+			if len(args) != 1 {
+				return nil, script.ErrUsage
+			}
+
+			bb, err := os.ReadFile(makeAbsolute(s.Getwd(), args[0]))
+			if err != nil {
+				return nil, err
+			}
+			src, err := flow.ParseSource(args[0], bb)
+			if err != nil {
+				return nil, fmt.Errorf("parsing %q: %w", args[0], err)
+			}
+			return nil, f.LoadSource(src, nil)
+		},
+	)
+
+	cmds["controller.start"] = script.Command(
+		script.CmdUsage{
+			Summary: "start the Flow controller",
+			Async:   true,
+			Detail: []string{
+				"This command should almost always be run in the background by appending a &, otherwise the controller will exit immediately after the command finishes'",
+			},
+		},
+		func(s *script.State, args ...string) (script.WaitFunc, error) {
+			if len(args) != 0 {
+				return nil, script.ErrUsage
+			}
+
+			var wg sync.WaitGroup
+			wg.Add(1)
+
+			ctx, cancel := context.WithCancel(s.Context())
+			go func() {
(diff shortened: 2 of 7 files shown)


pkg/flow/flowtest/commands.go
```diff
@@ -0,0 +1,136 @@
...
+			return nil, f.LoadSource(src, nil)
+		},
+	)
+
+	cmds["controller.start"] = script.Command(
+		script.CmdUsage{
+			Summary: "start the Flow controller",
+			Async:   true,
+			Detail: []string{
+				"This command should almost always be run in the background by appending a &, otherwise the controller will exit immediately after the command finishes'",
```
thampiotr, line 57: if & is missing, wouldn't the script block execution until controller exits (which could be forever)?

pkg/flow/flowtest/commands.go
```diff
@@ -0,0 +1,136 @@
...
+	registerControllerCommands(f, cmds)
+
+	for name, cmd := range opts.cmds {
+		cmds[name] = cmd
+	}
+
+	return cmds
+}
+
+func registerControllerCommands(f *flow.Flow, cmds map[string]script.Cmd) {
```
thampiotr, line 29: It's quite a long function, maybe we should put each command in its own function.

pkg/flow/flowtest/commands.go
```diff
@@ -0,0 +1,136 @@
...
+			}
+			src, err := flow.ParseSource(args[0], bb)
+			if err != nil {
+				return nil, fmt.Errorf("parsing %q: %w", args[0], err)
+			}
+			return nil, f.LoadSource(src, nil)
+		},
+	)
+
+	cmds["controller.start"] = script.Command(
```
thampiotr, line 52: I thought it would make most sense to be as close to user experience as possible with those tests so things like quirks of CLI parsing or conflicting flag combinations or environment variables can be tested. Also, we can completely refactor things inside and validate that the users' interface remains the same.

The users call `grafana-agent run <config file>` and they don't start a controller themselves, so I would think that instead of operating on the controller level we should operate on agent's CLI level? We'd still need a way to extract controller from a running agent for the purposes of asserting on conditions like 'healthy', which is fine.

pkg/flow/flowtest/commands.go
```diff
@@ -0,0 +1,136 @@
...
+
+	for name, cmd := range opts.cmds {
+		cmds[name] = cmd
+	}
+
+	return cmds
+}
+
+func registerControllerCommands(f *flow.Flow, cmds map[string]script.Cmd) {
+	cmds["controller.load"] = script.Command(
```
thampiotr, line 30: How would testing of two agents work? for example a proxy scenario or a cluster of agents?

pkg/flow/flowtest/testdata/basic.txtar
```diff
@@ -0,0 +1,15 @@
+# This file performs a basic test of loading a component and asserting its
+# health.
+
+controller.load main.river
+controller.start &
+
+assert.health local.file.example healthy
+
+-- main.river --
```
thampiotr, line 9: I do like the embedded files :)

pkg/flow/flowtest/flowtest.go
```diff
@@ -0,0 +1,195 @@
...
+// available default commands.
+//
+// # Flow commands
+//
+// In addition to the default commands provided by [rsc.io/script], the
+// following Flow-specific commands are available:
+//
+//   - `controller.load [file]`: Load a file into the Flow controller.
+//   - `controller.start`: Start the Flow controller.
+//   - `assert.health [component] [health]`: Assert the health of a specific component.
```
thampiotr, line 17: I think we should just point to the file that registers all these commands. This is likely to get outdated fast.

pkg/flow/flowtest/flowtest_test.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func Test(t *testing.T) {
+	filepath.WalkDir("testdata/", func(path string, d fs.DirEntry, err error) error {
+		if filepath.Ext(path) != ".txtar" {
+			return nil
+		}
+		t.Run(d.Name(), func(t *testing.T) {
+			require.NoError(t, flowtest.TestScript(path))
```
thampiotr, line 18: Could we instead pass the `testing.T` to `TestScript` and further, so that we can use the rich assert/require library? These can facilitate stuff like comparing maps, checking numbers are within a range, checking something is contained in the other thing, using the 'eventually' functions... the error messages produced by these are usually quite clear and useful. So I think it's worth to make it possible to use them - especially will be useful for all the asserting commands.

pkg/flow/flowtest/flowtest.go
```diff
@@ -0,0 +1,195 @@
...
+	defer os.RemoveAll(tmpDir)
+
+	archive, err := txtar.ParseFile(filename)
+	if err != nil {
+		return fmt.Errorf("loading script: %w", err)
+	}
+
+	// Create a test controller for the script to interact with. The test
+	// controller's logs are buffered and printed to stderr on exit. This
+	// prevents log mangling when both the script engine and the Flow controller
```
thampiotr, line 90: We could put prefixes in logs and print at the same time? Splitting them becomes a problem when we need to correlate the logs of script with the logs of the controller to see what happened in what order.

go.mod
```diff
@@ -583,7 +583,7 @@ require (
 	golang.org/x/mod v0.14.0 // indirect
 	golang.org/x/sync v0.5.0 // indirect
 	golang.org/x/term v0.16.0 // indirect
-	golang.org/x/tools v0.15.0 // indirect
+	golang.org/x/tools v0.15.0
```
ptodev, line 586: It'd be nice if this this goes in the first import block with direct dependencies

pkg/flow/flowtest/flowtest.go
```diff
@@ -0,0 +1,195 @@
...
+// working directory of the process is changed to the temporary directory for
+// the duration of the call.
+func TestScript(filename string, opts ...TestOption) error {
+	var o options
+	for _, opt := range opts {
+		opt(&o)
+	}
+
+	// Create a temporary directory for the scope of our test to operate in.
+	tmpDir, err := os.MkdirTemp("", "flowtest-*")
```
ptodev, line 77: Are those directories cleaned up?

ptodev (commented):
Thank you, this looks interesting and I look forward to using it! I wonder how convenient it will be with debuggers too. E.g. I guess there won't be any issues with the temp folders if I terminate the process mid-session while debugging.

rfratto:
I'm going to close this for now. I'm still interested in a script-based testing framework, but it sounds like we need more thoughts around what the scripts should look like (and how to use them to test things like clustering).

@thampiotr expressed interest in taking inspiration from this exploration, so I'm handing it over to him for future work on this :)

PR #6412, opened 2024-02-21
Fix import tests windows

wildum:
#### PR Description

- use path.join to create a path instead of building it manually
- add eol rule for txtar files in .gitattributes to fix the parsing on windows


tpaschalis (approved):
LGTM, thanks!

PR #6417, opened 2024-02-21
[FEAT #6416] nodePort value can be set now

### LINKED ISSUE #6416: Possibility to set the value of NodePort with the Helm Chart
### Request

If someone choose `NodePort` as service type in the [helm chart service.yaml template](https://github.com/grafana/agent/blob/main/operations/helm/charts/grafana-agent/templates/service.yaml), then the port is chosen randomly.
We need to be able to set it.

### Use case

Mostly for `minikube` or `kind` Kubernetes cluster where Ingress might be too much and where we don't want to port forward the service each time a cluster is created.

ryayon:
#### PR Description

Allow to set a port when the service is `NodePort` type

Fixes #6416

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added


tpaschalis (commented):
Hey there @ryayon! Apologies for the delayed reply here!

I think this does make sense as a PR. I'm still on the fence on if we should provide a random-ish default like you did here, or default to an empty value that preserves the previous behavior (but also allow setting it).

In the meantime, could you please fix the linting error about the README file, and also possibly add a test?

ryayon:
> Hey there @ryayon! Apologies for the delayed reply here!
> 
> I think this does make sense as a PR. I'm still on the fence on if we should provide a random-ish default like you did here, or default to an empty value that preserves the previous behavior (but also allow setting it).
> 
> In the meantime, could you please fix the linting error about the README file, and also possibly add a test?

No worries.
Thank you for your feedback and let me know for the default port value.
In the meantime, I fixed the README (although I am not sure what was the issue) and I created a specific test.

tpaschalis:
cc @ryayon sorry for the hassle here; I've fixed a merge conflict (so you might have to pull first), but can you run `make generate-helm-tests` to also regenerate the test folders?

cc @captncraig in case he has any strong opinions here, but I think that it'd be best if we let the default value empty and just set it if the user provided a non-default value instead.

ryayon:
No need to feel sorry @tpaschalis .
Thank you for your help and time.
I just pushed a new commit with the test folders regenerated.
In addition, I pulled all previous commits as suggested.
Let me know for any issue.

operations/helm/charts/grafana-agent/values.yaml
```diff
@@ -254,6 +254,8 @@ service:
   enabled: true
   # -- Service type
   type: ClusterIP
+  # -- NodePort port
+  nodePort: 31128
```
captncraig, line 258: Perhaps the comment could indicate this only takes effect when type is `NodePort`.

ryayon: Absolutely.
I just fixed it.

captncraig (approved):

captncraig:
The ci wants you to run `make generate-helm-docs`.

ryayon:
> The ci wants you to run `make generate-helm-docs`.

Done

PR #6462, opened 2024-02-21
Fix data race in custom component registry

wildum:
#### PR Description

The race was found in drone ci pipeline: https://drone.grafana.net/grafana/agent/16673/4/2

### DIFF
--- a/pkg/flow/internal/controller/custom_component_registry.go
+++ b/pkg/flow/internal/controller/custom_component_registry.go
@@ -28,6 +28,20 @@ func NewCustomComponentRegistry(parent *CustomComponentRegistry) *CustomComponen
 	}
 }
 
+func (s *CustomComponentRegistry) getDeclare(name string) (ast.Body, bool) {
+	s.mut.Lock()
+	defer s.mut.Unlock()
+	declare, ok := s.declares[name]
+	return declare, ok
+}
+
+func (s *CustomComponentRegistry) getImport(name string) (*CustomComponentRegistry, bool) {
+	s.mut.Lock()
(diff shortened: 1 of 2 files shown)


pkg/flow/internal/controller/custom_component_registry.go
```diff
@@ -28,6 +28,20 @@ func NewCustomComponentRegistry(parent *CustomComponentRegistry) *CustomComponen
 	}
 }
 
+func (s *CustomComponentRegistry) getDeclare(name string) (ast.Body, bool) {
+	s.mut.Lock()
+	defer s.mut.Unlock()
```
spartan0x117, line 33: nit, non-blocking: since this is only reading, could it be an RLock and RUnlock?
```suggestion
	s.mut.RLock()
	defer s.mut.RUnlock()
```

pkg/flow/internal/controller/custom_component_registry.go
```diff
@@ -28,6 +28,20 @@ func NewCustomComponentRegistry(parent *CustomComponentRegistry) *CustomComponen
...
+	declare, ok := s.declares[name]
+	return declare, ok
+}
+
+func (s *CustomComponentRegistry) getImport(name string) (*CustomComponentRegistry, bool) {
+	s.mut.Lock()
+	defer s.mut.Unlock()
```
spartan0x117, line 40: nit, non-blocking: since this is only reading, could it be an RLock and RUnlock?
```suggestion
	s.mut.RLock()
	defer s.mut.RUnlock()
```

spartan0x117 (approved):
LGTM!

PR #6464, opened 2024-02-21
otelcolconvert: support converting zipkinreceiver

### LINKED ISSUE #6423: Support converting `zipkinreceiver`

rfratto:
Closes #6423

### DIFF
--- a/converter/internal/otelcolconvert/converter_zipkinreceiver.go
+++ b/converter/internal/otelcolconvert/converter_zipkinreceiver.go
@@ -1,0 +1,54 @@
+package otelcolconvert
+
+import (
+	"fmt"
+
+	"github.com/grafana/agent/component/otelcol"
+	"github.com/grafana/agent/component/otelcol/receiver/zipkin"
+	"github.com/grafana/agent/converter/diag"
+	"github.com/grafana/agent/converter/internal/common"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/receiver/zipkinreceiver"
+	"go.opentelemetry.io/collector/component"
+)
+
+func init() {
+	converters = append(converters, zipkinReceiverConverter{})
+}
+
+type zipkinReceiverConverter struct{}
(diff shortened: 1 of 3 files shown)


converter/internal/otelcolconvert/converter_zipkinreceiver.go
```diff
@@ -0,0 +1,54 @@
...
+
+func toZipkinReceiver(state *state, id component.InstanceID, cfg *zipkinreceiver.Config) *zipkin.Arguments {
+	var (
+		nextTraces = state.Next(id, component.DataTypeTraces)
+	)
+
+	return &zipkin.Arguments{
+		ParseStringTags: cfg.ParseStringTags,
+		HTTPServer:      *toHTTPServerArguments(&cfg.HTTPServerSettings),
+
```
erikbaranowski, line 49: It's probably not worth it since the DebugMetrics default is currently the golang default but you could set DebugMetrics on the flow argument to a copy of its default value explicitly if you wanted to protect against DebugMetrics getting a different default in the future. This is probably overengineering for something that is not likely to happen but is the safest way to deal with arguments that exist in flow and not the original format.

erikbaranowski: Whatever you decide here would apply in general to anywhere we have DebugMetrics

erikbaranowski (approved):
LGTM! 1 comment to consider for future-proofing

converter/internal/otelcolconvert/converter_zipkinreceiver.go
rfratto: I talked offline with Erik about this; I agree with his suggestion but I'd prefer to do it in a follow-up PR so I can handle all of the components at once.

rfratto: Resolved in #6469

PR #6466, opened 2024-02-21
otelcolconvert: support converting kafkareceiver

### LINKED ISSUE #6425: Support converting `kafkareceiver`

rfratto:
Converting this component is more complicated than other receivers due to its usage of internal struct types. We work around internal structs by converting to `map[string]any` when necessary, including a conversion back to a concrete struct type where it's exposed to us.

Due to the nature of this component using internal types, the test covers more fields than normal to ensure the mapstructure usage is correct.

Closes #6425.

### DIFF
--- a/converter/internal/otelcolconvert/converter_kafkareceiver.go
+++ b/converter/internal/otelcolconvert/converter_kafkareceiver.go
@@ -1,0 +1,195 @@
+package otelcolconvert
+
+import (
+	"fmt"
+
+	"github.com/davecgh/go-spew/spew"
+	"github.com/grafana/agent/component/otelcol"
+	"github.com/grafana/agent/component/otelcol/receiver/kafka"
+	"github.com/grafana/agent/converter/diag"
+	"github.com/grafana/agent/converter/internal/common"
+	"github.com/grafana/river/rivertypes"
+	"github.com/mitchellh/mapstructure"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/exporter/kafkaexporter"
+	"github.com/open-telemetry/opentelemetry-collector-contrib/receiver/kafkareceiver"
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/config/configtls"
+)
+
+func init() {
+	converters = append(converters, kafkaReceiverConverter{})
+}
+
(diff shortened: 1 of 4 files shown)


component/otelcol/receiver/kafka/kafka.go
```diff
@@ -81,7 +81,7 @@ var DefaultArguments = Arguments{
 	},
 	HeaderExtraction: HeaderExtraction{
 		ExtractHeaders: false,
-		Headers:        []string{},
+		Headers:        nil,
```
erikbaranowski, line 84: The flow docs have a default of `[]` so I think we need to update that with this change.

converter/internal/otelcolconvert/converter_kafkareceiver.go
```diff
@@ -0,0 +1,183 @@
...
+		Username:  cfg["username"].(string),
+		Password:  rivertypes.Secret(cfg["password"].(string)),
+		Mechanism: cfg["mechanism"].(string),
+		Version:   cfg["version"].(int),
+		AWSMSK:    toKafkaAWSMSK(encodeMapstruct(cfg["aws_msk"])),
+	}
+}
+
+func toKafkaAWSMSK(cfg map[string]any) kafka.AWSMSKArguments {
+	return kafka.AWSMSKArguments{
```
erikbaranowski, line 113: are we missing a nil check at the start of this one or is it unneeded here?

erikbaranowski (approved):
LGTM, a couple very small things to take a look at

component/otelcol/receiver/kafka/kafka.go
rfratto: I chatted with Erik offline about this; `[]string{}` and `nil` are identical in River (both are `[]`), so the docs wouldn't need to change. The difference is only meaningful specifically from a Go context when trying to see if the Go values match defaults exactly.

converter/internal/otelcolconvert/converter_kafkareceiver.go
rfratto: I'll add one to be safe but it shouldn't be necessary since it's a non-pointer in OpenTelemetry; it'll always be available.

PR #6469, opened 2024-02-21
otelcolconvert: automatically sync DebugMetrics with Flow defaults

rfratto:
Follow up on https://github.com/grafana/agent/pull/6464#discussion_r1498140245

### DIFF
--- a/converter/internal/common/river_utils.go
+++ b/converter/internal/common/river_utils.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/grafana/river"
 	"github.com/grafana/river/parser"
 	"github.com/grafana/river/printer"
 	"github.com/grafana/river/scanner"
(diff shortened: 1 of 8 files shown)


erikbaranowski (approved):

converter/internal/common/river_utils.go
```diff
@@ -124,3 +125,16 @@ func SanitizeIdentifierPanics(in string) string {
...
 }
+
+// DefaultValue returns the default value for a given type. If T implements
+// river.Defaulter, a value will be returned with defaults applied. If T does
+// not implement river.Defaulter, the zero value is returned.
+//
+// T must not be a pointer type.
+func DefaultValue[T any]() T {
+	var val T
+	if defaulter, ok := any(&val).(river.Defaulter); ok {
```
rfratto, line 136: Go generics don't make this easy; only pointer types implement river.Defaulter, and T isn't a pointer type, so we have to get the address of val here. 

I also tried requiring T to be a pointer type, but then there's no way to allocate a value for T, so this requires it to _not_ be a pointer type.

PR #6486, opened 2024-02-22
Fix an issue where services were included in the graph and updated in a child flow controller

erikbaranowski:
#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,6 +100,8 @@ Main (unreleased)
 
 - Fix bug in `pyroscope.ebpf` component when elf's PT_LOAD section is not page aligned . [PR](https://github.com/grafana/pyroscope/pull/2983)  (@korniltsev)
 
+- Fix an issue where services were included in the graph and updated in a child flow controller. (@erikbaranowski)
+
 ### Other changes
 
 - Removed support for Windows 2012 in line with Microsoft end of life. (@mattdurham)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,7 +100,7 @@ Main (unreleased)
 
 - Fix bug in `pyroscope.ebpf` component when elf's PT_LOAD section is not page aligned . [PR](https://github.com/grafana/pyroscope/pull/2983)  (@korniltsev)
 
-- Fix an issue where services were included in the graph and updated in a child flow controller. (@erikbaranowski)
+- Fix an issue where the configuration of the `http` and `remotecfg` blocks get ignored after loading a module. (@erikbaranowski)
 
 ### Other changes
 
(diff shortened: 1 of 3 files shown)


pkg/flow/internal/controller/loader.go
```diff
@@ -828,10 +832,9 @@ func multierrToDiags(errors error) diag.Diagnostics {
 	return diags
 }
 
-// If the definition of a module ever changes, update this.
-func (l *Loader) isModule() bool {
-	// Either 1 of these checks is technically sufficient but let's be extra careful.
-	return l.globals.OnExportsChange != nil && l.globals.ControllerID != ""
+// isRootController returns true if the loader is for the root flow controller.
```
erikbaranowski, line 835: We considered separating the module vs non root controller concept but decided it wasn't worth the potential complexity of additional code paths. With the new `isRootController` note that the logical operator was inverted and usage was updated everywhere.

pkg/flow/internal/controller/loader.go
```diff
@@ -402,7 +406,7 @@ func (l *Loader) populateServiceNodes(g *dag.Graph, serviceBlocks []*ast.BlockSt
 	for _, block := range serviceBlocks {
 		blockID := BlockComponentID(block).String()
 
-		if l.isModule() {
+		if !l.isRootController() {
```
rfratto, line 409: This loop is likely to panic now; line 419 tries to get a node in the graph that will never exist in non-root controllers, and then immediately calls `Block` on that nil node.

CHANGELOG.md
```diff
@@ -100,6 +100,8 @@ Main (unreleased)
 
 - Fix bug in `pyroscope.ebpf` component when elf's PT_LOAD section is not page aligned . [PR](https://github.com/grafana/pyroscope/pull/2983)  (@korniltsev)
 
+- Fix an issue where services were included in the graph and updated in a child flow controller. (@erikbaranowski)
```
rfratto, line 103: Can we rephrase this so it only talks about concepts users are familiar with, instead of implementation details? Something like: 

```suggestion
- Fix an issue where the configuration of the `http` and `remotecfg` blocks get ignored after loading a module. (@erikbaranowski)
```

rfratto (commented):
LGTM minus the potential pointer issue

pkg/flow/internal/controller/loader.go
erikbaranowski: there's a continue inside this block so I think we are ok. we did bump into this exact panic before getting the code right so I know what you mean.

tpaschalis (approved):
LGTM, I think the reversed conditional is easier to read and reason about. Thanks!

PR #6502, opened 2024-02-23
Modules import file folder

### LINKED ISSUE #6460: import.file: support importing folders as a single module
#5547 mentions that import.file should support importing folders, but it currently only supports importing single files. import.file should be updated to support folders.

IMO, this should _not_ include folder support in local.file, as that would require much more thinking beyond the use case of what we need for import.file.

Related to #6461, which is the same issue but for import.git.

wildum:
#### PR Description

Add support for importing folders as modules via the import.file config block.

Fixes #6460 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


pkg/flow/internal/controller/node_config_import.go
```diff
@@ -236,6 +241,21 @@ func (cn *ImportConfigNode) onContentUpdate(importedContent string) {
...
+
+	for key, value1 := range map1 {
+		value2, ok := map2[key]
+		if !ok || value1 != value2 {
+			return false
+		}
+	}
+
+	return true
+}
```
rfratto, line 257: nit: can we use [maps.Equal](https://pkg.go.dev/maps#Equal) from the standard library instead?

docs/sources/flow/reference/config-blocks/import.file.md
```diff
@@ -15,16 +15,20 @@ title: import.file
...
+in the same folder.
+
 [module]: {{< relref "../../concepts/modules.md" >}}
 
 ## Usage
 
 ```river
 import.file "NAMESPACE" {
-  filename = FILENAME
+  filename = FILENAME_OR_FOLDERNAME
```
rfratto, line 31: nit: the documentation for the [run command](https://grafana.com/docs/agent/latest/flow/reference/cli/run/) calls this `PATH_NAME`; should we use the same here to make it consistent?

pkg/flow/internal/importsource/import_file.go
rfratto, line 1: I'm concerned that the complexity being introduced in this file is avoidable. As it is now, I have three major pieces of feedback:

* Subjectively: it feels like we're turning ImportFile into a mini-controller, given that it has to manage a set of components.
* Subjectively: it also feels like we're reimplementing logic in `local.file` for directory watching. 
* Objectively: You're not using the same mechanisms to detect changes to directories that `local.file` uses; `local.file` detects changes quickly using fsnotify but directory changes can only be detected after the poll interval, which is slow by default (60s), and may surprise users when they have to wait 60 seconds for a new file to be used. 

Have you considered abstracting common logic that `local.file` uses out into a new package, and having that package optionally support folders? That way both `local.file` and `import.file` can make use of the same common code, while only `import.file` supports folders.

rfratto (commented):
Thanks! Nearly LGTM, but I have some concerns about the way we've modified `import.file` to handle multiple folders.

pkg/flow/internal/importsource/import_file.go
wildum: Thanks for the review, that's a good point. We could extract the detector.go into its own pkg and extend the FSNotify detector to optionally support folders.
I agree that this should reduce the complexity of import_file and after a brief look at the fsNotify pkg it seems quite easy to handle folders. + It solves the update delay that my solution has between the dir changes and the file changes. 
I'm changing the PR into a Draft while I make the changes to prevent people from reviewing code that will change and I will reopen it when it's ready

CHANGELOG.md
```diff
@@ -10,9 +10,12 @@ internal API changes are not present.
 Main (unreleased)
 -----------------
 
+### Enhancements
+
+- Add support for importing folders as single module to `import.file`. (@wildum)
 
 v0.40.0-rc.2 (2024-02-26)
--------------------------
+--------------------
```
wildum, line 18: fix this

docs/sources/flow/concepts/modules.md
```diff
@@ -26,7 +26,7 @@ Modules can be [imported](#importing-modules) to enable the reuse of [custom com
 A module can be _imported_, allowing the custom components defined by that module to be used by other modules, called the _importing module_.
 Modules can be imported from multiple locations using one of the `import` configuration blocks:
 
-* [import.file]: Imports a module from a file on disk.
+* [import.file]: Imports a module from a file or a folder on disk.
```
clayton-cornell, line 29: Do we really mean `folder` here? or should this be `directory`?  My instinct is to say `directory` here and in the other instances.

Grafana defaults to the Google wordlist: https://developers.google.com/style/word-list#directory which says... it should be `directory`

docs/sources/flow/reference/config-blocks/import.file.md
```diff
@@ -15,26 +15,32 @@ title: import.file
 
 {{< docs/shared lookup="flow/stability/beta.md" source="agent" version="<AGENT_VERSION>" >}}
 
-The `import.file` block imports custom components from a file and exposes them to the importer.
+The `import.file` block imports custom components from a file or a folder and exposes them to the importer.
```
clayton-cornell, line 18: Same question as above.. do we really mean `folder` here? or is this actually a `directory`?

CHANGELOG.md
wildum, line 18:
```suggestion
-------------------------
```

internal/file-detector/detector.go
```diff
@@ -84,7 +84,7 @@ type fsNotifyOptions struct {
 
 // newFSNotify creates a new fsnotify detector which uses filesystem events to
 // detect that a file has changed.
-func newFSNotify(opts fsNotifyOptions) (*fsNotify, error) {
+func NewFSNotify(opts FsNotifyOptions) (*fsNotify, error) {
```
rfratto, line 87: nit: it's bad form for an exported function/method to return a non-exported type; you should export `fsNotify`

internal/file-detector/detector.go
```diff
@@ -75,7 +75,7 @@ type fsNotify struct {
 	watcher    *fsnotify.Watcher
 }
 
-type fsNotifyOptions struct {
+type FsNotifyOptions struct {
```
rfratto, line 78: nit: it's conventional in Go for acronyms to be all uppercase; this should be called `FSNotifyOptions`

internal/file-detector/detector.go
```diff
@@ -159,18 +159,18 @@ func (fsn *fsNotify) Close() error {
...
-type pollerOptions struct {
+type PollerOptions struct {
 	Filename      string
 	ReloadFile    func() // Callback to request file reload.
 	PollFrequency time.Duration
 }
 
 // newPoller creates a new poll-based file update detector.
-func newPoller(opts pollerOptions) *poller {
+func NewPoller(opts PollerOptions) *poller {
```
rfratto, line 173: ditto as above: `poller` should be exported as `Poller`

pkg/flow/internal/controller/node_config_import.go
```diff
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"go.uber.org/atomic"
+	"golang.org/x/exp/maps"
```
rfratto, line 14: nit: maps is available in stdlib now: https://pkg.go.dev/maps, that should be used when possible

internal/file-detector/detector.go
rfratto, line 1: nit: it's conventional in Go for folder names to match the package name; `internal/file-detector` should be `internal/filedetector`

pkg/flow/internal/importsource/import_file.go
```diff
@@ -3,96 +3,236 @@ package importsource
...
+	return err
 }
 
 func (im *ImportFile) Run(ctx context.Context) error {
-	return im.fileComponent.Run(ctx)
+	defer func() {
+		if err := im.detector.Close(); err != nil {
+			level.Error(im.managedOpts.Logger).Log("msg", "failed to shut down detector", "err", err)
+		}
+		im.detector = nil
```
rfratto, line 126: `im.detector` is protected by a mutex, so you should grab the mutex here before closing and resetting it

pkg/flow/internal/importsource/import_file.go
```diff
@@ -3,96 +3,236 @@ package importsource
...
+			PollFrequency: im.args.PollFrequency,
+		})
+	case filedetector.DetectorFSNotify:
+		im.detector, err = filedetector.NewFSNotify(filedetector.FsNotifyOptions{
+			Logger:        im.managedOpts.Logger,
+			Filename:      im.args.Filename,
+			ReloadFile:    reloadFile,
+			PollFrequency: im.args.PollFrequency,
+		})
+	}
```
rfratto, line 116: If `Update` is called twice, the old value for `im.detector` leaks, no? It should be closed here before setting a new one.

pkg/flow/internal/importsource/import_file.go
```diff
@@ -3,96 +3,236 @@ package importsource
...
+	files := make([]string, 0)
+	im.importDir = fi.IsDir()
+	if im.importDir {
+		files, err = collectFilesFromDir(fpath)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		files = append(files, fpath)
+	}
```
rfratto, line 209: nit: it doesn't seem necessary to store `importDir` on the struct, since it's only ever written to and read on the same code path (`readFile`). 

You could change `collectFiles` to return whether it was a directory instead:

```go
func (im *ImportFile) collectFiles() (content []string, dir bool, err error)
```

rfratto (approved):
LGTM, I have some real comments and some nits; nits are prefixed as such so you can choose to ignore them.

(This is still pending an approval from Clayton)

docs/sources/flow/concepts/modules.md
rfratto: We mean directory cc @wildum, these should change to say "directory"

docs/sources/flow/reference/config-blocks/import.file.md
```diff
@@ -15,26 +15,32 @@ title: import.file
...
 `import.file` blocks must be given a label that determines the namespace where custom components are exposed.
 
+Imported folders are treated as single modules to support composability.
+That means that you can define a custom component in one file and use it in another custom component in another file
+in the same folder.
+
+[module]: {{< relref "../../concepts/modules.md" >}}
```
clayton-cornell, line 25:
```suggestion

```

I don't think we need this relref definition here.

docs/sources/flow/reference/config-blocks/import.file.md
```diff
@@ -15,26 +15,32 @@ title: import.file
...
 The following arguments are supported:
 
-Name             | Type       | Description                                         | Default      | Required
------------------|------------|-----------------------------------------------------|--------------|---------
-`filename`       | `string`   | Path of the file on disk to watch.                  |              | yes
-`detector`       | `string`   | Which file change detector to use (fsnotify, poll). | `"fsnotify"` | no
-`poll_frequency` | `duration` | How often to poll for file changes.                 | `"1m"`       | no
+| Name             | Type       | Description                                         | Default      | Required |
+| ---------------- | ---------- | --------------------------------------------------- | ------------ | -------- |
+| `filename`       | `string`   | Path of the file or folder on disk to watch.        |              | yes      |
```
clayton-cornell, line 41:
```suggestion
| `filename`       | `string`   | Path of the file or directory on disk to watch.        |              | yes      |
```

rfratto:
@clayton-cornell Since you're off today and William is OOO next week, I asked for this to get merged so we can be unblocked. If you have any docs feedback for us to change, I'll open a PR for it next week

clayton-cornell:
@rfratto All good for docs. There's nothing that needs attention.

PR #6504, opened 2024-02-23
otelcolconvert: support converting span processor

### LINKED ISSUE #6455: Support converting `spanprocessor`

tpaschalis:
Closes #6455

I had to add a few more encode helper methods to access internal slice/string types.

I was worried about the number of nested fields, but I used all the options with a span.yaml like the following and got eveything ported over to the River config without any panics.

<details>

```
receivers:
  otlp:
    protocols:
      grpc:
      http:

exporters:
  otlp:
    # Our defaults have drifted from upstream, so we explicitly set our
    # defaults below (balancer_name and queue_size).
    endpoint: database:4317
    balancer_name: pick_first
    sending_queue:
      queue_size: 5000

processors:
  span:
    name:
      from_attributes: ["db.svc", "operation"]
      separator: "::"
      to_attributes:
        rules:
          - ^\/api\/v1\/document\/(?P<documentId>.*)\/update$
    status:
      code: Error
      description: "some error description"
    include:
      match_type: "strict"
      attributes:
        - key: "key1"
          value: "value1"
      span_names: ["span1", "span2"]
      span_kinds: ["spankind1", "spankind2"]
      log_bodies: ["lb1", "lb2"]
      log_severity_texts: ["ls1", "ls2"]
    exclude:
      match_type: "regex"
      services: ["svc1", "svc2"]
      resources:
        - key: "key1"
          value: "value1"
      libraries:
        - name: "name1"
          version: "version1"
      log_severity_number:
        min: 1
        match_undefined: false
      metric_names: ["mn1", "mn2"]

service:
  pipelines:
    metrics:
      receivers: [otlp]
      processors: []
      exporters: [otlp]
    logs:
      receivers: [otlp]
      processors: []
      exporters: [otlp]
    traces:
      receivers: [otlp]
      processors: [span]
      exporters: [otlp]
```

</details>

### DIFF
--- a/converter/internal/otelcolconvert/converter_spanprocessor.go
+++ b/converter/internal/otelcolconvert/converter_spanprocessor.go
@@ -44,20 +44,20 @@ func toSpanProcessor(state *state, id component.InstanceID, cfg *spanprocessor.C
 		nextTraces = state.Next(id, component.DataTypeTraces)
 	)
 
-	setStatus := &span.Status{}
-	if cfg.SetStatus == nil {
-		setStatus = nil
-	} else {
-		setStatus.Code = cfg.SetStatus.Code
-		setStatus.Description = cfg.SetStatus.Description
+	var setStatus *span.Status
+	if cfg.SetStatus != nil {
+		setStatus = &span.Status{
+			Code:        cfg.SetStatus.Code,
+			Description: cfg.SetStatus.Description,
+		}
 	}
 
-	toAttributes := &span.ToAttributes{}
-	if cfg.Rename.ToAttributes == nil {
-		toAttributes = nil
-	} else {
-		toAttributes.Rules = cfg.Rename.ToAttributes.Rules
-		toAttributes.BreakAfterMatch = cfg.Rename.ToAttributes.BreakAfterMatch
+	var toAttributes *span.ToAttributes
+	if cfg.Rename.ToAttributes != nil {
+		toAttributes = &span.ToAttributes{
+			Rules:           cfg.Rename.ToAttributes.Rules,
+			BreakAfterMatch: cfg.Rename.ToAttributes.BreakAfterMatch,
+		}
 	}
 
 	return &span.Arguments{
--- a/converter/internal/otelcolconvert/converter_spanprocessor.go
+++ b/converter/internal/otelcolconvert/converter_spanprocessor.go
@@ -9,6 +9,7 @@ import (
 	"github.com/grafana/agent/converter/internal/common"
 	"github.com/open-telemetry/opentelemetry-collector-contrib/processor/spanprocessor"
 	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/pdata/plog"
 )
 
 func init() {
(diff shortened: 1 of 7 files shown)


converter/internal/otelcolconvert/converter_spanprocessor.go
```diff
@@ -0,0 +1,139 @@
...
+func toMatchProperties(cfg map[string]any) *otelcol.MatchProperties {
+	if cfg == nil {
+		return nil
+	}
+
+	var regexpConfig *otelcol.RegexpConfig
+	if cfg["regexp_config"] != nil {
+		regexpConfig.CacheEnabled = cfg["regexp_config"].(map[string]any)["cache_enabled"].(bool)
+		regexpConfig.CacheMaxNumEntries = cfg["regexp_config"].(map[string]any)["cache_max_num_entries"].(int)
+	}
```
rfratto, line 86: This is a little gnarly to read. I know it'll be a small pain, but I think having a `toRegexpConfig` function might help here to avoid the double-type-cast in a single expression (even though there will still be two type casts, just spread out across functions)

Ditto for log severity below and anywhere else we're doing a string of more than one type casting in a single expression

converter/internal/otelcolconvert/converter_spanprocessor.go
```diff
@@ -0,0 +1,139 @@
...
+	if cfg["log_severity"] != nil {
+		ls.Min = otelcol.SeverityLevel(cfg["log_severity"].(map[string]any)["min"].(string))
+		ls.MatchUndefined = cfg["log_severity"].(map[string]any)["match_undefined"].(bool)
+	}
+	a := cfg["attributes"]
+	attributes := toOtelcolAttributes(encodeMapslice(a))
+	r := cfg["resources"]
+	resources := toOtelcolAttributes(encodeMapslice(r))
+	l := cfg["libraries"]
+	libraries := toOtelcolInstrumentationLibrary(encodeMapslice(l))
```
rfratto, line 98: tiny nit: why are we moving these out to different variables instead of inlining them?

```suggestion
	attributes := toOtelcolAttributes(encodeMapslice(cfg["attributes"]))
	resources := toOtelcolAttributes(encodeMapslice(cfg["resources"))
	libraries := toOtelcolInstrumentationLibrary(encodeMapslice(cfg["libraries"]))
```

converter/internal/otelcolconvert/testdata/span.yaml
```diff
@@ -0,0 +1,33 @@
...
+  otlp:
+    # Our defaults have drifted from upstream, so we explicitly set our
+    # defaults below (balancer_name and queue_size).
+    endpoint: database:4317
+    balancer_name: pick_first
+    sending_queue:
+      queue_size: 5000
+
+processors:
+  span:
```
rfratto, line 17: Given the amount of type casting being down here, and it's not obvious to me how much of the code paths are being hit by default, maybe we should explicitly define all the options here so we have more code coverage? 

(Probably would need a comment justifying why this test case has so many options when the others just use the default settings)

rfratto (commented):
Thanks, just a few small comments

converter/internal/otelcolconvert/testdata/span.yaml
tpaschalis: Yeah, I tested a complete configuration manually just to verify I didn't get any panics.

Will update the testdata file, add a relevant comment and clean up the conversions.

converter/internal/otelcolconvert/converter_spanprocessor.go
tpaschalis: I've went ahead and inlined them directly onto the final struct.

rfratto (approved):

PR #6505, opened 2024-02-23
otelcolconvert: support converting loki exporter

tpaschalis:
Fixes grafana/alloy#243

### DIFF
--- a/converter/internal/otelcolconvert/converter_lokiexporter.go
+++ b/converter/internal/otelcolconvert/converter_lokiexporter.go
@@ -1,0 +1,126 @@
+package otelcolconvert
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+
+	"github.com/grafana/agent/component/common/config"
+	"github.com/grafana/agent/component/loki/write"
+	"github.com/grafana/agent/component/otelcol/exporter/loki"
+	"github.com/grafana/agent/converter/diag"
+	"github.com/grafana/agent/converter/internal/common"
+	"github.com/grafana/river/rivertypes"
+	pconfig "github.com/prometheus/common/config"
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/config/configtls"
+
+	"github.com/open-telemetry/opentelemetry-collector-contrib/exporter/lokiexporter"
+)
+
+func init() {
+	converters = append(converters, lokiExporterConverter{})
+}
+
+type lokiExporterConverter struct{}
+
+func (lokiExporterConverter) Factory() component.Factory {
+	return lokiexporter.NewFactory()
+}
+
+func (lokiExporterConverter) InputComponentName() string { return "otelcol.exporter.loki" }
+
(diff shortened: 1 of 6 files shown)


converter/internal/otelcolconvert/converter_lokiexporter.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	block2 := common.NewBlockWithOverride([]string{"loki", "write"}, label, args2)
+
+	diags.Add(
+		diag.SeverityLevelInfo,
+		fmt.Sprintf("Converted %s into %s", stringifyInstanceID(id), stringifyBlock(block2)),
+	)
+	diags.Add(
+		diag.SeverityLevelInfo,
+		fmt.Sprintf("Created a loki.write block with a best-effort conversion of the lokiexporter's confighttp, retry and queue configuration settings. You may want to double check the converted configuration as most fields do not have a 1:1 match"),
```
rfratto: nit: This makes it sound like a different event from the above diagnostic; can we use `diags.AddWithDetail` on the first one and add the best-effort notice as the detail line?

converter/internal/otelcolconvert/converter_lokiexporter.go
```diff
@@ -0,0 +1,126 @@
...
+		ForwardTo: toTokenizedLogsReceivers(ids),
+	}
+}
+
+func toLokiWrite(name string, cfg *lokiexporter.Config) (*write.Arguments, error) {
+	// Defaults for MaxStreams and WAL should be handled on the Flow side.
+	res := &write.Arguments{}
+
+	if cfg.Endpoint != "" {
+		// TODO(@tpaschalis) Wire in auth from auth extension.
```
rfratto, line 85: (Just a note) 

😬 Handling auth here will be tricky, because we can't pass `otelcol.auth.basic` to `loki.write`; we'll have to actually probe the auth extension that lokiexporter was using and copy settings from the definition of the auth extension into `loki.write`.

tpaschalis, line 85: Not sure if there's a great way to do this yet.

rfratto (commented):
This is sick! I love seeing a converter which takes an interpretation of the upstream OpenTelemetry Collector components and tries to convert them best-effort into the native equivalents. This is exactly the kind of stuff I thought we'd do to make migrating to Flow easy, and I think we should have this. 

That being said, I'm wondering if we should put this on hold until after the more literal components converters are finished (e.g., converters for otelcol components where we have 1:1 equivalents). Especially considering my comment about what will be necessary to handle with auth, it's not clear to me how much remaining dev time it would take to fully finish this, and whether the literal components have a higher priority. 

It's up to you what you want to do; I just wanted to raise the idea that maybe this type of converter falls into a different category that may have a different priority.

converter/internal/otelcolconvert/converter_lokiexporter.go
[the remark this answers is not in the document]
erikbaranowski: 👍 definitely a warning at minimum

rfratto:
I'm going to close this for now. We can listen to feedback to see if this is still needed and if so, port this to Alloy with feedback addressed.

PR #6512, opened 2024-02-23
Update generated Compatible Components topic to remove relrefs

clayton-cornell:
#### PR Description

This PR updates the doc generation tooling to use relative URL links instead of relrefs. See: https://grafana.com/docs/writers-toolkit/write/shortcodes/#relref  Relrefs will cause a problem when this topic is mounted in Grafana Cloud. Relative URLs will work better, especially in this generated topic with more than 100 generated links.

Changes:
 - Updated Go files used to generate links.
 - Regenerated all topics with `make generate-docs`
 - Minor tweaks to various files to use semantic line breaks, clean up tables, and fix up a few minor style/grammar issues I spotted while reviewing. This is not an in-depth style review, so some style and grammar issues remain.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


docs/generator/links_to_types.go
```diff
@@ -84,7 +84,7 @@ func outputComponentsSection(name string, meta metadata.Metadata) string {
 	for _, outputDataType := range meta.AllTypesExported() {
 		if list := allComponentsThatAccept(outputDataType); len(list) > 0 {
 			section += fmt.Sprintf(
-				"- Components that consume [%s]({{< relref \"../compatibility/%s\" >}})\n",
+				"- Components that consume [%s](../compatibility%s)\n",
```
jdbaldry, line 87: We need to go up one more directory to get to the parent of `compatibility` from these component pages.

I think we've lost the trailing slash accidentally. Although it was not desirable in the `relref` (https://grafana.com/docs/writers-toolkit/review/doc-validator/errata/#relref-has-trailing-slash), it avoids a redirect if you are using URLs instead.

```suggestion
				"- Components that consume [%s](../../compatibility/%s)\n",
```

docs/generator/links_to_types.go
```diff
@@ -101,7 +101,7 @@ func acceptingComponentsSection(componentName string, meta metadata.Metadata) st
 	for _, acceptedDataType := range meta.AllTypesAccepted() {
 		if list := allComponentsThatExport(acceptedDataType); len(list) > 0 {
 			section += fmt.Sprintf(
-				"- Components that export [%s]({{< relref \"../compatibility/%s\" >}})\n",
+				"- Components that export [%s](../compatibility%s)\n",
```
jdbaldry, line 104: We need to go up one more directory to get to the parent of `compatibility` from these component pages.

I think we've lost the trailing slash accidentally. Although it was not desirable in the `relref` (https://grafana.com/docs/writers-toolkit/review/doc-validator/errata/#relref-has-trailing-slash), it avoids a redirect if you are using URLs instead.

```suggestion
				"- Components that export [%s](../../compatibility/%s)\n",
```

docs/generator/links_to_types.go
clayton-cornell: Ooops I missed testing that link in each component topic... I focused just on `compatibility/_index.md`  Fixed and tested.

rfratto (approved):

PR #6513, opened 2024-02-24
feat(otel/proc/k8sattr): sync default configs for `k8sattributesprocessor` to upstream

### LINKED ISSUE #6499: Sync defaults in otelcol.processor.k8sattributes
The upstream processor introduces some default excludes that we don't have set up in otelcol.processor.k8sattributes

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/ee728f70e403a4f3347a30be5fc2944ab17ce8e4/processor/k8sattributesprocessor/factory.go#L21

hainenber:

Fixes #6499 

#### Notes to the Reviewer

I don't see any default arguments for the component so my approach is just to skip converting these default excludes to static configurations. Do let me know if this is sensible enough :D

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


docs/sources/flow/reference/components/otelcol.processor.k8sattributes.md
```diff
@@ -225,6 +225,8 @@ Name | Type     | Description
 
 The `exclude` block configures which pods to exclude from the processor.
 
+> **Note**: Pods with name `jaeger-agent` and `jaeger-collector` are excluded by default.
```
clayton-cornell, line 228:
```suggestion
{{< admonition type="note" >}}
Pods with the name `jaeger-agent` or `jaeger-collector` are excluded by default.
{{< /admonition >}}
```

converter/internal/otelcolconvert/converter_k8sattributesprocessor.go
```diff
@@ -80,6 +80,11 @@ func toExclude(cfg k8sattributesprocessor.ExcludeConfig) k8sattributes.ExcludeCo
 	}
 
 	for _, c := range cfg.Pods {
+		// These are default excludes from upstream opentelemetry-collector-contrib, we don't have to convert them.
+		// Source: https://github.com/open-telemetry/opentelemetry-collector-contrib/blame/main/processor/k8sattributesprocessor/factory.go#L21
+		if c.Name == "jaeger-agent" || c.Name == "jaeger-collector" {
+			continue
+		}
```
rfratto, line 87: I don't think this is quite right; my read on what upstream is doing is that if excludes _aren't_ set, then it defaults to `jaeger-agent` and `jaeger-collector`, not that they are always excluded.

hainenber: Yes, I misinterpreted the intent and made changes in wrong place as well. I've made changes in the component in HEAD instead. PTAL if you have time to do so :D

internal/component/otelcol/processor/k8sattributes/types.go
```diff
@@ -157,6 +157,20 @@ func (args ExcludeConfig) convert() map[string]interface{} {
...
+	if len(pods) == 0 {
+		pods = append(pods,
+			map[string]interface{}{
+				"name": "jaeger-agent",
+			},
+			map[string]interface{}{
+				"name": "jaeger-collector",
+			},
+		)
+	}
```
rfratto, line 172: I think this works, but it doesn't really align with how we handle defaulting in other components. You should be able to implement the river.Defaulter interface on the [Arguments type](https://github.com/grafana/agent/blob/4cd7f745d9411130abc783ec3c82674eba9cb6e6/internal/component/otelcol/processor/k8sattributes/k8sattributes.go#L34) to set defaults like so:

```go 
func (args *Arguments) SetToDefault() {
  args.Exclude = ExcludeConfig{
    Pods: []ExcludePodConfig{
      {Name: "jaeger-agent"},
      {Name: "jaeger-collector"}, 
    },
  }
}
```

Let me know what you think :)

hainenber: I implemented it in previous manner as I think `SetToDefault()` would only be tested via E2E/Integration. Boy that was horribly wrong :D.

Thank you for informing me of this pattern. I'll remember well next time for existing/new components!

Btw, I've made the proposed change in the HEAD :D

rfratto (approved):
LGTM! Will merge after we get a 👍 from @clayton-cornell

docs/sources/flow/reference/components/otelcol.processor.k8sattributes.md
clayton-cornell: I don't see this suggestion applied.  Notes should not be formatted as a block quote. They are not correctly styled in the generated web page unless we use the `admonition` shortcode.

clayton-cornell (approved):
All good for docs. We are ready for merge @rfratto

PR #6514, opened 2024-02-24
fix(*): prevent  `duplicate metrics collector registration` panic for components with possible surface area

hainenber:
#### PR Description

I saw many [issues with `duplicate metrics registration` panics](https://github.com/grafana/agent/issues?q=is%3Aissue+%22duplicate+metrics+collector%22+is%3Aclosed) and saw that there are many places in the codebase not having such guardrails. The PR helps fixing that and would hopefully prevent any similar issues from happening.

N/A

#### PR Checklist

- [x] CHANGELOG.md updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -122,6 +122,8 @@ v0.40.0-rc.1 (2024-02-23)
 
 - Fix an issue where the configuration of the `http` and `remotecfg` blocks get ignored after loading a module. (@erikbaranowski)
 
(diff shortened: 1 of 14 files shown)


ptodev:
I'm not sure if reusing a previously registered metric is something we should always do. Failing to register can be a sign of a bug, so being able to catch the issue is a good thing. Using the AlreadyRegistered pattern would prevent us from catching such bugs. There was a similar discussion recently in the Prometheus repo: https://github.com/prometheus/prometheus/pull/13610

PR #6521, opened 2024-02-26
otelcolconvert: support converting transform processor

### LINKED ISSUE #6456: Support converting `transformprocessor`

tpaschalis:
Closes #6456.

I've tried this with the full example configuration available over at [the upstream processor's README](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/transformprocessor) and it converted without issues.

It required adding the same helpers as in #6504 so there might a temporary conflict.

<details>

```
transform:
  error_mode: ignore
  trace_statements:
    - context: resource
      statements:
        - keep_keys(attributes, ["service.name", "service.namespace", "cloud.region", "process.command_line"])
        - replace_pattern(attributes["process.command_line"], "password\\=[^\\s]*(\\s?)", "password=***")
        - limit(attributes, 100, [])
        - truncate_all(attributes, 4096)
    - context: span
      statements:
        - set(status.code, 1) where attributes["http.path"] == "/health"
        - set(name, attributes["http.route"])
        - replace_match(attributes["http.target"], "/user/*/list/*", "/user/{userId}/list/{listId}")
        - limit(attributes, 100, [])
        - truncate_all(attributes, 4096)

  metric_statements:
    - context: resource
      statements:
      - keep_keys(attributes, ["host.name"])
      - truncate_all(attributes, 4096)
    - context: metric
      statements:
        - set(description, "Sum") where type == "Sum"
    - context: datapoint
      statements:
        - limit(attributes, 100, ["host.name"])
        - truncate_all(attributes, 4096)
        - convert_sum_to_gauge() where metric.name == "system.processes.count"
        - convert_gauge_to_sum("cumulative", false) where metric.name == "prometheus_metric"
        
  log_statements:
    - context: resource
      statements:
        - keep_keys(attributes, ["service.name", "service.namespace", "cloud.region"])
    - context: log
      statements:
        - set(severity_text, "FAIL") where body == "request failed"
        - replace_all_matches(attributes, "/user/*/list/*", "/user/{userId}/list/{listId}")
        - replace_all_patterns(attributes, "value", "/account/\\d{4}", "/account/{accountId}")
        - set(body, attributes["http.route"])
```

</details>

### DIFF
--- a/internal/component/otelcol/processor/transform/transform.go
+++ b/internal/component/otelcol/processor/transform/transform.go
@@ -53,7 +53,7 @@ func (c *ContextID) UnmarshalText(text []byte) error {
 	}
 }
 
-type contextStatementsSlice []ContextStatements
+type ContextStatementsSlice []ContextStatements
 
 type ContextStatements struct {
 	Context    ContextID `river:"context,attr"`
@@ -64,9 +64,9 @@ type ContextStatements struct {
 type Arguments struct {
 	// ErrorMode determines how the processor reacts to errors that occur while processing a statement.
 	ErrorMode        ottl.ErrorMode         `river:"error_mode,attr,optional"`
-	TraceStatements  contextStatementsSlice `river:"trace_statements,block,optional"`
-	MetricStatements contextStatementsSlice `river:"metric_statements,block,optional"`
-	LogStatements    contextStatementsSlice `river:"log_statements,block,optional"`
+	TraceStatements  ContextStatementsSlice `river:"trace_statements,block,optional"`
+	MetricStatements ContextStatementsSlice `river:"metric_statements,block,optional"`
+	LogStatements    ContextStatementsSlice `river:"log_statements,block,optional"`
 
 	// Output configures where to send processed data. Required.
 	Output *otelcol.ConsumerArguments `river:"output,block"`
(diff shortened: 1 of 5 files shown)


component/otelcol/processor/transform/transform.go
```diff
@@ -53,9 +53,9 @@ func (c *ContextID) UnmarshalText(text []byte) error {
 	}
 }
 
-type contextStatementsSlice []contextStatements
+type contextStatementsSlice []ContextStatements
```
rfratto, line 56: IMO this type should be exported too; I think it's a mistake that it's not already.

converter/internal/otelcolconvert/testdata/transform.yaml
```diff
@@ -0,0 +1,33 @@
+receivers:
+  otlp:
+    protocols:
+      grpc:
+      http:
+
+processors:
+  transform:
```
rfratto, line 8: Can we use a real-world example here? This is unlikely to be a common usage of the transform processor, I feel

rfratto (approved):
LGTM % two pieces of feedback, sorry it took so long to review this

converter/internal/otelcolconvert/testdata/transform.yaml
tpaschalis: Yeah, I've added the example from the [upstream component's README](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/transformprocessor/README.md).

PR #6523, opened 2024-02-26
Feature stability gating on a component level

### LINKED ISSUE #6063: Proposal: gate non-stable features behind a flag
### Background

Non-stable features are documented as such, but there's nothing preventing a user from using the feature. This can be a problem if a user didn't realize the feature was non-stable, and would either prefer to only use stable features or are running the agent in a corporate environment where non-stable features aren't permitted. 

This proposal outlines an approach to help users avoid accidentally using features with an undesired stability level.

### Proposal

I propose a new flag to `grafana-agent run` which specifies the minimum stability of features that can run in that process:

```
--stability.level=experimental # allow usage of experimental, beta, and stable features 
--stability.level=beta         # allow usage of beta and stable features 
--stability.level=stable       # allow usage of stable features 
```

If a Flow config file or any loaded module uses a feature at a level lower than the minimum stability level, that config file or module should fail to load. 

As part of this proposal, code will need to be updated to reflect the stability level of various things. To start with, the following will support stability levels:

* Components
* Component blocks 
* Component attributes 
* Services 

In the future, the stability concept may be extended to include language-level features or standard library functions in River.

If this proposal is accepted, the default value should be `--stability.level=experimental` for backwards compatibility with existing users. The default value should be changed to `--stability.level=stable` alongside the eventual 1.0 release.

thampiotr:
#### PR Description

- Adds the CLI flag that allows to set the minimum stability level for the features that the agent must enforce.
- For now this is on an entire component level only, but is possible to extend in the future to finer-grained features.
- The component's stability levels must be provided at registration. If they're not, the tests will fail.
- Default min stability level for now is experimental.
- Users get a nice error when they use wrong stability levels. Errors when loading modules or reloading config are handled the same way as other invalid config errors.

Fixes #6063 

#### Notes to the Reviewer

How errors look when stability level is not allowed:
image

How help message looks like:
```
      --server.http.ui-path-prefix string      Prefix to serve the HTTP UI at (default "/")
      --stability.level <stability_level>      Minimum stability level of features to enable. Supported values: "stable", "beta", "experimental" (default "experimental")
      --storage.path string                    Base directory where components can store data (default "data-agent/")
```

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/component/otelcol/connector/host_info/host_info.go
+++ b/component/otelcol/connector/host_info/host_info.go
@@ -8,6 +8,7 @@ import (
 	"github.com/grafana/agent/component"
 	"github.com/grafana/agent/component/otelcol"
 	"github.com/grafana/agent/component/otelcol/connector"
+	"github.com/grafana/agent/internal/featuregate"
 	"github.com/grafana/river"
 	otelcomponent "go.opentelemetry.io/collector/component"
 	otelextension "go.opentelemetry.io/collector/extension"
@@ -15,9 +16,10 @@ import (
 
 func init() {
 	component.Register(component.Registration{
-		Name:    "otelcol.connector.host_info",
-		Args:    Arguments{},
-		Exports: otelcol.ConsumerExports{},
+		Name:      "otelcol.connector.host_info",
+		Stability: featuregate.StabilityStable,
+		Args:      Arguments{},
+		Exports:   otelcol.ConsumerExports{},
 
 		Build: func(opts component.Options, args component.Arguments) (component.Component, error) {
 			fact := NewFactory()
--- a/component/otelcol/processor/resourcedetection/resourcedetection.go
+++ b/component/otelcol/processor/resourcedetection/resourcedetection.go
@@ -22,6 +22,7 @@ import (
 	kubernetes_node "github.com/grafana/agent/component/otelcol/processor/resourcedetection/internal/k8snode"
 	"github.com/grafana/agent/component/otelcol/processor/resourcedetection/internal/openshift"
 	"github.com/grafana/agent/component/otelcol/processor/resourcedetection/internal/system"
+	"github.com/grafana/agent/internal/featuregate"
 	"github.com/grafana/river"
 	"github.com/mitchellh/mapstructure"
 	"github.com/open-telemetry/opentelemetry-collector-contrib/processor/resourcedetectionprocessor"
@@ -31,9 +32,10 @@ import (
 
 func init() {
 	component.Register(component.Registration{
-		Name:    "otelcol.processor.resourcedetection",
-		Args:    Arguments{},
-		Exports: otelcol.ConsumerExports{},
+		Name:      "otelcol.processor.resourcedetection",
+		Stability: featuregate.StabilityStable,
+		Args:      Arguments{},
+		Exports:   otelcol.ConsumerExports{},
 
 		Build: func(opts component.Options, args component.Arguments) (component.Component, error) {
 			fact := resourcedetectionprocessor.NewFactory()
--- a/component/pyroscope/java/java.go
+++ b/component/pyroscope/java/java.go
@@ -12,6 +12,7 @@ import (
 	"github.com/grafana/agent/component"
 	"github.com/grafana/agent/component/pyroscope"
 	"github.com/grafana/agent/component/pyroscope/java/asprof"
+	"github.com/grafana/agent/internal/featuregate"
 	"github.com/grafana/agent/pkg/flow/logging/level"
 )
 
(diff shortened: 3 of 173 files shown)


pkg/flow/internal/controller/component_registry.go
```diff
@@ -1,27 +1,55 @@
...
 type RegistryMap map[string]component.Registration
 
+var _ ComponentRegistry = (*RegistryMap)(nil)
+
 // Get retrieves a component using [component.Get].
-func (m RegistryMap) Get(name string) (component.Registration, bool) {
+func (m RegistryMap) Get(name string) (component.Registration, error) {
 	reg, ok := m[name]
-	return reg, ok
+	if !ok {
```
mattdurham, line 51: I dislike having this not check stability since its exported. Thoughts on making it an actual struct and passing in a stability of experimental. It looks like its only used in two places so the blast radius is pretty small.

mattdurham (approved):
One small comment about registrymap but looks good otherwise.

pkg/flow/internal/controller/component_registry.go
thampiotr: It's only used in tests so I didn't change it, but I can without too much effort.

component/otelcol/connector/host_info/host_info.go
```diff
@@ -8,16 +8,18 @@ import (
...
 	otelextension "go.opentelemetry.io/collector/extension"
 )
 
 func init() {
 	component.Register(component.Registration{
-		Name:    "otelcol.connector.host_info",
-		Args:    Arguments{},
-		Exports: otelcol.ConsumerExports{},
+		Name:      "otelcol.connector.host_info",
+		Stability: featuregate.StabilityStable,
```
rfratto, line 20: @thampiotr The docs say this is experimental, not stable: https://grafana.com/docs/agent/latest/flow/reference/components/otelcol.connector.host_info/

component/otelcol/processor/resourcedetection/resourcedetection.go
```diff
@@ -31,9 +32,10 @@ import (
 
 func init() {
 	component.Register(component.Registration{
-		Name:    "otelcol.processor.resourcedetection",
-		Args:    Arguments{},
-		Exports: otelcol.ConsumerExports{},
+		Name:      "otelcol.processor.resourcedetection",
+		Stability: featuregate.StabilityStable,
```
rfratto, line 36: @thampiotr The docs say this is beta, not stable: https://grafana.com/docs/agent/latest/flow/reference/components/otelcol.processor.resourcedetection/

component/pyroscope/java/java.go
```diff
@@ -21,8 +22,9 @@ const (
 
 func init() {
 	component.Register(component.Registration{
-		Name: "pyroscope.java",
-		Args: Arguments{},
+		Name:      "pyroscope.java",
+		Stability: featuregate.StabilityStable,
```
rfratto, line 26: @thampiotr The docs say this is beta, not stable: https://grafana.com/docs/agent/latest/flow/reference/components/pyroscope.java/

component/pyroscope/java/java_stub.go
```diff
@@ -6,13 +6,15 @@ import (
...
+	"github.com/grafana/agent/internal/featuregate"
 	"github.com/grafana/agent/pkg/flow/logging/level"
 )
 
 func init() {
 	component.Register(component.Registration{
-		Name: "pyroscope.java",
-		Args: Arguments{},
+		Name:      "pyroscope.java",
+		Stability: featuregate.StabilityStable,
```
rfratto, line 16: ditto above; this is beta, not stable

pkg/flow/internal/controller/loader_test.go
```diff
@@ -143,6 +149,24 @@ func TestLoader(t *testing.T) {
...
+		diags := applyFromContent(t, l, []byte(testFile), nil)
+		require.ErrorContains(t, diags.ErrorOrNil(), "component \"testcomponents.tick\" is at stability level \"beta\", which is below the minimum allowed stability level \"stable\"")
+	})
+
+	t.Run("Load with undefined minimum stability level", func(t *testing.T) {
+		l := controller.NewLoader(newLoaderOptionsWithStability(featuregate.StabilityUndefined))
+		diags := applyFromContent(t, l, []byte(testFile), nil)
+		require.ErrorContains(t, diags.ErrorOrNil(), "stability levels must be defined: got \"beta\" as stability of component \"testcomponents.tick\" and <invalid_stability_level> as the minimum stability level")
+	})
+
```
rfratto, line 169: Can you also add test cases for modules using a component with a disallowed stability level so we can make sure that the setting properly propagates to children controllers?

rfratto (commented):
Sorry for the post-merge review, but there were some things I noticed

PR #6527, opened 2024-02-26
Convert tailsampling processor

### LINKED ISSUE #6437: Support converting `tailsamplingprocessor`

erikbaranowski:

Closes #6437

#### PR Checklist

- [ ] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/component/otelcol/processor/tail_sampling/types.go
+++ b/component/otelcol/processor/tail_sampling/types.go
@@ -86,15 +86,20 @@ type NumericAttributeConfig struct {
 	MinValue int64 `river:"min_value,attr"`
 	// MaxValue is the maximum value of the attribute to be considered a match.
 	MaxValue int64 `river:"max_value,attr"`
+	// InvertMatch indicates that values must not match against attribute values.
+	// If InvertMatch is true and Values is equal to '123', all other values will be sampled except '123'.
+	// Also, if the specified Key does not match any resource or span attributes, data will be sampled.
+	InvertMatch bool `river:"invert_match,attr,optional"`
 }
 
 func (numericAttributeConfig NumericAttributeConfig) Convert() tsp.NumericAttributeCfg {
 	var otelConfig tsp.NumericAttributeCfg
 
 	mustDecodeMapStructure(map[string]interface{}{
-		"key":       numericAttributeConfig.Key,
-		"min_value": numericAttributeConfig.MinValue,
-		"max_value": numericAttributeConfig.MaxValue,
+		"key":          numericAttributeConfig.Key,
+		"min_value":    numericAttributeConfig.MinValue,
+		"max_value":    numericAttributeConfig.MaxValue,
+		"invert_match": numericAttributeConfig.InvertMatch,
 	}, &otelConfig)
 
 	return otelConfig
@@ -195,6 +200,7 @@ func (rateLimitingConfig RateLimitingConfig) Convert() tsp.RateLimitingCfg {
 type SpanCountConfig struct {
 	// Minimum number of spans in a Trace
 	MinSpans int32 `river:"min_spans,attr"`
+	MaxSpans int32 `river:"max_spans,attr"`
 }
 
 func (spanCountConfig SpanCountConfig) Convert() tsp.SpanCountCfg {
@@ -202,6 +208,7 @@ func (spanCountConfig SpanCountConfig) Convert() tsp.SpanCountCfg {
 
 	mustDecodeMapStructure(map[string]interface{}{
 		"min_spans": spanCountConfig.MinSpans,
+		"max_spans": spanCountConfig.MaxSpans,
 	}, &otelConfig)
 
 	return otelConfig
@@ -448,7 +455,8 @@ func (andSubPolicyConfig AndSubPolicyConfig) Convert() tsp.AndSubPolicyCfg {
 	return otelConfig
 }
 
-// TODO: Why do we do this? Can we not just create the Otel types directly?
+// mustDecodeMapStructure decodes a map into a structure. It panics if it fails.
+// This is necessary for otel types that have private fields such as sharedPolicyCfg.
 func mustDecodeMapStructure(source map[string]interface{}, otelConfig interface{}) {
 	err := mapstructure.Decode(source, otelConfig)
 
--- a/component/otelcol/processor/tail_sampling/types.go
+++ b/component/otelcol/processor/tail_sampling/types.go
@@ -68,13 +68,9 @@ type LatencyConfig struct {
 }
 
 func (latencyConfig LatencyConfig) Convert() tsp.LatencyCfg {
-	otelConfig := tsp.LatencyCfg{}
-
-	mustDecodeMapStructure(map[string]interface{}{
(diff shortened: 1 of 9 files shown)


component/otelcol/processor/tail_sampling/types.go
```diff
@@ -68,13 +68,9 @@ type LatencyConfig struct {
...
 
 func (latencyConfig LatencyConfig) Convert() tsp.LatencyCfg {
-	otelConfig := tsp.LatencyCfg{}
-
-	mustDecodeMapStructure(map[string]interface{}{
-		"threshold_ms": latencyConfig.ThresholdMs,
-	}, &otelConfig)
-
-	return otelConfig
+	return tsp.LatencyCfg{
```
erikbaranowski, line 71: removed unnecessary use of `mustDecodeMapStructure` here and elsewhere

component/otelcol/processor/tail_sampling/types.go
```diff
@@ -86,18 +82,19 @@ type NumericAttributeConfig struct {
 	MinValue int64 `river:"min_value,attr"`
 	// MaxValue is the maximum value of the attribute to be considered a match.
 	MaxValue int64 `river:"max_value,attr"`
+	// InvertMatch indicates that values must not match against attribute values.
+	// If InvertMatch is true and Values is equal to '123', all other values will be sampled except '123'.
+	// Also, if the specified Key does not match any resource or span attributes, data will be sampled.
+	InvertMatch bool `river:"invert_match,attr,optional"`
```
erikbaranowski, line 88: There are 2 new config options from otel I wired in as part of this PR. You will find them both documented in the component docs.

converter/internal/otelcolconvert/converter_helpers.go
```diff
@@ -48,7 +48,14 @@ func toTokenizedConsumers(components []componentID) []otelcol.Consumer {
 // in an internal package.
 func encodeMapstruct(v any) map[string]any {
 	var res map[string]any
-	if err := mapstructure.Decode(v, &res); err != nil {
+	var decoderConfig mapstructure.DecoderConfig = mapstructure.DecoderConfig{Squash: true, Result: &res}
```
erikbaranowski, line 51: I kept this in as we discussed in case there are squashed fields we need it for. Since we don't specifically need it yet I can remove this change if we prefer.

converter/internal/otelcolconvert/converter_tailsamplingprocessor.go
```diff
@@ -0,0 +1,241 @@
...
+			SharedPolicyConfig: toSharedPolicyConfig(cfg),
+			CompositeConfig:    toCompositeConfig(cfg.CompositeCfg),
+			AndConfig:          toAndConfig(cfg.AndCfg),
+		})
+	}
+	return out
+}
+
+func toSharedPolicyConfig(cfg tailsamplingprocessor.PolicyCfg) tail_sampling.SharedPolicyConfig {
+	return tail_sampling.SharedPolicyConfig{
```
erikbaranowski, line 82: This will look repeated down below but we are in fact getting the same fields on 3 different structs. I'm open to suggestion but this is probably not worth more complexity or effort to save a little code.

converter/internal/otelcolconvert/testdata/tail_sampling.river
```diff
@@ -0,0 +1,229 @@
+otelcol.receiver.otlp "default" {
```
erikbaranowski, line 1: This is taken right from the examples from otel: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/tailsamplingprocessor/README.md

converter/internal/otelcolconvert/converter_tailsamplingprocessor.go
```diff
@@ -0,0 +1,241 @@
...
+
+	state.Body().AppendBlock(block)
+	return diags
+}
+
+func toTailSamplingProcessor(state *state, id component.InstanceID, cfg *tailsamplingprocessor.Config) *tail_sampling.Arguments {
+	var (
+		nextMetrics = state.Next(id, component.DataTypeMetrics)
+		nextTraces  = state.Next(id, component.DataTypeTraces)
+		nextLogs    = state.Next(id, component.DataTypeLogs)
```
tpaschalis, line 50: The processor only works for traces, should make sure other signals are not taken into account?

erikbaranowski: good catch, I swear I checked this and still got it wrong 🤦

docs/sources/flow/reference/components/otelcol.processor.tail_sampling.md
```diff
@@ -229,6 +230,9 @@ The following arguments are supported:
 Name | Type | Description | Default | Required
 ---- | ---- | ----------- | ------- | --------
 `min_spans` | `number` | Minimum number of spans in a trace. | | yes
+`max_spans` | `number` | Maximum number of spans in a trace. | `0` | no
+
+The `max_spans` can be set to 0 to indicate max spans should not be considered.
```
clayton-cornell, line 235:
```suggestion
Set `max_spans` to `0`, if you do not want to limit the policy samples based on the maximum number of spans in a trace.
```

The suggestion is an attempt to make the sentence active and be explicitly clear what's happening. I'm not sure if I got the technical meaning correct though.

Also... do we need to adjust the text in line 226 and define the policy samples based on the maximum (upper limit?) number of spans in a trace? Is it an upper limit to the number of spans? What happens if you have more than X spans? Is the trace not sampled if the max is hit (similar to min)?

erikbaranowski: you've got it, if the max spans is not equal to 0 and the number of spans is greater than the max spans then it does not get sampled.

converter/internal/otelcolconvert/converter_tailsamplingprocessor.go
```diff
@@ -0,0 +1,237 @@
...
+		SubPolicyCfg:           toSubPolicyConfig(cfg.SubPolicyCfg),
+		RateAllocation:         toRateAllocationConfig(cfg.RateAllocation),
+	}
+}
+
+func toSubPolicyConfig(cfgs []tailsamplingprocessor.CompositeSubPolicyCfg) []tail_sampling.CompositeSubPolicyConfig {
+	var out []tail_sampling.CompositeSubPolicyConfig
+	for _, cfg := range cfgs {
+		out = append(out, tail_sampling.CompositeSubPolicyConfig{
+			SharedPolicyConfig: tail_sampling.SharedPolicyConfig{
```
rfratto, line 107: Should `cfg.AndCfg` be converted here? (CompositeSubPolicyCfg has a field for AndCfg)

rfratto (approved):
LGTM but there's one comment that should be looked at before merging

converter/internal/otelcolconvert/converter_tailsamplingprocessor.go
erikbaranowski: yes! really good catch on that one.

PR #6529, opened 2024-02-27
Modules import git folder

### LINKED ISSUE #6461: import.git: support importing folders as a single module
#5547 mentions that import.git should support importing folders, but it currently only supports importing single files. import.git should be updated to support folders.

IMO, this should not include folder support in remote.git, as that would require much more thinking beyond the use case of what we need for remote.git.

Related to #6460, which is the same issue but for import.file.

wildum:
#### PR Description

Add support for importing directories as modules via the import.git config block.

Fixes #6461 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


clayton-cornell (commented):
Docs look OK to me

docs/sources/flow/reference/config-blocks/import.git.md
```diff
@@ -48,8 +48,10 @@ with a `git clone REPOSITORY_ADDRESS` command, such as `https://github.com/grafa
 When provided, the `revision` attribute must be set to a valid branch, tag, or
 commit SHA within the repository.
 
-You must set the `path` attribute to a path accessible from the repository's root,
-such as `FILE_NAME.river` or `FOLDER_NAME/FILE_NAME.river`.
+You must set the `path` attribute to a path accessible from the repository's root.
+It can either be a river file such as `FILE_NAME.river` or `DIR_NAME/FILE_NAME.river` or
+a directory containing river files such as `DIR_NAME` or `.` if the river files are stored at the root
+of the repository.
```
rfratto, line 54: River is a proper noun:

```suggestion
You must set the `path` attribute to a path accessible from the repository's root.
It can either be a River file such as `FILE_NAME.river` or `DIR_NAME/FILE_NAME.river` or
a directory containing River files such as `DIR_NAME` or `.` if the River files are stored at the root
of the repository.
```

pkg/flow/internal/controller/node_config_import.go
rfratto, line 1: (I'm not going to review this, assuming that it'll be synced with the changes from #6502)

pkg/flow/internal/importsource/import_git.go
```diff
@@ -235,16 +236,42 @@ func (im *ImportGit) pollFile(ctx context.Context, args GitArguments) error {
...
+	// try to read as a directory
+	filesInfo, err := im.repo.ReadDir(args.Path)
 	if err != nil {
+		var pathErr *os.PathError
+		// if it's not a directory, then try to read as a file
+		if errors.As(err, &pathErr) && errors.Is(pathErr.Err, syscall.ENOTDIR) {
+			return im.handleFile(args.Path)
+		}
 		return err
 	}
```
rfratto, line 248: For consistency with import.file, I'd prefer if we could add a Stat method to `im.repo` and then call `im.repo.Stat(args.Path)` here to check if the path exists and if it's a file or directory.

rfratto (approved):
LGTM after #6502 is merged (there's more feedback there on the shared code) and this is rebased

docs/sources/flow/reference/config-blocks/import.git.md
clayton-cornell: I missed that in the review :-(

PR #6546, opened 2024-02-28
Fix stability label in loki kubernetes rules doc

wildum:
The stability label in the doc was different from the one in the code


docs/sources/flow/reference/components/loki.rules.kubernetes.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: loki.rules.kubernetes
 labels:
-  stage: beta
+  stage: experimental
```
rfratto, line 4: You need to change the line 9 too, which also references beta

rfratto (approved):

PR #6549, opened 2024-02-28
Improve converter diagnostic output

### LINKED ISSUE #6539: `convert` does not tell you when errors prevent generating a file
### What's wrong?

When running the `convert` subcommand in Flow, if there was a conversion error, it doesn't tell you that the conversion was a no-op because of the presence of errors. Instead, it prints the diagnostics list. 

If the diagnostic lists contains both error and info diagnostics, the user may not understand that the conversion has failed.

The workaround is to pass `--bypass-errors`, but this is not obvious from the command output.

I would expect to see a clear message at the bottom indicating that exporting failed due to the presence of one or more error-level diagnostics, and that I should provide `--bypass-errors`. (Or that there is a fatal issue in case of fatal diagnostics) 

### Steps to reproduce

1. Run the `convert` command with a file that should return errors. 
2. Examine the console output; noting that there is nothing indicating that the conversion has failed, other than the error-level diagnostics.

### System information

N/A

### Software version

v0.40.0

### Configuration

_No response_

### Logs

```text
Error: (Error) The converter does not support converting the provided rule_files config.
(Info) Converted scrape_configs job_name "prometheus" into...
	A prometheus.scrape.prometheus component
(Info) Converted scrape_configs job_name "node" into...
	A prometheus.scrape.node component
(Info) Converted scrape_configs job_name "influxtest" into...
	A prometheus.scrape.influxtest component
(Info) Converted scrape_configs job_name "examplehttp" into...
	A prometheus.scrape.examplehttp component
(Info) Converted scrape_configs job_name "alertmanager" into...
	A prometheus.scrape.alertmanager component
(Info) Converted scrape_configs job_name "dht22" into...
	A prometheus.scrape.dht22 component
(Info) Converted scrape_configs job_name "weatherflow" into...
	A prometheus.scrape.weatherflow component
(Info) Converted scrape_configs job_name "analyticsmetrics" into...
	A prometheus.scrape.analyticsmetrics component
(Info) Converted scrape_configs job_name "grafanametrics" into...
	A prometheus.scrape.grafanametrics component
(Info) Converted scrape_configs job_name "hue" into...
	A prometheus.scrape.hue component
(Info) Converted scrape_configs job_name "unifipoller" into...
	A prometheus.scrape.unifipoller component
(Info) Converted scrape_configs job_name "dump1090" into...
	A prometheus.scrape.dump1090 component
(Info) Converted scrape_configs job_name "network" into...
	A prometheus.scrape.network component
	A discovery.relabel.network component
(Info) Converted 2 remote_write[s] "," into...
	A prometheus.remote_write.default component
```

erikbaranowski:
by including a Footer and removing lower level diagnostics when a configuration fails to generate.

Fixes #6539

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated


converter/diag/report.go
```diff
@@ -6,14 +6,47 @@ import (
 
 const Text = ".txt"
 
-// generateTextReport generates a text report for the diagnostics.
-func generateTextReport(writer io.Writer, ds Diagnostics) error {
-	content := ds.Error()
+const criticalErrorFooter = `
```
erikbaranowski, line 9: Wordsmithing is welcome on these 3 if we like the concept of a footer under the diags. @clayton-cornell @rfratto 

Check out the new tests below for how this will print out in practice.

converter/diag/report.go
```diff
@@ -6,14 +6,47 @@ import (
...
+
+A configuration file was not generated due to critical issues. Review the diagnostics above.`
+
+const errorFooter = `
+
+A configuration file was not generated due to errors. Review the diagnostics above.
+
+These may be bypassed which is not recommended for production use.`
```
rfratto, line 17: Can this say how to bypass them?

converter/diag/report.go
```diff
@@ -6,14 +6,47 @@ import (
...
-// generateTextReport generates a text report for the diagnostics.
-func generateTextReport(writer io.Writer, ds Diagnostics) error {
-	content := ds.Error()
+const criticalErrorFooter = `
+
+A configuration file was not generated due to critical issues. Review the diagnostics above.`
+
+const errorFooter = `
+
+A configuration file was not generated due to errors. Review the diagnostics above.
```
clayton-cornell, line 15:
```suggestion
Grafana Agent couldn't generate a configuration file. Refer to the error message for more information.
```

If we switch things around, we make the information active vs passive, and are specific about who couldn't do what.

What is output? is it actually diagnostics output? or just an error message? We should try to avoid directions (above/below) and be explicit about where the user needs to look.

converter/diag/report.go
clayton-cornell:
```suggestion
You can bypass the errors by using the --bypass-errors flag. Bypassing errors isn't recommended for production environments.`
```

Maybe something like this (the second sentence is passive, but... hard to make active and keep it short)

rfratto (approved):

PR #6552, opened 2024-02-28
misc: move all importable code to internal/

rfratto:
Our [backwards compatibility guarantees](https://github.com/grafana/agent/blob/main/docs/rfcs/0008-backwards-compatibility.md) cover importable Go code. This applies once we cut the 1.0 release, which we are rapidly approaching. 

The Grafana Agent maintainers have agreed that it is more important for 1.0 to focus on stability of the configuration file and usage of the Grafana Agent binaries rather than on the stability of importable code. 

This PR moves all importable Go code to `internal/` so that it is no longer externally importable. This gives us time to finalize Go APIs without blocking the 1.0 release or removing our commitment to having Go APIs follow semantic versioning. Some time after the 1.0 release, we will have public Go APIs available for consumption. 

There are some slightly odd artifacts in this PR for the purposes of making sure we have no public Go API surface:

* Non-binary tools have been moved to `internal/tools`:
  * Code for generating documentation has been moved to `internal/tools/docs_generator`. 
  * Code for integration testing system packages has been moved to `internal/tools/packaging_test`. 
* Binaries (i.e., `package main`) not a part of our release assets have similarly moved to `internal/cmd`:
  * Code for integration testing Flow components has been moved to `internal/cmd/integration-tests` 
  * Code for linting the codebase has been moved to `internal/cmd/agentlint` 
  * Code for evaluating River files has been moved to `internal/cmd/rivereval`
* All flow mode UI code has been moved from `web` to `internal/web`. 

I don't know if all of these changes are necessary, or if they have been moved to a sensible location. However, for the scope of this PR, I would like to keep the folders in their new location, and follow-up PRs and discussions can be used to reorganize `internal/` as needed.


tpaschalis (approved):
📪 ▶️

thampiotr (approved):

PR #6558, opened 2024-02-29
Fix logging node buffer

### LINKED ISSUE #6542: 0.4.0-rc.2: Follow-up error on start-up: lokiWriter failed to forward entry, channel was blocked
### What's wrong?

I'm using import.http (new syntax in 0.40) together with `logging` and `loki.write`.

`logging`, `import.http` and `loki.write` are on the same local configuration. 

If there is an error in a imported component, I getting an error from the logging block, too. This may confused people.

### Steps to reproduce

1. setup config with `import.file` , `logging` and `loki.write`
2. have a syntax error or something else inside the imported module cause an `could not perform the initial load successfully` error

### System information

Linux

### Software version

Grafana Agent v0.40.0-rc.2

### Configuration

```text
Let me know, if I should provide one.
```

### Logs

```text
{"ts":"2024-02-28T09:36:39.878497528Z","level":"error","msg":"failed to evaluate config","controller_id":"all.config.this","trace_id":"7fe423b9ede149b939a284690834b4b3","node":"prometheus.remote_write.prometheus","err":"decoding River: all:107:17: unrecognized attribute name \"remote_username\""}
{"ts":"2024-02-28T09:36:39.878560328Z","level":"error","msg":"failed to evaluate config","controller_id":"","trace_id":"2b75bd4dd148b8c86d06d4c36fcadd83","node":"logging","err":"could not update logger: lokiWriter failed to forward entry, channel was blocked"}
Error: all:107:17: unrecognized attribute name "remote_username"

Error: /config/azure-server.river:48:1: Failed to evaluate node for config block: could not update logger: lokiWriter failed to forward entry, channel was blocked

47 |   
48 |   logging {
   |  _^^^^^^^^^
49 | |     level  = "warn"
50 | |     format = "json"
51 | |     write_to = [loki.write.remote.receiver]
52 | | }
   | |_^
53 |   loki.write "remote" {
interrupt received

Error: could not perform the initial load successfully
```

wildum:
#### PR Description

During the initial evaluation, the logging node is updated with the correct format and tries to log the data that it had buffered. This works for the standard output but it doesn't for the other receivers because the components that own them are not running yet (they are not ready to receive data from the receiver channel). This results in an error.

This fix ignores the error, which is the same behavior as before the introduction of the buffer via #5992 

This is still not ideal because the first logs won't reach the receivers. The buffer should allow us to only send the logs to the components when they are ready but this requires much more effort.

Fixes #6542

#### PR Checklist

- [x] CHANGELOG.md updated

### DIFF
--- a/pkg/flow/logging/logger.go
+++ b/pkg/flow/logging/logger.go
@@ -126,9 +126,9 @@ func (l *Logger) Update(o Options) error {
 
 	// Print out the buffered logs since we determined the log format already
 	for _, bufferedLogChunk := range l.buffer {
-		if err := slogadapter.GoKit(l.handler).Log(bufferedLogChunk...); err != nil {
-			return err
-		}
+		// the buffered logs are currently only sent to the standard output
+		// because the components with the receivers are not running yet
+		slogadapter.GoKit(l.handler).Log(bufferedLogChunk...)
 	}
 	l.buffer = nil
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ Main (unreleased)
 
 - A new `loki.rules.kubernetes` component that discovers `PrometheusRule` Kubernetes resources and loads them into a Loki Ruler instance. (@EStork09)
 
+### Bugfixes
+
+- Fix an issues where the logging config block would trigger an error when trying to send logs to components that were not running. (@wildum)
+
 v0.40.0 (2024-02-27)
 --------------------
 


rfratto (approved):

pkg/flow/logging/logger.go
```diff
@@ -126,9 +126,9 @@ func (l *Logger) Update(o Options) error {
 
 	// Print out the buffered logs since we determined the log format already
 	for _, bufferedLogChunk := range l.buffer {
-		if err := slogadapter.GoKit(l.handler).Log(bufferedLogChunk...); err != nil {
-			return err
-		}
+		// the buffered logs are currently only sent to the standard output
+		// because the components with the receivers are not running yet
```
tpaschalis, line 130: Can we rephrase that comment? It is an explanation of why we ignore the error, but isn't true for the entire rationale behind l.buffer.

wildum: it's true for the entire lifetime of the buffer because it's set to nil and wont be used anymore after this

wildum: > does it mean that the buffered logs will never reach a loki.write receiver, for example? 

yes because the logs are only buffered till the config node is evaluated, after that the logs are not buffered anymore. And no component can run before all nodes have been through the first evaluation cycle

PR #6559, opened 2024-02-29
Port promtail changes part 1

thampiotr:
#### PR Description

Ports https://github.com/grafana/loki/commit/6387bf765c374468b10c39d235dc135fdb95b50d


tpaschalis (approved):
LGTM! 🙌

component/common/loki/client/batch.go
```diff
@@ -201,3 +201,11 @@ func (b *batch) reportAsSentData(h SentDataMarkerHandler) {
...
 	}
 }
+
+func entrySize(entry logproto.Entry) int {
+	structuredMetadataSize := 0
+	for _, label := range entry.StructuredMetadata {
+		structuredMetadataSize += label.Size()
+	}
+	return len(entry.Line) + structuredMetadataSize
+}
```
tpaschalis, line 211: nit: do we know what `entry.Parsed` is used for? Should it also be included here?

thampiotr: I would like to merge it as-is, because my goal here is to just match the upstream: https://github.com/grafana/loki/blob/main/clients/pkg/promtail/client/batch.go#L150-L156 

But you have a valid question. I wonder why `Parsed` is not included in size or why the existing Size method is not used: https://github.com/grafana/loki/blob/main/pkg/push/types.go#L720 

I think @vlad-diachenko will be able to answer the above.

tpaschalis: I see. Yeah, I think we should move forward with this

PR #6564, opened 2024-02-29
Add test for creating a file in a path that is excluded

### LINKED ISSUE #6158: Port grafana/loki#9831 to loki.source.file
@cstyan pointed out that a new bugfix got merged to Promtail (grafana/loki#9831), and that we may need to port it. 

We need to investigate whether the bugfix needs to be ported, and port it if it's relevant. If it's not relevant to our code, we can close this issue.

thampiotr:
#### PR Description

It seems like `local.file_match` doesn't need the fix from https://github.com/grafana/loki/pull/9831. 
Adding a test to prove this and keep it working. 

Fixes #6158 

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/local/file_match/file_test.go
+++ b/internal/component/local/file_match/file_test.go
@@ -238,10 +238,14 @@ func TestMultiLabels(t *testing.T) {
 	require.True(t, contains([]discovery.Target{foundFiles[1]}, "t1.txt"))
 }
 
+// createComponent creates a component with the given paths and labels. The paths and excluded slices are zipped together
+// to create the set of targets to pass to the component.
 func createComponent(t *testing.T, dir string, paths []string, excluded []string) *Component {
 	return createComponentWithLabels(t, dir, paths, excluded, nil)
 }
 
+// createComponentWithLabels creates a component with the given paths and labels. The paths and excluded slices are
+// zipped together to create the set of targets to pass to the component.
 func createComponentWithLabels(t *testing.T, dir string, paths []string, excluded []string, labels map[string]string) *Component {
 	tPaths := make([]discovery.Target, 0)
 	for i, p := range paths {
--- a/internal/component/local/file_match/file_test.go
+++ b/internal/component/local/file_match/file_test.go
@@ -116,6 +116,43 @@ func TestAddingFileInSubDir(t *testing.T) {
 	require.True(t, contains(foundFiles, "t3.txt"))
 }
 
+func TestAddingFileInAnExcludedSubDir(t *testing.T) {
+	dir := path.Join(os.TempDir(), "agent_testing", "t3")
+	os.MkdirAll(dir, 0755)
+	writeFile(t, dir, "t1.txt")
+	t.Cleanup(func() {
+		os.RemoveAll(dir)
+	})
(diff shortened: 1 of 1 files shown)


internal/component/local/file_match/file_test.go
```diff
@@ -207,17 +244,13 @@ func createComponent(t *testing.T, dir string, paths []string, excluded []string
...
 		for k, v := range labels {
 			tar[k] = v
 		}
-		tPaths = append(tPaths, tar)
-	}
-	for _, p := range excluded {
-		tar := discovery.Target{"__path_exclude__": p}
-		for k, v := range labels {
-			tar[k] = v
+		if i < len(excluded) {
```
mattdurham, line 252: I am not sure I follow this logic. Why do paths and excluded rely on the same indexer?
IE
Included:
Path A: /tmp/a/**/*.txt
Path B /tmp/b

Excluded:
/tmp/b/exc/*.txt

Then the exclusion would get added to path a and not path b?

thampiotr: Let me explain via an example. Suppose you pass the following arguments to this function:
```
paths=[/tmp/a/**/*.txt, /tmp/b/**/*.txt], excluded=[/tmp/a/exc/*.txt]
```

Before my change, it would produce the following path targets - the `tPaths` variable:
```
{__path__=/tmp/a/**/*.txt}
{__path__=/tmp/b/**/*.txt}
{__path__exclude__=/tmp/b/exc/*.txt}
```

Each of these targets will run a watcher which will independently discover & exclude the files. So the `{__path__=/tmp/b/**/*.txt}` would include all the files, even the ones under `/tmp/b/exc/`. The exclusion is not effective.

My understanding is that in order to have exclusion working properly, one would need to create the following targets instead:
```
{__path__=/tmp/a/**/*.txt}
{__path__=/tmp/b/**/*.txt, __path__exclude__=/tmp/b/exc/*.txt}
```

And the change to using the same indexer achieves this: the included and excluded paths are effectively zipped together.

I think that these tests could use some refactor too, but I wanted to focus on verifying the potential bug ahead of release we're working on.

mattdurham: With this test the exclude for B is attached to A which, for tests it may not matter that its got some fragility but might need to be documented.

```
func TestCreateComponent(t *testing.T) {
	c := createComponentWithLabels(t, "tmp", []string{"tmp/a/**/*.txt", "tmp/b/**/*.txt"}, []string{"tmp/b/exc/*.txt"}, nil)
	assert.True(t, len(c.args.PathTargets) == 2)
	assert.True(t, len(c.args.PathTargets[0].Labels()) == 1)
	assert.True(t, len(c.args.PathTargets[1].Labels()) == 2)
}
```

mattdurham: image

tpaschalis (approved):
LGTM! (some whitespace would help with readability 😅)

thampiotr:
> LGTM! (some whitespace would help with readability 😅)

I agree, but I think that this test could use a bit of refactor anyway, to reduce repetition. I would prefer to do that in a follow-up :)

PR #6575, opened 2024-02-29
wip: add count connector

rlankfo:
#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/internal/component/otelcol/connector/connector.go
+++ b/internal/component/otelcol/connector/connector.go
@@ -18,15 +18,18 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	otelcomponent "go.opentelemetry.io/collector/component"
 	otelconnector "go.opentelemetry.io/collector/connector"
+	otelconsumer "go.opentelemetry.io/collector/consumer"
 	otelextension "go.opentelemetry.io/collector/extension"
 	sdkprometheus "go.opentelemetry.io/otel/exporters/prometheus"
 	"go.opentelemetry.io/otel/sdk/metric"
 )
 
+type Type uint16
+
 const (
-	ConnectorTracesToTraces = iota
-	ConnectorTracesToMetrics
-	ConnectorTracesToLogs
+	ConnectorTracesToTraces  Type = 1 << iota // 1
+	ConnectorTracesToMetrics                  // 2
+	ConnectorTracesToLogs                     // 4
 	ConnectorMetricsToTraces
 	ConnectorMetricsToMetrics
 	ConnectorMetricsToLogs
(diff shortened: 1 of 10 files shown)


internal/component/otelcol/connector/count/count.go
```diff
@@ -0,0 +1,105 @@
...
+	})
+}
+
+// Arguments configures the otelcol.connector.count component.
+type Arguments struct {
+	Spans      []MetricInfo `river:"span,block,optional"`
+	SpanEvents []MetricInfo `river:"spanevent,block,optional"`
+	Metrics    []MetricInfo `river:"metric,block,optional"`
+	DataPoints []MetricInfo `river:"datapoint,block,optional"`
+	Logs       []MetricInfo `river:"log,block,optional"`
```
ptodev, line 32: It would be good to write a test which uses all of these values at once. I wonder whether it would work. Issues such as #6329 might prevent the component from working.

internal/component/otelcol/connector/connector.go
```diff
@@ -33,6 +33,7 @@ const (
 	ConnectorLogsToTraces
 	ConnectorLogsToMetrics
 	ConnectorLogsToLogs
+	ConnectorAnyToMetrics
```
ptodev, line 36: I don't mind having a `ConnectorAnyToMetrics`, but I think it would be cleaner to use a bit field like [this one](https://github.com/grafana/agent/blob/0e36aa27c95b7d600be1722ba80d56e8227d048b/internal/component/otelcol/exporter/exporter.go#L47-L72). This will also prevent us from needing a separate branch in the `switch` statement below for every permutation of inputs and outputs.

Having it as part of this `iota` makes it feels like it's a completely different connector type, which is also not very accurate.

internal/component/otelcol/connector/count/count.go
rlankfo: added the `Test_ComponentIO` test and it seems to work

internal/component/otelcol/connector/connector.go
```diff
@@ -177,22 +196,44 @@ func (p *Connector) Update(args component.Arguments) error {
 	var metricsConnector otelconnector.Metrics
 	var logsConnector otelconnector.Logs
 
-	switch pargs.ConnectorType() {
-	case ConnectorTracesToMetrics:
+	ct := pargs.ConnectorType()
+	if ct.SupportsAnyToMetrics() {
```
ptodev, line 200: In theory a connector can be both AnyToMetrics and AnyToLogs, in which case we'd have to remove the `"this connector can only output metrics"` error below. But we can leave the code as it is for now... and think about those other cases as they arise. Flow is flexible enough so that we can rework our code in the future to accommodate other connector types.

internal/component/otelcol/connector/count/count.go
```diff
@@ -0,0 +1,195 @@
...
+	cfg, err := args.Convert()
+	if err != nil {
+		return err
+	}
+	return cfg.(*countconnector.Config).Validate()
+}
+
+// SetToDefault implements river.Defaulter.
+func (args *Arguments) SetToDefault() {
+	*args = DefaultArguments
```
ptodev, line 194: If there is more than one such connector in the config file, I wonder if there'd be a bug due to reusing the same `DefaultArguments`. E.g. if one connector overrides some defaults, but another one doesn't, then would the connector which only used defaults pick up the settings from the other one? I wonder if we should copy the slices with defaults every time we call `SetToDefault`

tpaschalis:
It's something for later on, when we're ready to merge, but let's make sure we also include a converter in the [otelconvert](https://github.com/grafana/agent/tree/main/internal/converter/internal/otelcolconvert) package? We can help with that.

internal/component/otelcol/connector/count/count.go
rlankfo: I can test this out. I noticed several other components following this pattern though, I'd assume they have the same issue if this is the case.

internal/component/otelcol/connector/connector.go
rlankfo: Yes, this makes sense. I thought about it as I was implementing this but decided not to tackle it yet as there isn't a use case for it yet.

PR #6583, opened 2024-03-02
otelconvert: extend framework for extension components.

tpaschalis:
> NOTE: This PR can be reviewed commit-by-commit.

This PR extends the current framework for converting OpenTelemetry Collector configs into Flow mode configs to support extensions.

I'm not 100% happy yet with how we're overloading some of the concepts used by the first framework commit, but I hope we get to a good place after a round or two.

As an addition, this PR converts the basicauth extension into the otelcol.auth.basic component and wires it in the current set of exporters that make use of GRPCClientArguments and HTTPClientArguments.

### DIFF
--- a/internal/component/otelcol/auth/auth.go
+++ b/internal/component/otelcol/auth/auth.go
@@ -45,32 +45,19 @@ type Arguments interface {
 type Exports struct {
 	// Handler is the managed component. Handler is updated any time the
 	// extension is updated.
-	Handler Handle `river:"handler,attr"`
+	Handler Handler `river:"handler,attr"`
 }
 
-type Handler interface {
-	ID() otelcomponent.ID
-	Extension() otelextension.Extension
+// Handler combines an extension with its ID.
+type Handler struct {
+	ID        otelcomponent.ID
+	Extension otelextension.Extension
 }
 
-// Handle combines an extension with its ID.
-type Handle struct {
-	Id  otelcomponent.ID
-	Ext otelextension.Extension
-}
-
-var _ river.Capsule = Handle{}
+var _ river.Capsule = Handler{}
 
 // RiverCapsule marks Handler as a capsule type.
-func (Handle) RiverCapsule() {}
-
-func (h Handle) ID() otelcomponent.ID {
-	return h.Id
-}
-
-func (h Handle) Extension() otelextension.Extension {
-	return h.Ext
-}
+func (Handler) RiverCapsule() {}
 
 // Auth is a Flow component shim which manages an OpenTelemetry Collector
 // authentication extension.
@@ -182,9 +169,9 @@ func (a *Auth) Update(args component.Arguments) error {
 
 	// Inform listeners that our handler changed.
 	a.opts.OnStateChange(Exports{
-		Handler: Handle{
-			Id:  otelcomponent.NewID(otelcomponent.Type(a.opts.ID)),
-			Ext: ext,
+		Handler: Handler{
+			ID:        otelcomponent.NewID(otelcomponent.Type(a.opts.ID)),
+			Extension: ext,
 		},
 	})
 
--- a/internal/component/otelcol/auth/auth.go
+++ b/internal/component/otelcol/auth/auth.go
@@ -45,19 +45,32 @@ type Arguments interface {
 type Exports struct {
 	// Handler is the managed component. Handler is updated any time the
 	// extension is updated.
-	Handler Handler `river:"handler,attr"`
+	Handler Handle `river:"handler,attr"`
 }
 
-// Handler combines an extension with its ID.
-type Handler struct {
-	ID        otelcomponent.ID
-	Extension otelextension.Extension
+type Handler interface {
+	ID() otelcomponent.ID
+	Extension() otelextension.Extension
 }
 
-var _ river.Capsule = Handler{}
+// Handle combines an extension with its ID.
+type Handle struct {
+	id        otelcomponent.ID
+	extension otelextension.Extension
+}
+
+var _ river.Capsule = Handle{}
 
 // RiverCapsule marks Handler as a capsule type.
-func (Handler) RiverCapsule() {}
+func (Handle) RiverCapsule() {}
+
+func (h Handle) ID() otelcomponent.ID {
+	return h.id
+}
+
+func (h Handle) Extension() otelextension.Extension {
+	return h.extension
+}
 
 // Auth is a Flow component shim which manages an OpenTelemetry Collector
 // authentication extension.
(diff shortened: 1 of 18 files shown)


internal/component/otelcol/auth/auth.go
```diff
@@ -45,19 +45,32 @@ type Arguments interface {
...
+	Handler Handle `river:"handler,attr"`
 }
 
-// Handler combines an extension with its ID.
-type Handler struct {
-	ID        otelcomponent.ID
-	Extension otelextension.Extension
+type Handler interface {
+	ID() otelcomponent.ID
+	Extension() otelextension.Extension
```
tpaschalis, line 53: All of this is for being able to extend `*auth.Handler` to make use of the same neat trick we use in `toTokenizedConsumers`.

An alternative would be to define a RiverTokenize method here, but it would really be out of place and it couldn't be called correctly outside of the otelconvert context.

internal/converter/internal/otelcolconvert/converter_basicauthextension.go
```diff
@@ -0,0 +1,48 @@
...
+	converters = append(converters, basicAuthConverterConverter{})
+}
+
+type basicAuthConverterConverter struct{}
+
+func (basicAuthConverterConverter) Factory() component.Factory {
+	return basicauthextension.NewFactory()
+}
+
+func (basicAuthConverterConverter) InputComponentName() string { return "otelcol.auth.basic" }
```
tpaschalis, line 24: We're potentially overloading the InputComponentName method with a different use case here; this is not meant to forward data to, but only to build valid references to the Flow component.

internal/converter/internal/otelcolconvert/otelcolconvert.go
```diff
@@ -167,6 +168,39 @@ func appendConfig(file *builder.File, cfg *otelcol.Config) diag.Diagnostics {
 	// twice?
 	converterTable := buildConverterTable()
 
+	extensionTable := make(map[component.ID]componentID, len(cfg.Service.Extensions))
+
+	for _, ext := range cfg.Service.Extensions {
```
tpaschalis, line 173: We build the list of extensions 'activated' (defined in the service) as Flow components and keep a mapping of their OTel IDs to the blocks we've built.

There's no concept of multiple extensions per group or telemetry signals. Should we explain what we're doing here and why a little better?

internal/component/otelcol/auth/auth.go
rfratto: Hm, I'm not sure about this change; I'm worried it makes the API for component authors more confusing. 

There's a way to avoid needing this, but it may require some extra thoughtfulness around changes to the converter API. In particular, `github.com/grafana/river/token/builder` accepts being given a [ValueOverrideHook](https://github.com/grafana/river/blob/main/token/builder/builder.go#L88-L93), which allows you to change values at tokenization time with something else. 

One could hypothetically do this:

```go
block.Body().SetValueOverrideHook(func(val any) any {
  switch val.(type) {
  case auth.Handler: // old struct type, not interface type in this PR
    // return something which implements builder.Tokenizer
    //
    // (if the component has more than one auth.Handler value, you
    // will need to do some kind of comparison before returning to 
    // build the proper tokens) 
  }
})
```

This is used in the converters for Prometheus; a similar thing can be done in the otelcol converters to avoid needing to change the component API.

tpaschalis: Ooh, I was looking around for other ways to achieve this, but didn't spot this, thanks for the hint!

tpaschalis: cc @rfratto this worked really easily in https://github.com/grafana/agent/pull/6583/commits/719f46e0906188269dd7a76e1b3f1643043af8a8 thanks for the suggestion!

internal/converter/internal/otelcolconvert/otelcolconvert.go
rfratto: $0.02: I think needing to explain this in a GitHub comment is usually an indicator that it should be a code comment :) The next person reading this code isn't going to be able to easily find the PR and this comment explaining why this is happening here.

internal/converter/internal/otelcolconvert/converter_otlphttpexporter.go
```diff
@@ -32,9 +34,19 @@ func (otlpHTTPExporterConverter) ConvertAndAppend(state *state, id component.Ins
 	var diags diag.Diagnostics
 
 	label := state.FlowComponentLabel()
+	overrideHook := func(val interface{}) interface{} {
+		switch val.(type) {
+		case auth.Handler:
+			if val != nil {
```
rfratto, line 40: Can this condition ever be true? Go interfaces are fat pointers, containing type information and a pointer to the underlying data. However, in this case,  the type information is a non-pointer type. I don't think it's possible for a nil interface to have type information denoting that it's storing a non-nil type.

rfratto: (I would find this less confusing if the nil check was moved before the switch statement)

internal/converter/internal/otelcolconvert/converter.go
```diff
@@ -177,6 +183,14 @@ func (state *state) nextInstances(c component.InstanceID, dataType component.Dat
 	}
 }
 
+func (state *state) lookupExtension(id component.ID) componentID {
```
rfratto, line 186: nit: lookupExtension should be an exported if converter implementations are expected to use it, since idiomatically other exported methods on `state` signal to converter implementations that it is a safe entrypoint to use.

rfratto (approved):
LGTM % final comments

internal/converter/internal/otelcolconvert/otelcolconvert.go
tpaschalis: Wrote this for myself but didn't follow up on it, thanks! 👀

internal/converter/internal/otelcolconvert/converter_basicauthextension.go
rfratto: Sorry, I missed this comment; yeah I think InputComponentName is a weird name given this. I don't want to introduce a different concept or method, but maybe we can rename the existing method to fit both cases?

internal/converter/internal/otelcolconvert/converter_otlphttpexporter.go
tpaschalis: Ah, we don't need that check.. I preemptively added it for components where no auth is being converted, but it doesn't hit that code path. Thanks for the heads up!

PR #6589, opened 2024-03-04
agent-static-mixin: include custom 'all' value

tpaschalis:
This makes it for easier querying on the backend of your choice. I've verified by comparing the resulting dashboards in main and the fix branch
```
mixtool generate dashboards mixin.libsonnet --directory fix1
```

I'm adding the new parameter on the existing helper instead of refactoring to use `addMultiTemplateWithAll`, should be enough for now.

### DIFF
--- a/operations/agent-static-mixin/dashboards.libsonnet
+++ b/operations/agent-static-mixin/dashboards.libsonnet
@@ -286,6 +286,7 @@ local template = grafana.template;
             value: '$__all',
           },
           includeAll=true,
+          allValues='.+',
         ),
       )
       .addTemplate(
@@ -300,6 +301,7 @@ local template = grafana.template;
             value: '$__all',
           },
           includeAll=true,
+          allValues='.+',
         ),
       )
       .addTemplate(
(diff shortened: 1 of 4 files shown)


operations/agent-static-mixin/dashboards.libsonnet
```diff
@@ -286,6 +286,7 @@ local template = grafana.template;
             value: '$__all',
           },
           includeAll=true,
+          allValues='.*',
```
pracucci, line 289: I know I suggested you `.*` but `.+` may actually be slightly more accurate. That's also what `jsonnet-libs` does by default.

pracucci (approved):
Thanks for working on it. I left a nit.

operations/agent-static-mixin/dashboards.libsonnet
tpaschalis: Yeah, let's change it. I had both `addMultiTemplate` and `addMultiTemplateWithAll` helpers with a different default, changing it now.

rfratto (approved):

pracucci (approved):
Thanks. I think what you're doing in this PR is consistent with `jsonnet-libs`.

PR #6592, opened 2024-03-04
feat(rules/k8s): refactor common `rules.k8s` logic into `common/kubernetes` package

### LINKED ISSUE #6545: Loki kubernetes rules refactor follow-up
The component `loki.rules.kubernetes` introduced by https://github.com/grafana/agent/pull/3936 has a lot of copied logic from the `mimir.rules.kubernetes` component (if the mimir implementation had bugs then those were ported to the loki implementation) 

We should avoid redundancy as much as possible to keep the code manageable. 

I suggest to put the common logic in the common kubernetes pkg.

hainenber:
#### PR Description

Refactor 100% sharable logic in `diff.go`, some custom types in `event.go` and `rules.go`. Expect no changed behaviour as this is a refactoring.

Method of `Component` implementations for both components are not refactored to avoid convoluted abstraction.

Closes #6545 

#### PR Checklist

- [x] Tests updated


wildum (approved):
LGTM, thanks a lot for taking care of this!

PR #6603, opened 2024-03-05
[windows exporter] Clarify the use of use_api in docs.

ptodev:
I tried to add as much information as possible, based on what I understand from the [upstream PR](https://github.com/prometheus-community/windows_exporter/pull/812). It would be good if someone with knowledge of the Windows exporter double checks my wording.


mattdurham:
From a windows perspective looks good, @clayton-cornell can add context around the verbiage itself.

mattdurham (approved):
Looks reasonable to me

docs/sources/flow/reference/components/prometheus.exporter.windows.md
```diff
@@ -172,14 +172,19 @@ For a server name to be included, it must match the regular expression specified
...
-`where_clause` | `string` | WQL 'where' clause to use in WMI metrics query.       | `""`    | no
+Name           | Type     | Description                                                 | Default | Required
+---------------|----------|-------------------------------------------------------------|---------|---------
+`use_api`      | `string` | Use the Windows API to collect service data instead of WMI. | `false` | no
+`where_clause` | `string` | WQL 'where' clause to use in WMI metrics query.             | `""`    | no
 
 The `where_clause` argument can be used to limit the response to the services you specify, reducing the size of the response.
 If `use_api` is enabled, 'where_clause' won't be effective.
 
+The Windows API is more performant than WMI. Set `use_api` to `true` in situations when the WMI takes too long to get the service information.
```
clayton-cornell, line 183:
```suggestion
The WMI is slow to query data, but it provides detailed results. The Windows API is fast but you must make several API calls to get the same amount of information. If speed is important and the WMI takes too long to retrieve information, you can set `use_api` to `true` to use the API to query your data.
```

I gave it a go at tweaking the text here a little to add a little more contextual info and details. Just saying the API is more performant doesn't tell you very much.

docs/sources/flow/reference/components/prometheus.exporter.windows.md
```diff
@@ -172,14 +172,19 @@ For a server name to be included, it must match the regular expression specified
...
+Setting `use_api` to `true` does have a few disadvantages compared to using WMI:
+* WMI queries in `where_clause` won't work.
```
clayton-cornell, line 185: Is there any alternative here? The second bullet provides an alternate solution.. so I'm wondering what happens in this case.

docs/sources/flow/reference/components/prometheus.exporter.windows.md
```diff
@@ -172,14 +172,19 @@ For a server name to be included, it must match the regular expression specified
...
+* The `status` field of the service won't be reported. Instead, use the `state` property. It has the same information, albeit formatted differently.
```
clayton-cornell, line 186:
```suggestion
* The `status` field of the service is not available. You can use the `state` property to retrieve status information. This property provides the same information, but it is formatted differently.
```

clayton-cornell (commented):
Some suggestions to expand the text just a little

clayton-cornell (approved):

clayton-cornell:
@mattdurham @ptodev Is this still valid? Can it be merged?

docs/sources/flow/reference/components/prometheus.exporter.windows.md
ptodev: I don't think there are alternatives. But I also don't think it's worth elaborating too much on this, since it seems natural that the "WMI queries" feature won't work when "WMI" is disabled.

The second bullet point is not related to WMI queries IIUC.

docs/sources/flow/reference/components/prometheus.exporter.windows.md
ptodev: TBH I think the extra text feels a bit vague and creates more questions than answers.

> The WMI is slow to query data, but it provides detailed results.

Why is it more detailed?

> The Windows API is fast but you must make several API calls to get the same amount of information.

Do the number of API calls matter? And I'm not sure where we got this information from?


I'm personally not an expert on this exporter so I'd prefer not to add too much information that I'm not sure about 😅

PR #6604, opened 2024-03-05
flow: add native histograms to Flow controller metrics

rfratto:
This commit adds native histogram to Flow controller metrics (component evaluation time and component evaluation wait time).

For the scope of this commit, both classic and native histograms are used for these metrics. A future commit will remove the classic histogram.

The settings I picked came from the [Mimir documentation for native histograms](https://grafana.com/docs/mimir/latest/send/native-histograms/).


mattdurham (approved):
LGTM

PR #6607, opened 2024-03-05
agent-flow-mixin: fix visualization for concurrent classic and native histograms

rfratto:
If both classic and native histograms are visualized concurrently, the old query would not consistently work properly in Grafana, as the `or` operator would return both classic and native histograms.

This would cause Grafana to visualize the histograms inconsistently, sometimes showing nothing for time ranges where native histogram data should be available.


tpaschalis (approved):
🙈

PR #6622, opened 2024-03-05
On new windows installs, remove default read permissions from agent c…

erikbaranowski:
…onfig

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,9 +25,9 @@ v0.40.2 (2024-03-05)
 - Set permissions on the `Grafana Agent [Flow]` folder when installing via the
   windows installer rather than relying on the parent folder permissions. (@erikbaranowski)
 
-- Set restricted viewing permission on the `agent-config.yaml` (Static mode) or
-  `config.river` (Flow mode) when installing via the windows installer if the
-  file does not already exist. (@erikbaranowski)
+- Set restricted viewing permissions on the `agent-config.yaml` (static mode) or
+  `config.river` (flow mode) when installing via the Windows installer if the
+  configuration file does not already exist. (@erikbaranowski)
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -25,6 +25,10 @@ v0.40.2 (2024-03-05)
 - Set permissions on the `Grafana Agent [Flow]` folder when installing via the
   windows installer rather than relying on the parent folder permissions. (@erikbaranowski)
 
+- Set restricted viewing permission on the `agent-config.yaml` (Static mode) or
+  `config.river` (Flow mode) when installing via the windows installer if the
+  file does not already exist. (@erikbaranowski)
```
rfratto, line 30:
```suggestion
- Set restricted viewing permissions on the `agent-config.yaml` (static mode) or
  `config.river` (flow mode) when installing via the Windows installer if the
  configuration file does not already exist. (@erikbaranowski)
```

packaging/grafana-agent-flow/windows/install_script.nsis
```diff
@@ -111,6 +111,14 @@ Function CreateConfig
     Return
   CreateNewConfig:
     File "config.river"
+
+    # Set permissions on the config file
```
rfratto, line 115: (Discussed offline, but writing it here for transparency)

I like that this applies only when the config file is first created, since it allows users to have custom permissions and not worry about them getting overwritten. Right now, the permissions for INSTDIR get updated every time one of these installers is ran, and I think similar logic should apply; only apply permissions the very first time the folder is created.

erikbaranowski: See the latest for this

rfratto (approved):

PR #6625, opened 2024-03-05
dockertarget: fix flaky test

rfratto:
The callback for one of the assert.EventuallyWithT calls incorrectly passed the test-wide *testing.T instance. This causes the assert.EventuallyWithT call to always return on the first tick, regardless of whether the assertion passed.

This means that on slower machines, the first assertion would fail, the assert.EventuallyWithT call would return prematurely, and the rest of the test would fail in turn.


rfratto:
I'll retrigger the CI a bunch of times to validate my theory that this is the issue. Unfortunately I was never able to run it locally because my machine seems to be too fast to trigger this issue.

spartan0x117 (approved):

rfratto:
Ok, the tests passed five times in a row, I'm going to call this mystery solved.

This is a very unfortunate one, and something that a linter really should've caught, but IIRC we either ignore unused parameters in tests or we don't lint for unused parameters at all. We should revisit whether that decision makes sense; this was a fairly innocuous mistake that ended up costing everyone a bunch of time.

PR #6627, opened 2024-03-06
add nonroot user to docker image

captncraig:
This adds a new `grafana-agent` user in the dockerfile.

I arbitrarily chose UID 473 to be one higher than the uid the grafana/grafana image uses.

The container does not run as this user by default, but can be used in k8s by adding this to your container spec:

```
securityContext:
  runAsNonRoot: true 
  runAsUser: 473 
  runAsGroup: 473
  fsGroup: 473
```


cmd/grafana-agent/Dockerfile
```diff
@@ -44,6 +44,15 @@ EOF
 COPY --from=build /src/agent/build/grafana-agent /bin/grafana-agent
 COPY cmd/grafana-agent/agent-local-config.yaml /etc/agent/agent.yaml
 
+ARG UID=0
+ARG GF_UID=473
+ARG GF_UNAME="grafana-agent"
```
thampiotr, line 49: Nits:
- Should we put args higher up, under `FROM`?
- The names of args are not self-explanatory like the other args we use. Maybe add a few words of comment what these do?

thampiotr (approved):
LGTM

rfratto:
* When will the documentation come for this? 
* Is it even practical to run the container as a non-root user? Grafana Agent is a system daemon, and needs root-level access for a bunch of things, including `loki.source.file` and `prometheus.exporter.unix`.

cmd/grafana-agent/Dockerfile
```diff
@@ -44,6 +44,15 @@ EOF
...
+
+RUN groupadd --gid $GF_UID $GF_UNAME
+RUN useradd -m -u $GF_UID -g $GF_UID $GF_UNAME
+RUN chown -R $GF_UNAME:$GF_UNAME /etc/agent
+RUN chown -R $GF_UNAME:$GF_UNAME /bin/grafana-agent
+USER $UID:$UID
```
rfratto, line 55: Do we really need a second Docker container? You can specify a non-root user to run as when running the normal container, can't you?

captncraig: Good point. Some users reported having issues with that, but maybe having a valid user in the container will make that easier?

captncraig:
> * Is it even practical to run the container as a non-root user

I think so? A lot of use cases don't need that access, like k8s discovery scrape and forwarding. And volume mounts can be set to a given uid too:

```
spec:
  securityContext:
    runAsUser: 473
    runAsGroup: 473
    fsGroup: 473
```

Might be sufficient for most of the cases you specify.

cmd/grafana-agent/Dockerfile
captncraig: https://stackoverflow.com/a/57917406 gives an example of security context with something like this.

rfratto:
Just to make my stance clear here: I'm not really opinionated on whether there's two or one image, or support for non-root at all, but if we're going to support non-root I do think it needs to be documented and we need to identify what can break and help people understand the risks of running as non-root.

captncraig:
I agree we should go with the least impact. The current version adds the user to the container, but does not make it the default. Which doc pages do you think the non-root info could be added to cleanly?

rfratto:
> Which doc pages do you think the non-root info could be added to cleanly?

I took a quick scan, these seem like the places to mention it:

* How to run as non-root on the [Docker install page](https://grafana.com/docs/agent/latest/flow/get-started/install/docker/) 
* How to run as non-root on the [Kubernetes install page](https://grafana.com/docs/agent/latest/flow/get-started/install/kubernetes/)
* Update component references pages for components that require root, including (not necessarily exhaustive):
  * discovery.docker
  * discovery.dockerswarm
  * discovery.file (conditionally depending on file permissions) 
  * local.file (conditionally depending on file permissions) 
  * local.file_match (conditionally depending on file permissions)
  * loki.source.docker
  * loki.source.file (conditionally depending on file permissions)
  * loki.source.journal
  * loki.source.syslog
  * module.file (conditionally depending on file permissions)
  * otelcol.processor.resourcedetection (conditionally depending on detector used)
  * prometheus.exporter.process
  * prometheus.exporter.unix 
  * pyroscope.ebpf 
  * pyroscope.java

captncraig:
I'm going to merge this now so we can validate it works, and add onto the docs asap after.

PR #6628, opened 2024-03-06
Remove invalid pipe escapes in prometheus.exporter.unix docs

DragoonAethis:
In documentation, default values for filesystem.fs_types_exclude and netstat.fields in separate multiline blocks contain escapes for the "|" character, while they're not actually escaped in the config (and indeed, the agent will not be happy should you copy and paste that value with escapes directly). Remove them.

Additionally, link to the current integration configuration source file for up-to-date config values on each OS the agent runs on, as the listed defaults are for Linux only.


docs/sources/flow/reference/components/prometheus.exporter.unix.md
```diff
@@ -135,6 +135,8 @@ The following blocks are supported inside the definition of
 
 ### filesystem block
 
+The default values can vary by the operating system the agent runs on - refer to the [integration source](https://github.com/grafana/agent/blob/main/internal/static/integrations/node_exporter/config.go) for up-to-date values on each OS.
```
clayton-cornell, line 138: Should we call out the OS-specific defaults (instead of linking to the source code)?

docs/sources/flow/reference/components/prometheus.exporter.unix.md
```diff
@@ -144,7 +146,7 @@ The following blocks are supported inside the definition of
 `fs_types_exclude` defaults to the following regular expression string:
 
 ```
-^(autofs\|binfmt_misc\|bpf\|cgroup2?\|configfs\|debugfs\|devpts\|devtmpfs\|fusectl\|hugetlbfs\|iso9660\|mqueue\|nsfs\|overlay\|proc\|procfs\|pstore\|rpc_pipefs\|securityfs\|selinuxfs\|squashfs\|sysfs\|tracefs)$
+^(autofs|binfmt_misc|bpf|cgroup2?|configfs|debugfs|devpts|devtmpfs|fusectl|hugetlbfs|iso9660|mqueue|nsfs|overlay|proc|procfs|pstore|rpc_pipefs|securityfs|selinuxfs|squashfs|sysfs|tracefs)$
```
clayton-cornell, line 149: @rfratto Can you take a look at this?

docs/sources/flow/reference/components/prometheus.exporter.unix.md
rfratto: I'd also prefer not linking to the source code here; if you don't want to do that in this PR the work can be split up.

docs/sources/flow/reference/components/prometheus.exporter.unix.md
rfratto: LGTM, I'm not sure why they were escaped in the first place and it doesn't look like they should've been since we weren't trying to literally match the `|` character.

docs/sources/flow/reference/components/prometheus.exporter.unix.md
clayton-cornell: I've created https://github.com/grafana/agent/issues/6701 and will follow up on this.

clayton-cornell (approved):
Looks good as-is. A followup issue has been created to update the docs so we are not linking into the source code.

PR #6629, opened 2024-03-06
docs: remove `beta` label from clustering

tpaschalis:
This PR removes the `beta` label from the clustering. Components _may_ change how they opt-in to participating via the dedicated block in the future.


thampiotr (approved):
LGTM! 🥳

PR #6632, opened 2024-03-07
Allow configuring hostNetwork for integrations

iarlyy:
#### PR Description

This PR implements the ability of setting `hostNetwork` pod attribute for integration pods. This may be required for example by `node_exporter` pods, that need to access [/proc/sys/net](https://docs.kernel.org/admin-guide/sysctl/net.html) for gathering host level network metrics.

Fixes https://github.com/grafana/alloy/issues/250

#### PR Checklist

- [ ] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/static/operator/resources_integrations.go
+++ b/internal/static/operator/resources_integrations.go
@@ -81,6 +81,8 @@ func integrationsPodTemplateOptions(name string, d gragent.Deployment, daemonset
 		inst := i.Instance
 		volumePrefix := fmt.Sprintf("%s-%s-", inst.Namespace, inst.Name)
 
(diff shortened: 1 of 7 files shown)


internal/static/operator/resources_integrations.go
```diff
@@ -231,6 +233,11 @@ func mergePodTemplateOptions(inputs ...*podTemplateOptions) podTemplateOptions {
 			res.ExtraEnvVars = append(res.ExtraEnvVars, ev)
 			varNames[ev.Name] = struct{}{}
 		}
+
+		if input.HostNetwork {
```
iarlyy, line 237: i am not quite sure that this actually will fit here, any advice is welcome.

PR #6634, opened 2024-03-07
Tweak release cadence description 

rfratto:
Tweak the release cadence description to: 

* Allow performing "emergency" minor releases; minor releases performed outside of the normal release cycle. As the name implies, this is a tool intended to be used in emergencies and not something that we regularly rely on. 
* Include a promise that minor releases performed within the cadence (e.g., _not_ emergency releases) will always include a version bump for our OpenTelemetry Collector dependencies.


thampiotr (approved):
FWIW, LGTM

PR #6643, opened 2024-03-09
feat(otelcol/exporter/otlp*): resync drifted defaults with upstream

### LINKED ISSUE #6467: Resync defaults for otelcol.exporter.otlp with upstream
This file notes that our defaults have become desynced with upstream:

https://github.com/grafana/agent/blob/744a4b7723eb4068d65e289a3ba709408868cede/converter/internal/otelcolconvert/testdata/otlp.yaml#L9-L14

Where possible, we should resync. If resyncing a particular field would be classified as a breaking change, we should be careful and consider whether resyncing is worth it.

### LINKED ISSUE #6485: Resync defaults for otelcol.exporter.otlphttp with upstream
#6484 found that our defaults for otelcol.exporter.otlphttp have desynced with upstream (similar to #6467 for otelcol.exporter.otlp). 

Where possible, we should resync. If resyncing a particular field would be classified as a breaking change, we should be careful and consider whether resyncing is worth it.

hainenber:

Fixes #6485
Fixes #6467 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added


docs/sources/shared/flow/reference/components/otelcol-queue-block.md
```diff
@@ -16,14 +16,14 @@ Name            | Type      | Description
...
+`queue_size`    | `number`  | Maximum number of unwritten batches allowed in the queue at the same time. | `1000`  | no
 
 When `enabled` is `true`, data is first written to an in-memory buffer before sending it to the configured server.
 Batches sent to the component's `input` exported field are added to the buffer as long as the number of unsent batches doesn't exceed the configured `queue_size`.
 
 `queue_size` determines how long an endpoint outage is tolerated.
-Assuming 100 requests/second, the default queue size `5000` provides about 50 seconds of outage tolerance.
-To calculate the correct value for `queue_size`, multiply the average number of outgoing requests per second by the time in seconds that outages are tolerated.
+Assuming 100 requests/second, the default queue size `1000` provides about 10 seconds of outage tolerance.
+To calculate the correct value for `queue_size`, multiply the average number of outgoing requests per second by the time in seconds that outages are tolerated. Note that a very high figure can cause unexpected Out Of Memory (OOM) kills.
```
clayton-cornell, line 26:
```suggestion
To calculate the correct value for `queue_size`, multiply the average number of outgoing requests per second by the time in seconds that outages are tolerated. A very high value can cause Out Of Memory (OOM) kills.
```

clayton-cornell (commented):
One suggestion for consistency and simplifying.

clayton-cornell:
Over to @grafana/grafana-agent-maintainers for code review

clayton-cornell (dismissed):

hainenber:
Looks like I need to update some defaults as well. I'm at work atm, would troubleshoot and remediate later

internal/component/otelcol/config_grpc.go
```diff
@@ -13,6 +13,8 @@ import (
 	otelextension "go.opentelemetry.io/collector/extension"
 )
 
+const DEFAULT_BALANCER_NAME = "pick_first"
```
rfratto, line 16: Idiomatically, it's unusual for constants in Go to be SNAKE_CASE like this. Can this be called `DefaultBalancerName` instead?

rfratto (commented):
Thanks! Just one comment about the constant name, then this should be good to go :)

internal/component/otelcol/config_grpc.go
hainenber: Done :D

docs/sources/shared/flow/reference/components/otelcol-queue-block.md
clayton-cornell: I don't see this suggestion applied. Generally we do not want to state `Note that` in the documentation. If it's a note, then we should make it a note... In this case a factual statement is enough.

hainenber: My bad, I failed to pull the changes here to my local. Commited!

clayton-cornell (approved):
Docs look good. Over to you @rfratto for final check :-)

rfratto (approved):
Thank you!

PR #6647, opened 2024-03-11
docs: remove beta labels from import.http, import.git, import.string

tpaschalis:
This PR removes the beta label from the import.http and import.string components. The new modules have started picking up use and we don't anticipate breaking changes in the foundational way that these components work.


clayton-cornell (approved):
LGTM

erikbaranowski (approved):
We mentioned maybe not marking import.git or a subset of import.file features when we talked. Approved but I'm mentioning as a sanity check

PR #6659, opened 2024-03-12
remotecfg: fix race condition in test

tpaschalis:
This PR is to fix a race condition in the tests of the remotecfg package that renders our build pipeline flaky.

I'll be pushing along a series of commits to verify that the Drone and Github action runners are in a better state as well.

Apologies in advance for not picking this up earlier and the lost amount of dev time.


tpaschalis:
This had 6 sequential CI runs pass, which makes me think this has some merit.

erikbaranowski (approved):
LGTM, thanks!

PR #6665, opened 2024-03-12
agentseed: make sure we have a valid uid, or regenerate

captncraig:
I suspect this is the cause of some inconsistent data we are seeing. This return should always have been here.


ptodev (approved):

ptodev:
Hi @captncraig, would you like to merge the PR? I rebased it for you :)

PR #6668, opened 2024-03-12
Update to fetch head

mattdurham:
#### PR Description

Internally for git components/config blocks we are using fetch which updates the FETCH_HEAD and not HEAD. So whatever the most recent pull was gets permanently pinned since HEAD never changes.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,8 @@ Main (unreleased)
 
 - Fix a bug where structured metadata and parsed field are not passed further in `loki.source.api` (@marchellodev)
 
+- Git operations used `HEAD` which could lead to incorrect pulls since it uses fetch and not pull, it now uses `FETCH_HEAD` by default. (@mattdurham)
+
 ### Other changes
 
 - Clustering for Grafana Agent in Flow mode has graduated from beta to stable.
--- a/docs/sources/flow/reference/components/module.git.md
+++ b/docs/sources/flow/reference/components/module.git.md
@@ -41,12 +41,12 @@ module.git "LABEL" {
 
 The following arguments are supported:
 
-Name | Type | Description | Default | Required
----- | ---- | ----------- | ------- | --------
-`repository` | `string` | The Git repository address to retrieve the module from. | | yes
-`revision` | `string` | The Git revision to retrieve the module from. | `"HEAD"` | no
-`path` | `string` | The path in the repository where the module is stored. | | yes
-`pull_frequency` | `duration` | The frequency to pull the repository for updates. | `"60s"` | no
+Name | Type | Description | Default        | Required
+---- | ---- | ----------- |----------------| --------
+`repository` | `string` | The Git repository address to retrieve the module from. |                | yes
+`revision` | `string` | The Git revision to retrieve the module from. | `"FETCH_HEAD"` | no
+`path` | `string` | The path in the repository where the module is stored. |                | yes
+`pull_frequency` | `duration` | The frequency to pull the repository for updates. | `"60s"`        | no
 
 The `repository` attribute must be set to a repository address that would be
 recognized by Git with a `git clone REPOSITORY_ADDRESS` command, such as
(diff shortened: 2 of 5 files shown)


rfratto (approved):
LGTM, but can we add a test for this to reproduce the old issue and make sure this fixes it? Even after looking up the difference between HEAD and FETCH_HEAD, I don't really grok what the issue was here so this seems prone to regressions without a test. 

Do we have any theories for why we haven't seen people report this yet? module.git has a fair bit of usage behind it, and hopefully someone would've noticed it not updating. Do we think people just weren't relying on `HEAD` and explicitly setting a revision?

mattdurham:
There is a bug in the above, its going to be a bigger change. Basically FETCH_HEAD isnt set on initial pull for some reason and it needs to be HEAD. One solution might be to use PULL instead.

mattdurham:
Added a complex test to ensure that we could reproduce the behavior and also ensure it was fixed. Not in love with the pull based mechanic but it works without any gotchas.

internal/vcs/git.go
```diff
@@ -30,7 +30,7 @@ type GitRepo struct {
 // managed at storagePath.
 //
 // If storagePath is empty on disk, NewGitRepo initializes GitRepo by cloning
-// the repository. Otherwise, NewGitRepo will do a fetch.
+// the repository. Otherwise, NewGitRepo will do a pull.
```
rfratto, line 33: I feel like I don't really understand what the problem with fetching was, especially given the documentation of [git pull](https://git-scm.com/docs/git-pull) saying:

> More precisely, `git pull` runs `git fetch` with the given parameters and then depending on configuration options or command line flags, will call either `git rebase` or `git merge` to reconcile diverging branches.

Given this, it seems like `git fetch && git reset SHA` should've done it, and so I'm not sure `git fetch` was the issue so much as the `git reset SHA` part was.

But either way, as long as the test case reproduced the error prior to the fix, and it's fixed now, we can push this through.

CHANGELOG.md
```diff
@@ -32,6 +32,8 @@ Main (unreleased)
 
 - Fix a bug where structured metadata and parsed field are not passed further in `loki.source.api` (@marchellodev)
 
+- Git operations used fetch which updated `FETCH_HEAD` which could lead to incorrect pulls since the default is `HEAD` switched to pull for better compatibility. (@mattdurham)
```
rfratto, line 35: I had a hard time following this message, what about something like:

```suggestion
- Change `import.git` to use Git pulls rather than fetches to fix scenarios where the local code did not get updated. (@mattdurham)
```

rfratto (approved):

internal/vcs/git.go
mattdurham: I could see the reset being the core problem. Thought about aliasing HEAD to FETCH_HEAD if different but that felt harder to explain and more error prone.

mattdurham: Example below, so if git reset HEAD will just reset to what you have locally, you would need to call reset FETCH_HEAD for it to work properly with fetch.
❯ git fetch
❯ git rev-parse HEAD
db656133b17b3b16ae406406089d8c9303c5d595
❯ git rev-parse FETCH_HEAD
954f41122b2fa387ef5e6b2d5845009925d9d169

Running a git pull changes to
❯ git pull
❯ git rev-parse HEAD
954f41122b2fa387ef5e6b2d5845009925d9d169
❯ git rev-parse FETCH_HEAD
954f41122b2fa387ef5e6b2d5845009925d9d169

internal/flow/import_test.go
```diff
@@ -250,6 +251,101 @@ func TestImportError(t *testing.T) {
 	}
 }
 
+func TestPullUpdating(t *testing.T) {
```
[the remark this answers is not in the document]
mattdurham: Will take a look at this after merging the pr to see if I can clean up the test.

PR #6674, opened 2024-03-13
service: allow services to have defined stability levels

rfratto:
This commit adds stability levels to services. Every service, with the exception of remotecfg, is then given a stability level of stable. remotecfg is given a stability level of "beta."

### DIFF
--- a/internal/component/registry.go
+++ b/internal/component/registry.go
@@ -124,6 +124,7 @@ type Registration struct {
 	// sure the user is not accidentally using a component that is not yet stable - users
 	// need to explicitly enable less-than-stable components via, for example, a command-line flag.
 	// If a component is not stable enough, an attempt to create it via the controller will fail.
+	// This field must be set to a non-zero value.
 	Stability featuregate.Stability
 
 	// An example Arguments value that the registered component expects to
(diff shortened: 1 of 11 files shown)


internal/component/registry.go
```diff
@@ -124,7 +124,6 @@ type Registration struct {
 	// sure the user is not accidentally using a component that is not yet stable - users
 	// need to explicitly enable less-than-stable components via, for example, a command-line flag.
 	// If a component is not stable enough, an attempt to create it via the controller will fail.
-	// The default stability level is Experimental.
```
thampiotr, line 127: thanks for spotting

internal/service/cluster/cluster.go
```diff
@@ -162,6 +163,7 @@ func (s *Service) Definition() service.Definition {
 			// Cluster depends on the HTTP service to work properly.
 			http_service.ServiceName,
 		},
+		Stability: featuregate.StabilityBeta,
```
thampiotr, line 166: In the PR comment I think you said this should be stable. I think clustering should be marked as stable, that's what the docs currently say.

thampiotr (approved):
LGTM % stability of the clustering service

PR #6681, opened 2024-03-14
otelcolconvert: support converting servicegraph connector

### LINKED ISSUE #6443: Support converting `servicegraphconnector`

tpaschalis:
Closes #6443


internal/converter/internal/otelcolconvert/testdata/servicegraph.river
```diff
@@ -0,0 +1,33 @@
+otelcol.receiver.otlp "default" {
+	grpc { }
+
+	http { }
+
+	output {
+		metrics = [otelcol.exporter.otlp.default.input]
```
erikbaranowski, line 7: Is this correct to have the otlp receiver send to the otlp exporter directly? I'm trying to understand the concept of connectors in Otel so let me know.

rfratto: Ooh, I think this is a bug in the framework. The otlp receiver is not configured in any case to send metrics to the otlp exporter, so no, this isn't right. Nice catch.

tpaschalis: Whoops, will take a look, thanks for not glancing over it.

erikbaranowski: This should be all fixed now

erikbaranowski (approved):

PR #6682, opened 2024-03-14
otelcol.receiver.opencensus: sync with upstream defaults

### LINKED ISSUE #6489: Sync defaults in otelcol.receiver.opencensus
This file notes that our defaults for otelcol.receiver.opencensus have desynced with upstream (or in this case, I'm not sure it was ever correct).

https://github.com/grafana/agent/blob/c58709a9625b361200183e3dff083a25e4e8ef3b/converter/internal/otelcolconvert/testdata/opencensus.yaml#L6-L8

Changing a listening port like this _would_ be a breaking change, but given how 4317 is used by another receiver by default, I think it's justifiable in this case; we should change our default for otelcol.receiver.opencensus to align with the upstream opencensusreceiver component.

rfratto:
This commit re-syncs `otelcol.receiver.opencensus` with the upstream defaults, namely changing the default listen port from 4317 to 55678.

This is considered a breaking change; a section has been written in the upgrade guide notifying users to explicitly set the endpoint to the previous port to retain old behavior when upgrading.

Closes #6489.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ internal API changes are not present.
 Main (unreleased)
 -----------------
 
+### Breaking changes
+
+- The default listen port for `otelcol.receiver.opencensus` has changed from
+  4317 to 55678 to align with upstream. (@rfratto)
+
 ### Enhancements
 
 - Add support for importing folders as single module to `import.file`. (@wildum)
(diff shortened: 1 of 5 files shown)


clayton-cornell (approved):
Looks OK to me.

PR #6700, opened 2024-03-14
Pass in a label prefix from static to otelcolconvert so multiple trac…

### LINKED ISSUE #6691: Pass in a unique prefix for each trace instance when converting to flow
The tracing subsystem for static mode allows us to have multiple otel collector instances running. Flow mode runs a single instance so add support for a unique component label prefix for each.

erikbaranowski:
…e configs don't bump into each other

Closes #6691

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/converter/internal/staticconvert/internal/build/builder.go
+++ b/internal/converter/internal/staticconvert/internal/build/builder.go
@@ -386,8 +386,7 @@ func (b *ConfigBuilder) appendTraces() {
 			b.diags.Add(diag.SeverityLevelCritical, fmt.Sprintf("failed to load otelConfig from agent traces config: %s", err))
 			continue
 		}
-		// TODO: what prefix should each generated flow label get? each trace instance will need something unique
-		b.diags.AddAll(otelcolconvert.AppendConfig(b.f, otelCfg))
+		b.diags.AddAll(otelcolconvert.AppendConfig(b.f, otelCfg, cfg.Name))
 	}
 }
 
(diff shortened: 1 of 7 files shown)


internal/converter/internal/staticconvert/internal/build/builder.go
```diff
@@ -386,8 +386,7 @@ func (b *ConfigBuilder) appendTraces() {
 			b.diags.Add(diag.SeverityLevelCritical, fmt.Sprintf("failed to load otelConfig from agent traces config: %s", err))
 			continue
 		}
-		// TODO: what prefix should each generated flow label get? each trace instance will need something unique
-		b.diags.AddAll(otelcolconvert.AppendConfig(b.f, otelCfg))
+		b.diags.AddAll(otelcolconvert.AppendConfig(b.f, otelCfg, cfg.Name))
```
rfratto, line 389: nit: if there's only one instance, do we really need the prefix? it definitely makes the labels much longer

rfratto (approved):

internal/converter/internal/staticconvert/internal/build/builder.go
erikbaranowski: good call I can fix that up

erikbaranowski: I added a multi vs single test to cover our scenarios on this new logic

PR #6709, opened 2024-03-15
docs: make deprecation of classic modules more clear

rfratto:
The deprecation status of classic modules was constrained to the concept page of modules. This commit adds the deprecation notice to each component reference page and mentions it in the release notes.

### DIFF
--- a/docs/sources/flow/reference/components/module.string.md
+++ b/docs/sources/flow/reference/components/module.string.md
@@ -14,7 +14,8 @@ title: module.string
 # module.string (deprecated)
 
 {{< admonition type="caution" >}}
-Modules were redesigned in v0.40 to simplify concepts. `module.string` has been deprecated as of v0.40 in favor of `import.string`, and will be removed in the following release.
+Starting with release v0.40, `module.string` is deprecated and is replaced by `import.string`.
+`module.string` will be removed in a future release.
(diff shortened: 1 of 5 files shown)


rfratto:
I'm also noticing we didn't communicate the prometheus.exporter.vsphere deprecation alongside v0.40; I'll add that in a follow-up PR.

docs/sources/flow/reference/components/module.string.md
```diff
@@ -11,7 +11,11 @@ labels:
 title: module.string
 ---
 
-# module.string
+# module.string (deprecated)
+
+{{< admonition type="caution" >}}
+Modules were redesigned in v0.40 to simplify concepts. `module.string` has been deprecated as of v0.40 in favor of `import.string`, and will be removed in the following release.
```
clayton-cornell, line 17:
```suggestion
Starting with release v0.40, `module.string` is deprecated and is replaced by `import.string`.
`module.string` will be removed in a future release.
```

Looking at the wording suggested here: https://grafana.com/docs/writers-toolkit/write/deprecate-remove/ maybe a small tweak to this to make it more active voice and simplify the message in the Caution (we don't need to explain the why it was replaced).

Will it be definitely be removed in v0.41? Then we could state this explicitly.

PR #6713, opened 2024-03-17
feat(converter/otelcol): support converting `sigv4authextension`

### LINKED ISSUE #6442: Support converting `sigv4authextension`

hainenber:

Closes #6442 

#### Notes to the Reviewer
I'm not sure this is the correct approach as the newly added `ConvertWithoutValidation` doesn't really reflect production usage and only for testing. I might open a new PR on the upstream to see if OtelCol folks are OK with opening the struct and/or add a setter to inject mock SigV4 credential provider.


internal/converter/internal/test_common/testing.go
```diff
@@ -45,7 +45,7 @@ const (
 func TestDirectory(t *testing.T, folderPath string, sourceSuffix string, loadFlowConfig bool, extraArgs []string, convert func(in []byte, extraArgs []string) ([]byte, diag.Diagnostics)) {
 	require.NoError(t, filepath.WalkDir(folderPath, func(path string, d fs.DirEntry, _ error) error {
 		if d.IsDir() {
-			return nil
+			return filepath.SkipDir
 		}
```
rfratto, line 49: This has the unfortunate side effect of skipping _all_ tests; the first callback in filepath.WalkDir when walking testdata will be the testdata directory itself; by skipping that directory, all test cases are accidentally skipped.

internal/converter/internal/otelcolconvert/testdata/otelcol_without_validation/sigv4auth.river
```diff
@@ -0,0 +1,29 @@
+otelcol.auth.sigv4 "default" {
+  	region  = "ap-southeast-1"
```
rfratto, line 2: fyi: There's a formatting error here that isn't reported on yet mainly because of the test_common change :)

rfratto (commented):
Thanks for the PR! Will approve after the two comments are resolved. 

I'm OK with the ConvertWithoutValidation workaround for now, this one seems like an edge case and it's probably not worth doing a more extensive change just to get the converter working.

internal/converter/internal/test_common/testing.go
hainenber: I'd be guilty not fixing this :D. Now only child directories are skipped and not the top-level one

internal/converter/internal/otelcolconvert/testdata/otelcol_without_validation/sigv4auth.river
hainenber: Thanks! Fixed in HEAD now :D

rfratto (approved):
LGTM, thank you!

PR #6714, opened 2024-03-17
feat(exporter/otelcol): support converting `vcenterreceiver`

### LINKED ISSUE #6459: Support converting `vcenterreceiver`

hainenber:

Close #6459


internal/converter/internal/otelcolconvert/converter_vcenterreceiver.go
```diff
@@ -0,0 +1,83 @@
...
+	var (
+		nextMetrics = state.Next(id, component.DataTypeMetrics)
+		nextTraces  = state.Next(id, component.DataTypeTraces)
+	)
+
+	return &vcenter.Arguments{
+		Endpoint: cfg.Endpoint,
+		Username: cfg.Username,
+		Password: rivertypes.Secret(cfg.Password),
+
```
erikbaranowski, line 52: It looks like we are missing the conversion of `MetricsBuilderConfig` here.

internal/converter/internal/otelcolconvert/converter_vcenterreceiver.go
```diff
@@ -0,0 +1,83 @@
...
+		DebugMetrics: common.DefaultValue[vcenter.Arguments]().DebugMetrics,
+
+		ScraperControllerArguments: otelcol.ScraperControllerArguments{
+			CollectionInterval: cfg.CollectionInterval,
+			InitialDelay:       cfg.InitialDelay,
+			Timeout:            cfg.Timeout,
+		},
+
+		TLS: otelcol.TLSClientArguments{
```
erikbaranowski, line 61: We can use a common way to convert the TLS settings:

```
TLS: toTLSClientArguments(cfg.TLSClientSetting),
```

erikbaranowski (changes requested):
Thanks for the PR 😄 . Looks solid and I've left a couple comments to finalize this. Let me know if you'd like me to work them, no pressure.

internal/converter/internal/otelcolconvert/converter_vcenterreceiver.go
hainenber: I've added the converter but it looks like the defaults for agent's `vcenter` doesn't seem to be in sync with upstream, hence the full-blown River converted result.

These blocks with `enabled = true` should be omitted, right? Just wanna make sure that I understand the concept of `upstream syncing` correctly here

internal/converter/internal/otelcolconvert/converter_vcenterreceiver.go
hainenber: Done :D

internal/converter/internal/otelcolconvert/converter_vcenterreceiver.go
erikbaranowski: Good catch - I researched it and it made sense to commit the change I tested. The way nested structs were being handled in the component led to this weird behavior and it should be all set now. There is a slight divergence in defaults but it is not so bad anymore.

erikbaranowski (approved):
Thank you!

erikbaranowski:
I disabled automerge so this doesn't cause conflict with a bug fix we are working on. Once that is shipped I can merge it into this PR and get this merged.

PR #6725, opened 2024-03-19
Upgrade OTel from v0.87.0 to v0.96.0

### LINKED ISSUE #6389: Upgrade OpenTelemetry to at least  versions v1.23.1/v0.94.1
We plan to integrate https://github.com/grafana/beyla in the Agent. The problem is that depends on version OpenTelemetry that are incompatible with Agent. If this helpful at all, [this is](https://github.com/grafana/agent/compare/add_beyla_component) what I had to comment and delete to make it work. The main problem might that we have a custom fork.

ptodev:
#### PR Description

OTel upgrade. I went through the changes in the OTel changelog. As far as I can tell, this PR updates all the relevant Agent bits. I hope I didn't miss anything.

Fixes #6389

#### Notes to the Reviewer

I tested this locally with the Static and River files according to our [docs](https://github.com/grafana/agent/blob/main/docs/developer/updating-otel.md).

There are some additional things I could do:
* Add tests for the new config arguments.
* There are a few TODOs in the changelog which I need to take a look at.
* When the [fork PR](https://github.com/grafana/opentelemetry-collector/pull/10) is merged, I need to update the go.mod file again.
* Update config converters?

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [ ] Tests updated


wildum (commented):
Thanks for taking care of this! I only have a few comments, nothing major. With this review I only checked the added code. I did not go through all opentelemetry release changes to see if something was missing

wildum:
Found a missing change: in extract-field-block.md we say that "The source of the labels or annotations. Allowed values are `pod` and `namespace`." But this change in v0.89.0 adds the value "node": https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/28570

wildum:
Scrolled through the rest of the open telemetry contrib release, I did not see any other missing changes, well done 👍 For the future I would recommend having atomic commits for every change to make the review easier

docs/sources/flow/reference/components/otelcol.exporter.otlp.md
```diff
@@ -177,6 +177,13 @@ information.
 
 * `exporter_sent_spans_ratio_total` (counter): Number of spans successfully sent to destination.
 * `exporter_send_failed_spans_ratio_total` (counter): Number of spans in failed attempts to send to destination.
+* `exporter_queue_capacity_ratio` (gauge): Fixed capacity of the retry queue (in batches)
+* `exporter_queue_size_ratio` (gauge): Current size of the retry queue (in batches)
+* `rpc_client_duration_milliseconds` (histogram): Measures the duration of inbound RPC.
+* `rpc_client_request_size_bytes` (histogram): Measures size of RPC request messages (uncompressed).
+* `rpc_client_requests_per_rpc` (histogram): Measures the number of messages received per RPC. Should be 1 for all non-streaming RPCs.
+* `rpc_client_response_size_bytes` (histogram): Measures size of RPC response messages (uncompressed).
+* `rpc_client_responses_per_rpc` (histogram): Measures the number of messages received per RPC. Should be 1 for all non-streaming RPCs.
```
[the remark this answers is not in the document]
ptodev: I ran the Agent locally and saw these metrics. They come from one of the libraries, but I can't remember which one.

ptodev:
@wildum Thank you for the review! I addressed some of your comments. I plan to address the other ones in a subsequent PR.

I also took out these lines from the CHANGELOG:
```
  * [otelcol.connector.servicegraph] Additional debug metrics
  <!-- TODO: Document those metrics -->
https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/29917
  * [otelcol.processor.filter] Additional debug metrics
  <!-- TODO: Document those metrics -->
https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/30736
  * Scope name for all generated Meter/Tracer funcs now includes full package name
  <!-- TODO: Does this affect the Agent's metrics and logs? -->
https://github.com/open-telemetry/opentelemetry-collector/issues/9494
  * [otelcol.connector.servicegraph] Measure latency in seconds instead of milliseconds
  <!-- TODO: This is a breaking change for Static mode? -->
  <!-- TODO: Is this a breaking change for Flow? -->
https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/27488
```

I will review them and reintroduce them in a subsequent PR.

internal/component/otelcol/config_tls.go
```diff
@@ -81,6 +82,7 @@ func (args *TLSSetting) Convert() *otelconfigtls.TLSSetting {
 		MinVersion:               args.MinVersion,
 		MaxVersion:               args.MaxVersion,
 		ReloadInterval:           args.ReloadInterval,
+		CipherSuites:             args.CipherSuites,
```
wildum, line 85: should we create a new slice instead of copying it like this?

internal/static/traces/config_test.go
```diff
@@ -1159,6 +1159,8 @@ remote_write:
       scopes: ["api.metrics"]
       timeout: 2s
 `,
+			//TODO(ptodev): Look into why we need to add a "cipher_suites: []" explicitly.
+			// The expected config should unmarshal it to [], but instead it sets it to nil.
```
wildum, line 1163: It seems correct to me that the config would unmarshal it to nil. Else maybe you can explicitly create a list during the convert part (converting from nil to {}string[]). I think that we should avoid adding "cipher_suites: []" explicitly

wildum (commented):
Second review, only two little comments

internal/static/traces/config_test.go
ptodev: > It seems correct to me that the config would unmarshal it to nil.

I used a debugger in one of the Collector's unit tests, to see if it's `nil` or `[]`. It was `[]`, so I assumed that this is the correct thing to do. I don't know if the unit tests deviates from their prod code path, but I guess it doesn't.

ptodev: I think the way Agent's unit test creates the Collector config isn't the same way the Collector does it. Something is different, but I'm not sure what exactly.

wildum (approved):
let's ship it!

wildum:
@clayton-cornell could you have a look at the docs when you have time? We are merging the branch because this PR is already very big and we don't want it to be stuck longer. Some of my comments and your future comments will be addressed in a subsequent PR

PR #6730, opened 2024-03-19
Fix an issue where the default values of some component's arguments change whenever that argument is explicitly configured

erikbaranowski:
#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated


rfratto (approved):
My understanding is that the Windows test has been flaking and not failing consistently, right? So I'm not sure this is the right long-term fix, but we can merge it to see anyway

erikbaranowski:
There's a consistently broken test from the new converter work. This change to make a squashed property not a pointer fixes general inconsistencies with how river compares fields to the default value.

internal/component/all/all_test.go
```diff
@@ -0,0 +1,175 @@
...
+			return "", true
+		}
+		return sharePointer(a.Elem(), b.Elem())
+
+	case reflect.Slice:
+		if pointersMatch(a, b) {
+			// If the slices are nil, we don't need to check for shared pointers
+			if a.Len() > 0 || b.Len() > 0 {
+				return "", true
+			}
```
rfratto, line 120: This isn't quite right, since it could be sharing a pointer where len=0 but cap>0, which would be an error. This playground link shows that the pointer for slices is fixed until either len>0 or cap>0: https://go.dev/play/p/vUsVSJ-Z9jo

So this condition should be:

```go 
if a.Len() > 0 || a.Cap() > 0 || b.Len() > 0 || b.Cap() > 0 {
```

Also note that that the comment here isn't quite right. `[]string{}` is not nil, it's just a preallocated immutable pointer, so it's safe for us to ignore.

(This longer condition may be worth a helper function so it can be explained)

PR #6731, opened 2024-03-19
add announcement that prometheus.exporter.vsphere was deprecated in v0.40

rfratto:
The intent was the prometheus.exporter.vsphere was deprecated in v0.40, but this was mistakenly omitted from the release. This commit adds the deprecation notice.

Note that while the replacement component, otelcol.receiver.vcenter, is still marked as experimental, it wasn't intentional that prometheus.exporter.vsphere received the "stable" label, and it should have similarly been marked experimental.

### DIFF
--- a/docs/sources/flow/release-notes.md
+++ b/docs/sources/flow/release-notes.md
@@ -36,10 +36,6 @@ Other release notes for the different {{< param "PRODUCT_ROOT_NAME" >}} variants
 The default listen port for `otelcol.receiver.opencensus` has changed from 4317 to 55678 to align with the upstream defaults.
 To retain the previous listen port, explicitly set the `endpoint` argument to `0.0.0.0:4317` before upgrading.
 
-### Breaking change: classic modules have been removed
-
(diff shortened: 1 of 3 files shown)


docs/sources/flow/release-notes.md
```diff
@@ -36,10 +36,6 @@ Other release notes for the different {{< param "PRODUCT_ROOT_NAME" >}} variants
 The default listen port for `otelcol.receiver.opencensus` has changed from 4317 to 55678 to align with the upstream defaults.
 To retain the previous listen port, explicitly set the `endpoint` argument to `0.0.0.0:4317` before upgrading.
 
-### Breaking change: classic modules have been removed
-
-Classic modules (the `module.git`, `module.file`, `module.http`, and `module.string` components) were initially deprecated in v0.40 in favor of the `import` and `declare` configuration blocks, and have been removed as of this release.
```
rfratto, line 41: I jumped the gun with adding this line; we don't really know what agent release will remove these components.

tpaschalis (approved):

PR #6737, opened 2024-03-20
chore(converter/otelcol): remove deep pretty printer clause added during implementation debugging

hainenber:

N/A

#### Notes to the Reviewer

I noticed there are `spew.Dump()`clauses in `otelcolconvert` directory. I'm assuming these were added to aid debugging when implementing the converters.


tpaschalis (approved):
Nice catch 🕸️

PR #6739, opened 2024-03-20
feat(loki/src/k8s): configurable default `instance` label for scraped pod logs

### LINKED ISSUE #6724: `loki.source.kubernetes` instance label conflicts with self defined instance labels
### What's wrong?

I'm using the `loki.source.kubernetes` component, and I have issues trying to replicate the original promtail helm-chart config.

The promtail config contains an "instance" label that is set **when** the pod has an "instance" label, as you can see here: https://github.com/grafana/helm-charts/blob/1338c741e3ba6fa2c7b17f20c4aced673fa74943/charts/promtail/values.yaml#L509-L514

Using the same config causes conflicts with Grafana Agent. The "instance" label doesn't always exist on every pod, and therefore this label is usually not set/defined.

The issue is that Grafana Agent always defines this when it’s missing, as you can see here:
https://github.com/grafana/agent/blob/a7153e4bb940f4a1d65e01b173a24e236d15c043/internal/component/loki/source/kubernetes/kubetail/target.go#L222-L226

I'd expect to have this configurable. Allow user to disable this and allow user to change the name of that label.

### Steps to reproduce

```hcl
discovery.relabel "pods" {
  targets = discovery.kubernetes.pods.targets

  // either use this rule to see my exact issue
  rule {
    source_labels = [
      "__meta_kubernetes_pod_label_app_kubernetes_io_instance",
      "__meta_kubernetes_pod_label_instance",
    ]
    regex         = "^;*([^;]+)(;.*)?$"
    target_label  = "instance"
  }

  // or run this rule and you'll see that "instance" label still exists
  rule {
    regex  = "instance"
    action = "labeldrop"
  }
}

loki.source.kubernetes "pods" {
  targets    = discovery.relabel.pods.output
}
```

Running with this config, you'll see that you have instance labels that are not Pod Labels.

### System information

_No response_

### Software version

v0.43.3

### Configuration

_No response_

### Logs

_No response_

hainenber:

Fixes grafana/alloy#211 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated
- ~~Config converters updated~~


docs/sources/flow/reference/components/loki.source.kubernetes.md
```diff
@@ -159,6 +161,19 @@ arguments.
...
+
+The `default_instance_label` block configures default "instance" label to be assigned with logs 
+scraped from inputted targets.
+If `default_instance_label` block is not provided, the target's address is used as the default `instance` label.
+
+The following arguments are supported:
+
+Name                     | Type                | Description                                                   | Default | Required
+------------------------ | ------------------- | ------------------------------------------------------------- | ------- | --------
+`enabled`                | `bool`              | Enable setting default "instance" label value.                | true    | no
```
clayton-cornell, line 174:
```suggestion
`enabled`                | `bool`              | Enable setting default instance label value.                | true    | no
```

docs/sources/flow/reference/components/loki.source.kubernetes.md
```diff
@@ -159,6 +161,19 @@ arguments.
 
 [using clustering]: {{< relref "../../concepts/clustering.md" >}}
 
+### default_instance_label block
+
+The `default_instance_label` block configures default "instance" label to be assigned with logs 
+scraped from inputted targets.
+If `default_instance_label` block is not provided, the target's address is used as the default `instance` label.
```
clayton-cornell, line 168:
```suggestion
If you don't provide the `default_instance_label` block, the target's address is used as the default instance label.
```

docs/sources/flow/reference/components/loki.source.kubernetes.md
```diff
@@ -159,6 +161,19 @@ arguments.
 
 [using clustering]: {{< relref "../../concepts/clustering.md" >}}
 
+### default_instance_label block
+
+The `default_instance_label` block configures default "instance" label to be assigned with logs 
+scraped from inputted targets.
```
clayton-cornell, line 167: I'm not really following what this is describing. If the `enabled` argument is set to `true` in the `default_instance_label` block then you can explicitly assign an instance label with `value` argument? 

So, how does this relate to "logs scraped from inputted targets"? Is this what happens when you don't set `enabled` to `true`?

clayton-cornell (commented):
Some doc input/comments

clayton-cornell:
Over to @grafana/grafana-agent-maintainers for a code review.

docs/sources/flow/reference/components/loki.source.kubernetes.md
hainenber: >  If the enabled argument is set to true in the default_instance_label block then you can explicitly assign an instance label with value argument?

yes, this is what I'm trying to convey

> So, how does this relate to "logs scraped from inputted targets"? Is this what happens when you don't set enabled to true?
The scraped logs from inputted targets are the pod logs to have the default "instance" label. If we don't set `enabled=true`, the behavior would stay the same as before, i.e. automatically assign an "instance" label

clayton-cornell:
```suggestion
You can use the `default_instance_label` block to explicitly define the default "instance" label assigned to the logs 
scraped from the targets.
```

So, something like this?

hainenber: Yes! That's the spot

clayton-cornell (approved):
Approved from a docs standpoint

docs/sources/flow/reference/components/loki.source.kubernetes.md
```diff
@@ -159,6 +161,19 @@ arguments.
...
+
+You can use the `default_instance_label` block to explicitly define the default "instance" label assigned to the logs 
+scraped from the targets.
+If you don't provide the `default_instance_label` block, the target's address is used as the default instance label.
+
+The following arguments are supported:
+
+Name                     | Type                | Description                                                   | Default | Required
+------------------------ | ------------------- | ------------------------------------------------------------- | ------- | --------
+`enabled`                | `bool`              | Enable setting default instance label value.                | true    | no
```
mattdurham, line 174: Why do we need a separate enabled value? Would having a value inherently make it true?

tpaschalis: AFAIK in both Prometheus and Loki, an empty value label is considered equivalent to a label that doesn't exist.

So in this case, you wouldn't really be able to have an empty instance label. I'd agree this makes sense as an attribute instead of a block. It should only override `defaultInstance` when it's expliticly supplied as a non-empty string.

hainenber: Very cool knowing the practice put in place by other parts of the ecosystem :D

I'd try implementing it that way.

hainenber: I just realized that the original requester would like to disable the behavior of setting default `instance` label too 🤦 

And having an empty string to disable the behavior would be breaking change, unless we choose a token to signify the act of disabling, like `defaultInstanceLabel="disabled"` or something, which doesn't seem right to me 🤔 

To have both the ability to disable the behavior and to set a custom default `instance` label, I think we gotta resort to block now (unless having a separate attribute :D)

thampiotr:
Closing as the linked issue is closed too.

PR #6740, opened 2024-03-20
Fix incorrect clustering beta labels

thampiotr:
#### PR Description

Clustering is now graduated to GA / stable. We have left accidentally some labels that say it's beta in ref docs.

### DIFF
--- a/docs/sources/flow/reference/components/loki.source.kubernetes.md
+++ b/docs/sources/flow/reference/components/loki.source.kubernetes.md
@@ -94,7 +94,7 @@ inside a `client` block.
 [authorization]: #authorization-block
 [oauth2]: #oauth2-block
 [tls_config]: #tls_config-block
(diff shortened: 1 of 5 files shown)


rfratto (approved):

tpaschalis:
These labeling does not apply to clustering itself, but rather on the component blocks; I'd left it here because we might opt for a different way to configure components to opt-in to clustering (eg. through an `owns` keyword like we'd discuss in the past as part of #4598).

I'm OK with removing it if you think there's a good reason for it.

thampiotr:
> I'd left it here because we might opt for a different way to configure components to opt-in to clustering 

Gotcha, thanks for the clarification. I think we can still make the `owns` or other changes in a backwards compatible way and it won't be too much effort to keep support for this current way of configuration. Will go ahead with merging this.

PR #6753, opened 2024-03-22
Reduce default sync_interval for mimir.rules.kubernetes

56quarters:
#### PR Description

Changes the default interval on which rules are synced from Mimir via `mimir.rules.kubernetes` when there are otherwise no changes to the `PrometheusRule` resources. This reduces load on the Mimir ruler APIs. Changes to `PrometheusRule` resources still result in immediately applying changes and syncing state with Mimir.

N/A

#### Notes to the Reviewer

N/A

#### PR Checklist

- [X] CHANGELOG.md updated
- [X] Documentation added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ Main (unreleased)
 - The default listen port for `otelcol.receiver.opencensus` has changed from
   4317 to 55678 to align with upstream. (@rfratto)
 
+- The default sync interval for `mimir.rules.kubernetes` has changed from `30s`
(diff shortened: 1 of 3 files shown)


clayton-cornell (approved):
LGTM

clayton-cornell:
Is this applicable only in the latest, or is this something we backport to at least v0.40?

56quarters:
> Is this applicable only in the latest, or is this something we backport to at least v0.40?

It's an optimization so I don't think it's too important to get into 0.40 as long it's the default going forward.

PR #6754, opened 2024-03-22
Upgrade agent to prometheus 2.51

### LINKED ISSUE #6580: Upgrade Agent to Prometheus 2.50
### Request

Upgrade Agent's go.mod file to the latest version of Prometheus - 2.50.

### Use case

- Keeping current with the latest Prometheus developments.
- Helps us to take advantage of the Prometheus changes implemented via #2406
- Helps us close #5905

ptodev:
#### PR Description

Upgrading to Prometheus 0.51.0.

Fixes grafana/alloy#227

#### Notes to the Reviewer

There are various TODOs in the code many of which need to be looked into. I'm also not sure how to implement [this change](https://github.com/prometheus/prometheus/pull/12738). It seems like the default list of `scrape_protocols` changes if you enable native histograms.

When I start static mode, I get an error such as this:
```
panic: failed to register "pkg.translator.prometheus.PermissiveLabelSanitization": gate is already registered
```
This seems to be because feature gate code from OTel was copied to the Prometheus codebase. I tried to resolve this by replacing...
```
import "github.com/open-telemetry/opentelemetry-collector-contrib/pkg/translator/prometheus"
```
...with...
```
"github.com/prometheus/prometheus/storage/remote/otlptranslator/prometheus"
```
...but it didn't work.

A big part of this PR is not using globally registered metrics for Prometheus Scrape and Service Discovery. The way we initialise the SD and scrape managers has changed. When testing this, it's important that we catch issues with double metric registrations and with improper unregistrations. Try this:
* Test as much relevant code in static mode and in flow mode as possible.
* Try set up multiple instances of the same Flow component, multiple instances of the static mode scrape jobs, etc.
* Try reloading the config by hitting the reload endpoint. This will tell you if unregistration was done properly.

For example, maybe something like this is a good start: [static.yml.txt](https://github.com/grafana/agent/files/14727216/static.yml.txt)

I also haven't checked if there are any other [config options](https://prometheus.io/docs/prometheus/2.45/configuration/configuration/#scrape_config) which we need to add.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [ ] Tests updated


thampiotr:
Had to use two forks:
* [opentelemetry-collector-contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib/compare/main...grafana:opentelemetry-collector-contrib:prometheusreceiver-update-to-prom-v2.51.0?expand=1) this is because prom 0.51.0 broke the methods that this package used and OTEL is still on previous prom version.
* [opentelemetry-collector](https://github.com/open-telemetry/opentelemetry-collector/compare/main...grafana:opentelemetry-collector:feature-gate-registration-error-handler) - this is to handle the duplicate featuregate registration problem.

internal/component/discovery/discovery.go
```diff
@@ -110,13 +110,36 @@ type Component struct {
...
+	// This means it's ok to do the registration now, with the initial config.
+	// This is similar ot how Prometheus does it.
+	discConfig := creator(args)
+	//TODO(ptodev): Change upstream Prometheus to have NewDiscovererMetrics outside of discovery.Config,
+	// to show that it's independent of the config.
+
+	//TODO(ptodev): It looks messy to have a refresh metrics separate from the other SD metrics.
+	// Can we make this cleaner? The reason it was made like this currently,
+	// is so that metrics exposed by Prometheus executables don't change.
+	refreshMetrics := discovery.NewRefreshMetrics(o.Registerer)
```
mattdurham, line 130: This feels ugly, so we dont have to Register the the refreshMetrics but we do need to unregister them? And then since we can get a newdiscover on the channel we have to unregister there. The fact they aren't balanced from looking at the code makes it rough.

internal/component/discovery/discovery.go
```diff
@@ -145,14 +171,25 @@ func (c *Component) Run(ctx context.Context) error {
 			c.discMut.Lock()
 			disc := c.latestDisc
 			c.discMut.Unlock()
-			go c.runDiscovery(newCtx, disc)
+			wg.Add(1)
+			go func() {
+				c.runDiscovery(newCtx, disc)
+				c.sdMetrics.Unregister()
```
mattdurham, line 177: This feels like it should break if we create the discovering multiple time in the Update.

docs/sources/flow/reference/components/prometheus.scrape.md
```diff
@@ -48,14 +48,14 @@ Name | Type | Description | Default | Required
 `forward_to`                  | `list(MetricsReceiver)` | List of receivers to send scraped metrics to. | | yes
 `job_name`                    | `string`   | The value to use for the job label if not already set. | component name | no
 `extra_metrics`               | `bool`     | Whether extra metrics should be generated for scrape targets. | `false` | no
-`enable_protobuf_negotiation` | `bool`     | Whether to enable protobuf negotiation with the client. | `false` | no
```
mattdurham, line 51: Should this go into breaking changes documentation?

thampiotr:
This is replaced now with https://github.com/grafana/alloy/pull/794 in Alloy.

PR #6763, opened 2024-03-24
feat(otelcol/extension/jaeger_remote_sampling): resync defaults with upstreams

### LINKED ISSUE #6661: Resync defaults for otelcol.extension.jaeger_remote_sampling with upstream
In our fork of the jaeger_remote_sampling extension we define the default http and grpc endpoints as 

https://github.com/grafana/agent/blob/2603bda1660254f57f22e42ac3b47ab2ef58eb6e/internal/component/otelcol/extension/jaeger_remote_sampling/internal/jaegerremotesampling/factory.go#L44-L52

whereas the upstream code either adds 0.0.0.0 or localhost as a prefix based on a feature flag.

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/a7ffded40d597a6c3e1f3e755f3cb9135778e5bc/internal/common/localhostgate/featuregate.go#L52-L58

Let's not forget to update the defaults once we update our OpenTelemetry dependency.

cc @ptodev

hainenber:
#### PR Description
As of writing, the [upstream](https://github.com/open-telemetry/opentelemetry-collector/issues/8510) has agreed upon limiting binding IP address to be `localhost` instead of `0.0.0.0` for security reasons. This PR is to reflect that upstream change.

Fixes grafana/alloy#219 

#### PR Checklist

- [x] CHANGELOG.md updated
- [ ] Documentation added
- [x] Tests updated

### DIFF
--- a/internal/component/otelcol/extension/jaeger_remote_sampling/internal/jaegerremotesampling/config_test.go
+++ b/internal/component/otelcol/extension/jaeger_remote_sampling/internal/jaegerremotesampling/config_test.go
@@ -27,9 +27,9 @@ func TestLoadConfig(t *testing.T) {
 		{
 			id: component.NewID(typeStr),
 			expected: &Config{
-				HTTPServerConfig: &confighttp.ServerConfig{Endpoint: ":5778"},
+				HTTPServerConfig: &confighttp.ServerConfig{Endpoint: "localhost:5778"},
 				GRPCServerConfig: &configgrpc.ServerConfig{NetAddr: confignet.AddrConfig{
-					Endpoint:  ":14250",
+					Endpoint:  "localhost:14250",
 					Transport: "tcp",
 				}},
 				Source: Source{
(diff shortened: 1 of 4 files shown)


internal/component/otelcol/extension/jaeger_remote_sampling/internal/jaegerremotesampling/config_test.go
```diff
@@ -27,9 +27,9 @@ func TestLoadConfig(t *testing.T) {
 		{
 			id: component.NewID(typeStr),
 			expected: &Config{
-				HTTPServerConfig: &confighttp.ServerConfig{Endpoint: ":5778"},
+				HTTPServerConfig: &confighttp.ServerConfig{Endpoint: "localhost:5778"},
```
tpaschalis, line 30: I think this might be surprising to users, as suddenly telemetry might be not be able to get pushed anymore.

Looking at the [upstream docs](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/jaegerremotesampling/README.md) I think the default is 0.0.0.0 for this reason. Can we align it with that value?

Once we start supporting OTel Collector feature gates (eg. such as `component.UseLocalHostAsDefaultHost`) then it might make sense to also support this.

rfratto: @tpaschalis If you look at [this discussion](https://github.com/open-telemetry/opentelemetry-collector/issues/8510) upstream, they're working towards changing the default to 127.0.0.1 in a future release. Currently if you enable a specific feature gate (`UseLocalHostAsDefaultHost`) then all the components default to 127.0.0.1 instead.

I think we need to figure out if we're going to do this before 1.0, since it would be a breaking change when it does get implememnted.

hainenber: Since this is pretty breaking, maybe we can park this one and accept the deviance as part of `grafana/agent` features. Not sure how we can reconcile with upstream for this one.

PR #6772, opened 2024-03-26
Finish otel upgrade

mattdurham:
#### PR Description

This partially resolves some comments from the otel upgrade. This specifically does not handle the otel kafka conversion, but that will be coming.

I did not update the changelog since functionally nothing changed.

- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/flow/reference/components/otelcol.exporter.otlphttp.md
+++ b/docs/sources/flow/reference/components/otelcol.exporter.otlphttp.md
@@ -100,6 +100,8 @@ Before enabling this option, consider whether changes to idle connection setting
 
 If `http2_ping_timeout` is unset or set to `0s`, it will default to `15s`.
 
+If `http2_read_idle_timeout` is unset or set to `0s`, then no health check will be performed.
+
 {{< docs/shared lookup="flow/reference/components/otelcol-compression-field.md" source="agent" version="<AGENT_VERSION>" >}}
 
 ### tls block
--- a/docs/sources/shared/flow/reference/components/extract-field-block.md
+++ b/docs/sources/shared/flow/reference/components/extract-field-block.md
@@ -12,13 +12,13 @@ headless: true
 
 The following attributes are supported:
 
-Name        | Type     | Description                                                                            | Default | Required
-------------|----------|----------------------------------------------------------------------------------------|---------|---------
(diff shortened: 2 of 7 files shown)


docs/sources/flow/reference/components/otelcol.exporter.otlphttp.md
```diff
@@ -177,4 +179,4 @@ Connecting some components may not be sensible or components may require further
 Refer to the linked documentation for more details.
 {{< /admonition >}}
 
-<!-- END GENERATED COMPATIBLE COMPONENTS -->
```
mattdurham, line 180: I am very confused why this is flagged has a change

clayton-cornell: It's tagged in the previous file as well. No idea. I don't see a difference.... maybe an empty line was removed? Git does odd things when empty lines are removed.

docs/sources/shared/flow/reference/components/extract-field-block.md
```diff
@@ -12,13 +12,13 @@ headless: true
...
-------------|----------|----------------------------------------------------------------------------------------|---------|---------
-`from`      | `string` | The source of the labels or annotations. Allowed values are `pod` and `namespace`.     | `pod`   | no
-`key_regex` | `string` | A regular expression used to extract a key that matches the regular expression.        | `""`    | no
-`key`       | `string` | The annotation or label name. This key must exactly match an annotation or label name. | `""`    | no
-`regex`     | `string` | An optional field used to extract a sub-string from a complex field value.             | `""`    | no
-`tag_name`  | `string` | The name of the resource attribute added to logs, metrics, or spans.                   | `""`    | no
+Name        | Type     | Description                                                                                   | Default | Required
+------------|----------|-----------------------------------------------------------------------------------------------|---------|---------
+`from`      | `string` | The source of the labels or annotations. Allowed values are `pod`, `namespace` and `node`.    | `pod`   | no
```
clayton-cornell, line 17:
```suggestion
`from`      | `string` | The source of the labels or annotations. Allowed values are `pod`, `namespace`, and `node`.    | `pod`   | no
```

wildum (approved):
lgtm, thanks for following up on this

PR #6778, opened 2024-03-27
Update kafka exporter dependency

wildum:
#### PR Description

The agent is currently using this [outdated fork](https://github.com/davidmparrott/kafka_exporter/) of a popular [kafka_exporter](https://github.com/danielqsj/kafka_exporter).

The fork that we use has not been updated with upstream for years and has a bug with the added lag metrics not being updated on metadata changes (this was reported via a customer escalation). These lag metrics are not in the upstream repo yet (see https://github.com/danielqsj/kafka_exporter/pull/195#issuecomment-1541854159).

I forked the upstream repo, added the lag metrics, fixed them to be updated on metadata changes and merged my fork to a "new" grafana repo that the agent uses with this PR.

I added an integration test that spawns a kafka cluster with a producer and a consumer. The test runs an agent with the prometheus exporter and verifies that the expected metrics are collected correctly.

Fixes grafana/alloy#464

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


CHANGELOG.md
```diff
@@ -15,11 +15,29 @@ Main (unreleased)
 - The default listen port for `otelcol.receiver.opencensus` has changed from
   4317 to 55678 to align with upstream. (@rfratto)
 
+- Remove the field `prune_interval_seconds` from the component `prometheus.exporter.kafka`. The interpolation table is now pruned
+  on `metadata_refresh_interval`. (@wildum)
+
+- The following metrics exposed by the component `prometheus.exporter.kafka` have been renamed:
+  * kafka_consumergroup_uncommitted_offsets -> kafka_consumergroup_lag
+  * kafka_consumergroupzookeeper_uncommitted_offsets_zookeeper -> kafka_consumergroupzookeeper_lag
+  * kafka_consumergroup_lag_sum -> kafka_consumergroup_lag_sum (@wildum)
```
ptodev, line 24: `kafka_consumergroup_lag_sum` is repeated twice on this line.

docs/sources/flow/reference/components/prometheus.exporter.kafka.md
```diff
@@ -27,30 +27,41 @@ prometheus.exporter.kafka "LABEL" {
...
+| `kafka_uris`                  | `array(string)` | Address array (host:port) of Kafka server.                                                                                                                                         |         | yes      |
+| `instance`                    | `string`        | The`instance`label for metrics, default is the hostname:port of the first kafka_uris. You must manually provide the instance value if there is more than one string in kafka_uris. |         | no       |
+| `use_sasl`                    | `bool`          | Connect using SASL/PLAIN.                                                                                                                                                          |         | no       |
+| `use_sasl_handshake`          | `bool`          | Only set this to false if using a non-Kafka SASL proxy.                                                                                                                            | `true`  | no       |
```
ptodev, line 35: Is the change in default value intentional? Should it be documented in the changelog?

internal/cmd/integration-tests/configs/kafka/main.go
```diff
@@ -0,0 +1,109 @@
...
+	config := sarama.NewConfig()
+	config.Producer.Return.Successes = true
+
+	var producer sarama.SyncProducer
+	var err error
+
+	for {
+		producer, err = sarama.NewSyncProducer([]string{brokerAddress}, config)
+		if err == nil {
+			break
```
ptodev, line 33: What if `NewSyncProducer` never succeeds? Does it retry indefinitely?

ptodev (commented):
Thank you for working on this! I had a quick look and noted a few things in comments.

Btw, it would have been nice if we could [document](https://grafana.com/docs/agent/latest/flow/reference/components/prometheus.exporter.kafka/) the metrics which the exporter produces. However, we don't seem to do this for any `prometheus.exporter` component, so it's ok not to do it in this PR.

internal/cmd/integration-tests/configs/kafka/main.go
wildum: yeah, I think that's okay because the producer/consumer lifetime is handled by the integration test. The tests have a timeout of 60 seconds and after that, the docker env is brought down. Whether this script retries indefinitely or sends/consumes messages indefinitely does not really matter

docs/sources/flow/reference/components/prometheus.exporter.kafka.md
wildum: This value should was actually true by default already, it was just wrongly documented (thanks for pointing it out, I did not even realised). https://github.com/grafana/agent/blob/main/internal/component/prometheus/exporter/kafka/kafka.go#L18
I think that because it's just a doc mistake we don't need to document it in the changelog

docs/sources/flow/reference/components/prometheus.exporter.kafka.md
```diff
@@ -27,30 +27,42 @@ prometheus.exporter.kafka "LABEL" {
...
+| `sasl_mechanism`              | `string`        | The SASL SCRAM SHA algorithm sha256 or sha512 as mechanism.                                                                                                                        |         | no       |
+| `sasl_disable_pafx_fast`      | `bool`          | Configure the Kerberos client to not use PA_FX_FAST.                                                                                                                               |         | no       |
+| `use_tls`                     | `bool`          | Connect using TLS.                                                                                                                                                                 |         | no       |
+| `tls_server_name`             | `string`        | Used to verify the hostname on the returned certificates unless tls.insecure-skip-tls-verify is given. The kafka server's name should be given.                                    |         | no       |
```
clayton-cornell, line 41:
```suggestion
| `tls_server_name`             | `string`        | Used to verify the hostname on the returned certificates unless tls.insecure-skip-tls-verify is given. The Kafka server's name should be given.                                    |         | no       |
```

Should or must? Should implies you don't need to... what happens if you don't provide the Kafka server name?

docs/sources/flow/reference/components/prometheus.exporter.kafka.md
```diff
@@ -27,30 +27,42 @@ prometheus.exporter.kafka "LABEL" {
...
+| `zookeeper_uris`              | `array(string)` | Address array (hosts) of zookeeper server.                                                                                                                                         |         | no       |
+| `kafka_cluster_name`          | `string`        | Kafka cluster name.                                                                                                                                                                |         | no       |
+| `metadata_refresh_interval`   | `duration`      | Metadata refresh interval.                                                                                                                                                         | `1m`    | no       |
+| `gssapi_service_name`         | `string`        | Service name when using kerberos Auth.                                                                                                                                             |         | no       |
```
clayton-cornell, line 51:
```suggestion
| `gssapi_service_name`         | `string`        | Service name when using Kerberos Authorization                                                                                                                                             |         | no       |
```

docs/sources/flow/reference/components/prometheus.exporter.kafka.md
wildum: (I did not test it locally because it's a bit annoying to set up)

internal/static/integrations/kafka_exporter/kafka_exporter.go
```diff
@@ -147,14 +188,21 @@ func New(logger log.Logger, c *Config) (integrations.Integration, error) {
 		return nil, fmt.Errorf("zookeeper lag is enabled but no zookeeper uri was provided")
 	}
 
+	// 30 is the default value
+	if c.PruneIntervalSeconds != 30 {
```
ptodev, line 192: Do we need to print the warning if PruneIntervalSeconds is set to 0?

ptodev (approved):
Thank you! Looks great. Do we need to review the exporter fork too?

wildum:
> Do we need to review the exporter fork too?

The exporter was reviewed by @gaantunes and has been merged to the official grafana fork already. If you have time you can also have a look at it if you want

internal/static/integrations/kafka_exporter/kafka_exporter.go
wildum: if PruneIntervalSeconds is different from 30 then it was explicitly set in the config by the user and that should trigger a warning (even if it is set to 0)

clayton-cornell (approved):

PR #6780, opened 2024-03-27
Update postgres exporter

### LINKED ISSUE #4749: Update github.com/prometheus-community/postgres_exporter to v0.13.2
An update for `github.com/prometheus-community/postgres_exporter` (version `v0.13.2`) is now available. Version `v0.10.0` is currently in use.

### LINKED ISSUE #6645: Update prometheus postgres exporter
### Request

The prometheus postgres exporter is on a forked version that hasn't been updated in over 2 years.
There have been [bugfixes such as these](https://github.com/prometheus-community/postgres_exporter/releases/tag/v0.13.2) that we don't have. This blocks some users. We need to update the postgres exporter to new version.

### Use case

Export metrics from postgres using an embedded exporter.

thampiotr:
#### PR Description

- New fork was created of v0.15.0, because we needed logger and configuration handling: https://github.com/prometheus-community/postgres_exporter/compare/master...grafana:postgres_exporter:exporter-package-v0.15.0
- The latest exporter is in the middle of a transition from `Exporter` to `collector` package and [both metric collection implementations run in parallel](https://github.com/prometheus-community/postgres_exporter/blob/master/cmd/postgres_exporter/main.go#L120-L140). We had to make sure that Agent starts them as well.
  - The `Exporter` style metrics work same as previous version. But the `collector` style metrics are configurable and work only with the first DSN. Had to flag this in the documentation. Also: I tried making `collector` work with multiple DSNs, but ran into issues with duplicate metric registrations, which were non-trivial to solve.
- Some metrics names changed, so flagging it in the changelog under breaking changes.
- There was a bug where multiple DSNs wouldn't work. Fixed it.
- There are tests, also tested locally with Postgres in docker.

Fixes #6645
Fixes #4749

#### Notes to the Reviewer

It wasn't actually possible to use multiple DSNs in Flow Mode previously, due to the following error, which was not possible to workaround:
image

By wiring an `instance` name, we are able to work around this problem. 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated


rfratto:
Based on our chats offline, we suspected that we'll need to eventually make a breaking change to remove multiple DSN support, since it was removed from upstream. Is that no longer true?

rfratto (approved):

clayton-cornell (approved):
No doc suggestions

thampiotr:
> Based on our chats offline, we suspected that we'll need to eventually make a breaking change to remove multiple DSN support, since it was removed from upstream. Is that no longer true?

Note for the record: It's only partially true - seems like some metrics work with multiple DSNs, others don't. See the snippet from the description:

> The latest exporter is in the middle of a transition from Exporter to collector package and [both metric collection implementations run in parallel](https://github.com/prometheus-community/postgres_exporter/blob/master/cmd/postgres_exporter/main.go#L120-L140). We had to make sure that Agent starts them as well.
> The Exporter style metrics work same as previous version. But the collector style metrics are configurable and work only with the first DSN. Had to flag this in the documentation.

PR #6786, opened 2024-03-28
flow: separate parent component and controller IDs into new labels

### LINKED ISSUE #6260: Less bloated collection of debug metrics from modules
### Request

"Debug metrics", is referring to metrics available on the Agent's `/metrics` endpoint. Those metrics indicate how various  internals are working, and are used for alerts and dashboards which monitor the state of Agent instances.

Unfortunately, Agents running with [modules](https://grafana.com/docs/agent/latest/flow/concepts/modules/) have a few issues with their metrics.

### Issue 1: The component_id labels can be very long. 

This is because the ID includes the "module" component (e.g. `module.string`) which imported that component. If a `module.string` imports a `module.string` which uses `prometheus.remote_write`, then the ID label on the metric will get quite long.

If the label is extremely long, it may even hit a limit in other systems such as Mimir. That said, by default Mimir sets its `max_label_value_length` [config parameter](https://grafana.com/docs/mimir/latest/references/configuration-parameters/#limits) to `2048` - this should be long enough for most uses.

Long component IDs can make dashboards look awkward if they want to show a component ID in a drop down or a graph legend:
* Maybe we could make those drop downs work by not using exact component names? E.g. if there is a `prometheus.remote_write` in a drop down, then the dashboards will show `prometheus.remote_write` metrics from any module.
* Alternatively, there could be separate labels for the "module path" and for the leaf component name? This would mean that there will no longer be a singe metric label with identifies a component. Losing such ID labels is not ideal because they have their own usefulness.

### Issue 2: Each Flow controller has its own set of metric series. 

If an Agent uses multiple Flow controllers, the controller metrics could bloat the `/metrics` endpoint.
To overcome the additional series, could we maybe consolidate controller functionality so that it's ran only once per process?

### Should we make debug metrics more configurable?

There might not be a "one size fits all" solution. We might have to solve this by adding some additional settings for how debug metrics should be gathered and transformed? E.g. could there be a metrics block similar to the existing [logging](https://grafana.com/docs/agent/latest/flow/reference/config-blocks/logging/) block?

### Use case

Ease of use.

tpaschalis:
#### PR Description

This PR amends the registerers and loggers we propagate into the graph to separate the so that the parent component/controller IDs are stored into a separate label.

This makes these metrics easier to read and manage and unlocks other solutions for large names in the case of nested modules in the future.

> NOTE: Please review this PR commit-by-commit and look at the extended commit messages. Each contains a snippet of how this change affects the Agent's logs and metrics

The metrics and logs examples are validated using the following example that uses nested modules.

```
--- entrypoint.river ---
logging {
	format = "json"
}

import.file "a_namespace" {
	filename = "a.river"
}

local.file "default" {
	filename = "entrypoint.river"
}

a_namespace.a "default" {
	a_argument = 47
}

--- a.river ---
declare "a" {
	import.file "b_namespace" {
		filename = "b.river"
	}

	local.file "default" {
		filename = "entrypoint.river"
	}

	argument "a_argument" {
		comment = "Where to send collected metrics."
	}

	prometheus.exporter.self "default" { }

	b_namespace.b "default" {
		b_argument = 147
	}

	export "a_export" {
		value = argument.a_argument.value + b_namespace.b.default.b_export + 1
	}
}

--- b.river ---
declare "b" {
	import.file "c_namespace" {
		filename = "c.river"
	}

	local.file "default" {
		filename = "entrypoint.river"
	}

	argument "b_argument" {
		optional = false
	}

	prometheus.exporter.self "default" { }

	c_namespace.c "default" {
		c_argument = 101
	}

	export "b_export" {
		value = argument.b_argument.value + c_namespace.c.default.c_export + 1
	}
}

--- c.river ---
declare "c" {
	argument "c_argument" {
		optional = false
	}

	local.file "default" {
		filename = "entrypoint.river"
	}

	prometheus.exporter.self "default" { }

	export "c_export" {
		value = argument.c_argument.value + 1
	}
}
```

Closes #6260

#### Notes to the Reviewer
The tracer implementation still takes the full `globals.ControllerID` since all it needs is to call path.Ext and get the final component name as the dot-style extension.

One thing to agree upon is whether path.Split makes sense for separating the parent and final IDs, or if we should add a custom helper. Currently, global IDs are currently constructed like `a_namespace.a.default/b_namespace.b.default/c_namespace.c.default/local.file.default`, i.e. _without_ a leading slash. So for the root controller, we're left with an empty label instead of `"/"` which was discussed at some point.

TODO:
- [ ] Update mixin

#### PR Checklist

- [ ] Tests updated (N/A) 
- [ ] Config converters updated (N/A)

### DIFF
--- a/internal/flow/internal/controller/node_builtin_component.go
+++ b/internal/flow/internal/controller/node_builtin_component.go
@@ -168,7 +168,7 @@ func getManagedOptions(globals ComponentGlobals, cn *BuiltinComponentNode) compo
 	parent, id := path.Split(cn.globalID)
 	return component.Options{
 		ID:     cn.globalID,
-		Logger: log.With(globals.Logger, "component", cn.globalID),
+		Logger: log.With(globals.Logger, "component_path", parent, "component_id", id),
 		Registerer: prometheus.WrapRegistererWith(prometheus.Labels{
 			"component_path": parent,
 			"component_id":   id,
--- a/internal/flow/internal/controller/node_builtin_component.go
+++ b/internal/flow/internal/controller/node_builtin_component.go
@@ -165,11 +165,13 @@ func NewBuiltinComponentNode(globals ComponentGlobals, reg component.Registratio
 
 func getManagedOptions(globals ComponentGlobals, cn *BuiltinComponentNode) component.Options {
 	cn.registry = prometheus.NewRegistry()
+	parent, id := path.Split(cn.globalID)
 	return component.Options{
 		ID:     cn.globalID,
 		Logger: log.With(globals.Logger, "component", cn.globalID),
 		Registerer: prometheus.WrapRegistererWith(prometheus.Labels{
-			"component_id": cn.globalID,
+			"component_path": parent,
+			"component_id":   id,
 		}, cn.registry),
 		Tracer: tracing.WrapTracer(globals.TraceProvider, cn.globalID),
 
--- a/internal/flow/internal/controller/node_builtin_component.go
+++ b/internal/flow/internal/controller/node_builtin_component.go
@@ -165,7 +165,7 @@ func NewBuiltinComponentNode(globals ComponentGlobals, reg component.Registratio
 
 func getManagedOptions(globals ComponentGlobals, cn *BuiltinComponentNode) component.Options {
 	cn.registry = prometheus.NewRegistry()
-	parent, id := path.Split(cn.globalID)
+	parent, id := splitPath(cn.globalID)
 	return component.Options{
 		ID:     cn.globalID,
 		Logger: log.With(globals.Logger, "component_path", parent, "component_id", id),
--- a/operations/agent-flow-mixin/dashboards/prometheus.libsonnet
+++ b/operations/agent-flow-mixin/dashboards/prometheus.libsonnet
@@ -105,13 +105,13 @@ local remoteWritePanels(y_offset) = [
     panel.withQueries([
       panel.newQuery(
         expr=|||
-          sum by (instance, component_id) (
-            prometheus_remote_storage_highest_timestamp_in_seconds{cluster="$cluster", namespace="$namespace", instance=~"$instance", component_id=~"$component"}
+          sum by (instance, component_path, component_id) (
+            prometheus_remote_storage_highest_timestamp_in_seconds{cluster="$cluster", namespace="$namespace", instance=~"$instance", component_path=~"$component_path", component_id=~"$component"}
             - ignoring(url, remote_name) group_right(instance)
-            prometheus_remote_storage_queue_highest_sent_timestamp_seconds{cluster="$cluster", namespace="$namespace", instance=~"$instance", component_id=~"$component", url=~"$url"}
+            prometheus_remote_storage_queue_highest_sent_timestamp_seconds{cluster="$cluster", namespace="$namespace", instance=~"$instance", component_path=~"$component_path", component_id=~"$component", url=~"$url"}
           )
         |||,
-        legendFormat='{{instance}} / {{component_id}}',
+        legendFormat='{{instance}} / {{component_path}} {{component_id}}',
       ),
     ])
   ),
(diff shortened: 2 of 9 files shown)


internal/flow/internal/controller/node_builtin_component.go
```diff
@@ -165,11 +165,13 @@ func NewBuiltinComponentNode(globals ComponentGlobals, reg component.Registratio
 
 func getManagedOptions(globals ComponentGlobals, cn *BuiltinComponentNode) component.Options {
 	cn.registry = prometheus.NewRegistry()
+	parent, id := path.Split(cn.globalID)
```
rfratto, line 168: Note that from the internal discussion about this, there was a mention to make `component_path` rooted to make it clearly indicate a global path:

> Global IDs start with a path separator, similarly to absolute file paths, to clearly indicate rooting.

Additionally, if there is no parent, `component_path` should be `/`: 

> component_path: the global ID of the containing subgraph for a component. If there is no containing subgraph, component_path is the rooted path /.

(This rooting logic applies to the other metrics as well wherever the labels are being split)

internal/flow/internal/controller/node_builtin_component.go
```diff
@@ -165,11 +165,13 @@ func NewBuiltinComponentNode(globals ComponentGlobals, reg component.Registratio
...
 	return component.Options{
 		ID:     cn.globalID,
-		Logger: log.With(globals.Logger, "component", cn.globalID),
+		Logger: log.With(globals.Logger, "component_path", parent, "component_id", id),
 		Registerer: prometheus.WrapRegistererWith(prometheus.Labels{
-			"component_id": cn.globalID,
+			"component_path": parent,
```
rfratto, line 173: I think the mixin will need to be updated alongside this, since `component_id` _is_ used in some dashboards (which is why we were doing this prior to 1.0 since it would be a breaking change to split out the labels after-the-fact).

tpaschalis: I've pushed a commit to include references to `component_path` eg when summing up component-level metrics for the Controller and Prometheus dashboards.

Here's how it looks like in practice. I'll check the dashboards in more environments with nested modules and more data.
Screenshot 2024-04-02 171206


Finally, the OpenTelemetry dashboard doesn't drill down to the component level so it's left as-is for now.

internal/flow/internal/controller/node_builtin_component.go
thampiotr: I do like the leading `/` too.

operations/agent-flow-mixin/grr-linux-amd64
thampiotr, line 1: Accidental commit?

thampiotr (approved):
LGTM!

operations/agent-flow-mixin/dashboards/prometheus.libsonnet
```diff
@@ -105,13 +105,13 @@ local remoteWritePanels(y_offset) = [
...
-            prometheus_remote_storage_highest_timestamp_in_seconds{cluster="$cluster", namespace="$namespace", instance=~"$instance", component_id=~"$component"}
+          sum by (instance, component_path, component_id) (
+            prometheus_remote_storage_highest_timestamp_in_seconds{cluster="$cluster", namespace="$namespace", instance=~"$instance", component_path=~"$component_path", component_id=~"$component"}
             - ignoring(url, remote_name) group_right(instance)
-            prometheus_remote_storage_queue_highest_sent_timestamp_seconds{cluster="$cluster", namespace="$namespace", instance=~"$instance", component_id=~"$component", url=~"$url"}
+            prometheus_remote_storage_queue_highest_sent_timestamp_seconds{cluster="$cluster", namespace="$namespace", instance=~"$instance", component_path=~"$component_path", component_id=~"$component", url=~"$url"}
           )
         |||,
-        legendFormat='{{instance}} / {{component_id}}',
+        legendFormat='{{instance}} / {{component_path}} {{component_id}}',
```
rfratto, line 114: nit: the full path is `{{component_path}}/{{component_id}}`, should we add the `/` character to combine them rather than using a space to make it easier to read? (Ditto with the other legends)

operations/agent-flow-mixin/dashboards/prometheus.libsonnet
```diff
@@ -409,6 +409,9 @@ local remoteWritePanels(y_offset) = [
       dashboard.newMultiTemplateVariable('component', |||
         label_values(agent_wal_samples_appended_total{cluster="$cluster", namespace="$namespace", instance=~"$instance", component_id=~"prometheus\\.remote_write\\..*"}, component_id)
       |||),
+      dashboard.newMultiTemplateVariable('component_path', |||
+        label_values(agent_wal_samples_appended_total{cluster="$cluster", namespace="$namespace", instance=~"$instance", component_id=~"prometheus\\.remote_write\\..*", component_path=~".*"}, component_path)
+      |||),
```
rfratto, line 414: nit: I think the order of these should be swapped; the order here is most broad to most specific, so based on that path goes before component ID.

rfratto (approved):
Thanks!

operations/agent-flow-mixin/dashboards/prometheus.libsonnet
tpaschalis: I tried that initially, but it looked weird for components on the root controller (`//prometheus.remote_write.default`). 
Maybe should write some jsonnet to make it pretty? 😅

rfratto: > I tried that initially, but it looked weird for components on the root controller (//prometheus.remote_write.default).
Maybe should write some jsonnet to make it pretty? 😅


🤔 Why is that what components in the root controller look like? It should be `/prometheus.remote_write.default` for rooted components, where `component_path=/`  and `component_id=prometheus.remote_write.default`

rfratto: Oh I understand now

rfratto: Sorry about that. I think it's fine for us to leave this as is and we can figure it out later. I don't think we can write Jsonnet to handle this since all the rendering is being done in Grafana at visualization time.

Marking as resolved :)

wildum (approved):

PR #6792, opened 2024-03-28
[Clustering] Filter the label "instance" from the hash computation

wildum:
#### PR Description

In clustering mode, the targets exposed by exporters are distributed amongst the agents.
For an agent to know whether it should scrape the target or not, it will compute a hash based on the labels of the target.

In https://github.com/grafana/support-escalations/issues/9923, we noticed that the agent adds a label "instance" corresponding by default to the host where the agent runs to the targets. This is a problem because if the agents are not running on the same host, they will have a different value for the label "instance". This means that they will compute different hashes and the targets won't be distributed correctly.

#### Notes to the Reviewer

Is there a scenario where a target can only be uniquely identified by the instance label?
If that's the case, then the agent would compute the same hash for different targets which would result in an unbalanced cluster.

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,8 @@ Main (unreleased)
   whenever that argument is explicitly configured. This issue only affected a
   small subset of arguments across 15 components. (@erikbaranowski, @rfratto)
 
+- Fix an issue where targets exposed by exporters were not distributed correctly between agents in clustering mode. (@wildum)
+
 ### Other changes
 
 - Clustering for Grafana Agent in Flow mode has graduated from beta to stable.
--- a/internal/component/discovery/discovery.go
+++ b/internal/component/discovery/discovery.go
@@ -52,7 +52,7 @@ func (t *DistributedTargets) Get() []Target {
 	res := make([]Target, 0, resCap)
 
 	for _, tgt := range t.targets {
-		peers, err := t.cluster.Lookup(shard.StringKey(tgt.NonMetaLabels().String()), 1, shard.OpReadWrite)
+		peers, err := t.cluster.Lookup(shard.StringKey(tgt.FilteredLabels().String()), 1, shard.OpReadWrite)
 		if err != nil {
 			// This can only fail in case we ask for more owners than the
 			// available peers. This will never happen, but in any case we fall
@@ -77,10 +77,15 @@ func (t Target) Labels() labels.Labels {
 	return lset
 }
 
-func (t Target) NonMetaLabels() labels.Labels {
+// FilteredLabels drops the label "instance" and the labels starting by MetaLabelPrefix.
+// The "instance" label is set by default to the host of the collector in exporters. If the
+// collectors are not running on the same host, they will have different default values.
+// If the targets have don't have the same set of labels between the collectors in the cluster,
+// they won't be able to correctly distribute the workload because they won't compute the same hashes.
+func (t Target) FilteredLabels() labels.Labels {
 	var lset labels.Labels
 	for k, v := range t {
-		if !strings.HasPrefix(k, model.MetaLabelPrefix) {
+		if k != "instance" && !strings.HasPrefix(k, model.MetaLabelPrefix) {
 			lset = append(lset, labels.Label{Name: k, Value: v})
 		}
 	}
--- a/internal/component/discovery/discovery_test.go
+++ b/internal/component/discovery/discovery_test.go
@@ -1,0 +1,18 @@
+package discovery
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestFilteredLabels(t *testing.T) {
+	target := Target{
+		"instance":    "instanceTest",
+		"__meta_test": "metaTest",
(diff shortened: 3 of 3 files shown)


wildum:
Converting to draft for now because we think it might be a risky change and we don't want it in 1.0

tpaschalis:
I'm not convinced this is the right solution; feels too arbitrary. One of clustering's main assumptions is that all nodes run with the same configuration files and that the upstream APIs (eg. service discovery) work in the same way so that the fully-local consistent hashing works properly. The prometheus.exporter.* components _may_ set a custom `instance` key related to the system they're running on, but this is settable.

IMHO this could be part of the documentation for running clustering together with exporter components.

wildum:
In this example the agents have the same config files, they just run on different hosts. Do you think that we should constrain the agents to run on the same host if they use clustering + prometheus exporter or should we fix the exporters so that we don't set the instance label to the host where the agent is running?
If possible I think that this "instance" label should be removed. Its behavior is not consistent amongst exporters and the agent's host should not be inside of the targets' labels. But I guess that it would be a breaking change if used in some pipeline or dashboard for filtering

ptodev:
> We noticed that the agent adds a label "instance" corresponding by default to the host where the agent runs to the targets.

Is this correct? According to [this doc page](https://prometheus.io/docs/concepts/jobs_instances/) and [this blog post](https://www.robustperception.io/controlling-the-instance-label/), the `instance` label should normally be set to the address of the scrape target.

ptodev:
I think the bug in the clustering can be fixed if SNMP exporter's `InstanceKey` [function](https://github.com/grafana/agent/blob/196ea058806a42c8b24723fde82e8dcfc5f3e919/internal/static/integrations/snmp_exporter/snmp_exporter.go#L58-L61) is changed to return the addresses of the targets which are actually being scraped, similarly to the MSSQL exporter `InstanceKey` [function](https://github.com/grafana/agent/blob/196ea058806a42c8b24723fde82e8dcfc5f3e919/internal/static/integrations/mssql/sql_exporter.go#L68-L76). I could raise a PR for this,, since I'm oncall? And then we could close this one?

wildum:
that would be a start but if we want the same behavior as this PR we need to make sure that all exporters (except the ones self, process, UNIX (maybe others?)) set the instance to the target's host and not to the agent's one

thampiotr:
After taking a closer look, I'm also not in favour of this solution, but I 100% agree we should do something to address this problem.

Thinking about this, it seems to me that `instance` is a problem because we run embedded exporters in a cluster of N collectors. This means we will run multiple instances of these exporters which may also be undesired due to increased load on systems from which metrics are exported. So I think we should provide means to run (or scrape) given exporter only on one instance at a time and then including the `instance` label should be harmless, it merely denotes which Alloy instance did the scraping. This would be accomplished by more fine-grained component scheduling. 

In case we don't want the `instance` label to denote which Alloy instance in a cluster did the scraping, we can opt to not set the `instance` label at all and [let Prometheus set it automatically](https://github.com/prometheus/prometheus/blob/9ba13de220f7b15e6b0554d7a045e4225e366e09/scrape/target.go#L535-L538). I wouldn't recommend this option, as this would be, AFAICT, an localhost network with a random port on which given exporter component exposes metrics.

To make this less error-prone, we could also check whether we are running in a cluster in exporters components and do something to warn the user?

ptodev:
@thampiotr If a `prometheus.exporter` only does work when it's scraped, then it won't be a performance issue to run them in clustered mode. I don't know how most exporters are designed in practice, but I suppose most would only do work when they are scraped. If there is only one collector which scrapes them, all will be good and there won't be a performance hit.

I think we need to go through all `prometheus.exporter` components and make sure that the `instance` label is useful, and that it also works with clustering. Maybe we can create a tracker issue for that which lists all exporter components. And we can gradually create component-specific "sub-issues" which look into each component's `instance` label.

thampiotr:
Thanks @ptodev for the discussion (offline). I'm going to summarise it here:
- The performance is not the biggest concern, but a nice-to-have. 
- We both prefer to have a solution that is general to all exporters and doesn't require going through all the exporters one-by-one. 
- With a standalone exporter and with Agent Flow Mode and Alloy without clustering, the `instance` label will be typically set to the host where the _exporter_ runs, which may be different to where the thing being exported runs (for example MySQL).
- We want to focus on fixing the clustering issue here - making the `instance` label more useful (for example, be set to the hostname of the thing being exported) can be done separately. 
- Since fine-grained component scheduling is something we will need for other use-cases, such as kubernetes events, we agree that it's a good idea to use that planned feature to address this problem.
- In the meantime, users can work around the problem quite easily by relabelling the `instance` label when using clustering to the same value in all instances in the cluster, for example:
```
discovery.relabel "replace_instance" {
  targets = discovery.file.targets.targets
  rule {
    action        = "replace"
    source_labels = ["instance"]
    target_label  = "instance"
	replacement   = "alloy-cluster"
  }  
}
```

With this, I will close this PR for now and add a note to fine-grained scheduling issue.

PR #6798, opened 2024-03-29
Refactor mimir.rules.kubernetes component

56quarters:
#### PR Description

This change refactors the `mimir.rules.kubernetes` component to move most mutable state into a separate `eventProcessor` struct. The existing code already split most behavior between the main `rules.go` file and `events.go` file. This takes the next logical step and wraps the state from `events.go` into its own struct.

This is a prerequisite for a future PR that will add clustering support to the `mimir.rules.kubernetes` component. In the future PR a particular agent instance will be the "leader" for syncing rules from Mimir or not. When not the leader, the `eventProcesor` will be a no-op.

N/A

#### Notes to the Reviewer

As discussed offline, this change and the future PR to add clustering support is a stop-gap until there's some sort of universal clustered mode support for all components. 

This has been tested with existing unit tests and locally by running Agents and Mimir in a k3s cluster via the Helm chart.

#### PR Checklist

- [ ] CHANGELOG.md updated
- [ ] Documentation added
- [X] Tests updated

### DIFF
--- a/internal/component/mimir/rules/kubernetes/events.go
+++ b/internal/component/mimir/rules/kubernetes/events.go
@@ -46,7 +46,7 @@ type eventProcessor struct {
 	logger  log.Logger
 
 	currentState    kubernetes.RuleGroupsByNamespace
-	currentStateMtx sync.Mutex
+	currentStateMtx sync.RWMutex
 }
 
 func (e *eventProcessor) run(ctx context.Context) {
@@ -237,8 +237,8 @@ func (e *eventProcessor) applyChanges(ctx context.Context, namespace string, dif
 
 // getMimirState returns the cached Mimir ruler state, rule groups indexed by Mimir namespace.
 func (e *eventProcessor) getMimirState() kubernetes.RuleGroupsByNamespace {
-	e.currentStateMtx.Lock()
-	defer e.currentStateMtx.Unlock()
+	e.currentStateMtx.RLock()
+	defer e.currentStateMtx.RUnlock()
 
 	out := make(kubernetes.RuleGroupsByNamespace, len(e.currentState))
 	for ns, groups := range e.currentState {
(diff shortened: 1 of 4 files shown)


56quarters:
> loki.rules.kubernetes is very similar to the mimir.rules.kubernetes component. Would it make sense to do the same changes for the loki component in this PR and keep the similarities between the two implementations? Or should it be done via a following PR?

Good point! My preference would be to:
* Make this change only to mimir.rules.kubernetes
* Make the change to mimir.rules.kubernetes to support clustering mode
* Make the same changes to loki.rules.kubernetes based on the above PRs

wildum:
Sounds good, would you mind creating an issue to track it and link it to this PR please?

56quarters:
> Sounds good, would you mind creating an issue to track it and link it to this PR please?

Done in https://github.com/grafana/agent/issues/6821

internal/component/mimir/rules/kubernetes/events.go
```diff
@@ -157,38 +201,71 @@ func convertCRDRuleGroupToRuleGroup(crd promv1.PrometheusRuleSpec) ([]rulefmt.Ru
...
 
 	// resync mimir state after applying changes
-	return c.syncMimir(ctx)
+	return e.syncMimir(ctx)
+}
+
+// getMimirState returns the cached Mimir ruler state, rule groups indexed by Mimir namespace.
+func (e *eventProcessor) getMimirState() kubernetes.RuleGroupsByNamespace {
+	e.currentStateMtx.Lock()
+	defer e.currentStateMtx.Unlock()
```
wildum, line 241: nit: could be an RLock

wildum (approved):
LGTM

internal/component/mimir/rules/kubernetes/events.go
56quarters: Fixed in 39d7ee98c511772dca751a6c79594ced837a9b03

wildum:
I moved this PR to Alloy: https://github.com/grafana/alloy/pull/158

PR #6803, opened 2024-04-01
This is to handle myriad ways that git revision can be used.

mattdurham:
#### PR Description

There were issues with the previous change when it came to handling branches and tags. The hard reset would cause a `cannot merge non-forwarded merge` error when pulling to see if any changes occurred.

Instead of a reset went with a checkout that worked with the pull context, though checkout needs some massaging when handling revision. Specifically the commit hash of a tag doesnt work, thankfully you can use the tag name as a branch and everything is happy. I think I have every use case covered now, HEAD, branch name, tag, and commit. Along with ensuring that if the value changes on HEAD then that will be tracked properly.

I also verified it worked using branch names from the agent-modules repo.

- [x] CHANGELOG.md updated
- [x] Tests updated

### DIFF
--- a/internal/flow/import_git_test.go
+++ b/internal/flow/import_git_test.go
@@ -28,14 +28,6 @@ func TestPullUpdating(t *testing.T) {
 	// file based git repo then committing a file, running the component, then updating the file in the repo.
 	testRepo := t.TempDir()
 
-	contents := `declare "add" {
-    argument "a" {}
-    argument "b" {}
-
-    export "sum" {
-        value = argument.a.value + argument.b.value
-    }
-}`
 	main := `
 import.git "testImport" {
 	repository = "` + testRepo + `"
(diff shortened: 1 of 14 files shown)


internal/flow/import_git_test.go
```diff
@@ -84,27 +142,178 @@ testImport.add "cc" {
...
+	go func() {
+		defer wg.Done()
+		ctrl.Run(ctx)
+	}()
+
+	// Check for final condition.
+	require.Eventually(t, func() bool {
+		export := getExport[map[string]interface{}](t, ctrl, "", "testImport.add.cc")
+		return export["sum"] == 2
+	}, 20*time.Second, 1*time.Millisecond)
```
wildum, line 287: in the file, you use in several places a pull-frequency of 10s and a timeout of 20 seconds. Could both durations be reduced?

wildum (approved):
lgtm, thanks for fixing this :)

internal/flow/import_git_test.go
mattdurham: Good call

PR #6812, opened 2024-04-02
Update log conversion for legacy file path

### LINKED ISSUE #6323: Document a migration path from Static mode to Flow for loki.source.file
### Request

Similarly to Static mode, [loki.source.file](https://grafana.com/docs/agent/latest/flow/reference/components/loki.source.file/) uses a "positions file" which tells it the latest logs that it ingested from a given file. This is useful in case of Agent restart, so that the Agent doesn't re-send all the logs from the log file to the database. Instead, the Agent can just resume where it left off before the restart.

Ideally, users should be able to stop their "static" agent and then start a new "flow" agent with an identical Flow config. However, this is not possible at the moment:
1. The location of the position file in Flow is not very configurable. It is not possible to easily point Flow to the Static mode's position file (unless a symlink is used, or unless Static mode points to the Agent Flow data directory)
1. The format of Flow's position file is not the same as Static mode's. Even if we add a feature to configure a custom path for a position file, Flow wouldn't be able to read a Static mode position file.

@tpaschalis suggested several workarounds:
* Use the `tail_from_end` argument.
* In case the logs have structured timestamps, use` loki.process` to drop logs older than X hours/days

Instead of stopping the static agent, it might be preferable to run both static and flow mode in parallel?
@tpaschalis suggested sending identical logs to Loki and letting Loki deduplicate them.
We need to clarify if this will work for all types of logs?

Alternatively, can we make it possible for Static mode to emit a 2nd positions file which is Flow compatible? Then a Flow agent could reuse it?

### Use case

Making migrating to Flow easier.

mattdurham:
#### PR Description

This automatically sets the legacy_positions_file config value, this should be safe since it will only convert if the legacy file is valid and the new positions file is empty or doesnt exist.

Closes #6323

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Tests updated

### DIFF
--- a/internal/converter/internal/promtailconvert/promtailconvert_test.go
+++ b/internal/converter/internal/promtailconvert/promtailconvert_test.go
@@ -6,88 +6,8 @@ import (
 	"github.com/grafana/agent/internal/converter/internal/promtailconvert"
 	"github.com/grafana/agent/internal/converter/internal/test_common"
 	_ "github.com/grafana/agent/internal/static/metrics/instance" // Imported to override default values via the init function.
-	"github.com/stretchr/testify/require"
 )
 
 func TestConvert(t *testing.T) {
 	test_common.TestDirectory(t, "testdata", ".yaml", true, []string{}, promtailconvert.Convert)
-}
-
-// TestGlobalPositions will ensure that legacy is set if there are multiple readers. Since the global positions file contains
-// all files but when converted only the files of the actual individual jobs should be tracked. This does lead to a case
-// where there are orphaned files but since they will not be read by the job they are noops. If they add the previously unused
-// filepath to a job, say for instance add the fun paths to fun2 then it will pick up where it left off from the conversion
-// but this is likely the best path and unlikely to be a common occurrence.
-func TestGlobalPositions(t *testing.T) {
-	test := `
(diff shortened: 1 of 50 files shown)


internal/converter/internal/promtailconvert/promtailconvert_test.go
```diff
@@ -6,8 +6,88 @@ import (
...
 func TestConvert(t *testing.T) {
 	test_common.TestDirectory(t, "testdata", ".yaml", true, []string{}, promtailconvert.Convert)
 }
+
+// TestGlobalPositions will ensure that legacy is set if there are multiple readers. Since the global positions file contains
+// all files but when converted only the files of the actual individual jobs should be tracked. This does lead to a case
+// where there are orphaned files but since they will not be read by the job they are noops. If they add the previously unused
+// filepath to a job, say for instance add the fun paths to fun2 then it will pick up where it left off from the conversion
+// but this is likely the best path and unlikely to be a common occurrence.
+func TestGlobalPositions(t *testing.T) {
```
erikbaranowski, line 21: can we do this using the existing framework for testing with a new river/yaml combo in the testdata directory?

PR #6827, opened 2024-04-05
feat(pyroscope/scrape): add support for configuring CPU profile's duration scraped by `pyroscope.scrape`.

### LINKED ISSUE #6819: Add a configuration option to pyroscope.scrape to define for how long a CPU profile should run
### Request

Right now, the only configuration option to customize scraping interval for CPU collection are `scrape_interval` and `delta`. If we want to use delta profiling, but reduce the amount of profiles we collect, there's no other options.

As the doc says 
> The seconds used will be equal to scrape_interval - 1. For example, if scrape_interval is "15s", seconds will be 14 seconds. If the HTTP endpoint is /debug/pprof/profile, then the HTTP query will become /debug/pprof/profile?seconds=14

- If we set a larger scrape interval (60s), `seconds` will be 59
- If we set a smaller scrape interval (15s), `seconds` will be 14

With this param, we are basically running profiles all the time (except for this 1s between profiles)

We could add a configuration option to let the user decide of this `seconds` value. 

for instance
```
scrape_interval = 15s
scrape_duration = 5s
```

would result in a call to `/debug/pprof/profile?seconds=5` every 15s. This would enable considerable savings on resources used by grafana-agent and pyroscope, if it's a tradeoff the user is willing to accept

### Use case

User might want to reduce the amount of CPU profiles collected in order to save on resources (CPU, network, storage...)

hainenber:

Fixes grafana/alloy#195 

#### PR Checklist

- [x] CHANGELOG.md updated
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/docs/sources/flow/reference/components/pyroscope.scrape.md
+++ b/docs/sources/flow/reference/components/pyroscope.scrape.md
@@ -76,6 +76,7 @@ Name                | Type                     | Description                    
 `params`            | `map(list(string))`      | A set of query parameters with which the target is scraped.        |                | no
 `scrape_interval`   | `duration`               | How frequently to scrape the targets of this scrape configuration. | `"15s"`        | no
 `scrape_timeout`    | `duration`               | The timeout for scraping targets of this configuration. Must be larger than `scrape_interval`. | `"18s"`        | no
+`profiling_duration`| `duration`               | The duration for a delta profiling to be scraped. Must be larger than 1 second. | `"14s"`        | no
 `scheme`            | `string`                 | The URL scheme with which to fetch metrics from targets.           | `"http"`       | no
 `bearer_token_file` | `string`                 | File containing a bearer token to authenticate with.               |                | no
 `bearer_token`      | `secret`                 | Bearer token to authenticate with.                                 |                | no
@@ -387,7 +388,8 @@ Name | Type | Description | Default | Required
 `delta` | `boolean` | Whether to scrape the profile as a delta. | `false` | no
 
 When the `delta` argument is `true`, a `seconds` query parameter is
-automatically added to requests. The `seconds` used will be equal to `scrape_interval - 1`.
+automatically added to requests. The `seconds` used will be, by default to `scrape_interval - 1`
+, or `profiling_duration` if specified. 
 
 ### clustering block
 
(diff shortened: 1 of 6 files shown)


docs/sources/flow/reference/components/pyroscope.scrape.md
```diff
@@ -387,7 +388,8 @@ Name | Type | Description | Default | Required
 `delta` | `boolean` | Whether to scrape the profile as a delta. | `false` | no
 
 When the `delta` argument is `true`, a `seconds` query parameter is
-automatically added to requests. The `seconds` used will be equal to `scrape_interval - 1`.
+automatically added to requests. The `seconds` used will be, by default to `scrape_interval - 1`
+, or `profiling_duration` if specified.
```
clayton-cornell, line 392:
```suggestion
automatically added to requests. The default value for the `seconds` query parameter is `scrape_interval - 1`. If you set `profiling_duration`, then `seconds` has the same value as `profiling_duration.  For example, if you set `scrape_interval` to `"15s"`, `seconds` defaults to `14s`. If you set `profiling_duration` to `16s`, then `seconds is set to `16s` regardless of the `scrape_interval` value.
```

This probably isn't the final wording. I'm trying to understand the logic first... is this close?

hainenber: Yes, that's the correct logic. I think your suggestion is sufficient but alas, there's a missing backtick here :D

image

clayton-cornell: I've updated the suggestion a little. I fixed the missing backtick, and adjusted the wording slightly.. and included semantic linebreaks.

docs/sources/flow/reference/components/pyroscope.scrape.md
```diff
@@ -421,8 +423,10 @@ When the `delta` argument is `false`, the [pprof][] HTTP query will be instantan
 When the `delta` argument is `true`:
 * The [pprof][] HTTP query will run for a certain amount of time.
 * A `seconds` parameter is automatically added to the HTTP request.
-* The `seconds` used will be equal to `scrape_interval - 1`.
-  For example, if `scrape_interval` is `"15s"`, `seconds` will be 14 seconds.
+* The `seconds` used, by default, will be equal to `scrape_interval - 1`
+  or to `profiling_duration` if specified.
+  For example, if `scrape_interval` is `"15s"`, `seconds` will be 14 seconds
+  If `profiling_duration` is set as `"16s"`, `seconds` will be 16 seconds regardless of `scrape_interval`
```
clayton-cornell, line 429:
```suggestion
* The default value for the `seconds` query parameter is `scrape_interval - 1`.
   If you set `profiling_duration`, then `seconds` is assigned the same value as `profiling_duration`.
   For example, if you set `scrape_interval` to `"15s"`, then `seconds` defaults to `14s`.
   If you set `profiling_duration` to `16s`, then `seconds is set to `16s` regardless of the `scrape_interval` value.
```

hainenber:
I'll need to switch the destination branch to be of `grafana/alloy` main branch and some corresponding refactors. I don't think I should be introduce new features into now deprecating `grafana/agent`. Pardon me if this feels icky on your side!

